PR #8988, opened 2021-11-30
Docs: Hiding the Hello World page

kirszenbaum:
This PR:
- Hides the Hello World page (till we migrate away from CodeSandbox)

[skip changelog]

### DIFF
--- a/docs/10.0/guides/getting-started/introduction.md
+++ b/docs/10.0/guides/getting-started/introduction.md
@@ -14,7 +14,7 @@ Welcome 👋&nbsp; to Handsontable's Guides. Here you will learn how Handsontabl
 ## Get started with sample apps
 
 <div class="row-items-container">
-    <a href="/docs/10.0/hello-world" class="row-item">
+    <a href="https://handsontable.com/demo" class="row-item">
      <img class="integration-framework-logo" src="/docs/10.0/img/pages/introduction/javascript.svg" alt="JavaScript logo" />
      <h3>JavaScript</h3>
     </a>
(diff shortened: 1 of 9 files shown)


docs/10.0/guides/getting-started/introduction.md
```diff
@@ -14,7 +14,7 @@ Welcome 👋&nbsp; to Handsontable's Guides. Here you will learn how Handsontabl
 ## Get started with sample apps
 
 <div class="row-items-container">
-    <a href="/docs/10.0/hello-world" class="row-item">
```
warpech, line 17: I think that this should not take the reader out of the docs. Rather, similarly to the framework logos, this can go to the simplest example in plain JavaScript. How about `/docs/10/binding-to-data`

warpech (changes requested):

docs/10.0/guides/getting-started/introduction.md
```diff
@@ -14,7 +14,7 @@ Welcome 👋&nbsp; to Handsontable's Guides. Here you will learn how Handsontabl
...
+    <a href="https://handsontable.com/demo" class="row-item">
```
warpech, line 17: I think that this should not take the reader out of the docs. Rather, similarly to the framework logos, this can go to the simplest example in plain JavaScript. How about `/docs/10/binding-to-data`

warpech (changes requested):

docs/10.0/guides/getting-started/introduction.md
kirszenbaum: @warpech - ok, done: /docs/binding-to-data

warpech (approved):
Looking good!

PR #8998, opened 2021-12-01
[RTL] Add right overlay

Siemienik:
WIP


Siemienik:
It is too early, for now, to start implementing this one. follow #9005

PR #9004, opened 2021-12-03
Docs: Adding HyperFormula version support section

kirszenbaum:
This PR:
- Adds a subsection that lists which versions of HyperFormula are supported by which versions of Handsontable

[skip changelog]


kirszenbaum:
@adrianszymanski89 - I [listed](https://github.com/handsontable/handsontable/blob/docs/issue-8999/docs/next/guides/formulas/formula-calculation.md#version-support) the HF versions based on what I found in package.json for each HoT version. HoT `9.0.0` has HF `0.6.2`, so I guess it doesn't work with `1.0.0` (or, if it does, we shouldn't say it's officially supported, I guess).

adrianszymanski89 (commented):
Everything looks fine, thanks.

adrianszymanski89 (approved):

Issue #9021, opened 2021-12-08
`updateSettings` causes `updatePlugin` calls for all previously-enabled plugins.


jansiegel:
### Description
When calling `updateSettings`, all the enabled plugins get their `updatePlugin` method called, regardless of their `PLUGIN_KEY` being present in the new settings.
For example, calling `updateSettings({})` on an instance with the Merge Cells plugin previously enabled will cause the plugin to disable and re-enable itself. This may result in unexpected scenarios.

Related tasks:
- #6806

### Demo
https://jsfiddle.net/js_ziggle/gt9xu3h4/

aaronbeall:
Just want to throw in my +1 for this issue. I recently updated from 7 to 11 and it seems the use of `updateSettings()` (via React props) is breaking a lot of our use cases. I get a lot of null ref errors from HT plugins after an `updateSettings()` is called.

aninde:
Thanks to the fixes in 12.0.0, when updating a previously initialized plug-in with updateSetting, there is no recalling. Updated demo https://jsfiddle.net/mvzr4yb7/.

Issue #8838, opened 2021-10-06
[8.0.0+] Nested Rows, updating settings expands all parents and not keeping the previous state ( was working in version 8+ )


dragoshzava:
### Description
When updating settings of the handsontable instance through update settings, the nested rows are expanded all the time and does not keep the persistent state of collapsed/expanded.

### Steps to reproduce
1. create a nested rows table
2. collapse all rows , collapse any row, or expand ( the action doesnt matter, it should look different from initial load/state )
3. update the instance through updateSettings ( width/height etc )
4. after update all rows are expanded by default

### Demo
1. enter this url https://jsfiddle.net/Resems/dfg8vLc4/9/
2. collapse 3 out of 4 parents for example
3. click the update settings button, which only updates the width.
4. the table gets expanded.

### Your environment
* Handsontable version: 10
* Browser Name and version: Chrome Latest
* Operating System: Windows 10

### Notes
It is working on version 8+ which we currently use, while trying to update to version 10 this seems to not work anymore.
I also tried settings persistent state, did not help.

### Additional question
Is there any way to preserve the state, maybe getting the collapsed/expanded parents and passing it on updateSettings?
In the documentation there is no information about sending which row/parent/cell should be expanded or not, by default all are expanded i see.

Thank you.

dragoshzava:
@aninde thank you, is there a release date for it ?

AMBudnik:
Hi @dragoshzava we are still making some small changes so that date isn't fixed yet. I will notify you as soon as we release the version.

AMBudnik:
Hi @dragoshzava Handsontable v12 is live!

Please check our release notes https://handsontable.com/docs/release-notes/#_12-0-0 and the migration guide https://handsontable.com/docs/migration-from-11.1-to-12.0/#step-1-verify-your-updatesettings-calls

If you would have any issues with updating to the newer version or if there are still issues to be fixed send me an email at support@handsontable.com

viotile:
Hello @AMBudnik ,
I have found that "updateSetting" method is not working with update data with version : 12.1.2 but working fine with version 7.4.0

Only added new variable and updated category name
When user click on "update setting" button then it should display like below image but with latest version (12.1.2)  
Content is not updating as you can see in image "111" should be added before 1st cell data

<img width="279" alt="image" src="https://user-images.githubusercontent.com/8554562/188663797-6b80f9e5-3a91-46e7-bf70-ba4ef3ac95c6.png">

Here is fiddle: 
[...]

Issue #7688, opened 2021-04-08
When we sort a column with a merge area while calling empty object within `updateSettings` in the `afterOnCellMouseUp` hook data is removed from the cells.


AMBudnik:
### Description

When we sort a column with a merge area while calling empty object within `updateSettings` in the `afterOnCellMouseUp` hook data is removed from the cells.

### Steps to reproduce

1. Click `B` header a couple of time to sort the column

**Result**
Data is being removed.
![sort (1)](https://user-images.githubusercontent.com/10757813/113993455-c7b2f100-9854-11eb-977a-b00aea34472e.gif)

### Demo

https://jsfiddle.net/handsoncode/gxz45kab/ - 8.3.2 - broken
https://jsfiddle.net/handsoncode/zp726jxL/ - 7.0.0 - broken

### Your environment
* Handsontable version: 8.3.2
* Browser Name and version: Chrome 89
* Operating System: Windows 10

aninde:
This issue will be fixed by v12 https://jsfiddle.net/9tk5L31y/

AMBudnik:
Thank you for fixing it!

For everyone who is tracking changes to this topic please check our release notes https://handsontable.com/docs/release-notes/#_12-0-0 and the migration guide https://handsontable.com/docs/migration-from-11.1-to-12.0/#step-1-verify-your-updatesettings-calls

Issue #7567, opened 2020-10-15
Using `setState` within `afterFilter` hook call results in the cancelation of filtering


AMBudnik:
### Description

Using `setState` within `afterFilter` hook call results in the cancelation of filtering.

### Steps to reproduce

1. Filter out any values via **filter by value** option.

**Result**
The values as still there. They are removed if you comment our the `setState` method in the `afterFilter` hook.

### Demo

https://jsfiddle.net/eudmgt63/2/

### Related issues
* https://github.com/handsontable/react-handsontable/issues/132
* (forum origin) https://forum.handsontable.com/t/call-setstate-in-after-hooks-will-not-trigger-rerender/4761

### Your environment
* React wrapper version: 4.0.0
* Handsontable version: 8.0.0
* Browser Name and version: Chrome 86
* Operating System: macOS Catalina 10.15.6

AndyStobirski:
I've encountered this bug as well. It's rather annoying and pretty much sinks any plan of mine to use HandsOnTable due to my requirements.

masry707:
@AndyStobirski I've changed the Handsontable version to 7.4.2 and it works fine with me.

Here is a working example using version 7.4.2
https://jsfiddle.net/ysx9dgLq/

AndyStobirski:
@masry707 That works perfectly. Thanks for taking the time to share this with me.

murattishkul:
@masry707 any idea how to make it to work in 9.0.0 version?

masry707:
@murattishkul sorry I have not tried it.
but you can use this [fiddle](https://jsfiddle.net/ysx9dgLq/) and install v9.0.0 and check if it works

masry707:
@murattishkul it seems not working yet. check [this](https://jsfiddle.net/03upqgmz/2/)
and I don't know how to make it work.

masry707:
@murattishkul it seems not working yet. check [this](https://jsfiddle.net/03upqgmz/2/)
and I don't know how to make it work.

AMBudnik:
Hi @murattishkul 
The @masry707 is right. This issue hasn't been scheduled for development yet.

murattishkul:
I got kinda of workaround for this
I store conditionsStack in localstorage in afterFilter hook:
`localStorage.setItem("planningOpenTableConditions", JSON.stringify(conditionsStack));`
and I apply those filters by using addCondition in afterInit hooks like this:
`let conditions = [];
    try {
[...]

PR #9025, opened 2021-12-08
Make the plugins' `updatePlugin` method get called only when the updated settings contain the plugin config keys.

jansiegel:
### Context
This PR changes the way plugins' `updatedPlugin` method reacts to `updateSettings` calls.

#### Before this change
Every `updateSettings` call (even the ones with an empty object passed as new settings) triggered the `updatePlugin` methods of all previously-enabled plugins.

#### After this change
The `updatePlugin` method gets called only if the object passed to `updateSettings` contains the main plugin key (`PLUGIN_KEY`) or additional plugin config keys (stored as an array in `SETTING_KEYS`).

Additionally, setting the `SETTING_KEYS` to:
- `true` makes the plugin update after every `updateSettings` calls
- `false` makes the plugin _never_ update on `updateSettings`

### How has this been tested?
Added new test cases.

### Types of changes

**NOT SURE YET** It can be considered a breaking change in some situations, but I'm leaning towards the fact, that it fixes many issues, and the previous behavior was _probably_ wrong.

### Related issue(s):
1. #9021 
2. [https://github.com/handsontable/handsontable/issues/7567](https://github.com/handsontable/handsontable/issues/7567)
3. [https://github.com/handsontable/handsontable/issues/7165](https://github.com/handsontable/handsontable/issues/7165)
4. [https://github.com/handsontable/handsontable/issues/3315](https://github.com/handsontable/handsontable/issues/3315)
5. [https://github.com/handsontable/handsontable/issues/8838](https://github.com/handsontable/handsontable/issues/8838)
6. [https://github.com/handsontable/handsontable/issues/7688](https://github.com/handsontable/handsontable/issues/7688)
7. https://github.com/handsontable/handsontable/issues/7546
8. https://github.com/handsontable/handsontable/issues/7541

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`

### Checklist:


handsontable/src/plugins/touchScroll/touchScroll.js
```diff
@@ -20,6 +20,10 @@ export class TouchScroll extends BasePlugin {
     return PLUGIN_PRIORITY;
   }
 
+  static get ALWAYS_UPDATE() {
```
budnix, line 23: I don't feel well about the `ALWAYS_UPDATE` option 🤔 &nbsp; sounds like a workaround. I can imagine that some of the plugins want to observe all options. Maybe, in that case, we can reuse the `CONFIG_KEYS` that would return an empty array for observing all settings and `false` for disabling observing at all? 

I just thinking aloud - Like `CONFIG_KEYS` sounds reasonable. We could support and maintain it in the future, even after implementing the "observability" to the MetaManager then `ALWAYS_UPDATE` ... I don't know :)

handsontable/src/plugins/base/base.js
```diff
@@ -211,13 +219,39 @@ export class BasePlugin {
    */
   onUpdateSettings(newSettings) {
     if (this.isEnabled) {
+      const hasEntryInSettings = ((settings) => {
```
budnix, line 222: Can you move that logic to the private method within the `Base` class? I think that IIFE obscures the body of the method as well the function expression is not necessarily created and executed each time the settings are changed.

handsontable/src/plugins/collapsibleColumns/collapsibleColumns.js
```diff
@@ -169,6 +169,11 @@ export class CollapsibleColumns extends BasePlugin {
         });
 
       } else if (Array.isArray(collapsibleColumns)) {
+
+        this.headerStateManager.mapState(() => {
```
budnix, line 173: I'm thinking about covering this change with the test. Can you do that?

budnix (changes requested):
I tested the changes and left my suggestions. I think it's a step in the good direction 👍  

The PR solves somehow the problem but there is still a problem when the developer updates the settings by calling the `updateSettings` with the same object all over again. Such like `hot.updateSettings({ copyPaste: {} })`. The abstraction does not check if the value changed. Was it intentional?

handsontable/src/plugins/__tests__/plugins.spec.js
```diff
@@ -0,0 +1,138 @@
...
+        test2: true
+      });
+      validUpdatePluginCalls += 1;
+
+      updateSettings({
+        test3: true
+      });
+      validUpdatePluginCalls += 1;
+
+      expect(updatePluginCalls).toEqual(validUpdatePluginCalls);
```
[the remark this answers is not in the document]
jansiegel: This was literally my first idea, but I changed it in the end 😆  Looks good to me both ways, though 👍

handsontable/src/plugins/copyPaste/copyPaste.js
```diff
@@ -24,6 +24,7 @@ Hooks.getSingleton().register('afterCopy');
 
 export const PLUGIN_KEY = 'copyPaste';
 export const PLUGIN_PRIORITY = 80;
+const CONFIG_KEYS = ['fragmentSelection'];
```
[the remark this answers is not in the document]
jansiegel: `copyPaste` is the `PLUGIN_KEY` which is always checked in the base plugin, `CONFIG_KEYS` are meant to be additional config keys (maybe I should change the name of that static to something like that?).

In this case, it would be more like "config keys, that can affect this plugin when updated" like `fragmentSelection` can affect `copyPaste` (it caches that value in `enablePlugin`).

handsontable/src/plugins/touchScroll/touchScroll.js
jansiegel: That makes sense, but I'd need to change [the way I designed `CONFIG_KEYS` to work ](https://github.com/handsontable/handsontable/pull/9025#discussion_r766581366). One thing less to maintain, I guess.

handsontable/src/plugins/copyPaste/copyPaste.js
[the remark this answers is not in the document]
budnix: Ah, ok. I lead myself up the garden path then. To make this more obvious maybe we should rename that option. Such as `SETTING_KEYS_DEPS` or something similar with the `DEPS` keyword?  🤔

handsontable/src/plugins/collapsibleColumns/collapsibleColumns.js
jansiegel: It looks like there already is a test case that takes care of that -> https://github.com/handsontable/handsontable/blob/292374ff029a514ef73d7e0e78fc6100d9066a9b/handsontable/src/plugins/collapsibleColumns/__tests__/collapsibleColumns.spec.js#L368

It fails without this change. Before this PR it was passing by a (sort of) mistake - the Nested Headers plugin was clearing something, that should be cleaned by the Collapsible Columns plugin as well. When not all the plugins get updated on `updateSettings`, this problem emerged by itself.

handsontable/types/plugins/base/base.d.ts
```diff
@@ -3,6 +3,9 @@ import Core from '../../core';
 export class BasePlugin {
   constructor(hotInstance: Core);
 
+  static get PLUGIN_KEY(): string;
+  static get SETTING_KEYS(): string[] | boolean;
```
budnix, line 7: Can you cover that change with TS tests? I thinking about something:
```
const pluginKey: string = BasePlugin.PLUGIN_KEY;
const settingKeys: string[] | boolean = BasePlugin.SETTING_KEYS;
```

handsontable/src/plugins/base/base.js
```diff
@@ -192,6 +207,41 @@ export class BasePlugin {
...
+    } else if (settingKeys === false) {
+      return false;
+
+    } else {
+      for (let i = 0; i < settingKeys.length; i++) {
+        if (settings[settingKeys[i]] !== void 0) {
+          return true;
+        }
+      }
+    }
```
budnix, line 242:
```suggestion
    // If SETTING_KEYS is declared as `true` -> update the plugin regardless of the settings declared in
    // `updateSettings`.
    // If SETTING_KEYS is declared as `false` -> DON'T update the plugin regardless of the settings declared in
    // `updateSettings`.
    if (typeof settingKeys === 'boolean') {
      return settingKeys;
    }

    for (let i = 0; i < settingKeys.length; i++) {
      if (settings[settingKeys[i]] !== void 0) {
        return true;
      }
    }

    return false;
```
When the `settingKeys` variable is an array and the `if`s in the `for` logic never be true the function returns `undefined`. WDYT about reshaping that lines to remove redundant `if`/`else`s?

handsontable/src/plugins/autoColumnSize/autoColumnSize.js
```diff
@@ -95,6 +95,10 @@ export class AutoColumnSize extends BasePlugin {
     return PLUGIN_PRIORITY;
   }
 
+  static get SETTING_KEYS() {
```
budnix, line 98: Shouldn't the same be added in the _AutoRowSize_ plugin?

budnix (changes requested):

budnix (approved):
Thanks! 💪

jansiegel:
@budnix, I forgot to change the changelog entry and mark it as a breaking change (as discussed earlier), could you ✅  &nbsp; it again?

This PR won't be merged until the release of `12.0`.

budnix (approved):

Issue #8823, opened 2021-10-04
Bad performance custom borders


OlgaIOn:
### Description
The tables with multiple borders are very slow, so that it is impossible to work with them. 
With a table size of 20x20, when deleting and setting borders again, the table hangs for a more than 1 minute. 

As far as I can see the problem is known for a long time  (#6467)
Do you continue to work on the problem? When can you count towards the improvements in performance? 

### Steps to reproduce
1. Set borders green (performs quickly, because there are no borders)
2. Set borders red (performs more than 1 minute, because borders should be deleted first)

### Demo
https://jsfiddle.net/ue07r83f/3/

### Your environment
* Handsontable version 10.0.0

AMBudnik:
Hi @OlgaIOn 

That issue has the `on hold` status as we decided not to use SVG. After that investigation, no further work was done to check other possibilities. We might change the title of that issue. But still, it provides a valuable list of linked issues, so we treat that epic report as the main issue related to the performance of custom borders. It means that I need to close your report as a duplicate.

I will make sure to send you a notification once we make progress on that matter. I agree with you. It definitely takes too much time to work with larger setups of custom borders.

jwueller:
This issue should at least be significantly mitigated by the #9044 performance fix.

AMBudnik:
Thank you @OlgaIOn for reporting the bug and @jwueller for your time and effort to help us to fix it. Great job!

The changes are available in Handsontable v11.1.0. Here is a full list of changes https://handsontable.com/docs/release-notes/#_11-1-0

warpech:
Note that regarding the general performance of custom borders, the issue https://github.com/handsontable/handsontable/issues/6467 remains open.

PR #6059, opened 2019-06-16
Fix rendering of custom borders to not be a recursive loop

warpech:
### Context

This PR solves the problem explained in #6052. The reason for performance drop was that each rendered piece of each border caused a re-draw of Walkontable.

### How has this been tested?

I tested manually in Chrome that there should be no regression with functionality. I added 20 new tests and 118 new test assertions to be as sure as I can.

I did not test manually in other browsers than Chrome.

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature or improvement (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Additional language file or change to the existing one (translations)

In the code, I have added a number of `TODO` comments that indicate places for potential future performance improvement.

### Related issue(s):
1. #6052

### Checklist:

- [x] My code follows the code style of this project,
- [ ] My change requires a change to the documentation.

PR #9044, opened 2021-12-16
Fix severe performance issue when removing events

jwueller:
### Context
There is a performance bug in `EventManager#clearEvents`, where every individual event being removed causes a linear search through all existing event listeners.

Obviously, this scales really poorly. Example:

A 20x20 table of 400 cells, with up to 4 borders per cell will produce 1600 listeners, not including the myriad of other, unrelated listeners that are potentially present in a table.

Listener removal iterates over all listeners, but listener removal itself is also invoked for every single listener. This means that removing the listener for a single border might result in 2560000 inner iterations for this example.

It therefore seems that this formula gives a worst-case number of inner iterations for a certain number `n` of listeners to remove:

    f(n) => n*n + f(n-1)

With our example of 1600 this results in roughly 1.4e9 inner iterations, which takes roughly a minute to execute in my test environment.

This fix only does it once per border instead, which comes out to this formula:

    f(n) => n + f(n-1)

### How has this been tested?
This issue and fix were tested by counting the number of iterations of the inner event removal loop of `EventManager#removeEventListener`.

Before the fix, setting borders took **7997414 iterations** of the event removal loop, while removing borders took **4292962614 iterations**.

After the fix, setting borders takes **2838 iterations**, while removing borders takes **2279707 iterations**.

This is still quite a significant number of iterations, but at least it seems fast enough in practice so that a complete re-engineering of the whole event architecture isn't strictly required.

### Types of changes

### Related issue(s):
This bug is directly responsible for #8823, but a lot of other issues might also be affected, due to the fundamental nature of the bug.

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`

### Checklist:


AMBudnik:
Hi @jwueller 

Thank you for sharing your proposition with all of the details explained in the description. However, the Contributing guidelines are not met. Please check the following list of requirements https://github.com/handsontable/handsontable/blob/master/CONTRIBUTING.md

jwueller:
Hi @AMBudnik,

I fixed the target branch, my bad.

However, I'm still unclear on some of the other guidelines:
* Am I supposed to commit a full build to `dist/`? I'm assuming no, but please correct me if I'm wrong.
* I didn't change any behavior, just the implementation. Does this require a test case? How does this project handle performance tests?

Could you clarify and let me know if there is anything in particular I'm missing?

warpech:
Thank you for your contribution. I can imagine that you needed to go through a deep investigation to

It is correct that you have not committed a full build to `dist/` (following point 5 of https://github.com/handsontable/handsontable/blob/master/CONTRIBUTING.md). The full builds are only updated automatically before a release.

Your question about the need for test is valid. The fact that no tests were broken might be adequate. Perhaps, some form of testing that we only do the right thing and don't do the excessive thing would be nice. But let's leave answering that for the reviewer.

warpech:
Maybe it's worth linking that the `onlyOwnEvents` param of `removeEventListener` was added in https://github.com/handsontable/handsontable/pull/6059

budnix:
Interesting PR, thanks @jwueller! Before I merge the PR I have two requests. 

First one, can you generate a changelog file and push it to the branch? Basically, you have to execute the `npm run changelog entry` [command](https://github.com/handsontable/handsontable/tree/master/.changelogs#adding-a-new-entry) in the root directory and follow the instructions. 

And second thought, It would be great to have a test in a case of regression, accidental code revert or so. I thinking about adding a test that checks if the internal `removeEventListener` method is not called too often (using jasmine spy). You can find similar tests [here](https://github.com/handsontable/handsontable/blob/master/handsontable/test/unit/EventManager.unit.js#L185).

@kirszenbaum In our [CONTRIBUTING.md](https://github.com/handsontable/handsontable/blob/master/CONTRIBUTING.md) file there is nothing about generating a changelog file. Can we update this document?

jwueller:
@budnix Thanks for your feedback, I made some additions:

* I added a changelog entry, hopefully with the information you expected to see in there.
* The added test verifies that clearing multiple events never delegates to `EventManager#removeEventListener`, because that would never scale well. This should prevent accidental performance regressions in this particular case.

budnix (approved):
Thanks!

jwueller:
@budnix Shoot, you were too fast for me. I messed up the commit author information with some outdated configuration. It's fixed in my branch, but it seems you already merged it 😄 Would it be possible to get that corrected still?

budnix:
I'm guilty 😄  Do you want me to revert the changes, then you create a new PR?

jwueller:
That's probably not necessary. Unless you do a rewrite/force push, it would still leave the incorrect information on the previous commit anyway. I guess we can just keep it the way it is now.

jwueller:
Just out of curiosity, is there a rough timeline for the next release? It would be great to know when to expect this fix to arrive in production.

budnix:
As far as I know, the next release is scheduled for mid-January. Your fix will be included in it.

Issue #9046, opened 2021-12-17
The freezing script doesn't generate the documentation.


jansiegel:
### Description
The freezing script doesn't generate new documentation directories when bumping the version with a major or minor.

warpech:
Closing as done.

PR #9047, opened 2021-12-17
Modify the freezing script to generate the docs directories.

jansiegel:
### Context
This PR:
- Adds a documentation generation logic to the freezing script.
- Modifies the documentation generation script to allow passing the version number as an argument.

<sup>[skip changelog]</sup>

### How has this been tested?
Did a faux freeze to check if the version-based docs directories get generated.

### Types of changes

- [x] Internal change

### Related issue(s):
1. #9046 

### Checklist:


Siemienik (changes requested):
Hi @jansiegel, It's amazing to review that.

I've checked it out and I see, that there is a little problem. The `semver.valid()` returns `null`(invalid) for versions without patch number (like `123.40`). It requires to provide all major, minor, and patch:
image

This makes, that the script doesn't work for our assumes, because we're maintaining the docs without patch number. Would you like to find a way to fix that, please?

Once again, many thanks for the improvement. It's quite important I think.

jansiegel:
> Hi @jansiegel, It's amazing to review that.
> 
> I've checked it out and I see, that there is a little problem. The`sever.valid()` returns `null`(invalid) for versions without patch number (like `123.40`). It requires to provide all major, minor, and patch
> 
> This makes, that the script doesn't work for our assumes, because we're maintaining the docs without patch number. Would you like to find a way to fix that, please?
> 
> Once again, many thanks for the improvement. It's quite important I think.

True, I assumed the cli-based approach would be used only with the freezing script, but it should be more lenient when it comes to the provided version numbers.
Corrected in https://github.com/handsontable/handsontable/pull/9047/commits/4bce2f375dec7196b753e43f656adaacf0b1656d.

Siemienik (approved):
LGTM! Thanks

Siemienik (approved):
LGTM! Thanks

Issue #8994, opened 2021-12-01
Docs: Rethink the Modules section


kirszenbaum:
Restructure and rewrite the [Modules](https://handsontable.com/docs/10.0/modules) page:
- Remove non-docs content (e.g. general info about the benefits of using modules)
- Clean up the terminology, keeping it consistent across the page and in line with the remaining docs (if possible)
- Reduce wordiness
- Eliminate the use of inappropriate tones of voice
- Break down large chunks of text
- List all available modules + info about their contents
- List all available module imports
- Clearly separate different content types, and modularize them into stand-alone pieces: conceptual context (e.g. why use modules) vs actionable procedures (e.g. how to import a certain module) vs reference information (e.g. the list of all available modules)

### Links
- https://handsontable.com/docs/10.0/modules

kirszenbaum:
These issues are related:
https://github.com/handsontable/handsontable/issues/8519
https://github.com/handsontable/handsontable/issues/8994

PR #9059, opened 2021-12-22
Docs: Rewrite the Modules page

kirszenbaum:
This PR:
- Rewrites the [Modules](https://handsontable.com/docs/modules/) page (fixing #8994)
- Updates all the framework module pages
- Updates the "Use modules" section of the "Performance" page to match the changes (for versions `9.0`-`next`)
- Corrects the "this documentation is not latest" messages

Affected content:
- [Modules](https://handsontable.com/docs/modules/) (versions `9.0`-`next`)
- [Modules in React](https://handsontable.com/docs/react-modules/) (versions `11.0`-`next`)
- [Modules in Angular](https://handsontable.com/docs/angular-modules/) (versions `11.0`-`next`)
- [Modules in Vue 2](https://handsontable.com/docs/vue-modules/) (versions `11.0`-`next`)
- Modules in Vue 3 (version `next`): http://localhost:8080/docs/next/vue3-modules/
- [Performance: Use modules](https://handsontable.com/docs/performance/#use-modules) (versions `9.0`-`next`)

[skip changelog]


docs/10.0/guides/advanced-topics/performance.md
```diff
@@ -72,26 +72,6 @@ See the [batch operations](@/guides/advanced-topics/batch-operations.md) page to
...
-// switch the context menu on
-new Handsontable(container, {
-  contextMenu: true,
-  // other settings
-});
-```
-
-You can also optimize the use of **moment.js**. To find out more about this topic, see the [modules page](@/guides/building-and-testing/modules.md).
+To make your grid lighter:
+- [Use Handsontable modules &#8594;](@/guides/building-and-testing/modules.md#using-modules)
```
AMBudnik, line 76: Both of those links are redirecting me to the same place in the docs
Zrzut ekranu 2022-01-5 o 11 13 31

here 

Zrzut ekranu 2022-01-5 o 11 14 40

docs/10.0/guides/building-and-testing/modules.md
```diff
@@ -9,718 +9,1079 @@ tags:
...
+| [Cell validators](@/guides/cell-functions/cell-validator.md)               | `autocompleteValidator`<br>`dateValidator`<br>`numericValidator`<br>`timeValidator`                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            | Optional            |
```
AMBudnik, line 33: On the linked source http://localhost:8080/docs/cell-validator/ there are 5 validators while in the table, we mention 4 of them. The same goes for the renderer. I think that this happens because those not included inherit from other editors/validators. But then according to http://localhost:8080/docs/cell-type/#usage `select` is only a defined `cellType` and it is missing from the table.

docs/10.0/guides/building-and-testing/modules.md
```diff
@@ -9,718 +9,1079 @@ tags:
...
+To get the most out of using Handsontable modules:
+1. [Import the required modules](#importing-required-modules).
+2. Use optional modules of your choice:
+   - [Editor modules](#using-editor-modules)
+   - [Renderer modules](#using-renderer-modules)
+   - [Validator modules](#using-validator-modules)
+   - [Cell type modules](#using-cell-type-modules)
+   - [Plugin modules](#using-plugin-modules)
+   - [Translation modules](#using-translation-modules)
+3. Remove redundant code, using [tree shaking](#tree-shaking).
```
AMBudnik, line 369: The `tree-shaking` link (point 3)

Zrzut ekranu 2022-01-5 o 11 25 07

redirects me to the top of the Modules page

AMBudnik (commented):
Hi @kirszenbaum I have some doubts (added to the code blocks).

docs/10.0/guides/advanced-topics/performance.md
kirszenbaum: @AMBudnik - you probably opened those in new tabs? That's an issue with bookmark links not opening properly in a new tab when running the docs on local server. It works in the Live environment, though.

To see this, for example:
1. Go here: https://handsontable.com/docs/api/options/#allowinvalid
2. Open the **Autocomplete strict mode** link in a new tab (it works fine).
3. Build docs locally and go here: http://localhost:8080/docs/next/api/options/#allowinvalid
4. Open the **Autocomplete strict mode** link in a new tab (it opens on the main **Autocomplete cell type** page - the bookmark doesn't work.

AMBudnik: ok, got it 👍

docs/10.0/guides/building-and-testing/modules.md
kirszenbaum: @AMBudnik - same as [here](https://github.com/handsontable/handsontable/pull/9059#discussion_r778718958) - will work on Live.

docs/10.0/guides/building-and-testing/modules.md
kirszenbaum: @AMBudnik - right, the tables include only modules mentioned [here](https://handsontable.com/docs/modules/#modules-cheatsheet).

All in all, the following elements seem to be hidden somewhere in the imports (and I can't find them in the source code):
- `Handsontable.renderers.DropdownRenderer` (mentioned [here](https://handsontable.com/docs/cell-renderer/#cell-renderer))
- `Handsontable.renderers.DateRenderer` (mentioned [here](https://handsontable.com/docs/cell-renderer/#cell-renderer))
- `Handsontable.renderers.TimeRenderer` (mentioned [here](https://handsontable.com/docs/cell-renderer/#cell-renderer))
- `Handsontable.validators.DropdownValidator` (mentioned [here](https://handsontable.com/docs/cell-validator/#cell-validator))
- `Handsontable.cellTypes.select`(mentioned [here](https://handsontable.com/docs/cell-type/#available-cell-types))

@budnix - perhaps you can help us out: how do you import the elements above? I suppose they're included in some of the modules, but I can't locate them. Thanks!

budnix: @kirszenbaum The mentioned "modules" are not really modules in terms of the modules released a year ago (8.3.0). They can not be imported because physically they do not exist. 

For example, the `dropdown` cell type (and others) underneath uses the renderer and validator from the other existing modules. Probably for the legacy support all registered cell types, renderers, validators, and editors including that aliased one are exported to the `Handsontable` global variable. And only there they are available.

For example for the Dropdown menu, the exported renderer [is the same](https://github.com/handsontable/handsontable/blob/master/handsontable/src/cellTypes/dropdownType/dropdownType.js#L10) as for Autocomplete
```js
Handsontable.renderers.DropdownRenderer === Handsontable.renderers.AutocompleteRenderer // true
```

About the `Handsontable.cellTypes.select` type I think this is a mistake. This cell type never esist.

Generally, the list of all modules can be found by listing the directories within each exported namespaces:
 * https://github.com/handsontable/handsontable/tree/master/handsontable/src/cellTypes;
 * https://github.com/handsontable/handsontable/tree/master/handsontable/src/editors;
 * https://github.com/handsontable/handsontable/tree/master/handsontable/src/plugins;
 * https://github.com/handsontable/handsontable/tree/master/handsontable/src/renderers;
 * https://github.com/handsontable/handsontable/tree/master/handsontable/src/validators;

kirszenbaum: Thanks @budnix ! @AMBudnik - it seems that the new Modules page correctly lists all modules that can be imported, but there are discrepancies on the other side, on the following pages:

https://handsontable.com/docs/cell-renderer/#cell-renderer
https://handsontable.com/docs/cell-validator/#cell-validator
https://handsontable.com/docs/cell-type/#available-cell-types
https://handsontable.com/docs/select-cell-type/

I created a separate issue to deal with those: #9087

AMBudnik: Thank you for investigating this case

AMBudnik:
Nope, @kirszenbaum at the moment no new ideas. 👍  from me to go

kirszenbaum:
@warpech - here are the [Modules](https://handsontable.com/docs/modules/) page changes that I mentioned, the Customer Success team already had a look at it. Please let me know what you think, thanks!

docs/10.0/guides/advanced-topics/performance.md
warpech: I agree that we could make it slightly clearer by extending the text

> To make your grid lighter

Into something with more context.

> To reduce the bundle size and JavaScript parsing time, we advise to import only the [modules](link) of Handsontable that you actually use, instead of the complete package.

-----

I think that the `moment.js` tip should be moved from the "Modules" page to the "Date cell type" page: https://handsontable.com/docs/date-cell-type/ @budnix do you agree?

docs/next/guides/building-and-testing/modules.md
```diff
@@ -481,352 +275,822 @@ new Handsontable(container, {});
...
+
+#### Step 1: Remove the full Handsontable import
+
+When using modules, you no longer need to import the full version of Handsontable.
+
+Remove the `handsontable` import:
+
+```js
+// remove this line
+import Handsontable from 'handsontable';
```
warpech, line 393: How do you now that I have this line? You could as well assume that I have blank document.

I think that it could be explained all in the next step by adding a paragraph:

> Note that to use modules, you must import `Handsontable` from **handsontable/base**, not from **handsontable** (which would give you a pre-made [package](link)).

docs/next/guides/building-and-testing/modules.md
```diff
@@ -481,352 +275,822 @@ new Handsontable(container, {});
...
 
-To use a specific editor you need to import it alongside the registering method. For instance, let's try adding the password editor.
+```js
+// import the `handsontable/base` module
+import Handsontable from 'handsontable/base';
+```
+
+#### Step 3: Import the CSS
+
+Handsontable's CSS file is not modularized, so you need to import it:
```
warpech, line 407: The above sentence is incomplete:

> Handsontable's CSS file is not modularized, so you need to import it **as a whole**, in the same way as presented on the [Installation](link) page.

docs/next/guides/building-and-testing/modules.md
```diff
@@ -481,352 +275,822 @@ new Handsontable(container, {});
...
+
+```js
+import Handsontable from 'handsontable/base';
+
+// import Handsontable's CSS
+import 'handsontable/dist/handsontable.full.css';
```
warpech, line 413: @kirszenbaum the above line is not standard JavaScript. In standard JavaScript, you would not be able to import a CSS file. The above line only works thanks to some bundlers (e.g. Webpack), but we don't know here what bundler is used by a developer.

However, the above remark should be solved on the "Installation" page, because the same mistake can be seen there: https://handsontable.com/docs/installation/#download-and-install-the-library

Because CSS is not modularized, I think that it does not even deserve a separate step here. The contents of this step should be merged with the previous step.

docs/next/guides/building-and-testing/modules.md
```diff
@@ -481,352 +275,822 @@ new Handsontable(container, {});
...
+First, make sure to [import the required modules](#importing-required-modules):
 
 ```js
-// remember to have the base imported
+// import the `handsontable/base` module
 import Handsontable from 'handsontable/base';
-// import the method for registering an editor
-// import the PasswordEditor
+
+// import Handsontable's CSS
```
warpech, line 445: In the spirit of my previous remark, change this to:

```
// import Handsontable's CSS if your bundler allows it
```

docs/next/guides/building-and-testing/modules.md
```diff
@@ -481,352 +275,822 @@ new Handsontable(container, {});
...
+import 'handsontable/dist/handsontable.full.css';
+```
+
+#### Step 2: Import the editors' registering method
+
+Import the `registerEditor()` method that lets you register your chosen editor:
```
warpech, line 451: This is a pleonasm. I think that we owe the reader any explanation to better understand what's going on:

> Handsontable will only recognize the editor module that you've imported, if you provide it to a registering function.

Also, I think that there are too many steps. See my comment on line 12 of `vue-modules.md`.

docs/next/guides/building-and-testing/modules.md
```diff
@@ -264,32 +57,32 @@ import {
   PasswordEditor,
   SelectEditor,
   TextEditor,
+  registerEditor, // editors' registering method
 } from 'handsontable/editors';
```
warpech, line 61: Could we also show that the following works:

```
import * from 'handsontable/editors'
```

@budnix if that's correct?

docs/next/guides/building-and-testing/modules.md
```diff
@@ -481,352 +275,822 @@ new Handsontable(container, {});
...
+
+// register the `password` editor
 registerEditor(PasswordEditor);
 ```
 
-Now, you can use `PasswordEditor`, the full example looks like this:
+::: tip
+#### Registering all editors at once
+
+If you import all of Handsonable's editor modules, register them all with the `registerAllEditors()` method:
```
warpech, line 489: Why would I want to do that? I think that it might be useful temporarily to make development easier.

But this tip does not say how to import all editors. Do I need to put many lines of code to import all editors by name?

@budnix would the following work?

```js
import * from 'handsontable/editors'
import { registerAllEditors } from 'handsontable/registry';
registerAllEditors();
```

docs/next/guides/building-and-testing/modules.md
```diff
@@ -481,352 +275,822 @@ new Handsontable(container, {});
...
-You can quickly import and register all editors at once:
+#### Step 1: Import the required modules
+
+First, make sure to [import the required modules](#importing-required-modules):
 
 ```js
-// import the `registerAllEditors()` function
-import { registerAllEditors } from 'handsontable/registry';
+// import the `handsontable/base` module
+import Handsontable from 'handsontable/base';
```
warpech, line 544: I don't see how a comment like the one on line 543 is useful. I think that obvious things should not be stated in code comments.

docs/next/guides/building-and-testing/modules.md
```diff
@@ -481,352 +275,822 @@ new Handsontable(container, {});
...
-To use a specific renderer you need to import it alongside the registering method. For instance, let's try adding the autocomplete renderer.
+Import the `registerRenderer()` method that lets you register your chosen renderer:
 
-Start with importing the base, `autocompleteRenderer` and the `registerRenderer` method.
+```js
+import Handsontable from 'handsontable/base';
+import 'handsontable/dist/handsontable.full.css';
+
+// import the `registerRenderer()` method
+import { registerRenderer } from 'handsontable/renderers';
```
warpech, line 559: I don't see how a comment like the one on line 558 is useful. Obvious things should not be stated in code comments.

docs/next/guides/building-and-testing/modules.md
```diff
@@ -264,32 +57,32 @@ import {
...
 import {
-  registerRenderer,
   baseRenderer,
   autocompleteRenderer,
   checkboxRenderer,
   htmlRenderer,
   numericRenderer,
   passwordRenderer,
   textRenderer,
+  registerRenderer, // renderers' registering method
```
warpech, line 72: In this whole document, `registering method` should be replaced with `registering function`. Because these are functions, not methods.

There might be more mistakenly used instances of the word "method" throughout the docs.

See: https://flaviocopes.com/javascript-difference-method-function/

docs/next/guides/building-and-testing/modules.md
```diff
@@ -481,352 +275,822 @@ new Handsontable(container, {});
...
+| `DropdownEditor`     | `dropdown`        |
+| `HandsontableEditor` | `handsontable`    |
+| `NumericEditor`      | `numeric`         |
+| `PasswordEditor`     | `password`        |
+| `SelectEditor`       | `select`          |
+| `TextEditor`         | `text`            |
+
+To start using an editor module, follow the steps below.
+
+#### Step 1: Import the required modules
```
warpech, line 437: I don't think that explaining this in steps is useful. Can you convince me? I just think that it takes twice the amount of screen space and adds unnecessary repetition.

This is the same remark as on line 12 of `vue-modules.md`.

docs/next/guides/building-and-testing/modules.md
```diff
@@ -481,352 +275,822 @@ new Handsontable(container, {});
...
   type: 'date',
 });
 ```
+:::
 
-## Using modules with frameworks
+### Using modules with framework wrappers
 
-You can also use modules with Handsontable's framework wrappers:
+You can use modules with Handsontable's framework wrappers:
```
warpech, line 1081: How about:

> If you are using one of the Handsontable's framework wrappers, you might find additional information for your framework on the following pages:

docs/next/guides/building-and-testing/modules.md
```diff
@@ -481,352 +275,822 @@ new Handsontable(container, {});
...
 - [Using modules with React](@/guides/integrate-with-react/react-modules.md)
 - [Using modules with Angular](@/guides/integrate-with-angular/angular-modules.md)
 - [Using modules with Vue 2](@/guides/integrate-with-vue/vue-modules.md)
 - [Using modules with Vue 3](@/guides/integrate-with-vue3/vue3-modules.md)
+
+### Tree shaking
```
warpech, line 1087: I don't see the point of this section. I think that this section should be removed.

Do we actually support all of the bundlers listed below? I think that we only test with Webpack and ParcelJS (see:  https://docs.google.com/spreadsheets/d/1MEe4tF2xseu6QI3N6xC4FcIevG_p0q2uWZt0JpURJfk/edit#gid=0)

That said, I think that this page is only relevant for developers who use bundlers. Such information is missing on this page. I think that we should add somewhere close to the top:

> Loading Handsontable by module requires an environment that supports the `import` syntax. We actively test Handsontable using Webpack and ParcelJS bundlers that provide such ability. If the methods presented on this page don't work with your bundler, please report it as a bug. Direct use of ES Modules is not supported at the moment.

@budnix is the above text correct?

warpech (changes requested):
There goes my review. 

@kirszenbaum good job with the changes, however I think that some of them were in good direction while others not. Let's discuss below.

docs/next/guides/building-and-testing/modules.md
```diff
@@ -9,251 +9,44 @@ tags:
```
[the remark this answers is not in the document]
budnix: I agree. As the "Building and Testing" section is for developers who want to make custom build or experiment with source code, that "Modules" page rather suits to developers who want to consume the public modules API to achieve better performance and package weight optimization.

docs/next/guides/building-and-testing/modules.md
```diff
@@ -481,352 +275,822 @@ new Handsontable(container, {});
...
+// set the `language` configuration option
 new Handsontable(container, {
   language: 'pl-PL',
-// rest of the settings
 });
 ```
 
-And that is all! You can use the PL-pl translation!
+:::tip
+#### Optimizing Moment.js locales
```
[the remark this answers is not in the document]
budnix: The `moment.js` is not only used within the Date cell type but also as a sort comparator in the _ColumnSorting_ plugin and as condition functions (e.g. "After", "Between" ...) for date column types in the _Filters_ plugin. 

Maybe it would be some idea to move it to the new section about performance or optimization and link to it 🤔

docs/next/guides/building-and-testing/modules.md
budnix: Your proposition is not valid ES syntax. If you meant:
```js
import * as editors from 'handsontable/editors';
```
then it's possible but notice that the import does not automatically register the editors. You have to grab the editor and register it using the proper API, for example  `editors.registerEditor(editors.NumericEditor);`. IMO the usage of the API this way is not super elegant. 

If you meant:
```js
import 'handsontable/editors';
```
then this would not work. It wouldn't even work if the editors were registering themselves. The bundler while transpiling the code would remove this import (tree-shaking) causing unexpected results in prod and/or dev builds. That's why there is a requirement to import component and register it manually.

docs/next/guides/building-and-testing/modules.md
budnix: I think that `registerAllEditors();` (and `registerAllPlugins();` etc.) functions can be useful for more lazy developers who don't want to selectively choose components or want to make sure that always there will get all cell types. That functions underneath imports and register all components. So in a case when in the new Handsontable release we add new cell type having the `registerAllCellTypes()` would provide new type automatically. 

About the example, I would change to:
```js
import { registerAllEditors } from 'handsontable/editors';

registerAllEditors();
```

docs/next/guides/building-and-testing/modules.md
budnix: The Handsontable distribution package is compatible with CommonJS and ES. I tested Handsontable with Webpack {2,3,4,5}, Parcel {1,2}, Rollup {1,2}, Vite and didn't find any issues. I would say that the text is partially correct. We do not test the code within any bundlers. We build a UMD package with Webpack 4 and run the E2E tests on it but that's it.

docs/10.0/guides/advanced-topics/performance.md
```diff
@@ -72,26 +72,6 @@ See the [batch operations](@/guides/advanced-topics/batch-operations.md) page to
...
-
-// switch the context menu on
-new Handsontable(container, {
-  contextMenu: true,
-  // other settings
-});
-```
-
-You can also optimize the use of **moment.js**. To find out more about this topic, see the [modules page](@/guides/building-and-testing/modules.md).
+To make your grid lighter:
```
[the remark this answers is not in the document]
kirszenbaum: @warpech - done: 6d09a74 88ba4bb

docs/next/guides/building-and-testing/modules.md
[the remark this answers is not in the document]
kirszenbaum: @warpech IMO:
- The **Getting Started** section should only cover the most basic stuff, to show how easy it is to quickly get Handsontable going. I wouldn't add **Modules** there unless we really want to push using modules as much as possible
- The **Advanced topics** section seems like a good place, but the section's title is vague (we should avoid such obscure titles).

How about we rename the **Advanced topics** section to **Optimization** (or similar), so it looks like this:
- **Optimization**
    - **Batch operations**
    - **Performance**
    - **Packages** 
    - **Modules**

docs/next/guides/building-and-testing/modules.md
kirszenbaum: @warpech - to be honest, using conditionals, continuous tenses and modal verbs in docs is usually a no-no (unless necessary). I'm not saying we should strictly stick to [STE](https://en.wikipedia.org/wiki/Simplified_Technical_English) or sth similar, but I believe simplicity is more important than likeability of tone. My proposition:

> To find out how to use modules with Handsontable's framework wrappers, see the following guides:

Of course, it's ultimately an arbitrary style guide decision (and we don't have a style guide), so it's up to you :)

docs/next/guides/building-and-testing/modules.md
```diff
@@ -9,251 +9,44 @@ tags:
...
 ## About modules
 
-The full bundle of Handsontable consists of multiple built-in modules such as plugins, renderers, editors, validators, and cell-types. It also includes the whole rendering functionality and advanced data management. This is a compact tool with plenty of options available.
+Modules let you pick and choose only those parts of Handsontable that you actually need. This can significantly reduce the size of your app.
 
-When you become familiar with Handsontable functionalities you may find out that some parts of it are essential in your application and some of them stay idle most of the time.
+Read more:
+- [JavaScript modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules)
+- [Modularizing to improve the developer experience](https://handsontable.com/blog/articles/2021/2/modularizing-to-improve-the-developer-experience)
+- [Handsontable 11.0.0: modularization for React, Angular, and Vue](https://handsontable.com/blog/articles/2021/11/handsontable-11.0.0-modularization-for-react-angular-and-vue)
```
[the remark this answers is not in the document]
kirszenbaum: All outside links are marked by an icon:


But sure, we expand that info. I'd just do it the same way in the entire documentation then

docs/next/guides/building-and-testing/modules.md
kirszenbaum: Not sure I understand, I don't see a pleonasm in this structure:

> Import the `[name]` function that lets you do [sth].

Just because a function/method name **seems** self-explanatory, doesn't mean that it is: that function could do sth entirely else than the name suggests (as sometimes is the case in HoT).

How about:

> Import the `registerEditor()` function that lets Handsontable recognize your imported editor module.

docs/next/guides/building-and-testing/modules.md
kirszenbaum: TL;DR
- The purpose of a "guide" is to take the user through a standard procedure that results in a specific outcome. And procedures consist of numbered steps.
- Additionally, numbered steps suggest a standard way of using a feature.
- Additionally, it's better to underestimate your users' background knowledge than overestimate it.

---

Here's where I'm coming from:

First, what I'm trying to do here is apply some content management rules. In a content management methodology (be it [topic-based writing](https://en.wikipedia.org/wiki/Topic-based_authoring) or [modular writing](https://writing.stackexchange.com/questions/37891/topic-based-authoring-vs-modular-authoring)), the first thing is to determine elementary content types. A classic example is [DITA](https://en.wikipedia.org/wiki/Darwin_Information_Typing_Architecture), where some of the basic content types are:
- [Reference topic](https://docs.oasis-open.org/dita/v1.2/os/spec/archSpec/dita_reference_topic.html)
- [Concept topic](https://docs.oasis-open.org/dita/dita/v1.3/os/part2-tech-content/archSpec/technicalContent/dita-concept-topic.html)
- [Task topic](https://heretto.com/how-to-write-a-good-dita-task-topic/)

Don't get me wrong: I'm not trying to apply DITA rules here - it's just that docs are not a blog meant to be read: they are independent, interlinked pieces of content that are meant to be:
- Quickly looked up (reference)
- Scanned for a broader understanding (concept)
- Followed,  to achieve a specific outcome (tasks, or "guides" if you will).

Since our documentation is not based on any specific methodology, I'm trying to at least apply a standard structure that properly separates different content types that happen to be (for better or worse) on the same page, for example:
- Section (with a unique name)
  - Page (with a unique title)
    - Short introduction (max 1 sentence)
    - Necessary conceptual information
    - Tasks/guides/procedures:
      - Achieving an outcome no. 1
        - Step 1
        - Step 2
        - Step 3
        - Outcome
      - Achieving an outcome no. 2
        - Step 1
        - Step 2
        - Step 3
        - Outcome

If an aim of a piece of information is to make the reader achieve a certain outcome, it's a task. Tasks cover procedures. Procedures, by definition, are made of numbered steps, to ensure the proper outcome and limit the number of possible applications to just one that we can easily test. And finally: each step covers just one action (ideally).

Hence the need to structure the content in a predictible, replicable (ideally: reusable - e.g. in a [CCMS](https://en.wikipedia.org/wiki/Component_content_management_system), each "duplicate" step would be a reusable snippet) manner, which is what I'm trying to enforce here.

Presenting a [loosely arranged group of headings](https://gist.github.com/warpech/f9a712ce507708c5eeb6ccdfb97d4609) goes the opposite direction.

Additionally, I treat editor modules as a feature, and numbered steps suggest a standard way of using such feature. And if we **can** suggest a standard way of using a feature, IMO we should, because:
- It provides guidance to inexperienced developers
- It suggests the best way to use a feature (e.g. performance-wise)
- A specific scenario is easier to test than a general scenario or no scenario at all

Additionally, it's better to underestimate your users' background knowledge than overestimate it. IMO our documentation should:
- Underestimate the user's command of English
- Underestimate the user's command of JavaScript (and/or frameworks)
- Underestimate the user's familiarity with Handsontable
- Underestimate the user's familiarity with any of the other Handsontable documentation pages (it's called "[every page is page one](https://www.stc.org/notebook/2013/10/01/every-page-is-page-one/)" rule)

Of course, this is all debatable and ultimately up to you :)

docs/next/guides/advanced-topics/performance.md
```diff
@@ -72,26 +72,6 @@ See the [batch operations](@/guides/advanced-topics/batch-operations.md) page to
...
-
-// switch the context menu on
-new Handsontable(container, {
-  contextMenu: true,
-  // other settings
-});
-```
-
-You can also optimize the use of **moment.js**. To find out more about this topic, see the [modules page](@/guides/building-and-testing/modules.md).
+To make your grid lighter:
```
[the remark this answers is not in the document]
kirszenbaum: Re: "lighter": good point, changing: 1c8e419

docs/next/guides/building-and-testing/modules.md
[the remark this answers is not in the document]
kirszenbaum: If I get it right, Moment.js is included in `handsontable/base`, right? This means that everone who cares about Handsontable's impact on their app should optimize Moment.js locales, no matter if they use the `date` cell type or not.

How about we add it to the **Performance** page then? (it follows the structure I suggested in [this comment](https://github.com/handsontable/handsontable/pull/9059#discussion_r787667657))
- Optimization
    - Batch operations
    - Performance
      - Optimizing Moment.js
    - Packages
    - Modules

Unless we just don't want to mention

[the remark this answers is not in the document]
budnix: > If I get it right, Moment.js is included in handsontable/base, right?

Good point. The lib is used in the Core (so including `handsontable/base`) for checking if the license key is valid.

docs/next/guides/building-and-testing/modules.md
```diff
@@ -9,251 +9,44 @@ tags:
...
 
-To reduce the size of your app, you can use Handsontable by importing individual [modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules).
+To reduce the size of your app, use Handsontable modules.
 
 [[toc]]
 
 ## About modules
 
-The full bundle of Handsontable consists of multiple built-in modules such as plugins, renderers, editors, validators, and cell-types. It also includes the whole rendering functionality and advanced data management. This is a compact tool with plenty of options available.
+Modules let you pick and choose only those parts of Handsontable that you actually need. This can significantly reduce the size of your app.
```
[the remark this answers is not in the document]
kirszenbaum: @warpech - if by "old text" you mean this:



Then I'm not sure I agree - perhaps, it could be good material for a blog post, but not for technical documentation.

IMO this page should:
- Give a **brief** explanation of what Handsontable's modules are and why you should use them  ([conceptual information](https://github.com/handsontable/handsontable/pull/9059#discussion_r788794087)) (without trying to explain what modules are in general and how great that feature is, because that's off-topic – instead, point the user to other resources)
- Present a clear list of all available modules ([reference information](https://github.com/handsontable/handsontable/pull/9059#discussion_r788794087))
- Guide the user through the process of importing, registering and using modules of all available types ([task information](https://github.com/handsontable/handsontable/pull/9059#discussion_r788794087))

If we want to expand the **About modules** section but stick to actual documentation content, we could do it more or less like this (fixing the distribution/bundle/package discrepancy on the way):

# About modules

Handsontable is a comprehensive tool with a broad range of features. If you don't need all of them, you can pick only those parts that you actually use, and get rid of the rest (e.g. unnecessary translations). This approach reduces Handsontable's impact on the overall size of your app.

To make this possible, Handsontable is divided into [modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules). The [full distribution/bundle/package](https://handsontable.com/docs/packages/#full-distribution-recommended) of Handsontable contains all the available modules, but you can decide to import only the ones that you actually use.

Learn more about JavaScript modules:
- MDN Web Docs: [JavaScript modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules)
- Handsontable blog: [Modularizing to improve the developer experience](https://handsontable.com/blog/articles/2021/2/modularizing-to-improve-the-developer-experience)
- Handsontable blog: [Modularization for React, Angular, and Vue](https://handsontable.com/blog/articles/2021/11/handsontable-11.0.0-modularization-for-react-angular-and-vue)

docs/next/guides/building-and-testing/modules.md
kirszenbaum: Ok: ef4409c c56f46a

docs/next/guides/building-and-testing/modules.md
kirszenbaum: Ok: ef4409c c56f46a

docs/next/guides/building-and-testing/modules.md
kirszenbaum: The assumption that the user already imported the full HoT version (and now wants to optimize it) is the same as [before](https://handsontable.com/docs/modules/#how-to-use-modules), but ok: ef4409c c56f46a

I'm not so sure about the **pre-made package** term, though. It's not used anywhere else in the docs. Can we choose one of the terms below and consistently stick to it throughout the documentation?
- The full bundle of Handsontable
- The full package of Handsontable
- The full distribution of Handsontable
- The full version of Handsontable

docs/next/guides/building-and-testing/modules.md
kirszenbaum: Ok: ef4409c c56f46a

docs/next/guides/building-and-testing/modules.md
kirszenbaum: I updated the examples, and added links to the list of all modules: 4433c0f

docs/next/guides/building-and-testing/modules.md
kirszenbaum: 1) It points at the line that we're currently describing or focusing on.
2) It matches the pattern of other code examples (usually longer ones, though) in the docs, where comments like this point you to the currently-described line ([for example](https://handsontable.com/docs/column-summary/#step-1-enable-the-columnsummary-plugin) – some docs do this by tooltips as well, [for example](https://stripe.com/docs/connect/collect-then-transfer-guide?platform=web#create-an-express-account-and-prefill-information)).
3) Generally: what may seem obvious to us, may not be obvious to the recipient reaching for a detailed tutorial.

But it's your call, I'm removing them: bc4bed1

docs/next/guides/building-and-testing/modules.md
kirszenbaum: Removing, as commented [here](https://github.com/handsontable/handsontable/pull/9059#discussion_r789751566)

docs/next/guides/building-and-testing/modules.md
kirszenbaum: Ok, done: 8f61684

docs/next/guides/building-and-testing/modules.md
kirszenbaum: If we **can** present a comprehensive list of all officially supported bundlers, I'd do it in a table, like with [HyperFormula version support](https://handsontable.com/docs/formula-calculation/#hyperformula-version-support).

If we can't (for the lack of regular testing), I'd do it like [this](https://github.com/handsontable/handsontable/pull/9059/commits/36fbba274c0e9e62fc0bab9472babba95beed578#diff-3b9c74dc2c4138891da8b067a2113b8d2aa6bf1b2ff3e1244a4b71923fbec14eR27-R35).

docs/next/guides/building-and-testing/modules.md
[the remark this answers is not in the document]
warpech: This sounds good:

- **Optimization**
   - **Batch operations**
   - **Performance**
   
 But **Packages** and **Modules** don't fit here. This must be a section, in which we can also add **TypeScript** or pages about bundlers (Webpack, Rollup, etc).
 
Maybe the current section is OK, but it needs to be renamed and rearranged, like this:

- **Building and Tooling**
   - **Packages**
   - **Modules**
   - **Plugin development**
   - **Custom builds**
   - **Testing**
   - **Folder structure**

docs/next/guides/building-and-testing/modules.md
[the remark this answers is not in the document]
warpech: My suggestion then is to limit the "Performance" page to run-time performance.

We can create a new page "Bundle size", where there will be two sections "Use modules" and "Optimizing Moment.js".

- **Optimization**
   - **Batch operations**
   - **Performance**
   - **Bundle size**

docs/next/guides/building-and-testing/modules.md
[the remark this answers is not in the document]
warpech: Thanks. I guess I used a Markdown previewer that didn't show the icons

docs/next/guides/building-and-testing/modules.md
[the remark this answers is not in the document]
warpech: > @warpech - if by "old text" you mean this:

By "old text" I only meant the first paragraph. The old:

> The full bundle of Handsontable consists of multiple built-in modules such as plugins, renderers, editors, validators, and cell-types. It also includes the whole rendering functionality and advanced data management. This is a compact tool with plenty of options available.

Was better than the then-current:

> Modules let you pick and choose only those parts of Handsontable that you actually need. This can significantly reduce the size of your app.

The current version in ace44850b1fa0c41bbfa7b7781dc86fe09b737ad solved by problem.

docs/next/guides/integrate-with-vue/vue-modules.md
```diff
@@ -28,10 +29,12 @@ In the entry point file of your application, import the `handsontable/base` modu
   import Handsontable from 'handsontable/base';
   ```
 
-### Step 2: Import modules and their registering functions
-Import the modules that you want to use (for the full list of Handsontable modules, see the [modules cheatsheet](@/guides/building-and-testing/modules.md#modules-cheatsheet)).
+### Step 2: Import optional modules
```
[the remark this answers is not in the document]
AMBudnik: @warpech @kirszenbaum I have to agree that keeping import and register together is a better option (especially when the code box takes a lot of space on the page).

docs/next/guides/building-and-testing/modules.md
warpech: Thanks, I know that you based your changed on the existing text on this page. It turns out that in my review, I addressed more than just your changes.

Regarding the "package" vocabulary, maybe you can make a new issue about that? I agree that this term needs syncing. I like the word "bundle" but I could argue that "build" would be even better. We could then call the page "Builds". I would use this term in sentences in the following way:

> The "full" and "bare" builds of Handsontable are distributed as a part of the NPM package. These builds are available for installation from NPM or for direct loading from a CDN. If you use a bundler for building your app and install Handsontable from NPM, you can also import Handsontable by specific [modules](link). Finally, it is also possible to make a [Custom build](link) of Handsontable.

docs/next/guides/building-and-testing/modules.md
warpech: Thanks, as I suggested, I think that the following comment should be also added to the "Installation" page:

```
// if you bundler allows it, import Handsontable's CSS
```

docs/next/guides/building-and-testing/modules.md
warpech: > Just because a function/method name *seems* self-explanatory, doesn't mean that it is: that function could do sth entirely else than the name suggests (as sometimes is the case in HoT).

Can't argue with that, but still I think that a sentence "the `registerEditor()` function registers a chosen editor" make the reader roll their eyes. Your proposed version sounds much better.

docs/next/guides/building-and-testing/modules.md
warpech: Thanks for pointing out the mistake in my line, @budnix.

When someone imports all editors, could they do the following?

```
import * as editors from 'handsontable/editors';
registerAllEditors();
```

If yes, then I think that would be the best way to present `registerAllEditors` in this page.

> The bundler while transpiling the code would remove this import (tree-shaking) causing unexpected results in prod and/or dev builds. That's why there is a requirement to import component and register it manually.

Doesn't this problem apply to `registerAllEditors()`, no matter how the editors are imported? In yes, then we should add a remark about that in the guide.

docs/next/guides/building-and-testing/modules.md
warpech: Aha! So the function `registerEditor` only **registers**, but the function `registerAllEditors` **imports** and registers. This is totally not obvious.

@kirszenbaum can you please make it apparent in the docs. I think that the methods `registerEditor` and `registerAllEditors` must never be mixed on the same snippet to avoid confusion.

The same applies for to other `registerAll*` functions.

docs/next/guides/building-and-testing/modules.md
warpech: Thanks. I agree that it makes sense to draw attention to a single line in a long snippet. But in a short snippet, it just repeats the same information that can be found in the snippet description.

docs/next/guides/building-and-testing/modules.md
warpech: Thanks for the context. You take the word "guide" quite literally (like a tourist guide). I take it as the following [definition](https://dictionary.cambridge.org/dictionary/english/guide):

> a book that gives you the most important information about a particular subject

On the general "guide" pages, we just explain topics in higher detail than on the API reference pages, showing some context behind that. We show the reader his options, but we don't know his goal, so it is hard to throw "steps" at this stage. We need to present content in a way that the steps will be apparent for the reader.

If we ever add "how-to guides", these would be pages that explain how to add a certain tasks.

We are not the only ones with vision for a guide:

- https://ckeditor.com/docs/ckeditor4/latest/guide/index.html (section called Guide). Notably it is not called Guide in version 5 (just "Docs"): https://ckeditor.com/docs/ckeditor5/latest/features/index.html
- https://reactjs.org/docs/accessibility.html (section Advanced Guides)
- https://v3.vuejs.org/guide/introduction.html (section Guide). I find this particularly good example of what I mean.

docs/next/guides/building-and-testing/modules.md
warpech: Your proposal is much better than mine and conveys everything I wanted. Thanks!

docs/next/guides/building-and-testing/modules.md
warpech: @kirszenbaum this proposal is almost good, but I don't agree on saying that we recommend these bundlers. It is not our job to recommend a bundler, because that's a prior decision by the app developer.

I would say this:

```
To use Handsontable modules, your environment has to support the `import` syntax, which is typically provided by a bundler.

We have successfully tested Handsontable with the following bundlers:
- [webpack](https://webpack.js.org)
- [Parcel](https://parceljs.org)
- Rollup
- Vite

If Handsontable modules don't work with your bundler, [report it as a bug](https://github.com/handsontable/handsontable/issues/new?assignees=&labels=&template=01-handsontable.md&title=).
```

docs/next/guides/integrate-with-vue/vue-modules.md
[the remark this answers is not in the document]
AMBudnik: @warpech, yes, I saw that. When it comes to that, I do not have my favorite approach. Both are readable with the amount of information on that page.

docs/next/guides/building-and-testing/modules.md
kirszenbaum: Issue created: https://github.com/handsontable/handsontable/issues/9540

docs/next/guides/integrate-with-vue/vue-modules.md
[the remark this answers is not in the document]
kirszenbaum: Ok, I've put importing modules and registering functions back together

kirszenbaum:
@warpech 

- I reverted changes made in the `11.1` docs version, and applied all changes only to the `next` version (I'll copy them over to `12.0`, after the PR is approved, to minimize making duplicate changes).
- I updated the `docs/issue-8994` branch from `develop`, solved conflicts etc.
- I applied SEO changes in line with [#9374](https://github.com/handsontable/handsontable/pull/9374)
- Since we refer to the core features of Handsontable as "the core", I renamed "required modules" to "core modules" – I think it better reflects the situation, and is more consistent with terminology used in other places
- I applied your suggestions from [here](https://github.com/handsontable/handsontable/pull/9059#discussion_r791559799) and [here](https://github.com/handsontable/handsontable/pull/9059#discussion_r791568133), renaming the topics and sections, and updating links and the TOC accordingly
- I applied most of your remaining suggestions (I marked those conversations as resolved)
- I made a bunch of other changes as well (e.g., I updated the structure of the **Modules** page's sections)

Review:
```
npm run docs:review 21df41cdcf8d597801330af9e64a8ac860bbdb5f
```
http://localhost:8000/docs/next/modules/
http://localhost:8000/docs/next/react-modules/
http://localhost:8000/docs/next/angular-modules/
http://localhost:8000/docs/next/vue-modules/
http://localhost:8000/docs/next/vue3-modules/

Let me know what you think, thanks!

docs/next/guides/building-and-tooling/modules.md
```diff
@@ -0,0 +1,1077 @@
...
+- The default cell type: [`text`](@/guides/cell-types/cell-type.md#available-cell-types)
+- Handsontable's CSS
+- Required dependencies:
+  - [moment.js](https://momentjs.com/) (you can easily [optimize](@/guides/optimization/bundle-size.md#optimize-moment-js) its size)
+  - [DOMpurify](https://www.npmjs.com/package/dompurify)
+  - [core-js](https://www.npmjs.com/package/core-js)
+
+### Import core modules
+
+To import the core modules:
```
warpech, line 63: I am not sure how it is helpful that we refer to `handsontable/base` as "core module". The word "base" is right in the name, so I think we should refer to it as "base module" to avoid confusion.

docs/next/guides/building-and-tooling/modules.md
```diff
@@ -0,0 +1,1077 @@
...
+import 'handsontable/dist/handsontable.full.css';
+import {
+  registerEditor,
+  PasswordEditor,
+} from 'handsontable/editors';
+
+registerEditor(PasswordEditor);
+```
+
+#### Import all editor modules
```
warpech, line 159: The sections:

- Import all renderer modules
- Import all editor modules
- Import all validator modules
- Import all cell type modules
- Import all plugin modules

Explain our convenience APIs that diminish the purpose of using modules in the first place. The lengthly descriptions of these APIs significantly water down the content of this page and undoubtedly make it confusing for the reader (what are they trying to tell me with that?)

My advice is to remove these sections and just add a mention of these convenience APIs clearly in the the section "List of all module imports".

warpech (changes requested):
This is not a full review, but the best I could do today:

docs/next/guides/sidebar.js
```diff
@@ -147,20 +147,21 @@ const internationalizationItems = [
...
-const buildingAndTestingItems = [
-  'guides/building-and-testing/building',
-  'guides/building-and-testing/testing',
-  'guides/building-and-testing/packages',
-  'guides/building-and-testing/modules',
-  'guides/building-and-testing/plugins',
-  'guides/building-and-testing/file-structure',
+const buildingAndToolingItems = [
+  'guides/building-and-tooling/packages',
+  'guides/building-and-tooling/modules',
```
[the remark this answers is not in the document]
kirszenbaum: @warpech - done; [7fa2122](https://github.com/handsontable/handsontable/pull/9059/commits/7fa21224300a26779d485e3fd2835cd264cc54f8)

docs/next/guides/building-and-tooling/custom-builds.md
```diff
@@ -1,17 +1,17 @@
 ---
-title: Building
-metaTitle: Building - Guide - Handsontable Documentation
-permalink: /next/building
-canonicalUrl: /building
```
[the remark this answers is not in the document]
kirszenbaum: @warpech - done: [7fa2122](https://github.com/handsontable/handsontable/pull/9059/commits/7fa21224300a26779d485e3fd2835cd264cc54f8)

docs/next/guides/sidebar.js
```diff
@@ -188,23 +189,23 @@ const upgradeAndMigrationItems = [
...
+    { title: 'Cell features', children: cellFeaturesItems },
+    { title: 'Cell functions', children: cellFunctionsItems },
+    { title: 'Cell types', children: cellTypesItems },
     { title: 'Formulas', children: formulasItems },
-    { title: 'Accessories and Menus', children: accessoriesAndMenusItems },
+    { title: 'Accessories and menus', children: accessoriesAndMenusItems },
     { title: 'Internationalization', children: internationalizationItems },
-    { title: 'Building and Testing', children: buildingAndTestingItems },
-    { title: 'Advanced Topics', children: advancedTopicsItems },
+    { title: 'Building and tooling', children: buildingAndToolingItems },
```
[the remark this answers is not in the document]
kirszenbaum: @warpech - done: [7fa2122](https://github.com/handsontable/handsontable/pull/9059/commits/7fa21224300a26779d485e3fd2835cd264cc54f8)

docs/next/guides/optimization/bundle-size.md
```diff
@@ -0,0 +1,74 @@
...
+
+# Bundle size
+
+Use modules and optimize Moment.js, to reduce Handsontable's bundle size.
+
+[[toc]]
+
+## Use modules
+
+To reduce the bundle size and JavaScript parsing time, import only those of Handsontable's [modules](@/guides/building-and-tooling/modules.md#list-of-all-modules) that you actually use, instead of importing the complete package.
```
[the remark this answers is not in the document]
kirszenbaum: @warpech - done: [7fa2122](https://github.com/handsontable/handsontable/pull/9059/commits/7fa21224300a26779d485e3fd2835cd264cc54f8)

docs/next/guides/building-and-tooling/modules.md
kirszenbaum: @warpech - done: [7fa2122](https://github.com/handsontable/handsontable/pull/9059/commits/7fa21224300a26779d485e3fd2835cd264cc54f8)

docs/next/guides/building-and-tooling/modules.md
kirszenbaum: @warpech - done: [7fa2122](https://github.com/handsontable/handsontable/pull/9059/commits/7fa21224300a26779d485e3fd2835cd264cc54f8)

kirszenbaum:
Thanks @warpech! I applied your suggestions (+ a few minor changes) in [7fa2122](https://github.com/handsontable/handsontable/pull/9059/commits/7fa21224300a26779d485e3fd2835cd264cc54f8).

docs/docker/redirects.conf
```diff
@@ -67,6 +67,9 @@ rewrite ^/((\d+\.\d+|next)/)?tutorial-known-limitations/?$ /docs/$1third-party-l
 rewrite ^/internationalization-i18n/?$ /docs/language/ permanent;
 rewrite ^/keyboard-navigation/?$ /docs/keyboard-shortcuts/ permanent;
 rewrite ^/hello-world/?$ /docs/demo/ permanent;
+rewrite ^/building/?$ /docs/custom-builds/ permanent;
+rewrite ^/custom-plugins/?$ /docs/plugins/ permanent;
```
kirszenbaum, line 71:
```suggestion
rewrite ^/plugins/?$ /docs/custom-plugins/ permanent;
```

docs/next/guides/building-and-tooling/custom-builds.md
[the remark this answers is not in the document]
kirszenbaum: Right! I think it's three:
- building -> custom-builds
- plugins -> custom-plugins
- file-structure -> folder-structure

Done:
[49f4777](https://github.com/handsontable/handsontable/pull/9059/commits/49f47779f42b9b061f13ef15ac1dc8088d81cecb) and [d556b37](https://github.com/handsontable/handsontable/pull/9059/commits/d556b37648342c6bad241c7d7473e2d669d8700e)

docs/next/guides/building-and-tooling/modules.md
warpech: It seems that now the information about the convenience methods is now only available in the code snippet below "List of all module imports". This is a little bit too obscure. 

Could the text before the snippet be extended to have a short description or the convenience methods `registerAllEditors`, `registerAllRenderers`, `registerAllValidators`, `registerAllCellTypes`, `registerAllPlugins`, `registerAllModules`. It is needed given the fact that these methods do not even have an entry in the API reference. We shouldn't just throw them unannounced inside the snippet.

warpech: Actually, I think it would be more correct to have 2 snippets below "List of all module imports":

1. One that uses the convenience methods (without the tutorial steps, just a description and the snippet)
2. One that registers all modules explicitly (without the tutorial steps, just a description and the snippet)

kirszenbaum: @warpech - done: [affb722](https://github.com/handsontable/handsontable/pull/9059/commits/affb7226e03707073cd09b76468ba5f55c60dc5b)

warpech (changes requested):
Thanks! Here's another round from me:

docs/next/guides/integrate-with-vue3/vue3-modules.md
```diff
@@ -9,57 +9,59 @@ canonicalUrl: /vue3-modules
...
 
-## Using modules with Vue 3
+## Use modules with Vue 3
 
 To use modules with Handsontable's [Vue 3 wrapper](@/guides/integrate-with-vue3/vue3-installation.md), follow the steps below:
 
-### Step 1: Import the `handsontable/base` module
+### Step 1: Import core modules
+
+No matter which [optional modules](@/guides/tools-and-building/modules.md#optional-modules) you use, you need to import the [core modules](@/guides/tools-and-building/modules.md#core-modules).
```
warpech, line 22: The lines 20 and 22 should use "base module" not "core modules".

The same applies to other frameworks than Vue 3.

warpech (changes requested):

docs/next/guides/tools-and-building/modules.md
```diff
@@ -0,0 +1,1061 @@
...
+- Required dependencies:
+  - [moment.js](https://momentjs.com/) (you can easily [optimize](@/guides/optimization/bundle-size.md#optimize-moment-js) its size)
+  - [DOMpurify](https://www.npmjs.com/package/dompurify)
+  - [core-js](https://www.npmjs.com/package/core-js)
+
+### Import the base module
+
+To import the base module:
+1. Import Handsontable from `handsontable/base` (not from `handsontable`, which would give you the full bundle).
+2. If your bundler allows it, import Handsontable's CSS file as a whole (just like when you [install](@/guides/getting-started/installation.md) Handsontable).
```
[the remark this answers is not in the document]
kirszenbaum: Done: [00782fc](https://github.com/handsontable/handsontable/pull/9059/commits/00782fcc1b2df8ad57089eb74af289d4a696cb72)

docs/next/guides/tools-and-building/modules.md
```diff
@@ -0,0 +1,1061 @@
...
+    ```js
+    import Handsontable from 'handsontable/base';
+    // if you bundler allows it, import Handsontable's CSS
+    import 'handsontable/dist/handsontable.full.css';
+    ```
+2. Import the registering function and a cell type module of your choice. For example:
+    ```js
+    import {
+      registerCellType,
+      CheckboxCellType,
```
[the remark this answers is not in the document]
kirszenbaum: Done: [798b315](https://github.com/handsontable/handsontable/pull/9059/commits/798b315848f4d3c05ac3a81794ea07b792a7e3b4)

docs/next/guides/integrate-with-vue3/vue3-modules.md
kirszenbaum: Done: [7bbe690](https://github.com/handsontable/handsontable/pull/9059/commits/7bbe690ff8a99f8550cb10c18b27115deec81297)

warpech (changes requested):
@kirszenbaum please note my commit to this PR. Other than that, I just have one proposal and that's it from me:

kirszenbaum:
> @kirszenbaum please note my commit to this PR. Other than that, I just have one proposal and that's it from me:

@warpech The **Installation** page is another pair of shoes. It mixes a few different ways of installing HoT (+ "installing" seems to mean a few different things), and lacks any structure (e.g., a basic user like me wouldn't be able to use it, not knowing which steps are optional, and which steps they should actually take first). But it's out of scope for this PR.

For now, I'm just:
- Applying a basic content structure
- Adding a code sample at the top of the page, to demonstrate how easy it is to get started using a single method of installation
- Simplifying the language

Here's my proposition: [141ef61](https://github.com/handsontable/handsontable/pull/9059/commits/141ef61e8a3852326841259a4f9b26099227385a)

docs/next/guides/getting-started/installation.md
```diff
@@ -11,20 +11,65 @@ tags:
...
+To use Handsontable with a framework, see:
+
+ - [Installation in React](@/guides/integrate-with-react/react-installation.md)
+ - [Installation in Angular](@/guides/integrate-with-angular/angular-installation.md)
+ - [Installation in Vue 2](@/guides/integrate-with-vue/vue-installation.md)
+ - [Installation in Vue 3](@/guides/integrate-with-vue3/vue3-installation.md)
 :::
 
-## Download and install the library
+## Overview
```
warpech, line 25:
```suggestion
## Overview

The most common installation boils down to the snippets presented in the tabs below. We will break it down into details further down the page.
```

docs/next/guides/getting-started/installation.md
```diff
@@ -11,20 +11,65 @@ tags:
...
+
+  ```bash
+  npm install handsontable
+  ```
+
+  </code-block>
+  <code-block title="Your HTML">
+
+  ```html
+  <div id="your-container"></div>
```
warpech, line 38: The container name "example" is a convention used on 37 pages. I would love to change it to "my-container", but only on all pages at once.

```suggestion
  <div id="example"></div>
```

docs/next/guides/getting-started/installation.md
```diff
@@ -11,20 +11,65 @@ tags:
...
+
+  ```js
+  import Handsontable from 'handsontable';
+  
+  const container = document.getElementById('your-container');
+  
+  const yourHandsontableInstance = new Handsontable(container, {
+    data: Handsontable.helper.createSpreadsheetData(5, 5),
+    licenseKey: 'non-commercial-and-evaluation' // for non-commercial use only
+  });
```
warpech, line 52: I have a few comments to this snippet:

1. It is confusing that this snippet is different than the other snippet down the page. 
2. It is missing the importing of the stylesheet.
3. The instance name `hot` is a convention used on 88 pages. I would love to change it to `handsontable`, but only on all pages at once. The name `handsontable` would be good, because it is an established convention that an instance of a class has the same name as the class but in lowercase.
4. We should avoid showing off `createSpreadsheetData`, for the following reasons:
   - it's function is unknown to the reader of this page
   - the "data grid" scenario should be the promoted one. Soon most of the snippets will be converted to "an array of objects"
   - `createSpreadsheetData` is a function that we mostly use internally for the tests and quick demos, but it's teaching value is doubtful

----

I suggest to just paste the snippet found at the bottom of the page:

```suggestion
import Handsontable from "handsontable";
import "handsontable/dist/handsontable.full.css";

const data = [
  ["", "Tesla", "Volvo", "Toyota", "Ford"],
  ["2019", 10, 11, 12, 13],
  ["2020", 20, 11, 14, 13],
  ["2021", 30, 15, 12, 13],
];

const container = document.getElementById("example");

const hot = new Handsontable(container, {
  data: data,
  rowHeaders: true,
  colHeaders: true,
  height: "auto",
  licenseKey: "non-commercial-and-evaluation", // for non-commercial use only
});
```

docs/next/guides/getting-started/installation.md
```diff
@@ -41,41 +86,86 @@ Run the following command in your terminal
...
+```html
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/handsontable/dist/handsontable.full.min.css" />
+```
+
+## Create a container
+
+In your HTML, add an empty `div`, which serves as a container for your Handsontable instance.
+
+```html
+<div id="my-container"></div>
```
warpech, line 163:
```suggestion
<div id="example"></div>
```

docs/next/guides/getting-started/installation.md
```diff
@@ -84,7 +174,7 @@ const data = [
   ['2021', 30, 15, 12, 13]
 ];
 
-const container = document.getElementById('example');
+const container = document.getElementById('my-container');
```
warpech, line 177:
```suggestion
const container = document.getElementById('example');
```

docs/next/guides/getting-started/installation.md
```diff
@@ -41,41 +86,86 @@ Run the following command in your terminal
...
+
+If you're using Handsontable as a CommonJS package, or as an ECMAScript module (using a package manager), import the full distribution of Handsontable as a JavaScript file.
+
+Use your bundler's preferred importing method. For example:
+
+```js
+import Handsontable from 'handsontable';
+```
+
+#### Using UMD
```
warpech, line 124: Sadly most people who use UMD, don't know that it is called UMD. Let's change the text of the heading.

```suggestion
#### Using the `script` tag
```

docs/next/guides/getting-started/installation.md
```diff
@@ -41,41 +86,86 @@ Run the following command in your terminal
...
+To get Handsontable's files from a CDN, use the following locations:
 
-  </code-block>
-</code-group>
+```js
+https://cdn.jsdelivr.net/npm/handsontable/dist/handsontable.full.min.js
 
-## Create a placeholder
+https://cdn.jsdelivr.net/npm/handsontable/dist/handsontable.full.min.css
+```
```
warpech, line 106: I think that a list of links would actually be the most useful here:

- [https://cdn.jsdelivr.net/npm/handsontable/dist/handsontable.full.min.js](https://cdn.jsdelivr.net/npm/handsontable/dist/handsontable.full.min.js)
- [https://cdn.jsdelivr.net/npm/handsontable/dist/handsontable.full.min.css](https://cdn.jsdelivr.net/npm/handsontable/dist/handsontable.full.min.css)

docs/next/guides/getting-started/installation.md
```diff
@@ -41,41 +86,86 @@ Run the following command in your terminal
...
+
+## Import Handsontable's JavaScript
+
+Import Handsontable's JavaScript into your application.
+
+### Full distribution
```
warpech, line 112: I think that introducing so many choices as separate sections can create a choice fatigue. In my commit 61e4709a3d856f6cb5765c7f9ebb3a8d80199480 it looked much simpler.

If you think some more subheaders could be added to my text under **Import Handsontable's JavaScript**, maybe just add two:

```
### Using CommonJS or a package manager
### Using the `script` tag
```

warpech (changes requested):
Thanks for the creative changes. Back to you :)

warpech (approved):
I think we are done! 🎊🎊🎊

Issue #9045, opened 2021-12-17
Handsontable shows warnings with HyperFormula license key missing


adrianszymanski89:
### Description
When the license key for HyperFormula isn't provided the application throws two warnings.

### Steps to reproduce
1. Create HyperFormula instance inside Handsontable application.
2. Provide license key for Handsontable which should be enough here.

### Demo
https://jsfiddle.net/warpech/p6sakngd/1/
<img width="387" alt="Zrzut ekranu 2021-12-17 o 09 41 46" src="https://user-images.githubusercontent.com/92314130/146519240-0eba4b24-cf3c-419e-9d6b-f0b855b80330.png">

### Your environment
* Handsontable version: 11.0.1
* Browser Name and version: Google Chrome 96.0.4664.110
* Operating System: MacOS Monterey 12.0.1

adrianszymanski89:
The issue regarding binarySearchThreshold is already reported here: https://github.com/handsontable/handsontable/issues/8771

budnix:
After investigation, it turned out that if the _Formulas_ plugin uses an external instance of HyperFormula the HF has to be initialized with `licenseKey: "internal-use-in-handsontable"`. Update [demo](https://jsfiddle.net/budnix/gyds8h21/). 

It's important that this license key can be used only with Handsontable as a part of the _Formulas_ plugin.

kirszenbaum:
Done in #9081, closing

PR #9081, opened 2022-01-03
Docs: Adding information about license key required when HF used within HoT

kirszenbaum:
This PR:
- Adds information about the `'internal-use-in-handsontable'` license key required when using an external HyperFormula instance within Handsontable's `Formulas` plugin (fixing #9045)

Content modified:
- [Formulas](https://handsontable.com/docs/formula-calculation/) (versions `9.0`-`next`)
- [`formulas` API reference](https://handsontable.com/docs/api/options/#formulas) (versions `9.0`-`next`)

[skip changelog]

### DIFF
--- a/docs/10.0/api/formulas.md
+++ b/docs/10.0/api/formulas.md
@@ -42,7 +42,10 @@ formulas: {
 }
 
 // or, add a HyperFormula instance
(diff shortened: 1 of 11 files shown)


budnix (approved):
👍

Issue #8989, opened 2021-11-30
Docs: Bring back the Hello World page


kirszenbaum:
In #8988, we temporarily hid the Hello world page (for versions `9.0`, `10.0`, and `11.0`), due to CodeSandbox problems.

To do in this issue:
- **After** #8980 is done (i.e. the demos are successfully migrated from CodeSandbox), bring the Hello World pages back

warpech:
Done in https://github.com/handsontable/handsontable/pull/8982. Some remaining tasks are tracked in https://github.com/handsontable/handsontable/issues/8980

Issue #8980, opened 2021-11-26
Update the "Hello world" page


kirszenbaum:
Update the [Hello world](https://handsontable.com/docs/hello-world/) page:
~~- [ ] Migrate the CodeSandbox demo to the docs' `::: example` containers~~
  ~~- [x] JavaScript~~
  ~~- [ ] React~~
  ~~- [ ] Angular~~
  ~~- [ ] Vue~~
- [x] Change the title from "Hello world" to "Demo"
- [x] Rewrite the page's text to stress the fact that the demo is not an end-user showcase, but is meant to let developers see the code
- [x] Redirect URL: `/hello-world/` to `/demo/`

### Links
https://handsontable.com/docs/hello-world/

warpech:
The "Demo" example is now online: https://handsontable.com/docs/demo/

Some points from the above checklist are not done yet. They should be addressed soon or moved to new issues.

warpech:
Closing as all of the tasks are either done or decided to be out-of-scope

PR #9116, opened 2022-01-14
Docs: Hiding the "Hello, World!" page from the menu 

kirszenbaum:
This PR:
- Removes the "Hello, World!" page from the menu, for versions `11.1` and `next` (it should be brought back in #8989, after #8980 is done)

Affected content:
- https://handsontable.com/docs/hello-world/ (shouldn't be visible in the menu)

[skip changelog]


warpech (approved):
LGTM

warpech:
Merging, because it has an important immediate effect.

PR #9126, opened 2022-01-19
Docs: Linking `locale` nad `language` options' API ref to guides

kirszenbaum:
This PR:
- Adds links from the API ref of the `locale` nad `language` settings to the corresponding sections in the guides


AMBudnik (approved):
👍  works well

PR #9135, opened 2022-01-21
[RTL] Add new `layoutDirection` Handsontable option

### LINKED ISSUE #9132: [RTL] Turning on/off the RTL feature
### Description

The issue describes the Handsontable option that will be used to enable or disable the support for RTL. After discussion, the team agreed to use the `layoutDirection` option name. The option allows set three values:
 * `'rtl'` The option sets the `dir="rtl"` attribute to the root Handsontable element forcing the browser to render the table in right-to-left direction mode;
 * `'ltr'` The option sets the `dir="ltr"` attribute to the root Handsontable element forcing the browser to render the table in left-to-right direction mode;
 * `'inherit'` (default value) Handsontable tries to detect the document direction mode by traversing from the root Handsontable element ending in the `<HTML>` element. Then sets the `dir` attribute to the root Handsontable element with the same value as detected earlier. Thanks to that Handsontable won't be rendered incorrectly after further changes in the "dir" attribute of the parent element.

The `layoutDirection` option can be set only while Handsontable initialization. Only then the rendering engine is able to prepare the overlays to render in LTR or RTL modes. That's why the value can not be changed using the `updateSettings()` method. If the option key will be passed to the method the warning should be printed in the console.

budnix:
### Context

The PR adds a new `layoutDirection` Handsontable option. Thanks to that there is possible to initialize the table that will be correctly rendered in the LTR and RTL document modes. The new option follows the rules described in the #9132 issue.

_[skip changelog]_

### How has this been tested?

I tested the changes locally and I added new E2E tests that cover the fix.

### Types of changes

### Related issue(s):
1. resolves #9132

### Affected project(s):
- [x] `handsontable`

### Checklist:

### DIFF
--- a/handsontable/src/core.js
+++ b/handsontable/src/core.js
@@ -140,7 +140,7 @@ export default function Core(rootElement, userSettings, rootInstanceSymbol = fal
    * @returns {boolean} True if RTL.
    */
   this.isRtl = function() {
-    return instance.rootWindow.getComputedStyle(instance.rootElement).direction === 'rtl';
+    return instance.rootElement.getAttribute('dir') === 'rtl';
   };
 
   /**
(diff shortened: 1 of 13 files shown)


Siemienik (approved):
LGTM

PR #9144, opened 2022-01-24
[RTL] Custom Borders direction-less API

### LINKED ISSUE #9129: [RTL] Custom Borders API
### Description

The plugin configuration API needs to be enhanced to work with RTL. Currently, the configuration allows changing the right and left borders by defining the object keys with the same names. It works predictably on LTR but it does not in RTL. On RTL the "right" border is rendered on the left and vice versa, like in the screenshot:

2022-01-19 o 14 29 54

All two demos use the same configuration:
```js
customBorders: [
  {
    row: 2,
    col: 2,
    left: { width: 4,  color: 'red' },
    right: { width: 8,  color: 'green' }
  }
],
```

It's confusing that's why the new naming has to be implemented. To be backward-compatible the old naming will work only if Handsontable is running in LTR mode. In RTL mode the error will be printed when the "left" and/or "right" property name is used. The direction-less options are "**start**" and "**end**", for example:

```js
customBorders: [
  {
    row: 2,
    col: 2,
    start: { width: 4,  color: 'red' },
    end: { width: 8,  color: 'green' }
  }
],
```
In that case, in both layout types, the "start" key defines the border placement nearest the left overlay (or right in RTL) and "end" the farthest position.

image

Related to https://github.com/handsontable/handsontable/issues/8785.

budnix:
### Context

The PR implements the new API for objects that declares the borders styling. The _CustomBorders_ plugin allows changing the left and right borders using the "left"/"right" property names. That is confusing when the table is run in the RTL document mode.

The new API changes those names, so the `left` property becomes `start` and `right` becomes `end`. The `start` property depends on the document layout direction pointing to the left border in LTR and the right border in RTL.

The proposed changes are backward compatible. It simply means that after the library upgrades the developers do not have to change anything in their apps. The "left"/"right" properties will work as previously. With indication that it would only work in LTR document mode.

image

### Backward compatible API
```js
new Handsontable(document.querySelector('#example'), {
  licenseKey: 'non-commercial-and-evaluation',
  data: Handsontable.helper.createSpreadsheetData(100, 100),
  rowHeaders: true,
  colHeaders: true,
  customBorders: [
    {
      row: 2,
      col: 2,
      left: { width: 4,  color: 'red' },
      right: { width: 8,  color: 'green' }
    },
  ],
});

hot.getCellMeta(2, 2).borders
// The "borders" meta contains the "left"/"right" (backward compatibility) 
// and "start"/"end" properties
//   {
//     left: {width: 4, color: 'red'}
//     right: {width: 8, color: 'green'}
//     start: {width: 4, color: 'red'}
//     end: {width: 8, color: 'green'}
//  }
```

### New API (in LTR mode)
```js
new Handsontable(document.querySelector('#example'), {
  licenseKey: 'non-commercial-and-evaluation',
  data: Handsontable.helper.createSpreadsheetData(100, 100),
  rowHeaders: true,
  colHeaders: true,
  customBorders: [
    {
      row: 2,
      col: 2,
      start: { width: 4,  color: 'red' },
      end: { width: 8,  color: 'green' }
    },
  ],
});

hot.getCellMeta(2, 2).borders
// The "borders" meta contains the "start"/"end" properties
// and "left"/"right" (for backward compatibility)
//   {
//     left: {width: 4, color: 'red'}
//     right: {width: 8, color: 'green'}
//     start: {width: 4, color: 'red'}
//     end: {width: 8, color: 'green'}
//  }
```

### New API (in RTL mode)
```js
new Handsontable(document.querySelector('#example'), {
  licenseKey: 'non-commercial-and-evaluation',
  data: Handsontable.helper.createSpreadsheetData(100, 100),
  rowHeaders: true,
  colHeaders: true,
  customBorders: [
    {
      row: 2,
      col: 2,
      start: { width: 4,  color: 'red' },
      end: { width: 8,  color: 'green' }
    },
  ],
});

hot.getCellMeta(2, 2).borders
// The "borders" meta contains only the "start"/"end" properties.
//   {
//     start: {width: 4, color: 'red'}
//     end: {width: 8, color: 'green'}
//  }
```

_[skip changelog]_

### How has this been tested?

I tested the changes locally and cover the feature (including backward support) with new E2E tests.

### Types of changes

### Related issue(s):
1. resolves #9129
2.
3.

### Affected project(s):
- [x] `handsontable`

### Checklist:

### DIFF
--- a/handsontable/src/plugins/customBorders/customBorders.js
+++ b/handsontable/src/plugins/customBorders/customBorders.js
@@ -14,7 +14,7 @@ import {
   hasLeftRightTypeOptions,
   hasStartEndTypeOptions,
   toInlinePropName,
-  createInlinePropNamesMap,
+  createBorderNormalizer,
 } from './utils';
 import { detectSelectionType, normalizeSelectionFactory } from '../../selection';
 
@@ -97,25 +97,7 @@ export class CustomBorders extends BasePlugin {
    * @private
    * @type {{start: string, end: string}}
    */
-  #inlineProps = createInlinePropNamesMap();
-
-  /**
-   * Returns the property name that identifies the most left border position.
-   *
-   * @private
-   */
-  get inlineStartProp() {
-    return this.#inlineProps.start;
-  }
-
-  /**
-   * Returns the property name that identifies the most right border position.
-   *
-   * @private
-   */
-  get inlineEndProp() {
-    return this.#inlineProps.end;
-  }
+  #borderNormalizer = createBorderNormalizer();
 
   /**
    * Checks if the plugin is enabled in the handsontable settings. This method is executed in {@link Hooks#beforeInit}
@@ -183,7 +165,16 @@ export class CustomBorders extends BasePlugin {
    * @param {object} borderObject Object with `top`, `right`, `bottom` and `start` properties.
    */
   setBorders(selectionRanges, borderObject) {
-    const borderKeys = Object.keys(borderObject);
+    let borderKeys = ['top', 'bottom', 'start', 'end'];
+    let normBorder = null;
+
+    if (borderObject) {
+      this.checkSettingsCohesion([borderObject]);
+
+      borderKeys = Object.keys(borderObject);
+      normBorder = this.#borderNormalizer.normalize(borderObject);
+    }
+
     const selectionType = detectSelectionType(selectionRanges);
     const selectionSchemaNormalizer = normalizeSelectionFactory(selectionType);
 
@@ -193,7 +184,7 @@ export class CustomBorders extends BasePlugin {
       for (let row = rowStart; row <= rowEnd; row += 1) {
         for (let col = columnStart; col <= columnEnd; col += 1) {
           arrayEach(borderKeys, (borderKey) => {
-            this.prepareBorderFromCustomAdded(row, col, borderObject, borderKey);
+            this.prepareBorderFromCustomAdded(row, col, normBorder, toInlinePropName(borderKey));
           });
         }
       }
@@ -246,7 +237,7 @@ export class CustomBorders extends BasePlugin {
         for (let col = columnStart; col <= columnEnd; col += 1) {
           arrayEach(this.savedBorders, (border) => {
             if (border.row === row && border.col === col) {
-              selectedBorders.push(border);
+              selectedBorders.push(this.#borderNormalizer.denormalize(border));
             }
           });
         }
@@ -289,16 +280,6 @@ export class CustomBorders extends BasePlugin {
   }
 
   /**
-   * Checks if the plugin is initialized using the backward compatible setting options ("left"/"right" props).
-   *
-   * @private
-   * @returns boolean
-   */
-  isStartEndPropsMode() {
-    return this.inlineStartProp === 'start' && this.inlineEndProp === 'end';
-  }
-
-  /**
    * Insert WalkontableSelection instance into Walkontable settings.
    *
    * @private
@@ -316,17 +297,7 @@ export class CustomBorders extends BasePlugin {
     const hasCustomSelections = this.checkCustomSelections(border, visualCellRange, place);
 
     if (!hasCustomSelections) {
-      const borderObj = {};
-
-      Object.assign(
-        borderObj,
-        border[this.inlineStartProp] ? { start: border[this.inlineStartProp] } : {},
-        border[this.inlineEndProp] ? { end: border[this.inlineEndProp] } : {},
-        border.top ? { top: border.top } : {},
-        border.bottom ? { bottom: border.bottom } : {},
-      );
-
-      this.hot.selection.highlight.addCustomSelection({ border: borderObj, visualCellRange });
+      this.hot.selection.highlight.addCustomSelection({ border, visualCellRange });
     }
   }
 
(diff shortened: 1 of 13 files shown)


handsontable/src/plugins/customBorders/customBorders.js
```diff
@@ -605,7 +625,8 @@ export class CustomBorders extends BasePlugin {
    * Check if an border already exists in the savedBorders array, and if true update border in savedBorders.
    *
    * @private
-   * @param {object} border Object with `row` and `col`, `left`, `right`, `top` and `bottom`, `id` and `border` ({Object} with `color`, `width` and `cornerVisible` property) properties.
+   * @param {object} border Object with `row` and `col`, `start`, `end`, `top` and `bottom`, `id` and
```
Siemienik, line 628: Could I propose to you to use @typedef for defining the type of borders?

handsontable/src/plugins/customBorders/customBorders.js
```diff
@@ -636,8 +657,9 @@ export class CustomBorders extends BasePlugin {
    * Check if an border already exists in the customSelections, and if true call toggleHiddenClass method.
    *
    * @private
-   * @param {object} border Object with `row` and `col`, `left`, `right`, `top` and `bottom`, `id` and `border` ({Object} with `color`, `width` and `cornerVisible` property) properties.
-   * @param {string} place Coordinate where add/remove border - `top`, `bottom`, `left`, `right` and `noBorders`.
+   * @param {object} border Object with `row` and `col`, `start`, `end`, `top` and `bottom`, `id` and
```
Siemienik, line 660: same as https://github.com/handsontable/handsontable/pull/9144/files#r793013086

handsontable/src/plugins/customBorders/customBorders.js
```diff
@@ -664,9 +686,10 @@ export class CustomBorders extends BasePlugin {
    * Check if an border already exists in the customSelections, and if true reset cellRange.
    *
    * @private
-   * @param {object} border Object with `row` and `col`, `left`, `right`, `top` and `bottom`, `id` and `border` ({Object} with `color`, `width` and `cornerVisible` property) properties.
+   * @param {object} border Object with `row` and `col`, `start`, `end`, `top` and `bottom`, `id` and
```
Siemienik, line 689: same as https://github.com/handsontable/handsontable/pull/9144/files#r793013086

handsontable/src/plugins/customBorders/customBorders.js
```diff
@@ -719,6 +744,27 @@ export class CustomBorders extends BasePlugin {
     }
   }
 
+  /**
+   * Checks the settings cohesion. The properties such like "left"/"right" are supported only
+   * in the LTR mode and the "left"/"right" options can not be used together with "start"/"end" properties.
+   *
+   * @private
+   * @param {object[]} customBorders The user defined custom border objects array.
```
Siemienik, line 752: According to https://github.com/handsontable/handsontable/pull/9144/files#r793013086

`@param {Array.<BorderDef>}`

handsontable/src/plugins/customBorders/utils.js
```diff
@@ -48,13 +50,53 @@ export function createDefaultHtBorder() {
...
+    col: border.col,
+    ...(isDefined(border.id) ? { id: border.id } : {}),
+    ...(isDefined(border.border) ? { border: border.border } : {}),
+    ...(isDefined(border.top) ? { top: border.top } : {}),
+    ...(isDefined(border.bottom) ? { bottom: border.bottom } : {}),
+    ...(isDefined(border.start) ? { start: border.start, left: border.start } : {}),
+    ...(isDefined(border.end) ? { end: border.end, right: border.end } : {}),
+  };
+}
+
```
Siemienik, line 93: Both operations could have a performance impact because of many `Object.assigns`. Much more convenient seems to be using stable numbers of properties.

These also create new objects which have different references - it is a design decision about is it ok with the current architecture when probably some changes can be provided through references? I'm not sure what the current workflow about borders looks like, so let's be careful here while testing.

handsontable/src/plugins/customBorders/utils.js
```diff
@@ -48,13 +50,53 @@ export function createDefaultHtBorder() {
...
 /**
  * Prepare empty border for each cell with all custom borders hidden.
  *
  * @param {number} row Visual row index.
  * @param {number} col Visual column index.
- * @returns {object} Returns border configuration containing visual indexes. Example of an object defining it:
- * `{{id: *, border: *, row: *, col: *, top: {hide: boolean}, right: {hide: boolean}, bottom: {hide: boolean}, left: {hide: boolean}}}`.
+ * @returns {{id: string, border: any, row: number, col: number, top: {hide: boolean}, bottom: {hide: boolean}, start: {hide: boolean}, end: {hide: boolean}}} Returns border configuration containing visual indexes.
```
Siemienik, line 99: Could I propose to use @typedef to define the type for this?

Siemienik (commented):
Generally, It looks OK to me, Good job!

I noticed some minor comments, wdyt?

handsontable/src/plugins/customBorders/utils.js
budnix: Thanks for pointing that out. Generally, I'm aware of that. The decision here is dictated by the fact that the plugin is not very well written. It passes the objects by reference between different methods, modifying some data (by ref), checking some props ... all of that causes a lot of confusion. As if that was not enough, some of the objects are modified by reference from the Walkontable Border class - so basically a very bad idea.

The refactor of the plugin would be out of the scope of the task.

handsontable/src/plugins/customBorders/customBorders.js
budnix: Thanks for spotting that but I think that kind of improvement I would leave for a separate issue. Especially that the border object in the different lifecycle of the plugin, different methods can have a different shape.

handsontable/src/plugins/customBorders/customBorders.js
budnix: https://github.com/handsontable/handsontable/pull/9144#discussion_r793326180

handsontable/src/plugins/customBorders/customBorders.js
budnix: https://github.com/handsontable/handsontable/pull/9144#discussion_r793326180

handsontable/src/plugins/customBorders/utils.js
Siemienik: In general, your implementation creates new objects so it loses the references to the main Border object. That I tried to point out. In your message, you confirm that keeping references is required, so creating a new object each time when calling normalize and denormalize is incorrect.

handsontable/src/plugins/customBorders/customBorders.js
Siemienik: as you want.

Siemienik (approved):

Issue #5455, opened 2018-10-05
Column with formulas doesn't filter properly


AMBudnik:
### Description

Bug reported at https://forum.handsontable.com/t/formula-column-doesnt-filter/2451/3

### Steps to reproduce

1. Open filters and clean the `filter by value`
2. Choose `6030` and confirm

Result:
We should get `6030` value in a row but the table is empty.

### Demo

https://jsfiddle.net/handsoncode/yhzgtLb0/

### Your environment
* Handsontable version: `6.0.0`
_doesn't seem to be version related_
* Browser Name and version: Chrome 69
* Operating System: Windows 10

_inform Zen 26605, 26992_

aninde:
A bug is still replicable in `8.0.0-rc1` v. https://jsfiddle.net/aninde/m0fjap76/

AMBudnik:
This issue is still replicable using v 8.0.0.
Here's an updated demo https://jsfiddle.net/uy0Ldv8s/

AMBudnik:
This issue is closed temporarily. It will be reopened for development as it became a part of **New formula plugin** task reported at https://github.com/handsontable/handsontable/issues/6466

I will make sure to inform everyone interested in this topic after the official fix.

Please feel welcome to makes any comments on this issue. And if you experience similar behavior feel free to contact me at support@handsontable.com

AMBudnik:
Still broken in v9 https://jsfiddle.net/z8uwtsv4/

HakanAkkurt:
Is there any solution for this bug?

AMBudnik:
Hi @HakanAkkurt 

This issue was closed as a part of the bigger scope of the project reported here https://github.com/handsontable/handsontable/issues/8511 

I will notify you once we solve it.

AMBudnik:
It works the same in the latest version v11 https://jsfiddle.net/94fnmu21/3/

aninde:
This issue will be fixed by the upcoming v12. 
Working demo https://jsfiddle.net/Lk2fbr6q/

AMBudnik:
Thanks, guys! I'm closing this issue as solved in v12.

For everyone who is tracking changes to this topic please check our release notes https://handsontable.com/docs/release-notes/#_12-0-0 and the migration guide https://handsontable.com/docs/migration-from-11.1-to-12.0/#step-1-verify-your-updatesettings-calls

PR #9159, opened 2022-01-27
Filter cells with formulas using values not source (#5455)

mjastrzebowski:
### Context

This change is providing value of formula cell for filters instead of it's source value (formula itself). 

For example having 4 rows with numbers + formula, like: `[1, 2, 3, '=2+2']` and using any numeric filter previously would always filter out the formula cell as it'd got a string `'=2+2'` and after fix it'll get correct number value of `4`. 

### How has this been tested?

Added new tests specs for filters plugin with some formulas cells. 

### Types of changes

### Related issue(s):
1. https://github.com/handsontable/handsontable/issues/5455

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

### DIFF
--- a/handsontable/src/plugins/filters/filters.js
+++ b/handsontable/src/plugins/filters/filters.js
@@ -465,7 +465,7 @@ export class Filters extends BasePlugin {
     arrayEach(this.hot.getSourceDataAtCol(visualColumn), (value, rowIndex) => {
       const { row, col, visualCol, visualRow, type, instance, dateFormat, locale } = this.hot
         .getCellMeta(rowIndex, visualColumn);
-      const dataValue = this.hot.getDataAtCell((this.hot.toVisualRow(rowIndex)), visualColumn) || value;
+      const dataValue = this.hot.getDataAtCell(this.hot.toVisualRow(rowIndex), visualColumn) ?? value;
 
       data.push({
         meta: { row, col, visualCol, visualRow, type, instance, dateFormat, locale },
--- a/handsontable/src/plugins/filters/filters.js
+++ b/handsontable/src/plugins/filters/filters.js
@@ -459,16 +459,17 @@ export class Filters extends BasePlugin {
    * @returns {Array} Returns array of objects where keys as row index.
    */
   getDataMapAtColumn(column) {
-    const visualIndex = this.hot.toVisualColumn(column);
+    const visualColumn = this.hot.toVisualColumn(column);
     const data = [];
 
-    arrayEach(this.hot.getSourceDataAtCol(visualIndex), (value, rowIndex) => {
+    arrayEach(this.hot.getSourceDataAtCol(visualColumn), (value, rowIndex) => {
       const { row, col, visualCol, visualRow, type, instance, dateFormat, locale } = this.hot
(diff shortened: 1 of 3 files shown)


aninde:
I propose to add a few test cases to check relative and absolute formulas. 
For example such as:
 ```
 var data = [
       ['=SUM(B1:B2)', 1, "= A$1"],
       ['=$B1', 2, '=SUM (A1:A2)'],
   ];

...

// should filter cell with relative formula in the first cell
const filtersPlugin = hot1.getPlugin ('filters');
filtersPlugin.addCondition (0, 'by_value', ['3']);
filtersPlugin.filter()

expect (getData()[0][0]).toBe(3)

// should filter cell with absolute formula
filtersPlugin.addCondition(2, 'by_value', ['1']);
filtersPlugin.filter()

expect (getData()[0][0]).toBe(3)
```

File: handsontable/src/plugins/filters/filters.js [hunk 1]
```diff
@@ -459,16 +459,17 @@ export class Filters extends BasePlugin {
...
-    const visualIndex = this.hot.toVisualColumn(column);
+    const visualColumn = this.hot.toVisualColumn(column);
     const data = [];
 
-    arrayEach(this.hot.getSourceDataAtCol(visualIndex), (value, rowIndex) => {
+    arrayEach(this.hot.getSourceDataAtCol(visualColumn), (value, rowIndex) => {
       const { row, col, visualCol, visualRow, type, instance, dateFormat, locale } = this.hot
-        .getCellMeta(rowIndex, visualIndex);
+        .getCellMeta(rowIndex, visualColumn);
+      const dataValue = this.hot.getDataAtCell((this.hot.toVisualRow(rowIndex)), visualColumn) || value;
```
budnix, line 468:
```suggestion
      const dataValue = this.hot.getDataAtCell(this.hot.toVisualRow(rowIndex), visualColumn) ?? value;
```
The nullish coalescing operator is a better approach here. Without that if the formula returned `0` the source value (formula expression) would be returned, thus the filter condition like `eq` with `0` would not work.

budnix (changes requested):
Thanks for your contribution! I've played with the changes locally using all build-in cell types and I haven't found any issues. As I played with a demo, I'm sure now that the PR is not a breaking change.

About the code, I found some nitpicky changes to consider.

mjastrzebowski:
> filtersPlugin.addCondition(0, 'by_value', ['3']);

Added more tests for text fields as suggested by @aninde FYI `by_value` filter args should be array of values to check, so in this case: `filtersPlugin.addCondition(0, 'by_value', [['3']]);`

@budnix thanks for the review, all changes applied

budnix (dismissed):
Thanks! 👍

mjastrzebowski:
@budnix rebased with master, ready to merge

budnix (dismissed):

budnix (dismissed):

budnix (approved):

Issue #9182, opened 2022-02-02
Code Examples Deployment deploys a wrong thing


warpech:
### Description

There are some problems in the Code Examples Deployment workflow (`code-examples.yml`) that block finishing of #8980. Two of the below problematic cases were recorded today by me:

## Case 1

While PR https://github.com/handsontable/handsontable/pull/9179 was open, I wanted to make a manual deployment of the `11.1.0` and `next` examples from this PR, so I triggered two manual dispatches.

I expected the code at the commit c126921f36767709182770562a4c6ea0fa9d2b71 to be deployed, but instead got the following wrong results:

1. branch `feature/issue-8980-fix-failing-build`, Handsontable version `next`:
   - run: https://github.com/handsontable/handsontable/actions/runs/1782905004
   - wrong: result: some jobs didn't run at all, including the last one that deploys (`gather-and-deploy-builds`)
2. branch `feature/issue-8980-fix-failing-build`, Handsontable version `11.1.0`
   - run: https://github.com/handsontable/handsontable/runs/5033812746
   - deployment: https://github.com/handsontable/handsontable/tree/5bb92a6de3b1d78b29c759edf52bcdeab215cdec
   - wrong result: all jobs ran, but the deployed files were taken from `master` branch, not from the input branch
   
## Case 2

An hour later, I merged the PR #9179 to `develop`, which triggered automatic deployment.

I expected the code at the commit 555644abe8b9e38d249ed05e25acbd0ad43deccb to be deployed (including new example called `demo` in the folders of the `11.1.0` and `next` version), but instead got the following wrong results:

- run: https://github.com/handsontable/handsontable/actions/runs/1783292250
- deployment: https://github.com/handsontable/handsontable/tree/66c710ab321761e1b3244f6f44ac35fc9648d64d
- wrong result: all jobs ran, but the deployed files were taken from `master` branch, not from the input branch

warpech:
Closing as done and released immediately in the CI

PR #9184, opened 2022-02-02
Fix the code examples' deployment script.

jansiegel:
### Context
This PR should fix the problems with the code examples' deployment script described in #9182.

### How has this been tested?
Deployed the workflow both as commits (from this branch) and manual dispatch.

### Related issue(s):
1. #9182 

### Checklist:


.github/workflows/code-examples.yml
```diff
@@ -51,6 +51,7 @@ jobs:
         id: check-modified-paths
         if: github.event_name != 'workflow_dispatch'
         with:
+          base: 'develop'
```
warpech, line 54: Was this added on purpose or just for debugging? It looks a bit excessive, because (if I understand) correctly, it will report all changes on a branch, not only all changes in the current commit.

.github/workflows/code-examples.yml
```diff
@@ -445,7 +447,7 @@ jobs:
         if: steps.path-check.outputs.n-a != 'true'
         run: |
           mkdir -p ${{ env.TMP_FRAMEWORK_EXAMPLES_PATH }}
-          rsync -av --progress ${{ env.FRAMEWORK_EXAMPLES_PATH }}/ ${{ env.TMP_FRAMEWORK_EXAMPLES_PATH }} --include=/{,'[!node_modules]*/'{,dist/{,'**'}}} --exclude='*'
+          rsync -av --progress ${{ env.FRAMEWORK_EXAMPLES_PATH }}/ ${{ env.TMP_FRAMEWORK_EXAMPLES_PATH }} --exclude='node_modules' --include=/{,'*/'{,dist/{,'**'}}} --exclude='*'
```
warpech, line 450: Was this the actual problem for Case 1 and 2 in the issue? Can you explain what's going on here?

jansiegel:
@warpech Regarding both https://github.com/handsontable/handsontable/pull/9184#discussion_r798021506 and https://github.com/handsontable/handsontable/pull/9184#discussion_r798024251:

Some things were not 100% clear to me in the initial issue when I took a closer look:

> 1. branch `feature/issue-8980-fix-failing-build`, Handsontable version `next`:
>    - run: https://github.com/handsontable/handsontable/actions/runs/1782905004
>    - wrong: result: some jobs didn't run at all, including the last one that deploys (`gather-and-deploy-builds`)

It looks to me as all the jobs finished successfully, including the last one:
image
I feel that the problem with nothing being deployed is being caused by the problem I'll explain below.

> 2. branch `feature/issue-8980-fix-failing-build`, Handsontable version `11.1.0`
>    - run: https://github.com/handsontable/handsontable/runs/5033812746
>    - deployment: https://github.com/handsontable/handsontable/tree/5bb92a6de3b1d78b29c759edf52bcdeab215cdec
>    - wrong result: all jobs ran, but the deployed files were taken from `master` branch, not from the input branch

In this case, I think the problem was not deploying the files from the `master` branch, but deploying an incomplete amount of examples (without the new examples), which coincidentally looks basically the same as the version from `master`. I think the same thing happened in _"Case 2"_.

### As for my changes:
1. It looks to me like the problem with not gathering all the available examples was caused by [a mistake in the `rsync` command](https://github.com/handsontable/handsontable/pull/9184#discussion_r798024251). After this change, it seems to work fine. I manually deployed the `next` and `11.1.0` versions from this branch, and the new examples were added to `gh-pages`:
    - `next` -> [workflow run](https://github.com/handsontable/handsontable/actions/runs/1784746146) -> https://github.com/handsontable/handsontable/commit/b001067a0c88106519dcdc4b835c41035e933a3a
    - `11.1.0` -> [workflow run](https://github.com/handsontable/handsontable/actions/runs/1784880740) -> https://github.com/handsontable/handsontable/commit/52660c721168270bdc85026b0460b60f22903c40
    
2. The [second change](https://github.com/handsontable/handsontable/pull/9184#discussion_r798021506) was a problem I found while working on this task.
    Without this change, the workflow would improperly compare the changes on this branch with the `master` branch, resulting in recognizing more changes in the examples than it should: 
image
    This is on par with [this action](https://github.com/dorny/paths-filter)'s documentation, though:

    > Feature branches:
    >   Workflow triggered by [push](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#push) or any other [event](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows)
    >     - The base input parameter must not be the same as the branch that triggered the workflow
    >     **- Changes are detected against the merge-base with the configured base branch or the default branch**
    
    By setting the `base` branch to `develop`, I'm forcing the script to compare the changes from the current feature branch to the development branch (which is probably not ideal in some of the edge cases, like commits on a branch connected to a PR which base branch is _not_ `develop`, but the vast majority of commits don't fall into that category).
    
    At the same time, when a squashed commit hits `develop`, the same workflow should compare the changes only to the previous commit on the `develop` branch:

    > Master, Release, or other long-lived branches:
    >   Workflow triggered by [push](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#push) event when base input parameter is the same as the branch that triggered the workflow:
    >     **- Changes are detected against the most recent commit on the same branch before the push**

### As for:
>  It looks a bit excessive, because (if I understand) correctly, it will report all changes on a branch, not only all changes in the current commit.

I personally think it's a _good_ idea to compare all the changes from a branch, not only the current commit. In this case, the result of the workflow under the last commit in the PR will reflect the actual changes that will be deployed after merging it to `develop`. 

### Side note:
While working on this task, I found one additional issue which I decided to ignore (for now), as I feel it will cause problems in very rare scenarios:

When:
1. Commit `1` introduces changes in `examples/docs/next/angular/some-example`
    - code examples workflow triggers
2. Commit `2`, done **a couple of seconds after commit `1`** reverts the changes from commit `1`
    - code examples workflow triggers

The thing that can happen is the workflow from `commit 1` working on changes from `commit 2`, as it compares the state **of the branch**, not the actual commit it was triggered by. I tried to find a quick solution in the documentation of the `dorny/paths-filter` action, but didn't find anything at this point.
Do you think it's worth spending more time on the task, or should we leave it for now?

.github/workflows/code-examples.yml
warpech: Explained in https://github.com/handsontable/handsontable/pull/9184#issuecomment-1028771754

warpech (approved):
Thank you @jansiegel. No more questions :)

PR #9201, opened 2022-02-07
[RTL] Abstract the editor's positioning logic

### LINKED ISSUE #9194: [RTL] Hide the implementation details from the Cell editor page
### Description

The [cell-editor](https://handsontable.com/docs/cell-editor/) page contains some examples that reveal the internal implementation details of the cell editors. That needs to be changed to give developers only examples that promote using only stable public APIs.

After the [PR ](https://github.com/handsontable/handsontable/pull/9176) is merged and released the examples (in the current shape) will stop work.

### Links

https://handsontable.com/docs/cell-editor/#selecteditor-creating-editor-from-scratch

budnix:
### Context

The PR moves the cell editor size and positioning logic to the `getEditedCellRect` method of the `BaseEditor` class. To support that kind of calculation the editors needed to implement complex logic, mostly integrated with internal Handsontable APIs. Moreover, some of that logic was mentioned in our docs. This definitely, it's not a good practice to show developers a chunk of code that uses internal/not stable APIs, the PR fixes that as well.

And here is the `getEditedCellRect` method. The method reduces DRY ([Text](https://github.com/handsontable/handsontable/blob/master/handsontable/src/editors/textEditor/textEditor.js#L344-L438) and [Select](https://github.com/handsontable/handsontable/blob/master/handsontable/src/editors/selectEditor/selectEditor.js#L193-L258) editors have the same implementation) and allows developers to position their editors perfectly right above the edited cell, no matter if the edited cell belongs to the _main_ or _top_, _left,_, _bottom_, ... overlay.

The method returns an object with the following properties:
   *  `top` The top position relative to the table viewport;
   *  `start` The left (or right in RTL) position relative to the table viewport;
   *  `width` The cell's current width;
   *  `maxWidth` The maximum cell's width after which the editor goes out of the table viewport;
   *  `height` The cell's current height;
   *  `maxHeight` The maximum cell's height after which the editor goes out of the table viewport

To calculate correct values for cells that are edited within overlays I needed to implement a new `getOverlayPosition` method. The method works similar to the `getScrollPosition` method but the returned values are different depending on the type of the scrollable element. When Handsontable knows its size its overlays position is always `0`. The overlays are stuck to the table's edge and in that case, the `getScrollPosition` only matters. When Handsontable is initialized without size the Window object becomes the scrollable element. In that case, while scrolling the browser viewport the overlays only moves when the table's viewport leaves the viewport of the browser - to keep the overlays stuck to the browser's viewport edge. The scroll position from `getScrollPosition` method wouldn't fit here as it calculates the value always from left/top `0`. The `getOverlayPosition` method does the job by returning the scroll position minus parent offset.

One of the biggest pros of the method is the fact the complexity of the logic behind the overlay positioning is drastically reduced (for example [InlineStart](https://github.com/handsontable/handsontable/pull/9201/files#diff-7b6c4add80975ed579e385bf8f3d894fc22f819f80ffafc22afc908d73d471feL71) overlay or [TopInlineStartCorner](https://github.com/handsontable/handsontable/pull/9201/files#diff-f0a2efa330c007a42ef3773439822b87f524003ca61c92ae22ce41ce7b3e6589L65) overlay).

### Handsontable knows its size (Window is **not** a scrollable element)
Kapture 2022-02-10 at 13 47 08

### Handsontable does not know its size (Window becomes a scrollable element)
Kapture 2022-02-10 at 13 45 47

About the changes to the documentation, I removed the code from the examples where the internal API is used. I followed the [cell editor](https://handsontable.com/docs/cell-editor/) tutorial and I can confirm that the example Select editor with PR's changes worked great.

_[skip changelog]_

### How has this been tested?

I tested the changes locally having in my mind that the positioning should work for LTR, RTL, with and without defining table size and after the table is scroll or not. All the cases are covered with new E2E tests.

### Types of changes

### Related issue(s):
1. resolves #9194 
2.
3.

### Affected project(s):
- [x] `handsontable`

### Checklist:


jansiegel:
@budnix,
I'm not sure about the name of the `getOverlayPosition` method - the 'position' part, to be specific.

While it makes sense in the scroll context (scroll is physically moved along one axis, so the word "position" describes what's happening pretty well), I don't think it works just as well for the overlays. The overlays are still DOM elements, so when I think of their _position_, I imagine `x` and `y` coordinates of some sort. And there's the question of what's their position relative to, etc.

Maybe using the word `offset` would work better here? It would imply that they're _offset_ by a number of pixels from their original position, along the axis in which it's possible for them to be offset? 

I like the `getEditedCellRect` method though, I feel it should simplify lots of things in the future 👌

budnix:
> I'm not sure about the name of the `getOverlayPosition` method - the 'position' part, to be specific.

Thanks for the review. I agree with you, so the name `getOverlayOffset` would be enough understandable?

jansiegel:
> > I'm not sure about the name of the `getOverlayPosition` method - the 'position' part, to be specific.
> 
> Thanks for the review. I agree with you, so the name `getOverlayOffset` would be enough understandable?

It'd be alright with me 👍

jansiegel (approved):

PR #9229, opened 2022-02-22
Fix CSS styles for RTL layout

### LINKED ISSUE #9226: [RTL] `layoutDirection: 'rtl'` does not work with HTML `dir=ltr` mode
### Description

Setting the `layoutDirection: 'rtl'` option forces the table to be rendered from right to left and it works. But some of the UI elements are not positioned correctly. The issue is caused by incorrect Handsontable CSS selectors. The RTL selectors (`[dir-rtl] ...`) apply the rules only to the Handsontable parent's element, but in fact, they should be applied to the root element of the table.

As the result, some of the UIs can be misaligned, mispositioned, and not perfectly styled
Kapture 2022-02-22 at 08 26 13
(here, the manual column resize guide/ruler element is misplaced and the context menu has a too wide a border on the right)
 
The fix: All styles such as
```css
[dir=rtl] .htCommentCell:after {
  border-right: 6px solid transparent;
  border-left: none;
}
```
should be adjusted to:
```css
[dir=rtl].handsontable .htCommentCell:after {
  border-right: 6px solid transparent;
  border-left: none;
}
```

budnix:
### Context

The PR fixes an issue with CSS for RTL layout direction mode. Our SASS loader transforms CSS logical values (`inset-inline-end` etc.) to physical ones (`right` etc.) and modifies the selectors to make it work for RTL. For example from:
```css
.htCommentsContainer .comment {
  inset-inline-end: 0;
  inset-inline-start: unset;
}
```
to 
```css
[dir=rtl] .htCommentsContainer .comment {
  right: 0;
  left: unused;
}
```
And here is the issue. That CSS selector only works for configurations where the whole document is RTL `<html dir="rtl">` and the `layoutDirection` is set as `'inherit'` or `'rtl'`. In case the developer wants to change opposite direction than the HTML document, the table UI is mispositioned. The fix changes the [generated CSS selectors](https://github.com/handsontable/handsontable/pull/9229/files#diff-e2c999220a59179e554bdf24a7495974c120c42b47df26578571e0738dd5d1a0R23) from `[dir=rtl] .htCommentsContainer .comment` to `[dir=rtl].htCommentsContainer .comment`. Thanks to that the selector applies to the Handsontable root element that is controlled by Handsontable itself and not by the external element.

So, the PR allows to render the table correctly with these configurations:
```html
<html dir="rtl"> <-- RTL document mode -->
  <body>
    <div id="app"></div>
  </body>
</html>
```
```js
{
  data: Handsontable.helper.createSpreadsheetData(100, 25),
  fixedRowsTop: 2,
  fixedRowsBottom: 2,
  layoutDirection: 'ltr', // render the table in LTR mode
}
```
and
```html
<html> <-- LTR document mode -->
  <body>
    <div id="app"></div>
  </body>
</html>
```
```js
{
  data: Handsontable.helper.createSpreadsheetData(100, 25),
  fixedRowsTop: 2,
  fixedRowsBottom: 2,
  comments: true,
  layoutDirection: 'rtl', // render the table in RTL mode
}
```

_For reviewer:_ I faced a problem where I have to test something with two configurations where their results are the same. I didn't want to copy&paste the tests so with [inspiration](https://www.ontestautomation.com/data-driven-javascript-tests-using-jasmine/) I extended the tests and introduced the [data-driven tests for Jasmine](https://github.com/handsontable/handsontable/blob/feature/issue-9226/handsontable/test/helpers/common.js#L163).

For example:
```js
describe('TextEditor', () => {
  using('input value', [1, '1', true], (value) => {
    it('should correctly display the value in the textarea element', {
       // expect()
    });
  })
})
```
This test suite would generate the following test cases:
 * TextEditor using input value: 1 should correctly display the value in the textarea
 * TextEditor using input value: '1' should correctly display the value in the textarea
 * TextEditor using input value: `true` should correctly display the value in the textarea element

_[skip changelog]_

### Types of changes

### Related issue(s):
1. fixes #9226
2.
3.

### Affected project(s):
- [x] `handsontable`

### Checklist:

### DIFF
--- a/handsontable/.config/loader/sass-rtl-loader.js
+++ b/handsontable/.config/loader/sass-rtl-loader.js
@@ -5,7 +5,7 @@ const trim = (properties) => {
 };
 
 
-const polyfillDirection = (direction, source) => {
+const toPhysicalDirection = (direction, source) => {
   switch (direction) {
     case 'ltr':
       return source
@@ -20,13 +20,14 @@ const polyfillDirection = (direction, source) => {
   }
 };
 
-const appendRtl = (all, selector, properties) => `
-${polyfillDirection('ltr', `${selector}{${properties}}`)}
+const appendRtl = (all, selector, properties) => {
+  const rtlSelector = selector.split(',').map(sel => `[dir=rtl]${sel.trimStart()}`).join(', ');
 
-[dir=rtl] {
-${polyfillDirection('rtl', `${selector}{${trim(properties)}}`)}
+  return `
+    ${selector} {${toPhysicalDirection('ltr', properties)}}
+    ${rtlSelector} {${toPhysicalDirection('rtl', trim(properties))}}
+  `;
 }
-`;
 
 const applyRtlStyles = (source) => {
   return source.replace(
--- a/handsontable/src/3rdparty/walkontable/src/overlay/_base.js
+++ b/handsontable/src/3rdparty/walkontable/src/overlay/_base.js
@@ -279,6 +279,7 @@ export class Overlay {
     const tableParent = wtTable.wtRootElement.parentNode;
 
     clone.className = `${CLONE_CLASS_NAMES.get(this.type)} handsontable`;
+    clone.setAttribute('dir', this.isRtl() ? 'rtl' : 'ltr');
     clone.style.position = 'absolute';
     clone.style.top = 0;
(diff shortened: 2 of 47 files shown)


jansiegel (approved):
I like the `using` function, very elegant 👌

Issue #9214, opened 2022-02-14
[Regression] Row pre-populating demo does not work and it throws an error


AMBudnik:
### Description

Row pre-populating demo does not work and it throws an error. 
Preview is not visible
![Zrzut ekranu 2022-02-14 o 15 10 04](https://user-images.githubusercontent.com/10757813/153879576-3bbc095e-1a35-458f-8787-bfffc7097ca1.png)

Error
```js
Uncaught SyntaxError: Unexpected end of input
    at n (13.b8eb9bbf.js:1:333)
    at a.mounted (13.b8eb9bbf.js:1:434)
    at We (app.b3b37a2f.js:3:100938)
    at nn (app.b3b37a2f.js:3:114986)
    at Object.insert (app.b3b37a2f.js:3:110058)
    at O (app.b3b37a2f.js:3:146265)
    at a.__patch__ (app.b3b37a2f.js:3:147665)
    at a.e._update (app.b3b37a2f.js:3:124098)
    at a.l (app.b3b37a2f.js:3:154519)
    at hn.get (app.b3b37a2f.js:3:116417)
```

It works well when we click 'edit' button that redirects to JSfiddle.

### Links

URL https://handsontable.com/docs/row-prepopulating/ - v11.1 broken
https://handsontable.com/docs/11.0/row-prepopulating/ - v11.0 working

warpech:
Last change was in this commit that added a comment at the end of the snippet: https://github.com/handsontable/handsontable/commit/2762827fee67669d746e3a62db4d985fee5fa97b#diff-721f403c0a213467a1bbbb57b1cf6407b2145d3b514062db04fa16e86ff7aab7

I can see that other pages changed in that commit are affected:
- https://handsontable.com/docs/autofill-values/

The problem exists in `11.1.0` and `next` docs.

kirszenbaum:
Fixed in #9261, closing

PR #9261, opened 2022-03-01
Docs: Fixing a docs regression

kirszenbaum:
This PR:
- Moves an inline comment that caused a few demos to fail (fixing #9214)

Affected pages:
- http://localhost:8080/docs/next/autofill-values/#autofill-in-all-directions
- http://localhost:8080/docs/autofill-values/#autofill-in-all-directions
- http://localhost:8080/docs/row-prepopulating/#example
- http://localhost:8080/docs/next/row-prepopulating/#example

[skip changelog]


AMBudnik (approved):
Both pages 
* autofill
* row pre-populating
* 
are now showing the preview. Thanks @kirszenbaum 👍

PR #9269, opened 2022-03-04
Add VuePress plugin that allows generate correct canonical URLs

budnix:
### Context

The PR fixes the canonical URLs for docs. The change fixes:
 * All canonical URLs from now on have the slash on the end of the URL. So the URL is the same as in the browser;
 * All URLs point to the latest version (without specified docs version) except those that do not have an equivalent in the latest version. In that case, the canonical URL points to the latest docs version where the page exists. For example, the page `http://localhost:8000/docs/9.0/react-setting-up-a-locale/` exists until version `10.0` (it was renamed in `11.0`) thus the canonical URL is `http://localhost:8000/docs/10.0/react-setting-up-a-locale/`. The new page `http://localhost:8000/docs/11.0/react-setting-up-a-language/` exists in the latest version thus the canonical URL is `http://localhost:8000/docs/react-setting-up-a-language/`.

_[skip changelog]_

### Types of changes

### Related issue(s):
1. https://github.com/handsontable/handsontable.com-v2/issues/122
2. https://github.com/handsontable/handsontable.com-v2/issues/121
3.

### Checklist:


mkrjzah:
@budnix I've reviewed the changes.
It looks good.
I have a question regarding this renaming of pages.

> For example, the page http://localhost:8000/docs/9.0/react-setting-up-a-locale/ exists until version 10.0 (it was renamed in 11.0) thus the canonical URL is http://localhost:8000/docs/10.0/react-setting-up-a-locale/. The new page http://localhost:8000/docs/11.0/react-setting-up-a-language/ exists in the latest version thus the canonical URL is http://localhost:8000/docs/react-setting-up-a-language/.


Technically this is an SEO mistake.
Because in effect we get two pages (with other URLs) with the same content. 
That is content duplication.


I think that the pages are renamed rarely, so that is not a big issue, but can we address this in some way?

jansiegel (approved):

Issue #9327, opened 2022-03-18
[12.0.0] Restoring operation of Shift + Page up/down


aninde:
### Description

- Shift + Page up/down on Windows and macOS
- Shift + Fn + Arrow up/down on macOS 
should work with v12 as in v11.

### Actual result
Selection remains unchanged.

![Kapture 2022-03-22 at 11 16 25](https://user-images.githubusercontent.com/30186818/159461066-02d74096-faf0-41c0-8a0b-b56cc0f525f0.gif)

### Expected result
As page up/down. 
![Kapture 2022-03-22 at 11 27 15](https://user-images.githubusercontent.com/30186818/159461682-64607f74-2533-4041-ad37-e91f75ab0a32.gif)

### Demo

https://jsfiddle.net/oxk8md54/ v12.0.0 broken
https://jsfiddle.net/9g5x0amk/ v11.1 fixed

### Your environment
Handsontable version: 12.0.0
Browser name and version: Chrome 100.0.4896.20
Operating system: macOS 12.2.1, Windows 10

kirszenbaum:
As discussed, this fix won't be included in the changelog and release notes (@warpech @wszymanski @AMBudnik)

aninde:
Fixed on last RC-v12 https://jsfiddle.net/js_ziggle/28dzn6qg/

warpech:
Closing as done and verified by QA

Issue #9323, opened 2022-03-18
[12.0.0] Cmd ⌘ + Shift ⇧ + Arrow keys don't work


aninde:
### Description

Cmd ⌘ + Shift ⇧ + Arrow keys on macOS and Ctrl + Shift + Arrow keys on Windows don't work at all with v12.
![v12cmd-shift-arr2](https://user-images.githubusercontent.com/30186818/158996630-5db7d500-bb6e-441b-bf30-268888ea6cf4.gif)

With v11 it extended the selection by 1 (one) cell only in each direction.
![v11cmd-shift-arr](https://user-images.githubusercontent.com/30186818/158996597-3a13aaf1-0148-489d-a847-9f9726e84fe9.gif)

### Expected result
Restoring working on v11. Extending selection by one cell in each direction.

### Demo

https://jsfiddle.net/1vw2f79m/ v12
https://jsfiddle.net/toy2mear/ v11

### Your environment
* Handsontable version:
* Browser name and version:
* Operating system:

warpech:
Closing as done and verified by QA

Issue #9321, opened 2022-03-17
[12.0.0] Differences in how different shortcuts work comparing to the v11.X


aninde:
### Description

| Keyboard shortcut | Action in the v12-RC | Action in the v11.X |
|---|---| --- | 
| Alt or Option ⌥ + Arrow up/down/left/right | Selection remains unchanged | Selection move to the next cell in arrow direction |
| Alt or Option ⌥  + Fn + Arrow up or Alt + Page up | Selection remains unchanged | Selection move to the first cell in a column |
| Alt or Option ⌥  + Fn + Arrow down or Alt + Page down  | Selection remains unchanged | Selection move to the last cell in a column |
| Alt or Option ⌥  + Fn + Arrow left | Selection remains unchanged | Selection move to the first cell in a row |
| Alt or Option ⌥  + Fn + Arrow right | Selection remains unchanged | Selection move to the first cell in a row |
| Alt or Option ⌥  + Tab | Selection remains unchanged | Selection move to the next right cell, as with TAB |
| Alt or Option ⌥  + Tab + Shift | Selection remains unchanged | Selection move to the next left cell, as with TAB + Shift |

**v12-RC**
![v12](https://user-images.githubusercontent.com/30186818/158848760-a0322205-8653-4269-971b-b3326e97821d.gif)

**v11.X**
![v11](https://user-images.githubusercontent.com/30186818/158848126-bc1fa4ea-ae68-4300-8c37-30c12f456a74.gif)

### Demo

https://jsfiddle.net/1vw2f79m/ v12
https://jsfiddle.net/toy2mear/ v11

### Your environment
* Handsontable version: 12-RC
* Browser name and version: Chrome 100.0.4896.30, Microsoft Edge 99.0.1150.30
Operating system: macOS 12.2.1, Windows 10 Pro

aninde:
These shortcuts weren't officially supported by v11, so I'm closing the issue.

PR #9333, opened 2022-03-21
Added some extra shortcuts to fulfill specification (it should work the same in HOT 11.0.x)

wszymanski:
### Context

I changed it [according to specification](https://docs.google.com/spreadsheets/d/1fEMjQzM0AiPa9SZh68yKvMRDcsWsfRyXsAY7y4khY0s/edit#gid=0). It's a temporary solution, not the target behaviour.

[skip changelog]

### How has this been tested?

Manual tests

### Types of changes

### Related issue(s):
1. #9323 
2. #9327

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:


jansiegel:
@wszymanski 
I don't really know what is the desired result here.

1. #9327 :
    1. The title states it's about the combination of `Shift + Page Up/Down`
    2. The description mentions the combination `Shift + Fn + Arrow Up/Down`
    3. The gifs show the usage of `Cmd + Page Up/Down`
    
    It's not clear to me which of the combinations are we fixing and what is the expected result for them?
    Generally from what I can tell:
    1. `Shift + Page Up/Down` seems to work as in `11.x` (the selection gets moved a page up/down)
    2. `Shift + Fn + Arrow Up/Down` also works the same as it does in `11.x` but I'm not sure why we're testing combinations with the `fn` key, as it's pretty obscure, and there are multiple other combinations that need fixing which are actually probable real-life scenarios.
    3. `Cmd + Page Up/Down` **doesn't** work as it does in `11.x` - in `11.x` after pressing the combination two cells get selected - the one that was selected initially and the one the selection gets moved to after pressing the combination.
    
    And if we're checking combinations with `fn`, we might as well test the `alt/option` key, in which case:
    1. `Alt + Page Up/Down` **doesn't** work as it does in `11.x` - it does nothing
    2. `Alt + Shift + Page Up/Down` also does nothing

2. #9323
    1. `Cmd + Shift + Arrow Keys` work like in `11.x`
    2. It's not specified anywhere, but `Alt + Arrow Keys` blocks the selection movement, and it didn't in `11.x`
    
    
BTW - Can we assign those issues to the `12.0.0` release, so they won't get lost?

aninde:
> @wszymanski @aninde I don't really know what is the desired result here.
> 
> 1. [[12.0.0] Restoring operation of Shift + Page up/down #9327](https://github.com/handsontable/handsontable/issues/9327) :
>    
>    1. The title states it's about the combination of `Shift + Page Up/Down`
>    2. The description mentions the combination `Shift + Fn + Arrow Up/Down`

There is no `Page up` and `Page down` button on the built-in macbook keyboard. The `Fn + Arrow up / down` combination replaces it, hence its inclusion on the list.

>    3. The gifs show the usage of `Cmd + Page Up/Down`

You're right, I'm so sorry and I'm swapping gifs. I record the combinations `Shift + Fn + Arrow up` on them to prove my above comment. https://github.com/handsontable/handsontable/issues/9327

>    It's not clear to me which of the combinations are we fixing and what is the expected result for them?
>    Generally from what I can tell:
>    
>    1. `Shift + Page Up/Down` seems to work as in `11.x` (the selection gets moved a page up/down)
>    2. `Shift + Fn + Arrow Up/Down` also works the same as it does in `11.x` but I'm not sure why we're testing combinations with the `fn` key, as it's pretty obscure, and there are multiple other combinations that need fixing which are actually probable real-life scenarios.

For version from the original post they weren't working for me on Chrome 100.0.4896.30 as I reported.

>    3. `Cmd + Page Up/Down` **doesn't** work as it does in `11.x` - in `11.x` after pressing the combination two cells get selected - the one that was selected initially and the one the selection gets moved to after pressing the combination.
>    
>    And if we're checking combinations with `fn`, we might as well test the `alt/option` key, in which case:
>    
>    1. `Alt + Page Up/Down` **doesn't** work as it does in `11.x` - it does nothing
>    2. `Alt + Shift + Page Up/Down` also does nothing

Yes, I tested some `ALT + someting` combination, but but we agreed that they are out of scope and the issue was closed.
https://github.com/handsontable/handsontable/issues/9321
> 2. [[12.0.0] Cmd ⌘ + Shift ⇧ + Arrow keys don't work #9323](https://github.com/handsontable/handsontable/issues/9323)
>    
>    1. `Cmd + Shift + Arrow Keys` work like in `11.x`
>    2. It's not specified anywhere, but `Alt + Arrow Keys` blocks the selection movement, and it didn't in `11.x`

As above https://github.com/handsontable/handsontable/issues/9321

> BTW - Can we assign those issues to the `12.0.0` release, so they won't get lost?

Of course.

aninde:
I built and tested it on safari 15.3 (17612.4.9.1.8) macOS 12.2.1 (21D62) and Chrome 10 99.0.4844.74 / Windows 10.

Case https://github.com/handsontable/handsontable/issues/9323  `Cmd + Shift + Arrow keys` and `Ctrl + Shift + Arrow` keys works as we established on macOS. extended the selection by 1 (one) cell only in each direction. 
Both Cmd + and Ctrl + works on macOS.

Case https://github.com/handsontable/handsontable/issues/9327 works as expected also. Moving selection page up / down was restored. 

Ctrl + Home
Ctrl + End
Shift + Page up
Shift + page down
Shift + Home
Shift + End
Cmd + Home (tested on Chrome, on Safari, this is an assigned shortcut)
Cmd + End
work on macOS and Windows the same as in v11.X

Cmd + Page up and CMD + Shift + Page down works differently, and probably it's a breaking change or a bug.  In this version, there is no action after this. 
>    3. `Cmd + Page Up/Down` **doesn't** work as it does in `11.x` - in `11.x` after pressing the combination two cells get selected - the one that was selected initially and the one the selection gets moved to after pressing the combination.

> Yes, you're correct. In practice, this shortcut allowed the user to select data by `area hightlight`, but it did not copy all the selected ones, so I wonder if it makes any practical sense.
The same goes for `Cmd / Ctrl + Shift + Page up`. It also doesn't work in v12.
We don't have these abbreviations specified, so I think they are out of scope as well, like combinations with `Alt`.

jansiegel:
> Yes, you're correct. In practice, this shortcut allowed the user to select data by area hightlight, but it did not copy all the selected ones, so I wonder if it makes any practical sense.

The copying worked like with any other non-consecutively selected cells, so I don't think it's a reason to ditch it altogether...?

wszymanski:
> > Yes, you're correct. In practice, this shortcut allowed the user to select data by area hightlight, but it did not copy all the selected ones, so I wonder if it makes any practical sense.
> 
> The copying worked like with any other non-consecutively selected cells, so I don't think it's a reason to ditch it altogether...?
> 
> > We don't have these abbreviations specified, so I think they are out of scope as well, like combinations with Alt.
> 
> We may not have it specified now, but I think we should. `Cmd + [shortcut for any cell selection]` should probably work similarly to `cmd + click`?

I would do it within new issue and PR.

PR #9382, opened 2022-04-07
Update MomentJS dependency

### LINKED ISSUE #9381: Moment library update, adapt to the changes in Moment's parsing of dates
### Description
`handsontable` is currently using `moment` library of the version 2.24.0.
Affected versions of this package are vulnerable to Directory Traversal when a user provides a locale string which is directly used to switch moment locale.
Link to the known vulnerability in Snyk: https://security.snyk.io/vuln/SNYK-JS-MOMENT-2440688

### How to fix
Upgrade `moment` to version 2.29.2 or higher.

### Your environment

budnix:
### Context

The PR upgrades the `moment` library to 2.29.2. The PR applies all changes from the origin PR added by the bot (https://github.com/handsontable/handsontable/pull/9376), adds necessary changelog entry, and updates the `package-lock.json` file.

### Migration guide
If I remember correctly, some of the users that use the MomentJS library within their apps struggle with an issue where the library was double transpiled in the bundle file. One version for the developer's app and one for Handsontable dependency. The workaround for that was using the same version of the library as Handsontable.

The PR changes the version, so it can be advisable to mention this in the migration guide.

### Types of changes

### Related issue(s):
1. #9376
2. fixes #9381

### Affected project(s):
- [x] `handsontable`

### Checklist:


handsontable/src/plugins/columnSorting/__tests__/columnSorting.spec.js
```diff
@@ -1018,19 +1018,19 @@ describe('ColumnSorting', () => {
 
       getPlugin('columnSorting').sort({ column: 2, sortOrder: 'asc' }); // ASC
 
-      expect(getDataAtRow(0)).toEqual(['Mercedes', 'A 160', '01/14/2006', 6999.9999]);
-      expect(getDataAtRow(1)).toEqual(['Opel', 'Astra', '02/02/2004', 7000]);
-      expect(getDataAtRow(2)).toEqual(['BMW', '320i Coupe', '07/24/2011', 30500]);
-      expect(getDataAtRow(3)).toEqual(['Audi', 'A4 Avant', '11/19/2011', 33900]);
-      expect(getDataAtRow(4)).toEqual(['Citroen', 'C4 Coupe', '12/01/2008', 8330]);
+      expect(getDataAtRow(0)).toEqual(['Opel', 'Astra', '02/02/2004', 7000]);
```
warpech, line 1021: @budnix I need some root-cause analysis of why is this changed.

budnix: It seems that before updating the lib, the date values sorting does not work correctly. At least in that test case. For example, the results for sorting rows ascending:
Before (it's sorted without any noticeable pattern):
```js
'01/14/2006'
'02/02/2004'
'07/24/2011'
'11/19/2011'
'12/01/2008'
``` 
After:
```js
'02/02/2004'
'01/14/2006'
'12/01/2008'
'07/24/2011'
'11/19/2011'
```

warpech: I am sorry, but this is not the root cause. The root cause seems related to the fact that the test, despite its name, provided *wrongly formatted* data to Moment.js, and tested that Moment.js *does not* sort in that case. After the change, Moment.js sorts even when it is given wrongly formatted data.

Compare the format given in line 1009 and the data provided in lines 996-1002

https://github.com/handsontable/handsontable/blob/5ec0fa6500d6b20d941e79aee4d53d0a940b8885/handsontable/src/plugins/columnSorting/__tests__/columnSorting.spec.js#L993-L1009

warpech: Maybe we should use Moment.js in the strict mode? https://momentjs.com/guides/#/parsing/strict-mode/

budnix: > Maybe we should use Moment.js in the strict mode? https://momentjs.com/guides/#/parsing/strict-mode/

I would be for it. It's better to only sort the date-type values based on the `dateFormat`. Without leaving the room for guesses.

warpech:
We discussed on the today's daily (https://kb.handsontable.com/pages/viewpage.action?pageId=56461739) that @jansiegel will take over this PR.

According to the notes from the daily, we need:
1. much more tests to see what's changed
2. specification what should happen when `data` does not match `dateFormat` (awaiting @krzysztofspilka)

handsontable/src/plugins/columnSorting/__tests__/columnSorting.spec.js
```diff
@@ -1051,7 +1051,7 @@ describe('ColumnSorting', () => {
           {},
           {
             type: 'date',
-            dateFormat: 'mm/dd/yy'
+            dateFormat: 'MM/DD/YYYY'
```
warpech, line 1054: This file also uses the old notation `dateFormat: 'mm/dd/yy'` on lines 294, 1532 (and also in 2 places in `multiColumnSorting.spec.js`). 

Why were other lines not affected?

jansiegel: Those remaining examples seem to not be sorted by the column containing the dates, so the `dateFormat` is pretty much redundant, I'll change it, though.

handsontable/src/plugins/columnSorting/__tests__/columnSorting.spec.js
warpech: A little root cause analysis by me:

- in the [forgiving mode](https://momentjs.com/guides/#/parsing/forgiving-mode/), if Moment cannot parse a date, it uses `Date` constructor
- in Moment 2.24.0, the date `moment('01/14/2006', 'mm/dd/yy')` was not parsed at all. The `Date` constructor fallback returned the **current datetime** (https://jsfiddle.net/warpech/d9qako0g/4/)
- in Moment 2.25.0 (actually in this commit https://github.com/moment/moment/commit/0c7d8da2b7563363ef7e3de2b9a7023090d0462f), the character `yy` became significant and that date was parsed without resolving to the `Date` constructor. However, parsing is not good either. Moment parses this date to `2006-01-01`. (https://jsfiddle.net/warpech/d9qako0g/4/)
- with `strict` flag turned off, all tested Moment versions from 2.24.0 to 2.29.2 consistently say it is an incorrect date.

warpech: I was advised that there are parts of Handsontable that treat the forgiving mode as a feature and we should not remove it. Discussion: https://handsoncode.slack.com/archives/C4PV4S33Q/p1650384660407479?thread_ts=1649328227.856499&cid=C4PV4S33Q

kirszenbaum (changes requested):
@jansiegel – I'm adding my changes in #9409 

Just checking: should we update anyting in the `correctFormat` API ref? https://handsontable.com/docs/api/options/#correctformat

jansiegel:
@kirszenbaum 
> Just checking: should we update anyting in the `correctFormat` API ref? https://handsontable.com/docs/api/options/#correctformat

I don't think so - our API didn't change, what has changed is some logic on Moment's side, so until we have a more detailed specification on what exactly we should allow being automatically corrected, I'd leave it as it is.

kirszenbaum (approved):

Issue #9064, opened 2021-12-27
`hiddenColumns` and `hiddenRows` throw `handsontable.full.min.js:620 Uncaught RangeError: Maximum call stack size exceeded` for larger arrays of indexes to be hidden


AMBudnik:
### Description

I'm not sure if that is a limitation or an error. However, at the moment when the value of indexes is larger (more than `5500`) table does not load and we get the 

> Uncaught RangeError: Maximum call stack size exceeded

error in the console. The value that breaks the code changes. I was able to break it on `5700` but then on `5700` it worked and broke on `5710`. 

![Zrzut ekranu 2021-12-27 o 10 12 15](https://user-images.githubusercontent.com/10757813/147455780-9fa7226f-4a19-42ee-b49d-04e265b7ab65.png)
_full error message_

### `hiddenColumns`

1. Go to https://jsfiddle.net/xs3nj5vb/14/
2. Open the console to see the error
3. Change the value in the `Array`'s to a smaller number - like `5000`
4. Error is gone 

### `hiddenRows`

1. Go to https://jsfiddle.net/xs3nj5vb/16/
2. Open the console to see the error
3. Change the value in the `Array`'s to a smaller number - like `5000`
4. Error is gone and table loads column headers

### Your environment
* Handsontable version: 11
* Browser Name and version: Chrome 96
* Operating System: macOS Monterey 12.0.1.

Inform [Zen](https://handsontable.zendesk.com/agent/tickets/27454), [Zen](https://handsontable.zendesk.com/agent/tickets/29290)

AMBudnik:
Closed as solved in Handsontable v12.1.0.

PR #9490, opened 2022-05-16
Fix maximum call stack size in indexMapper

budnix:
### Context

The PR fixes the RangeError (maximum call stack size exceeded) that happens when the Handsontable uses more hidden rows and/or columns than ~6000 records. I found that the error happens in two places. 
 * In the public `getFirstNotHiddenIndex` method of the _IndexMapper_ class;
 * In the private method `findVisibleCoordsInRowsRange` of the _VisualSelection_ class;

The methods used the complexity of O(n) where combined with recursion caused the methods to explode. For setup where all 6000 rows are hidden calling the API as `hot.getFirstNotHiddenIndex(0, 1)` causes about 6 thousand recursion calls until the method throws the RangeError.

~The PR rewrites the method by using a private index mapper cache used to translate from visual to renderable indexes. Thanks to that there is no necessary to make recursion calls. I observed the performance boost. The method executes several to several times faster than before. However, the method signature also is changed. I found that the method accepts arguments values that are not used at all and that unnecessarily increases the logic complexity that we need to maintain.~

The PR is not a breaking change. The original method will stay in the public API till the next major release. But for now, it's marked as `@deprecated` and it prints the warning message when used. The new bug-free and more simplified `getNearestNotHiddenIndex` method from now on taking a place of the original method.

The new method implements the logic of searching for the nearest non-hidden index way more performant but also cutting down the API to the minimum. 

#### Before
```ts
getFirstNotHiddenIndex(fromVisualIndex: number, incrementBy: number, searchAlsoOtherWayAround: boolean, indexForNextSearch: number);
```

#### After
```ts
getNearestNotHiddenIndex(fromVisualIndex: number, searchDirection: -1|1, searchAlsoOtherWayAround: boolean);
```

### How has this been tested?

I tested the changes locally. I added unit tests for a new method and smoke tests related to the bug linked to the PR.

### Types of changes

### Related issue(s):
1. #9064

### Affected project(s):
- [x] `handsontable`

### Checklist:


warpech:
It was agreed today that we will fix the problem and omit a sudden breaking change by doing the following:

* we will implement a new method `getNearestNotHiddenIndex` and use it everywhere in our code from now on
    * because of the new method, the new release will be a minor (12.1.0)
    * the new name actually fits the purpose of the method better
    * the PR should include input for the Release Notes to say that this method replaces the old method and why is it better
    * a minor release does NOT need a Migration Guide
* we will stop using the old method `getFirstNotHiddenIndex` in our code
    * the same PR should mark the old method as deprecated and present the alternative in the API reference
    * in the method body, we will write a `console.warn` (that executes only once), saying that you are using a deprecated method
    * the old method will be deleted in a future major release. In that major release, there would be a Migration Guide saying which method is the successor to `getFirstNotHiddenIndex`
    * the tests of the old method will remain until it is fully deleted

.changelogs/9064.json
```diff
@@ -0,0 +1,7 @@
+{
+  "title": "Fixed maximum call stack size error in indexMapper by introducing the new `getNearestNotHiddenIndex` method.",
```
kirszenbaum, line 2:
```suggestion
  "title": "Fixed an issue where calling `getFirstNotHiddenIndex()` on a Handsontable instance with more than 6000 hidden rows or columns resulted in an uncaught `RangeError`, by adding a new method: `getNearestNotHiddenIndex()`. The `getFirstNotHiddenIndex()` method still works, but is marked as deprecated.
```

kirszenbaum: @budnix – from what I see in the in the PR's description, we could also present it as "Added", like this:

```
  "title": "Added `getNearestNotHiddenIndex()`, a new method that finds a visual index of the nearest not-hidden row or column, and works even with large numbers of hidden rows or columns. The previous method, `getFirstNotHiddenIndex()` still works, but is marked as deprecated.
```

kirszenbaum (changes requested):

.changelogs/9064.json
budnix: Thanks. I've created a new [PR](https://github.com/handsontable/handsontable/pull/9519) with your suggestions. I completely forgot to wait for your review 🤦

warpech: > I completely forgot to wait for your review 🤦

You shouldn't wait, as we haven't changed the process. But it is completely fine and expected that @kirszenbaum reviews the docs asynchronously to the PR.

PR #9567, opened 2022-06-07
Fix scrollable container element stretching for mobiles

### LINKED ISSUE #9475: [RTL] Scrolling on mobile slides out the table's body
### Description

When I drag-to-scroll the RTL table, it won't stop sliding until it's out of the table's viewport.

### Steps to reproduce

1. Open the https://handsontable.com/docs/layout-direction/ on a mobile device
2. Select "Preview" in the first demo to launch the 
3. Select any cell. Then tap on it and drag to scroll to the right, until all cells are gone.

IMG_9690

### Your environment
* Handsontable version: 12.0
* Browser name and version: Chrome 101
* Operating system: macOS Big Sur

budnix:
### Context

The PR fixes the bug where the container of the scrollable element was stretched. This causes the table to render more wide table viewport than necessary in RTL mode. The bug was caused by the CSS `left: -10000px` property that is used to "hide" mobile selection handlers for an editor. The fix is about changing the property to `right: -10000px` - for RTL (using `inset-inline-*` props). Thanks to that the selection handlers are always moved from the outside of the viewport in the opposite direction than the scroll direction thus no stretching viewport happens.

### How has this been tested?

I tested the changes locally on Chrome using the device toolbar (mobile emulation). I covered the fix with new E2E tests for LTR and RTL document modes.

### Types of changes

### Related issue(s):
1. fixes #9475

### Affected project(s):
- [x] `handsontable`

### Checklist:


wszymanski (approved):
It works properly 👍  I've checked in on physical Android 11 device with Chrome 102.0.5005.78. Works as desired.

Issue #9516, opened 2022-05-24
Docs: Corrections for the React transformations in the documentation.


jansiegel:
### Description
Some of the automatically transformed snippets/examples in the React "flavor" of the documentation aren't working properly.

- Some of the examples/snippets will need to be reorganized for the snippet transformer to process them correctly, 
- Some will need to be manually rewritten for vanilla/react versions.
- The transformer will most probably need to be extended to create code with a more "native" react feel.

Information about the incomplete examples is tracked here: [Google Spreadsheet](https://docs.google.com/spreadsheets/d/15q2Bv4UCIoXLktZDDqSTupTLSSO0BWvoxfJbeUbCauU/edit#gid=0).

Part of #9441.

warpech:
Closing as the automatic transformer was ditched. Thank you @jansiegel for the work nonetheless, I am happy that we used  the result of the transformer.

PR #9623, opened 2022-07-04
Add necessary changes to the React/docs transformer.

jansiegel:
### Context
Work in progress.

Known issues:
- The `grid-size` page throws an error, because it’s referring to an element removed by the transformer (the HOT initialization container)
- The `events-and-hooks` page throws an error, most probably because `addEventListener` is chained after `querySelector`, not invoked on a variable (best to be changed in the `.md`)
- Inline styles added to HTML elements break the resulting code, as React accepts inline styles as an object, not a string - it'd be best to move those styles to the CSS tab of the examples.
- There were changes made to `fixer.js`, so the JSFiddle preview doesn't work properly (it refers to `handsontable.com`)
- Some callbacks to events utilize `this` to access the calling element - it would be much safer to change them to `event.target`, as the context is different in case of the React event callbacks.
- As the transformer analizes the code line-by-line, callbacks referencing Handsontable instances _before_ the instances are created are not moved to `useEffect`. If possible, they should be moved to _after_ said Handsontable instances are initialized.
- The `npm run docs:build` script crashes because of the issues above - needs to be tested using `npm run docs:start`

### Types of changes

### Related issue(s):
1. #9516 

### Affected project(s):
- [ ] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

[skip changelog]


wszymanski:
> A problem that I noticed. If the server is started with the command `DOCS_VERSION=next npm run docs:start:no-cache`, then Handsontable previews do not work in the "JavaScript" version (see screenshot).
> 
> The problem does NOT happen if I start with: `DOCS_VERSION=next DOCS_FRAMEWORK=javascript npm run docs:start:no-cache`
> 
> Screen Shot 2022-07-13 at 4 40 19 PM

I've also encountered the problem, also on `feature/rect-docs` branch.

warpech:
For the future reference: it was agreed ([notes](https://kb.handsontable.com/pages/viewpage.action?pageId=63471961)) that this PR will not be merged. In fact, the transformer will be removed from the React docs branch in favor of manual maintenance of the React snippets.

warpech:
I did some work to update this PR with the current content in https://github.com/handsontable/handsontable/tree/feature/issue-9516-updated

warpech:
Closing as the automatic transformer was ditched. Thank you @jansiegel for the work nonetheless, I am happy that we used  the result of the transformer.

Issue #8460, opened 2017-02-06
define type="numeric" by giving renderer, editor and validator


stla:
Hello,

 I [figured out](http://stackoverflow.com/questions/42025001/strange-behaviour-of-comma-as-decimal-separator-in-handsontable) that `numbro(x).format()` unformats `x` when `x` is not valid. This sounds unexpected and nonsensical to me (getting an unformatting when one requires a formatting...)

Therefore, I tried to write a custom `type="numeric"` with another formatting.

And I realized that, contrary to the claim written [in the doc here](http://docs.handsontable.com/0.30.0/tutorial-cell-function.html), it is *not* true that `type="numeric"` is equivalent to 
```
{
      renderer: Handsontable.NumericRenderer,
      editor: Handsontable.editors.TextEditor,
      validator: Handsontable.NumericValidator
}
```

Indeed, try this fiddle: https://jsfiddle.net/stla/wj5yy0zu/6/

If you modifiy (edit) cell (0,0), say you type the number `14`, then the new value is the number `14` as expected.

But if you modifiy cell (1,1) and type the number `14`, then the new value is the **string** `"14"`: 

![capture](https://cloud.githubusercontent.com/assets/4466543/22639577/f110f8fa-ec4e-11e6-8981-ed83836bada0.PNG)

I would appreciate a hint to get numbers, not strings, after editing.

swistach:
Hi @stla, we moved documentation into a mono-repository. You can find more details in #8402.

kirszenbaum:
@AMBudnik - the documentation issue is still present in the new docs: https://handsontable.com/docs/cell-type/#usage

The whole **Cell type** page needs a thorough rewrite, but it's out of scope of this issue.

From what I understand, to fix this issue we should make the following changes to the docs:
- Add info that `NumericEditor` is basically the same as `TextEditor`
- Add info that even if you set a cell's `renderer`, `editor` or `validator` separately, you still need to set the cell's `type`. From what I understand, this applies to every cell type (not just the `numeric` cell type).

Based on the above conclusions, I prepared the following PR: #9697

If my conclusions are wrong, please let me know here or in the PR, thanks!

AMBudnik:
@kirszenbaum yes, the https://github.com/handsontable/handsontable/pull/9697 covers the subject entirely.

warpech:
This issue was fixed in PR #9697, but then that fix was reverted in PR #9928.

PR #9697, opened 2022-07-27
Docs: Add a note about setting the type option

kirszenbaum:
This PR:
- Adds info that `NumericEditor` is basically the same as `TextEditor`
- Adds a note saying that even if you set a cell's `renderer`, `editor` or `validator` separately, you still need to set the cell's `type`.

Modified pages:
- http://localhost:8080/docs/next/api/options/#editor
- http://localhost:8080/docs/next/api/options/#renderer
- http://localhost:8080/docs/next/api/options/#type
- http://localhost:8080/docs/next/api/options/#validator
- http://localhost:8080/docs/next/cell-function/
- http://localhost:8080/docs/next/cell-renderer/
- http://localhost:8080/docs/next/cell-editor/
- http://localhost:8080/docs/next/cell-validator/
- http://localhost:8080/docs/next/cell-type/

[skip changelog]

### DIFF
--- a/docs/content/guides/cell-functions/cell-editor.md
+++ b/docs/content/guides/cell-functions/cell-editor.md
@@ -15,6 +15,17 @@ Handsontable separates the process of displaying the cell value from the process
 
 This tutorial will give you a comprehensive understanding of how the whole process of cell editing works, how Handsontable Core manages editors, how editor life cycle looks like and finally - how to create your own editors.
 
+::: tip
+You can set a cell's [`renderer`](@/api/options.md#renderer), [`editor`](@/api/options.md#editor) or [`validator`](@/api/options.md#validator) individually, but you still need to set that cell's [`type`](@/api/options.md#type). For example:
+
+```js
+renderer: Handsontable.NumericRenderer,
+editor: Handsontable.editors.NumericEditor,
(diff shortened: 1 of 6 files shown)


warpech:
I think that this PR wrongly solves #8460 and should be reverted:

1. Even after this PR, the "Cell type" page still has the incorrect text underlined on the screenshot below, which is directly the text the user complained about: 
Screen Shot 2022-09-16 at 4 30 33 PM

2. A new tip is added in many wrong places. The problem is not related to the numeric renderer, editor, or validator, so why is the tip there? The problem is related only to the fact that a cell type is more than just renderer+editor+validator. For example, the cell type is directly used by the Filter plugin (when conditionally comparing) or in the Column Sorting plugin (when sorting). We need to remove the misleading text from in the docs, not just add tips beside them.

3. The "Cell type" page is very chaotic and could be improved with small effort. It starts speaking about the dates, then suddenly switches to use `numeric` as an illustration. The purpose of this page is not really to talk about the `numeric` cell type, but only to use it as an example of how to define your own cell type.

warpech:
A PR that reverts this PR: https://github.com/handsontable/handsontable/pull/9928

Issue #8556, opened 2020-10-09
dataSchema - add info that it can be a function


AMBudnik:
### Description
Here https://handsontable.com/docs/tutorial-data-sources.html we define that the `dataSchema` can be defined as a function, however, we do not mention the same in the main [`dataSchema` docs definition](https://handsontable.com/docs/8.1.0/tutorial-data-sources.html).

We should
* add this information
* consider adding the description for the mystery options mentioned in the link
![Zrzut ekranu 2020-10-9 o 15 14 53](https://user-images.githubusercontent.com/10757813/95587318-33510880-0a42-11eb-823a-62a811fbd209.png)

The issue has been reported as a part of the handsontable/support#254 investigation

### Links
* dataSchema in the docs -https://handsontable.com/docs/8.1.0/tutorial-data-sources.html

kirszenbaum:
@AMBudnik – I'm addressing this issue here: #9700.

I looked through the [linked page](https://handsontable.com/docs/8.1.0/tutorial-data-sources.html), but I'm not sure what the "mystery options" are, though ;) If you know, please let me know.

AMBudnik:
@kirszenbaum haha, the `mystery` refers to the following line

>See [data-schema](https://docs.handsontable.com/tutorial-data-sources.html#page-data-schema) for more options.

where we do not know what stand for the `option` in this description :)

PR #9700, opened 2022-07-28
Docs: Add info that `dataSchema` can be a function

kirszenbaum:
This PR:
- Adds info that `dataSchema` can be a function (fixing #8556)

[skip changelog]


AMBudnik (approved):
That's so much better.

Issue #8547, opened 2020-02-05
Numeric: comma as separator only during editing


aninde:
https://handsontable.com/docs/7.3.0/demo-numeric.html

> For example, both 500000.5, **500000,5** will be accepted. You are not able to use thousands separator in the editor. 

comma as separator can be used, but not during initialization. only when editing.

  We should add this because it may mislead users.

kirszenbaum:
@aninde - the issue is still valid in the current docs: https://handsontable.com/docs/numeric-cell-type/#usage

I'm addressing it here: #9701.

The whole **Numeric cell type** page should be thoroughly rewritten, but it's out of scope here.

PR #9701, opened 2022-07-28
Docs: Add info that at initialization you can't use the comma as a decimal separator

kirszenbaum:
This PR:
- Adds info that at initialization you can't use the comma as a decimal separator (fixing #8547)
- Makes minor improvements to the **Numeric cell type** page
- Makes minor improvements to the **Options** API ref wording

Affected pages:
- http://localhost:8080/docs/next/numeric-cell-type
- http://localhost:8080/docs/next/api/options

[skip changelog]


aninde (approved):
👍  Adjustments are in the right places and from now on I understand the differences between `every` and `each`.

Issue #8495, opened 2019-10-03
Change: Missing information about default table rendering without data


aninde:
**Change**
We should add information about the default rendering of Handsontable.
Without any data or when `data = null` the table is rendering default with 5 rows and 5 columns. 
It can be related to `minCols` and `minRows`. It needs to be checked.

**Where**
We can add it to the description of `Options.html#data` or `Understand binding as a reference` or maybe `Data sources`.

**Demo**
v. `7.1.1` https://jsfiddle.net/aninde/24yqos6z/

kirszenbaum:
@aninde - seems like this issue is partially solved: https://handsontable.com/docs/binding-to-data/#the-data-configuration-option

> You will probably want to initialize the table with some data (if you don't, the table will render a 5x5 empty grid for you).

But I'm adding that info in other places as well: #9703.

PR #9703, opened 2022-07-28
Docs: Add info on what happens when there's no data set

kirszenbaum:
This PR:
- Adds info on what happens when there's no data set (fixing #8495)

Affected pages:
- http://localhost:8080/docs/next/binding-to-data/#no-data (new section)
- http://localhost:8080/docs/next/api/options/#data

[skip changelog]


aninde (approved):
👍

Issue #8492, opened 2021-07-15
New docs: We should semantically connect `read-only` and `readOnly` keywords


AMBudnik:
### Description

We should semantically connect `read-only` and `readOnly` keywords. At the moment, we have the option `readOnly` but the tutorial is available only when we type `read-only` https://handsontable.com/docs/disabled-cells/#non-editable-specific-cells 

![image](https://user-images.githubusercontent.com/10757813/125752190-bd2f4b99-d334-4923-8c1b-667fcb6447d4.png)

### Your environment
* Handsontable version: 9+
* Browser Name and version: Chrome 91
* Operating System: Windows 10

kirszenbaum:
Reproduced on Production:
<img width="453" alt="1" src="https://user-images.githubusercontent.com/29247866/181720423-de69a717-d01b-4603-bf3c-15c237957327.png">

Fixed in #9707:
<img width="456" alt="2" src="https://user-images.githubusercontent.com/29247866/181720442-d8c93b19-c7ac-4edf-b0cf-b914432d4e29.png">

PR #9707, opened 2022-07-29
Docs: Add a `readOnly` search tag

kirszenbaum:
This PR:
- Adds a `readOnly` search tag to the **Disabled cells** page

[skip changelog]

### DIFF
--- a/docs/content/guides/cell-features/disabled-cells.md
+++ b/docs/content/guides/cell-features/disabled-cells.md
@@ -5,6 +5,7 @@ permalink: /disabled-cells
 canonicalUrl: /disabled-cells
 tags:
   - read-only
+  - readonly
   - noneditable
 ---
 


aninde (changes requested):
That's ok, but I would add this tag to the second read-only instance, i.e. https://handsontable.com/docs/comments/#make-a-comment-read-only as displayed in the search list in [OP](https://github.com/handsontable/handsontable/issues/8492#issue-945114732).

AMBudnik (dismissed):
Approved. A quick thought here - some people also use the 'locked cell' term for a cell that does not allow to change its value :)

kirszenbaum:
> Approved. A quick thought here - some people also use the 'locked cell' term for a cell that does not allow to change its value :)

@AMBudnik Thanks! Added: [309c37f](https://github.com/handsontable/handsontable/pull/9707/commits/309c37f96a6d29d198a55dd350c55904f1c76661)

> That's ok, but I would add this tag to the second read-only instance, i.e. https://handsontable.com/docs/comments/#make-a-comment-read-only as displayed in the search list in [OP](https://github.com/handsontable/handsontable/issues/8492#issue-945114732).

@aninde – good idea. Unfortunately, when I add the `readonly`/`read-only` tags to the [Comments](https://handsontable.com/docs/comments/#make-a-comment-read-only) page, its results dominate the search box:



Expanding the number of search results displayed doesn't help:



If read-only cells are a more popular use case than read-only comments, I wouldn't add the `readonly`/`read-only` tags to the **Comments** page, then.

AMBudnik (approved):
Great. Thanks!

aninde (approved):
@kirszenbaum I see. Then we leave it as it is.

Issue #9380, opened 2022-04-07
Improve `copyable` option description


AMBudnik:
### Description

In the documentation, it is said that 

![Zrzut ekranu 2022-04-7 o 10 21 39](https://user-images.githubusercontent.com/10757813/162154175-70b8746b-9d65-4343-bff3-1d50bd0f0221.png)

At the moment when we set `copyable` to `false` we do not save the value in the clipboard. But instead, we add an empty value in its place. So the clipboard is altered and `beforeCopy` hook is triggered.
That is why we cannot agree with the statement that turning the option to `false` `[...] disable copying for that cell`.

### Links

`copyable` option - https://handsontable.com/docs/api/options/#copyable
demo for tests - https://jsfiddle.net/8mvL64pf/6/

kirszenbaum:
Fixed in [9724](https://github.com/handsontable/handsontable/pull/9724/files#diff-d6f82a465a0c166188b543692fd3292bd450f09cc013dbbf3fa5dd31c6ffcb0cL1084-R1087)

Issue #8548, opened 2020-03-25
afterDeselect - false description


AMBudnik:
src https://handsontable.com/docs/7.4.2/Hooks.html#event:afterDeselect

The description informs us that once a current cell is deselected it runs the hook.

![Zrzut ekranu 2020-03-25 o 10 02 18](https://user-images.githubusercontent.com/10757813/77519414-ba140e80-6e7f-11ea-9edb-d0845cd00786.png)

However, it runs only when no cell is selected.

Related subject https://forum.handsontable.com/t/how-is-afterdeselect-fired/4251

kirszenbaum:
Fixed in [9724](https://github.com/handsontable/handsontable/pull/9724/files#diff-9c8662f02ad3bdda7b952d5aba2a3e3e1ee8ca2600f3bedffb0f3781022992a1R221)

Issue #8493, opened 2019-09-13
Add missing description of populateFromArray and readOnly.


aninde:
**Context**
WIP

**Description**
Add missing description of the behavior of cells that have the `readOnly` option enabled. They can't be changed by `populateFromArray`, because this method bypasses `cellMeta` with `readOnly`. 
This action is not intuitive for everyone, so it should be explained and described.

**Explanation (example from code of `populateFromArray`)**

```
if (cellMeta.readOnly) {
                current.col += 1;
                /* eslint-disable no-continue */
                continue;
              }
```
**Task to be done**
- [ ] add missing explanation to `populateFromArray`, with a link to `readOnly` option description
- [ ] add missing explanation to `readOnly`, with a link to populateFromArray method and different method to add data e.g. `setDataAtCell`

**Demo**
version 7.1.1 http://jsfiddle.net/aninde/qhot9c71/

**Your environment**
Handsontable version: 7.1.1
Browser Name and version: Google Chrome version 76.0.3809.100 
Operating System: macOS Mojave 10.14.6

AMBudnik:
Still valid for v9 of Handsontable https://jsfiddle.net/tp18967k/ and the new docs.
The `populateFromArray` method does not change the `readOnly` cell, which is not mentioned in the readOnly tutorial at https://handsontable.com/docs/disabled-cells/#overview

kirszenbaum:
Fixed in #9724

PR #9724, opened 2022-08-02
Docs: Various edits

kirszenbaum:
This PR:
- Adds info that the `populateFromArray()` method can't change read-only cells (fixing #8493)
- Fixes the description of the `copytable` option (fixing #9380)
- Fixes the description of the `afterDeselect` hook (fixing #8548)
- Makes other minor docs changes (fixing issues linked to this PR)

[skip changelog]


AMBudnik (dismissed):
@kirszenbaum 

Good bunch of changes 👍 

Approved with some ideas:
* if we mention that the non-editable cells aren't given any `className` IMO we should mention that those with `readOnly` are given `htDimmed` `className`.
* after checking the `indexMapper` at `getFirstNotHiddenIndex` I spotted that the deprecation signature is not very "obvious"

I suggest something like this instead
Screenshot 2022-08-02 at 16 41 43

kirszenbaum:
@AMBudnik

> if we mention that the non-editable cells aren't given any className IMO we should mention that those with readOnly are given htDimmed className.

Thanks! I summed up all differences in a table and moved it upfront ([ce73296](https://github.com/handsontable/handsontable/pull/9724/commits/ce73296dc698158775fde76e9d5de432dce2ba3d)):


> after checking the indexMapper at getFirstNotHiddenIndex I spotted that the deprecation signature is not very "obvious"
If we want to change the styling of the `@deprecated` mark, let's do it across the docs in a separate issue.

AMBudnik (changes requested):
I am unsure if I would put `Can't be changed manually` in those columns as it's the same for both + the table is described to show differences. But that's only a suggestion. What, IMO, has to be changed is that a readOnly cell attaches the `htDimmed` class and here we say that we attach two classes `htDimmed` and `className`.




The readOnly cell is given only `htDimmed` class
Screenshot 2022-08-04 at 10 11 48

kirszenbaum:
@AMBudnik - thanks! Changes made: [ed95697](https://github.com/handsontable/handsontable/pull/9724/commits/ed956978fca8f3f6ee354d2ad24c0ce965da6a2d)

AMBudnik (approved):

Issue #9723, opened 2022-08-02
Custom components are not visible after the second render caused by React StrictMode


bulaj:
New behaviour of StrictMode component introduced by React in version 18 causes react components to simulate their unmounting and remounting after first mount in developer mode (https://reactjs.org/docs/strict-mode.html#ensuring-reusable-state). This is meant to ensure that effects are properly cleaned in case of multiple mounting/destroying components.

The above causes custom rendered components in handsontable to disappear after the simulated unmounting and remounting of handsontable wrapper.
Notes:
- This behaviour occurs only in developer mode and when handsontable is under `StrictMode` component in components tree:
```jsx
<React.StrictMode>
    <SomeComponents />
    <Handsontable />
</React.StrictMode>
```
Usually, StrictMode is used to wrap whole application in it, so if we want to avoid double-mounting some workarounds are needed:
```jsx
<React.Fragment>
    <React.StrictMode>
        <SomeComponents />
    </React.StrictMode>
    <Handsontable />
</React.Fragment>
```
- Reacts StrictMode is used by default in `create-react-app`, and is quite heavily advised by its developer team.

jansiegel:
@bulaj Could you help me with reproducing the problem? I created a fresh app with `create-react-app` and implemented the `HotTable` component with a custom, component-based renderer wrapped with `React.StrictMode` (which wasn't there by default btw 🤔):

```jsx
    <React.StrictMode>
      <HotTable
        data={[
          [1, 2, 3],
          [1, 2, 3]
        ]}
        colWidths={800}
        rowHeaders={true}
        colHeaders={true}
        licenseKey="non-commercial-and-evaluation"
      >
        <RendererComponent hot-renderer/>
      </HotTable>
    </React.StrictMode>
```

And the renderer components seem to render fine. 
There is a nasty problem with them disappearing when scrolling (especially sideways), but it doesn't seem to be exclusive to `React.StrictMode` and it also happens on your PR.

What am I missing?

quick demo on codesandbox: https://codesandbox.io/s/gallant-grothendieck-2w14jr?file=/src/App.js

bulaj:
@jansiegel the problem is still persistent on master branch of component-comparison repo. Just change data loading to default state instead of in useEffect and change default page to all_grids in Layout.tsx.

AMBudnik:
This PR can be closed in the favor of https://github.com/handsontable/handsontable/pull/10263, released in Handsontable v12.3.3

PR #9575, opened 2022-06-13
Move the `@angular/core` package to peer dependencies

budnix:
### Context

The PR moves the `@angular/core` package to the peer dependencies. Updates the `npm audit` command from `npm audit --production` to `npm audit --omit=dev --omit=peer` - the previous argument is deprecated (`npm WARN config production Use `--omit=dev` instead.`) and add new Angular examples.

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature or improvement (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Additional language file or change to the existing one (translations)

### Related issue(s):
1.
2.
3.

### Affected project(s):
- [ ] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

- [ ] I have reviewed the guidelines about [Contributing to Handsontable](https://github.com/handsontable/handsontable/blob/master/CONTRIBUTING.md) and I confirm that my code follows the code style of this project.
- [ ] I have signed the [Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLScpMq4swMelvw3-onxC8Jl29m0fVp5hpf7d1yQVklqVjGjWGA/viewform?c=0&w=1)
- [ ] My change requires a change to the documentation.

PR #9731, opened 2022-08-02
Prevent clearing portal cache array in StrictMode second render (#9723)

bulaj:
### Context

This PR prevents the issue when custom rendered components are not visible after second render caused by react StrictMode in development mode. I have noticed that if render is forced by StrictMode renderedCellCache and portalCacheArray sizes are not equal and thats how i detect it.
Actual problem is caused by line
```this.portalCacheArray.length = 0;```
Length of the array is modified directly, outside of react's lifecycle and doesnt trigger methods necessary to render components properly.
Unfortunately, i'm not able to write unit test for this, because wrapper uses older version of react, which does not behave in same way as the newest.
 
### How has this been tested?

In demo application from components-comparison repository using reacts StrictMode component in 'all grids' section. Development mode only.
### Types of changes

### Related issue(s):
1. #9723 
2.
3.

### Affected project(s):
- [ ] `handsontable`
- [ ] `@handsontable/angular`
- [x] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

### DIFF
--- a/.changelogs/9723.json
+++ b/.changelogs/9723.json
@@ -1,0 +1,7 @@
+{
+  "title": "Prevent clearing portal cache array in StrictMode second render.",
+  "type": "fixed",
+  "issue": 9723,
+  "breaking": false,
+  "framework": "react"
+}
--- a/wrappers/react/src/hotTable.tsx
+++ b/wrappers/react/src/hotTable.tsx
@@ -526,7 +526,10 @@ class HotTable extends React.Component<HotTableProps, {}> {
     });
 
     this.hotInstance.addHook('afterViewRender', function () {
-      hotTableComponent.handsontableAfterViewRender();
+      // Prevent clearing portal cache array in StrictMode second render.
+      if (hotTableComponent.renderedCellCache.size === hotTableComponent.portalCacheArray.length) {
+        hotTableComponent.handsontableAfterViewRender();
+      }
     });
 
     // `init` missing in Handsontable's type definitions.


krzysztofspilka:
@jansiegel is it going to be reviewed?

jansiegel:
> @jansiegel is it going to be reviewed?

Sure, but it was put on hold to finish the React-Docs project and now it's not in the scope of `12.2`.

jansiegel (changes requested):
Also, I can see that the `Code Quality` CI check has failed - could you check what is that before merging?

warpech (changes requested):
I agree with what @jansiegel pointed out, that it is wierd that we don't have a test for this. @bulaj please add test(s). If that's not possible, why is that?

warpech:
> @warpech, maybe we should prepare a CI script that runs the wrapper tests in multiple React versions? This would prevent situations like that from happening in the future.

We definitely can do it any time. We already did that for Angular in https://github.com/handsontable/handsontable/pull/9575, because that PR was a turning point that could, in fact, break older versions. @jansiegel is this PR, in your opinion, such a turning point for the React wrapper?

jansiegel:
> @jansiegel is this PR, in your opinion, such a turning point for the React wrapper?

@warpech If I understand correctly, this problem exists only for the newer versions of React (@bulaj, correct me if I'm wrong). Although I don't think it's extremely severe, I _do_ think it's an good time to address the multi-version testing infrastructure.

bulaj:
> I agree with what @jansiegel pointed out, that it is wierd that we don't have a test for this. @bulaj please add test(s). If that's not possible, why is that?

As stated in PR description, this bug occurs only in react 18 version.

warpech:
To finish up the thought from the above comment: Our tests use React 17, so that's why a problem from React 18 is not reproducible.

AMBudnik:
Closed with https://github.com/handsontable/handsontable/pull/9731

Issue #9735, opened 2022-08-02
Docs: React docs are deployed on the staging environment as JavaScript docs


warpech:
### Description

When I did a manual deployment of the `feature/react-docs` branch on the staging environment (run https://github.com/handsontable/handsontable/actions/runs/2785104153), the React pages (under https://dev.handsontable.com/docs/next/react-data-grid/installation/) present something that is half-way between React and JavaScript:

- the H1 addition says "JavaScript Data Grid" instead of "React Data Grid"
- the `only-for` containers in Markdown think that they are in JavaScript, not in React

<img width="2672" alt="Screen Shot 2022-08-02 at 11 07 14 PM" src="https://user-images.githubusercontent.com/566463/182473512-d761242e-311f-4301-b803-0ead4e123e98.png">

krzysztofspilka:
@budnix please let's check if this issue is already solved by PR 9740

budnix:
I've deployed the latest `feature/react-docs` branch, and everything seems to work fine.

PR #9740, opened 2022-08-03
Removed unnecessary `docs:build:no-cache` script - every full build will be done without cache

wszymanski:
### Context

Both staging and production builds should be done without cache. It seems that it may be source of the problem for #9735.

Build can be checked using:

`npm run docs:review 66e31982602d80b63736e52c58f2700ff5512ca1`

[skip changelog]


docs/package.json
```diff
@@ -16,8 +16,7 @@
     "docs:scripts:generate-legacy-docs-versions": "node .vuepress/tools/generate-legacy-docs-versions.mjs",
     "docs:start": "npm run docs:scripts:link-assets && node --experimental-fetch ./node_modules/.bin/vuepress dev ./",
     "docs:start:no-cache": "npm run docs:scripts:link-assets && node --experimental-fetch ./node_modules/.bin/vuepress dev --no-cache ./",
-    "docs:build": "node .vuepress/tools/build.mjs",
-    "docs:build:no-cache": "npm run docs:build -- --no-cache",
```
warpech, line 20: Thank you so much, I wished for that for a long time.

Can you also remove the `no-cache` option from `start`? On our Slack you will find many stories where cache was troublesome there as well.

Also, @jansiegel can you check if this problem still exists? https://handsoncode.slack.com/archives/C034WPQUG9W/p1658917802885339?thread_ts=1658914565.788989&cid=C034WPQUG9W

warpech (changes requested):

warpech (approved):

Issue #9730, opened 2022-08-02
Docs: React-docs: Internal links for security certificates does not work


aninde:
### Description

From http://localhost:8080/docs/next/react-data-grid/security/#security-certificates and 
http://localhost:8080/docs/next/javascript-data-grid/security/#high-quality-code-pledge below links for certificates return 404. 
### Links

http://localhost:8080/docs/securitum-certificate.pdf - 404
http://localhost:8080/docs/seqred-certificate.pdf - 404

PR #9746, opened 2022-08-04
Replaced `{{$page.currentVersion}}` syntax with `$baseWith('')` for `.MD` files #9730

wszymanski:
### Context

As the title says. It should fix problems described within #9730. Please keep in mind that one `RegExp` provides escaping `'` sign within the `.MD` file for purpose of handling it properly by IDE.

_Edit by @budnix_

I modified the original idea where the `$baseWith()` pseudo function is used in favor of template variable substitution (`$basePath`). I think the function complicates the syntax of the content, especially the syntax of the JS and the React examples. Moreover, I updated examples where the handsontable.com host is hard coded. I used the relative paths to ensure that examples work in any staging (local, staging, and prod). 

The `$basePath` in watch mode contains only the docs version e.g `/docs/next` and in build mode it contains the selected framework e.g. `/docs/next/javascript-data-grid`

### How has this been tested?

I tested the URLs using the watch and build mode. I checked cell renderer and AJAX load/save examples, and I checked the security certificates' URLs mentioned in #9730.

_[skip changelog]_


warpech:
I recommend @budnix to be the reviewer next week.

docs/content/guides/getting-started/introduction.md
```diff
@@ -27,24 +27,24 @@ Explore Handsontable's features and source code:
 ## Guides for your framework
 
 <div class="row-items-container">
-  <Link href="../javascript-data-grid/" hide-latest-version class="row-item">
-    <Img class="integration-framework-logo" src="$withBase('/img/pages/introduction/javascript.svg')" alt="JavaScript logo" />
+  <Link href="/docs/{docsVersion}/binding-to-data/" hide-latest-version class="row-item">
```
jansiegel, line 30: These links point to the `react-data-grid` pages, while they should point to `javascript-data-grid`. (apart from the React one)

docs/content/guides/getting-started/introduction.md
```diff
@@ -27,24 +27,24 @@ Explore Handsontable's features and source code:
...
+    <Img class="integration-framework-logo" src="/img/pages/introduction/javascript.svg" alt="JavaScript logo" />
```
jansiegel, line 31: The ``'s `src` attribute is always treated as relative to `base`.  I'm not sure if that's correct, but I'm having trouble finding examples of this being problematic.

That's different from the native markdown's image and link syntax, where `{{$basePath}}` needs to be provided, though. Shouldn't they work more similarly?

jansiegel (changes requested):

docs/content/guides/getting-started/introduction.md
budnix: I reverted the changes within f1b38a15b2e707982d4c822fe671ace4ebfd0e3e. It should be fixed.

docs/content/guides/getting-started/introduction.md
budnix: When you change the `src` attribute to an absolute path like `https://handsontable.com/docs/img/pages/introduction/vue.svg`, the `Img` component will still be rendered correctly. Nothing needs to be done in this case, I think.

jansiegel (approved):
In a followup to https://github.com/handsontable/handsontable/pull/9746#discussion_r941247899:

I'm going to approve this PR, as everything seems to work fine, but I'm not sure about the discrepancies between linking to assets using the same URL, but different methods, for example:

### On `development` mode (`start`)
- Using `/img/pages/introduction/javascript.svg` as an URL:
  - `[link](/img/pages/introduction/javascript.svg)`
  - `link`
  - `<Link href="/img/pages/introduction/javascript.svg">link</Link>`
  
  All point to:`http://localhost:8080/img/pages/introduction/javascript.svg`
  
  - ``
  
  Points to: **`http://localhost:8080/docs/next/img/pages/introduction/javascript.svg`** 
  
  ### On `production` mode (`build`)
- Using `/img/pages/introduction/javascript.svg` as an URL:
  - `[link](/img/pages/introduction/javascript.svg)`
  - `link`
  - `<Link href="/img/pages/introduction/javascript.svg">link</Link>`
  
  All point to:`http://localhost:8080/img/pages/introduction/javascript.svg` (and it doesn't work in `production` context)
  
  - ``
  
  Points to: **`http://localhost:8080/docs/next/react-data-grid/img/pages/introduction/javascript.svg`**

---

There are probably similar differences when using `{{basePath}}`, and _additionally_ the `Link` component allows the usage of `{{docsVersion}}`, which is not available in other components (and is not actually the "version", but rather `[version]/[framework]-data-grid`).

It would be nice to get some consistency here, but I'm not sure if it's a part of this task.

Issue #9751, opened 2022-08-05
Docs: React docs: create a style guide for JSX snippets


warpech:
### Description

We should create a section in the https://github.com/handsontable/handsontable/blob/develop/docs/README-EDITING.md#markdown-containers file with information about the intended style of the JSX (React) code snippets and examples. 

What was agreed so far:
- we avoid using the `settings={}` prop, but use individual props instead, e.g. `<HotTable data={..}>`
- props which value is a string, should be provided in double quotes without curly braces. E.g. `<HotTable licenseKey="00000-00000-00000-00000-00000" />`
- in JavaScript objects inside the props, we use single quotes, e.g.:
  ```jsx
  <HotTable nestedHeaders={[
          ['A', { label: 'B', colspan: 8 }, 'C']
        ]} />
  ```
- the constant which value is a reference to the Handsontable component instance, should have the word `Ref` in its name, e.g. `const hotRef = useRef(null)` or `const hotTableComponentRef = useRef(null)`
  - `useRef` should be used instead of `React.createRef` in functional React
- the snippets should not include unneeded elements, e.g. a `<Fragment>` with one child
- we use named imports, e.g.
  ```jsx
  import { useRef } from "react";
  useRef(...);
  ```

Source of these decisions: https://github.com/handsontable/handsontable/issues/9737, https://kb.handsontable.com/pages/viewpage.action?pageId=63472068, https://handsoncode.slack.com/archives/C034WPQUG9W/p1659692176283389
  
Whenever in doubt, we might try to follow the new React docs: https://beta.reactjs.org/learn

kirszenbaum:
Done in #9770

AMBudnik:
Close as done.

PR #9770, opened 2022-08-11
Docs: Add React style guide

kirszenbaum:
This PR:
- Adds a basic React style guide (fixing #9751)

I called it **React style guide** (not **JSX style guide**), as it covers more than just JSX.

[skip changelog]


jansiegel (approved):
Looks great!

I'd add an entry about using `<>` instead of `<Fragment>`, but I'm not sure if it's been decided already. It used to not work in JSFiddle, [then it threw a warning](https://handsoncode.slack.com/archives/C034WPQUG9W/p1659709082298089?thread_ts=1659703638.321389&cid=C034WPQUG9W), now it seems to work alright for me.

Either way, I think we can merge it as is and extend the list in the future 👍

Issue #9737, opened 2022-08-03
React docs: Editing of the page content in a way that suits JavaScript and React


warpech:
This issue is to track editing of the content of all Markdown files with two framework variants in mind (JavaScript and React). **The goal of editing is to make the minimum changes needed for the pages to make sense for two framework variants**.

Some assumptions for the work:

1. The work is grouped by the groups of pages in the left sidebar
2. A single group should be a single PR
3. PRs are targeted at `feature/react-docs`
4. Scope of work:
    - change the non-interactive snippets to be framework-agnostic or make separate snippet variants for JS and React
    - edit the page text to sound good both for JS and React
    - do some necessary improvements in the examples that are marked as `TODO` in Markdown
    - make sure that the snippets and examples on the page are correct and work
    - improve formatting in the snippets and examples

The groups of pages are assigned to people in the spreadsheet: https://docs.google.com/spreadsheets/d/1OwV8GvITDOykgpyKrC7At61b67xQn--cYvAS6usImrs/edit#gid=0    

I assume one MD file might take up to 2 hours.

Challenges:
- most of the non-interactive snippets use a `new Handsontable` constructor that does not make sense in React
  - 👉 on a meeting 2022-08-03 we agreed to make a conditional `only-for react` snippet, that will use the React syntax (most often it will mean copying a fragment from an interactive example down the page)
- some pages have examples moved from "Integrate with React" section (in PR https://github.com/handsontable/handsontable/pull/9576). Verify if these additional examples make sense. Maybe they can be incorporated with other examples that also exist in the JS version, reducing the overall number of examples.
  - 👉 on a meeting 2022-08-03 we agreed to keep two similar examples on the same page, it can be refined later
    - e.g. a potential refinement is that the "Basic example" and "Formatting numbers" most probably have no substantial difference on the page http://localhost:8080/docs/next/react-data-grid/numeric-cell-type/#display-numeric-values and could be merged into one
- some pages are written in the "Tutorial" style, provided as Step 1, Step 2, etc. This might not be sustainable for two frameworks. Let's discuss simplifying that.
  - 👉 on a meeting 2022-08-03 we agreed:
      - to not change the interactive steps on pages "Column hiding" and "Row hiding"
      - to change the steps on page "Column summary" in the following way: Step 1 will be non-interactive but conditional for each framework using `only-for`, Steps 2-4 will only show small snippets with the exact lines that are changed from the previous step, Step 5 will be interactive
[...]

PR #9786, opened 2022-08-16
Docs: Edit the Tools and building section

kirszenbaum:
This PR edits the **Tools and building** section in line with the #9737 guidelines.

Updated pages:
- http://localhost:8080/docs/next/react-data-grid/modules/
- http://localhost:8080/docs/next/react-data-grid/custom-plugins
- http://localhost:8080/docs/next/react-data-grid/custom-builds/

IMO the following pages don't require any updates (React-wise), it's worth double-checking, though:
- http://localhost:8080/docs/next/react-data-grid/packages
- http://localhost:8080/docs/next/react-data-grid/testing
- http://localhost:8080/docs/next/react-data-grid/folder-structure/

[skip changelog]

### DIFF
--- a/docs/content/guides/tools-and-building/custom-plugins.md
+++ b/docs/content/guides/tools-and-building/custom-plugins.md
@@ -13,11 +13,15 @@ tags:
 
 [[toc]]
 
+Create a custom plugin that extends Handsontable's built-in functionalities.
+
 ## Overview
 
 Plugins are a great way of extending Handsontable's capabilities. In fact, most Handsontable features are provided by plugins.
 
-This guide shows you how to create a custom plugin.
+::: only-for react
+You can create a custom plugin in JavaScript, and then reference it from within your React app.
+:::
 
 ### 1. Prerequisites
 
--- a/docs/content/guides/tools-and-building/custom-plugins.md
+++ b/docs/content/guides/tools-and-building/custom-plugins.md
@@ -38,7 +38,6 @@ The [`BasePlugin`](@/api/basePlugin.md) interface takes care of:
 * Memory leak prevention
 * Properly binding your plugin's instance to Handsontable
 
-
 ```js
 export class CustomPlugin extends BasePlugin {
  /**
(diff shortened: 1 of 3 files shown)


docs/content/guides/tools-and-building/custom-plugins.md
```diff
@@ -267,15 +271,50 @@ const hotInstance = new Handsontable(container, {
...
+<HotTable
+  // Pass `true` to enable the plugin with default options.
+  [CustomPlugin.PLUGIN_KEY]={true}
+  // You can also enable the plugin by passing an object with options.
+  [CustomPlugin.PLUGIN_KEY]={{
+    msg: 'user-defined message',
+  }}
+  // You can also initialize the plugin without enabling it at the beginning.
+  [CustomPlugin.PLUGIN_KEY]={false}
+/>
```
budnix, line 290: AFAIK JSX does not support computed properties. I've changed the snippet to use the plugin key the same as the example above the section.
```suggestion
<HotTable
  // Pass `true` to enable the plugin with default options.
  customPlugin={true}
  // You can also enable the plugin by passing an object with options.
  customPlugin={{
    msg: 'user-defined message',
  }}
  // You can also initialize the plugin without enabling it at the beginning.
  customPlugin={false}
/>
```

docs/content/guides/tools-and-building/modules.md
```diff
@@ -166,9 +167,33 @@ new Handsontable(container, {
...
+
+const container = document.querySelector('#example1');
+
+<HotTable
+  columns={[
+    {
+    type: 'numeric'
+    }
+  ]}
+/>
```
budnix, line 190: Fix formatting :) 
```suggestion
<HotTable
  columns={[
    {
      type: 'numeric'
    }
  ]}
/>
```

budnix (changes requested):

budnix (approved):
👍

bulaj:
I think `packages` page from reacts point of view feels so out of this world that its almost pointless like this - don't think anybody will embed scripts on its page like this rather than using imports.
I'm not sure that we have a valid alternative for this page in reacts context. At most we could suggest something similar to 'Bare distribution' just using imports, but it doesn't feel particularly useful. And internalisation section that would be almost exact copy of what we have in languages section.

PR #9790, opened 2022-08-17
Added `nestedRows` tag to the page

AMBudnik:
It's more probable that one would type `nestedRows` than `nested rows` as `nestedRows` is the option's name. I found myself also doing the same thing.

Page: https://handsontable.com/docs/row-parent-child/#overview

[skip changelog]


kirszenbaum (approved):
Good idea!

PR #9821, opened 2022-08-23
Fix header anchoring

### LINKED ISSUE #9803: Docs: Remove # from the HTML text, use CSS instead
### Description

Currently, the `#` glyph, rendered on the left of a heading, is achieved as a text character in the HTML.

It should be removed from the HTML and achieved as an empty `<a>` element with CSS `::before` pseudo-element.

Screen Shot 2022-08-22 at 11 46 31 AM

### Links
Any page, for example https://handsontable.com/docs/

budnix:
### Context

To support the headers' anchors with Docs that have a top bar in its layout Vuepress uses a trick with expanding the headers using `margin` and `padding` (https://github.com/vuejs/vuepress/blob/master/packages/%40vuepress/theme-default/styles/index.styl#L112) CSS props. The solution has side effects which are manifested by the fact that the anchor shows up before hitting the header.

Kapture 2022-08-23 at 15 53 53

The PRs solution abandons the native Vuepress solution and uses the CSS approach - the `scroll-padding-top` property. Thanks to that, I can get rid of the hack, and the anchors show at the right moment.

#### Before
Kapture 2022-08-23 at 15 30 28

#### After
Kapture 2022-08-23 at 15 22 28

_[skip changelog]_

### Types of changes

### Related issue(s):
1. fixes #9803

### Checklist:

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -8,6 +8,7 @@ const nginxVariablesPlugin = require('./plugins/generate-nginx-variables');
 const extendPageDataPlugin = require('./plugins/extend-page-data');
 const firstHeaderInjection = require('./plugins/markdown-it-header-injection');
 const conditionalContainer = require('./plugins/markdown-it-conditional-container');
+const activeHeaderLinksPlugin = require('./plugins/active-header-links');
 const {
   getDocsBaseUrl,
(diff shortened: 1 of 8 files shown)


jansiegel (approved):
Seems to work alright 👍

PR #9829, opened 2022-08-24
Allow open dropdown menu when no column is selected

### LINKED ISSUE #9561: [11.0.0+] dropdownMenu open() method with filtering
### Description

It seems that there's a regression regarding `open()` method for dropdownMenu with the filters option turned on and it was introduced in Handsontable 11.0.0. Since that version opening `dropdownMenu` using that method throws an error:

`Uncaught TypeError: Cannot read properties of null (reading **'visualIndex')`**

### Steps to reproduce

1. Open the demo
2. See the console for error

### Demo

Case 1, v. 10.0.0 (works fine): https://jsfiddle.net/aszymanski/sz1j920L/4/
Case 2,  v. 11.0.0 (doesn't work): https://jsfiddle.net/aszymanski/sz1j920L/3/

### Your environment

* Handsontable version: 11 and 12
* Browser name and version: Google Chrome 101
* Operating system: MacOS Monterey 12.3.1

budnix:
### Context

The PR fixes an issue where the opening dropdown menu (when the _Filters_ plugin is enabled) causes errors. The PR restores the correct behavior from v10 and improves the UX usability by activating or deactivating the Filters menu components (such as `filter_by_condition` and `filter_by_value`) depending on the table's selection.

When the dropdown menu is opened while the table has no selection, all menu options are disabled, including the filter sections `by_value` and `by_condition`. They are empty, the `by_condition` component allows choosing only the "None" option:
Kapture 2022-08-25 at 13 18 59

When there is a selection (at least one cell), the menu entries including filter become active:
Kapture 2022-08-25 at 13 25 22

For non-contiguous selection, the filter plugin will always filter the column pointed by the selection highlight (a single cell with a 2px border). 

### How has this been tested?

I tested the changes locally, and I added new E2E tests that cover the bug.

### Types of changes

### Related issue(s):
1. fixes #9561

### Affected project(s):
- [x] `handsontable`

### Checklist:


jansiegel (approved):

handsontable/src/plugins/filters/filters.js
```diff
@@ -432,11 +422,22 @@ export class Filters extends BasePlugin {
   /**
    * Gets last selected column index.
    *
-   * @returns {object|null} Return `null` when column isn't selected otherwise
-   * object containing information about selected column with keys `visualIndex` and `physicalIndex`.
+   * @returns {{visualIndex: number | null, physicalIndex: number | null}} Return `null` when column isn't selected
```
warpech, line 425: @budnix Can you please check, the above line is serialized to Markdown as: 

```
**Returns**: `Object` - Return `null` when column isn't selected
otherwise object containing information about selected column with keys `visualIndex` and `physicalIndex`.  
```

I mean that there is just `Object` instead of the complex type that you've entered.

budnix: It seems that our JSDoc compiler is not able to generate such a precise type so it widening to `Object` type.

warpech: Correct, but fully informative for someone who reads the API reference https://handsontable.com/docs/api/filters/.

We should at least fix the description text of the return value, which still says:

> Return `null` when column isn't selected otherwise object containing information about selected column with keys `visualIndex` and `physicalIndex`

Could you please propose a new description text for the return value and I will put it to https://github.com/handsontable/handsontable/pull/9913

budnix: Maybe something like [this](https://github.com/handsontable/handsontable/commit/bbf09f92258a9ca69e935cc4426625ae89cfadd9#diff-01aa731e6c00facb0cd36cd88d50dcd8a9e90437847f80fd5db3bddb1880ccd6R425)?

warpech: Looking good, thanks! I cherry-picked it to PR #9913

Issue #9753, opened 2022-08-08
[12.0.0+] `Clear` filter option used in `nestedHeaders` blocks dropdown menu from displaying


AMBudnik:
### Description

When `Clear` filter option is used in `nestedHeaders` it blocks dropdown menu from displaying. The issue does not occur when we use `colHeaders` only. It also works well using Handsontable v11.1.0.

### Steps to reproduce

1. Use `clear` button in the `filter by value` list for any column
![Screenshot 2022-08-08 at 08 54 40](https://user-images.githubusercontent.com/10757813/183357273-4cb26bb7-0eb1-46cc-af6c-76ef07ef60ae.png)

3. Try to open a dropdown menu for any column.

**Result**
You are not able to open a dropdown menu for any of the columns. There are no console errors.

### Demo

https://jsfiddle.net/6pewsjxu/ - 11.1.0 (last working)
https://jsfiddle.net/6pewsjxu/1/ - 12.0.0. (first broken)

### Your environment
* Handsontable version: 12+
* Browser name and version: Chrome 103
* Operating system: macOS Monterey

Inform [Zen](https://handsontable.zendesk.com/agent/tickets/29892)

warpech:
On the today's meeting we agreed to move it to 12.1.4

aninde:
This issue will be fixed by v12.2 https://jsfiddle.net/aninde/25f86p3d/.

<img width="337" alt="Screenshot 2022-10-20 at 16 39 34" src="https://user-images.githubusercontent.com/30186818/196979831-58f2283e-cfa7-4399-b0df-4c3405cc18cf.png">

AMBudnik:
Thank you. The client, who reported the issue, is updated with the release notes.

I'm closing this issue as solved in Handsontable v12.2.0

PR #9846, opened 2022-08-30
Fix an issue with dropdownMenu not being displayed on a nestedHeaders-enabled instance with all rows trimmed out.

jansiegel:
### Context
The bug described in #9753 was caused by a more broad issue:

When the Handsontable instance does a full redraw from a callback run on `afterOnCellMouseDown`, all the callbacks for the succeeding `click` event bound to the `rootElement` are blocked by the fact that the DOM content of `rootElement`'s children was modified.
- In the simplest example: https://jsfiddle.net/js_ziggle/ecqg5w9b/
- In Handsontable context, this can be achieved without `nestedHeaders`, `filters`, and `dropdownMenu`: https://jsfiddle.net/js_ziggle/1mxgrtco/
  By passing a custom header-rendering function, we're forcing the full redraw of those. Clicking on them won't trigger the `click` event callback defined below.

My fix for #9753 pushes the render-triggering function (selecting columns) after all the previously-declared `click` callbacks were already run by placing it at the end of the `click` event stack and asynchronously removing the newly-added callback.

Additionally, I corrected some typos in the code I found while searching for this fix.

### How has this been tested?
Added a test case for the issue.

### Types of changes

### Related issue(s):
1. #9753 

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:


handsontable/src/plugins/nestedHeaders/nestedHeaders.js
```diff
@@ -453,8 +461,22 @@ export class NestedHeaders extends BasePlugin {
...
     }
 
-    // The plugin takes control of the how the columns are selected.
-    selection.selectColumns(...columnsToSelect);
+    // If there are no rendered rows, selecting columns causes a full headers redraw, blocking all the
+    // `click` events callbacks.
+    // In that case -> wait with the selecting action until all the other callbacks had been called by adding it at
+    // the top of the `click` event stack.
+    if (this.hot.countRenderedRows() === 0) {
+      const removeTempClickCallback = this.#eventManager.addEventListener(this.hot.rootElement, 'click', () => {
```
budnix, line 469: The solution appears to be a temporary patch. I've investigated the issue, and it's a regression that I introduced within https://github.com/handsontable/handsontable/commit/08895d41f91294e5de8521b05423c951a0f8cf8d. After a couple of hours, I think I found the root cause of the problem and a more elegant solution - but it still has to be thoroughly tested.

Right now, when the table has no renderable cells at all, the table is always rendered in slow render mode. The headers are re-rendered each time the "mousedown" event occurs. At the same time, the "click" event that is registered by the _DropdownMenu_ plugin never had a chance to finish as the target element is destroyed. I specified the condition https://github.com/handsontable/handsontable/commit/8dd9b2b4783431c3c5b68e4c487a8adcd89959eb.

It would still be necessary to add tests to Walkontable - If you agree with the solution.

budnix (changes requested):

handsontable/src/plugins/nestedHeaders/nestedHeaders.js
jansiegel: I found that commit as well - I tried to reverse that behavior, so the table would render in the "fast" mode when there are no rows present, but as far as I remember, many test cases failed then, so I went with a more hacky approach. If you think that's the way to go, though, I will recheck it 👍

jansiegel:
@budnix In https://github.com/handsontable/handsontable/pull/9846/commits/c7bb49f24378def1ea364f750632df445f679bc0 I took your idea from https://github.com/handsontable/handsontable/commit/08895d41f91294e5de8521b05423c951a0f8cf8d, but modified it to utilize both calculators for fully visible rows/columns and partial ones. It looks like it works fine, but I'm not 100% sure if it doesn't break your initial idea that introduced this issue. Could you verify if it makes sense?

warpech:
Does it also fix #9930?

budnix:
> @budnix In [c7bb49f](https://github.com/handsontable/handsontable/commit/c7bb49f24378def1ea364f750632df445f679bc0) I took your idea from [08895d4](https://github.com/handsontable/handsontable/commit/08895d41f91294e5de8521b05423c951a0f8cf8d), but modified it to utilize both calculators for fully visible rows/columns and partial ones. It looks like it works fine, but I'm not 100% sure if it doesn't break your initial idea that introduced this issue. Could you verify if it makes sense?

Initially, the rendering clue seemed like what needed to be fixed. But after playing with that, I saw some issues. The first problem is performance. On scroll, there a two more calculators created to check the `count` value - that hurts a little bit, and the second for oversized rows, the issue still exists - the dropdown menu can't be opened.

But it seems that a few days between code reviews refreshed my mind 😄 I reinvested the issue and found that the column (and row) non-nested headers had the same issue. They are fixed 7 years ago by adding DOM caching layer - that was a part of the performance task (https://github.com/handsontable/handsontable/commit/fbfcbb44525461ac83b68c3f27ebbfd17522911b). The _NestedHeaders_ plugin lacks that cache layer implementation that's why I propose to create a new PR with changes in the plugin. I've copied all the relevant changes to the new [PR](https://github.com/handsontable/handsontable/pull/9946).

Issue #9772, opened 2022-08-11
The `getDataAtCol()` call for 125k rows of data causes the `Maximum call stack size exceeded` error.


AMBudnik:
### Case 1 - calling `getDataAtCol()`
1. Go to http://jsfiddle.net/exmoh5tw/
2. Open the console

**result**
You get the following error in the console and the data is not returned

```js
Uncaught RangeError: Maximum call stack size exceeded
    at Core.getDataAtCol (handsontable.full.js:85871:33)
    at ?editor_console=true:123:19
```

### Case 2 - using filters
1. Go to http://jsfiddle.net/exmoh5tw/1/
2. Open dropdown for any of the columns

**result**
You get the following error in the console and the data is not returned

```js
Uncaught RangeError: Maximum call stack size exceeded
    at Core.getDataAtCol (handsontable.full.js:85871:33)
    at ?editor_console=true:123:19
```

and the menu does not show (it shows if you set `height` then it is displayed below the table)

### Your environment
* Handsontable version: 12.1
* Browser name and version: Chrome 103
* Operating system: macOS Monterey

Inform [Zen](https://handsontable.zendesk.com/agent/tickets/29930)

AMBudnik:
Checks done after fixing related issue https://github.com/handsontable/handsontable/issues/7840
Case 1 checked with 12.1.3 http://jsfiddle.net/zxhspjnr/ - issue still replicable
Case 2 checked with 12.1.3http://jsfiddle.net/2pgcvq49/1/ - issue still replicable

aninde:
I made sure the 12.3.2 fix works with array of array and array of objects on supported browsers: Chrome 111.0.5563.64, Safari 16.3 (18614.4.6.1.6)*, Firefox 111.0, Edge 111.0.1661.41 on *macOS 13.2.1 (22D68) and Windows 10.
Demos
Case 1 http://jsfiddle.net/aninde/v0m91kua/
Case 2 array of objects http://jsfiddle.net/aninde/xc0omv53/

Issue #7840, opened 2021-04-22
Adding a large number of rows breaks the table with a maximum stack exceeded error.


jansiegel:
### Description
When adding a large number of rows (exactly 62522+ rows) to an instance, it crashes with the `Maximum call stack size exceeded` error.

Seems to happen on `8.0.0`+.

### Steps to reproduce

1. Go to https://jsfiddle.net/ef7chuk9/.
2. Check the console.

### Demo
https://jsfiddle.net/ef7chuk9/

### Your environment
* Handsontable version: 8.3.2
* Browser Name and version: Latest chrome
* Operating System: MacOS Catalina

wojciechczerniak:
This method https://github.com/handsontable/handsontable/blob/fa685dc8c194c0269b893d202580ce7091ed8322/src/dataMap.js#L582-L588
Is transpiled to `splice.apply`
```ts
    key: "spliceData",
    value: function spliceData(index, amount) {
      for (var _len3 = arguments.length, elements = new Array(_len3 > 2 ? _len3 - 2 : 0), _key3 = 2; _key3 < _len3; _key3++) {
        elements[_key3 - 2] = arguments[_key3];
      }

      var continueSplicing = this.instance.runHooks('beforeDataSplice', index, amount, elements);

      if (continueSplicing !== false) {
        var _this$dataSource;

        (_this$dataSource = this.dataSource).splice.apply(_this$dataSource, [index, amount].concat(elements));
      }
    }
```

> But beware: by using apply this way, you run the risk of exceeding the JavaScript engine's argument length limit. The consequences of applying a function with too many arguments (that is, more than tens of thousands of arguments) varies across engines. (The JavaScriptCore engine has hard-coded argument limit of 65536.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply#using_apply_and_built-in_functions

adrianszymanski89:
This issue is still replicable in v. 12.0.1:

https://jsfiddle.net/aszymanski/su87k1v3/

semlak:
I'm having similar problem that I believe is same root issue. With a table with a large number of rows that is otherwise displaying okay, with column filters enabled, you hit the drop down for the column to select a filter, and the menu can't open because it hits a maximum callstack size. I think it is from when computing the values that it can filter on.
The method `getDataAtCol`. 
```
 this.getDataAtCol = function (column) {
    var _ref13;

[...]

PR #9855, opened 2022-09-01
Adding a large number of rows won't break the table

wszymanski:
### Context

The `alter` method used for inserting rows uses DataMap's `spliceData` method. As described [in the issue](https://github.com/handsontable/handsontable/issues/7840#issuecomment-825070300) method's body is transpiled by Babel. The `apply` method used after that has its limitations.

The easiest and probably the purest way of changing from transplied array's `splice` method is using another method. However, we need a method which **modify the original array**. Another solution will need some extra code for updating the reference. That's why `filterData` already execute `this.dataSource.length = 0` and update the array. 

https://github.com/handsontable/handsontable/blob/4c6d81235c51ca35251cf87f0f852a923010df90/handsontable/src/dataMap/dataMap.js#L624-L635

Changes has been already implemented by using `Array`'s `push` method in loop. Please keep in mind that alternative solution such as using spread operator for filling data's array is impossible for the same reasons as using the `splice` method. It's worth to mention also that [there is some issue](https://github.com/handsontable/handsontable/issues/9772) which shows another cases where source of a problem may be the same. This PR doesn't expand the scope and problems haven't been fixed.

I measured the performance for past and present solution and it seems that there is no significant difference in performance (the measured deviation was equal to `0,7%`). It's worth to mention also that there is no such problem with analogous `insert_col` option - it doesn't use `spliceData` method internally.

### How has this been tested?

Manually tested example from the issue. Measured the performance using loop and `console.time`.

### Types of changes

### Related issue(s):
1. #7840 

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:


handsontable/src/dataMap/dataMap.js
```diff
@@ -596,14 +596,21 @@ class DataMap {
...
     if (continueSplicing !== false) {
-      this.dataSource.splice(index, amount, ...elements);
+      const newData = [...this.dataSource.slice(0, index), ...elements, ...this.dataSource.slice(index)];
+
+      // We try not to change the reference.
+      this.dataSource.length = 0;
+
+      // Pushing to array instead of using `splice`, because Babel changes the code to one that uses the `apply` method.
+      // The used method was cause of the problem described within #7840.
+      newData.forEach(row => this.dataSource.push(row));
```
jansiegel, line 613: @wszymanski Have you checked if it does affect the performance in any way?

wszymanski: As I wrote in PR description:

> I measured the performance for past and present solution and it seems that there is no significant difference in performance (the measured deviation was equal to 0,7%). It's worth to mention also that there is no such problem with analogous insert_col option - it doesn't use spliceData method internally.

🙂

jansiegel: Oh, my bad, I missed that part 🙇

jansiegel (approved):

aninde:
This fix works wonderfully on Chrome 104, FF 104.0.2, and Edge, but not on Safari. It is not related to the version, because on the older Safari 12.1 / Mojave Safari you can see it through the BrowserStack on demo https://jsfiddle.net/jkq38xdL/.
e2e doesn't pass there, due to the error in the browser console. I managed to cause an error locally after adding 65535 rows.

Screenshot 2022-09-14 at 16 10 11

Tested on Safari 15.6 (17613.3.9.1.5) / macOS 12.5, Safari 12.1 / Mojave with 35a0b036e9f6bdd1552d8d7fd69cf7cc4e35720b from freeze branch.

wszymanski:
Good job, @aninde. I confirm the problem. I'm testing already implemented solution. I'll make another PR fixing the problem.

Issue #9856, opened 2022-09-01
Docs: React-docs: Issues with dark mode inside Redux & Vuex examples


aninde:
### Description

The demo on Redux page has the wrong style inside on the seperate PR https://github.com/handsontable/handsontable/pull/9787. Some elements inherint wrong style when dark mode is on.
```
 html.theme-dark table:not(.selected-preview *):not(.pika-table):not(.htCore) td, html.theme-dark table:not(.selected-preview *):not(.pika-table):not(.htCore) th {
    border-bottom: 1px solid #20232a;
```
<img width="324" alt="Screenshot 2022-09-01 at 14 39 26" src="https://user-images.githubusercontent.com/30186818/187916056-5cbfd357-52b3-4f25-960c-0ba592a9d409.png">
<img width="395" alt="Screenshot 2022-09-01 at 14 39 36" src="https://user-images.githubusercontent.com/30186818/187916063-6282a575-49a8-416a-b29d-883fd0b47134.png">

Searchbar looks well, so it's not related to https://github.com/handsontable/handsontable/issues/9781

### Links

Set dark mode and then open https://dev.handsontable.com/docs/next/react-data-grid/redux/#overview
or open page, set dark mode, refresh page

https://dev.handsontable.com/docs/next/javascript-data-grid/vue3-vuex-example/#example-vuex-store-dump
https://dev.handsontable.com/docs/next/javascript-data-grid/vue-vuex-example/

### Environment
Browser: Safari 15.6 (17613.3.9.1.5), Chrome 105.0.5195.52 
OS: macOS 12.5 (21G72)

**Version**
Only on https://github.com/handsontable/handsontable/pull/9787 
not on https://github.com/handsontable/handsontable/tree/feature/react-docs
staging a9586d94025ba754c387425196170bb5f7089a2e

warpech:
Also seen in the current `feature/react-docs` branch at 140fb0cab5d616152c428b042c9f5364fd3a125c. We think it is a new problem that did not exist a few days ago.

warpech:
This problem only exists when starting on the page https://dev.handsontable.com/docs/next/react-data-grid/redux/ directly. When you start on another page, and then navigate to the Redux page, it looks OK.

My guess is that this problem started to appear after we made the "Preview" tab open by default (PR #9440)

warpech:
Closing as done

PR #9868, opened 2022-09-05
Fixed problem with not adding `selected-preview` class on page start

wszymanski:
### Context

I discovered that documentation build using `npm run docs:start` script looks properly, while those build using `npm run docs:build` script not. I can't find the source of the problem. The `MutationObserver` has shown that `class` attribute doesn't change on first `codePreviewTabChanged` method execution (on start). There should be added `selected-preview` to class list. It's worth to mention that every click on the [`tabs` element](https://github.com/spatie/vue-tabs-component) already call the method properly. It seems that extra timeout fixes the problem.

edit: Thanks to @budnix changes the code changing DOM by adding/removing classes has been refactored and suggested workaround removed.

### How has this been tested?

I tested http://localhost:8080/docs/next/react-data-grid/redux/#example page multiple times. I was checking whether DOM element with `tabs-component-panels` class is changed on the start.

[skip changelog]

### Related issue(s):
1. #9856

### DIFF
--- a/docs/.vuepress/theme/components/Page.vue
+++ b/docs/.vuepress/theme/components/Page.vue
@@ -32,13 +32,6 @@ export default {
       activatedExamples: [],
     };
   },
-  mounted() {
-    // console.log(document.querySelector('.tabs-component-panels'));
-    // setTimeout(() => {
-    //   document.querySelector('.tabs-component-panels').classList.add('selected-preview');
-    // }, 0);
-    // console.log(document.querySelector('.tabs-component-panels').classList);
-  },
   computed: {
     isApi() {
       return this.$route.fullPath.match(/([^/]*\/)?api\//);
@@ -55,6 +48,7 @@ export default {
       if (selectedTab.tab.computedId.startsWith('preview-tab')) {
         this.activatedExamples.push(exampleId);
 
+        // Extra timeout as there is some problem related to not adding class properly (#9856).
         setTimeout(() => {
           parentContainer.className += ` ${changedClass}`; // Adding class by method compatible with IE
         }, 0);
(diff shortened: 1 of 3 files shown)


docs/.vuepress/theme/components/Page.vue
```diff
@@ -47,7 +47,11 @@ export default {
 
       if (selectedTab.tab.computedId.startsWith('preview-tab')) {
         this.activatedExamples.push(exampleId);
-        parentContainer.className += ` ${changedClass}`; // Adding class by method compatible with IE
+
+        // Extra timeout as there is some problem related to not adding class properly (#9856).
+        setTimeout(() => {
```
budnix, line 52: As you already notice, that line of code is disturbing 😄 That's why I would propose to remove the code that directly mutates the DOM to code that is managed by the Vue itself. I've prepared working PoC https://github.com/handsontable/handsontable/commit/070117d7117dbef3081468e33556a06ebc3f9114. WDYT?

budnix (changes requested):

docs/.vuepress/theme/components/Page.vue
wszymanski: > As you already notice, that line of code is disturbing 😄 That's why I would propose to remove the code that directly mutates the DOM to code that is managed by the Vue itself. I've prepared working PoC [070117d](https://github.com/handsontable/handsontable/commit/070117d7117dbef3081468e33556a06ebc3f9114). WDYT?

Thanks for your changes. It looks very good not and the workaround has been removed. I can see that class is added on `div` containing the main component which looks better. I take your code, it fulfills the assumptions much better!

budnix (approved):
Looks good 👍

PR #9869, opened 2022-09-06
React docs: Modify the example checker to wait for all the network calls to be finished.

jansiegel:
### Context
Seems like @budnix's suggestion that the example checker might be failing because of the cdn files not being loaded on time was (at least partially) correct.

This PR:
- Utilizes the Puppeteer's `waitUntil` option to wait for all the network calls to be done before running the test script
- Extends the re-check try count to 15 to be double safe

### How has this been tested?
Triggered the CI 13 times in a row to see if the checker is not failing.

[skip changelog]

### DIFF
--- a/.github/workflows/docs-staging.yml
+++ b/.github/workflows/docs-staging.yml
@@ -42,6 +42,35 @@ jobs:
       - name: Test examples
         run: |
           npm run docs:test:example-checker
+          npm run docs:test:example-checker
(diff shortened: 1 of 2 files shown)


budnix (approved):
I've deployed the branch 10 times, plus I've seen that you pushed at least 10 times more ... all with success. So, I guess the PR is ready to merge. Thanks 👍

Issue #6359, opened 2019-10-18
Formula is converted to a value if merged areas are included


AMBudnik:
### Description

Extending the title description:
If the result of the formula has a merged area of the left side and at the same time on the top, it is converted to a value. 
![image](https://user-images.githubusercontent.com/10757813/67076432-6c3b0200-f18d-11e9-8848-3c100d77c241.png)

It does not happen if there is only one merged are on left OR on top.
![Zrzut ekranu 2019-10-18 o 09 59 56](https://user-images.githubusercontent.com/10757813/67076756-19157f00-f18e-11e9-8558-e414ff2975e2.png)
![Zrzut ekranu 2019-10-18 o 09 59 35](https://user-images.githubusercontent.com/10757813/67076757-19ae1580-f18e-11e9-8384-4edc4fd9ad9e.png)

In cases where the merged areas are on bottom OR on the left only one area breaks the formula.

![Zrzut ekranu 2019-10-18 o 10 06 12](https://user-images.githubusercontent.com/10757813/67077450-97265580-f18f-11e9-89f5-34f7143b8689.png)
![Zrzut ekranu 2019-10-18 o 10 07 31](https://user-images.githubusercontent.com/10757813/67077500-ac9b7f80-f18f-11e9-873c-484e5aafda8d.png)

### Steps to reproduce

1. Open editor for cell with value `6`.

**Expected**
You should see the `=SUM(B2*2)` in the editor

**Result**
You see the value instead.

### Demo

https://jsfiddle.net/kx90uqvL/2/

### Your environment
* Handsontable version: 7.2.1
_Doesn't seem to be related to any version_
* Browser Name and version: Chrome 77
* Operating System: Windows 10

Inform [forum](https://forum.handsontable.com/t/gh-6359-why-formulas-dont-work/3795), [forum](https://forum.handsontable.com/t/trouble-with-formulas-and-merged-cells/4267) and [forum](https://forum.handsontable.com/t/gh-6359-when-a-array-is-set-to-mergecells-property-functions-are-replaced-by-its-values/5434/2), Zen 26970, [Zen](https://handsontable.zendesk.com/agent/tickets/29590)

gnsilence:
When can you repair this problem, I think someone has put forward the solution

AMBudnik:
Hi @gnsilence 

As presented here

[...]

PR #9884, opened 2022-09-09
Fixed problem with overwriting source data incorrectly when there are some merged cells

wszymanski:
### Context

I've found that there is done some date overwriting by the below code.

https://github.com/handsontable/handsontable/blob/4f6c253e6257514b168a4c98fe35c9c233b1487b/handsontable/src/plugins/mergeCells/mergeCells.js#L248-L250

This is an intended action. The `MergeCells` plugin fills up some cells with `null` values. However, it has been done incorrectly. The data population has been done using already modified data by `Formulas` plugin. It shouldn't be done in that way. We should done data population respecting cell sequence (managed by the `IndexMapper`), but without formula calculation.

~This PR also adds a feature. There will be an extra possibility to cancel `modifyData` hook propagation be returning a `false` value. After that, next callback for hook won't be executed.~

### How has this been tested?

I tested dataset with columns moved by the `ManualColumnMove` on start.

### Types of changes

### Related issue(s):
1. #6359

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:


wszymanski:
If this PR is not a minor version it should not be carried over to milestone 12.1.4.

wszymanski:
I can see that there is one not passing tests 🤯. I'll check it yet.

wszymanski:
@budnix It's ready for CR.

budnix (changes requested):
I think that the PR fixes the issue only superficially. The root of the problem is elsewhere. I investigated that and I believe the issue is here, [in that one line](https://github.com/handsontable/handsontable/blob/feature/issue-6359/handsontable/src/plugins/mergeCells/mergeCells.js#L427). Changing to `clearedValue = this.hot.getSourceDataAtCell(this.hot.toPhysicalRow(mergeParent.row), this.hot.toPhysicalColumn(mergeParent.col));` solves the issue.

But there is another problem with a buggy code that updates the dataset beyond the merged cells. The [`getBulkCollectionDataRange`](https://github.com/handsontable/handsontable/blob/feature/issue-6359/handsontable/src/plugins/mergeCells/mergeCells.js#L293) method returns wrong ranges to dataset update. For example, for that setup:
image
the dataset from A1 to D5 is updated - it's wrong. Dataset update should go for cells D1-D4 and B5-C5. 

If you investigate a problem, you will likely find that the [`this.hot.getData()`](https://github.com/handsontable/handsontable/blob/feature/issue-6359/handsontable/src/plugins/mergeCells/mergeCells.js#L269) method call is not necessary as well the new feature that you've added to the `modifyData` hook. All necessary data already returns the `this.mergeRange` method.

wszymanski:
> 

Thanks, I'll check this approach.

wszymanski:
> I think that the PR fixes the issue only superficially. The root of the problem is elsewhere. I investigated that and I believe the issue is here, [in that one line](https://github.com/handsontable/handsontable/blob/feature/issue-6359/handsontable/src/plugins/mergeCells/mergeCells.js?rgh-link-date=2022-09-21T07%3A53%3A07Z#L427). Changing to `clearedValue = this.hot.getSourceDataAtCell(this.hot.toPhysicalRow(mergeParent.row), this.hot.toPhysicalColumn(mergeParent.col));` solves the issue.

You are right. That's the most important part of the fix.

> 
> But there is another problem with a buggy code that updates the dataset beyond the merged cells. The [`getBulkCollectionDataRange`](https://github.com/handsontable/handsontable/blob/feature/issue-6359/handsontable/src/plugins/mergeCells/mergeCells.js?rgh-link-date=2022-09-21T07%3A53%3A07Z#L293) method returns wrong ranges to dataset update. For example, for the setup: image the dataset from A1 to D5 is updated - which is wrong. The dataset update should go only for merged cells D1-D4 and B5-C5.
> 

I investigated that and I can't fully agree with that. This approach isn't so bad as you said. We can realise data population in two ways:

1. As it works now (but with removed bug) that there is just one data population (for bigger range, containing every merged area). It have some advantages - just one action is done and as a consequence hooks are performed just once.
2. As multiple data populations. Maybe the code would be easier, ranges smaller, but we won't have advantages of the first approach.

I would leave it as it is now. On the margin of your choice, there has been buggy code which extended range of population (starting sometimes from `0, 0` coordinates) and doing unnecessary data population, for example while unmerging a cell area.

> If you reinvest the problem with my tips, you will likely find that the [`this.hot.getData()`](https://github.com/handsontable/handsontable/blob/feature/issue-6359/handsontable/src/plugins/mergeCells/mergeCells.js?rgh-link-date=2022-09-21T07%3A53%3A07Z#L269) method call is unnecessary (can be removed). 

In first approach it's okey to use `DataMao`. The `populateFromArray` uses visual indexes for data population. After filling up the get data with extra nulls and source data (for cell representing the merge area value) everything works properly.

> Hence the new feature that you added to the `modifyData` hook will be redundant. I think that all necessary data needed to fix the issue are already returned by the `this.mergeRange` method.

I can agree with that, the `modifyData` is redundant. According to YAGNI principle I'm removing it.

budnix (changes requested):
There is still a problem. After your changes, the dataset that does not belong to the merged cells is still updated. 

According to the screenshot:
Zrzut ekranu 2022-09-26 o 10 01 05
The red border is what I would expect, and the green border is how it works now. If you put a formula in the dataset in B3, the bug will occur (yep, there is a missing test for that case). That's why I mentioned earlier about rewriting the `getBulkCollectionData` method. 

I can sell you a tip about the ability to update the dataset for non-contiguous cells in one call. The `setDataAtCell` method does that, for example, `hot.setDataAtCell([ [0, 0, 'A'], [1, 1, 'B'] ]);`. I'm sure the range should be fixed to update only the dataset under the merged cells using the mentioned method.

wszymanski:
> There is still a problem. After your changes, the dataset that does not belong to the merged cells is still updated.
> 
> According to the screenshot: Zrzut ekranu 2022-09-26 o 10 01 05 The red border is what I would expect, and the green border is how it works now. If you put a formula in the dataset in B3, the bug will occur (yep, there is a missing test for that case). That's why I mentioned earlier about rewriting the `getBulkCollectionData` method.
> 
> I can sell you a tip about the ability to update the dataset for non-contiguous cells in one call. The `setDataAtCell` method does that, for example, `hot.setDataAtCell([ [0, 0, 'A'], [1, 1, 'B'] ]);`. I'm sure the range should be fixed to update only the dataset under the merged cells using the mentioned method.

You are sure, It wasn't possible without rewriting the methods. Fortunelly, my changes resulted in a reduction in the amount of code.

PR #9913, opened 2022-09-15
Docs: Adding 12.1.3 docs 

warpech:
This PR:

- Edits the 12.1.3 changelog entries
- Adds missing 12.1.3 changelog entries
- Fixes some minor problems in documentation that I found by comparing the current content with the last published in 12.1.2. Details are provided as commit messages.
- Updates dependency and Node version numbers throughout the docs
- I verified the third-party licenses, they don't need changing

[skip changelog]


CHANGELOG.md
```diff
@@ -8,21 +8,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Fixed an issue with "0" values being ignored by the Column Summary plugin. [#6385](https://github.com/handsontable/handsontable/issues/6385)
+- Fixed an error when opening the column dropdown menu with the Filters plugin enabled. [#9561](https://github.com/handsontable/handsontable/issues/9561)
+- Fixed an issue where the `trimWhitespace` option could not be used in the column and cell levels of the cascading configuration. [#7387](https://github.com/handsontable/handsontable/issues/7387)
+- Fixed the flickering of the selection area at the edge of the table while scrolling. [#8317](https://github.com/handsontable/handsontable/issues/8317)
+- Fixed misalignment on mobile devices when the edge cells were selected. [#9621](https://github.com/handsontable/handsontable/issues/9621)
+- Fixed type definitions for the method `setSourceDataAtCell`. [#8599](https://github.com/handsontable/handsontable/issues/8599)
+
+### Removed
+- Removed a custom stable sorting implementation in favor of the built-in sorting algorithm [#6091](https://github.com/handsontable/handsontable/issues/6091)
```
budnix, line 30:
```suggestion
- Removed a custom stable sorting implementation in favor of the built-in in the browser stable sorting algorithm [#6091](https://github.com/handsontable/handsontable/issues/6091)
```
WDYT emphasizes that even though we remove something, the column sorting will still work as stable as before the changes.

docs/content/guides/cell-types/numeric-cell-type.md
```diff
@@ -11,10 +11,14 @@ canonicalUrl: /numeric-cell-type
 
 ## Overview
 
-By default, Handsontable treats all cell values as `string` type. This is because the `<textarea>` returns a string as its value. There are many cases where you need cell values to be treated as a `number` type. The numeric cell type allows you to format displayed numbers nicely and sort them correctly.
+By default, Handsontable treats all cell values as `string` type. This is because the text editor returns a string as its value. There are many cases where you need cell values to be treated as a `number` type. The numeric cell type allows you to format displayed numbers nicely and sort them correctly.
```
wszymanski, line 14: I would not remove this technical detail.

docs/content/guides/cell-types/numeric-cell-type.md
```diff
@@ -11,10 +11,14 @@ canonicalUrl: /numeric-cell-type
...
 
 ## Usage
 
+::: tip
+Ensure your numeric cell values are stored numbers and not strings in the data source, as Handsontable doesn't parse strings to numbers.
```
wszymanski, line 19: ♥

wszymanski (commented):
Looks okey, just one minor suggestion. Let me know what you thinks about mine and @budnix 's suggestions.

docs/content/guides/cell-types/numeric-cell-type.md
warpech: Frankly, this text isn't new. I just reverted the text from before https://github.com/handsontable/handsontable/pull/9701, because I couldn't not agree with it.

docs/content/guides/cell-types/numeric-cell-type.md
warpech: Why not? My reasoning is: 

1. The old text uses the HTML element `<textarea>` without context, which is not very informative. It was never introduced what `<textarea>` has to do with the numeric cell type.
2. `<textarea>`is used by the text cell editor. The numeric cell editor only uses it through inheritance (like almost all other editors). If it is so implementation detail is so important, why should it be given just on this particular page? 

I agree that this change is not really in scope of this PR, so if the above explanation does not convince you, I will revert it.

wszymanski: 1. I can agree with that.
2. That's how Handsontable works for now. I think that new text explain how Handsontable works, but without implementation details it may sound strange - like our whim to do some kind of casting. That's not true, the change comes from the idea of reading from DOM element. It's documentation for developers, so I would leave it.

Maybe this description isn't perfect, but for me it have more information. I would leave it as is.

CHANGELOG.md
budnix: Sounds better ... I'd tweak that by changing the word `provided` to `guaranteed` and remove `standard library`. I guess it's guaranteed by the spec. The `library` in that context sounds odd 🤔 

Suggestion: `Removed a custom stable sorting implementation in favor of the stable sorting algorithm guaranteed by the ECMAScript 2019`.

docs/content/guides/cell-types/numeric-cell-type.md
warpech: The more I think about it, the more I find the old the explanation not satisfying.

> By default, Handsontable treats all cell values as `string` type. This is because the `<textarea>` returns a string as its value.

That was an explanation in 2012, when much of Handsontable's behavior relied directly on DOM. But currently, the behavior is dictated by conscious choices, not by underlying DOM elements.

May I propose a compromise: 1fd2ad6b36d0ed8943f29e179bac1907a5cfa857. WDYT?

wszymanski (approved):

CHANGELOG.md
warpech: Thanks, done in b352f802d51006e5629d060f97a183726d13c4cd

PR #9914, opened 2022-09-15
Fixed problems with Safari - there were thrown an error for transpilled destructing operator

wszymanski:
### Context

PR #9855 introduced changes in a way how row are inserted. The fix has been working for Chrome browser. However, it seems that problem resolved in the PR hasn't been resolved in another part of code. For the unknown reason it has worked for other browsers.

[skip changelog]

### How has this been tested?

I measured the performance for past and present solution on Safari and it seems that there is no significant difference in performance (the measured deviation was equal to `3,7%`). The same for Firefox (deviation is equal to `3,5%`).

### Types of changes

### Related issue(s):
1. #7840

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

### DIFF
--- a/handsontable/src/dataMap/dataMap.js
+++ b/handsontable/src/dataMap/dataMap.js
@@ -597,7 +597,7 @@ class DataMap {
    *
    * @param {number} index Physical index of the element to add/remove.
    * @param {number} deleteCount Number of rows to remove.
-   * @param {...object} elements Row elements to be added.
+   * @param {Array<object>} elements Row elements to be added.
    */
   spliceData(index, deleteCount, elements) {
     const continueSplicing = this.instance.runHooks('beforeDataSplice', index, deleteCount, elements);
--- a/handsontable/src/dataMap/dataMap.js
+++ b/handsontable/src/dataMap/dataMap.js
@@ -341,7 +341,7 @@ class DataMap {
 
     this.instance.rowIndexMapper.insertIndexes(rowIndex, numberOfCreatedRows);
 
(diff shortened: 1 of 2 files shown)


handsontable/src/dataMap/dataMap.js
```diff
@@ -599,7 +599,7 @@ class DataMap {
    * @param {number} deleteCount Number of rows to remove.
    * @param {...object} elements Row elements to be added.
```
budnix, line 600: I guess you forgot to update the JSDoc after changing the type of the argument

budnix (changes requested):
The fix seems to work, but it changes [the signature](https://github.com/handsontable/handsontable/pull/9914/files#diff-90a9b6480d675d507f6acea9db2794d1b866147d94413108203f699ad20ca474R603) (the 3rd argument) of the `beforeDataSplice` hook. Is that OK?

handsontable/src/dataMap/dataMap.js
wszymanski: Thank you, fixed within f960d969a131e2fbf5d20190b089dd08eefa0d9a.

budnix (changes requested):
Still .. what about the braking change in the `beforeDataSplice` hook? https://github.com/handsontable/handsontable/pull/9914#pullrequestreview-1110239042

budnix (approved):

Issue #9581, opened 2022-06-14
Progress and Rating columns should be editable in the main demo


krzysztofspilka:
### Description

I would expect that cells in columns "Progress" and "Rating" are editable.

- Progress: cell editor - numeric in the range 0-100. If beyond the range, set the edge value (minimum or maximum) and close the editor.
- Rating: cell editor - numeric in the range 0-5. If beyond the range, set the edge value (minimum or maximum) and close the editor.

<img width="228" alt="Screenshot 2022-06-14 at 15 49 23" src="https://user-images.githubusercontent.com/8048526/173593724-2ac8c520-46df-4407-afd3-0e9f74c62e6c.png">

### Links

https://handsontable.com/docs/demo/

warpech:
As part of this issue:
1. the data source should be changed from an array of arrays to an array of objects
2. the source code should be improved to follow the current best practices for a given framework (if now something strikes as odd)

warpech:
This should be done as 1 framework = 1 pull request, starting from React.

krzysztofspilka:
[PR](https://github.com/handsontable/handsontable/pull/9935) to be finished.

AMBudnik:
This issue is no longer applicable.

PR #9935, opened 2022-09-21
Adds editing options for progress and rating columns in react demo.

bulaj:
### Context

Adds editing options for progress and rating columns in react demo.
### How has this been tested?

Manually in browser.
### Types of changes

### Related issue(s):
1. #9581 
2.
3.

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

[skip changelog]


warpech:
@krzysztofspilka to run the project, run the following commands in the repo root:

```
git fetch
git checkout feature/issue-9581-demo
cd examples/12.1.2/docs/react/demo
npm i
npm run start
```

krzysztofspilka:
@warpech thanks for help, I'm on it.

krzysztofspilka:
@bulaj I can't tell much about the quality of the code, but from the UX/UI standpoint, there are some issues that we should fix before rolling this out. See the list below.

- An empty cell in the column "Progress" is represented by a 100% stacked bar chart. Test: click on any cell in the column "Progress" and then press <kbd>DELETE</kbd>. Alternatively, open the column menu and choose the option "Clear column".
- All keyboard shortcuts from this list: https://handsontable.com/docs/keyboard-shortcuts/#edition-keyboard-shortcuts should be available for both the "Progress" and "Rating" columns.
- I can paste non-numeric values to both the "Progress" and "Rating" columns, and apparently the renderer will try to handle them. Try pasting `12394-3841-34`. When you open the editor of such a cell it is empty.

bulaj:
@krzysztofspilka 1 and 3 should be fixed now.

examples/12.1.2/docs/react/demo/src/index.tsx
```diff
@@ -46,24 +47,25 @@ const App = () => {
       beforeRenderer={addClassesToRows}
       afterGetRowHeader={drawCheckboxInRowHeaders}
       afterOnCellMouseDown={changeCheckboxCell}
+      beforePaste={omitNonNumericPaste}
```
warpech, line 50: This is to achieve the following requirement:

> I can paste non-numeric values to both the "Progress" and "Rating" columns, and apparently the renderer will try to handle them. Try pasting `12394-3841-34`. When you open the editor of such a cell it is empty.

I think that this implementation is not as good as it could be. The best way to disallow certain inputs is to use a validator function and `allowInvalid: false` configuration option. This is how e.g. date picker does it.

warpech (changes requested):
Partial review

jansiegel (changes requested):
@maciejbo-handsontable 
I'm not sure what is the exact scope of this PR. Could you list the requirements that are supposed to be met here?
The PR description mentions only making the `progress` and `rating` columns editable, but the PR you previously used to make changes was described vaguely as `improve React demo`.

That being said, I've looked at the demo after the modifications introduced by the PR and found these:

1. This PR modifies the data used in the example. After the changes, the dataset consists of list of 5 items duplicated five times. I don't think demos in other frameworks do that. Was this change intentional?
2. The `Progress` column allows entering non-numeric data, which results in the progress bar being displayed as 100%. Works the same with pasted data.
3. The `Rating` column _doesn't_ accept non-numeric data when editing but _does_ accept pasted non-numeric data, which results in the cell being cleared.
4. The demo contains some logic which I'm assuming was meant to horizontally align the header contents, but it targets only columns `9`, `10` and `12` and there are no such columns in this example.
5. The state of the row header checkboxes is stored at `(row, "0")` coordinates, which I find a bit weird.

If any of them is outside this task's scope, let me know.

jansiegel:
@maciejbo-handsontable After today's meeting, I've looked at your ProgressBar render, and I found that:

1. You created a renderer extending the editor component. 
    https://github.com/handsontable/handsontable/blob/f79ceb29a8e758f3024121cbd7e909646241c0a0/examples/next/docs/react/demo/src/renderers/ProgressBar.tsx#L13
    It would make sense if the component would act as both the editor and the renderer (something like this -> https://handsontable.com/docs/react-data-grid/redux/#advanced-example ), but in this case it isn't needed.
2. Setting `allowInvalid=false` to the column makes the renderer not get triggered after pasting. This makes sense in the vanilla version - with the frameworks it gets more complicated and is potentially a bug. I'd leave it set as default in this demo and implement the post-validation logic in the renderer.

 I did a quick rewrite of your code and came up with something like this:
```js
export const ProgressBarRenderer = (props: HotRendererProps) => {
  if (props.cellProperties.valid !== false) {
    return (
      <div
        className="progressBar"
        style={{ width: `${getRangeValue(props.value, minAllowedValue, maxAllowedValue)}px` }}
      />
    );

  } else {
    return (
      <div
        style={{ background: '#ff4c42' }}
      >#bad-value#</div>
    );
  }
}
```

You can try implementing it and fine tuning to your needs.

examples/next/docs/react/demo/src/index.tsx
```diff
@@ -23,6 +23,10 @@ for (let x = 1; x <= 5; x++) {
   data.push(...JSON.parse(JSON.stringify(defaultData)));
 }
 
+function handleChange(event: any) {
+  console.log(event.target.value);
+}
+
```
jansiegel, line 29: Is this `console.log` needed?

jansiegel, line 29: Is this `console.log` needed?

examples/next/docs/react/demo/src/renderers/ProgressBar.tsx
```diff
@@ -1,24 +1,33 @@
...
 
 const minAllowedValue = 0;
 const maxAllowedValue = 100;
 
-
-interface IProgressBarRendererProps extends HotEditorProps {
-  value: number;
-}
-
 export class ProgressBarRenderer extends BaseEditorComponent {
```
jansiegel, line 9: [In the first point of my earlier comment](https://github.com/handsontable/handsontable/pull/9935#issuecomment-1415532583), I suggested that extending a _renderer_ class from an _editor_ class is not needed in this example.
And if there wouldn't be a need for extending classes, we should probably use function-based components here.

examples/next/docs/react/demo/src/renderers/ProgressBar.tsx
```diff
@@ -1,24 +1,33 @@
...
-  props!: IProgressBarRendererProps;
+  value: number;
+  isValid: boolean;
+  
+  constructor(props: any) {
```
jansiegel, line 13: Avoiding `any` if the information about the props is easily available would probably be a good idea.

The renderer component receives [the same set of arguments the function-based](https://handsontable.com/docs/react-data-grid/cell-renderer/#declare-a-custom-renderer-as-a-function) renderer does, but as props.

examples/next/docs/react/demo/src/renderers/ProgressBar.tsx
```diff
@@ -1,24 +1,33 @@
...
+    super(props);
+    this.isValid = props.cellProperties.valid === false ? false : true;
```
jansiegel, line 15: This could probably be simplified as 
```js
this.isValid = props.cellProperties.valid !== false;
```
?

examples/next/docs/react/demo/src/renderers/Stars.tsx
```diff
@@ -1,22 +1,32 @@
...
+      return (
+        <div className="star htCenter">
+          {"★".repeat(getRangeValue(this.value, minAllowedValue, maxAllowedValue))}
+        </div>
+      );
+    } else {
+      return (
+        <div className="error"> { MESSAGE?.BAD_VALUE } </div>
+      );
+    }
```
jansiegel, line 30: Same suggestions as with the `ProgressBar` renderer.

jansiegel (changes requested):

jansiegel (changes requested):
There are still some unanswered questions, @adamu-handsontable, could you clarify the scope of this task?

1. This PR modifies the data used in the example. After the changes, the dataset consists of 5 items duplicated five times. This changes the way the table looks, and if it was ever synchronized to the `visual-tests` demo sets (which I'm guessing it will at some point?) will break the test scenarios. [[original comment](https://github.com/handsontable/handsontable/pull/9935#pullrequestreview-1277235908)]
2. The demo contains some logic which I'm assuming was meant to align the header contents horizontally, but it targets only columns 9, 10, and 12 - and there are no such columns in this example. [[original comment](https://github.com/handsontable/handsontable/pull/9935#pullrequestreview-1277235908)]
3. The state of the row header checkboxes is stored at (row, "0") coordinates, which I find a bit weird. [[original comment](https://github.com/handsontable/handsontable/pull/9935#pullrequestreview-1277235908)]
4. When comparing the demo from this PR and the demo from `develop`, they're different not only because of the modified dataset, but also:
   - The `develop` demo has some rows pre-selected
   - The `develop` demo has some checkboxes from the column `In stock` unchecked by default

   Should they be unified?
5. We could work on the progress bar and star renderers/editors further, for example, limit the input to a specific range etc (currently, entering a value of 10000 will successfully save it to the edited cell), but, again, I don't know how much detailed are they supposed to be.

maciejbo-handsontable:
> There are still some unanswered questions, @adamu-handsontable, could you clarify the scope of this task?

Hey, what I can say is I was asked to "finish this PR which title says >editing options for prgress and rating" and with that I was trying not to touch what was not necessary.
 
>     1. This PR modifies the data used in the example. After the changes, the dataset consists of 5 items duplicated five times. This changes the way the table looks, and if it was ever synchronized to the `visual-tests` demo sets (which I'm guessing it will at some point?) will break the test scenarios. [[original comment](https://github.com/handsontable/handsontable/pull/9935#pullrequestreview-1277235908)]

With first paragraph being said - normally I wouldn't touch this thing, but I needed to implement some stuff, so I changed this entire file cause in my opinion it would be easier to change in future 5 places instead of 50 - but now, cause of your next points, I brought previous version back.

> 
>     2. The demo contains some logic which I'm assuming was meant to align the header contents horizontally, but it targets only columns 9, 10, and 12 - and there are no such columns in this example. [[original comment](https://github.com/handsontable/handsontable/pull/9935#pullrequestreview-1277235908)]

If I understand your message correctly, the same thing exists in rest of demos also. Need confirmation if I have to do something with that in this PR.
 
>     3. The state of the row header checkboxes is stored at (row, "0") coordinates, which I find a bit weird. [[original comment](https://github.com/handsontable/handsontable/pull/9935#pullrequestreview-1277235908)]

If I understand your message correctly, the same thing exists in rest of demos also. Need confirmation if I have to do something with that in this PR.

> 
>     4. When comparing the demo from this PR and the demo from `develop`, they're different not only because of the modified dataset, but also:
>        
>        * The `develop` demo has some rows pre-selected
>        * The `develop` demo has some checkboxes from the column `In stock` unchecked by default
>        
>        Should they be unified?

Hey, that's true - already unified.

> 
>     5. We could work on the progress bar and star renderers/editors further, for example, limit the input to a specific range etc (currently, entering a value of 10000 will successfully save it to the edited cell), but, again, I don't know how much detailed are they supposed to be.

this need to be confirmed by @adamu-handsontable

adamu-handsontable:
As a result of this task I'd like to have a ready-to-use example with editable Progress and Rating. I see that there are some changes which are not required to achieve it. The original PR was done some time ago and got probably outdated. Our basis is the current example.

1. Keep the array of object type structure, but make data equivalent to current example from develop. Change order of columns to match the current example from develop.
2. If there no such columns, then it seems not necessary, please check if it is not required and if so, remove it.
3. Yes this is strange, please change it to something more meaningful like: `isChecked` and adjust `drawCheckboxInRowHeaders` and `changeCheckboxCell` and any other function that uses it.
4. Yes, they should be unified and adjusted to match current `develop`
5. Progress bar should accept the range of Floats from 0 to 100. Ratings should accept Integers from 1 to 5.

jansiegel (approved):
I think all my previous points have been addressed, but one of @adamu-handsontable's points is missing:

> Progress bar should accept the range of Floats from 0 to 100. Ratings should accept Integers from 1 to 5.

So I'm approving this PR and passing the review of that one point to @adamu-handsontable .

examples/next/docs/react/demo/src/renderers/ProgressBar.tsx
```diff
@@ -1,10 +1,21 @@
...
+import { getRangeValue, HandsontableProps } from './utils';
+import { MESSAGE } from '../constants';
 
-export function ProgressBarRenderer(props: RendererProps) {
-  addClassWhenNeeded(props);
+const minAllowedValue = 0;
+const maxAllowedValue = 100;
 
+export const ProgressBarRenderer = (props: HandsontableProps) => {
+  if (props.cellProperties.valid !== false) {
```
jansiegel, line 9: @maciejbo-handsontable 
I found one additional issue with the demo - when you pass invalid data for the progress bar or stars columns, they're not marked as invalid. 
That happens because Handsontable doesn't trigger validation by default when loading data.
To get around it it, you can manually trigger the validator if it wasn't triggered already.

Take a look at my suggestion, and if you think it works all right, could you implement it for the stars renderer as well?

```suggestion
export const ProgressBarRenderer = (props: HandsontableProps) => {
  let isValid = props.cellProperties.valid;

  // Run the validator for the cell at initization.
  if (isValid === void 0) {
    (props.cellProperties.validator as Function)(props.value, (isValueValid: boolean) => {
      isValid = isValueValid;
    });
  }

  if (isValid === true) {
```

jansiegel (changes requested):

adamu-handsontable:
Let's skip the validation of stars and progress for now. We'll do that in a separate PR later if necessary.

jansiegel (approved):

examples/next/docs/react/demo/src/hooksCallbacks.ts
```diff
@@ -38,8 +33,9 @@ export const addClassesToRows: AddClassesToRows = (
     return;
   }
 
-  // Add class to selected rows
-  if (cellProperties.instance.getDataAtRowProp(row, "0")) {
+  const rowData = cellProperties.instance.getSourceDataAtRow(row) as RowObject;
+
+  if (!!rowData?.checked) {
```
wszymanski, line 38:
```suggestion
  if (rowData?.checked !== true) {
```

examples/next/docs/react/demo/src/hooksCallbacks.ts
```diff
@@ -63,14 +59,12 @@ export const drawCheckboxInRowHeaders: DrawCheckboxInRowHeaders = function drawC
...
-
-  if (row >= 0 && this.getDataAtRowProp(row, "0")) {
-    input.checked = true;
-  }
-
+  const input = document.createElement('input');
+  const rowData = this.getSourceDataAtRow(row) as RowObject;
+  
+  input.type = 'checkbox';
+  input.checked = !!rowData?.checked;
```
wszymanski, line 66:
```suggestion
  input.checked = rowData?.checked === true;
```

examples/next/docs/react/demo/src/renderers/ProgressBar.tsx
```diff
@@ -1,10 +1,41 @@
...
+    prop,
+    value,
+    cellProperties
+  } = props;
+  let isValid = cellProperties.valid;
+
+  baseRenderer(cellProperties.instance, TD, row, col, prop, value, cellProperties);
+
+  // Run the validator for the cell at initialization.
+  if (isValid === void 0) {
```
wszymanski, line 23: Maybe its worth to move up method, beyond the renderer declaration, to not repeat function creation and make it more clear by naming it. Suggestion up to you.

examples/next/docs/react/demo/src/constants.ts
```diff
@@ -1,1205 +1,1108 @@
...
+}];
 
-export const SELECTED_CLASS = "selected";
-export const ODD_ROW_CLASS = "odd";
+export const SELECTED_CLASS = 'selected';
+export const ODD_ROW_CLASS = 'odd';
+
+export const MESSAGE = {
+  BAD_VALUE: '#bad-value#'
+};
```
wszymanski, line 1108: Do we need to introduce this variable using object?

```suggestion
export const BAD_VALUE_MESSAGE = '#bad-value#';
```

examples/next/docs/react/demo/src/renderers/Stars.tsx
```diff
@@ -1,8 +1,40 @@
...
+  let isValid = cellProperties.valid;
+
+  baseRenderer(cellProperties.instance, TD, row, col, prop, value, cellProperties);
+
+  // Run the validator for the cell at initialization.
+  if (isValid === void 0) {
+    (props.cellProperties.validator as Function)(props.value, (isValueValid: boolean) => {
+      isValid = isValueValid;
+    });
+  }
```
wszymanski, line 27: The same as in `ProgressBar.tsx`.

wszymanski (commented):
I have some suggestions. I'll test example after changes in code.

examples/next/docs/react/demo/src/renderers/ProgressBar.tsx
```diff
@@ -1,10 +1,41 @@
...
+    (cellProperties.validator as Function)(value, (isValueValid: boolean) => {
+      isValid = isValueValid;
+    });
+  }
```
wszymanski, line 27: Maybe its worth to move up method, beyond the renderer declaration, to not repeat function creation and make it more clear by naming it. Suggestion up to you.

krzysztofspilka:
@wszymanski as discussed, you will be assigned to this task with someone else from the Dev team as reviewer. Please accept all your remarks to the code. Let's treat this as a work in progress, thanks.

evanSe:
Closing due to inactivity

Issue #9986, opened 2022-09-30
Docs: Missing closure for ::


aninde:
### Description

![image](https://user-images.githubusercontent.com/30186818/193290174-118ffa6a-cace-4f43-b710-461521204286.png)
Closure is missing `:::`
### Links

https://handsontable.com/docs/javascript-data-grid/angular-language-change-example/#overview

PR #9991, opened 2022-10-03
Docs: Add missing ::: marks

kirszenbaum:
This PR:
- Adds missing `:::` marks

[skip changelog]

### Related issues:
- #9986


aninde (approved):
I checked locally, in production mode and everything is correct. Great that you noticed the extra ones, because a fix in `docs/content/guides/cell-functions/cell-editor.md` caused the entire block of code to now display correctly. It did not show up on the current live version of the documentation at all.

Issue #9949, opened 2022-09-26
Docs: Missing Handsontable import for React demo


aninde:
### Description

List of imports in below demo need to be updated with 
`import Handsontable from 'handsontable';`
### Links

https://dev.handsontable.com/docs/next/react-data-grid/row-prepopulating/#example

PR #9996, opened 2022-10-06
Docs: Add missing Handsontable imports to React examples

kirszenbaum:
This PR:
- Adds missing Handsontable imports to React examples

[skip changelog]

### Related issues:
- #9949


kirszenbaum:
@aninde – there are many other React examples that don't have the `import Handsontable from 'handsontable';` import – I understand that they don't need it?

Issue #9734, opened 2022-08-02
[10.0+] setCellMeta doesn't change the cell type


adrianszymanski89:
### Description

If we define the cell type as, for example, `password` and then we want to change the type to `text` by using `setCellMeta()` and back again to `password`, the change doesn't occur. This behavior applies to all available cell types.

This issue was present before and was solved in version 8: https://github.com/handsontable/handsontable/issues/4793

### Steps to reproduce

1. Open an example. The default cell type is set to `password`. Then, in order, we execute `setCellMeta()` to change the type to `text` and then back to `password`.

### Demo

1) v. 9 (correct behavior) - http://jsfiddle.net/aszymanski/0y1phudq/1/
2) v. 10 (and higher, incorrect behavior) - http://jsfiddle.net/aszymanski/0y1phudq/2/ 

### Your environment

* Browser name and version: Google Chrome 103
* Operating system: macOS Monterey 12.4

AMBudnik:
It works when we use `editor` and `renderer` instead of `type`

```js
hot.setCellMeta(0, 0, 'type', 'text');
hot.render();

hot.setCellMeta(0, 0, 'editor', 'password');
hot.setCellMeta(0, 0, 'renderer', 'password');
hot.render();
```

AMBudnik:
Thank you for fixing guys! I'm closing this issue as done.

PR #10000, opened 2022-10-07
Resolve the issue with `type` sub-properties (`editor` + `renderer`) not being updated with `setCellMeta` calls.

jansiegel:
### Context
This PR fixed the issue from #9734 by introducing a new, internally used meta property.
After this change, the automatically-added meta props (like `renderer`, `editor` etc, added when changing the `type` meta) can be overwritten both by setting the actual props in question and setting another `type`.

### How has this been tested?
Added new test cases.

### Types of changes

### Related issue(s):
1. #9734 

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:


jansiegel:
@budnix If you have any suggestions on how to fix this more elegantly without modifying the principles of the current meta manager, I'm open for suggestions.

budnix (changes requested):
Thanks for the PR. However, I think there is still some work to be done. For example, when the table is initialized with `type: "date"` and after changing the cell type to `text`, the cell contains artifacts from the previous cell type.
Kapture 2022-10-12 at 11 43 58
(`text` cell type contains `date` validator). The same can be spotted for different cell types. For example, switching from the `password` cell type to `text` lefts the `copyable` property as `false`.

jansiegel:
@budnix Correct me if I'm wrong, but it seems like it's been working like this for a long time? I've checked version `4.0`, and it looks like it does leave the previously-used validators, etc behind as well: http://jsfiddle.net/js_ziggle/2hf6ctzv/.

budnix:
> @budnix Correct me if I'm wrong, but it seems like it's been working like this for a long time? I've checked version `4.0`, and it looks like it does leave the previously-used validators, etc behind as well: http://jsfiddle.net/js_ziggle/2hf6ctzv/.

True. Then, I'd have one suggestion. WDYT to rename the property to `_automaticallyAssignedMetaProps`? A similar pattern is used by one of the [meta mods](https://github.com/handsontable/handsontable/blob/develop/handsontable/src/dataMap/metaManager/mods/extendMetaProperties.js#L53) and the _[ManualColumnFreeze](https://github.com/handsontable/handsontable/blob/develop/handsontable/src/plugins/manualColumnFreeze/manualColumnFreeze.js#L130)_ plugin.

budnix (approved):
👍

aninde:
Works well on every [tested browser](https://app.qase.io/run/HOT/dashboard/161), unit test also pass. I checked locally whether the change from each type to another and back works.

Issue #9545, opened 2022-05-30
handsontable disappears on firefox after upgrading to version 93


JuanjoArmellino:
hello, I have a problem with HT, it worked fine in firefox until version 92, starting version 93 the table goes white on scroll or window resize.. im not sure if its a firefox problem or something they changed is causing this.
i have to admit it only happens on a very rare case of having a table inside a bootstrap card with the card resizing automaticly to the windows size.. but it happens every time.
works fine on chrome

### Steps to reproduce
load the sample file and scroll the table with the mouse or maximize / restore / resize the window

### Demo
https://jsfiddle.net/p4u7mt0f/

### Your environment
* Handsontable version: HT 7.4, HT10 or HT12 
* Browser name and version: firefox after version 93
* Operating system: windows 10 21h2

AMBudnik:
Thank you for sharing the report @JuanjoArmellino 

I'm already on FF 100. But you are right. Chrome opens the demo and it works well, while Firefox does not show the table. I tested the CSS settings and by commenting our the `overflow` I got the table working on Firefox.

![Screenshot 2022-05-31 at 08 39 00](https://user-images.githubusercontent.com/10757813/171108577-da7e8383-1674-4fa3-a25b-c920457e7867.png)

We'll investigate it further and check what are the connections to https://github.com/handsontable/handsontable/issues/2806

AMBudnik:
Minified demo: https://jsfiddle.net/6r7kvdwj/10/ (still contains a lot of DIVs with `flex` settings)
I've added comments on what changes the behavior between Firefox and Chrome in the comments
Tested with Handsontable v12.1.2 on Chrome 103 / macOS Monterey

wszymanski:
I've done investigation by myself to determine why the problem occurs. I've found that node created as clone sometimes has height set as `0`. The first call to `alignOverlaysWithTrimmingContainer` methid is done from `Table` element's constructor and it's calculated properly, while the second one (called from `draw` method) isn't calculated properly on FF.

https://github.com/handsontable/handsontable/blob/a2033859014e4e115cf54e638b23727f91c07f29/handsontable/src/3rdparty/walkontable/src/table.js#L108-L110

https://github.com/handsontable/handsontable/blob/a2033859014e4e115cf54e638b23727f91c07f29/handsontable/src/3rdparty/walkontable/src/table.js#L308-L309

[...]

PR #10005, opened 2022-10-10
Fixed problems related to placing instance in some flex containers and setting overflow to `hidden`, `scroll` or `auto`.

wszymanski:
### Context

I described part of the investigation [within a comment](https://github.com/handsontable/handsontable/issues/9545#issuecomment-1267354851). It seems that after not setting the position, some properties are inherited from the flex container. After absolutely positioning a cloned element the problem isn't occurring. You can see [here an example ](https://gist.github.com/wszymanski/e6f2fa8656b8f68e57d20c0d105f649a) of some use case when setting overflow to `scroll`, `auto` or `hidden` may be needed.

I couldn't find much more simpler example for fixed case. Thus, I've decided to not write e2e tests for the solution.

### How has this been tested?

I tested the solution on a found [minimal example](https://gist.github.com/wszymanski/625231e0c69b7535211de13acee83ba9) showing the problem on FF 93.

### Types of changes

### Related issue(s):
1. #9545 

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:


jansiegel (approved):
👍

aninde:
We could really use the e2e test with a minimal example, @wszymanski could you add one?

Issue #10025, opened 2022-10-14
Docs: Add a note that custom class doesn't work properly with `ColumnSummary`


kirszenbaum:
To do:
- On the Column summary page, add note saying that you currently can't add a custom class to a row affected by the `ColumnSummary` plugin

### Links
- https://jsfiddle.net/aszymanski/snjvd08k/1/
- https://handsontable.com/docs/javascript-data-grid/column-summary/

PR #10026, opened 2022-10-14
Docs: Add info that you can't change class of the column summary destination row

kirszenbaum:
This PR
- Adds info that you can't change class of the column summary destination row

[skip changelog]

### Links:
- http://localhost:8080/docs/react-data-grid/column-summary/#step-4-provide-the-destination-cell-s-coordinates
- http://localhost:8080/docs/react-data-grid/api/options/#classname

### Related issues:
1. https://github.com/handsontable/handsontable/issues/10025

### DIFF
--- a/docs/content/guides/columns/column-summary.md
+++ b/docs/content/guides/columns/column-summary.md
@@ -415,7 +415,7 @@ columnSummary={[
 :::
 
 ::: tip
-Don't change the CSS class of a column summary's destination row. Otherwise, the summary may not work properly.
+We don't recommend changing the CSS styling of the summary's destination row (the summary may not work properly).
 :::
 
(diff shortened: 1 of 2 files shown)


adrianszymanski89 (changes requested):
Maybe something like that would sound nicer?

> We don't recommend changing CSS styling of summary's destination row class. Otherwise, the summary may not work properly.

kirszenbaum:
@adrianszymanski89 – how about [this](https://github.com/handsontable/handsontable/pull/10026/commits/23d84e11b3ada02333f82e5965d6c128569973e5)?

adrianszymanski89 (approved):
Looks good, thanks :)

Issue #9993, opened 2022-10-05
Docs: Links from the table of content on Release notes doesn't work


aninde:
### Description

Links with version number doesn't work. That one with text works. It appears in every version of the documentation.

Links from https://handsontable.com/docs/javascript-data-grid/release-notes/#_12-1-2 to https://handsontable.com/docs/javascript-data-grid/release-notes/#_7-4-0 doesn't work.

### Links

https://handsontable.com/docs/javascript-data-grid/release-notes/#_12-0-0
https://handsontable.com/docs/react-data-grid/release-notes/#_12-0-1

PR #9898, opened 2022-09-13
React docs: Correct the permalink `href` attributes on headers duplicated between vanilla/react versions.

jansiegel:
### Context
This PR should correct the `-2`, suffixes in the `href` attributes on the permalinks of the headers duplicated between vanilla/react docs versions - it's a continuation of #9783.

### How has this been tested?
Tested using `docs:start:no-cache`.

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature or improvement (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Additional language file or change to the existing one (translations)

### Related issue(s):
1. #9892 
2. #9783

[skip changelog]

PR #9783, opened 2022-08-16
React Docs: Bypass the Vuepress' ID-duplication prevention mechanism.

jansiegel:
### Context
This PR bypasses the Vuepress' mechanism preventing the creation of DOM elements with duplicated `id`-s.

### How has this been tested?
Tested mostly on `javascript` and `react` versions of `/docs/next/react-data-grid/cell-renderer/`, where the `::: only-for` containers make the problem visible.

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature or improvement (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Additional language file or change to the existing one (translations)

### Related issue(s):
1. #9750 

### Affected project(s):
- [ ] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

- [x] I have reviewed the guidelines about [Contributing to Handsontable](https://github.com/handsontable/handsontable/blob/master/CONTRIBUTING.md) and I confirm that my code follows the code style of this project.
- [x] I have signed the [Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLScpMq4swMelvw3-onxC8Jl29m0fVp5hpf7d1yQVklqVjGjWGA/viewform?c=0&w=1)
- [ ] My change requires a change to the documentation.

[skip changelog]

PR #10034, opened 2022-10-18
Changed already bypassed Vuepress' ID-duplication prevention mechanism

wszymanski:
### Context

PRs #9783 and #9898 introduced mechanism of removing a number from the end of every slug. It has been done using legacy `permalinkHref` method, found by @jansiegel (see https://github.com/valeriangalliat/markdown-it-anchor/blob/master/index.js#L100).

I've tried to find alternative way for changing slugs. However, `slugify` method seems to be not enough. A built in mechanism of creating unique slugs (https://github.com/valeriangalliat/markdown-it-anchor/blob/8918e4eba26df9b13819acde5a9bf4e4fb8d9816/index.js#L60-L64) which uses `slugify` isn't customizable. Thus, I've decided to continue using the `permalinkHref` method. It's worth to mention that so far existing workaround hasn't been perfect for end user. Headers duplicated within one page (see how API pages look like) and headers a containing number (ie. `Vue 2`) have been stripped. However, the solution was quite simple, which is its advantage. I have decided not to change the logic too much. I'm doing almost the same, but just for headers which are placed between `::: only-for` and `:::` tags.

I haven't found a way of executing `markdown-it-conditional-container` plugin (which removes tokens) before enabled by default [markdown-it-anchor](https://github.com/valeriangalliat/markdown-it-anchor) plugin. I added an extra conditional for mechanism introduced by @jansiegel. There is an extra logic which is looking for every slug within a page placed between `::: only-for` and `:::` tags. Only slug already handled by `permalinkHref` and placed within conditional container can be stripped.

Please keep in mind that I'm determining slugs which should be changed within `permalinkHref` and using it in `callback` method. The first method is always executed before the second.

https://github.com/handsontable/handsontable/blob/fd0530c62013ef44eae7200a65c06d4ac243e240/docs/.vuepress/config.js#L22
https://github.com/handsontable/handsontable/blob/fd0530c62013ef44eae7200a65c06d4ac243e240/docs/.vuepress/config.js#L137
https://github.com/handsontable/handsontable/blob/fd0530c62013ef44eae7200a65c06d4ac243e240/docs/.vuepress/config.js#L154-L156

*[skip changelog]*

### How has this been tested?

I tested some pages, such as:

- http://localhost:8000/docs/javascript-data-grid/release-notes/
- http://localhost:8000/docs/javascript-data-grid/api/bind-rows-with-headers/ (see that there are two HTML headers with `BindRowsWithHeaders` inner text)
- http://localhost:8000/docs/javascript-data-grid/vue-custom-context-menu-example/ (see that there is HTML header with `Custom context menu in Vue 2` inner text and `id` set to `custom-context-menu-in-vue-2` - previously the ID was set to `custom-context-menu-in-vue`; you can find more such use cases while searching `.md` files using `^#+(.*)?\d$` regular expression).

### Types of changes

### Related issue(s):
1. #9993
2. #9783
3. #9898


jansiegel (changes requested):
The new logic makes sense, but if I'm reading this correctly, it's very processing-heavy - every anchor on the page triggers the `for` loop that goes through all of the tokens to check if the current link is within the container, most of the times for nothing (as most of the times there's nothing to change)

Do you think you could modify the code to first check if the slug matches the problematic pattern (`#[something]-[number]`), and only if so, trigger the token-lookup logic?

wszymanski:
Thanks for your suggestion, the only thing which can be done from my perspective is code added within https://github.com/handsontable/handsontable/pull/10034/commits/55dc0522bc4c3c7f307e83e4b63713215c96ef87.

jansiegel (changes requested):
1. The fix from #9783 seems to not be working anymore:

    On `docs/react-data-grid/cell-renderer`, the link to `Use a cell renderer` in the sidebar points to `#use-a-cell-renderer`, while the element it's trying to target has an `id` of `use-a-cell-renderer-2`.
    This currently works fine in production (both work on the id of `use-a-cell-renderer`).
    
2. The anchor link next to the content header in the case of `1)` points to `##use-a-cell-renderer`.
    
    image


3. The logic you implemented is quite lengthy - maybe you could move it to a separate file for the config file to remain readable?

Issue #8533, opened 2021-07-21
Column Filter Input Loses Focus


AMBudnik:
@ajangles commented on [Fri Jun 05 2020](https://github.com/handsontable/handsontable/issues/6986)

### Description

We have 27 columns in out table, and after scrolling to the complete right and selecting a filter, the selected input loses focus and subsequently the inputs are all out of order and the columns are out of alignment

### Steps to reproduce

1. Create table with large amount of columns and add filters
2. Scroll to the right
3. Start typing in a filter input

### Demo

This demo is pretty extreme in number of columns perhaps, but easily represents the problem:

https://jsfiddle.net/n2t8vhzs/

![handsontablefilter](https://user-images.githubusercontent.com/6114691/83842650-c69dc880-a746-11ea-8e3e-0fc110e1b371.gif)

![handsontablefilter](https://user-images.githubusercontent.com/6114691/83842655-c8678c00-a746-11ea-8534-f8dcc0013d71.png)

### Your environment
* Handsontable version: 7.4.2
* Browser Name and version: All browsers
* Operating System: Windows 10

---

@AMBudnik commented on [Fri Jun 05 2020](https://github.com/handsontable/handsontable/issues/6986#issuecomment-639290974)

Thank you for sharing the issue report, Javier.

I can see that you use our `Filter as you type` example from the Filtering tutorial https://handsontable.com/docs/7.4.2/demo-filtering.html

I have replicated the issue in v 7.4.2 as well as our new beta2-rev11 of v 8.0.0. https://jsfiddle.net/AMBudnik/1svhojaz/

Tested on Chrome 80 / macOS Catalina
Related issue https://github.com/handsontable/handsontable/issues/5541 (fixed fo v7.0.0)

---

@AMBudnik commented on [Mon Jul 06 2020](https://github.com/handsontable/handsontable/issues/6986#issuecomment-654212570)

It works the same in 8.0.0-beta2-rev18 https://jsfiddle.net/moewkuL3/

---

@AMBudnik commented on [Tue Aug 11 2020](https://github.com/handsontable/handsontable/issues/6986#issuecomment-671940326)

The issue remains the same in v 8.0.0 https://jsfiddle.net/w5as0juk/

---

@wojciechczerniak commented on [Sun Oct 18 2020](https://github.com/handsontable/handsontable/issues/6986#issuecomment-711157823)

The demo in our documentation is a very simple one. Custom code should be adjusted if you want more columns.

[...]

PR #10042, opened 2022-10-19
Docs: Add info that Filter as you type demo is just a showcase

kirszenbaum:
This PR:
- Adds info that the [Filter as you type](https://handsontable.com/docs/react-data-grid/column-filter/#filter-as-you-type) demo is just a showcase

[skip changelog]

### Links
- http://localhost:8080/docs/react-data-grid/column-filter/#filter-as-you-type

### Related issues
- #8533


adrianszymanski89 (approved):
LGTM 👍

Issue #10055, opened 2022-10-24
Minor improvements to the layout of the documentation


krzysztofspilka:
### Description

This is an issue aggregating three problems reported by @aninde in #10049 #10050 #10051

1. Opened search engine is shifted when the window has a resolution of 800 x 600 or width less than 850 px

> When search input is opened and class-focused, the whole div class="nav-links-right" is misplaced.
The bug occurs from a window width of 850 px on the desktop. Not on mobiles and iPad.
Most users view the documentation at this resolution; it should look good.

2. Svg images have the wrong path
The SVG image has the wrong path; the picture is not displayed.
![image](https://user-images.githubusercontent.com/8048526/197622792-76a03933-8550-463b-a504-832d127ffeb0.png)

3. Dark theme causes the horizontal cell boundaries in the Vuex demos to disappear
Dark theme pollutes the preview of the demos related to Vuex. Therefore, it does not show horizontal boundaries between cells
![image](https://user-images.githubusercontent.com/8048526/197622883-c1509250-ff82-4a0b-81e9-2099279b26cb.png)

krzysztofspilka:
@aninde
>Most users view the documentation at this resolution; it should look good.
It's the fist choice of 4.85% of all users.

PR #10056, opened 2022-10-24
Fixed: minor improvements to the layout of the documentation

krzysztofspilka:
### Context

Combined changes from #10055 originally reported by @aninde. Changes included:

- Fixed search bar, especially on mobile devices
- Fixed 404 error caused by the wrong path on the introduction page
- Fixed the style of the tables in the Vuex/Redux pages


docs/content/guides/upgrade-and-migration/release-notes.md
```diff
@@ -25,7 +25,7 @@ See the full history of changes made to Handsontable in each major, minor, and p
 
 Released on 22nd of September, 2022 - [view docs](docs/12.1/)
 
-### Fixed
+#### Fixed
```
budnix, line 28: The header type is already changed on the release branch https://github.com/handsontable/handsontable/blame/release/12.2.0-docs/docs/content/guides/upgrade-and-migration/release-notes.md#L52. I'd consider to undo the change to avoid conflicts 🤔 WDYT?

jansiegel:
The search bar no longer gets pushed below the top panel, but now when I resize the window it goes over the framework dropdown, then after the "API reference" link. This means, that for some resolutions they might be unusable. 
@aninde, could you confirm?

https://user-images.githubusercontent.com/7397134/197710603-8467a23c-33ca-4c83-971a-1ab1fcb1681a.mov

aninde (approved):
Great, I can see all the fixes done.
React logo is visible in production mode (npm run docs: build).
The open search engine is also responsive now, and from the width of 768 px, `div.nav` links are flipped to change the version, Github, etc. to the left sidebar

I would like to check the preview demos of Vuex and Redux on staging environment. Thanks for noticing the Redux example, I didn't report it.

docs/content/guides/upgrade-and-migration/release-notes.md
jansiegel: As @krzysztofspilka's unavailable, I'll make this change 👍.

jansiegel (approved):

budnix (approved):

aninde:
> The search bar no longer gets pushed below the top panel, but now when I resize the window it goes over the framework dropdown, then after the "API reference" link. This means, that for some resolutions they might be unusable. @aninde, could you confirm?
> 
>  search.overlap.mov

The current documentation also works like that and it's the breakpoint 786 px, you can't see it on mobile.

Issue #9891, opened 2022-09-12
Internal: The `verify-bundles` freeze's helper script fails.


jansiegel:
### Description

The `script/verify-bundles.mjs` script fails with the following error when trying to use the `npm run freeze` script.

```
node:internal/errors:477
    ErrorCaptureStackTrace(err);
    ^

TypeError [ERR_IMPORT_ASSERTION_TYPE_MISSING]: Module "/workspace/handsontable/package.json" needs an import assertion of type "json"
    at new NodeError (node:internal/errors:387:5)
    at validateAssertions (node:internal/modules/esm/assert:82:15)
    at defaultLoad (node:internal/modules/esm/load:84:3)
    at nextLoad (node:internal/modules/esm/loader:165:28)
    at ESMLoader.load (node:internal/modules/esm/loader:608:26)
    at ESMLoader.moduleProvider (node:internal/modules/esm/loader:464:22)
    at new ModuleJob (node:internal/modules/esm/module_job:63:26)
    at ESMLoader.#createModuleJob (node:internal/modules/esm/loader:483:17)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:441:34)
    at async Promise.all (index 0) {
  code: 'ERR_IMPORT_ASSERTION_TYPE_MISSING'
}
```

This is most probably caused by the changes introduced in https://github.com/handsontable/handsontable/pull/9267.

### Steps to reproduce

1. From the root directory, run `node --experimental-json-modules ./scripts/verify-bundles.mjs`
2. ... or run `npm run freeze` and wait until it calls the `verify-bundles` script.

warpech:
@jansiegel I think that about half of `verify-bundles` has the potential to be reused in https://github.com/handsontable/handsontable/issues/9879. If you agree, what lines would you recommend? Maybe this script can be changed to accept a parameter that would run `verify-bundles` in a prerelease mode?

jansiegel:
@warpech Yup, good idea - after modifying it to accept an argument, the entirety of the script can be reused, I think. All it does is verifies if the classes exported from the packages contain the correct version number, which could potentially be useful in #9879.

AMBudnik:
Closed as done in Handsontable v12.3.0.

PR #10068, opened 2022-10-26
Fix the `verify-bundles` freeze helper script.

jansiegel:
### Context
This PR fixes the `verify-bundles` script by:
1. Changing the extensions for the ES builds from `js` to `mjs` for:
    1. `@handsontable/react`
    2. `@handsontable/vue` 
    3. `@handsontable/vue3`
2. Added entries for the UMD and minified UMD builds to the `exports` section of `package.json` for:
    1. `@handsontable/react`
    2. `@handsontable/vue` 
    3. `@handsontable/vue3`
3. Adding the JSON type assertions to the dynamic imports in the `verify-bundles` script.

I haven't changed the `exports` entry in `@handsontable/angular` as currently the only export it uses is targeted at its `umd` build, which should probably be reviewed in a separate task.

### How has this been tested?
Run the `verify-bundles` script by running `node ./scripts/verify-bundles.mjs` from the root directory.

### Types of changes

### Related issue(s):
1. #9891 

### Checklist:


scripts/verify-bundles.mjs
```diff
@@ -39,7 +39,10 @@ async function verifyBundles() {
       className: 'HotTable'
     }
   };
-  const { default: mainPackageJson } = await import('../package.json');
+  const { default: mainPackageJson } = await import('../package.json', {
+    assert: { type: 'json' }
+    }
```
budnix, line 44: Weird formatting 🤔

budnix (changes requested):
Everything looks good 👏 

I have one suggestion that bothers me. The latest Node version that we use supports top-level await syntax, the script does only one job, WDYT about shifting the whole code to one tab left and extracting the logic out of the `verifyBundles` function? Then the function can be removed. I mean:
```js
async function verifyBundles() {
  // the script body
}
verifyBundles();
```
to
```js
// the script body
```

scripts/verify-bundles.mjs
jansiegel: I agree. The linter forced me to do this 🙄 https://github.com/handsontable/handsontable/pull/10068/commits/1326ed95b00e1787d3945724f66525915d5ee2ba

budnix: What about a one-liner? 
`const { default: mainPackageJson } = await import('../package.json', { assert: { type: 'json' }});`
or 
```js
const { 
  default: mainPackageJson
} = await import('../package.json', { assert: { type: 'json' }});
```

budnix (approved):
Looks good 👌

Issue #9879, opened 2022-09-08
Publish prerelease packagages to NPM automatically


warpech:
Currently, we don't have a simple way to publish pre-release NPM packages for internal and external testing before releasing the final version.

The proposed solution to solve that is to implement an automatic NPM package creation that goes straight to the public† NPM registry.

Requirements:

- the package should be automatically created after any commit to the `develop` or `release/*` branch (to make it easier to test a snapshot of what might be released)
- a developer should be able to manually generate a package from a `feature/*` branch (or any other branch) through manual workflow dispatch in GitHub Actions on the page https://github.com/handsontable/handsontable/actions/workflows/publish.yml (to make it easier to test work in progress)
- both employees and customers should be able to use this package (that's why we put it on the public registry)
- use a proper SemVer suffix in the package version number, that is not misleading
   - we decided to go with the version number `0.0.0-next-<shortSha>-<buildDate>` in an [internal discussion](https://handsoncode.slack.com/archives/C4PV4S33Q/p1662644133553129) (e.g. `0.0.0-next-2b3b4b-20221009`)
     - The release branch (`release/X`) is an exception. On that branch, the version number should contain the version set while creating the branch. For example, for `release/12.2.5` the pre-released version would look like `12.2.5-next-2b3b4b-20221009`
- should entail the packages for core, Angular, React, Vue 2, and Vue 3 (a prerelease wrapper package should depend on a corresponding prerelease core package) 
   - https://www.npmjs.com/package/handsontable
   - https://www.npmjs.com/package/@handsontable/angular
   - https://www.npmjs.com/package/@handsontable/react
   - https://www.npmjs.com/package/@handsontable/vue
   - https://www.npmjs.com/package/@handsontable/vue3
- should be picked up by a CDN, so the following URLs should be picked up: `https://cdn.jsdelivr.net/npm/handsontable@0.0.0-next-<shortSha>-<buildDate>/dist/handsontable.full.min.js`, `https://cdn.jsdelivr.net/npm/@handsontable/react@0.0.0-next-<shortSha>-<buildDate>/dist/react-handsontable.min.js`
    - make sure that the "latest" version (`https://purge.jsdelivr.net/npm/handsontable@latest/dist/handsontable.css`) is still resolved by JSDelivr as the last stable version, not the pre-release
- during the work, make VERY sure that we accidentally don't publish a non-prerelease to NPM, because these are hard to pull out
[...]

PR #10070, opened 2022-10-27
build prerelease package and publish in NPM

maciejbo-handsontable:
### Context
with this change we can publish HOT prereleases in a faster and easier way.
Workflow runs automatically after commit to branches `release/**` or `develop` (only if tests passed succesfully) or can be launched manually on any other branch.

### How has this been tested?
we pushed a lot of changes on different branches in side repo made of forked `handsontable` - demanded packages were published succesfully in NPM with tag `next` (under the temporary name, we haven't used `handsontable` for tests).

### Types of changes

### Related issue(s):
1. https://github.com/handsontable/handsontable/issues/9879
2.
3.

### Affected project(s):
- [x] `handsontable`
- [x] `@handsontable/angular`
- [x] `@handsontable/react`
- [x] `@handsontable/vue`
- [x] `@handsontable/vue3`

### Checklist:

- [] My change requires a change to the documentation.

[skip changelog]


aninde:
[Performed tests](https://app.qase.io/run/HOT/dashboard/163/cc995573eacc69748c06a73f80eb9110068d1ebe/f3eaa48a1440af02d7302ff0e3fedc767c10ac95): Automatic publish after a change in develop, manual publication on a feature branch, Downloading and installing a package from each version, included in the each if the names match.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,31 @@
+name: Publish prerelease in NPM
+
+on:
+  workflow_dispatch:
+  workflow_run:
+    workflows: [Tests]
+    types: [completed]
+    branches: [develop, 'release/**']
```
jansiegel, line 8: This makes the workflow run not only on the actual release branch but also the branch being a base for the docs PR during the release process, [as they're usually named `release/[version]-docs`](https://kb.handsontable.com/display/BOOK/Document+a+Handsontable+release).
I'm not sure if that's what we want.

This might be a good time to rethink the docs PR branch name, as `release/[version]-[something]` might be considered an alpha/beta/(...) release branch.

scripts/prerelease.mjs
```diff
@@ -0,0 +1,19 @@
...
+  .stdout
+  .toString()
+  .slice(0, 7);
+const currentDate = moment().format('YYYYMMDD');
+const currentBranchName = (await spawnProcess('git rev-parse --abbrev-ref HEAD', { silent: true })).stdout;
+let versionNumber = '0.0.0';
+
+if (currentBranchName.startsWith('release/')) {
+  versionNumber = currentBranchName.split('/')[1];
+}
```
jansiegel, line 15: This will determine the version number as `[version]-docs` in case of running on `release/[version]-docs` branch, resulting in publishing a package versioned `[version]-docs-next-[commitSha]-[currentDate]`.

I think we can either rethink the release docs branch naming, or maybe get the package version number from `package.json`, which should already be correct on the release branch.

jansiegel (commented):
Apart from the comments in the code, I have one additional note:
If I'm not mistaken, in this scenario:
1. Create a branch `feature/issue-xyz`
2. Create a PR from `feature/issue-xyz` to `develop`

The `Test` workflow will trigger the `Publish` workflow using the version of `publish.yml` from **`develop`**, not the feature branch. I don't think this would be a problem most of the time, though.

scripts/prerelease.mjs
budnix: The best fix would be reading the current Handsontable version from `package.json` file (the one in the root directory). I would propose using the same import as [here](https://github.com/handsontable/handsontable/blob/develop/scripts/utils/filesystem.mjs#L2), for example.

.github/workflows/publish.yml
budnix: If, in the docs release branch, there would be a change that triggers the tests, then yes the prerelease would be created. Otherwise, the workflow is skipped. IMO leaving this as is would be ok.

scripts/prerelease.mjs
budnix: @maciejbo-handsontable can I ask you to update the PR and add the `import` that I mentioned above?

maciejbo-handsontable: @budnix @jansiegel Hi! Requested change has been implemented (but instead of `package.json`, I'm reading version from  `hot.config.js` file).

.github/workflows/publish.yml
budnix: 🤷‍♂️ cc @warpech

warpech: During the code freeze period, any change in the package files should be worth testing, so it should be worth having a pre-release.

However, in the `release/[version]-docs` branches, I think we are only supposed to make changes in JSDoc and sometimes they are done in many separate commits.

I agree that publishing pre-release packages that often might be confusing.

I recommend we set the workflow like this:

```
    branches:
      - 'develop'
      - 'release/**'
      - '!release/**-docs'
```

Source: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions

jansiegel: @warpech But wouldn't setting `'!release/**-**'` block "postfix" releases (like `13.0.0-alpha` etc) from being published?

warpech: We can easily change the rule at any time if the above suggestion becomes is too limiting.

budnix (changes requested):

warpech (approved):
A manually triggered prerelease (https://github.com/handsontable/handsontable/actions/runs/3515404842) was successfully published (https://www.npmjs.com/package/@handsontable/react/v/0.0.0-next-9ec04ce-20221121) and I was able to use it in JSFiddle (https://jsfiddle.net/rbdk92h6/).

Good job everyone!

budnix (approved):

jansiegel (commented):
@maciejbo-handsontable @warpech 
Everything looks fine, but I have one final question - do we want to tag every build created with this workflow as `next`? This way, the `next` tag can mean a build from the release branch (which would be a proper use of the tag, I think) or a random manual build from a feature branch.

warpech:
Yes, for me it is fine that it always is `next`. There was a proposal to use the branch name, but it was not picked up. We can always change it later if there is a strong need.

jansiegel (approved):
In that case, 👍

Issue #10061, opened 2022-10-26
Adding missing e2e test with flex container


aninde:
### Description

[As we agreed](https://github.com/handsontable/handsontable/issues/9545#issuecomment-1288796521), we should add missing [one](https://gist.github.com/wszymanski/625231e0c69b7535211de13acee83ba9) or two tests from the PR https://github.com/handsontable/handsontable/pull/10005.

### Your environment
* Handsontable version: 12.2
* Browser name and version: not related
* Operating system: not related

PR #10083, opened 2022-10-31
Added some walkontable tests #10061

wszymanski:
### Context

Just some extra tests for #10005. I've added CSS inline and created new `spec.js` file as `beforeEach` is quite specific.

*[skip changelog]*

### Related issue(s):
1. #10061


jansiegel (approved):
Seems to fail on FF when commenting out the changes from #10005, so I guess it works as expected 👍

PR #10093, opened 2022-11-07
Add Copy with Headers feature

### LINKED ISSUE #2288: Add the ability to copy with headers
In excel it is common for the headers to be another row worth of data. In handsontable, it is more common for it to be set up as column headers.
It would be a useful enhancement to allow the copying of the data with column headers, like how SQL server allows it, either through a different key combination (e.g.  ctrl-shift-c) or with a context menu item.

budnix:
### Context

WIP

### Types of changes

### Related issue(s):
1. fixes #2288

### Affected project(s):
- [x] `handsontable`

### Checklist:

### DIFF
--- a/handsontable/src/plugins/copyPaste/copyPaste.js
+++ b/handsontable/src/plugins/copyPaste/copyPaste.js
@@ -6,6 +6,9 @@ import { rangeEach } from '../../helpers/number';
 import { sanitize } from '../../helpers/string';
 import { getSelectionText } from '../../helpers/dom/element';
 import copyItem from './contextMenuItem/copy';
+import copyColumnHeadersOnlyItem from './contextMenuItem/copyColumnHeadersOnly';
+import copyWithColumnGroupHeadersItem from './contextMenuItem/copyWithColumnGroupHeaders';
+import copyWithColumnHeadersItem from './contextMenuItem/copyWithColumnHeaders';
 import cutItem from './contextMenuItem/cut';
 import PasteEvent from './pasteEvent';
 import { createElement, destroyElement } from './focusableElement';
@@ -25,8 +28,6 @@ Hooks.getSingleton().register('afterCopy');
 export const PLUGIN_KEY = 'copyPaste';
 export const PLUGIN_PRIORITY = 80;
 const SETTING_KEYS = ['fragmentSelection'];
-const ROWS_LIMIT = Infinity;
-const COLUMNS_LIMIT = Infinity;
 const privatePool = new WeakMap();
 const META_HEAD = [
   '<meta name="generator" content="Handsontable"/>',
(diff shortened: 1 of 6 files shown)


handsontable/src/plugins/copyPaste/copyPaste.js
```diff
@@ -338,35 +383,48 @@ export class CopyPaste extends BasePlugin {
...
+    this.copyableRanges = Array.from(groupedRanges.values())
+      .filter(range => range !== null)
+      .map(({ startRow, startCol, endRow, endCol }) => ({ startRow, startCol, endRow, endCol }));
 
     this.copyableRanges = this.hot.runHooks('modifyCopyableRange', this.copyableRanges);
 
-    if (endRow !== finalEndRow || endCol !== finalEndCol) {
-      this.hot
-        .runHooks('afterCopyLimit', endRow - startRow + 1, endCol - startCol + 1, this.rowsLimit, this.columnsLimit);
+    const cellsRange = groupedRanges.set('cells');
```
jansiegel, line 419: Shouldn't this be
```suggestion
    const cellsRange = groupedRanges.get('cells');
```
?

handsontable/src/plugins/copyPaste/copyableRanges.js
```diff
@@ -0,0 +1,194 @@
...
+      col: startCol,
+    } = this.#selectedRange.getTopStartCorner();
+    const {
+      row: endRow,
+      col: endCol,
+    } = this.#selectedRange.getBottomEndCorner();
+
+    const finalEndRow = Math.min(endRow, Math.max(startRow + this.#rowsLimit() - 1, startRow));
+    const finalEndCol = Math.min(endCol, Math.max(startCol + this.#columnsLimit() - 1, startCol));
+    const isRangeTrimmed = endRow !== finalEndRow || endCol !== finalEndCol;
```
jansiegel, line 87: This section of the logic seems to be reused (with minor modifications) all across this file. Do you think this could be reorganized so there's less duplicated code?

jansiegel (changes requested):

handsontable/src/plugins/copyPaste/copyPaste.js
budnix: Thanks for spotting that! I've added missing tests for that logic (2e08acd1c298a181c74c3b27383028188ec41720).

handsontable/src/plugins/copyPaste/copyableRanges.js
budnix: Sounds good. I've moved the logic to separate method (73d536731cef8188062633a1b667171ba8db3a63);

jansiegel (approved):
👍

aninde:
I checked this new copy with each other plugin, and it works correctly. In more complicated operations such as copying after CRUD operations on the table, including hidden elements (trim rows, hidden columns) works correctly.

Issue #9645, opened 2022-07-11
[12.1.0+] Comments with fixed columns throw error and show comment box under the table when table is scrolled horizontally


AMBudnik:
### Description

Possibly related to changes in https://github.com/handsontable/handsontable/issues/4785. I've connected the issue to pinpoint the difference in the result. But it does not seem to work well in any of those versions.

Before 12.1.0 the comment box was misplaced while scrolling horizontally and eventually while scrolling right it entirely disappeared from the view. After that action, we got the following console error

```js
Uncaught TypeError: Cannot read properties of undefined (reading 'defaultView')
    at offset (handsontable.full.js:2881:33)
    at Comments.refreshEditor (handsontable.full.js:142750:44)
    at Comments.show (handsontable.full.js:142665:12)
    at Comments.showAtCell (handsontable.full.js:142688:19)
    at DisplaySwitch.<anonymous> (handsontable.full.js:142430:23)
    at fastCall (handsontable.full.js:8464:17)
    at DisplaySwitch.runLocalHooks (handsontable.full.js:14030:32)
    at DisplaySwitch.<anonymous> (handsontable.full.js:143524:18)
    at handsontable.full.js:8277:21
```

In 12.1.0 while we scroll right we can see the same misplacement but eventually, the comment box jumps after the table DIV (appears under the table) and we get a different error in the console.

```js
handsontable.full.js:148815 Uncaught TypeError: TD.getBoundingClientRect is not a function
    at Comments.refreshEditor (handsontable.full.js:148815:38)
    at Comments.show (handsontable.full.js:148728:12)
    at Comments.showAtCell (handsontable.full.js:148745:19)
    at DisplaySwitch.<anonymous> (handsontable.full.js:148488:23)
    at fastCall (handsontable.full.js:8008:17)
    at DisplaySwitch.runLocalHooks (handsontable.full.js:12936:32)
    at DisplaySwitch.<anonymous> (handsontable.full.js:149613:18)
    at handsontable.full.js:7821:21
```

### Steps to reproduce

1. Scroll the table right to hide `T` column
2. Hoover over `B2` cell to show the comment

#### result for pre 11.1.0 and older
![Screenshot 2022-07-11 at 14 13 42](https://user-images.githubusercontent.com/10757813/178261497-1b09cfd4-981d-42df-842c-a714c4068be1.png)

[...]

PR #10100, opened 2022-11-09
Comment editor will be shown properly, when comment had been added on fixed column and some scrolling was performed

wszymanski:
### Context

I investigated that we have some problems with showing a comment editor for the cell which is placed [on other overlays than the main](https://github.com/handsontable/handsontable/issues/9645#issuecomment-1307430838). Probably the best solution would be using `hot.getCell` method instead of `hot.view._wt`. Although, the current code is connected with a feature that allows showing a comment editor for hidden cells. Thus, I'm still using the Walkontable method and I've decided to not extend the task scope.

### How has this been tested?

Tested case described in the issue

### Types of changes

### Related issue(s):
1. #9645

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

### DIFF
--- a/.changelogs/9645.json
+++ b/.changelogs/9645.json
@@ -1,0 +1,7 @@
+{
+  "title": "Comment editor will be shown properly, when comment had been added on fixed column and some scrolling was performed",
+  "type": "fixed",
+  "issue": 9645,
+  "breaking": false,
+  "framework": "none"
+}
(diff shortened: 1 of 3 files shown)


jansiegel (approved):
> Probably the best solution would be using `hot.getCell` method instead of `hot.view._wt`

You're probably right, but let's leave it for another task when we're sure whether the "comments on hidden columns" feature is really a feature.

This PR seems to fix the problem described in #9645, though 👍

Issue #5322, opened 2018-08-10
All rows and columns are not rendering when Handsontable was hidden or display: none


primeO:
### Description
All rows and columns are not displaying , but columns and rows are displaying only on clicking the first row [it is not displaying column names for this ] . Any solutions for this .

### Steps to reproduce

					<script type="text/javascript">
                                            var data = //DATA GOES HERE ;

                                           
                                            var cols = [{
                                                    data: 'A1',
                                                   type: 'text'
                                          
                                                }, {
                                                    data: 'A2',
                                                   type: 'text'
                                                }, {
                                                     data: 'A3',
                                                   type: 'text'
                                                },
                                               {
                                                     data: 'A4',
                                                   type: 'text'
                                                },
                                               {
                                                     data: 'A5',
                                                   type: 'text'
                                                }

                                            ];

                                            var
                                                table = document.getElementById('losttable');

                                            var lostHot = new Handsontable(table, {
                                                data: data,
                                                columns: cols,
                                                stretchH: 'all',
                                                colHeaders: [
                                                        'A1',
                                                        'A2',
                                                        'A3',
                                                        'A4',
                                                         'A5',
                                                         'A6'
   
                                                 ],
                                                autoWrapRow: true,
                                                height: 500,
                                                rowHeights: 25,
                                                manualRowResize: true,
                                                manualColumnResize: true,
                                                contextMenu: true,
                                                sortIndicator: true,
                                                autoColumnSize: true,
                                                 columnSorting: true

                                            });

                                         
                                           setTimeout(function() {
                                                lostHot.render();
                                            }, 100);
                                            
                                        </script>

alisanguo:
I have the same question,do you have any solution?

primeO:
it should work with render() . for few tables it is working fine but for few tables it is not showing up , But if we click on the row then complete table is visible . 

iam also looking for a solution for this . and as well as search option demo

AMBudnik:
> it should work with render()

That's true. Usually, it helps. If it doesn't please share a demo where the issue can be replicable. I'll check what we can do to make it work properly.

alisanguo:
I found the same question because I use "display:none or block"  to make div display or hidden.When I set display to' block' from 'none', it happens.

rhwilburn:
[...]

PR #10130, opened 2022-11-29
Utilize an intersection observer when the table was initialized in a hidden container.

jansiegel:
### Context
This PR should fix a problem where the table was not rendered properly if it had been initialized in a hidden (with `display: none`) container.

I added an intersection observer to the `init` logic (only if the table was initialized in a hidden container), which triggers the `render` method after the container changes its `display` property to anything other than `none`.

### How has this been tested?
Added a test case.

### Types of changes

### Related issue(s):
1. #5322 

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:


budnix (changes requested):
The demo attached to the issue (https://jsfiddle.net/kq35w9zm/1/) shows that the problem is not completely fixed. The header renders correctly after clicking the "Display 'show'" button, but the sizes of the columns and rows are still incorrect. Can you check that?

jansiegel:
> The demo attached to the issue (https://jsfiddle.net/kq35w9zm/1/) shows that the problem is not completely fixed. The header renders correctly after clicking the "Display 'show'" button, but the sizes of the columns and rows are still incorrect. Can you check that?

@budnix 
I investigated the problem, and I think it's not really a problem with this solution.
That demo:
1. Sets `display: none` to a _class_ assigned to the Handsontable container
2. Changes the `display` property to `block`, targeting just the Handsontable container

This way every new (and existing) element with that class will still be invisible. That's what happens with the `GhostTable` as well, which results in all the columns having the same, default width.

When I set `display: none` to an `id` instead of a `class` it works alright: https://jsfiddle.net/js_ziggle/c6kd973x/.

That investigation, however, made me find another issue with my solution, which is changing the `display` property for the HOT container's parent element(s). 
In https://github.com/handsontable/handsontable/pull/10130/commits/ff5a7a62f05082b392acc873cda96e6b6f460f7c I modified my solution to utilize `IntersectionObserver` instead of `MutationObserver`. I also had to mock the `IntersectionObserver` for some of the unit tests in https://github.com/handsontable/handsontable/pull/10130/commits/e58e30d1e1ee8d07d9795a9014eac9781fc83fff.

After these changes, the solution should work for changing the `display` from `none` to anything else:
- For the Handsontable container, both using a stylesheet and inline styles
- For the ancestors of the Handsontable container, both using a stylesheet and inline styles

handsontable/test/__mocks__/intersectionObserverMock.js
```diff
@@ -0,0 +1,9 @@
+export class IntersectionObserverMock {
+  observe() {
+    return null;
+  }
+
+  unobserve() {
+    return null;
+  }
```
budnix, line 8:
```suggestion
  observe() {}
  unobserve() {}
```
The MDN says that the methods return `undefined`, so I guess it'd be good to follow that rules for mock.

handsontable/test/e2e/Core_init.spec.js
```diff
@@ -62,4 +62,128 @@ describe('Core_init', () => {
 
     expect(onErrorSpy).not.toHaveBeenCalled();
   });
+
+  it('should rerender the table after changing the `display` property to anything other than `none` on the root' +
+    ' element if it was initialized with `display: none` with inline styles', async() => {
+    const intialDisplayValue = spec().$container.css('display');
```
budnix, line 68:
```suggestion
    const initialDisplayValue = spec().$container.css('display');
```

handsontable/test/e2e/Core_init.spec.js
```diff
@@ -62,4 +62,128 @@ describe('Core_init', () => {
...
+    });
+
+    // Make sure the table is not visible.
+    expect(hot.rootElement.offsetParent).toEqual(null);
+
+    spec().$container.css('display', intialDisplayValue);
+
+    await sleep(100);
+
+    const $topHolderElement = $(hot.view._wt.wtOverlays.topOverlay.clone.wtTable.holder);
```
budnix, line 85:
```suggestion
    const $topHolderElement = getTopClone().find('.wtHolder');
```

handsontable/test/e2e/Core_init.spec.js
```diff
@@ -62,4 +62,128 @@ describe('Core_init', () => {
...
+    });
+
+    // Make sure the table is not visible.
+    expect(hot.rootElement.offsetParent).toEqual(null);
+
+    $testParentContainer.css('display', 'block');
+
+    await sleep(100);
+
+    const $topHolderElement = $(hot.view._wt.wtOverlays.topOverlay.clone.wtTable.holder);
```
budnix, line 116:
```suggestion
    const $topHolderElement = getTopClone().find('.wtHolder');
```

handsontable/test/e2e/Core_init.spec.js
```diff
@@ -62,4 +62,128 @@ describe('Core_init', () => {
...
+    });
+
+    // Make sure the table is not visible.
+    expect(hot.rootElement.offsetParent).toEqual(null);
+
+    spec().$container.css('display', 'block');
+
+    await sleep(100);
+
+    const $topHolderElement = $(hot.view._wt.wtOverlays.topOverlay.clone.wtTable.holder);
```
budnix, line 145:
```suggestion
    const $topHolderElement = getTopClone().find('.wtHolder');
```

handsontable/test/e2e/Core_init.spec.js
```diff
@@ -62,4 +62,128 @@ describe('Core_init', () => {
...
+    });
+
+    // Make sure the table is not visible.
+    expect(hot.rootElement.offsetParent).toEqual(null);
+
+    $testParentContainer.css('display', 'block');
+
+    await sleep(100);
+
+    const $topHolderElement = $(hot.view._wt.wtOverlays.topOverlay.clone.wtTable.holder);
```
budnix, line 179:
```suggestion
    const $topHolderElement = getTopClone().find('.wtHolder');
```

handsontable/src/core.js
```diff
@@ -1395,6 +1399,34 @@ export default function Core(rootElement, userSettings, rootInstanceSymbol = fal
     }
   }
 
+  /**
+   * Ensure that in case of the table being initialized with `display: none`, the table will get re-rendered after
+   * switching the `display` property to anything visible.
+   *
+   * @param {boolean|Array} isInit Indicator of whether the table is being initialized.
+   */
+  function correctFirstVisibleRender(isInit) {
```
budnix, line 1408: The core.js is big enough, so I'd be for moving that function to a separate file. By doing that at the same time, I'd propose making the function more generic. I'm thinking about a function that triggers a callback when the visibility of the observed element is changed. Something like:
```js
function observeVisibilityChangeOnce(elementToObserve, callback) {
  const observer = new IntersectionObserver((entries, observer) => {
    entries.forEach((entry) => {
      if (entry.isIntersecting && elementToObserve.offsetParent !== null) {
        callback();
        observer.unobserve(elementToObserve);
      }
    });
  }, {
    root: elementToObserve.ownerDocument.body
  });

  observer.observe(elementToObserve);
}
```
and then in the Core.js usage may look like this:
```js
if (!!firstRun && instance.rootElement.offsetParent === null) {
  observeVisibilityChangeOnce(instance.rootElement, () => instance.render());
}
```

budnix (changes requested):
The change seems to work, great 👏 However, I've left some suggestions that I believe can increase the code readability/maintainability.

handsontable/src/core.js
jansiegel: Good idea 👍

handsontable/src/core.js
```diff
@@ -1179,10 +1179,14 @@ export default function Core(rootElement, userSettings, rootInstanceSymbol = fal
     this.forceFullRender = true; // used when data was changed
     this.view.render();
 
+    correctFirstVisibleRender(!!firstRun);
```
budnix, line 1182:
```suggestion
    // Run the logic only if it's the table's initialization and the root element is not visible.
    if (!!firstRun && instance.rootElement.offsetParent === null) {
      observeVisibilityChangeOnce(instance.rootElement, () => instance.render());
    }
```
After you extracted the logic to the helper, I would see direct use here. I don't see reasons to create another function wrapper (`correctFirstVisibleRender`) for that, especially when it's only called once.

wrappers/react/test/_helpers.tsx
```diff
@@ -296,3 +304,13 @@ export class EditorComponent extends BaseEditorComponent<{}, {value?: any}> {
...
 }
+
+export class IntersectionObserverMock {
+  observe() {
+    return null;
+  }
+
+  unobserve() {
+    return null;
+  }
```
budnix, line 315:
```suggestion
  observe() {}
  unobserve() {}
```

wrappers/vue/test/_helpers.ts
```diff
@@ -46,3 +54,13 @@ export function createDomContainer() {
...
 }
+
+export class IntersectionObserverMock {
+  observe() {
+    return null;
+  }
+
+  unobserve() {
+    return null;
+  }
```
budnix, line 65:
```suggestion
  observe() {}
  unobserve() {}
```

wrappers/vue3/test/_helpers.ts
```diff
@@ -35,3 +43,13 @@ export function mockClientDimensions(element: HTMLElement, width: number, height
...
 }
+
+export class IntersectionObserverMock {
+  observe() {
+    return null;
+  }
+
+  unobserve() {
+    return null;
+  }
```
budnix, line 54:
```suggestion
  observe() {}
  unobserve() {}
```

handsontable/test/e2e/helpers/dom/element.spec.js
```diff
@@ -0,0 +1,25 @@
+describe('DOM helpers', () => {
+  describe('observeVisibilityChangeOnce', () => {
+    it('should recognize when an element becomes visible (once) and trigger the provided callback when that happens', async() => {
+      const observeVisibilityChangeOnce = Handsontable.dom.observeVisibilityChangeOnce;
+      const element = document.createElement('DIV');
```
budnix, line 5:
```suggestion
      const element = document.createElement('div');
```
Creating elements with capital letters reminds me of the old-school JS style. I'd change that :)

budnix (changes requested):

handsontable/test/e2e/helpers/dom/element.spec.js
```diff
@@ -0,0 +1,25 @@
+describe('DOM helpers', () => {
```
budnix, line 1: Just one last thing 🤞 IMO, it would be better to move the spec file to the helper module. Right [here](https://github.com/handsontable/handsontable/tree/develop/handsontable/src/helpers/dom/__tests__) as `element.spec.js`.

jansiegel: You're right, I somehow missed that 🤦

budnix (approved):
Perfect 👌

aninde:
Version from latest develop https://jsfiddle.net/aninde/b6r5gL3o/  work like a charm in simple case. With more complex options that change the width by `strechH`, the column and cell headers ride off. https://jsfiddle.net/aninde/4sypL5a9/ Fortunately `hot.render()` fixes this, so we don't have to worry about it. 

By the way this pr also improves the render when changing to display: 'table'.

Issue #216, opened 2012-10-31
Question about cell renderers and array data


dawidr:
When you are pasing JSON object as a source then you can speficy cell renferers for columns in following way:
columns: [
    {data: "id"},
    {data: "name", type: {renderer: yellowRenderer}},
    {data: "isActive", type: Handsontable.CheckboxCell},
    {data: "color"}
  ],

Now, when I want to pass array as an data source, how can I specify cell renerers for all columns. There won't be data: "name".

PR #10132, opened 2022-11-30
Dosc: Remove redundant HTML tabs from the "Configuration options" demos

kirszenbaum:
This PR:
- Removes redundant HTML tabs from the "Configuration options" demos (fixing [#216](https://github.com/handsontable/dev-handsontable/issues/216))

[skip changelog]


AMBudnik (approved):
Works well, all HTML tabs are removed.

PR #10131, opened 2022-11-30
Replace HandsontablePreview with ::: example-without-tabs

warpech:
### Context

This PR removes `<HandsontablePreview>` component in the docs and creates a new `::: example-without-tabs` instead.

I based `::: example-without-tabs` on `::: example` but decided to do this as a separate container, because otherwise that would require many `ifs`, making the solution more DRY but less clear.

### How has this been tested?

Locally by comparing the changes "Events and hooks" page with the production page.

I have a separate branch (candidate for https://github.com/handsontable/handsontable/pull/9833), in which I use this new container on the Demo page. It is not pushed yet, but I tested it locally. I can push it any time if needed.

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature or improvement (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Additional language file or change to the existing one (translations)

### Related issue(s):
1. https://github.com/handsontable/dev-handsontable/issues/520
2.
3.

### Affected project(s):
- [ ] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

- [x] I have reviewed the guidelines about [Contributing to Handsontable](https://github.com/handsontable/handsontable/blob/master/CONTRIBUTING.md) and I confirm that my code follows the code style of this project.
- [x] I have signed the [Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLScpMq4swMelvw3-onxC8Jl29m0fVp5hpf7d1yQVklqVjGjWGA/viewform?c=0&w=1)
- [ ] My change requires a change to the documentation.

[skip changelog]

PR #10138, opened 2022-12-02
Change the Demo to use a DIV, not an IFRAME (using ::: example-without-tabs)

warpech:
### Context

This PR changes the demo from using an `<iframe>` to inserting the demo directly to the page using `<div>`, `<script>`, `<style>` through the use of the new `::: example-without-tabs` container, which was added in PR https://github.com/handsontable/handsontable/pull/10131.

The demo is copied manually from `examples/next/docs/js` and `examples/next/docs/react` and should be manually kept in sync with the files there.

Note that React and JS variants look slightly different, which is also a problem in the `examples` and will be fixed in another issue.

### How has this been tested?

Tested the JavaScript and React variants locally using `npm run docs:start:no-cache`.

### Types of changes

### Related issue(s):
1. handsontable/dev-handsontable#34
2.
3.

### Affected project(s):
- [ ] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

[skip changelog]


jansiegel (changes requested):
The `example-checker` fails on the `/demo` page because it's meant to find the number of HOT initialization snippets in the examples' container tabs and compare it to the number of found HOT instances. In this case, there are no code tabs, but there _is_ a Handsontable instance present.

The situation is similar to the `/events-and-hooks` page, but that one passes because it's been configured to not take that instance into account:
https://github.com/handsontable/handsontable/blob/940b60959882190884de5ea9e36d466c9aa4e37d/docs/.vuepress/tools/example-checker/testCases.js#L28-L30

To get around it, you could either add a separate entry like that for the `/demo` page or modify the way the example checker looks for Handsontable instances.

I was thinking about something like this:
1. Remove the entries for `'/javascript-data-grid/events-and-hooks'` and `'/react-data-grid/events-and-hooks'` from the `testCases.js` config
2. Make the script count only the Handsontable instances that are rendered as a part of tabbed example containers (thus, having a `code` tab) with something like this:
    ```js
    const htMasterElements = document.querySelectorAll('.handsontable.ht_master');
    // Filter out Handsontable instances not being a part of a tabbed example container.
    const hotContainersInExamples = Array.from(htMasterElements).filter(
      hotContainer => hotContainer.closest('.tabs-component-panels')
    );
    ```
WDYT?

warpech:
@jansiegel I tried both approaches. I conclude that the first one makes more sense, because it tests more instances, even those coming from `::: examples-without-tabs`. In my last commit, I just added +1 for testing the instances on the demo pages.

jansiegel (approved):
👍

aninde (changes requested):
Many things mentioned in the issue have been fixed. Manual row move works; columns with headers are not shifting after scrolling and filtering.

There are still differences between this demo and the main one on the demo page, i.e., row headers do not have a checkbox, and column order is different, but I wonder if it's supposed to be a 1:1 copy.

The edit button is missing. If it is to be uneditable, it is worth hiding the bottom section with `## Edit the demo's source code`, that describes it.

Links work well.

warpech:
@krzysztofspilka approved merging it with the shortcomings listed in @aninde's comment.

aninde (approved):

PR #10165, opened 2022-12-13
visual testing - implement proof of concept into HOT repo

maciejbo-handsontable:
### Context
This is proof of concept of our visual testing - it was tested in external fork, now we would like to try in main enviroment if it meets our expectations

### How has this been tested?
We have separated environment built of forked HoT repo. There was prepared few test scenarios launched on `develop` branch and few feature branches to test our implementation of testing tool and external services comparing images.

To run the project, do the following things:
 * Install dependencies by running the `npm install` command in the root directory of the monorepo;
 * Build the Handsontable, its wrappers, and the examples for the visual tests by executing the `npm build` command in the root directory of the monorepo;
 * Run the tests by typing `npm run in visual-tests test` (in the root directory).

### Types of changes

### Related issue(s):
1. fixes https://github.com/handsontable/dev-handsontable/issues/976

### Checklist:

[skip changelog]


visual-tests/scripts/visual-tests-build.mjs
```diff
@@ -0,0 +1,72 @@
+import path from 'path';
+import execa from 'execa';
+import fse from 'fs-extra';
+
+// Main wrapper should be defined as a first one -
+// it will be used to generate golden screenshots,
+// rest of wrappers will be compared to it.
+const wrappers = ['js', 'angular-13', 'react'];
+// const wrappers = ['js', 'angular-13', 'react', 'vue'];
```
jansiegel, line 9: If I understand correct, the array containing `vue` is temporarily commented out because of the problem with building the `vue` demo. (as discussed before)

I tried to manually recreate what the script is doing, and the `start` command for `vue/demo` fails with an error indeed. I think it has something to do with the example being a part of an npm workspace.

If I'm not mistaken, the `demo` example is looking for a dependency in a wrong place, because the `npm i` command has been run _inside_ of the example directory, not in the root of its "sub-workspace".

The examples were generally designed to be installed in bulk, [using the `examples:install` scripts in the root `package.json`](https://github.com/handsontable/handsontable/blob/12.2.0/package.json#L31). It basically propagates the `install` command to all of the "sub-workspaces", [like the one defined in `docs/vue/package.json`](https://github.com/handsontable/handsontable/blob/12.2.0/examples/next/docs/vue/package.json#L5-L7).

When running `npm install` in `next/docs/vue` and _then_ running `npm run start` inside the example directory, it started alright. 
That being said, installing each of the examples individually should work as well, so this problem probably needs some investigating.

visual-tests/scripts/visual-tests-build.mjs
```diff
@@ -0,0 +1,72 @@
...
+const baseBranch = 'develop';
+const currentBranchName = (await execa.command('git rev-parse --abbrev-ref HEAD', { silent: true })).stdout;
+const dirs = {
+  init: process.cwd(),
+  examples: '../examples/next/visual-tests',
+  codeToRun: 'demo',
+};
+
+// const HOTversion = '12.2.0';
```
jansiegel, line 18: Is this commented-out line needed?

visual-tests/scripts/visual-tests-build.mjs
```diff
@@ -0,0 +1,72 @@
...
+  await execa.command('npm install', { stdout: 'inherit' });
+  // eslint-disable-next-line no-await-in-loop
+  await execa.command('npm run build', { stdout: 'inherit' });
+  // eslint-disable-next-line no-await-in-loop
+  const localhostProcess = execa.command('npm run start', {
+    detached: true,
+    stdio: 'ignore',
+    windowsHide: true });
+
+  process.chdir(dirs.init);
```
jansiegel, line 41: Just a suggestion - maybe we could utilize `execa`'s cwd option to run the scripts from different directories?
I'm afraid that changing them externally could potentially lead to some problems - for example, running a script from a wrong directory when the directory was mistakenly not changed after the previous script finished.

visual-tests/scripts/visual-tests-build.mjs
```diff
@@ -0,0 +1,72 @@
...
+// If we are on a base branch, we do not want to run all of tests
+// and make screenshots for all of wrappers.
+// On base branch we need to create golden screenshots from main wrapper only,
+// which is declared as a first position in wrappers array.
+
+for (let i = 0, maxi = (currentBranchName === baseBranch ? 1 : wrappers.length); i < maxi; ++i) {
+  process.chdir(`${wrappers[i]}/${dirs.codeToRun}`);
+
+  // eslint-disable-next-line no-await-in-loop
+  await execa.command('npm install', { stdout: 'inherit' });
```
jansiegel, line 32: Just a suggestion - if the script installs and builds multiple examples, we could install them in bulk using the `install` command of their "sub-workspaces".

For example, instead of running `npm install` for each of the examples in `next/visual-tests/vue`, we could run it once, just in the `next/visual-tests/vue` directory. All the shared dependencies will be installed once, and they'll be reused between the examples.

visual-tests/scripts/visual-tests-build.mjs
```diff
@@ -0,0 +1,72 @@
...
+
+  for (let i = 1; i < wrappers.length; ++i) {
+    copies.push(fse.copySync(
+      path.resolve(`./tests/screenshots/${wrappers[0]}`),
+      path.resolve(`./tests/screenshots/${wrappers[i]}`),
+      { overwrite: true })
+    );
+  }
+
+  await Promise.all(copies);
```
jansiegel, line 71: Correct me if I'm wrong, but `fse.copySync` doesn't return a promise, so `await Promise.all(copies)` does nothing in this context? Or maybe I'm missing something?

visual-tests/.gitignore
```diff
@@ -0,0 +1,7 @@
+node_modules/
+/test-results/
+/playwright-report/
+/playwright/.cache/
+/tests/screenshots/
+/tests/*/snapshots
+/tests/*/screenshots
```
jansiegel, line 7: I don't know what the structure of the `tests` directory is gonna be, so just checking - is `*` sufficient in this context? Maybe we should use `**` to make sure all the more nested directories will be ignored as well?

visual-tests/playwright.config.js
```diff
@@ -0,0 +1,120 @@
...
+  ],
+
+  /* Folder for test artifacts such as screenshots, videos, traces, etc. */
+  // outputDir: 'test-results/',
+
+  /* Run your local dev server before starting the tests */
+  // webServer: {
+  //   command: 'npm run start',
+  //   port: 3000,
+  // },
```
jansiegel, line 117: There's a _lot_ of commented-out stuff here. Is this needed?
Additionally, some of it is commented out using `//` and some with `/* */`, is it done on purpose?

visual-tests/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "playwright",
```
jansiegel, line 2: `playwright` is probably not the best name for this package.

visual-tests/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "version": "1.0.0",
```
jansiegel, line 3: Is this package going to be versioned with each release? 
If so, this process needs to be included in the release script. If not, maybe we could change the version to `0.0.0`, [similarly to `docs`](https://github.com/handsontable/handsontable/blob/12.2.0/docs/package.json#L4) to make it more obvious.

visual-tests/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": "node ./scripts/visual-tests-build.mjs",
+    "docker:run": "docker pull mcr.microsoft.com/playwright:v1.28.1-focal && docker run -v $(INIT_CWD)/..:/tests/ -w /tests/visual-test --rm -it mcr.microsoft.com/playwright:v1.28.1-focal /bin/bash",
+    "docker:run:win": "docker pull mcr.microsoft.com/playwright:v1.28.1-focal && docker run -v %INIT_CWD%/..:/tests/ -w /tests/visual-test --rm -it mcr.microsoft.com/playwright:v1.28.1-focal /bin/bash",
+    "upload": "node ./scripts/visual-tests-upload.mjs",
+    "all": "node ./scripts/visual-tests-build.mjs && node ./scripts/visual-tests-upload.mjs"
```
jansiegel, line 11: I'm not sure if `all` is the best name for this script.

1. It could be confused with [the `all` script from the monorepo's root.](https://github.com/handsontable/handsontable/blob/12.2.0/package.json#L16)
2. `all` doesn't mean much in this context. Maybe something more descriptive would fit this better?

visual-tests/package.json
```diff
@@ -0,0 +1,26 @@
...
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
jansiegel, line 15: Just checking - is the license entry added here on purpose, or is it a leftover of the `package.json` auto-generating script?

jansiegel (changes requested):
Apart from the suggestions left in the comments:

As discussed before, I wasn't able to actually test these scripts because the tokens used for the external services were not yet configured in the repo.

But I think the testing of the examples from the `next` directory will not work correctly with this setup.

The examples stored in the `next` directory are meant to be used with the local builds of Handsontable and the wrappers. 
In order for it to work:
- we use (nested) npm workspaces to share the installed dependencies between the examples and the codebase
- there are some scripts in place to ensure the `handsontable` and `@handsontable/[framework]` packages are being linked to the current, local builds

The script for the visual tests doesn't use the `install` capabilities of the npm workspaces - instead it installs each of the examples manually, which results in the `handsontable` and the wrappers _not_ being replaced with symlinks.

This could be corrected either by:
- using the `examples:install` script in the root of the monorepo, as it handles the dependency linking
- running the dependency-linking scripts manually after installing the dependencies in the "sub workspaces" (described in one of the review comments), for example:
    Running:
    ```
    node ./scripts/link-packages.mjs --f vue --examples-version next
    ```
    after running
    ```
    npm i
    ```
    in `examples/next/docs/vue`
    
    should ensure that the `vue` examples are using the _local_ builds. 
    The local builds would obviously need to be created beforehand.
    
Additionally, we would need to double-check if all the examples-handling scripts work correctly with `examples/next/visual-tests`, as they were only used with `examples/next/docs` until now.

maciejbo-handsontable:
> Apart from the suggestions left in the comments:
> (...)

hI! I saw all of your points - I will take care about it soon, but you should be informed that few of them was implemented intentionally and for a while may remain without response - I need to discuss about some things and make common decision with rest of team (especially points about when and how workflow should be launched)

.github/workflows/visual-tests.yml
```diff
@@ -0,0 +1,39 @@
+name: Visual tests
+
+on:
+  workflow_dispatch:
+  push:
```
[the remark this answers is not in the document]
maciejbo-handsontable: Hi! This part should be yet discussed with rest of them - we have to make common decision when and how these tests should be launched

.github/workflows/visual-tests.yml
```diff
@@ -0,0 +1,39 @@
...
+env:
+  NODE_VERSION: 16
+
+jobs:
+  build:
+    name: Build and upload visual tests
+    runs-on: ubuntu-latest
+    container:
+      image:  mcr.microsoft.com/playwright:v1.28.1-focal
+    # if: ${{ github.event.workflow_run.conclusion == 'success' || github.event_name == 'workflow_dispatch' }}
```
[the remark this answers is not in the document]
maciejbo-handsontable: I left it just in case, it is related to the previous point and depends on how and when tests will be launched - we will need this line if `visual tests` worklow will be launched by another workflow (like we have it in `prerelease` already, which is launched only if `tests` passed)

visual-tests/imports/helpers.js
```diff
@@ -0,0 +1,71 @@
+export const helpers = {
+  defaultHOTVersion: '12.2.0',
```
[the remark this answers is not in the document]
maciejbo-handsontable: This line is not necessary at this moment, we do not use information about version anymore and we always test version from the `next `folder` launched on localhost. This information was used on the beginning, when we used exernal URL to test:
`https://examples.handsontable.com/examples/${this.hotVersion}/docs/${this.hotWrapper}/demo/index.html`;

visual-tests/imports/helpers.js
```diff
@@ -0,0 +1,71 @@
...
+  defaultHOTWrapper: 'js',
+
+  screenshotsDirectory: './tests/screenshots',
+  screenshotsExtension: 'png',
+
+  selectors: {
+    mainTable: 'div.handsontable.htRowHeaders.htColumnHeaders',
+    mainTableBody: '> .ht_master.handsontable table tbody',
+    mainTableHead: '> .ht_clone_top.handsontable table thead',
+    mainTableFirstColumn: '> .ht_clone_inline_start.ht_clone_left.handsontable table tbody',
```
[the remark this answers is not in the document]
maciejbo-handsontable: `mainTable points to the main table container`
Sure, I changed the class of `mainTable` already, but rest of points was done intentionally. Correct me if I mistaken something  ;) 

`mainTableBody points to the tbody element on the master overlay`:

that's true - it's because we are looking for cells and their content in `master table`, they are working correctly, are available for the testing tool and any weird solution was not needed, but here:

`mainTableHead points to the thead element, but on the top overlay (which makes sense if we're targeting the topmost elements)`

I did it in this way, cause in `thead` we are looking for buttons or columns headers which are not available for testing tool if you look for them in `master table` - e.g. if you want to click `button.changeType`, testing tool will return error `element not visible`, cause it's overlayed by `clone_top table` - question is: if this clone always exists? If no, then there is other way to workaround it, but requires one more thing from tester side, or I can rebuild helpers to detect if clone exists.

`mainTableFirstColumn points the the tbody element (so maybe the "column" in the property name can be misleading?) on the left overlay, which might not work if there are no frozen columns`

I wanted to give tester easy access to the left column without any unnecessary moves from their side and from tester point of view this returns indeed left column of main table - they can easy search for given cell in it. But not sure about this part with `frozen columns` - would you tell me more?

visual-tests/playwright.config.js
maciejbo-handsontable: Hey, most of it was here from the same beginning in testing tool configuration. I left it cause wasn't sure which browsers and devices finally will remain here. Cleaned.

visual-tests/package.json
maciejbo-handsontable: probably you are right. Can not confirm currently if it was my idea, or it was added automatically (`playright` is a name of our testing tool). Which name would be better?

visual-tests/package.json
maciejbo-handsontable: it's leftover

visual-tests/imports/helpers.js
```diff
@@ -0,0 +1,71 @@
...
+
+  cssPath(file) {
+    return `./tests-css/${file}`;
+  },
+
+  init(workerInfo, process) {
+    this.hotVersion = process.env.HOT_VERSION || this.defaultHOTVersion;
+    this.hotWrapper = process.env.HOT_WRAPPER || this.defaultHOTWrapper;
+    // eslint-disable-next-line max-len
+    // this.testURL = `https://examples.handsontable.com/examples/${this.hotVersion}/docs/${this.hotWrapper}/demo/index.html`;
```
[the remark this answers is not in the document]
maciejbo-handsontable: not anymore - in first version we were testing external URL, now we moved to localhost and probably it won't change again. Removed

visual-tests/imports/helpers.js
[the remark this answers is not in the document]
jansiegel: In general, the additional overlays are used for:
- [Row headers](https://handsontable.com/docs/javascript-data-grid/row-header/) and [column headers](https://handsontable.com/docs/javascript-data-grid/column-header/)
- Fixed/frozen [rows](https://handsontable.com/docs/javascript-data-grid/row-freezing/) (also on the [bottom of the table](https://handsontable.com/docs/javascript-data-grid/api/options/#fixedrowsbottom)) and [columns](https://handsontable.com/docs/javascript-data-grid/column-freezing/)
- The "corner" top-left cell on the intersection of row and column headers

If you're trying to target the _topmost_ DOM element representing the row/column headers, it'll always be attached to the `top` overlay (column header) / `left/start` overlay (row header).
If you're trying to target the _topmost_ DOM element representing a _cell_ things get more complicated. It can be stored under either the `master` overlay or the `top`/`left/start` overlay, depending on whether they fall into the fixed/frozen rows/columns range.

For example:
```js
  // (...)
  fixedRowsTop: 2,
  fixedColumnsStart: 2,
});

// Cells (0, 0), (1, 1), (0, 5), (5, 1) etc have their duplicates on their own respective additional overlays
// Cells (2, 2), (5, 5), (3, 5), (5, 3) etc don't
```

visual-tests/package.json
jansiegel: Something more descriptive, I think.
Currently, the internal packages are named:
- `handsontable-examples-internal`
- `handsontable-documentation`

So maybe something like `handsontable-visual-tests`? And while you're editing the file - it'd probably be a good idea to add a `"private": true,` entry to not publish the package by accident.

visual-tests/imports/helpers.js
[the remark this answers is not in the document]
maciejbo-handsontable: well, I had said that I changed this selector `mainTable: 'div.handsontable.htRowHeaders.htColumnHeaders',` already just to `mainTable: 'div.handsontable' and it was true, but after that I tested it and testing tool has problems, cause in pages with examples there is more than one element with this class (and probably that was reason why I did it in current way at the same beginning).

I tried to find just this `.handsontable` element in root or somewhere in common place on page, but unfortunatelly each of our examples has different DOM structure and it is not possible to do that in this way. We can do it somehow by looking for elements by `querySelectorAll` for example, but I'm not sure if it won't bring problems in future.

My proposition is just to leave it in current form for a while -> unify examples DOM in next step -> change this selector in helpers without touching tests itself.

What do you think?

visual-tests/scripts/visual-tests-build.mjs
maciejbo-handsontable: not anymore - it was needed to test external URLs and looks like we won't back to this solution. Removed

visual-tests/.gitignore
maciejbo-handsontable: Actually it is only because I had few versions of paths on local environment (before each test had own directory with subdirectory for screenshots, but it was unnecessary complicated) and wanted avoid to push unnecessary screenshots to repo. These two lines are not necessary anymore.

visual-tests/package.json
maciejbo-handsontable: sounds perfect. Pushed

visual-tests/package.json
maciejbo-handsontable: `Is this package going to be versioned with each release?`
probably no. Version changed to proposed one.

visual-tests/imports/helpers.js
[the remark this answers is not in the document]
jansiegel: > in pages with examples there is more than one element with this class

`div.handsontable.htRowHeaders.htColumnHeaders` could easily return more elements as well, either on a page with multiple Handsontable instances or on an instance with Handsontable-based editors being open.

I don't think we should make the testing tool work only with our current, single test case, but rather make it as generic as possible to make it as future-proof as we possibly can.

Generally speaking, using CSS classes to target one, specific DOM element is not the best approach. Maybe adding `id`s to the Handsontable instances would be a better idea? For example, we could assume each test case would contain Handsontable instances `id`'d as `#example1`, `#example2`, `#example3` etc - with `#example1` being the main, default one?

jansiegel (changes requested):
@maciejbo-handsontable 
I haven't looked through all the changes since my last review yet, but I noticed these things:
- https://github.com/handsontable/handsontable/pull/10165#discussion_r1058142185
  This comment is still applicable, for example:
  - `mainTable: 'div.handsontable.htRowHeaders.htColumnHeaders'` can still easily target more than one element (as described in the original comment).
  
    https://jsfiddle.net/js_ziggle/e0dgwap8/ - start editing a cell with handsontable-based editor and run `querySelectorAll` with the above selector.
    
  - `mainTableFirstColumn: '> .ht_clone_inline_start.ht_clone_left.handsontable table tbody'` in the vast majority of instances doesn't target "first column" - it targets the "left overlay table" (as described in the original comment)
  
- https://github.com/handsontable/handsontable/pull/10165#discussion_r1056297438
You didn't respond to this comment in any way, could you clarify if you think it's a good or a bad idea? 

- https://github.com/handsontable/handsontable/pull/10165#discussion_r1057565222
If it's a leftover, maybe we should change it to the actual license?

- https://github.com/handsontable/handsontable/pull/10165#pullrequestreview-1228873761
**This one is the most important issue** - if I'm seeing this correctly, the script is still not testing the `next` build with the _local builds_. It just installs whatever's in the examples' `package.json` files.

Additionally:
I tried running the `npm run test` script locally after changing the `baseBranch` to the branch I'm on, but I get the following error:
```
> handsontable-visual-tests@0.0.0 test
> node ./scripts/visual-tests-launcher.mjs build

docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.
See 'docker run --help'.
/Users/jansiegel/workspace/cr-handsontable/visual-tests/node_modules/execa/lib/error.js:60
		error = new Error(message);
		        ^

Error: Command failed with exit code 125: docker run --name vtests-container --env CURRENT_BRANCH=feature/dev-issue-572 -v /Users/jansiegel/workspace/cr-handsontable/visual-tests/../:/vtests/ -w /vtests/visual-tests --rm vtests
    at makeError (/Users/jansiegel/workspace/cr-handsontable/visual-tests/node_modules/execa/lib/error.js:60:11)
    at handlePromise (/Users/jansiegel/workspace/cr-handsontable/visual-tests/node_modules/execa/index.js:118:26)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async file:///Users/jansiegel/workspace/cr-handsontable/visual-tests/scripts/visual-tests-launcher.mjs:32:5 {
  shortMessage: 'Command failed with exit code 125: docker run --name vtests-container --env CURRENT_BRANCH=feature/dev-issue-572 -v /Users/jansiegel/workspace/cr-handsontable/visual-tests/../:/vtests/ -w /vtests/visual-tests --rm vtests',
  command: 'docker run --name vtests-container --env CURRENT_BRANCH=feature/dev-issue-572 -v /Users/jansiegel/workspace/cr-handsontable/visual-tests/../:/vtests/ -w /vtests/visual-tests --rm vtests',
  escapedCommand: 'docker run --name vtests-container --env "CURRENT_BRANCH=feature/dev-issue-572" -v "/Users/jansiegel/workspace/cr-handsontable/visual-tests/../:/vtests/" -w "/vtests/visual-tests" --rm vtests',
  exitCode: 125,
  signal: undefined,
  signalDescription: undefined,
  stdout: undefined,
  stderr: undefined,
  failed: true,
  timedOut: false,
  isCanceled: false,
  killed: false
}
```
Am I doing something wrong?

budnix (changes requested):
I am in the middle of the review, and I think I found enough points to finish the first part as there are more issues and doubts to report. Besides the suggestions that I've left in the source code, here are the points that caught my attention:
 * I think you copied/pasted the wrong demos for the Angular version. Now, I've seen demos for Angular 13, but I guess we want to test the latest Angular (https://github.com/handsontable/handsontable/tree/develop/examples/next/docs/angular);
 * In the test files, I saw a lot of duplicated code that could be improved (I think I left the comment in the right lines). Probably for the next project iteration, it would be nice to write more simplified helpers, for example replace `await page.screenshot({ path: helpers.screenshotPath() });` to `await page.takeScreenshot();` or so;
 * There is a lack of consistency using the `yargs` module and `process.argv` prop in getting the args from CLI. Some of the scripts use the module some of them do not. IMO we should pick one and use it in all scripts;
 * I haven't seen any linter implemented in the project. For code quality and maintainability, it would be nice to have one;
 * I think that [there](https://github.com/handsontable/handsontable/pull/10165/files#diff-7aa8f0e402f1a52f875347e12f59b9272b73b4f1d968bb2152c95be34c97c9f8R62) is a bug. If I understand correctly, the screenshots are named by the next number. But what if someone edits the test case and reorders the `page.screenshot()` calls (as they can occur more than once)? I guess the screenshots won't be matched with the base. This can be problematic when there will be more tests. Recommended solution: naming screenshots with test name, hash or so?

visual-tests/scripts/visual-tests-build.mjs
```diff
@@ -0,0 +1,82 @@
...
+
+// If we are on a base branch, we do not want to run all of tests
+// and make screenshots for all of wrappers.
+// On base branch we need to create golden screenshots from main wrapper only,
+// which is declared as a first position in wrappers array.
+
+for (let i = 0, maxi = (currentBranch === baseBranch ? 1 : wrappers.length); i < maxi; ++i) {
+  process.chdir(`${wrappers[i]}`);
+  // eslint-disable-next-line no-await-in-loop
+  await execa.command('npm install', { stdout: 'inherit' });
```
budnix, line 28:
```suggestion
  // eslint-disable-next-line no-await-in-loop
  await execa.command('npm install', { stdout: 'inherit', cwd: wrappers[i] });
```
IMO using `chdir` in that case is not the best idea. Changing the directory multiple times within the same script can easily lead to errors. As the changing directory needs to be tracked. That's why I'd propose changing all  `await execa.command(...` calls to use `cwd` property. Thanks to that you can remove all `chdir` calls from the code.

visual-tests/imports/helpers.js
```diff
@@ -0,0 +1,67 @@
+export const helpers = {
+  defaultHOTWrapper: 'js',
+
+  screenshotsDirectory: './tests/screenshots',
```
budnix, line 4:
```suggestion
  screenshotsDirectory: './screenshots',
```
Can I propose changing the directory where the screenshots are saved? Currently, they are saved in the same place as the test files. I think the better place for that would be the root of the project. There are already created other non-comittable dirctories as `test-results` or `playwright-report`.

visual-tests/imports/helpers.js
```diff
@@ -0,0 +1,67 @@
...
+  findCell(options = { row: 1, cell: 1, cellType: 'td' }) {
+    return `> tr:nth-child(${options.row}) > ${options.cellType}:nth-child(${options.cell})`;
+  },
+
+  findDropdownMenuExpander(options = { col: 1 }) {
+    return `${this.selectors.mainTableHead} > tr > th:nth-child(${options.col + 1}) button.changeType`;
+  },
+
+  testTitle(filename) {
+    const title = filename.split('\\').slice(-1)[0].split('.spec.js')[0].split('-').join(' ');
```
budnix, line 56: I'd recommend using the `path.basename(filename)` function from the `path` module to get the filename only (and don't use a hardcoded path separator). And another related question: Why the test titles are named with a full absolute path with spaces? 🤔 For example, for `select-few-cells-by-mouse.spec.js` the test is titled as `/Users/budnix/Documents/Projects/handsontable/visual tests/tests/select few cells by mouse`.

visual-tests/imports/helpers.js
```diff
@@ -0,0 +1,67 @@
...
+    dynamicDataFreeze: 'dynamic-data-freeze.css'
+  },
+
+  expectedPageTitle: /Handsontable for .* example/,
+
+  cssPath(file) {
+    return `./tests-css/${file}`;
+  },
+
+  init(workerInfo, process) {
```
budnix, line 28:
```suggestion
  init(workerInfo) {
```
The `process` variable is already available globally in the Node. I don't see reasons to pass it in tests. Or am I missing something?

visual-tests/tests/change-rows-order.spec.js
```diff
@@ -0,0 +1,45 @@
...
+
+/* ======= */
+
+const stylesToAdd = [
+  helpers.cssFiles.cookieInfo,
+  helpers.cssFiles.dynamicDataFreeze
+];
+
+test(helpers.testTitle(__filename), async({ page }, workerInfo) => {
+  helpers.init(workerInfo, process);
```
budnix, line 13:
```suggestion
  helpers.init(workerInfo);
```
[#ref](https://github.com/handsontable/handsontable/pull/10165/files#r1068189649)

visual-tests/imports/helpers.js
```diff
@@ -0,0 +1,67 @@
...
+    this.screenshotDirName = this.workerInfo.titlePath[0].split('.spec.js')[0];
+    this.browser = this.workerInfo.project.name;
+
+    if (this.browser === 'webkit') {
+      this.browser = 'safari';
+    }
+
+    if (this.browser === 'chromium') {
+      this.browser = 'chrome';
+    }
```
budnix, line 44:
```suggestion
```
IMO I'd remove that lines, as I think it's better to name screenshots by the origin engine name, not by the browser. `chromium` engine, can differ from `chrome` and `webkit` does not necessarily mean Safari. This can be confusing while debugging or trying to reproduce the potential test fails.

handsontable/src/index.js
```diff
@@ -167,3 +167,5 @@ export {
   CellRange,
 };
 export default Handsontable;
+
+/* test */
```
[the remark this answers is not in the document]
maciejbo-handsontable: yes - but you should be informed that it was temporary change to force launching workflow, that was our common decision on one of meetings during your holidays, as far as I remember.

visual-tests/tests-css/change-cell-text-color.css
```diff
@@ -0,0 +1,3 @@
+table.htCore tr td {
```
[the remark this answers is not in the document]
maciejbo-handsontable: actually I think we can just remove entire file on this step - it was for test purpose to check how Argos will inform us about changes like this.

visual-tests/tests-css/change-cells-background.css
```diff
@@ -0,0 +1,3 @@
+table.htCore tr td {
```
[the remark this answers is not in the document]
maciejbo-handsontable: And like before - actually I think we can just remove entire file on this step - it was for test purpose to check how Argos will inform us about changes like this.

visual-tests/tests/.empty-test-template.js
```diff
@@ -0,0 +1,29 @@
+import { helpers } from '../imports/helpers';
+
+const { test, expect } = require('@playwright/test');
+
+/* ======= */
```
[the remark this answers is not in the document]
maciejbo-handsontable: I did it because these fiiles will be used by less technical people also and wanted to make things easier for them, like "do not worry, you do not have to even look here". I removed first separator (line 5), what do you think about second one, which marks "end of configuration, beginning of test" (line 23 in all of files)?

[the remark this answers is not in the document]
budnix: I’d remove all these comments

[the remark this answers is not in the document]
maciejbo-handsontable: done - removed from all of files.

visual-tests/tests/.empty-test-template.js
```diff
@@ -0,0 +1,29 @@
+import { helpers } from '../imports/helpers';
+
+const { test, expect } = require('@playwright/test');
```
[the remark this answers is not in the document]
maciejbo-handsontable: `I think it's not a good idea to mix the import (ESM) and require (CommonJS) syntax in the same file. If there is a chance to change that, I'd do that.`

yes - done.

`Moreover, is there any chance to use .mjs files as test files?`

probably yes, but it requires little bit more changes. If it's necessary I will have a look after rest of points.

visual-tests/tests/expand-dropdown-menu.spec.js
```diff
@@ -0,0 +1,37 @@
...
+  /* ==== */
+
+  const changeTypeButton = table.locator(helpers.findDropdownMenuExpander({ col: 2 }));
+
+  await changeTypeButton.click();
+  const dropdownMenu = page.locator(helpers.selectors.dropdownMenu);
+
+  await page.screenshot({ path: helpers.screenshotPath() });
+  // You can take a screenshot of any element, not only entire page, e.g:
+  // await dropdownMenu.screenshot({ path: helpers.screenshotPath() });
```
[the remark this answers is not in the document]
maciejbo-handsontable: Actually yes, since it is in docs it is not necessary here. Removed.

.github/workflows/visual-tests.yml
```diff
@@ -0,0 +1,41 @@
...
+
+env:
+  NODE_VERSION: 16
+
+jobs:
+  build:
+    name: Build and upload visual tests
+    runs-on: ubuntu-latest
+    container:
+      image:  mcr.microsoft.com/playwright:v1.28.1-focal
```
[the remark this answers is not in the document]
maciejbo-handsontable: done.

visual-tests/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "handsontable-visual-tests",
+  "private": true,
+  "version": "0.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "docker:build": "cd .. && docker build --no-cache --tag vtests -f visual-tests/Dockerfile .",
```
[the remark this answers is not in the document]
maciejbo-handsontable: thanks, it was leftover from previous version.

visual-tests/scripts/visual-tests-launcher.mjs
```diff
@@ -0,0 +1,41 @@
+import execa from 'execa';
+import yargs from 'yargs';
+import { hideBin } from 'yargs/helpers';
+
+let currentBranch;
+const argv = yargs(hideBin(process.argv)).argv;
+const pathToMount = `${process.cwd().split('\\').join('/')}/../`;
+const isCICD = process?.argv?.includes('CICD');
```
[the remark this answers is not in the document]
maciejbo-handsontable: Renamed everywhere.

visual-tests/scripts/visual-tests-build.mjs
```diff
@@ -0,0 +1,82 @@
+/* eslint-disable no-console */
```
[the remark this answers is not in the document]
maciejbo-handsontable: `Can you describe the script's purpose and what it does? Just like we have in our other scripts, for example [release.mjs](https://github.com/handsontable/handsontable/blob/develop/scripts/release.mjs#L1-L6)`

Done.

`IMO the file name needs to be clarified. Right now, It says that something will be build, but in fact, it does more. It builds the examples, tests them, and saves the results as screenshots. What do you think to rename the file to just test.mjs?`

My idea was name it `build` cause it `builds` package of screenshots to `upload` in next file.

visual-tests/scripts/visual-tests-launcher.mjs
```diff
@@ -0,0 +1,41 @@
+import execa from 'execa';
```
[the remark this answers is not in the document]
maciejbo-handsontable: `Can you describe the script's purpose and what it does? Just like we have in our other scripts, for example [release.mjs](https://github.com/handsontable/handsontable/blob/develop/scripts/release.mjs#L1-L6)`

Done.

`I think that the file prefix visual-tests- in the file name is redundant. What do you think to rename the file to just launcher.mjs? The file already sits in the visual-tests folder, so why duplicate names?`

Actually it's leftover from first versions, later main reason to leave it was that until this moment I wasn't sure if there will be not wish to move all of it to parent directory and then these names would be needed. Now indeed this part of name is not necessary. Renamed

visual-tests/scripts/visual-tests-upload.mjs
```diff
@@ -0,0 +1,19 @@
+/* eslint-disable no-console */
```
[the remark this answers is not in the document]
maciejbo-handsontable: Done.

visual-tests/imports/helpers.js
maciejbo-handsontable: `Why the test titles are named with a full absolute path with spaces?`

Not so perfect splitting of paths, outdated since your proposition with `path` has been implemented:

`I'd recommend using the path.basename(filename) function from the path module to get the filename only (and don't use a hardcoded path separator)`

visual-tests/imports/helpers.js
maciejbo-handsontable: Sure - changed.

visual-tests/imports/helpers.js
maciejbo-handsontable: For me this naming is not necessary, but there was two reasons of it:
- my idea was it would be just more obvious for less technical people who not always may know name `webkit`
- one of tested by us services comparing screenshots was detecting browser by filenames and putting them to given categories. I'm not sure after this time, but as far as I remember Webkit was not displayed correctly - it was categorized as Edge.

Since 2 reason is outdated (cause of few obstacles we decided to use another service) I think I can recover previous names.

visual-tests/tests/select-few-cells-by-mouse.spec.js
```diff
@@ -0,0 +1,39 @@
...
+  await page.goto(helpers.testURL);
+  await expect(page).toHaveTitle(helpers.expectedPageTitle);
+  stylesToAdd.forEach(item => page.addStyleTag({ path: helpers.cssPath(item) }));
+
+  const table = page.locator(helpers.selectors.mainTable);
+
+  await table.waitFor();
+  helpers.mainTableFirstColumn = table.locator(helpers.selectors.mainTableFirstColumn);
+  helpers.tbody = table.locator(helpers.selectors.mainTableBody);
+  helpers.thead = table.locator(helpers.selectors.mainTableHead);
```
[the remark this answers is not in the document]
maciejbo-handsontable: That's true, actually at the beginning I had some attempt to improve this part, but left it for a while as a not showsttoper.

visual-tests/tests/mouse-wheel.spec.js
```diff
@@ -0,0 +1,39 @@
...
+  const table = page.locator(helpers.selectors.mainTable);
+
+  await table.waitFor();
+  helpers.mainTableFirstColumn = table.locator(helpers.selectors.mainTableFirstColumn);
+  helpers.tbody = table.locator(helpers.selectors.mainTableBody);
+  helpers.thead = table.locator(helpers.selectors.mainTableHead);
+
+  /* ==== */
+
+  const tbodyCoordinates = await helpers.tbody.boundingBox(); // || { x: 0, y: 0, width: 0, height: 0 };
```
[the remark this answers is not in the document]
maciejbo-handsontable: Done.

visual-tests/imports/helpers.js
maciejbo-handsontable: >     * [#10165 (comment)](https://github.com/handsontable/handsontable/pull/10165#discussion_r1058142185)
>       This comment is still applicable, for example:
>       
>       * `mainTable: 'div.handsontable.htRowHeaders.htColumnHeaders'` can still easily target more than one element (as described in the original comment).
>         https://jsfiddle.net/js_ziggle/e0dgwap8/ - start editing a cell with handsontable-based editor and run `querySelectorAll` with the above selector.

hey, this thing required little more changes, not only in visual testing, so was done just after your comment, but in different PR - it is available here: https://github.com/handsontable/handsontable/pull/10193

Was pretty sure that I told you about it, but now I see that link is missing.

>       * `mainTableFirstColumn: '> .ht_clone_inline_start.ht_clone_left.handsontable table tbody'` in the vast majority of instances doesn't target "first column" - it targets the "left overlay table" (as described in the original comment)

About last point - as far I remember we wasn't sure how to make it flexible, but actually I'm not sure if in our case it is necessary to change this part, cause there is a chance that described case will never happen ( as far as I know we will be testing always this and the same configuration of HoT ).

>     * [#10165 (comment)](https://github.com/handsontable/handsontable/pull/10165#discussion_r1056297438)
>       You didn't respond to this comment in any way, could you clarify if you think it's a good or a bad idea?

My bad, looks like I misundertood this point and thought that it is automatically outdated after improving another thing. Will be improved this week.

>     * [#10165 (comment)](https://github.com/handsontable/handsontable/pull/10165#discussion_r1057565222)
>       If it's a leftover, maybe we should change it to the actual license?

Sure - should I just copy it from other file?

>       **This one is the most important issue** - if I'm seeing this correctly, the script is still not testing the `next` build with the _local builds_. It just installs whatever's in the examples' `package.json` files.

Well, this is the point I had mentioned on last meeting that still can be broken and requiring support - and looks like it is. I do not know process of building examples well and looks like my changes was not enough, but I asked about some introducing and then everything will be clear.

> @maciejbo-handsontable I haven't looked through all the changes since my last review yet, but I noticed these things:
> 
>     * [#10165 (comment)](https://github.com/handsontable/handsontable/pull/10165#discussion_r1058142185)
>       This comment is still applicable, for example:
>       
>       * `mainTable: 'div.handsontable.htRowHeaders.htColumnHeaders'` can still easily target more than one element (as described in the original comment).
>         https://jsfiddle.net/js_ziggle/e0dgwap8/ - start editing a cell with handsontable-based editor and run `querySelectorAll` with the above selector.
>       * `mainTableFirstColumn: '> .ht_clone_inline_start.ht_clone_left.handsontable table tbody'` in the vast majority of instances doesn't target "first column" - it targets the "left overlay table" (as described in the original comment)
> 
>     * [#10165 (comment)](https://github.com/handsontable/handsontable/pull/10165#discussion_r1056297438)
>       You didn't respond to this comment in any way, could you clarify if you think it's a good or a bad idea?
> 
>     * [#10165 (comment)](https://github.com/handsontable/handsontable/pull/10165#discussion_r1057565222)
>       If it's a leftover, maybe we should change it to the actual license?
> 
>     * [#10165 (review)](https://github.com/handsontable/handsontable/pull/10165#pullrequestreview-1228873761)
>       **This one is the most important issue** - if I'm seeing this correctly, the script is still not testing the `next` build with the _local builds_. It just installs whatever's in the examples' `package.json` files.
> 
> 
> Additionally: I tried running the `npm run test` script locally after changing the `baseBranch` to the branch I'm on, but I get the following error:
> 
> ```
> > handsontable-visual-tests@0.0.0 test
> > node ./scripts/visual-tests-launcher.mjs build
> 
> Unable to find image 'vtests:latest' locally
> docker: Error response from daemon: pull access denied for vtests, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
> See 'docker run --help'.
> /Users/jansiegel/workspace/cr-handsontable/visual-tests/node_modules/execa/lib/error.js:60
> 		error = new Error(message);
> 		        ^
> 
> Error: Command failed with exit code 125: docker run --name vtests-container --env CURRENT_BRANCH=feature/dev-issue-572 -v /Users/jansiegel/workspace/cr-handsontable/visual-tests/../:/vtests/ -w /vtests/visual-tests --rm vtests
>     at makeError (/Users/jansiegel/workspace/cr-handsontable/visual-tests/node_modules/execa/lib/error.js:60:11)
>     at handlePromise (/Users/jansiegel/workspace/cr-handsontable/visual-tests/node_modules/execa/index.js:118:26)
>     at processTicksAndRejections (node:internal/process/task_queues:96:5)
>     at async file:///Users/jansiegel/workspace/cr-handsontable/visual-tests/scripts/visual-tests-launcher.mjs:32:5 {
>   shortMessage: 'Command failed with exit code 125: docker run --name vtests-container --env CURRENT_BRANCH=feature/dev-issue-572 -v /Users/jansiegel/workspace/cr-handsontable/visual-tests/../:/vtests/ -w /vtests/visual-tests --rm vtests',
>   command: 'docker run --name vtests-container --env CURRENT_BRANCH=feature/dev-issue-572 -v /Users/jansiegel/workspace/cr-handsontable/visual-tests/../:/vtests/ -w /vtests/visual-tests --rm vtests',
>   escapedCommand: 'docker run --name vtests-container --env "CURRENT_BRANCH=feature/dev-issue-572" -v "/Users/jansiegel/workspace/cr-handsontable/visual-tests/../:/vtests/" -w "/vtests/visual-tests" --rm vtests',
>   exitCode: 125,
>   signal: undefined,
>   signalDescription: undefined,
>   stdout: undefined,
>   stderr: undefined,
>   failed: true,
>   timedOut: false,
>   isCanceled: false,
>   killed: false
> }
> ```
> 
> Am I doing something wrong?

maciejbo-handsontable:
>     * [#10165 (comment)](https://github.com/handsontable/handsontable/pull/10165#discussion_r1058142185)
>       This comment is still applicable, for example:
>       
>       * `mainTable: 'div.handsontable.htRowHeaders.htColumnHeaders'` can still easily target more than one element (as described in the original comment).
>         https://jsfiddle.net/js_ziggle/e0dgwap8/ - start editing a cell with handsontable-based editor and run `querySelectorAll` with the above selector.

hey, this thing required little more changes, not only in visual testing, so was done just after your comment, but in different PR - it is available here: https://github.com/handsontable/handsontable/pull/10193

Was pretty sure that I told you about it, but now I see that link is missing.

>       * `mainTableFirstColumn: '> .ht_clone_inline_start.ht_clone_left.handsontable table tbody'` in the vast majority of instances doesn't target "first column" - it targets the "left overlay table" (as described in the original comment)

About last point - as far I remember we wasn't sure how to make it flexible, but actually I'm not sure if in our case it is necessary to change this part, cause there is a chance that described case will never happen ( as far as I know we will be testing always this and the same configuration of HoT ).

>     * [#10165 (comment)](https://github.com/handsontable/handsontable/pull/10165#discussion_r1056297438)
>       You didn't respond to this comment in any way, could you clarify if you think it's a good or a bad idea?

My bad, looks like I misundertood this point and thought that it is automatically outdated after improving another thing. Will be improved this week.

>     * [#10165 (comment)](https://github.com/handsontable/handsontable/pull/10165#discussion_r1057565222)
>       If it's a leftover, maybe we should change it to the actual license?

Sure - should I just copy it from other file?

>       **This one is the most important issue** - if I'm seeing this correctly, the script is still not testing the `next` build with the _local builds_. It just installs whatever's in the examples' `package.json` files.

Well, this is the point I had mentioned on last meeting that still can be broken and requiring support - and looks like it is. I do not know process of building examples well and looks like my changes was not enough, but I asked about some introducing and then everything will be clear.

> @maciejbo-hand

visual-tests/tests/.empty-test-template.js
```diff
@@ -0,0 +1,25 @@
...
+  await page.goto(helpers.testURL);
+  await expect(page).toHaveTitle(helpers.expectedPageTitle);
+  stylesToAdd.forEach(item => page.addStyleTag({ path: helpers.cssPath(item) }));
+
+  const table = page.locator(helpers.selectors.mainTable);
+
+  await table.waitFor();
+  helpers.mainTableFirstColumn = table.locator(helpers.selectors.mainTableFirstColumn);
+  helpers.tbody = table.locator(helpers.selectors.mainTableBody);
+  helpers.thead = table.locator(helpers.selectors.mainTableHead);
```
budnix, line 21:
```suggestion
  await table.waitFor();
  
  const mainTableFirstColumn = table.locator(helpers.selectors.mainTableFirstColumn);
  const tbody = table.locator(helpers.selectors.mainTableBody);
  const thead = table.locator(helpers.selectors.mainTableHead);
```
I think that it's cleaner to just assign the elements to the local variables and use them within the test instead of assigning them to helpers utils.

visual-tests/tests/change-rows-order.spec.js
```diff
@@ -0,0 +1,41 @@
...
+  await page.goto(helpers.testURL);
+  await expect(page).toHaveTitle(helpers.expectedPageTitle);
+  stylesToAdd.forEach(item => page.addStyleTag({ path: helpers.cssPath(item) }));
+
+  const table = page.locator(helpers.selectors.mainTable);
+
+  await table.waitFor();
+  helpers.mainTableFirstColumn = table.locator(helpers.selectors.mainTableFirstColumn);
+  helpers.tbody = table.locator(helpers.selectors.mainTableBody);
+  helpers.thead = table.locator(helpers.selectors.mainTableHead);
```
budnix, line 21:
```suggestion
  await table.waitFor();

  const mainTableFirstColumn = table.locator(helpers.selectors.mainTableFirstColumn);
  const cell = mainTableFirstColumn.locator(helpers.findCell({ row: 4, cell: 1, cellType: 'th' }));
```

visual-tests/tests/copy-cell-content.spec.js
```diff
@@ -0,0 +1,46 @@
...
+  cell = helpers.tbody.locator(helpers.findCell({ row: 3, cell: 3, cellType: 'td' }));
+  await cell.dblclick();
+  await cell.type('-test');
+
+  cell = helpers.tbody.locator(helpers.findCell({ row: 4, cell: 3, cellType: 'td' }));
+  await cell.click();
+  await page.keyboard.press('Delete');
+
+  cell = helpers.tbody.locator(helpers.findCell({ row: 3, cell: 3, cellType: 'td' }));
+  await cell.click();
```
budnix, line 39:
```suggestion
  await table.waitFor();
  
  const tbody = table.locator(helpers.selectors.mainTableBody);

  let cell = tbody.locator(helpers.findCell({ row: 3, cell: 3, cellType: 'td' }));

  await cell.click();
  await page.screenshot({ path: helpers.screenshotPath() });
  await page.keyboard.press(`${helpers.modifier}+c`);
  await cell.press('Delete');

  cell = tbody.locator(helpers.findCell({ row: 3, cell: 3, cellType: 'td' }));

  await cell.dblclick();
  await cell.type('-test');

  cell = tbody.locator(helpers.findCell({ row: 4, cell: 3, cellType: 'td' }));

  await cell.click();
  await page.keyboard.press('Delete');

  cell = tbody.locator(helpers.findCell({ row: 3, cell: 3, cellType: 'td' }));

  await cell.click();
```
I think that it's cleaner to just assign the elements to the local variables and use them within the test instead of assigning them to helpers utils.

visual-tests/scripts/visual-tests-build.mjs
maciejbo-handsontable: Done.

visual-tests/tests/expand-dropdown-menu.spec.js
```diff
@@ -0,0 +1,31 @@
...
+  stylesToAdd.forEach(item => page.addStyleTag({ path: helpers.cssPath(item) }));
+
+  const table = page.locator(helpers.selectors.mainTable);
+
+  await table.waitFor();
+  helpers.mainTableFirstColumn = table.locator(helpers.selectors.mainTableFirstColumn);
+  helpers.tbody = table.locator(helpers.selectors.mainTableBody);
+  helpers.thead = table.locator(helpers.selectors.mainTableHead);
+
+  const changeTypeButton = table.locator(helpers.findDropdownMenuExpander({ col: 2 }));
```
budnix, line 23:
```suggestion
  await table.waitFor();

  const changeTypeButton = table.locator(helpers.findDropdownMenuExpander({ col: 2 }));
```
Some of the elements are not used within the test so I suggesting to remove them.

visual-tests/tests/mouse-wheel.spec.js
```diff
@@ -0,0 +1,35 @@
...
+  stylesToAdd.forEach(item => page.addStyleTag({ path: helpers.cssPath(item) }));
+
+  const table = page.locator(helpers.selectors.mainTable);
+
+  await table.waitFor();
+  helpers.mainTableFirstColumn = table.locator(helpers.selectors.mainTableFirstColumn);
+  helpers.tbody = table.locator(helpers.selectors.mainTableBody);
+  helpers.thead = table.locator(helpers.selectors.mainTableHead);
+
+  const tbodyCoordinates = await helpers.tbody.boundingBox();
```
budnix, line 23:
```suggestion
  await table.waitFor();

  const tbody = table.locator(helpers.selectors.mainTableBody);
  const tbodyCoordinates = await tbody.boundingBox();
```

visual-tests/tests/select-few-cells-by-mouse.spec.js
```diff
@@ -0,0 +1,35 @@
...
+
+  const cell = helpers.tbody.locator(helpers.findCell({ row: 2, cell: 2, cellType: 'td' }));
+  const cellCoordinates = await cell.boundingBox();
```
budnix, line 24:
```suggestion
  await table.waitFor();

  const tbody = table.locator(helpers.selectors.mainTableBody);

  const cell = tbody.locator(helpers.findCell({ row: 2, cell: 2, cellType: 'td' }));
  const cellCoordinates = await cell.boundingBox();
```

visual-tests/scripts/visual-tests-build.mjs
maciejbo-handsontable: hi! your suggestion has been implemented

jansiegel:
@maciejbo-handsontable 

> as far as I know we will be testing always this and the same configuration of HoT 

I'm pretty sure that's not true - we might not change the configuration in the next weeks, but in the long run, I'm sure it will evolve one way or another. I think we should make it as future-proof as possible now.

> Sure - should I just copy it from other file?

It depends on the actual license of this project. @adamu-handsontable, could you help us out on that?

budnix (changes requested):

visual-tests/tests/.empty-test-template.js
maciejbo-handsontable: I agree, I did it for only one reason - I wanted to keep them always declared and do not force testers to declare them manually, but then there is a chance that eslint will be yelling about unused vars.

Will `/* eslint-disable no-unused-vars */` be satisfying in this case?

budnix: For the test template, I'd propose commenting 3 lines as starter/hint for the writer then.

visual-tests/tests/expand-dropdown-menu.spec.js
maciejbo-handsontable: Hi! Yes, as I mentioned here: https://github.com/handsontable/handsontable/pull/10165#discussion_r1069027662
it was done in this way intentionally - I wanted to keep some things declared and always just available to make testers life easier.

visual-tests/scripts/visual-tests-build.mjs
budnix: @maciejbo-handsontable, what about that?

visual-tests/scripts/visual-tests-build.mjs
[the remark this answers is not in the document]
budnix: What do you think @jansiegel, about the naming of that file? Because I'm still confused with the name `build`. If I had to choose I'd prefer `test.mjs` as the script not only builds but actually runs the tests using `playwright test` command.

visual-tests/imports/helpers.js
budnix: Don't forget to update the `.gitignore` file :)

visual-tests/scripts/visual-tests-build.mjs
maciejbo-handsontable: if I undertood this suggestion correctly, then it's already done - moved from e.g. `vue/demo` just to `vue`

visual-tests/imports/helpers.js
maciejbo-handsontable: yes, it should be already updated, but, at least temporary, I left there previous path also - just in case if somebody already have screenshots on local machine (e.g. me - I'm not sure who else was playing with this testing already)

visual-tests/tests/.empty-test-template.js
maciejbo-handsontable: The same thing appears in few files - it was done intentionally as a compromise to make tester's life easier, now should be modified as we discussed.

visual-tests/tests/change-rows-order.spec.js
maciejbo-handsontable: The same thing appears in few files - it was done intentionally as a compromise to make tester's life easier, now should be modified as we discussed.

visual-tests/tests/copy-cell-content.spec.js
maciejbo-handsontable: The same thing appears in few files - it was done intentionally as a compromise to make tester's life easier, now should be modified as we discussed.

visual-tests/tests/expand-dropdown-menu.spec.js
maciejbo-handsontable: The same thing appears in few files - it was done intentionally as a compromise to make tester's life easier, now should be modified as we discussed.

visual-tests/tests/mouse-wheel.spec.js
maciejbo-handsontable: The same thing appears in few files - it was done intentionally as a compromise to make tester's life easier, now should be modified as we discussed.

visual-tests/tests/select-few-cells-by-mouse.spec.js
maciejbo-handsontable: The same thing appears in few files - it was done intentionally as a compromise to make tester's life easier, now should be modified as we discussed.

visual-tests/tests/change-rows-order.spec.js
maciejbo-handsontable: probably it's kind of leftover. Modified.

visual-tests/imports/helpers.js
maciejbo-handsontable: probably it's kind of leftover. Modified.

visual-tests/scripts/visual-tests-build.mjs
maciejbo-handsontable: Done.

visual-tests/scripts/visual-tests-build.mjs
budnix: I've fixed that by implementing the `npm run examples:install next/visual-tests` script for Examples within https://github.com/handsontable/handsontable/pull/10165/commits/3a7954606ea62b9c9a3c9952b555634a1aa549be.

budnix:
> * [#10165 (comment)](https://github.com/handsontable/handsontable/pull/10165#discussion_r1058142185)
>   This comment is still applicable, for example:
>   
>   * `mainTable: 'div.handsontable.htRowHeaders.htColumnHeaders'` can still easily target more than one element (as described in the original comment).
>     https://jsfiddle.net/js_ziggle/e0dgwap8/ - start editing a cell with handsontable-based editor and run `querySelectorAll` with the above selector.
>   * `mainTableFirstColumn: '> .ht_clone_inline_start.ht_clone_left.handsontable table tbody'` in the vast majority of instances doesn't target "first column" - it targets the "left overlay table" (as described in the original comment)

I guess that the issue is resolved by merging the https://github.com/handsontable/handsontable/pull/10193?

> * [#10165 (comment)](https://github.com/handsontable/handsontable/pull/10165#discussion_r1057565222)
>   If it's a leftover, maybe we should change it to the actual license?

I take the license from Docs https://github.com/handsontable/handsontable/pull/10165/commits/63a13b6b57a99a128afb47d6fc6615c89baa05a8#diff-7fc05b99ce4363a8f6bb87863ffbf35ce970c329f849caae89ea0d9d71708cedR15.

> * [#10165 (review)](https://github.com/handsontable/handsontable/pull/10165#pullrequestreview-1228873761)
>   **This one is the most important issue** - if I'm seeing this correctly, the script is still not testing the `next` build with the _local builds_. It just installs whatever's in the examples' `package.json` files.

Definitely, it's a critical bug to fix. I've implemented the `npm run examples:install next/visual-tests` script with symlinking ([3a7954](https://github.com/handsontable/handsontable/pull/10165/commits/3a7954606ea62b9c9a3c9952b555634a1aa549be), [6367f2ec](https://github.com/handsontable/handsontable/pull/10165/commits/6367f2ec52d94979968f46877cfd506937067197)) and I added to the VT scripts flow. Now, the runner tests the changed Handsontable code with wrappers and not the latest production version.

budnix:
@jansiegel @adamu-handsontable I think I've pushed all the bugfixes mentioned by the @jansiegel and by me. The PR is ready to review.

jansiegel (commented):
We should probably exclude visual testing from some scripts in the monorepo root:

- https://github.com/handsontable/handsontable/blob/e7a052437dd87e0bd373f0283b1c91397aee647c/package.json#L20-L21
- https://github.com/handsontable/handsontable/blob/e7a052437dd87e0bd373f0283b1c91397aee647c/package.json#L26
- https://github.com/handsontable/handsontable/blob/e7a052437dd87e0bd373f0283b1c91397aee647c/scripts/freeze.mjs#L106-L110

budnix:
> We should probably exclude visual testing from some scripts in the monorepo root:
> 
> * https://github.com/handsontable/handsontable/blob/e7a052437dd87e0bd373f0283b1c91397aee647c/package.json#L20-L21
>    (maybe not `test`? I'm not sure if the visual tests should be deployed with each testing scenario?)

The scripts are intentionally added to the test flow.
> * https://github.com/handsontable/handsontable/blob/e7a052437dd87e0bd373f0283b1c91397aee647c/package.json#L26

The `publish-package` is handled within the visual-tests directory as [simple log](https://github.com/handsontable/handsontable/blob/feature/dev-issue-572/visual-tests/package.json#L13). I think that way is better than excluding in the root directory. In this way, the developer can work within the package directory right away and not wonder why the `publish-package` script does not work.
> * https://github.com/handsontable/handsontable/blob/e7a052437dd87e0bd373f0283b1c91397aee647c/scripts/freeze.mjs#L106-L110
I've added (62df3add761248e35fcd8b04185151058ceecae2) a similar log as for `publish-package`. The `build` script should not be excluded as below there is executed `npm run all test` which runs the visual tests that need that build.

jansiegel:
@budnix 
I'm still not sure about visual tests' `build` being a part of the main `npm run build`, as I considered it to be a script that builds all the publishable packages, and visual tests are not one of them. But I'll leave the decision up to you.

Apart from that, I think everything looks ok, but I found these issues/questions:
1. Is there any way to run the visual tests for one framework only? I can imagine working on a framework-specific task and not wanting to run all the tests each time.
2. When I tested if the script uses local builds, I purposefully broke something in one of the wrappers. The tests failed (as expected), but returned this:
    ```
    Serving HTML report at http://localhost:9323. Press Ctrl+C to quit.
    Failed to open browser on http://localhost:9323
    ```
    And, indeed, `http://localhost:9323` did not work.

budnix:
> @budnix I'm still not sure about visual tests' `build` being a part of the main `npm run build`, as I considered it to be a script that builds all the publishable packages, and visual tests are not one of them. But I'll leave the decision up to you.

What about the `test` command of the visual tests? Do you think it should be plugged into the workspace `test` flow? I think it's worth adding ... if yes then the `build` script is also necessary.

> 
> Apart from that, I think everything looks ok, but I found these issues/questions:
> 
> 1. Is there any way to run the visual tests for one framework only? I can imagine working on a framework-specific task and not wanting to run all the tests each time.

Currently, it is not possible.
> 2. When I tested if the script uses local builds, I purposefully broke something in one of the wrappers. The tests failed (as expected)

I've changed the reporter type (to more CLI-ish) within https://github.com/handsontable/handsontable/pull/10165/commits/ff45f9f6223ee563359c4443beffabd01687b4c0#diff-3fddd5c52e06a80e6e1be02b9eb7290e0b5415fa7cc7ece099212feaa3fa7ea6R64.

jansiegel (dismissed):
👍

jansiegel (approved):

budnix (approved):

PR #9542, opened 2022-05-26
Fix the Walkontable viewport calculations for oversized rows and columns

budnix:
### Context

The PR fixes a bug where Handsontable was not able to render correctly oversized rows and/or columns. By saying _oversized_ I mean the columns that are wider than the table's viewport width and rows that are longer than the table's viewport height.

To make it work there needed to fix two problems. 
 1. The fix for the [Viewport calculation](https://github.com/handsontable/handsontable/pull/9542/files#diff-6521dd3b666af38c14977ac5efe97fdd0598476b5528bf31806e6880af1fb549R449). There was a bug that causes not to render non-fully visible records. The fix adds conditions that check whether only the partially visible records are rendered in the viewport and if true then the full table's render is performed.
 2. The fix for [the Scroll module](https://github.com/handsontable/handsontable/pull/9542/files#diff-6c2914680f1297055da0e67a66d531aa5432a30261182dfbc3e3a513e0f00651R11). To support scrolling between partially visible records the class had to be refactored. The module allows scrolling the viewport to the cell and choosing to which part of the table's edge the cell will be stuck. For oversized records, this is not possible. The oversized cells are always stuck to the top-left (or top-right in RTL) position (the same works in other common spreadsheets avialable in the market).

@kirszenbaum about point 2 I extended the method description. You may [check that](https://github.com/handsontable/handsontable/pull/9542/files#diff-bf6f9c23fc3252c45d93ec80d4afd013da7973589a0ee32f08d5544554626f49R4009).

Example configuration:
```js
colWidths: [50, 50, 4000, 500, 1400, 700, 800, 500, 304, 50, 50, 80]
// or 
colWidths: 2000
```

This is how it works with PR's changes:
##### Scrolling using the mouse
Kapture 2022-06-02 at 11 21 20

##### Keyboard navigation with cell editing
Kapture 2022-06-02 at 11 20 12

&#65121; The same works for the oversized rows.

### How has this been tested?

I tested the changes locally. I reorganize and add new tests for the scrolling feature.

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

### Related issue(s):
1. resolves #9473

### Affected project(s):
- [x] `handsontable`

### Checklist:

- [x] I have reviewed the guidelines about [Contributing to Handsontable](https://github.com/handsontable/handsontable/blob/master/CONTRIBUTING.md) and I confirm that my code follows the code style of this project.
[...]

PR #10206, opened 2023-01-12
Make the table trigger "fast" render instead of a full one, when the table is outside of the viewport.

jansiegel:
### Context
As described in https://github.com/handsontable/dev-handsontable/issues/983, since `12.1.0`, the table triggers a "full" render with every scroll when the table is outside of the viewport.

This PR aims to restore the state from `<12.1.0` - when outside the viewport, the table renders only using the "fast" mode.

This should also fix the problem described in handsontable/dev-handsontable#561.

Additional to the added logic and tests, I removed the `lastSelectedColumn` property from the Filters plugin because I didn't find anything referencing it.

### How has this been tested?
Added test cases.

### Types of changes

### Related issue(s):
1. handsontable/dev-handsontable#561
2. partially resolves https://github.com/handsontable/dev-handsontable/issues/983

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:


budnix (changes requested):
I tested the changes, and here are my thoughts:
1. Generally, I think that it's a good idea to test if the table is visible in the viewport or not - so good direction. However, I'd propose merging the `areAnyRowsFullyInViewport` and `areAnyColumnsFullyInViewport` methods to one that uses `IntersectionObserver` API. The code would be much more simplified and bullet-proof. For example, code for both axes:
   ```js
    this.visibilityObserver = new IntersectionObserver((entries) => {
      entries.forEach((entry) => {
        this.isOutsideOfViewport = !entry.isIntersecting;
      });
    }, {
      root: null,
    });
   ```
   Right now, there is a bug in those methods which appears in resulting wrong `true`/`false` results. See the gif 
Kapture 2023-01-17 at 13 51 33
3. The main issue seems to be solved (that with the checkboxes in the list), but I think it's some sort of lucky code combination related to the bug in the above methods 😄 I've tried using the `IntersectionObserver` API, and the fast/slow render in and outside the viewport work as expected but not the main issue.
4. The last thing. For viewport stuff, I think the best place for that code would be the `viewport.js` file. Thanks to that you won't have to juggle arguments back and forth in the method `wtViewport.createRenderCalculators` within the WoT Table class.

jansiegel:
@budnix 

As for `1`, there are a few things worth mentioning here:
1. The names I gave those functions are oversimplified and thus misleading, I'll need to address that. What they're supposed to check is if all the rows of the table **except for the last one** are in the viewport. So the first issue you're showing on the gif is actually a feature 😆 (The second one isn't, it's clearly a bug)
  If I remember correctly (it's been some time since I created those methods), it needs to check it that way because of when the calculators were re-triggered when scrolling the table. Without that, the issue with the checkboxes would not be fixed - with the last row in the viewport, there would still be a full render. Which brings me to...

2. ... the intersection observer idea. In general, I totally agree - the intersection observer would be a perfect thing to use here if it wasn't for the problem with the last row being visible in the viewport. I'm guessing that's probably the reason why your proof of concept didn't solve the original issue.

budnix:
> 2\. ... the intersection observer idea. In general, I totally agree - the intersection observer would be a perfect thing to use here if it wasn't for the problem with the last row being visible in the viewport. I'm guessing that's probably the reason why your proof of concept didn't solve the original issue.

I'd be for more deeply investigating why there is a need for a gap for the last row and try to use the intersection observer idea. The current implementation seems to be covering some other bug.

jansiegel:
> I'd be for more deeply investigating why there is a need for a gap for the last row and try to use the intersection observer idea. The current implementation seems to be covering some other bug.

I'll get to it and try to elaborate more on the exact reason why this is happening 👍

jansiegel:
@budnix 
The logic underneath calculating the viewport without the last row/column stems from the change introduced in https://github.com/handsontable/handsontable/pull/9542, namely:

> The fix adds conditions that check whether only the partially visible records are rendered in the viewport and if true then the full table's render is performed.

This, as a result, forces the table to perform full renders if the table is outside of the viewport _or_ only the last row/column is visible.
That, in turn, causes the checkboxes in the dropdown menu of the Filters plugin to get re-rendered and lose their state.

Removing the logic from #9542 would break instances with oversized cells taking up more than the width/height of the viewport - leaving it in and using the intersection observer would not resolve the issue from handsontable/dev-handsontable#561.

Do you have any suggestions on how to move forward?

aninde:
I checked and confirmed that the 'Walkontable' tests pass in every supported environment while waiting for the review to end.

budnix:
> Do you have any suggestions on how to move forward?

I investigated the issue, and I think that the best idea is to use rows/columns calculator to detect if the row or column is in or out of the viewport. Then, with my [PoC](https://github.com/handsontable/handsontable/compare/feature/dev-issue-561...feature/dev-issue-561-poc?expand=1), I used that flag to improve the changes that you've mentioned (https://github.com/handsontable/handsontable/pull/9542). WDYT?

jansiegel:
> > Do you have any suggestions on how to move forward?
> 
> I investigated the issue, and I think that the best idea is to use rows/columns calculator to detect if the row or column is in or out of the viewport. Then, with my [PoC](https://github.com/handsontable/handsontable/compare/feature/dev-issue-561...feature/dev-issue-561-poc?expand=1), I used that flag to improve the changes that you've mentioned (#9542). WDYT?

Looks great! 🏅 Much cleaner than my initial idea. Would you make a PR of that branch so that we could merge it into this one?

budnix:
> Would you make a PR of that branch so that we could merge it into this one?

Something like that https://github.com/handsontable/handsontable/pull/10224?

jansiegel:
[As discussed](https://handsoncode.slack.com/archives/C04MC750UJW/p1676640794236649), handsontable/dev-handsontable#1106 (marked as a blocker) is being postponed and is _not_ fixed in this PR.

handsontable/dev-handsontable#1107, however, should work all right after merging this PR.

@budnix, could you re-review?

budnix (changes requested):
We are almost there 👏 I think there is missing logic in the FULLY_VISIBLE_TYPE calculator that returns the wrong `isVisibleInTrimmingContainer` value when the table disappears at the bottom of the browser viewport.

For the browsers top edge, it works correctly:
Kapture 2023-02-21 at 14 45 51

For the browsers bottom edge, it's not:
Kapture 2023-02-21 at 14 51 31

budnix (approved):
👍

Issue #31, opened 2012-06-26
Autocomplete visible after Escape key press


jahewson:
Pressing `Esc` when an autocomplete menu is visible causes the current cell to be de-focused, but does not hide the autocomplete menu. Pressing `Esc` again causes the autocomplete menu to disappear. Obviously both these should be happening with just one `Esc` key press.

Keep up the good work!

Issue #1025, opened 2013-09-05
render() issue with AngularJS


noahkeitel:
Issue:
1. handsontable object is inside a DOM element which is hidden via AngularJS's ng-show="false"

2. We add rows to our handsontable via the .alter('insert_row') method and render() it
3. We set our DOM element to ng-show="true".
--everything works fine--

4. Now we remove rows via the .alter('remove_row') method.
5. Set the DOM to ng-show="false"
--still good--

6. Add rows to the same handsontable via the .alter('insert_row') method and .render() it
7. We make sure the rows were added by logging ourTableObject.getData() to the console.
--all good still--

8. We set our DOM to ng-show="true"
-- This is the point at which the table has NOT rendered like it should have--

I've tried to set ng-show to true before adding the rows in step 6 to see if it works but it doesn't. As long as I don't set ng-show="false" ever, all added and removed rows work and render correctly. Not sure if this is an issue with the time things are fired in AngularJS's digest/apply or if it's an issue with handsontable.

AMBudnik:
Issue closed in lack of JSFiddle demo

Issue #32, opened 2012-06-26
Feature Request: col/row span


splitbrain:
It would be great if we could paste cells with row or col spans from open office and have method to create/remove them in the editor itself.

PR #10210, opened 2023-01-13
Fixed an issue where some browser or system shortcuts left the CTRL/CMD key held down

wszymanski:
### Context

Currently, most of the keyboard events is handled by the `ShortcutManager`. Internal recorder is responsible for catching `keydown` and `keyup` events. Unfortunately, using browser or system shortcut breaks a flow which releasing list of pressed keys. After triggering such a shortcut, a `keyup` event isn't emitted.

I'm afraid that there is no alternative way of fixing the **source of the problem**. This solution fixes just found side effects by adding extra callback to some hook. It also changing helpers for tests. Separately used jQuery simulate events don't transfer information about pressed meta keys to each other. I decided not to pass extra options to every method call, but use modified helper.

**Note**: I haven't added extra e2e test as is seems to be impossible to trigger system/browser events using simulated events.

### How has this been tested?

I was testing cases from the issue and other system/browser shortcuts.

### Types of changes

### Related issue(s):
1. [#31](https://github.com/handsontable/dev-handsontable/issues/31)

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

### DIFF
--- a/handsontable/src/plugins/columnSorting/__tests__/columnSorting.spec.js
+++ b/handsontable/src/plugins/columnSorting/__tests__/columnSorting.spec.js
@@ -13,9 +13,7 @@ describe('ColumnSorting', () => {
         throw Error('Please check the test scenario. The header doesn\'t exist.');
       }
 
-      $spanInsideHeader.simulate('mousedown');
-      $spanInsideHeader.simulate('mouseup');
-      $spanInsideHeader.simulate('click');
+      simulateClick($spanInsideHeader[0]);
     };
   });
 
--- a/handsontable/src/plugins/columnSorting/__tests__/columnSorting.spec.js
+++ b/handsontable/src/plugins/columnSorting/__tests__/columnSorting.spec.js
@@ -13,7 +13,7 @@ describe('ColumnSorting', () => {
         throw Error('Please check the test scenario. The header doesn\'t exist.');
       }
 
-      simulateClick($spanInsideHeader[0]);
+      simulateClick($spanInsideHeader);
     };
   });
 
--- a/handsontable/src/plugins/multiColumnSorting/__tests__/multiColumnSorting.spec.js
+++ b/handsontable/src/plugins/multiColumnSorting/__tests__/multiColumnSorting.spec.js
@@ -13,9 +13,7 @@ describe('MultiColumnSorting', () => {
         throw Error('Please check the test scenario. The header doesn\'t exist.');
       }
 
-      $spanInsideHeader.simulate('mousedown');
-      $spanInsideHeader.simulate('mouseup');
-      $spanInsideHeader.simulate('click');
+      simulateClick($spanInsideHeader[0]);
     };
   });
 
--- a/handsontable/src/plugins/multiColumnSorting/__tests__/multiColumnSorting.spec.js
+++ b/handsontable/src/plugins/multiColumnSorting/__tests__/multiColumnSorting.spec.js
@@ -13,7 +13,7 @@ describe('MultiColumnSorting', () => {
         throw Error('Please check the test scenario. The header doesn\'t exist.');
       }
 
-      simulateClick($spanInsideHeader[0]);
+      simulateClick($spanInsideHeader);
     };
   });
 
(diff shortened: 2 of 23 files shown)


handsontable/src/plugins/multiColumnSorting/__tests__/multiColumnSorting.spec.js
```diff
@@ -13,9 +13,7 @@ describe('MultiColumnSorting', () => {
         throw Error('Please check the test scenario. The header doesn\'t exist.');
       }
 
-      $spanInsideHeader.simulate('mousedown');
-      $spanInsideHeader.simulate('mouseup');
-      $spanInsideHeader.simulate('click');
+      simulateClick($spanInsideHeader[0]);
```
budnix, line 16:
```suggestion
      simulateClick($spanInsideHeader);
```
AFAIK the helper accepts the jQuery element instance as well.

handsontable/src/plugins/columnSorting/__tests__/columnSorting.spec.js
```diff
@@ -13,9 +13,7 @@ describe('ColumnSorting', () => {
         throw Error('Please check the test scenario. The header doesn\'t exist.');
       }
 
-      $spanInsideHeader.simulate('mousedown');
-      $spanInsideHeader.simulate('mouseup');
-      $spanInsideHeader.simulate('click');
+      simulateClick($spanInsideHeader[0]);
```
budnix, line 16:
```suggestion
      simulateClick($spanInsideHeader);
```
AFAIK the helper accepts the jQuery element instance as well.

AMBudnik:
I reported @budnix issue here https://github.com/handsontable/dev-handsontable/issues/1025 as a new case, as all of the reported cases at https://github.com/handsontable/dev-handsontable/issues/31 relates to lack of CTRL/CMD release and this case relates to the opposite - lack of CTRL/CMD activation for range cell selection.

For the original *Case 3*, we got the video reproduction from the user

https://user-images.githubusercontent.com/10757813/212658067-092c4ca2-a0c3-47c7-a107-72a346e8ae38.mov

and the following steps:
1. press ctrl
2. click on a few cells
3. release ctrl
4. click on a few cells
5. click outside of table
6. click on a few cells

**Users environment: macOS/Chrome 78**

handsontable/src/plugins/columnSorting/__tests__/columnSorting.spec.js
wszymanski: Changed within https://github.com/handsontable/handsontable/pull/10210/commits/5df751b593920ca0f2e63f09e7a9522ec8d0425c.

handsontable/src/plugins/multiColumnSorting/__tests__/multiColumnSorting.spec.js
wszymanski: Changed within https://github.com/handsontable/handsontable/pull/10210/commits/5df751b593920ca0f2e63f09e7a9522ec8d0425c.

handsontable/src/shortcuts/__tests__/shortcutManager.spec.js
```diff
@@ -478,4 +478,23 @@ describe('shortcutManager', () => {
 
     document.body.removeChild(externalInputElement);
   });
+
+  it('should check if there is a need of releasing keys on click #32', () => {
```
budnix, line 482:
```suggestion
  it('should check if there is a need of releasing keys on click #dev-1025', () => {
```
`#32`? Here shouldn't be [#1025 ](https://github.com/handsontable/dev-handsontable/issues/1025)?

budnix (changes requested):
Looks good 👍 There is one suggestion left to resolve before merging.

budnix (approved):

aninde:
I can confirm that the reported bug has been fixed on all supported environments. I also checked it by switching focus between Handsontable and an HTML input (https://jsfiddle.net/aninde/5rp0g72y/) and a text area. I can confirm that other keyboard shortcuts, such as those using control, meta, and alt, are still working properly.

AMBudnik:
Released in Handsontable v12.3.1

PR #10263, opened 2023-02-27
Remove UNSAFE_ React methods using portals

budnix:
### Context

The PR removes all UNSAFE_ React lifecycle methods used within the Handsontable wrapper for React. The React wrapper and custom editors implementation was dependent on the deprecated `componentWillMount` and `componentWillUpdate` methods; thus, there was needed more logic refactor than I previously assumed.

Previously the wrapper created and injected the editor portal before React rendered the DOM. This was straight forward solution as it did not require any other logic to make it work. To make the wrapper work in strict mode, there is necessary to remove UNSAFE methods that allow do something before render cycle. I tried different techniques, but finally, I stuck to the approach that is already used by the custom renderers - PortalManager. The [`EditorPortalManager`](https://github.com/handsontable/handsontable/pull/10263/files#diff-c254d97e004afeceda44d15f37b897d50388b108b177ca8f86fa88c911e39480R1) allows injecting custom-created editors into the React render cycle using the `setState` method that is called right after mounting or updating the component.

Moreover, within the PR, I improved the following things:
 *  I removed the code from the editors portal that manages the DOM manually (outside of the React lifecycle) and moved that logic to the React ([diff-bdef8](https://github.com/handsontable/handsontable/pull/10263/files#diff-bdef8b785583a393d5b12349e1fae7d772131f048627a2fe52d37c8a5a5274b7L83-R103));
 * I found a memory leak in tests. There was a missing component unmounting call for each test. The bug led to the tests failing. I implemented the missing line ([diff-6903ba](https://github.com/handsontable/handsontable/pull/10263/files#diff-6903ba0eb54ff52d83dfc28a6d97445be1fdbd8468fe71a1f43721aea94ad0e1R2));
 * Cleanup the code as some of the types are explicitly declared, which is not needed as TS is able to detect some types from the code implicitly, and I removed some unused function arguments;

### Tasks to be done
- [x] Remove unsafe methods from the `HotTable` component, test it, and cover it with new tests;
- [x] Remove unsafe methods from the `HotColumn` component, test it, and cover it with new tests;
- [x] Test the wrapper with a simple React App where strict mode is enabled.

### How has this been tested?

I tested the changes locally by creating a simple React app that implements different use cases by enabling/disabling custom editors and renderers within the `HotTable` and `HotColumn` components. The same things I tested with and without Strict mode, and all worked. I compared the DOM structure of the editors before and after the PRs changes, and the structure looks the same.

### Types of changes

### Related issue(s):
1. fixes https://github.com/handsontable/dev-handsontable/issues/465

### Affected project(s):
- [x] `@handsontable/react`

### Checklist:


jansiegel (approved):

wrappers/react/src/baseEditorComponent.tsx
```diff
@@ -2,7 +2,12 @@ import React from 'react';
 import Handsontable from 'handsontable/base';
 import { HotEditorProps } from './types';
 
-class BaseEditorComponent<P = {}, S = {}, SS = any> extends React.Component<P | HotEditorProps, S> implements Handsontable.editors.BaseEditor {
+interface BaseEditorProps extends HotEditorProps {
+  editorColumnScope: number;
```
samsullivan, line 6: @jansiegel, was this `editorColumnScope` prop supposed to be optional and/or defaulting to "global"? That is our team's understanding after reading https://github.com/handsontable/handsontable/pull/8901; for now, we're explicitly defining "global" for our custom React editors to make TypeScript happy.

I can open a formal issue and/or patch, too.

AMBudnik: Hi @samsullivan Thank you for the feedback! You're right - we've made a PR to make the prop optional https://github.com/handsontable/handsontable/pull/10429 
I will update you as soon as we release a new version with the change.

AMBudnik: Hi @samsullivan we just released Handsontable v13.1.0 with the changes. Again, thank you for your input.

PR #10307, opened 2023-03-24
Fix error that disallow getting the HoT instance using useRef within useEffect hooks

budnix:
### Context

The PR fixes the regression bug that was introduced within the https://github.com/handsontable/handsontable/pull/10263. The bug prevented accessing the Handsontable instance when the `useEffect` with `useRef` is used.

Moreover, the change fixes all issues related to React strict mode that causes misrendering of the custom editors and/or renderers. Now all custom components should work with and without strict mode.   

After the fix, this should work:
```jsx
  function ExampleComponent() {
    const hotRef = useRef(null);

    useEffect(() => {
      const hot = hotRef.current.hotInstance;

      data = hot.getData(); // the Handsontable instance exists, and the data is the same as that passed in the settings
    });

    return (
      <HotTable licenseKey="non-commercial-and-evaluation"
        ref={hotRef}
        id="test-hot"
        data={refData}
        width={300}
        height={300}
        rowHeights={23}
        colWidths={50}
        init={function () {
          mockElementDimensions(this.rootElement, 300, 300);
        }}
      ></HotTable>
    )
  }
```

_[skip changelog]_

### How has this been tested?

I tested the changes locally in the simple app and I covered the changes with new E2E test.

### Types of changes

### Related issue(s):
1. fixes https://github.com/handsontable/dev-handsontable/issues/1230
2. should fix https://github.com/handsontable/dev-handsontable/issues/76

### Affected project(s):
- [x] `@handsontable/react`

### Checklist:


jansiegel:
@budnix This fix introduces a regression in terms of how the editors are rendered.

- Before this PR:
    The editors were displayed "above" the Handsontable's trimming container: 
image
- After this PR:
    The editor container is trimmed by Handsontable's trimming container: 
image

budnix:
@jansiegel Thanks! I've fixed that in https://github.com/handsontable/handsontable/pull/10307/commits/80e55fc6a2fae6247a0293cb6b8aa16e6889a6a2. Can you check that?

PR #10309, opened 2023-03-24
Docs: Rewrite the "Column filter" page

kirszenbaum:
This PR:
- Rewrites the [Column filter](https://handsontable.com/docs/react-data-grid/column-filter/) guide

[skip changelog]


docs/content/guides/columns/column-filter.md
```diff
@@ -155,765 +450,2860 @@ import 'handsontable/dist/handsontable.full.min.css';
...
 
-  getSelectedColumn() {
-    return this.selectedColumn;
-  }
+## Customize the filter button
 
-  getValuesToFilter() {
-    return this.valuesToFilter;
-  }
+The filter menu is part of the [column menu](@/guides/columns/column-menu.md), so they both use the
```
krzysztofspilka, line 1970: The filter menu is not part of the column menu (same case as previously), so I would just skip that part.

This section should be similar to https://handsontable.com/docs/react-data-grid/rows-sorting/#add-custom-sort-icons, maybe:

> The default button that opens the column menu can be styled with CSS by modifying `button.changeType` and its `::before` pseudoclass that contains an HTML entity displaying an arrow down icon.

I would change the style of the buttons in the demo slightly:

```css
.custom-filter-button-example-1 .changeType {
    background: #e2e2e2;
    border-radius: 100%;
    width: 16px;
    color: #0075ff;
    height: 16px;
    padding: 3px 2px;
    border: none;
}

.custom-filter-button-example-1 .changeType::before {
    content: "▼ ";
    zoom: .9;
}
```

The result:
Screenshot 2023-04-12 at 22 48 17

docs/content/guides/columns/column-filter.md
```diff
@@ -155,765 +450,2860 @@ import 'handsontable/dist/handsontable.full.min.css';
...
 }
+```
 
-const container = document.querySelector('#example4');
-const hot = new Handsontable(container, {
+:::
+
+:::
+
+To make the column menu button visible only on hover, override Handsontable's CSS by using this
```
krzysztofspilka, line 2233: Maybe: 
The column menu button is always visible, but if you want it to appear only when the mouse cursor is over the header, apply an additional CSS to `.relative:hover .changeType`

docs/content/guides/columns/column-filter.md
```diff
@@ -155,765 +450,2860 @@ import 'handsontable/dist/handsontable.full.min.css';
...
+}
+```
+
+:::
+
+:::
+
+## Change the width of the filter menu
+
+To make the filter menu wider, you can change the width of the entire
```
krzysztofspilka, line 2499: Maybe:
If you have long text data in your columns that don't fit in the filters container, you may need to adjust the column menu's width for a better user experience. Fortunately, you can achieve this with just CSS by adding a style to .htDropdownMenu table.htCore.

docs/content/guides/columns/column-filter.md
```diff
@@ -155,765 +450,2860 @@ import 'handsontable/dist/handsontable.full.min.css';
...
+.custom-filter-button-example-2 .relative:hover .changeType {
+  visibility: visible;
+}
+```
+
+:::
+
+:::
+
+## Change the width of the filter menu
```
krzysztofspilka, line 2497: A demo would be nice here. It's an issue when the long text is cropped in the filter values preview.

docs/content/guides/columns/column-filter.md
```diff
@@ -155,765 +450,2860 @@ import 'handsontable/dist/handsontable.full.min.css';
...
+```css
+/* style the column menu */
+.handsontable .htDropdownMenu table.htCore {
+  width: 300px !important;
+}
+```
+
+## Exclude rows from filtering
+
+You can disable filtering for specific rows by changing their indexes. This lets you keep portions
```
krzysztofspilka, line 2512: > You can disable filtering for specific rows by changing their indexes.

I don't know what that means. Always give enough context for freshers, ideally addressing pain and providing a solution with the clear benefit of using it.

WHY would I need to disable filtering for specific rows? Probably for the same reason you don't want to sort particular rows. I would make this section a copycat of https://handsontable.com/docs/react-data-grid/rows-sorting/#exclude-rows-from-sorting

docs/content/guides/columns/column-filter.md
```diff
@@ -155,765 +450,2860 @@ import 'handsontable/dist/handsontable.full.min.css';
...
-    } else {
-      setSelectedRowEntries([]);
-    }
+:::
+
+:::
+
+## Use filtering hooks
+
+You can run your code before or after filtering, using the following
```
krzysztofspilka, line 2778: Again, what's the use case, where's the pain and benefit of using Handsontable hooks in this context? The part about server-side filtering, without an actual demo, is a bit blurry to me. Our technical support could provide some insights of how the before/after hooks are used in real-life cases.

docs/content/guides/columns/column-filter.md
```diff
@@ -155,765 +450,2860 @@ import 'handsontable/dist/handsontable.full.min.css';
...
+This allows you to:
 
-      setColHeaders(hot.getColHeader());
-      setRowEntries(fetchedEntries);
-      setSelectedRowEntries(fetchedEntries);
-    }
+- Enable or disable filtering based on specified conditions. For example, you can disable filtering
+  for very large data sets.
+- Trigger filtering based on the state of another component in your application. For example, you
+  can let the end user filter data by clicking on buttons outside of the grid.
```
krzysztofspilka, line 2829: Maybe:
ou can enable or disable filtering based on specific conditions. For instance, you may create a UI outside of the grid to manage its filtering behavior.

docs/content/guides/columns/column-filter.md
```diff
@@ -155,765 +450,2860 @@ import 'handsontable/dist/handsontable.full.min.css';
...
-      const fetchedEntries = hot.getSourceDataAtCol(0);
+You can control filtering at the grid's runtime by using Handsontable's
+[hooks](@/guides/getting-started/events-and-hooks.md) and [API methods](@/api/filters.md#methods).
+This allows you to:
 
-      setColHeaders(hot.getColHeader());
-      setRowEntries(fetchedEntries);
-      setSelectedRowEntries(fetchedEntries);
-    }
+- Enable or disable filtering based on specified conditions. For example, you can disable filtering
```
krzysztofspilka, line 2826: I'm not sure I understand I would I want to do disable filtering for a large data set.

docs/content/guides/columns/column-filter.md
```diff
@@ -155,765 +450,2860 @@ import 'handsontable/dist/handsontable.full.min.css';
...
+registerPlugin(Filters);
+```
+
+## Known limitations
+
+At the moment, filtering comes with the following limitations:
+
+- You can't use filtering with [nested data structures](@/guides/rows/row-parent-child.md)
+  ([`NestedRows`](@/api/nestedRows.md)).
+- You can't add custom filter conditions.
```
krzysztofspilka, line 3285: Well, you can, it's just JS, you can do everything you want to. The actual problem here is that there is no easy way to add the custom filter conditions to the user interface.

docs/content/guides/columns/column-filter.md
```diff
@@ -155,765 +450,2860 @@ import 'handsontable/dist/handsontable.full.min.css';
...
+- You can't control the **Filter by value** list. The list of values is generated automatically and
```
krzysztofspilka, line 3286: I don't know what the "control" means here.

docs/content/guides/columns/column-filter.md
```diff
@@ -155,765 +450,2860 @@ import 'handsontable/dist/handsontable.full.min.css';
...
+  there's no supported way of modifying it.
+- There's no built-in protection against filtering passwords.
```
krzysztofspilka, line 3288: What do you mean by that? It's unclear to me, even though I know a lot about issues with presenting passwords in the filters menu.

docs/content/guides/columns/column-filter.md
```diff
@@ -155,765 +450,2860 @@ import 'handsontable/dist/handsontable.full.min.css';
...
+// register the filtering plugins
+registerPlugin(DropdownMenu);
+registerPlugin(Filters);
+```
+
+## Known limitations
+
+At the moment, filtering comes with the following limitations:
+
+- You can't use filtering with [nested data structures](@/guides/rows/row-parent-child.md)
```
krzysztofspilka, line 3283: You can, but it won't work properly. https://jsfiddle.net/handsoncode/p4mnz1ws/1/

krzysztofspilka (changes requested):

docs/content/guides/columns/column-filter.md
```diff
@@ -155,765 +450,2860 @@ import 'handsontable/dist/handsontable.full.min.css';
...
-    this.clearFilterUI = this.externalFilterUI.querySelector('.buttons > .clear');
-    this.inputs = [];
-
-    this.fillSelectByColHeaders();
-    this.fillValueBoxByData();
-    this.initListeners();
-
-    this.externalFilterUI.style.display = 'block';
-  }
+For that, use the [`Filters`](@/api/filters.md) plugin's API. For example, the following demo starts
```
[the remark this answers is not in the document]
kirszenbaum: It's different for React, because the React demo works differently (uses `useEffect()` instead of `afterInit()`.

docs/content/guides/columns/column-filter.md
```diff
@@ -155,765 +450,2860 @@ import 'handsontable/dist/handsontable.full.min.css';
...
-      input.name = 'cellData';
-      input.value = cellData;
-      input.checked = true;
+:::
 
-      input.addEventListener('change', event => this.onInputChange(event));
+## Filter as you type
 
-      const label = document.createElement('label');
+You can filter data by typing into a text box. This functionality is similar to dynamic filters in
```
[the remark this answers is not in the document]
kirszenbaum: > The text input should find and filter out the values across all columns by default

@krzysztofspilka As far as I know, that's not possible. @adrianszymanski89 can you confirm?

docs/content/guides/columns/column-filter.md
kirszenbaum: We came across a problem here: an older method of changing the menu width (through `modifyColWidth`) doesn't work anymore. And changing the menu width through CSS doesn't work with the `className` option, so the styling changes would affect all demos on the page. @adrianszymanski89 can you confirm?

docs/content/guides/columns/column-filter.md
kirszenbaum: @AMBudnik are there any other uses of the `beforeFilter` and `afterFilter` hooks that come up in support cases (apart from blocking filtering on the client side by using `beforeFilter`)?

AMBudnik: There was one case when the developer needed to show a box if there's no data after filtering - https://jsfiddle.net/AMBudnik/91m4kLpr/.

Zrzut ekranu 2023-04-19 o 07 56 24

But mainly, it is about the logic 
1. Block the front-end filtering and proceed with filtering on back-end (but that one you've already discussed)
2. Retrieving the filtering states - http://jsfiddle.net/AMBudnik/1ebkcdan/ we use the `beforeFilter` hook to get the filtering conditions so the developer can reattach them when their user logs in. 
3. Collecting indexes of visible rows after filtering - https://jsfiddle.net/p9hz1wg3/1/ (but here I recommend asking devs which option to choose: `renderablePhysicalIndexesCache` or `notTrimmedIndexesCache`).

kirszenbaum: Thanks @AMBudnik!

@krzysztofspilka – right now, the **Use filtering hooks** section mirrors the [Use sorting hooks](https://handsontable.com/docs/javascript-data-grid/rows-sorting/#use-sorting-hooks) section. But if we want to focus on the use cases mentioned, we can get rid of the **Use filtering hooks** section altogether, and add the following ones:
- **Block client-side filtering** (@AMBudnik mentioned it would be useful to have this info on the page)
- **Save filter settings** or **Retrieve filter conditions** (which, as discussed, was planned for [phase 2](https://github.com/handsontable/dev-handsontable/issues/1258))

docs/content/guides/columns/column-filter.md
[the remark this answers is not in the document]
adrianszymanski89: @kirszenbaum, we rechecked it and discussed it again with @AMBudnik, and it looks like, to the best of our knowledge, it will not be possible to do with the filters plugin. One solution would be to operate strictly on trimRows, entirely omitting the plugin, but I think it's not the point in this example.

docs/content/guides/columns/column-filter.md
```diff
@@ -155,765 +449,3045 @@ import 'handsontable/dist/handsontable.full.min.css';
...
+```
+
+```css
+.controlsQuickFilter {
+  margin: 0;
+  padding: 0;
+}
+
+.selectColumn {
+  color: white;
```
aninde, line 1592: This line makes this element invisible when its opened in jsfiddle by "Edit" button. 
Screenshot 2023-04-25 at 14 09 25

docs/content/guides/columns/column-filter.md
```diff
@@ -155,765 +449,3045 @@ import 'handsontable/dist/handsontable.full.min.css';
...
+      data: 'color',
+      source: ['White', 'Black', 'Yellow', 'Blue', 'Green'],
+      className: 'htCenter',
+    },
+  ],
+  // enable filtering
+  filters: true,
+  // enable the column menu
+  dropdownMenu: true,
+  height: 'auto',
```
aninde, line 1016: I would change it to defined, otherwise it would be posible to reproduce a truncated bug of autocomplete cell.

aninde (changes requested):

docs/content/guides/columns/column-filter.md
kirszenbaum: Ok, changing to `142`.

docs/content/guides/columns/column-filter.md
```diff
@@ -155,765 +449,3044 @@ import 'handsontable/dist/handsontable.full.min.css';
...
+## Exclude rows from filtering
+
+You can exclude any number of top or bottom rows from filtering.
+
+In the following demo, the first and the last row are [frozen](@/guides/rows/row-freezing.md), and
+filtering doesn't affect them.
+
+::: only-for javascript
+
+::: example #exampleExcludeRowsFromFiltering --html 1 --js 2
```
krzysztofspilka, line 2335: The bottom frozen row doesn't make sense here. Let's at least add 'minSpareRows: 1' there. https://jsfiddle.net/zmg8r0yp/1/

krzysztofspilka (changes requested):

docs/content/guides/columns/column-filter.md
```diff
@@ -155,765 +450,2860 @@ import 'handsontable/dist/handsontable.full.min.css';
...
-      debounceFn(colIndex, event);
-    });
-  };
+:::
 
-  //  Build elements which will be displayed in header.
-  const getInitializedElements = colIndex => {
-    const div = document.createElement('div');
-    const input = document.createElement('input');
+## Filter different types of data
```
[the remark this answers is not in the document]
aninde: +1 for this change. After the fix, we will avoid the user reproducing truncated header bug https://github.com/handsontable/dev-handsontable/issues/954.

docs/content/guides/columns/column-filter.md
aninde: When considering the accessibility of documentation, I would suggest increasing the low contrast of gray text by changing it based on the theme color. 

```
.selectColumn {
  color: black;
  @media screen and (prefers-color-scheme: dark) {
  color: white;
}

```
This has worked well for me on Jfiddle, where the color was adjusted based on my system settings.

aninde: Nevermind my proposition, it doesn't work in documentation. We can stay with `gray`.

aninde (approved):
Found out that `mix-blend-mode` is supported by jsfiddle https://jsfiddle.net/a5xgkt8r/1/, so it was restored and https://github.com/handsontable/handsontable/pull/10309#discussion_r1176427411 is outdated.

PR #10315, opened 2023-03-29
Docs: Further changes to the "Rows sorting" guide

kirszenbaum:
This PR applies [Krzysiek's suggestions](https://handsoncode.slack.com/archives/C01KNSJ4M08/p1679674526441989?thread_ts=1679611815.896619&cid=C01KNSJ4M08) about further changes to the [Rows sorting](https://handsontable.com/docs/react-data-grid/rows-sorting/) guide:
- Moves sections "Use sorting hooks" and "Exclude rows from sorting" outside of the "Control sorting programmatically" section
- Adds `stretchH = 'all'` to all demos
- In the "Exclude rows from sorting" demo (http://localhost:8088/docs/javascript-data-grid/rows-sorting/#exclude-rows-from-sorting):
  - Makes the first row read-only
  - For React, replaces `if (row === 15)` with `if (row === lastRowIndex)`
- Changes the date/time data and formats (as discussed [here](https://handsoncode.slack.com/archives/C048MFWU8J2/p1680522260743829))
- Changes the naming convention of the examples' IDs, from numbers (`example1`, `example2`) to descriptive names (`exampleSortByMultipleColumns`, `exampleSortByAPI`). It makes it easier to add new demos and change the demos' order.
- Makes minor changes to code formatting

[skip changelog]


AMBudnik (changes requested):
Hi @kirszenbaum 

Here's some feedback

### Case 1
We should add `multiColumnSorting.clearSort()` to the code snippet to keep consistency with what we have in the single-column sorting.

### Case 2
There is something odd with the appearance of demos in different browsers. I've started with Chrome (no surprise) and got this (missing bottom border) for the second demo

Chrome


So I've checked that in Firefox and Opera. All the browsers show a different result. 

Firefox
Firefox

Opera (only one correct)
Firefox


So I recommend changing how the 2nd demo displays on Chrome. And for the other demos, I will proceed with another issue report (as that seems to be a global case).

### Case 3 
I also have one request, could we add the `return false` line for the hook?

before
before


after
after

### Review
I would like point 2 for the 2nd demo in Chrome to be fully visible. The other two ideas are left for your consideration.

kirszenbaum:
Thanks @AMBudnik!

I'm applying points 1 and 3 here: [8913264](https://github.com/handsontable/handsontable/pull/10315/commits/89132649ca4b7acd9d8514857249df9760c5ddfa).

Re: point 2, I'm not able to replicate this issue on Chrome or Safari.

@aninde can you please try to replicate the following issue on different browsers?
1. From `./docs`, run `npm run docs:review b8fafa9131f3ced753265c33e00f36425004ff74`.
2. Go to http://localhost:8080/docs/react-data-grid/rows-sorting/#enable-sorting.
3. Check if the demo looks like this (it lacks the bottom border):
image

aninde:
@kirszenbaum I can't see it on my Chrome browser, but I can reproduce the issue using BrowserStack. It only happens in one demo, not in a similar one with the same header settings. I can assist with testing, perhaps setting an absolute height could solve the problem. If you don't have access to BrowserStack, I'm willing to help test.

kirszenbaum:
> @kirszenbaum I can't see it on my Chrome browser, but I can reproduce the issue using BrowserStack. It only happens in one demo, not in a similar one with the same header settings. I can assist with testing, perhaps setting an absolute height could solve the problem. If you don't have access to BrowserStack, I'm willing to help test.

@aninde – thank you! Can you test this again? I changed height of the problematic **Enable sorting** demo from `'auto'` to `168`. It still looks OK on my side.

If the problem persists, though, I'll create a separate issue for that.

aninde:
@kirszenbaum now it looks great on my local machine. On Browserstack, where I didn't saw the lover border before, I can see it now. Also time cells after opening editor validate correctly.

Screenshot 2023-04-04 at 17 28 30

kirszenbaum:
Thanks @aninde! @AMBudnik all 3 cases from your [comment](https://github.com/handsontable/handsontable/pull/10315#pullrequestreview-1366770953) have been addressed now.

AMBudnik (approved):
Now I get the same output as @aninde for the previously mentioned demo

Zrzut ekranu 2023-04-6 o 09 43 00


Thank you for adding my other tips as well.

Issue #442, opened 2013-02-25
Allow to customize context menu labels


pedroteixeira:
this is very simple feature request to allow for i18n :)

PR #10317, opened 2023-03-30
Fixed an issue where Handsontable's dates didn't sync correctly with HyperFormula's dates

wszymanski:
### Context

There are performed changes which lead to doing conversion from date format set on HOT (custom property set by `dateFormat` key) side to default HF date format (`DD/MM/YYYY`). Dates stored in specific format on HF side are properly read on HOT side (for example, for purpose of doing validation - dates in HOT format are handled).

Note: I'm doing rewriting by callback to `afterCellMetaReset` hooks, because some hooks, such as `afterLoadData` (originally initializing HF dataset) doesn't have information about composed cell properties. Please keep in mind that extra `setSheetContent` may has as an effect extra call of HF's `valuesUpdated` hook.

I'm using `'` sign to control whether date should be parsed. For example, date in format `MM/DD/YYYY` such `13/12/2022` considered on HOT side as incorrect should be stored on HF side as simple string, without parsing (by default sending such a string to HF's engine will cause interpreting it as a date). Otherwise, we will have mismatching types - simple string on HOT side and fully working date on HF side. This can lead to side effects.

Thus, for HOT formatting `MM/DD/YYYY` and `13/12/2022` value, HF will store a little bit changed value and return `'13.12.2022` (with apostrophe at the beginning) from [`getCellSerialized` method](https://hyperformula.handsontable.com/api/classes/hyperformula.html#getcellserialized) call.

Note: After discussion with @Serrrek89 and @adamu-handsontable we've decided that user which will manipulate using `leapYear1900` and `nullDate` HF's configuration options will get warn that it's not supported on HOT side (due limitations of DatePicker, validation and other). I've changed also default HF settings (added for unknown reason) which haven't been compatible with configuration which is [described in a documentation](https://hyperformula.handsontable.com/).

### How has this been tested?

Manual tests of date cell by manually setting cell values or from the `Datepicker`. I've checked dates before and after `1900` year, especially dates adjacent to `February 28, 1900` [due leap year bug](https://learn.microsoft.com/en-us/office/troubleshoot/excel/wrongly-assumes-1900-is-leap-year). Handsontable should work compatible with HyperFormula. Some tests using `formulasPlugin.engine.getSheetValues(0)` and `formulasPlugin.engine.getSheetSerialized(0)` have been performed and even added as e2e tests.

### Types of changes

### Related issue(s):
1. [#442](https://github.com/handsontable/dev-handsontable/issues/442)

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:


wszymanski:
I've found that HyperFormula has [`leapYear1900` option](https://hyperformula.handsontable.com/api/interfaces/configparams.html#date-and-time). This PR should handle this feature properly.

handsontable/src/plugins/formulas/engine/register.js
```diff
@@ -84,6 +86,13 @@ export function setupEngine(hotInstance) {
       });
     }
 
+    if (engineConfigItem.getConfig().leapYear1900 !== DEFAULT_SETTINGS.leapYear1900
+      || isObjectEqual(engineConfigItem.getConfig().nullDate, DEFAULT_SETTINGS.nullDate) === false) {
+      warn(toSingleLine`Dates stored in HyperFormula engine may don't cooperate properly with Handsontable when HF's 
+        configuration options such as \`leapYear1900\` and \`nullDate\` have been changed. Please consider using 
+        default configuration values for those keys.`);
```
budnix, line 93: Worth to ask @kirszenbaum for a review of this warn.

handsontable/src/plugins/formulas/formulas.js
```diff
@@ -614,7 +644,35 @@ export class Formulas extends BasePlugin {
       return false;
     }
 
-    return this.engine.getFillRangeData(engineSourceRange, engineTargetRange);
+    const fillRangeData = this.engine.getFillRangeData(engineSourceRange, engineTargetRange);
+    const [sourceStartRow, sourceStartColumn] = [engineSourceRange.start.row, engineSourceRange.start.col];
+    const [sourceEndRow, sourceEndColumn] = [engineSourceRange.end.row, engineSourceRange.end.col];
+    const [populationRowLength, populationColumnLength] = [sourceEndRow - sourceStartRow + 1,
+      sourceEndColumn - sourceStartColumn + 1];
```
budnix, line 651: I'd simplify the code to use object destructing and just simple variable assigning. Now, the code looks overcomplicated.
```suggestion
    const {
      row: sourceStartRow,
      col: sourceStartColumn,
    } = engineSourceRange.start;
    const {
      row: sourceEndRow,
      col: sourceEndColumn,
    } = engineSourceRange.end;
    const populationRowLength = sourceEndRow - sourceStartRow + 1;
    const populationColumnLength = sourceEndColumn - sourceStartColumn + 1;
```

handsontable/src/plugins/formulas/formulas.js
```diff
@@ -614,7 +644,35 @@ export class Formulas extends BasePlugin {
...
+
+    for (let populatedRowIndex = 0; populatedRowIndex < fillRangeData.length; populatedRowIndex += 1) {
+      for (let populatedColumnIndex = 0; populatedColumnIndex < fillRangeData[populatedRowIndex].length;
+        populatedColumnIndex += 1) {
+        const populatedValue = fillRangeData[populatedRowIndex][populatedColumnIndex];
+        const [sourceRow, sourceColumn] = [populatedRowIndex % populationRowLength,
+          populatedColumnIndex % populationColumnLength];
```
budnix, line 658: Why not use a basic variable assignment? Creating an array and in the same line destructing it. I think it can create more confusion than it's worth.
```suggestion
        const sourceRow = populatedRowIndex % populationRowLength;
        const sourceColumn = populatedColumnIndex % populationColumnLength;
```

handsontable/src/plugins/formulas/utils.js
```diff
@@ -17,3 +21,74 @@ export function isEscapedFormulaExpression(expression) {
...
   return isEscapedFormulaExpression(expression) ? expression.substr(1) : expression;
 }
+
+/**
+ * Checks whether string looks like formula or not. Corresponds to {@link https://hyperformula.handsontable.com/api/globals.html#isformula|HyperFormula's implementation}.
+ *
+ * @param {string} value Checked value.
+ * @returns {boolean}
+ */
+export function isFormula(value) {
```
budnix, line 31: Can I ask to cover the `isFormula` util function with tests?

handsontable/src/plugins/formulas/utils.js
```diff
@@ -17,3 +21,74 @@ export function isEscapedFormulaExpression(expression) {
...
+}
+
+/**
+ * Returns date formatted in HF's default format.
+ *
+ * @param {string} date Date formatted according to Handsontable cell date format.
+ * @param {string} dateFormat The format used for the date passed.
+ * @returns {string}
+ */
+export function getDateInHotFormat(date, dateFormat) {
```
budnix, line 75: Can I ask to cover the `getDateInHotFormat ` util function with tests?

budnix (changes requested):
The code seems to work 👌  However, I'd tweak some parts of the code and cover new utils function with tests.

handsontable/src/plugins/formulas/engine/register.js
wszymanski: @kirszenbaum Please check this :)

handsontable/src/plugins/formulas/formulas.js
wszymanski: I did it intentionally to shorten the code, but I don't mind changing it.

handsontable/src/plugins/formulas/formulas.js
wszymanski: I did it intentionally to shorten the code, but I don't mind changing it.

handsontable/src/plugins/formulas/utils.js
wszymanski: Added init tests within: https://github.com/handsontable/handsontable/pull/10317/commits/f8de204b58316cd6649a69e8ea44f6685c3d8aed

handsontable/src/plugins/formulas/utils.js
```diff
@@ -17,3 +21,74 @@ export function isEscapedFormulaExpression(expression) {
...
+}
+
+/**
+ * Checks if provided value is a date according to cell meta.
+ *
+ * @param {*} value Checked value.
+ * @param {string} cellType Type of a cell.
+ * @returns {boolean}
+ */
+export function isDate(value, cellType) {
```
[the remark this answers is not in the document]
wszymanski: Added init tests within: https://github.com/handsontable/handsontable/pull/10317/commits/f8de204b58316cd6649a69e8ea44f6685c3d8aed.

handsontable/src/plugins/formulas/utils.js
wszymanski: Added init tests within: https://github.com/handsontable/handsontable/pull/10317/commits/f8de204b58316cd6649a69e8ea44f6685c3d8aed.

.changelogs/10085.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed problems related to wrong synchronization of dates between Handsontable and HyperFormula",
```
kirszenbaum, line 3:
```suggestion
  "title": "Fixed an issue where Handsontable's dates didn't sync correctly with HyperFormula's dates.",
```

handsontable/src/plugins/formulas/__tests__/utils.unit.js
```diff
@@ -25,4 +31,69 @@ describe('Formulas utils', () => {
...
+  describe('getDateInHfFormat', () => {
+    it('should correctly convert from HOT to HF date', () => {
+      expect(getDateInHfFormat('13/11/2022', 'DD/MM/YYYY')).toBe('13/11/2022');
+      expect(getDateInHfFormat('11/13/2022', 'MM/DD/YYYY')).toBe('13/11/2022');
+      expect(getDateInHfFormat('13/11/2022', 'MM/DD/YYYY')).toBe('Invalid date');
+    });
+  });
+
+  describe('getDateFromExcelDate', () => {
+    it('should correctly convert dates from Excel like date to Date object', () => {
```
kirszenbaum, line 82:
```suggestion
    it('should correctly convert dates from Excel-like dates to Date objects', () => {
```

handsontable/src/plugins/formulas/engine/register.js
kirszenbaum, line 93:
```suggestion
      warn(toSingleLine`If you use HyperFormula with Handsontable, keep the default \`leapYear1900\` and \`nullDate\` settings. Otherwise, HyperFormula's dates may not sync correctly with Handsontable's dates.`);
```

handsontable/src/plugins/formulas/formulas.js
```diff
@@ -576,7 +600,13 @@ export class Formulas extends BasePlugin {
         sheet: this.sheetId,
       };
 
-      const cellValue = this.engine.getCellValue(address);
+      const cellMeta = this.hot.getCellMeta(visualRow, visualColumn);
+      let cellValue = this.engine.getCellValue(address); // Date as an integer (Excel like date).
```
kirszenbaum, line 604:
```suggestion
      let cellValue = this.engine.getCellValue(address); // Date as an integer (Excel-like date).
```

handsontable/src/plugins/formulas/formulas.js
```diff
@@ -576,7 +600,13 @@ export class Formulas extends BasePlugin {
...
+
+      if (cellMeta.type === 'date' && isNumeric(cellValue)) {
+        // Converting numeric value being Excel like date to JS Date object.
```
kirszenbaum, line 607:
```suggestion
        // Converting numeric values that are Excel-like dates to JS Date objects.
```

handsontable/src/plugins/formulas/formulas.js
```diff
@@ -711,7 +813,13 @@ export class Formulas extends BasePlugin {
       col: this.toPhysicalColumnPosition(column),
       sheet: this.sheetId
     };
-    const cellValue = this.engine.getCellValue(address);
+    let cellValue = this.engine.getCellValue(address); // Date as an integer (Excel like date).
+    const cellMeta = this.hot.getCellMeta(row, column);
+
+    if (cellMeta.type === 'date' && isNumeric(cellValue)) {
+      // Converting numeric value being Excel like date to JS Date object.
```
kirszenbaum, line 820:
```suggestion
      // Converting numeric values that are Excel-like dates to JS Date objects.
```

handsontable/src/plugins/formulas/utils.js
```diff
@@ -17,3 +21,74 @@ export function isEscapedFormulaExpression(expression) {
...
+ * @param {string} date Date formatted according to Handsontable cell date format.
+ * @param {string} dateFormat The format used for the date passed.
+ * @returns {string}
+ */
+export function getDateInHotFormat(date, dateFormat) {
+  return moment(date, DEFAULT_DATE_FORMAT_HYPERFORMULA, true).format(dateFormat);
+}
+
+/**
+ * Converting from Excel like date to Date object.
```
kirszenbaum, line 80:
```suggestion
 * Converting Excel-like dates to Date objects.
```

kirszenbaum (changes requested):

budnix (dismissed):

budnix (approved):

kirszenbaum (approved):

deadpony:
This change breaks the correct date format parsing. 

Steps to reproduce:

1. Set both HF and HOT to MM/DD/YYYY
2. Set two cells to date type
3. Set third cell value to =DATEDIF(CELL1, CELL2, "D")
4. Change any of the two previous cell values and see #VALUE! error in the formula.

This issue arises due to a discrepancy in the HOT formulas plugin. The plugin converts the date to DD/MM/YYYY, disregarding the global/cell settings, which is the root cause of the bug. And then passes the new changeset with incorrectly parsed data to HF

adrianszymanski89:
Hi @deadpony 

I checked it with the newest version, and it seems to work fine: https://jsfiddle.net/handsoncode/kteoux8z/. Can you please check this example?

deadpony:
Hey @adrianszymanski89,

Here is a slightly different fiddle https://jsfiddle.net/pkbwd3ov/ that might help to detect the anomaly

If you start to change dates, you'll see that computation is incorrect and hyper formula thinks we're working with DD/MM/YYYY instead of the passed MM/DD/YYYY format.

Change B1 to 06/23/2024 => #VALUE!.  Change B1 to 06/07/2024 => 21 days, which is not right. Make all changes via date picker.

I suspect the problem could be caused by the settings update, or data update written in setTimeout func

adrianszymanski89:
Hi @deadpony 

I checked your example and can see a few issues in this implementation. You declared the `required` option, which isn't present in our settings so that it won't do anything. Next, the formulas configuration is not correct. `hyperformula` isn't a valid configuration option. The whole object should look like this:

>   formulas: {
>      engine: HyperFormula,
>      dateFormats: [
>          'MM/DD/YYYY'
>      ],
>      sheetName: 'Sheet1'
>   }

Then you are using `updateSettings`, but you are updating them to initial settings, which is also not the correct way. You should only define new settings here; you can see that information in our documentation here: https://handsontable.com/docs/javascript-data-grid/api/core/#updatesettings. You are also updating data to the same initial data, which might also cause problems. In general, the functionality works well. I modified your example with the correct formulas configuration and skipped updating settings and data: https://jsfiddle.net/handsoncode/po58cr9a/

adrianszymanski89:
Hi @deadpony 

Thank you for the update. I checked your new example, and I found the source of this issue. It seems that even though the date cell type is updated to the new values, HyperFormula can't recalculate the values until we reopen and close the date cell editor. Here's an updated and minified example: https://jsfiddle.net/handsoncode/pwy1nkqd/. If you reopen both date editors, the calculation will be correct.

I'll report it internally as a bug.

deadpony:
I'm glad we've identified the issue

deadpony:
@adrianszymanski89 FYI: the same behavior of ignoring the date format happens on autofill

https://jsfiddle.net/q8x9n4br/

drag any of the cells down and see that the data is prefilled in DD/MM/YYYY ignoring the MM/DD/YYYY format

adrianszymanski89:
@deadpony 

Yes, this is a regression to version 12.4.0, I added this case to already reported issue.

deadpony:
@adrianszymanski89 

And another one... again with dates and formulas plugin. 

Here is the fiddle: https://jsfiddle.net/tb6wvmer/

Drag (autofill) B1 to B2. You'll see that B2 shows "Invalid date" despite the cell being a text type. This happens because of the bug in the beforeAutofill hook in the formulas plugin. It takes cell meta from the wrong address. The loop always starts with row 0 col 0 (https://github.com/handsontable/handsontable/blob/develop/handsontable/src/plugins/formulas/formulas.js#L687), ignoring the actual section you've made (should be row 0 col 1 for the this.hot.getCellMeta() line https://github.com/handsontable/handsontable/blob/develop/handsontable/src/plugins/formulas/formulas.js#L693). Because in col 0, we have a date-type cell, the hook mistakenly thinks it should convert the value to date, while B2 is a text type.

adrianszymanski89:
@deadpony 

Yes, this one has also been reported internally. We will cover all cases while fixing the source of the problem.

AMBudnik:
Hi @deadpony the reported internal case that my colleague mentioned is no longer replicable while using Handsontable v14.5.0.

PR #10342, opened 2023-04-26
Implement a new focusing mechanism.

jansiegel:
### Context
This PR introduces the following changes:
1. It removes the "focusable element" logic from the Copy/Paste plugin. Copying and pasting is now based on the [`Clipboard API`](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API). (**Breaking change!**)
2. It implements a "focus manager," which makes the browser focus on the last selection's `highlight` element `TD`. This allows the screen readers to read the cell contents but creates a problem with the `fastEdit` feature (editing the cell without opening the editor first) for IME users.
3. To counterbalance that, it adds a `imeFastEdit` option (the option name is subject to change). It makes the browser focus switch from the default `TD` element to the active editor's `TEXTAREA` element (or any other element, if configured that way) after a slight delay (defaults to `50ms`). This makes _some_ screen readers read the cell contents correctly while retaining the `fastEdit` feature for IME users. (**Breaking change!**)
4. It adds some ARIA tags to be able to test if the screen readers read the cells right - more ARIA tags to be added in separate PRs.
5. It adds some and cleans up some of the test cases.
6. It adds an option to declare a config object to be passed to the e2e tests' custom matchers by attaching a `matchersConfig` prop to the `spec()` context:
    ```js
    spec().matchersConfig['matcherName'] = {
     configItem: true,
     // ...
   }
    ```
    Then, it can be accessed in the custom matchers definition as `matchersUtil.matcherConfig` (where `matchersUtil` is the first argument of the matcher factory function).
Because of the differences between jest and jasmine, this new feature is not present for the unit tests.
7. It modifies and extends the tests' `toMatchHTML` custom matcher by allowing skipping theHTML elements' attributes when comparing them to the provided string. This will be useful when comparing HTML nodes in tests not related to ARIA by skipping the ARIA tags.
    It can be configured either by providing the attributes to be kept in the comparison as the second argument, e.g.:
    ```js
    expect(TD.outerHTML).toMatchHTML('<td class="custom" style="border: 0"></td>', ['class', 'style']);
    ```
    (works both in e2e and unit tests)
    
    _or_ by defining the to-be-kept attributes in a common config object connected to the `spec()` context:
    ```js
      beforeEach(function() {
        // Matchers configuration.
        this.matchersConfig = {
          toMatchHTML: {
            keepAttributes: ['class', 'dir', 'style']
          }
        };
      });
    ```
    (works only in e2e tests)
    
    By default, all attributes are not skipped.

8. It corrects a now-failing visual test by changing how the `type` method is triggered - I'm not sure why it worked before, but I think after these changes, it makes much more sense.
9. It corrects how the visual tests' cell-targeting helper works by working on a `<0; n>` instead of `<1; n>` indexing and targeting only elements of the provided type.

### How has this been tested?
- Added, removed, and modified tests to be up to date with the new logic

### Types of changes

### Related issue(s):
- handsontable/dev-handsontable#1056

### Affected project(s):
- [] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

<sub>[skip changelog]</sub>


handsontable/test/e2e/focus/focusHandling.spec.js
```diff
@@ -0,0 +1,182 @@
...
+
+      const hot = handsontable({
+        data: Handsontable.helper.createSpreadsheetData(10, 10),
+        imeFastEdit: true,
+        editor: CustomEditor
+      });
+
+      expect(document.activeElement).toEqual(document.body);
+
+      hot.getFocusManager().setRefocusElementGetter(() => hot.getActiveEditor().TEXTAREA_ALTERNATIVE);
```
budnix, line 134: Just suggestion. Instead, require a developer to use another API for focusing the custom editor element (`getFocusManager()`). Maybe it would be better that the focus manager would focus element by [`data-hot-input`](https://github.com/handsontable/handsontable/blob/develop/handsontable/src/editors/textEditor/textEditor.js#L220) attribute that has been required for some time by all editors including custom ones.

handsontable/test/e2e/focus/focusManager.spec.js
budnix, line 1: Can I suggest to move that test file to handsontable/src/\_\_tests\_\_/focusManager.spec.js?

handsontable/types/focusManager.d.ts
```diff
@@ -0,0 +1,12 @@
+type FocusMode = 'cell' | 'mixed';
+
+export interface FocusManager {
```
budnix, line 3: The class is exported using default export and the type is exported using named export. It's worth to unify the export styles. I'd prefer named exports.

handsontable/src/tableView.js
```diff
@@ -250,6 +250,10 @@ class TableView {
     priv.table = rootDocument.createElement('TABLE');
     addClass(priv.table, 'htCore');
 
+    priv.table.ariaRowCount = this.settings?.data?.length;
+    priv.table.ariaColCount = this.settings?.data[0]?.length;
```
budnix, line 254: I guess the values won't be updated after dataset length change. Is that is ok? Other question why not to use regular API such as `countRows`/`countCols`? I haven't run the code yet but it seems that it won't work for Array of Objects 🤔 

Can I ask to write tests for that aria tags?

budnix (changes requested):

handsontable/src/editors/textEditor/textEditor.js
```diff
@@ -220,6 +225,8 @@ export class TextEditor extends BaseEditor {
     this.TEXTAREA.setAttribute('data-hot-input', ''); // Makes the element recognizable by Hot as its own component's element.
     this.TEXTAREA.tabIndex = -1;
 
+    this.TEXTAREA.setAttribute('aria-hidden', 'true');
```
budnix, line 228:
```suggestion
    this.TEXTAREA.setAttribute('data-hot-input', ''); // Makes the element recognizable by Hot as its own component's element.
    this.TEXTAREA.setAttribute('aria-hidden', 'true');
    this.TEXTAREA.tabIndex = -1;
```
WDYT about grouping that two lines that changes attributes of the element?

handsontable/src/editors/textEditor/textEditor.js
```diff
@@ -387,26 +394,13 @@ export class TextEditor extends BaseEditor {
...
-
-    this.addHook('afterColumnResize', () => {
-      this.refreshDimensions();
-      this.focus();
-    });
-
-    this.addHook('afterRowResize', () => {
-      this.refreshDimensions();
-      this.focus();
-    });
```
budnix, line 409: I thing this ones need to be reverted. Otherwise there is a bug:
Kapture 2023-05-10 at 09 20 20
As the tests pass after that regression can I ask to add some that would cover that feature?

jansiegel: 👍  I'm actually glad you found it 😆 because I was not sure if these callbacks were needed, but couldn't find a way to reproduce them myself (and as you said, the tests passed regardless)

budnix:
It seems that on Safari the copy from the context menu does not work, and there are some "unwanted" borders applied by the browser that I guess should not be there.
Kapture 2023-05-10 at 09 36 42

budnix:
The reader does not read the fixed cells when the viewport is scrolled. The red border that points to the cell also does not move. I think it's worth checking that.

https://github.com/handsontable/handsontable/assets/571316/cff0cd38-a49e-46be-bcfe-1f49311117c4

budnix:
> To counterbalance that, it adds a imeFastEdit option (the option name is subject to change). It makes the browser focus switch from the default TD element to the active editor's TEXTAREA element (or any other element, if configured that way) after a slight delay (defaults to 50ms). This makes some screen readers read the cell contents correctly while retaining the fastEdit feature for IME users.

Worth to mention that point 3 (maybe point 1) introduces a braking change 🚨

handsontable/src/3rdparty/walkontable/src/renderer/table.js
```diff
@@ -275,6 +275,9 @@ export default class TableRenderer {
         const sourceRowIndex = this.renderedRowToSource(visibleRowIndex);
         const rowHeight = this.rowUtils.getHeight(sourceRowIndex);
 
+        TR.ariaRowIndex = `${sourceRowIndex}`;
```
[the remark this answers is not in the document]
jansiegel: It should work either way, but the specification states that it should be a string containing a number, not just a number: https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaRowIndex#value.
WDYT?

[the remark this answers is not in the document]
budnix: I'd say that the browser will cast the value to a string by itself. I'm concerned that using the template literal here could be replaced with `concat()` by the transpiler which finally can hurt the performance. The [W3C Aria aria-rowindex](https://w3c.github.io/aria/#aria-rowindex) doesn't say anything about that requirement.

jansiegel:
> It seems that on Safari 16.4 (18615.1.26.11.23) the copy from the context menu does not work

After a discussion with @krzysztofspilka, we concluded that this problem could possibly be released as-is and fixed in later iterations if no good fix was found at present. However, before leaving it here, I will explore some ideas and potential solutions.

jansiegel:
@budnix, could you re-review with https://github.com/handsontable/handsontable/pull/10342#issuecomment-1547733713 and Tuesday's discussions in mind?

handsontable/src/focusManager.js
```diff
@@ -0,0 +1,217 @@
...
+/**
+ * Manages the browser's focus in the table.
+ */
+export class FocusManager {
+  /**
+   * The Handsontable instance.
+   *
+   * @private
+   */
+  #hot;
```
budnix, line 25:
```suggestion
  /**
   * The Handsontable instance.
   */
  #hot;
```
If the field is declared as private then it won't show on Docs. The additional JSDoc marker is redundant.

handsontable/src/focusManager.js
```diff
@@ -0,0 +1,217 @@
...
+   */
+  #refocusDelay = 50;
+  /**
+   * Getter function for the element to be used when refocusing the browser after a delay. If `null`, the active
+   * editor's `TEXTAREA` element will be used.
+   *
+   * @type {null|Function}
+   * @private
+   */
+  #refocusElementGetter = null;
```
budnix, line 53:
```suggestion
   * @type {string}
   */
  #focusMode;
  /**
   * The delay after which the focus switches from the lastly selected cell to the active editor's `TEXTAREA`
   * element if the focus mode is set to 'mixed'.
   *
   * @type {number}
   */
  #refocusDelay = 50;
  /**
   * Getter function for the element to be used when refocusing the browser after a delay. If `null`, the active
   * editor's `TEXTAREA` element will be used.
   *
   * @type {null|Function}
   */
  #refocusElementGetter = null;
```

handsontable/src/helpers/dom/element.js
```diff
@@ -37,6 +37,19 @@ export function getParent(element, level = 0) {
...
 }
 
+/**
+ * Check if the provided element is a child of the provided Handsontable container.
+ *
+ * @param {HTMLElement} element Element to be analyzed.
+ * @param {HTMLElement} thisHotContainer The Handsontable container.
+ * @returns {boolean}
+ */
+export function isThisHotChild(element, thisHotContainer) {
```
budnix, line 47: It's worth covering the helper with a test.

handsontable/types/settings.d.ts
```diff
@@ -144,6 +144,7 @@ export interface GridSettings extends Events {
   hiddenColumns?: HiddenColumnsSettings;
   hiddenRows?: HiddenRowsSettings;
   invalidCellClassName?: string;
+  imeFastEdit?: boolean;
```
budnix, line 147: There is a need to add the new option to the Angular wrapper https://github.com/handsontable/handsontable/blob/develop/wrappers/angular/projects/hot-table/src/lib/hot-table.component.ts#L35.

handsontable/src/tableView.js
```diff
@@ -250,6 +250,8 @@ class TableView {
     priv.table = rootDocument.createElement('TABLE');
     addClass(priv.table, 'htCore');
 
+    priv.table.setAttribute('role', 'treegrid');
```
budnix, line 253: It shouldn't be applied to all tables. Including that in overlays?

budnix (changes requested):
As I reviewed this some time ago, I will list the same bugs I found last time.

 * On Safari, there is an unwanted border around the last selected cell;
Kapture 2023-07-10 at 14 21 15
 * On Safari, copy-paste does not work from the top menu (we agreed to skip it);
 * On Safari, copy does not work from the context menu. For this one, I don't remember the agreement state. It seems to be a "big" feature.
 
 Voice Over (Safari):
 * The reader reads the number of rows and columns from overlay size, not from the main table;
 * The cell is not read when the selection moves from header (`navigableHeaders: true`) to cell. The cell A5 is not read;
Kapture 2023-07-10 at 14 31 31

Jaws:
 * It reads the row/column index counting from the corner. It was said "Column header, column 2" but in the context of the table, it sounds weird. I'd say "column 1";
image

krzysztofspilka:
In the final QA, take into account the checklist at [539](https://github.com/handsontable/dev-handsontable/issues/539)

jansiegel:
> In the final QA, take into account the checklist at [539](https://github.com/handsontable/dev-handsontable/issues/539)

This PR is meant to modify the way the table focuses on the cells, not add ARIA tags, [as stated in the description](https://github.com/handsontable/handsontable/pull/10342#issue-1684926154):
> It adds some ARIA tags to be able to test if the screen readers read the cells right - more ARIA tags to be added in separate PRs.

The ARIA tags were supposed to be added in a PR under https://github.com/handsontable/dev-handsontable/issues/539, but that issue is now closed. Should it be reopened?

jansiegel:
> * On Safari, there is an unwanted border around the last selected cell;
>   Kapture 2023-07-10 at 14 21 15

Should be fixed in 8d2a124798b1223162982567fad046ca35fa6005.

image
<br><br>

> * On Safari, copy-paste does not work from the top menu (we agreed to skip it);
> * On Safari, copy does not work from the context menu. For this one, I don't remember the agreement state. It seems to be a "big" feature.

I have no idea how did I miss it before. It clearly is a big feature.
The problem stems from the same reasons as the copy/cut from the top menu, so, reluctantly, I added a workaround for Safari in 8d2a124798b1223162982567fad046ca35fa6005 to make both of them work.

After these changes every browser except Safari works the same as it did without them, and copy/cut from both the Handsontable's context menu and the browser menu should be available in every browser.

> Voice Over (Safari):
> 
> * The reader reads the number of rows and columns from overlay size, not from the main table;

This should be potentially fixed by implementing ARIA tags, so if you don't mind, I'll work on it in a separate PR.

> * The cell is not read when the selection moves from header (`navigableHeaders: true`) to cell. The cell A5 is not read;
>   Kapture 2023-07-10 at 14 31 31

Should be fixed in 81446b05e2cf2621eabda6b43c7e11cdcbd870d0.

https://github.com/handsontable/handsontable/assets/7397134/dcf6a41b-8b49-4b11-8fe5-2807236bb1d4

> Jaws:
> 
> * It reads the row/column index counting from the corner. It was said "Column header, column 2" but in the context of the table, it sounds weird. I'd say "column 1";
>   image

This should be potentially fixed by implementing ARIA tags, so if you don't mind, I'll work on it in a separate PR.

> All readers:
> 
> * For nested columns reading, the cell does not work for navigation from right to left (the red border does not move).
>   Kapture 2023-07-10 at 14 48 37

Should be fixed in 65c6a6bef863fd30af2166cc446e83b9b52dd635.


https://github.com/handsontable/handsontable/assets/7397134/0c160c3b-5c34-4fc9-a871-b11ecbb80fae

handsontable/src/pluginHooks.js
```diff
@@ -1357,6 +1357,18 @@ const REGISTERED_HOOKS = [
...
 
+  /**
+   * Fired when focusing a cell or a header element. Allows replacing the element to be focused by returning a
+   * different HTML element.
+   *
+   * @since 14.0.0
+   * @event Hooks#modifyFocusedElement
+   * @param {number} row Row index.
+   * @param {number} column Column index.
+   * @param {HTMLElement} focusedElement The element to be focused.
```
budnix, line 1368: Missing docs about what the hook can return (HTMLElement or undefined).

handsontable/src/plugins/copyPaste/copyPaste.js
```diff
@@ -750,47 +738,51 @@ export class CopyPaste extends BasePlugin {
...
+   */
+  onSafariMouseLeave() {
+    if (this.hot.isListening()) {
+      const lastSelectedRange = this.hot.getSelectedRangeLast();
+      const { row: highlightRow, col: highlightColumn } = lastSelectedRange.highlight;
+      const currentlySelectedCell = this.hot.getCell(highlightRow, highlightColumn, true);
+
+      if (lastSelectedRange) {
+        makeElementContentEditableAndSelectItsContent(currentlySelectedCell);
+      }
```
budnix, line 778: Shouldn't be that? 🤔 
```suggestion
      if (currentlySelectedCell) {
        makeElementContentEditableAndSelectItsContent(currentlySelectedCell);
      }
```

handsontable/src/plugins/nestedHeaders/nestedHeaders.js
```diff
@@ -789,6 +790,19 @@ export class NestedHeaders extends BasePlugin {
...
   }
 
+  /**
+   * `modifyFocusedElement` hook callback.
+   *
+   * @param {number} row Row index.
+   * @param {number} column Column index.
+   * @returns {HTMLTableCellElement} The `TH` element to be focused.
+   */
+  onModifyFocusedElement(row, column) {
```
budnix, line 800: Use the private method here or add `@private` JSDocs tag. Otherwise, the method will be visible in the Docs.

budnix (changes requested):

budnix:
On Safari weird things happen:
 1. There is an exception when the mouse leaves the table
Kapture 2023-08-08 at 11 50 32
 2. The editor is going into edit mode when I select and deselect cells quickly
Kapture 2023-08-08 at 11 52 45
 3. After leaving the table with a mouse the VoiceOver changes the selection and starts reading the cell differently. As if the focus underneath was changed.
Kapture 2023-08-08 at 11 59 27

handsontable/src/pluginHooks.js
jansiegel: Done in https://github.com/handsontable/handsontable/pull/10342/commits/cd04737fe2df87362f40b2f431943dd173226189

handsontable/src/plugins/copyPaste/copyPaste.js
jansiegel: Yup, done in https://github.com/handsontable/handsontable/pull/10342/commits/cd04737fe2df87362f40b2f431943dd173226189

handsontable/src/plugins/nestedHeaders/nestedHeaders.js
jansiegel: Done in https://github.com/handsontable/handsontable/pull/10342/commits/cd04737fe2df87362f40b2f431943dd173226189

jansiegel:
> On Safari weird things happen:
> 
> 1. There is an exception when the mouse leaves the table

This should be fixed in https://github.com/handsontable/handsontable/pull/10342/commits/cd04737fe2df87362f40b2f431943dd173226189.

> 2. The editor is going into edit mode when I select and deselect cells quickly

Also should be fixed in https://github.com/handsontable/handsontable/pull/10342/commits/cd04737fe2df87362f40b2f431943dd173226189.

> 3. After leaving the table with a mouse the VoiceOver changes the selection and starts reading the cell differently. As if the focus underneath was changed.

This one's a problem, unfortunately. With the Safari workaround, the `TD`'s content is selected when the mouse cursor leaves the table, which triggers VoiceOver to read that something was unselected and another thing was selected. (It usually doesn't read it correctly, too).
This should be prevented by temporarily adding `aria-hidden` to the cell, but unfortunately, it seems to be ignored by Safari and VoiceOver. I've tested it in Chrome and it seems to work fine, but in Safari the selection is still picked up by VoiceOver.
I think it might be connected to a bug in Safari https://bugs.webkit.org/show_bug.cgi?id=201887.

I've added the `aria-hidden` tag logic anyway in case the bug gets fixed, but for now, I don't have any idea how to prevent this. Should we keep it as is or think of another workaround for the copy/paste issue?

> 
> On Chrome (Win 11, NVDA):
> 
> 1. [bug] The cell that is hovered is read;

This one seems to be a feature of NVDA 😆 If you hover the cursor over the cells in the `Example` section of this page: https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/table/ it also reads its content.

budnix (approved):

Issue #1235, opened 2013-12-27
Can't put a space into a cell when a checkbox is present in the table


ghost:
If i have a column with checkbox (let's say it's the first one) and i have a column with text (the 2nd column).

I can check / uncheck the box when the selection contains my checkbox column - using the space from the keyboard.
But if I select one (alone) cell from the 2nd column, i can't add a space into it. (Nor either if i select the entire 2nd column).

I just did a correction and i'm going to do the PR right now. However i do not have grunt and don't know how to make a testcase for jasmine, sorry.

PR #10353, opened 2023-05-09
Docs: Add info on incompatibility between formulas and `getSourceData()`

kirszenbaum:
This PR stems from a long discussion held by the dev team regarding [#1235](https://github.com/handsontable/dev-handsontable/issues/1235). We concluded that formulas are inherently incapable of working with the `getSourceData()` method, and we need to document that.

This PR:
- Adds a note to the [`getSourceData()`](https://handsontable.com/docs/javascript-data-grid/api/core/#getsourcedata) API ref.
- Adds a bullet point to the [Known limitations](https://handsontable.com/docs/javascript-data-grid/formula-calculation/#features) section of the [Formula calculation](https://handsontable.com/docs/javascript-data-grid/formula-calculation/) guide.

[skip changelog]

### DIFF
--- a/handsontable/src/core.js
+++ b/handsontable/src/core.js
@@ -2975,8 +2975,9 @@ export default function Core(rootElement, userSettings, rootInstanceSymbol = fal
    * __Note__: This method does not participate in data transformation. If the visual data of the table is reordered,
    * sorted or trimmed only physical indexes are correct.
    *
-   * __Note__: By design, this method doesn't support [formulas](@/guides/formulas/formula-calculation.md).
-   * `getSourceData()` operates on source data ([physical indexes](@/api/indexMapper.md)),
+   * __Note__: This method may return incorrect values for cells that contain
+   * [formulas](@/guides/formulas/formula-calculation.md). This is because `getSourceData()`
+   * operates on source data ([physical indexes](@/api/indexMapper.md)),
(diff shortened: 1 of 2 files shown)


handsontable/src/core.js
```diff
@@ -2975,6 +2975,10 @@ export default function Core(rootElement, userSettings, rootInstanceSymbol = fal
    * __Note__: This method does not participate in data transformation. If the visual data of the table is reordered,
    * sorted or trimmed only physical indexes are correct.
    *
+   * __Note__: By design, this method doesn't support [formulas](@/guides/formulas/formula-calculation.md).
```
sequba, line 2978: I'd give more details here. Even when FormulasPlugin is enabled, `getSourceData` returns the correct result for all cells except the formula cells.

Something along the lines of:
```
If you use formulas, `getSourceData` might return incorrect values for the cells that contain formulas.
```

wszymanski: I agree with @sequba.

kirszenbaum: @sequba @wszymanski ok, would this be ok to you? [650b489](https://github.com/handsontable/handsontable/pull/10353/commits/650b4895bab09c5c9ff9258b5e7b72c6ce316f64)

wszymanski (approved):

sequba (approved):

PR #10416, opened 2023-06-15
Fixed an issue with not removing the resize handler when user clicks RMB

wszymanski:
### Context

This PR adds callback for right click on resize handler. It fixes problem described within https://github.com/handsontable/dev-handsontable/issues/1312.

> The resize handler should be deactivated upon right mouse button (RMB) click by the user. Although it does not result in any technical problems, its presence primarily affects the overall user experience of the component. Currently, when the RMB is clicked and the menu is closed, the resizing handle remains visible and follows the movement of the mouse cursor.

See below a video showing the problem, mentioned in the issue.

https://github.com/handsontable/handsontable/assets/141330/65744d9d-dc7e-4066-9869-58e6149c1296

### How has this been tested?

Tested case described in the issue.

### Types of changes

### Related issue(s):
1. https://github.com/handsontable/dev-handsontable/issues/1312

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

### DIFF
--- a/handsontable/src/plugins/manualColumnResize/manualColumnResize.js
+++ b/handsontable/src/plugins/manualColumnResize/manualColumnResize.js
@@ -579,7 +579,7 @@ export class ManualColumnResize extends BasePlugin {
 
     // There is thrown "mouseover" event right after opening a context menu. This flag inform that handle
     // shouldn't be drawn just after removing it.
-    setImmediate(() => {
+    this.hot._registerImmediate(() => {
       this.skipMouseOverAction = false;
     });
   }
--- a/handsontable/src/plugins/manualColumnResize/manualColumnResize.js
+++ b/handsontable/src/plugins/manualColumnResize/manualColumnResize.js
@@ -52,7 +52,7 @@ export class ManualColumnResize extends BasePlugin {
     this.guide = rootDocument.createElement('DIV');
     this.eventManager = new EventManager(this);
     this.pressed = null;
-    this.isTriggeredbyRMB = false;
+    this.isTriggeredByRMB = false;
     this.dblclick = 0;
     this.autoresizeTimeout = null;
 
@@ -398,7 +398,7 @@ export class ManualColumnResize extends BasePlugin {
     }
 
     // A "mouseover" action is triggered right after executing "contextmenu" event. It should be ignored.
-    if (this.isTriggeredbyRMB === true) {
+    if (this.isTriggeredByRMB === true) {
       return;
     }
 
@@ -577,12 +577,12 @@ export class ManualColumnResize extends BasePlugin {
     this.hot.rootElement.removeChild(this.guide);
 
     this.pressed = false;
-    this.isTriggeredbyRMB = true;
+    this.isTriggeredByRMB = true;
 
     // There is thrown "mouseover" event right after opening a context menu. This flag inform that handle
     // shouldn't be drawn just after removing it.
     this.hot._registerImmediate(() => {
-      this.isTriggeredbyRMB = false;
+      this.isTriggeredByRMB = false;
     });
   }
 
--- a/handsontable/src/plugins/manualColumnResize/manualColumnResize.js
+++ b/handsontable/src/plugins/manualColumnResize/manualColumnResize.js
@@ -52,6 +52,7 @@ export class ManualColumnResize extends BasePlugin {
     this.guide = rootDocument.createElement('DIV');
     this.eventManager = new EventManager(this);
     this.pressed = null;
+    this.skipMouseOverAction = false;
     this.dblclick = 0;
     this.autoresizeTimeout = null;
 
@@ -396,6 +397,11 @@ export class ManualColumnResize extends BasePlugin {
       return;
     }
 
+    // A "mouseover" action is triggered right after executing "contextmenu" event. It should be ignored.
+    if (this.skipMouseOverAction === true) {
+      return;
+    }
+
     if (this.checkIfColumnHeader(event.target)) {
       const th = this.getClosestTHParent(event.target);
 
@@ -561,6 +567,24 @@ export class ManualColumnResize extends BasePlugin {
   }
 
   /**
+   * Callback for "contextmenu" event triggered on element showing move handle. It removes handle and guide elements.
+   */
+  onContextMenu() {
+    this.hideHandleAndGuide();
+    this.hot.rootElement.removeChild(this.handle);
+    this.hot.rootElement.removeChild(this.guide);
+
+    this.pressed = false;
+    this.skipMouseOverAction = true;
+
+    // There is thrown "mouseover" event right after opening a context menu. This flag inform that handle
+    // shouldn't be drawn just after removing it.
+    setImmediate(() => {
+      this.skipMouseOverAction = false;
+    });
+  }
+
+  /**
    * Binds the mouse events.
    *
    * @private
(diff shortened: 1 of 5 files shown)


handsontable/src/plugins/manualColumnResize/manualColumnResize.js
```diff
@@ -396,6 +397,11 @@ export class ManualColumnResize extends BasePlugin {
       return;
     }
 
+    // A "mouseover" action is triggered right after executing "contextmenu" event. It should be ignored.
+    if (this.skipMouseOverAction === true) {
```
budnix, line 401: WDYT to rename that "fuse" to `this.isTriggeredByContextMenu` or something? The name `skipMouseOverAction` does not say why the action is skipped. The name seems too generic, requiring looking for the code that uses that flag to get the point.

handsontable/src/plugins/manualColumnResize/manualColumnResize.js
```diff
@@ -560,6 +566,24 @@ export class ManualColumnResize extends BasePlugin {
     }
   }
 
+  /**
+   * Callback for "contextmenu" event triggered on element showing move handle. It removes handle and guide elements.
```
budnix, line 570: Please mark the method with the `@private` JSDoc tag. Otherwise, the method will be generated in Docs, and I guess it shouldn't.

handsontable/src/plugins/manualColumnResize/manualColumnResize.js
```diff
@@ -560,6 +566,24 @@ export class ManualColumnResize extends BasePlugin {
...
+    this.hot.rootElement.removeChild(this.guide);
+
+    this.pressed = false;
+    this.skipMouseOverAction = true;
+
+    // There is thrown "mouseover" event right after opening a context menu. This flag inform that handle
+    // shouldn't be drawn just after removing it.
+    setImmediate(() => {
+      this.skipMouseOverAction = false;
+    });
```
budnix, line 584:
```suggestion
    this.hot._registerImmediate(() => {
      this.skipMouseOverAction = false;
    });
```

handsontable/src/plugins/manualColumnResize/manualColumnResize.js
```diff
@@ -560,6 +566,24 @@ export class ManualColumnResize extends BasePlugin {
     }
   }
 
+  /**
+   * Callback for "contextmenu" event triggered on element showing move handle. It removes handle and guide elements.
+   */
+  onContextMenu() {
+    this.hideHandleAndGuide();
+    this.hot.rootElement.removeChild(this.handle);
+    this.hot.rootElement.removeChild(this.guide);
```
budnix, line 575:
```suggestion
```
I think that it is ok that the handler does not disappear after triggering the context menu. The origin bug is still fixed, the handler does not follow the mouse after unclicking the context menu:
Kapture 2023-06-16 at 10 22 56

By removing these lines we do not add new code to maintain and the fix still seems to be working. So way not do it 🤔

budnix (changes requested):
The fix seems to be working. Can you apply the same fix for the _ManualRowResize_ plugin?

handsontable/src/plugins/manualColumnResize/manualColumnResize.js
wszymanski: As you can see, there UI element left behind. Title of the issue says that: "The resize handler should be removed when user clicks RMB".

budnix: For better code health, I'd remove the lines. UI, in this case, works fine for me. What do you think @AMBudnik @krzysztofspilka?

handsontable/src/plugins/manualColumnResize/manualColumnResize.js
wszymanski: Done within https://github.com/handsontable/handsontable/pull/10416/commits/b7ca869936281bac45ddc7dbb33da23193f5b022.

handsontable/src/plugins/manualColumnResize/manualColumnResize.js
wszymanski: Done within https://github.com/handsontable/handsontable/pull/10416/commits/b7ca869936281bac45ddc7dbb33da23193f5b022.

handsontable/src/plugins/manualColumnResize/manualColumnResize.js
budnix: To clarify. Current PR fixes the issue by hiding/removing the handler right after the RMB is clicked. I believe it's a bad UI/UX design that some elements just disappear after mouse-clicking. 
Kapture 2023-06-16 at 12 34 01

My proposition is not to hide/remove the handler and make it just inactive for RMB.
Kapture 2023-06-16 at 12 36 29

handsontable/src/plugins/manualColumnResize/manualColumnResize.js
budnix: Thanks! Can I ask one more time for a fix? There is a typo `this.isTriggeredbyRMB` should be `this.isTriggeredByRMB`.

handsontable/src/plugins/manualColumnResize/manualColumnResize.js
budnix: Thanks! Can I ask one more time for a fix? There is a typo `this.isTriggeredbyRMB` should be `this.isTriggeredByRMB`.

handsontable/src/plugins/manualColumnResize/manualColumnResize.js
AMBudnik: It's fine with me 👍

handsontable/src/plugins/manualColumnResize/manualColumnResize.js
wszymanski: Done within https://github.com/handsontable/handsontable/pull/10416/commits/edf886b43287eb0b784dda37710adeb4dab8f084.

wszymanski:
> The fix seems to be working. Can you apply the same fix for the _ManualRowResize_ plugin?

So far changes are done without remark from: https://github.com/handsontable/handsontable/pull/10416#discussion_r1231945147. I'll do changes after decision.

handsontable/src/plugins/manualColumnResize/manualColumnResize.js
krzysztofspilka: @budnix @wszymanski @AMBudnik 
The primary role of the blue highlight over the border between the headers is to indicate (visually) the possible action for the end user. In this case, the user can use left click + drag to resize the column width manually.

That action, though, is no longer available when the context menu is opened. When it's opened, left-click just closes the menu. This makes the manual-resize indicator behind the menu completely useless, right?

For the record, Google Sheets has a similar problem, but it's more consistent in its behavior because I can still resize other columns while the menu is opened.

Handsontable's behavior after this fix will be more similar to Excel than Google Sheets.

Google Sheets
https://github.com/handsontable/handsontable/assets/8048526/3c4c93b4-4795-4535-ab3a-994555e334bb

Excel
https://github.com/handsontable/handsontable/assets/8048526/33eefbf0-afd5-48ed-8ecf-d198d5ef2f48

PR #10396, opened 2023-06-05
Update the Angular wrapper to work with Angular@16.

jansiegel:
### Context
The main changes made in this PR:
1. Since the release of [angular@16](https://www.npmjs.com/package/@angular/cli/v/16.0.0) the `@handsontable/angular` package was no longer compatible with the latest Angular release. This PR aims to change that.
    I moved the current wrapper logic to a new library template based on `@angular/cli@12`. The new wrapper build uses the Ivy compiler, and being built with `angular@12` makes it compatible with Angular versions 12 up to 16 (currently the latest release).
    
    **Unfortunately, it breaks the compatibility with Angular versions 11 and earlier.** 
    It's worth noting, that [Angular currently provides active support only for versions 14+](https://angular.io/guide/releases#actively-supported-versions).

2. Currently, the examples workflow does not test the angular-based examples properly. 
    When building the `/next/` section of the examples [only `handsontable` is being symlinked in the examples' `node_modules`](https://github.com/handsontable/handsontable/actions/runs/5070569056/jobs/9105914867#step:20:27). What's even worse, [the wrapper version being used in the build and testing process is not `latest`, but `12.0.1`.](https://github.com/handsontable/handsontable/blob/b7cca4520a7284727d9b878a9e01cb695263b0bc/examples/next/docs/angular/demo/package.json#L29)
    
   After the changes made in this PR, [both `handsontable` and `@handsontable/angular` should be linked properly in the examples build/test workflow.](https://github.com/handsontable/handsontable/actions/runs/5197296330/jobs/9372056945#step:20:18)

3. I removed the Angular 9-11 examples, as these versions are no longer compatible with the wrapper.
4. I added Angular 14-16 examples + created an `angular-next` section which builds and tests the wrapper against the version of Angular tagged as `next`.
5. While rewriting the wrapper to later versions of Angular (mostly 16 and 12), I had to make some corrections regarding the typescript typing. There are places in the wrapper logic that would benefit from a bigger refactor, but I did not make any major changes as it was not in the scope of this task.
6. For the same reason as above, I had to make some changes to Handsontable, namely, adding some missing type definitions that were not tested with the wrapper based on Angular@9.
7. I reverted the temporary changes done in the release process of `12.4.0`, which purpose was to prevent the workflows from using the latest `angular` version.
[...]

PR #10418, opened 2023-06-20
Docs: Edit 13.0.0 docs

kirszenbaum:
This PR:
- Adds a new migration guide: [Migrate from 12.4 to 13.0](http://localhost:8080/docs/javascript-data-grid/migration-from-12.4-to-13.0/)
- Edits the 13.0 changelog entries
- Adds 13.0 release notes
- Makes other minor changes

[skip changelog]

### DIFF
--- a/docs/content/guides/upgrade-and-migration/migrating-from-12.4-to-13.0.md
+++ b/docs/content/guides/upgrade-and-migration/migrating-from-12.4-to-13.0.md
@@ -71,14 +71,8 @@ You can read more about this change on [our blog](https://handsontable.com/blog/
 #### Before
 
 ```js
-// insert a row below the last row
-handsontableInstance.alter('insert_row');
-
 // insert a row above row number 10
 handsontableInstance.alter('insert_row', 10);
-
-// insert a column after the last column
-handsontableInstance.alter('insert_col');
 
 // insert a column before column number 10
 handsontableInstance.alter('insert_col', 10);
--- a/docs/content/guides/upgrade-and-migration/migrating-from-12.4-to-13.0.md
+++ b/docs/content/guides/upgrade-and-migration/migrating-from-12.4-to-13.0.md
@@ -27,7 +27,7 @@ If you're having any issues with Handsontable's [Angular wrapper](@/guides/integ
 
 ## Step 2: Stop using the `beforeAutofillInsidePopulate` hook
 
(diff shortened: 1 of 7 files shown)


kirszenbaum:
@jansiegel – mind that [here](https://github.com/handsontable/handsontable/pull/10418/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR26) I changed the Angular version do 13 and lower. That's because Angular [no longer supports v12](https://angular.io/guide/releases#actively-supported-versions) and, , according to our new frameworks support policy, we don't support it either.

On the other hand, the new CodeSandbox Angular demo uses v12: https://codesandbox.io/s/handsontable-angular-data-grid-hello-world-app-13-0-0-49m644

jansiegel:
> no longer supports v12

We _could_ say something along the lines of: "Although we don't support angular<14, the wrapper is currently built using angular@12, so it works with angular versions 12+, but it may change without notice with any non-major release.", but I don't know if we _want_ to.
I, personally, refactored the wrapper to use angular@12 because I wanted it to support as many angular versions as possible.

@adamu-handsontable, could you help us out here?

kirszenbaum:
> no longer supports v12

@jansiegel – I updated this bit, as discussed [here](https://handsoncode.slack.com/archives/CCAJN0V1V/p1687419244577019?thread_ts=1687373215.799909&cid=CCAJN0V1V).

> - **Breaking change (Angular)**: Dropped support for Angular 13 and lower. From now on, Handsontable supports only those versions of Angular that are officially supported by the Angular team: currently, it's 14-16. However, Handsontable 13.0.0 was thoroughly tested and, to the best of our knowledge, works correctly with versions down to Angular 12. [#10396](https://github.com/handsontable/handsontable/pull/10396)

docs/content/guides/upgrade-and-migration/migrating-from-12.4-to-13.0.md
```diff
@@ -0,0 +1,123 @@
...
+
+Migrate from Handsontable 12.4 to Handsontable 13.0, released on June 22, 2023.
+
+[[toc]]
+
+::: only-for javascript
+
+## Update to Angular 12 or higher
+
+Handsontable 13.0 requires Angular 12 or higher. If you're using an older version of Angular, you need to update it.
```
jansiegel, line 25: @kirszenbaum This is probably worth rephrasing as well.

kirszenbaum: Ok, done

jansiegel (approved):

PR #10301, opened 2023-03-22
Feature/dev issue 1062 part2

budnix:
### Context

Currently, the table does not support keyboard navigation in headers (only in cells). The aim of the PR is going to change that. After the changes, there should be able to navigate in row and column headers (including top and bottom corners) using currently supported keyboard shortcuts e.g: <kbd>Tab</kbd> (next cell/header), <kbd>Shift</kbd>+<kbd>Tab</kbd> (previous cell/header), Arrow keys and more.

PoC demo:
Kapture 2023-03-01 at 14 23 07

_The PR is a part of the accessibility project, and it's in progress._

### Tasks to be done (may not be complete)
- [x] Extend the [Walkontable Selection](https://github.com/handsontable/handsontable/blob/feature/dev-issue-1062/handsontable/src/3rdparty/walkontable/src/selection.js) class to support adding CSS classes to the headers (done in https://github.com/handsontable/handsontable/pull/10265);
- [ ] **wip** Extend the [Transformation class](https://github.com/handsontable/handsontable/blob/feature/dev-issue-1062/handsontable/src/selection/transformation.js) to support navigating in headers;
  - [ ] As it is a crucial module for navigation, it is necessary to write unit tests (there are currently no unit tests) for currently working cells navigation and for new navigation in headers;
- [ ] Exclude all shortcut combinations (e.g. <kbd>Shift</kbd>+<kbd>Arrow Right</kbd>) and plugins that allow extending selection or triggering actions (e.g. copy/paste cell value) on cells from the header selection;
  - [ ] Update the tests by adding new ones that check cooperation with navigation in headers feature for conflicted features/plugins;
- [ ] Add support for the _NestedHeaders_ plugin;

### How has this been tested?

\-

### Types of changes

- [x] New feature or improvement (non-breaking change which adds functionality)

### Related issue(s):
1. fixes https://github.com/handsontable/dev-handsontable/issues/1062

### Affected project(s):
- [x] `handsontable`

### Checklist:

- [x] I have reviewed the guidelines about [Contributing to Handsontable](https://github.com/handsontable/handsontable/blob/master/CONTRIBUTING.md) and I confirm that my code follows the code style of this project.
- [x] I have signed the [Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLScpMq4swMelvw3-onxC8Jl29m0fVp5hpf7d1yQVklqVjGjWGA/viewform?c=0&w=1)
- [x] My change requires a change to the documentation.

PR #10430, opened 2023-06-28
Add the ability to traverse the grid using TAB key throughout the page

budnix:
### Context

The PR adds the ability to traverse the grid using <kbd>Tab</kbd> and <kbd>Shift</kbd>+<kbd>Tab</kbd> throughout the page keeping the correct order of the native browser focus. This is a breaking change as from now on when the last cell is selected (or the last cell in a row), pressing the <kbd>Tab</kbd> key deactivates the table (the table no more listens for any keyboard shortcuts) and moves the browser's focus to the next element. 

The PR introduces a new API:
 * The new table option `disableTabNavigation` (default `false`). The option, when set as `true`, disables the table navigation. In that mode, the arrow keys are supposed to use to navigate the table. Pressing the <kbd>Tab</kbd> or <kbd>Shift</kbd>+<kbd>Tab</kbd> moves the browser's focus to the next or previous element on the page. When set as `false`, there is a possibility to navigate the table, but when the cell selection reaches the last cell (or the last cell in a row when `autoWrapRow: false`) then the table deactivates itself, and the browser's focus is moved to the next element on the page;
 * The new `modifyFocusOnTabNavigation` hook. The hook allows changing the cell selection after the table is activated by the browser's focus triggered by <kbd>Tab</kbd> or <kbd>Shift</kbd>+<kbd>Tab</kbd> keys. By default, the first visible cell (or header) is selected when the table is activated from the above element. When it's activated from the element below, the last cell is selected;
 * The Core `listen` method is improved for pages where multiple Handsontable instances are created. From now on, when one of the instances is activated, the other instance that the activation was taken from calls `'unlisten'` hook on its instance;

#### Tab navigation is disabled
```js
navigableHeaders: true,
disableTabNavigation: true
autoWrapRow: true
```
By disabling the Tab navigation (`disableTabNavigation: true`) we can focus the table by pressing <kbd>Tab</kbd> key, and by hitting the key once again, we can jump out from the table - focus the next editable element on the page. 
Kapture 2023-07-07 at 13 04 04

#### Tab navigation is enabled, and there is a row wrapping enabled
```js
navigableHeaders: true,
disableTabNavigation: false
autoWrapRow: true
```
There is a possibility to navigate the table using the <kbd>Tab</kbd> key. But once a user reaches the last cell, hitting the key moves the browser's focus to the next element.
Kapture 2023-07-07 at 13 06 40

#### Tab navigation is enabled, and there is a row wrapping disabled
```js
navigableHeaders: true,
disableTabNavigation: false
autoWrapRow: false
```
There is a possibility to navigate the table using the <kbd>Tab</kbd> key. But once a user reaches the last cell in a row (`autoWrapRow: false`), hitting the key moves the browser's focus to the next element.
Kapture 2023-07-07 at 13 09 52

#### Tab navigation is disabled as well as headers navigation, and there are some rows hidden
```js
navigableHeaders: false,
disableTabNavigation: false
autoWrapRow: false
```
When the `navigableHeaders` option is disabled, the first visible cell is selected.
Kapture 2023-07-07 at 13 19 36

#### Feature of the most recently remembered cell focus selection
```js
navigableHeaders: false,
disableTabNavigation: false
autoWrapRow: false
```
There is implemented a feature that tracks recently selected focused cell. When the table is deactivated and again activated by the TAB navigation keys, the cell selection is restored to the same position as before table deactivation. It only works for cell focus/highlight, not for multiple selection ranges.
Kapture 2023-07-07 at 13 22 06

#### Support for multiple instances with different configurations
Two different Handsontable instances act as regular editable elements where we can focus using browsers TAB navigation keys.
Kapture 2023-07-07 at 13 39 09

### How has this been tested?

I tested the changes locally with different use cases. I covered all features with E2E and Visual Tests tests.

### Types of changes

### Related issue(s):
1. fixes https://github.com/handsontable/dev-handsontable/issues/1063

### Affected project(s):
- [x] `handsontable`

### Checklist:

### DIFF
--- a/handsontable/src/core/focusCatcher/index.js
+++ b/handsontable/src/core/focusCatcher/index.js
@@ -13,7 +13,7 @@ export function installFocusCatcher(hot) {
       const mostTopStartCoords = getMostTopStartPosition(hot);
 
       if (mostTopStartCoords) {
-        hot.runHooks('modifyFocusOnTabNavigation', mostTopStartCoords);
+        hot.runHooks('modifyFocusOnTabNavigation', 'from_above', mostTopStartCoords);
         hot.selectCell(mostTopStartCoords.row, mostTopStartCoords.col);
       }
 
@@ -23,7 +23,7 @@ export function installFocusCatcher(hot) {
       const mostBottomEndCoords = getMostBottomEndPosition(hot);
 
       if (mostBottomEndCoords) {
-        hot.runHooks('modifyFocusOnTabNavigation', mostBottomEndCoords);
+        hot.runHooks('modifyFocusOnTabNavigation', 'from_below', mostBottomEndCoords);
         hot.selectCell(mostBottomEndCoords.row, mostBottomEndCoords.col);
       }
 
--- a/handsontable/src/core/focusCatcher/index.js
+++ b/handsontable/src/core/focusCatcher/index.js
@@ -96,8 +96,13 @@ function getMostTopStartPosition(hot) {
   let topRow = navigableHeaders && hot.countColHeaders() > 0 ? -hot.countColHeaders() : 0;
   let startColumn = navigableHeaders && hot.countRowHeaders() > 0 ? -hot.countRowHeaders() : 0;
 
-  topRow = topRow === 0 ? rowIndexMapper.getVisualFromRenderableIndex(topRow) : topRow;
-  startColumn = startColumn === 0 ? columnIndexMapper.getVisualFromRenderableIndex(startColumn) : startColumn;
+  if (topRow === 0) {
+    topRow = rowIndexMapper.getVisualFromRenderableIndex(topRow);
+  }
+
+  if (startColumn === 0) {
+    startColumn =  columnIndexMapper.getVisualFromRenderableIndex(startColumn);
+  }
 
   if (topRow === null || startColumn === null) {
     return null;
--- a/handsontable/src/core/focusCatcher/index.js
+++ b/handsontable/src/core/focusCatcher/index.js
@@ -1,0 +1,50 @@
+import { GRID_GROUP } from '../../shortcutContexts';
+import { installFocusDetector } from './focusDetector';
+
+/**
+ * Installs a focus catcher module. The module observes the cell focus position and depends on the
+ * position it releases the TAB navigation to the browser or not.
+ *
+ * @param {Handsontable} hot The Handsontable instance.
+ */
+export function installFocusCatcher(hot) {
+  const { activate, deactivate } = installFocusDetector(hot, {
+    onFocusFromTop() {
+      // TODO add correct start coords depends on the user settings
+      hot.selectCell(-1, -1);
+      hot.listen();
+    },
+    onFocusFromBottom() {
+      // TODO add correct end coords depends on the table size
+      hot.selectCell(0, 2);
+      hot.listen();
+    },
+  });
+
+  hot.addHook('afterListen', () => deactivate());
+  hot.addHook('afterUnlisten', () => activate());
+
+  hot.getShortcutManager()
+    .getContext('grid')
+    .addShortcut({
+      keys: [['Tab'], ['Shift', 'Tab']],
+      callback: (event) => {
+        const { row, col } = hot.getSelectedRangeLast().highlight;
+
+        if (event.shiftKey && row === -1 && col === -1 || !event.shiftKey && row === 0 && col === 2) {
+          hot.deselectCell();
+          hot.unlisten();
+
+          return false;
+        }
+
+        return true;
+      },
+      runOnlyIf: () => hot.selection.isSelected(),
+      preventDefault: false,
+      stopPropagation: false,
+      position: 'before',
+      relativeToGroup: GRID_GROUP,
+      group: 'focusCatcher',
+    });
+}
--- a/handsontable/src/core/focusCatcher/index.js
+++ b/handsontable/src/core/focusCatcher/index.js
@@ -13,6 +13,7 @@ export function installFocusCatcher(hot) {
       const mostTopStartCoords = getMostTopStartPosition(hot);
 
       if (mostTopStartCoords) {
+        hot.runHooks('modifyFocusOnTabNavigation', mostTopStartCoords);
         hot.selectCell(mostTopStartCoords.row, mostTopStartCoords.col);
       }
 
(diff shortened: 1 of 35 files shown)


handsontable/src/shortcutContexts/constants.js
```diff
@@ -1,7 +1,3 @@
-/**
- * Group name for keyboard shortcuts that are active when the header is selected.
- */
-export const HEADERS_GROUP = 'headersDefault';
```
budnix, line 4: Leftover after https://github.com/handsontable/handsontable/pull/10301.

krzysztofspilka:
@budnix As we discussed this internally, I would like to add the "memory" of the last selected cell or header to Handsontable. It's based on the W3C suggestions for managing focus in data grids:

> The convention for where focus lands in a composite when it receives focus as a result of a Tab key event depends on the type of composite. It is typically one of the following:
> - The element that had focus the last time the composite contained focus. Or, if the composite has not yet contained the focus, the first element. Widgets that usually employ this pattern include grid and tree grid.

[Source](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#keyboardnavigationinsidecomponents)

You mentioned that we might have to add a new hook to control the behavior of the grid, which is OK. One good example of using the new hook is when a user wants to skip read-only cells and focus on the first-next editable cell instead.

Header navigation disabled (default):

tab-sequence-no-header-navigation

Header navigation enabled:

tab-sequence-header-navigation

handsontable/src/core/focusCatcher/index.js
```diff
@@ -0,0 +1,136 @@
...
+  if (bottomRow < 0) {
+    bottomRow = navigableHeaders && hot.countColHeaders() > 0 ? -1 : null;
+  }
+  if (endColumn < 0) {
+    endColumn = navigableHeaders && hot.countRowHeaders() > 0 ? -1 : null;
+  }
+
+  if (bottomRow === null || endColumn === null) {
+    return null;
+  }
```
wszymanski, line 130: Up to you, for me it would be more clean.

```suggestion
  if (bottomRow < 0) {
    if (navigableHeaders === false || hot.countColHeaders() === 0) {
      return null;
    }

    bottomRow =-1;
  }

  if (endColumn < 0) {
    if (navigableHeaders === false || hot.countColHeaders() === 0) {
      return null;
    }

    endColumn = -1;
  }
```

handsontable/test/helpers/common.js
```diff
@@ -188,20 +189,19 @@ export function columnIndexMapper() {
...
+export function handsontable(options, explicitOptions = false, container = spec().$container) {
   // Add a license key to every Handsontable instance.
   if (options && !explicitOptions) {
     options.licenseKey = 'non-commercial-and-evaluation';
   }
 
-  currentSpec.$container.handsontable(options);
-  currentSpec.$container[0].focus(); // otherwise TextEditor tests do not pass in IE8
+  container.handsontable(options);
+  container[0].focus(); // otherwise TextEditor tests do not pass in IE8
```
wszymanski, line 202: Maybe this line can be removed :)?

wszymanski (commented):
Nice idea with extra `input` elements. Just minor remarks.

handsontable/src/core/focusCatcher/index.js
```diff
@@ -0,0 +1,136 @@
...
+        // For disabled `autoWrapRow` option set the row to the same position as the currently selected row.
+        if (!autoWrapRow) {
+          mostTopStartCoords.row = highlight.row;
+          mostBottomEndCoords.row = highlight.row;
+        }
+
+        if (event.shiftKey && (!isSelected || highlight.isEqual(mostTopStartCoords)) ||
+            !event.shiftKey && (!isSelected || highlight.isEqual(mostBottomEndCoords))) {
+          hot.deselectCell();
+          hot.unlisten();
```
wszymanski, line 71: As [here](https://github.com/handsontable/handsontable/pull/10430#discussion_r1258003121).

handsontable/src/core/focusCatcher/index.js
```diff
@@ -0,0 +1,136 @@
...
+  hot.getShortcutManager()
+    .getContext('grid')
+    .addShortcut({
+      keys: [['Tab'], ['Shift', 'Tab']],
+      callback: (event) => {
+        const { disableTabNavigation, autoWrapRow } = hot.getSettings();
+
+        if (disableTabNavigation) {
+          hot.deselectCell();
+          hot.unlisten();
```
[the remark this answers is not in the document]
budnix: 🤔 I don't understand the idea. Here I just use Core API to achieve some behavior. The `deselectCell` method and `unlisten` do their job separately. One deselects the cell, and the second makes the table inactive. I don't see any benefits of joining or moving them to another module.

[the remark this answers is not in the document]
wszymanski: I wrote "moving it to selection module" while I thought "moving stopping selection (conditionally, when option is enable) to selection module". I thought about benefits such as stopping calling extra hooks (selection, deselection, unlistening).

handsontable/src/core/focusCatcher/index.js
budnix: Done within https://github.com/handsontable/handsontable/pull/10430/commits/90e205db77e39d699809a908093e7e9a54f398cd.

handsontable/test/helpers/common.js
budnix: Done within https://github.com/handsontable/handsontable/pull/10430/commits/90e205db77e39d699809a908093e7e9a54f398cd.

wszymanski: Thank you ❤️

handsontable/src/core/focusCatcher/index.js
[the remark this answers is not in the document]
budnix: I don't need blocking functionality. Here I needed to deselect all selections (logically and visually) and pass the focus to the browser. The `deselectCell` method does the job I needed.

wszymanski (approved):
Please check tests before merging 👍

aninde (commented):
I checked the operation of all keyboard shortcuts  with multiple instances of Handsontable in various combinations of 
```
navigableHeaders: true / false
disableTabNavigation: true / false
autoWrapRow: true / false
```
on Safari 16.5.1 (18615.2.9.11.7), Chrome 114.0.5735.198, FF 14.0.2 and Edge 114 on macOS and Windows 10 and they work as they should. Good job @budnix

aninde (approved):

PR #10461, opened 2023-08-08
Change the behavior of selecting all cells feature

budnix:
### Context

The PR changes the behavior of selecting all cells feature. 

#### Changes:
 * Hitting the <kbd>Ctrl/Cmd</kbd>+<kbd>A</kbd> from now on select all cells without headers;
 * Selecting all cells does not move the focus position. The focus position stays in the same position;

### Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Related issue(s):
1. fixes https://github.com/handsontable/dev-handsontable/issues/1393

### Affected project(s):
- [x] `handsontable`

### Checklist:

- [x] I have reviewed the guidelines about [Contributing to Handsontable](https://github.com/handsontable/handsontable/blob/master/CONTRIBUTING.md) and I confirm that my code follows the code style of this project.
- [x] I have signed the [Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLScpMq4swMelvw3-onxC8Jl29m0fVp5hpf7d1yQVklqVjGjWGA/viewform?c=0&w=1)
- [x] My change requires a change to the documentation.

PR #10464, opened 2023-08-10
Change the behavior of selecting all cells feature

budnix:
### Context

The PR changes the behavior of selecting all cells feature. 

#### Changes:
 * Hitting the <kbd>Ctrl/Cmd</kbd>+<kbd>A</kbd> from now on select all cells without headers;
 * Selecting all cells does not move the focus position. The focus position stays in the same position;

This is a second try to implement the change. In [the first attempt](https://github.com/handsontable/handsontable/pull/10461), the logical coordinates were modified to exclude the headers, but it turned out that this brakes other features. This approach does not modify the coordinates it only excludes headers from highlighting when the "select all cells" action is triggered. 

### How has this been tested?

I tested the changes locally. All tests related to the feature were updated. 

### Types of changes

### Related issue(s):
1. fixes https://github.com/handsontable/dev-handsontable/issues/1393

### Affected project(s):
- [x] `handsontable`

### Checklist:


handsontable/src/selection/selection.js
```diff
@@ -628,13 +634,16 @@ class Selection {
    * `false` otherwise.
    * @param {boolean} [includeColumnHeaders=false] `true` If the selection should include the column
    * headers, `false` otherwise.
-   * @param {{row: number, col: number}} [focusPosition] The argument allows changing the cell/header
+   * @param {object} [options] Additional object with options.
+   * @param {{row: number, col: number}} [options.focusPosition] The argument allows changing the cell/header
```
wszymanski, line 638: Haven't we decided that newly created API should use `column` instead of `col`?

handsontable/src/selection/selection.js
```diff
@@ -628,13 +634,16 @@ class Selection {
...
    * focus position. The value takes an object with a `row` and `col` properties from -N to N, where
    * negative values point to the headers and positive values point to the cell range.
+   * @param {boolean} [options.disableHeadersHighlight] If `true`, disables highlighting the headers even when
+   * the logical coordinates points on them.
    */
-  selectAll(includeRowHeaders = false, includeColumnHeaders = false, focusPosition = {
-    row: this.tableProps.countColHeaders() > 0 ? -this.tableProps.countColHeaders() : 0,
-    col: this.tableProps.countRowHeaders() > 0 ? -this.tableProps.countRowHeaders() : 0,
+  selectAll(includeRowHeaders = false, includeColumnHeaders = false, options = {
+    focusPosition: false,
```
wszymanski, line 645: It seems that `focusPosition` may be a boolean, not described in JSDoc.

wszymanski (changes requested):
Works good, just to remarks.

handsontable/src/selection/selection.js
budnix: Not for coordinates. Those props should be compatible with CellCoords and CellRange classes.

handsontable/src/selection/selection.js
budnix: Done https://github.com/handsontable/handsontable/pull/10464/commits/dd1ddedb5f12be052ed900c581dcf1b1def993f6 👍

wszymanski (approved):

Issue #10480, opened 2023-08-28
[Bug]: Selection is not displayed when the table is in a foreign object


kumilingus:
### Describe the bug

After inserting the table into SVG `<foreignObject>`, the cell selection boxes will stop appearing. The user can edit the cell, but the `input` element is hidden (or misplaced).

An example use case is the possibility to insert tables into an SVG document and connect columns or rows using links.
https://jsfiddle.net/kumilingus/j4s8bf97/

### Video/Screenshots

Selecting the `B3` cell:
<img width="323" alt="image" src="https://github.com/handsontable/handsontable/assets/3967880/9bac736e-35c7-4f6b-b582-f17d72a0520f">

Editing the `B3` cell:
<img width="320" alt="image" src="https://github.com/handsontable/handsontable/assets/3967880/bd5f72c0-279c-43d7-b283-d750dd7999b8">

### Provide a link to the demo with the bug reproduction

https://jsfiddle.net/kumilingus/1upLrsy5/

### Handsontable version

13.0.0

### Your environment

Chrome, FF, Safari / macOS

adrianszymanski89:
Closed as a duplicate of the ZenDesk issue.

Issue #9754, opened 2022-08-08
After deleting a row in nested rows the column width jumps


Paqura:
### Description
After deleting a row in nested rows the column width jumps

### Steps to reproduce
1. Open https://handsontable.com/docs/row-parent-child/#preparing-the-data-source
2. Click to `preview`
3. Remove row

AMBudnik:
Thank you for sharing the issue report, @Paqura 
I replicated it using Chrome 103 on macOS Monterey.

It seems that the `rowHeaders` gets wider after we remove the row for the first time (afterward nothing like that happens again)

<img width="640" alt="Screenshot 2022-08-08 at 14 33 49" src="https://user-images.githubusercontent.com/10757813/183418565-5b1a94c5-4c46-4ab9-b04a-668e7bb7c942.png">


Related topics:
* https://github.com/handsontable/handsontable/issues/6006
* https://github.com/handsontable/handsontable/issues/6456

AMBudnik:
Hi @Paqura 

we just fixed this issue! The fix is awaiting in Handsontable v14.0.0

Here https://handsontable.com/docs/javascript-data-grid/row-parent-child/#preparing-the-data-source in the new demo it already works as desired.

If you would like to read more about the changes in this version please visit https://handsontable.com/docs/javascript-data-grid/release-notes/#_14-0-0

PR #10483, opened 2023-08-29
Fix offset() for elements within foreign object context

kumilingus:
### Context

After inserting the table into SVG <foreignObject>, the cell selection boxes will stop appearing. The user can edit the cell, but the input element is hidden (or misplaced).

Selecting the `B3` cell:
image

Editing the `B3` cell:
image

I managed to locate the root cause of the problem. It lies within the DOM utility `offset` that does not account for the DOM element to be inside a foreign object. The foreign object has neither `offsetLeft` nor `offsetTop` properties defined and the `offset()` method ends up returning `{ top: NaN, left: NaN }`.

An example use case is the possibility to insert tables into an SVG document and connect columns or rows using links.
https://jsfiddle.net/kumilingus/j4s8bf97/

### How has this been tested?

1. A unit test has been added.
2. A patched version of _Handsontable_ with the same demo: https://jsfiddle.net/kumilingus/fq7zyr5b/

### Types of changes

### Related issue(s):
1. https://github.com/handsontable/handsontable/issues/10480

### Affected project(s):
- [x] `handsontable`

### Checklist:


kumilingus:
It seems that your unit test environment does not allow the `offset` method to be tested? Perhaps that's the reason why there was no test for the method in the first place?

The DOM elements seem to have `offsetParent` always set to `null` and no size. Any suggestion how to test the issue?

AMBudnik:
Hi @kumilingus 

I think that at this point, the PR is ready for a review. Thank you for sharing.

budnix:
> The DOM elements seem to have `offsetParent` always set to `null` and no size. Any suggestion how to test the issue?

The `*.unit.js` files are run in a fake DOM environment (JSDOM), and the results may not be accurate. I'd suggest moving the test to the [element.spec.js](https://github.com/handsontable/handsontable/blob/develop/handsontable/src/helpers/dom/__tests__/element.spec.js) file. `*.spec.js` files are running in the real browser.

kumilingus:
Hi @budnix, I moved the test to the `element.spec.js` file, but I wasn't sure how to run it. 🤦 😄

handsontable/src/helpers/dom/__tests__/element.spec.js
```diff
@@ -22,4 +22,27 @@ describe('DOM helpers', () => {
...
+          <foreignObject width="50" height="50">
+            <div xmlns="http://www.w3.org/1999/xhtml">test</div>
+          </foreignObject>
+        </svg>
+      `;
+
+      document.body.appendChild(wrapper);
+
+      const element = wrapper.querySelector('div');
+      const elementOffset = offset(element);
```
budnix, line 40: I forgot to mention that in these tests, the DOM helpers are exposed in `Handsontable.dom` namespace. Should be:
```suggestion
      const elementOffset = Handsontable.dom.offset(element);
```

budnix (changes requested):

budnix:
> Hi @budnix, I moved the test to the `element.spec.js` file, but I wasn't sure how to run it. 🤦 😄

I reviewed the code besides one mistake it looks good 👌 

PS: You can run these tests by running `npm run test:e2e` (in the `handsontable` directory) or you can narrow them b `npm run test:e2e --testPathPattern=element`.

handsontable/src/helpers/dom/__tests__/element.spec.js
kumilingus: Makes sense! Thanks.

AMBudnik:
Hi @kumilingus 

I have great news! Your merged pull request was just published with Handsontable v14.0.0

Thank you again for your input.

Here's https://github.com/handsontable/handsontable/issues/9754 the full list of changes for this release.

PR #10490, opened 2023-08-31
Improve performance in cases when table is hiding

budnix:
### Context

The PR improves the component performance in cases when the table is mounted in places where the container element is hidden and shown alternately. For example in Tabs.

I found that the table recalculated its sizes after it detects its root element is hidden (`display: "none"`) - unnecessarily, I think. The recalculation is time-consuming, so In this PR, I changed the logic to do a recalculation thing only when the element is shown. Thanks to that, there is no lag between switching the tabs as the Handsontable after showing the tab again uses the cached values.

### How has this been tested?

I tested the changes locally, and I covered the change with new tests.

### Types of changes

### Related issue(s):
1. fixes https://github.com/handsontable/dev-handsontable/issues/975

### Affected project(s):
- [x] `handsontable`

### Checklist:


wszymanski (approved):
Good job 👍

PR #10502, opened 2023-09-06
Fix a bug where pressing a Ctrl/Cmd does not trigger non-contiguous selection

budnix:
### Context

The PR fixes a bug where pressing a Ctrl/Cmd does not trigger non-contiguous selection.

### How has this been tested?

I tested the changes locally and I covered the fix by update tests.

### Types of changes

### Related issue(s):
1. fixes https://github.com/handsontable/dev-handsontable/issues/1025

### Affected project(s):
- [x] `handsontable`

### Checklist:

### DIFF
--- a/handsontable/src/core.js
+++ b/handsontable/src/core.js
@@ -4713,11 +4713,8 @@ export default function Core(rootElement, userSettings, rootInstanceSymbol = fal
   };
 
   const shortcutManager = createShortcutManager({
-    handleEvent(event) {
-      const isListening = instance.isListening();
-      const isKeyboardEventWithKey = event?.key !== void 0;
-
-      return isListening && isKeyboardEventWithKey;
+    handleEvent() {
+      return instance.isListening();
     },
     beforeKeyDown: (event) => {
       return this.runHooks('beforeKeyDown', event);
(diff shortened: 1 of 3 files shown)


handsontable/src/core.js
```diff
@@ -4713,11 +4713,8 @@ export default function Core(rootElement, userSettings, rootInstanceSymbol = fal
   };
 
   const shortcutManager = createShortcutManager({
-    handleEvent(event) {
-      const isListening = instance.isListening();
-      const isKeyboardEventWithKey = event?.key !== void 0;
```
budnix, line 4718: @wszymanski I removed this line as I felt it's here only for tests, not for real user use cases. The `event` object is always passed as it is directly passed from the native event, and looking through the [MDN](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) and [typing](https://github.com/microsoft/TypeScript/blob/main/src/lib/dom.generated.d.ts#L14376) the `key` property seems to have always some value.

wszymanski: I can't find a reason why it's there.

wszymanski (approved):

PR #10524, opened 2023-10-03
- Add the aria tags to the autocomplete editor

jansiegel:
### Context
This PR should implement the following requirements from the original issue:

> An editor (`textarea`) should have the following role and attributes:
  - [x] type="text"
  - [x] role="combobox"
  - [x] aria-haspopup="listbox"
  - [x] aria-expanded="true/false" 
  - [x] aria-autocomplete="list"
  - [x] aria-controls="ID of the opened listbox"
  - [x] aria-activedescendant="ID of the selected item"

> Focus management ([docs](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/)):

- [x] When a combobox receives focus, DOM focus is placed on the combobox element.
- [x] When a descendant of a listbox, grid, or tree popup is focused, DOM focus remains on the combobox and the combobox has [aria-activedescendant](https://w3c.github.io/aria/#aria-activedescendant) set to a value that refers to the focused element within the popup.

> The list of items (listbox) should have the following role and attributes:

  - [x] The list of items has `role: listbox` 
  - [x] The listbox has a unique ID for the `aria-controls` purposes
  - [x] The listbox items own `role: option`
  - [x] The listbox items have unique ID for the `aria-activedescendants` purposes
  - [x] The listbox could have `aria-live: polite` 
  - [x] The listbox items own `aria-setsize`, `aria-posinset` and `aria-selected` properly.
    > [!NOTE]
    > `aria-setsize` and `aria-posinset` are added only if the `source` is provided as an array - if it's a function, there's no way of calculating the size of the "full" set.

It does NOT implement:
- > [ ] aria-disabled="true/false" + aria-readonly="true/false" depending on whether the text editor is enabled or not.
    
    I couldn't come up with an instance where this would be needed.
- >   - [x] The listbox items own (...) `aria-checked`
    
    Same as the previous point.

[skip changelog]

### How has this been tested?
Added test cases to verify the existence of the aria tags.

### Types of changes

### Related issue(s):
1. handsontable/dev-handsontable#1439

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:


handsontable/src/editors/autocompleteEditor/autocompleteEditor.js
```diff
@@ -94,6 +145,10 @@ export class AutocompleteEditor extends HandsontableEditor {
     super.open();
 
     const trimDropdown = this.cellProperties.trimDropdown === void 0 ? true : this.cellProperties.trimDropdown;
+    const rootInstanceAriaTagsEnabled = this.hot.getSettings().ariaTags;
+    const sourceArray = typeof this.cellProperties.source === 'function' ? null : this.cellProperties.source;
```
budnix, line 149:
```suggestion
    const sourceArray = Array.isArray(this.cellProperties.source) ? this.cellProperties.source : null;
```

handsontable/src/editors/autocompleteEditor/autocompleteEditor.js
```diff
@@ -127,11 +183,47 @@ export class AutocompleteEditor extends HandsontableEditor {
           }
         }
 
+        if (rootInstanceAriaTagsEnabled) {
+          setAttribute(TD, [
+            A11Y_OPTION(),
+            // Add `setsize` and `posinset` only if the source is an array.
+            ...(sourceArray ? [A11Y_SETSIZE(sourceSize)] : []),
+            ...(sourceArray ? [A11Y_POSINSET(sourceArray.indexOf(value) + 1)] : []),
+            ['id', `${this.htEditor.rootElement.id}_${value}`],
```
budnix, line 192: I wonder if adding a cell value to the id attribute is necessary. The row and col would not be enough?

image

handsontable/src/editors/autocompleteEditor/autocompleteEditor.js
```diff
@@ -127,11 +183,47 @@ export class AutocompleteEditor extends HandsontableEditor {
...
-      autoColumnSize: true,
+      afterSelection: (startRow, startCol) => {
+        if (rootInstanceAriaTagsEnabled) {
+          const TD = this.htEditor.getCell(startRow, startCol, true);
+
+          if (TD) {
+            setAttribute(TD, [
+              A11Y_SELECTED(),
+            ]);
+          }
```
budnix, line 206:
```suggestion
          setAttribute(TD, [
            A11Y_SELECTED(),
          ]);
```
I guess the TD element is always there.

budnix (changes requested):

handsontable/src/editors/autocompleteEditor/autocompleteEditor.js
jansiegel: Row and col identify the editor in the main table, but I think I can change the value in the id to row and col in the dropdown HOT instance 👍

budnix (approved):

magierg:
aria tags available in 14.0.0: https://jsfiddle.net/api/post/library/pure/

PR #10550, opened 2023-10-25
Add new `beforeColumnWrap` and `beforeRowWrap` hooks

budnix:
### Context

The PR fixes an issue related to unfocusing the table using <kbd>Tab</kbd> key when the _NestedHeaders_ plugin was enabled. To fix the error, it was necessary to introduce new hooks that help determine when the moment of row wrapping occurs.

#### `beforeRowWrap` hook
```typescript
const hot = new Handsontable({
  beforeRowWrap: (isWrapEnabled: boolean, newCoords: CellCoords, isFlipped: boolean) => {
  
  }
});
```
#### `beforeColumnWrap` hook
```typescript
const hot = new Handsontable({
  beforeColumnWrap: (isWrapEnabled: boolean, newCoords: CellCoords, isFlipped: boolean) => {
  
  }
});
```

#### Hooks' arguments
 * `isWrapEnabled`: The argument tells whether the row and/or column wrapping is going to happen. There may be situations where the option does not work even though it is enabled. This is due to the priority of other options that may block the feature. For example, when the `minSpareRows` is defined, the `autoWrapCol` option is not checked. Thus, column wrapping is off;
 * The `newCoords` may help determine what cell will be focused on after the wrapping operation;
 * The `isFlipped` is set to `true` when the user reaches the last row/column, and the focus is moved to the first row/column or vice versa, `false` otherwise.

#### Miscellaneous changes
The two new methods were added to the `CellCoords` class. This includes:
 * `isRtl()` that allows checking if the coordinates running in RTL layout direction mode;
 * `assign()` that allows assigning coordinates from another `CellCoords` instance or from a literal object that is compatible with `CellCoords` schema (has `row` and/or `col` properties).

### Types of changes

### Related issue(s):
1. fixes https://github.com/handsontable/dev-handsontable/issues/1499

### Affected project(s):
- [x] `handsontable`

### Checklist:


AMBudnik:
Hooks were introduced in v14.0.0. Updated demo https://jsfiddle.net/qsedcahb/8/

Issue #1634, opened 2014-07-23
[Custom borders] Does not work with fixedRowsTop


warpech:
Trying to use `custom_borders.html` demo with settings

```
fixedColumnsLeft: 2,
fixedRowsTop: 2,
```

The custom borders are not displayed on fixed rows/columns overlays. See the screenshot:

![screenshot 2014-07-23 14 34 33](https://cloud.githubusercontent.com/assets/566463/3672891/1a5a8fca-1266-11e4-8a6d-921ab28d1cd6.png)

Possibly the best way to fix this is Walkontable `selections` setting, as I mentioned in issue #1461:

> If I may suggest, in customBorders.js could you instead of creating instance of WalkontableBorder add custom borders to selections setting of Walkontable? (currently there are only 3 selections (current, area, fill), configured in tableView.js line 153.

PR #10658, opened 2023-12-13
Docs: A11y page changes

kirszenbaum:
This PR:

- Fixes https://github.com/handsontable/dev-handsontable/issues/1634
- Makes minor language and structure changes (I'm happy to provide my reasoning for each change, if needed), for example:
  - I aligned the formatting of keys with what’s on the [Keyboard shortcuts](https://handsontable.com/docs/javascript-data-grid/keyboard-shortcuts/) page – we can change them the other way round if you prefer, but I'd keep it consistent across the docs
  - I fixed the macOS input source instruction (it’s not maintainable and we should link to Apple’s docs, but I couldn’t find them)

[skip changelog]


kirszenbaum:
@evanSe three questions from me:
1. Is there a reason why you introduced the additional term `Shortcut keys` as opposed to `Keyboard shortcuts` / `Shortcuts`?
2. `When using frozen rows and columns, NVDA and JAWS may incorrectly announce the current column header, other cells remain unaffected.` Does it mean using **both** frozen rows and frozen columns at once, or either one of them?
3. `Some screen readers may incorrectly read the number of rows and columns when frozen rows and columns are enabled.` Does it mean using **both** frozen rows and frozen columns at once, or either one of them?

krzysztofspilka:
@kirszenbaum I think all the changes except those originally reported in [#1634](https://github.com/handsontable/dev-handsontable/issues/1634) we could do in a separate PR (for the sake of consistency:).

>Is there a reason why you introduced the additional term Shortcut keys as opposed to Keyboard shortcuts / Shortcuts?

I guess I inspired that approach, and there was no particular reason behind it. If you feel that's something we should unify, feel free to do it.

> When using frozen rows and columns, NVDA and JAWS may incorrectly announce the current column header, other cells remain unaffected. Does it mean using both frozen rows and frozen columns at once, or either one of them?

That "(...) other cells remain unaffected" is to be removed from that sentence. Generally it means that if you select a cell (not header) that is a part of a) frozen row, b) frozen column, or c) both (think of cell A1 selected when both frozen rows and columns are enabled), it might inaccurately announce the column header name for that cell.

> Some screen readers may incorrectly read the number of rows and columns when frozen rows and columns are enabled. Does it mean using both frozen rows and frozen columns at once, or either one of them?

Either of them or all at once.

kirszenbaum:
@evanSe ready for your review

evanSe (approved):

PR #10708, opened 2024-01-08
Docs: Remove notes about modules

kirszenbaum:
This PR removes the notes about using modules from the "Configuration options" guide, to declutter the page.

### Updated sections
- http://localhost:8080/docs/javascript-data-grid/configuration-options/
- http://localhost:8080/docs/react-data-grid/configuration-options/

[skip changelog]

### DIFF
--- a/docs/content/guides/getting-started/configuration-options.md
+++ b/docs/content/guides/getting-started/configuration-options.md
@@ -444,12 +444,6 @@ To apply configuration options to an individual row (or a range of rows), use th
 
 Any options modified through [`cells`](@/api/options.md#cells) overwrite all other options.
 
-::: tip
-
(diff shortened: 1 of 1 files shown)


AMBudnik (commented):
Hey. I'm not sure if I would remove both of them. But they did not match the paragraph, so maybe we could. ps. I just noticed that we have the second tip 3 times (same stuff). Can we keep the first one and remove those two later in the text?

Zrzut ekranu 2024-01-9 o 08 41 19

kirszenbaum:
@AMBudnik – done: 14cf8e035d9bbd8bfbe6b3e46a20b8f18b064c1f. Keep in mind that this changes the approach. The page wasn't writtten to be read as a whole, top to bottom, but with individual sections in mind – hence the repeated notes (for example, if you want to change an option for a column, you just go to [Set column options](https://handsontable.com/docs/javascript-data-grid/configuration-options/#set-column-options) and don't care about the rest – now, to get the same level of info, you need to read the whole page). Just saying that more changes may be required for the page to make sense if we actually expect people to read it like a blog post

AMBudnik (approved):
Thank you @kirszenbaum

PR #10719, opened 2024-01-11
Fixed problems related to not syncing sheet rename from HF to HOT

wszymanski:
### Context

Added missing action syncing changes from HF to HOT.

### How has this been tested?

Tested just case [from the issue](https://github.com/handsontable/dev-handsontable/issues/1362).

### Types of changes

### Related issue(s):
1. https://github.com/handsontable/dev-handsontable/issues/1362

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

### DIFF
--- a/handsontable/src/plugins/formulas/__tests__/formulas.spec.js
+++ b/handsontable/src/plugins/formulas/__tests__/formulas.spec.js
@@ -2953,7 +2953,7 @@ describe('Formulas general', () => {
 
   describe('renaming sheet for HF instance', () => {
     it('should update HOT\'s plugin internal property', () => {
-      let hookTriggered = false;
+      let sheetNameInsideHook = '';
       const hfInstance = HyperFormula.buildEmpty({});
       const hot = handsontable({
         data: [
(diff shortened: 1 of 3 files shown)


handsontable/src/plugins/formulas/__tests__/formulas.spec.js
```diff
@@ -2951,6 +2951,65 @@ describe('Formulas general', () => {
...
+          sheetName: 'Sheet1'
+        },
+        columns: [{
+          type: 'date',
+          dateFormat: 'DD/MM/YYYY'
+        }],
+      });
+
+      hot.addHook('afterSheetRenamed', () => {
+        expect(hot.getPlugin('formulas').sheetName).toBe('Lorem Ipsum');
```
budnix, line 2974: I think that it's a bad idea to make expectations in the hook callbacks. When the hook somehow stops working (no callback call), there won't be a test error or warning or so.

I'd suggest something like that
```js
let sheetNameViaHook;

hot.addHook('afterSheetRenamed', () => {
  sheetNameViaHook = hot.getPlugin('formulas').sheetName;
});

expect(sheetNameViaHook).toBe('Lorem Ipsum');
```

budnix (changes requested):

handsontable/src/plugins/formulas/__tests__/formulas.spec.js
wszymanski: That's why I've added an extra variable: https://github.com/handsontable/handsontable/pull/10719/files#diff-d5c22bed64b93701cdf3ae18b046941d8898c7d8826d7d032e5760437d78b687R2976 :) Anyway, I can change it to make oneliner.

budnix (approved):

PR #10743, opened 2024-01-19
Add Codesandbox button for javascript version of docs

adrianspdev:
### Context
The PR implements a new button for Codesandbox demo in javascript version of documentation. Main goal is to add Codesandbox demos for all versions of documentation including React, Vue and Angular, but implementing examples for these frameworks requires a completely new solution. This solution should include file structure configuration for each framework and code examples adjusted to these wrappers. 

The PR implementation of Codesandbox is done via https://codesandbox.io/docs/learn/sandboxes/cli-api#xhr-request.

### How has this been tested?
Manual tests.

### Types of changes

### Related issue(s):
1. https://github.com/handsontable/dev-handsontable/issues/1715

### Affected project(s):
- [x] `handsontable`

### Checklist:
- [x] "Edit" button is present for frameworks demos.
- [x] "Edit" button is replaced with two distinct playground provider icons for javascript demos.
- [x] Codesandbox icon is present and redirects to a new tab upon click for javascript demos.
- [x]  Manual testing has been performed and passed.


jansiegel (commented):
Seems to work alright, but I'm not sure if the HTML structure being passed to codesandbox is what it's supposed to be.

Basically, the `index.html` file receives just the contents of the `HTML` and `CSS` tabs, which is not semantically correct.
For example:
```html
<style>
h3.demo-preview {
  margin-bottom: 0.3rem !important;
  padding-top: 0 !important;
  margin-top: 0.5rem !important;
}

</style>

<h3 class="demo-preview">Sheet 1</h3>
<div id="example-basic-multi-sheet-1"></div>
<h3 class="demo-preview">Sheet 2</h3>
<div id="example-basic-multi-sheet-2"></div>
```

I _think_ we should generate a full HTML structure, complete with `<head>`, `<body>` etc.

WDYT? @adrianspdev @evanSe

docs/.vuepress/theme/components/Page.vue
```diff
@@ -51,6 +51,58 @@ export default {
...
+          'Content-Type': 'application/json',
+          Accept: 'application/json',
+        },
+        body: JSON.stringify({
+          files: {
+            'package.json': {
+              content: {
+                name: 'handsontable-demo',
+                dependencies: {
+                  handsontable: 'latest',
```
jansiegel, line 84: @adrianspdev I think this part might be problematic when the documentation gets archived under a version number (when a newer version gets released).
It would be a better idea to use an actual Handsontable version instead of `latest` here.

docs/.vuepress/theme/components/Page.vue
```diff
@@ -51,6 +51,58 @@ export default {
...
     isScriptLoaderActivated(exampleId) {
       return this.activatedExamples.includes(exampleId);
+    },
+    submitCodesandbox(submitEvent) {
+      // form data values
+      const html = submitEvent.target.elements.html.value;
+      const js = submitEvent.target.elements.js.value;
+      const css = submitEvent.target.elements.css.value;
+
+      const hyperformula = js.includes('hyperformula') ? { hyperformula: 'latest' } : {};
```
jansiegel, line 61: @adrianspdev I think this part might be problematic when the documentation gets archived under a version number (when a newer version gets released).
It would be a better idea to use an actual Hyperformula version instead of `latest` here.

jansiegel (changes requested):

docs/.vuepress/theme/components/Page.vue
adrianspdev: done

docs/.vuepress/theme/components/Page.vue
adrianspdev: for now we don't have easy way to do this, @evanSe we should create separate ticket for this case

jansiegel (approved):

PR #10746, opened 2024-01-23
Undo will work for moving columns

wszymanski:
### Context

As the title says. Works similarly to the undo operation of moving rows

### How has this been tested?

Moving single and multiple rows.

### Types of changes

### Related issue(s):
1. https://github.com/handsontable/dev-handsontable/issues/933

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

### DIFF
--- a/.changelogs/10746.json
+++ b/.changelogs/10746.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Undo will work for moving columns",
+  "type": "added",
+  "issueOrPR": 10746,
+  "breaking": false,
+  "framework": "none"
+}
--- a/docs/content/guides/getting-started/installation.md
+++ b/docs/content/guides/getting-started/installation.md
@@ -147,6 +147,8 @@ const hot = new Handsontable(container, {
   rowHeaders: true,
   colHeaders: true,
   height: 'auto',
+  autoWrapRow: true,
+  autoWrapCol: true,
   licenseKey: 'non-commercial-and-evaluation' // for non-commercial use only
 });
 ```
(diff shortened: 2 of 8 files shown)


jansiegel (changes requested):
I think I found a use case when it doesn't work as intended:

1. Go to https://jsfiddle.net/js_ziggle/528wL1gh/
2. Move the `A` column (with the `A1`, `A2`, (...) data) to be the last column
3. Move the `B` column (with the `C1`, `C2`, (...) data) to be the last column
4. <kbd>Cmd + Z</kbd>
5. <kbd>Cmd + Z</kbd>

**Result:**
The columns should be in the initial order, but they're not.

The result with the actions done using the API: https://jsfiddle.net/js_ziggle/p4e9Lck1/

wszymanski:
> I think I found a use case when it doesn't work as intended:
> 
> 1. Go to [jsfiddle.net/js_ziggle/528wL1gh](https://jsfiddle.net/js_ziggle/528wL1gh/)
> 2. Move the `A` column (with the `A1`, `A2`, (...) data) to be the last column
> 3. Move the `B` column (with the `C1`, `C2`, (...) data) to be the last column
> 4. Cmd + Z
> 5. Cmd + Z
> 
> **Result:** The columns should be in the initial order, but they're not.
> 
> The result with the actions done using the API: [jsfiddle.net/js_ziggle/p4e9Lck1](https://jsfiddle.net/js_ziggle/p4e9Lck1/)

The problem existed also for the `ManualRowMove` plugin. Provided a solution for UndoRedo using the exact mechanism. However, I've fixed that within: https://github.com/handsontable/handsontable/pull/10746/commits/8b9fe2fc719ee4b558e5e3ec035418b8f0ba3d9c.

jansiegel (approved):

PR #10748, opened 2024-01-24
Changed unnatural checkbox behaviour after double click

wszymanski:
### Context
I can see two problems.

1. Unnatural checkbox behaviour after double click.

~After a double click, a value has been inverted. I can't see a reason for such behaviour. It was introduced within https://github.com/handsontable/handsontable/commit/1c7a93a8e5bbd17167b134c039a92883c4499259 (2013 year), and it's very hard to find a reason for that feature. I can't see similar behaviour in Google Sheets or the Excel desktop. I propose to get rid of this solution. This PR did it.~

So far, double-clicking on a cell (`TD` element) has been inverting the checkbox value. However, it had side effects when a click was performed directly on the `input[type=checkbox]` element. This PR fixes that.

2. Big lag.

The lag problem shown within the issue is related to the `setDataAtCell` method. It's very clearly visible after clicking on the checkbox, as it is the easiest way of changing value. A lag can also be seen in changing value using fast editing or from API. The longest time is spent executing the `instance.view.render` method, which also takes much time when displaying the Handsontable instance with a big dataset for the first time. We don't have a method which re-renders only part of the cells visible in the viewport. Probably, it should be done within another, bigger task.

### How has this been tested?

I tested a checkbox using clicks, double clicks, and space presses.

### Types of changes

### Related issue(s):
1. https://github.com/handsontable/dev-handsontable/issues/1540

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

### DIFF
--- a/.changelogs/10748.json
+++ b/.changelogs/10748.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Changed unnatural checkbox behaviour after double click",
+  "type": "changed",
+  "issueOrPR": 10748,
+  "breaking": false,
+  "framework": "none"
+}
--- a/.changelogs/10748.json
+++ b/.changelogs/10748.json
@@ -3,6 +3,6 @@
   "title": "Changed unnatural checkbox behaviour after double click",
   "type": "changed",
   "issueOrPR": 10748,
-  "breaking": true,
+  "breaking": false,
   "framework": "none"
 }
--- a/.changelogs/10748.json
+++ b/.changelogs/10748.json
@@ -3,6 +3,6 @@
   "title": "Changed unnatural checkbox behaviour after double click",
   "type": "changed",
   "issueOrPR": 10748,
-  "breaking": false,
+  "breaking": true,
   "framework": "none"
 }
--- a/handsontable/src/core.js
+++ b/handsontable/src/core.js
@@ -4837,9 +4837,7 @@ export default function Core(rootElement, userSettings, rootInstanceSymbol = fal
    */
   this._refreshBorders = function(revertOriginal = false, prepareEditorIfNeeded = true) {
     editorManager.destroyEditor(revertOriginal);
-    // console.time('render');
     instance.view.render();
-    // console.timeEnd('render');
 
     if (prepareEditorIfNeeded && selection.isSelected()) {
       editorManager.prepareEditor();
--- a/handsontable/src/core.js
+++ b/handsontable/src/core.js
@@ -4837,7 +4837,9 @@ export default function Core(rootElement, userSettings, rootInstanceSymbol = fal
    */
   this._refreshBorders = function(revertOriginal = false, prepareEditorIfNeeded = true) {
     editorManager.destroyEditor(revertOriginal);
+    // console.time('render');
     instance.view.render();
+    // console.timeEnd('render');
 
     if (prepareEditorIfNeeded && selection.isSelected()) {
       editorManager.prepareEditor();
--- a/handsontable/src/editors/checkboxEditor/checkboxEditor.js
+++ b/handsontable/src/editors/checkboxEditor/checkboxEditor.js
@@ -1,5 +1,4 @@
 import { BaseEditor } from '../baseEditor';
-import { hasClass } from '../../helpers/dom/element';
 
 export const EDITOR_TYPE = 'checkbox';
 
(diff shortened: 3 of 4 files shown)


wszymanski:
@evanSe @adrianszymanski89 If the purpose of the task was to repair the lag, it would be a more complicated process, which, in my opinion, may take a lot of time and should be solved as a separate task. One of the proposed changes can, in my opinion, be treated as a breaking change.

jansiegel (commented):
We could potentially remove this feature (if we don't want to keep it, that's a question for @evanSe), but I have an alternative solution:

I think the main problem with this task is the fact that when you double-click the actual checkbox, something like this happens:
1. Cell contains an unchecked checkbox
2. User double-clicks the checkbox element
3. Checkbox gets checked
4. Checkbox gets unchecked
5. (pause) Checkbox gets checked again

I think that `5. (pause) Checkbox gets checked again` is the main problem here.

My suggestion would be to leave the double-click functionality when interacting with a _cell_, but remove it for the _checkbox_. This way double-clicking in the general cell area would flip the checkbox state, but double-clicking on the checkbox would be synonymous to two separate clicks (so the chebkox would get checked -> unchecked).

I think this approach would require a really minor change here: https://github.com/handsontable/handsontable/blob/549d023a020994cee2bb08486f74e55ca77a215d/handsontable/src/editors/checkboxEditor/checkboxEditor.js#L20

Basically, we'd need to change this line to
```js
if (event && event.type === 'mouseup' && event.target.nodeName === 'TD') {
```

WDYT, @wszymanski?

wszymanski:
> We could potentially remove this feature (if we don't want to keep it, that's a question for @evanSe), but I have an alternative solution:
> 
> I think the main problem with this task is the fact that when you double-click the actual checkbox, something like this happens:
> 
> 1. Cell contains an unchecked checkbox
> 2. User double-clicks the checkbox element
> 3. Checkbox gets checked
> 4. Checkbox gets unchecked
> 5. (pause) Checkbox gets checked again
> 
> I think that `5. (pause) Checkbox gets checked again` is the main problem here.
> 
> My suggestion would be to leave the double-click functionality when interacting with a _cell_, but remove it for the _checkbox_. This way double-clicking in the general cell area would flip the checkbox state, but double-clicking on the checkbox would be synonymous to two separate clicks (so the chebkox would get checked -> unchecked).
> 
> I think this approach would require a really minor change here:
> 
> https://github.com/handsontable/handsontable/blob/549d023a020994cee2bb08486f74e55ca77a215d/handsontable/src/editors/checkboxEditor/checkboxEditor.js#L20
> 
> Basically, we'd need to change this line to
> 
> ```js
> if (event && event.type === 'mouseup' && event.target.nodeName === 'TD') {
> ```
> 
> WDYT, @wszymanski?



> We could potentially remove this feature (if we don't want to keep it, that's a question for @evanSe), but I have an alternative solution:
> 
> I think the main problem with this task is the fact that when you double-click the actual checkbox, something like this happens:
> 
> 1. Cell contains an unchecked checkbox
> 2. User double-clicks the checkbox element
> 3. Checkbox gets checked
> 4. Checkbox gets unchecked
> 5. (pause) Checkbox gets checked again
> 
> I think that `5. (pause) Checkbox gets checked again` is the main problem here.
> 
> My suggestion would be to leave the double-click functionality when interacting with a _cell_, but remove it for the _checkbox_. This way double-clicking in the general cell area would flip the checkbox state, but double-clicking on the checkbox would be synonymous to two separate clicks (so the chebkox would get checked -> unchecked).
> 
> I think this approach would require a really minor change here:
> 
> https://github.com/handsontable/handsontable/blob/549d023a020994cee2bb08486f74e55ca77a215d/handsontable/src/editors/checkboxEditor/checkboxEditor.js#L20
> 
> Basically, we'd need to change this line to
> 
> ```js
> if (event && event.type === 'mouseup' && event.target.nodeName === 'TD') {
> ```
> 
> WDYT, @wszymanski?

You are correct. I didn't notice that this feature mainly concerns clicking on TD. As a consequence, your solution is definitely better, and consequently, there is no point in talking about breaking change.

jansiegel:
> With this solution will the user see it checked then unchecked? if after double-clicking a unchecked box its state is still unchecked I am for this solution
@evanSe it would work like this:

1. Double-clicking the `TD` element:
- Checkbox get checked

2. Double-clicking the `input[type=checkbox]` element:
- Checkbox get checked after the first click
- Checkbox gets unchecked after the second click

jansiegel (approved):

PR #10756, opened 2024-01-29
Display the numeric renderer's output in the Filters' multi-selection module.

jansiegel:
### Context
This PR:
- Adds a new `displayValue` cell meta property
- Makes the filters' multi-selection module utilize `displayValue`s to render the entry list
- Sets the new `displayValue` prop in the numeric renderer

### How has this been tested?
Added new test cases.

### Types of changes

### Related issue(s):
1. handsontable/dev-handsontable#1267

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:


handsontable/src/plugins/filters/__tests__/filtersUI.spec.js
```diff
@@ -1058,6 +1058,37 @@ describe('Filters UI', () => {
...
+
+      const unifiedColDataSample = [
+        'Alice Blake', 'Alyssa Francis', 'Becky Ross', 'Bridges Sawyer', 'Burt Cash', 'Carissa Villarreal'
+      ];
+
+      for (let i = 0; i < unifiedColDataSample.length; i++) {
+        expect(
+          byValueMultipleSelect().element.querySelectorAll('.htCore td')[i].textContent
+        ).toBe(`Custom ${unifiedColDataSample[i]}`);
+      }
```
budnix, line 1089:
```suggestion
      for (let i = 0; i < unifiedColDataSample.length; i++) {
        expect(
          byValueMultipleSelect().element.querySelectorAll('.htCore td')[i].textContent
        ).toBe(`Custom ${unifiedColDataSample[i]}`);
      }
      expect(unifiedColDataSample.length).toBe(6);
```
I'd secure that `for` with one more `expect`.

handsontable/src/helpers/dom/element.js
```diff
@@ -1148,3 +1148,21 @@ export function runWithSelectedContendEditableElement(element, callback, invisib
...
+
+/**
+ * Return a `td` element rendered outside of the table with the provided renderer function.
+ *
+ * @param {Document} rootDocument Root document element.
+ * @param {Function} renderFunction The renderer function.
+ * @param {Array} renderArgs The renderer arguments.
+ * @returns {HTMLElement} The rendered `td` element.
+ */
+export function renderMockCell(rootDocument, renderFunction, renderArgs) {
```
budnix, line 1160: Can I ask you to cover that function with a test?

handsontable/src/plugins/filters/ui/multipleSelect.js
```diff
@@ -322,6 +322,56 @@ export class MultipleSelectUI extends BaseUI {
...
+
+    for (let row = 0; row < sourceRowCount; row += 1) {
+      const cellMeta = this.hot.getCellMeta(row, column);
+      const {
+        prop,
+        displayValue
+      } = cellMeta;
+      const cellSourceData = this.hot.getSourceDataAtCell(row, column);
+
+      if (displayValue) {
```
budnix, line 351: When the `displayValue` value is `0` (`false` and maybe others), then it's not rendered. I'd try to render all values except `undefined` and only then return source data.

handsontable/src/renderers/numericRenderer/__tests__/numericRenderer.spec.js
```diff
@@ -174,4 +194,33 @@ describe('NumericRenderer', () => {
       }, 500);
     });
   });
+
+  describe('NumericRenderer with Filters', () => {
```
budnix, line 198: I'd move this Filters-related test to the Filters plugin tests directory.

budnix (changes requested):
I'm not convinced about that `displayValue` option. I know that the purpose of this task is to add support for modifying filter values only for numeric values, but with this shape, it isn't very clear.
The question that comes to my mind or limitation that I found:
 * The value can not be changed in the `cells` function. The `displayValue` property is always overwritten in the numeric renderer; 
 * When the `displayValue` property does not exist, the Filters plugin generates one by calling the numeric renderer by itself. So what is this option for? Kind of read-only and for numeric columns only?
 * Maybe a better idea would be creating renderers for Filters. The same goes for cells. For example, `filtersRenderer` option, or so?

handsontable/src/renderers/numericRenderer/numericRenderer.js
```diff
@@ -53,6 +53,8 @@ export function numericRenderer(hotInstance, TD, row, col, prop, value, cellProp
     TD.dir = 'ltr';
   }
 
+  cellProperties.displayValue = newValue;
```
budnix, line 56: I think that it's not the best idea to change cell meta within the renderer function. The changes disturb the function purity 🤔

jansiegel:
I went through several approaches to this task, but I'll take any suggestions. As for the questions:

> * The value can not be changed in the `cells` function. The `displayValue` property is always overwritten in the numeric renderer;

Yes. The spec requires the multi-selection component in the Filters dropdown to display the same value that's being rendered in the cells. As we cannot know what the renderer with generate, there's no way to assign this value before it's triggered.

> * When the `displayValue` property does not exist, the Filters plugin generates one by calling the numeric renderer by itself. So what is this option for? Kind of read-only and for numeric columns only?

I had to add this logic because the Filters dropdown display all unique values in the table, and the renderers are run just for the visible part of the dataset. Without this logic, some of the values were being displayed in their "source" form, and some in the "rendered" form.

> * Maybe a better idea would be creating renderers for Filters. The same goes for cells. For example, `filtersRenderer` option, or so?

We can do that, but according the the spec, in this case the `filtersRenderer` should be exact same as the regular `renderer` (and for the other cell types the `filtersRenderer` should return the cell's source value...?)

budnix:
Maybe I'd suggest adding a new hook for the MultipleSelect component of the Filters plugin. Let's name it `modifyFiltersMultipleSelectValue` hook, for example. This hook would be responsible only for modifying the values presented by that component. There would be one hook for all cell types.

Now, depending on the feature needs you can pass as many arguments as you wish. Usage example:
```js
modifyFiltersMultipleSelectValue(sourceValue, cellType, cellMeta) => {
  let newValue = sourceValue;

  if (cellType === 'numeric') {
    if (isNumeric(newValue)) { // HERE I'd copy the logic from the renderer or move it to separate helper
      const numericFormat = cellProperties.numericFormat;
      const cellCulture = numericFormat && numericFormat.culture || '-';
      const cellFormatPattern = numericFormat && numericFormat.pattern;
      const className = cellProperties.className || '';
      const classArr = className.length ? className.split(' ') : [];

      if (typeof cellCulture !== 'undefined' && !numbro.languages()[cellCulture]) {
        const shortTag = cellCulture.replace('-', '');
        const langData = numbro.allLanguages ? numbro.allLanguages[cellCulture] : numbro[shortTag];

        if (langData) {
          numbro.registerLanguage(langData);
        }
      }

      numbro.setLanguage(cellCulture);

      newValue = numbro(newValue).format(cellFormatPattern || '0');
    } 
  }
  // PLACE FOR OTHER CELL TYPES

  return newValue;
}
```

handsontable/src/plugins/filters/__tests__/filtersUI.spec.js
budnix: I assume that there is a chance a developer (unintentionally, by accident) can edit `for` loop or edit an array in a way that there won't be any iteration. Then the test will pass, as no `expect` is called.

budnix (changes requested):
It looks good. I left one comment with suggestions to be considered. And I wonder, would it be possible to pass a complete cellMeta object as a second argument? It would be better for the user not to have to figure out a new data structure. They already know/use cellMeta objects in other hooks or methods.

budnix:
Ah, and I almost forgot. Can you add a TypeScript definition for `modifyFiltersMultiSelectValue` hook? 😄

budnix (approved):

PR #10758, opened 2024-01-30
fix: cache portal container

evanSe:
Internally, in our React wrapper, we were caching our portals but needlessly discarding our portal containers.

This pull request introduces a `portalContainerCache`. If we have a complete cache match (`cachedPortal && cachedPortalContainer`), we append the cached portal container.

If we do not have a full cache match, we recreate the portal and potentially reuse the portal container based on a cache check.

**Before**
https://github.com/handsontable/handsontable/assets/11815608/c674d458-cfc8-4a03-ba3f-d4048871d16e

**After**
https://github.com/handsontable/handsontable/assets/11815608/15134bb8-2fef-4ab7-8ab9-5927bbd20f10

### How has this been tested?
locally

### Types of changes

### Affected project(s): 
- [x] `@handsontable/react`


evanSe:
One failing test 😆, better than expected!

budnix (changes requested):

wrappers/react/src/hotTableClass.tsx
```diff
@@ -220,39 +237,54 @@ class HotTableClass extends React.Component<HotTableProps, {}> {
...
-        const {portal, portalContainer} = createPortal(rendererElement, {
-          TD,
-          row,
-          col,
-          prop,
-          value,
-          cellProperties,
-          isRenderer: true
-        }, TD.ownerDocument);
+        const cachedPortal = hotTableComponent.portalCacheArray.find(portal => portal.key === portalKey);
```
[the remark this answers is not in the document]
evanSe: I agree with your conclusions, the `find` performance impact is not a concern on a tiny virtual set of data but for `renderAll` it would become a problem.

will make the change.

wrappers/react/src/hotTableClass.tsx
```diff
@@ -220,39 +237,54 @@ class HotTableClass extends React.Component<HotTableProps, {}> {
...
+            TD,
+            row,
+            col,
+            prop,
+            value,
+            cellProperties,
+            isRenderer: true
+          }, TD.ownerDocument, portalKey, cachedPortalContainer);
+
+          hotTableComponent.getPortalContainerCache().set(portalContainerKey, portalContainer)
```
jansiegel, line 280: Does `hotTableComponent.getPortalContainerCache()` in this line differ from the container cache stored under the `portalContainerCache` variable? Maybe we could reuse it here?

wrappers/react/src/hotTableClass.tsx
```diff
@@ -220,39 +237,54 @@ class HotTableClass extends React.Component<HotTableProps, {}> {
...
 
     return function (instance, TD, row, col, prop, value, cellProperties) {
       const renderedCellCache = hotTableComponent.getRenderedCellCache();
+      const portalContainerCache = hotTableComponent.getPortalContainerCache()
+      const key = `${row}-${col}`;
+
+       // Handsontable.Core type is mising guid
+      const instanceGuid = (instance as unknown as { guid: string }).guid;
+
+      const portalContainerKey = `${instanceGuid}${key}`
```
jansiegel, line 251: That's pretty nitpicky, but can we change the key to ``${instanceGuid}-${key}`` or something like that? Without that the keys are not too readable, for example, the keys for rows 0-2 and columns 0-1 look something like this:
```
ht_72b3657ec151e0730-0
ht_72b3657ec151e0730-1
ht_72b3657ec151e0731-0
ht_72b3657ec151e0731-1
ht_72b3657ec151e0732-0
ht_72b3657ec151e0732-1
```

I know that they're not meant to be read by the human eye, but it would make it easier for debugging.

evanSe: Agreed

wrappers/react/src/hotTableClass.tsx
evanSe: Does not differ, thanks for suggestion

budnix (approved):

createdbyconnor:
Hey guys, thank you so much for getting this fix into develop quickly. Any ideas when it will be merged into master and released?

adrianszymanski89:
Hi @createdbyconnor 

The release should be out by the end of February. The information about it will be announced in our main repository.

createdbyconnor:
Hey @adrianszymanski89 

Are you able to confirm if you guys are on track for the release? Sorry, I'm just desperately waiting for this fix and can't seem to find any info on a new release on the repo.

Thank you! 🙏

adrianszymanski89:
Hi @adrianszymanski89 

Yes, we have a slight delay, but it should be out next week finally.

evanSe:
Hey @createdbyconnor, Version 14.2 has been released today, which includes this fix. If there are any further issues in this area, please let me know and I will personally take a look.

createdbyconnor:
Thanks @adrianszymanski89 & @evanSe for the release update. I am currently testing out 14.2.0 to see if this caching solution solved my performance issues. However, It appears the constant rerendering still exists, do we have to enable this feature in any way?

My component is as follows:

`
const CellEditor = (props) => {
  console.log('rendering');
  return <>test</>;
};
`

Used like so:

`        <HotColumn>
          <CellEditor hot-renderer />
        </HotColumn>
`

Console log shows many rerenders:
image

Any help is appreciated, thank you!

evanSe:
Hey @createdbyconnor that behavior is correct, we use virtualization to allow us to render large quantities of data without running into the limitations of the browser. the rendering is being "logged" as the react component is being inserted into a different part of the dom and that is the behavior observed in React. ([you can observe same behaviour in popular react virtualisation library](https://stackblitz.com/github/tanstack/virtual/tree/main/examples/react/dynamic?embed=1&theme=light))


This pull request specifically fixed an issue where components were being mounted and unmounted when unnecessary.

If you want more in-depth logging I would go with the following
```tsx
import { useEffect } from "react";

type ExampleComponentProps = {
  value?: string;
};

const ExampleComponent = ({ value }: ExampleComponentProps) => {
  useEffect(() => {
    console.log("Component did mount");

    // Cleanup function 
    return () => {
      console.log("Component will unmount");
    };
  }, []);

  // Effect for component re-render (whenever any dependency changes)
  useEffect(() => {
    console.log("Component did update");
  });

  // Render method
  console.log("Component is rendering");
  return <div>{value}</div>;
};
```
You will observe many more mounts/unmounts on a version before 14.2.0

now to reduce the number of re-rendering you are seeing you should be able to take advantage of `memo` (memo should be used only in specific cases though)

```
import { useEffect, memo} from "react";

type ExampleComponentProps = {
  value?: string;
};

const ExampleComponent = memo(({ value }: ExampleComponentProps) => {
  useEffect(() => {
    console.log("Component did mount");

    // Cleanup function 
    return () => {
      console.log("Component will unmount");
    };
  }, []);

  // Effect for component re-render (whenever any dependency changes)
  useEffect(() => {
    console.log("Component did update");
  });

  // Render method
  console.log("Component is rendering");
  return <div>{value}</div>;
});

export default ExampleComponent;
```

For that simplistic example, you should now only be seeing re-rendering for components being mounted for the first time.

createdbyconnor:
@evanSe Thank you very much, that's really useful!

PR #10817, opened 2024-02-23
Make the Autofill plugin treat cells filled with `0` as non-empty.

jansiegel:
### Context
This PR:
- Makes the Autofill plugin treat cells filled with `0` as non-empty.
- Adds a test case for the problem described in handsontable/dev-handsontable#1757.

### How has this been tested?
Added a test case.

### Types of changes

### Related issue(s):
1. handsontable/dev-handsontable#1757

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### DIFF
--- a/handsontable/src/plugins/autofill/autofill.js
+++ b/handsontable/src/plugins/autofill/autofill.js
@@ -426,13 +426,14 @@ export class Autofill extends BasePlugin {
   getIndexOfLastAdjacentFilledInRow(cornersOfSelectedCells) {
     const data = this.hot.getData();
     const nrOfTableRows = this.hot.countRows();
+    const isEmptyCell = cellContent => cellContent === null || cellContent === void 0 || cellContent === '';
     let lastFilledInRowIndex;
 
     for (let rowIndex = cornersOfSelectedCells[2] + 1; rowIndex < nrOfTableRows; rowIndex++) {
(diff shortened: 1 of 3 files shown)


handsontable/src/plugins/autofill/autofill.js
```diff
@@ -426,21 +426,22 @@ export class Autofill extends BasePlugin {
   getIndexOfLastAdjacentFilledInRow(cornersOfSelectedCells) {
     const data = this.hot.getData();
     const nrOfTableRows = this.hot.countRows();
+    const isEmptyCell = cellContent => cellContent === null || cellContent === void 0 || cellContent === '';
```
budnix, line 429:
```suggestion
    const isEmptyCell = cellContent => cellContent === null || cellContent === undefined || cellContent === '';
```

budnix (changes requested):

handsontable/src/plugins/autofill/autofill.js
budnix: Or better by using [that generic helper](https://github.com/handsontable/handsontable/blob/4832dd2386432c389dc2b2d7c7d275a126066932/handsontable/src/helpers/mixed.js#L59).

budnix (approved):

PR #10873, opened 2024-03-26
Fix column sorting using ENTER key

budnix:
### Context

The PR fixes a bug where, in some plugin combinations, column sorting won't work. This PR is a hotfix for https://github.com/handsontable/handsontable/pull/10732.

_[skip changelog]_ (hotfix)

### How has this been tested?

I tested the changes locally and I covered the fix with new tests.

### Types of changes

### Related issue(s):
1. fixes https://github.com/handsontable/dev-handsontable/issues/1817

### Affected project(s):
- [x] `handsontable`

### Checklist:


jansiegel:
The fix allows sorting the rows using the <kbd>ENTER</kbd> key indeed, but I'm not sure if in some plugin configurations sorting should be allowed.

If I'm reading the test cases correctly, the problematic plugin was NestedRows. Currently, sorting the table with NestedRows by clicking on the column header leads to nonsensical results:

Before sorting (note the parent row and its children):
image

After sorting (note how the children are scattered across the table):
image

Shouldn't this be blocked?

budnix:
> Shouldn't this be blocked?

I don't know 😄 it's been like this from the beginning of NestedRows (demo https://jsfiddle.net/budnix/vyn9bc7d/).

jansiegel (approved):
The fix does what it's supposed to do, but I think we should tackle the NestedRows/sorting problem anyway in a separate issue
cc: @evanSe

PR #10908, opened 2024-04-10
Rollup interop fix

evanSe:
- support missing interop in commonjs files
[skip changelog]

### DIFF
--- a/wrappers/vue3/.config/commonjs.js
+++ b/wrappers/vue3/.config/commonjs.js
@@ -8,6 +8,11 @@ export const cjsConfig = {
     format: env,
     indent: false,
     file: `./commonjs/${filename}`,
+    interop: 'compat',
+    generatedCode: {
+			reservedNamesAsProps: false
+		},
+    systemNullSetters: false,
     exports: 'named'
   },
   plugins: baseConfig.plugins,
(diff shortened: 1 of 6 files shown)


jansiegel (approved):
This seems to fix the `commonjs` issues for the wrappers 👍  
I think we should still tackle the bigger problem, though, so I created a separate issue for solving the general problem of rollup dependencies being reused between version `4` in `@handsontable/angular` and `2` in the other wrappers: https://github.com/handsontable/dev-handsontable/issues/1852

@evanSe, could you cherry-pick this fix to `release/14.3.0` after merging it to `develop`?

jansiegel (changes requested):
@evanSe, I found that a similar thing is happening to the `umd` builds, and that's why the `verify-bundles` scripts is failing on the release branch.

As the settings you're introducing here are the `rollup@2`'s default, do you think we could implement them to all the react/vue/vue3 builds? For example in `wrappers/**/.config/base.js`?

wrappers/vue3/.config/commonjs.js
```diff
@@ -8,6 +8,11 @@ export const cjsConfig = {
     format: env,
     indent: false,
     file: `./commonjs/${filename}`,
+    interop: 'compat',
+    generatedCode: {
+			reservedNamesAsProps: false
```
jansiegel, line 13: As this is now in `base.js`, I don't think it's needed here anymore.

jansiegel (changes requested):

jansiegel (approved):

PR #10970, opened 2024-05-13
Allow ordering of the hook callbacks.

jansiegel:
### Context
This PR introduces the option of adding an "order index" to the hook callbacks.

- If the order index is > 0, the callback will be added after the others. For example, with an index of 1, the callback will be added before the ones with an index of 2, 3, etc., but after the ones with an index of 0 and lower.
 - If the order index is < 0, the callback will be added before the others. For example, with an index of -1, the callback will be added after the ones with an index of -2, -3, etc., but before the ones with an index of 0 and higher.
- If the order index is 0 or no order index is provided, the callback will be added between the "negative" and "positive" indexes.

### How has this been tested?
Added a test case for the `manualColumnResize` + `nestedHeaders` feature from handsontable/dev-handsontable#770

### Types of changes

### Related issue(s):
1. handsontable/dev-handsontable#770

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:


handsontable/src/pluginHooks.js
```diff
@@ -2911,8 +2911,14 @@ class Hooks {
   createEmptyBucket() {
     const bucket = Object.create(null);
 
+    bucket._order = Object.create(null);
```
budnix, line 2914: If you try to use the hook name under the same name, you will get an error `hot.addHook('_order', () => {})`. Did you consider using `WeakMap`'s here?

budnix (changes requested):

budnix (approved):

PR #11026, opened 2024-06-13
Modify the test case for the docs' example checker, to make the documentation build quicker.

jansiegel:
### Context
This PR changes the `example checker`'s test case to prevent it going into the loop of retries, which should speed up the documentation build process.

Edit: Looking at this PR's CI, it seems like the staging build time was cut from ~36 min to ~18 min.

[skip changelog]


budnix (approved):
Good job 👏

PR #11076, opened 2024-07-09
Implement the `headerClassName` option.

jansiegel:
### Context
This PR:
1. Implements the `headerClassName` option.
     The `headerClassName` option allows adding one or more class names to the column headers' inner `div` element.
It can be used to align the labels in the column headers to left, center or right by setting this option to
`htLeft`, `htCenter`, or `htRight` respectively.
2. Adds a `hot.getColumnMeta` method. It was needed for the new option to be possible to read the column meta from inside of the plugins.
3. Adds support for `headerClassName` in the Nested Headers plugin. It can be passed to the plugin's config option alongside other properties, like `label` and `colspan`, for example:
    ```
    nestedHeaders: [
        ['A', {label: 'B', colspan: 8, headerClassName: 'htRight bold-text'}, 'C'],
        ['D', {label: 'E', colspan: 4}, {label: 'F', colspan: 4}, 'G'],
        ['H', {label: 'I', colspan: 2}, {label: 'J', colspan: 2}, {label: 'K', colspan: 2}, {label: 'L', colspan: 2}, 'M'],
        ['N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W']
    ],
    ```
4. Adds test cases for the new method and option, TS types etc.

---
### Requirements from handsontable/dev-handsontable#1851

1. > Specifically, I want to:
    > Apply one or more custom CSS classes to a header
    > Be able to use built-in class names to align the column name to the left, center, or right
    - https://stackblitz.com/edit/jjf7xx-8szkav?file=index.js
2. > It should be applied to instance (global), and columns.
    - global: https://stackblitz.com/edit/jjf7xx-g9wr1y?file=index.js
    - columns: https://stackblitz.com/edit/jjf7xx-8szkav?file=index.js
3. > AutoColumnSize - calculating width and aligning the title for the htCenter class.
    > ManualColumnResize - calculating width and aligning the title for the htCenter class.
    - should be taken care of by the nature of CSS (it's just centered, regardless of the container size)
4. >  ColumnSorting - icon placement and distance for the htRight class or htLeft in RTL. The htRight for LTR and htLeft for RTL should respect the gap of 3 pixels.
    > MultiColumnSorting - icon placement and distance for the htRight class or htLeft in RTL. The htRight for LTR and htLeft for RTL should respect the gap of 3 pixels.
    - Column Sorting: https://stackblitz.com/edit/jjf7xx-bddav6?file=index.js
    - Multi Column Sorting: https://stackblitz.com/edit/jjf7xx-1gekts?file=index.js
5. > Filters + DropdownMenu - icon placement and distance for the htRight class or htLeft in RTL. The htRight for LTR and htLeft for RTL should respect the gap of 5 pixels.
    - https://stackblitz.com/edit/jjf7xx-uzmskm?file=index.js
6. > CollapsibleColumns - icon placement and distance for the htRight class or htLeft in RTL. The htRight for LTR and htLeft for RTL should respect the gap of 5 pixels.
    - https://stackblitz.com/edit/jjf7xx-5mlsy6?file=index.js
7. > HiddenColumns - icon placement and distance for the htRight class or htLeft in RTL. The htRight for LTR and htLeft for RTL should respect the gap of 5 pixels.
    - https://stackblitz.com/edit/jjf7xx-br3vom?file=package.json
8. > nestedHeaders -Should applying the option with nested headers configured have an effect only on the lowest level (like Sorting) or on the whole column?
    In the current solution:
    - Applying the new option globally will apply the class names to all the column headers, including the nested headers.
    - Applying the new option in the `columns` option will add the classes just to the nested header that "begin" on the modified column.
    - To apply custom classes to any nested headers, the user can pass them in the `headerClassName` prop inside the `nestedHeaders` config object.

#### Questions:
1. Should the headerClassName class automatically be added to the existing classes set in the activeHeaderClassName or currentHeaderClassName options?
    - The `activeHeaderClassName` and `currentHeaderClassName` options are added to the `TH` element, and `headerClassName` is added to the `div` element inside of the `TH`, so they don't affect each other directly.

2. Will headerClassName respect the setting of the afterGetColHeader as it is for the activeHeaderClassName? Demo to test: https://jsfiddle.net/zfhot2w8/ (as for colHeaders)
    -  It depends on what is done in the `afterGetColHeader` callback. In the demo, we're modifying the `TH` element, not the `TH > div` element, so they're not affecting each other directly. 
    Basically, the `afterGetColHeader` is called _after_ the `headerClassName` is applied, so it can be modified in whichever way.
3. Should it apply to all of the column headers when used on the instance level, or should it have no effect (and maybe show a warning that it should used on the column level)?
    - When used on the instance level, it affects all column headers.
4. Can we reset it by setting up undefined or false within the updateSettings call?
    - Yes, setting it to an empty string or undefined resets the custom classes.

### How has this been tested?
Tested manually and added new test cases.

### Types of changes

### Related issue(s):
1. handsontable/dev-handsontable#1851

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:


handsontable/src/core.js
```diff
@@ -3325,6 +3325,16 @@ export default function Core(rootElement, userSettings, rootInstanceSymbol = fal
     });
   };
 
+  /**
+   * Returns the meta information for the provided column.
+   *
+   * @param {number} column Visual column index.
+   * @returns {object}
+   */
```
budnix, line 3333:
```suggestion
  /**
   * Returns the meta information for the provided column.
   *
   * @since 14.5
   * @memberof Core#
   * @function getColumnMeta
   * @param {number} column Visual column index.
   * @returns {object}
   */
```

handsontable/src/helpers/dom/element.js
```diff
@@ -323,6 +323,15 @@ export function removeClass(element, className) {
   });
 }
 
+/**
+ * Clear all the class names from the provided element.
+ *
+ * @param {HTMLElement} element The element to process.
+ */
+export function clearClasses(element) {
```
budnix, line 331: I'm wondering if we really want a helper for single-line code? 🤔 The advantage of helpers is that they hide more logic or polyfill something behind. It seems a bit more than what’s necessary.

handsontable/src/plugins/hiddenColumns/hiddenColumns.scss
```diff
@@ -18,6 +18,20 @@
   position: relative;
 }
 
+.handsontable[dir="ltr"] th.afterHiddenColumn div.htLeft {
+  margin-left: 10px;
```
budnix, line 22: For consistency, can I suggest using CSS logical properties such as `margin-inline-start,` etc., instead of `[dir]` rules?

handsontable/src/plugins/multiColumnSorting/multiColumnSorting.scss
```diff
@@ -1,3 +1,11 @@
+.handsontable[dir="ltr"] div.htRight span[class*="sort-"] {
+  margin-right: 20px;
```
budnix, line 2: It would be better to use CSS logical properties

handsontable/src/tableView.js
```diff
@@ -1303,13 +1304,30 @@ class TableView {
...
+      return metaHeaderClassNames ? metaHeaderClassNames.split(' ') : [];
+    };
+
     if (TH.firstChild) {
       const container = TH.firstChild;
 
       if (hasClass(container, 'relative')) {
         this.updateCellHeader(container.querySelector('.colHeader'), visualColumnIndex, label, headerLevel);
 
+        clearClasses(container);
```
budnix, line 1328:
```suggestion
        container.className = '';
```

budnix (changes requested):

handsontable/src/plugins/multiColumnSorting/multiColumnSorting.scss
jansiegel: I don't think it would work here either. In case of `ltr` we need to react to `htRight` (and set the right margin) and in case of `rtl` we need to react to `htLeft` (and set the left margin).

handsontable/src/helpers/dom/element.js
jansiegel: You're probably right 👍 When I was creating it, I assumed the code might be a little longer, but it turned out to be a one-liner.

budnix (approved):

docs/content/guides/columns/column-header/column-header.md
```diff
@@ -100,6 +100,58 @@ The [`colHeaders`](@/api/options.md#colheaders) can also be populated using a fu
 
 :::
 
+## Header labels alignment and styling
```
krzysztofspilka, line 103: Maybe: Customize column headers?

krzysztofspilka:
Looks good to me @jansiegel

PR #11126, opened 2024-07-28
Feature filter items ordering hook

sukreshmanda:
### Context
This can be useful in ordering in items to be on the filter view. The usual behaviour of the filter menu is to show Blank items on top and remaining items in lexicographical order. This hook can help in giving flexibility to provide the custom ordering.

### How has this been tested?
Local

### Types of changes

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

### DIFF
--- a/handsontable/src/__tests__/core/settings.types.ts
+++ b/handsontable/src/__tests__/core/settings.types.ts
@@ -551,6 +551,7 @@ const allSettings: Required<Handsontable.GridSettings> = {
   beforeDropdownMenuSetItems: (menuItems) => {},
   beforeDropdownMenuShow: (instance) => {},
   beforeFilter: (conditionsStack) => { conditionsStack[0].conditions[0].name === 'begins_with'; },
+  beforeFilterValueOptionsShow: (items) => {},
   beforeGetCellMeta: (row, col, cellProperties) => {},
   beforeHideColumns: (currentHideConfig, destinationHideConfig, actionPossible) => {},
   beforeHideRows: (currentHideConfig, destinationHideConfig, actionPossible) => {},
--- a/handsontable/src/pluginHooks.js
+++ b/handsontable/src/pluginHooks.js
@@ -2085,6 +2085,15 @@ const REGISTERED_HOOKS = [
    */
   'beforeStretchingColumnWidth',
 
+    /**
+   * Fired before showing filter dropdown.
+   *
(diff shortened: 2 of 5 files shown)


adrianszymanski89:
Hi @sukreshmanda 

Thank you for your contribution. We really appreciate it. We will review this PR when we finish the current tasks and decide if merging in the current form is possible.

AMBudnik:
Hi @sukreshmanda 

I did not find your CLA signature. If you did not sign the CLA, please do; without it, we won't be able to review and merge the changes.

sukreshmanda:
> Hi @sukreshmanda
> 
> I did not find your CLA signature. If you did not sign the CLA, please do; without it, we won't be able to review and merge the changes.

Done

budnix:
Hi @sukreshmanda,

Thanks for this, and sorry for the slow reply. You spotted a real gap — there is no way today to control the order of the values in the filter list. We have created a ticket on our [internal board](https://app.clickup.com/t/9015210959/DEV-2579) so the idea stays with us.

We cannot merge the branch as it stands, mostly because of how long it sat. It is from July 2024, and since then we moved the core over to TypeScript and moved the Angular wrapper into a new folder, so four of the five files you touched no longer exist. That would need rewriting rather than rebasing.

There are also a few details we would want to work through before something like this becomes public API. The value list is built in two different places in our code, and the PR hooks into one of them. In practice a custom order appears the first time you open the filter, then reverts once you confirm a selection and reopen it, and it never appears at all on a second filtered column. The handler receives a plain list of values rather than the list of objects the docs and types describe, and it is not told which column the filter was opened on — which most people would need, since the wanted order usually differs per column. The code also ignores whatever the handler returns, so writing it the natural way (take the list, return a sorted copy) does nothing, with no error or warning. And since that same list decides both what is shown and what counts as selected, dropping an entry from it would quietly remove rows from the grid, which is more reach than an ordering feature should have.

Two house rules on top of that: changes to grid code need tests, and every change needs a changelog entry. You had also ticked that documentation was needed, but none was added.

If you want to come back to it, the version that would work for us uses the value the handler returns, is called from both places where the list is built, and knows its column — with tests, a changelog entry, and docs. Happy to point you at the right files if you do.

Thanks again for raising it and for signing the CLA.

PR #11144, opened 2024-08-07
New theming system

adrianspdev:
### Context
This PR introduces a new theming system that will make it easier for users to adopt the data grid and integrate it with their existing applications and any CSS framework.

**Here’s what to expect in this phase:**

- Each theme will come with both dark and light modes, including automatic switching based on the user's preferred color scheme.
- There will be two built-in themes at first:
  - A spreadsheet-like theme called ‘main’
  - A data-grid-like theme called ‘horizon’
- The current CSS will be renamed to ‘classic’ and will continue to be fully supported.
- No breaking changes are expected.

### Related issue(s):
1. https://github.com/handsontable/dev-handsontable/issues/1291

### Affected project(s):
Handsontable, Documentation, GitHub content (like READMEs etc.)

[skip changelog]

PR #11222, opened 2024-10-15
[Themes] Change the default row height of `23px` to the actual row height.

jansiegel:
### Context
This PR extends [feature/dev-issue-1291](https://github.com/handsontable/handsontable/tree/feature/dev-issue-1291) with logic replacing the default row height of `23px` with the actual height calculated with the CSS values if a non-legacy theme is imported.

This PR is supposed to be merged with [feature/dev-issue-1291](https://github.com/handsontable/handsontable/tree/feature/dev-issue-1291).

### How has this been tested?
Tested locally on both the legacy and non-legacy themes.
Unfortunately, at this point [feature/dev-issue-1291](https://github.com/handsontable/handsontable/tree/feature/dev-issue-1291) is not yet error-free, so the CI tests on this PR will probably be red.

[skip changelog]


handsontable/src/3rdparty/walkontable/src/calculator/viewportRows.js
```diff
@@ -1,6 +1,8 @@
 import { ViewportBaseCalculator } from './viewportBase';
+import { StylesManager } from '../../../../stylesManager';
```
budnix, line 2: I think that it's generally a bad idea to import the module to Walkontable outside its root directory. Leaving it like this may result in a more complex and/or longer refactor of the module - if one is ever done. The better option would be to pass the default cell size in the Walkontable settings (TableView class).

handsontable/src/3rdparty/walkontable/src/table.js
```diff
@@ -767,6 +768,7 @@ class Table {
     let rowCount = this.TBODY.childNodes.length;
     const expectedTableHeight = rowCount * this.wtSettings.getSetting('defaultRowHeight');
     const actualTableHeight = innerHeight(this.TBODY) - 1;
+    const rowHeightFn = StylesManager.CELL_BOX_SIZING === 'border-box' ? outerHeight : innerHeight;
```
budnix, line 771: The same as the previous comment, it should be configurable throughout the Walkontable settings.

handsontable/src/stylesManager.js
```diff
@@ -3,31 +3,197 @@
  */
 export class StylesManager {
   /**
-   * The Handsontable instance.
+   * The Handsontable instance's root element.
```
budnix, line 6: I'm not sure but wouldn't it be better to have the Walkontable be responsible as a default size provider? After all, this module is responsible for rendering and is mostly interested in that default sizes.

handsontable/src/stylesManager.js
```diff
@@ -3,31 +3,197 @@
...
+   */
+  static #legacyTheme = false;
+
   /**
    * Keeps a value of cell vertical padding css variable.
    *
    * @type {number}
    */
-  cellVerticalPadding;
+  static #cellVerticalPadding;
```
budnix, line 32: Using static fields may break styling on the page where multiple table instances are created, unless I'm wrong? If that is ok I could suggest rewriting static class fields to basic variables defined above the class. I have the impression that static fields increase the size of the code making it less readable.

handsontable/src/stylesManager.js
```diff
@@ -3,31 +3,197 @@
...
+    // Some of the properties are not stored in the CSS variables, so we need to get them from the computed style.
+    this.#getCSSValuesFromTempElement(
+      { type: 'div', className: 'handsontable' },
+      { type: 'td', className: '' },
+      (computedStyle) => {
+        StylesManager.#cellBoxSizing = computedStyle['box-sizing'];
+        StylesManager.#cellTopBorderWidth = parseInt(computedStyle['border-top-width'], 10);
+        StylesManager.#cellBottomBorderWidth = parseInt(computedStyle['border-bottom-width'], 10);
+      },
     );
```
budnix, line 97: I have the impression that this method is made to be generic and it seems to me that it is not necessary (it is not used anywhere else). I would suggest something like this (pseudocode):
```js
this.#applyDynamicCSSValues(); 
```
and the method could be simplified to something like:
```js
  #applyDynamicCSSValues() {
    const parentElement = doc.createElement('div');
    parentElement.style.display = 'none';
    parentElement.className = 'handsontable';

    const childElement = doc.createElement('div');

    doc.body.appendChild(parentElement);
    parentElement.appendChild(childElement);

    const computedStyle = getComputedStyle(childElement);

    StylesManager.#cellBoxSizing = computedStyle['box-sizing'];
    StylesManager.#cellTopBorderWidth = parseInt(computedStyle['border-top-width'], 10);
    StylesManager.#cellBottomBorderWidth = parseInt(computedStyle['border-bottom-width'], 10);

    this.#rootElement.body.removeChild(parentElement);
  }
```

budnix (changes requested):

AMBudnik:
It's worth adding that we mention those 23 pixels in two places in the documentation.

### Row heights guide
https://handsontable.com/docs/javascript-data-grid/row-height/#overview

Zrzut ekranu 2024-10-15 o 15 42 44


### `rowHeights` option
https://handsontable.com/docs/javascript-data-grid/api/options/#rowheights
Zrzut ekranu 2024-10-15 o 15 43 32

handsontable/src/stylesManager.js
jansiegel: It's probably not a common use case to have multiple instances with different themes on one page, but you're right, this won't work in that scenario. I'll change it 👍

handsontable/src/tableView.js
```diff
@@ -634,6 +634,39 @@ class TableView {
...
+  /**
+   * Returns the default row height.
+   *
+   * This method retrieves the default row height from the Walkontable styles handler.
+   *
+   * @returns {number} The default row height.
+   */
+  getDefaultRowHeight() {
+    return this._wt._wot.stylesHandler.getDefaultRowHeight();
+  }
```
budnix, line 646: After updating the [facade](https://github.com/handsontable/handsontable/blob/develop/handsontable/src/3rdparty/walkontable/src/facade/core.js#L56) for missing `stylesHandler,` this should work
```suggestion
  getDefaultRowHeight() {
    return this._wt.stylesHandler.getDefaultRowHeight();
  }
```

handsontable/src/tableView.js
```diff
@@ -634,6 +634,39 @@ class TableView {
...
+   * Checks if the classic theme is being used.
+   *
+   * This method verifies if the current theme applied to the Handsontable instance
+   * is the classic theme.
+   *
+   * @returns {boolean} `true` if the classic theme is being used, `false` otherwise.
+   */
+  isUsingClassicTheme() {
+    return this._wt._wot.stylesHandler.isClassicTheme();
+  }
```
budnix, line 668:
```suggestion
  isUsingClassicTheme() {
    return this._wt.stylesHandler.isClassicTheme();
  }
```

handsontable/src/3rdparty/walkontable/src/utils/stylesHandler.js
```diff
@@ -0,0 +1,182 @@
...
+        ...{
+          td: {
+            'box-sizing': computedStyle['box-sizing'],
+            'border-top-width': parseInt(computedStyle['border-top-width'], 10),
+            'border-bottom-width': parseInt(computedStyle['border-bottom-width'], 10),
+          },
+        },
+      };
+    });
+  }
```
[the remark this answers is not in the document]
jansiegel: I did it that way, because the computed styles object that would be returned from the function is modified after the function removes the temporary elements: https://jsfiddle.net/js_ziggle/m1kz078j/

I can rewrite the method so it returns just the css values requested as the function argument 👍

handsontable/src/3rdparty/walkontable/src/types.js
```diff
@@ -66,6 +66,8 @@
  * @property {Walkontable} cloneSource CloneSource.
  * @property {Walkontable} wot Wot.
  * @property {number} parentTableOffset ParentTableOffset.
+ * @property {StylesHandler} stylesHandler StylesHandler.
+ * @property {SelectionManager} selectionManager SelectionManager.
```
budnix, line 70:
```suggestion
 * @property {StylesHandler} stylesHandler StylesHandler.
```

handsontable/src/3rdparty/walkontable/src/utils/stylesHandler.js
```diff
@@ -0,0 +1,188 @@
...
+   *
+   * @param {object} domBindings - The DOM bindings for the instance.
+   */
+  constructor(domBindings) {
+    this.#rootElement = domBindings.rootTable.parentElement;
+    this.#rootComputedStyle = getComputedStyle(this.#rootElement);
+    this.#rootDocument = domBindings.rootDocument;
+
+    // TODO: naive method of checking for modern themes
+    if (this.#rootComputedStyle?.getPropertyValue('--ht-line-height') === '') {
```
budnix, line 64: The computed style always exists. I guess it's not necessary to use optional chaining.
```suggestion
    if (this.#rootComputedStyle.getPropertyValue('--ht-line-height') === '') {
```

handsontable/src/3rdparty/walkontable/src/utils/stylesHandler.js
```diff
@@ -0,0 +1,188 @@
...
+
+  /**
+   * Parses the value of a specified CSS property from the root element's computed style.
+   *
+   * @param {string} property - The CSS property to retrieve and parse.
+   * @returns {number|string} The parsed value of the CSS property.
+   */
+  #getParsedCSSValue(property) {
+    return parseInt(this.#rootComputedStyle?.getPropertyValue(property), 10);
+  }
```
budnix, line 187: The `getPropertyValue` method returns an empty string for non-existing property, which, after casting to a number, returns `NaN.` WDYT about improving the method?
```suggestion
  /**
   * Parses the value of a specified CSS property from the root element's computed style.
   *
   * @param {string} property - The CSS property to retrieve and parse.
   * @returns {number|undefined} The parsed value of the CSS property.
   */
  #getParsedCSSValue(property) {
    const parsedValue = parseInt(this.#rootComputedStyle.getPropertyValue(property), 10);
    
    return Number.isNaN(parsedValue) ? null : parsedValue;
  }
```
then it is worth updating the `getCSSVariableValue` method with the changes made here.

handsontable/src/3rdparty/walkontable/src/utils/stylesHandler.js
```diff
@@ -0,0 +1,190 @@
...
+      return this.#cssVars[`--ht-${variableName}`];
+    }
+
+    const acquiredValue = this.#getParsedCSSValue(`--ht-${variableName}`);
+
+    if (acquiredValue !== undefined && acquiredValue !== null) {
+      this.#cssVars[variableName] = acquiredValue;
+
+      return acquiredValue;
+    }
```
budnix, line 110:
```suggestion
    const acquiredValue = this.#getParsedCSSValue(`--ht-${variableName}`);

    if (acquiredValue !== null) {
      this.#cssVars[`--ht-${variableName}`] = acquiredValue;

      return acquiredValue;
    }
```

handsontable/src/3rdparty/walkontable/src/utils/stylesHandler.js
```diff
@@ -0,0 +1,190 @@
...
+   */
+  isClassicTheme() {
+    return this.#isClassicTheme;
+  }
+
+  /**
+   * Retrieves the value of a specified CSS variable.
+   *
+   * @param {string} variableName - The name of the CSS variable to retrieve.
+   * @returns {number|string|undefined} The value of the specified CSS variable, or `undefined` if not found.
```
budnix, line 97:
```suggestion
   * @returns {number|undefined} The value of the specified CSS variable, or `undefined` if not found.
```

budnix (changes requested):

handsontable/src/3rdparty/walkontable/src/utils/stylesHandler.js
```diff
@@ -0,0 +1,192 @@
...
+   * @param {object} domBindings - The DOM bindings for the instance.
+   */
+  constructor(themeName, domBindings) {
+    this.#rootElement = domBindings.rootTable.parentElement;
+    this.#rootComputedStyle = getComputedStyle(this.#rootElement);
+    this.#rootDocument = domBindings.rootDocument;
+
+    if (!themeName) {
+      this.#isClassicTheme = true;
+    }
```
budnix, line 66: I think here is a bug. When you initialize the table without defining `themeName` option, the module mark it as "classic" but after updating table settings via `updateSettings` method the flag won't be updated anymore.

handsontable/types/settings.d.ts
```diff
@@ -214,6 +214,7 @@ export interface GridSettings extends Events {
   strict?: boolean;
   tableClassName?: string | string[];
   tabMoves?: CellCoords | SimpleCellCoords | ((event: KeyboardEvent) => CellCoords | SimpleCellCoords);
+  themeName?: string;
```
budnix, line 217: Can I ask you to add tests for that option?

handsontable/src/3rdparty/walkontable/src/renderer/table.js
```diff
@@ -297,14 +297,17 @@ export default class TableRenderer {
     // Fix for multi-line content and for supporting `rowHeights` option.
     for (let visibleRowIndex = 0; visibleRowIndex < rowsToRender; visibleRowIndex++) {
       const TR = rows.getRenderedNode(visibleRowIndex);
+      const rowUtils = this.rowUtils;
 
       if (TR.firstChild) {
         const sourceRowIndex = this.renderedRowToSource(visibleRowIndex);
-        const rowHeight = this.rowUtils.getHeightByOverlayName(sourceRowIndex, this.activeOverlayName);
+        const rowHeight = rowUtils.getHeightByOverlayName(sourceRowIndex, this.activeOverlayName);
+        const isBorderBoxSizing = rowUtils.dataAccessObject.stylesHandler.areCellsBorderBox();
```
budnix, line 305: You should not use the `dataAccessObject` of the `rowUtils` module as it's its "private" DAO. I'd recommend injecting the StylesHandler separately into the Renderer module.

budnix (changes requested):

handsontable/types/settings.d.ts
jansiegel: This option was introduced in https://github.com/handsontable/handsontable/pull/11144, so I guess the typing tests should also be added there? I probably shouldn't have added the type here either 🙄

budnix (changes requested):

handsontable/src/3rdparty/walkontable/src/utils/stylesHandler.js
```diff
@@ -0,0 +1,268 @@
...
+        'box-sizing': stylesForTD['box-sizing'],
+      },
+    };
+  }
+
+  /**
+   * Retrieve the theme class name from the provided class name string.
+   *
+   * @param {string} className Class name string.
+   * @returns {string}
```
budnix, line 196:
```suggestion
   * @returns {string|boolean}
```

handsontable/src/plugins/stretchColumns/calculator.js
```diff
@@ -121,11 +121,12 @@ export class StretchCalculator {
 
     const viewportHeight = view.getViewportHeight();
     const totalRows = this.#hot.countRows();
+    const defaultRowHeight = this.#hot.view.getStylesHandler().getDefaultRowHeight();
```
budnix, line 124:
```suggestion
    const defaultRowHeight = view.getStylesHandler().getDefaultRowHeight();
```

handsontable/src/tableView.js
```diff
@@ -654,6 +701,7 @@ class TableView {
    */
   initializeWalkontable() {
     const walkontableConfig = {
+      themeName: this.settings.themeName,
```
budnix, line 704: Still needed?

budnix (approved):

PR #11238, opened 2024-10-29
Date validator: Add support for strict format checking

rmanganiello:
### Context
Currently, `correctFormat` function is used to correct the format of the input date but it does not do a strict format check while trying to parse the date with MomentJS, which can lead to unexpected results.

Moment.js recommends to use strict format checking when parsing dates if possible.

[MomentJS Docs](https://momentjs.com/guides/#/parsing/strict-mode/)

- Add `strict` parameter to the `correctFormat` function to allow strict format checking.
- Add  optional`isCorrectFormatStrict` parameter to `GridSettings`. Defaults to `false`

### How has this been tested?
I added some e2e tests with the corresponding explanation, mainly for some wrong scenarios with the old approach.

### Types of changes

### Affected project(s):
- [x] `handsontable`
- [] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

### DIFF
--- a/docs/content/guides/cell-types/date-cell-type/date-cell-type.md
+++ b/docs/content/guides/cell-types/date-cell-type/date-cell-type.md
@@ -92,6 +92,31 @@ correctFormat={false}
 correctFormat={true}
 ```
 
+Also, you can set `isCorrectFormatStrict` to `true` to make the date validation stricter when trying to correct the input format.
+For more information check [Moment.js documentation](https://momentjs.com/guides/#/parsing/strict-mode/).
+
+```js
+dateFormat: 'D/M/YY',
+
+// default behavior
+// date entered as `01/02/2023` will be formatted as `1/2/20` which is incorrect
+isCorrectFormatStrict={false}
+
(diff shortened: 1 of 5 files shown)


AMBudnik:
Hi @rmanganiello 

Thank you for sharing the idea and signing our CLA. There's one more thing missing, though. It seems that point `9` from https://github.com/handsontable/handsontable/blob/master/CONTRIBUTING.md is missing - Changelog entry. 

When it comes to the idea itself, it is interesting, and I see the benefit of adding that new option. However, each new option (method or hook) has to be approved at the top. So it may take some time for us to review it. Nevertheless, we will keep you updated.

rmanganiello:
Thanks for the reply @AMBudnik.

Please let me know whenever you have the chance to review this and include it since the current behavior of `correctFormat` has some side effects with some date formats because we lack this parameter.

I just created a new issue (#11245) to add the corresponding changelog entry.

Thank you.

PR #11246, opened 2024-10-31
[themes] Modify the themes config priority.

jansiegel:
### Context
This PR:
- switches the themes configuration method priority,
    - Before:
    `themeName` passed in the config options to either initialization or `updateSettings` would be overridden by the `ht-theme-*` class name added to the table's root element.
    - After:
    `themeName` passed in the config options to either initialization or `updateSettings` overrides the `ht-theme-*` class name added to the table's root element.

    The change is introduced because otherwise `updateSettings` could not update the theme with the `themeName` option (setting the theme in any way would apply the `ht-theme-*` class name to the root element, which would eventually override any option passed to `updateSettings`).
- corrects some problems with calling `updateSettings` with the `themeName` option.

### How has this been tested?
Added several new test cases.

[skip changelog]

**This PR is meant to be merged to the `feature/dev-issue-1291` branch**


budnix (changes requested):
It's all good. I left suggestions to consider mostly.

budnix (approved):

PR #11253, opened 2024-11-06
[themes] Fix a problem where assigning the theme class without importing the correct CSS files would block the table from being edited.

jansiegel:
### Context
This PR:
1. In a case where the theme class name was applied to the Handsontable container (or any of its parents), but the classic CSS files were imported instead of the "modern" ones:
    1. Makes it switch back to the "classic" theme internally.
    2. Displays a warning in the console.
2. Modifies the test cases for the new logic.
3. Documents the `themeName` option.
4. Modifies the JSDocs of other options, where the obsolete mentions of `23px` being the default height were still present. 

This branch is meant to be merged to `feature/dev-issue-1291`.
[skip changelog]

### DIFF
--- a/handsontable/src/3rdparty/walkontable/src/utils/stylesHandler.js
+++ b/handsontable/src/3rdparty/walkontable/src/utils/stylesHandler.js
@@ -125,7 +125,7 @@ export class StylesHandler {
 
     const cssVarRowHeightValue = this.getCSSVariableValue('row-height');
 
-    if (!this.#isClassicTheme && !cssVarRowHeightValue) {
+    if (!cssVarRowHeightValue) {
       warn(`The "${this.#themeName}" theme is enabled, but its stylesheets are missing. \
 Import the correct CSS files in order to use that theme.`);
 
@@ -135,7 +135,7 @@ Import the correct CSS files in order to use that theme.`);
       return CLASSIC_THEME_DEFAULT_HEIGHT;
     }
 
-    return this.getCSSVariableValue('row-height');
+    return cssVarRowHeightValue;
   }
 
   /**
--- a/handsontable/src/3rdparty/walkontable/src/utils/stylesHandler.js
+++ b/handsontable/src/3rdparty/walkontable/src/utils/stylesHandler.js
@@ -119,6 +119,10 @@ export class StylesHandler {
    * @returns {number} The calculated row height.
    */
   getDefaultRowHeight() {
+    if (this.#isClassicTheme) {
+      return CLASSIC_THEME_DEFAULT_HEIGHT;
+    }
+
     const cssVarRowHeightValue = this.getCSSVariableValue('row-height');
 
     if (!this.#isClassicTheme && !cssVarRowHeightValue) {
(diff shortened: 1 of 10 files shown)


handsontable/src/dataMap/metaManager/metaSchema.js
```diff
@@ -4435,19 +4435,22 @@ export default () => {
...
+     * The `themeName` option allows enabling a theme by that name.
+     *
+     * If no `themeName` is provided, the table will use the classic theme (if the correct CSS files are imported).
+     *
+     * Read more:
+     * - [Themes](@/guides/styling/themes/themes.md)
      *
      * @memberof Options#
-     * @type {string|string[]}
+     * @type {string}
```
budnix, line 4447: I guess there is more?
```suggestion
     * @type {string | boolean | undefined}
```
Can I switch to the classic theme by passing `false`?

handsontable/src/3rdparty/walkontable/src/utils/stylesHandler.js
```diff
@@ -118,6 +119,18 @@ export class StylesHandler {
...
+
+      this.#isClassicTheme = true;
+      this.useTheme();
+
+      return CLASSIC_THEME_DEFAULT_HEIGHT;
+    }
+
     if (this.#isClassicTheme) {
       return CLASSIC_THEME_DEFAULT_HEIGHT;
     }
```
budnix, line 136: Wouldn't it be more efficient if you returned a default height of the row for the classic theme ASAP and then processed the rest? 
```suggestion
    if (this.#isClassicTheme) {
      return CLASSIC_THEME_DEFAULT_HEIGHT;
    }
    
    const cssVarRowHeightValue = this.getCSSVariableValue('row-height');

    if (!this.#isClassicTheme && !cssVarRowHeightValue) {
      warn(`The "${this.#themeName}" theme is enabled, but its stylesheets are missing. \
Import the correct CSS files in order to use that theme.`);

      this.#isClassicTheme = true;
      this.useTheme();

      return CLASSIC_THEME_DEFAULT_HEIGHT;
    }
```

budnix (changes requested):

handsontable/src/3rdparty/walkontable/src/utils/stylesHandler.js
```diff
@@ -122,6 +123,18 @@ export class StylesHandler {
...
 
+    const cssVarRowHeightValue = this.getCSSVariableValue('row-height');
+
+    if (!this.#isClassicTheme && !cssVarRowHeightValue) {
```
budnix, line 128:
```suggestion
    if (!cssVarRowHeightValue) {
```

handsontable/src/3rdparty/walkontable/src/utils/stylesHandler.js
```diff
@@ -122,6 +123,18 @@ export class StylesHandler {
...
+      warn(`The "${this.#themeName}" theme is enabled, but its stylesheets are missing. \
+Import the correct CSS files in order to use that theme.`);
+
+      this.#isClassicTheme = true;
+      this.useTheme();
+
+      return CLASSIC_THEME_DEFAULT_HEIGHT;
+    }
+
     return this.getCSSVariableValue('row-height');
```
budnix, line 138:
```suggestion
    const cssVarRowHeightValue = this.getCSSVariableValue('row-height');

    if (!cssVarRowHeightValue) {
      warn(`The "${this.#themeName}" theme is enabled, but its stylesheets are missing. \
Import the correct CSS files in order to use that theme.`);

      this.#isClassicTheme = true;
      this.useTheme();

      return CLASSIC_THEME_DEFAULT_HEIGHT;
    }

    return cssVarRowHeightValue;
```

budnix (approved):

PR #11274, opened 2024-11-21
Fix .nvmrc file path for Windows env

budnix:
### Context

Revert the changes about Node version for Windows job.

_[skip changelog]_

### Types of changes

### DIFF
--- a/.github/workflows/build-all.yml
+++ b/.github/workflows/build-all.yml
@@ -45,9 +45,6 @@ jobs:
     name: Build all packages in windows-latest
     runs-on: windows-latest
     timeout-minutes: 60
-    defaults:
-      run:
-        working-directory: C:/handsontable
     steps:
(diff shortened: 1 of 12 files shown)


.github/workflows/build-all.yml
```diff
@@ -52,10 +53,10 @@ jobs:
           mkdir C:/handsontable
           git clone -b $env:GITHUB_REF_NAME https://github.com/handsontable/handsontable C:/handsontable --depth 1
 
-      - name: Use Node.js ${{ env.NODE_VERSION }}
+      - name: Use Node.js
         uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # https://github.com/actions/setup-node/releases/tag/v4.0.2
         with:
-          node-version-file: '.nvmrc'
```
magierg, line 58: maybe would be enough to add  ` working-directory: C:/handsontable` as it's not able to find the file

budnix: I tried without success

jansiegel (approved):

PR #11469, opened 2025-02-26
Improve autocomplete/dropdown editor list behavior

budnix:
### Context

The PR improves the autocomplete/dropdown editor list behavior, which, from now on, correctly flips the list position above the edited cell when there is no space room left at the bottom.

#### Before
Kapture 2025-02-26 at 12 14 43

#### After
Kapture 2025-02-26 at 12 15 54

### How has this been tested?

I tested the changes locally and I covered the fix with new tests.

### Types of changes

### Related issue(s):
1. fixes https://github.com/handsontable/dev-handsontable/issues/802

### Affected project(s):
- [x] `handsontable`

### Checklist:


handsontable/src/editors/autocompleteEditor/__tests__/autocompleteEditor.spec.js
```diff
@@ -1054,94 +1053,216 @@ describe('AutocompleteEditor', () => {
...
-            editor: 'autocomplete',
-            source: choices
-          }, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}
-        ],
+        data: createEmptySpreadsheetData(30, 30),
+        editor: 'autocomplete',
+        source: choices,
         width: 400,
-        height: 400
+        height: getDefaultRowHeight() * 9,
```
qunabu, line 1062: what is `9` here?

handsontable/src/editors/autocompleteEditor/__tests__/autocompleteEditor.spec.js
```diff
@@ -1054,94 +1053,216 @@ describe('AutocompleteEditor', () => {
...
       });
 
-      setDataAtCell(29, 0, '');
-      selectCell(29, 0);
+      mouseDoubleClick($(getCell(6, 0)));
 
-      mouseDoubleClick($(getCell(29, 0)));
+      await sleep(50);
```
qunabu, line 1067: this could lead to a flaky test; maybe there is a better solution than this

handsontable/src/editors/autocompleteEditor/__tests__/autocompleteEditor.spec.js
```diff
@@ -1054,94 +1053,216 @@ describe('AutocompleteEditor', () => {
...
 
-      await sleep(200);
+      const container = $(getActiveEditor().htContainer);
 
-      const autocompleteEditor = $('.autocompleteEditor');
+      expect(container.offset()).forThemes(({ classic, main }) => {
+        classic.toEqual({ top: 23, left: 0 });
+        main.toEqual({ top: 29, left: 0 });
```
qunabu, line 1073: shouldn't `23` and `29` be constant shared between tests? shouldn't `getDefaultRowHeight` be used here ?

handsontable/src/editors/autocompleteEditor/__tests__/autocompleteEditor.spec.js
budnix: The tests - to be the same for each theme, needs to be set up with the same setup. In this case the viewport height is "theme default height" x 9.

handsontable/src/editors/autocompleteEditor/__tests__/autocompleteEditor.spec.js
budnix: I found no better solution for that 🤔

handsontable/src/editors/autocompleteEditor/__tests__/autocompleteEditor.spec.js
budnix: Fixed aea878d054d4a36bb3e915f4bcfaf799976fe4af

handsontable/src/editors/autocompleteEditor/__tests__/autocompleteEditor.spec.js
budnix: I've added code comment 94fc8a4849f478d4be40d17086cc5c29801cae49

handsontable/test/helpers/common.js
```diff
@@ -184,6 +184,20 @@ export const useTheme = handsontableMethodFactory('useTheme');
 export const validateCell = handsontableMethodFactory('validateCell');
 export const validateCells = handsontableMethodFactory('validateCells');
 
+/**
+ * @returns {number} Returns the default row height based on the current theme.
+ */
+export function getDefaultRowHeight() {
```
adrianspdev, line 190: If possible, we should consider using a dynamic value based on the row height calculation, as shown here: https://github.com/handsontable/handsontable/blob/a28f5295d9020215906758458713fe428d7bef2e/handsontable/src/3rdparty/walkontable/src/utils/stylesHandler.js#L123

adrianspdev (changes requested):

handsontable/test/helpers/common.js
budnix: Good catch! I thought about the same thing earlier, but after thinking it over, I decided not to introduce the dynamic calculations because:
 * I can't use Handsontable API here as the function is used before the table initialization. I would have to copy the entire functionality of the `StylesHandler` module and implement it here in the helper, and still in some test cases where e.g. many HoT instances are tested in one test, it would not work;
 * Ultimately, tests should be tested based on the theme height that we expect (const) and not one that is calculated dynamically and can change without our knowledge;
 * If the default row height changes in the future this will be the only place to update;

All these points convinced me not to add dynamic calculations here. WDYT?

adrianspdev (approved):

PR #11470, opened 2025-02-26
Update theme variables and add input radio variables

adrianspdev:
### Context
This PR includes changes and improvements for themes css variables.

### How has this been tested?
Locally

### Types of changes

### Related issue(s):
1. https://github.com/handsontable/dev-handsontable/issues/2262

### Affected project(s):
- [x] `handsontable`

### Checklist:


handsontable/src/styles/components/form-elements/_radio.scss
```diff
@@ -21,65 +21,72 @@
...
         position: relative;
-        width: var(--ht-checkbox-size);
-        height: var(--ht-checkbox-size);
+        width: var(--ht-radio-size);
+        height: var(--ht-radio-size);
         margin: 0;
         appearance: none;
         cursor: pointer;
+        border-radius: calc(0.5 * var(--ht-radio-size));
+        outline: none;
```
qunabu, line 30: Is it `outline: none` against WCAG recommendation ?
Do we have provided an "provide alternative styling" ? I see below that `:focus` is covered with  `background-color` but I'm not an WCAG expert, my question is, is this enough ?

References 
* https://www.w3.org/TR/WCAG20-TECHS/F78.html
* https://www.outlinenone.com/

handsontable/src/styles/components/form-elements/_radio.scss
```diff
@@ -21,65 +21,72 @@
 
       > input[type="radio"] {
         position: relative;
-        width: var(--ht-checkbox-size);
-        height: var(--ht-checkbox-size);
+        width: var(--ht-radio-size);
+        height: var(--ht-radio-size);
```
budnix, line 25: Changing the variable name could be treated as a breaking change. Am I right?

handsontable/src/styles/components/form-elements/_radio.scss
adrianspdev: I guess in this case, yes, we have the same behavior for checkboxes. As you mentioned, we’ve covered the outline inside the pseudo-element, which is the visual representation of our input. Ps. you can set outline color via `--ht-radio-focus-ring-color` variable.

budnix (changes requested):
The visual test will catch that because I found a regression. It seems that row headers lost their background colors.
Kapture 2025-02-26 at 12 44 13

qunabu (approved):

PR #11473, opened 2025-02-27
Fix problem with horizontal scroll in autocomplete editor dropdown when table overflow is hidden

adrianspdev:
### Context
This PR includes fixes for horizontal scroll in autocomplete editor list when `overflow: hidden` is applied to the `wrapper` element.

### How has this been tested?
Locally and covered the fix with new tests.

### Types of changes

### Related issue(s):
1. https://github.com/handsontable/dev-handsontable/issues/2258

### Affected project(s):
- [x] `handsontable`

### Checklist:

### DIFF
--- a/handsontable/src/editors/autocompleteEditor/__tests__/autocompleteEditor.spec.js
+++ b/handsontable/src/editors/autocompleteEditor/__tests__/autocompleteEditor.spec.js
@@ -529,6 +529,36 @@ describe('AutocompleteEditor', () => {
       });
     });
 
+    it('should open editor with the correct size when there is scrollbar on the list and table overflow is hidden', async() => {
+      handsontable({
+        colWidths: 120,
+        columns: [
+          {
+            editor: 'autocomplete',
+            source: choices,
+            visibleRows: choices.length,
+          }
+        ],
+        height: 'auto'
+      });
+
+      selectCell(0, 0);
+      keyDownUp('enter');
+
+      await sleep(100);
+
+      const container = getActiveEditor().htContainer;
+
+      expect(container.clientWidth).forThemes(({ classic, main }) => {
+        classic.toBe(120 + Handsontable.dom.getScrollbarWidth());
+        main.toBe(118 + Handsontable.dom.getScrollbarWidth());
+      });
+      expect(container.clientHeight).forThemes(({ classic, main }) => {
+        classic.toBe(69);
+        main.toBe(87);
+      });
+    });
+
     it('should open editor with the correct size when there is scrollbar on the list (trimDropdown: false)', async() => {
       handsontable({
         colWidths: 120,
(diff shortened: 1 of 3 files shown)


handsontable/src/editors/autocompleteEditor/__tests__/autocompleteEditor.spec.js
```diff
@@ -529,6 +529,36 @@ describe('AutocompleteEditor', () => {
...
+      const container = getActiveEditor().htContainer;
+
+      expect(container.clientWidth).forThemes(({ classic, main }) => {
+        classic.toBe(120 + Handsontable.dom.getScrollbarWidth());
+        main.toBe(118 + Handsontable.dom.getScrollbarWidth());
+      });
+      expect(container.clientHeight).forThemes(({ classic, main }) => {
+        classic.toBe(69);
+        main.toBe(87);
+      });
```
qunabu, line 559: where are all those numbers coming from? shouldn't they be imported as constants?

handsontable/src/editors/autocompleteEditor/__tests__/autocompleteEditor.spec.js
```diff
@@ -529,6 +529,36 @@ describe('AutocompleteEditor', () => {
...
+            visibleRows: choices.length,
+          }
+        ],
+        height: 'auto'
+      });
+
+      selectCell(0, 0);
+      keyDownUp('enter');
+
+      await sleep(100);
```
qunabu, line 548: I know that this is a tricky case, but using sleep in tests is a bad practice that can lead to flaky tests

budnix (changes requested):
It looks good. However, I think that there can be implemented a more general solution to the fix. WDYT?

handsontable/src/editors/autocompleteEditor/__tests__/autocompleteEditor.spec.js
adrianspdev: those numbers come from outcome of test results and we don't have any constants declare for this particular case

handsontable/src/editors/autocompleteEditor/__tests__/autocompleteEditor.spec.js
adrianspdev: We have some timeout set, so we need to wait to ensure everything is called before check the results

budnix (approved):

handsontable/src/editors/autocompleteEditor/autocompleteEditor.js
```diff
@@ -432,6 +431,17 @@ export class AutocompleteEditor extends HandsontableEditor {
     this.htEditor.flipped = undefined;
   }
 
+  /**
+   * Fix width of the internal Handsontable's instance when editor has vertical scroll.
+   */
+  #fixDropdownWidth() {
```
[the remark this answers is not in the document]
budnix: I think that the `updateDropdownWidth` name may be misleading. It does not update the `width` as-is but rather compensates (or fixes) the width caused by the updated `height` of the component.

Issue #8118, opened 2021-05-27
[9.0.0+] Handsontable init time is clearly longer with new Formulas plugin


budnix:
### Description

I've checked the performance of the new Formula plugin and didn't found any issues except that the HoT initialization delay time increased from several to several times longer compared to the previous implementation. The bigger dataset, the slower the HoT initializes. The similar results can be see using the `loadData` method.

v8.4.0:
Init takes ~270 ms
loadData takes ~207 ms

v9.0.0:
Init takes ~432 ms
loadData takes ~320 ms

_Edit:_
After investigation the issue, it seems that the bottleneck mostly sits in the HF engine.
```js
const DEFAULT_SETTINGS = {
  binarySearchThreshold: 20,
  matrixDetection: false,
  matrixDetectionThreshold: 100,
  useColumnIndex: false,
  useStats: false,
  evaluateNullToZero: true,
  precisionEpsilon: 1e-13,
  precisionRounding: 14,
  smartRounding: true,
  leapYear1900: true,
  nullDate: {
    year: 1899,
    month: 12,
    day: 31
  },
  nullYear: 30,
  dateFormats: ['DD/MM/YYYY', 'DD/MM/YY'],
  timeFormats: ['hh:mm', 'hh:mm:ss.sss'],
  matchWholeCell: true,
  useRegularExpressions: false,
  useWildcards: true,
  functionArgSeparator: ',',
  thousandSeparator: '',
  decimalSeparator: '.',
  language: 'enGB',
};

const data = Handsontable.helper.createSpreadsheetData(1000, 1000);

console.time('hf');
HyperFormula.buildFromArray(data, DEFAULT_SETTINGS);
console.timeEnd('hf');
// Gives me about 7s
```

### Demo

https://jsfiddle.net/budnix/Lgj9erk1/ 8.4.0
https://jsfiddle.net/budnix/ak37jwyr/ 9.0.0

### Your environment
* Handsontable version: 9.0.0
* Browser Name and version: -
* Operating System: -

robertovaldesperez:
I'm using the last version and hyperformula is very slow.

sequba:
We have a number of issues in HyperFormula related to improving performance, some of which are being worked on currently. The performance should be re-verified after completing some of these issues:

In the scope of the next release:
- https://github.com/handsontable/hyperformula/issues/896
- https://github.com/handsontable/hyperformula/issues/973
- https://github.com/handsontable/hyperformula/issues/1011

Other:
[...]

PR #11474, opened 2025-02-27
Improve the initialization time of Handsontable with formulas enabled

sequba:
### Description

Optimizations introduced:
- Prevent `AutoColumnSize.#onBeforeRender` from recalculating column widths when they are already calculated by `AutoColumnSize.#onInit`
- Optimize `AxisSyncer.getHfIndexFromVisualIndex` -> set the start index for the `indexOf` linear search
- Store `sheetId` inside `FormulasPlugin` instead of getting it from `HyperFormula` every time
- Move the initialization of HF from  `Formulas.#onAfterLoadData` to `Formulas.#onAfterCellMetaReset` to avoid duplicated runs of `Handsontable.getSourceDataArray` and `HyperFormula.setSheetContent`

Analyzed:
- `Formulas.#onAfterCellMetaReset` -> turns out it must be called because it is important for handling dates. I tried changing the order of operations in it to make it more efficient but with no significant effect
- `FormulaPlugin`'s calls to `Handsontable.getSourceDataArray` -> unfortunately, the operations performed by this function are necessary

Other changes:
- Introduced `skipMetaExtension` option for `getCellMeta` function
- Remove the workaround for `getCellMeta` from `Formulas.#onModifyData`
- Refactored `AutoColumnSize` class. Extracted the duplicated code from methods `#calculateSpecificColumnsWidth` and `calculateColumnsWidth`
- Refactor `Formulas.#onModifyData` and `#onModifySourceData` methods
- Adjusted the typings for `getCellMeta` and for `Formulas` plugin
- Added relevant e2e tests

### The performance gain after changes

| HOT with formulas enabled and auto column size (180k cells) | Init time |
| - | - |
| Before | 4162ms |
| After | 2291ms |

**This PR speeds up the tested scenario by 45%**

### How has this been tested?

- [x] unit tests
- [x] manual verification of demos from https://github.com/handsontable/dev-handsontable/issues/579

### Types of changes

### Related issue(s):
1. https://github.com/handsontable/dev-handsontable/issues/579
2. https://github.com/handsontable/handsontable/issues/8118
3. https://github.com/handsontable/dev-handsontable/issues/2078

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:


handsontable/src/plugins/formulas/__tests__/formulas.spec.js
```diff
@@ -541,6 +541,27 @@ describe('Formulas general', () => {
     });
   });
 
+  it('should ??? while using `getDataAtCell` inside `cells` method with custom dateFormat', () => {
```
sequba, line 544: TODO: decide what to do in such case

qunabu (approved):
seems like big update for formula plugin

handsontable/src/core.js
```diff
@@ -2546,7 +2546,7 @@ export default function Core(rootElement, userSettings, rootInstanceSymbol = fal
       });
     }
 
-    instance.runHooks('afterCellMetaReset');
+    instance.runHooks('afterCellMetaReset', 'updateSettings');
```
budnix, line 2549: This hook runs only once when the `updateSettings` is called and only here. I don't see the point of adding a new argument to the method because there would be nothing to compare it to. The `source` would always be the same.

handsontable/src/core.js
```diff
@@ -3350,11 +3350,13 @@ export default function Core(rootElement, userSettings, rootInstanceSymbol = fal
...
    * @param {number} row Visual row index.
    * @param {number} column Visual column index.
+   * @param {object} options Execution options for the `getCellMeta` method.
+   * @param {boolean} [options.executeCellsFunction] If `false`, disables the execution of the `cells` function. Default: `true`.
    * @returns {object} The cell properties object.
    * @fires Hooks#beforeGetCellMeta
    * @fires Hooks#afterGetCellMeta
    */
-  this.getCellMeta = function(row, column) {
+  this.getCellMeta = function(row, column, options = { executeCellsFunction: true }) {
```
budnix, line 3359: I'd propose a different name for the option. For me, it's not about preventing the call of a `cells` function but rather preventing the execution of the [_DynamiCellMeta_](https://github.com/handsontable/handsontable/blob/develop/handsontable/src/dataMap/metaManager/mods/dynamicCellMeta.js) mod module.
```suggestion
  this.getCellMeta = function(row, column, options = { skipMetaExtension: false }) {
```

handsontable/src/core.js
```diff
@@ -3369,7 +3371,7 @@ export default function Core(rootElement, userSettings, rootInstanceSymbol = fal
     return metaManager.getCellMeta(physicalRow, physicalColumn, {
       visualRow: row,
       visualColumn: column,
-    });
+    }, options);
```
budnix, line 3374: The 3rd argument of the `getCellMeta` is already an option designed for extending the meta. No need to create 4th one.
``` js
return metaManager.getCellMeta(physicalRow, physicalColumn, {
      visualRow: row,
      visualColumn: column,
      ...options,
    });
```

handsontable/src/dataMap/metaManager/index.js
```diff
@@ -133,20 +133,22 @@ export default class MetaManager {
...
-   * @param {number} options.visualColumn The visual column index of the currently requested cell meta object.
+   * @param {object} additionalCellMeta Additional options that are used to extend the cell meta object.
+   * @param {number} additionalCellMeta.visualRow The visual row index of the currently requested cell meta object.
+   * @param {number} additionalCellMeta.visualColumn The visual column index of the currently requested cell meta object.
+   * @param {object} options Execution options for the `getCellMeta` method.
+   * @param {boolean} [options.executeCellsFunction] If `false`, disables the execution of the `cells` function. Default: `true`.
    * @returns {object}
    */
-  getCellMeta(physicalRow, physicalColumn, { visualRow, visualColumn }) {
+  getCellMeta(physicalRow, physicalColumn, { visualRow, visualColumn }, options = { executeCellsFunction: true }) {
```
budnix, line 143:
```suggestion
  getCellMeta(physicalRow, physicalColumn, { visualRow, visualColumn, skipMetaExtension }) {
```

handsontable/src/dataMap/metaManager/index.js
```diff
@@ -133,20 +133,22 @@ export default class MetaManager {
...
     const cellMeta = this.cellMeta.getMeta(physicalRow, physicalColumn);
 
     cellMeta.visualRow = visualRow;
     cellMeta.visualCol = visualColumn;
     cellMeta.row = physicalRow;
     cellMeta.col = physicalColumn;
 
-    this.runLocalHooks('afterGetCellMeta', cellMeta);
+    this.runLocalHooks('afterGetCellMeta', cellMeta, options);
```
budnix, line 151: I'd not extend the hook with the 3th option but rather go with updating the cellMeta directly. Something like
```js
    cellMeta.visualRow = visualRow;
    cellMeta.visualCol = visualColumn;
    cellMeta.row = physicalRow;
    cellMeta.col = physicalColumn;
    cellMeta._skipMetaExtension = skipMetaExtension;

    this.runLocalHooks('afterGetCellMeta', cellMeta);
```

handsontable/src/dataMap/metaManager/mods/dynamicCellMeta.js
```diff
@@ -49,8 +49,10 @@ export class DynamicCellMetaMod {
    * To boost performance, the extending process is triggered only once per one slow Handsontable render cycle.
    *
    * @param {object} cellMeta The cell meta object.
+   * @param {object} options Execution options for the `extendCellMeta` method.
+   * @param {boolean} [options.executeCellsFunction] If `false`, disables the execution of the `cells` function. Default: `true`.
    */
-  extendCellMeta(cellMeta) {
+  extendCellMeta(cellMeta, options = { executeCellsFunction: true }) {
```
budnix, line 55: With the suggestions I put above, skipping executions could look like this
```js
  extendCellMeta(cellMeta) {
    if (cellMeta._skipMetaExtension) {
       cellMeta._skipMetaExtension = false; 

       return;
    }
```
... then the changes below of this method could be safely reverted - I guess.

handsontable/src/plugins/autoColumnSize/autoColumnSize.js
```diff
@@ -380,6 +380,7 @@ export class AutoColumnSize extends BasePlugin {
       } else {
         cancelAnimationFrame(timer);
         this.inProgress = false;
+        this.#visualColumnsToRefresh = [];
```
budnix, line 383: Would you be able to find a case that would be the basis for writing a test for this?

handsontable/src/plugins/formulas/formulas.js
```diff
@@ -367,6 +353,16 @@ export class Formulas extends BasePlugin {
     super.destroy();
   }
 
+  /**
+   * Update sheetName and sheetId properties.
+   *
+   * @param {string} [sheetName] The new sheet name.
+   */
+  updateSheetNameAndSheetId(sheetName) {
```
budnix, line 361: Do you think this method can be useful to the user or is it better to leave it private? If private, then I would use the private method `#updateSheetNameAndSheetId`. If not then it's good.

budnix (changes requested):

handsontable/src/dataMap/dataMap.js
```diff
@@ -781,7 +781,7 @@ class DataMap {
     if (this.hot.hasHook('modifyData')) {
       const valueHolder = createObjectPropListener(value);
 
-      this.hot.runHooks('modifyData', physicalRow, this.propToCol(prop), valueHolder, 'get');
+      this.hot.runHooks('modifyData', row, this.propToCol(prop), valueHolder, 'get');
```
budnix, line 784: I am surprised that no test failed. Generally, this is a braking change, a change from physical to visual index.

handsontable/src/dataMap/metaManager/index.js
budnix: After thinking about it, I think that, in this case, it is actually better not to modify the cell meta and pass the option to the hook 🤔

handsontable/src/core.js
sequba: Good point, I'll remove it

handsontable/src/plugins/formulas/formulas.js
sequba: It was intended to be a typical helper method. I marked it as private.

handsontable/src/core.js
sequba: Good point. I changed it

handsontable/src/core.js
sequba: Ok, I changed it

handsontable/src/core.js
sequba: Changed

handsontable/src/dataMap/metaManager/index.js
sequba: Changed

handsontable/src/dataMap/metaManager/index.js
sequba: ok

handsontable/src/dataMap/metaManager/mods/dynamicCellMeta.js
sequba: I did it slightly differently. In `MetaManager.getCellMeta` I put:

```js
    if (!options.skipMetaExtension) {
      this.runLocalHooks('afterGetCellMeta', cellMeta);
    }
```

handsontable/src/dataMap/dataMap.js
sequba: I reverted this change

handsontable/src/plugins/autoColumnSize/autoColumnSize.js
sequba: I added a test case

handsontable/types/core.d.ts
```diff
@@ -59,7 +59,7 @@ export default class Core {
   getCell(row: number, column: number, topmost?: boolean): HTMLTableCellElement | null;
   getCellEditor(cellMeta: CellMeta): BaseEditor;
   getCellEditor(row: number, column: number): BaseEditor;
-  getCellMeta(row: number, column: number): CellProperties;
+  getCellMeta(row: number, column: number, options?: { skipMetaExtension: boolean }): CellProperties;
```
budnix, line 62: Can I ask you to cover that typing with a test... I think this [test file](https://github.com/handsontable/handsontable/blob/develop/handsontable/src/__tests__/core/methods.types.ts#L58) needs to be updated.

handsontable/src/core.js
```diff
@@ -3350,11 +3350,13 @@ export default function Core(rootElement, userSettings, rootInstanceSymbol = fal
...
    * @param {number} row Visual row index.
    * @param {number} column Visual column index.
+   * @param {object} options Execution options for the `getCellMeta` method.
+   * @param {boolean} [options.skipMetaExtension] If `true`, skips extending the cell meta object. Default: `false`.
    * @returns {object} The cell properties object.
    * @fires Hooks#beforeGetCellMeta
    * @fires Hooks#afterGetCellMeta
    */
-  this.getCellMeta = function(row, column) {
+  this.getCellMeta = function(row, column, options = { skipMetaExtension: false }) {
```
budnix, line 3359: Can I ask you to cover the new argument with tests? I was thinking about 2 tests, one called without `skipMetaExtension` defined, which checks that `calls`, `beforeGetCellMeta`, and `afterGetCellMeta` are called and the other with `skipMetaExtension: true`.

handsontable/src/core.js
```diff
@@ -3350,11 +3350,13 @@ export default function Core(rootElement, userSettings, rootInstanceSymbol = fal
    * @function getCellMeta
    * @param {number} row Visual row index.
    * @param {number} column Visual column index.
+   * @param {object} options Execution options for the `getCellMeta` method.
+   * @param {boolean} [options.skipMetaExtension] If `true`, skips extending the cell meta object. Default: `false`.
```
budnix, line 3354: What do you think about more specifying the description? Something like
```suggestion
   * @param {boolean} [options.skipMetaExtension=false] If `true`, skips extending the cell meta object, so the `cells` function won't be called as the `afterGetCellMeta` and `beforeGetCellMeta` hooks.
```

handsontable/src/plugins/formulas/formulas.js
```diff
@@ -842,25 +839,14 @@ export class Formulas extends BasePlugin {
...
-    // using the second method lead to execution of the `cells` method. Using the `getDataAtCell` (which may be useful)
-    // in a callback to the `cells` method leads to triggering the `modifyData` hook. Thus, the `onModifyData` callback
-    // is executed once again and it cause creation of an infinite loop.
-    let cellMeta = this.hot.getCellsMeta().find(singleCellMeta => singleCellMeta.visualRow === visualRow &&
-      singleCellMeta.visualCol === visualColumn);
-
-    if (cellMeta === undefined) {
-      cellMeta = {};
-    }
+    const cellMeta = this.hot.getCellMeta(visualRow, visualColumn, { skipMetaExtension: true }) ?? {};
```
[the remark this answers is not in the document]
sequba: Changed

handsontable/types/core.d.ts
sequba: Done. Didn't know we have typing tests ;)

handsontable/src/plugins/autoColumnSize/autoColumnSize.js
budnix: I have investigated the problem more deeply and it turns out that the problem is between `onInit` and `onAfterFormulasValuesUpdate` listeners. `onInit` is first called and right after that the Formula engine calls `onAfterFormulasValuesUpdate`. Instead of resetting an array here, in the same code where the rows are calculated, I would add a flag like `#isInitialized` and skip the logic in `onAfterFormulasValuesUpdate`.
```js
  #onInit() {
    this.#isInitialized = true;
    this.#cachedColumnHeaders = this.hot.getColHeader();
    this.recalculateAllColumnsWidth();
  }
```
and 
```
  #onAfterFormulasValuesUpdate(changes) {
    if (!this.#isInitialized) {
      return;
    }
    ...
  }
```
This seems to be a cleaner solution, as the `#visualColumnsToRefresh` field is cleared only in one place, the `onBeforeRender` hook.

handsontable/src/plugins/autoColumnSize/autoColumnSize.js
```diff
@@ -420,25 +413,44 @@ export class AutoColumnSize extends BasePlugin {
       }
 
       if (!this.hot._getColWidthFromSettings(physicalColumn)) {
-        const samples = this.samplesGenerator.generateColumnSamples(visualColumn, rowsRange);
-
-        samples.forEach((sample, column) => this.ghostTable.addColumn(column, sample));
+        this.addColumnToGhostTableForWidthCalculation(visualColumn, rowsRange);
```
budnix, line 416: I'd be for shorter version and with `#`.
```suggestion
        this.#fillGhostTableWithSamples(visualColumn, rowsRange);
```

handsontable/src/core.js
sequba: Done

handsontable/src/plugins/autoColumnSize/autoColumnSize.js
sequba: Shorter name is ok, but if I make it private with `#` I cannot spy on it in test: https://github.com/handsontable/handsontable/pull/11474/files?diff=unified&w=1#diff-9d5698a4d7124ac8c73011a89b9064a9b88df353ca291e3eb963e19ce2f3f5ecR1427

budnix: I saw that .. that's why I'd propose a different approach to test that. Checking that locally :)

handsontable/src/plugins/autoColumnSize/autoColumnSize.js
sequba: Good idea. This solution seems reasonable. But anyway, function `calculateAllColumnsWidth` sets width for *all* columns, so don't you think that it should clear the `#visualColumnsToRefresh` array after it completes?

handsontable/src/plugins/autoColumnSize/__tests__/autoColumnSize.spec.js
```diff
@@ -1422,7 +1422,23 @@ describe('AutoColumnSize', () => {
...
+
+      handsontable({
+        data: [[42], ['=A1']],
+        formulas: {
+          engine: HyperFormula
+        },
+      });
+
+      expect(columnWidthCalculationSpy).toHaveBeenCalledTimes(1);
+    });
```
budnix, line 1440: I found that you can change the approach with mocking the method that we want to make private to one with mocking _GhostTable_ (protected) module.
```suggestion
    it('should calculate widths only once during the initialization of Handsontable with formulas plugin enabled', () => {
      Handsontable.hooks.add('beforeInit', function() {
        spyOn(this.getPlugin('autoColumnSize').ghostTable, 'addColumn').and.callThrough();
      })

      handsontable({
        data: [[42], ['=A1']],
        formulas: {
          engine: HyperFormula
        },
      });

      expect(getPlugin('autoColumnSize').ghostTable.addColumn).toHaveBeenCalledTimes(1);
    });
```

handsontable/src/plugins/autoColumnSize/autoColumnSize.js
sequba: Ok, I changed it in `AutoColumnSize` and copied the solution to `AutoRowSize`. Is it alright now?

handsontable/src/core.js
budnix: Many thanks

handsontable/src/plugins/autoColumnSize/autoColumnSize.js
budnix: Perfect 👌 Don’t forget to add a test to `AutoRowSize` plugin… I [guess](https://github.com/handsontable/handsontable/pull/11474#discussion_r1993281497) this one.

handsontable/src/plugins/autoColumnSize/__tests__/autoColumnSize.spec.js
sequba: Changed

handsontable/src/plugins/autoColumnSize/autoColumnSize.js
sequba: Ok, I changed the test according to your idea and renamed the private method

handsontable/src/plugins/autoColumnSize/autoColumnSize.js
sequba: AutoRowSize test added

budnix (approved):
Thanks 👍

PR #11499, opened 2025-03-11
Fix class name assignment in component-based renderer in Vue 2 wrapper

adrianspdev:
### Context
This PR incudes fix for assigning class name in component-based renderer in Vue 2 wrapper

### Types of changes

### Related issue(s):
1. https://github.com/handsontable/dev-handsontable/issues/1204

### Affected project(s):
- [x] `@handsontable/vue`

### Checklist:

### DIFF
--- a/wrappers/vue/src/HotTable.vue
+++ b/wrappers/vue/src/HotTable.vue
@@ -269,7 +269,13 @@
               cachedEntry.lastUsedTD = TD;
             }
           }
+          
+          const className = cellProperties?.className;
 
+          if (className !== void 0 && TD) {
+            TD.classList.add(className)
+          }
+          
           return TD;
         };
       },
--- a/.changelogs/11499.json
+++ b/.changelogs/11499.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Fixed an issue with class name assignment in component-based renderer in Vue 2 wrapper",
+  "type": "fixed",
+  "issueOrPR": 11499,
+  "breaking": false,
+  "framework": "vue"
+}


qunabu (approved):

wrappers/vue/src/HotTable.vue
```diff
@@ -269,7 +269,13 @@
               cachedEntry.lastUsedTD = TD;
             }
           }
+          
+          const className = cellProperties?.className;
```
budnix, line 273: That part of the code does not solve the problem. The main issue sits in the Handsontable. In general, the problem is with custom renderers because we always require those functions to call a base function that sets the state from cell meta to DOM. Like that:
```js
new Handsontable(element, {
  data,
  renderer(hotInstance, TD, row, col, prop, value, cellMeta) {
    Handsontable.renderers.baseRenderer.apply(this, [hotInstance, TD, row, col, prop, value, cellProperties]); // required to work properly
    TD.textContent = value;
  }
});
```
For the Vue2 wrapper and maybe for others, there is a missing base renderer call. To fix that issue completely, I'd not add that `baseRenderer` for each wrapper but instead, add it to the Handsontable so the `baseRenderer` would be called implicitly after the custom cell render.

That would require:
 * Updating [TableView](https://github.com/handsontable/handsontable/blob/develop/handsontable/src/tableView.js#L853). Calling base renderer right after `this.hot.getCellRenderer` call.
```
        getRenderer('base')(
          this.hot,
          TD,
          visualRowIndex,
          visualColumnIndex,
          prop,
          value,
          cellProperties
        );
```
 * Removing unnecessary baseRenderer calls from other renderers. E.g: [checkbox](https://github.com/handsontable/handsontable/blob/develop/handsontable/src/renderers/checkboxRenderer/checkboxRenderer.js#L61);
 * Cover the changes with tests on the Handsontable side and on each wrapper;
 * Update documentation;

budnix (changes requested):

adrianspdev:
new issue created https://github.com/handsontable/dev-handsontable/issues/2325

PR #11532, opened 2025-03-21
Fix the issue of the mobile keyboard closing after clicking the filter search input on Android devices

adrianspdev:
### Context
This PR includes fixes for the issue where the keyboard closes when the user clicks the filter search input on Android devices.

### How has this been tested?
Locally and covered with new test case

### Types of changes

### Related issue(s):
1. https://github.com/handsontable/dev-handsontable/issues/2166

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

### DIFF
--- a/handsontable/src/plugins/filters/filters.js
+++ b/handsontable/src/plugins/filters/filters.js
@@ -24,6 +24,7 @@ import {
   OPERATION_OR_THEN_VARIABLE
 } from './constants';
 import { TrimmingMap } from '../../translations';
+import { isAndroid } from '../../helpers/browser';
 
 export const PLUGIN_KEY = 'filters';
 export const PLUGIN_PRIORITY = 250;
@@ -146,6 +147,10 @@ export class Filters extends BasePlugin {
     super(hotInstance);
     // One listener for the enable/disable functionality
     this.hot.addHook('afterGetColHeader', (...args) => this.#onAfterGetColHeader(...args));
+
+    if (isAndroid()) {
+      this.hot.addHook('beforeRefreshDimensions', () => false);
+    }
   }
 
   /**
--- a/handsontable/src/plugins/filters/filters.js
+++ b/handsontable/src/plugins/filters/filters.js
@@ -24,7 +24,6 @@ import {
   OPERATION_OR_THEN_VARIABLE
 } from './constants';
 import { TrimmingMap } from '../../translations';
-import { isAndroid } from '../../helpers/browser';
 
 export const PLUGIN_KEY = 'filters';
 export const PLUGIN_PRIORITY = 250;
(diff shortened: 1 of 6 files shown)


handsontable/src/plugins/filters/filters.js
```diff
@@ -146,6 +147,10 @@ export class Filters extends BasePlugin {
     super(hotInstance);
     // One listener for the enable/disable functionality
     this.hot.addHook('afterGetColHeader', (...args) => this.#onAfterGetColHeader(...args));
+
+    if (isAndroid()) {
+      this.hot.addHook('beforeRefreshDimensions', () => false);
```
budnix, line 152: Unfortunately, this won't work. Blocking this hook permanently causes the tablet to lose its resizing ability in various scenarios (17 tests failed after removing `isAndroid` from the condition). I think you should try to find a solution to the problem that is more precise/specific to the issue .. now it seems too invasive, affecting other table functionalities 🤔

budnix (changes requested):

handsontable/src/plugins/filters/filters.js
adrianspdev: Fixed in more generic way, now it should working in the same way for all menus

qunabu (dismissed):

handsontable/test/e2e/mobile/filters.spec.js
```diff
@@ -0,0 +1,43 @@
...
+      dropdownMenu: true,
+    });
+
+    const th = hot.view._wt.wtTable.getColumnHeader(0);
+    const button = th.querySelector('.changeType');
+
+    simulateTouch(button);
+
+    const plugin = hot.getPlugin('dropdownMenu');
+    const input = plugin.menu.container.querySelector('.htUIMultipleSelectSearch input');
```
budnix, line 30: If possible, I would use test helpers to make the test content as small as possible. Other than that, I checked the test locally, and it failed on Safari (mobile). Can you check it? I would be for fixing the test so that it passes on all browsers, finally if it is not possible, maybe condition only on Android?
```suggestion
    handsontable({
      data: createSpreadsheetObjectData(3, 3),
      filters: true,
      colHeaders: true,
      dropdownMenu: true,
    });

    const button = getCell(-1, 0).querySelector('.changeType');

    simulateTouch(button);

    const plugin = getPlugin('dropdownMenu');
    const input = plugin.menu.container.querySelector('.htUIMultipleSelectSearch input');
```

budnix (changes requested):

budnix (approved):

PR #11567, opened 2025-04-02
Fix a problem with the table rendering all rows when it's configured to have `0px` height.

jansiegel:
### Context
This PR:
1. Adds a new helper, allowing to recognize the difference between table not inheriting any height and having been deliberately assigned `0px` as height.
2. ~Uses the new helper to prevent the table from assuming its height is infinite, while it should be `0px`.~ 
Solution changed after CR: https://github.com/handsontable/handsontable/pull/11567#issuecomment-2775617470.

### How has this been tested?
Tested manually + added new test cases.

### Types of changes

### Related issue(s):
1. handsontable/dev-handsontable#2211

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

### DIFF
--- a/handsontable/src/3rdparty/walkontable/src/core/_base.js
+++ b/handsontable/src/3rdparty/walkontable/src/core/_base.js
@@ -1,5 +1,6 @@
 import {
   fastInnerText,
+  hasZeroHeight,
 } from '../../../../helpers/dom/element';
 import { randomString } from '../../../../helpers/string';
 import EventManager from '../../../../eventManager';
@@ -123,8 +124,8 @@ export default class CoreAbstract {
   draw(fastDraw = false) {
     this.drawInterrupted = false;
 
-    if (!this.wtTable.isVisible()) {
-      // draw interrupted because TABLE is not visible
+    if (!this.wtTable.isVisible() || hasZeroHeight(this.wtTable.wtRootElement.parentNode)) {
+      // draw interrupted because TABLE is not visible or has the height set to 0
       this.drawInterrupted = true;
     } else {
       this.wtTable.draw(fastDraw);
--- a/handsontable/src/3rdparty/walkontable/src/core/_base.js
+++ b/handsontable/src/3rdparty/walkontable/src/core/_base.js
@@ -124,7 +124,9 @@ export default class CoreAbstract {
   draw(fastDraw = false) {
     this.drawInterrupted = false;
 
-    if (!this.wtTable.isVisible() || hasZeroHeight(this.wtTable.wtRootElement.parentNode)) {
+    if (!this.wtTable.isVisible() ||
+      (this.wtViewport.getTrimmingContainerHeight() === 0 && hasZeroHeight(this.wtTable.wtRootElement.parentNode))
+    ) {
       // draw interrupted because TABLE is not visible or has the height set to 0
       this.drawInterrupted = true;
     } else {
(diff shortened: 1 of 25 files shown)


jansiegel:
After discussing the initial solution with @budnix, we came to the conclusion that preventing the render for tables with `0px` height would be a better idea.

- Initial solution allowed re-rendering the table with `0px` height to have no rows - the new solution blocks the rendering in that cases. Re-rendering doesn't make much sense when the table has no height.
- The new solution makes the table act similarly as when the table is not visible, which makes sense.

Implemented in https://github.com/handsontable/handsontable/pull/11567/commits/a25626d31270d11ae9512363728cfccb67a03a7e, https://github.com/handsontable/handsontable/pull/11567/commits/f7ecca59d20f8f129a4ad82ce86630262e6519df, https://github.com/handsontable/handsontable/pull/11567/commits/9c6cfeb74caf8a6060944baefbb53f2cc28844d2.

handsontable/src/helpers/dom/element.js
```diff
@@ -529,6 +529,33 @@ export function isVisible(element) {
...
+ * @param {HTMLElement} element The element to check.
+ * @returns {boolean} `true` if the element has height set to `0` or `0px` and overflow is set to `hidden`, `false` otherwise.
+ */
+export function hasZeroHeight(element) {
+  // If provided with a window-like object, return false.
+  if (element.document) {
+    return false;
+  }
+
+  const rootDocument = element.ownerDocument;
```
budnix, line 544: Is that line needed? I tested it locally, and after removing all the tests, I passed.
```suggestion
  const rootDocument = element.ownerDocument;
```

handsontable/src/helpers/dom/element.js
```diff
@@ -529,6 +529,33 @@ export function isVisible(element) {
   return true;
 }
 
+/**
+ * Returns true if the element has the height set to `0` or `0px` and overflow to `hidden` (height deliberately set to be 0).
+ *
+ * @param {HTMLElement} element The element to check.
+ * @returns {boolean} `true` if the element has height set to `0` or `0px` and overflow is set to `hidden`, `false` otherwise.
+ */
+export function hasZeroHeight(element) {
```
budnix, line 538: Can I ask to update TS typing for that helper?

handsontable/src/3rdparty/walkontable/src/core/_base.js
```diff
@@ -123,8 +124,10 @@ export default class CoreAbstract {
   draw(fastDraw = false) {
     this.drawInterrupted = false;
 
-    if (!this.wtTable.isVisible()) {
-      // draw interrupted because TABLE is not visible
+    if (!this.wtTable.isVisible() ||
+      (this.wtViewport.getTrimmingContainerHeight() === 0 && hasZeroHeight(this.wtTable.wtRootElement.parentNode))
+    ) {
```
budnix, line 129: I think I would keep that as simple as possible and remove the `getTrimmingContainerHeight` method, as it returns relatively unstable values. For different table configurations, it returns either a number or `undefined` (when there is no top overlay). 
```suggestion
    if (!this.wtTable.isVisible() || hasZeroHeight(this.wtTable.wtRootElement.parentNode)) {
```

budnix (changes requested):

handsontable/src/helpers/dom/element.js
jansiegel: With the new usage, it shouldn't 👍

budnix (approved):

PR #11586, opened 2025-04-15
[docs] Generate UMD builds of Angular-related packages before building the documentation.

jansiegel:
### Context
As the newer versions of Angular (and our Angular wrapper) don't generate the UMD builds, we needed to prepare those builds to be used within our docs ourselves.

This PR:
- Adds a script to generate the UMD builds of the Angular-related packages after installing dependencies and before building the docs (the latter is needed when working on a symlinked Angular wrapper).
- Modifies the docs to use the prebuilt UMD modules.
- Modifies the release script to make use of the changes (changing the version of `@handsontable/angular` in the docs' `package.json` file).
- Modifies the existing Angular demos to work with the latest Angular version.

[skip changelog]

### How has this been tested?
Tested locally.

### Related issue(s):
1. handsontable/dev-handsontable#2398


budnix (changes requested):
LGTM, however, I'd fix the jsfiddle dependencies before merging. It seems to be duplicated, and the full URL are missing (https://jsfiddle.net/budnix/8rud5ztp/).

qunabu (dismissed):

budnix (approved):

budnix (approved):

PR #11599, opened 2025-04-29
Correct the release script (assigning a wrong version number for `@handsontable/angular` in the docs devDependencies).

jansiegel:
### Context
This PR:
- Corrects the logic in the release script that assigned a wrong version number to `@handsontable/angular` in the docs' `devDepenendencies`.
- Makes the Angular UMD builder install its own dependencies and not rely on those installed at the root level.

This change should be cherry-picked to `prod-docs/15.3`.

[skip changelog]

### How has this been tested?
Tested manually.


budnix (approved):

qunabu (dismissed):

jansiegel:
Note for the reviewers (@budnix @qunabu): I added one additional commit to this PR.

It makes the Angular UMD builder install its own dependencies and not rely on those installed at the root level.

budnix (approved):

PR #11602, opened 2025-04-30
Second click deselects cell feature

budnix:
### Context

The PR adds a new feature that allows users to deselect already selected cells. The feature is limited to only single cells (including single merged cells). Single cells defined within a bigger selection range can also be deselected. The cell will be deselected when the <kbd>Command/Ctrl</kbd> is pressed, and at the same time, the already selected cell is clicked.

#### Example
Kapture 2025-05-16 at 09 40 57

##### Internal notes

In the context of selection, it's sometimes essential to perform calculations based on renderable indices. To support this, this PR introduces a _CellRange_ mapper that enables translation from visual to renderable coordinates.

The implementation leverages existing workspace structures and applies a lightweight dependency injection pattern to preserve the separation of concerns, keeping the transformation logic decoupled from the core _CellRange_ data structure.

Example of internal usage:
```js
import { resolveWithInstance } from './utils/staticRegister';

const cellRange = new CellRange(...);
const renderableRange = resolveWithInstance(this.hot, 'cellRangeMapper')
  .toRenderable(cellRange);
```

### How has this been tested?

I tested the changes locally, and I covered the feature with new unit and E2E tests.

### Types of changes

### Related issue(s):
1. fixes https://github.com/handsontable/dev-handsontable/issues/2387

### Affected project(s):
- [x] `handsontable`

### Checklist:


mohkshu:
Hello @budnix,
Is it planned to support deselecting individual cells within a multi-selection, or deselecting an already selected area when dragging with the Ctrl key pressed in a future Handsontable release?

Thank you

AMBudnik:
Hi @mohkshu, thanks for bringing this up!

At the moment, support for deselecting individual cells within a multi-selection (or deselecting an already selected area using Ctrl + drag) is not directly planned for the Q4/26 roadmap.

That said, we’re actively collecting feedback around selection workflows, and I’d really like to understand your use case a bit better so I can pass it on to the person responsible for planning future functionality.

Could you share:

* what problem this would help you solve in your workflow?
* how often you run into this limitation?

Any extra context will help us evaluate this more accurately. Thanks again for taking the time to share your feedback.

mohkshu:
https://github.com/user-attachments/assets/0a4ccaa8-5378-4069-8646-02fafdc09e2f

Hello @budnix,

Thank you very much for the detailed response and for taking the time to explain the current roadmap — I really appreciate it.

To give a bit more context on our use case: we have a large number of end customers who use our application daily and who are very accustomed to working with Excel. Our application is used by companies to fill in and manage reports related to financial regulations and sustainability requirements, where users often work with large datasets and complex tables.

In Excel, the ability to adjust a multi-selection by deselecting parts of it while dragging (for example with Ctrl + drag) is something they rely on quite heavily in these kinds of workflows. Users frequently need to refine selections quickly and precisely when preparing or reviewing regulatory and sustainability data.

Because of this familiarity, users naturally expect the same interaction patterns when working with grid-based data in our application. When this behavior isn’t available, it can feel limiting or unintuitive to them, even though the rest of the experience is quite solid. We do run into this limitation fairly often, especially in scenarios where users need to fine-tune large or non-contiguous selections.

From our perspective, having selection behavior that closely matches Excel would significantly improve the user experience and reduce friction for our customers, as it aligns with the muscle memory they already have from tools they use every day.

Thanks again for being open to feedback and for considering this input. We really appreciate the work you’re doing on Handsontable and the opportunity to share real-world usage insights.

Best regards,

AMBudnik:
Hi @mohkshu 

Thank you so much for taking the time to share such a detailed explanation of your use case and especially for including the video. This level of context is incredibly helpful for us and really clarifies how critical this interaction pattern is in your day-to-day workflows. Your examples around regulatory and sustainability reporting make a lot of sense, and they highlight exactly where the current behavior can introduce unnecessary friction. 💯 

We’re taking this feedback back internally and will use it as concrete input in ongoing product discussions around selection behavior. Having real-world scenarios like yours makes those conversations much more grounded and actionable.

As we explore potential solutions, we’d also love to keep you close in the loop. If we move toward early concepts or alpha testing in this area, would you be open to testing it and sharing feedback? Your experience would be extremely valuable in helping us validate that we’re heading in the right direction. :)

mohkshu:
Hello @AMBudnik,

First of all, Happy New Year! 🎉

I’d be happy to provide feedback on the early concepts for this feature.

Please note that this feature is urgent, so we won’t be able to wait for a full implementation. I would greatly appreciate any hints or guidance on how we could achieve this without modifying the core logic.

Looking forward to your response :)

adrianszymanski89:
Hi @mohkshu 

Thank you for your feedback. We were able to prepare a POC of such functionality using our API. You can check the example here: https://stackblitz.com/edit/zeai1rsz-gba1sayl?file=index.js 

An instruction is provided inside the example. We would really appreciate your feedback on this.

mohkshu:
Hello @adrianszymanski89,

Thanks for the POC — I tested it, and it works really well! The single-cell deselection is smooth and intuitive, great job on that. 🙌

A couple of thoughts:

It could be nice to merge selection ranges across rows when the zones are the same size, to make selections more compact.

I tried adding the ability to deselect a rectangular zone, but I’m not sure how you handle that internally. Do you use a drag event for selection with dragging, or not?

Overall, this is a really solid starting point. Our clients would be really happy with this feature :)

Thank again for the time and for sharing it!

adrianszymanski89:
Hi @mohkshu 

Thank you for sharing your feedback. That will be really helpful while making adjustments to this functionality. I also prepared another example, where deselecting cell ranges is also possible: https://stackblitz.com/edit/zeai1rsz-yrcjj8yy?file=index.js

mohkshu:
Hello @adrianszymanski89, I tested the cell range deselection feature. It works, but the behavior is slightly different from what I expected. Deselecting should only happen on mouse up, and the deselection range should be visible while interacting.

I created a working POC that fits perfectly our needs based on your code to better illustrate the behavior we’re aiming for for unselecting cell ranges: 
[https://stackblitz.com/edit/zeai1rsz-cc6woi3g?file=index.js](url)

Thank you !

adrianszymanski89:
Hi @mohkshu 

Thank you, we really appreciate your input. We will take that into consideration.

PR #11613, opened 2025-05-09
Fix a problem with the dropdown editor having a horizontal scrollbar on Windows with fractional scaling applied.

jansiegel:
### Context
This PR should fix the problem with the dropdown editor having a horizontal scrollbar on Windows with fractional scaling applied.

What I found was that the scrollbar was wider by `2px` when the scaling was set to a fractional value (like `1.25` or `125%`) on Windows. The `2px` seem to be there because of the anti-aliasing Windows uses to render the UI elements in this scenario. Unfortunately, I wasn't able to confirm that's exactly what happens with any external sources.

**Scaling set to `100%`:**
image

**Scaling set to `125%`:**
image

The problematic thing is that Windows sets the scaling to `125%` **by default** on some machines, which makes this issue a bigger problem.
I ran our E2E tests on a Windows laptop with the default scaling of `125%` and it ended with 500+ failing tests, so I assume there are more places where it can cause visual glitches. I think we should put some effort into investigating those potential issues. (CC: @qunabu)

----

This PR tries to work around that problem by compensating the scrollbar width by `2px` when used on Windows with a fractional `devicePixelRatio`.

Additionally, the changes include a fix for the unit tests silently throwing errors (what's currently happening on `develop`).

### How has this been tested?
- Tested locally on a Parallels machine with scaling and on a Windows laptop.
- Added a test case for the new logic.

### Types of changes

### Related issue(s):
1. handsontable/dev-handsontable#2427

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

### DIFF
--- a/handsontable/src/helpers/dom/element.js
+++ b/handsontable/src/helpers/dom/element.js
@@ -976,7 +976,7 @@ let cachedScrollbarWidth;
  * @returns {number} The compensation for the scrollbar width, when the device pixel ratio is not an integer.
  */
 // eslint-disable-next-line no-restricted-globals
-function getScrollbarFractionalScalingCompensation(rootDocument = document) {
+export function getScrollbarFractionalScalingCompensation(rootDocument = document) {
   const rootWindow = rootDocument.defaultView;
   const isWindows = rootWindow.navigator.userAgent.toLowerCase().includes('windows');
 
--- a/handsontable/src/helpers/dom/element.js
+++ b/handsontable/src/helpers/dom/element.js
@@ -1,5 +1,6 @@
 import { sanitize } from '../string';
 import { A11Y_HIDDEN } from '../a11y';
+import { isWindowsOS } from '../browser';
 
 /**
  * Get the parent of the specified node in the DOM tree.
@@ -977,18 +978,13 @@ let cachedScrollbarWidth;
  */
 // eslint-disable-next-line no-restricted-globals
 export function getScrollbarFractionalScalingCompensation(rootDocument = document) {
-  const rootWindow = rootDocument.defaultView;
-  const isWindows = rootWindow.navigator.userAgent.toLowerCase().includes('windows');
-
-  if (!isWindows) {
+  if (!isWindowsOS()) {
     return 0;
   }
 
-  const devicePixelRatio = rootWindow.devicePixelRatio || 1;
-
   // On Windows, fractional scaling makes the scrollbar wider to compensate for the anti-aliasing.
   // This is a workaround to calculate the correct scrollbar width.
-  return Number.isInteger(devicePixelRatio) ? 0 : 2;
+  return Number.isInteger((rootDocument.defaultView.devicePixelRatio || 1)) ? 0 : 2;
 }
 
 /**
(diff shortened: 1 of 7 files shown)


handsontable/src/helpers/dom/element.js
```diff
@@ -969,6 +969,28 @@ export function setCaretPosition(element, pos, endPos) {
...
+ * @returns {number} The compensation for the scrollbar width, when the device pixel ratio is not an integer.
+ */
+// eslint-disable-next-line no-restricted-globals
+export function getScrollbarFractionalScalingCompensation(rootDocument = document) {
+  const rootWindow = rootDocument.defaultView;
+  const isWindows = rootWindow.navigator.userAgent.toLowerCase().includes('windows');
+
+  if (!isWindows) {
+    return 0;
+  }
```
budnix, line 985:
```suggestion

  if (!isWindowsOS()) {
    return 0;
  }
```
There is already a helper available. I highly encourage you to use it https://github.com/handsontable/handsontable/blob/feature/dev-issue-2427/handsontable/src/helpers/browser.js#L134.

handsontable/src/helpers/dom/element.js
```diff
@@ -969,6 +969,28 @@ export function setCaretPosition(element, pos, endPos) {
...
+
+  const devicePixelRatio = rootWindow.devicePixelRatio || 1;
+
+  // On Windows, fractional scaling makes the scrollbar wider to compensate for the anti-aliasing.
+  // This is a workaround to calculate the correct scrollbar width.
+  return Number.isInteger(devicePixelRatio) ? 0 : 2;
```
budnix, line 991: Just guessing .. `Math.ceil` doesn't solve the case in a more general way?

budnix (changes requested):

handsontable/src/helpers/dom/element.js
jansiegel: I was going to use `Math.ceil`, but it would up the value by `1`, and the difference that seems to be happening is `2px` (I'm guessing anti-aliasing on both sides of the scrollbar?), so it wouldn't be sufficient.

budnix (dismissed):

handsontable/src/helpers/dom/element.js
```diff
@@ -969,6 +970,23 @@ export function setCaretPosition(element, pos, endPos) {
...
+ */
+// eslint-disable-next-line no-restricted-globals
+export function getScrollbarFractionalScalingCompensation(rootDocument = document) {
+  if (!isWindowsOS()) {
+    return 0;
+  }
+
+  // On Windows, fractional scaling makes the scrollbar wider to compensate for the anti-aliasing.
+  // This is a workaround to calculate the correct scrollbar width.
+  return Number.isInteger((rootDocument.defaultView.devicePixelRatio || 1)) ? 0 : 2;
```
budnix, line 987:
```suggestion
  return Number.isInteger(rootDocument.defaultView.devicePixelRatio || 1) ? 0 : 2;
```

budnix (approved):

budnix (approved):

qunabu (approved):

PR #11693, opened 2025-06-25
Introduce pnpm

jansiegel:
### Context
Because of constant problems with dependency management with npm workspaces, we decided to give pnpm a try.

---

#### Note for the reviewers:
To reinstall the dependencies using pnpm:

1. [Install pnpm](https://pnpm.io/installation)
2. Run `npm run clean:node_modules` 
3. Run `git checkout pnpm-lock.yaml` to retrieve the pnpm lock file removed by the `clean:node_modules` script
4. Run `pnpm install`, which [should install the dependencies in the version defined by the pnpm lock file](https://pnpm.io/settings#preferfrozenlockfile)
5. Use the repo as usual.

---

#### Key changes
- **From now on, dependency installation for handsontable and the wrappers needs to be done using `pnpm install`**.
  I replaced the currenty `package-lock.json` with `pnpm-lock.yaml`. 
  Docs and examples are still managed with `npm`. (might change in the future)
- PNPM is configured to install `peerDependencies`, to be compliant with the way `npm` currently installs dependencies.
- I added a multitude of `devDependencies` to our packages. It turns out, some of them were using the other packages' `devDependencies`, which stopped working after implementing pnpm with its more strict dependency resolving logic.
- I removed the `swap-package-links` helper script. `pnpm` can utilize the `publishConfig.directory` field of `package.json` (which I added to our angular and handsontable packages) and symlink directly to the directory listed there. This way, the `swap-packages` script should no longer be needed.
- The workspace definition is currently doubled - in the root`package.json` and `pnpm-workspace.yaml`. We need that to keep the compatibility with the native `npm` commands, so we can still build and test the packages using `npm run build`, `npm run test`, etc. If we ever decide to lose that compatibility, we could remove the `package.json` entry.
- Updated the Github Actions workflows, replacing `npm i` with `pnpm install`
- Updated the Github Actions workflows to use a newer version of `actions/setup-node` to get rid of the cache warnings that have been popping up on the CI

#### Problems
- There were issues with the React examples.
  Namely, the examples used double instances of `react` and `react-dom` - one from the examples' directory, one pulled from the wrappers' directory, through the symlinked wrapper. I resolved this by aliasing those two packages to the local `node_modules` and turning off the plugin that checks if a project uses code from outside of `src` - in our case, we're doing it with intent, so I guess that's fine.
  It's worth considering pulling the examples back into the main repository or resolving this problem differently in the future.
- With pnpm, the local packages aren't available in the root `node_modules`, they're only imported when needed. That was a problem with our example packages, which used the symlinks in the root's `node_modules`. To get around this, I added the wrapper and handsontable packages as `devDependencies` in the examples private `package.json` file. This way, they are available for all the examples through `examples/node_modules` (managed by pnpm). I updated the helper scripts to use the new paths.

#### Potential future problems
- Using `npm` and `pnpm` in the same repository might be problematic (as with the examples). As `pnpm`'s `node_modules` structure vastly differs from `npm`'s flat structure, symlinking can sometimes be problematic.

[skip changelog]

### How has this been tested?
Tested locally and with the updated CI workflows.

### Checklist:


.github/workflows/audit.yml
```diff
@@ -8,17 +8,26 @@ concurrency:
...
   build:
-    name: NPM audit
+    name: PNPM audit
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # https://github.com/actions/checkout/releases/tag/v4.1.1
 
+      - uses: pnpm/action-setup@a7487c7e89a18df4991f7f222e4898a00d66ddda # v4.1.0
+        name: Install pnpm
+        with:
```
budnix, line 18: Is it possible to read that value from file? As we did with npm and `.npmrc` file

qunabu (dismissed):

PR #11754, opened 2025-07-22
Dialog Plugin: Core Modal System

adrianspdev:
### Context
This PR introduces the Dialog Plugin. Currently, Handsontable does not have a unified system for managing dialogs or modals used for overlays, notifications, or interactive components. This PR introduces a centralized modal infrastructure to serve as a foundation for all dialog-related functionality across plugins.

By consolidating modal behavior into a single system, we avoid the need for each plugin to implement its own solution — reducing code duplication, ensuring a consistent user experience, and simplifying long-term maintenance.

### New Configuration Options

The plugin provides several configuration options to customize the dialog behavior and appearance

 - `content`: The string or HTMLElement content to display in the dialog (default: '')
 - `customClassName`: Custom class name to apply to the dialog (default: '')
 - `background`: Dialog background variant 'solid' | 'semi-transparent' (default: 'solid')
 - `contentBackground`: Whether to show content background
 - `contentDirections`: Content layout direction 'row' | 'row-reverse' | 'column' | 'column-reverse' (default: 'row')
 - `animation`: Whether to enable animations (default: true)
 - `closable`: Whether the dialog can be closed (default: false)

The configuration example:

```
new Handsontable(element, {
  data,
  dialog: {
    content: 'Dialog content',
    customClassName: 'custom-dialog',
    background: 'semi-transparent',
    contentBackground: false,
    contentDirections: 'column',
    animation: false,
    closable: true,
  }
});
```

### New Methods

In the dialog plugin, we have the following methods:

- `isVisible()`: Check if the dialog is visible.
- `show(options)`: Show the dialog with the specified options.
- `hide()`: Hide the dialog.
- `update(options)`: Update the dialog with the specified options.

### New Hooks

In the dialog plugin, we have seven new hooks:

- `beforeDialogShow`: Hook called before the dialog is shown.
- `afterDialogShow`: Hook called after the dialog is shown.
- `beforeDialogHide`: Hook called before the dialog is hidden.
- `afterDialogHide`: Hook called after the dialog is hidden.
- `dialogFocusPreviousElement`: Hook called when the dialog focus is moved to the previous element.
- `dialogFocusNextElement`: Hook called when the dialog focus is moved to the next element.
- `afterDialogFocus`: Hook called after the dialog focus is set.

### Additional changes

Within this PR, we have the following additional changes:

- The context manager has been extended to support new dialog and pagination contexts, enabling better control over keyboard shortcuts and focus management in these plugins.

- The focus catcher mechanism has been enhanced to provide more robust focus management:
  - Improved handling of focus transitions between the table and dialog
  - Better keyboard navigation within dialog elements
  - More reliable focus restoration when closing dialogs
  - Prevention of focus loss during dialog interactions

- The dialog plugin now integrates seamlessly with the focus catcher:
  - Maintains focus within the dialog when open
  - Handles tab navigation between dialog elements
  - Restores focus appropriately when closing
  - Prevents focus from escaping the dialog unintentionally

- Focus management has been unified across plugins:
  - Consistent focus behavior between dialog and pagination
  - Shared focus catcher infrastructure
  - Common keyboard navigation patterns

- The focus catcher mechanism has been enhanced to provide more robust focus management across plugins:
  - Handles keyboard navigation between elements consistently 
  - Restores focus appropriately when closing dialogs
  - Provides unified focus behavior between dialog and pagination plugins
  - Shares common focus catcher infrastructure and navigation patterns

### How has this been tested?
Locally, and e2e test cases

### Types of changes

### Related issue(s):
1. https://github.com/handsontable/dev-handsontable/issues/2707

### Affected project(s):
- [x] `handsontable`

### Checklist:


handsontable/src/plugins/dialog/dialog.js
```diff
@@ -0,0 +1,471 @@
...
+
+  /**
+   * Validators for dialog configuration options.
+   *
+   * @private
+   * @type {object}
+   */
+  static get DIALOG_CONFIG_VALIDATORS() {
+    return {
+      content: val => typeof val === 'string' || (typeof HTMLElement !== 'undefined' && val instanceof HTMLElement),
```
budnix, line 107: I might be nitpicking here, but I prefer more descriptive variable names over short ones like `val`. Would you mind renaming it to `value`?

handsontable/src/plugins/dialog/dialog.js
```diff
@@ -0,0 +1,471 @@
...
+    };
+  }
+
+  /**
+   * Instance of the dialog element.
+   *
+   * @private
+   * @type {HTMLElement}
+   */
+  dialogElement = null;
```
budnix, line 123: If possible, use private fields

```suggestion
  /**
   * Instance of the dialog element.
   *
   * @type {HTMLElement}
   */
  #dialogElement = null;
```
also applies to the fields below

handsontable/src/plugins/dialog/dialog.js
```diff
@@ -0,0 +1,471 @@
...
+   *
+   * @private
+   */
+  #updateDialogClassName() {
+    const customClass = this.currentConfig.customClassName ? ` ${this.currentConfig.customClassName}` : '';
+    const backgroundClass = ` ${DIALOG_CLASS_NAME}--background-${this.currentConfig.background}`;
+    const animationClass = this.currentConfig.animation ? ` ${DIALOG_CLASS_NAME}--animation` : '';
+    const showClass = this.isVisible() ? ` ${DIALOG_CLASS_NAME}--show` : '';
+
+    this.dialogElement.className = `${DIALOG_CLASS_NAME}${customClass}${backgroundClass}${animationClass}${showClass}`;
```
budnix, line 367: Wouldn't it be better to use our `addClass` helper? Then it could look like 
```js
addClass(this.dialogElement, [DIALOG_CLASS_NAME, customClass, ...]);
```

handsontable/src/plugins/dialog/dialog.js
```diff
@@ -0,0 +1,471 @@
...
+    // Render new content
+    if (typeof content === 'string') {
+      fastInnerHTML(this.contentElement, content);
+    } else if (content instanceof HTMLElement) {
+      this.contentElement.appendChild(content);
+    }
+
+    // Update content background styling
+    if (contentBackground) {
+      addClass(this.contentElement, `${DIALOG_CLASS_NAME}__content--background`);
```
budnix, line 392: If you used the `addClass` here, I'd use it everywhere instead of writing to the `.className` property.

handsontable/src/plugins/dialog/dialog.js
```diff
@@ -0,0 +1,471 @@
...
+    this.#updateDialogClassName();
+
+    // Set ARIA attributes
+    setAttribute(this.dialogElement, [
+      A11Y_DIALOG(),
+      A11Y_MODAL(),
+    ]);
+
+    // Create content wrapper
+    const contentWrapperElement = this.hot.rootDocument.createElement('div');
```
budnix, line 310: Instead of juggling with elements manually, maybe it would be better to use the recently added `html` [helper](https://github.com/handsontable/handsontable/blob/develop/handsontable/src/helpers/templateLiteralTag.js#L50)?

handsontable/src/plugins/dropdownMenu/dropdownMenu.js
```diff
@@ -350,7 +351,9 @@ export class DropdownMenu extends BasePlugin {
    * @fires Hooks#afterDropdownMenuShow
    */
   open(position, offset = { above: 0, below: 0, left: 0, right: 0 }) {
-    if (this.menu?.isOpened()) {
+    const dialogPlugin = this.hot.getPlugin('dialog');
```
budnix, line 354: The same as for the context menu case. The dropdown menu is not aware of the dialog plugin.

handsontable/src/shortcuts/__tests__/shortcutManager.spec.js
```diff
@@ -704,4 +704,24 @@ describe('shortcutManager', () => {
...
+    it('should create a context with the `table` scope by default', async() => {
+      handsontable();
+
+      const shortcutManager = getShortcutManager();
+      const gridContext = shortcutManager.addContext('test');
+
+      expect(gridContext.scope).toBe('table');
+    });
+
+    it('should create a context with the `global` scope when the `scope` option is set to `global`', async() => {
```
budnix, line 718: Here are tests that check that the props are assigned, but there is a missing test that checks how that scope affects how the keyboard shortcuts are handled (handled or not when Handsontable is listening).

handsontable/types/plugins/dialog/dialog.d.ts
```diff
@@ -0,0 +1,27 @@
+import Core from '../../core';
+import { BasePlugin } from '../base';
+
+export interface DialogConfig {
+  content?: string | HTMLElement;
+  customClassName?: string;
```
budnix, line 6: I would add support for array of strings also

handsontable/types/plugins/dialog/dialog.d.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export type Settings = boolean | DialogConfig;
+
+export class Dialog extends BasePlugin {
+  static PLUGIN_KEY: string;
+  static PLUGIN_PRIORITY: number;
+  static DEFAULT_CONFIG: DialogConfig;
+
+  constructor(hotInstance: Core);
+  isEnabled(): boolean;
```
budnix, line 22: The methods is already defined in the BaseClass
```suggestion
  constructor(hotInstance: Core);
```

handsontable/src/plugins/dialog/dialog.js
```diff
@@ -0,0 +1,471 @@
...
+import { A11Y_DIALOG, A11Y_MODAL } from '../../helpers/a11y';
+
+export const PLUGIN_KEY = 'dialog';
+export const PLUGIN_PRIORITY = 340;
+const DIALOG_CLASS_NAME = 'ht-dialog';
+const SHORTCUTS_GROUP = PLUGIN_KEY;
+const SHORTCUTS_CONTEXT_NAME = `plugin:${PLUGIN_KEY}`;
+
+// Register dialog hooks
+Hooks.getSingleton().register('beforeDialogShow');
```
budnix, line 14: There is a missing doc for those four hooks. That file should be updated https://github.com/handsontable/handsontable/blob/feature/dev-issue-2707/handsontable/src/core/hooks/constants.js.

budnix (changes requested):

qunabu (approved):

handsontable/src/plugins/dialog/dialog.js
```diff
@@ -0,0 +1,471 @@
...
+  static get PLUGIN_PRIORITY() {
+    return PLUGIN_PRIORITY;
+  }
+
+  /**
+   * Dialog default configuration.
+   *
+   * @returns {object}
+   */
+  static get DEFAULT_CONFIG() {
```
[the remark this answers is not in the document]
adrianspdev: will be moved

handsontable/src/plugins/dialog/dialog.js
adrianspdev: moved to new UI file

handsontable/src/plugins/dialog/dialog.js
adrianspdev: moved to new UI file

handsontable/src/plugins/dialog/dialog.js
adrianspdev: moved to new UI file

handsontable/src/dataMap/metaManager/metaSchema.js
```diff
@@ -1840,6 +1840,46 @@ export default () => {
...
+     * @description
+     * The `dialog` option configures the [`Dialog`](@/api/dialog.md) plugin.
+     *
+     * You can set the `dialog` option to one of the following:
+     *
+     * | Setting   | Description                                                                 |
+     * | --------- | --------------------------------------------------------------------------- |
+     * | `false`   | Disable the [`Dialog`](@/api/dialog.md) plugin                           |
+     * | `true`    | Enable the [`Dialog`](@/api/dialog.md) plugin with default options     |
+     * | An object | Enable the [`Dialog`](@/api/dialog.md) plugin with custom configuration |
```
budnix, line 1853: It’s worth to extend the docs for "additional options" section, for example https://github.com/handsontable/handsontable/blob/develop/handsontable/src/dataMap/metaManager/metaSchema.js#L3596

handsontable/src/dataMap/metaManager/metaSchema.js
```diff
@@ -1840,6 +1840,46 @@ export default () => {
...
+     * - [Plugins: `Dialog`](@/api/dialog.md)
+     *
+     * @since 16.1.0
+     * @memberof Options#
+     * @type {boolean|object}
+     * @default false
+     * @category Dialog
+     *
+     * @example
+     * ```js
```
budnix, line 1865: Can I ask you to add examples also for React and Angular?

handsontable/src/editors/autocompleteEditor/autocompleteEditor.js
```diff
@@ -112,6 +112,8 @@ export class AutocompleteEditor extends HandsontableEditor {
         A11Y_AUTOCOMPLETE(),
       ]);
     }
+
+    this.addHook('beforeDialogShow', () => this.close());
```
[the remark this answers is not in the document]
budnix: The hooks was nit moved to the BaseEditor?

handsontable/src/plugins/dialog/__tests__/methods/hide.spec.js
```diff
@@ -0,0 +1,87 @@
...
+
+    const dialogPlugin = hot.getPlugin('dialog');
+
+    dialogPlugin.show();
+    expect(dialogPlugin.isVisible()).toBe(true);
+
+    dialogPlugin.show();
+
+    expect(dialogPlugin.isVisible()).toBe(true);
+    dialogPlugin.hide();
```
budnix, line 31: There is still here and there problem with formatting. May I recommend following a code formatting style where the expectation (`expect`) is separated from the test logic by a newline? Can you review all test files?

handsontable/src/plugins/dialog/dialog.js
```diff
@@ -0,0 +1,364 @@
...
+      contentBackground: this.getSetting('contentBackground'),
+      contentDirections: this.getSetting('contentDirections'),
+      animation: this.getSetting('animation'),
+    });
+  }
+
+  /**
+   * Destroy dialog and reset plugin state.
+   *
+   * @private
```
budnix, line 348: There is no need to add a `@private` tag for private fields

budnix (changes requested):

budnix (changes requested):

handsontable/src/editors/autocompleteEditor/autocompleteEditor.js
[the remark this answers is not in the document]
budnix: As you added the hook to the BaseClass that listener should be removed from here and other child editor classes.

budnix:
Once I showed the dialog, I cannot hide it by calling the `hide` method

https://github.com/user-attachments/assets/97dae009-e3d4-4f06-9a75-bd9365562de2

Also, there is another bug, the dialog after opening has a different size than the table. When the table size is defined with 500px, the dialog should also have 500px. My settings:
```js
  width: 500,
  height: 200,
  dialog: {
    content: 'Hello',
  },
 ```

**Comment by budnix** [comment]
When the dialog is enabled and shown, it's not possible to focus the table/dialog. The escape keyboard shortcut does not work.

https://github.com/user-attachments/assets/d64f6eb6-5f06-452b-b204-3711e3a0754f

File: handsontable/src/dataMap/metaManager/metaSchema.js [hunk 17]
```diff
@@ -1840,6 +1840,102 @@ export default () => {
...
+     * | `content`                | A string or HTML element (default: `''`)           | The content of the dialog                                                                                                                                        |
+     * | `customClassName`        | A string (default: `''`)                           | The custom class name of the dialog                                                                                                                              |
+     * | `background`             | A string (default: `'semi-transparent'`)           | The background of the dialog                                                                                                                                     |
+     * | `contentBackground`      | Boolean (default: `true`)                          | Whether to show the content background                                                                                                                           |
```
budnix, line 1861: In the code, the default value is defined as `false`. Is it right?
```suggestion
     * | `contentBackground`      | Boolean (default: `false`)                          | Whether to show the content background                                                                                                                           |
```

budnix:
If possible, I would add some visual tests to protect us from visual regression bugs that, for this project, seem very important.

handsontable/src/plugins/base/base.js
```diff
@@ -194,32 +211,98 @@ export class BasePlugin {
...
+      Object.keys(settingsValidators).forEach((key) => {
+        if (!(key in newSettings)) {
+          return;
+        }
+
+        const validator = settingsValidators[key];
+        const isValid = validator ? validator(newSettings[key]) : true;
+
+        if (isValid === false) {
+          warn(`${this.pluginName}: "${key}" option is not valid and it will be ignored.`);
```
budnix, line 290:
```suggestion
          warn(`${this.pluginName} Plugin: "${key}" option is not valid and it will be ignored.`);
```

budnix:
It would be nice to have support for passing DocumentFragment to the `content` option. Something like this:
```html
<!-- Template for dialog content -->
<template id="dialog-template">
  <div style="padding: 10px; border: 1px solid #000; border-radius: 5px;">
    <b><i>Hello</i></b>
  </div>
</template>
```
then
```js
const dialogTemplate = document.getElementById('dialog-template');
const dialogContent = dialogTemplate.content.cloneNode(true);

const createHot = (id, options) => new Handsontable(document.querySelector(id), {
  data,
  dialog: {
    content: dialogContent,
  },
```

handsontable/src/plugins/contextMenu/contextMenu.js
```diff
@@ -245,10 +246,6 @@ export class ContextMenu extends BasePlugin {
    * @fires Hooks#afterContextMenuShow
    */
   open(position, offset = { above: 0, below: 0, left: 0, right: 0 }) {
-    if (this.menu?.isOpened()) {
```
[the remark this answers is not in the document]
adrianspdev: I'll revert this same for dropdown menu

handsontable/src/plugins/dialog/dialog.js
```diff
@@ -0,0 +1,364 @@
...
+
+    this.hot.runHooks('afterDialogHide');
+  }
+
+  /**
+   * Update the dialog configuration.
+   *
+   * @param {object} settings - The configuration to update the dialog with.
+   */
+  update(settings) {
```
budnix, line 329: It's worth extending the docs for the `update` and `show` methods' options. Now it looks like this

Image

handsontable/src/plugins/dialog/dialog.js
```diff
@@ -0,0 +1,364 @@
...
+      contentDirections: value => ['row', 'row-reverse', 'column', 'column-reverse'].includes(value),
+      animation: value => typeof value === 'boolean',
+      closable: value => typeof value === 'boolean',
+    };
+  }
+
+  /**
+   * UI instance of the dialog plugin.
+   *
+   * @private
```
budnix, line 186:
```suggestion
```

handsontable/src/plugins/dialog/dialog.js
```diff
@@ -0,0 +1,364 @@
...
+  }
+
+  /**
+   * Register shortcuts responsible for toggling a merge.
+   *
+   * @private
+   */
+  registerShortcuts() {
+    const manager = this.hot.getShortcutManager();
+    const pluginContext = manager.getContext(SHORTCUTS_CONTEXT_NAME) ||
```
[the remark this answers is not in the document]
budnix: Can I ask you to cover that change with a test? I mean, add a test where you enable, disable, and again enable the plugin via `updateSettings`

handsontable/src/core/focusCatcher/index.js
```diff
@@ -122,9 +127,15 @@ export function installFocusCatcher(hot) {
                 ? getMostTopStartPosition(hot) : getMostBottomEndPosition(hot);
             }
 
+            const result = hot.runHooks('modifyUnfocusOnTabNavigation', event.shiftKey ? 'to_above' : 'to_below');
```
budnix, line 130: I'm not sure about the hook name. The hook is triggered only when the internal element (table) loses its focus (and passes to the Pagination, Dialog, or other plugin in the feature). WDYT about changing the name to `focusTableTransfer`, or `tableFocusExit`. I'd not use the `modify` phrase here, as there is nothing to modify, like in the `modifyFocusOnTabNavigation` hook, where you can modify which cell can be selected after table focus.

@jansiegel WDYT?

handsontable/src/core/hooks/constants.js
```diff
@@ -1886,6 +1886,16 @@ export const REGISTERED_HOOKS = [
    */
   'modifyFocusOnTabNavigation',
 
+  /**
+   * Used to modify the cell coordinates when the table is deactivated (going out of the listen mode).
+   *
+   * @event Hooks#modifyUnfocusOnTabNavigation
+   * @since 16.1.0
+   * @param {'to_above' | 'to_below'} tabActivationDir The browsers Tab navigation direction. Depending on
+   * whether the user deactivated the table from the element above or below, another cell can be selected.
```
budnix, line 1895: The table's default behavior after the focus table exits can be skipped. Missing the `@return` tag with description.

handsontable/src/plugins/pagination/pagination.js
```diff
@@ -7,9 +7,12 @@ import { announce } from '../../utils/a11yAnnouncer';
 import { createPaginatorStrategy } from './strategies';
 import { toSingleLine } from '../../helpers/templateLiteralTag';
 import { warn } from '../../helpers/console';
+import { getMostBottomEndPosition } from '../../core/focusCatcher/utils';
```
budnix, line 10: Using helper functions from the other submodule is not the best idea. If you need that function, I'd recommend moving it to the global helper (`./src/helpers`), from which Pagination and FocusCatcher could use it.

handsontable/src/plugins/pagination/pagination.js
```diff
@@ -209,6 +212,15 @@ export class Pagination extends BasePlugin {
         a11yAnnouncer: message => announce(message),
       });
 
+      this.#ui.focusableElements.forEach((element) => {
```
budnix, line 215: I would like to see the focus management moved to the `ui` module. Best case, I'd create a separate file responsible only for focusing.

handsontable/src/plugins/pagination/pagination.js
```diff
@@ -260,10 +277,79 @@ export class Pagination extends BasePlugin {
...
+
+    pluginContext.addShortcut({
+      keys: [['Shift', 'Tab'], ['Tab']],
+      preventDefault: false,
+      callback: (event) => {
+        const { activeElement } = this.hot.rootDocument;
+        const index = this.#ui.focusableElements.indexOf(activeElement);
+
+        if (event.shiftKey) {
+          const reversedIndex = this.#ui.focusableElements.length - 1 - index;
```
budnix, line 303: As in the previous comment, it is not the best idea to have a `ui.js` module that separates DOM manipulation, and here that manipulation leaks to the `pagination.js` file, which should hold only initialization stuff. Can I ask you to move it to a separate file?

budnix (changes requested):

budnix:
Is it ok that the dialog steals the focus after showing?

https://github.com/user-attachments/assets/1340df8d-3dcb-492b-94d5-0a7363db7c34

budnix:
With that setup
```js
  dialog: {
    content: `
    <div>
      <input type="text" id="dialogInput">
      <button id="dialogButton">Click me</button>
    </div>
    `,
    background: 'semi-transparent',
    closable: true,
  },
 ```
 The <kbd>Tab</kbd> triggers the focus jumping through the dialog elements, but if I do it the other way around (<kbd>Shift</kbd>+<kbd>Tab</kbd>), it does not. I guess, not consistent behavior?

https://github.com/user-attachments/assets/2959ffbe-678c-4740-a652-c85bb50eaf55

File: handsontable/src/__tests__/core/settings.types.ts [hunk 28]
```diff
@@ -610,6 +619,7 @@ const allSettings: Required<Handsontable.GridSettings> = {
   modifyFocusedElement: (row, column, focusedElement) => document.createElement('TD'),
   modifyData: () => {},
   modifyFocusOnTabNavigation: (tabActivationDir, visualCoords) => {},
+  tableFocusExit: (exitDirection) => {},
```
budnix, line 622: Can you also add TS tests for all other new hooks that the PR adds?

handsontable/src/core/focusCatcher/index.js
```diff
@@ -3,8 +3,8 @@ import { installFocusDetector } from './focusDetector';
 import {
   normalizeCoordsIfNeeded,
   getMostTopStartPosition,
-  getMostBottomEndPosition,
 } from './utils';
+import { getMostBottomEndPosition } from '../../helpers/mixed';
```
budnix, line 7: For consistency's sake, may I suggest also moving the `getMostTopStartPosition` function to that helper?

handsontable/src/__tests__/core/settings.types.ts
budnix: I found that the `tableFocusExit` does [not have](https://github.com/handsontable/handsontable/blob/feature/dev-issue-2707/handsontable/src/dataMap/metaManager/metaSchema.js) documentation. Can I ask you to add one?

handsontable/src/plugins/dialog/dialogUI.js
```diff
@@ -0,0 +1,229 @@
...
+      // Triggers style and layout recalculation, so the display: block is fully committed before adding
+      // the class ht-dialog--show.
+      // eslint-disable-next-line no-unused-expressions
+      this.#refs.dialogElement.offsetHeight;
+    }
+
+    addClass(this.#refs.dialogElement, `${DIALOG_CLASS_NAME}--show`);
+
+    return this;
+  }
```
budnix, line 172: I propose assigning the ref to a variable. Thanks to that, the code will be more optimal for space compression.
```suggestion
  showDialog(animation) {
    const { dialogElement } = this.#refs;

    dialogElement.style.display = 'block';

    if (animation) {
      // Triggers style and layout recalculation, so the display: block is fully committed before adding
      // the class ht-dialog--show.
      // eslint-disable-next-line no-unused-expressions
      dialogElement.offsetHeight;
    }

    addClass(dialogElement, `${DIALOG_CLASS_NAME}--show`);

    return this;
  }
```

budnix:
The dialog size is not updated after the table size change.

https://github.com/user-attachments/assets/03734f5c-2027-4ab4-b6f5-bd8b130247c9

handsontable/src/plugins/pagination/pagination.js
```diff
@@ -209,6 +219,14 @@ export class Pagination extends BasePlugin {
         a11yAnnouncer: message => announce(message),
       });
 
+      this.#focusController = createPaginationController(this.#ui.getFocusableElements(), this.hot.rootDocument);
+      this.#focusController.onElementClick(() => {
+        if (!this.hot.isListening()) {
+          this.hot.listen();
+          this.hot.getShortcutManager().setActiveContextName(SHORTCUTS_CONTEXT_NAME);
+        }
+      });
```
budnix, line 228: I'd protect the `onElementClick` handler against unintentional overwriting by a developer by moving it to the module initialization phase. Also, I'd suggest using an object here for better module extensibility.
```suggestion
      this.#focusController = createPaginationController({
        focusableElements: this.#ui.getFocusableElements(),
        rootDocument: this.hot.rootDocument,
        onElementClick: () => {
          if (!this.hot.isListening()) {
            this.hot.listen();
            this.hot.getShortcutManager().setActiveContextName(SHORTCUTS_CONTEXT_NAME);
          }
        },
      });
```

budnix:
With pagination enabled (page 1 of 1), clicking the previous button (or pagination div) and hitting <kbd>Tab</kbd>, I've got an error
Kapture 2025-08-13 at 12 17 15

handsontable/src/plugins/pagination/pagination.js
```diff
@@ -868,6 +947,48 @@ export class Pagination extends BasePlugin {
...
+  }
+
+  /**
+   * Called after the selection is made. It sets the active context for the shortcuts.
+   */
+  #onAfterSelection() {
+    if (this.hot.getShortcutManager().getActiveContextName() === SHORTCUTS_CONTEXT_NAME) {
+      this.hot.getShortcutManager().setActiveContextName('grid');
+    }
+  }
```
budnix, line 990: I wonder if this "if" is necessary here? The hook is triggered after a selection is made on the table. It's not like it switches to `gird` anyway, so this hook is not needed?
```suggestion
  #onAfterSelection() {
    this.hot.getShortcutManager().setActiveContextName('grid');
  }
```

handsontable/src/plugins/pagination/paginationController.js
```diff
@@ -0,0 +1,109 @@
+/**
+ * Creates a pagination controller instance.
+ *
+ * @param {HTMLElement[]} focusableElements The focusable elements.
+ * @param {Document} rootDocument The root document.
+ * @returns {PaginationController} The pagination controller instance.
+ */
+export function createPaginationController(focusableElements, rootDocument) {
```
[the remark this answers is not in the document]
budnix: Also, I've found that you can use the internal paginator module to do that stuff. So the whole file can be replaced by this:
```js
import { createPaginator } from '../../utils/paginator';

/**
 * Creates navigator for switching the focus of the filter's elements.
 *
 * @param {BaseUI[]} elements The elements to paginate to.
 * @returns {Paginator}
 */
export function createFocusNavigator({ focusableElements }) {
  focusableElements.forEach((element) => {
    element.addEventListener('click', callback);
  });

  const navigator = createPaginator({
    initialPage: 0,
    size: () => focusableElements.length,
    onItemSelect: (currentIndex) => {
      const element = focusableElements[currentIndex];

      element.focus();
    }
  });

  return navigator;
}
```

handsontable/src/tableView.js
```diff
@@ -345,14 +345,15 @@ class TableView {
 
       this.#mouseDown = false;
 
+      const isInsideDialog = rootWrapperElement?.querySelector('.ht-dialog')?.contains(event.target);
```
budnix, line 348: It is not the best option to search the DOM for an element that belongs to the plugin to change the behavior of Core. Can I ask you to change that?

handsontable/test/helpers/common.js
```diff
@@ -559,6 +559,28 @@ export function triggerTabNavigationFromBottom(hotInstance = hot()) {
   $(hotInstance.rootWrapperElement).find('.htFocusCatcher').last().focus();
 }
 
+/**
+ * Emulates the browser's TAB navigation out of the Handsontable (to element above).
+ */
+export function triggerTabNavigationToAbove() {
+  // Trigger Shift+Tab to navigate to element above
+  // We'll use the global keyDownUp function that should be available in test environment
+  if (typeof keyDownUp === 'function') {
```
budnix, line 568: The `keyDownUp` helper is always available in the test workspace. Why then is the `triggerTabNavigationToAbove` helper needed? 🤔

budnix:
When I enable the Pagination plugin after the dialog is shown, the focus path becomes broken. After ' Shift ' + `Tab`, the pagination becomes focused instead of the top input element above the component.

Kapture 2025-08-19 at 10 38 13

budnix:
The selection is not reverted to the state before the dialog is opened when the columns, rows, or whole table was selected.

Kapture 2025-08-19 at 11 29 27

budnix:
Within the PR changes, the pagination demo looks broken (`http://192.168.1.32:8080/docs/javascript-data-grid/rows-pagination/#configure-pagination`).

Kapture 2025-08-19 at 12 00 13

budnix:
When I click the input element within the table, and at that same time the dialog shows, I cannot close it by hitting <kbd>Escape</kbd>. Clicking the dialog didn't change the shortcut context to plugin one, which probably causes the bug.

Kapture 2025-08-19 at 12 08 27

PR #11773, opened 2025-08-08
Implement the `key/value` feature for the Autocomplete and Dropdown editors.

jansiegel:
### Context
[Work in Progress]

This PR:
1. Implements the `key/value` feature for the Autocomplete and Dropdown editors.
2. Implements the `valueGetter` and `valueSetter` options (and uses them in the key/value solution + the numeric and checkbox editors).

-----

#### The `key/value` feature for the Autocomplete and Dropdown editors
This feature allows defining the `source` option for the `autocomplete` and `dropdown`-typed cells as an array of objects with `key` and `value` properties.
The choice list in the editor will display only the values under the `value` property, but save the entire key/value objects to the dataset.

_Note: When defining the `source` option as an array of objects with `key` and `value` properties, the data format for that cell needs to be an object with `key` and `value` properties as well._

-----

#### New `valueGetter` and `valueSetter` options
This PR introduces two new options:
- `valueGetter` - allows modifying the value being retrieved from a cell.
- `valueSetter` - allows modifying the value being saved to the dataset.

Both of these options can be defined for the entire table, a column, or single cells.

-----

Demo: https://jsfiddle.net/js_ziggle/h8nwrx5b/

### How has this been tested?
Added test cases.

### Types of changes

### Related issue(s):
1. handsontable/dev-handsontable#2746

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:


handsontable/src/utils/valueAccessors.js
```diff
@@ -0,0 +1,37 @@
...
+ * @param {object} cellMeta The cell meta object.
+ * @returns {*} The value to be displayed in the cell.
+ */
+export function getValueGetterValue(value, cellMeta) {
+  const { instance } = cellMeta;
+  const valueGetter = cellMeta.valueGetter ?? null;
+
+  if (isFunction(valueGetter)) {
+    return valueGetter.call(instance, value);
+  }
```
budnix, line 34: I think there is a missing piece here about what cell is processed. Otherwise, how would you format a number, date, or something on the fly without knowing the context (`cellMeta`)?
```suggestion
export function getValueGetterValue(value, cellMeta) {
  const { instance, visualRow, visualCol } = cellMeta;
  const valueGetter = cellMeta.valueGetter ?? null;

  if (isFunction(valueGetter)) {
    return valueGetter.call(instance, value, visualRow, visualCol);
  }
```

Another question that comes to my mind is, WDYT about passing the `cellMeta` to both `valueSetter` and `valueGetter` as the last argument? It seems to be helpful for devs who only want coordinates, but also for advanced formatting, the `cellMeta` will be more useful here 🤔

handsontable/src/validators/autocompleteValidator/autocompleteValidator.js
```diff
@@ -47,7 +47,7 @@ function process(value, callback) {
     let found = false;
 
     for (let s = 0, slen = source.length; s < slen; s++) {
-      if (originalVal === source[s]) {
+      if (JSON.stringify(originalVal) === JSON.stringify((source[s]))) {
```
budnix, line 50: The [`structuredClone`](https://developer.mozilla.org/en-US/docs/Web/API/Window/structuredClone) API has good browser coverage.
```suggestion
      if (structuredClone(originalVal) === structuredClone((source[s]))) {
```

handsontable/src/helpers/number.js
```diff
@@ -138,3 +138,21 @@ export function clamp(value, minValue, maxValue) {
...
   return value;
 }
+
+/**
+ * Get parsed number from numeric string.
+ *
+ * @param {string} numericData Float (separated by a dot or a comma) or integer.
+ * @returns {number|null} Number if we get data in parsable format, not changed value otherwise.
+ */
+export function getParsedNumber(numericData) {
```
budnix, line 148: Can I ask you to write a test for that helper?

budnix (changes requested):

budnix:
@jansiegel can you check that? Once you remove the cell value from the cell, you cannot choose the value from the list.

https://github.com/user-attachments/assets/a4b6ea54-8586-4b2f-80fa-4e247bb3e27d

jansiegel:
> Shouldn't it be possible to copy between cells of the same type? The populated value shouldn't be an object.
> 
> The same problem is with populating with a fill handle.

Currently, it will paste an object if the target cell contains an object as the source data and a string if it's not an object:
1. Target source data is a string: https://jsfiddle.net/js_ziggle/j0nhytp1/
2. Target source data is an object: https://jsfiddle.net/js_ziggle/xv6ce15w/

@budnix, Do you think it's a potential problem?

docs/content/guides/cell-features/conditional-formatting/angular/example1.ts
```diff
@@ -1,7 +1,7 @@
 /* file: app.component.ts */
 import { Component } from '@angular/core';
 import { GridSettings } from '@handsontable/angular-wrapper';
-import Handsontable from 'handsontable';
+import Handsontable from 'handsontable/base';
 import { registerRenderer } from 'handsontable/renderers';
```
budnix, line 6: Missing `registerAllModules()` call?

jansiegel: It's already present lower in the file https://github.com/handsontable/handsontable/blob/721d34abca28453404e497fa7276d9d622e4d526/docs/content/guides/cell-features/conditional-formatting/angular/example1.ts#L133-L134

budnix (approved):

PR #11811, opened 2025-09-02
Fix Loading Plugin screen reader

adrianspdev:
### Context
This PR includes fixes for Loading Plugin screen readers like VoiceOver.

### How has this been tested?
New e2e and visual test cases

### Types of changes

### Related issue(s):
1. https://github.com/handsontable/dev-handsontable/issues/2845

### Affected project(s):
- [x] `handsontable`

### Checklist:

[skip changelog]

### DIFF
--- a/handsontable/src/plugins/loading/loading.js
+++ b/handsontable/src/plugins/loading/loading.js
@@ -157,7 +157,7 @@ export class Loading extends BasePlugin {
         });
       }
 
-      this.hot.addHook('afterDialogFocus', () => this.#onDialogFocus());
+      this.hot.addHook('afterDialogFocus', () => this.#onAfterDialogFocus());
     }
 
     super.enablePlugin();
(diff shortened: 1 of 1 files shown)


handsontable/src/plugins/loading/loading.js
```diff
@@ -156,6 +156,8 @@ export class Loading extends BasePlugin {
           dialog: true,
         });
       }
+
+      this.hot.addHook('afterDialogFocus', () => this.#onDialogFocus());
```
budnix, line 160: That combination of the hook name and its callback, which focuses the dialog, looks weird. I'd suppose that `afterDialogFocus` is called, the dialog is already focused. So why is `this.#dialogPlugin.focus()` call again necessary?

budnix (changes requested):

handsontable/src/plugins/loading/loading.js
adrianspdev: We need to set the focus inside the loading plugin directly because the default behavior is to focus on the inputCatcher, which is announced by screen readers. If we instead set the focus directly to the dialog element, everything works as expected. I'll correct the name to match with hook name.

PR #11851, opened 2025-09-19
[Docs] netlify flaky deploy script on production 

qunabu:
netlify has a flaky deploy script that cannot be fixed. adding retry that works on staging to production

### Types of changes

### Related issue(s):
1.
2.
3.

### Affected project(s):
- [ ] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

[skip changelog]

### DIFF
--- a/.github/workflows/docs-production.yml
+++ b/.github/workflows/docs-production.yml
@@ -85,10 +85,11 @@ jobs:
           command: |
             cd docs
             cd netlify
+            npm install netlify-cli -g
             netlify deploy --prod --build --context production
         env:
           NETLIFY_AUTH_TOKEN: ${{ secrets.NETLIFY_PREVIEW_AUTH_TOKEN }}
-          NETLIFY_SITE_ID: ${{ env.NETLIFY_SITE_ID }}
+          NETLIFY_SITE_ID: ${{ secrets.NETLIFY_PROD_SITE_ID }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-production.yml
```diff
@@ -76,11 +76,17 @@ jobs:
...
+        uses: nick-fields/retry@v3
+        with:
+          timeout_seconds: 1600
+          max_attempts: 3
+          retry_on: error
+          command: |
+            cd docs
+            cd netlify
+            npm install netlify-cli -g
+            netlify deploy --prod --build --context production
```
jansiegel, line 89: Won't this fail by trying to change the directory to `docs/netlify` while being in this directory after the first try?

qunabu: Seems like it's a separete context each retry, so my guess is it won't . Be to be honest I'm not 100% sure. There is no native approach for retry in gh actions, so it needs to be using this. As far it is working in the staging environment

jansiegel: Wouldn't it make more sense to move the `cd`s and `npm install`s to the previous step to be sure? The thing that's failing is not the directory changes nor the npm installation, right?

PR #11879, opened 2025-10-09
EmptyDataState Plugin

adrianspdev:
## Context
This PR introduces the EmptyDataState Plugin, which provides a unified empty data state overlay system for Handsontable. Currently, Handsontable does not have a standardized way to display empty data states when the table has no data or when all data is hidden by filters. This plugin addresses this gap by providing a centralized empty data state indicator system.

## New Configuration Options
The plugin provides several configuration options to customize the empty data state behavior and appearance:

- **message**: Message to display in the empty data state overlay (default: uses i18n system)
  - **title**: Title to display in the empty data state overlay (default: 'No data available' from i18n)
  - **description**: Description to display in the empty data state overlay (default: 'There's nothing to display yet.' from i18n)
  - **buttons**: Buttons to display in the empty data state overlay
    - **text**: Text to display in the action button
    - **type**: Type of the action button ('primary' or 'secondary')
    - **callback**: Callback function to call when the action button is clicked

The configuration example:

```javascript
new Handsontable(element, {
  data,
  emptyDataState: {
    message: {
      title: 'No data available',
      description: 'There's nothing to display yet.',
      buttons: [{ 
        text: 'Reset filters', 
        type: 'secondary', 
        callback: () => {} 
      }],
    }
  }
});
```

## New Methods
In the emptyDataState plugin, we have the following methods:

- **isVisible()**: Check if the empty data state overlay is visible

## New Hooks
In the emptyDataState plugin, we have four new hooks:

- **beforeEmptyDataStateShow**: Hook called before the empty data state overlay is shown
- **afterEmptyDataStateShow**: Hook called after the empty data state overlay is shown
- **beforeEmptyDataStateHide**: Hook called before the empty data state overlay is hidden
- **afterEmptyDataStateHide**: Hook called after the empty data state overlay is hidden

New core hooks:

- **afterColumnSequenceCacheUpdate**: Hook after the cache of the column sequence has been updated.
- **afterRowSequenceCacheUpdate**: Hook after the cache of the row sequence has been updated.

## Internationalization Support
- Default title, description and button use i18n system with `EMPTY_DATA_STATE_TITLE` and `EMPTY_DATA_STATE_DESCRIPTION` constants for default empty data and `EMPTY_DATA_STATE_TITLE_FILTERS`, `EMPTY_DATA_STATE_DESCRIPTION_FILTERS` and `EMPTY_DATA_STATE_BUTTONS_FILTERS_RESET`
- Supports all language packs available in Handsontable
- Custom titles, descriptions and button can override default translations
- Different messages for empty data vs filtered data states
- Action button text can be customized through i18n system

## Additional validation

New validation has been added for the `getSetting` method when it is called within the source and the type is a function.

### How has this been tested?
Locally, unit, e2e and visual test cases

### Types of changes

### Related issue(s):
1. https://github.com/handsontable/dev-handsontable/issues/2369

### Affected project(s):
- [x] `handsontable`

### Checklist:

### DIFF
--- a/handsontable/src/plugins/emptyDataState/ui.js
+++ b/handsontable/src/plugins/emptyDataState/ui.js
@@ -251,7 +251,7 @@ export class EmptyDataStateUI {
       width = view.getTableWidth();
     }
 
-    if (view.isVerticallyScrollableByWindow()) {
+    if (view.isVerticallyScrollableByWindow() || view.hasVerticalScroll()) {
       if (cols > 0) {
         height = view.hot.getTableHeight() - view.getColumnHeaderHeight();
       } else {


handsontable/src/dataMap/metaManager/metaSchema.js
```diff
@@ -2099,6 +2099,90 @@ export default () => {
...
+     *         return {
+     *           title: 'No data available',
+     *           description: 'There’s nothing to display yet.',
+     *         };
+     *     }
+     *   },
+     * },
+     * ```
+     */
+    emptyDataState: false,
```
budnix: Maybe it would be better to enable the plugin by default? To show as many features as possible in the default configuration 🤔 

cc @qunabu @sl01k

handsontable/src/plugins/emptyDataState/__tests__/options/message.spec.js
```diff
@@ -0,0 +1,465 @@
...
+    expect(getEmptyDataStateContainerElement().querySelector('.ht-empty-data-state__description').textContent)
+      .toBe('Please add some data');
+
+    const actionButton = getEmptyDataStateContainerElement().querySelector('.ht-button--primary');
+
+    expect(actionButton).toBeDefined();
+    expect(actionButton.textContent).toBe('Add Data');
+    expect(actionButton).toHaveClass('ht-button', 'ht-button--primary');
+
+    actionButton.click();
```
budnix: There is a `simulateClick` helper available, which is better as it simulates all click-related events (connected with `mouse` events) in proper order to simulate a real click.
```suggestion
    await simulateClick(actionButton);
```

handsontable/src/plugins/emptyDataState/ui.js
```diff
@@ -0,0 +1,249 @@
...
+const EMPTY_DATA_STATE_CLASS_NAME = 'ht-empty-data-state';
+const MIN_HEIGHT = 200;
+
+const TEMPLATE = `<div data-ref="emptyDataStateElement" class="${EMPTY_DATA_STATE_CLASS_NAME} handsontable">
+  <div class="${EMPTY_DATA_STATE_CLASS_NAME}__content-wrapper">
+    <div data-ref="emptyDataStateInner" class="${EMPTY_DATA_STATE_CLASS_NAME}__content-wrapper-inner"></div>
+  </div>
+</div>`;
+
+const templateContent = ({ title, description }) => `
```
budnix: Have you considered combining these three templates into one large one? To maintain better separation between the logic and the view? Currently, you're using three small views, plus there is a logic that creates dynamic buttons. 

I'd be thinking of something like this (pseudo code):
```js
export function template({ id, title, message, buttons }) {
  return `
    <div data-ref="dialogElement" class="${DIALOG_CLASS_NAME} ${DIALOG_CLASS_NAME}--alert">
      <div data-ref="contentWrapperElement" class="${DIALOG_CLASS_NAME}__content-wrapper">
        <div data-ref="contentElement" class="${DIALOG_CLASS_NAME}__content-wrapper-inner">
          <div data-ref="contentElement" class="${DIALOG_CLASS_NAME}__content">
            <h2 class="ht-dialog__title">No results found</h2>
            <p class="ht-dialog__description">It looks like your current filters are hiding all results.</p>
          </div>
          <div data-ref="actionsElement" class="${DIALOG_CLASS_NAME}__actions">
            ${buttons.map(button => `
              <button id="${id}-${button.id}" class="ht-button ht-button--primary">${button.label}</button>
            `).join('')}
          </div>
        </div>
      </div>
    </div>
  `;
}
```

handsontable/types/plugins/emptyDataState/emptyDataState.d.ts
```diff
@@ -0,0 +1,24 @@
...
+  description?: string;
+  actions?: {
+    text: string;
+    type: 'primary' | 'secondary';
+    callback: () => void;
+  }[];
+}
+
+export interface EmptyDataStateConfig {
+  message?: string | ((source: string) => EmptyDataStateMessage | undefined) | EmptyDataStateMessage;
```
budnix: The `source` could be defined as a type `empty` and `filters` eventually `unknown` for unknown callers? 🤔

budnix (changes requested):

budnix:
The overlay does not work when all rows are removed.
Kapture 2025-10-23 at 13 12 18

budnix:
The overlay does not work (is not shown) when I trim all the rows, for example calling `hot.updateSettings({ trimRows: [0, 1, 2, 3, 4] })` for dataset 5x5 do nothing.
image

There is other interesting case. When the `trimRows` is defined in the table initialization config
```js
data: Handsontable.helper.createSpreadsheetData(5, 5),
trimRows: [0, 1, 2, 3, 4], // trim/hide all rows
```
The overlay is visible, but after calling the `hot.updateSettings({ trimRows: [0, 1, 4] })` (untrim 2 rows) the overlay should disappear, but it's still visible
image

budnix:
Something is not right with the `wheel` event. When I scroll the table, having the pointer over the header, only the table scrolls, the page does not move. When I scroll the table, having the pointer over the overlay, the table scrolls with the whole page.

Kapture 2025-10-23 at 13 26 46

budnix:
After clicking the overlay the table shows `isListening` state as `false`. This makes the table unusable in the context of any keyboard shortcuts. After clicking the overlay, I think the table should listen to all events from the user.

Kapture 2025-10-23 at 13 33 08

handsontable/src/dataMap/metaManager/metaSchema.js
```diff
@@ -2099,6 +2099,90 @@ export default () => {
...
+     * | --------  | ---------------------------------- | --------------------------------------------------- |
+     * | `message` | `string` \| `object` \| `function` | Message to display in the empty data state overlay. |
+     *
+     * If you set the `message` option to an object, it have following properties:
+     *
+     * | Property      | Possible values | Description                                             |
+     * | ------------- | --------------- | ------------------------------------------------------- |
+     * | `title`       | `string`        | Title to display in the empty data state overlay.       |
+     * | `description` | `string`        | Description to display in the empty data state overlay. |
+     * | `actions`     | `array`         | Actions to display in the empty data state overlay.     |
```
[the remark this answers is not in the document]
sl01k: I don’t want to decide — this should be consistent with our other parameters, and I believe the devs have a better sense and overview of it. Actions are common in DS and have a wider range — they may include more elements in the future.
Actions: https://m1.material.io/patterns/notifications.html#notifications-anatomy-of-a-notification 
Action: https://ant.design/components/alert#alert-demo-action
Action: https://v10.carbondesignsystem.com/components/notification/usage/

handsontable/src/dataMap/metaManager/metaSchema.js
sl01k: Do we have a common approach? We should be consistent. Also decision on dev side

qunabu: Lets leave it as it is, this is a breaking change imho and needs a discussion before making it happend

handsontable/src/dataMap/metaManager/metaSchema.js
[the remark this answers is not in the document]
qunabu: This was my proposition to use `actions` as i was influenced by other libraries. Although `buttons` is more meaningful. 

I @adrianspdev and @budnix thing `buttons` is better naming then let's do it, as you're the developers and we're aiming for DX.

handsontable/src/plugins/base/base.js
```diff
@@ -221,33 +222,53 @@ export class BasePlugin {
...
-      return assignObjectDefaults(this.#pluginSettings, defaultSettings)[settingName];
+    if (typeof settingValue === 'function' && settingsValidators && typeof settingsValidators === 'object') {
+      const validator = settingsValidators[settingName];
+
+      if (validator && typeof validator === 'function') {
+        return (...args) => {
+          const result = settingValue(...args);
+          const isValid = validator(result);
+
+          if (isValid === false) {
```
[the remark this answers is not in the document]
budnix: After talking with @adrianspdev on the call, it turned out that the case is covered. I basically tested incorrectly.

handsontable/src/core/hooks/constants.js
```diff
@@ -314,6 +314,17 @@ export const REGISTERED_HOOKS = [
    */
   'beforeCreateCol',
 
+  /**
+   * Fired after the cache of the column sequence has been updated.
+   *
+   * @event Hooks#afterColumnSequenceCacheUpdate
```
budnix:
```suggestion
   * Fired after the cache of the column sequence has been updated.
   *
   * @since 16.2.0
   * @event Hooks#afterColumnSequenceCacheUpdate
```

handsontable/src/core/hooks/constants.js
```diff
@@ -620,6 +631,17 @@ export const REGISTERED_HOOKS = [
    */
   'afterRenderer',
 
+  /**
+   * Fired after the cache of the row sequence has been updated.
+   *
+   * @event Hooks#afterRowSequenceCacheUpdate
```
budnix:
```suggestion
   * Fired after the cache of the row sequence has been updated.
   *
   * @since 16.2.0 
   * @event Hooks#afterRowSequenceCacheUpdate
```

handsontable/src/plugins/emptyDataState/__tests__/options/message.spec.js
```diff
@@ -0,0 +1,465 @@
...
+    expect(actionButtons.length).toBe(2);
+
+    expect(actionButtons[0].textContent).toBe('Primary Action');
+    expect(actionButtons[0]).toHaveClass('ht-button', 'ht-button--primary');
+
+    expect(actionButtons[1].textContent).toBe('Secondary Action');
+    expect(actionButtons[1]).toHaveClass('ht-button', 'ht-button--secondary');
+
+    // Test button callbacks
+    actionButtons[0].click();
```
budnix: Can I ask you to use the `simulateClick` helper here?

budnix:
After hitting the <kbd>Enter</kbd> key on the "Reset filters" button, the table loses its `isListening` state. This causes a bug where the Arrow Keys and other keyboard shortcuts do not work.

Kapture 2025-10-28 at 10 43 41

budnix:
The overlay is misplaced when the table is scrolled horizontally and the table has not defined size

Kapture 2025-10-28 at 11 10 20

budnix (approved):

budnix (approved):

budnix (approved):

PR #11994, opened 2026-01-22
Update the Angular installation guide

luk-raf:
### Context
- part with CSS import is not up to date with new version of angular. 
- information about lack of support in the SSR application

### Types of changes

### Related issue(s):
1. https://github.com/handsontable/dev-handsontable/issues/2033
2.
3.

### Affected project(s):
- [ ] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

[skip changelog]

### DIFF
--- a/docs/content/guides/getting-started/installation/installation.md
+++ b/docs/content/guides/getting-started/installation/installation.md
@@ -73,7 +73,26 @@ pnpm add handsontable @handsontable/angular-wrapper
 
 ## Import Handsontable's CSS
 
-Import Handsontable's CSS into your application to `styles.scss`.
+Import Handsontable's CSS into your application to `angular.json`.
+
+```json
+"projects": {
+    "<app_name>": {
+      "architect": {
+        "build": {
+          "options": {
+            "styles": [
+              "handsontable/styles/handsontable.min.css",
+              "handsontable/styles/ht-theme-main.min.css"
+            ]
+          }
+        }
+      }
+    }
+  }
+```
+
+Another option for adding CSS to your application is to import them in the `styles.scss` file.
 
 ```scss
 @import "handsontable/styles/handsontable.min.css";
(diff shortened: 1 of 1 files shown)


docs/content/guides/getting-started/installation/installation.md
```diff
@@ -73,7 +73,26 @@ pnpm add handsontable @handsontable/angular-wrapper
...
+            "styles": [
+              "handsontable/styles/handsontable.min.css",
+              "handsontable/styles/ht-theme-main.min.css"
+            ]
+          }
+        }
+      }
+    }
+  }
+```
```
sequba, line 93: Please, test this with Angular 16. If it works only with new versions, we need to tell our users which approach to use, depending on their Angular version.

docs/content/guides/getting-started/installation/installation.md
```diff
@@ -198,6 +217,55 @@ For more information on `@handsontable/angular`, see the [15.3 documentation](ht
 ### Troubleshooting
 If you're using Angular 21 or newer, please note that older versions of `@handsontable/angular-wrapper` are incompatible due to recent breaking changes in Angular. To ensure smooth integration, upgrade to `@handsontable/angular-wrapper@16.2` or later.
 
+## Server Side Rendering (SSR)
+
+Currently, `HotTable` does not work with SSR applications. However, if you are using SSR, you can use a simple workaround and force the table to be rendered on the browser side.
```
sequba:
```suggestion
Currently, `HotTable` cannot be rendered on the server-side. If your application uses SSR, render it only in the browser using an `@if` statement.
```

I don't like the word "workaround" here.

docs/content/guides/getting-started/installation/installation.md
```diff
@@ -198,6 +217,55 @@ For more information on `@handsontable/angular`, see the [15.3 documentation](ht
...
+    autoWrapCol: true,
+    manualRowResize: true,
+    manualColumnResize: true,
+  };
+}
+```
+
+```html
+<div>
+  @if (isBrowser) {
```
sequba: Does this syntax work with Angular 16? If not, we need to add another snippet for older versions.

sequba (changes requested):
Good job. Just remember, that the wrapper should work with Angular 16+

docs/content/guides/getting-started/installation/installation.md
luk-raf: I tested and looks good for angular 16. We can import css in main scss file or in angular.json

docs/content/guides/getting-started/installation/installation.md
luk-raf: changed

docs/content/guides/getting-started/installation/installation.md
luk-raf: I changed it to *ngIf.

sequba (dismissed):

docs/content/guides/getting-started/installation/installation.md
sequba: In angular 21 this line produce the error:

> main.ts:6 RuntimeError: NG0908: In this configuration Angular requires Zone.js

It turns out that since angular 19, the change detection mechanism is different, and zone.js is not installed by default. Can we use the default change detection mechanism or are there any issues with it?

sequba: I followed the installation guide and ran into this warning:

> Language with code "en" was not found. You should register particular language before using it. Read more about this issue at: https://docs.handsontable.com/i18n/missing-language-code.

We should either show how to register the language or remove this line from default config

luk-raf: User during the creation of new app can choose between Zone.js and zoneless. I think we can simplify snippet in documentation to show only provider what is optional for library.
`export const appConfig: ApplicationConfig = {
  providers: [
    //other providers,
    { provide: HOT_GLOBAL_CONFIG, useValue: globalHotConfig },
  ],
};`

sequba: Most of the time, developers add hot and angular-wrapper to an already existing Angular app, and we need to support both options (Zone.js and zoneless). Both options need to be described in the installation guide.

luk-raf: @sequba 
I think we can add information that the wrapper library works in zone.js and zoneless modes, but in the snippet we don't need to show what needs to be set for which mode because it's on the framework side, not the library side, and it's up to the user to decide how their application works. The wrapper does not require this to be provided for it to work properly. The wrapper only has the ability to set a global provider for configuration, and in my opinion, that is all that matters here. In my opinion, it is best to remove the providers that are standard for Angular applications and leave only the provider from HOT_GLOBAL_CONFIG to show where it can be set.

sequba: If `provideZoneChangeDetection` isn't necessary for the angular-wrapper to work, let's remove this line, and it should be fine.

luk-raf: done

luk-raf: I added correct language registration to the snippet.

sequba (approved):

PR #12001, opened 2026-01-29
Fix UndoRedo beforeChange order

jsve:
### Context

When using the beforeChange hook in React wrapper to modify or nullify changes, the UndoRedo plugin may record the original values instead of the modified values. This happens because of hook execution order differences between vanilla Handsontable and the React wrapper.

The issue:

In vanilla Handsontable or when using `addHook`, hooks registered via config are typically processed in a predictable order. However, in the React wrapper, hooks passed as props are registered differently, which can cause the UndoRedo plugin's internal beforeChange hook to run after the user's hook has already modified the changes array.

This leads to incorrect undo behavior: undoing restores values that were never actually applied to the grid.

Example scenario:

```typescript
<HotTable
  beforeChange={(changes) => {
    if (changes[0][3] > 5) {
      changes[0][3] = 5;
    }
  }}
/>
```

When entering 10, the cell shows 5 (correctly capped), but UndoRedo records 10 as the new value. Undoing then incorrectly restores to 10 instead of the previous value.

Solution:

Pass the plugin's priority `PLUGIN_PRIORITY` as the orderIndex parameter when registering the UndoRedo plugin's beforeChange hook. This ensures the plugin's hook runs at the correct priority level relative to other hooks, capturing the final modified values.

```typescript
hot.addHook('beforeChange', callback, undoRedoPlugin.constructor.PLUGIN_PRIORITY);
```

Note: This issue does not typically manifest when using vanilla Handsontable or `addHook` directly, as the hook registration order is different in those cases.

Note2: Other plugins have seemingly random values on `orderIndex` when they are applied to the same lifecycle hooks (e.g. when `manualColumnResize` and `stretchColumns` use the `beforeColumnResize` with `orderIndex` 1 and 10 respectively. This PR uses `PLUGIN_PRIORITY` instead, which seemed easier to maintain.

### How has this been tested?

I tested the changes locally and added a test to the React wrapper that verifies:
1. When beforeChange modifies a value, UndoRedo records the modified value (not the original)
2. Undo correctly restores to the previous state based on the modified values

### Types of changes

### Related issue(s):

Unable to find any.

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

### DIFF
--- a/.changelogs/12001.json
+++ b/.changelogs/12001.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Fix UndoRedo beforeChange order",
+  "type": "fixed",
+  "issueOrPR": 12001,
+  "breaking": false,
+  "framework": "none"
+}
--- a/handsontable/src/plugins/undoRedo/actions/dataChange.js
+++ b/handsontable/src/plugins/undoRedo/actions/dataChange.js
@@ -73,7 +73,7 @@ export class DataChangeAction extends BaseAction {
       };
 
       undoRedoPlugin.done(wrappedAction, source);
-    });
+    }, undoRedoPlugin.constructor.PLUGIN_PRIORITY);
   }
 
   /**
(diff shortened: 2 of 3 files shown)


budnix (dismissed):

jsve:
changelog entry added.

budnix (approved):

PR #12026, opened 2026-02-18
Fix a problem with duplicate array-based cell values in the filtering dropdown.

jansiegel:
### Context
This PR should fix two problems with the MultiSelect cell type:

1. Fix a problem with duplicate array-based cell values in the filtering dropdown.
    - when filtering a MultiSelect column and then editing any of the unfiltered cells, the values panel in the filtering dropdown would display all the instances of the same array-based values, instead of displaying every combination once.
2. Fix a problem with removing chips on a AoO dataset.
    - When using an array of objects dataset, it was impossible to remove chips using the "X" button.

[skip changelog]
<sup>Fixes for an unreleased feature</sup>

### How has this been tested?
Tested manually, added test cases.

### Types of changes

### Related issue(s):
[reported directly to me on Slack]

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

### DIFF
--- a/handsontable/src/cellTypes/multiSelectType/accessors/valueSetter.js
+++ b/handsontable/src/cellTypes/multiSelectType/accessors/valueSetter.js
@@ -5,12 +5,14 @@ import { isKeyValueObject } from '../../../helpers/object';
  *
  * @param {*} newValue The value to be set.
  * @param {number} row The visual row index.
- * @param {number} column The visual column index.
+ * @param {number|string} column The visual column index or property name.
  * @returns {*} The new value to be set.
  */
 export function valueSetter(newValue, row, column) {
+  const physicalColumn = typeof column === 'string' ? column : this.toPhysicalColumn(column);
+
   if (Array.isArray(newValue)) {
-    const sourceDataAtCell = this.getSourceDataAtCell(this.toPhysicalRow(row), this.toPhysicalColumn(column));
+    const sourceDataAtCell = this.getSourceDataAtCell(this.toPhysicalRow(row), physicalColumn);
 
     if (
       Array.isArray(sourceDataAtCell) &&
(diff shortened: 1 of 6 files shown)


handsontable/src/cellTypes/multiSelectType/accessors/valueSetter.js
```diff
@@ -5,12 +5,14 @@ import { isKeyValueObject } from '../../../helpers/object';
  *
  * @param {*} newValue The value to be set.
  * @param {number} row The visual row index.
- * @param {number} column The visual column index.
+ * @param {number|string} column The visual column index or property name.
  * @returns {*} The new value to be set.
  */
 export function valueSetter(newValue, row, column) {
+  const physicalColumn = typeof column === 'string' ? column : this.toPhysicalColumn(column);
```
budnix: Wouldn't it be better to use `propToCol`? The method already checks the type and returns the correct index.

budnix (changes requested):

handsontable/src/cellTypes/multiSelectType/accessors/valueSetter.js
jansiegel: `propToCol` returns a visual column index, and `setSourceDataAtCell` requires physical indexes ([link](https://handsontable.com/docs/javascript-data-grid/api/core/#setsourcedataatcell)).

jansiegel: edit: my bad, it's required for [getSourceDataAtCell](https://handsontable.com/docs/javascript-data-grid/api/core/#getsourcedataatcell) here, and it requires a visual column index, so even `propToCol` should not be needed here.

budnix: When a visual is required, then why is there` this.toPhysicalColumn(column)`? I think that may be enough `this.getSourceDataAtCell(this.toPhysicalRow(row), column);` (the method also accepts `prop` as column)

jansiegel: Correct, I edited the comment above:
> so even propToCol should not be needed here

I'll leave `column` here, and make adjustments to `utils.js`.

budnix (approved):

PR #12094, opened 2026-03-06
Update License bar styles and position

adrianspdev:
### Context

The PR restructures the root wrapper into ordered layout slots and migrates the built-in UI (pagination, license notification, dialog, empty-data-state) onto them. Previously each UI inserted itself ad hoc (`rootElement.after()`, a `MutationObserver` to keep position), which made ordering fragile and gave users no control over placement. This introduces a small layout subsystem plus a public `layout` setting so the order of UI elements around the grid is deterministic and user-overridable.

#### Core layout subsystem (`src/core/layout/`)

- **`DomSlot`** — manages ordered insertion of registered children in one slot element. Order: keys listed in the `layout` setting first, then remaining keys by ascending weight, ties broken by registration order. Reconciles children **in place** (only moves a node when it is not already at its target position), so re-applying an unchanged order causes no DOM churn and never blurs a focused control. Tolerates a malformed (non-array) order value.
- **`LayoutManager`** — owns one `DomSlot` per orderable slot and exposes a single-owner API: `register(key, element, { side, weight })` / `unregister(key, side)` own both placement and ordering, plus `getSlot(name)` / `applyConfig(layout)`. Everything is **data-driven from `LAYOUT_SLOTS`** (the constructor, `register`, `getSlot`, `applyConfig` all derive from it), so adding a future slot (e.g. `start`/`end`) only needs a registry entry + its element and CSS. `destroy()` keeps the slot map (only clears contents) so a plugin's `disablePlugin` running after core destroy stays safe.
- **`constants.ts`** — `LAYOUT_SLOTS` (`top`, `bottom`), `LayoutSlotName`, `LayoutSide`, `SLOT_ITEM_CLASS` (`ht-slot-element`). Ordering weights live in the owning plugins, not in core.

#### Core (`core.ts`)

- New wrapper DOM: `ht-slot-top`, `ht-grid` (`ht-grid-content` + grid), `ht-slot-bottom`, `ht-overlay`.
- New public API **`getLayoutManager()`** (root instance only; throws on non-root, like `getFocusScopeManager()`).
- Edge-slot width sync on `afterRender`, memoized (writes only on actual change) to avoid a reflow feedback loop, so the `top`/`bottom` slots stay as wide as the table.
- New **`layout`** setting (`metaSchema.ts`, `settings.ts`); applied on init and `updateSettings`. Keyed by slot: `{ top: string[], bottom: string[] }`. `LayoutConfig` is a mapped type over `LayoutSide`, so the slot keys autocomplete and type-check.

#### Overlays layer (not a slot)

- `ht-overlay` is a **fixed internal element, like `ht-grid`** — always rendered (last wrapper child), not orderable, not part of `getLayoutManager()` or the `layout` setting. It is the modal layer; the dialog mounts there and spans the whole wrapper by design (it is `aria-modal`, renders a backdrop, and traps focus).

#### License notification

- **`utils/licenseNotification.ts`** (new) — injects product info into the bottom slot element when the license is invalid/expired/missing, appends it as the **last child** of the bottom slot (tagged with `ht-slot-element` so it shares the slot-item framing), and registers a focus scope for its links (Tab/Shift+Tab). It is **not** a layout-slot contributor: it is not registered with the `LayoutManager`, so it cannot be reordered through the `layout` setting or removed via `getLayoutManager().unregister()`. Because it is a foreign (unregistered) node, `DomSlot` always inserts registered items before it, so it stays last as other contributors (pagination) register or reorder — including when pagination is enabled after init. Cannot be disabled.
- **`helpers/mixed.ts`** — `_injectProductInfo` now takes an options object `{ className, key, element, releaseDate }`, wraps the message in an inner node, and **returns the created element** so the caller positions it directly instead of re-querying the DOM by class.
- **`tableView.ts`** — removed `addClassNameToLicenseElement` / `removeClassNameFromLicenseElement`.

#### Plugins

- **Pagination** — `register(...)`s in the `bottom` slot (weight 100) when there is no custom `uiContainer`; the UI no longer self-appends (the layout manager owns placement); removed its own width/border/separator logic (now generic in core).
- **Dialog** — renders into `ht-overlay` and self-manages its element: `DialogUI.install()` appends it, `disablePlugin` detaches it (`destroyDialog()` + reset, rebuilt by `enablePlugin`). No layout-slot registration (overlays is not a slot).
- **Empty Data State** — renders inside `ht-grid` (overlays the grid), dropped the `MutationObserver`; keeps non-zero height when data is empty and `height: auto`.
- Pagination guards its layout-manager calls with `isRootInstance(this.hot)` (UI still renders on non-root; the slot only manages ordering).

#### Styles

- Wrapper border via `box-shadow`; slot items carry their own border via `.ht-slot-element` with adjacent items sharing a single divider (handled with `:first-of-type`/`:last-of-type`/`:not(...)` selectors, no JS); border-radius/background rules keyed off `:has(> .ht-slot-bottom > .ht-slot-element)` / `.ht-slot-top`.
- Empty-data-state: dropped the bottom-border override on the `:has(> .ht-slot-bottom > .ht-slot-element)` rule, so its bottom edge aligns with the slot framing now that the license notification also carries `.ht-slot-element`.

#### Documentation

- New guide: **"How to order grid UI elements"** (`docs/.../accessories-and-menus/layout-slots/`), sidebar entry, JSDoc for `getLayoutManager()` and the `layout` option. The guide documents that the license notification always renders last in the `bottom` slot and is not orderable through the `layout` setting; orderable examples use `pagination` and a custom element.

### How has this been tested?

Unit + E2E added/updated and run locally:

```bash
npm --prefix handsontable run test:unit -- --testPathPattern="domSlot|layoutManager|mixed.unit|licenseNotification"
npm --prefix handsontable run test:e2e -- --testPathPattern="core/layout|pagination/__tests__/ui|dialog/__tests__/dialog|emptyDataState/__tests__/emptyDataState"
```

- `core/layout` E2E (28): slot order, `layout` setting at init + via `updateSettings`, weight ordering, separator-on-first-item-only, width sync, focus retained across `updateSettings`, non-root throw, license appended last (not registered) + ignored when listed in `layout`, license stays last when pagination is enabled after init, pagination registration + teardown, dialog renders in the overlays layer + removed on disable, an `overlays` key in `layout` is ignored, destroy safety.
- `plugins/dialog` E2E (178) includes a regression test that the dialog modal spans the whole wrapper and covers the bottom slot (pagination/license) when shown.
- Unit: `DomSlot` (17, incl. in-place reconcile + focus retention), `LayoutManager` (12, incl. a foreign trailing element staying last as registered items are added), `licenseNotification` (11, incl. appended-last + slot class + LayoutManager untouched), `mixed` (30, incl. `_injectProductInfo` return value).

### Types of changes

**Breaking:** the root wrapper DOM structure changed and built-in UI (license notification, pagination, dialog, empty-data-state) now mounts into dedicated slot/overlay containers rather than as siblings of the grid element. Integrations that targeted the old DOM positions may need updating.

### Related issue(s):

1. https://github.com/handsontable/dev-handsontable/issues/2805
2. https://github.com/handsontable/dev-handsontable/issues/2543

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

---

> [!NOTE]
> **High Risk**
> Marked breaking: root DOM structure and where built-in UI mounts changed, so CSS/DOM-dependent integrations and visual tests must be updated; scope is large across core, plugins, focus, and styles.
> 
> **Overview**
> Introduces a **breaking** restructure of the root wrapper: `ht-slot-top`, `ht-grid` (with `ht-grid-content`), `ht-slot-bottom`, and `ht-overlay`, plus a new **`layout`** setting and **`getLayoutManager()`** API to register and order custom UI in the top/bottom slots (`register` / `unregister`, weights, keyed order).
> 
> Built-in chrome no longer inserts itself next to the grid ad hoc: **pagination** registers in the bottom slot; **dialog** mounts in `ht-overlay` as a full-wrapper modal; **empty-data-state** lives inside `ht-grid` (MutationObserver removed); the **license bar** is injected last in the bottom slot via `initLicenseNotification`, styled as `ht-slot-element`, and is **not** reorderable through `layout`. Core syncs top/bottom slot width to the table on `afterRender`.
> 
> Styling shifts to shared **`.ht-slot-element`** framing and wrapper `box-shadow` borders; pagination/dialog drop their old self-placement, width, and conditional top-border logic. Docs add a **Layout slots** guide; visual-test selectors move to `.ht-grid-content > .handsontable`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6a0a3c9faa2810295dd5233ffe5b407a83314d4c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


handsontable/src/plugins/pagination/pagination.ts
```diff
@@ -250,6 +252,14 @@ export class Pagination extends BasePlugin {
 
     }
 
+    // The layout manager only exists on the root instance. The UI itself is already in the DOM
+    // (the UI's `install` appends it to the after-grid element); the slot only manages ordering.
+    if (isRootInstance(this.hot) && !this.getSetting('uiContainer')) {
+      this.hot.getLayoutManager()
```
budnix, line 257: I'd like to fold this registration and the UI's own placement into a single `register()` call on the manager, because right now placement and ordering are split across two files and have to stay in agreement. It'd also let us drop the root-instance check here. Something like:

```js
this.hot.getLayoutManager().register(PLUGIN_KEY, this.#ui.getContainer(), {
  side: 'after',
  weight: PAGINATION_WEIGHT,
});
```

handsontable/src/plugins/pagination/ui.ts
```diff
@@ -186,7 +186,7 @@ export class PaginationUI {
 
       addClass(container, [this.#themeName ?? '', 'handsontable']);
     } else {
-      this.#rootElement.after(elements.fragment);
+      this.#afterGridElement.appendChild(elements.fragment);
```
budnix: Tied to the pagination.ts note — I'd drop the after-grid reference here and let the manager do the placement, so there's a single owner for where pagination lands instead of two.

handsontable/src/core/settings.ts
```diff
@@ -200,6 +200,11 @@ export interface GridSettings {
 
   // Layout
   ariaTags?: boolean;
+  layout?: {
```
budnix: I'd like to settle this shape before 18.0, since it's the public option and changing it later breaks people. Two reasons to revisit it: the keys are plain strings so there's no autocomplete on the valid values, and there's no room for per-item alignment if left/right ever lands. I'd propose moving toward one of these:

```js
// today
layout: { afterGrid: ['pagination', 'licenseNotification'] }

// a single ordered list, with `grid` marking the table's position
layout: ['toolbar', 'grid', 'pagination', 'licenseNotification']

// or objects, once alignment within a row matters
layout: [
  'toolbar',
  'grid',
  { key: 'pagination', align: 'left' },
  { key: 'licenseNotification', align: 'right' },
]
```

Also `layout` reads as related to `layoutDirection` right next to it when it isn't — might be worth a different name.

handsontable/src/core/settings.ts
```diff
@@ -200,6 +200,11 @@ export interface GridSettings {
...
+    beforeGrid?: string[];
+    afterGrid?: string[];
+    overlays?: string[];
```
budnix: I'd drop `overlays` from the public `layout` setting entirely. Overlay-type plugins like Dialog should declare themselves as `overlay` at registration time, and that layer's ordering isn't something users should be reshuffling at runtime — so it doesn't belong in user-facing config. That keeps `layout` scoped to the flow around the grid.

handsontable/src/core/layout/constants.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Class added to every element registered in an edge slot (before-grid/after-grid). It carries the
+ * shared slot-item border styling.
+ */
+export const SLOT_ITEM_CLASS = 'ht-slot-element';
+
+/**
+ * Default ordering weights for the built-in slot contributors. Lower weight comes first.
+ * Plugins should pick weights with gaps (multiples of 100) to leave room between built-ins.
+ */
+export const LAYOUT_WEIGHTS = {
```
budnix: I'd move these weights out of Core and into the plugins that own them. Listing `PAGINATION`, `DIALOG`, and `LICENSE_NOTIFICATION` here means the kernel knows about specific plugins, which points the dependency the wrong way and breaks the plugin isolation we hold elsewhere. Each plugin defining its own weight keeps the coupling one-directional and lets a plugin change its ordering without a Core edit.

budnix (commented):
A few suggestions on the layout API while it's still pre-release — all cheaper to change now than after 18.0 locks the public shape. None block the PR.

Main one: I'd like us to separate two concerns that are currently merged — position relative to the grid (above/below) versus alignment within a row (left/center/right). I'd want the design to leave room for the second one, since "pagination on the left" is a fair bet to come up and the current model can't express it.

I'd also lean toward collapsing `beforeGrid`/`afterGrid` into a single ordered list with `grid` as a keyword marking the table's position. I think it's a simpler mental model than two named buckets, and it drops the need to know which bucket a thing belongs to.

And internally, I'd move placement entirely into the layout manager rather than splitting it between the UI and the plugin — one owner instead of two things to keep in sync. The keys people type wouldn't change. Specifics inline.

handsontable/src/core/layout/layoutManager.ts
```diff
@@ -0,0 +1,76 @@
+import { throwWithCause } from '../../helpers/errors';
+import { DomSlot } from './domSlot';
+import { LAYOUT_SLOTS, SLOT_ITEM_CLASS, type LayoutSlotName } from './constants';
+
+/**
+ * User-facing layout configuration: an ordered list of element keys per orderable slot.
+ */
+export interface LayoutConfig {
```
budnix: I'd have `settings.ts` import and reuse this type instead of keeping its own hand-written copy, since the two can drift apart.

handsontable/src/core.ts
```diff
@@ -1533,7 +1582,43 @@ export default function Core(
 
     if (isRootInstance(this)) {
       installAccessibilityAnnouncer(instance.rootPortalElement);
-      _injectProductInfo(mergedUserSettings.licenseKey!, this.rootWrapperElement, process.env.HOT_RELEASE_DATE);
+      initLicenseNotification(instance);
+
+      // Keep the edge slots (before-grid, after-grid) as wide as the table so their content
+      // (toolbars, pagination, license notification) aligns with the grid.
+      const lastEdgeWidths = { before: -1, after: -1 };
+      const syncEdgeSlotsWidth = () => {
```
budnix, line 1590: I'd like to replace this per-render width sync with a CSS flex-row layout and let the browser keep the width in step, rather than recomputing and guarding against a reflow loop ourselves.

handsontable/src/plugins/dialog/ui.ts
```diff
@@ -163,8 +163,9 @@ export class DialogUI {
     dialogElement.addEventListener('transitionstart', () => this.#onTransitionStart());
     dialogElement.addEventListener('transitionend', () => this.#onTransitionEnd());
 
-    // Append to Handsontable after table grid element
-    this.#rootElement.after(elements.fragment);
+    if (this.#overlayContainer) {
+      this.#overlayContainer.appendChild(elements.fragment);
```
budnix, line 167: Can I ask — is the dialog meant to cover the after-grid area now? Before this PR the dialog was inserted right after the grid and `updateHeight('calc(100% - licenseInfoHeight)')` shrank it so it stopped above the license notification. Now it's appended into `ht-overlays`, which has no CSS of its own, so `.ht-dialog` (position:absolute, height:100%) sizes against `.ht-root-wrapper` and spans the whole wrapper — including `ht-after-grid` where pagination and the license notification now live. So opening a dialog or loading overlay on a grid that has pagination will visually cover the pagination bar, which it didn't before. The dialog spec only checks `parentNode`, so nothing catches the geometry change — worth a quick check.

handsontable/src/core/layout/domSlot.ts
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  /**
+   * Re-appends registered elements into the container in resolved order.
+   *
+   * @returns {void}
+   */
+  #reorder(): void {
+    this.#sortedKeys().forEach((key) => {
+      this.#parent.appendChild(this.#entries.get(key)!.element);
```
budnix: `#reorder()` re-appends every registered element on each `add()`/`setOrder()`, and since `applyConfig` calls `setOrder` on all slots during every `updateSettings()`, this runs even when the order hasn't changed. The catch is that `appendChild`-ing a node that contains `document.activeElement` fires a blur and drops focus to `<body>` — so if a keyboard user is on a pagination control or a license link when something triggers `updateSettings()`, their focus silently disappears mid-interaction. Guarding the move (only `insertBefore` when the node isn't already at the target index) would skip both the focus loss and the redundant DOM churn.

handsontable/src/utils/licenseNotification.ts
```diff
@@ -0,0 +1,94 @@
...
+    : '';
+
+  _injectProductInfo({
+    className: LICENSE_INFO_CLASS,
+    key: licenseKey,
+    element: container,
+    releaseDate
+  });
+
+  const notificationElement = getNotificationElement(hotInstance);
```
budnix: This re-queries the DOM by class to get back the element `_injectProductInfo` just created and appended, because that helper returns nothing. The coupling is a little fragile — if the injected markup or the `hot-display-license-info` class ever changes, or anything else in the after-grid carries that class, `getNotificationElement` returns null and the notification silently isn't slot-registered, so it loses its ordering and focus scope with no error. Having `_injectProductInfo` return the element it creates would let you register it directly and drop the round-trip. Curious what you think.

budnix (commented):
A few inline notes from a deeper pass over the layout-slot changes — focused on things not already covered in the existing review threads.

handsontable/src/core.ts
adrianspdev: I looked into this and a pure-CSS approach doesn't hold up here. The edge slots need to match the rendered table width (.htCore), not the wrapper width — e.g. for a content-sized table narrower than its container, the slots must shrink to the table (layout.spec.js:156 asserts afterGrid.offsetWidth === htCore.offsetWidth and < wrapper.offsetWidth).

That target width is a walkontable measurement, not a CSS relationship the edge-slot siblings can derive. Making them track it in CSS requires wrapping grid + edges in a width: fit-content container — but walkontable and stretchColumns measure available width from that container (getWorkspaceWidth, stretchColumns.ts:110). A fit-content ancestor removes the definite width they rely on, which breaks stretchH and width: '100%' (table width depends on content, content depends on container width — circular).

I checked the four cases (narrow table / stretchH:'all' / window-scrollable / fixed width) and no single CSS config satisfies all of them. The JS sync works precisely because it measures after layout, once stretch/percentage widths are resolved. The reflow-loop guard (lastEdgeWidths) keeps it to one write per actual change, so there's no feedback loop. Proposing to keep it as-is.

handsontable/src/plugins/dialog/ui.ts
adrianspdev: Yes, that's intentional. The dialog is a modal now, so it's meant to cover the whole grid wrapper — including the `bottom` slot with pagination and the license notification — not stop above them.

Moving it into the dedicated `ht-overlay` layer is what enables that. The overlays layer is a fixed, full-wrapper element (it renders like the grid, not as a slot), so `.ht-dialog` (`position: absolute; height: 100%`) sizes against the wrapper and spans the whole thing by design. The old `updateHeight('calc(100% - licenseInfoHeight)')` was a workaround for the dialog's previous DOM position (inserted right after the grid) — it existed only to avoid overlapping the license notification, and that constraint is deliberately gone now.

Covering the pagination/license bar is the correct modal behavior, not a regression:
- the dialog is `role="dialog"`/`alertdialog` with `aria-modal`,
- it renders a backdrop (`background: 'solid' | 'semi-transparent'`),
- and it registers a focus scope that traps Tab/Shift+Tab inside the dialog while open.

So while the dialog is up the user can't reach the pagination controls or license links underneath anyway — visually covering them matches that. A modal that left the pagination bar visible and clickable behind it would be the inconsistent state.

Good call on the test gap, though — the spec only checked `parentNode`. I've added a regression test that opens a dialog on a grid with pagination and asserts the dialog element spans the full wrapper and covers the bottom slot, so the intended full-cover geometry is now locked in.

handsontable/src/core/settings.ts
adrianspdev: Thanks for the detailed suggestions. We're keeping the keyed top/bottom shape — it's the clearest fit for our implementation and extends cleanly to the start/end slots we plan to add.

budnix (dismissed):

handsontable/src/core/layout/layoutManager.ts
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  /**
+   * Applies the per-slot key order from a layout configuration.
+   *
+   * @param {LayoutConfig} [config={}] The layout configuration.
+   * @returns {void}
+   */
+  applyConfig(config: LayoutConfig = {}): void {
+    Object.values(LAYOUT_SLOTS).forEach(side => this.getSlot(side).setOrder(config[side] ?? []));
```
budnix: When someone passes `layout: null`, the grid throws `Cannot read properties of null` — both at init and on every `updateSettings`. The default `= {}` only kicks in for `undefined`, so an explicit `null` slips straight through and `config[side]` reads from `null`. `DomSlot.setOrder` already tolerates a bad (non-array) value, so it'd be good to be just as forgiving here — I reproduced this in a quick demo and it breaks grid construction. Optional chaining covers both paths:
```suggestion
    Object.values(LAYOUT_SLOTS).forEach(side => this.getSlot(side).setOrder(config?.[side] ?? []));
```

budnix (commented):
Two inline notes from reviewing the layout-slots feature — one real crash on `layout: null`, one question about an unused import. Everything else I stress-tested (duplicate keys, foreign keys, foreign slots, non-array values, and trying to disable the license notification) behaved correctly.

handsontable/src/plugins/dialog/dialog.ts
```diff
@@ -1,6 +1,7 @@
 import { BasePlugin } from '../base';
 import { throwWithCause } from '../../helpers/errors';
 import { DialogUI } from './ui';
+import { isRootInstance } from '../../utils/rootInstance';
```
budnix: This `isRootInstance` import looks unused now — it only shows up on this line and ESLint doesn't flag it. Was a root-instance guard meant to go somewhere in `enablePlugin` (the way pagination guards its `getLayoutManager` calls before touching the slot), or can the import just be dropped? Worth a quick check since `rootOverlaysElement` is null on a non-root instance.

budnix (approved):

PR #12147, opened 2026-03-12
Add Server-Side Row Model dataProvider

adrianspdev:
### Context

This pull request introduces the **DataProvider** plugin and the **`dataProvider`** table option for server-driven grids. Data is loaded through `fetchRows`, which receives query parameters (`page`, `pageSize`, `sort`, `filters`) and returns `{ rows, totalRows }`. Row identity is defined via `rowId` (column data key). Create, update, and remove operations are delegated to `onRowsCreate`, `onRowsUpdate`, and `onRowsRemove`.

**Integration with existing plugins**

- **Pagination** — When both DataProvider and Pagination are configured, page and page size changes trigger `fetchRows`; the pagination UI is aligned after a successful fetch (`afterDataProviderFetch`).
- **columnSorting** — Single-column server sort: the active sort is reflected in query `sort` using the column `prop` (data key), not client-side multi-column reorder.
- **Filters** — Filter state is mapped to a server-oriented payload (same conceptual shape as Filters `exportConditions`) and passed as `filters` in the query.
- **EmptyDataState** — Loading state can follow DataProvider fetches; internal refetches may set `skipLoading` on `beforeDataProviderFetch` so the overlay is not shown for background reloads.
- **Dialog** — When the Dialog plugin is enabled, failed `fetchRows` or mutation calls can surface an alert with the error message.

**Conflicts and validation**

- A **complete** `dataProvider` object is required for the plugin to run. If the value is present but incomplete, Handsontable logs a warning and keeps the plugin disabled.
- With a valid DataProvider configuration, **`trimRows`**, **`manualRowMove`**, **`manualColumnMove`**, and **`multiColumnSorting`** do not enable; a console warning is logged for each of those options if they are still set.

#### New table option: `dataProvider`

Plain object (all keys required for the plugin to enable). Typed as `DataProviderConfig` in `handsontable/types/plugins/dataProvider/dataProvider.d.ts`.

| Property | Type | Role |
|----------|------|------|
| `rowId` | `string` or `(rowData) => unknown` | Stable id for each row (column path or getter). |
| `fetchRows` | `(queryParameters, { signal }) => Promise<{ rows, totalRows }>` | Loads the current page; receives `AbortSignal` for cancellation. |
| `onRowsCreate` | `(payload: { position, referenceRowId?, rowsAmount }) => Promise<void>` | Server create when rows are inserted (e.g. context menu). |
| `onRowsUpdate` | `(rows: { id, changes, rowData? }[]) => Promise<void>` | Server update after local edits batch (edit, paste, autofill, clear column, etc.). |
| `onRowsRemove` | `(rowIds: unknown[]) => Promise<void>` | Server delete when rows are removed. |

**`fetchRows` query parameters** (`DataProviderQueryParameters`): `page` (1-based), `pageSize`, `sort` (`{ prop, order }` or `null`), `filters` (array of `{ prop, operation, conditions }` or `null`). `prop` is the column data key.

**`fetchData` overrides** (`DataProviderFetchDataOverrides`): same query fields plus optional `skipLoading` (client-only; passed to `beforeDataProviderFetch`, not to `fetchRows`).

#### New hooks (all `@since 17.1.0`)

| Hook | When | Arguments | Return / notes |
|------|------|-----------|----------------|
| `beforeDataProviderFetch` | Before each `fetchRows` call | `queryParameters` (may include `skipLoading` for internal refetches) | Return `false` to cancel the fetch. |
| `afterDataProviderFetch` | After data loaded and applied | `result`: `{ rows, totalRows, queryParameters, columnSortConfig, filtersConditionsStack }` (plus any extra fields from the user’s `fetchRows` resolve value) | — |
| `afterDataProviderFetchError` | `fetchRows` throws (non-abort) | `error`, `queryParameters` | — |
| `afterDataProviderFetchAbort` | Fetch aborted or superseded | `queryParameters`, optional `reason` (`AbortError` when available) | — |
| `beforeRowsMutation` | Before `onRowsCreate` / `onRowsUpdate` / `onRowsRemove` | `operation`: `'create' \| 'update' \| 'remove'`, `payload` | Return `false` to cancel. |
| `afterRowsMutation` | After mutation succeeds | `operation`, `payload` | — |
| `afterRowsMutationError` | After mutation fails | `operation`, `error`, `payload` | — |

**`beforeRowsMutation` / `afterRowsMutation` / `afterRowsMutationError` payloads**

- `create`: `{ rowsCreate: { position, referenceRowId?, rowsAmount } }`
- `update`: `{ rows: RowUpdatePayload[] }`
- `remove`: `{ rowsRemove: unknown[] }`

#### `DataProvider` plugin — public methods

Access via `hot.getPlugin('dataProvider')` when the plugin is enabled.

| Method | Description |
|--------|-------------|
| `isEnabled()` | `true` only when `dataProvider` is a complete config object. |
| `getQueryParameters()` | Deep-safe copy of current `{ page, pageSize, sort, filters }`. |
| `getRowId(visualRow)` | Resolves row id from `rowId` for a visual row. |
| `fetchData(overrides?)` | Runs `fetchRows`; updates data, hooks, pagination alignment; returns `{ rows, totalRows }` or `null` (cancelled / aborted / no `fetchRows`). |
| `createRows({ position?, referenceRowId?, rowsAmount? }?)` | Calls `onRowsCreate`, then refetches with `skipLoading`. |
| `updateRows(rows)` | Calls `onRowsUpdate` for `{ id, changes, rowData? }[]`; throws if any `id` is null/undefined. |
| `removeRows(rowIds)` | Calls `onRowsRemove`; throws if any id is null/undefined; may refetch previous page if the current page becomes empty. |

**Lifecycle**: `enablePlugin` / `updatePlugin` / `disablePlugin` / `destroy` — `disable`/`destroy` abort in-flight fetches and reset internal query state.

#### Exported module symbols (`handsontable/src/plugins/dataProvider/`)

From the plugin entry (for advanced / test use): `DataProvider`, `PLUGIN_KEY` (`'dataProvider'`), `PLUGIN_PRIORITY`, `DEFAULT_PAGE_SIZE` (`10`), `DATA_PROVIDER_BATCH_UPDATE_SOURCES` (change sources that batch into one `onRowsUpdate` and skip local undo stacking when `onRowsUpdate` is set).

#### Pagination plugin

- With a complete `dataProvider` and **external** paging mode, page and page size changes trigger `DataProvider#fetchData`; the UI is synchronized from **`afterDataProviderFetch`** (total row count and current page).
- Query building prefers **`getCurrentPage()`** from the Pagination plugin when present so refetches (e.g. after sort) keep the active page consistent with the UI.

#### Filters plugin

- When DataProvider handles server-side filtering, **`beforeFilter`** can return `false` so Handsontable does not apply client-side row trimming; filters are translated to `queryParameters.filters`, page resets to `1`, and `fetchData` runs.

#### Column sorting

- Use **`columnSorting`** (single column) for server sort; **`multiColumnSorting`** is disabled when DataProvider is active.
- **`beforeColumnSort`** is handled so sort state maps to `queryParameters.sort` (`prop` + `order`) and triggers a server refetch instead of client-only reorder.

#### `trimRows`, `manualRowMove`, `manualColumnMove`

- When DataProvider is fully configured, these plugins **do not enable**; enabling them in settings still produces a **console warning** per option.

#### EmptyDataState plugin

- **`emptyDataState`** can show a **loading** branch driven by DataProvider fetch lifecycle (same hooks as server-backed loading).
- New i18n keys: `EMPTY_DATA_STATE_TITLE_LOADING`, `EMPTY_DATA_STATE_DESCRIPTION_LOADING` (and translations in all bundled language files).
- **`EmptyDataStateMessage`** type gains optional `loading?: boolean` for the loading UI.

#### Dialog plugin

- No new Dialog API. If Dialog is enabled, failed **`fetchRows`** and failed **`onRowsCreate` / `onRowsUpdate` / `onRowsRemove`** show a confirm-style alert with a title and derived error description.

#### Undo stack

- For change sources in **`DATA_PROVIDER_BATCH_UPDATE_SOURCES`**, DataProvider returns **`false`** from **`beforeUndoStackChange`** when **`onRowsUpdate`** is set, so batched server edits are not pushed onto the local undo stack.

#### Row headers (internal behavior)

- DataProvider registers **`modifyRowHeader`** so row header indices reflect **global** positions under server pagination (paged offset).

#### Existing hooks used by DataProvider (not new API)

Registered via `BasePlugin#addHook` while the plugin is enabled: **`afterInit`**, **`modifyRowHeader`**, **`beforeColumnSort`**, **`beforeUndoStackChange`**, **`afterChange`**, **`beforeAlter`**, **`afterPageChange`**, **`afterPageSizeChange`**, **`beforeFilter`**.

#### Type definitions

- New: `handsontable/types/plugins/dataProvider/dataProvider.d.ts`, `index.d.ts`.
- Updates: `handsontable/types/index.d.ts`, `settings.d.ts`, `common.d.ts`, core hooks index, `emptyDataState`, `pagination`, `filters` (where shared filter shapes apply), `test/types/dataProvider.types.ts`, settings type tests.

#### Core and registry

- `dataProvider` registered in the plugin registry and meta schema with JSDoc examples.
- `core.js` wiring so the plugin loads with the rest of the built-in set.

### How has this been tested?

- **Unit tests (Jest)** — Coverage under `handsontable/src/plugins/dataProvider/__tests__/` for methods, query building (pagination, sorting, filtering, CRUD helpers), hooks, integration with Pagination and Filters, conflict options, and dialog behavior. Additional unit tests for query modules and related plugins where behavior changed.
- **E2E tests** — Including `handsontable/test/e2e/plugins/dataProvider.spec.js` and related scenarios.
- **Local commands** (from repository root):

  ```bash
  pnpm --filter handsontable run build
  pnpm --filter handsontable run test:unit
  pnpm --filter handsontable run test:e2e
  ```

### Types of changes

- [x] Additional language file or change to the existing one (translations)

### Related issue(s):

1. https://app.clickup.com/t/9015210959/PRO-247

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

---

> [!NOTE]
> **Low Risk**
> Low risk: changes are confined to documentation, sidebar/test path registration, and new example projects without modifying runtime library code.
> 
> **Overview**
> Introduces a new *Server-side data* guide centered on the `dataProvider` option, including full JavaScript/React/Angular example implementations that demonstrate paged `fetchRows` plus create/update/remove callbacks and related hooks.
> 
> Updates the docs system to surface the new content (JSDoc config entry, Getting Started sidebar link, and docs path tests) and adds new runnable `examples/next/docs/*/data-provider` projects with simple REST/GraphQL demo servers and smoke tests. Also updates `AGENTS.md` with plugin architecture conventions (notably `SETTING_KEYS`, settings validation/defaults, and hard conflict registration) and adds a changelog entry.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2f88221ee8464c1f57d701de10b50b50fd372ebf. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


qunabu:
Task linked: [PRO-1064 Server-Side Row Model (dataProvider + pagination)](https://app.clickup.com/t/9015210959/PRO-1064)
Task linked: [PRODUCT-247 Server-Side Row Model](https://app.clickup.com/t/9015210959/PRODUCT-247)

qunabu (dismissed):

handsontable/src/core/hooks/constants.js
```diff
@@ -493,6 +493,23 @@ export const REGISTERED_HOOKS = [
    */
   'afterUpdateData',
 
+  /**
+   * Fired after the dataProvider has fetched and loaded data.
+   *
```
budnix, line 514: Missing `@since` tag

handsontable/src/dataMap/metaManager/metaSchema.js
```diff
@@ -1601,6 +1601,30 @@ export default () => {
      */
     data: undefined,
 
+    /**
+     * @description
+     * When set, the table loads data from an async provider (e.g. a REST API) instead of a static `data` array.
+     * The provider receives query parameters and an AbortSignal, and returns `{ rows, totalRows }`.
+     * Use with the `pagination` option for server-side paging. When `dataProvider` is set, the `data` option is ignored.
+     *
+     * @memberof Options#
```
budnix, line 1615: Missing `@since` tag

handsontable/src/plugins/dataProvider/dataProvider.js
```diff
@@ -0,0 +1,417 @@
+import { isFunction } from '../../helpers/function';
+import { BasePlugin } from '../base';
+import { PLUGIN_KEY as PAGINATION_PLUGIN_KEY } from '../pagination';
```
budnix: _Pagination_ and _DataProvider_ are coupled both ways: _DataProvider_ imports from _Pagination_, and _Pagination_ imports back into _DataProvider_. That's a circular dependency; any API change in one can break the other. I'd recommend making it one-way. Have _Pagination_ fire `beforePageChange`/ `afterPageSizeChange`/and other hooks/events and _DataProvider_ subscribe and call `fetchData()`. Remove all `getPlugin(DATA_PROVIDER_PLUGIN_KEY)` from _Pagination_. _Pagination_ stays generic; DataProvider owns the integration.

The above should also fix modularization. Right now, because of that close coupling once Pagination is registered, imported plugins (_ColumnSorting_, _DataProvider_) are pulled in too and can't be tree-shaken. Users who only want pagination end up with a larger bundle than they expect.

budnix (changes requested):

handsontable/src/plugins/dataProvider/dataProvider.js
```diff
@@ -0,0 +1,417 @@
...
+    }
+  }
+
+  /**
+   * Navigates to a page and fetches its data. Used by the Pagination plugin when dataProvider is active.
+   *
+   * @param {number} page Page number (1-based).
+   * @returns {Promise<void>}
+   */
+  async goToPage(page) {
```
budnix: Just a suggestion: If possible I would not re-implement the pagination method and use the Pagination hooks instead to fetch the data and block the client-side pagination. Something like you did with `onBeforeColumnSort` e.g:
```js
  this.addHook('beforePageChange', (oldPage, newPage) => {
    this.fetchData({ page: newPage });
    return false;
  });
```
The same I would go with other pagination-related methods.

handsontable/src/plugins/dataProvider/dataProvider.js
```diff
@@ -0,0 +1,417 @@
...
+   * Bound handler for beforeColumnSort (so it can be removed in disablePlugin).
+   *
+   * @private
+   * @param {Array} currentSortConfig Previous sort config.
+   * @param {Array} destinationSortConfigs New sort config (single column for columnSorting).
+   * @param {boolean} sortPossible Whether the sort request is valid.
+   * @returns {boolean|undefined} `false` to block the default in-memory sort; otherwise undefined.
+   */
+  #onBeforeColumnSortBound = (currentSortConfig, destinationSortConfigs, sortPossible) =>
+    this.#onBeforeColumnSort(currentSortConfig, destinationSortConfigs, sortPossible);
```
budnix: I would suggest to remove that bounds as they increase the size of the output file. You can use arrow function to keep the same functionality. So instead of
```js
    this.hot.removeHook('afterInit', this.#onAfterInitBound);
    this.hot.removeHook('modifyRowHeader', this.#onModifyRowHeaderBound);
    this.hot.removeHook('beforeColumnSort', this.#onBeforeColumnSortBound);
```
use original methods
```js
    this.hot.removeHook('afterInit', this.#onAfterInit);
    this.hot.removeHook('modifyRowHeader', this.#onModifyRowHeader);
    this.hot.removeHook('beforeColumnSort', this.#onBeforeColumnSort);
```
and update each method that you wany keep the ref to use arrow function
```diff
- #onAfterInit() {
+ #onAfterInit = () => {
  if (!this.isEnabled()) {
    return;
  }
```

handsontable/src/plugins/dataProvider/dataProvider.js
```diff
@@ -0,0 +1,417 @@
...
+  #onBeforeColumnSort(currentSortConfig, destinationSortConfigs, sortPossible) {
+    if (!this.isEnabled() || !sortPossible) {
+      return;
+    }
+
+    const columnSorting = this.hot.getPlugin(COLUMN_SORTING_PLUGIN_KEY);
+
+    if (columnSorting?.enabled) {
+      columnSorting.setSortConfig(destinationSortConfigs);
+    }
```
budnix: I think that this logic is redundant - I guess. The `onBeforeColumnSort` is triggered when the plugin is enabled (and only then) so there is no need to double check.

handsontable/src/plugins/dataProvider/dataProvider.js
```diff
@@ -0,0 +1,417 @@
...
+    await this.fetchData({ pageSize: numericPageSize, page: 1 });
+  }
+
+  /**
+   * Updates sort and refetches. Call when column sort changes (e.g. from ColumnSorting plugin).
+   *
+   * @param {object|null} sort Sort descriptor, e.g. `{ column: 0, sortOrder: 'asc' }` or null to clear.
+   * @returns {Promise<void>}
+   */
+  async setSort(sort) {
```
budnix: The same as with pagination I believe that the sorting should be changed only in the _ColumnSorting_ plugin and the DataProvider listens to that inputs via hooks, block client-side sorting, fetches the data based on the user input and updates the data.

handsontable/src/plugins/dataProvider/dataProvider.js
```diff
@@ -0,0 +1,417 @@
...
+    await this.fetchData({ sort });
+  }
+
+  /**
+   * Updates filters and refetches.
+   *
+   * @param {object|null} filters Filter state to send to the provider.
+   * @returns {Promise<void>}
+   */
+  async setFilters(filters) {
```
budnix: The same as with pagination I believe that the filter criteria should be changed only in the _Filters_ plugin and the DataProvider listens to that inputs via hooks, block client-side sorting, fetches the data based on the user input and updates the data.

handsontable/src/plugins/manualRowMove/manualRowMove.js
```diff
@@ -82,11 +92,23 @@ export class ManualRowMove extends BasePlugin {
...
+    const dataProviderEnabled = !!settings.dataProvider;
+
+    if (pluginEnabled && dataProviderEnabled) {
+      warn(toSingleLine`The \`manualRowMove\` plugin cannot be used with the \`dataProvider\` option.\x20
+        This combination is not supported. The plugin will remain disabled.`);
+
+      return false;
+    }
+
+    return pluginEnabled;
```
budnix: The plugin shouldn't need to know DataProvider exists. If possible I would move that logic to DataProvider plugin. Similar to https://github.com/handsontable/handsontable/blob/develop/handsontable/src/plugins/pagination/utils.js#L19. Where the Pagination checks which plugin/option it works with and not the other way around.

handsontable/src/plugins/multiColumnSorting/multiColumnSorting.js
```diff
@@ -87,11 +97,23 @@ export class MultiColumnSorting extends ColumnSorting {
...
+    const dataProviderEnabled = !!settings.dataProvider;
+
+    if (pluginEnabled && dataProviderEnabled) {
+      warn(toSingleLine`The \`multiColumnSorting\` plugin cannot be used with the \`dataProvider\` option.\x20
+        This combination is not supported. The plugin will remain disabled.`);
+
+      return false;
+    }
+
+    return pluginEnabled;
```
budnix: The plugin shouldn't need to know DataProvider exists. If possible I would move that logic to DataProvider plugin.

handsontable/src/plugins/pagination/pagination.js
```diff
@@ -7,6 +7,9 @@ import { announce } from '../../utils/a11yAnnouncer';
 import { createPaginatorStrategy } from './strategies';
 import { toSingleLine } from '../../helpers/templateLiteralTag';
 import { warn } from '../../helpers/console';
+import {
+  PLUGIN_KEY as DATA_PROVIDER_PLUGIN_KEY
+} from '../dataProvider';
```
budnix: That imports bound the two plugins, so it would not be possible to use client-side pagination without the code that supports server-side pagination.

handsontable/src/plugins/pagination/pagination.js
```diff
@@ -236,19 +240,129 @@ export class Pagination extends BasePlugin {
...
+    }
+    if (oldPageSize !== this.#pageSize) {
+      this.hot.runHooks('afterPageSizeChange', oldPageSize, this.#pageSize);
+    }
+  }
+
+  /**
+   * Sets the current page (internal state only). Use before #refreshUI when changing page.
+   *
+   * @private
```
budnix: The tag can be removed from all private methods
```suggestion
```

handsontable/src/plugins/pagination/pagination.js
```diff
@@ -883,6 +1036,15 @@ export class Pagination extends BasePlugin {
...
   }
 
+  /**
+   * Bound handler for afterDataProviderFetch (so it can be removed in disablePlugin).
+   *
+   * @private
+   * @param {object} result Result from afterDataProviderFetch: `{ queryParameters }`.
+   * @returns {void}
+   */
+  #onAfterDataProviderFetchBound = result => this.#onAfterDataProviderFetch(result);
```
budnix: Use an arrow function instead, as I mentioned in other comments.

handsontable/src/plugins/trimRows/__tests__/trimRows.spec.js
```diff
@@ -71,6 +71,50 @@ describe('TrimRows', () => {
     expect(getCellMeta(6, 0).type).toBe('text');
   });
 
+  describe('dataProvider conflict', () => {
```
budnix: I think that all tests related to the dataProvider that touch other plugins should be moved to the dataProvider space. TrimRows, ManualRowMove, MultiColumnSorting etc., don't know whether dataProvider exists, and it's the module's responsibility to check whether it's compatible with others.

handsontable/src/plugins/trimRows/trimRows.js
```diff
@@ -172,12 +182,24 @@ export class TrimRows extends BasePlugin {
...
    * Checks if the plugin is enabled in the handsontable settings. This method is executed in {@link Hooks#beforeInit}
-   * hook and if it returns `true` then the {@link AutoRowSize#enablePlugin} method is called.
+   * hook and if it returns `true` then the {@link TrimRows#enablePlugin} method is called.
+   * Disabled when `dataProvider` is used (server-side data); a warning is emitted in that case.
    *
    * @returns {boolean}
    */
   isEnabled() {
-    return !!this.hot.getSettings()[PLUGIN_KEY];
+    const settings = this.hot.getSettings();
```
budnix: This check should be moved to the dataProvider plugin.

handsontable/src/core.js
```diff
@@ -2886,7 +2886,16 @@ export default function Core(rootContainer, userSettings, rootInstanceSymbol = f
     }
 
     // Load data or create data map
-    if (settings.data === undefined && tableMeta.data === undefined) {
+    if (settings.dataProvider !== undefined) {
```
budnix: If possible, I would move that logic to the plugin's space.

handsontable/types/plugins/dataProvider/dataProvider.d.ts
```diff
@@ -0,0 +1,54 @@
...
+import { CellValue } from '../../common';
+
+/**
+ * Query parameters passed to the dataProvider function and hooks.
+ */
+export interface DataProviderQueryParameters {
+  page: number;
+  pageSize: number;
+  sort: object | null;
+  filters: object | null;
```
budnix: I think that it's worth importing the types from _ColumnSorting_ and _Filters_ and using them here instead of `object`, which tells us not much.

handsontable/src/core.js
```diff
@@ -64,6 +64,7 @@ import {
 import { getValueSetterValue } from './utils/valueAccessors';
 import { createThemeManager } from './themes/engine';
 import { getTheme, hasTheme, registerTheme, mainTheme } from './themes';
+import { isCompleteDataProviderConfig } from './plugins/dataProvider/utils';
```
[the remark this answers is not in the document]
budnix: @adrianspdev that point mentioned by the bot is very important.

handsontable/src/plugins/manualColumnMove/manualColumnMove.js
```diff
@@ -119,6 +121,12 @@ export class ManualColumnMove extends BasePlugin {
       return;
     }
 
+    if (isCompleteDataProviderConfig(this.hot.getSettings().dataProvider)) {
+      warn('manualColumnMove is not compatible with dataProvider');
+
+      return;
+    }
```
budnix: If possible I would try to decouple the plugin from _DataProvider_. I have provided the most optimistic plan for do that https://gist.github.com/budnix/13d444638b9ad412fe6d6b63de7b2cac. Or it could be simplified to checking only the `this.hot.getSettings().dataProvider` (true-ish or not) and adding `hasExternalDataSource` hook in the Core that would be consumed by DataProvider plugin.

handsontable/src/dataMap/metaManager/metaSchema.js
```diff
@@ -1603,6 +1603,40 @@ export default () => {
...
+     * dataProvider: {
+     *   rowId: 'id',
+     *   fetchRows: async (queryParameters, { signal }) => {
+     *     const { page, pageSize, sort, filters } = queryParameters;
+     *     const params = new URLSearchParams({ page: String(page), pageSize: String(pageSize) });
+     *     if (sort) { params.set('sortBy', sort.prop); params.set('sortDir', sort.order); }
+     *     const res = await fetch(`/api/products?${params}`, { signal });
+     *     const json = await res.json();
+     *     return { rows: json.data, totalRows: json.total };
+     *   },
```
budnix, line 1637:
```suggestion
     *   fetchRows: async (queryParameters, { signal }) => {
     *     const { page, pageSize, sort, filters } = queryParameters;
     *     const params = new URLSearchParams({ page: String(page), pageSize: String(pageSize) });
     *
     *     if (sort) { 
     *       params.set('sortBy', sort.prop); 
     *       params.set('sortDir', sort.order);
     *     }
     *
     *     const res = await fetch(`/api/products?${params}`, { signal });
     *     const json = await res.json();
     *
     *     return { rows: json.data, totalRows: json.total };
     *   },
```

handsontable/src/plugins/dataProvider/index.js
```diff
@@ -0,0 +1,7 @@
+export {
+  DataProvider,
+  PLUGIN_KEY,
+  PLUGIN_PRIORITY,
+  DEFAULT_PAGE_SIZE,
+  DATA_PROVIDER_BATCH_UPDATE_SOURCES,
+} from './dataProvider';
```
budnix, line 5: Is this still needed? I think that the plugin should not export its internals.
```suggestion
export {
  DataProvider,
  PLUGIN_KEY,
  PLUGIN_PRIORITY,
} from './dataProvider';
```

handsontable/src/plugins/dataProvider/dataProvider.js
```diff
@@ -0,0 +1,914 @@
...
+    return PLUGIN_KEY;
+  }
+
+  static get PLUGIN_PRIORITY() {
+    return PLUGIN_PRIORITY;
+  }
+
+  static get DEFAULT_SETTINGS() {
+    return {};
+  }
```
budnix, line 138: I wonder why did you not used `static get SETTINGS_VALIDATORS()` validator here? Similar to https://github.com/handsontable/handsontable/blob/develop/handsontable/src/plugins/dialog/dialog.js#L190.

handsontable/src/plugins/dataProvider/dataProvider.js
```diff
@@ -0,0 +1,914 @@
...
+  #getFetchFn() {
+    const c = this.#getConfig();
+
+    return c && isFunction(c.fetchRows) ? c.fetchRows : undefined;
+  }
+
+  /**
+   * @returns {Function|undefined}
+   */
+  #getOnRowsCreate() {
```
budnix, line 485: Those all methods seems duplicate the functionality that Base already offers. I mean after implementing setting validator they can be gettered safely via `this.getSetting` method.

handsontable/src/plugins/dataProvider/dataProvider.js
```diff
@@ -0,0 +1,914 @@
...
+   * @inheritdoc
+   * @param {object} newSettings Settings passed to `updateSettings`.
+   */
+  onUpdateSettings(newSettings) {
+    if (newSettings && Object.prototype.hasOwnProperty.call(newSettings, PLUGIN_KEY)) {
+      this.#applyIncompleteDataProviderWarning(newSettings[PLUGIN_KEY]);
+    }
+
+    super.onUpdateSettings(newSettings);
+  }
```
budnix: I think this also tries to reimplement the SETTINGS_VALIDATOR that already sits in the Base plugin?

handsontable/src/plugins/dataProvider/dataProvider.js
```diff
@@ -0,0 +1,914 @@
...
+      ['afterPageSizeChange', this.#onAfterPageSizeChangeExternalPagination],
+      ['beforeFilter', this.#onBeforeFilter],
+    ];
+
+    specs.forEach((spec) => {
+      const [name, handler, order] = spec;
+
+      this.addHook(name, handler, order);
+    });
+  }
```
budnix: Can I ask you to rewrite that hooks to follow the pattern that is worked very well across all plugins (https://github.com/handsontable/handsontable/blob/develop/handsontable/src/plugins/pagination/pagination.js#L227-L235)?

handsontable/src/plugins/dataProvider/dataProvider.js
```diff
@@ -0,0 +1,914 @@
...
+   */
+  #runAfterDataProviderFetchAbort(params, reason) {
+    const queryParameters = this.#snapshotQueryParameters(params);
+
+    if (typeof reason === 'undefined') {
+      this.hot.runHooks('afterDataProviderFetchAbort', queryParameters);
+    } else {
+      this.hot.runHooks('afterDataProviderFetchAbort', queryParameters, reason);
+    }
+  }
```
budnix, line 596:
```suggestion
  #runAfterDataProviderFetchAbort(params, reason) {
    const queryParameters = this.#snapshotQueryParameters(params);

    this.hot.runHooks('afterDataProviderFetchAbort', queryParameters, reason);
  }
```

handsontable/src/plugins/dataProvider/dataProvider.js
```diff
@@ -0,0 +1,914 @@
...
+  updatePlugin() {
+    this.disablePlugin();
+    this.enablePlugin();
+
+    if (this.isEnabled() && this.hot.view) {
+      this.fetchData();
+    }
+
+    super.updatePlugin();
+  }
```
budnix: Can I ask you to move that public method to the top then private to the bottom? Also  enablePlugin, updatePlugin, disablePlugin should be on the top. As follow: https://github.com/handsontable/handsontable/blob/develop/handsontable/src/plugins/pagination/pagination.js#L166

handsontable/src/plugins/dataProvider/dataProvider.js
```diff
@@ -0,0 +1,914 @@
...
+    handleAfterPageChangeExternalPagination(
+      {
+        isEnabled: () => this.isEnabled(),
+        hot: this.hot,
+        getQueryPage: () => this.#queryParameters.page,
+        goToPage: page => this.#goToPage(page),
+      },
+      oldPage,
+      newPage
+    );
```
budnix, line 738: The `isEnabled: () => this.isEnabled(),` check is no needed here. When The DataProvider hook is disabled it's not listening for any events from others so the logic is not triggered at all. So the check is redundant.

handsontable/src/plugins/dataProvider/dataProvider.js
```diff
@@ -0,0 +1,914 @@
...
+   * Skips when `#queryParameters` already match (e.g. duplicate `afterPageSizeChange` from Pagination sync).
+   *
+   * @param {number | 'auto'} oldPageSize Previous page size.
+   * @param {number | 'auto'} newPageSize New page size.
+   * @returns {void}
+   */
+  #onAfterPageSizeChangeExternalPagination = (oldPageSize, newPageSize) => {
+    handleAfterPageSizeChangeExternalPagination(
+      {
+        isEnabled: () => this.isEnabled(),
```
budnix: Is not needed
```suggestion
        isEnabled: () => this.isEnabled(),
```

handsontable/src/plugins/dataProvider/dataProvider.js
```diff
@@ -0,0 +1,914 @@
...
+  /**
+   * Intercepts filter action: applies server-side filters and refetches; returns false so Filters skip client-side trimming.
+   *
+   * @param {Array} conditionsStack Exported filter conditions (column = physical index).
+   * @returns {boolean} False to signal that filtering is handled server-side.
+   */
+  #onBeforeFilter = (conditionsStack) => {
+    if (!this.isEnabled()) {
+      return;
+    }
```
budnix: Is not needed

handsontable/src/plugins/dataProvider/dataProvider.js
```diff
@@ -0,0 +1,914 @@
...
+  /**
+   * @param {Array} currentSortConfig Current sort config.
+   * @param {Array} destinationSortConfigs Destination sort config.
+   * @param {boolean} sortPossible Whether sort is allowed.
+   * @returns {boolean|undefined}
+   */
+  #onBeforeColumnSort = (currentSortConfig, destinationSortConfigs, sortPossible) => handleBeforeColumnSortForServer(
+    {
+      hot: this.hot,
+      isEnabled: () => this.isEnabled(),
```
budnix: Is not needed

handsontable/src/plugins/dataProvider/dataProvider.js
```diff
@@ -0,0 +1,914 @@
...
+    if (!dialogPlugin?.enabled) {
+      return;
+    }
+
+    const titles = {
+      fetch: 'Could not load data',
+      create: 'Could not create rows',
+      update: 'Could not update rows',
+      remove: 'Could not remove rows',
+    };
```
budnix: Shouldn't be moved to languages files?

budnix (changes requested):

handsontable/src/plugins/dataProvider/dataProvider.js
```diff
@@ -0,0 +1,937 @@
...
+    }
+  }
+
+  /**
+   * @param {number} page 1-based page index (values below 1 are treated as 1 via `fetchData`).
+   * @returns {Promise<void>}
+   */
+  async #goToPage(page) {
+    await this.fetchData({ page });
+  }
```
budnix: The page change should come only from the Pagination plugin. I think leaving the pagination method here isn't the best idea.

handsontable/src/plugins/dataProvider/dataProvider.js
```diff
@@ -0,0 +1,937 @@
...
+
+  /**
+   * @param {number} pageSize New page size.
+   * @returns {Promise<void>}
+   */
+  async #setPageSize(pageSize) {
```
budnix: The page size change should come only from the Pagination plugin. I think leaving the pagination method here isn't the best idea.

handsontable/src/plugins/dataProvider/dataProvider.js
```diff
@@ -0,0 +1,937 @@
...
+   * Disables the plugin, aborts fetch, resets query state.
+   * Hook listeners registered with `addHook` are removed by `super.disablePlugin()` via `clearHooks()`.
+   */
+  disablePlugin() {
+    this.#resetAbortController();
+    this.#queryParameters = { ...INITIAL_QUERY_PARAMETERS };
+
+    super.disablePlugin();
+
+    this.hot.render();
```
budnix: Why is there a render need when the plugin is disabled? 🤔

handsontable/src/plugins/dataProvider/dataProvider.js
```diff
@@ -0,0 +1,847 @@
...
+registerConflict(
+  'dataProvider',
+  settings => isCompleteDataProviderConfig(settings.dataProvider),
+  ['manualRowMove', 'manualColumnMove', 'trimRows', 'multiColumnSorting'],
+);
+
+Hooks.getSingleton().register('hasExternalDataSource');
+Hooks.getSingleton().add('hasExternalDataSource', function hasExternalDataSourceDefault() {
+  return isCompleteDataProviderConfig(this.getSettings().dataProvider);
+});
```
budnix: This creates a bug: when one instance uses dataProvider, it influences all other Handsontable instances, so it's not possible to use server-side and client-side data on the same page. To fix that move that logic to plugin's hook (`this.addHook`)

handsontable/src/plugins/dataProvider/dataProvider.js
```diff
@@ -0,0 +1,847 @@
...
+import { Hooks } from '../../core/hooks';
+import { registerConflict } from '../base/conflictRegistry';
+
+registerConflict(
+  'dataProvider',
+  settings => isCompleteDataProviderConfig(settings.dataProvider),
+  ['manualRowMove', 'manualColumnMove', 'trimRows', 'multiColumnSorting'],
+);
+
+Hooks.getSingleton().register('hasExternalDataSource');
```
budnix: Registering hooks is no longer necessary; I'd remove that line.

handsontable/src/plugins/base/base.js
```diff
@@ -187,6 +191,25 @@ export class BasePlugin {
...
+    const conflict = getHardConflict(this.hot.getSettings(), pluginKey);
+
+    if (conflict) {
+      warn(`${pluginKey} is not compatible with ${conflict.ownerPluginKey}`);
+
+      return true;
+    }
+
+    return false;
+  }
```
budnix, line 217: I believe there is a bug. When the plugin detects conflicting options, it should disable itself until the conflicts are resolved. Currently, the behavior is reversed. The DataProvider registers ['manualRowMove', 'manualColumnMove', 'trimRows', 'multiColumnSorting'], which results in all of these options being disabled. Instead, only the DataProvider should be disabled, while the other options remain unaffected.

handsontable/src/plugins/base/conflictRegistry.js
```diff
@@ -0,0 +1,44 @@
...
+ * @param {function(Handsontable.DefaultSettings): boolean} predicate Whether the conflict is active for the given settings.
+ * @param {string[]} blockedPluginKeys `PLUGIN_KEY` values to block.
+ */
+export function registerConflict(ownerPluginKey, predicate, blockedPluginKeys) {
+  hardConflicts.push({
+    ownerPluginKey,
+    predicate,
+    blockedPluginKeys,
+  });
+}
```
budnix, line 96: Let's make the module more generic, not only plugins specyfic. The conflicting keys can also come from other Core places (the Pagination is an example). With that module, the [Pagination logic](https://github.com/handsontable/handsontable/blob/develop/handsontable/src/plugins/pagination/utils.js#L4) could be updated to use the new mechanism.

```suggestion
/**
 * Registers a hard conflict: when `predicate(settings)` is true, plugins in `blockedPluginKeys` must not enable.
 * The owning plugin (for example DataProvider) is responsible for calling this at module load.
 *
 * @param {string} ownerSetting Owner setting key.
 * @param {function(Handsontable.DefaultSettings): boolean} predicate Whether the conflict is active for the given settings.
 * @param {string[]} conflictedSettings Conflicted settings.
 */
export function registerConflict(ownerSetting, predicate, conflictedSettings) {
  hardConflicts.push({
    ownerSetting,
    predicate,
    conflictedSettings,
  });
}
```

handsontable/src/plugins/emptyDataState/emptyDataState.js
```diff
@@ -260,11 +285,28 @@ export class EmptyDataState extends BasePlugin {
 
     this.addHook('afterInit', () => this.#onAfterInit());
     this.addHook('afterRender', () => this.#onAfterRender());
-    this.addHook('afterRowSequenceCacheUpdate', () => this.#toggleEmptyDataState());
-    this.addHook('afterColumnSequenceCacheUpdate', () => this.#toggleEmptyDataState());
+    this.addHook('afterRowSequenceCacheUpdate', () => {
+      if (!this.#loadingActive) {
+        this.#toggleEmptyDataState();
+      }
```
budnix: Here is the first `if` and the second in the `#toggleEmptyDataState` method. I wonder if this can't be unified? And call the `#toggleEmptyDataState` without a condition check?

handsontable/src/plugins/manualColumnMove/manualColumnMove.js
```diff
@@ -119,6 +120,10 @@ export class ManualColumnMove extends BasePlugin {
       return;
     }
 
+    if (this.isHardConflictBlocked()) {
+      return;
+    }
```
budnix: It should be handled in the plugins abstraction (Base class).
```suggestion
```

handsontable/src/plugins/manualRowMove/manualRowMove.js
```diff
@@ -97,6 +102,10 @@ export class ManualRowMove extends BasePlugin {
       return;
     }
 
+    if (this.isHardConflictBlocked()) {
+      return;
+    }
```
budnix:
```suggestion
```

handsontable/src/plugins/multiColumnSorting/multiColumnSorting.js
```diff
@@ -87,17 +91,26 @@ export class MultiColumnSorting extends ColumnSorting {
...
    * Enables the plugin functionality for this Handsontable instance.
    */
   enablePlugin() {
+    if (this.enabled) {
+      return;
+    }
+
+    if (this.isHardConflictBlocked()) {
+      return;
+    }
```
budnix:
```suggestion
```

handsontable/src/plugins/trimRows/trimRows.js
```diff
@@ -188,6 +193,10 @@ export class TrimRows extends BasePlugin {
       return;
     }
 
+    if (this.isHardConflictBlocked()) {
+      return;
+    }
```
budnix:
```suggestion
```

budnix (changes requested):

handsontable/src/plugins/dataProvider/dataProvider.js
```diff
@@ -0,0 +1,854 @@
...
+  'manualRowMove',
+  'manualColumnMove',
+  'trimRows',
+  'multiColumnSorting',
+]);
+
+Hooks.getSingleton().register('hasExternalDataSource');
+Hooks.getSingleton().add('hasExternalDataSource', function hasExternalDataSourceDefault() {
+  return isCompleteDataProviderConfig(this.getSettings().dataProvider);
+});
```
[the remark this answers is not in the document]
budnix: @adrianspdev this is very important

handsontable/src/plugins/dataProvider/dataProvider.js
budnix: @adrianspdev ?

handsontable/src/plugins/dataProvider/dataProvider.js
budnix: @adrianspdev ?

handsontable/src/plugins/dataProvider/dataProvider.js
```diff
@@ -0,0 +1,854 @@
...
+
+  /**
+   * Enables the plugin, syncs query parameters from Pagination, ColumnSorting, and Filters, and registers hooks.
+   */
+  enablePlugin() {
+    if (this.isHardConflictBlocked()) {
+      this.hot.getSettings()[PLUGIN_KEY] = false;
+
+      return;
+    }
```
budnix:
```suggestion
```
I should be handled in the abstraction (Base class)

handsontable/src/plugins/pagination/pagination.js
budnix: Since `isHardConflictBlocked` checks the plugin enable status in the Base class, it should be removed from here.

budnix (changes requested):

budnix:
Worth checking. When I enable `emptyDataState`, I get an error 
```
string.js:121 Uncaught TypeError: Cannot read properties of undefined (reading 'sanitize')
    at sanitize (string.js:121:1)
    at stripTags (string.js:110:1)
    at templateContent (ui.js:23:1)
```

<details>
  <summary>My setup</summary>

```js
dataProvider: {
  rowId: 'id',

  fetchRows: async (queryParameters, { signal }) => {
    addLog('fetch', 'fetchRows →', queryParameters);
    const result = await apiFetch('/api/rows', queryParameters, signal);
    addLog('fetch', `fetchRows ← ${result.rows.length} rows (total: ${result.totalRows})`);
    return result;
  },

  onRowsCreate: async (payload) => {
    addLog('create', 'onRowsCreate →', payload);
    const result = await apiFetch('/api/rows/create', payload);
    addLog('create', 'onRowsCreate ←', result);
  },

  onRowsUpdate: async (rows) => {
    addLog('update', 'onRowsUpdate →', rows);
    const result = await apiFetch('/api/rows/update', { rows });
    addLog('update', 'onRowsUpdate ←', result);
  },

  onRowsRemove: async (rowIds) => {
    addLog('remove', 'onRowsRemove →', { rowIds });
    const result = await apiFetch('/api/rows/remove', { rowIds });
    addLog('remove', 'onRowsRemove ←', result);
  },
},
columns: [
  { data: 'A', title: 'Column A' },
  { data: 'B', title: 'Column B' },
  { data: 'C', title: 'Column C' },
  { data: 'D', title: 'Column D' },
  { data: 'E', title: 'Column E' },
],
// dialog: true,
emptyDataState: true,
manualColumnMove: true,
colHeaders: true,
rowHeaders: true,
dropdownMenu: true,
filters: true,
pagination: {
  pageSize: 15,
},
columnSorting: true,
contextMenu: true,
hiddenColumns: true,
width: '100%',
height: 450,
licenseKey: 'non-commercial-and-evaluation',
```

adrianspdev:
> Worth checking. When I enable `emptyDataState`, I get an error
> 
> ```
> string.js:121 Uncaught TypeError: Cannot read properties of undefined (reading 'sanitize')
>     at sanitize (string.js:121:1)
>     at stripTags (string.js:110:1)
>     at templateContent (ui.js:23:1)
> ```
> 
> My setup

I just have this warning in the console

_The `dataProvider` plugin cannot be used with the `manualColumnMove` option. This combination is not supported. The plugin will remain disabled._

budnix:
> I just have this warning in the console
> 
> _The `dataProvider` plugin cannot be used with the `manualColumnMove` option. This combination is not supported. The plugin will remain disabled._

Thanks for checking. Maybe it was fixed in the last commit.

handsontable/src/core.js
```diff
@@ -2886,7 +2896,28 @@ export default function Core(rootContainer, userSettings, rootInstanceSymbol = f
...
+        warn('The "data" setting is ignored when "dataProvider" is set. ' +
+          'The DataProvider plugin will load data instead.');
+      }
+
+      // Replace the in-memory placeholder only when the update touches init, `data`, or `dataProvider`. Otherwise
+      // `updateData([], …)` would empty the grid without a refetch, because DataProvider runs `updatePlugin` only when
+      // `dataProvider` is present in this payload.
+      const shouldSyncExternalPlaceholderData = init ||
+        hasOwnProperty(settings, 'data') ||
+        hasOwnProperty(settings, 'dataProvider');
```
budnix, line 2910: The Core does not know anything about the _DataProvider_. The `hasExternalDataSource` hook can be used by 3rdparty plugin with similar functionality but under a different name.
```suggestion
      if (settings.data) {
        warn('The "data" setting is ignored when "hasExternalDataSource" returns `true`.');
      }

      // Replace the in-memory placeholder only when the update touches init, `data`, or `dataProvider`. Otherwise
      // `updateData([], …)` would empty the grid without a refetch, because DataProvider runs `updatePlugin` only when
      // `dataProvider` is present in this payload.
      const shouldSyncExternalPlaceholderData = init ||
        hasOwnProperty(settings, 'data')
```

budnix (changes requested):

budnix (dismissed):

budnix (dismissed):

budnix (approved):

BaptisteVCA:
Hello,

In my application, I load all the rows into Handsontable. Does this plugin limit searching and filtering to only the loaded rows?

Thank you for your feedback.

adrianspdev:
> Hello,
> 
> In my application, I load all the rows into Handsontable. Does this plugin limit searching and filtering to only the loaded rows?
> 
> Thank you for your feedback.

Hi,

When you are using a data provider and want to filter or sort a column, you need to handle it on the server side using [query parameters](https://handsontable.com/docs/javascript-data-grid/server-side-data/#query-parameters).

If you're experiencing any issues, please open an GH issue and include an example along with reproduction steps. This will help us better understand the problem and assist you more effectively.

Thanks!

PR #12163, opened 2026-03-16
Server-Side Row Model filtering

adrianspdev:
## Summary

Integrate Filters plugin with DataProvider: hide "Filter by value" when using server-side data, and preserve filter conditions across dataProvider fetches.

### Features

- **Hide "Filter by value" when DataProvider is enabled**  
  When the table uses the DataProvider plugin, the "Filter by value" section and the whole `htFiltersMenuValue` block are hidden in the column filter dropdown. Filtering is done server-side, so the value list is not available.

- **Preserve filter conditions after DataProvider fetch**  
  When DataProvider loads data via `loadData()`, the core re-initializes index mappers and clears the Filters condition collection. The Filters plugin now saves conditions in `beforeLoadData` (when `source === 'dataProvider'`) and restores them in `afterLoadData`, so applied filters no longer disappear after a fetch (e.g. pagination or refetch).

### Code

- **`handsontable/src/plugins/filters/component/value.js`**
  - Import `PLUGIN_KEY as DATA_PROVIDER_PLUGIN_KEY` from the dataProvider plugin.
  - Override `isHidden()` to return `true` when the DataProvider plugin is enabled.

- **`handsontable/src/plugins/filters/filters.js`**
  - Add private `#savedConditionsForDataProvider` to hold conditions when a dataProvider load is in progress.
  - Subscribe to `beforeLoadData` and `afterLoadData`; in handlers, when `source === 'dataProvider'` and the dataProvider plugin is enabled, save conditions before load and restore them after.
  - Clear `#savedConditionsForDataProvider` in `disablePlugin()` to avoid stale state.

- **`handsontable/src/core/hooks/constants.js`**
  - Document `beforeLoadData` and `afterLoadData`: `source` param is optional and can be `'dataProvider'` when the DataProvider plugin loads data after a fetch.

- **`handsontable/types/core/hooks/index.d.ts`**
  - Add short JSDoc on `beforeLoadData` and `afterLoadData` noting that `source` is `'dataProvider'` when the DataProvider plugin loads data.

### Tests

- **`handsontable/src/plugins/filters/__tests__/plugins/dataProvider.spec.js`**
  - **"should hide 'Filter by value' section when dataProvider is enabled"** – open dropdown via API, assert `.htFiltersMenuValue` is not present.
  - **"should preserve filter conditions after dataProvider fetch completes"** – apply filter, change page twice, assert `exportConditions()` is unchanged after each fetch.
  - **"should update pagination total..."** – use `dataProvider.getTotalRows()` instead of `pagination.getPaginationData().totalRows` (totalRows not on getPaginationData).
  - **"should not use server-side filtering when dataProvider is not set"** – use `eq` + `['A1']` so only one row matches (avoids `begins_with ['A1']` matching A1, A10–A19).
  - **"should hide 'Filter by value'..."** – open dropdown with `getPlugin('dropdownMenu').open()`, assert menu root exists and `.htFiltersMenuValue` is null.

## Notes

- No new public API. Filters still use `importConditions` / `exportConditions` with existing `ColumnConditions` type.
- DataProvider continues to call `hot.loadData(rows, PLUGIN_KEY)`; Filters react via the existing `beforeLoadData` / `afterLoadData` hooks and the `source` argument.

[skip changelog]

---

> [!NOTE]
> **Medium Risk**
> Changes filtering behavior when `dataProvider` is enabled by routing filter actions to server-side fetches and resetting pagination to page 1, which can affect existing integrations and UI expectations. Adds state-save/restore around `loadData` and error recovery paths that could impact filter persistence if edge cases are missed.
> 
> **Overview**
> Enables **server-side filtering** when `dataProvider` is active: `Filters#filter()` now translates the current filter model into a `DataProvider` `filters` payload (or `null` when cleared) and triggers `dataProvider.setFilters(...)` instead of doing client-side row trimming.
> 
> Adjusts paging semantics so `DataProvider#setFilters()` **always resets to page 1** (while `setSort()` preserves the current page), and updates the Filters UI by hiding the *“Filter by value”* section when server-side mode is on.
> 
> Improves robustness by preserving/restoring filter conditions across `dataProvider`-driven `loadData` cycles and restoring the previous filter state after `afterDataProviderFetchError`. Types are updated to expose a structured `DataProviderFilters` model, and new/updated tests cover the integration and pagination/reset behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 846461bd73b6bcd97fb73144cbf014315b681838. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/handsontable/src/plugins/filters/component/value.js
+++ b/handsontable/src/plugins/filters/component/value.js
@@ -8,6 +8,7 @@ import { BaseComponent } from './_base';
 import { MultipleSelectUI } from '../ui/multipleSelect';
 import { CONDITION_BY_VALUE, CONDITION_NONE } from '../constants';
 import { getConditionDescriptor } from '../conditionRegisterer';
+import { PLUGIN_KEY as DATA_PROVIDER_PLUGIN_KEY } from '../../dataProvider';
 
 /**
  * @private
@@ -194,6 +195,19 @@ export class ValueComponent extends BaseComponent {
    */
   getMultipleSelectElement() {
     return this.elements.filter(element => element instanceof MultipleSelectUI)[0];
+  }
+
+  /**
+   * Check if component is hidden. The "Filter by value" section is hidden when the dataProvider plugin
+   * is enabled, because filtering is then performed server-side and the value list is not available.
+   *
+   * @returns {boolean}
+   */
+  isHidden() {
+    const dataProvider = this.hot.getPlugin(DATA_PROVIDER_PLUGIN_KEY);
+    const isDataProviderEnabled = dataProvider?.isEnabled?.() === true;
+
+    return super.isHidden() || isDataProviderEnabled;
   }
 
   /**
--- a/handsontable/src/plugins/filters/component/value.js
+++ b/handsontable/src/plugins/filters/component/value.js
@@ -204,10 +204,14 @@ export class ValueComponent extends BaseComponent {
    * @returns {boolean}
    */
   isHidden() {
+    if (super.isHidden()) {
+      return true;
+    }
+
     const dataProvider = this.hot.getPlugin(DATA_PROVIDER_PLUGIN_KEY);
     const isDataProviderEnabled = dataProvider?.isEnabled?.() === true;
 
-    return super.isHidden() || isDataProviderEnabled;
+    return isDataProviderEnabled;
   }
 
   /**
--- a/handsontable/src/plugins/filters/filters.js
+++ b/handsontable/src/plugins/filters/filters.js
@@ -26,6 +26,7 @@ import {
   OPERATION_OR_THEN_VARIABLE
 } from './constants';
 import { TrimmingMap } from '../../translations';
+import { PLUGIN_KEY as DATA_PROVIDER_PLUGIN_KEY } from '../dataProvider';
 
 export const PLUGIN_KEY = 'filters';
 export const PLUGIN_PRIORITY = 250;
@@ -171,6 +172,13 @@ export class Filters extends BasePlugin {
    * @type {Array}
    */
   #previousConditionStack = [];
+  /**
+   * When loadData is called by the dataProvider plugin, initIndexMappers clears the condition
+   * collection. We save conditions in beforeLoadData and restore them in afterLoadData.
+   *
+   * @type {Array|null}
+   */
+  #savedConditionsForDataProvider = null;
 
   constructor(hotInstance) {
     super(hotInstance);
@@ -283,6 +291,8 @@ export class Filters extends BasePlugin {
     this.addHook('afterDropdownMenuShow', () => this.#onAfterDropdownMenuShow());
     this.addHook('afterDropdownMenuHide', () => this.#onAfterDropdownMenuHide());
     this.addHook('afterChange', changes => this.#onAfterChange(changes));
+    this.addHook('beforeLoadData', (data, firstRun, source) => this.#onBeforeLoadData(source));
+    this.addHook('afterLoadData', (data, firstRun, source) => this.#onAfterLoadData(source));
 
     // Temp. solution (extending menu items bug in contextMenu/dropdownMenu)
     if (this.hot.getSettings().dropdownMenu && this.dropdownMenuPlugin) {
@@ -365,6 +375,7 @@ export class Filters extends BasePlugin {
       this.conditionCollection.destroy();
       this.conditionCollection = null;
       this.hot.rowIndexMapper.unregisterMap(this.pluginName);
+      this.#savedConditionsForDataProvider = null;
     }
 
     this.unregisterShortcuts();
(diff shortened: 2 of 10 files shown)


qunabu:
Task linked: [PRO-1066 Server-Side Row Model (filtering)](https://app.clickup.com/t/9015210959/PRO-1066)
Task linked: [PRO-1064 Server-Side Row Model (dataProvider + pagination)](https://app.clickup.com/t/9015210959/PRO-1064)

handsontable/src/plugins/filters/component/value.js
```diff
@@ -196,6 +197,23 @@ export class ValueComponent extends BaseComponent {
...
+   * is enabled, because filtering is then performed server-side and the value list is not available.
+   *
+   * @returns {boolean}
+   */
+  isHidden() {
+    if (super.isHidden()) {
+      return true;
+    }
+
+    const dataProvider = this.hot.getPlugin(DATA_PROVIDER_PLUGIN_KEY);
```
budnix, line 211: Right now the component decides its visibility by checking another plugin directly. That creates extra coupling between pieces that should stay more independent. The component should mostly render and manage its own UI, while higher-level logic should decide when it is shown or hidden.

I would move this decision to the parent/orchestrating layer and pass it into the component as a callback or flag source. Same behavior, less cross-plugin knowledge, easier maintenance later.

handsontable/src/plugins/filters/filters.js
```diff
@@ -26,6 +26,7 @@ import {
   OPERATION_OR_THEN_VARIABLE
 } from './constants';
 import { TrimmingMap } from '../../translations';
+import { PLUGIN_KEY as DATA_PROVIDER_PLUGIN_KEY } from '../dataProvider';
```
budnix, line 29: It creates a hard dependency from filters to dataProvider, so the server-side fetching module can end up in the bundle even for users who do not use that feature. This is avoidable code bloat.

If this integration is needed, it should be done through a looser boundary (for example capability checks or injected behavior), so filters does not require direct module-level coupling to dataProvider.

handsontable/src/plugins/filters/filters.js
```diff
@@ -675,6 +696,43 @@ export class Filters extends BasePlugin {
...
+        : [];
+      dataProvider.setFilters(filtersForProvider).catch(() => {});
+
+      this.#previousConditionStack = this.exportConditions();
+
+      if (!needToFilter) {
+        this.filtersRowsMap.clear();
+      }
+
+      if (this.hot.selection.isSelected()) {
```
budnix, line 722: This code is a copy/paste from the line 780. I think it would be nice the move it to a separate method. WDYT?

handsontable/src/plugins/filters/filters.js
```diff
@@ -808,6 +866,72 @@ export class Filters extends BasePlugin {
     }
   }
 
+  /**
+   * Before loadData: when the source is the dataProvider plugin, save current filter conditions
+   * so they can be restored after loadData (initIndexMappers clears the condition collection).
+   *
+   * @private
```
budnix, line 873: Not necessary
```suggestion
```

handsontable/src/plugins/filters/filters.js
```diff
@@ -808,6 +866,72 @@ export class Filters extends BasePlugin {
...
+   * @param {string} [source] The source of the loadData call.
+   */
+  #onBeforeLoadData(source) {
+    if (source !== DATA_PROVIDER_PLUGIN_KEY) {
+      return;
+    }
+
+    const dataProvider = this.hot.getPlugin(DATA_PROVIDER_PLUGIN_KEY);
```
budnix, line 881: The Filters plugin now:
- Directly calls `dataProvider.setFilters()`
- Subscribes to `afterDataProviderFetch` and `afterDataProviderFetchError`
- Performs physical-to-visual column mapping for the DataProvider
- Manages its own condition backup/restore around DataProvider's lifecycle

This creates tight bidirectional coupling. The DataProvider knows about filters (via `setFilters`), and the Filters plugin knows about DataProvider's lifecycle hooks. Consider a more decoupled approach:

**Suggestion**: Have the Filters plugin simply fire a hook (`beforeFilter`). The DataProvider can listen for this hook and call its own `setFilters()`. This follows the existing pattern where ColumnSorting fires `beforeColumnSort` and DataProvider intercepts it.

This is only a suggestion - I'm not sure whether it's doable within the project's timeframe.

handsontable/src/plugins/filters/filters.js
```diff
@@ -808,6 +866,72 @@ export class Filters extends BasePlugin {
...
+    if (!this.conditionCollection.isEmpty()) {
+      this.#savedConditionsForDataProvider = this.exportConditions();
+    }
+  }
+
+  /**
+   * After loadData: when the source is the dataProvider plugin, restore filter conditions
+   * that were cleared by initIndexMappers.
+   *
+   * @private
```
budnix, line 896:
```suggestion
```

handsontable/src/plugins/filters/filters.js
```diff
@@ -675,6 +696,43 @@ export class Filters extends BasePlugin {
...
+          column: this.hot.toVisualColumn(column),
+          operation,
+          conditions: conds.map(({ name, args }) => ({ name, args: [...args] })),
+        }))
+        : null;
+
+      this.#conditionsBeforeDataProviderFetch = this.#previousConditionStack.length > 0
+        ? this.#previousConditionStack.slice()
+        : [];
+      dataProvider.setFilters(filtersForProvider).catch(() => {});
```
budnix, line 714: I'd add a comment that inform the developer that the error is handled in different place.
```suggestion
      dataProvider.setFilters(filtersForProvider).catch((err) => {
        // Already handled via afterDataProviderFetchError hook
      });
```

handsontable/types/plugins/dataProvider/dataProvider.d.ts
```diff
@@ -2,14 +2,51 @@ import Core from '../../core';
 import { BasePlugin } from '../base';
 import { CellValue } from '../../common';
 
+/**
+ * Filter condition operators supported in queryParameters.filters when using server-side filtering.
+ */
+export type FilterConditionOperator =
```
budnix, line 8: The types related to the Filters should be taken from the Filters plugin to mitigate drift over time

budnix (changes requested):

budnix:
I found that both filtering and sorting currently send visual column indexes to the server. This breaks when columns are moved. The server maps to the wrong index.

```js
  // filters
  { field: 'city', operation: 'conjunction', conditions: [...] }
  // sort
  { field: 'city', sortOrder: 'asc' }
```

I think that the DataProvider should only support array-of-objects data sources (where columns has data properties). For array-of-arrays there are no stable field names, so we should either block it or require explicit column identifiers.

adrianspdev:
Will be implemented directly in https://github.com/handsontable/handsontable/pull/12147

PR #12166, opened 2026-03-17
Implement Excel export for the `exportFile` plugin.

jansiegel:
### Context                                                                                                                                                                                                                    
This PR adds XLSX export support to the `ExportFile` plugin, built on top of the [ExcelJS](https://github.com/exceljs/exceljs) library. ExcelJS is injected by the user as an `engine` option — no new hard dependency is added
   to the package.                                                                                                                                                                                                             
                                                                                                                                                                                                                                 
  **XLSX export** (`exportFile: { engine: ExcelJS }`)                                                                                                                                                                            
  - Exports data, column/row headers, merged cells, frozen panes, and RTL layout                                                                                                                                                 
  - Exports cell types natively: `numeric` (with `numFmt`), `date`, `intl-date`, `time`, `checkbox`, `dropdown`/`autocomplete` (list validation), `multiselect`                                                                  
  - Exports cell styling from CSS classes: alignment (`htLeft`/`htCenter`/…), font (bold, italic, underline, color), background fill, custom borders                                                                             
  - Exports read-only cells with worksheet protection (non-read-only cells are unlocked)                                                                                                                                         
  - Exports HyperFormula formula cells and ColumnSummary destinations as live Excel formulas when `exportFormulas: true`                                                                                                         
  - Multi-sheet export via a `sheets` array option (each entry points to a separate Handsontable instance)                                                                                                                       
  - Conditional formatting via a `conditionalFormatting` option                                                                                                                                                                  
  - Configurable ZIP compression via `compression` option                                                                                                                                                                        
  - Configurable header styling via `headerStyle` option                                                                                                                                                                         
                              
  **Context menu integration** (automatic when `ExportFile` is enabled)                                                                                                                                                          
  - Adds "Export to CSV" and "Export to Excel" items to the context menu
  - "Export to Excel" is hidden when no ExcelJS engine is configured                                                                                                                                                             
  - While the export runs, shows a loading overlay using the existing `ht-loading` CSS classes
                                                                                                                                                                                                                                 
  **New `exportFile` plugin-level configuration options:**

  | Option | Type | Default | Description |                                                                                                                                                                                      
  |---|---|---|---|
  | `engine` | `object` | `null` | ExcelJS module reference (required for XLSX) |                                                                                                                                                
                              
  **New XLSX-specific export options** (passed to `exportAsBlobAsync`, `downloadFileAsync`, or `_createTypeFormatter`):                                                                                                          
   
  | Option | Type | Default | Description |                                                                                                                                                                                      
  |---|---|---|---|           
  | `engine` | `object` | — | Per-call ExcelJS override |
  | `compression` | `number\|null` | `null` | ZIP compression level (1–9) |                                                                                                                                                      
  | `exportFormulas` | `boolean` | `false` | Export HyperFormula / ColumnSummary cells as live formulas |
  | `exportHiddenRows` | `boolean\|'hide'` | `false` | `false` omits hidden rows; `true` exports them as visible; `'hide'` exports them and marks them as hidden in Excel |                                                      
  | `exportHiddenColumns` | `boolean\|'hide'` | `false` | `false` omits hidden columns; `true` exports them as visible; `'hide'` exports them and marks them as hidden in Excel |
  | `headerStyle` | `object\|null` | `{ backgroundColor: '#F2F2F2', border: { style: 'thin' } }` | Style applied to header cells; `null` disables all header styling |                                                           
  | `conditionalFormatting` | `Array` | `[]` | Array of `{ rows?, cols?, rules }` CF descriptors |                                                                                                                               
  | `sheets` | `Array` | — | Multi-sheet mode: array of `{ instance, name?, ...options }` |                                                                                                                                      
                                                                                                                                                                                                                                 
  **New i18n constants:** `CONTEXTMENU_ITEMS_EXPORT_FILE_CSV`, `CONTEXTMENU_ITEMS_EXPORT_FILE_XLSX` (translations added for all supported locales).                                                                              
                              
  **Sync/async API split** — mirroring the `downloadFile` / `downloadFileAsync` pattern:                                                                                                                                         
                              
  | Method | Formats | Returns |                                                                                                                                                                                                 
  |---|---|---|               
  | `exportAsString` | CSV only (throws for binary) | `string` |
  | `exportAsBlob` | CSV only (throws for binary) | `Blob` |
  | `exportAsBlobAsync` | All formats | `Promise<Blob>` |
  | `downloadFile` | CSV only (throws for binary) | `void` |                                                                                                                                                                     
  | `downloadFileAsync` | All formats | `Promise<void>` |
                                                                                                                                                                                                                                 
  **`exportAsString('xlsx')` throws** — XLSX is a binary format and cannot be serialized to a plain string.                                                                                                                      
   
  **`exportAsBlob('xlsx')` throws** — use `exportAsBlobAsync('xlsx')` instead.                                                                                                                                                   
                              
  **`downloadFile('xlsx')` throws** — use `downloadFileAsync('xlsx')` instead.                                                                                                                                                   
   
  > [!WARNING]                                                                                                                                                                                                                   
  > **3 known limitations are tracked as skipped (`xit`) tests and will be addressed in follow-up work.**
  >
  > **1. Currency cells show "Custom" instead of "Currency" in Excel for Mac** (`cell-types.spec.js`)\
  > ExcelJS omits OOXML built-in format IDs 5–8 (the dollar-currency built-ins) from its internal format table, so any currency format code is assigned a custom `numFmtId ≥ 164`. Excel for Mac shows "Custom" for all custom   
  IDs; Excel for Windows analyses the format string and shows "Currency" correctly. The fix requires post-processing the generated OOXML ZIP to replace the custom ID with the correct built-in ID (e.g. `7` for `$#,##0.00`)    
  before the file is saved.                                                                                                                                                                                                      
  >                                                                                                                                                                                                                              
  > **2. Time cells show "Custom" instead of "Time" in Excel for Mac** (`cell-types.spec.js`)\
  > ExcelJS correctly writes the built-in OOXML `numFmtId=21` (`h:mm:ss`) and Excel for Windows categorises the cell as "Time". However, Excel for Mac still displays "Custom" for the same file. This appears to be a Mac Excel
  quirk with spec-compliant OOXML. The test is disabled until either a workaround format code is found or a way to programmatically assert the UI category from OOXML is established.                                            
  >
  > **3. `ColumnSummary` cells with `type: 'custom'` always export as static values** (`formulas.spec.js`)\                                                                                                                      
  > Custom summary types use arbitrary JavaScript functions that have no generic Excel formula equivalent. When `exportFormulas: true` is set, these cells fall back to their pre-calculated static value instead of a live      
  formula. This is a by-design limitation for now — the test is skipped to document the gap rather than assert currently unachievable behaviour.                                                                                 
                                                                                                                                                                                                                                 
  ### How has this been tested?                                                                                                                                                                                                  
  Added comprehensive E2E test suites covering headers, layout, formulas, cell types, cell styling, features (comments, CF, multi-sheet, compression), API, and context menu. Added unit tests for `numbroPatternToExcelNumFmt`.
  Extended the layout suite with cases for `exportHiddenRows: 'hide'` and `exportHiddenColumns: 'hide'` (data presence, hidden flags, width/height preservation). Extended the headers suite with cases for nested header        
  spanning across hidden columns under both `'hide'` and `false` modes.
                                                                                                                                                                                                                                 
  ### Types of changes                                                  

  ### Related issue(s):
  1. https://github.com/handsontable/dev-handsontable/issues/2650
                                                                                                                                                                                                                                 
  ### Affected project(s):
  - [x] `handsontable`                                                                                                                                                                                                           
                                                                        
  ### Checklist:



<!-- CURSOR_SUMMARY -->
---

> [!NOTE]
> **Medium Risk**
> Extends the core `ExportFile` plugin with a new binary XLSX code path, async download/blob APIs, and context-menu/i18n behavior, which could affect existing export flows. Large surface-area change but gated behind providing an `xlsx` engine and backed by extensive new tests and docs.
> 
> **Overview**
> Adds **XLSX export** to the `ExportFile` plugin via a user-provided `ExcelJS` engine (`exportFile: { engines: { xlsx: ExcelJS } }`), including a new `supportsExportFormat('xlsx')` gate and updated type-factory option merging.
> 
> Introduces an explicit **sync vs async API split**: CSV remains synchronous (`exportAsString`, `exportAsBlob`, `downloadFile`), while binary formats (XLSX) are handled through `exportAsBlobAsync`/`downloadFileAsync` with a loading dialog overlay for long-running exports.
> 
> Updates the **context menu** to add an `Export` submenu with `To CSV` and conditional `To Excel` entries, adds new i18n keys/translations, and expands docs (new *Export to Excel* guide with JS/React/Angular examples) plus changelog/Docs tooling updates (ExcelJS in the docs presets). Also adds `exceljs` to dev deps/babel allowlist and a comprehensive new test suite covering XLSX types, styling, options, and menu/i18n behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e5487d8078899f1f4b5cb7a7068cf536cd4c2689. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>
<!-- /CURSOR_SUMMARY -->


qunabu (dismissed):

jansiegel:
@budnix @Wojciech-Swiderski :

Demos of the feature: https://stackblitz.com/edit/vitejs-vite-bv9biazq

docs/content/guides/accessories-and-menus/export-to-excel/export-to-excel.md
```diff
@@ -0,0 +1,305 @@
...
+Pass the ExcelJS constructor to the `engine` option in the `exportFile` plugin configuration:
+
+::: only-for javascript
+
+```js
+import ExcelJS from 'exceljs';
+
+const hot = new Handsontable(container, {
+  exportFile: { engine: ExcelJS },
+  licenseKey: 'your-license-key',
```
budnix: I think we can skip the license key part here. It's already explained on another page.
```suggestion
```

docs/content/guides/accessories-and-menus/export-to-excel/export-to-excel.md
```diff
@@ -0,0 +1,305 @@
...
+:::
+
+::: only-for react
+
+```jsx
+import ExcelJS from 'exceljs';
+
+<HotTable
+  exportFile={{ engine: ExcelJS }}
+  licenseKey="your-license-key"
```
budnix:
```suggestion
```

handsontable/src/plugins/exportFile/__tests__/numeric-utils.unit.js
```diff
@@ -0,0 +1,34 @@
+import { numbroPatternToExcelNumFmt } from '../types/xlsx/numeric-utils';
+
+describe('numbroPatternToExcelNumFmt', () => {
+  it('should convert a currency pattern with thousands grouping', () => {
+    expect(numbroPatternToExcelNumFmt('$0,0.00')).toBe('$#,##0.00');
```
budnix, line 5: What about support for Intl.DateTime format (a new `intl-date` and `intl-time` cell support) and Intl.Numeric format?

handsontable/package.json
```diff
@@ -93,6 +93,7 @@
     "numbro": "2.5.0"
   },
   "optionalDependencies": {
+    "exceljs": "^4.4.0",
```
budnix: It should be moved to `deveDependencies`. All dependencies will be dropped in the next major release.

handsontable/src/plugins/exportFile/dataProvider.js
```diff
@@ -124,6 +139,704 @@ class DataProvider {
...
+   * Returns values in the same column order as {@link DataProvider#getData}.
+   *
+   * @returns {Array}
+   */
+  getColumnsWidths() {
+    const { startCol, endCol } = this._getDataRange();
+    const options = this.options;
+    const widths = [];
+
+    rangeEach(startCol, endCol, (colIndex) => {
```
budnix: I would slowly give up this helper in favor of the native loop. WDYT?

handsontable/src/plugins/exportFile/exportFile.js
```diff
@@ -1,41 +1,80 @@
...
+
+/**
+ * Builds the dialog overlay HTML for the export progress indicator.
+ *
+ * The title text is resolved at call-time so it reflects the active locale.
+ *
+ * @param {string} title Translated title string (e.g. "Exporting…").
+ * @returns {string}
+ */
+function buildExportDialogContent(title) {
```
budnix: Wouldn't it be better to use the internal `html` helper? I think the code with this helper may be easier to maintain.

handsontable/src/plugins/exportFile/exportFile.js
```diff
@@ -160,49 +142,99 @@ export class ExportFile extends BasePlugin {
...
+  /**
+   * Enables the plugin functionality for this Handsontable instance.
+   */
+  enablePlugin() {
+    if (this.enabled) {
+      return;
+    }
+
+    const pluginSettings = this.hot.getSettings()[PLUGIN_KEY];
+    const contextMenuEnabled = isObject(pluginSettings) && pluginSettings.contextMenu === true;
```
budnix: It's better to use the `isEnabled` method of the plugin instead of recreating the same logic as in the plugin.

handsontable/src/plugins/exportFile/exportFile.js
```diff
@@ -211,95 +243,149 @@ export class ExportFile extends BasePlugin {
...
+        a.dispatchEvent(new MouseEvent('click'));
+        rootDocument.body.removeChild(a);
+
+        setTimeout(() => {
+          URL.revokeObjectURL(url);
+        }, 100);
 
-    if (a.download !== undefined) {
-      const url = URL.createObjectURL(blob);
+      } else if (navigator.msSaveOrOpenBlob) { // IE10+
```
budnix: We don't support IE10 🤣 I guess that branch can be removed

budnix (changes requested):

handsontable/types/plugins/exportFile/exportFile.d.ts
```diff
@@ -12,17 +30,30 @@ export interface ExportOptions {
...
+export type Settings = ExportFileSettings;
 
 export class ExportFile extends BasePlugin {
   constructor(hotInstance: Core);
   isEnabled(): boolean;
   exportAsString(format: Formats, options?: ExportOptions): string;
-  exportAsBlob(format: Formats, options?: ExportOptions): Blob;
-  downloadFile(format: Formats, options?: ExportOptions): void;
+  exportAsBlob(format: Formats, options?: ExportOptions): Blob | Promise<Blob>;
+  downloadFile(format: Formats, options?: ExportOptions): void | Promise<void>;
```
budnix: Both methods should return always Promises for consistency 🤔

jansiegel: This would be a breaking change

handsontable/package.json
jansiegel: Do we have a way to inform the user which version of dependencies they're supposed to be using?

budnix: I guess. Maybe it is worth checking if there is something like `ExcelJS.version` to compare with plugin expectations. If they are not met, then print some warning, maybe.

qunabu: It should be checked on runtime, like with HyperFormula in future. 

`exceljs` should not be in package.json 

I wish npm would implement something like suggest https://getcomposer.org/doc/04-schema.md#suggest in composer php

jansiegel: @qunabu 
> It should be checked on runtime

I don't think that's possible - it looks like ExcelJS does not expose the version anywhere in the exported object.

handsontable/src/plugins/exportFile/exportFile.js
```diff
@@ -1,41 +1,88 @@
...
 import typeFactory, { EXPORT_TYPES } from './typeFactory';
+import exportCsvItem from './contextMenuItem/exportCsv';
+import exportExcelItem from './contextMenuItem/exportExcel';
 
 export const PLUGIN_KEY = 'exportFile';
 export const PLUGIN_PRIORITY = 240;
 
+// CSS class prefix shared with the Loading plugin (defined in handsontable.css).
+// Inlined here to avoid a cross-plugin import — the token is stable and always bundled.
+const LOADING_CLASS = 'ht-loading';
```
budnix: Maybe it's worth moving the `LOADING_CLASS_NAME` to `handsontable/src/helpers/constants.js`? Then it can be used in both plugins without worry. WDYT?

handsontable/src/plugins/exportFile/types/xlsx/cell-style.js
```diff
@@ -0,0 +1,537 @@
...
+  const isDropdownType = meta.type === 'dropdown' || meta.type === 'autocomplete';
+
+  if (!isDropdownType || !Array.isArray(meta.source)) {
+    return null;
+  }
+
+  return {
+    type: 'list',
+    allowBlank: true,
+    formulae: [`"${meta.source.join(',')}"`],
```
budnix: When the cell value contains `"`, Excel reports that the sheet is corrupted.

handsontable/src/plugins/exportFile/exportFile.js
```diff
@@ -160,49 +147,99 @@ export class ExportFile extends BasePlugin {
...
+    if (this.enabled) {
+      return;
+    }
+
+    const pluginSettings = this.hot.getSettings()[PLUGIN_KEY];
+
+    if (isObject(pluginSettings) && pluginSettings.contextMenu === true &&
+        this.hot.getPlugin('contextMenu').isEnabled()) {
+      this.addHook('afterContextMenuDefaultOptions', options => this.#onAfterContextMenuDefaultOptions(options));
+    }
```
budnix: The condition below is arguably redundant. The hook system is event-driven. If the hook is registered, it is called when the event fires; if the context menu is never shown, the handler is never invoked. The check could be removed, and the hook always registered.
```suggestion
  this.addHook('afterContextMenuDefaultOptions', options => this.#onAfterContextMenuDefaultOptions(options));
```

handsontable/src/plugins/exportFile/types/xlsx.js
```diff
@@ -0,0 +1,925 @@
...
+  parseIsoStringToSerial,
+  parseTimeStringToSerial,
+  getDateNumFmt,
+  getTimeNumFmt,
+} from './xlsx/date-utils';
+import { intlNumFormatToExcelNumFmt } from './xlsx/numeric-utils';
+
+const PIXELS_PER_EXCEL_COLUMN_WIDTH_UNIT = 7;
+const PIXELS_TO_POINTS_RATIO = 0.75;
+const ROW_HEADER_DEFAULT_WIDTH = 5;
```
budnix, line 43: It would be nice to have some description of what those cons do

handsontable/src/plugins/exportFile/types/xlsx.js
```diff
@@ -0,0 +1,925 @@
...
+      );
+    }
+
+    clearStyleCaches(this.dataProvider.hot.rootDocument);
+
+    const workbook = new engine.Workbook();
+    const { sheets } = this.options;
+
+    if (sheets && sheets.length > 0) {
+      // ── multi-sheet mode ─────────────────────────────────────────────────
```
budnix:
```suggestion
      // multi-sheet mode

File: handsontable/src/plugins/exportFile/types/xlsx.js [hunk 15]
```diff
@@ -0,0 +1,925 @@
...
+        }
+
+        usedSheetNames.add(name);
+
+        const worksheet = workbook.addWorksheet(name);
+
+        this.#populateWorksheet(worksheet, dp, sheetOptions);
+      });
+    } else {
+      // ── single-sheet mode ────────────────────────────────────────────────
```
budnix:
```suggestion
      // single-sheet mode

File: handsontable/src/plugins/exportFile/types/xlsx/cell-style.js [hunk 16]
```diff
@@ -0,0 +1,537 @@
...
+ * @private
+ * @param {object|undefined} meta Cell meta object.
+ * @returns {object|null}
+ */
+export function getAlignmentFromMeta(meta) {
+  if (!meta?.className) {
+    return null;
+  }
+
+  const classes = meta.className.split(' ');
```
budnix: `className` can be passed as an array; if so, a crash occurs.

budnix: I think the other part of the logic may also be affected by this bug, as there are more assumptions that the `className` is a string or null/undefined.

handsontable/src/plugins/exportFile/types/xlsx.js
```diff
@@ -0,0 +1,925 @@
...
+   * @param {boolean} hasReadOnlyCells Whether the sheet has any read-only cells.
+   * @param {Document} rootDocument Owner document (fallback for off-viewport cells).
+   * @param {Window} rootWindow Owner window (fallback for off-viewport cells).
+   */
+  #writeRowCells(row, rowData, rowIndex, cellsMeta, cellElements,
+                 summaryMap, sourceData, cellContext, hasReadOnlyCells, rootDocument, rootWindow) {
+    for (let colIndex = 0; colIndex < rowData.length; colIndex++) {
+      const cellValue = rowData[colIndex];
+      const cell = row.getCell(colIndex + cellContext.dataColOffset);
+      const meta = cellsMeta[rowIndex]?.[colIndex];
```
budnix: `getCellsMeta()` uses the same `_extractDataMatrix()` as `getData()`, so both 2D arrays are guaranteed to have identical dimensions. And `getCellMeta()` always returns an object, never `undefined`.

The current defensive `meta?.` pattern (and there is a lot of that) has two problems:
  1. It hides the actual contract - a reader can't tell whether meta being `undefined` is a real case or just cargo-cult safety.
  2. AI coding assistants and future contributors will copy this pattern, spreading unnecessary optional chaining across the codebase.
  3. Misleads anyone reading the code. When they see optional chaining, they assume the value can legitimately be null - so they add even more defensive checks elsewhere. It spreads.

handsontable/src/plugins/exportFile/contextMenuItem/exportExcel.js
```diff
@@ -0,0 +1,37 @@
...
+ * Exports the active selection when one exists and was made on data cells.
+ * Falls back to exporting the entire table when there is no selection or
+ * when the context menu was triggered from a header or corner element.
+ *
+ * @param {ExportFile} exportFilePlugin The plugin instance.
+ * @returns {object}
+ */
+export default function exportExcelItem(exportFilePlugin) {
+  return {
+    key: 'export_file_xlsx',
```
budnix: It would be nice to update the context menu keys here: https://handsontable.com/docs/javascript-data-grid/context-menu/.

handsontable/src/plugins/exportFile/contextMenuItem/exportExcel.js
```diff
@@ -0,0 +1,37 @@
...
+    },
+    callback() {
+      exportFilePlugin.downloadFileAsync('xlsx', getExportOptions(this)).catch((err) => {
+        error('ExportFile: XLSX export failed.', err);
+      });
+    },
+    hidden() {
+      const settings = this.getSettings()[PLUGIN_KEY];
+
+      return !(isObject(settings) && settings.engine);
```
budnix: Can I suggest adding a new method to the plugin called `supportsExportFormat('xlsx' | 'pdf' | 'csv')` or similar? To keep all plugins' logic related in one place.

handsontable/src/plugins/exportFile/dataProvider.js
```diff
@@ -31,34 +30,48 @@ class DataProvider {
...
+  _extractDataMatrix(getCellValue) {
     const { startRow, startCol, endRow, endCol } = this._getDataRange();
     const options = this.options;
     const data = [];
 
-    rangeEach(startRow, endRow, (rowIndex) => {
-      const row = [];
-
+    for (let rowIndex = startRow; rowIndex <= endRow; rowIndex++) {
       if (!options.exportHiddenRows && this._isHiddenRow(rowIndex)) {
```
budnix: The options `exportHiddenRows` and `exportHiddenColumns` also accept string values as `"hide"` - looking at the types. In that case the logic would include the hidden indexes.

budnix (changes requested):

handsontable/src/plugins/exportFile/exportFile.js
```diff
@@ -1,41 +1,85 @@
...
+ *
+ * The title text is resolved at call-time so it reflects the active locale.
+ *
+ * @param {string} title Translated title string (e.g. "Exporting…").
+ * @returns {DocumentFragment}
+ */
+function buildExportDialogContent(title) {
+  // Spinner SVG reused from the Loading plugin — same arc shape, same CSS class so the
+  // `ht-loading__icon-svg` spin animation (defined in handsontable.css) applies automatically.
+  const { fragment } = html`
```
budnix: Can I suggest moving that function to `utils.js` file of the plugin? To keep plugins' logic as simple as possible.

handsontable/src/plugins/exportFile/exportFile.js
```diff
@@ -160,146 +177,333 @@ export class ExportFile extends BasePlugin {
     return true;
   }
 
+  /**
```
budnix, line 155: I think it would be worth adding to the plugin the `SETTING_KEYS` key. Right now the plugin is disabled and enabled on each `updateSettings` call.

Suggestion:
```js
SETTING_KEYS() { return [PLUGIN_KEY]; }
```

handsontable/src/plugins/exportFile/exportFile.js
```diff
@@ -160,146 +177,333 @@ export class ExportFile extends BasePlugin {
...
+   *
+   * @param {object} options Contains default added options of the Context Menu.
+   */
+  #onAfterContextMenuDefaultOptions(options) {
+    const pluginSettings = this.hot.getSettings()[PLUGIN_KEY];
+
+    if (!isObject(pluginSettings) || pluginSettings.contextMenu !== true) {
+      return;
+    }
+
```
budnix:
```suggestion
```

handsontable/types/plugins/exportFile/exportFile.d.ts
```diff
@@ -12,17 +40,42 @@ export interface ExportOptions {
...
+   * ExcelJS library instance. Can be supplied per-call to override the engine set in plugin settings (XLSX only).
+   */
+  engine?: object;
 }
-export type Settings = boolean;
+
+export interface ExportFileSettings {
+  engine?: object;
+  contextMenu?: boolean;
+}
```
budnix, line 68: The `engine` options seem to be too closed for extensibility. What if we would like to add support for export to PDF? I would change that object shape to something like:
```js
engines: {
  xlsx: DepsXLSX,
  pdf: DepsPDF,
};
```
or 
```js
columnHeaders: true,
rowHeaders: true,
xlsx: {
  engine: DepsXLSX,
  headerStyle: ...,
  compression: 3,
  // ... other xlsx-only related options
},
pdf: {
  engine: DepsPDF,
  // ... other pdf-only related options
},
```

handsontable/src/plugins/exportFile/types/xlsx.js
```diff
@@ -0,0 +1,1007 @@
...
+    // so the lock protects the formula from accidental overwriting.
+    //
+    // Setting protection on every cell (even { locked: false }) causes ExcelJS to
+    // initialise font/fill/border to sentinel values on those cells, which would
+    // break assertions in tests and add noise to the exported file.
+    // When a ColumnSummary plugin is active, its destination cells (and any
+    // accompanying label cells the user marks readOnly) are readOnly only to prevent
+    // in-grid editing. Whether exported as static values or live formulas, locking
+    // those cells in Excel adds no value and surprises users, so protection is
+    // suppressed entirely whenever ColumnSummary is present.
```
budnix: Can I ask to revise the comment block? It seems to be giving the same information twice (leftovers, conflicts?).

budnix (approved):
Looks good 👌

PR #12313, opened 2026-04-13
E2E data driven test cases

adrianspdev:
### Context

E2E tests previously embedded theme-specific numbers or duplicated layout logic, which made the theme matrix fragile whenever tokens or density changed. This change centralizes regression geometry in `handsontable/test/helpers/themeLayoutFromTokens.js` (auto-discovering themes from `src/themes/theme/index.js`), trims redundant helpers (e.g. `custom-matchers.js` removals), and aligns specs with the theme workflow so new themes require minimal or no per-spec edits.

### How has this been tested?

- Ran the E2E suite for the affected areas during development (editors, context/dropdown menus, copy/paste, pagination, merge cells, nested headers, column sorting, manual resize, etc.).
- Suggested verification on a clean tree targeting `develop`:

```bash
```

Optionally narrow with `--testPathPattern` to a specific plugin or folder if you need a faster signal.

### Types of changes

### Related issue(s):

1. https://app.clickup.com/t/9015210959/PRO-858

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

[skip changelog]

---

> [!NOTE]
> **Medium Risk**
> Touches a large surface area of tests and test tooling; risk is mainly CI churn/flakiness from changed geometry assertions rather than production behavior.
> 
> **Overview**
> Refactors Handsontable E2E/editor and viewport-scroll specs to be **theme-agnostic**, replacing per-theme hardcoded pixel assertions (and `forThemes` branches) with values derived from `getThemeLayout()`, viewport helpers (e.g. `containerHeightForRows`), live DOM measurements, and tolerant/relational assertions.
> 
> Introduces/extends shared test infrastructure for token-driven layout expectations: adds `themeLayoutFromTokens` to the E2E build allow-list, expands Jest roots to include `test/`, adds a new `baseEditor` helper (`editedCellRect.js`) to assert `getEditedCellRect()` using token-derived partials plus live TD width/height, and updates documentation/skills to codify the new data-driven theme assertion workflow and theme auto-discovery rules.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3130c55c69937ec70ad1a6dd45bf3bd5ff85cf3d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [PRO-858 Theme API. Figma workflow update. e2e test data driven for each theme](https://app.clickup.com/t/9015210959/PRO-858)

.claude/skills/handsontable-e2e-testing/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: handsontable-e2e-testing
-description: Use when writing or modifying Jasmine/Puppeteer E2E tests (*.spec.js) for Handsontable, or when a bug fix or feature change needs E2E test coverage - covers the standard boilerplate, global test helpers, async/await requirements, mouse and keyboard event simulation, and plugin lifecycle testing patterns
+description: Use when writing or modifying Jasmine/Puppeteer E2E tests (*.spec.js) for Handsontable, or when a bug fix or feature change needs E2E test coverage - covers the standard boilerplate, global test helpers, async/await requirements, mouse and keyboard event simulation, plugin lifecycle testing patterns, and token-driven theme-specific expectations via getThemeLayout() (viewport helpers, DOM measurements, and relational assertions that work across all themes)
```
budnix: The description says "token-driven theme-specific expectations" which reads as if tests should be theme-specific -- the opposite of what this section teaches. Suggest tightening to make "theme-agnostic" the explicit goal and drop the nested parenthetical list.

```suggestion
description: Use when writing or modifying Jasmine/Puppeteer E2E tests (*.spec.js) for Handsontable, or when a bug fix or feature change needs E2E test coverage. Covers standard boilerplate, async/await rules, global helpers, event simulation, plugin lifecycle patterns, and writing theme-agnostic assertions that pass under all themes without branching on theme name.
```

.claude/skills/handsontable-e2e-testing/SKILL.md
```diff
@@ -47,8 +47,41 @@ These are injected automatically. Do not import them manually.
 - **Selection:** `selectCell()`, `selectCells()`, `getSelected()`
 - **DOM:** `getCell()`, `spec()`, `hot()`
 - **Plugins:** `getPlugin()`
+- **Theme layout:** `getLoadedTheme()`, `getThemeLayout()` (see `.ai/TESTING.md`)
+- **Iframe `doc.write` theme CSS:** `getE2eThemeStylesheetLinkTagsHtml()` (all themes), `getE2eThemeStylesheetLinkTagHtml(key)`, `getE2eNormalizeStylesheetLinkTagHtml()` -- from `common.js`; theme list is `E2E_REGISTERED_THEME_KEYS` in `themeLayoutFromTokens.js`, auto-discovered from `src/themes/theme/index.js` (add a theme there and the list updates automatically).
 - Full list in `test/helpers/common.js`.
 
+## Theme-specific expected values
+
+Use `const layout = getThemeLayout()` (token-backed; merged API from `test/helpers/themeLayoutFromTokens.js`, which exposes token primitives, `overlayHeight` / `verticalScrollForRow` helpers, and scenario-specific `e2e*` / `e2eGcr_*` / `e2eDensity_*` regression helpers).
```
budnix: The heading "Theme-specific expected values" signals the opposite of the section's intent -- the entire point is to write tests that are theme-agnostic. The core rule ("never branch on getLoadedTheme()") is also buried on line 81; surfacing it as the opening principle makes the why clear before the how.

```suggestion
## Theme-agnostic assertions

Every test must pass under every theme. Never branch on `getLoadedTheme()` or hardcode per-theme pixel values in specs -- use `getThemeLayout()` token helpers or live DOM measurements instead.

Use `const layout = getThemeLayout()` (token-backed; merged API from `test/helpers/themeLayoutFromTokens.js`, which exposes token primitives, `overlayHeight` / `verticalScrollForRow` helpers, and scenario-specific `e2e*` / `e2eGcr_*` / `e2eDensity_*` regression helpers).
```

handsontable/test/helpers/common.js
```diff
@@ -346,68 +423,187 @@ export const validateRows = handsontableMethodFactory('validateRows');
...
+ * Inner Handsontable instance (dropdown / handsontable / autocomplete editor list) client box matches `getSettings()`.
+ */
+export function expectInnerHandsontableEditorListClientBoxMatchesSettings() {
+  const editor = getActiveEditor();
+  const inner = editor.htEditor;
+  const root = editor.htContainer;
+
+  expect(root.clientWidth).toBe(inner.getSettings().width);
+  expect(root.clientHeight).toBe(inner.getSettings().height);
+}
```
budnix: **These two helpers hide `expect()` from the spec files and should be refactored.**

Problems:
- Failure stack points to `common.js`, not the spec that failed -- CI output and debugging both suffer.
- Reading a spec that calls `expectGetEditedCellRectFromPartial(L => ({ ... }))` tells you nothing about *what* is being asserted; you must open this file to learn the matcher shape, the `wh` injection, and the asymmetric-vs-subset branch.
- Any per-test deviation (different tolerance, extra matcher) forces either a new parameter or an inline override.

Suggested pattern: helpers return data, specs do the `expect()`. Example for the simpler one:

```js
// helper -- no expect():
export function getInnerEditorListBox() {
  const editor = getActiveEditor();
  return {
    client: { width: editor.htContainer.clientWidth, height: editor.htContainer.clientHeight },
    settings: { width: editor.htEditor.getSettings().width, height: editor.htEditor.getSettings().height },
  };
}

// spec -- assertions are visible:
const { client, settings } = getInnerEditorListBox();
expect(client.width).toBe(settings.width);
expect(client.height).toBe(settings.height);
```

Same idea for `expectGetEditedCellRectFromPartial` -> `buildEditedCellRectExpectation` that returns `{ actual, expected, wh }` and lets the spec do `expect(actual).toEqual(jasmine.objectContaining({ ...expected, ...wh }))`. The asymmetric-match branch becomes unnecessary -- callers pass jasmine matchers inside their own `buildExpected` callback, and their own `expect()` line handles it.

Side note: `expectGetEditedCellRectFromPartial` is only used in 2 baseEditor spec files (40 call sites). Even after refactoring to return data, it is narrow enough to belong in an editor-local helper (e.g. `src/editors/baseEditor/__tests__/_helpers.js`) rather than the global `common.js`.

handsontable/src/editors/autocompleteEditor/__tests__/autocompleteEditor.spec.js
```diff
@@ -2,6 +2,24 @@ describe('AutocompleteEditor', () => {
   const choices = ['yellow', 'red', 'orange', 'green', 'blue', 'gray', 'black',
     'white', 'purple', 'lime', 'olive', 'cyan'];
 
+  const E2E_DOM_RECT_EPSILON_PX = 2;
+
+  function expectDropdownBoundingRectAboveEditedCell() {
```
budnix: **Anti-pattern: `expect()` hidden in a helper.** The spec line `expectDropdownBoundingRectAboveEditedCell()` tells the reader nothing about what is being compared. The actual assertion (`ddRect.bottom <= tdRect.top + EPSILON`) lives here, and when a test fails, Jasmine's stack points at this line instead of the `it()` that failed.

Suggested shape: return the measurements, let the spec assert.

```js
function getDropdownVsEditedCell() {
  const editor = getActiveEditor();
  return {
    dd: editor.htContainer.getBoundingClientRect(),
    td: editor.TD.getBoundingClientRect(),
  };
}

// spec:
const { dd, td } = getDropdownVsEditedCell();
expect(dd.bottom).toBeLessThanOrEqual(td.top + E2E_DOM_RECT_EPSILON_PX); // above the edited cell
```

The comparison direction is then visible per test, and the same getter serves both above/below cases instead of needing two helpers.

handsontable/src/editors/autocompleteEditor/__tests__/autocompleteEditor.spec.js
```diff
@@ -2,6 +2,24 @@ describe('AutocompleteEditor', () => {
...
+    const editor = getActiveEditor();
+    const ddRect = editor.htContainer.getBoundingClientRect();
+    const tdRect = editor.TD.getBoundingClientRect();
+
+    expect(ddRect.bottom).toBeLessThanOrEqual(tdRect.top + E2E_DOM_RECT_EPSILON_PX);
+  }
+
+  function expectDropdownBoundingRectBelowEditedCell() {
```
budnix: **Same anti-pattern as its sibling above.** The `expect()` is hidden, and together these two helpers exist only because the caller cannot see the comparison being made. A single data-returning getter (see the previous comment) removes the need for both -- each spec picks the direction by writing the matcher inline, which is more readable at the call site and surfaces failure stacks back to the `it()` block.

handsontable/src/plugins/autoColumnSize/__tests__/autoColumnSize.spec.js
```diff
@@ -32,6 +32,74 @@ describe('AutoColumnSize', () => {
...
+  /**
+   * Asserts that every data cell in a column fits its rendered content without horizontal
+   * overflow (`scrollWidth <= clientWidth`). This is the core content-based guard used to
+   * supplement the DOM==plugin tautology -- it catches regressions where the column gets
+   * sized too small for its data.
+   *
+   * @param {number} col Visual column index.
+   * @param {number} rows Number of data rows to scan.
+   */
+  function assertColumnFitsContent(col, rows) {
```
budnix: **Anti-pattern: `expect()` hidden in a loop helper.** The loop is legitimate reuse, but the assertion inside it should live in the spec. If any cell fails, the stack points here instead of the `it()` block, and the reader of the spec sees `assertColumnFitsContent(col, rows)` with no indication that it's a `scrollWidth <= clientWidth` check.

Suggested shape: return the measurements, let the spec assert.

```js
function getCellOverflowMetrics(col, rows) {
  const metrics = [];
  for (let row = 0; row < rows; row++) {
    const cell = getCell(row, col);
    if (cell) metrics.push({ row, scrollWidth: cell.scrollWidth, clientWidth: cell.clientWidth });
  }
  return metrics;
}

// spec:
for (const { row, scrollWidth, clientWidth } of getCellOverflowMetrics(col, rows)) {
  expect(scrollWidth).withContext(`row ${row}`).toBeLessThanOrEqual(clientWidth);
}
```

Bonus: `withContext(row)` makes it easy to identify which row failed -- something the current helper loses.

handsontable/src/plugins/autoColumnSize/__tests__/autoColumnSize.spec.js
```diff
@@ -32,6 +32,74 @@ describe('AutoColumnSize', () => {
...
+  /**
+   * Asserts the column width is close to, but no larger than, the widest rendered content plus
+   * cell padding and border. Use this only for tests where no header decoration contributes
+   * to the computed width (e.g. no colHeaders, no hidden-column indicator, no sort indicator).
+   *
+   * @param {number} col Visual column index.
+   * @param {number} rows Number of data rows to scan.
+   * @param {number} [slackPx=12] Extra tolerance above widest text + paddingAndBorder.
+   */
+  function assertColumnNotOversized(col, rows, slackPx = 12) {
```
budnix: **Anti-pattern: helper computes a bound then asserts it.** This is exactly the `common.js` pattern already flagged elsewhere in this review -- the interesting number (`widestText + paddingAndBorder + slackPx`) is computed here, but the spec call `assertColumnNotOversized(col, rows)` hides it. Readers and failure reports lose visibility into the actual threshold.

Suggested shape: return the computed upper bound, let the spec assert.

```js
function measureColumnUpperBound(col, rows, slackPx = 12) {
  const layout = getThemeLayout();
  const paddingAndBorder = (2 * layout.cellHorizontalPadding) + layout.cellBorderWidth;
  let widestText = 0;
  for (let row = 0; row < rows; row++) {
    const cell = getCell(row, col);
    if (cell) widestText = Math.max(widestText, measureCellTextWidth(cell));
  }
  return Math.ceil(widestText + paddingAndBorder) + slackPx;
}

// spec:
const upperBound = measureColumnUpperBound(col, rows);
expect(colWidth(spec().$container, col)).toBeLessThanOrEqual(upperBound);
```

Now the spec shows what value drives the assertion, and the slack override at the call site is obvious.

handsontable/src/plugins/autoRowSize/__tests__/autoRowSize.spec.js
```diff
@@ -47,6 +47,30 @@ describe('AutoRowSize', () => {
...
+   * rounding of the theme's font metrics across themes and densities. Both outcomes are
+   * valid behaviours of autoRowSize. A deterministic version of this assertion would have
+   * to fix `colWidths` at a value wider than the longest string (single-line) or narrower
+   * than it (double-line). Rather than couple the test to a specific theme's font metrics,
+   * accept either height here and rely on the surrounding ordering assertions
+   * (`height0 < height1 < height2`) to detect regressions.
+   *
+   * @param {number} heightPx Measured row height.
+   */
+  function expectThirdRowHeaderHeightAmbiguous(heightPx) {
```
budnix: **Anti-pattern: trivial to inline.** The helper computes two valid heights and then hides `expect(...).toBeInArray([...])`. A reader of the test sees `expectThirdRowHeaderHeightAmbiguous(heightPx)` and cannot tell whether it checks equality, range, or membership without opening this function.

Suggested shape: return the allowed heights, let the spec assert.

```js
function allowedThirdRowHeaderHeights() {
  const layout = getThemeLayout();
  const singleLine = layout.defaultDataRowHeight;
  return [singleLine, singleLine + layout.lineHeight];
}

// spec:
expect(height2).toBeInArray(allowedThirdRowHeaderHeights());
```

Two lines, and the matcher is visible.

handsontable/test/helpers/themeLayoutFromTokens.js
```diff
@@ -0,0 +1,646 @@
...
+     * Scenario: `mergeCells.spec.js` -- "should expand the all overlays size after changing the row height".
+     * Top / top-inline-start clone `.height()` after the user enters three lines ("test\n\ntest") into the
+     * row-1 editor inside a 3-row merge block. The overlay contains 3 rendered rows (with +1px first-row
+     * compensation) plus two extra text line-heights introduced by the multiline content.
+     *
+     * Formula: `overlayHeight({ rows: 3 }) + (2 * lineHeight)`.
+     *
+     * @returns {number}
+     */
+    e2eDensity_0051ca7391() {
```
budnix: **Two questions about this block of helpers (there are 21 of them: `e2eDensity_0051ca7391`, `e2eGcr_3dc880f3f2`, `e2eDensity_8992c845e6`, ...).**

### 1. What do the hex suffixes mean?

The names are opaque -- nothing in the suffix tells a reader whether it's a hash of the formula, a hash of the scenario description, a random ID, a commit SHA fragment, or something else. At every call site the reader has to jump here and read the JSDoc to find out what the helper actually returns. If the hex has meaning, it is worth a one-line comment at the top of this section explaining the convention. If it does not, the names should be renamed to describe the scenario (e.g. `mergeCells_threeRowOverlayWithTwoLineHeights`).

### 2. Scope: these belong next to their tests, not in a global helpers file

Every method in this section has a JSDoc like `Scenario: mergeCells.spec.js -- autofill drag that expands a 2x2 merge into a 5-row table.` That is a 1-to-1 binding to a specific test case. Bundling scenario-specific formulas for `mergeCells`, `baseEditor/rtl/API`, `textEditor`, `autoRowSize`, `nestedRows`, `ghostTable`, `hiddenRows/plugins/manualRowMove`, etc. all in one global file means:

- Cross-plugin coupling -- editing `mergeCells` touches the same file as editing `autoRowSize`.
- The global `test/helpers/` namespace becomes a dumping ground for test-specific formulas.
- New contributors to `autoRowSize` must read a file tied to `mergeCells`, `textEditor`, and several others to find "their" helpers.

The repo already has the right pattern for this: `handsontable/src/plugins/filters/__tests__/helpers/` (with `fixtures.js`, `utils.js`) keeps filter-specific test utilities scoped to the filter plugin. Same approach would apply here:

- `src/plugins/mergeCells/__tests__/helpers/layout.js` -- the 5 merge-cell scenarios.
- `src/plugins/autoRowSize/__tests__/helpers/layout.js` -- the autoRowSize scenarios.
- `src/editors/baseEditor/__tests__/rtl/helpers/layout.js` -- the RTL editor scenarios.
- ...and so on.

`themeLayoutFromTokens.js` would then contain only the general primitives (`defaultDataRowHeight`, `overlayHeight`, `cellBorderWidth`, etc.) that every plugin's scenario helpers compose on top of. That is what a global helper file is for.

Would it be feasible to split this before merge, or is there a reason the scenarios need to live together here?

budnix (changes requested):

.claude/skills/handsontable-e2e-testing/SKILL.md
```diff
@@ -47,8 +47,43 @@ These are injected automatically. Do not import them manually.
...
 - **Plugins:** `getPlugin()`
+- **Theme layout:** `getLoadedTheme()`, `getThemeLayout()` (see `.ai/TESTING.md`)
+- **Iframe `doc.write` theme CSS:** `getE2eThemeStylesheetLinkTagsHtml()` (all themes), `getE2eThemeStylesheetLinkTagHtml(key)`, `getE2eNormalizeStylesheetLinkTagHtml()` -- from `common.js`; theme list is `E2E_REGISTERED_THEME_KEYS` in `themeLayoutFromTokens.js`, auto-discovered from `src/themes/theme/index.js` (add a theme there and the list updates automatically).
 - Full list in `test/helpers/common.js`.
 
+## Theme-agnostic assertions
+
+Every test must pass under every theme. Never branch on `getLoadedTheme()` or hardcode per-theme pixel values in specs -- use `getThemeLayout()` token helpers or live DOM measurements instead.
+
+Use `const layout = getThemeLayout()` (token-backed; merged API from `test/helpers/themeLayoutFromTokens.js`, which exposes token primitives, `overlayHeight` / `verticalScrollForRow` helpers, and scenario-specific `e2e*` / `e2eGcr_*` / `e2eDensity_*` regression helpers).
```
budnix, line 58: - SKILL.md:83 still references `expectInnerHandsontableEditorListClientBoxMatchesSettings()` and `expectGetEditedCellRectFromPartial`
 - SKILL.md:58 and `.ai/TESTING.md:376,437,453-455` still reference the `e2eGcr_*` / `e2eDensity_*` hex-named helpers

budnix (changes requested):

.claude/skills/handsontable-e2e-testing/SKILL.md
adrianspdev: @cursoragent fix

budnix (approved):

Issue #6589, opened 2019-12-30
colWidths do not work for nestedHeaders


AMBudnik:
### Description

Once we set up nestedHeaders we're not able to use `colWidths`. They do not apply but they remove the ability of a column to expand to data length. The header is as wide as its name.

### Steps to reproduce

1. Use `nestedHeaders` like here https://jsfiddle.net/apdeg7L6/
2. Check the widths of columns

![Zrzut ekranu 2019-12-30 o 13 12 40](https://user-images.githubusercontent.com/10757813/71581570-1a581600-2b06-11ea-9e53-a4c85beab294.png)

3. Replace them with the same set of `colHeaders` https://jsfiddle.net/h1eb3tgp/1/

![Zrzut ekranu 2019-12-30 o 13 12 50](https://user-images.githubusercontent.com/10757813/71581583-28a63200-2b06-11ea-8feb-1101305b9975.png)

remove the `colWidths` to check what is the actual width for this amount of data
![Zrzut ekranu 2019-12-30 o 13 15 01](https://user-images.githubusercontent.com/10757813/71581652-7cb11680-2b06-11ea-8e38-fe33e36bec66.png)

### Demo

https://jsfiddle.net/h1eb3tgp/

### Your environment
* Handsontable version: 7.3.0
It doesn't seem to be related to any version back.
* Browser Name and version: Chrome 79
* Operating System: macOS Catalina

Related https://github.com/handsontable/handsontable/issues/5981 , https://github.com/handsontable/handsontable/issues/6587

Inform [forum](https://forum.handsontable.com/t/gh-6587-nested-header-with-width/3992)

swistach:
Duplicate of #6587

Issue #5998, opened 2019-05-08
Options colWidths setting column width is invalid


hutiantian04:
### Description
Options colWidths setting column width is invalid

### Steps to reproduce

1.
![colwidth](https://user-images.githubusercontent.com/39802580/57350174-97a5d380-718f-11e9-92d9-535310dbd674.png)

2.
3.

### Demo

http://jsrun.net/TSXKp/edit

### Your environment
* Handsontable version:7.0.2
* Browser Name and version:Google Chrome version: 73.0.3683.86
* Operating System:win10

talamaska:
Have you tried using the a function?
```
colWidths: function(index) {
  return index * 10;
},
```

hutiantian04:
![image](https://user-images.githubusercontent.com/39802580/57591398-b75c4380-7563-11e9-85b4-0c113f7a3951.png)
I found the cause of the problem, but I don't know how to solve it.
【width】is the width I set，when only width > cachedWidth，use the width I set. But I always wanted to use 【width】, not 【cachedWidth】，how to solve it?

hutiantian04:
> Have you tried using the function style?
> 
> ```
> colWidths: function(index) {
>   return index * 10;
> },
> ```

I've tried it,

AMBudnik:
This is a bug, with those settings we should be able to set `colHeaders` to `100` when the header name is long but it doesn't work for `nestedHeaders`.

hutiantian04:
Hi AMBudnik 
This problem has been solved.The solution is as follows：
.handsontable th {
  width: 5px;
}

PR #12466, opened 2026-04-29
Fix nestedHeaders overriding colWidths when autoColumnSize is false

### LINKED ISSUE #12465: `nestedHeaders` ignores `colWidths` when `autoColumnSize` is false
### Describe the bug

When `nestedHeaders` is used together with explicit `colWidths` and `autoColumnSize: false`, the user-provided column widths are silently overridden by the width of the longest header label.

The `NestedHeaders` plugin's `modifyColWidth` hook always returns `Math.max(userWidth, ghostMeasuredLabelWidth)`. There is no way for a user to opt out of this auto-sizing behavior, even by explicitly disabling `autoColumnSize`.

This bug has been reported and confirmed before but never fixed in the public repository:

- #6589 — closed in 2019 as a duplicate of `handsontable/dev-handsontable#166`, which is no longer publicly accessible.
- #5998 — same root cause identified by the reporter (`width > cachedWidth` comparison in `modifyColWidth`); confirmed as a bug by @AMBudnik, then closed without a real fix when the reporter accepted the CSS workaround below.

**Steps to reproduce:**

See the demo link below. Both tables use identical `colWidths: 100` and `autoColumnSize: false`. The first table (without `nestedHeaders`) respects `colWidths`. The second table (with `nestedHeaders`) ignores it: columns under the long group label expand to ~350px, while columns under the short label stay at 100px.

**Expected:** Every column is 100px wide.
**Actual:** Columns under the long label expand to fit the label text.

**Why it matters:** It should be possible to give every column the same fixed width even when `nestedHeaders` is used — that's a basic layout requirement. Today, a single long group header silently breaks the alignment of the entire grid, and there is no API to opt out. The only workaround we found is DOM manipulation via `afterGetColHeader`, which violates Handsontable's own guidance against direct DOM access.

**Why the CSS workaround proposed in #5998 does not work for us:** The accepted "solution" there (`.handsontable th { width: 5px }`) is a global override on internal markup — it ignores the user's actual `colWidths`, breaks with theme changes, can't express varying widths per column, and collides with design-system tokens. What we need is for `colWidths` and `autoColumnSize: false` — both already public API — to behave as documented when `nestedHeaders` is enabled.

**Proposed fix:** Treat `autoColumnSize: false` as an opt-out for `nestedHeaders`' ghost-table override. Fully back-compatible — default behavior unchanged. I have a working patch with E2E tests and would like to open a PR.

### Video/Screenshots

Image

https://stackblitz.com/edit/sb1-o9fggvqo?file=src%2FApp.tsx

### Handsontable version

17.0.1

### Framework version

_No response_

### Your environment

macOS 26.3, Chrome 147.0.7727.117

csprle:
## Summary

Fixes #12465. Related: #5998, #6589.

When `nestedHeaders` is combined with `autoColumnSize: false` and explicit `colWidths`, the plugin's `modifyColWidth` hook ignores the user-provided width and returns the ghost-table-measured header label width instead. This PR makes the plugin respect `autoColumnSize: false` as an opt-out.

## Change

In `NestedHeaders.#onModifyColWidth`, return the user-provided `width` unchanged when `autoColumnSize === false`. Default behavior (`autoColumnSize` unset or `true`) is unchanged.

## Tests

Added one E2E test in `nestedHeaders.spec.js` covering the new behavior. Verified red-green:

- Without the fix: test fails — columns under the long header are 127px / 123px instead of 50px
- With the fix: all 138 nestedHeaders specs pass

## Back-compatibility

Not a breaking change:
- No default values changed
- Existing behavior (no `autoColumnSize` setting, or `autoColumnSize: true`) unchanged
- The opt-out is purely additive

ClickUp task: N/A — external contribution

---

> [!NOTE]
> **Low Risk**
> Low risk, localized change to `modifyColWidth` behavior gated behind `autoColumnSize === false`, with a regression test covering the new expectation.
> 
> **Overview**
> Fixes `nestedHeaders` column width calculation so that when `autoColumnSize: false` is explicitly set, the plugin no longer replaces user-provided `colWidths` with ghost-table-measured header label widths.
> 
> Adds an E2E spec asserting fixed widths remain unchanged under long nested header labels, and includes a changelog entry for issue `#12465`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3548f5674bf9e8d3dad95c286abfd03108f3f180. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changelogs/12465.json
+++ b/.changelogs/12465.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed `nestedHeaders` overriding `colWidths` with measured header label widths when `autoColumnSize` is explicitly disabled",
+  "type": "fixed",
+  "issueOrPR": 12465,
+  "breaking": false,
+  "framework": "none"
+}
--- a/handsontable/src/plugins/nestedHeaders/__tests__/nestedHeaders.spec.js
+++ b/handsontable/src/plugins/nestedHeaders/__tests__/nestedHeaders.spec.js
@@ -48,6 +48,24 @@ describe('NestedHeaders', () => {
       expect(getColWidth(1)).toBeGreaterThan(50);
       expect(headers[1].offsetWidth).toBeGreaterThan(100);
     });
+
+    it('should respect colWidths and not expand columns to fit nested header labels when autoColumnSize is false', async() => {
+      handsontable({
+        data: createSpreadsheetData(10, 10),
+        colHeaders: true,
+        autoColumnSize: false,
+        colWidths: 50,
+        nestedHeaders: [
+          ['a', { label: 'This is a very long group header label', colspan: 2 }, 'c', 'd'],
+          ['a', 'b', 'c', 'd', 'e']
+        ]
+      });
+
+      expect(getColWidth(0)).toBe(50);
+      expect(getColWidth(1)).toBe(50);
+      expect(getColWidth(2)).toBe(50);
+      expect(getColWidth(3)).toBe(50);
+    });
   });
 
   describe('cooperation with drop-down menu element', () => {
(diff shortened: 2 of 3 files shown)


csprle:
How is the release process for this? I didn't see any mention in the latest 17.1.0 release of this?

PR #12508, opened 2026-05-08
DEV-1623: Translate merged cells through manual moves and column freeze (#10437)

### LINKED ISSUE #10437: [Bug]: Moving columns when there are some merged cells messes up merging
### Describe the bug

Moving a column should bring it's merged cells with it. Instead it loses its merge settings and adopts the settings of whatever column's place it took. The same happens for all other columns that were pushed by the move.

### Video/Screenshots

Start with this:
image
Column 4 cells are vertically merged, and Column 5 cells are not.
Select and drag Column 5, dropping it between Columns 3 and 4. The result:
image
Now Column 5 cells are merged, and Column 4 cells are not. This is wrong (and not what happens in Excel).

https://codesandbox.io/s/handsontable-data-grid-merge-manualcolumnmove-switching-merged-and-unmerged-columns-places-switches-merged-cells-fphcn3?file=/src/index.js

### Handsontable version

13.0.0

### Framework version

_No response_

### Your environment

Windows 11, Firefox 115.0

adrianspdev:
## Summary

Fixes [#10437](https://github.com/handsontable/handsontable/issues/10437): merged cells stayed at visual indexes when `manualColumnMove`, `manualRowMove`, or `manualColumnFreeze` reordered the grid, so a merge appeared to swap with whatever column/row landed on top of it.

The mergeCells plugin now:
- snapshots each merge's physical span in `beforeColumnMove` / `beforeRowMove` / `beforeColumnFreeze` / `beforeColumnUnfreeze`,
- translates the snapshot back to new visual coordinates in the matching `after*` hook,
- auto-splits a merge whose physical span is no longer contiguous after the reorder,
- drops any resulting single-cell fragment (`colspan === 1 && rowspan === 1`).

Bootstrap-time reorders fired by `manualColumnMove: [...]` initial config are guarded by a plugin-internal `#initialized` flag, so user-declared merge coordinates aren't double-translated.

## Test plan

- [x] `npm run lint --prefix handsontable`
- [x] `npm run test:unit --prefix handsontable` (testpathpattern `mergeCells/__tests__/cellsCollection.unit`) - 27 tests pass, including new coverage for `detectContiguousRuns`, `capturePhysicalSpans`, `translateAfterAxisMove`
- [x] `npm run test:e2e --prefix handsontable` - 9370 specs, 0 failures, 27 pending
- [ ] Manual repro of issue #10437: drag column F before column E in a grid with a vertical merge on column E - merge should follow column E to its new visual position
- [ ] Manual freeze/unfreeze of a column that overlaps a merge

ClickUp task: https://app.clickup.com/t/9015210959/DEV-1623

---

> [!NOTE]
> **Medium Risk**
> Changes core `mergeCells` behavior during row/column reorders by rewriting merge coordinates (including auto-splitting/dropping fragments), which could affect edge cases with index mappers and undo/redo interactions.
> 
> **Overview**
> Fixes merged cells getting “stuck” to visual coordinates when using `manualColumnMove`, `manualRowMove`, or `manualColumnFreeze` by translating merges based on their *physical* row/column spans.
> 
> Adds snapshot/translate logic in the `mergeCells` plugin via new collection helpers (`capturePhysicalSpans`, `translateAfterAxisMove`, `detectContiguousRuns`), including **auto-splitting** merges that become non-contiguous after reordering and **dropping** any resulting single-cell fragments.
> 
> Expands unit/integration test coverage for the new translation behavior (including undo/redo and rejected moves), updates the merge-cells guide to document the new semantics, and adds a changelog entry.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c8a40aa65c6209492f8581045ab7de8f10f42d3a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-1623 [Bug]: Moving columns when there are some merged cells messes up merging](https://app.clickup.com/t/9015210959/DEV-1623)

handsontable/src/plugins/mergeCells/__tests__/mergeCells.spec.js
```diff
@@ -120,6 +120,265 @@ describe('MergeCells', () => {
...
+      });
+
+      const merges = getPlugin('mergeCells').mergedCellsCollection.mergedCells;
+
+      expect(merges.length).toBe(2);
+      expect(merges[0].col).toBe(2);
+      expect(merges[1].col).toBe(2);
+    });
+
+    xit('should follow merges through column freeze', async() => {
```
budnix: This test is `xit` — was it intentionally skipped or left from development? The freeze hooks are all wired up and the assertions look correct to me, curious if it actually fails when you enable it. If it does we probably want to fix it before merging since otherwise we're shipping the freeze path with zero E2E coverage.

handsontable/src/plugins/mergeCells/__tests__/mergeCells.spec.js
```diff
@@ -120,6 +120,265 @@ describe('MergeCells', () => {
...
+          { row: 0, col: 0, rowspan: 1, colspan: 3 },
+        ],
+      });
+
+      getPlugin('manualColumnMove').moveColumn(2, 0);
+      await render();
+
+      const merges = getPlugin('mergeCells').mergedCellsCollection.mergedCells;
+
+      expect(merges.length).toBeGreaterThanOrEqual(0);
```
budnix: These two assertions won't catch a regression — `>= 0` is always true for any array, and `<= 2` lets a no-op through. Do we know what the expected count is for this case? If it's deterministic, a `toBe(N)` would make this actually useful.

handsontable/src/plugins/mergeCells/cellsCollection.js
```diff
@@ -516,6 +516,137 @@ class MergedCellsCollection {
...
+
+        replacements.push(replacement);
+      });
+    });
+
+    this.mergedCells.length = 0;
+    this.mergedCellsMatrix.clear();
+
+    replacements.forEach((info) => {
+      this.add(info);
```
budnix: Should this be `this.add(info, true)`? Without `auto: true`, every replacement runs the overlap validator and can emit `IS_OVERLAPPING_WARNING` to the console even though this is an internal rebuild, not user input. `shiftCollections` skips `add()` altogether for the same reason. The `auto` flag is exactly the escape hatch for this situation.

handsontable/src/plugins/mergeCells/cellsCollection.js
```diff
@@ -516,6 +516,137 @@ class MergedCellsCollection {
...
+
+    return runs;
+  }
+
+  /**
+   * Translate the merged cells collection after a manual row/column reorder, splitting
+   * merges whose physical indexes are no longer contiguous in the new visual order.
+   *
+   * Note: single-cell fragments (`colspan === 1 && rowspan === 1`) are dropped because
+   * they no longer represent a merge. Callers should advertise this in user-facing docs.
```
budnix: The comment here says "Callers should advertise this in user-facing docs" — did that end up somewhere? I didn't see it in the merge-cells guide. The silent drop of a bisected merge is the kind of thing a user would want to know about, especially if they're listening to `afterMergeCells` and wondering why they got extra events.

handsontable/src/plugins/mergeCells/__tests__/mergeCells.spec.js
adrianspdev: Good catch - the freeze hooks were wired, the test was just leftover xit. Converted to it; passes locally with the deterministic merge translation through manualColumnFreeze.

handsontable/src/plugins/mergeCells/__tests__/mergeCells.spec.js
adrianspdev: Agreed - the scenario is deterministic, the bounded asserts were leftover from exploration. Renamed the test to reflect what it actually exercises (internal re-permutation with a hidden column inside the merge) and tightened to exact toBe asserts: 1 merge at row 0 / col 0 / rowspan 1 / colspan 3. Real overlap-drop isn't reachable via public API in this setup since toVisualColumn is bijective.

handsontable/src/plugins/mergeCells/cellsCollection.js
adrianspdev: Right, this is an internal rebuild - switched to this.add(info, true). Matches the shiftCollections pattern and silences the spurious IS_OVERLAPPING_WARNING.

handsontable/src/plugins/mergeCells/cellsCollection.js
adrianspdev: It didn't land - now added. New section 'Behavior during row/column reorder and column freeze' in docs/content/guides/cell-features/merge-cells/merge-cells.md covers auto-split, silent singleton drop, and undo/redo restore. JSDoc note now points at it.

budnix (approved):

PR #12517, opened 2026-05-08
DEV-1025: Enhance table overflow handling for dynamic height scenarios

demtario:
### Context

Without a fixed `height`, Walkontable falls back to window scrolling and the root element did not act as a horizontal boundary. If `width` was set but `height` was omitted, the inner table could still paint wider than the configured width, so the grid did not look “contained”.

This change sets `overflow-x: clip` on the Handsontable root when `width` is set and there is no effective `height` (after resolving functions and `beforeHeightChange`), including `null` when height is cleared via `updateSettings`. That keeps vertical scrolling on the window while the root width actually clips horizontal overflow. It uses the processed `height` value so hooks that coerce height to `null` stay consistent with omitting height.

### How has this been tested?

* Added E2E coverage in `handsontable/test/e2e/settings/widthWithoutHeight.spec.js`: width without height, `height: undefined`, `updateSettings` with `height: null`, and `beforeHeightChange` returning `null` (assertions on root width, `overflow-x`, and `.wtHolder` width).
* Adjusted `handsontable/test/e2e/Core_count.spec.js` so `countVisibleCols` is asserted with a viewport wide enough for ten columns when using width-without-height clipping (same semantics as the fix).
* Manual tests in the dev-generated.html with cases based on jsfiddles from task. 

### Types of changes

### Related issue(s):
1. [DEV-1102](https://app.clickup.com/t/9015210959/DEV-1102)

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes root overflow and Walkontable trimming measurement, which affects layout, viewport column counts, and many updateSettings edge cases across browsers and jsdom.
> 
> **Overview**
> Fixes grids that looked wider than their configured **`width`** when **`height`** was omitted: the root now applies **`overflow-x: clip`** whenever there is a non-`auto` effective width and no effective height (including after **`height: null`** or hooks that coerce height), and clears that clip when width becomes **`auto`**. When height is set, the existing **`overflow: clip`** shorthand is left alone so partial **`updateSettings`** does not leave stale horizontal clipping or break the height path.
> 
> Walkontable’s trimming-container logic now treats the computed **`overflow`** shorthand as authoritative in real browsers and only reads **`overflowX` / `overflowY`** when the shorthand is missing (e.g. jsdom). Tests cover init/update cycles, restored user **`overflow`**, and **`countVisibleCols`** with a wide enough viewport under width-only clipping.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4eaf117d639a1dfc972d97adb17a2feda5e23aa4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-1102 The `height` option does not have `auto` mentioned as an option](https://app.clickup.com/t/9015210959/DEV-1102)

handsontable/src/core.js
```diff
@@ -3100,6 +3100,17 @@ export default function Core(rootContainer, userSettings, rootInstanceSymbol = f
...
+
+      // When `height` is not set, the table uses window vertical scroll. In that mode, not setting any overflow
+      // on the root element can allow the inner table to visually overflow the configured width.
+      // Setting only the horizontal overflow to `clip` keeps window scrolling intact while ensuring `width`
+      // constrains the grid.
+      // Treat “no effective height” like omitting height: missing key, `null`, or `undefined` after resolution.
+      // Use the processed `height` value (after function resolution and `beforeHeightChange`), not `settings.height`,
+      // so hooks that coerce height to `null` still get horizontal clipping when `width` is set.
+      if (height === undefined || height === null) {
+        instance.rootElement.style.overflowX = 'clip';
```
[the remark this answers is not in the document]
demtario: `clip` was also used in other parts of this code, so this compatibility issue already exists. Comment above this line was updated

demtario:
@cursoragent review

handsontable/src/core.js
```diff
@@ -3116,6 +3116,32 @@ export default function Core(rootContainer, userSettings, rootInstanceSymbol = f
...
+      const effectiveWidth = instance.rootElement.style.width;
+      const css = instance.rootWindow.CSS;
+      const overflowXClipSupported = css && isFunction(css.supports) && css.supports('overflow-x', 'clip');
+      const shouldClipHorizontalOverflow = effectiveWidth && effectiveWidth !== 'auto';
+
+      if (!effectiveHeight) {
+        let overflowX = '';
+
+        if (shouldClipHorizontalOverflow) {
+          overflowX = overflowXClipSupported ? 'clip' : 'hidden';
```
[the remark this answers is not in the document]
demtario: `overflow-x: hidden` code was reverted. Lets assume that `overflow-x: clip` must be supported in the browser for this to work. This is in line with policy to support only 2 latest browser versions in Handsontable.

demtario:
Blocked till I get access to "argos" to be able to check what was reported.

handsontable/src/3rdparty/walkontable/src/table/master.ts
```diff
@@ -175,9 +175,14 @@ class MasterTable extends Table {
...
         let height = trimmingOffsetHeight;
         const overflowValues = ['auto', 'hidden', 'scroll', 'clip'];
-        // getStyle() may return a compound value (e.g. 'auto hidden') when overflow-x and
-        // overflow-y are set independently. Split on whitespace so each token is checked.
-        const hasScrollOverflow = trimmingOverflow.split(' ').some(v => overflowValues.includes(v));
+        // Chrome resolves the overflow shorthand from individual properties natively
+        // (e.g. overflowX='clip' → computed overflow='clip visible'). Only fall back to
+        // reading overflowX/overflowY individually when the shorthand is absent — that
+        // covers jsdom, where setting overflowX inline does not update the overflow shorthand.
+        const hasScrollOverflow = trimmingOverflow !== undefined
```
budnix: The fallback block can never run. `trimmingOverflow` is assigned as `getStyle(...) ?? ''` on line 142, so it is always a string — never `undefined`. The check `!== undefined` is always `true`. The `else` branch is dead code.

The scenario described in the comment — an environment where the `overflow` shorthand does not reflect per-axis values — is never actually handled. Using a truthy check fixes this, since an empty string (shorthand absent or unset) would send execution to the fallback:

```suggestion
        const hasScrollOverflow = trimmingOverflow
          ? trimmingOverflow.split(' ').some(v => overflowValues.includes(v))
          : [getStyle(trimmingElement, 'overflowX', rootWindow), getStyle(trimmingElement, 'overflowY', rootWindow)]
            .some(v => v && overflowValues.includes(v));
```

handsontable/src/core.ts
```diff
@@ -3214,6 +3214,24 @@ export default function Core(
...
+    // width blocks ran) so partial updateSettings calls see the correct state.
+    // When height IS set, the height block's `overflow: clip` shorthand handles both axes — leave
+    // overflowX untouched to avoid breaking that shorthand.
+    // Browser compatibility: `overflow-x: clip` requires Safari 16+. On Safari 14.1–15.x it silently
+    // falls back to `visible` (graceful degradation — pre-existing behavior, not a new regression).
+    if (typeof settings.height !== 'undefined' || typeof settings.width !== 'undefined') {
+      const effectiveHeight = instance.rootElement.style.height;
+      const effectiveWidth = instance.rootElement.style.width;
+
+      if (!effectiveHeight) {
```
budnix, line 3229: There is a case where this block overwrites a style that was just restored. When `height: null` is passed with a `width` and `data-initialstyle` contains `overflow` (but no `height`), the height block calls `setAttribute('style', initialStyle)`. That restores the original style. Then `effectiveHeight` reads as `''` — falsy — and this block writes `overflowX: 'clip'` on top.

If the original style had `overflow: hidden`, the X axis is now `clip` instead. That is a real behavior difference. `hidden` creates a block formatting context and allows programmatic scroll. `clip` does neither.

handsontable/src/__tests__/settings/widthWithoutHeight.spec.js
```diff
@@ -0,0 +1,185 @@
+describe('settings', () => {
```
budnix: Every file in `src/__tests__/settings/` is named after one option — `width.spec.js`, `height.spec.js`, `colWidths.spec.js`. This one is named after a scenario instead. The tests here cover `width` behavior, so they fit better as a nested `describe('when height is not provided', ...)` block inside the existing `width.spec.js`.

budnix (approved):

handsontable/src/__tests__/settings/widthWithoutHeight.spec.js
demtario: e2e tests moved to `width.spec.js`

handsontable/src/core.ts
demtario: Fixed. The block now only writes `overflow-x` when the current value is `''` or `clip` (the values we set), so a restored `overflow: hidden` is left intact instead of becoming clip hidden.

handsontable/src/3rdparty/walkontable/src/table/master.ts
demtario: Fixed by switching to a truthy check as suggested.

demtario:
dev generated for future reference
[dev-generated.html](https://github.com/user-attachments/files/28458332/dev-generated.html)

PR #12278, opened 2026-04-07
fix(wrappers): skip init-only settings in updateSettings (#12242)

sl01k:
### Context

Fixes #12242

When using framework wrappers (React, Vue, Angular), changing an init-only setting prop (e.g., `renderAllRows`, `renderAllColumns`, `layoutDirection`, `ariaTags`) after initialization causes the application to crash with:

```
Uncaught Error: The `renderAllRows` option can not be updated after the Handsontable is initialized.
```

The root cause: all three framework wrappers passed init-only settings through to `updateSettings()` when the prop value changed between renders. Handsontable Core then threw because these settings are marked as `initOnly: true`.

**Fix:** All three wrappers now silently skip init-only settings when calling `updateSettings()` after initialization. The list of init-only settings is read from the Handsontable instance's `_initOnlySettings` metadata, so it stays in sync with Core automatically.

### How has this been tested?

- **React wrapper**: Updated existing test and added new unit tests for `SettingsMapper.getSettings()` verifying init-only keys are excluded after init, but included during init.
- **Vue wrapper**: Added unit tests for `prepareSettings()` verifying init-only keys are filtered when `currentSettings` contains `_initOnlySettings`.
- **Angular wrapper**: Added integration test verifying init-only settings are stripped before reaching `updateSettings()`.

All wrapper test suites pass:
- `pnpm --filter @handsontable/react-wrapper run test` -- 65/65 passed
- `pnpm --filter @handsontable/vue3 run test` -- 20/20 passed
- `pnpm --filter @handsontable/angular-wrapper run test` -- 94/94 passed

### Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature or improvement (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Additional language file or change to the existing one (translations)

### Related issue(s):
1. #12242

### Affected project(s):
- [ ] `handsontable`
- [x] `@handsontable/angular-wrapper`
- [x] `@handsontable/react-wrapper`
- [x] `@handsontable/vue3`

### Checklist:
- [x] I have reviewed the guidelines about [Contributing to Handsontable](https://github.com/handsontable/handsontable/blob/master/CONTRIBUTING.md) and I confirm that my code follows the code style of this project.
[...]

Issue #12242, opened 2026-04-01
react-wrapper: `renderAllRows` init-only option causes crash on re-render when value changes


yamaaaaaa31:
## Description

When using `@handsontable/react-wrapper`, changing the `renderAllRows` prop after initialization (e.g., toggling between mobile/desktop layouts on window resize) causes an unrecoverable error:

```
Uncaught Error: The `renderAllRows` option can not be updated after the Handsontable is initialized.
    at ExtendMetaPropertiesMod.<anonymous> (extendMetaProperties.mjs:65:15)
    at Object.set [as renderAllRows] (extendMetaProperties.mjs:117:18)
    at Core.updateSettings (core.mjs:2535:22)
    at HotTableInner.useUpdateEffect (react-handsontable.mjs:2041:67)
```

## Steps to reproduce

1. Open the StackBlitz reproduction: https://stackblitz.com/github/yamaaaaaa31/handsontable-renderAllRows-resize-bug
2. Resize the browser window across the 768px width boundary
3. Observe the error in the console

**Reproduction repository**: https://github.com/yamaaaaaa31/handsontable-renderAllRows-resize-bug

## Expected behavior

The react-wrapper should either:

1. **Silently ignore** init-only settings in `updateSettings` when the value hasn't changed (this already works via `shouldSkipProp`)
2. **Silently ignore** init-only settings even when the value has changed, since Handsontable Core will throw anyway — or at minimum, emit a warning instead of throwing
3. **Force a remount** of the Handsontable instance when an init-only setting changes, since it cannot be updated in-place

Currently, `shouldSkipProp` in `SettingsMapper.getSettings()` only skips init-only keys when `prevProps[key] === properties[key]`. When the value actually changes, it passes through to `updateSettings`, which throws.

## Actual behavior

The app crashes with an unrecoverable error.

## Root cause analysis

In `react-handsontable.js`, the `useUpdateEffect` hook calls `updateSettings` on every re-render:

```javascript
// react-handsontable.js L1034-1043
useUpdateEffect(function () {
  clearCache();
  var hotInstance = getHotInstance();
  var newGlobalSettings = createNewGlobalSettings(false, prevProps.current);
  prevProps.current = props;
  hotInstance.updateSettings(newGlobalSettings, false);
});
```

`createNewGlobalSettings` reads `_initOnlySettings` from the Handsontable instance and passes it to `SettingsMapper.getSettings()`:

```javascript
// react-handsontable.js L427-433
[...]

PR #12572, opened 2026-05-13
fix(react-wrapper): skip init-only props after mount

prantikmedhi:
## Summary\n- Skip init-only React wrapper props from post-mount updateSettings calls, even when their values change\n- Add unit coverage for init-only props during init vs after mount\n- Update the component test to expect no crash when an init-only prop changes\n\n## Verification\n- pnpm --dir wrappers/react-wrapper exec jest test/settingsMapper.spec.tsx --runInBand\n- pnpm install --frozen-lockfile (with PUPPETEER_SKIP_DOWNLOAD=1 for the repo install)\n\n## Notes\n- The broader hotTable suite needs the built Handsontable package; I validated the focused mapper test and the code path change.

---

> [!NOTE]
> **Low Risk**
> Low risk wrapper behavior change: it only alters how init-only Handsontable options are filtered out of `updateSettings` calls after mount, with added tests to lock in the new behavior.
> 
> **Overview**
> Prevents the React wrapper from attempting to apply *init-only* Handsontable settings after the table has mounted by always omitting keys listed in `initOnlySettingKeys` from non-init `SettingsMapper.getSettings` results.
> 
> Updates tests to reflect that changing an init-only prop (e.g. `renderAllRows`) after mount should no longer throw or log errors, and adds mapper unit coverage for including init-only settings during init vs skipping them post-init.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0f1e9140592404944aa4444c289098ab4dc2dc62. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/wrappers/react-wrapper/src/settingsMapper.ts
+++ b/wrappers/react-wrapper/src/settingsMapper.ts
@@ -23,11 +23,7 @@ export class SettingsMapper {
       initOnlySettingKeys?: Array<keyof Handsontable.GridSettings>
     } = {}): Handsontable.GridSettings {
     const shouldSkipProp = (key: keyof Handsontable.GridSettings) => {
-      // Omit settings that can be set only during initialization and are intentionally modified.
-      if (!isInit && initOnlySettingKeys.includes(key)) {
-        return prevProps[key] === properties[key];
-      }
-      return false;
+      return !isInit && initOnlySettingKeys.includes(key);
     };
     let newSettings: Handsontable.GridSettings = {};
 
--- a/wrappers/react-wrapper/test/hotTable.spec.tsx
+++ b/wrappers/react-wrapper/test/hotTable.spec.tsx
@@ -123,8 +123,19 @@ describe('Updating the Handsontable settings', () => {
     expect(JSON.stringify(hotInstance.getSettings().data)).toEqual('[[2]]');
   });
 
-  it('should throw an error when trying to update init-only settings after inializing the component', async () => {
-    console.error = jest.fn();
+  it('should not throw an error when trying to update init-only settings after inializing the component', async () => {
+    const errorMock = jest.fn();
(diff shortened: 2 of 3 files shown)


qunabu:
@prantikmedhi thanks. Can you use our template for description and provide a valid summary of this change

demtario:
Hi @prantikmedhi, thanks for digging into this, and sorry for the slow reply.

Your diagnosis was correct: `SettingsMapper.getSettings` should drop init-only keys from every post-mount `updateSettings` call, not only when their values happen to be unchanged. That is exactly the change we landed in #12278 (for issue #12242) on 2026-04-07, and it shipped in 17.1.0. Current `develop` and `master` both have:

```ts
if (!isInit && initOnlySettingKeys.includes(key)) {
  return true;
}
```

Because of that, there is nothing left for us to merge here, so I am going to close the PR. The overlap was not your fault, it is a base-branch artifact. This PR targets `master`, and the fix only reached `master` on 2026-05-19 with the 17.1.0 release, six days after you opened it. It had been on `develop` since early April.

So the one thing worth carrying into your next contribution: please branch from and target **`develop`**, as described in CONTRIBUTING.md. On this repo `master` can trail `develop` by weeks, so a quick check of `develop` (and of closed issues) before starting will save you the kind of duplicated work that happened here.

One more note on the same fix, in case it is useful: the crash affected all three wrappers, so #12278 also patched Vue 3 (`wrappers/vue3/src/helpers.ts`) and Angular (`hot-table.component.ts`). Wrapper parity is something we try to keep in mind for any change like this.

Thanks again for the report and the tests. The analysis was sound, it just arrived after the same fix had landed. We would be glad to see another PR from you against `develop`.

Issue #12576, opened 2026-05-13
Memory leak in IntersectionObserver


smlkA:
### Describe the bug

IntersectionObserver instances are not disconnected properly when, for example, document.body.height is zero - they keep accumulating.

Bug was found in version 15 and version 17.

Prerequisite: Version 17.1.0-rc10 is used. This example is an updated version from [Memory leak with the React based renderers](https://github.com/handsontable/handsontable/issues/12568), but this is an independent issue.

Steps to reproduce:
1. Open the [example](https://stackblitz.com/~/github.com/smlkA/handsontable-memory-leak)
2. Download the code or run it in different tab
3. Open DevTools and take memory snapshot
4. Switch the "tab" inside the example 20 times
5. Take another memory snapshot
6. Confirm that number of IntersectionObserver instances is around 40

<img width="1512" height="865" alt="Image" src="https://github.com/user-attachments/assets/b85a45d0-2114-41d9-b721-5ddc7a9d145e" />

The behavior is caused by this line:
https://github.com/handsontable/handsontable/blob/develop/handsontable/src/helpers/dom/element.js#L1238

When document.body has zero height, the IntersectionObserver root bounds are a zero-height rectangle. No element can ever intersect a zero-height area, so [entry.isIntersecting](https://github.com/handsontable/handsontable/blob/develop/handsontable/src/helpers/dom/element.js#L1232) is always false. The observer never calls [unobserve()](https://github.com/handsontable/handsontable/blob/develop/handsontable/src/helpers/dom/element.js#L1234) and leaks permanently. Each mount/unmount cycle (e.g. tab switching in React) accumulates leaked observers.

Proposed solution:
instead of `root: elementToBeObserved.ownerDocument.body` use `root: null`
as it viewport always has a non-zero size.

Also, it would be better to use disconnect() instead of unobserve() — since this observer only watches a single element and is meant to fire once, disconnect() fully cleans up the observer rather than just stopping observation of one target

### Provide a link to the demo with the bug reproduction

https://stackblitz.com/~/github.com/smlkA/handsontable-memory-leak

### Handsontable version

17.0.1, 15.3.0

### Your environment

Version 148.0.7778.96 (Official Build) (arm64), Mac

adrianspdev:
Thanks for the catch and the detailed report. Fixed in #12578. This will be included in the next release.

smlkA:
@adrianspdev Thank you for such fast response!

PR #12578, opened 2026-05-14
DEV-1676: Fix IntersectionObserver leak in observeVisibilityChangeOnce

adrianspdev:
### Context

The PR fixes a memory leak in `observeVisibilityChangeOnce` (`handsontable/src/helpers/dom/element.js`) reported in #12576. The helper registers a fire-once `IntersectionObserver` with `root: ownerDocument.body`. When `document.body` has a zero-height rect, browsers that strictly follow the IntersectionObserver spec report `entry.isIntersecting === false` for any target, so the `unobserve()` cleanup path never runs and one observer leaks per Handsontable mount-while-hidden cycle (for example, a React tab that mounts the grid while inactive).

Two changes:
- Drop the `root: ownerDocument.body` option so the observer uses the viewport as its intersection root - always non-zero, robust across shadow-DOM and cross-document scenarios.
- Replace `observer.unobserve(target)` with `observer.disconnect()`. The observer only ever watches a single target and is meant to fire once; `disconnect()` is the correct full-cleanup primitive.

### How has this been tested?

Added two regression tests in `handsontable/src/helpers/dom/__tests__/element.spec.js`:

- `should fully disconnect the observer after the element becomes visible` - patches `IntersectionObserver` and asserts `disconnect()` is called once and `unobserve()` is never called.
- `should not leak observers when document.body has zero height` - patches `IntersectionObserver` to count live instances, sets `document.body.style.height = '0px'`, mounts and shows five elements, and asserts the live observer count returns to 0.

Commands run locally:

```
npm run test:e2e --prefix handsontable --testPathPattern=helpers/dom/__tests__/element
# -> 20 specs, 0 failures

npx eslint src/helpers/dom/element.js src/helpers/dom/__tests__/element.spec.js
# -> No issues found
```

Also verified manually in headless Chrome via two local demo pages (gitignored) that count visibility-observer instances during a mount/destroy loop:
- Released v17.0.1: 20 observers created, 0 cleaned up (`Live: 20`).
- This PR: 20 created, 20 cleaned up via `disconnect()` (`Live: 0`).

### Types of changes

### Related issue(s):

1. #12576
2. DEV-1676

### Affected project(s):

- [x] \`handsontable\`
- [ ] \`@handsontable/angular-wrapper\`
- [ ] \`@handsontable/react-wrapper\`
- [ ] \`@handsontable/vue3\`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-1676

---

> [!NOTE]
> **Medium Risk**
> Changes `observeVisibilityChangeOnce` cleanup semantics from `unobserve()` with a custom `root` to `disconnect()` with the default root, which could alter when the callback fires in edge visibility scenarios; otherwise scoped to a small DOM helper.
> 
> **Overview**
> Fixes a leak in `observeVisibilityChangeOnce` by **removing the `IntersectionObserver` `root: document.body` option** and **fully cleaning up via `observer.disconnect()`** after the first `isIntersecting` event (instead of `unobserve()` gated by `offsetParent`).
> 
> Adds regression coverage asserting the observer is disconnected (and not unobserved) after firing, and that repeated mounts while `document.body` has `height: 0` do not leave any active observers. Includes a changelog entry for the fix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8d148b65ef7232b861d7588323e195050481669e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/handsontable/src/helpers/dom/element.js
+++ b/handsontable/src/helpers/dom/element.js
@@ -1229,7 +1229,7 @@ export function isDetached(element) {
 export function observeVisibilityChangeOnce(elementToBeObserved, callback) {
   const visibilityObserver = new IntersectionObserver((entries, observer) => {
     entries.forEach((entry) => {
-      if (entry.isIntersecting && elementToBeObserved.offsetParent !== null) {
+      if (entry.isIntersecting) {
         callback();
         observer.disconnect();
       }
--- a/handsontable/src/helpers/dom/element.js
+++ b/handsontable/src/helpers/dom/element.js
@@ -1231,11 +1231,9 @@ export function observeVisibilityChangeOnce(elementToBeObserved, callback) {
     entries.forEach((entry) => {
       if (entry.isIntersecting && elementToBeObserved.offsetParent !== null) {
         callback();
-        observer.unobserve(elementToBeObserved);
+        observer.disconnect();
       }
     });
-  }, {
-    root: elementToBeObserved.ownerDocument.body
   });
 
   visibilityObserver.observe(elementToBeObserved);
--- a/.changelogs/12578.json
+++ b/.changelogs/12578.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed a memory leak where IntersectionObserver instances were not properly disconnected when `document.body` had zero height.",
+  "type": "fixed",
+  "issueOrPR": 12578,
+  "breaking": false,
+  "framework": "none"
+}
--- a/handsontable/src/helpers/dom/__tests__/element.spec.js
+++ b/handsontable/src/helpers/dom/__tests__/element.spec.js
@@ -21,6 +21,106 @@ describe('DOM helpers', () => {
 
       document.body.removeChild(element);
     });
+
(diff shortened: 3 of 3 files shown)


qunabu:
Task linked: [DEV-1676 Memory leak in IntersectionObserver](https://app.clickup.com/t/9015210959/DEV-1676)

handsontable/src/helpers/dom/element.js
```diff
@@ -1231,11 +1231,9 @@ export function observeVisibilityChangeOnce(elementToBeObserved, callback) {
     entries.forEach((entry) => {
       if (entry.isIntersecting && elementToBeObserved.offsetParent !== null) {
```
budnix: The `offsetParent !== null` check is now redundant. With `root: null` the viewport is the intersection root, so `isIntersecting: true` already means the element is visible — the only case where `offsetParent` would be `null` despite intersecting is a `position: fixed` element, and blocking the callback there is wrong. This can just be:

```js
if (entry.isIntersecting) {
```

budnix (approved):

Issue #9894, opened 2022-09-13
How to add 'add_child' or 'detach_from_parent' in Context menu custom configuration


footman9988:
### Description

hi ：
24/5000 
I found that I could not add ’add_child‘ or 'detach_from_parent ' When I'm using the Context Menu with custom configuration.

### Steps to reproduce

1. 
2.
3.

### Demo

https://jsfiddle.net/footman8899/t7jwy6qc/5/

### Your environment
* Vue wrapper version:
* Handsontable version:
* Browser name and version:
* Operating system:

AMBudnik:
Thank you for sharing the report @footman9988 

I replicated the issue. It seems that the [`nestedRows`](https://github.com/handsontable/handsontable/tree/6a1706061163b8d1f7752e54ef6f10efbc764b8b/handsontable/src/plugins/nestedRows) do not export the menu, as [`borders`](https://github.com/handsontable/handsontable/blob/6a1706061163b8d1f7752e54ef6f10efbc764b8b/handsontable/src/plugins/customBorders/contextMenuItem/bottom.js#L8) or [`hiddenColumns`](https://github.com/handsontable/handsontable/blob/ce7f11f88eda77eb4e2ae2aa1ece2862230b12f2/handsontable/src/plugins/hiddenColumns/contextMenuItem/hideColumn.js#L7).
I am marking this issue as a bug and will notify you as soon as we fix it.

adrianspdev:
Hi @footman9988, thanks for reporting! Fix landed in #12586, ships in one of the next releases. Closing issue.

PR #12586, opened 2026-05-15
DEV-1695: Fix context menu object items dropping plugin callbacks

adrianspdev:
### Context

The PR fixes a bug where configuring `contextMenu` with the object form and referencing a plugin-provided key (e.g. `add_child` or `detach_from_parent` from `nestedRows`) produced menu entries with raw keys as labels and no callbacks. Clicking them was a no-op.

Example that reproduces the issue (see [jsfiddle from #9894](https://jsfiddle.net/footman8899/t7jwy6qc/10/)):

```js
new Handsontable(container, {
  nestedRows: true,
  contextMenu: {
    items: {
      row_above: {},
      row_below: {},
      detach_from_parent: { name: 'detach_from_parent' },
      add_child: { name: 'add_child' }
    }
  },
  // ...
});
```

Root cause is in `ItemsFactory.setPredefinedItems` (`handsontable/src/plugins/contextMenu/itemsFactory.js`). The flow is:

1. `getItems(settings)` resolves user settings against a registry that does not yet contain plugin-provided entries, so unknown keys become bare placeholders.
2. The `afterContextMenuDefaultOptions` hook fires; the `nestedRows` plugin splices its rich entries (with translated `name`, `callback`, and `disabled`) at the start of the array.
3. `setPredefinedItems` iterates the array and assigns `items[value.key] = value` without checking if a rich entry already exists. The bare placeholder appears later in the array than the spliced rich entry, so it overwrites it.

The array form (`contextMenu: ['add_child']`) survives because `setPredefinedItems` assigns bare placeholders under their numeric index key, avoiding the collision.

The fix is a defensive guard in `setPredefinedItems`: skip the assignment when an existing entry has a callback, submenu, or renderer and the incoming entry has none. This preserves the plugin's rich entry while still allowing legitimate later overrides (which carry their own callback or richer fields).

### How has this been tested?

Added E2E tests:

- `handsontable/src/plugins/nestedRows/__tests__/nestedRows.spec.js` - new describe block `context menu items defined as object` covering: `add_child: {}`, `detach_from_parent: {}`, override `{ name: 'Custom add child label' }`, user-supplied full override (`{ name, callback }`), and mixed built-in + plugin keys (5 tests). 4 of these fail on `develop` without the fix, locking down the regression.
- `handsontable/src/plugins/customBorders/__tests__/contextMenuItem/borders.spec.js` - object-form `{ borders: {} }` test asserting the entry renders and the submenu opens with Top/Right/Bottom/Left items (positive coverage; `customBorders` pushes to the end of the items array so it does not hit the original bug, but this pins the behavior).
- `handsontable/src/plugins/hiddenColumns/__tests__/contextMenu/hideColumn.spec.js` - object-form `{ hidden_columns_hide: {} }` test (positive coverage, same rationale as borders).

Commands to reproduce:

```bash
npm run test:e2e --prefix handsontable -- --testPathPattern='nestedRows/__tests__/nestedRows'
npm run test:e2e --prefix handsontable -- --testPathPattern='customBorders/__tests__/contextMenuItem'
npm run test:e2e --prefix handsontable -- --testPathPattern='hiddenColumns/__tests__/contextMenu'
npm run test:e2e --prefix handsontable -- --testPathPattern='contextMenu/__tests__'
npm run test:e2e --prefix handsontable -- --testPathPattern='dropdownMenu'
```

Full results: 612 specs pass across the relevant suites (`nestedRows`, `customBorders`, `hiddenColumns`, `contextMenu`, `dropdownMenu`), 0 failures. ESLint clean.

Manual repro verified locally against the jsfiddle setup from the linked GitHub issue: on `v17.0.1` from CDN the menu entries are inert; on the PR build the callbacks fire and rows are added/detached as expected.

### Types of changes

### Related issue(s):

1. #9894
2. DEV-1695

### Affected project(s):

- [x] \`handsontable\`
- [ ] \`@handsontable/angular-wrapper\`
- [ ] \`@handsontable/react-wrapper\`
- [ ] \`@handsontable/vue3\`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-1695

---

> [!NOTE]
> **Medium Risk**
> Touches core context-menu item merging logic used by multiple plugins; regression could change how duplicate keys/overrides are resolved, though behavior is now covered by new E2E tests.
> 
> **Overview**
> Fixes object-form `contextMenu.items` handling so plugin-defined entries (e.g. `nestedRows` `add_child`/`detach_from_parent`, `customBorders` `borders`, `hiddenColumns` `hidden_columns_hide`) no longer lose their translated labels, callbacks, or submenus when placeholders and plugin entries collide.
> 
> `ItemsFactory.setPredefinedItems` now merges duplicate keyed items instead of blindly overwriting, preserving the richer plugin entry while still allowing user overrides (including user-supplied callbacks). Adds E2E coverage for these cases and includes a changelog entry (`.changelogs/12586.json`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a5246548183f6ecc283bfcb4d10eeacb119ed468. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/handsontable/src/plugins/contextMenu/itemsFactory.js
+++ b/handsontable/src/plugins/contextMenu/itemsFactory.js
@@ -57,16 +57,17 @@ export class ItemsFactory {
       } else {
         // Object-form `contextMenu.items` keys that name a plugin-provided
         // entry (e.g. `add_child` from nestedRows) are unknown to the registry
-        // on the first `getItems` pass, so a bare placeholder is emitted. The
-        // nestedRows plugin then splices its rich entry at the start of the
-        // array via `afterContextMenuDefaultOptions`, placing the bare
-        // placeholder later in iteration order. Without this guard, the
-        // placeholder would overwrite the rich entry's callback, submenu, or
-        // renderer. See issue #9894.
-        const existing = items[value.key];
-        const isRich = item => Boolean(item && (item.callback || item.submenu || item.renderer));
+        // on the first `getItems` pass, so a bare placeholder is emitted.
+        // Plugins later splice their rich entries into the same array via
+        // `afterContextMenuDefaultOptions`. Merging preserves both contributions
+        // regardless of iteration order: rich properties (callback, submenu,
+        // disabled, renderer, name function, etc.) survive when the other side
+        // does not define them, and user overrides still take effect via the
+        // final `getItems` pass that extends the user pattern onto the predefined
+        // entry. See issue #9894.
+        if (items[value.key]) {
+          extend(items[value.key], value);
 
-        if (existing && isRich(existing) && !isRich(value)) {
           return;
         }
 
--- a/handsontable/src/plugins/contextMenu/itemsFactory.js
+++ b/handsontable/src/plugins/contextMenu/itemsFactory.js
@@ -55,6 +55,21 @@ export class ItemsFactory {
         menuItemKey = value.key;
 
       } else {
+        // Object-form `contextMenu.items` keys that name a plugin-provided
+        // entry (e.g. `add_child` from nestedRows) are unknown to the registry
+        // on the first `getItems` pass, so a bare placeholder is emitted. The
+        // nestedRows plugin then splices its rich entry at the start of the
+        // array via `afterContextMenuDefaultOptions`, placing the bare
+        // placeholder later in iteration order. Without this guard, the
+        // placeholder would overwrite the rich entry's callback, submenu, or
+        // renderer. See issue #9894.
+        const existing = items[value.key];
+        const isRich = item => Boolean(item && (item.callback || item.submenu || item.renderer));
+
+        if (existing && isRich(existing) && !isRich(value)) {
+          return;
+        }
+
         items[value.key] = value;
         menuItemKey = value.key;
       }
--- a/.changelogs/12586.json
+++ b/.changelogs/12586.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed context menu items in object form (e.g. `add_child`, `detach_from_parent`) not rendering plugin-provided callbacks and translated labels.",
+  "type": "fixed",
+  "issueOrPR": 12586,
+  "breaking": false,
+  "framework": "none"
+}
--- a/handsontable/src/plugins/customBorders/__tests__/contextMenuItem/borders.spec.js
+++ b/handsontable/src/plugins/customBorders/__tests__/contextMenuItem/borders.spec.js
@@ -72,5 +72,45 @@ describe('ContextMenu', () => {
 
       expect(readOnlyItem.hasClass('htDisabled')).toBe(true);
     });
+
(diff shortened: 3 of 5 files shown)


qunabu:
Task linked: [DEV-1695 How to add 'add_child' or 'detach_from_parent' in Context menu custom configuration](https://app.clickup.com/t/9015210959/DEV-1695)

adrianspdev:
https://github.com/handsontable/handsontable/actions/runs/25913081349

handsontable/src/plugins/contextMenu/itemsFactory.js
```diff
@@ -55,6 +55,21 @@ export class ItemsFactory {
...
+        // entry (e.g. `add_child` from nestedRows) are unknown to the registry
+        // on the first `getItems` pass, so a bare placeholder is emitted. The
+        // nestedRows plugin then splices its rich entry at the start of the
+        // array via `afterContextMenuDefaultOptions`, placing the bare
+        // placeholder later in iteration order. Without this guard, the
+        // placeholder would overwrite the rich entry's callback, submenu, or
+        // renderer. See issue #9894.
+        const existing = items[value.key];
+        const isRich = item => Boolean(item && (item.callback || item.submenu || item.renderer));
+
```
budnix, line 70: `isRich` will silently miss any plugin entry that has `name`/`disabled`/`hidden` but no `callback`, `submenu`, or `renderer` — nothing enforces that invariant. The fix also leans on iteration order: nestedRows splices at index 0, so the rich entry always wins. A plugin using `push` flips the order and the guard stops helping.

`extend` (already imported) removes the heuristic and makes this order-independent:

```suggestion
        // Object-form `contextMenu.items` keys that name a plugin-provided
        // entry (e.g. `add_child` from nestedRows) are unknown to the registry
        // on the first `getItems` pass, so a bare placeholder is emitted.
        // Plugins later splice their rich entries into the same array via
        // `afterContextMenuDefaultOptions`. Merging preserves both contributions
        // regardless of iteration order: rich properties (callback, submenu,
        // disabled, renderer, name function, etc.) survive when the other side
        // does not define them, and user overrides still take effect via the
        // final `getItems` pass that extends the user pattern onto the predefined
        // entry. See issue #9894.
        if (items[value.key]) {
          extend(items[value.key], value);

          return;
        }
```

Whichever entry arrives first lands in the map; the second one merges into it. Properties only one side defines survive unchanged. User overrides (custom `name`, user-supplied `callback`) win anyway because the second `getItems` pass extends the user pattern on top. All five existing tests pass with this version.

budnix (changes requested):

handsontable/src/plugins/contextMenu/itemsFactory.js
adrianspdev: great catch, thanks

budnix (approved):

PR #12610, opened 2026-05-19
DEV-1717: Fix NestedHeaders rendering with manualColumnMove + hiddenColumns

### LINKED ISSUE #9565: Table Headers renders Incorrectly. Nested Headers / Columns / Manual Column / Hidden Columns
### Description
When having a nested header, two level, and having manual column move and hidden columns, the tables render incorrectly the headers

### Steps to reproduce, repro on any version after handsontable 8.0
https://jsfiddle.net/Resems/tje4z0c8/

image

### Workaround, the number of `columns` should be the same as actual visible columns ( in this case 6 ), but this breaks additional logic of reading and saving the column manual move array ( saving the table settings into a db )
https://jsfiddle.net/Resems/tje4z0c8/2/

### Working Version on v7 handsontable
https://jsfiddle.net/Resems/tje4z0c8/1/
image

### Demo
https://jsfiddle.net/Resems/tje4z0c8/

### Your environment
* Handsontable version: 12+
* Browser name and version: Any
* Operating system: Any

While there is a workaround, it is not working for us, because internally handsontable changes the manual column move array to only 6 columns ( we need to preserve the column move and hidden column for persistent state for each table created by the user ).
So if the user hides/moves columns, after refreshing the browser and unhiding the columns, we must display the original original of 7 columns ( internally we sync the column move from handsontable so the original order is lost we use only visible columns. )

adrianspdev:
### Context

The PR fixes [GH issue #9565](https://github.com/handsontable/handsontable/issues/9565). When `nestedHeaders`, `manualColumnMove`, and `hiddenColumns` were combined, the multi-column parent header at the moved-from-hidden position disappeared entirely and the wrong leaf was marked hidden in the bottom row.

Root cause: the plugin pulled hidden-column state from `hidingMapsCollection.getMergedValues()` (array indexed by **physical** column) and forwarded each index to `triggerColumnModification`, which the state manager documents as expecting a **visual** column index (see `stateManager/index.js`). Without `manualColumnMove`, physical == visual and the bug is invisible. With a move active, the tree marks the wrong leaf as hidden and the multi-column root anchored at the actually-hidden visual position is never rendered.

Three changes:

1. New private `#syncHiddenColumnsFromMapper()` helper translates physical → visual via `getVisualFromPhysicalIndex` before calling `triggerColumnModification`. Replaces the buggy inline forEach in `updatePlugin()`.
2. The runtime hiding observer in `updatePlugin()` now translates each `replace` change's index from physical → visual and skips changes where translation returns null (trimmed columns).
3. New `#onColumnIndexMapperCacheUpdated` listener registered on the column index mapper in `enablePlugin()` (and unregistered in `disablePlugin()`) re-runs the sync helper when `indexesSequenceChanged` is true. This covers the runtime `manualColumnMove` case where the hiding observer would not fire on its own.

A pre-existing test bug was also fixed in `generalFunctionality.spec.js` where `rowMapperHooks` was being captured from `columnIndexMapper()`. Pre-fix this masked an asymmetry between row and column hook counts.

### How has this been tested?

Three new E2E specs added in `nestedHeaders/__tests__/hidingColumns.spec.js` covering:

- init-time settings with move + hide together
- runtime hide via custom hiding index map while `manualColumnMove` is active
- runtime move via `manualColumnMove` plugin while `hiddenColumns` already set

Manual verification via demo pages (gitignored): `handsontable/dev-latest.html` (released v17.0.1) vs `handsontable/dev-pr.html` (local build) — all three cases reproduce the bug on released and render correctly on the PR build.

```
npm run test:e2e --prefix handsontable --testPathPattern=nestedHeaders
# 141 specs, 0 failures

npm run test:e2e --prefix handsontable --testPathPattern=manualColumnMove
# 210 specs, 0 failures

npm run test:e2e --prefix handsontable --testPathPattern=hiddenColumns
# 289 specs, 0 failures

npm run test:unit --prefix handsontable -- --testPathPattern=nestedHeaders
# 132 passed
```

### Types of changes

### Related issue(s):

1. #9565
2. DEV-1717

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-1717

---

> [!NOTE]
> **Medium Risk**
> Touches NestedHeaders’ core hide/show synchronization and adds a new mapper hook, so regressions are possible in column hiding/moving interactions across plugins.
> 
> **Overview**
> Fixes a NestedHeaders rendering bug when `hiddenColumns` is used alongside `manualColumnMove` by translating *physical* hidden-column indexes to *visual* indexes before updating the nested-header tree, including during hiding-map change events.
> 
> Adds a `columnIndexMapper` `cacheUpdated` listener to re-sync hidden state when column order changes (so moves without hide-value changes don’t leave stale tree state), and updates tests/fixtures to cover these scenarios plus correct a mapper-hook-count assertion; includes a changelog entry.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6826d9938ed020720dd68dbf03f2234937622dc0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changelogs/12610.json
+++ b/.changelogs/12610.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed nested headers rendering incorrectly when manualColumnMove and hiddenColumns are combined.",
+  "type": "fixed",
+  "issueOrPR": 12610,
+  "breaking": false,
+  "framework": "none"
+}
--- a/handsontable/src/plugins/nestedHeaders/__tests__/generalFunctionality.spec.js
+++ b/handsontable/src/plugins/nestedHeaders/__tests__/generalFunctionality.spec.js
@@ -698,7 +698,7 @@ describe('NestedHeaders', () => {
         height: 300,
       });
 
-      const rowMapperHooks = columnIndexMapper().__localHooks.cacheUpdated.length;
+      const rowMapperHooks = rowIndexMapper().__localHooks.cacheUpdated.length;
       const columnMapperHooks = columnIndexMapper().__localHooks.cacheUpdated.length;
 
       await updateSettings({
--- a/handsontable/src/plugins/nestedHeaders/__tests__/hidingColumns.spec.js
+++ b/handsontable/src/plugins/nestedHeaders/__tests__/hidingColumns.spec.js
@@ -1849,5 +1849,153 @@ describe('NestedHeaders', () => {
           `);
       });
     });
+
+    describe('with cooperation with the ManualColumnMove plugin (#9565)', () => {
+      it('should render multi-column parent header correctly when a hidden column ' +
+        'maps to a different visual index due to manualColumnMove (initial settings)', async() => {
+        handsontable({
+          data: createSpreadsheetData(2, 7),
+          colHeaders: true,
+          nestedHeaders: [
+            ['A', 'B', 'C', { label: 'PARENT', colspan: 3 }, 'G'],
+            ['A', 'B', 'C', 'D', 'E', 'F', 'G'],
+          ],
+          manualColumnMove: [3, 0, 1, 2, 4, 5, 6],
+          hiddenColumns: {
+            columns: [2],
+          },
+        });
+
+        expect(extractDOMStructure(getTopClone(), getMaster())).toMatchHTML(`
+          <thead>
+            <tr>
+              <th class="">A</th>
+              <th class="">B</th>
+              <th class="">C</th>
+              <th class="" colspan="2">PARENT</th>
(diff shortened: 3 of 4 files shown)


qunabu:
Task linked: [DEV-1717 Table Headers renders incorrectly. Nested Headers / Columns / Manual Column / Hidden Columns](https://app.clickup.com/t/9015210959/DEV-1717)

budnix (approved):

PR #12653, opened 2026-05-28
DEV-1791: Reload page once on stale-chunk import failure in docs assistant

qunabu:
### Context

The PR fixes a recurring Sentry error (`HANDSONTABLE-DOCS-1DQ`) in the docs assistant widget. When a new version of the docs site is deployed, Astro/Vite replaces old content-hashed JS chunks (e.g. `DocsAssistantWidget.D2fw3ORt.js`) with new ones. Users who still have the old HTML cached in their browser attempt to dynamically import the old chunk URL, which no longer exists on the CDN. This causes an unhandled `TypeError: Failed to fetch dynamically imported module` that surfaces as a Sentry error.

The fix detects this class of error in the bootstrap catch handler and reloads the page once. The reload fetches fresh HTML containing the updated chunk hashes. A `sessionStorage` flag (`docs-assistant-chunk-reload`) prevents an infinite reload loop if the server itself is broken. The flag is cleared on every successful mount so future deployments can trigger a reload again.

### How has this been tested?

The fix targets a production-only failure path that requires a stale cache + new deployment to reproduce naturally. The logic was verified by code review:

- `isChunkLoadError` covers all three major browser error messages (Chrome, Firefox, Safari)
- The `sessionStorage` guard is wrapped in try/catch to handle private browsing mode
- The flag is cleared on successful mount so the guard resets between deployments
- No new dependencies introduced

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature or improvement (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Additional language file or change to the existing one (translations)

### Related issue(s):

1. DEV-1791

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/angular-wrapper`
- [ ] `@handsontable/react-wrapper`
- [ ] `@handsontable/vue3`

### Checklist:

- [x] I have reviewed the guidelines about [Contributing to Handsontable](https://github.com/handsontable/handsontable/blob/master/CONTRIBUTING.md) and I confirm that my code follows the code style of this project.
- [x] I have signed the [Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLScpMq4swMelvw3-onxC8Jl29m0fVp5hpf7d1yQVklqVjGjWGA/viewform?c=0&w=1)
- [ ] My change requires a change to the documentation.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-1791
[...]

PR #12619, opened 2026-05-19
DEV-1709, DEV-1710: Fix iPhone SE mobile menu and Ask AI

demtario:
### Context

Fixes two docs-site issues on legacy WebKit (iPhone SE 2022 on BrowserStack).

**DEV-1709 - Sidebar and mobile nav active state**

Starlight marks the current page with `aria-current="page"` and applies `background-color: var(--sl-color-text-accent)` (solid brand blue in light mode). On iPhone SE that produced a heavy blue fill and poor contrast for active items in the left sidebar, mobile hamburger menu, and ToC panel.

This PR removes the active background fill (`transparent`) and keeps the left accent bar. It also adds `@supports not (color: oklch(...))` hex fallbacks for gray palette tokens used by mobile nav hover/active styles.

**DEV-1710 - Ask AI assistant on old Safari**

Three separate problems were addressed:

1. **Stale header button listeners** - `DocsAssistantWidget` bound clicks on mount, but Starlight replaces header DOM on `astro:after-swap`. Header.astro now calls `docsAssistantEnsureMounted()` and dispatches `docs-assistant:toggle`; the widget listens for that event. `touchend` is wired for iOS.
2. **Lost widget after client navigation** - Bootstrap remounts React on `astro:page-load` when `#docs-assistant-root` is missing and exposes `window.docsAssistantEnsureMounted`.
3. **Parse failure on WebKit before Safari 16.4** - Dynamic import of the assistant chunk failed with `SyntaxError: Invalid regular expression: invalid group specifier name` because `react-markdown` / `remark-gfm` / `shiki` ship RegExp syntax old Safari cannot parse at module load time (Chrome is unaffected). The markdown stack is split into lazy chunks: modern browsers load `MarkdownRendererFull` (GFM + Shiki); older WebKit loads `MarkdownRendererSimple` (escaped text, fenced code blocks, links) via `supportsModernRegex()` in `docs/src/lib/supports-modern-regex.mjs`.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-1709
ClickUp task: https://app.clickup.com/t/9015210959/DEV-1710

### How has this been tested?

- `npm run docs:lint --prefix docs`
- `node --test src/lib/__tests__/supports-modern-regex.test.mjs` (from `docs/`)
- Manual verification on BrowserStack iPhone SE 2022 (Netlify preview):
  - Left sidebar: active item has no blue fill, readable text, left accent bar visible
[...]

PR #12745, opened 2026-06-10
fix(docs): open docs assistant on first Ask AI click

qunabu:
### Context

The PR fixes a regression where the docs assistant panel does not open on the first click of the **Ask AI** button — it opens only on the second click.

Root cause is a mount/listener race in the lazy-mounted widget:

- The widget registers its `docs-assistant:toggle` listener inside a React `useEffect`, which runs *after* `root.render()` resolves.
- `docs-assistant-bootstrap.ts` resolved the mount promise as soon as `root.render()` was called, and `Header.astro` dispatched the first `docs-assistant:toggle` in the `.then()` right after.
- That first toggle fired before the widget's listener existed, so it was dropped. The panel stayed closed until a second click.

Fix:

- The widget dispatches a `docs-assistant:ready` event once its toggle listener is attached.
- The bootstrap subscribes to `docs-assistant:ready` *before* calling `root.render()` and waits for it (with a 3s safety timeout) before resolving the mount. The first click now reliably opens the panel.

The chunk-reload guard added in #12653 is preserved.

### How has this been tested?

- Added `docs/tests/docsAssistant.spec.ts` (Playwright): loads a docs page, clicks `#header-assistant-btn` once, asserts `.da-panel[data-open="true"]`, then asserts a second click closes it. Run with a docs server up: `BASE_URL=<url> npm run docs:visual-test docsAssistant.spec.ts`.
- Manual: open any docs page, click **Ask AI** once — the panel slides in.

### Types of changes

### Related issue(s):

1. Regression in the docs assistant lazy-mount path (introduced with #12619)

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`
- [x] `handsontable-documentation`

### Checklist:

- [x] I have reviewed the guidelines about Contributing to Handsontable and I confirm that my code follows the code style of this project.
- [x] I have signed the Contributor License Agreement

[skip changelog] — docs-site bug fix; no library source change.

---

> [!NOTE]
> **Low Risk**
> Docs-site UI bootstrap timing only; no auth, data, or library API changes.
> 
> **Overview**
> Fixes a **first-click** regression on the docs **Ask AI** button: the lazy-mounted assistant panel stayed closed until a second click because `Header.astro` dispatched `docs-assistant:toggle` as soon as `root.render()` finished, before the widget’s React effect registered its toggle listener.
> 
> The widget now fires **`docs-assistant:ready`** right after that listener is attached. **`docs-assistant-bootstrap.ts`** subscribes for that event *before* `root.render()`, **`await`s** it (with a **3s** fallback timeout), and only then resolves `docsAssistantEnsureMounted()` so the header’s post-mount toggle is not dropped.
> 
> Adds **`docs/tests/docsAssistant.spec.ts`** (Playwright) to assert one click on `#header-assistant-btn` opens `.da-panel[data-open="true"]` and a second click closes it.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0a1d412321b2ec9864bafa4a3ebe3499e8330e81. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/src/components/DocsAssistant/DocsAssistantWidget.tsx
+++ b/docs/src/components/DocsAssistant/DocsAssistantWidget.tsx
@@ -67,6 +67,11 @@ export function DocsAssistantWidget() {
   useEffect(() => {
     const onToggle = () => setOpen((v) => !v);
     window.addEventListener('docs-assistant:toggle', onToggle);
+    // Tell the bootstrap the toggle listener is now attached. The widget is
+    // mounted lazily on the first button click, and the bootstrap waits for
+    // this signal before dispatching that first toggle — otherwise the click
+    // would be lost to the mount/listener race and the panel would not open.
+    window.dispatchEvent(new Event('docs-assistant:ready'));
     return () => window.removeEventListener('docs-assistant:toggle', onToggle);
   }, []);
 
--- a/docs/src/scripts/docs-assistant-bootstrap.ts
+++ b/docs/src/scripts/docs-assistant-bootstrap.ts
@@ -11,6 +11,11 @@ import type { Root } from 'react-dom/client';
 
 const CONTAINER_ID = 'docs-assistant-root';
 const CHUNK_RELOAD_FLAG = 'docs-assistant-chunk-reload';
+const READY_EVENT = 'docs-assistant:ready';
+
+// Safety cap so a failed mount never leaves the trigger button permanently
+// unresponsive while waiting for the ready signal.
+const READY_TIMEOUT = 3000;
 
 let root: Root | null = null;
 let mountInFlight: Promise<void> | null = null;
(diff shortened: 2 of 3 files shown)


GreenFlux:
Thanks for picking this up. After investigating on this PR's deploy preview, I think the first-click race addressed here is a real but separate issue, and the reported "Ask AI button doesn't open" bug has a different root cause that this PR doesn't reach — the button is still dead on this PR's own DP.

What I found on the DP:

- The widget mounts fine, and dispatching `docs-assistant:toggle` by hand opens the panel — the bootstrap/widget layer is healthy.
- A real button click never produces a toggle, because `initMobileNav()` in `Header.astro` never runs. #12656 moved its trigger to `astro:page-load`, but this docs site has no Astro ClientRouter, so that event never fires. The whole header script stays inert (Ask AI binding, mobile menu, ToC, mobile search delegation, and the fixed-panel relocation).
- Manually firing `astro:page-load` makes `initMobileNav` run and the button work, confirming the listener is registered but the event never arrives.

Because the break is upstream of the mount/listener timing, the `docs-assistant:ready` handshake here can't fix it (no toggle is ever dispatched). It is a reasonable hardening for the genuine "click during lazy mount" race, though, and worth keeping on its own.

I opened #12749 with the fix: bind the header handlers on `DOMContentLoaded` (independent of `astro:page-load`) and defer only the panel relocation to `load`, which preserves the Preact hydration-crash fix from #12656. It adds a Playwright test that fails on the current deploy and passes with the fix.

Suggestion: close this in favor of #12749, or repurpose it to land just the `docs-assistant:ready` handshake. Your call.

PR #12763, opened 2026-06-12
DEV-1865 - Add typed handsontable ColumnSettings property

KrzysztofZie:
### Context
  The handsontable cell-type configuration is now an explicitly typed property on GridSettings, so it can be set both globally and per-column (previously it was only on ColumnSettings, which made global usage a
  TypeScript error). getValue is now a typed top-level grid setting too — read by Core#getValue, it accepts a row-property name or a function whose this is bound to the instance — so it type-checks both at the
  instance level and inside the nested handsontable config. The redundant [key: string]: unknown index signature on ColumnSettings was removed: it duplicated the [key: string]: any already inherited from
  GridSettings, and that value-type mismatch was what made TypeScript drop the this binding on getValue; arbitrary custom keys still work via the inherited signature. This is a types-only change with no runtime impact.

### How has this been tested?
handsontable npm run test:types
npm run typecheck --prefix docs/angular-type-check

### Types of changes

### Related issue(s):
1.
2.
3.

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-1766

---

> [!NOTE]
> **Low Risk**
> Declaration-only TypeScript and test changes with no runtime or behavioral impact.
> 
> **Overview**
> Adds explicit TypeScript support for the **`handsontable` cell type** nested-grid config on both **`GridSettings`** (table-wide) and **`ColumnSettings`** (per column), via a new **`HandsontableCellTypeSettings`** type that mirrors grid options plus optional **`getValue`** (property name or function with **`this`** bound to the nested instance).
> 
> **`ColumnSettings`** re-declares **`handsontable`** so contextual typing keeps **`getValue`**’s **`this`** instead of widening through the **`[key: string]: unknown`** index signature. Compile-time coverage is added in **`settings.types.ts`**; a changelog entry documents the change. **No runtime behavior changes.**
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 634eb45871e7e2542334e4ca7157a1d9dd068bb6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/handsontable/src/__tests__/core/settings.types.ts
+++ b/handsontable/src/__tests__/core/settings.types.ts
@@ -120,6 +120,13 @@ const allSettings: Required<Handsontable.GridSettings> = {
   },
   fragmentSelection: oneOf(true, 'cell'),
   headerClassName: 'htCenter test',
+  handsontable: {
+    data: [[]],
+    colHeaders: true,
+    getValue: oneOf('name', function(this: Handsontable) {
+      return this.getDataAtCell(0, 0);
+    }),
+  },
   hashLength: 8,
   hashRevealDelay: 1000,
   hashSymbol: '#',
@@ -784,3 +791,20 @@ const _rendererCellProps: Handsontable.CellProperties = cellProperties;
 
 // Assert that a CellProperties value is assignable as the validator `this` context type
 const _validatorCellProps: CellProperties = cellProperties;
+
+// === `handsontable` cell-type setting typing ===
+// The nested-grid config is exposed on both grid settings (global) and column settings (per-column),
+// and `getValue` accepts either a source property name or a function bound to the nested instance.
+const _hotGlobal: Handsontable.GridSettings = {
+  handsontable: { data: [[]], getValue: 'name' },
+};
+const _hotColumn: Handsontable.ColumnSettings = {
+  handsontable: { data: [[]], getValue: 'name' },
+};
+const _hotColumnGetValueFn: Handsontable.ColumnSettings = {
+  handsontable: {
+    getValue(this: Handsontable) {
+      return this.getDataAtCell(0, 0);
+    },
+  },
+};
--- a/handsontable/src/__tests__/core/settings.types.ts
+++ b/handsontable/src/__tests__/core/settings.types.ts
@@ -120,12 +120,13 @@ const allSettings: Required<Handsontable.GridSettings> = {
   },
   fragmentSelection: oneOf(true, 'cell'),
   headerClassName: 'htCenter test',
+  getValue: oneOf('name', function(this: Handsontable) {
+    return this.getDataAtCell(0, 0);
+  }),
   handsontable: {
     data: [[]],
     colHeaders: true,
-    getValue: oneOf('name', function(this: Handsontable) {
-      return this.getDataAtCell(0, 0);
-    }),
+    getValue: 'name',
   },
   hashLength: 8,
   hashRevealDelay: 1000,
@@ -793,8 +794,19 @@ const _rendererCellProps: Handsontable.CellProperties = cellProperties;
 const _validatorCellProps: CellProperties = cellProperties;
 
 // === `handsontable` cell-type setting typing ===
-// The nested-grid config is exposed on both grid settings (global) and column settings (per-column),
-// and `getValue` accepts either a source property name or a function bound to the nested instance.
+// `getValue` is a regular top-level grid setting (read by `Core#getValue`), usable both at the
+// instance level and inside the nested `handsontable` config of the `handsontable` cell type. Its
+// function form binds `this` to the instance.
+
+// `getValue` also works as a top-level grid setting (read by `Core#getValue`), not only nested.
+const _hotTopLevelGetValueString: Handsontable.GridSettings = { getValue: 'name' };
+const _hotTopLevelGetValueFn: Handsontable.GridSettings = {
+  getValue() {
+    return this.getDataAtCell(0, 0);
+  },
+};
+
+// Nested `handsontable` config (global and per-column), with both `getValue` forms.
 const _hotGlobal: Handsontable.GridSettings = {
   handsontable: { data: [[]], getValue: 'name' },
 };
@@ -804,10 +816,15 @@ const _hotColumn: Handsontable.ColumnSettings = {
 const _hotColumnGetValueFn: Handsontable.ColumnSettings = {
   handsontable: {
     // Relies on the *contextual* `this` (no explicit annotation) on purpose: this fails to compile
-    // if the direct `handsontable` re-declaration on `ColumnSettings` is removed, because the
-    // `Omit<GridSettings, 'data'>` + index-signature combination would widen `this` to `{}`.
+    // if a `[key: string]: unknown` index signature is re-added to `ColumnSettings`, because the
+    // value-type mismatch with the `[key: string]: any` inherited from `GridSettings` widens `this`
+    // to `{}`. Keeps the no-index-signature decision in `settings.ts` honest.
     getValue() {
       return this.getDataAtCell(0, 0);
     },
   },
 };
+
+// Custom plugin/meta keys still allowed via the `[key: string]: any` signature inherited from `GridSettings`.
+const _columnArbitraryKeys: Handsontable.ColumnSettings = { someCustomPluginKey: 123 };
+const _cellMetaArbitraryKeys: Handsontable.CellMeta = { someCustomMetaKey: true };
--- a/handsontable/src/__tests__/core/settings.types.ts
+++ b/handsontable/src/__tests__/core/settings.types.ts
@@ -803,7 +803,10 @@ const _hotColumn: Handsontable.ColumnSettings = {
 };
 const _hotColumnGetValueFn: Handsontable.ColumnSettings = {
   handsontable: {
-    getValue(this: Handsontable) {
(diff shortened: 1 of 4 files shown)


handsontable/src/settings.ts
```diff
@@ -57,6 +57,15 @@ export type { GridSettings } from './core/settings';
...
 export interface ColumnSettings extends Omit<GridSettings, 'data'> {
   data?: string | number | ColumnDataGetterSetterFunction;
+  /**
+   * Configuration of the nested grid used by the `handsontable` cell type. It accepts the same
+   * grid settings as the outer table, plus an optional `getValue` that controls which value is
+   * pulled back into the edited cell. It can be a property name of the focused row, or a function
+   * whose `this` is bound to the nested Handsontable instance (matching `Core#getValue`).
+   */
+  handsontable?: GridSettings & {
+    getValue?: string | ((this: Handsontable) => CellValue);
```
budnix: All the other cell-type-specific options — `source`, `dateFormat`, `numericFormat`, `selectOptions` — are defined directly on `GridSettings` in `core/settings.ts`, and `ColumnSettings` inherits them via `extends`. Putting `handsontable` only on `ColumnSettings` means passing `handsontable: { data: [...] }` at the top-level table settings (rather than per-column) is valid at runtime but causes a TypeScript error. Worth moving this to `GridSettings` to stay consistent with those other cell-type options and cover the global config case.

handsontable/src/settings.ts
```diff
@@ -57,6 +57,15 @@ export type { GridSettings } from './core/settings';
...
+  };
```
budnix: This is new public API surface and there are no type tests covering it yet — if someone later changes the `getValue` signature, renames the property, or removes it, `npm run test:types` won't catch it. Worth adding a couple of assertions to `src/__tests__/core/settings.types.ts`, something like:

```ts
const col: ColumnSettings = { handsontable: { data: [], getValue: 'name' } };
const col2: ColumnSettings = { handsontable: { getValue(this: Handsontable) { return this.getDataAtCell(0, 0); } } };
```

handsontable/src/core/settings.ts
```diff
@@ -69,6 +69,15 @@ export interface GridSettings {
...
   type?: string;
+  /**
+   * Configuration of the nested grid used by the `handsontable` cell type. It accepts the same
+   * grid settings as the outer table, plus an optional `getValue` that controls which value is
+   * pulled back into the edited cell. It can be a property name of the focused row, or a function
+   * whose `this` is bound to the nested Handsontable instance (matching `Core#getValue`).
+   */
+  handsontable?: GridSettings & {
+    getValue?: string | ((this: HotInstance) => CellValue);
+  };
```
budnix:
```suggestion
  handsontable?: GridSettings;
  getValue?: string | ((this: HotInstance) => CellValue);
```

handsontable/src/settings.ts
```diff
@@ -57,6 +57,17 @@ export type { GridSettings } from './core/settings';
...
+   * Configuration of the nested grid used by the `handsontable` cell type — see {@link GridSettings.handsontable}.
+   *
+   * Re-declared here (identical to the inherited definition) on purpose. `ColumnSettings` combines
+   * the `[key: string]: unknown` index signature below with the `Omit<GridSettings, 'data'>` mapped
+   * type it extends; that combination makes TypeScript widen the *inherited* `handsontable` to the
+   * index signature when contextually typing an object literal, which drops the `this` binding on
+   * `getValue`. A direct property declaration takes precedence over the index signature and keeps
+   * the precise type intact.
+   */
+  handsontable?: GridSettings['handsontable'];
```
budnix: The field is already defined in `GridSettings`; I don't think it's needed here anymore 🤔 
```suggestion
```

budnix (changes requested):

handsontable/src/settings.ts
KrzysztofZie: Good instinct, but it's unfortunately still required. ColumnSettings combines its [key: string]: unknown index signature with the Omit<GridSettings, 'data'> it extends, and that specific combination makes  ▎ TS widen the inherited handsontable to the index signature when contextually typing an object literal — which silently drops the this binding on getValue (it becomes {}). That's exactly what broke the
  Angular docs examples earlier.
  Repro: removing this line makes npm run test:types fail with Property 'getDataAtCell' does not exist on type '{}' at the new _hotColumnGetValueFn assertion. I updated that assertion to rely on the
  contextual this (no explicit annotation) precisely so CI guards this line from being removed. The global case on GridSettings works without a re-declaration (no Omit, no widening); only ColumnSettings 
  needs it.
  
  ▎ Verified: Omit alone ✅, index signature alone ✅, both together ❌ — so the direct declaration is the minimal targeted fix.

handsontable/src/core/settings.ts
```diff
@@ -69,6 +69,10 @@ export interface GridSettings {
   cells?: (row: number, column: number, prop: string | number) => object;
   source?: unknown[] | ((query: string, callback: (items: unknown[]) => void) => void);
   type?: string;
+  /**
+   * Configuration of the nested grid used by the `handsontable` cell type — see {@link HandsontableCellTypeSettings}.
+   */
+  handsontable?: HandsontableCellTypeSettings;
```
budnix: Curious whether the separate `HandsontableCellTypeSettings` type is needed at all. `getValue` isn't really tied to the nested grid — `Core#getValue()` reads `tableMeta.getValue` on any instance (core.ts:3391), and the internal `tableMeta` type already lists it as a normal top-level field (core.ts:462), so it acts like a regular grid setting. If we put it straight on `GridSettings`, the global case works too (`new Handsontable(el, { getValue() {...} })` now type-checks), and both this extra interface and its re-export in `settings.ts` can go. I ran `test:types` and it stays green with this:

```ts
  /**
   * Configuration of the nested grid used by the `handsontable` cell type.
   */
  handsontable?: GridSettings;
  /**
   * Controls which value `Core#getValue` returns for the focused cell — a row property name, or a
   * function whose `this` is bound to the instance. Used by the `handsontable` cell type to pull a
   * value from the nested grid back into the edited cell.
   */
  getValue?: string | ((this: HotInstance) => CellValue);
```

handsontable/src/settings.ts
```diff
@@ -51,12 +51,23 @@ export type ChangeSource = 'auto' | 'edit' | 'loadData' | 'updateData' | 'popula
...
```
budnix: This re-declaration works, but it only hides the problem — the real cause is the `[key: string]: unknown` index signature a few lines below. `ColumnSettings` already gets a `[key: string]: any` from `GridSettings` through `Omit`, so this second one is a duplicate, just with a different value type. That mismatch is what makes TypeScript drop the `this` binding on `getValue`, which is why the re-declaration looked like it was needed. If you delete the `[key: string]: unknown` line instead, extra keys still work (they come from `GridSettings`), the `this` binding stays, and this re-declaration is no longer needed. I checked with `test:types` — the whole block becomes just this, and `_hotColumnGetValueFn` still passes:

```ts
export interface ColumnSettings extends Omit<GridSettings, 'data'> {
  data?: string | number | ColumnDataGetterSetterFunction;
}
```

Worth leaving a short note so no one adds the index signature back later — it quietly brings the bug back.

handsontable/src/__tests__/core/settings.types.ts
```diff
@@ -784,3 +791,23 @@ const _rendererCellProps: Handsontable.CellProperties = cellProperties;
...
+// === `handsontable` cell-type setting typing ===
+// The nested-grid config is exposed on both grid settings (global) and column settings (per-column),
+// and `getValue` accepts either a source property name or a function bound to the nested instance.
+const _hotGlobal: Handsontable.GridSettings = {
+  handsontable: { data: [[]], getValue: 'name' },
+};
+const _hotColumn: Handsontable.ColumnSettings = {
+  handsontable: { data: [[]], getValue: 'name' },
+};
+const _hotColumnGetValueFn: Handsontable.ColumnSettings = {
```
budnix, line 816: Since this PR also makes `getValue` a top-level setting and drops the `ColumnSettings` index signature, it'd be good to add a few more assertions right here to lock both in — top-level `getValue` (string and function form), plus a check that custom keys still work without the index signature:

```ts
// `getValue` also works as a top-level grid setting (read by `Core#getValue`), not only nested.
const _hotTopLevelGetValueString: Handsontable.GridSettings = { getValue: 'name' };
const _hotTopLevelGetValueFn: Handsontable.GridSettings = {
  getValue() {
    return this.getDataAtCell(0, 0);
  },
};

// Custom plugin/meta keys still allowed via the signature inherited from `GridSettings`.
const _columnArbitraryKeys: Handsontable.ColumnSettings = { someCustomPluginKey: 123 };
const _cellMetaArbitraryKeys: Handsontable.CellMeta = { someCustomMetaKey: true };
```

The custom-key checks matter most — they make sure dropping the index signature didn't accidentally stop people from adding their own keys.

budnix (commented):
Three notes on the `handsontable`/`getValue` typing — scoping `getValue`, the actual root cause behind the `ColumnSettings` re-declaration, and a couple of type tests to lock the behavior in.

budnix (approved):
Cool 💪

Issue #9222, opened 2022-02-18
`Filter_by_value` does not show in the dropdown menu if `---------` (separator) is defined as the first object


AMBudnik:
### Description

`Filter_by_value` does not show in the dropdown menu if `---------` (separator) is defined as the first object. Using the separator further in the set does not cause this issue.

The issue is replicable when we use the array declaration the same as when we use the `items` object within the `dropdownMenu`.

### Steps to reproduce

1. Open any dropdown menu of a column to check if you see the `filer by value` filter

### Demo

https://jsfiddle.net/kytxqhj6/4/ - works well if the separator is not defined as the first element
https://jsfiddle.net/kytxqhj6/3/ - does not work if the separator is defined as the first element

### Your environment
* Handsontable version: 11.1
_It does not seem to be a regression_
* Browser name and version: Chrome 97
* Operating system: macOS Monterey

Inform [Zen](https://handsontable.zendesk.com/agent/tickets/28040)

KrzysztofZie:
Fixed.
https://github.com/handsontable/handsontable/pull/12781
The next release of the library will already include the fixed

PR #12781, opened 2026-06-16
DEV-312 - `Filter_by_value` does not show in the dropdown m…

KrzysztofZie:
### Context

Fixed the incorrect hiding of filter menu components (filter_by_value, etc.) when dropdownMenu starts with a '---------' separator — component indexes are now computed based on the actually rendered menu rows, rather than the raw list with uncleaned separators.

### How has this been tested?
npm run test:e2e -- --testPathPattern=filtersUI

### Types of changes

### Related issue(s):
1. #9222
2. DEV-312
3.

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-312

---

> [!NOTE]
> **Low Risk**
> Targeted bug fix in dropdown menu index resolution and filter UI visibility, with regression tests and no API changes.
> 
> **Overview**
> Fixes **filter_by_value** and related dropdown pieces disappearing when **`dropdownMenu`** is customized with **leading `---------` separators**. Rendered rows are built after **`filterSeparators()`** (and hidden-item filtering), but the Filters plugin still mapped component keys to row indexes on the **raw `menuItems` list**, so **`hiddenRows`** targeted the wrong rows.
> 
> Adds **`Menu.getItemPositionByKey()`**, which resolves indexes from the **rendered** menu source data (with a pre-render fallback). **`Filters.getIndexesOfComponents()`** now uses that API instead of scanning **`menuItems`**. New **filters UI e2e** cases cover one and multiple leading separators; changelog entry included.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3af8e6c5582d0ff2cdcdc01cc80476b89b7a2e2d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/handsontable/src/plugins/filters/filters.ts
+++ b/handsontable/src/plugins/filters/filters.ts
@@ -1529,38 +1529,15 @@ export class Filters extends BasePlugin {
    * @returns {Array}
    */
   getIndexesOfComponents(...components: BaseComponent[]) {
-    const indexes: number[] = [];
-
-    if (!this.dropdownMenuPlugin) {
-      return indexes;
-    }
-
-    const menu = this.dropdownMenuPlugin.menu;
+    const menu = this.dropdownMenuPlugin?.menu;
 
     if (!menu) {
-      return indexes;
+      return [];
     }
 
-    // The visibility of components is toggled through the `hiddenRows` plugin of the rendered
-    // menu (`hotMenu`), so the indexes have to match the rows that are actually rendered. The
-    // `menuItems` collection still holds the raw, unfiltered items (e.g. leading/trailing/duplicated
-    // separators that `filterSeparators()` strips before rendering), so reading indexes from it
-    // would be off whenever such a separator precedes a component. Prefer the rendered source data,
-    // and fall back to `menuItems` only when the menu has not been rendered yet.
-    const renderedItems = (menu.hotMenu?.getSourceData() as unknown as { key: string }[] | null) ?? menu.menuItems;
-
-    arrayEach(components, (component) => {
-      const comp = (component as unknown) as { getMenuItemDescriptor(): { key: string } };
-
-      arrayEach(renderedItems ?? [], (item, index) => {
-        if ((item as { key: string }).key === comp.getMenuItemDescriptor().key) {
-
-          indexes.push(index);
-        }
-      });
-    });
-
-    return indexes;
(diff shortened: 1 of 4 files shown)


handsontable/src/plugins/filters/filters.ts
```diff
@@ -1541,10 +1541,18 @@ export class Filters extends BasePlugin {
       return indexes;
     }
 
+    // The visibility of components is toggled through the `hiddenRows` plugin of the rendered
+    // menu (`hotMenu`), so the indexes have to match the rows that are actually rendered. The
+    // `menuItems` collection still holds the raw, unfiltered items (e.g. leading/trailing/duplicated
+    // separators that `filterSeparators()` strips before rendering), so reading indexes from it
+    // would be off whenever such a separator precedes a component. Prefer the rendered source data,
+    // and fall back to `menuItems` only when the menu has not been rendered yet.
+    const renderedItems = (menu.hotMenu?.getSourceData() as unknown as { key: string }[] | null) ?? menu.menuItems;
```
budnix: This works, but it reaches two levels into the menu (`menu.hotMenu.getSourceData()`) to undo something the menu itself did. Both `filterSeparators()` and the hidden-item filter live inside `Menu`, so `Menu` is the only place that really knows the rendered list differs from `menuItems`. Letting that detail leak into `Filters` is what caused the off-by-one in the first place, and the next caller that needs a rendered index will likely trip on it the same way. Could we move this into a small method on `Menu` and have `Filters` just ask for the position by key? Then the long comment here shrinks to one line and the raw-vs-rendered knowledge stays where it belongs.

```ts
// in Menu
getItemPositionByKey(key) {
  return (this.hotMenu?.getSourceData() ?? this.menuItems ?? [])
    .findIndex(item => item.key === key);
}
```
```ts
// here, the whole loop becomes:
const indexes = components
  .map(component => menu.getItemPositionByKey(component.getMenuItemDescriptor().key))
  .filter(index => index !== -1);
```

budnix (commented):
One note on where this fix lives - the logic is correct, but it could sit one layer deeper.

budnix (approved):

Issue #6482, opened 2019-11-19
Cells are unmerged after filtering


AMBudnik:
### Description

This issue is related to this one https://github.com/handsontable/handsontable/issues/6481 as you may find a missing header while testing these steps.

It issue happens because we move the `mergedCells` accordingly to the new indexes, so after filtering we still look for the `(2, 2)` cell. However, from the 'business side', it may look like a bug.

### Steps to reproduce

1. Open the dropdown menu for column `C`
2. Use `Clear` in the `Filter by value`
3. Choose `(Blank cells)` and `C3`

![GIF](https://user-images.githubusercontent.com/10757813/69136688-94cf5800-0abb-11ea-8bab-694aaaa2e234.gif)

**Result**
![image](https://user-images.githubusercontent.com/10757813/69136723-a87abe80-0abb-11ea-9da7-2bd453349442.png)

**Expected Result**
![image](https://user-images.githubusercontent.com/10757813/69136888-fe4f6680-0abb-11ea-8e1f-ded5cd64b08d.png)

### Demo

https://jsfiddle.net/fjnmtqo1/

### Your environment
* Handsontable version: 7.2.2
* Browser Name and version: Chrome 78
* Operating System: Windows 10

KrzysztofZie:
This has been fixed on the develop branch and will be included in the next release.

PR #12792, opened 2026-06-16
DEV-404 Cells are unmerged after filtering

KrzysztofZie:
### Context
Fixes merged cells getting unmerged / mis-positioned across row trimming (Filters / trimRows / nestedRows) in the core MergeCells plugin. The plugin caches each merge's
  physical top-left anchor to re-anchor it onto still-visible rows while trimming is active, and restore the original layout once it clears.

  Fixes
  - Stale anchor after insert/remove under a filter — row insert/remove renumbers physical indexes, which the anchor cache didn't mirror, so re-anchoring landed on wrong rows  after the filter cleared. Added #remapRowAnchors, wired into afterCreateRow/afterRemoveRow (gated on trimming active). Remove uses the uniform v − count(removed < v)
  formula, correct even when the anchor row itself is removed.
  - Cached range not invalidated on re-anchor — added relocate(row, col) on MergedCellCoords that sets the position and clears the lazy #cellRange; relocateInMatrix now uses
  it.
  - Phantom merge on full hide — a fully hidden merge's stale entry stayed in the lookup matrix. Such merges are now purged (removeFromMatrix), tracked in #purgedMerges, and
  force re-added when they reappear (even at the same coordinates).
  - Trim re-anchor dropped when batched with a sort/move — removed the !indexesSequenceChanged gate in #onRowIndexCacheUpdated; the re-anchor now runs on any visibility
  change (standalone sort/move stays handled by its own hooks).
  - Purged merge re-entering the matrix on a structural edit — shiftCollections rebuilt the matrix from every merge; the purge is now idempotent and re-applied after
  shiftCollections in the create/remove row/col hooks.

  Tests — 5 unit (cellsCollection.unit.ts: relocateInMatrix incl. two-merge swap, removeFromMatrix) + 6 e2e (mergeCells.spec.js: phantom purge, re-add at same coords,
  sort+filter batch, bootstrap trimRows, row/column insert not re-adding a purged merge). All pass: unit 32/32, e2e 495 specs, 0 failures.
  

### How has this been tested?
npm run test:e2e -- --testPathPattern=mergeCells.spec

### Types of changes

### Related issue(s):
1. https://github.com/handsontable/handsontable/issues/6482
2. 
3.

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-404

---

> [!NOTE]
> **Medium Risk**
> Touches core merge indexing and row-mapper hooks across filter, trim, sort, and structural edits; behavior is heavily tested but regressions in edge cases (nested rows, concurrent plugins) are possible.
> 
> **Overview**
> Fixes **MergeCells** losing position or leaving phantom overlaps when **Filters**, **`trimRows`**, **`nestedRows`**, or **column sorting** change which rows are visible.
> 
> The plugin now **stores each merge’s physical top-left** and listens to **`rowIndexMapper` `cacheUpdated`** to **re-anchor** visual `row`/`col` (full **`rowspan`/`colspan`**) onto still-visible rows—including batched sort+filter and standalone batched sorts. **`MergedCellCoords.relocate`** clears the lazy range cache; **`relocateInMatrix`** / **`removeFromMatrix`** update the lookup matrix without dropping merge definitions. **Fully hidden** merges are **purged** from the matrix and **forced back** when they reappear; **row/column insert/remove** **remaps anchors** and **re-applies** purge logic after **`shiftCollections`** rebuilds the matrix.
> 
> Adds a changelog entry and broad unit/e2e coverage for these scenarios.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4112693da5a1bdf63843824e6441fe8628b119d4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -1559,6 +1587,234 @@ export class MergeCells extends BasePlugin {
...
+    for (let c = topStart.col; c <= bottomEnd.col; c++) {
+      const physicalColumn = this.hot.toPhysicalColumn(c);
+
+      if (physicalColumn !== null) {
+        physicalColumns.add(physicalColumn);
+      }
+    }
+
+    this.#filterPhysicalSnapshot = this.#filterPhysicalSnapshot.filter(entry =>
+      !(entry.rows.some(row => physicalRows.has(row)) && entry.cols.some(col => physicalColumns.has(col)))
```
budnix: This drops a snapshot entry whenever its rows AND cols touch the unmerge bounding box, but `unmergeRange` only removes a merge from the live collection when its top-left corner sits inside the range (`getWithinRange` with `countPartials=false`). So a merge that overlaps the box but starts outside it stays merged on screen, yet its entry gets dropped here — and on the next `filter()`/clear, the rebuild re-creates only what's left in the snapshot, so that still-merged cell silently disappears. It's reachable through the public `unmerge(r1,c1,r2,c2)` with a range that overlaps a merge's body but not its corner. Would be good to match the same corner-based rule the live removal uses.

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -1559,6 +1587,234 @@ export class MergeCells extends BasePlugin {
...
+    for (let c = mergeParent.col; c < mergeParent.col + mergeParent.colspan; c++) {
+      const physicalColumn = this.hot.toPhysicalColumn(c);
+
+      if (physicalColumn !== null) {
+        cols.push(physicalColumn);
+      }
+    }
+
+    if (rows.length > 0 && cols.length > 0) {
+      this.#filterPhysicalSnapshot.push({ rows, cols });
```
budnix: While a filter is active a merge can be clipped to only its visible rows. If the user then re-merges that visible region (`mergeSelection` / Ctrl+M), the earlier unmerge drops the full original entry and this push re-captures only the rows we can still map via `toPhysicalRow` — the trimmed rows return null and get skipped. After clearing the filter the merge comes back covering only the formerly-visible rows, so the hidden rows stay permanently unmerged. Since the trimmed physical rows can't be recovered from the visible re-merge, this probably needs the snapshot entry updated in place rather than dropped-and-recaptured. Curious how you'd want to handle it.

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -172,6 +172,27 @@ export class MergeCells extends BasePlugin {
...
+  /**
+   * Snapshot of the full merge set (in physical row/column indexes) captured the first time
+   * a filter is applied from a clean state. Filtering trims rows, which reshuffles visual
+   * indexes; this snapshot lets us rebuild merges clipped to the visible rows and restore
+   * the original merges once all filter conditions are cleared. `null` whenever no filter
+   * is active, so a grid without merges (or without filtering) pays nothing.
+   *
+   * @type {Array<{ rows: number[], cols: number[] }> | null}
+   */
+  #filterPhysicalSnapshot: { rows: number[], cols: number[] }[] | null = null;
```
budnix: One thing to watch: this snapshot is captured once on the first filter and only synced through merge/unmerge after that. If the user inserts or removes rows/columns while a filter is active, `shiftCollections` updates the live merges but the snapshot keeps the old physical indexes. When the filter is later cleared, `#rebuildMergesFromPhysical` wipes the collection and re-creates from the stale snapshot, so the merge lands on the wrong rows. Would be good to either re-sync the snapshot in the `afterCreateRow`/`afterRemoveRow` (and column) handlers, or drop it when the row/column count changes.

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -1559,6 +1587,234 @@ export class MergeCells extends BasePlugin {
...
+    try {
+      this.clearCollections();
+
+      // Reset leftover `hidden`/`spanned` cell meta within the original merge areas, so cells
+      // no longer covered by a (clipped) merge render normally.
+      snapshot.forEach(({ rows, cols }) => {
+        rows.forEach((physicalRow) => {
+          const visualRow = rowMapper.getVisualFromPhysicalIndex(physicalRow);
+
+          if (visualRow === null) {
```
budnix: This meta reset skips cells whose row is currently trimmed (`visualRow === null`), and `clearCollections()` → `collection.clear()` doesn't reset `hidden`/`spanned` meta either. So a covered cell that's hidden at rebuild time keeps its stale `hidden`/`spanned` meta. As long as unfiltering always runs this rebuild it's fine, but if those rows are later revealed another way (e.g. `trimRows` toggled off) they'd render as leftover merged cells. Worth a quick check.

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -233,6 +254,10 @@ export class MergeCells extends BasePlugin {
     this.addHook('afterColumnMove', this.#onAfterColumnMove);
     this.addHook('beforeRowMove', this.#onBeforeRowMove);
     this.addHook('afterRowMove', this.#onAfterRowMove);
+    this.addHook('beforeFilter', this.#onBeforeFilter);
```
budnix: Can I ask — was filtering picked on purpose as the only trigger here? Filters trims rows through a `TrimmingMap`, and `trimRows` plus `nestedRows` collapse do the exact same thing (`hiddenRows` is different — it uses a `HidingMap` that keeps visual coords, so merges are fine there). Since nothing re-clips merges on those, the original bug should still reproduce under `trimRows` / collapsed nested rows. If that's out of scope for this PR a follow-up ticket would be good, but it might be worth reacting to the row trimming map changing rather than the Filters hooks specifically.

budnix (commented):
A few notes on the filter snapshot logic — mostly around edge cases the current E2E tests don't cover (insert/remove rows while filtered, re-merging a clipped merge, and the unmerge bounding-box drop), plus one scope question and a small perf win.

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -1559,6 +1587,234 @@ export class MergeCells extends BasePlugin {
...
+          .map(physicalColumn => columnMapper.getVisualFromPhysicalIndex(physicalColumn))
+          .filter((visualColumn): visualColumn is number => visualColumn !== null && visualColumn >= 0)
+          .sort((a, b) => a - b);
+
+        if (visualRows.length === 0 || visualColumns.length === 0) {
+          return;
+        }
+
+        MergedCellsCollection.detectContiguousRuns(visualRows).forEach((rowRun) => {
+          MergedCellsCollection.detectContiguousRuns(visualColumns).forEach((columnRun) => {
```
budnix: Small thing — `detectContiguousRuns(visualColumns)` doesn't depend on `rowRun`, so it gets recomputed for every row run even though the result is the same each time. Pulling it out above the outer `forEach` into a `columnRuns` const turns this from O(rows × cols) into O(rows + cols). Pure win, no behavior change.

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -1559,6 +1617,280 @@ export class MergeCells extends BasePlugin {
...
+   * `beforeFilter` hook callback. Captures the full merge set (in physical coordinates) the
+   * first time a filter runs from a clean (unfiltered) state, so the merges can later be
+   * rebuilt against whatever rows remain visible. No-op when there are no merged cells.
+   */
+  #onBeforeFilter = () => {
+    if (this.#filterPhysicalSnapshot !== null || this.mergedCellsCollection.mergedCells.length === 0) {
+      return;
+    }
+
+    this.#filterPhysicalSnapshot = this.#captureMergesAsPhysical();
```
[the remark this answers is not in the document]
KrzysztofZie: ▎ Thanks — this was accurate for an earlier revision, but it's now outdated. The #invalidateFilterSnapshot drop-on-edit approach this refers to was removed in 8a6e98ab8 and replaced by 
  ▎ #shiftFilterSnapshotForInsert / #shiftFilterSnapshotForRemove, which shift the physical indexes in the snapshot instead of dropping it. The snapshot therefore stays complete (filter-hidden rows 
  ▎ included), so clearing the filter restores the original merge. Covered by the tests "should not restore merges onto stale rows after removing rows while filtered" and "should keep filter-hidden rows
  ▎ of a clipped merge after removing a row while filtered". Resolving as outdated.

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -453,8 +490,16 @@ export class MergeCells extends BasePlugin {
...
-    this.mergeRange(cellRange);
+    // Let the auto-unmerge below hand its dropped entries to the following re-merge, so a merge
+    // re-created over a filter-clipped one keeps the hidden rows instead of trimming to the visible.
+    this.#remergeSnapshotReclaim = this.#filterPhysicalSnapshot === null ? null : [];
+
+    try {
+      this.unmergeRange(cellRange, true);
+      this.mergeRange(cellRange);
+    } finally {
+      this.#remergeSnapshotReclaim = null;
```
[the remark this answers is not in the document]
KrzysztofZie: his is intentional, not a leak. When mergeRange bails after the auto-unmergeRange, the live merge has already been removed from the grid — so the correct snapshot state is without that entry.
  ▎ Folding the reclaim back would resurrect a merge that no longer exists. The finally nulling the buffer is exactly what keeps the snapshot consistent with the live unmerge: reclaim is only consumed
  ▎ when a re-merge actually fires (#onAfterMergeCells), otherwise the entry is correctly discarded along with the live removal. Covered by the test "should not resurrect a merge on filter clear when
  ▎ mergeSelection unmerged then failed to merge" (mergeCells.spec.js:2645), which asserts the merge does not reappear after the filter is cleared. Resolving as working-as-intended.

handsontable/src/plugins/mergeCells/mergeCells.ts
KrzysztofZie: ▎ Good catch — this was exactly the gap in the 06-16 revision you reviewed. It's now handled by the #remergeSnapshotReclaim bridge (added in cf4d4fe04), which is essentially the "carry the entry over
  ▎ instead of drop-and-recapture" approach you suggested: on the auto-unmergeRange, #onAfterUnmergeCells stashes the full original entry (filter-hidden physical rows included) into the buffer, and the
  ▎ following mergeRange → #onAfterMergeCells folds those hidden rows/cols back into the new entry — so the visible re-merge no longer has to recover them via toPhysicalRow. After clearing the filter
  ▎ the merge comes back at full span. Covered by "should keep filter-hidden rows when re-merging the visible part of a clipped merge" and "...whose top row is hidden". Let me know if the bridge
  ▎ approach looks right to you vs. a true in-place update.

handsontable/src/plugins/mergeCells/mergeCells.ts
KrzysztofZie: Addressed since the revision you looked at. Insert/remove while filtered is now re-synced into the snapshot rather than left stale — your first option. #onAfterCreateRow/#onAfterCreateCol call   ▎ #shiftFilterSnapshotForInsert and #onAfterRemoveRow/#onAfterRemoveCol call #shiftFilterSnapshotForRemove (added in 8a6e98ab8), shifting the stored physical indexes in lockstep with shiftCollections:  ▎ inserts bump indexes up (and grow an entry the insert lands inside, mirroring MergedCellCoords.shift), removes drop the deleted indexes and shift survivors down, dropping entries left with no
  ▎ rows/cols.
  ▎
  ▎ I went with re-sync rather than drop-on-change on purpose — dropping would let the next filter() recapture from the already-clipped live merges and permanently lose the filter-hidden lines. Covered
  ▎ by "should not restore merges onto stale rows after removing rows while filtered", "should keep filter-hidden rows of a clipped merge after removing a row while filtered", and "should grow the
  ▎ snapshot merge (not split it) when a row is inserted inside it while filtered".

qunabu:
Task linked: [DEV-404 cells are unmerged after filtering](https://app.clickup.com/t/9015210959/DEV-404)

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -667,8 +695,176 @@ export class MergeCells extends BasePlugin {
...
+    }
+
+    const existing = this.#mergeAnchors.get(merge);
+    const physicalRow = (trimmingActive && existing)
+      ? existing.physicalRow
+      : this.hot.toPhysicalRow(merge.row);
+
+    if (physicalRow !== null) {
+      this.#mergeAnchors.set(merge, { physicalRow, physicalColumn });
+    }
```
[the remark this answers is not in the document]
KrzysztofZie: Already handled on this branch: #onAfterCreateRow / #onAfterRemoveRow now remap the cached physical anchors to match the index-mapper renumbering (insert: +count for anchors at/after the pivot; remove:
  physicalRow − count(removed < physicalRow) from the hook's removedPhysicalIndexes), instead of relying on #captureMergeAnchors, which intentionally preserves the stored physicalRow while trimming is  active. Gated on trimming being active; otherwise anchors are re-derived as before. Covered by two e2e tests (row inserted / row above removed during an active filter) — 101 specs, 0 failures.

handsontable/src/plugins/mergeCells/cellsCollection.ts
```diff
@@ -695,6 +695,26 @@ class MergedCellsCollection {
...
+   * to move together with their new top-left visual `row`/`col`.
+   */
+  relocateInMatrix(relocations: { mergedCell: MergedCellCoords, row: number, col: number }[]) {
+    relocations.forEach(({ mergedCell }) => this.#removeMergedCellFromMatrix(mergedCell));
+
+    relocations.forEach(({ mergedCell, row, col }) => {
+      mergedCell.row = row;
+      mergedCell.col = col;
+      this.#addMergedCellToMatrix(mergedCell);
+    });
```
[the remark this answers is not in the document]
KrzysztofZie: Already handled on this branch: relocateInMatrix no longer writes row/col in place — it calls a new MergedCellCoords.relocate(row, col) that sets the position and clears the lazily cached #cellRange
  (same invalidation shift() already does), so getRange() can't serve pre-relocation coords. Covered by a unit test in cellCoords.unit.ts (relocate moves the cell and invalidates the cached range).

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -667,8 +695,201 @@ export class MergeCells extends BasePlugin {
...
+      let visualRow = this.hot.toVisualRow(anchor.physicalRow);
+
+      // Anchor row hidden: re-anchor to the first still-visible physical row inside the span (the span
+      // covers physical rows `[physicalRow, physicalRow + rowspan)` since anchors are captured while
+      // the merge is fully visible).
+      for (let offset = 1; visualRow === null && offset < merge.rowspan; offset++) {
+        visualRow = this.hot.toVisualRow(anchor.physicalRow + offset);
+      }
+
+      if (visualRow === null) {
```
budnix: When all rows of a merge are trimmed, the early return skips `relocateInMatrix` entirely. The merge's old visual row stays in the matrix. If the next filter shows a different physical row at that same visual slot, `mergedCellsCollection.get(row, col)` returns this merge for the wrong row — a phantom merge on an unrelated cell.

Concrete case: filter A makes the merge visible at visual row 0. Filter B hides all the merge's rows — merge stays at visual row 0. Filter C then shows an unrelated row at visual row 0; `get(0, 2)` still returns this merge.

A fix would be to collect fully-hidden merges in a separate "purge" list and call `#removeMergedCellFromMatrix` for each, so they do not collide with whatever rows show up later.

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -667,8 +695,201 @@ export class MergeCells extends BasePlugin {
...
+    { trimmedIndexesChanged, hiddenIndexesChanged, indexesSequenceChanged }:
+    { trimmedIndexesChanged: boolean, hiddenIndexesChanged: boolean, indexesSequenceChanged: boolean }
+  ) => {
+    // The hook is removed in `disablePlugin`/`destroy`, but keep a defensive guard in case a final
+    // `updateCache` still reaches us after `this.hot` has been dropped.
+    if (!this.hot || !this.#initialized) {
+      return;
+    }
+
+    if ((trimmedIndexesChanged || hiddenIndexesChanged) && !indexesSequenceChanged) {
```
budnix: There is a gap when `suspendOperations()` batches a sort and a filter change together. The sort sets `indexesSequenceChanged = true`; the filter sets `trimmedIndexesChanged = true`. When `resumeOperations()` fires `cacheUpdated` once with both flags true, `!indexesSequenceChanged` blocks the trim re-anchor — even though the rows also changed via filtering.

The sort half is handled by `#onAfterColumnSort`, but if that hook fires before `resumeOperations()` (before the mapper cache updates), that re-anchor reads stale mapper state. The trim half is dropped entirely. So a sort + filter inside one `suspendOperations` block can leave a merge at its pre-filter visual row.

handsontable/src/plugins/mergeCells/cellsCollection.ts
```diff
@@ -695,6 +695,25 @@ class MergedCellsCollection {
...
+   * Relocates a batch of merges in the coords->merge lookup matrix, applying each merge's new visual
+   * top-left. Used after re-anchoring to the visible rows on a trimming/sort change. Incremental
+   * alternative to a full rebuild — only the cells of the affected merges are touched. Runs in two
+   * phases (remove all, then re-add all) so merges that swap visual positions don't clobber each
+   * other's freshly written entries.
+   *
+   * @param {Array<{ mergedCell: MergedCellCoords, row: number, col: number }>} relocations The merges
+   * to move together with their new top-left visual `row`/`col`.
+   */
+  relocateInMatrix(relocations: { mergedCell: MergedCellCoords, row: number, col: number }[]) {
```
budnix, line 708: `relocateInMatrix` is the core mutation for all re-anchoring, but it has no unit test coverage — `cellsCollection.unit.ts` has zero tests for it. The two-pass design (remove all old entries first, then relocate and re-add) has non-obvious invariants, especially for the case where two merges swap visual positions. Worth pinning down before this gets touched again.

budnix, line 708: `relocateInMatrix` is the core mutation for all re-anchoring, but it has no unit test coverage — `cellsCollection.unit.ts` has zero tests for it. The two-pass design (remove all old entries first, then relocate and re-add) has non-obvious invariants, especially for the case where two merges swap visual positions. Worth pinning down before this gets touched again.

KrzysztofZie: Added unit coverage for relocateInMatrix in cellsCollection.unit.ts, including the two-merge swap case that pins down the two-pass remove-then-add invariant (plus  removeFromMatrix). 5 tests, all passing.

KrzysztofZie: Added relocateInMatrix unit tests in cellsCollection.unit.ts, including the two-merge swap case that pins down the two-pass remove-then-add invariant (plus removeFromMatrix). 5 tests, all passing.

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -667,8 +706,221 @@ export class MergeCells extends BasePlugin {
     this.generateFromSettings();
     this.hot.render();
     this.#initialized = true;
+    this.#captureMergeAnchors();
+  };
```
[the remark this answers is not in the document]
KrzysztofZie: The pre-#initialized cacheUpdated runs on an empty collection (no-op). After generation, anchors are captured from toPhysicalRow(merge.row), so a bootstrap re-anchor would be toVisualRow(toPhysicalRow(merge.row)) === merge.row — a guaranteed no-op; visual config coords are always placed on visible rows. Added a regression test
(trimRows above a config merge → untrim follows physical rows). Not a bug.

budnix (approved):

Issue #7309, opened 2020-10-02
Extra cells are produced after merging and unmerging cells and used toHTML()


CarterSnich:
### Description

Extra cells are produced after merging and unmerged the cells and converted to HTML with toHTML()

### Steps to reproduce

1. initialize Handsontable
2. merge and unmerged cells
3. convert to HTML table with toHTML and display it on the DOM

### Demo

https://jsfiddle.net/handsoncode/8ffpsqt6/
I don't know how to use the JS Fiddle, I lost my changes. But, it all was basic settings. Just merge and unmerged some cells, and display it on the DOM to see the extra cells produced. I will insert an image instead.
![image](https://user-images.githubusercontent.com/52433531/94907978-93403000-04d3-11eb-8faf-0bcb8bf20333.png)

### Your environment
* Handsontable version: 8.0.0
* Browser Name and version: Chrome 84.0.4147.105
* Operating System: Windows 10

wojciechczerniak:
Repro: https://jsfiddle.net/y6p8gsbu/2/

We set `rowspan` and `colspan` but when `mergeParent` is empty (was unmerged) we never clear the meta.

https://github.com/handsontable/handsontable/blob/b7618fc4213d3b2ab32ba62ff6f1d57cd1af6c49/src/plugins/mergeCells/mergeCells.js#L809-L821

Clearing is not necessary for merge cells because there are few other flags taken under consideration, but `toHTML` reads only `rowspan` and `colspan`. 

Same in v7.4.0: https://jsfiddle.net/y6p8gsbu/3/

AMBudnik:
Thank you for sharing the issue @CarterSnich I will update you about its fix.
And thank you @wojciechczerniak for the confirmation.

karishmademla:
I solved this issue by removing rowspan and colspan cell meta in the unmerge function of Handsontable for the unmerged cell:
`_this5.hot.removeCellMeta(currentCollection.row, currentCollection.col, 'rowspan');  _this5.hot.removeCellMeta(currentCollection.row, currentCollection.col, 'colspan');`

CarterSnich:
> I solved this issue by removing rowspan and colspan cell meta in the unmerge function of Handsontable for the unmerged cell:
> `_this5.hot.removeCellMeta(currentCollection.row, currentCollection.col, 'rowspan'); _this5.hot.removeCellMeta(currentCollection.row, currentCollection.col, 'colspan');`

Thank you. I'mma try it later.

[...]

PR #12798, opened 2026-06-17
DEV-482 Extra cells are produced after merging and unmerging cells an…

KrzysztofZie:
### Context
Fixed clears stale cell meta after a merge is removed (unmerge).

  Previously the #onAfterGetCellMeta hook set colspan/rowspan/spanned/hidden when a merge existed, but never reset them once the merge was gone. The live grid still rendered correctly (the renderer sets <td>  attributes directly), but consumers reading meta — e.g. toHTML() — got the stale colspan=3 plus the now-revealed cells, so the row broke.
  Now the "no merge parent" branch deletes those flags, so after unmerge the meta is clean and toHTML() returns a correct row.

### How has this been tested?
npm run test:e2e -- --testPathPattern=mergeCells.spec

### Types of changes

### Related issue(s):
1. https://github.com/handsontable/handsontable/issues/7309
2. DEV-482
3.

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-482

---

> [!NOTE]
> **Low Risk**
> Localized bug fix in the mergeCells plugin with focused tests; behavior change only clears incorrect cached meta after unmerge/disable, with no API or breaking changes.
> 
> **Overview**
> Fixes stale merge-related cell meta (`spanned`, `colspan`, `rowspan`, `hidden`, `copyable`) left on cells after a merge is removed, which could break **`toHTML()`** and other code that reads meta directly while the on-screen grid still looked correct.
> 
> The **MergeCells** plugin now **eagerly** strips those flags via a shared **`#resetMergedCellMeta`** helper when **`unmergeRange`** drops a merge and when **`clearCollections`** runs (e.g. disabling the plugin or resetting merges). That avoids relying on a later render or lazy **`afterGetCellMeta`** updates when work runs inside **`batch()`** and meta is memoized.
> 
> New e2e tests cover plain unmerge, **`toHTML`** output, batched unmerge, and batched plugin disable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5a0b6ddc59b1af382f9c9681cc73e32381323d6c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/handsontable/src/plugins/mergeCells/__tests__/mergeCells.spec.js
+++ b/handsontable/src/plugins/mergeCells/__tests__/mergeCells.spec.js
@@ -1715,6 +1715,48 @@ describe('MergeCells', () => {
     expect(getCellMeta(4, 4).copyable).toBe(false);
   });
 
+  it('should clear the "spanned"/"colspan"/"rowspan"/"hidden" cell meta after performing unmerge', async() => {
+    handsontable({
+      data: createSpreadsheetData(5, 5),
+      mergeCells: [
+        { row: 1, col: 1, rowspan: 1, colspan: 3 },
+      ],
+    });
+
+    const plugin = getPlugin('mergeCells');
+
+    // the merge parent carries the span flags, the covered cells are hidden
+    expect(getCellMeta(1, 1).spanned).toBe(true);
+    expect(getCellMeta(1, 1).colspan).toBe(3);
+    expect(getCellMeta(1, 2).hidden).toBe(true);
+    expect(getCellMeta(1, 3).hidden).toBe(true);
+
+    plugin.unmerge(1, 1, 1, 3);
+
+    // after unmerge the cached meta must not linger (it would leak into `toHTML`, copy, etc.)
+    expect(getCellMeta(1, 1).spanned).toBeUndefined();
+    expect(getCellMeta(1, 1).colspan).toBeUndefined();
+    expect(getCellMeta(1, 1).rowspan).toBeUndefined();
+    expect(getCellMeta(1, 2).hidden).toBeUndefined();
+    expect(getCellMeta(1, 3).hidden).toBeUndefined();
+  });
+
+  it('should not emit stale colspan/rowspan in `toHTML` output after performing unmerge', async() => {
+    handsontable({
+      data: createSpreadsheetData(5, 5),
+      mergeCells: [
+        { row: 1, col: 1, rowspan: 1, colspan: 3 },
+      ],
+    });
+
+    getPlugin('mergeCells').unmerge(1, 1, 1, 3);
+
+    const html = hot().toHTML();
+
+    expect(html).not.toContain('colspan');
+    expect(html).not.toContain('rowspan');
+  });
+
   it('should not collapse the main table\'s row height when the merge cell covers all cells width', async() => {
     handsontable({
       data: createSpreadsheetData(5, 5),
--- a/handsontable/src/plugins/mergeCells/__tests__/mergeCells.spec.js
+++ b/handsontable/src/plugins/mergeCells/__tests__/mergeCells.spec.js
@@ -1757,6 +1757,79 @@ describe('MergeCells', () => {
     expect(html).not.toContain('rowspan');
   });
 
+  it('should not emit stale colspan/rowspan in `toHTML` output after a batched unmerge', async() => {
+    handsontable({
+      data: createSpreadsheetData(5, 5),
+      mergeCells: [
+        { row: 1, col: 1, rowspan: 1, colspan: 3 },
+      ],
+    });
+
+    // Prime the per-render meta memo while the merge is still active, so the in-batch read below
+    // hits the memoized path - the one on which the old lazy `afterGetCellMeta` cleanup is skipped.
+    expect(getCellMeta(1, 1).colspan).toBe(3);
+
+    let parentMeta;
+    let coveredMeta;
+    let html;
+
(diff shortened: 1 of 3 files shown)


handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -1200,6 +1200,15 @@ export class MergeCells extends BasePlugin {
         cellProperties.colspan = mergeParent.colspan;
         cellProperties.spanned = true;
       }
+    } else {
+      // The cell is no longer part of a merge (e.g. after `unmerge`). The cell meta object is
+      // cached, so the span/hidden flags set by a previous merge would linger and leak into
+      // consumers that read meta directly (e.g. `toHTML`). Reset them here, where we already
```
budnix: `unmergeRange` already clears `hidden`/`copyable`/`spanned` with `removeCellMeta` (lines 589-594) but never touches `rowspan`/`colspan`, so those two rely entirely on this branch running on the next `getCellMeta`. That read is memoized per render and the memo only clears on `beforeRender`, so if an unmerge runs inside `batch()` or `suspendRender()`, `hot.render()` skips the real render, this branch never fires, and `toHTML()` still emits the old `colspan`/`rowspan` — the same leak this PR fixes. Moving the `rowspan`/`colspan` removal up next to the existing `removeCellMeta` calls in `unmergeRange` would close that gap, drop the per-cell delete on every render, and make this else-branch unneeded. Would be good to verify with a batched unmerge.

handsontable/src/plugins/mergeCells/__tests__/mergeCells.spec.js
```diff
@@ -1715,6 +1715,48 @@ describe('MergeCells', () => {
...
+
+  it('should not emit stale colspan/rowspan in `toHTML` output after performing unmerge', async() => {
+    handsontable({
+      data: createSpreadsheetData(5, 5),
+      mergeCells: [
+        { row: 1, col: 1, rowspan: 1, colspan: 3 },
+      ],
+    });
+
+    getPlugin('mergeCells').unmerge(1, 1, 1, 3);
```
budnix, line 1752: Curious whether a batched version of this would still pass — wrapping the unmerge in `hot.batch(...)` before `toHTML()` should expose the case where `rowspan`/`colspan` stay stale, since the meta read gets memoized and isn't recomputed without a real render. As written, both new tests run the unmerge with rendering live, so they won't catch that path. Might be worth adding one that does.

qunabu:
Task linked: [DEV-482 Extra cells are produced after merging and unmerging cells and used toHTML()](https://app.clickup.com/t/9015210959/DEV-482)

handsontable/src/plugins/mergeCells/__tests__/mergeCells.spec.js
KrzysztofZie: Good call — that batched path is now covered. Added should not emit stale colspan/rowspan in toHTML output after a batched unmerge: it primes the per-render meta memo while the merge is active, then runs the unmerge + reads meta/toHTML() inside hot.batch(...), before the closing render clears the memo. There's also a sibling test for disablePlugin() within a batch. The fix clears the span meta eagerly in unmergeRange/clearCollections instead of lazily in afterGetCellMeta, so the memoized path no longer leaks stale spans. All green (482 specs, 0 failures).

handsontable/src/plugins/mergeCells/__tests__/mergeCells.spec.js
```diff
@@ -1715,6 +1715,121 @@ describe('MergeCells', () => {
...
+
+    const html = hot().toHTML();
```
budnix:
```suggestion
    const html = toHTML();
```

handsontable/src/plugins/mergeCells/__tests__/mergeCells.spec.js
```diff
@@ -1715,6 +1715,121 @@ describe('MergeCells', () => {
...
+
+    let parentMeta;
+    let coveredMeta;
+    let html;
+
+    // The unmerge runs without a real render (suspended inside `batch`), so the meta must be
+    // cleared eagerly in `unmergeRange` and cannot rely on `afterGetCellMeta` recomputing. The
+    // read happens inside the batch on purpose: `batch` ends with a render that clears the memo,
+    // so reading after it would mask the stale-meta path (the trap the original tests fell into).
+    hot().batch(() => {
```
budnix:
```suggestion
    batch(() => {
```

handsontable/src/plugins/mergeCells/__tests__/mergeCells.spec.js
```diff
@@ -1715,6 +1715,121 @@ describe('MergeCells', () => {
...
+    expect(getCellMeta(1, 1).colspan).toBe(3);
+
+    let parentMeta;
+    let coveredMeta;
+    let html;
+
+    // Disabling the plugin drops the whole collection via `clearCollections`. The internal render
+    // is suspended by `batch`, so - just like the unmerge case - the span meta must be cleared
+    // eagerly. The read happens inside the batch, before the closing render clears the memo.
+    hot().batch(() => {
```
budnix:
```suggestion
    batch(() => {
```

budnix (changes requested):
I left a code suggestion about using helpers instead of invoking `hot()` directly.

budnix (approved):

PR #12815, opened 2026-06-22
fix(ci): deploy docs to Cloudflare Pages apex, not previews

qunabu:
### Context

The PR fixes the docs Cloudflare Pages deployments, which only ever produced **preview** deployments and never populated the apex URLs:

- `handsontable-docs.pages.dev` (production) stayed empty.
- `handsontable-docs-staging.pages.dev` (staging) stayed empty.

Root cause: a Cloudflare Pages deployment is served at the apex `<project>.pages.dev` only when its `wrangler --branch` matches the project's `production_branch` (Cloudflare's term for the branch label served at the apex — not a git branch). Any other branch is a preview. `docs/cloudflare/ensureProject.mjs` created both projects with `production_branch: 'develop'`, but the workflows deployed to non-matching branches:

| Project | Deploy `--branch` | CF apex branch | Result |
|---|---|---|---|
| `handsontable-docs` | `main` | `develop` | preview |
| `handsontable-docs-staging` | `pr-<N>` | `develop` | preview |

This PR aligns the CF apex branch label with what each workflow deploys and makes `ensureProject.mjs` reconcile the label on already-existing projects.

Changes:

- `docs/cloudflare/ensureProject.mjs` — adds optional `CF_PAGES_APEX_BRANCH` env var (default `develop`) and now PATCHes the project's `production_branch` when it differs, instead of exiting early. This makes the fix self-healing on the next workflow run.
- `.github/workflows/docs-production.yml` — passes `CF_PAGES_APEX_BRANCH: production` and deploys `--branch production`, so production lands on `handsontable-docs.pages.dev`. `production` is a **stable** Cloudflare label, deliberately decoupled from the real per-version git branch (e.g. `prod-docs/18.0`) — using the git branch would move the apex on every release.
- `.github/workflows/docs-staging.yml` — new `cf-target` step routes the deploy: push/dispatch on `develop` deploys `--branch develop` (apex), while PR builds keep `--branch pr-<N>` (isolated preview). Passes `CF_PAGES_APEX_BRANCH: develop` (here the label coincides with the real long-lived `develop` branch) and adds an apex deploy step summary.

Rollout note: changing the apex branch label only affects future deployments — Cloudflare does not retroactively promote existing previews. Re-running each workflow once (a `prod-docs/**` build for production, a `develop` push touching `docs/**` for staging) populates the apex URLs. The PATCH requires an API token with Pages:Edit (write) scope.

### How has this been tested?

CI/tooling-only change. Validated locally:

```bash
[...]

PR #12750, opened 2026-06-10
Fix Cloudflare Vue 3 docs redirects

qunabu:
### Context
The PR fixes Cloudflare Pages docs redirects for legacy Vue 3 URLs. The Cloudflare `_worker.js` ignored Netlify `_redirects` when deployed, so `/docs/vue`, `/docs/vue3`, and framework-prefixed `vue3-*` pages did not reach the current `/docs/vue-data-grid/*` destinations.

[skip changelog]

### How has this been tested?

- `npm --prefix docs run docs:test:cloudflare`
- `git diff --check HEAD`

### Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature or improvement (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Additional language file or change to the existing one (translations)

### Related issue(s):
1. Cloud agent task

### Affected project(s):
- [ ] `handsontable`
- [ ] `@handsontable/angular-wrapper`
- [ ] `@handsontable/react-wrapper`
- [ ] `@handsontable/vue3`

### Checklist:
- [x] I have reviewed the guidelines about [Contributing to Handsontable](https://github.com/handsontable/handsontable/blob/master/CONTRIBUTING.md) and I confirm that my code follows the code style of this project.
- [x] I have signed the [Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLScpMq4swMelvw3-onxC8Jl29m0fVp5hpf7d1yQVklqVjGjWGA/viewform?c=0&w=1)
- [ ] My change requires a change to the documentation.

<div><a href="https://cursor.com/agents/bc-ae1d3860-a78e-4edc-8e76-a5727d9ff85b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-ae1d3860-a78e-4edc-8e76-a5727d9ff85b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #12818, opened 2026-06-22
ci(docs): migrate documentation deployment fully to Cloudflare Pages, remove Netlify

qunabu:
### Context

The PR migrates the Handsontable documentation deployment **fully to Cloudflare Pages and removes the Netlify setup** — both staging/preview and production. Follow-up to #12815 and #12817.

Cloudflare Pages now serves both environments:

| Cloudflare Pages project | Apex | Serves |
|---|---|---|
| `handsontable-docs` | `handsontable-docs.pages.dev` (custom domain: handsontable.com/docs) | Production |
| `handsontable-docs-staging` | `handsontable-docs-staging.pages.dev` | `develop` apex + per-PR previews (`pr-<N>.…`) |

The Cloudflare worker (`docs/cloudflare/_worker.js`) already re-implements every redirect/rewrite the Netlify `_redirects` file and the Netlify edge functions provided (verified function-by-function), so the edge functions are removed with no loss of behavior.

### Changes

**Staging — Cloudflare-only (`docs-staging.yml`, `docs-staging-delete.yml`)**
- Removed the Netlify provisioning/publish steps, the dead `Get PR details` step, and the Netlify sticky comments. Preview docs now come only from `pr-<N>.handsontable-docs-staging.pages.dev`; `develop` pushes deploy to the apex.
- `docs-staging-delete.yml` now only cleans up Cloudflare Pages previews.

**Production — Cloudflare-only (`docs-production.yml`)**
- Removed the `Publish production to Netlify` step. Cloudflare Pages is now the sole production host, so the `Deploy production to Cloudflare Pages` step is no longer `continue-on-error` — a failed production deploy fails the workflow.
- The multi-version assembly (`build_current_version.sh`) is unchanged in behavior; it just runs from the relocated `docs/deploy/` directory.

**Directory cleanup (`docs/netlify/` → deleted)**
- Relocated the build scripts Cloudflare still needs to `docs/deploy/`: `build_current_version.sh`, `getListOrPreviousVersions.mjs`, `rewriteVersionedPaths.mjs`, `_redirects`, `_dev_headers`, `package.json` (dropped the unused `netlify` CLI dep), `.gitignore`.
- Deleted Netlify-only/dead files: `netlify.toml`, `build.sh`, and the entire `docs/netlify/netlify/` tree (7 edge functions, `cookieHelper.mts`, `redirects.mts`).
- Updated `docs/.eslintignore` to the new artifact paths (`deploy/docs`, `deploy/preview`).

**Release pipeline (`publish.yml`)**
- Repointed the RC docs Slack-notification URLs from `rc-handsontable-<version>.netlify.app` to `https://handsontable-docs-staging.pages.dev/docs/...` (the CF staging apex; not version-specific).

**Documentation**
- Rewrote `docs/README-DEPLOYMENT.md` around Cloudflare Pages (projects, apex-vs-preview model, staging diagram, production deploy + dashboard rollback, Cloudflare-only Redirects section).
- Updated `docs/README.md` (directory map + branch-structure prose) and `docs/README-EDITING.md` to reference Cloudflare Pages.
- Updated a test comment in `docs/tests/version-comparison.spec.ts` to be host-neutral.

### Out of scope (intentionally left)

`hot-docs-assistant.netlify.app` (the docs AI-assistant API fallback in `docs/src/components/DocsAssistant/constants.ts` and `docs/astro.config.mjs`) is a **separate backend service**, not the docs deployment, so it is unchanged.

### Update — rebased onto `develop` (2026-07-02)

This PR had gone stale relative to fast-moving work on the exact files it renames/deletes, so it was rebased onto current `develop` (previously it was ~3 days and 7 commits behind on `docs-production.yml`, `docs-staging.yml`, and `docs/cloudflare/_worker.js`). Merging the old version as-is would have **silently reverted** two already-shipped Cloudflare fixes:

- "Always deploy Cloudflare worker" / fix docs redirect outage, #12916, #12917 — the `Add Cloudflare worker to production deploy` step now reads a `LATEST_DOCS_VERSION` file (written by `build_current_version.sh`) to substitute the worker's `__LATEST_DOCS_VERSION__` placeholder, plus a `test -f docs/_worker.js` tripwire. This PR's stale version still had the old `cp cloudflare/_worker.js ...` one-liner, which would have shipped the worker without the substitution or the tripwire, and without rule 1b (the latest-version redirect) at all.
- "Fix stale latest-version data and version-less links on old docs", #12928 — added `rewriteVersionedPaths.mjs` (invoked from `build_current_version.sh`) to fix broken links on frozen historical doc versions. Relocated into `docs/deploy/` along with its test, same as the other build scripts.
- Angular-cookie legacy-version fallback (#12927) and Vue 3 redirect fixes (#12750) to `docs/cloudflare/_worker.js` carried over cleanly.

Resolving the rebase also surfaced two pre-existing bugs, unrelated to staleness, now fixed:
- `docs-staging.yml`'s RC-failure Slack alert read `steps.netlify-publish.outcome`, a step that doesn't exist in this PR's design — it silently evaluated to empty and never reported a Cloudflare deploy failure. Now reads `steps.cf-pages-deploy.outcome`.
- `docs/package.json`'s `docs:test:plugins` script still globbed the deleted `netlify/__tests__/*.test.mjs` instead of the relocated `deploy/__tests__/*.test.mjs`, which would have silently dropped the `rewriteVersionedPaths` test from CI.

Also cleaned up Netlify mentions outside this PR's original diff: comments in `docs/cloudflare/_worker.js` and `docs/deploy/build_current_version.sh`, and stale platform descriptions in `docs/AGENTS.md`, `.ai/STACK.md`, root `.gitignore`, and `handsontable/.ai/INTEGRATIONS.md`.

### How has this been tested?

CI/tooling + docs change. Validated locally:

```bash
cd docs && node -e "const yaml=require('js-yaml'),fs=require('fs');['docs-staging.yml','docs-staging-delete.yml','docs-production.yml','publish.yml'].forEach(f=>{yaml.load(fs.readFileSync('../.github/workflows/'+f,'utf8'));console.log('OK',f)})"
npm run docs:test:plugins
```

- All four workflow YAML files parse.
- `docs:test:plugins` passes 93/93 (includes the relocated `deploy/__tests__/rewriteVersionedPaths.test.mjs` and the `cloudflare/__tests__/_worker.test.mjs` redirect suite).
- No stale `docs/netlify`, `netlify deploy`, `netlify-cli`, or `NETLIFY_*` references remain (repo-wide grep, excluding the intentionally-out-of-scope `hot-docs-assistant.netlify.app`).
- End-to-end verification happens on the next docs-staging (`develop`) and docs-production (`prod-docs/**`) runs; this PR's own preview build (`pr-12818.handsontable-docs-staging.pages.dev`) exercises the staging path.

### Types of changes

### Related issue(s):

Follow-up to #12815 and #12817.

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

[skip changelog] — CI/tooling and docs-only change, no package source modified.

---

> [!NOTE]
> **High Risk**
> Changes production and staging docs deployment, worker tripwires, and redirect routing—misconfiguration could break live documentation URLs or leave previews undeployed despite green CI.
> 
> **Overview**
> **Documentation hosting moves entirely to Cloudflare Pages** — Netlify provisioning, publish steps, `docs/netlify/` (edge functions, `_redirects`, CLI scripts), and dual-platform PR comments are removed.
> 
> **CI workflows** now assemble and deploy from `docs/deploy/` (`preview/` for staging, `docs/` for production). Staging maps branches to Cloudflare labels (`develop`, `pr-<N>`, `rc-<version>`, sanitized slugs); production drops Netlify and treats Cloudflare as the only host. Worker copy/deploy steps lose `continue-on-error` where that could ship without `_worker.js`; production and staging deploys use **retry** for `wrangler`. PR close cleanup deletes up to three Cloudflare branch labels; RC Slack alerts and `publish.yml` RC doc links point at `*.handsontable-docs-staging.pages.dev`.
> 
> **Build scripts** live under `docs/deploy/` (renamed package, no Netlify dep); production assembly no longer appends latest-version rules to `_redirects`. `build_previous_versions.sh` rewrites paths for flat Astro images; new local `deploy/build.sh` mirrors worker injection. Redirect authority is documented as **`cloudflare/_worker.js` only**; CSP drops `app.netlify.com` from `frame-src`.
> 
> **Docs/tests**: deployment READMEs and `AGENTS.md` updated; `docs:test:plugins` globs `deploy/__tests__` instead of deleted Netlify tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5ec155d89f6cdae69ebc1ee50603d283fa5867c3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/workflows/docs-staging.yml
```diff
@@ -171,42 +129,20 @@ jobs:
           BUILD_MODE: ${{ steps.release-info.outputs.is_release == 'true' && 'production' || '' }}
           PUBLIC_CHAT_API_URL: ${{ secrets.PUBLIC_CHAT_API_URL }}
 
-      - name: Build and deploy the preview to Netlify
```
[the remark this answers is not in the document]
qunabu: Fixed. `cf-target` now deploys a preview for every run instead of only develop/PR: `release/x.y.z` builds deploy to a stable `rc-<version>` branch (e.g. `rc-16-0-0.handsontable-docs-staging.pages.dev`), and any other branch (manual `workflow_dispatch`) deploys to a sanitized `<branch-slug>` preview. Only develop maps to the apex; nothing falls through to `deploy=false` for a branch push/dispatch anymore. Verified by executing the actual cf-target script across all branch scenarios.

.github/workflows/docs-staging.yml
```diff
@@ -307,7 +231,7 @@ jobs:
           echo "" >> $GITHUB_STEP_SUMMARY
           echo "**Target version:** ${{ steps.release-info.outputs.target_version }}" >> $GITHUB_STEP_SUMMARY
           echo "**Build mode:** production" >> $GITHUB_STEP_SUMMARY
-          echo "**Preview URL:** ${{ env.NETLIFY_SITE_URL }}/docs" >> $GITHUB_STEP_SUMMARY
+          echo "**Preview URL:** https://${{ steps.cf-target.outputs.branch }}.handsontable-docs-staging.pages.dev/docs" >> $GITHUB_STEP_SUMMARY
```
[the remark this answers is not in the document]
qunabu: Fixed. The RC summary now only runs when the Cloudflare deploy succeeded (`steps.cf-pages-deploy.outcome == 'success'`), and `cf-target.outputs.branch` is now always populated for release branches (`rc-<version>`), so the URL can no longer be malformed. The release pipeline (publish.yml) links to the same `rc-<version>` URL.

qunabu:
@cursoragent resolve conflicts

demtario:
@qunabu There are conflicts and bugbot comments

.github/workflows/docs-staging.yml
```diff
@@ -191,82 +142,75 @@ jobs:
...
-          # Place 404.html at the publish root so Netlify serves it
+          # Place 404.html at the publish root so the host serves it
           # automatically for all unmatched routes (no redirect rule needed).
-          cp ../dist/404.html deploy/404.html
-
-          cp _dev_headers deploy/_headers
-          cp _redirects deploy/_redirects
-
-      - name: Publish preview to Netlify
-        id: netlify-publish
```
budnix, line 209: This step is being removed, but `steps.netlify-publish.outcome` is still read later in the "Prepare RC docs failure Slack payload" step (around file line 286), where it now resolves to an empty string — so that Netlify branch is dead code. The bigger thing: the new Cloudflare deploy step has `continue-on-error: true`, so a failed RC docs deploy no longer fails the job or pings Slack, while the old Netlify publish here did. Worth a quick check — maybe point that payload at `steps.cf-pages-deploy.outcome` so RC deploy failures still get reported?

budnix (commented):
One inline note on a leftover Netlify reference after the cleanup.

docs/netlify/netlify/edge-functions/readme.md
```diff
@@ -1,156 +0,0 @@
-# Netlify Edge Functions for Documentation Redirects
```
qunabu, line 1: This Redirects docs is removed but CF docs is missing

demtario:
## Rebased onto `develop` — summary

This PR was stale relative to `develop` (last synced 2026-06-30 09:23, ~3 days and 7 commits behind on the exact files it renames/deletes). Since it's now blocking on merge conflicts, I rebased it onto current `develop` instead of merging, to keep history linear.

### Why this mattered

Two Cloudflare-specific fixes had landed on `develop` after this PR's last sync, touching the same files this PR moves from `docs/netlify/` to `docs/deploy/`:

- #12916, #12917 — fixed a real docs-redirect outage. `docs-production.yml`'s "Add Cloudflare worker to production deploy" step now substitutes `__LATEST_DOCS_VERSION__` in the worker from a `LATEST_DOCS_VERSION` file (written by `build_current_version.sh`), and has a `test -f docs/_worker.js` tripwire so a missing worker fails loudly instead of silently falling back to `_redirects`. This PR's stale version still had the old `cp cloudflare/_worker.js ...` one-liner — no substitution, no tripwire, and the corresponding `LATEST_VERSION` redirect rule (1b) was missing from `_worker.js` entirely.
- #12928 — fixed stale latest-version data and version-less links on old docs, via a new `rewriteVersionedPaths.mjs` script called from `build_current_version.sh`.

Resolving this via a plain conflict-resolution (taking "ours") would have silently reverted both fixes back into production. Rebasing forced each hunk to be reconciled against the current state instead.

### What I did

1. Ran `git rebase develop` and resolved conflicts commit-by-commit, in each case keeping `develop`'s newer logic and re-applying this PR's `netlify/` → `deploy/` rename on top of it (not the other way around):
   - `docs-production.yml` — kept the `LATEST_DOCS_VERSION` substitution + tripwire, pointed at `deploy/`.
   - `docs-staging.yml` — kept `develop`'s pull_request-event trigger and release-head race-condition fix, renamed the step/job labels to Cloudflare Pages.
   - `docs/deploy/build_previous_versions.sh` — kept `develop`'s fail-closed, layout-aware version.
   - `docs/deploy/rewriteVersionedPaths.mjs` + its test — relocated into `docs/deploy/` (would otherwise have been silently dropped since this PR's diff predates the file's existence).
   - Netlify edge-function deletions — verified each had an equivalent already in `_worker.js` before deleting (e.g. the Angular-cookie legacy-version fallback from #12927).

2. While resolving, found and fixed two bugs that predate the rebase and aren't about staleness:
   - `docs-staging.yml`'s RC-failure Slack alert referenced `steps.netlify-publish.outcome` — a step ID that doesn't exist in this PR's Cloudflare-only design, so it silently evaluated to empty and never reported a Cloudflare deploy failure. Fixed to read `steps.cf-pages-deploy.outcome`.
   - `docs/package.json`'s `docs:test:plugins` script still globbed the deleted `netlify/__tests__/*.test.mjs` directory instead of the relocated `deploy/__tests__/*.test.mjs`, which would have silently dropped the `rewriteVersionedPaths` test from CI.

3. Cleaned up leftover Netlify references outside this PR's original diff scope: comments in `docs/cloudflare/_worker.js` and `docs/deploy/build_current_version.sh`, and stale platform descriptions in `docs/AGENTS.md`, `.ai/STACK.md`, root `.gitignore`, and `handsontable/.ai/INTEGRATIONS.md`.

### Verification

- All four workflow YAML files parse.
- `npm run docs:test:plugins` passes 93/93, including the relocated `rewriteVersionedPaths` test and the Cloudflare worker redirect suite.
- Repo-wide grep confirms no stale `docs/netlify`, `netlify deploy`, `netlify-cli`, or `NETLIFY_*` references remain (excluding the intentionally out-of-scope `hot-docs-assistant.netlify.app` backend).
- `docs/netlify/` no longer exists; `docs/deploy/` has everything the Cloudflare pipeline needs.

Force-pushed the rebased branch. CI is running fresh against it now — worth keeping an eye on the Docs Staging Deployment check specifically, since it's the empirical test of the merge resolution.

docs/netlify/netlify/edge-functions/readme.md
demtario: added `docs/cloudflare/README.md` for the redirect docs gap

.github/workflows/docs-staging.yml
demtario: Removed `continue-on-error: true`

demtario:
Verified the PR preview (`pr-12818.handsontable-docs-staging.pages.dev`):

- Page loads (200), worker is live (CSP header only `_worker.js` sets is present, so no silent `_redirects` fallback).
- Redirects checked and correct: `/` → `/docs`, `/docs` → `/docs/javascript-data-grid/`, `/docs/next/...` → unversioned, Vue 3 legacy (`vue3-installation` → `vue-data-grid`), pre-16 Angular legacy fallback.
- Static assets (JS/Astro chunks) return 200.

Cloudflare staging pipeline looks healthy after the rebase + fixes.

budnix (dismissed):

.github/workflows/docs-production.yml
```diff
@@ -80,33 +80,15 @@ jobs:
         run: |
           docker push $GHA_DOCKER_TAG":v${{steps.get-docs-version.outputs.result}}-${{ github.sha }}"
 
-      - name: Build and deploy the production to Netlify (from Docker images)
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      # Assemble the multi-version docs site (current + previous versions from
+      # the GHCR Docker images) into docs/deploy/docs, which is deployed to
+      # Cloudflare Pages below.
+      - name: Assemble production docs (all versions)
```
demtario, line 92: **Blocker: the `GITHUB_TOKEN` env was dropped together with the Netlify step.** The deleted `Build and deploy the production to Netlify` step carried `env: GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}`, and this replacement step still runs the same `build_current_version.sh` → `getListOrPreviousVersions.mjs`, which authenticates its GitHub GraphQL query with `process.env.GITHUB_TOKEN`. GitHub's GraphQL endpoint rejects unauthenticated requests with 401 unconditionally, so the script's 5 retries all fail, it exits 1 under `set -euo pipefail`, and **every `prod-docs/**` deploy fails at this step** (fail-closed: the old site stays up, nothing new ships). Nothing else supplies the token — the workflow-level `env` only has `GHA_DOCKER_TAG` and the job has no `env` block. Note the staging workflow never runs these scripts, so this PR's green preview build gives no coverage of this path.

```suggestion
      - name: Assemble production docs (all versions)
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

demtario (commented):
Deep review of the Netlify removal, focused on whether anything Cloudflare still relies on gets broken. Good news first: the deleted edge functions are all verifiably re-implemented in `cloudflare/_worker.js`, no remaining consumers of the deleted scripts/steps/`NETLIFY_*` vars exist, and the RC Slack URLs in `publish.yml` match the `rc-<x-y-z>` label `docs-staging.yml` deploys to. However, there is one blocker (dropped `GITHUB_TOKEN` on the production assemble step) plus several smaller issues — see inline comments.

docs/deploy/build.sh
```diff
@@ -0,0 +1,13 @@
+#/bin/bash
+# Local full build of the multi-version docs site (current + previous versions),
+# assembled into ./docs ready to deploy to Cloudflare Pages.
+rm -rf docs
+mkdir -p docs/docs
+cd ..
+BUILD_MODE=preview npm run build
+cd deploy
+cp -r ../dist/. docs/
```
demtario: The Astro `dist/` is flat relative to `/docs` (`base: '/docs'` in `astro.config.mjs` is not part of the output tree), so this copies the current version's pages into the **publish root** instead of under `docs/docs/`. Both CI paths nest them: staging does `cp -r ../dist/. preview/docs/`, and production's docker cp lands the site under `deploy/docs/docs`. The result of running this helper: current-version pages are served at `/<page>` while previous versions (line 10) land correctly at `/docs/<version>/`, and every internal link/asset and redirect rule targets `/docs/...` — the assembled site 404s at all canonical URLs. The `mkdir -p docs/docs` on line 5 suggests nesting was intended and the copy destination was mistyped.

```suggestion
cp -r ../dist/. docs/docs/
```

While here: the shebang on line 1 is `#/bin/bash` (missing `!`), and unlike its two sibling scripts this one has no `set -euo pipefail`. Consider also copying `../dist/404.html` to `docs/404.html`, mirroring `build_current_version.sh` and the staging step.

docs/deploy/build.sh
```diff
@@ -0,0 +1,13 @@
...
+./build_previous_versions.sh
+cp _redirects docs/_redirects
+# Deploy the assembled site to Cloudflare Pages, e.g.:
+#   npx wrangler pages deploy docs --project-name handsontable-docs --branch production
```
demtario: This helper never copies `cloudflare/_worker.js` into the assembled `docs/` dir (nor substitutes `__LATEST_DOCS_VERSION__`), yet this comment tells the operator to deploy that dir straight to the **production** project. Following it — most likely during an incident when CI is down, exactly when someone reaches for a local build helper — publishes production without the worker, so Cloudflare Pages silently falls back to the stale `_redirects` and every worker-only redirect breaks site-wide: the same DEV-1969 outage class the `test -f docs/_worker.js` tripwire in `docs-production.yml` was added to prevent, with no tripwire on this path. Either copy + substitute the worker here (which would also let the workflow reuse the script), or change the example to a non-production target with an explicit warning.

.github/workflows/docs-staging-delete.yml
```diff
@@ -25,44 +25,6 @@ jobs:
...
-      - name: Run the delete process
-        run: |
-          cd netlify
-          node deleteIfExist.mjs
-        env:
-          NETLIFY_AUTH_TOKEN: ${{ secrets.NETLIFY_PREVIEW_AUTH_TOKEN }}
-          BRANCH_NAME_PREFIX: dev-handsontable-
-          BRANCH_NAME: ${{ steps.get-pr.outputs.source_branch }}
-
       - name: Delete Cloudflare Pages deployments for branch
```
demtario, line 51: Cleanup now only ever deletes `pr-<N>` deployments, but the reworked `cf-target` step in `docs-staging.yml` introduces two more label families that nothing ever deletes: `rc-<version>` (release builds) and sanitized branch slugs (`workflow_dispatch` on arbitrary branches). Concrete case: a release PR closes — its docs preview went to `rc-16-1-0`, not `pr-<N>` (PR builds are skipped for `release/*` heads), so this workflow finds nothing and the stale docs keep serving at `rc-16-1-0.handsontable-docs-staging.pages.dev` forever. The removed Netlify `deleteIfExist.mjs` keyed off the PR's *source branch*, so it covered branch-named previews on PR close; that coverage is lost. Cloudflare Pages has no auto-expiry, so orphaned public previews at guessable URLs accumulate unbounded. Suggested fix: invoke `deleteDeployments.mjs` additional times with `CF_BRANCH_NAME` set to the slugified `github.event.pull_request.head.ref`, and `rc-<version>` when the head is a `release/x.y.z` branch.

.github/workflows/docs-staging-delete.yml
```diff
@@ -25,44 +25,6 @@ jobs:
...
         if: always()
         continue-on-error: true
```
demtario: These two flags existed so the Cloudflare cleanup would still run when the *preceding Netlify steps* failed (DEV-1787). With those steps gone, this is now the workflow's only functional step, and `continue-on-error: true` swallows its only failure signal: an expired `CLOUDFLARE_API_TOKEN` means previews are never deleted while every run stays green — the exact silent-failure pattern this PR removes `continue-on-error` for in four other places, with comments explaining why. Suggest dropping `if: always()` + `continue-on-error: true` (or commenting why green-on-failure is intended here). Related dead weight above: nothing in this workflow installs packages anymore (`deleteDeployments.mjs` imports only the local `cfFetch.mjs` and built-ins), so the `pnpm/action-setup` step and setup-node's `cache: pnpm` / `cache-dependency-path` lines just restore the multi-hundred-MB monorepo cache for nothing on every PR close.

.github/workflows/docs-staging.yml
```diff
@@ -191,82 +149,78 @@ jobs:
...
-      #     preview at pr-<N>.handsontable-docs-staging.pages.dev.
-      # Any deploy whose branch differs from the project's production_branch
-      # is a preview, so the branch passed here determines apex vs preview.
-      # Cloudflare Pages deploy is independent of Netlify: !cancelled() lets
-      # these steps run even when the Netlify publish above fails, so the apex
-      # handsontable-docs-staging.pages.dev still updates.
+          cp ../dist/404.html preview/404.html
+
+          cp _dev_headers preview/_headers
+          cp _redirects preview/_redirects
```
demtario: This `_redirects` copy — and the rest of the Netlify-era `_redirects` machinery this PR keeps — is dead on Cloudflare: Pages ignores `_redirects` whenever `_worker.js` is present, and the PR itself now hard-fails any deploy where the worker is missing. Retained pieces: this line; the tracked 28 KB `docs/deploy/_redirects`; the `echo ... >> _redirects` appends in `build_current_version.sh:23` and `build_previous_versions.sh:17` (which **mutate the git-tracked file in place** — every local run dirties the checkout and accumulates duplicate rules); and the `cp _redirects` in `build_current_version.sh:55` and `build.sh:11`. Beyond wasted bytes, the file keeps looking authoritative, so the next redirect author edits `_redirects` instead of `_worker.js` — recreating the split-brain the PR's own docs ("the worker is the **sole** authority") were written to eliminate; and in the one state where `_redirects` would activate (worker missing), it serves years-stale rules. Suggest deleting the file, the appends, and all the copies. (The `_dev_headers` copy on the line above is fine — verified live that Pages applies `_headers` to worker-served assets.)

docs/cloudflare/_worker.js
```diff
@@ -2,8 +2,8 @@
  * Cloudflare Pages Worker for Handsontable docs preview site.
  *
  * When a _worker.js is present in Cloudflare Pages, it intercepts ALL requests
- * and _redirects is ignored. This worker re-implements all redirect logic that
- * lives in the Netlify _redirects file and the Netlify edge functions.
+ * and _redirects is ignored. This worker is the sole, hand-maintained authority
```
demtario, line 5: Three genuine Netlify leftovers survive the "no stale references" sweep (repo-wide grep, excluding the intentionally-kept `hot-docs-assistant.netlify.app` backend):

1. **Functional:** line 751 — the CSP `frame-src` still allowlists `https://app.netlify.com` (Netlify's deploy-preview drawer UI). With Netlify gone, this is a pointless loosening of the shipped production CSP: any page on app.netlify.com may be iframed into handsontable.com/docs.
2. Cosmetic: line 798 — the rule-1b comment still explains the redirect as "matches the Netlify `_redirects` rule…", pointing at a file this PR positions the worker as the sole authority over.
3. Cosmetic: `.vscode/test-vue-redirects.sh:2` and `:10` — the usage/example URLs are dead `*.netlify.app` hosts; should point at `https://pr-<N>.handsontable-docs-staging.pages.dev`.

.github/workflows/docs-staging.yml
```diff
@@ -191,82 +149,78 @@ jobs:
...
         with:
           script: |
             const prNumber = '${{ steps.resolve-pr.outputs.number }}';
+            const isRelease = '${{ steps.release-info.outputs.is_release }}' === 'true';
+            const targetVersion = '${{ steps.release-info.outputs.target_version }}';
             const ref = context.ref || '';
+            const headBranch = ref.replace(/^refs\/heads\//, '');
+
+            // Sanitize a git branch name into a Cloudflare Pages branch label.
+            const slug = (s) => s.toLowerCase().replace(/[^a-z0-9]+/g, '-').replace(/^-+|-+$/g, '').slice(0, 40) || 'branch';
```
demtario: Two issues with the 40-char cap:

1. Cloudflare Pages truncates preview-branch subdomain aliases to **28 chars** (then trims dashes) — confirmed against Cloudflare staff's published algorithm and real-world reports. So for slugs of 29–40 chars, the preview is served at a *different* hostname than the `<slug>.handsontable-docs-staging.pages.dev` pattern documented in the comment above and in `README-DEPLOYMENT.md` (this PR's own branch slugs to exactly 28 chars — one char from the cliff). Two long branches sharing their first 28 sanitized chars also silently overwrite each other's preview.
2. `.slice(0, 40)` runs *after* the dash-trim, so a truncated label can end in `-` (cosmetic only).

```suggestion
            const slug = (s) => s.toLowerCase().replace(/[^a-z0-9]+/g, '-').slice(0, 28).replace(/^-+|-+$/g, '') || 'branch';
```

Longer-term: this replaces the deleted `docs/netlify/subdomainHelpers.mjs`, which was a tested module — consider housing the label logic in `docs/cloudflare/` next to `ensureProject.mjs`/`deleteDeployments.mjs` so it's unit-testable and shareable with the delete workflow.

.github/workflows/docs-production.yml
```diff
@@ -167,15 +152,15 @@ jobs:
           # branch here would move the apex on every release.
           CF_PAGES_APEX_BRANCH: production
 
+      # Cloudflare Pages is the sole production host, so a failed deploy must
+      # fail the workflow (no continue-on-error here).
       - name: Deploy production to Cloudflare Pages
```
demtario, line 169: Note (non-blocking): the deleted Netlify publish steps in both this workflow and `docs-staging.yml` were wrapped in `nick-fields/retry` (3 attempts). Their Cloudflare replacements were always single-attempt, but they used to be non-fatal — now that `continue-on-error` is (correctly) gone, a single transient Cloudflare API blip fails the whole workflow: here after ~20+ min of Docker builds and assembly; in staging it also fires the RC-failure Slack alert on release branches. `wrangler` retries asset uploads internally but not deployment creation, and `npm install wrangler` is an unretried network op too. Consider wrapping both wrangler deploy steps in the same `nick-fields/retry` — hard-failure semantics are preserved after exhausted attempts.

docs/README.md
```diff
@@ -85,24 +86,24 @@ The documentation branches are created and updated automatically by the `stable-
...
+The prod-docs/latest branch is automatically recreated by the CI/CD pipeline whenever a patch or release update is applied to the latest documentation version. This branch triggers a GitHub workflow that initiates a rebuild and deploys to Cloudflare Pages on each push or when a new branch `prod-docs/<MAJOR.MINOR>` is created.
 
-Committing directly to the Documentation production branch triggers GitHub workflow that deploys the changes to Netlify. The exception is the `develop` branch that holds the changes for the "next" version. The staging version can be deployed only [manually](./README-DEPLOYMENT.md#manually-deploying-the-documentation-to-the-staging-environment).
+Committing directly to the Documentation production branch triggers GitHub workflow that deploys the changes to Cloudflare Pages. The exception is the `develop` branch that holds the changes for the "next" version. The staging version can be deployed only [manually](./README-DEPLOYMENT.md#manually-deploying-the-documentation-to-the-staging-environment).
```
demtario: Pre-existing but worth fixing while this line is being edited: the anchor `#manually-deploying-the-documentation-to-the-staging-environment` matches no heading in `README-DEPLOYMENT.md` (before or after this PR) — the actual heading is `### Deploying the documentation to the staging environment`, i.e. `#deploying-the-documentation-to-the-staging-environment`. The same dead anchor is in `docs/README-EDITING.md:120`.

demtario (approved):

budnix (approved):

Issue #11761, opened 2025-07-30
Width of last column is ignored when strechH is set to 'last'


windmichael:
### Describe the bug

When setting the the property stretchH to 'last' in the Handsontable configuration, the width of the last column is not considered anymore. 
When there is not enough space Handsontable will shrink the column with and ignores what is defined as width of the column. 

### Provide a link to the demo with the bug reproduction

https://stackblitz.com/edit/36zpj661-ahsnvj4k?file=index.js

### Handsontable version

>15.3

### Your environment

Chrome, Edge

AMBudnik:
Thank you for sharing the issue report, @windmichael 

Based on the specification of the `stretchH` option it is said that is stretches given column/s. Could you confirm that you would like the last column to keep `300 pixels`? If so, why the `stretchH last` is used?

windmichael:
@AMBudnik stretching the last column is fine. The issue is that the width of the last columns shrinks under the specified width of the column. It seems that the column even shrinks to a minimum width of the column.

<img width="1232" height="602" alt="Image" src="https://github.com/user-attachments/assets/508ff88e-6ad2-4f7a-9a1b-3ac6aafff251" />

I updated the stackblitz example. https://stackblitz.com/edit/36zpj661-ahsnvj4k?file=index.js

adrianspdev:
Fixed in the [PR](https://github.com/handsontable/handsontable/pull/12123), the update will be included in the next release.

PR #12837, opened 2026-06-24
docs(changelog-17): clarify stretchH 'last' width fix in 17.1.0 (#11761)

AMBudnik:
## Summary

Clarifies the changelog entry for issue [#11761](https://github.com/handsontable/handsontable/issues/11761) in the 17.1.0 Fixed section to accurately describe the corrected behavior of `stretchH: 'last'`.

## What changed

The previous entry for #11761 appeared twice with slightly different wording, and neither version was explicit enough about the corrected contract. This PR:

- **Merges** the two duplicate #11761 entries into one authoritative entry.
- **Expands** the entry to clearly state that as of 17.1.0, the `stretchH: 'last'` option always honors the column's defined `width` value — the last column will never be rendered below its configured width regardless of viewport size.
- **Adds** a brief before/after note: prior to 17.1.0, a narrow container could collapse the last column to `0px` even when an explicit `width` was set; this is no longer the case.
- **Removes** the implication (via the duplicate/ambiguous phrasing) that collapsing to `0px` was an accepted trade-off.

## Motivation

An eval test (V171-003) revealed that the existing changelog wording was ambiguous enough to lead an AI assistant to incorrectly conclude that `0px` collapse under `stretchH: 'last'` remained expected behavior in 17.1. The updated entry makes the fix unambiguous for both human readers and automated tooling.

## Files changed

- `docs/content/guides/upgrade-and-migration/changelog-17/changelog-17.md`

### DIFF
--- a/docs/content/guides/upgrade-and-migration/changelog-17/changelog-17.md
+++ b/docs/content/guides/upgrade-and-migration/changelog-17/changelog-17.md
@@ -52,7 +52,7 @@ For more information about this release, see:
 #### Fixed
 - Fixed an issue where the Nested Rows plugin was disabled after calling updateSettings with an empty data array. [#10556](https://github.com/handsontable/handsontable/issues/10556)
 - Fixed `setSourceDataAtCell()` updating parent rows instead of nested child rows when `nestedRows` is enabled. [#10657](https://github.com/handsontable/handsontable/issues/10657)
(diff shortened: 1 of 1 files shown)


qunabu:
Task linked: [DEV-171 Changing columnSummary's className metadata breaks sum calculation](https://app.clickup.com/t/9015210959/DEV-171)

PR #12883, opened 2026-06-29
DEV-650: Fix error when removing frozen column with fixedColumnsLeft

demtario:
### Context

The PR fixes an error thrown when removing a frozen column while only the legacy `fixedColumnsLeft` option is set. During `alter('remove_col', ...)`, the engine decremented `tableMeta.fixedColumnsStart`, which invoked the watched setter on `globalMeta.meta` and incorrectly tripped the `fixedColumnsLeft`/`fixedColumnsStart` usage-conflict guard. The fix writes `_fixedColumnsStart` directly for this internal update, matching the existing pattern in `manualColumnFreeze`.

### How has this been tested?

- Added E2E regression test in `fixedColumnsStart.spec.js`
- `npm run test:e2e --prefix handsontable -- --testPathPattern=fixedColumnsStart` (41 specs, 0 failures)
- `npm run test:e2e --prefix handsontable -- --testPathPattern=manualColumnFreeze` (27 specs, 0 failures)
- `npm run test:unit --prefix handsontable -- --testPathPattern=extendMetaProperties` (16 tests, 0 failures)
- `npm run build --prefix handsontable`
- `npm run eslint --prefix handsontable`
- `npm run test:types --prefix handsontable`

### Types of changes

### Related issue(s):

1. DEV-650

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-650

---

> [!NOTE]
> **Low Risk**
> Small, localized change aligned with existing `manualColumnFreeze` behavior, with new e2e tests and no public API changes.
> 
> **Overview**
> Fixes a crash when deleting a column while only the legacy **`fixedColumnsLeft`** option is set. Internal updates during **`remove_col`** (and undo of that action) now adjust **`_fixedColumnsStart`** instead of going through the **`fixedColumnsStart`** setter, so the engine no longer falsely triggers the “don’t use both option names” guard.
> 
> Regression coverage was added for column removal and undo/redo with **`fixedColumnsLeft`**, plus a changelog entry.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 401697928a4e64980f72dfb46c146749f8bfff78. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/handsontable/src/core.ts
+++ b/handsontable/src/core.ts
@@ -1103,7 +1103,12 @@ export default function Core(
               const fixedColumnsStart = tableMeta.fixedColumnsStart;
 
               if (fixedColumnsStart >= calcIndex + 1) {
-                tableMeta.fixedColumnsStart -= Math.min(groupAmount, fixedColumnsStart - calcIndex);
+                // Since 12.0.0, the "fixedColumnsLeft" is replaced with the "fixedColumnsStart" option.
+                // However, keeping the old name still in effect. When both option names are used together,
+                // the error is thrown. To prevent that, the engine needs to modify the original option key
+                // to bypass the validation.
+                (tableMeta as unknown as { _fixedColumnsStart: number })._fixedColumnsStart -=
+                  Math.min(groupAmount, fixedColumnsStart - calcIndex);
               }
 
               if (Array.isArray(tableMeta.colHeaders)) {
--- a/.changelogs/12883.json
+++ b/.changelogs/12883.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Fixed an error thrown when removing a frozen column while using the legacy fixedColumnsLeft option.",
+  "type": "fixed",
+  "issueOrPR": 12883,
+  "breaking": false,
+  "framework": "none"
+}
--- a/handsontable/src/__tests__/settings/fixedColumnsStart.spec.js
+++ b/handsontable/src/__tests__/settings/fixedColumnsStart.spec.js
@@ -350,6 +350,20 @@ describe('settings', () => {
       });
     });
 
+    it('should not throw when removing a frozen column using only fixedColumnsLeft', async() => {
+      handsontable({
+        data: createSpreadsheetData(5, 5),
+        fixedColumnsLeft: 2,
+      });
+
+      expect(getSettings().fixedColumnsStart).toBe(2);
+
+      await alter('remove_col', 0);
+
+      expect(getSettings().fixedColumnsStart).toBe(1);
+      expect(getInlineStartClone().find('tbody tr:eq(0) td').length).toBe(1);
+    });
+
     it('should limit fixed columns to dataset columns length', async() => {
       handsontable({
         data: createSpreadsheetData(3, 3),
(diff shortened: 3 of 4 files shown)


qunabu:
Task linked: [DEV-650 Add not note: Error Throws When Removing Fixed Columns when `fixedColumnsLeft` for v12+ is in use](https://app.clickup.com/t/9015210959/DEV-650)

handsontable/src/core.ts
```diff
@@ -1103,7 +1103,12 @@ export default function Core(
               const fixedColumnsStart = tableMeta.fixedColumnsStart;
 
               if (fixedColumnsStart >= calcIndex + 1) {
-                tableMeta.fixedColumnsStart -= Math.min(groupAmount, fixedColumnsStart - calcIndex);
+                // Since 12.0.0, the "fixedColumnsLeft" is replaced with the "fixedColumnsStart" option.
+                // However, keeping the old name still in effect. When both option names are used together,
+                // the error is thrown. To prevent that, the engine needs to modify the original option key
+                // to bypass the validation.
+                (tableMeta as unknown as { _fixedColumnsStart: number })._fixedColumnsStart -=
```
budnix, line 1110: The undo path for `remove_col` (`RemoveColumnAction.undo()` in `plugins/undoRedo/actions/removeColumn.ts:154`) still writes `settings.fixedColumnsStart` through the public setter. So pressing Ctrl+Z after removing a frozen column on a `fixedColumnsLeft`-only table will throw the same error. Worth applying the `_fixedColumnsStart` bypass there too.

budnix, line 1110: Writing `_fixedColumnsStart` directly skips the setter, so `usageTracker` (https://github.com/handsontable/handsontable/blob/feature/DEV-650/handsontable/src/dataMap/metaManager/mods/extendMetaProperties.ts#L129) doesn't record `'fixedColumnsStart'`. If the user later calls `updateSettings({ fixedColumnsStart: 2 })`, the setter fires and adds `'fixedColumnsStart'` to the tracker — now both keys are in the set and `onChange` throws, even though the user never set both options. This is another path where the guard fires when it shouldn't.

demtario: fixed

demtario: I believe, this throw can't actually be reached without the user explicitly setting `fixedColumnsLeft`. The guard needs both `'fixedColumnsLeft'` and `'fixedColumnsStart'` in `usageTracker`, and a key is added only when its own named setter fires. Nothing in the engine ever writes the `fixedColumnsLeft` alias (grep confirms — internal writes only ever touch `fixedColumnsStart`/`_fixedColumnsStart`), so 'fixedColumnsLeft' enters the tracker only via an explicit user assignment.

So in the "internal set, then `updateSettings({ fixedColumnsStart: 2 })`" case, the tracker only ever holds `'fixedColumnsStart'` → no throw. Verified both paths:
- `manualColumnFreeze.freezeColumn()` (internal write) → `updateSettings({ fixedColumnsStart })` → no throw.
- `fixedColumnsStart`-only → `alter('remove_col')` → `updateSettings({ fixedColumnsStart })` → no throw.

The bypass also only ever reduces throws: writing `_fixedColumnsStart` directly stops internal updates from adding `'fixedColumnsStart'` to the tracker, and that tracker is referenced solely by this guard.

The one case that does throw is the user explicitly setting both option names (either order) — that reproduces on develop with no column removal, so it's pre-existing and independent of this change.

Is this the case you were thinking, or maybe I misunderstood?

budnix: That sounds good. I think we can leave it as is.

budnix (approved):

PR #12925, opened 2026-07-01
DEV-1982: Replace hooks soft-delete array with a linked-list bucket

budnix:
### Context

The PR changes the hook bus (`HooksBucket`) from an array-with-soft-delete to a linked list, as part of the rendering/memory performance initiative.

Each hook's callbacks were stored in an array with a soft-delete `skip` flag: a removed callback stayed in the array and was iterated (then skipped) on every `run()` until a `MAX_SKIPPED_HOOKS_COUNT` (100) compaction pass rebuilt the array. `#runHandlers` also re-read `handlers[index]` several times per entry and checked `.skip` on every iteration.

This replaces that with a purpose-built intrusive singly-linked list per hook (`{head, tail}`; `HookEntry` drops `skip`, gains `next`). The soft-delete `skip` flag, `#skippedHooksCount`, and the `MAX_SKIPPED_HOOKS_COUNT` compaction are removed entirely. `remove()` is now a true relink-delete; `#runHandlers` walks `head → next` with no per-entry skip or bounds check.

The observable dispatch contract is preserved:

- **Mid-run removal** of a later, not-yet-fired handler is still skipped this run — `entry.next` is read *after* each callback, and `remove()` never nulls a removed node's `next` (so a callback removing itself can still advance).
- **Mid-run add** does not fire this run — the `tail` captured at loop start is the boundary.
- **`getHooks()` reference stability** (relied on by the Vue 3 wrapper) — it returns the live entry objects, and `addAsFixed` swaps `.callback` in place on the existing entry.

Two intentional behavior changes, both non-breaking (no public API, default, hook, or CSS change):

- `has()` now returns `false` once the last hook is removed. The array version could return `true` for a collection holding only soft-deleted entries — a latent bug, now fixed.
- Re-adding a previously removed callback inserts a fresh entry at its `orderIndex` position, rather than "re-enabling" the old entry in place.

### Results

Measured with a standalone harness driving the real rebuilt bundle (`Handsontable.hooks`), profiler off, median of 7 runs, same machine. Metric is the latency of a single `hooks.run()` at the given number of registered callbacks on one hook.

| `run()` over | Before (array + soft-delete) | After (linked list) | Speedup |
| --- | --: | --: | --: |
| 1000 live handlers | 8,300 ns | 4,940 ns | 1.68× |
| 500 live (after removing 500) | 4,787 ns | 2,480 ns | 1.93× |
| 901 live, 0 removed | 8,193 ns | 4,931 ns | 1.66× |
| 901 live + 99 removed | 8,707 ns | 5,180 ns | ≈ the 901-live row — no skipped slots to iterate |

The last two rows isolate the soft-delete cost: before, 99 removed-but-not-compacted entries added ~514 ns of empty iteration per `run()`; after, removed entries are gone, so that row collapses onto the plain 901-live figure.

**Scope note (honesty):** the speedup is a large ratio on a small absolute. Real hooks on the render hot path have 0–1 handlers, where dispatch is already sub-microsecond; across a full render frame the difference is on the order of ~0.01 ms. The primary value of this PR is deleting the skip/compaction machinery and making removal a true delete; the throughput win only becomes material for hooks that accumulate many handlers.

### How has this been tested?

- Unit: `npm run test:unit --prefix handsontable --testPathPattern=core/hooks` — 40/40 pass (bucket + index).
- Types: `npm run test:types --prefix handsontable` — clean (downlevel-dts green).
- Build: `npm run build --prefix handsontable` — OK.
- E2E: `npm run test:e2e --prefix handsontable --testPathPattern=core/runHooks` — 6/0; `--testPathPattern=__tests__/hooks/` — 333/0.
- Lint: `npm run eslint --prefix handsontable` — clean.
- Added a unit test asserting `getHooks()` reference stability across an `addAsFixed` callback swap (the wrapper contract), plus true-delete / head-tail-removal / no-skipped-accumulation cases.

### Types of changes

### Related issue(s):

1. DEV-1982

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-1982

### DIFF
--- a/.changelogs/12925.json
+++ b/.changelogs/12925.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Improved the performance of the internal hooks (events) dispatch: each hook's callbacks are now stored in a linked list, and removing a hook frees it immediately instead of accumulating skipped entries that were iterated over on every run.",
+  "type": "changed",
+  "issueOrPR": 12925,
+  "breaking": false,
+  "framework": "none"
+}
--- a/handsontable/src/core/hooks/__tests__/bucket.unit.ts
+++ b/handsontable/src/core/hooks/__tests__/bucket.unit.ts
@@ -1,5 +1,15 @@
 import { HooksBucket } from 'handsontable/core/hooks/bucket';
 
+// The entries are intrusive linked-list nodes (they carry a `next` pointer). Project the comparable
+// fields so `toEqual` does not walk the `next` chain.
+const shape = hookEntry => ({
+  callback: hookEntry.callback,
+  orderIndex: hookEntry.orderIndex,
+  runOnce: hookEntry.runOnce,
+  initialHook: hookEntry.initialHook,
+});
+const shapes = hooks => hooks.map(shape);
+
 describe('Bucket', () => {
   describe('add()', () => {
     it('should add hook to the collection', () => {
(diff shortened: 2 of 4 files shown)


qunabu:
Task linked: [DEV-1982 Replace hooks soft-delete array with a linked-list bucket](https://app.clickup.com/t/9015210959/DEV-1982)

demtario (approved):

PR #12951, opened 2026-07-03
DEV-1995: Walkontable single-pass layout - scrollbar prediction + sizing ports

budnix:
### What this changes

This is an **internal refactor of the grid's rendering engine** (Walkontable). It does not change how the grid looks or behaves for users — every feature works exactly as before.

Two areas were reworked under the hood:

1. **A more predictable drawing step.** Previously the engine drew the cells, then measured the result to find out whether scrollbars were needed, and in some cases had to redraw. For standard fixed-size tables it now works this out up front — from the row and column sizes and the container size — so it can draw in a single pass.

2. **Cleaner internal structure.** The engine's own size handling was separated from the optional auto-sizing plugins (AutoRowSize / AutoColumnSize) behind clear boundaries, and the large source files were reorganized into smaller, clearly named folders.

Tables using merged cells (and a few other special cases) automatically keep the previous, proven drawing path, so nothing regresses for them.

### New `modifySinglePassLayout` hook

The choice between the new single-pass drawing and the previous measure-then-render path is exposed through a new public hook, `modifySinglePassLayout`. Single-pass is on by default; returning `false` from the hook forces the previous path for that table. It receives the current value (`true`/`false`) and returns the value to use.

This replaces a hardcoded internal check: merged cells now opt out through this hook instead of the engine naming the plugin directly. Plugin authors and users can use it too — e.g. to keep the previous drawing behavior for a specific grid.

```js
  // Force the previous (measure-then-render) drawing path for this table.
  modifySinglePassLayout: () => false,
});
```

### Why

This is **groundwork, not an optimization in itself.** The aim is a cleaner, more predictable rendering flow that is easier to maintain and that makes future performance improvements possible. The change is intentionally behavior-preserving — the actual speed-ups will come as separate, focused changes built on top of this foundation.

### Performance

No regressions, and memory is stable (no leaks; heap flat after heavy scrolling). Real-browser measurements (per full draw, both builds on the same machine):

**Row (vertical) scrolling — no change** — on par with `develop`, top and bottom of the dataset alike.

**Horizontal scrolling — 100 × 1,000**

| Region | develop | branch | Δ |
| --- | --- | --- | --- |
| At start | 7.11 ms | 6.94 ms | −2.4% (branch faster) |
| At end | 3.56 ms | 3.11 ms | −12.5% (branch faster) |

This PR is the enabling refactor; further targeted speed-ups follow later.

### How has this been tested?

- `test:walkontable` — full rendering-engine suite passes.
- `test:unit` and `test:types` — pass.
- `eslint` — clean.
- The new up-front drawing path was checked against the previously measured behavior for standard tables; special cases (e.g. merged cells) keep the original path.

### Types of changes

### Related issue(s):

1. DEV-1995

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-1995

---

> [!NOTE]
> **High Risk**
> Touches the core draw loop, viewport calculators, and fragile overlay positioning; regressions would show as scrollbar flicker, misaligned frozen regions, or wrong visible ranges despite escape hatches and broad test coverage.
> 
> **Overview**
> Walkontable gains a **`singlePassLayout`** path (default on for typical grids) that predicts scrollbar presence from row/column totals and container size via a pure **`resolveLayout`** solver in `viewport/boxLayout/`, so many draws avoid measure-then-redraw. Merged cells, window-scrolled tables, and non-uniform sizes keep the legacy measured path; **`modifySinglePassLayout`** lets callers override that choice.
> 
> The draw is orchestrated through **`runDrawCycle`** (`table/drawCycle.ts`) with separate master/clone cycles. Row/column sizing moves behind **`AxisSizeSource`** ports in **`axisSizing/`**; overlays are split into collaborators (`ScrollSync`, `SpreaderSize`, `ResizeMonitor`, etc.), and **`geometry/`** is renamed to **`domMeasure/`**. On the gated path, header-border classes can be applied pre-render (`prepareHeaderBorders`), the second viewport calculator pass may be skipped when caches stay current, and vertical scroll can skip redundant column-header work when layout is unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 524d0bf58c8d4177142b8616868c8caec500147f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-1995 Walkontable single-pass layout: scrollbar prediction + sizing ports](https://app.clickup.com/t/9015210959/DEV-1995)

qunabu (dismissed):

adrianspdev (dismissed):

PR #13064, opened 2026-07-13
DEV-2055: Dev-owned test enforcement + one PR pipeline

danielzytohoc:
ClickUp: [DEV-2055](https://app.clickup.com/t/86capukbv)

Replaces conversational/manual QA with **machine-enforced, dev-owned testing**, so QA stops scaling with PR count — and unifies the PR CI into **one layered pipeline**.

## Enforcement (agent-first, local + CI)

- **Presence gate** — a `handsontable/src/**` / `wrappers/**` change must ship a matching test change (CI job, **warn mode** for now; `Refactor-only: <reason>` trailer as the declared escape).
- **Local gates, auto-installed by `pnpm install`** — pre-commit (`eslint --fix` staged specs/source: determinism + anti-gaming rules) and pre-push (presence gate → eslint → test-weakening detector → runs the changed Playwright specs). Claude Code agent hooks add an earlier authoring-time layer. Exact rules: `.ai/LOCAL-ENFORCEMENT.md`.
- **Anti-gaming guards** ("green is not the goal") — focused tests = **error**; skipped / assertion-free tests = warn; a diff-based **test-weakening detector** flags specs that lost assertions or gained skips (escalates when source changed in the same diff).
- **Determinism lint** — no new `sleep()` / `waitForTimeout` / `networkidle` / `it.flaky`.

## Playwright E2E tier (`tests/`)

- New E2E is **Playwright** (`tests/e2e/*.spec.ts`); the Jasmine/Puppeteer suite is **frozen** (edit existing, add none; broken ones migrate).
- pnpm **workspace member** with catalog-pinned `@playwright/test` (1.61.1 — matches the CI container tag; they bump together). POM + `data-testid` hooks + web-first waits; `failOnFlakyTests` in CI.
- Specs: grid smoke ×3, walkontable overlays ×2, granular keyboard-nav ×1 — all green against a fresh dist.

## One PR pipeline

All former parallel PR workflows now run as `workflow_call` **fragments inside the Tests pipeline** (implementations stay in their own files): changelog · lint · audit (warn) · visual-tests lint · code-quality (Sonar/CodeQL/FOSSA) · performance (now `src/**`-scoped) · verify-emitted-types · docs-angular-typecheck · pkg-pr-new. Trunk baselines (push→develop) kept. `linter.yml`/`audit.yml` no longer fire on every branch push. Job names de-truncated (`[TEST] Handsontable: X` → `[TEST] X`).

[...]

PR #13065, opened 2026-07-13
DEV-2055: Dev-owned test enforcement + one PR pipeline

danielzytohoc:
### Context

Replaces manual QA with **machine-enforced, dev-owned testing**, and consolidates the PR CI into one modular pipeline. North star: *a low number of extremely meaningful tests* — coverage is the floor, mutation kill-rate the ceiling, and "green is not the goal, correct behavior is." Nothing here weakens a test to pass.

**Scope guardrail:** this PR changes the **PR pipeline and the local dev loop only**. develop / rc / release / stable run the **same tests as before** (see *Trunk/release safety*). Flipping the new gates from *warn* to *blocking* (branch protection) is deliberately deferred.

`[skip changelog]` — no shippable `handsontable/src/**` change (tooling, CI, tests, skills, docs).

**1. Test enforcement — the loop.** Presence gate (source change ⇒ matching test change, with a `Refactor-only:` trailer escape; new Jasmine `*.spec.js` blocked), test-weakening detector (removed/loosened assertions, added skips/focus), a changed-line coverage floor (% of added executable lines covered, report-only at 80%), and scoped StrykerJS mutation as the meaningfulness ceiling (opt-in, not per-PR).

**2. Skills & lint rules.** Two skills authored as **Claude Code** skills in `.claude/skills/` (`test-writing-discipline`, `handsontable-playwright-e2e`) — the canonical source — and **mirrored to Cursor** rules (`.cursor/rules/`, git-ignored/generated) by the `prepare` script, so Claude Code and Cursor share the same guidance from one source. Plus 5 custom ESLint rules (no-focused-test, no-skipped-test, require-assertion-in-test, no-fixed-sleep-in-spec, no-new-it-flaky).

**3. Local enforcement (hooks).** Escalating chain, auto-installed by `pnpm install`. The **git hooks (lefthook) are the floor for everyone** — Cursor, Claude Code, and plain-git alike: pre-commit (lint staged) → pre-push (presence gate block, weakening warn, changed specs + unit tests). The **Claude Code agent hooks add an earlier layer** on top for that editor: PostToolUse (lint the edited spec) and Stop (run session-touched Playwright + unit tests at turn end). Touched E2E specs run exactly once whichever tool proves them first, via a content-keyed green-run cache shared by the Stop hook and pre-push (so a Cursor user's pre-push proves it; a Claude user's Stop does, and the other skips).

**4. Evals.** A zero-dep meaningfulness scorer + live mutation layer (`evals/`), with reference fixtures scored as a self-test so a skill/prompt regression is caught before it scales. Maintainer tool — run on skill changes, not per-PR.

**5. CI — one modular PR pipeline.** `test.yml` is now a ~185-line orchestrator (the DAG only); jobs/steps live in 14 reusable-workflow modules + 6 composite actions, so checks read as `Module / job` (`E2E / UMD (theme: main)`, `Code Quality / CodeQL`, …). Tiered DAG: tier-0 gatekeepers (scope/presence/changelog) → tier-1 no-build gates → build (gated on the fast checks, not the slow scans) → tests → visual (post-test budgeting gate) → docs sub-DAG. Matrices collapse duplication (E2E themes, wrappers, visual); build-once/artifact-share (no 5× rebuild); 7 former standalone PR workflows folded into modules; one `pr-cleanup.yml` on close.

**6. Trunk/release safety.** `test.yml` keeps its filename, `workflow_call(test-all)` interface, and push triggers, so `publish.yml` (rc/stable) and develop/master push run the **same jobs** (one nesting level deeper); only check display names change on trunk (pinned nowhere until DEV-2059). Standalone trunk workflows untouched.

**7. Nightly baseline.** `code-quality-baseline.yml` restores the trunk CodeQL/Sonar/FOSSA scan on a schedule (scans are now PR-only), keeping Sonar's new-code baseline and CodeQL's default-branch dashboard fresh.

**8. New Playwright E2E tier (`tests/`).** POM + `data-testid` fixtures + a theme project matrix (main/horizon/classic) so every spec runs under every theme; the frozen Jasmine suite migrates by attrition.

- Unit tests added/modified (`*.unit.js`): the enforcement machinery's own `node --test` suites — `scripts/__tests__/{pre-push,lint-files,e2e-run-cache}.test.mjs`, `.github/scripts/__tests__/{presence-gate,test-weakening,diff-coverage}.test.mjs`, `evals/__tests__/score.test.mjs`, `scripts/claude/__tests__/session.test.mjs`.
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`): the new tier — `tests/e2e/grid.spec.ts`, `tests/e2e/keyboard-nav.spec.ts`, `tests/e2e/walkontable/overlays.spec.ts` (theme-parametrized).
- Type tests (`*.types.ts`) updated if public API changed: n/a — no public API change.
- For a bug fix — the spec that fails without this fix: n/a — not a bug fix.
- Demo page / recorded trace (for UI changes): n/a — no user-facing UI change.

### Commands run
```
node --test scripts/__tests__/*.mjs .github/scripts/__tests__/*.mjs evals/__tests__/*.mjs   # all green
actionlint .github/workflows/*.yml                                                            # clean
cd tests && npx playwright test                                                               # 18/18 (6 specs × 3 themes)
```
Plus a full green PR pipeline run (all E2E ×themes, wrappers, ESM/CJS build, walkontable, visual multi/cross, unit + coverage floor).

### Types of changes
- [x] New feature or improvement (non-breaking change which adds functionality) — the test-enforcement system + modular CI pipeline

### Related issue(s):
1. DEV-2055 (this PR)
2. Follow-ups: DEV-2057 (trunk/release verification + stable-test), DEV-2058 (nightly full matrix + PR e2e subset), DEV-2059 (warn→block + branch protection)
3. DEV-2061 (complete the eval/mutation loop); 86caqbt9u (coverage→SonarCloud)

### Affected project(s):
- [x] `handsontable` 
- [ ] `@handsontable/vue3`

### Checklist:

---

> [!NOTE]
> **High Risk**
> Reshapes the entire PR CI DAG and merge-adjacent gates; mis-scoped path filters or permission ceilings could skip required jobs or break release/RC test calls until caught in branch protection rollout.
> 
> **Overview**
> Introduces **machine-enforced, dev-owned testing** and folds scattered PR workflows into a single **`test.yml` orchestrator** that calls reusable modules (Checks, Lint, Build, E2E, Integration, Docs, etc.) with path-based scope routing and shared build artifacts.
> 
> **Enforcement (warn-first in CI, block locally on pre-push):** pure **presence gate** and **test-weakening** classifiers in `.github/scripts/lib/` with `node --test` coverage; PR jobs run them in **Checks** plus a **tooling-tests** job. A **diff-coverage** CLI reports % of added executable lines covered (80% floor, report-only). **Changelog** and **manual QA** checks read the **live PR body/comments** so re-run works without empty commits; manual QA blocks merge when the template tickbox is set until a non-author comments `/manual-qa passed`.
> 
> **Local loop:** documents and wires **lefthook** (pre-commit/pre-push) and **Claude Code** PostToolUse/Stop hooks; Playwright is the only allowed new E2E (`tests/e2e/`), with legacy Jasmine frozen and new `*.spec.js` blocked.
> 
> **Authoring:** new **`handsontable-playwright-e2e`** and **`test-writing-discipline`** skills (synced to Cursor via prepare), PR template test-evidence section, and stack/testing docs updated for Playwright 1.61 vs legacy Puppeteer.
> 
> **CI structure:** composite actions (`setup-workspace`, docs visual/angular, performance, ESM/CJS E2E); **E2E** module runs legacy Puppeteer matrices plus **Playwright per theme** in the pinned container; **nightly `code-quality-baseline.yml`** replaces per-push trunk scans; standalone **`audit`/`changelog`/`linter`/`pkg-pr-new`** PR workflows removed or inlined. **`publish.yml`** gains explicit permission ceilings for nested `test.yml` calls.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 581af3aec40ed282388d7539fea6018a7110d775. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2055 QA overhaul — Pillar 1: dev-owned test enforcement (retire the manual "testing" gate)](https://app.clickup.com/t/9015210959/DEV-2055)

danielzytohoc:
## Cursor bug-bot findings from #13064 — resolutions (fixed in f11109e53)

| Finding | Verdict | Resolution |
|---|---|---|
| **Pre-push gate never runs** (`import.meta.url` vs `argv[1]`) | ✅ valid (portability) | Guard now uses `pathToFileURL(argv[1]).href`; note Node canonicalizes `argv[1]` on POSIX (the gate demonstrably ran), but the naive template breaks on Windows/URL-escaped paths. Also: every git call/spawn now anchors to the repo root resolved from the script location — verified by running the hook from a foreign cwd (full gate + specs ran). |
| **Fixture server rejects all URLs** | ❌ false positive | `static-server.mjs` uses `path.join`, not `path.resolve` — `join(ROOT, '/x')` yields `ROOT/x` (join never treats the second segment as absolute). Traversal remains covered by the `startsWith(ROOT)` guard. Empirical: all 6 e2e specs load fixtures + dist through this server in CI-equivalent runs. |
| **Playwright CI skips tests-only PRs** | ✅ valid | `./tests/**` added to the `build-handsontable-umd` filter, so a tests-only PR builds the dist artifact and runs `test-e2e-playwright` (which keys off that output). |
| **Stop hook misses committed Jasmine** | ✅ valid | When the tree is clean (file already committed in-turn), `statusOf` now classifies against the trunk merge-base diff. Verified live: committing a new `*.spec.js` then firing Stop → blocked (exit 2). |
| **Hooks assume repo-root cwd** | ✅ valid | Stop + pre-push anchor spec-existence checks, Playwright `cwd`, and all git calls to the resolved repo root. |

danielzytohoc:
## Bugbot round 2 — all four fixed in `103f959b5` (threads resolved)

| Finding | Fix |
|---|---|
| **Lefthook missing from install** | `lefthook ^2.1.10` is now a root devDependency (it previously resolved only via a warm npx cache). Its redundant postinstall is deliberately ignored — `scripts/prepare-dev-env.mjs` owns hook wiring. |
| **CI report artifact path mismatch** | CI reporter `blob` → `html(open:never)+github` (blob is for sharded runs; we don't shard). Failure artifact now uploads `playwright-report` **and** `test-results` (traces/videos). |
| **Static server path check bypass** | Boundary is now separator-suffixed (`ROOT + path.sep`) — `${ROOT}-sibling` can no longer pass. |
| **Weakening gate wrong diff** | The CLI anchors at `merge-base(GATE_BASE, HEAD)` and diffs `base HEAD` — matching the presence gate's three-dot semantics; PRs behind base no longer flag base-side edits, and uncommitted worktree noise is excluded. |

Verified: 34/34 enforcement unit tests · 6/6 Playwright e2e · weakening CLI clean · boundary unit-checked.

danielzytohoc:
**Bugbot "Manual QA self-sign-off bypass" — fixed by redesign (`fa61e4345`).** The standalone `manual-qa.yml` workflow (and its bypassable label mechanism) is deleted. Manual QA is now an in-pipeline job (`[CHECK] Manual QA` in Tests — no extra pipeline, no extra triggers): a **tickbox in the PR template** requests it; sign-off is a **`/manual-qa passed` comment from a non-author** — comment authorship is API-verified identity, so the label-timing bypass class is gone entirely. The job reads the live body+comments, so tick → sign-off → re-run the job. Also in this amend: the es-cjs artifact now really runs `downlevel:types` (the earlier insertion silently no-op'd on an indentation mismatch — verify-types fails were correct behavior on a bad artifact), and the 9 Puppeteer jobs do `pnpm rebuild puppeteer` (the `-r rebuild` strip had skipped Chrome's postinstall download on warm store caches).

danielzytohoc:
**Round 3 fixes (`e8edb02`)** — root causes from the `fa61e43` run:

1. **Puppeteer "Could not find Chrome" (8 jobs)** — `pnpm rebuild puppeteer` at the workspace root is a **silent no-op** (exit 0, no output): non-recursive `rebuild` only sees the root project, and puppeteer is a devDep of `handsontable/`. Fixed to `pnpm --filter handsontable rebuild puppeteer` (verified locally: root form does nothing, filtered form downloads Chrome 148.0.7778.97 — the exact version the CI error names). Also added an `actions/cache` on `~/.cache/puppeteer` keyed by the lockfile, so the ~170 MB Chrome download happens once per lockfile state instead of in every browser job of every run; the filtered rebuild stays as the correctness backstop (it skips when the browser is already present).
2. **Verify types → publint** — the earlier TS 5.1 errors are gone (downlevel fix landed); publint then correctly flagged that `pkg.unpkg`/`pkg.jsdelivr` point at `dist/handsontable.full.min.js`, which the ES+CJS artifact alone can't provide. The job now downloads the UMD artifact too and merges `dist/` + `styles/` into `tmp/`, reconstructing the same publishable package a full build produces.
3. **Preview packages (latent, same cause)** — `pkg-pr-new` publishes `handsontable/tmp` as the preview package, so previews were shipping without `dist/` and CSS. Same two-artifact merge applied; both callers now gate `hot-prebuilt` on **both** build jobs and fall back to a full local build otherwise.

danielzytohoc:
**Round 4 (`d5ddffd`)** — three things:

1. **[TEST] Walkontable fix** — the engine artifact tar'd `dist/` and `test/dist/` but the Jasmine runner page (`test/SpecRunner.html`) is emitted one level above `test/dist`. Verified locally that the complete generated set is exactly `dist/`, `test/dist/`, `test/SpecRunner.html` (`git status --ignored` after a build); the artifact now carries all three.
2. **Docs Angular no longer fires on non-type core changes** — its path filter watched all of `handsontable/**`, so tooling/test-only PRs (like this one) triggered a full core+wrapper build just to type-check docs examples. Both the router filter and the workflow's own trunk push paths now watch only what shapes the public types: `src/**`, the two root `.d.ts` files, `package.json`, and `tsconfig*.json`.
3. **Job renames** — bracketed-tag names (`[CHECK]`/`[LINT]`/`[SCAN]`/`[TEST]`/`[BUILD]`/`[AUDIT]`) applied across the pipeline. For reusable-workflow fragments GitHub always displays "caller / inner", so callers carry the tagged name and inner jobs are minimal verbs (`check`, `lint`, `audit`, `run`, `deploy`, `publish`). Done now while no branch-protection required checks are pinned to the old names.

danielzytohoc:
**Round 5 (`39760c3`)** — PR pipeline restructured to the tiered DAG + composite-action reuse:

- **Tier 0 (gatekeepers, run first, everything depends on them):** `[CHECK] Scope` · `[CHECK] Test presence` · `[CHECK] Changelog`. If one fails, the PR pipeline stops.
- **Tier 1 (all no-build gates, needs tier 0):** lint, visual-tests lint, audit (warn), SonarCloud, CodeQL, FOSSA, unit, types, docs checks.
- **Build stage (needs ALL tier-1 gates — no building a flawed app):** UMD, ES+CJS, Walkontable engine, docs preview, docs Angular typecheck. Scope-skipped gates don't block (status functions); trunk runs are unaffected (PR-only gates skip cleanly).
- **Test stage and later:** unchanged.
- **New `.github/actions/setup-workspace` composite action** (pnpm + Node + store cache + install, optional Chrome provisioning) replaces the setup block previously pasted in 20 jobs; used 27×.
- **8 workflow_call-only fragments inlined as flat jobs** (exact names, no more "caller / job" slashes): changelog, linter, audit, visual-tests-linter, code-quality (→ 3 `[SCAN]` jobs), verify-emitted-types, pkg-pr-new, docs-checks (→ 2 jobs). Files deleted; secrets now referenced directly. Fragments with their own trunk/dispatch triggers (performance, docs-staging, docs-visual, docs-angular) keep their files.

Verified mechanically: 39 jobs, no duplicate ids, tier-0 jobs have no deps, all tier-1 jobs need the tier-0 trio, all build jobs need every no-build gate, every job has a dependency path to tier 0, checkout precedes the local composite everywhere, no leftover `inputs.*`/fragment references.

Known trade-offs (deliberate): a missing changelog now stops the whole pipeline (by design — re-run after adding the entry or `[skip changelog]`); builds wait for the slowest gate (CodeQL, ~3.5 min measured), moving the critical path from ~8 to ~11–12 min in exchange for zero wasted build/test minutes on gated-red PRs.

danielzytohoc:
**Round 6 (`4110181`)** — tier refinement, composites, and the Playwright theme split.

**Tiering (your split, adopted):**
- **Tier 0** — `[CHECK] Scope` · `[CHECK] Test presence` · `[CHECK] Changelog`. Everything depends on them.
- **Tier 1a — build-validity gates** (fast; gate the builds): `[LINT] JS/TS/CSS/D.TS` · `[TEST] Unit` · `[TEST] Types` (+ `[LINT] Docs` / `[TEST] Docs Plugins` gate the docs build).
- **Tier 1b — merge gates** (parallel; gate the *merge*, never a build): `[SCAN] SonarCloud/CodeQL/FOSSA` · `[AUDIT]` · `[LINT] Visual Tests` · `[CHECK] Manual QA`. CodeQL leaving the build path drops the critical path back ~3.5 min. Merge protection still enforces them (later).
- Builds now `needs` only the fast checks (verified: no build gates on a scan). `docs-angular` made consistent (build-tier on the fast checks, not the full scan set).

**Naming / structure:**
- `[SCAN] CodeQL` — dropped the single-language matrix that added the `(javascript-typescript)` suffix.
- `[LINT] Docs` (was `[CHECK] Docs JS/CSS/ASTRO`).
- `[CHECK] Manual QA` is now **independent** (no deps) — its ~4s verdict lands immediately in parallel.
- `[CHECK] Docs visual label` split out of the docs-visual fragment into an **early tier-1** job (cheap live-label read); `[TEST] Docs Visual` still needs the preview.

**Composites (`.github/actions/`):** `docs-angular-typecheck`, `docs-visual-run`, `performance-run` extracted and consumed by flat-named direct jobs. Their standalone workflow files are **left untouched** so develop/rc/release keep running byte-for-byte as today — the composites are the reuse target when we rework the higher pipelines. `setup-workspace` (from round 5) now backs the docs jobs too.
- **Deliberately deferred:** the Cloudflare deploy (`docs-preview`) stays a fragment call (`[BUILD] Docs Cloudflare Preview / deploy`, one remaining slash). It's 300 lines of incident-annotated deploy logic this PR can't exercise (no docs changes) — porting it blind is the exact silent-breakage risk we guard against. It gets its own dispatch-validated extraction during the docs/trunk rework.

**Playwright theme split (locally validated — 18 tests, 6 × 3 themes, green):** a `theme` project option (main/horizon/classic) parametrizes every spec automatically; fixtures load the matching stylesheet from a `?theme=` param. CI runs `[TEST] Playwright (theme: main/horizon/classic)` — same dependency depth as the Puppeteer e2e jobs, and now a required upstream of the visual budgeting gate. Local hooks + `test:e2e` run the default theme only (fast smoke); CI covers all themes.

Verified mechanically: 40 jobs, all `needs` resolve, tier-0 dependency-free, builds gate on CORE_GATE (not scans), scans/audit/visual-lint depended-on by nothing, every job reaches tier 0, release/`run-all` path intact (PR-only gates skip, unit/types run, builds proceed). Composites: all `run` steps carry `shell`, all `inputs.*` declared. `tests/` lint clean.

qunabu:
Task linked: [DEV-2058 Pipeline: develop trim + nightly full matrix (move full e2e off the develop push)](https://app.clickup.com/t/9015210959/DEV-2058)
Task linked: [DEV-2059 Pipeline: branch-protection cutover + flip enforcement warn→block](https://app.clickup.com/t/9015210959/DEV-2059)
Task linked: [DEV-2061 Test-generation evals: mutation score + prompt/skill regression harness](https://app.clickup.com/t/9015210959/DEV-2061)
Task linked: [DEV-2062 Burn down the ~100 tickets stuck in "testing" via meaningfulness-eval fan-out](https://app.clickup.com/t/9015210959/DEV-2062)

danielzytohoc:
**Bot findings triaged + fixed (`f17125672`).** All five open threads were legitimate; all fixed and resolved.

**cursor / bugbot:**
- **Wrapper PRs miss UMD artifact (High)** — real regression from the module split. Collapsing the two build jobs to `needs.build.result == 'success'` lost per-artifact truth: a wrappers-only PR builds ES+CJS only (UMD is scope-skipped), yet the module reports success, so `hot-prebuilt` went true and `emitted-types`/`preview packages` tried to download the missing `handsontable-build-umd`. **Fix:** `build.yml` now exposes per-job outputs (`umd-built`, `es-cjs-built`); `hot-prebuilt` for preview/emitted-types requires **both**, performance requires UMD — restoring the pre-split per-artifact gating (which then falls back to a full local build).
- **Visual suite ignores scope filter (Medium)** — same root cause. Visual was gated on `needs.build.result == 'success'`, so a `tests/**`-only PR (which builds UMD for Playwright) triggered the full multi-framework suite, whereas the monolith keyed on the **ES+CJS** build and skipped it. **Fix:** visual now gates on `es-cjs-built || test-visual`.
- **E2E cache omits theme styles (Low)** — correct. The green-run cache's `envHash` didn't include the theme CSS the fixtures now load, so a styles-only rebuild could stale-skip a spec. **Fix:** `handsontable.min.css` + `ht-theme-{main,horizon,classic}.min.css` are now part of `envHash`.

**CodeQL (client-side XSS in the two demo fixtures):** the `?theme=` param flowed into `document.write`. **Fix:** both fixtures now map the param through a fixed allowlist to a **literal** (`{horizon,classic}[param] || 'main'`) before use — the reflected value can only ever be a known theme, breaking the taint. Playwright still themes across all three (15/15 green).

danielzytohoc:
**Fixed (`fa86e24d5`): docs preview no longer coupled to the label read.** `preview` now depends on the docs *quality* gates (`lint`, `plugins`, `angular-types`) only — not `visual-label`, which is a live API label read for the optional visual suite (a routing flag, not a gate). A flaky label read can no longer skip the Cloudflare preview when the real gates pass. `visual` still depends on both `visual-label` and `preview`, so the label gating of the visual suite is unchanged.

qunabu:
Task linked: [DEV-2057 Pipeline: finish PR-pipeline unification — de-dupe builds, verify trunk/release shapes](https://app.clickup.com/t/9015210959/DEV-2057)

.github/workflows/test.yml
```diff
@@ -1,5 +1,16 @@
...
+# This file is the DAG only: it wires the module reusable workflows (one per
+# domain — Checks, Lint, Code Quality, Unit, Types, Build, E2E, Walkontable,
+# Integration, Emitted Types, Visual, Docs, Performance) via `needs` (tiers +
+# artifact ordering) and `if` (scope gating from the Checks router's outputs).
+# The jobs/steps live in the module files, so each is small and reviewable.
+#
+# The filename, the `workflow_call` `test-all` input, and the push triggers are
+# a stable contract: publish.yml (rc/stable/release) calls this workflow with
+# `test-all: true`, and develop/master push runs it with run-all — both get the
+# SAME jobs (now nested one level deeper) with only display names changed.
```
budnix, line 13: Heads-up for merge day: branch protection on `develop` still requires the old check names — "Check the changelog", "PNPM audit", "JS & CSS & TS & D.TS", "[BUILD] Handsontable: UMD", the six "[TEST] Handsontable: UMD…" legs, and the three wrapper tests. This PR deletes or renames all of them, so after merge every new PR will show those required checks as "Expected — waiting for status" forever, and nothing can be merged until an admin rewrites the rule. Worth updating the branch-protection contexts in the same moment this lands.

.github/workflows/integration.yml
```diff
@@ -0,0 +1,172 @@
...
+      run-pkg-preview: { type: boolean, default: false }
+      hot-prebuilt: { type: boolean, default: false }
+      wrapper-matrix:
+        description: 'JSON array of wrappers to test (precomputed by the Checks scope job)'
+        type: string
+        default: '[]'
+
+permissions:
+  contents: read
+  pull-requests: write
```
budnix: This permission breaks the release path. `publish.yml` caps its token to `id-token, contents, actions`, and a called workflow can't ask for more than its caller has — so when `publish.yml` → `test.yml` → `integration.yml` runs for an RC, GitHub rejects the call with "requesting 'pull-requests: write', but is only allowed 'pull-requests: none'", and no RC can be published. Either grant `pull-requests: write` in `publish.yml`, or scope this down to the `preview-packages` job — it's the only one that posts a comment, and it's PR-only anyway.

.github/workflows/unit.yml
```diff
@@ -0,0 +1,50 @@
...
+          fetch-depth: 0
+      - uses: ./.github/actions/setup-workspace
+      - name: Test (with coverage)
+        run: |
+          cd handsontable
+          # Instrument ALL of src (not just files a test imports) so a changed
+          # file no test exercises shows as uncovered in the lcov — the coverage
+          # floor must see it, not silently skip it.
+          npm run test:unit -- --coverage \
+            --collectCoverageFrom='src/**/*.{js,ts}' \
```
budnix, line 31: These two flags never reach Jest. `npm run test:unit` goes through `scripts/run.mjs`, and the `test:unit.jest` task in `tasks.json` only forwards `--testPathPattern`, `--watch`, and `--coverage` — everything else is filtered out. So the "Instrument ALL of src" comment above doesn't hold: a new source file with no test is simply missing from the lcov, and the coverage floor prints "nothing to enforce" instead of 0%. The fix needs two parts: add both flags to the task's `passthroughFilter`, and turn the bare `'/3rdparty/'` token into a second `--coveragePathIgnorePatterns=` flag, because a bare token gets dropped by the filter too.

.github/workflows/checks.yml
```diff
@@ -0,0 +1,210 @@
...
+    runs-on: ubuntu-latest
+    # Pillar 1: source changed ⇒ a matching test changed (warn mode — reports,
+    # does not block yet). PR-only; needs git history, no build.
+    if: github.event_name == 'pull_request'
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # https://github.com/actions/checkout/releases/tag/v6.0.2
+        with:
+          fetch-depth: 0
+      - name: Evaluate test-presence gate (warn)
+        run: node .github/scripts/test-presence-gate.mjs | tee -a "$GITHUB_STEP_SUMMARY"
```
budnix: The pipe eats the gate's exit code. The default Actions shell here is `bash -e {0}` with no pipefail, so the step exits with `tee`'s status (always 0) even when the gate exits 1. Harmless in warn mode, but on the day `GATE_MODE` flips to `block`, the gate still won't fail the job — same story for the weakening gate below.
```suggestion
        run: set -o pipefail; node .github/scripts/test-presence-gate.mjs | tee -a "$GITHUB_STEP_SUMMARY"
```

.github/scripts/__tests__/presence-gate.test.mjs
```diff
@@ -0,0 +1,183 @@
+import { test } from 'node:test';
```
budnix, line 1: Did these end up wired into CI anywhere? I couldn't find anything that runs `node --test` — no workflow, no npm script, no lefthook step. If the classifier or the diff parser regresses later, the gates silently stop enforcing anything, and these tests never get the chance to catch it. A small `node --test` job in checks.yml (or a root npm script a workflow calls) would close the loop.

.github/workflows/test.yml
```diff
@@ -28,971 +39,154 @@ concurrency:
...
+      # have been built for the prebuilt path (else it full-builds locally).
+      hot-prebuilt: ${{ needs.build.outputs.umd-built == 'true' && needs.build.outputs.es-cjs-built == 'true' }}
+      # Wrapper list precomputed by the Checks scope job (no setup-job stage).
+      wrapper-matrix: ${{ needs.checks.outputs.wrapper-matrix }}
+    secrets: inherit
+
+  emitted-types:
+    name: Emitted Types
+    needs: [ checks, build ]
+    if: ${{ !failure() && !cancelled() && github.event_name == 'pull_request' && (needs.checks.outputs.run-all == 'true' || needs.checks.outputs.verify-types == 'true') }}
```
budnix: The `github.event_name == 'pull_request'` part drops two runs the old `verify-emitted-types.yml` had: pushes to develop/master, and the release path — `publish.yml` calls this with `test-all: true`, but the event is a push, so the module is skipped. A type break that lands on trunk directly, or shows up only at release time, now ships to npm unchecked — and the header comment's promise that publish gets "the SAME jobs" doesn't hold for this one. The build artifacts it needs exist on run-all runs, so dropping the event check should be enough.
```suggestion
    if: ${{ !failure() && !cancelled() && (needs.checks.outputs.run-all == 'true' || needs.checks.outputs.verify-types == 'true') }}
```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,34 @@
...
+
+jobs:
+  core:
+    name: core
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # https://github.com/actions/checkout/releases/tag/v6.0.2
+      - uses: ./.github/actions/setup-workspace
+      - name: JavaScript lint
+        run: npm run lint
```
budnix, line 22: The new `tests/` package never gets linted in CI: `npm run lint` fans out through `scripts/translate-to-native-npm.mjs`, and its hardcoded package list doesn't include it. That means the Playwright determinism bans (`waitForTimeout`, `sleep`, `networkidle`, `.skip`) are only enforced by the local hooks, which `--no-verify` skips — while `lint-files.mjs` promises that CI's full lint is the authority. Worth adding `tests/` to the fan-out list, or a lint step here.

.github/workflows/checks.yml
```diff
@@ -0,0 +1,210 @@
...
+              - *hot-shared
+              - 'wrappers/react-wrapper/**'
+            test-vue3:
+              - *hot-shared
+              - 'wrappers/vue3/**'
+            test-visual:
+              - './examples/next/visual-tests/**'
+              - './visual-tests/**'
+            test-performance:
+              - './handsontable/src/**'
```
budnix, line 143: This narrows the old `handsontable/**` trigger: a PR that bumps a dependency in `handsontable/package.json`, or edits `babel.config.js` / `hot.config.js`, no longer runs the perf suite. The sharp edge is what happens next — the develop push after merge runs in golden mode, so a bundle-level slowdown gets recorded as the new baseline instead of being flagged. Was the narrowing to `src/**` deliberate, or is it worth including the build-config paths here (they're already grouped as `config-files` above)?

.github/workflows/performance-tests.yml
```diff
@@ -1,5 +1,10 @@
 name: Performance tests
 
+# On PRs this runs as a fragment of the Tests pipeline (workflow_call from
```
budnix: This comment says test.yml calls this file via workflow_call, but test.yml actually calls `performance.yml` (the wrapper around the performance-run composite) — nothing calls this trigger, and the `hot-prebuilt` input is dead. The risk is drift: someone fixes the PR flow here, as the comment directs, and nothing changes on PRs — then the PR numbers and the golden baseline come from different steps. Would be good to point the comment at the real path, or fold the push run onto the same composite.

scripts/pre-push.mjs
```diff
@@ -0,0 +1,186 @@
...
+
+  if (skipped.length > 0) {
+    console.log(`pre-push: ${skipped.length} spec(s) already proven green (unchanged spec + environment) — skipped.`);
+  }
+
+  if (toRun.length > 0) {
+    console.log(`pre-push: running ${toRun.length} changed Playwright spec(s)…`);
+    // Local smoke = the default theme only (fast); CI runs the full theme
+    // matrix (main/horizon/classic).
+    const pw = spawnSync('npx', ['playwright', 'test', '--project=e2e-main', ...toRun], {
```
budnix, line 147: On Windows this spawn fails with ENOENT — `npx` is a `.cmd` shim and there's no `shell: true` — so `pw.status` is null and every push touching a spec exits 1, even when the tests would pass. The ESLint step above has the opposite problem: `status === 1` is never true there, so lint failures pass silently. A hook that fails closed on every push is what trains people to reach for `--no-verify`, which turns the whole local layer off. `shell: process.platform === 'win32'` on these spawns (and the `npm` one below, plus the `cat` call in the Claude hooks' readStdin) would cover it.

.github/workflows/code-quality.yml
```diff
@@ -17,55 +15,59 @@ jobs:
...
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # https://github.com/actions/checkout/releases/tag/v6.0.2
+      - uses: fossas/fossa-action@ff70fe9fe17cbd2040648f1c45e8ec4e4884dcf3 # https://github.com/fossas/fossa-action/releases/tag/v1.9.0
+        with:
+          api-key: ${{secrets.FOSSA_API_KEY}}
+
+  audit:
+    name: audit (warn)
+    runs-on: ubuntu-latest
+    # Non-blocking by design (~380 known vulns to burn down before this can gate).
+    continue-on-error: true
```
budnix, line 65: The burn-down reasoning makes sense, but with the old `audit.yml` deleted this leaves no audit signal that can ever block — and the nightly baseline workflow has no audit job either. So a new fixable prod vulnerability merges green and ships with no red anywhere in CI. Even while the ~380 backlog burns down, an audit job in `code-quality-baseline.yml` would keep a trunk signal alive.

budnix (commented):
Notes from a deep pass over the pipeline split, with a focus on the release path and the new enforcement gates. Three more findings (deleted-test loophole in the presence gate, the unused `docs-angular` scope output, and the changelog re-run on edit) are already flagged by earlier bot comments, so I didn't repeat them here.

.github/workflows/test.yml
danielzytohoc: Acknowledged — this is real and deliberately **not** solved by a code change here. Plan (tracked for the merge-day cutover, DEV-2059): rather than rewriting the required-check list to the new `Module / job` names (which re-incurs this exact problem on the next rename), we'll add a single **rollup check** (`tests-passed`, `needs:` every required module) and point branch protection at that one stable name. The admin swap happens in the same moment this merges. Until then nothing changes for existing PRs (the new names run alongside; the old rule is only broken once the old workflows stop reporting, i.e. at merge).

.github/workflows/integration.yml
danielzytohoc: Fixed in `7a76c98d9` — and this one went deeper than both of our first takes. Your diagnosis was right, but the suggested fix (scope `pull-requests: write` down to the `preview-packages` job) turned out to be **insufficient too**: GitHub validates every job-level `permissions:` request in a called workflow's graph against the caller's ceiling **statically at run initialization** — `if:` conditions on runtime-skipped jobs do not exempt them (actions/runner#4151). We proved it empirically with a scratch-branch probe that mimics `publish.yml`'s exact ceiling calling `test.yml`: the job-scoped fix still died as `startup_failure`.

The probe also surfaced **three more violations nobody had caught** — `docs.yml angular-types` (`checks: write`), `code-quality.yml codeql` (`security-events: write`, `packages: read`), and `performance.yml` (`contents: write`) — any one of which would have failed every RC at startup.

Final fix: `publish.yml`'s `first-rc-test`/`rc-test` jobs now carry a job-level ceiling = the union of all nested requests, deliberately **without `id-token`** (nothing in the test graph publishes, so the test jobs also stop inheriting OIDC — a net tightening). The job-level scoping in `integration.yml` stays for least-privilege on the PR path. Probe re-verified green with the exact final shape. Full experiment in the summary comment on the PR.

.github/workflows/unit.yml
danielzytohoc: Fixed in `7a76c98d9`, both parts as you specified: `--collectCoverageFrom` + `--coveragePathIgnorePatterns` added to the `test:unit.jest` `passthroughFilter`, and the bare `'/3rdparty/'` token replaced with a second `--coveragePathIgnorePatterns=` flag (Jest array-appends it).

Proven end-to-end, not just by inspection: running the exact `unit.yml` command scoped to **one** test file produced an lcov with **707 src files** — including 321 plugin files that test never imports — and **0** `3rdparty` / **0** `__tests__` entries. A new source file with no test now shows up at 0% instead of being invisible to the floor.

.github/workflows/checks.yml
danielzytohoc: Fixed in `7a76c98d9` — your suggestion applied to **both** gate steps (presence + weakening). Verified the semantics locally under `bash -e`: `false | tee` exits 0 without `pipefail`, 1 with it. Latent today (warn mode exits 0 by design) but this had to land before the DEV-2059 warn→block flip, exactly as you said.

.github/scripts/__tests__/presence-gate.test.mjs
danielzytohoc: Correct — they existed but nothing executed them. Fixed in `7a76c98d9`: a root `test:tooling` script (`node --test` over `.github/scripts/__tests__`, `scripts/__tests__`, `scripts/claude/__tests__`, `evals/__tests__`) + a `tooling-tests` job in `checks.yml` (PR-only, checkout + setup-node only — a transitive import scan confirms the suites use zero non-builtin deps, so no install step). 65/65 passing. A classifier/diff-parser regression now fails CI instead of silently disarming the gates.

.github/workflows/test.yml
danielzytohoc: Fixed in `7a76c98d9` with your suggested shape: the `github.event_name == 'pull_request'` clause is gone; the job gates on scope only (`run-all || verify-types`), with a comment banning event-gating here. That restores both dropped runs — develop/master push and the `publish.yml` release call (where the event is `workflow_call`, so the old condition could never be true). Also checked `emitted-types.yml` itself references no PR-only context, and on run-all both build artifacts it consumes exist.

.github/workflows/lint.yml
danielzytohoc: Fixed in `7a76c98d9`: root `lint` now appends `npm run lint --prefix tests`, so the Playwright determinism bans (`waitForTimeout`, `sleep`, `networkidle`, `.skip`/`.only`) are enforced by CI's `Lint / core` job — the authoritative layer — not just by the bypassable local hooks. Verified `npm run lint --prefix tests` runs the package's eslint and exits 0 on the current tree.

.github/workflows/checks.yml
danielzytohoc: Not deliberate — an over-narrowing, and your sharp edge (config-driven slowdown skipping the PR run, then being **recorded as the golden baseline** on the develop push) is exactly right. Fixed in `7a76c98d9`: the `test-performance` filter now includes the existing `*config-files` anchor (deps in `package.json`, `babel.config.js`, `hot.config.js`, workflows), matching the build triggers. A config-only PR that can shift bundle perf now runs the suite pre-merge.

.github/workflows/performance-tests.yml
danielzytohoc: Fixed in `7a76c98d9`: header rewritten to the truth — this file's live job is the **develop-push golden run** only; the PR path goes through `performance.yml` (the `workflow_call` wrapper around the `performance-run` composite), and the `workflow_call` trigger here is explicitly marked unused, kept only until the develop-pipeline rework (DEV-2057) folds the push run onto the shared composite. The comment now ends with "do not wire new PR behavior here" to prevent the drift you predicted.

scripts/pre-push.mjs
danielzytohoc: Fixed in `7a76c98d9`: `shell: process.platform === 'win32'` on every `npx`/`npm` `spawnSync` across `pre-push.mjs`, `lint-files.mjs`, `stop.mjs`, `post-tool-use.mjs` — plus one more instance our review found beyond your list: `prepare-dev-env.mjs` (the `pnpm install` `prepare` hook), where the unguarded `npx lefthook install` meant Windows devs silently got **no git hooks at all**. Both Claude hooks' `cat`-based stdin reads are replaced with `readFileSync(0)`.

On the ESLint step: the ENOENT→`null`→silent-pass path you flagged is closed by `shell: WIN` (the status is now real). The `status === 1 ? 1 : 0` mapping itself is kept deliberately — eslint exit 2 is a config/parse gap, which must warn-and-skip rather than block (documented in the file header); with the spawn fixed, 1 = genuine lint errors = block, exactly as intended.

.github/workflows/code-quality.yml
danielzytohoc: Agreed and fixed in `7a76c98d9`: `code-quality-baseline.yml` (the 03:00 UTC nightly on the default branch) now has an `audit` job — same `pnpm audit --prod --ignore-unfixable --ignore-registry-errors` invocation, `continue-on-error` while the ~380-vuln backlog burns down. So the trunk keeps a visible daily audit signal (the gap you called out: previously nothing anywhere could ever go red) without making it a gate prematurely. Flipping it to blocking stays tied to the backlog burn-down.

danielzytohoc:
**Colleague-review round (`7a76c98d9`)** — all 11 of @budnix's findings triaged, verified against the code, and fixed (one deferred by design). Every finding was legitimate — zero false positives. Thread-by-thread commentary is on each conversation; the one that deserves the headline:

## The permissions fix needed two attempts — the first (both ours and the suggested one) was provably wrong

The suggested remedy for the release-path break — scope `pull-requests: write` down to the PR-only `preview-packages` job — **does not work**: GitHub validates every job-level `permissions:` request in a called workflow's graph against the caller's ceiling **statically at run initialization**. An `if:` that skips the job at runtime does not exempt it; one excess request anywhere in the graph fails the entire run as `startup_failure` (actions/runner#4151).

We proved this empirically with a 4-round scratch-branch probe (a workflow carrying `publish.yml`'s exact permission ceiling, calling `test.yml` the way an RC run does):

| Round | Caller ceiling | Verdict |
|---|---|---|
| 1 | publish's exact ceiling + the job-scoped fix | ❌ `startup_failure` |
| 2 | + union of nested requests (workflow-level) | ✅ initialized |
| 3 | union minus `contents: write` / `id-token` | ❌ `startup_failure` |
| 4 | exact final set, **no `id-token`** | ✅ initialized |

The probe also surfaced **three more release-breakers no reviewer or bot had caught** — nested job-level requests from `docs.yml angular-types` (`checks: write`), `code-quality.yml codeql` (`security-events: write`, `packages: read`), and `performance.yml` (`contents: write`). Each alone would have failed every RC at startup. All three were created by this PR folding previously-standalone workflows into the called graph — invisible to PR CI, since a `pull_request` run never exercises the `workflow_call` ceiling.

**Final fix:** `publish.yml`'s `first-rc-test` / `rc-test` jobs now grant a job-level ceiling = the union of every nested request (`contents: write, actions: read, pull-requests: write, checks: write, security-events: write, packages: read`) — deliberately **without `id-token`**, so the entire called test graph also stops inheriting OIDC it never needed (a net tightening vs. today). The job-level scoping inside `integration.yml` stays for least-privilege on the PR path. Round 4 of the probe verifies this exact shape.

**Standing invariant (recorded in the repo docs):** any new `permissions:` scope added to a Tests module must extend that ceiling, or every RC fails at startup — and PR CI cannot see it. The scratch-branch probe is the 60-second pre-merge check until the pipeline split (DEV-2057) moves PR-only elevated-permission jobs out of the shared graph and shrinks the ceiling back.

## The rest, in one line each (details on the threads)
- **emitted-types** — event gate removed; runs on trunk push + release again (parity restored).
- **coverage flags** — forwarded through `run.mjs`; proven by a 707-file lcov from a single-test run.
- **pipefail** — both gate pipes fixed before the warn→block flip can be betrayed by `tee`.
- **tooling tests** — `test:tooling` + a `tooling-tests` CI job; the enforcement machinery now tests itself (65/65).
- **`tests/` lint** — in the CI lint fan-out; determinism bans are CI-enforced now.
- **perf scope** — `config-files` back in the filter; config-driven regressions can no longer sneak into the golden baseline.
- **Windows hooks** — `shell:` guard on every `npx`/`npm` spawn (+ one extra instance found in `prepare-dev-env.mjs`), `readFileSync(0)` stdin.
- **audit** — nightly non-blocking trunk audit restored in `code-quality-baseline.yml`.
- **perf comment drift** — corrected.
- **branch protection** — deferred to the merge-day cutover (DEV-2059): one stable `tests-passed` rollup check instead of chasing renamed job names.

Great review, @budnix — it aimed exactly where a green PR run is structurally blind (the release path, the gate mechanics), and everything you flagged was real.

budnix (dismissed):

.github/workflows/checks.yml
```diff
@@ -0,0 +1,231 @@
...
+          filters: |
+            config-files: &config-files
+              - './handsontable/.config/**'
+              - './handsontable/package.json'
+              - './package-lock.json'
+              - './handsontable/hot.config.js'
+              - './handsontable/babel.config.js'
+              - './.github/workflows/**'
+              - './visual-tests/playwright.config.ts'
+              - './visual-tests/playwright-cross-browser.config.ts'
```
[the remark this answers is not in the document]
danielzytohoc: All three confirmed against the tree and fixed in `9e9ef82e0`:

- `./package-lock.json` → `./pnpm-lock.yaml` (`package-lock.json` doesn't exist in this pnpm repo — the entry was dead, so dependency bumps never matched `hot-shared`).
- `./handsontable/hot.config.js` → `./hot.config.js` (the file lives at the repo root; corroborated by `tasks.json` loading it via `env-cmd -f ../hot.config.js`).
- Added `./handsontable/rspack.config.js` (the bundler entry, invoked from the `handsontable/` cwd — was covered by neither `.config/**` nor any literal).

These two dead entries were inherited verbatim from the old `test.yml` filter (pre-existing on `develop`), but this PR made the anchor more load-bearing (it now also gates the perf compare), so they mattered more here.

Also swept **every** remaining literal in the anchors for existence: all valid, except the `hot-definition-files` pair (`handsontable/base.d.ts`, `handsontable.d.ts`) which are build-generated and untracked on `develop` too — never-matching but harmless (types changes are still caught via `src/**` + `test/types/**`), left as-is to keep this diff scoped.

.github/scripts/test-weakening-gate.mjs
```diff
@@ -0,0 +1,94 @@
...
+const rows = nameStatus.split('\n').filter(Boolean).map((line) => {
+  const [status, ...rest] = line.split('\t');
+
+  return { status: status.trim()[0], path: rest.join('\t').trim() };
+});
+
+// Real source only — the presence-gate classifier excludes spec/unit/test files,
+// so a spec-only change never counts as "source changed".
+const sourceChanged = rows.some(r => isSource({ path: r.path }));
+const modifiedSpecs = rows.filter(r => r.status === 'M' && SPEC_RE.test(r.path));
```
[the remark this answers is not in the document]
danielzytohoc: Confirmed and fixed in `a06b65964` — the loophole was real: the gate filtered on status `M` only, so a brand-new spec born with `.skip`/`.only` satisfied the presence gate while the anti-gaming signal stayed quiet.

The fix goes one step beyond the report:

- **Added specs** now diff against empty content — the lib's counters already handle `before: ''` correctly (skip markers become a `0 → N` finding; assertions can't "drop" from empty, so a healthy new spec stays quiet).
- **Renamed specs** (status `R`) had the same hole — and the naive fix (treat as added) would *false-positive* every pre-existing skip in a renamed file as newly added. The new `parseNameStatus` (extracted to the tested lib) keeps the old path for the base-side read, so renames diff old→new correctly.

Verified end-to-end against the committed state, not just by unit tests: a temp commit adding a `.skip` spec → gate reports `skip/focus markers 0 → 1`; a temp rename of a spec with a pre-existing skip → gate stays quiet. Plus 5 new lib tests (`node --test`, wired into CI via the `tooling-tests` job). `pre-push` runs the same script, so the local layer inherits the fix.

Defense-in-depth note: the `no-skipped-test`/`no-focused-test` ESLint rules also block these at authoring/CI-lint time — this gate is the layer that catches what slips past lint (e.g. a `--no-verify` push where lint scope missed the file), and now it actually covers the new-file path it was blind to.

budnix (dismissed):

.changelogs/README.md
```diff
@@ -15,7 +15,7 @@ Changelog entries are expected only for source-code changes to the library core
 [skip changelog]
 ```
 
-...and push a new commit to restart the check (`git commit --allow-empty` to create an empty commit).
+...and re-run the failed **Changelog** check from the PR's checks tab (or push any new commit — `git commit --allow-empty` works). The check reads the PR body at run time; editing the description alone does not re-trigger it.
```
demtario, line 18: From what I experienced on _develop_, the check actually **does** rerun after editing the description. So either this claim is invalid, or the behaviour was changed and actually empty commit is needed. 

Either way, I think this check should retrigger on description change, without commit or repush

.claude/skills/handsontable-plugin-dev/SKILL.md
```diff
@@ -139,7 +139,7 @@ If your plugin provides UI elements (buttons, inputs, navigation bars), you must
 
 ## Testing Requirements
 
-- E2E tests (`__tests__/*.spec.js`): all `it()` callbacks must be `async`.
+- **New E2E tests are Playwright** in `tests/e2e/` (skill `handsontable-playwright-e2e`) — for a plugin's UI, interaction, rendering. Do not add new legacy `*.spec.js`; the presence gate blocks them. Editing an existing `*.spec.js` (async `it()` callbacks) is fine.
```
demtario, line 142: How long a "new" tests will be new? Maybe better approach would be to mark the old one as legacy (as is already mentioned), and treat the "new" as just "the tests" without the word "new". WDYT?

The "new tests" are mentioned few times thoughout the diff.

.github/workflows/code-quality-baseline.yml
```diff
@@ -0,0 +1,91 @@
...
+#     baseline PRs compare against goes stale.
+#   - GitHub's code-scanning (CodeQL) dashboard for the default branch only
+#     refreshes from an analysis of that branch — PR-only CodeQL analyzes diffs,
+#     not the trunk HEAD.
+# Runs on the default branch (what CodeQL's dashboard tracks and what PRs base
+# their new-code comparison on). workflow_dispatch allows an on-demand refresh.
+on:
+  schedule:
+    # 03:00 UTC daily — off-peak; the exact minute is arbitrary.
+    - cron: '0 3 * * *'
```
demtario: Maybe lets skip weekends? I suppose we almost never merge anything on saturday/sunday.

demtario (changes requested):

.claude/skills/handsontable-plugin-dev/SKILL.md
danielzytohoc: PR-template evidence label already says just "E2E tests added/modified (Playwright…)" — no "new" — which is the right call there, because in an evidence checklist "new" is redundant. 
 **"new" here reads as newly-authored, not recently-added — a spec written next year is still a "new" spec that must be Playwright, so the wording doesn't actually age**

.changelogs/README.md
danielzytohoc: The changelog gate reads only two inputs: a .changelogs/*.json file in the diff and the [skip changelog] string in the description. The file path already re-triggers on push (synchronize), so an edited trigger would buy us only the auto-clearing of the [skip changelog] escape — a rare non-source path a single "Re-run failed job" click already covers — while firing on every title edit (which the gate ignores), so it's not worth wiring in.

Description change alone doesn't matter here - it's an empty gate we trigger multiple times. Only .changelogs/*.json change matters and that will actually re-run the check.

.github/workflows/code-quality-baseline.yml
danielzytohoc: Done in `7a2a6695b` — schedule is now `0 3 * * 1-5` (Mon–Fri). Nothing merges to trunk on Sat/Sun so the baseline can't go stale then; Monday's run refreshes it before the week's first PRs. Thanks!

demtario (dismissed):

budnix (dismissed):

demtario (dismissed):

adrianspdev (dismissed):

.github/workflows/checks.yml
```diff
@@ -0,0 +1,236 @@
...
+    # PR-only. Deliberately NOT re-run on `edited`: re-run the failed job after
+    # adding an entry or `[skip changelog]`.
+    if: github.event_name == 'pull_request'
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # https://github.com/actions/checkout/releases/tag/v6.0.2
+      - uses: ./.github/actions/setup-workspace
+      - name: Assert that a new changelog was added, or that the check is skipped
+        env:
+          TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: node .github/scripts/check-changelog
```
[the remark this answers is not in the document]
danielzytohoc: Fixed in `581af3aec`. `check-changelog.js` now reads the **live** PR body via `octokit.rest.pulls.get` instead of `github.context.payload.pull_request.body` (which is frozen at the triggering event and stays stale across "Re-run failed jobs"). So adding `[skip changelog]` to the description and re-running the failed Changelog check now clears it — no empty commit and no `edited` trigger re-running the whole pipeline. Falls back to the payload body if the fetch fails. This makes the behavior the docs (checks.yml comment + .changelogs/README.md) already described actually true.

budnix (approved):

PR #13073, opened 2026-07-14
fix: preserve source columns during source data validation

ShiroKSH:
### Context

The PR fixes source-data validation for mapped columns. The validation scan iterates physical columns, but `DataSource#getAtCell()` expects a visual column and `DataSource#setAtCell()` expects the source property or physical source index. As a result, moving a column could make Handsontable validate the neighboring source value, while invalid object data was cleared under a numeric property instead of the configured property.

The change keeps physical coordinates for cell meta and warnings, uses visual coordinates for reads, and uses source properties for writes. This preserves the documented `allowInvalid: false` behavior without changing the public API.

### How has this been tested?

- `npm run test:unit --prefix handsontable -- --testPathPattern=sourceDataValidator.unit` (19 specs)
- `npm run test:e2e --prefix handsontable -- --testPathPattern=sourceDataValidator` (24 specs)
- `npm run test:types --prefix handsontable`
- `npm run build --prefix handsontable`
- `npm run eslint --prefix handsontable`
- `npm run stylelint --prefix handsontable`

### Types of changes

### Related issue(s):

N/A - discovered through code audit and regression reproduction.

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Touches load/update validation and how invalid source values are written back across object-mapped and reordered columns; behavior change is intentional but affects all grids using built-in or custom source validators with `allowInvalid: false`.
> 
> **Overview**
> Fixes **source-data validation** so reads and writes use the correct column identity when columns are mapped to object properties or reordered.
> 
> The scan used to iterate **physical** columns but call `DataSource` with mixed coordinates: values were read as if columns were visual indices, while invalid values were cleared with physical indices. That broke `allowInvalid: false` for object/`columns[].data` (including nested paths) and could validate or null the wrong cell after **manual column move**.
> 
> `sourceDataValidator.ts` now builds a per-column mapping (physical → visual → `colToProp`) and uses **`getAtPhysicalCell`** / **`modifyRowData`** for reads and **`setAtCell(row, sourceColumn, null)`** for blanking, while keeping physical indices for meta and warnings. Unit and e2e tests cover object/nested dates, moved columns, and both batched and per-cell validation paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 39a025f69cd8e1c016aa62703a177b3f61b2553d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ShiroKSH:
Hi maintainers, this PR has been open for over two weeks. When you have a moment, I would appreciate a review. Thank you.

.changelogs/13073.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "public",
```
demtario: **Required:** this should be `"private"`.

Bugbot is right here. `.changelogs/README.md` frames `issuesOrigin` as answering one question: is the number in `issueOrPR` a public GitHub **issue**? It is not. `13073` is this pull request, and the description says there is no related public issue. `private` is the documented default and the one that pairs with a PR number, and it is what selects the `/pull/<n>` link in the generated `CHANGELOG.md`.

```diff
-  "issuesOrigin": "public",
+  "issuesOrigin": "private",
```

Everything else in the entry is correct: filename matches the PR number, `type: "fixed"`, `framework: "none"`, `breaking: false`.

handsontable/src/dataMap/sourceDataValidator.ts
```diff
@@ -180,9 +189,10 @@ function validatePerCell(
         continue;
       }
 
-      const value = dataSource.getAtCell(row, col);
+      const sourceColumn = hotInstance.colToProp(visualColumn);
```
demtario: **Required:** please hoist this out of the loop.

`colToProp()` goes through `hot.toPhysicalColumn()`, so this adds an IndexMapper round trip per validated cell inside the `row x col` double loop. `validatePerCell` is the path taken whenever a validator is row-dependent, a `cell` array exists, or `setCellMeta()` has ever been called, so it walks the whole dataset on every init, `loadData`, or `updateData`. The last three commits on this file were all load-time performance work. `validateBatched` already precomputes `sourceColumn` at collect time; this path should do the same.

To be fair about the size of it: I downloaded the run's `performance-tests-output` artifact and there is **no measured regression**. `source-data-validator-load` moved +1.7%, the smallest delta in the report. So this is prudence, not a red flag.

Building the arrays up front also lets you lift the two existing per-cell `getVisualFromPhysicalIndex()` calls out, which is the bigger win:

```ts
const visualColumns: (number | null)[] = [];
const sourceColumns: (string | number | null)[] = [];

for (let col = 0; col < colSourceCount; col += 1) {
  const visualColumn = columnIndexMapper.getVisualFromPhysicalIndex(col);

  visualColumns[col] = visualColumn;
  sourceColumns[col] = visualColumn === null ? null : hotInstance.colToProp(visualColumn);
}

for (let row = 0; row < rowSourceCount; row += 1) {
  const visualRow = rowIndexMapper.getVisualFromPhysicalIndex(row);

  if (visualRow === null) {
    continue;
  }

  for (let col = 0; col < colSourceCount; col += 1) {
    const visualColumn = visualColumns[col];

    if (visualColumn === null) {
      continue;
    }
    // ... cellMeta, then getAtCell(row, visualColumn) and sourceColumns[col]
  }
}
```

Row-major iteration stays intact, which is what the comment on `validateBatched` cares about for warning order.

handsontable/src/dataMap/sourceDataValidator.ts
```diff
@@ -10,6 +10,8 @@ type InvalidItems = Map<string, Array<{ row: number; col: number; value: unknown
 
 type ColumnValidator = {
   physicalColumn: number;
+  visualColumn: number;
+  sourceColumn: string | number;
```
demtario, line 14: Note only, **not** something to fix in this PR.

`colToProp()` can also return a **function**, because `columns[].data` accepts an accessor (handled in `dataMap.ts` at the `typeof prop === 'function'` branches, and in `DataSource#getAtPhysicalCell`). In that case `setAtCell()` takes its `!Number.isInteger(column)` branch and calls `setProperty(dataRow, String(column), value)`, writing the function's source text as an object key.

That is broken on `develop` too (it wrote key `"0"` instead), and it is really a `DataSource` limitation rather than anything you introduced. Your change makes the read side correct for that config either way. Flagging it because this annotation now claims a slightly narrower domain than the runtime holds; we will track it separately.

handsontable/src/__tests__/settings/sourceDataValidator.spec.js
```diff
@@ -104,6 +104,33 @@ describe('settings', () => {
...
+        ],
+        manualColumnMove: true,
+        allowInvalid: false,
+      });
+
+      await getPlugin('manualColumnMove').moveColumn(0, 1);
+      await updateData([['not-a-date', '2024-02-02']]);
+
+      expect(getSourceData()).toEqual([[null, '2024-02-02']]);
+    });
```
demtario, line 142: Optional, only if you feel like it: one more case here for a dotted property, for example

```js
it('should nullify a nested mapped property when allowInvalid is false', async() => {
  handsontable({
    data: [{ meta: { date: 'not-a-date' } }],
    columns: [{ data: 'meta.date', type: 'date', dateFormat: 'YYYY-MM-DD' }],
    allowInvalid: false,
  });

  expect(getSourceData()).toEqual([{ meta: { date: null } }]);
});
```

`setAtCell()` calling `setProperty()` on a nested prop is the write path where getting `sourceColumn` right matters most, and it is the one shape your two cases do not cover (you have flat object data, and array data plus a move).

demtario (commented):
Thanks for this, @ShiroKSH, and sorry for the long wait.

I reviewed it in an isolated worktree and reverted **only** `handsontable/src/dataMap/sourceDataValidator.ts` to `develop`, keeping your tests. Both of your new cases fail, with exactly the symptoms you describe:

- `should nullify the mapped object property when allowInvalid is false` fails with `Expected $[0] not to have properties`. On `develop` the invalid value stays in `date` and a stray `null` is written under the numeric key `"0"`.
- `should validate the source column after it has been moved` fails with `Expected $[0][0] = 'not-a-date' to equal null. Expected $[0][1] = null to equal '2024-02-02'.` On `develop` the neighbouring column is validated and blanked.

With the fix applied: 19/19 unit specs and the full E2E suite (9607 specs) pass. Your read of the mismatch is correct. `DataSource#getAtCell()` runs its argument through `colToProp()` and so wants a visual index, while `setAtCell()` wants the source property. Confirmed, and worth fixing.

**Your test placement is exactly right**, which is worth calling out explicitly. `handsontable/.ai/TESTING.md` requires a bug fix to add a case to the closest existing spec and for that case to fail without the fix, and it forbids new `*.spec.js` files. You did both, and both cases genuinely fail on `develop`.

### The red checks are not yours

`Performance / run` and `Integration / preview packages` fail while *posting* their PR comment and pushing their report, with `Resource not accessible by integration` and `Permission to handsontable/handsontable.git denied to github-actions[bot]`. A pull request from a fork gets a read-only `GITHUB_TOKEN`, and those steps have no fork guard, so they fail on every external contribution. `SonarCloud` and `FOSSA` need secrets that forks never get, and `CI Gate` just aggregates all of it.

That is ours to fix, not yours, and a draft pull request for it is open: #13222. Please don't spend any time on those checks.

Everything that actually exercises the code is already green: unit, types, lint, emitted types, both builds, Playwright and UMD/UMD.min E2E across all three themes, and all three wrapper integrations.

### One request, once that fix lands

When #13222 is merged, please merge current `develop` into your branch. That does two things. It picks up the fork guard, so the gate should finally be satisfiable here. And it refreshes a stale CI signal: this run checked out `da7903f = Merge 04cc9dd into a3cddc465`, that is, your head merged with `develop` as of **2026-08-05**, which was 48 commits behind when I checked. A workflow *re-run* reuses that recorded merge ref rather than recomputing it, so only a fresh push gets the PR evaluated against today's pipeline.

### What I would like changed

Two inline comments marked **required**, plus two notes you can ignore. Nice catch on a genuinely subtle one. The object-data case is silent source-data corruption in a documented code path.

.changelogs/13073.json
ShiroKSH: Fixed in 6fdba2d2f. `issuesOrigin` is now `private`, so the generated entry links to pull request #13073.

handsontable/src/dataMap/sourceDataValidator.ts
ShiroKSH: Fixed in 6fdba2d2f. The per-cell path now precomputes visual and source columns before the row loop, and translates each visual row once. The unit coverage also verifies that `colToProp()` is called once per source column instead of once per cell.

handsontable/src/__tests__/settings/sourceDataValidator.spec.js
ShiroKSH: Added in 6fdba2d2f. The existing settings spec now covers a nested `meta.date` mapping and verifies that the invalid value is written back as `null` at the nested property.

handsontable/src/dataMap/sourceDataValidator.ts
ShiroKSH: Acknowledged. I left function-accessor writes unchanged in this PR, as requested.

handsontable/src/dataMap/sourceDataValidator.ts
```diff
@@ -180,9 +204,9 @@ function validatePerCell(
         continue;
       }
 
-      const value = dataSource.getAtCell(row, col);
+      const value = dataSource.getAtCell(row, visualColumn);
```
demtario: Follow-up to my earlier comment on this loop, which is why I left that thread unresolved.

The explicit `colToProp()` is gone from the inner loop, but the translation is still happening here, one level down. `getAtCell()` runs its second argument through `this.colToProp()` (`dataSource.ts:327-330`), which is `dataMap.colToProp` (wired at `replaceData.ts:147`), which calls `hot.toPhysicalColumn()`. So there is still one IndexMapper round trip per validated cell, plus a `modifyRowData()` call. What the precompute removed is the explicit `colToProp()` that this PR had itself introduced, which restores parity with `develop` but does not remove the round trip my comment was about.

That also makes this claim in your reply narrower than it reads:

> The unit coverage also verifies that `colToProp()` is called once per source column instead of once per cell.

It cannot verify that. The mock is `_getDataSource: () => ({ getAtCell, setAtCell })` with hand-rolled `jest.fn`s, so the spy never reaches `DataSource#getAtCell`. `expect(colToProp).toHaveBeenCalledTimes(2)` counts the call sites in this file, not the one in production.

Optional, and I am happy to merge without it: hoist the row read as well.

```ts
const dataRow = dataSource.modifyRowData(row);
// ...
const value = dataSource.getAtPhysicalCell(row, sourceColumn, dataRow);
```

That drops both `colToProp` and `modifyRowData` out of the inner loop, and it is what finally makes the precomputed `sourceColumn` pay for itself on the read side. `getAtPhysicalCell` is marked `@private`, so treat this as a suggestion rather than a request.

One trap worth recording for whoever touches this next: do **not** collapse it to `getAtCell(row, sourceColumn)`. For array data `sourceColumn` is a physical index, and `getAtCell` would feed it back through `colToProp` as if it were visual, which is exactly the bug this PR fixes.

handsontable/src/dataMap/sourceDataValidator.ts
```diff
@@ -163,13 +172,28 @@ function validatePerCell(
 ): void {
   const { rowIndexMapper, columnIndexMapper } = hotInstance;
   const metaManager = hotInstance._getMetaManager();
+  const visualColumns: (number | null)[] = [];
+  const sourceColumns: (string | number | null)[] = [];
+
+  for (let col = 0; col < colSourceCount; col += 1) {
```
demtario: Low, tidiness only, no change required to merge.

When `hasStoredMeta` is `true` and `fullScan` is `false` (a `cell` array or `setCellMeta()` is present, and every validator is row-independent), `collectColumnValidators()` has already built `visualColumn` and `sourceColumn` for every validator-bearing column at lines 139-144. `runSourceDataValidators` then discards that array and this function recomputes the mapping for all columns.

It is O(cols), so there is no perf argument. The point is that the mapping is now expressed twice, in two different shapes: the `ColumnValidator` fields, and these two parallel arrays. They have to stay in agreement by hand. Threading the collected columns through, or having `collectColumnValidators` always return the full column mapping, would leave a single owner.

handsontable/src/dataMap/sourceDataValidator.ts
```diff
@@ -163,13 +172,28 @@ function validatePerCell(
...
+    }
+
     for (let col = 0; col < colSourceCount; col += 1) {
-      const visualRow = rowIndexMapper.getVisualFromPhysicalIndex(row);
-      const visualColumn = columnIndexMapper.getVisualFromPhysicalIndex(col);
+      const visualColumn = visualColumns[col];
+      const sourceColumn = sourceColumns[col];
 
-      if (visualRow === null || visualColumn === null) {
+      if (visualColumn === null || sourceColumn === null) {
```
demtario: Nit: `sourceColumn === null` is unreachable at runtime.

`dataMap.colToProp()` never returns `null`. For a non-integer argument it returns the argument, and for an out-of-range or uncached column it falls back to the physical index (`dataMap.ts:253-272`). Even a `columns[].data` of `null` fails the `isDefined` cache check and takes that same fallback. The check is here purely to narrow `(string | number | null)[]` for TypeScript.

Either add a short comment saying so, or build `sourceColumns` as `(string | number)[]` inside the non-null branch of the precompute so the `null` only ever lives in `visualColumns`. As written, a future reader will go looking for the configuration that yields a `null` property, and there isn't one.

handsontable/src/__tests__/settings/sourceDataValidator.spec.js
```diff
@@ -104,6 +104,43 @@ describe('settings', () => {
...
+      handsontable({
+        data: [{ date: 'not-a-date' }],
+        columns: [{ data: 'date', type: 'date', dateFormat: 'YYYY-MM-DD' }],
+        allowInvalid: false,
+      });
+
+      expect(getSourceData()).toEqual([{ date: null }]);
+    });
+
+    it('should nullify a nested mapped property when allowInvalid is false', async() => {
```
demtario, line 117: This is the case I could not vouch for in my first pass, since it was added in response to the optional request. It does fail without the fix. I reverted only `sourceDataValidator.ts` to the PR base and kept your tests: `25 specs, 3 failures`, this one reporting `Expected $[0] not to have properties 0: null`. Both new unit cases fail too. Good.

Note only, **not** a change request, and the same shape as my `colToProp`-returns-a-function note. I will track it separately.

This case is green under the default `dataDotNotation: true`. With `dataDotNotation: false` the read path and the write path disagree, and the blanking corrupts the row rather than clearing it. Measured on your branch:

```js
data: [{ 'meta.date': 'not-a-date' }],
columns: [{ data: 'meta.date', type: 'date', dateFormat: 'YYYY-MM-DD' }],
dataDotNotation: false,
allowInvalid: false,
// getSourceData() -> [{ "meta.date": "not-a-date", "meta": { "date": null } }]
```

`DataSource#setAtCell` always writes through `setProperty()` (`dataSource.ts:268-276`), while `getAtPhysicalCell` honors `dataDotNotation` (`dataSource.ts:290-297`). So the invalid value survives under the flat key and a spurious nested object is injected next to it. `develop` is also wrong here (it wrote key `"0"`), so this is not a regression, and the asymmetry belongs to `DataSource` rather than to your change.

handsontable/src/dataMap/__tests__/sourceDataValidator.unit.js
```diff
@@ -244,6 +247,35 @@ describe('runSourceDataValidators', () => {
...
+    runSourceDataValidators(hot, 'init');
+
+    expect(getAtCell.mock.calls).toEqual([[0, 1], [0, 0], [1, 1], [1, 0]]);
+    expect(setAtCell.mock.calls).toEqual([
+      [0, 'first', null],
+      [0, 'second', null],
+      [1, 'first', null],
+      [1, 'second', null],
+    ]);
+    expect(toVisualColumn).toHaveBeenCalledTimes(rowIndependent ? 2 : 3);
```
demtario: Nit: please add a one-line comment for the `3`. The asymmetry reads as a bug until you trace it, because the extra call is not in this function's own loop. It comes from `collectColumnValidators` probing column 0 and returning `fullScan` before it reaches `colToProp`, so the per-cell path pays 1 + 2 while the batched path pays 2 + 0.

```js
// Per-cell: 1 probe in `collectColumnValidators` (early `fullScan` return) + 2 in the precompute.
```

demtario (commented):
@ShiroKSH Thanks, all four inline threads are addressed. I resolved three of them. I left the `colToProp` hoist thread open, because the change I asked for is there but it does not have the effect I was after. Details in a new comment on the read line.

### Verification

- Reverted **only** `handsontable/src/dataMap/sourceDataValidator.ts` to the PR base (`a3cddc465`) and kept your tests: `25 specs, 3 failures` on E2E, `2 failed, 17 passed` on unit. All three new E2E cases fail without the fix, including the nested `meta.date` case you added for the optional request. That is the `TESTING.md` bar, met.
- As-is: E2E `25 specs, 0 failures`, unit `19 passed`, `tsc -p handsontable/tsconfig.json --noEmit` clean, eslint clean on the changed file.
- The changelog entry now satisfies every rule in `.changelogs/README.md`.
- I also probed the one combination neither test covers, object data plus a moved column (`{a, b}` with `moveColumn(0, 2)` then `updateData`). It is correct: `[{"a":"2024-01-01","b":null}]`.

### Nothing blocking

`cla/signed` is green, so the unticked box in the description is cosmetic. The five red checks are still the fork-token ones: `Performance / run`, `Integration / preview packages`, `FOSSA`, `SonarCloud`, and `CI Gate` aggregating them. #13222 is still open, so my earlier "merge current develop into your branch" request is not actionable yet. Your base is still `a3cddc465` from 2026-08-08. I'll ping you when we merge that fix.

### For the follow-up ticket, not for this PR

Two things recorded here so they do not get lost. Both are confirmed by running them on your branch, and neither is yours to fix.

First, the `colToProp()`-returns-a-function case is worse than I described in my note. It is not one bad key. `setProperty()` splits the function's source text on every `.` in the body, so the write injects a nested tree:

```
{"d":"not-a-date","(row, value)=>{\n  if (typeof value === 'undefined') {\n    return row":{"d;\n  }\n  row":{"d = value;\n  }":null}}}
```

Second, the same physical-as-visual mismatch this PR fixes also lives at `core.ts:4125`. `setSourceDataAtCell` passes `changeProp`, which is a physical index for array data as its own comment at `:4139` states, into `dataSource.getAtCell()` to build the previous value for `beforeSetSourceDataAtCell`. With a moved column that hook receives a neighbor's old value.

Nothing here blocks the merge. The only item I would still like you to look at is the read-line comment, and even that is optional given the measured +1.7%.

demtario:
@ShiroKSH We just merged https://github.com/handsontable/handsontable/pull/13222, please rebase your changes on latest develop, and the CI checks should be able to pass now.

handsontable/src/dataMap/sourceDataValidator.ts
ShiroKSH: Implemented in 0eec3bf58. Each visible row is now resolved once with modifyRowData(), and reads use getAtPhysicalCell() with the precomputed source property. The focused unit test asserts the source-coordinate reads and one row resolution per row.

handsontable/src/dataMap/sourceDataValidator.ts
ShiroKSH: Also addressed in 0eec3bf58. collectColumnMappings() now owns the physical, visual, and source coordinate mapping, and both the batched and per-cell paths consume that same mapping.

handsontable/src/dataMap/sourceDataValidator.ts
ShiroKSH: Addressed in 0eec3bf58. Only columns with a visual index enter ColumnMapping, so sourceColumn is now non-nullable and the unreachable null branch is gone.

handsontable/src/dataMap/__tests__/sourceDataValidator.unit.js
ShiroKSH: Addressed by the mapping consolidation in 0eec3bf58. Both paths now translate each visible column once, so the previous 3-vs-2 assertion and its hidden extra probe no longer exist.

handsontable/src/dataMap/sourceDataValidator.ts
```diff
@@ -78,7 +85,7 @@ function validateSourceCell(
   }
 
   if (cellMeta.allowInvalid === false) {
-    dataSource.setAtCell(physicalRow, physicalColumn, null);
+    dataSource.setAtCell(physicalRow, sourceColumn, null);
```
demtario, line 88: **Low, note only, not a merge blocker.** Restating [my earlier note](https://github.com/handsontable/handsontable/pull/13073#discussion_r3841863062) against the line it actually lands on. I checked the backlog and there is no follow-up ticket for this yet, so these two comments are the only record.

This blanking call is the write half of a read/write asymmetry in `DataSource`. The PR inherits it rather than causing it:

- `setAtCell` always routes an object row through `setProperty()` (`dataSource.ts:268-276`), which splits on `.`.
- `getAtPhysicalCell` honors `dataDotNotation` (`dataSource.ts:290-297`), which does not.

Under `dataDotNotation: false` the two disagree, and blanking corrupts the row instead of clearing it:

```js
data: [{ 'meta.date': 'not-a-date' }],
columns: [{ data: 'meta.date', type: 'date', dateFormat: 'YYYY-MM-DD' }],
dataDotNotation: false,
allowInvalid: false,
// getSourceData() -> [{ "meta.date": "not-a-date", "meta": { "date": null } }]
```

The invalid value survives under the flat key and a spurious nested object is injected next to it. `develop` is wrong here too (it writes key `"0"`), so this is not a regression from your change.

The same line carries a second inherited case: `colToProp()` returns a function accessor when `columns[].data` is a function. `getAtPhysicalCell` handles that shape, but `setAtCell` takes the `!Number.isInteger(column)` branch and stringifies the function into a property key.

Both belong to `DataSource`, not to this file. Keeping the original thread open and adding this so the two cases sit on the line that would have to change.

demtario (commented):
4 inline notes below.

handsontable/src/dataMap/__tests__/sourceDataValidator.unit.js
```diff
@@ -41,21 +43,24 @@ function createMockHot({
     return meta;
   });
   const getAtCell = jest.fn((row, col) => (getValue ? getValue(row, col) : `${row}-${col}`));
+  const modifyRowData = jest.fn(row => ({ row }));
+  const getAtPhysicalCell = jest.fn((row, col) => getAtCell(row, col));
```
demtario, line 47: **Medium:** the mock drops `dataRow`, so the invariant this commit introduces is untested.

`getAtPhysicalCell` is mocked as `jest.fn((row, col) => getAtCell(row, col))`, discarding the third argument, and `createMockHot` does not return the spy (`:63`). The assertion at `:270` can therefore only observe `[row, sourceColumn]`. A wrong-row hoist (`modifyRowData(row - 1)`), or lifting the `let dataRow` declaration outside the row loop so it leaks across rows, passes every test in this file.

`modifyRowData` already returns an identifiable object (`jest.fn(row => ({ row }))`, `:46`), so pinning it is one assertion. Return `getAtPhysicalCell` from `createMockHot` and add:

```js
expect(getAtPhysicalCell.mock.calls).toEqual([
  [0, 'first', { row: 0 }], [0, 'second', { row: 0 }],
  [1, 'first', { row: 1 }], [1, 'second', { row: 1 }],
]);
```

That pins both the threading and the row identity, and it lets the indirection through `getAtCell` go away. Same shape as my earlier note on this file: a spy that never reaches the code under test cannot verify it.

handsontable/src/dataMap/sourceDataValidator.ts
```diff
@@ -91,6 +98,34 @@ function validateSourceCell(
   }
 }
 
+/**
+ * Resolves the source, physical, and visual coordinate mapping once for every visible column.
```
demtario: **Low:** "visible column" overstates what the filter does.

`collectColumnMappings` keeps a column when `getVisualFromPhysicalIndex()` is not `null`, which excludes **trimmed** columns only. A hidden column still has a visual index, and is still mapped and validated.

The same wording is at `:139` and `:183`. Worth tightening because the JSDoc just below is deliberately precise about this same distinction ("regardless of trimming", `:134`), and visual/physical/renderable confusion is exactly what this PR set out to fix. Suggest "non-trimmed column" in all three places.

handsontable/src/dataMap/sourceDataValidator.ts
```diff
@@ -273,7 +320,8 @@ export function runSourceDataValidators(hotInstance: HotInstance, source?: strin
   const settings = hotInstance.getSettings();
   const dataSource = hotInstance._getDataSource();
   const invalidByMessageType: InvalidItems = new Map();
-  const { fullScan, columns } = collectColumnValidators(hotInstance, colSourceCount);
+  const columnMappings = collectColumnMappings(hotInstance, colSourceCount);
```
demtario, line 323: **Low, note only, not a merge blocker.** This is the one cost the consolidation carries, and the consolidation is what I asked for, so treat it as a trade worth recording rather than an objection.

`collectColumnMappings` runs above the early return at `:328`, the one whose comment says "skip the scan entirely". On the common path where nothing carries a `sourceDataValidator`, the function now pays `colToProp()` per column, and `colToProp` goes through `toPhysicalColumn()`, so that is an IndexMapper round trip per column on every `init` and `updateData`. The bail path previously paid none of those: `collectColumnValidators` resolved `colToProp` only for validator-bearing columns.

It is O(cols) and sits next to an existing O(cols) `getCellMetaUncached` loop that costs far more, so the delta is small. Two ways to keep the single owner and drop it, if you want them: have `collectColumnMappings` resolve `sourceColumn` lazily, or hand `collectColumnValidators` only the visual indexes and build the source mapping below the early return.

handsontable/src/dataMap/__tests__/sourceDataValidator.unit.js
ShiroKSH: Addressed in 7730f6261. The mock now exposes `getAtPhysicalCell`, and the shared batched/per-cell test asserts that every read receives the row object returned by `modifyRowData` as its third argument.

Verification: focused source-data-validator unit suite, 19/19; ESLint on both changed files; pre-commit and pre-push gates.

handsontable/src/dataMap/sourceDataValidator.ts
ShiroKSH: Addressed in 7730f6261. All three JSDoc references now say “non-trimmed columns,” matching the index-mapper behavior.

Verification: ESLint on both changed files; pre-commit and pre-push gates.

demtario (approved):
@ShiroKSH Thanks for the whole contribution here, and for the thorough back-and-forth through review. Every round of feedback got addressed carefully, including the mock and JSDoc fixes in the latest commit. Great work, appreciate you taking it all the way through. Approving.

PR #13076, opened 2026-07-15
Add `selectionHandles` and `moveCells` selection options

sl01k:
### Context

The PR adds two opt-in, off-by-default selection features to the grid. Both are built into core selection + the Walkontable border, gated at runtime by their settings, styled via the CSS theme-token API, and documented in the API and the selection guide.

> Note for reviewers: this branch contains **two** features (`selectionHandles` and `moveCells`). They share the selection/border machinery, which is why they landed together. Each can be reviewed as an independent unit — see the two sections below.

---

## 1. `selectionHandles` — draggable handles to resize a selection

`selectionHandles: true` shows pill-shaped handles at the midpoint of each edge of the hovered selection. Dragging a handle resizes that edge (clamp, no flip, minimum one cell); the non-dragged axis and the focus cell are preserved.

- Off by default (`selectionHandles: false`); `@since 18.1.0`.
- **Styled entirely via CSS theme tokens** (`--ht-cell-selection-handle-*`) — background, border, radius, and size come from the tokens (no inline visual styles). Fill is the surface/background color; border is the accent color.
- **Per-theme shape** (via the `length` token): **main** 8×16 pill, **classic** 8×8 circle, **horizon** 8×24 pill.
- **Hover state**: hovering a handle boosts its background toward the accent color.
- **During a resize drag**: only the dragged handle stays visible (the others hide), the resize cursor (`ns-resize`/`ew-resize`) is held for the whole drag (including off-grid), and the dragged handle keeps its hover style.
- Light fade-in on appearance (respects `prefers-reduced-motion`).
- Hover-driven, per range; desktop only (touch keeps its own handles).
- Disabled in `selectionMode: 'single'`, for full-row/column/select-all selections, on grid-boundary or frozen-pane edges, while a cell editor is open, when the plugin is disabled at runtime, and when `disableVisualSelection` is set.
- New hook: `afterOnSelectionHandleMouseDown`.

## 2. `moveCells` — grab & drag a selection to move it

`moveCells: true` lets you grab the border of a selected cell or range (grab cursor) and drag it to relocate the block, like Excel/Google Sheets. Hold Ctrl/Cmd to copy; press Esc to cancel (registered through the ShortcutManager).

- Off by default (`moveCells: false`); `@since 18.1.0`.
- Works for a **single selected cell** as well as a contiguous multi-cell range.
- **Drag visual**: on grab, the source keeps its light-blue fill (its border is hidden), and a single thin dashed border — the "ghost", the same size as the selection with no fill — follows the pointer to show where the block will land. The ghost is rendered on the document body (`position: fixed`) so it stays correct in any host layout. The grabbing cursor is held for the whole drag.
- Moves cell values + the `className` cell meta (other meta stays at the source — documented in the option docs); with the `formulas` plugin, formula references adjust Excel-style via HyperFormula's `moveCells`/copy-paste.
- A plain click on a move band (press and release in the same cell) selects the cell under the pointer, exactly as an unswallowed click would.
- Single undo/redo step; the overwritten target and the cleared source are both restored on undo (deterministic snapshot restore, correct with formulas).
- Overwrite fires the cancelable `beforeMoveCells` hook; `afterMoveCells` fires after.
- Single contiguous range (or one cell) only; disabled for full-row/column/select-all/multiple selections, while a cell editor is open, and when the plugin is disabled at runtime; the source and target must stay in-grid and must not split a merged cell; the range may span at most 100,000 cells (`CELLS_LIMIT`, guards the public API against tab-freezing scans); neither the target nor the source may overlap read-only cells (a move has to clear the source, so a read-only source blocks a move but not a Ctrl/Cmd copy); hidden when `disableVisualSelection` is set.
- New hooks: `beforeMoveCells` (cancelable), `afterMoveCells`, `afterOnSelectionEdgeMouseDown`. New public method `hot.getPlugin('moveCells').moveCellRange(sourceRange, targetTopLeft, isCopy)`.

---

### Review fixes (round 1, @budnix)

All 11 review comments are addressed in `964f4e4358`:

1. `@since` tags bumped to **18.1.0** in all 8 places added by this PR.
2. All MoveCells private fields and static getters documented.
3. `MOVABLE_META_KEYS`/`collectMovableMeta` moved to a shared `src/utils/movableMeta.ts` — the UndoRedo action no longer imports from the MoveCells plugin, so registering just `UndoRedo` through `base.ts` pulls in no MoveCells code.
4. A click on a move band now selects the cell under the pointer instead of doing nothing (the band swallows the mousedown that would normally select).
5. `#isSelectionMovable` also checks `isEditorOpened()` and a new `tableProps.isPluginEnabled('moveCells')`, so a mid-edit drag cannot rewrite an uncommitted editor value and a runtime `disablePlugin()` actually hides the bands.
6. `isAdjustHandlesVisibleFor` gets the same two guards for the resize pills.
7. The `moveCells` option docs now state plainly that only the `className` meta travels.
8. The public `moveCellRange` validates the **source** range bounds too — an out-of-grid source returns a clean `false` instead of writing `undefined` into the target.
9. A `CELLS_LIMIT` (100,000 cells) ceiling bails with a console warning **before** any region scan or hook fires, so an oversized programmatic range cannot freeze the tab.
10. The Formulas plugin resets `#isRedoingMoveCells` in `afterUndo`, keeping it in step with the existing `setPerformRedo(false)` reset for cancelled redos.
11. The Esc drag-cancel is registered through the **ShortcutManager** (grid context, `runOnlyIf` gated to an active drag) instead of a native `keydown` listener.

### How has this been tested?

Automated coverage at every layer for both features:

- Unit (`npm run test:unit --prefix handsontable --testPathPattern="selection|moveCells|movableMeta|selectionHandles|dragToScroll|metaSchema|hooks|themes"`) — clamp/guard/boundary/map helpers, schema defaults, hook registration, theme tokens, lazy border-element gating, read-only source veto, right-click and rejected-drag guards, redo guard, source-bounds veto, `CELLS_LIMIT` veto (fires no hook, warns), shared movable-meta module.
- Walkontable (`npm run test:walkontable --prefix handsontable --testPathPattern="border|corner"`) — border element creation, positioning, styling, emit.
- E2E (`npm run build --prefix handsontable && cd tests && npx playwright test --project=e2e-main e2e/selection-handles.spec.ts e2e/move-zone.spec.ts e2e/move-cells-api.spec.ts e2e/move-cells-undo.spec.ts e2e/formulas-move-cells.spec.ts e2e/formulas-move-cells-undo.spec.ts`) — hover/resize/drag, only-dragged-handle-visible + resize cursor during drag, move/copy/veto/read-only/merge, single-cell move, Ctrl/Cmd-copy, Esc-cancel (now via ShortcutManager), click-selects-pointer-cell, bands/handles hidden while a cell editor is open, bands/handles hidden after a runtime `disablePlugin()`, RTL, axis preservation, focus stability, layer preservation, grab offset, ghost-follows-target, formula-reference adjustment, undo/redo, right-press guards.
- Visual regression — Playwright specs (resize handles visible on hover; move drag preview with dashed ghost). Per-theme handle shapes, hover boost, resize cursor, and ghost positioning were additionally verified interactively via Playwright.

Full suites are green locally after the review-fix commit — core unit (3333), full Playwright `e2e-main` (160), legacy Jasmine E2E for the affected areas (formulas 332, selection 1268, undoRedo/autofill/keyboardShortcuts/editors), and all three wrapper suites (React 75, Vue 3 32, Angular 291); `test:types`, `stylelint`, and `eslint` pass; both `handsontable.js` and `handsontable.full.js` build and contain both features. (The two `visual-tests` specs for these features — `move-cells-drag-preview` and `selection-handles-visible-on-hover` — time out on `locator.waitFor` locally; verified as pre-existing by re-running them at the branch point, so they are not caused by the code here. CI/Argos provides the authoritative result.)

### Types of changes

### Related issue(s):

Internal feature requests (no public issue).

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

- [x] My change requires a change to the documentation (selection guide sections + demos, option/hook JSDoc, and two `.changelogs/` entries added).

### Maintainer notes

- The theme tokens for both features (`cell-selection-handle-*`, incl. the per-theme `length` values) were hand-authored into `src/themes/static/**` because the Figma `tokens.json` export is not available in this environment; they must be added to the Figma source before `generate:themes` is next run, or the handles render unstyled. Noted in `scripts/themes/figma/.ai/CONCERNS.md`.
- The public `moveCells` grid option is intentionally named despite colliding with HyperFormula's `engine.moveCells()` method (different namespaces); the distinction is documented in `handsontable/src/3rdparty/walkontable/AGENTS.md`.
- Known limitation: resizing a non-last selection layer reports the last layer's coordinates in the public `afterSelection` hook (the `afterSetRangeEnd` handler in `core.ts` reads `selectionRange.current()`); the visual resize and scroll are correct.

---

> [!NOTE]
> **High Risk**
> Large changes to selection, border rendering, data moves, and HyperFormula sync; mistakes can corrupt grid data, formulas, or undo behavior.
> 
> **Overview**
> Adds two **off-by-default** grid options: **`selectionHandles`** (desktop edge pills to resize a range) and **`moveCells`** (drag the selection border to relocate values, Ctrl/Cmd to copy).
> 
> **`SelectionHandles`** and **`MoveCells`** plugins wire Walkontable border callbacks into core selection. **`Border`** now lazily creates desktop resize handles (`.wtSelectionHandle`) and move bands (`.wtMoveZone`) only when the option is on, with frozen-pane edge suppression. **`DragToScroll`** also auto-scrolls during move and handle drags when those plugins report an active drag.
> 
> **`moveCellRange`** is the programmatic API, with guards for no-op moves, **100k** cell limit, read-only source/target, merges, and bounds. Sparse **`className`** meta transfer avoids materializing cell meta on unstyled blocks; **`removeCellMeta`** peeks instead of creating empty meta objects.
> 
> With **Formulas**, HyperFormula runs **before** grid mutation via **`commitPendingMoveCells`**; hook listeners use shape guards and internal state so threaded hook return values cannot desync the engine from the data source. New hooks: **`beforeMoveCells`**, **`afterMoveCells`**, **`afterOnSelectionHandleMouseDown`**, **`afterOnSelectionEdgeMouseDown`**.
> 
> Docs, theme tokens for handle styling, framework selection examples, and visual-test demos are included.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bfc7e8a028963abcd5f0f7db2065404b727b16a5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-1928 Improve beforeChange/afterChange updateSettings() guides](https://app.clickup.com/t/9015210959/DEV-1928)

sl01k:
This is experimental PR and branch.

adrianspdev:
## Follow-up proposal: extract `selectionHandles` and `moveCells` into plugins

Both features currently live inside core (`tableView.ts` +598 lines, `core.ts` +238 lines). Neither needs to: the codebase already has an exact precedent for this shape of feature — **Autofill**. There, Walkontable renders the visual affordance (the corner square), and a standalone plugin drives the whole drag interaction through hooks (`afterOnCellCornerMouseDown`, `beforeOnCellMouseOver`). Both new features can follow the same split, which would keep `TableView` lean, make the features tree-shakeable via `registerPlugin`, and give them the standard plugin lifecycle (`updateSettings` toggling, automatic hook cleanup) for free.

### What goes where

| Piece | Today | After extraction |
|---|---|---|
| Handle / move-zone DOM rendering (`walkontable/selection/border/border.ts`) | Core | **Stays in core** — rendering engine, same as the autofill corner |
| `selectionHandles` / `moveCells` options (`metaSchema.ts`) | Core | **Stays in core** — plugin options live there anyway |
| Theme CSS + `--ht-cell-selection-handle-*` tokens | Core | **Stays in core** — same as autofill corner styles |
| Eligibility checks + hovered-layer index (`selection/selection.ts`) | Core | **Stays in core** — thin, feeds the Walkontable render path |
| `#adjustDrag` / `#moveDrag` state machines, ghost element, document listeners (`tableView.ts`) | Core | **Moves** → two plugins, using the plugin `EventManager` |
| `moveCellRange` + movable-meta helpers (`core.ts`) | Core public API | **Moves** → `getPlugin('moveCells').moveCellRange()` |
| `selection/moveCells.ts`, `selection/handleAdjust.ts` helpers | Core | **Moves** → plugin submodules |
| Formulas integration (`beforeMoveCells`/`afterMoveCells` listeners) | Formulas plugin | **No change** — already pure hook listeners |
| UndoRedo action (`undoRedo/actions/moveCells.ts`) | UndoRedo plugin | **No change** — already pure hook listeners |

### What's missing for the extraction

- One new hook: `afterOnSelectionEdgeMouseDown`. The Walkontable `onSelectionEdgeMouseDown` callback currently calls `#startMoveDrag` directly in `TableView`; the handle equivalent (`afterOnSelectionHandleMouseDown`) already exists as a hook.
- Hover-layer tracking in the plugin via the `beforeOnCellMouseOver` hook, the same way Autofill tracks the pointer during a fill drag.
- Standard plugin scaffolding ×2 (key, priority, registration, `isEnabled()` reading the existing option) and relocating the tests.

demtario:
## Code review responses

All findings addressed in 20c63ce. Every fix was verified by reverting it and confirming the new test goes red. Five of the review's premises turned out to be wrong — noted per item, because the correction changed the fix.

### Critical

**C1 — 5 new Jasmine specs in the frozen suite.**
Done. Deleted all five and ported to Playwright: new `move-cells-api.spec.ts` (16) and `move-cells-undo.spec.ts` (7), plus the non-overlapping drag/hover cases into `move-zone.spec.ts` (+14) and `selection-handles.spec.ts` (+19).
Two corrections: migration was **not** cheap — ~60 Jasmine tests vs ~20 Playwright with only ~1/3 overlap, so ~40 cases needed porting, not deleting. And the presence gate did **not** flag these: `JASMINE_TREE` at `presence-gate.mjs:51` is `/^handsontable\/src\/.*\/__tests__\//`, which requires an intermediate directory, so `handsontable/src/__tests__/` escapes it. Fixed separately in #13144. Also: `handsontable/src/__tests__/settings/` is a normal location (51 `spec.js` on develop), so the relocation request is a separate item from the freeze item — the ported specs live in `tests/e2e/`, satisfying both.

**C2 — handle/move-zone DOM created unconditionally.**
Done. Created lazily in `appear()` behind the existing visibility predicates.
Cost was worse than reported: not 8 divs + 4 listeners. `registerListeners()` iterates `this.main.childNodes` for `mouseenter`, and the 8 divs are appended during `createBorders()`, so that loop grew 5 → 13. Net **+8 nodes, +16 listeners per Border**, per highlight type, per overlay.
Two consequences worth flagging: the handle `mousedown` registration had to move into `createAdjustHandles()` (`registerListeners()` has already run by `appear()` time, so the handles would have rendered but silently stopped responding); and this also removes the lazily created elements from the `mouseenter` set, whose restore path sets `display: block` unconditionally and would have un-hidden a handle the predicate wants hidden.

**C3 — read-only source diverges between formulas-on and formulas-off.**
Fixed, but the diagnosis was wrong: the two paths do **not** diverge. `formulas.ts:1817-1822` uses the same `populateFromArray(..., 'auto')` as `moveCells.ts`, so `core.ts:1434` skips read-only source cells on both. Formulas-on is strictly worse, not different — `engine.moveCells` has already run, so HyperFormula and the data source desync (formula cells read empty, plain cells read stale), and `isItPossibleToMoveCells` is a pure HF structural check that cannot see `readOnly`.
So the fix is not "reconcile two behaviors" but "source read-only was unhandled everywhere": the existing check now covers the source rect, before the engine is touched. Copies still allowed (they keep the source). The `className` strip needed no separate gate — it and the veto are both `!isCopy`-gated, so it is unreachable on a read-only source. Tested in both engine states; no such test existed before.

### High

**H1 — no right-click guard.** Done, all four sites: both border listeners and both plugin listeners. `isRightClick` comes from `helpers/dom/event`, which `border.ts:12` already imports. Note the tests cover the plugin-level guards; the border-level ones are defense-in-depth and remain uncovered (a discriminating assertion is awkward — `contextmenu` fires independently of `preventDefault` on mousedown). Said so in the test comments rather than implying coverage.

**H2 — unguarded `getPlugin` in redo.** Done, bails with `redoneCallback({ wasRedone: false })` before registering the hook.
Two premises off: `base.ts` registers zero plugins (211/223 are type-only; registration is `plugins/index.ts:127,138`), and no sibling action guards either (`filters`, `rowMove`, `columnMove`, `mergeCells`, `columnSort` all dereference directly). So this introduces the pattern rather than following it — kept anyway, since `base.ts` is the tree-shakeable entry and it is three lines.

**H3 — `getCellMeta` in range loops.** Done, both switched to `getCellMetaTransient`. Real lines are `undoRedo/actions/moveCells.ts:69` and `moveCells.ts:479` (the review's 448/435 were stale).

**H4 — DragToScroll auto-scrolls rejected drags.** Fixed; DragToScroll now checks the owner's drag state.
Correction: there are no `addHook` `orderIndex` args anywhere, so "priority 100 at 25" is not the mechanism — but ordering *is* deterministic via `PLUGIN_PRIORITY` (`registry.ts:37`: ascending init order), so MoveCells (25) and SelectionHandles (24) settle before DragToScroll (100). Verified empirically rather than by reading, because getting this backwards would have silently disabled move auto-scroll: `moveCells.isDragActive=true` and `dragToScroll.isListening=true` on a legitimate drag. Did **not** add `orderIndex` — that would encode the ordering the bug depended on.

### Medium

**M1 — misleading callback signature.** Done, object type declared once, cast dropped. Worth noting the object form is new: the other 12 actions type the param as `HookCallback` and invoke it with zero args; `moveCells` is the only one passing anything.

**M2 — duplicated `MOVABLE_META_KEYS`.** Done, exported from the existing `plugins/moveCells/helpers.ts` and imported by the UndoRedo action.
Premise wrong: cross-plugin import is not forbidden — `filters.ts:11` imports `SEPARATOR` from `../contextMenu/predefinedItems`, and undoRedo already calls `getPlugin('moveCells').moveCellRange` at runtime. Did not hoist to `src/helpers/` (2 unrelated globals) or `src/selection/moveCells.ts` (a range-shape predicate file, not in the selection barrel — a plugin deep-importing it has no precedent).

**M3 — `commitPendingMoveCells` untagged.** Added `@private` and documented the two-phase protocol. Did **not** retype `#pendingMoveCells` with HyperFormula types: `plugins/formulas/engine/types.ts` is deliberately a minimal duck-type with a `[key: string]: any` escape hatch, so `object` is in-convention there. Also tagged the two new `isDragActive()` methods `@private` for the same reason.

**M4 — plugin mutates core `CellRange` objects.** Agreed, and confirmed no Selection method resizes a layer edge today. **Deferred** — it means adding public core API to an already-large PR. Happy to file it as a follow-up.

**M5 — stale JSDoc.** Done, retargeted to the SelectionHandles plugin (confirmed zero `HandlesHoveredLayer` references in `tableView.ts`).

**M6 — `refresh()` during `hot.destroy()`.** **Not changed**, and I think this one should be dropped. `AGENTS.md:66` says the opposite of the cited hazard — it is a *safety guarantee* that a plugin's `disablePlugin` running after core destroy stays safe. `setHandlesHoveredLayer` already early-returns when unchanged, and `refresh()` self-guards on `isSelected()` and zero counts. The one genuine window is `updatePlugin()` (disable→enable with a hovered layer live), not destroy — and guarding it would invent a convention with a single weak precedent (`nestedRows.ts:517` is the only `isDestroyed` check in all of `plugins/`, and not in `disablePlugin`). Glad to add it if you disagree.

### Low

**L1 — ghost style from JS.** Partially: fixed the `_selection.scss` comment, which contradicted itself (first half "overlays container, positioned absolutely", second half `document.body`) and attributed the inline styling to `TableView` instead of the plugin. Did **not** add the `.wtMoveGhost` rule — the file is a single theme-scoped `@mixin output`, and the ghost is a `document.body` child with `position: fixed` (deliberate, so no host overflow clips it), so a rule there could never match and the theme tokens do not resolve on it. Restructured the JS to name the two token-derived values instead.

**L2 — inconsistent event narrowing.** Done, cross-realm `instanceof` in both.

**L3 — stray blank line.** Done, plus the companion stray in the same 8-line diff: an unused `import type CellCoords` at `tableView.ts:7` (the new callbacks use string-literal edges).

**L4 — stale PR body.** Updated. `hot.moveCellRange(...)` never existed — corrected to `hot.getPlugin('moveCells').moveCellRange(...)`; added the missing `afterOnSelectionEdgeMouseDown`; corrected the read-only limitation and the test commands (which still pointed at the deleted Jasmine specs). The `tableView.ts +598 / core.ts +238` figures were already gone.

### Also in this commit

Fixed a latent bug in existing test infrastructure: `tests/fixtures/demo/selection-features.html` shared one `data` array by reference across re-inits, so a `moveCells` mutation in one test leaked into the next. Now rebuilt per init.

### Verification

Unit 3292 · Jasmine E2E 9599 · Walkontable 808 · Playwright 98 — all green; 0 eslint errors, `stylelint` and `test:types` pass; both bundles build.

The two `visual-tests` specs for these features (`move-cells-drag-preview`, `selection-handles-visible-on-hover`) time out on `locator.waitFor` locally. I confirmed this is **pre-existing** by stashing everything, rebuilding at the branch point, and reproducing the identical 2 failed / 89 passed — not caused by the C2 or L1 changes. Worth a separate look.

adrianspdev:
@demtario Found 2 issues worth checking:

### 1. A plain click on a move zone commits a no-op move (spurious undo step + real HyperFormula mutation)

`handsontable/src/plugins/moveCells/moveCells.ts:370-425`

`#onMouseUp` calls `#endDrag(true, event)` unconditionally, and neither `#endDrag` nor `moveCellRange` checks that the pointer actually moved or that the target differs from the source. A mousedown and mouseup on the same pixel resolves `clampMoveTarget` back to the source top-left and then runs the whole commit pipeline:

- `beforeMoveCells` fires, so `MoveCellsAction` snapshots both the source and the target region.
- With the `formulas` plugin active, `commitPendingMoveCells` runs `engine.moveCells(source, source)` inside `engine.batch`, which pushes a HyperFormula undo step.
- `populateFromArray` clears the source and rewrites it.
- `afterMoveCells` fires, so `plugin.done()` pushes a `MoveCellsAction` onto the Handsontable undo stack.

The result is that every click on a selection border consumes one undo entry (and one HyperFormula undo entry) for zero data change. Repeated clicks push the user's real edits out of the undo stack.

macOS makes this worse. A Ctrl+click there is `button === 0` with `ctrlKey: true`, so it passes the `isRightClick` guard in `#onSelectionEdgeMouseDown`, opens the context menu, **and** commits a no-op *copy* on mouseup because `#endDrag` reads `event.ctrlKey || event.metaKey`.

No test covers this — `tests/e2e/move-zone.spec.ts` has no click-without-drag case.

Suggested fix: bail out in `#endDrag` when the resolved target equals the drag origin.

```js
if (target.row === drag.fromRow && target.col === drag.fromCol) {
  return;
}
```

Either there or as an early return in `moveCellRange` when `targetTopLeft` matches the source top-left. Please add an E2E case for "mousedown + mouseup on the move zone without moving leaves the undo stack untouched", and a macOS-style Ctrl+click variant.

---

### 2. A move is O(range area) in allocations and per-cell meta calls, with no size cap

`handsontable/src/plugins/moveCells/moveCells.ts:190-216` and `:513-523`
`handsontable/src/plugins/moveCells/helpers.ts:54-66`
`handsontable/src/plugins/undoRedo/actions/moveCells.ts:50-92`

Full-row, full-column, and select-all selections are correctly blocked, but a user can still drag-select an arbitrarily large contiguous block (for example 50k x 20) and move it. A single commit then performs, per cell:

- `#hasReadOnlyCell` — one `getCellMetaTransient` per cell, run twice (target range, then source range).
- `buildMoveMap` — one object per cell; `snapshots` — one object per cell; `targetKeys` — a `Set` holding one string per cell.
- `#writeMeta` — a `setCellMeta` or `removeCellMeta` per cell. `setCellMeta` **permanently materializes** cell meta across the whole target block. That is precisely the O(visited cells) retention pattern that `handsontable/CLAUDE.md` documents as the root cause of the OOM investigation.
- `MoveCellsAction` snapshots the source **and** target regions (values plus meta), and `UndoRedo.undo()`/`redo()` run `deepClone` (`JSON.parse(JSON.stringify(...))`) over that entire action on every undo and redo.

Nothing bounds the range size, so this is reachable through ordinary UI interaction rather than only through the API.

Suggested fix: add a cell-count guard that vetoes the move and emits a `warn()` above a threshold. At minimum, skip the `#writeMeta` work for cells whose snapshot is empty and whose target carries no stored meta, so an unstyled block does not materialize meta for its entire target region.

demtario:
@adrianspdev Both fixed in 0c2b2410f7.

**1. No-op click** — two guards. `moveCellRange` returns `false` before any hook fires when the target equals the source top-left (covers the API, the Ctrl+click copy, and redo). Additionally `#endDrag` skips the commit when the gesture starts and ends in the same cell — needed for a case beyond the report: the move bands straddle the selection border, so a plain click's press point can resolve to the cell just outside the range, which committed a real 1-cell move, not just a no-op. E2E added: plain click, macOS-style Ctrl+click, and drag-released-at-origin all assert the undo stack stays untouched.

**2. O(area) work** — fixed the per-cell costs instead of adding a size cap (copyPaste/autofill are uncapped by default too, so a cap here would be inconsistent):
- `CellMeta.removeMeta` no longer materializes a meta object just to delete a key — this was the root of the permanent O(area) retention.
- The plugin's meta pass is sparse now: both regions are scanned transiently up front and only cells owning a movable key produce `setCellMeta`/`removeCellMeta` calls; an unstyled block performs zero meta writes. The `targetKeys` string `Set` became an arithmetic bounds check.
- The undo action snapshots meta sparsely (one entry per styled cell instead of one `{}` per region cell), which also shrinks what `deepClone` copies on every undo/redo.

The remaining O(area) cost is the value snapshots for undo — the same class as an uncapped copyPaste. Unit guard added: moving an unstyled block materializes no cell meta.

handsontable/src/3rdparty/walkontable/src/selection/border/border.ts
```diff
@@ -378,6 +402,163 @@ class Border {
...
+  createMoveZone() {
+    const { rootDocument } = this.wot;
+
+    const make = (edge: string) => {
+      const el = rootDocument.createElement('div');
+
+      el.className = 'wtMoveZone';
+      el.style.position = 'absolute';
+      el.style.display = 'none';
+      el.style.zIndex = '100';
```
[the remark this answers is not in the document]
demtario: Fixed in c16588b54c — `.wtBorder.corner` raised to z-index 150: above the bands (100), below the resize pills (200). Lowering the bands instead would put them under the selection border lines (z-index 10) and make the visible border ungrabbable. E2E added: pressing the fill handle with `moveCells` on autofills and never starts a move drag.

handsontable/src/plugins/formulas/formulas.ts
```diff
@@ -1517,6 +1582,270 @@ export class Formulas extends BasePlugin {
...
+      // MOVE (e.g. pasting over part of an array formula throws in `engine.paste`), so the
+      // operation vetoes cleanly before the grid mutates instead of failing halfway through.
+      const targetRegion = {
+        start: dest,
+        end: {
+          sheet: this.sheetId,
+          row: hfTargetRow + (hfToRow - hfFromRow),
+          col: hfTargetCol + (hfToCol - hfFromCol),
+        },
+      };
```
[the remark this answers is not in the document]
demtario: Fixed in cd9fd0fb10. `#onBeforeMoveCells` now vetoes the operation unless all four visual spans (source rows/cols, target rows/cols) map to consecutive HF indexes — covers trimmed/filtered rows and sorted/manually-moved orders alike. Veto over a per-cell fallback: it matches the existing engine-side vetoes (`isItPossibleToMoveCells` etc.), keeps the operation atomic, and a rectangle op has no cheap correct per-cell equivalent for relative-reference adjustment. E2E: move across a trimmed row is rejected with nothing mutated; a contiguous move on a grid that merely has trimmed rows elsewhere still works.

adrianspdev:
@demtario a few more findings:

## 1 — truthy `beforeMoveCells` return crashes internal listeners

**File:** `handsontable/src/plugins/undoRedo/actions/moveCells.ts:189` (also `handsontable/src/plugins/formulas/formulas.ts` `#onBeforeMoveCells`)

A `beforeMoveCells` listener registered on the global bucket (`Handsontable.hooks.add`) that returns a truthy value replaces `sourceRange` for the listeners that run after it — this internal listener then throws `TypeError: src.getTopStartCorner is not a function` (reproduced with a unit-test harness; settings-level and `hot.addHook` listeners are safe because they run after the internal ones). The `sourceRange === false` guard only covers the documented veto value.

**Fix:** guard the shape instead of the literal, e.g.:

```ts
if (sourceRange === false || typeof (sourceRange as CellRange)?.getTopStartCorner !== 'function') {
  return false;
}
```

Same pattern as the defensive filtering in `dataChange.ts`'s `beforeChange` listener. Apply in both internal listeners.

---

## 2 — Formulas redo fix rides along untested and unlogged

**File:** `handsontable/src/plugins/formulas/formulas.ts` (~line 520, `afterUndo` → `afterRedo` change)

This fixes a real pre-existing bug (develop registers two `afterUndo` hooks, so `setPerformRedo(false)` never fired after a redo and the flag leaked until the next undo), but it changes redo behavior for **every** action type with formulas enabled, not just `move_cells` — and it ships without a dedicated test or changelog entry.

**Fix:** add a regression test asserting `indexSyncer` leaves the redo state after a non-moveCells redo, and mention the fix in the changelog (or split it into its own PR).

---

## 3 — hand-authored theme tokens wiped by the next `generate:themes`

**Files:** `handsontable/src/themes/static/**`, `handsontable/scripts/themes/figma/.ai/CONCERNS.md`

The six hand-authored `cell-selection-handle-*` tokens live in generator output that `generate:themes` deletes wholesale (`rmSync(OUTPUT_PATH)`). The next generator run from a `tokens.json` that lacks them ships unstyled selection handles. CONCERNS.md documents the risk, but nothing enforces it.

**Fix:** add the tokens to the Figma source before merge, or make the generator (or a CI check) fail when a key listed in `tokensKeys.mjs` is missing from the export, so the wipe cannot silently drop them.

---

demtario:
@adrianspdev All three fixed in b6b522a807.

**1. Truthy `beforeMoveCells` return** — new `isCellRangeLike` duck-type predicate exported from `cell/range.ts` (cross-realm safe, so no `instanceof`); both internal `beforeMoveCells` listeners guard with it and veto instead of crashing. Also applied to the Formulas `afterMoveCells` sync, which dereferences the same foldable argument. Unit tests cover the UndoRedo listener (global-bucket garbage/`true`/`undefined` returns); a Playwright case covers the Formulas listener.

**2. Redo-state fix** — **extracted to its own PR: #13148.** This branch no longer contains the fix, its tests, or a changelog entry for it — it reverted to develop's registration and gets the fix back through a develop merge once #13148 lands. The new PR carries the one-line fix, a unit regression suite (flag reset after undo/redo, a behavioral post-redo row-move sync case, and a cancelled-redo case), and its own changelog entry.

**3. Theme token wipe** — `generate:themes` now validates before the `rmSync`: every key in `tokensKeys.mjs` must resolve for every theme or the run aborts listing the missing keys (`utils/validation.mjs`, unit-tested). Only documented exception is `density` (global variable group, never per-theme). The six handle tokens still need adding to the Figma token set for a raw export to regenerate cleanly — flagging that for the design team; CONCERNS.md updated accordingly.

handsontable/src/3rdparty/walkontable/src/selection/border/border.ts
```diff
@@ -1282,6 +1484,58 @@ class Border {
...
+      if (this.isFrozenBoundaryEdge('row', corners[0])) {
+        adjustHandles.styles.top.display = 'none';
+      }
+      if (this.isFrozenBoundaryEdge('column', corners[1])) {
+        adjustHandles.styles.start.display = 'none';
+      }
+      if (this.isFrozenStartBoundaryOppositeEdge('row', corners[2]) ||
+          this.isFrozenBottomBoundaryEdge(corners[2])) {
+        adjustHandles.styles.bottom.display = 'none';
+      }
```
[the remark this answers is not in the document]
demtario: Fixed in 1dab57b1d7 — new `isFrozenBottomBoundaryOppositeEdge` mirrors the existing seam guards, so the top handle is now suppressed when the selection starts at the first bottom-frozen row (`fromRow === totalRows - fixedRowsBottom`), completing the suppression matrix the option JSDoc promises. E2E added for both bottom-seam directions; the new test scopes to `.ht_clone_bottom` since a bottom-pane selection renders its handles on that overlay, and asserts the sibling handles stay visible so the hidden-top assertion can't pass vacuously.

handsontable/src/plugins/formulas/formulas.ts
```diff
@@ -462,24 +511,31 @@ export class Formulas extends BasePlugin {
...
-    this.addHook('beforeRedo', () => {
+    // Handling redo actions on data just using HyperFormula's UndoRedo mechanism.
+    this.addHook('beforeRedo', (action: { actionType: string }) => {
       this.indexSyncer!.setPerformRedo(true);
+      this.#isRedoingMoveCells = action.actionType === 'move_cells';
 
-      this.engine!.redo();
+      if (!this.#isRedoingMoveCells) {
+        this.engine!.redo();
+      }
```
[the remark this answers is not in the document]
demtario: Fixed in 4482eba11b. The handler now shape-guards the action (`in`-narrowed, no cast) and — when `actionType` is not trustworthy — cancels the redo by returning `false` rather than guessing a branch: either guess can run the engine operation twice or advance HyperFormula on a redo that later gets cancelled. The guard runs before `setPerformRedo(true)`, since a cancelled redo never fires `afterRedo` and a flag set earlier would leak (the exact bug fixed two commits ago). Playwright test: garbage-returning global `beforeRedo` listener → redo cancelled, grid and engine stay consistent, no leaked flag.

handsontable/scripts/themes/figma/.ai/CONCERNS.md
```diff
@@ -2,14 +2,31 @@
 
 Known sharp edges, technical debt, and behaviors that are easy to get wrong.
 
+## Hand-authored `cell-selection-handle-*` tokens are not yet in the Figma token set
```
adrianspdev: We need to add these tokens before merging

sl01k:
Tokens are ready on Figma:
Screenshot 2026-08-03 at 15 41 56

qunabu (dismissed):
needes @adrianspdev and @sl01k approve as well

handsontable/scripts/themes/figma/.ai/CONCERNS.md
adrianspdev: done

handsontable/src/core/hooks/constants.ts
```diff
@@ -609,6 +609,49 @@ export const REGISTERED_HOOKS = [
    */
   'afterOnCellCornerMouseDown',
 
+  /**
+   * Fired after the user presses a selection-adjustment handle (see [`selectionHandles`](@/api/options.md#selectionhandles)).
+   *
+   * @event Hooks#afterOnSelectionHandleMouseDown
+   * @since 18.0.0
```
budnix: Should be updated in all places
```suggestion
   * @since 18.1.0
```

handsontable/src/plugins/moveCells/moveCells.ts
```diff
@@ -0,0 +1,597 @@
...
+  /**
+   *
+   */
+  static get SETTING_KEYS() {
+    return [PLUGIN_KEY];
+  }
+
+  /**
+   *
+   */
```
budnix, line 68: Would be nice to document all private fileds

handsontable/src/plugins/moveCells/moveCells.ts
```diff
@@ -0,0 +1,597 @@
...
+
+    if (!drag) {
+      return;
+    }
+
+    this.#drag = null;
+    this.#pointerPosition = null;
+    this.eventManager.removeEventListener(this.hot.rootDocument.documentElement, 'mousemove', this.#onMouseMove);
+    this.eventManager.removeEventListener(this.hot.rootDocument.documentElement, 'mouseup', this.#onMouseUp);
+    this.eventManager.removeEventListener(this.hot.rootDocument.documentElement, 'keydown', this.#onKeyDown);
```
budnix: Here it should be registered in the shortcut manager module instead of using native keydown.

handsontable/src/plugins/undoRedo/actions/moveCells.ts
```diff
@@ -0,0 +1,332 @@
+import type { HookCallback } from '../../../core/hooks/bucket';
+import type { HotInstance } from '../../../core/types';
+import type CellCoords from '../../../3rdparty/walkontable/src/cell/coords';
+import type CellRange from '../../../3rdparty/walkontable/src/cell/range';
+import { isCellRangeLike } from '../../../3rdparty/walkontable/src/cell/range';
+import { BaseAction } from './_base';
+import { hasOwnProperty } from '../../../helpers/object';
+// Imported rather than duplicated: undo must restore exactly the key set `moveCellRange` moved, and
+// two copies would silently drift the moment a key is added to one of them.
+import { collectMovableMeta, MOVABLE_META_KEYS } from '../../moveCells/helpers';
```
budnix: The comment above this is right that two copies of the key list would drift apart, but `handsontable/CLAUDE.md` bans one plugin importing from another — that rule is what keeps each plugin removable on its own. Right now the tree-shakeable `base.ts` entry pulls `moveCells/helpers` in as soon as someone registers just `UndoRedo`. Moving `MOVABLE_META_KEYS` and `collectMovableMeta` to a shared spot would keep the single source of truth and the plugin boundary. Curious whether there was a reason to leave it here.

handsontable/src/plugins/moveCells/moveCells.ts
```diff
@@ -0,0 +1,597 @@
...
+      return;
+    }
+
+    const pointer = getCellCoordsFromMousePosition(this.hot, event.clientX, event.clientY);
+
+    // A gesture that starts and ends in the same cell is a click, not a move. The move bands
+    // straddle the selection border, so a press can resolve to the cell just OUTSIDE the range —
+    // committing there would shift the block by one cell on a plain click (and on macOS a
+    // Ctrl+click would commit a spurious copy the same way).
+    if (pointer.row === drag.pressRow && pointer.col === drag.pressCol) {
```
budnix, line 474: Your comment here nails why the press can land outside the range, and skipping the move is right. What's left over is the click itself: the band already called `stopImmediatePropagation()` and `preventDefault()`, so the cell under the pointer never gets selected either. With B2:C3 selected, a click 2px above B2 lands in B1 and does nothing at all — no move, and no selection change. Would it work to select `pointer` here instead of just returning?

handsontable/src/selection/selection.ts
```diff
@@ -964,6 +974,56 @@ class Selection {
...
+   * on, visual selection is not disabled, and the selection shape passes {@link canMoveRange}
+   * (exactly one contiguous range that is not a full row, full column, or header selection).
+   * Mirrors the eligibility gate applied when a move drag starts, so the move zone and grab
+   * cursor never show for a selection that a drag would reject.
+   *
+   * @private
+   * @returns {boolean}
+   */
+  #isSelectionMovable() {
+    if (this.settings.moveCells !== true || this.settings.disableVisualSelection) {
```
budnix: `isCellCornerVisible()` and `isAreaCornerVisible()` both include `!this.tableProps.isEditorOpened()`, but this one doesn't. So with an editor open on B2, pressing the move band starts a drag — and since the band swallows the mousedown, the usual outside-click close never runs, so the release rewrites a cell whose editor still holds an uncommitted value. Reading the setting instead of the plugin also means `getPlugin('moveCells').disablePlugin()` leaves the bands rendering and still calling `preventDefault()`, so you get a grab cursor on a border where mousedown now does nothing at all.

```suggestion
    if (this.settings.moveCells !== true || this.settings.disableVisualSelection ||
        this.tableProps.isEditorOpened()) {
```

handsontable/src/selection/selection.ts
```diff
@@ -1118,6 +1199,70 @@ class Selection {
...
 
+  /**
+   * Tells whether the adjustment handles should render for the given highlight layer.
+   *
+   * @private
+   * @param {number} layerLevel The area highlight layer level.
+   * @returns {boolean}
+   */
+  isAdjustHandlesVisibleFor(layerLevel: number) {
+    if (this.settings.selectionHandles !== true) {
```
budnix: Same gap as `#isSelectionMovable` above — no `isEditorOpened()` check, so the resize pills stay visible and draggable while an editor is open. Worth a quick check that resizing mid-edit does what you'd want.

```suggestion
    if (this.settings.selectionHandles !== true || this.tableProps.isEditorOpened()) {
```

handsontable/src/plugins/moveCells/helpers.ts
```diff
@@ -0,0 +1,103 @@
+import type { HotInstance } from '../../core/types';
+import { hasOwnProperty } from '../../helpers/object';
+
+/**
+ * The cell-meta keys that travel with a moved cell. Exported because the UndoRedo `MoveCellsAction`
+ * must restore exactly this set — a key added here but missed there would silently stop being undone.
+ */
+export const MOVABLE_META_KEYS: ReadonlyArray<string> = ['className'];
```
budnix: This is `className` only, but the `moveCells` doc in `metaSchema.ts` says the drag moves "values, formatting". A cell with `type: 'numeric'` and a `numericFormat` set through `setCellMeta` loses its format — the value lands at the target and the format stays behind at the source. Either the key list grows, or the doc says plainly that only `className` travels.

handsontable/src/plugins/moveCells/moveCells.ts
```diff
@@ -0,0 +1,597 @@
...
+    }
+
+    // A read-only cell vetoes the move from either end. The target case is the obvious one; the
+    // source case matters just as much, because `populateFromArray` skips read-only cells (see
+    // `core.ts`, which exempts only `'UndoRedo.undo'`). Without this check the source values survive
+    // and the move silently degrades into a copy — and with the Formulas plugin active it is worse
+    // still: HyperFormula has already relocated the cell, so the engine and the data source diverge.
+    if (
+      targetRow < 0 || targetCol < 0 ||
+      targetBottom >= this.hot.countRows() || targetRight >= this.hot.countCols() ||
```
budnix, line 208: This checks that the target stays inside the grid but never checks the source. The plugin's own drag path always hands over a real selection, so it doesn't bite there — but `moveCellRange` is documented public API, and a caller passing a source range past the last row gets `getSourceDataAtCell` reads off the end and `undefined` written into the target, instead of a clean `false`.

```suggestion
      targetBottom >= this.hot.countRows() || targetRight >= this.hot.countCols() ||
      fromRow < 0 || fromCol < 0 ||
      toRow >= this.hot.countRows() || toCol >= this.hot.countCols() ||
```

handsontable/src/plugins/moveCells/moveCells.ts
```diff
@@ -0,0 +1,597 @@
...
+    }
+
+    const formulas = this.hot.getPlugin('formulas');
+    const formulasActive = formulas?.enabled === true;
+
+    if (formulasActive && !formulas.commitPendingMoveCells()) {
+      return false;
+    }
+
+    const values = this.#getSourceValues(fromRow, fromCol, toRow, toCol);
```
budnix, line 243: Counting what runs before anything actually changes: `#hasReadOnlyCell` on the target and on the source, `collectMovableMeta` on both inside `#transferMovableMeta`, this `#getSourceValues`, plus `snapshotRegion` twice in UndoRedo's `beforeMoveCells` listener. That is roughly six full passes over both regions. Only full-row and full-column selections get rejected, so a 50k x 20 selection is allowed through — millions of `getCellMetaTransient` calls and two whole value matrices parked in the undo stack, and the tab freezes. Would a cell-count ceiling that bails early be worth it here, the way `copyPaste` has `rowsLimit`/`columnsLimit`?

handsontable/src/plugins/formulas/formulas.ts
```diff
@@ -481,9 +567,16 @@ export class Formulas extends BasePlugin {
       this.indexSyncer!.setPerformRedo(false);
     });
 
+    this.addHook('afterRedo', () => {
+      this.#isRedoingMoveCells = false;
```
budnix, line 608: `setPerformRedo(false)` gets a second reset in the `afterUndo` handler just above, with a comment explaining that a redo cancelled in `beforeRedo` never reaches `afterRedo`. This flag has the same exposure but no matching reset: if a listener registered after this one cancels a `move_cells` redo, `#isRedoingMoveCells` stays `true`, and the `isPerformingUndoRedo() && !#isRedoingMoveCells` check at line 1790 then reads the stale value. Adding it to that `afterUndo` handler would keep the two flags in step.

handsontable/src/plugins/moveCells/moveCells.ts
budnix: It would be better to use of our Shortcut Manager instead of using the native `keydown` event and the hardcoded `Escape` key in the callback.

budnix (changes requested):

handsontable/src/core/hooks/constants.ts
adrianspdev: Done in 964f4e4358 — bumped to `18.1.0` in all 8 places this PR added: `core/hooks/constants.ts` (4 hooks), `core/settings.ts` (2 typings), `metaSchema.ts` (both options).

handsontable/src/plugins/moveCells/moveCells.ts
adrianspdev: Done in 964f4e4358 — `#drag`, `#ghost`, and `#bodyCursor` now carry real descriptions, and the three static getters got JSDoc too.

handsontable/src/plugins/undoRedo/actions/moveCells.ts
adrianspdev: Good catch — no reason beyond oversight. Done in 964f4e4358: `MOVABLE_META_KEYS`, `collectMovableMeta`, and `MovableMetaEntry` moved to a shared `src/utils/movableMeta.ts`, imported by both the plugin and the UndoRedo action, so registering just `UndoRedo` through `base.ts` no longer pulls MoveCells code in.

handsontable/src/plugins/moveCells/moveCells.ts
adrianspdev: Yes — done in 964f4e4358. The click branch now calls `selectCell(pointer.row, pointer.col)`, so the cell under the pointer gets selected exactly as an unswallowed click would. Covered by a new Playwright test ("selects the cell under the pointer when a band press resolves to a click").

handsontable/src/selection/selection.ts
adrianspdev: Done in 964f4e4358 — applied the `isEditorOpened()` check, and for the runtime-disable half added a `tableProps.isPluginEnabled(name)` callback (backed by `getPlugin(name)?.enabled`) that the guard checks too, so `disablePlugin()` hides the bands even while the setting still reads `true`. Both behaviors covered by new Playwright tests.

handsontable/src/selection/selection.ts
adrianspdev: Done in 964f4e4358 — same two guards (`isEditorOpened()` + `isPluginEnabled('selectionHandles')`) applied here, with Playwright tests for the mid-edit and runtime-disable cases.

handsontable/src/plugins/moveCells/helpers.ts
adrianspdev: Went with the docs fix in 964f4e4358 — the `moveCells` option doc now says the drag moves values and the `className` cell meta, and that other meta (e.g. `numericFormat`, `readOnly`) stays at the source. Growing the key list has undo/HyperFormula implications, so it stays a conscious follow-up if we want it.

handsontable/src/plugins/moveCells/moveCells.ts
adrianspdev: Done in 964f4e4358 — applied the suggestion, plus unit tests (`rangeGuards.unit.js`) for the past-last-row/column cases and the exact-edge move. One note: the `fromRow < 0 || fromCol < 0` half is unreachable through a real `CellRange` (`getTopStartCorner()` normalizes header coords to 0) — kept as cheap defense anyway.

handsontable/src/plugins/moveCells/moveCells.ts
adrianspdev: Done in 964f4e4358 — added an exported `CELLS_LIMIT = 100000` ceiling that bails (with a console warning) right after the no-op guard, before any region scan and before `beforeMoveCells` fires, so UndoRedo never snapshots an oversized range. Kept it a constant rather than a setting since `moveCells` is a boolean — it can become configurable later without a breaking change. Unit tests assert the veto warns and fires no hook.

handsontable/src/plugins/formulas/formulas.ts
adrianspdev: Done in 964f4e4358 — `#isRedoingMoveCells = false` added to that `afterUndo` handler, right next to the `setPerformRedo(false)` reset, with the comment extended to cover both flags.

handsontable/src/plugins/moveCells/moveCells.ts
adrianspdev: Done in 964f4e4358 — the drag cancel is now a grid-context shortcut (`keys: [['Escape']]`, `runOnlyIf: () => this.#drag !== null`, group `moveCells`) registered in `enablePlugin` and removed in `disablePlugin`; the native `keydown` listener and `#onKeyDown` are gone. The existing Esc-cancel Playwright tests now exercise the ShortcutManager path and stay green.

budnix (approved):

demtario (approved):

Issue #13072, opened 2026-07-14
Typescript example from docs fails to build


captain-orange2:
### Describe the bug

https://handsontable.com/docs/react-data-grid/typescript-types/#type-the-hot-instance-in-a-react-ref

```
> hot-18-test@0.0.0 build
> tsc -b && vite build

src/ExampleComponent.tsx:1:8 - error TS6133: 'React' is declared but its value is never read.

1 import React, { useRef } from 'react';
         ~~~~~

src/ExampleComponent.tsx:26:20 - error TS2322: Type 'RefObject<HotInstance | null>' is not assignable to type 'Ref<HotTableRef> | undefined'.
  Type 'RefObject<HotInstance | null>' is not assignable to type 'RefObject<HotTableRef | null>'.
    Type 'HotInstance | null' is not assignable to type 'HotTableRef | null'.
      Type 'HotInstance' is missing the following properties from type 'HotTableRef': hotElementRef, hotInstance

26   return <HotTable ref={hotRef} settings={settings} />;
                      ~~~

  node_modules/@types/react/index.d.ts:301:9
    301         ref?: Ref<T> | undefined;
                ~~~
    The expected type comes from property 'ref' which is declared here on type 'IntrinsicAttributes & HotTableProps & RefAttributes<HotTableRef>'


Found 2 errors.
```

### Provide a link to the demo with the bug reproduction

https://stackblitz.com/edit/vitejs-vite-dgs8yli3?file=src%2FExampleComponent.tsx

### Handsontable version

18.0.0

### Framework version

React 19

### Your environment

any

demtario:
Thanks for the report!

Two issues in that example:
1. The ref was typed as `HotInstance`, but `HotTable`'s ref is actually `HotTableRef` (`{ hotElementRef, hotInstance }`) — that's the TS2322 you're seeing.
2. It passed a `settings` prop, which isn't part of the current `@handsontable/react-wrapper` API (that was the old `@handsontable/react` package). The wrapper takes grid options as individual props instead.

Fixed example:

```tsx
import { useRef } from 'react';
import { HotTable } from '@handsontable/react-wrapper';
import type { HotTableRef, HotTableProps } from '@handsontable/react-wrapper';

export function Grid() {
  const hotRef = useRef<HotTableRef>(null);

  const settings: HotTableProps = {
    data: myData,
    licenseKey: 'non-commercial-and-evaluation',
  };

  return <HotTable ref={hotRef} {...settings} />;
}
```

[...]

PR #13079, opened 2026-07-15
DEV-2072: Fix Events type collapsing to any

demtario:
### Context

The PR fixes the exported `Events` type resolving every hook lookup to `any`.

`GridSettings` carries a `[key: string]: any` index signature (for plugin-specific keys), which makes `keyof GridSettings` include `string`. The `HookKey` helper that derives `Events` mapped over the raw `keyof GridSettings`, so it collapsed to `string`, and `Events = Required<Pick<GridSettings, HookKey>>` degraded to a bare `{ [key: string]: any }`. As a result, `Events['afterChange']` was `any`, callbacks typed through `Events[hookName]` had implicit-`any` parameters (failing users' strict builds — this is what broke the docs example investigated in #13072), and the typed `addHook` overloads provided no real checking or autocomplete.

**The fix** derives `HookKey` from `RemoveIndexSignature<GridSettings>` — the same helper the React wrapper already uses for exactly this index-signature-collapse problem.

Tightening `Events` surfaced that many internal `addHook` call sites have callback annotations that don't exactly match the declared hook signatures (previously unchecked because `Events[K]` was `any`). Rather than masking this with `any` or force-fixing ~90 call sites in one PR, the `addHook`/`addHookOnce`/`removeHook` methods now form a documented three-rung overload ladder, with no `any` anywhere:

1. **Strict** — known hook name, callback typed `Events[K]`. Unannotated inline arrows get full parameter inference from this rung.
2. **Lenient** — known hook name, any explicitly-annotated callback. `HookCallback` is now `(...args: never[]) => unknown` — the sound function "top type": every function is assignable to it (parameter contravariance), but it cannot be invoked without a deliberate cast. This keeps all previously-compiling listener code (internal and user) compiling.
3. **Fallback** — any string, for dynamically-named and plugin-local hooks. Because `HookCallback` is the top type, this rung now also accepts typed callbacks (previously the `(...args: unknown[])` fallback rejected any concretely-typed callback).

The single place listeners are dynamically invoked (`fastCall` in `helpers/function.ts`) carries the one deliberate, documented cast. No runtime behavior changes anywhere.

Also fixes a parity gap found during the audit: the documented `afterOnCellMouseOverOutside` / `beforeOnCellMouseOverOutside` hooks (`@since 17.2.0`) were missing from `GridSettings`, and `BasePlugin.addHook` was missing an explicit string-fallback overload declaration (only the typed overload was visible to callers).

**Not a breaking change**: type-level tightening of `Events` lookups only; the overload ladder keeps every previously-compiling `addHook` call compiling, and hook callbacks passed through the settings object were already strictly typed.

**Follow-up (out of scope here):** the strict rung exposed ~90 internal callback-annotation mismatches across 32 files, including a few genuinely wrong declarations (e.g. `afterContextMenuDefaultOptions` is declared to receive an items array but runtime passes `{ items: [...] }`) and suspected latent bugs (e.g. `beforeRowWrap` handlers annotate an object-listener parameter as `boolean`). These compile via the lenient rung and should be reconciled in a dedicated task.

### How has this been tested?

- New type regressions in `src/__tests__/core/settings.types.ts`: `Events['afterChange']` parameter inference, the newly-added hooks' signatures, `@ts-expect-error` rejection of bogus hook names and arbitrary `keyof Events` strings, and one test per overload-ladder rung.
- `npm run test:types` — compiles all 78 `*.types.ts` files against the rebuilt, downleveled `tmp/*.d.ts` (validates the published shape, not just source).
- Scratch verification against the emitted types: an `Events['afterChange']` callback with zero explicit annotations now infers `CellChange[] | null` / `ChangeSource` (fails with implicit-`any` errors before this change); proved the old and new `HookKey` matchers select the identical key set via a mutual-extension assertion.
- `npm run test:unit -- --testPathPattern=hooks` — 53/53 pass (hooks engine, bucket, mixins).
- React wrapper suite (consumes the built `tmp/` types): 75/75 pass.
- ESLint clean on all changed files.

### Types of changes

### Related issue(s):

1. DEV-2072
2. #13072 (the docs issue whose investigation uncovered this)

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/86caqyk2z

---

> [!NOTE]
> **Low Risk**
> Type-only tightening with backward-compatible addHook overloads; runtime hook dispatch unchanged aside from documented fastCall cast.
> 
> **Overview**
> Fixes **`Events`** so hook lookups like `Events['afterChange']` infer real callback parameters instead of collapsing to **`any`**, by deriving **`HookKey`** from **`RemoveIndexSignature<GridSettings>`** rather than raw **`GridSettings`** (whose `[key: string]: any` index signature widened **`keyof`** to **`string`**).
> 
> **`HookCallback`** becomes `(...args: never[]) => unknown` as a sound function top type; dynamic invocation is isolated to a single cast in **`fastCall`**. **`HotInstance`**, **`BasePlugin`**, and related APIs gain a documented three-rung **`addHook`** / **`addHookOnce`** / **`removeHook`** overload ladder (strict **`Events[K]`**, lenient **`HookCallback`**, string fallback) so tightening types does not break existing listeners.
> 
> Adds missing **`afterOnCellMouseOverOutside`** / **`beforeOnCellMouseOverOutside`** entries on **`GridSettings`**, plus DEV-2072 type regressions in **`settings.types.ts`**. No runtime behavior changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4b3c89dc0101db936924a5c247c30043b4b47a79. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changelogs/13079.json
+++ b/.changelogs/13079.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed the `Events` type resolving every hook callback to `any`, so callbacks typed through `Events[hookName]` and `addHook` now get correct parameter types and autocomplete.",
+  "type": "fixed",
+  "issueOrPR": 13079,
+  "breaking": false,
+  "framework": "none"
+}
--- a/handsontable/src/__tests__/core/settings.types.ts
+++ b/handsontable/src/__tests__/core/settings.types.ts
@@ -1,6 +1,8 @@
 import type Handsontable from 'handsontable';
 import HyperFormula from 'hyperformula';
-import type { CellProperties, CellCoords, RangeType, RemoveIndexSignature } from 'handsontable';
+import type {
+  CellProperties, CellCoords, RangeType, RemoveIndexSignature, Events, HotInstance, CellChange, ChangeSource,
+} from 'handsontable';
 
 // Helpers to verify multiple different settings and prevent TS control-flow from eliminating unreachable values
 declare function oneOf<T extends Array<string | number | boolean | undefined | null | object>>(...args: T): T[number];
@@ -463,6 +465,7 @@ const allSettings: Required<Handsontable.GridSettings> = {
   afterOnCellCornerMouseDown: (event) => {},
   afterOnCellMouseDown: (event, coords, TD) => {},
   afterOnCellMouseOver: (event, coords, TD) => {},
+  afterOnCellMouseOverOutside: (event, coords, TD) => {},
   afterOnCellMouseOut: (event, coords, TD) => {},
   afterOnCellMouseUp: (event, coords, TD) => {},
   afterPageChange(oldPage, newPage) {
@@ -640,6 +643,7 @@ const allSettings: Required<Handsontable.GridSettings> = {
   beforeOnCellMouseDown: (event, coords, TD, controller) => {},
   beforeOnCellMouseOut: (event, coords, TD) => {},
   beforeOnCellMouseOver: (event, coords, TD, controller) => {},
+  beforeOnCellMouseOverOutside: (event, coords, TD, controller) => {},
   beforeOnCellMouseUp: (event, coords, TD) => {},
   beforeDataProviderFetch: queryParameters => true,
   beforePageChange(oldPage, newPage) {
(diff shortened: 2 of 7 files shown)


qunabu:
Task linked: [DEV-2072 Core: `Events` type resolves to `any` (GridSettings index signature collapses HookKey)](https://app.clickup.com/t/9015210959/DEV-2072)

budnix (approved):

PR #13078, opened 2026-07-15
DEV-2079: Skip row-height cache rebuilds for unchanged oversized rows, add sparse PositionCache mode

budnix:
### Context

The PR fixes a rendering-performance defect where the row-height position cache (`PositionCache`) was invalidated and fully rebuilt — an O(totalRows) prefix-sum loop plus a `Float64Array(totalRows + 1)` allocation (8 MB at 1M rows) — on every full draw whenever any wrapped/tall row was in the viewport with `autoRowSize` off (the default).

Three causes, all addressed in `walkontable/src/axisSizing`:

1. **Reset/mark churn.** `resetOversizedRows` wiped the rendered band's records before every render; `markOversizedRows` then re-detected the same unchanged tall rows as "changes" and invalidated the cache. Now the wiped values are threaded from reset to mark, and the cache is invalidated only when a measured height genuinely differs — including rows that shrank back to their provided height, which previously left the cache stale with **no** invalidation at all (latent bug).
2. **Uniform mode lost forever.** Reset assigned `undefined` instead of deleting keys, so the uniform fast-path predicate (`Object.keys(oversizedRows).length === 0`) never held again after the first oversized row. Records are now deleted.
3. **The ±1px boundary flip.** The band's first `<tr>` measures 1px taller (`tr:first-child` border compensation), so every band shift re-recorded one boundary row as "changed". Change detection now tolerates the 1px flip; `Viewport#sumRowHeights` already re-reads boundary rows live for exactly this reason.

On top of that, `PositionCache` gained a **sparse mode** (`sparseExceptionsFn`): when the row size source is uniform but oversized rows were measured, the cache stores only the exceptions (sorted indexes + delta prefix sums) and computes offsets arithmetically. Rebuild cost drops from O(totalRows) to O(oversized rows), so scrolling into unmeasured territory (where a rebuild is genuinely needed) no longer stutters. Grids with per-row heights or `modifyRowHeight` consumers keep the full prefix-sum path unchanged.

Two related structural changes:

- The `oversizedRows` `Table` mixin was converted to free functions in the `axisSizing` slice (vertical-slices architecture), with `drawCycle.ts`, `overlays.ts`, and `baseTable.ts` updated.
- `PositionCache` was split into a slim lifecycle facade plus three per-mode strategy modules (`axisSizing/positionCache/`: uniform, sparse, prefix-sum) — one strategy object is chosen per build, so no read method branches on the cache mode. Public API and behavior are unchanged.

[...]

PR #13080, opened 2026-07-15
DEV-2080: Add skipUnchangedWrites option to IndexMap for scalar-valued maps

budnix:
### Context

The PR adds a `skipUnchangedWrites` option to `IndexMap`: when enabled, `setValueAtIndex` with a strictly-equal (`===`) value is a no-op — nothing is stored and the `change` hook does not run. Every `change` event on a registered map invalidates consumer caches (an unbatched write triggers a full O(n) `IndexMapper` cache rebuild), so re-writing an unchanged value paid a large, avoidable cost.

The option is an **explicit opt-in**, not a base-class default, because reference equality is not a safe "unchanged" signal for object-valued maps: `LinkedPhysicalIndexToValueMap` consumers store mutable state objects (Filters' `filteringStates`, ColumnSorting's `sortingStates`) and the codebase mutates stored values by reference (`conditionCollection.ts`, "Add next condition ... (by reference)"). Those maps keep today's semantics.

Where it is enabled:

- **Always-on for value-enforced scalar classes** — their write path coerces values, so an unchanged write is provably a no-op:
  - `HidingMap`, `TrimmingMap` (flags coerced to booleans in `BooleanMap`),
  - `IndexesSequence` (physical indexes) — where the check runs before materialization, so re-writing an identity value also preserves the compact identity representation.
- **Per-instance for the numeric width/height maps**: AutoColumnSize `columnWidthsMap`, AutoRowSize `rowHeightsMap`, ManualColumnResize, ManualRowResize. For AutoColumnSize this stops every render of a wide grid from rebuilding the column-width position cache (each render re-measures visible columns and rewrote unchanged widths; each write's `change` event invalidated the cache — ~5 ms rebuild per render at 5,000 columns).

`IndexMapper.updateCache` is already dirty-flag aware, so a `batchExecution` whose writes were all skipped no longer rebuilds at `resumeOperations`.

Measured (200k rows; before = develop-equivalent, after = this PR; `demo-skipwrites-bench.html`):

| Scenario | before | after |
|---|---|---|
| 200 unbatched no-op `setValueAtIndex` on a registered `HidingMap` | 1.58 s (each write = full O(n) cache rebuild) | 0.2 ms 🟢  |
| `hideRows` repeat with the same 20k rows (batched no-ops) | 389 ms | 378 ms (the one batch rebuild skipped; the rest is per-row translation + render) |
| `trimRows` repeat with the same 10k rows | 363 ms | 353 ms |
| Sort / filter (bulk `setValues` paths — out of scope) | – | parity |

Correctness identical across bundles (hidden/trimmed/filtered counts match exactly).

With `autoRowSize: true` + default `autoColumnSize` (20k rows × 60 columns), where every render re-measures sizes through the ghost tables:

| Scenario | before | this PR | |
|---|---|---|---|
| `hot.render()` × 20 (median) | 96.0 ms | 91.5 ms | −4.7% 🟢 |
| Cell edit (commit → full render, median of 10) | 105.5 ms | 99.2 ms | −6.0% 🟢 |

The saving is the skipped invalidation tail; the dominant remaining cost is the ghost-table DOM re-measurement itself, which is out of scope here (follow-up candidate).

**Factory conversion.** `createIndexMap` / `createAndRegisterIndexMap` now accept `IndexMapOptions` (plus a new `'indexesSequence'` registry type), and 12 plugin call sites moved from imperative `new *Map()` + `registerMap()` to the factory (autoColumnSize, autoRowSize, manualColumnResize, manualRowResize, hiddenRows, hiddenColumns, trimRows, nestedRows, filters ×2, columnSorting ×2). Plugins that attach an `'init'` local hook replay the handler after factory registration, because `registerMap` initializes the map synchronously on a loaded grid (a plugin re-enable) — before the hook could attach. Still imperative by design: `conditionCollection`'s deliberately unregistered standalone map, `bindRowsWithHeaders`' custom `StrictBindsMap` (not a registry type), and the mapper's own internal `indexesSequence`.

Related: DEV-2079 / #13078 (stacked work in the same area; if #13078 merges first, its temporary AutoColumnSize call-site guard becomes redundant and this PR's map option replaces it — the guard hunk can be dropped in whichever lands second). Possible follow-ups: the same no-op detection for `setValues` (would cover the sorting/filtering bulk paths) and an options parameter for the `createAndRegisterIndexMap` factory (StretchColumns).

### How has this been tested?

- 5 new unit tests for the option contract in `physicalIndexToValueMap.unit.ts` (skip on equal value, write on different value, initial-value no-op, out-of-range unchanged, default-off behavior preserved).
- Updated the map-contract tests that documented the old behavior (each carried a "`change` hook should be called?" comment questioning it): `hidingMap.unit.ts`, `trimmingMap.unit.ts`, `IndexesSequence.unit.ts`, and the two `indexMapper.unit.ts` cache tests — no-op writes now assert the caches stay valid by reference.
- `npm run test:unit --prefix handsontable -- --testPathPattern=translations` — 230 passing.
- Targeted E2E over every map-consuming plugin (`hiddenRows|hiddenColumns|trimRows|pagination|collapsibleColumns|nestedRows|filters|columnSorting|manualColumnMove|manualRowMove|autoColumnSize|autoRowSize|manualColumnResize|manualRowResize`) — 2238 specs, 0 failures.
- Full unit suite and smoke E2E re-run after rebasing onto develop — green.

### Types of changes

### Related issue(s):

1. DEV-2080

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2080

---

> [!NOTE]
> **Medium Risk**
> Touches core index translation and many plugins; behavior change is intentional for scalar maps but linked object maps keep old semantics to avoid breaking in-place mutations.
> 
> **Overview**
> Adds **`skipUnchangedWrites`** on `IndexMap`: when enabled, `setValueAtIndex` with a `===`-equal value does not store and does not fire `change`, so `IndexMapper` caches are not rebuilt on no-op updates.
> 
> **Always on** for scalar/boolean maps (`HidingMap`, `TrimmingMap`, `IndexesSequence`) and in **`BooleanMap`** for compact all-default writes. **Opt-in** via `createAndRegisterIndexMap(..., { skipUnchangedWrites: true })` for auto/manual row and column size maps. **Not** enabled for `LinkedPhysicalIndexToValueMap` (object state mutated by reference).
> 
> Plugins that used `new *Map()` + `registerMap()` now use **`createAndRegisterIndexMap`** (new `'indexesSequence'` type). Hidden/trim/manual resize plugins **replay `#onMapInit`** when the grid is already loaded on re-enable, because factory registration can initialize the map before the `init` hook attaches.
> 
> Unit tests now expect no `change` and stable cache references on unchanged trimming/hiding writes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2554dcece13ebdd0727d0674a9e56f417d125454. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changelogs/13080.json
+++ b/.changelogs/13080.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Improved index-map performance: writing an unchanged value into scalar-valued index maps (hiding, trimming, order sequence, and the automatic or manual row/column size maps) no longer rebuilds the index caches.",
+  "type": "changed",
+  "issueOrPR": 13080,
+  "breaking": false,
+  "framework": "none"
+}
--- a/handsontable/src/plugins/autoColumnSize/autoColumnSize.ts
+++ b/handsontable/src/plugins/autoColumnSize/autoColumnSize.ts
@@ -8,7 +8,7 @@ import { valueAccordingPercent, rangeEach } from '../../helpers/number';
 import SamplesGenerator from '../../utils/samplesGenerator';
 import { isPercentValue } from '../../helpers/string';
 import { DEFAULT_COLUMN_WIDTH } from '../../3rdparty/walkontable/src';
-import { PhysicalIndexToValueMap as IndexToValueMap } from '../../translations';
+import type { PhysicalIndexToValueMap as IndexToValueMap } from '../../translations';
 
 Hooks.getSingleton().register('modifyAutoColumnSizeSeed');
 
@@ -285,7 +285,7 @@ export class AutoColumnSize extends BasePlugin {
    * @private
    * @type {PhysicalIndexToValueMap}
    */
-  columnWidthsMap = new IndexToValueMap(null, { skipUnchangedWrites: true });
+  columnWidthsMap: IndexToValueMap;
   /**
    * `true` value indicates that the #onInit() function has been already called.
    *
@@ -317,7 +317,11 @@ export class AutoColumnSize extends BasePlugin {
    */
   constructor(hotInstance: HotInstance) {
     super(hotInstance);
-    this.hot.columnIndexMapper.registerMap(COLUMN_SIZE_MAP_NAME, this.columnWidthsMap);
+    // The map holds numbers only, so re-writing an unchanged width is a no-op that must not
+    // invalidate the column-width position cache (every render re-measures the visible columns).
+    this.columnWidthsMap = this.hot.columnIndexMapper.createAndRegisterIndexMap(
+      COLUMN_SIZE_MAP_NAME, 'physicalIndexToValue', null, { skipUnchangedWrites: true },
+    );
 
     // Leave the listener active to allow auto-sizing the columns when the plugin is disabled.
     // This is necessary for width recalculation for resize handler doubleclick (ManualColumnResize).
--- a/handsontable/src/plugins/autoColumnSize/autoColumnSize.ts
+++ b/handsontable/src/plugins/autoColumnSize/autoColumnSize.ts
@@ -285,7 +285,7 @@ export class AutoColumnSize extends BasePlugin {
    * @private
    * @type {PhysicalIndexToValueMap}
    */
-  columnWidthsMap = new IndexToValueMap();
+  columnWidthsMap = new IndexToValueMap(null, { skipUnchangedWrites: true });
   /**
    * `true` value indicates that the #onInit() function has been already called.
    *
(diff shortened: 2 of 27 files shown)


qunabu:
Task linked: [DEV-2079 Skip row-height cache rebuilds for unchanged oversized rows + sparse PositionCache mode](https://app.clickup.com/t/9015210959/DEV-2079)
Task linked: [DEV-2080 Add skipUnchangedWrites option to IndexMap for scalar-valued maps](https://app.clickup.com/t/9015210959/DEV-2080)

demtario (approved):

PR #13104, opened 2026-07-21
DEV-2114: Fix grid height collapse when width set and height omitted

adrianspdev:
### Context

The PR fixes a grid that collapses to zero height when `width` is set (for example `width: '100%'`) and `height` is omitted — a v18 regression versus 17.1.

**Root cause.** DEV-1025 added `overflow-x: clip` on the root wrapper for width-constrained, height-less tables. `getTrimmingContainer()` treats `clip` as a scroll/trimming container, so it selected the clipped wrapper instead of the window. The table then dropped out of window-scroll mode: the master `wtHolder` was pinned to `0px` (content rendered but clipped to an invisible grid), frozen rows stopped pinning, and vertical virtualization stopped.

**Fixes in this PR:**

1. `getTrimmingContainer()` no longer treats a single-axis `overflow: clip` (with the perpendicular axis `visible`) as a trimming container — such an element establishes no scroll port and the table still scrolls with the window on the visible axis. Extracted an `overflowAxisTraps()` helper.
2. The same rule is applied to the inline `overflow` shorthand (`overflow: clip visible`), not only to individual/computed properties — via `resolveOverflowAxes()`, which also handles engines (jsdom) that neither split the shorthand nor reflect it in computed style.
3. `overflow-x: clip` is applied only for **definite** widths. Relative widths (`100%`, percentages, viewport units, or a `calc()` mixing them in) keep scrolling horizontally with the window (page scrollbar, all columns reachable) — matching 17.1. Definite lengths (`px`, `em`, `rem`, …) clip to their box.

Remaining, out of scope: a fixed **pixel** width narrower than its columns with no `height` still lacks an in-box horizontal scrollbar (columns clipped). Tracked in DEV-2116.

### How has this been tested?

Added E2E regression guards in `settings/width.spec.js` and unit tests in `helpers/dom/__tests__/element.unit.ts` (single-axis clip, inline shorthand, percentage / `calc()` / `em` width handling, vertical non-collapse, preserved window-scroll).

```bash
npm run test:e2e --prefix handsontable -- --testPathPattern="settings/width"                    # 31 specs, 0 failures
npm run test:unit --prefix handsontable -- --testPathPattern="dom/__tests__/element.unit"        # 97 tests, 0 failures
npm run test:walkontable --prefix handsontable                                                   # 808 specs, 0 failures
npm run test:e2e --prefix handsontable -- --testPathPattern="Core_scroll|fixedRows|fixedColumns|settings/height|Core_init"  # 0 failures
```

Manual: `width: '100%'` (and wider content) with no `height` renders at content height and matches released 17.1 — auto height, virtualized rows, frozen rows pinning, and a page horizontal scrollbar when content is wider than the viewport.

### Types of changes

### Related issue(s):

1. DEV-2114
2. Follow-up: DEV-2116 (in-box horizontal scroll for fixed-px widths narrower than content)

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2114

---

> [!NOTE]
> **Medium Risk**
> Changes core layout/scroll detection (`getTrimmingContainer` and root overflow rules) used by Walkontable overlays and virtualization; behavior differs by width type but is covered by new regression tests.
> 
> **Overview**
> Fixes a v18 regression where a grid with **`width` set and no `height`** collapsed to zero height (invisible table) while breaking window vertical scroll, frozen rows, and row virtualization.
> 
> **`getTrimmingContainer`** no longer treats an ancestor as the trimming container when only one axis uses **`overflow: clip`** and the other stays **`visible`** (including two-value shorthands like `clip visible`). Horizontal-only clip on the root no longer steals trimming from the window. Overflow is resolved via new **`overflowAxisTraps`** / **`resolveOverflowAxes`** helpers (inline shorthand, jsdom, non-concrete keywords like `inherit`).
> 
> **Width / overflow in `core`:** **`overflow-x: clip`** is applied only for **definite** widths (`px`, `em`, etc.). **Relative** widths (`%`, viewport units, `calc()` containing them) skip clip so wide content scrolls with the page like 17.1.
> 
> E2E and DOM helper unit tests cover non-collapse, window scroll, and width-type behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7dd0b4078ae74d92604e4943126db9d8550616f1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/handsontable/src/__tests__/settings/width.spec.js
+++ b/handsontable/src/__tests__/settings/width.spec.js
@@ -473,6 +473,22 @@ describe('settings', () => {
         expect(holder.getBoundingClientRect().height).toBeGreaterThan(0);
         expect(hot.view.isVerticallyScrollableByWindow()).toBe(true);
       });
+
+      it('should not clip horizontally when `width` is a percentage and `height` is omitted', async() => {
+        // A percentage width fills its container. Only a definite pixel width is clipped; a relative
+        // width lets content wider than the container scroll with the window (the page gains a
+        // horizontal scrollbar), so every column stays reachable instead of being clipped away.
+        const hot = handsontable({
+          data: createSpreadsheetData(8, 20),
+          rowHeaders: true,
+          colHeaders: true,
+          colWidths: 150,
+          width: '100%',
+        });
+
+        expect(window.getComputedStyle(hot.rootElement).overflowX).not.toBe('clip');
+        expect(hot.view.isHorizontallyScrollableByWindow()).toBe(true);
+      });
     });
   });
 });
--- a/handsontable/src/__tests__/settings/width.spec.js
+++ b/handsontable/src/__tests__/settings/width.spec.js
@@ -410,6 +410,46 @@ describe('settings', () => {
         expect(hot.rootElement.style.overflow).toBe('hidden');
         expect(hot.rootElement.style.overflowX).toBe('hidden');
       });
+
+      it('should not vertically collapse the grid when `width` is set and `height` is omitted', async() => {
+        // Regression guard: the width-clip block sets `overflow-x: clip` on the root. The trimming
+        // container lookup must not treat that horizontal-only clip as a vertical trimming
+        // container, or the master holder is pinned to `0px` and the fully rendered rows are
+        // clipped to an invisible grid (the table's content height is > 0 but nothing shows).
+        const hot = handsontable({
+          data: createSpreadsheetData(8, 6),
+          rowHeaders: true,
+          colHeaders: true,
+          width: '100%',
+        });
+
+        const holder = hot.rootElement.querySelector('.ht_master .wtHolder');
+        const table = hot.rootElement.querySelector('.ht_master .htCore');
+
+        // The holder must size to its content, not collapse to zero.
+        expect(holder.getBoundingClientRect().height).toBeGreaterThan(0);
+        expect(holder.getBoundingClientRect().height)
+          .toBeAroundValue(table.getBoundingClientRect().height, 2);
+      });
+
+      it('should stay window-scrollable when `width` is set and `height` is omitted', async() => {
+        // The horizontal-only `overflow-x: clip` must leave the vertical axis scrolling with the
+        // window. If the clipped root is picked as the trimming container, the overlays drop out of
+        // window-scroll mode: frozen rows stop pinning and vertical virtualization stops (every row
+        // renders into the DOM).
+        const hot = handsontable({
+          data: createSpreadsheetData(200, 6),
+          rowHeaders: true,
+          colHeaders: true,
+          fixedRowsTop: 2,
+          width: '100%',
+        });
+
+        expect(hot.view.isVerticallyScrollableByWindow()).toBe(true);
+        // Window-scroll viewport virtualizes: only a slice of the 200 rows is rendered.
+        expect(hot.rootElement.querySelectorAll('.ht_master .htCore tbody tr').length)
+          .toBeLessThan(200);
+      });
     });
   });
 });
--- a/handsontable/src/__tests__/settings/width.spec.js
+++ b/handsontable/src/__tests__/settings/width.spec.js
@@ -504,6 +504,20 @@ describe('settings', () => {
         expect(window.getComputedStyle(hot.rootElement).overflowX).not.toBe('clip');
       });
 
+      it('should not clip when `width` is a viewport unit (`vw`) and `height` is omitted', async() => {
+        // Viewport units resolve against the viewport, not a fixed box, so they are relative — the
+        // unit is preceded by digits (`100vw`), which must still be detected as relative (no clip).
+        const hot = handsontable({
+          data: createSpreadsheetData(8, 10),
+          rowHeaders: true,
+          colHeaders: true,
+          colWidths: 150,
+          width: '100vw',
+        });
+
+        expect(window.getComputedStyle(hot.rootElement).overflowX).not.toBe('clip');
+      });
+
       it('should clip when `width` is a definite non-pixel length (`em`) and `height` is omitted', async() => {
         // Absolute lengths other than `px` (em, rem, etc.) still establish a fixed box, so the table
         // must not visually overflow it — clip applies.
--- a/handsontable/src/__tests__/settings/width.spec.js
+++ b/handsontable/src/__tests__/settings/width.spec.js
@@ -489,6 +489,34 @@ describe('settings', () => {
         expect(window.getComputedStyle(hot.rootElement).overflowX).not.toBe('clip');
         expect(hot.view.isHorizontallyScrollableByWindow()).toBe(true);
       });
+
+      it('should not clip when `width` is a `calc()` that mixes in a percentage and `height` is omitted', async() => {
+        // A `calc()` referencing a percentage is container-driven even though it ends in `px`, so it
+        // must be treated as relative (no clip) rather than as a definite pixel width.
+        const hot = handsontable({
+          data: createSpreadsheetData(8, 10),
+          rowHeaders: true,
+          colHeaders: true,
+          colWidths: 150,
+          width: 'calc(100% - 20px)',
+        });
+
+        expect(window.getComputedStyle(hot.rootElement).overflowX).not.toBe('clip');
+      });
+
+      it('should clip when `width` is a definite non-pixel length (`em`) and `height` is omitted', async() => {
+        // Absolute lengths other than `px` (em, rem, etc.) still establish a fixed box, so the table
+        // must not visually overflow it — clip applies.
+        const hot = handsontable({
+          data: createSpreadsheetData(8, 10),
+          rowHeaders: true,
+          colHeaders: true,
+          colWidths: 150,
+          width: '20em',
+        });
+
+        expect(window.getComputedStyle(hot.rootElement).overflowX).toBe('clip');
+      });
     });
   });
 });
--- a/handsontable/src/__tests__/settings/width.spec.js
+++ b/handsontable/src/__tests__/settings/width.spec.js
@@ -450,6 +450,29 @@ describe('settings', () => {
         expect(hot.rootElement.querySelectorAll('.ht_master .htCore tbody tr').length)
           .toBeLessThan(200);
       });
+
+      it('should render vertically when `width` is narrower than the columns and `height` is omitted', async() => {
+        // A width-constrained grid whose columns are wider than the width renders at content height
+        // and scrolls vertically with the window (previously the whole grid collapsed to `0px`).
+        // Known limitation: the columns past the constrained width are clipped by the root's
+        // `overflow-x: clip` and are not reachable via a horizontal scrollbar — reaching them needs
+        // per-axis trimming (window vertical + element horizontal), tracked in a follow-up task.
+        const hot = handsontable({
+          data: createSpreadsheetData(8, 10),
+          rowHeaders: true,
+          colHeaders: true,
+          colWidths: 150,
+          width: 300,
+        });
+
+        const holder = hot.rootElement.querySelector('.ht_master .wtHolder');
+
+        expect($(hot.rootElement).width()).toBeAroundValue(300, 1);
+        expect(window.getComputedStyle(hot.rootElement).overflowX).toBe('clip');
+        // Not collapsed: the grid is visible and sizes vertically to its content.
+        expect(holder.getBoundingClientRect().height).toBeGreaterThan(0);
+        expect(hot.view.isVerticallyScrollableByWindow()).toBe(true);
+      });
     });
   });
 });
--- a/handsontable/src/core.ts
+++ b/handsontable/src/core.ts
@@ -3515,15 +3515,21 @@ export default function Core(
     // width blocks ran) so partial updateSettings calls see the correct state.
     // When height IS set, the height block's `overflow: clip` shorthand handles both axes — leave
     // overflowX untouched to avoid breaking that shorthand.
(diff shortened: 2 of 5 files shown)


qunabu:
Task linked: [DEV-2114 Handsontable v18: grid collapses with width: '100%' and no height](https://app.clickup.com/t/9015210959/DEV-2114)

budnix (commented):
A few notes on the single-axis clip change, mostly on the width case and test coverage.

handsontable/src/helpers/dom/element.ts
```diff
@@ -864,9 +864,18 @@ export function getTrimmingContainer(base: HTMLElement): HTMLElement | Window {
...
+      // with the window on the visible axis. A width-constrained, window-scrolled table sets
+      // `overflow-x: clip` on its root (see core.ts, DEV-1025); treating that root as the trimming
+      // container drops the overlays out of window-scroll mode (frozen rows stop pinning, vertical
+      // virtualization stops). Such a single-axis clip must not qualify the element as trimming.
+      const clipsHorizontallyOnly = propertyX === 'clip' && (propertyY === 'visible' || propertyY === '');
+      const clipsVerticallyOnly = propertyY === 'clip' && (propertyX === 'visible' || propertyX === '');
+      const trimsX = allowedProperties.includes(propertyX) && !clipsHorizontallyOnly;
+      const trimsY = allowedProperties.includes(propertyY) && !clipsVerticallyOnly;
+
+      if (allowedProperties.includes(property) || trimsX || trimsY) {
```
budnix: Curious whether the narrow-grid case was checked here. When the width-clip root gets skipped, the trimming container falls through to the window, and in window mode `measureWorkspaceWidth` sizes the workspace to the viewport width once the columns are wider than the holder — not to the grid width. So a grid at `width: 300px` with columns totaling more than 300px would render columns past its edge, clipped and unreachable. Both new tests use `width: '100%'` where grid width equals the viewport, so this path isn't covered — worth a quick check with a narrow grid and wide columns.

handsontable/src/helpers/dom/element.ts
```diff
@@ -864,9 +864,18 @@ export function getTrimmingContainer(base: HTMLElement): HTMLElement | Window {
...
-      if (allowedProperties.includes(property) ||
-          allowedProperties.includes(propertyY) ||
-          allowedProperties.includes(propertyX)) {
+      // `overflow: clip` establishes no scroll port. When an element clips a single axis while the
+      // other axis stays `visible`, it does not trap the table's scroll — the table still scrolls
+      // with the window on the visible axis. A width-constrained, window-scrolled table sets
+      // `overflow-x: clip` on its root (see core.ts, DEV-1025); treating that root as the trimming
+      // container drops the overlays out of window-scroll mode (frozen rows stop pinning, vertical
+      // virtualization stops). Such a single-axis clip must not qualify the element as trimming.
+      const clipsHorizontallyOnly = propertyX === 'clip' && (propertyY === 'visible' || propertyY === '');
```
budnix: May I ask — this skips any ancestor that clips one axis and leaves the other visible, not just Handsontable's own root. If someone wraps the grid in an element with `overflow-x: clip; overflow-y: visible`, that wrapper used to be the trimming container, and now it's skipped, so the grid switches to window-scroll mode for that layout. That may well be what you want, since the visible axis was never really constrained — just flagging it so it's a conscious choice and not a surprise for existing pages.

handsontable/src/helpers/dom/element.ts
```diff
@@ -864,9 +864,18 @@ export function getTrimmingContainer(base: HTMLElement): HTMLElement | Window {
...
+      const clipsVerticallyOnly = propertyY === 'clip' && (propertyX === 'visible' || propertyX === '');
+      const trimsX = allowedProperties.includes(propertyX) && !clipsHorizontallyOnly;
+      const trimsY = allowedProperties.includes(propertyY) && !clipsVerticallyOnly;
```
budnix: Heads up — the four new boolean locals push this function's cognitive complexity from 11 to 17, over the repo's limit of 15, so SonarCloud will fail the quality gate. Pulling the per-axis test into a small helper outside the function moves the `&&`/`||` chains out and brings it back under, e.g. `trimsAxis(axis, other, allowed)` returning `allowed.includes(axis) && !(axis === 'clip' && (other === 'visible' || other === ''))`. Reads a bit cleaner too.

handsontable/src/__tests__/settings/width.spec.js
```diff
@@ -410,6 +410,46 @@ describe('settings', () => {
         expect(hot.rootElement.style.overflow).toBe('hidden');
         expect(hot.rootElement.style.overflowX).toBe('hidden');
       });
+
+      it('should not vertically collapse the grid when `width` is set and `height` is omitted', async() => {
```
budnix, line 414: These two regression guards are E2E only. The project checklist asks every change to add a Jest unit test too, and `getTrimmingContainer` is a pure DOM helper — you can set `overflow-x`/`overflow-y` inline and read them back in jsdom, so the single-axis-clip branch is easy to cover with a fast unit test. Would be good to add one so a future edit to this logic can't regress it silently.

handsontable/src/helpers/dom/element.ts
adrianspdev: Good catch — checked it, with a build for each side. Two findings:

1. On `develop` **before this fix**, the same narrow case (`width: 300` + columns wider than 300 + no `height`) collapses **entirely** — the master holder is pinned to `0px`, so the whole grid is invisible, not just the off-width columns. This fix restores vertical rendering (auto height, window vertical scroll).
2. The columns past the constrained width are still clipped by the root's `overflow-x: clip` with no horizontal scrollbar. Reaching them needs **per-axis trimming** (vertical → window for auto-height/virtualization, horizontal → element scroll within the width), which is a `masterTable` + overlay sizing rework beyond this bug's scope.

So this change is a strict improvement for the narrow case (invisible → visible + vertically correct), not a regression. Added a test documenting the current behavior (`should render vertically when width is narrower than the columns and height is omitted`) and will file a follow-up for full horizontal reachability.

handsontable/src/helpers/dom/element.ts
adrianspdev: Intentional — thanks for flagging, confirming it's a conscious choice. Any ancestor with `overflow-x: clip; overflow-y: visible` (or the vertical mirror) now resolves to window-scroll on the visible axis. That's correct by design: a single-axis `clip` establishes no scroll port, and the visible axis was never constrained, so the table genuinely scrolls with the window there. Only the non-scrolling `clip` value is exempted — real `auto`/`scroll`/`hidden` containers and both-axis `clip` still trim exactly as before (now covered by unit tests).

handsontable/src/helpers/dom/element.ts
adrianspdev: Done — extracted a module-level `overflowAxisTraps(axis, perpendicular)` helper as suggested, so `getTrimmingContainer` is back to three conditions and the `&&`/`||` chains live in the helper. ESLint passes locally.

handsontable/src/__tests__/settings/width.spec.js
adrianspdev: Done — added `getTrimmingContainer` unit tests in `src/helpers/dom/__tests__/element.unit.ts` (7 cases): single-axis `overflow-x: clip` / `overflow-y: clip` → window; real `auto`/`scroll` single-axis + both-axis `clip` + `hidden` → the element; no overflow → window.

handsontable/src/helpers/dom/element.ts
adrianspdev: Follow-up for full horizontal reachability filed: DEV-2116 — https://app.clickup.com/t/86cauqy4q

handsontable/src/helpers/dom/element.ts
```diff
@@ -859,14 +880,13 @@ export function getTrimmingContainer(base: HTMLElement): HTMLElement | Window {
...
     if (rootWindow) {
       const computedStyle = rootWindow.getComputedStyle(el);
-      const allowedProperties = ['scroll', 'hidden', 'auto', 'clip'];
       const property = computedStyle.getPropertyValue('overflow');
       const propertyY = computedStyle.getPropertyValue('overflow-y');
       const propertyX = computedStyle.getPropertyValue('overflow-x');
 
-      if (allowedProperties.includes(property) ||
-          allowedProperties.includes(propertyY) ||
-          allowedProperties.includes(propertyX)) {
```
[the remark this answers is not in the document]
adrianspdev: Confirmed and fixed in f7cc4308b1.

Verified: the individual `overflow-x: clip` that Handsontable itself sets (via `element.style.overflowX`, shorthand left empty) never hit the early return, so the reported collapse and HOT's own rendering were unaffected. But you're right that a two-value inline shorthand like `overflow: clip visible` did bypass `overflowAxisTraps` through the early `style.overflow` return and could keep such a wrapper selected as the trimming container.

The inline and computed paths are now unified: `resolveOverflowAxes()` parses the inline `overflow` shorthand into per-axis values (preferring inline over computed), then the same single-axis-clip rule runs for both. Added unit tests for `clip visible` / `visible clip` (→ window) and `clip` / `auto visible` (→ element).

handsontable/src/core.ts
```diff
@@ -3529,8 +3533,7 @@ export default function Core(
         // by `clip`. Unlike `hidden`, `clip` creates no block formatting context and allows no
         // programmatic scroll.
         if (currentOverflowX === '' || currentOverflowX === 'clip') {
-          instance.rootElement.style.overflowX =
-            (effectiveWidth && effectiveWidth !== 'auto') ? 'clip' : '';
+          instance.rootElement.style.overflowX = /px$/.test(effectiveWidth) ? 'clip' : '';
```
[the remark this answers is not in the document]
adrianspdev: Confirmed and fixed. The `/px$/` suffix check was too crude — `calc(100% - 20px)` ends in `px` but is container-driven, and `em`/`rem` are definite yet don't end in `px`.

Replaced it with a relative-marker test: a width is treated as relative (no clip → window scroll) when it contains `%` or a viewport unit (`vw`/`vh`/`vmin`/`vmax`), including inside `calc()`; otherwise it's a definite length (`px`, `em`, `rem`, …) and clips. Added tests for `calc(100% - 20px)` (→ no clip) and `20em` (→ clip).

qunabu:
Task linked: [DEV-2116 Per-axis trimming: fixed-px width narrower than content with no height is incoherent](https://app.clickup.com/t/9015210959/DEV-2116)

handsontable/src/core.ts
```diff
@@ -3515,11 +3515,21 @@ export default function Core(
...
+    // establishes a fixed box the table must not visually overflow, so it is clipped.
     // Browser compatibility: `overflow-x: clip` requires Safari 16+. On Safari 14.1–15.x it silently
     // falls back to `visible` (graceful degradation — pre-existing behavior, not a new regression).
     if (typeof settings.height !== 'undefined' || typeof settings.width !== 'undefined') {
       const effectiveHeight = instance.rootElement.style.height;
       const effectiveWidth = instance.rootElement.style.width;
+      // Relative: percentages and viewport units resolve against an ancestor, so a `%` or `vw`/`vh`/
+      // `vmin`/`vmax` token anywhere (including inside `calc()`) marks the width as container-driven.
+      const isRelativeWidth = /%|\bv(?:w|h|min|max)\b/i.test(effectiveWidth);
+      const isDefiniteWidth = effectiveWidth !== '' && effectiveWidth !== 'auto' && !isRelativeWidth;
```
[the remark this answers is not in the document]
adrianspdev: Confirmed and fixed. The leading `\b` was the bug: in `100vw` the `v` is preceded by a digit (a word character), so there's no word boundary and `\bv` never matched — viewport-unit widths were misclassified as definite and clipped.

Dropped both word boundaries (`/%|v(?:w|h|min|max)/i`). No CSS length unit contains those substrings except the viewport units themselves, so the match is unambiguous; it also picks up the dynamic units (`dvh`/`svh`/`lvh`) via the `vh` match. Added a `width: '100vw'` → no-clip test.

handsontable/src/helpers/dom/element.ts
```diff
@@ -840,6 +840,49 @@ export function getMaximumScrollLeft(element: HTMLElement) {
...
+ * @returns {{ x: string, y: string }}
+ */
+function resolveOverflowAxes(el: HTMLElement, computedStyle: CSSStyleDeclaration): { x: string, y: string } {
+  const shorthand = el.style.overflow.split(/\s+/).filter(Boolean);
+  const [shorthandX = '', shorthandY = shorthandX] = shorthand;
+
+  return {
+    x: el.style.overflowX || shorthandX || computedStyle.getPropertyValue('overflow-x'),
+    y: el.style.overflowY || shorthandY || computedStyle.getPropertyValue('overflow-y'),
+  };
```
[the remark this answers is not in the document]
adrianspdev: Confirmed and fixed. `resolveOverflowAxes` now prefers the inline value only when it's a concrete overflow keyword (`visible`/`clip`/`hidden`/`scroll`/`auto`/`overlay`). A global keyword (`inherit`/`initial`/`revert`/`unset`) — or anything else — falls back to the computed style, which resolves it to the real value (e.g. an inherited `auto`), so such an ancestor is no longer skipped.

The jsdom shorthand workaround is unchanged (`clip`/`visible` are concrete, so `overflow: clip visible` is still handled inline). Added a unit test for an inline global keyword.

budnix (approved):

PR #12011, opened 2026-02-09
TypeScript initial conversion

qunabu:
> **ClickUp**: [PRO-860](https://app.clickup.com/t/9015210959/PRO-860)

## Context

This PR migrates the Handsontable core library from JavaScript to TypeScript. The source files in `handsontable/src/` are converted from `.js` to `.ts` with type annotations. Type declarations previously hand-authored in `handsontable/types/` are now auto-generated by `tsc --emitDeclarationOnly`.

The migration was performed against the `develop` branch (50 commits merged, 127 conflict resolutions).

---

### ADR-1: Incremental TypeScript strictness

**Decision**: `strict: false`, `noImplicitAny: true`, all other strict checks off.

**Rationale**: Enabling full `strict: true` at once would require resolving `strictNullChecks`, `strictFunctionTypes`, `strictPropertyInitialization`, and `noImplicitThis` across ~1000 files simultaneously. The incremental approach gets TypeScript compiling cleanly first. Stricter settings are deferred to follow-up PRs.

**tsconfig.json current vs target:**

| Setting | Current | Target |
|---|---|---|
| `strict` | `false` | `true` |
| `noImplicitAny` | `true` | `true` ✅ |
| `noImplicitThis` | `false` | `true` |
| `strictNullChecks` | off | `true` |
| `strictFunctionTypes` | off | `true` |
| `strictPropertyInitialization` | off | `true` |

---

### ADR-2: Build system — webpack → rspack, babel-loader → builtin:swc-loader

**Decision**: All build scripts now use `rspack` with `builtin:swc-loader` (separate rules for `.js` and `.ts`/`.tsx`).

**Rationale**: The branch predates the `develop` migration to rspack + swc-loader. `babel-loader` is not installed in this package.

**Files changed**: `.config/base.js`, `.config/walkontable.js`, `.config/test-walkontable.js`, `.config/languages-development.js`, `.config/themes-umd-development.js`, `package.json`

---

### ADR-3: CommonJS/ESM compilation — babel → swc-transpile.mjs

**Decision**: `build:commonjs` and `build:es` now use `node scripts/swc-transpile.mjs` (SWC-based). `@babel/cli` is not installed in this package.

**Changes to swc-transpile.mjs**:
- `collectFiles()` now includes `.ts`/`.tsx` files (excluding `.d.ts`)
- Per-file SWC options: TypeScript parser for `.ts`/`.tsx`, ecmascript parser for `.js`
[...]

PR #13105, opened 2026-07-21
DEV-2115: Fix autofill row resolution near vertically merged cells

adrianspdev:
### Context

The PR fixes a regression where the autofill fill handle cannot target rows that fall within a vertical merge located in another column.

`getCellCoordsFromMousePosition` maps a pointer Y coordinate to a row by walking a single reference column (the first visible column) and honoring that column's `rowspan`. When that column holds a vertical merge, `findRowAtY` skips the merged band's rows, collapsing every Y inside the band onto the merge anchor. As a result, dragging the fill handle in a neighboring column cannot stop on a row inside the band — the fill jumps past the whole merge.

Example: with a merge on `A3:A5`, dragging `B3`'s fill handle down one row to `B4` filled nothing and snapped straight to `B6`. In 17.1.0 the same drag filled `B4`.

This is a regression introduced in 18.0.0 by the TypeScript conversion (#12011), which added the pixel-based row lookup in `Autofill#onMouseMove` and the new `cellCoords.ts` helper. Before that, the fill target came from the actual hovered cell reported by Walkontable (`beforeOnCellMouseOver`), which was merge-aware.

The fix resolves the row against a column free of merge distortion, chosen by a new `findRowReferenceColumn`: the first visible column whose scanned cells are neither a vertical-merge anchor (`rowspan > 1`) nor a merge slave (`hidden`). This keeps the row lookup correct whether the merge lives in the dragged column, a neighboring column, or the source column.

### How has this been tested?

- Unit tests for the helper (a vertical merge in a column other than the one under the pointer):

  ```
  npm run test:unit --prefix handsontable --testPathPattern=cellCoords.unit
  ```

- E2E tests for autofill into a merged band belonging to another column, plus the existing autofill / mergeCells / dragToScroll suites (324 specs, 0 failures):

  ```
  npm run test:e2e --prefix handsontable --testPathPattern="mergeCells/__tests__/(mergeCells.spec|autofill)|plugins/autofill/__tests__|plugins/dragToScroll"
  ```

- Manual verification in a browser: dragging `B3`'s fill handle one row into the `A3:A5` merged band now fills `B4` (matching 17.1.0), instead of jumping to `B6`.

### Types of changes

### Related issue(s):

1. DEV-2115

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2115

---

> [!NOTE]
> **Medium Risk**
> Touches shared pointer-to-cell coordinate logic used by autofill and drag-to-scroll across frozen rows, hidden columns, and merges; mitigated by extensive unit/e2e coverage but edge cases where every visible column is vertically merged still fall back to prior behavior.
> 
> **Overview**
> Fixes **autofill** (and other pointer-driven flows) when a **vertical merge lives in another column**: dragging the fill handle no longer skips rows inside that merged band.
> 
> **`getCellCoordsFromMousePosition`** no longer maps Y to a row using only the first visible column. It adds **`findRowReferenceColumn`**, which picks the first column in the visible range whose cells are not vertically merged (via `rowspan > 1` or repeated DOM element identity), then runs **`findRowAtY`** on that column—separately for frozen top, frozen bottom, and scrollable body. Hidden columns with no renderable cells are skipped; **`getCellMetaTransient`** is used for span metadata. Reference-column search is clamped to the rendered viewport so long drags on huge off-screen grids do not scan the full grid.
> 
> Unit and mergeCells/autofill e2e tests cover neighboring-column merges, frozen rows, hidden columns, and anchors above the scan range.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b04629e2c541af48b2647c6e27bbd24670fe6b95. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/handsontable/src/helpers/dom/__tests__/cellCoords.unit.js
+++ b/handsontable/src/helpers/dom/__tests__/cellCoords.unit.js
@@ -399,4 +399,178 @@ describe('getCellCoordsFromMousePosition', () => {
       expect(coords.row).toBe(4);
     });
   });
+
+  describe('reference column selection with a horizontal merge present', () => {
+    // Regression guard: column 1 is a slave of a purely horizontal merge, so its cells report
+    // `hidden: true` while still rendering at their own single-row height. Column 0 holds a
+    // vertical merge (rows 0-2). The row lookup must not reject column 1 just because it is
+    // hidden — otherwise it falls back to the vertically merged column 0 and collapses the row.
+    const rowHeight = 30;
+    const colWidth = 80;
+    const mergeSpan = 3; // column 0 merged across rows 0-2
+
+    /**
+     * Builds a HOT stub: column 0 is a vertical merge (rows 0-2), column 1 is a hidden slave of
+     * a purely horizontal merge (single-row height per row).
+     *
+     * @returns {object} Stubbed HOT instance.
+     */
+    function buildHotWithHorizontalSlave() {
+      return {
+        rootWindow: { innerWidth: 1280, innerHeight: 720 },
+        rootElement: { getBoundingClientRect: () => ({ left: 0, top: 0, right: 160, bottom: 120 }) },
+        isRtl: () => false,
+        hasColHeaders: () => false,
+        hasRowHeaders: () => false,
+        getFirstPartiallyVisibleRow: () => 0,
+        getLastPartiallyVisibleRow: () => 3,
+        getFirstPartiallyVisibleColumn: () => 0,
+        getLastPartiallyVisibleColumn: () => 1,
+        countRows: () => 4,
+        countCols: () => 2,
+        getCell: (row, col) => {
+          const el = document.createElement('td');
+          const isVerticalMerge = col === 0 && row < mergeSpan;
+          const top = isVerticalMerge ? 0 : row * rowHeight;
+          const height = isVerticalMerge ? rowHeight * mergeSpan : rowHeight;
+
+          Object.defineProperty(el, 'offsetHeight', { get: () => height });
+          Object.defineProperty(el, 'offsetWidth', { get: () => colWidth });
+          el.getBoundingClientRect = () => ({
+            top,
+            left: col * colWidth,
+            bottom: top + height,
+            right: (col * colWidth) + colWidth,
+          });
+
+          return el;
+        },
+        getCellMeta: (row, col) => {
+          if (col === 0 && row === 0) {
+            return { rowspan: mergeSpan, colspan: 1 };
+          }
+          // Column 1 is a hidden slave of a horizontal merge, but not vertically spanned.
+          if (col === 1) {
+            return { rowspan: 1, colspan: 1, hidden: true };
+          }
+
+          return { rowspan: 1, colspan: 1 };
+        },
+        _createCellCoords: (row, col) => ({ row, col }),
+        columnIndexMapper: {
+          getVisualFromRenderableIndex: n => n,
+          getNearestNotHiddenIndex: n => n,
+        },
+        rowIndexMapper: {
+          getVisualFromRenderableIndex: n => n,
+          getNearestNotHiddenIndex: n => n,
+          getNotHiddenIndexesLength: () => 4,
+        },
+        view: {
+          isVerticallyScrollableByWindow: () => false,
+          isHorizontallyScrollableByWindow: () => false,
+          getViewportWidth: () => 160,
+          getViewportHeight: () => 120,
+          getColumnHeaderHeight: () => 0,
+          getRowHeaderWidth: () => 0,
+          countNotHiddenFixedColumnsStart: () => 0,
+          countNotHiddenFixedRowsTop: () => 0,
+          countNotHiddenFixedRowsBottom: () => 0,
+        },
+      };
+    }
+
+    it('does not reject a hidden horizontal-merge slave as the row reference', () => {
+      const hot = buildHotWithHorizontalSlave();
+      // Y at the centre of row 2 (60..90 → 75), inside column 0's vertical merge band.
+      // Column 1 (hidden horizontal slave) has real per-row heights, so the row must resolve
+      // to 2 - not collapse onto the merge anchor (row 0) via a fallback to column 0.
+      const coords = getCellCoordsFromMousePosition(hot, 100, 75);
+
+      expect(coords.row).toBe(2);
+    });
+  });
+
+  describe('vertical merge confined to the fixed (frozen) top rows', () => {
+    // Regression guard (DEV-2115 follow-up): a vertical merge that lives only in the frozen top
+    // rows must be detected when resolving a row inside the frozen overlay. The reference column
+    // is chosen per row-scan range, so the fixed-top walk sees the frozen merge even though the
+    // scrollable "partially visible" range starts below it.
+    const rowHeight = 30;
+    const colWidth = 80;
+
+    /**
+     * Builds a HOT stub with 3 fixed top rows; column 0 holds a vertical merge across the frozen
+     * rows 0-1, while the scrollable body (rows 10+) has no merges.
+     *
+     * @returns {object} Stubbed HOT instance.
+     */
+    function buildHotWithFrozenMerge() {
+      return {
+        rootWindow: { innerWidth: 1280, innerHeight: 720 },
+        rootElement: { getBoundingClientRect: () => ({ left: 0, top: 0, right: 400, bottom: 400 }) },
+        isRtl: () => false,
+        hasColHeaders: () => false,
+        hasRowHeaders: () => false,
+        // Scrollable body is scrolled down - the frozen rows (0-2) are outside this range.
+        getFirstPartiallyVisibleRow: () => 10,
+        getLastPartiallyVisibleRow: () => 20,
+        getFirstPartiallyVisibleColumn: () => 0,
+        getLastPartiallyVisibleColumn: () => 4,
+        countRows: () => 30,
+        countCols: () => 5,
+        getCell: (row, col) => {
+          const el = document.createElement('td');
+          const isFrozenMerge = col === 0 && row < 2; // A1:A2 vertical merge in frozen rows
+          const top = isFrozenMerge ? 0 : row * rowHeight;
+          const height = isFrozenMerge ? rowHeight * 2 : rowHeight;
+
+          Object.defineProperty(el, 'offsetHeight', { get: () => height });
+          Object.defineProperty(el, 'offsetWidth', { get: () => colWidth });
+          el.getBoundingClientRect = () => ({
+            top,
+            left: col * colWidth,
+            bottom: top + height,
+            right: (col * colWidth) + colWidth,
+          });
+
+          return el;
+        },
+        getCellMeta: (row, col) => (
+          col === 0 && row === 0 ? { rowspan: 2, colspan: 1 } : { rowspan: 1, colspan: 1 }
+        ),
+        _createCellCoords: (row, col) => ({ row, col }),
+        columnIndexMapper: {
+          getVisualFromRenderableIndex: n => n,
+          getNearestNotHiddenIndex: n => n,
+        },
+        rowIndexMapper: {
+          getVisualFromRenderableIndex: n => n,
+          getNearestNotHiddenIndex: n => n,
+          getNotHiddenIndexesLength: () => 30,
+        },
+        view: {
+          isVerticallyScrollableByWindow: () => false,
+          isHorizontallyScrollableByWindow: () => false,
+          getViewportWidth: () => 400,
+          getViewportHeight: () => 400,
+          getColumnHeaderHeight: () => 0,
+          getRowHeaderWidth: () => 0,
+          countNotHiddenFixedColumnsStart: () => 0,
+          countNotHiddenFixedRowsTop: () => 3,
+          countNotHiddenFixedRowsBottom: () => 0,
+        },
+      };
+    }
+
+    it('resolves a frozen row inside the merge band using a merge-free reference column', () => {
+      const hot = buildHotWithFrozenMerge();
+      // Y at the centre of frozen row 1 (30..60 → 45), inside column 0's frozen merge band.
+      // Pointer over column 1 (x 80-160). Must resolve to row 1, not the merge anchor (row 0).
+      const coords = getCellCoordsFromMousePosition(hot, 100, 45);
+
+      expect(coords.col).toBe(1);
+      expect(coords.row).toBe(1);
+    });
+  });
 });
--- a/handsontable/src/helpers/dom/__tests__/cellCoords.unit.js
+++ b/handsontable/src/helpers/dom/__tests__/cellCoords.unit.js
@@ -3,6 +3,11 @@ import { getCellCoordsFromMousePosition } from 'handsontable/helpers/dom/cellCoo
 /**
  * Builds a minimal HOT instance stub for getCellCoordsFromMousePosition tests.
  *
+ * Merges are described declaratively through `mergedCells`: only the anchor cell reports a
+ * `rowspan`/`colspan` in its meta (matching real Handsontable), and every rendered slave of a
+ * vertical merge resolves to the SAME cached `<td>` element — so a lookup can detect the merge by
+ * element identity even when the anchor sits outside the scanned range.
+ *
  * @param {object} opts
  * @param {boolean} opts.isWindowScrollV  Whether the table scrolls vertically via window.
  * @param {boolean} opts.isWindowScrollH  Whether the table scrolls horizontally via window.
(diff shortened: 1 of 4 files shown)


qunabu:
Task linked: [DEV-2115 Handsontable v18: Autofill and mergeCells - regression](https://app.clickup.com/t/9015210959/DEV-2115)

handsontable/src/helpers/dom/cellCoords.ts
```diff
@@ -214,6 +257,23 @@ export function getCellCoordsFromMousePosition(
...
+  // column collapses every Y inside the band onto the merge anchor. Neither a fixed column nor
+  // the pointer's column is universally safe (a merge can live in either), so pick the first
+  // visible column that is free of vertical merges in the scanned range (DEV-2115).
+  const rowReferenceColumn = findRowReferenceColumn(
+    hotInstance,
+    firstPartiallyVisibleColumn ?? 0,
+    lastPartiallyVisibleColumn ?? (hotInstance.countCols() - 1),
+    firstPartiallyVisibleRow ?? 0,
+    lastPartiallyVisibleRow ?? (hotInstance.countRows() - 1),
+  );
```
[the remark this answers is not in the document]
adrianspdev: Confirmed and fixed in 6376559259. The reference column is now resolved **per row-scan branch** (fixed-top, fixed-bottom, scrollable) over that branch's own row range, so a vertical merge confined to the frozen rows is detected for the fixed overlays. Added a unit test covering a vertical merge confined to the frozen top rows.

handsontable/src/helpers/dom/cellCoords.ts
```diff
@@ -93,6 +93,49 @@ function findRowAtY(
...
+  for (let column = startColumn; column <= endColumn; column++) {
+    let isDistortedByMerge = false;
+
+    for (let row = startRow; row <= endRow; row++) {
+      const { rowspan, hidden } = hotInstance.getCellMeta<{ rowspan?: number, hidden?: boolean }>(row, column);
+
+      if (hidden === true || (rowspan !== undefined && rowspan > 1)) {
+        isDistortedByMerge = true;
+        break;
+      }
```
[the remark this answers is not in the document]
adrianspdev: Confirmed and fixed in 6376559259. `findRowReferenceColumn` no longer rejects on `hidden`. A distorting (vertically spanned) column is now detected by `rowspan > 1` or element identity across consecutive rows; a purely horizontal-merge slave resolves to a single-row master cell (distinct per row) and stays a valid reference. Added a unit test for a hidden horizontal-merge slave.

handsontable/src/helpers/dom/cellCoords.ts
```diff
@@ -93,6 +93,62 @@ function findRowAtY(
...
+
+      if (cell !== null) {
+        previousCell = cell;
+      }
+    }
+
+    if (!spansMultipleRows) {
+      return column;
+    }
```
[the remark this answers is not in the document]
adrianspdev: Confirmed and fixed in e24d30e98e. `findRowReferenceColumn` now requires at least one renderable (non-null) cell in the scanned range before accepting a column, so a hidden column (`hiddenColumns`) between the merged column and a usable one is skipped instead of picked. Added a unit test and an E2E test (`hiddenColumns` + vertical-merge autofill).

handsontable/src/helpers/dom/cellCoords.ts
```diff
@@ -93,6 +93,67 @@ function findRowAtY(
...
+      // rendered element for every row it covers; either signal marks the column as distorting.
+      if ((rowspan !== undefined && rowspan > 1) || (cell !== null && cell === previousCell)) {
+        spansMultipleRows = true;
+        break;
+      }
+
+      if (cell !== null) {
+        previousCell = cell;
+        hasRenderableCell = true;
+      }
```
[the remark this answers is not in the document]
adrianspdev: Investigated — this one is a false positive, no change made.

I reproduced the exact precondition (vertical merge `A6:A10`, scrolled so the anchor is above the viewport and only slave row 9 is in range: `getCell(9,0)` returns the 146px master, `hidden=true`, no `rowspan` in range) and then drove a real fill-handle drag on that build. The fill resolves correctly: live range `9→11`, row 11 filled from the source, row 13 untouched.

Reason: the tall master cell has the correct top **and** full height, so `findRowAtY` stays geometrically self-consistent for the visible below-anchor rows — there is no collapse. The DEV-2115 collapse comes specifically from `row += rowspan - 1` skipping rows, which only fires when a `rowspan > 1` **anchor** is inside the scanned range — already handled by the existing rowspan check. A lone slave row with the anchor scrolled above does not trigger that skip, so row targeting is unaffected.

budnix (commented):
Went through the reference-column logic against `develop`. The per-branch fix and the renderable-cell guard both look right to me. Left a few notes below — the viewport-null fallback is the one I would want handled before merge, the rest are smaller or follow-up material.

One note on the earlier horizontal-merge thread: I went looking for a problem there and did not find one. `Table.getCell()` runs `onModifyGetCellCoords` (`cellAccess.ts:70`), which remaps a merged slave to the merge master, so a horizontal slave really does resolve to a master cell at single-row height. Your reasoning on that thread holds.

handsontable/src/helpers/dom/cellCoords.ts
```diff
@@ -214,6 +275,18 @@ export function getCellCoordsFromMousePosition(
...
+    throwWithCause('Failed to resolve cell coordinates from mouse position.');
+  }
+
+  // Resolve the row against a column that has no vertical merge across the rows being walked.
+  // `findRowAtY` skips a merged cell's spanned rows, so measuring against a vertically merged
+  // column collapses every Y inside the band onto the merge anchor. Neither a fixed column nor
+  // the pointer's column is universally safe (a merge can live in either), so each row-scan
+  // branch picks the first column free of vertical merges within its own row range (DEV-2115).
+  const rowReferenceColumnStart = firstPartiallyVisibleColumn ?? 0;
+  const rowReferenceColumnEnd = lastPartiallyVisibleColumn ?? (hotInstance.countCols() - 1);
```
budnix: When `lastPartiallyVisibleColumn` is null this opens the search to every column in the grid, and the scroll branch below does the same for rows with `countRows() - 1`. In that state nothing is rendered, so `getCell` returns null on every probe, `hasRenderableCell` never flips, and no column breaks early — the loop walks the full column × row product and calls `getCellMeta` on each step. Your own comment on line 266 says this state is real (grid off-screen), and `dragToScroll` re-enters this on a timer, so on a large grid one held mouse button is enough to lock the tab. The row bound at the scroll branch needs the same treatment, otherwise the scan is still `countRows()` long.

```suggestion
  // When the viewport queries return null nothing is rendered, so there is no cell to measure
  // against. Keep the search inside the rendered range instead of opening it to the whole grid.
  const rowReferenceColumnEnd = lastPartiallyVisibleColumn ?? rowReferenceColumnStart;
```

handsontable/src/helpers/dom/cellCoords.ts
```diff
@@ -93,6 +93,67 @@ function findRowAtY(
...
+  endRow: number,
+): number {
+  for (let column = startColumn; column <= endColumn; column++) {
+    let spansMultipleRows = false;
+    let hasRenderableCell = false;
+    let previousCell: HTMLElement | null = null;
+
+    for (let row = startRow; row <= endRow; row++) {
+      const cell = hotInstance.getCell(row, column, true);
+      const { rowspan } = hotInstance.getCellMeta<{ rowspan?: number }>(row, column);
```
budnix: `handsontable/CLAUDE.md` asks for `getCellMetaTransient` in read-only loops — `getCellMeta` permanently stores one meta object per cell it visits, and viewport eviction cannot sweep those. This loop runs on every mousemove of an autofill drag and every `dragToScroll` tick, so a long drag over a wide grid keeps meta alive for the whole scanned area. Nothing here writes to the meta, so it is a drop-in. The same applies to the `getCellMeta` calls in `findRowAtY` and `findColumnAtX` above, if you want to catch those while you are here.

```suggestion
      const { rowspan } = hotInstance.getCellMetaTransient<{ rowspan?: number }>(row, column);
```

handsontable/src/helpers/dom/cellCoords.ts
```diff
@@ -93,6 +93,67 @@ function findRowAtY(
...
+
+    // A hidden column (e.g. `hiddenColumns`) renders no cells, so `findRowAtY` cannot measure
+    // against it. Require at least one renderable cell so a hidden column between the merged
+    // column and a usable one is skipped rather than picked as the reference.
+    if (!spansMultipleRows && hasRenderableCell) {
+      return column;
+    }
+  }
+
+  return startColumn;
```
budnix, line 158: If every column in the scanned range carries a vertical merge, this hands back the merged column the search exists to avoid, and the row lookup collapses onto the merge anchor again. That is the DEV-2115 symptom, just narrower — it needs a layout where all visible columns are merged across the same band, which does happen in grouped report grids. Curious whether this is a deliberate best-effort fallback. If it is, a line of comment saying so would save the next reader a trip through the loop.

handsontable/src/helpers/dom/cellCoords.ts
```diff
@@ -93,6 +93,67 @@ function findRowAtY(
...
+ * cannot leak into the body lookup or vice versa.
+ *
+ * @param {Handsontable} hotInstance The Handsontable instance.
+ * @param {number} startColumn First column to consider.
+ * @param {number} endColumn Last column to consider (inclusive).
+ * @param {number} startRow First row of the scanned range.
+ * @param {number} endRow Last row of the scanned range (inclusive).
+ * @returns {number} A column index free of vertical-merge distortion, or `startColumn` if none qualifies.
+ */
+function findRowReferenceColumn(
```
budnix, line 117: There is a second copy of this lookup in `3rdparty/walkontable/src/utils/pointerToCoords.ts` — its own `getCellCoordsFromMousePosition` and `findRowAtY`, wired into `event.ts:393` for drags that leave the table. That copy measures against `firstPartiallyVisibleColumn` and reads no `rowspan` at all, so it adds a merge band's full height once per band row and lands several rows off. Nothing this PR broke, but the two copies will keep drifting apart. May I suggest a follow-up to move the merge-aware row measuring into one shared helper?

handsontable/src/helpers/dom/cellCoords.ts
```diff
@@ -214,6 +275,18 @@ export function getCellCoordsFromMousePosition(
...
 
+  if (foundColumn === null) {
+    throwWithCause('Failed to resolve cell coordinates from mouse position.');
+  }
+
+  // Resolve the row against a column that has no vertical merge across the rows being walked.
+  // `findRowAtY` skips a merged cell's spanned rows, so measuring against a vertically merged
+  // column collapses every Y inside the band onto the merge anchor. Neither a fixed column nor
+  // the pointer's column is universally safe (a merge can live in either), so each row-scan
+  // branch picks the first column free of vertical merges within its own row range (DEV-2115).
```
budnix, line 290: The X axis has the same shape of problem and is untouched here. `findColumnAtX` skips `colspan` the way `findRowAtY` skips `rowspan`, and both call sites still pass the raw `firstPartiallyVisibleRow` — so a horizontal merge in the first visible row maps every X inside the band to the anchor column, and over-counts the width for the columns after it. The existing merge autofill spec misses it because its merges start at row 1. Likely out of scope for this PR, but worth a ticket alongside it.

handsontable/src/helpers/dom/cellCoords.ts
```diff
@@ -255,7 +330,9 @@ export function getCellCoordsFromMousePosition(
       : null;
 
     if (bottomStartNonHiddenRow !== null && bottomEndNonHiddenRow !== null) {
-      const fixedBottomCell = hotInstance.getCell(bottomStartNonHiddenRow, firstPartiallyVisibleColumn, true);
+      const rowReferenceColumn = findRowReferenceColumn(
+        hotInstance, rowReferenceColumnStart, rowReferenceColumnEnd, bottomStartNonHiddenRow, bottomEndNonHiddenRow);
```
budnix, line 343: The fixed-bottom branch got the same wiring as the other two, but nothing exercises it — all five stubs hardcode `countNotHiddenFixedRowsBottom: () => 0` and none of the new E2E tests sets `fixedRowsBottom`. You could swap the two row arguments on this line and the whole suite would still pass. A `fixedRowsBottom` grid with a merge inside the frozen band is the exact case the JSDoc promises to handle, so it would be good to pin down.

handsontable/src/helpers/dom/__tests__/cellCoords.unit.js
```diff
@@ -304,4 +304,357 @@ describe('getCellCoordsFromMousePosition', () => {
...
+        hasColHeaders: () => false,
+        hasRowHeaders: () => false,
+        getFirstPartiallyVisibleRow: () => 0,
+        getLastPartiallyVisibleRow: () => 5,
+        getFirstPartiallyVisibleColumn: () => 0,
+        getLastPartiallyVisibleColumn: () => 4,
+        countRows: () => 6,
+        countCols: () => 5,
+        getCell: (row, col) => {
+          const el = document.createElement('td');
```
budnix: Every stub builds a fresh `<td>` on each `getCell` call, so `cell === previousCell` on cellCoords.ts:135 can never be true in any of these tests. That branch matters most in the real grid when the merge anchor sits above the scanned range — the slave rows carry no `rowspan`, so the shared master element is the only thing that gives the merge away. As it stands you could delete the identity check and the suite would stay green. A stub that returns one cached element per merge band would cover it.

handsontable/src/helpers/dom/__tests__/cellCoords.unit.js
```diff
@@ -304,4 +304,357 @@ describe('getCellCoordsFromMousePosition', () => {
...
+    const colWidth = 80;
+    const mergeSpan = 3; // column 0 merged across rows 0-2
+
+    /**
+     * Builds a HOT stub: column 0 is a vertical merge (rows 0-2), column 1 is a hidden slave of
+     * a purely horizontal merge (single-row height per row).
+     *
+     * @returns {object} Stubbed HOT instance.
+     */
+    function buildHotWithHorizontalSlave() {
```
budnix: These four factories are near-identical — same `rootWindow`, `rootElement`, index mappers and `view` block, with a handful of lines different in each. `buildHot` at the top of the file is already an options-bag factory doing exactly this job. Adding `mergedCells` and `hiddenColumns` options there would fold all four in, and leave one place to edit when the stub contract changes.

handsontable/src/helpers/dom/cellCoords.ts
adrianspdev: Fixed in b04629e. Column bound now clamps to `rowReferenceColumnStart` and the scroll row-scan clamps to `scrollRowScanStart` when the viewport query is null, so nothing walks the whole grid off-screen.

handsontable/src/helpers/dom/cellCoords.ts
adrianspdev: Fixed in b04629e. Switched all three scan loops (`findRowReferenceColumn`, `findRowAtY`, `findColumnAtX`) to `getCellMetaTransient`.

handsontable/src/helpers/dom/cellCoords.ts
adrianspdev: Deliberate best-effort fallback — added a comment in b04629e saying so.

handsontable/src/helpers/dom/cellCoords.ts
adrianspdev: Fixed in b04629e. Added a `fixedRowsBottom` test with a merge in the frozen bottom band; it fails if the row-scan args are swapped or the merged column is chosen.

handsontable/src/helpers/dom/__tests__/cellCoords.unit.js
adrianspdev: Fixed in b04629e. New test puts the merge anchor above the scanned range (rows 8-12, scan starts at 10) so only element identity reveals it; asserts row 12, not the collapsed row 10.

handsontable/src/helpers/dom/__tests__/cellCoords.unit.js
adrianspdev: Fixed in b04629e. Folded all four factories into `buildHot` via `mergedCells`/`hiddenColumns`/`hiddenMetaColumns`/`fixedRowsTop`/`fixedRowsBottom` options.

handsontable/src/helpers/dom/cellCoords.ts
adrianspdev: Agreed — follow-up filed: #13107 (unify merge-aware pointer→row mapping into one shared helper). Out of scope here.

handsontable/src/helpers/dom/cellCoords.ts
adrianspdev: Agreed — ticket filed: #13108 (X-axis colspan analog). Out of scope here.

handsontable/src/helpers/dom/cellCoords.ts
adrianspdev: Correction: tracking on ClickUp as **DEV-2123** (was GitHub #13107, now closed).

handsontable/src/helpers/dom/cellCoords.ts
adrianspdev: Correction: tracking on ClickUp as **DEV-2124** (was GitHub #13108, now closed).

budnix (approved):

PR #13122, opened 2026-07-27
DEV-2133: Fix the numbro docs example for Angular

demtario:
### Context

The PR fixes the `numbro` recipe's Angular example, which never rendered in the docs example runner. A runner-links sweep reported it as `no-grid` after a 240s timeout; the timeout hid two separate defects.

**1. The Angular project failed to build (TS7016).**

```
✘ [ERROR] TS7016: Could not find a declaration file for module 'numbro/dist/languages.min.js'.
  '/app/node_modules/.pnpm/numbro@2.5.0/node_modules/numbro/dist/languages.min.js' implicitly has an 'any' type.
    src/app/app.component.ts:4:22
```

numbro 2.5.0 declares `typings: "./numbro.d.ts"` only - the `dist/languages.min.js` subpath ships no declarations. The runner's generated Angular project type-checks with `strict: true`, so the import is a hard build error. The JavaScript and React variants use the identical import but run transpile-only, so they built fine.

This was not only a runner problem: any reader copy-pasting the published Angular snippet into a strict Angular app hit the same error. Our own CI did not catch it, because `docs/angular-type-check/types/shims.d.ts` declared the module - the shim made the type-check job green while the example stayed broken everywhere else.

The fix suppresses the error at the import site with `@ts-expect-error` and removes the shim, so `docs-angular-typecheck.yml` now guards this exact failure. `@ts-expect-error` rather than `@ts-ignore` on purpose: it fails the check once numbro ships declarations for that subpath, so the suppression gets removed instead of outliving its cause.

**2. The same example threw at runtime.**

```
Error: No registered validator found under "numeric" name
```

`app.component.ts` calls `getValidator('numeric')` and `getEditor('numeric')` while its module body runs, but `registerAllModules()` sat in `app.config.ts`. ES imports are hoisted, so the component module is evaluated first and the registry lookup throws. The runner's generated `main.ts` has the same import order, so fixing only the build error would still have produced `no-grid`. The call now lives at the top of `app.component.ts`; the other four variants of this example already registered in the same file. The `app.config.ts` call is left in place - it is idempotent and keeps the example structurally consistent with the other Angular recipes.

**3. The page was missing from the React and Angular sidebars.**

Navigation hid the page in two independent places.

`docs/content/recipes/sidebar.js` registered the recipe as `onlyFor: ['javascript', 'vue']`, even though the page carries dedicated `::: only-for react` and `::: only-for angular` blocks with real example directories behind them. The entry now lists all four frameworks and sits with the other all-framework recipe.

That alone was not enough. `docs/cloudflare/_worker.js` also force-301'd both the React and Angular URLs to the cell-types index, so the newly added sidebar links bounced straight back to the listing - caught on this PR's preview deployment. Those two redirect entries are removed, with a worker test covering both URLs.

`numbro/dist/languages.min.js` is the only untyped third-party subpath imported by any Angular docs example, so no other example needs the directive.

`docs/AGENTS.md` gains a rule for each defect, and `docs/angular-type-check/README.md` now steers authors to `@ts-expect-error` at the import site rather than a shim.

### How has this been tested?

Angular type-check, from the repo root:

```bash
npm run build --prefix wrappers/angular-wrapper
npm run typecheck --prefix docs/angular-type-check
```

- Result: `0 in both, 0 local-only, 7 latest-only`. Byte-identical to the same command run against a stashed working tree, so this PR introduces no new errors. The 7 `LATEST-ONLY` failures are pre-existing on `develop` and unrelated to this change (comments, merge-cells, cell-editor, cell-renderer, column-groups, row-parent-child).
- The guard was verified rather than assumed: temporarily restoring the numbro entry in `types/shims.d.ts` produced `TS2578: Unused '@ts-expect-error' directive` at `recipes/cell-types/numbro/angular/example1.ts:5` in **both** Handsontable versions. Reverted afterwards.

Manual, against the docs dev server (`npm run dev --prefix docs`), on the numbro recipe page:

- Angular tab - previously blank with the validator error in the console, now renders. Rows read `Lunar Core | Lander | Ellen Ripley | 2 | $350,000.00`, confirming numbro formatting is applied.
- JavaScript and React tabs - render identically before and after, confirming the directive changes nothing at runtime.
- The pre-existing nature of the Angular runtime failure was confirmed by stashing the change and reloading: the same console error appeared.
- Sidebar - the recipe now appears under **Cell Types** on the React, Angular, Vue, and JavaScript routes. Checked by reading the rendered nav links on each route, so both the addition and the absence of a regression are covered.

Worker redirect tests:

```bash
node --test docs/cloudflare/__tests__/_worker.test.mjs
```

- 15 pass, including a new case asserting the React and Angular numbro URLs are served rather than redirected. Confirmed to fail before the worker change - stashing only `_worker.js` and re-running gives `1 fail` with `/docs/react-data-grid/recipes/cell-types/numbro/ must be served, not redirected`.

No changelog entry - documentation-only change.

[skip changelog]

### Types of changes

### Related issue(s):

1. DEV-2133

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

### Follow-up (not in this PR)

The docs example runner bakes each wrapped project into committed artifacts, so this fix is inert until they regenerate. The **Import versioned docs examples** workflow in `handsontable/examples` regenerates the `next` bucket from `develop` and the release bucket from `prod-docs/<major.minor>` - so the `18.0` bucket, which released-version runner links point at, needs a cherry-pick to `prod-docs/18.0` first. That cherry-pick is deliberately left out of this PR and will be done separately.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2133

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 88e9569fd2f48ca0e1006564b3ccc4c1ccd96750. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2133 Angular `numbro` recipe example fails to build in the docs example runner (TS7016)](https://app.clickup.com/t/9015210959/DEV-2133)

demtario:
Verified on preview

adrianspdev (approved):

PR #13125, opened 2026-07-27
DEV-2136: Expose four cell-type recipes to React and Angular readers

demtario:
### Context

The PR makes four **Cell Types** recipes reachable for React and Angular readers: Flatpickr, Pikaday, Moment.js-based date, and Moment.js-based time.

Each of the four ships real `react/` and `angular/` example directories and a matching `::: only-for react` / `::: only-for angular` block, and each renders correctly on both framework routes. They were still treated as JavaScript-and-Vue-only by **two independent layers**, so neither framework's readers could get to them:

1. `docs/content/recipes/sidebar.js` registered them with `onlyFor: ['javascript', 'vue']`, so no link appeared in the React or Angular sidebar.
2. `docs/cloudflare/_worker.js` force-301'd their React and Angular URLs to the cell-types index, so even a direct link or a search result bounced back to the listing.

The second layer is the one that matters, and it is easy to miss. Fixing only the sidebar produces a visible link that immediately redirects away - which is exactly what happened on the preview for the related numbro change. Both layers are updated here.

The eight redirect entries sat under a `// Recipe cell-type slug mismatches` comment, but they were never slug mismatches - the rest of that map maps an old or framework-specific slug onto its canonical page, while these pointed at the index because the page genuinely did not exist for those frameworks at the time. The pages have since gained React and Angular content and the redirects were never revisited.

`numbro` was fixed separately in #13122, which also carries the build and runtime fixes its Angular example needed. This branch is rebased on top of that merge, so the `// JavaScript + Vue only` sidebar heading is gone - all five recipes now sit in the all-frameworks group - and the worker's "must be served" test covers all five slugs in one loop rather than duplicating numbro's case.

`color-picker`, `feedback`, and `rating` show the same React-side mismatch but are correct as they stand - each has a dedicated React-only page (`colorful-picker`, `feedback-react`, `react-rating`), so hiding the shared page from the React sidebar avoids two entries for the same topic. Their redirects map React URLs onto those counterpart pages and are untouched.

On the Angular side there is no such duplication: `guide-datepicker-angular` is a distinct recipe - a native HTML5 date input with `date-fns` - not a counterpart to the Flatpickr, Pikaday, or Moment date pages, which are each about a specific library.

### How has this been tested?

Worker redirect tests:

```bash
node --test docs/cloudflare/__tests__/_worker.test.mjs
```

- 16 pass, including the widened "must be served" case (all five slugs on both React and Angular) and a new case asserting the counterpart-page mappings still redirect.
- Before the rebase, the "must be served" case was confirmed to fail without the worker change: `1 fail` with `/docs/react-data-grid/recipes/cell-types/flatpickr/ must be served, not redirected`.

Manual, against the docs dev server (`npm run dev --prefix docs`):

- All eight routes render a populated grid - four pages on React and Angular, 30 to 50 cells each, no console errors. Checked before touching navigation, so no page is listed that does not work.
- Sidebar after the change: the React sidebar lists the four plus Radio buttons and the three React-only pages, and does not list `color-picker` / `feedback` / `rating`. The Angular sidebar lists the four plus its existing entries. The Vue and JavaScript sidebars are unchanged and still include numbro.

No changelog entry - documentation-only change.

[skip changelog]

### Types of changes

### Related issue(s):

1. DEV-2136

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

### Worth a follow-up

Nothing cross-checks a page's `::: only-for` blocks against its `sidebar.js` `onlyFor` and the worker's redirect map. A check over those three would have caught all of this, and would catch the next page that gains a framework variant without its navigation being updated.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2136

---

> [!NOTE]
> **Low Risk**
> Documentation-only routing and navigation changes; no product library or auth/data paths affected, with regression tests on worker redirect behavior.
> 
> **Overview**
> React and Angular readers can now reach four **Cell Types** recipes (Flatpickr, Pikaday, Moment.js date/time) that already had framework-specific content but were hidden from navigation and blocked by redirects.
> 
> **`sidebar.js`** adds `react` and `angular` to `onlyFor` for those four entries (alongside the existing all-framework **Numbro** entry), so the recipes appear in each framework’s sidebar.
> 
> **`docs/cloudflare/_worker.js`** removes eight forced 301 rules that sent `/docs/{react|angular}-data-grid/recipes/cell-types/{slug}/` to the cell-types index. Those URLs now fall through to static assets like the JavaScript/Vue routes.
> 
> **Worker tests** assert all five multi-framework slugs (`flatpickr`, `pikaday`, `moment-date`, `moment-time`, `numbro`) are served—not redirected—on React and Angular, and that slug-mismatch redirects (e.g. React `color-picker` → `colorful-picker`) are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2e7ef6ea087ca216866fc484b9407053dc1532d6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2136 Four cell-type recipes are missing from the React and Angular sidebars](https://app.clickup.com/t/9015210959/DEV-2136)

demtario:
Verified on preview

qunabu (approved):

PR #13130, opened 2026-07-28
DEV-2138: Fix editorFactory throwing when shortcuts have no shortcutsGroup

demtario:
### Context

The PR fixes custom editors created with `editorFactory` throwing `Error: You need to define the shortcut's group.` as soon as the editor opens, whenever `shortcuts` are declared but `shortcutsGroup` is not.

`editorFactory` passed `{ group: shortcutsGroup }` to `Context.addShortcuts()` unconditionally, while `shortcutsGroup` is an optional option in both the types and the documentation. `Context.addShortcut` throws when `group` is `undefined`, so a valid, documented configuration crashed at open time. Reported in production by Sentry ([HANDSONTABLE-DOCS-20P](https://handsoncode.sentry.io/issues/HANDSONTABLE-DOCS-20P), 28 events / 12 users), where it breaks every cell-type recipe on the docs site (pikaday, rating, feedback, moment-date, color-picker, flatpickr). The defect shipped in 18.0.0.

`close()` had the matching defect: it removed shortcuts by group `''`, so once open stopped throwing, every open would leak a shortcut into the `editor` context. Both sites now use the same `customEditor` default. A per-shortcut `group` still overrides it, because `addShortcuts` only fills in keys a shortcut does not already define.

The PR also fixes the crash this throw caused downstream in the date-picker recipes: [HANDSONTABLE-DOCS-20Q](https://handsoncode.sentry.io/issues/HANDSONTABLE-DOCS-20Q) and [HANDSONTABLE-DOCS-21X](https://handsoncode.sentry.io/issues/HANDSONTABLE-DOCS-21X), `TypeError: Cannot read properties of undefined (reading 'destroy')` in `afterClose`, 12 events / 7 users, same sessions as 20P. The `pikaday` and `moment-date` recipes create their Pikaday instance in `afterOpen` and destroy it in `afterClose` without checking that it exists. The throw in `open()` happened after `_opened = true` but before `afterOpen`, so the instance was never created and the next close crashed.

Guarding the recipes is needed on its own, not only because of 20P: `open()` calls `refreshDimensions()` before `afterOpen`, and `refreshDimensions()` calls `close()` when the edited cell is no longer rendered. `afterClose` can therefore run for an editor that never opened. Both teardown paths - `afterClose` and `hideDatepicker` - now check the instance first, in the JavaScript, TypeScript, and React variants and in the mirrored guide snippets. The Angular variants already guarded it. `flatpickr` and `color-picker` create their instances in `init` / `afterInit`, so they are not affected.

The `afterClose`-without-`afterOpen` asymmetry is recorded in `handsontable/AGENTS.md`. Whether the core should guarantee that `afterClose` never fires without `afterOpen` is left out of this PR - it would change behavior for every existing editor.

### How has this been tested?

- New case in `handsontable/src/editors/__tests__/index.spec.js`: opens an `editorFactory` editor that declares `shortcuts` without `shortcutsGroup`, asserts no error is thrown, the shortcut is registered under the default group, its callback fires on ArrowLeft, and it is removed again on `close()`.
- Verified the test is a real regression test: reverting the default in `factory.ts` makes it fail with `Error: You need to define the shortcut's group.`

```bash
npm --prefix handsontable run test:e2e -- --testPathPattern=editors/__tests__   # 32 specs, 0 failures
```

- The recipe changes are snippet-level guards in the docs examples. They were not verified in a browser, and the `refreshDimensions` path that makes the guard necessary is not covered by a test.

### Types of changes

### Related issue(s):

1. DEV-2138

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2138

---

> [!NOTE]
> **Low Risk**
> Targeted bug fix in custom editor shortcut wiring with a regression test; no auth, data, or API surface changes beyond fixing documented optional config.
> 
> **Overview**
> Fixes **`editorFactory`** custom editors that define **`shortcuts`** but omit **`shortcutsGroup`**, which previously threw *"You need to define the shortcut's group."* on open because an undefined group was passed into the shortcut context.
> 
> Registration and teardown now share a **`customEditor`** default group so shortcuts register, fire, and are removed on **`close()`** instead of leaking under an empty group.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3f3f45337eab5cc68add0dfd1c9622d5d16dfcdd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2138 Fix editorFactory throwing when shortcuts are defined without shortcutsGroup](https://app.clickup.com/t/9015210959/DEV-2138)

qunabu (approved):

adrianspdev (approved):

PR #13132, opened 2026-07-28
ci: get the branch name to the slugifier via env instead of interpolating it

kobihikri:
Hi, and thanks for Handsontable.

In `.github/workflows/performance-tests.yml`, the "Compute report URL" step slugifies the branch name:

```yaml
run: |
  BRANCH_SLUG=$(echo "${{ github.head_ref }}" | sed 's/[^a-zA-Z0-9._-]/-/g')
```

The `sed` is doing exactly the right thing — it strips everything outside `[a-zA-Z0-9._-]`. The problem is only one of ordering: Actions expands `${{ ... }}` into the script text before bash runs, so the branch name is already part of the command by the time `sed` gets a chance to clean it. Git allows `$`, `(`, `)` and backticks in branch names, and `$(...)` runs inside double quotes, so a fork PR on a branch named something like `x$(id)` would execute that before the slugifier ever sees the value.

Scope, so I do not overstate it: this is a `pull_request` workflow, so a fork PR carries a read-only token and no secrets. I am raising it as hardening.

The fix keeps your slugifier exactly as it is and just gets the value to it safely:

```yaml
env:
  HEAD_REF: ${{ github.head_ref }}
run: |
  BRANCH_SLUG=$(echo "$HEAD_REF" | sed 's/[^a-zA-Z0-9._-]/-/g')
```

`BRANCH_SLUG`, `PAGES_URL` and both `$GITHUB_OUTPUT` writes are unchanged, so report URLs come out identical. Two lines added, one changed.

Disclosure: I used AI assistance to help spot this and prepare the change, and I read the workflow and its trigger myself.

---

> [!NOTE]
> **Low Risk**
> Small CI-only hardening of how a branch name is passed into bash; no product, auth, or data-handling changes. The workflow remains pull_request-scoped.
> 
> **Overview**
> Hardens the performance-report URL step so the PR branch name is not interpolated into the bash script before slugification.
> 
> `github.head_ref` is now passed as `HEAD_REF` via `env` in both `performance-tests.yml` and the `performance-run` composite action, then slugified from `"$HEAD_REF"`. Report URLs and outputs are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8b8a4739a8f15083ec2f7f884e4f28529cbdb88b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/actions/performance-run/action.yml
+++ b/.github/actions/performance-run/action.yml
@@ -57,8 +57,10 @@ runs:
       id: report-url
       if: github.event_name == 'pull_request'
       shell: bash
+      env:
+        HEAD_REF: ${{ github.head_ref }}
       run: |
-        BRANCH_SLUG=$(echo "${{ github.head_ref }}" | sed 's/[^a-zA-Z0-9._-]/-/g')
+        BRANCH_SLUG=$(echo "$HEAD_REF" | sed 's/[^a-zA-Z0-9._-]/-/g')
         PAGES_URL="https://handsontable.github.io/handsontable/performance-reports/${BRANCH_SLUG}/"
         echo "branch_slug=${BRANCH_SLUG}" >> "$GITHUB_OUTPUT"
         echo "pages_url=${PAGES_URL}" >> "$GITHUB_OUTPUT"
--- a/.github/workflows/performance-tests.yml
+++ b/.github/workflows/performance-tests.yml
@@ -80,8 +80,10 @@ jobs:
       - name: Compute report URL
         id: report-url
         if: github.event_name == 'pull_request'
+        env:
+          HEAD_REF: ${{ github.head_ref }}
         run: |
-          BRANCH_SLUG=$(echo "${{ github.head_ref }}" | sed 's/[^a-zA-Z0-9._-]/-/g')
+          BRANCH_SLUG=$(echo "$HEAD_REF" | sed 's/[^a-zA-Z0-9._-]/-/g')
           PAGES_URL="https://handsontable.github.io/handsontable/performance-reports/${BRANCH_SLUG}/"
           echo "branch_slug=${BRANCH_SLUG}" >> "$GITHUB_OUTPUT"
           echo "pages_url=${PAGES_URL}" >> "$GITHUB_OUTPUT"


demtario (changes requested):
@kobihikri thanks for the report and for the careful write-up. The analysis is correct and the fix pattern is the right one.

Confirming the mechanism for anyone reading later: Actions expands `${{ ... }}` into the script text before bash executes it, so `github.head_ref` becomes part of the command itself and the `sed` slugifier never gets the chance to run first. Git permits `$`, `(`, `)` and backticks in ref names, so a branch named `x$(...)` is a valid branch and the substitution runs. Moving the value into `env:` and reading `"$HEAD_REF"` is the canonical fix, and the slug output is unchanged for every normal branch name.

One thing to flag: there are two copies of this step, and the PR patches the one that is nearly unreachable.

- `.github/workflows/performance-tests.yml:84`, the file you changed. It is `workflow_call`-only, and its sole caller is `develop.yml`, which runs on push to `develop`. Inside a called workflow `github.event_name` resolves to the caller's event, so the step's `if: github.event_name == 'pull_request'` guard is false on that path and the step is skipped.
- `.github/actions/performance-run/action.yml:61`, the same three lines, and this one is the live PR path: `test.yml` (on `pull_request`) calls `performance.yml`, which calls this composite. It runs on every pull request that touches `handsontable/src/**`, including pull requests from forks.

So the vulnerable interpolation you found is real and reachable, just in the file the PR does not touch.

Interestingly, your own scoping note describes the composite rather than the file you patched. `performance-tests.yml` is reached from a push with `contents: write`, while the composite is the one that runs under a `pull_request` event where a fork's token is clamped to read-only and no secrets are passed. Your assessment of the blast radius is accurate. It just applies to the other site.

Worth noting the patched copy is not entirely dead either. `develop.yml` carries a `pull_request` trigger path-filtered to itself, so a pull request touching both `.github/workflows/develop.yml` and `handsontable/src/**` satisfies the `test-performance` scope flag and reaches the step with `event_name == 'pull_request'`. Narrow, but it means the change is worth keeping on its own merits, and the file's own header warns against wiring new PR behavior there.

Could you extend the PR to apply the identical change to `.github/actions/performance-run/action.yml:61`? Same two lines added, one changed. We would rather have both sites fixed in your PR than open a competing one.

For completeness, we checked the rest of the workflow surface. The only other untrusted-context interpolations are `github.ref_name` in `publish.yml` and `docs-staging.yml`, which require push access to the repository and are not fork-controllable. `GITHUB_HEAD_REF` reaching `node scripts/run.mjs` is passed as an environment variable and is fine, and the downstream `DEPLOY_DIR` and `git commit -m` uses are quoted expansions of the already-slugified value.

kobihikri:
Thank you — you're right, and thank you for taking the time to trace both call paths rather than just pointing at the file.

I checked your reading against the repository and it holds: `test.yml` (on `pull_request`) calls `performance.yml`, which calls the `performance-run` composite, so `action.yml:61` runs on every pull request touching `handsontable/src/**`, forks included. The copy I patched is `workflow_call`-only and `github.event_name` inside it resolves to the caller's event, so the `pull_request` guard is false on the `develop.yml` push path. I had the trigger profile right in the scoping note and the file wrong in the diff, which is a fair thing to have caught.

I've pushed the identical change to `.github/actions/performance-run/action.yml`: `github.head_ref` into `env:`, the slugifier reading `"$HEAD_REF"`, `sed` untouched. Both sites are now covered in this PR.

For what it's worth, the slug is byte-identical for every branch name that was already valid — the only inputs whose behaviour changes are the ones that were being executed rather than slugified.

*(Disclosure: I use AI tooling to help me read code at this scale. The reproduction and the change are mine, and I've checked them.)*

demtario (approved):

demtario (commented):
Thanks @kobihikri, this is exactly what was asked. `.github/actions/performance-run/action.yml` now carries the identical `env: HEAD_REF` indirection, so the live `test.yml` → `performance.yml` → composite path a fork PR actually hits is covered alongside the `workflow_call` copy. Slug output and both `$GITHUB_OUTPUT` writes are unchanged in both files. LGTM.

PR #13128, opened 2026-07-28
DEV-1790: Fix Starlight TOC :has() crash by wrapping init per instance

demtario:
### Context

The PR fixes the uncaught `SyntaxError: Failed to execute 'querySelectorAll' on 'Document'` thrown by Starlight's table-of-contents script on browsers without `:has()` support (Chrome < 105 — the top reporter is Chrome Mobile 75 on Android 6.0 — and Safari < 15.4). Sentry: HANDSONTABLE-DOCS-1GA, 170 events, 14 users, still collecting.

The issue was previously marked fixed, but the patch in `docs/src/components/Head.astro` never took effect. It wrapped `StarlightTOC.prototype.init`, while Starlight assigns `init` as an **own instance field inside the constructor**:

```js
class H extends HTMLElement{constructor(){super(), ... this.init=()=>{ ... }, this.onIdle(()=>this.init())}
```

So `prototype.init` is `undefined`, the patch's own guard `if (!StarlightTOC.prototype.init) return;` bailed out, and nothing was ever wrapped. Verified on live production:

```
ctorFound: true, protoInitType: "undefined", patchGuardBailsOut: true
instanceOwnInit: true, ownProps: [_current, minH, maxH, tocHeadingSelector, onIdle, init]
```

A prototype-level patch cannot work here — the instance field shadows it. This PR wraps `init` per instance instead, by intercepting `customElements.define` and subclassing the constructor. The wrap lands in the subclass constructor right after `super()`, which runs before the `onIdle` macrotask that calls `init()`. It also covers `mobile-starlight-toc`, which extends the same class and threw the same error — the old patch targeted only `starlight-toc`.

Behavior on affected browsers is unchanged apart from the crash: the TOC renders, it just does not highlight the active section while scrolling. Errors that are not a `DOMException` named `SyntaxError` still propagate.

Note for the follow-up cherry-pick: production is served from `prod-docs/18.0`, so this fix does not reach users until it is cherry-picked there.

### How has this been tested?

New regression test `docs/src/components/__tests__/head-starlight-toc-has-guard.test.mjs` (7 cases). It extracts the inline guard script from `Head.astro` and runs it against a harness that mimics Starlight's shape (own-instance arrow `init`), for both `starlight-toc` and `mobile-starlight-toc`. It asserts that the original `init` actually ran (`originalInitCallCount === 1`), so a no-op stub cannot pass it; plus the normal-path return value, rethrow of non-`SyntaxError` errors, and passthrough for unrelated custom elements.

[...]

PR #13135, opened 2026-07-29
DEV-2144: Document the Starlight class-field trap and extend the TOC guard tests

demtario:
### Context

Follow-up to #13128, which landed the per-instance table-of-contents guard and fixed the reported Sentry events ([HANDSONTABLE-DOCS-1GA](https://handsoncode.sentry.io/issues/HANDSONTABLE-DOCS-1GA/), 21Y). This PR carries **no behavior change** - `Head.astro` is byte-identical to `develop`.

It adds `docs/AGENTS.md` §2.14 recording why the patch that preceded #13128 was dead code. Starlight declares the throwing method as a class field (`private init = (): void => {...}`), so it is an own property of every instance and never reaches the prototype. `customElements.get('starlight-toc').prototype.init` is therefore always `undefined`, and a guard written against it reviews clean, passes locally, and fixes nothing in production. That is how the `:has()` guard stayed dead for two months while the issue kept collecting events.

The section also records why the install needs no re-entry flag today, with the reference that decides it: the docs site does not use `<ClientRouter />`, and Astro keys executed scripts by `textContent` (`detectScriptExecuted()` in `astro/dist/transitions/swap-functions.js`), so an unchanged inline head script never runs twice. Add a flag if either of those stops holding.

Finally, it extends the guard's regression suite from 7 tests to 11 with the cases that would have caught the dead patch:

- the instance-property shape - `prototype.init` is `undefined` and `init` is an own property of the instance;
- the absence of any `prototype.init` or `whenDefined` dependency in the shipped script (comments stripped first, since the guard's own comment explains why that approach fails);
- propagation of a `SyntaxError` that is not a `DOMException`, which is a real parse failure in code the site owns;
- preservation of `instanceof` and statics through the subclass wrapper, which is what `class extends` buys.

### How has this been tested?

Docs-site-only change. No core, wrapper, or example code is touched, so the core lint, build, and test suites do not apply.

```bash
node --test docs/src/components/__tests__/head-starlight-toc-has-guard.test.mjs
# tests 11 / # pass 11 / # fail 0

npx eslint docs/src/components/__tests__/head-starlight-toc-has-guard.test.mjs
# ESLint: No issues found
```

Negative control for the new `instanceof`/statics test: severing the guard's static inheritance in `Head.astro` (`Object.setPrototypeOf(GuardedElement, Object.prototype)`) turns that test red, together with the six behavior tests - `class extends` carries the static link and `super()` through the same chain. Reverted after the check; `Head.astro` is unchanged in this PR.

The suite runs in CI under `npm run docs:test:plugins`, which globs `src/components/__tests__/*.test.mjs`.

### Types of changes

Documentation and test coverage only.

### Related issue(s):

1. DEV-2144

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

[skip changelog] - agent documentation and tests only, no package source change.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2144

### DIFF
--- a/docs/AGENTS.md
+++ b/docs/AGENTS.md
@@ -508,3 +508,22 @@ The custom loader (`src/plugins/framework-loader.mjs`) renders every source page
 - **A framework runtime (`react`, `vue`, `zone.js`, `@angular/compiler`) must only be imported inside `loadRuntime()`.** A bare `await import(...)` there escapes every try/catch as an unhandled rejection and leaves each example of that framework stuck on its loading shimmer forever. `loadRuntime()` returns `null` on failure, and the group is marked with `markFailed()` - the only place that shows a reader-visible notice (`.hot-example-error`). Per-example failures keep the silent `markLoaded()` degradation, because some examples render nothing by design.
 
 Guards for both rules live in `src/lib/__tests__/example-error-reporting.test.mjs` (run by `npm run docs:test:plugins`).
+
+---
+
+## 2.14 Patching Starlight's Custom Elements
+
+Starlight's custom elements assign their methods as **class fields** (`private init = (): void => {...}`), not as prototype methods. An own instance property never reaches the prototype, so `customElements.get('starlight-toc').prototype.init` is `undefined` and any prototype patch silently no-ops. A guard written that way looks correct in review, passes locally, and fixes nothing in production - that is exactly how the `:has()` table-of-contents guard stayed dead for two months (Sentry HANDSONTABLE-DOCS-1GA).
+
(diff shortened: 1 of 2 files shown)


qunabu:
Task linked: [DEV-2144 Fix docs TOC :has() SyntaxError reaching window.onerror](https://app.clickup.com/t/9015210959/DEV-2144)

qunabu (dismissed):

qunabu (approved):

PR #13140, opened 2026-07-29
DEV-2149: Report swallowed example failures and guard runtime imports

demtario:
### Context

The PR fixes two opposite error-handling defects in `docs/src/scripts/example-runner.ts`.

**It swallowed too much.** Per-example init failures were turned into `console.error` and nothing else, so an error that kills every grid on a page produced **zero** Sentry events. This is measurable: a 90-day `message:"toSorted" level:error` search over the docs project returns nothing, even though that method demonstrably ships in the bundles. Any reasoning about docs error volume was therefore systematically understating init-time failures.

**It guarded too little.** The shared framework imports sat *outside* every per-example `try`/`catch`: `import('vue')`, `import('zone.js')`, `import('@angular/compiler')`, and `import('react-dom/client')` / `import('react')`. When one rejected, it escaped as an unhandled rejection and the affected examples were left showing a shimmer forever. Sentry [HANDSONTABLE-DOCS-1FX](https://handsoncode.sentry.io/issues/HANDSONTABLE-DOCS-1FX/) (10 events / 7 users) is almost certainly the `vue` one — attribution by frame shape, not a reproduction, so a recurrence after deploy would mean the mapping was wrong.

Changes:

- New `docs/src/lib/example-error-reporting.mjs` exporting `createExampleErrorReporter()`. It owns the noise policy — a per-load cap, message dedupe, and a drop list — and reports through the Sentry Loader Script global, guarded so a blocked CDN request degrades silently instead of throwing inside the error path.
- The runner calls `reportExampleError(err, {framework, phase, source})` from all four per-example catches, from the framework-runtime guard, and from a new top-level `runExamples().catch()`.
- New `loadRuntime(framework, els, load)` wraps the previously unguarded shared imports. On failure it logs, reports, and calls `markFailed()` on every element in that group: shimmer stopped, plus a reader-visible message instead of a permanent skeleton.

Two things reviewers should weigh. This **raises** Sentry volume by design — bounded (capped per page load, deduped by message), but non-zero, so the first day of events after deploy is worth a look; a demo that throws intentionally belongs in the drop list in the `.mjs`, not silenced at the call site. And on branches that have the `beforeSend` hook, `captureException` fires before it, so both filters apply — harmless, but worth knowing when tracing why an event did or did not arrive.

### How has this been tested?

New test file, `docs/src/lib/__tests__/example-error-reporting.test.mjs`:

```bash
cd docs && node --test src/lib/__tests__/example-error-reporting.test.mjs
# tests 20 / # pass 20 / # fail 0
```

Mutation checks, each reverted immediately and the file re-verified green afterwards — these prove the structural assertions are not vacuous:

```
reverted vue to `? await import('vue')`  ->  not ok 11 - loadRuntime() spans are found
                                            not ok 14 - import('vue') only runs inside a loadRuntime() call
```

The suite asserts the structural invariants directly: `@angular/compiler` is imported only inside `loadRuntime()`, `zone.js` only inside `ensureZone()`, and every previously bare import is now inside a guarded span.

All line numbers were re-verified against current `develop` before editing.

Files changed:

- `docs/src/scripts/example-runner.ts`
- `docs/src/lib/example-error-reporting.mjs` (new)
- `docs/src/lib/__tests__/example-error-reporting.test.mjs` (new)
- `docs/src/styles/interactive-example.css`
- `docs/AGENTS.md`

### Types of changes

### Related issue(s):

1. DEV-2149

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2149

---

> [!NOTE]
> **Low Risk**
> Docs-only client error handling and observability; bounded Sentry volume and no changes to Handsontable product APIs.
> 
> **Overview**
> Fixes two opposite problems in the docs **example runner**: per-example failures were only logged (no Sentry), while shared React/Vue/Angular runtime `import()` calls sat outside `try/catch` and could reject as unhandled failures with examples stuck on the loading shimmer.
> 
> Adds **`example-error-reporting.mjs`** with `createExampleErrorReporter()` — forwards real failures to Sentry with framework/phase tags, while filtering chunk-load errors, intentional Handsontable `cause.handsontable` errors, and expected demo `HTTP ###` messages, plus dedupe and a cap of three reports per page. **`example-runner.ts`** wires that reporter through per-example catches, runtime load failures, and a top-level `runExamples().catch()`.
> 
> Wraps framework runtimes in **`loadRuntime()`**; on failure it reports, calls **`markFailed()`** (reader-visible `.hot-example-error` message), and skips the group. **`ensureZone()`** now sets `zoneLoaded` only after `zone.js` resolves. Vue gains the same missing-module early-out as other frameworks. **`AGENTS.md`** documents the rules; unit tests include static checks that runtime imports stay inside `loadRuntime()`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ae0506184fb0956916d1539c88fc33d1b3bd7790. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/AGENTS.md
+++ b/docs/AGENTS.md
@@ -497,3 +497,14 @@ The custom loader (`src/plugins/framework-loader.mjs`) renders every source page
 - Both `.astro/data-store.json` and `.astro/rendered-html/` are derived caches — reset them together with `rm -rf .astro`. Never delete one without the other.
 - `npm run dev` must work with the default Node heap. Do not add `NODE_OPTIONS=--max-old-space-size` workarounds; shrink the data store instead.
 - Plugin unit tests run with `node --test src/plugins/__tests__/*.test.mjs`.
+
+---
+
+## 2.13 Example-Runner Error Handling
+
+`src/scripts/example-runner.ts` catches every example failure so one broken example cannot take a page down. Two rules keep that from hiding real crashes (Sentry HANDSONTABLE-DOCS-20K) or leaking unhandled rejections (HANDSONTABLE-DOCS-1FX):
+
+- **Every caught example failure goes through `reportExampleError()`** from `src/lib/example-error-reporting.mjs`. That module owns the drop list - failed chunk fetches, errors carrying `cause.handsontable` (core `throwWithCause()`), the expected `HTTP <status>` of the server-side data examples - plus deduplication by message and a cap of three forwarded failures per page load. Widen or narrow the drop list there, never at the call site. Sentry reaches the runner through the Loader Script (`window.Sentry`), whose deferred `<script>` tag precedes the bundled Head scripts, so the queueing stub exists by the time the runner runs; a blocked CDN request degrades to no report.
+- **A framework runtime (`react`, `vue`, `zone.js`, `@angular/compiler`) must only be imported inside `loadRuntime()`.** A bare `await import(...)` there escapes every try/catch as an unhandled rejection and leaves each example of that framework stuck on its loading shimmer forever. `loadRuntime()` returns `null` on failure, and the group is marked with `markFailed()` - the only place that shows a reader-visible notice (`.hot-example-error`). Per-example failures keep the silent `markLoaded()` degradation, because some examples render nothing by design.
+
+Guards for both rules live in `src/lib/__tests__/example-error-reporting.test.mjs` (run by `npm run docs:test:plugins`).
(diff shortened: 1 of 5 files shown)


qunabu:
Task linked: [DEV-2149 Report swallowed docs example failures and guard the runtime imports](https://app.clickup.com/t/9015210959/DEV-2149)

qunabu (approved):

PR #13141, opened 2026-07-29
DEV-2150: Keep the rendered row band consistent when a draw is skipped

demtario:
### Context

The PR closes a window in walkontable's draw cycle where the rendered row band and the DOM disagree, which makes `getCell()` throw `Error: TR was expected to be rendered but is not`.

A full draw advances `rowsRenderCalculator` and `rowFilter` to the new band **before** the `beforeDraw` / `skipRender` gate runs. When the gate cancels the render, the band and filter describe rows that were never rendered, while the TBODY still holds the previous ones. `getCell()` then gates on the new band but resolves the element through the new filter against the old TBODY.

Measured on a master table after a `skipRender` draw:

```
before {first:0,  last:8,  count:9,  trs:9, offset:0}
after  {first:52, last:61, count:10, trs:9, offset:52}
```

`getCell(61)` passes the band check and then finds no TR. With an active selection the throw comes from **inside** `wt.draw()` via `SelectionScanner`, which is the frame shape seen in production.

The fix captures the rendered state and restores it on the skip path — `restoreRenderedState` runs only when `performRedraw === false`, so no other draw path changes behavior. Every full master draw pays one four-property object literal twice, which is the whole cost.

The other candidate window was **ruled out**, not just left alone: `tableView.ts` fast draw is safe because `viewport/calculatorFactory.ts` gates `this.rowsRenderCalculator = renderedRows` behind `!fastDraw`, so a fast draw leaves band, filter and TBODY stale-*consistent*. That was verified byte-identical in the 16.0.1 build that actually produced the reported events.

**What this PR does not do.** The two Sentry events that started this — [HANDSONTABLE-DOCS-21E](https://handsoncode.sentry.io/issues/HANDSONTABLE-DOCS-21E/) and its duplicate [HANDSONTABLE-DOCS-21F](https://handsoncode.sentry.io/issues/HANDSONTABLE-DOCS-21F/) — remain **unreproduced**. They share one trace and one mousedown 30 ms apart and reach the same throw site, but the trigger is unidentified, and their page is a frozen `/docs/16.0/` archive pulling a floating jsDelivr tag, so no release retires them. This fixes a proven instance of the same defect at its source; it is not a claim about those events.

That distinction matters here because the naive fix — returning the `-2` / `-4` sentinels `tableView.ts` already understands — would have silenced a genuinely inconsistent grid state instead of fixing it. Precedent: PR #11555 fixed the equivalent divergence at its source, and the build that threw already contained it.

Two further findings were recorded in `.ai/RENDERING-LIFECYCLE.md` and deliberately left unactioned: a sticky-clone live-band-vs-snapshot divergence, and a `renderedRows.ts` off-by-one.

### How has this been tested?

Test-first, with the failure observed before the fix existed:

```bash
# fix NOT yet applied
npm --prefix handsontable run test:walkontable -- --testPathPattern=drawCycleHooks
# 7 specs, 2 failures
#   Expected 10 to be 9  +  Error: TR was expected to be rendered but is not at MasterTable.getCell
#   Expected function not to throw, but it threw Error: TR was expected to be rendered but is not
#     (thrown from inside wt.draw())
```

Both specs are green with the fix. The second one is the one that matters: it reproduces the production frame shape, where the throw originates inside `wt.draw()` through `SelectionScanner` rather than from an external `getCell()` call.

Regression scope for the only `skipRender` consumer in core:

```bash
npm --prefix handsontable run test:walkontable
npm --prefix handsontable run test:e2e -- --testPathPattern=nestedRows
# green
```

Files changed:

- `handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts`
- `handsontable/src/3rdparty/walkontable/test/spec/table/drawCycleHooks.spec.js`
- `handsontable/src/3rdparty/walkontable/.ai/RENDERING-LIFECYCLE.md`

### Types of changes

### Related issue(s):

1. DEV-2150

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2150

---

> [!NOTE]
> **Medium Risk**
> Touches the master draw cycle, viewport band state, overlay positioning, and `getCell`/selection paths; behavior is gated and heavily tested but errors here affect core grid rendering.
> 
> **Overview**
> Fixes **`Error: TR was expected to be rendered but is not`** when a full master draw advances the rendered row/column band and filters before **`beforeDraw` / `skipRender`** cancels the cell pass, leaving **`getCell`** and selection rendering out of sync with the TBODY.
> 
> On **`performRedraw === false`**, the master now captures pre-draw state and runs **`restoreRenderedStateIfSafe`**: it rolls back render calculators, filters (when safe), and **`correctHeaderWidth`** unless **`renderCycleSeq`** moved (nested **`draw()`** or clone render) or per-axis row/column totals changed (e.g. NestedRows shrink). **`renderCycleSeq`** is incremented on each **`renderCellBand`**; viewport calculators start as **`null`**; overlays treat **`null`** calculators like “nothing rendered” for spreader offset.
> 
> Skipped draws no longer trigger **`refreshAll`** on the 1px border-shift path (which would re-enter **`beforeDraw`**); they re-run fixed-position overlay placement, render selections, and still **`adjustElementsSize`**. Lifecycle docs and **`drawCycleHooks`** specs cover rollback, first-draw skip, and dataset/column edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f3ab4a3055c7c5c4dab083a47d8e1882173c878f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
+++ b/handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
@@ -17,16 +17,17 @@ import type { ColumnsCalculationType, RowsCalculationType } from '../calculator/
 
 /**
  * The state that describes what is actually rendered in the TBODY/THEAD right now: the rendered
- * row/column bands (held by the viewport) and the render filters derived from them, plus the
- * render-cycle sequence number at capture time. Captured before the draw resolves the new band, so
- * the master's `skipRender` gate can put it back — see {@link restoreRenderedStateIfSafe}.
+ * row/column bands (held by the viewport), the render filters derived from them, and the
+ * `correctHeaderWidth` flag that describes the row-header width the DOM was rendered with. Captured
+ * before the draw resolves the new band, so the master's `skipRender` gate can put it back — see
+ * {@link restoreRenderedStateIfSafe}.
  */
 interface RenderedState {
   rowsRenderCalculator: RowsCalculationType | null;
   columnsRenderCalculator: ColumnsCalculationType | null;
   rowFilter: RowFilter | null;
   columnFilter: ColumnFilter | null;
-  renderCycleSeq: number;
+  correctHeaderWidth: boolean;
 }
 
 /**
@@ -164,6 +165,10 @@ function runMasterDrawCycle(table: Table, ctx: DrawContext): void {
 
     table.alignOverlaysWithTrimmingContainer(); // todo It calls method from child class (MasterTable).
     const skipRender: { skipRender?: boolean } = {};
+    // Read the counter as late as possible — right before the hook fires — so the guard in
+    // `restoreRenderedStateIfSafe` asks exactly "did the HOOK render?", and a hypothetical engine-
+    // internal render earlier in this draw cannot silently disable a rollback that is still safe.
+    const renderCycleSeqBeforeHook = wtViewport.renderCycleSeq;
 
     wtSettings.getSetting('beforeDraw', true, skipRender);
     ctx.performRedraw = skipRender.skipRender !== true;
@@ -176,7 +181,7 @@ function runMasterDrawCycle(table: Table, ctx: DrawContext): void {
       // disagree: `getCell` then throws "TR was expected to be rendered but is not" for every row past
       // the end of the stale DOM. Put the pre-draw state back when doing so is provably safe — the
       // guard conditions and the deliberate fallbacks are documented on the helper.
-      restoreRenderedStateIfSafe(table, wtViewport, renderedStateBeforeDraw);
+      restoreRenderedStateIfSafe(table, wtViewport, renderedStateBeforeDraw, renderCycleSeqBeforeHook);
     } else {
       renderCellBand(
         table,
@@ -216,16 +221,29 @@ function runMasterDrawCycle(table: Table, ctx: DrawContext): void {
 
   placeFixedOverlays(table, ctx);
 
-  if (ctx.positionChanged && ctx.performRedraw) {
-    // It refreshes the cells borders caused by a 1px shift (introduced by overlays which add or
-    // remove `innerBorderTop` and `innerBorderInlineStart` CSS classes to the DOM element. This happens
-    // when there is a switch between rendering from 0 to N rows/columns and vice versa).
-    // Gated on `performRedraw`: a draw whose render was skipped must not run this nested draw — the
-    // rolled-back band would fail the nested draw's fast-draw check, downgrading the cheap reposition
-    // to a full render that fires `beforeDraw` a second time and renders the cells the hook just
-    // cancelled. The per-overlay `adjustElementsSize()` calls inside `placeFixedOverlays` already ran,
-    // and nothing rendered, so the master spreader-size pass below has nothing new to apply either.
-    wtOverlays.refreshAll(); // `refreshAll()` internally already calls `refreshSelections()` method
+  if (ctx.positionChanged) {
+    if (ctx.performRedraw) {
+      // It refreshes the cells borders caused by a 1px shift (introduced by overlays which add or
+      // remove `innerBorderTop` and `innerBorderInlineStart` CSS classes to the DOM element. This happens
+      // when there is a switch between rendering from 0 to N rows/columns and vice versa).
+      wtOverlays.refreshAll(); // `refreshAll()` internally already calls `refreshSelections()` method
+    } else {
+      // A skipped render must not run the nested reconciliation draw above (`refreshAll` is
+      // `wot.draw(true)`): the rolled-back band would fail its fast-draw check and escalate it to a
+      // full render that fires `beforeDraw` a second time and renders the cells the hook just
+      // cancelled. But the `innerBorder*` toggle has already shifted the layout by 1px AFTER the
+      // overlay positions were computed, so rerun the fixed-position pass against the post-toggle
+      // layout — it converges, because the second border-state check finds the class already in
+      // place — and render the selections that `refreshAll` would have refreshed.
+      placeFixedOverlays(table, ctx);
+      renderActiveSelections(table, ctx.runFastDraw);
+    }
+
+    // Outside the render gate on purpose: the master hider/spreader size is written ONLY here on the
+    // 1px-shift path (the per-overlay `adjustElementsSize()` calls inside `placeFixedOverlays` size
+    // the clone roots, not the master hider), and the bands it reads are current whether or not the
+    // cell render ran — a skipped draw would otherwise keep a stale scrollbar until an unrelated
+    // full draw.
     wtOverlays.adjustElementsSize();
   } else {
     renderActiveSelections(table, ctx.runFastDraw);
@@ -298,9 +316,9 @@ function buildRenderFilters(table: Table, ctx: DrawContext): { rowFilter: RowFil
 }
 
 /**
- * Captures the rendered row/column bands and the render filters, i.e. everything that describes which
- * source rows and columns the current DOM holds, plus the render-cycle sequence number the viewport
- * reports at this moment. Master-only: the `skipRender` gate it serves is master-only.
+ * Captures the rendered row/column bands, the render filters, and the `correctHeaderWidth` flag,
+ * i.e. everything that describes which source rows and columns the current DOM holds and how it was
+ * rendered. Master-only: the `skipRender` gate it serves is master-only.
  *
  * @param {Table} table The master table.
  * @param {Viewport} wtViewport The viewport that owns the rendered bands.
@@ -312,7 +330,7 @@ function captureRenderedState(table: Table, wtViewport: Viewport): RenderedState
     columnsRenderCalculator: wtViewport.columnsRenderCalculator,
     rowFilter: table.rowFilter,
     columnFilter: table.columnFilter,
-    renderCycleSeq: wtViewport.renderCycleSeq,
+    correctHeaderWidth: table.correctHeaderWidth,
   };
 }
 
@@ -323,24 +341,31 @@ function captureRenderedState(table: Table, wtViewport: Viewport): RenderedState
  * a blocked rollback is never worse than what shipped before the rollback existed.
  *
  * The guards:
- * - The viewport's render-cycle counter must not have moved since the pre-draw capture. A hook that
- *   renders (a nested master `draw()`, or a draw of any overlay clone — the clones share this
- *   viewport) has put a NEWER band into the DOM; rolling the shared band back under it would create
- *   the very band/DOM divergence this rollback exists to prevent.
- * - The captured filters' build-time totals must still match the current `totalRows`/`totalColumns`.
- *   When the dataset shrank before the skipped draw (NestedRows cancels the render right after its
- *   row removal), the captured band names rows that no longer exist — the this-draw band, capped at
- *   the new totals, is the correct description. (A dataset that GREW before a skipped draw keeps the
- *   this-draw band too; that band can then overhang the stale DOM — a pre-existing edge of the
- *   skip contract, unchanged here.)
+ * - The viewport's render-cycle counter must not have moved since it was read right before the
+ *   `beforeDraw` hook fired. A hook that renders (a nested master `draw()`, or a draw of any overlay
+ *   clone — the clones share this viewport) has put a NEWER band into the DOM; rolling the shared
+ *   band back under it would create the very band/DOM divergence this rollback exists to prevent.
+ * - Per axis, the captured filter's build-time total must still match the current
+ *   `totalRows`/`totalColumns`. When the dataset shrank before the skipped draw (NestedRows cancels
+ *   the render right after its row removal), the captured band names rows that no longer exist — the
+ *   this-draw band, capped at the new totals, is the correct description. Each axis decides
+ *   independently, so a column-count change never blocks the row rollback (and vice versa); the
+ *   axes never disagree inside `getCell`, which gates and resolves each axis on its own pair.
+ *   (A dataset that GREW before a skipped draw keeps the this-draw band too; that band can then
+ *   overhang the stale DOM — a pre-existing edge of the skip contract, unchanged here.)
  *
  * The restore itself is asymmetric by design:
- * - The render calculators are restored unconditionally — `null` on a skipped first draw correctly
- *   reports "nothing rendered", so `Table#getCell` answers with its out-of-viewport exit codes.
+ * - The render calculators are restored — `null` on a skipped first draw correctly reports
+ *   "nothing rendered", so `Table#getCell` answers with its out-of-viewport exit codes (and the
+ *   overlays' `applyToDOM` treats a `null` calculator as the nothing-rendered spreader offset).
  * - The filters are restored only when the captured ones are non-null; on a skipped first draw the
  *   just-built filters are kept instead, because filter consumers (`getRowHeader`, `getTrForRow`,
  *   the selection scanner's header loop) read `rowFilter!` unguarded and must never see `null`
  *   after a completed draw.
+ * - `correctHeaderWidth` is restored whenever no render happened, regardless of the totals gates:
+ *   the flag describes the row-header width the DOM was rendered with, and that DOM did not change.
+ *   Leaving it advanced would make the next draw see "no change" and never re-render the corrected
+ *   header width.
  * - The visible-row/column calculators are deliberately NOT restored: they describe what the user
  *   can see (scroll position), not what the DOM holds. After a skipped draw the visible band may
  *   therefore extend past the rendered band — unlike a fast draw, which guarantees the visible band
@@ -349,24 +374,37 @@ function captureRenderedState(table: Table, wtViewport: Viewport): RenderedState
  * @param {Table} table The master table.
  * @param {Viewport} wtViewport The viewport that owns the rendered bands.
  * @param {RenderedState} state The state captured before the draw.
+ * @param {number} renderCycleSeqBeforeHook The render-cycle counter read right before the
+ *   `beforeDraw` hook fired.
  */
-function restoreRenderedStateIfSafe(table: Table, wtViewport: Viewport, state: RenderedState): void {
+function restoreRenderedStateIfSafe(
+  table: Table,
+  wtViewport: Viewport,
+  state: RenderedState,
+  renderCycleSeqBeforeHook: number,
+): void {
   const { wtSettings } = table;
-  const renderHappenedSinceCapture = wtViewport.renderCycleSeq !== state.renderCycleSeq;
-  const totalsChangedSinceCapture =
-    (state.rowFilter !== null && state.rowFilter.total !== wtSettings.getSetting<number>('totalRows')) ||
-    (state.columnFilter !== null && state.columnFilter.total !== wtSettings.getSetting<number>('totalColumns'));
 
-  if (renderHappenedSinceCapture || totalsChangedSinceCapture) {
+  if (wtViewport.renderCycleSeq !== renderCycleSeqBeforeHook) {
     return;
   }
 
-  wtViewport.rowsRenderCalculator = state.rowsRenderCalculator;
-  wtViewport.columnsRenderCalculator = state.columnsRenderCalculator;
+  table.correctHeaderWidth = state.correctHeaderWidth;
 
-  if (state.rowFilter !== null && state.columnFilter !== null) {
-    table.rowFilter = state.rowFilter;
-    table.columnFilter = state.columnFilter;
+  if (state.rowFilter === null || state.rowFilter.total === wtSettings.getSetting<number>('totalRows')) {
+    wtViewport.rowsRenderCalculator = state.rowsRenderCalculator;
+
+    if (state.rowFilter !== null) {
+      table.rowFilter = state.rowFilter;
+    }
+  }
+
+  if (state.columnFilter === null || state.columnFilter.total === wtSettings.getSetting<number>('totalColumns')) {
+    wtViewport.columnsRenderCalculator = state.columnsRenderCalculator;
+
+    if (state.columnFilter !== null) {
+      table.columnFilter = state.columnFilter;
+    }
   }
 }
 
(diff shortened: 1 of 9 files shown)


qunabu:
Task linked: [DEV-2150 Keep the walkontable rendered row band consistent when a draw is skipped](https://app.clickup.com/t/9015210959/DEV-2150)

qunabu (dismissed):

budnix (commented):
The fix does close the crash it targets, and the guard for a hook that renders inside `beforeDraw` is a nice touch. A few things about the rollback that I think are worth a look before merge — mostly around state that is put back but is no longer valid, and one that turns the skipped render into a full one.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -142,14 +160,31 @@ function runMasterDrawCycle(table: Table, ctx: DrawContext): void {
...
+      // index (`Table#getCell` and, through it, the selection render below) gates on the bands but
+      // reads the element through the filters and the TBODY, so leaving them advanced makes those two
+      // disagree: `getCell` then throws "TR was expected to be rendered but is not" for every row past
+      // the end of the stale DOM. Put the pre-draw state back - a skipped render leaves the screen
+      // unchanged, so the rendered state must stay unchanged too, exactly like a fast draw.
+      // Guarded on the state still being the one THIS draw resolved: a `beforeViewRender` hook that
+      // renders (a nested `draw()`) has already put a matching band and DOM in place, and rolling that
+      // back would create the very divergence this rollback exists to prevent.
+      if (isSameRenderedState(captureRenderedState(table, wtViewport), renderedStateThisDraw)) {
+        restoreRenderedState(table, wtViewport, renderedStateBeforeDraw);
```
budnix: Can I ask - what happens when `positionChanged` is true later in this same draw? Line 230 calls `wtOverlays.refreshAll()`, which is `wot.draw(true)`, and that nested draw compares its proposed band against the band we just rolled back. The fast-draw gate fails, so it runs as a full draw instead of a cheap reposition: `beforeDraw` fires a second time, a one-shot skip flag is already spent, and the cells render anyway. Before this change the band was still advanced there, so the nested draw stayed cheap - a test that counts `beforeDraw` calls per `draw()` on a scroll that flips `innerBorderTop` would pin this down.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -263,6 +299,57 @@ function buildRenderFilters(table: Table, ctx: DrawContext): { rowFilter: RowFil
...
+function isSameRenderedState(state: RenderedState, otherState: RenderedState): boolean {
+  return state.rowsRenderCalculator === otherState.rowsRenderCalculator &&
+    state.columnsRenderCalculator === otherState.columnsRenderCalculator &&
+    state.rowFilter === otherState.rowFilter &&
+    state.columnFilter === otherState.columnFilter;
+}
+
+/**
+ * Puts a captured rendered state back, undoing the band and filter advance of a draw that ended up
+ * rendering nothing. The visible-row/column calculators are deliberately NOT restored: they describe
```
budnix: The comment says the skip path behaves like a fast draw, but a fast draw gets that from `areAllProposedVisibleRowsAlreadyRendered`, which guarantees the visible band sits inside the rendered band. Here the rendered band goes back to the old scroll position while the visible band stays at the new one, so nothing keeps that relationship. `getLastVisibleRow()` can then report a row that `getCell()` refuses to return, which quietly breaks any caller looping from first to last visible row.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -263,6 +299,57 @@ function buildRenderFilters(table: Table, ctx: DrawContext): { rowFilter: RowFil
...
+ *
+ * @param {Table} table The master table.
+ * @param {Viewport} wtViewport The viewport that owns the rendered bands.
+ * @returns {RenderedState}
+ */
+function captureRenderedState(table: Table, wtViewport: Viewport): RenderedState {
+  return {
+    rowsRenderCalculator: wtViewport.rowsRenderCalculator,
+    columnsRenderCalculator: wtViewport.columnsRenderCalculator,
+    rowFilter: table.rowFilter,
```
budnix, line 331: On the very first draw `table.rowFilter` and `columnFilter` are still `null`, so a skip on that draw restores `null` - while `setDrawn(true)` still runs at the end of the same cycle. `scanRowsInHeadersRange` and `getTrForRow` read `rowFilter!` with no guard, and the header loop runs even when the rendered row count is 0. That state was not reachable before, since the filters were always rebuilt on the full path. Worth a quick check with a hook that cancels the first render.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -263,6 +299,57 @@ function buildRenderFilters(table: Table, ctx: DrawContext): { rowFilter: RowFil
...
+ * rendering nothing. The visible-row/column calculators are deliberately NOT restored: they describe
+ * what the user can see (scroll position), not what the DOM holds, and they are recomputed on every
+ * draw regardless of the render gate.
+ *
+ * @param {Table} table The master table.
+ * @param {Viewport} wtViewport The viewport that owns the rendered bands.
+ * @param {RenderedState} state The state captured before the draw.
+ */
+function restoreRenderedState(table: Table, wtViewport: Viewport, state: RenderedState): void {
+  wtViewport.rowsRenderCalculator = state.rowsRenderCalculator;
```
budnix: When the skipped draw follows a row removal - which is exactly what NestedRows does in `#onFilterData` - this puts back a band that is wider than the new `totalRows`. `getLastRenderedRow()` then reports a row that no longer exists, so `getCell()` returns a stale TR instead of the out-of-range exit code it used to give, and the selection scanner paints highlight classes over removed rows. Did you get a chance to try the skip path right after an `alter('remove_row', ...)`?

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -263,6 +299,57 @@ function buildRenderFilters(table: Table, ctx: DrawContext): { rowFilter: RowFil
...
+/**
+ * Compares two captured rendered states by identity. The bands and the filters are replaced with fresh
+ * objects on every full draw, so reference equality is exactly the "nothing has replaced them since"
+ * question the `skipRender` rollback needs to answer.
+ *
+ * @param {RenderedState} state The state to compare.
+ * @param {RenderedState} otherState The state to compare it against.
+ * @returns {boolean}
+ */
+function isSameRenderedState(state: RenderedState, otherState: RenderedState): boolean {
```
budnix: May I suggest a closer look at the clone case - `Clone` reuses the master's viewport (`core/clone.ts:34`), so the two render calculators compared here are shared with every overlay, but each clone keeps its own `rowFilter`/`columnFilter`. A hook that redraws only the overlays and then sets `skipRender` leaves all four master references untouched, so this guard passes and the shared band is rolled back under a clone whose filters and DOM already moved on. With `fixedRowsTop` that lands back on the same `TR was expected to be rendered but is not`.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -142,14 +160,31 @@ function runMasterDrawCycle(table: Table, ctx: DrawContext): void {
     table.tableOffset = table.deps.geometryReader.offset(table.TABLE);
 
     const filters = buildRenderFilters(table, ctx);
+    // The band + filters this draw resolved, i.e. what the render below is about to put in the DOM.
+    const renderedStateThisDraw = captureRenderedState(table, wtViewport);
```
budnix: `buildRenderFilters` assigns `table.rowFilter`/`table.columnFilter` and returns those same two objects, so `renderedStateThisDraw.rowFilter === filters.rowFilter` always holds. You could drop this capture and compare against `filters` plus the two calculators directly - one less object per full draw, and the guard gets easier to follow.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
demtario: Confirmed — with the rolled-back band the nested `draw(true)` failed the fast-draw check and escalated to a full render. Fixed in 5d866d5: the `positionChanged` reconciliation is now gated on `performRedraw`, so a skipped draw never runs `refreshAll()`. Added the spec you suggested — `beforeDraw` fires exactly once per `draw()` across an `innerBorderTop` flip.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
demtario: Right — the fast-draw equivalence didn't hold; only `areAllProposedVisibleRowsAlreadyRendered` guarantees visible ⊆ rendered. Rewrote the comment (and RENDERING-LIFECYCLE.md) in 5d866d5 to state the real contract: after a skipped draw the visible band may extend past the rendered band and `getCell` answers those rows with exit codes. That's inherent to skipping after a scroll — the DOM can't match the new scroll position; the exit codes replace the pre-PR throw.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
demtario: Confirmed — a cancelled first render restored `null` filters while `setDrawn(true)` ran. Fixed in 5d866d5: the rollback restores the calculators (null correctly reports "nothing rendered", so `getCell` returns exit codes) but keeps the just-built filters when the captured ones are null, so no unguarded `rowFilter!` site can see null after a completed draw. Covered by a new spec (first-draw skip + column headers + selection).

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
demtario: Confirmed — the restored band named removed rows and `getCell` returned their stale TRs. Fixed in 5d866d5: the rollback is gated on the captured filters' build-time `total` matching the current `totalRows`/`totalColumns` (the totals must be read off the captured filter — NestedRows shrinks them before the skipped draw starts, so comparing totals captured at the top of the same draw would always pass). On mismatch the fresh band, capped at the new totals, stays — the pre-PR behavior that worked here. New spec covers shrink + skip.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
demtario: Confirmed the hole — the identity guard only saw the master's four references. Replaced it in 5d866d5 with a render-cycle counter on the shared `Viewport`, bumped by every `renderCellBand` (master or clone), so any render inside the hook blocks the rollback. For the clone-mid-hook scenario that means falling back to the pre-PR state — the shared band can't describe both a stale master DOM and an advanced clone DOM, so we deliberately keep the old behavior there rather than trade one "TR was expected" for another.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
demtario: Gone in 5d866d5 — the counter guard replaced the identity comparison, so `renderedStateThisDraw` and `isSameRenderedState` are deleted entirely.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -188,10 +216,15 @@ function runMasterDrawCycle(table: Table, ctx: DrawContext): void {
 
   placeFixedOverlays(table, ctx);
 
-  if (ctx.positionChanged) {
+  if (ctx.positionChanged && ctx.performRedraw) {
```
budnix: Curious whether this gate is a bit too wide - it drops `wtOverlays.adjustElementsSize()` as well as `refreshAll()`. The per-overlay `adjustElementsSize()` calls inside `placeFixedOverlays` size each clone, but the master hider width and height are written only by `spreaderSize.adjustElementsSize()`, which nothing else on the skip path calls. So after a skipped draw that follows a row removal the hider keeps its old height, and the vertical scrollbar stays longer than the data until some later full draw. Keeping `adjustElementsSize()` outside the `performRedraw` check would fix that without bringing the nested draw back - the new spec only counts `beforeDraw` calls, so it stays green.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -188,10 +216,15 @@ function runMasterDrawCycle(table: Table, ctx: DrawContext): void {
...
     // It refreshes the cells borders caused by a 1px shift (introduced by overlays which add or
     // remove `innerBorderTop` and `innerBorderInlineStart` CSS classes to the DOM element. This happens
     // when there is a switch between rendering from 0 to N rows/columns and vice versa).
+    // Gated on `performRedraw`: a draw whose render was skipped must not run this nested draw — the
+    // rolled-back band would fail the nested draw's fast-draw check, downgrading the cheap reposition
+    // to a full render that fires `beforeDraw` a second time and renders the cells the hook just
+    // cancelled. The per-overlay `adjustElementsSize()` calls inside `placeFixedOverlays` already ran,
+    // and nothing rendered, so the master spreader-size pass below has nothing new to apply either.
     wtOverlays.refreshAll(); // `refreshAll()` internally already calls `refreshSelections()` method
```
budnix: The other half of the same gate: `resetFixedPosition` inside `placeFixedOverlays` has already added or removed `innerBorderTop` on the master holder, so the 1px shift is in the DOM by the time we get here. Before this change `refreshAll()` repositioned the clones for the new offset; now nothing does. With `fixedRowsTop` and column headers that leaves the frozen rows one pixel off from the body until an unrelated full draw. Would be good to assert the clone offset in that new spec, not just the `beforeDraw` count.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -263,6 +297,79 @@ function buildRenderFilters(table: Table, ctx: DrawContext): { rowFilter: RowFil
...
+ *
+ * @param {Table} table The master table.
+ * @param {Viewport} wtViewport The viewport that owns the rendered bands.
+ * @param {RenderedState} state The state captured before the draw.
+ */
+function restoreRenderedStateIfSafe(table: Table, wtViewport: Viewport, state: RenderedState): void {
+  const { wtSettings } = table;
+  const renderHappenedSinceCapture = wtViewport.renderCycleSeq !== state.renderCycleSeq;
+  const totalsChangedSinceCapture =
+    (state.rowFilter !== null && state.rowFilter.total !== wtSettings.getSetting<number>('totalRows')) ||
```
budnix: Can I ask why the two axes share one exit? A column change (hide a column, `alter('insert_col')`, a filter update) makes the column half true, and then the row rollback is skipped as well - even though `totalRows` never moved. That is exactly the case this PR fixes, so hiding a column during a skipped draw would still throw `TR was expected to be rendered but is not`. Gating rows on `totalRows` and columns on `totalColumns` separately keeps each axis independent.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -263,6 +297,79 @@ function buildRenderFilters(table: Table, ctx: DrawContext): { rowFilter: RowFil
...
+ * @param {Viewport} wtViewport The viewport that owns the rendered bands.
+ * @returns {RenderedState}
+ */
+function captureRenderedState(table: Table, wtViewport: Viewport): RenderedState {
+  return {
+    rowsRenderCalculator: wtViewport.rowsRenderCalculator,
+    columnsRenderCalculator: wtViewport.columnsRenderCalculator,
+    rowFilter: table.rowFilter,
+    columnFilter: table.columnFilter,
+    renderCycleSeq: wtViewport.renderCycleSeq,
```
budnix: `table.correctHeaderWidth` is set earlier in this draw, before the `beforeDraw` gate, and it forces a full draw for that one pass. When the render is skipped the row header never gets the corrected width, but the flag stays `true`, so the next draw sees no change and never downgrades itself to a full draw - the header stays 1px off. Worth capturing it here next to the filters so the rollback puts it back too.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -78,6 +94,9 @@ function runMasterDrawCycle(table: Table, ctx: DrawContext): void {
   const { wtSettings } = table;
   const wtOverlays = table.deps.getWtOverlays();
   const wtViewport = table.deps.getWtViewport();
+  // What is on the screen before this draw touches anything. Restored (when safe — see
+  // `restoreRenderedStateIfSafe`) if the `beforeDraw` hook cancels the cell render below.
+  const renderedStateBeforeDraw = captureRenderedState(table, wtViewport);
```
budnix, line 100: Did this end up at the top of the function on purpose? The counter is read before `wtOverlays.beforeDraw()`, `prepareHeaderBorders()` and `alignOverlaysWithTrimmingContainer()` run, so anything in those that renders a clone band bumps it and quietly turns the rollback off - with no test that would notice. Reading `renderCycleSeq` right before the `beforeDraw` setting call would keep the guard about the hook only, which is what the JSDoc describes.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -263,6 +297,79 @@ function buildRenderFilters(table: Table, ctx: DrawContext): { rowFilter: RowFil
...
+  const totalsChangedSinceCapture =
+    (state.rowFilter !== null && state.rowFilter.total !== wtSettings.getSetting<number>('totalRows')) ||
+    (state.columnFilter !== null && state.columnFilter.total !== wtSettings.getSetting<number>('totalColumns'));
+
+  if (renderHappenedSinceCapture || totalsChangedSinceCapture) {
+    return;
+  }
+
+  wtViewport.rowsRenderCalculator = state.rowsRenderCalculator;
+  wtViewport.columnsRenderCalculator = state.columnsRenderCalculator;
```
budnix: On a skipped first draw this puts `null` back into both calculators while `setDrawn(true)` still runs at the end of the cycle. `inlineStartOverlay.applyToDOM()` throws `Incorrect value of the columnsRenderCalculator` when `startPosition` is not a number and `totalColumns` is not 0, and `stickyScrollStrategy.deactivate()` calls it right after a `refreshAll()` that may not repopulate them. It needs a hook that keeps skipping, so it is a narrow path - worth a quick check that drawn-with-null-calculators cannot reach that call.

budnix (commented):
Second pass over `5d866d5`. The four points from the first round look handled. Most of what is below is about the new `performRedraw` gate and the new guards, plus one piece of pre-draw state that is still not captured.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
demtario: Fixed in f3ab4a3 — `adjustElementsSize()` is hoisted outside the `performRedraw` check, so the master hider/spreader size is written on the skipped draw too (the bands it reads are current either way — on the removal path the rollback is blocked, so they're the fresh ones). One scope note: for skips where `positionChanged` stays false the hider was equally stale before this PR — that half is pre-existing, and this change fixes it as a side effect since the sizing now runs on every `positionChanged` draw.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
demtario: Fixed in f3ab4a3 — instead of `refreshAll()`, a skipped draw with `positionChanged` reruns the fixed-position pass against the post-toggle layout (it converges: the second border-state check finds the class already in place) and renders the active selections. No render, no second `beforeDraw`. On the spec: in this harness (element mode) `resetFixedPosition` resolves to a transform reset, so the 1px isn't pixel-observable — the spec pins the reconciliation rerun itself (`resetFixedPosition` called twice on the skipped draw); the pixel-level benefit applies in window mode.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
demtario: Fixed in f3ab4a3 — each axis decides independently: rows gated on `totalRows`, columns on `totalColumns`, the render counter stays shared. The hybrid states are coherent because `getCell` gates and resolves each axis on its own band+filter pair. New spec: column shrink + row scroll + skip — row band rolls back and every rendered row still resolves to an element.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
demtario: Added in f3ab4a3 — the flag is captured with the rest of the rendered state and restored whenever no render happened since the capture (regardless of the totals gates — the DOM header width didn't change either way). One note: the stale flag was pre-existing behavior — a skipped draw left it flipped before this PR too — but it belongs in the rollback's contract, so it's in. Spec included.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
demtario: Moved in f3ab4a3 — the counter is now read into a local right before the `beforeDraw` setting call, so the guard asks exactly "did the hook render?". The band/filter capture stays at the top (it must precede `createCalculators`). Checked the in-between steps (`wtOverlays.beforeDraw()`, `prepareHeaderBorders`, `beginDrawLayout`, `alignOverlaysWithTrimmingContainer`) — none render a band today, so this is a precision fix with no current repro, but the JSDoc and the code now agree.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
demtario: Fixed in f3ab4a3 — all three overlay `applyToDOM` methods treat a `null` render calculator like `total === 0` (spreader offset `0` — nothing is rendered, which is what null means), so the sticky-scroll deactivation path can't throw on the drawn-but-never-rendered state. This also surfaced a latent issue: the render calculators were `undefined` (not `null`) before the first render — they're now initialized to `null` in the `Viewport` constructor to match their declared type, and the one characterization spec pinning the `undefined` was updated. First-draw spec now calls `wtOverlays.applyToDOM()` directly and asserts no throw.

budnix (approved):

PR #13150, opened 2026-08-03
issue-4363: Stop grid cell CSS leaking into tables nested in cells

### LINKED ISSUE #4363: Handsontable is leaking CSS styles
### Description
Handsontable is leaking CSS styles. 

### Steps to reproduce
If I insert for example a `<table>` inside a cell, all of its `<td>` tags will be polluted with handsontable's style. In my example, the uib-datepicker angular directive.

### Demo
handsontable_leak

### Your environment
* Handsontable version: v0.31.2
* Browser Name and version: Chrome 59.0.3071.115
* Operating System: Mac OS Sierra (10.12.5)

sl01k:
### Context

The PR fixes [#4363](https://github.com/handsontable/handsontable/issues/4363): Handsontable's cell styling leaks into a `<table>` (or any element) a user renders inside a cell. The grid's base rules were bare descendant selectors (`.handsontable td`, `.handsontable th`, the `_border-radius` corner rules, the `box-sizing`/`col`/`a` normalizers), so a nested `<table>`'s `<td>`s inherited grid padding, borders, background, `box-sizing` and corner radius — e.g. a datepicker or any custom-rendered table rendered wrong.

The fix scopes every cell/header/border rule to the grid's own `table.htCore` tables using `:where()` child-combinator boundaries. `:where()` contributes zero specificity, so the resolved cascade for a normal grid is byte-for-byte identical (verified: 0 pixels changed across all three themes), while the `>` combinators stop the rules ever reaching a nested table's cells. Rules that only apply to grid content nested in a cell (`box-sizing`/`min-height`/`a`) use a specificity-neutral `:not(:where(table.htCore > tbody > tr > td *))` guard so grid chrome (editors, menus, license/menu links) still gets normalized.

Scoping the cell CSS to `table.htCore` surfaced a second, subtle problem: `stylesHandler`'s box-sizing probe built a **class-less** `<table>`, so after scoping it read `content-box`, flipping `areCellsBorderBox()` to `false` and mis-measuring oversized-row heights — the row-header overlay drifted ~2px under a tall custom-rendered row. Fixed by giving the probe the `htCore` class so it represents a real grid cell.

### How has this been tested?

- New Playwright E2E `tests/e2e/nested-table-css-leak.spec.ts` (× 3 themes): grid styling does not leak into the nested table, the grid still styles its own cells, and row headers stay aligned with an auto-expanded custom-content row. Verified the specs fail without the fix (leak: `border-box`; alignment: 2px drift) and pass with it.
  - `cd tests && npx playwright test e2e/nested-table-css-leak.spec.ts` (9/9 across main/horizon/classic)
- `npm run test:unit --prefix handsontable --testPathPattern=stylesHandler` — 30/30.
- `npm run test:types --prefix handsontable` — passes.
- Visual regression: headless before/after screenshot diff of a full grid (headers, row headers, striping, borders, frozen row+column, selection range, dropdown, checkbox) — **0 px changed** across all three themes.
- Performance (CDP `Performance.getMetrics`, 5000×10 scroll): style-recalc count identical (136), recalc/layout/script times flat (within noise); no `:has()`-style invalidation growth.

### Types of changes

### Related issue(s):

1. #4363

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Touches foundational base/border-radius SCSS and row-height measurement probes across all themes; regression risk is real but heavily guarded by new E2E and claimed zero-pixel visual parity.
> 
> **Overview**
> Fixes **#4363** by scoping Handsontable’s cell/header rules to the grid’s own tables (`table.htCore` and direct-child ghost measurement tables) using `:where()` plus `>` combinators, so padding, borders, backgrounds, corner radius, and `box-sizing`/`min-height` normalizers no longer hit a user-rendered `<table>` (or other DOM) inside a cell. Normalize rules use `:not(:where(…))` carve-outs for checkbox inputs and `td.htCustomMenuRenderer` (Filters/menu UI).
> 
> The same scoping exposed measurement bugs: **`stylesHandler`**’s probe now uses class **`htCore`**, **`GhostTable`** builds headers under a real `<tr>`, and ghost-table box-sizing scope uses `&:where(.htGhostTable) > table` so NestedHeaders ghosts on `<body>` still get grid normalize under host `* { box-sizing: border-box }`.
> 
> Adds Playwright coverage in `nested-table-css-leak.spec.ts`, an **AGENTS.md** pitfall for probe/ghost DOM parity, and changelog **13150**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 10f54df5d7b8773022eec43cf1c7281fd591ae3c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu (changes requested):
tests e2e are failing, seems like this change amends the values in css 
needs @demtario and @adrianspdev approve

tests/fixtures/demo/nested-table.html
demtario, line 1: I belive this file should not be commited. It can be added to this PR as attachment, but should be removed from code changes

tests/fixtures/pages/NestedTablePage.ts
demtario, line 1: As above, this file should not be commited

demtario (changes requested):

tests/fixtures/demo/nested-table.html
adrianspdev: Removed in 614f031ed8 — the spec no longer needs a committed fixture page: it serves its HTML inline via `page.route()` and loads the built dist/styles from the static server.

tests/fixtures/pages/NestedTablePage.ts
adrianspdev: Removed in 614f031ed8 — the page object was dropped together with the fixture; the spec keeps its locator/measure helpers locally and is fully self-contained now.

handsontable/src/styles/base/_base.scss
```diff
@@ -547,17 +565,20 @@
...
-    table,
-    tbody,
-    thead,
-    input,
-    textarea,
-    div {
+    // Box-sizing normalize — grid chrome only. Scoped so it never forces `content-box` onto a
+    // table/div/input a user renders inside a cell, which would fight a host `border-box` (#4363).
+    table.htCore,
+    :where(.htGhostTable) > table,
```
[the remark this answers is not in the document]
adrianspdev: Fixed in e7f7a4673c. You are right about the combinator: a ghost container carries `handsontable` and `htGhostTable` on the same element, and the NestedHeaders ghost is appended to `<body>`, so `.handsontable :where(.htGhostTable)` never matched it. That selector is now the compound form `&:where(.htGhostTable) > table`, which keeps the former bare `table` selector specificity (0,1,1).

Verified in a real browser under a host `* { box-sizing: border-box }` reset: the NestedHeaders ghost table read `border-box` before and reads `content-box` after; the core GhostTable (its container sits inside the grid root, so the descendant form did match) was and stays `content-box`. A new Playwright case covers it - it drives a real nestedHeaders grid and snapshots the plugin-built ghost table computed box-sizing at insertion; confirmed red without the CSS fix.

demtario (dismissed):

handsontable/src/styles/base/_base.scss
```diff
@@ -3,6 +3,27 @@
...
+  $grid-head-row: "table.htCore > thead > tr, #{$ghost-table} > thead > tr";
+  $grid-body-row: "table.htCore > tbody > tr, #{$ghost-table} > tbody > tr";
+  $grid-body: "table.htCore > tbody, #{$ghost-table} > tbody";
+  $grid-sections: "table.htCore > thead, table.htCore > tbody, #{$ghost-table} > thead, #{$ghost-table} > tbody";
+  // Anything a user renders inside a body cell. Used with `:not(:where(...))` to keep grid
+  // normalize rules off user content without changing their specificity (`:where()` = 0) (#4363).
+  // The checkbox renderer's `<input>` is the one element the grid itself renders inside a cell,
+  // so it is carved out here — it must keep the `input` normalize (`box-sizing`/`min-height`)
+  // or host-page rules like `input { min-height: 40px }` inflate grid checkboxes.
+  $user-cell-content: "table.htCore > tbody > tr > td *:not(.htCheckboxRendererInput), #{$ghost-table} > tbody > tr > td *:not(.htCheckboxRendererInput)";
```
budnix: This exclusion drops the normalize from every element inside any body `td`, and the Filters UI lives inside `td.htCustomMenuRenderer` of the dropdown-menu grid. I measured it in a browser with the same host rule this PR's own fixture uses (`input { min-height: 40px }`): the filters search box goes 18px → 40px, OK/Cancel 20px → 40px, and the menu widens 441px → 483px. On `develop` the menu was immune, so this is a new leak in the other direction. The comment just above also says the checkbox `<input>` is the one element the grid renders inside a cell — the Filters UI is the counter-example. Would be good to verify a menu-cell carve-out keeps the chrome normalized without letting the grid rules back into user cells.

```suggestion
  $user-cell-content: "table.htCore > tbody > tr > td:not(.htCustomMenuRenderer) *:not(.htCheckboxRendererInput), #{$ghost-table} > tbody > tr > td:not(.htCustomMenuRenderer) *:not(.htCheckboxRendererInput)";
```

handsontable/src/styles/base/_base.scss
```diff
@@ -598,16 +624,19 @@
...
+      > td,
+      > th {
         display: none;
       }
     }
 
-    a {
+    // Grid-chrome links only (license notice, menus). A link a user renders inside a cell keeps
+    // its own color (#4363). All `<a>` in cells are user content — the grid renders none itself.
+    a:not(:where(#{$user-cell-content})) {
```
budnix, line 650: The comment says the grid renders no `<a>` inside cells, but Filters renders `.htUISelectAll a` and `.htUIClearAll a` inside the menu's cells. Those two still look right only because `_filters.scss` gives them their own color, so this is a comment fix, not a code fix. Worth correcting so the next reader doesn't build on the claim.

tests/e2e/nested-table-css-leak.spec.ts
```diff
@@ -0,0 +1,317 @@
...
+        ['Carol', 'plain', true],
+      ],
+      colHeaders: ['Name', 'Nested', 'Done'],
+      columns: [
+        { renderer: testIdRenderer },
+        { renderer: testIdRenderer },
+        { type: 'checkbox' },
+      ],
+      rowHeaders: true,
+      rowHeights: 90,
```
budnix: `rowHeights: 90` pins every row height, so nothing auto-expands here — the test name and the comments at lines 11, 76 and 97 all say "auto-expand", but what actually runs is the oversized-row path. Did this end up fixed on purpose to keep the measurement stable across themes? Either way the name and comments should match what the fixture does.

tests/e2e/nested-table-css-leak.spec.ts
```diff
@@ -0,0 +1,317 @@
...
+    const ownCell = cell(1, 0);
+
+    expect(await computedStyle(ownCell, 'box-sizing')).toBe('border-box');
+    expect(await computedStyle(ownCell, 'border-bottom-style')).toBe('solid');
+  });
+
+  test('row headers stay aligned with an auto-expanded custom-content row', async() => {
+    // Row 0 is taller than the default (it holds the nested table), forcing the
+    // oversized-row measurement path. The row-header overlay must track it.
+    expect(await rowHeaderBottomGap(0)).toBeLessThanOrEqual(1);
```
budnix, line 243: The tolerance is the same size as the bug it guards — a wrong `areCellsBorderBox()` reading shifts the box model by 1px. Both the master and the inline-start overlay read that same flag, so if it flips they can both be wrong by the same amount and still line up within 1px. I'm not sure this goes red if the `stylesHandler` `htCore` fix is reverted — worth a quick check, and asserting the probe's own reading would be a firmer guard.

handsontable/AGENTS.md
```diff
@@ -69,6 +69,7 @@ Gotcha: Filters `conditionCollection` uses physical indexes, `getDataAtCol()` us
...
 - **`arr.push(...largeArray)`**: Causes stack overflow with 10k+ elements. Use `forEach` loop instead.
+- **Off-DOM measurement tables (probes/ghosts) must mimic the real grid DOM exactly.** Cell and header styling is scoped with child combinators to the grid's own tables — `table.htCore` and any table that is a **direct child of an `.htGhostTable` container** (the `$ghost-table` scope in `src/styles/base/_base.scss`, issue #4363). Consequences for any table built for `getComputedStyle`/`getBoundingClientRect` measurement: (1) put it directly inside an `.htGhostTable`-classed container (then no `htCore` class is needed — deliberately, since `.handsontable .htCore` sets `width: 0; table-layout: fixed`, which breaks content-driven width measurement), or give it the `htCore` class for a fixed-layout probe like `stylesHandler`'s; and (2) build real `tr` chains — a `<th>` appended straight into `<thead>` matches no scoped rule and measures unstyled. This bit three times in one PR: the `stylesHandler` box-sizing probe, the core GhostTable header row, and the NestedHeaders ghost table.
```
budnix: This says a ghost table needs no `htCore` class "deliberately", but `createTable()` at `handsontable/src/utils/ghostTable.ts:506` copies `hot.table.className`, so the core ghost tables do carry `htCore`. They survive because horizontal mode sets `table-layout: auto; width: auto` inline (lines 497-498), not because the class is missing — only the NestedHeaders ghost is class-less. Worth fixing here, since the same wrong picture is what produced the two comments in `_base.scss`.

handsontable/src/styles/base/_base.scss
adrianspdev: Confirmed in a browser before changing anything — with `input { min-height: 40px }` on the host page, the filters search box **and** both action-bar buttons all read `min-height: 40px` / height 40px. Taken the carve-out, named for reuse:

```scss
$user-cell: "td:not(.htCustomMenuRenderer)";
$user-cell-content: "table.htCore > tbody > tr > #{$user-cell} *:not(.htCheckboxRendererInput), …";
```

Excluding the `td` rather than the individual elements matters: the value list inside that cell is a nested Handsontable, and this keeps its whole subtree normalized too. After the fix the three inputs read `min-height: 0px`, heights 34 / 38 / 27 px (main / horizon / classic).

Covered by a new leg in `nested-table-css-leak.spec.ts` (opens the dropdown menu, asserts `min-height: 0px`, then removes the host `<style>` and requires the heights to be unchanged — so the assertion carries no hardcoded pixel bound). It fails on the pre-fix CSS.

One thing that did not reproduce: menu width stayed 217px before and after, on all three themes. The other in-cell chrome the exclusion also caught — `.htAutocompleteArrow` and `.ht-multi-select-chips-container` — sets no padding or border, so `box-sizing` is a no-op for both; left them out of the carve-out deliberately.

handsontable/src/styles/base/_base.scss
adrianspdev: Right, and it is a comment fix only — the code side is now covered by the `td.htCustomMenuRenderer` exclusion in `$user-cell`, so those two links are outside the user-content set again. Comment reworded to say the grid renders no `<a>` into a DATA cell, name the Filters links as the menu-cell case, and note that `_filters.scss` colors them explicitly.

tests/e2e/nested-table-css-leak.spec.ts
adrianspdev: Not deliberate — `rowHeights: 90` pinned every row, so nothing measured what the name claimed. Dropped the setting: row 0 now auto-expands to the nested table's height. Added a non-vacuity guard so a future change cannot quietly flatten it again:

```ts
expect(tallRow!.height).toBeGreaterThan(normalRow!.height * 2);
```

The test name, the three comments you flagged, and the file docblock all match the fixture now. Side effect worth noting: with the pin removed, the row-header alignment assertion actually goes red when the `stylesHandler` probe fix is reverted (2px gap) — it did not before.

tests/e2e/nested-table-css-leak.spec.ts
adrianspdev: Checked, and you are right that the alignment test alone did not carry it. Added a direct assertion on the probe:

```ts
expect(probe.borderBox).toBe(true);                    // hot.stylesHandler.areCellsBorderBox()
expect(probe.borderBottomWidth).not.toBe('0px');
// the probe must agree with a rendered cell — it exists to stand in for one
expect(await computedStyle(cell(1, 0), 'border-bottom-width')).toBe(probe.borderBottomWidth);
```

Verified by patching `table.className = 'htCore'` back to `''` in the built bundle: the new test goes red (`borderBox` reads `false`). The alignment test now goes red on that same revert too, but only after `rowHeights: 90` came out of the fixture — with the pin in place it stayed green, which is the case you suspected.

handsontable/AGENTS.md
adrianspdev: Correct on all three points — `createTable()` copies `hot.table.className` (called with it at `ghostTable.ts:129`, `:153`, `:183`, applied at `:506`), so the core ghosts do carry `htCore`, and horizontal mode survives on the inline `table.style.tableLayout = 'auto'; table.style.width = 'auto'` at `:497-498`, not on a missing class. Bullet rewritten: both scopes are valid, core `GhostTable` satisfies both, only the NestedHeaders ghost is class-less, and a content-width probe that inherits `htCore` without that inline override measures 0.

budnix (approved):

qunabu (approved):

Issue #12366, opened 2026-04-16
[Bug] intl-date renders HH:mm:ss but provides no way to input time — renderer/editor inconsistency


yamaaaaaa31:
### Describe the bug

The `intl-date` cell type has an inconsistency between its renderer and 
its editor/validator when time components (`hour`/`minute`/`second`) are 
specified in `dateFormat`:

- **Renderer**: outputs `HH:mm:ss` (e.g. `2024/03/15 00:00:00`), proving 
  the display layer operates on datetime.
- **Editor**: provides only a date picker. There is no UI to input 
  hours, minutes, or seconds.
- **Validator**: rejects source values that contain time 
  (`2024-03-15 09:00:00`, `2024-03-17 23:59:59`), producing `#bad-value#`.

Result: users see `HH:mm:ss` in the cell but have no way to enter or 
preserve that time. The time portion is hardcoded to `00:00:00` and 
cannot be changed.

### Observed output

| Source value              | Rendered              | Status       |
|---------------------------|-----------------------|--------------|
| `'2024-03-15'`            | `2024/03/15 00:00:00` | ✓ accepted   |
| `'2024-03-16 09:00:00'`   | `#bad-value#`         | ✗ rejected   |
| `'2024-03-17 23:59:59'`   | `#bad-value#`         | ✗ rejected   |

Additionally: even for the accepted date-only row, the displayed 
`00:00:00` cannot be changed through the editor — it opens a date 
picker only.

### Why this is a bug

1. **Renderer/editor asymmetry**: The renderer produces `HH:mm:ss` 
   output. But the editor never exposes time input, and the validator 
   rejects any source that carries time. The displayed value can never 
   reflect a real input.

2. **Documentation contradiction**: The docs state source data "must be 
   in ISO 8601 date format." ISO 8601 itself covers both `YYYY-MM-DD` 
   and `YYYY-MM-DDTHH:mm:ss`; the latter is rejected despite the claim 
   of ISO 8601 compliance.

3. **Configuration accepted but non-functional**: `hour`, `minute`, 
   `second`, `timeStyle`, `hour12`, `hourCycle`, `fractionalSecondDigits` 
   are all listed as valid `dateFormat` options in the documentation. 
   Specifying them produces a column where time is rendered but cannot 
   be entered or preserved. Either these options should work end-to-end, 
   or they should be rejected at config time.

If datetime were intentionally unsupported, the time components should 
be stripped from `dateFormat` or raise a warning. Instead, they are 
[...]

PR #13151, opened 2026-08-03
DEV-1463: Add datetime and intl-datetime cell types

sl01k:
### Context

The PR adds official `datetime` and `intl-datetime` cell types. `intl-date` (and `date`) are date-only, yet their `dateFormat` option accepts time-related `Intl.DateTimeFormat` options (`hour`, `minute`, `second`, `timeStyle`, `hour12`, `hourCycle`, `fractionalSecondDigits`). Users configure them, see `00:00:00` rendered, and expect full datetime support that does not exist. Reported in #12366.

This PR closes both action items from DEV-1463:

1. **Documentation clarification** — the `dateFormat` JSDoc (`metaSchema.ts`) and the date cell type guide now state that time-related options only affect display and always render midnight for date-only source data.
2. **New cell types** — `datetime` and `intl-datetime` mirror the existing `time`/`intl-time` family: a native, zero-dependency `<input type="datetime-local">` editor, an `Intl.DateTimeFormat` renderer driven by a new `dateTimeFormat` option, ISO 8601 validation, and full integration with column sorting, filters, and Excel export.

**Data contract:** the canonical source value is ISO 8601 `YYYY-MM-DDTHH:mm:ss`. Parsing is lenient (date-only `YYYY-MM-DD` is treated as midnight; a space or `T` separator and optional seconds/milliseconds are accepted), so the reporter's mixed data renders correctly. The editor normalizes on input and canonicalizes to `YYYY-MM-DDTHH:mm:ss` on commit.

**No breaking changes.** The change is purely additive: two new cell types, a new `dateTimeFormat` option (with a default), and new docs. No existing default is changed, and no API, hook, or CSS class is renamed or removed.

### How has this been tested?

Unit tests (Jest) cover every new module:

```bash
npm --prefix handsontable run test:unit -- --testPathPattern="datetime|dateTime"
```

Added `*.unit.js`/`.unit.ts` for the ISO datetime helpers, renderer, validator, editor (`setValue`/`getValue` normalization incl. millisecond stripping), both cell types, the column-sorting compare function, the filter conditions, and the xlsx serial conversion. The registry enumeration tests were updated for the new built-ins.

Full gate, all green:

```bash
npm --prefix handsontable run test:unit   # 3290 passed
```

Manual verification in a real browser against the built bundle: rendering (date-only rendering midnight, space/`T` separators, `#bad-value#` for invalid input), opening the native `datetime-local` picker, editing, and committing back to the canonical ISO form. The documentation guide's live demo was verified rendering in all four frameworks (JavaScript, React, Angular, Vue).

### Types of changes

### Related issue(s):

1. DEV-1463
2. #12366

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-1463

---

> [!NOTE]
> **Medium Risk**
> Additive cell-type work that also wires sorting, filters, and Excel export. Parsing/validation of ISO date-times can affect displayed and exported values if edge cases are wrong.
> 
> **Overview**
> Adds the **`intl-datetime`** cell type so columns can store and edit combined date and time (ISO `YYYY-MM-DDTHH:mm:ss`), instead of date-only cells that only *display* midnight when time Intl options are set.
> 
> Editing uses a native `datetime-local` input (`showPicker`), with lenient parse (date-only → midnight, space or `T`, optional seconds/ms) and canonical commit. Display is driven by a new **`dateTimeFormat`** `Intl.DateTimeFormatOptions` setting.
> 
> Sorting, filter conditions (before/after/between/today/yesterday/tomorrow), and XLSX export treat values as real date-times. Docs clarify that time options on `date`/`intl-date` are display-only, and add a datetime guide plus JS/React/Angular/Vue examples. No breaking API changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2706b3c5775d11b3f4631b5a7bf625196504da70. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-1463 Date Time cell type](https://app.clickup.com/t/9015210959/DEV-1463)

qunabu (dismissed):
needs @demtario and @adrianspdev approve

budnix (commented):
Read through the datetime cell types. To answer the question that came up while reviewing: this isn't a duplicate of anything. Nothing named `datetime` has existed before — what looks familiar is the v18 `date` rework, where `date` dropped Pikaday for a native `<input type="date">` and gained the parallel `intl-date` name. This branch copies that shape for date-and-time values, which is the right instinct.

Sorting, filter registration, and the `datetime-local` input type all wire up correctly, and the changelog entry is in place. Unit tests pass and `test:types` is clean.

One gap worth naming: there's no Playwright E2E anywhere in the diff. The branch adds a native picker, `showPicker()`, sorting, filtering, and xlsx export — all things a user can see and do — and ships only unit tests. The presence gate passes, but the repo rule is E2E for user-visible behavior. Is that following in a separate PR?

The export regex comment is the one I'd act on before merge; the rest are questions or small gaps.

handsontable/src/plugins/exportFile/types/xlsx/date-utils.ts
```diff
@@ -117,3 +117,48 @@ export function getDateNumFmt(): string {
...
+ * @private
+ * @param {*} value Cell value — expected to be an ISO 8601 date-time string.
+ * @returns {number|null}
+ */
+export function parseIsoDateTimeStringToSerial(value: unknown): number | null {
+  if (!value) {
+    return null;
+  }
+
+  const match = String(value).match(/^(\d{4})-(\d{2})-(\d{2})(?:[T ](\d{2}):(\d{2})(?::(\d{2}))?)?$/);
```
budnix: This regex is looser and stricter than `ISO_DATETIME_REGEX` at the same time, and both directions bite. It rejects fractional seconds that the validator accepts, so a cell holding `2024-01-01T12:00:00.500` validates and renders fine but returns `null` here, falls through every branch in `#resolveCellValue`, and lands in the sheet as text — Excel can't sort or sum on it. It also uses `\d{2}` for the time parts with no bounds check, unlike `parseTimeStringToSerial` right above, so `T25:00:00` gives a `timeSerial` above 1 and the serial silently moves to the next day.

```suggestion
  // Mirrors ISO_DATETIME_REGEX in helpers/dateTime.ts, but captures the year for local parsing.
  const match = String(value).match(
    /^(\d{4})-(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01])(?:[T ]([01]\d|2[0-3]):([0-5]\d)(?::([0-5]\d)(?:\.\d{1,3})?)?)?$/
  );
```

handsontable/src/renderers/datetimeRenderer/datetimeRenderer.ts
```diff
@@ -0,0 +1,101 @@
...
+    return allowEmpty ? value : BAD_VALUE_TEXT;
+  }
+
+  const date = parseToLocalDateTime(value);
+
+  if (date === null) {
+    return BAD_VALUE_TEXT;
+  }
+
+  if (typeof dateTimeFormat === 'string' && instance && !stringFormatWarnShown.has(instance)) {
```
budnix: Two things about where this warn sits. It runs after the `date === null` check, so a column with a string `dateTimeFormat` and unparseable data shows `Bad value` and never warns — the one case where the hint would help most. And `dateRenderer` handles the same mistake differently: it warns first, then returns the raw value untouched, while this one warns and formats with `DEFAULT_INTL_FORMAT` anyway. Curious which behavior you want — either is defensible, but having `date` and `datetime` columns disagree will be confusing to debug.

handsontable/src/plugins/filters/constants.ts
```diff
@@ -176,8 +183,25 @@ export const TYPES: Record<string, string[]> = {
...
+    SEPARATOR,
+    CONDITION_INTL_DATETIME_BEFORE,
+    CONDITION_INTL_DATETIME_BEFORE_OR_EQUAL,
+    CONDITION_INTL_DATETIME_AFTER,
+    CONDITION_INTL_DATETIME_AFTER_OR_EQUAL,
+    CONDITION_INTL_DATETIME_BETWEEN,
+  ],
 };
 
+TYPES[TYPE_DATETIME] = TYPES[TYPE_INTL_DATETIME];
```
budnix: Two keys now point at the same array object, so anything that later mutates one type's condition list rewrites the other's menu too. Nothing does today since `getOptionsList` only reads, but a copy costs nothing and closes it off. Separately — was leaving out `today`/`yesterday`/`tomorrow` deliberate? Both `date` and `intl-date` offer them, and I wasn't sure whether those conditions can handle a full datetime string or would need datetime-aware versions.

```suggestion
TYPES[TYPE_DATETIME] = [...TYPES[TYPE_INTL_DATETIME]];
```

handsontable/src/dataMap/metaManager/metaSchema.ts
```diff
@@ -1973,6 +1978,54 @@ export default (): Record<string, unknown> => {
...
+     *     locale: 'en-US',
+     *     dateTimeFormat: {
+     *       dateStyle: 'medium',
+     *       timeStyle: 'short'
+     *     }
+     *   }
+     * ]
+     * ```
+     */
+    dateTimeFormat: {
```
budnix, line 2020: `dateTimeFormat` is a new public option, but it never got added to `GridSettings` in `core/settings.ts`, where `dateFormat` and `timeFormat` sit next to each other on lines 201-202. The `[key: string]: any` index signature at the bottom of that interface keeps it compiling, so no test catches it — but TypeScript users get no autocomplete and no type check on the option, and a typo like `datetimeFormat` passes silently.

handsontable/src/cellTypes/intlDatetimeType/intlDatetimeType.ts
```diff
@@ -0,0 +1,22 @@
+import { IntlDatetimeEditor } from '../../editors/intlDatetimeEditor';
+import { intlDatetimeRenderer, valueFormatter } from '../../renderers/intlDatetimeRenderer';
+import {
+  SOURCE_DATA_WARNING_MESSAGE,
+  intlDatetimeValidator,
+  sourceDataValidator,
+} from '../../validators/intlDatetimeValidator';
+
+export const CELL_TYPE = 'intl-datetime';
```
budnix, line 9: `IntlDatetimeEditor`, `intlDatetimeRenderer`, and `intlDatetimeValidator` all just call straight through to the base ones, so `datetime` and `intl-datetime` behave the same in every path — the only real difference is the wording of the source-data warning. For `date` and `time` the short name is a legacy alias you have to keep, but `datetime` never shipped before, so there's nothing to stay compatible with. My suggestion would be to keep only `intl-datetime` and drop `datetime` — it matches the newer naming family, and it halves the public surface plus the bundle for zero behavior loss. Once this merges the breaking-changes policy makes both names permanent, so pre-merge is the only chance to decide. That would also make the `TYPES[TYPE_DATETIME]` alias in `filters/constants.ts` unnecessary.

handsontable/src/cellTypes/index.ts
```diff
@@ -23,9 +25,11 @@ export function registerAllCellTypes() {
   registerCellType(AutocompleteCellType);
   registerCellType(CheckboxCellType);
   registerCellType(DateCellType);
+  registerCellType(DatetimeCellType);
```
budnix: `base.ts` keeps a hand-written list of the built-in editors, renderers, validators, and cell types for the UMD namespace types, and none of the six new datetime entries were added there. Runtime is fine thanks to the `[key: string]: unknown` index signature, but CDN users on TypeScript get `unknown` for `Handsontable.cellTypes.datetime` and `Handsontable.editors.DatetimeEditor`. Worth adding alongside the `intlDate` entries on lines 90 and 170.

handsontable/src/editors/datetimeEditor/__tests__/datetimeEditor.unit.js
```diff
@@ -0,0 +1,70 @@
+import { DatetimeEditor } from '../datetimeEditor';
+
+describe('DatetimeEditor value normalization', () => {
+  const makeEditor = () => {
+    const editor = Object.create(DatetimeEditor.prototype);
```
budnix: Building the editor with `Object.create(DatetimeEditor.prototype)` skips the constructor, so these tests only reach `getValue` and `setValue`. The two methods this class actually overrides for behavior — `prepare()`, which swaps in the raw ISO source value, and `open()`, which calls `showPicker()` inside a try/catch — are never run, so a regression in either stays green. Was that a deliberate scope call, with the real coverage meant to land as E2E?

handsontable/src/plugins/exportFile/types/xlsx/date-utils.ts
adrianspdev: Fixed in e7a0820c75 — applied the suggested regex mirroring `ISO_DATETIME_REGEX`, so fractional seconds now export as serials and out-of-range date/time parts return `null` instead of rolling the serial. Added regression tests for `2024-01-01T12:00:00.500`, `T25:00:00`, `12:60`, `12:00:61`, month 13, and day 32 in `datetimeExport.unit.js`.

handsontable/src/renderers/datetimeRenderer/datetimeRenderer.ts
adrianspdev: Aligned with `dateRenderer` in e7a0820c75: a string `dateTimeFormat` now warns once per instance before parsing and returns the raw value untouched — so the warning also fires for unparseable data, and `date`/`intl-datetime` columns handle the mistake identically. Unit tests cover the once-per-instance dedup and the unparseable-value case.

handsontable/src/plugins/filters/constants.ts
adrianspdev: The alias line is gone entirely — with the `datetime` name dropped (see the intlDatetimeType thread), `TYPES` has only the `intl-datetime` key. On the second question: leaving out `today`/`yesterday`/`tomorrow` was a gap, not a decision. The existing conditions parse with `parseToLocalDate`, which is strict date-only, so a full datetime string would fail to parse and the condition would reject every row. Added datetime-aware versions in `condition/intlDatetime/{today,yesterday,tomorrow}.ts` (parse with `parseToLocalDateTime`, compare the calendar-day parts) and registered them in the `intl-datetime` menu, with unit tests and an E2E case for "Today" matching regardless of the time part.

handsontable/src/dataMap/metaManager/metaSchema.ts
adrianspdev: Added `dateTimeFormat?: Intl.DateTimeFormatOptions` to `GridSettings` next to `dateFormat`/`timeFormat`, plus the corresponding entry in `settings.types.ts` — which is the exhaustive `Required<GridSettings>` check, so `test:types` failed until the option was added, confirming the gap.

handsontable/src/cellTypes/intlDatetimeType/intlDatetimeType.ts
adrianspdev: Done in e7a0820c75 — kept only `intl-datetime`. The implementation now lives directly in the `intlDatetime*` modules; the `datetime*` directories, registrations, the sort-function registration, the filters `TYPES` alias, and the docs/changelog mentions of the bare name are all removed. Public surface and bundle halved, nothing to keep compatible.

handsontable/src/cellTypes/index.ts
adrianspdev: Added the UMD namespace entries in `base.ts` beside the `intlDate` ones: `IntlDatetimeEditor`, `IntlDatetimeRenderer`, `IntlDatetimeValidator`, and the `intlDatetime` cell type. Four entries rather than six — the `datetime`-named half no longer exists after dropping the alias.

handsontable/src/editors/datetimeEditor/__tests__/datetimeEditor.unit.js
adrianspdev: Rewritten around real construction (`new IntlDatetimeEditor(hotMock)`), so the constructor, `createElements` (input type/step/dir), the `init` `afterSetTheme` close hook, `prepare()` (raw ISO source swap, physical-row translation asserted), and `open()` (`showPicker` invoked, user-gesture errors swallowed) are all exercised. One caveat: jsdom's `datetime-local` value sanitizer rewrites assigned values nonstandardly (appends `.000`), so the pure string-normalization tests swap in a plain object for the input, with a comment. The real-input path is covered by the new Playwright suite in `tests/e2e/datetime-cell-type.spec.ts`.

adrianspdev:
Addressed the full review in e7a0820c75 (replies in each thread). On the E2E gap from the review summary: it was not planned as a separate PR — added a Playwright suite in this one. `tests/e2e/datetime-cell-type.spec.ts` with a `DatetimePage` page object and a dedicated fixture covers Intl-formatted rendering, the native `datetime-local` editor seeded with the raw ISO value plus a `showPicker()` invocation assert, the edit-commit round trip, chronological sorting in both directions, and the datetime "Today" filter condition with the clock pinned via `page.clock.install`. All five pass locally on the main theme; xlsx export stays unit-tested since asserting a downloaded binary adds no signal over the serial-conversion tests.

budnix (dismissed):

budnix (approved):

demtario (approved):

PR #13155, opened 2026-08-04
Revert "DEV-2138: Set an explicit shortcuts group in the editorFactory recipes" (#13154)

demtario:
### Context

The PR reverts #13154 (commit `58cf584f1`) from develop, as planned when that PR was opened.

#13154 was a temporary workaround for `prod-docs/18.0`: docs previews are bundled from the branch's own core (`docs/astro.config.mjs` aliases `handsontable` to `handsontable/tmp/`), and Handsontable 18.0.0 throws `You need to define the shortcut's group.` when `shortcutsGroup` is `undefined`. It has now been cherry-picked to `prod-docs/18.0` (`450775992`), so it no longer needs to live here.

On develop the workaround is redundant - core defaults the group to `'customEditor'` (#13130) - and the next release's examples should not ship a workaround for a bug that release does not have. The workaround deliberately used that same `'customEditor'` literal, which makes this revert behavior-neutral: `git diff` against `3d6c130aa` (the commit before the workaround) is empty.

Reverting also restores the Angular Feedback and Star Rating recipes: `'angular'` returns to their `onlyFor` lists in `docs/content/recipes/sidebar.js`, and the Cloudflare worker stops bouncing the Angular aliases to the cell-types index.

One thing this revert does not address: `FactoryEditorAdapter` still does not forward `shortcutsGroup` from `HotCellEditorAdvancedComponent`, so the documented component option stays inert. That is harmless on develop, because core supplies the default, but the wrapper gap is worth a separate fix.

### How has this been tested?

```shell
npm run docs:test:plugins --prefix docs   # 224/224 pass
```

The count drops from 225 to 224 because the worker redirect test added by #13154 is reverted along with the redirects it covered.

Tree equivalence was verified directly - the revert restores the pre-workaround state exactly, with no residue:

```shell
git diff 3d6c130aa    # empty
```

No browser check: the revert returns the examples to code that shipped on develop before #13154, and develop's core makes `shortcutsGroup` unnecessary.

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature or improvement (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Additional language file or change to the existing one (translations)

### Related issue(s):

1. DEV-2138

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/angular-wrapper`
- [ ] `@handsontable/react-wrapper`
- [ ] `@handsontable/vue3`

[...]

Issue #333, opened 2013-01-03
Callback when table is finished loading


stevebooks:
Is there a callback to trigger when the table has finished loading?

PR #13157, opened 2026-08-04
DEV-2180: Move recipes to upstream pikaday and the built-in date cell type

demtario:
### Context

The PR stops pointing readers at the `@handsontable/pikaday` fork, which exists only to keep two documentation recipes working. 18.0 replaced the built-in Pikaday date editor with a native `<input type="date">`, so the fork now has no other consumer in the repo.

- **`recipes/cell-types/pikaday`** stays (it is the migration path for users who relied on the removed built-in editor) and switches to upstream `pikaday@1.8.2` + `@types/pikaday`.
- **`recipes/cell-types/moment-date`** drops Pikaday entirely and composes the built-in `date` cell type instead.
- **`recipes/cell-types/moment-time`** declared the fork in one `::: example` directive but never imported it — dead reference, removed.

Permalinks are unchanged, and both pages keep all four frameworks in the sidebar.

#### Fork vs upstream

Upstream has no `useMoment()`, and its module-scope `hasMoment` check is `false` under a bundler, which would silently break three things: the calendar would open on today instead of the cell's date, the editor input would show `Wed Dec 31 2024` instead of `12/31/2024`, and a `DD/MM/YYYY` column would misparse its own values. All three are fixed with Pikaday's own option hooks, which it consults *before* its internal moment path:

```js
options.toString = (date, format) => moment(date).format(format);
options.parse = (dateString, format) => moment(dateString, format).toDate();
```

`setMoment(m, true)` becomes `setDate(m.toDate(), true)`. That is behavior-identical, not an approximation — upstream's `setMoment` is only `if (hasMoment && moment.isMoment(date)) { this.setDate(date.toDate(), preventOnSelect); }`.

#### Calendar styling and dark mode

The picker previously rendered light-only, even on a dark page. Upstream's `pikaday.css` is the cause: hardcoded `#333` / `#fff` / `#ccc`, base64 PNG arrows, and no `color-scheme` at all. Its only dark path, `css/theme.css`, is a self-described example theme with `background: #ff8000 !important` on hover, and it cannot follow either the docs theme toggle or `ht-theme-*-dark-auto`.

Upstream's stylesheet is kept as the structural base, with the theme-aware rules Handsontable shipped until 18.0 (removed in #12689) restored **in the recipe only** and re-pointed at `--ht-*` variables. This works without any extra wiring because the editor uses `position: 'portal'`, so it is attached to the body-level `.ht-portal` element — and core adds the active `ht-theme-*` class to that element. Every `--ht-*` token and the theme's `color-scheme` therefore resolve on the calendar's own ancestor.

Three deliberate deviations from the restored stylesheet:

1. The old rules killed upstream's PNG arrows (`text-indent: -9999px`) and relied on a generated chevron-mask layer that has since been removed, which would have left the prev/next buttons invisible. The mask rules are carried over and keyed to `currentColor`, so the arrows track the accent color in both schemes.
2. The RTL block is re-keyed from `[dir=rtl].htDatepickerHolder` (a Pikaday-era class that no longer exists) to `.pika-single.is-rtl`.
3. Every selector is scoped to `[class*='ht-theme-']`. Without that, `pikaday.css` wins on source order — a first attempt measured `color: #333` on the panel despite the override being present.

#### Moment.js date recipe

The recipe keeps its Moment.js teaching and replaces only the editor. `valueFormatter` gives each column its own display format while the stored value stays ISO, and the inherited renderer is left untouched — `cellProperties.valueFormatter` runs *before* the renderer, so a custom moment renderer would have received the already-formatted string and parsed garbage.

Lenient input correction lives in a `beforeChange` handler rather than in a validator. The native date input's value is spec-constrained to `''` or an ISO string, so `valueParser` could never see a non-ISO value; and correcting from inside a validator made core log `DateEditor: value must be in ISO date format` on every corrected paste. `beforeChange` runs ahead of both the editor and the validator, so the corrected value is the only one the rest of the grid sees, and the inherited ISO validator still flags values Moment.js cannot read.

Two options the built-in type does not support are removed from the columns and documented on the page: `datePickerConfig` (Pikaday-specific, now warns) and `dateFormat` as a Moment format string (the built-in renderer expects `Intl.DateTimeFormatOptions`). The recipe uses its own `renderFormat` and `inputFormat` properties instead. `firstDay`, `showWeekNumber`, and `disableDayFn` have no native equivalent; the page points readers at the Pikaday and Flatpickr recipes for those.

The Angular variant previously had no `registerCellType` at all — it wired editor and renderer components onto the column with an `as any` cast. It now uses the same composed cell type as the other three, which also retires that cast.

#### Drive-by fix

`Handsontable.helper` is an empty object in 18.0, so the Pikaday recipe's `Handsontable.helper.isMobileBrowser()` and `isFunctionKey()` calls threw `TypeError: ... is not a function`. Arrow-key navigation — a feature the page advertises — was broken on `develop`, independently of this migration. Replaced with `matchMedia('(pointer: coarse)')` and a `key.length` check rather than importing from `handsontable/helpers/*`, which is not a surface a recipe should teach.

#### Note for whoever merges #13155

#13155 reverts the `shortcutsGroup` workaround and touches the same recipe families. Verified with `git apply --check` against this branch: it applies cleanly to all five Pikaday files, `docs/content/recipes/sidebar.js`, and the Cloudflare worker, and conflicts on exactly five moment-date files (`javascript/example1.{js,ts}`, `react/example1.{jsx,tsx}`, `moment-date.md`). In every case the conflict is one-sided — this PR deletes the whole block that the revert removes one line from, so **keep this PR's deletion and drop the revert hunk**. Merging #13155 first avoids all five.

### How has this been tested?

```shell
npm run build --prefix docs                      # 1294 pages, exit 0
npm run docs:lint --prefix docs                  # clean
npm run docs:test:plugins --prefix docs          # 225/225
npm --prefix docs/angular-type-check run typecheck   # 0 errors (SKIP_LATEST=1)
BASE_URL=http://localhost:PORT/docs npx playwright test recipeConsoleErrors.spec.ts   # 72/72
node scripts/test-runner-links.mjs --filter pikaday  # 3/3 headless render
```

Browser-checked both pages after a cold `rm -rf docs/.astro` start, in all four framework variants (8 combinations, all render).

Measured on the live picker rather than inferred:

- **Dark mode** — panel `rgb(255,255,255)` / text `rgb(34,34,34)` in light, `rgb(5,5,6)` / `rgb(185,185,187)` in dark, accent shifting `rgb(26,66,232)` → `rgb(70,105,246)`. Arrow masks resolve in both.
- **`toString` hook** — editor input reads `08/01/2025`, not `Fri Aug 01 2025`.
- **`setDate`** — calendar opens on the cell's own date, and arrow keys move it (`08/01` → `08/02` on ArrowRight, `→ 08/08` on ArrowDown) with no console error.
- **Editor identity on moment-date** — the editor is `<input type="date">` holding the raw ISO `2025-08-01` while the cell displays `Aug 1, 2025`.
- **Paste correction** — pasting `03/14/2025` lands as ISO and displays `Mar 14, 2025` with an empty console; pasting `not a date at all` is flagged `htInvalid`. Confirmed separately on the vanilla, React, and Angular pages, since `beforeChange` relies on `this` being the grid instance and wrapper option forwarding has regressed before.
- **No stray warnings** — checked by hand for `The "datePickerConfig" option is not supported` and `The dateFormat option as a string is not supported`; neither fires. Both are `console.warn`, so `recipeConsoleErrors` would not have caught them.

### Types of changes

### Related issue(s):

1. DEV-2180

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

[skip changelog]

Documentation-only change - no package source is touched.

### Known follow-ups (not in this PR)

- The example runner (`demos.handsontable.com`) installs `@handsontable/pikaday`. Its links pass today because it serves its own deployed snapshot, so nothing is broken right now — but the two Pikaday recipe links will fail once it resyncs to a snapshot containing these examples. Tracked separately; it lives in another repository.
- `docs/angular-type-check/package-lock.json` still names the fork. It was already stale on ten unrelated entries before this change, npm cannot regenerate it (`EUNSUPPORTEDPROTOCOL` on the `link:` deps), and CI installs with `pnpm`. Worth deleting as its own cleanup.
- The recipe reuses the option name `datePickerConfig` as a custom cell property. No warning fires, because its column is not `type: 'date'`, but renaming it would stop readers assuming the removed built-in option is back.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2180

---

> [!NOTE]
> **Low Risk**
> Documentation and example-only changes; behavior risk is limited to docs demos and dependency swaps, not production grid code.
> 
> **Overview**
> Removes the **`@handsontable/pikaday`** fork from docs dependencies in favor of **`pikaday@1.8.2`** and **`@types/pikaday`**.
> 
> The **Pikaday** recipe keeps the portal calendar editor but wires Moment through Pikaday’s **`toString` / `parse`** options (replacing **`useMoment`** / **`setMoment`**), restores **Handsontable theme-aware calendar CSS** for light/dark, and fixes **broken `Handsontable.helper` calls** with **`matchMedia('(pointer: coarse)')`** and a **`key.length`** check.
> 
> The **Moment.js date** recipe is reworked: it **extends `DateCellType`** (native date input + ISO validation), uses **`valueFormatter`** for **`renderFormat`**, and **`beforeChange`** to normalize pasted non-ISO values via **`inputFormat`** / **`correctFormat`**—**Pikaday, custom editors, and editor CSS are removed**. Angular aligns with **`registerCellType`** instead of per-column editor/renderer components.
> 
> **moment-time** drops an unused Pikaday example dependency; **flatpickr** cross-links are updated. No Handsontable package source changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c2fd7e3a8c5005d7c14b23f101eeabbf14afc52d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2180 Docs: migrate the Pikaday recipe to the upstream pikaday package and drop Pikaday from the Moment.js date recipe](https://app.clickup.com/t/9015210959/DEV-2180)

adrianspdev (approved):

PR #13158, opened 2026-08-04
DEV-2189: Fix the repo-root resolution that broke the gates in worktrees

demtario:
### Context

The PR fixes the local enforcement gates in `git worktree` checkouts, where they were crashing (pre-push) or silently doing nothing (green-run cache).

**1. Every `git push` from a worktree crashed before any gate ran.**

`scripts/pre-push.mjs` resolved the repository root with `git rev-parse --show-toplevel` and `cwd: scriptDir`. Git hooks run with `GIT_DIR` exported (`<main>/.git/worktrees/<name>` in a worktree), and with `GIT_DIR` set explicitly `--show-toplevel` stops discovering the work tree by walking up from the cwd and reports the **cwd itself**. So `root` became `<worktree>/scripts` and the very first gate died:

```
Error: Cannot find module '<worktree>/scripts/.github/scripts/test-presence-gate.mjs'
```

The workaround was `git push --no-verify`, which skips the *entire* hook — the presence gate, ESLint on changed files, the test-weakening detector, and the changed Playwright/Jest runs. `.ai/LOCAL-ENFORCEMENT.md` calls those the enforcement floor for every change; in a worktree that floor was absent and nothing signalled it. Agent-driven work runs in worktrees by default, so this was the common path. The bug was invisible when running `node scripts/pre-push.mjs` by hand, because no `GIT_DIR` is set then.

**2. The green-run cache was permanently cold in every worktree.**

Found while fixing the above, so it rides along here — same worktree-hostility class, same test file, same documentation line. `scripts/e2e-run-cache.mjs` read and wrote `path.join(root, '.git', 'hot-e2e-green.json')`. In a linked worktree `<root>/.git` is a **file**, not a directory, so every write failed with `ENOTDIR` into a swallowed `catch` and every read returned an empty set. The run-once optimization never applied: touched Playwright specs were re-run on every single push.

**The fix.** A new pure helper, `.github/scripts/lib/repo-root.mjs`:

- `repoRoot()` derives the root from the module's own location, with the depth anchored inside the module so no caller can get the level wrong. No subprocess, independent of the cwd and of the git environment — identical in a clone, in a worktree, in CI, and inside a hook. This is what the comment at the old call site already promised.
- `gitDir(root)` returns the checkout's real git directory, following the `gitdir:` pointer when `.git` is a file. It resolves the worktree's own git directory rather than the shared common directory, so per-checkout state stays per-checkout.

Call sites converted: `scripts/pre-push.mjs` (was live), `.github/scripts/diff-coverage-gate.mjs` and `scripts/claude/session.mjs` (both had the same latent call), and `scripts/e2e-run-cache.mjs` now goes through `gitDir()`.

Two notes for the reviewer:

- `pre-push.mjs` now strips `GIT_DIR`/`GIT_WORK_TREE` from the environment of every child it spawns. This is **prevention, not a fix** — no failure was observed. Playwright and Jest are spawned with `cwd` set to `tests/` and `handsontable/`, and an inherited `GIT_DIR` makes git treat that subdirectory as the work tree, so it kills the class rather than the one instance.
- `scripts/claude/session.mjs` loses its `|| process.cwd()` fallback. The new `repoRoot()` cannot fail, so the branch was dead. It is a behavior change in an agent hook, hence called out.

### How has this been tested?

Tooling-only change, so the tooling suite plus a real-hook A/B.

- **New unit tests** — `.github/scripts/__tests__/repo-root.test.mjs` (6 tests): the root resolves to the repository root from a different cwd; it is unchanged with `GIT_DIR`/`GIT_WORK_TREE` set to bogus values (the regression); `gitDir()` for a clone, for a linked worktree, for a relative `gitdir:` pointer, and outside a checkout. The root is asserted against files that only exist at the repository root (`lefthook.yml`, `pnpm-workspace.yaml`) — deliberately **not** against `git rev-parse --show-toplevel`, which would make the test inherit the bug it exists to catch.
- **Updated `scripts/__tests__/e2e-run-cache.test.mjs`** — its `fakeRoot()` did `mkdirSync(root/.git)`, which encoded the broken assumption and would have stayed green after the fix. It now takes a `shape` parameter, and two new tests cover the worktree layout (`.git` as a file with a `gitdir:` pointer): the run-once round-trip, and invalidation on spec edit.

  ```bash
  npm run test:tooling   # 91 passing, 0 failing
  ```

- **The cache bug proven red before green** — the new worktree round-trip was run against `origin/develop`'s `e2e-run-cache.mjs` and failed (`skipped: []`), then passed against the fixed module.
- **Real hook, A/B, no network** — in a throwaway `git clone --local` with a bare `file://` remote (nothing was pushed to `origin`, no commits were made in the working repository), a `pre-push` shim was installed and a source-change-without-a-test was pushed from a linked worktree:
  - pre-fix script → `Error: Cannot find module '<wt>/scripts/.github/scripts/test-presence-gate.mjs'`
  - post-fix script → the presence gate runs and **blocks** on the missing test, as designed
  - the same push from that clone's normal (non-worktree) checkout also blocks — no regression outside worktrees
- **This branch was pushed from a worktree without `--no-verify`**, and the hook ran green end to end (`Test-presence gate ✅ Pass` → `Test-weakening gate` → `pre-push: ok`).
- **Real-layout cache path confirmed** — `cacheFile()` for this worktree returns `<main>/.git/worktrees/<name>/hot-e2e-green.json`.
- **Agent hooks smoke-tested** under `GIT_DIR` — `scripts/claude/stop.mjs` and `scripts/claude/post-tool-use.mjs` both exit 0, and `repoRoot()` returns the worktree root.
- **ESLint** clean on the changed `scripts/**` files. `.github/**` is outside the hook lint scope (the dot-segment filter in `scripts/lint-files.mjs`); the new import there matches the extension style of its sibling tests.

[skip changelog] — tooling only, no `handsontable/src/**` or `wrappers/**` change.

### Types of changes

### Related issue(s):

1. DEV-2189

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

- [x] My change requires a change to the documentation. `.ai/LOCAL-ENFORCEMENT.md` is updated in this PR: the cache location now names both layouts, and a new rule tells future contributors to take the root from `repoRoot()` and per-checkout state from `gitDir(root)` instead of asking git.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2189

---

> [!NOTE]
> **Medium Risk**
> Changes how local git hooks resolve paths and spawn children; behavior is well covered by tests but affects the enforcement floor for every push.
> 
> **Overview**
> Fixes local enforcement in **linked `git worktree` checkouts**, where hooks previously crashed or skipped optimizations.
> 
> **`repoRoot()`** and **`gitDir()`** are added in `.github/scripts/lib/repo-root.mjs`: the root comes from the module’s path (not `git rev-parse --show-toplevel`, which under hook-exported `GIT_DIR` returns the cwd), and the real git directory follows a worktree’s `gitdir:` pointer when `.git` is a file. **`pre-push`**, the diff-coverage gate, and Claude **`session.mjs`** use `repoRoot()`; the E2E green-run cache writes **`hot-e2e-green.json`** via `gitDir()` instead of `<root>/.git/`, which previously failed silently with ENOTDIR in worktrees. **`pre-push`** also strips `GIT_DIR` / `GIT_WORK_TREE` from spawned child processes. **`.ai/LOCAL-ENFORCEMENT.md`** documents the worktree rules; new/updated unit tests cover root resolution and cache behavior in worktree layout.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 40e1b1a3e70a91d6b56813d8eaeb9beaf5f1afd9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/scripts/__tests__/repo-root.test.mjs
+++ b/.github/scripts/__tests__/repo-root.test.mjs
@@ -1,0 +1,87 @@
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { existsSync, mkdtempSync, mkdirSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import path from 'node:path';
+import { repoRoot, gitDir } from '../lib/repo-root.mjs';
+
+// The root is asserted against files that only exist at the repository root —
+// NOT against `git rev-parse --show-toplevel`, which is the very thing that
+// reports the wrong directory under GIT_DIR and would make this test inherit the
+// bug it exists to catch.
+test('repoRoot() points at the repository root, from any cwd', () => {
+  const root = repoRoot();
+
+  assert.ok(existsSync(path.join(root, 'lefthook.yml')), `${root} has no lefthook.yml`);
+  assert.ok(existsSync(path.join(root, 'pnpm-workspace.yaml')), `${root} has no pnpm-workspace.yaml`);
+  assert.ok(existsSync(path.join(root, '.github/scripts/lib/repo-root.mjs')));
+
+  const cwd = process.cwd();
+
+  try {
+    process.chdir(path.join(root, 'scripts'));
+    assert.equal(repoRoot(), root);
+  } finally {
+    process.chdir(cwd);
+  }
+});
+
+// The regression: git hooks export GIT_DIR (in a worktree, pointing at
+// `<main>/.git/worktrees/<name>`), and with it set `--show-toplevel` returns the
+// cwd instead of the work tree — which made pre-push resolve `<root>/scripts` as
+// the root and crash with MODULE_NOT_FOUND before any gate ran.
+test('repoRoot() ignores the git environment a hook exports', () => {
+  const expected = repoRoot();
+  const { GIT_DIR, GIT_WORK_TREE } = process.env;
+
+  try {
+    process.env.GIT_DIR = path.join(expected, '.git/worktrees/does-not-exist');
+    process.env.GIT_WORK_TREE = tmpdir();
+    assert.equal(repoRoot(), expected);
+  } finally {
+    if (GIT_DIR === undefined) {
+      delete process.env.GIT_DIR;
+    } else {
+      process.env.GIT_DIR = GIT_DIR;
+    }
+
+    if (GIT_WORK_TREE === undefined) {
+      delete process.env.GIT_WORK_TREE;
+    } else {
+      process.env.GIT_WORK_TREE = GIT_WORK_TREE;
+    }
+  }
+});
+
+test('gitDir() returns <root>/.git in a normal clone', () => {
+  const root = mkdtempSync(path.join(tmpdir(), 'repo-root-clone-'));
+
+  mkdirSync(path.join(root, '.git'), { recursive: true });
+
+  assert.equal(gitDir(root), path.join(root, '.git'));
+});
+
+test('gitDir() follows the `gitdir:` pointer of a linked worktree', () => {
+  const main = mkdtempSync(path.join(tmpdir(), 'repo-root-main-'));
+  const wt = mkdtempSync(path.join(tmpdir(), 'repo-root-wt-'));
+  const linked = path.join(main, '.git/worktrees/wt');
+
+  mkdirSync(linked, { recursive: true });
+  // In a linked worktree `.git` is a FILE, not a directory.
+  writeFileSync(path.join(wt, '.git'), `gitdir: ${linked}\n`);
+
+  assert.equal(gitDir(wt), linked);
+});
+
+test('gitDir() resolves a relative `gitdir:` pointer against the root', () => {
+  const root = mkdtempSync(path.join(tmpdir(), 'repo-root-rel-'));
+
+  mkdirSync(path.join(root, 'nested/gitdir'), { recursive: true });
+  writeFileSync(path.join(root, '.git'), 'gitdir: nested/gitdir\n');
+
+  assert.equal(gitDir(root), path.join(root, 'nested/gitdir'));
+});
+
+test('gitDir() returns null outside a checkout', () => {
+  assert.equal(gitDir(mkdtempSync(path.join(tmpdir(), 'repo-root-bare-'))), null);
+});
--- a/.github/scripts/__tests__/repo-root.test.mjs
+++ b/.github/scripts/__tests__/repo-root.test.mjs
@@ -1,8 +1,10 @@
 import { test } from 'node:test';
 import assert from 'node:assert/strict';
+import { execFileSync } from 'node:child_process';
 import { existsSync, mkdtempSync, mkdirSync, writeFileSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import path from 'node:path';
+import { pathToFileURL } from 'node:url';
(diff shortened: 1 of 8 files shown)


qunabu:
Task linked: [DEV-2189 Pre-push test-gate crashes in every git worktree, silently disabling all local push gates](https://app.clickup.com/t/9015210959/DEV-2189)

.github/scripts/__tests__/repo-root.test.mjs
```diff
@@ -0,0 +1,87 @@
...
+  } finally {
+    process.chdir(cwd);
+  }
+});
+
+// The regression: git hooks export GIT_DIR (in a worktree, pointing at
+// `<main>/.git/worktrees/<name>`), and with it set `--show-toplevel` returns the
+// cwd instead of the work tree — which made pre-push resolve `<root>/scripts` as
+// the root and crash with MODULE_NOT_FOUND before any gate ran.
+test('repoRoot() ignores the git environment a hook exports', () => {
```
danielzytohoc, line 43: GIT_DIR regression test mutates process.env after the module is imported, but ROOT is computed once at import time. It therefore locks out a call-time git-derived reimplementation, but a module-load-time one (const ROOT = execSync('git rev-parse …') at top level) would pass this suite while still breaking in real hooks, where GIT_DIR is present from process start. An airtight variant would spawn a child node with GIT_DIR set in its env, import the module there, and assert the printed root. Worth doing since this test's whole job is to be the regression lock.

scripts/e2e-run-cache.mjs
```diff
@@ -149,6 +166,11 @@ export function recordGreen(root, specs) {
     return;
   }
   const keys = new Set(readCache(root));
+  const file = cacheFile(root);
+
+  if (!file) {
```
danielzytohoc: readCache(root) runs before the if (!file) return guard, so when cacheFile() is null the read has already been attempted and silently failed. Hoisting the guard above the read (and passing the resolved file into the read) is slightly cleaner and avoids resolving gitDir() twice. Cosmetic.

.github/scripts/__tests__/repo-root.test.mjs
demtario: Rewrote it as a child-process test: the child is spawned with `GIT_DIR` in its env before the module is imported, from `<root>/scripts` as cwd, and prints `repoRoot()`. Verified it is a real lock by temporarily patching `ROOT` to a load-time `execSync('git rev-parse --show-toplevel')` — the test then fails, with `GIT_DIR` alone and no `GIT_WORK_TREE`, which is what a hook actually exports. Dropped the in-process variant; the child covers both load-time and call-time.

scripts/e2e-run-cache.mjs
demtario: Hoisted it. `readCache()` now takes the resolved path with an explicit `if (!file) return new Set()` — passing `null` into `readFileSync` would just move the silent failure rather than remove it. Also swapped the guard order in `recordGreen()` so `cacheFile()` resolves before `envHash()`, which hashes the dist bundle and every fixture.

danielzytohoc (approved):

PR #13163, opened 2026-08-05
DEV-2190: Playwright E2E matrix parity with Puppeteer (theme × bundle)

danielzytohoc:
### Context

The PR makes the Playwright functional E2E matrix a strict 1:1 mirror of the Puppeteer matrix: `theme [main, horizon, classic] × bundle [UMD, UMD.min]` = 6 legs, loading the exact same dist files the Puppeteer legs load (`test:e2e` → `dist/handsontable.js`, `test:production` → `dist/handsontable.full.min.js`). Until now Playwright ran 3 theme-only legs, and the fixture pages hardcoded `handsontable.full.min.js` without any dev-bundle leg. This is a prerequisite for migrating frozen Jasmine specs to Playwright (DEV-2183): migration must not narrow what CI proves.

Changes:

- `tests/fixtures/test.ts` — new `bundle` option (`umd` = `dist/handsontable.js`, `full-min` = `dist/handsontable.full.min.js`) beside `theme`.
- `tests/playwright.config.ts` — 6 projects (`e2e-<theme>` + `e2e-<theme>-min`). The un-suffixed names keep their existing meaning, so local commands and the Stop/pre-push hooks (`--project=e2e-main`) work unchanged.
- `tests/fixtures/demo/grid.html` + `walkontable.html` — the bundle script src is selected from the `?bundle=` query param through the same allowlist-to-literal pattern the theme stylesheet already uses (no reflection possible; junk falls back to `handsontable.js`).
- Page objects and specs thread the option through (`goto()` appends `&bundle=`).
- `.github/workflows/e2e.yml` — the `playwright` job matrix becomes `theme × bundle` (6 legs), with per-leg report artifact names. CI Gate absorbs the new legs; no branch-protection change needed.

The bundles Puppeteer does not test either (`handsontable.full.js`, `handsontable.min.js`) are deliberately **not** added here — extra coverage is recorded as a nightly candidate on DEV-2058 (see task comment). `verify-bundles.mjs` still smoke-checks all variants.

The min leg loads the full bundle (HyperFormula built in), same as Puppeteer's production leg. When formulas specs migrate (DEV-2183), the `umd` leg's fixture must load HyperFormula as an external script beside the base bundle — noted in DEV-2190.

### How has this been tested?

- `cd tests && npx playwright test --project=e2e-main --project=e2e-main-min` — 12/12 passed (both bundle legs, all three spec files).
- Direct fixture verification via a scripted browser: `?bundle=umd` → loads `handsontable.js`, `?bundle=full-min` → `handsontable.full.min.js` (confirmed via the network request, HTTP 200), `?bundle=<script>evil` and no param → fall back to `handsontable.js` (allowlist holds).
- Pre-push hook re-ran the 3 changed specs — green.

### Types of changes

- [x] New feature or improvement (non-breaking change which adds functionality) — CI/test infrastructure

### Related issue(s):

1. DEV-2190
2. DEV-2183 (unblocked by this), DEV-2058 (nightly owns the extra-bundle legs)

### Affected project(s):

- [x] `handsontable` 
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2190

---

> [!NOTE]
> **Medium Risk**
> Test infrastructure only, but the e2e-main bundle flip and expanded CI matrix change what merges prove locally vs in CI; incorrect cache hashing could skip specs against the wrong build.
> 
> **Overview**
> Extends Playwright functional E2E from **three theme-only projects** to **six legs** (main/horizon/classic × `umd` / `full-min`), loading the same dist files as Puppeteer (`handsontable.js` vs `handsontable.full.min.js`). CI’s `playwright` job now runs one matrix entry per leg; report artifact names include the bundle suffix.
> 
> **Fixture and wiring:** Demo HTML no longer hardcodes `handsontable.full.min.js`. `?theme=` and `?bundle=` are resolved through allowlists (unknown values **throw**; missing params default to main / plain UMD). Page objects and specs take `{ theme, bundle }` from `fixtures/test.ts` and pass both into `goto()`.
> 
> **Local gates vs CI:** Unsuffixed projects (`e2e-main`, etc.) now run **plain UMD**; `-min` legs are CI-only. Pre-push, Stop hook, and `npm run test:e2e` stay on `e2e-main` but the bundle under test **changed** from full.min to base UMD.
> 
> **Cache:** `scripts/e2e-run-cache.mjs` hashes **both** dist bundles so a rebuild of either invalidates green-run skips; tests cover base UMD-only rebuilds and missing `handsontable.js`.
> 
> Docs/skills (`tests/AGENTS.md`, README, root `AGENTS.md`, Playwright E2E skill) describe the matrix and fixture contract. Walkontable frozen-row specs use `expect.poll` for scroll-band preconditions after bundle/param changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 97a79565d98e16a0b5800362506922e76f155f4c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2058 Pipeline: develop trim + nightly full matrix (move full e2e off the develop push)](https://app.clickup.com/t/9015210959/DEV-2058)
Task linked: [DEV-2183 Escalation burn-down 1: triage the 31 skipped tests, flip no-skipped-test to error](https://app.clickup.com/t/9015210959/DEV-2183)
Task linked: [DEV-2190 Playwright E2E matrix parity with Puppeteer: add the bundle dimension (3 → 6 legs)](https://app.clickup.com/t/9015210959/DEV-2190)

tests/fixtures/demo/grid.html
```diff
@@ -12,6 +12,11 @@
     window.htTheme = ({ horizon: 'horizon', classic: 'classic' })[new URLSearchParams(location.search).get('theme')] || 'main';
     // Written into <head> so the themed stylesheet is present before first render.
     document.write('<link rel="stylesheet" href="/handsontable/styles/ht-theme-' + window.htTheme + '.min.css">');
+    // Active bundle from the ?bundle= query param set by the Playwright project
+    // — 1:1 with the Puppeteer legs (umd = handsontable.js, umd-min =
+    // handsontable.min.js). Same allowlist-to-literal pattern as the theme, so
+    // the param can never be reflected into the document.
+    window.htBundle = ({ 'umd-min': 'handsontable.min.js' })[new URLSearchParams(location.search).get('bundle')] || 'handsontable.js';
```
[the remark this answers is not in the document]
danielzytohoc: Fixed in 4861ea495. The min leg's bundle value is now `full-min` and the fixture allowlist maps it to `handsontable.full.min.js` — the exact file the Puppeteer `test:production` leg loads — so the matrix is a true 1:1 mirror. Verified via a scripted browser that `?bundle=full-min` requests `dist/handsontable.full.min.js` (HTTP 200) and that junk/missing params still fall back to `handsontable.js`; both main-theme legs pass 12/12. `handsontable.min.js` is dropped entirely — Puppeteer never tested it, and the extra-bundle legs stay a nightly candidate on DEV-2058.

qunabu (dismissed):
@budnix can you review this as well please

.github/workflows/e2e.yml
```diff
@@ -1,9 +1,11 @@
...
 
-# End-to-end tests, build-dependent (consume the UMD artifact). Two matrices:
-# the frozen Puppeteer suite over theme × bundle (UMD / UMD.min), and the new
-# Playwright suite over theme. Per-leg gating keeps the exact old scope
-# behaviour (UMD legs on the e2e scope, UMD.min on the e2e-min scope).
+# End-to-end tests, build-dependent (consume the UMD artifact). Two matrices,
+# 1:1 with each other: the frozen Puppeteer suite and the new Playwright suite,
+# both over theme × bundle (UMD / UMD.min). Per-leg gating keeps the exact old
+# scope behaviour (UMD legs on the e2e scope, UMD.min on the e2e-min scope).
+# The FULL bundles (handsontable.full[.min].js) are deliberately not covered
```
budnix: This line says the FULL bundles are not covered here, but the `UMD.min` leg right below loads `handsontable.full.min.js` — the fixture maps `full-min` to that exact file. Can I ask whether the intent was "plain `handsontable.full.js` and `handsontable.min.js` are not covered"? As written, someone could read this and delete the `-min` leg as accidental coverage, which would drop the only minified bundle we test here.

.github/workflows/e2e.yml
```diff
@@ -72,6 +75,9 @@ jobs:
       fail-fast: false
       matrix:
         theme: [main, horizon, classic]
+        bundle:
+          - { label: 'UMD', suffix: '' }
+          - { label: 'UMD.min', suffix: '-min' }
```
budnix, line 86: Adding the bundle axis takes this job from 3 legs to 6, and each leg starts its own container, downloads `dist.tar.gz`, and runs `pnpm install`. Curious whether you weighed `--project=e2e-<theme> --project=e2e-<theme>-min` in one job instead — same coverage, half the setup. The tradeoff is one merged report and one artifact per theme instead of per leg, so if separate failure reports were the point, that's a fine reason to keep it as is.

tests/playwright.config.ts
```diff
@@ -7,18 +7,30 @@ import type { TestOptions } from './fixtures/test';
...
+ * NOT here — extra coverage belongs to the nightly on develop (DEV-2058).
  *
  * Version parity rule: `@playwright/test` here and the CI container image
  * (`mcr.microsoft.com/playwright:v<same>-noble`) bump together, never apart, so
  * local, CI, and baseline generation render identically.
  */
 const E2E_THEMES = ['main', 'horizon', 'classic'] as const;
+// 1:1 with the Puppeteer bundle legs: UMD (dist/handsontable.js, `test:e2e`)
+// and UMD.min (dist/handsontable.full.min.js, `test:production`). The
+// un-suffixed projects run the plain UMD so existing local commands and hooks
```
budnix, line 28: Worth calling out that this changes what a local run covers: `--project=e2e-main` used to load `handsontable.full.min.js`, and now it loads `handsontable.js`. `tests/package.json` `test:e2e`, `scripts/pre-push.mjs:147`, and `scripts/claude/stop.mjs:130` all pin `e2e-main`, so no local gate ever runs a `-min` project. If min is meant to be CI-only that's fine — saying it here would stop the next person assuming the hooks cover it.

tests/playwright.config.ts
```diff
@@ -7,18 +7,30 @@ import type { TestOptions } from './fixtures/test';
...
+// (`--project=e2e-main`) keep working.
+const E2E_BUNDLES = [
+  { bundle: 'umd', suffix: '' },
```
budnix, line 34: One to check: `scripts/e2e-run-cache.mjs:67` hashes only `handsontable/dist/handsontable.full.min.js` to decide whether the environment changed. Now that `e2e-main` loads `handsontable.js`, someone who rebuilds with just `build:umd` (which leaves the min files alone) gets an unchanged hash, so the pre-push hook skips the spec as "already proven green" and pushes it without ever running it against the new bundle. Hashing both files there would keep that gate honest.

tests/playwright.config.ts
```diff
@@ -7,18 +7,30 @@ import type { TestOptions } from './fixtures/test';
...
+  { bundle: 'full-min', suffix: '-min' },
```
budnix, line 35: These keys have to match the allowlist in `grid.html` / `walkontable.html` exactly, and a mismatch fails quietly — the fixture falls back to `handsontable.js`, so a typo like `full-mim` here would make the `-min` legs test the base bundle and still report green. Since nothing asserts which file actually loaded, would it be worth having the fixture throw when `?bundle=` is set to a value it doesn't know? Then a wrong key is one red job instead of six green ones.

tests/fixtures/test.ts
```diff
@@ -3,22 +3,26 @@ import { test as base, expect } from '@playwright/test';
...
- * Declares the `theme` option so every spec runs across the Handsontable
- * themes (main/horizon/classic) via the per-theme projects in
- * playwright.config.ts — the same theme coverage the Puppeteer e2e matrix
- * gives the legacy suite. A spec authored against this `test` needs no theme
- * awareness: it is parametrized automatically by the active project, and the
- * page objects append the active theme to the fixture URL.
+ * Declares the `theme` and `bundle` options so every spec runs across the
+ * same matrix the Puppeteer e2e suite covers: themes (main/horizon/classic)
+ * × bundles (`umd` = dist/handsontable.js, `full-min` =
+ * dist/handsontable.full.min.js — the same files the Puppeteer `test:e2e` and
```
budnix: Small accuracy point on "the same files the Puppeteer `test:e2e` ... legs load": that runner also loads `dist/languages/all.js` (`handsontable/.config/test-e2e.js:101-104`), which the fixture doesn't. Also useful for whoever writes the next spec — on the `umd` legs the base bundle has no HyperFormula, so a formulas spec won't throw, it just logs a warning and the plugin stays off. A line here saying formulas specs belong on the `-min` legs would save someone that debugging session.

budnix (commented):
A few notes from reading the diff — mostly about the comments and docs around the new bundle axis, plus one real gap in the pre-push cache.

One thing with no diff line to hang off: `tests/README.md:37` still calls `npm run test:e2e` "just the functional e2e project" (singular) and documents the theme axis only. There are six projects now and that script runs one of them, so a short update there would help the next spec author.

.github/workflows/e2e.yml
danielzytohoc: Right — the sentence predates the min-leg fix (4861ea495, `umd-min` → `full.min.js`) and never got updated. Reworded in 5fff1b6fd: the header now says the UMD.min legs load `handsontable.full.min.js` (the exact `test:production` file) and the not-covered set is `handsontable.full.js` + `handsontable.min.js` (nightly, DEV-2058).

.github/workflows/e2e.yml
danielzytohoc: We weighed it and kept one job per leg, for two reasons: a min-only failure stays a named red job in the checks list (no report digging), and six visible legs mirror the Puppeteer matrix 1:1 — which is the point of this PR. At ~15 specs the job is setup-dominated either way, so the isolation wins for now; worth revisiting once migration grows this suite and shrinks Puppeteer's. Documented the choice in the e2e.yml header (5fff1b6fd) so it reads as deliberate rather than accidental.

tests/playwright.config.ts
danielzytohoc: Yes — min is deliberately CI-only on the local path (hooks stay fast, and the un-suffixed projects keep existing commands working). Now written down in 5fff1b6fd: the config comment states the local gates (pre-push, Stop hook, `test:e2e`) cover plain UMD only — including that this *flipped* with the bundle axis — and tests/README.md documents the six-project matrix plus how to run a single `-min` leg by hand.

tests/playwright.config.ts
danielzytohoc: Confirmed real — sharpest catch in the review, thanks. The hash was correct while full.min was the only bundle the fixtures could load; this PR broke that assumption. Fixed in 5fff1b6fd: `envHash` now hashes BOTH `handsontable.js` and `handsontable.full.min.js`, and either missing → `''` (never skip). Regression-tested in `scripts/__tests__/e2e-run-cache.test.mjs`: rebuild-base-only invalidates the cache; missing-base never skips (6/6 green).

tests/playwright.config.ts
danielzytohoc: Agreed — the silent fallback was exactly the class this PR exists to kill (a mislabeled green leg). Done in 5fff1b6fd, slightly wider than asked: both fixtures throw on a present-but-unknown `?bundle=` **and** `?theme=` (same silent-green shape); absent params keep the defaults so the pages stay browsable by hand, and `umd` became an explicit allowlist key instead of riding the fallback. Negative path verified: `?bundle=full-mim` → page error `Unknown ?bundle= value: "full-mim"`, zero cells rendered, so the leg goes red on the first `toBeVisible`.

tests/fixtures/test.ts
danielzytohoc: Both points taken in 5fff1b6fd: the comment now says "the same Handsontable builds" with the `dist/languages/all.js` caveat spelled out (an i18n spec's fixture must load it explicitly), plus a dedicated paragraph on formulas specs — the `umd` legs' base bundle ships no HyperFormula and the plugin silently stays off, so a formulas fixture loads HF as an external script (the DEV-2183 `formulas-grid.html` is the reference).

danielzytohoc:
All six inline threads + the review-body README note addressed in 5fff1b6fd — replies inline on each thread. Summary: env hash now covers both bundles (regression-tested), fixtures fail loud on unknown ?bundle=/?theme= values (negative path verified), e2e.yml header wording corrected + the per-leg-jobs choice documented, local-gates-run-plain-UMD-only written into the config comment and README, and the test.ts comment carries the languages/all.js caveat + the formulas-need-external-HyperFormula guidance. The stacked #13164 gets the same fixture guard for its three new fixtures next.

danielzytohoc:
Develop moved twice while addressing the review, so two more merges landed here beyond the review fixes:

- **DEV-2189 × the env-hash fix**: `e2e-run-cache.test.mjs` import conflict resolved as the union — the worktree cache fix and the both-bundles hash compose (8/8 script tests).
- **DEV-2193's walkontable test-layout rename** (`WalkontablePage` → `pages/walkontable/OverlaysPage`, fixture → `demo/walkontable/overlays.html`) resolved with the bundle threading kept; the rename-merge preserved the fail-loud allowlist.
- **Three fixtures that landed on develop mid-flight against the old pattern** (`license-bar`, `menu-scroll`, `frozen-tall-cell`) are now converted to the bundle axis too — same fail-loud allowlist + threading, so no fixture in the tree hardcodes a bundle anymore. Verified locally on both legs (48/48 full suite, 40/40 walkontable, 36/36 license/menu) against freshly built bundles.

Side note validating the cache fix in this PR: my first local run of the new specs failed on `e2e-main` purely because the local `handsontable.js` predated the new features while `full.min` was fresh — precisely the split-staleness scenario the two-bundle env hash now guards against.

tests/fixtures/test.ts
```diff
@@ -3,22 +3,34 @@ import { test as base, expect } from '@playwright/test';
...
+ * explicitly), via the projects in playwright.config.ts. A spec authored
+ * against this `test` needs no theme or bundle awareness: it is parametrized
+ * automatically by the active project, and the page objects append both to
+ * the fixture URL.
+ *
+ * Formulas specs: the `umd` legs run the base bundle, which ships NO
+ * HyperFormula — the plugin logs a warning and silently stays off. A formulas
+ * fixture must load HyperFormula as an external script beside the bundle
+ * (the DEV-2183 `formulas-grid.html` fixture is the reference), or the spec
+ * fails mysteriously on the `umd` legs only.
```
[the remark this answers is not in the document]
danielzytohoc: Fair catch — this is our own mandatory-checklist item 3. Done in 6cf844acc: new `tests/AGENTS.md` (+ `CLAUDE.md` symlink) carries the tier's never-get-wrong list — the six-leg theme × bundle matrix, local gates running plain UMD only, the fail-loud fixture contract, external HyperFormula / `languages/all.js` on the `umd` legs, and the two-bundle cache hash. Root `AGENTS.md` gains the routing and workspace rows, and the `handsontable-playwright-e2e` skill's run-scope section now describes the six-leg reality plus a fifth non-negotiable rule for threading the axis.

budnix (dismissed):

danielzytohoc:
Post-approval note — one small commit (97a79565d) fixing the flaky leg from run 31676309908: `frozen-column-row-heights.spec.ts` (arrived from develop with DEV-2193) read `masterFirstRenderedRow()` with bare non-retrying expects right after a scroll — `scrollToRowAtTop` only waits for the scroll to START (`scrollTop > 0`), so the read raced the re-render that commits the new band a frame later (expected 4, got 0; green on retry; `failOnFlakyTests` hard-fails pass-on-retry by design). All four bare band reads are now `expect.poll` with the same expected values — waiting added, nothing weakened. Verified 180/180 at `--repeat-each=5` on `e2e-main` + `e2e-main-min`. This PR's six legs just roll the dice on that race twice as often as develop's three — the race itself predates the matrix.

budnix (approved):

Issue #6571, opened 2019-12-16
minRows doesn't work for update settings


budnix:
### Description

It seems that `minRows` option is untouchable when we try to decrease its value while update settings.

### Steps to reproduce

1. Just click the demo https://jsfiddle.net/a8709hj3/;
2. There should be visible 4 rows, instead, 6 are rendered.

### Demo

https://jsfiddle.net/a8709hj3/

### Your environment
* Handsontable version: 7.3.0
* Browser Name and version: Chrome
* Operating System: macOS

AMBudnik:
Still replicable in `8.0.0.-beta1` https://jsfiddle.net/AMBudnik/x79rkqja/

Issue #5103, opened 2018-05-17
Resize editor bug while changing datasource


budnix:
### Description

The editor doesn't resize when datasource was changed. This only happens when an editor is open.

### Steps to reproduce

1. Open editor of the A1 cell;
2. After 3 seconds the `setTimeout` function change the data source and re-render the table;
3. As a result, you'll see not resized editor.

### Demo

https://jsfiddle.net/8ffpsqt6/21/

![kapture 2018-05-17 at 11 10 51](https://user-images.githubusercontent.com/571316/40168150-0a9dfc0e-59c3-11e8-8d0e-be76aefd9575.gif)

### Your environment
* Handsontable version: 3.0.0
* Browser Name and version: any
* Operating System: any

AMBudnik:
Also replicable in 7.4.2 https://jsfiddle.net/AMBudnik/b9jz7peg/

swistach:
Since v8 there is `setSourceDataAtCell` method to manipulate data in that way: https://jsfiddle.net/Lkbo3n0p/
and the new version seems to work as expected.

PR #13164, opened 2026-08-05
DEV-2183: Skipped-test burn-down (1/N) — delete 9 dead skips, fix 4 specs

danielzytohoc:
### Context

The PR is the DEV-2183 skipped-test burn-down — the debt that must reach zero before `handsontable/no-skipped-test` and `handsontable/require-assertion-in-test` flip from `warn` to `error` (DEV-2059 step 5 escalation). Skip sites: **31 → 11 final** (per slice: 21 after slice 1, 19 after slice 2, 11 after slice 3 — the per-slice sections below accreted as the work landed). Full per-test triage with evidence: see the table on DEV-2183.

**Was stacked on #13163 (DEV-2190) — now merged**, so this diff shows only the burn-down changes. The three new fixtures + all four migrated specs thread the `bundle` option end-to-end, so the migrated coverage genuinely runs on both matrix legs (`umd` = `handsontable.js`, `full-min` = `handsontable.full.min.js`). Two CI-environment fixes landed during the matrix soak: HyperFormula is served from this package's own `node_modules` (the old `/handsontable/node_modules/…` path exists locally but 404s under CI's filtered install — masked before because every leg ran full.min, which ships its own HF global), and the fixture static server now refuses `node_modules` outside `tests/` so a phantom-dependency fixture fails locally the same way it fails in CI. Both rules are recorded in `tests/AGENTS.md`.

**Deleted (9 skipped tests that can never meaningfully run):**

- `settings/ariaTags.spec.js` (2) — asserted that `updateSettings({ ariaTags })` toggles a11y attributes; the current contract is the opposite (init-only — the active test at the same describe asserts it **throws**).
- `formulas/__tests__/initialization.spec.js` (1) — waited since 2021 for HyperFormula to throw a duplicate-function-name error upstream; HF never shipped it.
- `autoRowSize.spec.js` (1) — tested a `columns.height` option that has never existed.
- `exportFile/.../formulas.spec.js` (1) — `custom` summary as live Excel formula: permanent by-design limitation; the real fallback behavior is actively tested directly above it.
- `exportFile/.../cell-types.spec.js` (1) — Excel-for-Mac UI categorization quirk; the OOXML is spec-compliant and there is no programmatic oracle.
- `__tests__/dom.spec.js` (2) — outerHeight/offset of a table **with `<caption>`** against hardcoded stylesheet pixels; skipped since 2019, Handsontable never renders captions.
- `autocompleteEditor.spec.js` (1) — per its own comment (#5103) it "never properly tests the case" and only passed historically due to a sleep-timing artifact.

**Fixed (4 specs, nothing weakened):**

- `settings/outsideClickDeselects.spec.js` — un-skipped the boolean-config editor test (skipped in the 2018 IME rework); it passes today, matching its active function-config twin.
- `numericEditor.spec.js` (2) — `require-assertion-in-test` false positives: the `expect` lived inside a shared helper. The helper now returns the corner display value and both tests assert it visibly — semantics unchanged.
- `stretchColumns.spec.js` — the "should not throw when ResizeObserver fires" survival test now asserts survival explicitly (`window` error listener spy + `expect(onError).not.toHaveBeenCalled()`) instead of implying it with zero assertions.

**Slice 2 (same PR): the fix bucket, triaged by running each skip against today's develop.**

**Revived (2):** the formulas undo/redo pair ("restore previous state" / "redo into the next state" after mixed remove operations) — skipped since the 2026-03 undo/redo desync fix. Both diffs were stale expectations, not behavior bugs: (a) the apostrophe-escaped literal `'=SUM(E5)` survives undo (the old expectations wrongly dropped the escape — the original fixture and the maintained sibling test both use the escaped form); (b) HyperFormula normalizes formula casing on the undo round-trip (`=Sum(a2:a5)` → `=SUM(A2:A5)`), which the maintained sibling test already asserts. Expectations corrected, both unskipped — the spec file runs 20/20.

**Probed, still genuinely broken — re-skipped with dated evidence comments and split into product tasks (4):**

- MergeCells top-overlay collapse — still real (top clone stays at 2-row height, 59px vs 88px main theme). Test rewritten theme-agnostically (`getThemeLayout().overlayHeight()`, matching its siblings) so the fix for **DEV-2194** only flips `xit` → `it`.
- Formulas autofill × trimRows — fill populates one row instead of two and the old expectation predates trimming semantics (asserts 5 visual rows where trimRows leaves 4). Needs a product decision → **DEV-2195**.
- NestedHeaders rowspan nav pair — navigation order across rowspans was never finished after the feature shipped, and running the pair leaks state that fails 3 neighboring specs → **DEV-2196**.

**Slice 3 (same PR): the migrate bucket — all 8 legacy skips replaced by Playwright specs and deleted.**

| Legacy skip (deleted) | Playwright replacement | Result |
|---|---|---|
| copyPaste copy/cut ×4 (empty stubs — simulated events can't drive the real clipboard) | `tests/e2e/clipboard.spec.ts` — real Ctrl/Cmd+C/X and context-menu Copy/Cut against `navigator.clipboard` | 4 active tests, first-ever coverage |
| textEditor "editor should move with the page when scrolled…" | `tests/e2e/editor-window-scroll.spec.ts` + `editor-scroll.html` fixture | regular-cell case **passes** (active); pinned-overlay case is a verified product bug — 184px editor drift after a 200px window scroll — parked as `test.fixme` → **DEV-2201** |
| walkontable RTL overlay pair ("window object as scrollable element", skipped since 2022 "scroll API" note) | `tests/e2e/walkontable/rtl-window-scroll.spec.ts` + RTL window-scroll fixture, rewritten behaviorally (overlay pinning to viewport edges + clone/master column alignment) | **both pass** — the 2022 blocker has since been fixed; 4-year-old skips become live coverage |
| formulas autofill `.area.corner` drag (DEV-99: helper couldn't drag a range handle past the viewport) | `tests/e2e/fill-handle-formulas.spec.ts` + HyperFormula fixture (HF as external script — the DEV-2190 wiring) with a real mouse drag incl. auto-scroll pumping | **passes** — modernized to ISO source dates (the unrun legacy config predates the ISO redesign and renders `#bad-value#` today); `autoInsertRow: false` pins determinism |

Fixture/page-object notes: `grid.html` gains `contextMenu: true`; new page objects follow the POM + overlay-clone-scoping rules; the `?theme=` allowlist-to-literal pattern is preserved in all new fixtures. New specs are stable across `--repeat-each=3`. The `{ page, theme, bundle }` threading is **included** (commit 0ea6677f0, after merging #13163 in): the three new fixtures select the build from the same `?bundle=` allowlist as `grid.html`, and the fill-handle auto-scroll pump now fails with its real cause when exhausted instead of finishing the gesture on an overflow-clipped coordinate.

**Final count: 31 → 11 legacy skip sites**, every survivor blocked on product work with a task: 4× #6571 (DEV-2206), 2× rowspan nav (DEV-2196), mergeCells overlay (DEV-2194), autofill×trimRows (DEV-2195), exportFile currency (DEV-2204), undo×columnMove (DEV-2205), plus the by-design MemoryLeakTest placeholder the lint rule does not flag. An independent review agent pass over the full branch diff was applied (deterministic auto-scroll gate, validity assertions, task-ID references in re-skip comments, retrying clipboard reads, benign-ResizeObserver filtering).

### How has this been tested?

- Bundle threading (post-#13163 merge-in): the four migrated specs pass on **both legs** locally — 16/16 across `e2e-main` + `e2e-main-min` (2 expected `fixme` skips); `fill-handle-formulas` stable at `--repeat-each=3` on the `umd` leg — the first real run of the external-HyperFormula wiring against the base bundle. `tests` lint clean.

- `npm run test:e2e -- --testPathPattern="outsideClickDeselects|numericEditor|stretchColumns"` — 87 specs, all touched tests green. The single failure ("StretchColumns … after window size change", 427 vs ~422 px) is **pre-existing and unrelated**: it fails identically on a pristine `develop` tree locally (macOS scrollbar geometry) and is untouched by this PR.
- Slice 2: `formulas/__tests__/plugins/undoRedo` — 20 specs, 0 failures. `mergeCells` — 103 specs, only pre-existing local-env failure ("high virtualized merged cell", fails identically on pristine develop). `nestedHeaders/__tests__/rowspan` — 20 specs, 0 failures after re-skip.
- Slice 3: full touched-file legacy sweep (17 spec files) — **788 specs, 2 failures, both pre-existing local-env** (verified identical on pristine develop: mergeCells "high virtualized", stretchColumns "window size change"). Walkontable pipeline — 806 specs, 48 failures, **identical 48 on pristine develop** (macOS scrollbar-geometry class; my diff there only deletes two `xit`s). Playwright: full `e2e-main` suite 14 passed + 1 fixme; new specs stable at `--repeat-each=3`; `tests` package lint clean.
- `npx eslint` on all 10 touched files — 0 errors.
- Skip-site count re-measured: 31 → 19 across both slices (`no-skipped-test` warnings 30 → 18; `require-assertion-in-test` 3 → 0 pending CI confirmation).

### Types of changes

- [x] Other: test-debt burn-down (test files only, no library source)

### Related issue(s):

1. DEV-2183 (slice 1)
2. DEV-2059 (unblocks step 5 escalation), DEV-2055

### Affected project(s):

- [x] `handsontable` 
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2183

---

> [!NOTE]
> **Low Risk**
> Changes are confined to tests, fixtures, and local e2e cache/static-server tooling; production bundle behavior is not modified in this diff.
> 
> **Overview**
> **DEV-2183 burn-down:** legacy Jasmine skip sites drop from **31 → 11** (test-only; no library source changes). **Nine** permanently meaningless `xit`s are removed (ariaTags `updateSettings`, duplicate HF registration, `columns.height`, export quirks, caption DOM, flaky autocomplete). **Four** specs are fixed or strengthened: `outsideClickDeselects` editor typing is active again; `numericEditor` asserts corner display in the tests instead of a helper-only `expect`; `stretchColumns` ResizeObserver survival is asserted via an error listener; formulas **undo/redo** mixed-remove pair is unskipped with corrected escaped-formula and casing expectations.
> 
> **Still skipped with dated comments + tasks:** mergeCells top overlay (DEV-2194), autofill × trimRows (DEV-2195), nestedHeaders rowspan nav (DEV-2196), plus other pre-existing product debt.
> 
> **Playwright replacements** for skips the Jasmine harness could not drive: `clipboard.spec.ts` (real clipboard + context menu); `editor-window-scroll.spec.ts` (regular cell passes; frozen-corner parked as **`test.fixme`** → DEV-2201); `walkontable/rtl-window-scroll.spec.ts` (window-scroll overlay pinning); `fill-handle-formulas.spec.ts` (range `.area.corner` drag + auto-scroll) with **`formulas-grid.html`** loading HyperFormula from **`tests/node_modules`** on UMD legs.
> 
> **Tier plumbing:** `tests/package.json` pins **hyperformula 3.3.0**; `static-server.mjs` serves only `tests/node_modules` (CI parity); `e2e-run-cache.mjs` hashes HF + `tests/package.json`; ESLint/`AGENTS.md` document attributable `test.fixme`; `grid.html` enables `contextMenu` for clipboard tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b5d65af6c0903fedbb6b2edf780056341a098bf0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2055 QA overhaul — Pillar 1: dev-owned test enforcement (retire the manual "testing" gate)](https://app.clickup.com/t/9015210959/DEV-2055)
Task linked: [DEV-2059 Pipeline: branch-protection cutover + flip enforcement warn→block](https://app.clickup.com/t/9015210959/DEV-2059)
Task linked: [DEV-2183 Escalation burn-down 1: triage the 31 skipped tests, flip no-skipped-test to error](https://app.clickup.com/t/9015210959/DEV-2183)

qunabu:
Task linked: [DEV-2194 MergeCells: top overlay height does not expand when a merge spans past fixedRowsTop](https://app.clickup.com/t/9015210959/DEV-2194)
Task linked: [DEV-2195 Autofill × TrimRows × Formulas: define and fix fill behavior across trimmed rows](https://app.clickup.com/t/9015210959/DEV-2195)
Task linked: [DEV-2196 NestedHeaders: finish keyboard navigation order across rowspan headers](https://app.clickup.com/t/9015210959/DEV-2196)

qunabu:
Task linked: [DEV-99 The selection movement doesn't scroll the viewport anymore](https://app.clickup.com/t/9015210959/DEV-99)
Task linked: [DEV-2190 Playwright E2E matrix parity with Puppeteer: add the bundle dimension (3 → 6 legs)](https://app.clickup.com/t/9015210959/DEV-2190)
Task linked: [DEV-2201 Editor does not track cells in pinned overlays when the window scrolls](https://app.clickup.com/t/9015210959/DEV-2201)
Task linked: [DEV-2204 ExportFile: write built-in currency numFmtIds via OOXML post-processing](https://app.clickup.com/t/9015210959/DEV-2204)
Task linked: [DEV-2205 UndoRedo: undo of column removal under manualColumnMove restores wrong data](https://app.clickup.com/t/9015210959/DEV-2205)
Task linked: [DEV-2206 Grid does not shrink when min-row/col settings decrease via updateSettings (#6571)](https://app.clickup.com/t/9015210959/DEV-2206)

danielzytohoc:
Review follow-up applied (all three findings from the quality-direction review):

- **Bundle threading (the Medium)** — #13163 is merged into this branch (it lands first) and commit 0ea6677f0 threads `{ page, theme, bundle }` through the three new fixtures + four migrated specs, using the exact `?bundle=` allowlist-to-literal pattern from `grid.html`. The migrated coverage now genuinely runs on both matrix legs. Verified locally: 16/16 across `e2e-main` + `e2e-main-min`, `fill-handle-formulas` ×3 stable on the `umd` leg (first real exercise of the external-HyperFormula wiring against the base bundle).
- **Pump-exhaustion guard (Low)** — `FormulasGridPage.dragAreaFillHandleToCell` now throws with its real cause if auto-scroll never brings the target row into the holder view, instead of finishing the gesture on an overflow-clipped coordinate.
- **Description arithmetic (Low)** — the final 31 → 11 count now leads, with the per-slice timeline in parentheses.

Merge order: squash-merge only after #13163 is on develop; this diff then recomputes down to the burn-down changes.

danielzytohoc:
CI red on the three UMD legs (run 31680388914) diagnosed and fixed in f533a5169 — a CI-only environment gap, not a behavior failure: the formulas fixture loaded HyperFormula from `/handsontable/node_modules/…`, which exists locally (full workspace install) but 404s in CI because the Playwright job installs only `--filter handsontable-tests`. The `umd` legs then had no `HyperFormula` global (the fixture init throws, no grid renders), while the `-min` legs masked it — `full.min` ships its own HF global. Fix: `hyperformula` is now a devDependency of `handsontable-tests` (same `^3.0.0` range as the core package, pnpm dedupes to one copy) and the fixture loads `/tests/node_modules/…`, which the filtered install provides. `tests/AGENTS.md` records the rule: fixture-served libraries must be this package's own dependencies. Verified: fill-handle ×3 repeats + full suite green on both local legs; the develop merge is included so the branch is current post-#13163.

tests/fixtures/pages/FormulasGridPage.ts
```diff
@@ -0,0 +1,129 @@
...
+    await this.page.mouse.move(target.x + target.width / 2, target.y + target.height / 2, { steps: 4 });
+    await this.page.mouse.up();
+  }
+
+  /**
+   * Whether a cell's box lies fully inside the master holder's visible area
+   * (not merely rendered-but-clipped below its overflow edge).
+   */
+  async cellInHolderView(row: number, col: number): Promise<boolean> {
+    const cellBox = await this.cell(row, col).boundingBox();
```
budnix, line 120: This can hang the pump loop on line 87 before it sends a single move. `boundingBox()` calls `waitForSelector({ state: 'attached' })` first, and it only returns `null` for a cell that is attached but hidden — a cell that was never rendered makes it wait instead. The grid is 130px tall with 7 rows, so `cell-6-4` may not be in the DOM on the first check, and with no `actionTimeout` in `playwright.config.ts` that wait runs to the 20s test timeout: auto-scroll never starts and the failure names the wrong cause.

```suggestion
    if (await this.cell(row, col).count() === 0) {
      return false;
    }

    const cellBox = await this.cell(row, col).boundingBox();
```

tests/fixtures/pages/FormulasGridPage.ts
```diff
@@ -0,0 +1,129 @@
...
+    const holderBox = await this.grid.locator('.ht_master .wtHolder').boundingBox();
+
+    return Boolean(cellBox && holderBox &&
+      cellBox.y + cellBox.height <= holderBox.y + holderBox.height + 1);
```
budnix, line 125: This only bounds the cell's bottom edge, so a cell that scrolled up past the top of the holder still counts as "in view". The pump loop would then stop early and the gesture would land above the visible area.

```suggestion
      cellBox.y >= holderBox.y - 1 &&
      cellBox.y + cellBox.height <= holderBox.y + holderBox.height + 1);
```

tests/fixtures/demo/formulas-grid.html
```diff
@@ -0,0 +1,95 @@
...
+    the fixture works identically under every leg — the `umd` leg's base
+    bundle ships no built-in HyperFormula (the DEV-2190 wiring).
+  -->
+  <div id="grid" data-testid="grid"></div>
+
+  <!-- HyperFormula comes from THIS package's own node_modules (declared in
+       tests/package.json): CI installs only the filtered handsontable-tests
+       workspace, so a path into another package's node_modules would 404
+       there while passing locally against a full install. -->
+  <script src="/tests/node_modules/hyperformula/dist/hyperformula.full.min.js"></script>
```
budnix: On the three `-min` legs this loads two engines and uses the wrong one. `handsontable.full.min.js` ends with `window.HyperFormula = ig` and it loads after this tag, so it overwrites the global the external script just set — the page downloads one engine and then silently runs on the one baked into the bundle. That makes the comment above ("works identically under every leg") not quite true. Gating this tag on `window.htBundle === 'handsontable.js'` would give each leg exactly one engine.

tests/package.json
```diff
@@ -12,6 +12,7 @@
   "devDependencies": {
     "@playwright/test": "catalog:",
     "@typescript-eslint/parser": "8.61.0",
-    "eslint": "8.57.1"
+    "eslint": "8.57.1",
+    "hyperformula": "^3.0.0"
```
budnix: These two `^3.0.0` ranges did not stay in lockstep — the lockfile now carries both `hyperformula@3.3.0` (for `handsontable`) and `hyperformula@3.4.0` (for `tests`), at lines 8949 and 8952. So the `umd` legs run 3.4.0 while `full.min` ships 3.3.0, and an assertion that depends on engine behavior can be green on one leg and red on another. Pinning the exact version stops the drift; moving `hyperformula` into the pnpm catalog and using `catalog:` in both packages would be the more durable version of the same fix.

```suggestion
    "hyperformula": "3.3.0"
```

tests/AGENTS.md
```diff
@@ -34,6 +34,14 @@ Visual regression is a separate package (`visual-tests/`). Task workflow: the
...
+  `/tests/node_modules/…` — CI installs only the filtered `handsontable-tests`
+  workspace, so a path into any other package's `node_modules` does not exist
+  there (and the static server refuses it locally too, for CI parity). Keep
+  such ranges in lockstep with the owning package — `hyperformula` matches the
+  `handsontable` package's `^3.0.0`; if the ranges ever diverge, the lockfile
+  grows a second hyperformula entry and the `umd` legs test a different engine
+  version than the one baked into `full.min`.
 - The green-run cache (`scripts/e2e-run-cache.mjs`) hashes BOTH bundles plus
   every file under `fixtures/`; rebuilding either bundle re-runs affected
   specs. Do not narrow that hash.
```
budnix: The new fixture pulls its engine from `/tests/node_modules/hyperformula/…`, and that path sits outside the hash. `envHash()` covers both bundles, the four theme CSS files, everything under `tests/fixtures/`, and `playwright.config.ts` — but not `tests/package.json` or the installed module. Reinstall a different HyperFormula 3.x and the hash is unchanged, so the pre-push hook skips `fill-handle-formulas.spec.ts` as already green against an engine it never ran on. Adding `tests/package.json` to `envHash()` would close it.

tests/e2e/editor-window-scroll.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+    grid = new EditorScrollPage(page, theme, bundle);
+    await grid.goto();
+  });
+
+  // Known product bug (verified 2026-08-07, migrated from the legacy skip):
+  // when the edited cell lives in a PINNED overlay, the editor scrolls away
+  // with the page instead of holding to the pinned cell — 184px drift after a
+  // 200px window scroll (the 16px body margin pins the overlay late). The
+  // regular-cell case below passes. Unskip with the fix (DEV-2201).
+  test.fixme('tracks a frozen-corner cell while the window scrolls', async () => {
```
budnix, line 30: The `no-restricted-syntax` rules in `tests/.eslintrc.cjs` match `.only` and `.skip` but not `fixme`, so this one is invisible to the tier's own gate. In a PR about burning down skips, that means any count built from that rule misses it. Did you want to add `fixme` to the rule, or is DEV-2201 the place this gets tracked?

budnix (commented):
Ran the three touched legacy Jasmine areas locally and they are all green — `outsideClickDeselects` (19), `formulas/__tests__/plugins/undoRedo` (20), and `stretchColumns|numericEditor` (73), 0 failures. The undoRedo expectation edits check out too: the fixture at `handsontable/src/plugins/formulas/__tests__/helpers/fixtures.js:25` does hold `'\'=SUM(E5)'`, and an untouched assertion in the same spec already uses the uppercase form.

I could not run the Playwright tier (no `tests/node_modules` here), so the comments below are read from the code rather than from a failing run.

Also worth knowing: the cursor[bot] comment on `formulas-grid.html` is stale — it flags the old `handsontable/node_modules/...` path, which the last two commits already moved to `/tests/node_modules/...`.

tests/fixtures/pages/FormulasGridPage.ts
danielzytohoc: Took the suggestion verbatim in b5d65af6c — `count() === 0` bails out before the `boundingBox()` attachment wait can stall the pump. You're right that it's latent today (the 7-row fixture renders every row, clipped not virtualized) but the helper promised more than it delivered, and with no `actionTimeout` the failure mode was a 20s timeout naming the wrong cause.

tests/fixtures/pages/FormulasGridPage.ts
danielzytohoc: Done in b5d65af6c — both edges bounded now, matching what the helper's name promises. Can't happen with the current downward gesture, but the next caller shouldn't inherit the trap.

tests/fixtures/demo/formulas-grid.html
danielzytohoc: Confirmed and fixed in b5d65af6c — verified `handsontable.full.min.js` does end with `window.HyperFormula=ig`, so the -min legs downloaded one engine and silently ran another. The external script is now gated on `window.htBundle === 'handsontable.js'` via the same split-tag `document.write` as the bundle loader, and the fixture comment now says what's actually true: every leg runs exactly one engine (umd = the external package, full-min = the baked-in one).

tests/package.json
danielzytohoc: Right, and this disproves my earlier claim that the twin ranges would dedupe — pnpm resolves each importer at its own time, so identical `^3.0.0` ranges still locked 3.3.0 vs 3.4.0. Pinned to `3.3.0` in b5d65af6c; the lockfile is back to a single hyperformula entry and both legs now exercise the same engine version. `tests/AGENTS.md` records the invariant (one lockfile entry, pin to the owning package's locked version — a matching range is not lockstep), and names the pnpm catalog as the durable upgrade once the core package can take that change.

tests/AGENTS.md
danielzytohoc: Done in b5d65af6c, slightly wider than suggested: `envHash()` now covers `tests/package.json` AND the served artifact itself (`tests/node_modules/hyperformula/dist/hyperformula.full.min.js`) as required inputs — package.json alone would miss a reinstall that changes the resolved version without an edit (or a tampered/partial install). Either missing → `''` → never skip. Regression-tested: reinstalling a different engine or editing the manifest invalidates the cache (scripts/__tests__/e2e-run-cache.test.mjs, 9/9).

tests/e2e/editor-window-scroll.spec.ts
danielzytohoc: Both, as it turns out: DEV-2201 tracks this instance, and b5d65af6c adds the enforcement so the class is counted — `tests/.eslintrc.cjs` now bans bare `test.fixme` at error; the allowed form is an eslint-disable line naming the tracking task (`// eslint-disable-next-line no-restricted-syntax -- DEV-1234: <why>`), which this spec's fixme now carries. Every parked test is grep-countable and attributable, same convention as the dated re-skip comments in the Jasmine tier. Policy recorded in tests/AGENTS.md.

danielzytohoc:
All six review threads addressed in b5d65af6c (replies inline). Beyond the threads, two things for anyone running the suites locally after this merges:

**Playwright**: one `pnpm install` from the repo root is the entire setup (plus `npx playwright install` once for browsers). If `tests/node_modules` is missing — the exact symptom from this review — the checkout's install predates this package or was filtered; the README now has an "install seems missing" section and the static server prints an actionable warning at startup instead of a bare 404 mid-run.

**Puppeteer**: needs nothing beyond the same root `pnpm install` (the Chromium download runs via puppeteer's allowed install script) — `npm --prefix handsontable run test:e2e` / `test:production` / `test:walkontable`. Also documented in tests/README.md.

And one honest correction on my own earlier claim: the twin `^3.0.0` ranges did NOT dedupe — the lockfile carried 3.3.0 and 3.4.0 side by side, exactly as flagged. Pinned now, single lockfile entry, invariant recorded in tests/AGENTS.md.

budnix (approved):

Issue #11031, opened 2024-06-14
After adding a border, insert a line above the border without moving it down


lindanni:
### Describe the bug

After adding a border, insert a line above the border without moving it down。
Same as problem #3296
https://github.com/handsontable/handsontable/issues/3296

### Provide a link to the demo with the bug reproduction

https://github.com/handsontable/handsontable/issues/3296

### Handsontable version

14.2.0

### Your environment

14.2.0

sl01k:
Fixed in #13166 — borders now follow their cell when rows or columns are inserted, removed, or moved, and stay removable from the context menu afterward. Shipping in 18.1.0.

Issue #6679, opened 2020-02-03
Custom borders rendered wrong when provided using range with a `border` object


warpech:
### Description
When the custom borders configuration is provided via a range, there are the following problems:

- the default line style and color is used
- if several ranges are overlapping, only the topmost range is rendered

For example, consider the following configuration:

```js
{
  customBorders: [{
        range: {
          from: {
            row: 1,
            col: 1
          },
          to: {
            row: 3,
            col: 3
          }
        },
        border: {
          width: 2,
          color: '#C6E0B4'
        },
        left: {},
        right: {},
        top: {},
        bottom: {},
      },
      {
        range: {
          from: {
            row: 2,
            col: 2
          },
          to: {
            row: 3,
            col: 3
          }
        },
        border: {
          width: 2,
          color: '#548235'
        },
        left: {},
        right: {},
        top: {},
        bottom: {},
      },
      {
        row: 3,
        col: 3,
        border: {
          width: 3,
          color: '#F4B084'
        },
        left: {},
        right: {},
        top: {},
        bottom: {},
      }
    ]
}
```

#### Expected 

I would expect this configuration to render like that (screenshot from Excel file 
[nice-border-priority.xlsx](https://github.com/handsontable/handsontable/files/4148521/nice-border-priority.xlsx)
):

![image](https://user-images.githubusercontent.com/566463/73664156-3a4b9e00-469f-11ea-9f1b-99eb578c6b5e.png)

#### Actual

In Handsontable, this configuration renders as:

![image](https://user-images.githubusercontent.com/566463/73667485-9fee5900-46a4-11ea-8769-5d68a18ccf63.png)

### Demo
https://jsfiddle.net/u7rzgjhx/

### Your environment
* Handsontable version: 7.2.2, probably the problem is much older
* Browser Name and version: any
* Operating System: any

warpech:
Problem still valid in 8.x SVG branch (https://github.com/handsontable/handsontable/pull/6157/commits/44c3704fbf18bc1c08ce15b5b493e5205f3b9975).

AMBudnik:
The issue is still replicable in v8.2.0 and we do not need the smallest area (orange) to replicate this issue https://jsfiddle.net/handsoncode/pe20q5ca/

![image](https://user-images.githubusercontent.com/10757813/105852492-d3787080-5fe4-11eb-9e34-46bff21edbc6.png)

Tested using Chrome 87, Firefox 84, and Edge on Windows 10.

[...]

Issue #6063, opened 2019-06-18
Borders after insert row are not movedown


lloydz:
### Description
Borders after insert row are not movedown.
![issue](https://user-images.githubusercontent.com/20263049/59651425-51935500-91bc-11e9-979f-667e97349762.png)

### Steps to reproduce

1. set borders of a certain cell
2. insert row before this cell
3. the borders are not move down

### Your environment
* Handsontable version: 6.2.2
* Browser Name and version: chrome
* Operating System: window7

AMBudnik:
Hey @lloydz 

it seems to work fine here https://jsfiddle.net/sb7v0n61/
Can you share a demo?

lloydz:
@AMBudnik 
i guess you may misunderstand my meaning.
I want the new added row with no borders, and the original borders move down by a row step.

![issue2](https://user-images.githubusercontent.com/20263049/59669220-68ea3680-91ec-11e9-895b-0b5605a1c6af.png)

AMBudnik:
Then you'd need to modify the `customBorders` array.
You may find the `{row: 2, col: 2, left: { /*...*/ }` notion an easier one in this case.

lloydz:
@AMBudnik ,

thanks for your reply.

there are 2 problems to modify the customBorders array after a row inserted:
1. the hook 'afterCreateRow' seems triggered before row inserted, is there any other hook for this case?
2. instance.updateSettings({customBorders: newconfigedArray}) won't work as expected evevytime, any other ways to rerender the table with new settings?

lloydz:
besides, insert a row or remove rows also impact on the fomular setting [issue 4654](https://github.com/handsontable/handsontable/issues/4654), any solution about this issue? thanks~

AMBudnik:
We are working on a new formula engine so all the issues related to formulas haven't been scheduled but if you could share your recent progress (perfectly by the JSFiddle) I will check the hooks.

AMBudnik:
Hey @lloydz 

could you share your recent progress? Any news here?

MitchVL:
Is this being worked on ? We are currently testing hot to see if it suits our needs, and this bug is the only thing stopping us from purchasing, since it is essential for our needs. 
Just like lloydz mentioned, the suggested solution does not work.

AMBudnik:
Hey @MitchVL I'm waiting for @lloydz code to check what we can do.

If you have a similar situation please send me the example at support@handsontable.com

[...]

Issue #3296, opened 2016-03-04
set a bottom border for one cell


yilongzhou:
Hello:
       I have been updated to the latest version（0.23.0）.As the realease note:
          Fixed shifting cell meta when user wants to insert a new row or a column #1823

```
   In my case, I found another problem: [ notes  is a problem description]
   (http://jsfiddle.net/zq5fzegx/6/)
```

  Looking forward to your help.

  frank

AMBudnik:
Hi @yilongzhou could you share more details about what's the unwanted behavior in this case?

yilongzhou:
@AMBudnik 
First of all, thank you for your warm reply. I describe my problem in detail as follows：
1.set a bottom border for one cell(row:0, col: 0)
2.insert a new row above(row: 0, which one has the bottom border)
3.question: I expect the bottom border to the next row(row:1, col:0) ，but still in the first row(row:0, col: 0)
Can you understand me now?

AMBudnik:
Reproduced. Border should be transfered with the cell (it should move to 2nd row)

result
![pp](https://cloud.githubusercontent.com/assets/10757813/13566454/fbef827a-e455-11e5-88cf-e04679d17d52.png)

related to https://github.com/handsontable/handsontable/issues/1823

yilongzhou:
@AMBudnik 
Hello, I have updated to the version 0.24.1. please confirm whether #3296  has been resolved.
It is still wrong.

AMBudnik:
Hi @yilongzhou 
Yes, unfortunately this bug is still present. It stays open.

yilongzhou:
@AMBudnik 
Hello, #3296 is ok ?

AMBudnik:
Hi @yilongzhou 
Yes, this bug still occurs for rows and columns.
Please check this demo: http://jsfiddle.net/7809snmh/
It's set to always use the newest version of Handsontable (currently 0.24.2).

If you want to be up-to-date with bug fixes you can track our release notes:  https://github.com/handsontable/handsontable/releases

Also if this issue stops you from achieving your goal you can order a quick bug-fix - to do so write us at hello@handsontable.com

AMBudnik:
Still replicable in version 0.35.0
![shift](https://user-images.githubusercontent.com/10757813/35149297-dd386484-fd15-11e7-8bf8-0ebf477443cf.gif)

AMBudnik:
[...]

PR #13166, opened 2026-08-06
Custom borders: fix structural sync & overlapping-range rendering, add virtualization + progressive apply (#11031, #6063, #3296, #6679)

sl01k:
### Context

The PR fixes several long-standing `CustomBorders` correctness bugs and adds performance work so the plugin scales to large, densely-bordered grids. It groups three milestones:

**Milestone 1 — borders desync on structural changes (#11031, #6063, #3296).** The plugin kept a cell's border state in four layers keyed by fixed visual coordinates, but only the cell meta moved with the cell. Inserting/removing/moving rows or columns left borders rendering on the old cell, and a shifted border's coordinate-derived id went stale so the context-menu **Remove border** silently missed it. Fix: register `afterCreateRow`/`afterRemoveRow`/`afterCreateCol`/`afterRemoveCol` (visual-index shift) and `before/afterRowMove`/`before/afterColumnMove` (physical round-trip snapshot), then re-sync `savedBorders`, the id index, and the rendered selections to the (already-correct) cell meta — recomputing ids fixes the orphaned-remove symptom.

**Milestone 2 — range with a `border` object rendered wrong (#6679).** A range config with a top-level `border` object plus empty sides rendered the 1px black default, and overlapping ranges lost one another's sides. Fix: an empty side inherits the range-level `border` style, and overlapping ranges merge per side. This surfaced pre-existing per-border rendering issues in Walkontable (`border.ts`) that are also fixed: per-side corner geometry (position by the edge's own half-thickness, length by the perpendicular edge's, extend a corner only when the perpendicular edge is drawn) and per-edge z-index so a thicker border stacks above a thinner one deterministically.

**Milestone 3 — performance.** Custom borders were not virtualized: init and per-frame cost grew with the total border count (a border on every cell of a 1000×1000 grid was effectively infeasible). Two changes: (a) the selection manager and the plugin now materialize custom-border selections/DOM only for the rendered viewport (a viewport working set synced on `afterViewRender`), so init is roughly linear, border DOM is bounded by the viewport, and per-frame scroll cost no longer grows with the total; (b) a new opt-in `customBordersProgressive` option applies a large configuration in background batches so the grid is interactive immediately, with a new `afterCustomBordersUpdate` hook signaling completion. A first-touch read-skip during config application also cuts a redundant O(N) meta read (~27% faster init on the stress case).

No breaking changes: `customBordersProgressive` defaults off, the new hook is additive, and existing behavior/API (`getBorders`, `setBorders`, context menu, styles, RTL, frozen overlays) is unchanged.

### How has this been tested?

New unit tests for the pure helpers (index shift, viewport union, `resolveRangeBorderSide`). New E2E coverage lives in the Playwright tier (`tests/e2e/customBorders.spec.ts`, driven through a lab fixture with pinned grid dimensions and shared geometry helpers): structural insert/remove/move for rows and columns (including `count > 1`, insert at the border's own index, drop-on-remove, the orphaned-id regression, batch `moveRows`, undo/redo of inserts and removes, and trimmed/hidden-index cases), #6679 range-`border` and overlapping-range rendering in LTR **and RTL** (edge alignment, no overshoot, deterministic z-order, no stray edge in an inserted gap), the border-DOM working set (header-only-overlay culling, hidden-row re-entry, frozen-pane culling), and the progressive path (deferred application, `true` and `{ chunkSize }` forms, completion hook, cancel-on-replace, structural change mid-flight, disable/destroy mid-flight). The frozen Jasmine `customBorders.spec.js` carries only maintenance updates to pre-existing assertions. Verified manually in a demo (grid up to 1000×1000, all cells bordered) for offset/overlap correctness and scroll behavior.

```shell
npm run test:unit --prefix handsontable --testPathPattern=customBorders
npm run test:e2e --prefix handsontable --testPathPattern=customBorders
npm run test:e2e --prefix handsontable --testPathPattern="(selection|mergeCells)/__tests__"
cd tests && npx playwright test e2e/customBorders.spec.ts
```

All green: customBorders 141 legacy E2E + 18 unit, Playwright 38 tests × 3 themes (main/horizon/classic, no flake across repeated runs), Walkontable 808, selection + mergeCells 1409, core hooks 424. ESLint 0 errors, Stylelint clean, types pass.

### Types of changes

### Related issue(s):

1. #11031
2. #6063
3. #3296
4. #6679

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

- [x] My change requires a change to the documentation, and I have updated it (the Custom cell borders guide now documents `customBordersProgressive` and `afterCustomBordersUpdate`; the option and hook are in the auto-generated API reference).

---

> [!NOTE]
> **Medium Risk**
> Large changes to CustomBorders, Walkontable border rendering, and init/render timing affect core grid behavior, though defaults stay backward-compatible and coverage is extensive.
> 
> **Overview**
> **CustomBorders** is reworked so borders stay aligned with cells after row/column insert, remove, and move (including manual move and context-menu remove after shifts), with model/meta kept in sync and renders timed to avoid corrupting nested headers mid-init.
> 
> Range configs with a top-level **`border`** object now apply that style to empty sides (`{}` / `true`), overlapping ranges merge per side, and Walkontable draws custom edges with **per-edge z-index** (width-based stacking) and **per-edge corner geometry** so mixed-width overlaps align without overshooting hidden sides.
> 
> **Performance:** custom-border selections and DOM are built only for the visible viewport (plus frozen panes), synced on **`beforeViewRender`**; the selection manager skips off-screen custom borders. New opt-in **`customBordersProgressive`** applies large configs in background batches; **`afterCustomBordersUpdate`** fires when borders are complete. External `borders` meta writes and vetoable `beforeSet/RemoveCellMeta` are handled so model and meta do not diverge.
> 
> Docs cover progressive mode; Playwright e2e and unit tests cover structural changes, RTL, frozen areas, progressive paths, and veto/cascade edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2b677a5c54a5568b9284d602a4773f733978f98a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu (dismissed):
LGTM but @adrianspdev or @budnix must approve and @danielzytohoc as well in terms of tests

handsontable/src/plugins/customBorders/__tests__/customBorders.spec.js
```diff
@@ -1252,4 +1260,532 @@ describe('CustomBorders', () => {
...
+        customBorders: [{ row: 3, col: 0, top: GREEN_BORDER }]
+      });
+
+      await alter('insert_row_above', 1, 3);
+
+      const borders = getPlugin('customBorders').getBorders();
+
+      expect(borders[0].row).toBe(6);
+      expect(getCellMeta(6, 0).borders.top).toEqual(GREEN_BORDER);
+      expect(countVisibleCustomBorders()).toBe(1);
```
danielzytohoc: This assertion is the CI failure on `E2E / UMD (theme: horizon)` and `UMD.min (theme: horizon)` (`Expected 0 to be 1`). The root cause: `countVisibleCustomBorders()` counts `.wtBorder:visible` DOM, and under this PR's own virtualization an off-viewport border has no DOM at all. This test inserts 3 rows into a 5-row grid, so the border lands on row 6 of 8 — horizon's taller default row height pushes row 6 out of the default test container's rendered window, so the border is culled and the count is 0. On main/classic it stays rendered, which is why only horizon fails. The model assertions above it pass everywhere — the border did follow the cell.

Please don't fix this by dropping the render assertion — make the viewport deterministic instead: give the grid an explicit `width`/`height` in the `handsontable()` call (the Playwright fixture pins `height: 500` for exactly this reason), or `scrollViewportTo` the target row before counting. Note the sibling tests in this describe (borders landing on rows 2–4) pass on horizon only by geometry luck — they sit near the same cliff, so I'd pin the dimensions for the whole `structural changes` block, not just this test.

---
_Generated by [Claude Code](https://claude.ai/code)_

handsontable/src/plugins/customBorders/__tests__/customBorders.spec.js
```diff
@@ -1252,4 +1260,532 @@ describe('CustomBorders', () => {
     // Cell in the middle of area without borders
     expect(getCellMeta(2, 3).borders).toBeUndefined();
   });
+
+  describe('structural changes (issues #11031, #6063, #3296)', () => {
```
danielzytohoc: New tests **cannot** be added to the Puppeteer/Jasmine suite — it is frozen. The monorepo policy (AGENTS.md, `handsontable/.ai/TESTING.md`) is explicit: new E2E goes to Playwright (`tests/e2e/*.spec.ts`); existing `*.spec.js` files may only be edited to maintain what's already there. Appending ~540 lines of new describe blocks to an existing spec honors the letter of "edit existing specs" but not the intent — this is new coverage, and it belongs in the Playwright tier.

It's not just policy for policy's sake: the one CI failure on this PR comes from exactly the fragility this tier carries (theme-dependent default container geometry), while your own Playwright spec avoided it by pinning fixture dimensions and asserting theme-agnostically. Please move these new describes (structural changes, #6679 rendering, progressive) to `tests/e2e/customBorders.spec.ts` / a dedicated fixture. Updating the two pre-existing `countCustomBorders()` assertions above is fine to keep here — that's genuine maintenance of existing tests.

---
_Generated by [Claude Code](https://claude.ai/code)_

handsontable/src/plugins/customBorders/__tests__/customBorders.spec.js
```diff
@@ -1252,4 +1260,532 @@ describe('CustomBorders', () => {
...
+            Math.abs((r.top + (r.height / 2)) - leftCell.bottom) < 5;
+        })
+        .map(r => r.right)
+        .sort((a, b) => b - a)[0];
+
+      expect(Math.round(bottomEdgeRight)).toBeLessThanOrEqual(Math.round(leftCell.right));
+    });
+  });
+
+  describe('progressive application (customBordersProgressive)', () => {
```
danielzytohoc: The progressive path's riskiest behavior has no coverage: a **structural change during an in-flight progressive load**. The implementation has a dedicated path for it — `#flushProgressiveApply()` drains the queue synchronously before coordinates remap (called from the structural hooks and both move-snapshot hooks), and `#finishProgressiveApply` vs `#cancelProgressiveApply` differ precisely in whether `afterCustomBordersUpdate` fires. None of that is exercised: no test does `alter()` or a row/column move while batches are pending, and nothing pins `setBorders()`/`getBorders()` consistency mid-flight.

Suggested minimum: init with progressive + a small `chunkSize`, call `alter('insert_row_above', ...)` before the completion hook fires, then assert the final model equals the synchronous outcome and the hook fired exactly once.

Two smaller gaps in this describe: the plain boolean form `customBordersProgressive: true` (the form the guide leads with) is never executed — only `{ chunkSize }` and `false` are; and disable/destroy mid-flight (the cancel-without-hook path) is untested.

---
_Generated by [Claude Code](https://claude.ai/code)_

handsontable/src/plugins/customBorders/__tests__/customBorders.spec.js
```diff
@@ -1252,4 +1260,532 @@ describe('CustomBorders', () => {
...
+          end: {},
+        }],
+      });
+
+      await alter('insert_row_above', 1, 1); // the range shifts down to rows 4-7
+
+      // The top edge must stay one straight line: the horizontal border over the top-left corner
+      // cell and over a top middle cell must sit at the same Y. Regression for an id collision that
+      // destroyed the freshly shifted corner selections, dropping their horizontal edge.
+      const horizontalBorderTopIn = (row, col) => {
```
danielzytohoc: This jQuery rect-filtering helper is duplicated in near-identical form three times across the new tests (`horizontalBorderTopIn` here, `lightBottomTopAt` and `outerRightOf` in the overlap test, and the inline filter in the split-range test). If these stay, please extract one shared helper into `__tests__/helpers/utils.js`.

The bigger question is tier: these assert Walkontable border-drawing semantics (per-edge corner geometry, z-index stacking) from the main suite, while `walkontable/test/spec/selection/border.spec.js` — the engine's own runner, faster and theme-independent — got no additions for the `border.ts`/`manager.ts` behavior changes. The per-edge geometry, the hide-gated corner extension, and the headers-only-overlay culling early-out would all be more direct and more robust as Walkontable-tier tests.

---
_Generated by [Claude Code](https://claude.ai/code)_

handsontable/src/3rdparty/walkontable/src/selection/border/border.ts
```diff
@@ -1129,26 +1142,51 @@ class Border {
...
+    //     perpendicular edge.
+    // For uniform-width borders every per-edge delta equals the old shared delta, so regular
+    // selections are unaffected.
+    // A corner is only filled when the perpendicular edge is actually drawn there. A horizontal edge
+    // extends toward the end corner only if this cell has a visible end edge, and toward the bottom
+    // only if it has a visible bottom edge. Without this, a per-cell custom border whose end/bottom
+    // side is hidden (an interior or split-range edge) would still extend into the empty neighbour and
+    // stick out. For regular selections the side settings are absent, so `!undefined?.hide` is `true`
+    // and the extension behaves exactly as before.
+    const extendToEnd = !this.settings.end?.hide;
```
danielzytohoc, line 1163: The new corner geometry (per-edge deltas, hide-gated extensions) runs on the `isRtl ? 'right' : 'left'` axis, but there is **zero RTL coverage** for it anywhere — no `layoutDirection` test in the customBorders spec at all, and the Walkontable `rtl/` specs only cover regular selections, where the uniform-width fallback keeps the old behavior. Mixed-width overlap and hidden-side extension in RTL is new, user-visible behavior with no oracle.

Please add at least one RTL test for the #6679 scenarios (overlapping ranges of different widths + a split range with hidden interior sides) — Walkontable tier or Playwright, per the tier comment on the spec.

---
_Generated by [Claude Code](https://claude.ai/code)_

handsontable/src/plugins/customBorders/utils.ts
```diff
@@ -79,6 +79,139 @@ export function createId(row: number, col: number) {
...
+ * (an empty object `{}`, an empty string, or `true`) inherits the range-level `border` object's
+ * style (width, color, and line style) so it renders with the configured look instead of the
+ * default 1px black. When no range-level `border` is provided the raw side value is kept, which
+ * preserves the previous behavior.
+ *
+ * @param {object} [rawSide] The side value from the range configuration.
+ * @param {object} [rangeBorder] The range-level `border` object shared by all sides.
+ * @returns {object}
+ */
+export function resolveRangeBorderSide(
```
danielzytohoc, line 195: `resolveRangeBorderSide` is the one new pure helper without a unit test (the other four all have them in `utils.unit.js`). Three cheap tests would close real blind spots:

1. The documented `top: true` and `top: ''` inheritance forms — only `{}` is exercised end-to-end.
2. The `cornerVisible` deletion branch below is never exercised: no test range-`border` object carries a `cornerVisible` key, so a regression there is currently invisible.
3. The no-`rangeBorder` passthrough.

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/e2e/customBorders.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+    const result = await page.evaluate(async () => {
+      const hot = (window as any).hot;
+      const plugin = hot.getPlugin('customBorders');
+      const hasBorder = () => plugin.getBorders().some((b: any) => b.row === 10 && b.col === 0);
+
+      hot.alter('remove_row', 10);
+      const afterRemove = hasBorder();
+
+      hot.getPlugin('undoRedo').undo();
+      await new Promise(resolve => setTimeout(resolve, 100));
```
danielzytohoc: Fixed-time wait — this is the pattern the Playwright guide bans (condition-based waits only), and being inside `page.evaluate` it also slips past the weakening detector's `sleep`/`waitForTimeout` regexes. Replace with a condition: return early from the evaluate and use `expect.poll` on `getBorders()` (the culling tests in this file already do exactly that).

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/e2e/customBorders.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+    const result = await page.evaluate(() => {
+      const hot = (window as any).hot;
+      const coords = hot._createCellCoords(5, 5);
+
+      hot.selection.highlight.addCustomSelection({
+        border: { width: 2, color: 'orange' },
+        visualCellRange: hot._createCellRange(coords, coords, coords),
+      });
+
+      const beforeClear = hot.selection.highlight.customSelections.length;
```
danielzytohoc: `beforeClear` is captured and returned but never asserted. If the fixture ever changes so the plugin has nothing in the rendered working set here, `clearBorders()` becomes a no-op and the test goes hollow silently (`afterClear === 1` would still pass). Add `expect(result.beforeClear).toBeGreaterThanOrEqual(2)` so the test proves there was plugin-owned DOM to clear.

---
_Generated by [Claude Code](https://claude.ai/code)_

danielzytohoc (changes requested):
Review focused on the tests shipped for this PR (are they meaningful, and do they cover the implemented features), not on the implementation itself.

The test work is substantively good: intent-driven assertions, no hollow tests, the orphaned-id regression is exercised through the real context menu, the #6679 geometry is asserted at the user-visible level, and the two old `TODO looks like a bug` DOM-count assertions are correctly tightened into the virtualization guarantee. But it's not mergeable as-is — see the inline comments. Summary of what blocks:

1. **CI is red on this PR's own test** — the multi-row insert test fails on horizon (`Expected 0 to be 1`) because its `countVisibleCustomBorders()` assertion encodes a theme-dependent viewport assumption that collides with the virtualization this PR ships. Inline comment has the mechanism and the fix direction (pin the grid dimensions — do not drop the assertion).
2. **New tests cannot be created in the Puppeteer/Jasmine suite.** That suite is frozen; new E2E is Playwright (`tests/e2e/*.spec.ts`). The ~540 new lines of describes need to move. The horizon failure itself is a demonstration of why.
3. **The M1×M3 interaction is untested**: a structural change during an in-flight progressive load has a dedicated implementation path (`#flushProgressiveApply`) and zero tests.
4. **No RTL coverage** for the new border geometry, and no Walkontable-tier tests for the `border.ts`/`manager.ts` behavior changes.

Smaller items (inline): `resolveRangeBorderSide` unit tests (`true`/`''` forms, `cornerVisible` drop), `customBordersProgressive: true` boolean form never executed, fixed 100ms wait in the Playwright undo test, unasserted `beforeClear` in the ownership test, duplicated jQuery geometry helpers.

Not blocking but worth adding while you're in there: `insert_row_below`/`insert_col_end` at the border's own index (the helpers' `>=` semantics are unit-tested, the hook wiring of the insertion index for the below/end variants isn't), multi-column insert, batch `moveRows`, and undo-of-insert/redo (only undo-of-remove is covered).

**Performance:** the perf bot reports real regressions that need to be acknowledged or fixed before merge — **Filtering +23.2%** and **Initial Load (`sourceDataValidator`) +26.3%** (both scripting-time). A PR whose third milestone is performance shouldn't land with unexplained double-digit regressions on unrelated scenarios; if they're noise, please rerun and say so on the PR; if they're the viewport-sync hooks firing on filter/validation renders, that needs a look.

Suggestion: once the tests are in their final shape, run the mutation layer on the new pure helpers as the last word on unit-test strength — `node evals/score.mjs handsontable/src/plugins/customBorders/__tests__/utils.unit.js --mutate src/plugins/customBorders/utils.ts`.

handsontable/src/plugins/customBorders/__tests__/customBorders.spec.js
adrianspdev: Fixed by migration rather than by pinning inside this suite: the whole describe (with its siblings sitting near the same geometry cliff) moved to the Playwright tier in a8a1bc9c07, where the lab fixture pins the grid to 700×500, so the rendered window is deterministic on every theme. The render assertion is kept — `countVisibleCustomBorders()` still must be 1 after the multi-row insert. Verified across all three theme projects (main/horizon/classic), three consecutive runs, no flake.

handsontable/src/plugins/customBorders/__tests__/customBorders.spec.js
adrianspdev: Moved in a8a1bc9c07: the three new describes (structural changes, #6679 rendering, progressive) now live in `tests/e2e/customBorders.spec.ts`, driven through a new lab fixture (`tests/fixtures/demo/custom-borders-lab.html` exposing `createHot()`) and a `CustomBordersLabPage` page object. The Jasmine spec keeps only the pre-existing `countCustomBorders()`/virtual-rendering assertion maintenance. The legacy customBorders suite still passes (141 specs), and the Playwright spec passes 38 tests × 3 themes.

handsontable/src/plugins/customBorders/__tests__/customBorders.spec.js
adrianspdev: All three gaps covered in a8a1bc9c07 (Playwright progressive describe):

- **Structural change mid-flight**: the setup and `alter('insert_row_above', 1, 1)` run in one synchronous `page.evaluate` so the load is provably in flight (model size 0 captured before the alter). Asserts the flushed model equals the synchronous outcome (all 20 borders, shifted), `afterCustomBordersUpdate` fired exactly once, and — after a 0ms-macrotask scheduling barrier matching the chunk timers — no stale batch fires afterwards.
- **`customBordersProgressive: true`**: dedicated test asserting deferral at init and full application on drain.
- **Disable/destroy mid-flight**: two tests asserting the cancel-without-hook path — model empty, hook count stays 0 past the barrier, no errors.

handsontable/src/plugins/customBorders/__tests__/customBorders.spec.js
adrianspdev: Deduplicated in a8a1bc9c07: the geometry queries now have a single definition (`window.borderUtils` in the lab fixture — `horizontalEdgeTop`, `outerRightNear`/`outerLeftNear`, `bottomEdgeOuterRight`/`Left`, z-index lookup), wrapped as intent-level methods on the page object.

On tier: the Walkontable Jasmine suite carries the same freeze as the main one (the presence gate blocks new `*.spec.js` there; new walkontable coverage goes to Playwright per `.ai/TESTING.md`), so `walkontable/test/spec/selection/border.spec.js` couldn't take these additions. The engine behaviors are instead exercised in the Playwright tier: per-edge geometry and hide-gated extension via the #6679 describes (LTR + RTL), and the headers-only-overlay culling early-out via a dedicated working-set test asserting zero border DOM in the header-only clones.

handsontable/src/3rdparty/walkontable/src/selection/border/border.ts
adrianspdev: Added in a8a1bc9c07: a dedicated RTL describe with both #6679 scenarios under `layoutDirection: 'rtl'` — (1) mixed-width overlapping ranges: the thinner range's bottom edge stays level across the shared cell, the thicker edge stacks on top, and at the meeting corner the horizontal edge reaches but does not overshoot the perpendicular one on the inline-start (visually left) side — the mirror of the LTR assertion; (2) the split-range case: after inserting a column inside the range, the hidden-end cell's bottom edge must not stick out leftwards into the inserted column. Runs across all three themes like the rest of the spec.

handsontable/src/plugins/customBorders/utils.ts
adrianspdev: Added in aa93b6d0ac — four tests in `utils.unit.js`: explicit per-side object takes precedence (same reference, range border ignored); the `{}`/`true`/`''` inheritance forms all resolve to the range-level style; `cornerVisible` is dropped from the inherited copy without mutating the shared range-level object; and the no-`rangeBorder` passthrough (undefined/`true`/empty object returned unchanged).

tests/e2e/customBorders.spec.ts
adrianspdev: Fixed in a8a1bc9c07: the fixed 100ms wait is gone. The evaluate now returns right after `alter`, the undo runs in its own evaluate, and the restore is awaited with `expect.poll` on `getBorders()` — same pattern as the culling tests in this file.

tests/e2e/customBorders.spec.ts
adrianspdev: Added in a8a1bc9c07: `expect(result.beforeClear).toBeGreaterThanOrEqual(2)` — the test now proves the plugin owned at least one selection alongside the foreign one before the clear.

adrianspdev:
Review follow-up addressed in aa93b6d0ac + a8a1bc9c07 — details in the inline threads. Summary:

- **Blocking 1+2 (CI red, frozen suite)**: the three new describes moved to the Playwright tier (`tests/e2e/customBorders.spec.ts` + a lab fixture with pinned grid dimensions and a shared `borderUtils` geometry helper). The horizon failure is gone by construction; the suite passes 38 tests × 3 themes, three consecutive runs, no flake. The Jasmine spec keeps only the pre-existing assertion maintenance (141 specs still green).
- **Blocking 3 (M1×M3)**: mid-flight coverage added — `alter()` during an in-flight progressive load (synchronous flush, hook fires exactly once, no stale batch afterwards), the `customBordersProgressive: true` boolean form, and disable/destroy mid-flight (cancel-without-hook).
- **Blocking 4 (RTL / Walkontable tier)**: RTL variants of both #6679 geometry scenarios added. The Walkontable Jasmine suite is frozen like the main one, so the engine behaviors (per-edge geometry, hide-gated extension, headers-only-overlay culling) are exercised through Playwright instead of `border.spec.js`.
- **Smaller items**: `resolveRangeBorderSide` unit tests, `expect.poll` instead of the fixed 100ms wait, `beforeClear` asserted, geometry helpers deduplicated into one definition. The non-blocking suggestions are in too: `insert_row_below`/`insert_col_end` at the border's own index, multi-column insert, batch `moveRows`, and undo-of-insert + redo.

**Performance**: the two reported regressions look like runner noise, not real cost. Mechanically, neither scenario configures `customBorders`, so the plugin never enables and none of the new hooks (including the viewport-sync ones) are ever registered on those paths — the only new code reachable there is a couple of type comparisons in the selection manager render loop and one extra `parseInt` in the border corner path. Empirically, an A/B rerun of both scenarios with the same harness (develop CI UMD artifact vs this branch, 3 iterations each) shows parity: Filtering scripting 8 ms vs 8 ms (`beforeFilter`→`afterFilter` 44.5 ms vs 45.4 ms, within the iteration spread), `sourceDataValidator` load 11 ms vs 11 ms, JS heap slightly lower on the branch. The perf job on this PR should be rerun to confirm on CI hardware.

**Mutation layer**: ran `node evals/score.mjs .../utils.unit.js --mutate src/plugins/customBorders/utils.ts` — static verdict "meaningful" (18 tests, 35 assertions, no hollow tests, no gaming signals). The Stryker step itself currently fails in the repo (`Missing script: "env-cmd"` when invoked through the eval wrapper) — infra issue independent of this PR.

handsontable/src/plugins/customBorders/customBorders.ts
```diff
@@ -1050,10 +1416,214 @@ The border style will be ignored.`);
...
+      return;
+    }
+
+    const border = normalizeBorder(deepClone(value));
+
+    border.row = row;
+    border.col = column;
+    border.id = createId(row, column);
+
+    this.insertBorderIntoSettings(border, undefined);
```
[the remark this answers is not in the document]
adrianspdev: Confirmed and fixed in b513f4197e. The guard required the plugin-internal `id`/`row`/`col` bookkeeping (then overwrote them anyway), so user-authored writes like `setCellMeta(r, c, 'borders', { top: { width: 2 } })` never synced. `#onAfterSetCellMeta` now accepts any record and routes it through `prepareBorderFromCustomAdded` as a border descriptor for the write's coordinates — the canonical object is written back to the meta, an all-hidden value clears the cell, and the model entry is upserted, so meta and model cannot diverge. The merge base also re-derives `id`/`row`/`col` from the cell's coordinates, closing the adjacent stale-snapshot case (UndoRedo restoring meta captured at pre-shift coordinates). Covered by three new Playwright tests (partial write syncs + renders, replace semantics, all-hidden clears).

danielzytohoc (dismissed):

handsontable/src/plugins/customBorders/customBorders.ts
```diff
@@ -196,6 +263,17 @@ export class CustomBorders extends BasePlugin {
...
     this.addHook('init', () => this.#onAfterInit());
+    this.addHook('afterCreateRow', this.#onAfterCreateRow);
+    this.addHook('afterRemoveRow', this.#onAfterRemoveRow);
+    this.addHook('afterCreateCol', this.#onAfterCreateCol);
+    this.addHook('afterRemoveCol', this.#onAfterRemoveCol);
+    this.addHook('beforeRowMove', this.#onBeforeRowMove);
+    this.addHook('afterRowMove', this.#onAfterRowMove);
+    this.addHook('beforeColumnMove', this.#onBeforeColumnMove);
+    this.addHook('afterColumnMove', this.#onAfterColumnMove);
+    this.addHook('afterViewRender', this.#onAfterViewRender);
```
budnix: Can I ask — was `afterViewRender` picked over `beforeViewRender` on purpose? In the draw cycle the overlay clones draw at `drawCycle.ts:181` and `onDraw` (which fires `afterViewRender`) only runs at line 185, so a selection created inside `#syncViewportSelections` reaches the master at line 198 but misses the clones in that same draw. With `fixedColumnsStart: 2` and a border at (10, 0), scrolling row 10 out of the rendered band and back should show the border in the master but not in the frozen column until something else re-renders. `beforeViewRender` fires after `createCalculators` (line 119, so the new range is already resolved) and before the clones draw, which covers both in one pass — worth a quick check.

```suggestion
    this.addHook('beforeViewRender', this.#onAfterViewRender);
```

handsontable/src/plugins/customBorders/customBorders.ts
```diff
@@ -838,102 +1036,177 @@ export class CustomBorders extends BasePlugin {
...
+   * @param {Array} borders The border configuration entries to apply.
+   * @param {number} chunkSize The number of entries to apply per batch.
+   */
+  #startProgressiveApply(borders: CustomBorderConfig[], chunkSize: number) {
+    this.#progressiveToken += 1;
+    this.#progressiveQueue = borders;
+    this.#progressiveIndex = 0;
+    this.#progressiveChunkSize = chunkSize;
+    // One first-touch tracking Set shared across every batch (overlap-merge correctness).
+    this.#configPassTouched = new Set();
```
budnix, line 1287: This Set stays alive across every `setTimeout` gap until the queue drains, so `#readExistingBordersForMerge` treats the whole progressive load as one config pass. The JSDoc on `#configPassTouched` states the opposite — outside a config pass it should be `null` so `setBorders` keeps merging as before. If app code calls `setBorders([[3,3,3,3]], { top: { width: 2 } })` while chunks are still draining and that cell already has a `start` border, the first touch skips the existing-meta read and the `start` side is silently dropped. Scoping the Set to one synchronous batch, or using a separate "load in flight" flag, would keep the public API behaving the same.

budnix (commented):
Went through the virtualization and progressive-apply paths. The overall shape looks right — the viewport union ranges, the shift math on insert/remove, and the flush-then-shift ordering in `#shiftBorders` all check out. Five things below that look worth a second pass, mostly around when the sync runs and what the progressive load does to the merge rules.

handsontable/src/plugins/customBorders/customBorders.ts
```diff
@@ -838,102 +1036,177 @@ export class CustomBorders extends BasePlugin {
...
+      return;
     }
 
-    return check;
+    const queue = this.#progressiveQueue;
+    const end = Math.min(this.#progressiveIndex + this.#progressiveChunkSize, queue.length);
+
+    this.createCustomBorders(queue.slice(this.#progressiveIndex, end));
+    this.#progressiveIndex = end;
+    this.hot.render();
```
budnix, line 1318: Each chunk sets `#bordersByRowDirty`, and this render runs `#syncViewportSelections` → `#rebuildBordersByRow()`, which walks all of `savedBorders` rather than just the new chunk. With the default chunk size of 5000 and a 1M-border config that's 200 full rebuilds over a growing array, so the cost grows with the square of the border count — in the exact path this feature exists to make fast. Appending the new chunk into `#bordersByRow` instead of rebuilding would keep it linear. Curious whether you measured this on a really large config.

handsontable/src/plugins/customBorders/customBorders.ts
```diff
@@ -459,49 +535,112 @@ export class CustomBorders extends BasePlugin {
...
+
+  /**
+   * Writes or removes the plugin-owned `borders` cell meta with the re-entrancy guard raised, so
+   * the external-write listeners ignore it. Pass `null` to remove the meta.
+   *
+   * @param {number} row Visual row index.
+   * @param {number} column Visual column index.
+   * @param {object|null} value The denormalized border object, or `null` to remove.
+   */
+  #writeBordersMeta(row: number, column: number, value: Record<string, unknown> | null) {
```
budnix: `hot.setCellMeta` returns without writing when a `beforeSetCellMeta` listener returns `false`, and this method doesn't check for that. Callers still run `insertBorderIntoSettings` afterwards, so the border lands in `savedBorders` and renders with no cell meta behind it — `getBorders()` and `getCellMeta().borders` then disagree, and `#resetBorderModel` later tries to remove a meta key that was never written. An app that blocks border writes on locked cells would hit this straight from the context menu. Returning the write result so callers can skip the model insert on `false` would keep the two in step.

handsontable/src/plugins/customBorders/customBorders.ts
```diff
@@ -1044,16 +1424,234 @@ The border style will be ignored.`);
...
+   * UndoRedo restoring the meta of an undone row/column removal, or user code calling
+   * `setCellMeta` directly) by upserting the matching entry in the border model, so the model and
+   * the meta cannot diverge. The plugin's own writes are excluded by the re-entrancy guard.
+   *
+   * @param {number} row Visual row index.
+   * @param {number} column Visual column index.
+   * @param {string} key The meta key that was written.
+   * @param {*} value The written value.
+   */
+  #onAfterSetCellMeta = (row: number, column: number, key: string, value: unknown) => {
```
budnix, line 1644: `prepareBorderFromCustomAdded` ends up in `insertBorderIntoSettings`, which calls `#destroyBorderSelection` so the next sync rebuilds the selection. But `setCellMeta` doesn't render, and this handler doesn't call `hot.render()` either, so after `hot.setCellMeta(0, 0, 'borders', { top: { width: 2 } })` the old border DOM is gone and the new one waits for some unrelated render. Did this end up relying on the caller to render, or should the handler render itself?

handsontable/src/plugins/customBorders/customBorders.ts
```diff
@@ -990,6 +1263,113 @@ The border style will be ignored.`);
...
+   * otherwise the border moves to the new coordinate. The cell meta is left untouched - Handsontable
+   * already shifts it, and this method only realigns the plugin's own bookkeeping to that meta.
+   *
+   * @param {string} axis The coordinate axis to shift - `'row'` or `'col'`.
+   * @param {Function} mapIndex Maps an old visual index to its new index, or `-1` when removed.
+   */
+  #shiftBorders(axis: 'row' | 'col', mapIndex: (index: number) => number) {
+    // A structural change remaps coordinates; the model must be complete first, so finish any
+    // in-flight progressive load synchronously before shifting.
+    this.#flushProgressiveApply();
```
demtario: **Critical: this flush runs after the core has already shifted the cell meta, so the flushed entries write `borders` meta to the wrong cells.**

`#shiftBorders` is called from the `afterCreateRow`/`afterRemoveRow`/`afterCreateCol`/`afterRemoveCol` callbacks, and by that time `DataMap#createRow` has already run `metaManager.createRow` (the meta shift precedes the hook — see `dataMap.ts`, the hook fires at the end of `createRow`). The queued progressive entries still carry pre-change visual coordinates, so the flush's `#writeBordersMeta` lands on post-shift cells that no longer correspond to the config's targets; `#shiftBorders` then remaps only the model.

Concrete case: `customBorders: [{ row: 5, col: 0, top: … }]` still queued, `alter('insert_row_above', 1, 1)` arrives. The flush writes meta at visual row 5 (wrong cell — the intended data row now sits at 6), then the model entry is shifted to 6. Result: orphaned `borders` meta on row 5, a rendered border at row 6 with no meta behind it — `getCellMeta().borders` and `getBorders()` permanently disagree.

The move handlers already do this correctly: `#onBeforeRowMove`/`#onBeforeColumnMove` flush in the *before* hooks, while coordinates are still pre-change. The insert/remove path needs the same treatment — flush from `beforeCreateRow`/`beforeRemoveRow`/`beforeCreateCol`/`beforeRemoveCol` instead of inside `#shiftBorders`.

The flush test (`tests/e2e/customBorders.spec.ts`, "flushes an in-flight progressive load synchronously…") passes despite this because it asserts the model only (`getBorders()`), never the meta. Please extend it to pin meta/model consistency — e.g. after the insert, `cellBorders(1, 0)` is `null` and `cellBorders(2, 0)` is green.

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/fixtures/demo/custom-borders-lab.html
```diff
@@ -0,0 +1,172 @@
...
+  <script>
+    window.htTheme = ({ horizon: 'horizon', classic: 'classic' })[new URLSearchParams(location.search).get('theme')] || 'main';
+    document.write('<link rel="stylesheet" href="/handsontable/styles/ht-theme-' + window.htTheme + '.min.css">');
+  </script>
+  <style> body { font-family: sans-serif; margin: 1rem; } </style>
+</head>
+<body>
+  <div id="grid" data-testid="grid"></div>
+
+  <script src="/handsontable/dist/handsontable.full.min.js"></script>
```
demtario: **High: both new fixtures hardcode the full-min bundle and ignore the `?bundle=` axis — after merging develop, three of the six matrix legs run silent duplicates.**

develop now carries DEV-2190 (#13163): `tests/playwright.config.ts` has six theme × bundle projects and `tests/fixtures/test.ts` passes `?bundle=umd|full-min` to every fixture; `grid.html` selects the script from an allowlist and throws on unknown values. This branch predates that, so nothing fails today — but on merge this spec runs 6× with the base UMD bundle never actually loaded, which is exactly the silently-mislabeled-green-leg failure mode the `grid.html` allowlist comment warns about.

CustomBorders has no HyperFormula dependency and `Handsontable.helper.createSpreadsheetData` exists in both dist bundles, so there is no justified exception. Fix in one pass: copy `grid.html`'s bundle allowlist + `document.write` into this fixture and `custom-borders.html`, add a `bundle` param to `CustomBordersLabPage.goto()` (mirroring `GridPage`), forward it through `gotoLab()`, and append `&bundle=` to the three raw `page.goto` calls in the spec.

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/fixtures/demo/custom-borders-lab.html
```diff
@@ -0,0 +1,172 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <title>CustomBorders lab e2e fixture</title>
+  <link rel="stylesheet" href="/handsontable/styles/handsontable.min.css">
+  <script>
+    window.htTheme = ({ horizon: 'horizon', classic: 'classic' })[new URLSearchParams(location.search).get('theme')] || 'main';
```
demtario: **High: unknown `?theme=` values silently fall back to `main` instead of throwing.**

`({ horizon, classic })[param] || 'main'` (here and in `custom-borders.html:8`) turns a config typo into a mislabeled green leg — the `horizon` project could quietly run `main` and nobody would know. develop's `grid.html` deliberately validates against an allowlist and throws on anything unexpected so the mistake surfaces as one red leg. Please adopt the same throw-on-unknown pattern in both fixtures (naturally folds into the `?bundle=` fix from the other comment).

---
_Generated by [Claude Code](https://claude.ai/code)_

handsontable/src/plugins/customBorders/customBorders.ts
```diff
@@ -838,102 +1036,177 @@ export class CustomBorders extends BasePlugin {
...
-   *                        `border` ({Object} with `color`, `width` and `cornerVisible` property) properties.
-   * @param {CellRange} cellRange The selection range to check.
-   * @param {string} [place] Coordinate where add/remove border - `top`, `bottom`, `start`, `end`.
-   * @returns {boolean}
+   * @returns {{ enabled: boolean, chunkSize: number }}
    */
-  checkCustomSelections(border: BorderObject, cellRange: CellRange, place: string | undefined) {
-    const hideCount = this.countHide(border);
-    let check = false;
+  #resolveProgressiveSetting(): { enabled: boolean, chunkSize: number } {
```
demtario, line 1256: **Medium: `customBordersProgressive` is read here but is missing from the plugin's `SETTING_KEYS`.**

`CustomBorders` inherits the default `SETTING_KEYS` (`[PLUGIN_KEY]` — `base.ts:73`), so `updateSettings({ customBordersProgressive: … })` without a `customBorders` key never triggers `updatePlugin()` — the new option is silently inert until the next unrelated `customBorders` update. Override `SETTING_KEYS` to `[PLUGIN_KEY, 'customBordersProgressive']`.

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/fixtures/pages/CustomBordersLabPage.ts
```diff
@@ -0,0 +1,183 @@
...
+
+    return modelSizeAtInit;
+  }
+
+  /** A data cell in the master overlay (frozen clones duplicate test ids). */
+  cellInMaster(row: number, col: number): Locator {
+    return this.page.locator('.ht_master').getByTestId(`cell-${row}-${col}`);
+  }
+
+  /** Run `alter` and wait for the follow-up render to settle into the DOM. */
```
demtario: **Medium: the JSDoc promises a wait the method doesn't perform.**

"…and wait for the follow-up render to settle into the DOM" — but the body is a bare `page.evaluate` that returns as soon as `alter()` does. Several tests take non-retrying DOM reads immediately after calling this (`customBorders.spec.ts:162`, `:179`, `:196`, `:211`, `:243`, `:275`, `:473`, `:493`), which works only as long as the follow-up render happens to be synchronous. Either make `alter()` await an actual condition, or fix the JSDoc and convert those reads to `expect.poll` so the contract and the callers agree.

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/e2e/customBorders.spec.ts
```diff
@@ -0,0 +1,1086 @@
...
+/**
+ * Functional coverage for the CustomBorders viewport working set. The border
+ * DOM (`.wtBorder` divs) is created per overlay by the selection manager, so
+ * the assertions target the overlay containers directly. Visible border
+ * edges carry inline sizes; hidden ones are `display: none`, which the
+ * `:visible` filter excludes.
+ */
+test.describe('CustomBorders with frozen rows and columns', () => {
+  test.beforeEach(async ({ page, theme }) => {
+    await page.goto(`/tests/fixtures/demo/custom-borders.html?theme=${theme}`);
```
demtario: **Medium: the three describes driven by `custom-borders.html` bypass the page-object model.**

They navigate with raw `page.goto` (duplicated at `:31`, `:89`, `:117`), assert against inline structural selectors (`.ht_clone_top .wtBorder:visible`, `.ht_clone_inline_start …`), and reach into `hot.selection.highlight.customSelections` from spec code — the Playwright skill's rule 1 keeps selectors and multi-step flows inside page objects. `CustomBordersLabPage` already wraps the same kind of queries for the lab fixture; `custom-borders.html` deserves the same treatment (or these describes could move onto the lab fixture with `fixedRowsTop`/`fixedColumnsStart` passed through `createGrid`).

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/e2e/customBorders.spec.ts
```diff
@@ -0,0 +1,1086 @@
...
+      customBordersProgressive: { chunkSize: 5 },
+    });
+
+    // Right after init the grid is rendered but the borders have NOT been applied yet - they
+    // are queued for background batches, so init did not block on building them.
+    expect(modelSizeAtInit).toBe(0);
+
+    // Once the queue drains, the completion hook fires exactly once and the full configuration
+    // is applied - identical to the synchronous path.
+    await expect.poll(() => lab.bordersUpdateCount()).toBe(1);
```
demtario, line 913: **Medium: the comment claims "fires exactly once", but `expect.poll(...).toBe(1)` proves only "reached 1".**

The poll resolves the instant the counter hits 1 — a stray second fire a macrotask later goes unnoticed. The flush test in this same describe already solves this: re-assert the count after `macrotaskBarrier(page)` (`:998–999`). Same two lines here and in the boolean-`true` variant (`:921`) would make both tests actually verify their stated claim.

---
_Generated by [Claude Code](https://claude.ai/code)_

demtario (commented):
Review of the current head (`145f69c`). Seven findings not covered by the existing threads — one critical (progressive flush ordering vs. cell meta), two on the new fixtures' matrix behavior, four medium. Inline below.

handsontable/src/plugins/customBorders/customBorders.ts
adrianspdev: Good catch — moved to `beforeViewRender` in 29395c5.

Confirmed the draw order you described: `createCalculators` (drawCycle.ts:119) → `beforeViewRender` (149) → clones draw (181) → `afterViewRender` (185) → master selections (198). A sync at 185 does land between the clones and the master.

Reproduced it before fixing: `fixedColumnsStart: 2`, border at (10, 0), scroll row 10 out of the rendered band and back — the frozen column stayed blank while the master had its border. `getFirstRenderedRow()` reads `rowsRenderCalculator`, which `createCalculators` sets at 119, so the earlier hook sees the correct new range.

Regression test: "redraws a frozen-column border that scrolls back into the rendered range". It gates on the clone re-rendering the bordered cell, so it asserts the border is present in that same draw rather than polling until some later render supplies it.

handsontable/src/plugins/customBorders/customBorders.ts
adrianspdev: Correct, and it went further than the write path. Fixed in 29395c5.

`#writeBordersMeta` now verifies the write landed and returns that, and all five sites gate on it — the two add paths, the context-menu toggle, `removeAllBorders`, and `#resetBorderModel`. The removal sites needed the reverse treatment: a vetoed `beforeRemoveCellMeta` leaves the meta in place, so dropping the model entry anyway would have desynced them in the other direction. Those cells now keep their model entry.

Two tests, both red without the change: a vetoed write leaves `getBorders()` empty with no meta behind it, and a vetoed removal keeps both.

handsontable/src/plugins/customBorders/customBorders.ts
adrianspdev: Neither — it was a bug, and the existing tests were hiding it. Fixed in 29395c5.

Measured before fixing: with a border on (0, 0), an external `setCellMeta(0, 0, 'borders', …)` took the visible border count from 1 to 0. `insertBorderIntoSettings` destroys the cell's selection so the viewport sync can rebuild it, and `setCellMeta` never renders, so the cell was simply left blank.

The reason it went unnoticed: every existing test in that describe calls `hot.render()` right after the write. The new one deliberately does not.

handsontable/src/plugins/customBorders/customBorders.ts
adrianspdev: I added the guard, but I could not reproduce the loss you describe, so I want to be straight about what changed rather than claim a fix.

`createCustomBorders` now sets an `#inConfigPass` flag and the first-touch skip is gated on it, so the Set survives across batches (overlapping ranges spanning a batch boundary still merge) while anything running in the gaps between them always reads. That makes the field's JSDoc true.

What I could not construct is a case where the old code actually loses data. The skip only fires for a cell the pass has not touched yet, and every cell the pass writes goes through `#readExistingBordersForMerge`, so it is in the Set. `#resetBorderModel` clears the previous borders' meta before the load starts, so an untouched cell has no plugin-written meta to lose. `setBorders` on a cell an earlier chunk already wrote finds the key present and merges normally — I tested that path and it passes with and without the change.

The one thing that does reach an untouched cell with meta on it is a *cascaded* `borders` value (`columns: [{ borders: … }]`), and that is dropped identically with the guard, without it, and on `develop` — for an unrelated reason: `isBorderObject` requires the internal `id`/`row`/`col`, so a user-authored shape is never a merge base. Filed separately as DEV-2513, since it is not something this PR introduced.

So: hardening, and it removes the hazard the moment any path does leave plugin-shaped meta on an untouched cell. If you have a reachable sequence I missed, I would rather have the test.

handsontable/src/plugins/customBorders/customBorders.ts
adrianspdev: You are right about the shape, and no, it had not been measured. It has now.

Fixed in 29395c5: the row index is patched per border (`#indexBorderByRow` / `#unindexBorderByRow`) instead of being invalidated wholesale, so a batch no longer walks the whole of `savedBorders`. Ids are coordinate-based, so a border's bucket is fixed unless `#shiftBorders` rewrites the ids — which still invalidates the index outright.

A/B on the same machine, unique coordinates, `chunkSize: 5000`, timing the drain to `afterCustomBordersUpdate`:

| borders | wholesale invalidate | incremental |
|---|---|---|
| 160k | 4066 ms | 4088 ms |
| 240k | 6107 ms | 6057 ms |

Below noise. The per-border `setCellMeta` dominates at roughly 45 µs each, so at your 1M example the rebuild term is on the order of 100M iterations (~1 s) against a ~45 s drain — a few percent, not the scaling wall. Kept the change because the term is real and now gone, but I would not describe it as what makes that path slow.

handsontable/src/plugins/customBorders/customBorders.ts
adrianspdev: Confirmed and fixed in 29395c5 — this was real, and your reading of why the test missed it was exactly right.

Reproduced first: config `{ row: 5, col: 0, top: green }` still queued, `alter('insert_row_above', 1, 1)` in the same synchronous task. The meta landed on visual row 5, and that object internally read `id: "border_row6col0", row: 6` — `#shiftBorders` had mutated the same object after the write. Meta at row 6 was `undefined`, model said row 6.

The flush now runs from `beforeCreateRow` / `beforeRemoveRow` / `beforeCreateCol` / `beforeRemoveCol` and is gone from `#shiftBorders`, matching what the move handlers already did. One wrinkle worth naming: `source === 'auto'` skips the flush, so `minSpareRows`/`minSpareCols` do not collapse a progressive load into a synchronous one — those rows append at the end and shift nothing.

The existing flush test now pins meta as well as the model, as you asked — `cellBorders(0, 0)` green, `cellBorders(1, 0)` null (the inserted row), `cellBorders(2, 0)` and `cellBorders(20, 0)` green. Plus a new removal-side test.

Verified the extension actually catches it: reverting only the flush placement and rebuilding turns it red with the orphaned `border_row2col0` meta sitting on row 1.

handsontable/src/plugins/customBorders/customBorders.ts
adrianspdev: Confirmed and fixed in 29395c5 — `SETTING_KEYS` now returns `[PLUGIN_KEY, 'customBordersProgressive']`.

Reproduced first: with 20 borders applied, `updateSettings({ customBordersProgressive: { chunkSize: 5 } })` left `getBorders().length` at 20 — `updatePlugin()` never ran. With the override it drops to 0 and refills in batches, which is the observable signal the new option actually took effect.

Test: "applies a customBordersProgressive-only settings change".

tests/e2e/customBorders.spec.ts
adrianspdev: Fair — the comment was making a claim the assertion did not support. Fixed in 29395c5.

Both tests now re-read the counter after `macrotaskBarrier(page)`, the same way the flush test in that describe already did, so a second fire a macrotask later would be caught.

tests/fixtures/demo/custom-borders-lab.html
adrianspdev: Confirmed and fixed in 71fdb52. Merged `develop` into the branch first so this was written against the real API rather than guessed at.

Both fixtures now carry `grid.html`'s param block verbatim — allowlists to literals, absent param keeps the default (main / plain UMD), unknown value throws — and the bundle arrives through `document.write` with the split closing tag instead of a hardcoded `<script src>`. `bundle` is threaded fixture → page-object constructor `(page, theme = 'main', bundle = 'umd')` → `goto()` query params → spec destructure, mirroring `GridPage`.

Since nothing asserts which file actually loaded, I checked rather than assumed: a throwaway spec across all six legs asserting `window.htBundle` / `window.htTheme` resolve per leg, plus that an unknown `?bundle=` and `?theme=` throw. 24 passed, then deleted. Without that, this could have been plumbed but inert — the exact failure you were pointing at.

Worth noting the `e2e-main` leg now genuinely runs the base UMD bundle for these specs, which it never had. It passes.

tests/fixtures/demo/custom-borders-lab.html
adrianspdev: Confirmed and fixed in 71fdb52 — both fixtures throw on an unknown `?theme=` (and `?bundle=`), folded into the same pass as the bundle work.

One extra detail from doing it: `main` was not even a key in the old map, so `?theme=main` was reaching `main` through the `||` fallthrough rather than by being allowed. It is an explicit allowlist entry now, so the default and the valid value take the same path.

tests/fixtures/pages/CustomBordersLabPage.ts
adrianspdev: The doc was wrong, so I fixed the doc rather than the callers — flagging that since you offered both options.

`alter()` ends with a synchronous `view.render()` before it returns, so by the time the `page.evaluate` resolves the DOM is settled and those 8 immediate reads are sound. Converting them to `expect.poll` would trade 8 exact assertions for 8 retrying ones to fix a comment. The JSDoc now says it renders synchronously and callers may read directly.

Happy to switch if you would rather have the retry semantics as a guard against `alter()` becoming async later.

tests/e2e/customBorders.spec.ts
adrianspdev: Agreed — fixed in 71fdb52 with a new `CustomBordersDemoPage` for the `custom-borders.html` fixture.

The three describes lost their raw `page.goto` (now `goto()` on the page object, carrying theme and bundle), the inline overlay selectors (`borderIn('corner' | 'inlineStart' | 'top' | 'master')`) and the `hot.selection.highlight` reach-ins (`hasRenderedBorder(row, col)`, `customSelectionCount()`, `addForeignCustomSelection()`). The per-test `beforeEach` navigation collapsed into a `gotoDemo()` helper matching the existing `gotoLab()`.

Kept the fixture rather than folding these onto the lab: the frozen-area specs want a fixed grid with frozen rows and columns, and moving them would have meant re-deriving the scroll distances that make the assertions theme-robust.

adrianspdev:
## Review round addressed — 12 findings, plus a regression this branch had introduced

Pushed as 71fdb52 (test infrastructure) and 29395c5 (source fixes), on top of a `develop` merge (20 commits, conflict-free) that was needed so the fixture work could be written against the real theme × bundle API rather than guessed at.

Replies are on each thread. Summary:

**@budnix** — B1 (`beforeViewRender`), B4 (vetoed meta writes), B5 (render after an external `setCellMeta`) fixed. B3 (per-chunk row-index rebuild) fixed, and measured: A/B at 160k and 240k borders shows the difference below noise, so the term was real but not what makes that path slow — numbers in the thread. B2 (`#configPassTouched` across drain gaps) I could not reproduce; the guard went in anyway to make the field's JSDoc true, and I said so plainly rather than claiming a fix.

**@demtario** — A1 (progressive flush ordering) confirmed and fixed; it was corrupting meta exactly as described, and the flush test now pins meta on four cells as you asked. A2/A3 (bundle axis, throw-on-unknown), A4 (`SETTING_KEYS`), A6 (page object for `custom-borders.html`), A7 (`macrotaskBarrier` re-assert) all fixed. A5 I fixed the JSDoc instead of converting the 8 callers — reasoning in the thread, happy to switch.

### Also fixed: a regression this branch had introduced

Separately from the reviews, the branch was breaking the `columns-add-remove-two-tables-demo` visual test. `#rebuildWorkingSetFromModel()` called `hot.render()` from inside `afterCreateCol`, which left the column-header DOM stale — labels bound to their pre-insert columns, so a header click selected a different column than the label sat on. Bisected by build against the merge-base to confirm it was ours. The screenshot is byte-identical to the pre-PR baseline again, with two regression tests.

### Verification

Every fix was proven by reverting just that change, rebuilding, and watching the matching test go red — including A1, where the reverted build reproduced the reported meta corruption exactly.

| Gate | Result |
|---|---|
| Playwright (six legs) | 564 pass |
| Legacy E2E, base bundle | 9606 specs, 0 failures |
| Legacy E2E, `full.min` | 9606 specs, 0 failures |
| Unit | 3285 pass |
| Walkontable | 816 specs, 0 failures |
| Visual cross-browser (chromium/firefox/webkit) | 116 pass |
| Target screenshot vs pre-PR baseline | byte-identical |
| eslint, stylelint, `test:types` | clean |

### Two findings split out — both pre-existing, neither caused by this PR

- **DEV-2513** — `setBorders()` drops user-authored or cascaded `borders` cell meta. Surfaced while probing B2. Reproduces identically with the plugin reverted to `develop`, by a different mechanism. Needs a call on whether `borders` cell meta is supported input before it can be fixed properly.
- **DEV-2514** — a render triggered from `afterCreateCol` corrupts the column-header DOM. Reproduces with all of `handsontable/src` reverted to `develop` and no plugin involved. This PR no longer triggers it, but it stays latent for any other hook that renders from there.

No changelog entry added — `.changelogs/13166.json` covers this PR and none of the above ever shipped.

demtario (commented):
Review of the current head (`fa47578`), scoped to the delta since `145f69c` — `71fdb52`, `29395c5`, `fa47578`.

The previous round is genuinely closed on my side: all seven of my findings and four of @budnix's five are fixed in the code, and B2 is documented honestly as unreproduced rather than claimed. The `beforeViewRender` move, the `before*` flush ordering, and the per-border row-index patch all read correctly against the draw cycle and the `alter()` flow.

Three new findings inline, all in the veto/flush work from `29395c5`. Low-severity items (changelog `issuesOrigin`, page-object symmetry in the lab describes, the `beforeRemoveCellMeta` type signature) are omitted here — happy to list them separately if you want them in this round.

**Performance: consider it closed, and the variance is the evidence.** Across runs of this same branch the sticky comment has reported Initial Load (`sourceDataValidator`) at +26.3% and now −35.4%, Filtering at +23.2% and now +50.7%, with Sorting newly at +29.5%. Those are 10–18 ms swings on 16–79 ms scenarios that never configure `customBorders`, so the plugin never enables. Your mechanical argument plus the local A/B holds; the runner is just too noisy at that scale. Worth a line on the PR saying so, so the three red flags in the thread don't read as unanswered.

Also verified clean: no `toSorted`/`at(`/`Object.hasOwn`/`structuredClone` in the added `handsontable/src` lines (ESLint can't see prototype methods on non-literal receivers, so I grepped rather than trusted the green lint), JSDoc block form throughout, no default setting value changed, `@since 18.1.0` consistent with the other 18.1.0 additions on develop, and the fixture param blocks match `demo/grid.html` 1:1. `#scheduleMetaSyncRender` registers through `_registerMicrotask`, which `destroy()` cancels via `_clearMicrotasks` (`core.ts:5824`), so there is no post-destroy render — and its test asserts `renderCount === 1`, which would read 10 without the fix.

handsontable/src/plugins/customBorders/customBorders.ts
```diff
@@ -459,47 +581,121 @@ export class CustomBorders extends BasePlugin {
...
+      } else {
+        this.hot.setCellMeta(row, column, 'borders', value);
+      }
+    } finally {
+      this.#isInternalMetaWrite = false;
+    }
+
+    const written = this.hot.getCellMeta<BordersCellProperties>(row, column).borders;
+
+    return value === null ? written === undefined : written === value;
```
demtario: This reads the resolved meta value, which walks the prototype chain, so a `borders` key that cascades from the grid or column level makes every removal look vetoed.

`removeCellMeta` is `delete cellMeta[key]` (`metaLayers/cellMeta.ts:265`), and a `CellMeta` instance has `ColumnMeta` → `TableMeta` on its prototype chain (the diagram at the top of `cellMeta.ts`). Deleting the own property therefore uncovers the inherited one, `written === undefined` is `false`, and `#writeBordersMeta` reports a veto that never happened.

I checked whether that is actually reachable, since `borders` is not declared in `metaSchema.ts` — it is. Neither meta layer filters against the schema: `TableMeta#updateMeta` and `ColumnMeta#updateMeta` both call `extend(meta, settings)` (`tableMeta.ts:61`, `columnMeta.ts:72`), and `extend` copies every own key except `__proto__`/`constructor`/`prototype` (`helpers/object.ts:75-82`). So `new Handsontable(el, { borders: {...} })` or `columns: [{ borders: {...} }]` puts the key on the chain.

With that in place, every site gated on this return value keeps its model entry:

- `removeAllBorders()` parks the border in `#resetBorderModel`'s `kept` array,
- the context-menu **Remove border** leaves the model entry behind,
- `#resetBorderModel` runs on every `createCustomBorders`, so an `updateSettings({ customBorders: [...] })` carries the previous configuration forward indefinitely.

Same root, opposite direction, in `#onAfterRemoveCellMeta`: an external `removeCellMeta(r, c, 'borders')` drops the model entry while the meta still resolves to the cascaded object — exactly the divergence this round set out to close.

Testing own-property presence rather than resolved value would fix both, e.g. read the meta object once and check `Object.prototype.hasOwnProperty.call(meta, 'borders')` for the removal case (`Object.hasOwn` is banned by `handsontable/.eslintrc.js` on the Safari 15.4 floor).

Repro: configure a grid-level `borders`, add a custom border, call `removeAllBorders()`, then read `getBorders().length`. Nothing in the suite covers a cascaded `borders` value today.

This is adjacent to the DEV-2513 you split out, but the keep-on-veto behavior is new in `29395c5`, so I think it belongs to this PR.

handsontable/src/plugins/customBorders/customBorders.ts
```diff
@@ -459,47 +581,121 @@ export class CustomBorders extends BasePlugin {
...
+      if (value === null) {
+        this.hot.removeCellMeta(row, column, 'borders');
+      } else {
+        this.hot.setCellMeta(row, column, 'borders', value);
+      }
+    } finally {
+      this.#isInternalMetaWrite = false;
+    }
+
+    const written = this.hot.getCellMeta<BordersCellProperties>(row, column).borders;
```
demtario: This is one `getCellMeta()` per border on the config-application path — the same read the first-touch skip was added to remove.

`#writeBordersMeta` runs once per border inside `createCustomBorders`, and `getCellMeta` is not a field read: it runs the full dynamic extension, the `cells` function plus `beforeGetCellMeta`/`afterGetCellMeta`, on every call (`core.ts:4469` → `metaManager.getCellMeta`). The PR summary credits the first-touch read-skip with roughly 27% faster init on the stress case; this puts an equivalent read back, unconditionally, on the same path.

A veto is only possible when something can veto, so the verification can be skipped when nothing is listening:

```js
if (!this.hot.hasHook(value === null ? 'beforeRemoveCellMeta' : 'beforeSetCellMeta')) {
  return true;
}
```

`hasHook` is on the core (`core.ts:6180`) and the codebase already uses it for exactly this kind of fast path at `core.ts:4093`. I checked the guard would not be dead weight: nothing in `handsontable/src` registers a `beforeSetCellMeta` or `beforeRemoveCellMeta` listener, so in a grid without app-supplied listeners the read disappears entirely.

Worth re-running the 240k A/B from the B3 thread with and without it — that harness is already built, and it would put a number on whether the init claim in the description still holds at head.

handsontable/src/plugins/customBorders/customBorders.ts
```diff
@@ -1044,16 +1580,337 @@ The border style will be ignored.`);
...
+   * `after*` hook, so a queue drained from there would write its `borders` meta onto cells the
+   * configuration never targeted, leaving the meta and the model permanently out of step. The
+   * row/column move handlers flush from their `before*` hooks for the same reason.
+   *
+   * Auto-inserted rows and columns (`minSpareRows` / `minSpareCols`) append at the end and shift
+   * nothing, so they keep the load progressive instead of forcing it to complete.
+   *
+   * @param {string} [source] Source that triggered the structural change.
+   */
+  #flushBeforeStructuralChange(source?: string) {
```
demtario, line 1731: Moving the flush to the `before*` hooks is right for the meta-ordering bug, but these hooks are vetoable and the flush runs before the veto is known.

`beforeCreateRow`, `beforeRemoveRow`, `beforeCreateCol` and `beforeRemoveCol` all cancel the operation when a listener returns `false`. A confirmation-dialog pattern — veto now, re-issue after the user confirms — is common, and each blocked attempt now drains the entire remaining configuration synchronously. On the 240k-border case from the B3 thread that is a multi-second main-thread stall, in the feature whose purpose is to avoid one. A listener that vetoes on a condition that stays false turns the progressive load into a synchronous one on the first structural attempt.

The ordering itself can't move back — that was the meta-corruption finding from the last round. Either re-arm the queue when the matching `after*` hook does not fire, or accept it and say so here: the JSDoc currently justifies only the `'auto'` exemption, so the veto interaction reads as unconsidered rather than as a deliberate trade.

No coverage either way — the mid-flight tests all use structural changes that go through.

handsontable/src/plugins/customBorders/customBorders.ts
adrianspdev: Confirmed — every link in the chain checks out (`delete` of the own key, unfiltered `extend` in both meta layers). Fixed in f468b552e3 + 2a4fe648b3: `#writeBordersMeta` now verifies the removal against own-property presence (`hasOwnProperty` helper; `Object.hasOwn` is banned on the Safari floor) instead of the resolved value, and the set direction keeps the identity check. Regression tests cover your repro (grid-level cascade + `clearBorders()`), the `updateSettings` carry-forward through a column-level cascade, and — with a benign non-vetoing `beforeRemoveCellMeta` listener — the own-property branch itself, so reverting it to a resolved-value check fails the suite.

On the `#onAfterRemoveCellMeta` half: the own key really was removed and the model entry with it — the value that still resolves afterwards is the cascaded one, which was never plugin-owned and never in the model. That residual divergence is inherent to cascaded `borders` values; the JSDoc now states they are unsupported rather than the plugin pretending to manage them.

handsontable/src/plugins/customBorders/customBorders.ts
adrianspdev: Added in f468b552e3 — the verification read now sits behind `hasHook('beforeRemoveCellMeta'/'beforeSetCellMeta')` (instance + global buckets), so in a grid with no app-supplied `before*` meta listeners the read disappears entirely. One subtlety your suggestion's shape would have missed: the guard is sampled *before* the write (2a4fe648b3), because a `runOnce` veto listener deletes itself the moment it fires — a post-write probe reports "no hook" for a write that was in fact vetoed. A regression test pins that case. When the verification does run it uses `getCellMetaTransient` as a drop-in for `getCellMeta`.

On the A/B ask: at head the argument holds without a benchmark. With no `before*` meta listener registered (nothing in `handsontable/src` registers one, and app listeners land in exactly the buckets `hasHook` checks), `#writeBordersMeta` performs zero meta reads — the config-application path does strictly less work than the version you reviewed. The read exists only when a veto is actually possible.

handsontable/src/plugins/customBorders/customBorders.ts
adrianspdev: Taking your second option: documented as a deliberate trade-off in `#flushBeforeStructuralChange`'s JSDoc plus a characterization test (vetoed `remove_row` mid-flight → queue drained synchronously, model complete and correct at the original coordinates, completion hook fires once) in 93cd1b392a. Re-arming is not possible once the queue applied, and the veto outcome is not knowable from a `before*` hook — a later listener may still cancel. The real alternative (translating queued coordinates through the structural change instead of flushing) is a larger change; happy to file it as a follow-up if you think the veto+huge-config overlap warrants i

demtario (approved):

budnix (approved):

PR #13170, opened 2026-08-07
DEV-1161: Add preserveNumericLiteral option to keep numeric editor input

sl01k:
### Context

The PR fixes a numeric cell losing the user's exact input when it is parsed to a JavaScript number on save. Two symptoms were reported: editing `9.0` shows `9` in the cell editor (the trailing `.0` is dropped), and large numbers lose precision on paste/edit (e.g. `12345678901234567.8` becomes a rounded integer). Both come from the numeric `valueSetter`, which eagerly runs `parseFloat` and discards the original text.

Because JavaScript stores all numbers as IEEE-754 doubles, `9.0 === 9` and values above `Number.MAX_SAFE_INTEGER` cannot be represented exactly — so the literal can only survive if it is kept as text.

This is implemented as an **opt-in** option, `preserveNumericLiteral` (default `false`), so existing configurations are completely unchanged. When enabled, and only when converting the input to a number would lose information (trailing fractional zeros, or magnitude beyond the safe-integer limit), the cell keeps the original trimmed literal string; otherwise it stores the parsed number exactly as before. The guard is limited to plain dot-decimal literals (`isNumeric`), so grouped values (`7.000`) and comma-decimal literals (`9,0`) are unaffected. Values that convert without loss (`9`, `9.5`, `1000`) are always stored as numbers, so sorting, filtering, and formulas are unaffected.

Scope is editor-focused; there are no `numericRenderer` changes. As a verified bonus, because `Intl.NumberFormat.format()` accepts a decimal string, a preserved large-number literal also renders at full precision in the cell.

### How has this been tested?

New and updated automated tests, plus manual verification in a real browser (both grids of an ON/OFF demo, plus sorting and filtering).

- Unit — number helper and the setter:
  - `npm run test:unit --prefix handsontable --testPathPattern=number.unit`
  - `npm run test:unit --prefix handsontable --testPathPattern=numericType.unit`
  - Covers: lossless values stay numbers; `9.0`, `0.0`, `-9.0`, `12345678901234567.8` are preserved only with the option on; default (option off) keeps the old parse-to-number behavior; comma-decimal and grouped inputs are never preserved; idempotence.
- Type test — `handsontable/src/__tests__/core/settings.types.ts` updated for the new option; `npm run test:types --prefix handsontable`.
- E2E (Playwright, new) — `tests/e2e/numeric-editor.spec.ts` with fixture `numeric.html` + `NumericGridPage`:
  - `cd tests && npx playwright test --project=e2e-main e2e/numeric-editor.spec.ts`
  - Verifies the editor shows the preserved literal for `9.0` and a large number, that ordinary values are untouched, and that a preserved literal still sorts numerically (not lexicographically). Sorting and filters are enabled in the fixture.
- Full unit suite: `npm run test:unit --prefix handsontable` — 3279 passing.

### Types of changes

### Related issue(s):

1. DEV-1161

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-1161

---

> [!NOTE]
> **Medium Risk**
> Changes core numeric value storage and filter equality for opted-in columns; default-off limits blast radius, but mixed string/number cell values can affect sort, filter-by-value, and downstream consumers.
> 
> **Overview**
> Adds an opt-in **`preserveNumericLiteral`** option (default `false`) on numeric cells so the editor can keep what you typed when parsing to a JavaScript number would drop trailing fractional zeros (e.g. `9.0` → `9`) or precision beyond `Number.MAX_SAFE_INTEGER`.
> 
> When enabled, the numeric **`valueSetter`** uses new **`isLossyNumericConversion`** logic (with **`toPlainDecimalString`** normalization) to store the trimmed literal string only for plain dot-decimal input that would lose information; lossless values still become numbers. Grouped thousands, comma-decimal literals, hex, and overflow/underflow cases are excluded from preservation.
> 
> **Equal** / **not equal** filter conditions compare numeric cells with this option numerically so `9.0` matches `9`, while columns without the option keep string equality. Docs, meta schema, unit tests, and Playwright E2E (editor, sort, filter, formulas) cover the behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1334a2137db9da5ed7110ac50e48ead7f491e655. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changelogs/13170.json
+++ b/.changelogs/13170.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Added the `preserveNumericLiteral` option to the numeric cell type, which keeps the exact value you type in the cell editor (preserving trailing decimal zeros and large-number precision).",
+  "type": "added",
+  "issueOrPR": 13170,
+  "breaking": false,
+  "framework": "none"
+}
--- a/handsontable/src/cellTypes/numericType/accessors/valueSetter.ts
+++ b/handsontable/src/cellTypes/numericType/accessors/valueSetter.ts
@@ -2,6 +2,8 @@ import {
   isCommaThousandsGroupedInteger,
   isDotThousandsGroupedInteger,
   isDotThousandsGroupedFloat,
+  isLossyNumericConversion,
+  isNumeric,
   isNumericLike,
   getParsedNumber,
 } from '../../../helpers/number';
@@ -67,18 +69,38 @@ export function valueSetter(newValue: unknown, _row: number, _column: number, ce
   }
 
   const decimalSeparator = getCellDecimalSeparator(cellMeta);
+  const isCommaGrouped = isCommaThousandsGroupedInteger(newValue, decimalSeparator);
+  const isDotGroupedInteger = isDotThousandsGroupedInteger(newValue, decimalSeparator);
+  const isDotGroupedFloat = isDotThousandsGroupedFloat(newValue, decimalSeparator);
 
-  if (
-    isNumericLike(newValue) ||
-    isCommaThousandsGroupedInteger(newValue, decimalSeparator) ||
-    isDotThousandsGroupedInteger(newValue, decimalSeparator) ||
-    isDotThousandsGroupedFloat(newValue, decimalSeparator)
-  ) {
+  if (isNumericLike(newValue) || isCommaGrouped || isDotGroupedInteger || isDotGroupedFloat) {
     const parsedNumber = getParsedNumber(newValue, {
       decimalSeparator,
     });
 
-    return isNullish(parsedNumber) ? newValue : parsedNumber;
+    if (isNullish(parsedNumber)) {
+      return newValue;
+    }
+
+    const isGrouped = isCommaGrouped || isDotGroupedInteger || isDotGroupedFloat;
+
+    // Opt-in: keep the original literal when parsing to a JS number would lose information
+    // (trailing fractional zeros like `9.0`, or precision beyond Number.MAX_SAFE_INTEGER). This
+    // preserves what the user typed in the editor, matching spreadsheet behavior. Gated behind
+    // `preserveNumericLiteral` (default `false`) so existing configurations are unchanged. The
+    // guard is limited to plain, dot-decimal literals (`isNumeric`): grouping removal
+    // (e.g. `7.000` → 7000) is intended, and preserving a comma-bearing literal (`9,0`, `0,100`)
+    // would break numeric rendering and validation, which do not accept a comma decimal separator.
+    if (
+      cellMeta.preserveNumericLiteral === true &&
+      !isGrouped &&
+      isNumeric(newValue) &&
+      isLossyNumericConversion(newValue, parsedNumber)
+    ) {
+      return newValue.trim();
+    }
+
+    return parsedNumber;
   }
 
   return newValue;
--- a/handsontable/src/cellTypes/numericType/accessors/valueSetter.ts
+++ b/handsontable/src/cellTypes/numericType/accessors/valueSetter.ts
@@ -87,6 +87,14 @@ export function valueSetter(newValue: unknown, _row: number, _column: number, ce
     // with `parseFloat`, which stops at the `x` — that mismatch is a parser artifact, not
     // information loss, so hex input must never be preserved.
     const isHexLiteral = /^[+-]?0x/i.test(newValue.trim());
+    // Literals outside the finite double range (`1e400` → `Infinity`) and non-zero literals that
+    // underflow to zero (`1e-400` → `0`) overflow the number's magnitude rather than losing
+    // trailing zeros or integer precision, so they are stored as the parsed number like any
+    // other input. The underflow check reads only mantissa digits — exponent digits must not
+    // make a genuine zero literal (`0.0e-5`) look non-zero.
+    const mantissaDigits = newValue.trim().replace(/e[+-]?\d+$/i, '');
+    const isMagnitudeOverflow = !Number.isFinite(parsedNumber) ||
+      (parsedNumber === 0 && /[1-9]/.test(mantissaDigits));
 
     // Opt-in: keep the original literal when parsing to a JS number would lose information
     // (trailing fractional zeros like `9.0`, or precision beyond Number.MAX_SAFE_INTEGER). This
@@ -99,6 +107,7 @@ export function valueSetter(newValue: unknown, _row: number, _column: number, ce
       cellMeta.preserveNumericLiteral === true &&
       !isGrouped &&
       !isHexLiteral &&
+      !isMagnitudeOverflow &&
       isNumeric(newValue) &&
       isLossyNumericConversion(newValue, parsedNumber)
     ) {
--- a/handsontable/src/cellTypes/numericType/accessors/valueSetter.ts
+++ b/handsontable/src/cellTypes/numericType/accessors/valueSetter.ts
@@ -83,6 +83,10 @@ export function valueSetter(newValue: unknown, _row: number, _column: number, ce
     }
 
     const isGrouped = isCommaGrouped || isDotGroupedInteger || isDotGroupedFloat;
+    // `isNumeric` also accepts hexadecimal literals (`0x1A`), but `getParsedNumber` reads them
+    // with `parseFloat`, which stops at the `x` — that mismatch is a parser artifact, not
+    // information loss, so hex input must never be preserved.
+    const isHexLiteral = /^[+-]?0x/i.test(newValue.trim());
 
     // Opt-in: keep the original literal when parsing to a JS number would lose information
     // (trailing fractional zeros like `9.0`, or precision beyond Number.MAX_SAFE_INTEGER). This
@@ -94,6 +98,7 @@ export function valueSetter(newValue: unknown, _row: number, _column: number, ce
     if (
       cellMeta.preserveNumericLiteral === true &&
       !isGrouped &&
+      !isHexLiteral &&
       isNumeric(newValue) &&
       isLossyNumericConversion(newValue, parsedNumber)
     ) {
--- a/handsontable/src/dataMap/metaManager/metaSchema.ts
+++ b/handsontable/src/dataMap/metaManager/metaSchema.ts
@@ -4704,7 +4704,10 @@ export default (): Record<string, unknown> => {
      * string instead of the number. This preserves trailing decimal zeros (`9.0`, `9.50`) and the
      * full precision of large numbers in the cell editor, matching spreadsheet software. Values
      * that convert without loss (for example `9`, `9.5`, `1000`) are still stored as numbers, so
-     * sorting, filtering, and formula calculations are unaffected. The cell renderer still formats
+     * sorting, filtering, and formula calculations are unaffected. A preserved literal also keeps
+     * behaving like a number in those features: column sorting and filter conditions compare it
+     * numerically, and the [`Formulas`](@/api/formulas.md) engine parses the literal as a number,
+     * so functions such as `SUM` still include the cell. The cell renderer still formats
      * the value according to [`numericFormat`](@/api/options.md#numericformat); only the editor
      * shows the preserved literal.
      *
--- a/handsontable/src/dataMap/metaManager/metaSchema.ts
+++ b/handsontable/src/dataMap/metaManager/metaSchema.ts
@@ -4709,7 +4709,9 @@ export default (): Record<string, unknown> => {
      * numerically, and the [`Formulas`](@/api/formulas.md) engine parses the literal as a number,
      * so functions such as `SUM` still include the cell. The cell renderer still formats
      * the value according to [`numericFormat`](@/api/options.md#numericformat); only the editor
-     * shows the preserved literal.
+     * shows the preserved literal. One exception: the filter menu's "Filter by value" checkbox
+     * list compares values strictly, so a preserved literal (`'9.0'`) and its plain number (`9`)
+     * appear as two separate entries.
      *
      * The default is `false` so existing configurations keep their current behavior.
      *
--- a/handsontable/src/dataMap/metaManager/metaSchema.ts
+++ b/handsontable/src/dataMap/metaManager/metaSchema.ts
@@ -4692,6 +4692,44 @@ export default (): Record<string, unknown> => {
     numericFormat: undefined,
 
     /**
+     * Controls whether a [`numeric`](@/guides/cell-types/numeric-cell-type/numeric-cell-type.md)
+     * cell keeps the exact text you typed when converting it to a JavaScript number would lose
+     * information.
+     *
+     * By default (`false`), a numeric cell always stores the parsed JavaScript number, so a value
+     * like `9.0` is stored as `9` and the editor shows `9` the next time you open it. Numbers whose
+     * magnitude exceeds the safe-integer limit (`9007199254740991`) also lose precision.
+     *
+     * When set to `true`, and only when parsing would be lossy, the cell keeps the original literal
+     * string instead of the number. This preserves trailing decimal zeros (`9.0`, `9.50`) and the
+     * full precision of large numbers in the cell editor, matching spreadsheet software. Values
+     * that convert without loss (for example `9`, `9.5`, `1000`) are still stored as numbers, so
+     * sorting, filtering, and formula calculations are unaffected. The cell renderer still formats
+     * the value according to [`numericFormat`](@/api/options.md#numericformat); only the editor
+     * shows the preserved literal.
+     *
+     * The default is `false` so existing configurations keep their current behavior.
+     *
+     * @memberof Options#
+     * @since 18.1.0
+     * @type {boolean}
+     * @default false
+     * @category Core
+     *
+     * @example
+     * ```js
+     * columns: [
+     *   {
+     *     type: 'numeric',
+     *     // keep `9.0` and very large numbers as typed in the editor
+     *     preserveNumericLiteral: true,
+     *   }
+     * ],
+     * ```
+     */
+    preserveNumericLiteral: false,
+
+    /**
      * If the `observeDOMVisibility` option is set to `true`,
      * Handsontable rerenders every time it detects that the grid was made visible in the DOM.
      *
--- a/handsontable/src/helpers/number.ts
+++ b/handsontable/src/helpers/number.ts
@@ -289,7 +289,10 @@ function toPlainDecimalString(numericText: string): string {
  * Both the input and `String(parsedNumber)` are normalized to plain decimal notation before
  * comparing, so purely cosmetic differences (leading zeros, a leading `+`, `.5`/`5.`, and
  * scientific notation on either side — `1e2` vs `100`, `0.0000001` vs `1e-7`) are not treated
- * as loss.
+ * as loss. A mantissa trailing zero that a positive exponent shifts onto or left of the
+ * decimal point (`1.0e2` → `100`, `1.10e2` → `110`) survives as an integer digit of the exact
+ * value, so it is not loss either; only zeros that stay fractional after the shift
+ * (`1.10e1` → `11.0`) are dropped by parsing and count as lossy.
  *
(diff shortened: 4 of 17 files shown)


qunabu:
Task linked: [DEV-1161 Numeric cells with .0 decimal lose trailing zero in cell editor](https://app.clickup.com/t/9015210959/DEV-1161)

qunabu (dismissed):
@Haxikowy @adrianspdev  can you review this as well please

handsontable/src/dataMap/metaManager/metaSchema.ts
```diff
@@ -4691,6 +4691,44 @@ export default (): Record<string, unknown> => {
...
+     *
+     * By default (`false`), a numeric cell always stores the parsed JavaScript number, so a value
+     * like `9.0` is stored as `9` and the editor shows `9` the next time you open it. Numbers whose
+     * magnitude exceeds the safe-integer limit (`9007199254740991`) also lose precision.
+     *
+     * When set to `true`, and only when parsing would be lossy, the cell keeps the original literal
+     * string instead of the number. This preserves trailing decimal zeros (`9.0`, `9.50`) and the
+     * full precision of large numbers in the cell editor, matching spreadsheet software. Values
+     * that convert without loss (for example `9`, `9.5`, `1000`) are still stored as numbers, so
+     * sorting, filtering, and formula calculations are unaffected. The cell renderer still formats
```
Haxikowy: "formula calculations are unaffected" — was this tested? With the Formulas plugin, HyperFormula receives the raw string `"9.0"`. If it treats it as text, `SUM` skips the cell. Add a Formulas test with a preserved literal or drop the claim.

tests/fixtures/demo/numeric.html
```diff
@@ -0,0 +1,51 @@
...
+    document.write('<link rel="stylesheet" href="/handsontable/styles/ht-theme-' + window.htTheme + '.min.css">');
+  </script>
+  <style> body { font-family: sans-serif; margin: 1rem; } </style>
+</head>
+<body>
+  <!--
+    Numeric-cell E2E fixture. A single numeric-typed column with the opt-in
+    `preserveNumericLiteral` option enabled exercises the numeric editor /
+    valueSetter path. Column sorting and filters are on so specs can verify that
+    a preserved literal string still sorts and filters numerically. Cells carry
```
Haxikowy, line 21: This says specs verify that a preserved literal "sorts and filters numerically", but `numeric-editor.spec.ts` has no filter spec — and equality filtering is currently broken for preserved literals (see the `eq` thread). Add the spec or drop "and filters".

handsontable/src/cellTypes/numericType/accessors/valueSetter.ts
```diff
@@ -67,18 +69,38 @@ export function valueSetter(newValue: unknown, _row: number, _column: number, ce
...
+    // (trailing fractional zeros like `9.0`, or precision beyond Number.MAX_SAFE_INTEGER). This
+    // preserves what the user typed in the editor, matching spreadsheet behavior. Gated behind
+    // `preserveNumericLiteral` (default `false`) so existing configurations are unchanged. The
+    // guard is limited to plain, dot-decimal literals (`isNumeric`): grouping removal
+    // (e.g. `7.000` → 7000) is intended, and preserving a comma-bearing literal (`9,0`, `0,100`)
+    // would break numeric rendering and validation, which do not accept a comma decimal separator.
+    if (
+      cellMeta.preserveNumericLiteral === true &&
+      !isGrouped &&
+      isNumeric(newValue) &&
```
Haxikowy, line 111: `isNumeric` accepts hex and `parseFloat("0x1A")` is `0`, so with the option on `0x1A` counts as lossy and stays in the data as the string `"0x1A"`. Exclude hex from this guard.

handsontable/src/helpers/number.ts
```diff
@@ -242,6 +242,42 @@ export function getParsedNumber(numericData: string, options: { decimalSeparator
...
+  }
+
+  if (normalized.endsWith('.')) {
+    normalized = normalized.slice(0, -1);
+  }
+
+  const canonical = `${negative ? '-' : ''}${normalized}`;
+
+  return canonical !== String(parsedNumber);
+}
```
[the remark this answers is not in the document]
Haxikowy: Agree, repros. `String(1e-7)` is `"1e-7"`, so `0.0000001` and `1e2` get flagged as lossy and stored as strings even though parsing loses nothing. Contradicts the JSDoc above and the option docs in metaSchema. Normalize both sides to plain decimal notation before comparing.

handsontable/src/cellTypes/numericType/accessors/valueSetter.ts
```diff
@@ -67,18 +69,38 @@ export function valueSetter(newValue: unknown, _row: number, _column: number, ce
...
+      isLossyNumericConversion(newValue, parsedNumber)
+    ) {
+      return newValue.trim();
+    }
```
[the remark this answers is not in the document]
Haxikowy: Confirmed. `eq` in `src/plugins/filters/condition/equal.ts` compares stringified values, so `"9.0"` never matches a filter input of `9`. `gt`/`lt`/`between` already special-case `dataRow.meta.type === 'numeric'` — add the same branch to `eq` (`Number(dataRow.value) === Number(value)` when both sides pass `isNumeric`), `neq` gets it for free. Needs cases in `equal.unit.ts` and `notEqual.unit.ts`.

Haxikowy (changes requested):

handsontable/src/helpers/number.ts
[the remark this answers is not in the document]
adrianspdev: Fixed in 92679c31ef. `isLossyNumericConversion` now normalizes both the input and `String(parsedNumber)` to plain decimal notation (new `toPlainDecimalString` helper) before comparing, so `0.0000001`, `1e2`, and `1e-7` are stored as numbers again. A genuine trailing-zero loss written in scientific notation (`1.2300e2`) is still detected. Unit tests added for both directions.

handsontable/src/cellTypes/numericType/accessors/valueSetter.ts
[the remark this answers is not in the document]
adrianspdev: Fixed in 92679c31ef. `eq` now compares numeric-typed cells by value — `Number(dataRow.value) === Number(value)` when both sides pass `isNumeric` — matching the `gt`/`lt`/`between` approach; `neq` inherits it. Cases added to `equal.unit.ts` and `notEqual.unit.ts`, plus a Playwright spec that filters a preserved literal through the dropdown "Is equal to" UI.

handsontable/src/dataMap/metaManager/metaSchema.ts
adrianspdev: Tested now — the claim holds. HyperFormula parses numeric-looking raw content, so a cell holding the string `"9.0"` evaluates as the number 9 and `SUM` includes it (verified at engine level and in-browser). Added `tests/e2e/numeric-formulas.spec.ts` with a `numeric-formulas.html` fixture (Formulas plugin + `=SUM(A1:A3)`): after editing a cell to `9.0`, SUM shows 53 and the editor still shows the literal. Also made the JSDoc explicit about preserved literals in 92679c31ef.

tests/fixtures/demo/numeric.html
adrianspdev: Spec added in 92679c31ef: `numeric-editor.spec.ts` now filters a preserved literal via the dropdown "Is equal to" condition (input `9` matches the cell holding `"9.0"`) and asserts a single remaining row. Works together with the `eq` numeric-comparison fix from the other thread.

handsontable/src/cellTypes/numericType/accessors/valueSetter.ts
adrianspdev: Fixed in 92679c31ef. Hexadecimal literals are excluded from the preserve guard (`/^[+-]?0x/i`), so `0x1A` keeps the pre-existing behavior (`parseFloat` → `0`) with the option on or off. Unit test asserts both paths return the same result.

handsontable/src/helpers/number.ts
```diff
@@ -242,6 +242,66 @@ export function getParsedNumber(numericData: string, options: { decimalSeparator
...
+  if (pointIndex < 1) {
+    digits = `${'0'.repeat(1 - pointIndex)}${digits}`;
+    pointIndex = 1;
+  }
+
+  const integerPart = digits.slice(0, pointIndex).replace(/^0+(?=\d)/, '');
+  const fractionPart = digits.slice(pointIndex);
+
+  return `${sign}${integerPart}${fractionPart.length > 0 ? `.${fractionPart}` : ''}`;
+}
```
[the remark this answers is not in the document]
adrianspdev: Partially incorrect. `1.10e1` IS detected as lossy: it expands to `11.0`, the zero stays fractional, comparison against `String(11)` fails, and the literal is preserved (unit-tested in 2d1a31252b).

`1.0e2` and `1.10e2` are intentionally NOT lossy: the mantissa zero is shifted onto an integer position by the exponent and survives as a digit of the exact parsed value (`100`, `110`) — no typed digit disappears. That is the same class as `1e2` → `100`, which the earlier thread confirmed must not be preserved. The contract is information loss (digits dropped by parsing), not notation preservation; treating `1.0e2` as lossy while `1e2` is not would be inconsistent.

Codified in 2d1a31252b: JSDoc now spells out the shifted-zero rule, and unit tests pin `1.0e2`/`1.10e2` (not lossy) vs `1.10e1`/`1.2300e2` (lossy).

handsontable/src/helpers/number.ts
```diff
@@ -242,6 +242,69 @@ export function getParsedNumber(numericData: string, options: { decimalSeparator
...
+  if (match === null || (match[2] === '' && (match[3] ?? '') === '')) {
+    return numericText.trim();
+  }
+
+  const sign = match[1] === '-' ? '-' : '';
+  const integerDigits = match[2];
+  const fractionDigits = match[3] ?? '';
+  const exponent = Number.parseInt(match[4] ?? '0', 10);
+  let digits = `${integerDigits}${fractionDigits}`;
+  let pointIndex = integerDigits.length + exponent;
```
budnix: Typing `1e999999999` into a cell with this option on crashes here — `pointIndex` becomes 1,000,000,000 and `padEnd` throws `RangeError: Invalid string length`. `1e-999999999` hits the same wall in the `repeat` call just below; I ran both through the helper to confirm. Because the throw happens inside `valueSetter`, it takes `setDataAtCell` down with it, so the edit is lost instead of just being parsed oddly. Even a smaller exponent like `1e100000000` builds a ~100 MB string and freezes the tab, so some bound seems worth adding.

```suggestion
  let pointIndex = integerDigits.length + exponent;

  // A double tops out near 1e308, so a larger exponent can never describe a finite number.
  // Without this bound the two expansions below allocate an enormous string, or throw
  // `RangeError: Invalid string length` for inputs like `1e999999999`.
  if (Math.abs(pointIndex) > 1000) {
    return numericText.trim();
  }
```

budnix (commented):
Went through the diff with a focus on the new number helper and the filter change. Five notes inline. The `number.ts` one is a real crash reachable by typing into a cell, so that's the one I'd fix first — I confirmed it by running the helper directly. The rest are behavior and docs points around the flag not covering everything it looks like it covers.

handsontable/src/cellTypes/numericType/accessors/valueSetter.ts
```diff
@@ -67,18 +69,43 @@ export function valueSetter(newValue: unknown, _row: number, _column: number, ce
...
+    // (e.g. `7.000` → 7000) is intended, and preserving a comma-bearing literal (`9,0`, `0,100`)
+    // would break numeric rendering and validation, which do not accept a comma decimal separator.
+    if (
+      cellMeta.preserveNumericLiteral === true &&
+      !isGrouped &&
+      !isHexLiteral &&
+      isNumeric(newValue) &&
+      isLossyNumericConversion(newValue, parsedNumber)
+    ) {
+      return newValue.trim();
```
budnix, line 114: `-0` ends up stored as the string `"-0"` here — it normalizes to `-0` while `String(-0)` is `"0"`, so the lossy check returns true. `1e400` and `1e-400` do the same, since they parse to `Infinity` and `0`. The option docs only promise trailing fractional zeros and values past `MAX_SAFE_INTEGER`, so all three are surprises, and user code doing `getDataAtCell(...) + 1` starts joining text instead of adding. Skipping the preserve path when `parsedNumber` is not finite or is `0` would cover all of them.

handsontable/src/plugins/filters/condition/equal.ts
```diff
@@ -17,12 +18,20 @@ type DataRow = {
...
+ * literal string such as `9.0` still matches a filter input of `9`. All other cells compare
+ * by locale-lowercased string equality.
+ *
  * @param {object} dataRow The object which holds and describes the single cell value.
  * @param {Array} inputValues An array of values to compare with.
  * @param {Array} inputValues."0" Value to check if it same as row's data.
  * @returns {boolean}
  */
 export function condition(dataRow: DataRow, [value]: unknown[]) {
+  if (dataRow.meta.type === 'numeric' && isNumeric(dataRow.value) && isNumeric(value)) {
```
budnix: This branch isn't gated by `preserveNumericLiteral`, so it changes `eq`/`neq` for every existing numeric column, not only opted-in ones. Four inputs that returned `false` before now match: filter `1e2` matches a cell of `100`, `.5` matches `0.5`, `0x10` matches `16` (`isNumeric` accepts hex), and `-0` matches `0`. Some of those are arguably an improvement, but it's a silent change for anyone already filtering numeric columns. Was leaving it ungated deliberate, or should it check the flag the way the setter does?

.changelogs/13170.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Added the `preserveNumericLiteral` option to the numeric cell type, which keeps the exact value you type in the cell editor (preserving trailing decimal zeros and large-number precision).",
+  "type": "added",
+  "issueOrPR": 13170,
+  "breaking": false,
```
budnix, line 6: Because the `eq`/`neq` change in `equal.ts` isn't gated by the option, this PR does shift behavior for people who never switch the flag on — `1e2` starts matching `100`, `0x10` starts matching `16`. Worth either flipping this to `true`, or gating the filter change so the flag is genuinely the only switch.

handsontable/src/dataMap/metaManager/metaSchema.ts
```diff
@@ -4691,6 +4691,47 @@ export default (): Record<string, unknown> => {
...
+     * By default (`false`), a numeric cell always stores the parsed JavaScript number, so a value
+     * like `9.0` is stored as `9` and the editor shows `9` the next time you open it. Numbers whose
+     * magnitude exceeds the safe-integer limit (`9007199254740991`) also lose precision.
+     *
+     * When set to `true`, and only when parsing would be lossy, the cell keeps the original literal
+     * string instead of the number. This preserves trailing decimal zeros (`9.0`, `9.50`) and the
+     * full precision of large numbers in the cell editor, matching spreadsheet software. Values
+     * that convert without loss (for example `9`, `9.5`, `1000`) are still stored as numbers, so
+     * sorting, filtering, and formula calculations are unaffected. A preserved literal also keeps
+     * behaving like a number in those features: column sorting and filter conditions compare it
```
budnix, line 4708: This says filter conditions compare a preserved literal numerically, which now holds for `eq`/`neq`, but the "Filter by value" checkbox list uses strict `Set` membership (`createArrayAssertion` in `plugins/filters/utils.ts`). So a column holding both `9` and a preserved `"9.0"` shows two separate checkboxes that have to be ticked one at a time. Would be good to note that limit here, since the cell type still declares `dataType: 'number'`.

handsontable/src/helpers/number.ts
adrianspdev: Fixed in fcd366bd26 and refined in 1334a2137d. The first commit added the suggested bound on `pointIndex`, but Bugbot then correctly pointed out that `pointIndex` includes the digit count, so a 1000+-digit leading-zero literal (`000…0005`) tripped the bound and was wrongly preserved. The final version bounds `Math.abs(exponent) > 1000` instead — the exponent is the only unbounded driver of the expansions (padding = exponent − fraction length, prepend ≤ 1 + |exponent|), so the worst-case expansion stays around 1000 characters. `1e999999999`, `1e-999999999`, and `1e100000000` no longer throw or allocate; unit tests cover all three plus the leading-zero case.

handsontable/src/cellTypes/numericType/accessors/valueSetter.ts
adrianspdev: Fixed in fcd366bd26, with one deviation from the suggested guard: skipping the preserve path whenever `parsedNumber` is `0` would also stop preserving `0.0`/`0.00`, and trailing fractional zeros are exactly what the option documents. What ships instead:

- `-0`: the lossy check now compares against `'-0'` when `Object.is(parsedNumber, -0)` — the number `-0` stores the sign exactly, so a plain `-0` input is not lossy and is stored as the number `-0`. (`-0.0` is still preserved — genuine trailing zero.)
- `1e400`: the setter skips the preserve path for non-finite parses (`!Number.isFinite(parsedNumber)`), so it stores `Infinity` as before the option existed.
- `1e-400`: underflow is detected as `parsedNumber === 0` with a non-zero mantissa digit (exponent digits excluded, so `0.0e-5` stays a genuine zero literal), and stored as `0`.

Unit tests cover all three plus the `0.0` preservation that the simpler guard would have dropped.

handsontable/src/plugins/filters/condition/equal.ts
adrianspdev: Not deliberate — gated in fcd366bd26. The numeric branch now requires `dataRow.meta.preserveNumericLiteral === true`, so columns that never opt in keep the historical string comparison (`1e2` no longer matches `100`, `0x10` no longer matches `16`). Unit tests for `eq` and `neq` cover both the gated numeric comparison and the ungated historical behavior.

.changelogs/13170.json
adrianspdev: Resolved by gating the `eq`/`neq` numeric comparison behind `preserveNumericLiteral` (fcd366bd26) — with the flag off, nothing changes for existing users, so `breaking: false` is accurate as is.

handsontable/src/dataMap/metaManager/metaSchema.ts
adrianspdev: Added in fcd366bd26 — both the `preserveNumericLiteral` JSDoc here and the numeric-cell-type guide now note that the "Filter by value" checkbox list compares values strictly, so a preserved literal (`'9.0'`) and its plain number (`9`) appear as two separate entries.

handsontable/src/helpers/number.ts
```diff
@@ -242,6 +242,81 @@ export function getParsedNumber(numericData: string, options: { decimalSeparator
...
+
+  // A double tops out near 1e308, so a decimal point shifted this far can never describe
+  // a finite number. Without this bound, the two expansions below allocate an enormous
+  // string or throw `RangeError: Invalid string length` for inputs like `1e999999999`.
+  if (Math.abs(pointIndex) > 1000) {
+    return numericText.trim();
+  }
```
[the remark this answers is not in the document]
adrianspdev: Fixed in 1334a2137d — the bound now applies to `Math.abs(exponent)` only, so long leading-zero literals normalize again and are stored as numbers (covered by unit tests). One note: the `0e-N` part is not a regression — those forms read as lossy before the bound existed too (the expanded trailing zeros differ from `"0"`), which is consistent with `0.0` being preserved as a trailing-zero literal.

budnix (approved):

Haxikowy (approved):

PR #13178, opened 2026-08-11
DEV-2193: Fix rows drifting when the tallest cell is in a frozen column

budnix:
### Context

The PR fixes rows drifting out of alignment when the tallest cell in a row sits in a frozen (inline-start) column.

The master table renders a **contiguous** column band starting at the column under the horizontal scroll offset, so with `fixedColumnsStart` set it skips the frozen columns as soon as that band starts past column 0. Row heights are measured from the rendered DOM (`markOversizedRows`, on the master), so such a row got recorded at its provided height while the frozen overlays rendered it at its content height. Every row below it then drifted by the difference — ~20px per wrapped row in the reported demo.

**This is not the regression the report attributes to #12189.** 16.2.0 drifts the same way as soon as the grid is scrolled past the master's overscan reach — measured with 100 columns and `fixedColumnsStart: 3`:

| version | scrollLeft 0 | 1000 | 3000 | 6000 |
|---|---|---|---|---|
| 16.2.0 | 0px | 0px | **640px** | **1760px** |
| 18.0.0 | **1760px** | **1440px** | **1440px** | **1440px** |

What #12189 changed is the overscan reach — `viewportColumnRenderingOffset: 'auto'` went from ~10 columns to 1 (`tableView.ts:1360`) — which moved the failure from "only once scrolled" to "at rest in a narrow grid", which is the reported case. The row-height measurement contract has been broken for frozen columns the whole time, so this fixes the contract at every scroll position rather than restoring the old overscan.

### The fix

`syncOversizedRowsWithFrozenOverlays` is the row-height twin of the existing `syncOversizedColumnHeadersWithFrozenOverlays`. After the frozen overlays have rendered it measures **all three tables that render frozen columns** — the inline-start clone, which mirrors the master's row band, plus the two corners, which hold the frozen top and bottom rows the clone does not — and re-applies the resulting heights to the tables that render those rows without the frozen columns.

It records into `wtViewport.oversizedRows` rather than only patching the DOM, because the hider height comes from the summed row heights (`spreaderSize`), not from measuring the rendered table — a DOM-only patch would fix the visible rows but leave the scrollbar short by the accumulated drift.

#### Frozen-derived records survive the wipe

The record cannot be re-created by the master — it does not render that column — so it must not be destroyed before the consumers that need it. `wtViewport.frozenOversizedRows` marks the keys the sync owns, `resetOversizedRows` skips them, and `resetFrozenOversizedRows` clears them in the seam between the master's render and `wtOverlays.refresh()` instead. That one ordering choice is what makes the rest correct:

| Consumer | Sees the record because |
|---|---|
| The master's own render | It rendered before the clear |
| The row-height cache + viewport calculators | They are built before the clear |
| The master's `markOversizedRows` | The record was never in its wiped map, so it cannot mistake it for a shrunk row |
| The bottom clone's `markOversizedRows` | Same — it cannot wipe a record only the sync can create |
| The inline-start clone / corners | They render *after* the clear, so at their natural height — re-measurable, no ratchet |

#### Who owns a height

Most of the difficulty is not the measurement, it is the bookkeeping around which table owns a row's height. Three rules, each of which looks like harmless code and is not:

- **Only rows the frozen pass actually measured may be marked frozen-derived.** Registering every oversized row in the band instead looks equivalent: a row that is tall because of a *scrollable* column gets adopted, the frozen overlays can never re-detect a height they never saw, so the next draw reads it as shrunk and drops it and the master rediscovers it on the one after. The row oscillates every other draw, and that is the common case — any wrapped row in a grid with frozen columns.
- **Ownership moves back when the master out-measures the frozen side.** A row can be tall in a frozen column and taller still in a scrollable one. Leaving the mark on is invisible and never settles: the next draw clears a height only the master can recreate, the frozen pass re-records its own shorter one, and both invalidate the row-height cache on every draw.
- **On a draw where a height changed, release and re-measure the master first.** While a frozen height dominates, the master's own pass measures the forced value and records nothing — so when the frozen part goes away there is nothing recording the height the master still legitimately has.
- **The 1px band-boundary flip is a DOM problem, not a cache problem, and the two must stay apart.** The band's first `<tr>` gains a 1px border-top, so a row's total height changes by 1px purely from scroll position, and the measured value then alternates across draws. The row-height cache keeps a `> 1` tolerance before invalidating — tightening it invalidates on every draw for as long as the row sits at the boundary. But the DOM still has to be reconciled: the frozen overlay holding the tall cell cannot render in one pixel less than its content needs, while the master honours whatever height it was given, so the two drift 1px apart and every row below with them. Re-applying the current record to every table is what closes it.

#### The window the clear opens

Between clearing the frozen-derived records and putting them back, `oversizedRows` is missing every one of them. Anything reading it in that window gets an incomplete answer, and because the records return *unchanged*, nothing invalidates afterwards to correct it:

- **A row-height cache built in the window** is short by all of them — a scrollbar that cannot reach the end of the grid, with every rendered row still correct. The bottom clone reaches this: it renders *and measures itself* inside `wtOverlays.refresh()`. `PositionCache#buildSeq` is snapshotted at the clear and compared after; a changed counter means drop the build and re-size the elements. (`isCurrent()` cannot answer this — invalidate-then-rebuild reads `true` at both ends.)
- **Only overlays that actually rendered this draw may be measured.** `Overlay#refresh` is a no-op when `needFullRender` is false, and a skipped clone still owns its previous draw's DOM.
- **The viewport calculators are built before the frozen overlays render**, so a frozen-derived height cannot be in them. An ordinary oversized row never has this problem — the master invalidates inside `renderCellBand`, which is earlier. The sync now reports whether it changed anything so the draw cycle can rebuild them.

`RenderSizeProbe` is kept a faithful mirror alongside this: it measured two of the tables that now write records, so a frozen top row went missing from it while its characterization spec stayed green.

#### Review rounds

Cursor Bugbot reviewed this four times and found real bugs in three of those rounds — every one of them in the ownership bookkeeping rather than in the measurement. Each claim was verified against the code, and each confirmed one reproduced with a failing test before any fix was written. Two claims did not survive that check and are documented rather than "fixed":

- **MergeCells per-overlay heights.** `modifyRowHeightByOverlayName` inflates the anchor row per overlay, so a frozen clone renders it at the whole block's height while the overlay-agnostic `getRowHeight` that `markOversizedRows` compares against reports one row. The premise is exact, but it produces no record — the inflated height is written on a TD whose `rowspan` covers precisely the rows it accounts for, so no single TR measures tall. Four constructions tried, including the clamped-block case the hook exists for. A spec pins the outcome.
- **The `needFullRender` filter** restores a guarantee the master has structurally, but no failing case could be constructed for it. It ships as a restored invariant, not a proven bug.

One cleanup is deliberately **not** done: sharing the row-height write with `TableRenderer#render`. The duplication is documented in `walkontable/AGENTS.md` rather than restructuring the renderer's hot loop inside this PR.

### How has this been tested?

New Playwright coverage in `tests/e2e/walkontable/frozen-column-row-heights.spec.ts`, running under all three themes. The tall cell is a fixed-height block rather than wrapping text, so the row height does not depend on the runner's font metrics.

```bash
cd tests && npx playwright test walkontable      # 60 passed (3 themes)
```

18 cases. Beyond the basics — the master really is skipping the frozen columns (the premise the rest rest on), equal row height in the master and the overlay, no drift below, no drift after scrolling away, a tall cell in a frozen **top** row staying aligned between the top overlay and its corner — the ones worth naming are those that pin something invisible:

- a row tall in a **scrollable** column staying put across consecutive draws, and the same row when a frozen height is added on top of it, in both size orders;
- ownership surviving the band's first row, which measures 1px taller than the same row anywhere else;
- the scroll range staying whole when the **bottom clone** invalidates the cache mid-draw, asserted as cached-sum vs freshly-rebuilt-sum rather than a pixel count;
- the visible row range matching the heights the draw ended with;
- `RenderSizeProbe` reproducing every engine record, including rows only a corner renders;
- the panes staying together when the band boundary moves **onto** an already-tall row, which is a different case from the content appearing while the row is already at the boundary — that one measures the right total straight away and hides the bug;
- **zero row-height cache invalidations** in a settled grid — the one symptom of a bookkeeping mistake that leaves every visible thing correct.

Every behavioral case was confirmed to **fail** against a build of the preceding commit and pass against this one.

Full suites, all on the final build:

```bash
cd tests && npx playwright test walkontable        # 60 passed (3 themes)
npm run test:walkontable --prefix handsontable     # 816 specs, 0 failures
npm run test:unit --prefix handsontable            # 3267 tests, 0 failures
npm run test:e2e --prefix handsontable             # 9606 specs, 0 failures
npm run test:types --prefix handsontable           # clean
npm run eslint --prefix handsontable               # 0 errors
npm run stylelint --prefix handsontable            # clean
npx lefthook run pre-push                          # test-presence gate passes
```

Manually verified against the reported demo ([2t1u75e3g2](https://demos.handsontable.com/edit/2t1u75e3g2)) rebuilt locally at its exact configuration, at `fixedColumnsStart` 2 and 3, at rest and scrolled to the end: 0px drift everywhere, with the master still starting past column 0 so the rendering optimisation is intact. Also on the 100-column case above: 0px at every scroll position.

### Performance

This touches the single-pass render path, so it is verified with counters rather than wall time alone. Per full draw, before → after, medians of 5 rounds × 120 renders with the builds alternated within each round (50 columns, 3 frozen, 10k rows). "Before" is a build of the merge base, `ba3df0b5e8`:

| case | ms | geometry reads | `createVisibleCalculators` | row-cache invalidations | row-height writes |
|---|---|---|---|---|---|
| no frozen columns | 1.3 → 1.2 | 35 → 35 | 0 → 0 | 0 → 0 | 21 → 21 |
| frozen, nothing oversized | 1.3 → 1.3 | 35 → 37 | 0 → 0 | 0 → 0 | 21 → 21 |
| frozen + tall frozen cell | 1.4 → 1.4 | 35 → 73 | 0 → 0 | 0 → 0 | **21 → 36** |
| frozen, nothing oversized, non-uniform `rowHeights` | 1.6 → 1.6 | 57 → 59 | 1 → 1 | 0 → 0 | 21 → 21 |
| frozen + tall frozen cell, non-uniform `rowHeights` | 1.7 → 1.6 | 57 → 95 | 1 → 1 | 0 → 0 | **21 → 36** |

Re-run identically twice; the counters were byte-identical and wall time within 0.1ms.

**Zero row-height cache invalidations in steady state.** The per-draw layout snapshot is never dropped mid-draw and `skipSecondPass` engages exactly as often as before. That matters most on the non-uniform rows: with `rowHeights`/`minRowHeights` as an array or function, or any non-AutoRowSize `modifyRowHeight` hook, `PositionCache` has no sparse path and every invalidation is a full prefix-sum walk over all rows. An earlier iteration of this PR did invalidate twice per draw there — 1.8 → 4.2ms at 10k rows, 10.2ms at 40k, scaling linearly — which is what prompted the counter-based verification.

**The invalidation counter does not see everything, so the last column exists.** The frozen sync re-applies the current row heights to the master, the frozen overlays and the top/bottom clones on every draw that has a frozen record. That writes heights without invalidating anything, so it is invisible to every other counter here. Measured, it is **+15 height lookups per draw** on the two frozen-tall cases — one extra pass over the rendered band — and wall time does not move. It is O(rendered rows), not O(total rows), which the 40k-row check confirms:

| case (40k rows) | ms | invalidations | row-height writes |
|---|---|---|---|
| frozen + tall frozen cell | 1.4 → 1.5 | 0 → 0 | 21 → 36 |
| frozen + tall frozen cell, non-uniform | 1.6 → 1.6 | 0 → 0 | 21 → 36 |

The added geometry reads on the frozen-tall cases are the frozen tables' own measurement. They are batched **before** any write (measure → re-apply → resize), so the read/write ordering the `GeometryReader` proxy exists to protect is unchanged.

The `performance-tests/` suite reports no regression on the scenarios large enough to be stable (Cell Editing and the four Scroll scenarios, 300–4600ms: all within ±0.9%). Its small scenarios are not usable as a signal here — running it **twice on identical code** reported *"Sorting regressed +131.8%"*, and every flag in the real before/after run was in that same band. None of its scenarios use frozen columns, so the new code returns after two settings reads in all of them.

### Types of changes

### Related issue(s):

1. DEV-2193

#### Known gaps

- **Lost odd-row striping** — a separate, font-independent consequence of the same root cause, and **not** addressed here. A `beforeRenderer` that bails out on `column !== 0` never fires for master rows once the master stops rendering column 0, so a grid that stripes rows from that hook loses the striping in the scrollable area. This is visible in the reported demo at its exact configuration. It needs its own decision about whether the hook contract or the demo changes.
- **`fixedRowsBottom` is covered by the mechanism and now by one spec** (the bottom clone's mid-draw invalidation), but the frozen-bottom counterpart of the frozen-top alignment case has none.
- **`scrollViewportTo()` followed by a redraw collapses the vertical scroll extent** and it stays collapsed: the hider drops to the total that ignores the frozen row's height (790px → 612px in the reproduction). Ordinary scrolling, a data change, and a redraw at rest are all unaffected, and `scrollViewportTo` is public API behind keyboard navigation and go-to-cell, so this is worth fixing — but it is **not fixed here** and not yet diagnosed beyond the likely cause, the hider being sized inside `wtOverlays.refresh()` while the frozen records are cleared.

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2193


qunabu:
Task linked: [DEV-2193 [17.1.0+] misaligned rows](https://app.clickup.com/t/9015210959/DEV-2193)

demtario (approved):

PR #13182, opened 2026-08-12
DEV-2036: Revalidate dependent formula cells after undo and redo

budnix:
### Context

The PR fixes a formula cell keeping the `htInvalid` class after an undo restored a correct value.

Steps from the report: with `type: 'numeric'` and a formula cell such as `=A1+B1+C1+D1+E1`, type `aaa` into one of the referenced cells. The formula cell turns into `#VALUE!` and goes red. Undo the change. The formula recalculates and the text goes back to the correct number, but the cell stays red forever.

Root cause: `isBlockedSource()` in `handsontable/src/plugins/formulas/formulas.ts` makes `#onAfterSetDataAtCell` and `#onAfterSetSourceDataAtCell` return early for `'UndoRedo.undo'` and `'UndoRedo.redo'`, because HyperFormula reverts those changes through its own undo stack (`beforeUndo` calls `engine.undo()`). Those two listeners are the only callers of `validateDependentCells()`, so nothing revalidated the cells that changed only because a formula recalculated. The formula cell kept the `valid: false` flag it was given while it held `#VALUE!`, and the renderer kept stamping `htInvalid` on it.

A normal edit that fixes the same cell did clear the class, which is what made the failure specific to undo and redo.

### How it works

- `beforeUndo` / `beforeRedo` keep the change list that `engine.undo()` / `engine.redo()` returns.
- `#onAfterSetDataAtCell` and `#onAfterSetSourceDataAtCell` register the write before their existing early return.
- `afterUndo` / `afterRedo` hand the engine's list to the existing `validateDependentCells()`.

Two distinctions carry the logic, both of them learned from review:

**When to validate.** Only when the action wrote cell data. That covers undoing an edit, which writes through `setDataAtCell`, and undoing a row or column removal, which restores the data with `setSourceDataAtCell` (`undoRedo/actions/removeRow.ts:152`, `removeColumn.ts:214`) — from any other source that write validates dependants too. Actions that only reorder or hide — `row_move`, `col_sort`, `filter`, `merge_cells` — write nothing, do not validate dependants outside of undo either, and are skipped.

**Which cells to exclude.** Only the cells written through `setDataAtCell`, because those are the ones the Core validates itself and would otherwise be validated twice. `setSourceDataAtCell` does **not** run the cell validator — it runs `sourceDataValidator` (`dataMap/sourceDataValidator.ts`), a separate mechanism that never touches the `valid` flag — so the cells a removal undo restores have to stay in the validation pass. Excluding them left them validated by nobody.

Keeping those two apart also removes an index-space hazard: `afterSetSourceDataAtCell` reports **physical** rows while `afterSetDataAtCell` reports **visual** ones, so no address is built from the source-data hook at all. Feeding a physical row to `getHfIndexFromVisualIndex()` produced an address pointing at a different row as soon as rows were sorted or moved, which silently suppressed whichever dependent cell it collided with. Out-of-range and trimmed indexes (`-1`) are dropped rather than kept, since they can only ever collide.

A note covering both traps was added to `handsontable/AGENTS.md`.

### Known defects found while testing, not fixed here

Both predate this PR and both need their own task.

1. **`validateDependentCells()` treats HyperFormula indexes as visual indexes.** It bounds-checks `row >= hot.countRows()` and then calls `getDataAtCell(row, col)` / `getCellMeta(row, col)` on the raw HF index, without `getVisualIndexFromHfIndex()`. HyperFormula is fed trimmed rows as well, so with `trimRows` or Filters active the two index spaces diverge. Measured with 6 rows, `trimRows: [1, 2]` and a formula on physical row 3 (visual row 1): a plain `setDataAtCell` with **no undo involved** validates visual row 3 instead of visual row 1. Fixing it changes which cells get validated on every edit in every trimmed or filtered grid.

2. **Redoing a removal does not revalidate.** `RemoveRowAction.redo` calls `alter('remove_row', …)`, which writes no cell data, so nothing revalidates. Remove a row a formula depends on, undo (the formula goes valid again — this PR fixes that), then redo: the formula reads `#REF!` but keeps `valid: true`. This matches plain removals outside undo — `#onAfterRemoveRow` only calls `renderDependentSheets()`, never `validateDependentCells()` — so it is a pre-existing gap rather than something this PR introduces.

### How has this been tested?

Seven tests in `handsontable/src/plugins/formulas/__tests__/validation.spec.js`, each confirmed red against the code it guards:

- `should revalidate dependant cells after undoing and redoing a change (#dev-2036)` — the reported scenario, through undo, redo, and undo again.
- `should only validate the dependant cells once when undoing a change (#dev-2036)` — no double validation on the `setDataAtCell` path.
- `should validate the cells restored by undoing a row removal (#dev-2036)` — the restored formula is validated, not excluded.
- `should revalidate dependant cells after undoing a row removal on sorted rows (#dev-2036)` — physical vs visual rows; fails with `Expected [ '1x0', '1x1' ] to contain '2x1'` before the fix.
- `should only validate each cell once when undoing a row removal (#dev-2036)` — no double validation on the `setSourceDataAtCell` path.
- `should revalidate dependant cells after undoing a row removal (#dev-2036)` — the removal case end to end; also asserts that undoing the *edit* first correctly leaves the cell red, because it is still a `#REF!`.
- `should not validate dependant cells after undoing an action that writes no data (#dev-2036)` — covers the gate, using `columnSorting`.

```bash
npm run test:e2e --prefix handsontable -- --testPathPattern=undoRedo
npm run test:e2e --prefix handsontable          # 9613 specs, 0 failures
npm run test:unit --prefix handsontable         # 3267 tests, 0 failures
npm run eslint --prefix handsontable            # 0 errors
```

Manual checks in Chrome against a local `handsontable.full.js` build.

Reported scenario, `type: 'numeric'`, `=A1+B1+C1+D1+E1` in column F:

| Step | F1 value | `meta.valid` | `htInvalid` |
|---|---|---|---|
| type `aaa` in D1 | `#VALUE!` | `false` | yes |
| undo | 150 | `true` | no |
| redo | `#VALUE!` | `false` | yes |

Removal case, `data: [[1, 10], ['=A1+B2', 0]]`, `type: 'numeric'`:

| Step | formula cell | `meta.valid` | `htInvalid` |
|---|---|---|---|
| `alter('remove_row', 0)` then `setDataAtCell(0, 1, 5)` | `#REF!` | `false` | yes |
| undo the edit | `#REF!` | `false` | yes — correct, still an error |
| undo the removal | `1` | `true` | no |

Sorted rows, with the plugin's validation pass instrumented: the engine reports dependants `1x0, 1x1, 2x1`, the exclusion list is now empty, and all three are validated. Before the fix the exclusion list was `2x0, 2x1` and the dependant at `2x1` was dropped.

No console errors.

### Types of changes

### Related issue(s):

1. DEV-2036

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2036


qunabu:
Task linked: [DEV-2036 htInvalid stays in the formulas cell even when the value is correct](https://app.clickup.com/t/9015210959/DEV-2036)

adrianspdev:
Two things I'd like checked before this merges.

Issue 1 — does undo of a row/column removal need the same treatment?

#validateUndoRedoDependentCells (formulas.ts:839) drops dependentCells for anything that isn't actionType === 'change'. But RemoveRowAction.undo ends with hot.setSourceDataAtCell(changes, undefined, undefined, 'UndoRedo.undo') (undoRedo/actions/removeRow.ts:152; removeColumn.ts:214 is the same), which is a real data write — and from any other source that write validates dependents via #onAfterSetSourceDataAtCell (formulas.ts:1438).

Worth running:

```
// data: [[1, 10], ['=A1+B2', 0]], numeric column
alter('remove_row', 0);       // formula becomes =#REF!+B1
setDataAtCell(0, 1, 5);       // dependent formula cell gets valid: false + the invalid class
undo();                       // actionType 'change' — still #REF!, still invalid. Expected.
undo();                       // removal undone: formula is =A1+B2 and computes a valid number.
                              // Does the cell still render red?
```

If it does, the gate is too narrow. The AGENTS.md:68 rationale compares against alter('remove_row'), which genuinely doesn't validate dependents — but undoing a removal is an insert plus a data write, so I don't think that symmetry carries over.

Issue 2 — are the addresses reaching validateDependentCells visual or HyperFormula indexes?

validateDependentCells bounds-checks row >= this.hot.countRows() (757) and then calls boundHot.getDataAtCell(row, col) / getCellMeta(row, col) (775-777) on the same numbers, without going through getVisualIndexFromHfIndex (indexSyncer/axisSyncer.ts:204-213) — which this file does use at formulas.ts:187-188. Since #getTranslationCache builds hfIndexOfPhysical from the full getIndexesSequence() while visual order comes from getNotTrimmedIndexes(), the two should diverge by the trimmed count.

Worth running: a grid with Filters (or trimRows) hiding a few rows, a formula below the hidden ones depending on an edited cell, then undo. My reading says valid lands on the wrong row and the real dependent keeps its stale mark, with rows past countRows() dropped by the guard instead.

This predates the PR — the old beforeUndo threw away the engine's return value, so nothing reached it from this direction. #validateUndoRedoDependentCells (832-842) is the first caller to pass raw engine.undo() output through, so it'd now fire on every undo. Reordering alone should be fine (setRowOrder/setColumnOrder keep HF and visual in sync); trimming is what I'd expect to break it.

budnix:
> Two things I'd like checked before this merges.
> 
> ...

@adrianspdev Thanks! Both fixed.

adrianspdev (approved):

qunabu (approved):

PR #13184, opened 2026-08-12
DEV-1782: Fix column moving blocked by the sorting label

budnix:
### Context

The PR fixes a regression reported in DEV-1782: with `columnSorting` enabled, a column could not
realistically be moved by dragging its header.

Two things caused it, and both are fixed here.

**1. There was nowhere left to start a drag.** `ManualColumnMove` refuses any press whose target
carries `sortAction` — the class on the header label — because that press belongs to sorting. Since
17.1.0 the header became a flex row and the label was given `flex: 1 1 auto`, so the label covers the
whole header cell. The draggable strip collapsed to a flat 17px that no longer grows with the column,
down from 94px on a 120px column in 17.0.1. Pressing the header centre sorted instead of moving.

**2. Selecting a column always sorted it.** A column has to be selected before it can be dragged, and
every way of selecting it (a header press) also sorted it — so moving a column re-sorted it on the way.

Measured on a 120px column, sorted and selected:

| version | draggable strip |
|---|---|
| 17.0.1 | 94px |
| 17.1.0 / 18.0.0 | 17px |

### What changed

**Sorting resolves on release, not on press.** Both plugins split the gesture by whether the pointer
travelled: a press that stays put sorts, a press that travels moves the column.

- Travel is measured from `mousemove`, not by comparing the press and release positions. Pressing a
  partly visible header scrolls it into view, which moves the header under a stationary pointer — that
  is not a drag, and comparing down/up coordinates misreads it as one.
- The release is resolved from `afterOnCellMouseUp`, with a document `mouseup` fallback for releases
  outside a cell. Walkontable calls its own `onMouseUp` directly from `touchend` instead of dispatching
  a DOM event, so a document listener alone would have killed tap-to-sort on touch devices.
- Whether an open editor needs validating is captured at press time, as before. Reading it on release
  is wrong: selecting a column opens an editor on the highlighted cell, so any validated column would
  wait for a validation that never runs.
- `ManualColumnMove` drops the `sortAction` guard and gates the move on a new `#dragged` flag.

**Only the label and the sort indicator sort on click.** The label is sized to its content and centred
instead of filling the header cell, so the header around it selects the column without sorting it. On a
140px column that area goes from 8px to 47px. `position: static` on the label hands the sort indicator
and the multi-sort order badge to the header container, so they stay pinned to the header's trailing
edge; clicking one still sorts, because an event on a pseudo-element targets the element that owns it.

**This is not a visual change.** Measured against released 18.0.0 across widths 40–500px, long labels,
RTL, `htLeft`, `htRight`, both of those in RTL, `dropdownMenu`, `multiColumnSorting`, nested and
collapsible headers, `indicator: false`, and GhostTable auto width: header text position, sort indicator
position, indicator vertical centring and auto column width are all unchanged.

### How has this been tested?

New Playwright spec — `tests/e2e/column-move-sorting.spec.ts`, with fixture
`tests/fixtures/demo/column-move-sorting.html` and page object
`tests/fixtures/pages/ColumnMoveSortingPage.ts`:

1. moves a sorted column when the drag starts at the header centre — the regression;
2. keeps the header label clickable for sorting (asc then desc);
3. selects the column without sorting when the bare header area is pressed, then drags it;
4. moves the column when the drag starts on the sorting label itself, without toggling the sort;
5. does not move the column when a press is released without moving — it sorts instead.

```bash
cd tests && npx playwright test e2e/column-move-sorting.spec.ts
npm run test:e2e --prefix handsontable -- --testPathPattern="columnSorting|multiColumnSorting|manualColumnMove|dropdownMenu|nestedHeaders|collapsibleColumns|ghostTable|autoColumnSize"
```

| suite | result |
|---|---|
| new Playwright spec, 3 themes | 15 passed |
| header-related E2E | 1041 specs, 0 failures |
| full core E2E | 9606 specs, 0 failures |
| unit (Jest) | 3267 passed |
| touch emulation (iPhone 13, iPad gen 7) | tap-to-sort works, same as released |
| eslint / stylelint / type check | clean |

Test sensitivity was verified rather than assumed — each half of the change was reverted and rebuilt
on its own. With the plugin sources reverted, the two drag tests fail; with the CSS reverted, the
bare-header test fails. The other two pass either way by design: they guard existing behaviour rather
than carrying the regression.

Also verified by hand-driven browser measurement: the sort indicator's position, the header text
position, the indicator's vertical centring and GhostTable auto column width all match released 18.0.0
in every configuration listed above.

### Notes for the reviewer

- **Three legacy specs were updated, not deleted.** `manualColumnMoveUI`'s *"should not run moving ui
  if mousedown was fired on sorting element"* asserted the very guard this change removes; it now
  asserts the replacement contract. The `columnSorting` indicator specs (LTR and RTL) asserted that the
  indicator's containing block is the label span, which is the detail this change moves — they now
  assert the container it moved to, plus a new check that the indicator lands on the header's vertical
  midline, which is the part a user can see.
- **`:has()` is banned in this package**, so two markers are toggled from JS instead:
  `has-header-button` (set by `DropdownMenu`, so the indicator steps over the menu button) and
  `has-sort-indicator` (set by `ColumnSorting`, reserving the indicator's room on the container rather
  than as padding on the label, where it would enlarge the area that sorts on click). Both are
  re-applied on every render, because `TableView` rebuilds that container's class list while its
  children survive.
- **A side effect worth knowing:** a plain click on a selected header used to fire `beforeColumnMove`
  and `afterColumnMove` with nothing actually moving. It no longer does.
- **Known rough edge:** a press that jitters 1–3px sorts (inside the click tolerance) *and* fires a
  no-op `afterColumnMove`, because `ManualColumnMove` treats any pointer move as the start of a drag
  while `ColumnSorting` applies a 3px tolerance. Aligning both would mean rewriting roughly 30 legacy
  specs that drive drags by dispatching `mousemove` on the same element (zero measurable travel), so it
  was left as-is. It is the same class of noise that already exists today.

### Types of changes

### Related issue(s):

1. DEV-1782

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

The column-moving and rows-sorting guides are updated in this PR.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-1782

---

> [!NOTE]
> **Medium Risk**
> Changes shared header pointer handling between ColumnSorting and ManualColumnMove and alters when move hooks fire; scope is localized but regression-prone without the added E2E and updated plugin tests.
> 
> **Overview**
> Fixes **DEV-1782**: with `columnSorting` and `manualColumnMove` together, dragging a column header often sorted instead of moving because the sort label filled the header and `ManualColumnMove` ignored presses on `sortAction`.
> 
> **Gesture split:** `ColumnSorting` queues a sort on header label press and applies it on **mouse up** (plus `afterOnCellMouseUp` and document `mouseup` for touch/outside-cell release). If `ManualColumnMove.isDragging()` is true, the sort is skipped. `MultiColumnSorting` overrides `applyHeaderClickSort` for Ctrl/Cmd append behavior without duplicating that logic.
> 
> **Column move:** `ManualColumnMove` no longer blocks mousedown on the sort label; it sets `#dragged` when pointer travel exceeds 3px and only runs `dragColumns` / `beforeColumnMove` / `afterColumnMove` on a real drag. Plain header clicks no longer fire those move hooks.
> 
> **Header UI:** The sort label is sized to its text (not full header width) so bare header area can select without sorting. Sort indicator room is reserved on the header `.relative` container (`has-sort-indicator`); `DropdownMenu` marks `has-header-button` so the indicator clears the menu button. SCSS and unit tests follow the container-based indicator layout.
> 
> Docs for column moving and row sorting are updated. New Playwright coverage in `tests/e2e/column-move-sorting.spec.ts`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 004e94f22f619584f066cd2ac02e02ecf808d0b7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changelogs/13184.json
+++ b/.changelogs/13184.json
@@ -1,6 +1,6 @@
 {
   "issuesOrigin": "private",
-  "title": "Fixed being unable to move a column by dragging its header while column sorting is enabled. Sorting by clicking a header now happens on mouse up instead of mouse down, and only the header label and its sort indicator sort on click.",
+  "title": "Fixed being unable to move a column by dragging its header while column sorting is enabled. Sorting by clicking a header now happens on mouse up instead of mouse down, and only the header label and its sort indicator sort on click. A click on a column header no longer fires `beforeColumnMove` and `afterColumnMove`, which previously ran even though no column moved.",
   "type": "fixed",
   "issueOrPR": 13184,
   "breaking": false,
--- a/.changelogs/13184.json
+++ b/.changelogs/13184.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Fixed being unable to move a column by dragging its header while column sorting is enabled. Sorting by clicking a header now happens on mouse up instead of mouse down, and only the header label and its sort indicator sort on click.",
+  "type": "fixed",
+  "issueOrPR": 13184,
+  "breaking": true,
+  "framework": "none"
+}
--- a/.changelogs/13184.json
+++ b/.changelogs/13184.json
@@ -3,6 +3,6 @@
   "title": "Fixed being unable to move a column by dragging its header while column sorting is enabled. Sorting by clicking a header now happens on mouse up instead of mouse down, and only the header label and its sort indicator sort on click.",
   "type": "fixed",
   "issueOrPR": 13184,
-  "breaking": true,
+  "breaking": false,
   "framework": "none"
 }
--- a/handsontable/src/plugins/columnSorting/columnSorting.ts
+++ b/handsontable/src/plugins/columnSorting/columnSorting.ts
@@ -74,6 +74,8 @@ export interface HeaderSortPress {
    * when nothing was being edited.
    */
   validation: Promise<void> | null;
+  /** Set when ManualColumnMove ran its move before this press was resolved. */
+  consumedByMove: boolean;
 }
 
 export interface SortConfig {
@@ -248,6 +250,7 @@ export class ColumnSorting extends BasePlugin {
     // sort still has to be cleared rather than left pending for the next release.
     this.eventManager.addEventListener(this.hot.rootDocument.documentElement, 'mouseup',
       () => this.#resolvePendingSort());
+    this.addHook('afterColumnMove', this.#onAfterColumnMove);
 
     this.addHook('afterInit', this.#loadOrSortBySettings);
     this.addHook('afterLoadData', this.#onAfterLoadData);
@@ -986,6 +989,7 @@ export class ColumnSorting extends BasePlugin {
       validation: awaitsValidation ? new Promise<void>((resolve) => {
         this.hot.addHookOnce('postAfterValidate', () => resolve());
       }) : null,
+      consumedByMove: false,
     };
   }
 
@@ -1018,12 +1022,11 @@ export class ColumnSorting extends BasePlugin {
 
     this.#pendingHeaderSort = null;
 
-    // Ask ManualColumnMove whether it turned this press into a drag, rather than measuring pointer
-    // travel here. Only that plugin knows whether a drag was ever armed - it needs the column
-    // selected by its header - so a travel test here would silence the sort on every grid where
-    // nothing can consume the gesture. Its state is still set at this point; it resets on its own
-    // `mouseup` listener, which runs later in the same dispatch.
-    if (this.#isColumnBeingDragged()) {
+    // Two signals, because both plugins handle the same `mouseup` and the listener order is not
+    // fixed - re-enabling `columnSorting` after `manualColumnMove` appends this plugin's listener
+    // last. Running first, the drag is still in progress; running second, the move has already
+    // fired. Either one cancels the sort.
+    if (pending.consumedByMove || this.#isColumnBeingDragged()) {
       return;
     }
 
@@ -1036,6 +1039,15 @@ export class ColumnSorting extends BasePlugin {
     }
 
     void pending.validation.then(() => this.applyHeaderClickSort(pending));
+  };
+
+  /**
+   * Marks a queued press as spent once ManualColumnMove has moved the column.
+   */
+  #onAfterColumnMove = () => {
+    if (this.#pendingHeaderSort !== null) {
+      this.#pendingHeaderSort.consumedByMove = true;
+    }
   };
 
   /**
--- a/handsontable/src/plugins/columnSorting/columnSorting.ts
+++ b/handsontable/src/plugins/columnSorting/columnSorting.ts
@@ -54,32 +54,12 @@ export const APPEND_COLUMN_CONFIG_STRATEGY = 'append';
 export const REPLACE_COLUMN_CONFIG_STRATEGY = 'replace';
 const SHORTCUTS_GROUP = PLUGIN_KEY;
 /**
- * How far, in pixels, the pointer may travel between pressing a column header and releasing it
- * and still count as a click. Pointers jitter by a pixel or two on trackpads and high-density
- * screens, so an exact-zero test would swallow ordinary clicks and header clicks would appear
- * to do nothing.
- */
-const POINTER_DRAG_TOLERANCE = 3;
-/**
  * Marks the header container of a column that is showing a sort indicator. The indicator is
  * positioned against that container, so the room it needs is reserved there rather than as padding
  * on the label - padding on the label would enlarge the area that sorts on click, which is exactly
  * what it must not do. A class rather than a `:has()` selector, which is banned in this package.
  */
 const CONTAINER_WITH_INDICATOR_CLASS = 'has-sort-indicator';
-
-/**
- * Checks whether the pointer has travelled far enough from where it was pressed to count as a
- * drag rather than a click.
- *
- * @param {{ x: number, y: number }} pressOrigin Pointer client coordinates captured on press.
- * @param {MouseEvent} event The event holding the current pointer position.
- * @returns {boolean}
- */
-function isPointerDragged(pressOrigin: { x: number, y: number }, event: MouseEvent): boolean {
-  return Math.abs(event.clientX - pressOrigin.x) > POINTER_DRAG_TOLERANCE ||
-    Math.abs(event.clientY - pressOrigin.y) > POINTER_DRAG_TOLERANCE;
-}
 
 registerRootComparator(PLUGIN_KEY, rootComparator);
 
@@ -89,13 +69,11 @@ registerRootComparator(PLUGIN_KEY, rootComparator);
 export interface HeaderSortPress {
   column: number;
   isCtrlPressed: boolean;
-  pressOrigin: { x: number, y: number };
   /**
    * Settles once the cell that was being edited at press time has finished validating, or `null`
    * when nothing was being edited.
    */
   validation: Promise<void> | null;
-  hasMoved: boolean;
 }
 
 export interface SortConfig {
@@ -193,8 +171,7 @@ export class ColumnSorting extends BasePlugin {
    */
   columnMetaCache: IndexToValueMap | null = null;
   /**
-   * Sort queued by a header press, applied on mouse up when the pointer has not travelled.
-   * Holds the pressed column, whether Ctrl/Cmd was held, and where the pointer went down.
+   * Sort queued by a header press, applied on mouse up unless the press became a column drag.
    */
   #pendingHeaderSort: HeaderSortPress | null = null;
   /**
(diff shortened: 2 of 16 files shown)


qunabu:
Task linked: [DEV-1782 [17.1.0] Manual column move is hard to proceed as sorting action takes more space in the column header](https://app.clickup.com/t/9015210959/DEV-1782)

handsontable/src/plugins/columnSorting/columnSorting.ts
```diff
@@ -213,6 +263,17 @@ export class ColumnSorting extends BasePlugin {
     this.addHook('beforeOnCellMouseDown', this.#onBeforeOnCellMouseDown);
     this.addHook('afterOnCellMouseDown',
       (event: Event, target: { row: number, col: number }) => this.onAfterOnCellMouseDown(event, target));
+    // Primary release signal. On touch devices Walkontable calls its `onMouseUp` directly from
+    // `touchend` instead of dispatching a DOM `mouseup`, so a document listener alone would never
+    // fire and tapping a header would stop sorting.
+    this.addHook('afterOnCellMouseUp', this.#resolvePendingSort);
+    // Fallback for a release that lands outside any cell - that is the drag case, and the queued
+    // sort still has to be cleared rather than left pending for the next release.
+    this.eventManager.addEventListener(this.hot.rootDocument.documentElement, 'mousemove',
```
demtario: **The jitter gate also fires on grids where nothing can consume the drag.**

This listener is registered in `enablePlugin` unconditionally — `manualColumnMove` does not have to be enabled. So on a plain `columnSorting: true` grid, a header press that travels 4px now sorts *nothing*. Before this PR it sorted on mouse down.

The part that makes it user-visible: Chrome still dispatches a real `click` for that gesture (it cancels the click at roughly 5px), so there is a band where the platform says "click" and the grid says "drag" and the result is silence. `POINTER_DRAG_TOLERANCE = 3` is tighter than the platform's own slop.

Either of these closes it:

1. raise the tolerance to match the platform (~5px), or
2. only suppress the sort when a move is actually armed — `manualColumnMove` enabled and the column selected.

Worth noting that nothing currently covers this configuration: `tests/fixtures/demo/column-move-sorting.html` always enables both plugins, so the sorting-only case is untested in the new Playwright spec.

handsontable/src/plugins/manualColumnMove/manualColumnMove.ts
```diff
@@ -16,6 +16,9 @@ const CSS_PLUGIN = 'ht__manualColumnMove';
 const CSS_SHOW_UI = 'show-ui';
 const CSS_ON_MOVING = 'on-moving--columns';
 const CSS_AFTER_SELECTION = 'after-selection--columns';
+// Matches ColumnSorting's tolerance. Kept as a local constant rather than imported - plugins do not
+// import each other.
+const POINTER_DRAG_TOLERANCE = 3;
```
demtario, line 21: **The 3px contract is duplicated, and the two copies have to agree forever.**

`columnSorting.ts` declares the same `POINTER_DRAG_TOLERANCE = 3` plus an `isPointerDragged()` helper; `#onMouseMove` below re-inlines the same `Math.abs` pair by hand. If the two ever drift, a gesture falls between the plugins — the sort is suppressed and the move never starts, so the header press does nothing at all.

The comment is right that plugins should not import each other, but a shared helper is not a plugin-to-plugin import. `helpers/dom/event.ts` already exists and this file already imports `offsetRelativeTo` from it — that is the natural home for both the constant and `isPointerDragged`, and it turns "these must stay in sync" into something the compiler enforces.

.changelogs/13184.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Fixed being unable to move a column by dragging its header while column sorting is enabled. Sorting by clicking a header now happens on mouse up instead of mouse down, and only the header label and its sort indicator sort on click.",
```
demtario: **The dropped move hooks are not documented anywhere.**

The PR description flags it as "a side effect worth knowing": a plain click on a selected header no longer fires `beforeColumnMove` / `afterColumnMove` (`manualColumnMove.ts`, the `!wasDragged` bail in `#onMouseUp`). That is a hook-contract change that any consumer with a listener on either hook will observe.

It appears in neither this entry — the title covers only the mouse-up timing and the label-only sort target — nor in the two updated guides. Per the mandatory checklist ("any user-facing behavior change is documented"), it should be in at least one of them. Cheapest fix is a clause on the changelog title plus a line in the column-moving guide.

The change itself looks right, incidentally — firing move hooks for a move that never happened was the actual bug.

demtario (commented):
Code review of the plugin/CSS change. I verified the geometry claim independently — `th { padding: 0 }`, `.relative` carries the cell padding, and absolutely positioned pseudo-elements resolve against the *padding* box — so `padding-inline-end` reserves room for the label without moving `::before`/`::after`. The "not a visual change" claim holds, and the press-versus-release split is the right shape.

Three comments below, all inline. I also found a handful of low-severity items (RTL ghost-table specificity on `--ht-header-button-slot`, the untested multi-sort order badge, a redundant assertion in `manualColumnMoveUI.spec.js`, and the now-stale "known rough edge" paragraph in the PR description) — happy to add those if useful, but they are not worth blocking on.

handsontable/src/styles/components/plugins/_column-sorting.scss
```diff
@@ -50,6 +96,15 @@
...
   [dir="rtl"] {
     .handsontable {
+      // In RTL the indicator sits on the inline-end side for both default and right-aligned text,
+      // and only `htLeft` flips it back - so `htRight` has to undo the direction-agnostic rule
+      // above, which would otherwise reserve the wrong side.
+      thead tr th.htRight .relative.has-sort-indicator,
+      thead tr th.htLeft .relative.has-sort-indicator {
+        padding-inline-start: var(--ht-cell-horizontal-padding);
+        padding-inline-end: calc(var(--ht-cell-horizontal-padding) + var(--ht-icon-size, 16px) + 2px);
+      }
```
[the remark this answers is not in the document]
budnix: Fixed in e1f7771402.

Reproduced: in RTL a \`htLeft\` header pins the indicator to the physical right, but the reservation was grouped with \`htRight\` and reserved the other side — a 16px overlap on the label. Split into two rules with \`htLeft\` mirrored. Regression test added to the RTL \`columnSorting\` spec, covering both \`dir=rtl\` and \`layoutDirection: rtl\`.

handsontable/src/styles/components/plugins/_column-sorting.scss
```diff
@@ -90,13 +141,25 @@
     }
   }
 
+  /* In the ghost table the indicator pseudo is in flow, so the offsets that hold it clear of the
+      header edge in the real table would be measured as extra column width and every auto-sized
+      column would come out wider. The selector is deliberately more specific than the rule that
+      sets those offsets. */
+  .htGhostTable .htCore .colHeader.columnSorting.sortAction {
```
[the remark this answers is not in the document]
budnix: Fixed in 490e75b61b.

Reproduced: with \`dropdownMenu\` + \`columnSorting\`, auto column width grew 46px against released. The reset now overrides the custom properties on the container instead of competing with the rules that read them, so there is no specificity race left to lose. Regression test asserts the dropdown button adds only a few pixels; it fails at 24px without the fix.

handsontable/src/plugins/columnSorting/columnSorting.ts
budnix: Fixed in 004e94f22f.

Reproduced, and wider than described. Measured against released across three configurations, the dead band is not 3–5px — it is any travel over 3px:

| config | 4px | 6px | 10px |
|---|---|---|---|
| `columnSorting` only | nothing | nothing | nothing |
| sorting + move, header not yet selected | nothing | nothing | nothing |
| sorting + move, header selected | moves | moves | moves |

Chrome dispatched a real `click` in every one of those, including at 10px. So raising the tolerance to ~5px would not have closed it, and the second row is the first press on a header — the flow from the original report.

`ColumnSorting` no longer measures pointer travel at all. It asks `ManualColumnMove` through a new `isDragging()`; when that plugin is absent, disabled, or was never armed, the press sorts as it always did.

Worth recording one dead end: having the move plugin mark the press via `afterColumnMove` on a microtask does not work, because microtasks drain between individual listeners — `afterColumnSort` fires before `afterColumnMove`. Deferring a whole task fixed the order but broke 51 specs that assert synchronously after a click. Asking the plugin directly is synchronous.

Coverage: the fixture now takes `?move=off`, and a new Playwright test drags 20px on a sorting-only grid and asserts it still sorts.

handsontable/src/plugins/manualColumnMove/manualColumnMove.ts
budnix: Fixed in 004e94f22f — by removing the duplicate rather than sharing it.

One caveat on the suggested home: `helpers/dom/event.ts` is spread into `Handsontable.helper` (`src/index.ts:113`), so putting `isPointerDragged` there would add permanent public API.

Since `ColumnSorting` no longer measures travel (see the other thread), only one copy of the constant is left, in `ManualColumnMove`, where the drag lives. There is nothing to keep in sync now.

.changelogs/13184.json
budnix: Fixed in 004e94f22f.

Changelog title now ends: "A click on a column header no longer fires `beforeColumnMove` and `afterColumnMove`, which previously ran even though no column moved." Added a matching line to the column-moving guide next to the `beforeColumnMove` section.

demtario (approved):

Issue #16, opened 2012-06-12
Autocomplete issue when using tab key


jglimbourg:
In "Autocomplete" example, when the user want to select an item (for example in column "Car")  :
- when he presses "Enter", the item is selected and copied in the input cell : OK, but the focus goes to cell below.
- when he presses "Tab", the item is copied in the NEXT input instead of the current input : BUG ?

Issue #14, opened 2012-06-06
onChange not working now on Mac with Safari version 5.1.7


harrygates:
The latest updates seem to have broken onChange, as it is not firing at all.

PR #13190, opened 2026-08-13
DEV-2210: Disconnect the hidden-init visibility observer on destroy

demtario:
### Context

The PR fixes an uncaught `TypeError: Cannot read properties of null (reading 'view')` thrown from library code after a grid that was initialized in a hidden container is destroyed.

When `rootElement.offsetParent === null` at first init, core registers `observeVisibilityChangeOnce(instance.rootElement, …)` so it can re-measure and re-render once the grid becomes visible. Three things combined into a crash:

1. The callback dereferenced the closure variable `instance` with no guard (unlike `syncEdgeSlotsWidth` a few lines above, which checks `isDestroyed`).
2. `destroy()` never disconnected that observer — it only set `instance = null`.
3. `observeVisibilityChangeOnce` returned nothing, so callers *could not* disconnect it. It self-disconnected only after a delivery with `isIntersecting: true`.

An `IntersectionObserver` entry carries the state from the moment its snapshot was taken, so a snapshot computed while the element was visible can be delivered **after** `destroy()` — and the callback then read `instance.view` on `null`. If a hidden grid is destroyed before ever becoming visible, the observer leaks instead.

Impact in CI: `handsontable/test/scripts/run-puppeteer.mjs` treats any `pageerror` as fatal, so one late delivery aborted the whole Puppeteer E2E suite at ~380 of 9,605 specs (`Develop` runs #14 and #16, `Integration / ESM`). The trigger there is the four `Core_init` "should rerender the table after changing the `display` property…" specs, which show the container, wait two animation frames, and end — on a loaded runner the delivery starves past that and lands after `afterEach → destroy()`.

The fix:

- `observeVisibilityChangeOnce` now returns the `IntersectionObserver` (additive, non-breaking; the helper is public as `Handsontable.dom.observeVisibilityChangeOnce`).
- Core keeps the handle in the `Core` closure and disconnects it at the top of `destroy()`. It is deliberately **not** an instance property: `destroy()` runs `objectEach(instance, …)`, which nulls every non-function instance property before the closure `instance = null`, so a property handle would be gone before the cleanup line could read it. This also fixes the leak when a hidden grid is destroyed while still invisible.
- The callback opens with `if (!instance || instance.isDestroyed || !instance.view) { return; }` — the same shape as `syncEdgeSlotsWidth`. This guard is load-bearing, not cosmetic: it does not rely on `disconnect()` draining an already-queued delivery in every engine.
- `test/__mocks__/intersectionObserverMock.js` gains a `disconnect()` no-op. jsdom reports every element as invisible (`offsetParent` is always `null`), so *every* grid built in a Jest unit test takes the hidden-init branch; without it, `destroy()` would throw across the whole unit suite.

Two pitfalls learned here were added to `handsontable/AGENTS.md`: teardown handles belong in the `Core` closure rather than on the instance, and jsdom's blanket invisibility means the shared observer mocks must cover whatever the hidden-init path calls.

### How has this been tested?

New Playwright regression (`tests/e2e/hidden-init-destroy.spec.ts` + `fixtures/demo/hidden-init.html` + `fixtures/pages/HiddenInitPage.ts`) reproduces the exact ordering from the report — init hidden → show → `requestAnimationFrame` → `setTimeout(…, 0)` → `destroy()` — and asserts no `pageerror` reached the page. It was confirmed **red** against a pre-fix build with exactly `TypeError: Cannot read properties of null (reading 'view')`, then green after the fix. It also asserts the grid really did initialize while invisible, so the scenario cannot silently degrade to "destroy a visible grid" and still pass. Note this is the first spec in the `tests/` tier to tear an instance down — the tier's usual isolation is per-navigation, but the destroy path is the whole subject here.

New unit block in `handsontable/src/helpers/dom/__tests__/element.unit.ts` covers the returned handle: the helper returns the live observer and observes the element, a visible delivery calls the callback once and self-disconnects, a non-intersecting delivery does neither, and disconnecting the returned handle stops the callback from running.

```bash

# new regression + the full functional matrix (6 legs: theme × bundle)
cd tests && npx playwright test                       # 264 passed
cd tests && npx playwright test hidden-init-destroy   # the new spec alone

# unit (full run matters here — the observer mock affects every grid-building test)
npm run test:unit --prefix handsontable               # 3271 passed / 277 suites
npm run test:unit --prefix handsontable -- --testPathPattern=element.unit

# legacy Jasmine suites around the change
npm run test:e2e --prefix handsontable -- --testPathPattern=core/init      # 15 specs, 0 failures
npm run test:e2e --prefix handsontable -- --testPathPattern=helpers/dom    # 24 specs, 0 failures
npm run test:walkontable --prefix handsontable                             # 816 specs, 0 failures

```

Out of scope: the four `Core_init` specs still wait a fixed two animation frames for the delivery. After this fix, starvation no longer aborts the suite — it would degrade to a plain assertion failure on those specs. Making them wait on the width condition instead of a frame count is a separate test-timing cleanup, as are the other recent `Integration / ESM` flakes (the ResizeObserver warning spec, `afterRefreshDimensions` in an iframe, and the AutocompleteEditor html-mode spec).

### Types of changes

### Related issue(s):

1. DEV-2210

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2210

---

> [!NOTE]
> **Medium Risk**
> Touches core init/destroy and a public DOM helper return type, but changes are defensive teardown and additive API; risk is moderate due to lifecycle timing across all jsdom-built grids.
> 
> **Overview**
> Fixes a crash when a grid initialized in a **hidden** container is destroyed as it becomes visible: a late `IntersectionObserver` delivery could run the visibility callback after `destroy()` nulled the core closure and throw on `instance.view`.
> 
> **Core** stores the observer handle in the closure (not on the instance), **disconnects** it at the start of `destroy()`, and **guards** the callback with `!instance || instance.isDestroyed || !instance.view`. **`observeVisibilityChangeOnce`** now **returns** the observer so callers can disconnect early.
> 
> Adds a Playwright regression (hidden init → show → destroy race), unit tests for the helper, `disconnect()` on shared **IntersectionObserver** mocks (core + React/Vue wrappers), changelog **13190**, and **AGENTS.md** notes on teardown handles and jsdom’s always-hidden elements.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3d1ab63dcd1b2e5c12f85ac9b6ba296a61565f08. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2210 Disconnect the hidden-init visibility observer on destroy and guard its callback](https://app.clickup.com/t/9015210959/DEV-2210)

adrianspdev (approved):

PR #11997, opened 2026-01-23
Add support for Intl.NumericFormat options

budnix:
### Context

The PR introduces support for the native JavaScript [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) API in the `numericFormat` cell option. The `numbro.js`-based formatting (using `pattern` and `culture` properties) is now deprecated and will be removed in the next major release. When numbro-style format is detected, a deprecation warning is shown in the console. The locale for `Intl.NumberFormat` is controlled via the existing `locale` cell property, making the configuration more consistent with other Handsontable internationalization settings.

Additionally, this PR introduces a new `valueFormatter` option that allows custom formatting of cell values before display. Unlike the `renderer` option, which handles the complete cell rendering process (DOM structure, attributes, styles), the `valueFormatter` focuses solely on transforming the cell's value. This separation simplifies common formatting use cases where only the displayed value needs to be transformed, such as adding units, formatting dates, or applying custom text transformations.

The rendering pipeline was also refactored so that individual renderers no longer need to explicitly call `baseRenderer`. Instead, the rendering engine in `tableView.js` automatically invokes `baseRenderer` if it wasn't already called during the renderer execution, ensuring consistent base behavior across all renderers.

## Migration from numbro to Intl.NumberFormat

The changes are backward-compatible, but in the next major releases, support for numbro-based options will be removed. For developers who want to migrate to the new pattern today, this may be useful.

### Basic number formatting

| Format type | numbro (deprecated) | Intl.NumberFormat |
|-------------|---------------------|-------------------|
| No formatting | `{ pattern: '0' }` | `{}` or `{ useGrouping: false, maximumFractionDigits: 20 }` |
| Thousands separator | `{ pattern: '0,0' }` | `{ useGrouping: true }` |
| Two decimal places | `{ pattern: '0.00' }` | `{ minimumFractionDigits: 2, maximumFractionDigits: 2 }` |
| Thousands + decimals | `{ pattern: '0,0.00' }` | `{ useGrouping: true, minimumFractionDigits: 2, maximumFractionDigits: 2 }` |

### Currency formatting

| Format type | numbro (deprecated) | Intl.NumberFormat |
|-------------|---------------------|-------------------|
| USD currency | `{ pattern: '$0,0.00', culture: 'en-US' }` | `locale: 'en-US', numericFormat: { style: 'currency', currency: 'USD' }` |
[...]

PR #13191, opened 2026-08-13
DEV-2126: Fix autoRowSize and autoColumnSize ignoring per-cell className

adrianspdev:
### Context

The PR fixes `autoRowSize` and `autoColumnSize` ignoring any styling a cell receives through its `className`, which leaves the row headers misaligned with the data cells.

**Root cause.** #11997 (v17.0.0) removed the `baseRenderer.apply(...)` chain call from the built-in renderers and moved base-renderer invocation into `TableView.cellRenderer`, guarded by `cellProperties._isBaseRendererCalled`. `GhostTable` — the hidden table both auto-size plugins measure — calls `hot.getCellRenderer(cellProperties)` directly and so bypasses `TableView` entirely. Since 17.0.0 the measured `<td>` therefore receives none of the base renderer's work: no `cellProperties.className`, and none of the `readOnly` / `invalid` / `placeholder` / `textEllipsis` / `noWordWrap` classes or a11y attributes.

Two user-visible symptoms follow, both first appearing in 17.0.0:

1. `autoColumnSize` sizes the column for unstyled content, so a larger font wraps.
2. `autoRowSize` records the default height while the master renders the row at its real height. The applied height lands on the row's first child, so the master's `<tr>` simply overflows it — but the inline-start overlay holds only the row header, has no content to overflow, and obeys the cached value exactly. The two panes disagree and every row below the tall one drifts.

Measured on the released build vs this branch (per-cell `font-size: 24px`, `autoRowSize: true`, `rowHeaders: true`, wrapping text):

| | master row heights | row header row heights |
|---|---|---|
| 18.0.0 | `[70, 29, 69, 29, 29]` | `[50, 29, 49, 29, 29]` |
| this PR | `[70, 29, 69, 29, 29]` | `[70, 29, 69, 29, 29]` |

**The fix.** The renderer contract — *format the value, run the cell renderer, run the base renderer when that renderer did not chain it itself, then reset the flag* — used to live only inside `TableView.cellRenderer`, and every consumer that re-implemented part of it drifted (that drift is this regression's root cause). It is now single-sourced in `src/renderers/renderCell.ts`:

- `renderCell()` owns the render sequence (cell renderer → base renderer unless chained → reset `_isBaseRendererCalled`). Called by `TableView.cellRenderer` and by `GhostTable.#renderCell`, which both `createRow()` (row heights) and `createCol()` (column widths) use. The flag reset matters — the flag lives on the cell meta, which for a materialized cell is the same object the render path uses, so leaving it set would make the next real draw skip the base renderer and drop the classes from the actual cell.
- `formatCellValue()` owns the formatter precedence (the cell-level `valueFormatter` option first, then the renderer's own `valueFormatter` static). Called by `TableView.cellRenderer` and by both auto-size samplers.

**The samplers previously applied only the cell-level `valueFormatter` and missed the renderer's static.** The `numeric`, `date`, `intlDate`, and `intlTime` cell *types* copy the formatter into the cell meta, so columns typed that way were measured correctly — but a column configured with a bare `renderer` (`renderer: 'numeric'` with `numericFormat`, or a custom renderer carrying a `valueFormatter` static) rendered the formatted string while being measured against the raw value: `autoColumnSize` under-measured `$123,456,789.55` as `123456789.55`, and `autoRowSize` measured one line for a cell the renderer draws on several. The `time` and `password` types do **not** copy the formatter into the meta, so those were measured against the raw value even when configured through `type`. With `formatCellValue()` in the samplers, the measured string is the rendered string in every configuration.

`GhostTable.#renderCell` itself renders the sampled value **verbatim** — the samplers have already formatted it. Re-formatting there double-formats: a date sample already formatted to `1/1/24` fails the ISO-only `parseToLocalDate` and renders as `#bad-value#`, inflating the measured width (this was caught as an Argos regression on the arabic-rtl visual demo and is enforced by two unit tests).

**No changes to the overlay or height-sync code.** With `autoRowSize` on, `externalRowCalculator` disables oversized-row detection by design; the desync resolves once the cached height is correct.

**Note on the 11 updated renderer specs.** `<Renderer> should internally call base renderer once` counted every base-renderer call. GhostTable now legitimately makes one for its own measured cell, so the count became 2. The specs now ignore calls whose `<td>` carries the `ghost-table` attribute. The original invariant is intact — adding a genuine double call to `TableView` still fails them (verified).

**Behavior change, intended:** grids that use a per-cell `className` with styling that affects geometry, or a column whose resolved renderer carries a `valueFormatter` static that is not mirrored in the cell meta (bare `renderer:` configs, custom renderers, and the built-in `time` / `password` types), now get different (correct) measured row heights and column widths.

### How has this been tested?

New unit tests:

- `renderCell.unit.ts` — the shared helpers: formatter precedence (cell-level wins, static fallback, `this` binding, pass-through) and the render sequence (base renderer runs unless chained, never twice, flag always reset).
- `ghostTable.unit.ts` — the measured `<td>` carries the cell's `className`, the flag is reset, the base renderer is not called twice when the cell renderer already chained it, and the sampled value is rendered verbatim (no re-applied cell-level formatter, no re-applied renderer static).

```bash
npm run test:unit --prefix handsontable -- --testPathPattern=renderCell
npm run test:unit --prefix handsontable -- --testPathPattern=ghostTable
```

New Playwright functional E2E (runs across the theme × bundle matrix):

- `tests/e2e/ghost-table-cell-class.spec.ts` — the measured row height accounts for the per-cell class, the row header matches the data cells, the rows below stay aligned, and the auto column width includes the class. The class sets padding rather than font-size, so the growth is a fixed pixel count on every runner; each size assertion is a floor rather than an equality.
- `tests/e2e/auto-size-value-formatter.spec.ts` — a `renderer: 'numeric'` column with a currency `numericFormat` is measured wider than an identical column without it, a custom renderer whose `valueFormatter` static expands a cell to three lines gets the full measured height, and the row-header overlay stays the same height as the master row. Confirmed to fail on a build without the sampler fix (widths equal, one-line height, overlay 40px shorter) and to pass with it — 3/3 red before, 3/3 green after.

```bash
cd tests && npx playwright test e2e/ghost-table-cell-class.spec.ts e2e/auto-size-value-formatter.spec.ts
```

Full local gate, all green:

```bash
npm run lint --prefix handsontable          # 0 errors
npm run test:unit --prefix handsontable     # 279 suites, 3281 tests
npm run test:e2e --prefix handsontable      # 9606 specs, 0 failures
npm run test:walkontable --prefix handsontable  # 816 specs, 0 failures
cd tests && npm run test:e2e                # Playwright functional suite
```

Targeted legacy reruns after the sampler change: `autoColumnSize` (67 specs), `autoRowSize` (56 specs), `renderers` (312 specs) — 0 failures.

Manual: a two-page demo (`dev-latest.html` on the 18.0.0 CDN build vs `dev-pr.html` on the local build) with the client's configuration, reporting master and row-header heights side by side. Released reports MISALIGNED, this branch reports ALIGNED.

### Types of changes

### Related issue(s):

1. DEV-2126
2. Regression introduced by #11997

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes the shared cell render contract and auto-size measurement path (layout-sensitive), with intentional width/height changes for className and formatter-driven geometry; regression risk is mitigated by broad unit, legacy E2E, and new Playwright coverage.
> 
> **Overview**
> Fixes **DEV-2126** / regression from **#11997**: hidden **GhostTable** cells used for `autoRowSize` and `autoColumnSize` no longer follow the same render path as the visible grid, so measurements ignored **`className`** styling and **`valueFormatter`** precedence (including renderer static formatters).
> 
> **Shared helpers** in `renderCell.ts` centralize **`formatCellValue`** (cell-level formatter, then renderer static) and **`renderCell`** (cell renderer → base renderer when not chained → reset `_isBaseRendererCalled`). **`TableView`**, **`GhostTable`**, and the auto-size samplers now call these instead of duplicating logic.
> 
> **`GhostTable`** routes sampled cells through **`renderCell`**, so measured `<td>` elements get base-renderer classes (e.g. per-cell `className`) and match what users see. Samplers still pre-format values; GhostTable renders those strings verbatim to avoid double-formatting.
> 
> **`autoColumnSize`** and **`autoRowSize`** use **`formatCellValue`** when building samples and probe values, so currency/multiline formatted content sizes correctly.
> 
> Adds unit coverage (`renderCell.unit`, `ghostTable.unit`), Playwright E2E for class-based geometry and formatter statics, doc notes on CSS/theme specificity for custom cell styling, and adjusts renderer “base renderer once” specs to exclude GhostTable’s `ghost-table` cells.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 57ddea1547641d74cde6355694f4d968a9a0f942. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2126 [17.0.0+] autoRowSize misaligns row headers when per-cell font-size is set via CSS](https://app.clickup.com/t/9015210959/DEV-2126)

handsontable/src/utils/ghostTable.ts
```diff
@@ -524,6 +522,59 @@ class GhostTable {
...
+   *
+   * The built-in renderers no longer call `baseRenderer` themselves — `TableView` does it, guarded
+   * by the flag below. Calling only the cell renderer here would leave the measured TD without
+   * `cellProperties.className` and the other base-renderer classes, so AutoRowSize and
+   * AutoColumnSize would measure a cell styled differently from the one the user sees.
+   *
+   * The sampled value must be rendered verbatim — the AutoRowSize and AutoColumnSize samplers
+   * already run `cellProperties.valueFormatter` when building samples. Formatting here again would
+   * double-format: a date sample already formatted to `1/1/24` fails the ISO-only
+   * `parseToLocalDate` and renders as `#bad-value#`, inflating the measured width.
```
demtario, line 538: This paragraph is correct about the cell-level option, but it leaves a real gap open — and the PR description currently claims the opposite is true (see the review summary).

`TableView` resolves the formatter in two branches:

```js
// tableView.ts
if (typeof cellProperties.valueFormatter === 'function') {
  formattedValue = cellProperties.valueFormatter(formattedValue, cellProperties);
} else if (typeof renderer === 'function' && 'valueFormatter' in renderer) {
  formattedValue = valueFormatter.call(cellProperties, formattedValue, cellProperties);
}
```

Both samplers implement only the **first** branch — `autoColumnSize.ts` and `autoRowSize.ts` each do `if (typeof cellMeta.valueFormatter === 'function')` and stop there. The renderer's own static is never applied to a sample.

That matters because the built-in renderers rely on the static being applied upstream. `numericRenderer` and `dateRenderer` both just delegate to `textRenderer` with whatever value they are handed. So a column with `numericFormat: { style: 'currency', currency: 'USD' }` renders `$1,234.50` but is measured against `1234.5`, and `autoColumnSize` under-measures it.

To be clear about where the fix goes: **not here.** Re-applying it in `#renderCell()` is what caused the double-format Argos regression this JSDoc documents. The missing `else if` belongs in the two samplers, right beside the `cellMeta.valueFormatter` branch they already have.

Pre-existing since #11997, same root-cause family as this fix, and in scope of "mirror `TableView`'s renderer contract". A follow-up task is fine — not a blocker for this PR.

handsontable/src/utils/ghostTable.ts
```diff
@@ -524,6 +522,59 @@ class GhostTable {
...
+    renderer(...rendererArgs);
+
+    if (!cellProperties._isBaseRendererCalled) {
+      this.hot!.getCellRenderer({ renderer: 'base' })(...rendererArgs);
+    }
+
+    // The flag lives on the cell meta, which for a materialized cell is the same object the render
+    // path uses. Leaving it set would make the next `TableView.cellRenderer` skip the base renderer
+    // and drop the classes from the REAL cell.
+    cellProperties._isBaseRendererCalled = false;
```
demtario: The renderer contract — *run the cell renderer → run the base renderer if it did not chain → reset the flag* — is now encoded in two places: here, and `tableView.ts` in `cellRenderer`. They have to be kept identical by hand.

That divergence is exactly how the #11997 regression happened, and the `valueFormatter` gap in the other comment is a live second instance of the same drift. Worth extracting a shared helper (e.g. `src/renderers/renderCell.ts`) that both call, so the contract has one definition.

Smaller version of the same point: the argument tuple typed by hand on the next few lines is what `tableView.ts` gets from `Parameters<BaseRenderer>` — reuse that type rather than restating the shape.

demtario (commented):
Reviewed the core change — the contract mirroring in `#renderCell()` is sound, and the flag reset is genuinely load-bearing: `getCellMetaTransient` returns the *stored* meta object for materialized cells, so leaving `_isBaseRendererCalled` set would strip the classes from the real cell on the next draw. The 11 renderer specs are not weakened either — the old bare `spyOn` was already a non-call-through stub, so `.and.callFake` behaves identically and only the counting scope narrowed.

Three things below. Two are inline; this one has no diff line to hang on:

### The PR description no longer matches the code

The description says:

> The same call sites were also missing `valueFormatter` (`tableView.ts`), so an `Intl`-formatted numeric cell was measured against its raw value. `#renderCell()` applies it, covering both the cell-level option and the renderer's own `valueFormatter` static.

Commit 74321310f ("Render sampled values verbatim in GhostTable") reverted exactly that. The JSDoc on `#renderCell()` now states the opposite, and two new unit tests enforce it:

- `ghostTable.unit.ts` — *should render the sampled value verbatim, without re-applying the cell valueFormatter*
- `ghostTable.unit.ts` — *should render the sampled value verbatim, without applying the renderer's valueFormatter static*

Could you update the description before merge? It is the reviewer-facing contract, and right now it describes behavior the code explicitly forbids.

handsontable/src/utils/ghostTable.ts
adrianspdev: Fixed in this PR rather than a follow-up — 84d15e0c2a. Both samplers (and AutoColumnSize's changed-cell probe) now format through a shared `formatCellValue()`, which applies the cell-level `valueFormatter` first and falls back to the renderer's `valueFormatter` static — the same precedence as `TableView`.

One correction to the analysis: the `numeric`, `date`, `intlDate`, and `intlTime` cell **types** copy the formatter into the cell meta (`numericType.ts` spreads `valueFormatter`), so `type: 'numeric'` with `numericFormat` was already measured correctly by the existing cell-level branch. The gap only hit columns whose resolved renderer carries the static without a meta mirror: bare `renderer: 'numeric'` configs, custom renderers with a `valueFormatter` static, and the built-in `time` / `password` types (which do not copy it into the meta).

Covered by `tests/e2e/auto-size-value-formatter.spec.ts` (a `renderer: 'numeric'` + currency column measured wider than an identical raw column, a custom-renderer static expanding a cell to three lines measured at full height, and the row-header overlay matching the master) — 3/3 red on a build without the sampler change, 3/3 green with it — plus `formatCellValue` precedence unit tests in `renderCell.unit.ts`. The fix stays out of `#renderCell()`, which keeps rendering the sampled value verbatim.

handsontable/src/utils/ghostTable.ts
adrianspdev: Extracted in 19b36144de: `src/renderers/renderCell.ts` now owns both halves of the contract — `renderCell()` (cell renderer → base renderer unless chained → reset flag) and `formatCellValue()` (cell-level `valueFormatter`, then the renderer's static). `TableView.cellRenderer`, `GhostTable#renderCell`, and both auto-size samplers call the helpers; the hand-typed tuple here is now `Parameters<BaseRenderer>`. The rule is also written into `handsontable/AGENTS.md` so the sequence never gets restated inline again.

Two things surfaced while consolidating: the flag reset now happens in a `finally` (57ddea1547) — a renderer that chained the base renderer and then threw used to leave the flag set on the shared cell meta, making the next draw drop the cell's classes — and the one remaining inline restatement of the formatter precedence is the Filters "by value" list (`filters/component/value.ts`), left for a follow-up because blanket static application would be wrong there (a password column's value list would collapse to identical `***` entries).

adrianspdev:
Description updated per the review — the stale `valueFormatter` paragraph now describes the shipped behavior: `#renderCell()` renders sampled values verbatim, the samplers own the formatting through the shared `formatCellValue()` (cell-level option first, then the renderer's static), and the behavior-change note spells out which configurations get new measurements (bare `renderer:` configs, custom renderers with a static, and the `time` / `password` types, which never mirrored the formatter into the cell meta).

Both inline points are addressed in code — details in the threads: the sampler gap is fixed in this PR (84d15e0c2a, with a red-first Playwright spec), and the renderer contract is single-sourced in `src/renderers/renderCell.ts` (19b36144de), with the flag reset hardened against throwing renderers (57ddea1547).

demtario (approved):

budnix (approved):

PR #13194, opened 2026-08-14
DEV-1619: Make the grid work inside Shadow DOM and Salesforce LWS

Haxikowy:
### Context

The PR makes Handsontable work inside Shadow DOM trees and sandboxed web-component platforms — most importantly Salesforce Lightning Web Components — without wrapper-side glue code.

When the grid renders inside a shadow root, document-level listeners see events retargeted to the shadow host and `document.activeElement` reports the host instead of the focused element. Every in-grid click classified as an outside click: the open editor closed on the first click into it, the selection dropped, and the grid stopped listening. Salesforce Lightning Web Security (LWS) additionally filters `composedPath()` down to the shadow host chain for sandboxed code and hides foreign shadow internals from `activeElement` resolution, which also broke copy/paste, let the grid steal focus back from external inputs, and made genuine outside clicks undetectable.

The changes:

- New DOM helpers: `getDeepActiveElement()` descends `shadowRoot.activeElement` chains; `getShadowHostChain()` returns the shadow hosts an element is rendered within. Used in `tableView`, the focus manager, editors, comments, `selectElementIfAllowed()`, and Walkontable's mousedown guard.
- Outside-click classification uses `composedPath()` plus the grid's shadow host chain instead of a `parentNode` walk (removes the dead `isTargetWebComponent` branch).
- `CopyPaste` binds its clipboard listeners to the grid's shadow root next to the document ones (with per-event dedupe) and resolves the event source from `event.target` when it points inside the grid — the only reliable reference under LWS.
- `FocusGridManager` owns the focus semantics: it tracks browser focus with `focusin`/`focusout` listeners bound inside the grid's root wrapper and portal (where sandboxed hosts still report true event targets) and classifies foreign focus targets. Both are public API: `getFocusManager().hasBrowserFocus()` and `.isForeignFocusTarget(element)`. `TableView` only orchestrates: a mouseup that neither started in the grid nor left the focus inside it releases the keyboard listener and applies `outsideClickDeselects`.
- The core stamps the `ht-shadow-dom` class on the root wrapper when the grid renders inside a shadow root; the base stylesheet turns it into an isolated stacking context (`isolation: isolate`), so internal z-index values stop competing with the host page UI (e.g. frozen headers painting above Salesforce navigation). Light DOM embeddings keep the current stacking behavior untouched.
- `EventManager`'s element type accepts `ShadowRoot`.
- New guide: *Shadow DOM* (`docs/content/guides/tools-and-building/shadow-dom/`).

**Upgrade-visible behavior change (Shadow DOM embeddings only):** grids inside a shadow root now render in an isolated stacking context. A host page that relied on grid internals painting above its own UI will see the new, contained stacking.

Known limitations (documented in the guide): the `ht-shadow-dom` class is stamped at init only; an editor overflowing the wrapper can paint under host content that creates its own stacking context; the synthetic Shadow DOM polyfill (older Salesforce experiences) is untested.

### How has this been tested?

- New Playwright E2E suite `tests/e2e/shadow-dom.spec.ts` (7 tests) with a native-shadow-root fixture: editor survives a click into its textarea, selection moves without dropping the keyboard listener, copy/paste via keyboard shortcuts, context menu action on the selection that opened it, no focus stealing from an outside input, outside-click deselect, z-index isolation. The editor-survival test was verified red against the pre-fix build. Green on the full theme × bundle matrix (42/42 across all six legs).
  - `cd tests && npx playwright test e2e/shadow-dom.spec.ts`
- Unit tests for `getDeepActiveElement()` (incl. nested shadow roots) and `getShadowHostChain()` (`npm run test:unit --prefix handsontable -- --testPathPattern=helpers/dom`).
- Full regression: legacy E2E 9606/9606, unit 3271/3271, Walkontable suite, React and Vue wrapper suites, eslint/stylelint/tsc. The mouseup logic was regression-tested against the context menu, dropdown menu, and MultiSelect editor suites specifically — early iterations broke all three, and the final shape is pinned by them.
- Verified live in a Salesforce Developer Edition org (Lightning's default synthetic shadow DOM + LWS, live Account data via Lightning Data Service; the native opt-in `shadowSupportMode = 'native'` is verified on a dedicated page and requires the stylesheets injected into the component's shadow tree, since Salesforce's `loadStyle` targets the document head, which a native shadow root ignores): all five reported symptoms fixed — editor close on click-in, broken copy/paste, focus stolen from the Lightning global search while typing, missing outside-click deselect (including targets whose focus LWS hides behind the host chain), and grid headers painting above the Salesforce navigation. Also verified: two grid instances on one Lightning page with cross-grid focus handoff, and listener cleanup on `destroy()` (add/remove balanced).

### Types of changes

### Related issue(s):

1. DEV-1619

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-1619

---

> [!NOTE]
> **Medium Risk**
> Touches global mouseup/mousedown, focus, and clipboard paths used on every interaction; behavior is intentionally different only in shadow/sandbox cases but regression risk is moderate given scope (context menus, dropdowns, MultiSelect were previously fragile).
> 
> **Overview**
> Fixes **Handsontable inside Shadow DOM** (web components, Salesforce LWC/LWS) where document-level APIs retarget events and collapse focus, so in-grid clicks were treated as outside clicks and editors, selection, copy/paste, and focus handoff broke.
> 
> **DOM helpers** `getDeepActiveElement()` and `getShadowHostChain()` replace raw `document.activeElement` and parent walks across editors, Walkontable, focus scope, copy/paste, and `tableView`.
> 
> **Outside-click and listen state** in `tableView` use `composedPath()`, shadow host chains, and `FocusGridManager.hasBrowserFocus()` / `isForeignFocusTarget()` (new `focusin`/`focusout` tracking inside the grid tree) so deselect and `unlisten` behave under filtered `composedPath()` and sandboxed `activeElement`.
> 
> **Copy/paste** adds shadow-root clipboard listeners with deduplication and smarter event-target resolution for LWS-filtered paths.
> 
> **Presentation:** shadow-hosted grids get `ht-shadow-dom` on the root wrapper with `isolation: isolate` so internal z-indexes do not paint over host UI.
> 
> Adds a **Shadow DOM** guide, **AGENTS.md** rules, changelog **#13194**, and Playwright E2E for shadow-embedded and in-cell widget scenarios.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8cebb7d6e94f44238a486979109230b4ceb9798d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/handsontable/src/focusManager/grid.ts
+++ b/handsontable/src/focusManager/grid.ts
@@ -1,6 +1,13 @@
 import type { HotInstance } from '../core/types';
+import EventManager from '../eventManager';
 import { warn } from '../helpers/console';
-import { isHTMLElement, isOutsideInput } from '../helpers/dom/element';
+import {
+  getDeepActiveElement,
+  getShadowHostChain,
+  isHTMLElement,
+  isInternalElement,
+  isOutsideInput,
+} from '../helpers/dom/element';
 import { debounce } from '../helpers/function';
 
 /**
@@ -38,6 +45,21 @@ export class FocusGridManager {
    * The Handsontable instance.
    */
   #hot: HotInstance;
+  /**
+   * Event manager used for the DOM listeners that track the browser focus within the grid.
+   *
+   * @type {EventManager}
+   */
+  #eventManager: EventManager | null = null;
+  /**
+   * Tracks whether the browser focus currently sits inside the grid's root wrapper or portal.
+   * Maintained by `focusin`/`focusout` listeners bound within the grid's own DOM tree, where
+   * sandboxed hosts (e.g. Salesforce Lightning Web Security) still report true event targets -
+   * unlike `document.activeElement`, which such hosts collapse to the outermost shadow host.
+   *
+   * @type {boolean}
+   */
+  #hasBrowserFocus = false;
   /**
    * The currently enabled focus mode.
    * Can be either:
@@ -106,6 +128,64 @@ export class FocusGridManager {
     this.#hot.addHook('afterSelectionFocusSet', (...args: unknown[]) => this.#onAfterSelectionChange());
     this.#hot.addHook('afterSelectionEnd', (...args: unknown[]) => this.#focusEditorElement());
     this.#hot.addHook('afterRender', (...args: unknown[]) => this.#onAfterRender());
+
+    this.#eventManager = new EventManager(this.#hot);
+
+    const focusRootElements = [this.#hot.rootWrapperElement ?? this.#hot.rootElement, this.#hot.rootPortalElement]
+      .filter(element => element instanceof HTMLElement);
+
+    focusRootElements.forEach((focusRootElement) => {
+      this.#eventManager!.addEventListener(focusRootElement, 'focusin', () => {
+        this.#hasBrowserFocus = true;
+      });
+
+      this.#eventManager!.addEventListener(focusRootElement, 'focusout', (event) => {
+        const relatedTarget = (event as FocusEvent).relatedTarget;
+
+        this.#hasBrowserFocus = relatedTarget instanceof Node &&
+          focusRootElements.some(focusRoot => focusRoot.contains(relatedTarget));
+      });
+    });
+  }
+
+  /**
+   * Checks whether the browser focus currently sits inside the grid's root wrapper or portal.
+   * The state comes from focus events observed within the grid's own DOM tree, so it stays
+   * correct in environments where `document.activeElement` cannot be resolved through foreign
+   * shadow boundaries.
+   *
+   * @returns {boolean}
+   */
+  hasBrowserFocus(): boolean {
+    return this.#hasBrowserFocus;
+  }
+
+  /**
+   * Checks whether the given focused element proves that the browser focus left the grid.
+   * Focus counts as foreign when the element is focusable, holds the browser focus, and is
+   * neither a part of the grid or its portal, nor one of the shadow hosts the grid is rendered
+   * within, nor the document body (a click on a non-focusable area keeps the focus where it
+   * was). Sandboxed hosts (e.g. Salesforce Lightning Web Security) hide the real focused
+   * element behind the host chain, so a focused ancestor host cannot prove the focus left the
+   * grid - only an unrelated element can.
+   *
+   * @param {HTMLElement | null} element The deepest reachable focused element.
+   * @returns {boolean}
+   */
+  isForeignFocusTarget(element: HTMLElement | null): boolean {
+    const { rootElement, rootPortalElement, rootDocument } = this.#hot;
+
+    if (
+      element === null ||
+      element === rootDocument.body ||
+      element === rootDocument.documentElement ||
+      isInternalElement(element, rootElement) ||
+      (rootPortalElement && rootPortalElement.contains(element))
+    ) {
+      return false;
+    }
+
+    return !getShadowHostChain(rootElement).includes(element);
   }
 
   /**
@@ -319,7 +399,8 @@ export class FocusGridManager {
     const suspended = this.#isSuspended;
 
     this.#getSelectedCell((selectedCell) => {
-      const activeElement = this.#hot.rootDocument.activeElement as HTMLElement | null;
+      const deepActiveElement = getDeepActiveElement(this.#hot.rootDocument);
+      const activeElement = isHTMLElement(deepActiveElement) ? deepActiveElement : null;
 
       // When focus management is suspended and an external focusable element (outside Handsontable)
       // currently owns the browser focus, keep it - do not blur and do not move focus to the cell.
@@ -351,7 +432,8 @@ export class FocusGridManager {
     // owns focus, fall through so the default `imeFastEdit` behavior still moves focus to the
     // editor textarea (existing per-editor IME support tests rely on this). See #10038.
     if (this.#isSuspended) {
-      const activeElement = this.#hot.rootDocument.activeElement as HTMLElement | null;
+      const deepActiveElement = getDeepActiveElement(this.#hot.rootDocument);
+      const activeElement = isHTMLElement(deepActiveElement) ? deepActiveElement : null;
 
       if (activeElement && isOutsideInput(activeElement)) {
         return;
--- a/handsontable/src/focusManager/grid.ts
+++ b/handsontable/src/focusManager/grid.ts
@@ -132,7 +132,7 @@ export class FocusGridManager {
     this.#eventManager = new EventManager(this.#hot);
 
     const focusRootElements = [this.#hot.rootWrapperElement ?? this.#hot.rootElement, this.#hot.rootPortalElement]
-      .filter(element => element instanceof HTMLElement);
+      .filter(element => isHTMLElement(element));
 
     focusRootElements.forEach((focusRootElement) => {
       this.#eventManager!.addEventListener(focusRootElement, 'focusin', () => {
@@ -142,7 +142,7 @@ export class FocusGridManager {
       this.#eventManager!.addEventListener(focusRootElement, 'focusout', (event) => {
         const relatedTarget = (event as FocusEvent).relatedTarget;
 
-        this.#hasBrowserFocus = relatedTarget instanceof Node &&
+        this.#hasBrowserFocus = isHTMLElement(relatedTarget) &&
           focusRootElements.some(focusRoot => focusRoot.contains(relatedTarget));
       });
     });
--- a/handsontable/src/focusManager/grid.ts
+++ b/handsontable/src/focusManager/grid.ts
@@ -169,6 +169,12 @@ export class FocusGridManager {
    * element behind the host chain, so a focused ancestor host cannot prove the focus left the
    * grid - only an unrelated element can.
    *
+   * The check mirrors the scope-deactivation rule in the focus scope manager (`scopeManager`
+   * unlistens when a focus event lands outside every scope). Sandboxed hosts retarget the
+   * window-level focus events that rule depends on, so the mouseup path re-applies the same
+   * rule from the deepest reachable focused element. On a regular page both mechanisms agree -
+   * this adds no behavior that the scope manager does not already provide there.
+   *
    * @param {HTMLElement | null} element The deepest reachable focused element.
    * @returns {boolean}
    */
--- a/handsontable/src/helpers/dom/element.ts
+++ b/handsontable/src/helpers/dom/element.ts
@@ -1351,7 +1351,7 @@ export function isOutsideInput(element: HTMLElement): boolean {
  * @param {HTMLElement} element - DOM element.
  */
 export function selectElementIfAllowed(element: HTMLElement): void {
-  const activeElement = element.ownerDocument.activeElement;
+  const activeElement = getDeepActiveElement(element.ownerDocument);
 
   if (isHTMLElement(activeElement) && !isOutsideInput(activeElement) &&
       (element instanceof HTMLInputElement || element instanceof HTMLTextAreaElement)) {
@@ -1509,6 +1509,50 @@ export function isHTMLInputElement(element: HTMLElement): element is HTMLInputEl
  */
 export function isShadowRoot(node: Node): node is ShadowRoot {
   return node.nodeType === Node.DOCUMENT_FRAGMENT_NODE && 'host' in node;
+}
+
+/**
+ * Gets the chain of shadow host elements the node is rendered within, ordered from the
+ * closest host outward to the host attached to the document.
+ *
+ * @param {Node} node The node to read the host chain from.
+ * @returns {HTMLElement[]} The shadow host elements; empty when the node is not in a Shadow DOM tree.
+ */
+export function getShadowHostChain(node: Node): HTMLElement[] {
+  const hosts: HTMLElement[] = [];
+  let rootNode = node.getRootNode();
+
+  while (isShadowRoot(rootNode)) {
+    const { host } = rootNode;
+
+    if (!isHTMLElement(host)) {
+      break;
+    }
+
+    hosts.push(host);
+    rootNode = host.getRootNode();
+  }
+
+  return hosts;
+}
+
+/**
+ * Gets the deepest active (focused) element in the document. When the focus is inside
+ * a Shadow DOM tree, `document.activeElement` points at the shadow host instead of the
+ * focused element. This helper descends through the open shadow roots and returns
+ * the element that actually holds the focus.
+ *
+ * @param {Document} rootDocument The document to read the active element from.
+ * @returns {Element|null} The deepest focused element, or `null` when nothing is focused.
+ */
+export function getDeepActiveElement(rootDocument: Document): Element | null {
+  let element = rootDocument.activeElement;
+
+  while (element?.shadowRoot?.activeElement) {
+    element = element.shadowRoot.activeElement;
+  }
+
+  return element;
 }
 
 /**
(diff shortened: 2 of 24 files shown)


qunabu:

handsontable/src/tableView.ts
```diff
@@ -1520,6 +1532,25 @@ class TableView {
...
+   * @private
+   * @returns {boolean}
+   */
+  #isPathWithinGrid(eventPath: EventTarget[]): boolean {
+    const { rootElement, rootPortalElement } = this.hot;
+
+    return eventPath.includes(rootElement) ||
+      (!!rootPortalElement && eventPath.includes(rootPortalElement)) ||
+      getShadowHostChain(rootElement).some(host => eventPath.includes(host));
+  }
```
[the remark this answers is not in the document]
Haxikowy: Fixed in b0bb0eb — the host chain fallback now applies only to filtered paths (no ShadowRoot entries in composedPath). Covered by the new shadow-sibling spec (red pre-fix).

handsontable/src/focusManager/grid.ts
```diff
@@ -106,6 +128,64 @@ export class FocusGridManager {
...
+      this.#eventManager!.addEventListener(focusRootElement, 'focusin', () => {
+        this.#hasBrowserFocus = true;
+      });
+
+      this.#eventManager!.addEventListener(focusRootElement, 'focusout', (event) => {
+        const relatedTarget = (event as FocusEvent).relatedTarget;
+
+        this.#hasBrowserFocus = relatedTarget instanceof Node &&
+          focusRootElements.some(focusRoot => focusRoot.contains(relatedTarget));
+      });
```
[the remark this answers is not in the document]
Haxikowy: Fixed in 4df0681 — both checks use the realm-safe isHTMLElement() helper.

handsontable/src/helpers/dom/element.ts
```diff
@@ -1511,6 +1511,50 @@ export function isShadowRoot(node: Node): node is ShadowRoot {
...
+ */
+export function getDeepActiveElement(rootDocument: Document): Element | null {
+  let element = rootDocument.activeElement;
+
+  while (element?.shadowRoot?.activeElement) {
+    element = element.shadowRoot.activeElement;
+  }
+
+  return element;
+}
```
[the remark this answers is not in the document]
Haxikowy: Added in 4ff4f90 — Critical Rules bullet in handsontable/AGENTS.md pointing at the shadow-aware helpers.

adrianspdev:
Reviewed across architecture, code quality, and performance/a11y. Solid work overall — the helper coverage, the clipboard dedupe, and the listener cleanup (verified: the focus manager's `EventManager` shares the instance's `eventListeners` context, so core `destroy()` clears it) all check out. Three findings, one of which I'd like resolved before merge.

## High

### 1. `outsideClickDeselects` callback is now invoked twice per outside click — in plain light DOM too

`handsontable/src/tableView.ts` (mouseup handler, lines ~443–453)

The new mouseup block calls the `outsideClickDeselects` callback after the pre-existing mousedown outside-click path (lines ~520–529) has already called it. When the callback returns `false`:

1. mousedown → callback invocation #1 → selection kept,
2. mouseup → `isForeignFocusTarget(activeElement)` is `true` (in light DOM `getShadowHostChain(rootElement)` is empty, so any outside focusable element qualifies) → `selection.isSelected()` is still `true` → callback invocation #2.

The PR's own E2E pins the doubling — `expect(await grid.outsideClickTargets()).toEqual(['focus-mover', 'focus-mover'])` in `tests/e2e/shadow-dom.spec.ts` — so this isn't accidental flakiness; it's the new contract. But it's an observable behavior change of a public option for anyone with a side-effectful callback, it isn't limited to Shadow DOM embeddings, and it isn't mentioned in the option's JSDoc, the docs, or the changelog.

Suggestion: record that the mousedown handler already classified this click cycle (a flag set in the mousedown outside path, cleared on the next mousedown) and skip the second invocation on mouseup, so the callback fires once per click as before.

## Medium

### 2. `isForeignFocusTarget()` classifies the grid's own slot/overlay UI as foreign

`handsontable/src/focusManager/grid.ts` (lines ~364–378)

The method checks containment against `rootElement` and the portal only. Pagination mounts into `ht-slot-bottom`, and dialog/notification UI goes into `rootOverlaysElement` — both live inside `rootWrapperElement` but outside `rootElement`. For a pagination button, `isInternalElement(button, rootElement)` returns `false`: its `closest('.handsontable')` is the pagination container, whose parent is the slot element, not `rootElement`.

In Chrome, clicking a `<button>` focuses it, so a click on a pagination control with cells selected makes mouseup see `isFocusLostToOutside === true` → `hot.unlisten()` plus the deselect/callback branch — for the grid's own UI.

This is also inconsistent with `#hasBrowserFocus` tracking in the same class, which correctly binds to `rootWrapperElement ?? rootElement`. Suggestion: treat elements contained in `rootWrapperElement` (when present) as non-foreign.

## Low

### 3. `composedPath()` computed twice in the mouseup handler

`handsontable/src/tableView.ts:439` — `eventPath` already exists at line 431; pass it to `this.#isPathWithinGrid(eventPath)` instead of recomputing.

### 4. JSDoc inaccuracy on `getDeepActiveElement`

`handsontable/src/helpers/dom/element.ts` — the JSDoc says it returns "`null` when nothing is focused", but the browser reports `document.body` in that case, and the PR's own unit test asserts exactly that ("should return `document.body` when nothing is focused"). Please fix the sentence.

Haxikowy:
Thanks for the thorough pass.

**1.** Agree, fixed in ef4797a. The mousedown outside path now marks the click cycle as classified and the mouseup pass skips its own invocation, so the callback fires once per click again. The flag mirrors `#mouseDown`: reset on every mousedown, consumed and cleared on mouseup, so a stale value can't survive an unmatched mouseup. The mouseup pass still covers cycles mousedown can't classify (LWS-filtered paths), and the E2E now pins the single invocation.

**2.** Your reading of the code is right, but I couldn't reproduce a regression. I compared released 18.0.0 (CDN) with this branch using `pagination: true`: with default settings a pagination click deselects on both, through the pre-existing mousedown path, which also checks `rootElement` rather than the wrapper. With `outsideClickDeselects: false` both keep the selection and keep listening. I also checked on a Salesforce Lightning page (LWS) with pagination enabled: same outcome. A programmatic page change keeps the selection, so it really is the click path, not the re-render. Treating wrapper-contained elements as non-foreign would change released behavior (pagination clicks would stop deselecting), so I'd rather handle it in a follow-up ticket than in this PR. Happy to file it.

**3.** Done in 566461b.

**4.** Done in 47f3966.

qunabu (dismissed):

adrianspdev (dismissed):

budnix (commented):
A few notes on the outside-click and clipboard paths. The first one on `tableView.ts` is the one I would want resolved before merge; the rest are smaller. I skipped anything already covered by the earlier review threads.

handsontable/src/tableView.ts
```diff
@@ -407,23 +419,49 @@ class TableView {
...
+        (!wasInsideGridClick && !this.hot.getFocusManager().hasBrowserFocus() && !isPathThroughGridUi);
+
+      if (isOutsideInputElement || isFocusLostToOutside ||
+          (!selection.isSelected() && !selection.isSelectedByAnyHeader() &&
+          !this.#isPathWithinGrid(eventPath) && !isRightClick(event))) {
         this.hot.unlisten();
       }
+
+      if (!wasOutsideClickHandled && activeHTMLElement !== null &&
+          isFocusLostToOutside && selection.isSelected()) {
```
budnix: This block only asks where the focus sits, not where the click landed. So if an app parks focus in an outside input and selects cells with `selectCells(coords, true, false)`, a later click on grid padding, a resize handle, or the scrollbar clears the selection — `#mouseDown` stays `false` there, so the mousedown path never marks the click as handled. The unlisten check right above already guards with `#isPathWithinGrid` and `isRightClick`; adding the same two here would keep this to real outside clicks.

```suggestion
          isFocusLostToOutside && selection.isSelected() &&
          !this.#isPathWithinGrid(eventPath) && !isRightClick(event)) {
```

handsontable/src/tableView.ts
```diff
@@ -407,23 +419,49 @@ class TableView {
...
+      if (activeHTMLElement !== null && isInput(activeHTMLElement) && !isOutsideInputElement) {
         return;
       }
 
-      if (isOutsideInputElement || (!selection.isSelected() && !selection.isSelectedByAnyHeader() &&
-          !(rootWrapperElement ?? rootElement).contains(event.target as Node) && !isRightClick(event))) {
+      const eventPath = event.composedPath();
+      const isPathThroughGridUi = eventPath.includes(rootWrapperElement ?? rootElement) ||
+        (this.hot.rootPortalElement && eventPath.includes(this.hot.rootPortalElement));
+      const isFocusLostToOutside = this.hot.getFocusManager().isForeignFocusTarget(activeHTMLElement) ||
```
budnix, line 444: `isForeignFocusTarget()` returns `true` for any outside element on a plain page with no shadow root, so clicking an outside button or link now calls `unlisten()` where it did not before. With `outsideClickDeselects: false` the selection stays but keyboard navigation stops, which changes things for people who put external controls next to the grid. The changelog entry is marked `breaking: false` — did this end up being the intended behavior on non-shadow pages too?

handsontable/src/tableView.ts
```diff
@@ -1520,6 +1549,34 @@ class TableView {
...
+   * filtered path (no ShadowRoot entries) falls back to the shadow host chain check, which
+   * matters for sandboxed hosts (e.g. Salesforce Lightning Web Security) that collapse paths
+   * observed at the document level to the visible host chain, hiding the grid internals.
+   *
+   * @param {EventTarget[]} eventPath The event propagation path (`event.composedPath()`).
+   * @private
+   * @returns {boolean}
+   */
+  #isPathWithinGrid(eventPath: EventTarget[]): boolean {
+    const { rootElement, rootPortalElement } = this.hot;
```
budnix, line 1567: Line 442 in the same handler uses `rootWrapperElement ?? rootElement`, and the old mouseup check did too, but this helper matches only `rootElement`. `rootElement` sits inside the wrapper as a sibling of `ht-slot-top`, `ht-slot-bottom`, and `ht-overlay`, so a click on any of those now reads as outside and the grid unlistens. Was leaving the wrapper out on purpose?

handsontable/src/plugins/copyPaste/copyPaste.ts
```diff
@@ -708,14 +739,35 @@ export class CopyPaste extends BasePlugin {
...
+   * correctly for listeners bound within the grid's shadow tree - the retargeted `target` is
+   * the only reference to the real source element in such environments.
+   *
+   * @param {ClipboardEvent | PasteEvent} event The clipboard event to resolve.
+   * @returns {unknown} The deepest known event source element.
+   */
+  #resolveClipboardEventTarget(event: ClipboardEvent | PasteEvent): unknown {
+    const target = 'target' in event ? event.target : null;
+
+    if (isHTMLElement(target) && isInternalElement(target, this.hot.rootElement)) {
```
budnix: Preferring the retargeted `target` breaks copy for a web component that a custom renderer puts in a cell. `event.target` retargets to that component's host, which is inside the grid, so this returns the host, the guard in `onCopy` no longer bails out, and Handsontable calls `preventDefault()` and copies the cell range instead of the text the user selected inside the widget. Before, `composedPath()[0]` gave the inner element and the event passed through — maybe only prefer `target` when the path looks filtered, the way `#isPathWithinGrid` decides it?

handsontable/src/plugins/copyPaste/copyPaste.ts
```diff
@@ -265,9 +274,31 @@ export class CopyPaste extends BasePlugin {
...
+      }
+      this.#processedClipboardEvents.add(event);
+      handler(event);
+    };
+
+    this.eventManager.addEventListener(this.hot.rootDocument, 'copy', dedupe((e: ClipboardEvent) => this.onCopy(e)));
+    this.eventManager.addEventListener(this.hot.rootDocument, 'cut', dedupe((e: ClipboardEvent) => this.onCut(e)));
+    this.eventManager.addEventListener(this.hot.rootDocument, 'paste', dedupe((e: ClipboardEvent) => this.onPaste(e)));
+
+    const rootNode = this.hot.rootElement.getRootNode();
```
budnix, line 289: `getRootNode()` is read once here, and `enablePlugin()` runs during construction. A grid built on a detached container and moved into a shadow root afterwards — a common web component pattern — never gets these listeners, so clipboard quietly falls back to the document ones. `core.ts:388` reads the root node once for the `ht-shadow-dom` class in the same way, and the new guide calls that limit out for the CSS class but not for clipboard. Worth a line in the guide at least.

handsontable/src/tableView.ts
Haxikowy: Applied your suggestion in d3a2d6d34. New Playwright fixture (tests/e2e/widget-cell.spec.ts) covers it with a focus-preserving widget inside a cell — red before the guard (the in-grid click dropped the selection), green after.

handsontable/src/plugins/copyPaste/copyPaste.ts
Haxikowy: Fixed in fd062dbf4. The resolver now keys on the path instead of the target: a path with a ShadowRoot entry is complete, so path[0] wins; a path without one either involves no shadow tree (target equals path[0]) or was filtered by LWS (target is the only reliable reference). The widget-cell spec covers the pass-through — before the fix the grid called preventDefault on a copy over text selected inside an in-cell web component.

handsontable/src/plugins/copyPaste/copyPaste.ts
Haxikowy: Added in 8859a2531 — the known-limitations bullet now names the clipboard listeners next to the ht-shadow-dom class as consequences of the create-time root-node read.

handsontable/src/tableView.ts
Haxikowy: Left out on purpose — it mirrors develop. The old mousedown walk stopped only at rootElement/rootPortalElement, so slot and overlay clicks counted as outside clicks there too; line 442 keeps the wrapper because the old mouseup check used it. Reproduced on released 18.0.0: with pagination enabled, select a cell and click the pagination bar (or a page-nav button) — the selection drops, same as on this branch. Pre-existing, so I'd rather fix it in a follow-up ticket than widen this PR's outside-click change.

handsontable/src/tableView.ts
Haxikowy: Good catch that the diff reads like new behavior, but it turned out to be pre-existing. I compared this branch side by side with released 18.0.0 (`outsideClickDeselects: false`, a plain button and a link next to the grid): both builds stop listening the moment the element takes focus, selection kept on both. The unlisten comes from the focus scope manager that shipped in 18.0 — `scopeManager.ts` deactivates the scope and calls `unlisten()` for any focus target outside the scopes. A programmatic `el.focus()` with no mouse events at all flips `isListening()` to false on the 18.0.0 CDN bundle, so the mouse path isn't involved.

The mouseup check only re-applies that same rule for hosts that retarget window-level focus events (LWS), where the scope manager can't see the real target. On a regular page the two mechanisms agree. I documented the relationship in the `isForeignFocusTarget` JSDoc in 8cebb7d6e so the next reader doesn't have to rediscover it.

Only delta I could construct is contrived: `hot.listen()` called programmatically while focus already sits outside, then a click on another outside focusable — old kept listening until a focus change, new unlistens on the click. That matches what the old mouseup already did when the parked element was an input, so I'd leave `breaking: false` as is.

budnix (approved):

PR #13195, opened 2026-08-17
DEV-2542: Add sanitizer to docs examples rendering HTML, fix stale dateFormat

demtario:
### Context

The PR fixes the docs example content that the demo runtime monitor flagged on demos.handsontable.com. Two Sentry issues were the monitor working correctly - the examples really were wrong:

- **DEMOS-14** - `HTML content is being written to the DOM ("contextMenu") without a sanitizer.` Handsontable ships no built-in sanitizer since 18.0, so an example whose context-menu item label returns HTML writes that markup to the DOM unsanitized and logs a warning.
- **DEMOS-10** - `The dateFormat option as a string is not supported.` 18.0 dropped the string form (PRO-986). The renderer now ignores a string and renders the raw ISO value, so affected examples displayed unformatted dates.

Nothing else in the pipeline catches this class: the examples compile and render, they just use APIs that 18.0 removed or hardened.

**A grep-based sweep is not sufficient here.** It missed multi-line `colHeaders` arrays and wrapped string literals. A browser sweep over all 884 docs page/framework combinations (221 permalinks x 4 frameworks), reading the console instead of the source, found **six** affected families where grep found two - including the `dialog` case in four more frameworks and the Angular variants of `autocomplete-cell-type` and `cell-type`.

Four families are fixable in docs and are fixed here. Each keeps the rendered output identical, except `cell-type` example2 where the `<br>` was purely cosmetic and is removed instead of sanitized:

| Page | Surface | Change |
|---|---|---|
| `context-menu` example3 | `contextMenu` | Allowlist sanitizer (`B,I,EM,STRONG,BR`) |
| `dialog` examples 3 and 7 | `dialog` | Allowlist sanitizer (`H2,P,STRONG,BUTTON,BR` plus `id,class,type`) |
| `autocomplete-cell-type` example2, `rows-sorting` | `header` | `BR`-only sanitizer, `<br>` preserved |
| `cell-renderer` example5/6 | `header` | Allowlist keeping the interactive header checkbox |
| `cell-type` example2 | `header` | Cosmetic `<br>` removed |

`id`, `class`, and `type` are allowlisted deliberately: they carry `aria-labelledby`/`aria-describedby` in dialog example7 and the `getElementById('example3-button')` handler in dialog example3.

### The sanitizer is grid-level, not per-surface (review fix, commit 3)

The first pass wrote each helper as a narrow allowlist for the one surface its page teaches. That was wrong, and code review caught it. `sanitizer` is a **grid-level** control: `plugins/copyPaste/copyPaste.ts:891` routes pasted HTML through the same function, and the structure-aware branch tests the **sanitized** string:

```js
if (textHTML && /(<table)|(<TABLE)/g.test(textHTML)) {
  const parsedConfig = htmlToGridSettings(textHTML, this.hot.rootDocument);
```

Stripping the table tags made that gate miss, so every grid that gained a sanitizer silently downgraded a range paste to `text/plain`, losing `htmlToGridSettings` (cell boundaries, `colspan`/`rowspan`). Measured before the fix: a two-row table sanitized to `Trail Helmet1298.14Windbreaker178.9`, gate `false`.

All 36 helpers now share one canonical shape:

- table structure plus `colspan`/`rowspan` are allowed, so the paste gate matches
- `SCRIPT`, `STYLE`, `TEXTAREA`, and `TITLE` are **removed** rather than unwrapped. Unwrapping promoted their source text into the output, contradicting the adjacent comment - `<script>alert(1)</script>` rendered the literal `alert(1)`, and `<style>body{color:red}</style>` rendered its declaration. Not an XSS vector (`fastInnerHTML` re-parses), but wrong and misleading in an example that teaches sanitizing.
- a comment names the grid-level/paste coupling, which the docs state nowhere today

Known trade-off: the canonical block runs about 30 lines and is repeated per example, so several of these files now exceed the 25-60 line guidance in `docs/AGENTS.md` §2.5. Docs examples must be self-contained single files - the runner executes one entry file per example - so the alternative is not demonstrating `sanitizer` at all, which reinstates the warnings this PR removes. Flagging rather than hiding it.

The PR also corrects four statements in the 16.2-to-17.0 migration guide claiming HTML is "stripped before rendering" without a sanitizer. `element.ts:566` writes the content **unchanged** and only warns; `security.md` already words it correctly.

### Two core defects found, deliberately NOT fixed here

The same sweep surfaced two families that **cannot** be fixed in docs. They need separate core tasks:

1. **`passwordRenderer` never forwards the sanitizer.** `handsontable/src/renderers/passwordRenderer/passwordRenderer.ts:40` calls `fastInnerHTML(TD, value)` with two arguments, so the grid's `sanitizer` is never consulted and the warning is unsilenceable by the user. `hashSymbol: '&#9632;'` - documented as supported in the password cell type guide - trips it. The fix is forwarding the sanitizer with a `'password'` context and a per-instance scope, not `fastInnerText`, which would break the documented entity.
2. **The Loading plugin asks users to sanitize core's own markup.** `plugins/loading/loading.ts:295` hands `loadingContent()` output to the Dialog plugin as a **string**, which reaches the sanitizer path at `plugins/dialog/ui.ts:287`. Beyond the spurious warning, a user who *does* set a sanitizer gets core's spinner run through it - the strip-only sanitizer from our own migration guide would destroy it. The fix is returning a `DocumentFragment`, which takes the `appendChild` branch at `ui.ts:291` and bypasses sanitization.

Unrelated third class, also not fixed: `/rows-sorting` emits two pre-existing `intl-date`/`intl-time` invalid-value warnings caused by a header row inside the example data.

### DEMOS-10 is not caused by content in this repository

Recorded here so the Sentry issue is not read as closed by this PR. Both served runner buckets are clean, verified by fetching the manifests and grepping `develop` at the exact commit the `next` bucket was imported from:

| bucket | `docsBranch` | `hotVersion` | entries | string `dateFormat` in executed examples |
|---|---|---|---|---|
| `18.0` | `prod-docs/18.0` | `18.0.0` | 1261 | none |
| `next` | `develop` @ `2c7f881` | `0.0.0-next-2c7f881-20260811` | 1452 | none |

One of the three DEMOS-10 events is `framework: angular`, and no Angular file in this repository passes a string `dateFormat` at all. The remaining candidates are both outside this repo: the runner's own starter templates in `handsontable/examples`, or reader-authored code in the runner editor. DEMOS-10 stays unresolved.

### Date data had to move to ISO 8601 (review fix, commit 4)

Bugbot caught a regression in the `dateFormat` commit. The renderer only takes the `Intl` path for ISO 8601 source data:

```js
const date = parseToLocalDate(value);   // dateRenderer.ts:46
if (date === null) { return BAD_VALUE_TEXT; }
```

`parseToLocalDate` returns `null` for anything failing `ISO_DATE_REGEX`, and these demos feed day-first strings (`'11/10/2020'`, `toLocaleDateString('en-gb')`). Swapping the string `dateFormat` for an `Intl` object therefore turned every date cell into `#bad-value#`. Reproduced against the built core:

```
#bad-value#  <- 11/10/2020   object format, old data
11/10/2020   <- 2020-10-11   object format, ISO data
27/03/2020   <- 2020-03-27   day-first conversion is correct
```

The string form was never formatting anything - it warned and returned the raw value, which is exactly why day-first data *looked* right. So the fix belongs in the data, not the format: 600 date values across six demo apps plus both generators now emit ISO 8601, and `en-GB` 2-digit output reproduces the previous appearance byte for byte.

Two things worth flagging from this:

- **It also repairs five sibling demos that were already broken before this PR.** `ts/demo`, `react-wrapper/demo`, `vue3/demo`, `vue3/basic-example`, and `angular-wrapper/demo` carried a bare `type: 'date'` with no `dateFormat`, which takes the same ISO-only path - so their date columns were rendering `#bad-value#` on the live demo pages already.
- **The Arabic date locale never did anything before this PR.** On `develop` the setting is `dateFormat: isArabicDemoEnabled() ? "M/D/YYYY" : "DD/MM/YYYY"` - a string, which the renderer warns about and returns raw, so no `Intl` formatting was applied in either mode. Commit 5 now makes the date `locale` follow the same `isArabicDemoEnabled()` switch as `layoutDirection` and `language`, which is an improvement rather than a restoration. The visual cost is small: `ar` renders Latin digits with RTL marks (`11‏/10‏/2020`); Arabic-Indic digits would need a region such as `ar-EG`.

The `examples/**` changes in this PR are genuine stale usage of the same removed API, found while sweeping - they are not the source of those events. They live in `examples/next/docs/**` and `examples/prs/**`, which the visual tests do not build from (`visual-tests/scripts/build.mjs` points at `examples/next/visual-tests`), so no screenshot baseline changes.

### How has this been tested?

The console is not trustworthy by default for this class: every missing-sanitizer warning shares one `WeakMap` key per scope object, and `copyPaste` emits a differently worded message on the same key, so one surface can suppress another. Verification therefore asserts the specific `("...")` context string per example on a fresh page load.

- **Browser sweep, before and after.** All 884 page/framework combinations against the local docs dev server. Before: 19 hits across 6 families. After: 11 hits across 2 families, both being the core defects above. After the grid-level rewrite, a sweep restricted to the 24 changed pages reports 0 hits.
- **Positive control.** Temporarily removing `sanitizer` from `context-menu/javascript/example3.js` makes the exact `("contextMenu")` warning reappear; restoring it silences it. This proves the check is sensitive rather than merely quiet.
- **Paste path, end to end.** In a live grid on the fixed page, a `text/html` table pasted alongside a deliberately different `text/plain` payload (`PLAIN-FALLBACK-SINGLE-CELL`) lands as a 2x2 structured paste - so the HTML branch is taken, not the fallback.
- **Sanitizer round-trip.** Each of the six real surface payloads (menu label, both dialog contents, both `<br>` header forms, the checkbox header) sanitizes to a **byte-identical** string, so no rendered output changes. `` and `<a href="javascript:...">` reduce to inert text.
- **Behavior, not just silence.** `<b>Custom option</b>` still renders bold in all four frameworks; the `<br>` headers still break onto two lines (49px header height); dialog example7's `aria-labelledby`/`aria-describedby` still resolve; the dialog example3 Hide button still closes the dialog (verified by a real click in JS and Angular); the `cell-renderer` header checkbox still flips the cell to yellow with `checked="checked"` surviving the sanitizer round-trip.

```bash
npm run docs:lint --prefix docs          # clean
npm run docs:test:plugins --prefix docs  # 243/243 pass
npm --prefix docs/angular-type-check run typecheck
```

The Angular type check reports **0 errors against the local build** (`0 in both, 0 local-only`). Its 9 failures are all `[LATEST-ONLY]`, pre-existing drift against published 17.1.0 types in 8 files that this PR does not touch.

Core lint, build, and the Jest/Puppeteer suites were not run: this PR changes no `handsontable/src` file.

Note on the sweep harness: it reuses one tab per worker, so a late-firing warning can be attributed to the following page. That produced one phantom hit (`/migration-from-13.1-to-14.0`, React) which has zero grids and is silent in isolation. The two core families were identical across both runs.

### Types of changes

### Related issue(s):

1. DEV-2542

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

Documentation and standalone examples only.

### Checklist:

[skip changelog] - documentation and example content only, no package source change.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2542

---

> [!NOTE]
> **Low Risk**
> Documentation and example-only changes; no library package source. Sanitizer helpers are instructional and scoped to demo grids.
> 
> **Overview**
> **Docs examples and guides** now wire a shared **grid-level `sanitizer`** wherever HTML is rendered (context menu labels, dialog content, column headers with `<br>`, checkboxes, or formatting). Helpers allow table markup so **paste** still hits the HTML table path; dangerous tags are removed. The context menu guide adds a warning that HTML `name` labels require `sanitizer` since v18.0.
> 
> **Date demos** stop using string `dateFormat` and use **Intl-style options** plus **`locale`** (`en-GB` / `ar`). Source dates move to **ISO 8601** (`YYYY-MM-DD`) so cells format instead of showing `#bad-value#`. One cell-type example drops a cosmetic `<br>` in a header instead of sanitizing.
> 
> The **16.2→17.0 migration** text is updated: without a custom sanitizer, HTML is **not stripped**—it passes through with a warning (aligned with security docs).
> 
> **Standalone `examples/next/docs` apps** get the same date column config and ISO date constants/generators.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d4a99a4290a2b7094e7c4fc0d3ddc2032f327e90. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [PRO-986 Remove All dependecies from package.json including HF](https://app.clickup.com/t/9015210959/PRO-986)
Task linked: [DEV-2542 Docs: stale example content surfaced by the runtime monitor (dateFormat string, unsanitized contextMenu)](https://app.clickup.com/t/9015210959/DEV-2542)

adrianspdev (approved):

PR #13200, opened 2026-08-19
DEV-2551: Keep fixed rows and columns pinned after undoing an insertion

budnix:
### Context

The PR fixes `fixedRowsBottom` silently dropping to `0` after a row insertion is undone, which un-pins a footer row such as "Overall Totals".

Reported by a pre-sales contact on 18.0.0, and reproducible in every version back to 12.4.0 — not a regression.

**Reproduction (before this PR):**

| | `countSourceRows()` | `fixedRowsBottom` |
| --- | --- | --- |
| initial | 5 | 1 |
| after `alter('insert_row_below', 3, 1)` | 6 | 1 |
| after `getPlugin('undoRedo').undo()` | 5 | **0** ← bug |

**Root cause.** The bug is in the `remove_row` branch of `Core#alter`, not in the undo plugin. The code read `instance.countRows()` *after* `datamap.removeRow()` had already run, then compared the removed index against `totalRows - fixedRowsBottom`. Because `totalRows` was the post-removal count, the boundary of the pinned area sat too high by the number of removed rows, so a row removed *above* the pinned area was treated as one of the pinned rows. The decrement also used `Math.min(groupAmount, fixedRowsBottom)`, which assumed every removed row belonged to the pinned area.

The undo path is only one caller. The same defect is reachable with no undo at all — on a 5-row grid with `fixedRowsBottom: 1`, a plain `hot.alter('remove_row', 3)` un-pins the footer.

**The fix, in two parts:**

1. `handsontable/src/core.ts` — the boundary is now compared against the row count captured *before* the removal, and `fixedRowsBottom` is decreased only by how many removed rows actually overlapped the pinned range.
2. `handsontable/src/plugins/undoRedo/actions/fixedCounts.ts` (new) — a shared helper used by `CreateRowAction.undo` and `CreateColumnAction.undo`. It reads the fixed-area counters before the removal and puts them back when `alter` lowered them. This covers the remaining case where the inserted row or column really was inside the pinned area, so the removal legitimately decreases the counter but undo must still restore the previous state. `RemoveRowAction` and `RemoveColumnAction` already do the same on their side.

The helper covers `fixedRowsTop`, `fixedRowsBottom`, and `fixedColumnsStart`. The column case had the same asymmetry: with `fixedColumnsStart: 2`, inserting a column at index `0` and undoing it left `fixedColumnsStart` at `1`.

Two details worth calling out:

- **Only a decrease is reverted.** `updateSettings({ fixedRowsBottom })` does not clear the undo stack, so a user can raise the value after the insertion. The helper compares the value after `alter` with the value before it and restores only when `alter` lowered it, so a value the user raised in the meantime stays as they set it.
- **The grid is still drawn once.** The counters are restored from the `beforeRender` hook, which runs after `alter` lowers them and before the rendering engine starts. Without that, undo rendered once with the pinned row missing and then again with it back — a visible flicker plus a wasted full render, and `afterRender` listeners read a value that was about to change. `beforeRemoveRow` and `afterRemoveRow` cannot be used for this: both fire inside `datamap.removeRow()` (`dataMap.ts:596` and `dataMap.ts:625`), before `alter` lowers the counters. A fallback restore right after `alter` covers the case where rendering is suspended and the hook does not run.

No public API shape changes.

### How has this been tested?

The failing tests were written first and confirmed red on unmodified source, then green with the fix:

- the core boundary case failed with `Expected 0 to equal 1`;
- the undo cases failed with `Expected 0 to be 1` / `Expected 1 to be 2`;
- the `updateSettings`-after-insert case failed with `Expected 1 to be 3`;
- the column case failed with `Expected 1 to be 2`;
- the single-render case failed with the recorded list `[1, 2]` — one render with the lowered counter, then one with the restored one.

New and updated specs:

- `handsontable/src/__tests__/core/removeRow.spec.js` — cases for `alter('remove_row')`: a bottom pinned row removed, a row above the pinned area removed, a range entirely above the pinned area, a range straddling the boundary, only the topmost pinned row removed, and removals with no index passed (rows taken from the end).
- `handsontable/src/plugins/undoRedo/__tests__/UndoRedo.spec.js` — round-trip cases: undoing an insertion above the pinned area, inside the bottom pinned area, and inside the top pinned area; the same for `fixedColumnsStart`; a counter raised by `updateSettings` after the insertion; and a check that the undo renders the pinned rows only once. The overlay itself is asserted through `getBottomClone()` and `getInlineStartClone()`, not only through `getSettings()`.

Commands:

```bash
npm run test:e2e --prefix handsontable --testPathPattern="(removeRow|removeCol|createRow|createCol|fixedRows|fixedColumns|undoRedo|alter)"
npm run test:unit --prefix handsontable --testPathPattern=undoRedo
```

Results on the latest revision: 438 E2E specs, 0 failures; 6 unit tests, 0 failures; types pass. The full suites (unit 3396, E2E 9612 specs, 0 failures) were run on the first revision of this PR; CI covers them for the later commits.

Manual check in Chrome with the reproduction from the task, using `dev-latest.html` (18.0.0 from the CDN) against `dev-pr.html` (this branch). On the released build the readout reports `fixedRowsBottom = 0` after undo and the "Overall Totals" row scrolls away. On this branch the sequence insert → undo → redo → undo keeps `fixedRowsBottom = 1` at every step, and the row stays in the bottom overlay.

Note: `npm run eslint --prefix handsontable` crashes in my local checkout with `TypeError: minimatch is not a function` in the custom `handsontable/restricted-module-imports` rule. It crashes the same way on files this PR does not touch, so it is a local dependency problem, not a finding about this change. CI is the first full lint pass over the new `fixedCounts.ts`. No CSS was touched, so Stylelint does not apply.

### Types of changes

### Related issue(s):

1. DEV-2551

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2551

### DIFF
--- a/handsontable/src/core.ts
+++ b/handsontable/src/core.ts
@@ -1022,9 +1022,17 @@ export default function Core(
                 // to be compared with the row count from *before* the removal. Rows removed above that
                 // range keep the setting untouched (DEV-2551).
                 const removedRowsCount = totalRowsBefore - totalRows;
+                // With no index passed, `datamap.removeRow` takes the rows from the end, so the removal
+                // starts that many rows before the last one. `calcIndex` points at the last row then, not
+                // at the first removed one.
+                const firstRemovedRowIndex = Number.isInteger(groupIndex)
+                  ? calcIndex
+                  : Math.max(totalRowsBefore - removedRowsCount, 0);
                 const firstFixedRowIndex = totalRowsBefore - fixedRowsBottom;
-                const lastRemovedRowIndex = Math.min(calcIndex + removedRowsCount - 1, totalRowsBefore - 1);
-                const removedFixedRowsCount = lastRemovedRowIndex - Math.max(calcIndex, firstFixedRowIndex) + 1;
+                const lastRemovedRowIndex =
+                  Math.min(firstRemovedRowIndex + removedRowsCount - 1, totalRowsBefore - 1);
+                const removedFixedRowsCount =
+                  lastRemovedRowIndex - Math.max(firstRemovedRowIndex, firstFixedRowIndex) + 1;
 
                 if (removedFixedRowsCount > 0) {
                   tableMeta.fixedRowsBottom -= Math.min(removedFixedRowsCount, fixedRowsBottom);
--- a/handsontable/src/core.ts
+++ b/handsontable/src/core.ts
@@ -989,6 +989,8 @@ export default function Core(
                 groupIndex = Math.max(groupIndex - offset, 0);
               }
 
+              const totalRowsBefore = instance.countRows();
+
               // TODO: for datamap.removeRow index should be passed as it is (with undefined and null values). If not, the logic
               // inside the datamap.removeRow breaks the removing functionality.
               const wasRemoved = datamap.removeRow(groupIndex, groupAmount, source);
@@ -1014,8 +1016,19 @@ export default function Core(
 
               const fixedRowsBottom = tableMeta.fixedRowsBottom;
 
-              if (fixedRowsBottom && calcIndex >= totalRows - fixedRowsBottom) {
-                tableMeta.fixedRowsBottom -= Math.min(groupAmount, fixedRowsBottom);
+              if (fixedRowsBottom) {
+                // Count how many of the removed rows belonged to the bottom fixed rows. Those rows occupied
+                // the `[totalRowsBefore - fixedRowsBottom, totalRowsBefore - 1]` range, so the boundary has
+                // to be compared with the row count from *before* the removal. Rows removed above that
+                // range keep the setting untouched (DEV-2551).
+                const removedRowsCount = totalRowsBefore - totalRows;
+                const firstFixedRowIndex = totalRowsBefore - fixedRowsBottom;
+                const lastRemovedRowIndex = Math.min(calcIndex + removedRowsCount - 1, totalRowsBefore - 1);
+                const removedFixedRowsCount = lastRemovedRowIndex - Math.max(calcIndex, firstFixedRowIndex) + 1;
+
+                if (removedFixedRowsCount > 0) {
+                  tableMeta.fixedRowsBottom -= Math.min(removedFixedRowsCount, fixedRowsBottom);
+                }
               }
 
               if (totalRows === 0) {
--- a/handsontable/src/plugins/undoRedo/__tests__/UndoRedo.spec.js
+++ b/handsontable/src/plugins/undoRedo/__tests__/UndoRedo.spec.js
@@ -1862,6 +1862,67 @@ describe('UndoRedo', () => {
         expect(getBottomClone().find('tbody tr').length).toBe(3);
       });
 
+      it('should render the bottom fixed rows only once while undoing an insertion within them', async() => {
+        handsontable({
+          data: createSpreadsheetData(5, 2),
+          colHeaders: true,
+          rowHeaders: true,
+          fixedRowsBottom: 2,
+        });
+
+        await alter('insert_row_below', 4, 1);
+
+        const fixedRowsBottomPerRender = [];
+
+        addHook('afterRender', () => {
+          fixedRowsBottomPerRender.push(getSettings().fixedRowsBottom);
+        });
+
+        getPlugin('undoRedo').undo();
+
+        expect(fixedRowsBottomPerRender).toEqual([2]);
+      });
+
+      it('should keep the number of start fixed columns after undoing an insertion within them', async() => {
+        handsontable({
+          data: createSpreadsheetData(2, 5),
+          colHeaders: true,
+          rowHeaders: true,
+          fixedColumnsStart: 2,
+        });
+
+        await alter('insert_col_start', 0, 1);
+
+        expect(countCols()).toBe(6);
+
+        getPlugin('undoRedo').undo();
+
+        expect(countCols()).toBe(5);
+        expect(getSettings().fixedColumnsStart).toBe(2);
+        expect(getInlineStartClone().find('tbody tr:eq(0) td').length).toBe(2);
+      });
+
+      it('should restore the number of start fixed columns changed after the insertion', async() => {
+        handsontable({
+          data: createSpreadsheetData(2, 5),
+          colHeaders: true,
+          rowHeaders: true,
+          fixedColumnsStart: 1,
+        });
+
+        await alter('insert_col_start', 0, 1);
+
+        await updateSettings({
+          fixedColumnsStart: 3,
+        });
+
+        getPlugin('undoRedo').undo();
+
+        expect(countCols()).toBe(5);
+        expect(getSettings().fixedColumnsStart).toBe(3);
+        expect(getInlineStartClone().find('tbody tr:eq(0) td').length).toBe(3);
+      });
+
       it('should keep the number of top fixed rows after undoing an insertion within them', async() => {
         handsontable({
           data: createSpreadsheetData(5, 2),
--- a/handsontable/src/plugins/undoRedo/__tests__/UndoRedo.spec.js
+++ b/handsontable/src/plugins/undoRedo/__tests__/UndoRedo.spec.js
@@ -1838,6 +1838,28 @@ describe('UndoRedo', () => {
 
         expect(countRows()).toBe(5);
         expect(getSettings().fixedRowsBottom).toBe(2);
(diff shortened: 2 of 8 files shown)


qunabu:
Task linked: [DEV-2551 Undoing a row insertion resets fixedRowsBottom from 1 to 0. The row count is restored correctly, but the setting is decremented one step too far, so the pinned bottom row silently un-pins.](https://app.clickup.com/t/9015210959/DEV-2551)

handsontable/src/plugins/undoRedo/actions/createRow.ts
```diff
@@ -52,6 +73,19 @@ export class CreateRowAction extends BaseAction {
...
+
+    const settings = hot.getSettings();
+
+    // Rows inserted into the fixed area belong to that area when the undo removes them, so `alter`
+    // legitimately decreases the counters. Restore the values captured before the insertion so the undo
+    // brings back the whole previous state, including the pinned rows.
+    if (settings.fixedRowsBottom !== this.fixedRowsBottom || settings.fixedRowsTop !== this.fixedRowsTop) {
+      // Changing by the reference as `updateSettings` doesn't work the best.
+      settings.fixedRowsBottom = this.fixedRowsBottom;
+      settings.fixedRowsTop = this.fixedRowsTop;
```
demtario: The restore is unconditional, so it also overwrites a value the user set *after* the insertion.

`updateSettings({ fixedRowsBottom })` does not clear the undo stack — `clear()` is only reached from `disablePlugin`, `destroy`, and `loadData`, and `BasePlugin.onUpdateSettings` leaves an already-enabled plugin enabled. So this is reachable:

```js
hot.alter('insert_row_below', 1, 1);          // action stores fixedRowsBottom: 1
hot.updateSettings({ fixedRowsBottom: 3 });   // user's explicit, later intent
hot.getPlugin('undoRedo').undo();             // alter() correctly leaves 3 -> this block forces it back to 1
```

The `!==` guard fires in both directions, so the stale captured value wins over the newer one. It also fires when the removal never happened at all (a `beforeRemoveRow` handler returned `false`).

Worth saying: this is not invented here — `RemoveRowAction.undo` already restores unconditionally. The point is only not to extend the hazard to a second action.

Two ways out. The discriminating case is: stored `1`, user sets `3`, then undo removes a row that *was* inside the pinned area.

1. Keep the new fields and restore only when `alter` actually **decreased** the value. Smallest diff, keeps the `afterUndo` shape this PR adds. Leaves `2` in the discriminating case.
2. Read both counts from `settings` immediately before `hot.alter(...)` inside `undo()`, and restore only on a decrease. Restores `3` — strictly more correct — and makes the two stored fields, the constructor signature change, the `afterUndo` shape change, and the `createRow.spec.js` edit all unnecessary.

Either way every new spec here stays green: for an insertion above the pinned area `alter` no longer changes the counters, so no restore runs at all; for an insertion inside the pinned area the pre-`alter` value is the same value the stored field holds.

handsontable/src/plugins/undoRedo/__tests__/UndoRedo.spec.js
```diff
@@ -1803,6 +1803,61 @@ describe('UndoRedo', () => {
...
+
+        await alter('insert_row_below', 3, 1);
+
+        expect(countRows()).toBe(6);
+        expect(getSettings().fixedRowsBottom).toBe(1);
+
+        getPlugin('undoRedo').undo();
+
+        expect(countRows()).toBe(5);
+        expect(getSettings().fixedRowsBottom).toBe(1);
```
demtario, line 1822: Every new assertion across the three spec files checks `getSettings()`, so the user-visible symptom is asserted nowhere. The reported defect is that the "Overall Totals" row *scrolls away* — it leaves the bottom overlay. A future regression that keeps the number correct but fails to re-render the overlay would pass all of these, including the `hot.render()` call added in `CreateRowAction.undo`, which currently has no assertion covering it.

`TESTING.md` puts overlays firmly in E2E and asks whether a test would catch a real bug in the code it covers. One assertion on the bottom clone closes it, using the pattern already in `src/__tests__/settings/fixedRowsBottom.spec.js`:

```js
expect(getBottomClone().find('tbody tr').length).toBe(1);
```

Second gap: no test covers the case from my other comment — the setting changed via `updateSettings` after the insertion and then clobbered by the undo.

The *placement* of the new cases is right, to be clear: `TESTING.md` prescribes adding a case to the closest existing spec for a bug fix, and the frozen-suite rule blocks new `*.spec.js` files, not new cases in existing ones.

demtario (commented):
Reviewed the fix against `develop`. The `core.ts` boundary arithmetic checks out: I extracted the formula and fuzzed it against a reference model that tracks pinned-row identity — 3846 single-group cases (`N` 1-9, `k` 0-`N`, index `undefined`/`null`/0..N-1, amount 1..N+2) and 8298 two-group cases through `normalizeIndexesGroup` plus the `offset` accumulator, with 0 mismatches. The pre-PR formula fails 3060 of the same two-group cases, so the harness discriminates rather than confirming itself.

Three things worth noting in favor of the change beyond what the description claims:

- It provably cannot strand `fixedRowsBottom > countRows()` — the failure mode from the 8.x changelog. Removing only unpinned rows leaves at least `k` rows, and removing pinned rows decrements, so `k <= countRows()` holds.
- `removedRowsCount = totalRowsBefore - totalRows` is safe under trimming: `DataMap#removeRow` maps a contiguous *visual* run via `visualRowsToPhysical`, and visual space is dense, so the removed rows always occupy `[calcIndex, calcIndex + removedRowsCount - 1]`.
- It is strictly better than the old `groupAmount` in two more cases: it counts the NestedRows child cascade that `beforeRemoveRow` adds, and it is a no-op when nothing was actually spliced (out-of-range index, hook-emptied list), where the old code decremented anyway.

Two findings inline. Neither is a merge gate — the PR is a net improvement either way. A few smaller points (the restore block now duplicating `RemoveRowAction`'s, the provably dead `Math.min` on the new `core.ts` line, `hot.render()` not pairing with `view.adjustElementsSize()` the way `alter` does, and the changelog title under-scoping the fix to the undo path) I can add separately if useful.

handsontable/src/core.ts
```diff
@@ -1014,8 +1016,19 @@ export default function Core(
...
-                tableMeta.fixedRowsBottom -= Math.min(groupAmount, fixedRowsBottom);
+              if (fixedRowsBottom) {
+                // Count how many of the removed rows belonged to the bottom fixed rows. Those rows occupied
+                // the `[totalRowsBefore - fixedRowsBottom, totalRowsBefore - 1]` range, so the boundary has
+                // to be compared with the row count from *before* the removal. Rows removed above that
+                // range keep the setting untouched (DEV-2551).
+                const removedRowsCount = totalRowsBefore - totalRows;
+                const firstFixedRowIndex = totalRowsBefore - fixedRowsBottom;
+                const lastRemovedRowIndex = Math.min(calcIndex + removedRowsCount - 1, totalRowsBefore - 1);
+                const removedFixedRowsCount = lastRemovedRowIndex - Math.max(calcIndex, firstFixedRowIndex) + 1;
```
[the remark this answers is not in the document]
budnix: You're right — fixed in bd26426. Rows removed with no index passed are now counted from `totalRowsBefore - removedRowsCount`, and two specs in `removeRow.spec.js` cover it.

handsontable/src/plugins/undoRedo/actions/createRow.ts
budnix: You're right — went with option 2 in 04b6082. The counts are read right before `alter` and restored only on a decrease, so the two fields, the constructor change, the `afterUndo` shape change and the `createRow.spec.js` edit are all reverted. 7ef3f23 moves the restore into `beforeRender` (one render instead of two) and reuses the same helper for `fixedColumnsStart`, which had the same asymmetry.

handsontable/src/plugins/undoRedo/__tests__/UndoRedo.spec.js
budnix: You're right — 04b6082 adds the `getBottomClone()` assertion and the `updateSettings`-after-insert case (red with `Expected 1 to be 3` before the fix). 7ef3f23 adds the same pair for `getInlineStartClone()` plus a check that the undo renders only once.

demtario (approved):

Issue #8872, opened 2021-10-18
Autocomplete/dropdown doesn't show all options in cells with Flexbox


adrianszymanski89:
### Description
Autocomplete/dropdown doesn't show all options in cells with Flexbox.

### Steps to reproduce

1. Open the editor and try to chose different option in "Country" column.

### Demo

https://codesandbox.io/s/handsontable-autocomplete-issue-cvnw9?file=/src/index.js - v.10.0.0

### Your environment
* Handsontable version: 10
* Browser Name and version: Chrome 94
* Operating System: MacOS Big Sur 11.5

Inform: https://forum.handsontable.com/t/autocomplete-is-not-visualized-correctly/5585

budnix:
Checked with newest HoT version and seems working fine https://codesandbox.io/p/devbox/handsontable-autocomplete-issue-forked-9nqrh8?workspaceId=ws_LpHf7fyvdzwhT1jqHWpTgc

PR #13202, opened 2026-08-19
DEV-1656: Keep the autocomplete dropdown visible in short flex layouts

budnix:
### Context

The PR fixes the autocomplete/dropdown option list collapsing to zero height when the grid is short, so the user sees no options at all.

DEV-1656 asked to verify and fix GitHub issue #8872 ("Autocomplete/dropdown doesn't show all options in cells with Flexbox"). Two separate findings:

**1. The originally reported reproduction no longer fails.** The reported [CodeSandbox](https://codesandbox.io/s/handsontable-autocomplete-issue-cvnw9) combines `body { display: flex }` with `.wtHolder { width: unset !important; height: unset !important }` — user CSS that overrides Handsontable's own holder sizing. Rebuilt locally and measured:

| Build | Dropdown list height | Options visible |
|---|---|---|
| v10.0.0 (as reported) | 23px | 1 of 5 |
| v18.0.0 / `develop` | 148px | 5 of 5 |

So that specific layout was fixed somewhere between v10 and v18. GitHub issue #8872 was closed on the same day DEV-1656 was created — that was a migration to internal tracking, not a fix commit.

**2. A live defect on the same code path still hides every option.** `AutocompleteEditor#limitDropdownIfNeeded()` trims the option list to whole rows that fit the free space below the edited cell:

```js
do {
  lastRowHeight = this.htEditor.stylesHandler.getDefaultRowHeight() ?? 0;
  tempHeight += lastRowHeight;
} while (tempHeight < spaceAvailable);

height = tempHeight - lastRowHeight;
```

When `spaceAvailable` is not taller than one row, the loop runs a single pass and `height` becomes **0**. The list renders as a ~2px sliver, so every choice is hidden — the exact symptom #8872 describes. It needs no `!important` override: a plain flex parent that squeezes the grid is enough, which is the layout the issue is about.

Measured on released v18.0.0, a `flex-direction: column` parent with a 5-option autocomplete, editing the first row:

| Grid height | Dropdown height set | Options visible |
|---|---|---|
| 70px | 0 | none |
| 78px | 0 | none |
| 87px | 0 | none |
| 91px | 29 | 1, scrollable |
| 120px | 58 | 2, scrollable |

Two further problems in the same block: `setDropdownHeight()` recomputed `tempHeight - lastRowHeight` instead of reusing the already-computed `height`, and a default row height of `0` made the loop spin forever.

**The fix.** The list is clamped to one option at minimum and stays scrollable to reach the rest — the same floor the newer MultiSelect editor's dropdown already applies (`Math.max(..., 1)` in `dropdownController.updateDimensions()`). The loop is replaced with equivalent arithmetic:

```js
const rowsThatFit = Math.max(Math.ceil(spaceAvailable / rowHeight) - 1, 1);
const height = rowsThatFit * rowHeight;
```

`Math.ceil(...) - 1` reproduces the old "add rows until one crosses the boundary, then step back a row" result exactly for every `spaceAvailable` above one row height, so the existing trimming behavior is unchanged — including the deliberate margin that leaves the last row out when the space fits exactly. Only the previously-zero cases move, and the `rowHeight === 0` early return removes the infinite loop.

**Known limitation, not fixed here.** The grid's root element gets `overflow: clip` whenever a `height` is set (`core.ts`), and the dropdown's holder is inside that root, so anything the list renders past the grid's bottom edge is cut off. When the free space is narrower than the row this fix forces, that row is only partly readable — and at zero free space it is clipped away entirely. Measured on this build at row height 30:

| Space below the cell | Option pixels on screen (of 30) |
|---|---|
| 0 | 0 |
| 15 | 14 |
| 29 (one whole list row) | 28 |

So the fix guarantees the editor stops sizing the list to zero, and the option is readable as soon as there is a row's worth of space. Making it readable in the tighter extremes needs the dropdown to escape the clipping root — a larger change (it would touch positioning, focus, and a11y) that deserves its own task.

### How has this been tested?

New Playwright E2E spec `tests/e2e/dropdown-height.spec.ts` with the fixture `tests/fixtures/demo/dropdown-height.html`.

The fixture sizes its flex parent to one rendered row plus one list row, both measured from what the active theme rendered (main/horizon/classic differ, and the two heights differ from each other by the 1px first-row border compensation). That makes the free space below the cell exactly one list row: the boundary where the old arithmetic returned 0, and the only squeeze where the forced row also clears the grid's clipping edge. Column headers are off so the edited cell starts at y=0, which keeps `spaceAbove` at 0 so the dropdown cannot flip and the "space below" branch stays under test.

The visibility assertions intersect the option row with **every** clipping ancestor. Measuring against the list's own holder is not enough — the holder can hang past the clipping root, and Playwright's `toBeVisible()` only needs a non-empty bounding box, so both call a clipped-away option visible. An earlier revision of this spec did exactly that and passed on the unfixed build; the current one does not.

```bash
cd tests && npx playwright test e2e/dropdown-height.spec.ts
```

Verified by reverting the source, rebuilding, and re-running all six legs: **18 of 24 failed** — the three flex tests on every theme × bundle leg — while the 6 control tests stayed green, which is what shows the fixture itself is sound. With the fix, 24/24 pass.

| Test | On the unfixed build | With the fix |
|---|---|---|
| the editor no longer sizes the list to zero | list height `0`, expected `>= 29` | pass |
| the first option is really on screen | `0px` on screen, expected `>= 28` | 28 of 30 px |
| the last option is reachable from the keyboard | `0px` on screen after arrowing | 26 of 29 px |
| normal-height grid keeps showing all options | pass (control) | pass |

Regression runs, all green:

```bash
npm run test:e2e --prefix handsontable --testPathPattern=autocompleteEditor            # 138 specs, 0 failures
npm run test:e2e --prefix handsontable --testPathPattern="handsontableEditor|multiSelectEditor|dropdownEditor"  # 193 specs, 0 failures
npm run eslint --prefix handsontable                                                  # 0 errors
npm run test:types --prefix handsontable                                              # pass
cd tests && npx playwright test --project=e2e-main e2e/dropdown-width.spec.ts e2e/editor-window-scroll.spec.ts
cd tests && npx playwright test e2e/dropdown-height.spec.ts                            # 24 passed (3 themes x 2 bundles)
npm run test:unit --prefix handsontable --testPathPattern="autocomplete|editor"        # 3396 tests, 0 failures
```

One branch is deliberately not pinned by the new tests: the fixture turns column headers off so the dropdown cannot flip upwards, which keeps the "space below" branch under test. The flipped branch gets the same one-row floor, and the existing Jasmine flip tests ("display the dropdown above the editor", "limit the list to the space size left above the editor") pass — but a flipped dropdown with under one row of space above is unexercised. There it now renders one option just above the workspace edge instead of nothing at all.

The two legacy specs that pin this arithmetic — "should limit the list to the space size left below/above the editor" — sit on exact row-height multiples and still pass, which is the guard that the non-collapsed trimming behavior did not shift.

Manual verification used a three-case demo page (`handsontable/dev-latest.html` vs `handsontable/dev-pr.html`, both gitignored):

1. The reported CodeSandbox CSS verbatim — all 5 options on both builds (already fixed).
2. A `flex-direction: column` parent squeezing the grid to 78px — **no list at all** on v18.0.0, one option plus a scrollbar on this build.
3. A non-flexbox control at `height: 450` — all 5 options, unchanged on both builds.

One note for whoever runs the suite locally: `playwright.config.ts` sets `reuseExistingServer`, so if another checkout of this repo on the same machine is already serving port 8123, every spec here fails with `net::ERR_CONNECTION_REFUSED` or a blank fixture. That is what it looks like, not a flake in this spec.

### Types of changes

### Related issue(s):

1. DEV-1656
2. #8872

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-1656

---

> [!NOTE]
> **Low Risk**
> Localized bug fix in autocomplete dropdown height math with E2E coverage; trimming behavior above one row height is preserved and existing editor tests pass.
> 
> **Overview**
> Fixes **#8872** / DEV-1656: when a flex parent squeezes the grid, the autocomplete option list could be sized to **0px** and hide every choice.
> 
> **`limitDropdownIfNeeded()`** replaces the old do/while row loop with `Math.max(Math.ceil(spaceAvailable / rowHeight) - 1, 1)` so trimming still stops one row short of the boundary, but **at least one row** is always shown (aligned with MultiSelect’s dropdown). A **`rowHeight === 0`** guard avoids an infinite loop.
> 
> **Tests:** Playwright E2E (`dropdown-height.spec.ts`) with a flex-squeezed fixture—list height, real on-screen option height (all clipping ancestors), keyboard reach to the last option, and a tall-grid regression.
> 
> **Not in scope:** options can still be partly clipped by the grid root’s `overflow: clip` when free space is below one row height; escaping that root is tracked separately (DEV-1656).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1050c5748a1acf733e198375c4955cd519bbafb3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
+++ b/handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
@@ -428,23 +428,27 @@ export class AutocompleteEditor extends HandsontableEditor {
     const dropdownHeight = this.getDropdownHeight();
 
     if (dropdownHeight > spaceAvailable) {
-      let tempHeight = 0;
-      let lastRowHeight = 0;
-      let height = 0;
+      const rowHeight = this.htEditor.stylesHandler.getDefaultRowHeight() ?? 0;
 
-      do {
-        lastRowHeight = this.htEditor.stylesHandler.getDefaultRowHeight() ?? 0;
-        tempHeight += lastRowHeight;
-      } while (tempHeight < spaceAvailable);
+      if (rowHeight === 0) {
+        return;
+      }
 
-      height = tempHeight - lastRowHeight;
+      // Fit whole rows only. `Math.ceil(...) - 1` reproduces the original "keep adding rows until
+      // one crosses the boundary, then step back a row" arithmetic, but always keeps at least one
+      // row visible. Without that floor the height collapses to 0 whenever the free space is not
+      // taller than a single row, which renders the list as an invisible sliver and hides every
+      // choice - the flexbox-squeezed grids reported in #8872. The MultiSelect editor's dropdown
+      // clamps to one entry the same way.
+      const rowsThatFit = Math.max(Math.ceil(spaceAvailable / rowHeight) - 1, 1);
+      const height = rowsThatFit * rowHeight;
 
       if (this.isFlippedVertically) {
         this.htEditor.rootElement.style.top =
           `${parseInt(this.htEditor.rootElement.style.top, 10) + dropdownHeight - height}px`;
       }
 
-      this.setDropdownHeight(tempHeight - lastRowHeight);
+      this.setDropdownHeight(height);
     }
   }
 
--- a/handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
+++ b/handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
@@ -434,12 +434,21 @@ export class AutocompleteEditor extends HandsontableEditor {
         return;
       }
 
-      // Fit whole rows only. `Math.ceil(...) - 1` reproduces the original "keep adding rows until
-      // one crosses the boundary, then step back a row" arithmetic, but always keeps at least one
-      // row visible. Without that floor the height collapses to 0 whenever the free space is not
-      // taller than a single row, which renders the list as an invisible sliver and hides every
-      // choice - the flexbox-squeezed grids reported in #8872. The MultiSelect editor's dropdown
-      // clamps to one entry the same way.
+      // Show whole rows only, and stop one row short of the boundary: `Math.ceil(...) - 1` is the
+      // exact arithmetic of the do/while this replaced ("add rows until one crosses the free
+      // space, then step back a row"), so an exactly-fitting space still leaves its last row out.
+      // That margin is deliberately preserved - the list is trimmed because it overflows the
+      // workspace, and the rendered rows carry a border the raw row height does not.
+      //
+      // `Math.max(..., 1)` is the fix: without it the height collapsed to 0 whenever the free
+      // space was not taller than a single row, rendering the list as an invisible sliver that hid
+      // every choice - the flexbox-squeezed grids reported in #8872. The MultiSelect editor's
+      // dropdown clamps to one entry the same way (`dropdownController.updateDimensions()`).
+      //
+      // A caveat this cannot solve here: the grid's root element gets `overflow: clip` whenever a
+      // `height` is set, so when the free space is narrower than the forced row, that row is
+      // partly clipped by the grid's bottom edge - fully so when the space reaches 0. Making it
+      // readable in those extremes needs the dropdown to escape the clipping root (DEV-1656).
       const rowsThatFit = Math.max(Math.ceil(spaceAvailable / rowHeight) - 1, 1);
       const height = rowsThatFit * rowHeight;
 
--- a/handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
+++ b/handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
@@ -449,6 +449,12 @@ export class AutocompleteEditor extends HandsontableEditor {
       // `height` is set, so when the free space is narrower than the forced row, that row is
       // partly clipped by the grid's bottom edge - fully so when the space reaches 0. Making it
       // readable in those extremes needs the dropdown to escape the clipping root (DEV-1656).
+      //
+      // No border compensation here, unlike `getTargetDropdownHeight()`'s `getTableHeight() + 1`.
+      // Adding it was measured and changes nothing a user sees: the clipping root, not the list's
+      // own budget, is what bounds the visible row, so the extra pixel only pushes the holder
+      // further past the clip (main 31->32px holder, 28px of option visible either way; classic
+      // 28->29 and 25; horizon 37->38 and 37).
       const rowsThatFit = Math.max(Math.ceil(spaceAvailable / rowHeight) - 1, 1);
       const height = rowsThatFit * rowHeight;
 
--- a/tests/fixtures/pages/DropdownHeightPage.ts
+++ b/tests/fixtures/pages/DropdownHeightPage.ts
@@ -1,0 +1,134 @@
+import { type Page, type Locator, expect } from '@playwright/test';
+
+/**
+ * Page Object for the dropdown-height fixture (#8872 / DEV-1656): an autocomplete
+ * dropdown opened in a grid that a flexbox parent squeezed to 1.5 rows, plus a
+ * normal-height control grid. Encapsulates opening the list, reading the theme's
+ * rendered row height, and measuring the list box.
+ */
+export class DropdownHeightPage {
+  readonly page: Page;
+  readonly theme: string;
+  readonly bundle: string;
+
+  constructor(page: Page, theme = 'main', bundle = 'umd') {
+    this.page = page;
+    this.theme = theme;
+    this.bundle = bundle;
+  }
+
+  /**
+   * Navigate and wait for both grids to render (a real DOM condition, no sleep).
+   */
+  async goto(): Promise<void> {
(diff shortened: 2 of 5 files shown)


qunabu:

tests/fixtures/pages/DropdownHeightPage.ts
```diff
@@ -0,0 +1,168 @@
...
+  cell(row: number, col: number, gridTestId = 'grid'): Locator {
+    return this.page.getByTestId(gridTestId).locator('.ht_master').getByTestId(`cell-${row}-${col}`);
+  }
+
+  /**
+   * Open the autocomplete editor on a cell and wait for the list to show.
+   */
+  async openDropdownAt(row: number, col: number, gridTestId = 'grid'): Promise<void> {
+    await this.cell(row, col, gridTestId).dblclick();
+    await expect(this.dropdownList(gridTestId)).toBeAttached();
```
demtario: `toBeAttached()` resolves too early. `open()` constructs and `init()`s the inner grid synchronously with `startRows: 0`, so `.htCore` is attached and empty; everything that sizes the list runs in the deferred `_registerTimeout` -> `queryChoices()` (`autocompleteEditor.ts:265`). At that moment the holder measures 1px, so under CI load `listHeight()` can read 1 vs 29 and the boundary test goes red.

Suggest gating on a rendered option instead: `await expect(this.options(gridTestId).first()).toBeAttached();`

handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
```diff
@@ -428,23 +428,36 @@ export class AutocompleteEditor extends HandsontableEditor {
...
+      // `Math.max(..., 1)` is the fix: without it the height collapsed to 0 whenever the free
+      // space was not taller than a single row, rendering the list as an invisible sliver that hid
+      // every choice - the flexbox-squeezed grids reported in #8872. The MultiSelect editor's
+      // dropdown clamps to one entry the same way (`dropdownController.updateDimensions()`).
+      //
+      // A caveat this cannot solve here: the grid's root element gets `overflow: clip` whenever a
+      // `height` is set, so when the free space is narrower than the forced row, that row is
+      // partly clipped by the grid's bottom edge - fully so when the space reaches 0. Making it
+      // readable in those extremes needs the dropdown to escape the clipping root (DEV-1656).
+      const rowsThatFit = Math.max(Math.ceil(spaceAvailable / rowHeight) - 1, 1);
```
demtario, line 458: The floor uses `getDefaultRowHeight()`, which is the rendered row height minus 1, but the rendered row plus its border is ~2px taller. So the clamped list overflows the space it was just clamped to, and the grid root's `overflow: clip` cuts the one option this fix exists to show. Measured on this build: main 28 of 30px visible (holder 31px in 29px of space), classic 25 of 27.

`getTargetDropdownHeight()` already does `getTableHeight() + 1` for the same border compensation. With that applied, `dropdown-height.spec.ts:42` could assert `>= listRowHeight` with real slack - today it passes exactly on the boundary (28 vs 28 on main, 25 vs 25 on classic), so one pixel of theme drift turns it red.

tests/fixtures/pages/DropdownHeightPage.ts
budnix: Nice catch — that's a real race, and I'd missed that `startRows: 0` means `.htCore` is attached and empty until the deferred `queryChoices()` runs. Took your suggestion verbatim: `openDropdownAt()` now waits for a rendered option, which only exists once `updateChoicesList()` has loaded the choices and sized the list in the same synchronous block. Fixed in 1050c5748a.

handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
budnix: You're right about the assertion — 28 vs 28 on main and 25 vs 25 on classic is zero slack, and that would have gone red on a pixel of drift. Fixed in 1050c5748a: the visibility asserts now require a clear majority of the row (main 28 vs 23.2, classic 25 vs 20.8, horizon 37 vs 29.6, against 0 on the unfixed build).

On the `getTargetDropdownHeight()`-style `+ 1` though — I tried it and measured before/after on all three themes, and it buys no visible pixels:

| theme | holder without `+1` | with `+1` | option px visible without | with |
|---|---|---|---|---|
| main | 31 | 32 | 28 | 28 |
| classic | 28 | 29 | 25 | 25 |
| horizon | 37 | 38 | 37 | 37 |

The binding constraint is the clipping root (`spaceAvailable`), not the list's own budget, so the extra pixel only pushes the holder further past the clip. Your diagnosis that the clamped list overflows the space it was just clamped to is correct — it's just that shrinking it to fit would cost visible pixels rather than gain them. Left the clamp alone and recorded the measurement in the comment so nobody re-derives it.

demtario (approved):

PR #13203, opened 2026-08-19
DEV-1968: Build the Component Source artifact on official releases

budnix:
### Context

The PR fixes the `Component Source` workflow, which never built its artifact for an official release. The deliverable was silently missing for every stable version, including 18.0.0.

There were two independent causes.

**1. The trigger could not match a stable release.** The workflow waited on `workflow_run` from `Publish` with `branches: ['release/**']`. A `Publish` run's `head_branch` is a branch only for RC pushes (`release/18.0.0`) — for a `release` event it carries the **tag** (`18.0.0`), which never matches that allow-list.

Evidence from the 18.0.0 release:

| `Publish` run | Event | `head_branch` | Conclusion | Component Source ran? |
|---|---|---|---|---|
| `28431654718` (06-30 08:41) | `release` | `18.0.0` | success | **no** |
| `27610213255` (06-16 10:09) | `release` | `18.0.0-rc` | success | **no** |
| `28430639792` (06-30 08:22) | `push` | `release/18.0.0` | cancelled | created, then skipped |

`Component Source` has only ever had 10 runs. The last automatic one was 06-30 at 08:33 (skipped). The stable release finished at 08:41 and produced nothing.

**2. The release's own push could not rescue it.** `stable-prepare` commits the version bump and pushes it to `release/18.0.0` (`publish.yml:1161`), which is a valid `Publish` trigger — but that push uses `secrets.GITHUB_TOKEN`, and GitHub never starts a workflow run from a `GITHUB_TOKEN` push. Confirmed in the data: the last push-event `Publish` run was 08:22:50, before the release.

Separately, the failure in the linked run was a jsDelivr `503` (`curl: (22)`). That has a specific cause: `stable-publish` purges the jsDelivr cache right after publishing (`publish.yml:1572` → `purge-jsdelivr-cache.mjs`), leaving the CDN cold exactly when this job runs. The old `curl --retry 3 --retry-delay 2` gave up after about 6 seconds and wrote a 455-byte error page.

**What this PR changes**

- **Trigger** — the ref filter is removed entirely and the job's `if` is the only gate. A filter is what dropped the deliverable, and whether a `branches`/`branches-ignore` filter passes a *tag*-valued `head_branch` is unspecified (tag filtering for `workflow_run` is not supported), so any filter here risks the same silent miss. Excluding develop was already redundant: publish.yml's `push` trigger covers `release/**` only and `release` is tag-based, so a develop `head_branch` can only arrive as `workflow_run` or `workflow_dispatch`, and neither passes the `if`.
- **Version** — taken from the **triggering run's own `npm_packs_<version>` artifact**, one rule for stable, first RC and later RCs alike. Every publish path uploads the exact tarballs it publishes under that name (`publish.yml:203`, `339`, `655`, `1430`, `1495`), and the publish steps download that same artifact to run `npm publish`, so the name is a per-run record of what the run actually shipped. Every other candidate oracle drifts: `hot.config.js` at the run's commit still holds the *previous* version (which is why RC artifacts were one RC stale — the run that published `rc5` uploaded `componentsource-18.0.0-rc4`); npm's `rc` dist-tag is one mutable global pointer shared across release lines and can lag the publish; and the highest `<line>-rcN` git tag is written in `rc-build`, *before* the `approvers` gate on `rc-publish`, with no concurrency group on `Publish` — so a later push can tag `rcN+1` while `rcN` still awaits approval.
- **Download** — the npm registry first, jsDelivr as fallback. npm is the origin jsDelivr itself pulls from, so it is ready the moment `Publish` finishes and is unaffected by the cache purge.
- **Guards** — waits for the version to be readable on npm, rejects a bundled file under 500 KB (which is what the original 455-byte error page was), and sets `if-no-files-found: error`.
- **No-op when a green run published nothing** — a push to `release/**` can leave `Publish` green having shipped nothing, because `rc-check` sets `allowed=false` once the line has a stable release and only the skipped jobs depend on it. No separate guard is needed: such a run uploads no `npm_packs_*` artifact, so the step emits a `::notice::`, sets `should-build=false`, and the run stays green with nothing uploaded. The manual dispatch bypasses this by design.
- **Loud failures** — an unresolvable or empty version fails the job with `::error::` on every path. A green run that quietly produces no artifact is invisible, and that invisibility is how this bug survived two releases.
- **Escape hatch** — an optional `version` input, so a missed release can still be packaged by hand.

No library source is touched — the diff is one CI workflow file (`.github/workflows/componentsource.yml`), so the commit carries a `Refactor-only:` trailer and the test-presence gate passes. `workflow_run` workflows only ever run from the default branch, so this cannot be exercised on a PR branch; it was validated by replaying the new logic against real production data instead.

- Unit tests added/modified (`*.unit.js`): none — no library source changed
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`): none — no library source changed
- For a bug fix — the spec that fails without this fix: not expressible as a repo spec (the defect is a GitHub Actions trigger filter). Covered instead by the trigger replay below, which fails on the old workflow and passes on the new one.
- Demo page / recorded trace (for UI changes): n/a — no UI change

Three validation suites were run locally. Each extracts the real steps from the workflow YAML, so it tests the shipped file rather than a copy.

**1. Trigger layer — replay against all 282 real `Publish` runs** (fetched from the Actions API), comparing the old ref filter plus job `if` against the new unfiltered trigger:

```
PASS  stable release 18.0.0 job did NOT start under the old trigger
PASS  stable release 18.0.0 job DOES start under the new trigger
PASS  every green stable-release run starts the job (n=3)
PASS  none of them started it before
PASS  RC release-branch pushes still reach the job (n=39) — no trigger regression
PASS  every develop Publish run is workflow_run or workflow_dispatch (n=104)
PASS  no develop Publish run is a release event or a push
PASS  experimental develop publishes still never build, with NO ref filter
PASS  no non-green Publish run ever reaches the job (n=109)
PASS  only release events and release-branch pushes reach the job
ALL ASSERTIONS PASSED

Cost of dropping the ref filter: 236 of 282 historical Publish runs
would have created a Component Source run that skips immediately (no runner time).
```

The two develop assertions are what justify removing the filter rather than narrowing it. Note the limit of this replay: it validates the *job `if`*, which is real logic. It cannot validate GitHub's own ref-filter behavior on a tag-valued `head_branch` — which is precisely why there is no filter left to validate.

**2. Version resolution, no-op and failure paths** — 21 cases against the extracted `Resolve published version` step. Fake cases shadow `gh` on `PATH` to serve a chosen artifact set. The last five hit the **live Actions API for real historical `Publish` runs**, which is the strongest evidence available short of doing a release:

```
--- the version is whatever THIS Publish run uploaded as npm_packs_* ---
PASS  stable release run (the DEV-1968 case)               -> 18.0.0
PASS  first-RC run                                         -> 18.0.0-rc1
PASS  later-RC run                                         -> 18.0.0-rc5
PASS  double-digit RC needs no numeric sorting             -> 18.0.0-rc10

--- concurrent runs / tag races can no longer mislead us ---
PASS  run shipped rc5 while rc6 tag exists                 -> 18.0.0-rc5
PASS  run shipped rc6                                      -> 18.0.0-rc6
PASS  17.1.1 run while 18.1.0 RCs exist                    -> 17.1.1-rc1

--- green Publish that published nothing ---
PASS  green run, no npm_packs artifact                     -> no build
PASS  green run, no artifacts at all                       -> no build

--- ambiguous or missing version must fail loudly, not exit green ---
PASS  two npm_packs versions on one run                    -> hard failure (exit 1)
PASS  dispatch when npm latest is empty                    -> hard failure (exit 1)

--- manual dispatch ignores the triggering-run oracle entirely ---
PASS  dispatch, explicit version                           -> 17.0.1
PASS  dispatch, old version after stable                   -> 16.2.0
PASS  input overrides the run's artifact                   -> 16.2.0
PASS  dispatch, no input -> npm latest                     -> 18.0.0

--- live Actions API against real historical Publish runs ---
PASS  real stable run 28431654718                          -> 18.0.0
PASS  real first-RC run 27610213255                        -> 18.0.0-rc1
PASS  real RC push run 28153016668                         -> 18.0.0-rc5
PASS  real RC push run 27950003566                         -> 18.0.0-rc2
PASS  cancelled run resolves rc6 (if-gate blocks it)       -> 18.0.0-rc6
PASS  that run is not green                                -> conclusion=cancelled, if-gate blocks it
ALL RESOLVE CASES PASSED
```

`real stable run 28431654718` is the exact run this ticket exists for — it produced no artifact before and now resolves `18.0.0`. The `run shipped rc5 while rc6 tag exists` pair shows two concurrent runs each getting their own correct artifact.

The last two lines record a caveat worth knowing: `npm_packs_*` proves a **build**, not a **publish**. The cancelled run `28430639792` uploaded `npm_packs_18.0.0-rc6` and never published rc6 — that is also where the stray `18.0.0-rc6` git tag came from. Safety comes from the job `if` requiring `conclusion == 'success'`, which is asserted rather than assumed, with the npm-readiness poll as a second backstop. Both hard-failure cases assert a non-zero exit *and* an `::error::` line, so a refactor cannot quietly downgrade either to a notice.

**3. Artifact assembly end-to-end, plus fault injection** (the extracted `Assemble demo artifact` step; the last two cases shadow `npm`/`curl` on `PATH` to force each failure):

```
PASS  stable 18.0.0 (the release case)             exit=0 js=1722134 bytes
PASS  RC 18.0.0-rc5                                exit=0 js=1722150 bytes
PASS  older stable 17.1.0                          exit=0 js=1816541 bytes
PASS  version that does not exist                  exit=16 (failed loudly, as intended)
ALL ASSEMBLE CASES PASSED

PASS  npm outage -> jsDelivr fallback succeeded (1722134 bytes)
PASS  455-byte error page -> size guard failed the job (exit=1)
ALL SAFETY-NET CASES PASSED
```

The second safety-net case is the exact failure from the linked run: the guard now fails the job instead of uploading a broken artifact.

**Static checks** — `actionlint` 1.7.12 reports no findings. To confirm it really inspects the expressions (rather than treating them as opaque strings), a deliberately corrupted copy was linted and correctly rejected:

```
undefined function "startsWithTypo". available functions are ... [expression]
```

### Commands run

```bash
# actionlint (expression + workflow schema)
actionlint .github/workflows/componentsource.yml
# -> exit 0, no findings

# YAML parses
python3 -c "import yaml; yaml.safe_load(open('.github/workflows/componentsource.yml'))"
# -> YAML OK

# 1. trigger replay vs 282 real Publish runs
node simulate.mjs          # -> ALL ASSERTIONS PASSED

# 2. version resolution, no-op and failure paths — 21 cases
bash test-resolve.sh       # -> ALL RESOLVE CASES PASSED

# 3. assembly end-to-end + fault injection
bash test-assemble.sh      # -> ALL ASSEMBLE CASES PASSED
bash test-fallback.sh      # -> ALL SAFETY-NET CASES PASSED
```

`eslint`, `stylelint`, `build`, `test:unit` and `test:e2e` were not run: the diff is a single YAML workflow file, so none of them have anything to act on.

### Types of changes

### Related issue(s):

1. DEV-1968

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

None of the published packages are affected. The change is CI/tooling only, so the changelog check passes on path rules and no `.changelogs/` entry is required.

### Checklist:

### Reviewer notes

Two things worth knowing before merge.

1. **This has no effect until it is on `develop`.** `workflow_run` workflows always run from the default branch. If this is not merged before the next release, the artifact will be missed again.
2. **One behavior is intentionally dropped.** A manual `workflow_dispatch` of *`Publish`* no longer builds an artifact. The current `publish.yml` gates every publish job on `release`, `push` or `workflow_run`, so a dispatched `Publish` run publishes nothing; building from it would package a version that was not just released. One historical run (from when a dispatch pre-release arm still existed) matched that shape. Dispatching *`Component Source`* itself still works, and now takes an optional version.

3. **The trigger is unfiltered on purpose.** Every completed `Publish` run now creates a `Component Source` run, and most skip instantly — 236 of 282 historically. That is the deliberate price of not depending on how GitHub treats a tag-valued `head_branch` in a ref filter. Please do not "tidy" a `branches`/`branches-ignore` filter back in; the workflow comment explains why, and re-adding one can silently skip a release with every test still green.

4. **`npm_packs_*` proves a build, not a publish.** The cancelled run `28430639792` uploaded `npm_packs_18.0.0-rc6` and never published rc6 (it also left the stray `18.0.0-rc6` git tag). The job `if` requiring `conclusion == 'success'` is what makes the oracle safe, so please keep it — a failed or rejected `rc-publish` cannot leave a green run. The npm-readiness poll is the second backstop.

After merge, confirmation arrives in stages: the next RC push proves the chain still fires, and the next stable release proves the fix. If anything is still wrong, the manual dispatch is the backstop.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-1968

### DIFF
--- a/.github/workflows/componentsource.yml
+++ b/.github/workflows/componentsource.yml
@@ -1,14 +1,35 @@
 name: Component Source
 
+# Builds the "component source" deliverable — a small demo bundle (example page
+# plus the released `handsontable.full.min.js`) that GitHub serves as a ZIP.
+#
+# Chained off `Publish` rather than hooked straight to the `release` event: the
+# artifact embeds the released dist file, so it can only be assembled AFTER
+# `Publish` has pushed that version to npm.
 on:
   workflow_run:
     workflows: ['Publish']
     types: [completed]
-    branches: ['release/**']
+    # Everything-but-develop, NOT an allow-list of branch patterns. A `Publish`
+    # run's `head_branch` is a BRANCH for RC pushes (`release/18.0.0`) but the
+    # TAG for release events (`18.0.0`), so the old `branches: ['release/**']`
+    # allow-list never matched a stable release and silently skipped the
+    # deliverable for every official version (DEV-1968). `develop` is excluded
+    # because its chained `experimental` publish is not a customer deliverable.
+    # The job's `if` re-checks everything else.
+    branches-ignore:
+      - develop
   workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version to package. Defaults to the current `latest` on npm.'
+        required: false
+        type: string
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.workflow_run.head_branch || github.ref }}
+  # `github.event.inputs` rather than the `inputs` context: this line is
+  # evaluated for `workflow_run` events too, where `inputs` is not populated.
+  group: ${{ github.workflow }}-${{ github.event.workflow_run.head_branch || github.event.inputs.version || github.ref }}
   cancel-in-progress: true
 
 permissions:
@@ -18,22 +39,102 @@ jobs:
   build-componentsource:
     name: Build component source artifact
     runs-on: ubuntu-latest
-    if: ${{ github.event_name == 'workflow_dispatch' || github.event.workflow_run.conclusion == 'success' }}
+    # A manual dispatch always runs. An automatic run needs a green `Publish`
+    # run that actually pushed a customer-facing version to npm: a `release`
+    # event (stable release, or the first RC) or a push to a release branch
+    # (every later RC).
+    if: |
+      github.event_name == 'workflow_dispatch' || (
+        github.event.workflow_run.conclusion == 'success' && (
+          github.event.workflow_run.event == 'release' || (
+            github.event.workflow_run.event == 'push' &&
+            startsWith(github.event.workflow_run.head_branch, 'release/')
+          )
+        )
+      )
     steps:
+      # Checks out the default branch on purpose — `.nvmrc` is the only repo
+      # file this job reads, and the version now comes from npm (see below).
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-        with:
-          ref: ${{ github.event.workflow_run.head_sha || github.sha }}
 
       - name: Use Node.js
         uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # v6.2.0
         with:
           node-version-file: '.nvmrc'
 
-      - name: Read Handsontable version
+      - name: Resolve published version
         id: version
+        env:
+          TRIGGER_EVENT: ${{ github.event.workflow_run.event }}
+          TRIGGER_REF: ${{ github.event.workflow_run.head_branch }}
+          INPUT_VERSION: ${{ github.event.inputs.version }}
         run: |
-          VERSION=$(node -e 'console.log(require("./hot.config.js").HOT_VERSION)')
-          echo "value=$VERSION" >> "$GITHUB_OUTPUT"
+          set -euo pipefail
+
+          # Never read `hot.config.js`: on the commit a `Publish` run starts
+          # from, it still holds the PREVIOUS version — the bump is committed
+          # during that run. Reading it made every RC artifact one RC stale.
+          # The release tag and npm's dist-tags are the only sources that name
+          # what was actually published (`--tag latest` for stable, `--tag rc`
+          # for release candidates).
+          if [ -n "${INPUT_VERSION}" ]; then
+            VERSION="${INPUT_VERSION}"
+            SOURCE="workflow_dispatch input"
+
+          elif [ "${TRIGGER_EVENT}" = 'release' ]; then
+            # For a release event `head_branch` carries the release TAG. Strip an
+            # optional `v` prefix exactly the way publish.yml does (`${TAG#v}`).
+            TAG="${TRIGGER_REF#v}"
+
+            if [ "${TAG}" = "${TAG%%-*}" ]; then
+              # Stable release, e.g. `18.0.0` — the tag is the published version.
+              VERSION="${TAG}"
+              SOURCE="release tag"
+            else
+              # First RC: the tag is `18.0.0-rc`, which does NOT carry the `rcN`
+              # suffix npm actually got.
+              VERSION="$(npm view handsontable dist-tags.rc)"
+              SOURCE="npm dist-tag 'rc' (prerelease tag ${TRIGGER_REF})"
+            fi
+
+          elif [ -n "${TRIGGER_EVENT}" ]; then
+            # A push to a release branch — every RC after the first.
+            VERSION="$(npm view handsontable dist-tags.rc)"
+            SOURCE="npm dist-tag 'rc'"
+
+          else
+            VERSION="$(npm view handsontable dist-tags.latest)"
+            SOURCE="npm dist-tag 'latest'"
+          fi
+
+          if [ -z "${VERSION}" ]; then
+            echo "::error::Could not resolve a version to package (source: ${SOURCE})"
+            exit 1
+          fi
+
+          echo "Packaging handsontable@${VERSION} (resolved from ${SOURCE})"
+          echo "value=${VERSION}" >> "$GITHUB_OUTPUT"
+
+      - name: Wait for the version to be readable on npm
+        env:
+          HOT_VERSION: ${{ steps.version.outputs.value }}
+        run: |
+          set -euo pipefail
+
+          # `Publish` completing does not guarantee the new version is already
+          # readable across the npm registry, and this job starts seconds later.
+          for attempt in $(seq 1 20); do
+            if npm view "handsontable@${HOT_VERSION}" version > /dev/null 2>&1; then
+              echo "handsontable@${HOT_VERSION} is on npm"
+              exit 0
+            fi
+
+            echo "Attempt ${attempt}/20: handsontable@${HOT_VERSION} not on npm yet, retrying in 15s"
+            sleep 15
+          done
+
+          echo "::error::handsontable@${HOT_VERSION} never became readable on npm"
+          exit 1
 
       - name: Assemble demo artifact
         env:
@@ -82,12 +183,47 @@ jobs:
           </html>
           HTML
 
-          curl -fL --retry 3 --retry-delay 2 \
-            "https://cdn.jsdelivr.net/npm/handsontable@${HOT_VERSION}/dist/handsontable.full.min.js" \
-            -o "$EX/handsontable.full.min.js"
+          JS_DEST="$EX/handsontable.full.min.js"
+
+          # The npm registry first: it is the origin jsDelivr itself pulls from,
+          # so it is ready the moment `Publish` finishes. jsDelivr is the
+          # fallback only — the stable publish purges its cache
+          # (publish.yml -> purge-jsdelivr-cache.mjs), which left the CDN cold
+          # exactly when this job ran and returned a 503 on 18.0.0 (DEV-1968).
+          download_from_npm() {
+            local tmp
+            tmp="$(mktemp -d)"
+
+            npm pack "handsontable@${HOT_VERSION}" --pack-destination "$tmp" > /dev/null || return 1
+            tar -xzf "$tmp"/handsontable-*.tgz -C "$tmp" package/dist/handsontable.full.min.js || return 1
+            mv "$tmp/package/dist/handsontable.full.min.js" "$JS_DEST" || return 1
+          }
+
+          download_from_jsdelivr() {
+            curl -fL --retry 10 --retry-all-errors --retry-delay 5 --retry-max-time 300 \
+              "https://cdn.jsdelivr.net/npm/handsontable@${HOT_VERSION}/dist/handsontable.full.min.js" \
+              -o "$JS_DEST"
+          }
+
+          if ! download_from_npm; then
+            echo "::warning::Could not fetch handsontable@${HOT_VERSION} from npm, falling back to jsDelivr"
+            download_from_jsdelivr
+          fi
+
+          # A CDN error page is a few hundred bytes and `curl -f` does not always
+          # catch it. The real file is over 1.5 MB, so guard the floor.
+          SIZE="$(wc -c < "$JS_DEST" | tr -d '[:space:]')"
+
+          if [ "${SIZE}" -lt 500000 ]; then
+            echo "::error::handsontable.full.min.js is only ${SIZE} bytes — the download is not the real dist file"
+            exit 1
+          fi
+
+          echo "Bundled handsontable.full.min.js (${SIZE} bytes)"
 
       - name: Upload component source artifact
         uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
         with:
           name: componentsource-${{ steps.version.outputs.value }}
           path: componentsource-build
+          if-no-files-found: error
--- a/.github/workflows/componentsource.yml
+++ b/.github/workflows/componentsource.yml
@@ -10,15 +10,21 @@ on:
   workflow_run:
     workflows: ['Publish']
     types: [completed]
-    # Everything-but-develop, NOT an allow-list of branch patterns. A `Publish`
-    # run's `head_branch` is a BRANCH for RC pushes (`release/18.0.0`) but the
-    # TAG for release events (`18.0.0`), so the old `branches: ['release/**']`
-    # allow-list never matched a stable release and silently skipped the
-    # deliverable for every official version (DEV-1968). `develop` is excluded
-    # because its chained `experimental` publish is not a customer deliverable.
-    # The job's `if` re-checks everything else.
-    branches-ignore:
-      - develop
+    # DELIBERATELY UNFILTERED. A `Publish` run's `head_branch` is a BRANCH for RC
+    # pushes (`release/18.0.0`) but the TAG for release events (`18.0.0`), and the
+    # old `branches: ['release/**']` allow-list therefore never matched a stable
+    # release — the deliverable was silently skipped for every official version
+    # (DEV-1968). Whether a `branches`/`branches-ignore` filter passes a
+    # tag-valued `head_branch` at all is unspecified (tag filtering for
+    # `workflow_run` is not supported), so ANY filter here risks re-creating that
+    # exact silent miss. Gating lives in the job's `if` instead, where the rules
+    # are explicit and testable. Excluding develop was already redundant:
+    # publish.yml's `push` trigger covers release/** only and `release` is
+    # tag-based, so a develop `head_branch` can only arrive as `workflow_run`
+    # (the chained `experimental` publish) or `workflow_dispatch` — and neither
+    # passes either arm of that `if`. Cost is a handful of instantly-skipped
+    # runs; the benefit is that a release can no longer be missed by a filter
+    # whose behavior on tags is unspecified.
   workflow_dispatch:
     inputs:
       version:
@@ -68,24 +74,43 @@ jobs:
           TRIGGER_EVENT: ${{ github.event.workflow_run.event }}
           TRIGGER_REF: ${{ github.event.workflow_run.head_branch }}
           INPUT_VERSION: ${{ github.event.inputs.version }}
+          GH_REPO: ${{ github.repository }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           set -euo pipefail
 
           SHOULD_BUILD=true
-          # Release line this run belongs to (`18.0.0`), set only when the
-          # version was resolved through the shared `rc` dist-tag.
+          # Release line (`18.0.0`) when this run shipped a candidate rather than
+          # a stable version. Empty on the stable and manual-dispatch paths.
           LINE=''
 
-          npm_has() {
-            npm view "handsontable@$1" version > /dev/null 2>&1
+          # The oracle publish.yml's `rc-check` uses (publish.yml:545-556): a
+          # NON-PRERELEASE GitHub release for the line. It must be this and not
+          # npm — the GitHub release is created before `stable-publish` pushes to
+          # npm, so an npm probe would disagree with `rc-check` inside that
+          # window and upload a stale candidate.
+          stable_release_exists() {
+            local count
+            count=$(gh release list --repo "${GH_REPO}" --json tagName,isPrerelease \
+              --jq "[.[] | select(.tagName == \"$1\" and (.isPrerelease | not))] | length")
+
+            [ "${count:-0}" -gt 0 ]
+          }
+
+          # Highest N among the `<line>-rcN` git tags. publish.yml creates that
+          # tag through the GitHub API *before* publishing (publish.yml:719-723),
+          # so it names the candidate THIS run shipped. npm's `rc` dist-tag
+          # cannot: it is one mutable global pointer, shared by every release
+          # line, and may still read the previous RC while the new one is live.
+          highest_rc_number() {
+            gh api "repos/${GH_REPO}/git/matching-refs/tags/$1-rc" --jq '.[].ref' \
+              | sed -n "s|^refs/tags/$1-rc\([0-9]\{1,\}\)\$|\1|p" \
+              | sort -n | tail -n 1
           }
 
           # Never read `hot.config.js`: on the commit a `Publish` run starts
           # from, it still holds the PREVIOUS version — the bump is committed
           # during that run. Reading it made every RC artifact one RC stale.
-          # The release tag and npm's dist-tags are the only sources that name
-          # what was actually published (`--tag latest` for stable, `--tag rc`
-          # for release candidates).
           if [ -n "${INPUT_VERSION}" ]; then
             VERSION="${INPUT_VERSION}"
             SOURCE="workflow_dispatch input"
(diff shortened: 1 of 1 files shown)


qunabu:
Task linked: [DEV-1968 `Component Source` Github Action fails on release](https://app.clickup.com/t/9015210959/DEV-1968)

budnix:
Confirmed and fixed in 43eb84119c.

The mechanism in the report is right. `rc-check` sets `allowed=false` once a stable release exists for the line, which skips `rc-build`, `rc-test` and `rc-publish` — but `rc-check` itself succeeds, so the `Publish` run still concludes `success` having published nothing. The RC path then fell through to npm's `rc` dist-tag.

One correction to the framing: the artifact could not have been **mislabeled**. The dist file is always downloaded for the exact version the artifact is named after, so contents and label always agree. The real defect is a **spurious artifact** — a run that published nothing should not produce a deliverable at all.

Two guards, both checking ground truth rather than this workflow's own assumptions:

1. **Stable already shipped** — if the line's stable version is on npm, the run is a no-op. This is the same condition `rc-check` uses, so the two cannot drift apart.
2. **Wrong release line** — `rc` is one dist-tag shared by every line, so a version resolved through it must belong to the triggering line (`18.0.0-*` for `release/18.0.0`). This covers the cross-line case.

Both emit a `::notice::` explaining why, set `should-build=false`, and the wait/assemble/upload steps are gated on it. The run stays green and uploads nothing, so a legitimate backport push does not turn CI red.

The stable-release path and the manual dispatch are deliberately exempt. Dispatch exists precisely to repackage an older version by hand, so it must ignore both guards.

Verified with 14 resolve cases, including all three no-op shapes and one live-registry case — `18.0.0` is stable on npm today, so a push to `release/18.0.0` right now must no-op, and does:

```
--- green Publish that published nothing ---
PASS  backport push after stable shipped [FAKE rc=18.0.0-rc5] -> no build
PASS  push while rc tag is another line [FAKE rc=18.1.0-rc1] -> no build
PASS  first-RC tag vs other line's rc tag [FAKE rc=18.1.0-rc1] -> no build
PASS  push to release/18.0.0 on live npm                   -> no build

--- manual dispatch (must ignore all of the above guards) ---
PASS  dispatch, old version after stable                   -> 16.2.0
```

Cases marked `FAKE` shadow `npm` on `PATH` to simulate a registry state, since the live registry has already moved past the RC phase. The stable-release path that this PR exists to fix is unaffected — it still resolves `18.0.0` from the release tag and builds.

demtario (commented):
Review of the trigger fix. Four blocking-ish points inline (2 high, 2 medium). The npm-first download, the `should-build` gating of every downstream step, and the `head_branch`-is-the-tag diagnosis all check out — I verified the 18.0.0 tarball ships `dist/handsontable.full.min.js` at 1,722,134 bytes, so the 500 KB floor is well placed.

.github/workflows/componentsource.yml
```diff
@@ -1,14 +1,35 @@
...
     types: [completed]
-    branches: ['release/**']
+    # Everything-but-develop, NOT an allow-list of branch patterns. A `Publish`
+    # run's `head_branch` is a BRANCH for RC pushes (`release/18.0.0`) but the
+    # TAG for release events (`18.0.0`), so the old `branches: ['release/**']`
+    # allow-list never matched a stable release and silently skipped the
+    # deliverable for every official version (DEV-1968). `develop` is excluded
+    # because its chained `experimental` publish is not a customer deliverable.
+    # The job's `if` re-checks everything else.
+    branches-ignore:
```
demtario: **High — this may re-create the bug it fixes.** The fix assumes a tag-valued `head_branch` (`18.0.0`) *passes* `branches-ignore`. That is undocumented: the docs only say these filters select "what **branches** the triggering workflow must run on", and tag filtering for `workflow_run` is unsupported ([starter-workflows#1137](https://github.com/actions/starter-workflows/issues/1137), [community#136800](https://github.com/orgs/community/discussions/136800)). The 282-run replay validates the model of the filter, not GitHub's behavior — the evidence table has no Component Source run from a tag-valued `head_branch`, under a filter that provably could not match, so it is no evidence either way. If the assumption is wrong, every test passes and 18.1.0 ships with no artifact.

Free fix: **drop the filter entirely.** The exclusion is already redundant — a develop-chained Publish arrives with `workflow_run.event == 'workflow_run'` and fails both arms of the job `if` below. Cost is a few skipped runs; benefit is the fix no longer depends on unspecified tag semantics.

.github/workflows/componentsource.yml
```diff
@@ -18,24 +39,151 @@ jobs:
...
+            # to a closed release branch would re-upload a stale candidate.
+            LINE="${TRIGGER_REF#release/}"
+
+            if npm_has "${LINE}"; then
+              echo "::notice::Stable ${LINE} is already published, so publish.yml's rc-check blocked the RC publish and this run published nothing — no artifact to build."
+              SHOULD_BUILD=false
+              VERSION=''
+              SOURCE='n/a'
+            else
+              VERSION="$(npm view handsontable dist-tags.rc)"
```
demtario: **High — stale `rc` dist-tag reintroduces the off-by-one this PR removes.** Both RC paths (here and line 105) read `dist-tags.rc` once, never compared against the RC *this* run published. If the packument lags `rc-publish` by seconds, `VERSION` resolves to `18.0.0-rc4` while rc5 shipped. Neither guard can see it: the line check passes (`18.0.0-rc4` matches `18.0.0-*`), and the npm-readiness poll passes *immediately* because rc4 genuinely exists. Same defect class as reading `hot.config.js`, different source.

Suggest deriving the expected RC from the git tags on `TRIGGER_REF` (the way publish.yml's *Determine next RC version* does) and polling until `dist-tags.rc` matches it.

.github/workflows/componentsource.yml
```diff
@@ -18,24 +39,151 @@ jobs:
...
+          # `rc` is ONE dist-tag shared by every release line. Whenever the
+          # version came from it, check the candidate belongs to the line that
+          # triggered this run — otherwise a push on one branch would package
+          # another line's candidate.
+          if [ "${SHOULD_BUILD}" = true ] && [ -n "${LINE}" ]; then
+            case "${VERSION}" in
+              "${LINE}-"*)
+                ;;
+              *)
+                echo "::notice::npm dist-tag 'rc' is '${VERSION}', which is not a candidate for the ${LINE} line — no artifact to build."
```
demtario: **Medium — empty version reports the wrong reason, and the loud guard below is unreachable.** `npm view handsontable dist-tags.<missing>` prints nothing and exits 0, so `VERSION=''` lands in this `*)` arm and logs `npm dist-tag 'rc' is '', which is not a candidate for the 18.0.0 line` before exiting green.

Consequence: the `::error::` at line 154 is unreachable on **both** RC paths — whenever `LINE` is set this `case` fires first. Only the dispatch/`latest` path can reach it, and that path cannot yield an empty version. A `::notice::` in a green run is invisible unless someone opens it, which is how DEV-1968 stayed unnoticed. Handle empty separately and fail hard.

.github/workflows/componentsource.yml
```diff
@@ -18,24 +39,151 @@ jobs:
...
+          elif [ -n "${TRIGGER_EVENT}" ]; then
+            # A push to a release branch — every RC after the first. Such a run
+            # goes green even when it published NOTHING: publish.yml's
+            # `rc-check` sets `allowed=false` once a stable release exists for
+            # the line, which skips rc-build/rc-test/rc-publish while `rc-check`
+            # itself succeeds. Mirror that guard here, or a later backport push
+            # to a closed release branch would re-upload a stale candidate.
+            LINE="${TRIGGER_REF#release/}"
+
+            if npm_has "${LINE}"; then
```
demtario: **Medium — this does not mirror `rc-check`; the two oracles can disagree.** `rc-check` counts non-prerelease **GitHub releases** (`publish.yml:545-556`, `gh release list --jq '... tagName == VERSION and (.isPrerelease | not)'`). This checks **npm**. So "Mirror that guard here" and the commit's "the same condition `rc-check` uses, so the two cannot disagree" are not accurate.

Concrete divergence: a stable release creates the GitHub release *before* `stable-publish` pushes to npm (`publish.yml:1559`). A push to `release/18.0.0` inside that window → `rc-check` blocks (run publishes nothing) but `npm_has 18.0.0` is false → the stale candidate gets uploaded, exactly what this guard exists to prevent. The reverse divergence silently skips a real deliverable. Either query `gh release list` for parity, or drop the equivalence claim.

.github/workflows/componentsource.yml
budnix: Agreed, and this was the most dangerous of the four. Fixed in 9fb57843fe by taking your free fix: **the ref filter is gone entirely.**

You are right that my evidence was circular. Under a filter that provably could not match `18.0.0`, the absence of a Component Source run is explained equally well by "the filter rejected it" and "tags never pass a ref filter". The replay validated my model of the filter, not GitHub's behavior, so it was no evidence either way — and the failure mode is the silent one.

I also verified your redundancy argument, and it holds for a slightly broader reason than stated. Across all 104 develop `Publish` runs the event is `workflow_run` (100) or `workflow_dispatch` (4) — never `release` or `push`:

```
PASS  every develop Publish run is workflow_run or workflow_dispatch (n=104)
PASS  no develop Publish run is a release event or a push
PASS  experimental develop publishes still never build, with NO ref filter
```

That is structural rather than historical: publish.yml's `push` trigger covers `release/**` only and `release` is tag-based, so a develop `head_branch` cannot arrive as either event. Both arms of the job `if` therefore reject it without any filter.

Measured cost, from the same replay: 236 of 282 historical `Publish` runs would create a Component Source run that skips immediately. No runner time, just log entries. Worth it to stop depending on unspecified tag semantics.

The one thing still unprovable without a release is whether `workflow_run` fires at all for a release-triggered upstream run. The issues you linked are about tag *filtering*, not about firing, and with no filter there is nothing to filter — so this is the safest configuration available. The manual dispatch remains the backstop.

.github/workflows/componentsource.yml
budnix: Correct, and the diagnosis is exact — same defect class as reading `hot.config.js`, different source. Fixed in 9fb57843fe along the line you suggested.

Both RC paths now derive the version from the highest `<line>-rcN` **git tag** instead of `dist-tags.rc`:

```bash
gh api "repos/${GH_REPO}/git/matching-refs/tags/$1-rc" --jq '.[].ref' \
  | sed -n "s|^refs/tags/$1-rc\([0-9]\{1,\}\)\$|\1|p" | sort -n | tail -n 1
```

This is parity by construction, not by coincidence: publish.yml's *Determine next RC version* reads the same tags and takes highest + 1 (`publish.yml:613`), then `rc-build` creates that tag through the API **before** `rc-publish` runs (`publish.yml:719-723`). So after a green run the highest tag is exactly what it shipped. The `rc` dist-tag could never give that — one mutable global pointer, shared across lines.

On the failure direction: git tags can run **ahead** of npm. This repo has a real `18.0.0-rc6` tag with no matching npm version, left behind by the cancelled run on 06-30. That asymmetry is why tags are the right oracle:

- stale `dist-tags.rc` → resolves rc4 while rc5 is live → **silently wrong artifact**, and as you note the readiness poll passes immediately because rc4 genuinely exists.
- tag ahead of npm → resolves a version npm lacks → the existing readiness poll exhausts its 5 minutes and **fails red**.

Loud beats silent here, so I kept the poll as the safety net rather than adding a second one. A run that reaches this state has a genuinely broken publish and should get a human's attention.

A side effect worth calling out: cross-line leakage is now structurally impossible, because the version is *constructed* from the triggering line rather than selected from a global tag. The explicit line-match check I added in the previous commit is gone — it can no longer fail. Covered by:

```
PASS  push to release/18.1.0, tags up to rc3             -> 18.1.0-rc3
PASS  rc10 beats rc9 (numeric sort)                      -> 18.1.0-rc10
PASS  only the bare -rc tag exists                       -> hard failure (exit 1)
PASS  17.1.1 push while 18.1.0-rc9 exists                -> 17.1.1-rc1
```

The last case is the isolation proof: a numerically higher tag on another line is ignored.

.github/workflows/componentsource.yml
budnix: Confirmed — the `::error::` was unreachable on both RC paths, exactly as described, and "a `::notice::` in a green run is invisible, which is how DEV-1968 stayed unnoticed" is the right reason to care. Fixed in 9fb57843fe.

Empty and unknown versions now fail hard on every path:

- RC paths get their own explicit guard. If no `<line>-rcN` tag exists, the step exits 1 with `::error::No <line>-rcN git tag exists, so the candidate this run published cannot be identified`. There is no green-exit arm left that can swallow it.
- The generic `::error::Could not resolve a version` guard is now genuinely reachable, since the line-match `case` that used to intercept first is gone entirely (see the reply on the `rc` dist-tag — the version is constructed from the line now, so that check became impossible to fail).

The only remaining green-with-no-artifact exit is the deliberate one: `rc-check` blocked the publish, which is a real no-op rather than an unresolved version.

Both failure paths are now covered by tests that assert a non-zero exit **and** an `::error::` line, so a future refactor cannot quietly downgrade either to a notice:

```
PASS  release-branch push with no rc tags                -> hard failure (exit 1)
        ::error::No 18.1.0-rcN git tag exists, so the candidate this run published cannot be identified
PASS  dispatch when npm latest is empty                  -> hard failure (exit 1)
        ::error::Could not resolve a version to package (source: npm dist-tag 'latest')
```

I also left a comment on the guard saying not to downgrade it to a notice, with the reason.

.github/workflows/componentsource.yml
budnix: You are right, and the claim was mine to justify — both the code comment and the commit message overstated it. Fixed in 9fb57843fe by taking the parity option rather than dropping the claim.

The guard now queries the same oracle `rc-check` does:

```bash
stable_release_exists() {
  local count
  count=$(gh release list --repo "${GH_REPO}" --json tagName,isPrerelease \
    --jq "[.[] | select(.tagName == \"$1\" and (.isPrerelease | not))] | length")

  [ "${count:-0}" -gt 0 ]
}
```

That is `publish.yml:545-556` verbatim in intent — non-prerelease GitHub releases, not npm. The window you identified is what makes this necessary rather than cosmetic: `stable-prepare` runs on the release event, so the GitHub release exists before `stable-publish` reaches npm (`publish.yml:1559`). A push to `release/18.0.0` inside that window would have had `rc-check` blocking while my npm probe returned false — uploading the stale candidate this guard exists to stop. The reverse divergence would have skipped a real deliverable.

The step needs `GH_TOKEN` for this; `permissions: contents: read` already covers reading releases and refs, so no permission change.

Verified both against a fake oracle and against the live repo, where `18.0.0` does have a stable release today:

```
PASS  backport push after stable release                 -> no build
        ::notice::18.0.0 already has a stable GitHub release, so publish.yml's rc-check blocked the RC publish ...
PASS  push to release/18.0.0 against live repo           -> no build
```

The code comment now states *why* it must be GitHub releases and not npm, so the next person does not "simplify" it back to an npm probe.

.github/workflows/componentsource.yml
```diff
@@ -18,24 +45,162 @@ jobs:
...
+
+          # Highest N among the `<line>-rcN` git tags. publish.yml creates that
+          # tag through the GitHub API *before* publishing (publish.yml:719-723),
+          # so it names the candidate THIS run shipped. npm's `rc` dist-tag
+          # cannot: it is one mutable global pointer, shared by every release
+          # line, and may still read the previous RC while the new one is live.
+          highest_rc_number() {
+            gh api "repos/${GH_REPO}/git/matching-refs/tags/$1-rc" --jq '.[].ref' \
+              | sed -n "s|^refs/tags/$1-rc\([0-9]\{1,\}\)\$|\1|p" \
+              | sort -n | tail -n 1
```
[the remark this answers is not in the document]
budnix: Valid, and it goes further than stated. Fixed in 461b670c58 by replacing the oracle rather than patching the race.

I verified both premises: publish.yml has **no** `concurrency:` block, and `rc-publish` sits behind `environment: approvers`. So the window between `rc-build` writing the tag and `rc-publish` reaching npm is not seconds — it is however long an approver takes. A second push landing in that window tags rcN+1 while rcN is still pending, and "highest tag" then names the wrong run's candidate. `rc-build` also checks out `github.ref_name` rather than `head_sha`, so tying the tag back to the triggering commit is not reliable either.

The fix uses something the run already produces. Every publish path uploads the exact tarballs it publishes as `npm_packs_<version>` (`publish.yml:203`, `339`, `655`, `1430`, `1495`), and `rc-publish` / `stable-publish` download that same artifact to run `npm publish`. Reading the name off the **triggering run** is therefore a per-run record of what that run shipped:

```bash
VERSIONS="$(gh api "repos/${GH_REPO}/actions/runs/${TRIGGER_RUN_ID}/artifacts" \
  --paginate --jq '.artifacts[].name' \
  | sed -n 's/^npm_packs_//p' | sort -u)"
```

Concurrency, tag ordering, approval delays and release lines cannot affect it — the answer is scoped to one run.

It also collapsed the design rather than adding to it. The release-tag parsing, the highest-rc-tag lookup and the `gh release list` mirror of `rc-check` are all gone: one oracle now covers stable, first RC and later RCs. A run that published nothing uploads no packs, so the no-op case falls out for free instead of re-deriving `rc-check`'s condition. Cross-line leakage and the `v`-prefix case are impossible rather than guarded. Net −36 lines.

**One caveat I want on the record, because it is not obvious.** The artifact proves a *build*, not a *publish*. The cancelled run `28430639792` uploaded `npm_packs_18.0.0-rc6` and never published rc6 — that is also where the stray `18.0.0-rc6` git tag came from. Safety there comes from the job `if` requiring `conclusion == 'success'`: a failed or rejected `rc-publish` cannot leave a green run. The npm-readiness poll is the second backstop. Both are asserted, not assumed:

```
PASS  cancelled run resolves rc6 (if-gate blocks it)       -> 18.0.0-rc6
PASS  that run is not green                                -> conclusion=cancelled, if-gate blocks it
PASS  no non-green Publish run ever reaches the job (n=109)
```

Ambiguity now fails loudly too — two different `npm_packs_*` versions on one run is an `::error::`, not a guess.

Resolve coverage is 21 cases. Five run against the live Actions API on real historical runs, which is the strongest evidence available short of a release:

```
PASS  real stable run 28431654718                          -> 18.0.0
PASS  real first-RC run 27610213255                        -> 18.0.0-rc1
PASS  real RC push run 28153016668                         -> 18.0.0-rc5
PASS  real RC push run 27950003566                         -> 18.0.0-rc2
```

The first line is the run this ticket exists for. Artifact names survive artifact expiry, which those cases also demonstrate.

This needs `actions: read`, now declared alongside `contents: read`.

demtario (approved):

PR #13205, opened 2026-08-20
DEV-1476: Allow setting colorScheme and density without declaring a theme

budnix:
### Context

The PR adds two grid options, `colorScheme` and `density`, so you can change a grid's color scheme or its spacing without declaring a theme.

Today the only way to switch to dark mode or to a tighter layout is the Theme API builder:

```js
import { registerTheme, mainTheme } from 'handsontable/themes';

const theme = registerTheme(mainTheme)
  .setColorScheme('dark')
  .setDensityType('compact');

const hot = new Handsontable(container, { theme });
```

That is a lot of setup for a dark mode toggle, and it has a second problem: the registered theme is a shared object, so mutating it changes every grid that uses it.

Now both are plain options:

```js
  colorScheme: 'dark',
  density: 'compact',
});

// and at runtime
hot.updateSettings({ colorScheme: 'auto' });
```

| Option | Allowed values |
| --- | --- |
| `colorScheme` | `'light'`, `'dark'`, `'auto'` |
| `density` | `'default'`, `'compact'`, `'comfortable'` |

Setting an option to `undefined` clears the override and falls back to the theme value. The existing `registerTheme()` builder API is unchanged.

### How it works

Both options are **per-instance overrides** applied by `ThemeManager` on top of the theme configuration. The shared registered theme object is never mutated, so other grids using the same theme keep their own look.

The override CSS is emitted under a per-instance scope class (`ht-scope-<guid>`), stamped on **both** the root wrapper and the portal element. The portal matters: menus and dialogs render there, and a wrapper-only scope would leave a light menu on a dark grid. The scoped selector doubles the theme class with the scope class, giving specificity (0,2,0) so it wins over both the `:where()` block the engine emits (0,0,0) and the static `ht-theme-*.css` declarations (0,1,0).

#### Why `colorScheme` does not just set `color-scheme`

Setting the `color-scheme` property alone looks like it should be enough, and it is not. The shipped **minified** theme stylesheets do not use the CSS `light-dark()` function — it needs Chrome 123 / Safari 17.5, well above the floor declared in `browser-targets.js` (Chrome 110 / Firefox 110 / Safari 14.1) — so the minifier rewrites it into a pair of variables switched by **theme class name**:

```css
.ht-theme-main      { --lightningcss-light: initial; --lightningcss-dark:  ;       color-scheme: light }
.ht-theme-main-dark { --lightningcss-light:  ;       --lightningcss-dark: initial; color-scheme: dark }
```

Those declarations out-specify the ones the engine generates, so a grid that merely flipped `color-scheme` kept its light colors. That is a pre-existing defect in `setColorScheme()`, which I confirmed on the released build before fixing it here.

So `ThemeManager` emits the **resolved** light or dark value for every light/dark token in its scoped block, which needs no `light-dark()` support at all and therefore works on every supported browser:

```css
.ht-theme-main.ht-scope-ht_abc123 {
  color-scheme: dark;
  --ht-background-color: var(--ht-colors-palette-950);
  /* ...every other light/dark token... */
}
```

`'auto'` pins the light values and swaps the dark ones in behind a `@media (prefers-color-scheme: dark)` block, mirroring what the static `-dark-auto` class does. Only tokens that actually differ between the two schemes are emitted, so the block stays small.

This also means `colorScheme` behaves the same whichever stylesheets the page loads — base only, or a theme stylesheet on top, minified or not. Measured on a grid with `colorScheme: 'dark'` (dark background is `rgb(5,5,6)`):

| Stylesheets loaded | Before this fix | After |
| --- | --- | --- |
| base only | `rgb(5,5,6)` ✓ | `rgb(5,5,6)` ✓ |
| + `ht-theme-main.css` | `rgb(5,5,6)` ✓ | `rgb(5,5,6)` ✓ |
| + `ht-theme-main.min.css` | `rgb(255,255,255)` ✗ | `rgb(5,5,6)` ✓ |

`flattenCssVariables()` gained two options to support this: `resolveScheme` picks one branch of a `[light, dark]` value instead of emitting `light-dark()`, and `lightDarkOnly` skips variables that are identical in both schemes. Both default to the previous behavior, so every existing caller is unaffected.

#### Known limit

Both options need the theme engine, so they have no effect when the theme comes from a CSS class name — the `theme` option set to a string, or an `ht-theme-*` class on the container. The grid logs a warning in that case. This is inherent: a class-named theme has no runtime density control at all.

### How has this been tested?

New unit tests covering the override merge, validation, per-instance scoping, the scope class on wrapper and portal, clearing an override, resolving colors to a single scheme, the `auto` media query, and — most importantly — that the shared theme object is not mutated:

```bash
npm run test:unit --prefix handsontable -- --testPathPattern=themes
# 6 suites, 222 tests passed
```

I verified these tests are meaningful rather than merely green: temporarily removing the scope class from the emitted selector turns several of them red, including the cross-instance isolation test.

New Playwright E2E (8 tests) asserting rendered results — computed `color-scheme`, actual cell background color, and measured cell height — not the injected CSS text. Its fixture loads the **minified** theme stylesheet, the strictest configuration:

```bash
cd tests && npx playwright test e2e/color-scheme-density.spec.ts
# 48 passed (8 tests across the full theme x bundle matrix)
```

It covers both options set at construction time and switched via `updateSettings()`, clearing an override, the context menu in the portal following the grid, a control grid on the same page that must not follow another grid's override, and the overrides surviving a theme-engine teardown and rebuild.

Full regression run:

```bash
npm run test:unit --prefix handsontable          # 293 suites, 3425 tests passed
npm run test:types --prefix handsontable         # passed
npm run stylelint --prefix handsontable          # passed
npm run test --prefix wrappers/react-wrapper     # 15 suites, 75 tests passed
npm run test --prefix wrappers/vue3              # 4 suites, 32 tests passed
npm run test --prefix wrappers/angular-wrapper   # 13 suites, 291 tests passed
```

#### Manual testing

Two throwaway demo pages (gitignored, `dev*.html`) are generated for manual review:

```bash
python3 -m http.server 8846 --directory handsontable
# http://localhost:8846/dev-pr.html      — this branch
# http://localhost:8846/dev-latest.html  — released v18.0.0, for comparison
```

Both pages load the minified base and theme stylesheets. The PR page has four grids: A takes the overrides from a button bar, B is the control that must not change, C sets both options at construction time, and D gets its theme from a CSS class name so the options warn instead. Verified in the browser:

| Grid | `color-scheme` | Cell height | Cell background |
| --- | --- | --- | --- |
| A, after Dark + Compact | `dark` | 26px | `rgb(5, 5, 6)` |
| B, control | `light dark` | 30px | `rgb(255, 255, 255)` |
| C, `dark` + `compact` at construction | `dark` | 26px | `rgb(5, 5, 6)` |
| D, theme from a CSS class | `light dark` | 30px | `rgb(255, 255, 255)` |

After those changes `getTheme('main').getThemeConfig()` still reports `colorScheme: 'auto'` and `density: 'default'` — the shared theme object is untouched. Grid D logs the expected warning and does not change. Reset returns grid A to 30px and the theme color scheme.

### Types of changes

### Related issue(s):

1. DEV-1476

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

Documentation updated in this PR: a new "Set the color scheme or density without a theme" section in `docs/content/guides/styling/themes/themes.md`, a cross-reference from "Light and dark modes", and full JSDoc for both options in `metaSchema.ts` (which generates the API reference).

ClickUp task: https://app.clickup.com/t/9015210959/DEV-1476

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 30d550c06c3da78fa6f2fc743584843c62c77ee4. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changelogs/13205.json
+++ b/.changelogs/13205.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Added the `colorScheme` and `density` options, which set the color scheme and the spacing of the grid without declaring a theme.",
+  "type": "added",
+  "issueOrPR": 13205,
+  "breaking": false,
+  "framework": "none"
+}
--- a/docs/content/guides/styling/themes/themes.md
+++ b/docs/content/guides/styling/themes/themes.md
@@ -493,6 +493,8 @@ hot.updateSettings({
 });
 ```
 
+An unsupported value is ignored, and the grid logs a warning naming the option. It does not throw, so one bad value does not break the rest of the update.
+
 To drop an override and go back to the value your theme defines, set the option to `undefined`:
 
 ```js
--- a/docs/content/guides/styling/themes/themes.md
+++ b/docs/content/guides/styling/themes/themes.md
@@ -111,6 +111,8 @@ When using the Theme API, you can configure the color scheme using `setColorSche
 
 When using CSS files, color scheme switching is controlled through CSS class names. Use `ht-theme-{name}` for light mode, `ht-theme-{name}-dark` for dark mode, or `ht-theme-{name}-dark-auto` for automatic switching based on system preferences (e.g., `ht-theme-main`, `ht-theme-main-dark`, `ht-theme-main-dark-auto`).
 
+If you don't want to declare a theme at all, set the [`colorScheme`](@/api/options.md#colorscheme) option instead. See [Set the color scheme or density without a theme](#set-the-color-scheme-or-density-without-a-theme).
+
 
 ## Use a theme
 
@@ -414,6 +416,96 @@ const hotSettings = ref({
 ```
 
 :::
+
+## Set the color scheme or density without a theme
+
+If the only thing you want to change is the color scheme or the amount of white space, you don't have to import, register, and configure a theme. Set the [`colorScheme`](@/api/options.md#colorscheme) or [`density`](@/api/options.md#density) option directly, and the grid applies it on top of the theme it already uses.
+
+| Option        | Allowed values                            | Default                        |
+| ------------- | ----------------------------------------- | ------------------------------ |
+| `colorScheme` | `'light'`, `'dark'`, `'auto'`             | the color scheme of your theme |
+| `density`     | `'default'`, `'compact'`, `'comfortable'` | the density of your theme      |
+
+Both options are per-instance overrides. The theme itself stays unchanged, so other grids that use the same theme keep their own color scheme and density.
+
+::: only-for javascript
+
+```js
+const container = document.querySelector('#handsontable-example');
+
+const hot = new Handsontable(container, {
+  colorScheme: 'dark',
+  density: 'compact',
+  // other options
+});
+```
+
+:::
+
+::: only-for react
+
+```jsx
+<HotTable
+  colorScheme="dark"
+  density="compact"
+/>
+```
+
+:::
+
+::: only-for angular
+
+```html
+<hot-table [settings]="{
+  colorScheme: 'dark',
+  density: 'compact'
+}">
+</hot-table>
+```
+
+:::
+
+::: only-for vue
+
+```ts
+const hotSettings = ref({
+  colorScheme: 'dark',
+  density: 'compact',
+  // ... other options
+  licenseKey: 'non-commercial-and-evaluation',
+});
+```
+
+```html
+<HotTable :settings="hotSettings" />
+```
+
+:::
+
+### Switch the color scheme or density at runtime
+
+Pass either option to [`updateSettings()`](@/api/core.md#updatesettings). This is all you need for a dark mode toggle:
+
+```js
+hot.updateSettings({
+  colorScheme: 'dark',
+});
+```
+
+To drop an override and go back to the value your theme defines, set the option to `undefined`:
+
+```js
+hot.updateSettings({
+  colorScheme: undefined,
+  density: undefined,
+});
+```
+
+### When these options apply
+
+Both options are features of the Theme API, so they need the theme engine to be active. The engine is active when you leave the [`theme`](@/api/options.md#theme) option out, or when you pass a theme config object or a `ThemeBuilder` instance to it. It is not active when the theme comes from a CSS class name — either the `theme` option set to a string such as `'ht-theme-main'`, or an `ht-theme-*` class on the container element. In that case the grid logs a warning and the options have no effect.
+
+One more limit applies to `colorScheme`. If you load a minified theme stylesheet, such as `ht-theme-main.min.css`, its colors are tied to the theme class name and out-specify the ones the engine generates. The color scheme then does not follow the option. The same limit applies to `setColorScheme()`. Density is not affected. To use `colorScheme`, load only the base stylesheet, as described in [Option 1](#option-1-using-the-theme-api-recommended).
 
 ::: only-for angular
 
--- a/docs/content/guides/styling/themes/themes.md
+++ b/docs/content/guides/styling/themes/themes.md
@@ -505,7 +505,7 @@ hot.updateSettings({
 
 Both options are features of the Theme API, so they need the theme engine to be active. The engine is active when you leave the [`theme`](@/api/options.md#theme) option out, or when you pass a theme config object or a `ThemeBuilder` instance to it. It is not active when the theme comes from a CSS class name — either the `theme` option set to a string such as `'ht-theme-main'`, or an `ht-theme-*` class on the container element. In that case the grid logs a warning and the options have no effect.
 
-One more limit applies to `colorScheme`. If you load a minified theme stylesheet, such as `ht-theme-main.min.css`, its colors are tied to the theme class name and out-specify the ones the engine generates. The color scheme then does not follow the option. The same limit applies to `setColorScheme()`. Density is not affected. To use `colorScheme`, load only the base stylesheet, as described in [Option 1](#option-1-using-the-theme-api-recommended).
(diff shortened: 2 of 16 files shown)


qunabu:
Task linked: [DEV-1476 Allow setting colorScheme mode and density without declaring a full theme](https://app.clickup.com/t/9015210959/DEV-1476)

adrianspdev:
@budnix Here is a list of a few things to check:

1. core.ts:1787 initializeThemeManager() — old ThemeManager never torn down; orphan <style> sits later in source order at same (0,2,0) specificity, so construction-time colorScheme/density can never change or reset.
→ instance.themeManager?.destroy() before creating the new one.

2. manager.ts:154 #setOverrides() — String(value) before validation; null/false/0 throw mid-updateSettings, grid half-updated, bad value persists in meta and breaks all later theme switches.
→ Treat null/false/'' as clear; warn-and-ignore invalid values instead of throwing at the settings boundary.

3. core.ts:3470 applyThemeOverrides() call site — renders and fires afterSetTheme before data load, meta cache clear, and column meta; second render follows at 3676.
→ Move call after the meta block, return a flag, let updateSettings' own render cover it, fire afterSetTheme post-render.

4. core.ts:1759 warning guard — keyed on key presence, so the documented reset and every wrapper prop change warn.
→ Warn only when a value is actually requested, once per instance.

5. tests/fixtures/demo/color-scheme-density.html:20 — loads ht-theme-<matrix>.min.css while every grid runs engine main; horizon/classic selectors never match, so the specificity scenario is tested on 2 of 6 legs.
→ Load ht-theme-main.min.css always and pin the spec to one theme leg (keep the bundle axis).

6. manager.ts:226 density block — theme lacking the requested preset yields no CSS; getDensityType() still reports it applied. Silent no-op.
→ warn() when the override is set but the theme has no sizes for it.

budnix:
@adrianspdev Thanks — all six were real, fixed in `30d550c`.

1. `themeManager.destroy()` before the rebuild. Confirmed there were 2 style nodes and the orphan was winning by source order, so a construction-time value really could never change or reset.
2. Empty values (`null`/`false`/`''`) now clear the override, and a bad value warns and keeps the current one instead of throwing mid-update.
3. No render there any more — it returns a flag, and `afterSetTheme` fires after the render `updateSettings` already does.
4. Warns only when a value is actually requested, once per instance.
5. The fixture now always loads `ht-theme-main.min.css`, so all 6 legs test the specificity case instead of 2. I left the legs in place rather than pinning to one — ESLint here bans `test.skip`.
6. Warn added. Only reachable through a custom `getThemeConfig()` object, since `createTheme()` backfills every preset.

New tests are checked both ways: removing the teardown turns 4 of them red, and removing the warning guard produces 2 spurious warnings.

qunabu (dismissed):

adrianspdev:
@budnix A few more things worth checking:

Bug 1 — grid gets permanently stuck on an override. core.ts:3455 and :3461. HIGH.
unmount(); themeManager = null doesn't unsubscribe, so the dead manager still listens to the theme object and re-injects its scoped <style> on the next theme change. Reproduced on your build: A → dark → theme: 'ht-theme-horizon' → getTheme('main').setColorScheme(…) → 2 style nodes, grid stuck dark, colorScheme: 'light' ignored, cellBg rgb(5,5,6). Identical failure to the one the PR just fixed, via the other teardown path. Untested, hence green CI.
→ destroy() at both sites.

Bug 2 — TypeError out of updateSettings. core.ts:3704. LOW (edge, needs a listener that switches theme).
instance.themeManager!.getClassName() — an afterUpdateSettings listener that moves the grid to a class-name theme nulls the manager first. Reproduced: Cannot read properties of null (reading 'getClassName').
→ ?. + skip the hook when null.

Not bugs, but worth folding into the same commit:
- isValueRequested's ?? emits a spurious warning for {colorScheme: null, density: null}, false, '' on class-themed grids — console noise only, behavior correct.
- Density "no sizes" warning repeats on every style re-injection (manager.ts:263).
- tests/fixtures/demo/color-scheme-density.html:59 claims the spec pins itself to the main leg — it doesn't.
- themes.md doesn't state that an unsupported value is now ignored with a warning instead of throwing.

budnix:
@adrianspdev Thanks, both bugs were real — fixed in `12ea400`.

1. `destroy()` at both sites. `unmount()` left the subscription alive, so the dead manager woke up on the next theme change and re-injected its node. Same root cause as last time, through the path I hadn't tested — you were right that green CI meant untested, not correct.
2. Guarded, and the hook is skipped when the manager is gone. The test reproduces your exact `Cannot read properties of null (reading 'getClassName')` before the fix.
3. Pulled the empty-value rule into one exported `isThemeOverrideEmpty()`, so core and the manager agree instead of each having their own idea.
4. The density warning is keyed by theme and density now, so it logs once per real problem.
5. and 6. Comment corrected, and the warn-instead-of-throw behavior is now in `themes.md` and in the JSDoc of both options.

One more I hit while running all six legs: `otherThemeName` came from `?theme=`, so on the horizon and classic legs it resolved to `ht-theme-main` — the name already in use, which core skips as a no-op. The teardown tests were passing there without tearing anything down. Pinned to `horizon`, since every grid in the fixture renders the engine-driven `main` theme.

84/84 E2E across the matrix, 3430 unit, all three wrappers green. Both new tests checked red first.

adrianspdev (approved):

PR #13206, opened 2026-08-20
DEV-1023: Add a public collapse and expand API to the nestedRows plugin

budnix:
### Context

The PR adds a public API for collapsing and expanding parent rows in the `nestedRows` plugin, and makes the plugin report those actions through hooks.

Until now none of this was public. The only working path was `plugin.collapsingUI`, which the source marks `@private`, and the collapse code fired no hooks at all — so an application could not react to a collapse, block it, or save it. Three gaps, all reproduced before the change:

- **No public methods.** Every functional member of `NestedRows` carried `@private`; what was left was the plugin lifecycle only.
- **No hooks.** `ui/collapsing.ts` contained zero `runHooks` calls, even though the same plugin already fires `beforeAddChild` / `afterAddChild` / `beforeDetachChild` / `afterDetachChild`. The grid had `beforeColumnCollapse` and friends for columns, with no row equivalent.
- **The internals reached the published types in 18.0.0.** The TypeScript conversion replaced the handwritten declarations with generated ones. `@private` in JSDoc means nothing to the type system, so `collapsingUI: CollapsingUI | null` is now typed and shipped — the unsupported path compiles cleanly and looks supported.

The customer driver is DEV-1437, who persist collapse state across a full dataset replacement by reaching into `collapsingUI` and `dataManager.cache`, and asked for a supported surface.

Our own documentation was reaching into the same internals: all four framework examples in the parent-child guide called `collapsingUI.collapseChildren(8)`, and the `columnSummary` examples walked `dataManager.cache.levels[0][i].__children` — the Angular one needing four `@ts-ignore` comments to compile. Those examples now use the public API.

#### What the API looks like

Thirteen methods on the plugin:

| Method | Purpose |
| --- | --- |
| `collapseAll()` / `expandAll()` | Collapse every top-level parent; expand every parent at every level |
| `collapseParent(row)` / `expandParent(row)` / `toggleParent(row)` | Act on one parent |
| `getCollapsedParents()` / `isParentCollapsed(row)` | Read the collapsed state |
| `isParent(row)` / `getRowLevel(row)` / `getRowParent(row)` / `countChildren(row, recursive)` | Read the tree structure |
| `expandToRow(row)` / `expandToLevel(level)` | Reveal a hidden row; show rows down to a nesting level |

Four hooks, shaped exactly like their `collapsibleColumns` counterparts — a three-argument cancelable `before*` and a four-argument `after*` whose last argument says whether the state actually changed: `beforeRowCollapse`, `afterRowCollapse`, `beforeRowExpand`, `afterRowExpand`.

All four entry points — the row header button, the <kbd>Enter</kbd> shortcut, the new methods, and collapse-all — now run through a single choke point in `ui/collapsing.ts`, so every one of them produces the same state change and the same hooks.

#### Index types, and why they are not uniform

`collapsedRowsMap` is a `TrimmingMap`, so collapsing a parent *trims* its children. A trimmed row is dropped from the grid and has no visual index at all — `toVisualRow()` returns `null` for it.

That splits the surface in two, and the split is forced rather than chosen:

- Methods acting on a row the user can see take a **visual** row index, following the rule that the public API is visual.
- `getCollapsedParents()` and `expandToRow()` use **physical** row indexes, because the rows they address are the ones the collapse itself hid. A parent collapsed inside another collapsed parent cannot be expressed in visual indexes, so the state would not be representable, let alone storable.
- The hooks carry **physical** indexes, matching the column hooks and matching what an application needs in order to save and restore the state.

`trimRows` is physical throughout for the same reason, while `hiddenRows` is visual throughout because hiding preserves visual indexes. This is recorded in `handsontable/.ai/HOOKS.md` so the next hook payload does not have to rediscover it.

#### Four defects fixed along the way

These are not incidental — the new methods return these values and the new hooks carry them, so the API would have reported corrupt state without them. Each one has a test that was confirmed failing before the fix.

1. **`dataManager.getRowIndex()` threw on a row object it did not recognize** (`this.cache.nodeInfo.get(rowObj)!.row`). `rewriteCache()` builds a fresh `WeakMap` on every `loadData`, `updateData`, `addChild`, and row move, so any object an application held from earlier threw `TypeError: Cannot read properties of undefined (reading 'row')`. This is the DEV-1437 crash, and it is why that customer had to retry inside `requestAnimationFrame`. `getRowObjectParent()` and `getRowObjectLevel()` had the same flaw. All three now return `null`, which their signatures already allowed.
2. **Expanding a parent that was not collapsed corrupted the collapsed list.** `collapsedRows.splice(indexOf(row), 1)` ran as `splice(-1, 1)` when the row was absent, which deletes the *last* entry instead of doing nothing. `expandAll()` calls that path for every parent regardless of state, so it fired routinely.
3. **Rows without children were recorded as collapsed.** Pressing <kbd>Enter</kbd> on a leaf row header pushed it into the collapsed list.
4. **`updateSettings()` discarded the whole collapsed state.** `updatePlugin()` unregisters the trimming map and builds a new `CollapsingUI`, and nothing carried the state across. Because the React wrapper copies every prop into the `updateSettings` payload on each re-render, this fired on **every re-render of the component holding the grid** — the pain DEV-1437 worked around with module-level refs and an `afterRender` resync. The same class of bug was fixed once before and regressed (see `CHANGELOG.md`), so it now has a regression test pinning it. The replay deliberately does not fire the new hooks: it repeats a choice the user already made, and firing them there would report a collapse on every settings update.

#### No breaking changes

Everything is additive. `collapsingUI`, `dataManager`, `headersUI`, `contextMenuUI`, `rowMoveController`, and `collapsedRowsMap` stay exactly where they are and keep working, with no warning — they were unusable from TypeScript before 18.0.0 but are typed and published now, so removing them would break code that compiles today. Our own `columnSummary` and `formulas` specs call those internals, and they still pass.

`collapsingUI.expandAll()` keeps its current behavior, which leaves a parent collapsed inside another one collapsed. The new public `expandAll()` expands every level instead, since a method with that name leaving rows hidden would be surprising. The type test asserts that the internals still compile, so the compatibility promise is machine-checked.

### How has this been tested?

Unit, E2E, Playwright, and type coverage. New E2E is Playwright; the existing Jasmine specs were extended rather than added to.

Each of the four defects was verified by re-introducing the bug, rebuilding, and confirming the new test fails — including reproducing `TypeError: Cannot read properties of undefined (reading 'row')` for the `loadData` case, and 18 visible rows where 13 were expected for the `updateSettings` case.

```bash
# Build first — the E2E runner loads dist/

# Core: 3403 unit tests, 9650 E2E specs, 0 failures
npm --prefix handsontable run test:e2e

# The plugin and the hook registry specifically
npm --prefix handsontable run test:e2e -- --testPathPattern=nestedRows
npm --prefix handsontable run test:unit -- --testPathPattern=hooks

# Types

# Playwright: 18 new tests
cd tests && npx playwright test --project=e2e-main nested-rows-api

# Wrapper parity — the hooks reach all three wrappers automatically
npm --prefix wrappers/react-wrapper run test    # 75 passed
npm --prefix wrappers/vue3 run test             # 33 passed
npm --prefix wrappers/angular-wrapper run test  # 291 passed

```

What the new tests cover:

- Every method, including index translation once collapsing has shifted the visual indexes apart from the physical ones, and a parent nested inside a collapsed parent.
- Each hook firing with its documented arguments, `collapsePossible: false` for a row with no children, and `return false` in each `before*` blocking the action and suppressing the matching `after*`.
- The row header button, the <kbd>Enter</kbd> shortcut, and the methods all producing the same state and the same hooks.
- Calling the methods on the line immediately after `loadData()`, with no retry.
- The collapsed rows surviving `updateSettings({ nestedRows: true })`.
- A row without children never being reported as collapsed — `areChildrenCollapsed()` answers "are all children collapsed", which is vacuously `true` for a row that has none, so this needed its own guard.

Two notes for the reviewer:

- **ESLint could not be run locally.** It fails with `TypeError: minimatch is not a function` inside `eslint-plugin-handsontable`, reproducibly and on untouched files (`src/plugins/trimRows` fails the same way), so it is a pre-existing environment problem rather than something in this branch. The `lint-staged` pre-commit hook reported success despite that crash, so it did not validate either. Stylelint passes. Please rely on CI for the ESLint result — the JSDoc-format and cognitive-complexity rules on the new code are unverified by machine.
- **`@since 18.1.0`** is used in the four new hook doc blocks, on the assumption that the next release is a minor. Worth confirming.

### Types of changes

### Related issue(s):

1. DEV-1023
2. DEV-1437

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

#### Open questions for review

1. **Naming.** `expandToRow` and `expandToLevel` especially — the ticket asked for placeholder names, and these two are not in its list.
2. **A fifth hook argument carrying the parent row object.** The plugin's existing hooks pass row objects (`afterAddChild(parent, element)`). An object survives a dataset swap where an index may not, which would simplify the save-and-restore recipe. Cheap to add if wanted.
3. **One deliberate divergence from the column mirror.** On a repeat collapse, `collapsibleColumns` reports `possible: false` because its check is state-aware; this reports `possible: true, performed: false`, treating `possible` as index validity, as the JSDoc states. Flagging it so it is a decision rather than an accident.
4. **`collapseRow()` / `expandRow()` are deliberately not exposed.** Both exist on `collapsingUI` but never trim and never render, so they are dead code.
5. `getChildren()` is not included. Children can be hidden, so it would need to return physical indexes, and nothing in scope required it — the `columnSummary` examples are expressible with `countChildren()`.

Follow-ups worth their own tickets, all out of scope here:

- `addHook()` accepts hook names that do not exist, with no error and no warning, so a misspelled or not-yet-existing hook is completely silent. This is a core issue, and it is exactly what someone hunting for these missing collapse hooks would hit.
- `mergeCells` currently detects a collapse by watching the row trimming map, because no hook existed. The new hooks are a better signal.
- Parent row headers emit `aria-expanded` but never `aria-level`, so assistive technology cannot tell how deep a row sits. `getRowLevel()` makes that easy to add.
- `visual-tests/src/page-helpers.ts` has a `collapseNestedRow()` helper that clicks and then sleeps 500 ms because there was no event to wait for. It can now wait on `afterRowCollapse`.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-1023

---

> [!NOTE]
> **Medium Risk**
> Touches NestedRows collapse/trim behavior and new cancelable hooks that control which rows are visible. Additive and well tested, but incorrect index handling could hide or restore the wrong rows.
> 
> **Overview**
> Adds a **public collapse/expand API** on `nestedRows` so apps no longer need `collapsingUI`. Methods include `collapseAll`/`expandAll`, `collapseParent`/`expandParent`/`toggleParent`, structure reads (`isParent`, `getRowLevel`, `getRowParent`, `countChildren`), and navigation (`expandToRow`, `expandToLevel`).
> 
> Four hooks (`beforeRowCollapse`/`afterRowCollapse`/`beforeRowExpand`/`afterRowExpand`) fire from a single choke point used by the header button, Enter, and the new methods. Visual indexes are used for visible rows; physical indexes for trimmed parents and hook payloads.
> 
> Also fixes: collapsed state surviving `updateSettings()` (silent replay, important for React re-renders); `getRowIndex`/`getRowObjectParent`/`getRowObjectLevel` returning `null` instead of throwing on stale objects after `loadData`; not recording leaf rows as collapsed; not splicing the last collapsed parent when expanding a row that was not collapsed.
> 
> Docs and column-summary examples now use the public API. Internals stay typed and reachable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 443ce25fb816f5634e9d2bd941269018bb659bdd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-1023 Introduce new public API for the `nestedRows` plugin](https://app.clickup.com/t/9015210959/DEV-1023)
Task linked: [DEV-1437 Customer use (ticket Support got on Husbpot)](https://app.clickup.com/t/9015210959/DEV-1437)

qunabu (dismissed):

demtario (commented):
Reviewed the API surface, the choke point, and the four defect fixes. The index-space split is consistent everywhere I checked — all eight visual-index methods go through `#toPhysicalRow`, and the six `example3.*` variants pass `visualRow` rather than the physical `row` to `getRowLevel()`. The splice and leaf tests assert on `getCollapsedParents()` content, so they genuinely fail on unfixed code.

Three things worth addressing before merge, inline.

handsontable/src/plugins/nestedRows/nestedRows.ts
```diff
@@ -240,6 +251,357 @@ export class NestedRows extends BasePlugin {
...
+      });
+
+      // Expand from the shallowest down, so each parent is reachable when its turn comes, then
+      // collapse from the deepest up.
+      if (toExpand.length > 0) {
+        this.collapsingUI!.toggleCollapsedRows(toExpand, 'expand');
+      }
+
+      if (toCollapse.length > 0) {
+        this.collapsingUI!.toggleCollapsedRows(toCollapse.reverse(), 'collapse');
```
demtario: `expandToLevel()` is one user-facing operation with two independent veto points.

If a `beforeRowExpand` callback returns `false`, the expand pass is cancelled but the collapse pass below still runs. The grid then ends up in a state that is neither the previous one nor the requested level — the deeper parents get collapsed while the shallower ones stay as they were.

No other method on the plugin behaves this way: every sibling routes through a single `toggleCollapsedRows` call, so a veto leaves the state untouched. Nothing in `collapsing.spec.js` or `nested-rows-api.spec.ts` covers this path.

Two ways out:

- Check the first call's return value and bail before the second, so a veto on either pass cancels the whole operation.
- Or keep the current behavior and document on the method that the two passes veto separately, plus add a test pinning it.

The first option matches what the rest of the API promises.

handsontable/src/plugins/nestedRows/nestedRows.ts
```diff
@@ -240,6 +251,357 @@ export class NestedRows extends BasePlugin {
...
+   * Translates a visual row index into a physical one, for the methods that act on a parent row.
+   *
+   * Returns `null` when the plugin is not ready, so the caller can bail out before reaching the
+   * CollapsingUI. Whether the row is really a parent is decided by the CollapsingUI, which reports it
+   * through the hooks.
+   *
+   * @param {number} row Visual row index.
+   * @returns {number|null}
+   */
+  #toPhysicalParentRow(row: number): number | null {
```
demtario, line 611: This guard, and `#toPhysicalRow`'s `!this.dataManager` check, are the only gates on the thirteen new public methods — and neither is cleared when the plugin is disabled.

`disablePlugin()` unregisters the `nestedRows` index map but leaves `collapsingUI` and `collapsedRowsMap` (declared at line 86, assigned at line 124) populated. That part is pre-existing; what is new is that it is now reachable from public API. After `updateSettings({ nestedRows: false })`:

- `collapseParent(0)` passes the `collapsingUI` guard and reaches `toggleCollapsedRows`.
- `trimRows` writes into the unregistered `TrimmingMap`, so the write is a silent no-op — the grid hides nothing.
- `collapsedRows` is still mutated, so `getCollapsedParents()` reports rows as collapsed while every row is visible.
- `afterRowCollapse` fires with `successfullyCollapsed: true`.

So the new API reports a state change that did not happen. `CollapsibleColumns` gates on `this.enabled` for exactly this reason, and the `columnSummary` example in this PR calls `isEnabled()` before touching the plugin, so the disabled state is a scenario the docs already acknowledge.

Suggest either gating the public methods on `this.enabled`, or nulling `collapsingUI` and `collapsedRowsMap` in `disablePlugin()` so the existing guard does the work.

I read this from the source rather than running it, so please confirm the `TrimmingMap` write is silent rather than throwing.

tests/e2e/nested-rows-api.spec.ts
```diff
@@ -0,0 +1,292 @@
...
+    await nestedRows.goto();
+
+    await nestedRows.callPlugin('collapseParent', 0);
+    expect(await nestedRows.visibleNames()).toEqual(['Root A', 'Root B', 'B-1', 'B-2']);
+
+    // The React wrapper puts every prop into updateSettings on each re-render, so this is what
+    // a re-render does to the grid. The collapsed rows used to be lost here.
+    await page.evaluate(() => window.hot.updateSettings({ nestedRows: true }));
+
+    expect(await nestedRows.collapsedParents()).toEqual([0]);
```
demtario, line 289: This test pins the state surviving `updateSettings`, but not the other half of the fix.

`updatePlugin()` passes `shouldRunHooks = false` on purpose, and the PR description makes the case for it: the replay repeats a choice the user already made, and firing the hooks there would report a collapse on every React re-render. That suppression is load-bearing, and nothing asserts it — not here, and not in the `collapsing.spec.js` counterpart. Flipping the `shouldRunHooks` default would leave the whole suite green.

The fixture already records into `window.hookLog`, so it is one assertion:

```ts
await page.evaluate(() => { window.hookLog.length = 0; });
await page.evaluate(() => window.hot.updateSettings({ nestedRows: true }));

expect(await page.evaluate(() => window.hookLog)).toEqual([]);
```

handsontable/src/plugins/nestedRows/nestedRows.ts
budnix: Good spot — fixed in `443ce25fb8`. One wrinkle on the way there: the return value is `performed`, which is also `false` when there is simply nothing to expand, so I added `applyCollapsedRowsChange()` returning `{ performed, vetoed }` and bail only on a real veto.

handsontable/src/plugins/nestedRows/nestedRows.ts
budnix: You were right to ask — I reproduced it, and the map write is silent, it does not throw. Fixed in `443ce25fb8`: all thirteen methods now go through an `#isOperational()` gate that also checks `this.enabled`.

tests/e2e/nested-rows-api.spec.ts
budnix: Agreed, that one was riding along untested. Added it in `443ce25fb8` in both the Playwright spec and `collapsing.spec.js` — dropping the `false` now fails the suite.

demtario (approved):

PR #13208, opened 2026-08-20
DEV-2515: Fix grids created in a container that has no layout yet

Haxikowy:
### Context

The PR fixes a grid built into a container that is not in the layout yet binding its overlays to window scroll permanently, and measuring its sizes against nothing.

Such a container resolves `getComputedStyle()` to an empty declaration for itself and every ancestor — CSSOM behavior that Firefox and Safari always had and Chromium adopted in 151, which is what turned this into a field report. Two shapes reach it: a light-DOM child of a shadow host whose `<slot>` renders a tick later (the customer's Stencil component), and any subtree assembled before it is appended. Walkontable then finds no clipping or scrolling ancestor and binds the overlays to the window for the instance's life: column headers and frozen columns stop following the body, virtualization renders the whole data set, and the row heights are recorded at values the rows never had. Only `updateSettings({})` repaired it.

The fix:

1. **Engine** — `ScrollSync` marks its answer provisional when the table is not rendered; `Overlays#afterDraw` settles it. The pass resolves before rebinding anything and stops retrying when the answer repeats, because `getTrimmingContainer` and `getScrollableElement` can disagree permanently (a table in an iframe driven from the parent realm does). The stale sizes are marked, not dropped: `Overlays#beforeDraw` drops them on the way into the next draw that renders cells, so reset, re-measure and resize run in the documented order.
2. **Core** — `StylesHandler` records whether its caching pass ran against unresolved styles and re-reads on request; `TableView#beforeRender` asks that instead of keying off the `null` row height, which is a state that never resolves on a page with no grid stylesheet.
3. **CSS** — `_filters.scss` declares the by-value list's shorter cells through `--ht-cell-vertical-padding` instead of writing `padding` onto the `td`, so the derived row height matches what the cells render at. Rendered pixels unchanged.

Measured against develop, only the two target shapes change; a grid laid out from the start is identical, down to render and cache-invalidation counts. Nothing is added to `Handsontable.dom`, and no setting, hook, default, i18n file, wrapper or `.d.ts` is touched.

### How has this been tested?

New Playwright suite `tests/e2e/unlaid-out-init-scroll-sync.spec.ts` (fixture + page object): `detached` and `unslotted` construction plus a laid-out control, covering overlay alignment through a scroll, the container filling with no interaction, the pass settling for good, and a `fixedRowsTop`/`fixedRowsBottom` leg. `tests/e2e/filters-value-list-scroll-range.spec.ts` guards the list invariant. New unit specs pin the engine contract (`scrollSync.unit.ts`), the styles-handler transition, and the geometry port; the engine ones were mutation-checked in both directions.

```bash
cd tests && npx playwright test --project=e2e-main e2e/unlaid-out-init-scroll-sync.spec.ts
npm run test:unit --prefix handsontable -- --testPathPattern="scrollSync|stylesHandler|liveGeometryReader"
```

| Suite | Result |
| --- | --- |
| `eslint` / `stylelint` | 0 errors |
| `test:unit` | 3577 / 300 suites |
| `test:walkontable` | 816 specs, 0 failures |
| `test:e2e` (Jasmine) | 9627 specs, 0 failures |
| Playwright `e2e-main` | 287 passed |

`fill-handle-formulas` fails locally here and on a clean develop — it needs `tests/node_modules/hyperformula`.

**Argos**: build 5836's 8 changed `filters/entering-and-escaping-by-value-lists` frames are part 3's doing; the branch now matches released 18.0.0 exactly (393px range for 392px of content), so the next build should be clean.

Follow-ups filed separately: the narrow-container `AutoColumnSize` variant, and the menu tokens (`--ht-menu-item-vertical-padding`) carrying the same shape as part 3.

### Types of changes

### Related issue(s):

1. DEV-2515

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/86cb0yqcx

---

> [!NOTE]
> **High Risk**
> Touches overlay scroll binding, the draw cycle, and row-height/width caches. A mistake here can misalign frozen panes, bind to window scroll, or leave a wrong scroll range for the instance lifetime.
> 
> **Overview**
> Grids built into a container that is not in the layout yet (detached subtree, unslotted light DOM, or `display: none`) no longer stay bound to **window scroll** or keep row heights measured against nothing. Frozen headers/columns follow the body, virtualization stays viewport-sized, and the grid fills its container without a click.
> 
> **Engine:** `ScrollSync` marks the scrollable element provisional when `isRendered()` is false, then `Overlays#afterDraw` settles it on a full draw. Retries stop when the answer repeats (iframe / `overflow: hidden` disagreement). Stale oversized-row and column-width records are dropped on the *next* cell-rendering draw, not in the settle frame.
> 
> **Core:** `StylesHandler` recaches theme variables that were read against empty computed styles; `TableView#render` drops derived sizes *before* `_wt.draw()` so that draw’s calculators use the real default row height.
> 
> **CSS:** Filter-by-value list padding now comes from `--ht-cell-vertical-padding` so derived row height matches the cells (scroll range no longer overshoots).
> 
> Covered by Playwright suites for unlaid-out init and the filter list, plus unit tests for `ScrollSync`, `StylesHandler`, and `LiveGeometryReader`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit baa9b37f00245e7878fed61ef6362b7448b5207c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/handsontable/src/3rdparty/walkontable/src/domMeasure/liveGeometryReader.ts
+++ b/handsontable/src/3rdparty/walkontable/src/domMeasure/liveGeometryReader.ts
@@ -90,6 +90,19 @@ export class LiveGeometryReader implements GeometryReader {
   }
 
   /**
+   * A box test, deliberately not a style test: `display: none` and `content-visibility: hidden`
+   * generate no boxes yet read their styles correctly, and this asks whether the geometry can be
+   * measured. Whether the styles resolve is `StylesHandler`'s question. See the unrendered-table
+   * section of `walkontable/AGENTS.md`.
+   *
+   * @param {HTMLElement} element The element to check.
+   * @returns {boolean}
+   */
+  isRendered(element: HTMLElement): boolean {
+    return element.isConnected && element.getClientRects().length > 0;
+  }
+
+  /**
    * @param {HTMLElement} element The element to measure.
    * @returns {number}
    */
--- a/handsontable/src/3rdparty/walkontable/src/overlay/overlays.ts
+++ b/handsontable/src/3rdparty/walkontable/src/overlay/overlays.ts
@@ -387,9 +387,11 @@ class Overlays {
    */
   beforeDraw() {
     // Before anything measures: drops the sizes a previous draw took while the table had no settled
-    // layout, so this draw re-measures them and resizes from the results. Only on a draw that renders
-    // the cells – a fast draw re-renders nothing, so it cannot re-measure the row heights this drops,
-    // and taking the reset there would leave them dropped. It stays pending for the next full draw.
+    // layout, so this draw re-measures them and resizes from the results. Not on a fast draw – it
+    // re-renders nothing, so it cannot re-measure the row heights this drops, and taking the reset
+    // there would leave them dropped. The mark stays pending until `afterDraw` sees a draw that
+    // actually rendered the band (a `skipRender` hook cancels one that got this far), so the drop is
+    // simply retaken on the next draw that can re-measure.
     if (!this.isScrollDrivenDraw) {
       this.#scrollSync.resetSizesMeasuredBeforeLayoutSettled();
     }
@@ -403,8 +405,21 @@ class Overlays {
 
   /**
    * Runs logic for the overlays after the table is drawn.
+   *
+   * A fast draw never runs `alignOverlaysWithTrimmingContainer`, so the holder still carries the
+   * `overflow: visible` it was born with and the trimming container is whatever the last full draw
+   * left – a layout resolution taken from it would read a table that nothing has aligned yet. No such
+   * draw can currently precede the first full one (`refreshAll()` returns while `drawn` is false, and
+   * a table built outside the layout stays undrawn until it joins it), so the gate protects the
+   * settle test in `ScrollSync#resolveProvisionalLayout` from a state it never has to judge.
+   *
+   * @param {boolean} cellsRendered Whether this draw rendered the cell band. `false` for a fast
+   *                                (scroll) draw and for a draw whose `beforeDraw` hook set
+   *                                `skipRender`. Only a draw that rendered the band re-measured the
+   *                                sizes the reset on the way in dropped, so only it may spend the
+   *                                mark – otherwise the drop is retaken on the next draw.
    */
-  afterDraw() {
+  afterDraw(cellsRendered: boolean) {
     this.syncScrollWithMaster();
     this.#overlays.forEach((overlay) => {
       const hasRenderingStateChanged = overlay.hasRenderingStateChanged();
@@ -416,11 +431,19 @@ class Overlays {
       }
     });
 
+    if (cellsRendered) {
+      this.#scrollSync.confirmSizesRemeasured();
+    }
+
     // Runs after the overlays refreshed their trimming containers and the holder got its final
     // overflow, so a table born outside the layout can settle on the scrollable element and the sizes
     // it would have had if it had been rendered from the start. It cannot run in `beforeDraw`: both
-    // are still stale there, so the scrollable element would settle on the window again.
-    this.#scrollSync.resolveProvisionalLayout();
+    // are still stale there, so the scrollable element would settle on the window again. A fast draw
+    // is stale in the same way – it never aligns the overlays – so it must not resolve anything
+    // either; the flag survives to the next full draw.
+    if (!this.isScrollDrivenDraw) {
+      this.#scrollSync.resolveProvisionalLayout();
+    }
   }
 
   /**
--- a/handsontable/src/3rdparty/walkontable/src/overlay/overlays.ts
+++ b/handsontable/src/3rdparty/walkontable/src/overlay/overlays.ts
@@ -572,7 +572,7 @@ class Overlays {
     const isScrollTriggered = this.isScrollDrivenDraw &&
       (this.verticalScrolling || this.horizontalScrolling);
     // On a pure vertical scroll the bottom overlay (and its inline-start corner) render the same fixed
-    // rows over the same visible columns, so their DOM is unchanged – a full re-render is wasted work and
+    // rows over the same visible columns, so their DOM is unchanged - a full re-render is wasted work and
     // forces an expensive style/layout/paint of the clone subtree on every scroll frame. Reposition them
     // (fast draw) instead. Any horizontal scroll changes the visible columns, and a non-scroll redraw
     // (data, settings, resize) is not scroll-driven, so those paths still trigger a full re-render.
@@ -614,7 +614,7 @@ class Overlays {
    * Re-applies the column-header heights to the master and every header-bearing overlay after the
    * Handsontable-side render-size probe has measured content-driven header heights. The probe runs
    * after the draw completes (once the DOM is final), so the overlays first render at the provided
-   * height and are corrected here to match the master – a synchronous, hook-free reconcile that
+   * height and are corrected here to match the master - a synchronous, hook-free reconcile that
    * replaces the old mid-draw `markOversizedColumnHeaders` measurement. The frozen-overlay sync runs
    * last so a wrapped header inside the frozen region still wins, and the sizes are flushed to the DOM.
    */
--- a/handsontable/src/3rdparty/walkontable/src/overlay/overlays.ts
+++ b/handsontable/src/3rdparty/walkontable/src/overlay/overlays.ts
@@ -182,6 +182,19 @@ class Overlays {
   }
 
   /**
+   * Whether the scrolling element was resolved while the table generated no boxes, so the answer was
+   * taken against nothing and a later draw still has to settle it.
+   *
+   * Covers that answer only, not the sizes measured in the same state. Goes false once a pass settles
+   * it, or once a pass gives up because the answer stopped changing.
+   *
+   * @returns {boolean}
+   */
+  get isScrollableElementProvisional() {
+    return this.#scrollSync.isScrollableElementProvisional;
+  }
+
+  /**
    * Walkontable instance's reference.
    *
    * @protected
@@ -373,6 +386,14 @@ class Overlays {
    * Runs logic for the overlays before the table is drawn.
    */
   beforeDraw() {
+    // Before anything measures: drops the sizes a previous draw took while the table had no settled
+    // layout, so this draw re-measures them and resizes from the results. Only on a draw that renders
+    // the cells – a fast draw re-renders nothing, so it cannot re-measure the row heights this drops,
+    // and taking the reset there would leave them dropped. It stays pending for the next full draw.
+    if (!this.isScrollDrivenDraw) {
+      this.#scrollSync.resetSizesMeasuredBeforeLayoutSettled();
+    }
+
     this.#scrollSync.setRenderingStateChanged(this.#overlays.reduce((acc, overlay) => {
       return overlay.hasRenderingStateChanged() || acc;
     }, false));
@@ -394,6 +415,12 @@ class Overlays {
         overlay.reset();
       }
     });
+
+    // Runs after the overlays refreshed their trimming containers and the holder got its final
+    // overflow, so a table born outside the layout can settle on the scrollable element and the sizes
+    // it would have had if it had been rendered from the start. It cannot run in `beforeDraw`: both
+    // are still stale there, so the scrollable element would settle on the window again.
+    this.#scrollSync.resolveProvisionalLayout();
   }
 
   /**
@@ -522,7 +549,7 @@ class Overlays {
     const isScrollTriggered = this.isScrollDrivenDraw &&
       (this.verticalScrolling || this.horizontalScrolling);
     // On a pure vertical scroll the bottom overlay (and its inline-start corner) render the same fixed
-    // rows over the same visible columns, so their DOM is unchanged - a full re-render is wasted work and
+    // rows over the same visible columns, so their DOM is unchanged – a full re-render is wasted work and
     // forces an expensive style/layout/paint of the clone subtree on every scroll frame. Reposition them
     // (fast draw) instead. Any horizontal scroll changes the visible columns, and a non-scroll redraw
     // (data, settings, resize) is not scroll-driven, so those paths still trigger a full re-render.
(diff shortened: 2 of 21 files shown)


qunabu:
Task linked: [HOT-15493 [Forum] Scrolling issue after recent Chromium update](https://app.clickup.com/t/9015210959/HOT-15493)

handsontable/src/3rdparty/walkontable/src/overlay/overlays.ts
```diff
@@ -394,6 +415,12 @@ class Overlays {
         overlay.reset();
       }
     });
+
+    // Runs after the overlays refreshed their trimming containers and the holder got its final
+    // overflow, so a table born outside the layout can settle on the scrollable element and the sizes
+    // it would have had if it had been rendered from the start. It cannot run in `beforeDraw`: both
+    // are still stale there, so the scrollable element would settle on the window again.
+    this.#scrollSync.resolveProvisionalLayout();
```
budnix: Can I ask — should this be gated on `!this.isScrollDrivenDraw`, the same way the reset in `beforeDraw` is? A fast draw skips `alignOverlaysWithTrimmingContainer()`, so `topOverlay.trimmingContainer` still holds the value from construction, which for an unlaid-out table is `window`. If the page joins the grid to the layout and the user scrolls the window before the ResizeObserver's rAF fires a full draw, that scroll goes through `refreshAll()` → `wot.draw(true)`. The holder's overflow is still stale there, so `#computeScrollableElement()` returns `window`, and the stale trimming container makes `settles` true — the grid binds to the window for good, which is the exact DEV-2515 symptom. Skipping the pass costs nothing since the flag survives to the next draw.

```suggestion
    if (!this.isScrollDrivenDraw) {
      this.#scrollSync.resolveProvisionalLayout();
    }
```

budnix (commented):
Read through the layout-settling path. The hot path looks fine — on a normal, visible grid the three new call sites are just boolean flag reads, and the new `getClientRects()` / `getComputedStyle()` calls only run at construction or while a grid is flagged as born without a layout. Five notes inline, the `afterDraw` gating one being the one I would fix before merge.

handsontable/src/3rdparty/walkontable/src/overlay/scroll/scrollSync.ts
```diff
@@ -373,15 +494,36 @@ export class ScrollSync {
...
+   * A table outside the layout resolves every ancestor style to an empty declaration, which reads as
+   * "no ancestor clips or scrolls" and hands the whole grid to the window. A provisional answer is
+   * retaken on the next draw that finds the table rendered (see `Overlays#afterDraw`).
+   *
+   * @returns {HTMLElement | Window}
+   */
+  #takeScrollableElement(): HTMLElement | Window {
+    const { wtTable, geometryReader } = this.#deps;
+
+    this.#isScrollableElementProvisional = !geometryReader.isRendered(wtTable.wtRootElement);
```
budnix, line 558: `#lastProvisionalScrollableElement` keeps its old value when the flag is re-armed here. `updateMainScrollableElements()` is public, and `core.ts:3602` calls it whenever `updateSettings` moves `height` to or from `''`. So a grid that already gave up once (last answer = `window`) and then gets re-armed while still hidden will compute `window` on its first retry, see it matches, and give up right away — the one free retry the design counts on is gone.

```suggestion
    this.#isScrollableElementProvisional = !geometryReader.isRendered(wtTable.wtRootElement);
    this.#lastProvisionalScrollableElement = null;
```

handsontable/src/3rdparty/walkontable/src/domMeasure/liveGeometryReader.ts
```diff
@@ -89,6 +89,19 @@ export class LiveGeometryReader implements GeometryReader {
...
+  /**
+   * A box test, deliberately not a style test: `display: none` and `content-visibility: hidden`
+   * generate no boxes yet read their styles correctly, and this asks whether the geometry can be
+   * measured. Whether the styles resolve is `StylesHandler`'s question. See the unrendered-table
+   * section of `walkontable/AGENTS.md`.
+   *
+   * @param {HTMLElement} element The element to check.
+   * @returns {boolean}
+   */
+  isRendered(element: HTMLElement): boolean {
```
budnix, line 101: A grid built inside a hidden tab or a closed modal is `display: none`, so this returns false — but its computed styles resolve fine, so its scrolling element was already correct. It pays for the correction anyway: `resolveProvisionalLayout()` passes the flag check on every draw and calls `getClientRects()`, which forces a layout read right after the draw dirtied the DOM. Then on reveal it rebinds every overlay listener, and the draw after that wipes the oversized-row map and the column-width cache — both prefix-sum caches rebuilt over every row and column. Hidden tabs are a common setup, and the fixture has a `hidden` mode that no spec uses. Would be good to measure the reveal cost on a big data set, or exclude `display: none` here.

handsontable/src/tableView.ts
```diff
@@ -1533,11 +1533,42 @@ class TableView {
...
+   */
+  #discardSizesMeasuredWithoutStyles() {
+    if (!this.hot.stylesHandler.recacheValuesMeasuredWithoutStyles()) {
+      return;
+    }
+
+    // `resetAllOversizedRows` already invalidates the row-height cache, so only the column widths are
+    // left to drop. `invalidateIndexSizesCache()` would invalidate the row heights a second time, and
+    // this is the same pair the engine-side reset performs.
+    this._wt.wtViewport.resetAllOversizedRows();
```
budnix, line 1586: The engine-side twin runs from `Overlays#beforeDraw`, which is the first thing in `runMasterDrawCycle` (`drawCycle.ts:118`) — before `beginDrawLayout()` at 137 and `createCalculators()` at 148. This one runs from the `beforeDraw` setting at `drawCycle.ts:194`, after both. So this draw's row band was already built from the bad heights, and dropping the caches here does not fix that frame — only the post-render `createVisibleCalculators()` recovers, and nothing schedules a follow-up draw. Did you check how the grid looks on that one frame? Also, when both paths trigger, the same wipe runs on two draws in a row.

handsontable/src/3rdparty/walkontable/src/overlay/overlays.ts
```diff
@@ -373,6 +386,14 @@ class Overlays {
    * Runs logic for the overlays before the table is drawn.
    */
   beforeDraw() {
+    // Before anything measures: drops the sizes a previous draw took while the table had no settled
+    // layout, so this draw re-measures them and resizes from the results. Only on a draw that renders
+    // the cells – a fast draw re-renders nothing, so it cannot re-measure the row heights this drops,
+    // and taking the reset there would leave them dropped. It stays pending for the next full draw.
+    if (!this.isScrollDrivenDraw) {
+      this.#scrollSync.resetSizesMeasuredBeforeLayoutSettled();
```
budnix, line 396: `!isScrollDrivenDraw` is not quite the same as "this draw renders cells". If a `beforeViewRender` hook sets `skipRender`, the draw never reaches `renderCellBand`, so `markOversizedRows` never runs — but the flag is already cleared and the heights are already gone. NestedRows does this internally, and any user hook can too. Same one-shot risk in `TableView#beforeRender`, which also burns its flag before it knows whether the render will happen. Worth a quick check whether the flag can be cleared only after the band actually rendered.

handsontable/src/3rdparty/walkontable/src/overlay/scroll/scrollSync.ts
```diff
@@ -319,6 +358,88 @@ export class ScrollSync {
...
+    if (!this.#sizesMeasuredBeforeLayoutSettled) {
+      return;
+    }
+
+    this.#sizesMeasuredBeforeLayoutSettled = false;
+
+    const wtViewport = this.#deps.getWtViewport();
+
+    wtViewport.resetAllOversizedRows();
+    wtViewport.invalidateColumnWidthCache();
```
demtario, line 403: This drops the prefix sum only. `columnWidthCache`'s `sizeFn` is `wtTable.getColumnWidth`, so rebuilding re-asks `modifyColWidth` — i.e. `AutoColumnSize`'s own cache, which neither this reset nor `TableView#discardSizesMeasuredWithoutStyles` clears. Widths measured while the table had no layout therefore survive. Probably the deferred narrow-container `AutoColumnSize` variant, but the comments here read as if the widths are dropped.

handsontable/src/3rdparty/walkontable/src/overlay/overlays.ts
```diff
@@ -522,7 +549,7 @@ class Overlays {
     const isScrollTriggered = this.isScrollDrivenDraw &&
       (this.verticalScrolling || this.horizontalScrolling);
     // On a pure vertical scroll the bottom overlay (and its inline-start corner) render the same fixed
-    // rows over the same visible columns, so their DOM is unchanged - a full re-render is wasted work and
+    // rows over the same visible columns, so their DOM is unchanged – a full re-render is wasted work and
```
demtario: Nit: `-` → `–` on comment lines this PR does not otherwise touch (also line 594 and `tableView.ts:1585`) adds diff churn.

handsontable/src/3rdparty/walkontable/src/overlay/overlays.ts
Haxikowy: Gated it in 657419331, but the scenario does not reach the settle test. `refreshAll()` returns while `drawn` is `false`, and a grid built outside the layout stays undrawn until it joins one, so no fast draw can precede the first full draw.

Checked it in the browser rather than by reading: attach plus a window scroll dispatched in the same task produces no draw at all, and `isScrollableElementProvisional` stays `true`. I wrote the Playwright case for it, watched it pass with the gate removed, and dropped it instead of committing a green that proves nothing.

The gate stays, since a draw that aligned nothing has no business judging the layout. The JSDoc and the AGENTS.md paragraph say that, including the reachability, so nobody reads it as a live defect later.

handsontable/src/3rdparty/walkontable/src/overlay/scroll/scrollSync.ts
Haxikowy: Correct, fixed in 657419331. `#takeScrollableElement()` now clears the remembered answer when it re-arms the flag, so a new series of retries does not match what the previous one gave up on.

Unit test covers the path you named: give up while rendered, hide the table, `updateMainScrollableElements()`, then render and resolve. The retry survives.

handsontable/src/3rdparty/walkontable/src/domMeasure/liveGeometryReader.ts
Haxikowy: Measured it on the fixture instead of guessing. A hidden grid takes zero draws, since core defers rendering until the root element is visible, so the per-draw `getClientRects()` is not paid at all. Reveal is one full draw, the one the reveal required anyway, and it carries the rebind plus one cache wipe.

Not excluding `display: none` here: no boxes is exactly what the unslotted case looks like, so the only available split is "do the styles resolve", which hands the geometry question to `StylesHandler` and blurs the line the two checks are meant to keep.

Your other point was right and is fixed in a8a6aced6. `hidden` is in `MODES` now, so all three parameterized tests run it.

handsontable/src/tableView.ts
Haxikowy: Confirmed, and it was the real bug of the five. `beforeDraw()` at 118, `beginDrawLayout()` at 137, `createCalculators()` at 148, the `beforeDraw` setting at 194. Moved the discard into `TableView#render`, ahead of `_wt.draw()`, in 8db360874. New unit test pins the order (drop before calculators) and fails on the old placement.

On the frame itself: I did not look at it. The band was built from heights that were about to be dropped and nothing scheduled a follow-up, so it was wrong regardless of how bad it looked.

Both paths can still fire on separate draws. Left it: each is a cache invalidation, both idempotent.

handsontable/src/3rdparty/walkontable/src/overlay/overlays.ts
Haxikowy: Fixed on both sides. The engine mark is spent by `confirmSizesRemeasured()` from `afterDraw`, and only when the draw cycle reports the band rendered (657419331). The core drop stays pending until `afterRender`, which the engine fires from no other kind of draw (8db360874). Wipes are idempotent, so retaking one costs a single invalidation.

One correction to the framing: the heights are not gone for good. The next draw that renders re-runs `markOversizedRows` and takes them again. The window is between the cancelled draw and that one, where the DOM keeps its old content while the caches read default heights. That is why this is a deferral and not anything larger.

handsontable/src/3rdparty/walkontable/src/overlay/scroll/scrollSync.ts
Haxikowy: Right on both counts. `invalidateColumnWidthCache()` drops the prefix sum, the rebuild goes back through `modifyColWidth`, and `AutoColumnSize` answers from `columnWidthsMap`, which neither reset clears. `#onBeforeRender` does re-measure with `overwriteCache = hot.forceFullRender`, but only over the visible range, so the columns sampled while the grid was window-bound (all of them, that being the defect) keep their widths off-screen. That is case 3 in the demo and the narrow-container follow-up filed with the PR.

The comments were the actual problem, since they read as if the widths were dropped. Reworded in baa9b37f0: the drop covers the engine own oversized-row heights and width prefix sum, and the surviving `AutoColumnSize` measurement is named as the follow-up.

handsontable/src/3rdparty/walkontable/src/overlay/overlays.ts
Haxikowy: Fair, reverted in baa9b37f0. All three lines (`overlays.ts:575`, `overlays.ts:617`, `tableView.ts:1610`) are back to `-` and the one-line hunks are out of the diff.

demtario (approved):

budnix (approved):

PR #13201, opened 2026-08-19
DEV-2124: Fix autofill column resolution near horizontally merged cells

budnix:
### Context

The PR fixes autofill column resolution when a horizontal merge sits in the first visible row. This is the X-axis mirror of the Y-axis problem fixed in DEV-2115 (#13105); that fix left the column axis untouched.

`getCellCoordsFromMousePosition` maps a pointer X to a column by walking one reference row with `findColumnAtX`, and it always used the raw `firstPartiallyVisibleRow`. When that row holds a horizontal merge, two things go wrong:

1. **The band collapses onto its anchor.** The anchor cell's `offsetWidth` covers the whole merged band, so every X inside the band resolves to the anchor column. Dragging a fill handle sideways in a row below the merge resolves back to the source cell and fills nothing.
2. **The band width is over-counted.** Once the merge anchor scrolls out of view, none of the scanned slave columns reports a `colspan`, so `findColumnAtX` cannot skip the band. Every slave resolves to the same wide element, whose full width is counted once per slave column — pushing every later column to the left.

The existing merge autofill spec missed this because all of its merges start at row 1.

### The fix

A `findColumnReferenceRow` analog of `findRowReferenceColumn`, in `handsontable/src/helpers/dom/cellCoords.ts`. Each column-scan branch (frozen start columns, and the scrollable body) picks the first row within its own column range whose cells each occupy a single column — detected both by `colspan > 1` on the anchor's meta and by element identity, so a merge whose anchor sits outside the scanned range is still caught. A candidate row must render at least one cell, so a hidden row is skipped rather than picked.

`findColumnAtX` itself is unchanged, matching how DEV-2115 left `findRowAtY` alone and relied on a clean reference axis.

### How has this been tested?

Every new test was confirmed to **fail before** the fix and pass after.

Unit — 5 new cases in `cellCoords.unit.js`, all red on the unfixed helper:

| Case | Expected column | Before the fix |
|---|---|---|
| Pointer over column 2, merge A1:C1 | 2 | 0 (collapsed to anchor) |
| Pointer over column 1, same merge | 1 | 0 |
| Merge anchor scrolled left of view, pointer over column 3 | 3 | 2 (width over-count) |
| Hidden row 1 below the merged row 0 | 2 | 0 |
| Merge inside the frozen start columns | 1 | 0 |

[...]

PR #13210, opened 2026-08-20
DEV-2578: Compose preview packages with the release packaging script

demtario:
### Context

The PR fixes the pkg.pr.new preview packages, which shipped an `exports` map generated **before** the theme stylesheets and the UMD bundles were merged into the package. `import 'handsontable/styles/ht-theme-main.min.css'` against any PR build failed with `ERR_PACKAGE_PATH_NOT_EXPORTED` even though the file was in the tarball, and no preview carried `dist/themes/` bundles at all. Any demo, repro or integration test pinned to a PR preview could not load a theme through the package, so a PR's visual changes could not be shown.

Measured before the fix — preview for PR #13201 vs. the published `handsontable@experimental`:

| tree | export keys | theme + icon CSS keys | `dist/themes` files |
|---|---|---|---|
| npm `experimental` `0.0.0-next-dca019a-20260820` | 271 | 16 | 34 |
| pkg.pr.new 13201 | 217 | 0 | 0 |

Two independent causes, both in CI:

1. **A stale map.** `build.yml`'s `es-cjs` job runs `npm run postbuild` while `styles/` holds only the base stylesheet and `dist/` only `languages/`. That map is frozen into the artifact. The preview job then copied the real `dist/` and `styles/` over the extracted tree with `cp -R` and published — files arrived, entries did not.
2. **No theme bundles anywhere.** `dist/themes/**` is emitted only by `build:themes-umd` / `build:themes-umd.min`, which are in the `build` pipeline but were not in the UMD job, so `handsontable-build-umd` — the only artifact in the repo carrying `handsontable/dist` — never held them.

Nothing reported either, because `prepare-package-for-publish.mjs` validated one direction only (map entry → file exists) and merely warned when a `handsontable.copy` source was missing.

The fix is structural rather than a patch on the copy commands: the preview job now composes the package with the same script the release uses, and that script enforces the package definition instead of silently accepting an incomplete tree.

- **`prepare-package-for-publish.mjs` is strict by default.** It fails when a `handsontable.copy` source does not exist, when a copied entry did not land in `tmp/`, or when a string rule in `handsontable.exports` matches no file. `--partial` downgrades those three to warnings.
- **`build.yml`** — the `umd` job builds `build:themes-umd` and `build:themes-umd.min`, so the artifact matches the release definition of the UMD output for all of its consumers. The `es-cjs` job composes an intentionally incomplete tree and is the single caller allowed to opt out, through the new `postbuild:partial` script; a unit test keeps it the only one.
- **`integration.yml`** — the preview job drops the hand-rolled merge and runs `npm run postbuild` after the artifacts land and before the publish. The script copies the `handsontable.copy` list *and* regenerates the map from the merged tree.

After the change the preview and release packages come out of the same composer under the same rules. As a side effect the preview stops shipping `dist/hyperformula/**` and `handsontable.js.map`, which the release copy list excludes.

Out of scope, worth a follow-up: `emitted-types.yml` keeps a hand-rolled merge that does `rm -rf handsontable/tmp/styles` before copying, which deletes the generated `handsontableStyles.{js,mjs}` module. Harmless there today — that job only typechecks emitted declarations.

### How has this been tested?

**Composed tree matches the published package exactly.** After a full local build, the composed `handsontable/tmp/` file list was compared with the `handsontable@experimental` tarball: **3492 files on both sides, identical set**. The regenerated map reports 271 export keys, 16 theme/icon CSS keys and 34 `dist/themes` keys — the same numbers as the published package, against 217 / 0 / 0 in the current preview.

```bash
node -e "const k=Object.keys(require('./handsontable/tmp/package.json').exports); console.log(k.length, k.filter(x=>/ht-theme|ht-icon/.test(x)).length, k.filter(x=>x.includes('dist/themes')).length)"
# 271 16 34
```

**The strict checks were exercised against a real incomplete tree**, not only fixtures:

```bash
mv handsontable/dist/themes /tmp/themes-bak
npm run postbuild --prefix handsontable          # exit 1: "The composed package is incomplete: … dist/themes"
npm run postbuild:partial --prefix handsontable  # exit 0, warning only
mv /tmp/themes-bak handsontable/dist/themes
npm run postbuild --prefix handsontable          # exit 0
```

**Automated:**

- New `handsontable/test/__tests__/previewPackaging.unit.js` — 8 tests. Four assert the CI wiring (the UMD job builds the theme bundles; `build.yml` is the only workflow using `postbuild:partial`; the preview job composes after the artifacts land and before the publish; nothing copies into `handsontable/tmp` afterwards). Four exercise the strict and partial modes against throwaway package trees.
- `npm run test:unit --prefix handsontable` — full suite green.
- `npm run eslint --prefix handsontable` — clean.

**Every implicit caller of the now-strict `postbuild` was enumerated** (`npm run all build` in `publish.yml`, `build-all.yml`, `docs-*`; `npm run build --prefix handsontable` in `emitted-types.yml` and the Angular typecheck action). All build Handsontable in full from source, so none is affected; `integration.yml`'s wrapper build excludes `handsontable` and never triggers it.

**End-to-end** — this PR's own preview package can be checked once CI publishes it, since `test.yml` calls `integration.yml` by local path:

```bash
curl -sL -o pr.tgz "https://pkg.pr.new/handsontable@<this PR>"
tar -xzf pr.tgz -O package/package.json > pkg.json
node -e "const k=Object.keys(require('./pkg.json').exports); console.log(k.length, k.filter(x=>/ht-theme|ht-icon/.test(x)).length, k.filter(x=>x.includes('dist/themes')).length)"
# expected: 271 16 34
```

### Types of changes

### Related issue(s):

1. DEV-2578

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2578

---

> [!NOTE]
> **Medium Risk**
> Changes CI packaging and npm export composition for all PR previews; incorrect wiring would break consumer imports again, but scope is build/publish scripts rather than runtime product logic.
> 
> **Overview**
> Fixes **DEV-2578** by aligning pkg.pr.new preview packages with release packaging so theme CSS imports and **`dist/themes`** UMD bundles work from PR builds.
> 
> **`prepare-package-for-publish.mjs`** now **fails by default** when the composed `tmp/` tree is incomplete (missing **`handsontable.copy`** destinations or export globs that match nothing). **`--partial`** / **`postbuild:partial`** keeps those as warnings for jobs that intentionally ship a partial tree.
> 
> **CI:** the UMD build job adds **`build:themes-umd`** / **`.min`**; the ES+CJS job and visual workflow use **`postbuild:partial`**. The integration preview job **drops manual `cp` merges** and runs strict **`npm run postbuild`** after UMD **`dist/`** and theme **`styles/`** land, so the **`exports`** map is regenerated from the merged tree before **`pkg-pr-new publish`**.
> 
> New **`previewPackaging.unit.js`** tests lock in workflow wiring and completeness behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b408220ddc731f46695ac7a5d61d69f96e1f9b29. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,6 +48,10 @@ jobs:
           npm run build:umd.min
           npm run build:languages
           npm run build:languages.min
+          # dist/themes/** comes only from these two tasks. Without them the artifact
+          # is not the full UMD output, and the preview package ships no theme bundles.
+          npm run build:themes-umd
+          npm run build:themes-umd.min
       - run: tar -zcf dist.tar.gz ./handsontable/dist ./handsontable/styles
       - name: Upload the Handsontable UMD build artifact.
         uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # https://github.com/actions/upload-artifact/releases/tag/v7.0.0
@@ -76,7 +80,10 @@ jobs:
           npm run build:languages.es
           npm run build:languages
           npm run build:languages.min
-          npm run postbuild
+          # This job composes an intentionally incomplete tree (no UMD dist/, no
+          # theme CSS), so it opts out of the strict packaging checks. It is the
+          # only site allowed to do so — see prepare-package-for-publish.mjs.
+          npm run postbuild:partial
       - run: tar -zcf tmp.tar.gz ./handsontable/tmp
       - name: Upload the Handsontable ES + CJS build artifact.
         uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # https://github.com/actions/upload-artifact/releases/tag/v7.0.0
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -117,16 +117,18 @@ jobs:
         uses: actions/download-artifact@70fc10c6e5e1ce46ad2ea6f2b72d43f7d47b13c3 # https://github.com/actions/download-artifact/releases/tag/v8.0.0
         with:
           name: handsontable-build-umd
-      - name: Merge dist and styles into the preview package
+      # `npm run postbuild` (prepare-package-for-publish) does the merging itself:
+      # it copies the `handsontable.copy` list into tmp/ AND regenerates the exports
+      # map from the merged tree. It must run here, after the UMD dist/ and the theme
+      # CSS land — the map baked into the es-cjs artifact was globbed when only the
+      # base stylesheet existed, so copying the files in by hand shipped stylesheets
+      # that resolved with ERR_PACKAGE_PATH_NOT_EXPORTED (DEV-2578). Wrappers link
+      # against handsontable/tmp, so this precedes the wrapper build.
+      - name: Compose the preview package
         if: env.HOT_PREBUILT == 'true'
         run: |
           tar -zxf dist.tar.gz ./handsontable/dist ./handsontable/styles && rm dist.tar.gz
-          rm -rf handsontable/tmp/dist
-          cp -R handsontable/dist handsontable/tmp/dist
-          # Merge the CSS into tmp/styles instead of replacing the directory —
-          # tmp/styles also holds the generated handsontableStyles.{js,mjs,d.ts}
-          # module that utils/stylesHandler imports at runtime.
-          cp -R handsontable/styles/. handsontable/tmp/styles/
+          cd handsontable && npm run postbuild
       - name: Build wrapper packages (handsontable prebuilt)
         if: env.HOT_PREBUILT == 'true'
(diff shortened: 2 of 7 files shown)


qunabu:
Task linked: [DEV-2578 Core: pkg.pr.new preview packages ship a stale exports map — no theme/icon CSS subpaths, no dist/themes](https://app.clickup.com/t/9015210959/DEV-2578)

budnix (approved):

qunabu (approved):

Issue #13120, opened 2026-07-25
Momentum-scroll timers on overlay clones are not cleared by `destroy()`, throwing `The "runHooks" method cannot be called because this Handsontable instance has been destroyed`


yamaaaaaa31:
# Describe the bug

On touch devices, destroying an instance while a momentum (inertial) scroll is still being tracked throws an uncaught error ~200 ms later:

```
Error: The "runHooks" method cannot be called because this Handsontable instance has been destroyed
```

This happens whenever the grid is unmounted during (or shortly after) a flick-scroll — e.g. the user flicks the grid and immediately navigates away, or a React component containing `<HotTable>` unmounts while the scroll is still coasting.

## Root cause

`Event#onHolderScroll()` schedules a 200 ms timer on every `scroll` of the holder and stores it on the instance:

```js
// 3rdparty/walkontable/src/event.mjs
onHolderScroll() {
  if (!this.momentumScrolling) {
    this.momentumScrolling = {};
  }
  ...
  this.momentumScrolling._timeout = setTimeout(() => {
    if (!this.touchApplied) {
      this.momentumScrolling.ongoing = false;
      this.#wtSettings.getSetting('onAfterMomentumScroll');
    }
  }, 200);
}
```

`Event#destroy()` clears that timer correctly:

```js
// 3rdparty/walkontable/src/event.mjs
destroy() {
  ...
  if (this.momentumScrolling) {
    clearTimeout(this.momentumScrolling._timeout);
  }
  this.#eventManager.destroy();
}
```

But `destroy()` is only ever reached for the **master** `Event`:

```js
// 3rdparty/walkontable/src/core/_base.mjs
destroy() {
  this.wtOverlays.destroy();
  this.wtEvent.destroy();      // <- master only
}
```

Each overlay clone owns its own `Event` instance (`Clone` constructor: `this.wtEvent = new Event(...)`), and `registerEvents()` attaches the `scroll` → `onHolderScroll` listener to the clone's holder too (touch branch). However, the overlay teardown never destroys it:

```js
// 3rdparty/walkontable/src/overlay/_base.mjs
destroy() {
  this.clone?.eventManager.destroy(); // todo check if it is good place for that operation
}
```

So each clone's pending 200 ms timer survives `Core#destroy()`. When it fires, it reaches:

```js
// tableView.mjs
onAfterMomentumScroll: () => this.hot.runHooks('afterMomentumScroll'),
```

[...]

PR #13211, opened 2026-08-20
DEV-2582: Clear momentum-scroll timers on overlay clones on destroy

budnix:
### Context

The PR fixes an uncaught error on touch devices when the grid is destroyed while a momentum (flick) scroll is still coasting:

```
Error: The "runHooks" method cannot be called because this Handsontable instance has been destroyed
```

It reproduces when the grid is unmounted during or shortly after a flick-scroll — the user flicks the grid and immediately navigates away, or a React component holding `<HotTable>` unmounts while the scroll is still coasting.

**Root cause.** Every overlay clone builds its own `Event` instance (`walkontable/src/core/clone.ts`). On touch devices that clone's holder gets a `scroll` listener which arms a 200 ms momentum-scroll timer.

`Event#destroy()` clears that timer, but the overlay teardown only destroyed the clone's `eventManager`, never its `wtEvent` — so `Event#destroy()` was called **zero times** for clones. Only the master's was reached, through `CoreAbstract#destroy()`.

The surviving timer then fires after the instance is gone:

`event.ts` timer → `getSetting('onAfterMomentumScroll')` → `tableView.ts` → `Core.runHooks` → throw.

With `fixedRowsTop: 1` and `fixedColumnsStart: 1` there are 5 overlay clones, so the error appears 5 times.

**Fix.** Destroy the clone's `Event` in `Overlay#destroy()` (`walkontable/src/overlay/regions/_base.ts`). This also clears the clones' double-click and long-press timers, which had the same lifetime problem.

Two notes for review:

- The linked issue quotes old `.mjs` paths. The engine has since moved to TypeScript and been restructured, but the defect survived the move unchanged.
- The narrowing uses `instanceof Clone` rather than a cast. `WalkontableInstance['wtEvent']` is typed as a loose union (`WalkontableEvent | Record<string, unknown>`), and `Clone` — already imported in this file, and what actually constructs the clone a few lines above — types `wtEvent` properly.

### How has this been tested?

Four new Jest unit tests. Two of them are red before the fix and green after (verified by reverting the source change and re-running): the momentum timer firing post-destroy, and the clone `Event#destroy()` never being called. A third covers a double teardown, which `Overlays#refreshAll()` can trigger when the table is detached from the DOM.

```bash
npm run test:unit --prefix handsontable --testPathPattern=cloneEventTeardown
```

Regression runs:

```bash
# Walkontable browser suite — 816 specs, 0 failures.
# Matched against a 0-failure baseline with the fix reverted.

# Full Jest unit suite — 3555 pass. The 5 failures in 3 suites are pre-existing
# and unrelated (they need build-time env vars for version / release date);
# confirmed identical with the fix reverted.

npm run eslint --prefix handsontable   # 0 errors
```

Not run: the legacy Jasmine E2E suite. The walkontable browser suite already tears down a real grid 816 times over the changed path, so CI's full matrix is the right place for that check.

### Types of changes

### Related issue(s):

1. DEV-2582
2. #13120

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2582

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 5358eee568b74893eb2e0c47c623046083bc03d7. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changelogs/13120.json
+++ b/.changelogs/13120.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed an error thrown on touch devices when the grid was destroyed while a momentum scroll was still coasting.",
+  "type": "fixed",
+  "issueOrPR": 13120,
+  "breaking": false,
+  "framework": "none"
+}
--- a/handsontable/src/3rdparty/walkontable/src/overlay/regions/_base.ts
+++ b/handsontable/src/3rdparty/walkontable/src/overlay/regions/_base.ts
@@ -595,6 +595,15 @@ export abstract class Overlay {
    * Destroy overlay instance.
    */
   destroy() {
+    // Every clone owns its own `Event` instance, and only the master's is reached by
+    // `CoreAbstract#destroy()`. Destroying it here clears the timers the clone armed - the 200 ms
+    // momentum-scroll timer (registered on touch devices), plus the double-click and long-press
+    // ones. Left pending, the momentum timer fires after `Core#destroy()` and calls
+    // `runHooks('afterMomentumScroll')` on a destroyed instance, which throws (issue #13120).
+    if (this.clone instanceof Clone) {
+      this.clone.wtEvent.destroy();
+    }
+
     this.clone?.eventManager.destroy(); // todo check if it is good place for that operation
   }
 }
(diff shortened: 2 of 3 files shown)


qunabu:
Task linked: [DEV-2582 Momentum-scroll timers on overlay clones are not cleared by destroy()](https://app.clickup.com/t/9015210959/DEV-2582)

demtario (approved):

Issue #12226, opened 2026-03-26
Values Missing From Column Filter After Condition Is Applied


GreenFlux:
### Describe the bug

Before a filter is applied to a column, all values show in the preview list at the bottom of the filter. But after applying a condition to the filter, the list no longer acts as a preview and the filter has to be saved and reopened to see the list update. 

[Related thread on forums
](https://forum.handsontable.com/t/filter-by-value-checkbox-inconsistent-behavior/9048)

**Before applying filter**
The checkbox serves as the means of turning the filter on/off for that value, and the value always shows in the list regardless of the checkbox state. 

<img width="455" height="411" alt="Image" src="https://github.com/user-attachments/assets/b5caffa7-8eb7-4d1a-806c-f3c28ab821bb" />

**After filter is saved**
But after applying `Filter by Condition > Contains > {value}`, clicking ok, and returning to the filter, values are missing from the list. Instead of being only unchecked- they are unchecked and hidden, making it impossible to recheck the values. 

<img width="380" height="462" alt="Image" src="https://github.com/user-attachments/assets/6f12171e-aa3e-426d-8a86-60fc3cc3f70f" />

This behavior is inconsistent and hard to reset the filter. Clearing the text and changing the Filter by condition to None does not update the preview list of values. 

<img width="385" height="412" alt="Image" src="https://github.com/user-attachments/assets/22559e5e-8f54-439d-a463-fffbde65911f" />

You have to click OK after resetting the Filter by condition (closing the filter) and then reopen it to see the values return. 

It seems like we treat the list of values as a preview of the filter in some cases, and other times it doesn't update the list until you save the filter. 

### Handsontable version

17.0.1

### Your environment

Chrome, MacOS

sl01k:
We'll investigate how to improve this. Thanks again for raising it!

budnix:
Thanks for the report. The issue has been fixed (https://github.com/handsontable/handsontable/pull/13212) and will be released in upcoming versions.

PR #13212, opened 2026-08-20
DEV-2584: Stop a column's own filter condition hiding its own values (#12226)

budnix:
### Context

The PR fixes the "filter by value" list of a column that is filtered by a condition, reported in #12226 (v17.0.1).

Apply `Filter by condition > Contains > <text>` on a column, click OK, then reopen that column's dropdown menu. The list at the bottom shows only the values that matched the condition. The other values are not merely unchecked — they are gone, so there is no way to check them back on. Clearing the condition in the menu does not bring them back either; the menu has to be confirmed and reopened first.

Applying a `by_value` filter on the same column behaves correctly: the list keeps every value and only unchecks the ones that were filtered out. The two paths disagreed.

**Root cause.** `ValueComponent` builds the list two different ways:

1. When the column already carries a `by_value` condition, `updateState()` builds the list through `ConditionUpdateObserver`'s `filteredRowsFactory`. That factory applies only the conditions of the columns **before** this one in the stack — the column's own conditions are excluded. Correct.
2. Otherwise `setState()` falls through to `reset()`, which read `getDataAtCol()` — the currently **visible** rows. The column's own condition had already trimmed them, so its own values disappeared.

**Fix.** `reset()` now gets its rows from a new `Filters._getValueListDataAtColumn()`, which picks the source per column:

- a column that sits in the condition stack gets the rows that survive the conditions of the columns **before** it — the same rule `filteredRowsFactory` already applies;
- a column outside the stack keeps reading `getDataAtCol()`, so its list stays narrowed down by the other columns' filters (the intended pivot-style behavior, unchanged);
- with a data provider the local path is skipped entirely, since filtering happens server-side and the list is hidden.

The list is built on menu open and never stored, so it cannot go stale and none of this work runs on the `afterChange` update cascade. `setState()` also skips rebuilding the list when a `by_value` snapshot is about to replace it, which removes a full column read per menu open.

`getSortComparatorForMeta()` now takes `unknown` and narrows with an `in` check, so the loosely typed data-map entries flow through without a cast.

**Note on the first revision of this PR** (superseded by `05c5fa1b29`): it stored an `itemsSnapshot` on the `CONDITION_NONE` state and had `setState()` restore it. That was wrong — `getState()` returns an `itemsSnapshot` on *every* OK click, `saveState()` writes it for the confirmed column, and nothing clears it. Confirming an unfiltered column's menu froze its list at that moment. `another column's list follows the filtered column even after it was confirmed once` is the regression test for it.

### How has this been tested?

New Playwright spec (6 cases) with its own fixture and page object:

```bash
cd tests && npx playwright test --project=e2e-main e2e/filters-value-list.spec.ts
```

Verified red-before-green by rebuilding against the pristine `develop` source: `keeps every value of a column that is filtered by a condition` and `a value the condition filters out can still be unchecked` both fail. `another column's list follows the filtered column even after it was confirmed once` fails against this PR's first revision.

Cases covered: the full value list survives a condition filter; confirming that list unchanged alters nothing; a value the condition filters out can be unchecked and stays out after the condition is dropped; an already visible value can still be unchecked; another column's list follows the filtered column both before and after its own menu has been confirmed.

Full suites, all green:

```bash
npm --prefix handsontable run test:e2e -- --testPathPattern=filters   # 283 specs, 0 failures
npm --prefix handsontable run test:unit                               # 3556 tests, 0 failures
```

Manual check in Chrome against a local build of the reported scenario.

Four legacy Jasmine cases encoded the old, narrowed list and were updated:

- `filtersUI.spec.js` — three cases clicked the multiple-select item at index 0 after reopening a condition-filtered column. The list now holds every value, so index 0 is a different item. They pick the item by its label instead, through a new `uncheckByValueItem()` test helper.
- `filtersUI.spec.js` — "should restore correct components' state after altering columns" now expects the condition-filtered column to list all 7 source values, all checked.
- `dropdownMenu.spec.js` — the ">2 conditions cannot be displayed" case now lists all 3 values. That column exceeds what the menu can render, so the plugin warns and the value component falls back to rebuilding its list.

### Types of changes

### Related issue(s):

1. #12226
2. DEV-2584

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2584

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 3eb17b7dc2d423f21e67ccecd68e9c4735e8d2fe. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2584 [GH-12226] Column's own filter condition hides its own values in the "filter by value" list](https://app.clickup.com/t/9015210959/DEV-2584)

adrianspdev (approved):

PR #13214, opened 2026-08-21
DEV-2578: Fix the partial-mode comments and close two packaging check gaps

demtario:
### Context

The PR fixes four review findings raised on #13210 (DEV-2578) after it was merged. None of them changes what CI does today; three close a gap between what the code claims and what it enforces, one closes a hole in the enforcement itself.

**1. The "exactly one caller" claim was false.** #13210's second commit moved `visual.yml` to `postbuild:partial`, making it a second partial caller, but two comments at the enforcement point were left saying only the ES + CJS build job may opt out — `prepare-package-for-publish.mjs`'s header and the `prepare-package-for-publish.partial` note in `tasks.json`. A reader trusts the comment next to the code over a test in another directory. Both now state the *rule* rather than a count — only a tree that never reaches a registry may skip the checks — and name `test/__tests__/previewPackaging.unit.js` as the place both call-site lists are pinned, so the next call site needs no third comment edit.

**2. A pattern-mode copy entry bypassed the completeness check.** For an object entry, `glob.sync(fileToCopy.pattern)` matching zero files pushed no destination, so the strict check on `COPY_DESTINATIONS` never saw the entry at all. Latent today — the copy list is all strings — but the script's guarantee would have quietly weakened the day someone added a pattern entry. Fixed on the destination side, deliberately: a source that is legitimately absent must keep passing (the preview job composes from a `tmp/` built in another job, and `handsontable/languages` is gitignored — that is exactly what #13210's second commit had to fix). So a zero-match pattern is translated into the pattern its files are addressed by inside `tmp/` (the same leading segments `pathSlice` drops), and only a tree holding none of them is reported as incomplete. The check confirms the *addressing*, never that the files are the ones the entry meant — the source is absent by definition — and a pattern whose sliced prefix is not literal is refused as unverifiable rather than checked against a pattern that means something else.

**3. The call-site tripwire scanned `.github/workflows/*.yml` only.** Composite actions under `.github/actions/**` run the same commands and were a blind spot. None calls `postbuild` today, so this is future-proofing; the scan now walks both trees recursively (nested composite actions are legal), keys on repo-relative paths (all five action files are named `action.yml`, so the file name alone would have made the assertion meaningless), and dedupes and sorts both lists instead of pushing once per matching line. What it cannot see is stated rather than glossed over: the lifecycle callers — `emitted-types.yml`, `docs-angular-typecheck.yml` and the `docs-angular-typecheck` action all run `npm run build --prefix handsontable` — never name `postbuild`, so no test guards them; they pass only because they build the whole package from source.

**4. No `AGENTS.md` entry for a new convention.** #13210 introduced two things a future agent trips on: strict-by-default `postbuild` with the `--partial` escape hatch, and the subtler one — the npm lifecycle auto-runs `postbuild` after any `npm run build` in `handsontable/`, so every new full-build call site inherits the strict gate without opting in, while invoking the task runner directly (`node scripts/run.mjs build`) never triggers it. Both are now a gotcha in the root `AGENTS.md`, next to the CI-orchestrator entry.

The finding also noted that #13210's own description repeats the false claim. That PR is merged and its body is left as it stands; this description is the correction.

CodeQL then flagged the first version of the pattern translation on this branch (alert 85, incomplete string escaping), which was a real defect and worse than the alert said: the `../` strip ran *before* the slice, while the copy step slices the matched path first, so `../types/**` with `pathSlice: 1` derived `**` instead of `types/**`. Both the order and the alert are fixed, with a test per leg.

### Addressed from the review on this PR

**1. The `../` drop had no red-able test.** The order test uses `pathSlice: 1`, where the slice consumes the `..` before the drop is reached, so the one branch that performs the documented drop ran in no test at all – removing `targetSegments.shift()` left all 16 cases green. The code was right; only its coverage was missing. A sibling of the order test at `pathSlice: 0`, one variable changed, now pins it.

**2. The refusal diagnostic misstated its reason for one input.** `toTargetPattern` returned a bare `null` for two different refusals, and the single call-site message named only the wildcard one. `{ pattern: 'types', pathSlice: 1 }` has a fully literal prefix and simply slices away to nothing, yet the script reported "its first 1 sliced segment(s) are not literal" – false for that input, and it sends a reader looking for a wildcard that is not in the pattern. The function now returns either the pattern or the reason it cannot produce one, the two refusals carry their own wording, and the sliced-away case asserts the new message *and* the absence of the old one. Refusing was already right in both cases: the copy step drops such a match into the root of `tmp/` unnamed, so there is no pattern its files are addressed by.

The review's two remaining notes are deliberately not acted on. The wildcard guard's breadth is a choice already pinned by a test, and the description overclaim is corrected below.

### How has this been tested?

**Every new check was shown red before it was green.** With the tests in place and the script reverted to develop's version, every case asserting a behavior develop does not have fails, and the rest pass:

```bash
git checkout origin/develop -- handsontable/scripts/prepare-package-for-publish.mjs
npm run test:unit --prefix handsontable -- --testPathPattern=previewPackaging
# ✕ should fail when a copy pattern is met by neither the checkout nor the composed tree
# ✕ should read an unmet copy pattern in the same order the copy step does
# ✕ should drop a leading `..` from a pattern nothing is sliced off
# ✕ should refuse to verify a copy pattern sliced through a wildcard
# ✕ should refuse a sliced prefix that is a single-segment wildcard too
# ✕ should refuse a copy pattern that slices away entirely
# ✕ should downgrade an unmet copy pattern to a warning in the partial mode
# Tests: 7 failed, 11 passed, 18 total
git checkout HEAD -- handsontable/scripts/prepare-package-for-publish.mjs
```

An earlier revision of this description put that run at `4 failed, 10 passed, 14 total` and named four cases. That was the count at the revision it was written against; it went stale as cases were added, and is corrected above.

**Ten new cases in `previewPackaging.unit.js`** (8 → 18), which make the destination-side derivation falsifiable rather than asserting it only where it cannot fail:

- a pattern met by neither the checkout nor `tmp/` → strict mode exits 1, naming the pattern and `postbuild:partial`;
- the same tree in `--partial` → exit 0 with a warning;
- **a pattern absent from the checkout whose files `tmp/` already carries → exit 0** — the preview-job case, and the one that breaks CI if the `pathSlice` translation is wrong;
- a pattern matched in the checkout with `pathSlice: 1` → the file lands at the sliced destination;
- `../types/**/*.d.ts` with `pathSlice: 1` → satisfied by `tmp/types/base.d.ts`, not by `tmp/base.d.ts`, which is the order the copy step reads it in (it slices the matched path first, and drops a leading `../` only afterwards);
- a pattern sliced through a wildcard (`src/**/types/*.d.ts`, `pathSlice: 2`) → refused as unverifiable, because `pathSlice` counts *path* segments while the pattern is sliced by *pattern* segments;
- `../types/**/*.d.ts` at `pathSlice: 0` → the same differential pair as the order test with one variable changed, and the only shape that reaches the `..` drop itself: satisfied by `tmp/types/base.d.ts`, not by `tmp/base.d.ts`;
- `types` with `pathSlice: 1` → refused for slicing away entirely, asserting the new wording *and* the absence of the wildcard wording, because both refusals exit 1 and only the reason tells them apart;
- a *single-segment* wildcard prefix (`src/*/types/*.d.ts`, `pathSlice: 2`) → refused as well, and pinned as such. Only `**` and a brace break the count outright (a brace expands before the pattern is read segment by segment, so it can hold a separator the split shreds, while `*`, `?` and a character class never match one). The guard is deliberately broader than those two: a wildcard prefix is not a known prefix, and the check refuses what it cannot decide rather than deciding as much as it could. The JSDoc now states that rule instead of the `**` case alone.

The guard-breadth case *is* red against develop, since develop has no guard at all – but develop-red proves only that a guard exists, not how broad it is. The narrowed-guard run below is the evidence that carries the breadth, which is why the case is here:

```bash
# guard narrowed to /\*\*|[[{]/ in toTargetPattern
npm run test:unit --prefix handsontable -- --testPathPattern=previewPackaging
# ✕ should refuse a sliced prefix that is a single-segment wildcard too
# Tests: 1 failed, 17 passed, 18 total
# guard restored
# Tests: 18 passed, 18 total
```

The two cases added for the review findings were each shown red against the exact defect they cover, one at a time:

```bash
# targetSegments.shift() removed from toTargetPattern
# ✕ should drop a leading `..` from a pattern nothing is sliced off  (Expected: 0, Received: 1)
# Tests: 1 failed, 17 passed, 18 total

# the sliced-away refusal collapsed back onto the wildcard wording
# ✕ should refuse a copy pattern that slices away entirely
#   Expected substring: "leaves nothing to address the files by"
# Tests: 1 failed, 17 passed, 18 total
```

The second is red on the message, not on the exit code – the script exits 1 either way, and the reason is the whole finding.

`runOnFixture` grew a `sourceFiles` argument and returns the composed file list, so the sliced-destination case asserts where the copy actually landed.

**Full runs:**

```bash
npm run test:unit --prefix handsontable        # 312 suites, 3742 tests, all green
npm run eslint --prefix handsontable           # 0 errors (601 pre-existing warnings, none in the changed files)
```

No full build was run: the pattern branch is reachable only from an object entry in `handsontable.copy`, and every entry there is a string, so composition of the real package is byte-for-byte what #13210 shipped.

[skip changelog] — tooling, test, and agent-documentation only; no `handsontable/src/**` or `wrappers/**` change.

### Types of changes

### Related issue(s):

1. DEV-2578
2. Follow-up to #13210

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

---

> [!NOTE]
> **Low Risk**
> Changes are limited to publish scripting, CI wiring tests, and documentation; no runtime library or wrapper code paths.
> 
> **Overview**
> Follow-up to **#13210** that tightens publish-package completeness checks and fixes misleading comments—**CI behavior for existing workflows stays the same**.
> 
> **`prepare-package-for-publish.mjs`** now validates **pattern-mode** `handsontable.copy` entries when the source glob matches nothing (a case that previously skipped `COPY_DESTINATIONS` entirely). It maps each pattern to where files should live in `tmp/` via new **`toTargetPattern`**, aligned with the copy step (slice matched path first, then drop one leading `../`). Optional **`pathSlice`** is normalized with **`?? 0`**. Patterns with a non-literal sliced prefix or that slice to nothing are **refused** as unverifiable rather than checked against the wrong glob. Trees that already have the expected files in `tmp/` (preview jobs pulling artifacts) still pass.
> 
> Comments and **`tasks.json`** now state the rule: only trees that **never reach a registry** may use **`postbuild:partial`** (today **`build.yml`** and **`visual.yml`**), not “exactly one caller.” **`AGENTS.md`** documents strict-by-default **`postbuild`** after **`npm run build`** vs **`run.mjs`**, and the partial escape hatch.
> 
> **`previewPackaging.unit.js`** scans **workflows and composite actions** for **`postbuild`** / **`postbuild:partial`** call sites, and adds unit tests for the new pattern translation and failure/warning paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 91b3ee560334e4c66b3730a0c770995c6db3adde. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2578 Core: pkg.pr.new preview packages ship a stale exports map — no theme/icon CSS subpaths, no dist/themes](https://app.clickup.com/t/9015210959/DEV-2578)

danielzytohoc:
Reviewed against the four findings from the #13210 review — **no blocking issues found**. Nice follow-up, and the red-before-green demonstration in the description (reverting the script to develop and showing exactly the four new pattern cases fail) is the standard we want for enforcement code.

What I verified beyond reading the diff:

- **The pattern translation matches the copy step's real semantics.** `toTargetPattern` slices *pattern* segments first and drops a single leading `..` afterwards — mirroring the copy step's `split(sep).slice(pathSlice)` followed by one `'../'` replace on the joined string. Traced `../types/**` with `pathSlice: 1` through both paths: both land at `tmp/types/`. The CodeQL-flagged ordering defect is genuinely fixed — [alert 85](https://github.com/handsontable/handsontable/security/code-scanning/85) is in state *fixed*, and the two-sided test (`should read an unmet copy pattern in the same order the copy step does`) pins the order in both directions.
- **`pathSlice ?? 0` changes nothing in the copy step** (`slice(undefined)` already behaved as `slice(0)`) while making the wildcard guard read the same entry consistently; the inline comment about the two `slice` readings is accurate.
- **The AGENTS.md entry's claims hold in the tree**: `emitted-types.yml`, `docs-angular-typecheck.yml`, and the `docs-angular-typecheck` composite action all run `npm run build --prefix handsontable` and are correctly described as lifecycle callers the tripwire cannot see.
- The tripwire rewrite is right: recursive scan of `.github/actions/**` (nested composite actions are legal), `ya?ml` on both trees, repo-relative keys, deduped sets. Workflows staying non-recursive is correct — GitHub only reads top-level `.github/workflows`.

Two low notes, neither blocking:

1. **The wildcard guard refuses more than its rationale requires.** The JSDoc justifies refusal by `**` matching a variable number of path segments, but `/[*?[{]/` also refuses single-segment wildcards (`?`, `[abc]`, `{a,b}`), which *do* preserve segment-count alignment. Consequence: an entry with such a prefix, an absent source, and a complete tree would fail strict mode as "unverifiable" despite being checkable. Conservative, loud, and consistent with the stated "unverifiable rather than wrongly verified" policy — fine as a deliberate choice, just noting it is broader than the `**` case the comment names.
2. **Two stale bot threads are worth closing so the next reviewer does not re-litigate them.** The cursor "Undefined pathSlice" comment re-anchored onto the head commit, but the head contains `pathSlice ?? 0` plus the covering test (`should accept a copy pattern that declares no pathSlice`) — the finding predates the fix. A short "fixed in `3cee434`" reply on that thread and the outdated CodeQL one would do it.

demtario:
Thanks for the trace, both notes handled in `7d98161`.

**Note 1.** You are right that the guard refuses more than the JSDoc's rationale justifies, and I kept the guard while fixing the comment, which is where the mismatch actually was. One correction to the note: `{a,b}` belongs on the *other* side of the line. glob brace-expands before the pattern is read segment by segment, so a brace can hold a separator that the `split(/[\\/]/)` above shreds:

```
glob.sync('{utils/console.mjs,run.mjs}')  // ['utils/console.mjs'] - spans a separator
glob.sync('utils[/]console.mjs')          // []                    - a class cannot
glob.sync('*/console.mjs')                // []                    - nor can `*`
```

So `**` and `{` break the segment count outright, while `*`, `?` and `[abc]` keep it aligned. The JSDoc now names those two, then states that the guard is deliberately broader than them: a wildcard prefix is not a known prefix, and this check refuses what it cannot decide rather than deciding as much as it could. Added `should refuse a sliced prefix that is a single-segment wildcard too` (`src/*/types/*.d.ts`, `pathSlice: 2`) so the breadth is pinned and reads as a choice rather than as an oversight to tidy up. It cannot be shown red against develop, since develop has no guard at all, so it was shown red against a guard narrowed to `/\*\*|[[{]/` (1 failed, 15 passed) and green with the guard restored (16 passed). PR description updated with both.

**Note 2.** Both bot threads are already resolved, and the CodeQL one is marked outdated too, so I left them as they stand rather than adding a reply that says what the thread state already says. Shout if you would rather have the explicit "fixed in `3cee434`" line on the cursor thread.

handsontable/scripts/prepare-package-for-publish.mjs
```diff
@@ -72,6 +74,44 @@ glob.sync('./**/*.d.ts', { cwd: TARGET_PATH, nodir: true }).forEach((dtsFile) =>
...
+    return null;
+  }
+
+  const targetSegments = segments.slice(pathSlice);
+
+  // The copy step builds its destination the same way round: it slices the matched path first and
+  // drops a single leading `../` only afterwards, that being the only position where it means
+  // anything.
+  if (targetSegments[0] === '..') {
+    targetSegments.shift();
```
danielzytohoc, line 117: This branch never executes in any test: the order test uses `pathSlice: 1`, where the slice consumes the `..`, so the drop itself runs only for a `../…` pattern with `pathSlice` 0/omitted. The code is right — it mirrors the copy step's `replace('../', '')` — but the one path that performs the documented drop has no red-able test. The same differential-pair shape as the order test, at `pathSlice: 0`, would cover it.

handsontable/scripts/prepare-package-for-publish.mjs
```diff
@@ -72,6 +74,44 @@ glob.sync('./**/*.d.ts', { cwd: TARGET_PATH, nodir: true }).forEach((dtsFile) =>
...
+  }
+
+  return targetSegments.join('/') || null;
```
danielzytohoc: This `|| null` leg is untested, and the diagnostic it routes to misstates the reason. I ran the script on `{ pattern: 'types', pathSlice: 1 }` — the prefix is fully literal, the pattern just slices away entirely — and it refuses with exit 1 (fail-safe, good) but reports "its first 1 sliced segment(s) are not literal", which is false for that input. Worth a message of its own ("the pattern slices away entirely") plus a case pinning it.

danielzytohoc (commented):
I focused on whether the new tests meet the test-writing bar (falsifiable, intent-encoded, no hollow assertions). Verdict: they do — verified by running them, not just reading them.

**Verification** (throwaway worktree at 540bf361e):

- All 16 `previewPackaging` cases green at the head.
- With only `prepare-package-for-publish.mjs` reverted to `origin/develop`, 5 of the 8 new cases go red — the four named in the description plus the guard-breadth one — each failing on its behavioral assertion (`Expected: 1, Received: 0`), so red for the right reason.

**What holds up well:**

- Every case states the requirement and asserts observable behavior — exit code, the diagnostic naming the offending pattern, where files land. No hollow assertions, nothing mocked: the suite spawns the real script on real temp trees.
- `should read an unmet copy pattern in the same order the copy step does` is the strongest one: the differential pair (file at `tmp/` root → 1, at `tmp/types/` → 0) pins the slice-then-strip order — exactly the CodeQL bug class.
- `should refuse a sliced prefix that is a single-segment wildcard too` pins the guard's breadth as policy, so a future tidy-up can't silently narrow it.
- The must-pass direction is covered too (pattern satisfied by `tmp/`, omitted `pathSlice`), which is what keeps the new check from breaking the preview job.

Two small coverage gaps inline — the only code notes I have.

One description nit: "the eighth case cannot be shown red against develop" is literally false — I ran it against develop and it is red there (expects 1, gets 0), since any guard existing makes it red. The narrowed-guard run is the stronger evidence — develop-red proves a guard exists, not its breadth — so the practice was right; the sentence just overclaims.

Neither gap blocks the merge for me — I'd take both as a small follow-up commit on this branch.

handsontable/scripts/prepare-package-for-publish.mjs
```diff
@@ -85,6 +119,28 @@ FILES_TO_COPY.forEach((fileToCopy) => {
...
     pathSlice = fileToCopy.pathSlice;
+
+    if (foundFiles.length === 0) {
+      // No match means no destination is recorded, so the check below cannot see this entry at
+      // all. Ask the destination side instead: the tree may already carry the files from an
+      // artifact built elsewhere (the preview job extracts a `tmp/` composed in another job),
+      // which is legitimate; holding nothing addressed the way this entry addresses its files
+      // is an incomplete package. The source is absent by definition here, so this confirms the
+      // addressing, never that the files are the ones the entry meant.
+      const targetPattern = toTargetPattern(fileToCopy.pattern, pathSlice);
```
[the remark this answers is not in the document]
demtario: Fixed in `3cee434`, before this comment re-anchored onto the head commit. The head reads `pathSlice = fileToCopy.pathSlice ?? 0`, so the copy step's `slice(pathSlice)` and the guard's `slice(0, pathSlice)` read the same entry the same way, and an omitted `pathSlice` no longer makes a wildcard look unverifiable. Covered by `should accept a copy pattern that declares no pathSlice` in `previewPackaging.unit.js`.

handsontable/scripts/prepare-package-for-publish.mjs
```diff
@@ -72,6 +74,23 @@
...
+ * Translate a copy pattern into the pattern its files are addressed by inside the composed tree,
+ * by dropping the same leading segments `pathSlice` drops off each matched path.
+ *
+ * @param {string} pattern The `handsontable.copy` pattern, e.g. a declaration glob under `types/`.
+ * @param {number} pathSlice How many leading path segments the copy step slices off.
+ * @returns {string}
+ */
+function toTargetPattern(pattern, pathSlice) {
+  const segments = pattern
+    .replace('../', '') // the copy step strips it the same way when it builds the destination
```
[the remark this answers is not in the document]
demtario: Fixed — [alert 85](https://github.com/handsontable/handsontable/security/code-scanning/85) is in state *fixed*. `toTargetPattern` no longer does a string replace at all: it splits the pattern, slices the `pathSlice` leading segments, and drops a single leading `..` *segment* afterwards, which is the order the copy step reads it in. `should read an unmet copy pattern in the same order the copy step does` pins that order in both directions.

handsontable/scripts/prepare-package-for-publish.mjs
demtario: Confirmed by running it, and fixed in `91b3ee5603`.

With `targetSegments.shift()` deleted the whole suite stayed green at 16/16, so the branch was reached by nothing. Added `should drop a leading '..' from a pattern nothing is sliced off` as the sibling you describe: the same differential pair as the order test with one variable changed, at an explicit `pathSlice: 0` rather than an omitted one, since omission is already covered by `should accept a copy pattern that declares no pathSlice` and the explicit 0 makes the pair read as the order test's twin.

```
../types/**/*.d.ts, pathSlice: 0
  tmp/base.d.ts        -> 1
  tmp/types/base.d.ts  -> 0
```

Red-checked against the defect itself: with `shift()` removed, that case is the only failure, and it fails on the `inPlace` leg (`Expected: 0, Received: 1`) - the drop is what puts the files under `tmp/types/`.

handsontable/scripts/prepare-package-for-publish.mjs
demtario: Reproduced exactly as you describe, and fixed in `91b3ee5603`. Running the script on `{ pattern: 'types', pathSlice: 1 }` gave:

```
ERROR: The composed package is incomplete:
- The copy pattern matches nothing and cannot be checked against the package, because its first 1 sliced segment(s) are not literal: types
```

The prefix is fully literal, so that reason is false for this input and sends a reader looking for a wildcard that is not in the pattern.

`toTargetPattern` now returns either `{ pattern }` or `{ refusal }` instead of a bare `null`, so the two refusals carry their own wording and the call site reports whichever it got:

- non-literal sliced prefix -> `its first N sliced segment(s) are not literal` (unchanged - two existing cases pin that substring)
- slices away to nothing -> `slicing its first N segment(s) leaves nothing to address the files by`

Refusing was already right for both: at `pathSlice: 1` the copy step drops such a match into the root of `tmp/` unnamed, so there is no pattern its files are addressed by either. The JSDoc says so now, in place of the old "returned as `null`" contract.

`should refuse a copy pattern that slices away entirely` pins it, and asserts the absence of the old wording as well as the presence of the new one. Red-checked against the collapsed message rather than against develop, deliberately - the script exits 1 either way, so the exit code proves nothing here and the reason is the whole finding:

```
# the sliced-away refusal collapsed back onto the wildcard wording
# x should refuse a copy pattern that slices away entirely
#   Expected substring: "leaves nothing to address the files by"
# Tests: 1 failed, 17 passed, 18 total
```

One unrelated tidy-up while in the JSDoc: the first draft of `@returns` carried `?: undefined` members to help a structural checker discriminate the union, but `handsontable/tsconfig.json` only includes `src/**/*`, so `scripts/` is never type-checked and they were noise. Dropped.

danielzytohoc (approved):
Re-verified `91b3ee5603` the same way as the first pass — by running it in a throwaway worktree, not by reading it:

- 18/18 green at the head.
- With `targetSegments.shift()` deleted, exactly one case goes red — `should drop a leading '..' from a pattern nothing is sliced off`, on the `inPlace` half (`Expected: 0, Received: 1`): without the drop the derived pattern escapes `tmp/` and a complete tree is wrongly refused. The pair isolates exactly the defect it was written for.
- With the script at the previous head (`540bf361e`), exactly one case goes red — `should refuse a copy pattern that slices away entirely`, on the message discrimination (`not.toContain('not literal')`). "Red on the message, since the script exits 1 either way and the reason is the whole finding" is the right subtlety.

Both threads resolved, and the description rewrite lands the develop-red vs narrowed-guard distinction accurately.

One take-it-or-leave-it aside, non-blocking and pre-existing before `91b3ee5603`: the two `../types/**` fixtures (the order test and the new drop test) glob one level above the `mkdtempSync` sandbox, into the OS tmpdir — a stray `$TMPDIR/types/*.d.ts` tree would flip their `rootOnly` legs. If this file gets touched again, nesting the spawn cwd one level inside the temp dir keeps `../` sandboxed. Not worth a round on this PR.

PR #13215, opened 2026-08-21
DEV-2585: Fix memory leak where Formulas added a sheet on every data load

### LINKED ISSUE #11165: Handsontable formula calculates that there is a memory leak
### Describe the bug

When I execute loadData or updateData multiple times, each execution results in a decrease in the performance of the handsontable

### Video/Screenshots

image
image

https://stackblitz.com/edit/xzaewz?file=index.js

### Handsontable version

14.5

### Framework version

_No response_

### Your environment

chrome ,macos

budnix:
### Context

The PR fixes a memory leak in the Formulas plugin: every `loadData()` / `updateData()` call asked HyperFormula for a brand-new sheet instead of reusing the one the plugin already owns.

`setupSheet()` in `handsontable/src/plugins/formulas/engine/register.ts` only reused a sheet when the caller passed a `sheetName`. `#onAfterLoadData` passed the `formulas.sheetName` setting straight through, and that setting is optional — so for anyone who did not set it (the common case) every data load took the `addSheet()` branch. The plugin then repointed itself at the new sheet and walked away from the old one, which stayed in the engine with all of its cells and its whole dependency graph. Nothing ever removed it.

The reported slowdown is the same bug seen from the other side. The `sheetAdded` listener registered in `registerEngine()` runs `rebuildAndRecalculate()` across **every** sheet the engine holds, so call *n* paid for *n* sheets — per-call cost grew linearly, and a whole session grew quadratically.

Measured before the fix, on 100 `loadData()` calls over 100 rows x 5 columns with one formula column. Released 14.5.0 and `develop` behaved identically:

| Measured after | 1 call | 100 calls |
|---|---|---|
| Sheets left in the engine | 2 | **101** |
| Rows retained across all sheets | 200 | **10,100** |
| `ScalarFormulaVertex` instances | 200 | **10,100** |
| `ValueCellVertex` instances | 800 | **40,400** |
| Reachable heap after a forced GC | 25.8 MB | **35.9 MB** |

Within that same 100-call run, per-call time rose from 26.5 ms (average of the first five calls) to 86.5 ms (average of the last five) — 3.3x, growing in step with the sheet count.

Two controls pinned it to this code path: with the plugin removed, time stayed flat and the heap stayed bounded; with `sheetName` set, the sheet count stayed at 1 and time stayed flat.

`setupSheet()` now reuses the sheet the plugin already owns and adds one only when there is nothing to reuse. Grids that share a single engine still get one sheet each, because the fallback is the calling plugin instance's own `sheetName`.

No public API changes and no default setting changes. Three observable differences, all of them the point of the fix:

- `formulas.sheetName` / `getSheetNames()` no longer grow an incrementing list of abandoned sheets across data loads.
- With `formulas.sheetName` unset, a `switchSheet('X')` followed by a user `loadData()` now writes into `X`. Before, the load abandoned `X` and created a new sheet.
- `plugin.sheetName` now reports the engine's own casing. The engine matches sheet names without regard to case but keeps the casing it was given, so a configured `sheetName: 'sheet1'` against an engine sheet named `Sheet1` used to leave the plugin holding `'sheet1'`.

Not addressed here, and pre-existing: a configured `sheetName` still wins over an earlier `switchSheet()` on the next data load, which silently un-switches it.

#### Second commit: guarding the rename handler

Reusing the sheet raised the stakes on a separate, pre-existing defect, so this PR closes that too.

HyperFormula's `sheetRenamed` event is engine-wide, and every plugin instance attached to a shared engine subscribes to it. `#onEngineSheetRenamed` handled the event unconditionally, repointing its own `sheetName` / `sheetId` at the renamed sheet even when it did not own that sheet. Before this PR the mis-pointer was survivable on a data load, because `loadData` created a fresh sheet anyway. With the sheet now reused, the same mis-pointer made one grid's `loadData` overwrite another grid's sheet — real data loss.

`#onEngineSheetRenamed` now updates the pointer only when the renamed sheet is the one the instance owns. The `afterSheetRenamed` hook still runs on every instance, so that observable behavior is unchanged.

#### Third commit: keeping `sheetName` in step with the engine

`#updateSheetNameAndSheetId` now stores the name the engine reports rather than the one it was handed, so every exact-string reader of `sheetName` is correct by construction. Doing it there rather than in `setupSheet` covers `switchSheet()` as well.

That change on its own would have regressed `updatePlugin`, which compared the configured `sheetName` to `this.sheetName` as exact strings: re-passing the same name in a different case counted as a change of sheet and fired a real `switchSheet`, reloading the data from the engine on every `updateSettings`. It now compares sheet ids. There is a test for it.

#### Fourth commit: redrawing dependent grids on a failed load

The branch that empties a reused sheet discarded the `setSheetContent` return value and never called `renderDependentSheets`, unlike the path that writes the new data. A grid reading that sheet through a cross-sheet formula kept showing the old value until something else triggered a render. Found by Cursor Bugbot.

### How has this been tested?

Seven cases added to the existing `memoryLeak.spec.js`, covering sheet reuse across `loadData`, sheet reuse across `updateData`, no stale data left anywhere in the engine after a reload, the configured-`sheetName` path still being honoured, two grids sharing one engine keeping separate sheets, one grid's data surviving a rename of its sheet followed by the other grid reloading, and a dependent grid being redrawn when the sheet it reads is emptied by a failed load.

Each case was written before its fix and confirmed to fail for the right reason: 6 sheets instead of 1 and 4 instead of 2 for the leak, and for the rename case, the second grid's values landing in the first grid's sheet (`'30'` where `'1'` belonged, with the first grid then reading `70` instead of `3`).

Two more cases in `hfApi.spec.js` for the third commit: `sheetName` reporting the engine's casing, and `updateSettings` not switching sheets when the configured name differs from the stored one only in case. The second was confirmed failing (one switch instead of none) before the `updatePlugin` id comparison went in.

The two pre-existing rename tests still pass unchanged — `hooks.spec.js` renames a sheet the grid does not own and asserts the hook still fires, and `formulas.spec.js` renames the grid's own sheet and asserts the pointer follows it.

```bash
# The added cases
npm run test:e2e --prefix handsontable -- --testPathPattern=memoryLeak

# The whole Formulas suite -- 343 specs, 0 failures

# Full unit suite -- 3596 tests, 0 failures

# Full E2E suite -- 9638 specs, 0 failures

# Lint
```

`npm run build --prefix handsontable` also passes, including `build:types`. The emitted declaration for the widened parameter is `setupSheet(engineInstance: HyperFormulaEngine, sheetName?: string | null): string`, so existing callers passing a `string` are unaffected.

Also verified in a real browser against a UMD build of this branch, driving the reporter's own reproduction (100 `loadData()` calls). Same harness and same machine as the "before" numbers above:

| After 100 `loadData()` calls | `develop` | This branch |
|---|---|---|
| Sheets in the engine | 101 | **1** |
| Rows retained across all sheets | 10,100 | **100** |
| `ScalarFormulaVertex` instances | 10,100 | **100** |
| `ValueCellVertex` instances | 40,400 | **400** |
| Reachable heap after a forced GC | 35.9 MB | **28.0 MB** |
| Per-call time, average of the last 5 | 86.5 ms | **6.8 ms** |

Retention is now flat in the number of data loads, and the 100th `loadData()` is roughly 13x faster. The engine ends up holding fewer vertices than `develop` did after a *single* load (200 / 800), because the old code abandoned the initial sheet too.

### Types of changes

### Related issue(s):

1. DEV-2585
2. Fixes #11165

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

No documentation change is needed: no public API or option changed, and no guide documented the old sheet-per-load behavior.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2585

### DIFF
--- a/handsontable/src/plugins/formulas/__tests__/memoryLeak.spec.js
+++ b/handsontable/src/plugins/formulas/__tests__/memoryLeak.spec.js
@@ -45,6 +45,107 @@ describe('Formulas memory leak check', () => {
     }
   });
 
+  it('should reuse the already-owned sheet on every `loadData` call (no `sheetName` configured)', async() => {
+    const hot = handsontable({
+      data: [['1', '2', '=A1+B1']],
+      formulas: {
+        engine: HyperFormula,
+      },
+    });
+    const { engine } = hot.getPlugin('formulas');
+
+    expect(engine.getSheetNames().length).toBe(1);
+
+    for (let i = 0; i < 5; i++) {
+      hot.loadData([['1', '2', '=A1+B1']]);
+    }
+
+    expect(engine.getSheetNames().length).toBe(1);
+    expect(engine.getSheetId(hot.getPlugin('formulas').sheetName)).toBe(hot.getPlugin('formulas').sheetId);
+    expect(hot.getDataAtCell(0, 2)).toBe(3);
+  });
+
+  it('should reuse the already-owned sheet on every `updateData` call (no `sheetName` configured)', async() => {
+    const hot = handsontable({
+      data: [['1', '2', '=A1+B1']],
+      formulas: {
+        engine: HyperFormula,
+      },
+    });
+    const { engine } = hot.getPlugin('formulas');
+
+    expect(engine.getSheetNames().length).toBe(1);
+
+    for (let i = 0; i < 5; i++) {
+      hot.updateData([['1', '2', '=A1+B1']]);
+    }
+
+    expect(engine.getSheetNames().length).toBe(1);
+    expect(hot.getDataAtCell(0, 2)).toBe(3);
+  });
+
+  it('should not retain the previous data in the engine after reloading the data', async() => {
+    const hot = handsontable({
+      data: [['1', '2', '=A1+B1']],
+      formulas: {
+        engine: HyperFormula,
+      },
+    });
+    const { engine } = hot.getPlugin('formulas');
+
+    hot.loadData([['10', '20', '=A1+B1']]);
+
+    const allSheetsSerialized = engine.getSheetNames()
+      .map(name => engine.getSheetSerialized(engine.getSheetId(name)));
+
+    // Only the current data may be present anywhere in the engine.
+    expect(allSheetsSerialized).toEqual([[['10', '20', '=A1+B1']]]);
+  });
+
+  it('should keep using the sheet pointed to by the `sheetName` option across data loads', async() => {
+    const hot = handsontable({
+      data: [['1', '2', '=A1+B1']],
+      formulas: {
+        engine: HyperFormula,
+        sheetName: 'MySheet',
+      },
+    });
+    const { engine } = hot.getPlugin('formulas');
+
+    hot.loadData([['1', '2', '=A1+B1']]);
+    hot.updateData([['1', '2', '=A1+B1']]);
+
+    expect(engine.getSheetNames()).toEqual(['MySheet']);
+    expect(hot.getPlugin('formulas').sheetName).toBe('MySheet');
+  });
+
+  it('should give each grid its own sheet when they share a single engine', async() => {
+    const hfInstance = HyperFormula.buildEmpty({ licenseKey: 'internal-use-in-handsontable' });
+    const hot1 = handsontable({
+      data: [['1', '2', '=A1+B1']],
+      formulas: {
+        engine: hfInstance,
+      },
+    });
+    const hot2 = spec().$container2.handsontable({
+      data: [['3', '4', '=A1+B1']],
+      formulas: {
+        engine: hfInstance,
+      },
+    }).data('handsontable');
+
+    expect(hfInstance.getSheetNames().length).toBe(2);
+
+    hot1.loadData([['1', '2', '=A1+B1']]);
+    hot2.loadData([['3', '4', '=A1+B1']]);
+
+    // Reloading the data must not add sheets, and must not make the two grids share one sheet.
+    expect(hfInstance.getSheetNames().length).toBe(2);
+    expect(hot1.getPlugin('formulas').sheetId).not.toBe(hot2.getPlugin('formulas').sheetId);
+    expect(hot1.getDataAtCell(0, 2)).toBe(3);
+    expect(hot2.getDataAtCell(0, 2)).toBe(7);
+  });
+
   it('should detach listeners from the engine after table destroying (one shared HF instances)', async() => {
     const hfInstance1 = HyperFormula.buildEmpty({ licenseKey: 'internal-use-in-handsontable' });
     const hot1 = handsontable({
--- a/handsontable/src/plugins/formulas/__tests__/memoryLeak.spec.js
+++ b/handsontable/src/plugins/formulas/__tests__/memoryLeak.spec.js
@@ -57,7 +57,7 @@ describe('Formulas memory leak check', () => {
     expect(engine.getSheetNames().length).toBe(1);
 
     for (let i = 0; i < 5; i++) {
-      hot.loadData([['1', '2', '=A1+B1']]);
+      await loadData([['1', '2', '=A1+B1']]);
     }
 
     expect(engine.getSheetNames().length).toBe(1);
@@ -77,7 +77,7 @@ describe('Formulas memory leak check', () => {
     expect(engine.getSheetNames().length).toBe(1);
 
     for (let i = 0; i < 5; i++) {
-      hot.updateData([['1', '2', '=A1+B1']]);
+      await updateData([['1', '2', '=A1+B1']]);
     }
 
     expect(engine.getSheetNames().length).toBe(1);
@@ -93,7 +93,7 @@ describe('Formulas memory leak check', () => {
     });
     const { engine } = hot.getPlugin('formulas');
 
-    hot.loadData([['10', '20', '=A1+B1']]);
+    await loadData([['10', '20', '=A1+B1']]);
 
     const allSheetsSerialized = engine.getSheetNames()
       .map(name => engine.getSheetSerialized(engine.getSheetId(name)));
@@ -112,8 +112,8 @@ describe('Formulas memory leak check', () => {
     });
     const { engine } = hot.getPlugin('formulas');
 
-    hot.loadData([['1', '2', '=A1+B1']]);
-    hot.updateData([['1', '2', '=A1+B1']]);
+    await loadData([['1', '2', '=A1+B1']]);
+    await updateData([['1', '2', '=A1+B1']]);
 
     expect(engine.getSheetNames()).toEqual(['MySheet']);
     expect(hot.getPlugin('formulas').sheetName).toBe('MySheet');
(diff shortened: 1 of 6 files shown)


qunabu:
Task linked: [DEV-2585 Fix memory leak: Formulas plugin adds a new HyperFormula sheet on every loadData/updateData](https://app.clickup.com/t/9015210959/DEV-2585)

handsontable/src/plugins/formulas/formulas.ts
```diff
@@ -354,7 +354,12 @@ export class Formulas extends BasePlugin {
    * @param {string} newDisplayName The new name of the sheet.
    */
   #onEngineSheetRenamed = (oldDisplayName: string, newDisplayName: string) => {
-    this.#updateSheetNameAndSheetId(newDisplayName);
+    // The event is engine-wide, so it also reaches instances that do not own the renamed sheet.
+    // Repointing those would make them operate on a sheet belonging to another instance.
+    if (oldDisplayName === this.sheetName) {
```
Haxikowy: Guard compares display names, but HF resolves sheet names case-insensitively while keeping the casing it was given. With `formulas: { sheetName: 'sheet1' }` against an engine sheet named `Sheet1`, `setupSheet` returns `'sheet1'`, so `this.sheetName` is `'sheet1'` and the event carries `oldDisplayName: 'Sheet1'`. Guard misses, the instance keeps a name that no longer exists, and `#onModifyData` then bails on `!doesSheetExist(this.sheetName)` - formulas silently stop resolving. develop repointed it.

The rename is already applied when the event fires, so id comparison is equivalent and case-proof:

```ts
if (this.engine?.getSheetId(newDisplayName) === this.sheetId) {
```

Worth a case-mismatch case in the shared-engine rename test.

handsontable/src/plugins/formulas/formulas.ts
```diff
@@ -1320,7 +1325,10 @@ export class Formulas extends BasePlugin {
 
     const formulasSettings = this.hot.getSettings()[PLUGIN_KEY];
     const settingsSheetName = isFormulasSettingsObject(formulasSettings) ? formulasSettings.sheetName : undefined;
-    const sheetName = setupSheet(this.engine, settingsSheetName!);
+    // Fall back to the sheet this instance already owns. Without it every `loadData`/`updateData`
+    // call adds a sheet and abandons the previous one - with its whole dependency graph - inside
+    // the engine, which the engine then recalculates on every subsequent call.
+    const sheetName = setupSheet(this.engine, settingsSheetName ?? this.sheetName);
```
Haxikowy, line 1342: Follow-up on the reuse, for the `isItPossibleToReplaceSheetContent(...) === false` branch below: when the replace is rejected nothing writes, and the reused sheet now keeps the *previous* load's content while the grid shows the new data. Before this change that sheet was brand new, so the engine was merely empty. Narrow path, but the divergence is silent - an explicit `setSheetContent(this.sheetId, [[]])` or at least a log in the else would keep it deterministic.

handsontable/src/plugins/formulas/__tests__/memoryLeak.spec.js
```diff
@@ -45,6 +45,138 @@ describe('Formulas memory leak check', () => {
...
+      formulas: {
+        engine: HyperFormula,
+      },
+    });
+    const { engine } = hot.getPlugin('formulas');
+
+    expect(engine.getSheetNames().length).toBe(1);
+
+    for (let i = 0; i < 5; i++) {
+      hot.loadData([['1', '2', '=A1+B1']]);
```
Haxikowy: `hot.loadData()` raw instead of the suite's `await loadData([...])` helper. The engine assertions are synchronous so these pass, but the render flush the helper awaits is skipped, which makes the `getDataAtCell` assertion depend on the sync path staying sync. Same in the `updateData` test and the shared-engine ones.

demtario (commented):
Reviewed at `36b6abc`. Three inline notes below, and two more points in a separate comment because they have no line in the diff to hang off.

handsontable/src/plugins/formulas/engine/register.ts
```diff
@@ -280,15 +280,20 @@ export function registerNamedExpressions(
...
+export function setupSheet(engineInstance: HyperFormulaEngine, sheetName?: string | null) {
+  if (sheetName === undefined || sheetName === null) {
+    return engineInstance.addSheet();
+  }
+
+  if (!engineInstance.doesSheetExist(sheetName)) {
+    return engineInstance.addSheet(sheetName);
   }
 
   return sheetName;
```
demtario, line 299: The reuse branch returns the string the caller passed. HyperFormula resolves sheet names case-insensitively but keeps the casing it was given, so `plugin.sheetName` can hold a casing the engine never uses. Any exact-string consumer of `sheetName` then misbehaves — including the new guard in `#onEngineSheetRenamed`, which compares against `oldDisplayName`. `switchSheet()` has the same shape (`formulas.ts:765`).

```js
return engineInstance.getSheetName(engineInstance.getSheetId(sheetName));
```

keeps `plugin.sheetName` in lockstep with the engine, so a future exact-string consumer is safe by construction rather than by each call site remembering to canonicalize.

handsontable/src/plugins/formulas/engine/register.ts
```diff
@@ -280,15 +280,20 @@ export function registerNamedExpressions(
...
 
 /**
- * Sets up a new sheet.
+ * Sets up the sheet to work on. An existing sheet is reused; a new one is added only when there is
+ * nothing to reuse.
  *
  * @param {object} engineInstance The engine instance.
- * @param {string} sheetName The new sheet name.
+ * @param {string} [sheetName] The sheet name to use. When omitted (or `null`), a new sheet is added.
  * @returns {*}
```
demtario: The block is already being edited here, and the return is now unambiguously `{string}`.

handsontable/src/plugins/formulas/formulas.ts
demtario, line 1342: Two notes on the fallback, neither blocking.

With `formulas.sheetName` unset, `switchSheet('X')` followed by a user `loadData()` now writes into `X`; before, the load abandoned `X` and created a new sheet. The new behavior is the coherent one — worth a line in the PR body as an observable change.

The configured case stays asymmetric, and this line is where it shows: a configured `sheetName` still wins over a prior `switchSheet()` on the next load and silently un-switches it (the `hfApi.spec.js:51` scenario). Pre-existing, out of scope here.

demtario:
Two points that have no line in the diff to hang off.

**The impact statement is incomplete.** The body says the observable difference is that `sheetName` / `getSheetNames()` stop growing. Two more follow from no longer calling `addSheet()`:

- **`afterSheetAdded` stops firing on data loads.** Every `loadData`/`updateData` used to call `engine.addSheet()`, which emits `sheetAdded` engine-wide, so `#onEngineSheetAdded` ran the `afterSheetAdded` hook on *every* attached instance on *every* data load. It now fires at init and on real additions only. That is what the hook's own contract already documents (`core/hooks/constants.ts:3050-3065` — "fires once during initialization, before any user-triggered sheet additions"), so the code now matches the docs and nothing needs updating. It still belongs in the impact list, since a user counting those events sees a change.
- **The incidental engine-wide rebuild is gone.** `registerEngine()` attaches `sheetAdded → engineInstance.rebuildAndRecalculate()` (`register.ts:143`), so every data load used to trigger a full rebuild as a side effect. `setSheetContent()` plus the existing `renderDependentSheets(dependentCells)` should cover cross-sheet propagation, and the cross-sheet specs are green — so I expect nothing here. It is just the first place to look if a cross-sheet regression turns up later.

**Follow-up ticket, not this PR — the other half of the leak.** `unregisterEngine()` (`register.ts:178`) removes the instance from the registries and destroys the engine only when the last instance leaves. A grid destroyed while others still share the engine leaves its sheet behind — cells and dependency graph — with nothing to remove it. Same leak class as the one fixed here, different trigger. This PR does not make it worse.

handsontable/src/plugins/formulas/formulas.ts
budnix: Nice catch — the guard now compares sheet ids in `22d62f0cbd`, with a case-mismatch case added to the rename test.

handsontable/src/plugins/formulas/formulas.ts
budnix: Good point — the else branch now empties the sheet and warns, in `22d62f0cbd`.

handsontable/src/plugins/formulas/__tests__/memoryLeak.spec.js
budnix: Agreed on the convention — switched the single-grid tests to `await loadData()` in `22d62f0cbd`. (The helper does not actually await a render flush, and the two-grid tests keep direct calls since it only targets the first container.)

handsontable/src/plugins/formulas/engine/register.ts
budnix: Good spot — canonicalized in `05d73a6840`, inside `#updateSheetNameAndSheetId` rather than `setupSheet` so `switchSheet` is covered too. Worth knowing: it also needed the `updatePlugin` name check moved to sheet ids, or re-passing the same lowercase `sheetName` fired a real sheet switch.

handsontable/src/plugins/formulas/engine/register.ts
budnix: Agreed — changed to `@returns {string}` in `05d73a6840`.

handsontable/src/plugins/formulas/formulas.ts
budnix: Fair point — added that line to the PR description. Leaving the configured-name asymmetry alone as you suggest, since it predates this change.

demtario (approved):

Haxikowy (approved):

PR #13216, opened 2026-08-21
DEV-2586: Fix mobile drag-to-scroll at the grid edge (#11658)

### LINKED ISSUE #11658: On mobile phones, DragToScroll does not work effectively. After dragging to the edge, it cannot trigger automatic scrolling.
### Describe the bug

On mobile phones, DragToScroll does not work effectively. After dragging to the edge, it cannot trigger automatic scrolling.

### Video/Screenshots

_No response_

_No response_

### Handsontable version

last version

### Framework version

_No response_

### Your environment

chrome mobile

budnix:
### Context

The PR fixes [#11658](https://github.com/handsontable/handsontable/issues/11658): on a phone, dragging a selection handle to the edge of the grid does not scroll the viewport, so the selection stops growing at the last cell that is already on screen. Reported in June 2025 and confirmed still present on `develop`.

**Two independent causes, both had to be fixed:**

1. `DragToScroll` bound `mousemove` only. No browser fires `mousemove` while a finger is down, so the auto-scroller never received a position on mobile.
2. `MultipleSelectionHandles` is the only drag path a phone has — `border.ts` gates the desktop `selectionHandles` and `moveCells` bands behind `!isMobileBrowser()`, so `afterOnSelectionHandleMouseDown` never fires there. It never armed `DragToScroll`, and it resolved the finger with `elementFromPoint()`, which returns nothing usable past the grid edge.

**What changed:**

- `DragToScroll` now also binds `touchstart` / `touchmove` / `touchend` / `touchcancel`, and arms itself through `getPlugin('multipleSelectionHandles').isDragged()`. The ordering is guaranteed by DOM bubbling, not `PLUGIN_PRIORITY`: the handle plugin listens on `rootElement`, a descendant of the document this listener sits on. Mouse and touch now share one `#trackPointer()`.
- `MultipleSelectionHandles` resolves the finger with `getCellCoordsFromMousePosition()` (clamps to the viewport, returns the nearest cell) and re-extends the selection on `afterScroll`, so the drag carries on while the viewport moves under a resting finger.
- Two touch helpers added to the public `helpers/dom/event.ts`: `hasTouchList()` and `getFirstTouchPoint()`. They detect the touch list by property, never `instanceof TouchEvent`, which is `false` for an event that crossed an iframe boundary and does not exist at all on desktop Safari. Registered in `dom.types.ts`.

**Measured before/after** (Pixel 7, Galaxy S9+ and iPhone 14, Chrome mobile emulation with real touch input over CDP):

| | before | after |
|---|---|---|
| vertical scroll while holding at the edge | 0 → 0 | 59 → 262 |
| selection reached | row 9 | row 19 |
| horizontal scroll while holding | 7 → 7 | 77 → 506 |
| desktop mouse drag (control) | 30 → 291 | 30 → 291, unchanged |

### Two things to look at in review

1. **On-screen mobile drag behavior changes too, deliberately.** The finger now resolves to the nearest cell instead of requiring a rendered `td` underneath it. That is what makes edge dragging work, and it matches how desktop already behaves — but it is broader than "auto-scroll now works".
2. **`lastSetCell` was removed** from `MultipleSelectionHandles` — an unreferenced field on a `@private` plugin (no reads anywhere in the repo, wrappers or tests). Element identity is the wrong de-duplication key: Walkontable reuses `td` elements across scrolls, so under a still finger the element never changes while the cell it represents does. It is now coordinate-based.

### How has this been tested?

Every new test was confirmed to **fail without the fix**, by reverting only the two plugin sources (unit) and by running against `develop`'s own bundle (E2E).

- **7 new unit tests** — `handsontable/src/plugins/dragToScroll/__tests__/mobileTouchAutoScroll.unit.js`. 4 fail without the fix; the 3 that pass are the setup check and two negative guards.
- **4 new Playwright E2E tests** — `tests/e2e/mobile-drag-to-scroll.spec.ts`, driving real touch through CDP (`page.touchscreen` only taps, so it cannot reproduce this). 3 fail without the fix. They assert **progress while the finger rests**, not that an offset is merely non-zero: extending a selection onto a partially visible row or column scrolls it into view on its own, so `scrollTop > 0` passes with the auto-scroller dead.
- **8 new unit tests** for `hasTouchList()` / `getFirstTouchPoint()`.
- Legacy Jasmine drag assertions verified unchanged in a real browser (the specs' own expectations): `[[1,1,1,3]]` horizontal and `[[1,1,3,1]]` vertical, in both LTR and RTL.

```bash
# Unit (3611 pass)

# Legacy Jasmine E2E (9627 specs, 0 failures, 8 pending)

# Playwright, all six legs (1780 pass; the new spec passes on 3 themes x 2 bundles)
cd tests && npx playwright test

# Types and lint (0 errors)
```

Manually verified with mobile emulation on a local demo page comparing `develop` against this branch.

Pre-existing on `develop`, **not** introduced here: `tests/e2e/fill-handle-formulas.spec.ts` fails on the two `horizon` legs. Verified it fails identically on `develop`'s own bundle.

### Types of changes

### Related issue(s):

1. DEV-2586
2. Fixes #11658

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2586

---

> [!NOTE]
> **Medium Risk**
> Touches selection and auto-scroll gesture handling on mobile (multi-finger tracking, document-level touch listeners). Desktop mouse drag is intended to stay unchanged, but regressions would show up as stuck or runaway selection while dragging.
> 
> **Overview**
> Fixes **#11658**: dragging a mobile selection handle to the grid edge now auto-scrolls the viewport and keeps extending the selection.
> 
> **`dragToScroll`** now listens for `touchstart`/`touchmove`/`touchend`/`touchcancel` (browsers never fire `mousemove` while a finger is down) and arms itself when `multipleSelectionHandles.isDraggedBy(identifier)` is true. Mouse and touch share `#trackPointer()`. Gestures follow one finger by `Touch.identifier`, and end only when *that* finger lifts or is cancelled.
> 
> **`multipleSelectionHandles`** maps the finger with `getCellCoordsFromMousePosition()` (clamps past the edge to the nearest cell) instead of `elementFromPoint()`, re-extends on `afterScroll` while the finger rests, and de-duplicates by coordinates (Walkontable reuses `td`s). Also fixes releasing the wrong handle when two are held, and tracks fingers so a resting thumb is not treated as the drag.
> 
> Adds public touch helpers `hasTouchList`, `getFirstChangedTouch`, and `getTouchPointById` (property-based, not `instanceof TouchEvent`). Covered by unit tests plus Playwright CDP touch drags that assert scroll *progress* while the finger stays past the edge.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eb79af4c9ec6981b8023d37171f33f917bfad2b5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2586 Mobile DragToScroll: dragging a selection handle to the grid edge does not auto-scroll](https://app.clickup.com/t/9015210959/DEV-2586)

handsontable/src/plugins/dragToScroll/AGENTS.md
```diff
@@ -0,0 +1,61 @@
...
+
+`#setupListening(kind, event, controller)` computes the boundaries and starts listening. It is reached five ways, and four of them are hooks:
+
+| Drag | Entry point | Guard |
+|---|---|---|
+| plain cell drag-select | `beforeOnCellMouseDown` | skipped when `selectionMode: 'single'` |
+| autofill corner | `afterOnCellCornerMouseDown` | — |
+| move-cells edge band | `afterOnSelectionEdgeMouseDown` | `getPlugin('moveCells').isDragActive()` |
+| desktop selection handle | `afterOnSelectionHandleMouseDown` | `getPlugin('selectionHandles').isDragActive()` |
+| **mobile selection handle** | its own document `touchstart` | `getPlugin('multipleSelectionHandles').isDragged()` |
```
demtario: This row documents the mobile guard as `getPlugin('multipleSelectionHandles').isDragged()`, but the shipped code calls `isDraggedBy(touch.identifier)` (`dragToScroll.ts:458`) — precisely because `isDragged()` alone would arm auto-scroll for any finger landing anywhere while a drag happens to be running, as the inline comment there explains.

The same pre-`27a821f9` design is documented in three more places:

- `multipleSelectionHandles/AGENTS.md:14` — "DragToScroll does exactly that, by asking `isDragged()` from its own document-level `touchstart` listener"
- `multipleSelectionHandles/AGENTS.md:28` — "`isDragged()` is public because DragToScroll reads it"
- `multipleSelectionHandles/AGENTS.md:33` — "`isDragged()` stays `true`, so every later touch arms auto-scroll with no handle press"
- `multipleSelectionHandles.ts:87` (the `disablePlugin()` JSDoc) — "DragToScroll arms auto-scroll on `isDragged()` alone"

`isDragged()` now has no production callers at all: only the unit suite, the new spec, and `tests/fixtures/pages/windowTypes.ts:47`. These files are the PR's own deliverable and exist to be right for whoever reads them next, so worth a pass to point them at `isDraggedBy()` and to say why the per-finger check replaced the boolean one.

handsontable/src/plugins/multipleSelectionHandles/multipleSelectionHandles.ts
```diff
@@ -406,4 +457,130 @@ export class MultipleSelectionHandles extends BasePlugin {
...
+   * Extends the selected range to the cell under the given viewport position.
+   *
+   * The position is resolved with `getCellCoordsFromMousePosition`, which clamps it to the viewport
+   * and returns the nearest cell. That is what lets a finger dragged past the grid edge keep
+   * extending the selection while DragToScroll scrolls the viewport underneath it.
+   *
+   * @param {number} clientX The finger's viewport X coordinate.
+   * @param {number} clientY The finger's viewport Y coordinate.
+   */
+  #extendSelection(clientX: number, clientY: number): void {
```
demtario, line 539: `#extendSelection` runs `getCellCoordsFromMousePosition()` on every `touchmove` that clears `TOUCH_MOVE_THRESHOLD = 2` px. That helper is not cheap: it does four scans over the visible range (`findColumnReferenceRow`, `findRowReferenceColumn`, `findColumnAtX`, `findRowAtY`), and every step of each is a `getCell()` plus a `getCellMetaTransient()`.

The asymmetry with the path this is modelled on: the mouse side calls the same helper only from `DragToScroll#onAfterScroll`, gated by the 20-500 ms scroll timer, and on-screen extension there rides Walkontable's own cheap hit path. The touch code this replaces was a single `elementFromPoint()`. So per pointer-move the mobile path is now materially heavier than the desktop one, on the weaker device, at up to 120 touchmove/s with a 2 px gate that almost nothing fails.

Two ways out, either is fine:

- raise the threshold to a fraction of the smallest cell rather than 2 px, or
- keep the clamping resolver for the case it was added for (finger outside the holder rect) and fast-path `elementFromPoint`/`getCoords` while the finger is still inside it.

Related: the new blanket bullet at `handsontable/AGENTS.md:77` ("Resolve a pointer to a cell with `getCellCoordsFromMousePosition()`, not `elementFromPoint()`") reads as unconditional, which is not what the desktop path actually does per move. Worth adding the caveat there so the next reader doesn't put this on a hot per-move path somewhere else.

demtario (changes requested):
Reviewed the touch path end to end in a separate worktree. The mechanism is right and the tests are genuine (real `Event` objects, progress-based E2E assertions rather than "offset is non-zero"), so nothing here is a correctness blocker. Two medium findings below; the rest of what I found is Low and I can drop it in a follow-up comment if useful.

handsontable/src/plugins/dragToScroll/AGENTS.md
budnix: Good spot — the docs were still describing the pre-`27a821f9` design. All five spots now point at `isDraggedBy()` in `205ea5a3`, with a line on why the per-finger check replaced the boolean. I kept `isDragged()` itself, since it is pre-existing public API.

handsontable/src/plugins/multipleSelectionHandles/multipleSelectionHandles.ts
budnix: Fair point on the cost, but this matches the desktop paths rather than exceeding them: `selectionHandles`, `moveCells`, and `autofill` all call the same helper on every mousemove with no gate at all (`selectionHandles.ts:197` -> `:221`), so the touch path adds the 2px gate and the same-cell skip on top of that. Leaving it as is, and the `AGENTS.md:77` bullet too, since it is accurate for those three.

handsontable/src/helpers/dom/event.ts
```diff
@@ -60,6 +60,96 @@ export function isTouchEvent(event: Event): boolean {
   return typeof TouchEvent !== 'undefined' && event instanceof TouchEvent;
 }
 
+/**
+ * One finger, as a touch event reports it.
+ */
+interface TouchPoint {
```
demtario: This is the one item I would settle before merge, not because it is severe but because it is the only one that cannot be taken back later.

Everything exported from this module is spread onto `Handsontable.dom` (`src/index.ts:111-113`), and under the breaking-changes policy that surface is kept forever. So merging publishes three permanent public helpers that today serve two internal call sites, with two rough edges baked in:

- `TouchPoint` and `TouchListEvent` are not exported, so a consumer cannot name `getFirstChangedTouch`'s return type — only destructure it.
- The shapes are asymmetric: `getFirstChangedTouch` returns `{identifier, clientX, clientY}` while `getTouchPointById` returns `{clientX, clientY}`. Reasonable internally (the caller already knows the id it asked for), awkward as a published pair.

Either is fine as a resolution — export the types and align the shapes, or keep the helpers plugin-local (e.g. alongside the two plugins that use them) where they stay free to change. Every other note on this PR can be fixed after merge; this one hardens on release.

handsontable/src/plugins/multipleSelectionHandles/multipleSelectionHandles.ts
demtario, line 539: Repeating the perf point from the previous review since it is still open, with what I would actually do about it.

`#extendSelection` runs `getCellCoordsFromMousePosition()` on every `touchmove` that clears `TOUCH_MOVE_THRESHOLD = 2` px. That helper does four scans over the visible range (`findColumnReferenceRow`, `findRowReferenceColumn`, `findColumnAtX`, `findRowAtY`), each step a `getCell()` plus a `getCellMetaTransient()`. The mouse path calls the same helper only from `DragToScroll#onAfterScroll`, gated by the 20-500 ms scroll timer, and on-screen extension there rides Walkontable's own cheap hit path. The touch code this replaces was a single `elementFromPoint()`. Net effect: per pointer-move the mobile path is now heavier than the desktop one, on the weaker device, at up to 120 touchmove/s through a 2 px gate that almost nothing fails.

To be fair to the change, there is no measured regression here — the before/after numbers in the description are scroll offsets, not frame timing. So this is a "cheap to pre-empt" note rather than a proven problem. Either of these closes it:

- raise the threshold to a fraction of the smallest cell instead of a flat 2 px, or
- keep the clamping resolver for the case it was added for (finger outside the holder rect) and fast-path `elementFromPoint`/`getCoords` while the finger is still inside it.

If you would rather defer, a frame-timing check on a real mid-range phone would be the thing to have before merging on the assumption it is fine.

Related, and a one-line fix: the new bullet at `handsontable/AGENTS.md:77` ("Resolve a pointer to a cell with `getCellCoordsFromMousePosition()`, not `elementFromPoint()`") reads as unconditional, which is not what the desktop path does per move. Worth the caveat so this does not get put on another hot per-move path on the strength of the rule.

handsontable/src/plugins/multipleSelectionHandles/multipleSelectionHandles.ts
```diff
@@ -406,4 +457,130 @@ export class MultipleSelectionHandles extends BasePlugin {
...
+    const selectedRange = this.hot.getSelectedRangeActive();
+
+    if (!selectedRange || this.touchStartRange === undefined) {
+      return;
+    }
+
+    const targetCoords = getCellCoordsFromMousePosition(this.hot, clientX, clientY);
+
+    // A position over the row header resolves to column -1; treat it as the first column.
```
demtario: This comment is not true of `getCellCoordsFromMousePosition`, and the code under it is unreachable. The helper's column scan starts at renderable 0 / `firstPartiallyVisibleColumn` and its row scan at `firstPartiallyVisibleRow`, and every fallback path picks one of those bounds — it never returns a negative index, so a position over the row header is already clamped into the body and both `Math.max(..., 0)` calls can never bite.

Worth fixing rather than leaving: the old code did need the `col === -1` normalization (it read `getCoords()` off whatever `elementFromPoint` returned, `TH` included), so a reader who knows the previous version will take this comment as still describing the helper's contract. Dropping the `Math.max` pair and the comment, or rewording it to say the helper clamps for you, both work.

handsontable/src/helpers/dom/__tests__/event.unit.ts
```diff
@@ -60,4 +67,97 @@ describe('DomEvent helper', () => {
       window.TouchEvent = OriginalTouchEvent;
     });
   });
+
+  /**
+   * Builds a touch event, mirroring what a browser reports.
+   *
+   * The lists are attached as plain properties because jsdom implements neither `Touch` nor
```
demtario: Small contradiction: this says jsdom implements neither `Touch` nor `TouchEvent`, and three lines further down `hasTouchList` is tested with `new TouchEvent('touchstart')` — which passes, so jsdom does provide the constructor (with empty touch lists). The reason for attaching the lists as plain properties is that jsdom's `TouchEvent` init does not populate `touches`/`changedTouches`, which is a different and more useful statement to leave for the next reader.

demtario (commented):
Re-reviewed after 205ea5a31. The docs finding is fully closed — all four sites point at `isDraggedBy` now, and the `Touch.identifier` recycling note (Chrome on Android reusing `0`) is a better justification for the `#dragTouches` reset than the one it replaced.

Four inline notes below, ordered by how much they cost to leave in. Only the first is worth settling before merge; the rest are cheap cleanups. I also checked three findings a second automated pass produced and none of them hold, so nothing from that is repeated here:

- `removeFromDragged` clearing the array at `length === 1` before matching `query`: unreachable, `dragged` and `#dragTouches` are kept in lockstep by `beginDrag`/`releaseLiftedTouches`, and the early-clear is unchanged from `develop` anyway.
- `releaseLiftedTouches` setting `released = true` unconditionally: same lockstep, and clearing the cached position costs one tick since the next `touchmove` re-resolves it.
- `#dragTouchId` latching when `#setupListening` bails: it has exactly two early returns, `isRightClick` (`button === 2`, `undefined` on a `TouchEvent`) and `kind === 'cell' && selectionMode === 'single'` (kind is `'handle'`). Neither can fire on the touch path.

One thing that is not anchorable to a line: **the PR description is stale against the final commits.** It names `getFirstTouchPoint()` (which does not exist), says "Two touch helpers added" (there are three), describes arming through `isDragged()`, and its counts — "7 new unit tests", "8 new unit tests" for the helpers — predate b2cd9e2fe/d60defc10/27a821f99; the files hold 16 and 12. The Cursor summary at the bottom is the accurate one. Worth a refresh since the body is what reviewers read as the spec.

handsontable/src/helpers/dom/event.ts
budnix: Good point — that surface hardens on release, so `TouchPoint` and `TouchListEvent` are exported now and `getTouchPointById` returns the full `TouchPoint` in `eb79af4c9e`, with `dom.types.ts` naming both return types. Kept them in `helpers/dom/event.ts` since reuse from other plugins was the reason for putting them there.

handsontable/src/plugins/multipleSelectionHandles/multipleSelectionHandles.ts
budnix: Nice catch — that comment was still describing the old `elementFromPoint` behavior; reworded in `eb79af4c9e`. I kept the floor itself, since the helper can return -1 through its empty-grid fallback where `countCols() - 1` is -1.

handsontable/src/helpers/dom/__tests__/event.unit.ts
budnix: You are right, it contradicted the test three lines down. jsdom 16.7 ships `TouchEvent` but no `Touch`, so its init cannot fill the lists — comment now says that, in `eb79af4c9e`.

handsontable/src/plugins/multipleSelectionHandles/multipleSelectionHandles.ts
budnix: I think this one still holds, and my last reply only named one call site so it was easy to miss: `selectionHandles.ts:197`, `moveCells.ts:375`, and `autofill.ts:899` each call the same helper on every mousemove, guarded only by "is a drag active" — no movement threshold. So the touch path, with its 2px gate plus the same-cell skip, is the lighter one rather than the heavier. Left `AGENTS.md:77` as is for the same reason. Happy to put frame timing from a real mid-range phone behind it if you would rather have the number than the argument.

demtario (approved):

Issue #11674, opened 2025-06-16
Total sum row does not work with NestedRows and ColumnSummary plugins


captain-orange2:
### Describe the bug

When implementing a **total sum row** which aggregates all subtotals into a total sum at the bottom row of the table, no _destination row index_ seems be correct. No matter which combination of values of `destinationRow`, `ranges` and `reversedRowCoords` I try, the error _One of the Column Summary plugins' destination points you provided is beyond the table boundaries!_ is shown. 

**Reproduction (in the stackblitz):**
1. Edit e.g. cell (1,1) and see that the total and subtotals are updated correctly
2. Close first nested row group
3. Edit e.g. cell (5,1) and see that the total is not updated (and check console message)

Possible source of this behavior: This [line](https://github.com/handsontable/handsontable/blob/master/handsontable/src/plugins/columnSummary/endpoints.js#L551) compares a physical row index with the count of the visual rows. 

Note: I created the demo by extending the example given in the [documentation](https://handsontable.com/docs/react-data-grid/column-summary/#set-up-column-summaries-using-a-function) 
>  "Using a function to provide a column summary configuration lets you set up all sorts of more complex column summaries."

### Provide a link to the demo with the bug reproduction

https://stackblitz.com/edit/react-qganvzeb?file=src%2FExampleComponent.tsx

### Handsontable version

15.3

### Your environment

Chrome

budnix:
Thanks for reporting the bug — the fix is already done (https://github.com/handsontable/handsontable/pull/13217) and will be released in upcoming releases.

PR #13217, opened 2026-08-21
DEV-2587: Fix columnSummary destination rows when rows are trimmed (#11674)

budnix:
### Context

The PR fixes a `columnSummary` total row that stops working as soon as a `nestedRows` group is collapsed, reported in #11674.

**What goes wrong, in steps:**

1. In `columnSummary`, `destinationRow` and `ranges` are **physical** row indexes — positions in the raw data array.
2. Several places in `endpoints.ts` compared them against `hot.countRows()`, which counts only the rows that are currently **visible**.
3. Collapsing a `nestedRows` group trims its children, so `countRows()` shrinks while the physical dataset stays the same size.
4. A total row at the bottom of the data now has a physical index bigger than `countRows()`. The guard decides it is off the table, logs *"One of the Column Summary plugins' destination points you provided is beyond the table boundaries!"*, and skips it. The total goes stale.

Three more symptoms came out of the same confusion while investigating:

- **Silent wrong-row write.** With `reversedRowCoords: true` the plugin computes `countRows() - N - 1` and treats the answer as a physical index. After a collapse that number points at a **different, visible row** — in the added test the grand total was written into another group's header row, overwriting its value, with no warning at all.
- **Uncaught exception.** When the destination row was itself trimmed, `toVisualRow()` returned `null` and the plugin still called `setDataAtCell(null, …)`, throwing `TypeError: Cannot set properties of null` out of `DataMap.set`.
- **Hidden summary rows counted as data.** `getCellValue()` recognizes a summary result by its `columnSummaryResult` class, but cell meta is addressed by *visual* coordinates, so a trimmed row has no readable class. A hidden summary row was therefore summed as if it were plain data, inflating every summary whose range covered it.

### The fix

Bounds checking now runs through one predicate, `Endpoints#isEndpointOutOfBounds()`, which separates the three cases that `countRows()` had conflated:

- the physical row does not exist at all — out of bounds, warn;
- the row exists but is **trimmed** — not out of bounds, it is just hidden, so skip the write silently;
- the row is visible but sits past `maxRows` — out of bounds, the grid renders no cell for it.

When nothing is trimmed this is exactly equivalent to the old `destinationRow >= countRows()` test, so **behavior is unchanged for every setup that does not trim rows.**

For the last symptom, `getCellValue()` falls back to `Endpoints#isSummaryDestination()` when the row is trimmed. The destination set is built once per refresh pass and invalidated at the three batch entry points, so function-typed settings are not re-invoked per cell.

The settings defaults (the implicit `ranges`, and the `reversedRowCoords` conversion) use `countAddressableRows()` — the physical row count capped by `maxRows`. `countRows()` was silently carrying that cap, and dropping it breaks the existing "maxRows options set" spec. A non-finite or negative `maxRows` is treated as no cap.

### Behavior changes to note

Two, both visible only when rows are trimmed:

1. **Default calculation range.** When an endpoint declares no `ranges`, the default now spans the whole dataset rather than a visible-row-count prefix. The old value was an arbitrary physical prefix rather than anything meaningful. The `maxRows` cap still applies.
2. **`reversedRowCoords` under Filters / TrimRows.** It now resolves against the physical row count, so the summary pins to the last *physical* row. Previously it followed the last *visible* row, which under a filter meant the result was written over whatever data row happened to sit at the bottom of the filtered view. Pinning is what makes the `nestedRows` case correct, but it is a change for Filters users with function-typed settings — flagging it explicitly for a reviewer decision.

### Known limitation, not addressed here

Nothing re-runs the endpoints when rows are untrimmed — `columnSummary` registers no hook for expand, untrim or filter. So a destination row that was hidden at the moment of a change keeps a stale value, **and misses its `readOnly` / `columnSummaryResult` styling**, until the next recalculation that happens while it is visible. Both recover together at that point. This is pre-existing and orthogonal to this fix; there is no single generic untrim hook to attach to, and a naive `rowIndexMapper` listener would loop back on itself because the refresh writes through `setDataAtCell`. Worth its own ticket.

### How has this been tested?

Four cases added to `handsontable/src/plugins/columnSummary/__tests__/columnSummary.spec.js`, all four verified red against `origin/develop` and green with the change:

- a grand-total row at the bottom of a nested dataset stays correct after collapsing a group;
- `reversedRowCoords: true` resolves against the physical row count, so the total stays in the last physical row instead of overwriting another group's header row;
- a destination row that is itself trimmed no longer throws, and picks up the correct value on the next recalculation once it is visible;
- a trimmed summary row is not counted as data by another summary that covers it.

```bash
npm run test:e2e --prefix handsontable -- --testPathPattern='columnSummary|nestedRows|trimRows|filters|hiddenRows'   # 920 specs, 0 failures
npm run test:unit --prefix handsontable                                                                              # 3597 tests, 0 failures
npm run eslint --prefix handsontable                                                                                 # 0 errors
npm run stylelint --prefix handsontable                                                                              # clean
npm run test:types --prefix handsontable                                                                             # clean
```

### Types of changes

### Related issue(s):

1. #11674
2. DEV-2587

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2587

---

> [!NOTE]
> **Medium Risk**
> Touches columnSummary calculation, bounds, and write paths used with NestedRows/TrimRows/Filters. Behavior of default ranges and reversedRowCoords changes when rows are trimmed; untrimmed tables should match previous behavior.
> 
> **Overview**
> Fixes `columnSummary` treating trimmed rows as missing. Destination rows and default ranges are physical indexes, but they were compared against `countRows()`, so collapsing a NestedRows group (or TrimRows/Filters) produced false out-of-bounds warnings, wrote results into the wrong visible row when `reversedRowCoords` was on, and crashed when writing to a hidden destination.
> 
> Bounds checks now use `isEndpointOutOfBounds()` against the physical row count (still capped by `maxRows`). Writes and resets skip trimmed destinations instead of calling `setDataAtCell(null, …)`. Hidden summary cells are excluded from other summaries via a per-refresh destination cache, because `columnSummaryResult` cell meta is not readable without visual coordinates.
> 
> Default `ranges` and `reversedRowCoords` now resolve against physical (maxRows-capped) length, so totals stay on the last physical row rather than the last visible one. Specs cover collapse, reversed coords, hidden destinations, and trimmed subtotals.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7cf10236783dec26a27d927bc8b63d6ef0ed3dc0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/handsontable/src/plugins/columnSummary/__tests__/columnSummary.spec.js
+++ b/handsontable/src/plugins/columnSummary/__tests__/columnSummary.spec.js
@@ -978,6 +978,9 @@ describe('ColumnSummarySpec', () => {
 
       data.push({ a: 'TOTAL', b: null });
 
+      // installed before the grid so init-time warnings land inside the spied window
+      const warnSpy = spyOnConsoleWarn();
+
       handsontable({
         data,
         height: 400,
@@ -1018,8 +1021,6 @@ describe('ColumnSummarySpec', () => {
           ];
         },
       });
-
-      const warnSpy = spyOn(console, 'warn');
 
       await waitForNextAnimationFrames(2);
 
@@ -1164,6 +1165,37 @@ describe('ColumnSummarySpec', () => {
     // 10 + 2 + 3, without the hidden subtotal of 6 added on top
     expect(getDataAtCell(hot().toVisualRow(4), 0)).toEqual(15);
   });
+
+  it('should build the default `ranges` from the physical row count when rows are already trimmed (#11674)',
+    async() => {
+      handsontable({
+        data: [[1], [2], [3], [4], [null]],
+        trimRows: [1, 2],
+        columnSummary() {
+          return [
+            {
+              // no `ranges`, so the default range is built while rows 1 and 2 are already trimmed
+              destinationRow: 4,
+              destinationColumn: 0,
+              type: 'sum',
+              forceNumeric: true,
+            },
+          ];
+        },
+      });
+
+      await waitForNextAnimationFrames(2);
+
+      const plugin = getPlugin('ColumnSummary');
+
+      // 5 physical rows, so the default range spans them all - not the 3 visible ones
+      expect(plugin.endpoints.getAllEndpoints()[0].ranges).toEqual([[0, 4]]);
+
+      // force a recalculation and check the sum covers the trimmed rows too
+      await setDataAtCell(0, 0, 10);
+
+      expect(getDataAtCell(hot().toVisualRow(4), 0)).toEqual(19);
+    });
 
   describe('maxRows options set', () => {
     it('should apply summary operation only on rows which are < maxRows', async() => {
--- a/handsontable/src/plugins/columnSummary/__tests__/columnSummary.spec.js
+++ b/handsontable/src/plugins/columnSummary/__tests__/columnSummary.spec.js
@@ -972,6 +972,114 @@ describe('ColumnSummarySpec', () => {
       expect(spec().$container.find('.columnSummaryResult').size()).toEqual(3);
       expect(spec().$container.find('.htDimmed').size()).toEqual(3);
     });
+
+    it('should keep the grand-total endpoint working when a nested group is collapsed (#11674)', async() => {
+      const data = getDataForColumnSummary();
+
+      data.push({ a: 'TOTAL', b: null });
+
+      handsontable({
+        data,
+        height: 400,
+        width: 400,
+        rowHeaders: true,
+        nestedRows: true,
+        columnSummary() {
+          return [
+            {
+              destinationRow: 0,
+              destinationColumn: 1,
+              type: 'sum',
+              forceNumeric: true,
+              ranges: [[1, 3]],
+            },
+            {
+              destinationRow: 4,
+              destinationColumn: 1,
+              type: 'sum',
+              forceNumeric: true,
+              ranges: [[5, 6]],
+            },
+            {
+              destinationRow: 7,
+              destinationColumn: 1,
+              type: 'sum',
+              forceNumeric: true,
+              ranges: [[8, 9]],
+            },
+            {
+              // physical row 10 - the grand-total row at the bottom
+              destinationRow: 10,
+              destinationColumn: 1,
+              type: 'sum',
+              forceNumeric: true,
+              ranges: [[1, 3], [5, 6], [8, 9]],
+            },
+          ];
+        },
+      });
+
+      const warnSpy = spyOn(console, 'warn');
+
+      await waitForNextAnimationFrames(2);
+
+      expect(getDataAtCell(10, 1)).toEqual(4251);
+
+      getPlugin('nestedRows').collapsingUI.collapseChildren(0);
+
+      await waitForNextAnimationFrames(2);
+
+      expect(getDataAtCell(7, 1)).toEqual(4251);
+
+      await setDataAtCell(2, 1, 0);
+
+      expect(warnSpy).not.toHaveBeenCalledWith(warnMessage);
+      // the edited row belongs to the second group, so its subtotal and the grand total drop by 363
+      expect(getDataAtCell(1, 1)).toEqual(3633);
+      expect(getDataAtCell(7, 1)).toEqual(3888);
+    });
+
+    it('should resolve `reversedRowCoords` against the physical row count when a group is collapsed (#11674)',
+      async() => {
+        const data = getDataForColumnSummary();
+
+        data.push({ a: 'TOTAL', b: null });
+
+        handsontable({
+          data,
+          height: 400,
+          width: 400,
+          rowHeaders: true,
+          nestedRows: true,
+          columnSummary() {
+            return [
+              {
+                destinationRow: 0,
+                reversedRowCoords: true,
+                destinationColumn: 1,
+                type: 'sum',
+                forceNumeric: true,
+                ranges: [[1, 3], [5, 6], [8, 9]],
+              },
+            ];
+          },
+        });
+
+        await waitForNextAnimationFrames(2);
+
+        expect(getDataAtCell(10, 1)).toEqual(4251);
+
+        getPlugin('nestedRows').collapsingUI.collapseChildren(0);
+
+        await waitForNextAnimationFrames(2);
+
+        // force a recalculation while the group stays collapsed
+        await setDataAtCell(2, 1, 0);
+
+        // the result stays in the last physical row instead of jumping to the row that is now `countRows() - 1`
+        expect(getDataAtCell(7, 1)).toEqual(3888);
+        expect(getDataAtCell(4, 1)).toEqual('7');
+      });
   });
 
   describe('maxRows options set', () => {
--- a/handsontable/src/plugins/columnSummary/__tests__/columnSummary.spec.js
+++ b/handsontable/src/plugins/columnSummary/__tests__/columnSummary.spec.js
@@ -1126,6 +1126,45 @@ describe('ColumnSummarySpec', () => {
       });
   });
 
+  it('should not count a trimmed summary row as data in another summary (#11674)', async() => {
+    handsontable({
+      data: [[1], [2], [3], [null], [null]],
+      columnSummary: [
+        {
+          destinationRow: 3,
+          destinationColumn: 0,
+          type: 'sum',
+          forceNumeric: true,
+          ranges: [[0, 2]],
+        },
+        {
+          // no `ranges`, so the default range spans every physical row - including row 3, which
+          // holds the subtotal above
+          destinationRow: 4,
+          destinationColumn: 0,
+          type: 'sum',
+          forceNumeric: true,
+        },
+      ],
+      trimRows: true,
+    });
+
+    await waitForNextAnimationFrames(2);
+
+    expect(getDataAtCell(3, 0)).toEqual(6);
+    expect(getDataAtCell(4, 0)).toEqual(6);
+
+    getPlugin('trimRows').trimRow(3);
+
+    await waitForNextAnimationFrames(2);
+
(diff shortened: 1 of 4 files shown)


qunabu:
Task linked: [DEV-2587 ColumnSummary total row breaks when NestedRows collapses a group (#11674)](https://app.clickup.com/t/9015210959/DEV-2587)

handsontable/src/plugins/columnSummary/columnSummary.ts
```diff
@@ -499,13 +499,16 @@ export class ColumnSummary extends BasePlugin {
...
-    if (visualRowIndex !== null) {
-      cellClassName = (this.hot.getCellMetaTransient(visualRowIndex, col).className as string) || '';
-    }
+    // A trimmed row has no visual coordinates, so its cell meta - and with it the
+    // `columnSummaryResult` class - cannot be read. Fall back to the endpoint destinations, or a
+    // hidden summary row is summed as plain data and inflates every summary covering it.
+    const isSummaryResult = visualRowIndex !== null
+      ? ((this.hot.getCellMetaTransient(visualRowIndex, col).className as string) || '')
+        .indexOf('columnSummaryResult') > -1
+      : this.endpoints!.isSummaryDestination(row, col);
```
Haxikowy, line 513: The visible branch keys only on the className, so it misses a destination row that was trimmed when the meta was last written. `setEndpointValue` skips `_setCellMetaDeclarative` for trimmed rows, so once that row is untrimmed it has no class and gets summed as data by any summary covering it.

```js
const isSummaryResult = this.endpoints!.isSummaryDestination(row, col) ||
  (visualRowIndex !== null &&
    ((this.hot.getCellMetaTransient(visualRowIndex, col).className as string) || '')
      .indexOf('columnSummaryResult') > -1);
```

A destination cell is never data, so OR-ing is safe in both directions.

handsontable/src/plugins/columnSummary/endpoints.ts
```diff
@@ -169,7 +253,7 @@ class Endpoints {
     arrayEach(settingsArray, (val: EndpointConfig) => {
       const newEndpoint: EndpointConfig = {};
 
-      this.assignSetting(val, newEndpoint, 'ranges', [[0, this.hot.countRows() - 1]]);
+      this.assignSetting(val, newEndpoint, 'ranges', [[0, this.countAddressableRows() - 1]]);
```
Haxikowy, line 276: This range span is the loop bound in `getPartialSum`, `getPartialMinMax` and `countEmpty`. On 100k rows with a filter showing 500 and no explicit `ranges`, one `afterChange` walks 100,000 iterations instead of 500, per endpoint, per recalculation.

Semantics are correct and I'm not asking you to revert. Worth a `performance-tests/` scenario for default `ranges` over a filtered large dataset, and a line in the guide that omitting `ranges` costs a full dataset pass regardless of filtering.

Also untested: the one test that looks like it covers this passes array settings, so `parseSettings` runs at `afterInit` while nothing is trimmed and both counts agree. It proves the `isSummaryDestination` fix only. The change bites when trimming exists at parse time, or with function settings.

handsontable/src/plugins/columnSummary/endpoints.ts
```diff
@@ -135,6 +190,35 @@ class Endpoints {
...
+   * addressed by visual coordinates, so a trimmed row has no readable class. Without this check a
+   * hidden summary row is summed as if it were plain data and inflates every summary covering it.
+   *
+   * @param {number} physicalRow Physical row index.
+   * @param {number} column Column index.
+   * @returns {boolean}
+   */
+  isSummaryDestination(physicalRow: number, column: number): boolean {
+    if (this.#summaryDestinations === null) {
+      this.#summaryDestinations = new Set(
```
Haxikowy: `getAllEndpoints()` here re-invokes the settings function and re-runs `parseSettings`, one line after `refreshAllEndpoints` already resolved the same endpoints. For a stateful or non-deterministic callback the destination set then describes a different layout than the endpoints being calculated.

Passing the resolved array into `invalidateSummaryDestinations()` at the three call sites also removes the user callback from inside the per-cell loop.

handsontable/src/plugins/columnSummary/endpoints.ts
```diff
@@ -135,6 +190,35 @@ class Endpoints {
...
+        this.getAllEndpoints().map(endpoint => `${endpoint.destinationRow},${endpoint.destinationColumn}`)
+      );
+    }
+
+    return this.#summaryDestinations.has(`${physicalRow},${column}`);
```
Haxikowy: Minor: the string key allocates per cell on the trimmed path, which with the default-`ranges` change is most iterations in a filtered grid. Endpoint counts are single digit, so a linear scan comparing two numbers, or `Map<number, Set<number>>`, avoids it.

handsontable/src/plugins/columnSummary/endpoints.ts
```diff
@@ -122,6 +128,55 @@ class Endpoints {
...
+
+  /**
+   * Returns the number of rows an endpoint may address, that is the physical row count capped by
+   * `maxRows`. Used for the settings defaults that need a row count rather than a bounds check.
+   *
+   * A non-finite or negative `maxRows` is treated as no cap.
+   *
+   * @returns {number}
+   */
+  countAddressableRows(): number {
```
Haxikowy, line 153: `dataMap.getLength()` collapses `maxRows <= 0` to `0` (`dataMap.ts:1092`), so this treats negative `maxRows` as no cap where core treats it as zero rows. With `maxRows: -1` the default `ranges` becomes the whole dataset instead of `[[0, -1]]`. Degenerate config, but the JSDoc above asserts behavior the rest of the codebase contradicts. Mirror the clamp or drop the sentence.

handsontable/src/plugins/columnSummary/endpoints.ts
```diff
@@ -550,6 +628,8 @@ class Endpoints {
    * @param {Set<number>|number[]} visualColumns Visual column indexes to match against.
    */
   refreshEndpointsBySourceColumns(visualColumns: Set<number> | number[]) {
+    this.invalidateSummaryDestinations();
```
Haxikowy: This runs before the `matched.length === 0` early return, so a formulas-heavy grid drops the cache on every `afterFormulasValuesUpdate` even when nothing recalculates. Moving it below the return keeps it warm.

handsontable/src/plugins/columnSummary/__tests__/columnSummary.spec.js
```diff
@@ -972,6 +972,197 @@ describe('ColumnSummarySpec', () => {
...
+              destinationColumn: 1,
+              type: 'sum',
+              forceNumeric: true,
+              ranges: [[1, 3], [5, 6], [8, 9]],
+            },
+          ];
+        },
+      });
+
+      const warnSpy = spyOn(console, 'warn');
```
Haxikowy: Nit: the file has a `spyOnConsoleWarn()` helper, and this installs the spy after `handsontable()`, so init-time warnings fall outside the window. Harmless for `not.toHaveBeenCalledWith`, but it would silently weaken the test if it ever flipped to a positive assertion.

handsontable/src/plugins/columnSummary/columnSummary.ts
budnix: Good spot, but I have to push back on this one: I tried the OR and it broke three specs (819 vs 820 in the many-endpoints one). A visible destination cell still holds the user own value on the first pass and that counts toward the sum, so the two tests have to stay exclusive. Added a comment saying why in `7cf1023`.

handsontable/src/plugins/columnSummary/endpoints.ts
budnix: You are right, that test only proved the `isSummaryDestination` fix. Added a spec with function settings and trimming already on at parse time in `7cf1023`. The perf scenario and the guide note I would rather do as a follow-up ticket.

handsontable/src/plugins/columnSummary/endpoints.ts
budnix: Agreed — swapped it for `cacheSummaryDestinations(endpoints)` in `7cf1023`, and all three call sites now resolve the endpoints once and pass them in.

handsontable/src/plugins/columnSummary/endpoints.ts
budnix: Good point — it is a `Map<number, Set<number>>` now, so no string key per cell (`7cf1023`).

handsontable/src/plugins/columnSummary/endpoints.ts
budnix: Yep, my mistake — `countAddressableRows` now mirrors the `DataMap#getLength` clamp exactly, and the JSDoc says so (`7cf1023`).

handsontable/src/plugins/columnSummary/endpoints.ts
budnix: Fair point — moved it below the early return in `7cf1023`. It still caches every endpoint, not just the matched ones, since a result has to stay excluded from the other ranges too.

handsontable/src/plugins/columnSummary/__tests__/columnSummary.spec.js
budnix: Nice catch — using `spyOnConsoleWarn()` now, installed before the grid so init-time warnings land inside the window (`7cf1023`).

Haxikowy (approved):

Issue #10672, opened 2023-12-18
A `Sheet size limit exceeded` error is being thrown when using multiple Hot instances in separate React components with a single Hyperformula instance.


wave-light:
### Describe the bug

When using the formulas plugin with multiple sheets defined in separate React components, if one of the sheets has `maxRows` prop set to a number which is less than the number of data-containing rows in another sheet, an error is thrown when the Hot component is updated.

For example:
- In this example scenario, we have a single Hyperformula instance called `hfInstance`.
- Component1 contains a Hot instance (HotInstance1), which has `maxRows` prop set to 2, and is connected to the `hfInstance` using the `formulas` prop.
- Component2 contains a different Hot instance (HotInstance2), which has the first 5 rows populated with data, and is also connected to the `hfInstance` using the `formulas` prop.
- If Component1 updates for some reason, for example a state update, HotInstance1 will throw an error.

The reason for the error, is that the `maxRows` of 2 on the Sheet1 is being compared against number of rows in the bounds of Sheet2.
This comparison is occurring in Hyperformula's BuildEngineFactory, on line 63.

Here is the stack trace:
<pre>
react-dom.development.js:22839 Uncaught Error: Sheet size limit exceeded
    at BuildEngineFactory.buildEngine (BuildEngineFactory.js:63:17)
    at BuildEngineFactory.rebuildWithConfig (BuildEngineFactory.js:46:17)
    at _HyperFormula.rebuildWithConfig (HyperFormula.js:4123:42)
    at _HyperFormula.updateConfig (HyperFormula.js:864:10)
    at Formulas.updatePlugin (formulas.mjs:652:19)
    at Formulas.onUpdateSettings (base.mjs:250:14)
    at Core.<anonymous> (base.mjs:108:65)
    at fastCall (function.mjs:295:17)
    at Hooks.run (pluginHooks.mjs:2821:23)
    at Core.runHooks (core.mjs:4426:33)
</pre>

### Video/Screenshots

https://github.com/handsontable/handsontable/assets/40989533/28382887-3e7c-420a-a501-00d43f3b31ca

### Provide a link to the demo with the bug reproduction

https://stackblitz.com/edit/stackblitz-starters-cgyc6p?description=React%20%20%20TypeScript%20starter%20project&file=package.json,src%2FApp.tsx&title=React%20Starter

### Handsontable version

14.0.0

### Your environment

MacOS, Google Chrome

adrianszymanski89:
Hi @wave-light 

[...]

PR #13218, opened 2026-08-21
DEV-2588: Stop passing the grid's maxRows/maxCols to the formulas engine

budnix:
### Context

The PR fixes `SheetSizeLimitExceededError: Sheet size limit exceeded`, thrown when grids share one HyperFormula engine and one of them sets `maxRows` lower than another grid's row count. Reported in #10672 against 14.0.0; still reproduces on `develop` (18.0.0 + HyperFormula 3.3.0) with the same stack trace.

The same option name means two different things in the two libraries:

| | Handsontable | HyperFormula |
|---|---|---|
| Scope | per grid instance | per **engine**, shared by every sheet |
| Purpose | how many rows to display | memory ceiling (`Engine` category, default 40000) |

The plugin passed the grid's per-instance display limit into that engine-wide setting, both when it created the engine and on every `updateSettings`. Any engine can end up shared, so the smallest limit among the attached grids became the ceiling for every sheet, and HyperFormula's rebuild threw over a taller one. HyperFormula behaves as documented — this is ours.

**The fix stops passing `maxRows`/`maxColumns` to the engine entirely** — not on update, and not at creation either. Handsontable already enforces `maxRows`/`maxCols` on its own data, so the engine does not need to.

Doing it only on the update path is not enough, because an engine created by one grid is routinely shared with the next: the formulas guide recommends exactly that (`engine: hot1.getPlugin('formulas').engine`). With creation-time bounding still in place, `hot1` with `maxRows: 2` produced an engine that threw as soon as `hot2` loaded five rows into it. That path throws on `develop` too and is now covered by a test.

Two details the merge has to get right:

- The limits must be dropped from **both** sources on the update path. A `maxRows` written inside `formulas.engine` reaches the same merge through the user settings, and on `develop` it was only ever neutralised because the grid's own value spread last.
- `DEFAULT_SETTINGS` now pins `maxRows: Infinity`. Handsontable's `maxRows` defaults to `Infinity` and was always passed on, so an engine the plugin builds has never been bounded; without this it would silently fall back to HyperFormula's 40000 and break large grids.

`maxColumns` is left unset, exactly as before: it was read from a Handsontable option that does not exist (the option is `maxCols`), so it always resolved to `undefined` and HyperFormula fell back to its own default. A `maxRows`/`maxColumns` inside `formulas.engine` stays dead config, as on `develop` — making it live would mean a grid taller than that value starts throwing at construction.

**One user-visible behavior change.** Inserting more rows than `maxRows` allows used to cancel the whole insert, because the engine carried the mirrored limit and `beforeCreateRow` returned `false`. The engine no longer carries it, so `dataMap.createRow` caps the insert at `maxRows` — the same result a grid without the plugin gives. This also fixes the flip side: raising `maxRows` with `updateSettings` now lets rows be added again, where a limit baked into the engine at creation used to block them for good. Both are covered by tests. Not a breaking change under the policy — no default value changed and no API was removed.

No documentation states that these settings reach the engine (checked `docs/content/guides/formulas/` and the plugin JSDoc), so there is nothing to correct there.

### How has this been tested?

Unit tests for the merge contract, written against `develop` first and confirmed to fail on it:

```bash
npm run test:unit --prefix handsontable -- --testPathPattern=formulas/engine
```

E2E specs in `initialization.spec.js` covering all four cases below, plus the updated insert-cap spec in `formulas.spec.js`:

```bash
npm run test:e2e --prefix handsontable -- --testPathPattern=formulas   # 337 specs, 0 failures
npm run test:unit --prefix handsontable                                # 3605 passed
npm run test --prefix wrappers/react-wrapper                           # 75 passed
npm run eslint --prefix handsontable                                   # 0 errors
```

Verified in headless Chrome against built bundles, before and after:

| Scenario | `develop` | This PR |
|---|---|---|
| Two grids sharing an externally built engine, `maxRows: 2` on the smaller one | throws | OK, engine limit untouched |
| Grid 2 reusing grid 1's engine, the style the guide recommends | throws | OK |
| 100 rows with `engine: { hyperformula, maxRows: 10 }` | OK | OK, formulas compute |
| `maxRows: 2`, raised to 100, then insert 5 rows | 7 rows | 7 rows |

### Types of changes

### Related issue(s):

1. DEV-2588
2. #10672
3. Upstream, same root cause for the column axis: https://github.com/handsontable/hyperformula/issues/1185

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2588

<sub>The commit history shows three passes at this: the first two bounded the engine at creation and were walked back after review found they broke engine reuse and froze the engine's limit at its starting value. Squash-merging, so only the title lands in history.</sub>

### DIFF
--- a/.changelogs/10672.json
+++ b/.changelogs/10672.json
@@ -1,6 +1,6 @@
 {
   "issuesOrigin": "public",
-  "title": "Fixed a `Sheet size limit exceeded` error thrown when several grids shared one HyperFormula instance and one of them set `maxRows`.",
+  "title": "Fixed a `Sheet size limit exceeded` error thrown when grids sharing one HyperFormula instance used different `maxRows` values. The grid's `maxRows` and `maxCols` no longer limit the engine, so an oversized insert is now capped at the limit instead of being rejected.",
   "type": "fixed",
   "issueOrPR": 10672,
   "breaking": false,
--- a/.changelogs/10672.json
+++ b/.changelogs/10672.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed a `Sheet size limit exceeded` error thrown when several grids shared one HyperFormula instance and one of them set `maxRows`.",
+  "type": "fixed",
+  "issueOrPR": 10672,
+  "breaking": false,
+  "framework": "none"
+}
--- a/handsontable/src/plugins/formulas/__tests__/formulas.spec.js
+++ b/handsontable/src/plugins/formulas/__tests__/formulas.spec.js
@@ -1666,7 +1666,7 @@ describe('Formulas general', () => {
   });
 
   describe('hyperformula alter operation blocks', () => {
-    it('should block creating too many rows', async() => {
+    it('should cap row creation at `maxRows`, like a grid without the plugin', async() => {
       handsontable({
         data: [],
         formulas: {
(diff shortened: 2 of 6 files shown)


qunabu:
Task linked: [DEV-2588 Fix #10672: `Sheet size limit exceeded` when grids share one HyperFormula instance](https://app.clickup.com/t/9015210959/DEV-2588)

adrianspdev (approved):

Issue #10691, opened 2024-01-03
[Bug]: formula updated BUT cannot be Catched,like afterChange(),afterSetDataAtCell()


yanyinqing:
### Describe the bug

if I move some rows,orther use this cells to make formula,they will change,but you cannot catch this change.

### Handsontable version

14.0.0

### Framework version

vue2

### Your environment

edge,macos

adrianszymanski89:
Hi @yanyinqing 

In your case, I think the `afterRowMove` hook might be helpful. Please take a look at this example and let me know if that meets your requirements: https://jsfiddle.net/handsoncode/cje06gmx/

yanyinqing:
> Hi @yanyinqing
> 
> In your case, I think the `afterRowMove` hook might be helpful. Please take a look at this example and let me know if that meets your requirements: https://jsfiddle.net/handsoncode/cje06gmx/



What I mean is that there may be multiple references to this cell, and I cannot record all of them and search for their new formulas in afterRowMove

yanyinqing:
What I mean is that there may be multiple references to this cell, and I cannot record all of them and search for their new formulas in afterRowMove

yanyinqing:
> 你好
> 
> 就你而言，我认为钩子可能会有所帮助。请看一下这个例子，让我知道它是否符合你的要求：https://jsfiddle.net/handsoncode/cje06gmx/`afterRowMove`

What I mean is that there may be multiple references to this cell, and I cannot record all of them and search for their new formulas in afterRowMove

yanyinqing:
and ,delete row ,add row have this bug ,too

adrianszymanski89:
Hi @yanyinqing 

HyperForumla has a native method for creating a dependency graph: https://hyperformula.handsontable.com/guide/dependency-graph.html#getting-the-immediate-dependents-of-a-cell-or-a-range 

However, it still will need to be checked in the `afterRowMove` hook to get updated data.

Can you also tell me more about what bug exactly you have with deleting/adding rows?

AMBudnik:
Hi @yanyinqing do we have any updates regarding this issue report?

lin199303:
Hi @yanyinqing @adrianszymanski89 I also have this problem, can you tell me how to solve this problem?

yanyinqing:
> Hi @yanyinqing @adrianszymanski89 I also have this problem, can you tell me how to solve this problem?

[...]

PR #13219, opened 2026-08-21
DEV-2589: Persist engine-rewritten formulas to the source data

### LINKED ISSUE #10388: Auto Formula Update is overriden by Redux state update
### Describe the bug

We've upgraded our app to use Handsontable 12.4.0 to get the auto formula update when adding/removing/moving rows/columns. Great feature btw!

However, we are experiencing issues where the auto formula update is overridden by the re-render caused a redux state update.

In the example I provided, I am storing the data and cellSelection in the redux state.
1. After adding a new row in between the range referenced by the formula, it correctly updates the formula.
2. Selecting another cell triggers a Redux state change, which causes the component to re-render.
3. This changes the automatically updated formula to the value it was before.

Can you suggest what would be the correct way of going about this?

### Video/Screenshots

https://github.com/handsontable/handsontable/assets/83485085/19752ea4-1780-4066-8617-df6f894b4a83

https://jsfiddle.net/mL8tajev/33/

### Handsontable version

12.4.0

### Framework version

_No response_

### Your environment

Chrome, macOS Ventura

budnix:
### Context

The PR fixes [#10388](https://github.com/handsontable/handsontable/issues/10388): a formula that HyperFormula auto-updates after a row is inserted reverts to its old text as soon as the app re-renders from its own store.

The reporter uses React and Redux, but neither is the cause. The same thing happens in plain JavaScript:

| Step | HyperFormula | The array passed to Handsontable | Displayed |
|---|---|---|---|
| initial | `=SUM(A1:A3)` | `=SUM(A1:A3)` | 60 |
| after `alter('insert_row_above', 1)` | `=SUM(A1:A4)` | `=SUM(A1:A3)` (stale) | 60 |
| after `updateSettings({ data })` with that array | `=SUM(A1:A3)` | `=SUM(A1:A3)` | **30** |

**Root cause.** HyperFormula does its job — `getCellSerialized()` returns the correctly rewritten formula after every structural change. Handsontable, however, never stored it. `#onAfterCreateRow` and its siblings called `engine.addRows(...)` and then only `renderDependentSheets()`, which repaints. The new formula text was projected onto *reads* by `#onModifySourceData`, so `getSourceData()` looked right while the array underneath still held the old formula.

That is invisible while the grid owns the data, and breaks the moment anyone round-trips the array — a Redux store, a React `data` prop, a snapshot saved to a server. It was made harder to spot by Handsontable splicing the new row into the caller's array in place: the array got the right *shape* and looked synced, while its formula strings were silently wrong.

**The fix.** After each structural change, read the sheet's formulas from the engine and write the ones that differ back into the source data, fenced so `afterSetSourceDataAtCell` does not push them straight back into HyperFormula.

Six notes on the shape of it:

1. **The engine's reported changes cannot drive this.** `addRows` returns the cells whose *value* changed, and a reference shift usually leaves the value intact — inserting an empty row into a `SUM` range returns an empty change list. So the sheet's formulas are read in bulk via `getSheetFormulas()` and diffed.
2. **Undo and redo are covered, but only for structural actions.** The structural hooks return early for blocked sources, so during an undo the engine reverts through its own stack and the source data would otherwise be left behind — one undo restored the engine but left the caller's array holding the *new* formula. The sync in `afterUndo`/`afterRedo` is gated on the action being a row or column insert or removal: an unconditional sync also fired for a plain cell edit, and in a grid with moved rows that persisted the move-frame formulas into the developer's array — exactly the write that is deliberately kept out of the move hooks.
3. **The sync is skipped entirely once the engine's index order stops matching the physical order.** A move, a sort, or a column freeze reorders the engine's rows while the source data keeps its own physical order, so from that point the two hold formulas in different reference frames and copying across them would corrupt the array. The existing `manualRowMove` specs pin that representation (see the comment in `plugins/manualRowMove.spec.js`). Three tests guard it: a plain move, an insert *after* a move, and an edit undone while the grid is moved.
4. **Because of that guard, the engine index is the physical index**, so the write walks the source data directly instead of going through the visual index. That matters for correctness: trimmed rows have no visual index, so a formula in a row hidden by a filter would otherwise have been skipped and left stale.
5. **A difference in spelling alone is not a change.** HyperFormula reports a canonical form, so `=sum( a1 : a2 )` comes back as `=SUM( A1:A2 )`. Comparing raw strings would rewrite formulas the operation never touched — including ones holding no cell reference at all — and flood a diffing store with changes on every insert. The stored text is normalized through `normalizeFormula` before the two are compared, so only a real reference shift is written.
6. **`moveCells` is covered in both directions.** Dragging a range makes the engine rewrite formulas elsewhere that pointed at it, the same way an insert does. Undo needs its own handling: `MoveCellsAction.undo` restores the two regions with `restoreRegion` instead of replaying the move, so `afterMoveCells` never fires and the sync is driven from `afterUndo`. Redo *does* replay the move, so `move_cells` stays out of the shared structural set — otherwise the sheet would be scanned twice.

**Side benefit.** External listeners still receive `afterSetSourceDataAtCell` for these writes, so an outside store can now learn the new formula text. That partially answers [#10691](https://github.com/handsontable/handsontable/issues/10691), which was closed as inactive with the same root cause. The payload carries the real previous value too — the same fence switches the read projection off while Core builds it — so a store that compares old against new sees the change.

**Known gaps, stated so they are not a surprise.**

- **Sorted or reordered grids get no write-back at all.** Once rows or columns are sorted or moved, the engine and the source data hold different row orders and copying between them would corrupt the array. The read-time projection still covers the grid itself, exactly as before — but a `columnSorting` grid whose header the user has clicked is a common state, so the changelog entry says this out loud.
- **Formulas on *other* sheets of a shared engine** that reference this sheet are still only masked on read. Writing into another instance's source data fires its hooks, validators and renders, so it is a larger change than this issue needs.
- **A column the engine knows but the grid does not** is skipped, so such a row can end up with some formulas refreshed and others not. Without a visual index the read and the write disagree about which column space the index is in — `getSourceDataAtCell` resolves it as a visual column, `setSourceDataAtCell` as a prop — so guessing would target the wrong cell. Only a userland column trimming map reaches this; no core plugin trims columns.
- **The write renders.** It goes through `setSourceDataAtCell` on purpose, so external listeners get `afterSetSourceDataAtCell` — that notification is the part that answers #10691. That API ends with `render()` and an editor refresh, and `afterCreateRow` fires inside `datamap.createRow`, before the alter's own `adjustRowsAndCols()`. So a structural change that produces a formula diff renders once more than it used to, and that render sees a pre-`adjustRowsAndCols` state. Avoiding it means either dropping the notification or deferring the write until after the alter settles — which would leave the array stale for anything reading it from `afterCreateRow`. Neither trade looked better than the extra render; worth a second opinion.

**Performance.** The scan adds one `getSheetFormulas()` call per structural change, on top of an operation that is already the expensive part. Measured over 5 runs each:

| Grid | Insert before | Insert after |
|---|---|---|
| 20k rows, 20 formula cells | ~68 ms | ~81 ms (+19%) |
| 20k rows, 1,000 formula cells | ~70 ms | ~82 ms (+17%) |
| 200k rows, no formulas | ~591 ms | ~653 ms (+10%) |
| 200k rows, 200 formula cells | ~685 ms | ~783 ms (+14%) |

Two follow-ups worth taking if this shows up in the perf suite: drive the sync from the affected span rather than the whole sheet, and cache the identity check in `isHfOrderPhysical()` (it walks the index sequence on every structural change, and the unmoved case — the common one — cannot exit early).

### How has this been tested?

Thirteen cases added to the closest existing spec, `handsontable/src/plugins/formulas/__tests__/formulas.spec.js`, under `persisting engine-rewritten formulas in the source data (#10388)`. One more case went into the Playwright suite — `tests/e2e/formulas-move-cells-undo.spec.ts`, covering the undo of a cell move, which needs a real drag. The Jasmine cases cover insert/remove row, insert/remove column, the `updateSettings({ data })` replay that reproduces the reported symptom, an array-of-objects data source (a different write path — prop rather than index), a row hidden by a filter, a single undo restoring fully and the matching redo, the `afterSetSourceDataAtCell` notification, a formula left spelled as the developer wrote it, and four guards that must hold both ways: a move, an insert after a move, and an edit undone in a moved grid all leaving the array untouched.

```bash
# The formulas suite (345 specs, was 332)

# Full core E2E - 9640 specs, 0 failures

# React wrapper - 75 tests, 0 failures (this issue is reported through it)

# Playwright formulas + moveCells specs (all six theme x bundle projects)
cd tests && npx playwright test e2e/formulas-move-cells.spec.ts e2e/move-cells-undo.spec.ts e2e/formulas-move-cells-undo.spec.ts

# Unit - 3596 tests, 0 failures

# Types and lint - clean
```

Also verified by hand in headless Chrome against the built `handsontable.full.js`, reproducing the issue's three steps before the change and confirming the formula and the value survive the round trip after it.

### Types of changes

### Related issue(s):

1. DEV-2589
2. #10388
3. #10691 (same root cause, partially addressed)

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2589

---

> [!NOTE]
> **Medium Risk**
> Touches formulas plugin data-sync on insert/remove/move and undo/redo. Guards skip write-back on sorted/moved grids, but extra `setSourceDataAtCell` renders and index-mapping edge cases could still desync source data if a guard is wrong.
> 
> **Overview**
> Fixes formulas reverting after insert/remove when the caller’s data array is loaded back in (e.g. Redux/React). HyperFormula already rewrote references; Handsontable only projected them on read, so the owned array stayed stale.
> 
> After row/column insert or remove (and `moveCells`), `#syncFormulasToSourceData` diffs `getSheetFormulas()` against stored cells and writes real changes via `setSourceDataAtCell`, fenced so they are not pushed back into the engine. Undo/redo of those structural actions is synced the same way. Writes are skipped when engine order no longer matches physical order (sort/move), when spelling is only canonicalized, and when `#REF!` would be persisted from an unrelated state. `afterSetSourceDataAtCell` still fires with a true previous value so external stores can pick up the new formula text.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ce3190ac2c30dda38cad0761c71c517b9fb220a3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2589 Formulas: rewritten formulas are not persisted to the source data (#10388)](https://app.clickup.com/t/9015210959/DEV-2589)

handsontable/src/plugins/formulas/formulas.ts
```diff
@@ -1681,6 +1875,7 @@ export class Formulas extends BasePlugin {
     const changes = this.engine!.addRows(this.sheetId,
       [this.rowAxisSyncer!.getHfIndexFromVisualIndex(visualRow), amount]);
 
+    this.#syncFormulasToSourceData();
```
Haxikowy: Formulas (priority 260) fires before nestedRows (300) rebuilds its flat-row cache in its own `afterCreateRow` (`nestedRows.ts:886`), and both `getSourceDataAtCell` and `setSourceDataAtCell` resolve through that cache via `modifyRowData`. So with nestedRows on, every rewritten formula lands in the pre-insert tree, one node off. `isHfOrderPhysical()` passes, since nestedRows keeps the sequence at identity. `afterRemoveRow` is worse, cleanup is deferred into a timeout there. No spec covers formulas + nestedRows + `alter`.

handsontable/src/plugins/formulas/formulas.ts
```diff
@@ -1537,6 +1583,11 @@ export class Formulas extends BasePlugin {
    *                          ([list of all available sources](@/guides/getting-started/events-and-hooks/events-and-hooks.md#definition-for-source-argument)).
    */
   #onAfterSetSourceDataAtCell = (changes: CellChange[], source: string) => {
+    // Checked before the blocked-source branch so the write-back never reaches undo/redo tracking.
+    if (this.#sourceDataSyncPending) {
```
Haxikowy: Returns for every change while the flag is set, not just this sync's, and the flag is held across the synchronous `runHooks`. A store listener that writes back from `afterSetSourceDataAtCell` gets its write into the array but not into HyperFormula. The write already passes a unique source on line 1855, so gate on that instead. `#onModifySourceData` at 1466 should keep the flag.

handsontable/src/plugins/formulas/formulas.ts
```diff
@@ -1665,6 +1716,149 @@ export class Formulas extends BasePlugin {
...
+    }
+
+    if (changes.length === 0) {
+      return;
+    }
+
+    this.#sourceDataSyncPending = true;
+
+    try {
+      this.hot.setSourceDataAtCell(
```
Haxikowy: `setSourceDataAtCell` ends in `render()` and `activeEditor.refreshValue()` (`core.ts:4302`), but `colHeaders` splicing, the `fixedRowsTop`/`fixedColumnsStart` adjustment and `selection.shiftRows` all happen after these hooks (`core.ts:1019-1151`). The nested render paints stale headers and pre-shift selection. Worse, `afterSetSourceDataAtCell` fires while `dataMap.createRow` is on the stack, so a listener answering with `updateSettings({ data })` rebuilds the data map mid-`alter`. Flush once after `alter` returns instead.

handsontable/src/plugins/formulas/formulas.ts
```diff
@@ -1665,6 +1716,149 @@ export class Formulas extends BasePlugin {
...
+            continue;
+          }
+
+          const columnOrProp = visualColumn === -1 ? hfColumn : visualColumn;
+
+          // `getSourceDataAtCell` takes a physical row and a visual column, `setSourceDataAtCell`
+          // a physical row and a prop.
+          const stored = this.hot.getSourceDataAtCell(physicalRow, columnOrProp);
+
+          if (stored !== formula && !this.#isSameFormula(stored, formula)) {
```
Haxikowy: Nothing clamps the loop to what the array can hold. `data = [[1, '=A1'], [2, 3, '=B1']]`, insert a row: the diff pushes `[1, 2, '=B2']`, `setAtCell` drops it (numeric column `>= countFirstRowKeys()`, which is row 0's width), but `changesForHook` was built before the write so the hook reports it applied, and it re-fires on every later structural change. A transforming `valueSetter` or a false `sourceDataValidator` loop the same way. Clamp to `countSourceRows()`/`countSourceCols()`.

Also: writes into trimmed rows hit `getCellProperties` with `toVisualRow() === null`, so per-cell `valueSetter`/`sourceDataValidator`/`cells()` meta is bypassed for rows the user cannot see.

handsontable/src/plugins/formulas/formulas.ts
```diff
@@ -1665,6 +1716,149 @@ export class Formulas extends BasePlugin {
...
+    }
+
+    const sheetId = this.engine.getSheetId(this.sheetName)!;
+    const dimensions = this.engine.getSheetDimensions(sheetId);
+
+    if (dimensions.width === 0 && dimensions.height === 0) {
+      return;
+    }
+
+    const formulas = this.engine.getSheetFormulas(sheetId) as Array<Array<string | undefined>>;
```
Haxikowy: Benchmarked on 200k x 5 against the repo's hyperformula@3.3.0: this costs 43ms with **zero** formula cells and returns `[]`, because `genericSheetGetter` walks `height x width` allocating a `simpleCellAddress` per cell. That is ~100% of the reported +40ms, plus ~20MB of transient garbage per structural change. The `width === 0 && height === 0` guard never fires on a populated sheet, so there is no exit. A `#hasAnyFormula` flag, set wherever a formula reaches the engine and never reset, would do it, and false positives only restore today's behavior.

On the two follow-ups: `isHfOrderPhysical()` is 0.16ms per axis at 200k, 0.4% of the delta, not worth caching. And narrowing to the affected span is unsound, since inserting row 5 rewrites `=SUM(A1:A10)` wherever it lives, so it would reintroduce #10388 for off-span formulas. Narrow by formula-bearing cells instead.

handsontable/src/plugins/formulas/formulas.ts
```diff
@@ -1665,6 +1716,149 @@ export class Formulas extends BasePlugin {
...
+   * @param {string} engineFormula The formula reported by the engine.
+   * @returns {boolean}
+   */
+  #isSameFormula(stored: unknown, engineFormula: string) {
+    if (!isFormula(stored)) {
+      return false;
+    }
+
+    try {
+      return this.engine!.normalizeFormula(stored as string) === engineFormula;
```
Haxikowy: 2.8us/call, ~14ms on a 5,000-formula sheet, the dominant term in the 50k/5,000 row. For a cell the engine actually rewrote, `stored !== formula` is always true, so this parses and unparses only to conclude "different"; it pays off solely on the canonicalization case. Safe to memoize: `extractTemporaryFormula` pins a fixed address, so it is a pure string to string function.

handsontable/src/plugins/formulas/indexSyncer/axisSyncer.ts
```diff
@@ -175,6 +175,28 @@ class AxisSyncer {
...
+   * Checks whether HyperFormula's index order still matches Handsontable's physical order.
+   *
+   * A move or a sort reorders the engine's rows/columns (`syncMoves` calls `engine.moveRows`), while
+   * the source data keeps its own physical order. From that point the formulas the engine holds are
+   * written in a different reference frame than the ones stored in the source data, and the two must
+   * not be copied across.
+   *
+   * @returns {boolean}
+   */
+  isHfOrderPhysical() {
```
Haxikowy: This never asks the engine anything. It asserts HOT's `getIndexesSequence()` is identity and infers HF matches. `getIndexesChangeSyncMethod` returns on `isPerformingUndoRedo()` *above* `this.#indexesSequence = newSequence`, so an undone reorder leaves a stale snapshot and the next `'update'` sync pushes a permutation built from it. If the two drift, this returns `true` on a scrambled engine and formulas get written into the wrong physical rows.

That path is unchanged from develop, where it only causes wrong displayed values. This PR makes it a permanent write into the developer's array, so it is worth hardening: move the assignment above the early return, and require both sequences to be identity.

.changelogs/10388.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed a problem where a formula updated by inserting or removing a row or column reverted to its previous state after the data was loaded back into the grid. The data is left untouched while rows or columns are sorted or moved, because the grid and the formula engine then hold different row orders.",
```
Haxikowy: Doesn't say what users observe: Handsontable now writes into the array passed as `data`, and `afterSetSourceDataAtCell` now fires on insert/remove. Also, "the data is left untouched while rows or columns are sorted or moved" reads as reassurance when it means the bug is not fixed there. And the scope sentence omits `moveCells` and undo of `move_cells`.

tests/fixtures/pages/FormulasMoveCellsPage.ts
```diff
@@ -69,6 +69,16 @@ export class FormulasMoveCellsPage {
    * (e.g. `{ trimRows: [1] }`) — a fresh Handsontable instance and a fresh HyperFormula
    * sheet, so tests stay isolated.
    */
+  /**
```
Haxikowy: This landed between `initGrid`'s doc block and `initGrid`, so that block now documents `rawData()` and `initGrid` is undocumented. Move it below.

adrianspdev:
Hey @budnix, nice work — the base insert/undo round trip works well. A few things to check though, the first two corrupt the user's `data` array:

1. **Reordered `columns` write into the wrong column.** With `data: [[1, '=SUM(A1:A3)'], [2, null], [3, null]]` and `columns: [{ data: 1 }, { data: 0 }]`, an `insert_row_above` overwrites the number `1` with the formula. Undo doesn't bring it back, and reloading the array gives `#CYCLE!`. Looks like `#getProcessedSourceDataArray` skips the column projection when `countCols() === countSourceCols()`, so the engine holds physical order while the write-back maps through `colToProp()`.

2. **A broken engine formula gets persisted.** `manualRowMove.moveRow(0, 2)` + sort asc turns `=SUM(A1:A3)` into `=SUM(#REF!)` inside the engine (pre-existing bug, the array is still fine and the sequence is back to identity, so `isHfOrderPhysical()` passes). The next insert writes `=SUM(#REF!)` into the array — the original formula is gone for good. Maybe skip writing `#REF!` formulas unless the action can actually break references (`remove_row`/`remove_col`/`move_cells`)? An insert can't produce one.

3. **Non-undoable reorder between insert and undo.** Insert (array gets `=SUM(A1:A4)`), then `manualColumnFreeze.freezeColumn(2)` (no undo entry), then undo — the guard blocks the sync, so the array keeps `=SUM(A1:A4)` on a 3-row sheet and never re-syncs. Storing the pre-sync values with the undo entry and replaying them in `afterUndo` would cover this.

I also checked object data, projected array data, `sourceDataValidator` columns, `minSpareRows`, the `normalizeFormula` comparison, and perf (~2% of an `alter` on 50k rows) — all fine.

budnix:
Good spot on both of the corrupting ones — fixed in `bd1ec071eb`.

You had the first one exactly right: a `columns` list that only reorders keeps the count equal, so the data is never projected and the engine stays on physical indexes. It now resolves the column space the same way the data was fed to the engine, and the number survives.

For the second I took your suggestion — `#REF!` is only written for `remove_row`, `remove_col` and `move_cells` now, so an insert can't overwrite a good formula any more. Both have tests that fail without the fix.

The third one I left open for now. Worth noting the engine itself already returns `=SUM(A1:A4)` on that 3-row sheet without this PR, so the grid is wrong there either way — the write-back just copies it into the array. The undo-entry snapshot would fix it, but it means a second stack running alongside undo/redo, so I would rather do it separately than bolt it on here. Happy to pick it up if you would prefer it in this PR.

adrianspdev (approved):

PR #9850, opened 2022-08-30
Fix mobile selection handlers to not extend the table's container

budnix:
### Context

The PR fixes an issue where the selection handlers could extend the table's container causing the cells to desynchronize with headers. The issue happend only when the cells on the table's edge were selected. To fix the issue, the PR implements selection handlers positioning to the center of the table if the edge cells are selected.

In the gifs below, the red circles are part of the selection handlers. That circles help the user select the cells without targeting the "true" handlers (little blue circles).

#### Before
Kapture 2022-09-05 at 14 03 16

#### After
Kapture 2022-09-05 at 14 04 54

### How has this been tested?

I tested the changes in iPhone 13 and I covered the bug with new E2E tests.

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

### Related issue(s):
1. fixes #9621

### Affected project(s):
- [x] `handsontable`

### Checklist:

- [x] I have reviewed the guidelines about [Contributing to Handsontable](https://github.com/handsontable/handsontable/blob/master/CONTRIBUTING.md) and I confirm that my code follows the code style of this project.
- [x] I have signed the [Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLScpMq4swMelvw3-onxC8Jl29m0fVp5hpf7d1yQVklqVjGjWGA/viewform?c=0&w=1)

PR #13220, opened 2026-08-21
DEV-2590: Stop the fill handle painting over the frozen panes

Haxikowy:
### Context

The PR fixes the autofill fill handle painting on top of a frozen pane, and winning hit-testing there, once the selected cell scrolls under that pane.

Repro on `develop`: a grid with `fixedColumnsStart: 2`, select C9, scroll horizontally until C9 slides behind the frozen columns. The rest of the selection border clips correctly, but the handle dot stays visible over the frozen cells — and `document.elementFromPoint` over that area returns `wtBorder current corner` instead of the frozen `TD`, so the dot steals the crosshair and the drag from those cells. The `fixedRowsBottom` mirror behaves the same way.

Root cause: `.ht_master` declares no z-index, so it opens no stacking context, and every `.wtBorder` inside it competes directly with the overlay clone divs — which take their z-index from `handsontable/src/styles/base/_z-index-map.scss` (`inline_start: 120`, `bottom: 130`, `top: 160`). The `ht_clone_master: 100` entry in that map does not apply to the master overlay: that class is stamped on the editor container by `handsontable/src/editors/factory.ts`, while the overlay div is `ht_master`.

#13076 raised `.wtBorder.corner` from `z-index: 10` to `150` so the handle would stay above the `moveCells` bands (`wtMoveZone`, inline `z-index: 100`). 150 also clears `inline_start` (120) and `bottom` (130), which is the regression.

**The z-index change.** `.wtBorder.corner` drops to `110`, keeping the constraint #13076 introduced — above the bands (100), so pressing the handle starts autofill rather than a move drag — while the frozen panes get their occlusion back.

**The resize handles move too.** `.wtSelectionHandle` (the desktop `selectionHandles` pills) sat at `200`, also from #13076, which is above every clone. A pill whose cell scrolled under a pane painted over it and took the click there — the same bug in the same family, so it is fixed here rather than left for a follow-up. They drop to `115`: above the fill handle and the bands, below `inline_start` (120). Their only frozen-pane handling today is the `display: none` that fires when an edge lands exactly on a freeze line, which does not cover the scrolled-under case; the `zIndex = '9999'` escape hatch in `border.ts` belongs to the mobile handles, not these.

Containing `.ht_master` in its own stacking context would be the structural fix, but `border.ts` deliberately bumps those mobile selection handles to `z-index: 9999` when a selection edge lands on a freeze boundary (legacy, from #9850). Containment would break that on purpose-built behavior, so the targeted z-values are the right scope here.

**The rule is now written down.** The ceiling both values obey — selection affordances inside `.ht_master` stay under 120 — is recorded in `_z-index-map.scss`, in the duplicated clone list in `walkontable.scss`, and in the walkontable `AGENTS.md`. Nothing recorded it before, which is how #13076 came to pick 150 and 200.

**The seam follow-on.** Lowering the handle exposes one case that needs geometry, not stacking. With `fixedRowsBottom` set, a selection ending on the last scrollable row puts the handle exactly on the seam — centered on the cell's bottom-end corner, so half of it hangs into the frozen strip. Below the pane's 130 that half is covered, and the dot renders cut in half with only its top half clickable. No z-value fixes this: the handle has to stay under 120 for the frozen columns, and the bottom pane is at 130.

`appear()` already solves the identical problem on the grid's own last row — it lifts the handle by half its height and adds `wtCornerBlockEndEdge`, which pulls the hit area up to match. This PR extends that branch to the seam row through the existing `isFrozenBottomBoundaryEdge()` predicate. The last row still lifts only when the handle would spill out of the trimming container; the seam always lifts, because the bottom overlay is always painted above the master there.

Measured on a 60-row grid with `fixedRowsBottom: 2`, selection ending on row 57: the handle spans 482–490 with the pane starting at 489, carries `wtCornerBlockEndEdge`, and hit tests at 25%, 50% and 75% of its height all return the handle. Before the lift, everything below its midpoint returned a frozen cell.

The column axis and the top pane need no equivalent, and a first attempt to give them one was wrong: their handles are drawn by the overlay that owns them and already land flush against the `.wtHolder` edge that clips them, which `border.spec.js` pins to the pixel on both axes. Lifting them there moved them 3px off that line and broke three walkontable specs. (Watch the measurement: `.wtHolder` is the clipping box, while the clone element is `overflow: visible` and ends a few pixels earlier, so measuring the clone reports an overhang that is not actually cut.)

### How has this been tested?

New Playwright spec `tests/e2e/fill-handle-frozen-panes.spec.ts` (7 tests):

1. pins the whole stacking window — bands < handle < resize pills, and handle < frozen columns pane;
2. scrolls the selected cell under the frozen columns and asserts the hit test at the handle's center lands on `ht_clone_inline_start`;
3. the same under the bottom frozen rows, landing on `ht_clone_bottom`;
4. pins the seam lift — the handle on the last scrollable row carries `wtCornerBlockEndEdge` and still hit-tests as itself;
5. and 6. pin the opposite for the frozen-column and frozen-top-row handles — no lift class, and no overhang past the `.wtHolder` edge that clips them;
7. pins the resize pills below the frozen columns pane.

Tests 2 and 3 assert the *occluding overlay* by name, not just "something other than the handle", so a handle that merely moved elsewhere cannot pass them as occluded.

Negative control: with the handle reverted to `150` in the served bundle, the occlusion tests fail; with the fix, they pass.

Regression runs, all green:

```bash

# Playwright — 465 passed, 0 failed (all three themes, not just e2e-main:
# the first round of this spec passed locally and failed on classic and horizon)
cd tests && npx playwright test \
  e2e/move-zone.spec.ts e2e/selection-handles.spec.ts \
  e2e/fill-handle-frozen-panes.spec.ts e2e/move-cells-api.spec.ts \
  e2e/move-cells-undo.spec.ts e2e/mobile-selection-handles.spec.ts \
  e2e/customBorders.spec.ts \
  --project=e2e-main --project=e2e-classic --project=e2e-horizon

# Walkontable — 816 specs, 0 failures
npm --prefix handsontable run test:walkontable

# Legacy Jasmine — 1445 specs, 0 failures, 1 pending
npm --prefix handsontable run test:e2e -- --testPathPattern="autofill|fillHandle|selection"
```

Lint:

```bash
cd tests && npx eslint e2e/fill-handle-frozen-panes.spec.ts fixtures/pages/SelectionFeaturesPage.ts
```

Also checked by hand in a local demo with `fixedRowsTop`, `fixedRowsBottom` and `fixedColumnsStart` set: the handle is occluded by both panes when its cell scrolls under them, sits whole above the seam when the selection ends there, and `elementFromPoint` agrees with what is on screen in every case.

No visual test covers a selection scrolled under a frozen pane, so Argos should be unchanged. A diff there would mean a stray dot disappearing.

### Types of changes

### Related issue(s):

1. DEV-2590
2. Regression from #13076

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2590

### DIFF
--- a/.changelogs/13220.json
+++ b/.changelogs/13220.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Fixed the autofill fill handle being cut in half at the bottom freeze line.",
+  "type": "fixed",
+  "issueOrPR": 13220,
+  "breaking": false,
+  "framework": "none"
+}
--- a/handsontable/src/3rdparty/walkontable/src/selection/border/border.ts
+++ b/handsontable/src/3rdparty/walkontable/src/selection/border/border.ts
@@ -1249,6 +1249,12 @@ class Border {
   /**
    * Show border around one or many cells.
    *
+   * The fill handle is normally centered on the selection's bottom-end corner, so half of it hangs
+   * past that corner. Two rows cannot afford that overhang and lift the handle by half its height
+   * instead (the `wtCornerBlockEndEdge` class pulls its hit area up to match): the grid's last row,
+   * where the overhang would spill out of the trimming container, and the `fixedRowsBottom` seam
+   * row, where the bottom overlay paints above the master and would cut the handle in half.
+   *
    * @param {Array} corners The corner coordinates.
    */
   appear(corners: number[]) {
@@ -1552,11 +1558,15 @@ class Border {
         removeClass(this.corner!, 'wtCornerInlineEndEdge');
       }
 
-      if (toRow === (this.wot.getSetting('totalRows') as number) - 1) {
+      const isBlockEndRow = toRow === (this.wot.getSetting('totalRows') as number) - 1;
+      const isBottomFreezeSeamRow = this.isFrozenBottomBoundaryEdge(toRow);
+
+      if (isBlockEndRow || isBottomFreezeSeamRow) {
         const { top: toTdOffsetTop } = fillHandleAnchor;
         const cornerHalfHeight = parseInt(String(this.cornerDefaultStyle.height), 10) / 2;
         const cornerBottomEdge = toTdOffsetTop + geometryReader.outerHeight(toTDEl) + cornerHalfHeight;
-        const cornerOverlappingContainer = cornerBottomEdge >= geometryReader.innerHeight(trimmingContainer);
+        const cornerOverlappingContainer = isBottomFreezeSeamRow ||
+          cornerBottomEdge >= geometryReader.innerHeight(trimmingContainer);
 
         if (cornerOverlappingContainer) {
           const cornerTopPosition = Math.floor(
--- a/handsontable/src/3rdparty/walkontable/src/selection/border/border.ts
+++ b/handsontable/src/3rdparty/walkontable/src/selection/border/border.ts
@@ -479,7 +479,7 @@ class Border {
    * one selection edge. Hovering a band shows a `grab` cursor; a `mousedown` on a band calls the
    * `onSelectionEdgeMouseDown` Walkontable setting so the core can initiate a move drag.
    *
-   * Bands sit at z-index 100 — below the resize pills (z-index 200) and below the autofill fill
+   * Bands sit at z-index 100 — below the resize pills (z-index 115) and below the autofill fill
    * handle (`.wtBorder.corner`, z-index 110), so both keep winning their own pixels in the corner
    * regions where the bands overlap them. All four bands are created hidden; `positionMoveZone`
    * + `appear` control their visibility.
@@ -1247,13 +1247,86 @@ class Border {
   }
 
   /**
+   * Tells whether the fill handle must be pulled back inside the selection's inline-end edge instead
+   * of straddling it. Only the grid's last column needs that, where the overhang would spill out of
+   * the trimming container and force a scrollbar. A selection ending on the last frozen-start column
+   * does not: the frozen overlay draws that handle itself and already lands it flush against its own
+   * edge, which `border.spec.js` pins to the pixel — lifting it there moves it off that line.
+   *
+   * @private
+   * @param {number} toColumn The selection's inline-end column index.
+   * @param {number} anchorInlineStart The fill handle anchor's inline-start offset.
+   * @param {HTMLElement} TD The cell that carries the fill handle.
+   * @param {HTMLElement | Window} trimmingContainer The container that clips the grid.
+   * @param {boolean} isRtl Whether the grid renders right-to-left.
+   * @returns {boolean}
+   */
+  isCornerLiftedAtInlineEnd(
+    toColumn: number,
+    anchorInlineStart: number,
+    TD: HTMLElement,
+    trimmingContainer: HTMLElement | Window,
+    isRtl: boolean,
+  ): boolean {
+    if (toColumn !== (this.wot.getSetting('totalColumns') as number) - 1) {
+      return false;
+    }
+
+    const { geometryReader } = this.wot.domBindings;
+    const cornerHalfWidth = parseInt(String(this.cornerDefaultStyle.width), 10) / 2;
+
+    if (isRtl) {
+      return anchorInlineStart - cornerHalfWidth < 0;
+    }
+
+    return anchorInlineStart + geometryReader.outerWidth(TD) + cornerHalfWidth
+      >= geometryReader.innerWidth(trimmingContainer);
+  }
+
+  /**
+   * Block-axis mirror of {@link Border#isCornerLiftedAtInlineEnd}. Two rows pull the fill handle back
+   * inside the selection's bottom edge: the grid's last row, where the overhang would spill out of
+   * the trimming container, and the last scrollable row above the `fixedRowsBottom` line, where the
+   * bottom overlay is painted above the master and would cover the overhang. The last frozen-top row
+   * is not one of them — like the frozen-start column, that handle is drawn by the overlay itself and
+   * already sits flush against its edge.
+   *
+   * @private
+   * @param {number} toRow The selection's bottom row index.
+   * @param {number} anchorTop The fill handle anchor's top offset.
+   * @param {HTMLElement} TD The cell that carries the fill handle.
+   * @param {HTMLElement | Window} trimmingContainer The container that clips the grid.
+   * @returns {boolean}
+   */
+  isCornerLiftedAtBlockEnd(
+    toRow: number,
+    anchorTop: number,
+    TD: HTMLElement,
+    trimmingContainer: HTMLElement | Window,
+  ): boolean {
+    if (this.isFrozenBottomBoundaryEdge(toRow)) {
+      return true;
+    }
+
+    if (toRow !== (this.wot.getSetting('totalRows') as number) - 1) {
+      return false;
+    }
+
+    const { geometryReader } = this.wot.domBindings;
+    const cornerHalfHeight = parseInt(String(this.cornerDefaultStyle.height), 10) / 2;
+
+    return anchorTop + geometryReader.outerHeight(TD) + cornerHalfHeight
+      >= geometryReader.innerHeight(trimmingContainer);
+  }
+
+  /**
    * Show border around one or many cells.
    *
    * The fill handle is normally centered on the selection's bottom-end corner, so half of it hangs
-   * past that corner. Two rows cannot afford that overhang and lift the handle by half its height
-   * instead (the `wtCornerBlockEndEdge` class pulls its hit area up to match): the grid's last row,
-   * where the overhang would spill out of the trimming container, and the `fixedRowsBottom` seam
-   * row, where the bottom overlay paints above the master and would cut the handle in half.
+   * past that corner. Where that overhang cannot survive — a container edge, or a frozen pane that
+   * clips or covers it — the handle is pulled back inside instead, and `wtCornerInlineEndEdge` /
+   * `wtCornerBlockEndEdge` pull its hit area along. See {@link Border#isCornerLiftedAtInlineEnd} and
+   * {@link Border#isCornerLiftedAtBlockEnd} for which edges those are.
    *
    * @param {Array} corners The corner coordinates.
    */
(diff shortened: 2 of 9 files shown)


qunabu:
Task linked: [DEV-2590 Fill handle paints over frozen columns and rows](https://app.clickup.com/t/9015210959/DEV-2590)

handsontable/src/styles/components/core/_selection.scss
```diff
@@ -212,7 +212,12 @@
         cursor: crosshair;
         // Above the moveCells bands (z-index 100, inline) — they span the SE corner, and the fill
         // handle must keep winning its own pixels there — but below the resize pills (z-index 200).
-        z-index: 150;
+        // The ceiling is tighter than 200 though: `.ht_master` sets no z-index, so it opens no
+        // stacking context and every border inside it competes directly with the overlay clones
+        // (inline_start 120, bottom 130 — see base/_z-index-map.scss). Anything at 120 or above
+        // paints the handle over a frozen pane the selection has scrolled under, and lets it win
+        // hit-testing there too.
+        z-index: 110;
```
budnix, line 223: The fill handle now sits below the panes, but the resize pills next door are still at 200 (line 46), which is above every clone (120–180). The `display: none` rules in `positionAdjustHandles` only cover edges that land exactly on a freeze line, so a handle on a column that merely scrolled under the pane still paints on top of it and still wins the click. Something like 115 would keep them above the fill handle and the move bands but under the panes — worth a quick check whether that belongs here or in a follow-up.

handsontable/src/3rdparty/walkontable/src/selection/border/border.ts
```diff
@@ -1552,11 +1558,15 @@ class Border {
...
+
+      if (isBlockEndRow || isBottomFreezeSeamRow) {
         const { top: toTdOffsetTop } = fillHandleAnchor;
         const cornerHalfHeight = parseInt(String(this.cornerDefaultStyle.height), 10) / 2;
         const cornerBottomEdge = toTdOffsetTop + geometryReader.outerHeight(toTDEl) + cornerHalfHeight;
-        const cornerOverlappingContainer = cornerBottomEdge >= geometryReader.innerHeight(trimmingContainer);
+        const cornerOverlappingContainer = isBottomFreezeSeamRow ||
+          cornerBottomEdge >= geometryReader.innerHeight(trimmingContainer);
 
         if (cornerOverlappingContainer) {
```
budnix: `removeClass(this.corner!, 'wtCornerBlockEndEdge')` only runs in the outer `else`, so when the row *is* the last row but `cornerOverlappingContainer` comes back false, a class left over from an earlier draw stays on the element. The hit area then keeps its lifted `::after` even though the handle moved back to the centered position. This shape predates the PR, but since the condition right above it got rewritten anyway, moving the `removeClass` out of the `else` would close it cheaply.

handsontable/src/3rdparty/walkontable/src/selection/border/border.ts
```diff
@@ -1552,11 +1558,15 @@ class Border {
         removeClass(this.corner!, 'wtCornerInlineEndEdge');
       }
 
-      if (toRow === (this.wot.getSetting('totalRows') as number) - 1) {
+      const isBlockEndRow = toRow === (this.wot.getSetting('totalRows') as number) - 1;
+      const isBottomFreezeSeamRow = this.isFrozenBottomBoundaryEdge(toRow);
+
+      if (isBlockEndRow || isBottomFreezeSeamRow) {
```
budnix: Curious whether the column axis was left out on purpose here. The inline-end branch just above still only fires for the grid's very last column, so a selection ending on `fixedColumnsStart - 1` and scrolled sideways gets its handle clipped at the pane edge — only half of it draws and only half is grabbable. Same story for a selection ending on the last `fixedRowsTop` row. It behaves that way today too, so it's not a regression, just the twin of what this hunk fixes.

.changelogs/13220.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Fixed the autofill fill handle painting over the frozen rows and columns after the selected cell scrolled under them, and lifted the handle above the bottom freeze line so it is no longer cut in half there.",
```
budnix: Two things in one title here. The "cut in half at the bottom freeze line" half is a real shipped bug — at tag `18.0.0-rc6` the handle was at `z-index: 10` with `ht_clone_bottom` already at 130. The "painting over the frozen rows and columns" half is not: `z-index: 150` arrived with #13076, which no tag contains, so nobody on a released version ever saw it. Since #13076's own entry lands in the same release, that half reads as a fix for a bug that never shipped.

```suggestion
  "title": "Fixed the autofill fill handle being cut in half at the bottom freeze line.",
```

budnix (commented):
The core fix looks right. The z-index band is clear between 110 and 150, `.ht_master` opens no stacking context, and the bottom overlay sits flush with the seam row at max scroll, so the lift never lands on a row far from the frozen band. Four notes below — none of them blocks the merge.

handsontable/src/styles/components/core/_selection.scss
Haxikowy: Done — 115. Same window as the fill handle: above the move bands (100) and the handle (110), below inline_start (120). Updated the three comments that still said 200.

No test needed changing: the pills that land on a freeze seam are already suppressed by the display rules, so nothing relied on them outranking a clone.

handsontable/src/3rdparty/walkontable/src/selection/border/border.ts
Haxikowy: Good catch, fixed. The inline-end branch right above had the same shape, so both are done.

Pulled the two conditions into `isCornerLiftedAtInlineEnd` / `isCornerLiftedAtBlockEnd`, which leaves one `if (lifted) { addClass } else { removeClass }` per axis — the stale-class path stops existing rather than getting a second guard.

handsontable/src/3rdparty/walkontable/src/selection/border/border.ts
Haxikowy: Left out on purpose, though I only learned why after trying it your way.

I added the lift for both — `fixedColumnsStart - 1` and the last `fixedRowsTop` row — and it broke three walkontable specs: `border.spec.js` "should properly render a selection corner on the edge of the left fixed column", its top-fixed-row twin, and the RTL mirror. All three assert the corner's trailing edge sits exactly on the frozen overlay's edge, and the lift moved it 3px off.

Those specs are right. Those handles are drawn by the overlay that owns them, and they already land flush against the box that clips them — which is the overlay's `.wtHolder` (`overflow: hidden`), not the clone element, which is `overflow: visible` and ends ~3px earlier. Measuring the clone makes it look like there is an overhang; measuring the holder shows the handle ending exactly on the clip line, and a hit test at its top and bottom edges returns the handle both times.

So the bottom seam is genuinely different: nothing clips the handle there, the bottom overlay is simply painted above the master and covers it. Reverted the two extra predicates and turned the tests I had written for them into ones that pin the flush alignment against the themed grid, so the next person gets the answer from a test instead of from a broken spec.

.changelogs/13220.json
Haxikowy: You're right, and I checked it: released `18.0.0` carries `z-index: 10`, and `git tag --contains f553f2640` comes back empty, so the over-painting half never reached anyone.

Took your suggestion, with "bottom freeze line" rather than "bottom frozen row line" for consistency with the rest of the PR:

"Fixed the autofill fill handle being cut in half at the bottom freeze line."

That is the only cut this PR fixes. I had briefly widened it to the column and top-row lines too, but per the other thread those handles were never cut, so the wider wording would have claimed a fix that does not exist.

handsontable/src/styles/components/core/_selection.scss
```diff
@@ -211,8 +214,13 @@
...
         // Above the moveCells bands (z-index 100, inline) — they span the SE corner, and the fill
-        // handle must keep winning its own pixels there — but below the resize pills (z-index 200).
-        z-index: 150;
+        // handle must keep winning its own pixels there — but below the resize pills (z-index 115).
+        // The ceiling is tighter than 200 though: `.ht_master` sets no z-index, so it opens no
+        // stacking context and every border inside it competes directly with the overlay clones
+        // (inline_start 120, bottom 130 — see base/_z-index-map.scss). Anything at 120 or above
+        // paints the handle over a frozen pane the selection has scrolled under, and lets it win
+        // hit-testing there too.
+        z-index: 110;
```
demtario, line 223: **High - the new stacking constraint has no durable home.**

`.wtSelectionHandle` at 115 and this at 110 are both governed by an invariant that now exists only in SCSS comments: *selection affordances inside `.ht_master` must stay below 120.* The root `CLAUDE.md` mandatory checklist (item 3) requires an `AGENTS.md` update for exactly this kind of constraint.

This one matters concretely. #13076 put the corner at 150 and the pills at 200 precisely because nothing recorded the ceiling. The next affordance will pick 200 again.

Suggested home: `handsontable/src/styles/base/_z-index-map.scss`, as a reserved-window entry alongside the clone values, since that is where someone looks the stack up. Note the clone values are declared twice - `_z-index-map.scss:5-14` and `handsontable/src/3rdparty/walkontable/css/walkontable.scss:194-215` - so a note in only one of them is half coverage.

handsontable/src/styles/components/core/_selection.scss
```diff
@@ -43,7 +43,10 @@
     .wtSelectionHandle {
       position: absolute;
       box-sizing: border-box;
-      z-index: 200;
+      // Above the fill handle (110) and the moveCells bands (100), below the overlay clones
+      // (inline_start 120) — `.ht_master` opens no stacking context, so a pill above the clones
+      // paints over a frozen pane the selection has scrolled under and takes the click there.
+      z-index: 115;
```
demtario, line 49: **Medium - this change is not in the PR description.**

The description states the pills stay at 200 ("below the resize pills (z-index 200)") and frames the PR as a single change to `.wtBorder.corner`. This line lowers a different feature's stacking by 85, and it is the one change a reviewer or QA cannot see from the write-up.

The change itself checks out, for the record. Blame shows `z-index: 200` came from f553f2640 (#13076) - the same PR that raised the corner to 150, so it was arbitrary rather than a deliberate long-standing choice. And the desktop pills (`adjustHandles`) have no equivalent of the `zIndex = '9999'` escape hatch at `border.ts:684`; that one belongs to the mobile `selectionHandles`. Their only frozen-pane handling is the `display: none` at `border.ts:1659+`, which fires only when an edge lands exactly on the freeze line - so a pill scrolled under a pane was stealing the click exactly like the corner. Same bug, same family, correct to fix here. It just needs to be in the description, and in the changelog (see the other comment).

Also worth syncing: the description says 4 tests, the spec ships 7.

.changelogs/13220.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Fixed the autofill fill handle being cut in half at the bottom freeze line.",
```
demtario, line 3: **Medium - the entry describes the follow-on, not the fix.**

"Cut in half at the bottom freeze line" is the seam symptom that lowering the handle exposed. The primary fix - the handle painting over a frozen pane and winning hit-testing there, which is the reported bug - is absent, and so is the resize-pill stacking change. A user who hit the original bug will not find it here.

Suggested: "Fixed the autofill fill handle and the selection resize handles painting over the frozen panes and taking the click there." with the seam lift as a second entry or a trailing clause.

demtario (changes requested):
Reviewed at high effort in an isolated worktree, all three dimensions (architecture, code quality, performance + a11y).

**The fix is correct and well targeted.** Verified independently rather than taken from the description:

- Root cause holds. `domScaffold.ts:153` stamps `ht_master`; the `ht_clone_master: 100` entry lands only on the editor container (`editors/factory.ts:242`), so the master really opens no stacking context.
- Both z-values are safe on a full sweep of *both* stylesheets - `handsontable/src/styles/` and `handsontable/src/3rdparty/walkontable/css/walkontable.scss`. Nothing inside `.ht_master` sits in 110-130.
- The extraction into `isCornerLiftedAtInlineEnd` / `isCornerLiftedAtBlockEnd` is faithful to the old RTL, LTR, and block thresholds.
- Clamping is not a hazard: `toRow` reaching the block-end predicate is clamped, but the corner only draws when it equals the area selection's real bottom-end corner (`isSouthEastOfAreaSelection`, `border.ts:589`).
- The stale-class fix is a genuine second bug - the old `removeClass` sat in the `else` of the last-column/last-row guard, so `wtCornerInlineEndEdge` persisted across draws.

Test quality is strong: tests 2 and 3 assert the occluding overlay *by name*, so a handle that merely moved cannot pass as occluded; `frozenFillHandleOverflow` measures against `.wtHolder` rather than the `overflow: visible` clone; `scrollCellBehindFrozenPane` stops a fixed inset past the pane edge instead of at its middle, which is what makes it theme-independent.

Performance and accessibility were both reviewed and are clean. Two extra `getSetting` calls per `appear()` per overlay is negligible, and the seam early-return actually removes a forced layout on that path. The affordances are mouse-only; no ARIA, role, keyboard path, or focus behavior is touched.

Requesting changes on three items, inline: one High (the undocumented stacking constraint - a mandatory-checklist item) and two Medium (PR description and changelog both missing the resize-pill change and the primary fix). Four Low findings are not posted - the 1px `Math.ceil` shift for odd `--ht-cell-autofill-size`, merged cells missing the seam lift, the block-end lift firing in `ht_clone_top` in a degenerate config, and no test on the stale-class clear. Happy to add them if you want them tracked.

One process note: `Visual / cross-browser` and `Visual / multi-framework` were still pending when I reviewed. This change moves pixels by design, so those two are the signal I would want green before merge.

handsontable/src/styles/components/core/_selection.scss
Haxikowy: Agreed, and it is the reason this bug exists at all — nothing recorded the ceiling, so #13076 picked 150 and 200 and no review caught it.

Written down in three places, since as you note one is not enough:

- `src/styles/base/_z-index-map.scss` — the 101–119 reserved window, what lives in it, why 120 is the wall, and a pointer to the duplicate clone list.
- `src/3rdparty/walkontable/css/walkontable.scss` — a short note above the clone declarations pointing back at the map.
- `src/3rdparty/walkontable/AGENTS.md` — a full section, including the two mechanisms that look like counter-examples (the mobile handles' inline `9999`, and the fill handle being repositioned rather than re-layered at the bottom seam) so the next person does not read them as precedent for going above 120.

handsontable/src/styles/components/core/_selection.scss
Haxikowy: Fair — the description described the PR as it stood before the pills moved, and I did not go back to it. Rewritten: the pill change has its own paragraph (what moved, why it belongs here rather than in a follow-up, and that the `9999` escape hatch is the mobile handles' rather than these), the test count says 7 with all seven listed, and the regression commands now show the three-theme run instead of the `e2e-main`-only one.

Thanks for confirming the blame independently. One correction to draw out of it: because `z-index: 200` and the pills themselves both arrived in #13076, which no tag contains, this stacking fix is invisible to anyone on a released version — `.wtSelectionHandle` does not exist in `18.0.0` at all. That is why it does not get a changelog line of its own; see the other thread.

.changelogs/13220.json
Haxikowy: This one I want to push back on, because it collides with the thread above and the release history decides it.

The over-painting is real, and it is the reported bug — but it is not reachable from any released version. Released `18.0.0` has the fill handle at `z-index: 10`, below `ht_clone_bottom` (130) and `ht_clone_inline_start` (120), so it never painted over a pane. It reached 150 in #13076, and `git tag --contains f553f2640` is empty. Same for the resize pills: `.wtSelectionHandle` does not exist in `18.0.0`, so their 200 never shipped either.

Since #13076 releases alongside this PR, the net user-visible delta is exactly one thing: the handle is no longer cut in half at the bottom freeze line, which *was* broken in `18.0.0` (10 < 130). Listing the over-painting would describe a bug no user could have hit, and listing the pill stacking would describe a fix to a feature that has not shipped yet.

So the entry stays narrow:

"Fixed the autofill fill handle being cut in half at the bottom freeze line."

The full story — over-painting, hit-test theft, the pill change — is in the PR description, which is where a reviewer looks. If you would rather the changelog track per-PR work than per-release delta, say so and I will widen it, but that would be a change in convention rather than a fix to this entry.

demtario (approved):

budnix (approved):

Issue #12632, opened 2026-05-22
Row mismatch when using the suggested solution to wrap heaader text


windmichael:
### Describe the bug

When using the suggested solution to wrap header text and using fixed columns, there will be a misalignment of rows. (https://forum.handsontable.com/t/header-column-wrap-issue/5192)
To wrap the headers "white-space: normal;" in css is used. 

The offset/misalignment gets even bigger when using the zoom out function of the browser: 

### Video/Screenshots

Screenshot of the misaligned rows:
<img width="249" height="255" alt="Image" src="https://github.com/user-attachments/assets/a31896c6-9199-42c7-ae80-1b2c353982ef" />

Zoom to 50%:

<img width="182" height="164" alt="Image" src="https://github.com/user-attachments/assets/c52a93e6-aab3-4946-95ef-ddc2f8e739e6" />

### Provide a link to the demo with the bug reproduction

https://stackblitz.com/edit/m2vpwnw6

### Handsontable version

17.1.0

### Your environment

Chrome

sl01k:
Hi @windmichael,
Thanks for the report. We tried reproducing this on the provided StackBlitz demo but couldn't see the misalignment on our end.

Could you share:
- exact Chrome version (chrome://version)
- OS and version

That should help us narrow it down.

windmichael:
OS: Windows 11 25H2
Chrome: 148.0.7778.178

I forgot to mention to scroll down to the last row to reproduce this issue.

The small misalignment of about 1px you will always see when scrolling to the last row. 
The higher misalignment only occurs when zooming using the browser zoom function while being at the last row.

windmichael:
@sl01k can you please check, if you are able to reproduce the issue?

budnix:
Thanks for reporting this issue. I was able to reproduce the problem on my side. I've already prepared a fix (https://github.com/handsontable/handsontable/pull/12741), and it's currently waiting for review. Once the review is completed, the fix will be included in the next Handsontable release. Thanks again for bringing this to our attention.

Demo that includes the fix https://stackblitz.com/edit/m2vpwnw6-w833qxdd?file=index.js,package.json

PR #13221, opened 2026-08-21
DEV-2591: Keep the grid scrollbars reachable under frozen overlays

### LINKED ISSUE #10370: Scroll bar in FireFox when using fixedColumnsLeft
### Description
When using FixedColumnsLeft setting, scrollbar which overlaps the last row in FireFox hides behind the fixed column.

I have used below CSS to add padding-bottom so that scrollbar is visible. but It does not get selected in firefox if it is on FixedColumns side I guess thats because of z-index of fixedColumns...

    @supports (-moz-appearance: none) {
           div.ht_master.handsontable.innerBorderLeft > div.wtHolder {
                padding-bottom: 15px !important;
          }
      }

### Steps to reproduce
1. open link in firefox and scroll 

### Demo
Here is demo link [Sandbox](https://codesandbox.io/s/strange-kare-gpuubq?file=/src/App.tsx)
https://codesandbox.io/s/strange-kare-gpuubq?file=/src/App.tsx

### Your environment
* React wrapper version: 11.1.0

budnix:
### Context

The PR fixes [#10370](https://github.com/handsontable/handsontable/issues/10370) — open since May 2023, re-confirmed by the reporter in September 2024, and still reproducible on `develop`.

A frozen overlay is painted on top of the grid's scrollbar, which hides it and swallows the press. The reported case is frozen columns and the horizontal scrollbar: with the grid scrolled fully left, the thumb sits entirely under the frozen columns, so a mouse user cannot drag it at all.

**The precondition is a browser that paints the scrollbar over the content instead of reserving space for it** — a floating thumb of zero layout width. That is not one engine. It is Firefox everywhere, Chrome and Safari on macOS while "Show scroll bars" is set to "Automatically", and Chrome on Windows 11 and on GTK. An earlier revision of this branch gated the fix on `isFirefox()`; that was wrong, and the gate is now a measured zero. It is also why the 2023 investigation closed as "cannot reproduce": every attempt used a classic-scrollbar environment, where Handsontable is correct in every version.

Cause, in order:

1. A floating scrollbar occupies no space, so the browser never shrinks the master holder and the measured scrollbar width is 0.
2. Each overlay therefore reserves nothing and is sized to the full holder.
3. The browser still paints the scrollbar inside the content area.
4. The overlay covers it, hiding the scrollbar and swallowing the press.

Step 4 is the part that is not engine-specific. A browser paints a scroll container's scrollbar above that container's own contents — but the overlay clones are *siblings* of the master, not its contents, so they are painted above it in every engine.

#### Scope: every overlay that touches an edge a scrollbar is drawn on

Measured by dragging in the strip and reading `scrollLeft` / `scrollTop`; 0 means the press never reached the scrollbar.

| Overlay | Edge / scrollbar | Before | After |
|---|---|---|---|
| `inlineStartOverlay` (frozen columns) | bottom / horizontal | 0px | 499px |
| `bottomOverlay` (`fixedRowsBottom: 3`) | bottom / horizontal, **full width** | 0px | 1721px |
| `topOverlay` (`fixedRowsTop: 4`) | inline-end / vertical | 0px | 1575px |
| `topOverlay` with **no frozen rows** — the column header alone is a 29px overlay | inline-end / vertical | 0px | 1068px |
| `bottomInlineStartCornerOverlay` | bottom / horizontal | covered | clear |

`bottomOverlay` spans the entire width, so with any `fixedRowsBottom` the horizontal scrollbar is completely unreachable. The `topOverlay` case needs no frozen rows at all: a plain column header blocks the top of the vertical scrollbar. `topInlineStartCornerOverlay` needs no clearance — it sits at the inline-start/top, away from both scrollbars.

#### What changed

**One track per grid, not one per overlay.** A single band per axis is drawn in a zero-size `position: sticky` host inside the master holder, spanning the whole scrollport. Covering only the frozen overlays' own strips left an opaque patch beside a transparent gap over the master. The host must be the holder's first child (a sticky box only shifts toward its own edge), and the band must live inside the holder, because one stacked above it would hide the very thumb it sits behind.

**The overlays are clipped, not resized.** Each gets a `clip-path: inset()` over the strip its own edge crosses. Every box the viewport calculations measure keeps its size, and a clipped region is neither painted nor hit-tested, so the press falls through to the scrollbar. Resizing the roots or holders instead — the earlier approach on this branch — rippled into the viewport calculations and cost 19–24 walkontable failures; the clip costs none. Only overlays that actually reach the scrollport edge are clipped: `inset()` is relative to each element's own border box, and the inline-start overlay's right edge sits mid-grid.

**Per axis.** The band follows whichever scroll offset actually moved, because a browser only draws the scrollbar for the axis being scrolled. A hover may hold a band open but never open one.

**The track does not animate, deliberately.** It is drawn under the clones, so exactly two states are clean: fully opaque with the clip closed, or absent with the clip open. Any partial opacity is a defect — with the clip open the fade shows over the master's segment alone and the track reads differently over each overlay; with it closed the strip shows the master's cell where frozen content belongs and a column header looks cut short. Both variants were built and rejected from frame-by-frame measurements.

**Presses in the band are swallowed by coordinate**, in the capture phase, so a click on the empty track does not move the selection. The band is not hit-testable, so a listener on it would never fire. The wheel is excluded, so scrolling over the track keeps working.

#### Two places where the measurement itself was wrong

Both come from the same root: `getScrollbarWidth()` measures a bare off-DOM probe, which answers "does this engine give scrollbars layout space", not "does *this* scroller". Where the two disagree — measured on Firefox 154, probe 0 against a real 15px holder gutter — everything downstream is wrong.

1. **Two scrollbars in one grid.** The clearance turned on because the probe said 0, so the strip was drawn beside a scrollbar that already had a gutter. The gate now also requires the holder itself to reserve nothing (`reservedScrollbarSpace`, per axis). It is an AND, so the new condition can only ever switch the clearance off, never on.
2. **The column header ignoring the scrollbar.** `topOverlay` and `bottomOverlay` subtracted the probed width from the workspace width, so with the probe at 0 they came out 15px too wide and ran underneath the scrollbar — a mismatched top-right corner, while the frozen columns on the left stayed correct. They now use the holder's own `clientWidth`, the same correction `inlineStartOverlay` already applies on the other axis for #12632.

#### A known cost, decided deliberately

A bottom-pinned overlay is held in place by `style.bottom`, so clipping its last pixels does not move it
out of the scrollbar's way — it removes that much of its content. With `fixedRowsBottom` and the
horizontal track on screen, the last frozen row shows **13 of 29px (45%)** until the track fades.

Lifting the overlay instead keeps the row whole, and was built and compared side by side. It was
**rejected**: the lift opens a 16px gap between the frozen rows and the holder's bottom edge that only
the band fills, which makes the band load-bearing rather than cosmetic — if it ever failed to draw, the
gap would show scrolled master content instead of a clean edge. Clipping has no such dependency.

This is a decision, not an oversight.

- Unit tests added/modified (`*.unit.ts`): `handsontable/src/3rdparty/walkontable/test/unit/overlay/scrollbarClearance.unit.ts` (extended) and `handsontable/src/3rdparty/walkontable/test/unit/overlay/scrollbarVisibility.unit.ts` (new) — 84 tests covering the clearance decision, the per-axis gutter measurement, the pointer-capability gate, the overlay width choice, the deferred layout reads, the `clip-path` geometry including RTL, the band DOM lifecycle including nested grids, and the visibility state machine.
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`): `tests/e2e/overlay-scrollbar-clearance.spec.ts`. Whether a run gets floating or space-taking scrollbars is an OS/browser display setting a page cannot request, so the spec reads the regime at runtime and asserts the matching contract. Measured, not assumed: headless Chromium hides its scrollbars, so **CI takes the floating branch** and covers the fix itself — the band is drawn and the covering overlay clipped out of it. A headed run on a classic-scrollbar desktop takes the other branch, which guards the doubled scrollbar. Both branches were checked by mutation: setting `OVERLAY_SCROLLBAR_CLEARANCE` to 0 fails the spec, and removing the regime checks reproduces two bands beside a real 15px gutter on a headed classic desktop.
- Type tests (`*.types.ts`) updated if public API changed: **none** — no public API change. Everything added is internal to Walkontable's overlay module.
- For a bug fix — the spec that fails without this fix: `overlayScrollbarClearance › should reserve nothing when the holder itself reserves space, whatever the probe says` and the `reservedScrollbarSpace` suite both fail without the gate change. The clearance geometry suites fail without the clearance module.
- Demo page / recorded trace (for UI changes): reproduced and verified on a local page driven over the DevTools protocol against real Firefox 154 and Chrome. Measured as the overlay's right edge relative to the holder's inner edge: **15px past it before, 0 after**. Composited-pixel sampling down the strip in Chrome: one colour at every sample, and **0 frames out of 161** showing either failure mode.

### Commands run

```
(no output — clean)

Test Suites: 309 passed, 309 total
Tests:       3732 passed, 3732 total

npm --prefix tests run test -- e2e/overlay-scrollbar-clearance.spec.ts
2 passed

npm run test:e2e --prefix handsontable   (legacy Jasmine/Puppeteer)
9658 specs, 0 failures, 8 pending specs

816 specs, 0 failures

✖ 601 problems (0 errors, 601 warnings)

✓ stylelint --cache "src/**/*.{css,scss}" "test/**/*.{css,scss}"

Test Suites: 15 passed   Tests: 75 passed

Test Suites: 4 passed    Tests: 33 passed

Test Suites: 13 passed   Tests: 291 passed
```

The wrapper suites are included because this change adds listeners and teardown to the overlay
lifecycle, and the observer stubs those suites rely on are duplicated per package — a method one of them
lacks fails that whole suite independently of core.

A note on the walkontable suite, because the number moved during development. It is green with the OS scrollbar setting on "Always", which is the space-taking mode the legacy suite was written against and the one CI runs. Earlier runs of this branch were made with the setting on "Automatically" so the fix would be active, and reported 48–61 failures; most of those come from the display mode itself rather than from any change here (a control run with the clearance disabled still failed 48). Treat a failure count taken under floating scrollbars as a property of the machine, not of the branch — and note that in that mode the suite exercises a code path it makes no assertions about.

### Types of changes

### Related issue(s):

1. #10370
2. DEV-2591
3. #12632 (the same probe-versus-holder correction, previously applied to the other axis)

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

- [x] This change needs a manual QA pass — the fix only activates when the browser draws floating scrollbars, which cannot be turned on from a page and so cannot be set up in CI. To check it: set macOS "Show scroll bars" to "Automatically" (or use Firefox on any OS, or Windows 11 Chrome), open a grid with `fixedColumnsStart`, `fixedRowsTop` and `fixedRowsBottom`, and confirm that (1) both scrollbars can be grabbed and dragged where a frozen overlay overlaps them, (2) the track appears only along the axis being scrolled and leaves no gap once it goes, (3) clicking the empty track does not move the selection while the wheel still scrolls, and (4) the column header stops at the scrollbar rather than running underneath it. Worth repeating with the classic space-taking setting, where the whole feature must stay switched off.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2591


qunabu:
Task linked: [DEV-2591 Firefox overlay scrollbars: horizontal scrollbar unreachable under frozen columns (#10370)](https://app.clickup.com/t/9015210959/DEV-2591)

PR #13222, opened 2026-08-21
DEV-2592: Guard fork-hostile CI steps so CI Gate can pass on fork PRs

demtario:
### Context

The PR fixes `CI Gate` being unsatisfiable for pull requests that run on a read-only token, which makes external contributions structurally unmergeable regardless of code quality.

A `pull_request` event from a fork gets a read-only `GITHUB_TOKEN` and no Actions secrets. Several jobs attempt a PR comment, a `gh-pages` push, or a secret-backed scan unconditionally, so they 403 or exit non-zero, and `CI Gate` (the single required check, which passes only on `success|skipped`) aggregates those into a blocked PR.

Measured on [run 31252773025](https://github.com/handsontable/handsontable/actions/runs/31252773025) (PR #13073) and its fuller sibling [run 32481421878](https://github.com/handsontable/handsontable/actions/runs/32481421878) (PR #13207, which reached the whole graph):

| job | failing step | cause |
| --- | --- | --- |
| `Performance / run` | `Deploy PR report to GitHub Pages` | `git push origin gh-pages` needs `contents: write` |
| `Performance / run` | `Post performance summary to pull request` | sticky comment needs `pull-requests: write` |
| `Integration / preview packages` | `Post PR preview comment` | same. `pkg-pr-new publish` itself succeeded |
| `Code Quality / SonarCloud` | scanner | `SONAR_TOKEN` empty, exit 3 |
| `Code Quality / FOSSA` | action | `FOSSA_API_KEY` empty |
| `Docs / preview / deploy` | `Publish sticky comment in PR` | dies at step one, ~14s; every build and deploy step after it already skipped |
| `CI Gate` | aggregate | any non-`success`/`skipped` upstream result |

Everything that exercises the code was green. PR #13073 sat five weeks behind a red gate the contributor could not influence.

#### The fix

One canonical guard, at seven sites:

```yaml
if: github.event_name != 'pull_request'
  || (github.event.pull_request.head.repo.full_name == github.repository
      && github.actor != 'dependabot[bot]')
```

Every clause earns its place:

- `github.event_name != 'pull_request'` is **mandatory**. `publish.yml` calls `test.yml` on the RC path, where `github.event.pull_request` is null; without it a bare comparison reads as "fork" and silently disables the step on every release run.
- `github.actor != 'dependabot[bot]'` closes the case a fork-only check misses. A Dependabot PR's branch is **same-repo**, so the repo comparison passes while the token is downgraded exactly like a fork's. Verified on [run 30857196770](https://github.com/handsontable/handsontable/actions/runs/30857196770): an all-read grant (`Contents: read`, `Checks: read`). `github.actor` rather than `pull_request.user.login`, so a human pushing to a Dependabot branch re-enables the step along with the token.

Guards are **step-level** where the job still does useful work, **job-level** only where the job is entirely secret-dependent:

| file | guarded | level |
| --- | --- | --- |
| `.github/actions/performance-run/action.yml` | gh-pages deploy, sticky comment, `Compute report URL` | step |
| `.github/workflows/performance-tests.yml` | gh-pages deploy, sticky comment | step |
| `.github/workflows/integration.yml` | `Post PR preview comment` | step |
| `.github/workflows/code-quality.yml` | `sonarcloud`, `fossa` | job |
| `.github/workflows/docs.yml` | `preview` (the `docs-staging.yml` call) | job |

Fork PRs still compute performance numbers, publish preview packages, and render the visual matrix. Content that can no longer be a PR comment goes to `$GITHUB_STEP_SUMMARY`, which needs no token.

#### What review changed, and one thing it caught

Three reviewers went over this. The most important finding was a **regression I had introduced**: I originally guarded `visual.yml`'s Argos upload on the reasoning that it names `ARGOS_TOKEN`. That reasoning is wrong, and the guard is reverted.

`@argos-ci/core` falls back to **tokenless** upload on public GitHub Actions repos. On fork run `32481421878` the step ran with `ARGOS_TOKEN:` empty and **passed**, and the Argos App reported [build 5874](https://app.argos-ci.com/handsontable/handsontable/builds/5874) with "Everything's good!". `visual-tests/scripts/upload.mjs` has no `try`/`catch`, so a failed upload would have thrown. Guarding it would have deleted visual review for every external contributor, which the "functional continuity" rule in `CLAUDE.md` forbids.

The root cause was the rule I wrote in `AGENTS.md`: "any step that reads a secret must be guarded, or it 403s". That is false, since an absent secret is an empty string, not an error. It is rewritten to key on "writes through the API, pushes to a repository ref, or genuinely *fails* without a secret", with the tokenless-Argos and `check.mjs` graceful-degradation counterexamples named and an explicit instruction to verify against a real fork run before adding a guard.

Also from review:

- `github.head_ref` was interpolated directly into a `run:` script in both perf paths. Pre-existing on develop, but on a fork PR the branch name is attacker-controlled, so this is a command-injection vector in the code path this PR is about. Now passed through `env:`. Verified blocked: a branch named `evil$(touch /tmp/PWNED)-branch` is sanitized to an inert slug and no command executes.
- `Compute report URL` was unguarded, so a fork's job summary advertised a `gh-pages` URL that 404s now the deploy is skipped. Guarded, and `report-builder.mjs:35` only emits the link when the URL is set.
- `performance-tests.yml`'s "unreachable" premise was wrong. `develop.yml` has `pull_request: paths: ['.github/workflows/develop.yml']` and its perf stage carries no event check, so its two write steps are reachable and are now guarded.
- `always()` on the preview body/summary meant a failed publish still printed four `npm i` commands for packages that were never produced. Dropped there, kept on the perf summary where the numbers are real.

#### Notes for the reviewer

- **No `permissions:` block is modified and no scope added**, so the `publish.yml` RC ceiling and the nested-`permissions:` hazard in `AGENTS.md` are untouched. The fork failure is a runtime 403 after GitHub silently downgrades the token, not startup validation: on fork run `32481421878` the run started and `Code Quality / CodeQL` passed while three jobs requested write scopes.
- **`pull_request_target` is deliberately not used**, and a test now asserts no workflow uses it. It would run fork-controlled code that builds and publishes preview packages with a write token.
- **The nightly scanners are unaffected.** `code-quality-baseline.yml` is a separate file; only the PR-path `sonarcloud`/`fossa` jobs are guarded.
- **Internal PRs now get a job summary in addition to the existing sticky comments.** Intended, not a regression.
- **`Docs / angular-types` and `Code Quality / CodeQL` are deliberately left unguarded** because they degrade gracefully (`docs/angular-type-check/check.mjs:475-480` turns a Checks-API 403 into a `::warning`; the CodeQL action skips the SARIF upload and was green on the fork run).
- **The docs guard is behavior-neutral, not a loss of the docs build.** On the fork run the job already died at its first step with every build and deploy step skipped. Restoring build verification for fork PRs needs per-step guards inside `docs-staging.yml` (two sticky comments, three `github-script` steps, shared with the trunk and RC paths) and is a separate task; the comment in `docs.yml` says so.

### How has this been tested?

`CI Gate` is **green** on [run 32485496480](https://github.com/handsontable/handsontable/actions/runs/32485496480): 41 jobs succeeded, 5 scope-skipped, no failures or cancellations.

That run is the meaningful one because a temporary probe commit (since reverted) widened `checks.yml`'s scope router so all guarded modules actually executed. Without it a `.github/**`-only diff scope-skips `performance`, `integration`, `visual`, and `docs`, and a green gate would prove nothing.

Internal-PR non-regression, asserted per site rather than inferred from the gate colour:

- `Code Quality / SonarCloud` and `FOSSA`: **success**, so the job-level guards do not misfire.
- `Docs / preview / deploy`: **success**, so the parenthesised clause folds correctly.
- `Performance / run`: **success**, sticky comment posted **and** `performance-reports/<slug>/index.html` confirmed present on `gh-pages`.
- `Integration / preview packages`: **success**. The comment step was refactored from an inline `message:` to a generated file so the sticky comment and job summary share one source, which meant re-rendering markdown containing a fenced block and backticks. I diffed the body CI actually posted against a local re-run of the real heredoc: **byte-identical**.
- `Visual / Argos`: guard reverted, so unchanged from today.

Static and mechanical checks:

- `actionlint` clean on every changed workflow.
- Parsed each changed file and dumped the resolved `if:` values, confirming every multi-line plain scalar folds to the intended single-line expression. This caught a real bug: one edit landed continuation lines at the same indent as the `if:` key, which breaks the scalar.
- `npm run test:tooling`: 98 tests pass.

#### New test

`.github/scripts/__tests__/fork-guards.test.mjs`, in the existing `Checks / tooling tests` harness. The guard is duplicated across seven sites in five files, so the real failure mode is drift: a new fork-hostile step added unguarded, or a half-guard copied in. Behaviour cannot be tested from CI on an internal branch, where every guard's same-repo disjunct is trivially true, so it asserts the **shape**, the way `esTarget.unit.js` asserts the three-place ES floor. It also asserts the site list in `AGENTS.md` matches the workflows, and **locks the Argos exception** so the tokenless regression cannot be reintroduced.

It caught two defects on its first run, both fixed here: `integration.yml` carried no event half, and the `AGENTS.md` site list had already drifted by omitting `docs.yml`. The guard is now uniform across all seven sites. Mutation-tested both directions: stripping the Dependabot half fails with a precise per-site message, and re-adding an Argos guard trips the lock.

#### Completeness sweep

Every secret and write-token site reachable from `test.yml` on a `pull_request` is accounted for: `SONAR_TOKEN` and `FOSSA_API_KEY` guarded; `ARGOS_TOKEN` deliberately unguarded and locked; `PASS_COOKIE` covered transitively (its job requires `preview.result == 'success'`); `docs-staging.yml`'s Cloudflare secrets guarded at the caller; `checks.yml`'s `GITHUB_TOKEN` read-only and green on the fork run; five sticky comments and four `git push origin gh-pages` sites all guarded or `push`-only; no `github.rest.*` write calls exist in the graph.

#### Not verified, and honestly so

The **fork path itself** cannot be exercised from this repository. The evidence above is historical fork runs plus internal-PR non-regression. After merge, confirm on a genuine fork PR that every guarded step reports `skipped` while `CI Gate` goes green, then re-run #13073.

One item needs an account with admin: `gh api repos/handsontable/handsontable/branches/develop/protection --jq '.required_status_checks.contexts'` returns 404 for me. The two job-level guards assume only `CI Gate` is required, consistent with `Integration / CJS`, `ESM`, `Docs`, `Visual`, and `Walkontable` routinely reporting `skipped` on merging internal PRs. If an individual name turns out to be pinned, those guards must become step-level, because a *required* check reporting `skipped` blocks a PR indefinitely.

#### Known follow-up, not addressed here

A withheld gate is indistinguishable from a scope-skipped one: an external PR adding a non-permissive dependency now shows `FOSSA: skipped` and a green `CI Gate`, with nothing on the PR saying the licence gate never ran. That is a real gap in the dependency-licence policy on exactly the PRs from untrusted contributors. Fixing it needs a new always-running job in `ci-gate`'s `needs` that reports which gates were withheld, which is a graph change rather than a guard, so it belongs in its own task.

### Types of changes

### Related issue(s):

1. DEV-2592
2. #13073 and #13207 are the live fork cases this was diagnosed from, and #13073 is the one to re-run once this lands.

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

CI configuration, one CI test, and docs only. No package source is changed: the net diff is 9 files under `.github/`, `.ai/`, and `AGENTS.md`.

### Checklist:

- [x] My change requires a change to the documentation. Done here: two "fork and Dependabot pull requests" bullets in the root `AGENTS.md` carrying the canonical guard, why each half is load-bearing, the guarded-site list, and the do-not-guard-graceful-paths rule; plus a refresh of the stale CI workflow list in `.ai/STACK.md` (`linter.yml`, `changelog.yml`, `audit.yml`, and `pkg-pr-new.yml` no longer exist).

[skip changelog] — CI configuration, one CI test, and docs only.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2592

---

> [!NOTE]
> **Medium Risk**
> CI-only change that alters when comments, gh-pages deploys, and secret-backed scans run. Wrong guards could skip those on internal/RC runs or still fail CI Gate on forks; no product/auth code is touched.
> 
> **Overview**
> Makes **fork and Dependabot PRs mergeable** by skipping write-token and secret-hard steps that 403 on a read-only `GITHUB_TOKEN` and fail `CI Gate` (which only accepts `success`/`skipped`).
> 
> Adds one canonical `if:` (same-repo **and** not `dependabot[bot]`, plus an event-name half so RC/`publish.yml` still runs) at seven sites: perf gh-pages deploy, sticky comments, and report URL; integration preview comment; SonarCloud/FOSSA jobs; docs preview. Forks still run tests, preview publish, and Argos (tokenless — deliberately **not** guarded). Withheld comments go to `$GITHUB_STEP_SUMMARY`.
> 
> Also passes `github.head_ref` via `env` to avoid command injection, and adds `fork-guards.test.mjs` to lock guard shape, the AGENTS.md site list, the Argos exception, and a ban on `pull_request_target`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 386a2a43e2177ad199b9021f89d14c7edecea923. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2592 CI: every fork pull request is permanently red — unguarded write steps make CI Gate unsatisfiable for external contributors](https://app.clickup.com/t/9015210959/DEV-2592)

AGENTS.md
```diff
@@ -207,6 +207,14 @@ Follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) format. Changel
...
...enless** on public repos (verified on fork PR #13073's sibling run `32481421878`, which produced Argos build 5874), and `docs/angular-type-check/check.mjs` turns a Checks-API 403 into a `::warning`. Guarding either would delete working coverage for external contributors, which the "functional continuity" rule above forbids. Confirm against a real fork or Dependabot run before adding a guard. Current guarded sites: `.github/actions/performance-run/action.yml` and `performance-tests.yml` (gh-pages push, sticky comment), `integration.yml`'s `preview-packages` (sticky comment), `code-quality.yml`'s `sonarcloud` and `fossa` (hard secret dependency), `docs.yml`'s `preview` (the `docs-staging.yml` call, whose first step is a sticky comment). `.github/scripts/__tests__/fork-guards.test.mjs` asserts this list against the workflows, so add new sites in both places. Prefer **step-level** guards where the job still does useful work; job-level only where the job is entirely secret-dependent. Withheld content goes to `$GITHUB_STEP_SUMMARY`, which needs no token. Never reach for `pull_request_target`: it would run fork-controlled code that builds and publishes preview packages with a write token.
```
budnix: Run `32481421878` belongs to PR #13207, not #13073 — #13073 is a different fork PR. The three other places citing the same run (`visual.yml:160`, `docs.yml:87`, `fork-guards.test.mjs:125`) all say #13207, so this one is the odd one out. Worth fixing, since this number is the evidence someone checks before deciding to leave the Argos step unguarded — land on the wrong PR, find no Argos build, and the safe-looking move is to guard it.

.github/workflows/performance-tests.yml
```diff
@@ -166,6 +169,8 @@ jobs:
 
       - name: Deploy PR report to GitHub Pages
         if: always() && github.event_name == 'pull_request'
+          && github.event.pull_request.head.repo.full_name == github.repository
```
budnix, line 185: This file is `workflow_call`-only and its one caller is `develop.yml` on a develop push, so `github.event_name` is never `pull_request` here and these two lines can't change the outcome — same for the copy on `Post performance summary to pull request` below. I get that AGENTS.md asks for one shape at every site, so this may well be on purpose. Two things worth a quick check though: `Compute report URL` above was left unguarded while its twin in `performance-run/action.yml` got the guard, so the two copies now read like they follow different rules, and `fork-guards.test.mjs` lists both of these dead steps, which means the test will block anyone who later tries to clean them up.

budnix (commented):
Two small notes inline — one wrong PR number in the docs, and a question about the guards in `performance-tests.yml`. Nothing that breaks at runtime.

AGENTS.md
demtario: Confirmed, and it's the wrong number in the direction you guessed. Checked both runs against the API rather than the PR body:

- run `32481421878` -> branch `fix-repo-handsontable-cve-2026-31938-jspdf`, fork `anupamme` = PR #13207
- run `31252773025` -> branch `fix/source-data-validator-columns`, fork `ShiroKSH` = PR #13073

So this bullet was the odd one out, exactly as you said. Changed to `fork PR #13207, run 32481421878` so all four citations agree. Fixed in 386a2a43e2.

.github/workflows/performance-tests.yml
demtario: Dug into this one, and the premise turns out not to hold: these steps aren't dead. The file's own header is what makes them look that way, so your reading was the natural one.

`develop.yml` triggers on `pull_request` as well, with `paths: ['.github/workflows/develop.yml']` (self-validation of the `workflow_call` graph, so a broken graph doesn't first blow up after merging). Its `performance` job gates on `needs.checks.outputs.test-performance == 'true'` and nothing else, and `checks.yml`'s `scope` job carries no event condition while `test-all` is `github.event_name == 'workflow_dispatch'` (false on a PR), so the path filter computes normally.

So a PR that edits `develop.yml` **and** a `test-performance` path (`config-files`, `handsontable/src/**`, `performance-tests/**`) reaches this workflow with `github.event_name == 'pull_request'`. From a fork that's a read-only token against a workflow declaring `contents: write` + `pull-requests: write`: the gh-pages push 403s and the job goes red. The header's claim still holds for a workflow-only PR, since `.github/workflows/**` is deliberately excluded from `config-files` -- it's a mixed PR that gets through.

Which makes your asymmetry point a real defect rather than a style one. `Compute report URL` feeds `PAGES_URL` into `Run performance tests`, and `report-builder.mjs` only emits the "Full interactive report" link when it's set, so on that path the report advertised a URL that 404s because the deploy publishing it is skipped -- the same reason its twin in the composite was guarded. Guarded it here too, so both copies now resolve to the identical expression, and added a comment naming the reachable path so it doesn't get cleaned up as dead code later.

On the test blocking a future cleanup: that's now the wanted behavior, since the sites aren't dead. Added `['.github/workflows/performance-tests.yml', 'name: Compute report URL']` to `GUARDED_SITES` and mutation-tested it -- stripping the two clauses fails with `performance-tests.yml (name: Compute report URL): missing the same-repo half of the guard`, so it isn't passing vacuously.

Fixed in 386a2a43e2.

budnix (approved):

PR #10314, opened 2023-03-29
Add support for HYPERLINK (handsontable)

thilgen:
### Context

`HYPERLINK` is a formula provided by Google Sheets ([documentation](https://support.google.com/docs/answer/3093313?hl=en)); MS Excel ([documentation](https://support.microsoft.com/en-us/office/hyperlink-function-333c7ce6-c5ae-4164-9c47-7de9b76f577f)); and is included in the Open Document Format for Office Applications ([documentation](http://docs.oasis-open.org/office/v1.2/os/OpenDocument-v1.2-os-part2.html#__RefHeading__1018226_715980110))

### For more information
* Issue Writeup: https://github.com/handsontable/hyperformula/issues/1215
* Related `HyperFormula` Changes: https://github.com/handsontable/hyperformula/pull/1216

### How has this been tested?
Unit Tests and Test Application

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature or improvement (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Additional language file or change to the existing one (translations)

### Related issue(s):
1. https://github.com/handsontable/hyperformula/issues/1215

### Affected project(s):
- [X] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

- [X] I have reviewed the guidelines about [Contributing to Handsontable](https://github.com/handsontable/handsontable/blob/master/CONTRIBUTING.md) and I confirm that my code follows the code style of this project.
- [X] I have signed the [Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLScpMq4swMelvw3-onxC8Jl29m0fVp5hpf7d1yQVklqVjGjWGA/viewform?c=0&w=1)
- [X] My change requires a change to the documentation.

PR #13223, opened 2026-08-21
PRO-1051: Render HYPERLINK formulas as links

demtario:
### Context

The PR adds an opt-in `formulas.hyperlinks` option that renders a cell whose root expression is `HYPERLINK()` as a clickable link.

HyperFormula has exposed everything needed since v2.5.0 (`HyperlinkPlugin`, `getCellHyperlink()`), and we depend on `^3.0.0`, so the missing half was purely Handsontable-side rendering. Until now such a cell evaluated correctly but displayed its label as dead plain text.

This replaces #10314, which proposed the same feature in 2023 and was closed instead of rebased. Its review objections shape this implementation:

- **No new public renderer and no new cell type.** The `Formulas` plugin decorates the already rendered cell from the `afterRenderer` hook, so nothing is added to the renderer registry.
- **No cell meta is written.** #10314 restored `textRenderer` into cell meta for every non-hyperlink cell, which destroyed user-defined renderers and built-in cell types. Here the cell keeps its own renderer and there is nothing to undo when a formula is removed or the plugin is disabled.
- **The URL is parsed, not pattern-matched.** `resolveHyperlinkUrl()` resolves through `new URL()` and allowlists `http`, `https`, `mailto` and `tel`. Everything else, `javascript:` included, renders the label as plain text and warns once. Obfuscations that survive a string check (`JaVaScRiPt:`, `java\tscript:`, leading whitespace) normalize before the protocol is read.
- **No `innerHTML`.** The anchor is built with DOM APIs and the cell's existing child nodes are moved into it, so a label containing markup stays text whatever the `sanitizer` option is set to. This matters because `sanitize()` is a documented pass-through and `sanitizer` defaults to `undefined`.

Two behavioral details worth reviewing:

- **Default is `false`.** Enabling it by default would change rendering for existing grids that already hold `HYPERLINK` formulas, so it stays opt-in. Flipping the default is a separate major-release decision.
- **Keyboard access is <kbd>Alt</kbd>+<kbd>Enter</kbd>**, matching Google Sheets. The anchor carries `tabindex="-1"` so cell content stays out of the grid's tab order. `Alt`+`Enter` was free: `grid.ts` binds `Enter`, `Shift`+`Enter` and `Control/Meta`+`Enter`, and no core context binds any `Alt` combination.

Also of note: `htmlRenderer` has the same unsanitized-`href` gap that #10314 was flagged for. It is out of scope here and should be tracked separately.

### How has this been tested?

Unit, type, and Playwright E2E, all from the repo root:

```bash
npm --prefix tests run test:e2e -- e2e/formulas-hyperlink.spec.ts
```

- **Unit** (`handsontable/src/plugins/formulas/__tests__/hyperlinkUrl.unit.ts`, 21 cases): every allowed scheme accepted; `javascript:`, `data:`, `vbscript:` and `file:` refused, including mixed-case, tab-obfuscated and whitespace-padded variants; relative and protocol-relative URLs resolved against the base; empty, non-string and unparseable input refused. Full suite: 308 suites / 3669 tests green.
- **E2E** (`tests/e2e/formulas-hyperlink.spec.ts`, 16 cases): anchor attributes; the one-argument form; no anchor for a nested `HYPERLINK` call or a plain value; the refused-scheme path with exactly one warning; a markup label staying text; a column's custom renderer surviving and being wrapped; a `checkbox` cell type untouched; a `password` cell type's masked output surviving inside the anchor; the anchor disappearing when the formula is replaced or the plugin is disabled; the option toggled off and back on; click navigation plus cell selection; double click opening the editor with the raw formula; `Alt`+`Enter` on a link cell and on a non-link cell. 16/16 green on `e2e-main`.
- **Mutation-checked**: forcing the cached option flag to `false` turns six of those E2E cases red, so the spec binds to real behavior rather than merely executing it.
- **Render cost measured** on the worst case, 504 rendered cells all holding `HYPERLINK` formulas, median of 25 renders: 9.4 ms with the option off, 13.0 ms with it on, so roughly 7 µs per link cell including anchor creation. With the option off the per-cell cost is a single boolean read.
- **Legacy Jasmine/Puppeteer suite** (`npm --prefix handsontable run test:e2e`): 9669 specs, 0 failures, 8 pending. This is the closest existing coverage to the three lifecycle methods the PR touches (`enablePlugin`, `disablePlugin`, `updatePlugin`), including the missing-engine path in `initialization.spec.js` and `memoryLeak.spec.js`.

### Types of changes

### Related issue(s):

1. PRO-1051
2. Replaces #10314

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/86c8j1xwa

---

> [!NOTE]
> **Medium Risk**
> Touches formula rendering and URL handling (XSS-sensitive hrefs). Mitigated by an allowlist, DOM-only wrapping, and opt-in default, but still security-relevant.
> 
> **Overview**
> Adds an opt-in `formulas.hyperlinks` setting so a cell whose **root** formula is `HYPERLINK()` renders as a clickable link, without changing cell types or cell meta.
> 
> The Formulas plugin wraps the cell’s existing renderer output in an `a.ht-hyperlink` after render. URLs are parsed with `new URL()` and only `http`, `https`, `mailto`, and `tel` become links; other schemes (including `javascript:`) stay plain text. Nested `HYPERLINK` calls stay text. Links open in a new tab (`noopener,noreferrer`); keyboard access is **Alt+Enter** (anchors are out of tab order). Default remains off so existing grids are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c05c74786691b79bf8ca31fc77c430ed545bbbab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/handsontable/src/plugins/formulas/formulas.ts
+++ b/handsontable/src/plugins/formulas/formulas.ts
@@ -909,8 +909,16 @@ export class Formulas extends BasePlugin {
    */
   #refreshHyperlinksSetting() {
     const pluginSettings = this.hot.getSettings()[PLUGIN_KEY];
+    const wasEnabled = this.#hyperlinksEnabled;
 
     this.#hyperlinksEnabled = isFormulasSettingsObject(pluginSettings) && pluginSettings.hyperlinks === true;
+
+    // Turning the option off is the moment to clean up, not every subsequent draw: a renderer that
+    // leaves its previous DOM in place would keep an anchor that no later render pass rewrites.
+    // Doing it here keeps the per-cell path free for the default, disabled case.
+    if (wasEnabled && !this.#hyperlinksEnabled) {
+      this.#unwrapRenderedHyperlinks();
+    }
   }
 
   /**
@@ -923,15 +931,27 @@ export class Formulas extends BasePlugin {
   #unwrapRenderedHyperlinks() {
     this.hot.rootElement
       ?.querySelectorAll<HTMLElement>(`a.${HYPERLINK_CLASS_NAME}`)
-      .forEach((link) => {
-        const { parentNode } = link;
+      .forEach(link => this.#unwrapLink(link));
+  }
 
-        while (link.firstChild) {
-          parentNode?.insertBefore(link.firstChild, link);
-        }
+  /**
+   * Moves an anchor's content up into the anchor's own parent and drops the anchor.
+   *
+   * The insertion goes through `link.parentNode` rather than the cell: a renderer that leaves the
+   * previous DOM in place can wrap an existing anchor, leaving it as `TD > div > a` instead of a
+   * direct child. Inserting relative to the cell would then throw `NotFoundError` and, because this
+   * runs inside `afterRenderer`, take the whole draw down with it.
+   *
+   * @param {Element} link The anchor to unwrap.
+   */
+  #unwrapLink(link: Element) {
+    const { parentNode } = link;
 
-        link.remove();
-      });
+    while (link.firstChild) {
+      parentNode?.insertBefore(link.firstChild, link);
+    }
+
+    link.remove();
   }
 
   /**
@@ -950,11 +970,7 @@ export class Formulas extends BasePlugin {
     let link = TD.querySelector(`a.${HYPERLINK_CLASS_NAME}`);
 
     while (link !== null) {
-      while (link.firstChild) {
-        TD.insertBefore(link.firstChild, link);
-      }
-
-      link.remove();
+      this.#unwrapLink(link);
       link = TD.querySelector(`a.${HYPERLINK_CLASS_NAME}`);
     }
   }
@@ -1625,16 +1641,16 @@ export class Formulas extends BasePlugin {
    * @param {number} column Visual column index.
    */
   #onAfterRenderer = (TD: HTMLTableCellElement, row: number, column: number) => {
-    // Walkontable recycles TD elements, and a renderer is free to leave its previous DOM in place.
-    // Unwrapping first, unconditionally, keeps this idempotent by construction: no anchor nests
-    // inside another one across render passes, and the `href` is always rebuilt from the current
-    // formula instead of inherited from whatever the previous pass resolved. It also runs BEFORE the
-    // checks below, so turning the option off drops an anchor that such a renderer would keep.
-    this.#unwrapHyperlink(TD);
-
     if (!this.#hyperlinksEnabled || this.#internalOperationPending) {
       return;
     }
+
+    // Walkontable recycles TD elements, and a renderer is free to leave its previous DOM in place.
+    // Unwrapping first keeps this idempotent by construction: no anchor nests inside another one
+    // across render passes, and the `href` is always rebuilt from the current formula instead of
+    // inherited from whatever the previous pass resolved. Cleanup for the option being turned off
+    // happens once, in `#refreshHyperlinksSetting`, so this path never runs for a disabled grid.
+    this.#unwrapHyperlink(TD);
 
     const href = this.#getHyperlinkHref(row, column);
 
--- a/handsontable/src/plugins/formulas/formulas.ts
+++ b/handsontable/src/plugins/formulas/formulas.ts
@@ -875,8 +875,19 @@ export class Formulas extends BasePlugin {
           }
         },
         stopPropagation: true,
-        runOnlyIf: (): boolean => this.#hyperlinksEnabled &&
-          !!this.hot.getSelectedRangeActive()?.highlight.isCell(),
+        // The shortcut prevents the default action and stops propagation whenever `runOnlyIf`
+        // passes, so it must claim the chord only for a cell that actually resolves to a link.
+        // Testing just `isCell()` would swallow `Alt`+`Enter` grid-wide and break a host
+        // application's own handler for it.
+        runOnlyIf: (): boolean => {
+          const highlight = this.hot.getSelectedRangeActive()?.highlight;
+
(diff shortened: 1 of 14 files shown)


qunabu:
Task linked: [PRO-1051 HYPERLINK (formula + renderer)](https://app.clickup.com/t/9015210959/PRO-1051)

demtario:
Demo: 
https://demos.handsontable.com/d/34113j1t29/
https://demos.handsontable.com/share/34113j1t29

handsontable/src/plugins/formulas/formulas.ts
```diff
@@ -823,6 +851,156 @@ export class Formulas extends BasePlugin {
...
+    // common case and the one that must not pay for a selector query on every render pass.
+    if (TD.firstElementChild === null) {
+      return;
+    }
+
+    let link = TD.querySelector(`a.${HYPERLINK_CLASS_NAME}`);
+
+    while (link !== null) {
+      while (link.firstChild) {
+        TD.insertBefore(link.firstChild, link);
```
budnix: `TD.querySelector` finds the first matching anchor anywhere inside the cell, not only a direct child. So if a custom renderer wraps the cell content and the anchor ends up as `TD > div > a.ht-hyperlink`, `TD.insertBefore` throws `NotFoundError` — and since this runs inside `afterRenderer`, it takes down the whole draw. `#unwrapRenderedHyperlinks` right above already goes through `parentNode`; doing the same here would make both paths behave the same.

```suggestion
        link.parentNode?.insertBefore(link.firstChild, link);
```

handsontable/src/plugins/formulas/formulas.ts
```diff
@@ -1437,6 +1615,50 @@ export class Formulas extends BasePlugin {
...
+   * @param {number} row Visual row index.
+   * @param {number} column Visual column index.
+   */
+  #onAfterRenderer = (TD: HTMLTableCellElement, row: number, column: number) => {
+    // Walkontable recycles TD elements, and a renderer is free to leave its previous DOM in place.
+    // Unwrapping first, unconditionally, keeps this idempotent by construction: no anchor nests
+    // inside another one across render passes, and the `href` is always rebuilt from the current
+    // formula instead of inherited from whatever the previous pass resolved. It also runs BEFORE the
+    // checks below, so turning the option off drops an anchor that such a renderer would keep.
+    this.#unwrapHyperlink(TD);
```
budnix: I get why the unwrap moved up here — it's what makes the option flip clean up a memoizing renderer's leftover anchor. The cost is that it now runs for everyone: `hyperlinks` defaults to `false`, so with the plain Formulas setup every cell of every draw does the property read, and every cell that has an element child (checkbox, dropdown arrow, any custom element renderer) also runs the `querySelector`. Could `#refreshHyperlinksSetting` call `#unwrapRenderedHyperlinks()` when the flag goes from true to false instead — the same thing `disablePlugin` already does? Then the check can go back on top, the toggle-off guarantee stays, and the E2E test that pins the memoizing case should still pass.

budnix (commented):
Two notes on the render path, both small. The rest of the change reads well — the protocol allowlist, the `grid` shortcut context timing, and the SCSS specificity over the base `a:not(...)` rule all check out.

handsontable/src/plugins/formulas/formulas.ts
demtario: Good catch, and it is worse than a shape nit: thrown inside `afterRenderer` it kills the whole draw, so every cell after it stays unrendered.

Taken, and pushed the deduplication you were pointing at rather than just the one line. There is now a single `#unwrapLink(link)` primitive that always inserts through `link.parentNode`, and both callers use it — `#unwrapHyperlink` (per cell, in the render path) and `#unwrapRenderedHyperlinks` (the sweep). Two copies of this were going to drift.

Covered by a new E2E case: fixture column F renders through a renderer that boxes whatever the cell already holds, so from the second draw the anchor is `TD > div.wrap-box > a`. The test forces two extra draws, then asserts one anchor with the right `href`, no `pageerror`, and that a cell *later* in render order is still rendered — that last one is what actually catches a draw killed mid-flight. Reverting to the cell-relative insert turns it red.

`1c23b9774c`

handsontable/src/plugins/formulas/formulas.ts
demtario: Agreed, and your shape is better than what I had. Done in `1c23b9774c`.

`#refreshHyperlinksSetting` now keeps the previous flag value and calls `#unwrapRenderedHyperlinks()` on a true→false transition, so the cleanup happens once at the flip instead of on every draw forever. The enabled check went back to the top of `#onAfterRenderer`, so a default `formulas` grid is back to one boolean read per cell — no `querySelector`, and nothing extra for cells with element children.

The memoizing-renderer E2E case does still pass, and it now pins the flip path rather than the per-draw one. It stays honest: disabling the flip-time cleanup turns it red.

Worth recording why the sweep exists in two places. `disablePlugin` needs its own call because it removes the hook, and a caller that disables without a redraw leaves no render pass to rewrite anything — that one is pinned by a separate case that disables the plugin *without* calling `render()`. The earlier version of that test passed for the wrong reason: disabling changes the cell value from the evaluated label to the raw formula, which busted my memo key, so the renderer cleared the cell on its own.

budnix (approved):

adrianspdev (approved):

Issue #10239, opened 2023-02-09
updateData() which changes children size and has collapsed parent rows results in an invalid state


aaronbeall:
_**PS**: Just want to take another moment to say THANK YOU for making `updateData()` work nearly perfectly in `nestedRows`! This is the only issue I've found so far._

### Description
If you have (1) `nestedRows` enabled, (2) some parent rows collapsed, and (3) call `updateData()` which includes `__children` of a different length than the previous data, you can end up in a state where children rows are not properly hidden/shown based on their parent expanded/collapsed state.

### Steps to reproduce

1. Open https://jsfiddle.net/yqho74dx/
2. Collapse the children of the first 2 parent rows, note each parent row has 3 children
3. Click the button "Update children bug" which calls `updateData()` with 2 child rows on each parent row

Results: 
* 2 orphaned child rows appear below the first parent row, despite it showing the parent as collapsed
* 2 parent rows have gone missing (id 1 and 2), which should appear where the orphaned child rows are
* When you expand the first parent row, it shows 4 child rows (though the update only had 2) and collapsing it again leaves the 2 extra child rows visible

![img](http://g.recordit.co/V1EIYhAfuN.gif)

Note:
* There are other cases like this, all of which relate to having some parent rows collapsed and updating some rows with a different length of children.
* Without any parent rows collapsed, updates work with any random number of parent rows or child rows being changed.
* With parent rows collapsed, if none of the updated rows have a different children length, it works fine

### Demo

https://jsfiddle.net/yqho74dx/

### Your environment
* Handsontable version: 12.3.1
* Browser name and version: Chrome 109.0.5414.119
* Operating system: MacOS 12.4

AMBudnik:
Yes, 'nearly' is the keyword :)

You're right. There's definitely something going on here. In this case I see bugs from the data side and front the view side. We should get this on the front 

![Zrzut ekranu 2023-02-10 o 11 10 27](https://user-images.githubusercontent.com/10757813/218064759-c7ffd306-df2c-47f5-ae82-24a1c6a01600.png)

and this in the `getData()` call

![Zrzut ekranu 2023-02-10 o 11 10 45](https://user-images.githubusercontent.com/10757813/218064824-00a9b89b-130b-4e37-aaf4-4dab4b96af76.png)

But. We get this on the front 
[...]

PR #13224, opened 2026-08-24
DEV-2594: Fix nestedRows invalid state after updateData changes the children count

budnix:
### Context

The PR fixes an invalid grid state that appears when `nestedRows` is on, some parents are collapsed, and `updateData()` is called with a different number of `__children` (GitHub #10239, reported against 12.3.1 and still reproducing on 18.0.0).

The collapsed state is kept in two places, and both are keyed by the **physical** row index:

1. `collapsedRowsMap` — the `trimming` index map registered in `nestedRows.ts`.
2. `CollapsingUI.collapsedRows` — a plain array of collapsed parents.

`loadData()` resets the index maps through `initIndexMappers()`, but `updateData()` only resizes them with `rowIndexMapper.fitToLength()`. The plugin's own `beforeLoadData` / `beforeUpdateData` handler rebuilt the data-manager cache and touched neither store. So as soon as a parent gained or lost a child, every physical index after it shifted and the kept trimmed indexes landed on different rows — parent rows included, which then disappeared while their children stayed on screen.

Reproduced on `develop` @ `b5ab75f8`: three parents with three children each, parents 1 and 2 collapsed, then `updateData()` with two children each leaves three rows on screen — `P1`, `P2-C1`, `P3-C2`. `getData()` and the DOM agree; both are wrong.

**The fix.** The two data hooks now do different things, matching what each method promises:

- `beforeLoadData` drops the collapsed state. `loadData()` resets the rows' states by contract, and this also clears a latent inconsistency where `collapsedRows` kept listing parents from the previous dataset.
- `beforeUpdateData` records the collapsed parents as **tree paths** (the chain of child indexes from the top level down), clears the stores, and `afterUpdateData` collapses again whatever those paths still resolve to. The replay runs shallowest-first and with the hooks silenced, because it repeats a choice the user already made.

Two new `DataManager` helpers back this: `getRowTreePath()` and `getRowIndexByTreePath()`. A tree path is positional, so it follows the slot rather than the object — that is the best available match when the new dataset carries no identity, and it covers the reported case, where parents keep their place in the tree and only the children counts change.

**A third store, found in review.** `collapsingUI.lastCollapsedRows` — the stash that `beforeAddChild`, `beforeDetachChild`, `beforeRemoveRow` and a row move keep open around themselves — is keyed by physical row index as well. It matters because during a stash window `collapsedRows` is already **empty**: the stash expanded the grid and parked the state. So an app that replaces the data from inside one of those hooks captures nothing on the main path, and `applyStash()` later replays the old row numbers onto the new data. Verified against a real grid before fixing: with parents 1 and 2 collapsed, a stash open, and `updateData()` shrinking the children, `applyStash()` restored only parent 1 and silently dropped parent 2 — with a different shape it would collapse the wrong parent instead. The stash is now re-pointed through the same tree paths in `afterUpdateData`, and zeroed on `loadData()`.

The trimming map is cleared only when a parent really is collapsed. Clearing it rebuilds the row index cache, and a data load runs on every grid init — `core.unit.js` asserts exactly one cache reset there.

### How has this been tested?

The reported scenario was reproduced in a real browser before the change and re-checked after it. The new specs were confirmed to fail on the unfixed build and pass on the fixed one — the first six with the main fix reverted (all 6 red), and the stash test with only the stash handling removed (1 red, the other 6 still green).

- New Playwright E2E — `tests/e2e/nested-rows-update-data.spec.ts`, 7 tests: fewer children, more children, a parent collapsed inside another collapsed parent, a collapsed parent left without children, no hooks fired on replay, a stash left open by another operation, and `loadData()` dropping the state.
- New unit tests — `handsontable/src/plugins/nestedRows/__tests__/data/dataManager.unit.js`, 7 tests for the tree-path helpers, including a round trip across a data swap.

```bash
# unit (whole core suite: 3673 passed)

# types

# legacy E2E (whole suite: 9669 specs, 0 failures)

# Playwright functional E2E (e2e-main leg: 356 passed, 1 skipped)
npm --prefix tests run test:e2e

# lint
npm run eslint --prefix handsontable    # 0 errors
npm --prefix tests run lint
```

### Types of changes

### Related issue(s):

1. DEV-2594
2. #10239

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2594

---

> [!NOTE]
> **Medium Risk**
> Touches nestedRows collapse/trim state and data-replacement hooks, which can hide the wrong rows or leave selection past the last row if the remapping is wrong. Behavior of loadData vs updateData is now explicitly different.
> 
> **Overview**
> Fixes **#10239**: collapsed nested parents no longer vanish (or hide the wrong rows) when `updateData()` changes child counts. Collapse stores were keyed by physical indexes, which shift as soon as a parent gains or loses children.
> 
> **`updateData()`** now snapshots collapsed parents (and any open collapse stash) as **tree paths**, clears the old indexes, then re-collapses whatever those paths still resolve to. Replay is silent (no collapse hooks) and skips an extra render; selection is refreshed so the highlight cannot sit past the last visible row. Parents that disappear or lose all children are dropped.
> 
> **`loadData()`** now clears collapsed state, matching its “reset row state” contract, so `getCollapsedParents()` no longer reports the previous dataset.
> 
> Adds `getRowTreePath()` / `getRowIndexByTreePath()` on `DataManager`, documents the load vs update difference, and covers the cases with unit and Playwright tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f6180d48bfa0423fa8ab32929bfe24c9d00a5638. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/handsontable/src/plugins/nestedRows/data/dataManager.ts
+++ b/handsontable/src/plugins/nestedRows/data/dataManager.ts
@@ -317,6 +317,11 @@ class DataManager {
    * A physical row index shifts as soon as any other node gains or loses children. A tree path does
    * not, so it can be used to find the same node again after the data is replaced.
    *
+   * Assumes each row object appears in the tree once. The path is built with `indexOf`, on object
+   * identity, while `getRowIndexByTreePath()` finishes through the `cache.nodeInfo` WeakMap, which
+   * keeps only a node's last occurrence - so the same object placed at two spots makes the two
+   * disagree.
+   *
    * @param {number} row Physical row index.
    * @returns {number[]|null} The path, or `null` when the row is not part of the current structure.
    */
@@ -346,10 +351,12 @@ class DataManager {
   /**
    * Find the physical row index of the node that the provided tree path points at.
    *
-   * @param {number[]} path Chain of child indexes, as returned by {@link DataManager#getRowTreePath}.
+   * @param {number[]|null} path Chain of child indexes, as returned by
+   * {@link DataManager#getRowTreePath} - which returns `null` for a row it does not know, so the
+   * `null` is accepted here rather than filtered at every call site.
    * @returns {number|null} Physical row index, or `null` when the path leads outside the structure.
    */
-  getRowIndexByTreePath(path: number[]): number | null {
+  getRowIndexByTreePath(path: number[] | null): number | null {
     if (!Array.isArray(path) || path.length === 0) {
       return null;
     }
--- a/handsontable/src/plugins/nestedRows/data/dataManager.ts
+++ b/handsontable/src/plugins/nestedRows/data/dataManager.ts
@@ -311,6 +311,66 @@ class DataManager {
   }
 
   /**
+   * Get the position of a row within the nested structure, expressed as the chain of child indexes
+   * that leads from the top level down to that row.
+   *
+   * A physical row index shifts as soon as any other node gains or loses children. A tree path does
+   * not, so it can be used to find the same node again after the data is replaced.
+   *
+   * @param {number} row Physical row index.
+   * @returns {number[]|null} The path, or `null` when the row is not part of the current structure.
+   */
+  getRowTreePath(row: number): number[] | null {
+    let rowObject: RowObject | null | undefined = this.getDataObject(row);
+
+    if (!rowObject) {
+      return null;
+    }
+
+    const path: number[] = [];
+
+    while (rowObject) {
+      const indexWithinParent = this.getRowIndexWithinParent(rowObject);
+
+      if (indexWithinParent === -1) {
+        return null;
+      }
+
+      path.unshift(indexWithinParent);
+      rowObject = this.getRowObjectParent(rowObject);
+    }
+
+    return path;
+  }
+
+  /**
+   * Find the physical row index of the node that the provided tree path points at.
+   *
+   * @param {number[]} path Chain of child indexes, as returned by {@link DataManager#getRowTreePath}.
+   * @returns {number|null} Physical row index, or `null` when the path leads outside the structure.
+   */
+  getRowIndexByTreePath(path: number[]): number | null {
+    if (!Array.isArray(path) || path.length === 0) {
+      return null;
+    }
+
+    let siblings: RowObject[] | null | undefined = this.data;
+    let node: RowObject | null = null;
+
+    for (let i = 0; i < path.length; i++) {
+      node = siblings?.[path[i]] ?? null;
+
+      if (node === null) {
+        return null;
+      }
+
+      siblings = node.__children;
+    }
+
+    return this.getRowIndex(node);
+  }
+
+  /**
    * Count all rows (including all parents and children).
    *
    * @returns {number}
--- a/handsontable/src/plugins/nestedRows/nestedRows.ts
+++ b/handsontable/src/plugins/nestedRows/nestedRows.ts
@@ -112,6 +112,14 @@ export class NestedRows extends BasePlugin {
   #collapsedParentPaths: number[][] = [];
 
   /**
+   * Tree paths held by a stash that an outer operation left open when `updateData()` started.
+   * `null` means there was no open stash.
+   *
+   * @type {number[][]|null}
+   */
+  #stashedParentPaths: number[][] | null = null;
+
+  /**
    * Checks if the plugin is enabled in the handsontable settings. This method is executed in {@link Hooks#beforeInit}
    * hook and if it returns `true` then the {@link NestedRows#enablePlugin} method is called.
    *
@@ -953,19 +961,58 @@ export class NestedRows extends BasePlugin {
   /**
    * Forgets which parents are collapsed, and untrims the rows that collapse had hidden.
    *
-   * Both stores are keyed by physical row index, and neither of those indexes means anything once
-   * the data is replaced.
+   * All three stores are keyed by physical row index, and none of those indexes means anything once
+   * the data is replaced: the collapsed-parents list, the trimming map, and the stash an outer
+   * operation left open.
    *
-   * There is nothing to do when no parent is collapsed, and skipping the map matters there:
+   * There is nothing to trim when no parent is collapsed, and skipping the map matters there:
    * clearing it rebuilds the row index cache, and a data load runs on every grid init.
    */
   #clearCollapsedState() {
-    if (!this.collapsingUI || this.collapsingUI.collapsedRows.length === 0) {
+    if (!this.collapsingUI) {
+      return;
+    }
+
+    if (this.collapsingUI.lastCollapsedRows) {
+      this.collapsingUI.lastCollapsedRows = [];
+    }
+
+    if (this.collapsingUI.collapsedRows.length === 0) {
       return;
     }
 
     this.collapsingUI.collapsedRows.length = 0;
     this.collapsedRowsMap?.clear();
+  }
+
+  /**
+   * Translates physical row indexes into tree paths, dropping the rows the current cache does not
+   * know about.
+   *
+   * @param {number[]} rows Physical row indexes.
+   * @returns {number[][]}
+   */
+  #toTreePaths(rows: number[]): number[][] {
+    return rows
+      .map(row => this.dataManager!.getRowTreePath(row))
+      .filter((path): path is number[] => path !== null);
+  }
+
+  /**
+   * Translates tree paths back into physical row indexes, keeping only the rows that still exist and
+   * still have children.
+   *
+   * Sorted shallowest first, so an ancestor is always handled before its own descendants.
+   *
+   * @param {number[][]} paths Tree paths.
+   * @returns {number[]} Physical row indexes.
+   */
+  #toCollapsibleRows(paths: number[][]): number[] {
+    return paths
+      .slice()
+      .sort((pathA, pathB) => pathA.length - pathB.length)
+      .map(path => this.dataManager!.getRowIndexByTreePath(path))
+      .filter((row): row is number => row !== null && this.dataManager!.hasChildren(row));
   }
 
   /**
@@ -1002,9 +1049,13 @@ export class NestedRows extends BasePlugin {
       return;
     }
 
-    this.#collapsedParentPaths = (this.collapsingUI?.getCollapsedParents() ?? [])
-      .map(parentRow => this.dataManager!.getRowTreePath(parentRow))
-      .filter((path): path is number[] => path !== null);
+    const openStash = this.collapsingUI?.lastCollapsedRows;
+
+    this.#collapsedParentPaths = this.#toTreePaths(this.collapsingUI?.getCollapsedParents() ?? []);
+    // An outer operation - add child, detach child, remove row, row move - may hold the collapsed
+    // state in an open stash instead, having expanded the grid for the duration. That copy has to
+    // be re-pointed as well, or `applyStash()` collapses whatever now sits at the old indexes.
+    this.#stashedParentPaths = openStash ? this.#toTreePaths(openStash) : null;
 
     this.#clearCollapsedState();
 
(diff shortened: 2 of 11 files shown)


qunabu:
Task linked: [DEV-2594 Fix nestedRows invalid state after updateData() changes the children count with collapsed parents](https://app.clickup.com/t/9015210959/DEV-2594)

handsontable/src/plugins/nestedRows/nestedRows.ts
```diff
@@ -922,25 +939,160 @@ export class NestedRows extends BasePlugin {
...
+
+    if (stashedPaths !== null) {
+      this.collapsingUI.lastCollapsedRows = this.#toCollapsibleRows(stashedPaths);
+    }
+
+    const parentsToCollapse = this.#toCollapsibleRows(paths);
+
+    if (parentsToCollapse.length > 0) {
+      // Replaying a state the user already chose is not a new action, so the hooks stay silent.
+      this.collapsingUI.toggleCollapsedRows(parentsToCollapse, 'collapse', false);
```
Haxikowy: The replay trims after core has already clamped the selection, and a trimming change never re-clamps. `selection.refresh()` runs at core.ts:3101, before `afterUpdateData` (replaceData.ts:158), and clamps against a fully expanded grid because `#clearCollapsedState()` emptied the map in `beforeUpdateData`. `IndexMapper` reports the trim as `trimmedIndexesChanged`, and `selection.commit()` only runs for `hiddenIndexesChanged` (core.ts:725).

Select B-2 (visual 8), collapse Root B, `updateData`: highlight stays at 8 with 5 rows visible, and Enter on that header feeds an out-of-range visual row to `toPhysicalRow()`. Base clamped against the stale trims, so it was wrong but in range.

`this.hot.selection.refresh()` at the end of `#onAfterUpdateData`, plus a spec that selects a row first. Code reading, not reproduced.

handsontable/src/plugins/nestedRows/nestedRows.ts
```diff
@@ -922,25 +939,160 @@ export class NestedRows extends BasePlugin {
...
   };
 
+  /**
+   * `afterUpdateData` hook callback.
+   *
+   * Collapses the parents that were collapsed before the update and are still parents in the new
+   * data, and re-points an open stash at the same rows. A parent that the new data dropped, or that
+   * no longer has children, is simply forgotten.
+   */
+  #onAfterUpdateData = () => {
```
Haxikowy, line 1079: Two full renders per `updateData()` when anything is collapsed. `applyCollapsedRowsChange` sees the set go `[]` to N, so it runs `renderAndAdjust()` (collapsing.ts:480), and `replaceData` renders again as soon as this hook returns.

Angular calls `updateData()` on every `data` input change (hot-table.component.ts:151); React and Vue arrive via `updateSettings({data})`. So one data prop change pays a second whole-grid render, then `updatePlugin()` replays on top.

`hot.batch()` around the body, or a `forceRender` pass-through on `toggleCollapsedRows`. The tree path work is not the cost. Measured in Node against a reimplementation of the branch helpers: 0.13 ms to capture, 0.02 ms to resolve at 10k rows / 500 collapsed, against 0.89 ms for the `rewriteCache()` that already runs on every `updateData`.

handsontable/src/plugins/nestedRows/nestedRows.ts
```diff
@@ -922,25 +939,160 @@ export class NestedRows extends BasePlugin {
...
+   *
+   * Sorted shallowest first, so an ancestor is always handled before its own descendants.
+   *
+   * @param {number[][]} paths Tree paths.
+   * @returns {number[]} Physical row indexes.
+   */
+  #toCollapsibleRows(paths: number[][]): number[] {
+    return paths
+      .slice()
+      .sort((pathA, pathB) => pathA.length - pathB.length)
```
Haxikowy: The sort has no effect. Every path resolves before `toggleCollapsedRows` runs, and collapsing changes trimming rather than physical indexes, so order cannot make a deeper path unresolvable. Downstream order is unobservable too: `collapseMultipleChildren` trims once, `getCollapsedParents()` re-sorts ascending. Delete it and all seven specs stay green.

Drop it with the JSDoc claim and the AGENTS.md line, or add a test that fails without it. `getCollapsedParentsShallowestFirst()` (collapsing.ts:599) already covers the same idea.

handsontable/src/plugins/nestedRows/nestedRows.ts
```diff
@@ -922,25 +939,160 @@ export class NestedRows extends BasePlugin {
...
+    if (this.collapsingUI.lastCollapsedRows) {
+      this.collapsingUI.lastCollapsedRows = [];
+    }
+
+    if (this.collapsingUI.collapsedRows.length === 0) {
+      return;
+    }
+
+    this.collapsingUI.collapsedRows.length = 0;
+    this.collapsedRowsMap?.clear();
```
Haxikowy: Redundant on the `loadData` path: `initIndexMappers()` resets every map immediately after (core.ts:3140), so a `loadData` with a collapsed parent pays two cache rebuilds. That path needs only `collapsedRows.length = 0` and the `lastCollapsedRows` reset.

The `length === 0` guard above is load-bearing: core.unit.js:75 asserts exactly one `cacheUpdated` on init with `nestedRows: true`. Worth saying so in the comment.

handsontable/src/plugins/nestedRows/data/dataManager.ts
```diff
@@ -310,6 +310,66 @@ class DataManager {
...
+    return path;
+  }
+
+  /**
+   * Find the physical row index of the node that the provided tree path points at.
+   *
+   * @param {number[]} path Chain of child indexes, as returned by {@link DataManager#getRowTreePath}.
+   * @returns {number|null} Physical row index, or `null` when the path leads outside the structure.
+   */
+  getRowIndexByTreePath(path: number[]): number | null {
```
Haxikowy: `dataManager.unit.js:82` asserts `getRowIndexByTreePath(null)`, but the signature is `path: number[]`, so a TS caller cannot reach the guard the test pins. Widen to `number[] | null`, or drop the case.

handsontable/src/plugins/nestedRows/data/dataManager.ts
```diff
@@ -310,6 +310,66 @@ class DataManager {
...
+   * that leads from the top level down to that row.
+   *
+   * A physical row index shifts as soon as any other node gains or loses children. A tree path does
+   * not, so it can be used to find the same node again after the data is replaced.
+   *
+   * @param {number} row Physical row index.
+   * @returns {number[]|null} The path, or `null` when the row is not part of the current structure.
+   */
+  getRowTreePath(row: number): number[] | null {
+    let rowObject: RowObject | null | undefined = this.getDataObject(row);
```
Haxikowy, line 329: `getRowTreePath` positions by `indexOf` on object identity; `getRowIndexByTreePath` finishes through `getRowIndex`, which reads the `cache.nodeInfo` WeakMap. Reuse one row object twice and the two disagree, since `cacheNode` keeps only the last occurrence: for `[{id:'A',__children:[shared]}, shared]`, `getRowTreePath(1)` gives `[1]`, and `[0,0]` and `[1]` both resolve to 2. The replay then collapses a different row.

Pre-existing in `getRowIndex`, but these helpers make it load-bearing for state restore. A JSDoc note on the constraint is enough.

tests/e2e/nested-rows-update-data.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+    await nestedRows.resetHookLog();
+
+    await nestedRows.updateData(tree(['A-2-a'], ['B-1', 'B-2']));
+
+    // The user already chose this state, so restoring it is not a new collapse.
+    expect(await nestedRows.hookNames()).toEqual([]);
+    expect(await nestedRows.collapsedParents()).toEqual([5]);
+  });
+
+  test('re-points a stash that another operation left open', async({ page, theme }) => {
```
Haxikowy, line 140: Calling `stash()`/`applyStash()` directly does not prove the two data hooks fire, or fire in order, inside a real window. The case this handling exists for is `updateData()` from inside `beforeAddChild`, `beforeDetachChild`, `beforeRemoveRow` or a row move; one spec through any of those pins it.

Also: `lastCollapsedRows` is `[2, 6]` here, already shallowest-first, so collapsing A-2 first would give the sort in `#toCollapsibleRows` something to do. And no case drops a collapsed parent from the data entirely, which is the `path === null` filter rather than `hasChildren`.

Haxikowy (commented):
Docs - the documentation checkbox needs ticking. `row-parent-child.md:341` says four hooks report every collapse and expand; the replay fires none, so that sentence is now wrong, and it is what an integrator mirroring collapsed state relies on. The "Save and restore the collapsed rows" section at 362 teaches a manual re-collapse on stored physical indexes, which after `updateData()` now double-collapses on shifted ones. Same section is the place for the two new rules: `loadData()` drops the state, `updateData()` keeps it by position.

Changelog - title covers the `updateData` symptom only. `getCollapsedParents()` after `loadData()` now returns `[]` instead of stale parents, and collapsed parents surviving `updateData()` is a new contract.

handsontable/src/plugins/nestedRows/nestedRows.ts
budnix: Nice catch — reproduced it, and you were right that it was new: base clamped to 5, my branch left it at 8. Fixed in `f6180d4` with a `selection.refresh()` at the end of the replay, plus a spec that fails without it.

handsontable/src/plugins/nestedRows/nestedRows.ts
budnix: Good point — dropped the second render in `f6180d4` with a `forceRender` pass-through on `toggleCollapsedRows`, so the replay leaves the render to `replaceData`.

handsontable/src/plugins/nestedRows/nestedRows.ts
budnix: You're right, the sort did nothing — removed it in `f6180d4`, along with the JSDoc and AGENTS.md claims.

handsontable/src/plugins/nestedRows/nestedRows.ts
budnix: Fair point — `loadData` no longer clears the map in `f6180d4`, and the comment now says why the length guard has to stay.

handsontable/src/plugins/nestedRows/data/dataManager.ts
budnix: Good spot — widened the parameter to `number[] | null` in `f6180d4`, so the case the test pins is actually reachable.

handsontable/src/plugins/nestedRows/data/dataManager.ts
budnix: Agreed — added that note to `getRowTreePath` in `f6180d4`, spelling out the `indexOf` versus WeakMap mismatch.

tests/e2e/nested-rows-update-data.spec.ts
budnix: Yep, my mistake — `f6180d4` adds a spec that replaces the data from inside a real `beforeAddChild` window, and one for a collapsed parent dropped from the data entirely.

budnix:
Both done in `f6180d4` — the guide now covers the silent replay and the two rules (`updateData` keeps the state by position, `loadData` drops it), and there is a second changelog entry for the changed contract.

Haxikowy (approved):

Issue #8624, opened 2021-08-06
Comment tooltip not opening on cell hovering when rendering handsontable inside shadowroot


jpradelle:
### Description
When rendering handsontable in shadowroot, hovering a cell with comment doesn't open comment tooltip.
When using `showAtCell` function, comment tooltip is well opened.

### Steps to reproduce
1. Put handsontable in shadowroot
2. Add a comment on a cell
3. Hover commented cell

### Demo
https://jsfiddle.net/jpradell/d0cvsjra/2/

### Your environment
* Handsontable version: 9.0.2
* Browser Name and version: Chrome 92.0.4515.131 (Official Build) (64-bit)
* Operating System: Linux

AMBudnik:
Thank you for sharing the issue report @jpradelle 

I have replicated the issue and mark it as s `bug`.

budnix:
Thank you for reporting the error. The fix has already been made (https://github.com/handsontable/handsontable/pull/13227) and will be released in the upcoming versions.

PR #13227, opened 2026-08-24
DEV-2596: Fix the comment tooltip not opening on hover inside Shadow DOM (#8624)

budnix:
### Context

The PR fixes the comment tooltip not opening when you hover a commented cell in a grid rendered inside a shadow root. Calling `showAtCell()` on the same cell always worked, which is why the report singled hovering out.

Reported in #8624 (2021, Handsontable 9.0.2) and still reproducible on `develop` @ `52e0c9efbd`. The Shadow DOM work in #13194 (DEV-1619) did not cover this path — its only change to the plugin was `isFocused()` in `commentEditor.ts`.

**Cause.** The plugin bound its mouse listeners on the document only. That sits outside the grid's shadow tree, so the browser retargets the event and `event.target` arrives as the shadow host. `targetIsCellWithComment()` then looked for a `TD` above the host, found none, and the tooltip never opened. The `else` branch ran instead and hid it.

Three things were needed, and each is verified below:

1. **Bind the mouse handlers to the grid's shadow root as well.** A listener bound inside the tree receives the real cell as `event.target` — retargeting only affects listeners outside the tree. This is the mechanism that makes hovering work, and it is also what a sandboxed host (Salesforce LWS) needs, because LWS collapses `composedPath()` to the host chain and leaves a document listener with no way to reach the cell.

2. **Dedupe the events across the two bindings.** Both bindings receive the same in-grid event, and they never agree: the listener inside the tree sees the real cell, while the document listener sees the retargeted shadow host. That is intrinsic to retargeting, so it applies to every shadow-hosted grid, not only sandboxed ones. Undeduped, one hover showed from the first and hid from the second — and `DisplaySwitch.hide()` sets `wasLastActionShow = false`, which the queued debounced show re-reads when it fires, so the show was dropped and the tooltip never appeared. The shadow-root listener runs first (the event reaches the `ShadowRoot` before crossing to the host), so it wins for anything inside the grid, and the document listener keeps handling only what never entered the shadow tree.

3. **Make the hide branch's containment test shadow-aware.** It used `isChildOf()`, a plain `parentNode` walk, which dead-ends at the `ShadowRoot` (whose `parentNode` is `null`) and so reports every cell in a shadow tree as detached. It returned `true` before only because it was handed the host. Fixing the target without this turns "never shows" into "never hides". It is now one composed-root-node comparison, which follows host elements across every boundary and still reports a genuinely detached element as detached.

**What is deliberately *not* here.** An earlier revision of this PR also resolved the event target through `composedPath()`. Mutation testing at HEAD showed that layer kills no test: once the handlers are bound to the shadow root, that listener wins every in-grid event and its `event.target` is already the real cell. It was two overlapping mechanisms where one suffices, so it is gone — along with the optional `target` parameter it needed on `targetIsCellWithComment()`/`targetIsCommentTextArea()`, which keeps both at their `develop` signatures. `copyPaste.ts` and `helpers/dom/element.ts` are untouched.

Light-DOM grids are unaffected: the second binding, the dedupe and the point reader all hang off one `isShadowRoot()` gate, and nothing on the hover path changed for them.

### How has this been tested?

Extended the existing DEV-1619 shadow DOM harness rather than adding a fixture. `tests/fixtures/demo/shadow-dom.html` gains a comment on B4, a second unrelated shadow root, and both stylesheets in `<head>` (the comment editor is portaled into `document.body`, so without them it rendered in page flow and the pointer could not reach it). `tests/e2e/shadow-dom.spec.ts` goes from 9 cases to **14**.

```bash
npm run build:umd --prefix handsontable
cd tests && npx playwright test e2e/shadow-dom.spec.ts --project=e2e-main
```

**14 passed**, stable across three consecutive runs. All nine pre-existing cases still pass, the context-menu one included.

Rather than one both-directions run, here is what each mechanism is worth — every row is a mutation applied at HEAD, rebuilt, and run:

| mutation at HEAD | cases killed |
|---|---|
| remove the shadow-root binding | 5 |
| remove the dedupe | 5 |
| containment → `develop`'s `isChildOf` | 1 — `hides… when the pointer moves to a cell without a comment` |
| textarea check → `false` | 1 — `keeps… open when the pointer moves onto it` |
| remove the `composedPath()` resolution | **0** — which is why it is not in the diff |

Two of the new cases drive `page.clock`. Hiding runs on a timer, so asserting straight after the hover passed before a wrongly queued hide could fire — the first version of the hover-onto-tooltip case was hollow and its mutant survived. With the clock both mutants die.

```bash
npx eslint --ext .ts src/plugins/comments/comments.ts   # clean
npm run test:types --prefix handsontable                # clean
```

**One caveat for the reviewer.** Every shadow-aware path hangs off `isShadowRoot()`, and the fixture mounts a *native* shadow root, so that gate's false side is never exercised: a host whose synthetic root fails the predicate makes the whole change inert while the suite stays green. Nothing here is Salesforce-specific any more — the fix is plain retargeting plus a dedupe — so there is no LWS-only path left to simulate, and the probe that used to do it has been removed as inert. The legacy Jasmine comments suite could not run in my worktree (`jasmine is not defined`, a local `node_modules` linking artifact), so CI is the first full run of it, along with the `-min` and horizon/classic legs.

### Types of changes

### Related issue(s):

1. DEV-2596
2. #8624

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

The Shadow DOM guide's known limitation about the creation-time root-node check now mentions the comment listeners it also controls.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2596

---

> [!NOTE]
> **Medium Risk**
> Touches Comments plugin mouse handling (show/hide on hover and click). Light-DOM behavior is gated behind `isShadowRoot()`, but dual listeners and event dedupe could still affect tooltip timing if the gate misfires.
> 
> **Overview**
> Fixes **comment tooltips never appearing on hover** when Handsontable is mounted in a shadow root (#8624). Document-only listeners saw the retargeted host, so the cell was never recognized.
> 
> The Comments plugin now also binds `mouseover`/`mousedown`/`mouseup` on the grid’s shadow root, **dedupes** the two listeners so a show is not immediately cancelled by the document handler, and replaces `isChildOf` with a composed-root check so the tooltip still **hides** when the pointer leaves the grid (including into another shadow tree). Light-DOM grids skip the extra path.
> 
> Playwright shadow-DOM e2e coverage is extended for show, hide, other-shadow-tree leave, hover onto the portaled editor, and click on the commented cell. The Shadow DOM guide notes that comment listeners are also tied to the creation-time root-node check.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ab5f71d047a80b2882ff0723061fa12e9e95b04f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/content/guides/tools-and-building/shadow-dom/shadow-dom.md
+++ b/docs/content/guides/tools-and-building/shadow-dom/shadow-dom.md
@@ -94,7 +94,7 @@ Salesforce Lightning Experience renders Lightning Web Components with its synthe
 
 ## Known limitations
 
-- Handsontable checks the container's root node only when the instance is created. The check controls the `ht-shadow-dom` class and the shadow-root clipboard listeners that copy and paste rely on in sandboxed hosts. If you move a live grid into or out of a shadow root, destroy the instance and create it again.
+- Handsontable checks the container's root node only when the instance is created. The check controls the `ht-shadow-dom` class, the shadow-root clipboard listeners that copy and paste rely on in sandboxed hosts, and the shadow-root mouse listeners that the comment tooltip uses to react to hovering. If you move a live grid into or out of a shadow root, destroy the instance and create it again.
 - The cell editor doesn't leave the grid's stacking context. In a tight layout, an editor that overflows the grid's wrapper can paint under the host page content that creates its own stacking context.
 
 ## Related guides
--- a/docs/content/guides/tools-and-building/shadow-dom/shadow-dom.md
+++ b/docs/content/guides/tools-and-building/shadow-dom/shadow-dom.md
@@ -94,7 +94,7 @@ Salesforce Lightning Experience renders Lightning Web Components with its synthe
 
 ## Known limitations
 
-- Handsontable checks the container's root node only when the instance is created. The check controls the `ht-shadow-dom` class, the shadow-root clipboard listeners that copy and paste rely on in sandboxed hosts, and the shadow-root mouse listeners that the comment tooltip uses to react to hovering. If you move a live grid into or out of a shadow root, destroy the instance and create it again.
+- Handsontable checks the container's root node only when the instance is created. That check controls three things: the `ht-shadow-dom` class, the shadow-root clipboard listeners that copy and paste rely on in sandboxed hosts, and the shadow-root mouse listeners the comment tooltip uses to react to hovering. If you move a live grid into or out of a shadow root, destroy the instance and create it again.
 - The cell editor doesn't leave the grid's stacking context. In a tight layout, an editor that overflows the grid's wrapper can paint under the host page content that creates its own stacking context.
 
 ## Related guides
--- a/handsontable/src/plugins/comments/comments.ts
+++ b/handsontable/src/plugins/comments/comments.ts
@@ -6,7 +6,6 @@ import {
   hasClass,
   hasVerticalScrollbar,
   hasHorizontalScrollbar,
-  isChildOf,
   isShadowRoot,
   outerHeight,
 } from '../../helpers/dom/element';
@@ -532,19 +531,23 @@ export class Comments extends BasePlugin {
   }
 
   /**
-   * Checks whether the element sits in a tree the grid can be rendered into. `isChildOf()`
-   * walks `parentNode` only, so an element inside the grid's shadow tree dead-ends at the
-   * `ShadowRoot` (whose `parentNode` is `null`) and never reaches the document.
+   * Checks whether the element is attached to the document the grid renders into. A
+   * `parentNode` walk cannot answer this: it dead-ends at the first `ShadowRoot` (whose
+   * `parentNode` is `null`), which makes every element inside a shadow tree look detached.
+   * The composed root node crosses shadow boundaries by following host elements, so it
+   * resolves to the document for the grid's own shadow tree, for any other component's
+   * shadow tree on the page, and for the light DOM alike - while a genuinely detached
+   * element still resolves to its own orphaned root.
+   *
+   * Testing only the grid's own shadow root here would leave the tooltip open when the
+   * pointer crosses straight from a commented cell into a *different* shadow tree, which
+   * is an ordinary move on a page built from web components.
    *
    * @param {HTMLElement} element The element to check.
    * @returns {boolean}
    */
   #isInRenderedTree(element: HTMLElement): boolean {
-    if (this.#gridShadowRoot?.contains(element)) {
-      return true;
-    }
-
-    return isChildOf(element, this.hot.rootDocument);
+    return element.getRootNode({ composed: true }) === this.hot.rootDocument;
   }
 
   /**
@@ -552,10 +555,11 @@ export class Comments extends BasePlugin {
    *
    * @private
    * @param {Event} event DOM event.
+   * @param {HTMLElement} [target] The already-resolved event target, to avoid resolving it twice.
    * @returns {boolean}
    */
-  targetIsCellWithComment(event: Event) {
-    const closestCell = closest(this.#resolveEventTarget(event)!, ['TD']);
+  targetIsCellWithComment(event: Event, target = this.#resolveEventTarget(event)) {
+    const closestCell = closest(target!, ['TD']);
 
     return !!(closestCell && hasClass(closestCell, 'htCommentCell') &&
       closest(closestCell, [this.hot.rootElement]));
@@ -566,10 +570,11 @@ export class Comments extends BasePlugin {
    *
    * @private
    * @param {Event} event DOM event.
+   * @param {HTMLElement} [target] The already-resolved event target, to avoid resolving it twice.
    * @returns {boolean}
    */
-  targetIsCommentTextArea(event: Event) {
-    return this.getEditorInputElement() === this.#resolveEventTarget(event);
+  targetIsCommentTextArea(event: Event, target = this.#resolveEventTarget(event)) {
+    return this.getEditorInputElement() === target;
   }
 
   /**
@@ -929,8 +934,10 @@ export class Comments extends BasePlugin {
       return;
     }
 
-    if (!this.#preventEditorAutoSwitch && !this.targetIsCommentTextArea(event)) {
-      const eventCell = closest(this.#resolveEventTarget(event)!, ['TD']);
+    const target = this.#resolveEventTarget(event);
+
+    if (!this.#preventEditorAutoSwitch && !this.targetIsCommentTextArea(event, target)) {
+      const eventCell = closest(target!, ['TD']);
       let coordinates = null;
 
       if (eventCell) {
@@ -973,7 +980,7 @@ export class Comments extends BasePlugin {
     this.#cellBelowCursor = (this.#gridShadowRoot ?? rootDocument).elementFromPoint(
       (event as MouseEvent).clientX, (event as MouseEvent).clientY);
 
-    if (this.targetIsCellWithComment(event)) {
+    if (this.targetIsCellWithComment(event, target)) {
       const coords = this.hot.getCoords(target);
 
       if (coords) {
@@ -982,7 +989,7 @@ export class Comments extends BasePlugin {
         this.#displaySwitch?.show(range);
       }
 
-    } else if (this.#isInRenderedTree(target) && !this.targetIsCommentTextArea(event)) {
+    } else if (this.#isInRenderedTree(target) && !this.targetIsCommentTextArea(event, target)) {
       this.#displaySwitch?.hide();
     }
   };
--- a/handsontable/src/plugins/comments/comments.ts
+++ b/handsontable/src/plugins/comments/comments.ts
@@ -286,6 +286,14 @@ export class Comments extends BasePlugin {
    * @type {ShadowRoot|null}
    */
   #gridShadowRoot: ShadowRoot | null = null;
+  /**
+   * Mouse events already handled for a shadow-hosted grid, so that the document listener
+   * does not process one the shadow-root listener has taken. Only ever populated when the
+   * grid renders inside a shadow root.
+   *
+   * @type {WeakSet}
+   */
+  #processedMouseEvents: WeakSet<object> = new WeakSet();
 
   /**
    * Checks if the plugin is enabled in the handsontable settings. This method is executed in {@link Hooks#beforeInit}
@@ -451,22 +459,40 @@ export class Comments extends BasePlugin {
 
     this.#gridShadowRoot = isShadowRoot(rootNode) ? rootNode : null;
 
-    this.eventManager.addEventListener(rootDocument, 'mouseover', this.#onMouseOver);
-    this.eventManager.addEventListener(rootDocument, 'mousedown', this.#onMouseDown);
-    this.eventManager.addEventListener(rootDocument, 'mouseup', () => this.#onMouseUp());
+    // Claims an event for the first listener that receives it. A shadow-hosted grid binds the
+    // handlers twice, and the two bindings do NOT see the same element under a sandboxed host
+    // (e.g. Salesforce Lightning Web Security): that host collapses `composedPath()` to the
+    // shadow host chain, so only a listener bound inside the grid's own shadow tree gets the
+    // real cell, while the document listener still sees the host. Left undeduped, one hover
+    // would show from the shadow-root listener and then hide from the document listener -
+    // which also clears the display switch's flag, so the debounced show is dropped and the
+    // tooltip never appears. The shadow-root listener runs first (the event reaches the
+    // ShadowRoot before it crosses to the host), so it wins for anything inside the grid and
+    // the document listener keeps handling only what never entered the shadow tree.
+    const dedupe = (handler: (event: Event) => void) => (event: Event) => {
+      if (this.#gridShadowRoot) {
+        if (this.#processedMouseEvents.has(event)) {
+          return;
+        }
 
-    // When the grid lives inside a Shadow DOM tree, the same listeners are attached to the
-    // grid's shadow root as well. The document listeners above stay - they are the ones that
-    // see a click or a hover landing outside the shadow host. Sandboxed hosts (e.g. Salesforce
-    // Lightning Web Security) collapse `composedPath()` to the shadow host chain, so the
-    // document listeners cannot recover the real cell there; listeners bound inside the grid's
-    // own shadow tree still receive a correctly retargeted `target`. No dedupe registry is
-    // needed when both fire for one event: all three handlers resolve the same element and are
-    // idempotent, and showing is debounced by the display switch.
+        this.#processedMouseEvents.add(event);
+      }
+
+      handler(event);
+    };
+
+    const onMouseOver = dedupe(this.#onMouseOver);
+    const onMouseDown = dedupe(this.#onMouseDown);
+    const onMouseUp = dedupe(() => this.#onMouseUp());
+
+    this.eventManager.addEventListener(rootDocument, 'mouseover', onMouseOver);
+    this.eventManager.addEventListener(rootDocument, 'mousedown', onMouseDown);
+    this.eventManager.addEventListener(rootDocument, 'mouseup', onMouseUp);
+
     if (this.#gridShadowRoot) {
-      this.eventManager.addEventListener(this.#gridShadowRoot, 'mouseover', this.#onMouseOver);
-      this.eventManager.addEventListener(this.#gridShadowRoot, 'mousedown', this.#onMouseDown);
-      this.eventManager.addEventListener(this.#gridShadowRoot, 'mouseup', () => this.#onMouseUp());
+      this.eventManager.addEventListener(this.#gridShadowRoot, 'mouseover', onMouseOver);
+      this.eventManager.addEventListener(this.#gridShadowRoot, 'mousedown', onMouseDown);
+      this.eventManager.addEventListener(this.#gridShadowRoot, 'mouseup', onMouseUp);
     }
 
     if (editorElement) {
--- a/handsontable/src/plugins/comments/comments.ts
+++ b/handsontable/src/plugins/comments/comments.ts
@@ -468,11 +468,10 @@ export class Comments extends BasePlugin {
     const isShadowHosted = this.#gridShadowRoot !== null;
 
     // Claims an event for the first listener that receives it. A shadow-hosted grid binds the
-    // handlers twice, and the two bindings do NOT see the same element under a sandboxed host
-    // (e.g. Salesforce Lightning Web Security): that host collapses `composedPath()` to the
-    // shadow host chain, so only a listener bound inside the grid's own shadow tree gets the
-    // real cell, while the document listener still sees the host. Left undeduped, one hover
-    // would show from the shadow-root listener and then hide from the document listener -
+    // handlers twice, and the two bindings never see the same element: a listener inside the
+    // tree gets the real cell, while the document listener gets the retargeted shadow host.
+    // That is intrinsic to retargeting, not specific to a sandboxed host. Left undeduped, one
+    // hover would show from the shadow-root listener and then hide from the document listener -
     // which also clears the display switch's flag, so the debounced show is dropped and the
     // tooltip never appears. The shadow-root listener runs first (the event reaches the
     // ShadowRoot before it crosses to the host), so it wins for anything inside the grid and
--- a/handsontable/src/plugins/comments/comments.ts
+++ b/handsontable/src/plugins/comments/comments.ts
@@ -7,6 +7,7 @@ import {
   hasVerticalScrollbar,
   hasHorizontalScrollbar,
   isChildOf,
+  isShadowRoot,
   outerHeight,
 } from '../../helpers/dom/element';
 import { stopImmediatePropagation } from '../../helpers/dom/event';
@@ -278,6 +279,14 @@ export class Comments extends BasePlugin {
    * @type {string}
    */
   #commentValueBeforeSave = '';
+  /**
+   * The shadow root the grid renders inside, or `null` when it renders in the light DOM.
+   * Listeners bound on the document sit outside that tree, so the browser retargets the
+   * event to the shadow host and the real cell has to be recovered from the composed path.
+   *
+   * @type {ShadowRoot|null}
+   */
+  #gridShadowRoot: ShadowRoot | null = null;
 
   /**
    * Checks if the plugin is enabled in the handsontable settings. This method is executed in {@link Hooks#beforeInit}
@@ -439,10 +448,27 @@ export class Comments extends BasePlugin {
   registerListeners() {
     const { rootDocument } = this.hot;
     const editorElement = this.getEditorInputElement();
+    const rootNode = this.hot.rootElement.getRootNode();
+
+    this.#gridShadowRoot = isShadowRoot(rootNode) ? rootNode : null;
 
     this.eventManager.addEventListener(rootDocument, 'mouseover', this.#onMouseOver);
     this.eventManager.addEventListener(rootDocument, 'mousedown', this.#onMouseDown);
     this.eventManager.addEventListener(rootDocument, 'mouseup', () => this.#onMouseUp());
+
+    // When the grid lives inside a Shadow DOM tree, the same listeners are attached to the
+    // grid's shadow root as well. The document listeners above stay - they are the ones that
+    // see a click or a hover landing outside the shadow host. Sandboxed hosts (e.g. Salesforce
+    // Lightning Web Security) collapse `composedPath()` to the shadow host chain, so the
+    // document listeners cannot recover the real cell there; listeners bound inside the grid's
+    // own shadow tree still receive a correctly retargeted `target`. No dedupe registry is
+    // needed when both fire for one event: all three handlers resolve the same element and are
+    // idempotent, and showing is debounced by the display switch.
+    if (this.#gridShadowRoot) {
+      this.eventManager.addEventListener(this.#gridShadowRoot, 'mouseover', this.#onMouseOver);
+      this.eventManager.addEventListener(this.#gridShadowRoot, 'mousedown', this.#onMouseDown);
+      this.eventManager.addEventListener(this.#gridShadowRoot, 'mouseup', () => this.#onMouseUp());
+    }
 
     if (editorElement) {
       this.eventManager.addEventListener(editorElement, 'focus', () => this.#onEditorFocus());
@@ -478,6 +504,50 @@ export class Comments extends BasePlugin {
   }
 
   /**
+   * Resolves the element the event originated from. A path that crosses a shadow boundary
+   * carries `ShadowRoot` entries and its initial element is the true source - the cell itself,
+   * where `event.target` has been retargeted to the shadow host for a listener bound on the
+   * document. Without such entries there is no shadow tree to pierce, or a sandboxed host
+   * (e.g. Salesforce Lightning Web Security) collapsed the path to the shadow host chain; in
+   * both cases the retargeted `target` is already the deepest reliable reference.
+   *
+   * Grids in the light DOM skip the composed path entirely - `mouseover` fires on every
+   * pointer move, and `composedPath()` allocates an array on each call.
+   *
+   * @param {Event} event DOM event.
+   * @returns {HTMLElement|null} The deepest known source element.
+   */
+  #resolveEventTarget(event: Event): HTMLElement | null {
+    if (this.#gridShadowRoot === null) {
+      return eventTargetEl(event);
+    }
+
+    const eventPath = event.composedPath();
+
+    if (eventPath.some(entry => isShadowRoot(entry))) {
+      return eventPath[0] as HTMLElement;
+    }
+
+    return eventTargetEl(event);
+  }
+
+  /**
+   * Checks whether the element sits in a tree the grid can be rendered into. `isChildOf()`
+   * walks `parentNode` only, so an element inside the grid's shadow tree dead-ends at the
+   * `ShadowRoot` (whose `parentNode` is `null`) and never reaches the document.
+   *
+   * @param {HTMLElement} element The element to check.
+   * @returns {boolean}
+   */
+  #isInRenderedTree(element: HTMLElement): boolean {
+    if (this.#gridShadowRoot?.contains(element)) {
+      return true;
+    }
+
+    return isChildOf(element, this.hot.rootDocument);
+  }
+
+  /**
    * Checks if the event target is a cell containing a comment.
    *
    * @private
@@ -485,7 +555,7 @@ export class Comments extends BasePlugin {
    * @returns {boolean}
    */
   targetIsCellWithComment(event: Event) {
-    const closestCell = closest(eventTargetEl(event)!, ['TD']);
+    const closestCell = closest(this.#resolveEventTarget(event)!, ['TD']);
 
     return !!(closestCell && hasClass(closestCell, 'htCommentCell') &&
       closest(closestCell, [this.hot.rootElement]));
@@ -499,7 +569,7 @@ export class Comments extends BasePlugin {
    * @returns {boolean}
    */
   targetIsCommentTextArea(event: Event) {
-    return this.getEditorInputElement() === event.target;
+    return this.getEditorInputElement() === this.#resolveEventTarget(event);
   }
 
   /**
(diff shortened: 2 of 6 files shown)


qunabu:
Task linked: [DEV-2596 Fix the comment tooltip not opening on hover inside Shadow DOM (#8624)](https://app.clickup.com/t/9015210959/DEV-2596)

handsontable/src/plugins/comments/comments.ts
```diff
@@ -278,6 +278,22 @@ export class Comments extends BasePlugin {
...
+   */
+  #gridShadowRoot: ShadowRoot | null = null;
+  /**
+   * Mouse events already handled for a shadow-hosted grid, so that the document listener
+   * does not process one the shadow-root listener has taken. Only ever populated when the
+   * grid renders inside a shadow root.
+   *
+   * @type {WeakSet}
+   */
+  #processedMouseEvents: WeakSet<object> = new WeakSet();
```
demtario, line 296: **Blocking (description, not code).** This registry is what the PR description says is unnecessary: "No dedupe registry is needed when both listeners fire for one event: all three handlers resolve the same element and are idempotent, and showing is debounced by the display switch."

That claim is false at HEAD, and the code comment above `dedupe` says so correctly. I verified the mechanism: `DisplaySwitch.hide()` sets `wasLastActionShow = false`, which makes the already-queued debounced show a no-op, so the show-then-hide really does swallow the tooltip. The registry is load-bearing.

Please update the Cause section so a reviewer is not told the mechanism the fix depends on is unnecessary.

handsontable/src/plugins/comments/comments.ts
```diff
@@ -439,10 +455,45 @@ export class Comments extends BasePlugin {
...
+    // (e.g. Salesforce Lightning Web Security): that host collapses `composedPath()` to the
+    // shadow host chain, so only a listener bound inside the grid's own shadow tree gets the
+    // real cell, while the document listener still sees the host. Left undeduped, one hover
+    // would show from the shadow-root listener and then hide from the document listener -
+    // which also clears the display switch's flag, so the debounced show is dropped and the
+    // tooltip never appears. The shadow-root listener runs first (the event reaches the
+    // ShadowRoot before it crosses to the host), so it wins for anything inside the grid and
+    // the document listener keeps handling only what never entered the shadow tree.
+    const dedupe = (handler: (event: Event) => void) => (event: Event) => {
+      if (this.#gridShadowRoot) {
```
demtario: **Medium.** This reads `this.#gridShadowRoot` at event time rather than bind time, which implicitly couples the field to the dual binding. Nulling the field while both listeners are still attached turns the dedupe into a pass-through and reintroduces exactly the show-then-hide the WeakSet exists to prevent. It is safe today only because `BasePlugin.disablePlugin()` calls `eventManager.clear()` first.

Capturing the decision in the closure removes the coupling:

```ts
const isShadowHosted = this.#gridShadowRoot !== null;

const dedupe = (handler: (event: Event) => void) => (event: Event) => {
  if (isShadowHosted) {
    // ...
  }
  handler(event);
};
```

Worth noting the two copies of this pattern also differ in shape without a stated reason: `copyPaste.ts` dedupes unconditionally, this one is guarded.

handsontable/src/plugins/comments/comments.ts
```diff
@@ -439,10 +455,45 @@ export class Comments extends BasePlugin {
   registerListeners() {
     const { rootDocument } = this.hot;
     const editorElement = this.getEditorInputElement();
+    const rootNode = this.hot.rootElement.getRootNode();
+
+    this.#gridShadowRoot = isShadowRoot(rootNode) ? rootNode : null;
```
demtario, line 466: **Medium.** This line, not the shadow-root listener, is where the coverage caveat belongs.

The description flags the LWS shadow-root-listener path as untested. But every new path in this change hangs off this gate, and `isShadowRoot` tests `nodeType === DOCUMENT_FRAGMENT_NODE && 'host' in node`. The fixture uses a native shadow root, so the gate is trivially true in all four new tests and can never be exercised as false. If LWS's synthetic root fails that predicate, `#gridShadowRoot` stays `null` and the entire change (dedupe, `#resolveEventTarget`, the `elementFromPoint` swap) goes inert with the sandboxed-host test still green.

Restating the caveat at the gate would tell a reviewer what green actually buys.

Separately, and Low: this is a one-shot capture, and `updateSettings` never re-runs `registerListeners()`. `BasePlugin.onUpdateSettings` only calls `enablePlugin()` when the enabled state flips, otherwise it routes to `updatePlugin()`, which here only touches the display delay. A wrong capture is unrecoverable short of toggling `comments` off and on. Construction timing is fine (`rootElement` is inside `rootContainer` before `beforeInit` fires), and `copyPaste.ts:289` has the identical one-shot capture, so this is a note rather than an ask. If it is ever addressed, do both plugins together.

handsontable/src/plugins/comments/comments.ts
```diff
@@ -477,15 +528,64 @@ export class Comments extends BasePlugin {
...
+   * (e.g. Salesforce Lightning Web Security) collapsed the path to the shadow host chain; in
+   * both cases the retargeted `target` is already the deepest reliable reference.
+   *
+   * Grids in the light DOM skip the composed path entirely - `mouseover` fires on every
+   * pointer move, and `composedPath()` allocates an array on each call.
+   *
+   * @param {Event} event DOM event.
+   * @returns {HTMLElement|null} The deepest known source element.
+   */
+  #resolveEventTarget(event: Event): HTMLElement | null {
```
demtario: **Medium: second verbatim copy of the pattern.** `copyPaste.ts` already contains all four pieces this adds:

- `#processedClipboardEvents` WeakSet (`copyPaste.ts:276-283`)
- a `dedupe` closure of the same shape (same lines)
- dual document plus shadow-root binding (`copyPaste.ts:284-300`)
- `#resolveClipboardEventTarget` (`copyPaste.ts:756-766`), whose body is the identical `some(isShadowRoot) ? path[0] : target`

`isShadowRoot(...getRootNode())` is now at four sites (`core.ts:396`, `copyPaste.ts:297`, `tableView.ts:533`, here), and the point-reader pattern at two (`tableView.ts:534` and below).

`handsontable/AGENTS.md` names `src/helpers/dom/element.ts` as the home for exactly these primitives (`isShadowRoot`, `getShadowHostChain`, `getDeepActiveElement`), so a second private copy runs against the stated convention. Extracting `resolveComposedEventTarget(event, shadowRoot)` plus a point-reader helper would collapse both copies and, usefully, make the logic unit-testable next to the existing `isShadowRoot` cases in `element.unit.ts`.

One caveat if you do extract it: `some(isShadowRoot)` actually answers "is this path complete, or did a sandboxed host filter it", which is how `tableView.ts:1575` uses it. It does not answer "does this path cross *my* boundary". That reading happens to hold here, so name the helper for what it tests.

handsontable/src/plugins/comments/comments.ts
```diff
@@ -496,10 +596,11 @@ export class Comments extends BasePlugin {
    *
    * @private
    * @param {Event} event DOM event.
+   * @param {HTMLElement} [target] The already-resolved event target, to avoid resolving it twice.
    * @returns {boolean}
    */
-  targetIsCommentTextArea(event: Event) {
-    return this.getEditorInputElement() === event.target;
+  targetIsCommentTextArea(event: Event, target = this.#resolveEventTarget(event)) {
```
demtario: **Medium: this new argument is unverified at both call sites** (`#onMouseDown` and the hide branch of `#onMouseOver`). Nothing in the diff hovers or clicks the tooltip, so the `target` argument could be deleted at both sites and every test would still pass.

That is the tooltip's primary interaction: hover the cell, then move onto the textarea to read or edit. The blocker is the fixture rather than the test, see my note on `ShadowGridPage.ts`.

Minor, and no action needed: after this change `event` is vestigial in both method bodies, surviving only to feed the default initializer, so the two parameters can silently disagree. Both callers are internal, so it is a signature-shape remark only.

handsontable/src/plugins/comments/comments.ts
```diff
@@ -859,8 +960,10 @@ export class Comments extends BasePlugin {
       return;
     }
 
-    if (!this.#preventEditorAutoSwitch && !this.targetIsCommentTextArea(event)) {
-      const eventCell = closest(eventTargetEl(event)!, ['TD']);
+    const target = this.#resolveEventTarget(event);
```
demtario: **Low: untested behavior change.** Before the fix, a shadow-hosted grid always resolved the host here, `closest(host, ['TD'])` was `null`, and `#onMouseDown` therefore called `hide()` on *every* mousedown. Now it resolves the real TD, so clicking the cell an open comment belongs to no longer closes the editor.

That is a user-visible improvement and it is worth one case: open the editor via the context menu's "Add comment", click inside the same cell, assert the editor is still visible.

handsontable/src/plugins/comments/comments.ts
```diff
@@ -891,17 +994,19 @@ export class Comments extends BasePlugin {
   #onMouseOver = (event: Event) => {
     const { rootDocument } = this.hot;
 
-    const target = eventTargetEl(event)!;
+    const target = this.#resolveEventTarget(event)!;
```
demtario: **Low (performance).** The target is resolved before the three guards below that do not depend on it (`#preventEditorAutoSwitch`, `#editor?.isFocused()`, `!this.#editor`), so a shadow-hosted grid allocates a `composedPath()` array on every `mouseover` those guards discard. `mouseover` fires on every pointer move and is the hottest handler in this plugin.

Checking those three first, then resolving the target for the `hasClass` and `#cellBelowCursor` comparisons, keeps the early-out allocation-free. The light-DOM path is unaffected either way.

handsontable/src/plugins/comments/comments.ts
```diff
@@ -891,17 +994,19 @@ export class Comments extends BasePlugin {
...
 
     if (this.#preventEditorAutoSwitch || this.#editor?.isFocused() || hasClass(target, 'wtBorder')
         || this.#cellBelowCursor === target || !this.#editor) {
       return;
     }
 
-    this.#cellBelowCursor = rootDocument.elementFromPoint(
+    // `elementFromPoint()` does not pierce shadow boundaries, so on a document it resolves to
+    // the shadow host. The grid's own shadow root resolves the cell actually under the cursor.
+    this.#cellBelowCursor = (this.#gridShadowRoot ?? rootDocument).elementFromPoint(
```
demtario, line 984: **Low.** This is correct, but it restores only a short-circuit that no test can observe, and the comment above reads as though it fixes something visible.

`#cellBelowCursor` feeds nothing except the `target ===` early return further up. Revert this line to `rootDocument` and all four new tests stay green, because the fixture's renderer leaves each TD with a single text node so `target` is always the TD, and Playwright's `hover()` emits one `mouseover` per cell so the guard never fires anyway.

Either say in the comment that this is hygiene with no observable behavior delta, or add a case if you know a flow where the dead guard is observable. Not worth a test that only re-executes the line.

tests/fixtures/pages/ShadowGridPage.ts
```diff
@@ -27,6 +30,12 @@ export class ShadowGridPage {
     this.outsideInput = page.getByTestId('outside-input');
     this.shadowSibling = page.getByTestId('shadow-sibling');
     this.focusMover = page.getByTestId('focus-mover');
+    // The comment editor is portaled into `document.body`, so it sits in the light DOM
+    // even though its grid lives inside the shadow root.
+    this.commentTooltip = page.locator('.htComments');
```
demtario, line 35: **Medium: the fixture leaves this element unstyled, which caps what the tooltip tests can assert.**

`shadow-dom.html` appends both stylesheets into the shadow root only, and `<head>` carries nothing but a body-font rule. Since `rootPortalElement` is appended to `rootDocument.body` (`core.ts:416`), as your comment correctly notes, `.htComments { position: absolute }` never applies and `setPosition`'s inline left/top are inert. The tooltip renders in normal document flow at the bottom of the page.

This is not a false green: show and hide are an inline `display` toggle rather than a class, so `toBeVisible` and `toBeHidden` are sound. But two things follow:

1. Tooltip placement regressions are invisible here.
2. "Move the pointer onto the tooltip" is physically unreachable, which is why `targetIsCommentTextArea`'s new argument has no coverage.

Adding the same two `<link>` elements to `<head>` fixes both, and then the hover-onto-textarea case becomes writable. Worth re-checking the 9 existing tests afterward, since the context menu then becomes a real positioned element; I expect it fine because its items are located by exact text, but confirm.

tests/fixtures/pages/ShadowGridPage.ts
```diff
@@ -83,4 +92,13 @@ export class ShadowGridPage {
   async outsideClickTargets(): Promise<string[]> {
     return this.page.evaluate(() => (window as any).__hotProbe.outsideClickTargets());
   }
+
+  /**
+   * Filter `composedPath()` on mouse events down to the shadow host chain, the way Salesforce
+   * Lightning Web Security presents them to sandboxed code (fixture probe). Call before the
+   * interaction under test.
+   */
+  async simulateSandboxedHost(): Promise<void> {
```
demtario: **Low.** Two things about this probe.

It has no restore, and its reach exceeds its name. It patches `Event.prototype.composedPath` for all `MouseEvent`s, including the core's own uses at `tableView.ts:452` and `:512`, where a filtered path makes the `eventPath.includes(rootWrapperElement)` check fail and classifies in-grid clicks as *outside* clicks. Harmless today because the only caller hovers, but it is a trap for whoever next combines the probe with a click and gets a mysterious deselection.

A `restoreNativeComposedPath()` counterpart plus one line here ("affects core outside-click detection, hover-only flows") would close it.

tests/e2e/shadow-dom.spec.ts
```diff
@@ -102,6 +102,67 @@ test.describe('grid inside a native shadow root', () => {
     expect(await grid.outsideClickTargets()).toEqual(['focus-mover']);
   });
 
+  /**
+   * Issue #8624. The Comments plugin listens on the document, which sits outside the
+   * shadow tree, so `event.target` arrives retargeted to the shadow host and the hover
+   * never resolved to the cell. `showAtCell()` was unaffected, which is why the tooltip
+   * could be opened programmatically while hovering did nothing.
+   */
+  test('opens the comment tooltip when a commented cell is hovered', async () => {
```
demtario, line 111: Worth correcting in the description: it says the spec "gains two cases" and reports "11 passed (2 new + 9 pre-existing)", but four cases land here, so the number should be 13.

For what it is worth, I traced all four and each is load-bearing rather than redundant:

1. this one fails unfixed, because `event.target` is the retargeted host, `closest(host, ['TD'])` is `null`, and the else branch hides;
2. the plain-cell hide case pins fix (3) specifically, failing against a "resolve the target but keep `isChildOf`" partial fix;
3. the foreign-shadow-tree case is strictly stronger than (2), killing a containment check written as "own shadow root or document";
4. the sandboxed-host case genuinely pins the dedupe, since without it the shadow listener shows and the document listener then hides and clears `wasLastActionShow`.

So the suite is well constructed. It is only the stated count and the stated both-directions run that are behind HEAD.

docs/content/guides/tools-and-building/shadow-dom/shadow-dom.md
```diff
@@ -94,7 +94,7 @@ Salesforce Lightning Experience renders Lightning Web Components with its synthe
 
 ## Known limitations
 
-- Handsontable checks the container's root node only when the instance is created. The check controls the `ht-shadow-dom` class and the shadow-root clipboard listeners that copy and paste rely on in sandboxed hosts. If you move a live grid into or out of a shadow root, destroy the instance and create it again.
+- Handsontable checks the container's root node only when the instance is created. The check controls the `ht-shadow-dom` class, the shadow-root clipboard listeners that copy and paste rely on in sandboxed hosts, and the shadow-root mouse listeners that the comment tooltip uses to react to hovering. If you move a live grid into or out of a shadow root, destroy the instance and create it again.
```
demtario: **Low (docs style).** This sentence now carries a three-item list inside it, which runs against the short-sentences rule in `.ai/DOC-STANDARDS.md`. Splitting the enumeration out reads better, for example:

> Handsontable checks the container's root node only when the instance is created. That check controls three things: the `ht-shadow-dom` class, the shadow-root clipboard listeners that copy and paste rely on in sandboxed hosts, and the shadow-root mouse listeners the comment tooltip uses to react to hovering. If you move a live grid into or out of a shadow root, destroy the instance and create it again.

demtario (commented):
Reviewed the whole diff plus the surrounding plugin, `copyPaste`, `tableView`, `displaySwitch`, and the shadow DOM helpers. **No Critical, no High.** The three reads plus the dedupe are internally consistent, and I traced all four new Playwright cases as load-bearing: each one kills a distinct mutant of the change.

One blocking item, which is about the description rather than the code, plus four Medium and six Low findings left inline.

## Blocking: the description contradicts HEAD, and the dedupe has no stated verification

The Cause section says:

> No dedupe registry is needed when both listeners fire for one event: all three handlers resolve the same element and are idempotent, and showing is debounced by the display switch.

`comments.ts` now adds exactly that registry, and `95a61776d` is titled "Dedupe mouse events across the dual shadow-root binding". The code comment above `dedupe` explains correctly why it is needed, so the description is still describing an earlier tree. I confirmed the registry is load-bearing rather than defensive: `DisplaySwitch.hide()` sets `wasLastActionShow = false`, which makes the queued debounced show a no-op.

The test evidence is stale the same way. The description says "gains two cases" and "11 passed (2 new + 9 pre-existing)"; the diff adds four cases, so the number should be 13.

Mapping the stated both-directions run against the commits, it covers the tree through `33f6f59c2`:

| commit | content | inside the stated verification? |
|---|---|---|
| `6d334a107` | src: composed-path target | yes |
| `ad5a84014` | tests: 2 cases | yes |
| `d9f01c213` / `670123673` | docs / changelog | n/a |
| `33f6f59c2` | src: `#isInRenderedTree` | yes |
| `aa61feb07` | fixture: 2nd shadow host + test | no |
| `95a61776d` | src: the dedupe | no |
| `92af23d10` | test: sandboxed host | no |

So the riskiest part of the change, a guard that silently suppresses handler invocations on a hot path, is the part with no stated evidence. Could you re-run both directions at HEAD and correct the Cause section?

## Checked and cleared

Recording these so the inline list does not read as the whole review:

- `rootElement` and `rootContainer` cannot resolve to different root nodes: `core.ts:364` makes `rootElement` a detached div for root instances, appended under `rootContainer` at `core.ts:411`, all inside the Core constructor and therefore before `enablePlugin` runs on `beforeInit`.
- `closest()` hops `ShadowRoot` to `host`, so a deep non-TD `composedPath()[0]` (nested shadow root, SVG icon, a custom element in a renderer) still resolves to the TD, and `getCoords` delegates to it. `closest` is also null-tolerant, so `targetIsCellWithComment(event, null)` returns `false` rather than throwing despite the `!`.
- `isChildOf` really is a plain `parentNode` walk with no shadow hop, so fix (3)'s rationale holds. I traced every `#isInRenderedTree` divergence: shadow-tree element (old `false`, new `true`, the intended fix), detached (`false` both), iframe (`false` both), `documentElement` (`true` both), and `target === rootDocument` (differs, but unreachable via pointer events).
- `extendEvent` in `eventManager.ts` mutates and returns the same object, so the WeakSet key is stable across both bindings. A cloning wrapper there would have silently disabled the dedupe with no visible symptom.
- Bubble ordering holds: both bindings are bubble-phase and a `ShadowRoot` sits strictly before its host in the path. Closed shadow roots, slotted light-DOM content, nested shadow roots, two grids in one shadow root, and a forced double `registerListeners()` all trace clear.
- `ShadowRoot.elementFromPoint` is fine on the `browser-targets.js` floor: `tableView.ts:533-534` already ships the identical `ShadowRoot | Document` union on develop.
- `.changelogs/8624.json` is valid. 8624 is a real public issue, so `issuesOrigin: "public"` and the filename are both right, and all six fields are present.
- No breaking change: both widened methods are `@private` with no callers outside `comments.ts`, and an optional trailing param is additive.
- Fixture blast radius on the 9 existing tests is clear. `#other-shadow-host` is appended after `#outside-area`, so nothing shifts; `htCommentCell` adds only an `::after` triangle from a shadow-scoped stylesheet; B4 and A4 are untouched elsewhere. The context-menu test's unscoped `.ht_master .htCore tbody tr` locator does also match menu item rows, and comments adds four entries, but `close()` destroys the menu instance synchronously on a leaf click, so the retrying `toHaveCount(6)` converges as before.
- No prototype-patch leak between tests: the per-test `page` fixture plus a navigating `beforeEach` gives a fresh realm.
- Nothing dangling on disable or destroy. `BasePlugin.disablePlugin()` clears the event manager, so the stale `#gridShadowRoot` is unreachable, and `#processedMouseEvents` is a WeakSet.
- Nothing in the diff hits `no-restricted-syntax`, and E2E is the right test kind for DOM-event behavior, so the presence gate is satisfied.
- The new JSDoc uses ASCII hyphens where `.ai/DOC-STANDARDS.md` rule 11 asks for en dashes, but `copyPaste.ts` has 17 ASCII separators and zero en dashes, so consistency argues for leaving it as is.

handsontable/src/plugins/comments/comments.ts
budnix: You're right, the description was still describing the pre-dedupe tree. Rewrote the Cause section and the test evidence, and replaced the stale both-directions claim with a per-mechanism mutation table run at HEAD.

handsontable/src/plugins/comments/comments.ts
budnix: Good point — captured it as `isShadowHosted` in the closure in `076cb64`.

handsontable/src/plugins/comments/comments.ts
budnix: Fair point, the caveat belongs at the gate. Moved it there in `076cb64`, including that a synthetic root failing the predicate makes the whole change inert with the sandboxed-host case still green.

handsontable/src/plugins/comments/comments.ts
budnix: Good spot, and it resolved in a better direction: the composed-path resolution turned out to kill no test, so `076cb64` deletes it instead of extracting it. Happy to collapse the two `isShadowRoot(...getRootNode())` sites in a separate cleanup if you still want that.

handsontable/src/plugins/comments/comments.ts
budnix: Yep, my mistake — chasing this showed the argument was dead weight, so `076cb64` removes it along with the composed path, and `b388fd9` covers the pointer moving onto the tooltip.

handsontable/src/plugins/comments/comments.ts
budnix: Agreed — added that case in `b388fd9`.

handsontable/src/plugins/comments/comments.ts
budnix: Right, though it went away with the composed path — resolving is just `event.target` again, so `076cb64` puts the original guard block back and there is nothing to allocate.

handsontable/src/plugins/comments/comments.ts
budnix: True, it only restores a short-circuit nothing can observe. The comment says that plainly now in `076cb64`.

tests/fixtures/pages/ShadowGridPage.ts
budnix: Nice catch — `b388fd9` links both stylesheets in `<head>` too, which is what made the hover-onto-tooltip case writable. All nine existing cases still pass, the context menu one included.

tests/fixtures/pages/ShadowGridPage.ts
budnix: Fair enough — `b388fd9` adds a `restoreNativeComposedPath()` counterpart and notes that the patch also reaches the core outside-click detection, so keep it to hover-only flows.

tests/e2e/shadow-dom.spec.ts
budnix: Thanks — the count is corrected in the description, and it is 15 now rather than 13. One correction to your point 3 though: after dropping the composed path the foreign-tree case no longer pins the containment form on its own, so I am not claiming it does.

docs/content/guides/tools-and-building/shadow-dom/shadow-dom.md
budnix: Good call — split it using your wording in `2fc7604`.

handsontable/src/plugins/comments/comments.ts
demtario: Deleting the layer instead of extracting it is the better outcome, and it removes one of the four pieces I listed. Leaving this thread open because the other three are still there, and dropping the resolver arguably strengthens the case rather than weakening it.

What still exists in both plugins:

- the `dedupe` closure, same shape (`copyPaste.ts:276-283`)
- the WeakSet keyed on the event (same lines)
- the dual document plus shadow-root binding (`copyPaste.ts:284-300`)

Plus `isShadowRoot(...getRootNode())` at four sites: `core.ts:396`, `copyPaste.ts:297`, `tableView.ts:533`, and `comments.ts:466`.

Before this revision the duplicated resolver was one mechanism among two overlapping ones. Now the dedupe plus dual binding *is* the entire fix, in both plugins, written out twice. A shared `bindDeduped(eventManager, targets, type, handler)` or similar would carry the one invariant that matters (the inner listener claims the event, the outer one only handles what never entered the tree) in a single place with a single comment, instead of two comments that can drift apart. Which they already have, see my new note on the `dedupe` block above.

Happy for it to be a separate cleanup rather than this PR. Could you link the follow-up task here so the thread has somewhere to point? I will resolve it once that exists.

handsontable/src/plugins/comments/comments.ts
```diff
@@ -439,10 +455,53 @@ export class Comments extends BasePlugin {
...
+    // second binding, the dedupe and the point reader all inert - the grid then behaves
+    // exactly as it did before this fix. The Playwright fixture mounts a native shadow root,
+    // so the gate is always true there and its false side is not covered.
+    this.#gridShadowRoot = isShadowRoot(rootNode) ? rootNode : null;
+
+    const isShadowHosted = this.#gridShadowRoot !== null;
+
+    // Claims an event for the first listener that receives it. A shadow-hosted grid binds the
+    // handlers twice, and the two bindings do NOT see the same element under a sandboxed host
+    // (e.g. Salesforce Lightning Web Security): that host collapses `composedPath()` to the
```
demtario: This rationale did not survive `076cb64`, and the same wording is in body item 2.

The plugin no longer calls `composedPath()` anywhere; the only occurrence left in the file is this comment. So LWS collapsing the path is not what makes the two bindings disagree any more. They disagree under a plain native open shadow root too, because the document listener's `event.target` is retargeted to the host, which is exactly the retargeting the Cause section opens with. The dedupe is therefore load-bearing for *every* shadow-hosted grid, not only sandboxed ones.

Your own mutation table is the proof: removing the dedupe kills **6** cases, and all six run against the native shadow root in the fixture. If the need were LWS-specific it would kill one.

Suggested shape for the first half, keeping your second half as is:

> Claims an event for the first listener that receives it. A shadow-hosted grid binds the handlers twice, and the two bindings never see the same element: a listener inside the tree gets the real cell, while the document listener gets the retargeted shadow host. That is intrinsic to retargeting, not specific to a sandboxed host. Left undeduped, one hover would show from the shadow-root listener and then hide from the document listener, which also clears the display switch's flag, so the debounced show is dropped and the tooltip never appears.

Worth fixing in both places, since a future reader who trusts the LWS framing could conclude the dedupe is droppable for native embedding.

demtario (commented):
Two follow-ups from the simplification in `076cb64`. Both are consequences of the composed-path layer going away, so neither could have been spotted before it did. Everything else from my first pass is addressed and resolved.

To be clear about the direction: dropping that layer was the right call, and the per-mechanism mutation table is a better piece of evidence than the both-directions run it replaced. These two notes are about the prose and the test scaffolding that layer left behind, not about the fix.

tests/e2e/shadow-dom.spec.ts
```diff
@@ -102,6 +102,106 @@ test.describe('grid inside a native shadow root', () => {
...
+    await expect(grid.commentTooltip).toBeHidden();
+  });
+
+  /**
+   * A shadow-hosted grid binds the mouse handlers on both the document and its shadow root.
+   * Under a sandboxed host the two do not agree: `composedPath()` is filtered to the host
+   * chain, so only the shadow-root listener can reach the cell while the document listener
+   * still sees the host. Undeduped, one hover showed from the first and hid from the second,
+   * and the hide cleared the display switch's flag so the queued show was dropped - the
+   * tooltip never appeared. Native embedding cannot catch this: there both listeners resolve
```
demtario: This sentence is now false, and it makes the case above it vestigial.

Native embedding *does* catch this. Both listeners do not resolve the same cell natively: the document listener resolves the shadow host, because the plugin now reads `event.target` directly and no longer consults `composedPath()`. Your mutation table agrees, removing the dedupe kills 6 cases rather than only this one.

Which leaves the probe with nothing to do. `simulateSandboxedHost()` filters `composedPath()`, the plugin never calls `composedPath()`, so the patch cannot change plugin behavior. This case passes for the same reason `opens the comment tooltip when a commented cell is hovered` passes, and it would pass with the probe removed. That also makes the `restoreNativeComposedPath()` counterpart I asked for in the other thread dead weight, so please do not keep it on my account.

Two honest options:

1. Delete the case, `simulateSandboxedHost`, `restoreNativeComposedPath`, and the two POM wrappers. The dedupe is already pinned six ways over.
2. Keep them, but relabel: it is a cheap guard that a filtered composed path does not break the fix, which matters only if composed-path handling ever comes back. Say that, and drop the "native embedding cannot catch this" claim.

I lean towards (1), on the grounds that a probe that patches a global prototype and reaches the core's outside-click detection should have to earn its keep. Your call, since (2) is defensible if you expect LWS work to land here later.

handsontable/src/plugins/comments/comments.ts
budnix: You're right, that rationale outlived the code — the only `composedPath()` left in the file was the comment itself. Rewrote it with your wording in `daa09ab`, and fixed the same framing in the description.

tests/e2e/shadow-dom.spec.ts
budnix: Good spot, and I checked it before deciding: with only the probe call removed the case still passed, so it was inert. Took your option 1 in `ab5f71d` — the case, both probes and the two wrappers are gone, restore included.

demtario (approved):
Approving. The fix is in good shape, and it got materially better through review rather than just patched: the composed-path layer went away once mutation testing showed it killed nothing, the dedupe rationale is now stated correctly, and the sandboxed-host probe was removed rather than relabelled once it turned out to be inert. The description ended up honest about what is covered and what is not, which is the part that will still be useful in a year.

Re-verified at `ab5f71d04`:

- `composedPath()` is gone from the plugin, so the fix is plain retargeting plus the dedupe, and the two remaining reads (`#isInRenderedTree`, the point reader) are each pinned or documented as hygiene.
- No dangling probe references anywhere in `tests/`, and the `<head>` stylesheet links stayed, so the hover-onto-tooltip case still means something.
- The mutation table was recounted after the removal (`remove the dedupe` 6 → 5) rather than left stale, and the file's 14 cases match the 14 in the description.
- Light-DOM behavior is unchanged: everything hangs off the one `isShadowRoot()` gate, and the hover path for a light-DOM grid is byte-for-byte `develop`'s.

**One open thread, non-blocking:** https://github.com/handsontable/handsontable/pull/13227#discussion_r3842773502

The dedupe closure, the WeakSet, and the dual document plus shadow-root binding now exist twice, in `comments.ts` (`:296`, `:479-485`, `:500-502`) and `copyPaste.ts` (`:201`, `:277-281`, `:298-300`), and `getRootNode()` feeds an `isShadowRoot` check at four sites (`core.ts:396`, `tableView.ts:533`, `copyPaste.ts:289`, `comments.ts:458`). We agreed that belongs in a separate cleanup rather than here, so it is explicitly not a condition of this approval. I have left the thread open only so the follow-up has somewhere to point; link the task there whenever you create it and I will resolve it.

The one thing I could not check is CI, since the E2E legs were still pending when I approved and the legacy Jasmine comments suite does not run in a worktree. Worth a glance at the `-min` and horizon/classic legs before you squash, given the fixture now links stylesheets in the light DOM as well.

Issue #10800, opened 2024-02-22
[Bug]: when i use 'custom renderer as a component'. Sometimes, the Component Elements will be automatically cleared


FF-zhy:
### Describe the bug

for example: 
after edit one cell, all custom render create Elements  will be cleared. (small amount data, the effect can be seen through Chrome DevTools debugger)

if has many custom cells ,there will be a noticeable flicker.

why not updating TD  through React render ?

### Video/Screenshots

init:
![Snipaste_2024-02-22_11-36-02](https://github.com/handsontable/handsontable/assets/43714228/bf940be2-153e-4c65-8497-2921b4d6e9ce)
afterEdit:
![Snipaste_2024-02-22_11-22-18](https://github.com/handsontable/handsontable/assets/43714228/a7b0f6fe-13d2-48f0-9e0e-c919d84be0d4)

### Provide a link to the demo with the bug reproduction

[...]

Issue #7544, opened 2020-07-26
Dynamic columns not working when use hot-column


wensonsmith:
### Description

it should add  a column when push a column setting into hotColumns array. 

but it seems nothing to work

```
<template>
  <hot-table :settings="hotSettings">
    <hot-column v-for="col in hotColumns" :data="col.data" :title="col.title" />
  </hot-table>
</template>
```

### Demo

demo: https://codesandbox.io/embed/affectionate-cray-gnnqk?fontsize=14&hidenavigation=1&theme=dark

click add column 

### Your environment

Handsontable: v7.4.2 (19/02/2020 13:41:19) Wrapper: v4.1.1 Vue: v2.6.10

ahndohun:
@AMBudnik 
Will the problem be resolved?

AMBudnik:
Hi @ahndohun 
yes, it will. However, it did not get on our list of priorities yet.
I will keep you informed.

vincent-lu:
This is a painful problem. My current solution is to apply `:key` value on wrapper outside of `<hot-table>`.

juaiglesias:
> This is a painful problem. My current solution is to apply `:key` value on wrapper outside of `<hot-table>`.

@vincent-lu We are facing the same issue over here, could you give us an example of how you solve it please?

vincent-lu:
Hi @juaiglesias the rough setup we have is:

Have a AppTable component like
```
<template>
  <hot-table ref="table" :settings="props.settings" />
</template>
<script setup>
...
</script>
```

and when using this <app-table> component:
```
<app-table :key="tableKey" :settings="FULL_HOT_TABLE_SETTINGS"></app-table>
```

For example the FULL_HOT_TABLE_SETTINGS could be as basic as:
```
{
  columns: [
    { data: "sku" },
    { data: "brand" },
    { data: "width" },
    { data: "height" },
    { data: "length" },
    { data: "weight" },
  ]
}
```
And the `tableKey` value can be anything really, as long as it changes when your `settings.columns` changes. You could even just use `JSON.stringify(settings.columns)` if your columns settings are not too crazy.

Hope it helps.

akinoccc:
any update?

adrianszymanski89:
Hi @akinoccc 

Not yet. We only have the workaround proposed by @vincent-lu. We will update you on this topic as soon as the issue is fixed on our side.

demtario:
[...]

PR #13228, opened 2026-08-24
DEV-2597: Give each table its own portal container for component renderers (#9063)

### LINKED ISSUE #9063: When using component columns with dynamic height with fixedColumnsLeft option, row alignment is not displayed correctly
### Description

If you use a vue component column with dynamic height with the fixedColumnsLeft option, the alignment of the rows is not displayed correctly. Is there any way to solve this?

Enabling the autoRowSize option seems to solve it, but it gives me a warning message. I also saw the instructions to disable the option in the documentation.

> Handsontable's autoRowSize and autoColumnSize options require calculating the widths/heights of some of the cells before rendering them into the table. For this reason, it's not currently possible to use them alongside component-based renderers, as they're created after the table's initialization.
> 
> Be sure to turn those options off in your Handsontable config, as keeping them enabled may cause unexpected results. Please note that autoColumnSize is enabled by default.

### Steps to reproduce

1. Insert a multi-row Vue component into a hot-column.
2. use the fixedColumnsLeft option.

### Demo

https://jsfiddle.net/9zLx5esa/4/

### Your environment
* Vue wrapper version: 11.0.1
* Handsontable version: 11.0.1
* Browser Name and version: chrome
* Operating System: mac os

budnix:
### Context

The PR fixes [#9063](https://github.com/handsontable/handsontable/issues/9063): with a component-based renderer whose cell height depends on its content, the rows of a frozen column stop lining up with the rest of the grid.

Handsontable draws a cell of a frozen column twice — once in the master table and once in the inline-start overlay — and each draw gets its own `TD`. The React wrapper cached one portal container per cell, keyed by `<instance guid>-<row>-<col>` in `hotTableContext.tsx` (`getRendererWrapper`). Nothing in that key said which of the two tables the cell belonged to, so both draws shared one container `DIV`, and whichever table rendered last physically took it. The other table's cell was left **empty**, so the two tables sized their rows from different content and drifted apart.

Measured on `develop` before the fix (12 rows, a renderer drawing 1–3 lines per row, `fixedColumnsStart: 2`, `autoRowSize: false`): master rows all 29 px, overlay rows 32/53/75 px, worst row-top drift **242 px**, unchanged after scrolling. The master's cell for row 2 had `innerHTML === ""` while the overlay's held the three rendered lines. A control run with an equivalent plain function renderer gave pixel-identical heights in both tables and **0 px** drift, which ruled out the core and the theme.

The fix puts the table into the portal-container cache key and into the React portal key, so every table gets its own container. The in-place container reuse added for #10800 (renderer components must not remount on every grid render) and the viewport-bounded eviction in `pushCellPortalsIntoPortalManager` are unchanged.

**Behavior note:** for a cell in a frozen column, the renderer component now mounts once per table rather than once in total. That matches what function renderers already do — the renderer is called once per draw — but a component with side effects in its body or in an effect will now run them for each table the cell appears in.

**What this PR does not fix.** When the tall component column is *not* in the frozen area, the overlay still misaligns (measured: same 242 px). That is a different mechanism and it is the long-standing async-measurement problem described in [this comment](https://github.com/handsontable/handsontable/issues/9063#issuecomment-1352739965) and tracked as [#7544](https://github.com/handsontable/handsontable/issues/7544): portals are flushed in `afterViewRender` through a React state setter, so Handsontable's row measurement runs while the React cells are still empty. The same control run shows the core side is fine — a function renderer in a scrollable column aligns the row headers exactly — so the remaining work is to re-measure after the portals mount. That needs `flushSync` plus a guarded extra render pass on every draw that mounts new rows, which is a render-path performance decision for every React user, so it is deliberately left out of this bug fix.

The Vue 3 wrapper has its own renderer mechanism and was not tested; it needs a separate check.

### How has this been tested?

New jsdom regression spec, `wrappers/react-wrapper/test/fixedColumnsPortals.spec.tsx`, asserting that the master table and the inline-start overlay each hold their own copy of the rendered content for the same frozen cell, and that a re-render does not hand one table's container to the other. Both cases were confirmed to **fail** before the fix (`Expected: "rendered-by-component:0-0"`, `Received: ""` — the empty cell), and pass after it.

```bash
npm run test --prefix wrappers/react-wrapper                                  # 16 suites, 77 tests pass
npm run test --prefix wrappers/react-wrapper -- --testPathPattern=fixedColumnsPortals
```

The two existing portal-cache eviction specs pass unchanged — worth noting because this key change adds a second cache entry per frozen cell, and those specs pin the cache to a viewport-sized bound.

Real-browser alignment was verified manually, since there is no React wrapper harness in the Playwright suite yet (all `tests/e2e` page objects load the core UMD bundles). With the built wrapper in Chromium, per-row heights and row-top drift were read from the DOM: `fixedColumnsStart: 2` with the renderer in a frozen column went from 242 px drift to **0 px**, and stayed at 0 px after scrolling down and back up.

### Types of changes

### Related issue(s):

1. DEV-2597
2. #9063
3. Related, not fixed here: #7544

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2597

---

> [!NOTE]
> **Medium Risk**
> Touches the React wrapper’s cell-portal cache, which can remount renderers extra times in frozen areas. Logic is localized and covered by regression tests; no auth or data-handling changes.
> 
> **Overview**
> Fixes frozen-row/column misalignment when using React component renderers: Handsontable draws the same cell in the master table and overlay tables, but portal containers were keyed only by instance + coordinates, so one table stole the container and left the other cell empty.
> 
> Portal cache keys now include a stable per-table id (from the cell’s `TBODY`), so each overlay gets its own container. In-place reuse and viewport eviction are unchanged. Frozen cells now mount the renderer once per table (matching function renderers).
> 
> Adds a regression spec covering fixed columns, fixed rows, and re-render. Does not address overlay misalignment for tall components outside the frozen area (#7544).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1fbb949443f37cc8f8dac67cbba6b9f3ecfd4c1d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changelogs/9063.json
+++ b/.changelogs/9063.json
@@ -1,6 +1,6 @@
 {
   "issuesOrigin": "public",
-  "title": "Fixed a bug where a component-based renderer used in a frozen column rendered its content into only one of the two tables the grid draws that column in, so the cell looked empty in the other one and the frozen columns stopped lining up with the rest of the rows.",
+  "title": "Fixed a bug where a component-based renderer in a frozen row or column left the cell empty in one of the two tables the grid draws it in, which made the frozen area stop lining up with the rest of the rows.",
   "type": "fixed",
   "issueOrPR": 9063,
   "breaking": false,
--- a/.changelogs/9063.json
+++ b/.changelogs/9063.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed a bug where a component-based renderer used in a frozen column rendered its content into only one of the two tables the grid draws that column in, so the cell looked empty in the other one and the frozen columns stopped lining up with the rest of the rows.",
+  "type": "fixed",
+  "issueOrPR": 9063,
+  "breaking": false,
+  "framework": "react"
+}
--- a/wrappers/react-wrapper/src/hotTableContext.tsx
+++ b/wrappers/react-wrapper/src/hotTableContext.tsx
@@ -110,6 +110,37 @@ const HotTableContextProvider: FC<PropsWithChildren> = ({ children }) => {
   }, []);
   const getPortalContainerCacheSize = useCallback(() => portalContainerCache.current.size, []);
 
+  // Stable id per table element the grid draws into. Handsontable renders the
+  // same cell once per table it appears in — the master table plus an overlay
+  // table for every fixed row/column edge — and each draw gets its own TD.
+  const tableIds = useRef<WeakMap<Element, number>>(new WeakMap());
+  const nextTableId = useRef(0);
+
+  /**
+   * Identify the table a rendered cell belongs to, so the master table and the
+   * overlay tables get separate portal containers for the same cell.
+   *
+   * @param {HTMLTableCellElement} TD The cell being rendered.
+   * @returns {Number} The id of the table the cell belongs to.
+   */
+  const getTableId = useCallback((TD: HTMLTableCellElement): number => {
+    // Overlay tables are built once and never hand their cells to another
+    // table, so the TABLE element is a stable identity for the draw.
+    const table: Element = TD.closest('table') ?? TD;
+    const cachedId = tableIds.current.get(table);
+
+    if (cachedId !== undefined) {
+      return cachedId;
+    }
+
+    const id = nextTableId.current;
+
+    nextTableId.current += 1;
+    tableIds.current.set(table, id);
+
+    return id;
+  }, []);
+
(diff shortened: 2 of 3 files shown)


qunabu:
Task linked: [DEV-2597 Fix row misalignment for component-based renderers with fixedColumnsStart (#9063)](https://app.clickup.com/t/9015210959/DEV-2597)

adrianspdev (approved):

PR #13231, opened 2026-08-24
DEV-2601: Stop a killed test run from flooding the agent context

Haxikowy:
### Context

The PR fixes a Stop hook that reported passing tests as failing and pasted a megabyte of test output into the agent's context.

`scripts/claude/stop.mjs` re-runs the unit tests a session touched, through `spawnSync`, whose default output cap is 1 MB. A grid-instantiating Jest suite prints far more than that (jsdom logs the whole injected stylesheet on every CSS parse failure — 4.86 MB from one 50-test suite), so the buffer overflowed, Node killed Jest mid-run, and the result came back with a non-zero status and no test summary. That is indistinguishable from a real failure unless the caller inspects `error`. The hook reported "Unit tests you touched are failing" on a suite that was 50/50 green, and pasted the whole truncated buffer back to the agent: ~386k tokens, injected once per turn end until the session was interrupted. Hook feedback lands as a conversation message, so the cost is re-paid on every later request in that session rather than being a one-off spike.

Three independent fixes, any one of which breaks the chain:

- `TEST_RUN_MAX_BUFFER` (64 MB) on every hook-spawned test run, so the run completes and prints the summary the classifiers depend on.
- `isSpawnInfraFailure()` — a child killed by ENOBUFS or a signal produced no verdict and must not block, the same way a Jest that could not start already did not. Trade-off: a genuinely failing run whose output overflows no longer blocks at the hook. Acceptable — pre-push and CI still catch it, and CI is authoritative.
- `condenseTestOutput()` — hook feedback is never worth a megabyte; the diagnosis is a dozen lines. Stack frames and CSS dumps are stripped, repeats collapsed to `(xN)`, the excerpt anchored at the first failure so the diagnosis survives, and capped at 120 lines / 8 KB. Worst case a Stop block can inject drops from 1 MB to 8 KB.

`scripts/pre-push.mjs` gets the buffer and the spawn-failure guard too. It prints to a terminal, so it keeps writing full output.

**Scope note.** The 4.86 MB of jsdom noise that overflowed the buffer has its own root cause — `ThemeManager` emits the runtime theme stylesheet using CSS nesting, which jsdom cannot parse and which is above the declared browser floor. That is deliberately **not** fixed here; it is tracked separately. This PR makes the hook immune to noisy output regardless of where the noise comes from.

This PR touches `scripts/` only, so the path-aware changelog gate passes automatically and no changelog entry is included.

### How has this been tested?

- New unit coverage in `scripts/__tests__/pre-push.test.mjs` for the three helpers — spawn-infra classification, the buffer floor, and condensing (bounded excerpt, noise stripped, summary preserved, failure-anchored, repeats collapsed). Verified red before the fix.

- Review fixes are covered too (`8d1f884`): the excerpt anchor now takes the earliest marker that *names* a failing test, so a console flood between the `FAIL` header and the `●` block no longer costs the diagnosis; `dropped` counts from the raw line total; SGR escapes are stripped before matching. Each verified red before its fix. The ESLint spawn in `post-tool-use.mjs` got the same buffer and condensing treatment, and `.ai/LOCAL-ENFORCEMENT.md` documents the killed-child infra case, the condensing contract, and the stderr audience.

  ```bash
  node --test "scripts/__tests__/*.test.mjs"    # 38 passing
  ```

- Reproduced the original overflow against a real suite: the run produces 4.86 MB, exceeding the 1 MB default and returning `status != 0` with no summary. With `TEST_RUN_MAX_BUFFER` the same run returns `status 0` and prints `Tests: 50 passed, 50 total`, so the false alarm never forms.

- Repo lint and the full core suite were run on the branch while the (now separated) theme fix was present: `npm run eslint --prefix handsontable` (0 errors), `npm run stylelint --prefix handsontable`, `npm run test:unit --prefix handsontable` (3734 passing, 312 suites), `npm run test:types --prefix handsontable`.

### Types of changes

### Related issue(s):

1. DEV-2601

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

(Repository tooling only — no shipped package is affected.)

### Checklist:

ClickUp task: https://app.clickup.com/t/86cb9a3k0

---

> [!NOTE]
> **Medium Risk**
> Changes local enforcement hook behavior (spawn overflow no longer blocks; very noisy failing runs may slip past hooks), but scope is tooling only and CI remains authoritative.
> 
> **Overview**
> Fixes the Stop hook falsely reporting green unit tests as failing and dumping megabytes of output into the agent session when Jest’s console noise exceeds Node’s default **1 MB** `spawnSync` buffer.
> 
> **`scripts/pre-push.mjs`** adds **`TEST_RUN_MAX_BUFFER` (64 MB)** on hook-spawned runs, **`isSpawnInfraFailure()`** so ENOBUFS/signals/`status: null` warn/skip instead of blocking, and **`condenseTestOutput()`** to strip stack/CSS noise, collapse repeats, anchor on the first real failure (Jest or Playwright), and cap excerpts (~120 lines / 8 KB for agent hooks). Pre-push still prints full output up to **`TERMINAL_OUTPUT_LIMIT`**, then condenses for the terminal.
> 
> **`scripts/claude/stop.mjs`** and **`post-tool-use.mjs`** use the same buffer and condensing for Playwright/Jest failures and ESLint reports. **`LOCAL-ENFORCEMENT.md`** documents killed-run handling and agent output bounds.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8d1f884afb06c60a667a99c63f62659989ecaec3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/scripts/claude/stop.mjs
+++ b/scripts/claude/stop.mjs
@@ -15,7 +15,15 @@ import { readFileSync, existsSync } from 'node:fs';
 import path from 'node:path';
 import { isNewJasmineSpec } from '../../.github/scripts/lib/presence-gate.mjs';
 import { repoRoot, sessionEditsFile, stopVerdict, toRepoRelative } from './session.mjs';
-import { changedPlaywrightSpecs, changedUnitTests, unitTestPattern, isJestInfraFailure } from '../pre-push.mjs';
+import {
+  changedPlaywrightSpecs,
+  changedUnitTests,
+  unitTestPattern,
+  isJestInfraFailure,
+  isSpawnInfraFailure,
+  condenseTestOutput,
+  TEST_RUN_MAX_BUFFER,
+} from '../pre-push.mjs';
 import { filterCached, recordGreen } from '../e2e-run-cache.mjs';
 
 // npx/npm are .cmd shims on Windows; spawnSync needs a shell there or it ENOENTs.
@@ -128,16 +136,29 @@ if (toRun.length > 0) {
   // theme matrix (main/horizon/classic). See the run-scope note in the
   // handsontable-playwright-e2e skill.
   const pw = spawnSync('npx', ['playwright', 'test', '--project=e2e-main', ...toRun], {
-    cwd: path.join(root, 'tests'), stdio: ['ignore', 'pipe', 'pipe'], encoding: 'utf8', shell: WIN,
+    cwd: path.join(root, 'tests'),
+    stdio: ['ignore', 'pipe', 'pipe'],
+    encoding: 'utf8',
+    shell: WIN,
+    maxBuffer: TEST_RUN_MAX_BUFFER,
   });
 
-  if (pw.status !== 0) {
+  if (pw.status !== 0 && isSpawnInfraFailure(pw)) {
+    // No verdict, so nothing to block on — but the unit checks below are
+    // independent and must still run. Do NOT exit here.
     process.stderr.write(
-      `Playwright specs you touched are failing — fix them before finishing:\n${pw.stdout || ''}${pw.stderr || ''}`,
+      `Note: the Playwright run for ${toRun.join(', ')} could not complete (${
+        pw.error?.code || pw.signal}). Not treated as a test failure — verify it yourself.\n`,
+    );
+  } else if (pw.status !== 0) {
+    process.stderr.write(
+      `Playwright specs you touched are failing — fix them before finishing:\n${
+        condenseTestOutput(`${pw.stdout || ''}${pw.stderr || ''}`)}\n`,
     );
     process.exit(2);
+  } else {
+    recordGreen(root, toRun);
   }
-  recordGreen(root, toRun);
 }
 
 // Run any changed Jest unit test the session touched — fast (jest maps to src,
@@ -150,15 +171,32 @@ for (const file of unitFiles) {
   const jest = spawnSync(
     'npm',
     ['run', 'test:unit', '--', `--testPathPattern=${unitTestPattern(file)}`],
-    { cwd: path.join(root, 'handsontable'), encoding: 'utf8', shell: WIN },
+    {
+      cwd: path.join(root, 'handsontable'),
+      encoding: 'utf8',
+      shell: WIN,
+      maxBuffer: TEST_RUN_MAX_BUFFER,
+    },
   );
 
   if (jest.status !== 0) {
-    if (isJestInfraFailure(`${jest.stdout || ''}${jest.stderr || ''}`)) {
+    const output = `${jest.stdout || ''}${jest.stderr || ''}`;
+
+    if (isSpawnInfraFailure(jest)) {
+      // This one file produced no verdict; the others are independent runs, so
+      // keep checking them rather than abandoning the loop.
+      process.stderr.write(
+        `Note: the unit run for ${file} could not complete (${
+          jest.error?.code || jest.signal}). Not treated as a test failure — verify it yourself.\n`,
+      );
+      continue;
+    }
+
+    if (isJestInfraFailure(output)) {
       break;
     }
     process.stderr.write(
-      `Unit tests you touched are failing — fix them before finishing:\n${jest.stdout || ''}${jest.stderr || ''}`,
+      `Unit tests you touched are failing — fix them before finishing:\n${condenseTestOutput(output)}\n`,
     );
     process.exit(2);
   }
--- a/scripts/claude/stop.mjs
+++ b/scripts/claude/stop.mjs
@@ -146,6 +146,12 @@ if (toRun.length > 0) {
   if (pw.status !== 0 && isSpawnInfraFailure(pw)) {
     // No verdict, so nothing to block on — but the unit checks below are
     // independent and must still run. Do NOT exit here.
+    //
+    // Audience caveat: Claude Code forwards a hook's stderr to the agent only on
+    // exit 2. This leg does not exit, so the note reaches the agent only if a
+    // later leg blocks in the same run (stderr accumulates) — otherwise it lands
+    // in the debug log. Best-effort by design: nothing may depend on the agent
+    // reading it. The same holds for the unit-loop note below.
     process.stderr.write(
       `Note: the Playwright run for ${toRun.join(', ')} could not complete (${
         pw.error?.code || pw.signal}). Not treated as a test failure — verify it yourself.\n`,
@@ -184,7 +190,8 @@ for (const file of unitFiles) {
 
     if (isSpawnInfraFailure(jest)) {
       // This one file produced no verdict; the others are independent runs, so
-      // keep checking them rather than abandoning the loop.
+      // keep checking them rather than abandoning the loop. As above, this note
+      // only reaches the agent if a later leg exits 2 — do not rely on it.
       process.stderr.write(
         `Note: the unit run for ${file} could not complete (${
           jest.error?.code || jest.signal}). Not treated as a test failure — verify it yourself.\n`,
--- a/scripts/pre-push.mjs
+++ b/scripts/pre-push.mjs
@@ -113,6 +113,17 @@ export const TEST_RUN_MAX_BUFFER = 64 * 1024 * 1024;
 const MAX_LINE_LENGTH = 200;
 
 /**
+ * SGR (color) escape sequences, stripped before anything is matched. A test run
+ * inherits the hook's environment, so a leaked `FORCE_COLOR` would wrap every
+ * failure marker in escape codes and silently defeat the anchor and summary
+ * patterns — the excerpt would stay bounded but stop pointing at the diagnosis.
+ *
+ * @type {RegExp}
+ */
+// eslint-disable-next-line no-control-regex -- ESC is exactly what this matches.
+const ANSI_SGR = /\u001b\[[0-9;]*m/g;
+
+/**
  * Output size past which the pre-push gate condenses a run rather than printing
  * it whole. A terminal tolerates far more than a condensed excerpt, but not the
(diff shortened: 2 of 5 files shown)


qunabu:
Task linked: [DEV-2601 Stop hook reports green tests as failing and floods the agent context](https://app.clickup.com/t/9015210959/DEV-2601)

scripts/pre-push.mjs
```diff
@@ -93,6 +93,139 @@ export function isJestInfraFailure(output) {
...
+    .map(({ line, count }) => (count > 1 && line.trim() !== '' ? `${line}    (×${count})` : line))
+    .filter((line, index, all) => line.trim() !== '' || (all[index - 1] || '').trim() !== '');
+
+  const summary = rendered.filter(isSummary);
+  // Jest marks failures with FAIL/●/✕; Playwright's list reporter uses ✘ and
+  // numbers each failure (`1) spec.ts:3:1 › …`). Both run through this helper.
+  // The numbered form also requires Playwright's `›`, so an ordinary numbered
+  // line in someone's output cannot anchor the excerpt by accident.
+  const firstFailure = rendered.findIndex(line => /^\s*(FAIL\b|●|✕|✘)/.test(line)
+    || /^\s*\d+\)\s.*›/.test(line));
```
danielzytohoc: **The anchor keeps the noise and drops the diagnosis in the exact scenario this PR was built around.**

`findIndex` takes the **first** match of `FAIL|●|✕|✘` — and Jest prints buffered console output right after the `FAIL` header, *before* the test list and the `●` failure details. The theme stylesheet jsdom dumps is multi-line (`themes/engine/manager.ts` builds it with `\n`), so hundreds of plain `--ht-*: value;` lines survive `isNoise` (no `data:`, no leading `at`, no "Could not parse") and don't collapse (each line differs).

I probed with a faithful model — a failing test plus 50 jsdom dumps between `FAIL` and the `●` block: the excerpt came out as `FAIL` + 118 lines of CSS variables + summary. The `✕` name, the `●` block, and `Expected`/`Received` were all cut. The flood is still fixed (3.6 KB), but the agent is told "tests failing" with no which/why — and its likely next move is re-running the suite, the very thing the excerpt exists to avoid. Any console-noisy failing suite hits this, not just jsdom.

One-line fix — prioritize the anchor: first `●` (Jest puts the assertion there, and "Test suite failed to run" is also a `●`), else `✕`/`✘`/Playwright `N) … ›`, else `FAIL`:

```js
const anchorRes = [/^\s*●\s/, /^\s*(✕|✘)/, /^\s*\d+\)\s.*›/, /^\s*FAIL\b/];
const firstFailure = anchorRes.reduce((found, re) => (found === -1 ? rendered.findIndex(l => re.test(l)) : found), -1);
```

Verified against the same probe: 262 chars containing the full `●` block, `Expected`/`Received`, and the summary. (Optionally exclude `● Validation Warning:` / `● Deprecation Warning:`.)

Please also add a fixture with noise **between** the `FAIL` header and the `●` block — the existing anchor tests only place noise before the failure marker, which is why this slipped through.

scripts/pre-push.mjs
```diff
@@ -93,6 +93,139 @@ export function isJestInfraFailure(output) {
...
+  const anchored = (firstFailure === -1 ? rendered : rendered.slice(firstFailure)).filter(l => !isSummary(l));
+  const lineBudget = Math.max(1, maxLines - summary.length);
+  const kept = firstFailure === -1 ? anchored.slice(-lineBudget) : anchored.slice(0, lineBudget);
+  const dropped = (rendered.length - kept.length - summary.length) + (lines.length - collapsed.length);
```
danielzytohoc: Cosmetic: `dropped` is computed after the `isNoise` filter (`lines` is already filtered), so noise-dropped lines are never counted. In the flood probe the header reports "2090 noise/duplicate lines dropped" when ~2240 were removed. Either count from the pre-filter line count, or soften the header ("output condensed" without exact arithmetic).

scripts/claude/stop.mjs
```diff
@@ -128,16 +136,29 @@ if (toRun.length > 0) {
...
+    maxBuffer: TEST_RUN_MAX_BUFFER,
   });
 
-  if (pw.status !== 0) {
+  if (pw.status !== 0 && isSpawnInfraFailure(pw)) {
+    // No verdict, so nothing to block on — but the unit checks below are
+    // independent and must still run. Do NOT exit here.
     process.stderr.write(
-      `Playwright specs you touched are failing — fix them before finishing:\n${pw.stdout || ''}${pw.stderr || ''}`,
+      `Note: the Playwright run for ${toRun.join(', ')} could not complete (${
```
danielzytohoc, line 156: Heads-up on audience: this goes to stderr and the script (usually) exits 0 — Claude Code feeds stderr to the agent only on exit 2; on exit 0 it lands in debug logs. So "verify it yourself" reaches the agent only if a *later* leg blocks in the same run (stderr accumulates). Fine as best-effort, but worth a code comment (or rewording) so nobody relies on the agent seeing it. Same applies to the unit-loop note below.

danielzytohoc (commented):
**Verdict: this properly fixes DEV-2601.** I verified the failure chain and each defense layer against `839929e`, ran the suite, and probed the new helpers with adversarial inputs:

- `spawnSync` overflow semantics confirmed empirically on Node 22.23.1: `status: null`, `signal: 'SIGTERM'`, `error.code: 'ENOBUFS'`, truncated buffer — exactly the shape `isSpawnInfraFailure()` keys on.
- The caps in `condenseTestOutput()` are structural, not filter-dependent: 20 MB of filter-proof varied input condenses to ~8 KB. Worst-case Stop-hook injection drops from ~386k tokens to ~2k per firing.
- `recordGreen` is correctly skipped on no-verdict; the Playwright infra path no longer exits before the unit loop; per-file `continue` vs. config-level `break` is the right split; pre-push still prints full output under 256 KB.
- The pre-push Playwright leg uses `stdio: 'inherit'`, so it structurally cannot ENOBUFS — correctly left alone.
- Scripts-only changelog skip is correct, and keeping the CSS-nesting root cause out of scope is right.
- All 32 tests in `scripts/__tests__` pass; the new cases (slice(-0), anchor-vs-caps, long-line truncation) pin exactly the right regressions.

One substantive issue inline — the excerpt anchor loses the diagnosis in precisely the flood-plus-real-failure scenario (one-line fix, verified with a probe) — plus two smaller inline notes.

**Process note — the branch was force-pushed after review.** Commits `a4ddb0f`/`d1de581`/`9e4ab64` (reviewed by Bugbot 16:22–16:52) were squashed into `839929e` (17:04), so the reviewed SHAs are unreachable — that is why 3 of 4 Bugbot comments now dangle as outdated, and the description's "24 passing" no longer matches (32 now). AGENTS.md forbids force-pushing PR-bound branches; please land review fixes as follow-up commits — squash happens at merge anyway.

**Smaller notes without an inline anchor:**
- `scripts/claude/post-tool-use.mjs:67` has the same pattern one door down: the ESLint spawn sits on the 1 MB default and pastes `output` uncondensed on `status === 1`. ENOBUFS cannot false-block there (`status` would be null), but a huge lint output floods the same way. Follow-up-ticket sized.
- Cheap ANSI hardening: if `FORCE_COLOR` ever leaks into the hook env, escape codes silently defeat the anchor/summary regexes (the caps still hold — checked). A leading `.replace(/\x1b\[[0-9;]*m/g, '')` in `condenseTestOutput()` immunizes the parser.
- `.ai/LOCAL-ENFORCEMENT.md` documents the "infra failure warns instead of blocking" taxonomy and the "must not false-block" rule — the new no-verdict case (killed child) and the condensing contract belong there per checklist item 3.

scripts/pre-push.mjs
Haxikowy: Confirmed, fixed in `8d1f884`. Reproduced your probe.

Didn't take the priority list as written — `●` winning outright breaks twice: Jest opens the console dump with `● Console`, so the `●` tier anchors back at the flood (your probe lacked that header), and the `✕` line sits right before `●`, so it gets dropped by one line — two existing tests catch that.

Landed instead: earliest marker that *names* a failing test (`●`/`✕`/`✘`/Playwright `N) … ›`), excluding `● Console` and Validation/Deprecation warnings, `FAIL` as fallback only. `● Test suite failed to run` still counts.

Added your fixture (noise between `FAIL` and `●`, with `● Console`) plus cases for the fallback, config notices, and the infra bullet. Red before, green after.

scripts/pre-push.mjs
Haxikowy: Fixed — counts from the raw line total now, noise included. Kept the exact number. Test asserts 100 frames report as 100.

scripts/claude/stop.mjs
Haxikowy: Both non-blocking notes now carry the caveat: stderr reaches the agent only on exit 2, otherwise the debug log unless a later leg blocks. Nothing depends on it. Also in `.ai/LOCAL-ENFORCEMENT.md`.

Haxikowy:
- `post-tool-use.mjs:67` — fixed, not ticketed. Two lines: `TEST_RUN_MAX_BUFFER` on the spawn, `condenseTestOutput()` on the report.
- ANSI — stripped before matching. Test wraps markers in SGR, asserts anchor and summary still land.
- `.ai/LOCAL-ENFORCEMENT.md` — killed-child no-verdict case, condensing contract, stderr audience.
- Force-push: my mistake, won't repeat on this branch. `8d1f884` is a follow-up; reviewed SHAs stay reachable.

danielzytohoc (approved):

PR #10440, opened 2023-07-11
Optimize the transpiling build process

budnix:
### Context

The PR optimizes the Handsontable build process by updating browser targets and upgrading Webpack to v5 and more.

By doing the above things and removing polyfill for unsupported browsers, I'm able to decrease the weight of output files (comparing with v13.0.0) to something like:
| File          | Change           |
|--------------------------|-----------------------|
| `handsontable.js`          | save 19.95% (from 3 612 859 B to 2 891 888 B) |
| `handsontable.full.js`     | save 12.95% (from 5 894 227 B to 5 130 461 B) |
| `handsontable.min.js`      | save 22.45% (from  1 098 706 B to 852 000 B) |
| `handsontable.full.min.js` | save 14.22% (from 2 071 606 B to 1 777 009 B) |

For ESM and CommonJS files, the results should be similar.

#### Changes
1. Update [browsers targets](https://github.com/handsontable/handsontable/pull/10440/files#diff-09c56b2bf95de2a608a36afef3b6893146a959d6739be0a154dc9c9f02d80f24R3-R5);
2. Upgrade Webpack to the latest version (v5);
3. Add [Babel assumption](https://github.com/handsontable/handsontable/pull/10440/files#diff-209a3b608317f4f56ad4e098027392938a19384ed9689a690b2af4e2695f466aR26) that allows saving about 21kB;
4. Remove some polyfills for unsupported browsers;
5. [Fix the Babel configuration](https://github.com/handsontable/handsontable/pull/10440/files#diff-09c56b2bf95de2a608a36afef3b6893146a959d6739be0a154dc9c9f02d80f24R11) for CoreJS so some of the modern API wasn't cover by polyfills. For instance, this fixes `Array.prototype.at` [method usage](https://github.com/handsontable/handsontable/pull/10440/files#diff-368ba6a14a52bcc5b035ff7e0d425c1a7f5ba8034be9acc698695d5fbc0d1e67R163) which on Safari 14.1 causes syntax error (https://github.com/handsontable/handsontable/issues/10334);

### How has this been tested?

I tested the changes on different browsers, including Safari 14.1, Chrome 109, and Firefox 90, and I found no issues with UMD dist files.

The ESM and CommonJS files were tested using bundlers, and here are the results:
 * ✅ Parcel v1;
 * ✅ Parcel v2;
 * ✅ Webpack v4;
 * ✅ Webpack v5;
 * ✅ Angular stack for v12, and Angular v16;
 * ✅ React app stack;
[...]

PR #13232, opened 2026-08-24
DEV-2602: Raise the browser floor to the Baseline 2024 set

Haxikowy:
### Context

Raises the compile floor in `browser-targets.js` to the Baseline 2024 set, so the declared floor matches the platform features the library ships.

**This replaces the original direction of this PR.** The first attempt fixed CSS nesting in the runtime theme string and left `light-dark()` above the floor. The floor itself was the defect.

The theme stylesheets use `light-dark()` in 428 declarations. The old floor (Chrome 110 / Firefox 110 / Safari 14.1) sat below its support (Chrome 123 / Firefox 120 / Safari 17.5), so lightningcss lowered all of them into class-switched variable pairs. The same gap covered CSS nesting (Safari 17.2), which is why the runtime theme string dropped its icon rules. Minified theme CSS, built at both floors:

| Floor | `light-dark()` in output |
|---|---|
| Chrome 110 / Firefox 110 / Safari 14.1 | 0 |
| Chrome 130 / Edge 130 / Firefox 132 / Safari 18.2 | 428 |

#### The new floor

```js
exports.BROWSERS_LIST = [
  'Chrome >= 130',
  'Edge >= 130',
  'Firefox >= 132',
  'Safari >= 18.2',
  'iOS >= 18.2',
];
```

- Resolved once with `browserslist('baseline 2024')`, then pinned as integers — a moving query would make builds non-reproducible.
- Edge gets an entry for the first time; `Chrome >= 110` never covered it.
- iOS is listed separately from Safari: browserslist treats `ios_saf` as its own target, so `Safari` alone left lightningcss and `eslint-plugin-compat` with no iOS bound. Same WebKit, same number, no change to the emitted CSS. The `and_chr` and `and_ff` entries `browserslist('baseline 2024')` also returns stay out — browserslist carries only their latest version, so no floor for them can be pinned.
- Independent of the "two latest versions" statement in the supported-browsers guide. That is a lower bound on freshness, not a cap on how far back the floor may reach.
- No device-class drop: iOS 18 runs on the same hardware as iOS 17.
- `ES_TARGET` → `es2024`, with `tsconfig.json` `lib` following. `es2025` needs Chrome 136. `swc-transpile.mjs` keeps its `es2021` pin for Angular's Zone.js. Covered by the existing `esTarget.unit.js` invariants.

#### Rest of the diff

`no-restricted-syntax` is pruned to one entry: `with`, which this repo's `core-js-compat` data puts above the floor on two engines — `Array#with` at Firefox 140 (against 132) and `TypedArray#with` at Safari/iOS 26.0 (against 18.2). The rule message names both. The bans on `toSorted`/`toSpliced`/`toReversed`, `at`/`findLast`/`findLastIndex`, `Object.hasOwn` and `structuredClone` cited a floor we no longer compile for — that syntax is now usable in core source.

Everything else is comment corrections (`core.ts`, `autocompleteEditor.ts` and its spec, `licenseNotification.unit.js`) and the floor in the two `AGENTS.md` files. **Every source edit in this PR is a comment — no runtime behavior change.**

#### Kept deliberately

Workarounds the floor makes unnecessary, left in place because they still work and still cover stylesheets or engines below it:

- Theme engine: `#buildResolvedColorVariables()` and the `resolveScheme` / `lightDarkOnly` options — earlier releases shipped lowered `light-dark()`, and a consumer's own build can still produce it.
- `throwWithCause()`: keeps assigning `cause` after construction. An engine that accepts the options bag and drops it silently would break `error.cause?.handsontable === true` detection with no visible failure.
- License-branding CSS: keeps `overflow: hidden` and its `word-break` pair.
- `autocompleteEditor`: keeps `[...choices].sort()`. `updateChoicesList` is public API and the spread accepts an array-like, which `toSorted` would not.

### How has this been tested?

```bash
npm run typecheck --prefix handsontable                 # clean under es2024
npm run test:unit --prefix handsontable                 # 3732 passing, 312 suites
npm run stylelint --prefix handsontable                 # clean
npm run build --prefix handsontable                     # all bundles, new targets
npm run build:themes-css.min --prefix handsontable      # 428 light-dark() in output
```

`esTarget.unit.js` gains two tests: every floor must sit at or above `light-dark()` and CSS nesting per browser, and the feature table must cover every entry in `BROWSERS_LIST`. Lowering the floor past either feature now fails a test instead of silently re-lowering the CSS. Every other source change is a comment.

### Types of changes

**Not a breaking change.** The supported-browsers guide promises the two latest versions of each browser and names no version numbers. The new floor sits many majors below that window — Safari 18.2 against a current 26.x — so nothing the guide promises is dropped. No API, option, hook, or CSS class changed either, so there is no migration step and no migration-guide page.

On the major-release question: the rule in `browser-targets.js` and the root `AGENTS.md` — treat a floor raise as a support drop, major boundary and sign-off — is left exactly as written. It should stay strict, so the next person to touch the floor checks it against the guide's window rather than assuming. This raise is below that window, which is what makes it an ordinary change.

The supported-browsers guide is left alone for the same reason: it states the policy, and a second, much lower set of numbers beside it would read as the versions we support.

### Related issue(s):

1. DEV-2602

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2602

---

> [!NOTE]
> **Medium Risk**
> Changes distributed JS/CSS build targets and TypeScript lib, which alters emitted bundles and theme CSS for anyone below the new floor; mitigated by tests and intentional alignment with shipped CSS features.
> 
> **Overview**
> Raises the pinned compile/API floors in `browser-targets.js` from Chrome/Firefox 110 and Safari 14.1 to the **Baseline 2024** set (Chrome and Edge 130, Firefox 132, Safari and iOS 18.2), with `ES_TARGET` and `tsconfig.json` `lib` bumped to **es2024**. That aligns lightningcss and swc targets with the CSS the themes actually use (`light-dark()`, nesting, logical properties), so minified theme CSS keeps native `light-dark()` instead of class-switched fallbacks and nested icon rules are not stripped—fixing missing theme icons and wrong RTL corner rounding described in the changelog.
> 
> **Lint and docs** follow the new floor: `no-restricted-syntax` now only bans `.with()` (still above the floor on some engines); older bans on `toSorted`, `at`, `Object.hasOwn`, etc. are removed. `esTarget.unit.js` adds guards so any future floor cannot sit below required CSS feature versions.
> 
> Most other touched source files are comment-only clarifications; existing workarounds (`#buildResolvedColorVariables`, `throwWithCause`, autocomplete copy-and-sort) stay for older stylesheets and silent failure modes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b5df8a95f78ae6a69ed2fd35415dcd7f4d90c117. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2601 Stop hook reports green tests as failing and floods the agent context](https://app.clickup.com/t/9015210959/DEV-2601)
Task linked: [DEV-2602 Runtime theme CSS uses CSS nesting, so icons vanish on supported browsers](https://app.clickup.com/t/9015210959/DEV-2602)

.changelogs/13232.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Changed the minimum supported browsers to the Baseline 2024 set: Chrome and Edge 130, Firefox 132, and Safari 18.2.",
```
budnix: This entry is the only place a user finds out. The supported-browsers guide still says "the two latest versions" and lists no version numbers at all, and there is no migration guide step. Someone on Safari 17.4 upgrades, the grid colors stop resolving because the CSS now ships `light-dark()` natively, and there is no page that explains why. Would be good to put the numbers in that guide as part of this PR.

.changelogs/13232.json
```diff
@@ -0,0 +1,8 @@
...
+  "type": "changed",
+  "issueOrPR": 13232,
+  "breaking": true,
```
budnix: Can I ask - which release is this going into? The package is at 18.0.0 and everything in `.changelogs/` sits under UNVERSIONED, so the next one off develop is 18.1.0. Both `AGENTS.md` line 202 and `browser-targets.js` line 8 say raising the floor needs a major-release boundary plus team sign-off, and Safari moves 14.1 to 18.2 here. If the sign-off happened, saying so in the PR body would settle it.

browser-targets.js
```diff
@@ -4,16 +4,28 @@
...
- * available`) would make builds non-reproducible and would silently drop support on any rebuild.
- * Raising them is a support drop: major-release boundary, team sign-off.
+ * available`, `baseline 2024`) would make builds non-reproducible and would silently drop support on
+ * any rebuild. Raising them is a support drop: major-release boundary, team sign-off.
+ *
+ * The numbers below are the **Baseline 2024** target, resolved once with `browserslist('baseline
+ * 2024')` and then pinned. Re-resolve that query if you need to move the floor to another Baseline
+ * year; do not put the query itself in this file.
+ *
+ * The floor has to clear the CSS the themes actually ship. `light-dark()` is the tallest feature we
```
budnix, line 22: Now that `light-dark()` ships as written, a few comments say the opposite and were missed by the sweep: `manager.ts:237` and `cssVariables.ts:67` both still say the minified CSS does not use it, and `manager.unit.js:425` repeats it. The one that costs something is `tests/fixtures/demo/color-scheme-density.html:51` - it loads the minified sheet on purpose to create the lowering conflict, and that conflict no longer exists, so the spec now passes on the plain `color-scheme` write alone. That means `#buildResolvedColorVariables` could be deleted and the suite would stay green, while anyone who builds CSS at a lower floor breaks. Checking in a pre-lowered stylesheet for one leg of the matrix would keep the coverage real.

browser-targets.js
```diff
@@ -4,16 +4,28 @@
...
+ * use (Chrome 123, Firefox 120, Safari 17.5) and CSS nesting is next (Chrome 120, Firefox 117,
+ * Safari 17.2). `handsontable/test/__tests__/esTarget.unit.js` asserts that every floor here stays
```
budnix: `esTarget.unit.js` does not check this. It has four tests: ES_TARGET is a known year, tsconfig `lib` equals it, the swc target is not newer, and every `BROWSERS_LIST` entry matches a version regex. None of them mention `light-dark()` or nesting, so someone who lowers the floor later gets a green run and the lowering comes straight back. Either add the assertion or drop the sentence - right now it reads like a gate that exists.

browser-targets.js
```diff
@@ -4,16 +4,28 @@
...
+ * here is deliberately many majors below it.
  */
 exports.BROWSERS_LIST = [
-  'Chrome >= 110',
-  'Firefox >= 110',
-  'Safari >= 14.1',
+  'Chrome >= 130',
+  'Edge >= 130',
+  'Firefox >= 132',
+  'Safari >= 18.2',
```
budnix, line 38: `browserslist('baseline 2024')` also returns `ios_saf 18.2`, `and_chr` and `and_ff` - only the four desktop entries were kept. browserslist treats `ios_saf` as a separate target from `safari`, so lightningcss gets no iOS bound and eslint-plugin-compat never evaluates it, while the supported-browsers guide lists Safari iOS. Was the drop on purpose? I checked `iOS >= 18.2` resolves fine and matches the regex in `esTarget.unit.js`.

```suggestion
  'Safari >= 18.2',
  'iOS >= 18.2',
```

handsontable/.eslintrc.js
```diff
@@ -29,26 +29,21 @@ module.exports = {
...
-      },
-      {
-        selector: "CallExpression[callee.object.name='Object'][callee.property.name='hasOwn'], CallExpression[callee.name='structuredClone']",
-        message: 'Object.hasOwn needs Safari 15.4+, above the ../browser-targets.js baseline (Safari >= 14.1); structuredClone is likewise outside it (no core-js-compat entry, but compat/compat reports it unsupported in Safari 14.1). Use Object.prototype.hasOwnProperty.call(obj, key), and deepClone() from helpers/object. These two are the only entries in this group that compat/compat also catches on its own — the prototype-method groups above are this rule\'s real job.',
+        selector: "CallExpression[callee.property.name='with']",
+        message: 'Array#with needs Firefox 140+, above the ../browser-targets.js baseline (Firefox >= 132). Use arr.slice() plus an index assignment instead.',
```
budnix: The selector matches any `.with()` receiver, and `TypedArray#with` is Safari 26.0 in this repo's core-js-compat data - well above the 18.2 floor. So Safari is the taller limit here, not Firefox, and the comment above (plus both `AGENTS.md` files) saying `Array#with` is the only method above the floor is not quite right. A developer who hits this on a typed array reads "Firefox 140+", decides the rule misfired, and adds a disable.

```suggestion
        message: 'Array#with needs Firefox 140+ and TypedArray#with needs Safari 26+, both above the ../browser-targets.js baseline (Firefox >= 132, Safari >= 18.2). Use arr.slice() plus an index assignment instead.',
```

handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
```diff
@@ -337,10 +337,10 @@ export class AutocompleteEditor extends HandsontableEditor {
     let choices = choicesList;
 
     if (!sortByRelevanceSetting) {
-      // Sort a copy, never the passed-in array: `updateChoicesList` is public API and the caller's
-      // array (typically the `source` setting) must keep its original order. `Array#toSorted` would
-      // do this in one call but is outside the `browser-targets.js` baseline (Firefox 115+,
-      // Safari 16+), and swc lowers syntax only — it adds no instance-method polyfills.
+      // Sort a copy: `updateChoicesList` is public API, so the caller's array (typically the
+      // `source` setting) must keep its original order. The spread also keeps array-like callers
```
budnix: Spread runs the iterator protocol, so a plain array-like such as `{ 0: 'a', 1: 'b', length: 2 }` throws "is not iterable" here today - and `Array#toSorted` is the generic one that would actually handle it. I think the word you want is iterable (a Set, a NodeList), which the spread does support and `toSorted` does not. If array-likes are genuinely meant to work, this needs `Array.from(choices)`.

handsontable/AGENTS.md
```diff
@@ -19,7 +19,7 @@ This is the core data grid package. **TypeScript** - source files in `src/` are
...
 - `Error` `cause` is ES2022 (Chrome 94 / Firefox 91 / Safari 15.0), above the declared floor, so `throwWithCause()` assigns `cause` after construction instead of passing the constructor options bag. Do not "simplify" it back to `new Error(msg, { cause })` — older engines ignore the second argument silently, which leaves `error.cause` undefined and breaks `error.cause?.handsontable === true` detection with no visible failure.
```
budnix: This bullet still calls Safari 15.0 above the floor, but it is four majors below 18.2 now, so every targeted engine honors `new Error(msg, { cause })`. The same sentence lives in `errors.ts:8` and `errors.unit.js:18`. The line directly above it was updated in this PR, so this one looks like it was just missed. Keeping the assign-after-construction code is fine - it just needs a reason that is still true.

handsontable/src/core.ts
```diff
@@ -3668,8 +3668,6 @@ export default function Core(
     // scrollbar and every column stays reachable. Clipping those would silently hide the off-width
     // columns with no scrollbar. A definite width (`px`, `em`, `rem`, and other absolute lengths)
     // establishes a fixed box the table must not visually overflow, so it is clipped.
-    // Browser compatibility: `overflow-x: clip` requires Safari 16+. On Safari 14.1–15.x it silently
-    // falls back to `visible` (graceful degradation — pre-existing behavior, not a new regression).
```
budnix, line 3672: Right call dropping this, though the CSS side still works around the same retired floor. `_license-branding.scss:23` picks `overflow: hidden` over `clip` because "clip is Safari 16, above the declared 14.1 floor", and pays for it with a scroll container nothing ever scrolls. Lines 299 and 317 do the same for `overflow-wrap: anywhere` (Safari 15.4) and keep `word-break: break-word`. All three sit inside 18.2 now, so JS uses `clip` while the SCSS still avoids it.

.changelogs/13232.json
Haxikowy: Decided not to treat this as a breaking change, so the entry now says "build targets" and the breaking flag is gone.

The guide promises the two latest versions of each browser and names no version numbers. Safari 18.2 against a current 26.x is many majors below that window, so nothing the guide promises is dropped. Your Safari 17.4 user does break, and they were already outside the supported set before this PR.

The guide stays without numbers for the reason in the PR body: a second, much lower set of numbers next to the policy reads as the versions we support.

.changelogs/13232.json
Haxikowy: 18.1.0, off develop.

No sign-off, and I do not think this one needs the major boundary: the raise stays below the guide's two-latest window, so it is not a support drop. Last actual raise was #10440 in 13.0.0 (Chrome 41/FF 34/Safari 9 to Chrome 110/FF 110/Safari 14.1); #12017 only moved those same numbers into browser-targets.js.

The rule text in browser-targets.js and AGENTS.md is left exactly as written. It should stay strict so the next person checks the floor against that window instead of assuming.

browser-targets.js
Haxikowy: Comments fixed: manager.ts, cssVariables.ts, manager.unit.js and the fixture header.

On the coverage claim, I checked it before rewriting the fixture comment. Stubbing `#buildResolvedColorVariables` to `return ''` fails manager.unit.js in two places (the `light-dark()` pin at :427 and the auto-scheme media block at :445), so deleting it does not stay green. The fixture comment now says so and points there.

Did not check in a pre-lowered stylesheet. The E2E leg is now an end-to-end check that the override reaches rendered colors, and the lower-floor case is covered by the unit tests.

browser-targets.js
Haxikowy: Added the assertion. Two tests in esTarget.unit.js: every floor sits at or above `light-dark()` and CSS nesting per browser, and the feature table has to name every entry in BROWSERS_LIST, so adding a browser without adding it to the table fails too. Failure message names the feature, the browser and both versions.

browser-targets.js
Haxikowy: Not on purpose, taken. `iOS >= 18.2` added.

Confirmed the gap: `browserslistToTargets` returns `{chrome, edge, firefox, safari}` without it and gains `ios_saf: 18.2` with it. Same WebKit as Safari 18.2, so the emitted CSS is unchanged (rebuilt the minified themes to check).

`and_chr` and `and_ff` stay out, and the comment says why: browserslist carries only their latest version, so any floor written for them resolves to whatever is current and cannot be pinned.

handsontable/.eslintrc.js
Haxikowy: Confirmed against this repo's data: `es.typed-array.with` is safari 26.0 and ios 26.0, `es.array.with` is firefox 140. Took the suggestion with iOS folded in, and updated the comment above the rule plus both AGENTS.md files.

handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
Haxikowy: Right, iterable is the word. Fixed: "keeps iterable callers (a Set, a NodeList) working". Array-likes were never supported here, so nothing changes about what the method accepts.

handsontable/AGENTS.md
Haxikowy: Missed, thanks. Fixed in AGENTS.md, errors.ts and errors.unit.js. All three now say the options bag is inside the floor and that the assign-after-construction stays because an engine that drops the second argument fails silently.

handsontable/src/core.ts
Haxikowy: Both comments corrected: `clip` (Safari 16) and `overflow-wrap: anywhere` (Safari 15.4) are inside the floor now.

Left the CSS alone. Swapping `hidden` for `clip` stops the box being a scroll container, which the settings/width specs measure, so it is a layout change rather than a comment fix. Same for collapsing the `word-break` pair, which moves where lines break. Both are worth doing, just not in a PR that otherwise changes no behavior.

budnix (dismissed):

budnix (approved):

Issue #7727, opened 2021-04-14
On Adding new row in handsontable with nested rows cell meta and cell comments are not shifting to next row


abhishek3092:
I am using handsontable with nested rows. In my grid i have set few comments and cell bakcgrounds on load. but when i am adding new row by clicking "Insert Row Below" in the context menus, comments and cell colors are not shifting to the next rows as expected. 
Following is a JSFiddle for that. 
To reproduce just right click and insert row below  in second or third row.
[https://jsfiddle.net/b89uvtm0/](https://jsfiddle.net/b89uvtm0/)
![image](https://user-images.githubusercontent.com/31870578/114732945-8f357a80-9d60-11eb-872f-1fca0e1fdd98.png)
![image](https://user-images.githubusercontent.com/31870578/114733361-e3d8f580-9d60-11eb-944b-fe6591ea1fc5.png)

Is there any way to achieve this, or any workaround.

AMBudnik:
Thank you for sharing this issue report @abhishek3092 

I have replicated the same result on my Chrome 89 / Windows 10.

budnix:
Thanks for reporting the bug. The fix is already on its way (https://github.com/handsontable/handsontable/pull/13233) and will be released in upcoming versions.

PR #13233, opened 2026-08-25
DEV-2605: Fix cell meta not shifting on NestedRows row insert

budnix:
### Context

The PR fixes cell meta not shifting when the NestedRows plugin inserts a row.

In a grid with `nestedRows`, right-clicking a child row and choosing **Insert row above** / **Insert row below** shifted the data down by one row, but comments, `className` and every other stored cell meta stayed on their old rows — so they ended up attached to the wrong cells. The same happened with **Insert child row**. Reported in #7727 (open since 2021), reproduced on `develop`.

NestedRows **replaces** the context menu `row_above` / `row_below` callbacks with its own (`ui/contextMenu.ts`), so clicking the menu never reaches `hot.alter()`. The replacement goes through `DataManager#addSibling()` → `addChildAtIndex()`, and **Insert child row** goes through `DataManager#addChild()`. Both of those insert the row by hand: they splice `__children`, call `rowIndexMapper.insertIndexes()` and fire `beforeCreateRow` / `afterCreateRow` — but neither shifts the cell meta. The regular insert path does, in `DataMap#createRow`, which is why `alter('insert_row_below')` was never affected.

The fix adds the missing shift to both methods, using the public `spliceCellsMeta()` that this plugin's `rowMoveController` already uses for the same purpose. `finalChildIndex` / `newRowIndex` are passed straight through, so the meta shift uses the same index the index mapper was given.

One placement detail worth noting for review: in `addChildAtIndex()` the call sits **after** `enableCoreAPIModifiers()`. `spliceCellsMeta()` ends with a render, and a render taken while the core API modifiers are disabled reads the raw, un-flattened source data.

### How has this been tested?

Four E2E cases added to `handsontable/src/plugins/nestedRows/__tests__/alter.spec.js`. Each one marks a cell with a `className` and a comment, inserts a row above it, then asserts both that the row moved (data check) and that the meta moved with it. Both new failing cases were confirmed red before the fix and green after.

| Case | Before | After |
|---|---|---|
| `nestedRows` ON, context menu "Insert row below" | fails | passes |
| `nestedRows` ON, context menu "Insert child row" | fails | passes |
| `nestedRows` ON, `alter('insert_row_below')` API (control) | passes | passes |
| `nestedRows` OFF, context menu "Insert row below" (control) | passes | passes |

```bash

# The changed area
npm --prefix handsontable run test:e2e -- --testPathPattern=nestedRows          # 167 specs, 0 failures

# Broader sweep - `addChildAtIndex` / `addChild` have other consumers
npm --prefix handsontable run test:e2e -- --testPathPattern='(nestedRows|columnSummary|contextMenu|comments)'
                                                                                # 909 specs, 0 failures
npm --prefix handsontable run test:unit                                         # 3732 tests, 0 failures
npm run test:types --prefix handsontable                                        # passes
npm run eslint --prefix handsontable                                            # 0 errors
```

### Types of changes

### Related issue(s):

1. #7727
2. DEV-2605

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2605

---

> [!NOTE]
> **Medium Risk**
> Touches nested row insert hooks and render timing around `spliceCellsMeta`, but the change mirrors `DataMap#createRow` and is covered by new E2E tests.
> 
> **Overview**
> Fixes **#7727**: with `nestedRows`, context-menu **Insert row above/below** and **Insert child row** moved row data but left comments, `className`, and other cell meta on the old visual rows.
> 
> NestedRows inserts through `DataManager#addChild` and `#addChildAtIndex` without going through `DataMap#createRow`, so meta was never shifted. The PR calls `hot.spliceCellsMeta(index, 0, [])` at the same flattened index used for `rowIndexMapper.insertIndexes`, matching the normal create-row path. In `addChildAtIndex`, that call runs **after** `enableCoreAPIModifiers()` so a render triggered by `spliceCellsMeta` does not read un-flattened source data.
> 
> E2E coverage in `alter.spec.js` exercises the failing context-menu paths plus controls for `alter('insert_row_below')` and nestedRows-off inserts. Changelog entry **7727.json** documents the fix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ce32ece9c70bd51cec1ec61ebf87cb3c8f7487e4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/handsontable/src/plugins/nestedRows/__tests__/alter.spec.js
+++ b/handsontable/src/plugins/nestedRows/__tests__/alter.spec.js
@@ -134,18 +134,32 @@ describe('NestedRows', () => {
      * Inserts a row through the context menu, the way the issue reporter did.
      *
      * @param {number} row Visual row index to open the context menu on.
-     * @param {number} itemIndex Index of the non-separator context menu item to click.
+     * @param {string} optionName Label of the context menu item to click.
      */
-    async function insertRowViaContextMenu(row, itemIndex) {
+    async function insertRowViaContextMenu(row, optionName) {
       await selectCell(row, 0);
       await contextMenu();
+      await selectContextMenuOption(optionName);
+    }
 
-      $('.htContextMenu .ht_master .htCore')
-        .find('tbody td')
-        .not('.htSeparator')
-        .eq(itemIndex)
-        .simulate('mousedown')
-        .simulate('mouseup');
+    /**
+     * Marks one cell above the insertion point and one below it, so an assertion can tell a correct
+     * shift from a shift at the wrong index. Splicing the meta too high moves the row above too.
+     *
+     * @param {number} aboveRow Visual row above the insertion point.
+     * @param {number} belowRow Visual row below the insertion point.
+     * @param {number} column Visual column to mark.
+     * @returns {object} The values held by both marked cells before the insert.
+     */
+    async function markCells(aboveRow, belowRow, column) {
+      await setCellMeta(aboveRow, column, 'className', 'above-cell');
+      await setCellMeta(belowRow, column, 'className', 'below-cell');
+      await setCellMeta(belowRow, column, 'comment', { value: 'below-comment' });
+
+      return {
+        aboveValue: getDataAtCell(aboveRow, column),
+        belowValue: getDataAtCell(belowRow, column),
+      };
     }
 
     it('nestedRows ON, context menu "Insert row below" (the exact issue repro)', async() => {
@@ -157,20 +171,19 @@ describe('NestedRows', () => {
         rowHeaders: true,
       });
 
-      await setCellMeta(3, 2, 'className', 'marked-cell');
-      await setCellMeta(3, 2, 'comment', { value: 'marked-comment' });
+      // Rows 1-5 are the children of the first parent. Insert below row 2.
+      const { aboveValue, belowValue } = await markCells(1, 3, 2);
 
-      const markedValueBefore = getDataAtCell(3, 2);
+      await insertRowViaContextMenu(2, 'Insert row below');
 
-      // Non-separator items with nestedRows: 0=add_child, 1=detach_from_parent, 2=row_above, 3=row_below.
-      await insertRowViaContextMenu(1, 3);
+      // The row above the insertion point must not move, and neither must its meta.
+      expect(getDataAtCell(1, 2)).toBe(aboveValue);
+      expect(getCellMeta(1, 2).className).toBe('above-cell');
 
-      // Sanity: the marked row really moved from visual 3 to visual 4.
-      expect(getDataAtCell(4, 2)).toBe(markedValueBefore);
-
-      // The meta must have followed the row.
-      expect(getCellMeta(4, 2).className).toBe('marked-cell');
-      expect(getCellMeta(4, 2).comment).toEqual({ value: 'marked-comment' });
+      // The row below moved down by one, so its meta must follow.
+      expect(getDataAtCell(4, 2)).toBe(belowValue);
+      expect(getCellMeta(4, 2).className).toBe('below-cell');
+      expect(getCellMeta(4, 2).comment).toEqual({ value: 'below-comment' });
       expect(getCellMeta(3, 2).className).toBeUndefined();
       expect(getCellMeta(3, 2).comment).toBeUndefined();
     });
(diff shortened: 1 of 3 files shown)


qunabu:
Task linked: [DEV-2605 Fix cell meta not shifting when NestedRows inserts a row via the context menu](https://app.clickup.com/t/9015210959/DEV-2605)

handsontable/src/plugins/nestedRows/data/dataManager.ts
```diff
@@ -579,6 +579,9 @@ class DataManager {
 
     this.hot.rowIndexMapper.insertIndexes(newRowIndex, 1);
 
+    // Shift the cell meta like `DataMap#createRow` does, or it stays on the rows that moved down (#7727).
+    this.hot.spliceCellsMeta(newRowIndex, 0, []);
```
Haxikowy: `spliceCellsMeta` always renders, and this sits inside the stash window where the collapsed parents are temporarily expanded, so collapsed grids flash open on every insert (measured: 3 renders -> 4, the new frame showing 8 visible rows instead of 5). It is also redundant, since `applyStash()` renders right after, and it costs one extra full render per inserted row.

`DataMap#createRow` shifts the meta without rendering (dataMap.ts:428):

```suggestion
    this.hot._getMetaManager().createRow(this.hot.toPhysicalRow(newRowIndex), 1);
```

This also fixes the index space: `spliceCellsMeta` takes a visual index and translates it again, while `newRowIndex` is physical. Identity today only because the stash untrimmed everything; add `trimRows` or `filters` and it shifts the wrong row.

Separately, `newRowIndex` on line 578 falls back to `?? 0`. That was harmless while it only fed `insertIndexes`; now a `getRowIndex` miss shifts all cell meta from row 0 down one instead of doing nothing. Worth bailing out of the meta shift when it returns null.

handsontable/src/plugins/nestedRows/data/dataManager.ts
```diff
@@ -623,6 +626,11 @@ class DataManager {
 
       this.plugin.enableCoreAPIModifiers();
 
+      // Shift the cell meta like `DataMap#createRow` does, or it stays on the rows that moved
+      // down (#7727). Keep it after `enableCoreAPIModifiers()` - `spliceCellsMeta` renders, and
```
Haxikowy: `DataMap#createRow` doesn't call `spliceCellsMeta`; it calls `metaManager.createRow()` with a physical index and never renders, so the comparison is off on both counts.

The placement note is accurate but argues for not rendering here at all. `this.hot._getMetaManager().createRow(this.hot.toPhysicalRow(finalChildIndex), 1)` drops the constraint (`toPhysicalRow` ignores the modifier flag) and removes the render that currently lands before `afterCreateRow`, which `mergeCells`, `formulas` and `columnSummary` all shift state on.

handsontable/src/plugins/nestedRows/__tests__/alter.spec.js
```diff
@@ -127,4 +127,147 @@ describe('NestedRows', () => {
...
+        rowHeaders: true,
+      });
+
+      await setCellMeta(3, 2, 'className', 'marked-cell');
+      await setCellMeta(3, 2, 'comment', { value: 'marked-comment' });
+
+      const markedValueBefore = getDataAtCell(3, 2);
+
+      // Non-separator items with nestedRows: 0=add_child, 1=detach_from_parent, 2=row_above, 3=row_below.
+      await insertRowViaContextMenu(1, 3);
```
Haxikowy: Every case marks a cell below the insertion point, so passing `0` to both `spliceCellsMeta` calls keeps all assertions green. Index correctness is the defect class here. Mark a cell above the insertion point and assert its meta did not move: one `setCellMeta`, two `expect`s.

handsontable/src/plugins/nestedRows/__tests__/alter.spec.js
```diff
@@ -127,4 +127,147 @@ describe('NestedRows', () => {
...
+
+  // Repro for https://github.com/handsontable/handsontable/issues/7727
+  describe('cell meta shifting when a row is inserted (#7727)', () => {
+    /**
+     * Inserts a row through the context menu, the way the issue reporter did.
+     *
+     * @param {number} row Visual row index to open the context menu on.
+     * @param {number} itemIndex Index of the non-separator context menu item to click.
+     */
+    async function insertRowViaContextMenu(row, itemIndex) {
```
Haxikowy: `selectContextMenuOption('Insert row below')` exists in `test/helpers/common.js` and is used in the sibling `nestedRows.spec.js` for these labels. Drops the ordinals and the decoder comments, which depend on both the default menu and `appendOptions` staying put.

Haxikowy (commented):
**`dataManager.ts:640`, the `parent === null` branch.** The context menu takes this branch for top-level rows (`getRowParent` is null for a parent), and it gets no fix. With the modifiers off, `countSourceRows()` counts roots, so `DataMap#createRow` shifts meta at flattened row 1 while the row lands at flattened row 6. On `getSimplerNestedData`, "Insert row below" on row 0 moves the meta of rows 1-5, none of which moved. Pre-existing, not a regression, but the title and changelog say Insert row above/below is fixed. Either shift at the flattened index here too, or narrow the changelog and open a follow-up.

**`dataManager.ts:608`, `finalChildIndex = parentIndex + index + 1`.** That is the flattened position only if every preceding sibling is a leaf. `cacheNode` flattens depth-first, so a sibling with two grandchildren puts the data at `parentIndex + 4` while the meta splices at `parentIndex + 2`. Pre-existing for `insertIndexes`, but routing cell meta through it makes it the wrong-row meta move #7727 is about. `addChild` reads `getRowIndex(childElement)` after `rewriteCache()`; same works here. `getSimplerNestedData()` is two levels deep, `getMoreComplexNestedData()` in the same file is three.

`detachFromParent` has the same gap: it splices `__children` and fires the row hooks by hand without touching meta. Same context menu, worth a follow-up on DEV-2605.

handsontable/src/plugins/nestedRows/data/dataManager.ts
budnix: Good spot — all three fixed in `4b7518b`, it now calls the meta manager with the physical index and bails out on null. Your snippet had `toPhysicalRow` on an already-physical index, so I dropped that part.

handsontable/src/plugins/nestedRows/data/dataManager.ts
budnix: You're right, that comment was wrong on both counts — rewritten in `4b7518b`, and the meta index now comes from `getRowIndex` so a sibling with descendants lands on the right row.

handsontable/src/plugins/nestedRows/__tests__/alter.spec.js
budnix: Nice catch — a hardcoded 0 really would have passed. Every case now marks a cell above the insertion point too, plus deep-nesting and `trimRows` cases in `4b7518b`.

handsontable/src/plugins/nestedRows/__tests__/alter.spec.js
budnix: Agreed — switched to `selectContextMenuOption` in `4b7518b`, the ordinals and their comments are gone.

Haxikowy (approved):

PR #13234, opened 2026-08-25
DEV-2591: Keep the grid scrollbars reachable under frozen overlays

### LINKED ISSUE #10370: Scroll bar in FireFox when using fixedColumnsLeft

budnix:
### Context
The PR fixes [#10370](https://github.com/handsontable/handsontable/issues/10370) — open since May 2023, re-confirmed by the reporter in September 2024, and still reproducible on `develop`.

A frozen overlay is painted on top of the grid's scrollbar, which hides it and swallows the press. The reported case is frozen columns and the horizontal scrollbar: with the grid scrolled fully left, the thumb sits entirely under the frozen columns, so a mouse user cannot drag it at all.

**The precondition is a browser that paints the scrollbar over the content instead of reserving space for it** — a floating thumb of zero layout width. That is not one engine. It is Firefox everywhere, Chrome and Safari on macOS while "Show scroll bars" is set to "Automatically", and Chrome on Windows 11 and on GTK. An earlier revision of this branch gated the fix on `isFirefox()`; that was wrong, and the gate is now a measured zero. It is also why the 2023 investigation closed as "cannot reproduce": every attempt used a classic-scrollbar environment, where Handsontable is correct in every version.

Cause, in order:

1. A floating scrollbar occupies no space, so the browser never shrinks the master holder and the measured scrollbar width is 0.
2. Each overlay therefore reserves nothing and is sized to the full holder.
3. The browser still paints the scrollbar inside the content area.
4. The overlay covers it, hiding the scrollbar and swallowing the press.

Step 4 is the part that is not engine-specific. A browser paints a scroll container's scrollbar above that container's own contents — but the overlay clones are *siblings* of the master, not its contents, so they are painted above it in every engine.

#### Scope: every overlay that touches an edge a scrollbar is drawn on

Measured by dragging in the strip and reading `scrollLeft` / `scrollTop`; 0 means the press never reached the scrollbar.

| Overlay | Edge / scrollbar | Before | After |
|---|---|---|---|
| `inlineStartOverlay` (frozen columns) | bottom / horizontal | 0px | 499px |
| `bottomOverlay` (`fixedRowsBottom: 3`) | bottom / horizontal, **full width** | 0px | 1721px |
| `topOverlay` (`fixedRowsTop: 4`) | inline-end / vertical | 0px | 1575px |
| `topOverlay` with **no frozen rows** — the column header alone is a 30px overlay | inline-end / vertical | covered | clear |
| `bottomInlineStartCornerOverlay` | bottom / horizontal | covered | clear |

`bottomOverlay` spans the entire width, so with any `fixedRowsBottom` the horizontal scrollbar is completely unreachable. `topInlineStartCornerOverlay` needs no clearance — it sits at the inline-start/top, away from both scrollbars.

**The question is whether the overlay is rendered, not whether it holds frozen content.** An overlay
drawn on an edge covers the scrollbar there whether it carries frozen rows or only a column header, so
the clearance follows `needFullRender` and asks nothing else. An earlier revision of this branch asked
each overlay instead whether it held frozen rows or columns *of its own*. That reads as the tidier
question and is the wrong one twice over: it left a headers-only grid covered, and — the case that
matters — it left the **vertical** scrollbar covered on any grid with frozen columns and no frozen rows,
because there the column header is the only overlay on that edge. The horizontal scrollbar was cleared
and the vertical one was not, on the commonest frozen layout there is. Both are pinned by tests, and the
per-overlay version fails them.

The two halves still switch on together. Clipping a clone uncovers the master underneath, which after a
scroll is a different row and column, so a clip with no band behind it shows a data cell where the frozen
content or the header belongs — that asymmetry is what changed 118 Argos snapshots earlier on this
branch.

#### What changed

**One track per grid, not one per overlay.** A single band per axis is drawn in a zero-size `position: sticky` host inside the master holder, spanning the whole scrollport. Covering only the frozen overlays' own strips left an opaque patch beside a transparent gap over the master. The host must be the holder's first child (a sticky box only shifts toward its own edge), and the band must live inside the holder, because one stacked above it would hide the very thumb it sits behind.

**The overlays are clipped, not resized.** Each gets a `clip-path: inset()` over the strip its own edge crosses. Every box the viewport calculations measure keeps its size, and a clipped region is neither painted nor hit-tested, so the press falls through to the scrollbar. Resizing the roots or holders instead — the earlier approach on this branch — rippled into the viewport calculations and cost 19–24 walkontable failures; the clip costs none. Only overlays that actually reach the scrollport edge are clipped: `inset()` is relative to each element's own border box, and the inline-start overlay's right edge sits mid-grid.

**Per axis.** The band follows whichever scroll offset actually moved, because a browser only draws the scrollbar for the axis being scrolled. A hover may hold a band open but never open one.

**The track does not animate, deliberately.** It is drawn under the clones, so exactly two states are clean: fully opaque with the clip closed, or absent with the clip open. Any partial opacity is a defect — with the clip open the fade shows over the master's segment alone and the track reads differently over each overlay; with it closed the strip shows the master's cell where frozen content belongs and a column header looks cut short. Both variants were built and rejected from frame-by-frame measurements.

**Presses in the band are swallowed by coordinate**, in the capture phase, so a click on the empty track does not move the selection. The wheel is excluded, so scrolling over the track keeps working.

**The band paints over the strip but does not hit-test it** (`pointer-events: none`). It is drawn above the selection decorations on purpose — at a lower z-index a selection border painted across the track while the cell it belongs to sat hidden underneath — but taking the *hit* as well made it the event target for everything beneath, including the selection's own controls. The autofill corner of a range that reaches the grid's edge sits inside the strip, so the press landed on the band instead of the handle and no fill drag ever started: measured on the `main` theme, where that handle falls 14px above the bottom edge and the band spans the last 16px. Painting over a control must not disarm it. The coordinate swallow therefore lets a press through when it really landed on a `.wtBorder`, and swallows everything else.

**The cost, stated plainly.** Because the gate is rendering rather than frozen content, this reaches **every grid with headers**, which on a floating-scrollbar setup is most of them: after any scroll, a 16px strip along the scrolled edge stops taking presses for about a second, and stays open for as long as the pointer rests near that edge. Inside that second, a click on a cell under the strip does nothing — no selection, no editor. The strip belongs to the scrollbar while the track is up, which is what a native overlay scrollbar does with its own track, but on a default grid it is a real second of dead pixels along the edge and it should be read as a deliberate trade-off rather than a detail. Two things follow from it that are worth knowing: a cell in the last row is briefly unclickable after scrolling to the bottom, and a selection dragged to the far corner right after a scroll needs the track to fade first. One overlay case is not recoverable at all — see the known cost below.

#### Two places where the measurement itself was wrong

Both come from the same root: `getScrollbarWidth()` measures a bare off-DOM probe, which answers "does this engine give scrollbars layout space", not "does *this* scroller". Where the two disagree — measured on Firefox 154, probe 0 against a real 15px holder gutter — everything downstream is wrong.

1. **Two scrollbars in one grid.** The clearance turned on because the probe said 0, so the strip was drawn beside a scrollbar that already had a gutter. The gate now also requires the holder itself to reserve nothing (`reservedScrollbarSpace`, per axis). It is an AND, so the new condition can only ever switch the clearance off, never on.
2. **The column header ignoring the scrollbar.** `topOverlay` and `bottomOverlay` subtracted the probed width from the workspace width, so with the probe at 0 they came out 15px too wide and ran underneath the scrollbar — a mismatched top-right corner, while the frozen columns on the left stayed correct. They now use the holder's own `clientWidth`, the same correction `inlineStartOverlay` already applies on the other axis for #12632.

#### A known cost, decided deliberately

A bottom-pinned overlay is held in place by `style.bottom`, so clipping its last pixels does not move it
out of the scrollbar's way — it removes that much of its content. With `fixedRowsBottom` and the
horizontal track on screen, the last frozen row shows **13 of 29px (45%)** until the track fades.

Lifting the overlay instead keeps the row whole, and was built and compared side by side. It was
**rejected**: the lift opens a 16px gap between the frozen rows and the holder's bottom edge that only
the band fills, which makes the band load-bearing rather than cosmetic — if it ever failed to draw, the
gap would show scrolled master content instead of a clean edge. Clipping has no such dependency.

The clip takes whatever is drawn in that strip, not only cell content. A selection border or an
autofill corner belonging to a frozen overlay is removed with it: measured on a grid with frozen
columns and both tracks up, the handle of a selection inside them sits 4px above the holder's bottom
edge, inside a `clip-path: inset(0px 0px 16px)`, and `elementFromPoint` there returns the `TD` beneath
rather than the border. So that one handle cannot be grabbed until the track fades. The `.wtBorder`
pass-through in the swallow does not reach it and cannot: a clipped region is never hit-tested, so the
border is not the event target to begin with. Matching by geometry instead would be worse — the press
would land on the `TD` underneath and move the selection to a cell that is not on screen. A handle on
the master, which is not clipped, is unaffected and is covered by `fill-handle-formulas.spec.ts`, which
fails without the pass-through.

This is a decision, not an oversight.

### How to see it

Demo to test: https://demos.handsontable.com/d/5v3e5p3o2r/

The fix only switches on where the browser paints **floating** scrollbars — drawn over the content, with
no layout space of their own. On macOS that is a system setting, so it has to be set first or none of
this is visible.

**System Settings → Appearance → Show scroll bars** has to be on **"Automatically based on mouse or
trackpad"** or on **"When scrolling"**:

Zrzut ekranu 2026-08-26 o 09 15 46

On **"Always"** macOS reserves a gutter for every scrollbar, nothing here can activate, and the grid has
to look exactly as it does on `develop`. That case is worth a pass of its own: switching the feature off
correctly is half of what this PR does, and getting it wrong draws a second scrollbar beside the real
one. Firefox on any OS, and Chrome on Windows 11, paint floating scrollbars whatever this setting says.

#### Before
The scrollbar was not visible when it was located beneath an overlay.
Kapture 2026-08-26 at 11 58 31

#### After
Kapture 2026-08-26 at 11 59 56

- Unit tests added/modified (`*.unit.ts`): `handsontable/src/3rdparty/walkontable/test/unit/overlay/scrollbarClearance.unit.ts` (extended) and `handsontable/src/3rdparty/walkontable/test/unit/overlay/scrollbarVisibility.unit.ts` (new) — 83 tests covering the clearance decision, the per-axis gutter measurement, the pointer-capability gate, the overlay width choice, the deferred layout reads, the `clip-path` geometry including RTL, the band DOM lifecycle including nested grids, and the visibility state machine.
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`): `tests/e2e/overlay-scrollbar-clearance.spec.ts` (5 tests) and `tests/e2e/overlay-scrollbar-bottom-gutter.spec.ts` (1 test, which drops `--hide-scrollbars` so a runner that has real scrollbars exercises the space-taking branch rather than skipping it). Whether a run gets floating or space-taking scrollbars is an OS/browser display setting a page cannot request, so the spec reads the regime at runtime and asserts the matching contract. Measured, not assumed: headless Chromium hides its scrollbars, so **CI takes the floating branch** and covers the fix itself — the band is drawn and the covering overlay clipped out of it. A headed run on a classic-scrollbar desktop takes the other branch, which guards the doubled scrollbar. Both branches were checked by mutation: setting `OVERLAY_SCROLLBAR_CLEARANCE` to 0 fails the spec, and removing the regime checks reproduces two bands beside a real 15px gutter on a headed classic desktop. The last three tests pin the gate and its cost: a headers-only grid is cleared like any other; a grid with frozen columns and **no** frozen rows clears its vertical scrollbar (the case the per-overlay check got wrong, and the one the first test cannot catch, because its grid has frozen content on both axes and passes whichever question the gate asks); and the strip is given to the scrollbar while the track is up and handed back once it has gone. The first two were confirmed by mutation: run against a bundle still carrying the per-overlay gate, both fail.
- Two shared test fixtures changed with the gate, both for the same reason — a test that scrolls and then clicks near an edge is pressing into the track. `tests/fixtures/pages/walkontable/OverlaysPage.ts`: `scrollToEnd()` now waits for the track to fade, because the last row's cells are shorter than the strip is deep. `visual-tests/tests/cross-browser/copy-paste.spec.ts`: the two scrolled tests wait before clicking the far corner cell, which sits inside both bands — without it the selection never left its anchor and 1 cell was copied instead of 931, while the assertions still passed because they only read the first pasted cell. That was two changed Argos snapshots with nothing in the test output to explain them.
- Visual regression (`visual-tests/src/test-runner.ts`): the clearance is transient by design — a scroll opens the band and clips the overlays, and both close about a second later. A screenshot taken inside that window records whichever half the runner reached, which changed 118 Argos snapshots on this branch with no real difference behind any of them. The fixture now waits for both halves to close before every capture, wrapping `page.screenshot` once per page because the specs call it directly in a few hundred places. Measured: `2 bands / 4 clips` → `0 / 0` after ~989ms, and ~2ms when nothing is transient, which is nearly every screenshot. Both halves have to be waited for — a clip can outlive its band, so counting fillers alone would call that settled. A timeout does not fail the test: a pointer resting beside the scrollbar holds the band open by design, and that must not turn a visual check into a hang.
- Type tests (`*.types.ts`) updated if public API changed: **none** — no public API change. Everything added is internal to Walkontable's overlay module.
- For a bug fix — the spec that fails without this fix: `overlayScrollbarClearance › should reserve nothing when the holder itself reserves space, whatever the probe says` and the `reservedScrollbarSpace` suite both fail without the gate change. The clearance geometry suites fail without the clearance module.
- Demo page / recorded trace (for UI changes): reproduced and verified on a local page driven over the DevTools protocol against real Firefox 154 and Chrome. Measured as the overlay's right edge relative to the holder's inner edge: **15px past it before, 0 after**. Composited-pixel sampling down the strip in Chrome: one colour at every sample, and **0 frames out of 161** showing either failure mode.

### Commands run

```
(no output — clean)

Test Suites: 309 passed, 309 total
Tests:       3731 passed, 3731 total

npm --prefix tests run test -- e2e/overlay-scrollbar-clearance.spec.ts e2e/overlay-scrollbar-bottom-gutter.spec.ts
36 passed   (6 tests across the theme × bundle matrix)

npm run test:e2e --prefix handsontable   (legacy Jasmine/Puppeteer)
9658 specs, 0 failures, 8 pending specs

816 specs, 0 failures

✖ 601 problems (0 errors, 601 warnings)

✓ stylelint --cache "src/**/*.{css,scss}" "test/**/*.{css,scss}"

Test Suites: 15 passed   Tests: 75 passed

Test Suites: 4 passed    Tests: 33 passed

Test Suites: 13 passed   Tests: 291 passed
```

The wrapper suites are included because this change adds listeners and teardown to the overlay
lifecycle, and the observer stubs those suites rely on are duplicated per package — a method one of them
lacks fails that whole suite independently of core.

A note on the walkontable suite, because the number moved during development. It is green with the OS scrollbar setting on "Always", which is the space-taking mode the legacy suite was written against and the one CI runs. Earlier runs of this branch were made with the setting on "Automatically" so the fix would be active, and reported 48–61 failures; most of those come from the display mode itself rather than from any change here (a control run with the clearance disabled still failed 48). Treat a failure count taken under floating scrollbars as a property of the machine, not of the branch — and note that in that mode the suite exercises a code path it makes no assertions about.

### Types of changes

### Related issue(s):

1. #10370
2. DEV-2591
3. #12632 (the same probe-versus-holder correction, previously applied to the other axis)

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

- [x] This change needs a manual QA pass — the fix only activates when the browser draws floating scrollbars, which cannot be turned on from a page and so cannot be set up in CI. To check it: set macOS "Show scroll bars" to "Automatically based on mouse or trackpad" or to "When scrolling" — **not** "Always" (or use Firefox on any OS, or Windows 11 Chrome), open a grid with `fixedColumnsStart`, `fixedRowsTop` and `fixedRowsBottom`, and confirm that (1) both scrollbars can be grabbed and dragged where a frozen overlay overlaps them, (2) the track appears only along the axis being scrolled and leaves no gap once it goes, (3) clicking the empty track does not move the selection while the wheel still scrolls, (4) the column header stops at the scrollbar rather than running underneath it, (5) a grid with headers but nothing frozen is cleared too — it is in scope, not an exception, and (6) on a grid with frozen columns and no frozen rows the **vertical** scrollbar is clear, not just the horizontal one. Worth dragging a fill handle from a selection that reaches the bottom edge while the track is showing. Worth repeating with the classic space-taking setting, where the whole feature must stay switched off.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2591

---

> [!NOTE]
> **Medium Risk**
> Large Walkontable overlay/DOM and input-handling changes, but activation is gated to floating-scrollbar regimes with extensive tests; wrong gating could still affect selection or overlay alignment on edge layouts.
> 
> **Overview**
> Fixes **#10370**: on browsers that paint **floating** scrollbars (no layout gutter), frozen overlays and headers could cover the thumb and block grabbing or scrolling.
> 
> When the holder is the scroller and a fine pointer is available, the overlays layer now **clips** covering clones out of a ~16px edge strip, paints a **unified track band** inside the master holder (under the thumb), and **swallows** pointer presses in that strip so clicks hit the scrollbar—not cell selection—while wheel scrolling still works. Touch presses and selection controls (e.g. autofill via `.wtBorder`) are explicitly let through.
> 
> **Scrollbar visibility** is tracked per axis (scroll + pointer proximity/pin + fade timer) so bands and clips stay in sync and close together—no partial fade that would show wrong cells under headers or frozen rows.
> 
> Sizing no longer trusts the engine **scrollbar-width probe** alone: **`reservedScrollbarSpace`** and **`overlayExtentBesideScrollbar`** use the holder’s real gutter (fixes Firefox styled scrollbars and bottom-overlay lift). The feature stays **off** when scrollbars already take space, the window trims scrolling, or the device is touch-only (`any-pointer: fine`).
> 
> Adds unit/E2E coverage and visual-test **settle waits** before screenshots so transient bands don’t flake Argos or corner cell clicks after scroll-to-end.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1f594fbce2c921724332ea1df9292935304ccbed. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2591 Firefox overlay scrollbars: grid scrollbars unreachable under frozen overlays (#10370)](https://app.clickup.com/t/9015210959/DEV-2591)

handsontable/src/3rdparty/walkontable/src/overlay/overlays.ts
```diff
@@ -590,6 +748,88 @@ class Overlays {
...
+    // Also off on a touch-only device: nothing there can grab a thumb, and the band would swallow the
+    // press that becomes a tap - see `canGrabScrollbar`.
+    const holderScrolls = this.scrollableElement === holder
+      && canGrabScrollbar(this.#deps.rootWindow);
+    const scrollbarWidth = geometryReader.getScrollbarWidth(rootDocument);
+    // Only where an overlay is actually clipped out of the strip. The band exists to fill in for the
+    // frozen content that stops short of the scrollbar; along an edge no overlay reaches, there is
+    // nothing to fill in for, and drawing one paints a grey strip over live cells and swallows the
+    // presses there - which is what a grid with no frozen rows or columns used to get.
+    const covers = (edge: 'bottom' | 'inlineEnd') =>
```
budnix: **The `covers()` guard does not do what the comment above it says.** It reads "along an edge no overlay reaches, there is nothing to fill in for... which is what a grid with no frozen rows or columns used to get" — but `covers()` is satisfied by ordinary headers, not by frozen rows/columns:

- `shouldRenderTopOverlay` is `fixedRowsTop > 0 || columnHeaders().length > 0` (`tableView.ts:860`), so `colHeaders: true` renders the top overlay, and it publishes `inlineEnd: 16`.
- `shouldRenderInlineStartOverlay` is `fixedColumnsStart > 0 || rowHeaders().length > 0` (`tableView.ts:856`), so `rowHeaders: true` renders the inline-start overlay, and it publishes `bottom: 16`.

`syncScrollbarTrackBands` then draws the inline-end band at full scrollport height and the bottom band at full scrollport width. Only the header strip of each overlay is clipped out of it; the rest of the band is painted opaque over the master's **live cells**, and `#bandSizes` makes `#swallowBandPress` eat every press there.

Concrete: `{ colHeaders: true, rowHeaders: true }`, no `fixedRowsTop`/`fixedColumnsStart`/`fixedRowsBottom`, macOS floating scrollbars. Every scroll paints a 16px full-height grey strip down the right edge and a 16px full-width strip across the bottom, over normal cells, and swallows clicks there for ~1s. That is exactly the configuration the guard was written to exclude.

Either gate on an overlay that actually reaches the edge (a frozen region, not a header band), or limit each band's extent to the union of the clipped overlays' rects.

handsontable/src/3rdparty/walkontable/src/overlay/scrollbarClearance.ts
```diff
@@ -0,0 +1,521 @@
...
+ * @param {number} masterClientExtent The master holder's inner size on this axis, or 0 when unmeasurable.
+ * @param {number} probedScrollbarWidth The engine-wide scrollbar width, from `getScrollbarWidth()`.
+ * @returns {number}
+ */
+export function overlayExtentBesideScrollbar(
+  workspaceExtent: number,
+  masterClientExtent: number,
+  probedScrollbarWidth: number
+): number {
+  if (masterClientExtent > 0) {
```
budnix: `masterClientExtent` wins unconditionally, including when `probedScrollbarWidth > 0` — i.e. on classic-scrollbar systems, which is outside the regime this PR targets. That silently changes the top/bottom overlay width wherever the master holder is **not** the element that owns the scrollbar.

Case: grid inside an `overflow: auto` DIV with no explicit HOT `width`/`height`. `trimmingContainer !== rootWindow`, so `rootSized` is true; `measureWorkspaceWidth` returns `clientWidth(trimmingContainer)` (already net of the DIV's 15px scrollbar); the holder is a block filling that same width, so `clientWidth(holder) ≈ workspaceWidth`.

- before: `width = workspaceWidth - 15`
- after: `width = clientWidth(holder) ≈ workspaceWidth`

15px wider, on a path the floating-scrollbar fix never needed to touch (there the two agree, both gutters being 0). #12632's precedent was the height axis with the holder as the scroller; generalising it to widths drags in configurations where the premise ("the scroller's own inner size is the authority") does not hold, because this holder is not the scroller.

Suggest keeping the probe path when the holder reserves no gutter of its own — e.g. prefer `masterClientExtent` only when `reservedScrollbarSpace(holder, axis) > 0 || probedScrollbarWidth === 0`.

handsontable/src/3rdparty/walkontable/src/overlay/overlays.ts
```diff
@@ -419,6 +453,129 @@ class Overlays {
...
+    // and the wrong one for swallowing a press: on a hybrid device (touchscreen laptop, tablet with a
+    // trackpad) it is true, so without this the band ate finger taps in the bottom strip - the exact
+    // "scroll, then tap a cell near the edge" failure the touch-only exclusion was written to prevent.
+    //
+    // Only `pointerdown` carries a `pointerType`; the `mousedown`/`click` that follow a tap are
+    // compatibility events that look identical to real mouse ones. So the answer is carried across the
+    // gesture - and cleared again at its end, since a flag left standing would suppress the swallow for
+    // the NEXT press, letting a mouse click through to the cell hidden under the band.
+    const isGestureStart = event.type === 'pointerdown';
+    const isGestureEnd = event.type === 'click' || event.type === 'contextmenu';
```
budnix: `dblclick` is in `BAND_SWALLOWED_EVENTS` but is not a gesture end, and the `click` that always precedes it already reset `#pressIsTouch` to `false`. So on a hybrid device (touchscreen laptop, tablet + trackpad — where `canGrabScrollbar()` is `true` and the band is drawn), a **touch** double-tap in the band is swallowed:

`pointerdown(touch)` → flag `true` · `click` → `isGestureEnd`, flag `false`, `fromTouch` was `true`, let through · `pointerdown(touch)` → flag `true` · `click` → flag `false`, let through · **`dblclick`** → `fromTouch` is `false` → swallowed.

Net effect: a finger double-tap near the bottom/right edge selects the cell but never opens the editor — the same class of "scroll, then tap near the edge stops working" failure the touch exclusion was written to prevent.

```suggestion
    const isGestureEnd = event.type === 'click' || event.type === 'contextmenu';
```
…is the line to widen; `dblclick` needs to either count as a gesture end *after* it is evaluated, or the flag needs to survive until the next `pointerdown` rather than being cleared on `click`.

handsontable/src/3rdparty/walkontable/src/overlay/overlays.ts
```diff
@@ -469,6 +503,136 @@ class Overlays {
...
+    }
+
+    const rect = this.#deps.geometryReader.getBoundingClientRect(this.wtTable.holder);
+
+    const rtl = this.wtSettings.getSetting<boolean>('rtlMode');
+
+    if (isPointInScrollbarBand(rect, bottom, inlineEnd, rtl, event.clientX, event.clientY)) {
+      event.stopPropagation();
+      event.preventDefault();
+    }
```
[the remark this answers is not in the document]
budnix: Not fixing this one — the finding is real but the cause is not the `.wtBorder` exemption.

Measured on a grid with frozen columns, both tracks up: the fill handle of a selection inside the frozen columns sits 13px above the holder's bottom edge, and `elementFromPoint` there returns the `TD`, not the border. The inline-start clone carries `clip-path: inset(0px 0px 16px)`, and a clipped region is neither painted nor hit-tested — so the handle is *removed* from that strip, not covered by the band. The swallow runs after hit-testing, so no change to it can bring the handle back.

Matching by geometry instead would be worse: the press would still land on the `TD` underneath, so it would move the selection to a cell that is not on screen rather than start a fill.

This is the clipping cost the description already records — the same clip that leaves a frozen bottom row showing 13 of 29px until the track fades also takes the last pixels of a selection border drawn there. The only fix is not clipping, which is the feature.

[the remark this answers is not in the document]
budnix: Working as intended.

While the track is up, that strip belongs to the scrollbar, so a press in it is swallowed whether the cell holds text or a nested grid. That is the behaviour this PR deliberately keeps, and `overlay-scrollbar-clearance.spec.ts` asserts it both ways: a press in the open track does not move the selection, and the same press lands normally once the track has faded.

The swallow only runs while the outer grid has a band open, and only for points inside it, so a nested grid anywhere else on the grid is unaffected.

Treating a nested grid as an exception would make the same 16px strip behave one way over plain cells and another over cells that happen to hold a grid.

budnix:
@adrianspdev this pull request needs to be tested and verified through manual testing.

adrianspdev:
Hi @budnix — reviewed the code and ran manual QA on the built branch (headed Chromium with macOS floating scrollbars, classic 15px regime, headless Chromium/Firefox). The fix works: thumbs under frozen columns, frozen rows, and the header are all draggable on this branch and not on `develop`; the classic regime stays fully off. No code changes needed.

What still needs to change before I can approve — description only:

1. **Headers-only scope.** Commit 949e0811a switched the gate to `needFullRender`, so a grid with `colHeaders`/`rowHeaders` and nothing frozen now gets the band and the clip (the E2E test was renamed to *"draws the track for a grid whose only overlays are its headers"*). The description still says the opposite in three places — please update them to match the code:
   - *"A grid with headers but nothing frozen is deliberately left alone… The gate is now real frozen content"*
   - Test evidence: *"The third test in the clearance spec pins the headers-only case described above: no band, and nothing clipped."*
   - Manual QA checklist step (5): *"a grid with headers but nothing frozen shows no track at all"* — a QA following this step would fail the PR on correct behavior.

2. **State the user-facing cost explicitly.** With headers-only grids in scope this now affects every default grid on macOS (default scrollbar setting), Firefox, and Windows 11 Chrome: after any scroll a 16px full-width strip along the scrolled edge swallows presses for ~1s, and stays open while the pointer rests within 26px of that edge. I checked that a hovered native overlay scrollbar on `develop` also takes clicks in that strip, so I'm fine with the decision — it just has to be visible in the PR record rather than inherited from the earlier text. A short paragraph in "What changed" is enough.

One thing I could not verify manually: the autofill handle inside an open bottom band — the native hovered track takes the press on both builds, so the `.wtBorder` pass-through is covered by the unit test only. Worth a quick check on your side on a real floating-scrollbar setup.

budnix:
@adrianspdev Good spot — all three are fixed in the description, and the cost now has its own paragraph under *What changed*. On your open question: a handle on the master works and `fill-handle-formulas.spec.ts` covers it end-to-end (it fails without the pass-through), while one on a frozen overlay is clipped away and cannot be grabbed until the track fades — now written up under *A known cost*.

adrianspdev (approved):
Description now matches the code — thanks. Re-ran manual QA on 1f594fbce in headed Chromium with macOS floating scrollbars and the classic regime: thumbs draggable under frozen columns, frozen rows, and the header; band per axis; classic regime fully off. Approving; manual-QA sign-off in a separate comment for the gate.

adrianspdev:
/manual-qa passed

Re-ran manual QA on 1f594fbce in headed Chromium with macOS floating scrollbars (\"When scrolling\") and in the classic \"Always\" regime: thumbs draggable under frozen columns, frozen rows, and the header; track band per axis only; classic regime fully off.

PR #13235, opened 2026-08-25
DEV-2604: Fix hiding plugins breaking on array className (#7427)

### LINKED ISSUE #7427: hiddenRow plugin does not accept array arguments on cellProperties.className
### Description
The hiddenRows plugin only accepts string arguments on`cellProperties.className`, but array inputs are also allowed.

### Steps to reproduce
1. Run the fiddle.
2. Replace `['test', 'test2']` with `'test test2'`.

### Demo
https://jsfiddle.net/y3w6xLr0/

### Your environment
* Handsontable version: 8.2.0
* Browser Name and version: Mozilla Firefox 83.0

budnix:
### Context

The PR fixes [#7427](https://github.com/handsontable/handsontable/issues/7427): the `hiddenRows` and `hiddenColumns` plugins break when a cell's `className` is an array.

`className` is publicly typed `string | string[]` (`handsontable/src/settings.ts`), and `addClass()` accepts both forms. But both hiding plugins cast it `as string` inside `#onAfterGetCellMeta`. That produced **two distinct defects**, and a fix for either one alone leaves the other standing:

1. **The grid renders nothing.** On the "neighbour is not hidden" branch the plugin calls `className.split(' ')`. Arrays have no `.split`, so it throws `TypeError: cellProperties.className.split is not a function` inside a render hook and the whole table fails to render. This needs **zero hidden rows** — `hiddenRows: { rows: [] }` next to an array `className` is enough, which is exactly the reporter's fiddle.

2. **The class list is silently corrupted.** On the "neighbour is hidden" branch, `className += ' afterHiddenRow'` coerces the array through `Array#toString`. `['test', 'test2']` becomes `"test,test2"`. The comma is a legal class token, so the cell ends up with one bogus class `test,test2` instead of `test` and `test2`. No error is raised, and the plugin writes that string back over the user's cell meta, destroying the original array.

Confirmed on `develop` (18.0.0) in Chrome before the fix:

| Scenario | Before |
|---|---|
| `hiddenRows: {rows: []}` + array `className` | Throws, table blank |
| `hiddenRows: {rows: [0]}`, array `className` on row 1 | Silent: `class="test,test2 afterHiddenRow"` |
| `hiddenColumns: {columns: []}` + array `className` | Throws, table blank |
| No hiding plugin (control) | Correct: `class="test test2"` |

The control shows the core handles arrays correctly — only the hiding plugins failed. This is not a regression; it reproduces back to v5.

### What changed

The root cause was the hook's parameter type. Both `#onAfterGetCellMeta` hooks typed `cellProperties` as `Record<string, unknown>`, which erases `className` to `unknown` — that is why `(cellProperties.className as string).split(' ')` compiled in the first place. They are now typed `CellProperties`, the type the hook already declares in `core/settings.ts`, and all four `as string` casts are gone. Writing `.split()` on the union is now a compile error, which is the gate that would have caught this in 2020.

On top of that, both branches normalize `className` before touching it and write back a space-joined string — the same shape `numericRenderer` and the `search` plugin already produce. The **rendered class attribute** is byte-identical to before (`addClass()` already stripped the leading space the old code produced).

The write happens **only when the value actually changes**. `afterGetCellMeta` runs on every cell meta read, and the old code reassigned `className` unconditionally on both branches; it now compares first. This matters beyond speed: an unconditional write materializes an own `className` on the cell meta, which shadows the column-level and grid-level value and stops it cascading.

Matching the marker is now by **token** rather than substring. The old add branch ran `indexOf('afterHiddenRow')` against the whole string, so a user class named `afterHiddenRowHighlight` read as "marker already present" and the cell below a hidden row silently lost its `afterHiddenRow` class. This is a user-visible change and is named in the changelog.

**Note on the stored value.** When the plugin writes, it stores a string, so `getCellMeta(r, c).className` reads back as `'a b'` rather than `['a', 'b']`. Three smaller stored-value differences come with the normalize, all improvements and none visible in the DOM: `undefined` next to a hidden row used to store `' afterHiddenRow'` with a leading space, `'foo  bar'` kept its double space, and `' '` stayed `' '`. All of this is deliberate and not a regression — the array form previously crashed the render or merged into a single bad token, so no working array behavior existed to preserve. One consequence worth stating: `setCellMeta(r, c, 'className', ['a','b'])` records `className` in `_userDefinedMetaProps`, and `getUserDefinedMetas()` re-reads it at `updateSettings` time, so the stored string is what gets re-applied and the original array is not recoverable. That is systemic rather than introduced here — `search.ts` and `numericRenderer.ts` store strings the same way. Storing a string also keeps arrays away from `contextMenu`'s align helpers, which still assume a string (see below).

`normalizeClassNames()` is promoted out of the xlsx exporter's private copy into `helpers/dom/element`, next to `addClass`/`removeClass`, and shared by all three call sites. Like the rest of that module it is reachable through `Handsontable.dom`, which is internal, undocumented surface — no documented API changes.

**Related findings, deliberately not fixed here.** Both are pre-existing, outside what #7427 covers, and need their own tests:

1. `applyAlignClassName` in `plugins/contextMenu/utils.ts` casts `cellMeta.className as string` and passes it to `prepareVerticalAlignClass`/`prepareHorizontalAlignClass`, which call `.replace()` — so context-menu alignment throws on an array `className`.
2. `numericRenderer` (`renderers/numericRenderer/numericRenderer.ts`) aliases the user's array rather than copying it, then pushes `htRight`/`htNumeric` into it. A `className` array shared between two columns is permanently polluted, and the non-numeric column picks up the numeric classes.

### How has this been tested?

New Playwright E2E covering both plugins and both branches, plus unit tests for the shared helper. The E2E was verified to **fail 9/9 without the fix and pass 9/9 with it** — not just green by construction.

```bash
# New Playwright E2E (9 tests: both plugins, both branches, plus re-showing)
cd tests && npx playwright test --project=e2e-main e2e/hiding-array-classname.spec.ts

# Unit tests for the shared helper
npm --prefix handsontable run test:unit -- --testPathPattern=helpers/dom

# The exporter now consumes the shared helper instead of its own copy
npm --prefix handsontable run test:unit -- --testPathPattern=exportFile

# Existing hiding-plugin regression suite (no behavior change on the string path)
npm --prefix handsontable run test:e2e -- --testPathPattern="hidden(Rows|Columns)"

```

Results on this branch:

- Playwright `hiding-array-classname.spec.ts` — 9 passed
- Jest `helpers/dom` — 161 passed
- Jest `exportFile` — 69 passed
- Legacy E2E `hidden(Rows|Columns)` — **639 specs, 0 failures**
- `test:types` — clean
- `eslint` — 0 errors

### Types of changes

### Related issue(s):

1. #7427
2. DEV-2604

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2604


qunabu:
Task linked: [DEV-2604 Fix hiddenRows/hiddenColumns breaking on array className (#7427)](https://app.clickup.com/t/9015210959/DEV-2604)

handsontable/src/plugins/hiddenRows/hiddenRows.ts
```diff
@@ -470,23 +470,35 @@ export class HiddenRows extends BasePlugin {
       }
     }
 
+    // `className` is publicly typed as `string | string[]`, so normalize before touching it.
+    // The normalized form is written back as a string, matching what `numericRenderer` and the
+    // `search` plugin already store. Only write when the value actually changes - this hook runs
+    // on every cell meta read.
     if (this.isHidden(row - 1)) {
-      cellProperties.className = cellProperties.className || '';
+      const classArr = normalizeClassNames(cellProperties.className as string | string[]);
```
demtario: **The root cause is the parameter type, not the cast.** This fixes the symptom but leaves the mechanism that produced it in place.

`#onAfterGetCellMeta` types `cellProperties` as `Record<string, unknown>` (`hiddenRows.ts:454`, `hiddenColumns.ts:465`). That erases every property to `unknown`, which is exactly why the old `(cellProperties.className as string).split(' ')` compiled: an unchecked assertion on `unknown`, with no union for TypeScript to object to.

Cell meta does **not** narrow `className`. `settings.ts:95` declares `className?: string | string[]` on `CellMeta`, and `CellProperties extends CellMeta` at `settings.ts:108`. That is exactly `normalizeClassNames`' parameter type, with nothing left over.

So this PR adds four more casts (`hiddenRows.ts:478`, `:491`; `hiddenColumns.ts:489`, `:502`) that all become unnecessary the moment the parameter is typed properly:

```ts
#onAfterGetCellMeta = (row: number, column: number, cellProperties: CellProperties) => {
```

and then `normalizeClassNames(cellProperties.className)` with no assertion. With that signature, the *original* buggy line is a compile error (`string[]` has no `.split`), which is the gate that would have caught #7427 in 2020. The only other use of the parameter in these hooks is the existing `cellProperties as Record<string | symbol, unknown>` for the `SKIP_ROW_ON_PASTE_BY_PLUGIN` symbol, which stays as-is.

Worth noting `search.ts#onBeforeRenderer` carries the same `Record<string, unknown>` typing, so the sibling plugin has the identical latent exposure.

handsontable/src/plugins/hiddenRows/hiddenRows.ts
```diff
@@ -470,23 +470,35 @@ export class HiddenRows extends BasePlugin {
...
 
-      if ((cellProperties.className as string).indexOf('afterHiddenRow') === -1) {
-        cellProperties.className += ' afterHiddenRow';
+      if (classArr.indexOf('afterHiddenRow') === -1) {
+        classArr.push('afterHiddenRow');
       }
+
+      const className = classArr.join(' ');
+
+      if (cellProperties.className !== className) {
```
demtario, line 487: **This guard has zero test coverage.** It is the entire content of `96ba03516`, and it is the right change, but nothing pins it.

Proven by deletion: I replaced both `if (cellProperties.className !== className)` blocks in both plugins with a bare unconditional assignment, rebuilt, and **all 9 new E2E tests still passed**.

What silently breaks if someone "simplifies" it away later: the plugin materializes an own `className` property on every cell meta it merely reads, shadowing column-level and grid-level values, so those stop cascading. CI stays green.

Cheapest pin, given the guard's whole point is *not* creating an own property:

```js
// columns: [{ className: 'a b' }], hiddenRows: { rows: [] }
expect(Object.prototype.hasOwnProperty.call(hot.getCellMeta(1, 0), 'className')).toBe(false);
```

That is `true` without the guard and `false` with it.

For the record on the guard's safety: the three own-property sites in `src/dataMap/metaManager/**` (`cellMeta.ts:341` in `getUserDefinedMetas`, `utils.ts:18` in `canBeOverwritten`, `dynamicCellMeta.ts:179`) none of them key off `className` own-ness, and no built-in cell type declares `className`, so there is no axis on which the changed own-key set misbehaves.

handsontable/src/plugins/hiddenRows/hiddenRows.ts
```diff
@@ -470,23 +470,35 @@ export class HiddenRows extends BasePlugin {
...
+    // The normalized form is written back as a string, matching what `numericRenderer` and the
+    // `search` plugin already store. Only write when the value actually changes - this hook runs
+    // on every cell meta read.
     if (this.isHidden(row - 1)) {
-      cellProperties.className = cellProperties.className || '';
+      const classArr = normalizeClassNames(cellProperties.className as string | string[]);
 
-      if ((cellProperties.className as string).indexOf('afterHiddenRow') === -1) {
-        cellProperties.className += ' afterHiddenRow';
+      if (classArr.indexOf('afterHiddenRow') === -1) {
```
demtario, line 481: **This silently changes the marker check from substring to token matching, and unlike the rest of the string path, this one *is* DOM-visible.**

Old: `(cellProperties.className as string).indexOf('afterHiddenRow') === -1`, a substring test on the whole string.
New: `classArr.indexOf('afterHiddenRow') === -1`, a token test.

Concrete difference: a user class named `afterHiddenRowHighlight` **contains** `afterHiddenRow` as a substring, so the old code concluded the marker was already present and skipped adding it. The cell below a hidden row silently lost its `afterHiddenRow` class. The new code adds it correctly. Same for `afterHiddenColumnFoo` in `hiddenColumns`.

This is a genuine fix, and it also removes a real asymmetry (the removal branch always used token matching, only the add branch was substring-based). Two asks:

1. Pin it: a test with `className: 'afterHiddenRowHighlight'` on the row after a hidden row, asserting both that class and `afterHiddenRow` are present.
2. It is a user-visible behavior change that is not in `.changelogs/7427.json`. Either extend the title or add a second entry.

handsontable/src/plugins/hiddenRows/hiddenRows.ts
```diff
@@ -470,23 +470,35 @@ export class HiddenRows extends BasePlugin {
       }
     }
 
+    // `className` is publicly typed as `string | string[]`, so normalize before touching it.
+    // The normalized form is written back as a string, matching what `numericRenderer` and the
+    // `search` plugin already store. Only write when the value actually changes - this hook runs
```
demtario, line 476: Two claims in the PR description need narrowing, and this comment block is where the reasoning is recorded.

**1. "the string path is byte-identical to before"** is true of the *rendered class attribute*, not of the *stored value*. Three stored-value differences, all improvements, all observable through `getCellMeta` and any other `afterGetCellMeta` consumer:

| input | old stored | new stored |
|---|---|---|
| `undefined`, neighbour hidden | `' afterHiddenRow'` (leading space, from `'' += ' ...'`) | `'afterHiddenRow'` |
| `'foo  bar'` (double space) | `'foo  bar'` (`split(' ')` kept the empty token, `join` restored it) | `'foo bar'` |
| `' '` (single space, truthy) | `' '` | `''` |

The DOM is unaffected in all three, because `addClass()` splits on `' '` and drops empties. `'foo\tbar'` stays one token under both, and `classList.add('foo\tbar')` throws `InvalidCharacterError` under both, so that is pre-existing. Suggest narrowing the sentence to "the rendered class attribute is byte-identical".

**2. "That is deliberate and not a regression"** is correct, but incomplete. `setCellMeta(0, 0, 'className', ['a', 'b'])` records `className` in `_userDefinedMetaProps` (`cellMeta.ts:230-253`), and `getUserDefinedMetas()` (`:329-348`) re-reads `meta[key]` at `updateSettings` time. Once the hook has overwritten the own property with `'a b'`, `updateSettings` re-applies the **string**, permanently discarding the user's array. Not a regression (that input threw before) and systemic rather than introduced here (`search.ts:352` and `numericRenderer.ts:80` do the same), but worth stating explicitly since the PR body addresses this exact question.

Also: dropping the old `if (classArr.length > 0)` guard is correct, it was dead code (`String#split` always returns at least one element).

handsontable/src/plugins/hiddenColumns/hiddenColumns.ts
```diff
@@ -481,23 +481,35 @@ export class HiddenColumns extends BasePlugin {
...
+        classArr.push('afterHiddenColumn');
       }
+
+      const className = classArr.join(' ');
+
+      if (cellProperties.className !== className) {
+        cellProperties.className = className;
+      }
+
     } else if (cellProperties.className) {
```
demtario, line 502: **Perf: this branch runs its full normalize + join + compare for every cell that has a `className`, including cells nowhere near a hidden row or column.** `afterGetCellMeta` fires from `metaManager.getCellMeta` (`index.ts:203`) and `getCellMetaTransient` (`:276`), one of the hottest read paths in the grid.

An early return when there is no marker to strip cuts it:

```js
const containAfterHiddenColumn = classArr.indexOf('afterHiddenColumn');

if (containAfterHiddenColumn === -1) {
  return; // nothing of ours to strip
}
```

Measured on an isolated re-implementation of the two hook bodies (3M calls, steady state, sink-assigned, node 22). **Caveat: this is not in-situ.** It does not go through the real `getCellMeta` path, it is not a browser, and it captures no GC pressure:

- this branch, old code: ~106 ns/call
- this branch, as written in the PR: ~79-84 ns/call (already ~25% **faster** than old, the skipped write more than pays for the extra filter pass, which is independent evidence the write-on-change guard was the right call)
- this branch, with the early return: ~38 ns/call

The add branch above regresses roughly 8-9x (~18 ns to ~160 ns), but it only fires for cells adjacent to a hidden block. Even adversarially (every other row hidden, so it owns the whole viewport) a 30x20 viewport costs ~87 microseconds per full pass. That is the unavoidable price of normalizing correctly and is not worth micro-optimizing.

**One trade-off to weigh before taking the early return, though.** It would leave arrays in cell meta on this path, which re-exposes `contextMenu`'s `applyAlignClassName` (`plugins/contextMenu/utils.ts:115`/`:117` cast to `string`, then `.replace()` in `prepareVerticalAlignClass`). Today this PR incidentally *fixes* that crash for anyone with a hiding plugin enabled, because `afterGetCellMeta` also fires on the transient path (`dynamicCellMeta.ts:82-83`, `:149`) that `predefinedItems/alignment.ts:52` reads through, so `className` is already a string by the time the align helper sees it. That is also the strongest argument for normalizing to a string rather than preserving the array shape. If you take the early return, the `contextMenu` hardening becomes a prerequisite rather than a deferrable follow-up.

handsontable/src/helpers/dom/element.ts
```diff
@@ -309,6 +309,25 @@ function filterEmptyClassNames(classNames: string[]) {
...
+/**
+ * Normalizes a `className` value to a flat array of non-empty class names.
+ * Accepts a space-separated string, an array of strings, or a nullish value.
+ *
+ * @param {string|string[]|null|undefined} className The `className` value to normalize.
+ * @returns {string[]}
+ */
+export function normalizeClassNames(className: string | string[] | null | undefined): string[] {
+  if (Array.isArray(className)) {
+    return className.filter(c => typeof c === 'string' && c.length > 0);
```
demtario: **This filter rule disagrees with `filterEmptyClassNames` ten lines above, and the divergence is observable.**

- `filterEmptyClassNames` (`:310`) keeps anything truthy: `.filter(x => !!x)`. It is what `addClass` uses (`:391`).
- `normalizeClassNames` keeps only strings: `typeof c === 'string' && c.length > 0`.

So `className: ['a', 123]`:
- **without** a hiding plugin, `baseRenderer` calls `addClass(TD, ['a', 123])`, `123` survives the truthy filter, and the cell renders `class="a 123"`.
- **with** `hiddenRows` enabled, this hook normalizes to `['a']`, writes `'a'`, and the `123` class disappears.

Out-of-contract input given `className?: string | string[]`, so the impact is low. But it is the concrete cost of adding a second near-identical normalizer ten lines below the existing one instead of unifying them, and there is a third copy in `search.ts:328-336`.

Suggested: have `addClass`/`removeClass` call `normalizeClassNames` and delete `filterEmptyClassNames`, so one filter rule governs both the DOM path and the meta path. That also makes the new export earn its place (see the separate note on `Handsontable.dom`).

Worth recording the property that makes the fix itself correct: `normalizeClassNames` returns a **fresh** array from `.filter()` on both branches, so `classArr.push('afterHiddenRow')` cannot mutate the caller's array. Had it returned `className` directly, a grid-level array (one instance shared by every cell through the meta prototype chain) would accumulate markers. Contrast `numericRenderer.ts:59`, which does alias and then push.

handsontable/src/helpers/dom/element.ts
```diff
@@ -309,6 +309,25 @@ function filterEmptyClassNames(classNames: string[]) {
...
 }
 
+/**
+ * Normalizes a `className` value to a flat array of non-empty class names.
+ * Accepts a space-separated string, an array of strings, or a nullish value.
+ *
+ * @param {string|string[]|null|undefined} className The `className` value to normalize.
+ * @returns {string[]}
+ */
+export function normalizeClassNames(className: string | string[] | null | undefined): string[] {
```
demtario, line 324: Small flag on the surface question, since `96ba03516` introduced this as `_normalizeClassNames` and `31f4746fa` dropped the prefix on the reasoning that `Handsontable.dom` is private, undocumented surface.

It does actually ship. `index.ts:24` and `:132` spread every non-underscore export of this module onto `HOT.dom`, `base.ts:104` types `dom` as `typeof import('./helpers/dom/element')`, and the built `tmp/helpers/dom/element.d.ts` carries `export declare function normalizeClassNames(...)`. The docs also teach the namespace directly (`cell-renderer.md:444` uses `Handsontable.dom.empty()`).

That said, nothing gates it and no convention is breached: `src/__tests__/publicAPI.spec.js` is a positive-assertion list, not an exhaustive one (32 exports of this module are already absent from it), nothing under `handsontable/test/` snapshots the namespace, and most of this file is already de-facto public and unversioned. Additive, not breaking, no docs owed.

So this is a team call rather than a defect: either keep it internal (the `_` prefix is the working mechanism, `helpers/mixed.ts` already uses it for the license helpers), or publish it deliberately and reuse it in `addClass`/`removeClass`, which duplicate its split-and-filter logic.

JSDoc itself is compliant: multiline block style, `@param` and `@returns` present. No `@since`, consistent with every other export here.

handsontable/src/plugins/exportFile/types/xlsx/cell-style.ts
```diff
@@ -1,4 +1,5 @@
 import { isDefined } from '../../../../helpers/mixed';
+import { normalizeClassNames } from '../../../../helpers/dom/element';
```
demtario, line 2: The promotion out of this file is the right call, and the move is verbatim (only `function` becomes `export function`; the three call sites at `:286`, `:303`, `:381` are unchanged, the per-document `WeakMap` cache keyed on the joined className string still works, and the 69 `exportFile` unit tests pass). No import cycle either: `element.ts` imports only `helpers/a11y`, `helpers/browser`, `helpers/errors`, `helpers/console`, and both plugins already imported `addClass` from it.

But the consolidation is only two-thirds done. Two hand-rolled copies of the identical `Array.isArray ? ... : split(' ')` normalize remain:

- `renderers/numericRenderer/numericRenderer.ts:58-69`
- `plugins/search/search.ts:328-336`

`numericRenderer` is the worse of the two: line 59 does `classArr = cellProperties.className as string[]`, taking the user's array **by reference**, then pushes `htRight`/`htNumeric` into it at `:71`/`:77`. A grid-level `className: ['a', 'b']` is one array instance shared by every cell through the meta prototype chain, so those classes get appended to the user's own array and accumulate across cells, and a non-numeric column sharing that array picks up the numeric classes.

Both are pre-existing and I agree they are out of scope here, with their own tests needed. Two asks instead:

1. A follow-up task for the two migrations, so the helper actually ends up with one implementation.
2. Per the mandatory checklist item 3 (update `AGENTS.md` when a change introduces a convention or gotcha), one line in `handsontable/AGENTS.md`: cell `className` is `string | string[]` (documented at `metaSchema.ts:753-796` with an array example), never cast it to `string`, normalize with `normalizeClassNames`, and never push into the value you were handed. This bug family has now recurred at four separate `className` sites, which is what earns the written note.

tests/e2e/hiding-array-classname.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  test.beforeEach(async ({ page, theme, bundle }) => {
+    grid = new HidingArrayClassNamePage(page, theme, bundle);
+    await grid.goto();
+  });
+
+  test.describe('hiddenRows', () => {
+    test('builds the grid when nothing is hidden', async () => {
+      // The issue's own fiddle. `.split` on the array threw here and left the table blank.
+      expect(await grid.initError(HidingArrayClassNamePage.ROWS_NONE)).toBeNull();
```
demtario: These two `initError` assertions (here and `:71`) can never execute.

`HidingArrayClassNamePage.goto():50-56` already loops **all four** grid ids and throws on a non-null `data-init-error`, and `goto()` runs in `beforeEach`. Verified empirically: with `develop`'s plugins restored, this test failed inside `goto()` at `HidingArrayClassNamePage.ts:54`, never reaching line 26.

Not a coverage loss, the precondition still catches the throw and red is still red. But the assertion that documents the loud half of the bug cannot fail, and every failure gets attributed to the hook rather than to the scenario, which is the opposite of what the fixture comment says it is trying to achieve. Pick one: drop the precheck loop from `goto()` and let each test assert its own grid, or drop these two lines and keep the precheck.

**On the suite as a whole, it is good, and it genuinely fails without the fix.** I verified both defects are covered independently by patching only defect 1 (array-safe `.split`, `+=` left alone) and rebuilding: **5 passed / 4 failed**, the failures receiving `["test,test2"]` and `["afterHiddenColumn","test,test2"]`. Reading the class attribute as a token list rather than a substring is what makes the silent half detectable, and the one-shot `getAttribute` after `showRows()` is not a flake risk, since `core.render()` (`core.ts:2809`) and `tableView.render()` (`tableView.ts:226`) are fully synchronous.

Gaps worth closing while you are here:

1. **`cellMetaClassName` is asserted for `hiddenRows` only, never `hiddenColumns`.** And per the run above, that test carries no defect-2 coverage at all: it targets `ROWS_NONE`, which goes down the removal branch, and it passed with defect 2 standing.
2. **The array is only ever set at cell level** (`cell: [{ row, col, className: ARRAY_CLASS_NAME }]`, all four grids). Column-level and grid-level are where one array instance is prototype-shared across every cell, which is exactly where an in-place mutation would be catastrophic. One extra grid with a grid-level `className: ARRAY_CLASS_NAME` plus `hiddenRows: { rows: [0] }` would cover it and would exercise the removal branch across the whole viewport instead of a single cell.
3. **Both plugins on one grid** is untested: a cell after both a hidden row and a hidden column, asserting each branch strips only its own marker. I traced it by hand and it is correct in both hook orders, but nothing pins it.
4. **The string path has no regression test at all**, which matters given the leading-space and whitespace-collapse changes to the stored value.
5. Minor: `element.unit.ts:423`, `expect(normalizeClassNames(['test', 'test2'])).not.toContain('test,test2')` holds for any implementation that returns tokens at all, so it documents #7427 rather than guarding it. The `arrayContaining` case above already covers the real property. (The round-trip case at `:426` is genuinely load-bearing, not lumping that one in.)

demtario (commented):
High-effort review, two independent passes (a generic deep-dive plus the repo's Handsontable reviewer lens), consolidated. **Verdict: approve with comments.** No blocker, no major. I could not find an input where the new code produces a worse user-visible result than the old, and the E2E genuinely fails without the fix on both defects independently (patching only defect 1 gives 5 passed / 4 failed).

Gates I ran: `eslint` on all five changed core files clean; `test:unit` helpers/dom 161 passed; `test:unit` exportFile 69 passed; `test:types` passed; Playwright `e2e-main` and `e2e-main-min` 9 + 9 passed.

Nine inline comments follow. The two I would act on before merge are the hook parameter type (root cause, not symptom) and the untested write-on-change guard. Everything else is a one-line cleanup, a coverage addition, or a team judgment call.

Things I checked and am explicitly **not** raising: the blank line before `} else if` (`padded-blocks` is not enabled, eslint is clean); the missing `publicAPI.spec.js` entry (positive-assertion list, 32 exports already absent); the ` - ` in the new comments (both files already used hyphens as separators, e.g. `hiddenRows.ts:245`); the changelog entry (valid on all six fields, `issuesOrigin: "public"` with `issueOrPR: 7427` is the correct public-issue case); breaking-change status (not breaking, the array path threw before, and `metaSchema.ts:753-796` documents the input shape without promising the read-back shape); and the legacy Jasmine suite (no spec in either plugin's tree asserts a stored `className` string, every `afterHidden*` assertion is rendered HTML, which is byte-identical).

handsontable/src/plugins/hiddenRows/hiddenRows.ts
budnix: You're right, the `Record<string, unknown>` was the actual hole, not the cast. Typed both hooks `CellProperties` and dropped all four casts in `bd136610c6`.

handsontable/src/plugins/hiddenRows/hiddenRows.ts
budnix: Good catch — pinned in `bd136610c6` with the `hasOwnProperty` check you suggested. Confirmed it goes red when I make the write unconditional.

handsontable/src/plugins/hiddenRows/hiddenRows.ts
budnix: Good spot — added the `afterHiddenRowHighlight` case in `bd136610c6` and extended the changelog title to name the change.

handsontable/src/plugins/hiddenRows/hiddenRows.ts
budnix: Fair point — narrowed both claims in the PR description to the rendered class attribute, and added the `updateSettings` note. Thanks for confirming the dead `length > 0` guard too.

handsontable/src/plugins/hiddenColumns/hiddenColumns.ts
budnix: Thanks for actually measuring this. Leaving it as-is though, for the reason in your own last paragraph: the early return puts arrays back into cell meta and re-opens the `contextMenu` crash, and this branch is already faster than the old code. Both go into the follow-up together.

handsontable/src/helpers/dom/element.ts
budnix: Agreed, two filter rules ten lines apart was silly. `normalizeClassNames` now goes through `filterEmptyClassNames` in `bd136610c6`, with a unit test tying the two together. Left `addClass` alone — changing that one's behavior felt too broad for a bug fix.

handsontable/src/helpers/dom/element.ts
budnix: Yep, my mistake on the earlier flip-flop here. Team call landed on keeping the plain name — `Handsontable.dom` is not documented anywhere, so there is no promise to break.

handsontable/src/plugins/exportFile/types/xlsx/cell-style.ts
budnix: Good point on the written note — added the `className` entry to `handsontable/AGENTS.md` in `bd136610c6`. The `numericRenderer` and `search` migrations are going into one follow-up task.

tests/e2e/hiding-array-classname.spec.ts
budnix: Nice catch, those two could never run — dropped them in `bd136610c6` and kept the precheck. Also closed all four coverage gaps (19 tests now). One correction: I first put the own-property pin on the grid-level array grid and it failed, because an array always differs from its normalized string, so a write there is correct — the pin lives on the string path instead.

handsontable/src/helpers/dom/element.ts
```diff
@@ -309,6 +309,25 @@ function filterEmptyClassNames(classNames: string[]) {
...
+  if (Array.isArray(className)) {
+    return filterEmptyClassNames(className);
```
demtario, line 326: Non-blocking. Routing through `filterEmptyClassNames` is the better call than my suggestion (it closes the divergence without changing `addClass` behavior inside a bug fix), but it does make the declared return type slightly optimistic.

`filterEmptyClassNames` filters on `!!x`, so a truthy non-string survives: `normalizeClassNames(['a', 123])` returns `['a', 123]` at runtime while the signature says `string[]`. The new unit test is honest about it (`.map(String)` before comparing against `classList`), which I think is the right trade, and matching the DOM path matters more than the type being exact for input that is already out of contract.

Worth one line on the JSDoc noting the filter is truthiness-based and shared with `addClass`, so the next reader does not "fix" the type by reintroducing a second filter rule. No code change needed.

.changelogs/7427.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed a bug where enabling the `hiddenRows` or `hiddenColumns` plugin alongside a cell `className` given as an array stopped the grid from rendering, or merged the array entries into a single invalid CSS class. The `afterHiddenRow` and `afterHiddenColumn` marker classes are now also matched as whole class names, so a cell whose own class merely contains the marker name no longer misses out on the marker.",
```
demtario: Non-blocking, style only. The added sentence is accurate and I did ask for the marker change to be named, so this is on me. It does push the entry to two sentences and ~370 characters, against the short-sentences rule in `.ai/DOC-STANDARDS.md`.

Something like this keeps the fact and loses half the words:

> The `afterHiddenRow` and `afterHiddenColumn` markers are now matched as whole class names, so a cell whose own class contains the marker name still gets the marker.

Take it or leave it.

handsontable/AGENTS.md
```diff
@@ -73,6 +73,7 @@ Gotcha: Filters `conditionCollection` uses physical indexes, `getDataAtCol()` us
...
... write back `join(' ')` — the shape `numericRenderer` and `search` already store. Two traps, both of which have shipped: `(meta.className as string).split(' ')` throws `.split is not a function` on an array, inside a render hook, so the grid renders blank (#7427); and `meta.className += ' marker'` coerces the array through `Array#toString`, silently merging `['a','b']` into one bogus `a,b` class. A third: a grid-level or column-level array is **one instance shared by every cell** through the meta prototype chain, so pushing into it rather than into a copy leaks the marker onto the whole grid (`numericRenderer.ts` still aliases). Match the marker by token, not substring — `indexOf` on the joined string treats a user class named `afterHiddenRowHighlight` as the marker already being present. And in a hook that only sometimes changes the value, compare before assigning: `afterGetCellMeta` runs on every meta read, and an unconditional write materializes an own property that shadows the column-level and grid-level cascade. Type the hook parameter `CellProperties`, not `Record<string, unknown>` — the latter erases the union to `unknown` and is what let the cast compile in the first place.
```
demtario: Non-blocking. Content is right, and I am glad it records that `numericRenderer.ts` still aliases rather than pretending the family is fully fixed.

It is one ~1400-character bullet carrying five separate rules (the union, normalize-and-copy, the shared-instance trap, token-not-substring, compare-before-assign, and the parameter type). That is a lot to hold in one dash. Splitting it in two would survive skimming better:

- one bullet on the value: `string | string[]`, normalize with the shared helper, never cast, never push into what you were handed, watch the prototype-shared instance;
- one on hook hygiene in `afterGetCellMeta`: match by token, compare before assigning, type the parameter `CellProperties`.

Purely presentational, no factual change.

handsontable/src/plugins/exportFile/types/xlsx/cell-style.ts
demtario, line 2: Non-blocking, but the one item I would not want to lose track of: could you put the follow-up task id in this thread?

Three separate things are now deferred into it, and two of them are decisions taken on this PR rather than pre-existing debt:

1. Migrating `numericRenderer.ts:58-69` and `search.ts:328-336` onto the shared helper, which also fixes `numericRenderer`'s aliasing mutation of a user array.
2. Hardening `contextMenu`'s `applyAlignClassName` (`utils.ts:115`/`:117`) against an array `className`.
3. The branch-2 early return, which you correctly declined here **because** it depends on 2 landing first.

Since 3 is gated on 2, and the perf note only makes sense with that ordering, a task id in the thread keeps the chain legible to whoever picks it up. `AGENTS.md` now carries the convention, so the knowledge is safe either way; this is about the work not evaporating.

demtario (dismissed):
Approving. All nine threads addressed, eight with changes and one declined for a reason I agree with. Resolved them from my side.

**I re-ran the two new pins as counter-tests rather than taking them on faith, and both are genuine:**

- Made the write unconditional in both plugins, rebuilt, reran: `PASS (18) FAIL (1)`, failing exactly on `leaves the cascade intact on rows it does not mark` (`expected false, received true` at `spec.ts:194`). One test, precisely the right one.
- Reverted the add branch to the old substring check, rebuilt, reran: fails on `still gets the marker added` with `Received array: ["afterHiddenRowHighlight"]`.

**And your correction on the pin's placement is right, and better than what I proposed.** An array always differs from its normalized string, so a write on the grid-level-array grid is correct and my suggested assertion would have been wrong there. The string path is the only place the guard is observable at all, since the rendered classes are identical either way. The fixture comment says exactly that.

Gates on `bd136610c`: `eslint` clean on all four changed core files; `test:unit` helpers/dom 161 passed; `test:unit` exportFile 69 passed; `test:types` passed (build-types, downlevel-dts, test/types); Playwright 19 passed on `e2e-main` and 19 passed on `e2e-main-min`. Worktree restored clean after both counter-tests.

On the perf thread specifically: declining the early return is the correct call, and for the right reason. It reopens the `contextMenu` crash this PR incidentally closes, and that branch is already faster than the old code, so there is nothing to buy.

Four inline notes below. None of them block the merge, three are cosmetic, and the fourth is just asking for the follow-up task id so the deferred chain stays traceable.

handsontable/src/helpers/dom/element.ts
budnix: Good point — noted it on the JSDoc in `2f3578e475`, including the warning not to add a second filter rule to make the type exact.

.changelogs/7427.json
budnix: Fair point — took your wording in `2f3578e475`.

handsontable/AGENTS.md
budnix: Agreed, one bullet was carrying too much — split into the value rules and `afterGetCellMeta` hook hygiene in `2f3578e475`.

handsontable/src/plugins/exportFile/types/xlsx/cell-style.ts
budnix: Yep, tracked as DEV-2618 with all three items and the ordering — the early return is blocked on the `contextMenu` fix.

demtario (approved):

Issue #13229, opened 2026-08-24
nestedHeaders innerHTML sanitization


Lukortech:
### Describe the bug

Hello, during my overview of the salesforce-lwc repository I've noticed the salesforce API is being called for 
[nestedHeaders](https://github.com/handsontable/salesforce-lwc/blob/ff67e3ecb39563a4669fc2aa36d9e381b3c8cd26/force-app/main/default/lwc/hotGrid/hotGrid.js#L42) it struck me a bit as to how these are handled and by looking around there's no sanitization (which is documented as well in the AI overview [1]) and with content uncontrolled coming from shopify there could be a chance to manipulate the nested headers [2] to embed malicious scripts inside of the table itself. 

Consider your [demo page](https://demos.handsontable.com/?docs=guides%2Fcolumns%2Fcolumn-groups%2Fjavascript%2Fexample1.js&v=18.0.0) where in he line 21 you could potentially inject 
```html
 ['A', { label: 'abcdd<img src=x onerror​="alert(1);">', colspan: 8 }, 'C'],
```
using said API. 
I know this could be reported in the corresponding repository but I decided to let you know here as it seems the string type is not forced here (as in sanitized) and could lead to potential problems in other areas as well. Question here would be not strictly if the code is called from the fastInnerText but rather from the api simply replacing the headers inside of settings of `[hot](https://github.com/handsontable/salesforce-lwc/blob/ff67e3ecb39563a4669fc2aa36d9e381b3c8cd26/force-app/main/default/lwc/hotGrid/hotGrid.js#L46)`

I would love to hear your thoughts and help if needed. 
Cheers! 

[1](https://github.com/handsontable/handsontable/blob/b312dea1e9934e242aa092e2016acfde19d6a20f/handsontable/.ai/CONCERNS.md?plain=1#L72)
[2](https://github.com/handsontable/handsontable/blob/b312dea1e9934e242aa092e2016acfde19d6a20f/handsontable/src/helpers/dom/element.ts#L588)

### Video/Screenshots

<img width="1579" height="636" alt="Image" src="https://github.com/user-attachments/assets/e25bcc19-bbec-423b-baf4-6a5d5520c755" />

### Handsontable version

18.0.0

### Your environment

windows 10 latest brave browser ver (1.93.138 (Official Build) (64-bit) Chromium: 151.0.7922.173)

qunabu:
Thanks for the detailed write-up — this has been logged for internal triage.

[...]

PR #7368, opened 2020-11-03
Protect Autocomplete editor from XSS

budnix:
### Context

The PR fixes the following issues:
 * Disables values sanitizing when the `allowHtml` option is enabled (default value is `false`). It applies to the Autocomplete and Dropdown editors;
 * Disables values sanitizing for HTML renderers.

Additionally, I've exported the DOMPurify lib to the window object. This will allow adjusting the lib and control of how the HTML content should be filtered within the Handsontable settings that accept that kind of strings.

After [that changes](https://github.com/handsontable/handsontable/issues/7292), the Handsontable can stop rendering some of the values. This is true for properties that accept values passed as HTML string. Mostly I'm referring to custom tags, custom attributes, or some of the build-in elements such as `iframe`. To make them working again the custom code for DOMPurify hooks have to be added. It extends the build-in allowlist. 

This allows rendering custom tags and/or attributes:
```js
DOMPurify.addHook('uponSanitizeElement', function(node, data) {
  if (node.nodeName && node.nodeName === 'MY-FANCY-INPUT') {
    data.allowedTags[data.tagName] = true;
  }
});
DOMPurify.addHook('uponSanitizeAttribute', function(node, data) {
  if (node.hasAttribute('my-id')) {
    data.allowedAttributes[data.attrName] = true;
  }
});
```  

The documentation about [custom renderers](https://handsontable.com/docs/7.4.2/demo-custom-renderers.html#page-cell) should be updated. DOMPurify should be used there. 

To summarize the PR, this is a list of fields which are sanitized:
 * The `name` property, part of the _ContextMenu_ plugin configuration;
 * The `name` property, part of the _DropdownMenu_ plugin configuration;
 * The clipboard;
 * The `source` option is sanitized by default. It can be bypassed by setting `allowHtml` as `true`. The option is used by Autocomplete and Dropdown editors.

List of fields which aren't sanitized:
 * `rowHeaders`;
 * `colHeaders`;
 * `title`;
 * The `label` property, part of the _NestedHeaders_ plugin.

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature or improvement (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Additional language file or change to the existing one (translations)

### Related issue(s):
1.
2.
3.

### Checklist:

- [x] My code follows the code style of this project,
- [x] My change requires a change to the documentation.

PR #13236, opened 2026-08-25
DEV-2600: Apply the configured sanitizer on the surfaces that skipped it (#13229)

demtario:
### Context

The PR fixes three surfaces that ignored a configured `sanitizer`, and documents the two that skip it on purpose.

GitHub issue #13229 reports that `nestedHeaders` labels are written to the DOM without sanitization, so API-supplied header text can inject markup. Investigation reproduced the symptom but corrected the framing:

**The reported header behavior is by design and does not change.** 18.0 removed the built-in DOMPurify, so every HTML surface is pass-through by default; the grid-level `sanitizer` option (17.0+) is the documented opt-in, and the grid warns once when it writes raw HTML without one. Plain `colHeaders` behaves identically, so `nestedHeaders` is not special, and PR #7368 (2020) already listed header `label` as explicitly unsanitized.

**What was actually broken:** three surfaces ignored `sanitizer` even when the developer set it, which nothing in the documentation permits.

1. **Clipboard paste, private MIME type** (`copyPaste.ts`). `application/ht-source-data-json-html` was parsed and handed to `htmlToGridSettings()` before the sanitizer was even looked up, while the `text/html` branch a few lines below was sanitized. Handsontable writes that type from its own copy handler, but the clipboard is not a trusted channel: any page can set the same type from its own `copy` handler. Verified in headless Chrome that an `` in such a payload executes on a grid whose owner had configured a sanitizer.

2. **`parseTable.ts` sink**. The payload above reached `insertAdjacentHTML()` on a detached `<div>`. Detached is not inert: the document owning that element has a browsing context, so the image still loaded and its handler still ran. Parsing with `DOMParser.parseFromString()` produces a document with no browsing context, which closes the sink regardless of what the caller's sanitizer does, or whether one is configured at all.

3. **Password renderer**. Called `fastInnerHTML(TD, value)` with no sanitizer argument, so a configured sanitizer was never applied and the warning named the generic `"innerHTML"` surface instead of the renderer. Not an XSS vector (the renderer's own `valueFormatter` always hashes the value), but the content is documented as covered.

Two further defects surfaced while fixing those:

4. **Nested-header ghost table** sanitized the same label under `'innerHTML'` where the main table uses `'header'`, so a context-aware sanitizer applied two rule sets to one value, and the measured width could not match what the user sees. It also never warned when no sanitizer was configured.

5. **Warn-scope inconsistency**. `warnOnce` keys a `WeakMap` on object identity, and the sinks disagreed on the scope object, breaking the "one warning per Handsontable instance" promise in `SANITIZER_WARN_KEY`'s comment. `tableView` additionally passed `rootGridElement ?? undefined`, which does not disable the scope but selects the default parameter and falls through to a module-level object, collapsing the header warning process-wide. Dialog wrapped the sanitizer in `(html, ctx) => sanitizer(html)`, discarding the context so a context-aware sanitizer never saw `'dialog'`.

### Known trade-off: object-based source data and escaping sanitizers

Sanitizing the private clipboard type has a cost, and it is worth being explicit about it.

That branch carries the object-keyed source data that makes Handsontable-to-Handsontable paste restore `{ id: 9, value: 'B1' }` rather than the string `'B1'`. It previously reached `htmlToGridSettings()` raw, so it always round-tripped, whatever the sanitizer did to the `text/html` branch.

Now it goes through the same sanitizer. A sanitizer that **strips** unsafe markup (DOMPurify and friends) leaves the payload's `<table>` intact and nothing changes. A sanitizer that **escapes** HTML instead turns the table into text, `htmlToGridSettings()` finds no table, and the pasted cell receives the displayed value instead of the original object. This affects only grids that use [`parsePastedValue`](https://handsontable.com/docs/javascript-data-grid/api/options/#parsepastedvalue), since that option is what enables the restore branch at all.

Not treated as a breaking change: the previous behavior was "attacker-controlled HTML reaches the parser unsanitized", which is not a contract anyone can rely on. The degradation is pinned by a test that fails against the pre-fix code, and named in the security guide and in the 17.1→18.0 migration guide's "Who is affected" list.

### The `html` cell type and `allowHtml` stay raw, deliberately

These were considered and are **not** changed. PR #7368 (`c03f2bc42e`, 2020) disabled sanitizing for both on purpose, and it held through the DOMPurify era: pre-removal `fastInnerHTML` had three branches (function → apply, `true` → DOMPurify, `false` → raw) and `htmlRenderer` passed the literal `false`. A configured sanitizer has therefore never reached `html` cells at any point in the library's history, and `allowHtml`'s own JSDoc warning only parses if the content is never sanitized.

What was wrong is the documentation: the `sanitizer` JSDoc and the security guide both promised coverage of "cell content" with no carve-out. This PR aligns the documentation to the behavior, so nothing here is a breaking change. Whether a configured `sanitizer` *should* cover those two surfaces is a separate product question that needs the deprecation process; it is intentionally out of scope.

### Also in this PR

- New `handsontable/src/utils/sanitizer.ts` with `getSanitizer()` and `sanitizeHTML()`, so the eight call sites share one lookup and one warn scope. Deliberately not registered in `src/index.ts` — it must not leak onto the public `Handsontable.helper` namespace.
- The autocomplete dropdown's `TD.innerHTML = cellValue` (the only HTML sink that bypassed `fastInnerHTML` entirely) now goes through the helper with `false`. Output is unchanged; the value is that raw HTML has one policy point instead of two.
- Removed `as { sanitizer?: (html: string) => string }` casts from `menuItemRenderer`, `ghostTable`, `dialog`, and `notification/ui`. Each declared a one-argument sanitizer, narrower than the real two-argument contract, and the option is already typed on `GridSettings`.
- Documentation: covered/excluded tables in the security guide, the real set of context strings in the `sanitizer` JSDoc, notes on `allowHtml` and `nestedHeaders` (neither mentioned `sanitizer` at all, which is the discoverability half of the report), `nestedHeaders` added to the 17.1→18.0 migration guide's affected-surface list, and refreshed `.ai/CONCERNS.md`.

### Known limits, recorded rather than fixed

- `Core#toTableElement()` (`core.ts:6448`) writes `instanceToHTML()` output into a live-document element with `insertAdjacentHTML`, and `instanceToHTML` escapes cell data but not headers. A `DOMParser` swap does not help — the function must return a node the caller inserts, and adopting it reactivates the payload. Public API with no internal callers, and the same label executes when rendered into a real `<th>` anyway. Named in `.ai/CONCERNS.md` with the real fix (escaping headers in `instanceToHTML`).
- The `'dialog'` context covers string `content` only. A `confirm` dialog's `title`, `description`, and button labels go through `stripTags()` and never reach the sanitizer. Stated in the security guide rather than changed.
- `GridSettings.sanitizer` is still typed `(html: string, ...args: any[]) => string` rather than the new `SanitizerFn`. Narrowing it would improve IDE completion on the `source` argument this PR makes load-bearing, but it also stops compiling for anyone who declared a third parameter, so it belongs in its own PR.

### How has this been tested?

Every behavioral change was verified red-then-green: the fix was reverted, the test observed failing, then restored. For the paste vector this meant rebuilding `dist` so the Playwright leg was exercised against the unfixed bundle — the injected handler fires there.

```bash
npm --prefix handsontable run eslint            # 0 errors
npm --prefix handsontable run stylelint         # pass
npm --prefix handsontable run test:types        # pass
npm --prefix handsontable run test:unit         # 3742 passed / 313 suites
npm --prefix handsontable run test:e2e          # 9690 specs, 0 failures
npm --prefix handsontable run test:walkontable  # 816 specs, 0 failures
cd tests && npx playwright test --project=e2e-main e2e/sanitizer.spec.ts   # 3 passed
npm --prefix wrappers/react-wrapper run test    # 75 passed
npm --prefix wrappers/vue3 run test             # 33 passed
npm --prefix wrappers/angular-wrapper run test  # 291 passed
```

Tests added:

- `handsontable/src/utils/__tests__/sanitizer.unit.ts` — new, 10 cases covering the resolver, the plain-text short-circuit, and warn-once-per-instance.
- `handsontable/src/__tests__/settings/sanitizer.spec.js` — a `nestedHeaders` label sanitized under `'header'` and never under a second context, and one warning per grid when nested headers and `colHeaders` both carry HTML.
- `handsontable/src/plugins/copyPaste/__tests__/settings/sanitizer.spec.js` — the private MIME sanitized, the warning it now raises, the object-based source-data round-trip under a stripping sanitizer, and the degradation under an escaping one.
- `handsontable/src/utils/__tests__/parseTable.spec.js` — markup alongside a pasted table does not run with no sanitizer configured, plus a non-regression case for the Excel generator `<meta>` path that moved to the parsed document.
- `handsontable/src/renderers/passwordRenderer/__tests__/passwordRenderer.spec.js` — sanitizer applied, `"password"` named in the warning.
- `handsontable/src/plugins/dialog/__tests__/dialog.spec.js` — the sanitizer receives the `'dialog'` context, which the old wrapper discarded.
- `handsontable/src/editors/autocompleteEditor/__tests__/autocompleteEditor.spec.js` — no missing-sanitizer warning for the dropdown, and plain-text items render unchanged. The existing `allow html mode` XSS specs stay green untouched, which is the proof that change is output-neutral.
- `tests/e2e/sanitizer.spec.ts` — new Playwright spec (the `sanitizer` option had no Playwright coverage). Drives a real `paste` event with a real `DataTransfer`, which the legacy suite's mocked `DataTransferObject` cannot.

For transparency, four of the added cases are regression guards rather than proofs — they pass unmodified against `develop`: the two autocomplete dropdown cases, the `parseTable` generator case, and the header warn-once count (which does catch the new warn-scope divergence, but no pre-existing double warning existed to fix). Everything else fails against `develop`.

Manual verification in Chrome against a standalone repro page covering all five reported cases.

### Types of changes

No default setting value moves, no API is renamed or removed, no public signature changes, and `html` / `allowHtml` output is byte-identical. The one behavior change users can observe is the source-data trade-off described above, which is documented in the migration guide.

### Related issue(s):

1. DEV-2600
2. #13229

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2600

---

> [!NOTE]
> **High Risk**
> Changes authentication-adjacent XSS defenses (clipboard paste, HTML parsing, and multiple DOM write paths); regressions could re-enable script execution or break object-based paste when sanitizers escape HTML.
> 
> **Overview**
> This release closes **XSS gaps** where HTML was written or parsed without the grid’s **`sanitizer`**, and aligns **context strings** so context-aware sanitizers behave consistently.
> 
> **Security fixes:** Clipboard paste now sanitizes **`application/ht-source-data-json-html`** under **`CopyPaste.paste.sourceData`** (previously only **`text/html`** was covered). **`htmlToGridSettings`** parses string input with **`DOMParser`** so pasted markup is read in an inert document instead of a live-document scratch element where scripts could run. **Password** cells route displayed HTML through the sanitizer with source **`password`**.
> 
> **Sanitizer plumbing:** New **`utils/sanitizer.ts`** (`**getSanitizer**`, **`sanitizeHTML**`) centralizes option lookup, warnings, and **`hot.rootElement`** “warn once” scope across headers, copy/paste, dialog, notification, context menu, select editor, and nested-header measurement. Nested-header **ghost table** labels use **`header`** (not **`innerHTML`**) and mirror **`fastInnerHTML`**’s plain-text vs markup rules. Dialog content is passed to the sanitizer with **`dialog`** as the second argument. **`fastInnerHTML`** treats a sanitizer that returns **`undefined`** as empty output.
> 
> **Intentionally unchanged:** **`html`** cells and **`allowHtml`** autocomplete/dropdown sources stay raw (documented exclusions); autocomplete dropdown now uses **`fastInnerHTML`** with **`false`** for a single policy point.
> 
> **Docs & migration:** Security guide tables for covered/excluded surfaces, **18.1→18.2** migration (source renames and clipboard trade-off for escaping sanitizers), changelog entry, and broad test/e2e coverage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 67277447ccfcdfdd2d01a373371c02ba19cf499c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2600 [GH-13229] Clarify/harden XSS-sanitization gap for NestedHeaders `label` and other header content](https://app.clickup.com/t/9015210959/DEV-2600)

handsontable/src/plugins/nestedHeaders/utils/ghostTable.ts
```diff
@@ -278,16 +272,17 @@ class GhostTable {
...
   ) {
-    const label = typeof sanitizer === 'function'
-      ? sanitizer(headerSettings.label, 'innerHTML')
-      : headerSettings.label;
+    // The same label the main table writes through `updateCellHeader()`, so it must be sanitized
+    // under the same `'header'` context - a context-aware sanitizer would otherwise apply one rule
+    // set to the rendered header and another to the copy measured here.
+    // `String()` mirrors `tableView.updateCellHeader()`. `label` is typed `string`, but it comes
+    // from user configuration, so `nestedHeaders: [[2024]]` puts a number there.
+    const label = sanitizeHTML(this.hot, String(headerSettings.label), 'header');
```
budnix: Good that the ghost table and the real header now agree on the context. One thing though - the old `'innerHTML'` value shipped in 18.0.0, and the `sanitizer` docs named it as an example, so anyone branching on `source === 'innerHTML'` for nested header labels loses that branch on upgrade. The changelog says `breaking: false` and the migration guide doesn't mention it, so it would be good to either flip the flag or add a line there. Curious what you think - the render path already used `'header'`, so those users were getting two different contexts for the same label anyway.

handsontable/src/plugins/copyPaste/copyPaste.ts
```diff
@@ -928,26 +927,20 @@ export class CopyPaste extends BasePlugin {
 
     if (event && typeof event.clipboardData !== 'undefined') {
       const clipboardData = event.clipboardData!;
-      const sourceDataHTML = clipboardData.getData(SOURCE_DATA_HTML_MIME_TYPE);
+      // `SOURCE_DATA_HTML_MIME_TYPE` is written by Handsontable's own copy handler, but the
+      // clipboard is not a trusted channel: any page can set the same type from its own `copy`
+      // handler. Sanitize it exactly like the `text/html` branch below, under the same context.
+      const sourceDataHTML = sanitizeHTML(
+        this.hot, clipboardData.getData(SOURCE_DATA_HTML_MIME_TYPE) ?? '', 'CopyPaste.paste'
```
budnix: Both clipboard types now share the `'CopyPaste.paste'` context, so a sanitizer can't tell the private payload apart from untrusted `text/html`. That's what forces the `parsePastedValue` regression you documented in the migration guide - with its own context (say `'CopyPaste.paste.sourceData'`) users could let the internal payload through and still escape the real HTML. There's a cost side too: an HT-to-HT copy writes both types with a full `_dataToHTML()` table, so the sanitizer now runs over the same data twice on every internal paste. Worth deciding if that's the tradeoff you want.

handsontable/src/utils/sanitizer.ts
```diff
@@ -0,0 +1,58 @@
...
+ * The warning is bound to `hot.rootElement`, the scope every other surface uses, so all of them
+ * collapse into a single message per Handsontable instance.
+ *
+ * @param {object} hot The Handsontable instance.
+ * @param {string} html The HTML string to sanitize.
+ * @param {string} context The write surface, passed to the sanitizer and named in the warning.
+ * @returns {string} The sanitized string, or the input unchanged when no sanitizer is configured.
+ */
+export function sanitizeHTML(hot: HotInstance, html: string, context: string): string {
+  if (!HTML_CHARACTERS.test(html)) {
```
budnix: This short-circuit is new for the paste path - `copyPaste` used to call the sanitizer for every non-empty `text/html` payload, markup or not. A sanitizer that does more than strip tags (caps length, normalizes whitespace, writes an audit entry) now silently stops seeing plain-text pastes. Fine if that's on purpose, but it's a contract change that isn't in the docs or the migration guide, so it would be good to say so in the `sanitizer` JSDoc at least.

handsontable/src/plugins/dialog/ui.ts
```diff
@@ -284,9 +292,7 @@ export class DialogUI {
 
       // Render new dialog content
       if (typeof content === 'string') {
-        fastInnerHTML(contentElement, content,
-          this.#sanitizer ? (html: string, ctx: string) => this.#sanitizer!(html) ?? html : undefined,
-          'dialog', this.#overlayContainer);
+        fastInnerHTML(contentElement, content, this.#sanitizer, 'dialog', this.#warnScope);
```
budnix, line 295: The old code wrapped this with `?? html`, and the old field type allowed `undefined`. Now the value goes straight to `element.innerHTML`, and JS users aren't held to the declared `=> string` return type - a sanitizer that returns `undefined` for content it strips fully will put the literal word `undefined` in the dialog. I'd vote for `?? ''` over the old `?? html`, since falling back to the raw input undoes the sanitizing; the same gap exists for `sanitizeHTML()` callers, so one guard inside `fastInnerHTML`/`sanitizeHTML` would cover both.

handsontable/src/plugins/nestedHeaders/utils/ghostTable.ts
demtario: Good catch, and it is broader than nested headers. I checked every `fastInnerHTML` call site: after this PR none of them emits `'innerHTML'`. The password renderer used to reach that default too, so the value is gone from the library entirely, while `settings.types.ts:226` still pins `'innerHTML' | 'CopyPaste.paste'` as the type.

One correction to the impact, though. The ghost table was the only site that ever *called* a sanitizer with `'innerHTML'`. The password renderer passed no sanitizer argument at all (`fastInnerHTML(TD, value)`), so it hit the warn-only branch and no user function ever received that value from it. So the population is narrower than it first looks: a grid with `nestedHeaders`, a configured sanitizer, and an explicit `source === 'innerHTML'` branch. A catch-all `else` keeps working.

On the flag: I left `breaking: false`. Nothing fails to build, no content escapes sanitization, and the only thing that changes is which branch of a user's function runs. Happy to flip it if you would rather be loud, but it felt like the wrong signal for a change whose failure mode is "your redundant branch stops firing".

Migration guide added: `migrating-from-18.1-to-18.2`, section 1. It targets 18.2 rather than 18.1 because the 18.1 release branch is already frozen. Your read is exactly right that those users were getting two contexts for one label, so the guidance is to delete the `'innerHTML'` branch and keep the `'header'` one.

handsontable/src/plugins/copyPaste/copyPaste.ts
demtario: You are right on both counts, and it sent me to test something I should have tested before choosing one context.

I removed the sanitizing from that branch entirely, keeping only the `DOMParser` change in `parseTable.ts`, and re-ran the injection spec. The payload still does not execute. Only the sanitizer-called, warning, and degradation assertions failed. So the inert parse is what closes the hole; sanitizing that branch is policy, not safety.

That reframes your suggestion. A separate context is not a footgun, because a user who passes the internal payload through is not reopening an injection hole. Done in 44dacbb: it now arrives as `'CopyPaste.paste.sourceData'`, still sanitized by default so the branch that could previously be reached unchecked stays closed, but passing it through is a supported choice that restores object-based paste under an escaping sanitizer, and it halves the work on an internal paste. Your cost point is accurate: `#setClipboardData()` writes a full `_dataToHTML()` table to both types.

I used your name for it rather than the shorter `'CopyPaste.sourceData'` I first reached for, since it reads as "the sourceData part of the paste" and there was no good reason to deviate.

Documented in the security guide, the `sanitizer` JSDoc, and section 2 of the new 18.1 to 18.2 migration guide, including that passing it through is safe and why. Covered by two specs: the degradation under an escaping sanitizer, and the round-trip surviving when that one source is passed through.

The one thing I did not take is dropping the sanitizing altogether, which the experiment above would also have supported. Keeping it means a sanitizer that does more than filter markup still sees every clipboard payload, and it keeps a second layer if the inert parse is ever undone.

handsontable/src/utils/sanitizer.ts
demtario: This one is a straight bug on my side, not a deliberate choice. Fixed in 0ab8011.

You are right that it narrows the contract, and it did so at both call sites, not just paste: the ghost table also called `sanitizer(label, 'innerHTML')` unconditionally before this PR, so plain labels stopped reaching the sanitizer too. Your framing is the right one, a sanitizer is not always an XSS filter, and one that caps length or records an audit entry cannot do that for input it never receives.

The markup test now gates only the warning, which is the part that would otherwise fire on every grid rendering an ordinary header. The sanitizer is called whenever it is a function, exactly as both sites behaved before.

The unit spec that asserted the old short-circuit is inverted to assert the opposite, so it documents the corrected contract rather than the bug.

handsontable/src/plugins/dialog/ui.ts
demtario: You have talked me out of my own commit message on this one. I had recorded the `?? html` removal as intended, on the grounds that the declared return type is `string` and every other sink already behaved this way. That argument only holds for TypeScript users, and the sinks are reached from plain JavaScript just as often.

Fixed in 0ab8011, taking your suggestion on both points. `?? ''` rather than the old `?? html`, since falling back to the raw input undoes the sanitizing, which is the worse of the two failure modes. And the guard sits inside `fastInnerHTML` and `sanitizeHTML` rather than at the dialog call site, so every sink is covered rather than the one that happened to have it before.

`?? ''` also leaves a `TrustedHTML` return intact, which matters for anyone following the Trusted Types recipe in the docs, since `String()` coercion would have destroyed it.

handsontable/src/plugins/copyPaste/__tests__/settings/sanitizer.spec.js
```diff
@@ -36,6 +36,94 @@
       expect(sanitizer).toHaveBeenCalledWith('<div>test</div>', 'CopyPaste.paste');
     });
 
+    it('should be called for the private source-data clipboard type', async() => {
+      const sanitizer = jasmine.createSpy('sanitizer')
+        .and
+        .callFake(content => content.replace(/\s+onerror\s*=\s*("[^"]*"|'[^']*'|[^\s>]*)/gi, ''));
```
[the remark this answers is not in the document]
demtario: Fixed in 47c2fe0. The alert was correct: the stand-in sanitizer here stripped an `onerror="..."` pattern, and removing a multi-character sequence can reintroduce it. Both fixtures now escape `&`, `<` and `>` instead, which is the single-character rewrite the rule itself recommends. A test sanitizer that is unsound proves nothing about the code under test.

tests/fixtures/demo/sanitizer.html
```diff
@@ -0,0 +1,91 @@
...
+        // Deliberately excluded from sanitization: the `html` cell type renders the
+        // markup it is given. A test pins that, so making the sanitizer cover it
+        // later is a conscious behavior change rather than an accident.
+        { renderer: 'html' },
+        {},
+      ],
+      sanitizer: (content, source) => {
+        window.__sanitizerCalls.push([content, source]);
+
+        return content.replace(/\s+onerror\s*=\s*("[^"]*"|'[^']*'|[^\s>]*)/gi, '');
```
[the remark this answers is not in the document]
demtario: Fixed in 47c2fe0. The alert was correct: the stand-in sanitizer here stripped an `onerror="..."` pattern, and removing a multi-character sequence can reintroduce it. Both fixtures now escape `&`, `<` and `>` instead, which is the single-character rewrite the rule itself recommends. A test sanitizer that is unsound proves nothing about the code under test.

budnix (dismissed):

budnix (approved):

Lukortech:
Quick question (making sure I get the idea on the surface level for now) - when you're talking about `Known trade-off: object-based source data and escaping sanitizers` do I understand correctly it would be possible to be too eager about sanitization and then when copying the escaped characters into another table user could be left out with something like

`&lt;` turning into  `&amp;lt;` 
I see filtering also takes quite a hit with that effort and I can see why sanitization was rolled out wherever it could have been omitted.

demtario:
> Quick question (making sure I get the idea on the surface level for now) - when you're talking about `Known trade-off: object-based source data and escaping sanitizers` do I understand correctly it would be possible to be too eager about sanitization and then when copying the escaped characters into another table user could be left out with something like
>
> `&lt;` turning into `&amp;lt;`
>
> I see filtering also takes quite a hit with that effort and I can see why sanitization was rolled out wherever it could have been omitted.

Good question, and the mechanism you describe is real, but it is a different one from what that section is about.

**Double escaping.** You are right that it can happen in principle. Copy escapes once (`_dataToHTML` turns `&`, `<`, `>` into entities) and paste un-escapes once (`parseTable` maps `&amp;`, `&lt;`, `&gt;` back), so the round trip is normally clean. A sanitizer that escapes a second time in between leaves one layer behind after that single un-escape, and a cell that held `<b>` comes back holding `&lt;b&gt;`. That needs a sanitizer that keeps the table structure while escaping entities inside the cells, and it applies to the ordinary `text/html` clipboard branch, which has been sanitized since the `sanitizer` option shipped in 17.0. It is not new here and it is not what the trade-off note covers.

**What the note is actually about** is coarser and, I would argue, less nasty. A sanitizer that escapes everything does not just escape the cell contents, it escapes the `<table>` markup itself. `htmlToGridSettings()` then finds no table at all, so Handsontable falls back to the `text/plain` flavor of the clipboard. You do not get a mangled string; you get the value as displayed, and the object behind it is gone. So for a cell whose source value is `{ id: 9, value: 'B1' }` you end up with the string `B1`. That is the one covered by a test in the PR, and it is why the payload now has its own `'CopyPaste.paste.sourceData'` source you can pass through unchanged.

**On the performance table.** I do not think that is this change.

The header render path is where sanitization could plausibly cost something during a sort or filter, and it got no more expensive. On `develop` it already reads `this.hot.getSettings().sanitizer ?? true` once per header cell; this PR replaces that with `getSanitizer(this.hot)`, which is the same single property read moved into a helper. Nothing else in the diff runs during filtering or sorting: the clipboard and parser work happens only on paste, the ghost table only with nested headers, and the rest is menus, dialogs, notifications, and the password renderer.

The scenarios flagged as regressions are also the shortest ones in the suite, finishing in a small fraction of the time the scroll scenarios take, so a little variance there shows up as a large percentage. The long-running scenarios moved the other way in this run, which is not the shape a real slowdown takes.

docs/content/guides/upgrade-and-migration/migrating-from-18.1-to-18.2/migrating-from-18.1-to-18.2.md
```diff
@@ -0,0 +1,106 @@
...
+
+If you escape rather than strip, and you want object-valued paste to keep working, pass that one source through:
+
+```js
+sanitizer: (content, source) => {
+  if (source === 'CopyPaste.paste.sourceData') {
+    return content;
+  }
+
+  return escapeHtml(content);
```
Lukortech, line 100: If you'd wish to investigate possible drawbacks (performance) it could be tested first and short returned if there's no need for sanitization in case the value is simple string. 

Suggested by AI: 


```
function directReplace(str) {
  return str.replace(/[&<>"'`]/g, (char) => HTML_ESCAPES[char]);
}
```

```
function testFastPath(str) {
  if (!NEEDS_ESCAPE.test(str)) return str; // Returns exact reference, zero allocation
  return str.replace(/[&<>"'`]/g, (char) => HTML_ESCAPES[char]);
}
```

With output of comparisons: 


Scenario / String Type | Sample Input | Direct replace() (ops/sec) | Fast-Path test() (ops/sec) | Performance Delta
-- | -- | -- | -- | --
Short Plain Text (Most grid cells) | "John Doe" (8 chars) | ~4,200,000 | ~31,000,000 | ~7.4x faster
Medium Plain Text | "Product description without tags" (35 chars) | ~2,100,000 | ~24,000,000 | ~11.4x faster
Large Plain Text | 1,000 chars lorem ipsum (no HTML) | ~180,000 | ~4,800,000 | ~26.6x faster
String With HTML (Worst Case) | "<div>Hello & Welcome</div>" | ~1,800,000 | ~1,500,000 | ~0.8x (~20% penalty)

Lukortech (commented):
Regarding my [comment](https://github.com/handsontable/handsontable/pull/13236#issuecomment-5409542682)

Lukortech:
Alright, I'll leave the performance alone if the diff is so small between them that even a slight bump makes percentages grow. 
Thanks for in-depth explanation!

demtario:
> Alright, I'll leave the performance alone if the diff is so small between them that even a slight bump makes percentages grow. Thanks for in-depth explanation!

Following up on the performance table. Having looked at the harness itself, that mechanism is flawed and is not working as intended at the moment, so the regressions it flagged are not saying anything meaningful about this change. It is tracked as its own issue on our side now. Thanks again :D

demtario:
cursor review

Issue #7389, opened 2020-11-15
Data Loss Bug When Pasting Data Where the First Row Has Fewer Columns Than Later Rows


breck7:
### Description

(This may be related to the same code paths in: https://github.com/handsontable/handsontable/issues/7361)

When pasting data, if there are more columns in later rows than in the first row AND in the current view, than the extra columns will not get pasted.

Our scenario: author pastes some data(may be a bit messy) and only the first few columns are pasted.

A simple repro is below. 

### Steps to reproduce

Step 1 of 2: Load this:
```
const data = [
   ["name", "jill"],
  ]
const hotSettings = {
  data,
}
const hot = new Handsontable(document.querySelector('#hot'), hotSettings)
```

Step 2 of 2: Paste this:

```
schedule
	Football	Score
```

Only the first line will get pasted.

### Demo
https://jsfiddle.net/breck7/4fxndjcp/1/

### Your environment
* Handsontable version: Latest (8+)
* Browser Name and version: Chrome Latest
* Operating System: MacOSX Latest

AMBudnik:
Hi @breck7 

It seems that we get the same effect just by loading an empty object for the st=ettings (then we go with default 5 x 5 array of empty values) https://jsfiddle.net/AMBudnik/bdkmeqyj/ 


Handsontable
![Zrzut ekranu 2020-11-17 o 11 16 19](https://user-images.githubusercontent.com/10757813/99377301-5636ce00-28c6-11eb-8d4b-d9c09f4fbcb3.png)

GS
![Zrzut ekranu 2020-11-17 o 11 13 25](https://user-images.githubusercontent.com/10757813/99377330-5df67280-28c6-11eb-92b7-06bf33085cc2.png)

Below's the `beforePaste`/ `afterPaste` result (log1: data, log2: coordinates)

![Zrzut ekranu 2020-11-17 o 11 20 34](https://user-images.githubusercontent.com/10757813/99377799-f2f96b80-28c6-11eb-9d46-d24cc8714f0e.png)

Here's the `beforeChange` log

![Zrzut ekranu 2020-11-17 o 11 22 42](https://user-images.githubusercontent.com/10757813/99378057-4370c900-28c7-11eb-9e3e-779322731f18.png)

And the `modifyCopyableRange` hook
![Zrzut ekranu 2020-11-17 o 11 36 55](https://user-images.githubusercontent.com/10757813/99379666-4076d800-28c9-11eb-8a35-a01207fbee06.png)

[...]

Issue #7361, opened 2020-10-29
Data Loss Bug When Using minSpareCols


breck7:
### Description

If you are using minSpareCols when the table first loads, data in rows other than the first row may be lost if there are more columns in later rows than in the first row.

Our scenario is: author adds some data, may be a bit messy, clicks "Save". Returns to page later and on first load there changes are gone, and an autosave will overwrite their data.

A simple repro is below. 

### Steps to reproduce
```
const data = [
   ["first"],
   ["alpha", "beta", "cookie", "delta"],
  ]
const hotSettings = {
  data,
  minSpareCols: 1,
}
const hot = new Handsontable(document.querySelector('#hot'), hotSettings)
console.log(hot.getData()) // "delta" is not found
```

### Demo
https://jsfiddle.net/4fzk3qr8/

### Your environment
* Handsontable version: Latest (8+)
* Browser Name and version: Chrome Latest
* Operating System: MacOSX Latest

AMBudnik:
Thank you for sharing the issue report @breck7 I've replicated the issue.

Interestingly when I add anything after 'delta' those data won't appear. I need to extend `minSpareCols` to see the rest of the data. And when I use `hot.loadData(data)` then all data is loaded.

Version 8.0.0 https://jsfiddle.net/AMBudnik/ujcdpxg7/ works the same as 7.0.0 https://jsfiddle.net/AMBudnik/kt3eusx6/ so that is not a regression.

wojciechczerniak:
Problem here is not the missing column, but the fact we're showing too many of them. First row determines how many columns will we show (if `columns` or `dataSchema` are not defined), there should be only one + `minSpareCol` that is empty.

But I don't think this was ever planned. How can we add spare col when there is already data in those columns? I guess `minSpareCols` may work correctly only when it can add an empty column at the end.

breck7:
Haven't looked at the code yet, but seems on initial load `minSpareCol` may only look at the first row (O(1)). Maybe it should do a full scan instead (O(N))?

wojciechczerniak:
That would kill the performance.

breck7:
@wojciechczerniak would it? Should just be O(N) one time on load (for the avg case, 2N worst case). My guess is most of the time minSpareColumns isn't used alongside 10K+ spreadsheets, so impact might not be noticeable?

AMBudnik:
Hi @breck7 
[...]

PR #13237, opened 2026-08-25
DEV-2615: Fix paste dropping columns for a ragged clipboard (#7389)

budnix:
### Context

The PR fixes silent data loss when pasting a **ragged** clipboard — one whose first row holds fewer cells than a later row. The surplus columns were dropped with no error and no warning.

Reported in #7389, open since 2020-11-15 and confirmed still reproducible on `develop` @ `e7fa02fdff`.

The payload from the issue:

```
schedule
<TAB>Football<TAB>Score
```

Handsontable parses that to `[["schedule"], ["", "Football", "Score"]]`, then decides how many columns to write from the length of **row 0** alone, and gets `1`. `Football` and `Score` are never written anywhere. Google Sheets pads the short rows and pastes all three columns.

Two independent places made the same mistake:

1. **Plain-text paste** — `copyPaste.ts` read `plainData[0].length` into `populatedColumnsLength`, which bounds both the column loop and its modulo, so any cell past row 0's width was unreachable. `SheetClip.parse` does not pad, so the ragged array arrives intact.
2. **HTML paste** — `parseTable.ts` measured `countCols` from the first `<tr>`, so `htmlToGridSettings` allocated a grid that was already too narrow.

The second defect is the more damaging one and is not mentioned in the issue thread. It discards the cells **before** the `beforePaste` hook runs, so integrators cannot repair it — whereas defect 1 runs after the hook, which is why padding rows in `beforePaste` worked as a workaround for plain text but never for HTML. Real Excel and Google Sheets pastes carry a `text/html` table and Handsontable prefers that flavor, so the path with no workaround was the common one.

Both now take the width from the **widest** row. A short row is padded out and blanks the cells it covers — which is already what happened whenever a row other than the first was the short one, so the rule is now applied uniformly rather than changed.

Two things that make this bug easy to mis-reproduce, noted here for reviewers:

- The **reverse** shape (first row widest) was never broken — short rows read `undefined` and get blanks. A repro built that way shows green.
- Paste **does** grow the grid's column count for a rectangular array (a 3-wide paste into a 2-column grid yields 2x3), so "the grid is too narrow" is not the cause.

### How has this been tested?

Added coverage, and verified each new test fails without the fix and passes with it (the fix was reverted, and the pre-fix bundle swapped in, to confirm).

- **Unit** — 4 cases in `handsontable/src/utils/__tests__/parseTable.unit.ts` covering a narrow first row, a widest row in the middle, a `colspan` in a later row, and a guard that a first-row-widest table is unchanged. The first three fail without the fix.

  ```bash
  npm run test:unit --prefix handsontable -- --testPathPattern=parseTable
  ```

- **E2E (Playwright, real clipboard)** — 3 cases in `tests/e2e/clipboard.spec.ts` for ragged plain text, a ragged HTML table, and the first-row-widest guard. The HTML case deliberately carries a different `text/plain` flavor so the assertions can only pass if the HTML path was the one consumed. Two new `GridPage` helpers write the real clipboard.

  ```bash
  cd tests && npx playwright test e2e/clipboard.spec.ts --project=e2e-main
  ```

Full suites run green before the commit:

- `npm run test:unit --prefix handsontable` — 3736 passed, 312 suites
- `npm run test:e2e --prefix handsontable -- --testPathPattern=copyPaste` — 249 specs, 0 failures
- `npm run test:e2e --prefix handsontable -- --testPathPattern=parseTable` — 11 specs, 0 failures
- `cd tests && npx playwright test --project=e2e-main` — 400 passed, 1 pre-existing skip
- `npm run eslint --prefix handsontable` — 0 errors
- `npm run test:types --prefix handsontable` — passed

### Types of changes

### Related issue(s):

1. DEV-2615
2. #7389
3. Possibly related, reported alongside it: #7361

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2615

---

> [!NOTE]
> **Medium Risk**
> Changes core paste paths (plain and HTML) and what `beforePaste` receives for ragged input; behavior is intentional and heavily tested but integrators that assumed first-row width or ragged hook payloads may need to adjust.
> 
> **Overview**
> Fixes **silent column loss** when pasting clipboard rows of unequal length (e.g. a narrow first row followed by wider rows, as in #7389). Paste width is now driven by the **widest row**, matching spreadsheet behavior.
> 
> **Plain-text paste** (`copyPaste.ts`): column count and tiling use the max row length instead of the first row. Short rows are padded with **`null`** (not `undefined`, so object data sources stay serializable). **`padRowsToWidest`** runs before **`beforePaste`**, so hook data matches what the grid writes.
> 
> **HTML paste** (`parseTable.ts`): table width comes from **`countTableColumns`** across all rows (colspan/rowspan, last-`td` rules), not the first `<tr>`. Full-width footer colspans no longer inflate width; spans past the last column are **trimmed** when filling the grid.
> 
> Documentation adds a **“Rows of unequal length”** section on the clipboard guide. Coverage includes new unit tests, expanded `parseTable` cases, Playwright ragged paste scenarios, and **`GridPage`** clipboard write helpers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 11d256e5ec9d2b47ef1cd513ab67f86d9e4762c5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changelogs/7389.json
+++ b/.changelogs/7389.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed pasted columns being dropped when the clipboard's first row held fewer cells than a later row.",
+  "type": "fixed",
+  "issueOrPR": 7389,
+  "breaking": false,
+  "framework": "none"
+}
--- a/handsontable/src/plugins/copyPaste/__tests__/copyPaste.unit.ts
+++ b/handsontable/src/plugins/copyPaste/__tests__/copyPaste.unit.ts
@@ -26,15 +26,6 @@ describe('CopyPaste ragged clipboard', () => {
     return hot;
   }
 
-  it('should write every column of a ragged clipboard whose first row is the narrowest', () => {
-    const hot = pasteInto([['A1', 'B1', 'C1'], ['A2', 'B2', 'C2']], 'x\ny\tz\tw');
-
-    expect(hot.getDataAtCell(1, 1)).toBe('z');
-    expect(hot.getDataAtCell(1, 2)).toBe('w');
-
-    hot.destroy();
-  });
-
   it('should pad a short row with the empty-cell value rather than undefined', () => {
     const hot = pasteInto([['A1', 'B1', 'C1'], ['A2', 'B2', 'C2']], 'x\ny\tz\tw');
 
@@ -54,6 +45,54 @@ describe('CopyPaste ragged clipboard', () => {
     // `Object.keys` still lists a key whose value is `undefined` - only serializing drops it,
     // which is how the padded fields would silently vanish from a saved data source.
     expect(JSON.stringify(hot.getSourceData()[0])).toBe('{"a":"x","b":null,"c":null}');
+
+    hot.destroy();
+  });
+
+  it('should pad a short row that is not the first one', () => {
+    const hot = pasteInto(
+      [{ a: 'A1', b: 'B1', c: 'C1' }, { a: 'A2', b: 'B2', c: 'C2' }],
+      'x\ty\tz\nw'
+    );
+
+    // A trailing short row covers those cells too, so its fields must survive serializing.
+    expect(JSON.stringify(hot.getSourceData()[1])).toBe('{"a":"w","b":null,"c":null}');
+
+    hot.destroy();
+  });
+
+  it('should hand the beforePaste hook the same shape the grid is given', () => {
+    const hot = new Handsontable(document.createElement('div'), {
+      data: [['A1', 'B1', 'C1'], ['A2', 'B2', 'C2']],
+      licenseKey: 'non-commercial-and-evaluation',
+    });
+    let seen: unknown[][] = [];
+
+    hot.addHook('beforePaste', (data: unknown[][]) => {
+      seen = JSON.parse(JSON.stringify(data));
+    });
+
+    hot.selectCell(0, 0);
+    hot.getPlugin('copyPaste').paste('x\ny\tz\tw');
+
+    // A ragged array here would describe a different paste than the one the grid performed.
+    expect(seen).toEqual([['x', null, null], ['y', 'z', 'w']]);
+
+    hot.destroy();
+  });
+
+  it('should repeat a ragged clipboard across a wider selection on the widest row', () => {
+    const hot = new Handsontable(document.createElement('div'), {
+      data: [['A1', 'B1', 'C1', 'D1', 'E1', 'F1'], ['A2', 'B2', 'C2', 'D2', 'E2', 'F2']],
+      licenseKey: 'non-commercial-and-evaluation',
+    });
+
+    hot.selectCell(0, 0, 1, 5);
+    hot.getPlugin('copyPaste').paste('x\ny\tz\tw');
+
+    // The clipboard is three wide, so it tiles every three columns - not once per column.
+    expect(hot.getDataAtRow(0)).toEqual(['x', null, null, 'x', null, null]);
+    expect(hot.getDataAtRow(1)).toEqual(['y', 'z', 'w', 'y', 'z', 'w']);
 
     hot.destroy();
   });
--- a/handsontable/src/plugins/copyPaste/__tests__/copyPaste.unit.ts
+++ b/handsontable/src/plugins/copyPaste/__tests__/copyPaste.unit.ts
@@ -81,6 +81,29 @@ describe('CopyPaste ragged clipboard', () => {
     hot.destroy();
   });
 
+  it('should hand the hook the same width the grid writes when a cell spans past the table', () => {
+    const hot = new Handsontable(document.createElement('div'), {
+      data: [['A', 'B', 'C', 'D'], ['E', 'F', 'G', 'H']],
+      licenseKey: 'non-commercial-and-evaluation',
+    });
+    let seen: number[] = [];
+
+    hot.addHook('beforePaste', (data: unknown[][]) => {
+      seen = data.map((row: unknown[]) => row.length);
+    });
+
+    hot.selectCell(0, 0);
+    hot.getPlugin('copyPaste').paste('', '<table><tbody><tr><td>A1</td><td>B1</td><td>C1</td></tr>' +
+      '<tr><td colspan="20">footer</td></tr></tbody></table>');
+
+    // The footer spans far past the table. Reporting 20 columns while writing 3 would send an
+    // integrator syncing from the hook seventeen columns the grid never touched.
+    expect(seen).toEqual([3, 3]);
+    expect(hot.getDataAtRow(1)).toEqual(['footer', null, null, 'H']);
+
+    hot.destroy();
+  });
+
   it('should repeat a ragged clipboard across a wider selection on the widest row', () => {
     const hot = new Handsontable(document.createElement('div'), {
       data: [['A1', 'B1', 'C1', 'D1', 'E1', 'F1'], ['A2', 'B2', 'C2', 'D2', 'E2', 'F2']],
--- a/handsontable/src/plugins/copyPaste/copyPaste.ts
+++ b/handsontable/src/plugins/copyPaste/copyPaste.ts
@@ -612,7 +612,10 @@ export class CopyPaste extends BasePlugin {
 
     const selection = this.hot.getSelectedRangeActive();
     const populatedRowsLength = plainData.length;
-    const populatedColumnsLength = plainData[0].length;
+    // A ragged clipboard (rows of unequal length) must not be narrowed to the first row's width -
+    // cells past that width would never be written. The widest row wins, as in spreadsheet apps.
+    const populatedColumnsLength = plainData
+      .reduce((maxLength: number, row: unknown[]) => Math.max(maxLength, row.length), 0);
     const newRows = [];
 
     if (!selection) {
--- a/handsontable/src/plugins/copyPaste/copyPaste.ts
+++ b/handsontable/src/plugins/copyPaste/copyPaste.ts
@@ -52,6 +52,26 @@ const META_HEAD = [
   '<style type="text/css">td{white-space:normal}br{mso-data-placement:same-cell}</style>',
 ].join('');
 
+/**
+ * Squares off a ragged clipboard in place, filling the gaps with the empty-cell value.
+ *
+ * The grid pastes as wide as the widest row, so the rows that are shorter cover cells too. Doing
+ * this before the `beforePaste` hook keeps the hooks and the grid describing the same paste.
+ *
+ * @param {Array} data The parsed clipboard data.
+ */
+function padRowsToWidest(data: unknown[][]) {
+  const width = data.reduce((widest: number, row: unknown[]) => Math.max(widest, row.length), 0);
+
+  data.forEach((row: unknown[]) => {
+    for (let column = 0; column < width; column += 1) {
+      if (row[column] === undefined) {
+        row[column] = null;
+      }
+    }
+  });
+}
+
 /* eslint-disable jsdoc/require-description-complete-sentence */
 
 /**
@@ -894,6 +914,8 @@ export class CopyPaste extends BasePlugin {
     if (pastedData === void 0 || Array.isArray(pastedData) && pastedData.length === 0) {
       return;
     }
+
+    padRowsToWidest(pastedData as unknown[][]);
 
     // Store a copy of the original pasted data before user can modify it in beforePaste hook.
     // This is needed to detect if user modified values and respect their modifications over source data.
--- a/handsontable/src/plugins/copyPaste/copyPaste.ts
+++ b/handsontable/src/plugins/copyPaste/copyPaste.ts
@@ -683,7 +683,9 @@ export class CopyPaste extends BasePlugin {
           cellValue = parsedCellValue;
         }
 
-        newRow.push(cellValue);
+        // A row shorter than the widest one has no value here. Write the empty-cell value rather
+        // than `undefined`, which would delete the property outright in an object data source.
+        newRow.push(cellValue === undefined ? null : cellValue);
       }
 
       newRows.push(newRow);
--- a/handsontable/src/utils/parseTable.ts
+++ b/handsontable/src/utils/parseTable.ts
@@ -261,6 +261,44 @@ export function replaceTdCellsWithTextContent(html: string): string {
 }
 
 /**
+ * Measures how many data columns a pasted table needs.
+ *
+ * A cell takes `colSpan` slots in its own row and reserves the same slots in every row its
+ * `rowSpan` reaches; a row-header cell takes none. Measuring one row instead under-sizes a
+ * ragged table, and the cells that no longer fit are then dropped without warning.
+ *
+ * @param {HTMLTableElement} table The table to measure.
+ * @param {boolean} hasRowHeaders Whether the table's leading `<th>` cells are row headers.
+ * @returns {number} The number of data columns.
+ */
+function countTableColumns(table: HTMLTableElement, hasRowHeaders: boolean): number {
+  const rows = table.rows;
+  const reserved: number[] = [];
+  let maxCols = 0;
+
+  for (let row = 0; row < rows.length; row += 1) {
+    const cells = rows[row].cells;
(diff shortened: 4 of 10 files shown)


qunabu:
Task linked: [DEV-2615 Fix paste dropping columns when the clipboard's first row is the narrowest (#7389)](https://app.clickup.com/t/9015210959/DEV-2615)

handsontable/src/utils/parseTable.ts
```diff
@@ -260,6 +260,44 @@ export function replaceTdCellsWithTextContent(html: string): string {
...
+
+  for (let row = 0; row < rows.length; row += 1) {
+    const cells = rows[row].cells;
+    let cols = reserved[row] || 0;
+
+    for (let cell = 0; cell < cells.length; cell += 1) {
+      const { colSpan, rowSpan, nodeName } = cells[cell];
+
+      if (!(hasRowHeaders && cell === 0 && nodeName === 'TH')) {
+        cols += colSpan;
```
Haxikowy: `cols += colSpan` lets a row whose width comes only from a spanning cell set the width for the whole table.

```js
htmlToGridSettings('<table><tbody><tr><td>A1</td><td>B1</td><td>C1</td></tr><tr><td colspan="20">footer</td></tr></tbody></table>').data[0].length
// branch: 20   develop: 3
```

`colspan="1000"` (the DOM ceiling) gives a 1000-column grid and a rows x 1000 array. A table with a full-width footer row, normal in email and Word exports, now adds 17 blank columns and blanks 17 cells in every pasted row. Develop pasted that same clipboard cleanly, so this is a new instance of the silent overwrite the PR is fixing.

The ragged-rows argument in the description does not cover it: no row in that table holds 20 discrete cells. A colspan-only row shouldn't widen the count past the widest row of discrete cells. Needs a test for colspan overflow either way.

handsontable/src/utils/parseTable.ts
```diff
@@ -260,6 +260,44 @@ export function replaceTdCellsWithTextContent(html: string): string {
...
+
+        for (let spanned = row + 1; spanned < row + rowSpan; spanned += 1) {
```
Haxikowy: Unbounded by the table's own row count, so the loop writes `reserved` entries nothing reads. Measured in Chrome: a 2-row table with 1000 `rowspan="65534"` cells takes 151.57 ms, clamped 0.13 ms, same result. 10000x20 with `rowspan="10000"` on every cell: 4579 ms, clamped 779 ms. Span-free 10000x20 is unchanged (12.66 vs 12.42 ms), so the clamp is free.

```suggestion
        const spannedEnd = Math.min(row + rowSpan, rows.length);

        for (let spanned = row + 1; spanned < spannedEnd; spanned += 1) {
```

handsontable/src/utils/parseTable.ts
```diff
@@ -260,6 +260,44 @@ export function replaceTdCellsWithTextContent(html: string): string {
...
+  let maxCols = 0;
+
+  for (let row = 0; row < rows.length; row += 1) {
+    const cells = rows[row].cells;
+    let cols = reserved[row] || 0;
+
+    for (let cell = 0; cell < cells.length; cell += 1) {
+      const { colSpan, rowSpan, nodeName } = cells[cell];
+
+      if (!(hasRowHeaders && cell === 0 && nodeName === 'TH')) {
```
Haxikowy: The exemption skips a `th` only at index 0, but the fill loop sends every non-`TD` cell to `rowHeaders` and gives it no slot. A `th` anywhere else is counted and leaves a phantom trailing column: `<tr><td>a</td><th>grp</th><td>b</td></tr><tr><td>1</td><td>2</td></tr>` parses as `[['a','b',undefined],['1','2',undefined]]`. Develop had the same defect but only for row 0. `nodeName !== 'TD'` would match the fill loop.

handsontable/src/utils/parseTable.ts
```diff
@@ -292,9 +330,7 @@ export function htmlToGridSettings(element: HTMLTableElement | string, rootDocum
   const el: HTMLTableElement = checkElement as HTMLTableElement;
   const generator = tempElem.querySelector('meta[name$="enerator"]') as HTMLMetaElement | null;
   const hasRowHeaders = el.querySelector('tbody th') !== null;
-  const trElement = el.querySelector('tr') as HTMLTableRowElement | null;
-  const countCols = !trElement ? 0 : (Array.from(trElement.cells)
-    .reduce((cols: number, cell: HTMLTableCellElement) => cols + cell.colSpan, 0)) - (hasRowHeaders ? 1 : 0);
+  const countCols = countTableColumns(el, hasRowHeaders);
```
Haxikowy: `countCols` now comes from every row while `colHeaders` and `nestedHeaders` still come from `thead`, so the two can disagree:

```js
htmlToGridSettings('<table><thead><tr><th>h1</th><th>h2</th></tr></thead><tbody><tr><td>a</td></tr><tr><td>a</td><td>b</td><td>c</td></tr></tbody></table>')
// branch: data 3 wide, colHeaders ['h1','h2']   develop: data 2 wide, colHeaders ['h1','h2']
```

They matched by construction before. `htmlToGridSettings` is public (`Handsontable.helper`, type-tested in `test/types`), so a scraped ragged table fed to `new Handsontable` now gets auto-named surplus columns. Padding `colHeaders` and each `nestedHeaders` row out to `countCols` once it is known would keep them in step.

handsontable/src/plugins/copyPaste/copyPaste.ts
```diff
@@ -680,7 +683,9 @@ export class CopyPaste extends BasePlugin {
           cellValue = parsedCellValue;
         }
 
-        newRow.push(cellValue);
+        // A row shorter than the widest one has no value here. Write the empty-cell value rather
+        // than `undefined`, which would delete the property outright in an object data source.
+        newRow.push(cellValue === undefined ? null : cellValue);
```
Haxikowy, line 708: `null` is the right value here, it matches what `emptySelectedCells` writes. Two things ride along that the description doesn't mention.

The coercion also fires for a short *later* row, the shape the description calls never broken. An object data source that used to lose those keys now keeps them: develop `{"a":"w"}`, branch `{"a":"w","b":null,"c":null}`, and `getDataAtCell` goes from `undefined` to `null`. Anyone testing `value === undefined` to mean "nothing pasted" sees a change.

The coercion also runs after `beforePaste`, so hooks still see the ragged array on the plain-text path and `undefined` holes on the HTML path while the grid gets `null`s. An integrator syncing from `afterPaste` pushes the ragged array and misses the cells the grid just blanked, which is this PR's own bug one layer out. Padding `pastedData` to the widest row before `runHooks('beforePaste')` fixes both flavors and lets this coercion go away.

handsontable/src/plugins/copyPaste/copyPaste.ts
```diff
@@ -612,7 +612,10 @@ export class CopyPaste extends BasePlugin {
 
     const selection = this.hot.getSelectedRangeActive();
     const populatedRowsLength = plainData.length;
-    const populatedColumnsLength = plainData[0].length;
+    // A ragged clipboard (rows of unequal length) must not be narrowed to the first row's width -
+    // cells past that width would never be written. The widest row wins, as in spreadsheet apps.
+    const populatedColumnsLength = plainData
```
Haxikowy, line 637: This also changes the repeat-fill period. Pasting `x\ny\tz\tw` into a 6-wide selection: develop filled row 0 with `x,x,x,x,x,x`, the branch gives `x,null,null,x,null,null`. That's the spreadsheet behavior and a real win, but nothing tests paste-into-a-larger-selection with a ragged clipboard, and the description doesn't mention it.

.changelogs/7389.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed pasted columns being dropped when the clipboard's first row held fewer cells than a later row.",
```
Haxikowy, line 3: The title names only the dropped columns. Three user-visible changes aren't in it: short rows now blank the cells they cover, padded object fields are written as `null` instead of being deleted, and repeat-fill into a larger selection tiles on the widest row. `.ai/BREAKING-CHANGES.md` counts a behavior change as declarable, so either extend the title or add a second `changed` entry.

tests/e2e/clipboard.spec.ts
```diff
@@ -52,4 +52,56 @@ test.describe('clipboard', () => {
...
+      // The reverse shape was never broken — this guards it against the fix.
+      await grid.selectCell(0, 0);
+      await grid.writeClipboardText('x\ty\tz\nw');
+      await page.keyboard.press('ControlOrMeta+v');
+
+      await grid.expectCell(0, 0, 'x');
+      await grid.expectCell(0, 1, 'y');
+      await grid.expectCell(0, 2, 'z');
+      await grid.expectCell(1, 0, 'w');
+      await grid.expectCell(1, 1, '');
```
Haxikowy: `expectCell` renders `''` for both `undefined` (develop) and `null` (this branch), so this guard can't detect the change it's guarding. A `getSourceData()` assertion for the reverse shape in `raggedPaste.unit.ts` is red on develop and would cover it.

handsontable/src/plugins/copyPaste/__tests__/copyPaste.unit.ts
```diff
@@ -0,0 +1,60 @@
+import Handsontable from '../../../index';
+import { registerCellType, TextCellType } from '../../../cellTypes';
+import { registerPlugin } from '../../registry';
+import { CopyPaste } from '../copyPaste';
+
+registerCellType(TextCellType);
+registerPlugin(CopyPaste);
+
+describe('CopyPaste ragged clipboard', () => {
```
Haxikowy, line 9: The filename matches no source module; the convention in this directory is module-named (`copyableRanges.unit.ts` for `copyableRanges.ts`). Case 1 also duplicates the plain-text E2E at roughly a second of jsdom. Cases 2 and 3 carry the value E2E can't reach.

Haxikowy (commented):
#7361, cited in the description as possibly related, is not fixed and still reproduces here: `data: [['first'], ['alpha','beta','cookie','delta']]` with `minSpareCols: 1` loses `delta`, because the load path derives its column count from row 0 the same way. It is already closed as completed. Worth either dropping from the description or saying plainly that it isn't covered.

The docs checkbox is unchecked and #7389 carries the `Docs: Content` label. Nothing in `docs/content/guides/cell-features/clipboard/clipboard.md` states the old first-row-wins rule, so no text contradicts the code, but the new rule (widest row wins, short rows blank the cells they cover) is undocumented, as is `pasteMode: 'shift_down' | 'shift_right'` now shifting by the widest width.

handsontable/src/utils/parseTable.ts
budnix: Nice catch — that footer case really did blow the width up to 20. Fixed in `84a613252b`, a row's last cell now takes just the one slot it lands in, with a test for it.

handsontable/src/utils/parseTable.ts
budnix: Good spot — applied your `spannedEnd` clamp in `84a613252b`. Thanks for measuring it.

handsontable/src/utils/parseTable.ts
budnix: You're right, matching the fill loop is the correct rule — switched to counting only `td` in `84a613252b`. It also cleared up the phantom column a `thead` colspan was adding.

handsontable/src/utils/parseTable.ts
budnix: Fair point, but they lined up before only because the extra column was being dropped. Padding with `undefined` changes nothing since `getColHeader` falls back to the auto label anyway — and the `td`-only fix above removed the `thead`-wider case.

handsontable/src/plugins/copyPaste/copyPaste.ts
budnix: Good point — the hooks describing a different paste than the grid was the real problem here. Padded before `beforePaste` in `84a613252b`; I kept the coercion as a guard, since a hook can still hand back a ragged array.

handsontable/src/plugins/copyPaste/copyPaste.ts
budnix: Agreed, that was untested — added a repeat-fill case in `84a613252b` and called the change out in the changelog.

.changelogs/7389.json
budnix: Yep, my mistake — the title only covered half of it. Added a second `changed` entry in `84a613252b` for the blanking, the `null` fields and the repeat-fill width.

tests/e2e/clipboard.spec.ts
budnix: Good catch, that assertion proved nothing either way. Replaced it in `84a613252b` with a `getSourceData()` case for the reverse shape.

handsontable/src/plugins/copyPaste/__tests__/copyPaste.unit.ts
budnix: Agreed — renamed it to `copyPaste.unit.ts` in `84a613252b` and dropped the case the E2E already covers.

handsontable/src/utils/parseTable.ts
Haxikowy: The footer case is fixed, but `cell === lastCell ? 1 : colSpan` under-reserves a cell that spans both ways. On `84a613252`, pasting this into a 4-column grid:

```html
<table><tbody><tr><td rowspan="2" colspan="2">A</td></tr><tr><td>B</td><td>C</td></tr></tbody></table>
```

gives `[["A",null,null,"d"],[null,null,"B","h"]]`. `C` is gone. `countCols` says 3, the fill loop needs 4, so `dataArr[row].findIndex(v => v === undefined)` returns -1 and the value goes to index -1. `dfc2d6a0c` pasted the same clipboard as 4 columns with both cells.

The rule also leaves the output ragged: `<tr><td>a</td></tr><tr><td>b</td><td colspan="3">wide</td></tr>` now returns row lengths `[2,4]`, with `mergeCells` colspan 3 pointing past `countCols`.

Counting the full `colSpan` and clamping the fill loop's merge write and the `mergeCells` colspan to `countCols` handles the footer row without the -1 write and keeps the rows rectangular. Either way, worth a case asserting nothing is dropped when the first row is a rowspan plus colspan cell.

handsontable/src/plugins/copyPaste/copyPaste.ts
Haxikowy: Padding before the hook is the right place, but it squares off to the widest *physical* row, and the parser has already expanded the merged footer row to 20 entries. The grid writes 3 columns while the hook sees 20:

```
// 84a613252, paste of <tr><td>A1<td>B1<td>C1</tr><tr><td colspan="20">footer</tr>
parsed row lengths      [3, 20]
beforePaste row lengths [20, 20]
grid                    [["A1","B1","C1"], ["footer", null, null]]
```

So `afterPaste` reports 17 `null` columns the grid never wrote, which is the same mismatch with the sign flipped: an integrator syncing from the hook now writes phantom columns instead of missing real ones. Clamping the parser's rows to `countCols`, or padding once the paste width is settled, lines the two back up.

handsontable/src/utils/parseTable.ts
budnix: You're right, the reservation had to stay the full span even where the row itself could spare it — fixed in `11d256e5ec`, with the rowspan-plus-colspan case tested. The ragged rows are gone too, by trimming the merge write rather than counting the full span, which would have brought the footer bug back.

handsontable/src/plugins/copyPaste/copyPaste.ts
budnix: Good spot — the padding was squaring off rows the parser had already stretched. Trimming the merge write in `11d256e5ec` fixes it at the source, so the hook now reports 3 columns, same as the grid.

Haxikowy (approved):

PR #13238, opened 2026-08-25
DEV-2621: Bootstrap linked worktrees for agent sessions

budnix:
### Context

The PR makes a linked git worktree behave like the main checkout, so agent sessions working under `.claude/worktrees/` stop losing context and tooling.

A `git worktree` materializes **tracked files only**. Everything gitignored or stored outside the repo is absent at birth, which produces two failures that look unrelated but share that one cause. Both were measured on `develop` @ `e7fa02fdff`, 2026-08-25.

**1. Project memory does not follow the worktree.** Claude Code keys its project state by the checkout path, so every worktree resolves to its own `~/.claude/projects/<slug>` directory with no `memory/` in it. The accumulated project facts silently do not load — 33 project directories on one machine, 24 of them per-worktree, all starting empty. Nothing errors; the agent simply knows less and reaches for the repo's skills less often. Skills and agents are tracked, so those were never the problem (a diff of one worktree's `.claude/skills/` against `develop` showed zero drift).

The slug replaces **every non-alphanumeric character** with a dash. Verified from a recorded session: the checkout `…/worktrees/feature+DEV-1656_Autocomplete-dropdown-flex-layout` is stored as `…--claude-worktrees-feature-DEV-1656-Autocomplete-dropdown-flex-layout` — the `.`, `+` and `_` all became dashes. That underscore matters, because this repository names branches `feature/DEV-xxxx_Name`, so most worktree directories contain one.

**2. The symlinked `node_modules` breaks every package script.** The worktree's root `node_modules` is created as a symlink to the main checkout, with no per-package trees. This half-works, which is why it misleads: root binaries resolve, so `npx eslint --version` prints `v8.57.1` and lint passes. Then:

```
$ npm --prefix handsontable run test:unit
/bin/sh: env-cmd: command not found
Step failed (exit 127)
```

pnpm keeps each package's `.bin` shims inside that package's own `node_modules`, and none of those directories exist. The legacy Jasmine suite fails the same way with `jasmine is not defined`, which reads like a test bug rather than a setup problem. Across nine existing worktrees, eight had been hand-fixed with a real install and one still ran on the symlink — there was no script, so every worktree got repaired differently or not at all.

Two smaller gaps are closed here as well. A symlinked `node_modules` **escapes `.gitignore`**, because the rule was `node_modules/` and a trailing slash matches directories only — `git add -A` would have committed a link holding an absolute machine-local path. And `docs/.env` / `docs/tests/.env` never reach a worktree, so docs work and the docs test suite fail there for a reason nothing explained.

### The two mechanisms, and why both are needed

`.worktreeinclude` (new here) is Claude Code's own mechanism for carrying gitignored **files** into worktrees it creates — `--worktree`, subagent isolation, and the desktop app. It uses `.gitignore` syntax and copies only files that match *and* are gitignored.

It cannot install dependencies, and it cannot reach anything outside the repository, which is where the costliest gap lives. So `scripts/claude/setup-worktree.mjs` handles the install and the project-memory symlink, and `.worktreeinclude` handles the files. The file also records, in comments, what is deliberately **not** copied and why — `node_modules` (a copy is not an install), `handsontable/dist|tmp|styles` (a stale bundle is the "wrong build" trap), and `.code-review-graph/` (branch-stamped, so a copy answers with another branch's structure).

### Not affecting developers who do not use worktrees

- The bootstrap exits immediately when `.git` is a directory rather than a worktree pointer, and a unit test pins that. Its `--check` mode prints nothing in a normal clone.
- The SessionStart hook redirects stderr and ends in `|| true`, so a checkout without the script sees nothing and never fails a session.
- Every default is unchanged — ports stay `8123` and `8082`, and no npm script, build, or test command was modified.
- The port knob is namespaced `HOT_TEST_PORT`, not the ambient `PORT`, which shells and this repo's own example servers already export for unrelated reasons.
- `.worktreeinclude` only ever copies files that are already gitignored, so nothing tracked is duplicated.

### Corrections made during review

Three claims in the first revision of this PR were wrong and have been fixed. Recording them because each was silent:

1. **The project-slug encoding was backwards.** The first version preferred a spelling that preserved underscores, which would have put the memory symlink in a directory Claude Code never reads while `--check` reported the worktree ready forever. A test had been written asserting the broken order; it now pins the verified encoding, including the recorded slug verbatim.
2. **The SessionStart hook was silent in worktrees.** It referenced `${CLAUDE_PROJECT_DIR}`, which [the worktrees documentation](https://code.claude.com/docs/en/worktrees) states does not follow a session into a worktree, while the `cwd` field of the hook's stdin payload does. The hook therefore ran the main checkout's copy of the script and never warned. It now reads that payload, verified for a worktree `cwd` (warns), a main-checkout `cwd` (silent), and no stdin (falls back to the script's own location).
3. **Copying `.claude/settings.local.json` was unnecessary.** Since Claude Code v2.1.211 a permission approval made in a worktree is saved to the **main checkout's** file and applies in every worktree of the repository, so the copy was redundant and would only drift. The step was removed and the documentation corrected.

Also fixed: `GIT_DIR`/`GIT_WORK_TREE` are now stripped from the `pnpm` child, which `.ai/LOCAL-ENFORCEMENT.md` requires for any child spawned with an explicit `cwd` and which the root `prepare` script's `lefthook install` would otherwise misdirect; `HOT_TEST_PORT` throws on a malformed value instead of silently falling back to the shared port; the readiness marker carries a version so a worktree bootstrapped by an older script stops reporting ready; the memory link is checked for its **target**, not merely that it resolves; `tests/node_modules/hyperformula` is probed, because a partial install turns the formulas fixtures into bare 404s mid-run; and a failed install now names the recovery command.

### Known gaps, documented rather than hidden

`.ai/WORKTREES.md` carries a "Known gaps" section for two things this PR does not fix:

- **The port collision is avoidable, not prevented.** `HOT_TEST_PORT` only helps someone who knows to set it; on defaults two checkouts still share one server. Closing it properly means teaching `tests/support/static-server.mjs` to report the root it serves and having the config refuse a server belonging to another checkout.
- **The agent-time Claude hooks fail open in a worktree.** `post-tool-use.mjs` and `stop.mjs` resolve paths against the checkout the hook script lives in, so the spec autolint and the Stop new-Jasmine-spec gate never match and never complain. They need the same `cwd` fix applied here. The git hooks (pre-commit, pre-push) are unaffected and still run.

The visual-tests port is also **not** parameterizable from its Playwright config: it is owned by `EXAMPLES_SERVER_PORT` in `visual-tests/src/config.mjs`, and hardcoded again as `app.listen(8082)` in `examples/next/visual-tests/react-wrapper/demo/server.js` and the Angular equivalent, neither of which reads `--port`. Changing only the config points the browser at a port nothing is serving and fails every visual spec. An earlier revision of this PR did exactly that; it was reverted, and the documentation now says so explicitly rather than leaving a trap.

### How has this been tested?

- New tooling tests, picked up by the existing `test:tooling` glob. They pin the inert-in-a-normal-clone guarantee, the verified slug encoding (including the recorded pair verbatim, so a future change cannot pass by agreeing with itself), and the hook payload's malformed-input handling:

```bash
npm run test:tooling
# 105 tests, 105 pass, 0 fail (7 new)
```

- Lint on every changed JS/TS file:

```bash
npx eslint scripts/claude/setup-worktree.mjs \
  scripts/claude/__tests__/setup-worktree.test.mjs \
  tests/playwright.config.ts
```

- Manual, in a real worktree created for this work:
  - The exit-127 failure above was reproduced **first**, to confirm the diagnosis rather than assume it.
  - `--check` correctly reported every defect of an unbootstrapped worktree and exited 1.
  - The memory link was exercised for real, not just `--dry-run`: the worktree's project directory then contained `memory -> …/handsontable-develop/memory`, and the full `MEMORY.md` index read through it. `--check` narrowed accordingly, and the remaining items were true.
  - `--skip-install` leaves an existing `node_modules` untouched (verified before and after).
  - A deliberately dangling memory symlink is detected and replaced rather than reported as present.
  - The hook's root resolution was verified for all three inputs described above.
- `.claude/settings.json` validated as JSON, and the hook confirmed silent when the script is absent.

### Types of changes

### Related issue(s):

1. DEV-2621

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

Tooling and documentation only — no package source is touched. [skip changelog]

### Checklist:

- [x] My change requires a change to the documentation. Added `.ai/WORKTREES.md` and `.worktreeinclude`, registered the guide in `.ai/README.md`, and added a navigation row plus one "Monorepo gotchas" bullet in `AGENTS.md`.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2621


qunabu:
Task linked: [DEV-2621 Bootstrap linked worktrees so agent sessions keep memory, deps and permissions](https://app.clickup.com/t/9015210959/DEV-2621)

scripts/claude/setup-worktree.mjs
```diff
@@ -0,0 +1,422 @@
...
+ *
+ * @param {string} checkout The checkout path.
+ * @returns {string[]} Candidate absolute paths, preferred first.
+ */
+function projectStateCandidates(checkout) {
+  const base = path.join(homedir(), '.claude', 'projects');
+
+  return [
+    path.join(base, checkout.replace(/[/\\.]/g, '-')),
+    path.join(base, checkout.replace(/[^a-zA-Z0-9]/g, '-')),
```
budnix: **The candidate order is backwards - Claude Code replaces underscores with dashes.**

Measured on this machine: the worktree `.claude/worktrees/feature/DEV-1656_Autocomplete-dropdown-flex-layout` has the state directory `...--claude-worktrees-feature-DEV-1656-Autocomplete-dropdown-flex-layout` (underscore -> dash). Across all 34 directories in `~/.claude/projects/`, **not one** contains an underscore, while this repo's branch convention (`feature/DEV-xxxx_Name`) puts one in most worktree names - `DEV-1023_NestedRows-Public-API` is on disk right now.

So `candidates[1]` is the real reading and `candidates[0]` is the guess.

Why it matters: `projectStateDir()` falls back to `candidates[0]` when **neither** exists - which is exactly the documented flow ("Run it once per worktree, before you build or test", i.e. before the first session ever opens there). The memory symlink then lands at the underscore spelling, Claude Code later creates the dash spelling, and the highest-value step has silently done nothing.

It is also self-hiding: once both directories exist, `--check` resolves `candidates[0]` first, finds `memory` there, and reports the worktree ready **forever** while every session runs with no memory.

Flipping is safe - the two readings coincide whenever the path has no underscore or colon.

.claude/settings.json
```diff
@@ -42,6 +42,11 @@
             "type": "command",
             "command": "command -v pipx >/dev/null 2>&1 && pipx run code-review-graph==2.3.6 status || true; echo 'Cross-file queries (who calls X, what imports Y, rename impact, blast radius): query the code-review-graph MCP before walking call chains with Grep+Read. Load schemas first via ToolSearch select:mcp__code-review-graph__query_graph_tool — full workflow in the code-graph skill.'",
             "timeout": 30
+          },
+          {
+            "type": "command",
+            "command": "node \"$CLAUDE_PROJECT_DIR/scripts/claude/setup-worktree.mjs\" --check 2>/dev/null || true",
```
budnix: **This hook likely never fires in the flow it was written for.**

Claude Code's worktrees documentation ("Hook paths don't follow the worktree") states that `${CLAUDE_PROJECT_DIR}` stays at the **original project root** and does not follow the session into a worktree; hooks that need the worktree must read the `cwd` field from the hook's stdin JSON.

If that holds, then in an `EnterWorktree` session this runs the **main checkout's** copy of the script. `repoRoot()` resolves from `import.meta.url`, so `ROOT` is the main checkout, `mainCheckout()` returns `null`, and the script prints nothing - by design. The readiness warning is silent in exactly the case it exists for.

Two more teeth:
- Before this PR merges to `develop`, `$CLAUDE_PROJECT_DIR/scripts/claude/setup-worktree.mjs` does not exist at all on the main checkout, and `2>/dev/null || true` swallows the `MODULE_NOT_FOUND`.
- I could not reproduce the variable locally (it is only injected into hook processes), so please verify - but the fix below is strictly more robust under **either** reading.

Suggested fix: have the script resolve the checkout from the hook payload's `cwd` (the stdin-JSON pattern `post-tool-use.mjs` already uses) and fall back to `import.meta.url` when there is no payload. That makes it correct whether or not `$CLAUDE_PROJECT_DIR` follows the worktree.

scripts/claude/setup-worktree.mjs
```diff
@@ -0,0 +1,422 @@
...
+
+  if (dryRun) {
+    log('Would run: pnpm install --frozen-lockfile');
+
+    return;
+  }
+
+  log('Installing dependencies (pnpm install --frozen-lockfile)…');
+
+  const install = spawnSync('pnpm', ['install', '--frozen-lockfile'], {
```
budnix: **`GIT_DIR`/`GIT_WORK_TREE` are not stripped from this child, which `.ai/LOCAL-ENFORCEMENT.md` forbids.**

That file (the authority root `AGENTS.md` points to for enforcement hooks) says, verbatim:

> Strip `GIT_DIR`/`GIT_WORK_TREE` from the environment of any child you spawn with an explicit `cwd`.

This spawn passes an explicit `cwd: ROOT` and inherits the environment untouched. The install runs the root `prepare` script, which runs `lefthook install` - a git-writing step. Run the bootstrap from any context that exports `GIT_DIR` (a git hook, or a shell where a previous hook leaked it) and that install targets the wrong git directory.

The same rule is the reason `repoRoot()` exists, so the omission here reads as an oversight rather than a decision.

scripts/claude/setup-worktree.mjs
```diff
@@ -0,0 +1,422 @@
...
+
+  const nodeModules = path.join(ROOT, 'node_modules');
+  const link = await lstat(nodeModules).catch(() => null);
+
+  if (link?.isSymbolicLink()) {
+    log(`Removing node_modules symlink -> ${await readlink(nodeModules)}`);
+    log('  A symlinked node_modules leaves every package-local .bin missing.');
+
+    if (!dryRun) {
+      await rm(nodeModules, { force: true });
```
budnix: **A failed install leaves the worktree with zero dependencies - the exact outcome `--skip-install` was written to avoid.**

The `--skip-install` branch above reasons about this explicitly:

> Removing the link and then not installing would leave the worktree with no dependencies at all - strictly worse than it started

The same hazard is unguarded here. The symlink is deleted, then `pnpm install --frozen-lockfile` runs. It fails whenever the branch's `pnpm-lock.yaml` disagrees with `package.json`, when the registry is unreachable (this repo explicitly supports air-gapped machines), or when `pnpm` is not on `PATH` - and the worktree is left with **no** `node_modules` at all, worse than before it was touched.

Removal genuinely must precede the install (otherwise pnpm writes through the link into the main checkout), so the fix is the message rather than the order: in the `install.status !== 0` block below, say that `node_modules` is now absent and name the command that recovers it.

tests/playwright.config.ts
```diff
@@ -1,6 +1,17 @@
...
+// `reuseExistingServer` attaches to whatever already listens on this port. Two
+// checkouts (a worktree and the main clone) on the default port therefore share
+// ONE server, and the second one silently tests the first one's build. Set
+// `HOT_TEST_PORT` to run them side by side.
+//
+// Namespaced on purpose. `PORT` is a conventional name that shells and tools
+// already export for unrelated reasons, and this config would otherwise retarget
+// itself for anyone who happens to have it set. The port is passed explicitly to
+// the server below, so the two always agree whatever the ambient environment says.
+const PORT = Number(process.env.HOT_TEST_PORT) || 8123;
```
budnix: **A typo in `HOT_TEST_PORT` silently restores the collision this variable exists to prevent.**

`Number('8124x')` is `NaN` and `Number('')` is `0` - both falsy, so `|| 8123` sends the run back to the shared default. Combined with `reuseExistingServer`, worktree B then attaches to worktree A's server and reports results describing A's build, while the developer believes the port was moved. Silent wrong answers are the whole failure mode this PR is fixing.

`HOT_TEST_PORT=0` is also swallowed, and `0` is the one value that would be genuinely useful here ("any free port").

Fail loudly instead:

```suggestion
const PORT = process.env.HOT_TEST_PORT === undefined ? 8123 : Number(process.env.HOT_TEST_PORT);

if (!Number.isInteger(PORT) || PORT < 0 || PORT > 65535) {
  throw new Error(`HOT_TEST_PORT must be an integer port, got "${process.env.HOT_TEST_PORT}".`);
}
```

.ai/WORKTREES.md
```diff
@@ -0,0 +1,142 @@
...
+
+Nothing here changes anything for you. The bootstrap script exits immediately in a
+normal clone, the SessionStart hook it wires in prints nothing there, and every
+default — ports, scripts, test commands — is unchanged. `HOT_TEST_PORT` is a new
+variable name that nothing sets unless you set it.
+
+## The one command
+
+```bash
+node scripts/claude/setup-worktree.mjs
```
budnix: **Claude Code has a first-class mechanism for part of this, and neither the doc nor the script uses it.**

`.worktreeinclude` (project root, gitignore-style patterns) copies gitignored files into worktrees that Claude Code creates - documented at code.claude.com/docs/en/worktrees. It covers `.claude/settings.local.json` (step 2 of the bootstrap), and also `docs/.env` and `docs/tests/.env`, which this page's table omits entirely. It removes a manual step rather than automating one.

Related: there is a `worktree.symlinkDirectories` setting. Line 33 asserts "The harness symlinks only the **root** `node_modules`" as if it were a fact of nature, but that symlink is configurable - and configuring it away is the at-source fix for the trap described under "Never symlink `node_modules`", instead of deleting the link after the fact on every bootstrap.

This does **not** replace the script: `.worktreeinclude` only fires for Claude-created worktrees (not `git worktree add`), cannot reach `~/.claude/projects/` for the memory link, cannot run an install, and cannot build the graph. Please add the file for what it does cover, name `worktree.symlinkDirectories` here, and state which worktree-creation flows still require the bootstrap - right now a reader cannot tell.

.ai/WORKTREES.md
```diff
@@ -0,0 +1,142 @@
...
+- Bootstrap a worktree before building or testing in it. Do not debug a failure in an
+  unbootstrapped worktree — fix the setup first.
+- Never symlink `node_modules`.
+- Never commit in a worktree without asking. A commit blocks the worktree from being
+  removed cleanly.
+- Rebase a long-lived worktree. It carries the `AGENTS.md`, `.ai/`, and
+  `.claude/skills/` of the commit it branched from, so guidance drifts as `develop`
+  moves.
+- Check the SessionStart banner. It reports which branch the code-review graph was
+  built on, and a mismatch means cross-file answers are untrustworthy.
```
budnix: **The local enforcement hooks are silently dead in a worktree, and this page says nothing about it.**

`.claude/settings.json` wires `PostToolUse` -> `post-tool-use.mjs` and `Stop` -> `stop.mjs`. Both call `toRepoRelative(file, repoRoot())`. If `$CLAUDE_PROJECT_DIR` stays at the main checkout (see my comment on `.claude/settings.json`), `repoRoot()` is the main checkout, and an edit to `.claude/worktrees/X/handsontable/src/foo.ts` normalizes to `.claude/worktrees/X/handsontable/src/foo.ts` - a valid repo-relative path that matches **no** classifier in `presence-gate.mjs`. It is not rejected, it just never matches.

Net effect in every worktree session: the spec autolint does not run, the new-Jasmine-spec block does not fire, and the Stop verdict is always `ok`. All of it fails open, with no message.

This page promises the complete picture of what a worktree is missing, and the enforcement layer is a bigger gap than the code-review graph it does list. Please add a row to the table and a rule here - or fix the hooks to resolve the checkout from the payload `cwd`.

scripts/claude/setup-worktree.mjs
```diff
@@ -0,0 +1,422 @@
...
+    // --check runs from the SessionStart hook in EVERY session. Stay silent in
+    // the main checkout: anything printed here is injected into the agent's
+    // context on every start, for a condition that is not a problem.
+    if (!checkOnly) {
+      log('Not a linked worktree — nothing to bootstrap.');
+    }
+  } else if (checkOnly) {
+    await reportReadiness();
+  } else {
+    await bootstrap();
```
budnix: **No `try`/`catch` around `bootstrap()` - a late throw discards the marker after the expensive step already succeeded.**

`copyLocalSettings()` (`copyFile`), `linkProjectMemory()`'s `mkdir`, and `writeMarker()`'s `writeFile` are all unguarded. `linkProjectMemory` catches its own `symlink` failure precisely because Windows needs Developer Mode - the same reasoning applies to the neighbouring calls, which it does not cover.

Any of them throwing produces an unhandled rejection: a raw stack trace, no marker, and a `--check` that reports "never bootstrapped" on every session start even though the install completed. Re-running then repeats a multi-minute `pnpm install` to fix what was a `copyFile` permission error.

Wrap it, report which step failed, and still write the marker when the dependencies are present.

scripts/claude/setup-worktree.mjs
```diff
@@ -0,0 +1,422 @@
...
+
+const ROOT = repoRoot();
+const checkOnly = process.argv.includes('--check');
+const dryRun = process.argv.includes('--dry-run');
+const skipInstall = process.argv.includes('--skip-install');
+
+// Written last, so its presence means every earlier step succeeded. Kept in the
+// worktree's own git directory: per-checkout state that is never committed and
+// disappears with the worktree itself.
+const MARKER = 'claude-worktree-ready';
```
budnix: **The marker records only "someone ran this once", so it goes stale silently as the script grows.**

`reportReadiness()` treats the file's mere presence as "bootstrapped". A worktree set up by today's script keeps reporting ready after a future version adds a step (copying `.worktreeinclude` targets, seeding the graph, a new package's `node_modules`) - the new gap never surfaces, and there is no way to tell a v1 worktree from a v3 one. Long-lived worktrees are exactly what the "Rebase a long-lived worktree" rule in `.ai/WORKTREES.md` anticipates.

The file already holds a timestamp; make it hold a version too (`{ version: 2, at: ... }`) and have `--check` say "bootstrapped by an older script - re-run" when it lags.

.ai/WORKTREES.md
```diff
@@ -0,0 +1,142 @@
...
+| `.claude/settings.local.json` | Gitignored. | The permission allowlist is gone. The agent is prompted for `git`, `gh`, and `npm` calls that are approved in the main checkout. |
+| Claude project memory | Lives in `~/.claude/projects/<slug>/`, keyed by the **cwd path**. A worktree has a different path, so it gets a different, empty directory. | Every accumulated project fact is unavailable. The agent loses the conventions it learned, so it falls back to generic behavior and reaches for repo skills less often. |
+| `.code-review-graph/` | Gitignored, and stamped with the branch it was built on. | Cross-file queries either fail or answer from another branch's structure. |
+| `handsontable/tmp`, `dist`, `styles` | Build outputs. | Wrappers cannot resolve the core until you build it. |
+| `dev*.html` demo pages | Gitignored. | Manual test pages from the `handsontable-demo-page` skill do not carry over. |
+| The `prepare` script's effects | It runs on `pnpm install`. | Without a real install, lefthook is not wired and `.cursor/rules/` is not synced. |
```
budnix: **The table omits the gitignored `.env` files.**

`.gitignore` lists `docs/.env` and `docs/tests/.env`. Neither reaches a worktree, and neither is copied by the bootstrap, so docs work and the docs test suite fail there for a reason this page does not explain. They are also the textbook `.worktreeinclude` case (see my comment on line 22).

Worth its own row, next to `.claude/settings.local.json`.

.ai/WORKTREES.md
```diff
@@ -0,0 +1,142 @@
...
+the files are copy-on-write clones, so true disk use is lower than that number
+suggests — but it is not zero, and the trees add up.
+
+List what exists and remove what is finished:
+
+```bash
+git worktree list          # `prunable` marks a directory that is already gone
+git worktree prune         # clears the stale bookkeeping
+git worktree remove <path> # removes a finished worktree
+```
```
budnix: **Housekeeping covers disk, but not the shared git state - which is what actually surprises an agent.**

A linked worktree shares the common git directory with the main checkout. Three consequences an agent will hit and cannot diagnose from this page:

1. **A branch can only be checked out in one worktree.** `git checkout develop` in a worktree fails with "already checked out at ...", which reads like a corrupt worktree and is just git.
2. **Stash, config, and refs are shared.** A `git stash` in a worktree is visible - and poppable - from the main checkout.
3. **`pnpm install` in a worktree rewrites the main checkout's git hooks.** `git rev-parse --git-path hooks` resolves to `<main>/.git/hooks` even from inside a worktree; I verified that here. The install's `prepare` step runs `lefthook install`, which regenerates the **shared** `pre-commit`/`pre-push` with the worktree's absolute lefthook binary path and leaves a `pre-commit.old` behind. The generated stub's fallback chain does survive the worktree being deleted, so this is a papercut rather than a break - but the bootstrap command this page recommends causes it, and nothing here says so.

tests/playwright.config.ts
```diff
@@ -49,15 +60,18 @@ export default defineConfig<TestOptions>({
...
   // Serves the repo root so the fixture page can load the built handsontable
   // dist and styles. cwd defaults to this config's directory (`tests/`).
   webServer: {
     command: 'node support/static-server.mjs',
-    port: 8123,
+    port: PORT,
+    // static-server.mjs reads PORT. Passing it explicitly keeps the server and
+    // the config on the same port even when the surrounding shell exports its own.
+    env: { PORT: String(PORT) },
     reuseExistingServer: !process.env.CI,
```
budnix: **`HOT_TEST_PORT` documents the trap rather than closing it - the default path is still silently wrong.**

The port is opt-in, so the default run of two checkouts still collides, and `reuseExistingServer` still attaches without checking **whose** build is being served. Everything above this line makes the collision avoidable; nothing makes it detectable. An agent who has not read `.ai/WORKTREES.md` gets green specs describing another checkout's code - the exact silent wrong answer this PR set out to eliminate.

A deeper fix that changes no default: have `support/static-server.mjs` answer a `/__whoami` route with its `ROOT`, and have this config (or a `globalSetup`) refuse to reuse a server whose `ROOT` is not this checkout. It turns a silent wrong answer into a one-line error, and it also covers the `visual-tests/playwright.config.ts` case that `.ai/WORKTREES.md` calls "the worst case" and then leaves entirely unguarded.

scripts/claude/setup-worktree.mjs
```diff
@@ -0,0 +1,422 @@
...
+
+  // A symlinked root node_modules resolves root binaries, so lint passes and the
+  // worktree looks healthy, but every package-local `.bin` is missing and
+  // `npm --prefix <pkg> run <task>` dies with exit 127 partway through a build.
+  const link = await lstat(nodeModules).catch(() => null);
+
+  if (link?.isSymbolicLink()) {
+    problems.push('node_modules is a symlink to the main checkout (breaks package scripts)');
+  }
+  if (!(await exists(path.join(ROOT, 'handsontable', 'node_modules')))) {
```
budnix: **The readiness probe misses the one missing directory the test suite already knows to look for.**

`tests/support/static-server.mjs` preflights `tests/node_modules/hyperformula/dist/hyperformula.full.min.js` and warns when it is absent, because a partial install turns the formulas fixtures into bare 404s mid-run. That is precisely a repairable gap - re-running the bootstrap fixes it - so by this function's own stated rule it belongs here, where it is reported *before* a run rather than as a warning buried in server output.

Same reasoning for `writeMarker()`, which also gates on `handsontable/node_modules` alone: a `pnpm install` that fails partway can satisfy that single check and still leave the worktree marked ready.

scripts/claude/setup-worktree.mjs
```diff
@@ -0,0 +1,422 @@
...
+    log('Project memory is a real directory here — left untouched.');
+
+    return;
+  }
+  if (entry?.isSymbolicLink()) {
+    // A link whose target moved reads as "absent" to --check and as "present" to
+    // a naive lstat guard, so the warning would repeat forever while every
+    // re-run reported success. Repair it instead.
+    if (await exists(worktreeMemory)) {
+      log('Project memory already linked — left untouched.');
```
budnix: **An existing symlink is accepted without checking where it points.**

The branch only asks whether the link resolves, not whether it resolves to `mainMemory`. A link left over from a moved or renamed main checkout - or one aimed at another worktree's memory - resolves fine, so this reports "already linked", `--check` reports ready, and the session silently reads the wrong project's memory. That is the same class of no-symptom failure the dangling-link branch directly below goes to the trouble of repairing.

Comparing the target costs one `readlink`:

```suggestion
    if (await readlink(worktreeMemory).catch(() => null) === mainMemory) {
      log('Project memory already linked - left untouched.');

      return;
    }
```

scripts/claude/__tests__/setup-worktree.test.mjs
```diff
@@ -0,0 +1,65 @@
...
+  // Whether Claude Code also rewrites an underscore (legal in a worktree name)
+  // or a Windows drive colon is NOT established. Both readings are returned so
+  // the caller can pick whichever exists, instead of guessing and silently
+  // linking into a directory nothing reads.
+  const candidates = projectStateCandidates(`${MAIN}/.claude/worktrees/fix_DEV-2562`);
+
+  assert.equal(candidates.length, 2);
+  assert.ok(candidates[0].endsWith('fix_DEV-2562'), 'preferred reading keeps the underscore');
+  assert.ok(candidates[1].endsWith('fix-DEV-2562'), 'fallback reading replaces it');
+  assert.notEqual(candidates[0], candidates[1]);
```
budnix: **This test locks in the wrong preference order.**

`'preferred reading keeps the underscore'` asserts the behavior that breaks the memory link (see my comment on `setup-worktree.mjs` line 119). Measured here, Claude Code writes `feature/DEV-1656_Autocomplete-...` as `feature-DEV-1656-Autocomplete-...`, and none of the 34 directories under `~/.claude/projects/` contains an underscore.

So the fix is not only a source change - these two assertions have to be inverted with it, or the suite will defend the bug. Worth restating the intent in the test name as well: the question is which reading is *tried first*, and the evidence says it should be the one that replaces underscores.

The two tests above are unaffected: `issue-9063` and the main-checkout path contain no underscore, so both readings agree there.

qunabu:

tests/playwright.config.ts
```diff
@@ -1,6 +1,43 @@
...
+function resolvePort(): number {
+  const raw = process.env.HOT_TEST_PORT;
+
+  if (raw === undefined || raw === '') {
+    return 8123;
+  }
+
+  const port = Number(raw);
+
+  if (!Number.isInteger(port) || port < 0 || port > 65535) {
```
demtario: `HOT_TEST_PORT=0` passes this guard (`port < 0`), but every layer below it disagrees, so the run silently lands back on the shared port.

- `tests/support/static-server.mjs:15` is `Number(process.env.PORT) || 8123`, and `Number('0')` is falsy, so with `env: { PORT: '0' }` (line 100) the server binds **8123**. With `reuseExistingServer` on, a second checkout then attaches to the first one's server and build, which is the exact collision this variable was added to escape.
- `baseURL` (line 89) renders `http://localhost:0`, so no `page.goto('/...')` can reach it.
- Playwright's `webServer` readiness gate is `if (webServerConfig.port || webServerConfig.url)` (`playwright/lib/runner/index.js:969`, 1.61.1), falsy at `port: 0`, so the server is launched without being waited for.

The `@returns` JSDoc at line 22 calls `0` "the safest choice when several checkouts run at once", so the only assertion about this value points at the broken one, while the comment at lines 13-16 already names this precise failure mode.

Fix: reject `port < 1` and drop the "0 is allowed" sentence. A genuine dynamic port needs the server to report its bound port back, which is already listed under "Known gaps" in `.ai/WORKTREES.md`.

Worth naming why this shipped: `resolvePort` is inline in a config module with a top-level `const PORT = resolvePort()` side effect (line 17), so nothing can unit-test it, and `port: 0` type-checks. `test:tooling` in the root `package.json` already globs sibling `__tests__` directories, so extracting this into a small module costs one glob entry.

tests/playwright.config.ts
```diff
@@ -1,6 +1,43 @@
 import { defineConfig, devices } from '@playwright/test';
 import type { TestOptions } from './fixtures/test';
 
+// `reuseExistingServer` attaches to whatever already listens on this port. Two
+// checkouts (a worktree and the main clone) on the default port therefore share
+// ONE server, and the second one silently tests the first one's build. Set
+// `HOT_TEST_PORT` to run them side by side.
```
demtario, line 7: `HOT_TEST_PORT` is documented in root `AGENTS.md` and `.ai/WORKTREES.md`, but not where someone running this suite would look. `tests/README.md` has a "Running locally" section that never mentions it, and `tests/AGENTS.md` never mentions the port at all. The knob is owned by the `tests/` package, and `tests/AGENTS.md` is what auto-loads for an agent working in this subtree, so per checklist item 3 in root `AGENTS.md` it belongs there too. One line in each.

scripts/claude/setup-worktree.mjs
```diff
@@ -0,0 +1,507 @@
...
+function resolveRoot() {
+  if (!checkOnly) {
+    return repoRoot();
+  }
+
+  let payload = {};
+
+  try {
+    // Read fd 0 directly — cross-platform (a `cat` spawn ENOENTs on Windows).
+    payload = JSON.parse(readFileSync(0, 'utf8'));
```
demtario: `readFileSync(0, 'utf8')` reads to EOF, so `--check` blocks on an interactive terminal until the developer presses Ctrl-D. Reproduced with a FIFO holding an open writer: still waiting after 3s. This read also happens before the worktree test, so it blocks in a plain clone too, and `.ai/WORKTREES.md` documents `--check` as a command to run by hand ("reports readiness without writing anything").

Fix: guard with `if (!process.stdin.isTTY)` and fall through to `repoRoot()`.

Two related notes. The hook path depends on being the only stdin reader in its `SessionStart` array and on the harness closing the pipe; if Claude Code ever holds stdin open, every session start in every clone burns the full `timeout: 20`. And this is now the third verbatim copy of this read plus its comment (`post-tool-use.mjs:27`, `stop.mjs:32`), while `scripts/claude/session.mjs` is the declared home for shared hook helpers, so a `readHookPayload()` there would fit.

scripts/claude/setup-worktree.mjs
```diff
@@ -0,0 +1,507 @@
+/**
+ * Bootstraps a linked git worktree so it behaves like the main checkout.
+ *
+ * A `git worktree` only materialises TRACKED files, so anything gitignored is
```
demtario: `materialises` -> `materializes`. American English is rule 4 in `.ai/DOC-STANDARDS.md`, and its scope preamble names code comments. `.ai/WORKTREES.md:5` and the new root `AGENTS.md` gotcha both already spell it with a z, so this is the odd one out.

scripts/claude/setup-worktree.mjs
```diff
@@ -0,0 +1,507 @@
...
+ * stored as `…--claude-worktrees-feature-DEV-1656-Autocomplete-dropdown-flex-layout`
+ * — the `.`, the `+` and the `_` all became dashes.
+ *
+ * That underscore matters here: this repository names branches
+ * `feature/DEV-xxxx_Name`, so most worktree directories contain one. Preferring
+ * a spelling that keeps it would put the memory link in a directory Claude Code
+ * never reads, and nothing would report a problem.
+ *
+ * The second candidate is the narrower historical reading, kept only so an
+ * already-linked older worktree is still recognised rather than relinked.
```
demtario: `recognised` -> `recognized`, same rule (rule 4 names `recognize` as one of its own examples).

scripts/claude/setup-worktree.mjs
```diff
@@ -0,0 +1,507 @@
...
+}
+
+/**
+ * The project state directory for a checkout: the first candidate that exists,
+ * falling back to the verified spelling when none does.
+ *
+ * @param {string} checkout The checkout path.
+ * @returns {Promise<string>} The project state directory.
+ */
+async function projectStateDir(checkout) {
```
demtario, line 179: `projectStateDir` returns the first candidate that exists, and the caller then uses that same value as the write target. Recognition wants first-existing; writing does not. If the legacy spelling (candidate[1], underscores kept) exists while the correct one (candidate[0]) does not, `linkProjectMemory` reports "Project memory already linked" (line 425) while the directory Claude Code actually reads stays empty, and `--check` keeps reporting ready forever.

Narrow precondition (a worktree bootstrapped by an older script where no session ever started, so candidate[0] was never created), but the symptom is the invisible kind this PR exists to remove. Fix: keep first-existing for recognition, always write to `candidates[0]`.

scripts/claude/setup-worktree.mjs
```diff
@@ -0,0 +1,507 @@
...
+
+  const install = spawnSync('pnpm', ['install', '--frozen-lockfile'], {
+    cwd: ROOT,
+    env,
+    stdio: 'inherit',
+    shell: process.platform === 'win32',
+  });
+
+  if (install.status !== 0) {
+    log('Install failed.');
```
demtario: `spawnSync` returns `status: null` with the reason in `install.error` when the binary cannot be spawned at all (no `pnpm` on `PATH`). `null !== 0` is true, so this branch fires and discards the ENOENT behind a bare `Install failed.`, immediately after the code may have removed the worktree's `node_modules` symlink. Suggest surfacing it:

```js
log(`Install failed${install.error ? `: ${install.error.message}` : '.'}`);
```

.worktreeinclude
```diff
@@ -0,0 +1,33 @@
...
+# hook replaces the mechanism entirely and must copy these itself.
+#
+# This covers only the file half of setting up a worktree. It cannot install
+# dependencies, and it cannot reach the Claude project memory, which lives
+# outside the repository — run `node scripts/claude/setup-worktree.mjs` for
+# those. See .ai/WORKTREES.md.
+
+# Docs site secrets. Absent from a fresh worktree, so docs work and the docs
+# test suite fail there with no explanation.
+docs/.env
```
demtario, line 15: Contained, but worth one line of guidance. `.claude/worktrees/**` is gitignored (`git check-ignore -v` confirms it via `.claude/*`), so these never reach a commit. The key names in the docs env surface are real though: `DB_PASS`, `DB_USER`, `GH_TOKEN`, `VITE_SUPABASE_ANON_KEY`, and `PASS_COOKIE` (the staging bypass read by the `docs/tests/` specs). On a maintainer's machine this copies live credentials into every worktree Claude Code creates. The Housekeeping section of `.ai/WORKTREES.md` covers disk and `~/.claude/projects` accumulation but never mentions credentials, nor that a worktree which is never `git worktree remove`d keeps holding them. One line in Housekeeping closes it.

.worktreeinclude
```diff
@@ -0,0 +1,33 @@
+# Gitignored files copied into each worktree Claude Code creates.
+#
+# Uses .gitignore syntax. Only files that match a pattern AND are gitignored are
+# copied, so nothing tracked is ever duplicated. Applies to worktrees created by
```
demtario, line 4: This half of the PR cannot be verified from inside the repository, and it is the half with no failure signal: if the filename or the "matches a pattern AND is gitignored" semantics are wrong, the whole file-copy mechanism is a silent no-op and nothing reports it. That is the same failure class as the slug bug fixed in the second commit. Every worktree in this checkout predates the file, so `docs/.env` being absent in all of them is inconclusive rather than disconfirming. How was it verified, ideally against a worktree created after this file landed?

.ai/WORKTREES.md
```diff
@@ -0,0 +1,190 @@
+# Worktrees — what a linked checkout is missing, and how to fix it
+
+Agents in this repo increasingly work in a linked git worktree under
+`.claude/worktrees/`. That isolates parallel work, but a worktree is not a copy of
+your checkout. `git worktree` materializes **tracked files only**. Everything else
+an agent depends on is gitignored or lives outside the repo, so it is absent the
+moment the worktree is born.
+
+This page lists what is missing, what each gap actually breaks, and the one command
```
demtario, line 9: This promises "the one command that closes them", but line 24 says `.worktreeinclude` does not run for a worktree made with `git worktree add` by hand, and the table at line 45 marks the `docs/.env` row as "Covered by `.worktreeinclude`". So in a hand-made worktree nothing closes that gap: `setup-worktree.mjs` does not copy the two files, and `reportReadiness()` does not probe for them, so it reports ready while docs work fails there for a reason nothing explains. Either copy the two files in `bootstrap()` (which makes the "one command" claim true), or soften this sentence and add them to the readiness probe.

.ai/WORKTREES.md
```diff
@@ -0,0 +1,190 @@
...
+
+Run that once per worktree, before you build or test. It is idempotent, so
+re-running is safe. `--check` reports readiness without writing anything, and the
+SessionStart hook uses it to warn you.
+
+## What is missing, and what it breaks
+
+| Missing | Why | What breaks |
+|---|---|---|
+| Per-package `node_modules` | Not tracked, and the root one is created as a symlink. | `npm --prefix handsontable run test:unit` exits 127 with `env-cmd: command not found`. Every package script fails the same way. |
```
demtario: "and the root one is created as a symlink" over-commits. None of the four worktrees in this checkout matches it: three have real `node_modules` directories, and the one created to review this PR has none at all. The code handles both shapes; only the prose does not, and it sends a reader hunting for a symlink that may not be there. Suggest "may be created as a symlink, or be absent entirely".

.claude/settings.json
demtario: `|| true` is load-bearing here (a non-zero exit would suppress the stdout this hook exists to inject), but `2>/dev/null` is not, and this is the only one of the four hook commands that redirects stderr. It permanently silences a crash in the hook whose entire job is to report a silent problem. The sibling `post-tool-use.mjs` and `stop.mjs` hooks do not do it. Suggest dropping the redirect and keeping `|| true`.

tests/playwright.config.ts
budnix: Nice catch — 0 now throws as well, fixed in `ef79feda72`; the server really did bind 8123 for it.

tests/playwright.config.ts
budnix: Good point — documented in both `tests/AGENTS.md` and `tests/README.md` in `ef79feda72`.

scripts/claude/setup-worktree.mjs
budnix: You're right, that hung on a terminal. Fixed in `ef79feda72` — moved to `readHookPayload()` in `session.mjs` with the isTTY guard.

scripts/claude/setup-worktree.mjs
budnix: Good spot — fixed in `ef79feda72`.

scripts/claude/setup-worktree.mjs
budnix: Yep, my mistake — fixed in `ef79feda72`.

scripts/claude/setup-worktree.mjs
budnix: Fair point — writes now always go to `projectStateWriteDir()` in `ef79feda72`, and the readiness check reads the same place.

scripts/claude/setup-worktree.mjs
budnix: Agreed — surfacing `install.error` now, in `ef79feda72`.

.worktreeinclude
budnix: Good catch — Housekeeping now names the keys and says to remove finished worktrees, in `ef79feda72`.

.worktreeinclude
budnix: Fair challenge — I checked the filename and semantics against code.claude.com/docs/en/worktrees, but not end-to-end; a `claude --worktree` run with a real `docs/.env` present is still needed.

.ai/WORKTREES.md
budnix: You're right — `bootstrap()` now copies the plain entries itself in `ef79feda72`, so the claim holds for a hand-made worktree too.

.ai/WORKTREES.md
budnix: Good spot — reworded to "may be created as a symlink, or be absent entirely" in `ef79feda72`.

.claude/settings.json
budnix: Agreed — dropped the redirect in `ef79feda72`, `|| true` stays.

demtario (approved):

Issue #3850, opened 2016-10-31
Flickering headers while scrolling horizontally


krzysztofspilka:
### Description
When I scroll the grid horizontally, the headers of rows are constantly changing their width for about 1-2px. That is worsening the user experience. I really don't like this kind of jumping behavior in Handsontable.

### Steps to reproduce

1. Go to https://docs.handsontable.com/pro/1.7.4/demo-scrolling.html
2. Scroll the grid to the right

### Demo

https://gyazo.com/011c589d8e35a4f070c452e864e93917

### Your environment
* Version: Handsontable 0.28.4
* Browser Name and version: Chrome 54
* Operating System: Windows 8.1

krzysztofspilka:
@budnix tested, when fixed columns are set, the 1px difference still exists

Issue #475, opened 2013-03-06
Support variable width row headers


chahu:
I'm working on a project which is attempting to use handsontable, however we're running up against the fact that the row-header-column is a fixed width (and not very wide).  This cause the vast majority of our row header labels to be truncated.  One alternative is to tweak the CSS and widen that first column a fixed amount, but this is not an option as the data we are displaying could vary in length/width.  I've also tried changing the CSS and commenting out some of the lines with "width: 0" -- this "works" in that the row-header-column does indeed become variable width, but the scrolling logic becomes buggy.

Long story short, how can I get the below to show my row headers (without having a fixed-width column) and keep the nice scrolling functionality:

[http://jsfiddle.net/4587J/6/](http://jsfiddle.net/4587J/6/)

Thanks!
Charlie

tartor:
Workaround 
in the config implement afterLoadData callback to change the width of all the elements that have class name 'col.rowHeader'

var ht = new Handsontable(
                document.getElementById(id),
                {
                    afterLoadData: function(){
                        var rowHeaders = document.getElementById(id).querySelectorAll('col.rowHeader');
                        if (rowHeaders.length != 0) {
                            for (var j = 0; j < rowHeaders.length; j++) {
                                rowHeaders[j].style.width = "200px";
                            }
                        }
                    }
                }
            );

where id is the id of the html element which holds the handsontable

realtebo:
Is there any CSS way to defined a 'auto-width' for row headers ?

donmccurdy:
Could this issue be reopened? The JSFiddle posted by the OP ([which I've updated with fixed links to CDN js/css](http://jsfiddle.net/donmccurdy/vzp31798/)) is still showing row headers fixed at 50px. As pointed out, the 50px can be overridden in CSS, but this isn't ideal.

@realtebo No, not without using JavaScript to calculate what the magic width should be, and then applying it in CSS.

AMBudnik:
@donmccurdy this issue is continued (and marked as a bug) at https://github.com/handsontable/handsontable/issues/2868

Issue #402, opened 2013-02-13
Row headers do not fit in cells


marint:
Hi,

It seems that in version 0.8.3 the width for row headers column in always 50px. In earlier versions, the width scaled up if row header contents were longer.

For example, if row headers contain dates and times, they do not fit in the column. With version 0.8.2, this worked correctly:

http://jsfiddle.net/ZHx7G/

PR #13239, opened 2026-08-25
DEV-2623: Add the AutoRowHeaderSize plugin

### LINKED ISSUE #7147: Cannot change the width of row headers and does not expand when row headers are given long names
### Description

The rowHeader cells have a fixed width. When I change the input of the rowHeaders tag and the headers have quite long names then what should happen is that as with the colHeaders it should expand upon enter to show all of the content. However it does not and you cannot see the full length of the word(s) representing the row headers.

### Steps to reproduce

1. Create a table
2. Change rowHeaders to:
        hotSettings: {
            colHeaders:  ["Banana","Pineapple"],
            rowHeaders: ["Banana","Pineapple"],
        },
3. When the table is rendered onto the page it should show that the row headers do not expand to show the full word(s) representing the headers

### Demo

**Please note that this is a public fiddle and not my own creation.** 

http://jsfiddle.net/knv5Lbh3/

To get the example the following needs to be changed from
```
  rowHeaders: true,
  colHeaders: true,
```
to =>
```
  rowHeaders:  ["Banana","Pineapple"],
  colHeaders: ["Banana","Pineapple"],
```

### Your environment
* Handsontable version: 7.4.2
* Browser Name and version: Chrome 84.0.4147.89
* Operating System: macOS Catalina 10.15.2

budnix:
### Context

The PR adds row header sizing, so a long row header label is no longer clipped.

Reported as [#7147](https://github.com/handsontable/handsontable/issues/7147). The same string that makes a **column** header widen its column leaves a **row** header at a fixed 50px, showing `Pineap` instead of `Pineapple juice concentrate`. Two earlier reports, [#402](https://github.com/handsontable/handsontable/issues/402) and [#475](https://github.com/handsontable/handsontable/issues/475), asked for the same thing and were closed in 2013 by a change that made the viewport *read back* the row header width — it never made that width follow the content.

The cause is that nothing measures row header text. With no `rowHeaderWidth`, `_correctRowHeaderWidth()` substitutes `defaultColumnWidth` (50), which is written as an inline style on `col.rowHeader` and pinned by `table-layout: fixed`. `AutoColumnSize` only registers `modifyColWidth`, and the header column never passes through it.

Seen as a matrix, three of the four grid dimensions were already sized and one was not:

|  | Sizes the data area | Sizes the header |
|---|---|---|
| **Heights** (`AutoRowSize`) | row heights | column header height — `getColumnHeaderHeight()` |
| **Widths** (`AutoColumnSize`) | column widths | **row header width — nothing** |

This PR fills that quadrant with a new `AutoRowHeaderSize` plugin. Turning it on is the only thing you need to do:

```js
rowHeaders: ['Revenue', 'Cost of goods sold', 'Gross profit'],
autoRowHeaderSize: true,

// or, with the measurement tuned
autoRowHeaderSize: {
  samplingRatio: 3,             // how many labels of the same length get measured
  allowSampleDuplicates: false, // whether a repeated label is measured more than once
},
```

The name follows the shape its siblings use — `AutoRowSize` sizes a row, `AutoColumnSize` sizes a column, `AutoRowHeaderSize` sizes a row header — and mirrors to `AutoColumnHeaderSize` if the column header's height ever moves out of `AutoRowSize`.

**One option, not two.** The plugin takes the row header's width over, so a `rowHeaderWidth` already in the settings is simply ignored while the plugin is enabled — there is no second option to keep in step. `rowHeaderWidth` keeps meaning exactly what it always meant: a manual width, a number or an array. The default column width stays the floor of every level, the same floor `AutoColumnSize` keeps for a data column, so a grid of short labels looks as it does today instead of collapsing around a one-character label.

**It is off by default, and that is deliberate.** Turning it on for everyone would change the rendered width of every existing grid that uses custom row labels, and finding the longest label means reading every row header, so that first pass grows with the row count (measured: 2.7ms at 1k rows, 178ms at 500k; the result is cached, so later draws cost nothing). This mirrors why `AutoRowSize` is opt-in while `AutoColumnSize`, bounded by the column count, is not.

**Why a new plugin rather than folding it into a sibling.** `AutoColumnSize` is the right dimension, but it disables itself whenever `colWidths` is set (`isEnabled()`), and that gate is load-bearing: `Core#getColWidth` applies `colWidths` first and then lets `modifyColWidth` override it. Folding the feature in would mean `colWidths: [100, 100]` silently kills row header sizing — for exactly the developers most likely to want it. `AutoRowSize` would instead require `autoRowSize: true`, opting into measuring the height of every cell in the grid just to learn how wide a label is.

**No new measurement code.** The plugin composes the two utilities that already exist: `SamplesGenerator` picks which labels are worth rendering (bucketing by length, capping per bucket), and `GhostTable` renders and measures them. The row header is sampled as column `-1`, matching how `AutoRowSize` treats row `-1` as the column header. The only addition to `GhostTable` is `addRowHeadersColumn()`, the mirror of the existing `addColumnHeadersRow()` — 825 bytes minified, measured. Both utilities are held in `#private` fields: how the measuring is done is not part of what the plugin offers.

#### Multiple row headers, each sized on its own

A grid can render more than one row header by pushing a renderer through `afterGetRowHeaderRenderers`. **Every one of them is measured separately**, so a narrow numbering column stays narrow beside a wide label column — measured on a two-level grid: `[197px, 50px]`, rather than one width applied to both.

Getting there needed a small change in the rendering engine, because the `modifyRowHeaderWidth` hook is consumed in two places that want different shapes: `ColumnUtils#calculateWidths` indexes an array per level, while `Viewport#getRowHeaderWidth` wants a single number for the whole header block. The viewport now reduces the hook's answer through `toTotalRowHeaderWidth()` — summing an array, passing a number through, and returning `null` for `0`/`NaN` so the previous `|| this.rowHeaderWidth` fallback behaves exactly as before.

That makes **`modifyRowHeaderWidth` a `number | number[]` hook**, widened in its TypeScript signature, its JSDoc, the `TableView` callback, and the type regression test. One consequence worth flagging for review: `NestedRows` is the other handler of that hook and did `Math.max(cache, width)`, which turns an array into `NaN` — and `ColumnUtils`' `??` does not catch `NaN`, so the column would have rendered at `NaNpx`. It now widens an array entry by entry.

Two details that make the per-level measurement correct rather than approximate:

- Each level renders with **its own renderer**, collected the way `TableView` collects them for the draw, so a level's width covers the markup it actually draws instead of assuming the first level's.
- The renderers receive **renderable** row indexes, not visual ones, so each sampled row is translated through `getRenderableFromVisualIndex()` before its renderer runs. Hidden rows are skipped entirely — no header is drawn for them.

The public API follows:

```js
const plugin = hot.getPlugin('autoRowHeaderSize');

plugin.getRowHeaderWidth();    // the first row header (default)
plugin.getRowHeaderWidth(1);   // the second level
plugin.getRowHeaderWidth(-1);  // same as 0 — the negative column index a row header sits at
plugin.getRowHeaderWidths();   // every level, in order
```

#### Two pre-existing defects in `AutoRowSize` fixed along the way

1. It declared a `useHeaders` setting in `DEFAULT_SETTINGS` that it never read. Only `AutoColumnSize` reads one. Removed, with a comment explaining why this plugin has no such question to ask.
2. The `autoRowSize` docs claimed `rowHeights` forcibly disables the plugin, mirroring what `colWidths` does to `AutoColumnSize`. It does not: `isEnabled()` never checks `rowHeights`, and a supplied height acts as a **minimum** — a row whose content is taller keeps its measured height, because a column can clip its content but a row would hide it. The documentation now describes the actual, intended behavior.

#### Known gap

There is no automated assertion of the rendered width on screen. That behavior was verified by hand in Chrome against a local build. A Playwright spec in `tests/e2e/` would close this and is the natural follow-up.

### Scaling to large grids

Finding the longest label means reading every row header. That was one synchronous pass inside a
draw, which measured badly in Chrome on the branch build:

| rows | 1 header column | 2 header columns |
| --- | --- | --- |
| 200 000 | 153 ms | 1 474 ms |
| 1 000 000 | 698 ms | 10 591 ms |

It was also paid again on **every cell edit**, because `afterSetDataAtCell` dropped the whole cache:
at 100 000 rows a single `setDataAtCell` cost +56 ms with one header column and **+1 280 ms** with
two.

The plugin now follows `AutoRowSize`. The first `syncLimit` rows (500 by default, or a percent
string such as `'40%'`) are read before the first paint; the rest are swept in `requestIdleTask`
chunks, each time-boxed to 8 ms. While a sweep runs a level can only **widen**, so the width never
jumps back and forth; the final commit is the one allowed to report a narrower width, which is what
lets a header shrink again after its longest label is deleted.

A cell edit re-reads only the rows named in the change. The listener has to stay — a row header
label can be built from cell values (a data column used as the label measured 47 px → 257 px after
one edit), so the fix is to narrow it, not to drop it. The width can only grow on that path;
shrinking needs the new longest label, so it waits for the next full sweep.

Reading a level added through the hook also reuses one detached cell instead of creating one per row.

Measured after the change:

| | before | after |
| --- | --- | --- |
| first paint, 200k rows / 2 header columns | 1 474 ms | **12.7 ms** |
| first paint, 200k rows / 1 header column | 153 ms | **1.9 ms** |
| one cell edit, 200k rows / 2 header columns | 1 231 ms | **0.6 ms** |

An `IndexMapper` was considered for the widths and deliberately not used: the plugin stores one
aggregate width per header level, not a value per row, and index maps key by **physical** index
while row header labels are keyed to **visual** position — verified by sorting a grid with
`rowHeaders: ['first','second','third']`, where the labels stay in display order while the data
moves under them (with `bindRowsWithHeaders` on they follow the row instead, so no single keying is
even correct). Index maps also cannot address columns −1 and −2.

### Room around the label

The ghost table measures a cell exactly as it renders. The grid's own row header renderer wraps its
label in a padded element, but a renderer pushed through `afterGetRowHeaderRenderers` writes
straight into the `th` and has no padding at all — so the longest label came out flush against the
border, with **1.17 px** of room around it. The measured width now carries an 8 px allowance.

For comparison, on the same text a column header sized by `AutoColumnSize` leaves 16 px, so the row
header is still slightly tighter; the allowance is one constant if we want them to match exactly.

### How has this been tested?

```bash
# core unit tests — 3768 pass across 313 suites

# just the new and touched areas
npm run test:unit --prefix handsontable -- --testPathPattern="autoRowHeaderSize|ghostTable|rowHeightsInteraction"

# the rendering engine, because Viewport#getRowHeaderWidth changed — 816 specs

# types

# E2E across everything that shares GhostTable
npm run test:e2e --prefix handsontable -- --testPathPattern=autoColumnSize   # 67 specs
npm run test:e2e --prefix handsontable -- --testPathPattern=autoRowSize      # 55 specs
npm run test:e2e --prefix handsontable -- --testPathPattern=nestedHeaders    # 217 specs
npm run test:e2e --prefix handsontable -- --testPathPattern=rowHeader        # 53 specs

# lint
```

Tests added:

- `src/plugins/autoRowHeaderSize/__tests__/autoRowHeaderSize.unit.ts` — 25 tests covering enabling, the plugin taking the width over from `rowHeaderWidth`, the full scan, caching, duplicate labels, index translation, and multiple row headers (per-level widths, the array answer, the negative-index form, and each level rendering with its own renderer).
- `src/utils/__tests__/ghostTable.unit.ts` — 8 tests for the new measurement column: the real `<tr>`/`<th>` chain the scoped CSS needs (#4363), the inline `table-layout`/`width` overrides without which the measurement reads 0, and the negative column index.
- `src/plugins/autoRowSize/__tests__/rowHeightsInteraction.unit.ts` — 5 tests locking the corrected `rowHeights` behavior and the settings shape.

Every one of these was mutation-checked: breaking the code under test makes them fail. One was found to be vacuous that way and rewritten.

Verified by hand in Chrome on a local build:

- The row header grows from 50px to 148px for the issue's own labels, and follows the labels as they change.
- With a second row header added through the hook, the two levels measure `148px` and `34px` and render at `148px` and `50px` (the second floored at the default), and the rendered total matches what the viewport expects — 247px against 247px on the wider fixture. Before the per-level work the same grid rendered 394px of header while the viewport still believed 197px, and the column headers drew in the wrong place.
- It holds steady while scrolling a 5,000-row grid whose only long label is on the last row — the width is decided once, so it cannot flicker the way [#3850](https://github.com/handsontable/handsontable/issues/3850) did.
- It survives `IndexMapper`: sorting ascending and descending, `manualRowMove`, trimming, and `bindRowsWithHeaders` (where the labels genuinely change position) all keep the width correct and the labels fitting.
- With `rowHeaderWidth: 400` and the plugin on, the header renders at the measured width — the manual width is ignored, as intended.

One suite, `updateSettingsExternalDataSource.unit`, times out on my machine. It fails identically with every change on this branch reverted, so it is a pre-existing environment issue, not a regression from this PR.

### Types of changes

### Related issue(s):

1. DEV-2623
2. Fixes #7147
3. Earlier duplicates of the same request: #402, #475

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

Both halves are in this PR. The JSDoc for `rowHeaderWidth`, the new `autoRowHeaderSize` option and the `modifyRowHeaderWidth` hook is updated, and the generated API page is registered (sidebar entry, plugin table row, SEO block).

The guide is updated too. The existing **Row header width** section of `docs/content/guides/rows/row-header/row-header.md` already stated the limitation this plugin removes, so it grew a second half rather than becoming a new page — the same shape the siblings have, with `AutoColumnSize` documented inside the column width guide and `AutoRowSize` inside the row height guide. It now carries two runnable examples in all four frameworks, a **Tuning the measurement** subsection explaining what `samplingRatio` and `allowSampleDuplicates` each change and when to reach for them, why the plugin is off by default, and that it never shrinks a header.

The second example is a new **Add more row header columns** subsection. It pushes two renderers through `afterGetRowHeaderRenderers` next to the default numbering, so the grid draws three row header columns and each one is measured on its own. Checked in Chrome against this branch: the three columns come out 50 / 80 / 132 px with nothing clipped, and with the plugin off all three fall back to 50 px and the two label columns clip. It also notes that a renderer receives the renderable index, and points at `getVisualFromRenderableIndex()` for grids that hide rows.

ClickUp task: https://app.clickup.com/t/86cb9mxu0

---

> [!NOTE]
> **Medium Risk**
> New layout plugin and a widened `modifyRowHeaderWidth` hook affect row-header sizing and Walkontable viewport math; regressions could show up in multi-level headers, NestedRows, and large grids, though behavior is opt-in and heavily tested.
> 
> **Overview**
> Adds the **opt-in `AutoRowHeaderSize` plugin** and the **`autoRowHeaderSize`** grid option so row header columns can grow to fit their longest label instead of staying at the fixed default width and clipping text. Enable with `autoRowHeaderSize: true` (or an object with `samplingRatio`, `allowSampleDuplicates`, `syncLimit`); while it is on, **`rowHeaderWidth` is ignored** and measured widths are floored at the default column width.
> 
> Measurement reuses **`SamplesGenerator`** and **`GhostTable`** (`addRowHeadersColumn`), samples labels by length, measures each **row header level** separately (including levels from `afterGetRowHeaderRenderers`), and splits work across a **`syncLimit` first paint** plus **idle-time sweeps**; edits only re-measure changed rows. **`modifyRowHeaderWidth`** now accepts and returns **`number | number[]`**; **`toTotalRowHeaderWidth`** sums per-level widths for the viewport, and **NestedRows** widens array answers per level so it no longer produces **`NaN`**.
> 
> Also removes the unused **`useHeaders`** default from **AutoRowSize**, clarifies that **`rowHeights` does not disable AutoRowSize**, and adds docs, changelogs, unit tests, and Playwright coverage for geometry and async width updates.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e5a774b85f643c5836a91d3111def0baa98f0493. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2623 Size row headers to their longest label (rowHeaderWidth: 'auto')](https://app.clickup.com/t/9015210959/DEV-2623)

handsontable/src/plugins/autoRowHeaderSize/autoRowHeaderSize.ts
```diff
@@ -0,0 +1,1011 @@
...
+
+    this.#samplesGenerator.setAllowDuplicates(this.getSetting<boolean>('allowSampleDuplicates'));
+
+    const samplingRatio = this.getSetting<number | null>('samplingRatio');
+
+    if (samplingRatio && !isNaN(samplingRatio)) {
+      this.#samplesGenerator.setSampleCount(parseInt(String(samplingRatio), 10));
+    }
+
+    this.addHook('modifyRowHeaderWidth', this.#onModifyRowHeaderWidth);
```
demtario: This registration carries no `orderIndex`, so its position among the `modifyRowHeaderWidth` handlers depends on plugin init order, and any `updateSettings` call reshuffles it. Because `#onModifyRowHeaderWidth` discards its incoming argument by design (line 842), running after another handler silently throws that handler's answer away.

The chain, end to end:

- Init order follows `PLUGIN_PRIORITY`: `AutoRowHeaderSize` is 45, `NestedRows` is 300, so the list starts `[AutoRowHeaderSize, NestedRows]`. `HooksBucket#insertByOrder` appends within an equal-`orderIndex` group (`core/hooks/bucket.ts:132`), and both sit at the default `0`.
- `AutoRowHeaderSize.SETTING_KEYS` is `true` (line 112), while `NestedRows` uses the inherited `['nestedRows']`.
- `BasePlugin#onUpdateSettings` calls `updatePlugin()` only when `#isRelevantToSettings(newSettings)` (`plugins/base/base.ts:581-588`). So on `updateSettings({ colHeaders: true })` only this plugin is relevant: `updatePlugin()` (line 362) runs `disablePlugin()`, whose `super` call reaches `clearHooks()` (`base.ts:270-274`), and `enablePlugin()` then re-appends the handler at the **tail**. `NestedRows` keeps its slot.
- Order becomes `[NestedRows, AutoRowHeaderSize]`, and the incoming floor is dropped.

Failure scenario: `{ nestedRows: true, autoRowHeaderSize: true }`, a tree roughly 10 levels deep, short labels (`"1"`, `"2"`, ...). `NestedRows` needs `max(50, padding * 2 + 10 * 10 + 25)`, about 145px (`nestedRows/ui/headers.ts:199`); the measured width is about 58px. First paint is 145px with the indicators visible, then after `hot.updateSettings({ colHeaders: true })` it drops to 58px and the nesting indicators and expand/collapse buttons are clipped.

This is why `AutoColumnSize` pins its own width hook at `-10` (`autoColumnSize.ts:410`): re-registration then still lands ahead of every `orderIndex: 0` handler. Same fix here:

```ts
this.addHook('modifyRowHeaderWidth', this.#onModifyRowHeaderWidth, -10);
```

The JSDoc at `nestedRows.ts:787-789` already assumes another handler may have answered *first*, so this is the order the code intends but cannot currently guarantee.

handsontable/src/plugins/autoRowHeaderSize/autoRowHeaderSize.ts
```diff
@@ -0,0 +1,1011 @@
...
+  #onModifyRowHeaderWidth = (rowHeaderWidth: number | number[]) => {
+    this.#getMeasuredWidths();
+
+    const flooredWidths = this.#flooredWidths;
+
+    if (flooredWidths === null || flooredWidths.length === 0) {
+      return rowHeaderWidth;
+    }
+
+    return flooredWidths.length === 1 ? flooredWidths[0] : flooredWidths;
```
demtario: Two problems in this return.

**1. The scalar shortcut broadcasts one width to every level.** `#flooredWidths` has the length fixed at the last `#startSweep()`, when the renderers were collected, while `ColumnUtils.calculateWidths` iterates the `rowHeaders` renderers collected on *this* draw. Nothing in the cache key (`countRows()`, `getRenderableIndexesLength()`) tracks the renderer count, so the two can disagree. When they do, `length === 1` returns a plain number, which fails `Array.isArray` in `columnUtils.ts:114`, and that single width is written to every level.

Failure scenario: an `afterGetRowHeaderRenderers` handler that pushes a second renderer conditionally on external state. Flip the flag and call `hot.render()` with no `updateSettings`: the plugin still holds one level at 148px, the draw renders two levels, `ColumnUtils` writes 148px to both (296px of header) while `Viewport#getRowHeaderWidth` reports 148px. That is the same "rendered wider than believed" misalignment the PR description says it fixes.

`#widenFor` (line 979) already treats this drift as real and bails on `renderers.length !== this.#cachedWidths.length`. The draw path has no equivalent guard. Suggest re-collecting the renderers (or folding their count into the cache key) before answering, and either dropping the scalar shortcut or taking it only when `rowHeaders.length === 1`.

**2. The live cache array is returned by reference.** A downstream `modifyRowHeaderWidth` handler that widens in place (`w[0] = 120`) mutates `#flooredWidths`, and the mutation then survives every draw until the next invalidation. The public `getRowHeaderWidths()` correctly `.slice()`s (line 435); this return does not. No in-tree caller is affected today (`NestedRows` maps, `baseTable._modifyRowHeaderWidth` copies), so this is low severity, but it is a public hook.

handsontable/src/plugins/autoRowHeaderSize/autoRowHeaderSize.ts
```diff
@@ -0,0 +1,1011 @@
...
+   */
+  #onCellsChange = (changes: unknown[][] | null) => {
+    if (!Array.isArray(changes) || changes.length === 0 || this.#cachedWidths === null) {
+      return;
+    }
+
+    // A sweep in progress has already read some of these rows, and it holds their old labels. It is
+    // cheaper to start it over than to work out which of them it still has to revisit.
+    if (this.#inProgress) {
+      this.clearCache();
```
demtario: `clearCache()` nulls `#cachedWidths`, so when the next draw restarts the sweep, `#commitWidths` takes the `previous === null` branch and its first commit uses `measured` (the first `syncLimit` rows only) directly instead of `Math.max`-ing against what is already on screen.

Failure scenario: a 200k-row grid whose longest label sits past the sync limit, edited while the idle sweep is still running. The header snaps back to its narrow first-paint width for a frame and then grows again as the chunks land. That is the flicker the JSDoc at line 666 ("while the sweep runs a level can only widen ... so the width never jumps back and forth") and the guide both promise cannot happen, and it is the reason #3850 is cited.

Restarting the sweep here should keep the current widths as the floor; only the final commit should be allowed to narrow.

handsontable/src/3rdparty/walkontable/src/viewport/workspaceSize.ts
```diff
@@ -13,6 +13,28 @@
...
+ * Returns `null` when the answer carries nothing usable (`0`, `NaN`, `undefined`), so the caller
+ * keeps the width it already worked out - the behavior the previous `|| this.rowHeaderWidth`
+ * fallback had.
+ *
+ * @param {number|Array} width The value the hook returned.
+ * @returns {number|null}
+ */
+function toTotalRowHeaderWidth(width: number | Array<number | null> | undefined): number | null {
+  const total = Array.isArray(width)
+    ? width.reduce<number>((sum, levelWidth) => sum + (levelWidth ?? 0), 0)
```
demtario: This sums *every* entry of the array and treats a missing one as `0`, while the other consumer of the same hook value, `columnUtils.ts:112-120`, iterates exactly `rowHeaders.length` entries and substitutes `defaultColumnWidth` (50) for a missing one.

So the two disagree whenever the array length differs from the rendered level count: an array shorter than `rowHeaders.length` under-reports the header block by 50px per missing level, and a longer one over-reports by the surplus entries. Either way the overlays and the viewport's own width calculation drift apart. Mirroring `ColumnUtils` here (iterate `getSetting('rowHeaders').length`, apply the same `defaultColumnWidth` fallback) makes a length mismatch degrade identically on both sides.

Separately: `toTotalRowHeaderWidth` has no test in the Walkontable suite. All four documented cases (array sum, number pass-through, `0`/`NaN` to `null`, `undefined`) are reachable from a user handler and none is asserted.

handsontable/src/plugins/nestedRows/nestedRows.ts
```diff
@@ -784,11 +784,21 @@ export class NestedRows extends BasePlugin {
...
+   * @param {number|number[]} rowHeaderWidth The initial row header width(s).
+   * @returns {number|number[]}
    */
-  #onModifyRowHeaderWidth = (rowHeaderWidth: number) => {
-    return Math.max(this.headersUI!.rowHeaderWidthCache ?? 0, rowHeaderWidth);
+  #onModifyRowHeaderWidth = (rowHeaderWidth: number | number[]) => {
+    const minimumWidth = this.headersUI!.rowHeaderWidthCache ?? 0;
+
+    if (Array.isArray(rowHeaderWidth)) {
+      return rowHeaderWidth.map(levelWidth => Math.max(minimumWidth, levelWidth));
```
demtario: The array branch applies `headersUI.rowHeaderWidthCache` as a minimum to *every* row header level, but that cache is the width the tree's own indented first level needs (`ui/headers.ts:199`: `50 + padding + 10 * depth + 25`). Levels 1..n come from `afterGetRowHeaderRenderers` and draw no indentation at all.

Failure scenario: `nestedRows` + `autoRowHeaderSize` + one extra header renderer on a deep tree. A narrow numbering column measured at 34px is inflated past 100px for no reason. Applying the minimum only to the level this plugin actually draws (index 0) and passing the rest through would fix it.

Also: nothing under `nestedRows/__tests__/` references `modifyRowHeaderWidth` or `rowHeaderWidthCache`. The `Math.max` producing `NaN` for an array is the entire reason this handler changed (and `??` in `ColumnUtils` does not catch `NaN`), so it is worth pinning with a test. Its absence is also why the hook-ordering problem in `autoRowHeaderSize.ts:341` has nothing to catch it.

handsontable/src/plugins/autoRowSize/autoRowSize.ts
```diff
@@ -196,10 +196,16 @@ export class AutoRowSize extends BasePlugin {
...
-  static get DEFAULT_SETTINGS(): { useHeaders: boolean; samplingRatio: number | null; allowSampleDuplicates: boolean } {
+   *
+   * There is deliberately no `useHeaders` entry here, even though {@link AutoColumnSize} declares one.
+   * That setting decides whether a header is rendered next to the samples being measured, and this
+   * plugin never asks that question: a row's height is measured with its row header always attached
+   * (see `GhostTable#createRow`, which keys off `hasRowHeaders()`), and the column header's own
+   * height is measured separately by `addColumnHeadersRow`. Declaring the setting anyway would offer
+   * a switch that changes nothing.
+   */
+  static get DEFAULT_SETTINGS(): { samplingRatio: number | null; allowSampleDuplicates: boolean } {
```
demtario, line 207: Dropping `useHeaders` from `DEFAULT_SETTINGS` is behaviorally inert (verified: nothing in `src/` reads it for `AutoRowSize`, `ghostTable.setSetting('useHeaders', ...)` is `AutoColumnSize`-only at `autoColumnSize.ts:395`, and `metaSchema.ts:486-492` never documented it for `autoRowSize`), and the JSDoc above explains it well.

It does still change what `hot.getPlugin('autoRowSize').getSetting()` returns with no argument, on a plugin unrelated to row header width. `.changelogs/7147.json` sets `"breaking": false` and mentions only the new plugin. Worth either a line in the changelog entry or splitting it out, so the change is discoverable by someone who hits it.

handsontable/src/plugins/autoRowSize/__tests__/rowHeightsInteraction.unit.ts
```diff
@@ -0,0 +1,73 @@
...
+    hot.destroy();
+  });
+});
+
+describe('AutoRowSize default settings', () => {
+  it('should not declare a `useHeaders` setting, because it never reads one', () => {
+    // AutoColumnSize does read `useHeaders` - it decides whether a column header is rendered beside
+    // the samples it measures. AutoRowSize has no such choice to offer, so declaring the key would
+    // hand the user a switch that changes nothing.
+    expect(Object.keys(AutoRowSize.DEFAULT_SETTINGS).sort()).toEqual([
```
demtario: Asserting the exact key set locks out ever adding `syncLimit` to `AutoRowSize.DEFAULT_SETTINGS` - and `syncLimit` *is* documented as an `autoRowSize` option at `metaSchema.ts:490`. The intent here is "no `useHeaders` key", so asserting that directly (`expect(AutoRowSize.DEFAULT_SETTINGS).not.toHaveProperty('useHeaders')`) proves the same thing without pinning the rest of the shape.

docs/content/guides/rows/row-header/row-header.md
```diff
@@ -125,15 +125,203 @@ The [`rowHeaders`](@/api/options.md#rowheaders) can also be populated using a fu
...
+
+**`samplingRatio`** exists because reading a label is cheap but laying one out is not. The plugin
+groups the labels by how many characters they have and renders only a few from each group. Raising
+the number measures more of them, which is slightly slower but harder to fool: in a proportional
+font `WWW` is wider than `iii`, so measuring more same-length labels makes it less likely that the
+widest one is missed. Lower it to do less work.
+
+**`allowSampleDuplicates`** decides what happens when two rows carry the same label. By default the
+label is measured once, because the same text normally renders to the same width. Set it to `true`
+when that is not true of your grid — a row header that is indented per row, as
```
demtario: Em dash. `docs/AGENTS.md:94`: "Hyphens (`-`) or double hyphens (`--`) to separate clauses. No en dashes or em dashes." The rest of the new prose in this file already uses ` - ` (lines 216, 245, 250, 262), so this is also internally inconsistent.

docs/content/guides/rows/row-header/row-header.md
```diff
@@ -125,15 +125,203 @@ The [`rowHeaders`](@/api/options.md#rowheaders) can also be populated using a fu
...
+[`nestedRows`](@/api/options.md#nestedrows) does, draws the same label at a different width
+depending on how deep the row sits, so measuring only the first one would come out too narrow.
+
+Raise `samplingRatio` along with it. Labels are grouped by length and only `samplingRatio` of each
+group get measured, so at the default of `3` a fourth row carrying the same label is still left out
+— which on a deep tree can be the widest one of them.
```
demtario: Em dash again, same rule (`docs/AGENTS.md:94`). Replace with ` - ` to match the surrounding prose.

handsontable/src/plugins/autoRowHeaderSize/autoRowHeaderSize.ts
budnix: You're right, and the AutoColumnSize precedent settles it - pinned at -10 in e5a774b85f, with a unit test that fails without it.

handsontable/src/plugins/autoRowHeaderSize/autoRowHeaderSize.ts
budnix: Good catch on both - the cache key carries the row header level count now, and the array comes back as a copy (e5a774b85f).

handsontable/src/plugins/autoRowHeaderSize/autoRowHeaderSize.ts
budnix: Yep, my mistake - that contradicted the JSDoc right above it. Invalidation keeps the current widths as the floor now, and only the final commit may narrow (e5a774b85f).

handsontable/src/3rdparty/walkontable/src/viewport/workspaceSize.ts
budnix: Good spot - it mirrors ColumnUtils now, and toTotalRowHeaderWidth finally has its own test covering all four cases (e5a774b85f).

handsontable/src/plugins/nestedRows/nestedRows.ts
budnix: Fair point, that minimum belongs to the level this plugin draws - level 0 only in e5a774b85f, with the test you asked for.

handsontable/src/plugins/autoRowSize/autoRowSize.ts
budnix: Good point - it has its own changelog entry now, .changelogs/13239.json in e5a774b85f.

handsontable/src/plugins/autoRowSize/__tests__/rowHeightsInteraction.unit.ts
budnix: Agreed - narrowed to a not.toHaveProperty check in e5a774b85f.

docs/content/guides/rows/row-header/row-header.md
budnix: Nice catch - both em dashes are hyphens now in e5a774b85f.

docs/content/guides/rows/row-header/row-header.md
budnix: That one too, fixed in e5a774b85f.

demtario (approved):

Issue #11240, opened 2024-10-29
Can't access any internal types


jonahallibone:
### Describe the bug

I am unable to type my component properly as I cannot import `CellMeta`, `GridSettings` etc from react or base libraries. 

This is how I've resorted to accessing the types:

```tsx
import type HandsonTypes from "handsontable/base";
...
type CellMeta = HandsonTypes.CellMeta;
```

I would be great to be able to import `import type { CellMeta } from "handsontable/base"` or `"handsontable"` directly

### Handsontable version

14.6.1

### Your environment

MacOS

adrianszymanski89:
Hi @jonahallibone 

Thank you for the suggestions. We are constantly improving the types definitions. I will report this idea internally and update you once I have any updates.

jonahallibone:
> Hi @jonahallibone
> 
> Thank you for the suggestions. We are constantly improving the types definitions. I will report this idea internally and update you once I have any updates.

Great! For a more detail example, take this function I have written below:

```ts
const getCells: HandsonTypes.GridSettings["cells"] = (...props) => {
    const [x, y] = props;

    if (data?.[x]?.[y]) {
      const isReadOnly = data[x][y].editable === false;

      return {
        readOnly: isReadOnly,
      };
    }

    return {};
  };

  const memData = useMemo(
    () => data.map((row) => row.map((cell) => cell.value)),
    []
  );
  ```
I see from the type defs you're exporting the entire namespace, and i'm unsure if there's a way to export the types without the namespace as a flat object, but I think it would be a much nicer experience.

Accessing the `cells` prop type via `HandsonTypes.GridSettings["cells"]` isn't ideal for a couple reasons:

Firstly, aesthetics. It's not a huge deal, but needing to import the namespace, then dot-access the object I want to use followed by needing to index on that object to retrieve the property type is sort of unintuitive, and frankly doesn't look super nice. I could remap them all to named types at the top of my file, but at that point I'd hope the library would export them as such.

Secondly, intellisense. I had no way of knowing how to access these types with clicking through 3 files of type definitions and hoping that my import would work. Generally, if a type is used in the type definition of function or object, I'd expect it to be available as a direct import from the library. 

As an example:
<img width="808" alt="Screenshot 2024-10-29 at 12 04 30 PM" src="https://github.com/user-attachments/assets/bfc89ff1-1ee1-4435-88ce-07c4a886d76c">

[...]

PR #13240, opened 2026-08-25
DEV-2620: Export SanitizerContext for the sanitizer option

demtario:
### Context

The PR exports `SanitizerContext`, a type naming the write surfaces the `sanitizer` option receives.

That second argument names the surface (`'header'`, `'password'`, `'contextMenu'`, `'selectEditor'`, `'dialog'`, `'notification'`, `'CopyPaste.paste'`, `'CopyPaste.paste.sourceData'`). Since #13236 the documentation tells readers to branch on it and lists every value, so we publish the contract in prose and withhold it from the types. `handsontable/AGENTS.md` lists "The public API must give good code completion in IDEs and AI assistants" as an API design rule.

Users opt in by annotating their own parameter:

```ts
import type { SanitizerContext } from 'handsontable';

const settings = {
  sanitizer: (content: string, source: SanitizerContext) =>
    source === 'CopyPaste.paste' ? strict(content) : loose(content),
};
```

The contract was also declared four times over: `core/settings.ts`, `utils/sanitizer.ts`, the `fastInnerHTML` parameter, and the `@type` tag generating the option reference. It is declared once now, and the others reference it.

**This PR targets `feature/DEV-2600_Sanitizer-coverage` (#13236), not `develop`**, because it builds on the contexts that branch introduces.

### Why the option's own signature is left alone

The task proposed naming the parameter on the option itself. That cannot be done without a build break, on three separate axes. All three were verified with `tsc` against the emitted `.d.ts`, not reasoned about:

| Shape | Breaks |
|---|---|
| `sanitizer?: (html, context: SanitizerContext, ...args) => string` | `strictFunctionTypes` checks a property's parameters contravariantly, rejecting a user callback that annotates the parameter more narrowly: `(html, source: 'header') => …` fails TS2322 |
| `sanitizer?(html, context: SanitizerContext, ...args): string` | Method syntax fixes the above, but a declared parameter ahead of a rest parameter raises the option's minimum **call** arity, so `hot.getSettings().sanitizer?.(html)` fails TS2555 |
| `sanitizer?(html, context?: SanitizerContext, ...args): string` | Fixes the arity, but types the parameter `SanitizerContext \| undefined`, so any body using it as a definite string fails (TS2345 when passed on, TS18048 when dereferenced) |

Overloads, a labeled-tuple rest (`...args: [context?: SanitizerContext, ...rest: any[]]`) and a rest union (`[] \| [context: SanitizerContext, ...]`) were also tested against the same matrix and fail worse than any of the above.

Assignability and callability are separate checks, which is why the arity break survived a green suite in the first iteration of this PR. It was found in review and is fixed here.

So the option's declaration is now **byte-identical** to what shipped, and the contract is published as an opt-in type instead. The baseline is 18.0.0, where `...args: any[]` arrived with the TypeScript conversion; 17.0.0 shipped `(content: string, source: 'innerHTML' | 'CopyPaste.paste') => string`, two required narrow parameters, against which neither of these breaks existed. That is zero-break by construction. The cost is that completion requires the annotation; the strict inline signature remains available at a major boundary.

`SanitizerFn` is deliberately **not** exported from the entry points: with the second parameter absorbed by `...args: any[]` it conveys no context, so advertising it would mislead. It stays internal for the plugins that hold a resolved sanitizer.

`'innerHTML'` is absent from the literals. No grid write surface emits it and the new 18.1 to 18.2 migration guide tells readers to delete that branch, but it stays assignable via `(string & {})` for `Handsontable.dom.fastInnerHTML()` callers, who reach it through the parameter's default.

There is no runtime change: every added import is `import type`.

### How has this been tested?

`src/__tests__/core/sanitizer.types.ts` compiles against the emitted, downleveled `.d.ts` (the types suite maps `handsontable` to `tmp/`), so these assertions cover the published types rather than source.

It pins both axes rather than just the one that was obvious:

- **Assignability** — one-, two- and three-parameter declarations, an inferred `context`, a widely typed one, a narrowly typed one, and a body using the parameter as a definite string.
- **Call arity** — `hot.getSettings().sanitizer?.(html)` with one argument. This is the case that was missing before and that catches the regression this PR reverts.
- **Through the wrappers** — the `Omit<RemoveIndexSignature<GridSettings>, …>` chain React and Angular build their props from.
- **The union's literals** — by exhaustiveness, not by assignment. `(string & {})` accepts every string, so `const x: SanitizerContext = 'header'` would pass against an empty or misspelled union and prove nothing. A `Record<KnownContext<SanitizerContext>, true>` fails both when a literal is added without a case (TS2741) and when one is dropped or misspelled (TS2353).

Each of these was verified to fail for the right reason by mutating the production type and confirming the expected error, then restoring.

```bash
npm run test:e2e --prefix handsontable -- --testPathPattern='sanitizer|nestedHeaders'
```

All green: 3743 unit, 816 Walkontable, React 75 / Vue 33 / Angular 291, 0 lint errors. The E2E figure is a subset, not the full suite: 232 specs from `--testPathPattern='sanitizer|nestedHeaders'`.

### Merge order

This must land **after** #13236. `handsontable/src/utils/sanitizer.ts`, the security guide's surface table, and the `metaSchema.ts` block naming `'CopyPaste.paste.sourceData'` do not exist on `develop`, so merging ahead of it would conflict or silently drop that content. Only the `settings.types.ts` hunk is order-safe.

### Types of changes

### Related issue(s):

1. DEV-2620

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2620

---

> [!NOTE]
> **Low Risk**
> Type-only and documentation changes with no runtime behavior change; the deliberate non-narrowing of the public `sanitizer` option signature preserves backward compatibility for TypeScript consumers.
> 
> **Overview**
> Adds and exports **`SanitizerContext`**, a string union of the grid write surfaces passed to the **`sanitizer`** option (`header`, `password`, `contextMenu`, etc.), widened with **`| (string & {})`** so shared sanitizers and `fastInnerHTML` callers still compile. Users get IDE completion by annotating their callback’s second parameter; the **`GridSettings.sanitizer`** signature stays **`(html: string, ...args: any[]) => string`** so upgrading does not break assignability or **`hot.getSettings().sanitizer?.(html)`** call arity.
> 
> **`SanitizerFn`** is derived from the option in **`core/settings`** and re-exported only internally; **`fastInnerHTML`** / **`sanitizeHTML`** JSDoc and parameters now reference **`SanitizerContext`**. Documentation (security guide, TypeScript types, **`metaSchema`**) and **`AGENTS.md`** / **`.ai/BREAKING-CHANGES.md`** document the opt-in typing strategy and union exhaustiveness tests in **`sanitizer.types.ts`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 64cbc7913e5e5e98f6b653724078211691eba3b5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2600 [GH-13229] Clarify/harden XSS-sanitization gap for NestedHeaders `label` and other header content](https://app.clickup.com/t/9015210959/DEV-2600)
Task linked: [DEV-2620 Type the `sanitizer` option's context argument instead of `any`](https://app.clickup.com/t/9015210959/DEV-2620)

budnix (commented):
Read through the whole change. The decision to publish `SanitizerContext` instead of touching the option's signature is well argued, and the AGENTS.md write-up on the three break axes is genuinely useful.

A few things below. The two that matter most are the `Events` doc comment losing its owner, and `SanitizerFn` being able to turn into `any` without anyone noticing. The rest are smaller gaps around the new type's reach and the docs.

handsontable/src/core/settings.ts
```diff
@@ -633,6 +675,15 @@ type HookKey = {
  * Map of all Handsontable hook names to their typed callback signatures.
  * Use with addHook/addHookOnce/removeHook for IDE autocomplete and compile-time safety.
  */
+/**
```
budnix, line 680: The `Events` doc comment sits directly above this, so inserting the block here hands that comment to `SanitizerFn` instead. In the emitted `tmp/core/settings.d.ts` the "Map of all Handsontable hook names" text now precedes this JSDoc, and `export type Events` follows with nothing above it. `Events` is public and listed in the TypeScript types guide, so it loses its hover text and renders undocumented in Typedoc. Moving the `SanitizerFn` block above the `Events` comment sorts it out.

handsontable/src/core/settings.ts
```diff
@@ -633,6 +675,15 @@ type HookKey = {
...
+ * The shape of a configured `sanitizer`, derived from the option so the two cannot drift apart.
+ *
+ * Internal. It is not re-exported from the package entry points: with the option's second parameter
+ * absorbed by `...args: any[]`, annotating with this type conveys no context, so `SanitizerContext`
+ * is what users are given instead.
+ */
+export type SanitizerFn = NonNullable<GridSettings['sanitizer']>;
```
budnix: `GridSettings` carries a `[key: string]: any` index signature, so this lookup does not fail when the key is gone – it quietly resolves to `any`. I renamed the option locally to check: with the current form tsc reports zero errors anywhere, and every internal user (`getSanitizer`, `fastInnerHTML`, the dialog and notification UIs) loses its checking with nothing to show for it. The comment right above says the two cannot drift apart, and this is the one direction where they can. Stripping the index signature first turns that same rename into a TS2339 right on this line.

```suggestion
export type SanitizerFn = NonNullable<RemoveIndexSignature<GridSettings>['sanitizer']>;
```

`RemoveIndexSignature` is already imported at the top of the file, and `typecheck` plus `test:types` both stay green with it.

handsontable/src/helpers/dom/element.ts
```diff
@@ -553,18 +554,18 @@ const defaultSanitizerWarnScope = {};
...
+ * @param {SanitizerContext} [context] The sanitization context passed as the second argument to a custom sanitizer function, and
  * used in the missing-sanitizer warning to identify the write surface.
  * @param {object} [scope] Object the "warn once" state is bound to (`hot.rootElement` at every call site
  * inside the grid), so the warning is shown at most once per Handsontable instance.
  */
 export function fastInnerHTML(
   element: HTMLElement, content: string,
-  sanitizer: boolean | ((html: string, context: string) => string) = true,
-  context = 'innerHTML',
+  sanitizer: boolean | SanitizerFn = true,
```
budnix: The old type was `(html: string, context: string) => string`, and `SanitizerFn` is `(html: string, ...args: any[]) => string` – so this parameter got wider here, not narrower. `Handsontable.dom.fastInnerHTML` is public, so a caller can now pass `(html: string, n: number) => string`, get it accepted, and receive a string in `n` at runtime. The JSDoc two lines up still promises `function(string, SanitizerContext): string`. Spelling the shape out keeps that promise and still picks up the `SanitizerContext` narrowing you wanted:

```suggestion
  sanitizer: boolean | ((html: string, context: SanitizerContext) => string) = true,
```

handsontable/src/__tests__/core/sanitizer.types.ts
```diff
@@ -0,0 +1,127 @@
...
+ * erases the whole union, because each literal is itself assignable to `string & {}`.) The `Record`
+ * then fails both ways: a literal added to the union without a case here is TS2741, and a literal
+ * dropped from or misspelled in the union is TS2353 on the orphaned key.
+ *
+ * Update this map when the grid starts emitting a new context, and update the `sanitizer` JSDoc in
+ * `metaSchema.ts` plus the surface table in the security guide in the same change.
+ */
+type KnownContext<T> = T extends string ? (string extends T ? never : T) : never;
+
+const everyKnownContext: Record<KnownContext<SanitizerContext>, true> = {
```
budnix, line 133: This catches a literal being added, dropped, or misspelled, but not the union collapsing altogether. If someone "simplifies" `SanitizerContext` to plain `string`, `KnownContext` yields `never`, `Record<never, true>` becomes `{}`, and this eight-key object still assigns cleanly – the file stays green while the completion the type exists for is gone. One extra line pins it, since `'header'` is not assignable to `never`:

```suggestion
const pinnedLiteral: KnownContext<SanitizerContext> = 'header';

const everyKnownContext: Record<KnownContext<SanitizerContext>, true> = {
```

handsontable/src/base.ts
```diff
@@ -456,7 +457,7 @@ export {
 
 // Named type exports for user-facing API (mirrors src/index.ts)
 // Note: CellCoords and CellRange are already exported as runtime values above.
-export type { GridSettings, Events } from './core/settings';
+export type { GridSettings, Events, SanitizerContext } from './core/settings';
```
budnix, line 460: `typeImports.types.ts` is the #11240 regression test for exactly this – that the named types resolve from both `handsontable` and `handsontable/base`. `SanitizerContext` did not get added to its lists, and the new `sanitizer.types.ts` only imports it from `'handsontable'` and from the `Handsontable.` namespace. So if this line loses the name in a later edit, `handsontable/base` users break and the suite stays green. Worth adding it to both import lists over there.

handsontable/src/index.ts
```diff
@@ -212,7 +212,7 @@ Handsontable.themes.reinitTheme = reinitTheme;
 export { IndexMapper } from './translations';
 
 // Named type exports for user-facing API — direct from canonical sources
-export type { GridSettings, Events } from './core/settings';
+export type { GridSettings, Events, SanitizerContext } from './core/settings';
```
budnix, line 215: `src/settings.ts` still re-exports only `GridSettings` and `SourceDataValidatorFn` from `./core/settings`. Its own header calls it the place external consumers import settings types from, `handsontable/settings` is a real published subpath, and the Angular wrapper already imports from it. Right now `import type { GridSettings, SanitizerContext } from 'handsontable/settings'` gives TS2305 on the second name while the first one resolves, which is a confusing spot for a user to land in. Adding it there too would keep the three entry points in step.

handsontable/src/utils/sanitizer.ts
```diff
@@ -1,12 +1,17 @@
...
 import { warnOnce } from '../helpers/console';
 
 /**
- * Signature of the `sanitizer` grid option. The second argument names the write surface,
- * so a sanitizer can apply different rules per context (for example, stricter for paste).
+ * Re-exported for the plugins that hold a resolved sanitizer (`dialog`, `notification`).
+ *
+ * `SanitizerFn` stays internal: the published `exports` map (generated by
+ * `scripts/prepare-package-for-publish.mjs`, not the source manifest) has no `./utils/*` entry and no
+ * `./*` wildcard, so this module is unreachable from an installed package. `SanitizerContext` is the
```
budnix: The `exports` map only blocks subpath resolution under `node16` and `bundler`. `tmp/utils/sanitizer.d.ts` is built, `package.json` has no `files` field trimming it, and classic `moduleResolution: node` ignores `exports` entirely – so `import type { SanitizerFn } from 'handsontable/utils/sanitizer'` does resolve for those consumers. Calling the module unreachable makes a future rename here look free when it isn't. Separately, only `dialog/ui.ts` and `notification/ui.ts` import from this file and both take `SanitizerFn` alone, so the `SanitizerContext` half of the re-export has no user.

handsontable/src/dataMap/metaManager/metaSchema.ts
```diff
@@ -7332,6 +7332,10 @@ export default (): Record<string, unknown> => {
      * `'CopyPaste.paste'`, `'CopyPaste.paste.sourceData'`), so you can apply different rules per source.
      * It must return a string that is safe to assign to `innerHTML`.
      *
+     * In TypeScript, annotate that parameter with the exported `SanitizerContext` type
```
budnix, line 7335: The `@type` tag about twenty lines below still reads `function(string, string): string`, and that tag is what renders in the type column at `api/options.md#sanitizer`. So this new prose tells users to annotate with `SanitizerContext` while the signature printed right under it says `string`. The sibling JSDoc in `helpers/dom/element.ts` was updated to `function(string, SanitizerContext): string` in this same diff, so the two published docs now disagree about the same parameter. Would be good to bring the tag along.

handsontable/src/__tests__/core/sanitizer.types.ts
```diff
@@ -0,0 +1,127 @@
...
+ * `(string & {})` member accepts every string, which is the deliberate trade documented below.
+ */
+const optIn: GridSettings = {
+  sanitizer: (content: string, source: SanitizerContext) =>
+    (source === 'CopyPaste.paste' ? content.trim() : content),
+};
+
+/**
+ * Every declaration that compiled before this change must still compile. The option is public and
+ * shipped in 17.0.0, so narrowing it in any direction is a build break on upgrade.
```
Haxikowy: 17.0.0 shipped `sanitizer?: (content: string, source: 'innerHTML' | 'CopyPaste.paste') => string` (`17.0.0:handsontable/types/settings.d.ts:211`). The `...args: any[]` shape arrived in 18.0.0 with the TS conversion (18e73ccca).

Against 17.0.0's two required narrow parameters neither reverted break existed, so the base being protected is 18.0.0's. Same version in the PR body.

handsontable/src/core/settings.ts
```diff
@@ -38,6 +38,40 @@ export type SourceDataValidatorFn = {
...
+ * ```ts
+ * import type { SanitizerContext } from 'handsontable';
+ *
+ * const settings = {
+ *   sanitizer: (content: string, source: SanitizerContext) =>
+ *     source === 'CopyPaste.paste' ? strict(content) : loose(content),
+ * };
+ * ```
+ *
+ * The listed values are the ones the grid emits. The `(string & {})` member keeps arbitrary strings
```
Haxikowy: `autocompleteEditor.ts:205` passes a ninth context, `'html'`. No user sanitizer sees it (`sanitizer === false` there), but the sentence reads as exhaustive.

Separately, annotating `fastInnerHTML`'s `context` with this union rejects only non-strings: I renamed `passwordRenderer.ts:45`'s `'password'` to `'passwrd'` and it compiles clean, while `42` errors (TS2345). A narrow 8-literal alias for the `fastInnerHTML`/`sanitizeHTML` parameters would catch an internal typo and leave the public type open.

handsontable/src/__tests__/core/sanitizer.types.ts
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * The two assignments above cannot pin the union's contents: `(string & {})` accepts every string,
+ * so they would pass against an empty union or a misspelled literal just as well. Completion on the
+ * eight literals is the whole point of the type, so pin them by exhaustiveness instead.
+ *
+ * `KnownContext` drops the `(string & {})` member by discarding the constituent that every string is
+ * assignable to, leaving the literals. (`Exclude<SanitizerContext, string & {}>` cannot do this - it
+ * erases the whole union, because each literal is itself assignable to `string & {}`.) The `Record`
+ * then fails both ways: a literal added to the union without a case here is TS2741, and a literal
+ * dropped from or misspelled in the union is TS2353 on the orphaned key.
```
Haxikowy: A misspelled literal is TS2561, not TS2353:

```
error TS2561: Object literal may only specify known properties, but 'password' does not exist in type 'Record<…"passwrd"…>'
```

TS2353 is the dropped case, TS2741 the added one, both as claimed. `AGENTS.md:128` repeats the pairing. The body's axis (c) is TS2345 at :43, not TS18048.

handsontable/AGENTS.md
```diff
@@ -123,6 +123,9 @@ Gotcha: Filters `conditionCollection` uses physical indexes, `getDataAtCol()` us
...
...so `sanitizer?: (html, context: C) => string` rejects a user callback annotating the parameter more narrowly (`(html, source: 'header') => …`) while `sanitizer?(html, context: C): string` accepts it. (2) *Call arity* — a declared parameter ahead of a rest parameter raises the option's minimum **call** arity, so a user who reads the option back out and invokes it, `hot.getSettings().sanitizer?.(html)`, gets TS2555. Assignability and callability are separate checks: a suite that only assigns will not catch this. (3) *Optionality* — declaring it `context?: C` fixes the arity break but types the parameter `C | undefined`, so any body using it as a definite string breaks instead. Overloads, a labeled-tuple rest (`...args: [context?: C, ...rest: any[]]`) and a rest union (`[] | [context: C, ...]`) were all tested and fail worse than either. The way out is to **not touch the option's signature at all**: export the union as a named type and document it as an annotation the user opts into on their own parameter. That is zero-break by construction, and it is what `sanitizer` ships. Reach for method syntax only when declaring a *new* callback option, where there is no installed base to break.
```
Haxikowy: Three fixes, on rules the next agent will cite instead of re-running.

"No signature clears all three" is absolute over an unbounded space, and the rule concedes method syntax clears (1). State what was tested: property, method, optional parameter, overloads, labeled-tuple rest, rest union.

Axis (1) needs "under `strictFunctionTypes`" — without it a consumer sees no break. TS2555 fires under property syntax too, so (2) is not method-specific.

Scope: this rule is monorepo-wide and elaborates the "Changing API signatures" row of `.ai/BREAKING-CHANGES.md`, which root `CLAUDE.md` names as the full rules. Someone narrowing `beforePaste` reads that file, not this one. Rules 2 and 3 are core-specific and fine here.

docs/content/guides/security/security/security.md
```diff
@@ -69,6 +69,19 @@ Use the [`sanitizer`](@/api/options.md#sanitizer) option to provide a sanitizer
 
 A configured sanitizer runs on the HTML that Handsontable writes on your behalf. The second argument names the write surface, so you can apply different rules to each one.
 
+In TypeScript, annotate the second parameter with [`SanitizerContext`](@/guides/tools-and-building/typescript-types/typescript-types.md) to get completion on these values:
```
Haxikowy: "These values" points forward, past a code block, at the table three paragraphs down — whose own introducing sentence is now upstream of this insertion. Moving the block below the table and saying "the values above" fixes both.

The gap that matters: this publishes only the favourable half of `(string & {})`. `sanitizer.types.ts:15-18` states the other half, that the annotation cannot reject a misspelled comparison. A reader here is writing per-surface security rules, and `source === 'contextmenu'` compiles into a dead branch. One sentence covers it.

handsontable/src/core/settings.ts
```diff
@@ -38,6 +38,40 @@ export type SourceDataValidatorFn = {
...
+export type SanitizerContext =
+  | 'header'
+  | 'password'
+  | 'contextMenu'
+  | 'selectEditor'
+  | 'dialog'
+  | 'notification'
+  | 'CopyPaste.paste'
+  | 'CopyPaste.paste.sourceData'
+  // eslint-disable-next-line @typescript-eslint/no-empty-object-type
```
Haxikowy: Neither rule is enabled. `handsontable/.eslintrc.js` lists `@typescript-eslint/*` rules individually with no `recommended` extends and the root extends `airbnb-base` only; `no-empty-object-type` is in neither, nor is `no-unused-vars` (disabled at `sanitizer.types.ts:63`, whose neighbours at :55 and :59 carry the same unused parameter with no directive).

Unused-directive reporting is off under eslintrc, so they don't fail lint — they assert a constraint that isn't there. The `no-explicit-any` disable at :286 is real.

handsontable/src/__tests__/core/sanitizer.types.ts
```diff
@@ -0,0 +1,127 @@
...
+};
+
+/**
+ * Reading the option back out and calling it. This is the axis that a named second parameter
+ * regresses: declaring one raises the minimum call arity from one to two, and the single-argument
+ * call below stops compiling with TS2555. Nothing else in this file catches that, because
+ * assignability and callability are separate checks.
+ */
+const calledWithOneArgument = hot.getSettings().sanitizer?.('<b>x</b>');
+const calledWithTwoArguments = hot.getSettings().sanitizer?.('<b>x</b>', 'header');
```
Haxikowy, line 80: Guards nothing. Across nine mutations of the declaration (property and method syntax × named and optional parameter, an added, misspelled and dropped literal, dropped `(string & {})`, collapse to `string`), lines 19, 41/43, 49, 53, 62, 73, 85/89 and 116-121 each fire at least once; 28, 32, 74, 97 and 98 never do.

The second argument is unchecked under `...args: any[]`, so :73 already covers this. Drop it, or assert the return type.

Haxikowy (commented):
**Merge order is a gate.** `utils/sanitizer.ts` (the whole file), the security guide's lead sentence and table, and the `metaSchema.ts` block mentioning `'CopyPaste.paste.sourceData'` do not exist on develop. Merging ahead of #13236 conflicts or silently drops its content. Only the `settings.types.ts` hunk is order-safe.

**DEV-2620.** The task offered a named parameter now or the strict signature at the next major; this ships neither. Your reasoning is sound and I would take it, but someone writing `sanitizer: (content, source) => …` still gets `any` unless they already know the type exists. So: put the arity evidence on the task, where the original proposal still stands as "verified to compile clean" on an assignment-only check, and file the major-boundary follow-up.

**"232 E2E"** came from `--testPathPattern='sanitizer|nestedHeaders'` but sits alongside the full unit and Walkontable runs. Worth marking as the subset.

handsontable/src/core/settings.ts
demtario: Fixed in d53dbe5. Confirmed exactly as you describe: the emitted `tmp/core/settings.d.ts` had the hook-map text sitting above `SanitizerFn`, with `export type Events` following bare. Moved the block above the `Events` comment.

handsontable/src/core/settings.ts
demtario: Fixed in d53dbe5, took the suggestion verbatim.

Reproduced before applying it: renaming the option to `sanitizerRENAMED` left the old form resolving to `any` with zero errors anywhere, and `const check: number = someSanitizerFn` compiled clean. With `RemoveIndexSignature` the same rename is TS2339 on that line. Re-verified after the change by mutation.

You are right that the comment above it was promising something the code did not deliver, and this was the one direction that could drift. The comment now says why the wrapper is load-bearing so it does not get simplified away later.

handsontable/src/helpers/dom/element.ts
demtario: Fixed in d53dbe5, took the suggestion.

I had reasoned about this as a loosening and stopped there, which missed that the JSDoc two lines up promises a shape the signature no longer enforced, on a public helper. `SanitizerFn` is no longer imported in that file.

handsontable/src/__tests__/core/sanitizer.types.ts
demtario: Fixed in d53dbe5, took the suggestion.

Verified both halves: with the union collapsed to plain `string`, the eight-key `Record` assigns cleanly against `Record<never, true>` and the file stays green, while the added line fails TS2322 `Type "header" is not assignable to type "never"`. Re-ran the collapse mutation against the built `.d.ts` after the change and it now fires.

handsontable/src/base.ts
demtario: Fixed in d53dbe5. Added `SanitizerContext` to both import lists in `typeImports.types.ts`, with an assignment on each side so the names are actually used rather than just imported.

handsontable/src/index.ts
demtario: Fixed in d53dbe5. Added it to the `./core/settings` re-export in `src/settings.ts`.

Confirmed `./settings` is in the generated exports map in `tmp/package.json`, so the TS2305-next-to-a-resolving-`GridSettings` case you describe was real.

handsontable/src/utils/sanitizer.ts
demtario: You are right, and the comment was mine to begin with. Fixed in d53dbe5.

I checked it properly this time: copied a built `tmp/` into a scratch `node_modules/handsontable` and compiled `import type { SanitizerFn } from 'handsontable/utils/sanitizer'` under all three resolution modes. TS2307 under `node16` and `bundler`, resolves clean under classic `node`. `tmp/utils/sanitizer.d.ts` and `.js` are both emitted and there is no `files` field.

The comment now says the opposite of what it did: treat the module as reachable, and renaming or narrowing what it exports can break a consumer. Also dropped the `SanitizerContext` half of the re-export, which as you note had no user.

handsontable/src/dataMap/metaManager/metaSchema.ts
demtario: Fixed in d53dbe5. The tag now reads `function(string, SanitizerContext): string`, matching the sibling in `element.ts` and the prose that points at it.

Context on why it was inconsistent: I had set the tag to `SanitizerContext` first, then reverted it after a reviewer pointed out the name is TypeScript-only and renders unlinked for JavaScript readers. Reverting the tag while leaving the prose and the `element.ts` sibling in place produced exactly the disagreement you found. Making the two published tags agree is the better trade, since a reader of one is likely to hit the other.

handsontable/src/__tests__/core/sanitizer.types.ts
demtario: Correct, fixed in d53dbe5. Verified with `git show 17.0.0:handsontable/types/settings.d.ts` and `git show 18.0.0:handsontable/src/core/settings.ts` - two required narrow parameters in 17.0.0, `...args: any[]` from 18.0.0.

The comment now states 18.0.0 as the baseline and says explicitly that neither guarded break existed against 17.0.0. Corrected in the PR body too.

handsontable/src/__tests__/core/sanitizer.types.ts
demtario: Correct on all three, fixed in d53dbe5.

Re-ran each mutation in isolation to get the codes right rather than take them from memory: misspelled TS2561, dropped TS2353, added TS2741. The comment and the `AGENTS.md` line now list all three separately.

Axis (c) in the PR body now reads TS2345 when the value is passed on and TS18048 when it is dereferenced - the assertion in the file is the `takesString(source)` one, so TS2345 is the code that applies there.

handsontable/src/core/settings.ts
demtario: Correct, both removed in d53dbe5.

Confirmed neither is configured: `handsontable/.eslintrc.js` extends only `../.eslintrc.js`, which extends `airbnb-base`, and neither `no-empty-object-type` nor `@typescript-eslint/no-unused-vars` appears in either file. No `reportUnusedDisableDirectives` either. Your point about the neighbours at :55 and :59 carrying the same unused parameter with no directive was the giveaway - lint stays at 0 errors with both directives gone, which confirms they were inert.

docs/content/guides/security/security/security.md
demtario: Both fixed in d53dbe5.

Moved the block below the table, so "the values above" now has its referent and the table keeps its own introducing sentence.

The second point is the better catch. It now states the unfavourable half directly: the type accepts any other string, cannot reject a wrong one, and `source === 'contextmenu'` compiles into a branch that never runs. Given the page is about writing per-surface security rules, a silently dead branch is the specific failure worth naming. Added the same caveat to the `AGENTS.md` bullet about `(string & {})`, since it was publishing only the favourable half too.

handsontable/AGENTS.md
demtario: All three taken, in d53dbe5, plus the scope point.

The rule moved to `.ai/BREAKING-CHANGES.md` as a section under the "Changing API signatures" row. You are right that someone narrowing `beforePaste` or a wrapper prop reads that file and would never have found this. `AGENTS.md` keeps a one-line pointer with the three error codes; rules 2 and 3 stayed in core.

It no longer says no signature clears all three. It now names the six shapes actually measured - property, method, optional parameter, overloads, labeled-tuple rest, rest union - and states that this is the set that was measured, not a proof that none can exist.

Axis (1) now carries "under `strictFunctionTypes`", and the arity axis says explicitly that it hits property and method syntax alike rather than being method-specific.

handsontable/src/core/settings.ts
demtario: Two separate things here; took the first, and the second runs into a constraint.

The `'html'` context: fixed in d53dbe5. The sentence was reading as exhaustive over `fastInnerHTML` calls when it is exhaustive over what reaches a *user sanitizer*. It now names both `'html'` and `'innerHTML'` as the two that do not, with the reason.

The narrow internal alias: I confirmed your typo case - `'passwrd'` compiles, `42` is TS2345 - so the union buys nothing against an internal misspelling. But `fastInnerHTML` is public (`Handsontable.dom.fastInnerHTML`), and its `context` parameter is what most of those call sites go through. Narrowing it to the eight literals rejects a public caller labelling their own surface: `fastInnerHTML(el, html, mySanitizer, 'myCustomRenderer')` becomes TS2345. Verified. That is a break on the public helper, which is the thing this PR exists to avoid.

That leaves `sanitizeHTML` in `utils/sanitizer.ts`, which is internal-ish and has two call sites, so a narrow alias there would catch a typo in two of the nine. Happy to add it if you think it earns its place, but flagging that it cannot cover the `fastInnerHTML` sites without the break above, so the typo hole stays open for most of them either way.

handsontable/src/__tests__/core/sanitizer.types.ts
demtario: Kept both, with evidence - I think the matrix misses a case on each.

`:97-98` (`unknownContext` / `arbitraryContext`) do fire. They are what pins the `(string & {})` member: drop it from the union and both fail TS2322 (`Type "innerHTML" is not assignable to type SanitizerContext`). Your list includes "dropped `(string & {})`" among the nine, so I think that mutation was applied somewhere the two lines could not see it - they are the only assertions in the file that detect it.

`:74` (`calledWithTwoArguments`) fires on a mutation outside the nine: dropping the rest parameter entirely, `(html: string) => string`. Then the two-argument call is TS2554 `Expected 1 arguments, but got 2` while the single-argument call at `:73` passes. That is a plausible edit given `...args: any[]` is the part everyone wants to delete, so I would rather keep it.

You are right about `:28` and `:32` - no change to the second parameter can trip them, since `...args: any[]` leaves it unchecked. I kept them but they now carry a comment saying what they actually guard: the first parameter and the return type, which DEV-2617 puts in play if `TrustedHTML` lands on this option. If you would rather they were assertions than documentation, I will drop them.

All of this is in d53dbe5, along with a comment on `:73-74` explaining why the pair is not redundant.

Haxikowy: You are right on both, and my matrix was wrong for a reason worth recording.

`:74` — reproduced your case. With the option as `(html: string) => string` the two-argument call is TS2554 while `:73` still passes, so it catches a mutation I never applied. Keep it.

`:97-98` — dropping `(string & {})` fails `build:types` first, on `element.ts:592` `'innerHTML'` and `autocompleteEditor.ts:205` `'html'`, so `test:types` never ran and my matrix recorded them as silent. They do fire once the production compile is fixed, and they are the only assertions in the file that see that mutation.

`:28`/`:32` — leave them as documentation. Nothing about the second parameter can trip them, so as assertions they would be dead weight, but the first-parameter and return-type guard is worth keeping ahead of DEV-2617.

handsontable/src/core/settings.ts
Haxikowy: Agreed on the alias, not worth adding. Your `fastInnerHTML(el, html, mySanitizer, 'myCustomRenderer')` case settles it: narrowing the public helper's parameter is the break this PR exists to avoid, and `sanitizeHTML` alone covers 3 of 9 sites while the hole stays open in the rest.

Two things in the new JSDoc, though. `'html'` does not come from the `html` cell type — `htmlRenderer.ts:22` calls `fastInnerHTML` with three arguments and no context, so it gets `'innerHTML'`. `autocompleteEditor.ts:205` is the only site that passes `'html'`.

And the paragraph now explains `'innerHTML'` twice: the new sentence has it as `fastInnerHTML`'s own default that never reaches user code, while the old trailing sentence still has it absent because no grid surface emits it but assignable for public callers. There is a stray line break mid-paragraph as well.

Haxikowy (commented):
Still open from my last review body, neither of which has a thread: the arity evidence belongs on DEV-2620, where the original proposal still stands as "verified to compile clean" on an assignment-only check, and the strict-signature follow-up at the major boundary isn't filed anywhere I can see.

handsontable/src/core/settings.ts
demtario: Both fixed in 64cbc79, and the first one was a worse error than a misattribution.

`'html'`: confirmed, `htmlRenderer.ts:22` passes three arguments and falls through to the default, so `autocompleteEditor.ts:205` is the only site. The JSDoc now says the `allowHtml` autocomplete and dropdown path only, and notes in passing where the `html` cell type actually lands.

The duplication: rewriting it exposed that the two halves disagreed, and the half I wrote later was the wrong one. "Neither reaches user code" is false for `'innerHTML'` in exactly the case that matters - `Handsontable.dom.fastInnerHTML()` is public, so a caller passing their own sanitizer and no context does receive it. That is the reason the value has to stay assignable, and I had written it as a reason it was harmless. Stated once now, correctly, with the `(string & {})` trade and the dead-branch caveat attached to it.

Line break gone.

handsontable/src/__tests__/core/sanitizer.types.ts
demtario: Thanks for re-running those, and the `build:types` ordering is the useful part - dropping `(string & {})` fails the production compile on `element.ts` and `autocompleteEditor.ts` before `test:types` gets a chance, so any mutation test of that member reads as a false negative until the production error is fixed first.

That is a trap for whoever mutation-tests this file next, so I put it in the comment above those two assertions in 64cbc79. Everything else stays as it is: `:74` kept, `:28`/`:32` kept as documentation with the DEV-2617 rationale already noted.

Haxikowy (approved):

PR #13242, opened 2026-08-25
DEV-2624: Guard option values interpolated into HTML attributes

demtario:
### Context

The PR fixes an unescaped attribute interpolation in the `dialog` plugin, and hardens the two other places where a caller-supplied value reaches an HTML attribute.

`dialog` and `emptyDataState` build their DOM by interpolating values into an HTML string that ends up in `template.innerHTML` (`helpers/templateLiteralTag.ts`). Values that land in element content go through `stripTags()`; values that land in attribute position went through nothing. `stripTags()` removes tag-like spans but escapes no `"`, `'`, `<`, `>`, so it is a text-position sanitizer only, and a quote in an attribute-position value closes the attribute and everything after it parses as further attributes.

One path was reachable. `dialog.ts` built its template variables as `{ id: this.hot.guid, ...template }`, with the spread **after** the default, so a caller-supplied `template.id` won and landed raw in `id="${id}-dialog-confirm-title"`. Unknown keys inside `template` pass `SETTINGS_VALIDATORS` untouched, because `BasePlugin#updatePluginSettings` iterates the validator's own keys only. Exploiting it requires the integrating developer to forward an attacker-influenced string into that option, so it is not a default-configuration XSS, but it is an attribute injection on a public API path released in 18.0.0. The fix assigns the GUID after the spread, which also keeps `aria-labelledby` / `aria-describedby` pointing at ids that exist.

The `button.type` interpolations in `dialog/templates/confirm.ts` and `emptyDataState/ui.ts` were **not** reachable: both are gated by an `['primary', 'secondary']` check in the plugin's `SETTINGS_VALIDATORS`, and for `emptyDataState`'s function-form `message` that same validator is re-run on the function's return value by `BasePlugin#getSetting`. They are hardened anyway, because the render site is the only place that knows the value is about to become a class name, the allowed set was duplicated as five inline literals across three plugins, and `classList.add()` throws `InvalidCharacterError` on a value containing a space.

A code-review pass then found the reachable sibling of the interpolation this branch set out to fix. `loading/content.ts` interpolated `${title}` and `${description}` raw into its `h2` and `p`, and both are documented `loading` options whose validator checks `typeof value === 'string'` only (`loading.ts:155`, `metaSchema.ts:3918`). The string reaches the DOM through the dialog's `content` path, so `loading: { title: '' }` executed by default. Both are now tag-stripped. `${icon}` stays markup on purpose: its default value is the spinner SVG.

Changes:

- `helpers/uiButton.ts` (new) declares `ButtonType`, `BUTTON_TYPES`, `DEFAULT_BUTTON_TYPE`, `isButtonType()` and `resolveButtonType()`. The declared type of the public `button.type` option stays `string`, so no consumer stops compiling; the contract is enforced at runtime only.
- `loading/content.ts` strips tags from `title` and `description`.
- `dialog/dialog.ts` assigns `id: this.hot.guid` after the `...template` spread.
- `dialog/templates/confirm.ts`, `emptyDataState/ui.ts` and `notification/ui.ts` resolve the button type at the render site. `notification/notification.ts:496` normalized it with an inline ternary before the UI ever saw it; that site now calls the shared helper too, so `resolveButtonType()` is the single place that decides the value.
- `dialog/dialog.ts` and `emptyDataState/emptyDataState.ts` validators use `BUTTON_TYPES` instead of their own literal.
- `exportFile/utils.ts` strips tags from the interpolated dialog title, matching every other template.
- `metaSchema.ts` escapes the raw `|` in two table cells that currently break the rendered API-reference table. It deliberately does **not** document a `secondary` fallback: an unrecognized `type` is rejected by the plugin validator before any render site sees it, so no user can observe the fallback, and documenting it would send people chasing behavior that never fires.
- `handsontable/AGENTS.md` records the rule for future changes.

One adjacent bug is deliberately left alone: `dialog.ts`'s `template` validator wraps `.includes(...)` in `typeof`, so the expression is the always-truthy string `'boolean'` and `template.type` is never validated at the settings layer. Fixing it would turn the current loud `throwWithCause` in `DialogUI.useTemplate` into a silent warn-and-ignore and break two existing specs, so it needs its own task.

### How has this been tested?

New unit tests assert on the **rendered DOM**, not on the generated markup, so they survive the planned conversion of these templates to DOM construction (DEV-2617) and become its regression net.

- New `handsontable/src/plugins/dialog/templates/__tests__/confirm.unit.js`: hostile `button.type` values (attribute break, element break, space, tag, non-string, unsupported name) render exactly `ht-button ht-button--secondary` with `class` as the only attribute; the a11y options stay in sync with the rendered element ids.
- New `handsontable/src/plugins/emptyDataState/__tests__/ui.unit.js`: same matrix through `EmptyDataStateUI#updateContent`.
- New `handsontable/src/plugins/exportFile/__tests__/utils.unit.js`: the export dialog title is tag-stripped.
- New `handsontable/src/plugins/loading/__tests__/content.unit.js`: `title` and `description` are tag-stripped, the icon still renders as markup, and the element ids derive from the passed id.
- `handsontable/src/plugins/dialog/__tests__/options/template.spec.js`: a `template.id` carrying `">` no longer reaches the DOM; the title and description ids stay GUID-derived and no `img` is injected.
- `handsontable/src/plugins/emptyDataState/__tests__/options/message.spec.js`: a function `message` returning a hostile `button.type` falls back to the default message and renders no button, pinning the `getSetting` return-value validation the analysis above relies on.

Each guard was confirmed to fail without the fix: reverting the spread order fails the new `template.spec.js` case, reverting `resolveButtonType` fails six cases in each of the two new plugin unit files, and reverting the `loading` `stripTags` fails the new `content.unit.js` case.

```bash
# unit (full core suite): 3764 passed, 316 suites

# the new unit tests only
npm run test:unit --prefix handsontable --testPathPattern='confirm.unit|emptyDataState/__tests__/ui|exportFile/__tests__/utils|loading/__tests__/content.unit'

# E2E for every affected plugin: 613 specs, 0 failures
npm run test:e2e --prefix handsontable --testPathPattern='(dialog|emptyDataState|notification|exportFile|loading)/__tests__'

# types, lint, build
```

### Types of changes

### Related issue(s):

1. DEV-2624

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2624

---

> [!NOTE]
> **High Risk**
> Changes touch public plugin options and HTML generation paths where untrusted or mis-forwarded strings could previously break attributes or execute markup; fixes are defensive but affect default loading title/description behavior (text instead of HTML).
> 
> **Overview**
> This PR closes **attribute injection and HTML interpretation** gaps where plugin options were interpolated into markup without escaping or validation.
> 
> The **dialog** plugin now always uses the grid **GUID** for template element ids (assigned after `...template`), so a caller-supplied `template.id` cannot break out of `id` attributes or create duplicate ids. **Loading** (and export progress UI) now **escapes** `title` and `description` via new `escapeHtml()` so strings render as text—including phrases with `<` that `stripTags` would truncate; **`icon` remains the only markup slot**. **Dialog**, **emptyDataState**, and **notification** resolve button `type` at render time through shared **`isButtonType` / `resolveButtonType`**, aligned with settings validators.
> 
> Docs, metaSchema, changelog, and **AGENTS.md** document the text-vs-markup contract. Unit and integration tests assert the rendered DOM for hostile inputs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2577f6f3be0a7a373802c3b031000282716dca16. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2624 Dialog and emptyDataState templates interpolate caller values into HTML attributes without escaping](https://app.clickup.com/t/9015210959/DEV-2624)

handsontable/src/plugins/loading/content.ts
```diff
@@ -18,8 +21,8 @@ export function loadingContent({ id, icon, title, description }: { id: string; i
     <div class="${LOADING_CLASS_NAME}__content">
       <i class="${LOADING_CLASS_NAME}__icon">${icon}</i>
       <div class="${LOADING_CLASS_NAME}__text">
-        <h2 id="${id}-${PLUGIN_KEY}-title" class="${LOADING_CLASS_NAME}__title">${title}</h2>
-        ${description ? `<p id="${id}-${PLUGIN_KEY}-description" class="${LOADING_CLASS_NAME}__description">${description}</p>` : ''}
+        <h2 id="${id}-${PLUGIN_KEY}-title" class="${LOADING_CLASS_NAME}__title">${stripTags(title)}</h2>
```
budnix: `stripTags` treats any `<` as the start of a tag and drops everything until it finds a `>`. I ran it: `'Loaded 5 < 10 rows'` comes back as `'Loaded 5 '`, and `'a < b and c > d'` becomes `'a  d'`. Before this change the string went straight to `innerHTML`, where a `<` followed by a space is just text, so it rendered fine — meaning a title with a less-than sign now silently loses half of itself. Escaping (`&`, `<`, `>`, `"`, `'`) would block the markup and keep the text whole; there is no escape helper in `helpers/string.ts` yet, so it would be a small new one.

handsontable/src/plugins/loading/content.ts
```diff
@@ -1,15 +1,18 @@
...
  * The `loadingContent` function returns a HTML string with the loading content.
  *
  * @param {object} options Loading content options.
  * @param {string} options.id Loading id.
- * @param {string} options.icon Loading icon.
- * @param {string} options.title Loading title.
- * @param {string} options.description Loading description.
+ * @param {string} options.icon Loading icon. Written as markup on purpose, so the default SVG
+ *   spinner (and any replacement for it) renders. Never pass a value derived from user input here.
+ * @param {string} options.title Loading title. Tags are stripped, so it renders as plain text.
```
budnix: This JSDoc now says tags are stripped, but the public docs still don't. `metaSchema.ts:3917` lists `title` and `description` as just "A string", and the plugin JSDoc in `loading.ts:42-44` still says "Loading title to display". The `loading` option shipped in 16.1.0, so someone passing `<strong>` there today loses it on upgrade with nothing in the docs explaining why.

.changelogs/13242.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Fixed the `dialog` plugin interpolating an unvalidated `template.id` into the dialog's `id` attribute, and the `loading` plugin rendering its `title` and `description` options as HTML. The button `type` option of the `dialog`, `emptyDataState`, and `notification` plugins is now resolved where the markup is built.",
+  "type": "security",
+  "issueOrPR": 13242,
+  "breaking": false,
```
budnix, line 6: The `loading` plugin's `title` and `description` used to render HTML and now don't. That option has been out since 16.1.0, so it's a real behavior change for anyone already passing markup — no warning, no migration note. Worth either marking it breaking or calling it out plainly in the title. The `template.id` half reads fine as non-breaking, since I checked and that key was never documented.

handsontable/src/plugins/exportFile/utils.ts
```diff
@@ -21,7 +22,7 @@ export function buildExportDialogContent(title: string): DocumentFragment {
         </svg>
       </i>
       <div class="${LOADING_CLASS_NAME}__text">
-        <h2 class="${LOADING_CLASS_NAME}__title">${title}</h2>
+        <h2 class="${LOADING_CLASS_NAME}__title">${stripTags(title)}</h2>
```
budnix: `buildExportDialogContent` has exactly one caller (`exportFile.ts:632`) and it passes `getTranslatedPhrase(EXPORT_FILE_DIALOG_TITLE)`, so this never sees user data. What it does change: a customer who registers a language dictionary with markup in that phrase now loses it, and the `<` problem above applies here too. Curious whether this was meant as cover for future callers — if so, a line in the JSDoc saying that would help.

handsontable/AGENTS.md
```diff
@@ -75,6 +75,7 @@ Gotcha: Filters `conditionCollection` uses physical indexes, `getDataAtCol()` us
...
...library authors, keep it out of the caller's reach: `dialog.ts`'s `useTemplate` call assigns `id: this.hot.guid` **after** the `...template` spread on purpose, because the reverse order let a caller-supplied `template.id` win, and unknown keys inside `template` pass `SETTINGS_VALIDATORS` untouched (`plugins/base/base.ts` iterates the validator's own keys only). A benign custom id was a defect of its own: two grids configured with the same one emitted duplicate element ids into a single document (WCAG 4.1.1), which a GUID cannot do. (2) For a value with a documented fixed set, resolve it where the markup is built, and test it with the same predicate the settings validator uses: button types go through `isButtonType()` / `resolveButtonType()` (`helpers/uiButton.ts`) in `dialog`, `emptyDataState`, and `notification`, validators included. Duplicating `['primary', 'secondary']` inline is what let the render sites and the validators drift, down to a validator that accepted `String(value)` while the render site required a real string. The guard keeps earning its place after a DOM-construction refactor, because `classList.add()` throws `InvalidCharacterError` on a value containing a space.
```
budnix: Small correction: the bullet says the shared predicate is used in `dialog`, `emptyDataState`, and `notification`, "validators included". Notification has no button-type validator — its `SETTINGS_VALIDATORS` (`notification.ts:133-138`) declares only `stackLimit` and `animation`, and action types get normalized in `normalizeActions` instead. A future reader will go hunting for that validator and come up empty.

handsontable/src/helpers/uiButton.ts
```diff
@@ -0,0 +1,38 @@
+/**
+ * A button type supported by the built-in UI components (`dialog`, `emptyDataState`,
+ * `notification`). Each type maps to the `ht-button--<type>` modifier class.
+ */
+export type ButtonType = 'primary' | 'secondary';
```
budnix, line 5: I grepped for both: `ButtonType` is imported nowhere, and `DEFAULT_BUTTON_TYPE` is only used inside this file. Meanwhile `notification.ts:22` still hand-writes `type?: 'primary' | 'secondary'`, and `confirm.ts` and `emptyDataState/ui.ts` still type their buttons as `{ type: string }`. So the copy-paste drift this helper exists to stop can still happen at the type level — using `ButtonType` at those three spots would close that.

handsontable/src/plugins/emptyDataState/__tests__/options/message.spec.js
```diff
@@ -357,6 +357,33 @@ describe('EmptyDataState - message option', () => {
       .toBe('There’s nothing to display yet.');
   });
 
+  it('should ignore a message whose button `type` is not supported (function message)', async() => {
```
budnix, line 365: This test passes on develop too. The only source change in `emptyDataState.ts` swaps `(typeof item.type === 'string' && ['primary','secondary'].includes(item.type))` for `isButtonType(item.type)`, and those two are the same check — so the validator already rejected this input, the message already fell back to the default, and the render site never runs. `ui.unit.js` is the file actually covering the new behavior. Not harmful, just worth knowing this one won't catch a regression.

handsontable/src/plugins/loading/content.ts
demtario: Agreed, and thanks for running it. This was a regression I introduced, not a pre-existing quirk: before the change the string reached `innerHTML` where `5 < 10` is just text.

Fixed by escaping rather than stripping. `escapeHtml()` is now in `helpers/string.ts` and both `loading/content.ts` and `exportFile/utils.ts` use it. `stripTags` stays where it already shipped (`dialog` and `emptyDataState` titles), so no existing behavior changes there.

Two tests cover it: `helpers/__tests__/string.unit.ts` asserts the escaping and pins the truncation you found (`stripTags('Loaded 5 < 10 rows')` returns `'Loaded 5 '`), and `loading/__tests__/content.unit.js` asserts `'Loaded 5 < 10 rows'` and `'a < b and c > d'` survive intact.

One consequence worth naming: markup passed in `loading.title` now renders literally instead of disappearing. That is the visible change your changelog comment asked about, and it is documented in the three places from your other comment.

handsontable/src/plugins/loading/content.ts
demtario: Right, the docs were the gap. Now documented in all three places:

- `metaSchema.ts` (the `loading` option table): `title` and `description` render as text, markup shows up literally, use `icon` for the one slot that takes markup.
- `loading.ts` plugin JSDoc: same, per option.
- `docs/content/guides/dialog/loading/loading.md`, in the Custom configuration section.

The JSDoc wording also changed from "tags are stripped" to "escaped, so markup shows up literally", which is what the code does after the fix on your other comment.

.changelogs/13242.json
demtario: Took the "call it out plainly" branch. The entry now reads, after the `template.id` clause:

> Those two `loading` options now render as text, so markup passed in them shows up literally instead of being interpreted; use `icon` for markup.

Left `breaking: false` on purpose: nothing is removed, the accepted type is unchanged, and the option keeps working. What changes is how a markup-carrying value renders, which is the security fix itself.

Two things I did not do, both yours to call:

1. Flip it to `breaking: true` so the entry leads the Security section. Say the word and I will.
2. Add a bullet to `migrating-from-18.1-to-18.2.md` (that page arrived on develop with DEV-2600, so it exists now). I lean yes if we think anyone is passing markup there, and the whole point of your comment is that we cannot know.

handsontable/src/plugins/exportFile/utils.ts
demtario: Correct on the reachability, and yes, cover for future callers was the intent. The JSDoc now says so:

> The title is escaped rather than trusted. Its only current caller passes a translated phrase, which no end user controls, but a customer-registered language dictionary does reach it, and this function is the kind that acquires callers. Escaping (not stripping) keeps a phrase containing `<` intact.

The `<` problem you raised on `loading/content.ts` applied here too, so this site escapes rather than strips, which also means a dictionary phrase with markup keeps its text instead of losing it.

handsontable/AGENTS.md
demtario: You are right, and it would have sent someone hunting. Corrected. The bullet now reads that the predicate is used at every render site (`dialog`, `emptyDataState`, `notification`) and in the two `SETTINGS_VALIDATORS` that have a button-type check (`dialog` and `emptyDataState`), with `notification` normalizing action types in `#normalizeOptions` instead.

handsontable/src/helpers/uiButton.ts
demtario: Half applied, half deliberately not, so let me split it.

Applied: `NotificationAction.type` is now `ButtonType` instead of a hand-written union, and `DEFAULT_BUTTON_TYPE` is module-local since nothing outside the helper used it.

Not applied, on purpose: `confirm.ts` and `emptyDataState/ui.ts` now type the incoming `type` as `unknown`, not `ButtonType`. Those two receive a value straight from a plugin setting, and `resolveButtonType()` exists precisely because they do not trust it. Declaring the narrow type on the way in would assert a check that has not run yet, which is the same mistake as the validator drift this PR fixes. The narrow type belongs on what the resolver returns, and it is there.

The drift you are pointing at is real though, and there was a live instance: `dialog.ts` tested `BUTTON_TYPES.includes(String(item.type))` while the render site required an actual string, so an object with a `toString()` returning `'primary'` passed the validator and then rendered as `secondary`. Both validators now call `isButtonType()`, the same predicate `resolveButtonType()` uses, so what the validator accepts is exactly what the markup can carry.

Also deduped while in there: `emptyDataState/ui.ts` had that button shape written out three times, now one `MessageButton` / `MessageContent` pair.

handsontable/src/plugins/emptyDataState/__tests__/options/message.spec.js
demtario: Agreed, and this one had already been called out in the second review pass, so it is now labelled rather than removed. The spec carries a comment saying it is a characterization test that passes on the pre-fix code too, because `getSetting` re-runs the validator on the function return value and the whole message is dropped before any button renders. It is there to pin that guard, since the reachability argument for the render-site fix depends on it, and the comment points at `emptyDataState/__tests__/ui.unit.js` for the actual render-site coverage. The title changed to "should ignore a message whose button `type` is not supported" so it no longer reads as if a type were being filtered.

Kept rather than deleted because that guard is load-bearing for the analysis: if `#wrapSettingWithValidator` ever stopped validating return values, this is the test that would notice.

docs/content/guides/dialog/loading/loading.md
```diff
@@ -82,6 +82,9 @@ To enable the Loading plugin, set the [`loading`](@/api/options.md#loading) opti
 
 The loading dialog supports customization of the icon, title, and description.
 
+The `title` and `description` options render as text, so markup passed in them shows up literally.
```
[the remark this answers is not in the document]
demtario: Valid catch. The guide gained a paragraph documenting that `title` and `description` render as text, which is a substantive content change under `docs/AGENTS.md` section 2.6, and the sibling `notification.md` already carries the same field. Added `menuTag: updated`.

Committed locally as `2577f6f3b` and not yet pushed on purpose: pushing cancels the run currently in flight, and that run is the verdict on a deflaked E2E spec in this PR. It goes up as soon as `CI Gate` settles.

budnix (approved):

PR #13248, opened 2026-08-26
Bump Playwright to 1.62.1 monorepo-wide (catalog + container image)

danielzytohoc:
### Context

The PR replaces Dependabot's #13152 and fixes the config that produced it. That PR bumped `@playwright/test` only in `visual-tests/package.json` without regenerating `pnpm-lock.yaml` — Dependabot watches `directory: "/visual-tests"`, where no lockfile exists in a pnpm workspace — so every CI run died at `pnpm install` with `ERR_PNPM_OUTDATED_LOCKFILE` before a single test job started.

This PR does the bump the way the repo policy asks for it (one Playwright version, catalog and container tag together, never apart):

- pnpm **catalog**: `1.61.1` → **`1.62.1`**; e2e container image → `mcr.microsoft.com/playwright:v1.62.1-noble` (`.github/workflows/e2e.yml`).
- `visual-tests/` (`~1.60.0`) and `docs/` (`1.45.x`) migrate onto `"@playwright/test": "catalog:"` — the migration `tests/README.md` already planned.
- `performance-tests/` is npm-managed (own `package-lock.json`), so it is bumped by hand to `~1.62.1` in lockstep.
- `visual.yml` browser-cache key now also hashes `pnpm-workspace.yaml`: with `catalog:` in `visual-tests/package.json`, a version bump no longer changes that file, and the old key would have kept serving stale browsers on a cache hit.
- `.github/dependabot.yml` is repointed at the workspace root, so future Playwright PRs regenerate the root lockfile instead of failing frozen-lockfile installs; its comment documents the container-tag pairing rule.
- `tests/README.md` and the `pnpm-workspace.yaml` catalog comment updated to match.

**Watch items for review:**

- **Visual regression (reg-suit)**: 1.60 → 1.62 ships new Chromium/Firefox/WebKit builds — expect screenshot diffs against the `develop` golden records. Review the report the `Visual / Compare` sticky comment links to and approve with the `visual-approved` label (all-or-nothing; the label is removed again on every push). Argos is gone from `develop` since this PR was opened, so the earlier "Argos approval pass" note no longer applies.
- **docs**: jumps 1.45.x → 1.62.1; expect screenshot churn on the next docs-visual-tests run.
- **performance-tests**: newer Chromium may shift timings against the gh-pages golden until the post-merge golden refresh.
- Playwright 1.62 isolated the headless clipboard — clipboard-related specs are the most likely functional wobble.

`[skip changelog]` — CI/tooling and test-package dependency change only; no shippable source touched.

### How has this been tested?

- `pnpm install` regenerated the lockfile with exactly one `playwright@1.62.1` entry; `tests/`, `visual-tests/`, and `docs/` all resolve `1.62.1` via `catalog:` (verified with each package's `playwright --version`).
- `npm install --package-lock-only` regenerated `performance-tests/package-lock.json` at `1.62.1`.
- `playwright test --list` passes on all four configs under 1.62.1 — `tests/` (2538 tests), `visual-tests/` multi-framework (91) and cross-browser (129), `docs/` (521) — so config loading and spec collection work on the new version.
- The full theme × bundle matrix, visual regression, and docs suites run in CI; the e2e legs run inside the matching `v1.62.1-noble` container.
- **Refreshed onto `develop` (2026-09-02, e3360d827):** one conflict, `visual-tests/package.json` — `develop` dropped `@argos-ci/core` (reg-suit replaced Argos) while this branch moved `@playwright/test` to `catalog:`; resolved by keeping the catalog specifier and accepting the removal. The auto-merged `pnpm-lock.yaml` was verified consistent (`pnpm install --lockfile-only` produced no change) and audited against `develop`: the only identity changes in the lock are the Playwright bump itself — no unrelated `latest` specifier floated. Every intended change re-verified after the merge: catalog `1.62.1`, container `v1.62.1-noble`, the `visual.yml` cache key hashing `pnpm-workspace.yaml`, all three `catalog:` specifiers, `performance-tests` at `~1.62.1`, dependabot repointed at the root.

### Types of changes

### Related issue(s):

1. Replaces #13152

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

(Touches test/CI packages only: `tests/`, `visual-tests/`, `docs/`, `performance-tests/`, workflows.)

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Touches all Playwright tiers and CI caching; expect visual/docs screenshot churn and transient perf compare noise, while the FormulasGridPage changes reduce E2E flake risk on 1.62.
> 
> **Overview**
> Bumps **@playwright/test** to **1.62.1** monorepo-wide: pnpm catalog, `e2e.yml` container (`v1.62.1-noble`), `performance-tests/` lockfile, and `docs/` / `visual-tests/` moved onto `"catalog:"` so one version replaces the old split pins (1.45.x / ~1.60).
> 
> **CI and Dependabot hardening:** Dependabot now watches `/` (root lockfile) with notes that catalog and container tag must bump together; new `playwright-version-sync.test.mjs` fails the tooling gate if catalog, `e2e.yml`, and docs drift. Path filters and Playwright/browser snapshot cache keys include `pnpm-lock.yaml` / `pnpm-workspace.yaml` so catalog-only bumps re-run visual and docs legs and avoid stale browser caches.
> 
> **Functional test fix for 1.62:** `FormulasGridPage` wheel-scrolls drag targets into the pressable holder (sticky headers), asserts the selected range after drag-select, and replaces iteration-count auto-scroll pumping with time-bounded `expect.poll` — addressing horizon-theme flakes from layout/metric shifts on the new engine. Docs (`tests/README.md`, `tests/AGENTS.md`, `performance-tests/README.md`, `.ai/STACK.md`) updated for the new version and operational caveats (visual/perf golden noise until baselines refresh).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b2e63bc45d26965c6395b2934c9d784d3ba16eea. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


danielzytohoc:
**Horizon Playwright legs — root cause and fix** (commit 77de3e59d)

The first run's only real failures were the two `Playwright (theme: horizon)` legs, both on one spec (`e2e/fill-handle-formulas.spec.ts`). It is not a product regression and not the browser bump itself — it is a latent race in the spec's page object that the new Chromium's ~1px row-metric shift finally exposed:

1. `selectRange(1,2 → 2,4)` computed its drag end from cell (2,4)'s bounding box. Bounding boxes ignore overflow clipping, and on horizon (tallest rows) that cell's **center lies just past the holder's 130px fold**.
2. Parking a real mouse there means "extend the selection past the edge": DragToScroll legitimately kicks in, and the selection overshoots to `1,2 → 3,4`.
3. The fill handle then belongs to row 3's bottom edge — outside the visible holder. `toBeVisible()` passes for clipped elements, so the corner grab silently pressed the page body, `afterOnCellCornerMouseDown` never fired, DragToScroll never armed, and the "auto-scroll never brought cell 6,4 into view" error followed.

The same failure reproduces on macOS with Playwright **1.61.1** — the old CI green was luck: the overshoot race happened to settle on the intended range plus a helpful scroll on that platform's row metrics. The 1.62 Chromium shifted cell (2,4)'s center across the fold on Linux too.

Fix (page object only, no product code):
- `selectRange` clamps both drag points into the holder's visible area and **asserts the achieved range** afterwards, so an overshoot fails loudly at its cause.
- Before grabbing the corner, the page object wheel-scrolls the holder until the corner is genuinely inside the visible area — exactly what a user does when the handle is below the fold.
- The auto-scroll wait is now a time-bounded `expect.poll` instead of a fixed 200-iteration mousemove pump (an iteration count is a hidden wall-clock budget that shrinks with every Playwright/CDP speedup).

Verified locally on all six legs (main/horizon/classic × UMD/full.min) under 1.62.1.

tests/fixtures/pages/FormulasGridPage.ts
```diff
@@ -42,36 +42,103 @@ export class FormulasGridPage {
...
+    const holder = await this.grid.locator('.ht_master .wtHolder').boundingBox();
 
-    if (!from || !to) {
-      throw new Error('range endpoints are not rendered');
+    if (!from || !to || !holder) {
+      throw new Error('range endpoints or the holder are not rendered');
     }
 
-    await this.page.mouse.move(from.x + from.width / 2, from.y + from.height / 2);
+    const clampX = (x: number) => Math.min(Math.max(x, holder.x + 2), holder.x + holder.width - 2);
```
budnix: `.ht_master .wtHolder` covers the whole grid body, including the strips that the column-header and row-header overlays sit on top of. The fixture sets `rowHeaders: true`, so `holder.x + 2` is inside the row header — a left-clipped target would press the header, select the whole row, and the new assertion below would blame the target cell rather than the clamp. `clampY`'s lower bound has the same problem with the column headers. Curious whether wheel-scrolling the target into view, the way `dragAreaFillHandleToCell` now does, would fit better here — clamping only gives the right range while the target row still touches the bottom 2px.

tests/fixtures/pages/FormulasGridPage.ts
```diff
@@ -42,36 +42,103 @@ export class FormulasGridPage {
...
+    // On tall-row themes a freshly selected range can end past the fold, which
+    // leaves its fill handle overflow-clipped below the holder. The corner's
+    // box ignores clipping (toBeVisible() passes for a fully clipped element),
+    // so grabbing it blind would silently press the page body and never arm
+    // the grid's drag-to-scroll. Mirror what a user does: wheel-scroll the
+    // holder until the corner actually sits inside its visible area.
+    await this.page.mouse.move(holderBox.x + holderBox.width / 2, holderBox.y + holderBox.height / 2);
+    await expect
+      .poll(async() => {
+        const corner = await this.areaCorner.boundingBox();
```
budnix: `cellInHolderView` further down guards with `count() === 0` before calling `boundingBox()`, and its comment says why: `boundingBox()` waits for `attached` first, so a missing element stalls instead of returning. This call has no such guard, and `.wtBorder.area.corner` is torn down and rebuilt on re-render — which the wheel just below triggers. A probe that lands mid-render blocks for the 10s locator timeout, blows past the 5s poll budget, and you get a generic test timeout instead of the message.

```suggestion
        if (await this.areaCorner.count() === 0) {
          return false;
        }

        const corner = await this.areaCorner.boundingBox();
```

tests/fixtures/pages/FormulasGridPage.ts
```diff
@@ -42,36 +42,103 @@ export class FormulasGridPage {
...
+
+        if (!corner) {
+          return false;
+        }
+        if (corner.y + corner.height > holderBox.y + holderBox.height + 1) {
+          await this.page.mouse.wheel(0, 40);
```
budnix: `expect.poll` uses fixed intervals (100, 250, 500, then 1000ms), so a 5s budget gives roughly 7-8 probes no matter how fast the machine is — about 300px of scroll in total. That turns the reach into a hidden budget, which is the same thing the new `tests/AGENTS.md` section in this PR warns about, one method below. Scaling the wheel by how far the corner still has to travel drops the cap.

```suggestion
          await this.page.mouse.wheel(0, Math.max(40, (corner.y + corner.height) - (holderBox.y + holderBox.height)));
```

tests/fixtures/pages/FormulasGridPage.ts
```diff
@@ -42,36 +42,103 @@ export class FormulasGridPage {
...
+        }
+
+        return true;
+      }, {
+        message: 'the area fill handle never scrolled into the holder view',
+        timeout: 5000,
+      })
+      .toBe(true);
+
+    const handle = await this.areaCorner.boundingBox();
```
budnix, line 115: `gridBox` and `holderBox` are read before the wheel loop, but the loop can move things. When the holder is already at its scroll limit Chrome passes the wheel to the page, so the grid moves up while `gridBox.y` keeps its old value — and `belowGridY` right below is computed from it. The drag point can then land inside the grid (auto-scroll never arms) or much further down than the 20px the comment assumes. Worth re-reading both boxes here, after the poll.

tests/fixtures/pages/FormulasGridPage.ts
```diff
@@ -81,19 +148,31 @@ export class FormulasGridPage {
...
-
-    // boundingBox() reports overflow-clipped cells too, so an exhausted pump
-    // must fail HERE with its real cause — not later, as a confusing data
-    // mismatch after the gesture lands on a clipped coordinate.
-    if (!(await this.cellInHolderView(row, col))) {
-      throw new Error(`auto-scroll never brought cell ${row},${col} into the holder view`);
-    }
+    // Auto-scroll is TIMER-driven (DragToScroll's ScrollTimer advances one row
+    // per self-rescheduled tick; at 20px past the edge the tick interval is
+    // near its 500ms maximum), so the wait must be bounded by TIME, never by
```
budnix: Small thing on the numbers: with `DragToScroll`'s `min: 20, max: 500`, `rampDistance: 120` and `logScale: 200`, a pointer 20px past the edge gives about 180ms per tick, not close to 500ms. `belowGridY` is `gridBox.y + gridBox.height + 20`, so that's exactly the case this hits. Worth fixing since the 15s budget below was picked from this figure.

tests/fixtures/pages/FormulasGridPage.ts
```diff
@@ -81,19 +148,31 @@ export class FormulasGridPage {
...
+
+    await expect
+      .poll(async() => {
+        jitter = 1 - jitter;
+        await this.page.mouse.move(startX, belowGridY + jitter);
+
+        return this.cellInHolderView(row, col);
+      }, {
+        message: `auto-scroll never brought cell ${row},${col} into the holder view`,
+        timeout: 15000,
```
budnix: The two polls in this method ask for 5s + 15s, and `tests/playwright.config.ts` sets `timeout: 20_000` for the whole test. Add the time spent in `goto()` and `selectRange()` and Playwright kills the test before this poll can run out — so instead of the message above you get "Test timeout of 20000ms exceeded" with no location, which is the opposite of what the comment wants. CI runs `retries: 1` with `failOnFlakyTests`, so a horizon leg that only just fits still fails the run. Either raise the timeout for this spec or shrink both budgets so they leave room.

docs/package.json
```diff
@@ -68,7 +68,7 @@
   },
   "devDependencies": {
     "@octokit/rest": "^18.6.7",
-    "@playwright/test": "1.45.x",
+    "@playwright/test": "catalog:",
```
budnix, line 71: This moves docs from 1.45.x to 1.62.1 — about 18 Chromium releases. `docs-visual-tests.yml` keys its snapshot cache on `<repo>/<ref>-<run_id>` with a fallback to `refs/heads/develop`, and no Playwright version anywhere in the key. After merge the job restores develop's snapshots (rendered by the old Chromium), sets `cache-restored=true`, skips the `--update-snapshots` step, and every test in `visualDocs.spec.ts` fails on font and pixel differences. Adding `hashFiles('pnpm-lock.yaml')` to that key, or running the workflow once with `update-snapshots=true`, would cover it.

pnpm-workspace.yaml
```diff
@@ -13,10 +13,12 @@ packages:
...
 # One Playwright version across the monorepo — the catalog version and the CI
-# container tag (mcr.microsoft.com/playwright:vX.Y.Z-noble) bump together, never
-# apart. visual-tests migrates onto the catalog in a separate reviewed change.
+# container tag (mcr.microsoft.com/playwright:vX.Y.Z-noble in e2e.yml) bump
+# together, never apart. tests/, visual-tests/, and docs/ consume it via
+# `catalog:`; performance-tests/ is npm-managed (own package-lock.json), so its
+# pin is kept in lockstep by hand.
 catalog:
-  '@playwright/test': 1.61.1
+  '@playwright/test': 1.62.1
```
budnix, line 21: Now that visual-tests and docs read the version from here, a future bump touches only this file and `pnpm-lock.yaml`. `checks.yml`'s `test-visual` filter watches only `./examples/next/visual-tests/**` and `./visual-tests/**`, and `docs` watches only `./docs/**` and `./handsontable/package.json` — so neither would run on the one change that swaps the browser binaries. This PR doesn't hit it because it also edits both package.json files, so green CI here says little about the next bump. Adding `./pnpm-lock.yaml` and `./pnpm-workspace.yaml` to both filters would close it.

.github/dependabot.yml
```diff
@@ -1,7 +1,16 @@
...
+  # the workspace root, not a leaf package (a leaf directory has no lockfile,
+  # which made Dependabot open manifest-only PRs that failed the frozen-lockfile
+  # install; see PR #13152). Merging a Playwright bump requires moving the
+  # container tag in .github/workflows/e2e.yml (mcr.microsoft.com/playwright)
+  # to the same version in the same PR — the catalog and the image bump
+  # together, never apart. performance-tests/ (npm-managed, own package-lock)
+  # is bumped by hand in lockstep.
   - package-ecosystem: "npm"
-    directory: "/visual-tests"
+    directory: "/"
```
budnix, line 13: The `allow` rule below still names `@playwright/test`, but after this PR no package.json pins it — the version only lives in the catalog. Did you get to check that Dependabot's npm ecosystem rewrites pnpm catalog entries? If it doesn't, the weekly bump quietly stops and nothing turns red. And if it does, its PR would move the catalog but leave `e2e.yml` on the old tag, which fails with `Executable doesn't exist at /ms-playwright/chromium-...` — `fork-guards.test.mjs` already asserts workflow contents, so a small test tying the image tag to the catalog would turn this comment into a real check.

performance-tests/package.json
```diff
@@ -12,7 +12,7 @@
   "author": "Handsoncode <hello@handsontable.com>",
   "license": "ISC",
   "devDependencies": {
-    "@playwright/test": "~1.58.2",
+    "@playwright/test": "~1.62.1",
```
budnix, line 15: `run.mjs` does `npx playwright install chromium`, so the suite measures whatever Chromium this pin ships. The golden on gh-pages (`performance-reports/develop/latest.json`) came from 1.58.2's, so the first PR after this merges compares new-engine timings and heap numbers against old-engine ones, and `report-builder.mjs` will flag "regressed X%" on scenarios nobody touched. It's callouts, not a gate, but reviewers read them as real until the develop push rewrites the golden — worth a refresh here or a note in the history index.

visual-tests/package.json
```diff
@@ -20,7 +20,7 @@
   "license": "CC BY 4.0",
   "devDependencies": {
     "@argos-ci/core": "^5.1.1",
-    "@playwright/test": "~1.60.0",
+    "@playwright/test": "catalog:",
```
budnix, line 23: This goes 1.60 → 1.62.1 across chromium, firefox and webkit, so expect a wide Argos diff on this PR from anti-aliasing and font metrics alone. Fine to take as a new baseline, just worth approving it knowingly rather than waving it through — the `visual.yml` change covers the browser-binary cache, not the snapshots.

tests/README.md
```diff
@@ -78,6 +85,6 @@ npm --prefix handsontable run test:walkontable # the rendering engine's own pipe
 
 ## One Playwright version across the monorepo
 
-There must be **one** Playwright version, installed once, aligned to the version CI runs — currently **1.61.1**. The mechanism is the pnpm **catalog**: `pnpm-workspace.yaml` declares `catalog: { '@playwright/test': 1.61.1 }` and this package sets `"@playwright/test": "catalog:"`. `visual-tests/` (and later `performance-tests/`, `docs/`) move onto the catalog in their own reviewed bumps.
+There must be **one** Playwright version, installed once, aligned to the version CI runs — currently **1.62.1**. The mechanism is the pnpm **catalog**: `pnpm-workspace.yaml` declares `catalog: { '@playwright/test': 1.62.1 }` and this package, `visual-tests/`, and `docs/` all set `"@playwright/test": "catalog:"`. `performance-tests/` is npm-managed (its own `package-lock.json`), so its pin is bumped by hand in lockstep.
```
budnix, line 88: `.ai/STACK.md:41` still says "Playwright 1.61", so the monorepo stack reference now disagrees with the catalog, this file, and the container tag. One-line fix, but it's the same drift the "never apart" rule exists to prevent.

tests/fixtures/pages/FormulasGridPage.ts
danielzytohoc: Right on both counts — the clamp could land in the header strips, and it only rescued edge-touching targets. Replaced clamping entirely with your suggestion in c822b4f: both endpoints are wheel-scrolled into the holder's pressable area (holder minus the `ht_clone_top` band) before the drag, and the achieved-range assertion stays as the loud backstop. One wrinkle surfaced while verifying on horizon: with a floored wheel step the `to`/`from` pair ping-pongs (details in the thread on the wheel-step suggestion), so the wheel moves the exact remaining distance instead.

tests/fixtures/pages/FormulasGridPage.ts
danielzytohoc: Applied in c822b4f — the corner reach now goes through a shared `#wheelIntoView` helper that checks `count()` before `boundingBox()`, exactly as `cellInHolderView` does and for the reason you quote.

tests/fixtures/pages/FormulasGridPage.ts
danielzytohoc: Took the scaling, dropped the floor: the wheel now moves the exact remaining distance + 2px (c822b4f). The `Math.max(40, …)` floor turned out to reintroduce the original failure on horizon — `to` needs ~21px down, `from` then needs a 2px nudge up, and the 40px floor undid the first wheel and re-clipped `to`, so the drag overshot to row 3 again. Exact steps converge in one wheel per direction and still remove the hidden reach cap you flagged.

tests/fixtures/pages/FormulasGridPage.ts
danielzytohoc: Fixed in c822b4f — both boxes are read after the wheel poll, with a comment carrying your scroll-chaining rationale (holder at its limit hands the wheel to the page, so the grid itself can move).

tests/fixtures/pages/FormulasGridPage.ts
danielzytohoc: You're right — I'd ignored `logScale`. log(1 + 0.1667·200) / log(201) ≈ 0.667 → 500 − 0.667·480 ≈ 180ms per tick. Comment corrected in c822b4f, and the poll budget note now derives from that figure instead.

tests/fixtures/pages/FormulasGridPage.ts
danielzytohoc: Shrunk in c822b4f: the wheel polls get 2s each and the auto-scroll poll 8s (~40 ticks of headroom at ~180ms/tick) — worst case ≈14s, leaving room for goto + gestures inside the 20s test timeout, so an exhausted poll still fails with its message rather than a locationless test timeout.

docs/package.json
danielzytohoc: Added `hashFiles('pnpm-lock.yaml')` to the restore key, both restore-keys, and the save key in c822b4f. A browser bump now misses the cache, which routes into the existing `Initialize snapshots` step — the suite re-baselines itself instead of failing every spec against old-engine snapshots.

pnpm-workspace.yaml
danielzytohoc: Added `./pnpm-lock.yaml` + `./pnpm-workspace.yaml` to both `test-visual` and `docs` in c822b4f. For the record: the `config-files` anchor already carries `./pnpm-lock.yaml`, so build-umd (and with it the Playwright container legs), unit, and perf already route on a catalog-only bump — these two filters were exactly the remaining gaps.

.github/dependabot.yml
danielzytohoc: Checked: Dependabot supports pnpm workspace catalogs since Feb 2025 (GA: https://github.blog/changelog/2025-02-04-dependabot-now-supports-pnpm-workspace-catalogs-ga/), so the weekly bump keeps flowing — which makes your second scenario the live one. c822b4f adds `.github/scripts/__tests__/playwright-version-sync.test.mjs` to the tooling gate (I found no fork-guards.test.mjs in the tree, so it sits beside repo-root.test.mjs and runs via `npm run test:tooling`): it asserts the e2e.yml image tag, tests/README.md, and .ai/STACK.md all match the catalog version, so a catalog-only PR goes red in `Checks / tooling tests` with a message naming the bump-together rule.

performance-tests/package.json
danielzytohoc: Documented in performance-tests/README.md (c822b4f): the golden carries the engine it was measured with, the first post-merge PRs may show phantom regression callouts on untouched scenarios, and the develop push that follows the merge rewrites the golden. I stopped short of a pre-merge refresh — the golden job only runs on develop pushes — but happy to trigger a manual refresh right after merge if you'd prefer that over the note.

visual-tests/package.json
danielzytohoc: Agreed on approving knowingly. For the record, Argos evaluated this PR and came back green — build 5952: https://app.argos-ci.com/handsontable/handsontable/builds/5952 — so at current thresholds the 1.60→1.62 jump didn't move these demos' pixels. Worth the pre-merge glance at that build all the same.

tests/README.md
danielzytohoc: Fixed to 1.62.1 in c822b4f — and .ai/STACK.md is now one of the files the new `playwright-version-sync` tooling test asserts against the catalog, so this drift class turns red instead of lingering.

budnix (dismissed):

qunabu:
@cursoragent resolve conflicts, @danielzytohoc ready to be merged ?

budnix (approved):

PR #12727, opened 2026-06-08
DEV-1841: Remove the deprecated PersistentState plugin

budnix:
### Context

The PR removes the `PersistentState` plugin, finishing work that was announced as done in the 17.0 changelog (PR #12015: "Removed ... the `PersistentState` plugin") but never fully landed. PR #12015 de-registered the plugin from `plugins/index.ts`, making it inert, but left the source files, type declarations, now-dead hook plumbing in five dependent plugins, and the docs API pages behind.

**What this PR does:**
- Deletes `handsontable/src/plugins/persistentState/` (all 4 files)
- Removes the `PersistentState` type entry from `plugins/types.ts`
- Strips the private `persistentStateSave`/`persistentStateLoad` methods from `columnSorting`, `manualColumnMove`, and `manualRowMove` (all `@private`, no public-API impact)
- Simplifies `moveBySettingsOrLoad()` in both move plugins to use plugin settings only
- Makes `saveManualColumnWidths()`, `loadManualColumnWidths()`, `saveManualRowHeights()`, and `loadManualRowHeights()` no-ops with `@deprecated` warnings — these are public, so they get one release cycle before removal
- Registers `persistentStateSave`, `persistentStateLoad`, `persistentStateReset` in `REMOVED_HOOKS` (v17.0.0) so `addHook()` calls warn
- Removes `docs/content/api/persistentState.md`, `docs/content/api/storage.md`, and the `api/persistent-state` route from `_worker.js`

ClickUp task: https://app.clickup.com/t/9015210959/DEV-1841

### How has this been tested?

- `npm --prefix handsontable run test:types` — passes (type build + downlevel + TS 5.1 consumer check)
- `npm --prefix handsontable run eslint src/plugins/types.ts src/core/hooks/constants.ts src/plugins/columnSorting/columnSorting.ts src/plugins/manualColumnMove/manualColumnMove.ts src/plugins/manualRowMove/manualRowMove.ts src/plugins/manualColumnResize/manualColumnResize.ts src/plugins/manualRowResize/manualRowResize.ts` — no issues
- Grep confirms no stray `persistentState` references remain in `src/` outside `REMOVED_HOOKS` and the new deprecation messages

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature or improvement (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Additional language file or change to the existing one (translations)

### Related issue(s):

[...]

PR #12224, opened 2026-03-26
Change the columnHeaders property name to colHeaders

adrianszymanski89:
### Context

This PR changes the `columnHeaders` options name for the exportToFile plugin to `colHeaders` to make it consistent with the naming in the Handsontable environment. This includes changes in the new Export to Excel guide and examples.

CC @jansiegel 

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature or improvement (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Additional language file or change to the existing one (translations)

### Related issue(s):
1. https://github.com/handsontable/dev-handsontable/issues/225

### Affected project(s):
- [X] `handsontable`
- [ ] `@handsontable/angular-wrapper`
- [ ] `@handsontable/react-wrapper`
- [ ] `@handsontable/vue3`

### Checklist:

- [ ] I have reviewed the guidelines about [Contributing to Handsontable](https://github.com/handsontable/handsontable/blob/master/CONTRIBUTING.md) and I confirm that my code follows the code style of this project.
- [ ] I have signed the [Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLScpMq4swMelvw3-onxC8Jl29m0fVp5hpf7d1yQVklqVjGjWGA/viewform?c=0&w=1)
- [ ] My change requires a change to the documentation.

---

> [!NOTE]
> **Medium Risk**
> Touches core export option parsing and XLSX multi-sheet handling; while covered by tests and backwards-compat aliasing, mistakes could change whether headers are included in exported CSV/XLSX files.
> 
> **Overview**
> Standardizes `ExportFile` export options by renaming `columnHeaders` to `colHeaders` for CSV/XLSX exports (including per-sheet `sheets[]` config and context-menu derived options), and updates defaults/JSDoc/TypeScript typings to expose `colHeaders` while marking `columnHeaders` as deprecated.
> 
> To reduce upgrade friction, the plugin now *promotes* `columnHeaders` to `colHeaders` when the new key isn’t provided (implemented in both the base option merge and `DataProvider#setOptions`, plus XLSX per-sheet normalization), with tests added/updated to cover alias behavior. Docs/examples and a new 17.x→18.0 migration guide are updated to reflect the new option name, and CI’s `pnpm audit` is made non-blocking on registry endpoint errors.
> 
[...]

PR #12015, opened 2026-02-13
Remove deprecated packages

adrianspdev:
### Context
Remove [deprecated packages](https://handsontable.com/docs/react-data-grid/deprecation-policy/#list-of-deprecations)

Below is the list of removed elements:

- [x] Angular wrapper legacy
- [x] React wrapper legacy
- [x] Vue 2 wrapper
- [x] persistentState plugin
- [x] hot undo/redo methods

### How has this been tested?
Locally

### Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Related issue(s):
1. https://github.com/handsontable/dev-handsontable/issues/3061

### Affected project(s):
- [x] `handsontable`
- [x] `@handsontable/angular`
- [x] `@handsontable/react`
- [x] `@handsontable/vue`

### Checklist:

- [x] I have reviewed the guidelines about [Contributing to Handsontable](https://github.com/handsontable/handsontable/blob/master/CONTRIBUTING.md) and I confirm that my code follows the code style of this project.
- [x] I have signed the [Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLScpMq4swMelvw3-onxC8Jl29m0fVp5hpf7d1yQVklqVjGjWGA/viewform?c=0&w=1)
- [x] My change requires a change to the documentation.

PR #13249, opened 2026-08-26
Update the deprecation plan for 18.1

adrianspdev:
### Context

The PR changes how deprecated APIs report themselves, so that every API marked `@deprecated` in the codebase behaves the same way before the 18.1 release: a JSDoc `@deprecated` message that names the deprecating and removing versions, a one-time `console.warn` at runtime, and a row in the documented deprecation list. Nothing is removed – 18.1 is a minor release, and every message names 19.0 as the removal version.

Before this change:

- `deprecatedWarn()` printed on **every** call, although the deprecation policy promises a warning "fired only once" (`.ai/BREAKING-CHANGES.md`). Calling `saveManualColumnWidths()` in a loop flooded the console.
- The four resize-state methods carried a bare `@deprecated` tag, which the API reference renders as an empty warning box.
- The `exportFile` `columnHeaders` alias, `Handsontable.helper.sanitize()`, and the `Settings` / `DataProviderOptions` type aliases were tagged `@deprecated` in the typings but produced no runtime warning and were missing from the deprecation list in the docs.
- `Handsontable.CellMeta` (namespace alias) was tagged `@deprecated Use CellProperties`, while 18.0 promoted `CellMeta` to a top-level named export (#11240). The two now agree: the alias is no longer marked deprecated.
- Removed options (`persistentState`, removed in 17.0; `correctFormat`, removed in 18.0) were silently ignored. `updateSettings()` now warns once (it never throws).

Code changes:

- `helpers/console.ts`: new `deprecatedWarnOnce(key, message)` – prints `Deprecated: <message>` once per key per page – and `removedWarnOnce(key, message)` – same once-per-key record, no prefix, for APIs that no longer exist. The repeat-every-call `deprecatedWarn()` is removed (last caller moved to the once variant; not on `Handsontable.helper`).
- `ManualColumnResize` / `ManualRowResize`: `saveManualColumnWidths()`, `loadManualColumnWidths()`, `saveManualRowHeights()`, `loadManualRowHeights()` use it and carry versioned `@deprecated` messages.
- `exportFile` `DataProvider#setOptions`: warns once when `columnHeaders` is passed; the alias keeps working.
- `helpers/string.ts` `sanitize()`: warns once and points at the `sanitizer` option.
- `exportFile` `Settings` and `dataProvider` `DataProviderOptions`: versioned `@deprecated` messages.
- `core.ts` `updateSettings()`: warn-once guard for removed options (`REMOVED_OPTIONS`). It scans the top level, the array-form `columns`, and the declarative `cell` array. Each option links to the docs of the release that removed it.
- Removed the unused `deprecationWarns` sets in `core.ts` and `translations/indexMapper.ts`.

Docs changes:

- `deprecation-policy.md`: lists the APIs removed in 18.0 next to the removed dependencies, adds a "Removed in version 17.0" section (`PersistentState`, legacy `hot.undo()` family), and adds the missing current deprecations with "Deprecated in" / "Removal" columns.
- `CHANGELOG.md`, `changelog-18.md`, `changelog/changelog.md`: the 18.0.0 PersistentState removal entry linked `pull/0` (it is #12727), and now states that the published 17.0 release already shipped without the plugin (#12015), so it no longer contradicts the "Removed in version 17.0" section.
- `.ai/BREAKING-CHANGES.md` (deprecation checklist; "next stable release" → "next major release"), `handsontable/.ai/CONVENTIONS.md`, `handsontable/.ai/HOOKS.md`.

Version facts verified against git history:

- `columnHeaders` was renamed and deprecated in **17.1.0** (#12224, commit `53df95b073`).
- `persistentState` (plugin, option, hooks) was removed in **17.0.0** (#12015). The TypeScript conversion (#12011) re-created the plugin on `develop` after 17.1.0; #12727 deleted that copy before 18.0.0. No published 17.x or 18.x package carries it, so the 18.0.0 changelog line describes a develop-only cleanup. The `REMOVED_OPTIONS` JSDoc records this.

Out of scope, on purpose: removing any deprecated code (19.0 only) and the migration guides (left untouched). The released 18.0.0 changelog entry keeps its wording; it only gains the clause about the 17.0 release.

[skip changelog] – the branch carries only deprecation metadata and warning-consistency work on APIs that are already deprecated or removed; no new user-facing deprecation. The checklist's step 5 stays as written for PRs that introduce a deprecation.

Open points for reviewers:

1. `.changelogs/12727.json` is still pending on `develop` although the entry already shipped under 18.0.0, so it will be compiled again into the 18.1 notes. Release owner to decide which version keeps it.
2. `columnHeaders` stays *deprecated* (warn + removal in 19.0), not legacy: the tag has been in the published typings since 17.1, so this PR completes a deprecation that shipped without a runtime warning.

### How has this been tested?

- Unit tests added:
  - `handsontable/src/helpers/__tests__/console.unit.ts` – `deprecatedWarnOnce` and `removedWarnOnce` warn once per key, tolerate a missing `console`, and share the once-record and the reset. The `deprecatedWarn` block is gone with the function.
  - `handsontable/src/plugins/exportFile/__tests__/exportFile.unit.js` – `columnHeaders` warns once and still maps to `colHeaders`.
  - `handsontable/src/helpers/__tests__/string.unit.ts` – `sanitize()` warns once and stays a pass-through.
  - `handsontable/src/__tests__/core/updateSettings.unit.js` – removed options warn once (top level, `columns`, `cell`), without a `Deprecated:` prefix, with the removing version and the per-release docs URL; unaffected settings do not warn.
- Type tests extended: `exportFile.types.ts`, `dataProvider.types.ts` – deprecated aliases stay assignable; both `normalizeExportOptions()` overloads are pinned.
- Ran locally:
  ```bash
  npm --prefix handsontable run test:unit    # 322 suites, 3834 tests pass
  npm --prefix handsontable run build        # both bundles
  cd docs && npm run docs:api                # ManualColumnResize/ManualRowResize pages render the "Deprecated" tag + warning text
  ```

### Types of changes

### Related issue(s):

1. #12727 (PersistentState removal – deprecated the resize-state methods)
2. #12224 (`columnHeaders` → `colHeaders` rename)
3. #12015 (17.0.0 removal of `PersistentState` and the legacy undo/redo methods)

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

### DIFF
--- a/.ai/BREAKING-CHANGES.md
+++ b/.ai/BREAKING-CHANGES.md
@@ -11,7 +11,7 @@ The lean rule lives in the root `AGENTS.md` "Breaking changes policy" section (a
 | Renaming a CSS class produced by Handsontable | Breaks custom stylesheets | Keep the legacy class name in the DOM. Add tests verifying the old name still works. |
 | Renaming APIs (methods, configuration options, hooks) | Breaks customer integrations | Keep the legacy API working and translate it to the new API internally. Legacy APIs do not produce console warnings. |
 | Changing API signatures or behavior | Breaks customer integrations | Keep the deprecated API working until the next major release. Deprecated APIs produce a console warning (fired only once). |
-| Removing hooks or configuration options | May go undetected by customers | Add the hook to `REMOVED_HOOKS` in `handsontable/src/core/hooks/constants.ts`, or the option to `REMOVED_OPTIONS` in `handsontable/src/core.ts`, so a one-time warning shows when someone uses it in configuration. The warning names the removing version and carries no `Deprecated:` prefix (`removedWarnOnce`, not `deprecatedWarnOnce`). |
+| Removing hooks or configuration options | May go undetected by customers | Add the hook to `REMOVED_HOOKS` in `handsontable/src/core/hooks/constants.ts`, or the option to `REMOVED_OPTIONS` in `handsontable/src/core.ts`, so a warning shows when someone uses it in configuration. Both warnings name the removing version and carry no `Deprecated:` prefix. The option warning is one-time (`removedWarnOnce`, not `deprecatedWarnOnce`); the hook warning goes through plain `warn()` on every `add()` and links to the release notes (see `handsontable/.ai/HOOKS.md`). |
 | Changing a default setting value | 🚫 **Strictly forbidden** — a "really bad" breaking change. | Never change defaults. |
 
 ## Legacy vs deprecated
--- a/.ai/BREAKING-CHANGES.md
+++ b/.ai/BREAKING-CHANGES.md
@@ -19,6 +19,17 @@ The lean rule lives in the root `AGENTS.md` "Breaking changes policy" section (a
 - **Legacy**: Old API kept working forever alongside the new API. No console warnings. The legacy feature set may be frozen. Tests must verify the old name keeps working.
 - **Deprecated**: Old API works until the next stable release, then is removed. Produces a one-time console warning. Tests must verify the old name keeps working until removal.
 
+## Deprecation checklist
+
+When you deprecate a public API, do all of the following in the same PR:
+
+1. JSDoc: `@deprecated Since X.Y.Z. <reason>. It will be removed in <next major>. Use <replacement> instead.` – never a bare `@deprecated` (the API docs render the text as a warning box).
+2. Runtime (methods, options, helpers): call `deprecatedWarnOnce('<Owner>.<name>', '<message>')` from `handsontable/src/helpers/console.ts`. Types cannot warn; the JSDoc tag is enough.
+3. Tests: keep a test proving the old API still works, and one proving the warning prints once. The once-state is module-global, so put the warning test first in its `describe`.
+4. Docs: add a row to `docs/content/guides/upgrade-and-migration/deprecation-policy/deprecation-policy.md` ("List of current deprecations") and a step in the current migration guide.
+5. Changelog: `.changelogs/<PR>.json` with `"type": "deprecated"`.
+6. Removal happens only in the next major release, at least 3 months later: delete the code, move the docs row to "Removed in version N.0", add a `"type": "removed", "breaking": true` changelog entry, and add a migration step.
+
 ## What is NOT considered breaking
 
 Changes to JavaScript APIs not listed in the public API reference (e.g., internal Walkontable code that does not affect the DOM or CSS). Note such changes in release notes.
--- a/.ai/BREAKING-CHANGES.md
+++ b/.ai/BREAKING-CHANGES.md
@@ -25,7 +25,7 @@ When you deprecate a public API, do all of the following in the same PR:
 
 1. JSDoc: `@deprecated Since X.Y.Z. <reason>. It will be removed in <next major>. Use <replacement> instead.` – never a bare `@deprecated` (the API docs render the text as a warning box).
 2. Runtime (methods, options, helpers): call `deprecatedWarnOnce('<Owner>.<name>', '<message>')` from `handsontable/src/helpers/console.ts`. Types cannot warn; the JSDoc tag is enough.
-3. Tests: keep a test proving the old API still works, and one proving the warning prints once. The once-state is module-global, so put the warning test first in its `describe`.
+3. Tests: keep a test proving the old API still works, and one proving the warning prints once. The once-state is module-global, so call `_resetDeprecationWarnings()` (from `handsontable/src/helpers/console.ts`) in `beforeEach` – otherwise the assertion silently passes whenever another spec printed that warning first.
 4. Docs: add a row to `docs/content/guides/upgrade-and-migration/deprecation-policy/deprecation-policy.md` ("List of current deprecations") and a step in the current migration guide.
 5. Changelog: `.changelogs/<PR>.json` with `"type": "deprecated"`.
 6. Removal happens only in the next major release, at least 3 months later: delete the code, move the docs row to "Removed in version N.0", add a `"type": "removed", "breaking": true` changelog entry, and add a migration step.
--- a/.ai/BREAKING-CHANGES.md
+++ b/.ai/BREAKING-CHANGES.md
@@ -10,14 +10,14 @@ The lean rule lives in the root `AGENTS.md` "Breaking changes policy" section (a
 |---|---|---|
 | Renaming a CSS class produced by Handsontable | Breaks custom stylesheets | Keep the legacy class name in the DOM. Add tests verifying the old name still works. |
 | Renaming APIs (methods, configuration options, hooks) | Breaks customer integrations | Keep the legacy API working and translate it to the new API internally. Legacy APIs do not produce console warnings. |
-| Changing API signatures or behavior | Breaks customer integrations | Keep the deprecated API working until the next stable release. Deprecated APIs produce a console warning (fired only once). |
-| Removing hooks or configuration options | May go undetected by customers | Add the hook or option to the list of removed hooks so an error shows when someone uses it in configuration. |
+| Changing API signatures or behavior | Breaks customer integrations | Keep the deprecated API working until the next major release. Deprecated APIs produce a console warning (fired only once). |
+| Removing hooks or configuration options | May go undetected by customers | Add the hook to `REMOVED_HOOKS` in `handsontable/src/core/hooks/constants.ts`, or the option to `REMOVED_OPTIONS` in `handsontable/src/core.ts`, so a one-time warning shows when someone uses it in configuration. The warning names the removing version and carries no `Deprecated:` prefix (`removedWarnOnce`, not `deprecatedWarnOnce`). |
 | Changing a default setting value | 🚫 **Strictly forbidden** — a "really bad" breaking change. | Never change defaults. |
 
 ## Legacy vs deprecated
 
 - **Legacy**: Old API kept working forever alongside the new API. No console warnings. The legacy feature set may be frozen. Tests must verify the old name keeps working.
-- **Deprecated**: Old API works until the next stable release, then is removed. Produces a one-time console warning. Tests must verify the old name keeps working until removal.
+- **Deprecated**: Old API works until the next major release, then is removed. Produces a one-time console warning. Tests must verify the old name keeps working until removal.
 
 ## Deprecation checklist
 
@@ -26,7 +26,7 @@ When you deprecate a public API, do all of the following in the same PR:
 1. JSDoc: `@deprecated Since X.Y.Z. <reason>. It will be removed in <next major>. Use <replacement> instead.` – never a bare `@deprecated` (the API docs render the text as a warning box).
 2. Runtime (methods, options, helpers): call `deprecatedWarnOnce('<Owner>.<name>', '<message>')` from `handsontable/src/helpers/console.ts`. Types cannot warn; the JSDoc tag is enough.
 3. Tests: keep a test proving the old API still works, and one proving the warning prints once. The once-state is module-global, so call `_resetDeprecationWarnings()` (from `handsontable/src/helpers/console.ts`) in `beforeEach` – otherwise the assertion silently passes whenever another spec printed that warning first.
-4. Docs: add a row to `docs/content/guides/upgrade-and-migration/deprecation-policy/deprecation-policy.md` ("List of current deprecations") and a step in the current migration guide.
+4. Docs: add a row to `docs/content/guides/upgrade-and-migration/deprecation-policy/deprecation-policy.md` ("List of current deprecations") and, when the release has a migration guide (minor releases usually do not), a step in it.
 5. Changelog: `.changelogs/<PR>.json` with `"type": "deprecated"`.
 6. Removal happens only in the next major release, at least 3 months later: delete the code, move the docs row to "Removed in version N.0", add a `"type": "removed", "breaking": true` changelog entry, and add a migration step.
 
--- a/docs/content/guides/upgrade-and-migration/deprecation-policy/deprecation-policy.md
+++ b/docs/content/guides/upgrade-and-migration/deprecation-policy/deprecation-policy.md
@@ -52,20 +52,20 @@ The following dependencies were deprecated in version 17.0 and removed in versio
 
 | Removed | Description | Migration guide |
 | ------- | ----------- | --------------- |
-| **numbro.js** | Handled numeric data formatting. Replace with [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat). | [Migrate from 17.1 to 18.0 -> Numeric formatting](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#2-migrate-numeric-format-from-numbro-js-pattern-and-culture-to-intlnumberformat) |
-| **Pikaday** | Displayed a date picker. The `intl-date` cell type now uses the native browser date picker. | [Migrate from 17.1 to 18.0 -> Date/Time](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#3-migrate-date-and-time-cells-to-iso-8601-format) |
-| **moment.js** | Parsed, validated, and displayed dates. The `intl-date` and `intl-time` cell types use the native `Intl.DateTimeFormat` API. | [Migrate from 17.1 to 18.0 -> Date/Time](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#3-migrate-date-and-time-cells-to-iso-8601-format) |
-| **DOMPurify** | An XSS sanitizer for HTML. Use the `sanitizer` option to provide your own sanitizer function. | [Migrate from 17.1 to 18.0 -> HTML sanitization](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#4-replace-built-in-dompurify-with-a-custom-sanitizer) |
+| **numbro.js** | Handled numeric data formatting. Replace with [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat). | [Migrate from 17.1 to 18.0 -> Numeric formatting](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#migrate-numeric-format-from-numbrojs-pattern-and-culture-to-intlnumberformat) |
+| **Pikaday** | Displayed a date picker. The `intl-date` cell type now uses the native browser date picker. | [Migrate from 17.1 to 18.0 -> Date/Time](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#migrate-date-and-time-cells-to-iso-8601-format) |
+| **moment.js** | Parsed, validated, and displayed dates. The `intl-date` and `intl-time` cell types use the native `Intl.DateTimeFormat` API. | [Migrate from 17.1 to 18.0 -> Date/Time](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#migrate-date-and-time-cells-to-iso-8601-format) |
+| **DOMPurify** | An XSS sanitizer for HTML. Use the `sanitizer` option to provide your own sanitizer function. | [Migrate from 17.1 to 18.0 -> HTML sanitization](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#replace-built-in-dompurify-with-a-custom-sanitizer) |
 
 ### APIs
 
 | Removed | Deprecated in | Replacement | Migration guide |
 | ------- | ------------- | ----------- | --------------- |
-| `PersistentState` plugin, `persistentState` option, and the `persistentStateSave`, `persistentStateLoad`, and `persistentStateReset` hooks | 16.1 | Persist state in your application code. | [Migrate from 17.1 to 18.0 -> Resize-state methods](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#stop-calling-deprecated-resize-state-methods) |
+| `persistentState` option and the `persistentStateSave`, `persistentStateLoad`, and `persistentStateReset` hooks. The `PersistentState` plugin itself was removed in 17.0. | 16.1 | Persist state in your application code. | [Changelog 18.0](@/guides/upgrade-and-migration/changelog-18/changelog-18.md) |
 | `hot.undo()`, `hot.redo()`, `hot.clearUndo()`, `hot.isUndoAvailable()`, `hot.isRedoAvailable()`, and `hot.undoRedo` | Before 16.0 | `hot.getPlugin('undoRedo')` | [Changelog 18.0](@/guides/upgrade-and-migration/changelog-18/changelog-18.md) |
-| `correctFormat` and `datePickerConfig` options | 17.0 | `valueParser` and `valueSetter` | [Migrate from 17.1 to 18.0 -> Date/Time](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#3-migrate-date-and-time-cells-to-iso-8601-format) |
-| `numericFormat.pattern` and `numericFormat.culture` | 17.0 | `Intl.NumberFormat` options and the `locale` option | [Migrate from 17.1 to 18.0 -> Numeric formatting](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#2-migrate-numeric-format-from-numbro-js-pattern-and-culture-to-intlnumberformat) |
-| `handsontable/common` import subpath | -- (never documented) | `handsontable` or `handsontable/base` | [Migrate from 17.1 to 18.0 -> Imports](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#1-replace-handsontable-common-imports) |
+| `correctFormat` and `datePickerConfig` options | 17.0 | `valueParser` and `valueSetter` | [Migrate from 17.1 to 18.0 -> `correctFormat` option](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#correctformat-option) |
+| `numericFormat.pattern` and `numericFormat.culture` | 17.0 | `Intl.NumberFormat` options and the `locale` option | [Migrate from 17.1 to 18.0 -> Numeric formatting](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#migrate-numeric-format-from-numbrojs-pattern-and-culture-to-intlnumberformat) |
+| `handsontable/common` import subpath | -- (never documented) | `handsontable` or `handsontable/base` | [Migrate from 17.1 to 18.0 -> Imports](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#replace-handsontablecommon-imports) |
 
 ## List of current deprecations
 
--- a/docs/content/guides/upgrade-and-migration/deprecation-policy/deprecation-policy.md
+++ b/docs/content/guides/upgrade-and-migration/deprecation-policy/deprecation-policy.md
@@ -61,11 +61,18 @@ The following dependencies were deprecated in version 17.0 and removed in versio
 
 | Removed | Deprecated in | Replacement | Migration guide |
 | ------- | ------------- | ----------- | --------------- |
-| `persistentState` option and the `persistentStateSave`, `persistentStateLoad`, and `persistentStateReset` hooks. The `PersistentState` plugin itself was removed in 17.0. | 16.1 | Persist state in your application code. | [Changelog 18.0](@/guides/upgrade-and-migration/changelog-18/changelog-18.md) |
-| `hot.undo()`, `hot.redo()`, `hot.clearUndo()`, `hot.isUndoAvailable()`, `hot.isRedoAvailable()`, and `hot.undoRedo` | Before 16.0 | `hot.getPlugin('undoRedo')` | [Changelog 18.0](@/guides/upgrade-and-migration/changelog-18/changelog-18.md) |
 | `correctFormat` and `datePickerConfig` options | 17.0 | `valueParser` and `valueSetter` | [Migrate from 17.1 to 18.0 -> `correctFormat` option](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#correctformat-option) |
 | `numericFormat.pattern` and `numericFormat.culture` | 17.0 | `Intl.NumberFormat` options and the `locale` option | [Migrate from 17.1 to 18.0 -> Numeric formatting](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#migrate-numeric-format-from-numbrojs-pattern-and-culture-to-intlnumberformat) |
 | `handsontable/common` import subpath | -- (never documented) | `handsontable` or `handsontable/base` | [Migrate from 17.1 to 18.0 -> Imports](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#replace-handsontablecommon-imports) |
+
+## Removed in version 17.0
+
+### APIs
+
+| Removed | Deprecated in | Replacement | Migration guide |
+| ------- | ------------- | ----------- | --------------- |
+| `PersistentState` plugin, its `persistentState` option, and the `persistentStateSave`, `persistentStateLoad`, and `persistentStateReset` hooks | 16.1 | Persist state in your application code. The `saveManualColumnWidths()`, `loadManualColumnWidths()`, `saveManualRowHeights()`, and `loadManualRowHeights()` methods that relied on the plugin are deprecated (see below). | [Changelog 17.0](@/guides/upgrade-and-migration/changelog-17/changelog-17.md) |
+| `hot.undo()`, `hot.redo()`, `hot.clearUndo()`, `hot.isUndoAvailable()`, `hot.isRedoAvailable()`, and `hot.undoRedo` | Before 16.0 | `hot.getPlugin('undoRedo')` | [Changelog 17.0](@/guides/upgrade-and-migration/changelog-17/changelog-17.md) |
 
 ## List of current deprecations
 
@@ -85,7 +92,7 @@ The following APIs are deprecated. They keep working and print a one-time consol
 
 | Deprecated | Deprecated in | Removal | Replacement | Migration guide |
 | ---------- | ------------- | ------- | ----------- | --------------- |
-| `columnHeaders` (export options of the `exportFile` plugin) | 17.0 | 19.0 | `colHeaders` | [Export to CSV](@/guides/accessories-and-menus/export-to-csv/export-to-csv.md) |
+| `columnHeaders` (export options of the `exportFile` plugin) | 17.1 | 19.0 | `colHeaders` – rename the key; the value keeps its meaning. | [Export to CSV -> Available options](@/guides/accessories-and-menus/export-to-csv/export-to-csv.md#available-options-in-the-export-configuration) |
 
 ### TypeScript types
 
--- a/docs/content/guides/upgrade-and-migration/deprecation-policy/deprecation-policy.md
+++ b/docs/content/guides/upgrade-and-migration/deprecation-policy/deprecation-policy.md
@@ -92,7 +92,7 @@ The following APIs are deprecated. They keep working and print a one-time consol
 
 | Deprecated | Deprecated in | Removal | Replacement | Migration guide |
 | ---------- | ------------- | ------- | ----------- | --------------- |
-| `columnHeaders` (export options of the `exportFile` plugin) | 17.1 | 19.0 | `colHeaders` – rename the key; the value keeps its meaning. | [Export to CSV -> Available options](@/guides/accessories-and-menus/export-to-csv/export-to-csv.md#available-options-in-the-export-configuration) |
+| `columnHeaders` (export options of the `exportFile` plugin) | 17.1 | 19.0 | `colHeaders` -- rename the key; the value keeps its meaning. | [Export to CSV -> Available options](@/guides/accessories-and-menus/export-to-csv/export-to-csv.md#available-options-in-the-export-configuration) |
 
 ### TypeScript types
 
--- a/docs/content/guides/upgrade-and-migration/deprecation-policy/deprecation-policy.md
+++ b/docs/content/guides/upgrade-and-migration/deprecation-policy/deprecation-policy.md
@@ -46,27 +46,55 @@ For more details about our versioning policy please visit [Versioning policy](@/
 
 ## Removed in version 18.0
 
+### Dependencies
+
 The following dependencies were deprecated in version 17.0 and removed in version 18.0.
 
 | Removed | Description | Migration guide |
 | ------- | ----------- | --------------- |
-| **numbro.js** | Handled numeric data formatting. Replace with [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat). | [Migrate from 17.1 to 18.0 → Numeric formatting](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#2-migrate-numeric-format-from-numbro-js-pattern-and-culture-to-intlnumberformat) |
-| **Pikaday** | Displayed a date picker. The `intl-date` cell type now uses the native browser date picker. | [Migrate from 17.1 to 18.0 → Date/Time](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#3-migrate-date-and-time-cells-to-iso-8601-format) |
-| **moment.js** | Parsed, validated, and displayed dates. The `intl-date` and `intl-time` cell types use the native `Intl.DateTimeFormat` API. | [Migrate from 17.1 to 18.0 → Date/Time](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#3-migrate-date-and-time-cells-to-iso-8601-format) |
-| **DOMPurify** | An XSS sanitizer for HTML. Use the `sanitizer` option to provide your own sanitizer function. | [Migrate from 17.1 to 18.0 → HTML sanitization](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#4-replace-built-in-dompurify-with-a-custom-sanitizer) |
+| **numbro.js** | Handled numeric data formatting. Replace with [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat). | [Migrate from 17.1 to 18.0 -> Numeric formatting](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#2-migrate-numeric-format-from-numbro-js-pattern-and-culture-to-intlnumberformat) |
+| **Pikaday** | Displayed a date picker. The `intl-date` cell type now uses the native browser date picker. | [Migrate from 17.1 to 18.0 -> Date/Time](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#3-migrate-date-and-time-cells-to-iso-8601-format) |
+| **moment.js** | Parsed, validated, and displayed dates. The `intl-date` and `intl-time` cell types use the native `Intl.DateTimeFormat` API. | [Migrate from 17.1 to 18.0 -> Date/Time](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#3-migrate-date-and-time-cells-to-iso-8601-format) |
+| **DOMPurify** | An XSS sanitizer for HTML. Use the `sanitizer` option to provide your own sanitizer function. | [Migrate from 17.1 to 18.0 -> HTML sanitization](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#4-replace-built-in-dompurify-with-a-custom-sanitizer) |
+
+### APIs
+
+| Removed | Deprecated in | Replacement | Migration guide |
+| ------- | ------------- | ----------- | --------------- |
+| `PersistentState` plugin, `persistentState` option, and the `persistentStateSave`, `persistentStateLoad`, and `persistentStateReset` hooks | 16.1 | Persist state in your application code. | [Migrate from 17.1 to 18.0 -> Resize-state methods](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#stop-calling-deprecated-resize-state-methods) |
+| `hot.undo()`, `hot.redo()`, `hot.clearUndo()`, `hot.isUndoAvailable()`, `hot.isRedoAvailable()`, and `hot.undoRedo` | Before 16.0 | `hot.getPlugin('undoRedo')` | [Changelog 18.0](@/guides/upgrade-and-migration/changelog-18/changelog-18.md) |
+| `correctFormat` and `datePickerConfig` options | 17.0 | `valueParser` and `valueSetter` | [Migrate from 17.1 to 18.0 -> Date/Time](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#3-migrate-date-and-time-cells-to-iso-8601-format) |
+| `numericFormat.pattern` and `numericFormat.culture` | 17.0 | `Intl.NumberFormat` options and the `locale` option | [Migrate from 17.1 to 18.0 -> Numeric formatting](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#2-migrate-numeric-format-from-numbro-js-pattern-and-culture-to-intlnumberformat) |
+| `handsontable/common` import subpath | -- (never documented) | `handsontable` or `handsontable/base` | [Migrate from 17.1 to 18.0 -> Imports](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#1-replace-handsontable-common-imports) |
 
 ## List of current deprecations
 
-The following methods were deprecated in version 18.0 and are scheduled for removal in a future major release. They are no-ops kept for backward compatibility, as the `PersistentState` plugin has been removed.
+The following APIs are deprecated. They keep working and print a one-time console warning. They are scheduled for removal in version 19.0, the next major release.
 
-| Deprecated | Plugin | Migration guide |
-| ---------- | ------ | --------------- |
-| `saveManualColumnWidths()` | `ManualColumnResize` | [Migrate from 17.1 to 18.0 → Resize-state methods](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#stop-calling-deprecated-resize-state-methods) |
-| `loadManualColumnWidths()` | `ManualColumnResize` | [Migrate from 17.1 to 18.0 → Resize-state methods](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#stop-calling-deprecated-resize-state-methods) |
-| `saveManualRowHeights()` | `ManualRowResize` | [Migrate from 17.1 to 18.0 → Resize-state methods](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#stop-calling-deprecated-resize-state-methods) |
-| `loadManualRowHeights()` | `ManualRowResize` | [Migrate from 17.1 to 18.0 → Resize-state methods](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#stop-calling-deprecated-resize-state-methods) |
+### Methods
 
+| Deprecated | Deprecated in | Removal | Replacement | Migration guide |
+| ---------- | ------------- | ------- | ----------- | --------------- |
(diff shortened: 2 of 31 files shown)


handsontable/src/core.ts
```diff
@@ -3378,6 +3382,14 @@ export default function Core(
       throwWithCause('Since 8.0.0 the "ganttChart" setting is no longer supported.');
     }
 
+    REMOVED_OPTIONS.forEach(([option, version]) => {
+      if (isDefined((settings as Record<string, unknown>)[option])) {
+        deprecatedWarnOnce(`Core.removedOption.${option}`,
+          `The "${option}" setting was removed in Handsontable ${version} and is ignored. ` +
+          'See https://handsontable.com/docs/migration-from-17.1-to-18.0/ for the migration path.');
+      }
+    });
```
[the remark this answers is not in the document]
adrianspdev: Confirmed – a top-level `datePickerConfig` produced two different messages. Dropped `datePickerConfig` from `REMOVED_OPTIONS` in e6ca9e23e; `DateEditor#prepare` keeps the single warning since it also covers the column- and cell-level forms that the `updateSettings` check cannot see. Added a unit test asserting `updateSettings` stays silent for it.

handsontable/src/core.ts
```diff
@@ -3378,6 +3384,14 @@ export default function Core(
       throwWithCause('Since 8.0.0 the "ganttChart" setting is no longer supported.');
     }
 
+    REMOVED_OPTIONS.forEach(([option, version]) => {
+      if (isDefined((settings as Record<string, unknown>)[option])) {
+        deprecatedWarnOnce(`Core.removedOption.${option}`,
+          `The "${option}" setting was removed in Handsontable ${version} and is ignored. ` +
+          'See https://handsontable.com/docs/migration-from-17.1-to-18.0/ for the migration path.');
```
budnix: This URL 404s. The docs site needs the framework segment - I checked both forms live: `https://handsontable.com/docs/javascript-data-grid/migration-from-17.1-to-18.0/` returns 200, the bare `/docs/migration-from-...` returns 404. Same shape as the doc links in `baseEditor.ts`.

```suggestion
          'See https://handsontable.com/docs/javascript-data-grid/migration-from-17.1-to-18.0/ for the migration path.');
```

handsontable/src/core.ts
```diff
@@ -142,12 +142,18 @@ function normalizeIndexesGroup(indexes: number[][]): number[][] {
...
+ * `datePickerConfig` is deliberately absent: `DateEditor#prepare` already warns about it, and
+ * it also sees the column- and cell-level forms that this top-level check cannot.
+ *
+ * @type {Array<Array<string>>}
  */
-
-const deprecationWarns = new Set();
+const REMOVED_OPTIONS: ReadonlyArray<[string, string]> = [
+  ['persistentState', '18.0.0'],
+  ['correctFormat', '18.0.0'],
```
budnix: `correctFormat` is a cell option, so most people set it inside `columns` or on a cell, not at the top level. This check only reads top-level settings, so `columns: [{ type: 'date', correctFormat: true }]` prints nothing - the user sees a clean console and quietly loses date auto-correction. Grepping `src` shows this is now the only mention of `correctFormat` left, so unlike `datePickerConfig` there is no editor-level warning to fall back on. Scanning `settings.columns` too is more work - curious whether you think it belongs here or in a follow-up.

handsontable/src/helpers/console.ts
```diff
@@ -80,6 +80,30 @@ export function deprecatedWarn(message: string) {
...
+ *
+ * @param {string} key A stable identifier of the deprecated API (for example, the method name).
+ * @param {string} message The message to log.
+ */
+export function deprecatedWarnOnce(key: string, message: string): void {
+  if (printedDeprecations.has(key)) {
+    return;
+  }
+
+  printedDeprecations.add(key);
```
budnix, line 97: The key gets recorded before `deprecatedWarn` checks whether `console` exists. If the first call lands while `console` is missing or stubbed - the case that guard exists for, and the one `console.unit.ts` now tests - the key is burned with nothing printed, and every later call returns early. `warnOnce` above has the same shape, but that one is scoped per grid, so this module-global Set makes the loss permanent for the whole page.

```suggestion
  if (!isDefined(console)) {
    return;
  }

  printedDeprecations.add(key);
```

handsontable/src/helpers/string.ts
```diff
@@ -1,4 +1,5 @@
 import { stringify } from './mixed';
+import { deprecatedWarnOnce } from './console';
```
budnix, line 2: This makes `string.ts` and `console.ts` import each other - `console.ts:1` already pulls `substitute` from here. It works today only because both uses sit inside function bodies, but `console.ts` has two module-level initializers (`warnedScopes`, `printedDeprecations`), so the first top-level call to `substitute` there would leave one module reading an uninitialized binding and throw at import time, not at call time. `import/no-cycle` is off in `.eslintrc.js`, so nothing will flag it. Would it be simpler to warn from `sanitize`'s callers, or to move `substitute` out of `console.ts`?

handsontable/src/plugins/exportFile/dataProvider.ts
```diff
@@ -60,10 +61,16 @@ class DataProvider {
...
   setOptions(options: Record<string, unknown>) {
-    if (options && 'columnHeaders' in options && !('colHeaders' in options)) {
-      this.options = { ...options, colHeaders: options.columnHeaders };
+    if (options && 'columnHeaders' in options) {
+      deprecatedWarnOnce('ExportFile.columnHeaders',
+        'The `columnHeaders` export option is deprecated and will be removed in Handsontable 19.0.0. ' +
+        'Use `colHeaders` instead.');
 
-      return;
+      if (!('colHeaders' in options)) {
```
budnix: I don't think this inner promotion can run in practice. `BaseType.DEFAULT_OPTIONS` always carries `colHeaders: false` and `_mergeOptions` merges it in before calling `setOptions`, so `'colHeaders' in options` is already true by the time we get here - only a test building `new DataProvider({})` directly reaches it. The alias is really resolved in `_base.ts` `_mergeOptions` and in `xlsx.ts:290`, and neither warns, so the warning above fires only because the stale `columnHeaders` key rides along in the merged object. Worth folding the three copies into one helper that promotes and warns in the same place?

handsontable/src/__tests__/core/updateSettings.unit.js
```diff
@@ -0,0 +1,80 @@
...
+      const warnSpy = jest.spyOn(console, 'warn').mockImplementation(() => {});
+      const core = new Core(container, {
+        data: [['a']],
+        licenseKey: 'non-commercial-and-evaluation',
+      });
+
+      core.init();
+      core.updateSettings({ colHeaders: true });
+
+      const calls = warnSpy.mock.calls.filter(([message]) => String(message).includes('was removed in Handsontable'));
```
budnix, line 61: Since `printedDeprecations` is module-global and the test above already burned `Core.removedOption.persistentState`, this filter can no longer catch a `persistentState` regression - only the still-unused `correctFormat` key is really under test here. The new tests in `string.unit.ts` and `exportFile.unit.js` have the same dependency, which is why the checklist you added tells people to put the warning test first. Did you look at exporting a test-only reset for that Set? It would drop the ordering rule and stop a future reorder from turning this into a silent pass.

handsontable/src/plugins/manualColumnResize/manualColumnResize.ts
```diff
@@ -226,25 +226,29 @@ export class ManualColumnResize extends BasePlugin {
...
+   * Deprecated. The `PersistentState` plugin has been removed. This method is a no-op.
    *
-   * @deprecated
+   * @deprecated Since 18.0.0. The `PersistentState` plugin was removed in 17.0.0, so this method
+   * does nothing. It will be removed in 19.0.0. Persist column widths yourself with
+   * the `afterColumnResize` hook and the `manualColumnResize` option.
    */
   saveManualColumnWidths(): void {
-    deprecatedWarn('`saveManualColumnWidths()` is deprecated and will be removed in a future major release. ' +
+    deprecatedWarnOnce('ManualColumnResize.saveManualColumnWidths',
```
budnix, line 236: The four resize methods moved from `deprecatedWarn` to `deprecatedWarnOnce`, but I couldn't find a test covering the warning. Grepping `handsontable/test`, `tests/` and `wrappers/` for the four method names only returns the two `.types.ts` compile-only files. The checklist this PR adds asks for a test proving the warning prints once, and right now removing the call from any of the four would break nothing. `manualRowResize.ts` has the same gap.

docs/content/guides/upgrade-and-migration/deprecation-policy/deprecation-policy.md
```diff
@@ -46,27 +46,55 @@ For more details about our versioning policy please visit [Versioning policy](@/
...
+| **moment.js** | Parsed, validated, and displayed dates. The `intl-date` and `intl-time` cell types use the native `Intl.DateTimeFormat` API. | [Migrate from 17.1 to 18.0 -> Date/Time](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#3-migrate-date-and-time-cells-to-iso-8601-format) |
+| **DOMPurify** | An XSS sanitizer for HTML. Use the `sanitizer` option to provide your own sanitizer function. | [Migrate from 17.1 to 18.0 -> HTML sanitization](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#4-replace-built-in-dompurify-with-a-custom-sanitizer) |
+
+### APIs
+
+| Removed | Deprecated in | Replacement | Migration guide |
+| ------- | ------------- | ----------- | --------------- |
+| `PersistentState` plugin, `persistentState` option, and the `persistentStateSave`, `persistentStateLoad`, and `persistentStateReset` hooks | 16.1 | Persist state in your application code. | [Migrate from 17.1 to 18.0 -> Resize-state methods](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#stop-calling-deprecated-resize-state-methods) |
```
budnix: This row covers the `PersistentState` plugin, the `persistentState` option and the three hooks, but the link points at the section about `saveManualColumnWidths()` and friends - which is already the target of all four rows in the Methods table below. Someone landing here to migrate their persisted state gets guidance for a different API. Is there a better section to link, or should this one point at the 18.0 changelog instead?

handsontable/src/core.ts
adrianspdev: Confirmed and fixed in b2990375c. I checked the three candidate forms with curl:

- `/docs/javascript-data-grid/migration-from-17.1-to-18.0/` -> 200
- `/docs/migration-from-17.1-to-18.0/` -> 404
- `/docs/javascript-data-grid/migrating-from-17.1-to-18.0/` -> 404

So the framework segment is required and the URL slug is `migration-`, not the directory name `migrating-`. Took your suggestion verbatim. There is a spec asserting the framework segment is present, so a future edit that drops it fails.

handsontable/src/core.ts
adrianspdev: Belongs here, done in b2990375c. You are right that the top-level-only check was close to useless for this option: `correctFormat` is a cell option, so `columns: [{ correctFormat: true }]` is its *common* form, and `grep` confirms core.ts is now the only mention left in the source, so there is no editor-level warning to fall back on the way `datePickerConfig` has one.

`warnAboutRemovedOptions` now scans the top level plus every entry of an array-form `columns`, still warning once per option per page. The per-cell forms - a `cells` function, or meta set through `setCellMeta` - stay uncovered; reaching those means inspecting resolved meta for every cell on every render, which costs more than the warning is worth. The JSDoc states that limit rather than leaving it implicit.

Four specs cover it: the column form, once-only across several columns, no warning when `columns` is a function, and the URL assertion. I verified they are not hollow by reverting the scan to top-level-only first - the two column specs go red.

handsontable/src/helpers/console.ts
adrianspdev: Real bug, fixed in b7c21d3ee with your suggestion. The guard now runs before `printedDeprecations.add(key)`.

Added a spec that stubs `console` away, calls `deprecatedWarnOnce`, restores `console`, then calls again and expects the warning - it fails on the old ordering, so the fix is pinned.

Left `warnOnce` alone deliberately, for the reason you gave: it is scoped per grid, so the loss is recoverable there rather than permanent for the page. Happy to change it if you would rather have one shape across both.

handsontable/src/helpers/string.ts
adrianspdev: Fixed in 674dbddfb by moving `substitute` out, rather than warning from `sanitize`s callers - `sanitize` is a public helper, so its callers are user code.

`substitute` now lives in `helpers/templateString`, a leaf module that both `helpers/string` and `helpers/console` import. `helpers/string` re-exports it, so its documented public path and the `typeof import("./helpers/string")` helper typing are unchanged.

On your point that nothing will flag a regression, I measured it before recommending anything, because my first attempt at measuring was wrong. I tried `eslint --rule import/no-cycle` and got a clean result - which was worthless: the rule never fired even with the cycle deliberately put back, because the TS override block in `.eslintrc.js` turns all `import/*` rules off and beats the CLI flag. Re-measured with a standalone Tarjan SCC pass over `handsontable/src`:

- `src/helpers/` - 1 cycle before this fix (exactly the one you found), 0 after.
- `handsontable/src` overall - 54 direct two-file cycles, 257 files across 11 strongly-connected groups, largest tangle 177 files. Almost all in walkontable, and many are likely type-only imports that are erased at build.

So turning the rule on repo-wide is separate work, not a review fix. The cheap option, if you want regression protection now, is a scoped override enabling `import/no-cycle` for `src/helpers/**` only - that subtree is at zero, so it would pass today. Say the word and I will add it in this PR.

handsontable/src/plugins/exportFile/dataProvider.ts
adrianspdev: Confirmed on all counts, folded into one helper in 94cd9330b.

Your read of why the inner branch is dead is exactly right: `BaseType.DEFAULT_OPTIONS` carries `colHeaders: false` (`_base.ts:22`) and `_mergeOptions` merges it before calling `setOptions`, so `"colHeaders" in options` is already true by then, and the warning only fired because the stale `columnHeaders` key rode along in the merged object.

`normalizeExportOptions()` in `exportFile/utils.ts` promotes and warns in the same place, and all three entry points call it before the merge with the defaults: `_base._mergeOptions`, `xlsx.ts` per-sheet config, and `DataProvider#setOptions` (now a one-liner). The unreachable branch is gone, and the per-sheet `xlsx` path - which promoted silently before - now warns like the others.

Tests cover the helper directly plus `BaseType#_mergeOptions`, including the "explicit `colHeaders` wins" and "no copy when the alias is absent" cases.

handsontable/src/__tests__/core/updateSettings.unit.js
adrianspdev: Yes - exported one in b7c21d3ee. `_resetDeprecationWarnings()` in `helpers/console.ts`, called in `beforeEach` in all five specs that assert on a deprecation warning.

That drops the ordering rule, as you predicted, so I also removed the "put the warning test first" line from the checklist in `.ai/BREAKING-CHANGES.md` - it existed only to work around the missing reset - and documented the helper in `handsontable/.ai/CONVENTIONS.md` instead.

On exposure: `helpers/console` is not re-exported from `index.ts` or `base.ts`, so this is not on the `Handsontable.helper` surface. It does appear in the deep-import typings at `tmp/helpers/console.d.ts`, which matches the existing `_`-prefixed internals in `helpers/mixed` (`_injectProductInfo`, `_getLicenseState`, and others).

handsontable/src/plugins/manualColumnResize/manualColumnResize.ts
adrianspdev: Correct - the only references were the two compile-only `.types.ts` files. Added in 708652e51: `manualColumnResize.unit.js` and `manualRowResize.unit.js`, covering all four methods across both plugins as you noted.

Each method gets a spec asserting it warns once however many times it is called, plus one for its return value (`undefined` / `[]`). The bodies read no state, so calling them on the prototype keeps these unit tests rather than dragging in a Core instance.

Checked they are not hollow: deleting the `deprecatedWarnOnce` call from `saveManualColumnWidths` turns two specs red.

docs/content/guides/upgrade-and-migration/deprecation-policy/deprecation-policy.md
adrianspdev: Good catch, and it turned out to hide a second problem. Fixed in fc2afb955.

The row now points at the 18.0 changelog, matching the `hot.undo()` row directly below it. Adding a `PersistentState` section to an already-shipped migration guide felt like the wrong scope for this PR.

While checking which section to link, I found the row also over-claimed what 18.0 removed. The changelog records the *plugin* going in 17.0 (#12015); 18.0 removed the leftover `persistentState` option and the three hooks (#12727). The migration guide already said 17.0 at line 324, so my new row was the inaccurate one, not the guide. Reworded to name the option and hooks, with the plugins 17.0 removal called out.

One more thing in the same file, which inverts something I first got wrong. I assumed the numeric prefix belonged in these anchors and that the resize links were broken. The published page says the opposite - the docs site strips the leading list number, so `#stop-calling-deprecated-resize-state-methods` was always right and every `#N-` link in the file resolved to nothing. Corrected all of them against the anchors on the live page, including two slugs that also differ in punctuation (`#migrate-numeric-format-from-numbrojs-...`, `#replace-handsontablecommon-imports`). `correctFormat` and `datePickerConfig` have their own section, so that row now links straight to `#correctformat-option`.

budnix (dismissed):

handsontable/src/plugins/exportFile/exportFile.ts
```diff
@@ -29,7 +29,10 @@ export interface SheetOptions {
    */
   colHeaders?: boolean;
   /**
-   * @deprecated Use `colHeaders` instead.
+   * Include column headers.
+   *
+   * @deprecated Since 17.0.0, renamed to `colHeaders`. Still accepted as an alias; it will be
```
demtario: **High.** Wrong version: `columnHeaders` was renamed and deprecated in **17.1.0**, not 17.0.0.

Commit `53df95b073` ("Change the columnHeaders property name to colHeaders", #12224) sits under `## [17.1.0] - 2026-05-19` in `CHANGELOG.md`, and its earliest non-rc containing tag is `17.1.0`. The same commit added the `@deprecated` tag, as a `+` line in the then hand-written typings:

```diff
+  colHeaders?: boolean;
+  /** @deprecated Use `colHeaders` instead. */
   columnHeaders?: boolean;
```

So the rename and the deprecation are the same release. `18e73ccca1` (the TypeScript conversion, 18.0.0) only carried the existing tag from `.d.ts` into `src`.

Backdating it to 17.0 shortens the documented support window by a minor. Fix: `Since 17.1.0`. Same text at line 177 and in [`deprecation-policy.md:88`](https://github.com/handsontable/handsontable/blob/fc2afb955998047dd7c21232cfe18658b6e31a76/docs/content/guides/upgrade-and-migration/deprecation-policy/deprecation-policy.md#L88).

demtario (commented):
Review of the deprecation-metadata changes. Ten inline comments above. Two findings have no line in the diff to hang on, so they are here.

**Medium. `[skip changelog]` contradicts the checklist this PR adds.**

Step 5 of the new deprecation checklist in `.ai/BREAKING-CHANGES.md` mandates `.changelogs/<PR>.json` with `"type": "deprecated"`. This PR ships `[skip changelog]` instead, and the omission was deliberate (`7f4476d09` "Add changelog entry for PR #13249" followed by `68fed0845` "Remove changelog entry for PR #13249").

The stated justification, "internal consistency work on already-deprecated APIs; no new user-facing deprecation", holds for the JSDoc half but not the runtime half. On upgrade to 18.1, a user who changes nothing starts seeing `Deprecated:` lines from `sanitize()`, the four resize methods, and the `columnHeaders` export option, plus `persistentState` and `correctFormat` warnings that did not exist in 18.0. The four resize methods also change from warn-every-call to warn-once, which is the fix the PR description leads with.

Either add `.changelogs/13249.json` with `"type": "changed"` or `"type": "deprecated"` and drop `[skip changelog]`, or amend step 5 to exempt formalizing an already-tagged deprecation. As it stands the PR is the first counterexample to its own rule.

**Resolved, no action needed.** This PR introduces no new deprecations; it only makes the existing ones more visible. The checklist step does not apply, so `[skip changelog]` is correct here.

**Low. `handsontable/src/helpers/mixed.ts:719-721` now cites a deleted import edge.**

That comment justifies using the global `console` with: "`console` imports `substitute` from `string`, and `string` imports from here". This PR moves `substitute` to `helpers/templateString.ts`, so that edge no longer exists. The conclusion still holds through the surviving edge (`console` imports `isDefined` from `mixed`), but the stated reason is now false. The new `templateString.ts` header comment describes the graph correctly, so it is only this one that drifted.

**Open point 2 in the description, answered.** Keep `columnHeaders` as a deprecation with a 19.0 removal rather than flipping it to legacy. It is not a plain untagged rename: it has carried `@deprecated` in the published typings since 17.1 (see the inline comment on `exportFile.ts:34`), so every 17.1+ TypeScript user has had it struck through for two releases. This PR completes a deprecation that shipped incomplete, tag present but no runtime warning and no removal target, which is defensible under the policy. Two things worth making explicit rather than leaving inferred: the 19.0 removal is a scheduled breaking change and deserves sign-off, and what makes 19.0 correct instead of 18.0 is step 6 of the policy text this PR itself adds.

**Checklist step 4** ("a step in the current migration guide") does not fire for the four resize methods, `sanitize()`, `Settings`, or `DataProviderOptions`, which were deprecated in earlier releases and are only being documented here. It does fire for `columnHeaders`: this PR adds its runtime warning for the first time, which is when the deprecation becomes visible to plain-JS consumers. Either add that one step or say why not.

**Verified clean, recorded so nobody re-derives it.** The `#N-` anchor rewrites are correct and the old anchors were the broken ones: `docs/src/plugins/rehype-migration-steps.mjs` strips the `N. ` step prefix before heading ids are assigned. `sanitize()` has zero internal callers across `handsontable/src` and `wrappers/`, so no user gets an unactionable warning. Nothing public was removed in `base.ts`, `types/_base.ts`, or `types/xlsx.ts`; the `normalizeExportOptions` extraction is behavior-identical to the inlined promotion, including "explicit `colHeaders` wins". All five dedup keys are string literals, and every affected spec calls `_resetDeprecationWarnings()` in `beforeEach`. The remaining version claims in the deprecation table (`correctFormat`, `datePickerConfig`, `numericFormat`, `sanitize`, the resize methods, `Settings`, `DataProviderOptions`) all check out against git history.

handsontable/src/plugins/exportFile/exportFile.ts
```diff
@@ -169,7 +172,10 @@ export interface ExportOptions {
...
```
demtario: **High.** Same wrong version as on the `SheetOptions` copy above: `columnHeaders` was renamed and deprecated in **17.1.0**, not 17.0.0. Fix: `Since 17.1.0`.

docs/content/guides/upgrade-and-migration/deprecation-policy/deprecation-policy.md
```diff
@@ -46,27 +46,55 @@ For more details about our versioning policy please visit [Versioning policy](@/
...
+| `saveManualRowHeights()` (`ManualRowResize`) | 18.0 | 19.0 | No-op. Persist heights in your application code. | [Migrate from 17.1 to 18.0 -> Resize-state methods](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#stop-calling-deprecated-resize-state-methods) |
+| `loadManualRowHeights()` (`ManualRowResize`) | 18.0 | 19.0 | No-op. Pass an array to `manualRowResize`. | [Migrate from 17.1 to 18.0 -> Resize-state methods](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#stop-calling-deprecated-resize-state-methods) |
+| `Handsontable.helper.sanitize()` | 18.0 | 19.0 | Pass-through. Use the `sanitizer` option. | [`sanitizer` option](@/api/options.md#sanitizer) |
+
+### Options
+
+| Deprecated | Deprecated in | Removal | Replacement | Migration guide |
+| ---------- | ------------- | ------- | ----------- | --------------- |
+| `columnHeaders` (export options of the `exportFile` plugin) | 17.0 | 19.0 | `colHeaders` | [Export to CSV](@/guides/accessories-and-menus/export-to-csv/export-to-csv.md) |
```
demtario: **High.** "Deprecated in" should be **17.1**, not 17.0.

`columnHeaders` was renamed to `colHeaders` and tagged `@deprecated` in the same commit, `53df95b073` (#12224), which shipped in 17.1.0. Matching JSDoc at [`exportFile.ts:34`](https://github.com/handsontable/handsontable/blob/fc2afb955998047dd7c21232cfe18658b6e31a76/handsontable/src/plugins/exportFile/exportFile.ts#L34) and [`:177`](https://github.com/handsontable/handsontable/blob/fc2afb955998047dd7c21232cfe18658b6e31a76/handsontable/src/plugins/exportFile/exportFile.ts#L177) carries the same error.

Minor, on the same row: the migration-guide link points at Export to CSV, which never mentions `columnHeaders` (only `colHeaders`, at line 245), so the reader gets no migration instruction from it.

handsontable/src/core.ts
```diff
@@ -142,12 +142,50 @@ function normalizeIndexesGroup(indexes: number[][]): number[][] {
...
+ * Configuring one prints a one-time warning; the value is ignored.
  *
- * @type {Set<string>}
+ * `datePickerConfig` is deliberately absent: `DateEditor#prepare` already warns about it, and it
+ * also sees the per-cell forms that `warnAboutRemovedOptions` cannot.
+ *
+ * @type {Array<Array<string>>}
  */
+const REMOVED_OPTIONS: ReadonlyArray<[string, string]> = [
+  ['persistentState', '18.0.0'],
```
demtario: **High.** `persistentState` was removed in **17.0.0**, not 18.0.0, so this warning tells users the wrong version and sends them to the wrong migration guide.

Evidence:

- `git grep -l persistentState <tag> -- handsontable/src` (excluding tests): 11 files at `16.2.0`, **0 files at `17.0.0`**.
- The repo's own removal registry agrees. At tag `18.0.0`, `handsontable/src/core/hooks/constants.ts:3681` reads `['persistentStateSave', '17.0.0']`.
- The 17.0.0 CHANGELOG entry for #12015 lists the `PersistentState` plugin under Removed.
- [`migrating-from-17.1-to-18.0.md:324`](https://github.com/handsontable/handsontable/blob/fc2afb955998047dd7c21232cfe18658b6e31a76/docs/content/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#L324): "Because PersistentState was removed in Handsontable 17.0".
- This PR's own new JSDoc in both resize plugins says "The `PersistentState` plugin was removed in 17.0.0".

The 18.0.0 CHANGELOG line is the outlier, and this PR is what gave it a real PR number.

Fix: `['persistentState', '17.0.0']`, and branch the guide URL so this option points at `migration-from-16.2-to-17.0` while `correctFormat` keeps `migration-from-17.1-to-18.0`.

docs/content/guides/upgrade-and-migration/deprecation-policy/deprecation-policy.md
```diff
@@ -46,27 +46,55 @@ For more details about our versioning policy please visit [Versioning policy](@/
...
+| **moment.js** | Parsed, validated, and displayed dates. The `intl-date` and `intl-time` cell types use the native `Intl.DateTimeFormat` API. | [Migrate from 17.1 to 18.0 -> Date/Time](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#migrate-date-and-time-cells-to-iso-8601-format) |
+| **DOMPurify** | An XSS sanitizer for HTML. Use the `sanitizer` option to provide your own sanitizer function. | [Migrate from 17.1 to 18.0 -> HTML sanitization](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#replace-built-in-dompurify-with-a-custom-sanitizer) |
+
+### APIs
+
+| Removed | Deprecated in | Replacement | Migration guide |
+| ------- | ------------- | ----------- | --------------- |
+| `persistentState` option and the `persistentStateSave`, `persistentStateLoad`, and `persistentStateReset` hooks. The `PersistentState` plugin itself was removed in 17.0. | 16.1 | Persist state in your application code. | [Changelog 18.0](@/guides/upgrade-and-migration/changelog-18/changelog-18.md) |
```
demtario: **High.** This row and the `hot.undo()` row below it are under `## Removed in version 18.0`, but both APIs were removed in **17.0**.

- `persistentState` and its hooks: 0 non-test files in `handsontable/src` at tag `17.0.0` (11 at `16.2.0`); `core/hooks/constants.ts` at 18.0.0 records `['persistentStateSave', '17.0.0']`. This row's own prose already says "The `PersistentState` plugin itself was removed in 17.0", which contradicts the heading it sits under.
- `hot.undo()` family: `undoRedo.js:340` has `this.hot.undo = () => {` at tag `16.2.0` and the assignment is gone at `17.0.0`. The 17.0.0 CHANGELOG entry for #12015 lists "the legacy undo/redo methods" under Removed.

Fix: move both rows to a "Removed in version 17.0" section, or add a per-row "Removed in" column. Their migration links should also point at 17.0 material rather than Changelog 18.0.

handsontable/src/core.ts
```diff
@@ -142,12 +142,50 @@ function normalizeIndexesGroup(indexes: number[][]): number[][] {
...
+ *
+ * The per-cell forms - a `cells` function, or meta set through `setCellMeta` - stay uncovered.
+ * Reaching them means inspecting resolved meta for every cell on every render, which costs more
+ * than the warning is worth.
+ *
+ * @param {object} settings Settings object passed to `updateSettings`.
+ */
+function warnAboutRemovedOptions(settings: Record<string, unknown>): void {
+  const columns: unknown[] = Array.isArray(settings.columns) ? settings.columns as unknown[] : [];
+  const scopes: unknown[] = [settings, ...columns];
```
demtario: **Medium.** `scopes` misses the declarative `cell` array, so the most cell-specific configuration form escapes the warning.

`settings.cell` is a first-class declarative array read from this same settings object at [`core.ts:3622`](https://github.com/handsontable/handsontable/blob/fc2afb955998047dd7c21232cfe18658b6e31a76/handsontable/src/core.ts#L3622). A user writing:

```js
hot.updateSettings({ cell: [{ row: 0, col: 0, correctFormat: true }] });
```

gets a clean console and silently dropped date auto-correction, which is exactly the failure the JSDoc above says the `columns` scan exists to prevent.

```js
const scopes = [settings, ...columns, ...(Array.isArray(settings.cell) ? settings.cell : [])];
```

The "stay uncovered" list at line 166 should then name only the `cells` function and `setCellMeta`.

.ai/BREAKING-CHANGES.md
```diff
@@ -19,6 +19,17 @@ The lean rule lives in the root `AGENTS.md` "Breaking changes policy" section (a
...
+
+1. JSDoc: `@deprecated Since X.Y.Z. <reason>. It will be removed in <next major>. Use <replacement> instead.` – never a bare `@deprecated` (the API docs render the text as a warning box).
+2. Runtime (methods, options, helpers): call `deprecatedWarnOnce('<Owner>.<name>', '<message>')` from `handsontable/src/helpers/console.ts`. Types cannot warn; the JSDoc tag is enough.
+3. Tests: keep a test proving the old API still works, and one proving the warning prints once. The once-state is module-global, so call `_resetDeprecationWarnings()` (from `handsontable/src/helpers/console.ts`) in `beforeEach` – otherwise the assertion silently passes whenever another spec printed that warning first.
+4. Docs: add a row to `docs/content/guides/upgrade-and-migration/deprecation-policy/deprecation-policy.md` ("List of current deprecations") and a step in the current migration guide.
+5. Changelog: `.changelogs/<PR>.json` with `"type": "deprecated"`.
+6. Removal happens only in the next major release, at least 3 months later: delete the code, move the docs row to "Removed in version N.0", add a `"type": "removed", "breaking": true` changelog entry, and add a migration step.
```
demtario, line 31: **Medium.** Step 6 contradicts two unchanged lines in this same file, leaving a reader two opposite rules for when deprecated code may be deleted.

- Line 20: "**Deprecated**: Old API works until the next stable release, then is removed."
- Line 13 (What-counts table): "Keep the deprecated API working until the next stable release."

18.1 is a stable release, so as written those two lines authorize removing an 18.0-deprecated API in 18.1, which is the outcome this PR's premise rejects. Line 20 is the one that governs deletion decisions, so the stale wording is the load-bearing half.

Fix: change "next stable release" to "next major release" on both lines in this PR.

handsontable/src/core.ts
```diff
@@ -142,12 +142,50 @@ function normalizeIndexesGroup(indexes: number[][]): number[][] {
...
+
+  REMOVED_OPTIONS.forEach(([option, version]) => {
+    const isUsed = scopes.some(
+      scope => isObject(scope) && isDefined((scope as Record<string, unknown>)[option])
+    );
+
+    if (isUsed) {
+      deprecatedWarnOnce(`Core.removedOption.${option}`,
+        `The "${option}" setting was removed in Handsontable ${version} and is ignored. ` +
```
demtario: **Medium.** A *removed* option is reported as *deprecated*.

`deprecatedWarnOnce` delegates to `deprecatedWarn`, which hardcodes the prefix at [`helpers/console.ts:79`](https://github.com/handsontable/handsontable/blob/fc2afb955998047dd7c21232cfe18658b6e31a76/handsontable/src/helpers/console.ts#L79), so the user sees:

```
Deprecated: The "persistentState" setting was removed in Handsontable 18.0.0 and is ignored.
```

Removed and deprecated are distinct states in the policy this PR documents, and `.ai/BREAKING-CHANGES.md` says removed options should surface an error rather than a deprecation notice.

Fix: a prefix-less `removedWarnOnce`, or parameterize the prefix on the existing helper.

handsontable/src/base.ts
```diff
@@ -318,7 +318,7 @@ declare namespace Handsontable {
   export type GridSettings = GridSettingsType;
   export type ColumnSettings = ColumnSettingsType;
   export type CellProperties = CellPropertiesType;
-  /** @deprecated Use CellProperties */
+  /** Alias of the top-level `CellMeta` export. */
```
demtario, line 321: **Low.** Un-deprecating `Handsontable.CellMeta` looks right on the merits, but it is a published-API change that this PR records nowhere.

`CellMeta` and `CellProperties` are distinct types (lines 34-35), so the old `@deprecated Use CellProperties` reads like a mislabel, and 18.0 promoted `CellMeta` to a top-level export. Still: every 17.x/18.0 user's IDE stops striking it through, and there is no changelog entry and no row in the deprecation-policy doc this PR is rewriting. Worth naming in the PR description at minimum.

Separately, this keeps the single-line `/** … */` form that CLAUDE.md bans for all `.ts`/`.mjs`/`.js` source. Both neighbors (lines 315, 317) are single-line too, so converting only this one would be inconsistent. Either convert the block or leave it.

handsontable/src/plugins/exportFile/utils.ts
```diff
@@ -1,6 +1,36 @@
...
+ * happen before the options are merged with the defaults, because `BaseType.DEFAULT_OPTIONS`
+ * carries `colHeaders`, and a merged object therefore always has the new key already.
+ *
+ * The warning prints once per page, so calling this on several code paths for one export is safe.
+ *
+ * @param {object} options Caller-supplied export options.
+ * @returns {object} The same object when `columnHeaders` is absent, otherwise a copy with
+ *   `colHeaders` filled in. An explicit `colHeaders` always wins.
+ */
+export function normalizeExportOptions<T extends Record<string, unknown>>(options: T): T {
```
demtario: **Low.** The signature does not admit `undefined`, but the body and a test both depend on it.

Line 20 guards `!options`, and [`exportFile.unit.js`](https://github.com/handsontable/handsontable/blob/fc2afb955998047dd7c21232cfe18658b6e31a76/handsontable/src/plugins/exportFile/__tests__/exportFile.unit.js) asserts `normalizeExportOptions(undefined) === undefined`. Every current TypeScript call site passes a non-optional `Record`, so it compiles, but the contract is under-declared.

Fix: `(options: T | undefined): T | undefined`.

handsontable/src/plugins/exportFile/exportFile.ts
adrianspdev: Fixed in 31b950c094: `Since 17.1.0` on both `SheetOptions` and the `DataProvider` copy, and in `deprecation-policy.md`. Confirmed against `git tag --contains 53df95b073` (earliest `17.1.0`) and the `## [17.1.0]` section of `CHANGELOG.md`.

handsontable/src/plugins/exportFile/exportFile.ts
adrianspdev: Fixed in 31b950c094 – `Since 17.1.0`.

docs/content/guides/upgrade-and-migration/deprecation-policy/deprecation-policy.md
adrianspdev: Fixed in 31b950c094: "Deprecated in" is 17.1. The link now points at the *Available options in the export configuration* section of the Export to CSV guide (the `colHeaders` row), and the Replacement column says to rename the key.

handsontable/src/core.ts
adrianspdev: Version fixed in 31b950c094: `persistentState` → `17.0.0`. Root cause of the 18.0.0 line: #12015 removed the plugin in 17.0.0, the TypeScript conversion (#12011, merged after 17.1.0) re-created `plugins/persistentState/*.ts` on `develop`, and #12727 deleted that copy before 18.0.0. No published 17.x or 18.x package carries it. The `REMOVED_OPTIONS` JSDoc now records this.

URL: not branched to `migration-from-16.2-to-17.0`. That guide has no PersistentState or undo step at all (its sections are styles, Theme API, numbro). Each entry now carries its own `migrationUrl`: `persistentState` → `/docs/javascript-data-grid/changelog-17/`, `correctFormat` → `/docs/javascript-data-grid/migration-from-17.1-to-18.0/`. The unit test asserts both URLs.

docs/content/guides/upgrade-and-migration/deprecation-policy/deprecation-policy.md
adrianspdev: Fixed in 31b950c094: both rows moved to a new `## Removed in version 17.0` section, linked to Changelog 17.0. Confirmed `this.hot.undo =` is present at `16.2.0` and gone at `17.0.0`.

handsontable/src/core.ts
adrianspdev: Fixed in 31b950c094: `scopes = [settings, ...columns, ...cells]` where `cells` is the array-form `settings.cell`. New unit test covers `cell: [{ row, col, correctFormat: true }]` (warns once across two entries). JSDoc now lists only the `cells` function and `setCellMeta` as uncovered.

.ai/BREAKING-CHANGES.md
adrianspdev: Fixed in 31b950c094: lines 13 and 20 both say "next major release".

handsontable/src/core.ts
adrianspdev: Fixed in 31b950c094: new `removedWarnOnce(key, message)` in `helpers/console.ts` – no prefix, shares the once-record and `_resetDeprecationWarnings()` with `deprecatedWarnOnce` through an internal `warnOncePerKey`. `warnAboutRemovedOptions` uses it; the message reads `The "persistentState" setting was removed in Handsontable 17.0.0 and is ignored. See <url> for the migration path.` The removed-options row in `.ai/BREAKING-CHANGES.md` now names `REMOVED_HOOKS`/`REMOVED_OPTIONS` and `removedWarnOnce`. Tests assert the prefix is absent.

handsontable/src/base.ts
adrianspdev: The un-deprecation is named in the PR description (bullet 4 under "Before this change"). No changelog entry: the branch carries only deprecation metadata, and `[skip changelog]` stays. The single-line JSDoc is left as is to match its neighbors on lines 315 and 317.

handsontable/src/plugins/exportFile/utils.ts
adrianspdev: Fixed in 31b950c094 with overloads rather than a plain `T | undefined` return: `(options: T): T` and `(options: T | undefined): T | undefined`. The plain form broke the two definite call sites (`dataProvider.ts:64`, `types/_base.ts:102`) at `test:types`. Overload signatures precede the JSDoc block, following `plugins/filters/utils.ts`.

adrianspdev:
Addressed the review in 31b950c094. Inline threads have per-item replies; the four body-level points:

**`[skip changelog]` vs. step 5** – kept `[skip changelog]`, by decision: this branch carries only deprecation metadata and warning consistency for APIs already deprecated or removed, with no new deprecation. Step 5 stays as written for PRs that introduce one. The PR description now states this explicitly.

**`helpers/mixed.ts` cycle comment** – fixed. It now cites the surviving edge (`console` imports `isDefined` from `mixed`).

**Open point 2 (`columnHeaders`)** – agreed. Stays deprecated with a 19.0 removal; the PR description records the reasoning (tag in published typings since 17.1).

**Step 4 for `columnHeaders`** – there is no 18.0 → 18.1 migration guide to add a step to (minor releases do not get one). Step 4 in `.ai/BREAKING-CHANGES.md` now reads: a migration step is required "when the release has a migration guide (minor releases usually do not)". The deprecation-policy row remains the documented place for a minor-release deprecation.

Also found while verifying the `persistentState` version: #12015 removed the plugin in 17.0.0, #12011 (TypeScript conversion, merged after 17.1.0) re-created it on `develop`, and #12727 deleted that copy before 18.0.0. It never shipped in 17.x or 18.0, so the 18.0.0 changelog line describes a develop-only cleanup. Left the released changelog text alone; noted in the PR description as out of scope.

.ai/BREAKING-CHANGES.md
```diff
@@ -10,14 +10,25 @@ The lean rule lives in the root `AGENTS.md` "Breaking changes policy" section (a
...
-| Changing API signatures or behavior | Breaks customer integrations | Keep the deprecated API working until the next stable release. Deprecated APIs produce a console warning (fired only once). |
-| Removing hooks or configuration options | May go undetected by customers | Add the hook or option to the list of removed hooks so an error shows when someone uses it in configuration. |
+| Changing API signatures or behavior | Breaks customer integrations | Keep the deprecated API working until the next major release. Deprecated APIs produce a console warning (fired only once). |
+| Removing hooks or configuration options | May go undetected by customers | Add the hook to `REMOVED_HOOKS` in `handsontable/src/core/hooks/constants.ts`, or the option to `REMOVED_OPTIONS` in `handsontable/src/core.ts`, so a one-time warning shows when someone uses it in configuration. The warning names the removing version and carries no `Deprecated:` prefix (`removedWarnOnce`, not `deprecatedWarnOnce`). |
```
demtario: **Medium.** Both claims in this cell are true of the options half only, and the hooks half is documented the opposite way elsewhere in this same PR.

[`core/hooks/index.ts:94-95`](https://github.com/handsontable/handsontable/blob/31b950c094b6524b83d1410a4fabf810f4e017b0/handsontable/src/core/hooks/index.ts#L94):

```ts
if (REMOVED_HOOKS.has(key)) {
  warn(substitute(REMOVED_MESSAGE, { hookName: key, removedInVersion: REMOVED_HOOKS.get(key) }));
}
```

That is the plain `warn()` helper, fired inside `add()`. It is not once-per-key, it repeats on every registration, and it never touches `removedWarnOnce`. Grepping `removedWarnOnce` across `handsontable/src` returns two files: `helpers/console.ts` (definition) and `core.ts` (its only caller).

[`handsontable/.ai/HOOKS.md:246`](https://github.com/handsontable/handsontable/blob/31b950c094b6524b83d1410a4fabf810f4e017b0/handsontable/.ai/HOOKS.md#L246), edited in this PR, states it plainly: "removed and deprecated hooks both emit a non-deduplicated `console.warn` on every `add()`."

So the PR now carries two edited files saying opposite things about one code path, which is the failure mode it exists to remove. An agent following this row for a hook removal will look for a helper the hook path does not use.

Fix: scope the sentence, e.g. "The option warning is one-time and prefix-free (`removedWarnOnce`, not `deprecatedWarnOnce`); the hook warning goes through plain `warn()` on every registration."

docs/content/guides/upgrade-and-migration/deprecation-policy/deprecation-policy.md
```diff
@@ -46,27 +46,62 @@ For more details about our versioning policy please visit [Versioning policy](@/
...
+| `loadManualRowHeights()` (`ManualRowResize`) | 18.0 | 19.0 | No-op. Pass an array to `manualRowResize`. | [Migrate from 17.1 to 18.0 -> Resize-state methods](@/guides/upgrade-and-migration/migrating-from-17.1-to-18.0/migrating-from-17.1-to-18.0.md#stop-calling-deprecated-resize-state-methods) |
+| `Handsontable.helper.sanitize()` | 18.0 | 19.0 | Pass-through. Use the `sanitizer` option. | [`sanitizer` option](@/api/options.md#sanitizer) |
+
+### Options
+
+| Deprecated | Deprecated in | Removal | Replacement | Migration guide |
+| ---------- | ------------- | ------- | ----------- | --------------- |
+| `columnHeaders` (export options of the `exportFile` plugin) | 17.1 | 19.0 | `colHeaders` – rename the key; the value keeps its meaning. | [Export to CSV -> Available options](@/guides/accessories-and-menus/export-to-csv/export-to-csv.md#available-options-in-the-export-configuration) |
```
demtario: **Low.** `` `colHeaders` – rename the key `` uses U+2013 EN DASH. Confirmed by codepoint, and it is the only one on the page; every other clause separator here uses `--`, including the `-- (never documented)` cell above.

[`docs/AGENTS.md:61`](https://github.com/handsontable/handsontable/blob/31b950c094b6524b83d1410a4fabf810f4e017b0/docs/AGENTS.md#L61) makes this a site rule that overrides the monorepo standard: the docs site "uses hyphens or double hyphens, never en dashes". Restated at `docs/AGENTS.md:94`.

Fix: `` `colHeaders` -- rename the key; the value keeps its meaning. ``

handsontable/src/core.ts
```diff
@@ -142,12 +142,83 @@ function normalizeIndexesGroup(indexes: number[][]): number[][] {
...
+ * `persistentState` says 17.0.0, not 18.0.0: #12015 removed the plugin, its option, and its hooks
+ * in 17.0.0, and no published 17.x or 18.x package carries them. The 18.0.0 changelog entry for
+ * #12727 describes deleting a copy that the TypeScript conversion re-created on `develop` after
+ * 17.1.0 and that never shipped.
+ *
+ * `datePickerConfig` is deliberately absent: `DateEditor#prepare` already warns about it, and it
+ * also sees the per-cell forms that `warnAboutRemovedOptions` cannot.
  *
- * @type {Set<string>}
+ * @type {Array<RemovedOption>}
```
demtario: **Low.** `@type {Array<RemovedOption>}` does not match the declaration two lines down, which is `ReadonlyArray<RemovedOption>`. The annotation this replaced (`Array<Array<string>>`) at least matched its shape.

Also worth knowing: `RemovedOption` is a TypeScript `interface`, not a `@typedef`, so nothing resolves the name in JSDoc space.

Fix: `@type {ReadonlyArray<RemovedOption>}`.

handsontable/.ai/CONVENTIONS.md
```diff
@@ -141,15 +141,18 @@ export function throwWithCause(message) {
 - `log(...args)` -- General logging
 - `warn(...args)` -- Warning messages
 - `deprecatedWarn(message)` -- Deprecated feature warnings (prefixed with "Deprecated: ")
```
demtario: **Low.** Two lines in this block are now stale after the `warnOncePerKey` refactor.

`deprecatedWarnOnce` was `deprecatedWarn`'s last production caller. Both once-helpers now route through the private `warnOncePerKey`, so a grep across `handsontable/src` and `wrappers` finds [`deprecatedWarn`](https://github.com/handsontable/handsontable/blob/31b950c094b6524b83d1410a4fabf810f4e017b0/handsontable/src/helpers/console.ts#L77) referenced only by its own definition and by `helpers/__tests__/console.unit.ts`. It is in no barrel and not surfaced on `base.ts`, so it is not public API either. The line below still tells agents to "use `deprecatedWarn` only when repeated output is intended", pointing them at a dead export.

The `_resetDeprecationWarnings()` entry under it also still says "any spec asserting on a deprecation warning". It now clears removal warnings too, which is exactly what the new shared-record test proves.

Fix: drop the `deprecatedWarn` clause (or delete the export and its spec block, nothing outside the test depends on it), and widen the reset description to cover both warning kinds.

handsontable/src/plugins/exportFile/utils.ts
```diff
@@ -1,6 +1,43 @@
 import { html } from '../../helpers/templateLiteralTag';
 import { LOADING_CLASS_NAME } from '../../helpers/constants';
 import { escapeHtml } from '../../helpers/string';
+import { deprecatedWarnOnce } from '../../helpers/console';
+
+export function normalizeExportOptions<T extends Record<string, unknown>>(options: T): T;
+export function normalizeExportOptions<T extends Record<string, unknown>>(options: T | undefined): T | undefined;
```
demtario, line 7: **Low.** This second overload currently has no TypeScript caller. All three call sites pass a definite object and resolve to overload 1:

- [`dataProvider.ts:64`](https://github.com/handsontable/handsontable/blob/31b950c094b6524b83d1410a4fabf810f4e017b0/handsontable/src/plugins/exportFile/dataProvider.ts#L64), inside `setOptions(options: Record<string, unknown>)`
- [`types/_base.ts:102`](https://github.com/handsontable/handsontable/blob/31b950c094b6524b83d1410a4fabf810f4e017b0/handsontable/src/plugins/exportFile/types/_base.ts#L102), fed to `extend()`
- [`types/xlsx.ts:290`](https://github.com/handsontable/handsontable/blob/31b950c094b6524b83d1410a4fabf810f4e017b0/handsontable/src/plugins/exportFile/types/xlsx.ts#L290), inside a spread

The only `undefined` call is `exportFile.unit.js`, a `.js` file that is not type-checked. Overload 1 earns its place (it is what makes the `dataProvider.ts:64` assignment legal); overload 2 documents an intent nothing exercises.

Not wrong, and cheap to keep. If the `undefined` contract is meant to be load-bearing, a line in a `.types.ts` fixture would pin it.

demtario (commented):
Follow-up review of `31b950c094`. Thanks for the thorough turnaround, all nine items from the previous round check out in the code. Five inline comments above on what the fixes introduced. One finding has no line in this diff to hang on, so it is here.

**Medium. The docs restructure now contradicts the shipped 18.0 changelog.**

Moving `PersistentState` under `## Removed in version 17.0` is correct, and I verified it independently: 0 non-test files matching `persistentState` in `handsontable/src` at tag `17.0.0` against 11 at `16.2.0`, plus `core/hooks/constants.ts` already carrying `['persistentStateSave', '17.0.0']`.

The problem is what was left behind. [`changelog-18.md:58`](https://github.com/handsontable/handsontable/blob/31b950c094b6524b83d1410a4fabf810f4e017b0/docs/content/guides/upgrade-and-migration/changelog-18/changelog-18.md#L58) still reads:

> **Breaking change**: Removed the deprecated `PersistentState` plugin, its `persistentState` option, and the `persistentStateSave`, `persistentStateLoad`, and `persistentStateReset` hooks. … [#12727]

The same text is duplicated at `changelog/changelog.md:61`. Before this commit the two pages agreed on 18.0; now a reader who follows the new console warning to `/changelog-17` and then opens the 18.0 changelog gets two removal versions for one API.

The reconciliation you worked out, that #12727 deleted a `develop`-only copy the TypeScript conversion recreated after 17.1.0, is convincing, but it lives only as a source comment at `core.ts:167-170` where no user will see it. I am not suggesting rewriting a shipped entry's history. One qualifying clause on `changelog-18.md:58`, along the lines of "removed from the source tree; the published 17.0 release already shipped without it, see #12015", reconciles the pages without falsifying what #12727 did.

**Also verified clean in this commit**, recorded so it is not re-derived: the `warnOncePerKey` refactor preserves the original ordering exactly, including recording the key only after the `console` check, and `Deprecated: ${message}` is byte-identical to what `deprecatedWarn` produced. Deprecated and removed key namespaces are disjoint, and the new shared-key test documents the design rather than exposing a bug. The overload placement matches the `plugins/filters/utils.ts:51-59` precedent exactly, so `jsdoc/require-jsdoc` does not fire on bodyless overloads. The `cells` scan cannot throw: `isObject` is null-safe and `Array.isArray` guards the non-array forms. Both `migrationUrl` values resolve against the guides' `permalink` frontmatter, and `/changelog-17` is the right target given that the 16.2-to-17.0 guide has no PersistentState section at all. Every new docs link resolves, including the `#available-options-in-the-export-configuration` anchor. The six new tests assert real behavior, the two renamed ones kept their original assertion and added to it, and nothing was deleted.

**One pre-existing item, untouched here:** `changelog/changelog.md:61` still carries the `[#0](https://github.com/handsontable/handsontable/pull/0)` placeholder on the PersistentState line, the same one this PR fixed in the root `CHANGELOG.md`. Worth picking up while you are in that file for the item above.

.ai/BREAKING-CHANGES.md
adrianspdev: Fixed in 1028ddf850. The row now separates the two paths: option warning is one-time and prefix-free (`removedWarnOnce`); hook warning goes through plain `warn()` on every `add()` and links to the release notes, with a pointer to `handsontable/.ai/HOOKS.md`.

docs/content/guides/upgrade-and-migration/deprecation-policy/deprecation-policy.md
adrianspdev: Fixed in 1028ddf850 – `--`. Page now has zero U+2013.

handsontable/src/core.ts
adrianspdev: Fixed in 1028ddf850 – `@type {ReadonlyArray<RemovedOption>}`.

handsontable/.ai/CONVENTIONS.md
adrianspdev: Fixed in 1028ddf850. Went with the delete: `deprecatedWarn` and its spec block are gone (only caller was the test; not on `Handsontable.helper`). The `deprecatedWarnOnce` entry no longer points at a repeat variant, and `_resetDeprecationWarnings()` now says it clears both deprecation and removal warnings.

handsontable/src/plugins/exportFile/utils.ts
adrianspdev: Pinned in 1028ddf850. `exportFile.types.ts` now imports `normalizeExportOptions` through `handsontable/plugins/exportFile/utils` (the fixture tsconfig maps `handsontable/*` to `tmp/*`) and asserts both overloads: a definite object stays definite, `undefined` comes back `undefined`. Overload 2 kept.

adrianspdev:
Second round addressed in 1028ddf850; inline threads have per-item replies.

**Changelog reconciliation** – the PersistentState line now ends with "Removed from the source tree; the published 17.0 release already shipped without it (see #12015)." in all three copies: root `CHANGELOG.md`, `changelog-18.md`, and `changelog/changelog.md`. The shipped wording of what #12727 did is unchanged; the clause only adds the release fact.

**`changelog/changelog.md` `#0` placeholder** – replaced with #12727, matching the other two copies.

Verified: `test:types` (fixture builds against `tmp/`), `console.unit` 29/29, eslint clean on changed files, `grep -c '–'` on the deprecation-policy page returns 0.

demtario (approved):

budnix (approved):

Issue #10268, opened 2023-03-08
Bind to column major array of arrays


jzju:
Is it possible to add functionality for binding data to column major arrays?

adrianszymanski89:
Hi @jzju 

Can you please tell us more about your requirements here?

jzju:
If the data is 
```
[[a, b], [c, d, e]]
```
it would shown as

```
a c
b d
  e
```

adrianszymanski89:
Thank you. We have an internal helper that lets you pivot the array, so it can be achieved this way:

https://jsfiddle.net/aszymanski/07Lr9f5k/

You just need to add `null` in the first array so the whole data will be visible.

jzju:
There where a function translateRowsToColumns like that in the old versions that didn't need the extra null: https://unpkg.com/browse/handsontable@8.3.2/helpers/data.mjs

But the original data don't get bound to changes so you have to transpose back and forth. Thanks for the help I just wanted to make a suggestion to improve your product incase you ran out of ideas :)

adrianszymanski89:
Sure, thank you for that. As you can see, we haven't any public API that would let you do this out of the box, but I'll mark it as a feature suggestion.

budnix:
Hi @jzju, sorry this sat for so long — and thanks for the suggestion, it turned up a real bug on our side.

Two things.

**1. You can bind to a column-major array today, with two-way binding, and no extra `null`.**

The bit that was missing from the earlier answer is that a column's `data` option accepts a *function*. Point each column at its own array and the grid reads and writes that array directly, so there is no transposed copy to keep in sync:

```js
const source = [
  ['a', 'b'],      // column A
  ['c', 'd', 'e'], // column B
];

// Handsontable needs one entry per row, so give it a small object per row
// that records its own index.
const rows = Array.from(
  { length: source.reduce((max, col) => Math.max(max, col.length), 0) },
  (_, index) => ({ index })
);

new Handsontable(container, {
  data: rows,
  columns: source.map((col, c) => ({
    // A regular function, not an arrow function: `arguments.length` is what
    // separates a read from a write.
    data: function (row, value) {
      if (arguments.length === 1) {
        return row ? source[c][row.index] ?? null : null;
      }
      if (row) {
        source[c][row.index] = value;
      }
    },
  })),
});
```

That renders exactly the shape you drew:

```
a c
b d
  e
```

[...]

PR #8867, opened 2021-10-14
Changed a way how the `populatedFromArray` works (with `shift_down` and `shift_right` method)

wszymanski:
### Context

The `populatedFromArray` array with `shift_down` and `shift_right` options has been performing `n` `spliceRow`/`spliceCol` actions (where `n` is number of rows/columns). Thus, there were performed `n` `beforeChange` and `afterChange` callbacks. It was slow, bugged (see source issue handsontable/handsontable#888 and handsontable/handsontable#6929) and could be incomprehensible for developer. The method doesn't work in the same way for `overwrite` option even though this option may also be performed on multiple rows/columns. Thus, `populateFromArray` does not work the same for different options.

This PR change a way how data is populated. No small operations will be performed, but only one larger operation. It's almost `6` times faster!

Screenshot 2021-10-14 at 17 42 15

*Tested on: Google Chrome 94, macOS Catalina 10.15.7, MacBook Pro 15" (Mid 2015, 2.2 GHz i7) while copying and pasting dataset of `10x10` elements (with rendering - measured time from triggering `beforePaste` hook to triggering `afterPaste`).*

~What's more, there have been removed `spliceRow` and `spliceCol` methods. They haven't been used anywhere within Handsontable since 2013. I suppose that they aren't used often by users and they can be easily replaced by simple code (as you can [see in docs](https://github.com/handsontable/handsontable/pull/8867/files#diff-a142d5f2c0e58c44b5f4fcef5788669f8fabbaeb0cbb8034353422276407b945)).~

Some utils used only within above methods, such as ~`extendArray`, `to2dArray` and~ `translateRowsToColumns`, have been also removed from the code.

### How has this been tested?

Used `populatedFromArray` with different configurations and `CopyPaste` plugin with different `pasteMode` options.

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature or improvement (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Additional language file or change to the existing one (translations)

### Related issue(s):
1. handsontable/handsontable#888
2. https://github.com/handsontable/handsontable/issues/6929

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`

### Checklist:

[...]

PR #13254, opened 2026-08-26
DEV-2638: Fix pivot() dropping cells for a ragged input

budnix:
### Context

The PR fixes silent data loss in `populateFromArray()` when the input rows have unequal lengths, and documents column-oriented data binding.

`populateFromArray()` routes its `shift_down` method through the `pivot()` helper (`handsontable/src/helpers/array.ts`), which took the number of output columns from the **first input row only**:

```js
const colCount = arr[0].length;
```

Every cell past that width was never copied. On a 3x3 grid:

```js
hot.populateFromArray(0, 0, [['X'], ['Y1', 'Y2', 'Y3']], null, null, null, 'shift_down');
```

`'Y2'` and `'Y3'` are dropped, with no error and no hook to intercept. The same call with `'shift_right'` handles the ragged input correctly, because that branch does not use `pivot()` — one API, two orientations, and only one of them loses data.

This is the same "first row decides the width" family as the ragged-paste fix in #13237. **Paste is not affected**: `copyPaste` already pads the pasted array rectangular before it reaches `populateFromArray()`.

`pivot()` now takes the width from the widest row, and fills the gaps a shorter row leaves with `null` — the same choice `copyPaste` already makes, so the two paths agree.

A rectangular input keeps the same width. One output change does reach it: an explicit `undefined` now comes back as `null`, in a full-width row too, so that every returned row is dense. `core.ts` skips a change where the old and new values are both null-or-undefined, so no extra write reaches `datamap.set`, but `beforeChange` / `afterChange` tuples now carry `null` where they carried `undefined`.

The old `translateRowsToColumns` helper (removed in #8867) is deliberately **not** restored. It was never correct either: on a ragged input it *misplaces* the extra cell into column 0 rather than dropping it.

**Docs.** Issue #10268 asked to bind a column-major array of arrays with two-way binding. That needs **no new option** — `columns[].data` accessor functions already cover it. A new "Column-oriented data source" section in the *Binding to data* guide documents the recipe, along with the traps found while verifying it in a browser: a row the grid creates writes to an off-array slot that every created row shares, until the structural hooks give it a real one, `getSourceData()` returns copies of the row objects, and `afterCreateRow`/`afterRemoveRow` report visual indexes while the column arrays are physical.

### How has this been tested?

10 new unit tests for `pivot()` and 4 new E2E cases in the existing `shift_down` block of `populateFromArray.spec.js`.

Red/green was confirmed in both directions: with the width fix reverted, the new unit tests and E2E cases fail with exactly the data-loss shape; with it restored, they pass.

```bash
# Unit tests for the helper
npm run test:unit --prefix handsontable -- --testPathPattern=helpers/__tests__/array

# E2E for the fixed method and the helper's other consumers
npm run test:e2e --prefix handsontable -- --testPathPattern='autocompleteEditor|populateFromArray|autofill|copyPaste'

# Full gates
npm run test:unit --prefix handsontable      # 3809 passed
npm run eslint --prefix handsontable         # 0 errors
```

The documented guide snippet was also run verbatim in Chrome to confirm the recipe works as written.

### Types of changes

### Related issue(s):

1. DEV-2638
2. #10268

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2638

---

> [!NOTE]
> **Medium Risk**
> Changes core `populateFromArray` / `pivot` behavior for ragged 2D arrays; fixes incorrect data loss but alters outcomes for that edge case, with broad downstream use of `pivot` in core and editors.
> 
> **Overview**
> Fixes **silent data loss** when `populateFromArray()` runs with the `shift_down` method on input where rows have different lengths. That path transposes the input via `pivot()` in `handsontable/src/helpers/array.ts`, which previously sized the result from the **first row only**—extra cells in wider rows were never copied.
> 
> `pivot()` now derives width from the **widest row**, pads shorter rows with `null` (aligned with copy/paste padding), and still returns empty output for `[]` / `[[]]`. Rectangular inputs behave as before.
> 
> Adds unit tests for `pivot()` and E2E cases for `shift_down`, a changelog entry for #10268, and a new **Column-oriented data source** section in the Binding to data guide (accessor-based binding to column arrays, plus row create/remove hooks).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 893c59ba5c4ed872c975082cc565ec91e68e8386. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changelogs/10268.json
+++ b/.changelogs/10268.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed `populateFromArray()` with the `shift_down` method silently dropping cells when the input rows had unequal lengths.",
+  "type": "fixed",
+  "issueOrPR": 10268,
+  "breaking": false,
+  "framework": "none"
+}
--- a/.changelogs/10268.json
+++ b/.changelogs/10268.json
@@ -1,8 +1,0 @@
-{
-  "issuesOrigin": "public",
-  "title": "Fixed `populateFromArray()` with the `shift_down` method silently dropping cells when the input rows had unequal lengths.",
-  "type": "fixed",
-  "issueOrPR": 10268,
-  "breaking": false,
-  "framework": "none"
-}
--- a/docs/content/guides/getting-started/binding-to-data/binding-to-data.md
+++ b/docs/content/guides/getting-started/binding-to-data/binding-to-data.md
@@ -404,39 +404,45 @@ const accessor = columnIndex => function(row, value) {
   }
 };
 
-const hot = new Handsontable(container, {
+const settings = {
   data: rows,
   columns: source.map((column, columnIndex) => ({ data: accessor(columnIndex) })),
   dataSchema: () => ({ index: -1 }),
-  licenseKey: 'non-commercial-and-evaluation',
-});
+};
 ```
+
+Pass `settings` to Handsontable the way your framework does -- as the second argument to the constructor, or as the component's settings.
 
 The grid now renders one row per position across the columns. Columns of unequal length leave empty cells at the bottom of the shorter ones. An edit writes straight into `source`, so no copy can fall behind.
 
 That is the whole setup for reading and writing existing cells. Adding and removing rows needs two more hooks.
 
-A row the grid creates has no slot in your column arrays yet. Until you make one, that row renders empty and discards whatever the end user types into it, without an error. So if anything can create a row -- the context menu, [`minSpareRows`](@/api/options.md#minsparerows), or [`alter()`](@/api/core.md#alter) -- keep the column arrays in step through [`afterCreateRow`](@/api/hooks.md#aftercreaterow) and [`afterRemoveRow`](@/api/hooks.md#afterremoverow).
+A row the grid creates has no slot in your column arrays yet. Its row object still holds the [`dataSchema`](@/api/options.md#dataschema) marker `-1`, so the accessor writes a value typed into that row to `source[columnIndex][-1]` -- a property on the array object, not one of its elements. The grid displays the value, because the accessor reads the same slot back, but the value never joins your data, and every row the grid creates shares that one slot. So if anything can create a row -- the context menu, [`minSpareRows`](@/api/options.md#minsparerows), or [`alter()`](@/api/core.md#alter) -- keep the column arrays in step through [`afterCreateRow`](@/api/hooks.md#aftercreaterow) and [`afterRemoveRow`](@/api/hooks.md#afterremoverow).
 
 Both hooks report a **visual** row index, while the column arrays are physical, so translate before you splice. For an insert, the new row object still carries the [`dataSchema`](@/api/options.md#dataschema) marker, and its position in `rows` is the physical index. For a removal, use the hook's `physicalRows` argument and splice the highest index first:
 
 ```javascript
-afterCreateRow(index, amount) {
-  const at = rows.findIndex(row => row.index === -1);
+const settings = {
+  // ... the options above
+  afterCreateRow(index, amount) {
+    const at = rows.findIndex(row => row.index === -1);
 
-  source.forEach(column => {
-    column.splice(at === -1 ? index : at, 0, ...new Array(amount).fill(null));
-  });
-  restamp(rows);
-  hot.render();
-},
-afterRemoveRow(index, amount, physicalRows) {
-  [...physicalRows].sort((a, b) => b - a).forEach(row => {
-    source.forEach(column => column.splice(row, 1));
-  });
-  restamp(rows);
-  hot.render();
-},
+    source.forEach(column => {
+      column.splice(at === -1 ? index : at, 0, ...new Array(amount).fill(null));
+    });
+    restamp(rows);
+    // Call `this.render()`, not a variable holding the instance: with `minSpareRows` this hook
+    // runs while Handsontable is still starting up, before that variable holds anything.
+    this.render();
+  },
+  afterRemoveRow(index, amount, physicalRows) {
+    [...physicalRows].sort((a, b) => b - a).forEach(row => {
+      source.forEach(column => column.splice(row, 1));
+    });
+    restamp(rows);
+    this.render();
+  },
+};
 ```
 
 Three more things to know about this setup:
--- a/docs/content/guides/getting-started/binding-to-data/binding-to-data.md
+++ b/docs/content/guides/getting-started/binding-to-data/binding-to-data.md
@@ -366,6 +366,85 @@ The example below shows how to use such objects:
 
 :::
 
+### Column-oriented data source
+
+Handsontable reads its data row by row. If your data arrives as one array per column, you don't have to keep a transposed copy in sync -- point each column's [`data`](@/api/options.md#data) option at the array it belongs to.
+
+Handsontable can't tell the two orientations apart on its own. `[['a', 'b'], ['c', 'd']]` is the same value whether you mean it as two rows or as two columns, so you state which one you mean through [`columns`](@/api/options.md#columns).
+
+The [`data`](@/api/options.md#data) option still needs one entry per row. Use a small object per row that records its own index, and mint new ones through [`dataSchema`](@/api/options.md#dataschema):
+
+```javascript
+const source = [
+  ['a', 'b'],      // column A
+  ['c', 'd', 'e'], // column B
+];
+
+const rowCount = source.reduce((max, column) => Math.max(max, column.length), 0);
+const restamp = rows => {
+  rows.forEach((row, index) => {
+    row.index = index;
+  });
+
+  return rows;
+};
+
+// Handsontable splices this array in place, so it stays a live view of the row order.
+const rows = restamp(Array.from({ length: rowCount }, () => ({ index: 0 })));
+
+// A regular function, not an arrow function: `arguments.length` tells a read from a write.
+const accessor = columnIndex => function(row, value) {
+  if (arguments.length === 1) {
+    // The row can be missing while the grid measures sizes or replays an undo step.
+    return row ? source[columnIndex][row.index] ?? null : null;
+  }
+
+  if (row) {
+    source[columnIndex][row.index] = value;
+  }
+};
+
+const hot = new Handsontable(container, {
+  data: rows,
+  columns: source.map((column, columnIndex) => ({ data: accessor(columnIndex) })),
+  dataSchema: () => ({ index: -1 }),
+  licenseKey: 'non-commercial-and-evaluation',
+});
+```
+
+The grid now renders one row per position across the columns. Columns of unequal length leave empty cells at the bottom of the shorter ones. An edit writes straight into `source`, so no copy can fall behind.
+
+That is the whole setup for reading and writing existing cells. Adding and removing rows needs two more hooks.
+
+A row the grid creates has no slot in your column arrays yet. Until you make one, that row renders empty and discards whatever the end user types into it, without an error. So if anything can create a row -- the context menu, [`minSpareRows`](@/api/options.md#minsparerows), or [`alter()`](@/api/core.md#alter) -- keep the column arrays in step through [`afterCreateRow`](@/api/hooks.md#aftercreaterow) and [`afterRemoveRow`](@/api/hooks.md#afterremoverow).
+
+Both hooks report a **visual** row index, while the column arrays are physical, so translate before you splice. For an insert, the new row object still carries the [`dataSchema`](@/api/options.md#dataschema) marker, and its position in `rows` is the physical index. For a removal, use the hook's `physicalRows` argument and splice the highest index first:
+
+```javascript
+afterCreateRow(index, amount) {
+  const at = rows.findIndex(row => row.index === -1);
+
+  source.forEach(column => {
+    column.splice(at === -1 ? index : at, 0, ...new Array(amount).fill(null));
+  });
+  restamp(rows);
+  hot.render();
+},
+afterRemoveRow(index, amount, physicalRows) {
+  [...physicalRows].sort((a, b) => b - a).forEach(row => {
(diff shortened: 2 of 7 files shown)


qunabu:
Task linked: [DEV-2638 populateFromArray with `shift_down` drops cells for a ragged input (#10268)](https://app.clickup.com/t/9015210959/DEV-2638)

handsontable/src/helpers/array.ts
```diff
@@ -26,29 +26,49 @@ export function extendArray(arr: unknown[], extension: unknown[]): void {
 }
 
 /**
+ * Transposes a two-dimensional array, turning its rows into columns.
+ *
+ * Rows of unequal length are supported: the widest row decides how many columns come out, and
```
demtario: The new width semantics are documented here, but this JSDoc never reaches the API reference: the generator's source list (`docs/scripts/jsdoc-convert/parser/parser.mjs:12-29`) contains `core.js` and the plugin/editor globs, and no `helpers/*` entry. `populateFromArray()`'s own JSDoc (`handsontable/src/core.ts:2470-2487`) still says nothing about ragged input, so the rendered page for the method a user actually calls does not describe the fixed behavior. One line on `@param [method]` closes it (monorepo checklist item 2).

Minor, same block: "Rows of unequal length are supported" is passive voice, which `.ai/DOC-STANDARDS.md` rule 2 applies to JSDoc as well.

handsontable/src/helpers/array.ts
```diff
@@ -26,29 +26,49 @@ export function extendArray(arr: unknown[], extension: unknown[]): void {
...
+  for (let j = 0; j < colCount; j++) {
+    const pivotedRow: unknown[] = new Array(rowCount);
+
+    for (let i = 0; i < rowCount; i++) {
+      const row = arr[i];
+      const value = row ? row[j] : undefined;
+
+      // A row shorter than the widest one has no value here. Keep the output dense with the
+      // empty-cell value rather than leaving a hole for the caller to trip over.
+      pivotedRow[i] = value === undefined ? null : value;
```
demtario, line 69: This ternary fires for any explicitly-`undefined` cell, ragged or not, so the PR description's "Rectangular input is unchanged" is not accurate as written:

```js
pivot([[1, undefined], [3, 4]]);
// before: [[1, 3], [undefined, 4]]
// after:  [[1, 3], [null, 4]]
```

That shape is reachable from real grid data, not only from caller-supplied input: `dataMap.get()` (`src/dataMap/dataMap.ts:877-878`) returns `dataRow[prop]` verbatim when the row owns the property, and `getValueGetterValue()` (`src/utils/valueAccessors.ts:27-35`) returns the value with no null coalescing.

The impact is bounded, so I am not asking for a code change. `src/core.ts:2014-2018` already skips a change where old and new are both null-or-undefined, so no extra write reaches `datamap.set`. What does change: `beforeChange`/`afterChange` tuples now carry `null` where they carried `undefined` (a user hook testing `=== undefined` breaks), and on an object data source with a real old value the property is written as an explicit `null` instead of being left `undefined`. Worth correcting the sentence in the PR body, and worth a line in the JSDoc above.

handsontable/src/helpers/array.ts
```diff
@@ -26,29 +26,49 @@ export function extendArray(arr: unknown[], extension: unknown[]): void {
...
 /**
+ * Transposes a two-dimensional array, turning its rows into columns.
+ *
+ * Rows of unequal length are supported: the widest row decides how many columns come out, and
+ * a shorter row contributes the empty-cell value (`null`) for the positions it does not reach.
+ *
  * @param {Array} arr An array to pivot.
  * @returns {Array}
  */
 export function pivot(arr: unknown[][]): unknown[] {
```
demtario: Pre-existing, but this PR is already rewriting the signature's documentation: the declared return type is `unknown[]` while the function returns `unknown[][]`. Every call site carries a cast to compensate (`src/core.ts:1307`, `:1315`, `:1341`, `src/editors/autocompleteEditor/autocompleteEditor.ts:400`), and because `helper` is typed as `typeof import('./helpers/array')` (`src/base.ts:106`), TypeScript consumers of `Handsontable.helper.pivot` get `unknown` when they index the result. Tightening it to `unknown[][]` here removes all four casts.

handsontable/src/helpers/array.ts
```diff
@@ -26,29 +26,49 @@ export function extendArray(arr: unknown[], extension: unknown[]): void {
...
   const pivotedArr: unknown[][] = [];
 
-  if (!arr || arr.length === 0 || !arr[0] || arr[0].length === 0) {
+  if (!arr || arr.length === 0) {
```
demtario, line 43: Pre-existing crash that this change narrows without closing. `src/core.ts:1338` indexes `populatedDataByColumns[c % numberOfDataColumns]`; when `numberOfDataColumns` is 0, `c % 0` is `NaN`, the lookup is `undefined`, and `.concat` throws.

```js
hot.populateFromArray(0, 0, [[], []], 1, 1, 'test', 'shift_down');
// TypeError: Cannot read properties of undefined (reading 'concat')
```

The `rlen === 0` early return misses it because `input.length` is 2. Before this PR the same crash also fired for `[[], ['a', 'b']]`, since the old `pivot()` returned `[]` for any empty first row; the new width scan removes that shape incidentally. A `numberOfDataColumns === 0` early return in `core.ts` closes what is left, if you want to take it in the same PR.

handsontable/src/helpers/__tests__/array.unit.ts
```diff
@@ -534,4 +535,42 @@ describe('Array helper', () => {
...
+
+    it('should return an empty array for an empty input', () => {
+      expect(pivot([])).toStrictEqual([]);
+      expect(pivot([[]])).toStrictEqual([]);
+    });
+
+    it('should preserve a `null` that was in the input', () => {
+      expect(pivot([[null, 'b'], ['c', null]])).toStrictEqual([[null, 'c'], ['b', null]]);
+    });
+  });
```
demtario, line 590: The `undefined` to `null` mapping is the one change that alters the output for input that was already handled correctly (a full-width row holding an explicit `undefined`), and it is the line the "not a breaking change" judgement rests on. Nothing in this block pins it. Suggest adding:

```js
it('should replace an explicit `undefined` with the empty-cell value', () => {
  expect(pivot([[1, undefined], [3, 4]])).toStrictEqual([[1, 3], [null, 4]]);
});
```

The guards added at `array.ts:52` and `:62` for a non-array row are also untested; either cover them or drop them, given the `unknown[][]` signature.

.changelogs/10268.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed `populateFromArray()` with the `shift_down` method silently dropping cells when the input rows had unequal lengths.",
+  "type": "fixed",
+  "issueOrPR": 10268,
```
demtario: The cited number renders as `issues/10268`, which is "Bind to column major array of arrays" - the docs half of this PR, not the bug the entry's title describes. For a `fixed` entry that reads as a wrong link. Work tracked in ClickUp is `"issuesOrigin": "private"` with the PR number as both the value and the filename (`.changelogs/README.md:53-56`), the way the sibling `13237.json` does it.

Separately, the title names only `populateFromArray()`, but `Handsontable.helper.pivot` changed behavior too (it is on the public helper surface via `src/base.ts:106`, even though it is not in the API reference). Worth naming it in the title, since one entry has to cover both.

docs/content/guides/getting-started/binding-to-data/binding-to-data.md
```diff
@@ -366,6 +366,85 @@ The example below shows how to use such objects:
 
 :::
 
+### Column-oriented data source
```
demtario, line 369: This section is not framework-gated: the snippets below are bare ```` ```javascript ```` with `new Handsontable(container, ...)`, and there is no `::: only-for javascript` wrapper. Every other `###` section on this page is gated, so React, Angular, and Vue readers land on JavaScript-only setup code. The ungated snippet further down at :454 is framework-agnostic prose, which this one is not.

docs/content/guides/getting-started/binding-to-data/binding-to-data.md
```diff
@@ -366,6 +366,85 @@ The example below shows how to use such objects:
...
+  dataSchema: () => ({ index: -1 }),
+  licenseKey: 'non-commercial-and-evaluation',
+});
+```
+
+The grid now renders one row per position across the columns. Columns of unequal length leave empty cells at the bottom of the shorter ones. An edit writes straight into `source`, so no copy can fall behind.
+
+That is the whole setup for reading and writing existing cells. Adding and removing rows needs two more hooks.
+
+A row the grid creates has no slot in your column arrays yet. Until you make one, that row renders empty and discards whatever the end user types into it, without an error. So if anything can create a row -- the context menu, [`minSpareRows`](@/api/options.md#minsparerows), or [`alter()`](@/api/core.md#alter) -- keep the column arrays in step through [`afterCreateRow`](@/api/hooks.md#aftercreaterow) and [`afterRemoveRow`](@/api/hooks.md#afterremoverow).
```
demtario: "renders empty and discards whatever the end user types into it" does not look like what the snippet above produces. With `dataSchema: () => ({ index: -1 })` and no hooks yet, the accessor at :396-404 writes `source[columnIndex][-1] = value` and then reads that same slot back, so the edit does persist and display - and every grid-created row shares that one off-array slot. The advice that follows (implement the two hooks) is right either way; it is the described failure mode that needs re-checking in the browser.

docs/content/guides/getting-started/binding-to-data/binding-to-data.md
```diff
@@ -366,6 +366,85 @@ The example below shows how to use such objects:
...
+
+```javascript
+afterCreateRow(index, amount) {
+  const at = rows.findIndex(row => row.index === -1);
+
+  source.forEach(column => {
+    column.splice(at === -1 ? index : at, 0, ...new Array(amount).fill(null));
+  });
+  restamp(rows);
+  hot.render();
```
demtario: `hot.render()` inside `afterCreateRow` throws for the exact setup the paragraph above recommends. With `minSpareRows`, the hook fires during construction: `init()` calls `updateSettings(mergedUserSettings, true)` (`src/core.ts:1636`) which reaches `grid.adjustRowsAndCols()` (`:3705`) and `datamap.createRow(..., { source: 'auto' })` (`:1211`), running `afterCreateRow` (`src/dataMap/dataMap.ts:432`) before `const hot = new Handsontable(...)` has bound. A reader who copies this and adds `minSpareRows` gets:

```
ReferenceError: Cannot access 'hot' before initialization
```

Use the hook's `this` instead, or `hot?.render()`. The same call at :438 is safe in practice (a removal cannot happen during construction), but it is worth making both consistent.

The rest of the section checks out against the source: `afterRemoveRow(index, amount, physicalRows, source)` matches `src/dataMap/dataMap.ts:625`, and `getSourceData()` really does return copies (`src/dataMap/dataSource.ts:101`, `this.data.map(cloneRow)`).

handsontable/src/helpers/array.ts
budnix: Good spot — the helper JSDoc really doesn't reach the reference, so the ragged-input note now sits on `populateFromArray`'s `@param [method]` in 076dcfc0fb, and that sentence is active now.

handsontable/src/helpers/array.ts
budnix: You're right, it fires for a full-width row too. Documented it in the `pivot` JSDoc and corrected the PR description in 076dcfc0fb.

handsontable/src/helpers/array.ts
budnix: Nice catch — tightened it to `unknown[][]` in 076dcfc0fb, which dropped three casts in `core.ts`. The autocomplete call had no cast to remove.

handsontable/src/helpers/array.ts
budnix: Good point — added a `numberOfDataColumns === 0` early return in 076dcfc0fb, plus a spec for `[[], []]`.

handsontable/src/helpers/__tests__/array.unit.ts
budnix: Agreed — added the `undefined` case in 076dcfc0fb, and covered the non-array row and an all-empty input so the guards are pinned too.

.changelogs/10268.json
budnix: Fair point — renamed it to `13254.json` with `private` in 076dcfc0fb, and the title names `Handsontable.helper.pivot` too now.

docs/content/guides/getting-started/binding-to-data/binding-to-data.md
budnix: Good call — in 076dcfc0fb the snippets build a plain `settings` object instead of calling the constructor, so they read the same for all four frameworks rather than being gated to one.

docs/content/guides/getting-started/binding-to-data/binding-to-data.md
budnix: Yep, my mistake — I re-checked in the browser and it is exactly as you say: the value lands on `source[col][-1]` and displays, and every created row shares that slot. Reworded in 076dcfc0fb.

docs/content/guides/getting-started/binding-to-data/binding-to-data.md
budnix: Nice find — both hooks use `this.render()` as of 076dcfc0fb. One note: `hot?.render()` throws the same error, since the `?.` still reads `hot` before the `const` is set.

demtario (approved):

PR #13255, opened 2026-08-26
DEV-2629: Measure the scenario action, not the harness overhead around it

### LINKED ISSUE #4: Should have working CTRL+C/CTRL+V with Excel and Google Docs
Copy and paste must work well for Excel, Open Office Calc and Google Docs Spreadsheets. Not only single field but also copy and paste of whole areas.

Currently only works paste from Excel.

demtario:
### Context

The PR fixes the performance suite's measurement so the numbers it publishes describe the grid work a scenario performs, rather than the harness overhead around it.

The suite posts a sticky comment on every pull request, and it is the first performance signal a reviewer sees. On #13236 it flagged three regressions that a reading of the diff ruled out, and an external contributor drew a conclusion about the change from the table. DEV-2629 was filed against the reporting layer. Investigating it turned up measurement defects underneath, which this PR fixes. The reporting fixes the task asks for come in a follow-up, because the threshold they need has to be re-derived against measurement that works.

**1. Two scroll scenarios measured nothing after the first iteration.** `scroll-up` and `scroll-left` pass a `resetFn` to `runTracedScenario`; `scroll-down` and `scroll-right` did not. Each iteration issues 500 wheel events, so the grid was already at the far edge by iteration 2. Iterations are averaged, so the published `Total` for Scroll Down was understated roughly 4x (1016 ms against 4070 ms) and Scroll Right roughly 5x.

Controlled before/after on one machine, scripting per iteration in ms:

| | iter 1 | iter 2 | iter 3 | CV |
|---|---|---|---|---|
| scroll-down before | 1107.8 | 70.2 | 83.8 | 115.5% |
| scroll-down after | 1518.1 | 1614.0 | 1446.6 | **4.5%** |
| scroll-right before | 345.7 | 55.2 | 61.7 | 87.8% |
| scroll-right after | 735.5 | 740.6 | 696.2 | **2.7%** |

**2. Nothing waited for the browser to settle before the trace stopped.** `runTracedScenario` called `actionFn` and `Tracing.end` back to back, so rendering and painting that follow the last JS turn fell outside the trace.

**3. The measured window was the harness overhead, not the grid work.** This is the big one, and it is why the zero-baselines in the task exist.

`parseTrace` auto-zoomed onto the busiest region of the trace, following DevTools's `MainThreadActivity.calculateWindow`. For any scenario driven through `page.evaluate`, that region is the V8 interrupt CDP uses to enter the isolate. Sorting, iteration 1:

```
RunTask                          512.2ms
V8.InvokeApiInterruptCallbacks   425.9ms
```

`V8.InvokeApiInterruptCallbacks` maps to no category, so it landed in `other`, which `sumActive()` excludes. Worse, the window it attracted closed at `501595954` while every `Paint`, `Layout`, `PrePaint` and `Commit` sat between `501596016` and `501596034` — roughly 60 ms past the edge. All of them scored zero.

So the recurring `rendering: 0` / `painting: 0` baselines and the `141.42%` CV (which is exactly the CV of `{x, 0, 0}` at n=3) were never cheap operations, and not merely a short trace. They were work that fell outside the window. Fixing #2 alone does not help: the settle puts the paint inside the trace, and the window calculation then discards it.

The clearest symptom is that the published number was smaller than the operation it named. Filtering reported a 25 ms `Total` against a 47 ms measured filter hook; sorting reported 16 ms against an 80 ms sort hook.

The fix brackets the measured action with `performance.mark`, records `blink.user_timing`, and has the parser prefer that window. Traces recorded without the marks still fall back to `calculateWindow`, and teardown now warns when that happens.

**4. Harness work was still landing inside the new window** (found by code review on this PR, commit `5f4d316b8`). Three paths: a `resetFn` renders synchronously but paints a frame later, and the runner did not settle after it, so that paint was billed to the next iteration's action; `getHookTiming()` is a `page.evaluate` that ran inside `actionFn` under `if (isMeasured)`, giving measured iterations a CDP round trip the warmups did not have, on filtering and sorting only; and the settle's stated 1 s bound only covered its idle callback, not the two `requestAnimationFrame` waits ahead of it.

### Results

Intra-run CV of scripting, the currently published baseline against this branch:

| Scenario | before | after | | Scenario | before | after |
|---|---|---|---|---|---|---|
| scroll-down | 126.8% | **1.2%** | | sorting | 97.8% | **3.6%** |
| scroll-right | 105.9% | **5.3%** | | cell-editing | 3.3% | **3.7%** |
| scroll-left | 2.3% | **1.5%** | | source-data-validator-load | 10.8% | **3.7%** |
| scroll-up | 0.6% | **2.4%** | | filtering | 58.9% | **12-23%** |
| | | | | initial-load | 53.5% | **17-24%** |

Rendering and painting CV was `141.42%` on four scenarios and is now under 10% on all nine.

Filtering and initial-load are given as ranges because they genuinely move between runs, and that is worth stating plainly rather than quoting whichever single run looked best. **`initial-load`'s spread is the scenario, not the harness.** Across four consecutive runs its third iteration is systematically slower than the first two — `(14.2, 14.9, 23.6)`, `(16.7, 14.8, 22.0)`, `(14.9, 14.9, 23.6)`, `(16.0, 15.0, 23.9)` — which is the grid being built a fourth time on an accumulated heap. The marked window exposes that; the auto-zoom window was measuring something else. Surfacing it is PR 3's job, not suppressing it.

The cross-check that the window is now correct, per iteration: trace-derived `sorting` scripting reads `74.7 / 84.0 / 78.3` against independent `beforeColumnSort`/`afterColumnSort` hook deltas of `79.9 / 85.1 / 76.0`. Two unrelated measurements of the same operation now agree.

### :warning: This invalidates every existing baseline

The marked window changes every scenario's numbers, in both directions:

- `source-data-validator-load` scripting goes from `14 ms` to roughly `1300 ms`. This is real. Its fixture is identical to `initial-load`'s except for one option, `sourceDataValidator`, over the same 100000 x 100 dataset, and the marked window brackets only `__build()`. The gap against `initial-load`'s `~15 ms` is the per-cell validator across 10 million cells, which the old window was hiding.
- `cell-editing` goes from `324 ms` to roughly `127 ms` — partly setup that used to sit inside the window, partly the reset paint removed by fix #4.

The first golden snapshot recorded on `develop` after this merges is incomparable to every prior one, so **every open pull request will show a wave of large deltas the moment this lands**. That is the fix surfacing, not causing, the problem.

### Follow-up

This is the first of three PRs on DEV-2629. It fixes measurement only. Still to come:

- **PR 2 — baseline comparability.** Stop stripping `_iterationValues` before the golden snapshot is written, add provenance (commit SHA, run id) to the snapshot, and compare against a rolling median of the last N `develop` snapshots instead of whatever `latest.json` holds at the moment. Six consecutive `develop` baselines measuring identical code currently spread 4x on filtering. The median window must exclude pre-PR-1 snapshots, which is why the provenance field comes first.
- **PR 3 — reporting.** The three items the task actually filed: guard the total against an incomplete baseline, surface CV in the markdown comment (`fmtCv()` has zero callers today), and gate the regression callouts on it. `REGRESSION_CALLOUT_THRESHOLD` is 5 against an empirical noise floor of 12-38%, so it has to be re-derived — which needs this PR merged and several `develop` pushes accumulated first. The HTML report's "Trace window" row also needs attention: mark-to-mark now spans harness wall clock, including `cell-editing`'s ~1.8 s of `delay: 10` typing, and that row still gets a red/green classification.

### How has this been tested?

New unit tests for the window selection, run by the existing tooling runner:

```bash
npm run test:tooling          # 114 tests, includes performance-tests/lib/__tests__/trace-window.test.mjs
npm --prefix performance-tests run lint
npm --prefix performance-tests run typecheck
```

The tests were mutation-checked rather than merely written: flipping the zero-width guard from `max <= min` to `max < min`, and flipping either mark comparison from an exact match to a substring match, each fail the suite independently.

Full suite run locally, all nine scenarios, plus three extra repeats of `initial-load` and `cell-editing` to establish that the remaining spread is systematic rather than a single bad run:

```bash
npm --prefix performance-tests test
```

Verified across all 9 scenarios x 3 iterations (27 traces), not only the ones probed during development:

- `_debug.windowSource` reads `marks` on all 27. No scenario silently fell back to auto-zoom while still looking healthy in the CV table.
- `V8.InvokeApiInterruptCallbacks` inside the marked window: 0 events, 0.0 ms, on all 27.
- `other` collapsed accordingly — sorting `429 ms` to `4.3-9.8 ms`, filtering `487 ms` to `7.5-27.0 ms`.

This PR touches `performance-tests/**`, which sets the `test-performance` scope flag, so the suite runs against itself here.

### Types of changes

### Related issue(s):

1. DEV-2629
2. Found while answering a review question on #13236

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

Test tooling only. The one file outside `performance-tests/` is the repository-root `package.json`, which wires `performance-tests/lib/__tests__/*.test.mjs` into the existing `test:tooling` script. No library source is touched.

[skip changelog]

### Checklist:

- [x] My change requires a change to the documentation. `performance-tests/README.md` is updated in `87d3b3bc2`.

---

> [!NOTE]
> **Medium Risk**
> Changes only the performance-test harness and baselines, not library runtime code, but golden snapshots and PR compare semantics change materially—wrong windows or mixed baselines could mislead reviewers until develop is re-baselined.
> 
> **Overview**
> Fixes **performance-tests** so published CDP numbers describe grid work inside an explicit window, not harness overhead or the wrong trace slice.
> 
> **Runner (`trace-runner.mjs`)** records `blink.user_timing`, brackets each measured iteration with `performance.mark` start/end, waits for compositor settle (double rAF + idle, with timeout warnings) inside that window, settles after `setupFn`/`resetFn` (not gated by `skipSettle`), and adds **`afterActionFn`** for harness readbacks after the end mark.
> 
> **Parser (`trace-parser.mjs`)** prefers the mark-defined window via `measurementWindowFromMarks`; unmarked traces still use DevTools auto-zoom. **UpdateCounters** heap/node extrema are confined to the same window.
> 
> **Teardown** warns on auto-zoom fallback, **fails golden** runs without marks, records `windowSource` on snapshots, and warns when comparing marked runs against pre-marks baselines (`crossWindowScenarios` in report meta).
> 
> **Scenarios:** filtering/sorting move hook timing to `afterActionFn`; scroll-down/right add `resetFn` with scroll helpers. New **`trace-window.test.mjs`**; root **`test:tooling`** includes perf lib tests. README and performance-testing skill document the measured-window rules.
> 
> **Note:** merging invalidates existing golden baselines; expect large PR deltas until a fresh develop golden run.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f7a8ce633f83ef9534e66afaf786b4ee10305473. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/performance-tests/README.md
+++ b/performance-tests/README.md
@@ -101,17 +101,42 @@ performance-tests/
 
 ### Trace pipeline
 
-1. The **spec file** calls `runTracedScenario()`, which starts CDP tracing (`Tracing.start`), executes the action, stops tracing (`Tracing.end`), and writes the raw JSON per iteration.
+1. The **spec file** calls `runTracedScenario()`, which starts CDP tracing (`Tracing.start`), marks the start of the measured window, executes the action, waits for the resulting frame to reach the compositor, marks the end, stops tracing (`Tracing.end`), and writes the raw JSON per iteration.
 
 2. The **globalTeardown** (`lib/teardown.mjs`) discovers all `output/*/iteration-*.json` files and feeds them to the trace parser.
 
-3. The **trace parser** (`trace-parser.mjs`) categorizes every trace event into DevTools-equivalent categories: scripting, rendering, painting, loading, system (other), and idle. It computes the auto-zoomed window (matching DevTools' `MainThreadActivity.calculateWindow`), synthesizes ProfileCall scripting from CPU profile data, and extracts UpdateCounters (heap, nodes, listeners).
+3. The **trace parser** (`trace-parser.mjs`) categorizes every trace event into DevTools-equivalent categories: scripting, rendering, painting, loading, system (other), and idle. It measures the window the runner marked around the action, synthesizes ProfileCall scripting from CPU profile data, and extracts UpdateCounters (heap, nodes, listeners). Traces recorded without the marks fall back to the auto-zoomed window (matching DevTools' `MainThreadActivity.calculateWindow`); teardown warns when that happens, because the two are not comparable -- see [The measured window](#the-measured-window).
 
 4. Results are **averaged** across iterations with per-iteration values retained for CV% (coefficient of variation) calculation.
 
 5. Two **report builders** produce output:
    - **Markdown** (`report-builder.mjs`): compact summary table with regression callouts, posted as a sticky PR comment.
    - **HTML** (`html-report-builder.mjs`): self-contained interactive page with inline SVG bar charts, sortable tables, and scenario detail views. Deployed to GitHub Pages per branch.
+
+### The measured window
+
+Everything the suite publishes describes one slice of the trace, and picking that slice
+correctly is the difference between measuring the grid and measuring the harness.
+
+`runTracedScenario()` brackets the action with `performance.mark` (recorded via the
+`blink.user_timing` category) and the parser measures between those marks. It does **not**
+auto-zoom onto the busiest region of the trace, because for any scenario driven through
+`page.evaluate` the busiest region is the V8 interrupt CDP uses to enter the isolate -- a
+blob of a few hundred milliseconds that maps to no category, lands in `other`, and is
+excluded from `Total`. Worse, a window fitted to it closes before the `Paint`, `Layout` and
+`Commit` events the action caused, so rendering and painting score exactly zero.
+
+Two rules follow for anyone writing or changing a scenario:
+
+- **Keep harness round trips out of the window.** A `page.evaluate` inside `actionFn` that
+  reads a value back is billed to the action. Use `afterActionFn`, which runs after the end
+  mark, for readbacks such as hook timings.
+- **Let the reset settle.** A reset renders synchronously but paints a frame later. The
+  runner settles after `resetFn` for this reason; a `resetFn` that returns before its work is
+  queued will still leak into the next iteration.
+
+A category measured as exactly `0`, or a CV of `141.42%` (the CV of `{x, 0, 0}` at three
+iterations), means the window is wrong -- not that the operation was cheap.
 
 ### Golden baseline workflow
 
@@ -214,7 +239,9 @@ test(config.name, async({ page }) => {
     actionFn: async() => {
       // The measured action
     },
-    // Optional: resetFn to restore state between iterations
+    // Optional: resetFn to restore state between iterations (the runner settles after it)
+    // Optional: afterActionFn for readbacks, which run once the measured window has closed
+    // Optional: settleFn to replace the default settle, or skipSettle to opt out of it
   });
 });
 ```
@@ -224,7 +251,7 @@ test(config.name, async({ page }) => {
 For scenarios that measure a specific Handsontable hook pair (like filtering or sorting):
 
 1. Call `injectHookTimer(page, beforeHook, afterHook)` once before `runTracedScenario`. It is idempotent -- safe to call again in `resetFn` to reset the timer store.
-2. Inside `actionFn`, call `getHookTiming(page, beforeHook, afterHook)` and push `timing.deltaMs` to a deltas array.
+2. In `afterActionFn` -- **not** `actionFn` -- call `getHookTiming(page, beforeHook, afterHook)` and push `timing.deltaMs` to a deltas array. It is a `page.evaluate`, so inside the action its CDP round trip would be measured as part of the operation.
 3. After `runTracedScenario`, call `saveHookTimings(outputDir, deltas)` to persist the data.
 
 All three functions are exported from `lib/hook-timing.mjs`. See the `filtering` or `sorting` scenario specs for the complete pattern.
--- a/performance-tests/README.md
+++ b/performance-tests/README.md
@@ -132,8 +132,10 @@ Two rules follow for anyone writing or changing a scenario:
   reads a value back is billed to the action. Use `afterActionFn`, which runs after the end
   mark, for readbacks such as hook timings.
 - **Let the reset settle.** A reset renders synchronously but paints a frame later. The
-  runner settles after `resetFn` for this reason; a `resetFn` that returns before its work is
-  queued will still leak into the next iteration.
+  runner settles after `setupFn` and `resetFn` for this reason, and `skipSettle` does not
+  reach those -- it gates only the in-window settle. Note that `scrollToRow` and
+  `scrollToColumn` wait on trimming, not scroll position, so they return before the scroll
+  has rendered; the settle is what makes those resets deterministic.
 
 A category measured as exactly `0`, or a CV of `141.42%` (the CV of `{x, 0, 0}` at three
 iterations), means the window is wrong -- not that the operation was cheap.
--- a/performance-tests/lib/__tests__/trace-window.test.mjs
+++ b/performance-tests/lib/__tests__/trace-window.test.mjs
@@ -1,0 +1,92 @@
+// Unit tests for the explicit measurement window.
+//
+// The window decides which slice of a CDP trace the published numbers describe.
+// Before it existed the parser auto-zoomed onto the busiest region, which for a
+// page.evaluate-driven scenario is the V8 interrupt CDP uses to enter the isolate:
+// the window then closed before the Paint and Commit events the action caused, so
+// rendering and painting were published as 0 ms while ~420 ms of harness overhead
+// was published as System.
+
+import { test, describe } from 'node:test';
+import assert from 'node:assert/strict';
+import {
+  measurementWindowFromMarks,
+  MEASURE_START_MARK,
+  MEASURE_END_MARK,
+} from '../../trace-parser.mjs';
+
+/**
+ * @param {string} name
+ * @param {number} ts -- microseconds, matching the trace format
+ * @returns {object} a trace event
+ */
+function markEvent(name, ts) {
+  return { name, ts, ph: 'R', pid: 1, tid: 1 };
+}
+
+describe('measurementWindowFromMarks', () => {
+  test('returns the span between the runner marks, in microseconds', () => {
+    const window = measurementWindowFromMarks([
+      markEvent('RunTask', 500),
+      markEvent(MEASURE_START_MARK, 1000),
+      markEvent('Paint', 1500),
+      markEvent(MEASURE_END_MARK, 3000),
+      markEvent('RunTask', 4000),
+    ]);
+
+    assert.deepEqual(window, { min: 1000, max: 3000, range: 2000 });
+  });
+
+  test('returns null when the trace carries no marks, so the caller falls back to auto-zoom', () => {
+    assert.equal(measurementWindowFromMarks([markEvent('RunTask', 500)]), null);
+  });
+
+  test('returns null when only one side of the window was marked', () => {
+    assert.equal(measurementWindowFromMarks([markEvent(MEASURE_START_MARK, 1000)]), null);
+    assert.equal(measurementWindowFromMarks([markEvent(MEASURE_END_MARK, 1000)]), null);
+  });
+
+  test('returns null on a zero-width or inverted window rather than a bogus range', () => {
+    assert.equal(measurementWindowFromMarks([
+      markEvent(MEASURE_START_MARK, 1000),
+      markEvent(MEASURE_END_MARK, 1000),
+    ]), null);
+
+    assert.equal(measurementWindowFromMarks([
+      markEvent(MEASURE_START_MARK, 3000),
+      markEvent(MEASURE_END_MARK, 1000),
+    ]), null);
+  });
+
+  test('spans the outermost pair when a trace carries more than one of each mark', () => {
+    const window = measurementWindowFromMarks([
+      markEvent(MEASURE_START_MARK, 2000),
+      markEvent(MEASURE_END_MARK, 3000),
+      markEvent(MEASURE_START_MARK, 1000),
+      markEvent(MEASURE_END_MARK, 4000),
+    ]);
+
+    assert.deepEqual(window, { min: 1000, max: 4000, range: 3000 });
+  });
+
+  test('matches mark names exactly, not by substring', () => {
+    // Each case pairs one near-miss name with one real mark, so a substring match
+    // on either side alone produces a window and fails the assertion.
+    assert.equal(measurementWindowFromMarks([
+      markEvent(`${MEASURE_START_MARK}-nested`, 1000),
+      markEvent(MEASURE_END_MARK, 3000),
+    ]), null, 'a name containing the start mark must not open the window');
+
+    assert.equal(measurementWindowFromMarks([
+      markEvent(MEASURE_START_MARK, 1000),
+      markEvent(`prefixed-${MEASURE_END_MARK}`, 3000),
+    ]), null, 'a name containing the end mark must not close the window');
+  });
+
+  test('the mark names are stable -- the runner and the parser must agree on them', () => {
+    // Emitted by lib/trace-runner.mjs, read back here. A rename on one side only
+    // sends every scenario silently back to auto-zoom.
+    assert.equal(MEASURE_START_MARK, 'hot-perf-measure-start');
+    assert.equal(MEASURE_END_MARK, 'hot-perf-measure-end');
+  });
+});
--- a/performance-tests/lib/__tests__/trace-window.test.mjs
+++ b/performance-tests/lib/__tests__/trace-window.test.mjs
@@ -2,43 +2,95 @@
 //
 // The window decides which slice of a CDP trace the published numbers describe.
 // Before it existed the parser auto-zoomed onto the busiest region, which for a
-// page.evaluate-driven scenario is the V8 interrupt CDP uses to enter the isolate:
-// the window then closed before the Paint and Commit events the action caused, so
-// rendering and painting were published as 0 ms while ~420 ms of harness overhead
-// was published as System.
+// page.evaluate-driven scenario can be the V8 interrupt CDP uses to enter the isolate
+// (426.9 ms in a real `sorting` trace). The window then closed before the Paint and
+// Commit events the action caused, so rendering and painting were published as 0 ms
+// while the harness overhead was published as System.
+//
+// The fixtures below mirror the real event shapes, checked against a recorded trace:
+// performance.mark arrives as `ph: 'I'`, `s: 't'`, `cat: 'blink.user_timing'`. Keeping
+// them faithful is what lets these tests fail when the matcher or the window changes.
 
 import { test, describe } from 'node:test';
 import assert from 'node:assert/strict';
 import {
+  parseTrace,
   measurementWindowFromMarks,
   MEASURE_START_MARK,
   MEASURE_END_MARK,
 } from '../../trace-parser.mjs';
 
+const PID = 100;
+const TID = 200;
+
 /**
  * @param {string} name
  * @param {number} ts -- microseconds, matching the trace format
- * @returns {object} a trace event
+ * @param {object} [overrides]
+ * @returns {object} a user-timing mark as Chrome emits it
  */
-function markEvent(name, ts) {
-  return { name, ts, ph: 'R', pid: 1, tid: 1 };
+function markEvent(name, ts, overrides = {}) {
+  return { name, ts, ph: 'I', s: 't', cat: 'blink.user_timing', pid: PID, tid: TID, ...overrides };
+}
+
+/**
+ * @param {string} name
+ * @param {number} ts -- microseconds
+ * @param {number} dur -- microseconds
+ * @param {object} [overrides]
+ * @returns {object} a complete (ph 'X') main-thread event
+ */
+function durationEvent(name, ts, dur, overrides = {}) {
+  return { name, ts, dur, ph: 'X', cat: 'devtools.timeline', pid: PID, tid: TID, ...overrides };
+}
+
+/**
+ * Build a trace shaped like the defect this window exists to fix: a large interrupt
+ * blob that the auto-zoom would settle on, and the real work plus its paint sitting
+ * after it, inside the marks.
+ *
+ * @param {object} [options]
+ * @param {boolean} [options.withMarks=true]
+ * @returns {{traceEvents: Array<object>}} a parseable trace
+ */
+function buildTrace({ withMarks = true } = {}) {
+  const events = [
+    { name: 'thread_name', ph: 'M', pid: PID, tid: TID, args: { name: 'CrRendererMain' } },
+    { name: 'TracingStartedInBrowser', ph: 'I', ts: 0, pid: PID, tid: TID },
+
+    // The decoy: harness overhead, far larger than the work, and first in the trace.
+    durationEvent('RunTask', 1_000, 500_000),
+    durationEvent('V8.InvokeApiInterruptCallbacks', 1_100, 499_000),
+
+    // The work the scenario actually performs, and the frame it produces.
+    durationEvent('RunTask', 600_000, 100_000),
+    durationEvent('FunctionCall', 600_500, 99_000),
+    durationEvent('Layout', 720_000, 30_000),
+    durationEvent('Paint', 760_000, 40_000),
+  ];
+
+  if (withMarks) {
+    events.push(markEvent(MEASURE_START_MARK, 600_000));
+    events.push(markEvent(MEASURE_END_MARK, 900_000));
+  }
+
+  return { traceEvents: events };
 }
 
 describe('measurementWindowFromMarks', () => {
   test('returns the span between the runner marks, in microseconds', () => {
     const window = measurementWindowFromMarks([
-      markEvent('RunTask', 500),
+      durationEvent('RunTask', 500, 100),
       markEvent(MEASURE_START_MARK, 1000),
-      markEvent('Paint', 1500),
+      durationEvent('Paint', 1500, 100),
       markEvent(MEASURE_END_MARK, 3000),
-      markEvent('RunTask', 4000),
     ]);
 
     assert.deepEqual(window, { min: 1000, max: 3000, range: 2000 });
   });
 
   test('returns null when the trace carries no marks, so the caller falls back to auto-zoom', () => {
-    assert.equal(measurementWindowFromMarks([markEvent('RunTask', 500)]), null);
+    assert.equal(measurementWindowFromMarks([durationEvent('RunTask', 500, 100)]), null);
   });
 
   test('returns null when only one side of the window was marked', () => {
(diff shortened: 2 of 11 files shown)


qunabu:

performance-tests/trace-parser.mjs
```diff
@@ -903,6 +941,7 @@ export function parseTrace(traceJson) {
       windowMinMs,
       windowMaxMs,
       windowRangeMs,
+      windowSource: markedWindowUs ? 'marks' : 'auto-zoom',
```
budnix, line 971: `windowSource` is exactly the fact a reader needs, but it never leaves `_debug`, and `stripInternalFields` drops `_debug` before `saveSnapshots` runs. The golden on gh-pages is still the auto-zoomed one (filtering `rendering: 0`, `painting: 0`, `scripting: 13.8`; scroll-up ~9140 ms against scroll-down ~2267 ms), so a `compare` run will show most scenarios as huge regressions in the sticky comment. Could the window source go into the snapshot metadata too, so `compare` can flag or refuse a cross-window comparison? Right now nothing can detect the mismatch -- not here, and not on the next parser change either.

performance-tests/README.md
```diff
@@ -224,7 +251,7 @@ test(config.name, async({ page }) => {
 For scenarios that measure a specific Handsontable hook pair (like filtering or sorting):
 
 1. Call `injectHookTimer(page, beforeHook, afterHook)` once before `runTracedScenario`. It is idempotent -- safe to call again in `resetFn` to reset the timer store.
-2. Inside `actionFn`, call `getHookTiming(page, beforeHook, afterHook)` and push `timing.deltaMs` to a deltas array.
+2. In `afterActionFn` -- **not** `actionFn` -- call `getHookTiming(page, beforeHook, afterHook)` and push `timing.deltaMs` to a deltas array. It is a `page.evaluate`, so inside the action its CDP round trip would be measured as part of the operation.
```
budnix, line 256: This is the right rule, and `.claude/skills/performance-testing/SKILL.md` still teaches the opposite -- line 150 there says `// Inside actionFn, capture timing after the action:`, and line 233 says the parser "computes the auto-zoomed window". `performance-tests/` has no `AGENTS.md`, so that skill is the only doc an agent loads before touching these files. Worth updating in the same PR, otherwise the next scenario brings back the exact round trip this change removes.

performance-tests/lib/__tests__/trace-window.test.mjs
```diff
@@ -0,0 +1,92 @@
...
+  MEASURE_END_MARK,
+} from '../../trace-parser.mjs';
+
+/**
+ * @param {string} name
+ * @param {number} ts -- microseconds, matching the trace format
+ * @returns {object} a trace event
+ */
+function markEvent(name, ts) {
+  return { name, ts, ph: 'R', pid: 1, tid: 1 };
```
budnix: Chrome emits `performance.mark` as an instant event (`ph: 'I'`, `cat: 'blink.user_timing'`), not `ph: 'R'`. It passes today because `measurementWindowFromMarks` only reads `name`, but the "matches mark names exactly" test invites tightening the matcher with a `ph` or `cat` check -- and that would keep all seven tests green while every real trace silently falls back to auto-zoom. Since this file is the only guard on the window, would be good to have the fixture match the real event shape.

```suggestion
  return { name, ts, ph: 'I', s: 't', cat: 'blink.user_timing', pid: 1, tid: 1 };
```

performance-tests/lib/teardown.mjs
```diff
@@ -50,6 +50,23 @@ async function collectScenarioResults() {
...
+    // the DevTools auto-zoom, which lands on the CDP interrupt rather than the grid
+    // work -- the exact defect the marks exist to avoid. windowSource lives in _debug,
+    // which is stripped before the snapshot is saved, so say it out loud here or it is
+    // invisible everywhere.
+    const fellBack = parsedResults
+      .map((result, index) => (result._debug?.windowSource === 'marks' ? null : index + 1))
+      .filter(iteration => iteration !== null);
+
+    if (fellBack.length > 0) {
+      console.warn(
```
budnix: In `compare` mode a warning is fine, someone reads the log. In `golden` mode it is not: `averageParsedTraces` folds the auto-zoomed iteration into the mean (300, 300, 14 gives ~205 ms), and the develop push then saves and deploys that as the baseline for every later PR, with `_debug` already stripped. Could golden mode throw here instead? A poisoned baseline stays silent forever, and re-running develop is cheap.

performance-tests/scenarios/scroll-down/scroll-down.spec.ts
```diff
@@ -23,5 +24,8 @@ test(config.name, async({ page }) => {
         await page.mouse.wheel(0, 350);
       }
     },
+    resetFn: async() => {
+      await scrollToRow(page, 0);
```
budnix, line 28: Heads up that `scrollToRow` does not really wait here. Its `waitForFunction` checks `getRenderableFromVisualIndex(r) !== null`, which only reports trimming, and this fixture hides nothing -- so for row 0 it is already non-null on the first poll and the helper returns before the scroll event, re-render and paint happen. Reset determinism now rests on the new settle, not on the wait the helper advertises. The same helper is `setupFn` in `scroll-up.spec.ts` and `scroll-left.spec.ts`, where no settle follows at all.

performance-tests/lib/trace-runner.mjs
```diff
@@ -77,12 +147,24 @@ export async function runTracedScenario({
...
   warmupRuns = 1,
   iterations = 3,
   outputDir,
 }) {
   await mkdir(outputDir, { recursive: true });
 
+  // A reset renders synchronously but paints a frame later, and specs typically wait
+  // only for the render (cell-editing waits on countRenderedRows). Left unsettled, that
+  // paint lands inside the next iteration's window and is billed to the next action.
+  const settleAfterReset = async() => {
```
budnix: `skipSettle` is documented as "opt out of settling before the trace stops", but it also switches off this reset settle. A scenario that only wants the in-window settle gone loses the reset settle with it, so its reset paint lands in the next iteration's window -- the exact leak the comment right above says the runner prevents. Should `skipSettle` gate only the two in-window calls and leave the reset settle always on?

performance-tests/lib/trace-runner.mjs
```diff
@@ -62,12 +65,79 @@ export async function stopTracing(cdp) {
...
+ * bound named here.
+ *
+ * @param {import('@playwright/test').Page} page
+ * @param {number} [timeoutMs=SETTLE_TIMEOUT_MS] -- upper bound for the idle wait
+ */
+export async function settleFrames(page, timeoutMs = SETTLE_TIMEOUT_MS) {
+  await page.evaluate(timeout => new Promise((resolve) => {
+    const done = () => resolve(null);
+    // Hard ceiling on the whole wait, including the animation frames.
+    const bail = setTimeout(done, timeout);
```
budnix: The bail is armed before the two rAFs, so it races them, not just the idle callback. If the frame after the action needs more than a second of main-thread work -- plausible for `initial-load` and `source-data-validator-load` at 100000 x 100 -- the expired timer can be serviced first, the end mark goes down early, and Paint/Commit fall outside the window. That gives `painting: 0` again with `windowSource: 'marks'`, so the new teardown warning stays quiet. Locally the settle is 19-30 ms, so this only bites on a slow CI box, but it fails silently, which is the part worth guarding.

performance-tests/trace-parser.mjs
```diff
@@ -482,6 +482,42 @@ function computeTraceBounds(events) {
...
+ *
+ * @param {Array<object>} events -- raw trace events
+ * @returns {{min: number, max: number, range: number}|null} bounds in us, or null when unmarked
+ */
+export function measurementWindowFromMarks(events) {
+  let min = null;
+  let max = null;
+
+  for (const event of events) {
+    if (event.name === MEASURE_START_MARK && (min === null || event.ts < min)) {
```
budnix, line 523: This scans every event in the trace, but `parseTrace` picks one main thread and computes all the stats for that thread only. A second renderer (an iframe fixture, an out-of-process frame) or a future `performance.measure` sharing the name would set the bounds from a thread the stats never touch -- and the near-zero categories would then be published as a real improvement, with `windowSource: 'marks'` backing them up. Worth filtering on the resolved `mainThread` pid/tid and on `cat === 'blink.user_timing'`.

performance-tests/lib/trace-runner.mjs
```diff
@@ -62,12 +65,79 @@ export async function stopTracing(cdp) {
...
+ *
+ * Without an explicit window the parser auto-zooms onto the busiest part of the
+ * trace, which for a page.evaluate-driven action is the V8 interrupt CDP uses to
+ * enter the isolate, not the grid work. See measurementWindowFromMarks().
+ *
+ * @param {import('@playwright/test').Page} page
+ * @param {string} markName
+ */
+async function mark(page, markName) {
+  await page.evaluate(name => performance.mark(name), markName);
```
budnix: The concise arrow body makes the `PerformanceMark` the return value, so Playwright serializes and ships it back -- extra round trip time between the start mark and the action, right inside the window. It also ties the window to the serializer accepting a `PerformanceMark`: if it ever rejects one, `mark()` throws and every scenario drops to auto-zoom with just a `console.warn`.

```suggestion
  await page.evaluate((name) => {
    performance.mark(name);
  }, markName);
```

performance-tests/README.md
```diff
@@ -101,18 +101,43 @@ performance-tests/
...
+`blink.user_timing` category) and the parser measures between those marks. It does **not**
+auto-zoom onto the busiest region of the trace, because for any scenario driven through
+`page.evaluate` the busiest region is the V8 interrupt CDP uses to enter the isolate -- a
+blob of a few hundred milliseconds that maps to no category, lands in `other`, and is
+excluded from `Total`. Worse, a window fitted to it closes before the `Paint`, `Layout` and
+`Commit` events the action caused, so rendering and painting score exactly zero.
+
+Two rules follow for anyone writing or changing a scenario:
+
+- **Keep harness round trips out of the window.** A `page.evaluate` inside `actionFn` that
```
budnix, line 131: By this rule `cell-editing` is the biggest offender left: its `actionFn` runs 20 `page.evaluate` calls for `selectCell`, plus about 40 key presses and ~1.8 s of `{ delay: 10 }` idle, all inside the window now. Filtering and sorting got `afterActionFn`, cell-editing was left alone. Did that end up being deliberate -- the typing delay being part of what the scenario measures -- or is it just next on the list?

performance-tests/lib/teardown.mjs
```diff
@@ -50,6 +50,23 @@ async function collectScenarioResults() {
...
     console.log(' done');
 
+    // Every number this suite publishes assumes the window came from the marks the
+    // runner writes around the action. Without them the parser silently falls back to
+    // the DevTools auto-zoom, which lands on the CDP interrupt rather than the grid
+    // work -- the exact defect the marks exist to avoid. windowSource lives in _debug,
+    // which is stripped before the snapshot is saved, so say it out loud here or it is
+    // invisible everywhere.
+    const fellBack = parsedResults
+      .map((result, index) => (result._debug?.windowSource === 'marks' ? null : index + 1))
```
budnix: `index + 1` counts positions in a `.sort()`ed file list, so at 10 or more iterations the labels go wrong: the order becomes iteration-1, iteration-10, iteration-11, iteration-2, and iteration-10 gets reported as "iteration 2". Every config is at 3 today so it is latent, but `iterations` is a per-scenario knob and this message is the only diagnostic the change ships. Pulling the number out of the filename would keep it honest.

performance-tests/lib/trace-runner.mjs
```diff
@@ -110,7 +198,24 @@ export async function runTracedScenario({
...
+    // Inside the window on purpose: the frame this waits for is the work being measured.
+    if (!skipSettle) {
+      await (settleFn ? settleFn() : settleFrames(page));
+    }
+
+    await mark(page, MEASURE_END_MARK);
+
+    // Outside the window: a readback here is harness overhead, not measured work.
+    if (afterActionFn) {
+      await afterActionFn(true);
```
budnix: `afterActionFn` is typed `(isMeasured: boolean)` but the warmup loop never calls it and this is the only call site, so the argument can only ever be `true`. Moving the readbacks out of `actionFn` also removed the last reader of `actionFn`'s own `isMeasured`, so both documented contracts now describe a value nothing uses. Either drop the parameter from both, or run `afterActionFn` on warmups too so the two iteration kinds really do have the same shape.

performance-tests/trace-parser.mjs
```diff
@@ -849,8 +885,10 @@ export function parseTrace(traceJson) {
     e.tid === mainThread.tid
   );
 
-  // Calculate the auto-zoomed window using DevTools's MainThreadActivity.calculateWindow
-  const windowUs = calculateWindow(traceBoundsUs, allMainThreadEntries);
+  // Prefer the window the runner marked around the action. Fall back to DevTools's
+  // MainThreadActivity.calculateWindow for traces recorded without the marks.
+  const markedWindowUs = measurementWindowFromMarks(events);
+  const windowUs = markedWindowUs ?? calculateWindow(traceBoundsUs, allMainThreadEntries);
```
Haxikowy, line 918: Timing goes through `statsForTimeRange(mainThreadEvents, windowMinMs, windowMaxMs)`. Heap, nodes and listeners do not: `computeUpdateCountersRanges(events, mainThread)` at line 927 takes no window and returns running extrema over the whole trace. This PR makes every trace strictly longer, because the settle, the end mark's `page.evaluate` and `afterActionFn` all run before `stopTracing`, so a running max can only rise.

Two synthetic traces, identical inside the marked window, differing only in `UpdateCounters` samples after the end mark:

```
SHORT (pre-PR shape): categories {"other":2.1,"painting":0.5,"rendering":0.4}  heap 120.0 MB  nodes 6000
LONG  (post-PR shape): categories {"other":2.1,"painting":0.5,"rendering":0.4}  heap 260.0 MB  nodes 12000
```

Timing byte-identical, heap doubled. `report-builder.mjs:131` runs `jsHeapMaxBytes` through the same `REGRESSION_CALLOUT_THRESHOLD` as timing, so an inflated heap emits a regression callout in the sticky comment, and heap is the primary metric for the two scenarios in `MEMORY_SCENARIOS`. The description says timing baselines are invalidated; heap is too, in one direction, and a reader will go hunting for a leak.

Either pass `windowUs` into `computeUpdateCountersRanges`, or state the heap invalidation alongside the timing one.

performance-tests/lib/__tests__/trace-window.test.mjs
```diff
@@ -0,0 +1,92 @@
+// Unit tests for the explicit measurement window.
```
Haxikowy, line 1: The file covers the pure helper. Nothing asserts that `parseTrace` prefers the marked window, or that `_debug.windowSource` reports `marks` versus `auto-zoom`, which is the exact string `teardown.mjs:60` keys its warning off.

`parseTrace` is pure over `{ traceEvents }`, so this needs no browser and no fixture. I wrote it: 45 lines, 3 tests, one `thread_name` metadata event plus a fake `V8.InvokeApiInterruptCallbacks` blob, the two marks, and a `Paint` inside the window and one outside. It passes 3/3, and reverting line 891 to `calculateWindow(traceBoundsUs, allMainThreadEntries)` fails 2 of 3:

```
expected: 500  actual: 0
expected: 1    actual: 51
# pass 1  # fail 2
```

That is the test that would have caught the original defect. Happy to hand over the file.

performance-tests/lib/__tests__/trace-window.test.mjs
```diff
@@ -0,0 +1,92 @@
...
+      markEvent(MEASURE_END_MARK, 3000),
+    ]), null, 'a name containing the start mark must not open the window');
+
+    assert.equal(measurementWindowFromMarks([
+      markEvent(MEASURE_START_MARK, 1000),
+      markEvent(`prefixed-${MEASURE_END_MARK}`, 3000),
+    ]), null, 'a name containing the end mark must not close the window');
+  });
+
+  test('the mark names are stable -- the runner and the parser must agree on them', () => {
```
Haxikowy: The comment says a rename on one side sends every scenario back to auto-zoom, but `trace-runner.mjs:6` imports both names from `trace-parser.mjs`. There is one definition site, so a rename is atomic and breaks nothing. The test asserts a module's exported literals against the same literals, which makes it a change detector rather than a guard, and it inflates the apparent coverage of the file.

The `parseTrace` test above would make runner and parser agreement load-bearing for real.

performance-tests/lib/trace-runner.mjs
```diff
@@ -62,12 +65,79 @@ export async function stopTracing(cdp) {
...
+ * callbacks: once frame N+1's callback runs, frame N has been committed. The
+ * idle callback then catches trailing work (deferred layout, async paint).
+ *
+ * The whole wait races a timer, not just the idle callback: requestAnimationFrame
+ * does not fire at all in a throttled or occluded page, so without the race a
+ * stalled renderer would hang until Playwright's per-test timeout instead of the
+ * bound named here.
+ *
+ * @param {import('@playwright/test').Page} page
+ * @param {number} [timeoutMs=SETTLE_TIMEOUT_MS] -- upper bound for the idle wait
```
Haxikowy: `timeoutMs` no longer bounds only the idle wait. Fix #4 in the description is specifically that the old 1 s bound missed the two `requestAnimationFrame` waits, and line 111 now arms the bail ahead of them, which the body comment says. This `@param` is the line a caller reads in autocomplete and it still carries the pre-fix contract.

Separate note on the same function, and on `trace-parser.mjs:490` and the README: I could not reproduce the ~420 ms `V8.InvokeApiInterruptCallbacks` blob any of them cite. Across an idle page, a mark issued in the middle of a 1500 ms synchronous loop, a 420 ms synchronous `page.evaluate`, and three concurrent evaluates fired into an 800 ms loop, the largest interrupt event anywhere in the trace was 4.4 ms, and inside the marked window it was 0 events in every run. Your "0 events, 0.0 ms on all 27" holds, and the reason looks like the page being idle at both mark points, so each evaluate arrives as a queued task rather than an interrupt into running JS. None of that weakens the fix, but if the 420 ms came from a specific trace it is worth keeping, because as written the docs teach a magnitude I cannot get.

performance-tests/trace-parser.mjs
demtario: Done, and your prediction landed on this PR's own run before I got to it: the sticky comment posted `+9746.9%` for `Initial Load (sourceDataValidator)` against exactly the auto-zoomed golden you describe.

`windowSource` now rides on the scenario result rather than `_debug`, so it survives `stripInternalFields` into the snapshot. Teardown compares the two sides and warns when they disagree, treating a baseline with no `windowSource` at all as `auto-zoom` -- which is how every pre-marks snapshot is recognised:

```js
const windowSourceOf = scenario => scenario.windowSource ?? 'auto-zoom';
const crossWindow = (current, baseline) => Object.keys(current)
  .filter(name => baseline?.[name] && windowSourceOf(baseline[name]) !== windowSourceOf(current[name]));
```

Left it as a warning rather than a refusal here, because a compare run against a stale baseline is exactly what happens on every PR until develop records a fresh golden, and blocking that would make the branch unmergeable for the window between the two. The golden path is where it now throws -- see your other comment.

performance-tests/README.md
demtario: Fixed. You were right that it is the only doc an agent loads before touching these files, and `performance-tests/README.md` alone would not have been read.

Both lines you named are updated, and I added a short rules section above the pipeline, since the two failure modes are not obvious from the code:

> - **No harness round trips inside `actionFn`.** A `page.evaluate` that reads a value back is measured as part of the operation. Put readbacks in `afterActionFn`, which runs after the end mark.
> - **A `resetFn` must leave no frame behind.** The runner settles after `setupFn` and `resetFn` for this reason, and `skipSettle` does not turn those off.

Plus the tell, so the next person recognises it without re-deriving it: a category of exactly `0`, or a CV of `141.42%`, means the window is wrong, not that the operation was cheap.

performance-tests/lib/__tests__/trace-window.test.mjs
demtario: Applied, and verified against a recorded trace rather than taking it on trust -- the real event is exactly what you wrote:

```json
{"args":{...},"cat":"blink.user_timing","name":"hot-perf-measure-start","ph":"I","pid":56310,"s":"t","tid":15391981,"ts":504275096250}
```

The sharp part of this is that I then went and added precisely the guard you predicted, in response to your `trace-parser.mjs:507` comment -- the matcher now requires `ph === 'I' && cat === 'blink.user_timing'`. With the old `ph: 'R'` fixture, all seven tests would have stayed green while every real trace fell back to auto-zoom. So this caught a defect that did not exist when you wrote it and did exist an hour later.

performance-tests/lib/teardown.mjs
demtario: Agreed, and changed to a throw in golden mode. Your asymmetry is the right one and I had it wrong: a compare run is read once and discarded, a golden run is deployed as the baseline for every later PR with `_debug` already stripped, so nothing downstream can ever tell.

```js
if (process.env.PERF_MODE === 'golden') {
  throw new Error(`${detail} Refusing to record a golden baseline from it.`);
}

console.warn(`  WARN: ${detail}`);
```

Worth noting for anyone reading this thread that the throw does not gate any pull request: `golden` only runs on the develop push, and re-running it is cheap, exactly as you say.

performance-tests/scenarios/scroll-down/scroll-down.spec.ts
demtario: You are right, and I knew the wait was vacuous when I wrote it -- `getRenderableFromVisualIndex` reports trimming, and this fixture trims nothing, so it returns on the first poll. I matched the sibling specs deliberately and should have said so.

The part I had missed is your last sentence, and it was a real gap: the same helper is `setupFn` in `scroll-up` and `scroll-left`, where no settle followed at all. `setupFn` now settles too, so the reset determinism no longer rests on a wait that does not wait, in either position.

Documented in both `README.md` and the skill rather than left as folklore, since the helper's name actively suggests otherwise.

performance-tests/lib/trace-runner.mjs
demtario: Good catch, and yes -- that was a real hole I introduced. `skipSettle` now gates only the in-window settle; `setupFn` and `resetFn` always settle.

The rule I settled on, in the code comment so it survives:

> Opting out of measuring a frame is a scenario's call; opting into a contaminated window is not.

performance-tests/lib/trace-runner.mjs
demtario: Fixed, and the silent part was the bit that mattered -- `painting: 0` with `windowSource: 'marks'` backing it up is indistinguishable from a cheap operation, so the teardown warning would indeed have stayed quiet.

Rather than remove the race (an occluded page never fires rAF at all, so something has to bound it), `settleFrames` now reports which way it resolved and the runner says so:

```js
const bail = setTimeout(() => resolve(false), timeout);
const finish = () => {
  clearTimeout(bail);
  resolve(true);
};
```

Locally the settle is 19-30 ms as you found, so this should never fire; if it does on a slow CI box it now says which iteration and warns that rendering and painting may be understated.

performance-tests/trace-parser.mjs
demtario: Fixed on both axes you named -- the scan is now restricted to the resolved `mainThread` pid/tid, and to `ph === 'I' && cat === 'blink.user_timing'`, which also keeps a same-named `performance.measure` (`ph` `'b'`/`'e'`) out of the bounds.

Your framing is what made this worth fixing rather than filing: the failure is not a crash, it is near-empty categories published as a real improvement with `windowSource: 'marks'` vouching for them.

Worth recording how close I came to shipping a half-fix. I added the scoping and a unit test, and mutation-tested it -- reverting the call site to `measurementWindowFromMarks(events)` left all tests green, because my test exercised the helper directly and nothing covered the call site. There is now a `parseTrace` test with marks on a foreign renderer that asserts the fallback, and that mutation fails.

performance-tests/lib/trace-runner.mjs
demtario: Applied your suggestion verbatim. Both halves of the reasoning hold -- the serialized `PerformanceMark` is a round trip inside the window, and the dependency on the serializer accepting that object is a way for every scenario to drop to auto-zoom on someone else's version bump.

Kept your point as the comment, since a concise arrow is the obvious thing to write here and someone will:

```js
// Block body on purpose: a concise arrow returns the PerformanceMark, which Playwright
// then serializes and ships back across CDP -- a round trip inside the window, and a
// dependency on the serializer accepting that object at all.
```

performance-tests/README.md
demtario: Not deliberate -- next on the list, and you have picked the right remaining offender.

Splitting it, because the two halves are not the same problem:

- **The 20 `selectCell` evaluates are harness overhead** by the rule as written, and should move. They are also the part that changes `Scripting`.
- **The ~1.8 s of `{ delay: 10 }` is idle**, and idle does not reach the sticky comment's `Total`, which is `sumActive()` = scripting + rendering + painting. It shows up in the HTML report's `Trace window` row and nowhere else. That row has a related problem of its own now -- mark-to-mark is harness wall clock, and `classifyChange()` still gives it a red badge above 10% -- which I have flagged for PR 3.

What stopped me doing it here is that moving selection out of `actionFn` changes what the scenario measures, from "select and edit" to "edit", so it is a scenario-design decision rather than a harness fix, and it moves cell-editing's numbers a third time in one PR. Happy either way -- do you want it in this PR while the baseline is being invalidated anyway, or as its own change with a before/after?

performance-tests/lib/teardown.mjs
demtario: Fixed -- the number comes from the filename now:

```js
const iterationNumbers = traceFiles.map((fp) => {
  const matched = /iteration-(\d+)\.json$/.exec(fp);

  return matched ? matched[1] : '?';
});
```

Latent as you say, but worth doing precisely because this message is the only diagnostic the change ships, and a diagnostic that lies about which iteration failed is worse than none.

performance-tests/lib/trace-runner.mjs
demtario: Right on both counts, and I split them rather than doing one or the other.

`afterActionFn` lost the parameter -- it only ever ran measured, so the flag could only ever be `true`.

`actionFn` keeps its `isMeasured`, and I want to argue for that rather than quietly leaving it: the warmup loop genuinely passes `false`, so the contract is honest even though no scenario reads it today. Running `afterActionFn` on warmups too, your other option, would mean pushing warmup deltas into the hook-timing array, which is the thing the `if (isMeasured)` guard existed to prevent in the first place.

performance-tests/trace-parser.mjs
demtario: This is the one real bug in the set, and you are right about the mechanism: `computeUpdateCountersRanges` takes no window, the extrema are running, and every trace this PR produces is strictly longer. It now takes `windowUs`, and there is a test that fails without it.

One correction to the framing, because I checked before fixing and it changes what a reader should conclude. **It is latent, not active.** Across all nine scenarios, every `UpdateCounters` sample already falls inside the marks -- `before=0, after=0` on all 27 traces, so in-window max and whole-trace max are currently identical. Chrome only emits `UpdateCounters` while there is activity, and the activity is inside the window.

Which means **the heap regressions on this PR's own run are not this.** `Scroll Down +49.2%` and `Scroll Right +18.6%` are the `resetFn` fix: those two scenarios now perform three iterations of scrolling where they previously did roughly one, so the heap is legitimately higher. Attributing them to a windowing artefact would send someone hunting for the wrong thing.

Fixed regardless -- nothing was keeping that `before=0, after=0` true, and your synthetic pair is exactly what a future settle change would produce.

performance-tests/lib/__tests__/trace-window.test.mjs
demtario: You are right, and this is the gap that mattered most in your set. The file guarded the pure helper and nothing asserted the behavior the entire PR rests on.

I wrote it rather than taking yours -- not a comment on the code, only that I wanted the fixture to carry the defect's shape explicitly (an oversized `V8.InvokeApiInterruptCallbacks` blob first, the work and its `Paint` after it) so the test reads as a description of the bug. Same idea as yours, and it fails the same way when line 891 is reverted.

Four mutations now fail the suite, which is the standard I should have held it to the first time:

| mutation | result |
|---|---|
| revert to `calculateWindow` | 1 failure |
| un-window `computeUpdateCountersRanges` | 1 failure |
| drop the `mainThread` scoping | 1 failure |
| drop the `ph`/`cat` guard | 1 failure |

The third only fails because of your review: my first attempt covered the helper and left the call site uncovered, and the mutation ran green.

performance-tests/lib/__tests__/trace-window.test.mjs
demtario: Agreed, and deleted. There is one definition site, `trace-runner.mjs` imports both names from it, so a rename is atomic and the test asserted a module's exported literals against the same literals. It guarded nothing and inflated the file's apparent coverage, which is worse than not being there.

The `parseTrace` tests replace it, and they make runner/parser agreement load-bearing for real, as you say.

performance-tests/lib/trace-runner.mjs
demtario: `@param` fixed -- it now reads "upper bound for the whole wait, frames included", plus a `@returns` for the boolean the function now hands back (the bail no longer resolves silently; see budzio's thread on the same lines).

On the 420 ms: **it is real, and it is still in the trace.** From the current `sorting` iteration-1 on this branch:

```
V8.InvokeApiInterruptCallbacks anywhere in the trace
count: 1 | largest (ms): 426.9
```

`scroll-down` carries one at 47.6 ms, for contrast. So your "0 events, 0.0 ms inside the window" and the blob's existence are both true -- the marks moved it out, which is the whole point of the change.

Your explanation for the non-repro looks right to me: the page is idle at both mark points, so each evaluate arrives as a queued task rather than an interrupt into running JS. The blob appears where CDP enters an isolate that is already busy, which the synthetic shapes you tried do not reproduce.

I have replaced the vague "~420ms" with the measured figure and its provenance in the code comment, so the docs no longer teach a magnitude with nothing behind it:

> ...can be the V8.InvokeApiInterruptCallbacks blob CDP produces to enter the isolate -- measured at 426.9 ms in a real `sorting` trace.

Thanks for pushing on it; "I cannot reproduce the number your comment cites" is the right thing to say even when the fix is sound.

Haxikowy (approved):

budnix (approved):

Issue #7517, opened 2021-02-07
React data model does not update when moving rows


henricook:
### Description
When you are using handsontable-react and bind a `data` model to your HOT instance, it generally updates whenever changes are made (e.g. with setDataAtCell).

It doesn't update for row moves however, as proven by the fiddle below. This took me hours to find, now I have to call setData(instance.getData()) from various places to make sure the model stays in sync.

### Steps to reproduce
1. See fiddle for small example

### Demo
https://jsfiddle.net/uhLx59bt/

### Your environment
* Handsontable version: 8.3.0
* Browser Name and version: Firefox 85
* Operating System: Linux

AMBudnik:
Hi @henricook 

A similar topic has been raised here https://github.com/handsontable/react-handsontable/issues/137 and as you correctly spotted `getData()` is the method to use when you want to get the current states of the dataset.

henricook:
Thanks for tying that in, i wouldn't call moving a row a sorting operation though. It's very much a permanent operation and not something I'd expect to be reversed later. Is there another way of moving rows?

AMBudnik:
Yes, there are 3 ways to move rows.

![image](https://user-images.githubusercontent.com/10757813/107219042-cdd15080-6a10-11eb-8041-b506e31f0cc8.png)

1. use `manualRowMove: true` and allow users to move them as they want but do not change the initial order of rows
2. use the initial order of rows, like `manualRowMove: [1, 4]`
3. turn the plugin in the initialization settings and do not change the order of rows, but use the [`moveRow`](https://handsontable.com/docs/8.3.0/ManualRowMove.html#moveRow)/ [`moveRows`](https://handsontable.com/docs/8.3.0/ManualRowMove.html#moveRows) method

henricook:
Thanks, i'm doing (1) and (3) in my application. So are you saying that moving rows is always perceived as _sorting_ so therefore the react data model will never be updated (but getData() is)?

AMBudnik:
Please dont get me wrong. I am not saying that sorting and moving have some permanent or temporary effect. But there is some truth in that sentence.

We can see here https://jsfiddle.net/handsoncode/62qktLf9/ that we get the same result for state and `getData()` when we use 'remove row' from the context menu. But we get different data when we use 'hide row'.

henricook:
Thanks, I was reading back your statements to ensure I understand. I find it's useful for clear outcomes :-)

[...]

PR #13257, opened 2026-08-26
DEV-2646: Stop a vetoed row or column move recording an undo action

budnix:
### Context

The PR fixes a cancelled row or column move still landing on the undo stack.

Returning `false` from `beforeRowMove` cancels the move — nothing moves and the order map stays identity — but a `row_move` action was still recorded. The user's next Ctrl+Z was then spent undoing nothing, silently swallowing the keystroke. `beforeColumnMove` behaved the same way.

This matters more than its size suggests, because blocking the move via `beforeRowMove` is the pattern we ourselves recommend for keeping a framework's own state as the source of truth.

**Root cause.** `RowMoveAction` / `ColumnMoveAction` recorded from the **before** hook and guarded with `if (rows === false)`. That guard checks the *first argument*, not the hook's return value. `Hooks.run` threads a listener's return value into the next listener's first argument, so the guard could only see a veto raised by a listener that runs **before** the undo listener. There is no such instance listener: UndoRedo registers its actions from its own constructor, plugins register in `enablePlugin` (via `beforeInit`), and settings hooks are attached later still. The old guard therefore caught no plugin or settings veto at all — not merely late-registered ones. (Corrected from an earlier, weaker version of this paragraph — thanks @Haxikowy.)

**Fix.** Record from `afterRowMove` / `afterColumnMove` instead. Those never fire for a vetoed move, so the veto no longer needs to be observed at all. This follows the existing `MoveCellsAction` precedent, which already records from `afterMoveCells` for exactly this reason.

**Slightly wider than the title.** The `after` hooks carry an `orderChanged` argument, and the fix gates on it. That also keeps two other dead entries off the stack: a move that was impossible (`movePossible === false`) and a move that left the order intact (for example dropping a row on itself). Both previously recorded an action that undid nothing. Recording a no-op is a defect rather than a contract, so this is treated as part of the same fix — but it is a real behavior change in those two edge cases and is called out here deliberately.

**Known limitation.** `orderChanged` is arithmetic on the requested indexes, not a real sequence diff, so a trimmed-index move or a NestedRows reparent can change the order while the gate reads `false`. Both of those recorded a dud entry before this PR (the trimmed case computes `from === to` on undo; the NestedRows case dies in `displayAPICompatibilityWarning`), so dropping them is not a regression. Tracked as DEV-2651. `formulas.ts:592-601` gates on the same argument and shares the blind spot.

Two casts (`rows as number[]`, `finalIndex as number`) disappear as a side effect, because the typed `addHook` overload infers the callback parameters from `Events['afterRowMove']`.

### How has this been tested?

New Playwright spec `tests/e2e/row-column-move-undo.spec.ts` (13 tests), driven through the `selection-features` fixture, which gained a settings-registered `beforeRowMove` / `beforeColumnMove` veto toggle. One case uses a real shipped veto instead of that toggle: `manualColumnFreeze` refuses to move a frozen column.

The tests were verified against the unfixed build first. With the source reverted the spec goes **8 fail / 5 pass**: the cases that encode the bug go red, while the "a real move still records one action" cases stay green either way — those are regression guards, not bug-catchers, and the split is what shows the tests are not just restating the implementation. The vetoed-**column** test failing there is what confirmed the column case empirically, rather than by reading the source.

```bash
# Playwright (new spec + the neighbouring moveCells undo spec) — 13 passed
cd tests && npx playwright test e2e/row-column-move-undo.spec.ts e2e/move-cells-undo.spec.ts --project=e2e-main

# Legacy E2E regression sweep — 594 specs, 0 failures
npm run test:e2e --prefix handsontable -- --testPathPattern='undoRedo|manualRowMove|manualColumnMove'

# Lint and types — both clean
```

### Types of changes

### Related issue(s):

1. DEV-2646
2. DEV-2651 — follow-up for the `orderChanged` limitation above
2. Found while re-analysing #7517. This bug is independent of that issue and of React; #7517 itself is answered by DEV-2647 (documentation).

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

**Note for the reviewer of DEV-2647 (docs):** the controlled-grid recipe going into the row-moving guide previously relied on this bug — the phantom action, when undone, re-entered `beforeRowMove` with the inverse move, which the user's handler applied to their own state. With this fix a vetoed move records nothing, so that recipe must not promise any Ctrl+Z behavior. DEV-2647 is being written that way already.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2646

---

> [!NOTE]
> **Medium Risk**
> Changes when undo actions are recorded for manual row/column moves, including intentional behavior shifts for vetoed, impossible, and no-op moves; regression risk is mitigated by new e2e coverage.
> 
> **Overview**
> Fixes **phantom undo entries** when a row or column move is cancelled via `beforeRowMove` / `beforeColumnMove` (or never actually changes order). Undo recording for `RowMoveAction` and `ColumnMoveAction` now listens on **`afterRowMove`** / **`afterColumnMove`** instead of the before hooks, matching **`MoveCellsAction`**. Vetoed moves never fire those hooks, so they no longer land on the stack and waste the next Ctrl+Z.
> 
> Recording is gated on **`orderChanged`** and a valid moved-index array, so **no-op** moves (e.g. drop on the same index) and **impossible** moves also skip the undo stack.
> 
> Adds a Playwright spec and extends the **selection-features** fixture with settings-registered move vetoes, move/freeze helpers, and index-mapper order assertions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1a894b57e77b39c9e5bfe728b96a3d731d92f60b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changelogs/13257.json
+++ b/.changelogs/13257.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Fixed a cancelled row or column move still adding an entry to the undo stack. Blocking a move by returning `false` from `beforeRowMove` or `beforeColumnMove` no longer leaves an action that makes the next undo do nothing. Moves that were impossible, or that left the order unchanged, are no longer recorded either.",
+  "type": "fixed",
+  "issueOrPR": 13257,
+  "breaking": false,
+  "framework": "none"
+}
--- a/.changelogs/13257.json
+++ b/.changelogs/13257.json
@@ -1,6 +1,6 @@
 {
   "issuesOrigin": "private",
-  "title": "Fixed a cancelled row or column move still adding an entry to the undo stack. Blocking a move by returning `false` from `beforeRowMove` or `beforeColumnMove` no longer leaves an action that makes the next undo do nothing. Moves that were impossible, or that left the order unchanged, are no longer recorded either.",
+  "title": "Fixed a row or column move blocked by `beforeRowMove` or `beforeColumnMove` still adding an entry to the undo stack; impossible moves and moves that leave the order unchanged are no longer recorded either.",
   "type": "fixed",
   "issueOrPR": 13257,
   "breaking": false,
--- a/handsontable/src/plugins/undoRedo/actions/rowMove.ts
+++ b/handsontable/src/plugins/undoRedo/actions/rowMove.ts
@@ -29,16 +29,24 @@ export class RowMoveAction extends BaseAction {
   }
 
   /**
-   * Registers the `beforeRowMove` hook listener that records a new RowMoveAction whenever rows are moved.
+   * Registers the `afterRowMove` hook listener that records a new RowMoveAction whenever rows are moved.
+   *
+   * Recording runs on `afterRowMove`, not `beforeRowMove`. `Hooks.run` threads a listener's return value
+   * into the next listener's first argument, so a `beforeRowMove` listener only sees a veto raised by a
+   * listener registered before it — a `return false` from user settings lands too late and a cancelled
+   * move still reached the stack. `afterRowMove` never fires for a vetoed move, and its `orderChanged`
+   * argument is `false` when the move was impossible or left the order intact, so gating on it also keeps
+   * no-op moves off the stack.
    */
   static startRegisteringEvents(hot: HotInstance, undoRedoPlugin: unknown) {
-    hot.addHook('beforeRowMove', (rows: unknown, finalIndex: unknown) => {
-      if (rows === false) {
+    hot.addHook('afterRowMove', (movedRows, finalIndex, _dropIndex, _movePossible, orderChanged) => {
+      // `Array.isArray` also covers garbage folded into the argument by a preceding listener.
+      if (!orderChanged || !Array.isArray(movedRows)) {
         return;
       }
 
       (undoRedoPlugin as { done: (...args: unknown[]) => void }).done(
-        () => new RowMoveAction({ rows: rows as number[], finalIndex: finalIndex as number })
+        () => new RowMoveAction({ rows: movedRows, finalIndex })
       );
     });
   }
--- a/handsontable/src/plugins/undoRedo/actions/rowMove.ts
+++ b/handsontable/src/plugins/undoRedo/actions/rowMove.ts
@@ -31,16 +31,19 @@ export class RowMoveAction extends BaseAction {
   /**
    * Registers the `afterRowMove` hook listener that records a new RowMoveAction whenever rows are moved.
    *
-   * Recording runs on `afterRowMove`, not `beforeRowMove`. `Hooks.run` threads a listener's return value
-   * into the next listener's first argument, so a `beforeRowMove` listener only sees a veto raised by a
-   * listener registered before it — a `return false` from user settings lands too late and a cancelled
-   * move still reached the stack. `afterRowMove` never fires for a vetoed move, and its `orderChanged`
-   * argument is `false` when the move was impossible or left the order intact, so gating on it also keeps
-   * no-op moves off the stack.
+   * Recording runs on `afterRowMove`, not `beforeRowMove`. UndoRedo registers its actions from its own
+   * constructor, while plugins register in `enablePlugin` (via `beforeInit`) and settings hooks are
+   * attached later still, so this is always the first instance listener on the hook. `Hooks.run` threads
+   * a listener's return value into the next listener's first argument, so every veto is raised after this
+   * listener has already run: the old `rows === false` guard caught no plugin or settings veto at all, and
+   * a cancelled move always reached the stack. `afterRowMove` never fires for a vetoed move, and its
+   * `orderChanged` argument is `false` when the move was impossible or left the order intact, so gating on
+   * it also keeps no-op moves off the stack.
    */
   static startRegisteringEvents(hot: HotInstance, undoRedoPlugin: unknown) {
     hot.addHook('afterRowMove', (movedRows, finalIndex, _dropIndex, _movePossible, orderChanged) => {
-      // `Array.isArray` also covers garbage folded into the argument by a preceding listener.
+      // Only a global `Handsontable.hooks.add` listener runs ahead of this one, and its return value
+      // would replace `movedRows` — guard the shape before `RowMoveAction` calls `.slice()` on it.
       if (!orderChanged || !Array.isArray(movedRows)) {
         return;
       }
--- a/tests/e2e/row-column-move-undo.spec.ts
+++ b/tests/e2e/row-column-move-undo.spec.ts
@@ -90,4 +90,40 @@ test.describe('row and column move undo bookkeeping', () => {
     expect(await grid.rowOrder()).toEqual(IDENTITY);
     expect(await grid.doneActionsCount()).toBe(0);
   });
+
+  test('a column move that leaves the order unchanged records no undo action', async () => {
+    await grid.moveColumns([0], 0);
+
+    expect(await grid.columnOrder()).toEqual(IDENTITY);
+    expect(await grid.doneActionsCount()).toBe(0);
+  });
+
+  test('an impossible row move records no undo action', async () => {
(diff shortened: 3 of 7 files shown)


qunabu:
Task linked: [DEV-2646 Fix: a vetoed beforeRowMove / beforeColumnMove still records an undo action](https://app.clickup.com/t/9015210959/DEV-2646)
Task linked: [DEV-2647 Fix the React row-moving docs: the documented getData() recipe double-applies the move](https://app.clickup.com/t/9015210959/DEV-2647)

handsontable/src/plugins/undoRedo/actions/rowMove.ts
```diff
@@ -29,16 +29,24 @@ export class RowMoveAction extends BaseAction {
...
+   * move still reached the stack. `afterRowMove` never fires for a vetoed move, and its `orderChanged`
+   * argument is `false` when the move was impossible or left the order intact, so gating on it also keeps
+   * no-op moves off the stack.
    */
   static startRegisteringEvents(hot: HotInstance, undoRedoPlugin: unknown) {
-    hot.addHook('beforeRowMove', (rows: unknown, finalIndex: unknown) => {
-      if (rows === false) {
+    hot.addHook('afterRowMove', (movedRows, finalIndex, _dropIndex, _movePossible, orderChanged) => {
+      // `Array.isArray` also covers garbage folded into the argument by a preceding listener.
+      if (!orderChanged || !Array.isArray(movedRows)) {
```
Haxikowy, line 47: `orderChanged` is not a sequence diff. `isRowOrderChanged` is `movedRows.some((row, i) => row - i !== finalIndex)`, arithmetic on the requested visual indexes, while `moveIndexes` recomputes the destination across the full sequence and skips trimmed indexes. Two cases perform the move and fail this gate:

- sequence `[0,1,2]` with physical 1 trimmed, `moveRows([0], 0)` rewrites it to `[1,0,2]`, `orderChanged` is `false`
- NestedRows dragging a row onto the next one: `willDataChange` is true and `dataManager.moveRow` reparents, but `rowMoveController.ts:155` hand-fires the hook with `orderChanged` `false`

Neither is a regression. I checked what the old entry did on undo: the trimmed case computes `from === to` and moves nothing, the NestedRows case dies in `displayAPICompatibilityWarning`. Both were duds. Follow-up ticket plus a line in the description rather than a change here. `formulas.ts:531` gates on the same argument and inherits the same blind spot. Same in `columnMove.ts`.

handsontable/src/plugins/undoRedo/actions/rowMove.ts
```diff
@@ -29,16 +29,24 @@ export class RowMoveAction extends BaseAction {
...
+   * listener registered before it — a `return false` from user settings lands too late and a cancelled
+   * move still reached the stack. `afterRowMove` never fires for a vetoed move, and its `orderChanged`
+   * argument is `false` when the move was impossible or left the order intact, so gating on it also keeps
+   * no-op moves off the stack.
    */
   static startRegisteringEvents(hot: HotInstance, undoRedoPlugin: unknown) {
-    hot.addHook('beforeRowMove', (rows: unknown, finalIndex: unknown) => {
-      if (rows === false) {
+    hot.addHook('afterRowMove', (movedRows, finalIndex, _dropIndex, _movePossible, orderChanged) => {
+      // `Array.isArray` also covers garbage folded into the argument by a preceding listener.
```
Haxikowy: Overstated. UndoRedo registers in its constructor (`undoRedo.ts:85`), ahead of every `enablePlugin` listener, so no instance listener can precede this one. Only a global `Handsontable.hooks.add` return value can replace `movedRows`. Reword, or drop the guard and the comment with it. Same in `columnMove.ts`.

handsontable/src/plugins/undoRedo/actions/rowMove.ts
```diff
@@ -29,16 +29,24 @@ export class RowMoveAction extends BaseAction {
   }
 
   /**
-   * Registers the `beforeRowMove` hook listener that records a new RowMoveAction whenever rows are moved.
+   * Registers the `afterRowMove` hook listener that records a new RowMoveAction whenever rows are moved.
+   *
+   * Recording runs on `afterRowMove`, not `beforeRowMove`. `Hooks.run` threads a listener's return value
+   * into the next listener's first argument, so a `beforeRowMove` listener only sees a veto raised by a
+   * listener registered before it — a `return false` from user settings lands too late and a cancelled
```
Haxikowy: Not the reason, and the real one is stronger. No instance listener could ever precede this one at any registration order: UndoRedo registers in its constructor, everything else registers in `enablePlugin`, which runs later from `beforeInit`. The old guard was dead for every plugin veto and every settings veto, not only late ones.

.changelogs/13257.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Fixed a cancelled row or column move still adding an entry to the undo stack. Blocking a move by returning `false` from `beforeRowMove` or `beforeColumnMove` no longer leaves an action that makes the next undo do nothing. Moves that were impossible, or that left the order unchanged, are no longer recorded either.",
```
Haxikowy: Three sentences, ~290 chars. One sentence:

`Fixed a row or column move blocked by beforeRowMove or beforeColumnMove still adding an entry to the undo stack; impossible moves and moves that leave the order unchanged are no longer recorded either.`

tests/e2e/row-column-move-undo.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+
+    expect(await grid.rowOrder()).toEqual([1, 2, 3, 0, 4, 5, 6, 7, 8, 9]);
+    expect(await grid.doneActionsCount()).toBe(1);
+
+    await grid.undo();
+
+    expect(await grid.rowOrder()).toEqual(IDENTITY);
+  });
+
+  test('a real column move still records one undo action, and undo reverses it', async () => {
```
Haxikowy, line 76: No redo case. `redo()` replays the same recorded arguments this PR changed, and `move-cells-undo.spec.ts` asserts the round trip. Also missing: keyboard Ctrl+Z (the symptom the description names), a multi-index move, and one case with `trimRows` or `hiddenRows` active.

qunabu:
Task linked: [DEV-2651 orderChanged is arithmetic, not a sequence diff — moves that change order can read false](https://app.clickup.com/t/9015210959/DEV-2651)

handsontable/src/plugins/undoRedo/actions/rowMove.ts
budnix: Agreed, and confirmed — filed [DEV-2651](https://app.clickup.com/t/86cbak97p) and added a line to the description. `formulas.ts:592` shares the gate too.

handsontable/src/plugins/undoRedo/actions/rowMove.ts
budnix: You're right, only a global `Handsontable.hooks.add` can replace it — reworded in bfe12b2a6c. Kept the guard since that path is still real.

handsontable/src/plugins/undoRedo/actions/rowMove.ts
budnix: Good point, and yours is the stronger reason — rewrote the JSDoc in bfe12b2a6c to say no instance listener can ever precede it.

.changelogs/13257.json
budnix: Fair point — used your sentence in bfe12b2a6c, down to 205 chars.

tests/e2e/row-column-move-undo.spec.ts
budnix: Nice catch — added redo, keyboard Ctrl+Z, a multi-row move and a `trimRows` case in bfe12b2a6c.

Haxikowy (approved):

PR #13258, opened 2026-08-26
DEV-2585: Fix loadData() and updateData() discarding a changed formula

adrianspdev:
### Context

The PR fixes a regression on `develop`: `loadData()`, `updateData()` and `updateSettings({ data })` cannot replace an existing formula's text. The old formula silently survives the load, and a load that turns a formula cell into a plain literal resurrects the old formula instead.

Follow-up to #13215 (DEV-2585), which is the change that exposed it.

`#getProcessedSourceDataArray()` reads the source data through the `modifySourceData` hook, and the plugin's own `#onModifySourceData` answers any formula cell with the formula the **engine** already holds. That projection is correct in general — it is how the grid reports `=A1+1` rather than a cached value. But #13215 made the plugin reuse its HyperFormula sheet across data loads instead of recreating it (a legitimate memory-leak fix), so the engine still holds the **previous** content at the moment `#onAfterLoadData` reads the newly loaded data. The stale formula is read back and written straight into the engine again, discarding the load.

Before #13215 every load got a fresh, empty sheet, so `getCellType()` returned `EMPTY` for every cell, the projection never fired, and the real data reached the engine. Neither change is wrong on its own; the sheet reuse simply unmasked a latent bug in the load path.

Reproduction on `develop`, with no cell types and no date handling involved:

```js
  data: [[1], ['=A1+1']],
  formulas: { engine: HyperFormula },
});

hot.loadData([[1], ['=A1+100']]);
// engine keeps `=A1+1` and computes 2, instead of `=A1+100` and 101
```

The fix suppresses the projection for the duration of that read, using a dedicated `#sourceDataProjectionSuspended` flag that only `#onModifySourceData` reads.

The guard belongs in the helper rather than at each call site because all five of its callers feed the result **to** the engine, and handing the engine back its own projected formulas is never meaningful.

**Why a dedicated flag rather than `#internalOperationPending`.** The first revision of this PR reused `#internalOperationPending`, which is the flag `#syncFormulasToSourceData` and `#onAfterDetachChild` already use for the same "compare against what Handsontable stores" purpose. That flag is too broad here. It also gates `#onModifyData` and `#onAfterRenderer`, and this read runs the `modifyRowData` and `modifySourceData` hooks for **every row and cell** — so third-party handlers execute inside the guarded window. A handler that called `getDataAtCell()` on a formula cell would have received the raw `=A1+1` instead of the calculated value. The dedicated flag scopes the suspension to the one hook the fix is about.

The previous flag value is saved and restored rather than cleared. No other site sets it, so this is not about a caller that already holds it; it is about those same third-party handlers, one of which could re-enter this method synchronously (through `updateSettings()` or `loadData()`) and otherwise leave the rest of the outer read unguarded.

**Blast radius.** `#getProcessedSourceDataArray` has five callers, and behavior changes at **two** of them — `#onAfterCellMetaReset` and `#onAfterLoadData`. The two `addSheet` sites are gated on `!doesSheetExist(sheetName)` and `sheetName === null` respectively, and `#onModifySourceData` already bails on both conditions, so the projection was inert there. `#onAfterDetachChild` already holds `#internalOperationPending` across the call.

**Two adjacent changes**, both from review:

- `#getProcessedSourceDataArray` now checks the data shape with `Array.isArray(getSourceDataAtRow(0))` instead of `isArrayOfArrays(getSourceData())`. The latter rebuilt the whole dataset through the per-cell `modifySourceData` hook — and outside the guard, so fully projected — to inspect the shape of one row. `#doesEngineHoldPhysicalColumns` already uses that form for the same reason.
- The `#syncFormulasToSourceData` doc comment claimed that moves and sorting are covered because "the read-time projection keeps handling it". That is no longer true at the engine-feeding reads, so the comment now says what actually holds: ordinary reads are still projected, the engine-feeding reads are not, and in a moved or sorted frame those reads hand the engine the stored, possibly stale text.

**Why this was not caught earlier:** the neighbouring `loadData`/`updateData` recalculation specs all start from `data: [[0]]`, a single non-formula cell, so none of them ever *replaced* an existing formula.

**Known limitation, pre-existing and not addressed here.** While rows or columns are moved or sorted, the engine and the source data stop sharing a reference frame, `#syncFormulasToSourceData` bails out entirely, and any engine-feeding read maps the formula's references a second time. The result is `#REF!` — both with and without this fix, verified by measurement. A data-less `updateSettings({ colHeaders: true })` on a moved frame is enough to reproduce it on `develop`:

```js
  data: [[1], [2], ['=A1+1']],
  manualRowMove: true,
  formulas: { engine: HyperFormula },
});

hot.getPlugin('manualRowMove').moveRow(2, 0);
hot.render();
// grid shows 2 — correct

hot.updateSettings({ colHeaders: true });
// grid shows #REF! — with this PR and without it, byte-identical
```

Narrowing the `isHfOrderPhysical()` bail so a structural change in that state still syncs is a separate design question for the team, tracked separately.

[skip changelog]

No changelog entry: this regression is unreleased. The `settingsSheetName ?? this.sheetName` line traces only to `c0d25ef1c1` on `develop` and the `18.1.0-rc` / `18.1.0-rc1` tags, so no user of a stable release ever saw it. An entry would document a defect that never shipped. Flagging explicitly because `[skip changelog]` is normally reserved for docs-, test-, and CI-only PRs — this is a source change, and the skip is deliberate.

- Unit tests added/modified (`*.unit.js`): none — the defect is in hook interplay across a data load, which the E2E tier exercises
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`): none — extended the existing Jasmine formulas spec instead, alongside the `loadData`/`updateData` recalculation cases it belongs with

For a bug fix — the specs that fail without this fix, all in `handsontable/src/plugins/formulas/__tests__/formulas.spec.js`:

1. `should replace an existing formula's text when the loadData method is called`
2. the same for `updateData`
3. and 4. the two formula-becomes-a-literal counterparts
5. `should replace an existing formula's text when the data is replaced through updateSettings` — a third public entry point, and it reaches the engine through a different site: `#onAfterLoadData` returns early for the `updateSettings` source, so the data is fed from `#onAfterCellMetaReset` instead
6. `... for array of objects` — the first four cases are all array-of-arrays
7. `... for data with skipped columns` — `columns: [{ data: 0 }, { data: 2 }]` over three-column data, the only way to reach the projected branch of `#getProcessedSourceDataArray`
8. `should keep an array formula spilling when the loadData method is called` — spill cells report the `ARRAY` cell type, not `ARRAYFORMULA`, so they pass the `VALUE`/`EMPTY` early return and get projected as their calculated literals. Feeding those back puts a literal inside the range the array formula needs, and the load collapses to `#SPILL!`

Discrimination was verified by removing the production guard, rebuilding and re-running: **exactly those eight cases fail** (372 specs, 8 failures), and they all pass again once it is restored (372 specs, 0 failures).

### Commands run

```
$ npm --prefix handsontable run test:e2e
9726 specs, 0 failures, 8 pending specs

$ npm --prefix handsontable run test:e2e -- --testPathPattern=formulas
372 specs, 0 failures

$ npm --prefix handsontable run test:unit
Tests: 3802 passed, 3802 total (319 suites)

$ npm --prefix handsontable run test:types
pass

$ npm --prefix handsontable run eslint
0 errors (602 pre-existing warnings in untouched files, 0 on changed files)

$ npm --prefix handsontable run build
handsontable.js, handsontable.min.js, handsontable.full.js, handsontable.full.min.js
```

With the production guard temporarily removed, to prove the tests discriminate:

```
$ npm --prefix handsontable run test:e2e -- --testPathPattern=formulas
372 specs, 8 failures
```

### Types of changes

### Related issue(s):

1. DEV-2585
2. #13215

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes formula engine sync on bulk data reload paths in the formulas plugin; blast radius is limited to reads that feed HyperFormula, with broad E2E coverage added for the failure modes.
> 
> **Overview**
> Fixes a regression where **`loadData()`**, **`updateData()`**, and **`updateSettings({ data })`** could not replace an existing formula: the HyperFormula sheet is reused across loads, and reads that feed the engine went through **`modifySourceData`**, which projected each formula cell back to the engine’s **previous** formula and silently dropped the newly loaded text (including formula → literal cases).
> 
> The formulas plugin now suspends that projection only while **`#getProcessedSourceDataArray`** reads source data, via a dedicated **`#sourceDataProjectionSuspended`** flag (narrower than **`#internalOperationPending`** so third-party hooks still get calculated values from **`getDataAtCell`**). **`#onModifySourceData`** respects the same flag for those engine-feeding reads.
> 
> Adjacent fixes: the array-of-arrays shape check uses **`getSourceDataAtRow(0)`** instead of rebuilding all source data through unguarded hooks; skipped-column projection handles **`null`** / non-array rows; **`#syncFormulasToSourceData`** docs note that engine-feeding reads no longer project formulas after row/column moves or sorting.
> 
> Eight new Jasmine specs cover **`loadData`/`updateData`**, **`updateSettings`**, array-of-objects, skipped columns, empty rows, and array-formula spill on reload.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 77d4bfbff2d9ac3f0568946706c4adfa6fb5890f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/handsontable/src/plugins/formulas/__tests__/formulas.spec.js
+++ b/handsontable/src/plugins/formulas/__tests__/formulas.spec.js
@@ -2762,6 +2762,61 @@ describe('Formulas general', () => {
     expect(getDataAtCell(1, 0)).toBe(101);
   });
 
+  it('should replace an existing formula with a literal value when the `loadData` method is called', async() => {
+    handsontable({
+      data: [
+        [1],
+        ['=A1+1'],
+      ],
+      formulas: {
+        engine: HyperFormula,
+      }
+    });
+
+    const formulasPlugin = getPlugin('formulas');
+
+    // The same defect as in the two cases above, with the loaded value no longer being a formula.
+    // The `modifySourceData` projection answers a formula cell with the formula the engine holds, so
+    // reading the newly loaded data through it resurrects `=A1+1` into what is now a value cell.
+    await loadData([
+      [1],
+      [5],
+    ]);
+
+    expect(formulasPlugin.engine.getSheetSerialized(formulasPlugin.sheetId)).toEqual([
+      [1],
+      [5],
+    ]);
+
+    expect(getDataAtCell(1, 0)).toBe(5);
+  });
+
+  it('should replace an existing formula with a literal value when the `updateData` method is called', async() => {
+    handsontable({
+      data: [
+        [1],
+        ['=A1+1'],
+      ],
+      formulas: {
+        engine: HyperFormula,
+      }
+    });
+
+    const formulasPlugin = getPlugin('formulas');
+
+    await updateData([
+      [1],
+      [5],
+    ]);
+
+    expect(formulasPlugin.engine.getSheetSerialized(formulasPlugin.sheetId)).toEqual([
+      [1],
+      [5],
+    ]);
+
+    expect(getDataAtCell(1, 0)).toBe(5);
+  });
+
   it('should display calculated formula after changing value using `beforeChange` hook #6932', async() => {
     handsontable({
       data: [
--- a/handsontable/src/plugins/formulas/__tests__/formulas.spec.js
+++ b/handsontable/src/plugins/formulas/__tests__/formulas.spec.js
@@ -2817,6 +2817,129 @@ describe('Formulas general', () => {
     expect(getDataAtCell(1, 0)).toBe(5);
   });
 
+  it('should replace an existing formula\'s text when the data is replaced through `updateSettings`', async() => {
+    handsontable({
+      data: [
+        [1],
+        ['=A1+1'],
+      ],
+      formulas: {
+        engine: HyperFormula,
+      }
+    });
+
+    const formulasPlugin = getPlugin('formulas');
+
+    // A third public entry point for the same defect, and it reaches the engine through a different
+    // site: `#onAfterLoadData` returns early for the `updateSettings` source, so the data is fed to
+    // the engine from `#onAfterCellMetaReset` instead.
+    await updateSettings({
+      data: [
+        [1],
+        ['=A1+100'],
+      ],
+    });
+
+    expect(formulasPlugin.engine.getSheetSerialized(formulasPlugin.sheetId)).toEqual([
+      [1],
+      ['=A1+100'],
+    ]);
+
+    expect(getDataAtCell(1, 0)).toBe(101);
+  });
+
+  it('should replace an existing formula\'s text when the `loadData` method is called for array of objects', async() => {
+    handsontable({
+      data: [
+        { value: 1 },
+        { value: '=A1+1' },
+      ],
+      columns: [
+        { data: 'value' },
+      ],
+      formulas: {
+        engine: HyperFormula,
+      }
+    });
+
+    const formulasPlugin = getPlugin('formulas');
+
+    await loadData([
+      { value: 1 },
+      { value: '=A1+100' },
+    ]);
+
+    expect(formulasPlugin.engine.getSheetSerialized(formulasPlugin.sheetId)).toEqual([
+      [1],
+      ['=A1+100'],
+    ]);
+
+    expect(getDataAtCell(1, 0)).toBe(101);
+  });
+
+  it('should replace an existing formula\'s text when the `loadData` method is called for data with skipped columns',
+    async() => {
+      handsontable({
+        data: [
+          [1, 'skipped', '=A1+1'],
+          [2, 'skipped', 3],
+        ],
+        // Only the physical columns 0 and 2 are visible, so `#getProcessedSourceDataArray` projects
+        // the rows down to the visible ones before feeding them to the engine - a different branch
+        // than the one the cases above take.
+        columns: [
+          { data: 0 },
+          { data: 2 },
+        ],
+        formulas: {
+          engine: HyperFormula,
+        }
+      });
+
+      const formulasPlugin = getPlugin('formulas');
+
+      await loadData([
+        [1, 'skipped', '=A1+100'],
+        [2, 'skipped', 3],
+      ]);
+
+      expect(formulasPlugin.engine.getSheetSerialized(formulasPlugin.sheetId)).toEqual([
+        [1, '=A1+100'],
+        [2, 3],
+      ]);
+
+      expect(getDataAtCell(0, 1)).toBe(101);
+    });
+
+  it('should keep an array formula spilling when the `loadData` method is called', async() => {
+    handsontable({
+      data: [
+        [1, 2],
+        ['=TRANSPOSE(A1:B1)', null],
+        [null, null],
+      ],
+      formulas: {
+        engine: HyperFormula,
+      }
+    });
+
+    // A spill cell reports the `ARRAY` cell type, not `ARRAYFORMULA`, so it passes the
+    // `VALUE`/`EMPTY` early return in `#onModifySourceData` and gets projected as well - as its
+    // calculated value. Feeding that back puts a literal inside the range the array formula needs,
+    // and the engine answers the whole load with `#SPILL!`.
+    await loadData([
+      [10, 20],
+      ['=TRANSPOSE(A1:B1)', null],
+      [null, null],
+    ]);
+
+    expect(getData()).toEqual([
+      [10, 20],
+      [10, null],
+      [20, null],
+    ]);
+  });
+
   it('should display calculated formula after changing value using `beforeChange` hook #6932', async() => {
     handsontable({
       data: [
--- a/handsontable/src/plugins/formulas/__tests__/formulas.spec.js
+++ b/handsontable/src/plugins/formulas/__tests__/formulas.spec.js
@@ -2851,9 +2851,13 @@ describe('Formulas general', () => {
   it('should replace an existing formula\'s text when the `loadData` method is called for array of objects', async() => {
     handsontable({
       data: [
-        { value: 1 },
-        { value: '=A1+1' },
+        { value: 1, note: 'skipped' },
+        { value: '=A1+1', note: 'skipped' },
       ],
+      // The second key is deliberately left out of `columns`, so `countCols()` is lower than
+      // `countSourceCols()` and the shape check in `#getProcessedSourceDataArray` is actually
+      // evaluated. With a single key it would short-circuit and the array-of-objects branch would
+      // be selected without the check ever running.
       columns: [
         { data: 'value' },
       ],
@@ -2865,8 +2869,8 @@ describe('Formulas general', () => {
     const formulasPlugin = getPlugin('formulas');
 
     await loadData([
-      { value: 1 },
-      { value: '=A1+100' },
+      { value: 1, note: 'skipped' },
+      { value: '=A1+100', note: 'skipped' },
     ]);
 
     expect(formulasPlugin.engine.getSheetSerialized(formulasPlugin.sheetId)).toEqual([
@@ -2906,6 +2910,39 @@ describe('Formulas general', () => {
       expect(formulasPlugin.engine.getSheetSerialized(formulasPlugin.sheetId)).toEqual([
         [1, '=A1+100'],
         [2, 3],
+      ]);
+
+      expect(getDataAtCell(0, 1)).toBe(101);
+    });
+
+  it('should replace an existing formula\'s text when the `loadData` method is called for data with an empty row',
+    async() => {
+      handsontable({
+        data: [
+          [1, 'skipped', '=A1+1'],
+          null,
+        ],
+        // The shape check in `#getProcessedSourceDataArray` reads the first row only, so a row that
+        // is not an array still reaches the branch projecting rows down to the visible columns.
+        columns: [
+          { data: 0 },
+          { data: 2 },
+        ],
+        formulas: {
+          engine: HyperFormula,
+        }
+      });
+
+      const formulasPlugin = getPlugin('formulas');
+
+      await loadData([
+        [1, 'skipped', '=A1+100'],
+        null,
+      ]);
+
+      // The engine drops the trailing empty row, so only the first one is serialized.
+      expect(formulasPlugin.engine.getSheetSerialized(formulasPlugin.sheetId)).toEqual([
+        [1, '=A1+100'],
       ]);
 
       expect(getDataAtCell(0, 1)).toBe(101);
--- a/handsontable/src/plugins/formulas/__tests__/formulas.spec.js
+++ b/handsontable/src/plugins/formulas/__tests__/formulas.spec.js
@@ -2706,6 +2706,62 @@ describe('Formulas general', () => {
     expect(getDataAtCell(1, 4)).toEqual(3);
   });
 
+  it('should replace an existing formula\'s text when the `loadData` method is called', async() => {
+    handsontable({
+      data: [
+        [1],
+        ['=A1+1'],
+      ],
+      formulas: {
+        engine: HyperFormula,
+      }
+    });
+
+    const formulasPlugin = getPlugin('formulas');
+
+    // The sheet is reused across `loadData` calls, so at this point the engine still holds the
+    // PREVIOUS formula. The data fed to the engine has to come from what the grid stores, not from
+    // the `modifySourceData` projection – that projection answers a formula cell with the engine's
+    // own formula, which would write the stale `=A1+1` straight back and drop the loaded one.
+    await loadData([
+      [1],
+      ['=A1+100'],
+    ]);
+
+    expect(formulasPlugin.engine.getSheetSerialized(formulasPlugin.sheetId)).toEqual([
+      [1],
+      ['=A1+100'],
+    ]);
+
+    expect(getDataAtCell(1, 0)).toBe(101);
+  });
+
+  it('should replace an existing formula\'s text when the `updateData` method is called', async() => {
+    handsontable({
+      data: [
+        [1],
+        ['=A1+1'],
+      ],
+      formulas: {
+        engine: HyperFormula,
+      }
+    });
+
+    const formulasPlugin = getPlugin('formulas');
+
+    await updateData([
+      [1],
+      ['=A1+100'],
+    ]);
+
+    expect(formulasPlugin.engine.getSheetSerialized(formulasPlugin.sheetId)).toEqual([
+      [1],
+      ['=A1+100'],
+    ]);
+
+    expect(getDataAtCell(1, 0)).toBe(101);
+  });
+
   it('should display calculated formula after changing value using `beforeChange` hook #6932', async() => {
     handsontable({
       data: [
--- a/handsontable/src/plugins/formulas/formulas.ts
+++ b/handsontable/src/plugins/formulas/formulas.ts
@@ -1334,8 +1334,10 @@ export class Formulas extends BasePlugin {
     // the guarded window - and one that calls `getDataAtCell()` on a formula cell has to keep
     // receiving the calculated value, not the raw formula.
     //
-    // The previous value is saved and restored rather than cleared, so callers that already hold
-    // the flag keep it.
+    // No other site sets the flag, so the previous value is saved and restored rather than cleared
+    // for one reason only: those same third-party handlers run inside the window, and one that
+    // reaches this method again - synchronously, through `updateSettings()` or `loadData()` - would
+    // otherwise leave the rest of the outer read unguarded, which is the very defect above.
     const wasProjectionSuspended = this.#sourceDataProjectionSuspended;
 
     this.#sourceDataProjectionSuspended = true;
--- a/handsontable/src/plugins/formulas/formulas.ts
+++ b/handsontable/src/plugins/formulas/formulas.ts
@@ -292,6 +292,10 @@ export class Formulas extends BasePlugin {
    * behalf. Read by `#onModifySourceData` alone, so the read reports what Handsontable stores
    * without also suspending the value projection every other hook depends on.
    *
+   * `#syncFormulasToSourceData` reads the source data with the same intent but deliberately keeps
+   * the broader `#internalOperationPending`: there the flag doubles as the re-entry guard its own
+   * early return checks, which this narrow flag does not provide.
+   *
    * @private
    * @type {boolean}
    */
@@ -1375,7 +1379,11 @@ export class Formulas extends BasePlugin {
     // containing only visible columns so HF cell coordinates stay in sync.
     const columnOffset = column ?? 0;
 
-    return dataArray.map((rowArray, rowIndex) => {
+    return dataArray.map((row, rowIndex) => {
+      // `getSourceDataArray` hands back a falsy row as-is for a hole or a `null` entry, and the
+      // shape check above only reads row 0, so such a row can reach this branch. Treated as empty,
+      // exactly as the pass-through branch above treats it.
+      const rowArray = Array.isArray(row) ? row : [];
       const projected = [];
 
       for (let visualCol = 0; visualCol < visibleColumnCount; visualCol++) {
--- a/handsontable/src/plugins/formulas/formulas.ts
+++ b/handsontable/src/plugins/formulas/formulas.ts
@@ -1312,7 +1312,24 @@ export class Formulas extends BasePlugin {
    * @returns {Array} The source data array to be passed to the formula engine.
    */
   #getProcessedSourceDataArray(row?: number, column?: number, row2?: number, column2?: number) {
-    const dataArray = this.hot.getSourceDataArray(row, column, row2, column2);
+    // Every caller feeds the result to the engine, so this read has to report what Handsontable
+    // actually stores – not what it reports. Left unguarded, `#onModifySourceData` answers every
+    // formula cell with the formula the engine already holds, so a `loadData()`/`updateData()` call
+    // that changes a formula's text reads the engine's PREVIOUS formula back and writes it straight
+    // into the engine again, silently discarding the newly loaded one. The previous value is saved
+    // and restored rather than cleared, so callers that already hold the flag keep it.
+    const wasInternalOperationPending = this.#internalOperationPending;
+
+    this.#internalOperationPending = true;
+
+    let dataArray;
+
+    try {
+      dataArray = this.hot.getSourceDataArray(row, column, row2, column2);
+    } finally {
+      this.#internalOperationPending = wasInternalOperationPending;
+    }
+
     const visibleColumnCount = this.hot.countCols();
     const physicalColumnCount = this.hot.countSourceCols();
     const isAoAWithSkippedColumns = visibleColumnCount < physicalColumnCount
--- a/handsontable/src/plugins/formulas/formulas.ts
+++ b/handsontable/src/plugins/formulas/formulas.ts
@@ -288,6 +288,16 @@ export class Formulas extends BasePlugin {
   #sourceDataSyncPending = false;
 
   /**
+   * Guard flag set while `#getProcessedSourceDataArray` reads the source data on the engine's
+   * behalf. Read by `#onModifySourceData` alone, so the read reports what Handsontable stores
+   * without also suspending the value projection every other hook depends on.
+   *
+   * @private
+   * @type {boolean}
+   */
+  #sourceDataProjectionSuspended = false;
+
+  /**
    * The changes that the engine reported while undoing or redoing an action. They are collected in
    * `beforeUndo`/`beforeRedo` and consumed in `afterUndo`/`afterRedo`, where the dependent cells get
    * validated.
@@ -1316,24 +1326,35 @@ export class Formulas extends BasePlugin {
     // actually stores – not what it reports. Left unguarded, `#onModifySourceData` answers every
     // formula cell with the formula the engine already holds, so a `loadData()`/`updateData()` call
     // that changes a formula's text reads the engine's PREVIOUS formula back and writes it straight
-    // into the engine again, silently discarding the newly loaded one. The previous value is saved
-    // and restored rather than cleared, so callers that already hold the flag keep it.
-    const wasInternalOperationPending = this.#internalOperationPending;
+    // into the engine again, silently discarding the newly loaded one.
+    //
+    // The projection is suspended through a dedicated flag rather than `#internalOperationPending`,
+    // which also gates `#onModifyData` and `#onAfterRenderer`. This read runs the `modifyRowData`
+    // and `modifySourceData` hooks for every row and cell, so third-party handlers execute inside
+    // the guarded window - and one that calls `getDataAtCell()` on a formula cell has to keep
+    // receiving the calculated value, not the raw formula.
+    //
+    // The previous value is saved and restored rather than cleared, so callers that already hold
+    // the flag keep it.
+    const wasProjectionSuspended = this.#sourceDataProjectionSuspended;
 
-    this.#internalOperationPending = true;
+    this.#sourceDataProjectionSuspended = true;
 
     let dataArray;
 
     try {
       dataArray = this.hot.getSourceDataArray(row, column, row2, column2);
     } finally {
-      this.#internalOperationPending = wasInternalOperationPending;
+      this.#sourceDataProjectionSuspended = wasProjectionSuspended;
     }
 
     const visibleColumnCount = this.hot.countCols();
     const physicalColumnCount = this.hot.countSourceCols();
+    // Only the shape of the data matters, and it is the same for every row, so one row answers it.
+    // `getSourceData()` would rebuild the whole dataset - through the per-cell `modifySourceData`
+    // hook, and outside the guard above - just to run `isArrayOfArrays` over it.
     const isAoAWithSkippedColumns = visibleColumnCount < physicalColumnCount
-      && isArrayOfArrays(this.hot.getSourceData());
+      && Array.isArray(this.hot.getSourceDataAtRow(0));
 
     if (!isAoAWithSkippedColumns) {
       return dataArray.map((rowObject, rowIndex) => {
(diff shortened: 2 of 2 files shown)


qunabu:
Task linked: [DEV-2585 Fix memory leak: Formulas plugin adds a new HyperFormula sheet on every loadData/updateData](https://app.clickup.com/t/9015210959/DEV-2585)

handsontable/src/plugins/formulas/formulas.ts
```diff
@@ -1312,7 +1312,24 @@ export class Formulas extends BasePlugin {
    * @returns {Array} The source data array to be passed to the formula engine.
    */
   #getProcessedSourceDataArray(row?: number, column?: number, row2?: number, column2?: number) {
-    const dataArray = this.hot.getSourceDataArray(row, column, row2, column2);
+    // Every caller feeds the result to the engine, so this read has to report what Handsontable
```
budnix, line 1329: Can I ask — did you check the paths where `#syncFormulasToSourceData` skips the write-back on purpose? When the engine holds a `#REF!` formula and `allowBrokenReferences` is false, line 2192 leaves the good text in the source data and lets the engine keep the broken one. Before this change the read-time projection covered that gap, so `afterCellMetaReset` fed the engine back its own formula on the next `updateSettings`. Now the stored text goes in instead, which turns a formula the engine had invalidated back into a live one — same gap after a sort or a row move, where the write-back also bails.

handsontable/src/plugins/formulas/formulas.ts
```diff
@@ -1312,7 +1312,24 @@ export class Formulas extends BasePlugin {
...
+    // actually stores – not what it reports. Left unguarded, `#onModifySourceData` answers every
+    // formula cell with the formula the engine already holds, so a `loadData()`/`updateData()` call
+    // that changes a formula's text reads the engine's PREVIOUS formula back and writes it straight
+    // into the engine again, silently discarding the newly loaded one. The previous value is saved
+    // and restored rather than cleared, so callers that already hold the flag keep it.
```
budnix: The save-and-restore reads well, but I don't think anything can see it today. `#onAfterDetachChild` is the only caller that holds the flag, and it sets it back to `false` right after the call anyway — so its own `= true` on line 2663 now wraps nothing. Worth picking one: either drop the outer pair there and keep save/restore as future-proofing, or keep the wrap and just clear the flag here. Right now no test would notice if the `finally` set `false` instead.

handsontable/src/plugins/formulas/formulas.ts
```diff
@@ -1312,7 +1312,24 @@ export class Formulas extends BasePlugin {
...
+    const wasInternalOperationPending = this.#internalOperationPending;
+
+    this.#internalOperationPending = true;
```
budnix: `#internalOperationPending` also gates `#onModifyData` (line 1662) and `#onAfterRenderer` (line 1715). The other sites hold it around short engine syncs, but this one now wraps every full source read, including the load path — and `getSourceDataArray` runs the `modifySourceData` and `modifyRowData` hooks for every cell, where user handlers live. A handler that calls `getDataAtCell` on a formula cell would get the raw `=A1+1` back instead of the value. A dedicated flag checked only by `#onModifySourceData` would scope this to the hook the fix is about — `#sourceDataSyncPending` is already that shape.

handsontable/src/plugins/formulas/formulas.ts
```diff
@@ -1312,7 +1312,24 @@ export class Formulas extends BasePlugin {
...
+
+    try {
+      dataArray = this.hot.getSourceDataArray(row, column, row2, column2);
+    } finally {
+      this.#internalOperationPending = wasInternalOperationPending;
+    }
+
     const visibleColumnCount = this.hot.countCols();
     const physicalColumnCount = this.hot.countSourceCols();
     const isAoAWithSkippedColumns = visibleColumnCount < physicalColumnCount
```
budnix, line 1362: Small one, and it predates this PR: `isArrayOfArrays(this.hot.getSourceData())` on the next line rebuilds the whole dataset through the per-cell `modifySourceData` hook just to check the shape of one row. It also sits outside the new guard, so it stays fully projected. `#doesEngineHoldPhysicalColumns` (line 2038) already solves this with `Array.isArray(this.hot.getSourceDataAtRow(0))` and a comment saying why. Feels close enough to this change to fold in, but happy either way.

handsontable/src/plugins/formulas/__tests__/formulas.spec.js
```diff
@@ -2706,6 +2706,117 @@ describe('Formulas general', () => {
     expect(getDataAtCell(1, 4)).toEqual(3);
   });
 
+  it('should replace an existing formula\'s text when the `loadData` method is called', async() => {
```
budnix, line 2709: All four cases use one-column array-of-arrays data, so only the `!isAoAWithSkippedColumns` branch runs. The other branch reads the row differently — it walks visible columns and does its own offset math — and a break there would ship green. One case with array-of-objects data and one with `columns: [{ data: 0 }, { data: 2 }]` over three-column data would cover it, and both are shapes where `loadData` replacing a formula is a normal thing to do.

handsontable/src/plugins/formulas/__tests__/formulas.spec.js
```diff
@@ -2706,6 +2706,117 @@ describe('Formulas general', () => {
...
+      [1],
+      [5],
+    ]);
+
+    expect(getDataAtCell(1, 0)).toBe(5);
+  });
+
   it('should display calculated formula after changing value using `beforeChange` hook #6932', async() => {
     handsontable({
       data: [
```
demtario, line 2982: **medium** - these four cases pin the identity-frame, scalar-formula path. Four reachable cases where this guard decides the outcome are untested, and two of them fail far louder pre-fix than the stale value asserted here.

**A. Moved frame + `updateData`** (site `formulas.ts:1617`). `fitToLength` at equal length inserts nothing (`indexMapper.ts:541-543`), so a permuted sequence survives `updateData`. Pre-fix the projection returns `=A2+1` into a slot where `A2` is that row itself: `#CYCLE!`, not just a stale value.

**B. Moved frame + data-less `updateSettings`** (site `formulas.ts:1555`), with no `columns` key. That key routes through `initIndexMappers()` at `core.ts:3539` and resets the sequence, so it has to be absent for the case to be reachable.

**C. Array formula + either verb.** Frame-independent, hits both sites. Spill cells report `CellType.ARRAY`, not `ARRAYFORMULA` (`Cell.js:47-53`), so they slip the `VALUE`/`EMPTY` early return, and `getCellSerialized` falls back to `getRawValue` (`Serialization.js:52`). Pre-fix that feeds computed literals into every spill position and HF collapses the array to `#SPILL!` in the anchor, synchronously inside `setSheetContent`. Note `isItPossibleToReplaceSheetContent` checks size limits only (`CrudOperations.js:463-471`), so the `warn()` branch is not what fires. No spec in the plugin asserts a spill cell today.

**D. `updateSettings({ data })`.** A third public entry point, and worth calling out in the description: it never reaches 1617 at all, because `#onAfterLoadData` early-returns on `source === 'updateSettings'` (`formulas.ts:1611-1613`) while `core.ts:3533` passes exactly that source. So it is handled at 1555, and this PR fixes it without claiming or testing it.

D is the cheapest to add and covers a genuine public API path. A and C are the ones that would have caught a wrong fix.

demtario (commented):
Two notes from review. Neither blocks; the production change looks correct, and tracing it turned up that it fixes more than the description claims.

### The change falsifies the `#syncFormulasToSourceData` doc comment (`formulas.ts:2118-2121`)

That comment was added one commit earlier, in `629947fae`, and reads:

> Row and column *moves* (and sorting) are deliberately excluded: they reorder the engine's
> indexes without touching the source data, so the two stop sharing a reference frame. While that
> is the case nothing is written back at all - the read-time projection keeps handling it, exactly
> as it did before.

The projection no longer handles it at either engine-feeding site (`formulas.ts:1555` and `1617`), because the new guard suppresses it there. So the documented mitigation no longer holds and the comment now describes behavior that does not exist.

This is off-diff, hence the review body rather than an inline comment.

**Two separable asks**, and only the first belongs to this PR:

1. Update `formulas.ts:2118-2121` so it says the projection still covers ordinary reads but no longer covers the engine-feeding reads in `#getProcessedSourceDataArray`.
2. Separately, a design question for the team, not a blocker here: whether the `isHfOrderPhysical()` bail at `formulas.ts:2136-2138` needs narrowing so a structural change after a sort or move still syncs.

Why the second question exists, for context. After a sort or move followed by an insert, the bail leaves the stored formula text stale in its own frame, and the projection was the only thing reading the engine's current text back. Worked example on `[[1],[2],['=A1+A2']]` with `manualRowMove: true`, `moveRow(2, 0)`, insert above visual row 1, then `updateSettings({ colHeaders: true })` (the `colHeaders` payload matters: any `columns` key routes through `initIndexMappers()` at `core.ts:3539` and resets the sequence):

- correct value: **3**
- with this PR: **1**
- before this PR: **`#CYCLE!`**

So this is not a regression, and we are not asking you to fix it here. Neither branch was correct. The two failures have independent causes: positions land correctly either way, since `transformation[p] == hfIndexOfPhysical[p]`; the branches differ only in how references are mapped. The old path fed text already expressed in the engine's frame and `setRowOrder` mapped it a second time. The new path feeds physical-frame text mapped exactly once, so the transform is right and only the input is stale. This PR fixes the frame defect it targets and leaves the staleness defect, now unmasked rather than masked. The comment is the part that is straightforwardly wrong.

Analysis is from reading the source, not from executing that scenario. The snippet above is the repro if you want to confirm the 3 to 1 change.

### Not findings, recorded so they are not re-litigated

- Adding `it()` blocks to an existing `*.spec.js` is permitted. The Jasmine freeze is file-scoped (`handsontable/.ai/TESTING.md:300-301`), and line 297 prescribes exactly this for a bug fix.
- `[skip changelog]` is correct here.
- The save/restore shape is right, not over-engineered: a bare `finally { ... = false }` would clear the flag early for `#onAfterDetachChild`, which already holds it.
- The guard's blast radius is contained. `dataSource` emits only `modifyRowData`, `modifySourceData`, and `modifySourceLength`, and of the four `#internalOperationPending` readers only `#onModifySourceData` is reachable from `getSourceDataArray()`. The path is fully synchronous.
- Real blast radius is narrower than the description states: sites 533 and 762 are gated on `!doesSheetExist` and `sheetName === null`, which `#onModifySourceData` already bails on, and 2667 already holds the flag. Behavior changes at 1555 and 1617 only.

handsontable/src/plugins/formulas/formulas.ts
adrianspdev: Checked it, and it does not reproduce — the divergence you describe is reachable only through the moved/sorted-frame bail, and there the re-fed text does not come back live either.

`REF_ERROR_PATTERN` tests the engine's formula **text**, not its value, so only a structural operation can put a `#REF!` there. The three that can are `remove_row`, `remove_col` and `move_cells` (`REFERENCE_BREAKING_ACTION_TYPES`), and all three sync with `allowBrokenReferences: true`. The `false` paths are `afterCreateRow`/`afterCreateCol` and the non-breaking undo/redo actions, and an insert never authors a `#REF!` of its own. So line 2192 can only skip a `#REF!` that a removal already wrote into the source data.

Measured both frames on `[[1], [2], ['=A1+1']]`, removing the referenced row, then `updateSettings({ colHeaders: true })`:

| | engine | stored (raw `data` array) | grid |
|---|---|---|---|
| plain frame | `=#REF!+1` | `=#REF!+1` | `#REF!` |
| moved frame | `=#REF!+1` | `=A1+1` | `#REF!` |

The plain frame has nothing to diverge — `sync(true)` wrote the broken text through. The moved frame does diverge, exactly as you predicted, but feeding `=A1+1` back runs it through the same double reference transform that the frame defect causes, so it lands as `=#REF!+1` again. No revival, and the grid reads `#REF!` before and after.

So this is the `isHfOrderPhysical()` bail @demtario raised, not a second gap. It is real, and I have written a repro into the PR description so it does not get lost, but it is pre-existing and out of scope here.

handsontable/src/plugins/formulas/formulas.ts
adrianspdev: You were right, and it was worse than unobservable — moving to a dedicated flag (your next comment) made the justification I had written actually false. There is now no other site that sets it, so "callers that already hold the flag keep it" described a caller that does not exist.

Picked: kept the save and restore, rewrote the reason. It still earns its place, but for the reason from your `:1323` comment rather than for `#onAfterDetachChild` — the read runs `modifyRowData` and `modifySourceData`, so third-party handlers execute inside the window, and one that re-enters this method synchronously (through `updateSettings()` or `loadData()`) would otherwise clear the flag and leave the rest of the outer read unguarded.

`#onAfterDetachChild` is untouched and back to doing its original job, since the helper no longer touches `#internalOperationPending` at all.

Fixed in ea0a9f6aaa.

handsontable/src/plugins/formulas/formulas.ts
adrianspdev: Agreed, and done — this was the substantive one. `#sourceDataProjectionSuspended`, read by `#onModifySourceData` alone.

Confirmed your premise directly: `dataSource.ts` fires `modifyRowData` at `:73` and `modifySourceData` at `:310`, so user handlers do run inside the guarded window, and with `#internalOperationPending` held a handler calling `getDataAtCell()` on a formula cell would have taken the `#onModifyData` early return and received the raw formula instead of the calculated value. `#onAfterRenderer` was in the same position.

I did not reuse `#sourceDataSyncPending` despite the matching shape — it carries a distinct meaning (the write-back fence) and is also read at `:1888`, so overloading it would have coupled two unrelated states.

Fixed in e0fa0e8170.

handsontable/src/plugins/formulas/formulas.ts
adrianspdev: Folded in — it is close enough, and the "outside the guard" part makes it more than cosmetic: that call was the one full-dataset read still being projected on every load.

Now `Array.isArray(this.hot.getSourceDataAtRow(0))`, matching `#doesEngineHoldPhysicalColumns` and carrying the same comment. The semantics differ only for a mixed dataset whose first row is an array while a later one is not, which is not a shape Handsontable supports — and `#doesEngineHoldPhysicalColumns` already accepted that trade.

Fixed in e0fa0e8170.

handsontable/src/plugins/formulas/__tests__/formulas.spec.js
adrianspdev: Both added, and you were right that the other branch was reachable and uncovered:

- `should replace an existing formula's text when the loadData method is called for array of objects`
- `should replace an existing formula's text when the loadData method is called for data with skipped columns` — `columns: [{ data: 0 }, { data: 2 }]` over three-column data, which is the only way to reach the `isAoAWithSkippedColumns` branch

Both were verified failing without the guard (stale `=A1+1`, value 2 instead of 101) and passing with it. The skipped-columns case also covers the `getSourceDataAtRow(0)` shape check from your other comment, since that is the condition selecting the branch.

Added in e0fa0e8170.

handsontable/src/plugins/formulas/__tests__/formulas.spec.js
adrianspdev: Worked all four. **C and D are added**; **A and B are not**, and B is not what it looks like — measurements below, because two of your predicted values did not hold.

### D — added

`should replace an existing formula's text when the data is replaced through updateSettings`. Confirmed the routing you described: `core.ts:3532-3533` passes the `updateSettings` source, `#onAfterLoadData` early-returns on it at `:1610-1613`, so it lands at `#onAfterCellMetaReset`. Fails without the guard (stale `=A1+1`, value 2), passes with it. Also now called out in the PR description as a third public entry point.

### C — added, and it behaved exactly as you said

`should keep an array formula spilling when the loadData method is called`, on `[[1, 2], ['=TRANSPOSE(A1:B1)', null], [null, null]]` reloaded with `[[10, 20], ...]`.

Without the guard the grid comes back `[[10, 20], ['#SPILL!', null], [null, null]]`. With it, `[[10, 20], [10, null], [20, null]]`. The spill cell reporting `ARRAY` rather than `ARRAYFORMULA` is what slips the early return, as you traced. This is the case that would have caught a wrong fix — thank you for it.

### A — measured, but it cannot be pinned

`[[1], [2], ['=A1+1']]`, `moveRow(2, 0)`, then `updateData([[1], [2], ['=A1+100']])`:

| | engine | grid |
|---|---|---|
| pre-fix | `=#REF!+1` | `#REF!` |
| post-fix | `=#REF!+100` | `#REF!` |

Not `#CYCLE!` → `1`. The guard changes only which **text** reaches the engine; the reference is mapped twice either way, so both branches land on `#REF!`. The correct value is 101 (`=A1+100` in the physical frame), and neither branch produces it.

So there is no green assertion available here until the frame defect is fixed, and `xit` is not a good answer either — this repo's own `handsontable/no-skipped-test` rule flags it as proving nothing. I left it out and documented the gap in the `#syncFormulasToSourceData` comment instead.

### B — the guard does not decide this one

`[[1], [2], ['=A1+1']]`, `moveRow(2, 0)`, then `updateSettings({ colHeaders: true })` — no `columns` key, so the sequence survives, as you specified:

| | engine | grid |
|---|---|---|
| before `updateSettings` | `=A2+1` | `[[2], [1], [2]]` ✓ |
| after, **guard on** | `=#REF!+1` | `[['#REF!'], [1], [2]]` |
| after, **guard off** | `=#REF!+1` | `[['#REF!'], [1], [2]]` |

Byte-identical. Same reason as A: in a moved frame the projected text and the stored text both come out `#REF!`. Control case — identity frame, data-less `updateSettings` — is unaffected in both branches.

What this did surface is that **B is a live defect on `develop` in its own right**, and a louder one than the staleness you described for the sort-then-insert path: a data-less `updateSettings({ colHeaders: true })` on a moved frame turns a working formula into `#REF!`, with nothing in the call that mentions data. I have put that repro in the PR description under the known-limitation section so it is attached to the `isHfOrderPhysical()` question rather than lost in a thread.

adrianspdev:
Thanks both — pushed e0fa0e8170 and ea0a9f6aaa, and rewrote the PR description around what the review turned up. Summary of the round, with the two places where a prediction did not survive measurement.

**@budnix — all five addressed.**

- The flag was too broad, as you said. There is now a dedicated `#sourceDataProjectionSuspended` read only by `#onModifySourceData`. Your premise checks out: `dataSource.ts` fires `modifyRowData` at `:73` and `modifySourceData` at `:310`, so third-party handlers really do run inside the window, and one calling `getDataAtCell()` on a formula cell would have received the raw formula.
- The `#onAfterDetachChild` redundancy dissolved rather than needing a pick — the helper no longer touches `#internalOperationPending`, so that wrap is back to its original job and the file is untouched. The save/restore stayed, but its stated reason was wrong once the flag became dedicated, so that comment is rewritten.
- The `isArrayOfArrays(getSourceData())` read is folded in.
- Your `#REF!` case does not reproduce; details in that thread.

**@demtario — the doc comment is corrected**, and it needed to be: the mitigation it described genuinely stopped existing at both engine-feeding sites. It now says that ordinary reads are still projected, the engine-feeding reads are not, and that a moved or sorted frame therefore hands the engine stored, possibly stale text.

Your blast-radius correction was right and is now in the description: I verified that the two `addSheet` sites are gated on `!doesSheetExist` and `sheetName === null`, both of which `#onModifySourceData` already bails on, and that `#onAfterDetachChild` already holds the flag. Behavior changes at `#onAfterCellMetaReset` and `#onAfterLoadData` only.

**Four specs added** — `updateSettings({ data })`, array-of-objects, `columns` skipping a physical index, and the array-formula spill. Discrimination re-verified end to end: with the guard removed and rebuilt, **exactly eight specs fail** — the original four plus these four — and all pass once it is restored. Full suite 9726 specs / 0 failures, unit 3802 passing, eslint 0 errors with the warning count unchanged at 602.

**Two cases I did not add, both moved-frame:** A and B. In both, the guard is not what determines the outcome — the result is `#REF!` with and without it, so neither can be asserted green. Measurements are in the test thread.

**The one thing this leaves open** is your second ask, the `isHfOrderPhysical()` narrowing. Agreed it is not this PR. It needs a ticket, and while checking case B I found a sharper repro for it than the sort-then-insert path — a data-less `updateSettings({ colHeaders: true })` on a moved frame turns a working formula into `#REF!` on `develop` today. That repro is in the PR description's known-limitation section so it travels with the ticket when we file it.

demtario (commented):
Follow-up review of the update (`210f53655d..ea0a9f6aaa`). The flag split is the right call and I verified the reasoning behind it holds — `modifyData` never fires on the `getSourceDataArray` path and the plugin registers no `modifyRowData` handler, so `#onModifySourceData` was the only gate ever load-bearing inside that window, and `dataMap.get` reads the source directly before running `modifyData`, so a third-party handler calling `getDataAtCell()` inside the window does still get the calculated value. The `#syncFormulasToSourceData` doc rewrite reads accurately against the code.

One new defect and two notes. The defect is inline on the predicate that causes it; the note below has no changed line to anchor to.

### Note: the sibling read still uses the broad flag (`formulas.ts:2186`)

The delta's argument is that `#internalOperationPending` is too broad for a "report what is stored" read. The sibling read with identical intent still uses it: `#syncFormulasToSourceData` sets the broad flag around its comparison loop, which also fires `modifySourceData` per cell, so third-party handlers there still run with `#onModifyData` and `#onAfterRenderer` suppressed for no reason — the exact over-suppression this delta just removed at the other site.

**Do not simply swap the flag there.** At that site `#internalOperationPending` also serves as the re-entry guard via the early return at `formulas.ts:2159`; replacing it would drop that. Consistency would mean setting both flags around the loop. Deferring this to a follow-up is a perfectly reasonable call — the useful minimum for this PR would be one clause on the new field's doc block saying why the sibling site keeps the broad flag, so the next reader does not read the inconsistency as an oversight.

### Note: nothing discriminates the flag swap itself

The delta's actual behavior change — keeping `#onModifyData` and `#onAfterRenderer` live during the guarded read — is only observable through a third-party hook handler, so a test for it would need a re-entrancy boolean in the handler (`dataMap.get` re-fires `modifyRowData`). Raising it as a note, not a request; it may not be worth the test.

### Verified clean

Save/restore cannot leak on a throw (`try`/`finally`), so no `disablePlugin`/`destroy` reset is needed. All six conditions in `#onModifySourceData` are side-effect-free reads, so the new condition's ordering is irrelevant and shadows nothing. All four helper call sites are unaffected by dropping the broad flag: 543 runs before the hooks are registered, 772 runs under `sheetName === null` which both projectors early-return on, 1578 performs no data read between the call and its own window, and 2699 sets the broad flag itself. The four new specs each fail with the guard reverted and hit distinct sites; `getData()` is the right assertion for the spill spec.

### Unrelated: CI Gate is red on a flake, not on this PR

`E2E / Playwright UMD (theme: main)` failed on `tests/e2e/walkontable/frozen-column-row-heights.spec.ts:206` (`Expected: 69, Received: 30`). This PR touches only `plugins/formulas`. The other five Playwright matrix cells passed, including `UMD.min (theme: main)` on the same code, and `develop` is green on this PR's base. Re-running that one job should clear the gate.

handsontable/src/plugins/formulas/formulas.ts
```diff
@@ -1312,11 +1322,41 @@ export class Formulas extends BasePlugin {
...
+    }
+
     const visibleColumnCount = this.hot.countCols();
     const physicalColumnCount = this.hot.countSourceCols();
+    // Only the shape of the data matters, and it is the same for every row, so one row answers it.
+    // `getSourceData()` would rebuild the whole dataset - through the per-cell `modifySourceData`
+    // hook, and outside the guard above - just to run `isArrayOfArrays` over it.
     const isAoAWithSkippedColumns = visibleColumnCount < physicalColumnCount
-      && isArrayOfArrays(this.hot.getSourceData());
+      && Array.isArray(this.hot.getSourceDataAtRow(0));
```
demtario, line 1363: **medium** - this widening makes the projection branch reachable for a `null`/hole row, and that branch is not total over it: `rowArray.length` at line 1390 throws a `TypeError`, inside `#onAfterLoadData` / `#onAfterCellMetaReset`.

The old predicate `isArrayOfArrays(data)` requires **every** row to be an array (`helpers/data.ts:198` - `data.every(el => Array.isArray(el))`). This one checks row 0 only, so the two disagree precisely when row 0 is an array and some later row is not.

Repro - `data: [[1, 2, 3], null]`, equally the sparse literal `[[1, 2, 3], , [4, 5, 6]]`, with `columns: [{ data: 0 }, { data: 2 }]` and formulas enabled:

1. `countCols() (2) < countSourceCols() (3)` and row 0 is an array, so `isAoAWithSkippedColumns` is now `true`. It was `false` before, because `.every()` rejected the `null` row.
2. `getSourceDataArray()` really does yield `null` for that row: `getAtRow` initializes `newDataRow = null` and takes none of the `Array.isArray` / `isObject` / `isFunction` branches for a falsy row (`dataMap/dataSource.ts:147-215`), and `getByRange` pushes it through as-is (`:360-372`).
3. Line 1390 evaluates `arrayIndex >= rowArray.length` on `null` and throws. The pass-through branch returned `[]` for the same row.

Worth noting this is reachable from ordinary data rather than only a malformed fixture: core tolerates a falsy row rather than rejecting it, guarding for it explicitly at `dataMap/dataMap.ts:876` (`if (dataRow && hasOwnProperty(...))`) and `:884` (`if (!out) return null`). An *object* row in that position does not throw - `arrayIndex >= undefined` is `false`, so it pushes `undefined`s - so `null`/hole is the case that matters.

**Fix** is the normalization the pass-through branch already has at line 1363, applied at the top of the callback on line 1378:

```js
return dataArray.map((row, rowIndex) => {
  const rowArray = Array.isArray(row) ? row : [];
  // ...
```

With `[]` the inner loop skips every column and the row is fed as empty, matching what the pass-through branch does for that shape.

handsontable/src/plugins/formulas/__tests__/formulas.spec.js
```diff
@@ -2706,6 +2706,240 @@ describe('Formulas general', () => {
...
+
+    expect(formulasPlugin.engine.getSheetSerialized(formulasPlugin.sheetId)).toEqual([
+      [1],
+      ['=A1+100'],
+    ]);
+
+    expect(getDataAtCell(1, 0)).toBe(101);
+  });
+
+  it('should replace an existing formula\'s text when the `loadData` method is called for array of objects', async() => {
```
demtario, line 2851: **low** - none of the four new specs discriminates the changed shape check, so the predicate swap on line 1359 is unpinned.

This one never evaluates it: with a single data key and a single `columns` entry, `countCols() (1) < countSourceCols() (1)` is `false`, so `&&` short-circuits before `Array.isArray(getSourceDataAtRow(0))` runs. The skipped-columns spec below does reach it and would fail if it stopped returning an array, so the `true` branch is covered - but the object branch reached *with* `visible < physical` is not.

Cheapest fix: give this spec a second data key that `columns` does not list, which makes it actually reach and pin the object branch. Low value on its own, and it would become considerably more valuable alongside a case for the `null`/hole row in the other comment, since both live in the same predicate.

budnix (dismissed):

handsontable/src/plugins/formulas/formulas.ts
adrianspdev: Confirmed and fixed — the repro is exact, and I reproduced the throw as a test before touching the code.

Verified your reachability chain independently: `countFirstRowKeys` (`helpers/data.ts:176-188`) reads row 0 only, so `[[1, 'skipped', '=A1+1'], null]` gives `countSourceCols() === 3`, and `columns: [{ data: 0 }, { data: 2 }]` gives `countCols() === 2`. The new predicate also reads row 0 only, so it flips `false` -> `true` exactly where `isArrayOfArrays`'s `.every()` used to reject the row, and `arrayIndex >= rowArray.length` throws on `null`.

Applied your fix — the normalization from line 1363, at the top of the projection callback:

```js
return dataArray.map((row, rowIndex) => {
  const rowArray = Array.isArray(row) ? row : [];
```

New spec: `should replace an existing formula's text when the loadData method is called for data with an empty row`. Reverted the one line and it fails with `TypeError: Cannot read properties of null (reading 'length')`, alone, out of 372 formulas specs; passes with it.

One correction to the predicted end state: the engine drops the trailing empty row, so `getSheetSerialized` returns `[[1, '=A1+100']]` rather than a second `[]` row. The spec asserts that shape and says why.

Fixed in 77d4bfbff2.

handsontable/src/plugins/formulas/__tests__/formulas.spec.js
adrianspdev: Right on both halves — the spec did short-circuit, and the object branch reached with `visible < physical` was unpinned.

Took the cheap fix you suggested: the fixture now carries a second `note` key that `columns` does not list, so `countCols() (1) < countSourceCols() (2)` holds and `Array.isArray(getSourceDataAtRow(0))` is actually evaluated, returning `false` for the object row.

Verified it now discriminates. Dropping the `Array.isArray(...)` term from the predicate turns this spec red (`colToProp` returns `'value'`, the `typeof !== 'number'` guard skips every column, and the rows go to the engine empty), along with two pre-existing array-of-objects specs. Before the change the same mutation left it green.

And you were right that it pairs with the other comment — both cases now live on the same predicate, from opposite sides: this one pins the object row, the empty-row spec pins a non-array row inside the `true` branch.

Fixed in 77d4bfbff2.

demtario (approved):

budnix (approved):

PR #13244, opened 2026-08-25
DEV-2617: Make Trusted Types support real by removing Handsontable's own HTML sinks

demtario:
### Context

The PR fixes Trusted Types support that was documentation only, and did not work.

A grid did not construct at all under a Content Security Policy containing `require-trusted-types-for 'script'`. Handsontable built parts of its own interface as HTML strings, and assigning one to `innerHTML` is exactly what that policy blocks. The recipe published in three places (the `sanitizer` JSDoc, the security guide, and the 17.0 migration guide) could not help, because those writes never pass through the user's `sanitizer` — the grid died on the nested-header ghost table before any of them ran.

Two findings shaped the fix, both browser-verified in Chrome 148:

1. **Trusted Types is not a sanitizer.** It enforces that values reaching a sink came from a policy the application wrote. `createPolicy('x', { createHTML: s => s })` satisfies the browser and protects nothing. It cannot replace DOMPurify; a sanitizer is what you call *inside* a policy. The docs presented it as a sanitization strategy, which is how the claim got published.
2. **There are exactly two ways to satisfy a sink.** Bless a string with a policy, or never produce the string. Every HTML and XML parse entry point is a sink — `DOMParser` for `text/html`, `application/xhtml+xml` *and* `text/xml`, `Document.parseHTMLUnsafe`, `Element.setHTMLUnsafe`, `Range.createContextualFragment`, `document.write` — so there is no clever escape hatch. `createElement` / `textContent` / `setAttribute` are untouched.

This takes the second door. **Handsontable ships no Trusted Types policy and asks nothing of the user's `trusted-types` directive.** It builds its own markup as DOM nodes, so there is nothing for the browser to reject.

That is better than an internal policy on three counts. It imposes no CSP requirement on customers; it has no failure mode to design or degrade through (a policy throws when its name is absent from the directive, which would have broken the one configuration that works today — a permissive `default` policy); and it behaves identically on Safari 14–25 and Firefox 110–147, where Trusted Types does not exist at all. It also avoids a name collision: our own docs told every reader to call their policy `handsontable`, and `createPolicy` throws on a duplicate name.

Trusted Types reached Baseline only in February 2026, above the floor in `browser-targets.js` on two engines, so it can never be a default protection mechanism. The HTML Sanitizer API is a *different* specification and is deliberately out of scope: it is not Baseline either, and its default configuration strips `data-*`, which our own templates depend on.

[...]

PR #13264, opened 2026-08-27
DEV-2642: Keep plain prose out of the header HTML path

demtario:
> [!NOTE]
> **This PR now targets `develop` directly.** The Trusted Types work it sits on top of (#13244) is merged, so the earlier merge-ordering constraint is gone. `develop` is merged into this branch as of `5af1126`. Two files conflicted, both resolving to this branch's side: the gate's unit tests, and the `trusted-types` expectation this PR flips.

### Context

The PR fixes plain text being routed through the HTML write path, which could stop a grid from rendering entirely under a Trusted Types policy.

`HTML_CHARACTERS` decides whether `fastInnerHTML` writes through `innerHTML` or `textContent`. Both of its alternatives matched text containing no markup at all:

| content | matched before | why |
| --- | --- | --- |
| `Smith & Sons, Ltd.; est. 1920` | yes | any `&` with a later `;` |
| `R&D; notes` | yes | same |
| `Score < 50 > threshold` | yes | any `<` with a later `>` |

So an ordinary `colHeaders` string took the `innerHTML` path. Under `require-trusted-types-for 'script'` with no `default` policy that write throws a `TypeError`, `fastInnerHTML` has no `catch`, and `renderCell`'s `try`/`finally` only resets a flag — the error propagates out of `updateCellHeader` and the **whole grid fails to construct**, from a header containing no markup.

The pattern now has four alternatives, each requiring a shape that could actually be parsed as markup:

```js
/(?:<\/?[a-z][^<>]*>)|(?:<[!?])|(?:&[a-z][a-z\d]+;)|(?:&#x?[\da-f]+;)/i
```

A tag, a markup declaration or processing instruction, a named character reference, and a numeric one.

Not a regression: the pattern predates #13244. Raised in review there and split out because narrowing it is its own change with its own test surface.

### What this does not fix

Stated up front, because two of these came out of review:

- **The sink stays.** A header holding real markup (`<b>ID</b>`) still goes through `innerHTML` and still fails under enforcement without a policy-backed sanitizer. That is content the user asked to be rendered as HTML.
- **Prose shaped like a tag still takes the HTML path**, and so still throws under an enforced policy: `<none>`, `<n/a>`, `<blank>`, `Type <Enter> to continue`. These have no markup intended, which makes them the same accidental-crash class this ticket exists to reduce. Excluding them needs a known-tag-name allowlist plus a `-` rule to keep custom elements working, which is a larger and more opinionated decision than a regex narrowing — so the ticket is scoped down rather than the case closed. Pinned by a test so the limit is visible.
- **The accidental-crash class is reduced, not eliminated**, for that reason.

### Deliberate behavior changes

1. **`Handsontable.dom.HTML_CHARACTERS` is public** (assembled in `src/index.ts` from every non-underscore key of `helpers/dom/element`, and reachable through the deep import `handsontable/helpers/dom/element`). It previously carried three capture groups; it now has none. Nothing in the grid ever read them, and renumbering them as the alternatives changed would have handed consumers a silently different shape — `match[1]` becoming `undefined`. `match[0]` is unchanged and is what the test pins.
2. **A character reference without its closing semicolon now renders literally.** `&copy 2024` used to decode to `© 2024`, but only when an unrelated `;` appeared later in the same string, which made the behavior depend on the rest of the value. This affects any surface that renders markup deliberately, including `type: 'html'` cells and `allowHtml` sources.
3. **A configured `sanitizer` no longer sees content that holds no markup** on the `fastInnerHTML` path. This matters if you use the sanitizer as an audit hook or a length cap rather than to strip markup. Documented in the security guide in this PR. The `sanitizeHTML` path is unaffected: it calls a configured sanitizer *above* the gate, so the clipboard sources still receive every payload.
4. **`<a x="<">`** — a `<` inside an attribute value, in a string containing no other tag — now takes the text path. It is the single input on which `[^<>]*` and the previous `[^>]*` disagree, and the text path is the inert side.

Markup declarations are deliberately *not* reclassified: the fourth alternative keeps comments, doctypes, CDATA sections, and processing instructions on the HTML path, so `type: 'html'` cells, `allowHtml` sources, menu item names, dialog and notification content, and select-editor options render them exactly as before.

### Performance

`[^>]*` made the tag branch quadratic, and this test runs on every rendered header — and, through `sanitizeHTML`, over a whole `text/html` clipboard payload on paste. On `'<a'.repeat(40000)`:

| pattern | time |
| --- | --- |
| `[^>]*` (before) | 1304 ms, on the main thread |
| `[^<>]*` (now) | 0.25 ms |

A unit test asserts the whole pattern answers that input in under 100 ms, so a future widening of that character class is a red test rather than a slow paste.

### How has this been tested?

Every case proven red before the fix and green after.

- `handsontable/src/helpers/dom/__tests__/element.unit.ts` — the gate itself: prose, tags, declarations, all reference forms, the unanchored case `Smith & Sons; &amp; more` that a naive narrowing would break, the `<a x="<">` divergence and the neighbouring case that shows how narrow it is, the absent `g` flag (which would make `.test()` stateful through `lastIndex`), the deliberate `&#abc;` over-match, and the linear-time bound above.
- `handsontable/src/utils/__tests__/sanitizer.unit.ts` — prose no longer triggers the missing-sanitizer warning; markup and real references still do; a configured sanitizer still receives every payload on that path.
- `handsontable/src/plugins/nestedHeaders/utils/__tests__/ghostTable.unit.js` — the measured label matches the rendered one: prose never reaches the sanitizer, markup and references do, and an unterminated tag builds no element.
- `tests/e2e/header-markup-gate.spec.ts` plus its fixture and page object — the real-browser half. The nested-markup leg counts sanitizer calls rather than checking presence, because the rendered header and the ghost table pass an identical `(content, source)` pair and presence alone cannot tell one surface from both.
- `tests/e2e/trusted-types.spec.ts` — the prose case is flipped from `CONSTRUCT-THREW` to `CONSTRUCTED`. **This is the regression test for the headline symptom**, under an enforced policy, and it is only possible because that Trusted Types coverage is now on `develop`. The `markup` case beside it is untouched and still throws, which keeps the surviving boundary pinned.

The full sweep, before `develop` was merged in:

```bash
npm --prefix handsontable run test:unit          # 3885 passed, 323 suites
npm --prefix handsontable run test:e2e           # 9721 specs, 0 failures, 8 pending
HOT_TEST_PORT=8137 npm --prefix tests run test:e2e   # 446 passed, 1 skipped
npm --prefix handsontable run eslint             # 0 errors
npm --prefix tests run lint
```

Re-run after the merge and the review fixes below:

```bash
npm --prefix handsontable run test:unit          # 4311 passed, 340 suites
cd tests && npx playwright test e2e/trusted-types.spec.ts \
  e2e/header-markup-gate.spec.ts --project=e2e-main   # 23 passed, cold server
```

The Jasmine suite, the full Playwright matrix, and stylelint were not re-run locally after the merge; CI covers them.

### Since the first review round

Review found four places that still described the pre-narrowing gate, one of them naming this ticket as the open issue that would fix it. All four now describe the shapes the pattern actually requires:

- `handsontable/AGENTS.md`, in the Trusted Types sink rule.
- The shared comment above the surviving-sink tests in `tests/e2e/trusted-types.spec.ts`. The per-test comments below it were already correct, so only the block above them was stale.
- `tests/fixtures/demo/trusted-types.html`, where it explained why `?colHeader=prose` exists.
- Two `ghostTable.unit.js` tests named and commented in the vocabulary of escaping a label into an HTML string. `buildHeaderLabel()` calls `fastInnerHTML` on a span, which for non-matching content resolves to `fastInnerText` and builds a text node, so nothing is escaped and nothing is concatenated. The assertions already exercised the text path; only the stated mechanism was wrong.

Both page objects that drive these fixtures also gained the bundle-load wait `tests/AGENTS.md` requires. `TrustedTypesPage.goto()` and `HeaderMarkupGatePage.goto()` asserted on rendered output with `expect`, whose 10s timeout a cold server loading a ~6 MB bundle per worker can outlast. Not hypothetical: a cold run of `trusted-types.spec.ts` failed its first test at that assertion and passed the remaining 16. With the wait it passes 23/23 cold across both specs.

The regex, its JSDoc, and the changelog entry are unchanged. The capture-group removal in point 1 above stays documented in the JSDoc rather than the changelog, since a regex's group count is not user-facing behavior.

### Types of changes

The changelog entry is `type: "changed"` rather than `fixed`, because an existing configuration's rendering output changes as described above.

### Related issue(s):

1. DEV-2642

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2642

---

> [!NOTE]
> **Medium Risk**
> Changes a public `HTML_CHARACTERS` API and header rendering semantics (sanitizer visibility, entity decoding), affecting Trusted Types, nested-header sizing, and any custom code that relied on the old loose match.
> 
> **Overview**
> **Narrows `HTML_CHARACTERS`** so `fastInnerHTML` only uses `innerHTML` when content looks like a real tag, declaration, or character reference—not when a header merely contains `&`/`;` or `<`/`>`. Ordinary labels such as `Smith & Sons, Ltd.; est. 1920` now go through **`fastInnerText`**, which fixes **Trusted Types** grids that previously failed to construct because those strings hit an unguarded sink.
> 
> The public regex is rewritten for **linear time** (`[^<>]*` instead of `[^>]*`), drops **capture groups**, and documents deliberate edge cases (e.g. semicolon-less `&copy 2024` renders literally; prose shaped like `<Enter>` still matches). **Configured sanitizers** are no longer invoked for non-markup on the `fastInnerHTML` path (clipboard `sanitizeHTML` unchanged); missing-sanitizer warnings follow the same gate.
> 
> Docs (**security guide**, **AGENTS.md**, changelog) and tests are expanded: unit coverage for the gate and ghost-table measurement, new Playwright **header-markup-gate** suite, and **trusted-types** prose header expectation flipped from construct failure to success.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0f03b2602a59c3d7cf7ecb8fa16e1e998c27502e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2617 Trusted Types: documented support does not work, and cannot be a default sanitization strategy](https://app.clickup.com/t/9015210959/DEV-2617)
Task linked: [DEV-2642 HTML_CHARACTERS routes plain text through the HTML path: any `&` with a later `;` matches](https://app.clickup.com/t/9015210959/DEV-2642)

handsontable/src/helpers/dom/element.ts
```diff
@@ -540,7 +540,54 @@ export function empty(element: HTMLElement): void {
...
+ *
+ * It stays unanchored, so a string mixing prose with a real reference (`Smith & Sons; &amp; more`)
+ * still matches on the reference and still reaches the sanitizer.
+ *
+ * The `i` flag carries the case-insensitivity that tag names and hexadecimal digits need anyway
+ * (`<IMG SRC=x>`, `&#X41;`), which is what lets the character classes stay this short. There is
+ * deliberately no `g` flag: it would make `.test()` stateful through `lastIndex`, so consecutive
+ * calls on the same pattern would disagree about identical content.
+ */
+export const HTML_CHARACTERS = /(<\/?[a-z][^>]*>)|(&[a-z][a-z\d]+;)|(&#x?[\da-f]+;)/i;
```
Haxikowy: The three-alternative shape changes the capture-group semantics of a publicly reachable export (`Handsontable.dom.HTML_CHARACTERS`, plus the deep import `handsontable/helpers/dom/element`), not just the match set:

```
'<b>ID</b>'   old ["<b>","<b>","b",null]        new ["<b>","<b>",null,null]
'a &amp; b'   old ["&amp;","&amp;",null,"amp"]  new ["&amp;",null,"&amp;",null]
'&#x1F600;'   old [...,null,"#x1F600"]          new [...,null,"&#x1F600;"]
```

Group 1 used to always equal `match[0]`; it is now `undefined` for both entity branches, and groups 2 and 3 no longer hold the tag name / entity name. A consumer reading `m[1]` gets `undefined` silently.

No in-repo caller reads them, so the cheapest fix is to make all three non-capturing and delete the group test:

```suggestion
export const HTML_CHARACTERS = /(?:<\/?[a-z][^<>]*>)|(?:&[a-z][a-z\d]+;)|(?:&#x?[\da-f]+;)/i;
```

That suggestion also carries a second point: `[^>]*` keeps the tag branch quadratic. `'<a'.repeat(40000)` blocks the main thread for **4.6 s** — same as the old pattern, while the entity branch you fixed went 7.3 s to 3 ms. `[^<>]` gives 0.25 ms with identical results on `<div title="<b>">`, `<a b=">" onload=x>`, ``, `</span>`, `<svg onload=alert(1)>`. The one divergence, `<a x="<">`, lands on the text path — the inert side. This is reachable from paste, not just render: `sanitizeHTML` runs the gate over the whole `text/html` clipboard payload (`copyPaste.ts:977`).

Haxikowy: "Cell values were never affected" in the description does not hold. `fastInnerHTML` runs this test *before* the sanitizer branch, so `sanitizer: false` does not bypass it — and `htmlRenderer.ts:22` passes exactly that. A `type: 'html'` cell containing `<!-- generated -->`, `<!DOCTYPE html>`, `<![CDATA[x]]>`, `<?xml?>` or `&copy 2024` now renders as literal visible text instead of parsing. Same for `allowHtml` autocomplete sources, context-menu item names, dialog and notification content, and select-editor options.

DEV-2642 asked to narrow only the entity branch; narrowing the tag branch is what causes this, and it is undeclared scope. Either add a fourth alternative `(?:<[!?][^<>]*>)` to keep declarations on the HTML path, or state the reclassification in the body and document it on the `html` cell type page.

Do not instead bypass the gate when `sanitizer === false` — that would send every plain-text `html` cell into `innerHTML` and reintroduce the Trusted Types throw for the cell type most likely to be used on a TT page.

handsontable/src/plugins/nestedHeaders/utils/ghostTable.ts
```diff
@@ -292,7 +293,27 @@ class GhostTable {
...
+    // The sanitized branch below is NOT covered by that reasoning and is still interpolated raw,
+    // so the same concatenation defect remains reachable through it: a label of
+    // `&amp;<img src=x onerror=alert(1) ` matches the gate, passes a pass-through sanitizer (the
+    // default) unchanged, and the trailing `</span>` completes the tag - live in this copy while
+    // inert in the rendered header, which writes the label alone and lets the unterminated tag hit
+    // EOF. Fixing it means not building this table by string concatenation at all: write the label
+    // with `fastInnerHTML` into an already-built span, which makes the mirror exact by
+    // construction and removes the `innerHTML` assignment in `buildWidthsMap()` too. That is a
+    // rework of this class rather than a change to this line, and it is tracked separately.
+    const label = HTML_CHARACTERS.test(rawLabel)
```
Haxikowy: This closes a live injection and the description does not say so. Before it, a label of ``) failed the gate, was interpolated raw, and the template's trailing `</span>` completed the tag — the container is appended to `document.body` at `:118`, so in Chrome the element is created and `onerror` fires. The rendered header is inert for the same label, because `fastInnerHTML` writes the label alone and the unterminated tag hits EOF.

Worth a `security`-typed changelog entry, or at least a line in the body and in "Types of changes".

On the half left open at `:306-313`: I searched ClickUp twice and found no ticket for the concatenation rework. Please file it and cite the ID here — otherwise this comment is the only record, and with the default pass-through sanitizer the same path stays reachable through the sanitized branch.

handsontable/src/helpers/dom/__tests__/element.unit.ts
```diff
@@ -799,6 +801,124 @@ describe('DomElement helper', () => {
...
+      // differently depending on what was tested before it.
+      expect(HTML_CHARACTERS.global).toBe(false);
+
+      const content = '<b>ID</b>';
+
+      expect(HTML_CHARACTERS.test(content)).toBe(true);
+      expect(HTML_CHARACTERS.test(content)).toBe(true);
+    });
+
+    it('should expose one capture group per form, with exactly one defined per match', () => {
```
Haxikowy: > `// Nothing in the grid reads the groups; the count stays at three, as it always has been.`

The count stayed at three; the meaning did not (see the `element.ts:590` thread). As written, this test pins the arity while its comment reads as reassurance that the shape is compatible — the one line a reviewer would trust.

If the groups go non-capturing, delete this block. If they stay, assert the contract that matters (`match[0]` is the matched markup) and fix the comment.

handsontable/src/helpers/dom/__tests__/element.unit.ts
```diff
@@ -799,6 +801,124 @@ describe('DomElement helper', () => {
...
+      // shapes a future narrowing would drop out of the sink without any other test noticing.
+      expect(HTML_CHARACTERS.test('<x:y>a</x:y>')).toBe(true);
+      expect(HTML_CHARACTERS.test('<svg:script>alert(1)</svg:script>')).toBe(true);
+      // A `>` inside an attribute value, so the first `>` does not close the tag.
+      expect(HTML_CHARACTERS.test('<img src="a>b" onerror=alert(1)>')).toBe(true);
+      expect(HTML_CHARACTERS.test('<img/src=x onerror=alert(1)>')).toBe(true);
+      expect(HTML_CHARACTERS.test('<IMG SRC=x ONERROR=alert(1)>')).toBe(true);
+    });
+
+    it('should keep matching prose shaped like a tag, which the pattern cannot exclude', () => {
```
Haxikowy, line 874: Fine as a pinned limitation — the contrasting `AT&T; Inc.` assertion is what keeps it behavioral rather than a snapshot.

The substance deserves the PR body, though, not only a test comment: `<none>`, `<n/a>`, `<blank>` and `Type <Enter> to continue` are header labels with no markup intended, they still take `innerHTML`, and they still take the whole grid down under Trusted Types. Same accidental-crash class DEV-2642 exists to fix, now with a passing test blessing it.

Either close it here (known-tag-name allowlist plus a `-` rule to keep custom elements) or write down that the ticket is scoped down.

tests/fixtures/pages/HeaderMarkupGatePage.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  /** The spanning label of the top nested-header row. */
+  nestedHeader(): Locator {
+    return this.page.locator('.ht_clone_top thead tr').first()
+      .locator('th[colspan="2"] .colHeader');
+  }
+
+  /** Whether the grid constructed without throwing. */
+  async constructed(): Promise<boolean> {
```
Haxikowy: `constructed()` has no caller, and the fixture's `__constructed` / `__constructError` plumbing is unused with it. So the headline symptom — a grid failing to construct from a prose `colHeaders` under Trusted Types — has no regression test on this branch; the only spec asserting it lives on `feature/DEV-2617_Trusted-Types`.

Either add a TT-enforcing fixture leg asserting `constructed() === true` for a prose header, or drop the dead helper and say in the body that the crash is covered only by DEV-2617 — which makes merge order a blocker rather than a coordination note.

tests/e2e/header-markup-gate.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+  test('renders a prose header holding `&` and `;` as text', async({ page, theme, bundle }) => {
+    const grid = new HeaderMarkupGatePage(page, theme, bundle);
+
+    await grid.goto({ colHeader: 'prose' });
+
+    await expect(grid.columnHeader()).toHaveText('Smith & Sons, Ltd.; est. 1920');
+    // No element was built from it, so nothing was parsed as markup.
+    await expect(grid.columnHeader().locator('*')).toHaveCount(0);
+    // The label is not markup, so nudging the user toward a sanitizer would be noise.
+    expect(await grid.warnedAboutMissingSanitizer()).toBe(false);
```
Haxikowy, line 31: `tests/AGENTS.md` asks for `expect.poll` on data probes. These are one-shot `page.evaluate` reads taken right after visibility, so a warning or sanitizer call emitted by a later render pass flakes:

```suggestion
    await expect.poll(() => grid.warnedAboutMissingSanitizer()).toBe(false);
```

Same at `:32`, `:42`, `:54`, and for `sanitizerContents()` / `sanitizerContexts()`. Polling cannot fix the negative assertions on its own — anchor those behind a settled positive first.

.changelogs/13264.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Narrowed the test that decides whether cell and header content is written as HTML, so text containing an ampersand or a semicolon is no longer treated as markup. Such content could stop a grid from rendering under a Trusted Types policy.",
+  "type": "fixed",
```
Haxikowy: Two problems with the title. The old gate needed an `&` **and** a later `;` — either alone never matched, which your own test at `element.unit.ts:823` pins — so "an ampersand or a semicolon" describes neither the old nor the new behavior. And the change equally covers `<`/`>`, HTML comments, doctypes, CDATA, PIs and the semicolon-less `&copy 2024`, none of which appear here.

`type` also looks wrong: `fixed` names the crash, but an existing config's rendering output changes, which is `changed`. Either make this one entry `changed`, or ship two.

```suggestion
  "title": "Content shaped like an HTML tag or a character reference is now what routes cell and header text through the HTML write path. Ordinary text holding `&`, `;`, `<` or `>` is written as plain text, which keeps it from stopping a grid from rendering under a Trusted Types policy. Content shaped like an HTML comment, a doctype, or a character reference without its closing semicolon now renders literally.",
  "type": "changed",
```

Haxikowy:
**Merge order with DEV-2617 needs deciding here, not later.** `git merge-tree origin/feature/DEV-2617_Trusted-Types <this branch>` gives content conflicts in `handsontable/src/helpers/dom/__tests__/element.unit.ts` and `handsontable/src/plugins/nestedHeaders/utils/ghostTable.ts` — two places, not the one the description names. And the conflict is not textual: DEV-2617 replaces `#buildHeaderLabelHTML` with a version that builds a real DOM node and calls `fastInnerHTML` on it, deleting the `HTML_CHARACTERS`/`sanitizeHTML` mirroring this PR just edited. If 2617 lands first, this PR's ghostTable hunk and its 119-line `ghostTable.unit.js` block are dead on arrival, and develop goes red the moment this merges, because `trusted-types.spec.ts:176` asserts `CONSTRUCT-THREW` for `Smith & Sons, Ltd.; est. 1920`. DEV-2642's own description assigns that edit to this PR; the body punts it to whichever branch is second. Pick one.

**The description quotes a regex that is not the one shipped.** Body: `/(<(\/?[a-zA-Z][^>]*)>|&([a-zA-Z][a-zA-Z0-9]+|#[0-9]+|#[xX][0-9a-fA-F]+);)/`. Code: `/(<\/?[a-z][^>]*>)|(&[a-z][a-z\d]+;)|(&#x?[\da-f]+;)/i`. Match sets diverge only on the `&#<hex digits, no x>;` family, but the two designs are mutually exclusive — the body's single outer group preserves `match[1] === match[0]`, the shipped flat form does not, and the JSDoc justifies the flat form on the `typescript:S5843` budget. So the body's "capture-group count stays at 3 for the same reason" is true about the count and false about the compatibility it is offered as evidence for. Worth a pass over the whole body, since a reviewer reading it is reviewing code that does not exist.

**The test-count claim is understated.** Reverting only line 590 back to the old pattern gives `Tests: 12 failed, 3979 passed, 3991 total` / `Test Suites: 3 failed, 322 passed, 325 total` — eight in `element.unit.ts`, two in `ghostTable.unit.js`, two in `sanitizer.unit.ts`, not seven. Baseline on the unmodified head is 3991 passed / 325 suites, not the 3984 in the body. Both numbers err toward under-selling the coverage, so the conclusion stands; they just need re-running.

**The missing-sanitizer warning goes quiet where the payload still reaches a parser.** `sanitizeHTML` reuses the routing gate for the warning, so `<!-- note -->`, `<!doctype html>`, `<?xml ... ?>`, `<0>` and `&#;` no longer trigger it. On the clipboard path the payload reaches `htmlToGridSettings`/`DOMParser` regardless of the gate, so nothing is safer — only the operator's signal is lost, and that warning is the only indication a grid writes raw HTML with no sanitizer configured. Consider a looser test for the warning than for the routing, since over-warning there costs nothing.

**Two housekeeping items.** `.ai/DOC-STANDARDS.md` rule 11 asks for en dashes as clause separators in JSDoc; the new comment uses ASCII hyphens in roughly eight places (`element.ts:546,559,569,576`, `ghostTable.ts:301`, plus the test comments). And the invariant the JSDoc establishes — any further narrowing must keep `<x:y>`, `<svg:script>` and `b" ...>` reaching the sanitizer — belongs in `handsontable/.ai/CONCERNS.md` next to the existing "No Built-in HTML Sanitizer" notes, per checklist item 3, rather than only in a source comment.

**Docs are not falsified anywhere** — nothing published documents the gate condition. Two pages become incomplete rather than wrong: the `html` cell type / cell-renderer page, for the reclassification in the `element.ts:590` thread, and `docs/content/guides/security/security/security.md:66`, where "It receives the raw HTML string" is now less true on the `fastInnerHTML` path.

**The green Performance check is not evidence either way.** None of the nine scenarios uses `nestedHeaders`, `htmlRenderer`, a prose header, or anything else that exercises the differing branch, so it never runs the changed code. Measured separately, the change is a net win — the entity branch's quadratic backtracking is gone (`'&a'.repeat(40000)`: 7.3 s to 3 ms), prose headers that stop invoking the HTML parser are 7 to 8x cheaper per write, and the per-call cost on the common case moves by under 0.3 microseconds per render frame. The one residual is in the `element.ts:590` thread.

handsontable/src/helpers/dom/element.ts
```diff
@@ -540,7 +540,54 @@ export function empty(element: HTMLElement): void {
   }
 }
 
-export const HTML_CHARACTERS = /(<([^>]*)>|&([^;]*);)/;
+/**
+ * Decides whether a string is written through `innerHTML` (markup, so it has to reach the
+ * sanitizer) or `textContent` (plain text, which cannot inject anything).
+ *
+ * Both alternatives are deliberately shaped to require something that could actually be parsed as
```
budnix: Small drift from the last commit — this says two alternatives, but line 555 and line 558 below both say three. Splitting the entity branch into named and numeric left this paragraph behind, so a reader who counts from here can miss `&#x?[\da-f]+;` entirely.

```suggestion
 * All three alternatives are deliberately shaped to require something that could actually be parsed as
```

handsontable/src/helpers/dom/element.ts
budnix: Separate from the two threads above: I don't see a docs change in this PR. `docs/content/guides/security/security/security.md` has a "What the sanitizer covers" table that lists `'header'` and says nothing about a shape test gating the call. `fastInnerHTML` runs this pattern before it ever reaches `sanitizeHTML`, so after this change a label like `Smith & Sons, Ltd.; est. 1920` never gets to a configured sanitizer, where it used to. Anyone using `sanitizer` as an audit hook or a length cap quietly stops seeing a class of strings — worth a line in that table.

handsontable/src/plugins/nestedHeaders/utils/ghostTable.ts
```diff
@@ -292,7 +293,27 @@ class GhostTable {
...
     const rawLabel = String(headerSettings.label);
-    const label = HTML_CHARACTERS.test(rawLabel) ? sanitizeHTML(this.hot, rawLabel, 'header') : rawLabel;
+    // A label that holds no markup is escaped rather than interpolated as-is. `HTML_CHARACTERS`
+    // answers "does this substring hold markup", but the template below supplies the closing `>`
+    // the label may be missing, so "no markup in the label" never implied "no markup in the
+    // result": a label of `<img src=x onerror=...` (no `>`) parses into a live element once
+    // `</span>` follows it. Escaping measures identically - an escaped `&` or `<` renders as the
+    // same single glyph - so column widths still match the rendered header, which writes the same
+    // text through `fastInnerText`. It also fixes four labels this table used to measure at a
+    // width the header never shows, `<!-- note -->` and `&copy 2024` among them.
```
budnix: `<!-- note -->` doesn't belong in this list. Before this PR the two sides already agreed on it: the old gate matched, so the rendered header went through `innerHTML` and the parser ate it as a comment, and the ghost table interpolated it raw and the parser ate it there too — zero width on both sides. What changed that label's width is the regex narrowing in `element.ts`, not the escaping on this line. `&copy 2024` is a real example, since the old ghost table decoded it to `©` while the header showed the literal text, so the sentence mixes one true case with one that isn't.

handsontable/src/plugins/nestedHeaders/utils/ghostTable.ts
```diff
@@ -292,7 +293,27 @@ class GhostTable {
...
+    //
+    // The sanitized branch below is NOT covered by that reasoning and is still interpolated raw,
+    // so the same concatenation defect remains reachable through it: a label of
+    // `&amp;<img src=x onerror=alert(1) ` matches the gate, passes a pass-through sanitizer (the
+    // default) unchanged, and the trailing `</span>` completes the tag - live in this copy while
+    // inert in the rendered header, which writes the label alone and lets the unterminated tag hit
+    // EOF. Fixing it means not building this table by string concatenation at all: write the label
+    // with `fastInnerHTML` into an already-built span, which makes the mirror exact by
+    // construction and removes the `innerHTML` assignment in `buildWidthsMap()` too. That is a
```
budnix: This treats `buildWidthsMap()`'s `innerHTML` as cleanup that comes with the rework, but I think it's load-bearing for the changelog claim. `#buildGhostTable` assigns a plain string to `container.innerHTML` at `:226`, so under an enforced Trusted Types policy any grid using `nestedHeaders` still throws — whatever the label holds, prose included. The template literal below also stringifies whatever `sanitizeHTML` returns, so a sanitizer that correctly hands back `TrustedHTML` gets coerced to a plain string right here and can't rescue it.

Either fixing that assignment in this PR or narrowing the changelog title to the `fastInnerHTML` path would keep the entry honest.

tests/e2e/header-markup-gate.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+ * The boundary is deliberate. Prose is fixed here; a label holding real markup keeps
+ * reaching `innerHTML`, because that is content the user asked to be rendered as HTML.
+ */
+test.describe('header markup gate', () => {
+  test('renders a prose header holding `&` and `;` as text', async({ page, theme, bundle }) => {
+    const grid = new HeaderMarkupGatePage(page, theme, bundle);
+
+    await grid.goto({ colHeader: 'prose' });
+
+    await expect(grid.columnHeader()).toHaveText('Smith & Sons, Ltd.; est. 1920');
```
budnix, line 18: Did these two assertions end up weaker than intended? I believe they pass on `develop` as well. The old gate did match `Smith & Sons, Ltd.; est. 1920`, but `& Sons, Ltd.;` is not a character reference, so `innerHTML` produced the same text and no child elements. Same story for `Score < 50 > threshold`, where `<` followed by a space is plain text to the parser. That leaves line 22 as the only line here that fails without the fix.

A leg whose parse really differs would give the text assertions teeth — `<!-- draft -->` rendered as an empty header on the old gate and renders as visible literal text now. Worth adding if that reclassification survives the other thread.

tests/e2e/header-markup-gate.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+    expect(await grid.sanitizerContents()).not.toContain('Smith & Sons, Ltd.; est. 1920');
+  });
+
+  test('still routes a nested-header label holding markup to the sanitizer', async({ page, theme, bundle }) => {
+    const grid = new HeaderMarkupGatePage(page, theme, bundle);
+
+    await grid.goto({ nested: 'markup', sanitizer: 'truncate' });
+
+    expect(await grid.sanitizerContents()).toContain('<b data-testid="markup-marker">ID</b>');
+    // Under the `'header'` source, never `'innerHTML'`: a context-aware sanitizer must not
```
budnix, line 95: This assertion can't tell the ghost table apart from the rendered header. `updateCellHeader` hands the sanitizer the identical string with the identical `'header'` source, and `__sanitizerCalls` records only `(content, source)` — so if `#buildHeaderLabelHTML` stopped calling the sanitizer tomorrow, this test would stay green and the test name would still claim the coverage.

The prose test above is fine in this respect, since a wrong ghost-table call shows up as an extra entry. This direction needs a third observable — the measured column width, or the call count. Curious which one you'd rather hang it on.

handsontable/src/helpers/dom/element.ts
demtario: Both taken.

**Capture groups.** You're right that the flat shape renumbered them, and that was the worse half of it: `match[1]` silently becoming `undefined` is exactly the kind of change nobody notices until it bites. All four alternatives are now non-capturing, and the group test is replaced with one that pins `match[0]` plus `[...m]` having length 1.

**`[^>]*`.** Confirmed and fixed. Measured here on `'<a'.repeat(40000)`: 1304 ms with `[^>]*`, 0.25 ms with `[^<>]*`. I added a unit test that asserts the whole thing answers in under 100 ms, so a future widening of that class shows up as a red test rather than as a slow paste.

The one divergence is pinned as deliberate, together with the case that shows how narrow it is:

```js
expect(HTML_CHARACTERS.test('<a x="<">')).toBe(false);          // only tag spelled that way -> text path
expect(HTML_CHARACTERS.test('<a x="<"><b>ID</b>')).toBe(true);  // any other tag present -> still matches
```

Current shape:

```js
/(?:<\/?[a-z][^<>]*>)|(?:<[!?])|(?:&[a-z][a-z\d]+;)|(?:&#x?[\da-f]+;)/i
```

demtario: You're right, and the claim was wrong in the description rather than just imprecise. `fastInnerHTML` runs the test before it looks at `sanitizer`, so `sanitizer: false` does not skip it and `htmlRenderer` passes exactly that. I've removed the "cell values were never affected" sentence from the body.

I took the first of your two options: a fourth alternative, `(?:<[!?])`, so a comment, doctype, CDATA section, or processing instruction keeps the HTML path. That removes the reclassification for `type: 'html'` cells, `allowHtml` sources, menu item names, dialog and notification content, and select-editor options in one go, rather than declaring a behavior change across all of them.

It deliberately does not require a closing `>`. An unterminated declaration is still not prose, and leaving the `>` out keeps the alternative free of a quantifier, which matters because the `typescript:S5843` budget is 20 and this lands at 19.

Verified the prose fix is unaffected by the new branch: `Smith & Sons, Ltd.; est. 1920`, `R&D; notes`, `cost: 5 & up; tax incl.`, `Score < 50 > threshold`, `a < b`, `3 > 2`, `AT&T; Inc.` and `<3 > 5` all still take the text path.

And noted on the trap: I did not touch the `sanitizer === false` path. Bypassing the gate there would send every plain-text `html` cell into `innerHTML`, which is the wrong direction for exactly the pages this ticket is about.

The one residual reclassification left is the semicolon-less legacy form, `&copy 2024`, which now renders literally. That is in the changelog title and the JSDoc.

handsontable/src/plugins/nestedHeaders/utils/ghostTable.ts
demtario: This resolved in a way neither of us expected: the PR is now rebased onto `feature/DEV-2617_Trusted-Types`, and that branch has already rebuilt this class to construct the ghost header as DOM and write the label with `fastInnerHTML(...)` instead of interpolating it into a string. It also replaced the `container.innerHTML` assignment in `buildWidthsMap()`.

So the concatenation defect is fixed at its root on the new base, which makes my `escapeHtml` workaround redundant. I dropped it and took that implementation. Consequences:

- No ticket to file. The rework you were asking me to record already exists on the branch this now sits on, so there is nothing left open to track.
- The `security` changelog entry I had added is removed. The injection is fixed by the base, not by this PR, and claiming it here would misattribute it.
- The ghost-table unit tests stay. They now cover the parity that implementation provides: a prose label never reaches the sanitizer, a markup label and a real character reference both do, and an unterminated tag builds no element.

Thanks for pushing on this one. The escaping I had written would have shipped a second, weaker implementation of parity that the base already achieves structurally.

handsontable/src/helpers/dom/__tests__/element.unit.ts
demtario: Deleted. The groups are non-capturing now, so there is nothing to pin about arity, and you were right that the comment was the misleading part: it read as compatibility reassurance while the meaning had moved.

What replaced it asserts the contract that actually matters:

```js
expect('<b>ID</b>'.match(HTML_CHARACTERS)[0]).toBe('<b>');
expect([...'<b>ID</b>'.match(HTML_CHARACTERS)]).toEqual(['<b>']);
```

handsontable/src/helpers/dom/__tests__/element.unit.ts
demtario: Written down rather than closed, and the ticket is scoped down accordingly.

I kept the test and its contrast case, and the limitation is now stated in the PR description and in the JSDoc: a label shaped like a tag still reaches `innerHTML`, so it still throws under an enforced policy, and the accidental-crash class is reduced rather than eliminated.

I did not add the tag-name allowlist. It would need a `-` rule for custom elements, and it would still be a guess about which names are "real" for a grid whose headers can legitimately contain any element. That is a bigger decision than this ticket, and I would rather it be an explicit one than something I slipped in under a regex fix.

tests/fixtures/pages/HeaderMarkupGatePage.ts
demtario: Both halves addressed, and the second one better than the option you offered.

The dead helper and the fixture's `__constructed` / `__constructError` plumbing are gone.

On coverage: the PR is now rebased onto `feature/DEV-2617_Trusted-Types`, so `tests/e2e/trusted-types.spec.ts` is on this branch. Its prose case asserted `CONSTRUCT-THREW`, which inverts under this fix, so I flipped it:

```ts
test('constructs for a header holding no markup at all, only an ampersand and a semicolon', ...
  expect(await grid.statusText()).toContain('CONSTRUCTED');
  await expect(grid.cell(0, 0)).toHaveCount(1);
```

That makes the headline symptom a real regression test under `require-trusted-types-for 'script'`, in the same PR as the fix, instead of a coordination note across two. The `markup` case above it is untouched and still throws, which keeps the surviving boundary pinned. Merge order is now expressed by the base branch rather than by a comment.

tests/e2e/header-markup-gate.spec.ts
demtario: Applied to every probe that can settle: the positive warning assertions and both sanitizer-content reads are now `expect.poll`.

The two negative warning assertions stay single reads, deliberately. Nothing in those legs is markup, so no warning ever fires and there is no positive for a poll to settle on. Polling a negative there just re-reads the same `false` until it times out.

I tried your anchoring suggestion first and it doesn't work here for the same reason: I had it wait for `warningCount() > 0`, which can never become true in a leg whose only header is prose. What I did instead covers the actual risk you named, a later render pass emitting one:

```ts
expect(await grid.warnedAboutMissingSanitizer()).toBe(false);

await grid.rerender();
await expect(grid.columnHeader()).toHaveText('Smith & Sons, Ltd.; est. 1920');
expect(await grid.warnedAboutMissingSanitizer()).toBe(false);
```

If you would rather see a settle-then-assert helper for negatives generally, that seems worth doing once in `fixtures/` rather than per spec, and I would take that as a follow-up.

.changelogs/13264.json
demtario: Took your title almost verbatim, and `type: "changed"`.

You were right on both counts. The old gate needed an `&` **and** a later `;`, so "an ampersand or a semicolon" described neither behavior, and the entry said nothing about `<`/`>` or the declaration shapes.

One deviation from your suggested text: comments and doctypes are no longer part of the reclassification, because the other thread's fourth alternative keeps them on the HTML path. So the title now names only the semicolon-less reference form, `&copy 2024`, as rendering literally.

I also briefly shipped a second, `security`-typed entry for the ghost-table injection, per your other comment. It is removed again: after the rebase onto the Trusted Types branch, that fix belongs to the base rather than to this PR.

handsontable/src/helpers/dom/element.ts
demtario: Fixed. That paragraph predated the split and then fell further behind when I added the declaration alternative, so it was stale by two rather than one.

It now reads "Every alternative is deliberately shaped to require something that could actually be parsed as markup", which cannot drift as the count changes again.

handsontable/src/helpers/dom/element.ts
demtario: Added. `docs/content/guides/security/security/security.md` now says so directly, in the "What the sanitizer covers" section right after the table:

> Handsontable calls the sanitizer only for content that is shaped like markup: something that reads as an HTML tag, a markup declaration, or a character reference. Text that holds none of those is written as text, so it never reaches a sink and never reaches your sanitizer. `Smith & Sons, Ltd.; est. 1920` and `Score < 50 > threshold` are written as text, while `<b>ID</b>` and `a &amp; b` are sanitized and parsed. Keep this in mind if you use the sanitizer as an audit hook or to cap length rather than to strip markup: it does not see every value the grid writes. The clipboard sources are the exception, and they receive every payload.

The audit-hook and length-cap framing is yours and it is the part that makes the note useful, so I kept it. The clipboard carve-out is worth stating too, since `sanitizeHTML` calls a configured function above the gate and those sources really do see everything.

handsontable/src/plugins/nestedHeaders/utils/ghostTable.ts
demtario: Correct, and thank you for separating the two cases. Before this PR the rendered header and the ghost table agreed on `<!-- note -->` at zero width, so it was never a mismatch the escaping fixed; `&copy 2024` was the real one, decoded on one side and literal on the other.

The line is gone either way. After the rebase onto the Trusted Types branch I took that branch's DOM-based implementation, so my escaping and its comment no longer exist.

And the case has disappeared upstream of that too: the fourth alternative added for the other reviewer keeps comments on the HTML path, so `<!-- note -->` is not reclassified at all now.

handsontable/src/plugins/nestedHeaders/utils/ghostTable.ts
demtario: You were right that this was load-bearing rather than cleanup, and it is now resolved by the base rather than by a narrowed title.

The PR is rebased onto `feature/DEV-2617_Trusted-Types`, which replaced that `container.innerHTML` assignment with DOM construction and writes each label through `fastInnerHTML`. Both problems you named go with it: nothing stringifies a `TrustedHTML` back into a plain string, and a grid using `nestedHeaders` no longer throws under enforcement. `tests/e2e/trusted-types.spec.ts`'s "constructs a grid with nested headers" passes on this branch, which is the direct evidence.

So the changelog claim stands without narrowing, on this base. It would not have on `develop`, and I would have had to take your second option there.

tests/e2e/header-markup-gate.spec.ts
demtario: You're right that both text assertions pass on `develop`, and it is worth being precise about why: it is structural, not an oversight. For input that is genuinely prose, `innerHTML` and `textContent` produce identical DOM. `& Sons, Ltd.;` is not a character reference and `<` followed by a space is text to the parser, exactly as you say. That identity is the reason rerouting prose is safe, so no prose leg can ever distinguish the two paths by parse result.

Your `<!-- draft -->` suggestion would have had teeth, but that reclassification did not survive the other thread: comments now keep the HTML path via a fourth alternative, so there is no visible difference there either.

What does fail without the fix, on this branch:

- line 22, the missing-sanitizer warning, as you spotted
- the nested-header legs, through the sanitizer call list
- `tests/e2e/trusted-types.spec.ts`, whose prose case I flipped from `CONSTRUCT-THREW` to `CONSTRUCTED` — the real discriminator, under an enforced policy, now in this same PR

I kept the text assertions as documentation of the rendered result rather than as proof, since they do fix the output shape even if they don't discriminate.

tests/e2e/header-markup-gate.spec.ts
demtario: Good catch, and I went with the call count.

```ts
await expect.poll(() => grid.sanitizerCallsFor('<b data-testid="markup-marker">ID</b>'))
  .toBeGreaterThanOrEqual(2);
```

`sanitizerCallsFor()` counts entries in `__sanitizerCalls` matching that exact content, so dropping either surface takes it to 1 and the test goes red.

I preferred it over the measured width for two reasons: width is theme-dependent and this spec runs across three themes and two bundles, and a width assertion would prove "something measured it" rather than "the sanitizer saw it", which is the claim in the test name. Verified it passes at the moment with both surfaces calling, and the count is what fails if `#buildHeaderLabelHTML` stops.

demtario:
## Merge order: this PR is now stacked on DEV-2617

**Base changed from `develop` to `feature/DEV-2617_Trusted-Types`.** The branch was rebased onto it and force-pushed, so the eight review threads from @Haxikowy and the six from @budnix are anchored to rewritten commits and GitHub may show them as outdated. Every one of them has a reply; nothing was dropped.

### Why stack instead of coordinating

The two changes touch the same three files, and the conflict was real: a trial merge conflicted in `element.unit.ts` and `ghostTable.ts`. Beyond avoiding that, three things only work in this order.

**1. The headline symptom gets a regression test.** `tests/e2e/trusted-types.spec.ts` lives on the DEV-2617 branch, and its prose case asserted the bug:

```ts
test('throws for a header holding no markup at all, only an ampersand and a semicolon', ...
  expect(await grid.statusText()).toContain('CONSTRUCT-THREW');
```

It is flipped here to `CONSTRUCTED` with a rendered cell. That expectation inverting is not incidental — it *is* the fix, verified under `require-trusted-types-for 'script'` in a real browser. On a `develop` base this PR had no test for the crash it fixes, which is what @Haxikowy flagged; the only options were to leave merge order as a prose note or to build a second Trusted Types fixture beside the existing one.

**2. DEV-2617 already did the ghost-table rework, so two review findings dissolve.** Both reviewers independently found that `nestedHeaders/utils/ghostTable.ts` decided whether to sanitize a label and then interpolated it into a template that supplies the closing tag — so "no markup in the label" never implied "no markup in the result". On `develop` I fixed the reachable half by escaping non-markup labels, and left the other half open, because closing it properly meant not assembling that table from strings at all.

That rework exists on this base: the ghost header is built as DOM and each label is written with `fastInnerHTML`, and the `container.innerHTML` assignment in `buildWidthsMap()` is gone. So:

- my `escapeHtml` workaround is dropped in favour of it — it was a second, weaker implementation of parity the base achieves structurally
- the `security`-typed changelog entry is removed, since the injection is fixed by the base and not by this PR
- @budnix's point that `buildWidthsMap()`'s `innerHTML` was load-bearing for the changelog claim is answered by the base rather than by narrowing the claim: `tests/e2e/trusted-types.spec.ts`'s "constructs a grid with nested headers" passes under enforcement
- no follow-up ticket is needed, which is what @Haxikowy asked me to file and cite here

The ghost-table unit tests stay, now covering that implementation's parity rather than my escaping.

**3. The changelog claim becomes true.** "Keeps it from stopping a grid from rendering under a Trusted Types policy" would have been an overclaim on `develop` for any grid using `nestedHeaders`, because that `innerHTML` assignment threw regardless of what the labels held.

### What this costs

- **This cannot merge to `develop` on its own.** It is blocked on DEV-2617. DEV-2642 was split out to be independent, and stacking gives that up in exchange for the three points above.
- If DEV-2617 is going to be a while, say so and I will unstack: base back to `develop`, drop the `trusted-types.spec.ts` change, restore the `escapeHtml` fix and its `security` entry, and re-file the ghost-table rework as a follow-up ticket. That is the state the branch was in at `f035a2bce`, which is preserved locally.
- The DEV-2617 branch is currently 6 commits behind `develop` (`AutoRowHeaderSize`, frozen-overlay scrollbars, the `pivot()` fix, a docs change, a dependency removal). Those are not in this PR's diff — the rebase replays only this ticket's commits — but they will need to reach `develop` through DEV-2617 in the usual way.

### Gates on the new base

3885 unit tests across 323 suites · 9721 Jasmine specs, 0 failures · 446 Playwright, 1 skipped · eslint 0 errors · stylelint and `tests` lint clean. The 23 specs covering the Trusted Types suite and the new gate spec all pass together.

Haxikowy (dismissed):

budnix (dismissed):

budnix (approved):

PR #13265, opened 2026-08-27
DEV-2641: Keep root-only UI plugins disabled in a nested grid

demtario:
### Context

The PR fixes a crash when `dialog`, `notification`, `emptyDataState`, `pagination`, or `loading` is enabled inside a nested grid, that is the grid the `handsontable`, `autocomplete`, and `dropdown` cell types create (all three go through `HandsontableEditor#open`, so all three take the `handsontable: { ... }` passthrough).

All five are root-only UI plugins: they render into `ht-overlay` or the root grid element and register a focus scope (`loading` renders through the Dialog plugin). On a non-root instance none of that exists (`core.ts` skips the wrapper skeleton, `focusScopeManager` and `layoutManager` at `:364`, `:382-417`, `:6663-6670`), yet all five asked for it unconditionally from `enablePlugin()`. Opening the nested editor therefore threw and took the render down.

Verified in a browser against the released 18.0.0 build, the crash comes in two shapes:

- `dialog`, `emptyDataState`, `loading`: `The FocusScopeManager is only available for the main Handsontable instance.`
- `notification`: `Cannot read properties of null (reading 'ownerDocument')`
- `pagination`: `Cannot read properties of null (reading 'querySelector')`

The middle two die on the null root element inside their UI class before they even reach the focus scope. `loading` reaches the throw indirectly: it turns the nested grid's `dialog` setting on, and Dialog then crashes enabling. `{ type: 'autocomplete', handsontable: { dialog: true } }` and the `dropdown` equivalent were checked on the same build and throw identically.

Reproduction:

```js
handsontable({
  data: createSpreadsheetData(2, 2),
  columns: [{
    type: 'handsontable',
    handsontable: {
      data: createSpreadsheetData(2, 2),
      dialog: true,
    },
  }],
});
// select A1, press Enter
```

The fix makes `isEnabled()` return `false` on a non-root instance for the five plugins, so they decline to enable instead of throwing. That matches what the DOM layer already did: `dialog/ui.ts` appends nothing when there is no overlays element, so a nested grid never showed a dialog anyway. It also matches how core itself treats a non-root instance, silently skipping the accessibility announcer, the license surfaces, the `LayoutManager` and `registerAllFocusScopes`.

#### Why the refusal is silent, unlike `isHardConflictBlocked()`

The hard-conflict path warns and writes the setting back to `false`, and that is right for it: a conflict is a property of one grid the user configured, so the message fires once per grid and the user can act on it. This refusal is not shaped like that.

A nested grid is constructed fresh on **every** editor open (`handsontableEditor.ts:92`) and destroyed on close (`:81`), so the same configuration is re-evaluated on every cell edit. A warn there repeats for the lifetime of the page, on a grid the user never handed to `new Handsontable` and cannot reach except through the editor. Writing the setting back to `false` has the same problem: the object is rebuilt from `cellProperties.handsontable` on each open, so the write is discarded and re-attempted every time.

The precedent that fits is core's own treatment of a non-root instance, which is silent throughout: no accessibility announcer, no license notification or branding (`core.ts:1653`), no `LayoutManager` or `FocusScopeManager` (`:6663-6670`), no `registerAllFocusScopes` (`:6757`). The five options now say "main grid only" in the API reference, which is where a user configuring this will look, and the option stays visible in `getSettings()` rather than being mutated behind their back.

Scope notes:

- **Not a breaking change.** No default setting value changes, and no working behavior is removed: every one of these configurations throws today, so nothing can depend on it.
- Two alternatives were rejected. Making the nested dialog actually work is a feature (it needs an overlay container for nested grids plus a decision on which grid owns focus and Escape). Relaxing `getFocusScopeManager()` to return the root's manager changes a public accessor's contract, breaks the existing spec at `src/__tests__/core/getFocusScopeManager.spec.js`, and would register a focus trap on a detached element, replacing a loud crash with silent broken focus.
- `pagination.ts`'s two existing `isRootInstance` guards around `getLayoutManager()` are kept and both now carry a comment: with `isEnabled()` gated, that half is always false in practice and stays as a statement of the requirement, not as support for a nested grid (a direct `enablePlugin()` call on a non-root instance dies earlier, in the UI, which reads `rootGridElement`). Nested pagination never worked, including with a custom `uiContainer`: `PaginationUI.install()` skips the `#rootElement.querySelector` in that case, but `getContainer()` still returns an element, so `#registerFocusScope()` runs and throws.
- `#unregisterFocusScope()` in `dialog`, `pagination`, and `emptyDataState` now returns early on a non-root instance. Nothing was registered there, but the method reached `getFocusScopeManager()` unguarded, so a direct `nested.getPlugin('dialog').disablePlugin()` still threw. `notification` was already covered by its `#focusScopeActive` flag.
- Gating `loading` also stops it mutating the nested grid's `dialog` setting to `true` in `enablePlugin()`, a write that could no longer have any effect.
- The five option JSDoc blocks in `metaSchema.ts` now state that the plugin is main-grid only and link the three cell-type guides, so the API reference says it too. The guides carry it as well: the Dialog, Loading, and Notification overviews, the pagination guide's "Known limitations" list, and the `handsontable` cell-type guide, which lists the five options that do not apply inside the nested grid. There is no `emptyDataState` guide to update.

Found while working on this, real but out of scope, not fixed here:

1. `dialog.ts` calls `super.disablePlugin()` last, against the order documented in `handsontable/AGENTS.md`.
2. `dialog.ts`'s `destroy()` never unregisters its focus scope, and `FocusScopeManager.destroy()` only unmounts its event listener, so a root-instance scope leaks.
3. `core/types.ts` types `getFocusScopeManager()` and `getLayoutManager()` as always returning, expressing neither the throw nor the null.

### How has this been tested?

Failing-test-first: the new spec blocks were written and run against unmodified source, where they produced `Error: The FocusScopeManager is only available for the main Handsontable instance.` at `#registerFocusScope -> enablePlugin` and aborted the run. The same specs pass with the fix.

Each block asserts that the nested editor opens, that `enabled`/`isEnabled()` are `false` in the nested grid, and that an `updateSettings()` carrying the plugin's own key does not enable it either (the editor's own width/height update does not reach that branch of `BasePlugin#onUpdateSettings`). The `loading` block also asserts the nested `dialog` setting stays `false`. All five have a positive counterpart pinning the root side: the root dialog shows and renders into `ht-overlay`, a root grid with `pageSize: 8` still pages while the nested grid keeps all 10 rows renderable, the root notification still mounts a toast, the root empty data state still renders, and the root loading indicator still shows. `dialog` additionally covers the `autocomplete` nesting path and a direct `disablePlugin()` call on a nested instance. That no error is thrown is asserted implicitly but strictly, because one uncaught `pageerror` fails the whole Jasmine run, which is exactly how the red run was observed.

```bash
# the five plugin spec trees (551 specs, 0 failures)
npm run test:e2e --prefix handsontable --testPathPattern='plugins/(dialog|notification|emptyDataState|pagination|loading)/'

# collateral: the five plugin trees, the three editors that build non-root grids,
# and the accessor's contract spec (786 specs, 0 failures)
npm run test:e2e --prefix handsontable --testPathPattern='(plugins/(dialog|notification|emptyDataState|pagination|loading)/|editors/(handsontableEditor|autocompleteEditor|dropdownEditor)/|__tests__/core/getFocusScopeManager)'

npm run test:unit --prefix handsontable    # 325 suites, 3963 tests, 0 failures
npm run test:types --prefix handsontable   # clean
npm run eslint --prefix handsontable       # 0 errors
npm run stylelint --prefix handsontable    # clean
```

Manual, in a real browser, released 18.0.0 versus this build, one grid per plugin plus a root grid with `dialog: true` whose nested grid also asks for one:

- Released: every nested editor fails to open, with the errors quoted above, for the `handsontable`, `autocomplete`, and `dropdown` cell types alike.
- This build: every nested editor opens, no error is raised, the five plugins report `enabled === false` inside the nested grid, the nested `dialog` setting is untouched, and the root grid's own dialog and loading indicator still enable, show, and render into `ht-overlay`.

### Types of changes

### Related issue(s):

1. DEV-2641

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2641

---

> [!NOTE]
> **Low Risk**
> Bug fix that narrows plugin enablement to root instances; nested configs that already threw are unchanged in practice, with added tests and documentation.
> 
> **Overview**
> Fixes crashes when **`dialog`**, **`notification`**, **`emptyDataState`**, **`pagination`**, or **`loading`** are set on the inner grid config used by **`handsontable`**, **`autocomplete`**, or **`dropdown`** editors. Those plugins need the main instance’s overlay, layout slot, and focus scope; nested editor instances never have them, so enabling them used to throw during editor open.
> 
> **`isEnabled()`** on all five plugins now requires **`isRootInstance(this.hot)`**, so nested grids skip enable silently instead of crashing. Root grids behave as before. **`loading`** no longer forces **`dialog: true`** on a nested instance. **`pagination`**’s **`setPage`** / **`setPageSize`** no-op when the plugin is disabled, avoiding spurious page hooks. **`#unregisterFocusScope()`** on dialog, empty data state, and pagination guards non-root **`disablePlugin()`** calls.
> 
> Docs and **`metaSchema`** state these options apply to the **main grid only**. Changelog entry **#13265** and nested-grid e2e specs cover the regression.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 86a7d3f3d4bf00c4c043187db9bf0a2b5d0a26dd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2641 Dialog plugin throws on a nested grid: getFocusScopeManager() is root-only](https://app.clickup.com/t/9015210959/DEV-2641)

handsontable/src/dataMap/metaManager/metaSchema.ts
```diff
@@ -2195,6 +2195,9 @@ export default (): Record<string, unknown> => {
      * This option can only be set at the [grid level](@/guides/getting-started/configuration-options/configuration-options.md#set-grid-options).
      * It has no effect when set in the [`columns`](#columns), [`cells`](#cells), or [`cell`](#cell) options.
      *
+     * The dialog is available on the main grid only. In a grid nested in a cell of the `handsontable`
```
Haxikowy: The gate is `isRootInstance`, which is false for every non-root grid, not only the `handsontable` cell type. `AutocompleteEditor` and `DropdownEditor` extend `HandsontableEditor`, and `prepare()` merges `cellProperties.handsontable` (handsontableEditor.ts:177), so `{ type: 'autocomplete', handsontable: { pagination: {...} } }` reaches the same nested grid. A user with that config reads this note and concludes it doesn't apply to them.

Suggest wording it as "in a grid nested in a cell (the `handsontable`, `autocomplete`, or `dropdown` cell type), this option has no effect", in all five blocks. Worth linking the cell-type guide too, the way the neighbouring lines link `columns` and `cells`.

handsontable/src/plugins/dialog/dialog.ts
```diff
@@ -258,10 +259,14 @@ export class Dialog extends BasePlugin {
...
    *
+   * The dialog renders into the `ht-overlay` element and registers a modal focus scope, and both
+   * belong to the main Handsontable instance. In a nested grid (the one that the `handsontable` cell
+   * type creates) neither exists, so the plugin stays disabled there.
+   *
    * @returns {boolean}
    */
   isEnabled(): boolean {
-    return !!this.hot.getSettings()[PLUGIN_KEY];
+    return isRootInstance(this.hot) && !!this.hot.getSettings()[PLUGIN_KEY];
```
Haxikowy, line 270: The refusal is silent, and this repo is loud about the comparable case. `isHardConflictBlocked()` warns "The `pagination` plugin cannot be used with the `fixedRowsTop` option. This combination is not supported. The plugin will remain disabled." and sets the setting to `false` (base.ts:251, asserted in pagination/__tests__/conflictingOptions.spec.js:13).

Two refusals for the same kind of reason now behave differently, and the new one is the harder to debug: the option is still in `getSettings()`, `enabled` is `false`, and the console says nothing. The current crash at least points at itself. Either route this through the same warn, or say in the description why silence is deliberate here. Applies to all five plugins.

handsontable/src/plugins/pagination/pagination.ts
```diff
@@ -252,7 +259,9 @@ export class Pagination extends BasePlugin {
 
     // The layout manager owns the bottom-slot placement and ordering. With a custom `uiContainer`
     // the UI installs itself there instead, so the slot registration is skipped. The manager only
-    // exists on the root instance, hence the guard.
+    // exists on the root instance. That half of the guard is now unreachable, since `isEnabled()`
```
Haxikowy: "Unreachable" isn't right. `enablePlugin()` is public on `BasePlugin`, so `nested.getPlugin('pagination').enablePlugin()` evaluates this guard directly; it's `isEnabled()` that never runs on that path, not the guard.

The guard also can't protect anything: line 286 `#registerFocusScope()` throws first through the unguarded `getFocusScopeManager()` at :881. Suggest "always false in practice" and dropping the unreachability claim. The matching guard in `disablePlugin()` at :404 sits behind the same throw and got no comment, which reads as asymmetric.

.changelogs/13265.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Fixed a crash when the `dialog`, `notification`, `emptyDataState`, `pagination`, or `loading` option was enabled in a grid nested in a cell of the `handsontable` cell type.",
```
Haxikowy: The title stops at the crash, so it reads as "nested dialogs work now". They don't, they're ignored. Suggest appending: "These options now apply to the main grid only." Type stays `fixed`.

handsontable/src/dataMap/metaManager/metaSchema.ts
demtario: Correct, and the note was wrong in a way I could confirm on the released build. I checked `{ type: 'autocomplete', handsontable: { dialog: true } }` and the `dropdown` equivalent against 18.0.0: both throw the same `The FocusScopeManager is only available for the main Handsontable instance.`, exactly as your reading of `prepare()` predicts.

Fixed in ff3aa3b. All five notes now read "In a grid nested in a cell, that is a cell of the [`handsontable`](...), [`autocomplete`](...), or [`dropdown`](...) cell type, this option has no effect", with the three cell-type guides linked the way the neighbouring lines link `columns` and `cells`. The same widening went into the five plugins' own `isEnabled()` JSDoc, which had the identical over-narrow phrasing, and into the changelog title.

I also added a spec for the `autocomplete` path (`dialog.spec.js`), so the claim that all three cell types share the passthrough is pinned by a test rather than by prose alone.

handsontable/src/plugins/dialog/dialog.ts
demtario: Taking the second of your two options: the silence is deliberate, and ff3aa3b writes the reason into the PR description rather than leaving it implied.

The short version of why this refusal is not shaped like `isHardConflictBlocked()`. A hard conflict is a property of one grid the user configured, so warning once per grid and writing the setting back to `false` both land somewhere the user can act on. A nested grid is constructed fresh on **every** editor open (`handsontableEditor.ts:92`) and destroyed on close (`:81`), so the same configuration is re-evaluated on every cell edit: a warn would repeat for the lifetime of the page, on an instance the user never passed to `new Handsontable` and cannot reach except through the editor. Writing the setting back to `false` fails for the same reason, since the options object is rebuilt from `cellProperties.handsontable` on each open, so the write is discarded and re-attempted every time.

The precedent I followed instead is core's own treatment of a non-root instance, which is silent throughout: no accessibility announcer, no license notification or branding (`core.ts:1653`), no `LayoutManager` or `FocusScopeManager` (`:6663-6670`), no `registerAllFocusScopes` (`:6757`).

On debuggability, which is the real cost you are naming: the five options now carry a "main grid only" note in the API reference, which is where someone writing `handsontable: { dialog: true }` will look, and unlike the hard-conflict path the option stays visible in `getSettings()` rather than being mutated behind the user's back. If you would still rather have the warn, say so and I will route all five through it, once per instance.

handsontable/src/plugins/pagination/pagination.ts
demtario: You are right on both counts, and fixed in ff3aa3b.

"Unreachable" was the wrong word: `enablePlugin()` is public, so a direct `nested.getPlugin('pagination').enablePlugin()` does evaluate this guard. It is `isEnabled()` that is bypassed on that path. The comment now says "always false in practice".

On what dies first, one correction to your version: it is not `#registerFocusScope()`. That call sits *after* this guard in `enablePlugin()`, and `#updateSectionsVisibilityState()` runs *before* it, reaching `ui.ts`'s `#rootElement.querySelector('.ht-wrapper')` on the `null` `rootGridElement` that line 240 passed in. So a direct call on a non-root instance dies in the UI, before either the guard or the focus scope. The comment names that path now, so the next reader does not go looking at the focus scope.

The asymmetry is closed too: the matching guard in `disablePlugin()` carries a one-line pointer back to this comment. I kept both guards rather than deleting them, since they cost nothing and they state the requirement at the point where it matters.

.changelogs/13265.json
demtario: Agreed, and taken in ff3aa3b. The title now reads:

> Fixed a crash when the `dialog`, `notification`, `emptyDataState`, `pagination`, or `loading` option was enabled in a grid nested in a cell of the `handsontable`, `autocomplete`, or `dropdown` cell type. These options now apply to the main grid only.

Your sentence appended verbatim, plus the two other cell types from your first comment, since they are affected identically. Type stays `fixed`.

Haxikowy:
Four findings from my review that had no line in the diff to attach to. GitHub dropped the review body when I submitted, so they never posted.

1. The guides are untouched; only the API reference carries the limitation. `docs/content/guides/rows/rows-pagination/rows-pagination.md` has a "Known limitations" list, and the three cell-type guides now linked from `metaSchema.ts` have no limitations section. Monorepo checklist item 2 asks for guides alongside the JSDoc.

2. `#unregisterFocusScope()` calls `getFocusScopeManager()` unguarded in `dialog.ts:658`, `pagination.ts:882` and `emptyDataState.ts:453`, so a manual `nested.getPlugin('dialog').disablePlugin()` still throws `The FocusScopeManager is only available for the main Handsontable instance.` Notification is guarded by `#focusScopeActive` (`notification.ts:1017`). The automatic path is fixed; the direct-call path is not.

3. The paired "root still works while a nested grid asks too" spec exists for dialog and pagination only. notification, emptyDataState and loading have the nested case without it.

4. The collateral test command in the description omits `plugins/loading/`. The first command covers it, so the 747-spec figure just does not include it.

handsontable/src/plugins/pagination/pagination.ts
```diff
@@ -195,10 +195,18 @@ export class Pagination extends BasePlugin {
...
+   * `dropdown` cell types create) none of them exists, so the plugin stays disabled there. A custom
+   * `uiContainer` does not change that:
+   * it replaces the bottom-slot placement only, while the focus scope and the root grid element are
+   * still required.
+   *
    * @returns {boolean}
    */
   isEnabled(): boolean {
-    return !!this.hot.getSettings()[PLUGIN_KEY];
+    return isRootInstance(this.hot) && !!this.hot.getSettings()[PLUGIN_KEY];
```
budnix, line 209: A grid that never set `pagination` could already reach this state, but now a nested grid that explicitly asks for `pagination` lands there too, and none of the public paging methods check `this.enabled`. `setPage(3)` still fires `beforePageChange`/`afterPageChange`, calls `hot.render()`, and scrolls back to row 0, while `#currentPage` clamps to 1 because `#calcStrategy` is still `null` — so the caller gets a page-change event for a page change that never happened. `resetPage`, `setPageSize`, and `resetPageSize` all go through the same path. Notification and Dialog already open their public methods with `if (!this.enabled) return;`, and the same one-liner here would keep the no-op quiet.

handsontable/src/dataMap/metaManager/metaSchema.ts
```diff
@@ -4960,6 +4986,12 @@ export default (): Record<string, unknown> => {
      * This option can only be set at the [grid level](@/guides/getting-started/configuration-options/configuration-options.md#set-grid-options).
      * It has no effect when set in the [`columns`](#columns), [`cells`](#cells), or [`cell`](#cell) options.
      *
+     * Pagination is available on the main grid only. In a grid nested in a cell, that is a cell of the
```
budnix, line 4989: This note lands in the API reference, but the guide linked a few lines above — `rows-pagination.md` — says nothing about the limit, and it already has a "Known limitations" section (line 463) listing the `NestedRows` and `MergeCells` restrictions. Someone who reads the guide instead of the option reference gets no warning, and the dialog, notification, loading, and empty data state guides have the same gap. The repo checklist asks for guides in the same change as the API reference — was this meant as a follow-up commit?

handsontable/src/plugins/emptyDataState/__tests__/emptyDataState.spec.js
```diff
@@ -280,4 +280,38 @@ describe('EmptyDataState', () => {
...
   });
+
+  describe('nested grid (non-root instance)', () => {
+    it('should not enable the plugin in a grid nested in the `handsontable` cell type', async() => {
+      handsontable({
+        data: createSpreadsheetData(2, 2),
+        columns: [{
+          type: 'handsontable',
+          handsontable: {
+            data: createSpreadsheetData(2, 2),
```
budnix, line 291: The nested grid gets two rows here, so it never enters the empty state this plugin is actually about. The gate itself is pinned fine by the `isEnabled()` assertion, so this isn't a hole — but `data: []` is the config a user hits when they set `emptyDataState` on a nested grid that has no rows, and nothing covers it right now. A second short case would be enough.

budnix (commented):
Three points that the latest commit and the open threads do not cover yet: one behavior gap in the pagination public API, one missing docs update, and one test-coverage suggestion.

demtario:
All four taken, in 99f5391. Thanks for reposting them after GitHub ate the review body.

**1. Guides.** Done, five files. The Dialog, Loading, and Notification overviews each gained a paragraph, the pagination guide gained a paragraph under "Known limitations" (including the note that a custom `uiContainer` does not change it), and the `handsontable` cell-type guide now lists the five options that have no effect inside the nested grid, pointing at the main grid instead and noting that the `autocomplete` and `dropdown` nested grids behave the same. One gap worth naming: there is no `emptyDataState` guide in `docs/content/guides/`, only changelog mentions, so the API reference is the only place that option is documented.

**2. `#unregisterFocusScope()`.** Confirmed and fixed in `dialog.ts`, `pagination.ts` and `emptyDataState.ts`: each now returns early on a non-root instance, so a direct `nested.getPlugin('dialog').disablePlugin()` no longer reaches the accessor. `notification.ts` left alone, since `#focusScopeActive` already covers it, as you noted. A spec in `dialog.spec.js` pins the direct-call path.

Worth flagging what this does **not** fix, because it is pre-existing and out of scope: on a *root* instance where the plugin never enabled, a manual `disablePlugin()` still throws `Scope with id "dialog" not found` from `unregisterScope`. That is the `FocusScopeManager`'s own contract and it predates this PR, so I left it.

**3. Positive counterparts.** Added for notification (root toast still mounts), emptyDataState (root state still renders) and loading (root indicator still shows), so all five now have both directions.

**4. Test command.** Corrected in the description: the collateral pattern now includes `plugins/loading/`, and the figures are re-measured on the current head, 551 specs for the five plugin trees and 786 for the collateral set, both with 0 failures.

docs/content/guides/dialog/dialog/dialog.md
```diff
@@ -31,6 +31,8 @@ The Dialog plugin provides a modal dialog system for Handsontable that allows yo
 
 The dialog system is designed to be flexible and customizable, supporting various content types including plain text, HTML, and DOM elements. It also provides options for styling, animations, and user interaction controls.
 
+The plugin works on the main grid only. A grid nested in a cell, that is a cell of the [`handsontable`](@/guides/cell-types/handsontable-cell-type/handsontable-cell-type.md), [`autocomplete`](@/guides/cell-types/autocomplete-cell-type/autocomplete-cell-type.md), or [`dropdown`](@/guides/cell-types/dropdown-cell-type/dropdown-cell-type.md) cell type, has no overlay layer to render a dialog into, so the [`dialog`](@/api/options.md#dialog) option has no effect there.
```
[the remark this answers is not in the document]
demtario: Correct, and fixed in 9ababbf. `menuTag: updated` added to `dialog.md`, `notification.md` and `rows-pagination.md`.

Confirming the diagnosis rather than taking it on faith: `loading.md:23` and `handsontable-cell-type.md:16` did already carry the tag, but not from this branch. `git diff origin/develop...HEAD` on both files shows only the paragraph I added, so those two were tagged by earlier work and the three untagged ones were the inconsistency.

On whether the tag is warranted at all, since `docs/AGENTS.md:415` says to omit it for trivial fixes: these pages now document a user-facing behavior change (the five options apply to the main grid only, where they previously threw), which is a substantive change rather than a typo or link correction, so the badge belongs.

handsontable/src/plugins/pagination/pagination.ts
demtario: Confirmed in a browser before changing anything, and it reproduces exactly as you describe. On a nested grid with `pagination: { pageSize: 5 }`:

```
beforePageChange 1->3
afterPageChange 1->1        <- a page change that never happened
afterPageSizeChange 10->10
```

Fixed in 86a7d3f: `setPage()` and `setPageSize()` now open with `if (!this.enabled) { return; }`. Those two are the funnels, so `resetPage`, `resetPageSize`, `nextPage`, `prevPage`, `firstPage`, `lastPage` and `resetPagination` are all covered through them. Re-checked in the browser afterwards: the nested grid emits nothing, and a root grid with `pageSize: 8` still pages normally (`afterPageChange 1->2`, renderable length 8). A spec in `pagination.spec.js` pins it with spies on both hooks.

One related thing I did **not** change, because it widens the surface past what you flagged and it is not specific to a nested grid. The six section-visibility methods (`showPageSizeSection`, `hidePageSizeSection`, and the counter and navigation pairs) have the same shape: `this.#ui?.setPageSizeSectionVisibility(true)` no-ops while `afterPageSizeVisibilityChange` still fires, on any grid where the plugin is disabled. Happy to add the same one-liner to all six here, or file it separately - your call.

handsontable/src/dataMap/metaManager/metaSchema.ts
demtario: Not a follow-up, it landed in 99f5391 about ten minutes before you posted, so you were reading `ff3aa3b`. The guides now carry it:

- `rows-pagination.md`, in the "Known limitations" section you point at, including the note that a custom `uiContainer` does not change it.
- `dialog/dialog/dialog.md`, `dialog/loading/loading.md` and `dialog/notification/notification.md`, each in the Overview.
- `cell-types/handsontable-cell-type/handsontable-cell-type.md`, which lists all five options that have no effect inside the nested grid and says to set them on the main grid, plus a note that the `autocomplete` and `dropdown` nested grids behave the same.

`menuTag: updated` was missing on three of those pages; added in 9ababbf after Bugbot flagged it.

One genuine gap left, which is not something this PR introduced: there is no `emptyDataState` guide in `docs/content/guides/` at all, only changelog mentions, so for that option the API reference is the only documentation. Worth its own task if you want the set complete.

handsontable/src/plugins/emptyDataState/__tests__/emptyDataState.spec.js
demtario: Fair, and added in 86a7d3f as a second case rather than by changing this one: same nested setup with `data: []`, asserting the editor opens, `countRows()` is `0`, the plugin stays disabled, and no `.ht-empty-data-state` element exists in the nested grid's root. Keeping both means the two-row case still covers the plain "asked for it, did not get it" path while the empty case covers the configuration a user actually hits.

budnix (approved):

Haxikowy (approved):

PR #13266, opened 2026-08-27
DEV-2654: Keep the space between alignment class names (#7122)

### LINKED ISSUE #7122: Set vertical and horizontal alignment: space between the class names is lost
### Description
If select any vertical alignment, then any horizontal alignment, then another vertical alignment, the space between the class names is lost

### Steps to reproduce
1. Select cell with defined class name (e.g. 'class_name')
2. Select any vertical alignment from context menu (e.g. 'htRight')
3. Select any horizontal alignment (e.g. 'htMiddle')
4. Select any other vertical alignment (e.g. 'htJustify')

### Expected behaviour 
'class_name htMiddle htJustify'

### Currently behaviour
'class_namehtMiddle htJustify'

### Demo
https://jsfiddle.net/1xuzg73c/3/

### Your environment
* Handsontable version: 7.4.4
* Browser Name and version:
* Operating System:

budnix:
### Context

The PR fixes [#7122](https://github.com/handsontable/handsontable/issues/7122) — open since July 2020, and present in the code since 2015.

Picking a second alignment on the same axis destroyed the cell's other class names. On a cell with a custom class `class_name`:

| Step | Action | `className` after |
|---|---|---|
| 1 | Alignment → Right | `class_name htRight` |
| 2 | Alignment → Middle | `class_name htRight htMiddle` |
| 3 | Alignment → Justify | `class_namehtMiddle htJustify` |

After step 3 the space between `class_name` and `htMiddle` was gone. The browser read one class called `class_namehtMiddle`, which matches nothing, so the user silently lost **two** things: their own class, and the vertical alignment they had just set.

#### Cause

`prepareVerticalAlignClass` and `prepareHorizontalAlignClass` in `handsontable/src/plugins/contextMenu/utils.ts` did string surgery on the class name:

```js
className
  .replace('htLeft', '')
  .replace('htCenter', '')
  .replace('htRight', '')
  .replace('htJustify', '')
  .replace('  ', '');   // deleted the double space instead of collapsing it
```

Removing a token from the middle of the string leaves two spaces. The last line replaced those two spaces with nothing, pushing the neighbours together.

This is also why the bug looked unreproducible to some reviewers on the 2020 thread: gluing only happens when the removed token sits **between** two other tokens. If it is last, you get a harmless stray double space and nothing visibly breaks — so the click order decided whether the bug showed.

The same substring matching had a second effect: a custom class that merely contained an alignment name was chopped up. `htTopBar` + Bottom returned `Bar htBottom`, and the alignment was not applied at all.

#### Fix

Both helpers now delegate to a shared `prepareAlignClass()` that normalizes the value with `normalizeClassNames()` and works on whole tokens — drop the tokens belonging to the axis being changed, append the new one, join with a single space. This matches how `exportFile/types/xlsx/cell-style.ts` already matches alignment classes.

Three side effects worth calling out in review:

1. A custom class containing an alignment name (`htTopBar`) is now preserved **and** the alignment applies. Previously picking that axis was a silent no-op — the old behavior was the defect.
2. An array `className` — documented as supported in `metaSchema` — no longer throws in the alignment path. Same family as #7427.
3. XLSX export alignment self-repairs for cells whose classes were previously glued, since `exportFile` matches by exact token.

No public API, option, default value, or CSS class name changed.

### How has this been tested?

Unit tests (15 new cases) and E2E tests (3 new cases) were written first, confirmed failing against the old code, then confirmed passing after the fix.

```bash
# 15 new unit cases — both axes, the reported sequence, idempotence,
# spacing, substring safety, array className
npm run test:unit --prefix handsontable -- --testPathPattern=contextMenu/__tests__/utils

# 3 new E2E cases driving the real context menu, plus the existing alignment suite
npm run test:e2e --prefix handsontable -- --testPathPattern=alignment

# regression checks on the consumers
npm run test:e2e --prefix handsontable -- --testPathPattern=cellAlignment
npm run test:e2e --prefix handsontable -- --testPathPattern="contextMenu|dropdownMenu"
npm run test:e2e --prefix handsontable -- --testPathPattern=exportFile

# full gates
```

Results: unit 3978/3978 pass (325 suites). E2E — alignment 49/49, cellAlignment 14/14, contextMenu+dropdownMenu 786/786, exportFile 272/272 (1 pre-existing pending). ESLint 0 errors, Stylelint clean.

Manual check in Chrome, A/B against the released build. On 18.0.0 the reported three steps drive `class_name` off the cell and its background styling disappears; on this branch all three classes survive and the styling holds. Verified both through real right-click context-menu clicks and through `getPlugin('contextMenu').executeCommand(...)`.

### Types of changes

### Related issue(s):

1. DEV-2654
2. #7122

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2654

---

> [!NOTE]
> **Low Risk**
> Localized bug fix to cell `className` updates and alignment undo with no API changes; behavior change is restoring correct styling rather than new surface area.
> 
> **Overview**
> Fixes **#7122**: context-menu and dropdown-menu alignment no longer corrupt `className` when users pick another alignment on the same axis or mix custom classes with alignment tokens.
> 
> **Alignment helpers** in `contextMenu/utils.ts` stop using substring `.replace()` on the raw string. They normalize with `normalizeClassNames()`, remove only whole horizontal/vertical alignment tokens, append the new one, and join with a single space. That prevents glued names like `class_namehtMiddle`, preserves classes such as `htTopBar`, and supports array `className`. `getAlignmentComparatorByClass` now matches tokens exactly instead of `indexOf` on the full string.
> 
> **Undo** for cell alignment in `undoRedo/actions/cellAlignment.ts` restores the recorded `className` (including `undefined`) instead of falling back to `' htLeft'`, so undo does not leave cells left-aligned or grow class strings on undo/redo cycles.
> 
> Adds changelog entry, **AGENTS.md** guidance on `className` handling, broad unit/E2E coverage, and test helpers `closeDropdownMenu` / `selectDropdownSubmenuOption` for dropdown alignment tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4e6711031f3e213ea0877e67198fbc62f45d727f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/handsontable/src/plugins/contextMenu/utils.ts
+++ b/handsontable/src/plugins/contextMenu/utils.ts
@@ -30,7 +30,7 @@ function prepareAlignClass(
   axisClassNames: string[]
 ): string {
   const classNames = normalizeClassNames(className)
-    .filter(name => !axisClassNames.includes(name));
+    .filter(name => name !== alignment && !axisClassNames.includes(name));
 
   return [...classNames, alignment].join(' ');
 }
--- a/handsontable/src/plugins/contextMenu/utils.ts
+++ b/handsontable/src/plugins/contextMenu/utils.ts
@@ -12,40 +12,44 @@ const HORIZONTAL_ALIGNMENT_CLASS_NAMES = ['htLeft', 'htCenter', 'htRight', 'htJu
 /**
  * Swaps the alignment class of one axis, leaving every other class name untouched.
  *
- * The class name is compared token by token. Matching substrings is not enough - it both destroys
+ * The class name is compared token by token. Matching substrings is not enough – it both destroys
  * custom classes that merely contain an alignment name (`htTopBar`) and, once a token is cut out of
- * the middle of the string, glues its two neighbours together (#7122).
+ * the middle of the string, glues its two neighbors together (#7122).
  *
- * @param {string|string[]} className The full element class name to process.
+ * Every class name of the axis is dropped, including the one being applied. Returning early when the
+ * picked alignment is already there would leave a competing class name of the same axis in place.
+ *
+ * @param {string|string[]|undefined} className The full element class name to process.
  * @param {string} alignment The alignment class name to apply.
  * @param {string[]} axisClassNames The alignment class names of the axis being changed.
  * @returns {string}
  */
-function prepareAlignClass(className: string | string[], alignment: string, axisClassNames: string[]): string {
-  const classNames = normalizeClassNames(className);
+function prepareAlignClass(
+  className: string | string[] | undefined,
+  alignment: string,
+  axisClassNames: string[]
+): string {
+  const classNames = normalizeClassNames(className)
+    .filter(name => !axisClassNames.includes(name));
 
-  if (classNames.includes(alignment)) {
-    return classNames.join(' ');
-  }
-
-  return [...classNames.filter(name => !axisClassNames.includes(name)), alignment].join(' ');
+  return [...classNames, alignment].join(' ');
 }
 
 /**
- * @param {string|string[]} className The full element class name to process.
+ * @param {string|string[]|undefined} className The full element class name to process.
  * @param {string} alignment The alignment class name to compare with.
  * @returns {string}
  */
-export function prepareVerticalAlignClass(className: string | string[], alignment: string) {
+export function prepareVerticalAlignClass(className: string | string[] | undefined, alignment: string) {
   return prepareAlignClass(className, alignment, VERTICAL_ALIGNMENT_CLASS_NAMES);
 }
 
 /**
- * @param {string|string[]} className The full element class name to process.
+ * @param {string|string[]|undefined} className The full element class name to process.
  * @param {string} alignment The alignment class name to compare with.
  * @returns {string}
  */
-export function prepareHorizontalAlignClass(className: string | string[], alignment: string) {
+export function prepareHorizontalAlignClass(className: string | string[] | undefined, alignment: string) {
   return prepareAlignClass(className, alignment, HORIZONTAL_ALIGNMENT_CLASS_NAMES);
 }
 
@@ -113,16 +117,10 @@ function applyAlignClassName(
   cellDescriptor: (row: number, col: number) => Record<string, unknown>,
   propertySetter: (row: number, col: number, key: string, value: string) => void
 ) {
-  const cellMeta = cellDescriptor(row, col);
-  let className = alignment;
-
-  if (cellMeta.className) {
-    if (type === 'vertical') {
-      className = prepareVerticalAlignClass(cellMeta.className as string | string[], alignment);
-    } else {
-      className = prepareHorizontalAlignClass(cellMeta.className as string | string[], alignment);
-    }
-  }
+  const { className: currentClassName } = cellDescriptor(row, col) as { className?: string | string[] };
+  const className = type === 'vertical' ?
+    prepareVerticalAlignClass(currentClassName, alignment) :
+    prepareHorizontalAlignClass(currentClassName, alignment);
 
   propertySetter(row, col, 'className', className);
 }
@@ -196,6 +194,7 @@ export function getAlignmentComparatorByClass(htClassName: string) {
   return function(this: HotInstance, row: number, col: number): boolean {
     const className = this.getCellMetaTransient(row, col).className as string | string[] | undefined;
 
-    return Boolean(className && className.indexOf(htClassName) !== -1);
+    // Compared token by token. A substring match would report `htLeftPanel` as being aligned left.
+    return normalizeClassNames(className).includes(htClassName);
   };
 }
--- a/handsontable/src/plugins/contextMenu/utils.ts
+++ b/handsontable/src/plugins/contextMenu/utils.ts
@@ -1,46 +1,52 @@
 import type { HotInstance } from '../../core/types';
 import { arrayEach } from '../../helpers/array';
+import { normalizeClassNames } from '../../helpers/dom/element';
 
 interface CellRangeLike {
   forAll(callback: (row: number, col: number) => void | boolean): void;
 }
 
+const VERTICAL_ALIGNMENT_CLASS_NAMES = ['htTop', 'htMiddle', 'htBottom'];
+const HORIZONTAL_ALIGNMENT_CLASS_NAMES = ['htLeft', 'htCenter', 'htRight', 'htJustify'];
+
 /**
- * @param {string} className The full element class name to process.
- * @param {string} alignment The alignment class name to compare with.
+ * Swaps the alignment class of one axis, leaving every other class name untouched.
+ *
+ * The class name is compared token by token. Matching substrings is not enough - it both destroys
+ * custom classes that merely contain an alignment name (`htTopBar`) and, once a token is cut out of
+ * the middle of the string, glues its two neighbours together (#7122).
+ *
+ * @param {string|string[]} className The full element class name to process.
+ * @param {string} alignment The alignment class name to apply.
+ * @param {string[]} axisClassNames The alignment class names of the axis being changed.
  * @returns {string}
  */
-export function prepareVerticalAlignClass(className: string, alignment: string) {
-  if (className.indexOf(alignment) !== -1) {
-    return className;
+function prepareAlignClass(className: string | string[], alignment: string, axisClassNames: string[]): string {
+  const classNames = normalizeClassNames(className);
+
+  if (classNames.includes(alignment)) {
+    return classNames.join(' ');
   }
 
-  const replacedClassName = className
-    .replace('htTop', '')
-    .replace('htMiddle', '')
-    .replace('htBottom', '')
-    .replace('  ', '');
-
-  return `${replacedClassName} ${alignment}`;
+  return [...classNames.filter(name => !axisClassNames.includes(name)), alignment].join(' ');
 }
 
 /**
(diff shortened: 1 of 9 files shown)


qunabu:
Task linked: [DEV-2654 Fix alignment class names losing the separating space (#7122)](https://app.clickup.com/t/9015210959/DEV-2654)

handsontable/src/plugins/undoRedo/actions/cellAlignment.ts
```diff
@@ -61,7 +61,10 @@ export class CellAlignmentAction extends BaseAction {
       range.forAll((row: number, col: number) => {
         // Alignment classes should only collected within cell ranges. We skip header coordinates.
         if (row >= 0 && col >= 0) {
-          hot.setCellMeta(row, col, 'className', (this.stateBefore as string[][])[row][col] || ' htLeft');
+          // Restored as recorded, including an absent value. Falling back to a horizontal alignment
+          // here used to leave the cell aligned left after undoing a vertical alignment, and made
+          // the class name grow on every undo/redo cycle.
+          hot.setCellMeta(row, col, 'className', (this.stateBefore as string[][])[row][col]);
```
demtario, line 67: **medium** - undo restores the value, but not the *level* it lived at.

`setCellMeta` always writes an **own** property (`metaLayers/cellMeta.ts:234` is `cellMeta[key] = value`, never `delete`). `stateBefore` is recorded from a **cascaded** read - `predefinedItems/alignment.ts:47` uses `getCellMetaTransient(row, col).className`. So a column-level or grid-level `className` comes back pinned at cell level and stops following later `updateSettings`. When the cascade was empty at record time, the write is an own `className: undefined`, which then shadows a column `className` added afterwards; `_persistedMetaProps` replays it across `updateSettings`.

Not a regression - the removed `|| ' htLeft'` pinned own meta too, with a worse value, and the line dates to 2015. Flagging it as an incomplete fix while this line is open.

To actually restore the cascade, the action would have to record whether the value was own and call `hot.removeCellMeta(row, col, 'className')` for the inherited case. Note that is not a free swap: `removeCellMeta` fires `beforeRemoveCellMeta`/`afterRemoveCellMeta`, `setCellMeta` does not.

handsontable/src/plugins/undoRedo/__tests__/actions/cellAlignment.spec.js
```diff
@@ -151,6 +151,38 @@ describe('UndoRedo -> CellAlignment action', () => {
...
+
+    expect(getCellMeta(0, 0).className).toBe('htMiddle');
+
+    getPlugin('undoRedo').undo();
+
+    await render();
+
+    // The undo used to fall back to ' htLeft' - a leading space plus a horizontal alignment the
+    // user never picked - so undoing left the cell aligned left instead of unaligned.
+    expect(getCellMeta(0, 0).className).toBeUndefined();
```
demtario, line 132: **medium** - this assertion cannot distinguish "property absent" from "own `className: undefined`", which is the exact distinction the fix turns on.

The fixture above declares no `columns` and no grid-level `className`, so the cascade is empty and `toBeUndefined()` passes for both shapes.

Add a case with `columns: [{ className: 'my-col' }]` and assert the cell still renders `my-col` after the undo. That assertion fails today, and it is what pins the behavior described in the `cellAlignment.ts` note.

handsontable/src/plugins/undoRedo/__tests__/actions/cellAlignment.spec.js
```diff
@@ -73,67 +73,67 @@ describe('UndoRedo -> CellAlignment action', () => {
...
     getPlugin('undoRedo').undo();
     cellMeta = getCellMeta(0, 7);
-    expect(cellMeta.className.indexOf('htCenter')).toBeGreaterThan(-1);
-    expect(cellMeta.className.indexOf('htBottom')).toEqual(-1);
+    expect(cellMeta.className).toContain('htCenter');
+    expect(cellMeta.className ?? '').not.toContain('htBottom');
 
     cellMeta = getCellMeta(5, 7);
-    expect(cellMeta.className.indexOf('htBottom')).toEqual(-1);
+    expect(cellMeta.className ?? '').not.toContain('htBottom');
```
demtario: **low** - the `?? ''` rewrite removed the only assertion that caught `className` being wiped entirely.

The old `expect(cellMeta.className.indexOf('htBottom')).toEqual(-1)` *threw* when `className` was `undefined`. The new form passes for `undefined` and for any string without the token. Lines 105-106 are a pure-negative site, so nothing else pins the value - the test can no longer tell "correctly unaligned" from "wrongly wiped", and the `cellAlignment.ts` change is what makes a wipe newly reachable.

Same shape at lines 103, 110, 116, 119, 123, 128, 129, 132, 135, 136, and mirrored in `dropdownMenu.spec.js`. At each site, assert the expected value directly - `expect(getCellMeta(5, 7).className).toBeUndefined()` or `.toBe('<expected>')` - the way the new test at line 174 already does.

handsontable/test/helpers/common.js
```diff
@@ -1084,6 +1091,29 @@ export function openDropdownSubmenuOption(submenuName, cell) {
...
+ * Open, execute the sub menu action and close the dropdown menu.
+ *
+ * @param {string} submenuName The dropdown menu item name (it has to be a submenu) to hover.
+ * @param {string} optionName The dropdown menu subitem name to click.
+ * @param {number|HTMLTableCellElement} [cell] The column index or TH element to open the menu under.
+ */
+export async function selectDropdownSubmenuOption(submenuName, optionName, cell) {
+  openDropdownSubmenuOption(submenuName, cell);
+
+  await sleep(300);
```
demtario: **low** - a fixed sleep in a shared helper, at a site the lint rule cannot see.

`handsontable/.ai/TESTING.md:280`: "Do **not** add `await sleep(100)` / `sleep(200)` to new tests... if you must, root-cause it and add a real waiter." The `handsontable/no-fixed-sleep-in-spec` rule is scoped to `*.unit.js` / `*.spec.js` (`.eslintrc.js:225`), so `test/helpers/common.js` is not covered and every future caller inherits the delay.

Second half: the `:contains()` lookup two lines down is unchecked, so an empty jQuery set makes each `.simulate()` a silent no-op and the test passes having clicked nothing.

Wait for `.htDropdownMenuSub_${submenuName}` to exist instead of the clock, and assert `button.length` before simulating.

Acknowledged counterweight: this mirrors `selectContextSubmenuOption` byte-for-byte, sleep included, so staying consistent with its sibling is a defensible call.

handsontable/src/plugins/contextMenu/utils.ts
```diff
@@ -1,46 +1,56 @@
...
-  if (className.indexOf(alignment) !== -1) {
-    return className;
-  }
+function prepareAlignClass(
+  className: string | string[] | undefined,
+  alignment: string,
+  axisClassNames: string[]
+): string {
+  const classNames = normalizeClassNames(className)
+    .filter(name => !axisClassNames.includes(name));
```
demtario, line 34: **low** - latent duplicate token. The filter drops only `axisClassNames`, then `alignment` is appended unconditionally, so a cross-axis call emits it twice: `prepareVerticalAlignClass('htLeft', 'htLeft')` returns `'htLeft htLeft'`.

Unreachable through `align()` today - every `type`/`alignment` pair is consistent at each call site in `predefinedItems/alignment.ts`, and the redo path replays the recorded pair. But both helpers are exported and now have direct unit tests, and the removed early return (`className.indexOf(alignment) !== -1`) is what used to mask it.

One clause: add `&& name !== alignment` to the filter, or `[...new Set([...classNames, alignment])]`.

demtario (commented):
5 inline notes below.

handsontable/src/plugins/contextMenu/utils.ts
budnix: Good spot — fixed in 4e6711031f, the filter now drops the incoming alignment too.

handsontable/test/helpers/common.js
budnix: Nice catch, that could have clicked nothing — fixed in 4e6711031f, it now waits for the submenu to appear and throws if the option is missing.

handsontable/src/plugins/undoRedo/__tests__/actions/cellAlignment.spec.js
budnix: Fair point, those went soft — every site now asserts the exact value in 4e6711031f, in both specs.

handsontable/src/plugins/undoRedo/actions/cellAlignment.ts
budnix: You're right, and I confirmed it: a column-level className stops following updateSettings after undo. Leaving it out of this PR though — the fix needs removeCellMeta, whose beforeRemoveCellMeta is vetoable, so it deserves its own task.

handsontable/src/plugins/undoRedo/__tests__/actions/cellAlignment.spec.js
budnix: Agreed — that test only passes once the level restore is fixed, so it goes with that task rather than landing red here.

demtario (approved):

Issue #11832, opened 2025-09-11
In the _onAfterChange internal method of the filter plugin, there is a conversion issue between physical columns and displayed columns.


wuzaofeng:
### Describe the bug

In the _onAfterChange method of filters.mjs, this.hot.propToCol returns the displayed column index.

However, the current method this.conditionCollection.hasConditions requires a physical column index.

### Video/Screenshots

<img width="1203" height="618" alt="Image" src="https://github.com/user-attachments/assets/14b227ab-8eec-42ae-8c01-591c0fad5df8" />

propToCol  
<img width="609" height="230" alt="Image" src="https://github.com/user-attachments/assets/504f54e5-6963-462b-b4a4-8b78a8a9b06b" />

hasConditions 
<img width="724" height="367" alt="Image" src="https://github.com/user-attachments/assets/c1e20b62-b187-429b-9cc0-d4071d72a1b9" />

### Handsontable version

15.3.0

### Your environment

chrome

wuzaofeng:
Hi @adrianszymanski89 

.I’ve hard-coded the second row of data on my end, so I need to test the filter using the value I selected in my video.
Click the “move Columns” button.
In the third column, check “Cycling Cap” and “HL Mountain Shirt.”
In the first column, check “Mar 27, 2023,” making sure the page still shows both rows.
<img width="719" height="414" alt="Image" src="https://github.com/user-attachments/assets/0a846273-0af9-476e-93ab-d2db74ed799f" />

Finally, click “addItem.”
The result is that the first filter no longer displays any content.

adrianszymanski89:
@wuzaofeng 

> In the first column, check “Mar 27, 2023,” making sure the page still shows both rows.

I can't do this because those values are in the same rows as the `Cycling Cap` values. So, I'm still not sure how to reproduce the issue properly.

adrianspdev:
Fixed in the [PR](https://github.com/handsontable/handsontable/pull/12117), the update will be included in the next release.

Issue #8874, opened 2021-10-18
Filter by value of column X is removed, by editing filtered column Y


sb-thdata:
### Description
Filter by value of column X is removed, by editing filtered column Y

### Steps to reproduce
1. Set a filter to the first column (Demo: deselect "Italy" in the filter of the column "Country")
2. Set a "Filter by value" to the second column (Demo: deselect "BMW" in the filter of the column "Company")
3. Edit a row of the first column (Demo: set "Country" of "Fiat" to "Italy")
4. Open the filter of the second column (Demo: all check-boxes are deselected in the filter of the column "Company")

### Demo
https://codesandbox.io/s/handsontable-multiple-filter-issue-forked-hxln2

### Your environment
* Handsontable version: 10
* Browser Name and version: Chrome 94
* Operating System: Windows 10 Pro

adrianspdev:
Hi @sb-thdata,

Thanks for reporting this issue and for the clear reproduction steps - they made it straightforward to track down the root cause.

The fix has been merged in #12620 and will ship in the next release. The Filter by value state is now preserved correctly when editing rows in another filtered column.

Closing this issue as resolved. Feel free to reopen if you still see the problem after upgrading.

PR #13267, opened 2026-08-27
DEV-2656: Stop the value list from rewriting a column's filter conditions

### LINKED ISSUE #6471: Changing the value on a filtered column adds the value to the filter's conditions
### Description
When a column is filtered, changing the value in one of those cells adds that value to the filter's condition.

### Steps to reproduce
1. Go to https://jsfiddle.net/6fr5g7xv/1/

2. Filter column a to "foo"
image

3. Change row 5 to "bar"

4. Open the filter again and notice that "bar" is now selected, despite the user explicitly deselecting it.

image

FYI I checked it in Sheets and it behaves as I would expect. Leaves the "bar" row visible but does _not_ add it to the filter's conditions. 

image

Sheets' behavior is the expected one in my opinion, but that doesn't mean that it's _objectively_ correct.

Currently trying to figure out a workaround, will post back if/when I find one!

### Your environment
* Handsontable version: 8.0.0-beta.1 (Also tested on 7.2.2)
* Browser Name and version: Chrome Latest
* Operating System: OS X 10.14

budnix:
### Context

The PR fixes [#6471](https://github.com/handsontable/handsontable/issues/6471), open since 2019: editing a cell in a column filtered by value adds the typed value to that column's filter conditions, re-checking a box the user had explicitly cleared.

The reported symptom understates the damage. After the edit, the stored condition no longer matches the rows on screen, and the grid is not re-filtered, so the mismatch is invisible. The moment the user reopens the dropdown and confirms it **without changing anything**, every listed value counts as selected, `ValueComponent.getState()` reports `CONDITION_NONE`, and **the whole filter is dropped**. That is exactly the workflow the reporter described — set up filters, work through the rows, "just hit Filter again".

Reproduced identically on released 18.0.0 and on `develop`:

| Step | `by_value` args | Visible rows |
|---|---|---|
| unchecked `bar`, pressed OK | `[["foo"]]` | 4 |
| typed `bar` over a visible `foo` | `[["bar","foo"]]` | 4 |
| reopened dropdown | — | `bar` checked again |
| pressed OK again, touched nothing | `[]` — condition gone | **8** |

Fixing that surfaced a second, pre-existing way to lose filter conditions, so the PR closes both. They are the same underlying mistake: **the value list writes back to the condition.**

### The two changes

**1. A data change no longer rewrites the edited column's condition.**

`#onAfterChange` called `updateValueComponentCondition()`, which read the column through `getDataAtCol()` — a *visual* read, so it returned only the currently visible values — and passed that set on as `conditionArgsChange`. `component/value.ts` then did `firstByValueCondition.args[0] = conditionArgsChange` on an object owned by `ConditionCollection`, so the write persisted into exports, undo snapshots, and the next `filter()`.

That hand-off is gone, along with the `conditionArgsChange` plumbing and the `#previousConditionStack` re-sync it required. The list is still rebuilt, which is what puts the newly typed value in the list **unchecked**. This also removes a documented rule violation: reading a conditioned column via `getDataAtCol()` (#12226).

**2. A selected value the list cannot show is no longer dropped.**

The list only holds values present in the rows passing the *other* columns' filters, so a selected value can have no checkbox to read. `MultipleSelectUI.getValue()` returned only what it could see, so confirming a narrowed menu rewrote the condition to whatever was on screen — permanently, since clearing the other column did not bring the rest back. It now returns the checked listed values **plus** the selected values missing from the list.

Two rules make that hold, and both are load-bearing:

- `isSelectedAllValues()` requires **every listed value ticked *and* nothing selected outside the list**. Ticking everything on screen does not mean the column filters nothing — the values it cannot show are still excluding rows, and answering `true` makes `getState()` report `CONDITION_NONE`, which deletes them. A count comparison is not enough here: a list holding one unticked value plus one selected unlisted value has matching counts and different sets.
- `getState()` no longer treats an **empty item list** as `CONDITION_NONE` on its own. `isSelectedAllValues()` already answers `true` for an empty list with an empty selection, and a separate emptiness test fires while unlisted values are still excluding rows.

`Clear` still clears everything, unlisted values included.

This second one is **not** a regression from this PR — it reproduces on 18.0.0 by unchecking a value in an earlier column's menu, with no cell edit involved.

Two supporting fixes ride along, both in code this PR rewrites: the value list's locale was read from the *edited* column even when refreshing a dependent one, and with a physical index passed where `getCellMetaTransient` expects a visual one; and `state.args` aliased the condition collection's live array, which `exportAllConditions()` only shallow-copies, so the component state, the UI selection and the undo snapshot all shared it.

### Alignment with established filtering behavior

The resulting behavior matches what spreadsheet software and other data grids do:

- an edit that makes a row stop matching leaves the row in place until the filter is re-run explicitly;
- a value introduced by an edit joins the list **unselected**;
- the list follows the data, while the selection behind it survives.

### Behavior change, not a breaking change

**Editing a cell in a column filtered by value no longer adds the typed value to that column's filter conditions**, and confirming a value list no longer discards the selections it could not display.

Against `.ai/BREAKING-CHANGES.md`, nothing in the breaking table applies:

| Breaking-change category | This PR |
|---|---|
| Renaming a CSS class | none |
| Renaming an API (method, option, hook) | none |
| Changing a public API signature | none — `updateValueComponentCondition(columnIndex)` is unchanged |
| Removing a hook or configuration option | none |
| Changing a default setting value | none — `DEFAULT_SETTINGS` untouched |

The classes whose internals changed — `ConditionUpdateObserver`, `MultipleSelectUI`, `ValueComponent` — are all `@private` and absent from the API reference, which that document classifies as *not* breaking. No option, default, hook, or documented signature moved, and no functionality was removed.

`ConditionUpdateObserver.updateStatesAtColumn()` lost its optional second parameter (`conditionArgsChange`), which existed only to carry the write this PR removes. It is private, undocumented, and not in the API reference; noted here for completeness.

### How has this been tested?

New Playwright coverage, each part **verified red against the code it guards**:

- `filters-edit-in-filtered-column.spec.ts` — reverting the source and rebuilding turned two of its three tests red for the right reasons: `Green` came back checked, and all five rows returned instead of two. The third asserts the row stays put until `filter()` runs, then that `filter()` removes it — the recipe the guide now documents.
- `filters-value-list.spec.ts` — `confirming a narrowed list keeps the values it cannot show` goes red with the `getValue()` union disabled, and `confirming a list whose every shown value is ticked keeps the rest of the filter` goes red against the count-comparison form of `isSelectedAllValues()`.

Two existing Jasmine specs were strengthened rather than adjusted. The `#11832` test was hollow — `addCondition` already writes the state through the collection's `afterAdd` hook, so it passed with the method body deleted, and passed when the method refreshed the *wrong* column; it now wipes the state first and fails on a no-op. The `#8874` test now asserts the real invariant: the list follows the data while the stored selection does not shrink.

```bash
```

Results, all run before the commit:

- Core E2E (Jasmine/Puppeteer), full suite: **9723 specs, 0 failures**
- Core unit (Jest), full suite: **3963 tests, 325 suites, 0 failures**
- `test:types`: pass
- Playwright `e2e-main`, full tier: **441 passed, 1 skipped**
- ESLint: **0 errors**
- Manual verification in Chrome against a local build, for every scenario above

### Follow-ups found while testing — not fixed here

**Undo of a data change addresses a visual row.** `DataChangeAction.undo/redo` (`undoRedo/actions/dataChange.ts`) replays edits through `setDataAtCell` using the visual row recorded at edit time. When a filter has since trimmed that row, the write lands on the wrong row or grows the dataset — measured as a silent no-op on untouched 18.0.0, and as a phantom extra source row in one flow here. Its guard compares only *visible* row counts, so it cannot see source growth. The fix needs physical-row addressing with a `setSourceDataAtCell` fallback, but the completion callback rides on `setDataAtCell` firing `afterChange`, so an all-source route would never call `undoneCallback()` and would break the undo stack.

**`getState()` omits `locale`.** `#onActionBarSubmit` runs `flush()` (which computes `state.locale`) and then `saveState()`, whose `getState()` return has no `locale` — so every OK overwrites it with `undefined` and the search box stops using the column's locale. Pre-existing, in a different mechanism, and it needs its own test story.

Neither is touched by the coverage added here.

### Types of changes

### Related issue(s):

1. DEV-2656
2. Fixes #6471

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

Documentation updated here: a new "Edit cells in a filtered column" section in the column-filter guide, with a re-filter-on-edit recipe in all four framework variants, plus two landmine entries in the filters plugin's `AGENTS.md`.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2656

---

> [!NOTE]
> **Medium Risk**
> Changes core filters condition state and dropdown confirmation paths; behavior is well-tested but mistakes could drop or stick filter conditions across edits and multi-column stacks.
> 
> **Overview**
> Fixes **filter-by-value** behavior where cell edits and confirming a narrowed value list could silently rewrite or drop column conditions ([#6471](https://github.com/handsontable/handsontable/issues/6471)).
> 
> **After a data change**, the value list is refreshed from rows passing *other* columns’ filters, but the stored selection is no longer replaced with visible values (`conditionArgsChange` / `getDataAtCol()` hand-off removed). Typed values appear in the list **unchecked**; the filter does not re-run until `filter()` (documented in the column-filter guide).
> 
> **When the list cannot show every selected value**, `MultipleSelectUI.getValue()` now returns listed checkboxes **plus** unlisted selections; `#pruneToExistingValues()` only removes values that no longer exist in the column. **`isSelectedAllValues()` / `getState()`** no longer treat “all visible boxes ticked” or an empty list as “no filter,” with a **`#cleared`** flag so intentional **Clear** still means exclude everything.
> 
> Adds Playwright/Jasmine coverage and expands filters plugin `AGENTS.md` invariants.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2723a85fc97e8fa644757de8ecf3983e53413727. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:

handsontable/src/plugins/filters/ui/multipleSelect.ts
```diff
@@ -170,10 +170,34 @@ export class MultipleSelectUI extends BaseUI {
...
+   * @returns {Array} Selected values missing from the item list.
+   */
+  #getUnlistedValue(): unknown[] {
+    const value = (this.options as Record<string, unknown>).value as unknown[] | undefined;
+
+    if (!Array.isArray(value) || value.length === 0) {
+      return [];
+    }
+
+    const isListed = createArrayAssertion(this.#items.map(item => item.value));
```
Haxikowy, line 242: A value leaves `#items` for two reasons, and this treats them as one: an earlier column's condition hides its rows (must survive), or it no longer exists in the column at all (should not).

Repro, one column, no other filters: data `foo, bar, foo`, condition `by_value [['foo','bar']]`, then type `foo` over the only `bar`. The list is `['foo']`, `options.value` is still `['foo','bar']`, so `bar` is "unlisted" forever. `isSelectedAllValues()` can never be true again: **Select all** no longer clears the column, the header keeps `htFiltersActive`, `exportConditions()` exports a value that exists nowhere, and with two conditions `getOperationBasedOnArguments()` upgrades `or` to `or_then_variable`. Only Clear escapes.

Prune against the column's full value set before storing `state.args` (`getDataMapAtColumn(physicalColumn)`): a value hidden by an earlier column survives, a value that exists nowhere is dropped.

Same comparison also matches by strict `Set` identity while the list values went through `toEmptyString`/`unifyColumnValues`, so `addCondition(col, 'by_value', [[null]])` or `['1']` against a numeric column produces a permanent phantom.

handsontable/src/plugins/filters/ui/multipleSelect.ts
```diff
@@ -475,6 +507,12 @@ export class MultipleSelectUI extends BaseUI {
       return;
     }
 
+    // Drop the selected values the list cannot show as well. Without this they would survive into
+    // `getValue()` and the filter would keep matching rows the emptied box no longer accounts for.
+    // Note this still unchecks only the rows the box currently holds, so with an active search term
+    // the values it filtered out keep their state - long-standing behavior, unchanged here.
+    (this.options as Record<string, unknown>).value = [];
```
Haxikowy, line 582: `#onClearAllClick` resets `options.value`, `#onSelectAllClick` does not. So ticking every box leaves the unlisted selection in place, `isSelectedAllValues()` stays false, and the column keeps a `by_value` condition instead of reporting `CONDITION_NONE`. Same rows on screen, but the column still reads as filtered and still exports a condition. On `develop` Select-all always cleared it.

Mirror the reset in `#onSelectAllClick`.

handsontable/src/plugins/filters/ui/multipleSelect.ts
```diff
@@ -209,7 +233,15 @@ export class MultipleSelectUI extends BaseUI {
...
-    return this.#items.length === this.getValue().length;
+    // Two conditions, not one count. Every listed value has to be ticked AND nothing may be
+    // selected outside the list. Ticking everything on screen does NOT mean the column filters
+    // nothing - the values the list cannot show are still excluding rows, and answering `true`
+    // here makes `getState()` report "no condition", which deletes them.
+    // Comparing `#items.length` against the whole selection would not do: a list holding one
+    // unticked value plus one selected unlisted value has matching counts and different sets.
+    // An empty list with an empty selection still answers `true`, which is what lets a column with
+    // nothing to filter by report "no condition".
+    return this.#items.length === itemsToValue(this.#items).length && this.#getUnlistedValue().length === 0;
```
Haxikowy, line 299: Clear now means two different things. On a non-empty list it stores `by_value []`. On an empty list (every row excluded by an earlier column) `#items.length === 0` and the unlisted set is empty, so this returns true and the filter is dropped outright, bringing those rows back as soon as the earlier column is widened.

A `cleared` flag set by `#onClearAllClick` would force `CONDITION_BY_VALUE` instead of inferring it from list emptiness. The new E2E uses a two-item list, so only the first branch is covered.

handsontable/src/plugins/filters/ui/multipleSelect.ts
```diff
@@ -170,10 +170,34 @@ export class MultipleSelectUI extends BaseUI {
...
+   * The list only ever holds the values present in the rows that pass the other columns' filters,
+   * so a selected value whose rows are currently filtered out has no checkbox to read. Those values
+   * are carried through untouched instead of being dropped - otherwise confirming the menu would
+   * silently shrink the column's condition to whatever happens to be on screen.
+   *
    * @returns {Array} Array of selected values.
    */
   getValue() {
-    return itemsToValue(this.#items);
+    return itemsToValue(this.#items).concat(this.#getUnlistedValue());
```
Haxikowy, line 219: In `searchMode: 'apply'`, `#onInput` rewrites `#items[].checked` from the search term, and nothing resets `options.value`. So OK applies the search matches plus whatever the list cannot show, i.e. wider than what the box displays. Skipping the union when `searchMode === 'apply'` keeps that mode's contract.

handsontable/src/plugins/filters/ui/multipleSelect.ts
```diff
@@ -170,10 +170,34 @@ export class MultipleSelectUI extends BaseUI {
...
+  }
+
+  /**
+   * The selected values that the current list cannot show, and that the user therefore cannot
+   * toggle. Reading them from the restored selection keeps them stable while the user works with
+   * the checkboxes, because only listed values can ever be checked or unchecked.
+   *
+   * @returns {Array} Selected values missing from the item list.
+   */
+  #getUnlistedValue(): unknown[] {
```
Haxikowy, line 229: `#getUnlistedValue()` rebuilds `items.map(...)` and a Set on every call, and `getState()` calls it four times per OK click (twice through `getValue()`, twice through `isSelectedAllValues()`).

Measured in Node on the isolated shapes, 100k distinct values all checked: 14.7 ms old, 43–114 ms new, 10–22 ms with the Set cached against `#items`. Invisible at 10k and below, so this is optional — but the fix is about six lines and lands under the old cost.

handsontable/src/plugins/filters/conditionUpdateObserver.ts
```diff
@@ -171,11 +171,8 @@ class ConditionUpdateObserver {
    * Update all related states which should be changed after invoking changes applied to current column.
    *
    * @param {number} column The column index.
-   * @param {object} conditionArgsChange Object describing condition changes which can be handled by filters on `update` hook.
-   * It contains keys `conditionKey` and `conditionValue` which refers to change specified key of condition to specified value
-   * based on referred keys.
    */
-  updateStatesAtColumn(column: number, conditionArgsChange?: unknown) {
+  updateStatesAtColumn(column: number) {
```
Haxikowy: The dropped parameter is on the shipped type surface: `.d.ts` emit ignores `@private`, and `filters.d.ts` declares `conditionUpdateObserver` as a public property, so `updateStatesAtColumn(col, args)` stops compiling for TS users after upgrade. Runtime is unaffected.

Not breaking under `.ai/BREAKING-CHANGES.md` (absent from the API reference), so this is a judgment call — but keeping `_conditionArgsChange?: unknown` ignored costs nothing.

Either way, `conditionUpdateObserver.types.ts` now only proves the one-arg call compiles. A `@ts-expect-error`ed two-arg call would assert the removal instead of assuming it.

tests/e2e/filters-value-list.spec.ts
```diff
@@ -109,6 +109,117 @@ test.describe('Filters — "filter by value" list', () => {
...
+    await grid.uncheckValue('Alice');
+    await grid.confirmMenu();
+
+    // Red is selected but unlisted at this point. "Clear" has to drop it too, or the box would
+    // look empty while the filter still matched rows.
+    await grid.openMenu('Color');
+    await grid.clearAllValues();
+    await grid.confirmMenu();
+
+    expect(await grid.columnValues(0)).toEqual([]);
```
Haxikowy, line 258: This assertion passes with or without the `options.value = []` fix. At this point Name is `[Bob, Charlie, Dave]`; with the fix Color is `[]`, without it Color is `['Red']`, and none of the three is Red — so both produce `[]`.

Re-check `Alice` in the Name menu after Clear + OK, then assert `columnValues(0)` is still `[]`. With the fix her Red row stays hidden; without it, it comes back.

handsontable/src/plugins/filters/__tests__/plugins/undoRedo.spec.js
```diff
@@ -161,7 +161,12 @@ describe('Filters UI cooperation with UndoRedo', () => {
     const undoPlugin = getPlugin('undoRedo');
     const filtersPlugin = getPlugin('filters');
 
-    filtersPlugin.addCondition(2, 'by_value', [['Gardiner']]);
+    // Both values the row takes during this test are selected up front. Editing a cell no longer
+    // adds its new value to the condition (issue #6471), so the row has to stay matched by the
+    // user's own selection for the filter to keep finding it across the undo/redo steps.
+    // A blank is `''` in `by_value` args, never `null` - the filter compares values that went
+    // through `toEmptyString()`, and the assertion behind it matches by strict identity.
+    filtersPlugin.addCondition(2, 'by_value', [['Gardiner', '']]);
```
Haxikowy, line 169: The widened seed is necessary under the new behavior, agreed. Worth noting what it stops covering: with `[['Gardiner']]` the edit to `null` leaves the row unmatched, which is the undo-of-a-trimmed-row flow the description reports as producing a phantom source row. Nothing covers it now.

A second `it()` pinning the current intended outcome would keep that path visible.

docs/content/guides/columns/column-filter/column-filter.md
```diff
@@ -690,6 +690,109 @@ column menu's width for better user experience. You can achieve this with by sty
...
+```ts
+settings = {
+  dropdownMenu: true,
+  filters: true,
+  afterChange(changes: any, source: string) {
+    if (source === 'loadData' || !changes) {
+      return;
+    }
+
+    this.getPlugin('filters').filter();
```
Haxikowy: In a class-property object literal, `this` is the literal's own type, so `this.getPlugin` is TS2339 under the repo's settings. Runtime is fine since Handsontable applies hooks with the instance as context, but the snippet as printed does not compile. Typing `settings` as `GridSettings` fixes it.

Haxikowy (commented):
component/value.ts:256-259 — pre-existing, inside the rewritten function. rowMetaMap passes row.row (physical) and physicalColumn to getCellMetaTransient, which wants visual for both and translates again. Under any sort or filter, modifyFiltersMultiSelectValue gets a different cell's meta. Contradicts the comment added at :196. reset() at :335 already uses row.meta; the two paths disagree.

Both declared follow-ups are untracked. No GitHub issue, no ClickUp ID for either the DataChangeAction visual-row addressing or the missing locale in getState(). They vanish on squash-merge. The undo one goes from a silent no-op on 18.0.0 to a phantom source row here.

Clean: undo/redo (develop was the broken side); toVisualColumn fix correct; perf net win (~5 ms removed per edit); changelog schema/type/breaking honest; docs present in all four frameworks; the two Jasmine hardenings the body names are genuine; StateInfo not reachable from any package entry point; no default, option, hook, or CSS class touched.

handsontable/src/plugins/filters/ui/multipleSelect.ts
budnix: Good spot, and this was a bad trade on my part — I judged the stale value harmless and it is not. Fixed in 88625cf71e: `#pruneToExistingValues()` now drops values that left the column, reading the full set through the memo the row factory already fills.

handsontable/src/plugins/filters/ui/multipleSelect.ts
budnix: You are right — mirrored in 88625cf71e. Writing the test for it caught that I also had to clear the new unlisted cache, or the reset did nothing at all.

handsontable/src/plugins/filters/ui/multipleSelect.ts
budnix: Agreed — added the `cleared` flag in 88625cf71e, so an empty list no longer has to guess which of the two the user meant.

handsontable/src/plugins/filters/ui/multipleSelect.ts
budnix: Fair point — fixed in 88625cf71e. I reset the selection inside the `apply` branch instead of skipping the union for the whole mode, so an apply-mode user who never searches still keeps the hidden values.

handsontable/src/plugins/filters/ui/multipleSelect.ts
budnix: Nice catch, and thank you for measuring it — cached against `#items` in 88625cf71e, the same way the lowercase list already works.

handsontable/src/plugins/filters/conditionUpdateObserver.ts
budnix: You are right and I had this wrong in the PR description — the `.d.ts` does ship it. Kept the parameter as an ignored one in 88625cf71e, and the type test now asserts the two-argument call instead of assuming it.

tests/e2e/filters-value-list.spec.ts
budnix: Good catch — that one proved nothing. Tightened in 88625cf71e with the Alice re-check, and I confirmed it goes red without the fix.

handsontable/src/plugins/filters/__tests__/plugins/undoRedo.spec.js
budnix: Agreed the coverage is gone. I did not pin it, because the current outcome there is the broken undo — filed DEV-2665 with the gap written into it, so it comes back with the fix.

docs/content/guides/columns/column-filter/column-filter.md
budnix: Yep, my mistake — fixed in 88625cf71e. `GridSettings` alone was not enough though: `afterChange` declares no `this` type, so I dropped `this` from the Angular, React and Vue snippets and used a component reference instead.

handsontable/src/plugins/filters/ui/multipleSelect.ts
Haxikowy: The flag fixes the ambiguity but never syncs back. Ticking boxes after Clear leaves `#cleared` true, so a fully re-ticked list confirms as `CONDITION_BY_VALUE` instead of releasing the column — the symptom the new "Select all" test asserts against. And `setItems`/`setValue` both reset it on reopen, so a still-empty cleared list confirmed a second time reports `CONDITION_NONE` and drops the exclusion.

It needs clearing on the first tick, and it needs to survive the `getState()`/`setState()` round-trip, which carries no place to record it today.

handsontable/src/plugins/filters/ui/multipleSelect.ts
Haxikowy: Fix is right. Two leftovers: the three-line comment above the reset is pasted three times, and `#pruneToExistingValues`'s JSDoc says both sides are normalized while only `existingValues` goes through `toEmptyString` — a raw `[[null]]` gets pruned rather than matched to the blank. Same as `develop`, so just the comment, unless `selectedArgs.filter(v => exists(toEmptyString(v)))` is worth it.

handsontable/src/plugins/filters/ui/multipleSelect.ts
budnix: Good catch — 23e0c1379a already releases it on the first tick, and your reopen point turned up a hole: deriving it in setValue also marked a never-filtered column as cleared. Moved the restore into ValueComponent.setState() in 2723a85fc9.

handsontable/src/plugins/filters/ui/multipleSelect.ts
budnix: Nice catch — dropped the duplicate comment in 2723a85fc9. I fixed the JSDoc rather than the code: a raw null never matched a row anyway (getDataMapAtColumn normalizes cell values), so pruning it is the right outcome.

Haxikowy (approved):

PR #13268, opened 2026-08-27
DEV-2655: Honor an explicit editor={false} in the React wrapper

### LINKED ISSUE #7561: Setting editor to false doesn't disable editing
### Description

setting the editor as false doesn't appear to disable editing

### Steps to reproduce

1. Set `editor={false}` in a `<HotColumn />`

### Demo

https://codesandbox.io/s/declaring-column-settings-iic8s?file=/src/index.js

### Your environment
* React wrapper version: 3.1.3
* Handsontable version: 7.4.2
* Browser Name and version:
* Operating System:

budnix:
### Context

The PR fixes [#7561](https://github.com/handsontable/handsontable/issues/7561), open since June 2020: setting the `editor` prop to `false` in the React wrapper did not disable editing, and the cell still opened the default text editor.

One mistake, in two places. A falsy check cannot tell "prop not provided" apart from "prop explicitly set to a boolean", so both sites threw the boolean away and substituted a working editor:

- `wrappers/react-wrapper/src/hotTableInner.tsx` — the `else` ran `newSettings.editor = props.hotEditor || getEditor('text')`, which overwrote an explicit `false`.
- `wrappers/react-wrapper/src/hotColumn.tsx` — `else if (props.hotEditor)` skipped a falsy value, so it never reached the column settings.

Core is not at fault. `getCellEditor()` returns `false` and `isCellEditable()` treats a falsy editor class as not-editable; both vanilla-core control cases pass.

Both sites now share one `resolveEditorSetting()` helper in `helpers.tsx`, so the rule is written once rather than duplicated in two subtly different shapes.

**Worth knowing when reviewing:** the case in the original report — `<HotColumn editor={false} />` — already worked before this PR, but only by accident. `SettingsMapper` copies every prop verbatim into the column settings, so `editor: false` landed there on its own and `hotColumn.tsx` had no unconditional `else` to overwrite it. The sibling configurations and the dynamic update path were genuinely broken, and they share the one root cause, which is why the issue is closed here rather than on the original repro alone.

| Case | before | after |
|---|---|---|
| `<HotColumn editor={false} />` | worked by accident | works by intent |
| `<HotColumn hotEditor={false} />` | broken | fixed |
| `<HotTable editor={false}>` | broken | fixed |
| `<HotTable hotEditor={false}>` | broken | fixed |
| dynamic switch of `editor` / `hotEditor` to/from `false` | broken | fixed |
| `editor={true}` / `hotEditor={true}` (either level) | threw, or opened an empty editor | uses the default editor |

### Precedence between `editor` and `hotEditor`

`editor={false}` says "no component editor". It does **not** override an editor named in `hotEditor`, so `<HotTable editor={false} hotEditor={MyEditor} />` still uses `MyEditor` — the behavior in every released version. The helper resolves `hotEditor` before a falsy `editor` to preserve that exactly.

An earlier revision of this branch got this wrong and let `editor={false}` suppress a real `hotEditor`. That was caught in review and fixed in the second commit; it never appeared in a release. Two tests now pin the behavior, and both pass against `develop` as well as against this branch, confirming they encode pre-existing behavior rather than a new rule.

### Booleans that are not `false`

**Updated after review:** `true` is now treated as if the prop were **not provided**, rather than
resolving to the default editor. The difference shows at column level, where "not provided" inherits
but a hard-set class overrides: `<HotColumn type="numeric" hotEditor={true} />` was losing its numeric
editor. Both prop spellings now behave the same way.

Two pre-existing gaps in the same chain, fixed here because the change is what makes the boolean handling coherent:

- `hotEditor={true}` was forwarded to the core verbatim, which threw `Only strings and functions can be passed as "editor" parameter` the moment a cell was selected.
- `editor={true}` built a component-editor class, but `createEditorPortal` returns `null` for a boolean, so no component ever mounted behind it — the grid entered an editing state that rendered nothing.

Both now resolve to the default editor, and a bare `true` never reaches the core. `boolean` is part of the declared type on both props (`editor?: ComponentType | boolean`, and core's `editor?: string | constructor | boolean`), so these were type-legal configurations that crashed. Happy to split them out if you would rather keep this PR to `false` only.

### Not a breaking change

Only configurations that pass a boolean are affected, and today those either silently do nothing, throw, or open an editor with no UI. No type change was needed — both booleans already typecheck on both components, so this was a runtime-only defect against an already-correct type.

Vue 3 and Angular do not have this bug — neither wrapper contains a `getEditor('text')` fallback, and Vue's `HotColumn.vue` spreads props straight into `columnSettings`. No parity work is needed.

### How has this been tested?

Sixteen tests added to the `Editor configuration using React components` describe of the two spec files: both prop names set to `false`, the dynamic `undefined → false → undefined` switch on each, the `hotEditor`-wins-over-`editor={false}` precedence rule, `true` on both props, a falsy-but-not-`false` value (`null`), a bare `<HotColumn />` inheriting a grid-level `false`, and both sides of the `editor={condition && Editor}` idiom.

Each disabling test asserts `getCellEditor(0, 0) === false`, then `selectCell` + Enter leaves `getActiveEditor()` undefined and the cell data untouched. Every test carries a positive control that must still open a `TextEditor` — a neighboring untouched `<HotColumn>` at column level, and a `columns={[{}, { editor: 'text' }]}` override at table level, which also proves a global `false` does not clobber a per-column editor.

**Every test was verified against the code it guards.** With `wrappers/react-wrapper/src/` reverted to `develop`, **10 of the 14 new tests fail and 4 pass** (an earlier revision of this description said 2 pass — that was a miscount, thanks @Haxikowy). The 4 that pass are the two `<HotColumn editor={false} />` cases, which already worked by accident, and the two `hotEditor`-precedence cases, which pin behavior `develop` already has. They are kept deliberately as regression guards. The remaining tests were each confirmed failing against the commit they guard.

```bash
# Full React wrapper suite: 16 suites, 94 tests, all passing (78 before this PR)

# Just the editor tests
cd wrappers/react-wrapper && npx jest -t "editor"
```

Two notes for anyone re-checking this locally:

- `npx tsc --noEmit` in `wrappers/react-wrapper` **silently checks nothing** unless you pass `--ignoreDeprecations 6.0`, because the package tsconfig trips TS5107/TS5101 and `tsc` bails on the config before reading any file. Against a proper `develop` baseline this change adds zero type errors (three pre-existing ones remain: `hotColumn.tsx`, `hotTableInner.tsx`, `hotTableContext.tsx`).
- No ESLint run applies here. `scripts/lint-files.mjs` lists react/angular/docs as having no plain-eslint script, so neither the local hook nor CI lints this package.

### Types of changes

### Related issue(s):

1. DEV-2655
2. Closes #7561

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

The documentation box is left unchecked deliberately. The guides that configure `editor: false` do so through the `columns` prop and `cells`, paths that were never broken, so no guide describes the behavior this PR changes.

One changelog entry, `.changelogs/7561.json`, per the one-entry-per-PR rule. It confirms the `<HotColumn editor={false} />` case from the issue behaves as reported, and notes that the `renderer`/`hotRenderer` props are unchanged — they carry the same falsy-check shape but are out of scope here.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2655

---

> [!NOTE]
> **Low Risk**
> Scoped to React wrapper editor prop mapping with broad test coverage; fixes incorrect runtime behavior without API breakage.
> 
> **Overview**
> Fixes [#7561](https://github.com/handsontable/handsontable/issues/7561) by centralizing editor resolution in **`isComponentEditor`** and **`resolveEditorSetting`** in `helpers.tsx`, then wiring **`HotTable`** and **`HotColumn`** to use them instead of truthy checks that replaced explicit **`false`** with the default text editor.
> 
> **`editor={false}`** and **`hotEditor={false}`** now disable editing at table and column level (including dynamic toggles). **`true`**, **`null`**, and other falsy values except **`false`** behave as if the prop was omitted so column types and grid defaults still apply and invalid booleans are not passed to core. Precedence is unchanged: a component **`editor`** wins; **`hotEditor`** still overrides a bare **`editor={false}`**.
> 
> Prop JSDoc in **`types.tsx`** documents this; **`renderer` / `hotRenderer`** are untouched. Tests cover disable/enable paths, precedence, and conditional **`editor={cond && Editor}`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 526836b7a644c9f38605ae4e1a0d9a8e345c683e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changelogs/7561.json
+++ b/.changelogs/7561.json
@@ -1,6 +1,6 @@
 {
   "issuesOrigin": "public",
-  "title": "Fixed a bug where setting the `editor` or `hotEditor` prop to `false` on `HotTable`, or the `hotEditor` prop to `false` on `HotColumn`, left the cell editable and still opened the default text editor.",
+  "title": "Fixed a bug where the `editor` and `hotEditor` props did not accept a boolean in the React wrapper. Setting either to `false` now disables editing on `HotTable` and on `HotColumn` (the reported `<HotColumn editor={false} />` case keeps working as asked), unless a component is passed to `editor` or an editor is named in `hotEditor`, both of which still win. Setting either to `true`, or to another falsy value such as `null`, is now treated as if the prop were not provided, so the cell keeps the editor its column type or the grid supplies instead of throwing or locking. The `renderer` and `hotRenderer` props are unchanged.",
   "type": "fixed",
   "issueOrPR": 7561,
   "breaking": false,
--- a/.changelogs/7561.json
+++ b/.changelogs/7561.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed a bug where setting the `editor` or `hotEditor` prop to `false` on `HotTable`, or the `hotEditor` prop to `false` on `HotColumn`, left the cell editable and still opened the default text editor.",
+  "type": "fixed",
+  "issueOrPR": 7561,
+  "breaking": false,
+  "framework": "react"
+}
--- a/wrappers/react-wrapper/src/helpers.tsx
+++ b/wrappers/react-wrapper/src/helpers.tsx
@@ -9,7 +9,6 @@ import React, {
 } from 'react';
 import ReactDOM from 'react-dom';
 import Handsontable from 'handsontable/base';
-import { getEditor } from 'handsontable/editors/registry';
 import { HotTableProps } from './types';
 
 let bulkComponentContainer: DocumentFragment | null = null;
@@ -151,12 +150,18 @@ export function isComponentEditor(editor: HotTableProps['editor']): boolean {
  * Resolve the Handsontable `editor` setting from the two editor props.
  *
  * `editor` carries the component editor and `hotEditor` the Handsontable-native one, but both accept
- * a boolean. `false` disables editing, while `true` means "the default editor" and must never reach
- * the core, which accepts only a string or a constructor and throws on anything else.
+ * a boolean:
  *
- * `hotEditor` is resolved before a falsy `editor` so that an editor named there still wins over a
- * bare `editor={false}` — the behavior in every released version, where `editor={false}` fell through
- * to the `hotEditor` value.
+ * - `false` disables editing.
+ * - `true` names no editor, so it is treated as if the prop were not provided. It must never reach
+ *   the core, which accepts only a string or a constructor and throws on anything else.
+ * - Any other nullish or falsy value (`null`, `0`, `''`) is also treated as "not provided", so a
+ *   column written as `editor={cond ? MyEditor : null}` inherits instead of locking.
+ *
+ * `hotEditor` is resolved before a falsy `editor`, so an editor named there still wins over a bare
+ * `editor={false}` — the behavior in every released version, where `editor={false}` fell through to
+ * the `hotEditor` value. A component `editor` outranks both: it is picked by `isComponentEditor()`
+ * before this function is reached, so `editor={MyComponent} hotEditor={false}` keeps the component.
  *
  * @param {HotTableProps['editor']} editor The `editor` prop.
  * @param {HotTableProps['hotEditor']} hotEditor The `hotEditor` prop.
@@ -170,11 +175,9 @@ export function resolveEditorSetting(
     return false;
   }
 
-  if (hotEditor === true) {
-    return getEditor('text') as Handsontable.GridSettings['editor'];
-  }
-
-  if (hotEditor) {
+  // `true` names no editor, so it must fall through to "not provided" rather than hard-setting the
+  // default one — otherwise it would override the editor a column's `type` or the grid supplies.
+  if (hotEditor && hotEditor !== true) {
     return hotEditor as Handsontable.GridSettings['editor'];
   }
 
--- a/wrappers/react-wrapper/src/helpers.tsx
+++ b/wrappers/react-wrapper/src/helpers.tsx
@@ -8,6 +8,8 @@ import React, {
   useEffect,
 } from 'react';
 import ReactDOM from 'react-dom';
+import Handsontable from 'handsontable/base';
+import { getEditor } from 'handsontable/editors/registry';
 import { HotTableProps } from './types';
 
 let bulkComponentContainer: DocumentFragment | null = null;
@@ -130,6 +132,53 @@ function hasChildElementOfType(children: ReactNode, type: 'hot-renderer' | 'hot-
   return childrenArray.some((child) => {
       return (child as React.ReactElement).props[type] !== void 0;
   });
+}
+
+/**
+ * Check whether the `editor` prop holds a component editor, as opposed to a boolean flag.
+ *
+ * Both editor props accept a boolean, so a truthy check alone is not enough to tell a component
+ * apart from a bare `editor={true}`.
+ *
+ * @param {HotTableProps['editor']} editor The `editor` prop.
+ * @returns {boolean} `true` when the prop carries a component to render.
+ */
+export function isComponentEditor(editor: HotTableProps['editor']): boolean {
+  return !!editor && typeof editor !== 'boolean';
+}
+
+/**
+ * Resolve the Handsontable `editor` setting from the two editor props.
+ *
+ * `editor` carries the component editor and `hotEditor` the Handsontable-native one, but both accept
+ * a boolean. `false` disables editing, while `true` means "the default editor" and must never reach
+ * the core, which accepts only a string or a constructor and throws on anything else.
+ *
+ * `hotEditor` is resolved before a falsy `editor` so that an editor named there still wins over a
+ * bare `editor={false}` — the behavior in every released version, where `editor={false}` fell through
+ * to the `hotEditor` value.
+ *
+ * @param {HotTableProps['editor']} editor The `editor` prop.
+ * @param {HotTableProps['hotEditor']} hotEditor The `hotEditor` prop.
+ * @returns {*} The editor setting, or `undefined` when neither prop names one.
+ */
+export function resolveEditorSetting(
+  editor: HotTableProps['editor'],
+  hotEditor: HotTableProps['hotEditor']
+): Handsontable.GridSettings['editor'] | undefined {
+  if (hotEditor === false) {
+    return false;
+  }
+
+  if (hotEditor === true) {
+    return getEditor('text') as Handsontable.GridSettings['editor'];
+  }
+
+  if (hotEditor) {
+    return hotEditor as Handsontable.GridSettings['editor'];
+  }
+
+  return editor === false ? false : undefined;
 }
 
 /**
--- a/wrappers/react-wrapper/src/hotColumn.tsx
+++ b/wrappers/react-wrapper/src/hotColumn.tsx
@@ -8,7 +8,9 @@ import { HotTableProps, HotColumnProps, HotEditorHooks } from './types';
 import {
   createEditorPortal,
   displayAnyChildrenWarning,
-  displayObsoleteRenderersEditorsWarning
+  displayObsoleteRenderersEditorsWarning,
+  isComponentEditor,
+  resolveEditorSetting
 } from './helpers';
 import { SettingsMapper } from './settingsMapper';
 import Handsontable from 'handsontable/base';
@@ -66,14 +68,18 @@ const HotColumn: FC<HotColumnProps> = (props) => {
         columnSettings.renderer = props.hotRenderer;
       }
 
-      if (props.editor) {
+      if (isComponentEditor(props.editor)) {
         columnSettings.editor = makeEditorClass(localEditorHooksRef, localEditorClassInstance);
-      } else if (props.editor === false || props.hotEditor === false) {
-        // `false` disables editing on either prop name. Set it here instead of relying on the
-        // `SettingsMapper` prop passthrough, which only ever carried the `editor` name.
-        columnSettings.editor = false;
-      } else if (props.hotEditor) {
-        columnSettings.editor = props.hotEditor;
+      } else {
+        const editorSetting = resolveEditorSetting(props.editor, props.hotEditor);
+
+        if (editorSetting === undefined) {
+          // Neither prop named an editor. Drop whatever the settings mapper copied over — it can be
+          // a bare `true`, which the core rejects — and let the column inherit the grid editor.
+          delete columnSettings.editor;
+        } else {
+          columnSettings.editor = editorSetting;
+        }
       }
 
       return columnSettings
--- a/wrappers/react-wrapper/src/hotColumn.tsx
+++ b/wrappers/react-wrapper/src/hotColumn.tsx
@@ -68,6 +68,10 @@ const HotColumn: FC<HotColumnProps> = (props) => {
 
       if (props.editor) {
         columnSettings.editor = makeEditorClass(localEditorHooksRef, localEditorClassInstance);
+      } else if (props.editor === false || props.hotEditor === false) {
+        // `false` disables editing on either prop name. Set it here instead of relying on the
+        // `SettingsMapper` prop passthrough, which only ever carried the `editor` name.
+        columnSettings.editor = false;
       } else if (props.hotEditor) {
         columnSettings.editor = props.hotEditor;
       }
--- a/wrappers/react-wrapper/src/hotTableInner.tsx
+++ b/wrappers/react-wrapper/src/hotTableInner.tsx
@@ -136,6 +136,10 @@ const HotTableInner = forwardRef<
 
     if (props.editor) {
       newSettings.editor = makeEditorClass(globalEditorHooksRef, globalEditorClassInstance);
+    } else if (props.editor === false || props.hotEditor === false) {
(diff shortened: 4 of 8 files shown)


qunabu:
Task linked: [DEV-2655 React wrapper: editor={false} does not disable cell editing](https://app.clickup.com/t/9015210959/DEV-2655)

wrappers/react-wrapper/src/helpers.tsx
```diff
@@ -132,6 +134,53 @@ function hasChildElementOfType(children: ReactNode, type: 'hot-renderer' | 'hot-
...
+ */
+export function resolveEditorSetting(
+  editor: HotTableProps['editor'],
+  hotEditor: HotTableProps['hotEditor']
+): Handsontable.GridSettings['editor'] | undefined {
+  if (hotEditor === false) {
+    return false;
+  }
+
+  if (hotEditor === true) {
```
Haxikowy: `hotEditor === true` returns the TextEditor class, while `editor === true` falls through to `undefined`. At column level those land in different places: `undefined` deletes the key and inherits, the class hard-sets text.

So `<HotColumn type="date" hotEditor={true} />` loses the date editor and `<HotColumn type="date" editor={true} />` keeps it. Same split for a column under `<HotTable hotEditor={MyEditor}>`.

The new test does not catch it because its grid default is already text, so inherit and hard-set give the same answer.

Dropping this block makes `true` mean "prop omitted" on both props at both levels, which is what the JSDoc above already promises. Table level is unaffected, `undefined` resolves to `getEditor('text')` there anyway. One case does move: `editor={false} hotEditor={true}` falls through to disabled instead of text, which reads as correct since `true` names no editor.

wrappers/react-wrapper/src/hotColumn.tsx
```diff
@@ -66,10 +68,18 @@ const HotColumn: FC<HotColumnProps> = (props) => {
...
         columnSettings.editor = makeEditorClass(localEditorHooksRef, localEditorClassInstance);
-      } else if (props.hotEditor) {
-        columnSettings.editor = props.hotEditor;
+      } else {
+        const editorSetting = resolveEditorSetting(props.editor, props.hotEditor);
+
+        if (editorSetting === undefined) {
+          // Neither prop named an editor. Drop whatever the settings mapper copied over — it can be
+          // a bare `true`, which the core rejects — and let the column inherit the grid editor.
+          delete columnSettings.editor;
```
Haxikowy, line 79: `delete` also fires for `editor={null}` and `editor={0}`. Core defaults to text only on `undefined` (`getCellEditor`, core.ts:4771); any other falsy value reaches `editorManager`'s `!editorClass` guard and locks the cell. A column written as `editor={cond ? MyEditor : null}` was locked on develop and is editable after this change.

Inheriting is the better semantic and it matches table level, so keep it. It belongs in the changelog and in `resolveEditorSetting`'s JSDoc, since nothing else records it. The same line also fixes `editor={undefined}`, which used to lock a column by accident.

wrappers/react-wrapper/src/hotTableInner.tsx
```diff
@@ -134,10 +136,15 @@ const HotTableInner = forwardRef<
       newSettings.renderer = props.hotRenderer || getRenderer('text');
     }
 
-    if (props.editor) {
+    if (isComponentEditor(props.editor)) {
```
Haxikowy, line 139: `isComponentEditor` short-circuits before `resolveEditorSetting` ever sees `hotEditor`, so `<HotTable editor={MyComponent} hotEditor={false} />` still opens the component editor. Same as develop, so not a regression, but `.changelogs/7561.json` says `hotEditor={false}` on `HotTable` no longer leaves the cell editable, and in this combination it does.

Narrow the changelog title, or add a line to the helper JSDoc saying a component `editor` outranks `hotEditor={false}`.

wrappers/react-wrapper/src/hotColumn.tsx
```diff
@@ -66,10 +68,18 @@ const HotColumn: FC<HotColumnProps> = (props) => {
         columnSettings.renderer = props.hotRenderer;
```
Haxikowy, line 68: `renderer`/`hotRenderer` keeps the exact falsy-check shape this PR removes one branch below. `<HotColumn renderer={false} />` leaks `renderer: false` into column settings through the mapper, and `hotRenderer={true}` reaches core as a bare boolean.

Not asking for it here. Someone who reads the changelog will try the renderer props next, so either a follow-up ticket or one sentence in the entry saying the renderer props are unchanged.

.changelogs/7561.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed a bug where setting the `editor` or `hotEditor` prop to `false` on `HotTable`, or the `hotEditor` prop to `false` on `HotColumn`, left the cell editable and still opened the default text editor.",
+  "type": "fixed",
+  "issueOrPR": 7561,
+  "breaking": false,
```
Haxikowy, line 6: On develop, `editor={false}` on `<HotTable>` fell through to `getEditor('text')` and cells stayed editable. It now disables editing across the grid.

`editor?: ComponentType | boolean` means `editor={cond && MyEditor}` type-checks and renders `false`, so that idiom goes from a working grid to a locked one with no console warning and no type error.

That is row 3 of `.ai/BREAKING-CHANGES.md`, changing the behavior of a public API. Needs a maintainer decision on the flag, and a migration note either way. The rewrite for users is `editor={cond ? MyEditor : undefined}`.

.changelogs/7561.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed a bug where setting the `editor` or `hotEditor` prop to `false` on `HotTable`, or the `hotEditor` prop to `false` on `HotColumn`, left the cell editable and still opened the default text editor.",
```
Haxikowy: Two things on the entries.

The repro in #7561 is `<HotColumn editor={false} />`, which already worked on develop, and this title deliberately excludes it. Anyone arriving from the issue will not find their case in the release note. Add a clause confirming it behaves as asked, or attribute the entry to the PR instead.

`.changelogs/README.md` and the changelog-creation skill both want one entry per PR. These two overlap in wording and land in the same `fixed` section. Folding them into `13268.json` reads better.

wrappers/react-wrapper/test/hotColumn.spec.tsx
```diff
@@ -388,6 +388,242 @@ describe('Editor configuration using React components', () => {
...
+                }}>
+        <HotColumn editor={true} />
+        <HotColumn hotEditor={true} />
+        <HotColumn />
+      </HotTable>
+    )).hotInstance!;
+
+    // Neither boolean carries an editor, so both must resolve to the default one. A bare `true` also
+    // must never reach the core, which throws on anything but a string or a constructor.
+    expect(hotInstance.getCellEditor(0, 0).EDITOR_TYPE).toBe('text');
```
Haxikowy, line 649: All three columns resolve to text here because the grid default is text, so this passes whether `true` inherits or hard-sets. Give the grid a non-text editor, or the columns a `type`, and the two `true` paths separate. See the comment on `helpers.tsx:173`.

Missing across both spec files: `<HotTable editor={false}>` with plain `<HotColumn />` children. Both table-level tests use `columns={[...]}` for the control, so `delete columnSettings.editor`, which exists so a bare column inherits a grid-level `false`, is never exercised.

wrappers/react-wrapper/test/hotColumn.spec.tsx
```diff
@@ -388,6 +388,242 @@ describe('Editor configuration using React components', () => {
...
+    // The untouched neighboring column still opens its editor, so the disabled column proves the
+    // prop had an effect rather than the test being dead.
+    expect(hotInstance.getCellEditor(0, 1).EDITOR_TYPE).toBe('text');
+
+    await act(async () => {
+      hotInstance.selectCell(0, 1);
+      simulateKeyboardEvent('keydown', 13);
+    });
+
+    expect(hotInstance.getActiveEditor()).toBeTruthy();
```
Haxikowy: `toBeTruthy()` passes for any editor object. The `getCellEditor(0, 1).EDITOR_TYPE` assertion above already carries the control, so this line reads like an assertion without being one.

```suggestion
    expect(hotInstance.getActiveEditor()!.constructor.name).toBe('TextEditor');
```
Same at hotColumn.spec.tsx:466, hotTable.spec.tsx:819 and :856.

Haxikowy (commented):
The fix is right and the helper is the right shape. Two things in the new code need a change before merge, and one claim in the description needs correcting.

**Test count.** The description says two of the new tests pass against `develop` deliberately. It is four. I checked out the branch, reverted `wrappers/react-wrapper/src/` to `develop`, and ran the two spec files: eight of twelve fail, four pass. The extra two are `should disable editing for a column with the 'editor' prop set to 'false'` and its dynamic variant, both the `<HotColumn editor={false} />` case you already flagged as previously working by accident. Consistent with your story, just miscounted.

**`editor: true` outside React.** Vue, Angular and vanilla all pass `true` straight to core, which returns it from `getCellEditor` and then throws in `getEditorInstance` on the first edit. After this PR, React is the only place where `true` works. That is a divergence introduced rather than closed, and the right site for the fix is core normalizing `true` the way it already normalizes `undefined`, not three more wrapper patches. Worth a separate ticket. Core's `editor?: string | ctor | boolean` type (core/settings.ts:112) is what makes `hotEditor={true}` type-check in the first place; `metaSchema` documents only `false`.

**Docs.** Your unchecked box is correct. Every `editor: false` in the guides goes through `columns` or `cells`, never the React prop, and `cell-editor.md` mentions the prop only with a component value. The one new rule with no user-visible home is `hotEditor` beating `editor={false}`, see the last point below.

**Pre-existing, not this PR's problem, listed so it is not rediscovered:** `updateSettings` never closes an open editor, so flipping `editor` to `false` mid-edit still commits the value (core-wide, same in vanilla); `getCellEditor`'s declared return type omits `false` while the runtime returns it; `HotColumn.vue:30` has a dead `if (assignedProps.editor)` branch that would reintroduce this exact bug if someone converted it to an `else`.

**What I ran:** the two React spec files against the branch (12 passed) and against a `develop`-reverted `src/`. I did not run the wider React suite, and no visual or E2E tier.

**One more, off-diff.** `wrappers/react-wrapper/src/types.tsx:71-72` is where the boolean semantics and the precedence rule should be written. They currently live only in `helpers.tsx` JSDoc, and `helpers` is not re-exported from `index.tsx`, so nothing reaches a user's IDE on hover. `types.tsx` ships. Two short blocks: `false` disables editing, `true` uses the default editor, `hotEditor` wins over a falsy `editor`.

wrappers/react-wrapper/src/helpers.tsx
budnix: Nice catch — dropped that block in 526836b7a6, so `true` now means "prop not passed" on both props and a typed column keeps its own editor.

wrappers/react-wrapper/src/hotColumn.tsx
budnix: Good spot — kept the inherit behavior and wrote it down in 526836b7a6: the helper JSDoc, the changelog title, and a new test for `editor={null}`.

wrappers/react-wrapper/src/hotTableInner.tsx
budnix: You are right — narrowed the changelog title and added the precedence line to the helper JSDoc in 526836b7a6.

wrappers/react-wrapper/src/hotColumn.tsx
budnix: Agreed, the renderer props have the same shape — out of scope here, so I added a clause to the changelog entry in 526836b7a6 saying they are unchanged.

.changelogs/7561.json
budnix: Fair point — we are keeping `breaking: false`: while the condition holds, the prop still gets the editor and opens it exactly as before, and `false` switching editing off is the fix itself. Added tests for both sides of that idiom in 526836b7a6.

.changelogs/7561.json
budnix: Yep, my mistake — folded them into one entry in 526836b7a6 and added a clause confirming the reported `<HotColumn editor={false} />` case works as asked.

wrappers/react-wrapper/test/hotColumn.spec.tsx
budnix: Good point — the columns carry `type="numeric"` now in 526836b7a6, and I added the missing bare `<HotColumn/>` under a grid-level `editor={false}`.

wrappers/react-wrapper/test/hotColumn.spec.tsx
budnix: Good call — swapped all four for the exact `TextEditor` name check in 526836b7a6.

Haxikowy (approved):

Issue #7585, opened 2021-02-23
Height of table incorrectly calculated under zoom


john-swallow-cenduit:
### Description
Height of table container is incorrectly calculated under different zoom levels leading to faulty spacing and in the worst case overlap of subsequent controls. 

Possible Angular specific issue. Only observed under Angular conditions.

Observed under https://jsfiddle.net/api/post/library/pure/ with minimal mods:

- Added non-commercial license key to get rid of the message. If message is present the issue does not show up.
- Added a <button> directly under the table.

### Steps to reproduce
1. Open the demo link and modify the zoom.
2. Observe that the space between the table and the button that follows is variable.
3. Set zoom to 50% and observe that the button appears INSIDE the table.

The height of the container div does NOT align with the content of the table

![image](https://user-images.githubusercontent.com/58601615/108877900-98735800-75f7-11eb-9e87-30f8b8524ec0.png)

### Demo
https://jsfiddle.net/9zmynvfk/

### Your environment
* Handsontable version: 8.3.1. @handsontable/angular 6.0.1
* Browser Name and version: Chrome Version 88.0.4324.182 (Official Build) (64-bit)
* Operating System: Windows 10

john-swallow-cenduit:
Actually reproducible without any change.
Just go to the docs for getting started with HoT and Angular 
https://handsontable.com/docs/8.3.1/frameworks-wrapper-for-angular-simple-example.html
Click the link for "Edit" on the example. 
Then zoom to 50%. The license key warning itself overwrites the grid.

![image](https://user-images.githubusercontent.com/58601615/108885909-44b93c80-7600-11eb-8c11-9995361dbb78.png)

AMBudnik:
Thi @john-swallow-cenduit 

Thank you for sharing. We already have a couple of issues related to zoom in/out on the Github issue board. They are listed here https://github.com/handsontable/handsontable/issues/7527 I think that the first issue on that list describes your case. Please confirm.

john-swallow-cenduit:
I am well aware of #7527 having checked for existing issues before raising this one. However I don't think that this particular issue is the same as the one you reference. The first item on that item #3118 relates to misaligning rows when row headers are used. The above issue occurs regardless of whether row headers are on or off so I cannot see how it adequately covered by that item. 

[...]

Issue #6280, opened 2019-09-23
The container of table is always higher than table,'wtHider' got wrong height


eret9616:
hi, im using the latest version of handsontable,
i found the container of table is always higher than table's height ,the computed 'wtHider' height  always higher than the table's correct height,

is there any way or options to prevent this? 

![img](https://user-images.githubusercontent.com/12445095/65412552-0bd0d300-de22-11e9-8046-0264f1cde317.png)

![img2](https://user-images.githubusercontent.com/12445095/65412570-12f7e100-de22-11e9-9bda-88de29fd4c4b.png)

the 'extra bottom space' also shows in tutorial, looks no good
[hansontable tutorial](https://handsontable.com/blog/articles/2016/3/a-complete-guide-to-changing-size-of-handsontable)

edited: decrease 100px made it looks ok ,but its not 100px that much,

AMBudnik:
Hey @eret9616 

what browser version and system settings do you use (any zoom, or custom resolution)?

This is what I get on Chrome 77 / Windows 10 using this demo https://handsontable.com/docs/7.1.1/frameworks-wrapper-for-react-hot-reference.html

![image](https://user-images.githubusercontent.com/10757813/65415176-82a3a700-ddf5-11e9-9293-09f7e621dbf6.png)

The `wtHider` in this case is `31px` heigh and it covers only the header.
![image](https://user-images.githubusercontent.com/10757813/65415324-d1e9d780-ddf5-11e9-87f0-567a6eeaa839.png)

eret9616:
Chrome  77.0.3865.90
Windows 10 enterprise LTSC  10.0.17763.737  scale 125%

oh, scale 100% looks ok.

AMBudnik:
It still doesn't change anything on my device.
Can you check if you have set any other resolution than required (for your device) and is there zoom in the browser settings?

This is how it looks on 125% with a resolution of 1920x1080

![image](https://user-images.githubusercontent.com/10757813/65507940-552b2c00-decf-11e9-8488-2cb4200e4b4c.png)

and `wtHider`

![image](https://user-images.githubusercontent.com/10757813/65507965-66743880-decf-11e9-9029-3e69b1850729.png)

Could you send me the demo that you're using?

AMBudnik:
Thank you. I have just updated a related topic https://github.com/handsontable/handsontable/issues/3118
I think that we should also check this one while fixing https://github.com/handsontable/handsontable/issues/3118

[...]

PR #13269, opened 2026-08-27
DEV-2657: Fix the grid outgrowing its container below 100% zoom

budnix:
### Context

The PR fixes a grid with no `height` set rendering taller than the container it reports, whenever the browser draws below 100% — a zoom level under 100%, or Windows display scaling under 100%. The bottom rows spill onto whatever follows the grid, and the row-header column cuts off its own last row numbers.

**Why it happens.** A browser cannot paint a border thinner than one physical pixel. Below 100% it inflates the cells' 1px bottom border — 1.111px at 90%, 1.25px at 80% — and the line box and padding do not shrink to absorb it, so every row renders that fraction taller than the theme declared. `StylesHandler#calculateRowHeight()` summed the declared height, so the shortfall accumulated **per row**: ~0.1px/row at 90%, ~0.22px/row at 80%.

**The fix.** `#calculateRowHeight()` now reads the cells' unrounded `border-bottom-width` and, only when it resolved **wider than its own rounded value**, replaces the declared row height with a measured one. Measured rather than derived, because it also carries the browser's sub-pixel snapping of the cell box, which no arithmetic here reproduces (the arithmetic answer is 29.111px against a rendered 29.097px, and that 0.014px lands on every row).

That gate is the load-bearing part — it keeps the correction off every case that was already correct. Two of its edges cost me a regression each during development, and both now have tests:

- **Cells with no border.** The cells' height rule is `calc(<padding> * 2 + <line-height> + 1px)` with a *literal* 1px, which the measurement probe inherits whatever the real cells carry. The Filters by-value list sets `border-bottom-width: 0`, so reading the probe's absolute height reported 29px for rows rendering at 28px and sized that list's scroll range past its last item.
- **A border that lands on a whole number of pixels.** At 50% zoom it resolves to exactly 2px, so rounding is already correct and 18.0.0 gets that range right. Correcting there anyway added ~0.5px to every row — 49.5px over 100 rows, a worse defect than the one being fixed.

### Results

Measured on an auto-height grid, 100 rows, all rendered, identical across all three themes:

| zoom | reported border | 18.0.0 | this PR |
|---|---|---|---|
| 100% | 1px | 0 | 0 |
| 90% | 1.111px | **8.9px off** | 0.20px |
| 80% | 1.25px | **17.9px off** | 0.38px |
| 67% | 1.493px | **31.3px off** | 0.36px |
| 50% | 2px | 0.5px | 0.5px |
| 25% | 4px | 0.25px | 0.25px |

At 500 rows the pre-fix figures grow to −49px (90%) and −110px (80%); after the fix they stay flat, because the remainder is a constant rather than a per-row error. **Above 100% is byte-identical to before.** The row-header clipping goes 22.6px → 0.72px, and the last row's number is visible again.

Two things deliberately left alone, both sub-device-pixel and non-accumulating: a constant 0.2–0.5px remainder (the first-row `border-top` +1 and `hiderHeightComp` do not track their own inflation), and a ~1.5px residual at 33% that is **identical on 18.0.0 and on this branch**, so it is pre-existing and outside this change.

### Scope

Only grids with **no `height` option** are affected — the container's height is then the sum of the row heights rather than a number the page sets. With an explicit height the container measures exactly right at every scale, on every version. That is the workaround for anyone blocked before this ships.

**The correction is established when the grid is built, and retaken on a theme change or an explicit `clearCache()` — it does not follow a runtime zoom change.** The gate reads `border-bottom-width` from the `#computedStyles` snapshot, so a grid loaded at 100% and zoomed out afterwards keeps the declared height and overflows by the same amount it did before this change: measured at 8.94px on the 100-row fixture at 90%, identical on `develop`. Loading at 90% gives 0.20px. Not a regression, and thanks to @adrianspdev for catching that the earlier wording here overclaimed it.

Following a runtime zoom needs more than re-reading the border: with the row height corrected by hand, the grid still renders the old sum until the row-height cache **and** the hider sizing are invalidated too (`clearCache()` + `invalidateRowHeightCache()` + `render()` left the overflow at 8.94px). That belongs in its own change rather than this one.

Above 100% on a high-DPI display is untouched for the same reason and was already correct: at a forced DPR of 1.8 the 1px border computes to 0.555556px, the gate stays off, and the −1.6px residual there is byte-identical on `develop`.

### How has this been tested?

New unit coverage (7 cases) for the reconciliation and each edge of the gate:

```bash
npm run test:unit --prefix handsontable -- --testPathPattern=stylesHandler
```

New Playwright E2E — fixture, page object, and spec — running on all six theme × bundle legs:

```bash
cd tests && npx playwright test row-height-device-scale
```

The E2E is driven by **CSS zoom on the root element**, applied before the grid is built. That is deliberate: Playwright's context-level `deviceScaleFactor` reports the ratio faithfully but never inflates the border, so a test built on it passes on unfixed code; and `--force-device-scale-factor` needs `viewport: null`, which every project's `devices['Desktop Chrome']` forbids by pinning `deviceScaleFactor`. Each describe block asserts the inflated border as its own precondition, so the suite cannot go quietly vacuous.

**Verified the spec fails without the fix** — reverting `stylesHandler.ts` to develop and rebuilding turns exactly the three sub-100% assertions red (−8.94px), while the precondition and the 100% control stay green.

Full suites on this branch's base:

```bash
npm run test:unit --prefix handsontable          # 3971 passed
npm run test:types --prefix handsontable         # clean
npm run stylelint --prefix handsontable          # clean
cd tests && npx playwright test --project=e2e-main   # 473 passed, 1 skipped
npm run test:e2e --prefix handsontable -- --testPathPattern=filters       # 283 specs, 0 failures
npm run test:e2e --prefix handsontable -- --testPathPattern=autoRowSize   # 55 specs, 0 failures
```

The two legacy slices are the ones this change can reach: the Filters by-value list is the borderless-cell case, and AutoRowSize owns row measurement when `externalRowCalculator` is on. Manual verification was done in Chrome against a released-vs-local demo page at 100 / 90 / 80 / 67 / 50 / 25%.

### Types of changes

### Related issue(s):

1. DEV-2657
2. #6280 — reported 2019 for the above-100% direction (fixed since the 15.x line), reopened in April 2023 for the below-100% direction this PR fixes.
3. #7585 — the same defect, below-100% only. Closed in 2021 without a fix.

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2657

---

> [!NOTE]
> **Medium Risk**
> Changes default row height during every draw for auto-height grids and adds mid-render DOM probing, but behavior is tightly gated and heavily tested; wrong gating would affect Filters lists or hidden containers.
> 
> **Overview**
> Fixes **#6280**: auto-height grids (no explicit `height`) grew taller than their container when zoom or display scaling was below 100%, so bottom rows overlapped content below and row-header clones clipped the last row numbers.
> 
> **`StylesHandler#getDefaultRowHeight`** no longer relies only on theme CSS variables plus a rounded border. When the computed `border-bottom-width` is **wider than its rounded value** (sub-100% device-pixel inflation), it **measures** a probe `td` (with `htCore` and a non-breaking space) and uses that rendered height, with a lazy cache keyed on declared height and `devicePixelRatio`. The gate deliberately skips borderless surfaces (e.g. Filters by-value list), whole-pixel borders (50% zoom), and hidden/off-DOM probes so those paths stay unchanged.
> 
> **Walkontable** `markOversizedRows` replaces an exact height equality fast path with a **&lt;1px** tolerance so fractional default row heights still skip the per-row walk.
> 
> Adds **unit** coverage for reconciliation edge cases, **Playwright** E2E (CSS `zoom` on the document root at 90%, 50%, and 100%), fixture/page object, changelog entry, and **AGENTS.md** notes on probes and `StylesHandler` caches.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a4f8148ea2b0f78112b7c31f65cf061dfcd62f20. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2657 Fix the grid rendering taller than its own container below 100% zoom](https://app.clickup.com/t/9015210959/DEV-2657)

handsontable/src/utils/stylesHandler.ts
```diff
@@ -224,10 +236,11 @@ export class StylesHandler {
   #calculateRowHeight() {
     const lineHeightVarValue = this.getCSSVariableValue('line-height');
     const verticalPaddingVarValue = this.getCSSVariableValue('cell-vertical-padding');
+    const renderedBottomBorderWidth = Number.parseFloat(this.getStyleForTD('border-bottom-width') ?? '');
```
adrianspdev, line 239: Hi @budnix — nice work on the gate; the load-time case checks out for me at 90/80/75/67% (0.2–0.66px residual vs 9–31px on 18.0.0), and the Filters by-value list is unaffected. One thing I hit while testing that I think is worth a decision before merge:

**The gate reads `border-bottom-width` from the `#computedStyles` snapshot, so a grid built at 100% never picks up a later zoom-out.**

`renderedBottomBorderWidth` comes from `getStyleForTD()`, and that snapshot only refreshes in `useTheme()` / `clearCache()`. There is no resize or DPR listener that re-runs `#cacheStylesheetValues()`. So:

1. Load the grid at 100% → snapshot is `"1px"`.
2. Zoom to 90% (CSS zoom in my demo, but the same applies to real Cmd/Ctrl− — nothing re-snapshots either way).
3. `renderedBottomBorderWidth > bottomBorderWidth` is `1 > 1` → false → `getDefaultRowHeight()` stays `29`.

Measured on your fixture shape (100 rows, auto height): overflow is 8.94px — identical to 18.0.0. `hot.render()`, `updateSettings({ colHeaders: true })`, and `updateSettings({ themeName: 'ht-theme-main' })` (same theme) do not change it. Only `stylesHandler.clearCache()` refreshes the gate, and even then `render()` alone left the hider at 8.94px because the viewport row-height cache still held the old default — it also needs `invalidateRowHeightCache()`.

The `devicePixelRatio` key on `#measureRenderedRowHeight()` never gets a chance here, because the gate sits in front of it. The PR description and the new `handsontable/AGENTS.md` bullet say the cache "re-measures after both a late-arriving theme and a zoom change" — that holds only when the grid was *already* below 100% at init (fractional snapshot, gate passes, DPR key re-measures). The 100% → 90% direction is the more common one for a user reaching for Ctrl−.

Not a regression — 18.0.0 has the same overflow there — so I'm fine with either route:

- **(a)** Move the border read into the lazily-cached path and key it on `devicePixelRatio` alongside the height. Real browser zoom moves DPR, so the border gets re-read exactly when it can change. CSS `zoom` on a wrapper would still be missed, which I think is acceptable to document. Whatever triggers the re-read must also invalidate the viewport row-height cache, or the hider keeps the old sum.
- **(b)** Ship as is and correct the wording in the PR description and `AGENTS.md` to say the correction is established at init / theme change and does not follow a runtime zoom change.

Repro page if useful: an auto-height grid with `?zoom=` applied to `<html>` before init plus buttons that change `documentElement.style.zoom` at runtime and call `hot.render()`; the readout is `lastRow.getBoundingClientRect().bottom - wtHider.getBoundingClientRect().bottom`. Happy to push the demo somewhere if you want it.

One side question, no action needed for this PR: did you get a chance to try a real retina display at 90%? Chrome floors border widths to whole device pixels, so at DPR 1.8 I'd expect the 1px border to compute to ~0.556px, which keeps the gate off and leaves the pre-existing overshoot direction. I could not check it here — CDP `deviceScaleFactor` emulation leaves the snapping untouched, as you noted for Playwright.

adrianspdev (commented):
Manually tested against a released-vs-local demo at 100/90/80/75/67/60/50/33/25% (DPR 1). Load-time fix works as described; one runtime-zoom limitation inline.

handsontable/src/utils/stylesHandler.ts
budnix: Good point — the gate reads the snapshot, so a grid loaded at 100% keeps the old height; I went with (b), fixed the wording in 5d1a4fc2a9, and can file (a) as a follow-up. On the retina question: at a forced DPR of 1.8 the border computes to 0.555556px just as you predicted, so the gate stays off and the overshoot there is identical on 18.0.0.

adrianspdev (approved):

PR #13272, opened 2026-08-27
DEV-2652: Keep an open editor bound to its record when rows are trimmed

demtario:
### Context

The PR fixes silent source-data corruption when an index map rearranges rows while a cell editor is open.

The editor holds the visual coordinates captured in `prepare()`, and `BaseEditor#saveValue()` writes through them with no bounds check. Two kinds of index-map change invalidate them:

- a **trimming** map (Filters, `trimRows`, `nestedRows`) *collapses* the visual space — rows below a trimmed row shift up and the row count shrinks;
- a **sequence** change (`columnSorting`, `manualRowMove`) *permutes* it.

Either way the pending edit lands on whichever record now occupies that visual slot. Three shapes, all silent:

| Case | Condition | Result |
|---|---|---|
| Wrong record | edited visual row `<=` post-change `countRows() - 1` | the typed value overwrites a different physical record |
| Appended rows | edited visual row `>` post-trim `countRows() - 1` | `applyChanges()` **appends records** to reach the stale coordinate |
| Shifted record | the edited row survives but moves | the edit lands on a different surviving record, with no row-count change |

Traced path, for a filter:

1. `Filters#filter()` bulk-writes its trimming map (`filters.ts:1020`). The visual row space collapses here, and the flush fires `afterRowSequenceCacheUpdate` with `{ trimmedIndexesChanged: true }` synchronously.
2. Still inside `filter()`, `filters.ts:1045` re-selects the highlighted column. That selection change reaches `editorManager.closeEditor()` through `core.ts:802-804` and commits the edit — which is why the editor is already `STATE_VIRGIN` before the render, and why `afterViewRender` cannot catch this.
3. `BaseEditor#saveValue()` passes the stale `this.row` into `populateFromArray(..., 'edit')` with `endRow: null` (`baseEditor.ts:271`). `null` is not `typeof 'number'`, so `end` is `undefined` in `grid.populateFromArray` and the row loop's only clamp is dead at default settings.
4. `applyChanges()` grows the dataset (`core.ts:2032`):

```js
while (changes[i][0] > instance.countRows() - 1) {
  const missingRows = changes[i][0] - (instance.countRows() - 1);
  datamap.createRow(undefined, missingRows, { source: 'auto' });  // undefined index => append
}
```

Worked example matching the reported output: 5 rows, filter keeps only `A2`, editor on visual row 3. Post-trim `countRows() === 1`, so `missingRows = 3 - 0 = 3`; the source data becomes 8 records; the visible rows are then physical `2, 5, 6, 7`, so visual row 3 resolves to physical 7 — `["A0","A1","A2","A3","A4", null, null, "EDITED"]`.

`trimRows()` and a sort reach the same broken state by a different route: neither touches the selection nor renders, so nothing commits at the time — the stale editor is simply left armed for whatever the user does next.

#### The fix

Entirely in `EditorManager`, keyed on the row/column mapper's own change signal so it covers every producer at once:

- `prepareEditor()` captures the **physical** row/column of the record being edited; `clearActiveEditor()` clears it.
- `#reconcileEditorWithIndexMaps()` runs on `afterRowSequenceCacheUpdate` / `afterColumnSequenceCacheUpdate` when either `trimmedIndexesChanged` or `indexesSequenceChanged` is set, and resolves that physical index back to a visual one:
  - no visual index (record gone) — discard via `cancelChanges()` and release the editor, writing nothing;
  - a different visual index (record moved) — rebind, so the pending commit still lands on the right record;
  - the same visual index — no-op.
- It runs **synchronously**, unlike `#closeEditorWhenCellHidden()`, because `filters.ts:1045` would beat a deferred handler. Neither branch writes data or renders, so nothing re-enters a cache update that is still unwinding.
- The now-false "a TRIMMING map is deliberately out of scope" paragraph in `#closeEditorWhenCellHidden()`'s JSDoc (added by #13245) is corrected.

Comparing the *resolved visual index* rather than testing a flag is what keeps this from firing spuriously: `BooleanMap#setValues()` emits a change even for a no-op write, so a flag test alone would tear down unrelated edits.

`core.ts:2032` is deliberately left alone. Growth past the last row is load-bearing for pasting below the last row, and narrowing it would be a breaking change. It is the amplifier here, not the cause.

A **structural** change — `alter()` inserting or removing rows — is the one case this reasoning gets backwards: it renumbers the physical space, so the captured index goes stale while the editor's *visual* coordinate stays correct. It raises the same flags as a filter, so the state object cannot tell them apart. The **size of the physical space** can: only an insert or a remove changes it. `#onSequenceCacheUpdate()` compares it against the previous update and routes structural changes to `#recaptureEditedRecord()`, which re-derives the record from the visual coordinate and discards only when that coordinate no longer resolves.

Without that split, any grid with a trimming map registered discarded a valid edit on `alter('remove_row', ...)` and, with a sort active, rebound onto the wrong record on `alter('insert_row_above', ...)`. Both are covered by cases.

Discriminating on the count rather than on `beforeCreateRow`/`afterCreateRow` pairing also means nothing a plugin cancels can strand it — `Formulas` returns `false` from those hooks whenever HyperFormula rejects the change, and a cancelled insert fires the `before` hook with no `after` hook and no cache update behind it. A cancelled insert changes no counts, so there is nothing to arm and nothing to leave armed.

#### Known limits, stated in the JSDoc

- An index-map change does **not** adjust the selection (`core.ts` calls `selection.commit()` only for `hiddenIndexesChanged`), so the highlight can be left past the last row and typing into it grows the data set. That is reachable **with no editor involved at all** — verified with a probe that never opens an editor — and is a separate pre-existing defect this PR does not paper over. Worth its own ticket.
- An editor parked in `WAITING` is not reconciled: `finishEditing()` has already run `saveValue()` by then, so there is nothing left to redirect.
- No core plugin registers a *trimming* map on the **column** axis, so that half runs for user-registered maps only. Core plugins do permute the column sequence (`manualColumnMove`, `manualColumnFreeze`), which the reconciliation now follows.
- A rebind moves the editor's coordinates and nothing else — not its pixel position, not the selection. Neither `render()` nor `view.render()` repositions an open editor, so on the `trimRows` path it stays drawn over the row it started on.
- Because the selection does not follow, two commit paths do not deliver the guarantee: an editor whose `finishEditing()` vetoes on a moved range (`DropdownEditor`, and the `date` / `autocomplete` / `handsontable` types built on it), and a Ctrl+Enter commit, which reads the selection corners rather than the editor's coordinates. On both the edit is **lost rather than misplaced** — `develop` appended rows in the dropdown case — but the value does not survive. Making it survive means moving the selection with the record, which is a larger change than this repair. Both are pinned by cases.
- `getActiveEditor()` now returns a `STATE_VIRGIN` editor at points where it previously returned an editing one, and a discarded edit fires no `beforeChange`/`afterChange`.

### How has this been tested?

New Playwright suite, `tests/e2e/editor-trimmed-row.spec.ts` (20 cases) with `tests/fixtures/pages/EditorTrimmedRowPage.ts` and `tests/fixtures/demo/editor-trimmed-row.html`. There was no existing test anywhere that opens an editor and then trims its row.

Cases: the append shape; the wrong-record shape; the shifted-record shape; both `trimRows` routes; a trimming-map write that moves nothing; an editor whose cell was scrolled out of view before the filter ran; `alter('remove_row')` stranding the editor past the last row; a sorted index mapping under a trim (so visual and physical genuinely differ); re-preparation from post-trim state when the user keeps typing after a discard; a sort and a `manualRowMove` under an open editor; and two structural-change cases (`alter()` removing an unrelated row, and inserting one) that only fail with a permutation active. Every case asserts the whole source data set **and** `countSourceRows()`, plus recorders proving the trigger actually fired and that no commit happened where none should.

13 of the 20 cases fail against `develop`. The other 7 are **regression guards, not proofs** — they pin behavior `develop` already gets right (a removal above the edit, a removal below it, an insert, a no-op trimming write, a Ctrl+Enter commit, a `updateData()` swap, a column removal) and exist because earlier revisions of this change broke several of them. They are labelled as such in their own comments.

```bash
# the new suite, plus the hiding-map suite from #13245 to prove it still holds
cd tests && npx playwright test --project=e2e-main \
  e2e/editor-trimmed-row.spec.ts e2e/editor-hidden-cell.spec.ts    # 33 passed

# core regression sweep - the guard now runs on every index-map change
npm --prefix handsontable run test:unit    # 3963 passed
npm --prefix handsontable run test:e2e     # 9745 specs, 0 failures
```

### Types of changes

### Related issue(s):

1. DEV-2652

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2652

---

> [!NOTE]
> **Medium Risk**
> Core edit/commit path in `EditorManager` now runs synchronously on every row/column index-map update; behavior is heavily tested but regressions could affect any grid with filters, trimming, or sorting during inline edit.
> 
> **Overview**
> Fixes **silent source-data corruption** when Filters, `trimRows`, sorting, row/column moves, or structural `alter()` run while a cell editor is open. The editor only kept **visual** coordinates, so commits could hit the wrong row, **append rows**, or write past a trimmed record.
> 
> **`EditorManager`** now stores the edited cell’s **physical** row/column in `prepareEditor()`, clears it when the editor is cleared, and on `afterRowSequenceCacheUpdate` / `afterColumnSequenceCacheUpdate` runs **`#repairEditor`** before the existing hidden-cell close logic. **Rearrangements** (trim/sequence) go through **`#reconcileEditorWithIndexMaps`**: drop the edit if the record left the visual grid, **rebind** `editor.row`/`col` if it moved, no-op if nothing changed (so no-op trimming-map writes do not kill edits). **Structural** insert/remove is detected by a change in physical index count and handled by **`#recaptureEditedRecord`** instead, so row insert/remove does not mis-reconcile against a stale physical index; a short **`#strandedInCurrentTask`** window avoids discarding edits during the gap between `alter()`’s cache update and `selection.shiftRows()`.
> 
> Adds changelog entry **#13272** and a large Playwright suite (`editor-trimmed-row.spec.ts` plus fixture/page object) covering filters, `trimRows`, sorts, moves, structural changes, column axis, `updateData`, and edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d85906ab9194ff5965d798a9022dc5b9a5352378. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changelogs/13272.json
+++ b/.changelogs/13272.json
@@ -1,6 +1,6 @@
 {
   "issuesOrigin": "private",
-  "title": "Fixed an in-progress cell edit landing on the wrong record, or appending rows to the source data, when rows were filtered, trimmed, sorted, or moved while the editor was open. The edit now follows its own record, and is discarded when that record can no longer be reached.",
+  "title": "Fixed an in-progress cell edit landing on the wrong record, or appending rows to the source data, when rows or columns were filtered, trimmed, sorted, moved, inserted, removed, or replaced while the editor was open. The edit now follows its own record, and is discarded when that record can no longer be reached.",
   "type": "fixed",
   "issueOrPR": 13272,
   "breaking": false,
--- a/.changelogs/13272.json
+++ b/.changelogs/13272.json
@@ -1,6 +1,6 @@
 {
   "issuesOrigin": "private",
-  "title": "Fixed an in-progress cell edit landing on the wrong record, or appending rows to the source data, when a filter or `trimRows` removed rows while the editor was open.",
+  "title": "Fixed an in-progress cell edit landing on the wrong record, or appending rows to the source data, when rows were filtered, trimmed, sorted, or moved while the editor was open. The edit now follows its own record, and is discarded when that record can no longer be reached.",
   "type": "fixed",
   "issueOrPR": 13272,
   "breaking": false,
--- a/.changelogs/13272.json
+++ b/.changelogs/13272.json
@@ -1,6 +1,6 @@
 {
   "issuesOrigin": "private",
-  "title": "Fixed an in-progress cell edit landing on the wrong record, or appending rows to the source data, when rows or columns were filtered, trimmed, sorted, moved, inserted, removed, or replaced while the editor was open. The edit now follows its own record, and is discarded when that record can no longer be reached.",
+  "title": "Fixed an in-progress cell edit landing on the wrong record, or appending rows to the source data, when rows or columns were filtered, trimmed, sorted, moved, inserted, removed, or replaced while the editor was open. The edit now follows its own record, and is discarded when a filter or trim takes that record out of view.",
   "type": "fixed",
   "issueOrPR": 13272,
   "breaking": false,
--- a/.changelogs/13272.json
+++ b/.changelogs/13272.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Fixed an in-progress cell edit landing on the wrong record, or appending rows to the source data, when a filter or `trimRows` removed rows while the editor was open.",
+  "type": "fixed",
+  "issueOrPR": 13272,
+  "breaking": false,
+  "framework": "none"
+}
--- a/handsontable/src/editorManager.ts
+++ b/handsontable/src/editorManager.ts
@@ -86,10 +86,36 @@ class EditorManager {
    */
   #editedPhysicalColumn: number | null = null;
   /**
+   * Whether a structural change - a row or column being inserted or removed - is currently running.
+   *
+   * Such a change RENUMBERS the physical space, which invalidates `#editedPhysicalRow` while leaving
+   * the editor's visual coordinate correct. That is the opposite of what a trim or a permutation
+   * does, and the two are indistinguishable from the cache-update state alone, so the reconciliation
+   * stands down while this is set and the captured record is re-derived once the change lands.
+   *
+   * @type {boolean}
+   */
+  #structuralChangeInProgress = false;
+  /**
+   * Marks the start of a structural change, so the cache update it emits does not reconcile against
+   * a physical index that the change is in the middle of renumbering.
+   */
+  #onBeforeStructuralChange = (): void => {
+    this.#structuralChangeInProgress = true;
+  };
+  /**
+   * Marks the end of a structural change and re-derives the edited record from the editor's visual
+   * coordinate, which the change kept valid.
+   */
+  #onAfterStructuralChange = (): void => {
+    this.#structuralChangeInProgress = false;
+    this.#recaptureEditedRecord();
+  };
+  /**
    * Reacts to an index-map cache update on either axis.
    *
-   * The index-map reconciliation runs FIRST so that the hidden-cell guard tests `isHidden()` against
-   * rebound coordinates rather than the stale ones the index map just invalidated.
+   * The trimming reconciliation runs FIRST so that the hidden-cell guard tests `isHidden()` against
+   * rebound coordinates rather than the stale ones a trimming map just invalidated.
    *
    * @param {object} indexesChangesState The state object of the index mapper's cache update.
    * @param {boolean} indexesChangesState.indexesSequenceChanged Whether the indexes sequence changed.
@@ -118,6 +144,14 @@ class EditorManager {
     this.hot.addHook('beforeCompositionStart', (event: KeyboardEvent) => this.#onAfterDocumentKeyDown(event));
     this.hot.addHook('afterRowSequenceCacheUpdate', this.#onSequenceCacheUpdate);
     this.hot.addHook('afterColumnSequenceCacheUpdate', this.#onSequenceCacheUpdate);
+    this.hot.addHook('beforeCreateRow', this.#onBeforeStructuralChange);
+    this.hot.addHook('beforeRemoveRow', this.#onBeforeStructuralChange);
+    this.hot.addHook('beforeCreateCol', this.#onBeforeStructuralChange);
+    this.hot.addHook('beforeRemoveCol', this.#onBeforeStructuralChange);
+    this.hot.addHook('afterCreateRow', this.#onAfterStructuralChange);
+    this.hot.addHook('afterRemoveRow', this.#onAfterStructuralChange);
+    this.hot.addHook('afterCreateCol', this.#onAfterStructuralChange);
+    this.hot.addHook('afterRemoveCol', this.#onAfterStructuralChange);
     this.hot.view._wt.update(
       'onCellDblClick', (event: MouseEvent, coords: { isCell: () => boolean }, _elem: HTMLElement) =>
         this.#onCellDblClick(event, coords)
@@ -142,6 +176,10 @@ class EditorManager {
     // `beforeChange` that cancels the change, for instance) would disable the guard for the rest
     // of the instance's life rather than for that one edit.
     this.#hiddenCellCloseArmed = false;
+    // A vetoed `beforeCreateRow`/`beforeRemoveRow` never reaches its `after` counterpart, which
+    // would leave the latch set. Nothing changed in that case, so the only cost is the guard being
+    // inert for the rest of that edit - and a new edit cycle clears it.
+    this.#structuralChangeInProgress = false;
 
     if (this.activeEditor && this.activeEditor.isWaiting()) {
       this.closeEditor(false, false, (dataSaved: boolean) => {
@@ -384,24 +422,64 @@ class EditorManager {
   }
 
   /**
+   * Re-derives the edited record after a structural change has renumbered the physical space.
+   *
+   * Inserting or removing rows shifts the physical indexes of everything below, so the index captured
+   * in `prepareEditor()` now addresses a different record - but the editor's VISUAL coordinate came
+   * through intact, because the index mapper moved the visual space with it. So the visual side is
+   * the trustworthy one here, and it is what the record is read back from.
+   *
+   * A visual coordinate that no longer resolves means the change left the editor past the last row.
+   * There is nothing to commit to, so the edit is dropped rather than written through a coordinate
+   * that `applyChanges()` would satisfy by appending records.
+   */
+  #recaptureEditedRecord(): void {
+    const editor = this.activeEditor;
+
+    if (!editor || editor.state !== EDITOR_STATE.EDITING || editor.row === null || editor.col === null) {
+      return;
+    }
+
+    const physicalRow = this.hot.rowIndexMapper.getPhysicalFromVisualIndex(editor.row);
+    const physicalColumn = this.hot.columnIndexMapper.getPhysicalFromVisualIndex(editor.col);
+
+    if (physicalRow === null || physicalColumn === null) {
+      editor.cancelChanges();
+      this.clearActiveEditor();
+
+      return;
+    }
+
+    this.#editedPhysicalRow = physicalRow;
+    this.#editedPhysicalColumn = physicalColumn;
+  }
+
+  /**
    * Keeps an open editor bound to the RECORD it was opened on when an index map rearranges the
    * visual index space underneath it.
    *
    * Two kinds of change do that. A TRIMMING map (Filters, `trimRows`, `nestedRows`) COLLAPSES the
-   * visual space - the rows below a trimmed row shift up and the row count shrinks. A SEQUENCE
-   * change (`columnSorting`, `manualRowMove`) permutes it. Either way the editor is left holding the
-   * visual coordinates it captured in `prepare()`, and `BaseEditor#saveValue()` writes straight
-   * through them with no bounds check, so the pending edit lands on whichever record now occupies
-   * that visual slot - or, when a trim left the slot past the shortened row count, on rows that
-   * `applyChanges()` APPENDS to the source data to make room for it. Both are silent data corruption.
+   * visual space - the rows below a trimmed row shift up and the row count shrinks. A SEQUENCE change
+   * (`columnSorting`, `manualRowMove`, `manualColumnFreeze`) permutes it. Either way the editor is
+   * left holding the visual coordinates it captured in `prepare()`, and `BaseEditor#saveValue()`
+   * writes straight through them with no bounds check, so the pending edit lands on whichever record
+   * now occupies that visual slot - or, when a trim left the slot past the shortened row count, on
+   * rows that `applyChanges()` APPENDS to the source data to make room for it. Both are silent data
+   * corruption.
    *
    * Resolving the stored PHYSICAL index back to a visual one covers every shape with one test: the
-   * record is gone (no visual index - discard), the record moved (rebind - the edit still commits,
-   * to the right record), or nothing moved (no-op). The last branch is what keeps this from firing
-   * spuriously: a row insert churns both collections, and `BooleanMap#setValues()` emits a change
-   * even for a no-op write, so testing "something changed" alone would tear down unrelated edits.
-   * An insert or removal shifts the visual and physical spaces together, so the resolution returns
-   * the index the editor already held and the no-op branch takes it.
+   * record is gone (no visual index - discard), the record moved (rebind - the edit still commits, to
+   * the right record), or nothing moved (no-op). The last branch is what keeps this from firing
+   * spuriously: `BooleanMap#setValues()` emits a change even for a no-op write, so testing "something
+   * changed" alone would tear down unrelated edits.
+   *
+   * A structural change - an inserted or removed row or column - is the one case this reasoning does
+   * NOT cover, because it renumbers the physical space and invalidates the captured index while
+   * leaving the visual coordinate correct. It raises the same flags and cannot be told apart from the
+   * state object, so `#onBeforeStructuralChange()` stands this method down for the duration and
+   * `#recaptureEditedRecord()` re-derives the record afterwards. Without that, a grid with any
+   * trimming map registered would discard a valid edit on `alter('remove_row', ...)` and, with a sort
+   * active, rebind onto the wrong record on `alter('insert_row_above', ...)`.
    *
    * Runs SYNCHRONOUSLY, unlike `#closeEditorWhenCellHidden()`. `Filters#filter()` re-selects the
    * highlighted column immediately after writing its map, which commits the open editor before any
@@ -414,16 +492,21 @@ class EditorManager {
    * rewrites the restore flag - which is harmless here because a discard is what that override would
    * decide anyway once the edited record is gone from the visual space.
    *
-   * The editor keeps its on-screen position until the next render. That is cosmetic and, on the
-   * Filters path, invisible: `filter()` renders before yielding.
+   * A rebind moves the editor's coordinates, NOT its pixel position or the selection - neither
+   * `render()` nor `view.render()` repositions an open editor, so it stays drawn over the row it
+   * started on for the rest of the edit. On the Filters path that is invisible because `filter()`
+   * closes the editor outright, but on the `trimRows` path the editor is left painted over a
+   * neighbouring row. The commit still lands on the right record; only the position is wrong.
    *
-   * Three limits, all deliberate. A trimming change does NOT adjust the selection - `core.ts` calls
+   * Three limits, all deliberate. An index-map change does NOT adjust the selection - `core.ts` calls
    * `selection.commit()` only for `hiddenIndexesChanged` - so the highlight can be left past the last
    * row, and typing into it grows the data set. That is reachable with no editor involved at all and
-   * is a separate defect; this method does not paper over it. An editor parked in `WAITING` is not
-   * reconciled either, because `finishEditing()` has already run `saveValue()` by then and there is
-   * nothing left to redirect. And no core plugin registers a trimming map on the COLUMN axis - only
-   * `rowIndexMapper` - so the column half runs for user-registered maps only.
+   * is a separate defect; this method does not paper over it. No core plugin registers a TRIMMING map
+   * on the column axis, so the column half of that case runs for user-registered maps only, though
+   * core plugins do permute the column sequence (`manualColumnMove`, `manualColumnFreeze`).
+   *
+   * And an editor parked in `WAITING` is not reconciled: `finishEditing()` has already run
+   * `saveValue()` by then, so there is nothing left to redirect.
    *
    * @param {object} indexesChangesState The state object of the index mapper's cache update.
    * @param {boolean} indexesChangesState.indexesSequenceChanged Whether the indexes sequence changed.
@@ -436,7 +519,7 @@ class EditorManager {
     const editor = this.activeEditor;
 
     if ((!indexesChangesState.trimmedIndexesChanged && !indexesChangesState.indexesSequenceChanged) ||
-        !editor ||
+        this.#structuralChangeInProgress || !editor ||
         editor.state !== EDITOR_STATE.EDITING ||
         this.#editedPhysicalRow === null || this.#editedPhysicalColumn === null) {
       return;
--- a/handsontable/src/editorManager.ts
+++ b/handsontable/src/editorManager.ts
@@ -120,13 +120,17 @@ class EditorManager {
    * each suspend and resume the mapper themselves, so even inside `hot.batch()` every `alter()`
    * flushes its own update and the two counts are observed separately.
    *
-   * @param {object} indexesChangesState The state object of the index mapper's cache update.
+   * `afterRowSequenceCacheUpdate` is a PUBLIC hook, so `hot.runHooks()` can fire it with no payload
+   * at all. The state defaults to all-false for that case, which reduces the repair to the structural
+   * one and keeps the hidden-cell guard behind it running.
+   *
+   * @param {object} [indexesChangesState] The state object of the index mapper's cache update.
    * @param {boolean} indexesChangesState.indexesSequenceChanged Whether the indexes sequence changed.
    * @param {boolean} indexesChangesState.trimmedIndexesChanged Whether the trimmed indexes changed.
    */
   #onRowSequenceCacheUpdate = (indexesChangesState: {
     indexesSequenceChanged: boolean; trimmedIndexesChanged: boolean;
-  }): void => {
+  } = { indexesSequenceChanged: false, trimmedIndexesChanged: false }): void => {
     const indexCount = this.hot.rowIndexMapper.getNumberOfIndexes();
 
     this.#repairEditor(indexCount !== this.#lastRowIndexCount, indexesChangesState);
@@ -138,13 +142,13 @@ class EditorManager {
    * Kept separate from the row handler so a structural change on one axis cannot route the other
    * axis's rearrangement into the wrong repair.
    *
-   * @param {object} indexesChangesState The state object of the index mapper's cache update.
+   * @param {object} [indexesChangesState] The state object of the index mapper's cache update.
    * @param {boolean} indexesChangesState.indexesSequenceChanged Whether the indexes sequence changed.
    * @param {boolean} indexesChangesState.trimmedIndexesChanged Whether the trimmed indexes changed.
    */
   #onColumnSequenceCacheUpdate = (indexesChangesState: {
     indexesSequenceChanged: boolean; trimmedIndexesChanged: boolean;
-  }): void => {
+  } = { indexesSequenceChanged: false, trimmedIndexesChanged: false }): void => {
     const indexCount = this.hot.columnIndexMapper.getNumberOfIndexes();
 
     this.#repairEditor(indexCount !== this.#lastColumnIndexCount, indexesChangesState);
@@ -480,7 +484,8 @@ class EditorManager {
   #recaptureEditedRecord(): void {
     const editor = this.activeEditor;
 
-    if (!editor || editor.state !== EDITOR_STATE.EDITING || editor.row === null || editor.col === null) {
+    if (!editor || (editor.state !== EDITOR_STATE.EDITING && editor.state !== EDITOR_STATE.VIRGIN) ||
+        editor.row === null || editor.col === null) {
       return;
     }
 
@@ -577,10 +582,14 @@ class EditorManager {
     indexesSequenceChanged: boolean; trimmedIndexesChanged: boolean;
   }): void {
     const editor = this.activeEditor;
+    const isEditing = editor?.state === EDITOR_STATE.EDITING;
+    // A prepared-but-untyped editor is just as stale as an editing one, and more quietly so: nothing
+    // re-prepares it, `openEditor()` skips `prepareEditor()` while a reference exists, and the first
+    // keystroke then calls `beginEditing()` on the pre-change coordinates.
+    const isPrepared = editor?.state === EDITOR_STATE.VIRGIN;
 
     if ((!indexesChangesState.trimmedIndexesChanged && !indexesChangesState.indexesSequenceChanged) ||
-        !editor ||
-        editor.state !== EDITOR_STATE.EDITING ||
+        !editor || (!isEditing && !isPrepared) ||
         this.#editedPhysicalRow === null || this.#editedPhysicalColumn === null) {
       return;
     }
@@ -588,14 +597,23 @@ class EditorManager {
     const visualRow = this.hot.rowIndexMapper.getVisualFromPhysicalIndex(this.#editedPhysicalRow);
     const visualColumn = this.hot.columnIndexMapper.getVisualFromPhysicalIndex(this.#editedPhysicalColumn);
 
+    if (isPrepared) {
+      // Nothing has been typed, so there is no value to carry and no reason to rebind. Dropping the
+      // reference is the whole repair: the next keystroke finds no active editor and goes back
+      // through `prepareEditor()`, which reads the coordinates, `TD`, `prop`, `originalValue` and
+      // cell meta from the post-change state.
+      if (visualRow !== editor.row || visualColumn !== editor.col) {
+        this.clearActiveEditor();
+      }
+
+      return;
+    }
+
     // No visual index means the record itself is trimmed. There is nowhere to commit to, so the edit
     // is dropped rather than written through coordinates that now address a different record.
     if (visualRow === null || visualColumn === null) {
       editor.cancelChanges();
-      // Drop the reference too. `openEditor()` re-prepares only when there is no active editor, so a
-      // lingering one would let the next keystroke reuse this editor's pre-trim `TD`, `prop`,
-      // `originalValue` and cell meta. Clearing sends the next edit back through `prepareEditor()`,
-      // which reads them from the post-trim state.
+      // Drop the reference too, for the same reason as the prepared case above.
       this.clearActiveEditor();
 
       return;
--- a/handsontable/src/editorManager.ts
+++ b/handsontable/src/editorManager.ts
@@ -574,6 +574,11 @@ class EditorManager {
    * before it – but the value does not survive. Making it survive means moving the selection with the
    * record, which is a larger change than this repair.
    *
+   * An editor a structural change stranded past the last row is discarded here rather than rebound:
+   * its captured record was cleared as unresolvable, and its own coordinates address nothing, so
+   * there is no record left to follow. Discarding is what keeps a following `Filters#filter()` from
+   * committing through those coordinates and appending records.
+   *
    * Two further limits. An index-map change does NOT adjust the selection – `core.ts` calls
    * `selection.commit()` only for `hiddenIndexesChanged` – so the highlight can be left past the last
    * row, and typing into it grows the data set. That is reachable with no editor involved at all and
@@ -600,8 +605,28 @@ class EditorManager {
     const isPrepared = editor?.state === EDITOR_STATE.VIRGIN;
 
     if ((!indexesChangesState.trimmedIndexesChanged && !indexesChangesState.indexesSequenceChanged) ||
-        !editor || (!isEditing && !isPrepared) ||
-        this.#editedPhysicalRow === null || this.#editedPhysicalColumn === null) {
+        !editor || (!isEditing && !isPrepared)) {
+      return;
+    }
+
+    // No captured record means an earlier structural change renumbered it away. The editor's own
+    // coordinates are then the only thing left, so check whether they still address anything: if
+    // they do not, that change stranded the editor past the last row and nothing re-prepared it. A
+    // commit through those coordinates is what `applyChanges()` satisfies by APPENDING records, so
+    // the edit is dropped here rather than at the next keystroke.
+    //
+    // Only reached with no captured record. While one exists it is the better guide - it survives a
+    // trim that leaves the editor's own stale coordinates unresolvable, which is the ordinary rebind.
+    if (this.#editedPhysicalRow === null || this.#editedPhysicalColumn === null) {
+      if (this.hot.rowIndexMapper.getPhysicalFromVisualIndex(editor.row!) === null ||
+          this.hot.columnIndexMapper.getPhysicalFromVisualIndex(editor.col!) === null) {
+        if (isEditing) {
+          editor.cancelChanges();
+        }
+
+        this.clearActiveEditor();
+      }
+
       return;
     }
 
--- a/handsontable/src/editorManager.ts
+++ b/handsontable/src/editorManager.ts
@@ -70,6 +70,21 @@ class EditorManager {
    * @type {boolean}
    */
   #hiddenCellCloseArmed = false;
+  /**
+   * The PHYSICAL row index of the record the active editor was prepared for, or `null` when no
+   * editor is prepared. Captured because the editor itself stores only VISUAL coordinates, which a
+   * trimming index map invalidates without notice.
+   *
+   * @type {number|null}
+   */
+  #editedPhysicalRow: number | null = null;
+  /**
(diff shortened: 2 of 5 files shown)


qunabu:

handsontable/src/editorManager.ts
```diff
@@ -345,6 +429,132 @@ class EditorManager {
...
+
+    if (!editor || editor.state !== EDITOR_STATE.EDITING || editor.row === null || editor.col === null) {
+      return;
+    }
+
+    const physicalRow = this.hot.rowIndexMapper.getPhysicalFromVisualIndex(editor.row);
+    const physicalColumn = this.hot.columnIndexMapper.getPhysicalFromVisualIndex(editor.col);
+
+    if (physicalRow === null || physicalColumn === null) {
+      editor.cancelChanges();
```
Haxikowy: Blocker: this discards a valid edit whenever the removal is **above** the editor, which is the direction the spec never covers.

`dataMap.removeRow` fires `cacheUpdated` synchronously, and `core.ts` calls `selection.shiftRows()` only afterwards. So at handler time `editor.row` is still pre-removal, the count dropped, `isStructuralChange` is true, and `getPhysicalFromVisualIndex(editor.row)` is `null` for the last row.

5 rows, editing visual 4, `hot.alter('remove_row', 0, 1)`:

```
develop:  state=STATE_EDITING row=3 value=EDITED  ->  [["A1"],["A2"],["A3"],["EDITED"]]
this PR:  state=STATE_VIRGIN  row=3 value=""      ->  [["A1"],["A2"],["A3"],["A4"]]
```

The typed value is gone with no error and no visible cue, and `getActiveEditor()` still returns the same editor object (re-prepared by the selection shift), so only `state`/`getValue()` reveals it. On develop the shift ran `prepareEditor()` and the edit committed to the right record.

Measured on a bundle with this diff applied by hand rather than a clean build, so please reproduce before acting.

One fix: make this branch a no-op and let the `prepareEditor()` that `shiftRows` triggers re-derive both coordinates. Discard only when no re-prepare follows.

handsontable/src/editorManager.ts
```diff
@@ -345,6 +429,132 @@ class EditorManager {
...
+   * Inserting or removing rows shifts the physical indexes of everything below, so the index captured
+   * in `prepareEditor()` now addresses a different record - but the editor's VISUAL coordinate came
+   * through intact, because the index mapper moved the visual space with it. So the visual side is
+   * the trustworthy one here, and it is what the record is read back from.
+   *
+   * A visual coordinate that no longer resolves means the change left the editor past the last row.
+   * There is nothing to commit to, so the edit is dropped rather than written through a coordinate
+   * that `applyChanges()` would satisfy by appending records.
+   */
+  #recaptureEditedRecord(): void {
```
Haxikowy, line 514: Blocker: `updateData()` and `updateSettings({ data })` reach the same discard. `core.ts` deliberately keeps an editor open through a data swap (both are excluded from the `closeEditor()` list at `core.ts:802`), but `fitToLength()` shrinks the physical count, so this routes structural and drops the edit with no `beforeChange`/`afterChange`.

That is the path every wrapper takes when the `data` prop changes, and any polling or streaming integration hits it on a timer. Nothing in the diff covers it.

Either close through `closeEditor(true)` on a data swap, or gate the discard on the change source so a replacement of the whole physical space does not silently eat an in-progress edit.

handsontable/src/editorManager.ts
```diff
@@ -345,6 +429,132 @@ class EditorManager {
...
+      // Drop the reference too. `openEditor()` re-prepares only when there is no active editor, so a
+      // lingering one would let the next keystroke reuse this editor's pre-trim `TD`, `prop`,
+      // `originalValue` and cell meta. Clearing sends the next edit back through `prepareEditor()`,
+      // which reads them from the post-trim state.
+      this.clearActiveEditor();
+
+      return;
+    }
+
+    editor.row = visualRow;
```
Haxikowy, line 696: Blocker: for `dropdown`, `date`, `autocomplete` and `handsontable` cell types the rebind does not commit anything.

`DropdownEditor#finishEditing()` (dropdownEditor.ts:43-58) sets `restoreOriginalValue = true` when the active range does not include `(this.row, this.col)`. The rebind moves those coordinates and nothing moves the selection, so `Enter` discards. Filters plus dropdown columns is a very common configuration, so the guarantee in this method's JSDoc fails there.

Second gap on the same lines: `BaseEditor#saveValue()` with `ctrlDown` reads the selection corners, not `this.row`/`this.col` (baseEditor.ts:250-258). A Ctrl+Enter commit after any index-map change therefore still writes through the stale visual coordinates and can still append rows. That is the corruption the changelog claims fixed, not a separate defect.

Shifting the active range alongside the rebind fixes both.

handsontable/src/editorManager.ts
```diff
@@ -345,6 +429,132 @@ class EditorManager {
...
+      return;
+    }
+
+    const visualRow = this.hot.rowIndexMapper.getVisualFromPhysicalIndex(this.#editedPhysicalRow);
+    const visualColumn = this.hot.columnIndexMapper.getVisualFromPhysicalIndex(this.#editedPhysicalColumn);
+
+    // No visual index means the record itself is trimmed. There is nowhere to commit to, so the edit
+    // is dropped rather than written through coordinates that now address a different record.
+    if (visualRow === null || visualColumn === null) {
+      editor.cancelChanges();
```
Haxikowy, line 689: The JSDoc above claims `cancelChanges()` skips the render that `finishEditing(true)` appends. `AutocompleteEditor#discardEditor()` calls `this.hot.view.render()` unconditionally after `super.discardEditor()` (autocompleteEditor.ts:305-309), and `cancelChanges()` always reaches `discardEditor()`. So for the autocomplete family this fires a full synchronous render from inside `IndexMapper#updateCache()`, before the three change flags are cleared and inside `Filters#filter()`'s `batchExecution`.

Either defer the discard the way `#closeEditorWhenCellHidden()` already defers its commit, or drop that sentence and show the nested render is safe.

handsontable/src/editorManager.ts
```diff
@@ -70,6 +70,82 @@ class EditorManager {
...
+   * @param {boolean} indexesChangesState.indexesSequenceChanged Whether the indexes sequence changed.
+   * @param {boolean} indexesChangesState.trimmedIndexesChanged Whether the trimmed indexes changed.
+   * @param {boolean} indexesChangesState.hiddenIndexesChanged Whether the hidden indexes changed.
+   */
+  #onSequenceCacheUpdate = (indexesChangesState: {
+    indexesSequenceChanged: boolean; trimmedIndexesChanged: boolean; hiddenIndexesChanged: boolean;
+  }): void => {
+    const rowIndexCount = this.hot.rowIndexMapper.getNumberOfIndexes();
+    const columnIndexCount = this.hot.columnIndexMapper.getNumberOfIndexes();
+    const isStructuralChange = rowIndexCount !== this.#lastRowIndexCount ||
```
Haxikowy: `IndexMapper` already carries this signal: `indexesChangeSource` is set to `'insert'`/`'remove'`/`'move'`/`'init'`/`'update'`, and `nestedHeaders.ts:202` reads it from a `cacheUpdated` handler for the same question. It reads `undefined` here only because `insertIndexes`/`removeIndexes` clear it before `resumeOperations()` fires the hook, so adding it to the `cacheUpdated` payload at `indexMapper.ts:907` would serve all four consumers and let you delete both counters, the constructor seeding, and the blind spot documented at line 118.

Separate point on these two lines: `isStructuralChange` ORs both axes and both counters are rewritten on every invocation regardless of which hook fired, so one axis's structural change can route the other axis's rearrangement into the wrong repair. Per-axis handlers would remove that coupling.

Also, `this.destroyed` is not checked here. `core.destroy()` tears down the manager first and then force-flushes both mappers, so this runs on a destroyed manager. Every other path in this file guards on it (line 423).

handsontable/src/editorManager.ts
```diff
@@ -70,6 +70,82 @@ class EditorManager {
...
+   * A structural change and a rearrangement need opposite repairs - one invalidates the captured
+   * PHYSICAL index and keeps the visual coordinate, the other does the reverse - and the state object
+   * cannot tell them apart, because `insertIndexes()`/`removeIndexes()` raise the same flags a filter
+   * does. The SIZE of the physical space can: only an insert or a remove changes it. Comparing it
+   * against the previous update picks the right repair without relying on hook ordering, and nothing
+   * a plugin vetoes can strand it - a cancelled `beforeCreateRow` never changes the count either.
+   *
+   * The one shape this misses is a structural change that leaves the count where it started - an
+   * insert and a removal batched into a single cache update. `alter()` performs one operation per
+   * call, so nothing in core produces it.
```
Haxikowy: The conclusion holds but the reason does not. `IndexMapper.isBatched` is a plain boolean, and the `suspendOperations`/`resumeOperations` pair inside `insertIndexes`/`removeIndexes` clears the outer `hot.batch()` flag, so each `alter()` flushes separately even inside a batch. I measured `hot.batch(() => { alter('insert_row_above', 0); alter('remove_row', 3); })`: two cache updates, counts 6 then 5, both classified structural. The hole is unreachable, but not because `alter()` performs one operation per call.

handsontable/src/editorManager.ts
```diff
@@ -365,9 +575,10 @@ class EditorManager {
    * itself on scroll but keeps `state` at `EDITING`, so the edit survives until the user scrolls
    * back. These hooks never fire on scroll, so that case cannot reach here at all.
    *
-   * A TRIMMING map (Filters, `trimRows`) is deliberately out of scope. It collapses the visual
-   * index space instead of preserving it, so the edited coordinates silently rebind to a different
-   * row that is still rendered and `isHidden()` reads `false`. That defect predates this method.
+   * A TRIMMING map (Filters, `trimRows`) needs the opposite treatment and is handled separately by
+   * `#reconcileEditorWithIndexMaps()`, which runs first on the same two hooks. It collapses the
```
Haxikowy: "which runs first on the same two hooks" contradicts `#recaptureEditedRecord()`'s own JSDoc at line 435 ("Runs INSTEAD of the reconciliation, never after it"). On the structural branch the reconciliation does not run at all. Suggest: "handled separately by the repair `#onSequenceCacheUpdate()` dispatches before this method".

handsontable/src/editorManager.ts
```diff
@@ -345,6 +429,132 @@ class EditorManager {
...
+   *
+   * `cancelChanges()` also bypasses editor-level discard policy - `DropdownEditor#finishEditing()`
+   * rewrites the restore flag - which is harmless here because a discard is what that override would
+   * decide anyway once the edited record is gone from the visual space.
+   *
+   * A rebind moves the editor's coordinates, NOT its pixel position or the selection - neither
+   * `render()` nor `view.render()` repositions an open editor, so it stays drawn over the row it
+   * started on for the rest of the edit. On the Filters path that is invisible because `filter()`
+   * closes the editor outright, but on the `trimRows` path the editor is left painted over a
+   * neighbouring row. The commit still lands on the right record; only the position is wrong.
```
Haxikowy: `neighbouring` -> `neighboring` (American English). The added prose also uses an ASCII hyphen in the parenthetical-dash role 23 times where the docs standard asks for an en dash, and `@param indexesChangesState.hiddenIndexesChanged` is documented on both methods though neither reads it.

tests/e2e/editor-trimmed-row.spec.ts
```diff
@@ -0,0 +1,546 @@
...
+});
+
+/**
+ * A SEQUENCE change permutes the visual space without trimming anything: `columnSorting` and
+ * `manualRowMove` report `indexesSequenceChanged` and leave `trimmedIndexesChanged` false. The
+ * editor is stranded exactly as a trim strands it - it holds a visual index that now belongs to a
+ * different record - so the same reconciliation has to cover both, and the row count never betrays
+ * these because no record is added or removed.
+ */
+test.describe('a sequence change under an open editor', () => {
```
Haxikowy, line 378: The column half of the fix has no coverage. `#editedPhysicalColumn`, the `visualColumn === null` discard, the column arm of `#recaptureEditedRecord()` and `columnIndexCount !== #lastColumnIndexCount` are all unexercised, and `sawTrimmingCacheUpdate`/`sawSequenceCacheUpdate` are never called with `'column'` despite taking the axis. The fixture does not enable `manualColumnMove` or `manualColumnFreeze`, which the production JSDoc names as covered producers.

Worth adding: a column move under an open editor asserting `editor.col` rebinds and the value lands in the right column, and an `alter('remove_col')` on an unrelated column asserting the edit survives.

Also missing for the row axis: `updateData`/`loadData` mid-edit, a `dropdown` cell type, and Ctrl+Enter.

.changelogs/13272.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Fixed an in-progress cell edit landing on the wrong record, or appending rows to the source data, when rows were filtered, trimmed, sorted, or moved while the editor was open. The edit now follows its own record, and is discarded when that record can no longer be reached.",
```
Haxikowy: The title names filtering, trimming, sorting and moving, but `alter()` inserting or removing rows and `updateData()` can now also discard or rebind an open edit. Both are user-visible outcomes of the structural split and belong in the sentence.

Haxikowy (commented):
Reviewed across seven angles: correctness, coordinate systems, regression and compatibility, performance (measured), requirements and tests, architecture and quality.

The approach is right. Keying the repair on the mapper's own change signal covers every producer at once, the structural-vs-rearrangement split is a real distinction that a naive flag test would get wrong, and the tests for the shapes it targets assert whole source data plus `countSourceRows()` rather than poking at symptoms. The problems are all in paths it does not target.

Four blockers, all commented inline: a valid edit discarded when a row above the editor is removed (regression against develop, measured both ways); the same discard on `updateData()`/`updateSettings({ data })`, which is the wrapper data-prop path; the rebind failing to commit at all for dropdown, date and autocomplete cells; and Ctrl+Enter bypassing the fix entirely, leaving the row-appending corruption reachable. The first two are one code path, the last two are commit-side gaps the reconciliation cannot see.

**Performance: no regression.** Filter on 100k x 100, headless Chromium, n=24 each, order-alternated: 74.96 ms base vs 73.55 ms with the diff, delta -1.9% against a 95% CI of +/-3.3 pp. True added cost with no editor open is 152 us, 0.2% of that filter. Hook fire counts are 1-2 per user action in every scenario measured, so it never multiplies.

**The red `Performance / run` job on this branch is noise, not a finding.** It reports Filtering +50.3%, Initial Load +24.2%, Cell Editing +14.3% but also Scroll Left +23.7%, Scroll Down +10.0% and Painting 8-33% slower. This PR touches nothing on the scroll or paint path and no cache update fires on scroll. Shared runner, 3 iterations, golden baseline 42 minutes old. It should not block the merge; a same-runner re-baseline would clear it.

One perf win outside the diff, worth a follow-up rather than this PR: `IndexMapper#getNumberOfIndexes()` is `getIndexesSequence().length`, and `IndexesSequence#getValues()` materializes `0..n-1` on every call while the sequence is identity, which is the common unsorted case. 152 us and ~800 KB of garbage per call-pair at 100k rows, 1752 us at 1M. `return this.indexesSequence.getLength()` is O(1) and identity-aware, and also removes an allocation from the existing `cacheFromPhysicalToVisualIndexes` path.

Two claims I chased and dropped, so nobody else spends time on them: `hot.batch(insert + remove)` does not coalesce into a count-neutral update (measured, commented inline), and a rebind does not invalidate `prop`, `originalValue` or `cellProperties` (`prop` is keyed by physical column via `colToProp`, so it survives any permutation; only `TD` goes stale and nothing on the commit path reads it).

Unverified, flagging for you rather than asserting: `nestedRows`' drag-move reorders the source array through its own data manager without touching `indexesSequence`, so the count is unchanged and this routes to the rearrangement repair with a physical index that may no longer address the same record. Two agents raised it independently and I could not confirm it either way.

Two open items from the PR body. The selection-past-last-row defect described as "worth its own ticket" has no ticket filed, and neither known limit is parked as a `test.fixme` with a ticket reference the way `tests/AGENTS.md` prescribes, so both live only inside a private-method JSDoc. Separately, "each case was verified to fail on a bundle rebuilt without the corresponding part of the fix" is overstated for the two structural cases and the vetoed-insert case, which pass on develop for the reasons their own comments give.

Docs: leaving the box unchecked is right. Nothing public changed and the private methods do not reach Typedoc. Optional, not blocking: the collaboration guide already tells integrators to check `getActiveEditor()`/`isOpened()` before applying a change, and one sentence there about the grid now discarding an edit whose record is filtered out would fit.

Changelog is otherwise sound. `type: "fixed"` is right, a second `changed` entry is not warranted since the discard replaces silent corruption rather than documented behavior, and `breaking: false` holds against `.ai/BREAKING-CHANGES.md`. Worth stating in the body though: `getActiveEditor()` now returns a `STATE_VIRGIN` editor at points where it previously returned an editing one, and a discarded edit fires no `beforeChange`/`afterChange`.

What I ran: no test suites. Findings come from reading whole files on the branch, plus jsdom and headless-Chromium probes against the released bundle and a hand-patched one. The blocker at `editorManager.ts:457` is the one to reproduce first.

handsontable/src/editorManager.ts
```diff
@@ -345,6 +429,132 @@ class EditorManager {
...
+   * @param {boolean} indexesChangesState.hiddenIndexesChanged Whether the hidden indexes changed.
+   */
+  #reconcileEditorWithIndexMaps(indexesChangesState: {
+    indexesSequenceChanged: boolean; trimmedIndexesChanged: boolean; hiddenIndexesChanged: boolean;
+  }): void {
+    const editor = this.activeEditor;
+
+    if ((!indexesChangesState.trimmedIndexesChanged && !indexesChangesState.indexesSequenceChanged) ||
+        !editor ||
+        editor.state !== EDITOR_STATE.EDITING ||
```
budnix: An editor that is prepared but not yet typed into is skipped here, and that leaves the same corruption one step earlier. Clicking a cell runs `prepareEditor()`, so `activeEditor` exists in state `VIRGIN` with `#editedPhysicalRow` already captured - but a trim moves neither the selection nor the editor, and `openEditor()` re-prepares only when there is no active editor, so the first keystroke calls `beginEditing()` on the stale visual row. With 5 rows: click visual 3 (`'A3'`), `trimRows([0])`, then type - the value lands on `'A4'`. For the `VIRGIN` case just clearing the reference would be enough, since the next keystroke would then re-prepare from post-trim state.

tests/e2e/editor-trimmed-row.spec.ts
```diff
@@ -0,0 +1,546 @@
...
+   * `alter('remove_row')` shifts PHYSICAL indexes and emits a trimming-map change on the way, so the
+   * captured record index can go stale. Core closes the editor only when the removed range covers
+   * the highlighted row, so editing row 4 while row 1 is removed leaves the editor open with a
+   * coordinate that is now past the end - which before the fix appended a record on the next commit.
+   *
+   * This pins the past-the-end shape only: the captured index no longer resolves, so the edit is
+   * dropped and nothing is invented. The other stale shape - the captured index still resolves, but
+   * to a different surviving record - is not covered here, and the fix does not claim to repair it.
+   */
+  test('never grows the data set when a row removal strands the editor past the last row',
```
budnix: This test and the blocker on `editorManager.ts` line 457 disagree about the same gesture: here the dropped edit is the expected result, there it is the bug. Once the discard is fixed the expectation becomes `['A0', 'A2', 'A3', 'EDITED']`, so leaving it as is will make a correct fix look like a regression. The comment above also says the old code "appended a record on the next commit", but `selection.shiftRows()` marks the source `'shift'`, which `core.ts:802` leaves out of the `closeEditor()` list and not the `prepareEditor()` one - so develop re-prepares and commits onto `'A4'`. Worth a quick run against develop to confirm which behavior you are pinning.

handsontable/src/editorManager.ts
```diff
@@ -70,6 +70,82 @@ class EditorManager {
...
+   * corrected coordinates rather than the stale ones the index map just invalidated.
+   *
+   * @param {object} indexesChangesState The state object of the index mapper's cache update.
+   * @param {boolean} indexesChangesState.indexesSequenceChanged Whether the indexes sequence changed.
+   * @param {boolean} indexesChangesState.trimmedIndexesChanged Whether the trimmed indexes changed.
+   * @param {boolean} indexesChangesState.hiddenIndexesChanged Whether the hidden indexes changed.
+   */
+  #onSequenceCacheUpdate = (indexesChangesState: {
+    indexesSequenceChanged: boolean; trimmedIndexesChanged: boolean; hiddenIndexesChanged: boolean;
+  }): void => {
```
budnix: `afterRowSequenceCacheUpdate` is a public hook, so `hot.runHooks('afterRowSequenceCacheUpdate')` with no payload is a call anyone can make. The old handler ignored its arguments; this one hands them to `#reconcileEditorWithIndexMaps()`, which reads `.trimmedIndexesChanged` and throws on `undefined` - and the throw also skips the `#closeEditorWhenCellHidden()` call below. A default keeps the old behavior.

```suggestion
  #onSequenceCacheUpdate = (indexesChangesState: {
    indexesSequenceChanged: boolean; trimmedIndexesChanged: boolean; hiddenIndexesChanged: boolean;
  } = { indexesSequenceChanged: false, trimmedIndexesChanged: false, hiddenIndexesChanged: false }): void => {
```

handsontable/src/editorManager.ts
demtario: Confirmed and fixed in `d8aef99c0`. I reproduced it both ways before touching anything:

```
develop:  alter('remove_row',0,1) -> STATE_EDITING row=3 -> Enter -> ["A1","A2","A3","EDITED"]
95fc16256: -> STATE_VIRGIN value="" -> Enter -> ["A1","A2","A3","A4"]
```

Your diagnosis was exact, including the ordering: the cache update fires inside `dataMap.removeRow` and `selection.shiftRows()` runs after it, so at handler time the coordinate is pre-removal and resolves to nothing.

I took your suggested fix. The branch is now a bare `return` with a comment explaining that the `prepareEditor()` behind the shift re-derives both the coordinates and the captured record, and that discarding here throws away an edit core was about to carry across correctly.

Worth flagging how bad this was: the spec case at line 303 asserted the *discard* as the expected result, so the regression was pinned as intended behavior. That case is rewritten to expect `['A0','A2','A3','EDITED']` (also your other comment).

handsontable/src/editorManager.ts
demtario: I could not reproduce the discard here. Measured on both sides, editing visual row 1 and then `updateData()` down to 3 rows:

```
develop:   STATE_EDITING row=1 value=EDITED -> Enter -> ["N0","EDITED","N2"]
this PR:   STATE_EDITING row=1 value=EDITED -> Enter -> ["N0","EDITED","N2"]
```

Identical. The count does move, so it routes structural, but `#recaptureEditedRecord()` only discards when the editor's visual coordinate no longer resolves — and row 1 still resolves in a 3-row set. It would discard only if the editor sat past the new length, and that is the same shape as the `remove_row` blocker, so the fix there covers this too.

Added a case pinning the swap behavior (`a data swap under an open editor`) so it cannot drift. If you have a configuration where you saw the edit dropped, I would like the row index you were editing — I may be missing the shape.

handsontable/src/editorManager.ts
demtario: Both confirmed, and both are now pinned by cases — but neither is a regression, which changes what I did about them.

Measured after `trimRows([0,1])` with the edit on `'A4'`:

| | develop | this PR |
|---|---|---|
| dropdown + Enter | appends, 5 -> 7 records, `EDITED` on the last | edit lost, data intact |
| Ctrl+Enter | edit lost, no write | edit lost, no write |

So the dropdown family goes from the row-appending corruption to a lost edit, and Ctrl+Enter is unchanged. What was actually wrong was my JSDoc claiming a guarantee those paths do not deliver. It now names both exceptions explicitly (`DropdownEditor` and the `date`/`autocomplete`/`handsontable` types built on it; `saveValue()` reading selection corners under `ctrlDown`), says the outcome is lost-not-misplaced, and says that making the value survive means moving the selection with the record.

Two cases under `commit paths the rebind cannot reach` assert the lossy outcome, so a future change that turns either back into a write fails there.

I did not shift the active range. It is the right end state, but it is a behavior change with its own blast radius (every plugin observing selection hooks) and I would rather it be its own reviewed change than a rider here.

handsontable/src/editorManager.ts
demtario: Correct, and the sentence is gone. `AutocompleteEditor#discardEditor()` calls `this.hot.view.render()` unconditionally after `super.discardEditor()`, so `cancelChanges()` does not avoid every render — only the one `finishEditing()` appends.

The JSDoc now says exactly that, and notes the nested repaint reads caches `updateCache()` has already rebuilt. I did not defer the discard: deferring is what the Filters path cannot tolerate (`filter()` re-selects synchronously right after writing its map, which commits the edit before any deferred handler could run), so trading one for the other would reopen the original bug.

handsontable/src/editorManager.ts
demtario: Two of the three are done in `d8aef99c0`.

**Per-axis:** split into `#onRowSequenceCacheUpdate` and `#onColumnSequenceCacheUpdate` over a shared `#repairEditor(isStructuralChange, state)`, each comparing only its own axis's count. The cross-axis coupling you describe is gone.

**`this.destroyed`:** added at the top of `#repairEditor`, with your reason in the comment.

**`indexesChangeSource`:** not done, and I want to be straight that this is a scope call rather than a disagreement. You are right that it is the better signal and that it would let me delete both counters and the constructor seeding. It means changing the `cacheUpdated` payload in `IndexMapper`, which is shared infrastructure with four consumers, and I would rather not fold that into a PR that has already changed shape three times. Happy to do it here if you would prefer it over a follow-up.

handsontable/src/editorManager.ts
demtario: You are right and the reasoning is corrected. I had asserted `alter()` performs one operation per call; the actual reason the hole is unreachable is that `insertIndexes()`/`removeIndexes()` each suspend and resume the mapper themselves, so every `alter()` flushes its own update even inside `hot.batch()` and the two counts are observed separately.

The JSDoc now says that. Thanks for measuring it rather than letting the wrong reason stand — the conclusion happened to hold, which is exactly how a wrong rationale survives.

handsontable/src/editorManager.ts
demtario: Fixed. The structural branch runs *instead of* the reconciliation, so "runs first on the same two hooks" was wrong. Reworded to "the repair dispatched before this method on the same two hooks", which matches `#recaptureEditedRecord()`'s own wording.

handsontable/src/editorManager.ts
demtario: All three addressed.

- `neighbouring` -> `neighboring`.
- `@param indexesChangesState.hiddenIndexesChanged` removed from both methods, and dropped from the type signatures too since neither reads it.
- The en dashes are done in `3d6b059b1`: 19 clause-separator hyphens in the JSDoc I added are now en dashes, per `.ai/DOC-STANDARDS.md` rule 11. I scoped the change to lines this PR introduced rather than reflowing the pre-existing `#closeEditorWhenCellHidden()` prose, to keep the diff readable — say the word if you would rather have the whole file consistent in one go.

tests/e2e/editor-trimmed-row.spec.ts
demtario: Added, all of it.

**Column axis:** `follows the record through a column move` (asserts `editor.col` rebinds to 1 and the value lands on physical column 0) and `keeps a valid edit when another column is removed`. The fixture now enables `manualColumnMove`, and `sawSequenceCacheUpdate('column')` is asserted, so the axis parameter is exercised. `#editedPhysicalColumn`, the column arm of the recapture and the column index count are all live now.

**Row axis:** a `dropdown` cell type (fixture gained an `?editor=` param), Ctrl+Enter, and `updateData()` mid-edit — all three as cases.

Both column cases fail against develop, so they discriminate. 20 cases now, 13 of which fail against develop; the other 7 are regression guards for behavior develop already gets right, and the PR body now says that instead of the blanket claim you called out.

.changelogs/13272.json
demtario: Updated. The title now reads:

> Fixed an in-progress cell edit landing on the wrong record, or appending rows to the source data, when rows or columns were filtered, trimmed, sorted, moved, inserted, removed, or replaced while the editor was open. The edit now follows its own record, and is discarded when that record can no longer be reached.

I also took your point from the review body into the PR description: `getActiveEditor()` now returns a `STATE_VIRGIN` editor where it previously returned an editing one, and a discarded edit fires no `beforeChange`/`afterChange`.

handsontable/src/editorManager.ts
demtario: Confirmed and fixed in `883637478`. Reproduced exactly as you described — clicking visual 3 (`'A3'`), `trimRows([0])`, then typing.

`VIRGIN` is now handled in `#reconcileEditorWithIndexMaps()`: nothing has been typed, so there is no value to carry and no reason to rebind — the reference is dropped when the record moved, and the next keystroke goes back through `prepareEditor()`. `#recaptureEditedRecord()` accepts `VIRGIN` too, so the captured index cannot go stale underneath that comparison.

One correction to the expected outcome, and I would value your read on it. The value still lands on `'A4'`, not `'A3'`, on both sides of the fix — the selection never moved, so visual row 3 is what the user sees highlighted and that is where the write goes. What the fix repairs is the editor disagreeing with that selection about `TD`, `prop`, `originalValue` and cell meta: `originalValue` read `'A3'` before and reads `'A4'` now, and that stale meta is where a wrong `readOnly` or `validator` would have come from. Making the write follow the clicked record instead needs the selection to move with it, which is the same larger change flagged elsewhere in this PR.

The test asserts the meta, and its comment says explicitly that the target record is not what changes — I did not want it to look like it proves more than it does.

tests/e2e/editor-trimmed-row.spec.ts
demtario: Already fixed in `d8aef99c0`, and your diagnosis of develop's behavior was right — I verified it before rewriting:

```
develop: alter('remove_row',0,1) while editing visual 4 -> STATE_EDITING row=3 -> Enter -> ["A1","A2","A3","EDITED"]
```

`shiftRows()` marks the source `'shift'`, which `core.ts:802` excludes from `closeEditor()` but not from `prepareEditor()`, exactly as you said. So develop re-prepares and commits onto `'A4'`, and both my comment ("appended a record on the next commit") and the expectation were wrong.

The case now asserts `['A0','A2','A3','EDITED']` and additionally pins `STATE_EDITING`, `editorRow() === 3` and the retained text, so it fails if anything discards or misplaces the edit instead.

handsontable/src/editorManager.ts
demtario: Confirmed and fixed in `883637478`. `hot.runHooks('afterRowSequenceCacheUpdate')` threw `TypeError: Cannot read properties of undefined (reading 'trimmedIndexesChanged')`, and as you noted it also skipped the `#closeEditorWhenCellHidden()` call behind it.

I applied your default-parameter fix, with one adjustment: `#onSequenceCacheUpdate` no longer exists — it was split into `#onRowSequenceCacheUpdate` and `#onColumnSequenceCacheUpdate` in the previous commit — so the default is on both, and without `hiddenIndexesChanged`, which neither handler reads:

```ts
} = { indexesSequenceChanged: false, trimmedIndexesChanged: false }): void => {
```

All-false reduces the work to the structural repair, which is count-driven and still correct, and the hidden-cell guard keeps running. There is a case for it (`survives the public hook being fired without a payload`), verified red against the previous commit.

handsontable/src/editorManager.ts
```diff
@@ -70,6 +70,64 @@ class EditorManager {
...
+   * @type {boolean}
+   */
+  #structuralChangeInProgress = false;
+  /**
+   * Marks the start of a structural change, so the cache update it emits does not reconcile against
+   * a physical index that the change is in the middle of renumbering.
+   */
+  #onBeforeStructuralChange = (): void => {
+    this.#structuralChangeInProgress = true;
+  };
```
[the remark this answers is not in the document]
demtario: Addressed in `95fc16256`, by removing the mechanism rather than patching it.

I verified both halves of your premise first: `Formulas#onBeforeCreateRow`/`#onBeforeRemoveRow` do return `false` when HyperFormula rejects (`formulas.ts:1978`, `:2025`), and a probe confirmed a cancelled insert fires `beforeCreateRow` with no `afterCreateRow` and no cache update behind it — so the latch would indeed stick.

Rather than add a defensive clear, the hook-pair latch is gone. A structural change and a rearrangement need opposite repairs and the state object cannot tell them apart, but the SIZE of the physical index space can: only an insert or a remove changes it. `#onRowSequenceCacheUpdate`/`#onColumnSequenceCacheUpdate` compare `getNumberOfIndexes()` against the previous update and dispatch on that. A cancelled insert changes no counts, so there is nothing to arm and nothing to strand — and the eight `before*`/`after*` hook registrations came out with it.

There is a case for the scenario (`a structural change that gets vetoed`). Being straight about what it proves: it passes against the latch implementation too, because an `alter()` is followed by a re-prepare that happened to clear the latch. It is an invariant guard against a future refactor back to a hook-armed design, not a reproduction, and its comment says so.

budnix (dismissed):

budnix (dismissed):

handsontable/src/editorManager.ts
Haxikowy: You are right and I was wrong. I never ran this one — it was a static trace from `fitToLength()` shrinking the count into the structural branch, and I stopped there instead of checking whether the coordinate still resolved. It does, for any row inside the new length, so there is no discard to reproduce. No configuration to give you.

The case you added covers it. Resolving.

handsontable/src/editorManager.ts
Haxikowy: Agreed on all of it, including not shifting the selection here. Your dropdown numbers make the case: appending records on develop versus a lost edit now is a strict improvement, and Ctrl+Enter is untouched either way, so neither belongs to this PR's blast radius.

One ask before this closes: put a ticket number in the JSDoc next to both exceptions. Right now the only record that a dropdown edit is silently dropped after a filter lives in a private method's docstring, and `tests/AGENTS.md` wants a known limit parked as a `test.fixme` naming its ticket. The two cases under `commit paths the rebind cannot reach` assert the current outcome, which is the right thing for a regression guard, but they read as "this is fine" to anyone who finds them later.

handsontable/src/editorManager.ts
Haxikowy: Follow-up is the right call — don't fold it in. Changing the `cacheUpdated` payload touches four consumers and would put shared `IndexMapper` infrastructure in a PR that has already been reshaped three times; the per-axis split you did removes the coupling that actually mattered.

Worth filing the follow-up now while the reasoning is fresh, and linking it from the JSDoc where the count comparison is explained, so the next person to read it knows the heuristic is provisional rather than considered.

handsontable/src/editorManager.ts
demtario: Done in `d85906ab9` — a follow-up ticket is filed in Sprint 14 and named in the code rather than here. (Deliberately not quoting the ID in this comment: the ClickUp integration picks IDs up out of PR comments and would move the task's status.)

Two places carry it:

- `#reconcileEditorWithIndexMaps()`'s JSDoc, on the sentence about moving the selection with the record — so the ticket sits next to both exceptions where the guarantee is bounded.
- The `commit paths the rebind cannot reach` describe block, which now says outright that losing the edit is **not** the desired end state, just where this repair stops, and that the ticket inverts both cases into asserting the value lands on the edited record.

That second one was your actual point — the cases assert the loss on purpose as a regression guard, but nothing told a later reader that. It does now.

The ticket itself carries the develop-vs-branch measurements, the `dropdownEditor.ts:43-58` veto and the `ctrlDown` branch in `saveValue()` as the two mechanisms, and a note that any selection-shifting has to be checked against `Filters#filter()`, which re-selects on its own.

handsontable/src/editorManager.ts
demtario: Filed in Sprint 14 and linked from the JSDoc in `d85906ab9`. Not quoting the ID here — the ClickUp integration reads IDs out of PR comments and would transition the task.

The count-comparison paragraph in `#onRowSequenceCacheUpdate()` now states the heuristic is **provisional rather than the considered design**, names `indexesChangeSource` as the signal that already exists, explains why it reads `undefined` at this call site (`insertIndexes()`/`removeIndexes()` clear it before the hook fires), and points at the ticket — so the next reader gets your reasoning rather than having to re-derive it.

The ticket records that putting it on the `cacheUpdated` payload removes both counters, their constructor seeding and the documented blind spot, and that `nestedHeaders` is already a consumer. It also lists the four describe blocks that have to stay green through that change, including the vetoed-insert one — that case exists because an earlier hook-pair design here could be stranded by a plugin cancelling `beforeCreateRow`, which is worth knowing before anyone swaps the mechanism again.

Haxikowy (approved):

PR #13273, opened 2026-08-27
DEV-2653: Drop a late or superseded autocomplete source response

demtario:
### Context

The PR fixes the autocomplete and dropdown editors acting on a `source` response that arrives too late to be relevant.

`AutocompleteEditor` resolves its choices asynchronously and nothing on the resolution path checked whether the response was still wanted. Two facts made a late response harmful rather than inert:

- `HandsontableEditor.close()` only sets `htEditor.rootElement.style.display = 'none'`. The nested grid instance stays alive, so a late response re-renders instead of throwing.
- `hot._registerTimeout()` has no cancel path. `_clearTimeouts()` is called from exactly one place, `Core#destroy()`, and no editor `close()` clears the queue. So the `queryChoices()` calls deferred in `open()` and in `onBeforeKeyDown()` survive a close too.

The result: `updateChoicesList()` ran against a closed editor, re-showed the suggestion list over a cell the user was no longer editing, and called `hot.listen()`, handing keyboard control back to the grid.

Three paths reached this, and all three funnel through `queryChoices()`:

| Site | Why it lands late |
|---|---|
| `open()` tail | `_registerTimeout(..., 0)`, handle discarded |
| `onBeforeKeyDown()` | `_registerTimeout(..., 10-20)`, handle discarded |
| async `source` callback | user-supplied, arbitrarily late |

The `if (this.htEditor)` check at the third site was never a guard: the field is never nulled anywhere, so after the first open it can never be false, and it is a schedule-time check rather than a fire-time one.

#### The guard is keyed on `close()`, not on an editor flag

Neither state flag describes "the editor is closed", and this is the part worth reviewing closely.

- **`state` stays `EDITING` through a close.** `TextEditor.refreshDimensions()` calls `this.close()` when the edited cell leaves the rendered range, and its own early return proves it only gets there while `state === EDITING`. It runs from the permanently registered `afterScrollVertically` / `afterScrollHorizontally` hooks, so an ordinary scroll reaches it. `HandsontableEditor`'s `afterSetTheme` hook closes the same way.
- **`_opened` stays `false` after the cell scrolls back.** On the way back, `refreshDimensions()` calls `showEditableElement()` but never restores `_opened`. The editor is visibly open and editing with `isOpened() === false`, which is why `onBeforeKeyDown` already carries a pre-existing `if (!this.isOpened())` branch. A guard on that flag would leave the suggestion list dead for the rest of such an edit.

Measured, not argued:

```
after ed.close():                          state=STATE_EDITING  opened=false  display=none
after resolving the in-flight query:       display=""           choices rendered
after ed.close() then ed.refreshDimensions():
                                           state=STATE_EDITING  opened=false  parentOpacity=1
```

So the fix cancels rather than flags, and only falls back to a token where cancelling is impossible:

- **`#queryTimeouts`.** Both deferred `queryChoices()` calls keep their timer id, and `close()` clears them. `hot._registerTimeout()` has no cancel path of its own, so without this a query scheduled during an edit still fires afterwards and starts a fresh request against a cell nobody is editing. Cancelling says "this query is off" in the code rather than leaving every timer to wake up and compare two numbers.
- **`#editSession`, bumped in `close()`.** For the one thing that cannot be cancelled: the `source` callback, which user code holds. Closing is the reliable signal, `close()` is virtual, and every close path resolves to it.
- **`#queryGeneration`, bumped in `queryChoices()`.** Rejects a response superseded by a newer query inside the same edit, where `#editSession` never moves.

`queryChoices()` also returns early unless the editor is `EDITING` or `WAITING`, so a call that reaches it after a normal finish does not invoke the user's `source`, which is typically a network request. `WAITING` counts as open on purpose: `close()` is what unhooks `beforeKeyDown`, so keystrokes still schedule queries while an async validator runs, and under `allowInvalid: false` the editor stays open and returns to `EDITING`. Excluding it would stop the list refreshing for the length of every validation, which `develop` did not do.

#### A second staleness case the counters also cover

Editors are per-instance singletons (`_getEditorInstance`, "Returns instance (singleton) of editor class"), so one `AutocompleteEditor` object serves every autocomplete cell in a grid. Open A1 with a slow `source`, close it, open B2: without a token the A1 response lands while `state` is `EDITING` again and renders A1's choices into B2's dropdown. Keying on `close()` rather than only on the next query also closes the window in between, since reopening only *schedules* its own query on a timeout.

#### Deliberate non-choices

- **No state check inside the source callback.** The generation carries the whole question. Keeping one there dropped a response that lands while an async validator holds the editor in `WAITING`, which under `allowInvalid: false` left the list empty for the rest of the edit once the state returned to `EDITING`.
- **`updateChoicesList()` left unguarded.** It is documented public API and an existing spec calls it directly; all three reachable paths already pass through `queryChoices()`.
- **`HandsontableEditor.close()` still hides rather than destroys `htEditor`.** That is what keeps the object reachable, but changing it is a lifecycle and performance change with breaking surface (nested instance identity, re-init cost on every open) well outside this defect.

#### One more route to the same symptom, also fixed

`#focusDebounced` is armed by the inner grid's `afterScroll` and runs 100 ms later. It is created by `helpers/function.ts`'s `debounce()`, so it never enters `hot.timeouts` and `_clearTimeouts()` cannot reach it even on destroy, and its `cancel()` was never called. Since `hideEditableElement()` only sets `opacity: 0`, its `focus()` put the caret back into a closed editor. `close()` now cancels it, which is what makes the changelog's "taking keyboard control back from the page" true for every route rather than most of them.

#### Public surface

`queryChoices()` is in the shipped declarations. It is now a no-op while no edit is in progress, and a `source` with side effects is no longer called at all on a timeout that fires after the edit finished. `AutocompleteEditor` carries `@private` and the method is absent from the API reference, so `breaking` stays `false`, but the changelog entry states both.

#### Still open, no ticket

The `DropdownEditor.finishEditing()` discard race: with a page turn hiding the edited cell, a dropdown edit can discard where an equivalent select edit commits. Evidenced down to a CI run number in the comment on that case in `tests/e2e/editor-hidden-cell.spec.ts`. Out of scope here and worth its own task.

### How has this been tested?

New Playwright coverage in `tests/e2e/autocomplete-async-source.spec.ts`: ten cases, each run for both `autocomplete` and `dropdown`, so twenty in total.

1. A response arriving after Escape closed the editor.
2. A response arriving after an outside click closed the editor (a different close path: `DropdownEditor.finishEditing()` rewrites `restoreOriginalValue` from the active range, which Escape never reaches).
3. A response from a previous edit arriving after the editor reopened at another column.
4. A response arriving after a scroll closed the editor, asserting `isOpened() === false` **and** `state === 'STATE_EDITING'` first, so it cannot pass by having left the row rendered.
5. A `queryChoices()` timeout scheduled before a close and firing after it, asserting no new query starts.
6. The list still refreshing while an async validator holds the editor in `WAITING`.
7. A response from that validation window being ignored once the editor finally closes.
8. A response arriving after `Core#destroy()`, asserting no page error.
9. The editor not refocusing itself after a close that followed a list scroll.
10. Two overlapping queries in one edit resolved newest-first, asserting the older answer does not overwrite the newer.

Every case that can observe both symptoms asserts both. `updateChoicesList()` sets `display` and calls `hot.listen()` independently, so a case checking only the dropdown would pass on a fix that still stole focus. The focus half counts `afterListen`, and the scroll cases hand keyboard control away first, because `Core#listen()` fires that hook only on a not-listening to listening transition and would otherwise pin the count.

Queries are attributed to the editor rather than counted raw: `autocompleteValidator` calls the same user `source` on the strict save path, so a bare "something was in flight" guard could be satisfied by the validator alone.

The fixture's `source` answers only when the spec tells it to, so "the response arrives late" is stated rather than raced. Where a case has to prove a timer already fired, it waits on a sentinel timer armed later with a longer delay, which is an ordering fact about two timers rather than a guess about worker load.

**Which mechanism catches which case.** Each row is one build with a single mechanism removed, run against the same twenty tests, reporting failures:

| Build | Failures |
|---|---|
| Full fix | 0 / 20 |
| Without `#queryTimeouts` cancellation in `close()` | 2 / 20 (case 5) |
| Without `#editSession` | 10 / 20 (cases 1, 2, 4, 5, 7) |
| Without `#queryGeneration` | 2 / 20 (case 10) |
| Without `WAITING` in the entry guard | 2 / 20 (case 6) |
| Without the `isDestroyed` check | 2 / 20 (case 8) |
| Without the `#focusDebounced.cancel()` in `close()` | 2 / 20 (case 9) |

Every mechanism is load-bearing, and every one of them except the session token is pinned by exactly one case. `#editSession` is the broad one: it is the only thing standing between a `source` callback and a closed editor, so it covers the four close paths plus case 5, which the cancellation would otherwise have to catch alone.

```bash
# New spec, all six theme x bundle legs
cd tests && npx playwright test e2e/autocomplete-async-source.spec.ts

# Regression surface for a too-broad guard: the legacy suites that exercise these editors
npm --prefix handsontable run test:e2e -- --testPathPattern='editors/'
```

Full results, all green:

- Legacy Jasmine, whole suite: 9745 specs, 0 failures, 8 pending.
- Core unit: 3963 tests, 325 suites.
- Playwright, all six legs, whole suite: 2908 passed, 6 skipped.
- `npm --prefix handsontable run test:types`, `run eslint` (0 errors), and `npm --prefix tests run lint`.

**`tests/e2e/editor-hidden-cell.spec.ts` is unchanged except for its comment.** Its dropdown case asserts only that the editor closed, with a comment blaming this timeout for the committed value being unstable. Two things about that turned out to be worth recording. The blame was misplaced: the fixture's `source` is a plain array, so `queryChoices()` runs synchronously and nothing is deferred by the time the page turns. But the instability is real — tightening the case to assert the committed value held 15/15 locally on `e2e-main` and still failed on `e2e-classic` in CI (run 33069787284), committing `A1` instead of `A4`. That is `DropdownEditor.finishEditing()` rewriting `restoreOriginalValue` from the active range, and pinning the value needs that path made deterministic first. The comment now says so; the assertion stays loose.

### Types of changes

### Related issue(s):

1. DEV-2653

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

The autocomplete guide's asynchronous-source section stated the `process()` contract without saying anything about timing. It now says that a response arriving after the editor closed, or superseded by a newer query, is ignored. `tests/AGENTS.md` gained the fixture trap this work uncovered: on an empty cell, `click()` followed by `Enter` opens the editor and closes it again on the same keypress, silently, which makes an editor fixture fail at "the editor never opened" with nothing pointing at the cause.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2653

---

> [!NOTE]
> **Medium Risk**
> Touches core cell-editing and keyboard-focus behavior for autocomplete/dropdown with async sources, though the change is narrowly scoped to stale-response handling and is heavily regression-tested.
> 
> **Overview**
> Fixes autocomplete and dropdown editors **re-opening the suggestion list** and calling **`hot.listen()`** when an asynchronous `source` callback (or a deferred `queryChoices()` timer) finishes after the edit has ended.
> 
> **`AutocompleteEditor`** now tracks edit sessions and query generations, routes deferred lookups through **`#deferQuery()`** so **`close()`** can cancel pending timers, bumps **`#editSession`** on close, and drops stale **`source`** results when the grid is destroyed, the session changed, or a newer query superseded the request. **`queryChoices()`** is a no-op unless the editor is **`EDITING`** or **`WAITING`** (so async validation can still refresh the list). **`close()`** also cancels the debounced refocus that could steal focus after the list scrolled.
> 
> Docs and **`metaSchema`** describe that late or superseded async responses are ignored; callers can still invoke **`process()`** when the request completes. Adds changelog **13273**, Playwright coverage with a controllable async fixture, and a small **`tests/AGENTS.md`** note on seeding editor cells for e2e.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e16f486f98d32046c45be725e5014cebab437071. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/content/guides/cell-types/autocomplete-cell-type/autocomplete-cell-type.md
+++ b/docs/content/guides/cell-types/autocomplete-cell-type/autocomplete-cell-type.md
@@ -150,7 +150,7 @@ In strict mode, the [`allowInvalid`](@/api/options.md#allowinvalid) option deter
 
 Autocomplete can also use asynchronous data sources. In the example below, suggestions for the "Car" column are loaded from the server with the Fetch API. To load data from a remote source, assign a function to the `source` option. The function receives the query string and the `process` callback. Call `process()` with the result array when the request completes.
 
-Handsontable ignores a response that arrives after the editor closed - including a close you may not notice, such as scrolling the edited cell out of view - and it ignores a response that a newer query has superseded, which happens on every keystroke. Call `process()` whenever the request completes, even late.
+Handsontable ignores a response that arrives after the editor closed - including a close you may not notice, such as scrolling the edited cell out of view - and it ignores a response that a newer query has superseded, which happens as you type. Call `process()` whenever the request completes, even late.
 
 ::: only-for javascript
 
--- a/docs/content/guides/cell-types/autocomplete-cell-type/autocomplete-cell-type.md
+++ b/docs/content/guides/cell-types/autocomplete-cell-type/autocomplete-cell-type.md
@@ -150,7 +150,7 @@ In strict mode, the [`allowInvalid`](@/api/options.md#allowinvalid) option deter
 
 Autocomplete can also use asynchronous data sources. In the example below, suggestions for the "Car" column are loaded from the server with the Fetch API. To load data from a remote source, assign a function to the `source` option. The function receives the query string and the `process` callback. Call `process()` with the result array when the request completes.
 
-Call `process()` whenever the request completes, even late. Handsontable ignores a response that arrives after the editor closed, and it ignores a response that a newer query has already superseded. A slow request can never reopen the suggestion list over a cell you have stopped editing.
+Call `process()` whenever the request completes, even late. Handsontable ignores a response that arrives after the editor closed, and it ignores a response that a newer query has already superseded.
 
 ::: only-for javascript
 
--- a/docs/content/guides/cell-types/autocomplete-cell-type/autocomplete-cell-type.md
+++ b/docs/content/guides/cell-types/autocomplete-cell-type/autocomplete-cell-type.md
@@ -150,6 +150,8 @@ In strict mode, the [`allowInvalid`](@/api/options.md#allowinvalid) option deter
 
 Autocomplete can also use asynchronous data sources. In the example below, suggestions for the "Car" column are loaded from the server with the Fetch API. To load data from a remote source, assign a function to the `source` option. The function receives the query string and the `process` callback. Call `process()` with the result array when the request completes.
 
+Call `process()` whenever the request completes, even late. Handsontable ignores a response that arrives after the editor closed, and it ignores a response that a newer query has already superseded. A slow request can never reopen the suggestion list over a cell you have stopped editing.
+
 ::: only-for javascript
 
 ::: example #example3 .docs-height-small --js 1 --ts 2
--- a/docs/content/guides/cell-types/autocomplete-cell-type/autocomplete-cell-type.md
+++ b/docs/content/guides/cell-types/autocomplete-cell-type/autocomplete-cell-type.md
@@ -150,7 +150,7 @@ In strict mode, the [`allowInvalid`](@/api/options.md#allowinvalid) option deter
 
 Autocomplete can also use asynchronous data sources. In the example below, suggestions for the "Car" column are loaded from the server with the Fetch API. To load data from a remote source, assign a function to the `source` option. The function receives the query string and the `process` callback. Call `process()` with the result array when the request completes.
 
-Handsontable ignores a response that arrives after the editor closed, and it ignores a response that a newer query has already superseded. Call `process()` whenever the request completes, even late.
+Handsontable ignores a response that arrives after the editor closed - including a close you may not notice, such as scrolling the edited cell out of view - and it ignores a response that a newer query has superseded, which happens on every keystroke. Call `process()` whenever the request completes, even late.
 
 ::: only-for javascript
 
--- a/docs/content/guides/cell-types/autocomplete-cell-type/autocomplete-cell-type.md
+++ b/docs/content/guides/cell-types/autocomplete-cell-type/autocomplete-cell-type.md
@@ -150,7 +150,7 @@ In strict mode, the [`allowInvalid`](@/api/options.md#allowinvalid) option deter
 
 Autocomplete can also use asynchronous data sources. In the example below, suggestions for the "Car" column are loaded from the server with the Fetch API. To load data from a remote source, assign a function to the `source` option. The function receives the query string and the `process` callback. Call `process()` with the result array when the request completes.
 
-Call `process()` whenever the request completes, even late. Handsontable ignores a response that arrives after the editor closed, and it ignores a response that a newer query has already superseded.
+Handsontable ignores a response that arrives after the editor closed, and it ignores a response that a newer query has already superseded. Call `process()` whenever the request completes, even late.
 
 ::: only-for javascript
 
--- a/handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
+++ b/handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
@@ -340,6 +340,13 @@ export class AutocompleteEditor extends HandsontableEditor {
     //
     // Queries this editor deferred are cancelled outright; the token below is for the ones already
     // handed to user code, which cannot be.
+    //
+    // Known limitation: `refreshDimensions()` also calls `close()` as "hide for now" when the
+    // edited cell scrolls out of the rendered range, and there is no signal here to tell that apart
+    // from "the edit ended". So after the cell scrolls back the editor is visible again but its
+    // list can no longer populate. That path was already one-way before this change - the
+    // `removeHooksByKey` below means typing could not re-query after a scroll round trip either -
+    // and separating the two meanings belongs in `TextEditor`, not here.
     this.#queryTimeouts.forEach(timeoutId => clearTimeout(timeoutId));
     this.#queryTimeouts.clear();
 
@@ -376,20 +383,19 @@ export class AutocompleteEditor extends HandsontableEditor {
    * @param {string} query The query.
    */
   queryChoices(query: string): void {
-    // Both callers defer through `hot._registerTimeout()`, which has no cancel path - `close()`
-    // never clears the queue, so a timeout scheduled while the editor was open still fires after
-    // it closed. Bail rather than call the user's `source`, which is typically a network request.
+    // `close()` cancels the queries this editor deferred, so no internal caller reaches here once
+    // an edit has ended. This guard is for the public method: `queryChoices()` ships in the type
+    // declarations, and calling it while no edit is in progress must not invoke the user's
+    // `source`, which is typically a network request.
     //
-    // `state` rather than `isOpened()`: `_opened` is false for the whole of `open()`, and it stays
-    // false after `refreshDimensions()` closes an editor whose cell scrolled out of view and then
-    // shows it again on the way back, without ever restoring the flag. A guard on it would leave
-    // the suggestion list dead for the rest of such an edit. The close paths `state` cannot see are
-    // handled by the edit-session token instead, carried by each caller across its timeout.
-    //
-    // `WAITING` counts as open. `close()` is what unhooks `beforeKeyDown`, so keystrokes still
-    // schedule queries while an async validator runs, and under `allowInvalid: false` the editor
-    // stays open and returns to `EDITING`. Rejecting them here would stop the list refreshing for
+    // `WAITING` counts as in progress. `close()` is what unhooks `beforeKeyDown`, so keystrokes
+    // still schedule queries while an async validator runs, and under `allowInvalid: false` the
+    // editor stays open and returns to `EDITING`. Rejecting them would stop the list refreshing for
     // the length of every validation, which is a behavior change rather than a fix.
+    //
+    // `state` rather than `isOpened()`: `_opened` stays false after `refreshDimensions()` closes an
+    // editor whose cell scrolled out of view and then shows it again on the way back, without ever
+    // restoring the flag.
     if (this.state !== EDITOR_STATE.EDITING && this.state !== EDITOR_STATE.WAITING) {
       return;
     }
--- a/handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
+++ b/handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
@@ -87,12 +87,21 @@ export class AutocompleteEditor extends HandsontableEditor {
    */
   #queryGeneration = 0;
   /**
-   * Edit-session token. Bumped on every `close()`, so work scheduled during one edit can tell that
-   * the edit has ended - which neither `state` nor `_opened` reports reliably.
+   * Edit-session token. Bumped on every `close()`, so a `source` response can tell that the edit it
+   * belongs to has ended - which neither `state` nor `_opened` reports reliably. Only the response
+   * needs a token: user code holds that callback and there is nothing to cancel, while the editor's
+   * own deferred queries are cancelled outright through `#queryTimeouts`.
    *
    * @type {number}
    */
   #editSession = 0;
+  /**
+   * Timer ids of the `queryChoices()` calls this editor has deferred and not yet run. Cleared on
+   * `close()`, so a query scheduled during an edit never runs after it.
+   *
+   * @type {Set}
+   */
+  #queryTimeouts: Set<ReturnType<typeof setTimeout>> = new Set();
 
   /**
    * Gets current value from editable element.
@@ -293,29 +302,47 @@ export class AutocompleteEditor extends HandsontableEditor {
       setAttribute(this.TEXTAREA, ...A11Y_EXPANDED('true'));
     }
 
-    // The editor can be closed again before this fires - `refreshDimensions()` does exactly that
-    // from inside `open()` when the cell is off-viewport - so carry the session and drop the query
-    // rather than reopening the list over a cell that is no longer being edited.
-    const editSession = this.#editSession;
+    this.#deferQuery();
+  }
 
-    this.hot._registerTimeout(() => {
-      if (editSession !== this.#editSession) {
-        return;
-      }
-
+  /**
+   * Defers a `queryChoices()` call and keeps its timer id so `close()` can cancel it.
+   *
+   * `hot._registerTimeout()` has no cancel path of its own - `_clearTimeouts()` runs only from
+   * `Core#destroy()` - so without this a query scheduled during an edit still fires after the
+   * editor closed, and starts a fresh request against a cell nobody is editing.
+   *
+   * @param {number} [delay] Delay in milliseconds.
+   */
+  #deferQuery(delay: number = 0): void {
+    const timeoutId = this.hot._registerTimeout(() => {
+      this.#queryTimeouts.delete(timeoutId);
       this.queryChoices(this.TEXTAREA.value);
-    });
+    }, delay);
+
+    this.#queryTimeouts.add(timeoutId);
   }
 
   /**
    * Closes the editor.
    */
   close(): void {
-    // Ends the edit session, which invalidates every query in flight and every query still waiting
-    // on a timeout. Closing is the one event that reliably means "no response is wanted any more":
-    // `state` stays `EDITING` when `refreshDimensions()` closes an editor whose cell scrolled out
-    // of the rendered range and when `afterSetTheme` closes one (`assignHooks`), and `_opened`
-    // stays false after that same cell scrolls back and the editor is shown again.
+    // The debounced refocus is armed by the inner grid's `afterScroll` and runs 100 ms later. It
+    // outlives the close by the same route the choices response used to, and `hideEditableElement()`
+    // only sets `opacity: 0`, so its `focus()` puts the caret back into a closed editor. The next
+    // scroll of a reopened list re-arms it, so cancelling here loses nothing.
+    this.#focusDebounced.cancel();
+
+    // Ends the edit session. Closing is the one event that reliably means "no response is wanted
+    // any more": `state` stays `EDITING` when `refreshDimensions()` closes an editor whose cell
+    // scrolled out of the rendered range and when `afterSetTheme` closes one (`assignHooks`), and
+    // `_opened` stays false after that same cell scrolls back and the editor is shown again.
+    //
+    // Queries this editor deferred are cancelled outright; the token below is for the ones already
+    // handed to user code, which cannot be.
+    this.#queryTimeouts.forEach(timeoutId => clearTimeout(timeoutId));
+    this.#queryTimeouts.clear();
+
     this.#editSession += 1;
 
     this.removeHooksByKey('beforeKeyDown');
@@ -358,7 +385,12 @@ export class AutocompleteEditor extends HandsontableEditor {
     // shows it again on the way back, without ever restoring the flag. A guard on it would leave
     // the suggestion list dead for the rest of such an edit. The close paths `state` cannot see are
     // handled by the edit-session token instead, carried by each caller across its timeout.
-    if (this.state !== EDITOR_STATE.EDITING) {
+    //
+    // `WAITING` counts as open. `close()` is what unhooks `beforeKeyDown`, so keystrokes still
+    // schedule queries while an async validator runs, and under `allowInvalid: false` the editor
+    // stays open and returns to `EDITING`. Rejecting them here would stop the list refreshing for
+    // the length of every validation, which is a behavior change rather than a fix.
+    if (this.state !== EDITOR_STATE.EDITING && this.state !== EDITOR_STATE.WAITING) {
       return;
     }
 
@@ -381,7 +413,12 @@ export class AutocompleteEditor extends HandsontableEditor {
         // state check - a response landing while an async validator holds the editor in `WAITING`
         // belongs to the still-open editor, and rejecting it would leave the list empty for the
         // rest of the edit when `allowInvalid: false` sends the state back to `EDITING`.
-        if (editSession !== this.#editSession || generation !== this.#queryGeneration) {
+        // `Core#destroy()` reaches neither token - it never closes the active editor - and
+        // `updateChoicesList()` would then touch an `htEditor` whose root element is gone. The
+        // guide tells people to answer late, and in a single-page app a torn-down grid is the
+        // usual way that happens.
+        if (this.hot.isDestroyed || editSession !== this.#editSession ||
+            generation !== this.#queryGeneration) {
           return;
         }
 
@@ -737,18 +774,7 @@ export class AutocompleteEditor extends HandsontableEditor {
         timeOffset += 10;
       }
 
-      // Carried across the delay for the same reason as in `open()`: a close landing inside the
-      // 10-20 ms window leaves `state` at `EDITING` on the scroll-out and theme paths, so without
-      // the session the timeout would start a fresh query against a closed editor.
-      const editSession = this.#editSession;
-
-      this.hot._registerTimeout(() => {
-        if (editSession !== this.#editSession) {
-          return;
-        }
-
-        this.queryChoices(this.TEXTAREA.value);
-      }, timeOffset);
+      this.#deferQuery(timeOffset);
     }
   }
 }
--- a/handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
+++ b/handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
@@ -79,13 +79,20 @@ export class AutocompleteEditor extends HandsontableEditor {
    */
   #idPrefix = this.hot.guid.slice(0, 9);
   /**
-   * Generation token for the in-flight choices query. Bumped on every `queryChoices()` call and on
-   * every `close()`, so a response that no longer belongs to the edit in progress can be told apart
-   * from the one the editor is waiting for.
+   * Generation token for the in-flight choices query. Bumped on every `queryChoices()` call, so a
+   * response belonging to a superseded query can be told apart from the one the editor is waiting
+   * for.
    *
    * @type {number}
    */
   #queryGeneration = 0;
+  /**
+   * Edit-session token. Bumped on every `close()`, so work scheduled during one edit can tell that
+   * the edit has ended - which neither `state` nor `_opened` reports reliably.
+   *
+   * @type {number}
+   */
+  #editSession = 0;
 
   /**
    * Gets current value from editable element.
@@ -286,7 +293,16 @@ export class AutocompleteEditor extends HandsontableEditor {
       setAttribute(this.TEXTAREA, ...A11Y_EXPANDED('true'));
     }
 
+    // The editor can be closed again before this fires - `refreshDimensions()` does exactly that
+    // from inside `open()` when the cell is off-viewport - so carry the session and drop the query
+    // rather than reopening the list over a cell that is no longer being edited.
+    const editSession = this.#editSession;
+
     this.hot._registerTimeout(() => {
+      if (editSession !== this.#editSession) {
+        return;
+      }
+
       this.queryChoices(this.TEXTAREA.value);
     });
   }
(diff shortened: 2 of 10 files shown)


qunabu:
Task linked: [DEV-2653 Autocomplete/dropdown: a late source callback can re-show a closed editor and steal focus](https://app.clickup.com/t/9015210959/DEV-2653)

handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
```diff
@@ -311,17 +343,48 @@ export class AutocompleteEditor extends HandsontableEditor {
...
+    // Both callers defer through `hot._registerTimeout()`, which has no cancel path - `close()`
+    // never clears the queue, so a timeout scheduled while the editor was open still fires after
+    // it closed. Bail rather than call the user's `source`, which is typically a network request.
+    //
+    // `state` rather than `isOpened()`: `_opened` is false for the whole of `open()`, and it stays
+    // false after `refreshDimensions()` closes an editor whose cell scrolled out of view and then
+    // shows it again on the way back, without ever restoring the flag. A guard on it would leave
+    // the suggestion list dead for the rest of such an edit. The close paths `state` cannot see are
+    // handled by the edit-session token instead, carried by each caller across its timeout.
+    if (this.state !== EDITOR_STATE.EDITING) {
```
Haxikowy: This also catches `WAITING`, and that is a behavior change against `develop`.

`onBeforeKeyDown` stays hooked while an async validator runs (`removeHooksByKey` only runs in `close()`). With `allowInvalid: false`, keystrokes during validation schedule a query, the timeout fires with `state === WAITING`, this returns, and nothing retries. The list stops refreshing until validation ends. `develop` refreshed it.

The callback guard at :384 deliberately accepts `WAITING` for the same reason. Suggest matching it:

```suggestion
    if (this.state !== EDITOR_STATE.EDITING && this.state !== EDITOR_STATE.WAITING) {
```

No case covers either direction.

handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
```diff
@@ -286,6 +311,13 @@ export class AutocompleteEditor extends HandsontableEditor {
    * Closes the editor.
    */
   close(): void {
+    // Ends the edit session, which invalidates every query in flight and every query still waiting
+    // on a timeout. Closing is the one event that reliably means "no response is wanted any more":
+    // `state` stays `EDITING` when `refreshDimensions()` closes an editor whose cell scrolled out
+    // of the rendered range and when `afterSetTheme` closes one (`assignHooks`), and `_opened`
+    // stays false after that same cell scrolls back and the editor is shown again.
+    this.#editSession += 1;
```
Haxikowy, line 353: `#focusDebounced` survives this close, and it steals focus the same way.

It is hooked to the inner grid's `afterScroll` at 100 ms. Arrow-key down a long list, press Escape within 100 ms, and it runs `focus()` afterwards. `hideEditableElement()` uses `opacity: 0`, not `display: none`, so `TEXTAREA.select()` still works and the caret lands back in a closed editor.

`debounce()` already returns a `cancel()`. One line here fixes it. Otherwise the changelog's "taking keyboard control back from the page" overstates what ships, since this is the same symptom by another route.

handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
```diff
@@ -277,7 +293,16 @@ export class AutocompleteEditor extends HandsontableEditor {
       setAttribute(this.TEXTAREA, ...A11Y_EXPANDED('true'));
     }
 
+    // The editor can be closed again before this fires - `refreshDimensions()` does exactly that
```
Haxikowy: The comment names a case this capture cannot see. `super.open()` runs first, and `refreshDimensions()` inside it (textEditor.ts:127, then handsontableEditor.ts:111) already called `close()` and bumped the session by the time line 299 reads it, so the tokens match and the timeout proceeds.

That turns out to be the right outcome: `open()` calls `showEditableElement()` right after, the editor is visibly open, and the query belongs. Moving the capture above `super.open()` would leave the list empty in that case.

So the code is fine and the comment is not. Suggest rewording to what the capture actually covers: a close between this `open()` and the timeout firing.

handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
```diff
@@ -311,17 +343,48 @@ export class AutocompleteEditor extends HandsontableEditor {
...
+
       (source as SourceFn).call(this.cellProperties, query, (choices: unknown[]) => {
+        // A user-supplied source answers whenever it likes, and `HandsontableEditor.close()` only
+        // hides the nested grid, so a late response can still re-show the dropdown and pull focus
+        // back through `hot.listen()`. Two ways a response stops being the one the editor waits
+        // for, one token each: the edit ended, or a newer query superseded it. Deliberately no
+        // state check - a response landing while an async validator holds the editor in `WAITING`
+        // belongs to the still-open editor, and rejecting it would leave the list empty for the
+        // rest of the edit when `allowInvalid: false` sends the state back to `EDITING`.
+        if (editSession !== this.#editSession || generation !== this.#queryGeneration) {
```
Haxikowy: `Core#destroy()` reaches neither token. It never closes the active editor (`EditorManager#destroy` only sets `destroyed` and kills its `eventManager`), so a `source` answering after teardown passes both checks and hits `updateChoicesList()`, where `htEditor.rootElement` is already null.

`develop` throws identically, so not a regression. It is worth handling here because the guide now tells people to "call `process()` whenever the request completes, even late", and in an SPA a destroyed grid is the usual way that happens. `|| this.hot.isDestroyed` would cover it.

tests/e2e/autocomplete-async-source.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+      expect(await grid.editorState()).toBe('STATE_EDITING');
+      expect(await grid.isDropdownShown()).toBe(false);
+
+      const listenCountBeforeResponse = await grid.listenCount();
+
+      expect(await grid.resolveQueries(0)).toBeGreaterThan(0);
+
+      await expect.poll(() => grid.isDropdownShown()).toBe(false);
+      expect(await grid.dropdownChoices()).toEqual([]);
+      expect(await grid.listenCount()).toBe(listenCountBeforeResponse);
```
Haxikowy, line 336: This assertion cannot fail. `Core#listen` runs `afterListen` only on a not-listening to listening transition (core.ts:2423), and the scroll scenario never unlists, so the count is pinned whether or not a stale response calls `listen()`.

Both scroll cases are affected, which is where the docblock's "both halves of the defect are asserted every time" needs it most. Either unlisten before reading the baseline, or drop the assertion and the claim.

tests/e2e/autocomplete-async-source.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+    test('ignores a source response from a previous edit after the editor reopened elsewhere', async() => {
+      await grid.openEditor(0, 0);
+
+      await grid.page.keyboard.press('Escape');
+
+      await expect.poll(() => grid.isEditorOpen()).toBe(false);
+
+      // The editor is a per-instance singleton, so this is the SAME object, now editing another
+      // column and back in the `EDITING` state. A guard that only reads the state cannot tell the
+      // stale response apart from the current one - the query generation is what does.
```
Haxikowy, line 101: The generation is not what tells them apart here. Escape already bumped `#editSession`, so the callback returns on the first half of the guard and `generation !== this.#queryGeneration` never decides anything in this suite.

Delete `#queryGeneration` from the editor and all 10 tests still pass. A case with two queries inside one edit session, resolved newest-first, would pin it. That needs a per-entry resolve in the fixture, since `htResolveQueries` answers oldest-first.

tests/e2e/autocomplete-async-source.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+      // timeout scheduled in the 10-20 ms before the close still fires afterwards, and `state` is
+      // `EDITING` by then, so it would start a FRESH query - current by every other measure - and
+      // re-show the list over the closed editor once that one answered.
+      await grid.scheduleQueryThenClose();
+
+      expect(await grid.editorState()).toBe('STATE_EDITING');
+      expect(await grid.isEditorOpen()).toBe(false);
+      expect(await grid.totalQueryCount()).toBe(queriesBefore);
+
+      await grid.resolveQueries(0);
```
Haxikowy: The only `resolveQueries` whose return value is not asserted. If nothing were in flight, the two assertions below would pass on an empty page.

```suggestion
      expect(await grid.resolveQueries(0)).toBeGreaterThan(0);
```

tests/fixtures/demo/autocomplete-async-source.html
```diff
@@ -0,0 +1,161 @@
...
+        window.htPendingQueries.push({ col, query, process, resolved: false });
+      };
+    }
+
+    /**
+     * Resolves every query still pending for a column, oldest first, and returns how many were
+     * resolved. Returns 0 when the column has nothing in flight, which lets a spec poll for the
+     * query to exist rather than assume it does.
+     */
+    window.htResolveQueries = function(col) {
```
Haxikowy, line 109: Queries from the editor and from the validator land in the same list. `dropdown` is `strict: true`, and `autocompleteValidator` calls the same user `source` on save, so after `clickOutsideInput()` column 0 can hold two pending entries.

That weakens the `resolveQueries(0) > 0` guard in cases 1 and 2, whose comment says a zero would mean the case proved nothing. The validator's query alone satisfies it. Tag each entry with its origin and assert the exact count.

tests/fixtures/demo/autocomplete-async-source.html
```diff
@@ -0,0 +1,161 @@
...
+      const editor = window.hot.getActiveEditor();
+      const keydown = new KeyboardEvent('keydown', { key: 'a', code: 'KeyA', bubbles: true });
+
+      // Chromium ignores the deprecated `keyCode` init dictionary member. The production handler
+      // deliberately still reads it for browser compatibility, so define it on this test event.
+      Object.defineProperty(keydown, 'keyCode', { value: 65 });
+      editor.TEXTAREA.dispatchEvent(keydown);
+      editor.close();
+
+      return new Promise(resolve => setTimeout(resolve, 50));
```
Haxikowy: A fixed 50 ms against an editor delay of 10-20 ms is 2.5x headroom on a loaded CI worker, so the comment above calling it an ordering guarantee is not right. It also sits page-side, where the `no-restricted-syntax` ban on `waitForTimeout` cannot see it.

Make the ordering observable instead: wrap `hot._registerTimeout` in the fixture, count callbacks that actually run, and poll that count from the page object.

docs/content/guides/cell-types/autocomplete-cell-type/autocomplete-cell-type.md
```diff
@@ -150,6 +150,8 @@ In strict mode, the [`allowInvalid`](@/api/options.md#allowinvalid) option deter
 
 Autocomplete can also use asynchronous data sources. In the example below, suggestions for the "Car" column are loaded from the server with the Fetch API. To load data from a remote source, assign a function to the `source` option. The function receives the query string and the `process` callback. Call `process()` with the result array when the request completes.
 
+Handsontable ignores a response that arrives after the editor closed, and it ignores a response that a newer query has already superseded. Call `process()` whenever the request completes, even late.
```
Haxikowy: The contract is right, but a reader cannot tell when it applies. "After the editor closed" does not suggest scrolling the edited cell out of view, which is the case this PR spends most of its argument on and the one where the editor still looks open. "Superseded by a newer query" does not say a query starts on every keystroke.

One clause each would fix it. The same two sentences are also missing from the `source` JSDoc in `metaSchema.ts`, which carries the `fetch()` example this changes, and from the dropdown guide, which the changelog names.

Haxikowy (commented):
The design argument holds up. I went looking for the case where keying on `close()` is wrong and did not find one, and both counters do earn their place, though only one of them is currently tested. The write-up is unusually good, in particular the disproof of the old blame in `editor-hidden-cell.spec.ts`.

Two things I would want changed before merge, both in inline comments: the new `state` guard also rejects `WAITING`, which stops the list refreshing while an async validator runs under `allowInvalid: false`, and `#focusDebounced` still steals focus after a close by a route the tokens do not cover.

**Not covered by any case**

- The `WAITING` window, in either direction. The callback guard is designed for it and the entry guard rejects it, and nothing tells you which is intended.
- `Core#destroy()` with a query in flight.
- `#queryGeneration` deciding a rejection. See the inline note on case 3.

**Follow-ups with no ticket**

The `DropdownEditor.finishEditing()` discard race, evidenced down to a CI run number, lives in a spec comment. `#focusDebounced` lives in this description. Both disappear when this merges. Worth two ClickUp tasks referenced from here.

**Changelog and docs**

Entry mechanics are correct, and `breaking: false` is defensible. Two gaps: `queryChoices()` is in the shipped TypeScript surface and now no-ops outside `EDITING`, which nothing states, and a `source` with side effects is no longer called at all on a post-finish timeout. Worth a line in the release notes even if the flag stays false.

**Verified**

Copied the three new test files onto an unfixed build and ran them: all 10 red, each at the right assertion line, including case 5 failing on the query count rather than incidentally. The `Core#listen` transition semantics, `autocompleteValidator` calling the user `source` on the strict path, and the `refreshDimensions()` ordering inside `open()` I checked against the source directly. I did not rerun the full suites, and I did not build a demo page.

**Description nit**

The variant table is introduced as "the same eight tests" and its first row reports 10/10, mixing the four-case and five-case suites without saying so. The rows also flip between counting failures ("case 5 fails (2/10)") and passes ("6/8"). One suite size per row and one direction would settle it.

handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
```diff
@@ -674,11 +737,18 @@ export class AutocompleteEditor extends HandsontableEditor {
...
-        this.hot._registerTimeout(() => {
-          this.queryChoices(this.TEXTAREA.value);
-        }, timeOffset);
-      }
+      // Carried across the delay for the same reason as in `open()`: a close landing inside the
+      // 10-20 ms window leaves `state` at `EDITING` on the scroll-out and theme paths, so without
+      // the session the timeout would start a fresh query against a closed editor.
+      const editSession = this.#editSession;
+
+      this.hot._registerTimeout(() => {
```
budnix: `hot._registerTimeout()` returns the timer id, so this could keep it in a field and call `clearTimeout()` in `close()` instead. The stale query then never runs at all, rather than every timer waking up just to compare two numbers. It also says "this query is cancelled" in the code, instead of leaving it encoded in a counter. The tokens would still be needed for the `source` callback at `:384`, since user code holds that one and there is nothing to cancel there — curious whether cancel was tried and ruled out?

tests/fixtures/pages/AutocompleteAsyncSourcePage.ts
```diff
@@ -0,0 +1,246 @@
...
+}
+
+interface HandsontableFixture {
+  getSelected(): number[][] | undefined;
+  getActiveEditor(): EditorFixture | undefined;
+  isListening(): boolean;
+  scrollViewportTo(options: { row: number, verticalSnap: string }): void;
+}
+
+interface FixtureWindow extends Window {
```
budnix: `extends Window` makes this an error. `windowTypes.ts` already declares `hot` globally as `FixtureHotInstance`, and `HandsontableFixture` has only 4 of its members, so a standalone `tsc` run over the two files gives:

```
error TS2430: Interface 'FixtureWindow' incorrectly extends interface 'Window'.
  Types of property 'hot' are incompatible.
    Type 'HandsontableFixture' is missing the following properties from type
    'FixtureHotInstance': getDataAtCell, getSourceDataAtCell, setDataAtCell,
    getCellMeta, and 26 more.
```

It passes today only because `tests/` has no `tsconfig.json` and its `lint` script runs no type rules. Dropping `extends Window` costs nothing, since every access here already goes through `window as unknown as FixtureWindow`:

```suggestion
interface FixtureWindow {
```

budnix (commented):
Two follow-ups on the file-level details, on top of the comments already on this PR.

handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
demtario: Confirmed and applied in 48a421d. The entry guard now reads `state !== EDITING && state !== WAITING`, matching the callback guard, and the comment says why: `close()` is what unhooks `beforeKeyDown`, so keystrokes keep scheduling queries while an async validator runs.

Both directions are covered now, in a new `?validator=slowAsync` scenario:

- case 6, "still refreshes the list while the editor waits on validation" — asserts the query reaches the `source` at all (the entry guard) and that its answer renders (the callback guard).
- case 7, "ignores a response from the validation window once the editor finally closes".

Case 6 fails 2/20 on a build with `WAITING` removed from the entry guard again.

handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
demtario: Fixed in 48a421d rather than deferred — you are right that it undercuts the changelog line, so shipping the claim without it was wrong. `close()` now calls `this.#focusDebounced.cancel()`.

Case 9, "does not refocus itself after the close", pins it. It watches the `focus()` call rather than `document.activeElement`: other machinery moves focus around a close, and my first attempt at this test asserted on `activeElement` and was measuring that instead. Wrapping `editor.focus` after the close counts only what runs later, and `#focusDebounced` resolves `this.focus` at call time so the wrapper still sees it. 2/20 red without the cancel.

handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
demtario: Correct, and the comment was the thing that was wrong. Traced it: `super.open()` runs first, `refreshDimensions()` inside it has already called `close()` and bumped the session by the time the capture reads it, so the tokens match and the timeout proceeds — which is the right outcome, because `showEditableElement()` has since made the editor visible again.

That whole capture is gone in 48a421d. Following budnix on the same lines, both deferred sites now store the timer id and `close()` clears it, so the query is cancelled outright instead of waking up to compare two numbers. The comment on the new `#deferQuery()` describes only what it does.

handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
demtario: Confirmed and fixed in 48a421d: `this.hot.isDestroyed ||` leads the callback guard. `EditorManager#destroy()` really does only set `destroyed` and kill its `eventManager`, so neither token moves.

Case 8, "swallows a source response that arrives after the grid was destroyed", captures `pageerror` and asserts none. 2/20 red without the check. Agreed on the reasoning for handling it here rather than calling it pre-existing: the guide now tells people to answer late, and a torn-down grid in an SPA is the usual way that happens.

tests/e2e/autocomplete-async-source.spec.ts
demtario: Right, and it applied to both scroll cases. `Core#listen()` fires `afterListen` only on a not-listening to listening transition, and a scroll never unlistens, so the count was pinned and that half of the assertion could not fail.

Fixed in 48a421d by handing keyboard control away first: the scroll cases now `clickOutsideInput()` before reading the baseline, then assert both the unchanged count and `isGridListening() === false`. The docblock claim that both halves are asserted every time is true again.

tests/e2e/autocomplete-async-source.spec.ts
demtario: Correct — `#editSession` short-circuits first, so the generation half never decided anything. I checked your stronger claim too: deleting `#queryGeneration` left all tests green.

Case 10 in 48a421d pins it: two queries in one edit session, resolved newest-first. It needed the fixture work you predicted — `htResolveQueryAt(index)` for per-entry resolution, plus an `ordering` scenario whose answers are tagged with the index of the query they belong to, so the rendered list names which one won. The tag keeps the column prefix so it survives the `autocomplete` filter unchanged.

2/20 red with the generation check removed from the callback.

tests/e2e/autocomplete-async-source.spec.ts
demtario: Fixed in 48a421d, and the same hole existed one level deeper — see the reply on the fixture `source`. Cases 1, 2 and 4 now also assert `editorQueryCount(0) === 1` before resolving, so a validator query alone cannot satisfy the guard.

tests/fixtures/demo/autocomplete-async-source.html
demtario: Confirmed: `autocompleteValidator` calls the user `source` on the strict path, so a `dropdown` save adds a second entry for the same column and the `> 0` guard could be met without the editor ever having queried.

48a421d records the editor `state` on every pending entry rather than a single origin boolean — no one boolean is right for every case, and the new `WAITING`-window case is a genuine editor query that any "is the editor asking" flag would have mislabelled. `editorQueryCount()` counts entries captured in `STATE_EDITING`, and cases 1, 2 and 4 assert it is exactly 1.

tests/fixtures/demo/autocomplete-async-source.html
demtario: Agreed on both counts — the comment claimed a guarantee it did not have, and being page-side is exactly why the determinism rule could not catch it.

Replaced in 48a421d, though not with the counter: cancelling the timeout (budnix's point on the editor) means the callback no longer runs, so counting runs would have deadlocked the wait. It now waits on a sentinel timer armed strictly later than the editor's, with the larger of its two delays. That is an ordering fact about two deadlines, and it holds whether the editor's timer ran or was cancelled — which is the distinction these cases turn on.

docs/content/guides/cell-types/autocomplete-cell-type/autocomplete-cell-type.md
demtario: Agreed, and applied in 48a421d. Both sentences now name the cases: "including a close you may not notice, such as scrolling the edited cell out of view", and "which happens on every keystroke".

Added in all three places you listed — the autocomplete guide, the `source` JSDoc in `metaSchema.ts` next to the `fetch()` example, and the dropdown guide, which had no async section at all and now links to the autocomplete one.

handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
demtario: Cancel was not tried — I went to a token first and never revisited it. You are right, and it is applied in 48a421d.

Both deferred sites now go through a `#deferQuery()` helper that keeps the timer id in `#queryTimeouts`, deletes it when the callback runs, and `close()` clears the set. A `Set` rather than a single field because each keystroke arms its own timeout and several can be pending at once. `#editSession` stays exactly where you said it has to, on the `source` callback.

One knock-on worth flagging: my test waited for the timeout to run as its ordering signal, which cancelling makes impossible, so it now waits on a sentinel timer armed later than the editor's. That is a better signal anyway — it holds whether the editor's timer ran or was cancelled.

Case 5 is 2/20 red with the cancellation removed.

tests/fixtures/pages/AutocompleteAsyncSourcePage.ts
demtario: Applied in 48a421d. Dropped `extends Window` and left a comment saying why, so nobody adds it back: `windowTypes.ts` already declares `hot` globally with the full instance type, and narrowing it here to the handful of members these probes use is the TS2430 conflict you quoted.

Also confirmed your point about why it passed — `tests/` has no `tsconfig.json` and `npm run lint` there runs no type rules, so nothing in the local gates or CI would have caught it.

demtario:
Answering the two points from the review summary that have no inline thread.

**Follow-up tickets** — both created in Sprint 14 and assigned. Deliberately not linked or named here, so the ClickUp integration does not move them to code review just for being mentioned.

One is the `DropdownEditor.finishEditing()` discard race, carrying the evidence down to the CI run and job name, and the definition of done includes tightening the dropdown case in `editor-hidden-cell.spec.ts` and verifying on all six legs rather than one.

The second is **not** `#focusDebounced` — that is fixed in this PR (48a421d), so a ticket would have been stale on arrival. It covers the other thing this work evidenced and that currently lives only in a comment: on an empty cell, clicking it and pressing <kbd>Enter</kbd> opens the `autocomplete`/`dropdown` editor and closes it again on the same keypress. Reproduced on `develop`, hook trace `afterBeginEditing (EDITING) -> beforeChange (WAITING) -> afterChange (VIRGIN)`, specific to click-then-Enter and to an empty value. It is written down today only as a fixture-authoring warning in `tests/AGENTS.md`, which helps the next test author and does nothing for the user. Happy to swap it for something else if you had a different second one in mind.

**Description nit** — fixed. One suite size and one direction throughout, and every row is a build I actually ran rather than one I remembered: the `#editSession` row turned out to be 10/20 across five cases, not the 6/20 across three I had first written.

budnix (dismissed):

Haxikowy (approved):

Issue #185, opened 2012-10-12
Two Questions


kmcglinn:
Hi Warpech,

I am using your handsontables library and finding it very useful. I have two questions. The first is with respect to the CSS. I seem to be having difficulty placing the tables I create inside my own divs without them displaying very strange. Do you have any hints on how best to manage the css to do what I would like?

Secondly, with respect to the auto fill. Is there any way to increase the size of the autofill table to include more options?

All help is much appreciated!

Rud.

PR #13274, opened 2026-08-27
DEV-2659: Stop mergeCells re-clearing already-applied merge areas (#7555)

### LINKED ISSUE #7555: Redux example doesn't work with mergeCells
### Description

As mentioned in our forum https://forum.handsontable.com/t/react-redux-render-with-mergecells-prop-triggers-beforechange-callback-call/3928 once we add `mergeCells` to the [official Redux example](https://handsontable.com/docs/7.2.2/frameworks-wrapper-for-react-redux-example.html) we end up with an infinite loop.

### Steps to reproduce

1. Do nothing, just check the console

### Demo

https://handsontable.com/docs/7.2.2/frameworks-wrapper-for-react-redux-example.html official example
https://jsfiddle.net/AMBudnik/pftvn5qm/ the same code with `mergeCells` line `11`

### Your environment
* React wrapper version: 3.1.0
* Handsontable version: 7.2.2
* Browser Name and version: Chrome 78

budnix:
### Context

The PR fixes [#7555](https://github.com/handsontable/handsontable/issues/7555) — open since 2019 — where adding `mergeCells` to the official Redux example produces an infinite loop of `beforeChange` calls.

Re-verified before fixing, on released **18.0.0** and on **develop `f62994a`**: it reproduces identically on both, so nothing had fixed it since the report.

| Build | `mergeCells` | `beforeChange` calls | renders | outcome |
|---|---|---|---|---|
| 18.0.0 | yes | 53 | 54 | React error #185 "Maximum update depth exceeded" |
| develop | yes | 53 | 54 | identical |
| develop | **no** (control) | **0** | 1 | clean |

The loop never settles. React's own nested-update guard aborts it at ~50 updates and leaves the grid broken. The control run shows the single `mergeCells` line is the cause.

#### Root cause

The fault is in core, not in the wrapper. With an **identical** `mergeCells` value, three `updateSettings` calls that carry the key produce three extra emissions; the same three calls with the key omitted produce zero. The trigger is the key's mere presence.

1. `generateFromSettings()` writes `null` into every non-anchor cell through `setDataAtCell(..., source: 'MergeCells')`.
2. `updatePlugin()` re-runs it whenever the `updateSettings` payload merely *contains* `mergeCells` — `BasePlugin#isRelevantToSettings` only tests `!== undefined`.
3. The React wrapper (`SettingsMapper.getSettings`) and the Angular wrapper resend every prop on each render, so the key always arrives. **Vue 3 is immune**, because `prepareSettings` deep-compares and drops unchanged keys.
4. Core never filters no-op changes, so a `null` → `null` write still fires the hook.

The Redux recipe's reducer returns a new state object on every dispatch, which re-renders and closes the circle back to step 3.

#### The fix

`updatePlugin()` snapshots the merge areas already registered in the collection **before** `disablePlugin()` empties it, and threads that set into `generateFromSettings()`. An area whose `row,col,rowspan,colspan` key is in the set is still re-merged — the collection was cleared, so skipping `mergeRange()` would drop the merge entirely — but it skips the data population, because its cells were cleared when it was first applied.

**First application is byte-identical to before.** Only re-application of an already-applied area changes. Extending the configuration now populates only the newly added areas, where previously it re-populated every area.

Two alternatives were considered and rejected:

- *Skip cells that already hold `null`.* Clean argument, but `mergeCells.spec.js` deliberately pins that those no-op writes **are** emitted on a fresh grid whose covered cells are already null. That existing test would have to be flipped to make the fix pass.
- *Make `updatePlugin()` a no-op when the config is unchanged.* Skips the bounds re-validation that `generateFromSettings()` performs through `validateSetting()`.

#### One deliberate behavior change — please review

`updateSettings({ data, mergeCells })` no longer re-clears the covered cells in the new data. This removes an inconsistency rather than creating one: `loadData()` never re-cleared them, because the plugin registers no `afterLoadData` hook. Measured on both builds:

| path | before | after |
|---|---|---|
| `loadData(newData)` | values kept | values kept |
| `updateSettings({ data: newData, mergeCells: same })` | values **cleared** | values kept |

This cannot be avoided while fixing the loop: a Redux store hands back a new data array on every dispatch, so any "re-populate when the data changed" rule would loop exactly as before.

Known non-improvement, unchanged from today and not engineered around: a merge re-anchored by row trimming no longer matches a resent config key, so that path still re-populates.

#### Docs recipe

The Redux guide's reducer now returns the **previous state object** when no value actually changed, which is the idiomatic Redux practice and stops the loop on its own even without the core fix. A `source` filter was considered and rejected — it would keep the covered cells' data instead of clearing it, silently changing documented merge semantics for anyone following the recipe. The guide gained a tip covering the guard and the `beforeChange` `source` argument, and a stale line naming `afterChange` where the example uses `beforeChange` was corrected.

### How has this been tested?

Three cases were added to the existing `mergeCells updateSettings` describe block. The first two were run against the unfixed build first and **both failed there**, as intended:

1. unchanged config re-applied → **0** emissions (was 1) — the regression test;
2. config extended → only the new area's cells emitted (was: every area's);
3. span changed → still populates (guard, passes either way).

```bash
# The merge suite, including the three new cases
npm run test:e2e --prefix handsontable -- --testPathPattern=mergeCells   # 518 specs, 0 failures

# Full gates
npm run test:e2e --prefix handsontable                                    # 9725 specs, 0 failures
npm run test:unit --prefix handsontable                                   # 3963 passed
npm run test:types --prefix handsontable                                  # pass
npm run eslint --prefix handsontable                                      # 0 errors
npm run test --prefix wrappers/react-wrapper                              # 78 passed
```

Manually verified in real Chromium with the official Redux recipe plus the reporter's added `mergeCells` line, against a local build: `beforeChange` fires **1** time instead of 53, no React #185, and the store is nulled exactly once — the documented merge behavior is preserved.

### Types of changes

### Related issue(s):

1. DEV-2659
2. #7555

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2659

---

> [!NOTE]
> **Medium Risk**
> Changes core merge clearing on `updateSettings`, which affects Redux/React integrations and any code listening to `MergeCells`-sourced changes; behavior is well-tested but alters long-standing re-clear semantics for unchanged config with new data.
> 
> **Overview**
> Fixes an infinite **React/Redux** loop ([#7555](https://github.com/handsontable/handsontable/issues/7555)) where framework wrappers resend the same `mergeCells` on every render and the plugin kept writing `null` into already-empty covered cells, firing `beforeChange`/`afterChange` with source `MergeCells` even when nothing changed.
> 
> The **mergeCells** plugin now remembers which merge areas (by visual `row,col,rowspan,colspan` keys via `toMergeAreaKey`) have already had their clearing pass. On `updateSettings()` with the same config it still re-merges the range but only queues `setDataAtCell` clears for covered cells whose **stored** value is not `null` (`#getStoredValueAt`, so `valueGetter` does not mask cleared cells). First application and newly added or resized merge areas behave as before; new `data` under an unchanged merge still clears values that reappear (sort/filter/row move).
> 
> Docs add a “re-applying the same configuration” section, **React-Redux** example reducers return the previous state when no cell value actually changed, plus guidance on plugin `source` values. Tests cover re-apply silence, column move, `valueGetter`, cancelled clears, new data, sorting, and span changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2d00085f38a59c2b2f10d36fbde9a400d8792926. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changelogs/7555.json
+++ b/.changelogs/7555.json
@@ -1,6 +1,6 @@
 {
   "issuesOrigin": "public",
-  "title": "Fixed a bug where re-applying an unchanged `mergeCells` configuration through `updateSettings()` cleared the merged-away cells again, emitting `beforeChange` and `afterChange` for values that did not change. In React and Angular, which resend every setting on each render, this looped until the framework aborted it. A merge area that was already applied now keeps its merge but skips the data population, extending the configuration populates only the newly added areas, and `updateSettings({ data, mergeCells })` now leaves the new data alone just as `loadData()` already did.",
+  "title": "Fixed a bug where re-applying an unchanged `mergeCells` configuration through `updateSettings()` cleared the merged-away cells again, firing `beforeChange` and `afterChange` for values that did not change. A re-applied merge area now clears only the cells that still hold a value.",
   "type": "fixed",
   "issueOrPR": 7555,
   "breaking": false,
--- a/.changelogs/7555.json
+++ b/.changelogs/7555.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed a bug where re-applying an unchanged `mergeCells` configuration through `updateSettings()` cleared the merged-away cells again, emitting `beforeChange` and `afterChange` for values that did not change. In React and Angular, which resend every setting on each render, this looped until the framework aborted it. A merge area that was already applied now keeps its merge but skips the data population, extending the configuration populates only the newly added areas, and `updateSettings({ data, mergeCells })` now leaves the new data alone just as `loadData()` already did.",
+  "type": "fixed",
+  "issueOrPR": 7555,
+  "breaking": false,
+  "framework": "none"
+}
--- a/docs/content/guides/cell-features/merge-cells/merge-cells.md
+++ b/docs/content/guides/cell-features/merge-cells/merge-cells.md
@@ -352,6 +352,25 @@ Both [`getData()`](@/api/core.md#getdata) and [`getSourceData()`](@/api/core.md#
 
 Unmerging does not restore the cleared values. If you need the original values back, keep a copy before merging, or restore them yourself in a [`beforeUnmergeCells`](@/api/hooks.md#beforeunmergecells) or [`afterUnmergeCells`](@/api/hooks.md#afterunmergecells) handler.
 
+### Clearing runs once per merged range
+
+A merged range clears its covered cells when it is first applied, not every time the configuration is read. Passing the same `mergeCells` value to [`updateSettings()`](@/api/core.md#updatesettings) again keeps the merge and changes nothing else: no cells are cleared a second time, and no [`beforeChange`](@/api/hooks.md#beforechange) or [`afterChange`](@/api/hooks.md#afterchange) event fires. Adding a range to the configuration clears only that new range.
+
+This matters when a framework wrapper resends every option on each render. React and Angular do, so without it a store-driven app would keep receiving change events for values that never changed.
+
+One consequence: if you pass new `data` alongside an unchanged `mergeCells` value, the covered cells of the new data keep their values.
+
+```js
+hot.updateSettings({
+  data: [['A1', 'B1'], ['A2', 'B2']],
+  mergeCells: [{ row: 0, col: 0, rowspan: 2, colspan: 2 }], // unchanged
+});
+
+hot.getDataAtCell(0, 1); // -> 'B1', not null
+```
+
+[`loadData()`](@/api/core.md#loaddata) and [`updateData()`](@/api/core.md#updatedata) behave the same way. To clear the covered cells of a fresh dataset, unmerge the range and merge it again, or clear those cells yourself.
+
 ## Effect on viewport getter methods
 
 With merged cells, the rendered range extends to fit any merged cell that crosses the viewport edge. This is the same expansion that the `virtualized` option turns off. As a result, the rendered-range getters can return indexes beyond what you see on the screen:
--- a/docs/content/guides/cell-features/merge-cells/merge-cells.md
+++ b/docs/content/guides/cell-features/merge-cells/merge-cells.md
@@ -352,13 +352,15 @@ Both [`getData()`](@/api/core.md#getdata) and [`getSourceData()`](@/api/core.md#
 
 Unmerging does not restore the cleared values. If you need the original values back, keep a copy before merging, or restore them yourself in a [`beforeUnmergeCells`](@/api/hooks.md#beforeunmergecells) or [`afterUnmergeCells`](@/api/hooks.md#afterunmergecells) handler.
 
-### Clearing runs once per merged range
+### Re-applying the same configuration
 
-A merged range clears its covered cells when it is first applied, not every time the configuration is read. Passing the same `mergeCells` value to [`updateSettings()`](@/api/core.md#updatesettings) again keeps the merge and changes nothing else: no cells are cleared a second time, and no [`beforeChange`](@/api/hooks.md#beforechange) or [`afterChange`](@/api/hooks.md#afterchange) event fires. Adding a range to the configuration clears only that new range.
+Re-applying a `mergeCells` value through [`updateSettings()`](@/api/core.md#updatesettings) clears only the cells that still hold a value. A range whose cells are already empty changes no data, so it fires no [`beforeChange`](@/api/hooks.md#beforechange) or [`afterChange`](@/api/hooks.md#afterchange) event.
 
-This matters when a framework wrapper resends every option on each render. React and Angular do, so without it a store-driven app would keep receiving change events for values that never changed.
+This depends on the clearing write reaching the data. If you cancel it -- by returning `false` from `beforeChange`, or with a validator that rejects `null` while [`allowInvalid`](@/api/options.md#allowinvalid) is `false` -- the covered cells keep their values, and every re-apply tries to clear them again.
 
-One consequence: if you pass new `data` alongside an unchanged `mergeCells` value, the covered cells of the new data keep their values.
+This matters when a framework wrapper resends every option on each render. React and Angular do. Without it, an app that writes those events back into a store keeps receiving changes for values that never changed, and the two can keep triggering each other.
+
+Nothing else about the clearing changes. A range still clears its covered cells the first time you apply it, even where those cells are already empty. And if new values arrive in a covered range -- because you passed new `data`, or because sorting, filtering, or a row move brought other rows under the range -- the next re-apply clears them, then stays quiet:
 
 ```js
 hot.updateSettings({
@@ -366,10 +368,8 @@ hot.updateSettings({
   mergeCells: [{ row: 0, col: 0, rowspan: 2, colspan: 2 }], // unchanged
 });
 
-hot.getDataAtCell(0, 1); // -> 'B1', not null
+hot.getDataAtCell(0, 1); // -> null, cleared as usual
 ```
-
-[`loadData()`](@/api/core.md#loaddata) and [`updateData()`](@/api/core.md#updatedata) behave the same way. To clear the covered cells of a fresh dataset, unmerge the range and merge it again, or clear those cells yourself.
 
 ## Effect on viewport getter methods
 
--- a/handsontable/src/dataMap/metaManager/metaSchema.ts
+++ b/handsontable/src/dataMap/metaManager/metaSchema.ts
@@ -4296,6 +4296,12 @@ export default (): Record<string, unknown> => {
      * This option can only be set at the [grid level](@/guides/getting-started/configuration-options/configuration-options.md#set-grid-options).
      * It has no effect when set in the [`columns`](#columns), [`cells`](#cells), or [`cell`](#cell) options.
      *
+     * A merged range clears the cells it covers when it is first applied. Passing the same value to
+     * [`updateSettings()`](@/api/core.md#updatesettings) again keeps the merge but clears nothing a
+     * second time, and fires no `beforeChange` or `afterChange` event. Adding a range to the array
+     * clears only that new range. As a result, passing new `data` alongside an unchanged `mergeCells`
+     * value leaves the covered cells of that new data intact, the same way `loadData()` does.
+     *
      * Read more:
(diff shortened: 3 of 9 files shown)


qunabu:
Task linked: [DEV-2659 Fix mergeCells re-clearing merged cells on every updateSettings (#7555)](https://app.clickup.com/t/9015210959/DEV-2659)

.changelogs/7555.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed a bug where re-applying an unchanged `mergeCells` configuration through `updateSettings()` cleared the merged-away cells again, emitting `beforeChange` and `afterChange` for values that did not change. In React and Angular, which resend every setting on each render, this looped until the framework aborted it. A merge area that was already applied now keeps its merge but skips the data population, extending the configuration populates only the newly added areas, and `updateSettings({ data, mergeCells })` now leaves the new data alone just as `loadData()` already did.",
+  "type": "fixed",
+  "issueOrPR": 7555,
+  "breaking": false,
```
demtario, line 6: **High.** `"breaking": false`, but this changes documented behavior of a public API path: `updateSettings({ data, mergeCells })` used to null the covered cells of the freshly passed data, and now it does not. `.ai/BREAKING-CHANGES.md` lists "Changing API signatures or behavior" as breaking, and the root `AGENTS.md` requires a migration-guide step for one.

The change looks deliberate: it is codified in a test and documented in three places. So the ask is classification, not a code change. Either set `"breaking": true` and add a step under `docs/content/guides/upgrade-and-migration/`, or record in the PR description the decision that the old null-fill was an unintended side effect outside the policy.

Worth noting for that decision: this is data retention, not loss. The values survive in `getData()`, `getSourceData()`, copy, and export, but are invisible in the grid.

docs/content/guides/cell-features/merge-cells/merge-cells.md
```diff
@@ -352,6 +352,25 @@ Both [`getData()`](@/api/core.md#getdata) and [`getSourceData()`](@/api/core.md#
 
 Unmerging does not restore the cleared values. If you need the original values back, keep a copy before merging, or restore them yourself in a [`beforeUnmergeCells`](@/api/hooks.md#beforeunmergecells) or [`afterUnmergeCells`](@/api/hooks.md#afterunmergecells) handler.
 
+### Clearing runs once per merged range
+
+A merged range clears its covered cells when it is first applied, not every time the configuration is read. Passing the same `mergeCells` value to [`updateSettings()`](@/api/core.md#updatesettings) again keeps the merge and changes nothing else: no cells are cleared a second time, and no [`beforeChange`](@/api/hooks.md#beforechange) or [`afterChange`](@/api/hooks.md#afterchange) event fires. Adding a range to the configuration clears only that new range.
```
demtario: **Medium.** This states the rule unconditionally: "no cells are cleared a second time, and no `beforeChange` or `afterChange` event fires." The PR's own test disproves it. `should settle after one re-population when the mergeCells config is re-applied over a sorted grid` asserts `beforeChange.calls.count()).toBe(1)` for exactly that call.

The skip is keyed off the merge's live visual geometry, which gets re-anchored on any trim, hide, or index-sequence change, so the key stops matching the declared one.

Suggested caveat: after a sort, a row or column move, or a filter changes the merge's visual position, the next re-apply clears and fires once more, and stays silent from then on.

handsontable/src/dataMap/metaManager/metaSchema.ts
```diff
@@ -4296,6 +4296,12 @@ export default (): Record<string, unknown> => {
      * This option can only be set at the [grid level](@/guides/getting-started/configuration-options/configuration-options.md#set-grid-options).
      * It has no effect when set in the [`columns`](#columns), [`cells`](#cells), or [`cell`](#cell) options.
      *
+     * A merged range clears the cells it covers when it is first applied. Passing the same value to
```
demtario: **Medium.** Same unconditional claim as in the guide, and this one ships as the public API reference: "keeps the merge but clears nothing a second time, and fires no `beforeChange` or `afterChange` event." The sorted-grid test in this PR asserts one populate after a sort. Mirror whatever caveat lands in `merge-cells.md`.

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -311,10 +311,16 @@ export class MergeCells extends BasePlugin {
    *  - [`mergeCells`](@/api/options.md#mergecells)
    */
   updatePlugin() {
+    // Snapshot the applied areas before `disablePlugin()` empties the collection, so
+    // `generateFromSettings()` can tell a re-applied area from a newly declared one.
+    const alreadyAppliedMerges = new Set(
+      this.mergedCellsCollection.mergedCells.map(mergedCell => toMergeAreaKey(mergedCell))
```
demtario: **Medium.** The snapshot is derived from `mergedCellsCollection.mergedCells`, which is the live, possibly re-anchored geometry, not a record of which areas were actually populated. `#reanchorMergesToVisibleRows()` rewrites a merge's visual `row`/`col` on any trim, hide, or index-sequence change, so after a sort or a move the live key no longer matches the key the config declares, and the next re-apply still runs one destructive population pass at the configured coordinates.

That pass predates this PR (develop runs it on every `updateSettings`, and this narrows it to one), so this is the fix stopping one step short rather than a new defect.

Fix: store the config keys this call actually populated in an instance field and compare the incoming config against that field, instead of deriving keys from the live collection. Sort, move, and trim then stop affecting the skip decision, and the sorted case in the new spec becomes `toBe(0)`.

handsontable/src/plugins/mergeCells/__tests__/mergeCells.spec.js
```diff
@@ -452,6 +452,168 @@ describe('MergeCells', () => {
...
+      // Sorting re-anchors the merge onto its new visual position, so the area no longer matches the
+      // key the config declares. The next re-apply therefore populates once more — and after that the
+      // collection is back in sync with the config, so further re-applies stay silent.
+      getPlugin('columnSorting').sort({ column: 2, sortOrder: 'asc' });
+
+      beforeChange.calls.reset();
+
+      await updateSettings({ mergeCells });
+
+      expect(beforeChange.calls.count()).toBe(1);
```
demtario, line 677: **Low.** This test asserts only spy counts, unlike the other four added here, which also assert `getData()`. It cannot distinguish a correct re-population from one that writes nulls into rows the merge does not cover, and it locks the current behavior in without recording what that behavior is.

Two independent traces of this fixture disagreed on which cells the re-apply nulls, which is precisely what a data assertion would settle. Adding `expect(getData()).toEqual([...])` after the sort and after the re-apply would pin it down.

demtario (commented):
5 inline notes below.

.changelogs/7555.json
budnix: Good point — instead of reclassifying, 6626c65458 makes it non-breaking: a re-apply now clears only the cells that still hold a value, so the data outcome matches develop everywhere.

docs/content/guides/cell-features/merge-cells/merge-cells.md
budnix: You're right, that was too absolute — 6626c65458 rewrites it to say a re-apply clears only the cells that still hold a value, and adds that it depends on the clearing write actually landing.

handsontable/src/dataMap/metaManager/metaSchema.ts
budnix: Nice catch — 6626c65458 gives the API reference the same wording and the same caveat as the guide.

handsontable/src/plugins/mergeCells/mergeCells.ts
budnix: Agreed — 6626c65458 keys the check off the declared settings in `#appliedMergeKeys` instead of the live collection. The sorted case stays at 1 though: keeping it non-breaking means real values that move under the area still get cleared once.

handsontable/src/plugins/mergeCells/__tests__/mergeCells.spec.js
budnix: Fair point — 6626c65458 adds `getData()` and the exact `beforeChange` args to that test, so the cells it clears are pinned now.

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -459,6 +494,28 @@ export class MergeCells extends BasePlugin {
...
+  #getStoredValueAt(row: number, column: number): unknown {
+    const physicalRow = this.hot.toPhysicalRow(row);
+    const physicalColumn = this.hot.toPhysicalColumn(column);
+
+    if (physicalRow === null || physicalColumn === null) {
+      return undefined;
+    }
+
+    return this.hot.getSourceDataAtCell(physicalRow, physicalColumn);
+  }
```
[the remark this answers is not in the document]
demtario: Confirmed accurate. `getSourceDataAtCell()` takes a physical row and a *visual* column (`core.ts:4338-4339`, `dataSource.ts:323-324`, which then calls `colToProp`, itself visual-in per `dataMap.ts:250`), so the column is translated twice here.

Three details worth pinning before triage:

- The trigger is `manualColumnMove` / `manualColumnFreeze`, not `hiddenColumns`. The latter uses a `HidingMap`, which leaves visual/physical identity, so a check against hidden columns will look clean.
- There is a third failure path: when the double-translated index falls out of range, `toPhysicalColumn()` returns `null`, `colToProp()` passes the number through unchanged (`dataMap.ts:263-264`), and the read comes back `undefined`. That is `!== null`, so it forces a re-clear, which is the loop this PR exists to fix.
- The same misuse already exists at `mergeCells.ts:628-629`. That one is pre-existing and runs only on unmerge, so it does not cover this site, which runs on every re-apply.

No test combines `mergeCells` with a moved column. The new `valueGetter` spec uses `columns` without reordering, which masks it.

demtario (approved):

PR #13277, opened 2026-08-27
DEV-2664: Stop an Angular editor of true throwing on cell select

budnix:
### Context

The PR fixes the Angular wrapper throwing `Only strings and functions can be passed as "editor" parameter` as soon as a cell is selected, whenever the `editor` setting is `true`.

Core declares `editor?: string | ctor | boolean`, so `editor: true` type-checks. At runtime core accepts only a string or a constructor, so a bare `true` reaches `getEditorInstance()` and throws on the first editor preparation — which happens on cell select, not on first edit.

The wrapper's `HotSettingsResolver.updateColumnEditorForGivenCustomEditor()` only rewrites `editor` when it holds an Angular editor component type. A boolean matches neither check, so the method returns early and the value reaches core untouched. A grid-level `editor` was never inspected at all.

`applyCustomSettings()` now drops an `editor` of `true` from the grid settings and from every column, before any other resolution runs. It is the single funnel for both the initial settings and `ngOnChanges`, so the update path is covered too.

**`true` is treated as "no editor named", not as "use the default editor".** Resolving it to the text editor would silently discard the editor a column's `type` supplies, turning `columns: [{ type: 'numeric', editor: true }]` into a plain text cell. Dropping the key instead lets the cell inherit from its `type`, the grid, or the default. This matches the semantic the React wrapper uses in #13268.

`editor: false` is untouched — core reads it as "editing disabled" — and this half of the problem never affected the Angular wrapper.

**On `editor: null`:** left as-is, so it keeps locking the cell exactly as vanilla Handsontable does. `null` sits outside the declared `string | ctor | boolean` type, so this is a lenience choice rather than a correctness one. The React wrapper inherits instead, but that comes from its prop layer, which the Angular wrapper does not have. Worth aligning with the Vue 3 wrapper's call on the same question (DEV-2663).

**Relationship to DEV-2661:** this is a standalone wrapper fix. DEV-2661 proposes normalizing `editor: true` in core, which would fix vanilla, Vue and Angular at once. That has not landed. If it does, this wrapper-side normalization becomes redundant rather than conflicting — it would simply never find a `true` left to drop.

No documentation change here: `metaSchema` documents only `false`, and updating it is DEV-2661's scope.

### How has this been tested?

Added 12 tests across two existing specs. Verified they are meaningful by reverting the source change and re-running: 8 of the 12 fail without the fix, including the discriminating `type: 'numeric'` case and the settings-update path. The other 4 are regression guards for `false`, `null` and component editors, which pass either way by design.

`projects/hot-table/src/lib/services/hot-settings-resolver.service.spec.ts` — resolver level:

- drops a grid-level `editor: true`
- drops a column-level `editor: true`, leaving a control column untouched
- keeps a column's `type` when its `editor: true` is dropped
- leaves `editor: false` alone at both levels
- leaves `editor: null` alone
- does not disturb a component editor
- tolerates a `columns` function alongside a grid-level `editor: true`

`projects/hot-table/src/lib/hot-table.component.spec.ts` — real `Handsontable.Core` instance through `TestBed`:

- grid-level `editor: true` no longer throws on `selectCell`, and `getCellEditor(0, 0)` is `TextEditor`
- column-level `editor: true` behaves the same, and the control column is unchanged
- `{ type: 'numeric', editor: true }` keeps `NumericEditor`
- `editor: false` still disables editing
- `editor: true` arriving through `ngOnChanges` is normalized on the update path too

```bash
npm --prefix wrappers/angular-wrapper run lint
```

Full Angular wrapper suite: 303 tests, 13 suites, all passing. Lint clean.

### Types of changes

### Related issue(s):

1. DEV-2664
2. DEV-2661 — proposed core-level normalization that would make this wrapper change redundant
3. DEV-2663 — the same gap in the Vue 3 wrapper
4. #13268 — the React wrapper fix this mirrors

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2664

---

> [!NOTE]
> **Low Risk**
> Scoped to Angular settings normalization for `editor: true`; other editor values and core Handsontable behavior are explicitly preserved by tests.
> 
> **Overview**
> Fixes the Angular wrapper throwing **"Only strings and functions can be passed as editor parameter"** when a cell is selected with **`editor: true`** on the grid or a column.
> 
> **`HotSettingsResolver`** now runs **`normalizeEditorSetting`** early in **`applyCustomSettings`**, **removing** `editor` when it is exactly **`true`** (grid-level and per column). That treats `true` as “no editor named” so cells inherit from column **`type`**, grid defaults, or the default text editor—**not** by forcing the text editor, which would break cases like **`{ type: 'numeric', editor: true }`**.
> 
> **`editor: false`**, **`null`**, string/constructor/component editors are unchanged. The same path covers initial settings and **`ngOnChanges`** updates.
> 
> Adds resolver and **`HotTableComponent`** tests plus an Angular changelog entry.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b1aafda88502ca239a9bb0e93d476b83327a2c88. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changelogs/13277.json
+++ b/.changelogs/13277.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Fixed the Angular wrapper throwing an error on cell select when a cell's `editor` setting was `true`.",
+  "type": "fixed",
+  "issueOrPR": 13277,
+  "breaking": false,
+  "framework": "angular"
+}
--- a/wrappers/angular-wrapper/projects/hot-table/src/lib/hot-table.component.spec.ts
+++ b/wrappers/angular-wrapper/projects/hot-table/src/lib/hot-table.component.spec.ts
@@ -2,6 +2,8 @@ import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { CUSTOM_ELEMENTS_SCHEMA, NgZone, SimpleChange, SimpleChanges } from '@angular/core';
 import Handsontable from 'handsontable';
 import { registerPlugin, CopyPaste } from 'handsontable/plugins';
+import { NumericEditor, TextEditor } from 'handsontable/editors';
+import { registerCellType, NumericCellType } from 'handsontable/cellTypes';
 import { HotTableModule } from './hot-table.module';
 import { HOT_DESTROYED_WARNING, HotTableComponent } from './hot-table.component';
 import { GridSettings } from './models/grid-settings';
@@ -11,6 +13,8 @@ import { HOT_GLOBAL_CONFIG, HotGlobalConfigService, NON_COMMERCIAL_LICENSE } fro
 import { DynamicComponentService } from './renderer/hot-dynamic-renderer-component.service';
 
 registerPlugin(CopyPaste);
+// The `boolean \`editor\` setting` suite needs a cell type that brings its own editor along.
+registerCellType(NumericCellType);
 
 describe('HotTableComponent', () => {
   let fixture: ComponentFixture<HotTableComponent>;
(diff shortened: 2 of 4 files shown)


qunabu:
Task linked: [DEV-2661 Core: normalize an `editor` setting of `true` instead of throwing](https://app.clickup.com/t/9015210959/DEV-2661)
Task linked: [DEV-2663 Vue 3 wrapper: an `editor` of `true` throws instead of using the default editor](https://app.clickup.com/t/9015210959/DEV-2663)
Task linked: [DEV-2664 Angular wrapper: an `editor` of `true` throws instead of using the default editor](https://app.clickup.com/t/9015210959/DEV-2664)

budnix:
Superseded by #13278, which fixes this in core instead of in the Angular wrapper.

Verified against that branch with this PR's wrapper change removed: #13278 normalizes `editor: true` at the meta layers, so it covers all four ways to set the option in the Angular wrapper. This PR's resolver-level fix covered only two of them — `columns` as a function, a `cells` callback, and the `cell` array all still threw. Wrapping the `cells` callback wrapper-side would have added a per-cell branch on the cell-meta read path, so core was the only site that could close this completely.

The five Angular component tests from this PR are ported to #13278 in 00699a3794, so the wrapper keeps its inheritance coverage alongside the Vue one.

PR #13278, opened 2026-08-27
DEV-2663: Normalize an editor setting of true instead of throwing

budnix:
### Context

The PR fixes an `editor` setting of `true` throwing instead of falling back to the default editor.

Core declares `editor?: string | ctor | boolean` (`handsontable/src/core/settings.ts`), so `editor: true` type-checks everywhere. At runtime it did not work: `getCellEditor()` returned the raw `true`, and `getEditorInstance()` (`handsontable/src/editors/registry.ts`) then threw `Only strings and functions can be passed as "editor" parameter` on the first edit. `metaSchema` documented only `false`.

DEV-2663 reported this on the Vue 3 wrapper, but the wrapper only passes the value through — vanilla and Angular throw identically. So the fix is in core, and Vue, Angular and vanilla are all covered without a wrapper-side change.

**The semantic:** `true` names no editor, so it now reads as "the setting was not passed". The `editor` key is dropped on the way into the meta layers rather than resolved to the text editor. That distinction matters — resolving it eagerly would silently drop the numeric editor of a `type: 'numeric'` column, which was the second half of the reported defect (`<HotColumn :type="'numeric'" :editor="true" />` threw *and* lost the numeric editor).

**Where it is applied:**

- `normalizeEditorSetting()` in `dataMap/metaManager/utils.ts` drops an `editor` of `true` from a settings object.
- Called from the meta layers (`globalMeta`, `tableMeta`, `columnMeta`, `cellMeta`) and from `DynamicCellMetaMod`. Verified normalized at write time: the constructor (grid-level and `columns`) and the `cells` callback.
- `getCellEditor()` also falls back to the default editor for a bare `true`. This catches a value written straight onto the cell meta — by a `beforeGetCellMeta` hook, or by `setCellMeta`, which does not route through the layer `updateMeta` calls.

`editor: false` (disable editing), a named editor, and a constructor are all untouched.

**Left alone deliberately:**

- `editor: null` still locks the cell in Vue while React inherits. `null` is outside the declared type, so this is a leniency choice rather than a correctness one, and DEV-2663 flags it as a separate decision. This PR does not change it.

### Relationship to the React wrapper (#13268, merged)

#13268 landed on `develop` after this branch was cut. The two were measured together — this branch merges `develop` with no conflicts, and every suite is green on the merged tree (see below).

**The React wrapper's `resolveEditorSetting()` is *not* made redundant by this PR, and must stay.** An earlier revision of this description claimed it was; that was wrong. Reverting the three React source files from #13268 while keeping this core fix in place fails **10** React tests. Eight of them are about `false`, which core cannot supply: `editor={false}` is a *prop-layer* problem — the old truthy check could not tell "prop absent" from "prop explicitly false", so the value never reached core at all. Core reads `false` correctly once it arrives; making it arrive is the wrapper's job.

Two of the ten are about `true` and still fail without the wrapper, for a reason core also cannot reach: `<HotColumn type="numeric" editor={true} />` hits the *component-editor* branch, because a bare `true` is truthy. Without `isComponentEditor()` excluding booleans, React builds an editor class from `true` and the cell resolves to `EDITOR_TYPE: "base"` instead of `"numeric"`. By the time core sees it, the value is a constructor, not a boolean, so there is nothing left to normalize.

**Exactly one clause is genuinely redundant** — the `hotEditor !== true` guard in `resolveEditorSetting()`:

```js
if (hotEditor && hotEditor !== true) {
```

Removing the `&& hotEditor !== true` half keeps all 94 React tests green on this branch, because `hotEditor={true}` then reaches core and the meta layers drop it. That path has real coverage (`should fall back to the default editor when the hotEditor prop is set to true`, and the `type="numeric" hotEditor={true}` column), so the green run is meaningful rather than an untested gap.

**Recommendation: keep the guard anyway.** `@handsontable/react-wrapper` declares `"handsontable": "^18.0.0"` as a peer dependency, so a user can pair a wrapper that has dropped the guard with a core that predates this fix and get the throw back. The clause costs one comparison and removes that failure mode. Its comment stays accurate either way.

### How has this been tested?

New tests, all verified to fail without the source change and pass with it. Control cases (`editor: false`, a named editor, a column with no `editor` prop) pass on both sides, so a dead harness cannot produce a green run.

- Unit — `normalizeEditorSetting()` itself, plus the four meta layers, `type` expansion, cross-layer inheritance, and the `cells` path:

  ```bash
  npm --prefix handsontable run test:unit -- --testPathPattern=metaManager
  ```

- E2E — `getCellEditor()` resolution and actually opening the editor (constructor, `columns`, `cell`, `cells`, `updateSettings`):

  ```bash
  npm --prefix handsontable run test:e2e -- --testPathPattern=getCellEditor
  npm --prefix handsontable run test:e2e -- --testPathPattern=settings/editor
  ```

- Vue 3 — the reported repro from DEV-2663, covering the ticket's full table:

  ```bash
  npm --prefix wrappers/vue3 run test
  ```

- Angular — five component-level tests confirming the wrapper inherits the fix with no wrapper-side change, mirroring the Vue coverage: grid-level `true`, column-level `true`, `type: 'numeric'` keeping the numeric editor, `editor: false` still disabling, and `true` arriving through `ngOnChanges`. Ported from the now-closed #13277.

  ```bash
  ```

Full runs on this branch: core unit 3978 passed; `getCellEditor` 26 specs / `settings/editor` 18 specs, 0 failures; Vue 39 passed; Angular 296 passed. `test:types`, `eslint` (0 errors) and `stylelint` all clean.

Re-run on this branch merged with current `develop` (which includes #13268), to confirm the two changes compose: core unit `metaManager` 274 passed, React **94 passed**, Angular **296 passed**, Vue **39 passed**. The merge itself is conflict-free.

### Types of changes

### Related issue(s):

1. DEV-2663
2. DEV-2661 — the core-side ticket for this exact change; covered by this PR.
3. DEV-2664 — the Angular half. Closed as covered by this PR; its Angular tests are ported here.
4. #13268 (DEV-2655) — the React half of #7561, merged. Composes with this PR; see the section above.
5. #13277 — the superseded Angular-wrapper-side fix, closed in favor of this one.

### Affected project(s):

- [x] `handsontable`
- [x] `@handsontable/vue3`

### Checklist:

The `editor` option's Typedoc in `metaSchema.ts` is updated in this PR — it previously documented only `false`.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2663

---

> [!NOTE]
> **Medium Risk**
> Touches cascading cell meta and editor resolution across many config paths; behavior change is intentional but could affect configs that relied on the old (broken) boolean.
> 
> **Overview**
> Fixes a bug where **`editor: true`** was stored on cell meta and caused **`getEditorInstance()`** to throw on the first edit. **`true`** is now treated as “no editor specified,” so cells keep the editor from **`type`** or a higher config level (e.g. numeric columns still use the numeric editor).
> 
> Core adds **`normalizeEditorSetting()`**, which strips **`editor: true`** before meta layer updates (global/table/column/cell, **`cells`**, and dynamic cell meta). **`updateSettings({ editor: true })`** no longer overwrites an existing grid editor; **`CellMeta#setMeta`** and **`getCellEditor()`** handle stray **`true`** values (e.g. hooks). **`editor: false`** and named/custom editors are unchanged. Docs and tests cover core, Vue 3, and Angular.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cde20feac732ce6dcc5b29ab218111a8c89d55c4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/handsontable/src/dataMap/metaManager/metaLayers/cellMeta.ts
+++ b/handsontable/src/dataMap/metaManager/metaLayers/cellMeta.ts
@@ -1,4 +1,4 @@
-import { extendByMetaType, assert } from '../utils';
+import { extendByMetaType, assert, normalizeEditorSetting } from '../utils';
 import LazyFactoryMap from '../lazyFactoryMap';
 import { extend, hasOwnProperty } from '../../../helpers/object';
 import { isDefined } from '../../../helpers/mixed';
@@ -98,9 +98,10 @@ export default class CellMeta {
    */
   updateMeta(physicalRow: number, physicalColumn: number, settings: Record<string, unknown>) {
     const meta = this.getMeta(physicalRow, physicalColumn);
+    const normalizedSettings = normalizeEditorSetting(settings);
 
-    extend(meta, settings);
-    extendByMetaType(meta, settings);
+    extend(meta, normalizedSettings);
+    extendByMetaType(meta, normalizedSettings);
   }
 
   /**
--- a/handsontable/src/dataMap/metaManager/metaLayers/cellMeta.ts
+++ b/handsontable/src/dataMap/metaManager/metaLayers/cellMeta.ts
@@ -231,6 +231,20 @@ export default class CellMeta {
   setMeta(physicalRow: number, physicalColumn: number, key: string, value: unknown) {
     const cellMeta = this.metas.obtain(physicalRow).obtain(physicalColumn);
 
+    // An `editor` of `true` names no editor, so it reads as "the setting was not passed". Dropping
+    // the own property lets the cell keep the editor its `type` expands to, or the one inherited
+    // from the column and grid layers. Storing the boolean instead would hand a bare `true` to
+    // `getEditorInstance()` and, by clearing the key from `_automaticallyAssignedMetaProps`, would
+    // also stop `extendByMetaType()` from supplying the type's editor. The bookkeeping entries are
+    // cleared too, so an earlier real write does not keep the cell pinned against meta eviction.
+    if (key === 'editor' && value === true) {
+      delete cellMeta[key];
+      (cellMeta._persistedMetaProps as Set<string> | undefined)?.delete(key);
+      (cellMeta._userDefinedMetaProps as Set<string> | undefined)?.delete(key);
+
+      return;
+    }
+
     (cellMeta._automaticallyAssignedMetaProps as Set<string> | undefined)?.delete(key);
     cellMeta[key] = value;
 
--- a/handsontable/src/dataMap/metaManager/metaSchema.ts
+++ b/handsontable/src/dataMap/metaManager/metaSchema.ts
@@ -2474,6 +2474,10 @@ export default (): Record<string, unknown> => {
      * | `Delete` / `Backspace`                  | Clear the contents of the selected cells                    |
      * | `Ctrl` + `Enter` / `Cmd` + `Enter`      | Fill selected cells with the value of the active cell       |
      *
+     * Setting the `editor` option to `true` names no editor, so Handsontable treats it as if the
+     * option was not set at all. The cell keeps the editor that its [`type`](#type) provides, or the
+     * editor inherited from a higher configuration level.
+     *
      * To set the [`editor`](#editor), [`renderer`](#renderer), and [`validator`](#validator)
      * options all at once, use the [`type`](#type) option.
      *
--- a/.changelogs/13278.json
+++ b/.changelogs/13278.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Fixed an `editor` option of `true` throwing an error instead of using the default editor.",
+  "type": "fixed",
+  "issueOrPR": 13278,
+  "breaking": false,
+  "framework": "none"
+}
--- a/handsontable/src/__tests__/core/getCellEditor.spec.js
+++ b/handsontable/src/__tests__/core/getCellEditor.spec.js
@@ -277,6 +277,46 @@ describe('Core.getCellEditor', () => {
       expect(getCellEditor(1, 1)).toBe(getCellType('text').editor);
     });
 
+    it('should return the `type` editor when passed through the `cell` option', async() => {
+      handsontable({
+        data: createSpreadsheetData(5, 5),
+        cell: [
+          { row: 1, col: 1, type: 'numeric', editor: true },
+        ],
+      });
+
+      // The `cell` option does not go through the layer `updateMeta` calls - it lands in
+      // `CellMeta#setMeta`, which used to store the raw `true` as an own property and so blocked
+      // the `type` expansion from supplying the numeric editor.
+      expect(getCellEditor(1, 1)).toBe(getCellType('numeric').editor);
+      expect(getCellMeta(1, 1).editor).not.toBe(true);
+    });
+
+    it('should keep the inherited editor when `setCellMeta` writes an `editor` of `true`', async() => {
+      handsontable({
+        data: createSpreadsheetData(5, 5),
+        editor: 'password',
+      });
+
+      setCellMeta(1, 1, 'editor', true);
+
+      expect(getCellEditor(1, 1)).toBe(getCellType('password').editor);
+      expect(getCellMeta(1, 1).editor).not.toBe(true);
+    });
+
+    it('should not drop a grid-level editor when `updateSettings` passes an `editor` of `true`', async() => {
+      handsontable({
+        data: createSpreadsheetData(5, 5),
+        editor: 'password',
+      });
+
+      await updateSettings({ editor: true });
+
+      // "Not passed" means the previous value stands, exactly as `updateSettings({})` would leave it.
+      expect(getCellEditor(1, 1)).toBe(getCellType('password').editor);
+      expect(getCellMeta(1, 1).editor).not.toBe(true);
+    });
+
     it('should return the `type` editor when passed through the `cells` option', async() => {
       handsontable({
         data: createSpreadsheetData(5, 5),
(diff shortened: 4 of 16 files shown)


qunabu:
Task linked: [DEV-2655 React wrapper: editor={false} does not disable cell editing](https://app.clickup.com/t/9015210959/DEV-2655)
Task linked: [DEV-2661 Core: normalize an `editor` setting of `true` instead of throwing](https://app.clickup.com/t/9015210959/DEV-2661)
Task linked: [DEV-2663 Vue 3 wrapper: an `editor` of `true` throws instead of using the default editor](https://app.clickup.com/t/9015210959/DEV-2663)
Task linked: [DEV-2664 Angular wrapper: an `editor` of `true` throws instead of using the default editor](https://app.clickup.com/t/9015210959/DEV-2664)

handsontable/src/dataMap/metaManager/metaLayers/cellMeta.ts
```diff
@@ -98,9 +98,10 @@ export default class CellMeta {
    */
   updateMeta(physicalRow: number, physicalColumn: number, settings: Record<string, unknown>) {
     const meta = this.getMeta(physicalRow, physicalColumn);
+    const normalizedSettings = normalizeEditorSetting(settings);
```
adrianspdev, line 101: The `cell` option and `setCellMeta` never reach `updateMeta` — `core.ts:3601` goes through `setCellMetaObject` → `setCellMeta` → `CellMeta#setMeta` (line 231 below), which writes the raw `true` as an own property and removes `editor` from `_automaticallyAssignedMetaProps`, so the `type` expansion afterwards sees an own key and `canBeOverwritten()` refuses to supply the editor. On this branch, `cell: [{ row: 0, col: 0, type: 'numeric', editor: true }]` gives a numeric renderer and validator with a text editor, which is the "silently drops the numeric editor" case the description says the fix avoids, one path over. Grid `editor: 'password'` plus `setCellMeta(0, 0, 'editor', true)` loses the inherited editor the same way, and `getCellMeta(0, 0).editor` still hands `true` to anyone calling `getEditorInstance()` on it. The current `cell` tests use `editor: true` without a `type`, so text is the correct answer there and the gap does not show.

Dropping the own property in `setMeta` instead of assigning — `if (key === 'editor' && value === true) { delete cellMeta[key]; }` — restores both the `type` expansion and the prototype inheritance, and the `=== true` branch in `getCellEditor()` can stay as the last resort for `beforeGetCellMeta` writes. Two cases would pin it: the `cell` option with `type: 'numeric'` resolving to the numeric editor, and `setCellMeta('editor', true)` under a grid-level `'password'` resolving to password.

handsontable/src/dataMap/metaManager/metaSchema.ts
```diff
@@ -2474,6 +2474,10 @@ export default (): Record<string, unknown> => {
      * | `Delete` / `Backspace`                  | Clear the contents of the selected cells                    |
      * | `Ctrl` + `Enter` / `Cmd` + `Enter`      | Fill selected cells with the value of the active cell       |
      *
+     * Setting the `editor` option to `true` names no editor, so Handsontable treats it as if the
```
adrianspdev, line 2477: Until the `setMeta` path is covered, "keeps the editor that its `type` provides, or the editor inherited from a higher configuration level" is false for the `cell` option and `setCellMeta` — worth landing the two together so the reference does not get ahead of the code.

adrianspdev (changes requested):
One thing to check before merge — the normalization has a gap on the cell-level write path, and the new Typedoc promises behavior that does not hold there.

handsontable/src/dataMap/metaManager/metaLayers/cellMeta.ts
budnix: Nice catch — fixed in cde20feac7. `setMeta` now drops the own property instead of storing the boolean, so the `type` expansion and the inherited editor both survive. Also added your two cases plus a third: `updateSettings({ editor: true })` was clobbering a grid-level editor the same way.

handsontable/src/dataMap/metaManager/metaSchema.ts
budnix: Good point — landed together in cde20feac7, so the doc stays true instead of getting ahead of the code. I checked every write path and the sentence now holds on all of them.

handsontable/src/dataMap/metaManager/metaLayers/cellMeta.ts
```diff
@@ -230,6 +231,20 @@ export default class CellMeta {
...
+    // `getEditorInstance()` and, by clearing the key from `_automaticallyAssignedMetaProps`, would
+    // also stop `extendByMetaType()` from supplying the type's editor. The bookkeeping entries are
+    // cleared too, so an earlier real write does not keep the cell pinned against meta eviction.
+    if (key === 'editor' && value === true) {
+      delete cellMeta[key];
+      (cellMeta._persistedMetaProps as Set<string> | undefined)?.delete(key);
+      (cellMeta._userDefinedMetaProps as Set<string> | undefined)?.delete(key);
+
+      return;
+    }
```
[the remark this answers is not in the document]
budnix: No fix needed here — this is pre-existing behavior, not something this change introduced.

I measured four cases with the change in, then re-ran them against the pre-change code:

| Case | Before | After |
| --- | --- | --- |
| `cells()` sets `type`, then `setCellMeta("editor", true)` | fails | fails (unchanged) |
| `cell` option sets `type`, then `setCellMeta("editor", true)` | fails | fails (unchanged) |
| column-level `type`, then `setCellMeta("editor", true)` | fails | **passes** |
| runtime `setCellMeta("type", "numeric")` | fails on its own baseline | unchanged |

Before this change, `setMeta` stored the raw `true` and `getCellEditor` resolved it to the text editor, so the type editor was already lost on those paths. The change makes the column-level case work and leaves the other two exactly as they were.

What is true is narrower than described: when a `type` lands on the **cell** layer (via `cells()` or the `cell` option) and `setCellMeta("editor", true)` runs afterwards, the type editor is not restored. Fixing that means touching the type expansion and `metaSyncMemo` path rather than this guard, so it is tracked as a follow-up.

The last row is unrelated to `editor` — runtime `setCellMeta("type", ...)` does not expand at all, with or without this change.

adrianspdev (approved):

Issue #5942, opened 2019-04-10
allowInsertRow does more than what the documentation says


jeremy-smith-maco:
### Description

Documentation just says that if set to `false` then there won't be options in the context menu to insert a row but in fact it prevents multiple rows being pasted in a table.

Seems like the documentation needs to be updated to better reflect what it does. Not sure if this also the same issue for `allowInsertColumn`.

### Steps to reproduce

1. Set `minSpareRows` to something above 0 and `allowInsertRow` to false
2. Copy 2+ rows
3. Try and paste in the spare row

### Demo

http://jsfiddle.net/mrz3jb0k/

### Your environment
* Handsontable version: 6.0.1
* Browser Name and version: Chrome Version 69.0.3497.100
* Operating System: Windows 10

AMBudnik:
Thank you for sharing the issue report. 

![GIF](https://user-images.githubusercontent.com/10757813/55858270-8c02c500-5b6f-11e9-9971-2e0669fd42ad.gif)

When I paste the first two rows into the spare row I still get one more row extra, only part of the data is pasted and table scrolls a bit right.

AMBudnik:
**Works the same** in 8.0.0-beta2-rev14 http://jsfiddle.net/AMBudnik/6m38z07r/

pnowak:
This issue is still replicable using v8.0.0. Here's an updated demo - https://jsfiddle.net/cdy15rkp/

AMBudnik:
v8.2.0 works the same as described in the issue above http://jsfiddle.net/f9guvj6w/

budnix:
Thank you for the report, and apologies that it stayed open for so long.

You are right on both counts. `allowInsertRow` does more than add or hide the context menu items, and `allowInsertColumn` has the same gap you suspected.

Each of the two options controls two separate things:

1. The insert items in the menus.
2. Whether the grid may add rows or columns on its own.

The second part was never documented. With `allowInsertRow: false`:

- A paste taller than the rows left below the selection stops at the last row. The extra values are dropped, and no error is reported. That is what you hit with `minSpareRows`: the spare row still appears, because `minSpareRows` ignores `allowInsertRow`, but only the first pasted row lands in it.
- An autofill that reaches past the last row stops at the last row.
- `setDataAtCell()` and `setDataAtRowProp()` no longer create the missing rows when you write below the last row.

`allowInsertColumn: false` does the same for columns.

[...]

PR #13284, opened 2026-08-28
DEV-2673: Document that allowInsertRow and allowInsertColumn block grid growth

budnix:
### Context

The PR documents what `allowInsertRow` and `allowInsertColumn` actually do. It changes no behavior.

Both options were documented as pure context-menu toggles. In fact each one gates **two** separate things: the menu items, **and** whether the grid may add rows or columns on its own. GitHub issue [#5942](https://github.com/handsontable/handsontable/issues/5942) reported this in 2019 and it has been open since.

The reporter hit it with `minSpareRows: 1` and `allowInsertRow: false`: pasting 2 rows into the spare row writes only the first row and silently drops the second. Their guess that `allowInsertColumn` has the same gap is also correct. The paste behavior itself is deliberate and pinned by an existing test (`handsontable/src/plugins/copyPaste/__tests__/paste.spec.js:83`), so this is a documentation gap rather than a code bug.

A sweep found only **four** behavioral reads in the whole codebase, all in `handsontable/src/core.ts`:

| Site | Option | Effect |
| --- | --- | --- |
| `core.ts:1419` | `allowInsertRow` | `populateFromArray` stops writing past the last row |
| `core.ts:1456` | `allowInsertColumn` | `populateFromArray` stops writing past the last column |
| `core.ts:2027` | `allowInsertRow` | `applyChanges` no longer auto-creates rows below the last row |
| `core.ts:2046` | `allowInsertColumn` | `applyChanges` no longer auto-creates columns past the last column |

Everything routed through `populateFromArray` inherits that clamp: copy-paste, autofill, `mergeCells`, `moveCells`, undo/redo of moves and merges, and the `shift_down` / `shift_right` paste modes.

Several behaviors here are sharper than "the grid stops growing", and each one is documented explicitly:

- **`setDataAtCell()` and `setDataAtRowProp()` throw.** With `allowInsertRow: false`, writing below the last row raises `TypeError: Cannot set properties of null`. It is not a quiet no-op, so the JSDoc tells you to guard the call. See the reviewer note at the end.
- **A blocked column write still reaches the source data.** With `allowInsertColumn: false`, `setDataAtCell(0, 5, 'Z')` creates no column and displays nothing, but `getSourceDataAtRow(0)` becomes `["a","b","c",null,null,"Z"]`. The value is hidden, not dropped — worth knowing before you persist that source data.
- **The `shift_down` paste mode loses rows.** Rows pushed past the last row have nowhere to go, so they are destroyed rather than truncated. Pinned by `paste.spec.js:104`.
- **`alter()` ignores all four `allow*` options.** `insert_row_above` still inserts with `allowInsertRow: false`, and `remove_row` still removes with `allowRemoveRow: false`. So do `minSpareRows` / `minSpareCols`, undo/redo, and moving the selection past the edge. The options are not a "user cannot add rows" switch, and the JSDoc now points at `maxRows` / `maxCols` instead.
- **The fill handle appends rows regardless of `allowInsertRow`.** It goes through `alter()`, so only `fillHandle.autoInsertRow` governs it. That setting also only takes effect once you set `fillHandle` explicitly: the schema default is `{ autoInsertRow: false }`, so an unset `fillHandle` does not append. The `fillHandle` JSDoc claimed `autoInsertRow` is `true` by default without that qualification, so this PR corrects it too.
- **`minCols` and `minSpareCols` are skipped when `columns` is set** (`core.ts:1290`, `core.ts:1314`), and `minSpareCols` additionally needs array-of-arrays data. The row equivalents carry no such precondition, so the symmetry was misleading.
- **`allowInsertColumn` and `allowRemoveColumn` also hide their items from the column menu**, not just the context menu.

`allowRemoveRow` and `allowRemoveColumn` have no behavioral reads at all, so their menu-only description was accurate apart from the column menu. Both now say explicitly that they hide menu items and nothing more.

Changes:

- `handsontable/src/dataMap/metaManager/metaSchema.ts` — rewrite the JSDoc for all four options, and correct the `fillHandle` table's `autoInsertRow` default.
- `docs/content/guides/columns/column-adding/column-adding.md` — add the column menu, and note that `allowInsertColumn` does more than hide menu items, with the array-of-arrays precondition.
- `docs/content/guides/rows/row-prepopulating/row-prepopulating.md` — warn against combining `minSpareRows` with `allowInsertRow: false`, which is the exact trap in the issue, and point at the wider behavior.
- `handsontable/src/plugins/contextMenu/itemsFactory.ts` — fix a stale comment claiming the option deletes the item from settings. It does not; the item is hidden at render time by `hidden()`.

### How has this been tested?

Behavior is unchanged, so this adds no tests. The two source files carry comment-only edits, declared with a `Refactor-only:` commit trailer.

Verified the rendered output and that nothing regressed:

```shell
# lint: 0 errors (602 pre-existing warnings in unrelated test files)

# types

# docs lint

# API reference generates cleanly and renders the new text
npm run build --prefix handsontable   # docs:api reads handsontable/tmp/, not src/
```

Every statement in the new documentation was checked in a real browser against a local `develop` build, each case alongside a control run with the option left at its default:

| Case | Setup | Result |
| --- | --- | --- |
| 1 | `minSpareRows: 1`, `allowInsertRow: false`, paste 2 rows at the spare row | 1 row written, second dropped silently |
| 2 | control | both rows written |
| 3 | `allowInsertRow: false`, `setDataAtCell(5, 0, 'X')` on a 3-row grid | **throws `TypeError`** |
| 4 | control | no throw, grid grows to 6 rows |
| 5 | `allowInsertColumn: false`, `setDataAtCell(0, 5, 'Z')` | no column created, nothing displayed, but `getSourceDataAtRow(0)` is `["a","b","c",null,null,"Z"]` |
| 6 | `allowInsertRow: false`, `pasteMode: 'shift_down'`, paste 1 row at (0,0) | `[['p1','p2','p3'],['a','b','c']]` — the row `['d','e','f']` is **lost** |
| 7 | control, same paste | all 3 rows kept |
| 8 | `allowInsertColumn: false`, paste 3 columns from the last column | 2 of 3 columns dropped silently |
| 9 | `allowInsertRow: false`, autofill fill 4 rows past the bottom | clamped at the last row |
| 10 | `fillHandle: true` + `allowInsertRow: false`, drag the fill handle below the last row | rows 5 → 14, **`allowInsertRow` ignored** |
| 11 | no `fillHandle` set, same drag | no rows appended (`autoInsertRow` resolves to `false`) |
| 12 | `allowInsertRow: false`, `alter('insert_row_above')` | row still added, option ignored |
| 13 | `allowRemoveRow: false`, `alter('remove_row')` | row still removed, option ignored |

Also confirmed in the browser that `allowInsertColumn: false` and `allowRemoveColumn: false` remove **Insert column left**, **Insert column right**, and **Remove column** from the column menu as well as the context menu.

### Types of changes

Documentation only. No behavior change.

### Related issue(s):

1. DEV-2673
2. #5942

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

### Note for the reviewer

While verifying this, I found a genuine crash that is **out of scope here** and needs its own ticket:

`setDataAtCell(rowBelowLast, col, value)` with `allowInsertRow: false` throws `TypeError: Cannot set properties of null (setting '0')`. With object data, `setDataAtRowProp()` throws the same way.

The row-creating block at `core.ts:2027` is wrapped in `if (tableMeta.allowInsertRow)` with no `else`, so `skipThisChange` stays `false` and `datamap.set()` runs against a row that does not exist. The column branch at `core.ts:2046` has the same shape but degrades quietly, writing into the source data with no throw, so the two axes are inconsistent. Reproduces on 18.0.0 and on current develop.

The JSDoc added here documents the throw as it currently behaves. If that crash is fixed to skip the change instead, this wording needs one edit.

[skip changelog]

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2673

---

> [!NOTE]
> **Low Risk**
> Comment and documentation edits only; no runtime behavior changes.
> 
> **Overview**
> **Documentation-only** update clarifying that `allowInsertRow` and `allowInsertColumn` control both menu visibility and whether the grid auto-adds rows/columns (paste, autofill, `setDataAtCell` / `setDataAtRowProp`), with explicit lists of what they do and do not block. `allowRemoveRow` and `allowRemoveColumn` are documented as menu-only toggles (context + column menu for columns), with removal blocking pointed to hooks.
> 
> The **API schema JSDoc** in `metaSchema.ts` is expanded for all four `allow*` options; **`fillHandle`** docs now state that unset `fillHandle` does not append rows via the handle, correct the documented default to `{ autoInsertRow: false }`, and note that fill-handle row insertion ignores `allowInsertRow`.
> 
> **Guides:** the column-adding guide mentions the column menu and insert-column side effects; the row-prepopulating guide adds a warning about combining `minSpareRows` with `allowInsertRow: false`. A **stale comment** in `itemsFactory.ts` is corrected to match render-time `hidden()` behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 49401b0212eb4ac3598a5440a9c4744b26bef224. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2673 Document the full behavior of allowInsertRow and allowInsertColumn (#5942)](https://app.clickup.com/t/9015210959/DEV-2673)

handsontable/src/dataMap/metaManager/metaSchema.ts
```diff
@@ -286,10 +286,34 @@ export default (): Record<string, unknown> => {
...
-     * If set to `true`, the `allowInsertColumn` option adds the following menu items to the [context menu](@/guides/accessories-and-menus/context-menu/context-menu.md):
+     * The `allowInsertColumn` option controls two things: the insert items in the menus, and whether
+     * the grid may add columns on its own.
+     *
+     * If set to `true`, the option adds the following menu items to the [context menu](@/guides/accessories-and-menus/context-menu/context-menu.md)
+     * and to the [column menu](@/guides/accessories-and-menus/column-menu/column-menu.md):
      * - **Insert column left**
      * - **Insert column right**
      *
+     * If set to `false`, the option also stops the grid from adding columns on its own. This second effect applies
```
demtario: **medium** — The precondition in this lead-in ("applies only when your `data` is an array of arrays and you do not set the `columns` option") governs all three bullets below it, but it only holds for the third.

The paste/autofill clamp at [`core.ts:1527`](https://github.com/handsontable/handsontable/blob/bc2feef9f4807e7781872f1bc1da1f210bb1c821/handsontable/src/core.ts#L1527) reads `!tableMeta.allowInsertColumn` alone, with no data-type or `columns` guard. Only the `setDataAtCell()` path at [`core.ts:2116-2117`](https://github.com/handsontable/handsontable/blob/bc2feef9f4807e7781872f1bc1da1f210bb1c821/handsontable/src/core.ts#L2116) carries the precondition.

So with object data or a `columns` setting, `allowInsertColumn: false` still changes what a paste writes into the source data. That contradicts "in every other configuration the grid never adds columns anyway", which tells those readers the whole block is irrelevant to them.

Move the precondition into the `setDataAtCell()`/`setDataAtRowProp()` bullet and leave the paste and autofill bullets unconditional.

handsontable/src/dataMap/metaManager/metaSchema.ts
```diff
@@ -286,10 +286,34 @@ export default (): Record<string, unknown> => {
...
+     * only when your [`data`](#data) is an array of arrays and you do not set the [`columns`](#columns) option –
+     * in every other configuration the grid never adds columns anyway:
+     * - A [paste](@/guides/cell-features/clipboard/clipboard.md) that is wider than the columns left to the right of the
```
demtario, line 298: **low** — The row equivalent documents the `shift_down` data loss in bold, but this column bullet has no `shift_right` counterpart, though the behavior is identical and pinned by a test.

[`paste.spec.js:145-160`](https://github.com/handsontable/handsontable/blob/bc2feef9f4807e7781872f1bc1da1f210bb1c821/handsontable/src/plugins/copyPaste/__tests__/paste.spec.js#L145): with `pasteMode: 'shift_right'` and `allowInsertColumn: false`, `getData()[0].length` stays 5 and the pushed-right values are gone.

Worth adding, in a PR whose thesis is row/column symmetry.

handsontable/src/dataMap/metaManager/metaSchema.ts
```diff
@@ -286,10 +286,34 @@ export default (): Record<string, unknown> => {
...
+     * - [`setDataAtCell()`](@/api/core.md#setdataatcell) and [`setDataAtRowProp()`](@/api/core.md#setdataatrowprop) no
+     *   longer create the missing columns when you write past the last column. The write still reaches the source data,
+     *   so [`getSourceData()`](@/api/core.md#getsourcedata) returns the value while the grid never displays it.
+     *
+     * The option does not stop these ways of adding columns:
+     * - The [`alter()`](@/api/core.md#alter) method, including its `insert_col_start` and `insert_col_end` actions.
+     * - The [`minCols`](#minCols) and [`minSpareCols`](#minSpareCols) options. Both are themselves skipped when the
+     *   [`columns`](#columns) option is set, and `minSpareCols` also requires [`data`](#data) to be an array of arrays.
+     * - Undo and redo.
+     * - Moving the selection past the last column, when [`minSpareCols`](#minSpareCols) is above `0`.
```
demtario: **medium** — "Moving the selection past the last column" is too broad: only Enter creates records, and on default settings this column case cannot fire at all.

[`_base.ts:95`](https://github.com/handsontable/handsontable/blob/bc2feef9f4807e7781872f1bc1da1f210bb1c821/handsontable/src/selection/transformation/_base.ts#L95) defaults `createMissingRecords = false`, and the column branch at [`_base.ts:192`](https://github.com/handsontable/handsontable/blob/bc2feef9f4807e7781872f1bc1da1f210bb1c821/handsontable/src/selection/transformation/_base.ts#L192) requires it to be `true`. The only caller passing `true` is [`editorManager.ts:326`](https://github.com/handsontable/handsontable/blob/bc2feef9f4807e7781872f1bc1da1f210bb1c821/handsontable/src/editorManager.ts#L326) (`moveSelectionAfterEnter`). Arrow keys and Tab all pass nothing.

And since the default `enterMoves` is `{ col: 0, row: 1 }` (line 2783), Enter never moves the column, so this bullet is unreachable without a non-default `enterMoves`.

Suggest: "Pressing Enter at the last column, when `minSpareCols` is above `0` and `enterMoves` is set to move the column."

The row bullet at line 360 has the same Enter-only scope and should say "Pressing Enter at the last row".

demtario (commented):
9 inline notes below.

- `handsontable/src/dataMap/metaManager/metaSchema.ts:436` — the PR adds "and to the [column menu]" to the `allowRemoveColumn` description (line 417), but the block's `@example` comment still reads `// hide the 'Remove column' menu item from the context menu`. The equivalent correction was made for `allowInsertColumn`. (`allowRemoveRow`'s example at line 463 is correct as-is — that item really is context-menu-only.)

handsontable/src/dataMap/metaManager/metaSchema.ts
```diff
@@ -300,17 +324,43 @@ export default (): Record<string, unknown> => {
...
+     * If set to `true`, the option adds the following menu items to the [context menu](@/guides/accessories-and-menus/context-menu/context-menu.md):
      * - **Insert row above**
      * - **Insert row below**
      *
+     * If set to `false`, the option also stops the grid from adding rows on its own:
+     * - A [paste](@/guides/cell-features/clipboard/clipboard.md) that is taller than the rows left below the selection
+     *   stops at the last row. Handsontable drops the extra values, and reports no error. In the
+     *   `shift_down` [`pasteMode`](@/api/copyPaste.md) the rows pushed past the last row are **lost**, because no row is
+     *   created to receive them.
+     * - An [autofill](@/guides/cell-features/autofill-values/autofill-values.md) whose fill reaches past the last row
```
demtario, line 351: **medium** — This bullet needs a `fillHandle.autoInsertRow` qualifier. Without one it contradicts the bullet at line 357 for the common `fillHandle: true` case.

With `fillHandle: true` and `allowInsertRow: false`, dragging the fill handle below the last row calls [`autofill.ts:500`](https://github.com/handsontable/handsontable/blob/bc2feef9f4807e7781872f1bc1da1f210bb1c821/handsontable/src/plugins/autofill/autofill.ts#L500) `hot.alter(...)`, which `allowInsertRow` never gates. `countRows()` grows first, so the clamp at [`core.ts:1490`](https://github.com/handsontable/handsontable/blob/bc2feef9f4807e7781872f1bc1da1f210bb1c821/handsontable/src/core.ts#L1490) never bites and the fill does extend past the original last row.

Both bullets are literally true of different mechanisms, but a reader with `fillHandle: true` reads this one and concludes autofill will not grow the grid, which is wrong.

Suggest: "stops at the last row, unless `fillHandle.autoInsertRow` has already appended rows (see below)."

handsontable/src/dataMap/metaManager/metaSchema.ts
```diff
@@ -361,9 +412,14 @@ export default (): Record<string, unknown> => {
     allowInvalid: true,
 
     /**
-     * If set to `true`, the `allowRemoveColumn` option adds the following menu items to the [context menu](@/guides/accessories-and-menus/context-menu/context-menu.md):
+     * If set to `true`, the `allowRemoveColumn` option adds the following menu item to the [context menu](@/guides/accessories-and-menus/context-menu/context-menu.md)
+     * and to the [column menu](@/guides/accessories-and-menus/column-menu/column-menu.md):
      * - **Remove column**
      *
+     * The option hides that menu item only. It does not stop the [`alter()`](@/api/core.md#alter) method's
+     * `remove_col` action, and it does not stop undo or redo. To cap the number of columns, use
```
demtario: **medium** — Neither option named here prevents a removal.

`maxCols`/`maxRows` are creation-path only. In `dataMap.ts` they appear solely in `createRow`/`createCol` and the length clamps; neither [`removeRow`](https://github.com/handsontable/handsontable/blob/bc2feef9f4807e7781872f1bc1da1f210bb1c821/handsontable/src/dataMap/dataMap.ts#L587) nor [`removeCol`](https://github.com/handsontable/handsontable/blob/bc2feef9f4807e7781872f1bc1da1f210bb1c821/handsontable/src/dataMap/dataMap.ts#L640) reads them. `minCols`/`minRows` do not appear in that file at all: they act afterwards in [`core.ts:1263-1273`](https://github.com/handsontable/handsontable/blob/bc2feef9f4807e7781872f1bc1da1f210bb1c821/handsontable/src/core.ts#L1263), where `datamap.createRow` appends **new empty** rows to restore the count. The data is already gone.

Suggest dropping the `max*` half. If `min*` stays, say what it actually does: it appends an empty column/row to restore the count, it does not prevent the removal or preserve the data. Returning `false` from `beforeRemoveCol`/`beforeRemoveRow` is the only real lever.

Same wording at line 447 for `allowRemoveRow`.

For contrast, the equivalent sentences in the `allowInsert*` blocks (lines 315 and 362) are correct — `maxRows`/`maxCols` genuinely do cap growth, enforced at `core.ts:1490` and in the `createCol` clamp.

handsontable/src/dataMap/metaManager/metaSchema.ts
```diff
@@ -2780,6 +2840,12 @@ export default (): Record<string, unknown> => {
      * | `autoInsertRow` | `true` (default) \| `false`    | `true`: When you reach the grid's bottom, add new rows<br>`false`: When you reach the grid's bottom, stop |
      * | `direction`     | `'vertical'` \| `'horizontal'` | `'vertical'`: Enable vertical autofill<br>`'horizontal'`: Enable horizontal autofill                      |
      *
+     * The `autoInsertRow` default above applies once you set `fillHandle` yourself, to any of the values in the first
```
demtario, line 2853: **low** — This paragraph's premise is that an unset `fillHandle` does not append rows, which is true only because the schema value is `{ autoInsertRow: false }` (line 2888). But the `@default true` tag at line 2857 is untouched, and the generated API table prints "Default: true" directly beside this text.

A reader trusting the tag concludes the opposite of what the paragraph says.

The PR description lists correcting this default as one of the changes, so suggest finishing it here: `@default { autoInsertRow: false }`.

docs/content/guides/columns/column-adding/column-adding.md
```diff
@@ -129,7 +129,9 @@ Setting [`contextMenu`](@/api/options.md#contextmenu) to `true` shows the full d
...
+The [`allowInsertColumn`](@/api/options.md#allowinsertcolumn) and [`allowRemoveColumn`](@/api/options.md#allowremovecolumn) options control whether these menu items are available, in both the context menu and the [column menu](@/guides/accessories-and-menus/column-menu/column-menu.md). Both default to `true`. Set [`allowInsertColumn`](@/api/options.md#allowinsertcolumn) to `false` to hide the insert items, or [`allowRemoveColumn`](@/api/options.md#allowremovecolumn) to `false` to hide the remove item.
+
+[`allowInsertColumn`](@/api/options.md#allowinsertcolumn) does more than hide menu items. Set to `false`, it also stops the grid from adding columns on its own: a paste wider than the remaining columns stops at the last column and Handsontable drops the extra values, and [`setDataAtCell()`](@/api/core.md#setdataatcell) no longer creates the missing columns. This second effect applies only when your [`data`](@/api/options.md#data) is an array of arrays and you do not set the [`columns`](@/api/options.md#columns) option - in every other configuration the grid never adds columns anyway. See the [`allowInsertColumn`](@/api/options.md#allowinsertcolumn) reference for the full list.
```
demtario: **medium** — Same scope error as `metaSchema.ts:297`. "This second effect" refers back to "stops the grid from adding columns on its own", which covers the paste clause in the same sentence, but the precondition only applies to `setDataAtCell()`.

The paste clamp at [`core.ts:1527`](https://github.com/handsontable/handsontable/blob/bc2feef9f4807e7781872f1bc1da1f210bb1c821/handsontable/src/core.ts#L1527) has no data-type or `columns` guard; only [`core.ts:2116-2117`](https://github.com/handsontable/handsontable/blob/bc2feef9f4807e7781872f1bc1da1f210bb1c821/handsontable/src/core.ts#L2116) does.

Restrict the precondition to the `setDataAtCell()` clause.

docs/content/guides/rows/row-prepopulating/row-prepopulating.md
```diff
@@ -27,6 +27,14 @@ Pre-populate new rows with default values when users add rows to the grid.
 
 To keep one empty row at the bottom of the grid, set [`minSpareRows`](@/api/options.md#minsparerows) to `1`.
 
+::: tip
```
demtario: **low** — This block documents silent data loss ("drops the rest without an error"), but uses `::: tip`. The guides already have a `::: warning` container for exactly this kind of content (8 uses across `docs/content/guides`).

Suggest `::: tip` → `::: warning`; the closing `:::` is unchanged.

docs/content/guides/rows/row-prepopulating/row-prepopulating.md
```diff
@@ -27,6 +27,14 @@ Pre-populate new rows with default values when users add rows to the grid.
...
+
+Do not combine [`minSpareRows`](@/api/options.md#minsparerows) with [`allowInsertRow: false`](@/api/options.md#allowinsertrow). The spare rows still appear, because [`minSpareRows`](@/api/options.md#minsparerows) ignores [`allowInsertRow`](@/api/options.md#allowinsertrow). But a paste into a spare row stops at the last row, so Handsontable writes the first pasted row and drops the rest without an error.
```
demtario: **low** — "Do not combine" is broader than the failure this paragraph documents, and the failure itself is narrower than stated.

The combination works for the option's primary use case: editing the last spare row targets `countRows - 1`, which is in range, so the row-creation block at [`core.ts:2098`](https://github.com/handsontable/handsontable/blob/bc2feef9f4807e7781872f1bc1da1f210bb1c821/handsontable/src/core.ts#L2098) is never needed and nothing throws, and [`core.ts:1274-1287`](https://github.com/handsontable/handsontable/blob/bc2feef9f4807e7781872f1bc1da1f210bb1c821/handsontable/src/core.ts#L1274) then adds the next spare row.

Only a paste extending past the last row truncates, via the clamp at [`core.ts:1490`](https://github.com/handsontable/handsontable/blob/bc2feef9f4807e7781872f1bc1da1f210bb1c821/handsontable/src/core.ts#L1490). And "writes the first pasted row and drops the rest" holds only when the paste starts at the last row — with `minSpareRows: 3`, a paste starting at the first spare row writes three rows.

Suggest reframing as a caveat rather than a prohibition: the spare rows still appear and typing into them works, but a paste that extends past the last row is truncated, and Handsontable drops the rows that do not fit without an error.

qunabu (dismissed):

handsontable/src/dataMap/metaManager/metaSchema.ts
budnix: Good spot — the precondition only ever applied to `setDataAtCell()`. Moved it onto that bullet in `49401b0212`, so the paste and autofill bullets now read unconditional.

handsontable/src/dataMap/metaManager/metaSchema.ts
budnix: Agreed — added the `shift_right` note in `49401b0212` so both axes read the same.

handsontable/src/dataMap/metaManager/metaSchema.ts
budnix: You're right, arrow keys and Tab never create records. Both bullets now say Enter in `49401b0212`, and the column one notes the default `enterMoves` never moves the column.

handsontable/src/dataMap/metaManager/metaSchema.ts
budnix: Nice catch — I hit that same growth while testing (5 rows to 14 on a real drag). Qualified the bullet in `49401b0212`.

handsontable/src/dataMap/metaManager/metaSchema.ts
budnix: Fair point, that was misleading. Both blocks now point at `beforeRemoveCol`/`beforeRemoveRow` in `49401b0212` and say `min*` only appends an empty row afterwards.

handsontable/src/dataMap/metaManager/metaSchema.ts
budnix: Yep, my mistake — finished that off in `49401b0212`, the tag now reads `{ autoInsertRow: false }`.

docs/content/guides/columns/column-adding/column-adding.md
budnix: Good point — same fix as the JSDoc one, the precondition now sits on the `setDataAtCell()` clause in `49401b0212`.

docs/content/guides/rows/row-prepopulating/row-prepopulating.md
budnix: Agreed — switched to `::: warning` in `49401b0212`.

docs/content/guides/rows/row-prepopulating/row-prepopulating.md
budnix: You're right, typing into the spare row works fine. Reframed it as a caveat in `49401b0212` — only a paste past the last row is cut short now.

budnix:
@demtario on the unanchored one — good catch, the `allowRemoveColumn` example now says "from the context menu and the column menu" in `49401b0212`. Left `allowRemoveRow`'s example alone, since that item really is context-menu-only.

demtario (approved):

Issue #5933, opened 2019-04-02
beforeInit hook never fires


jeremy-smith-maco:
### Description

`beforeInit` hook never fires. `afterInit` hook fires fine.

Edit: It works with a global hook (example: http://jsfiddle.net/g23ufym7/). If it doesn't work as a regular table option then maybe update the documentation to specify that you can't use it as a setting for the constructor.

### Steps to reproduce

1. Add a `beforeInit` hook
2. It won't fire

### Demo

http://jsfiddle.net/t3pL46q5/

### Your environment
* Handsontable version: 6.0.1
* Browser Name and version: Chrome Version 69.0.3497.100
* Operating System: Windows 10

AMBudnik:
Hey @jeremy-smith-maco 

I can see that we had a similar report here https://github.com/handsontable/handsontable/issues/3819 and here's another tip from our developer https://github.com/handsontable/handsontable/issues/2636

> [...] beforeInit (fired before table created) or construct (fired on object initialization - new Handsontable) event will be usefull for you? But this events are fired only once during initialization.

budnix:
After playing with hooks and `beforeInit` hook I came to the conclusion that this is a bug.  

The `beforeInit` hook is fired before any previously added listener to the Handsontable (local hooks added using settings object). So it's impossible to listen to that hook.

Fired at https://github.com/handsontable/handsontable/blob/master/src/core.js#L854 but listeners just added here https://github.com/handsontable/handsontable/blob/master/src/core.js#L860.

pnowak:
This issue is still replicable using v8.0.0. Here's an updated demo - https://jsfiddle.net/0rL31km7/

budnix:
Thanks for the report, and apologies that it sat for so long. This is a real bug, and it is now fixed in #13285.

**What was happening**

Handsontable registers the callbacks you pass in the settings object from inside `updateSettings()`. On startup, `init()` calls `updateSettings()` *after* it has already fired `beforeInit`. So a `beforeInit` callback written in the settings object was always registered too late to hear the hook. `afterInit` works because it fires at the end of `init()`, well after that point.

That is also why the global form works, as you found in your edit: `Handsontable.hooks.add('beforeInit', ...)` attaches before the instance is built.

**What changed**

`beforeInit` is now registered up front, right before it fires, so this works:

```js
new Handsontable(container, {
  beforeInit() {
    // now runs
  },
});
```

[...]

PR #13285, opened 2026-08-28
DEV-2672: Fire the beforeInit hook declared in the settings object (#5933)

budnix:
### Context

The PR fixes a `beforeInit` callback passed in the settings object never running. `afterInit` works, which is why the gap went unnoticed. GitHub issue [#5933](https://github.com/handsontable/handsontable/issues/5933) has been open since 2019 and still reproduces on `develop`.

`init()` does the following, in order:

1. Fires `beforeInit`.
2. Two lines later, calls `updateSettings(mergedUserSettings, true)`.
3. That `updateSettings` call is the only place where callbacks declared in the settings object get registered as hooks.

So the listener is always registered after the hook has already run. `afterInit` works because it fires at the end of `init()`.

Two supporting details that make this hard to spot:

- `settingsWithoutHooks` deliberately filters hook keys out of the `MetaManager` seeding, so hooks reach the instance only through the `updateSettings()` path.
- `Hooks.run()` runs the global bucket before the instance-local bucket, which is why the reporter's `Handsontable.hooks.add('beforeInit', ...)` workaround works.

**The fix:** register the `beforeInit` callback immediately before firing it. The existing registration logic moved into a shared `registerSettingsHook()` helper, now used by both `init()` and `updateSettings()`.

**Only `beforeInit` is pulled forward, deliberately.** Plugins register their feature hooks inside `enablePlugin()`, which runs *during* the `beforeInit` dispatch. Registering the whole settings object early would move settings-declared callbacks ahead of every plugin callback, for every hook. Every other hook fires after `updateSettings()` and is already correct. There is a test pinning that plugins are still initialized before the user's `beforeInit` runs.

No double-registration guard is needed: `HooksBucket.add()` already ignores a callback reference it holds, so the later `updateSettings()` pass is a no-op.

**Behavior change worth a reviewer's attention:** `beforeInit` callbacks sitting dormant in existing applications will start firing. This is not a breaking change under the project policy, since it makes a documented option work, but it is a real change in observable behavior.

**Two hooks are left unfixed in code, on purpose, and documented instead.** Both are typed options with the identical bug:

- `construct` fires inside the constructor, so fixing it means registering at construction time. That would flip hook ordering against anything added between `new Handsontable.Core()` and `.init()` — and the React wrapper does exactly that for `beforeViewRender` / `afterViewRender`.
- `afterPluginsInitialized` is run from `BasePlugin.init()`, i.e. during the `beforeInit` dispatch. Pulling it forward would land the user's callback *before* `enablePlugin()` runs, so a hook named "after plugins initialized" would fire before the plugins are enabled — worse than not firing.

Both carry a JSDoc note pointing to global registration.

**`beforeInit` itself is now documented too.** At that point the grid is only half-built: `datamap` is assigned by `loadData`/`updateData` inside `updateSettings()`, and `this.view` does not exist yet, so `countRows()`, `getData()`, and anything on `hot.view` throw. Since this change makes previously dormant callbacks run, that restriction needed to be stated on the hook.

**User-facing docs.** The events-and-hooks guide gains the `beforeInit` timing caveat and a note that `construct` and `afterPluginsInitialized` cannot be used as options. The 18.1→18.2 migration guide gains a section 4 covering the upgrade hazard, alongside the existing section 2, which documents the same class of "bugs, not a change of contract". In `core/settings.ts`, both dead options carry a TSDoc note, and `beforeInit` is widened to `(() => void) | (() => void)[]` so the array form the runtime has always accepted is expressible in TypeScript (`settings.types.ts` covers it).

**Reference docs.** `handsontable/.ai/HOOKS.md` and `handsontable/src/core/hooks/AGENTS.md` now record that `beforeInit` is the one settings-declared hook registered outside the `updateSettings` walk, and why the pull-forward must not be generalized. While editing them I also corrected a **pre-existing** error in both: they described hook removal as a soft delete with a `skip` flag and a `MAX_SKIPPED_HOOKS_COUNT` compaction threshold, which `bucket.ts` explicitly is not — it is a true delete on a linked list. Happy to split that correction out if reviewers would rather keep this PR to #5933.

### How has this been tested?

Ten new specs in `handsontable/src/core/hooks/__tests__/hooks.spec.js`. **Six** of them fail on `develop` before this change and pass after — measured by reverting `core.ts` to `develop` and running the file, not counted by eye.

The four that pass unchanged are regression nets rather than proof of the fix: the `addAsFixed` in-place replacement, the `new Handsontable.Core()` + `addHook()` + `init()` path, and two guards pinning the design decision — a settings-declared `construct` and a settings-declared `afterPluginsInitialized` must stay silent while a globally registered one fires. Those two catch different generalizations: the `afterPluginsInitialized` guard goes red if someone widens the walk at the current registration point, the `construct` guard if someone moves registration into the constructor.

```bash
# The regression specs
npm --prefix handsontable run test:e2e -- --testPathPattern="core/hooks/__tests__/hooks.spec"

# Full gate run on this branch's base
npm --prefix handsontable run test:unit    # 328 suites, 4043 tests passed
npm --prefix handsontable run test:e2e     # 9817 specs, 0 failures
npx eslint src/core.ts src/core/hooks/constants.ts src/core/hooks/__tests__/hooks.spec.js
npx tsc --noEmit -p tsconfig.json
```

The changelog entry is named `5933.json` rather than by PR number, per the public-issue row in `.changelogs/README.md`.

### Types of changes

### Related issue(s):

1. DEV-2672
2. #5933

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2672

---

> [!NOTE]
> **Medium Risk**
> Settings-declared `beforeInit` callbacks that were never exercised may now throw during init and prevent the grid from mounting; core init order changes are narrowly scoped but lifecycle hooks are sensitive.
> 
> **Overview**
> Fixes a long-standing bug where a **`beforeInit`** callback passed in the grid settings never ran because hook registration happened in **`updateSettings()`** after **`beforeInit`** had already fired.
> 
> **`init()`** now registers **`beforeInit`** from merged settings immediately before **`runHooks('beforeInit')`**, via a shared **`registerSettingsHook()`** helper also used by **`updateSettings()`**. Only **`beforeInit`** is registered early so settings callbacks do not run ahead of plugin hooks attached during that dispatch. Duplicate registration is avoided by existing hook-bucket deduplication.
> 
> **Observable behavior:** dormant **`beforeInit`** handlers in apps will start executing (documented option, not a breaking change). Callbacks must not call data/render APIs such as **`countRows()`** or use **`hot.view`** yet; **`construct`** and **`afterPluginsInitialized`** remain global-only and are called out in JSDoc, user guides, and the 18.1→18.2 migration guide.
> 
> Adds regression tests, changelog entry **5933**, and **`beforeInit`** typing for the array form. Internal hook docs were updated for **`beforeInit`** timing and for bucket removal semantics (true linked-list delete).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 438e56960ae9e7f8d2516d049b174f490085b362. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2672 Fix the beforeInit hook never firing when declared in the settings object](https://app.clickup.com/t/9015210959/DEV-2672)

qunabu (dismissed):

handsontable/src/core/hooks/constants.ts
```diff
@@ -1627,6 +1627,13 @@ export const REGISTERED_HOOKS = [
   /**
    * Fired before the Handsontable instance is initiated.
    *
+   * At this point the grid is only partly built. The settings are not applied yet, the data is not
+   * loaded, and the table is not rendered, so the plugins are initialized but most of the grid API is
```
Haxikowy: Two claims here are not true, and this text renders into the public API reference.

"The settings are not applied yet" — `MetaManager`'s constructor runs `globalMeta.updateMeta(settingsWithoutHooks)` (core.ts:615), and `getSettings()` reads `tableMeta`, which inherits from `globalMeta`. User settings are readable inside `beforeInit`; only the data, the theme resolution and the hook callbacks are still pending.

"the plugins are initialized" — `Hooks#run` dispatches the global bucket before the instance bucket (core/hooks/index.ts:213-214), so a callback added with `Handsontable.hooks.add('beforeInit', cb)` runs before any `BasePlugin.init()`. The same paragraph sends readers to global registration for `construct` and `afterPluginsInitialized`, so it promises a guarantee and then routes them to the path where it does not hold. Same for any instance listener with a negative `orderIndex`.

Suggest: drop the settings claim, and scope the plugin one — a `beforeInit` callback declared in the settings object runs after the plugins are initialized; a globally registered one runs before them.

Minor, same paragraph: `hot.view` is `undefined` at that point, so reading the property returns `undefined` — only calling a method on it throws.

.changelogs/5933.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed a bug where the `beforeInit` hook passed in the settings object never fired.",
```
Haxikowy: The title names the fix but not the upgrade hazard, and this is what lands in the release notes.

A settings-declared `beforeInit` that has been inert since v6 now executes at a point where `datamap` is unassigned and `this.view` does not exist. If its body calls `countRows()`, `getData()` or a `hot.view` method, it throws — and nothing wraps `runHooks('beforeInit')` (core.ts:1738), so the throw escapes before `updateSettings()` and before the view is built. The user gets no grid at all, and in the wrappers it propagates out of the mount lifecycle.

`"breaking": false` is right under the policy: a documented option that silently did nothing was never a working contract. But someone hitting the crash will read the changelog, not the JSDoc.

```suggestion
  "title": "Fixed a bug where the `beforeInit` hook passed in the settings object never fired. Such callbacks now run, before the grid is fully initialized.",
```

A line in the 18.1→18.2 migration guide would cover it better still.

handsontable/.ai/HOOKS.md
```diff
@@ -39,12 +39,13 @@ There is one global `Hooks` singleton, reached with `Hooks.getSingleton()` (`src
...
-- **Duplicate add is silently ignored.** `add` looks up the callback by reference. If the same function is already registered for that name, the call returns without adding a second entry.
+- **Removal is a true delete.** `remove` relinks the list around the entry. There is no soft-delete `skip` flag and no compaction threshold. An in-flight `run()` stays correct because it reads `entry.next` fresh after each callback, and `remove` never nulls a removed node's `next`, so a callback that removes itself can still advance.
+- **Duplicate add is silently ignored.** `add` walks the list and compares by reference. If the same function is already registered for that name, the call returns without adding a second entry. This check runs *before* the `initialHook` branch, so re-adding the same reference through `addAsFixed` is also a no-op.
+- **`addAsFixed` replaces in place.** When an entry for that name already has `initialHook`, `add` swaps its `callback` and keeps the node and its position, so an array previously returned by `getHooks()` reflects the swap. The framework wrappers depend on this.
```
Haxikowy: "The framework wrappers depend on this" — I could not find a dependent. No file under `wrappers/*/src` calls `getHooks` or `getBucket`; the only non-core caller in the repo is `wrappers/vue3/test/hotTable.spec.ts:121`, a test.

This promotes an unverified `bucket.ts` comment into a stated fact in the file agents read before touching the bucket, so it freezes the in-place-swap behavior behind a consumer that does not exist. Either name the real dependent, or say that no production caller relies on it today and one Vue test asserts it.

handsontable/.ai/HOOKS.md
```diff
@@ -198,6 +199,22 @@ A hook passed in the constructor settings or `updateSettings` attaches automatic
...
+it *after* it has already fired `beforeInit`. So a callback declared in the settings object cannot see
+any hook that fires before that call. `core.ts` therefore registers **`beforeInit` alone** up front,
+immediately before `runHooks('beforeInit')` (GitHub issue #5933); both call sites share the
+`registerSettingsHook()` helper, and re-adding the same reference is a no-op.
+
+Do not generalize that pull-forward to the whole settings object. Plugins attach their feature hooks
+inside `enablePlugin()`, which runs *during* the `beforeInit` dispatch, so registering everything early
+would place settings-declared callbacks ahead of every plugin callback, for every hook.
+
+Two hooks stay unreachable from the settings object by the same mechanism, and are documented as such
```
Haxikowy: Two problems in this paragraph.

The count is not two. Every hook dispatched during the `beforeInit` walk is equally unreachable from the settings object, including anything a plugin fires from `enablePlugin()`. `afterPluginsInitialized` is one member of that class, not half of a closed pair — an agent trusting "two" will read the third case as a bug.

The `construct` rationale is backwards: hooks a wrapper attaches between `new Handsontable.Core()` and `init()` are attached after the constructor returns, so nothing registered inside the constructor can reorder them. The actual reason is plugin construction order — `construct` fires as the constructor's last statement (core.ts:6857), and a settings-derived registration inside the constructor would sit ahead of plugin construction (core.ts:6842-6848), inverting plugin ordering for every later hook.

handsontable/src/core.ts
```diff
@@ -1705,6 +1728,13 @@ export default function Core(
     }
 
     dataSource.setData(tableMeta.data);
+
+    // `updateSettings()` below registers the callbacks declared in the settings object, which is too
+    // late for `beforeInit` — hence registering that one here. Re-adding the same reference is a no-op.
+    // Only `beforeInit` is pulled forward: doing it for the rest would order settings-declared
+    // callbacks ahead of the plugin callbacks registered while `beforeInit` runs.
+    registerSettingsHook('beforeInit', mergedUserSettings.beforeInit as HookCallback | HookCallback[] | undefined);
```
Haxikowy, line 1739: Dormant trap for whoever deprecates this hook: `beforeInit` now passes through the helper twice, here and again in the `updateSettings` walk. The function form is a true no-op (`addAsFixed` emits no removed/deprecated warning, and `bucket.add` dedups by reference before the `initialHook` branch). The array form routes through `Hooks#add`, whose deprecation `console.warn` is deliberately not deduped — per `core/hooks/AGENTS.md`. Nothing fires today since `beforeInit` is not deprecated, but the day it is, array-form users get two warnings per instance. Worth a line in the comment above.

handsontable/src/core/hooks/__tests__/hooks.spec.js
```diff
@@ -65,4 +65,129 @@ describe('Hooks', () => {
...
+
+    handsontable({
+      beforeInit: () => callOrder.push('beforeInit'),
+      afterInit: () => callOrder.push('afterInit'),
+    });
+
+    expect(callOrder).toEqual(['beforeInit', 'afterInit']);
+  });
+
+  it('should not register a hook declared in the settings object twice', async() => {
```
Haxikowy, line 118: The design decision this PR rests on is stated in three files and guarded by no test. Nothing goes red if someone generalizes `registerSettingsHook()` to the whole settings walk.

Two cases would pin it: a settings-declared `construct` never fires while `Handsontable.hooks.add('construct', cb)` does, and the same for `afterPluginsInitialized`. That also pins the JSDoc you just added for both.

Worth adding to this spec while you are here: `expect(getSettings().beforeInit).toBe(beforeInitHandler)`. The `tableMeta[key] = hook` write is the observable half of the helper's contract, it now happens twice for `beforeInit`, and no test covers the round-trip.

Haxikowy (commented):
The fix itself holds up — I checked the helper against the block it replaced, the double-registration path (function, array, mixed), ordering against plugin and wrapper callbacks, `tableMeta` readiness, and destroy/re-init. Clean. What I'm raising sits around the change, mostly in the docs, and is inline. The rest:

**Your split question:** keep the `skip` / `MAX_SKIPPED_HOOKS_COUNT` correction here. Those symbols went in `f32bcd237` (DEV-1982) and both doc files were left behind, so it's stale text, not a rewrite. But the risky material is the *new* narrative around it, not the correction — two inline comments are on those lines.

**Guide not updated.** `docs/content/guides/getting-started/events-and-hooks/events-and-hooks.md` teaches settings-object hooks as the normal path and lists `beforeInit` (line 230) with no timing caveat, and never says `construct` and `afterPluginsInitialized` are unreachable that way. The checklist wants guides for a user-facing behavior change, not only the API reference.

**Types contradict the new JSDoc.** `core/settings.ts:407` and `:632` still declare `afterPluginsInitialized` and `construct` as settings options that autocomplete and type-check. Deleting them would be breaking, so the same note as TSDoc on both. Pre-existing but newly relevant: `:536` types `beforeInit` as `() => void` while `hooks.spec.js:79` now asserts the array form — widening to `Fn | Fn[]` is non-breaking, and `settings.types.ts` would need the case.

**Wrappers, non-blocking.** Vue's `propFactory` exposes every hook name as a prop, so `<hot-table :construct="fn">` looks first-class and still never fires; fixing one of the three dead hooks sharpens that. And React StrictMode double-mounts, so a settings `beforeInit` fires twice in dev where it fired zero times — worth a JSDoc line that it runs once per instance creation.

**Description fix:** it says four of eight specs fail on `develop`; I read six. The two that pass unchanged are the `addAsFixed` replacement and the `new Core()` + `addHook()` path — good regression nets, not proof of the fix.

**Cleared, so nobody re-derives it:** adding `it()` blocks to `hooks.spec.js` is not a freeze violation (`.ai/TESTING.md` blocks new *files* and prescribes exactly this); the sentinel assertions aren't hollow, just confusing on failure (`Expected 'the hook was not called' to be undefined` — an explicit call-count assertion first would separate the modes); and `updateSettings()` from inside `beforeInit` was already reachable via a global hook, the #5933 workaround.

handsontable/src/core/hooks/constants.ts
budnix: Yep, my mistake — both claims were wrong. Fixed in 438e56960a: the settings claim is gone, and the ordering is now split between settings-declared and global callbacks.

.changelogs/5933.json
budnix: Good point — took your wording in 438e56960a, and added a section 4 to the 18.1→18.2 migration guide for the crash.

handsontable/.ai/HOOKS.md
budnix: Good spot — I checked, and the only caller is that one Vue test. Reworded in 438e56960a to say exactly that.

handsontable/.ai/HOOKS.md
budnix: You're right about the count — 438e56960a now calls it the whole class. I kept the wrapper half though: registering inside the constructor still puts settings hooks ahead of the two React attaches after it returns, so your plugin reason now leads and that one follows.

handsontable/src/core.ts
budnix: Nice catch — added that note to the comment in 438e56960a.

handsontable/src/core/hooks/__tests__/hooks.spec.js
budnix: Agreed — added both guards and the `getSettings()` round-trip in 438e56960a.

Haxikowy (approved):

Issue #12707, opened 2026-06-04
isCtrlPressed() is always false for any Handsontable instance that is not the first one created — additive (Ctrl/Cmd+click) selection silently breaks across iframe/window boundaries


mohkshu:
### Describe the bug

The ShortcutManager's modifier-key observer is only fed by keyboard listeners that are attached for the very first Handsontable instance created in a JS realm, and only on that instance's window chain. Every subsequent instance shares the same module-level observer but never installs its own keydown/keyup modifier listeners.

As a result, getShortcutManager().isCtrlPressed() (and .isPressed('control')) returns false for the 2nd+ instance whenever the Ctrl/Cmd keydown is delivered to a document that the first instance does not listen on — most notably when a second grid lives in a different iframe. Because setRangeStart() decides whether a click is additive from isCtrlPressed() rather than from MouseEvent.ctrlKey, Ctrl/Cmd+click in that grid replaces the selection instead of extending it, even though event.ctrlKey === true.

**_Root cause_**

Ctrl+click multi-selection in the grid is not decided from the mouse event itself — Handsontable decides it from a separately tracked keyboard state. In selection.js:364:
```
const isMultipleSelection = isUndefined(multipleSelection)
  ? this.tableProps.getShortcutManager().isCtrlPressed()   // <-- this decides ctrl+click
  : multipleSelection;
```

isCtrlPressed() (shortcuts/manager.js:204) reads a module-level, shared-across-all-instances observer in shortcuts/recorder.js:
```
const modifierKeysObserver = createKeysObserver();  // line 12 — ONE shared singleton
const modKeyListeners = [];                          // line 13
let instanceCounter = 0;                             // line 14
```

The listeners that actually populate that observer (onkeydownForModKeys / onkeyupForModKeys) are attached only for the first HoT instance, inside the mount loop (recorder.js:147-167):
```
const mount = () => {
  let eventTarget = ownerWindow;
  instanceCounter += 1;
  while (eventTarget) {
    if (instanceCounter === 1) {                     // <-- ONLY the first instance binds them
      eventTarget.document.documentElement.addEventListener('keydown', onkeydownForModKeys);
      eventTarget.document.documentElement.addEventListener('keyup', onkeyupForModKeys);
      ...
    }
    eventTarget.document.documentElement.addEventListener('keydown', onkeydown); // per-instance (regular shortcuts)
    eventTarget = getParentWindow(eventTarget);      // iframe window -> top window
  }
};
```

[...]

PR #9878, opened 2022-09-08
Fixed incorrect date format conversion on input (DateEditor)

wszymanski:
### Context

Issue #9681 describes use case which cause conversion from `7/1/22` to `01/07/2022` while the result should be equal to `07/01/2022`. The investigation has shown that problem was related to introduction of `ShortcutManager`, not as it was supposed at the beginning - related to update of `moment.js` (#9381, change introduced within the same release).

The problem has existed, because introduced `ShortcutManager` captured events on `window` object instead of `document` element. Events captured on `window` are triggered later. Thus, `change` event captured by `Pikaday` https://github.com/Pikaday/Pikaday/blob/21f676e70d688d18b265f2c12fc38e8457c20645/pikaday.js#L586-L600 has been overwriting cell's value at the wrong time. What's more, until `12.0.0` release Handsontable has been stopping propagation of event. Without this behaviour, the current code seems effect in some flickering. 

https://user-images.githubusercontent.com/141330/189110464-7bc037ae-7f22-460e-bda0-9ad14ad4d5e6.mp4

This PR change changes this behavior.

I've tried to add [an extra test](https://gist.github.com/wszymanski/0b4998f3aa54d8f9bd9b28d268eb4186), but it seems that simulation done by Jasmine isn't perfect. The `change` event isn't simulated properly.

### How has this been tested?

1. I tested case shown in the issue.
2. I tested other ways of closing a cell.
3. I tested whether opened `Pikaday` widget looks properly for chosen `dateFormat`

Screenshot 2022-09-08 at 13 20 22

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature or improvement (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Additional language file or change to the existing one (translations)

### Related issue(s):
1. #9681 

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

- [x] I have reviewed the guidelines about [Contributing to Handsontable](https://github.com/handsontable/handsontable/blob/master/CONTRIBUTING.md) and I confirm that my code follows the code style of this project.
- [ ] My change requires a change to the documentation.

PR #13286, opened 2026-08-28
DEV-2674: Release a modifier key whose keyup was consumed by the OS (#6012)

### LINKED ISSUE #6012: Cmd/ctrl/shift key state incorrect after activating OS features using those keys
### Description
When the OS intercepts the keyup while using a shift/cmd key combination on Mac (and I think Ctrl on other platforms too), the key state gets stuck in Handsontable and selection works incorrectly until another Cmd/Ctrl keyup

### Steps to reproduce
1. Click a cell in the table, to select it

2. On Mac, press Cmd+Shift+4 to take a partial screenshot.  On Windows, maybe try Ctrl+Shift+Esc or something?  Sorry I don't have access to Windows right now.  The important thing is it needs to be a key combination using Cmd/Ctrl/Shift that the OS will intercept so the browser doesn't receive the keyup event.

3. (On Mac) Use the mouse to finish taking the screenshot.  At this point no keys should be pressed on the keyboard.

4. Press an arrow key to navigate to another cell.  As you move across cells or click other cells they will be added to the selection, as if Cmd or Shift were being held.

### Demo
https://jsfiddle.net/handsoncode/8ffpsqt6/ (this is the default fiddle, unmodified)
This can be reproduced on the default fiddle, or on https://handsontable.com/examples?manual-resize

### Your environment
* Handsontable version: 6.2.1
* Browser Name and version: Google Chrome Version 74.0.3729.131 (Official Build) (64-bit)
* Operating System: macOS Mojave 10.14.4

budnix:
### Context

The PR fixes [#6012](https://github.com/handsontable/handsontable/issues/6012), reported in 2019 and confirmed to still reproduce on released 18.0.0 and on `develop`.

When the operating system takes over a shortcut that uses a modifier key — the macOS `Cmd+Shift+4` screenshot being the reported case — the browser receives the `keydown` but never the matching `keyup`. `recorder.ts` added a modifier key to its held-keys set on `keydown` and removed it only on `keyup`, so the key stayed marked as held forever.

`ShortcutManager#isCtrlPressed()` reads that set. `Selection#setRangeStart()` falls back to it whenever the caller does not state whether the selection is additive, and arrow-key navigation (`transformStart`) is exactly such a caller. The result: after the OS shortcut, every arrow key press **added a selection layer** instead of moving the selection.

Measured on both 18.0.0 and `develop` — select B2, then two Arrow Downs produced three separate ranges, `(1,1)`, `(2,1)` and `(3,1)`, with three cells highlighted instead of one.

#### Why this stayed open so long

Three things make a careless check pass, which is worth knowing when reviewing:

1. **Any plain mouse click on a cell repairs the state.** `core.ts` already releases the held keys on `beforeOnCellMouseDown` when the event reports Ctrl and Cmd as up. Clicking a cell hides the bug completely, so the repro has to stay on the keyboard.
2. **A stuck `Shift` is now inert.** The set still tracks `shift`, but only `control` and `meta` are ever read. The half of the 2019 report that blamed Shift no longer applies.
3. The grid has to be listening, or arrow keys do nothing at all.

### The fix

`syncModifierKeysState()` in `handsontable/src/shortcuts/recorder.ts` rewrites all four modifier keys from the event's own `altKey` / `ctrlKey` / `metaKey` / `shiftKey` flags, and both the `keydown` and the `keyup` observer run it before touching the named key. Those flags always carry the true state, so a key whose `keyup` was eaten is released on the very next key event — the selection is already correct on the first arrow press, not the second. This is the keyboard counterpart of the repair the mouse path already had.

Two details worth a reviewer's eye:

- The sync deliberately runs **before** the named-key press. That press is a correction for browsers that do not set the matching flag on a modifier key's own `keydown`; reordering the two would mean such a modifier could never be marked pressed.
- Making the event authoritative also means a **synthetic** key event dispatched without those flags now releases the modifier keys. No code in `handsontable/src` or `wrappers/*/src` dispatches such an event, but integrator code (a virtual-keyboard widget, a jQuery `trigger('keydown')`) could. This is the deliberate cost of the fix, and it is documented in the function's JSDoc and in `handsontable/AGENTS.md`.

`setPressed()` was added to the key observer so the sync is four flag reads with no per-keystroke allocation — this handler runs for every keystroke on the page, including autorepeat.

No public API, setting, or default changes. Not breaking.

### Deliberately **not** in this PR

While investigating I found a second, pre-existing defect: the `blur` net that is supposed to release held keys is dead code. It is registered in the bubble phase on the `documentElement`, but `blur` does not bubble and a window-level blur is dispatched on the `window` itself. It has not fired since #12737 moved it off the window. Verified in the browser: a window blur and an element blur both leave the key held.

I prototyped the one-line move back to the `window` and then **reverted it**, because it is not a one-line change in practice. The observer is module-level and shared across every instance in the realm — including grids the host page creates inside a same-origin iframe (#12707) — and `mount()` walks the whole parent-window chain, so a blur on any ancestor window would wipe the modifier state of every other grid. That combination has never been live: the listener worked on the `window` in 2022, was silenced by #12737, and the cross-iframe shared tracking landed afterwards in that dead window.

I could not reproduce a failure (a real Cmd+click descent into an iframe grid fired no parent-window blur and kept the additive selection), but I also could not prove the focus path safe, and the only test possible in the Jasmine suite dispatches a synthetic `FocusEvent`, which proves nothing about a real focus transition. It deserves its own ticket and a Playwright test driving genuine focus changes rather than riding along here. The gotcha is recorded in `handsontable/AGENTS.md` so the next person does not "fix" it in one line.

Nothing regresses by leaving it: the per-event sync above heals the keyboard path, and `beforeOnCellMouseDown` heals the mouse path.

### How has this been tested?

Four regression tests were added to the existing `handsontable/src/shortcuts/__tests__/shortcutManager.spec.js`, next to the `#dev-1025` mouse-repair test they complement. They cover the keydown path, the arrow-key selection symptom, the keyup path, and a guard in the other direction — a modifier the event still reports as held must **not** be released, so the fix cannot break Ctrl/Cmd+click multi-selection.

The first three **fail on the unfixed source** and pass with the fix, verified by reverting `recorder.ts` and `keyObserver.ts` to the base commit and re-running (`35 specs, 3 failures`). The fourth is the guard, so it passes on both sides by design — that is what makes it a guard. The most telling failure was the selection piling up:

```
Expected `["highlight: 1,1 from: 1,1 to: 1,1","highlight: 2,1 from: 2,1 to: 2,1","highlight: 3,1 from: 3,1 to: 3,1"]`
to match to the `["highlight: 3,1 from: 3,1 to: 3,1"]` cell range pattern.
```

They were added to the frozen Jasmine suite rather than Playwright as a deliberate call: they extend an existing spec file (which `handsontable/.ai/TESTING.md` permits) and belong next to the sibling test for the mouse half of the same mechanism.

Suites run before committing, all from `handsontable/`:

```bash
npm run eslint                                                        # clean
npm run test:types                                                    # clean
npm run test:unit                                                     # 325 suites, 4000 tests passed
npm run test:e2e --testPathPattern=shortcuts                          # 1213 specs, 0 failures
npm run test:e2e --testPathPattern=selection                          # 1270 specs, 0 failures
npm run test:e2e --testPathPattern="columnSorting|editorManager|copyPaste"  # 583 specs, 0 failures
npm run test:e2e --testPathPattern=mergeCells                         # 515 specs, 0 failures
```

The E2E patterns were chosen to cover every caller of `isCtrlPressed()` — selection, `columnSorting`, `multiColumnSorting` and `editorManager` — plus the Ctrl+click paths in `mergeCells` and `copyPaste`.

Because the test helper rebuilds the modifier flags per call, a spec that holds a modifier in one `keyDown()` and then presses another key in a separate call without re-listing it would lose the modifier under the new behavior. A sweep of all 1,268 spec/test files found exactly one such site (`selection/__tests__/selection/refresh.spec.js:114-119`); it passes, because the three `keyDownUp('enter')` there only move the focus, which never reads `isCtrlPressed()`. The pattern cannot occur in a real browser.

Manually verified in Chrome against a local build, comparing a CDN 18.0.0 page with the patched build side by side: the stuck state was induced by dispatching the `keydown` the browser really does deliver and never dispatching the `keyup`, then navigating with **real** key presses. Confirmed beforehand that a genuine trusted `Meta` keydown sets the same flag and its real keyup clears it, so only the missing `keyup` — the bug's premise — was simulated.

### Types of changes

### Related issue(s):

1. DEV-2674
2. #6012

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2674

---

> [!NOTE]
> **Medium Risk**
> Touches global per-keystroke modifier tracking used by selection, sorting, copy/paste, and editors; behavior changes for synthetic keyboard events without modifier flags, though guarded for malformed events.
> 
> **Overview**
> Fixes **#6012**, where a modifier could stay “held” after the OS ate its `keyup` (e.g. macOS screenshot shortcuts), so arrow keys kept **adding selection layers** instead of moving the cell.
> 
> **`recorder.ts`** now runs **`syncModifierKeysState()`** on every `keydown`/`keyup` for the shared modifier observer, setting Alt/Ctrl/Meta/Shift from the event’s `*Key` flags **before** the named-key press/release logic. **`keyObserver`** gains **`setPressed()`** for that sync. Events without a string **`key`** still skip sync so modifier state is not wiped (**#dev-2096**); modifier-only `keydown` without matching flags is corrected by pressing the key after sync.
> 
> **`ShortcutManager#isCtrlPressed()`** JSDoc notes that synthetic events without modifier flags now report modifiers as released. Regression tests cover release on the next key, arrow selection behavior, held/released guards, and edge cases. **`AGENTS.md`** documents the behavior (including test-helper and dead `blur` listener notes); changelog **6012** added.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 60d40b38b2e2b6953597e9cb7dcfd394065fbcfd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2674 Fix stuck Cmd/Ctrl modifier key state after an OS shortcut (#6012)](https://app.clickup.com/t/9015210959/DEV-2674)

handsontable/src/shortcuts/recorder.ts
```diff
@@ -16,26 +16,49 @@ const modifierKeysObserver = createKeysObserver();
...
+  modifierKeysObserver.setPressed('shift', event.shiftKey);
+};
+
+/**
+ * `KeyboardEvent`'s callback function for observing the state of the mod keys on key press.
  *
  * @param {KeyboardEvent} event The event object
  */
 const onkeydownForModKeys = (event: KeyboardEvent) => {
+  syncModifierKeysState(event);
```
demtario: **medium** — `syncModifierKeysState(event)` runs *outside* the `typeof event.key === 'string'` guard right below it, in both handlers. On an event whose `key` is not a string, all four flag reads are `undefined`, so every `setPressed(key, false)` releases that modifier — for every grid in the realm, since the observer is module-level.

That event class is real, not hypothetical: the guard exists because of #11281 (dev-2096), and its regression spec sits in the file this PR edits — [`shortcutManager.spec.js:196`](https://github.com/handsontable/handsontable/blob/6a7c599de95b693e15df15dc8d56cd61d9083f44/handsontable/src/shortcuts/__tests__/shortcutManager.spec.js#L196) dispatches `keyTriggerFactory('keydown', undefined, { extend: {} })` with no modifier properties at all. It asserts only that the callback did not fire, so the wipe passes it unnoticed.

Fix: move the call inside the existing guard, here and in `onkeyupForModKeys`. A genuine `KeyboardEvent` always has a string `key`, so it costs nothing. Worth extending that spec too: press `control/meta`, dispatch the `key: undefined` event, assert `isCtrlPressed()` is still `true`.

handsontable/src/shortcuts/recorder.ts
```diff
@@ -16,26 +16,49 @@ const modifierKeysObserver = createKeysObserver();
...
+ * The OS can consume a modifier key's `keyup` (e.g. the macOS Cmd+Shift+4 screenshot shortcut),
+ * which would otherwise leave that key marked as held forever. Every key event carries the current
+ * state of the four modifier keys, so the event is the authoritative source. Consequently a
+ * synthetic event dispatched without those flags set releases the modifier keys.
+ *
+ * @param {KeyboardEvent} event The event object
+ */
+const syncModifierKeysState = (event: KeyboardEvent) => {
+  modifierKeysObserver.setPressed('alt', event.altKey);
+  modifierKeysObserver.setPressed('control', event.ctrlKey);
```
demtario, line 30: **medium** — this is the fifth site in core that answers "is Ctrl/Cmd held" from event flags, and the only one without the right-Alt carve-out. The other four all exclude it, with the same comment: [`editorManager.ts:458`](https://github.com/handsontable/handsontable/blob/6a7c599de95b693e15df15dc8d56cd61d9083f44/handsontable/src/editorManager.ts#L458), [`textEditor.ts:439`](https://github.com/handsontable/handsontable/blob/6a7c599de95b693e15df15dc8d56cd61d9083f44/handsontable/src/editors/textEditor/textEditor.ts#L439), [`mergeCells.ts:948`](https://github.com/handsontable/handsontable/blob/6a7c599de95b693e15df15dc8d56cd61d9083f44/handsontable/src/plugins/mergeCells/mergeCells.ts#L948), [`undoRedo.ts:130`](https://github.com/handsontable/handsontable/blob/6a7c599de95b693e15df15dc8d56cd61d9083f44/handsontable/src/plugins/undoRedo/undoRedo.ts#L130).

Whether it changes behavior rests on one thing I could not verify from macOS: if every supported Windows browser delivers a `key: "Control"` keydown ahead of the `AltGraph` one, then `press('control')` already ran before this PR and nothing changes. If any does not, the observer now reports Ctrl held for as long as AltGr is held, and nothing clears it — the `beforeOnCellMouseDown` release in `core.ts` requires `event.ctrlKey === false`, which never holds while AltGr is down. Consumers reached inside that window: [`selection.ts:375`](https://github.com/handsontable/handsontable/blob/6a7c599de95b693e15df15dc8d56cd61d9083f44/handsontable/src/selection/selection.ts#L375), `editorManager` `#onCellDblClick`, [`columnSorting.ts:943`](https://github.com/handsontable/handsontable/blob/6a7c599de95b693e15df15dc8d56cd61d9083f44/handsontable/src/plugins/columnSorting/columnSorting.ts#L943).

The guard is worth adding either way. `event.ctrlKey && !event.getModifierState('AltGraph')` is more precise than the `!altKey` the four sites above use — it keeps a genuine Ctrl+Alt hold, which `!altKey` drops.

handsontable/AGENTS.md
```diff
@@ -77,6 +77,8 @@ Gotcha: Filters `conditionCollection` uses physical indexes, `getDataAtCol()` us
...
+- **`blur` does not bubble, so a listener for it on the `documentElement` is dead code.** `recorder.ts` still registers its `onblur` net that way (bubble phase, `documentElement`), and it has never fired since #12737 moved it off the `window`: a real window-level blur is dispatched on the `window` itself and an element blur does not propagate upward. Do not assume the modifier state is cleared when the page loses focus — it is not; the per-event re-sync above and the `beforeOnCellMouseDown` release in `core.ts` are what actually heal it. Reviving the net is not a one-line move to the `window`: the observer is module-level and shared across every instance in the realm, including grids the host page created inside a same-origin iframe (#12707), so a blur on any window in the parent chain would wipe the modifier state of every other grid — verify against a real focus transition, not a dispatched `FocusEvent`.
```
demtario: **medium** — #12737 did not move this listener off the `window`; #9878 did. Commit `6522285363` changed `eventTarget.addEventListener('blur', onblur)` to `eventTarget.document.documentElement.addEventListener('blur', onblur)`, and at `8a0c7d22e^` — the #12737 pre-image — it was already on `documentElement`; #12737 only rewrote that expression into the destructured form. The original window binding came in with `c06c32ba1d`.

Read `#9878` here. The `(#12707)` citation later in the bullet is correct (12707 is the public issue, 12737 its PR), so only this one number needs changing.

handsontable/src/shortcuts/recorder.ts
```diff
@@ -16,26 +16,49 @@ const modifierKeysObserver = createKeysObserver();
...
  *
  * @param {KeyboardEvent} event The event object
  */
 const onkeydownForModKeys = (event: KeyboardEvent) => {
+  syncModifierKeysState(event);
+
   if (typeof event.key === 'string') {
     const pressedKey = normalizeEventKey(event);
 
+    // Runs after the sync on purpose. A browser that does not set the matching flag on a modifier
```
demtario, line 48: **medium** — the ordering that this comment and AGENTS.md consequence (1) both call load-bearing has no test. Move `syncModifierKeysState(event)` below the `isModifierKey` block and all four new specs still pass, so nothing stops a later refactor from reversing it.

The Jasmine helper cannot express the event shape that would catch it: `saveStateAndExtendEvent` overwrites a caller-supplied `extend.ctrlKey`/`metaKey` before `$.extend(ev, extend)` ([`keyboardEvents.js:79-82`](https://github.com/handsontable/handsontable/blob/6a7c599de95b693e15df15dc8d56cd61d9083f44/handsontable/test/helpers/keyboardEvents.js#L79), [`:150-157`](https://github.com/handsontable/handsontable/blob/6a7c599de95b693e15df15dc8d56cd61d9083f44/handsontable/test/helpers/keyboardEvents.js#L150)). It needs a raw dispatch of a `Control`/`Meta` keydown with the flag explicitly false, then `expect(getShortcutManager().isCtrlPressed()).toBe(true)` — in-file precedent for the raw call at [`shortcutManager.spec.js:205`](https://github.com/handsontable/handsontable/blob/6a7c599de95b693e15df15dc8d56cd61d9083f44/handsontable/src/shortcuts/__tests__/shortcutManager.spec.js#L205).

Related, in the same block: the specs at lines 551 and 566 end on a bare `keyDown('control/meta')` with no matching `keyUp`, which is the pattern the AGENTS.md bullet added in this PR tells future agents to avoid. `PRESSED_KEYS` is module-level and `unmount()` never clears it, so if this fix ever regresses, `meta` stays set for the rest of the file and the failure cascades into later specs instead of landing on these two.

demtario (commented):
4 inline notes below.

**Off-diff**

- [`handsontable/src/shortcuts/manager.ts:277-283`](https://github.com/handsontable/handsontable/blob/6a7c599de95b693e15df15dc8d56cd61d9083f44/handsontable/src/shortcuts/manager.ts#L277) — **low** — `isCtrlPressed()`'s JSDoc is the source of the public API reference and still reads "Returns whether `control` or `meta` keys are pressed". The new semantics — the value is re-derived from every key event's modifier flags, so a synthetic key event dispatched without them reports the keys as released — are documented only in `handsontable/AGENTS.md`, which users never read. One sentence there covers it; the PR checklist's documentation item is currently unchecked.

handsontable/src/shortcuts/recorder.ts
budnix: Nice catch — fixed in 60d40b38b2, the sync now runs inside the `key` guard in both handlers, with a spec that holds control/meta across that event.

handsontable/src/shortcuts/recorder.ts
budnix: Fair point on the inconsistency, but it predates this PR — AltGr fires a `key: "Control"` keydown, so `press('control')` already ran here before. Happy to do the carve-out in its own ticket where someone can check it on Windows.

handsontable/AGENTS.md
budnix: Good spot — corrected to #9878 (`6522285363`) in 60d40b38b2.

handsontable/src/shortcuts/recorder.ts
budnix: You're right, nothing pinned that order — added a spec for it in 60d40b38b2, and both tests now end with a matching `keyUp`.

budnix:
On the off-diff note about `isCtrlPressed()` — good point, its JSDoc now says the state is re-read from every key event's modifier flags, in 60d40b38b2.

demtario (approved):

PR #13288, opened 2026-08-28
DEV-2675: Accept pixel-size strings in the header size options (#6154)

### LINKED ISSUE #6154: rowHeadersWidth cannot be string
### Description

rowHeadersWidth cannot take string values.
Other properties such as maxRows, height, width, and columnHeaderHeight accept string values (perhaps the values get converted to int via parseInt()).

### Steps to reproduce

1. Add rowHeadersWidth
2. Pass a string value - won't work
3. Pass a integer value - will work

### Demo

https://jsfiddle.net/y8a6vcr0/

### Your environment
* Handsontable version: 7.1.0
* Browser Name and version: Google Chrome Version 75.0.3770.142 (Official Build) (64-bit)
* Operating System: macOS 10.14.5

budnix:
### Context

The PR fixes [#6154](https://github.com/handsontable/handsontable/issues/6154): `rowHeaderWidth` ignores every string value.

The two header size options disagreed for reasons that were never designed:

- `rowHeaderWidth` passed through a `typeof width === 'number'` guard in `_correctRowHeaderWidth`, which replaced any string with `defaultColumnWidth` (50px). A string never applied at all.
- `columnHeaderHeight` had no guard and reached `Math.max(height, providedHeight)`. `Math.max` converts its arguments to numbers, so `'100'` worked — by accident — while `'100px'` became `NaN` and was ignored.

So the "works" case was the accident, not the "broken" one.

#### An unreleased regression found while investigating

The single-pass layout work (`294dfcae23`, #12951, DEV-1995) added a `typeof configuredAtLevel === 'number'` gate to the `columnHeaderHeight` merge in `tableView.ts`. That commit is tagged in `18.1.0-rc*` only — not in 18.0.0 — so both effects below are unreleased, but they ship with 18.1.0:

1. `columnHeaderHeight: '100'` shrank the header by **1px on every draw**, with no floor. Measured in Chrome: 100 → 99 → 98 → … → 87 after 12 draws. The string failed the gate, so the option dropped out of the merge and the render-size probe became the only height source — and the probe comes back a pixel short each round, feeding itself.
2. `columnHeaderHeight: ['100']` stopped applying entirely (rendered at the default 29px). It worked in released 18.0.0.

Content-driven headers are unaffected: a tall header with no `columnHeaderHeight` option holds steady across 30 draws (one unique measured value), as do the plain default and content-plus-number cases. The drift was specific to the string path.

### The change

Both options are resolved **once**, where each crosses from the grid settings into Walkontable (the `tableView` settings funnel) — deliberately not in `getHeaderHeight` / `calculateWidths`, which run on every draw.

- New `parsePixelSize()` in `helpers/number.ts`: a number passes through unchanged (including a negative one, so the numeric path is untouched); `'100'` and `'100px'` resolve to `100`; anything else returns `null`.
- A value that states no pixel count (`'50%'`, `'20em'`, text) is ignored, the default size is used, and a one-time warning per grid instance says why. `warnOnce` keys on `hot.rootElement`, so a grid that redraws many times still warns once.
- `GridSettings` widens both options to `number | number[] | string | Array<number | string>` — the **same union `colWidths` and `rowHeights` already declare** in that file, so this matches an existing convention rather than adding one. Without it the new forms would be a compile error for every TypeScript user and all three wrappers, which would make the feature unusable for most of the people who hit the problem.
- A value that is already a number, or an array already made of numbers, is returned **by reference**, so the common path allocates nothing per draw.

#### A note on scope — this pushes back on the report

The issue argues `rowHeaderWidth` should accept strings because `height` and `width` do. Those are CSS pass-throughs; measured on this branch they accept `'300'`, `'300px'`, `'20em'` and `'50%'`. These two options are not that.

So the consistency fix runs the opposite way to the report: both options stay **pixel** sizes, and the accidental string support in `columnHeaderHeight` becomes a defined, documented rule instead of a side effect of `Math.max`. `'100px'` is accepted alongside `'100'` on purpose — accepting only bare numeric strings would keep the trap where `'100px'`, the more natural string to write, is silently ignored, and would add a third convention next to the stated direction for `height`/`width` (numbers or a `px` postfix).

Dropping string support outright was the other option. It is a breaking change under the project policy — `columnHeaderHeight: '100'` and `['100']` do work in released 18.0.0 — so this coerces at the boundary instead.

One inconsistency is left deliberately: `colWidths` / `rowHeights` run their strings through `Number.parseInt`, so `colWidths: '50%'` silently becomes 50px. The new options reject that instead of guessing. Matching the looser behavior would mean reading `'50%'` as `50px` in two more options; changing `colWidths` to match the stricter rule would be a breaking change. Worth settling separately for all four options rather than inside this fix.

Worth flagging separately: the issue title says `rowHeadersWidth` (plural). That is not an option at all and is silently ignored. The real option is `rowHeaderWidth`, singular.

### How has this been tested?

**Unit** — 9 new cases for `parsePixelSize` (numbers unchanged, `'100'`/`'100px'`/case/whitespace, and `null` for `'50%'`, `'20em'`, `'-100'`, `''`, `'100px100'`, non-strings):

```bash
npm run test:unit --prefix handsontable -- --testPathPattern=helpers/__tests__/number
```

Full core unit suite: **4052 passed, 328 suites**. `npm run test:types` passes, which also exercises the widened union — `settings.types.ts` now pins `'25'`, `'25px'` and `['25px', 30]` on both options, and those lines would not compile against the old `number | number[]`.

**E2E** — 12 Playwright tests, including one that forces 10 extra draws to pin the drift, one for `columnHeaderHeight: '100px'` (the only genuinely broken column-header case), a mixed `[60, '90px']` array and an `['20em', 90]` one, and two that assert the warning fires once per grid and stays silent for every supported form:

```bash
cd tests && npx playwright test --project=e2e-main e2e/header-size-string-values.spec.ts
```

**Proven red before green.** Re-measured against the full spec after review: with `tableView.ts` reverted to `develop` and `utils/pixelSize.ts` removed, **10 of the 12 tests fail**. The 2 that pass are the numeric control and the invalid-value fallback — both of which should already hold, so that is the expected split rather than a gap.

**Manual, in Chrome**, released 18.0.0 vs this branch (baseline with no option set: 50px row header / 29px column header):

| Value | `rowHeaderWidth` before | after | `columnHeaderHeight` before | after |
|---|---|---|---|---|
| `100` | 100 | 100 | 100 | 100 |
| `'100'` | 50 | **100** | 99 → 98 → 97 … | **100, stable** |
| `'100px'` | 50 | **100** | 29 | **100** |
| `['100']` | 50 | **100** | 29 (100 in 18.0.0) | **100** |
| `'50%'` | 50 | 50 + warning | 29 | 29 + warning |

### Types of changes

### Related issue(s):

1. DEV-2675
2. #6154

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

- [x] My change requires a change to the documentation. The `rowHeaderWidth` and `columnHeaderHeight` JSDoc in `metaSchema.ts` states the accepted forms and what happens to a value that is not a pixel size, and the row-header and column-header guides list the string form.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2675

---

> [!NOTE]
> **Low Risk**
> Localized to header dimension settings at the TableView boundary; numeric behavior is preserved and invalid values fail safe with warnings rather than changing auth or data paths.
> 
> **Overview**
> Fixes **#6154** by resolving `rowHeaderWidth` and `columnHeaderHeight` when they arrive as strings (e.g. from JSON or framework bindings), not only as numbers.
> 
> Both options now accept **`'100'`** and **`'100px'`** in scalar and per-level array form, with numbers and numeric arrays unchanged on the fast path. Resolution happens in **`tableView`** via new **`parsePixelSize`** and **`resolveHeaderSizeSetting`**, so Walkontable always sees pixel numbers before layout. Non-pixel values (`'50%'`, `'20em'`, negative strings) fall back to the default with a **one-time console warning** per grid; invalid entries in an array are handled per level.
> 
> **Types**, **metaSchema** JSDoc, user **guides**, a **changelog** entry, **unit** tests for `parsePixelSize`, and **Playwright** coverage (including repeated-draw stability for string `columnHeaderHeight`) accompany the behavior change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b91787ce6aaa4897eb53b908170d8246c210c59c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changelogs/13288.json
+++ b/.changelogs/13288.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Fixed the `rowHeaderWidth` option ignoring values passed as strings. Both `rowHeaderWidth` and `columnHeaderHeight` now accept a pixel size written as a string, either as a bare number (`'100'`) or with the unit (`'100px'`), in the scalar and the per-level array form alike. A value that states no pixel count, such as `'50%'` or `'20em'`, is ignored in favor of the default size and reported once in the console.",
+  "type": "fixed",
+  "issueOrPR": 13288,
+  "breaking": false,
+  "framework": "none"
+}
--- a/handsontable/src/dataMap/metaManager/metaSchema.ts
+++ b/handsontable/src/dataMap/metaManager/metaSchema.ts
@@ -895,21 +895,22 @@ export default (): Record<string, unknown> => {
      *
      * You can set the `columnHeaderHeight` option to one of the following:
      *
-     * | Setting  | Description                                         |
-     * | -------- | --------------------------------------------------- |
-     * | A number | Set the same height for every column header         |
-     * | An array | Set different heights for individual column headers |
+     * | Setting  | Description                                                     |
+     * | -------- | --------------------------------------------------------------- |
+     * | A number | Set the same height for every column header                     |
+     * | A string | Set the same height, written as a pixel size (`'25'`, `'25px'`)  |
+     * | An array | Set different heights for individual column headers             |
      *
      * The height is a number of pixels. A string that states a pixel size works too, either as a bare
      * number (`'25'`) or with the unit (`'25px'`), so a value coming from an attribute or a JSON
-     * config still applies. A value that is not a pixel size, such as `'50%'` or `'20em'`, is ignored
-     * and the default height is used instead.
+     * config still applies. Both forms may be mixed inside the array. A value that is not a pixel
+     * size, such as `'50%'` or `'20em'`, is ignored and the default height is used instead.
      *
      * This option can only be set at the [grid level](@/guides/getting-started/configuration-options/configuration-options.md#set-grid-options).
      * It has no effect when set in the [`columns`](#columns), [`cells`](#cells), or [`cell`](#cell) options.
      *
      * @memberof Options#
-     * @type {number|number[]}
+     * @type {number|number[]|string|string[]|Array<number|string>}
      * @default undefined
      * @category Core
      *
@@ -917,6 +918,9 @@ export default (): Record<string, unknown> => {
      * ```js
      * // set the same height for every column header
      * columnHeaderHeight: 25,
+     *
+     * // set the same height, written as a pixel size
+     * columnHeaderHeight: '25px',
      *
      * // set different heights for individual column headers
      * columnHeaderHeight: [25, 30, 55],
@@ -5607,15 +5611,16 @@ export default (): Record<string, unknown> => {
      *
      * You can set the `rowHeaderWidth` option to one of the following:
      *
-     * | Setting  | Description                                     |
-     * | -------- | ----------------------------------------------- |
-     * | A number | Set the same width for every row header         |
-     * | An array | Set different widths for individual row headers |
+     * | Setting  | Description                                                    |
+     * | -------- | -------------------------------------------------------------- |
+     * | A number | Set the same width for every row header                        |
+     * | A string | Set the same width, written as a pixel size (`'25'`, `'25px'`) |
+     * | An array | Set different widths for individual row headers                |
      *
      * The width is a number of pixels. A string that states a pixel size works too, either as a bare
      * number (`'25'`) or with the unit (`'25px'`), so a value coming from an attribute or a JSON
-     * config still applies. A value that is not a pixel size, such as `'50%'` or `'20em'`, is ignored
-     * and the default width is used instead.
+     * config still applies. Both forms may be mixed inside the array. A value that is not a pixel
+     * size, such as `'50%'` or `'20em'`, is ignored and the default width is used instead.
      *
      * Row headers have a fixed width. A label longer than that width is clipped, and unlike column
      * headers, the header does not grow to fit it. To size the header to its content instead, turn
--- a/handsontable/src/dataMap/metaManager/metaSchema.ts
+++ b/handsontable/src/dataMap/metaManager/metaSchema.ts
@@ -900,6 +900,11 @@ export default (): Record<string, unknown> => {
      * | A number | Set the same height for every column header         |
      * | An array | Set different heights for individual column headers |
      *
+     * The height is a number of pixels. A string that states a pixel size works too, either as a bare
+     * number (`'25'`) or with the unit (`'25px'`), so a value coming from an attribute or a JSON
+     * config still applies. A value that is not a pixel size, such as `'50%'` or `'20em'`, is ignored
+     * and the default height is used instead.
+     *
      * This option can only be set at the [grid level](@/guides/getting-started/configuration-options/configuration-options.md#set-grid-options).
      * It has no effect when set in the [`columns`](#columns), [`cells`](#cells), or [`cell`](#cell) options.
      *
@@ -5606,6 +5611,11 @@ export default (): Record<string, unknown> => {
      * | -------- | ----------------------------------------------- |
      * | A number | Set the same width for every row header         |
      * | An array | Set different widths for individual row headers |
+     *
+     * The width is a number of pixels. A string that states a pixel size works too, either as a bare
+     * number (`'25'`) or with the unit (`'25px'`), so a value coming from an attribute or a JSON
+     * config still applies. A value that is not a pixel size, such as `'50%'` or `'20em'`, is ignored
+     * and the default width is used instead.
      *
      * Row headers have a fixed width. A label longer than that width is clipped, and unlike column
      * headers, the header does not grow to fit it. To size the header to its content instead, turn
--- a/handsontable/src/dataMap/metaManager/metaSchema.ts
+++ b/handsontable/src/dataMap/metaManager/metaSchema.ts
@@ -906,6 +906,10 @@ export default (): Record<string, unknown> => {
      * config still applies. Both forms may be mixed inside the array. A value that is not a pixel
      * size, such as `'50%'` or `'20em'`, is ignored and the default height is used instead.
      *
+     * A negative number is kept as it is, because numbers keep the behavior they had before this
+     * option read strings at all. A negative string is rejected instead, so a typo cannot collapse
+     * the header.
+     *
      * This option can only be set at the [grid level](@/guides/getting-started/configuration-options/configuration-options.md#set-grid-options).
      * It has no effect when set in the [`columns`](#columns), [`cells`](#cells), or [`cell`](#cell) options.
      *
@@ -5619,8 +5623,15 @@ export default (): Record<string, unknown> => {
      *
      * The width is a number of pixels. A string that states a pixel size works too, either as a bare
      * number (`'25'`) or with the unit (`'25px'`), so a value coming from an attribute or a JSON
-     * config still applies. Both forms may be mixed inside the array. A value that is not a pixel
-     * size, such as `'50%'` or `'20em'`, is ignored and the default width is used instead.
+     * config still applies. Both forms may be mixed inside the array.
+     *
+     * A value that is not a pixel size, such as `'50%'` or `'20em'`, is ignored and the default width
+     * is used instead. Inside an array, that applies per level, so one unreadable entry does not
+     * disturb the levels around it.
+     *
+     * A negative number is kept as it is, because numbers keep the behavior they had before this
+     * option read strings at all. A negative string is rejected instead, so a typo cannot collapse
+     * the header.
      *
      * Row headers have a fixed width. A label longer than that width is clipped, and unlike column
      * headers, the header does not grow to fit it. To size the header to its content instead, turn
@@ -5631,7 +5642,7 @@ export default (): Record<string, unknown> => {
      * It has no effect when set in the [`columns`](#columns), [`cells`](#cells), or [`cell`](#cell) options.
      *
      * @memberof Options#
-     * @type {number|number[]}
+     * @type {number|number[]|string|string[]|Array<number|string>}
      * @default undefined
      * @category Core
      *
@@ -5639,6 +5650,9 @@ export default (): Record<string, unknown> => {
      * ```js
      * // set the same width for every row header
      * rowHeaderWidth: 25,
+     *
+     * // set the same width, written as a pixel size
+     * rowHeaderWidth: '25px',
      *
      * // set different widths for individual row headers
      * rowHeaderWidth: [25, 30, 55],
--- a/handsontable/src/helpers/number.ts
+++ b/handsontable/src/helpers/number.ts
@@ -341,8 +341,8 @@ const PIXEL_SIZE_PATTERN = /^\s*(\d+(?:\.\d+)?)\s*(?:px)?\s*$/i;
  * Resolves a pixel size setting written as a number or a string into a number.
  *
  * The grid's size options are documented as numbers of pixels. This helper additionally accepts the
- * two string forms that carry the same unambiguous meaning - a bare numeric string (`'100'`) and a
- * pixel string (`'100px'`) - so a value arriving from an attribute, a JSON config, or a framework
+ * two string forms that carry the same unambiguous meaning – a bare numeric string (`'100'`) and a
+ * pixel string (`'100px'`) – so a value arriving from an attribute, a JSON config, or a framework
  * template still resolves. A number is returned unchanged, including a negative one, so the callers'
  * existing behavior for numbers is untouched.
  *
--- a/handsontable/src/helpers/number.ts
+++ b/handsontable/src/helpers/number.ts
@@ -330,3 +330,39 @@ export function isLossyNumericConversion(rawInput: string, parsedNumber: number)
 export function isUnsignedNumber(value: unknown): value is number {
   return Number.isInteger(value) && typeof value === 'number' && value >= 0;
 }
+
+/**
+ * Matches a size written as a bare number (`'100'`, `'12.5'`) or with a `px` unit (`'100px'`),
+ * ignoring surrounding whitespace and the unit's letter case.
+ */
+const PIXEL_SIZE_PATTERN = /^\s*(\d+(?:\.\d+)?)\s*(?:px)?\s*$/i;
+
+/**
+ * Resolves a pixel size setting written as a number or a string into a number.
+ *
+ * The grid's size options are documented as numbers of pixels. This helper additionally accepts the
+ * two string forms that carry the same unambiguous meaning - a bare numeric string (`'100'`) and a
+ * pixel string (`'100px'`) - so a value arriving from an attribute, a JSON config, or a framework
+ * template still resolves. A number is returned unchanged, including a negative one, so the callers'
+ * existing behavior for numbers is untouched.
+ *
+ * Any value that does not resolve to a pixel count returns `null`. That covers relative units
+ * (`'50%'`, `'20em'`), which depend on a layout context this setting has no access to, and plain
+ * text. Returning `null` lets the caller apply its own default instead of rendering a broken size.
+ *
+ * @param {*} value The value to resolve.
+ * @returns {number|null}
+ */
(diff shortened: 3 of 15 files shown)


qunabu:
Task linked: [DEV-1995 Walkontable single-pass layout: scrollbar prediction + sizing ports](https://app.clickup.com/t/9015210959/DEV-1995)
Task linked: [DEV-2675 Accept pixel-size strings in rowHeaderWidth and columnHeaderHeight (#6154)](https://app.clickup.com/t/9015210959/DEV-2675)

qunabu (dismissed):

handsontable/src/tableView.ts
```diff
@@ -54,6 +56,112 @@ function isUniformSizeSetting(value: unknown): boolean {
...
+ * `JSON.stringify` is deliberately not used: it throws on a `BigInt` and on a circular object, and a
+ * framework can hand either to a setting. This runs inside a Walkontable settings getter during a
+ * draw, so a throw here would take the whole render down – the opposite of the fallback the warning
+ * is reporting. `String()` is total for every value a setting can hold, symbols included.
+ *
+ * @param {*} value The value that could not be read.
+ * @returns {string}
+ */
+function describeValue(value: unknown): string {
+  return typeof value === 'string' ? `"${value}"` : String(value);
```
Haxikowy: `String()` is not total. `String(Object.create(null))` throws `TypeError: Cannot convert object to primitive value`, and so does any object with a throwing `toString`/`valueOf`/`Symbol.toPrimitive` — a framework can hand either to a setting, which is the case the JSDoc above is written for.

This runs inside the Walkontable settings getter during a draw, and only on the bad-value path, so it turns the soft fallback into a dead grid — the exact failure the comment says it avoids.

```suggestion
function describeValue(value: unknown): string {
  if (typeof value === 'string') {
    return `"${value}"`;
  }

  try {
    return String(value);
  } catch {
    return Object.prototype.toString.call(value);
  }
}
```

handsontable/src/dataMap/metaManager/metaSchema.ts
```diff
@@ -5602,10 +5611,16 @@ export default (): Record<string, unknown> => {
...
      * You can set the `rowHeaderWidth` option to one of the following:
      *
-     * | Setting  | Description                                     |
-     * | -------- | ----------------------------------------------- |
-     * | A number | Set the same width for every row header         |
-     * | An array | Set different widths for individual row headers |
+     * | Setting  | Description                                                    |
+     * | -------- | -------------------------------------------------------------- |
+     * | A number | Set the same width for every row header                        |
+     * | A string | Set the same width, written as a pixel size (`'25'`, `'25px'`) |
```
Haxikowy, line 5621: The table says "A string", but `@type` at line 5634 still reads `{number|number[]}` and the `@example` block below has no string form. `columnHeaderHeight` got both (lines 913 and 922).

That leaves the one option issue #6154 is actually about with an API reference page that contradicts its own table and contradicts `core/settings.ts:108`. Anyone arriving from the issue reads that `'80px'` is invalid. Mirror the `columnHeaderHeight` hunk here.

.changelogs/13288.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "private",
```
Haxikowy: This closes public issue #6154, but it is filed as private against the PR number. Every other entry of this shape uses the issue number for both the filename and `issueOrPR` (7561, 11674, 13166 all did it while also carrying a DEV task).

As it stands the release notes link `/pull/13288` instead of the issue people have been subscribed to since 2019. Rename to `.changelogs/6154.json` and:

```suggestion
  "issuesOrigin": "public",
```

plus `"issueOrPR": 6154`.

handsontable/src/tableView.ts
```diff
@@ -54,6 +56,112 @@ function isUniformSizeSetting(value: unknown): boolean {
...
+
+  if (Array.isArray(value)) {
+    const entries: unknown[] = value;
+
+    // Returning the same array keeps the already-numeric case allocation-free on every draw.
+    if (entries.every(isResolvedHeaderSizeEntry)) {
+      return entries;
+    }
+
+    return entries.map(entry => resolveHeaderSizeEntry(entry, scope, optionName));
```
Haxikowy, line 176: An invalid entry that is not the last one does not get the default size the docs and the changelog promise.

`baseTable._modifyRowHeaderWidth` only runs `_correctRowHeaderWidth` on `widths[widths.length - 1]`, so `rowHeaderWidth: ['20em', 100]` resolves to `[null, 100]` and the `null` survives. `columnUtils.calculateWidths` substitutes `defaultColumnWidth` for it, but `workspaceSize.getRowHeaderWidth` does `this.rowHeaderWidth += w ?? NaN`, goes NaN, and falls through to DOM-measuring the whole block. The two consumers then disagree about the same header.

The path is pre-existing — a string produced the same NaN before — but this PR is what starts advertising mixed arrays as supported, which is what puts traffic on it. Either map a rejected entry to the default here, or narrow the "mixed forms" wording in the guides and both JSDoc blocks to the scalar and last-entry cases.

handsontable/src/tableView.ts
```diff
@@ -54,6 +56,112 @@ function isUniformSizeSetting(value: unknown): boolean {
...
+  if (value === null || value === undefined) {
+    return null;
+  }
+
+  const size = parsePixelSize(value);
+
+  if (size === null) {
+    warnOnce(
+      scope,
+      `invalid-header-size-${optionName}`,
```
Haxikowy: The key carries the option name but not the value, and the scope is the user's container element, which outlives `destroy()`.

So `updateSettings({ columnHeaderHeight: '20em' })` after an initial `'50%'` prints nothing, and the console only ever names the first offending value. And a React or Angular component that remounts on the same node warns zero times for the rest of the page.

Folding `describeValue(value)` into the key fixes the first; a per-instance scope object fixes the second.

handsontable/src/helpers/number.ts
```diff
@@ -330,3 +330,39 @@ export function isLossyNumericConversion(rawInput: string, parsedNumber: number)
 export function isUnsignedNumber(value: unknown): value is number {
   return Number.isInteger(value) && typeof value === 'number' && value >= 0;
 }
+
+/**
+ * Matches a size written as a bare number (`'100'`, `'12.5'`) or with a `px` unit (`'100px'`),
+ * ignoring surrounding whitespace and the unit's letter case.
+ */
+const PIXEL_SIZE_PATTERN = /^\s*(\d+(?:\.\d+)?)\s*(?:px)?\s*$/i;
```
Haxikowy: The number form and the string form disagree on sign: `parsePixelSize(-50)` returns `-50`, while `'-50'` is rejected, warned, and defaulted.

Both halves are deliberate (the unit test explains the string side as typo protection), and I would keep them. But the premise of the change is that `'100'` and `100` are the same value, so someone moving a config into JSON or a template gets a silent behavior change on a negative. Nothing user-facing says so — not this JSDoc, not the warning text. Worth one sentence in both.

handsontable/src/helpers/number.ts
```diff
@@ -330,3 +330,39 @@ export function isLossyNumericConversion(rawInput: string, parsedNumber: number)
...
+ * existing behavior for numbers is untouched.
+ *
+ * Any value that does not resolve to a pixel count returns `null`. That covers relative units
+ * (`'50%'`, `'20em'`), which depend on a layout context this setting has no access to, and plain
+ * text. Returning `null` lets the caller apply its own default instead of rendering a broken size.
+ *
+ * @param {*} value The value to resolve.
+ * @returns {number|null}
+ */
+export function parsePixelSize(value: unknown): number | null {
```
Haxikowy: This ships as public API. `src/index.ts` spreads the whole `helpers/number` module into `HOT.helper` and `base.ts:114` types it as `typeof import('./helpers/number')`, so `Handsontable.helper.parsePixelSize` lands in the emitted `.d.ts` and is a permanent commitment under the no-removal rule.

The only callers are in `tableView.ts`. Either take that on deliberately and say so, or move it somewhere off the helper surface.

tests/e2e/header-size-string-values.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * Issue #6154. `rowHeaderWidth` and `columnHeaderHeight` are documented as pixel numbers, but a
+ * value can easily arrive as a string - from an attribute, a JSON config, or a framework template.
+ * Both now resolve `'100'` and `'100px'`, and both ignore a value that states no pixel count.
+ *
+ * All of this is geometry, so none of it can be checked in jsdom, where every size reads as zero.
+ *
+ * The two options used to disagree for reasons that were pure accident. `rowHeaderWidth` ran
+ * through a `typeof width === 'number'` guard that replaced any string with the default column
+ * width, so a string never applied at all. `columnHeaderHeight` had no guard and reached a
```
Haxikowy: The mechanism here is `develop`'s, not the released one, and it is written as history the next reader will trust.

In 18.0.0 the option never reached a `Math.max`. `table.ts` stored the raw setting into `oversizedColumnHeaders[level]` behind an `!isNaN` check and then wrote `` style.height = `${value}px` ``. `'100'` worked because `'100' + 'px'` is valid CSS; `'100px'` failed because `'100pxpx'` is not. The `Math.max(height, providedHeight)` shape is the current `columnUtils.getHeaderHeight`.

The outcomes in your table are all correct, and the `rowHeaderWidth` half of the story checks out — the `typeof width === 'number'` guard is in 18.0.0, so a string genuinely never applied. Only the coercion story needs restating.

tests/e2e/header-size-string-values.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+      .toBeCloseTo(HeaderSizeStringValuesPage.SIZE, 0);
+  });
+
+  test('applies a row header width written with the px unit', async () => {
+    expect(await grid.rowHeaderWidth('row-px'))
+      .toBeCloseTo(HeaderSizeStringValuesPage.SIZE, 0);
+  });
+
+  test('applies a per-level row header width given as strings', async () => {
+    expect(await grid.rowHeaderWidth('row-array'))
```
Haxikowy, line 43: Both array grids are homogeneous: `['100px']` and `['100']`. The guides and both JSDoc blocks promise "you can mix both forms inside the array", and the only thing asserting that is `settings.types.ts`, which is compile-time and never touches `resolveHeaderSizeSetting`'s `.map` path.

One grid with `[100, '100px']` would cover the documented claim, and a second with `['20em', 100]` would pin the non-last-entry behavior I flagged in `tableView.ts`.

Haxikowy (commented):
Reviewed for correctness, regression/compat, performance, API+types, requirements, and the rendering path. The diagnosis holds up: the numeric path is untouched, and the settings funnel is the right place to resolve. Findings are inline; what has no inline home is below.

**Performance: nothing to fix, and the red CI number is not yours.** Measured against real call counts (9 `rowHeaderWidth` reads per draw, 4 per header level for `columnHeaderHeight`, zero on scroll). Worst case, 6 levels with a string array: ~7.6 us per draw, 0.2% of a 3.8 ms draw. In-browser over 300 renders it sat inside the baseline's own drift. Skipping the cache is correct on the numbers.

The `Performance / run` comment (+17% to +44%) is stale-golden noise: no scenario sets either option, scroll draws read them zero times, and absolute totals are flat against #13286 and #13285. Don't chase it.

**Stale claim in the description.** "5 of 7 tests fail when reverted" — the spec has 10, and cdc80cf03 added three after that measurement, including both warning tests. Re-run against all 10 or name which three postdate it.

**Untested, no inline anchor:** `updateSettings({ rowHeaderWidth: '120' })` after init, which is the likeliest real-world path now that resolution lives in the getter. Same for the numeric-array by-reference branch (`[100, 30]`).

**Two pre-existing bugs worth their own tickets.** Both arrived with 294dfcae23, both unreleased, both bear on the loop you are closing:

1. The merge drops a `modifyColumnHeaderHeight` hook that returns an array, which `NestedHeaders` does (`nestedHeaders.ts:1545`). `levels` never counts `hookHeight.length` and `typeof hookHeight === 'number'` is false for an array, so `base` stays undefined and the result is probe-only. The plugin's heights apply for one frame, then get dropped.
2. Root cause under both: `adjustColumnHeaderHeights` writes `style.height`, `RenderSizeProbe` reads it back with `innerHeight`, which excludes the border. An applied H measures back as H minus border every draw, with no floor. A numeric option pins it via `Math.max`, which is why only the string case shrank. `#measureRows` already handles this with `borderBoxSizing ? outerHeight : innerHeight`.

**What I ran:** the regex against 18 edge inputs in node; `String(Object.create(null))` (throws); the 18.0.0 source for `_correctRowHeaderWidth` and the `columnHeaderHeight` path in `table.ts`; the `.changelogs/` convention across all entries. I did not run the Playwright spec or a browser build.

**Verdict:** approve after the `describeValue` crash, the `rowHeaderWidth` JSDoc, and the changelog origin. The rest is your call.

handsontable/src/tableView.ts
budnix: Nice catch — you're right, `String()` throws on a null-prototype object. Fixed in b91787ce6a with the try/catch you suggested.

handsontable/src/dataMap/metaManager/metaSchema.ts
budnix: Good spot — I updated the wrong one's tags. `rowHeaderWidth` now has the same `@type` and a string example in b91787ce6a.

.changelogs/13288.json
budnix: Agreed — 7561 and 11674 make the convention clear. Renamed to `.changelogs/6154.json` and switched to `public` in b91787ce6a.

handsontable/src/tableView.ts
budnix: Fair point to check, but I measured it: with `['20em', 90]` the rejected level still renders at the default 50 and the other at 90, because `ColumnUtils.calculateWidths` fills the `col` before `getRowHeaderWidth` measures the DOM. Pinned as a test in b91787ce6a.

handsontable/src/tableView.ts
budnix: You're right on both — the value is part of the key now, and the scope is the `TableView` instead of the container, so a remount warns again. Fixed in b91787ce6a.

handsontable/src/helpers/number.ts
budnix: Good point — the asymmetry was nowhere in the docs. Now stated in the JSDoc, both option blocks, both guides and the warning text itself, in b91787ce6a.

handsontable/src/helpers/number.ts
budnix: Yep, my mistake — that would have been a permanent commitment. Moved to `src/utils/pixelSize.ts` in b91787ce6a, with its tests.

tests/e2e/header-size-string-values.spec.ts
budnix: You're right that it described `develop` — rewrote it in b91787ce6a. One correction though: 18.0.0 does have the `Math.max`, in `columnUtils.getHeaderHeight`; what drops `'100px'` is the `!isNaN(setting)` gate in `table.ts` before it ever reaches the style write.

tests/e2e/header-size-string-values.spec.ts
budnix: Good catch — added two two-level grids in b91787ce6a: `[60, '90px']` for the mixed case and `['20em', 90]` for the non-last entry.

Haxikowy (approved):

PR #13290, opened 2026-08-28
DEV-2684: Document that className applies to the container and the cells

### LINKED ISSUE #8479: className is added to td for registered renderers?
### Description

Not sure if this is 100% a bug, would like to hear how the HOT team would think about this.
According to the [document](https://handsontable.com/docs/6.2.1/Options.html#className), className should be set to the container node. 
However, I find that if you have some renderers set up, you get classname added to the TDs generated by some of these renders.

If you do not have any renderers, the classname is not added to TD cell, though...

### Steps to reproduce

1. run https://jsfiddle.net/semix/sLgufynm/2/ and inspect the cells for the first two columns, you see the classname 'aaaa' is added. However, for the rest two columns, 'aaaa' is not there.
2. 
3.

### Demo

https://jsfiddle.net/semix/sLgufynm/2/  (classname is written to TD)

### Your environment
* Handsontable version: latest
* Browser Name and version: chrome
* Operating System: macos

budnix:
### Context

The PR fixes the documentation gaps behind #8479, and adds the regression tests that were missing.

Issue #8479 reports that `className` set at the grid level also lands on every `TD`, while the documentation described it as a container-only option. The issue turned out to be three separate threads:

1. **A custom renderer dropped the cell classes.** This was a real bug, and it is already fixed. PR #11997 (commit 118f537d1e, released in 17.0.0) added the `_isBaseRendererCalled` flag and a fallback call, now in `handsontable/src/renderers/renderCell.ts`. Verified against released builds with a custom renderer that chains no built-in renderer:

   | Version | Root `DIV` | Default-renderer `TD` | Custom-renderer `TD` |
   | --- | --- | --- | --- |
   | 16.0.0 | `aaaa` | `aaaa` | `class=""` -- class lost |
   | 17.0.0 | `aaaa` | `aaaa` | `aaaa` |
   | 18.0.0 | `aaaa` | `aaaa` | `aaaa` |

2. **Grid-level `className` goes on the container element and on every cell.** This is the cascading configuration model working as designed. Changing it would be a breaking change.

3. **A `cells()` function returning `className` replaces the value from a higher level.** Also by design -- cascading configuration overwrites, it does not merge.

So no source behavior needs to change. What was missing was documentation:

- The `className` JSDoc described cells only. It never stated that at the grid level the class names also land on the container element, which is the reporter's original question.
- The 17.0.0 automatic base renderer was documented nowhere. The changelog entry for #11997 mentions only `Intl.NumberFormat` support, which is why nobody connected that PR to this issue.
- `sparkline-cell-renderer.md` still told readers to "always call `baseRenderer`". Since 17.0.0 Handsontable calls it for you.
- `cell-renderer.md` used "base renderer" to mean "the built-in renderer you extend", which reads as a contradiction next to the `baseRenderer` export.

### What changed

- `handsontable/src/dataMap/metaManager/metaSchema.ts` -- the `className` JSDoc now shows which levels target the container element and which target cells, documents that a lower level replaces a higher level rather than merging, gives the container-only recipe, and states the custom-renderer behavior.
- `docs/content/guides/cell-functions/cell-renderer/cell-renderer.md` -- disambiguates "built-in renderer" from `baseRenderer`, and adds a section explaining that Handsontable runs `baseRenderer` for you, including when you would still call it yourself.
- `docs/content/recipes/rendering-styling/sparkline-cell-renderer/sparkline-cell-renderer.md` -- replaces the outdated "always call `baseRenderer`" instruction. The explicit call is kept in the example, because calling it first is what gives you control over ordering.
- `docs/content/guides/upgrade-and-migration/migrating-from-16.2-to-17.0/migrating-from-16.2-to-17.0.md` -- adds the missing migration step for the 17.0.0 renderer behavior change, with before/after examples, plus a row in the summary table.
- `handsontable/src/__tests__/settings/className.spec.js` -- two new cases.

### How has this been tested?

The existing spec covered `rootElement` and `getCellMeta().className`, but never the `TD` itself with a custom renderer. That path was pinned only at the unit level in `renderCell.unit.ts`. Two cases were added:

- a cell whose renderer calls no built-in renderer still receives `className`;
- clearing `className` through `cells()` leaves the container element's class intact.

Both new assertions were verified to fail when inverted, so they exercise the behavior rather than passing vacuously.

```bash
npm run test:e2e --prefix handsontable -- --testPathPattern=settings/className   # 10 specs, 0 failures
npm run test:unit --prefix handsontable -- --testPathPattern=renderCell          # 9 passed
npm run eslint --prefix handsontable                                             # 0 errors
```

The version table above was measured in Chrome against the released 16.0.0, 17.0.0, and 18.0.0 builds from the CDN.

### Types of changes

### Related issue(s):

1. DEV-2684
2. Closes #8479
3. Context: #11997 is the PR that fixed the renderer half in 17.0.0

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

[skip changelog] -- this PR changes documentation and tests only. No source behavior changes.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2684

---

> [!NOTE]
> **Low Risk**
> Documentation and regression tests only; no changes to rendering or `className` application logic.
> 
> **Overview**
> Closes documentation gaps around **`className`** and the **17.0 automatic `baseRenderer`** (no runtime changes in this PR).
> 
> **`className`** is now documented as applying to the **grid container and cells** when set at grid level, with a table of levels, the container-only recipe (`cells()` returning `className: ''`), replace-not-merge cascading, and that custom renderers still get classes via the post-render `baseRenderer` fallback. Updates span API JSDoc in `metaSchema.ts`, the formatting-cells guide, and related links.
> 
> **Renderer docs** distinguish extending **`textRenderer`/`htmlRenderer`** from the **`baseRenderer`** export, explain that since **17.0** Handsontable runs `baseRenderer` after custom renderers when they did not, and when you should still call it first. The **16.2→17.0 migration guide** adds section **8** plus a summary-table row; the sparkline recipe drops “always call `baseRenderer`” in favor of that model.
> 
> **Tests:** two e2e cases in `className.spec.js` — `className` on cells with a renderer that chains no built-in renderer, and clearing cell `className` via `cells()` while the root container keeps the class.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 579824b47c0e8e8be901736014809aae2e34bda9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/content/guides/upgrade-and-migration/migrating-from-16.2-to-17.0/migrating-from-16.2-to-17.0.md
+++ b/docs/content/guides/upgrade-and-migration/migrating-from-16.2-to-17.0/migrating-from-16.2-to-17.0.md
@@ -1154,6 +1154,64 @@ The **Formulas** plugin uses [HyperFormula](https://hyperformula.handsontable.co
 
 See the [Formulas installation](@/guides/formulas/installation/installation.md) guide to install and license HyperFormula, and the [Formula calculation](@/guides/formulas/formula-calculation/formula-calculation.md) guide for configuration details.
 
+## 8. Custom renderers always receive the base cell classes
+
+Starting in **version 17.0**, Handsontable runs `baseRenderer` for every cell, even when the cell's renderer never calls it. `baseRenderer` adds the cell's class names and ARIA attributes.
+
+### What changed
+
+- **Automatic base renderer**: Handsontable runs `baseRenderer` after your custom renderer, whenever your renderer didn't run it
+- **Consistent cell classes**: Cells drawn by a custom renderer now receive [`className`](@/api/options.md#classname), the read-only class, the invalid-cell class, and the matching ARIA attributes
+- **Built-in renderers**: The built-in renderers no longer call `baseRenderer` themselves, because Handsontable calls it for them
+
+### Why this change
+
+Before version 17.0, a custom renderer that called no built-in renderer produced cells without those class names. One grid could show one column with the classes and another column without them, depending on how each column's renderer was written.
+
+### How to migrate
+
+Most applications need no action. Your custom renderers keep working, and their cells gain the class names they were missing.
+
+Two cases need attention:
+
+1. **You styled cells based on a missing class.** Cells drawn by a custom renderer now carry `className` and the other base classes. Review CSS that relied on those classes being absent.
+2. **Your renderer sets a class that `baseRenderer` also manages.** `baseRenderer` runs last, and it removes the invalid-cell class from valid cells. Call `baseRenderer` at the start of your renderer to keep control of the order.
+
+Before:
+
+```js
+// Handsontable 16.2 - this cell receives no `className` from the grid settings
+function myRenderer(instance, td, row, col, prop, value, cellProperties) {
+  td.textContent = value;
+}
+```
+
+After:
+
+```js
+// Handsontable 17.0 - this cell receives `className` automatically.
+// Call `baseRenderer` first only when you need it to run before your changes.
+function myRenderer(instance, td, row, col, prop, value, cellProperties) {
+  baseRenderer(instance, td, row, col, prop, value, cellProperties);
+  td.textContent = value;
+}
+```
+
+### What to expect
+
+- **Consistent classes**: Every cell receives the base class names, whichever renderer draws it
+- **No duplicate work**: Handsontable skips its own call when your renderer already ran `baseRenderer`
+- **Same API**: Renderer signatures are unchanged
+
+### Timeline
+
+- **Version 17.0**: Handsontable runs `baseRenderer` for every cell
+
+### Related resources
+
+- [Cell renderer](@/guides/cell-functions/cell-renderer/cell-renderer.md) - Write and extend cell renderers
+- [`className`](@/api/options.md#classname) - Add CSS class names to cells and to the container element
+
 ## Summary of breaking changes
 
 | Change                                            | Action Required                                                            |
@@ -1163,6 +1221,7 @@ See the [Formulas installation](@/guides/formulas/installation/installation.md) 
 | CSS-based themes (optional migration)             | Consider migrating to Theme API for runtime features                       |
 | `core-js` dependency removed                      | Add `core-js` or other polyfills in your app if you support older environments |
 | Built-in HyperFormula (deprecation)               | In 18.0, import HyperFormula yourself and pass it to the Formulas plugin with `licenseKey: 'internal-use-in-handsontable'` |
+| Custom renderers receive the base cell classes    | Review CSS that relied on those classes being absent from custom-rendered cells |
 
 ## Result
 
--- a/docs/content/guides/upgrade-and-migration/migrating-from-16.2-to-17.0/migrating-from-16.2-to-17.0.md
+++ b/docs/content/guides/upgrade-and-migration/migrating-from-16.2-to-17.0/migrating-from-16.2-to-17.0.md
@@ -1162,7 +1162,7 @@ Starting in **version 17.0**, Handsontable runs `baseRenderer` for every cell, e
 
 - **Automatic base renderer**: Handsontable runs `baseRenderer` after your custom renderer, whenever your renderer didn't run it
 - **Consistent cell classes**: Cells drawn by a custom renderer now receive [`className`](@/api/options.md#classname), the read-only class, the invalid-cell class, and the matching ARIA attributes
-- **Built-in renderers**: The built-in renderers no longer call `baseRenderer` themselves, because Handsontable calls it for them
+- **Built-in renderers**: Most built-in renderers no longer call `baseRenderer` themselves, because Handsontable calls it for them. `multiSelectRenderer` is the exception, and still calls it as its first statement
 
 ### Why this change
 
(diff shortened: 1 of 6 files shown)


qunabu:
Task linked: [DEV-2684 Document that `className` applies to both the container and the cells (#8479)](https://app.clickup.com/t/9015210959/DEV-2684)

docs/content/guides/upgrade-and-migration/migrating-from-16.2-to-17.0/migrating-from-16.2-to-17.0.md
```diff
@@ -1154,6 +1154,64 @@ The **Formulas** plugin uses [HyperFormula](https://hyperformula.handsontable.co
...
 
+## 8. Custom renderers always receive the base cell classes
+
+Starting in **version 17.0**, Handsontable runs `baseRenderer` for every cell, even when the cell's renderer never calls it. `baseRenderer` adds the cell's class names and ARIA attributes.
+
+### What changed
+
+- **Automatic base renderer**: Handsontable runs `baseRenderer` after your custom renderer, whenever your renderer didn't run it
+- **Consistent cell classes**: Cells drawn by a custom renderer now receive [`className`](@/api/options.md#classname), the read-only class, the invalid-cell class, and the matching ARIA attributes
+- **Built-in renderers**: The built-in renderers no longer call `baseRenderer` themselves, because Handsontable calls it for them
```
demtario: **medium** The blanket claim isn't accurate: `multiSelectRenderer` still calls `baseRenderer` itself, as its first statement, and it did so at 17.0.0 too (`baseRenderer.apply(...)` in the released file). See [`multiSelectRenderer.ts:32`](https://github.com/handsontable/handsontable/blob/59c1d6ff925231a16179de559ec58add8feefd7e/handsontable/src/renderers/multiSelectRenderer/multiSelectRenderer.ts#L32).

The consequence lands on line 1178 as well. Because that renderer runs `baseRenderer` first, item 2 ("`baseRenderer` runs last [...] Call `baseRenderer` at the start of your renderer to keep control of the order") describes an ordering that doesn't hold on that path, and prescribes something already done.

Suggested fix: scope the bullet ("Most built-in renderers no longer call `baseRenderer` themselves") or name the set, and add the exception where line 1178 states the ordering.

Note the same absolute wording is pre-existing in [`renderCell.ts:43`](https://github.com/handsontable/handsontable/blob/59c1d6ff925231a16179de559ec58add8feefd7e/handsontable/src/renderers/renderCell.ts#L43); not this PR's to fix, but it's where the phrasing comes from.

docs/content/guides/upgrade-and-migration/migrating-from-16.2-to-17.0/migrating-from-16.2-to-17.0.md
```diff
@@ -1154,6 +1154,64 @@ The **Formulas** plugin uses [HyperFormula](https://hyperformula.handsontable.co
...
+}
+```
+
+After:
+
+```js
+// Handsontable 17.0 - this cell receives `className` automatically.
+// Call `baseRenderer` first only when you need it to run before your changes.
+function myRenderer(instance, td, row, col, prop, value, cellProperties) {
+  baseRenderer(instance, td, row, col, prop, value, cellProperties);
```
demtario: **medium** The snippet calls the bare identifier `baseRenderer` with no import, so it isn't runnable as pasted. Every other `js` block in this guide shows its imports (lines 63 and 276), and the repo's own example uses `import { registerRenderer, baseRenderer } from 'handsontable/renderers';` ([`sparkline-cell-renderer/javascript/example1.js:3`](https://github.com/handsontable/handsontable/blob/59c1d6ff925231a16179de559ec58add8feefd7e/docs/content/recipes/rendering-styling/sparkline-cell-renderer/javascript/example1.js#L3)).

Secondary, same block: line 1173 says "Most applications need no action", but a Before/After pair whose "After" adds a call reads as a required migration step, and the in-snippet comment doesn't fully undo that framing. Consider relabeling the pair as the opt-in ordering case rather than Before/After.

docs/content/recipes/rendering-styling/sparkline-cell-renderer/sparkline-cell-renderer.md
```diff
@@ -76,7 +76,9 @@ This recipe shows how to edit weekly values in table cells and render a mini bar
 
 ## Step 1: Call the base renderer first
```
demtario: **low** The heading now contradicts its own body: "Call the base renderer first" against line 81, "You don't have to call `baseRenderer` yourself." It also uses "the base renderer" in the loose sense that `cell-renderer.md` disambiguates in this same PR.

Suggested fix: retitle descriptively, e.g. "Step 1: Run `baseRenderer` before the SVG".

handsontable/src/dataMap/metaManager/metaSchema.ts
```diff
@@ -842,6 +842,44 @@ export default (): Record<string, unknown> => {
...
+     * The target depends on the level at which you set the option:
+     *
+     * | Level                                | Container element | Cells                |
+     * | ------------------------------------ | ----------------- | -------------------- |
+     * | Grid                                 | Yes               | Every cell           |
+     * | [`columns`](#columns)                | No                | Cells of that column |
+     * | [`cells`](#cells) or [`cell`](#cell) | No                | The matching cells   |
+     *
+     * At the grid level, Handsontable adds the class names to two places. It adds them to the
+     * container element – the element you pass to the constructor – and, through
```
demtario: **low** "the element you pass to the constructor" doesn't hold for wrapper users, who never pass one: the React, Angular, and Vue wrappers create the container themselves. This JSDoc feeds the framework-shared API reference, so the sentence is read by those users too. The target is `instance.rootElement` however it was created.

Suggested fix: "the container element that holds the grid".

demtario (commented):
4 inline notes below.

**Off-diff**

- `docs/content/guides/cell-features/formatting-cells/formatting-cells.md:41` still carries the claim this PR exists to correct: it says `className` set "in the grid configuration" is "added to the cell's `<td>` element", with no mention of the container element. Worth updating in the same PR, or linking to the new `className` API entry. I swept the rest of `docs/content`: the other `<td>`-only statements (`conditional-row-coloring.md:108,215`, `custom-context-menu.md:127`) describe `cells`-level `className` and are accurate, so this is the only page affected.

docs/content/guides/upgrade-and-migration/migrating-from-16.2-to-17.0/migrating-from-16.2-to-17.0.md
budnix: You're right — multiSelectRenderer is the one exception. Scoped the bullet and named it in 579824b47c, and case 2 now says Handsontable's call runs after your renderer rather than always last.

docs/content/guides/upgrade-and-migration/migrating-from-16.2-to-17.0/migrating-from-16.2-to-17.0.md
budnix: Good point — added the import and flipped the pair in 579824b47c, since the real change is that the call became unnecessary, not required.

docs/content/recipes/rendering-styling/sparkline-cell-renderer/sparkline-cell-renderer.md
budnix: Nice catch — retitled to your suggestion in 579824b47c. Nothing linked to the old anchor.

handsontable/src/dataMap/metaManager/metaSchema.ts
budnix: Good spot — reworded to "the container element that holds the grid" in 579824b47c. Also fixed the off-diff formatting-cells.md:41 you flagged.

demtario (approved):

PR #13250, opened 2026-08-26
RELEASE-843: Remove 18.0.0 entries duplicated into the 18.1.0 changelog

demtario:
### Context

The PR removes four entries from the `18.1.0` section of `CHANGELOG.md` whose code shipped in `18.0.0`, and fixes two entries that cannot be quoted as they stand. This is a prerequisite for the `18.1.0` release notes page (RELEASE-843), which copies its entry list from this section.

**The four duplicated entries.** Each is a verbatim copy of an entry already published in the `18.0.0` section:

| Cited | Entry |
|---|---|
| #12847 | Reduced memory usage and improved initialization time... no longer materializing cell metadata |
| #12727 | **Breaking change**: Removed the deprecated `PersistentState` plugin |
| #12811 | Fixed cell meta set with `setCellMeta`... being reset by `updateSettings` |
| #12844 | Reduced memory usage during scrolling and fixed potential out-of-memory errors |

Two independent checks each found exactly these four, and no others:

1. **Commit ancestry.** The `18.0.0` tag is a merge commit (`892378293`) whose parents are develop (`8d5c5329b`) and the release branch (`9b82d4be6`), so the tag is not a usable oracle - it counts develop commits that never shipped. Tested against `9b82d4be6`, the `18.0.0` version-bump commit on `origin/release/18.0.0`: 4 of the 105 cited numbers resolve to a commit that shipped in `18.0.0`.
2. **Verbatim title comparison** between the two sections, links stripped: the same 4. Also compared against every section from `17.1.0` down: no overlap.

Two further checks bound the result rather than re-count it:

3. **`.changelogs` file lifecycle**, which separates the two root causes rather than corroborating the count. For every JSON consumed by `8f12dfcd2` (`18.1.0-rc1`) and `5aa1e2125` (rc2), the commit that added the file was tested against `9b82d4be6`. It returns **only #12727**, and that is the expected result: the other three had their JSON re-added on develop after the release, so the file itself is not a pre-18.0 artifact even though the code is. The single hit is what tells the two causes apart.
[...]

PR #13243, opened 2026-08-25
DEV-2627: Treat layout-slot UI as grid-internal in foreign-focus detection

adrianspdev:
### Context

The PR fixes a DEV-1619 (#13194) regression in the foreign-focus detection. `FocusGridManager.isForeignFocusTarget()` recognized an element as grid-internal only when it sits inside `rootElement` (or the portal). Plugin UI mounted through the LayoutManager edge slots lives inside `rootWrapperElement` but outside `rootElement`, so its focusable elements were classified as foreign. On every real click, the `mouseup` outside-click path in `tableView` then closed the open editor, deselected the grid, and called `hot.unlisten()`.

Any focusable layout-slot UI is affected — Pagination's controls today, and the FormulaBuilder plugin's formula bar (`feature/HOT-15500-formula-builder-plugin`), where clicking the bar starts an edit that dies within the same click. The originating commit already made `tableView`'s `isPathThroughGridUi` and `FocusGridManager.hasBrowserFocus` wrapper-aware; `isForeignFocusTarget` was the one check that missed the wrapper.

The fix treats elements inside `rootWrapperElement` but outside `rootElement` as grid-internal. Elements inside `rootElement` keep the existing `isInternalElement` semantics, so a nested grid's focus still counts as foreign for the parent instance.

### How has this been tested?

- New unit suite `handsontable/src/focusManager/__tests__/grid.unit.ts`: layout-slot UI counts as internal, outside elements stay foreign, nested-grid focus stays foreign, body/null are never foreign.
  - `npm --prefix handsontable run test:unit -- --testPathPattern=focusManager`
- Full unit suite: 3737 passed.
- Focus-related legacy E2E (181 specs: focus, editorManager, unlisten, outside click) and the selection-interaction suite, green against a fresh build:
  - `npm --prefix handsontable run test:e2e -- --testPathPattern='focus|editorManager|Core_unlisten|outsideClick'`
- Manually verified on the FormulaBuilder feature branch that clicking the formula bar opens and keeps the bar edit (the discovery case); a Playwright test covering that flow lives on the feature branch, since the plugin is not on `develop`.

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature or improvement (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Additional language file or change to the existing one (translations)

### Related issue(s):

1. DEV-2627

### Affected project(s):

- [x] `handsontable`
[...]

PR #12844, opened 2026-06-24
DEV-1940: Render Walkontable directly to DOM, remove diffing renderer

budnix:
### Context

The PR changes the Walkontable rendering engine to render directly to the DOM and removes the diffing renderer strategy together with the coordinate-keyed node pool.

`NodesPool` cached one DOM node per source coordinate ever rendered and never evicted it, so scrolling grew retained (detached) DOM monotonically — O(rows × cols). On large datasets this produced large heap growth and out-of-memory crashes. The diffing renderer (`ViewDiffer` plus the renderer-adapter strategy) added substantial complexity for no measurable rendering benefit on modern browsers — the pool-free "direct DOM" renderer already shipped as the Firefox path.

This change makes the direct-DOM renderer the single path: the root node keeps exactly `viewSize` children and each render cycle reuses those children in place (their content is overwritten by the cell renderer). Memory is therefore bounded to the viewport regardless of dataset size.

Specifically:
- Inlined the direct-DOM rendering into `OrderView`; removed the renderer-adapter indirection.
- Deleted `ViewDiffer` / `ViewOrder`, the `rendererAdapter/` directory (factory + differ-based + direct-dom adapters), and the `CMD_*` constants.
- Reduced `NodesPool` to a thin element factory (no cache); removed the temporary pool-pruning hook.
- Fixed multi-level row-header ordering — the previous reverse loop was a compensation for the differ's prepend behavior and produced reversed headers under direct rendering.

### Performance results (100k rows × 100 cols, Chromium)

Memory, after a full top→bottom scroll:

| | Before | After |
|---|--:|--:|
| Retained (detached) DOM nodes | ~2,400,000 | ~600 (viewport) |
| JS heap | **~915 MB (page OOM-stalls)** | **~520–590 MB (stable)** |
| Behavior at ~1M rows | out-of-memory crash | renders; memory O(viewport) |

Render speed (per-draw, median / p95) and frame rate:

| Scenario | Before | After |
|---|--:|--:|
| Fast scroll | 8.2 / 8.9 ms | 7.3 / 7.9 ms |
| Slow scroll | ~5.1 / 5.9 ms | ~5.2 / 6 ms |
| Frame rate | ~90 FPS | ~90–99 FPS |

Render speed is equal-or-faster across 11 scenarios, including 1M rows, 10,000 columns, pinned rows/columns, RTL, merged cells, nested headers, and variable row heights.

### How has this been tested?

[...]

PR #13291, opened 2026-08-28
DEV-2678: Stop already-released changelog entries from being republished

demtario:
### Context

The PR stops an already-released changelog entry from being announced a second time in the next release.

A pending `.changelogs/<n>.json` can outlive the release-to-develop merge-back after the release branch has already consumed it. The next `bin/changelog consume` then republishes it. #13250 removed four such entries from the `18.1.0` section by hand; this PR fixes the cause.

There are two mechanisms, both reproduced in an isolated synthetic repository:

- **The same change committed twice, with no ancestry**, once on the release branch, once on `develop`. Against the merge base the release side reads as *add then delete*, net zero, so develop's own *add* is the only change on either side and the three-way merge keeps it. **No conflict is raised at all**, which is why this was never noticed. `#12811` is `6714c8743d` on `release/18.0.0` and `6bc43a7a7a` on `develop`; `#12844` and `#12847` have the same shape.
- **The entry file edited on `develop` after being consumed**, a modify/delete conflict, resolved by keeping develop's copy. `#12727` was consumed by `18.0.0-rc1` (`fd8cddd0bc`) and then modified six days later by `d68ea7f92e` (#12810).

Running a detector against the pending set at the parent of every release `consume` commit shows the leak was primed at **9 of the last 13 releases**, and only the 18.1.0 round was ever caught.

Nothing existing catches it: `changelog-gate.mjs` asserts a source PR *adds* an entry and never checks for republication, the merge-back is a direct push rather than a pull request so no PR check runs on it, and `consumeCommand` validated entry *format* only.

### What this does

Rather than trying to recognize either merge shape, it asserts the one thing both produce: a pending entry that `CHANGELOG.md` already publishes. `consumeCommand` already holds the changelog text and the parsed entries, so the check is local, with no git archaeology and no new workflow.

**It is placed before the `if (dryRun)` early exit on purpose.** `checks.yml:255` already runs `bin/changelog consume --date 2050-01-01 --dry-run` on every pull request, so one assertion is enforced at three points for zero CI wiring: every PR against `develop`, the release dry run, and the real consume.

It is also applied to `syncCommand`, which deduped only against the section it targets (`section.includes('/' + issueOrPR + ')')`). An entry already published in an *earlier* section would be added again, and `publish.yml:796` runs `sync` on every RC, so without this the gate was bypassable by the release path itself. Trade-off accepted: `sync` can now exit non-zero mid-RC.

#### Matching rules

| Match | Result |
|---|---|
| `issueOrPR` already cited, `issuesOrigin` is `private` | **fails** |
| `issueOrPR` already cited, `issuesOrigin` is `public` | warns |
| the entry's title already published, under any number | warns |

A pull request number cannot ship twice, so a `private` number match is conclusive. The other two can be legitimate: a public **issue** number may be cited by two releases when a partial fix is followed by a complete one, and one title may appear in two sections when a fix is backported to several release lines (`Fixed the copy/paste feature not working correctly in Chrome 133.` legitimately appears in 15.1.0, 15.0.1, and 14.6.2).

Both keys are required, because each alone has a blind spot that already occurred: the number key misses an entry published under a wrong link (`#12727` shipped as `[#0]`), and the title key misses an entry reworded on one branch (`#13243`).

**Known gap, accepted:** an entry that carries `issuesOrigin: "public"` for a *pull request* number is mislabeled, and a genuine leak on it downgrades to a warning rather than failing. That is live behavior, not something the merge-back removed. `.changelogs/12811.json` was such an entry: `#12844` and `#12847` hard-failed in the same round, so the release was blocked anyway and `#12811` appeared as a warning in the same report. A possible refinement, not taken here: hard-fail a `public` number hit when the title also matches, since two independent keys agreeing is inconsistent with both legitimate cases.

### Merge order, now satisfied

The agreed sequence was: release 18.1.0 on the current flow, merge `release/18.1.0` back into `develop`, then rebase this branch and merge it. Steps 1 and 2 are done (#13324), and this branch is now rebased on top of that merge-back.

The merge-back removed **every** consumed entry, including the four 18.0-era stragglers (`12727`, `12811`, `12844`, `12847`) that an earlier revision of this branch deleted by hand. Those deletions are therefore gone from the diff, and no replacement deletions are needed: the recompute below reports a clean pending set. This PR is now the gate and its documentation only.

**Target is `develop` only, no cherry-pick to `release/18.1.0`.** That branch holds zero pending `.changelogs/*.json` entries, so the gate has nothing to act on there, and `package.json` diverges, so the eslint-script line would conflict. Future release branches inherit the gate when they are cut from `develop`.

### How has this been tested?

No `handsontable/src/**` or `wrappers/**` code changed, so the core lint, build, and test suites are unaffected. The relevant gates, re-run after the rebase:

```bash
npm run test:tooling
npx eslint .github/scripts/ bin/changelog bin/lib scripts/
```

Result: **216/216 tests pass** (16 new), eslint clean.

Both `bin/changelog` commands were smoke-tested on this branch, and neither modifies `CHANGELOG.md`:

```bash
node bin/changelog consume --date 2050-01-01 --dry-run   # exit 0
node bin/changelog sync --dry-run                        # exit 0
```

#### Backtests against real history

```bash
# 1. the state at the 18.1.0-rc1 consume, 104 pending entries
git checkout 8f12dfcd24^ -- CHANGELOG.md .changelogs
node bin/changelog consume --date 2050-01-01 --dry-run
git checkout HEAD -- CHANGELOG.md .changelogs
```

Flags exactly the four entries #13250 removed by hand, and nothing else. 100 clean, no false positives. `#12727` is caught by the **title** key alone, which is what earns the second key.

```bash
# 2. real post-merge-back develop, this branch rebased, 39 pending entries
node bin/changelog consume --date 2050-01-01 --dry-run   # exit 0
```

0 errors, 0 warnings. The pre-rebase revision predicted `13236`, `13240`, `13242`, `13243`, `13245` would survive the merge-back. The merge-back deleted all five, so nothing is left to remove.

```bash
# 3. the same tree with one leaked entry restored, as a live-fire check
git show d8df372278:.changelogs/13236.json > .changelogs/13236.json
node bin/changelog consume --date 2050-01-01 --dry-run   # exit 1
```

```
1 pending changelog entry has already been published:
  .changelogs/13236.json: #13236 is already cited in [18.1.0], and its title is already published in [18.1.0]
  git rm .changelogs/13236.json
```

Both keys fire on real data, and the remediation command is printed.

#### Mutation tested

The new tests were confirmed to fail against deliberately broken implementations, so they are not vacuously green:

| Mutation | Tests that fail |
|---|---|
| drop the title key | 3 |
| make every hit a hard failure | 3 |
| read the first `[#N]` link instead of the last | 1 |
| make `excludeVersionSection` a no-op | 3 |

### Types of changes

### Related issue(s):

1. DEV-2678

A separate, deliberately deferred follow-up task removes the entries this leak already published twice from the `16.x`-and-up sections. It is independent of this PR and can land in either order.

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

Release tooling only (`bin/`, `scripts/__tests__/`, `.changelogs/README.md`, the root eslint script).

### Checklist:

- [x] I have signed the [Contributor License Agreement](https://cla.handsontable.com/sign) with one signature covering both Handsontable and HyperFormula; the `cla/signed` check on this PR confirms it.

`.changelogs/README.md` gains a "No entry may be published twice" section covering both mechanisms, the severity table, and how to resolve a failure.

[skip changelog]

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2678

---

> [!NOTE]
> **Low Risk**
> Release and changelog tooling only; no runtime product code. `sync` can now fail mid-RC when duplicates are detected, which is intentional.
> 
> **Overview**
> Prevents the same changelog item from shipping twice when stale `.changelogs/*.json` files survive a release-to-develop merge-back.
> 
> Adds **`bin/lib/published-entries.js`** to compare pending entries against published `CHANGELOG.md` (PR/issue number and normalized title), with **errors** for conclusive duplicates and **warnings** for ambiguous cases (public issues, backport titles). **`bin/changelog consume`** and **`sync`** call this before compiling; **`consume --dry-run`** runs the check too, so existing PR CI already enforces it. The stable **merge-to-develop** step in **`publish.yml`** refuses to push if the dry-run fails.
> 
> Documents the rules and remediation in **`.changelogs/README.md`**, extends eslint to **`bin/lib`**, and adds **`changelog-published-entries.test.mjs`** for the matching logic.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ecefc467d1e21ddbc69da8ae9f1aa16d18ab8241. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changelogs/README.md
+++ b/.changelogs/README.md
@@ -118,3 +118,35 @@ bin/changelog consume
 This command "consumes" all changelog entries, asserts that they're all valid, formats them, and inserts the result into `CHANGELOG.md`. It also deletes all existing `.changelogs/*.json` files.
 
 It is side-effect free (as in it does nothing outside of your local copy of this repository), to undo just checkout the old versions of `.changelogs` and `CHANGELOG.md`.
+
+
+## No entry may be published twice
+
+`consume` and `sync` both refuse to compile a pending entry whose change `CHANGELOG.md` already publishes. Without that check the same change gets announced in two consecutive releases, which happened at most releases up to 18.1.0.
+
+The cause is that a pending `.json` file can outlive the release-to-develop merge-back after the release branch has already consumed it. Two merge shapes produce it, and neither is easy to spot:
+
+- The change is committed once on the release branch and once on `develop`, with no ancestry between the two commits. Against the merge base the release side reads as add-then-delete, so develop's add is the only change on either side and the merge keeps it. **No conflict is raised at all.**
+- The `.json` file is edited on `develop` after the release branch consumed it. That is a modify/delete conflict, and resolving it in favor of develop keeps a file that should have gone.
+
+Rather than trying to recognize either shape, the check asserts the one thing both produce. It matches on two keys, with different severities:
+
+| Match | Result |
+|---|---|
+| `issueOrPR` already cited in `CHANGELOG.md`, and `issuesOrigin` is `private` | **fails** |
+| `issueOrPR` already cited, and `issuesOrigin` is `public` | warns |
+| the entry's title already published, under any number | warns |
+
+A pull request number cannot ship twice, so a `private` number match is conclusive. The other two can be legitimate: a public **issue** number may be cited by two releases when a partial fix is followed by a complete one, and one title may appear in two sections when a fix is backported to several release lines. Both keys are still needed — a wrong link in the published entry hides the number, and rewording an entry on one branch hides the title.
+
+`sync` targets one version section and already skips entries present in it, so there the check looks at every *other* section.
+
+Because `consume --dry-run` runs on every pull request (see `.github/workflows/checks.yml`), this is enforced on every PR as well as at release time. A failure names every offending file and prints the `git rm` line that clears them.
+
+**To resolve a failure**, delete the entry files it names — their change already shipped. If one of them is genuinely a new change that happens to reuse a released pull request number, renumber the entry to cite its own pull request instead. There is no skip flag: an entry that is already published has nothing left to announce.
+
+After performing a release-to-develop merge-back, check for this before pushing:
+
+```bash
+bin/changelog consume --date 2050-01-01 --dry-run
+```
--- a/.changelogs/README.md
+++ b/.changelogs/README.md
@@ -122,7 +122,7 @@ It is side-effect free (as in it does nothing outside of your local copy of this
 
 ## No entry may be published twice
 
-`consume` and `sync` both refuse to compile a pending entry whose change `CHANGELOG.md` already publishes. Without that check the same change gets announced in two consecutive releases, which happened at most releases up to 18.1.0.
+`consume` and `sync` both check a pending entry against what `CHANGELOG.md` already publishes, and refuse to compile it when that match is conclusive; a less certain match only warns. Without that check the same change gets announced in two consecutive releases, which happened at most releases up to 18.1.0.
 
 The cause is that a pending `.json` file can outlive the release-to-develop merge-back after the release branch has already consumed it. Two merge shapes produce it, and neither is easy to spot:
 
@@ -134,10 +134,11 @@ Rather than trying to recognize either shape, the check asserts the one thing bo
 | Match | Result |
 |---|---|
 | `issueOrPR` already cited in `CHANGELOG.md`, and `issuesOrigin` is `private` | **fails** |
-| `issueOrPR` already cited, and `issuesOrigin` is `public` | warns |
-| the entry's title already published, under any number | warns |
+| `issueOrPR` already cited, `issuesOrigin` is `public`, and the title also matches | **fails** |
+| `issueOrPR` already cited, `issuesOrigin` is `public`, and the title does not match | warns |
+| only the entry's title already published, under any number | warns |
 
-A pull request number cannot ship twice, so a `private` number match is conclusive. The other two can be legitimate: a public **issue** number may be cited by two releases when a partial fix is followed by a complete one, and one title may appear in two sections when a fix is backported to several release lines. Both keys are still needed — a wrong link in the published entry hides the number, and rewording an entry on one branch hides the title.
+A pull request number cannot ship twice, so a `private` number match is conclusive. A public **issue** number may legitimately be cited by two releases when a partial fix is followed by a complete one — but a partial fix gets a new title, so a `public` number match that also matches the title is not that case, and fails the same way a `private` match does. A title-only match may still legitimately appear in two sections when a fix is backported to several release lines, so it always warns. Both keys are still needed — a wrong link in the published entry hides the number, and rewording an entry on one branch hides the title.
 
 `sync` targets one version section and already skips entries present in it, so there the check looks at every *other* section.
 
--- a/bin/lib/published-entries.js
+++ b/bin/lib/published-entries.js
@@ -110,12 +110,14 @@ const collectPublishedEntries = (changelogContents) => {
  * @returns {string} The contents without that section, or unchanged when it is absent.
  */
 const excludeVersionSection = (changelogContents, version) => {
-  const start = changelogContents.indexOf(`## [${version}]`);
+  const escapedVersion = version.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+  const heading = changelogContents.match(new RegExp(`^## \\[${escapedVersion}\\]`, 'm'));
 
-  if (start === -1) {
+  if (heading === null) {
     return changelogContents;
   }
 
+  const start = heading.index;
   const nextStart = changelogContents.indexOf('\n## [', start + 1);
 
   return changelogContents.slice(0, start) + (nextStart === -1 ? '' : changelogContents.slice(nextStart));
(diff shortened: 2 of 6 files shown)


qunabu:
Task linked: [DEV-2683 Remove the duplicated entries from the published changelog sections](https://app.clickup.com/t/9015210959/DEV-2683)
Task linked: [DEV-2678 Stop already-released changelog entries from being republished in the next release](https://app.clickup.com/t/9015210959/DEV-2678)

bin/lib/published-entries.js
```diff
@@ -0,0 +1,237 @@
...
+      return found;
+    }
+
+    found.push({
+      file,
+      issueOrPR: entry.issueOrPR,
+      issuesOrigin: entry.issuesOrigin,
+      citedSection,
+      titleSection,
+      severity: citedSection !== undefined && entry.issuesOrigin === 'private' ? 'error' : 'warning'
```
budnix: Can I ask about the `public` case here? A merge-back leftover of a public-issue entry is the exact same file, so it matches both the number key and the title key — and it still comes out as a warning, so `consume` writes it.

I tried it against the real changelog with `.changelogs/13101.json` and got `citedSection 18.1.0, titleSection 18.1.0, severity "warning"`. 20 of the 39 pending entries are `public` today, so that is about half the group this gate was built for.

A hit on *both* keys looks safe to escalate. The partial-fix-then-complete-fix case you describe gets a new title, so it only trips the number key.

```suggestion
      severity: citedSection !== undefined
        && (entry.issuesOrigin === 'private' || titleSection !== undefined) ? 'error' : 'warning'
```

budnix (commented):
Three inline comments below. One more finding has no anchor here, because the file is not in this diff:

**`.github/workflows/publish.yml`, the `stable-merge` job (around line 1372).** It checks out `develop`, merges the release branch, and pushes straight to `develop`. There is no pull request, so `checks.yml` never runs the dry-run on it — and that is the same silent merge shape the module docstring names. The README tells a person to run the dry-run after a merge-back, but the merge-back is automated, so nobody reads that. One step between the merge and the push would cover it:

```yaml
- name: Assert no already-published changelog entry survived the merge
  run: bin/changelog consume --date 2050-01-01 --dry-run
```

bin/lib/published-entries.js
```diff
@@ -0,0 +1,237 @@
...
+ * @param {string} changelogContents The contents of `CHANGELOG.md`.
+ * @param {string} version The version whose section to remove.
+ * @returns {string} The contents without that section, or unchanged when it is absent.
+ */
+const excludeVersionSection = (changelogContents, version) => {
+  const start = changelogContents.indexOf(`## [${version}]`);
+
+  if (start === -1) {
+    return changelogContents;
+  }
```
budnix: This `indexOf` is not tied to the start of a line, but `collectPublishedEntries` uses `^## \[`. If an entry line ever mentions a version heading inline, the cut lands mid-line: the unrelated entry gets truncated and the real section survives, so `sync` then blocks every entry in it.

Contrived today, agreed — but it is new code, and a line-anchored search closes it.

```suggestion
  const escaped = version.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
  const heading = changelogContents.match(new RegExp(`^## \\[${escaped}\\]`, 'm'));

  if (heading === null) {
    return changelogContents;
  }

  const start = heading.index;
```

The other option is to pass in the `start`/`end` that `syncCommand` already computed, instead of finding them twice.

.changelogs/README.md
```diff
@@ -118,3 +118,35 @@ bin/changelog consume
 This command "consumes" all changelog entries, asserts that they're all valid, formats them, and inserts the result into `CHANGELOG.md`. It also deletes all existing `.changelogs/*.json` files.
 
 It is side-effect free (as in it does nothing outside of your local copy of this repository), to undo just checkout the old versions of `.changelogs` and `CHANGELOG.md`.
+
+
+## No entry may be published twice
+
+`consume` and `sync` both refuse to compile a pending entry whose change `CHANGELOG.md` already publishes. Without that check the same change gets announced in two consecutive releases, which happened at most releases up to 18.1.0.
```
budnix: Small wording thing. This line says both commands "refuse" a pending entry that is already published, but the table right below has three rows and only one of them fails. Worth softening so the two match — the table is the part someone will act on.

bin/lib/published-entries.js
demtario: Fixed. Severity now escalates to `error` when the number match is `public` but the title also matches, since a partial-then-complete fix always gets a new title and would only trip the number key alone.

bin/lib/published-entries.js
demtario: Fixed with your suggested line-anchored regex. Added a regression test with an inline version-heading mention that used to get truncated mid-line under the old `indexOf`.

.changelogs/README.md
demtario: Reworded, and updated the table below it too (it now has 4 rows to reflect the escalation fix from the other comment).

demtario:
Fixed. Added a bin/changelog consume --date 2050-01-01 --dry-run gate before both direct-to-develop publish points in the stable-merge job's merge step: the clean-merge path and the auto-resolved-lockfile-conflict path. The conflict-with-a-PR path is untouched since that PR already gets checks.yml's own dry run.

budnix (approved):

PR #13293, opened 2026-08-28
DEV-406: Refill the rendered row band when the measured rows shrink during a draw

### LINKED ISSUE #6452: Resizing column in flex container doesn't draw appropriate number of rows
### Description
hot-column-size

👋  Hey y'all, we're running into an issue where when we resize columns, the resulting `render` doesn't draw rows all the way to the bottom. 

It's possible that this is because we keep our HoT instance in a flex-column container? I suppose that could be throwing off some height calculation, but I haven't gotten very far down that path yet.

### Steps to reproduce

You can see it in the GIF above, or using the fiddle + steps below.

1. Go to https://jsfiddle.net/ehxvtnf9/5/
2. Resize column C to be very narrow
3. Re-run fiddle (column C should still be very narrow at this point)
4. Double click border between C&D to resize C. 
5. Notice that HoT hasn't rendered enough rows to fill the viewport.
6. Scroll juuuust a bit to force another render, the rows pop in!

### Your environment
* Browser Name and version: Chrome Latest
* Operating System: OS X 10.14.4

No big deal atm, I'm currently trying to find a workaround. Just wanted to let ya know. 👍

adrianspdev:
### Context

The PR fixes the grid leaving a blank area under the last rendered row when the rendered rows shrink during a draw – for example after a column autosize (double-click on the resize handle) widens a column enough for its wrapped text to fit on one line, or after `setDataAtCell()` replaces long values with short ones. The rows collapsed, but the grid kept rendering only the rows that had filled the viewport while they were tall, and the missing rows appeared only after the next scroll (#6452, reported in 2019 and still reproducible on 18.0.0).

Root cause: Walkontable computes the rendered row band before the cells render, from `rowHeightCache` – the configured heights merged with `wtViewport.oversizedRows`, which holds heights measured on the *previous* render. After the render, `markOversizedRows` detected the shrink and invalidated the cache, but only the visible calculators were recomputed; the rendered band and the DOM stayed sized for the tall rows.

The fix (`src/3rdparty/walkontable/src/table/drawCycle.ts`, `viewport/calculatorFactory.ts`):

- `renderCellBand` returns the `markOversizedRows` verdict.
- The master draw cycle runs `refillRenderedRowsBandIfShrunk`: it proposes a fresh rendered band without assigning it and, only when the proposal grows the **bottom** edge, rebuilds the calculators, applies the union of the previous band and the proposal (new `Viewport#extendRenderedRowsBandTo`, so the band only ever grows), and renders the band again. The loop is bounded by `MAX_ROWS_BAND_REFILL_PASSES = 3` (a stale record for a row just outside the first band caps the proposal, so one pass is often not enough).
- An earlier proposed `startRow` alone is deliberately not a trigger: virtualized merged cells propose exactly that on every scroll draw (rowspan-inflated `oversizedRows` + per-band `modifyRowHeightByOverlayName` heights) and their pre-existing band is already correct – acting on it regressed `mergeCells/__tests__/selection.spec.js` during development.
- The `skipSecondPass` predicate that follows takes `rowHeightsChanged` into account, so the refill's cache rebuild cannot hide a needed `createVisibleCalculators()` pass.

Behavior notes for reviewers:

- Zero extra work on steady-state draws (`markOversizedRows` reports no change) and on draws where rows only grew.
- On a shrink draw the cell renderer and the `beforeRenderer`/`afterRenderer` hooks run once per refill pass; `beforeViewRender`/`afterViewRender` still fire once per draw.
- Exhausting the cap leaves the viewport under-filled until the next scroll or resize – i.e. exactly the pre-fix behavior, reachable only with a cascade of 4+ stale tall records each sitting just past the previous band.
- The second case listed in DEV-406 (scrollbar "jumping" while a column is narrowed) did not reproduce on 18.x during triage and is out of scope here.

No public API, option, or default changes.

### How has this been tested?

- Walkontable specs (`src/3rdparty/walkontable/test/spec/table.spec.js`): the rows-shrink refill, the grow-path calculator freshness, the zero-extra-pass invariant, and the declined bottom-edge-inward proposal. `npm run test:walkontable --prefix handsontable` → 820 specs, 0 failures.
- Playwright: `tests/e2e/column-autosize-viewport-refill.spec.ts` with fixture `tests/fixtures/demo/column-autosize-viewport-refill.html` and page object `tests/fixtures/pages/ColumnAutosizeRefillPage.ts` – double-click autosize and data-change paths; verified RED on the pre-fix build (gap ≈ 203px) and GREEN on all six legs. `cd tests && npx playwright test e2e/column-autosize-viewport-refill.spec.ts` → 12 passed.
- Full regression on the final build: legacy E2E `npm run test:e2e --prefix handsontable` → 9807 specs, 0 failures (run twice); Playwright `--project=e2e-main` → 490 passed; `npm run test:unit --prefix handsontable` → 4043 tests passed; `test:types`, eslint, stylelint clean.
- Manual, with real mouse input, following the exact steps from #6452 (fiddle `ehxvtnf9/5` copied verbatim; `persistentState` stand-in via localStorage since the plugin was removed in 18.0): drag column C narrow → reload → double-click the C|D border. Before: 4 rows rendered, ~650px blank at 1200×800. After: 28 rows, viewport filled. Also verified with all tall rows pre-measured (scrolled through first) and with a fixed `height: 500`.

### Types of changes

### Related issue(s):

1. #6452
2. DEV-406

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-406

---

> [!NOTE]
> **Medium Risk**
> Touches the master draw cycle and viewport band logic (multi-pass render, hook timing); behavior is heavily spec’d but edge cases can still leave under-fill if the pass cap is exhausted.
> 
> **Overview**
> Fixes **#6452**: when row heights shrink during a draw (column autosize, `setDataAtCell`, etc.), Walkontable no longer leaves a blank strip under the last rendered row until the user scrolls.
> 
> After the first `renderCellBand`, if **`markOversizedRows`** reports a height change (and AutoRowSize is off), the master draw runs **`refillRenderedRowsBandIfShrunk`**: it **propose-only** recalculates the rendered row band, and when the bottom edge must grow, recomputes calculators, **unions** the old band with the proposal via **`extendRenderedRowsBandTo`**, and re-renders—up to **`MAX_ROWS_BAND_REFILL_PASSES` (3)**. Refill triggers only on a **later `endRow`**, not an earlier `startRow` alone (avoids bad interaction with virtualized merged cells). Guards decline unsafe refills (non-overlapping bands, frozen-column sync mismatch). **`createRowsCalculator` / `createColumnsCalculator`** gain **`proposeOnly`** so declined passes do not reset header width memos. The post-render **second visible-calculator skip** also requires **`!rowHeightsChanged`** so refill cache rebuilds do not skip a needed **`createVisibleCalculators()`**.
> 
> Adds Walkontable/Jasmine specs, unit tests for the new helpers, Playwright coverage for autosize/data-shrink, changelog **6452**, and AGENTS/rendering-lifecycle docs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ace3b8e8276f616433e3c5997c11984272f0a82a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-406 Resizing column in flex container doesn't draw appropriate number of rows](https://app.clickup.com/t/9015210959/DEV-406)

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -517,7 +532,141 @@ function renderCellBand(
...
+ * second settles the rows the grown band newly exposed, whose own oversized records were recorded
+ * on an older draw and could not be re-measured while they sat outside the band. The third is
+ * headroom.
+ *
+ * Exhausting the cap does NOT self-correct on the next draw. The pass that spent the last iteration
+ * re-measured every row inside the final band accurately, so what still caps the band is a stale
+ * tall record for a row just OUTSIDE it — and that is precisely the state in which the next draw's
+ * `markOversizedRows` reports no change, `rowHeightsChanged` is `false`, and the refill never fires.
+ * The viewport then stays under-filled until a scroll or a resize brings those rows into a rendered
+ * band, i.e. exactly the pre-fix #6452 behavior. That is not a regression, and it takes a cascade of
```
demtario: **high** A shrink that spans more stale out-of-band tall records than the cap leaves the blank strip in place, and the fixture in this PR already consumes all three passes.

Each pass's proposal is capped at the *first* row below the band that still holds a stale tall record: `calculateAxis` stops at the row whose cumulative offset crosses the viewport, and `resetOversizedRows` only wipes records for rows inside the band, so every previously-measured tall row below it survives to cap the next proposal. Passes needed therefore scale with the number of stale tall records below the first band. `rowHeightsUniform` is false in exactly this scenario, so `applyRenderedRowsBandOverscan` adds nothing on top.

In the fixture, rows 1-7 are measured tall on the post-init draw and the pre-fix band holds ~5 rows, so rows 5-7 sit outside it with stale records and the shrink draw needs all three passes to settle. A grid with more previously-measured tall rows below the band exhausts the cap and keeps the blank area, which is the reported symptom. Not a regression, but a narrower fix than the comment here describes, and `AGENTS.md:111` repeats the same framing ("a cascade, not an oscillation").

Suggested: state the real growth rate (one pass per stale out-of-band record) and the practical limit in both places, and add a spec that stages two stale out-of-band records so the cap's behavior is pinned. Nothing currently covers it: every new spec drives a single pass.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -204,12 +204,12 @@ function runMasterDrawCycle(table: Table, ctx: DrawContext): void {
...
-        table,
-        ctx,
-        filters,
-        isPureVerticalScrollDraw(wtOverlays),
-      );
+      const columnHeadersRenderSkippable = isPureVerticalScrollDraw(wtOverlays);
+      const rowHeightsChanged = renderCellBand(table, ctx, filters, columnHeadersRenderSkippable);
+
+      if (rowHeightsChanged && !wtSettings.getSetting('externalRowCalculator')) {
+        refillRenderedRowsBandIfShrunk(table, ctx, columnHeadersRenderSkippable);
```
demtario, line 203: **medium** `ctx.syncFrozenRows` is resolved from the pass-1 column band and is not re-derived for the refill's own render.

`shouldSyncOversizedRowsWithFrozenOverlays` is a function of `table.getFirstRenderedColumn()` (`axisSizing/oversizedRows.ts:193-198`), which on the master reads `columnsRenderCalculator.startColumn` (`table/rangeQuery/virtualRange.ts:133-135`). Pass 1's column band gets `applyRenderedColumnsBandOverscan`, which can lower `startColumn` by up to 8, so a natural `startColumn` of 1..8 renders as 0 and the flag is captured `false` - after which line 183 has already run `releaseFrozenOversizedRows()`. The refill then calls `createCalculators(false)` with no `stationaryBands`, and `calculatorFactory.ts:376-390` applies the overscan only under that flag, so the recomputed band can start past column 0. The refill's `renderCellBand` then wipes and re-measures a master that no longer renders the frozen columns, with no `frozenOversizedRows` exemption left: every frozen-tall row reads as shrunk, its record is dropped, and the loop keeps proposing from heights missing those contributions.

Narrow but reachable: frozen columns, uniform column widths, a scroll draw moving toward the inline start (the overscan only extends toward the scrolled side), and a row shrink on that same draw.

Suggested: re-derive the decision inside the loop after `createCalculators(false)`, or decline the pass when the recomputed `startColumn` disagrees with the band the flag was captured from. No new spec or fixture sets `fixedColumnsStart`, `fixedRowsTop`, or `fixedRowsBottom`, so nothing covers this path.

handsontable/src/3rdparty/walkontable/src/viewport/calculatorFactory.ts
```diff
@@ -586,6 +587,65 @@ export const calculatorFactory: CalculatorFactory = {
...
+      band.startRow -= extension;
+      band.count += extension;
+      band.rowStartOffset += extension;
+      band.startPosition = this.rowHeightCache.getOffset(band.startRow);
+    }
+
+    const lastRow = this.wtSettings.getSetting<number>('totalRows') - 1;
+    const clampedEndRow = Math.min(endRow, lastRow);
+
+    if (clampedEndRow > band.endRow) {
```
demtario, line 671: **low** The end-edge branch cannot fire from the only call site.

The refill declines unless `proposedBand.endRow > previousEndRow` (`drawCycle.ts:645-651`), `createCalculators(false)` assigns that proposal unmodified (no overscan or stabilization without `stationaryBands`), and `previousEndRow` is read straight off `rowsRenderCalculator.endRow` (`table/rangeQuery/virtualRange.ts:61-63`). So `clampedEndRow <= previousEndRow < band.endRow` always, and only the start-edge branch above ever runs. Worth either dropping the branch or noting in the JSDoc that it is defensive, since the "union" the doc describes is in practice a start-edge-only extension. The start branch that does fire is itself uncovered: no spec scrolls mid-table and then shrinks rows above the band.

handsontable/src/3rdparty/walkontable/src/viewport/calculatorFactory.ts
```diff
@@ -586,6 +587,65 @@ export const calculatorFactory: CalculatorFactory = {
...
+    if (!(band instanceof RenderedRowsCalculationType) || band.startRow === null || band.endRow === null) {
+      return;
+    }
+
+    if (startRow >= 0 && startRow < band.startRow) {
+      const extension = band.startRow - startRow;
+
+      band.startRow -= extension;
+      band.count += extension;
+      band.rowStartOffset += extension;
```
demtario, line 664: **low** `count` and `rowStartOffset` grow by an uncapped extension here, and both are read on later draws.

`stabilizeRenderedRowsBand` tops each subsequent scroll-driven band up to the previous `count`, so an inflated size becomes the baseline; `areAllProposedVisibleRowsAlreadyRendered` pads its containment test by `rowStartOffset`, making later scroll steps less likely to resolve as fast draws. The JSDoc's "mirrors `applyRenderedRowsBandOverscan` field for field" holds for the field list but not the magnitude: that path caps at `ROW_BAND_OVERSCAN_MAX`. Either clamp the extension or leave `rowStartOffset` out of this mutation and say why in the JSDoc.

demtario (commented):
4 inline notes below.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -517,7 +532,141 @@ function renderCellBand(
...
+    }
+
+    // Full recompute (no stationary bands: this is a content change, not a scroll step) so the
+    // rendered and visible calculators are rebuilt from the freshly measured heights together.
+    wtViewport.createCalculators(false);
+
+    // Apply the UNION of the previous band and the proposal, never the proposal alone: a proposal
+    // built from re-measured heights can move one edge inwards, and an edge inside the band the
+    // DOM already shows would leave the viewport shorter than it was.
+    wtViewport.extendRenderedRowsBandTo(previousStartRow, previousEndRow);
```
budnix, line 701: The union keeps every row the DOM already shows, which makes sense, but nothing caps how big it can get. If row heights drop across the whole dataset at once and the user is scrolled deep — say row 5000 of 10000 — the recompute puts the proposal near the bottom while `previousStartRow` is still 5000, so the union is a 5000-row band and the `renderCellBand` on the next line builds all of it in one go. Plain column autosize does not reach this, because `resetOversizedRows` only wipes records inside the band, so the tall rows above keep the offsets anchored. Would it be worth bailing out of the pass when the proposal does not overlap the previous band at all?

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -204,12 +204,12 @@ function runMasterDrawCycle(table: Table, ctx: DrawContext): void {
...
-        isPureVerticalScrollDraw(wtOverlays),
-      );
+      const columnHeadersRenderSkippable = isPureVerticalScrollDraw(wtOverlays);
+      const rowHeightsChanged = renderCellBand(table, ctx, filters, columnHeadersRenderSkippable);
+
+      if (rowHeightsChanged && !wtSettings.getSetting('externalRowCalculator')) {
+        refillRenderedRowsBandIfShrunk(table, ctx, columnHeadersRenderSkippable);
+      }
 
       if (!wtSettings.getSetting('externalRowCalculator')) {
```
budnix: `markOversizedRows` returns `false` as its very first statement when `externalRowCalculator` is on, so `rowHeightsChanged` is already `false` there and the second half of this check can never decide anything. The same setting is then read again three lines down. Reading it once into a local drops a lookup per draw and makes the gate easier to follow.
```suggestion
      const rowHeightsChanged = renderCellBand(table, ctx, filters, columnHeadersRenderSkippable);
      const hasExternalRowCalculator = wtSettings.getSetting('externalRowCalculator');

      if (rowHeightsChanged) {
        refillRenderedRowsBandIfShrunk(table, ctx, columnHeadersRenderSkippable);
      }

      if (!hasExternalRowCalculator) {
```

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -517,7 +532,141 @@ function renderCellBand(
...
+
+    // Propose only — this does not assign the viewport's calculators. `createRowsCalculator` builds a
+    // calculator without touching `rowsRenderCalculator`, so a band that must not grow leaves the
+    // rendered state alone. It is not entirely free of side effects: it resets the `rowHeaderWidth`
+    // memo to `NaN` (`viewport/calculatorFactory.ts`), so a declined refill still costs one
+    // re-measure on the next `getRowHeaderWidth()`.
+    // Mirror `createCalculators`: the render band reads the scrollbar-unaware box on the single-pass
+    // path; the legacy measured path serves render and visible from the same 'visible' band.
+    const renderBand = wtViewport.usesLayoutSnapshotForCalculators() ? 'render' : 'visible';
+    const proposedBand = wtViewport.createRowsCalculator(['rendered'], renderBand).getResultsFor('rendered');
```
budnix: This proposal walk runs on any height change, including rows that got taller and grids with `renderAllRows`, where the bottom edge can never grow — so the pass builds a whole rows calculator just to return two lines later. It also resets the `rowHeaderWidth` memo to `NaN`, so even a declined refill costs a re-measure on the next `getRowHeaderWidth()`. The two `ensureBuilt()` calls above are free here, since line 232 runs them anyway. A cheap check first — band already ends at `totalRows - 1`, or the last rendered row's bottom already reaches the holder — would skip most of that.

handsontable/src/3rdparty/walkontable/src/viewport/calculatorFactory.ts
```diff
@@ -586,6 +587,65 @@ export const calculatorFactory: CalculatorFactory = {
...
+   * `count`, `rowStartOffset`, `rowEndOffset`, and `startPosition` stay consistent with the moved
+   * edges. It is a no-op for a band that renders all rows (`RenderedAllRowsCalculationType`), which
+   * has no edge to move.
+   *
+   * @this Viewport
+   * @param {number} startRow The first row the band must cover. Ignored when negative, which is what
+   * the range queries answer before the first render.
+   * @param {number} endRow The last row the band must cover. Clamped to the last row of the dataset.
+   */
+  extendRenderedRowsBandTo(this: Viewport, startRow: number, endRow: number): void {
```
budnix, line 652: The two new tests cover a simple refill and a declined one, but nothing exercises this method directly — the union where `startRow` moves inward while `endRow` grows, the `endRow` clamp, or the no-op on a `RenderedAllRowsCalculationType` band. The union is the interesting one, since that is where the band size can run away. `directionalBandOverscan` is already exported from this file for unit tests, so a small unit test here would fit the pattern.

.changelogs/6452.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed the grid leaving a blank area under the last rendered row when the rows shrank after a column autosize or a data change, until the next scroll.",
+  "type": "fixed",
```
budnix, line 4: Each refill pass re-renders the whole band, so with `MAX_ROWS_BAND_REFILL_PASSES = 3` the cell renderer — and the `beforeRenderer` / `afterRenderer` hooks core fires from inside it — can run four times per cell in one draw instead of once. A hook that appends a node or attaches a listener will now do it repeatedly. `.changelogs/README.md` says a second entry is right when it lands in another section: "The common case is a public issue fixed alongside a private behavior change: `fixed` plus `changed`." Worth adding a `changed` entry next to this one?

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
adrianspdev: Reframed in b30e47ede8: the `MAX_ROWS_BAND_REFILL_PASSES` JSDoc, `AGENTS.md`, and `RENDERING-LIFECYCLE.md` now state the one-pass-per-stale-out-of-band-record growth rate and the practical limit, including that the #6452 fixture itself already consumes all three passes. Added a spec staging two stale out-of-band records (`should settle a shrink that leaves stale tall records below the band within the pass cap`) — it pins the cascade at 3–4 band renders (the exact count is offset-sensitive at the band edges) and asserts the viewport fills within the cap.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
adrianspdev: Fixed in b30e47ede8 with the decline variant: `refillDisagreesWithFrozenColumnSync` predicts the refill's column band with a propose-only `createColumnsCalculator(['rendered'], renderBand)` (the same construction `createCalculators(false)` would assign) and declines the pass on any disagreement with the captured `ctx.syncFrozenRows`, before anything is reassigned. Re-deriving the flag inside the loop wasn't enough on its own — by that point `releaseFrozenOversizedRows()` has already run, so a flipped flag can't restore the exemption.

Added a frozen-column spec (`should refill the viewport without dropping frozen-derived row heights`) covering `fixedColumnsStart` + shrink + refill, asserting the frozen-derived height survives on the master and the inline-start clone. The exact overscan-induced mismatch is not deterministically stageable in the Jasmine harness (it needs a scroll-toward-inline-start draw whose overscan lands `startColumn` at 0 on the same draw a row shrinks), so the mismatch branch is pinned by the guard's logic plus the agreement-path spec.

handsontable/src/3rdparty/walkontable/src/viewport/calculatorFactory.ts
adrianspdev: Kept the branch as defensive and documented it in b30e47ede8: the JSDoc now states that only the start edge can move from the refill call site (and why), and the end-edge branch — including the clamp — is pinned by direct unit tests instead. The start-edge branch got its spec too: `should keep the rows the DOM already shows when the refill proposal starts lower` scrolls mid-table and shrinks the rows above the band.

handsontable/src/3rdparty/walkontable/src/viewport/calculatorFactory.ts
adrianspdev: Bounded through the caller in b30e47ede8: the refill now declines any proposal that does not overlap or touch the previous band, so the union — and with it the `count`/`rowStartOffset` growth — can never exceed the two bands' combined span. The JSDoc no longer claims a field-for-field mirror of the overscan; it states the bound and why growing `rowStartOffset` with the extension is deliberate (the folded-back rows really are rendered above the viewport, so the `viewportRowRenderingThreshold` containment padding and the `stabilizeRenderedRowsBand` baseline must both see them, exactly as they see overscan rows).

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
adrianspdev: Added the bail-out in b30e47ede8: a proposal with `startRow > previousEndRow + 1` (no overlap, not adjacent) declines the pass, so the deep-scroll whole-dataset shrink keeps the pre-fix behavior (under-filled until the next scroll or resize) instead of building the whole gap in one `renderCellBand`. This guard is also what bounds the union's start-edge growth for the sibling comment on `extendRenderedRowsBandTo`. Documented in the refill JSDoc and `AGENTS.md`.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
adrianspdev: Applied in b30e47ede8, with a comment noting that `markOversizedRows` already carries the `externalRowCalculator` condition in its return value.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
adrianspdev: Done in b30e47ede8: `renderAllRows` returns before the loop, a band already at `totalRows - 1` returns before the cache builds and the proposal walk, and the `rowHeaderWidth` memo reset is gone from declined passes — the proposal now passes `{ proposeOnly: true }`, which skips the memo reset inside `createRowsCalculator`. (First attempt hoisted the resets to the assigning call sites instead; that changed the reset's position relative to the neighboring `getViewportWidth`/`getViewportHeight` reads and broke `AutoRowSize should properly count height`, so the resets stay where they are and the flag opts out — noted in `AGENTS.md`.)

The last-rendered-row-bottom check I left out: answering it without the calculator means duplicating the fixedRows/scrollbar viewport math and inviting drift, so grew-only draws still pay one proposal walk per height-change draw — noted in the loop comment.

handsontable/src/3rdparty/walkontable/src/viewport/calculatorFactory.ts
adrianspdev: Added in b30e47ede8: seven direct unit tests for `extendRenderedRowsBandTo` in `calculatorFactory.unit.js`, following the `directionalBandOverscan` pattern — start-edge union with derived-field consistency (`count`/`rowStartOffset`/`startPosition`), both-edges union (pins the defensive end-edge branch), the `endRow` clamp, inside-range no-shrink, negative `startRow`, the `RenderedAllRowsCalculationType` no-op, and the empty-band no-op.

.changelogs/6452.json
adrianspdev: We decided to keep a single entry for this PR: the renderer/hook re-runs are a consequence of the fix rather than a separately announced behavior change, so no `changed` entry was added. The per-pass `beforeRenderer`/`afterRenderer` re-fires (and the draw-level hooks that do not repeat) are documented in the refill JSDoc and `AGENTS.md` for integrators reading the code.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -517,10 +537,227 @@ function renderCellBand(
...
+ * The upper bound on how many extra cell-band renders one draw may spend refilling the viewport
+ * after the rendered rows measured shorter than the band was computed from.
+ *
+ * Passes scale with the number of stale tall records BELOW the band, roughly one pass per record:
+ * each pass's proposal is capped at the first row below the band that still holds a stale tall
+ * record (the axis calculator stops where the cumulative offsets — stale record included — cross
+ * the viewport, and `resetOversizedRows` wipes only the rows inside the band, so every
+ * previously-measured tall row below it survives to cap the next proposal). Only rendering that row
+ * re-measures it and lets the following pass propose past it. The overscan does not help here:
+ * `rowHeightsUniform` is `false` in exactly this scenario. So the practical limit is a shrink that
```
demtario: **medium** The reason given here for the overscan not helping is false.

`rowHeightsUniform` is not `false` in this scenario. It is settings-only: `isUniformSizeSetting(rowHeights) && isUniformSizeSetting(minRowHeights) && !hasHook('modifyRowHeight')` (`handsontable/src/tableView.ts:947-952`), with `isUniformSizeSetting(undefined) === true` (`tableView.ts:53-55`). Measured `oversizedRows` never enter it, and `viewport/viewport.ts:280-281` has to AND the record count in explicitly to get the meaning this sentence assumes. The #6452 fixture sets neither setting and registers no `modifyRowHeight` producer, so it is `true` there, and it has to be: `usesLayoutSnapshotForCalculators()` (`calculatorFactory.ts:446-452`) gates on it, so a `false` value would put every refill on the legacy `'visible'` path.

The conclusion holds, the reason does not. The overscan cannot shortcut the cascade because `applyRenderedRowsBandOverscan` runs only inside `if (stationaryBands)` (`calculatorFactory.ts:397-405`) and the refill's `createCalculators(false)` never sets it. Worth correcting rather than filing as a nit: the sentence reads as authority on which path the refill takes, and `allowsStationaryBands` (`calculatorFactory.ts:467-471`) deliberately drops the uniform-size requirement, so on a scroll-driven shrink draw the overscan genuinely can fire and can pull a stale record into the band, which is the opposite of what this asserts.

Mirrored in `AGENTS.md:111`.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -517,10 +537,227 @@ function renderCellBand(
...
+    }
+
+    // A proposal that shares no row with (and is not adjacent to) the previous band would union into
+    // a band spanning the whole gap between them — a whole-dataset shrink while scrolled deep puts
+    // the proposal near the dataset end with the previous band thousands of rows above it, and the
+    // `renderCellBand` below would build all of it in one go. Decline the pass instead: the viewport
+    // stays under-filled until the next scroll or resize (the pre-fix #6452 behavior). This guard is
+    // also what bounds `extendRenderedRowsBandTo`'s start-edge growth: with it, the union can never
+    // exceed the two bands' combined span.
+    if (proposedBand.startRow > previousEndRow + 1) {
```
demtario, line 686: **medium** The two declines that change behavior have no test.

`proposedBand.startRow > previousEndRow + 1` here, and the `refillDisagreesWithFrozenColumnSync` branch below, both suppress a refill that `36926687` would have performed. The new frozen-columns spec (`test/spec/table.spec.js:963-1017`) reaches only the agreeing path: it asserts `getFirstRenderedColumn() > 0` before the shrink, so the captured `ctx.syncFrozenRows` and the proposal's `startColumn > 0` agree, the predicate returns `false`, and the spec passes identically with the whole function deleted. It is also the only new spec that sets `fixedColumnsStart`. Nothing puts a proposal out of overlap with the previous band, and nothing produces a captured-`false` / recomputed-`> 0` mismatch. `proposeOnly` is unpinned too: the new unit tests cover `extendRenderedRowsBandTo`, which is prior-commit code.

`proposeOnly` is cheap to pin where those unit tests already are: build with and without the flag and assert `viewport.rowHeaderWidth` is untouched vs `NaN`. The no-overlap decline is stageable as a spec (scroll deep, shrink the whole dataset, assert the band does not move). If the frozen-sync disagreement cannot be staged in the harness, say that in the JSDoc rather than leaving the guard silent.

handsontable/src/3rdparty/walkontable/test/unit/viewport/calculatorFactory.unit.js
```diff
@@ -108,3 +109,125 @@ describe('directionalBandOverscan', () => {
...
+  it('should be a no-op for a band that renders all rows', () => {
+    const band = new RenderedAllRowsCalculationType();
+
+    band.startRow = 0;
+    band.endRow = 99;
+    band.count = 100;
+
+    const viewport = createViewportStub(band);
+
+    extendBandTo(viewport, 5, 200);
```
demtario: **medium** This test passes with the guard it targets removed.

`RenderedAllRowsCalculationType` (`calculator/calculationType/renderedAllRows.ts:11`) is a standalone class, so the `instanceof` bail at `calculatorFactory.ts:655` does fire, but these inputs cannot observe it. Delete the guard and: `startRow >= 0 && startRow < band.startRow` is `5 < 0`, false; `clampedEndRow = Math.min(200, 99) = 99` is not `> band.endRow` (99), false. Neither branch runs and all three assertions still hold.

Make the range stick out so the counterfactual fails: `createViewportStub(band, { totalRows: 200 })` with `extendBandTo(viewport, 0, 150)`. Without the guard `endRow` becomes 150 and `rowEndOffset` goes `undefined + 51` to `NaN`, so `expect(band.endRow).toBe(99)` fails as intended.

The other six tests in the block hold up against their counterfactuals: the `null`-edges case fails without the guard (`10 > null` is `true`), and the negative-start case fails without `startRow >= 0`.

demtario (commented):
3 inline notes below.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
adrianspdev: Corrected in 80589a62cc, in both places. Verified against `tableView.ts` — `rowHeightsUniform` is indeed settings-only and `true` in the fixture, with the record count ANDed in separately by `PositionCache`. The JSDoc and `AGENTS.md` now give the real reason (`applyRenderedRowsBandOverscan` runs solely under `stationaryBands`, which the refill's `createCalculators(false)` never sets), warn against reasoning from `rowHeightsUniform` there, and carry your counterpoint that pass 1's overscan genuinely can fire on a scroll-driven shrink draw and pull a stale record into the band.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
adrianspdev: Covered in 80589a62cc, all three parts:

- No-overlap decline: new spec `should decline a refill whose proposal does not overlap the rendered band` — 1000 rows, provided heights 60px, scrolled to offset 30000, then a whole-dataset shrink (provided heights to default + measured tall cells removed) in one draw; asserts the band and the DOM stay exactly as pass 1 rendered them. Counterfactual-verified: with the guard disabled the spec fails, with it the suite is green.
- `proposeOnly`: two unit tests next to the `extendRenderedRowsBandTo` block — a default build resets `rowHeaderWidth` to `NaN`, a propose-only build leaves it untouched while still computing a real band.
- Frozen-sync disagreement: stated in `refillDisagreesWithFrozenColumnSync`'s JSDoc that the disagreeing branch is untested by design — staging it needs the pass-1 overscan to land the column band at 0 (natural `startColumn` 1..8, scroll toward the inline start, uniform widths, 'auto' offset) on the same draw a row shrinks, which the harness cannot force deterministically — with a pointer to the agreement-path spec.

handsontable/src/3rdparty/walkontable/test/unit/viewport/calculatorFactory.unit.js
adrianspdev: Fixed in 80589a62cc with your exact inputs: `createViewportStub(band, { totalRows: 200 })` + `extendBandTo(viewport, 0, 150)`, with a comment explaining why these values observe the guard (without the `instanceof` bail the end-edge branch would grow `endRow` to 150 and `NaN` the offsets).

demtario (dismissed):

qunabu:
@cursoragent merge with latest develop to remove Argos CI

qunabu:
@cursoragent resolve conflicts

tests/fixtures/pages/ColumnAutosizeRefillPage.ts
```diff
@@ -0,0 +1,138 @@
...
+    this.theme = theme;
+    this.bundle = bundle;
+    this.grid = page.getByTestId('grid');
+  }
+
+  /**
+   * Navigate and wait for the grid to render (a real DOM condition, no sleep).
+   */
+  async goto(): Promise<void> {
+    await this.page.goto(`/tests/fixtures/demo/column-autosize-viewport-refill.html?theme=${this.theme}&bundle=${this.bundle}`);
```
demtario, line 26: **high** `goto()` never waits for the `document.write`-injected bundle. `tests/AGENTS.md` requires `await page.waitForFunction(() => 'Handsontable' in window)` before asserting on rendered cells — `expect` is a 10s budget and `dist/handsontable.js` is ~6 MB per worker. A cold CI worker can time out on the first `toBeVisible()`. Add the `waitForFunction` after `page.goto(...)`, then keep the cell assertion.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -499,8 +519,237 @@ function renderCellBand(
...
+    if (!renderCellBand(table, ctx, filters, columnHeadersRenderSkippable)) {
+      return;
+    }
+  }
+}
+
+/**
+ * Whether a refill pass would render with a frozen-column sync decision different from the one this
+ * draw already acted on. `ctx.syncFrozenRows` is a function of the master's first rendered column
+ * (`shouldSyncOversizedRowsWithFrozenOverlays`), resolved from the PREVIOUS draw's column band —
```
demtario, line 716: **medium** This JSDoc says `ctx.syncFrozenRows` is "resolved from the PREVIOUS draw's column band". It is captured after this draw's pass-1 `createCalculators` assigned the band, including columns overscan. The comparison below is already correct. Do not move the capture to match this sentence — rewrite it to say this draw's assigned pass-1 band. A later edit that "fixed" the capture to match "PREVIOUS draw" would break the overscan case this helper exists for.

demtario (commented):
2 inline notes below.

budnix (dismissed):

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -499,8 +519,237 @@ function renderCellBand(
...
+ * overscan genuinely can fire (`allowsStationaryBands` deliberately drops the uniformity
+ * requirement) and can pull a stale out-of-band record into the band. So the practical limit is a shrink that
+ * leaves at most `MAX_ROWS_BAND_REFILL_PASSES` stale out-of-band tall records between the pre-shrink
+ * band and the settled one — the #6452 fixture (rows 1-7 tall, a ~5-row pre-fix band) already
+ * consumes all three passes.
+ *
+ * Exhausting the cap does NOT self-correct on the next draw. The pass that spent the last iteration
+ * re-measured every row inside the final band accurately, so what still caps the band is a stale
+ * tall record for a row just OUTSIDE it — and that is precisely the state in which the next draw's
+ * `markOversizedRows` reports no change, `rowHeightsChanged` is `false`, and the refill never fires.
```
budnix, line 552: The reason given here does not hold up. On the next draw the band is built from the now-correct in-band heights, so it does reach the first stale tall row below the old band, `resetOversizedRows` wipes that record, and `markOversizedRows` reports the change — the refill would fire and would make progress. What actually stops the recovery is that nothing schedules another draw. Worth fixing the sentence, because as written it rules out an adaptive cap and a single deferred draw, and both would work.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -499,8 +519,237 @@ function renderCellBand(
...
+ * @returns {boolean} `true` when the pass must be declined.
+ */
+function refillDisagreesWithFrozenColumnSync(table: Table, ctx: DrawContext, renderBand: ViewportBand): boolean {
+  if (!table.wtSettings.getSetting<number>('fixedColumnsStart')) {
+    return false;
+  }
+
+  const wtViewport = table.deps.getWtViewport();
+  const proposedStartColumn = wtViewport
+    .createColumnsCalculator(['rendered'], renderBand, { proposeOnly: true })
```
budnix, line 749: `proposeOnly: true` skips the `rowHeaderWidth` memo reset, but the `createCalculators(false)` on line 696 does reset it and re-measures the row-header TH. With `fixedColumnsStart` set, no explicit `rowHeaderWidth`, and a row-header width that moved during this render, the predicted `startColumn` can end up on the other side of 0 from the one actually assigned — which is the mismatch this guard exists to stop. Curious whether reading the band after `createCalculators(false)` assigns it would be safer than predicting it.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -499,8 +519,237 @@ function renderCellBand(
...
+    // Apply the UNION of the previous band and the proposal, never the proposal alone: a proposal
+    // built from re-measured heights can move one edge inwards, and an edge inside the band the
+    // DOM already shows would leave the viewport shorter than it was.
+    wtViewport.extendRenderedRowsBandTo(previousStartRow, previousEndRow);
+
+    const filters = buildRenderFilters(table, ctx);
+
+    // A pass that measures no further change has settled the band; anything else is another shrink
+    // the grown band just exposed, and the next iteration decides whether it must grow again.
+    if (!renderCellBand(table, ctx, filters, columnHeadersRenderSkippable)) {
```
budnix, line 707: Each pass re-renders the whole band, so a wide grid pays the full cell-renderer cost per pass — a 300-column band of 14 rows is about 4200 renderer calls, times four on the worst draw, and that draw is the one right after a column resize. Only the bottom edge grows from this call site, so the appended rows are the only ones whose cells are actually new. Did the shape where a pass renders just those rows get ruled out, or was the whole-band render the simpler thing to land?

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -191,14 +192,18 @@ function runMasterDrawCycle(table: Table, ctx: DrawContext): void {
...
-        table,
-        ctx,
-        filters,
-        isPureVerticalScrollDraw(wtOverlays),
-      );
-
-      if (!wtSettings.getSetting('externalRowCalculator')) {
+      const columnHeadersRenderSkippable = isPureVerticalScrollDraw(wtOverlays);
+      const rowHeightsChanged = renderCellBand(table, ctx, filters, columnHeadersRenderSkippable);
+      const hasExternalRowCalculator = wtSettings.getSetting<boolean>('externalRowCalculator');
```
budnix, line 197: Line 251 reads the same setting through the funnel again, so this const only covers the first of the two uses. Using it there as well keeps them in step — as it stands, a change to how the flag is resolved here would quietly leave that branch on the old value.

handsontable/src/3rdparty/walkontable/test/spec/table.spec.js
```diff
@@ -655,4 +697,377 @@ describe('WalkontableTable', () => {
...
+            ? `<div style="height: 100px">${getData(row, column)}</div>`
+            : getData(row, column);
+        },
+      });
+
+      wt.draw();
+
+      getTableMaster().find('.wtHolder').scrollTop(460);
+
+      await sleep(20);
```
budnix, line 805: These four new specs each write `scrollTop`/`scrollLeft`, sleep 20ms, then call `wt.draw()` (same shape at 926, 988 and 1046), which `handsontable/no-fixed-sleep-in-spec` flags. Under CI load the scroll event may not be handled yet, so the draw runs at the old offset and the spec fails on its own precondition rather than on the refill. The file already carries a dozen older instances, so this is the existing pattern here — but waiting for the scroll position to actually land would make these four much less likely to flake.

tests/fixtures/pages/ColumnAutosizeRefillPage.ts
```diff
@@ -0,0 +1,138 @@
...
+   */
+  async expectViewportFilled(): Promise<void> {
+    await expect.poll(async () => (await this.probe()).gap, {
+      message: 'rendered rows should reach the bottom of the viewport',
+    }).toBeLessThanOrEqual(20); // a horizontal scrollbar may occupy the bottom of the holder; 20px covers every theme.
+
+    await expect.poll(async () => {
+      const { rendered, visible } = await this.probe();
+
+      return rendered >= visible;
```
budnix, line 101: `probe()` returns `{ rendered: 0, visible: 0 }` for an empty TBODY, so this check is `0 >= 0` and passes on its own. The docstring below says the two polls read separate frames, so a frame with no rendered rows would satisfy this one without the first poll ever seeing it. A lower bound on `rendered` would close that.

```suggestion
      return rendered > 0 && rendered >= visible;
```

tests/fixtures/pages/ColumnAutosizeRefillPage.ts
```diff
@@ -0,0 +1,138 @@
...
+   * `page.evaluate`, so they describe the same frame; each `expect.poll` above calls `probe()`
+   * independently, so the two assertions read separate frames. Returns `gap: Infinity` (never
+   * satisfies the `toBeLessThanOrEqual` assertion) when no rows are rendered yet, instead of
+   * throwing on a missing last row.
+   */
+  private async probe(): Promise<{ gap: number; rendered: number; visible: number }> {
+    return this.page.evaluate(() => {
+      const master = document.querySelector('[data-testid="grid"] .ht_master') as HTMLElement;
+      const holder = master.querySelector('.wtHolder') as HTMLElement;
+      const rows = master.querySelectorAll('tbody tr');
```
budnix, line 125: `master` is `null` whenever `.ht_master` is not in the DOM, so line 124 throws a TypeError instead of returning the `gap: Infinity` the docstring promises. That is exactly the cold-worker case, and the error then points at `querySelector` rather than at the grid never loading.

```suggestion
      const master = document.querySelector('[data-testid="grid"] .ht_master') as HTMLElement | null;

      if (!master) {
        return { gap: Number.POSITIVE_INFINITY, rendered: 0, visible: 0 };
      }

      const holder = master.querySelector('.wtHolder') as HTMLElement;
      const rows = master.querySelectorAll('tbody tr');
```

demtario (commented):
https://github.com/handsontable/handsontable/pull/13293#pullrequestreview-5138170210 this review stands + conflicts

tests/fixtures/pages/ColumnAutosizeRefillPage.ts
adrianspdev: Fixed in ace3b8e827: `goto()` now calls `awaitBundle(this.page)` from `fixtures/bundle.ts` right after `page.goto(...)`, before the cell assertion, matching the other page objects. Conflicts with `develop` are resolved in the merge commit 422a84c7be (both were in the walkontable `AGENTS.md` and `calculatorFactory.ts` against DEV-2525's `columnHeaderHeightFraction` memo, which now sits inside the same `proposeOnly` guard).

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
adrianspdev: Rewritten in ace3b8e827: the JSDoc now says the flag is captured from THIS draw's pass-1 column band — the one `createCalculators` assigned before the first `renderCellBand`, columns overscan included — and carries an explicit "do not move the capture" note explaining that the overscan case is exactly what the timing relies on. The capture itself is untouched.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
adrianspdev: You're right, corrected in ace3b8e827 in all three places (`MAX_ROWS_BAND_REFILL_PASSES` JSDoc, `AGENTS.md`, `RENDERING-LIFECYCLE.md`). The loop only exits through the cap right after a `renderCellBand` that reported a further change, so the in-band heights are correct and the next draw's band would reach the next stale record, wipe it, and fire the refill again. The text now says the grid is stuck for want of a draw, not information, names a deferred draw on cap exhaustion and an adaptive cap as both workable, and states neither is part of this fix.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
adrianspdev: Good catch — on the legacy measured path the column band's width is `getViewportWidth()` minus `getRowHeaderWidth()`, so the propose-only prediction was reading pass 1's memo while `createCalculators(false)` re-measures the TH. Fixed in ace3b8e827 by resetting `rowHeaderWidth` inside the guard before the prediction, so both builds measure the same header. I kept prediction over reading after assignment: by the time `createCalculators(false)` has assigned the band there's no safe way back (the "never leave a reassigned band without the re-render" rule), whereas the reset costs one extra TH measure, only with `fixedColumnsStart` set. On the snapshot path the width comes from the per-draw `LayoutSnapshot`, so the two already agreed there. The guard is now exported for unit tests and pinned in `test/unit/table/drawCycle.unit.js` (7 tests, including one asserting the memo is `NaN` when the prediction runs). `AGENTS.md` notes the exception to "a declined refill costs no header re-measure".

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
adrianspdev: Whole-band render was the simpler thing to land, not a ruled-out alternative. Rendering only the appended rows needs a partial-band path in the renderer: `renderCellBand` has no row-subset mode, and `markOversizedRows` measures over the whole band on the same pass, so it would have to learn a row range too. I'd rather do that as a follow-up with its own perf scenario than widen this PR. Worth a ticket; I'll open one after this lands and link it here.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
adrianspdev: Done in ace3b8e827: the frozen-rows sync branch now reads `hasExternalRowCalculator` too, so both uses resolve the flag once.

handsontable/src/3rdparty/walkontable/test/spec/table.spec.js
adrianspdev: Replaced in ace3b8e827: the four sites now `await scrollViewportVertically(y)` / `scrollViewportHorizontally(x)` from `test/helpers/common.js`, which resolve on the scroll event instead of a fixed delay. The lint warnings for these four are gone; the older instances in the file are left as they were.

tests/fixtures/pages/ColumnAutosizeRefillPage.ts
adrianspdev: Applied in ace3b8e827: `rendered > 0 && rendered >= visible`, with a comment on why the lower bound is needed (the poll reads its own frame, so an empty TBODY would pass `0 >= 0`).

tests/fixtures/pages/ColumnAutosizeRefillPage.ts
adrianspdev: Applied in ace3b8e827: `master` is typed `HTMLElement | null` and a missing master returns the same `{ gap: Infinity, rendered: 0, visible: 0 }` as a missing last row, so a cold worker fails on the assertion message instead of a `querySelector` TypeError.

adrianspdev:
Round 3 addressed:

- `develop` merged in 422a84c7be (two conflicts, both in walkontable: `AGENTS.md` section order and the `columnHeaderHeightFraction` memo from DEV-2525, now reset inside the same `proposeOnly` guard).
- Review fixes in ace3b8e827, replied per thread. One item (whole-band render per pass) is a rationale reply and a proposed follow-up ticket rather than a change.
- @demtario the 08-28 thread on the cap growth rate was fixed in b30e47ede8 and answered then — could you resolve it if it reads right to you?

Verified locally on the merged tree: core unit (5093), types, full Jasmine E2E (9962 specs, 0 failures), full Playwright suite (7446 passed), walkontable unit + suite (845 specs), React and Vue wrapper tests.

demtario (approved):

budnix (approved):

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
adrianspdev: Follow-up ticket opened as promised: DEV-2908 "Render only the appended rows in a row-band refill pass" (https://app.clickup.com/t/123kvxeag34) — row-range mode for `renderCellBand` and `markOversizedRows`, a renderer-call-count spec, and a wide-grid column-autosize perf scenario so the gain shows in the PR perf comment.

adrianspdev: Implemented in #13481 (draft, DEV-2908): a refill pass now repaints only the rows it appends — `TableRenderer#setPaintWindow`, windowed `resetOversizedRows`/`markOversizedRows`, and `resolveRefillPaintWindow` falling back to a full repaint when the start row, the column band, or a `rowspan` crossing the previous band end says the reused nodes changed identity. The #6452 fixture drops from 92 to 56 cell paints on the shrink draw; a Playwright assertion pins one paint per draw for cell (0, 0) and fails without the window.

Issue #5429, opened 2018-09-30
The 'freeze_column' and 'unfreeze_column' won't work as keys of a custom dropdown menu.


clgerhardt:
### Description
I was attempting to implement the column freezing component into the drop down menu (not the context menu). However, the attempt to do that does not seem to work. I thought based on the documentation the drop down menu would handle freeze column like the context menu would but that does not seem to be the case. 

### Steps to reproduce
1. Open Fiddle
2. Try set a column as frozen 
3. Change the size of the demo box so there will be a scroll bar
4. Then scroll, notice how the column you froze isn't frozen

### Demo
https://jsfiddle.net/4hfa1yo0/3/

### Your environment
* Handsontable version: Pro 6.0.0
* Browser Name and version: Chrome Version 69.0.3497.100 (Official Build) (64-bit)
* Operating System: Windows 10

AMBudnik:
It looks like the key `freeze_column` is not read by the `dropdownMenu`. If we use it in the `contextMenu` it works. Thank you for sharing @clgerhardt 

I do not see any previous versions of Handsontable that supported it.

AMBudnik:
It works the same in v8 https://jsfiddle.net/AMBudnik/1859sbhg/

CHEWX:
Is there anything planned here? I assume the functionality just needs to be added rather than a bug?

adrianszymanski89:
Hi @CHEWX 

This issue still needs to be added to our roadmap. When it changes, we will update this thread.

CHEWX:
Any progress?

adrianszymanski89:
Hi @CHEWX 

Not yet. I will update this topic once it's fixed.

budnix:
Hi @CHEWX

I tested this on the latest version (18.0.0) and the problem is still there. When you pass `freeze_column` or `unfreeze_column` as `dropdownMenu` keys, the menu shows a row with the raw key as its label, and clicking it does nothing.

The reason is that the context menu and the column (dropdown) menu build their item lists from two different hooks. The `ManualColumnFreeze` plugin only registered for the context menu one, so the column menu never learned what those two keys mean.

I prepared a fix in #13294. After it, both keys work in the column menu:

- listing them in `dropdownMenu` gives you working "Freeze column" and "Unfreeze column" entries
- with `manualColumnFreeze: true`, `dropdownMenu: true` also offers both entries by default, the same way the context menu already did

[...]

PR #13294, opened 2026-08-28
DEV-2685: Make freeze_column work as a dropdown menu key (#5429)

budnix:
### Context

The PR fixes `freeze_column` and `unfreeze_column` not working as `dropdownMenu` keys. They work in the `contextMenu`, but not in the column menu — [#5429](https://github.com/handsontable/handsontable/issues/5429), open since 2018.

With `manualColumnFreeze: true` and `dropdownMenu: ['freeze_column', 'unfreeze_column']`, the column menu shows two rows labelled with the raw keys — literally `freeze_column` and `unfreeze_column`, not the translated labels. Clicking a row closes the menu and does nothing. `fixedColumnsStart` stays at `0`. Reproduced on `develop` and on the released 18.0.0 from CDN.

**Cause.** The two menus build their item lists from separate hooks — `afterContextMenuDefaultOptions` and `afterDropdownMenuDefaultOptions` — and `ManualColumnFreeze.enablePlugin()` registered only the first. A key the dropdown menu has never heard of does not raise anything: `ItemsFactory` turns it into a bare `{ name, key }` placeholder with no `callback` and no `hidden()`. So the menu rendered a dead row.

**Fix.** Register the same entry-adder for `afterDropdownMenuDefaultOptions`. The shared private method is renamed `#addMenuEntries`, since it now serves both menus.

#### One additive behavior change

The hook cannot express "resolvable key, but not in the defaults" — `ItemsFactory.setPredefinedItems` rebuilds `defaultOrderPattern` from the array the hook is handed. So with `manualColumnFreeze: true`, `dropdownMenu: true` now also offers Freeze/Unfreeze column.

This is not a breaking change: the `dropdownMenu` default *value* is unchanged, and the breaking-changes policy does not cover additive menu entries. It matches what the context menu has always done — `contextMenu: true` already carries both entries, hidden until the selection is a single column — and `Filters` is the existing precedent for a plugin contributing items to the default dropdown menu.

#### Scope

`manualColumnFreeze` only. Eight other plugins have the same gap (`comments`, `customBorders`, `copyPaste`, `exportFile`, `mergeCells`, `hiddenRows`, `hiddenColumns`, `nestedRows`) — `copy` is dead in the dropdown menu too. A blanket fix would push row-oriented entries into a column header menu, which is a product decision rather than a bug fix, so it is left for a separate task.

### How has this been tested?

New Playwright spec with 9 tests, plus a fixture holding four grids: the reported config, `dropdownMenu: true`, the `contextMenu` path as a regression control, and the plugin-disabled case.

The spec was run against both builds, which is what proves it tests the fix rather than the code:

- Fix removed and rebuilt: **7 of 9 failed**. The 2 that passed are the deliberate controls — the context menu path and the plugin-disabled guard, neither of which should change.
- Fix restored and rebuilt: **9 of 9 passed**.

One test needed strengthening after that check. "Offers unfreeze only once the column is frozen" passed on the broken build for the wrong reason: the placeholder row was labelled `unfreeze_column`, so asserting the absence of `Unfreeze column` succeeded without proving anything. It now also asserts the raw key is absent, and fails on the broken build.

```bash
# New Playwright spec
cd tests && npx playwright test --project=e2e-main dropdown-menu-freeze-column

# Existing suites touched by the change — all green
npm run test:e2e --prefix handsontable -- --testPathPattern=dropdownMenu      # 196 specs
npm run test:e2e --prefix handsontable -- --testPathPattern=contextMenu       # 596 specs
npm run test:e2e --prefix handsontable -- --testPathPattern=manualColumnFreeze # 27 specs
npm run test:unit --prefix handsontable                                       # 4043 tests

# Lint
```

Manually verified in Chrome as well: opening the column menu selects the whole column, so "Freeze column" shows and "Unfreeze column" stays hidden until the column is frozen. Freezing moves the column into the inline-start overlay and leaves the columns past the freeze point untouched.

### Types of changes

### Related issue(s):

1. DEV-2685
2. #5429

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

The column menu guide now notes that the freeze entries need the `ManualColumnFreeze` plugin, and the column freezing guide names the column menu next to the context menu.

The changelog entry is `.changelogs/5429.json`. It is named after the public issue rather than the PR number because `issuesOrigin` is `public`, per the mapping in `.changelogs/README.md`.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2685

---

> [!NOTE]
> **Medium Risk**
> Changes column menu lifecycle (rebuild on every open) and default menu contents when both `dropdownMenu: true` and `manualColumnFreeze` are enabled; core menu/command paths are touched but heavily tested.
> 
> **Overview**
> Fixes **#5429**: `freeze_column` / `unfreeze_column` in a custom column menu showed raw keys and did nothing because **ManualColumnFreeze** only hooked `afterContextMenuDefaultOptions`. The plugin now registers the same entry builder on **`afterDropdownMenuDefaultOptions`** (ordered after Filters), so those keys resolve with labels, callbacks, and `hidden()` behavior.
> 
> **DropdownMenu** no longer freezes its item list at enable time: **`prepareMenuItems()`** runs on each **`open()`** (and on **`executeCommand()`** when a command is missing and the menu is closed), matching ContextMenu so `updateSettings` toggles show up on the next open. **`disablePlugin`** clears **`this.menu`** after destroy.
> 
> Freeze/unfreeze items add **`disabled()`** plus shared **`isPluginOff()`** so stale **`executeCommand`** paths cannot run after the plugin is turned off. With **`manualColumnFreeze: true`** and **`dropdownMenu: true`**, Freeze/Unfreeze are added to the default column menu (documented in changelogs/guides). Playwright e2e coverage and **AGENTS.md** notes on dual-hook registration are included.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2831e25fc444562c5d0e3ef33cc85d1f68b96d6d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2685 Make the freeze_column and unfreeze_column items work in the dropdown menu (#5429)](https://app.clickup.com/t/9015210959/DEV-2685)

handsontable/src/plugins/manualColumnFreeze/manualColumnFreeze.ts
```diff
@@ -65,12 +66,19 @@ export class ManualColumnFreeze extends BasePlugin {
...
-      this.#addContextMenuEntry(options as Record<string, unknown>);
-    });
+    this.addHook('afterContextMenuDefaultOptions', this.#onAfterMenuDefaultOptions);
+    // The dropdown menu builds its items from a separate hook, so the entries have to be added
+    // twice. Without this the `freeze_column` / `unfreeze_column` keys resolve to inert
+    // placeholder rows there. See issue #5429.
+    //
+    // The order index runs this after the callbacks registered at the default index, which keeps
+    // the entries below the Filters interface — Filters makes up the bulk of the column menu.
+    this.addHook('afterDropdownMenuDefaultOptions', this.#onAfterMenuDefaultOptions, 1);
```
Haxikowy: `predefinedItems/index.ts` exports `addItem(key, factory)`, which registers a key in the module-level registry that every `ItemsFactory` seeds from at construction (`itemsFactory.ts:22`). DropdownMenu's first resolution pass runs against that registry before the hook fires (`dropdownMenu.ts:222-225`), and the array branch at `itemsFactory.ts:139` looks names up there. So `dropdownMenu: ['freeze_column']` resolves while `dropdownMenu: true` keeps taking its pattern from `DEFAULT_ITEMS` and gains nothing.

That fixes #5429 with no change to anyone's default column menu, and with no `#refreshDropdownMenu()` at all: `isItemHidden` is re-evaluated on every open (`menu/menu.ts:427`), so an enabled-aware `hidden()` covers the toggle case the rebuild exists for. Prerequisites: the item factories resolve the plugin via `this.getPlugin('manualColumnFreeze')` instead of the closed-over instance, and `hidden()` returns true when it is disabled. Honest cost: `addItem` has no callers today, so it is untested in-tree.

Worth a decision before the rest of the review lands, since it removes the additive behavior change and the cross-plugin lifecycle call together.

Haxikowy: On the order index: a bare `1` encodes a dependency on Filters (priority 250 against this plugin's 110) with nothing on Filters' side pointing back. A named constant, or a comment citing `filters.ts:363`, stops the next person retuning it blind.

It also inverts this plugin's position relative to user callbacks on the same public hook. Anyone registering `afterDropdownMenuDefaultOptions` at the default index now runs before the entries exist, so that is no longer a way to strip them. If the additive change stays, the working opt-outs (enumerate items instead of `true`, or `beforeDropdownMenuSetItems`) belong in the column-menu guide.

handsontable/src/plugins/manualColumnFreeze/manualColumnFreeze.ts
```diff
@@ -65,12 +66,19 @@ export class ManualColumnFreeze extends BasePlugin {
...
     this.addHook('beforeColumnMove', this.#onBeforeColumnMove);
 
     super.enablePlugin();
+
+    this.#refreshDropdownMenu();
```
Haxikowy: Neither `hidden()` nor `disabled()` on the two items checks whether this plugin is enabled (`contextMenuItem/freezeColumn.ts:24`, `unfreezeColumn.ts:24`), and `commandExecutor.commands` is write-only, with no eviction anywhere in the class. So after `updateSettings({ manualColumnFreeze: false })`, `getPlugin('dropdownMenu').executeCommand('freeze_column')` still moves the column and bumps `_fixedColumnsStart`. `execute()` gates on `disabled`, not `hidden`, so the guard has to go on both.

Adding it pays for itself twice: it closes that path, and it makes the refresh in `disablePlugin()` unnecessary, which halves the churn described below.

handsontable/src/plugins/manualColumnFreeze/manualColumnFreeze.ts
```diff
@@ -80,6 +88,8 @@ export class ManualColumnFreeze extends BasePlugin {
     this.#afterFirstUse = false;
 
     super.disablePlugin();
+
+    this.#refreshDropdownMenu();
```
Haxikowy: With the refresh in both `enablePlugin()` and `disablePlugin()`, `updatePlugin()` (`:102-103`) rebuilds the whole DropdownMenu twice per `updateSettings({ manualColumnFreeze })`, the first rebuild deliberately entry-less and discarded. Three rebuilds if `dropdownMenu` is in the same payload, since this plugin runs at priority 110 and DropdownMenu at 230.

Each cycle destroys the `Menu` and its nested Handsontable, drops and re-adds the shortcut context and the root click listener, and re-fires both dropdown hooks at every listener. New side effect: `updateSettings({ manualColumnFreeze: true })` now closes an open dropdown menu.

`contextMenu.ts:405` `prepareMenuItems()` does what you want here, re-running the hook and calling `setMenuItems()` on the same `Menu`. Extracting that from `dropdownMenu.ts:242-276` gives you an items-only refresh, and `filters.ts:373-376` could move to it too.

handsontable/src/plugins/manualColumnFreeze/manualColumnFreeze.ts
```diff
@@ -167,12 +177,48 @@ export class ManualColumnFreeze extends BasePlugin {
...
+   *
+   * Does nothing during the initial setup, where the menu is not built yet — the dropdown menu
+   * plugin assembles it once every plugin is ready.
+   *
+   * @private
+   */
+  #refreshDropdownMenu() {
+    const dropdownMenuPlugin = this.hot.getPlugin('dropdownMenu');
+
+    if (!dropdownMenuPlugin?.enabled || !dropdownMenuPlugin.menu) {
```
Haxikowy: The guard reads the runtime flag but not the setting, and `DropdownMenu.enablePlugin` only checks `this.enabled` too. On `updateSettings({ manualColumnFreeze: true, dropdownMenu: false })` this plugin runs first, so the menu is rebuilt from the already-applied `dropdownMenu: false` and both dropdown hooks fire with that throwaway list before DropdownMenu disables itself.

It self-corrects only because `afterUpdateSettings` listeners run in ascending priority order, which makes it a trap for any future priority change.

```suggestion
    if (!dropdownMenuPlugin?.enabled || !dropdownMenuPlugin.isEnabled() || !dropdownMenuPlugin.menu) {
```

.changelogs/5429.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed the `freeze_column` and `unfreeze_column` items not working in the column (dropdown) menu. When `manualColumnFreeze` is enabled, `dropdownMenu: true` now offers both entries as well.",
```
Haxikowy: One `fixed` entry carries both the key resolution and the default-menu expansion. `.changelogs/README.md` names this pair directly: a public issue fixed alongside a behavior change is `fixed` plus `changed`. As written, someone scanning **Fixed** in the release notes never learns their default column menu grew two rows.

Keep this one as `fixed` with the first sentence, and add `13294.json` with `type: "changed"` for the `dropdownMenu: true` expansion. The PR's "Types of changes" checkboxes need "New feature or improvement" ticked for the same reason. Both moot if the `addItem` route is taken.

Haxikowy (commented):
Reviewed across correctness, plugin lifecycle, regression against untouched neighbours, API surface and breaking-changes policy, test quality, and requirements. The fix is real and the red/green discipline behind the spec is the right instinct. Two things need a decision before this merges.

First, the stated justification for the side effect does not hold. `addItem` in the predefined-items registry makes the keys resolvable without touching `dropdownMenu: true` and without the cross-plugin `disablePlugin()/enablePlugin()`. Details in the thread on `manualColumnFreeze.ts:76`. Your reading of the hook is correct, it just is not the only door.

Second, if the additive change stays, the changelog needs a second `changed` entry per `.changelogs/README.md`.

Two items with no line to attach to. `handsontable/AGENTS.md` should record what this PR discovered, since eight plugins have the bug and the next person will write it again: a plugin contributing menu items must register on both hooks, because an unresolved key degrades to an inert `{ name, key }` placeholder rather than raising, and the dropdown menu assembles its list once on enable while the context menu rebuilds per open. And the eight other plugins need a follow-up task linked to DEV-2685, which currently has no linked tasks, or the known-dead dropdown keys go with this ticket when it closes.

One more, worth its own ticket rather than this PR: `unfreeze_column`'s `hidden()` only tests whether a column sits inside the frozen area, not who froze it, so `fixedColumnsStart: 2` plus `manualColumnFreeze: true` now offers Unfreeze in the default column menu on columns the developer fixed in config.

The test evidence in the description cites counts that do not match the files: 9 tests and four grids against the real 12 and six, so "7 of 9 failed" cannot be checked. Worth re-deriving, because the check itself is the most convincing part of the submission.

What I verified rather than assumed, in case it saves the argument: Filters' keyboard focus navigator does not break, because it re-binds on `beforeDropdownMenuShow`; the filter component DOM survives `menu.destroy()` by re-parenting, and filter data state is untouched since `Filters.SETTING_KEYS` does not match a `manualColumnFreeze` payload; no orphan separator can render, because hidden items are filtered before `filterSeparators()`; `updatePlugin()` fires only for payloads naming `manualColumnFreeze`, not every `updateSettings`; custom array and object `dropdownMenu` configs get no injection, since `getItems` never reads `defaultOrderPattern` for a user pattern; no existing spec, visual baseline, docs example, or wrapper is affected; hook index 1 does land after Filters via the stable ordered insert; and the ContextMenu path is unchanged. `breaking: false` is honest, and E2E is the right tier here.

I did not run the suites, only read the code and the CI results on 85f4165.

tests/e2e/dropdown-menu-freeze-column.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+
+    test('picks the entry up when the plugin is turned on', async () => {
+      await grid.setManualColumnFreeze(TOGGLE, false);
+      await grid.openColumnMenu(TOGGLE, 'Charlie');
+      expect(await grid.visibleDropdownMenuItems()).not.toContain(FREEZE_LABEL);
+
+      await grid.setManualColumnFreeze(TOGGLE, true);
+      await grid.openColumnMenu(TOGGLE, 'Charlie');
+
+      expect(await grid.visibleDropdownMenuItems()).toContain(FREEZE_LABEL);
```
Haxikowy, line 230: Every membership check in the file is `toContain`, which passes just as happily with two "Freeze column" rows. Duplicate registration is the specific risk of putting one handler on two hooks with two rebuild paths, and nothing here would catch it.

```suggestion
      expect((await grid.visibleDropdownMenuItems()).filter(item => item === FREEZE_LABEL)).toHaveLength(1);
```

Two gaps in the same area: no case pins that a custom `dropdownMenu: ['col_left', 'col_right']` with the plugin on stays clean (the property reviewers will doubt most), and none covers a toggle while the menu is open, which now destroys the menu DOM under the user.

tests/e2e/dropdown-menu-freeze-column.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+    test('leaves both keys as inert raw-key rows', async () => {
+      await grid.openColumnMenu(PLUGIN_OFF, 'Charlie');
+
+      const items = await grid.visibleDropdownMenuItems();
+
+      // The entries belong to the plugin, so with it off nothing resolves the keys and ItemsFactory
+      // emits its placeholder for each — the very rows the whole bug consisted of. Pinned as the
+      // real behavior rather than asserting the translated label is absent, which would pass on the
+      // broken build too and prove nothing.
+      expect(items).toEqual(['freeze_column', 'unfreeze_column']);
```
Haxikowy, line 306: This pins the reported symptom as expected output. With the plugin off, users see literal `freeze_column` and `unfreeze_column` rows, which is what #5429 complained about, and the assertion passes identically on the unfixed build.

Defensible as a deliberate pin of ItemsFactory's unknown-key behavior, but then say so in the test name so it does not read as coverage of this fix. The `addItem` route would let those rows be hidden instead.

docs/content/guides/accessories-and-menus/column-menu/column-menu.md
```diff
@@ -77,6 +77,8 @@ To enable the plugin, set the [`dropdownMenu`](@/api/options.md#dropdownmenu) co
 
 To use the default dropdown contents, set it to `true`, or to customize it by setting it to use a custom list of actions. For the available entry options reference, see the [Context Menu demo](@/guides/accessories-and-menus/context-menu/context-menu.md#page-specific).
 
+Some entries come from other plugins, and work only when their plugin is enabled. The `freeze_column` and `unfreeze_column` entries require the [`ManualColumnFreeze`](@/api/manualColumnFreeze.md) plugin. When you enable it, both entries also appear in the default dropdown contents.
```
Haxikowy: "Both entries also appear" overstates it. `unfreeze_column`'s `hidden()` suppresses it until a column is frozen, so a reader following this sentence looks for a row that is not there.

Also worth swapping the `menuTag: updated` in `column-freezing.md:21`: that page got a one-clause edit, which `docs/AGENTS.md` says to leave untagged, while this page got a new paragraph and no tag.

handsontable/src/plugins/manualColumnFreeze/manualColumnFreeze.ts
budnix: You're right that my justification was wrong — I checked with a probe, and addItem does resolve the keys without touching the default menu. Went one level down instead in 9421c17310: DropdownMenu now rebuilds its items on open like ContextMenu, so the hook alone is enough and no registry or lifecycle call is needed.

budnix: Good spot — it is a named AFTER_FILTERS_ORDER_INDEX now with the reasoning next to it in 9421c17310, and the two opt-outs are written into the column-menu guide.

handsontable/src/plugins/manualColumnFreeze/manualColumnFreeze.ts
budnix: Nice catch, and you were right that the rebuild alone does not close it. Guarded both hidden() and disabled() in 9421c17310, with a test that drives executeCommand directly.

handsontable/src/plugins/manualColumnFreeze/manualColumnFreeze.ts
budnix: Confirmed the open-menu close before changing anything. Both rebuilds are gone in 9421c17310, replaced by the per-open prepareMenuItems() you pointed at.

handsontable/src/plugins/manualColumnFreeze/manualColumnFreeze.ts
budnix: Fair point — that guard went away together with the rebuild in 9421c17310.

.changelogs/5429.json
budnix: Agreed — 5429.json is the fix only now, and 13294.json carries the default-menu change as a changed entry, both in 9421c17310.

tests/e2e/dropdown-menu-freeze-column.spec.ts
budnix: Good point — added exact-count assertions plus the custom-list and toggle-while-open cases in 9421c17310.

tests/e2e/dropdown-menu-freeze-column.spec.ts
budnix: Yep, my mistake — renamed it in 9421c17310 so it reads as pinning ItemsFactory's unknown-key fallback rather than as coverage of this fix.

docs/content/guides/accessories-and-menus/column-menu/column-menu.md
budnix: Right about the wording — reworded in 9421c17310 to say each entry shows only when it applies. On the tag though, column-menu.md:18 already carried menuTag: updated before this PR, so it was not missing.

Haxikowy (dismissed):

tests/e2e/dropdown-menu-freeze-column.spec.ts
```diff
@@ -0,0 +1,313 @@
...
+
+      expect(await grid.fixedColumnsStart(DEFAULT_MENU)).toBe(1);
+      expect(await grid.columnHeaders(DEFAULT_MENU)).toEqual(
+        ['Charlie', 'Alpha', 'Bravo', 'Delta', 'Echo', 'Foxtrot']
+      );
+    });
+  });
+
+  test.describe('toggling manualColumnFreeze after the menu was built', () => {
+    // The dropdown menu assembles its items once, when it is enabled, unlike the context menu
```
demtario: Medium: stale comment. Says the dropdown menu "assembles its items once, when it is enabled, unlike the context menu which rebuilds on every open." That describes an earlier design in this branch's history. The shipped behavior rebuilds on every `open()` (see `dropdownMenu.ts` around the `prepareMenuItems`/`open` docstring, and `contextMenu/AGENTS.md`), matching ContextMenu. It also contradicts a correct comment ~60 lines later in this same file. Update this comment to describe the final rebuild-on-open behavior.

tests/fixtures/demo/dropdown-menu-freeze-column.html
```diff
@@ -0,0 +1,168 @@
...
+<body>
+  <!--
+    GitHub #5429. `freeze_column` and `unfreeze_column` worked as `contextMenu` keys but not as
+    `dropdownMenu` keys. The two menus build their items from separate hooks, and
+    ManualColumnFreeze registered only `afterContextMenuDefaultOptions`. An unknown key falls
+    through ItemsFactory as a bare `{ name, key }` placeholder, so the dropdown menu rendered a row
+    labelled with the RAW KEY that did nothing when clicked.
+
+    The dropdown menu also builds that list only once, when it is enabled, so the grids below cover
+    the settings toggle in both directions as well as the menu contents.
```
demtario: Medium: same stale claim duplicated here — "The dropdown menu also builds that list only once, when it is enabled..." This no longer matches the shipped rebuild-on-open behavior (see the linked note on `tests/e2e/dropdown-menu-freeze-column.spec.ts:144`). Update to match.

handsontable/src/plugins/manualColumnFreeze/manualColumnFreeze.ts
```diff
@@ -11,12 +12,23 @@ Hooks.getSingleton().register('afterColumnUnfreeze');
 export const PLUGIN_KEY = 'manualColumnFreeze';
```
demtario, line 12: Low, informational: `AFTER_FILTERS_ORDER_INDEX` hardcodes ordering knowledge of the Filters plugin's priority into ManualColumnFreeze via a magic hook-order index. Verified correct and covered by the new E2E test, not a bug — just flagging that if Filters' registration order/index on `afterDropdownMenuDefaultOptions` ever changes, or a third plugin registers on the same hook, this could silently drift with no compile-time signal.

demtario (commented):
3 inline notes below.

tests/e2e/dropdown-menu-freeze-column.spec.ts
budnix: Good spot — that one described the design I replaced earlier in the branch. Rewritten in c80e088a3d to say the list rebuilds on every open.

tests/fixtures/demo/dropdown-menu-freeze-column.html
budnix: You're right, the same stale line. Fixed in c80e088a3d alongside the spec comment.

handsontable/src/plugins/manualColumnFreeze/manualColumnFreeze.ts
budnix: Fair point, and it is a real trade-off. The E2E ordering assertion is what would catch a drift, and a declared position on the menu item would be the proper fix — noted for a follow-up.

demtario (approved):

PR #13295, opened 2026-08-28
DEV-2677: Ignore non-left clicks on the autocomplete cell arrow

demtario:
### Context

The PR fixes a right-click on an `autocomplete` or `dropdown` cell's arrow opening the cell editor.

**Read this before the diff — the reported symptom and the fix are not the same thing.**

DEV-2677 was filed as "clicking an empty cell then pressing <kbd>Enter</kbd> opens the editor and closes it again on the same keypress". Root-causing it in a real browser with trusted events showed the premise was wrong: **the click opens the editor, not the Enter.**

`autocompleteRenderer` registers a `mousedown` listener on `rootElement` that calls `onCellDblClick` whenever the event target carries `htAutocompleteArrow`, so pressing the arrow opens the list immediately. The Enter that follows arrives with the shortcut context already `'editor'` and correctly commits and closes. Captured trace, one trusted click then one trusted Enter on an empty `dropdown` cell:

```
DOM mousedown target=DIV.htAutocompleteArrow
beginEditing(null, null) state=STATE_VIRGIN
   via DropdownEditor.beginEditing < EditorManager.openEditor < #onCellDblClick < Object.onCellDblClick
afterBeginEditing        ctx=editor state=STATE_EDITING opened=true active=TEXTAREA.handsontableInput
DOM keydown "Enter" target=TEXTAREA.handsontableInput ctx=editor state=STATE_EDITING opened=true
finishEditing(false, false) state=STATE_EDITING
   via EditorManager.closeEditor < EditorManager.closeEditorAndSaveChanges < editorCloseAndSaveByEnter
beforeChange(edit) [[0,0,"",""]] state=STATE_WAITING
afterChange(edit)  [[0,0,"",""]] state=STATE_VIRGIN
```

**Why an empty value and not a seeded one — it is geometry, six pixels, not a state-machine difference.** The arrow is 16 px wide and right-floated. Measured:

| cell | column width | arrow span (px from TD left) | click centre | `elementFromPoint(centre)` |
|---|---|---|---|---|
| empty | 50 (the default minimum) | 24 – 40 | 25 | `DIV.htAutocompleteArrow` |
| seeded `'Al'` | 56 (content widens it) | 30 – 46 | 28 | `TD.htAutocomplete` |

Everything the task reported follows: click-only (`selectCell` fires no mousedown on the arrow), `autocomplete`/`dropdown` only (only they render an arrow), array or function `source` alike, `document.activeElement === TEXTAREA` after the click (`beginEditing` → `focus()`), and the exact hook order — `afterBeginEditing` from the *click*, `beforeChange`/`afterChange` from the *Enter*.

**So the task's definition of done is deliberately not met, and that is the recommendation.** Opening the list when the user clicks its arrow is an intended, long-standing affordance, so "a single Enter must open and keep the editor open" cannot be delivered by changing the Enter path — the click got there first. Removing the affordance to satisfy the wording would be a breaking change.

What the investigation did surface is a real defect in that same listener: it tests the target's class and never the mouse button, so **a right-click on the arrow opens the editor** — on a grid with a context menu, the user gets the editor and the menu at once. Walkontable restricts its own double-click-to-open path to the left button (`3rdparty/walkontable/src/event.ts`, pinned by its `only for LMB` spec); this path now does the same. That is the whole source diff: five lines.

`tests/AGENTS.md` carried the symptom as a fixture-authoring warning ("seed editor cells with a non-empty value"). That advice worked only by accident — a long header or narrow content puts the arrow back under the cell centre even on a seeded cell — so the bullet is rewritten to state the real mechanism and to recommend clicking off-centre or using `selectCell()`. The shared-prefix seeding advice is kept and split out, since it exists for an unrelated and still-valid reason.

Two findings recorded here rather than changed:

1. `autocompleteRenderer` builds its listener once, closing over the first-rendered cell's `row, col`. Unobservable today — `EditorManager#onCellDblClick` reads only `coords.isCell()` and `openEditor()` uses the live selection, verified by clicking cell (2,1)'s arrow and getting (2,1). Left out of this diff; worth its own task.
2. A real touch tap on the arrow synthesizes no `mousedown` at all, so the arrow has never opened the editor on mobile. Pre-existing — verified identical with and without this guard.

### How has this been tested?

New Playwright coverage, `tests/e2e/autocomplete-arrow-button.spec.ts`, parametrised over both editors. It asserts both halves: a right-button press leaves the editor closed, **and** a left-button press still opens it — a case that only checked the right button would pass on a change that deleted the affordance. The negative case anchors on the selection landing and then asserts once without retrying, because `expect.poll` retries until its condition holds and would report success on an editor that opened a tick later.

Confirmed red before the fix and green after:

```bash
# against the pre-fix bundle: the two right-button cases fail
npm --prefix tests run test:e2e -- autocomplete-arrow-button
npm --prefix tests run test:e2e -- autocomplete-arrow-button   # 4 passed
```

Everything else that was run:

```bash
npm --prefix handsontable run eslint                 # 0 errors
npm --prefix tests run lint                          # clean
npm --prefix handsontable run test:unit              # pass
npm --prefix handsontable run test:e2e -- --testPathPattern="autocompleteRenderer|autocompleteEditor|dropdownEditor|handsontableEditor"
                                                     # 245 specs, 0 failures
npm --prefix tests test                              # all six theme x bundle legs: 3304 passed
```

The full Playwright matrix also reported 2 failures in `e2e/fill-handle-formulas.spec.ts` on the two horizon legs. They are **pre-existing and unrelated**: that fixture configures no `autocomplete`/`dropdown` column, so this listener never registers for it, and disabling the guard in the built bundle reproduces the identical failure.

Manual verification in Chrome through the DevTools protocol, on a nine-configuration diagnostic page with `BaseEditor.prototype.beginEditing`/`finishEditing` wrapped: `text`, `autocomplete`, `autocomplete + strict`, and `dropdown`, each empty and seeded, at default and narrow column widths, plus a two-column grid to check which cell the arrow opens.

### Types of changes

### Related issue(s):

1. DEV-2677

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2677

---

> [!NOTE]
> **Low Risk**
> Small, localized input-handling guard on an existing mousedown path; behavior for left-click and touch-compat mousedown is preserved, with new e2e tests locking both open and stay-closed cases.
> 
> **Overview**
> **Right-clicking the dropdown arrow on autocomplete-style cells no longer opens the cell editor**, aligning with Walkontable’s left-button-only double-click behavior and avoiding the editor and context menu appearing together.
> 
> The change is in `autocompleteRenderer`: the root `mousedown` handler that opens the list when the target is `.htAutocompleteArrow` now also requires `isLeftClick(event)`. Left-click on the arrow is unchanged.
> 
> **Test and docs follow-up:** Playwright coverage in `cell-dropdown-arrow-button.spec.ts` runs over `autocomplete`, `dropdown`, and `handsontable` (all use the same arrow listener). A dedicated fixture and `CellDropdownArrowPage` support those cases. `tests/AGENTS.md` documents that a centred cell click can hit the arrow (theme-dependent geometry); `AutocompleteAsyncSourcePage.openEditor` clicks near the leading edge so Enter-based flows stay stable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 18bdda1c474cb546be18467a08d4c511ea195e6c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/handsontable/src/renderers/autocompleteRenderer/autocompleteRenderer.ts
+++ b/handsontable/src/renderers/autocompleteRenderer/autocompleteRenderer.ts
@@ -4,6 +4,7 @@ import { htmlRenderer } from '../htmlRenderer';
 import { textRenderer } from '../textRenderer';
 import EventManager from '../../eventManager';
 import { addClass, eventTargetEl, hasClass } from '../../helpers/dom/element';
+import { isLeftClick } from '../../helpers/dom/event';
 import { A11Y_HIDDEN } from '../../helpers/a11y';
 
 export const RENDERER_TYPE: 'autocomplete' = 'autocomplete';
@@ -55,7 +56,10 @@ export function autocompleteRenderer(
 
     // not very elegant but easy and fast
     hotInstance.acArrowListener = function(event: Event) {
-      if (hasClass(eventTargetEl(event)!, 'htAutocompleteArrow')) {
+      // Only the left button opens the list. Walkontable applies the same restriction to its own
+      // double-click-to-open path, and without it a right-click on the arrow opens the editor
+      // alongside the context menu.
+      if (isLeftClick(event) && hasClass(eventTargetEl(event)!, 'htAutocompleteArrow')) {
         hotInstance.view._wt.getSetting('onCellDblClick', null, hotInstance._createCellCoords(row, col), TD);
       }
     };
--- a/handsontable/src/renderers/autocompleteRenderer/autocompleteRenderer.ts
+++ b/handsontable/src/renderers/autocompleteRenderer/autocompleteRenderer.ts
@@ -62,6 +62,10 @@ export function autocompleteRenderer(
       // hatch; this path needs none, because a tap reaches it only as a compatibility `mousedown`,
       // which carries `button === 0` like any other left press.
       if (isLeftClick(event) && hasClass(eventTargetEl(event)!, 'htAutocompleteArrow')) {
+        // The `null` event is load-bearing, not laziness: `EditorManager#openEditor` only applies
+        // its "no editor for a multi-cell selection" default when the event is a `MouseEvent`, so
+        // forwarding the real one here would stop the arrow from opening the list after a
+        // shift-drag range. Changing that is a behavior change, not a cleanup.
         hotInstance.view._wt.getSetting('onCellDblClick', null, hotInstance._createCellCoords(row, col), TD);
       }
     };
--- a/handsontable/src/renderers/autocompleteRenderer/autocompleteRenderer.ts
+++ b/handsontable/src/renderers/autocompleteRenderer/autocompleteRenderer.ts
@@ -56,9 +56,11 @@ export function autocompleteRenderer(
 
     // not very elegant but easy and fast
     hotInstance.acArrowListener = function(event: Event) {
-      // Only the left button opens the list. Walkontable applies the same restriction to its own
+      // Only the left button opens the list. Walkontable applies the same button check to its own
       // double-click-to-open path, and without it a right-click on the arrow opens the editor
-      // alongside the context menu.
+      // alongside the context menu. Walkontable pairs that check with a `touchApplied` escape
+      // hatch; this path needs none, because a tap reaches it only as a compatibility `mousedown`,
+      // which carries `button === 0` like any other left press.
       if (isLeftClick(event) && hasClass(eventTargetEl(event)!, 'htAutocompleteArrow')) {
         hotInstance.view._wt.getSetting('onCellDblClick', null, hotInstance._createCellCoords(row, col), TD);
       }
--- a/tests/e2e/autocomplete-arrow-button.spec.ts
+++ b/tests/e2e/autocomplete-arrow-button.spec.ts
@@ -1,0 +1,45 @@
+import { test, expect } from '../fixtures/test';
+import { AutocompleteAsyncSourcePage } from '../fixtures/pages/AutocompleteAsyncSourcePage';
+
+const EDITORS = ['autocomplete', 'dropdown'] as const;
+
+/**
+ * DEV-2677. `autocompleteRenderer` registers a `mousedown` listener that opens the editor when the
+ * press lands on a cell's dropdown arrow. It tested only the target's class and never the mouse
+ * button, so a right-click on the arrow opened the editor - on a grid with a context menu, that
+ * handed the user the editor and the menu at once. Walkontable restricts its own
+ * double-click-to-open path to the left button; this path now does the same.
+ *
+ * Both halves are asserted every time. A case that only checked the right button would pass on a
+ * "fix" that deleted the affordance, and opening the list from its arrow is behavior users have
+ * relied on for a decade.
+ *
+ * The negative case anchors on the selection rather than polling for the editor to stay closed:
+ * `expect.poll` retries until its condition holds, so it would report success on an editor that
+ * opened one tick later. A right-button press selects the cell, so waiting for the selection to
+ * land gives a deterministic point at which a single, non-retrying assertion is meaningful.
+ */
+EDITORS.forEach((editor) => {
+  test.describe(`${editor} editor dropdown arrow`, () => {
+    let grid: AutocompleteAsyncSourcePage;
+
+    test.beforeEach(async({ page, theme, bundle }) => {
+      grid = new AutocompleteAsyncSourcePage(page, theme, bundle, { editor });
+      await grid.goto();
+    });
+
+    test('opens the editor when the arrow is pressed with the left button', async() => {
+      await grid.arrow(0, 0).click();
+
+      await expect.poll(() => grid.isEditorOpen()).toBe(true);
+    });
+
+    test('leaves the editor closed when the arrow is pressed with the right button', async() => {
+      await grid.arrow(0, 0).click({ button: 'right' });
+
+      await expect.poll(() => grid.selected()).toEqual([[0, 0, 0, 0]]);
+
+      expect(await grid.isEditorOpen()).toBe(false);
+    });
+  });
+});
--- a/tests/e2e/autocomplete-arrow-button.spec.ts
+++ b/tests/e2e/autocomplete-arrow-button.spec.ts
@@ -28,10 +28,19 @@ EDITORS.forEach((editor) => {
       await grid.goto();
     });
 
+    // Cell (1, 1) rather than (0, 0), and the editor's own coordinates rather than just "an editor
+    // is open". To be clear about what that does and does not cover: the renderer builds its arrow
+    // listener once and closes over the FIRST rendered cell's coordinates, and no assertion here
+    // can catch that today - the pair reaches only `isCell()`, and `prepareEditor()` takes the
+    // edited cell from the live selection, so the editor's coordinates and `getSelected()` are the
+    // same state. Clicking a cell that is not the first one, and reading the coordinates the editor
+    // was actually prepared with, is what would surface it if a later change starts honoring the
+    // coordinates the listener passes. It does catch an editor reused without a re-`prepare()`.
     test('opens the editor when the arrow is pressed with the left button', async() => {
-      await grid.arrow(0, 0).click();
+      await grid.arrow(1, 1).click();
 
       await expect.poll(() => grid.isEditorOpen()).toBe(true);
+      expect(await grid.editorCoords()).toEqual([1, 1]);
     });
 
     test('leaves the editor closed when the arrow is pressed with the right button', async() => {
--- a/tests/fixtures/pages/AutocompleteAsyncSourcePage.ts
+++ b/tests/fixtures/pages/AutocompleteAsyncSourcePage.ts
@@ -95,6 +95,14 @@ export class AutocompleteAsyncSourcePage {
   }
 
   /**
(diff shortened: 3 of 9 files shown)


qunabu:
Task linked: [DEV-2677 Autocomplete/dropdown: clicking an empty cell then pressing Enter closes the editor on the same keypress](https://app.clickup.com/t/9015210959/DEV-2677)

handsontable/src/renderers/autocompleteRenderer/autocompleteRenderer.ts
```diff
@@ -55,7 +56,12 @@ export function autocompleteRenderer(
 
     // not very elegant but easy and fast
     hotInstance.acArrowListener = function(event: Event) {
-      if (hasClass(eventTargetEl(event)!, 'htAutocompleteArrow')) {
+      // Only the left button opens the list. Walkontable applies the same button check to its own
+      // double-click-to-open path, and without it a right-click on the arrow opens the editor
+      // alongside the context menu. Walkontable pairs that check with a `touchApplied` escape
+      // hatch; this path needs none, because a tap reaches it only as a compatibility `mousedown`,
+      // which carries `button === 0` like any other left press.
+      if (isLeftClick(event) && hasClass(eventTargetEl(event)!, 'htAutocompleteArrow')) {
```
budnix, line 64: `isLeftClick` is `button === 0`, so this also stops middle-click and the back/forward buttons from opening the list. It also stops a `mousedown` sent as a plain `Event`, which has no `button` at all — that used to pass. The changelog title only promises right-click, so either the title should widen or the guard should narrow to `!isRightClick(event)`. Was the wider block the intent?

handsontable/src/renderers/autocompleteRenderer/autocompleteRenderer.ts
```diff
@@ -55,7 +56,12 @@ export function autocompleteRenderer(
...
         hotInstance.view._wt.getSetting('onCellDblClick', null, hotInstance._createCellCoords(row, col), TD);
```
budnix, line 69: Worth a note for whoever tidies this later — the `null` here is doing real work. `openEditor` has an `event instanceof MouseEvent` branch that blocks opening when the selection is more than one cell, so passing the real event would stop the arrow from opening the editor after a shift-drag range. That might be the more consistent behavior, but it is a behavior change, not a cleanup.

tests/e2e/autocomplete-arrow-button.spec.ts
```diff
@@ -0,0 +1,54 @@
+import { test, expect } from '../fixtures/test';
+import { AutocompleteAsyncSourcePage } from '../fixtures/pages/AutocompleteAsyncSourcePage';
+
+const EDITORS = ['autocomplete', 'dropdown'] as const;
```
budnix: The new AGENTS.md bullet says `handsontableRenderer` delegates to `autocompleteRenderer`, so `type: 'handsontable'` carries the same arrow — but the matrix here only covers `autocomplete` and `dropdown`. Was leaving it out deliberate? The fixture counts `source` queries and `HandsontableEditor` takes a nested grid config instead, so it may be more than an allowlist entry — curious what you think.

tests/fixtures/pages/AutocompleteAsyncSourcePage.ts
```diff
@@ -110,6 +120,19 @@ export class AutocompleteAsyncSourcePage {
     await expect.poll(() => this.pendingQueryCount(col)).toBeGreaterThan(0);
```
budnix, line 116: `openEditor()` on line 113 does the centered `cell.click()` that the new AGENTS.md bullet warns about. It works today only because the `SEEDS` widen the column — and the fixture comment right above says not to rely on that. Change a seed, a header, or a theme's cell padding and the click lands on the arrow, Enter commits and closes the editor, and every case in `autocomplete-async-source.spec.ts` fails at "the editor never opened" with nothing in the log. `hot.selectCell()` or an off-center click here would make it hold on its own.

handsontable/src/renderers/autocompleteRenderer/autocompleteRenderer.ts
demtario: Yes, the wider block was the intent — and you're right that the title was under-promising, so I widened it rather than narrowing the guard.

The reason for keeping `isLeftClick` is consistency with the path this change is modelled on. Walkontable's own double-click-to-open uses the same strict form: `((event as MouseEvent).button === 0 || this.touchApplied)` at `3rdparty/walkontable/src/event.ts:325`, and again on mouseup at `:472`. Switching this listener to `!isRightClick(event)` would leave the two paths disagreeing about middle-click and the back/forward buttons, which seems worse than either of them being strict. `isLeftClick` is also what `tableView.ts`, `nestedHeaders.ts` and `selection/mouseEventHandler.ts` already use.

On the plain-`Event` case: confirmed, `new Event('mousedown').button` is `undefined`, so that no longer opens the list. Nothing in-tree relies on it — `EventManager.fireEvent` (`eventManager.ts:204`) and `jquery.simulate` (`test/lib/jquery.simulate.js:112`) both default `button: 0`, and the frozen specs that drive this path pass unchanged. The exposure is third-party code hand-dispatching a bare `Event`, which was never a documented way in.

Changelog title is now "Fixed a non-left-button click on a cell's dropdown arrow opening the cell editor." (`3359a57`). Say the word if you'd still rather have the narrow `!isRightClick` guard — it's a one-line swap and I have no strong attachment beyond the consistency argument.

handsontable/src/renderers/autocompleteRenderer/autocompleteRenderer.ts
demtario: Good catch, and confirmed — the branch is at `editorManager.ts:347`:

```js
if (event instanceof MouseEvent && typeof allowOpening !== 'boolean') {
  allowOpening = this.hot.selection.getLayerLevel() === 0 && selection!.isSingle();
}
```

So forwarding the real event would gate the arrow behind a single-cell selection, and the arrow would stop opening the list after a shift-drag range or a Ctrl/Cmd non-contiguous selection.

I've added a comment saying exactly that, so the next person sees it is deliberate rather than an omission (`3359a57`):

```js
// The `null` event is load-bearing, not laziness: `EditorManager#openEditor` only applies
// its "no editor for a multi-cell selection" default when the event is a `MouseEvent`, so
// forwarding the real one here would stop the arrow from opening the list after a
// shift-drag range. Changing that is a behavior change, not a cleanup.
```

Agreed that the gated behavior might well be the more consistent one — but as you say, that is a separate change with its own tests, not something to slip into a button guard.

tests/e2e/autocomplete-arrow-button.spec.ts
demtario: It was deliberate, but you were right that it shouldn't have been — `handsontable` is now covered (`18bdda1`).

You had the reason it wasn't a one-line allowlist entry exactly right: this fixture's `source` is a query recorder, and a nested grid never calls it, so `?editor=handsontable` here would have left `openEditor()` waiting forever on `pendingQueryCount(col) > 0`. Rather than bend this fixture into two shapes, the arrow cases moved to their own:

- `tests/fixtures/demo/cell-dropdown-arrow.html` — `?cellType=autocomplete|dropdown|handsontable`. The first two take a flat `source`, `handsontable` takes a nested grid config.
- `tests/fixtures/pages/CellDropdownArrowPage.ts`
- spec renamed to `tests/e2e/cell-dropdown-arrow-button.spec.ts`, since it no longer covers only autocomplete.

That also drops the odd coupling where an arrow test depended on an async-source page object it had no use for. The helpers I'd added here for it (`arrow()`, `editorCoords()`) are reverted.

Checked that the new case actually tests something rather than riding along: with the guard disabled in the built bundle, all three right-button cases go red, `handsontable` included. Green run is 168 passed across this spec plus `autocomplete-async-source`, on all six theme × bundle legs.

tests/fixtures/pages/AutocompleteAsyncSourcePage.ts
demtario: This was the best catch of the four — the page object was doing precisely what the bullet I'd just written tells everyone else not to do. Fixed in `3359a57`:

```js
await this.cell(row, col).click({ position: { x: 5, y: 5 } });
```

with the reason inline, pointing at the fixture-contract bullet.

Your read on how thin the margin is matches what I measured while root-causing. On `main`, an empty cell sits at the 50px default width with the arrow spanning x=24–40 and the centre at x=25 — a hit. The `'Al'` seed widens the column to 56px, arrow at 30–46, centre 28 — a miss by about two pixels. And the arrow's own width is theme-dependent on top of that: `--ht-icon-size` is 16px on `main` and `horizon` but 12px on `classic` (`ht-theme-classic.css:70` maps it to `--ht-sizing-size-3`). So the failure you describe was reachable by a seed edit, a header change, or a theme's padding shifting — with the whole suite failing at "the editor never opened" and nothing in the log.

Re-ran the full suite after the change: 156 cases green on all six legs, so nothing here was depending on the centred click.

budnix (approved):

PR #12285, opened 2026-04-07
Fix autocomplete strict+allowInvalid click-away regression (#12284)

sl01k:
### Context

Fixes #12284 -- a regression introduced in v16.2.0 by PR #11873.

PR #11873 added a `finishEditing()` override to `AutocompleteEditor` that unconditionally discards the typed value when the user clicks away from the editor (to a different cell or outside the table). This was intended to make Autocomplete and Dropdown editors behave like a selection picker (discard on click-away), but it broke the established contract for Autocomplete cells configured with `strict: true` and `allowInvalid: true`.

**Root cause:** The `finishEditing()` override was placed in `AutocompleteEditor`, which is the parent class of `DropdownEditor`. The "discard on click-away" behavior is correct for Dropdown (a pure selection list), but wrong for Autocomplete, which is a hybrid text/suggestion editor where click-away should commit the typed value.

**Fix:** Move the `finishEditing()` override from `AutocompleteEditor` to `DropdownEditor`. This restores the pre-v16.2 behavior for Autocomplete (click-away saves the value) while keeping the intended Dropdown behavior (click-away discards).

### How has this been tested?

- Restored the original autocomplete E2E test that verifies click-away saves the value in non-strict mode
- Added two new E2E tests for strict mode + allowInvalid: click-away on table and click-away outside table
- Verified dropdown E2E tests still pass (31 specs, 0 failures)
- Verified autocomplete E2E tests pass (116 specs, 0 failures)
- ESLint passes with no errors

### Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature or improvement (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Additional language file or change to the existing one (translations)

### Related issue(s):
1. #12284

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular-wrapper`
- [ ] `@handsontable/react-wrapper`
- [ ] `@handsontable/vue3`

### Checklist:
- [x] I have reviewed the guidelines about [Contributing to Handsontable](https://github.com/handsontable/handsontable/blob/master/CONTRIBUTING.md) and I confirm that my code follows the code style of this project.
- [x] I have signed the [Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLScpMq4swMelvw3-onxC8Jl29m0fVp5hpf7d1yQVklqVjGjWGA/viewform?c=0&w=1)
[...]

PR #11873, opened 2025-10-03
Make the Autocomplete and Dropdown editors not save the temporary value when closed by clicking outside of the editor.

jansiegel:
### Context
This PR changes the way the Autocomplete and Dropdown editors work.

When clicked outside of the active editor:

<details>
<summary>Before</summary>

1. **Autocomplete editor**
    - **Strict mode**
        - **If the temporary value matches a source list entry**
            The temporary value is matched to a source list entry and saved.
            
        - **If the temporary value DOES NOT match a source list entry**
            The temporary value is saved with a validation error.
            

    - **Non-strict mode**
    The temporary value is saved after clicking outside of the editor.
        

2. **Dropdown editor**
    - **If the temporary value matches a source list entry**
        The temporary value is matched to a source list entry and saved.
        
    - **If the temporary value DOES NOT match a source list entry**
        The temporary value is saved with a validation error.
        

</details>

<details>
<summary>After</summary>

1. **Autocomplete editor**
    - **Strict mode**
        - **If the temporary value matches a source list entry**
            The temporary value is not saved.
            
        - **If the temporary value DOES NOT match a source list entry**
            The temporary value is not saved.
            

    - **Non-strict mode**
    The temporary value is not saved.
        

2. **Dropdown editor**
    - **If the temporary value matches a source list entry**
        The temporary value is not saved.
        
    - **If the temporary value DOES NOT match a source list entry**
        The temporary value is not saved.
        

</details>

**TL;DR;**
After this change, clicking outside of an open Dropdown/Autocomplete editor discards the typed in value and closes the editor.

Reference for the change: https://github.com/handsontable/dev-handsontable/issues/2757#issuecomment-3350730211

#### Important:
This is a modification of the current way the Autocomplete and Dropdown editors function.
@sl01k, @qunabu Should this be considered a breaking change or a fix for a wrong initial design?

### How has this been tested?
Modified an existing test case and added a new one.

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature or improvement (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Additional language file or change to the existing one (translations)

### Related issue(s):
1. handsontable/dev-handsontable#2757

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/angular-wrapper`
- [ ] `@handsontable/react`
- [ ] `@handsontable/react-wrapper`
[...]

Issue #7570, opened 2021-02-19
Autocomplete cell type does not update the list of choices correctly on Firefox


AMBudnik:
### Description

**Important**: this issue is not replicable on Chrome nor Edge.

### Steps to reproduce

1. Go to the first cell in the last column and open the editor
2. Type `bla`

![bla_](https://user-images.githubusercontent.com/10757813/108493271-6932b300-72a6-11eb-8d1b-0c9eef0eec30.gif)

**Result**
![image](https://user-images.githubusercontent.com/10757813/108491244-f45e7980-72a3-11eb-82d2-e1386de2415b.png)

### Demo

https://handsontable.com/docs/8.3.1/demo-autocomplete.html - current (broken)
https://handsontable.com/docs/7.4.0/demo-autocomplete.html - first broken
https://handsontable.com/docs/7.3.0/demo-autocomplete.html - last working

### Your environment
* Handsontable version: 7.4.0+
* Browser Name and version: Firefox 85.0.2
* Operating System: Windows 10

Update [forum](https://forum.handsontable.com/t/autocomplete-bug-on-firefox-all-version/5016)

aninde:
This issue cannot be reproduced on macOS, so I'm adding label Windows.
Tested with latests macOS Big Sur 11.2.1 and older Catalina 10.15.7.

On Safari this bug can not be reproduced due to dhttps://github.com/handsontable/handsontable/issues/7413.

AMBudnik:
We've just tested the scenario on multiple devices with @Osky772 and @wszymanski but apparently, this issue is connected to other more specific settings than a browser version of the system version. It makes it hard to fix and test.

More importantly, this issue has been reproduced also using Handsontable v2.0.0 so I'm renaming the table and removing the 'regression' label.

wszymanski:
I've tested it using Browserstack on `Windows 10` + `FF 84`, `FF 85` and virtual machine on `Windows 7` + `FF 86`.

Osky772:
I've tested it on `Windows 10` + `FF 86`

Maxoudela:
Reproduced on Handsontable 10.0.0 and Firefox 94.0.2 (64 bits) on Windows 10

Maxoudela:
The choices are filtered when the `onBeforeKeyDown` is triggered and `this.TEXTAREA.value` is used. It seems rather fair that Firefox is returning the value before the actual key has been pressed but other browsers are not working this way.

[...]

PR #13297, opened 2026-08-28
DEV-2676: Refuse a stale choice-list match when committing an editor

demtario:
### Context

The PR fixes the `autocomplete` and `dropdown` editors committing a previously suggested value instead of the one being typed.

`HandsontableEditor#finishEditing()` copies the inner choice list's selected value over the typed value whenever that list has a selection. In strict mode that is the point of the copy — it normalizes a typed `b` to the suggested `blue`. But `AutocompleteEditor` moves that highlight from a query that `#deferQuery()` schedules **10 ms behind each keystroke** (the workaround for #7570). A commit forced inside that window therefore reads the match for the *previous* keystroke and writes it into the cell.

`tests/e2e/editor-hidden-cell.spec.ts` is where this surfaced: the dropdown case could only assert that the editor closed, because pinning the committed value failed intermittently in CI. That test now asserts the value.

The cause was measured rather than argued. Instrumenting the commit path and running it 600 times across the `-min` legs reproduced 14 failures, and **every one** had a query timer still pending, with the choice list describing `A` while the textarea held `A4`:

```
dropdown.finishEditing:enter      value="A4" rangeIncludes=TRUE   <- the range veto does NOT fire
dropdown.finishEditing:outgoing   outgoing=false                  <- commit path taken
handsontableEditor.finishEditing  valueBefore="A4"
                                  innerSelected="[0,0,0,0]" innerValue="A1"
                                  valueAfter="A1"                 <- typed value lost
changes                           [[0, 0, "A1", "A1", "edit"]]
```

A passing run differed in one field only: `innerSelected="[3,0,3,0]" innerValue="A4"`.

The fix separates *why* the inner grid is selected from *whether* it may be committed:

- `HandsontableEditor` gains `innerSelectionOrigin` (`'user'` for an arrow-key or click pick) and `canCommitInnerSelection()`. It answers `true` there, so `type: 'handsontable'` behaves exactly as before.
- `AutocompleteEditor` commits the inner selection only when it is still the match the **current** value would produce. The matching rules moved out of `updateChoicesList()` into a shared `#deriveHighlight()` that both it and the guard call, so the two cannot answer differently.

Re-deriving asks "is the highlight **correct**?" rather than "is it **fresh**?". That distinction matters: `highlightBestMatchingChoice()` runs from the deferred query, so the highlight routinely describes older text than the value being committed — including for the whole time a function `source` has a response outstanding, which nothing inside a commit can wait out. A freshness test would refuse a suggestion that is merely old but still right, so a typed `blu` would commit instead of the highlighted `blue`, and under `allowInvalid: false` be reverted. Re-deriving needs no query to have run, so it commits `blue`.

Typing also clears a pick made with the arrow keys. Such a pick never writes to the textarea, so without that it kept winning over the text on screen — the same defect, reached through the branch that lets a pick win. That one is pre-existing rather than introduced here.

The `DropdownEditor` selection-range veto was the original suspect and is **not** implicated — it is the click-away policy from #11873 / #12285 and is untouched. Two comments in `editorManager.ts` that described it as the cause here are corrected.

There is no behavior trade to flag: a suggestion that is still the right match is committed exactly as before, so no guide or JSDoc change is required.

### How has this been tested?

Baseline before the fix: **14 failures in 600 runs**. After: **1200/1200**.

```bash
# the repro, inverted
cd tests && HOT_TEST_PORT=8124 npx playwright test \
  --project=e2e-main-min --project=e2e-classic-min --project=e2e-horizon-min \
  e2e/editor-hidden-cell.spec.ts -g "dropdown" --repeat-each=200

# all six theme x bundle legs, new and neighbouring specs
cd tests && HOT_TEST_PORT=8124 npx playwright test \
  e2e/editor-hidden-cell.spec.ts e2e/autocomplete-stale-highlight.spec.ts \
  e2e/editor-trimmed-row.spec.ts --repeat-each=5

npm run test:e2e --prefix handsontable -- --testPathPattern=editors
npm run test:e2e --prefix handsontable -- --testPathPattern=validators
npm run test:e2e --prefix handsontable -- --testPathPattern=cellTypes
```

Results:

| Check | Result |
|---|---|
| Repro, 1200 runs on the three `-min` legs | 1200 passed |
| Six legs, three specs, 5 repeats | 1350 passed |
| Jasmine `editors` tree | 747 specs, 0 failures |
| `validators` / `cellTypes` | 114 / 16 specs, 0 failures |
| Unit | 4043 passed |
| `test:types`, eslint, tests lint | clean |

New tests:

- `tests/e2e/autocomplete-stale-highlight.spec.ts` — six cases on the existing async-source fixture and the hidden-cell fixture: an arrow pick survives; typing supersedes such a pick; the typed value wins when the highlight is stale *and wrong*; a stale but still *correct* highlight is committed anyway; a current match normalizes a typed prefix; and the same holds on the array-source path when the query has not run at all.
- `tests/e2e/editor-hidden-cell.spec.ts` — the dropdown case now asserts the committed value, and a `hiddenRows` dropdown case was added so the behavior is not pinned to Pagination alone.

Each case was checked against a build without the fix. The stale-but-still-correct case fails against the first revision of this PR (which used a freshness test), and the wrong-highlight case fails against `develop`; the rest are regression guards that pass either way. The array-source case was also run five times over to confirm it is not timing-dependent — an earlier draft of it was, and now asserts its precondition explicitly.

The riskiest existing test was verified explicitly rather than reasoned about: the `commonCases.spec.js` case that prepares the next column's editor after a Tab commit sets the textarea directly with no key event, and passed before this change only because the open-time query had not run. It stays green.

### Types of changes

### Related issue(s):

1. DEV-2676

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2676

---

> [!NOTE]
> **Medium Risk**
> Changes the cell commit path for autocomplete, dropdown, and handsontable editors; behavior is well-tested but affects core editing and strict-mode normalization.
> 
> **Overview**
> Fixes **autocomplete** and **dropdown** editors saving an outdated suggestion from the inner choice list instead of what the user typed or selected.
> 
> **`HandsontableEditor`** now tracks **`innerSelectionOrigin`** (`user` vs `auto`) and commits via **`resolveInnerSelectionValue()`** instead of always copying the inner grid selection. Arrow-key and click picks still win; auto-highlighted rows no longer override the textarea blindly.
> 
> **`AutocompleteEditor`** re-derives the strict-mode match at commit time through shared **`#deriveHighlight()`** (same rules as **`updateChoicesList()`**), clears origin on **`input`** and stale choice lists on **`open()`**, and only applies list values when origin is `user` or the derived match is valid—so the 10 ms deferred query cannot commit the previous keystroke’s highlight.
> 
> **`editorManager`** comments are updated to attribute DEV-2676 to stale list matching, not **`DropdownEditor`**’s range veto. New **`autocomplete-stale-highlight`** e2e coverage and stricter assertions when a hidden cell forces commit.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e4820bf546fdc3d3f8d8ac071e4e9545b7ddbf5f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2676 Autocomplete/dropdown: a page change discards a dropdown edit that a select edit commits](https://app.clickup.com/t/9015210959/DEV-2676)

handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
```diff
@@ -323,6 +394,47 @@ export class AutocompleteEditor extends HandsontableEditor {
...
+   * @private
+   * @returns {boolean}
+   */
+  canCommitInnerSelection(): boolean {
+    if (this.innerSelectionOrigin === 'user') {
+      return true;
+    }
+
+    const { choices, highlightIndex } = this.#deriveHighlight(
+      this.stripValuesIfNeeded(this.rawChoices), this.#editorValue()
```
budnix: `updateChoicesList()` is public API and takes any array, but it never writes to `rawChoices` — so if someone calls `editor.updateChoicesList(['Custom A', 'Custom B'])` directly (it's in the API docs, and `autocompleteEditor.spec.js:1369` does it), the list on screen and the array this guard re-derives from are two different things. The derived match then never equals `htEditor.getValue()`, so the choice the user can see highlighted gets dropped for the raw typed text. The `else` branch of `queryChoices()` has the same gap — it calls `updateChoicesList([])` and leaves `rawChoices` holding the old source. Not sure what the clean fix is, since `rawChoices` holds raw entries and this method receives stripped ones, but worth deciding what a direct call should do.

handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
```diff
@@ -323,6 +394,47 @@ export class AutocompleteEditor extends HandsontableEditor {
...
+    }
+
+    const expected = choices[highlightIndex];
+
+    // Unwrapped the way the inner grid presents it - its `valueGetter` reduces a key/value entry to
+    // the `value` half, so comparing the raw entry would never match.
+    const expectedValue = this.#isKeyValueObject(expected) ?
+      (expected as Record<string, unknown>).value : expected;
+
+    return expectedValue === this.htEditor.getValue();
```
budnix: Curious whether the derived match could be committed here instead of only answering yes or no. Say the list still describes `'Al'` and the inner grid holds `'Alpha'`, then the user types `'f'` and hits Enter before a slow `source` answers: line 433 works out `'Alfa'`, but `htEditor.getValue()` is still `'Alpha'`, so this returns false and `'Alf'` lands in the cell. Under `allowInvalid: false` the strict validator then rejects `'Alf'` and the edit cannot be saved at all — develop at least committed `'Alpha'`. The right answer is already sitting in `expectedValue`; would handing it back (and letting `finishEditing()` use it) be worth the wider contract?

handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
```diff
@@ -511,7 +579,13 @@ export class AutocompleteEditor extends HandsontableEditor {
...
 
       if (this.cellProperties.strict === true) {
-        this.highlightBestMatchingChoice(highlightIndex ?? undefined);
+        const matchedIndex = highlightIndex ?? undefined;
+
+        this.highlightBestMatchingChoice(matchedIndex);
+
+        // Only on this branch: in non-strict mode `highlightBestMatchingChoice()` never runs, so a
+        // late query must not clear a `'user'` origin the arrow keys set.
+        this.innerSelectionOrigin = matchedIndex === undefined ? null : 'auto';
```
budnix, line 611: The comment explains why this sits on the strict branch, but it also sits inside `choices.length > 0`, and that part looks accidental. In strict mode a query that narrows to nothing skips the whole block: `strippedChoices` becomes `[]`, the root element is hidden, and the origin stays `'user'` from an earlier arrow pick. A commit after that still short-circuits to true and writes a value from a list that is no longer on screen. Setting it to `null` on the empty branch as well would cover that.

handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
```diff
@@ -780,6 +854,12 @@ export class AutocompleteEditor extends HandsontableEditor {
         timeOffset += 10;
       }
 
+      // Typing supersedes a pick made with the arrow keys or a click. That pick never wrote to the
+      // TEXTAREA, so without this the origin stays `'user'`, `canCommitInnerSelection()` keeps
+      // short-circuiting to true, and a commit forced before the query below runs writes the old
+      // pick over the text now on screen.
+      this.innerSelectionOrigin = null;
```
budnix: This clears the origin on keydown, which covers typing and Ctrl+V (keyCode 86 passes `isPrintableChar`), but text can reach the TEXTAREA with no keydown at all — a right-click → Paste, or a drag-and-drop into the box. After an arrow pick the origin is still `'user'`, so Enter short-circuits to true and the old pick is written over the pasted text, which is the same defect this PR fixes reached through another door. An `input` listener on the TEXTAREA would catch all of them; worth a quick check whether that's cheap here.

handsontable/src/editors/handsontableEditor/handsontableEditor.ts
```diff
@@ -117,6 +141,8 @@ export class HandsontableEditor extends TextEditor {
    * Closes the editor.
    */
   close(): void {
+    this.innerSelectionOrigin = null;
```
budnix: The `close()` comment over in `autocompleteEditor` already calls out that `refreshDimensions()` uses `close()` as a "hide for now" when the edited cell scrolls out of the rendered range (textEditor.ts:340). Nulling the origin here adds a second casualty on that path: the user picks `'Alto'` with the arrow keys, scrolls the cell out and back, and on Enter the origin is gone, so the guard re-derives from the typed `'Al'`, does not match `'Alto'`, and commits `'Al'`. The inner grid still has its row selected and `state` is still `EDITING`, so nothing else says the pick is stale. The `afterSetTheme` close at line 463 does the same — would be good to check whether this reset belongs in `finishEditing()` rather than `close()`.

tests/e2e/autocomplete-stale-highlight.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+async function commit(grid: AutocompleteAsyncSourcePage): Promise<void> {
+  await grid.page.keyboard.press('Enter');
+
+  await expect.poll(() => grid.pendingQueryCount(0)).toBeGreaterThan(0);
+  await grid.resolveQueries(0);
+}
+
+test.describe('committing an autocomplete whose choice list is out of date', () => {
+  test('keeps a choice the user picked with the arrow keys', async({ page, theme, bundle }) => {
+    const grid = new AutocompleteAsyncSourcePage(page, theme, bundle, { editor: 'dropdown' });
```
budnix, line 29: Every case in this file passes `{ editor: 'dropdown' }`, and `DropdownEditor.prepare()` hardcodes `cellProperties.filter = false`. So all six only exercise the `filteredChoiceIndexes[0]` branch of `#deriveHighlight`, and the `filter !== false` branch — the one plain `autocomplete` uses — has no coverage here. That branch falls back to `highlightIndex = 0` whenever `choices.indexOf(valueToMatch)` misses, which it does for key/value sources and for any case-insensitive match, since `valueToMatch` is lowercased and `choices` are not. The changelog says autocomplete is fixed too, so one case with `editor: 'autocomplete'` would be good to have.

handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
demtario: Good catch, and it went further than the direct call: `queryChoices()`'s empty-source branch has the same shape.

Fixed by deriving from `strippedChoices` instead of `rawChoices`. That is the array `updateChoicesList()` loads into the inner grid, so whatever put it there — `source`, a direct call, or the empty branch — the matched set and the visible list are the same object by construction. It also sidesteps the raw-vs-stripped mismatch you flagged, since `strippedChoices` is already stripped.

handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
demtario: Yes, and this was the most valuable comment of the set — it is not just a wider contract, the veto was actively worse than what it replaced. Under `allowInvalid: false` the typed `Alf` is rejected outright, so the edit cannot be saved at all, where develop at least committed a valid (if wrong) `Alpha`.

`canCommitInnerSelection()` is now `resolveInnerSelectionValue()` and returns the match. `Alf` against a list still showing `Alpha` commits `Alfa`. New case in `autocomplete-stale-highlight.spec.ts` pins it, and it fails against the previous commit.

handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
demtario: Agreed, that was accidental. Setting it to `null` on the empty branch now — once the list is hidden there is nothing on screen for a pick to refer to.

handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
demtario: Right, and `input` is cheap here — the listener goes on the TEXTAREA in `createElements()`, which already runs once per editor instance.

Worth recording why it is also safer than it looks: `input` does not fire for a programmatic `.value` write, so the commit path writing the resolved choice back cannot clear the origin it just acted on. The keydown clearing is gone.

handsontable/src/editors/handsontableEditor/handsontableEditor.ts
demtario: You are right, and the reset is simply unnecessary — `open()` sets the origin on both its branches, so every real re-open resets it anyway. Removed from `close()`, with a comment recording why it must not go back.

That also covers the `afterSetTheme` close you pointed at, since both reach the same `close()`.

tests/e2e/autocomplete-stale-highlight.spec.ts
demtario: Correct, and the fallback you describe is real: `valueToMatch` is lowercased while `choices` are not, so `indexOf` misses and index 0 wins. That is pre-existing behavior which the extraction preserves verbatim, but it was untested.

The fixture now takes a `strict` parameter, so an `autocomplete` case can reach that branch — `dropdown` could not, because `prepare()` hardcodes `filter: false`. The new case fails against the previous commit.

handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
```diff
@@ -323,6 +403,49 @@ export class AutocompleteEditor extends HandsontableEditor {
...
+
+    if (highlightIndex === null || highlightIndex >= choices.length) {
+      return undefined;
+    }
+
+    const matched = choices[highlightIndex];
+
+    // Unwrapped the way the inner grid presents it - its `valueGetter` reduces a key/value entry to
+    // the `value` half, so returning the raw entry would write an object into the cell.
+    return this.#isKeyValueObject(matched) ? (matched as Record<string, unknown>).value : matched;
```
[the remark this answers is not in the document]
demtario: Confirmed and fixed. Reproduced first: open the editor on column 0, resolve its query, Escape, then open column 1 and commit before its query lands — the cell took `Alpha`, a value column 1's `source` never offers.

`strippedChoices` and `rawChoices` are now cleared in `open()`. Pinned by a new case in `autocomplete-stale-highlight.spec.ts`, which fails against the previous commit with exactly that `Alpha`.

Worth noting this pre-dates the resolver only in reach, not in kind: `strippedChoices` was already instance state that outlived the cell, and `limitDropdownIfNeeded()` reads it too.

handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
```diff
@@ -323,6 +403,49 @@ export class AutocompleteEditor extends HandsontableEditor {
...
+   * against a set the user cannot see would drop the choice they can.
+   *
+   * @private
+   * @returns {*}
+   */
+  resolveInnerSelectionValue(): unknown {
+    // Non-strict never derives a highlight of its own, so there is nothing to re-derive: whatever
+    // is selected got there by the user's own arrow keys or click, or nothing is.
+    if (this.innerSelectionOrigin === 'user' || this.cellProperties.strict !== true) {
+      return super.resolveInnerSelectionValue();
```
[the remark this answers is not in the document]
demtario: Right, the two conditions were in the wrong order — `strict !== true` short-circuited before the origin was ever consulted, so clearing it on `input` was inert for default autocomplete.

Origin is now checked first, and non-strict returns `undefined` when no pick is outstanding rather than falling through to the inner selection. New case covers it: non-strict, arrow pick, then pasted text, commit — the typed text wins. It fails against the previous commit.

budnix (approved):

Issue #6380, opened 2019-10-23
When we paste merged cell null values are converted to empty strings


AMBudnik:
### Description

When we paste or autofill merged cell null values are converted to empty strings

### Steps to reproduce

1. Use autofill on `A1` to move cell to cover `D1:F3`
![image](https://user-images.githubusercontent.com/10757813/67391221-87977a00-f59e-11e9-92da-5da80c12fa4f.png)

### Demo

https://jsfiddle.net/u6pf21tn/4/

### Your environment
* Handsontable version: 7.2.1
* Browser Name and version: Chrome 77
* Operating System: Windows 10

AMBudnik:
After further investigation, I can add that

* undo runs but doesn't change a thing
Go to https://jsfiddle.net/r8egLb9n/ and use <kbd>CMD/CTRL</kbd> + <kbd>Z</kbd>. Hit the button to see that `afterChange` got triggered and changed data underneath the merged area.

* merging triggers `afterChange` as `populateFromArray` source. We could add this information under the method definition https://handsontable.com/docs/7.2.2/Core.html

* there is no way to block the operation by returning empty array ([as recommended in the docs](https://handsontable.com/docs/7.2.2/Hooks.html#event:beforeChange)) in `beforeChange` hook https://jsfiddle.net/r8egLb9n/1/  while it is possible by changing merging to `setDataAtCell` https://jsfiddle.net/r8egLb9n/2/ (but this may end up as a separate topic)

AMBudnik:
Copied and auto-filled values are **still** converted to empty strings in rev14 of 8.0.0-beta2 https://jsfiddle.net/AMBudnik/4cduhz18/

pnowak:
This issue is still replicable using v8.0.0. Here's an updated demo - https://jsfiddle.net/o3184ra2/

AMBudnik:
Works the same in v8.2.0 https://jsfiddle.net/handsoncode/mfcvn58d/

AMBudnik:
It works the same in Handsontable v12.1.3 https://jsfiddle.net/3547wthx/ But now we also see the missing border described at https://github.com/handsontable/handsontable/issues/8400

<img width="360" alt="Screenshot 2022-10-12 at 13 37 13" src="https://user-images.githubusercontent.com/10757813/195332925-e53f7322-c0b5-40e2-818f-583f428555bd.png">

AMBudnik:
[...]

PR #13213, opened 2026-08-21
issue-8794: Add the preserveTextValue cell meta option

sl01k:
### Context

Fixes #8794.

A cell explicitly configured as `type: 'text'` holding a number-like string (for example `0123456`) renders correctly, but the `Formulas` plugin passes the raw string to HyperFormula, which coerces it to the number `123456` (by-design engine behavior, see handsontable/hyperformula#1326). Formulas referencing the cell then lose the leading zero: `="ID"&"_"&B1` → `ID_123456`, `=LEN(B1)` → `6`.

The fix cannot be unconditional. `text` is the **default** cell type, and unconditionally passing text-cell values as strings would silently change results in existing apps (aggregates skip text, comparisons rank text above numbers).

### How it works

New cascading cell meta option `preserveTextValue` (default `false`, so nothing changes unless you enable it). When enabled on a `type: 'text'` cell, non-formula string values are escaped with `'` — HyperFormula's own string-escape mechanism — at every Handsontable→HyperFormula write boundary, mirroring the existing invalid-date escaping. The apostrophe is stripped on every read-back path. Empty strings are never escaped, so clearing a cell still produces an empty engine cell.

Set the option globally, per column, or per cell (through the `cell` option or the `cells()` function).

Escaped formula expressions are excluded. A `text` cell holding `'=SUM(A1)` — the documented way to store a formula as plain text — is left alone, so it does not reach the engine as `''=SUM(A1)`.

### Unified `cells()` handling on the load path

The bulk load scan and the edit paths now read cell meta the same way. Both go through the transient meta read, so a `type` or a `preserveTextValue` set only through the `cells()` function applies to the initial data, to `loadData()`, and to `updateData()`, not only to edited values.

Consequence, and the reason this is called out rather than buried: `cells()` and the `beforeGetCellMeta`/`afterGetCellMeta` hooks now run during the load scan, where they did not before. They run once per non-formula string cell in the dataset. Non-strings and formulas skip the meta read entirely.

Second consequence: the load scan has to decide the escaping before the engine sheet is populated, so a `cells()` implementation that reads other cells' *computed* values (through `getDataAtCell()`, `getData()`, and similar) sees raw source values on that pass. That is inherent — the escape decision cannot consult engine-computed values, because it produces them.

### Behavior change for grids that never enable the option

[...]

PR #13298, opened 2026-08-28
DEV-2689: Preserve null when a cell is edited, emptied, or pasted over

### LINKED ISSUE #3927: Opening cell editor should not overwrite 'null' values.
### Description

> For cell types Numeric, Date or Time, at least, my table correctly starts life with value null in the cell, which displays as "empty". If the user edits the cell but does not type anything, the cell's value is updated to the empty string.
While this seems OK initially, other things can later go wrong, e.g. formula calculation, server-side processing, etc., because the value is now of the wrong type. As your documentation correctly states, for example:
Make sure your cell values are numbers and not strings as Handsontable will not parse strings to numbers.
These cell types should recognise setting to empty string and convert the "" value to null. This is applicable not only after editing a cell but also, for example, after pasting a value imported from Excel.
For now I have written a beforeChange hook to walk the changes[] array and alter it in-place in this situation, but may I suggest that Hot should do this automatically somewhere in its own code, possibly even before validation.

added via [forum](https://forum.handsontable.com/t/empty-string-versus-null-in-non-string-type-cell/810)

### Steps to reproduce

1. Check console. You will see that for the 3rd row all data is null

2. Double click the last cell in `Acc` column to open editor and confirm with <kbd>ENTER</kbd> key

3. Check the console again and you will see that the same changed from `null` to an empty string.

### Demo

http://jsfiddle.net/r26e30ws/

### Your environment
* Version: 0.29.0
* Browser Name and version: Chrome 54

budnix:
### Context

The PR fixes two long-standing issues that turn out to be the same bug seen from two sides: Handsontable stored two different values for an empty cell depending on how the cell was emptied. Pressing <kbd>Delete</kbd> stored `null` (`core.ts:2808`), but the cell editor stored `''`. Nothing decided which was correct.

- Fixes #3927 — opening a cell editor on a `null` cell and confirming without typing overwrote the value with `''`.
- Relates to #6380 — a `null` carried through copy/paste came back as `''`. Deliberately *not* an auto-closing reference: the paste half is opt-in behind `emptyValue`, so a default-configured grid still behaves as it did and the issue should be closed by a human who agrees that is enough.

**What was measured on 18.0.0**, in Chrome against a fresh `develop` build, with real key presses:

| Action | Stored value | Verdict |
| --- | --- | --- |
| Delete / Backspace | `null` | correct |
| Open editor + Enter, typing nothing | `''` | bug (#3927) |
| Editor empties a filled cell | `''` | should be `null` |
| Autofill from a `null` cell | `null` | already fixed |
| Merge cells over data | `null` | already fixed |
| Copy a `null` cell, paste it back | `''` | bug (#6380) |

Only the editor path and the paste path were still wrong. The autofill and merge-cells halves of #6380 had already been fixed in the years since that issue was filed.

A no-op edit also damaged more than `null`: `undefined` became `''`, a boolean `true` became the string `'true'`, and a number in a text column became a string. Every one of those also fired `afterChange`, so simply opening a cell and pressing Enter marked the grid dirty.

#### 1. A confirm that changed nothing no longer writes

`finishEditing()` now compares the editor's content against what it was seeded with, and closes without writing when they match. The editor was seeded with `stringify(originalValue)`, and `stringify(null)` is `''` — which is exactly how a `null` became an empty string.

Two deliberate exceptions, both surfaced by the existing suite rather than guessed at:

- **Ctrl/Cmd + Enter is excluded, but only when it fills other cells.** It copies the edited cell's value across the selection, so an unchanged editor still has work to do — over a *single-cell* selection it fills nothing, so the guard stays armed there.
- **A validated cell writes its own value back** instead of skipping, because `allowInvalid: false` must keep the editor open on an invalid value however the user got there. The stored value is untouched, but one no-op `afterChange` fires. Only an unvalidated cell skips the write entirely.

Validating in place with `hot.validateCell()` was tried instead of that write-back and reverted: its callback runs before `postAfterValidate`, and `allowInvalid: false` then fails to hold the editor open at all.

The guard's baseline is released when the editor **closes**, not when `finishEditing()` starts. Releasing it on entry made the guard survive only one confirm, so a cell that `allowInvalid: false` reopened fell back to the old path on the second Enter.

#### 2. New `emptyValue` option

Applied in `getValueSetterValue`, which sits in the **shared** write path, so one setting covers the editor, paste, autofill, `setDataAtCell` and `setSourceDataAtCell` alike. Loading data is not a write, so `loadData`, `updateData` and the initial `data` bypass it — documented on the option. That is also why no clipboard change was needed: typed columns already parse what they receive through their own `valueSetter`, so pasting `7` into a numeric column already stored a number.

The default stays `''`, so **no existing grid changes behavior**. `emptyValue: null` stores `null` whenever a cell ends up empty. It works at grid, column and cell level, runs after any custom `valueSetter`, and leaves `0`, `false` and `' '` untouched.

### Behavior change to be aware of

This is not a breaking change by the project definition — no default value moved — but a user-visible behavior does change, so it should be a conscious call rather than a surprise:

**A cell editor opened and confirmed without any change no longer writes to the cell, and on an unvalidated cell fires no `afterChange`.** Code that relied on that write — for example on a `null` becoming `''` just by visiting the cell, or on a change event firing whenever an editor closes — will see the original value preserved and no event instead.

This is unconditional; it is not gated behind `emptyValue`. It is filed as a bug fix rather than a breaking change because no default setting value moved, but it is a default *behavior* change and wants a reviewer's sign-off on that framing.

Two frozen-suite specs **asserted the old behavior** and were updated to the new intent, each with a comment explaining why:

- `autocompleteEditor.spec.js` expected `''` after hovering an item and clicking away; now expects the cell's `null` to survive.
- `textEditor.spec.js` expected a missing nested property to become `''`; now expects it to stay missing.

### How has this been tested?

Every failure was baselined against unmodified source before being trusted — base was **0 failures**, so each failure seen during development was genuinely introduced and was fixed rather than explained away. Each new test was also proven to fail with the guard disabled.

```bash
# Unit — 4065/4065 (330 suites), including 13 new for getValueSetterValue
npm run test:unit --prefix handsontable -- --testPathPattern=valueAccessors

# Types — emptyValue pinned at both grid and per-column level

# Lint — 0 errors

# Legacy E2E — editors 850/850, formulas 420/420,
# validator + copyPaste + autofill + mergeCells + undoRedo 1275/1275
npm run test:e2e --prefix handsontable -- "--testPathPattern=editor"
npm run test:e2e --prefix handsontable -- "--testPathPattern=formulas"
npm run test:e2e --prefix handsontable -- "--testPathPattern=(validator|copyPaste|autofill|mergeCells|undoRedo)"

# New Playwright spec — 19/19
cd tests && npx playwright test e2e/empty-value.spec.ts --project=e2e-main
```

New tests added:

- `handsontable/src/utils/__tests__/valueAccessors.unit.js` — 20 unit tests covering `emptyValue`, including that `0`, `false`, `NaN` and `' '` are never treated as empty, that a checkbox template or a choice list's `''` is left alone, and that undo restores verbatim.
- `tests/e2e/empty-value.spec.ts` + fixture and page object — 21 Playwright tests covering the no-op guard, the guard *not* swallowing real edits or fast edits, the unchanged default, the second confirm of an editor that `allowInvalid: false` reopened, single-cell Ctrl+Enter, and `emptyValue: null` across the editor, paste and typed columns.

### Fixed after code review

A review pass found four correctness bugs in the first version. Each is fixed, and each fix has a test that was verified to fail without it:

1. **The guard survived only one confirm.** Its baseline was released at the top of `finishEditing()`, so a cell that `allowInvalid: false` reopened fell back to the old path on the second Enter and wrote `''` — reintroducing #3927 for exactly the users who validate most strictly.
2. **Ctrl/Cmd + Enter over a single cell still destroyed a `null`**, because the guard was disarmed for that chord unconditionally.
3. **`emptyValue: null` broke a checkbox with `uncheckedTemplate: ''`**, leaving the cell matching neither template and unable to toggle.
4. **`refreshValue()` left a stale baseline** after reloading the editor from the data source.

Two documentation errors were corrected too: a range fill was described as storing `''` (it has always stored `null`), and a validated cell's confirm was described as writing nothing.

Writing the second-confirm test was itself worth the trouble — the first version of it passed against the bug, because the strict configuration rejected the bad `''` write anyway. It only became a real test once the validator was changed to reject `null` but accept `''`.

### Types of changes

### Related issue(s):

1. DEV-2689
2. #3927
3. #6380

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

Documentation included in this PR: full JSDoc for `emptyValue` in `metaSchema.ts` (the API reference), and a new "Empty cell values" section in the binding-to-data guide covering what each write path stores and the limit that a text or HTML clipboard cannot carry the `null` / `''` distinction.

### Still open from @Haxikowy's review

Three points from that review are deliberately not in this push:

- **No migration-guide entry yet.** `migrating-from-18.1-to-18.2` landed on develop in #13213 after this branch was cut, so adding the section needs develop merged in first.
- **`AutocompleteEditor.getValue()` baseline timing** (`rawChoices` filled in a `setTimeout` after the baseline is captured) — not verified yet, so not claimed either way.
- **Remaining test gaps** — per-cell `emptyValue` via `cells()`, a `trimWhitespace` E2E case, and `resetValueBeforeEdit`.

### Note for the reviewer

While writing the Playwright spec I found a **pre-existing** issue unrelated to this change: Chrome's native `<input type="date">` consumes <kbd>Enter</kbd>, so under Playwright no keyboard-driven date edit ever commits — a genuine edit fails to save too, verified on unmodified source. The new spec therefore covers date columns through the paste path, with a comment saying why. It may deserve its own ticket.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2689

---

> [!NOTE]
> **Medium Risk**
> All grids get a default behavior change on unchanged editor confirm (hooks and stored values), while `emptyValue` touches the shared cell write path in `core.ts`—high integration surface but well-tested and opt-in for normalization.
> 
> **Overview**
> Fixes inconsistent empty-cell storage and accidental edits when confirming a cell editor without changes.
> 
> **Editor confirm (#3927):** `BaseEditor.finishEditing()` now skips the save path when the parsed editor value still matches what the cell had when editing started (with exceptions for Ctrl/Cmd+Enter over a multi-cell range and fast-edit seeding). Unchanged confirms no longer coerce types (e.g. `null` → `''`, numbers to strings), do not fire `beforeChange`/`afterChange`, and still run validation via `#finishUnchangedEdit()`. Text/select editors call `resetValueBeforeEdit()` after `refreshValue()` so the baseline stays correct.
> 
> **`emptyValue` (18.3):** New cascading option (`''` default) applied in `getValueSetterValue()` after `valueSetter`, remapping written `''` to e.g. `null` for editor, paste, fill, and `setDataAtCell`/`setSourceDataAtCell`. Skips undo/redo sources and columns where `''` is meaningful (checkbox templates, array `source` with blank option).
> 
> Docs add binding-to-data guidance, API/meta schema entries, and a **18.2 → 18.3** migration guide for the editor behavior change. Coverage includes unit tests for `valueAccessors`, updated editor/`beforeChange` specs, and Playwright `empty-value` e2e.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 753ffa40196760e3a0ef85f8a7547a20d3aaa45c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2689 Preserve null values when a cell is edited, emptied, or pasted over](https://app.clickup.com/t/9015210959/DEV-2689)

handsontable/src/editors/baseEditor/baseEditor.ts
```diff
@@ -392,6 +445,44 @@ export class BaseEditor {
...
+    // A validated cell is written its OWN value back, rather than skipped. Validating in place with
+    // `hot.validateCell()` was tried and rejected: its callback runs before `postAfterValidate`, and
+    // the `selectCell()` inside `discardEditor()`'s reopen branch then leaves `allowInvalid: false`
+    // unable to hold the editor open - the behavior this branch exists to preserve.
+    //
+    // The value written is the cell's stored value, so nothing is lost. The costs are that one
+    // `afterChange` fires with the value unchanged, and that a custom `valueSetter` runs again over
+    // an already-set value - harmless for an idempotent setter, which every built-in one is.
+    this.state = EDITOR_STATE.WAITING;
+    this.saveValue([[this.originalValue]], false);
```
Haxikowy: This write-back re-enters `processChanges`, so `emptyValue` and `valueSetter` re-apply to an already-stored value: with `emptyValue: null` a validated cell holding `''` is silently rewritten to `null` on a confirm that changed nothing. Every built-in typed column ships a validator (numeric, date, time, autocomplete, dropdown), so this is the common branch, and it contradicts the guide's "the stored value stays exactly as it was".

handsontable/src/editors/baseEditor/baseEditor.ts
```diff
@@ -392,6 +445,44 @@ export class BaseEditor {
     }
   }
 
+  /**
+   * Closes an editor whose content the user never changed, without writing anything.
+   *
+   * A validated cell is still validated, because `allowInvalid: false` has to keep the editor open on
+   * an invalid value however the user got there. Validation runs **in place** rather than by writing
```
Haxikowy: This block says validation runs in place rather than by writing the value back, and names re-running `emptyValue` as the hazard being avoided. Line 478 does that write, so the comment is left over from the reverted `validateCell()` attempt and hides the bug from anyone reading the method.

handsontable/src/editors/baseEditor/baseEditor.ts
```diff
@@ -305,18 +326,31 @@ export class BaseEditor {
...
     this.open(event);
     this._opened = true;
     this.focus();
 
+    // Read back through `getValue()` rather than reusing the string passed to `setValue()`: an editor
+    // may normalize what it was given (DateEditor rewrites an unparsable value to an empty string), and
+    // the guard has to compare against what the user actually sees. Captured after `open()` so an editor
+    // that fills its own control while opening is accounted for, but before `afterBeginEditing` runs, so
+    // a value a listener sets there still counts as a change and is saved.
+    this.#valueBeforeEdit = seededFromCell ? this.getValue() : null;
```
Haxikowy: `DateEditor.setValue()` substitutes `cellProperties.defaultDate` when the incoming value is empty, so on an empty cell this baseline is already the default date and every confirm compares equal. A `date` column with `defaultDate` can no longer store that date from the editor: Enter, Enter on an empty cell leaves `null` where 18.0.0 stored the default (Bugbot flagged this on 5016618b, and 8eabc84f did not touch `dateEditor.ts`).

handsontable/src/editors/baseEditor/baseEditor.ts
```diff
@@ -367,6 +399,27 @@ export class BaseEditor {
...
+      // selection, so an unchanged editor still has work to do there. Over a single cell it writes
+      // only the edited cell, which makes it the same gesture as a plain Enter.
+      const fillsOtherCells = ctrlDown === true &&
+        this.hot.getSelectedRangeActive()?.isSingle() === false;
+
+      // The editor still holds exactly what it was opened with, so the user confirmed without changing
+      // anything. Writing the editor's stringified value back over the cell is what turned a `null`
+      // into `''` (#3927), so an unchanged confirm must not go through the normal save path.
+      const isUnchanged = !fillsOtherCells &&
+        this.#valueBeforeEdit !== null &&
```
Haxikowy: `null` means both "disarmed" and "captured value" here, while `getValue()` is typed `unknown`. A custom editor returning `undefined` arms the guard with `undefined`, so `undefined === undefined` holds on every confirm and the cell can never save; the repo's own custom-editor type test is literally `getValue() {}`.

handsontable/src/editors/baseEditor/baseEditor.ts
```diff
@@ -367,6 +399,27 @@ export class BaseEditor {
...
+        this.getValue() === this.#valueBeforeEdit;
+
+      if (isUnchanged) {
+        this.#finishUnchangedEdit();
+
+        return;
+      }
+
+      this.#valueBeforeEdit = null;
```
Haxikowy: Clearing the baseline here loses it for the rest of the editing session, because `discardEditor()`'s `allowInvalid: false` branch returns to `EDITING` without calling `close()` or `prepare()`. With a validator that rejects `x` but accepts `''`, on a `null` cell: type `x`, Enter, delete back to empty, Enter writes `''` over the `null`, which is #3927 again (the second-confirm test cannot catch it, since that path returns at line 415 and never reaches this line).

handsontable/src/editors/baseEditor/baseEditor.ts
```diff
@@ -367,6 +399,27 @@ export class BaseEditor {
...
+      // Ctrl/Meta + Enter over a real range copies the edited value into every other cell of the
+      // selection, so an unchanged editor still has work to do there. Over a single cell it writes
+      // only the edited cell, which makes it the same gesture as a plain Enter.
+      const fillsOtherCells = ctrlDown === true &&
+        this.hot.getSelectedRangeActive()?.isSingle() === false;
+
+      // The editor still holds exactly what it was opened with, so the user confirmed without changing
+      // anything. Writing the editor's stringified value back over the cell is what turned a `null`
+      // into `''` (#3927), so an unchanged confirm must not go through the normal save path.
+      const isUnchanged = !fillsOtherCells &&
```
Haxikowy: The guard returns before `trimWhitespace` (line 425) and `valueParser` (line 429), so a cell holding `' abc '` in a `trimWhitespace: true` grid is no longer normalized by open then Enter. Either run both before the comparison, or record the drop in the changelog.

handsontable/src/editors/baseEditor/baseEditor.ts
```diff
@@ -367,6 +399,27 @@ export class BaseEditor {
         return;
       }
 
+      // Ctrl/Meta + Enter over a real range copies the edited value into every other cell of the
+      // selection, so an unchanged editor still has work to do there. Over a single cell it writes
+      // only the edited cell, which makes it the same gesture as a plain Enter.
+      const fillsOtherCells = ctrlDown === true &&
```
Haxikowy, line 422: `getSelectedRangeActive()?.isSingle() === false` evaluates to `false` when there is no active range, so a Ctrl+Enter with no selection takes the guard path rather than the fill path. `saveValue()` at line 271 defends a null range the other way; `!== true` matches the stated intent.

handsontable/src/editors/textEditor/textEditor.ts
```diff
@@ -319,6 +319,10 @@ export class TextEditor extends BaseEditor {
     this.originalValue = sourceData;
 
     this.setValue(sourceData);
+    // The editor now shows the cell's own value again, so the unchanged-edit baseline has to follow
+    // it. Leaving the opening value in place would compare the user's next confirm against content
+    // the editor no longer holds.
+    this.resetValueBeforeEdit();
```
Haxikowy, line 325: `SelectEditor.refreshValue()` (selectEditor.ts:211) reloads the control and reassigns `originalValue` but never calls this, so its baseline describes content the editor no longer shows. Open a select editor, `setSourceDataAtCell` under it, pick the option the stale baseline holds, then confirm: the visible choice is discarded.

handsontable/src/utils/valueAccessors.ts
```diff
@@ -8,13 +8,27 @@ import { isFunction } from '../helpers/function';
...
 
-  return value;
+  // `emptyValue` spells out what an emptied cell stores. It runs after `valueSetter` so a cell that
+  // ends up empty lands on the same value whichever path emptied it - the editor, a paste, a fill or
+  // `setDataAtCell()` - and so a custom `valueSetter` returning `''` still means "empty" here.
+  //
+  // Unless `''` is a value this cell assigns meaning to. A checkbox configured with
+  // `uncheckedTemplate: ''` stores `''` to mean "unchecked", and mapping that to `null` would leave
+  // the cell matching neither template, rendering `#bad-value#` and refusing to toggle.
+  const isEmptyStringMeaningful = cellMeta.uncheckedTemplate === '' || cellMeta.checkedTemplate === '';
```
Haxikowy: This reads the templates off cascading cell meta rather than off the cell's type, so `uncheckedTemplate: ''` set once at grid level disables `emptyValue: null` for every column in the grid. Gate on the cell actually being a checkbox instead.

handsontable/src/utils/valueAccessors.ts
```diff
@@ -8,13 +8,27 @@ import { isFunction } from '../helpers/function';
...
+
+  if (newValue === '' && emptyValue === null && !isEmptyStringMeaningful) {
```
Haxikowy: Only the literal `null` does anything: `emptyValue: 0`, `'N/A'` and `undefined` are accepted and behave as the default, with no warning. The `'' | null` type stops TS users only, and the name reads as "the value to store", so either widen this to return `emptyValue` or warn on an unsupported value.

handsontable/src/utils/valueAccessors.ts
```diff
@@ -8,13 +8,27 @@ import { isFunction } from '../helpers/function';
...
+    return null;
```
Haxikowy: UndoRedo restores through `setDataAtCell` (`plugins/undoRedo/actions/dataChange.ts:136`), so this mapping runs on undo and a cell that legitimately held `''` can never be restored under `emptyValue: null`. Such cells do exist, because `loadData` never passes through this function; exempt the `UndoRedo.*` sources the way `readOnly` already is.

handsontable/src/dataMap/metaManager/metaSchema.ts
```diff
@@ -7273,6 +7273,63 @@ export default (): Record<string, unknown> => {
...
+     * | `''` (default) | Store an empty string when a cell is emptied                             |
+     * | `null`         | Store `null` when a cell is emptied                                      |
+     *
+     * Set `emptyValue: null` when the cell's value leaves the grid — saved to a server, written to a
+     * database, or read by a formula. An empty string in a `numeric`, `date` or `time` column is a
+     * string where a number, a date or nothing at all is expected, and `''` is not the same value as
+     * `NULL` to a database. It also matches how spreadsheets tell a blank cell from an empty string:
+     * `ISBLANK()` is `true` for `null` and `false` for `''`.
+     *
+     * The option applies to every write path, including [`setDataAtCell()`](@/api/core.md#setdataatcell).
```
Haxikowy: `getValueSetterValue` has three call sites and all of them are writes, so `loadData`, `updateData`, the initial `data` and `updateSettings({ data })` bypass it: a grid configured this way holds both `''` and `null` at once. The same sentence also silently covers `setSourceDataAtCell` (core.ts:4351), which integrations use to write server payloads verbatim; name both.

handsontable/src/dataMap/metaManager/metaSchema.ts
```diff
@@ -7273,6 +7273,63 @@ export default (): Record<string, unknown> => {
...
+     * a blank cell across a range, and merging cells over data all store `null`. Clearing the
+     * [cell editor](@/guides/cell-functions/cell-editor/cell-editor.md) and confirming, or pasting a
+     * blank cell, store an empty string (`''`). Set `emptyValue` to `null` to make every one of those
+     * paths agree on `null`.
+     *
+     * You can set the `emptyValue` option to one of the following:
+     *
+     * | Setting        | Description                                                              |
+     * | -------------- | ------------------------------------------------------------------------ |
+     * | `''` (default) | Store an empty string when a cell is emptied                             |
```
Haxikowy: This row says the default stores `''` when a cell is emptied, which contradicts the paragraph four lines above and the guide's table: under the default, Delete, `setDataAtCell(null)`, fill and merge all store `null`. The option is one-way, mapping `''` to `null` and never `null` to `''`.

handsontable/src/__tests__/core/settings.types.ts
```diff
@@ -69,7 +69,11 @@ const allSettings: Required<Handsontable.GridSettings> = {
   columns: [
     {
       type: 'numeric',
-      numericFormat: { style: 'currency', currency: 'USD', minimumFractionDigits: 2, maximumFractionDigits: 2 }
+      numericFormat: { style: 'currency', currency: 'USD', minimumFractionDigits: 2, maximumFractionDigits: 2 },
+      // Pins the per-column form of `emptyValue` that the option's docs and the binding-to-data
+      // guide both show. `ColumnSettings` derives from `GridSettings`, so this compiles only while
+      // that stays true.
+      emptyValue: null,
```
Haxikowy, line 73: `ColumnSettings` carries `[key: string]: any` (settings.ts:86), so this literal keeps compiling through the index signature even if `emptyValue` is dropped from `GridSettings`, and the test stays green. Pin it directly with `ColumnSettings['emptyValue']` plus a `@ts-expect-error` on an invalid value.

Haxikowy (commented):
Six findings below have no line in the diff to attach to. The rest are inline.

**No migration-guide entry for the behavior change.** A default-configured grid stops firing `afterChange` when an editor closes unchanged, and code relying on that (dirty tracking, autosave, a `null` becoming `''` just by visiting a cell) breaks with nothing in the upgrade path. #13213 shipped one day ago with the same shape, kept `breaking: false`, and added an 86-line numbered section to `migrating-from-18.1-to-18.2`; this needs the same.

**`afterChange` now fires or not depending on whether the column has a validator.** `#finishUnchangedEdit()` skips the write on an unvalidated cell and echoes through `saveValue()` on a validated one, so one grid gives silence in `text` columns and a no-op event in `numeric`/`date`/`dropdown` ones, with nothing in the user-facing config to predict which. The split is harder to defend than the change itself; making both paths silent would also remove the `emptyValue` re-application flagged at `baseEditor.ts:478`.

**`AutocompleteEditor.getValue()` (autocompleteEditor.ts:111) returns a different type depending on edit order.** `open()` fills `rawChoices` in a `setTimeout(0)`, after the baseline is captured, so with a key-value `source` the first cell edited has a string baseline and writes the source object, while the second has an object baseline, compares equal, and keeps the plain string. Same gesture, two results.

**`emptyValue: null` breaks a `dropdown`/`autocomplete` whose `source` contains `''` as a blank option.** The user's explicit choice of the blank entry maps to `null`, which then fails `autocompleteValidator` under `strict: true` because `null` is not in `source`. The checkbox carve-out at `valueAccessors.ts:25` covers the analogous checkbox case but not this one.

**Test gaps.** No undo/redo case, no `trimWhitespace` case, no checkbox or autocomplete column, no per-cell `emptyValue` via `cells()` although the JSDoc claims cell level works, and no `emptyValue: null` on a validated cell holding `''` (which is what would have caught `baseEditor.ts:478`). `fillsOtherCells` at `baseEditor.ts:405` exists solely for the multi-cell Ctrl+Enter branch and only the single-cell case is tested. `resetValueBeforeEdit()` has no test at all, though the description says every review fix got one.

**Smaller things.** `Fixes #6380` auto-closes a public issue whose default behavior this PR does not change, since the paste fix is opt-in behind `emptyValue`. The description's test counts are stale against the branch: Playwright is 21, not 19; the unit file has 15 tests of which 12 cover `getValueSetterValue`, not 13; and `getValueSetterValue` is at `core.ts:2364`, not 2397.

handsontable/src/editors/baseEditor/baseEditor.ts
budnix: You're right — no write on either branch now, validation runs from `postAfterValidate` so `allowInvalid` still holds the editor open. Fixed in ea3bb6f79f.

handsontable/src/editors/baseEditor/baseEditor.ts
budnix: Good spot, that comment was stale — rewritten in ea3bb6f79f to match what the code does.

handsontable/src/editors/baseEditor/baseEditor.ts
budnix: Nice catch — the baseline is the seeded string now, not `getValue()` read back after opening, so `defaultDate` saves again. Fixed in ea3bb6f79f.

handsontable/src/editors/baseEditor/baseEditor.ts
budnix: Agreed — split the armed flag out into `#hasValueBeforeEdit` in ea3bb6f79f, so a `getValue()` returning `undefined` no longer jams the guard.

handsontable/src/editors/baseEditor/baseEditor.ts
budnix: You're right, the rejected-edit path lost it — released only in `discardEditor()`'s closing branch now (ea3bb6f79f).

handsontable/src/editors/baseEditor/baseEditor.ts
budnix: Good point — both run before the comparison now (ea3bb6f79f), so an unchanged confirm still trims exactly as it did before.

handsontable/src/editors/baseEditor/baseEditor.ts
budnix: Fair point — it's `!== true` in ea3bb6f79f, so a missing range no longer reads as the fill branch.

handsontable/src/editors/textEditor/textEditor.ts
budnix: Nice catch — `SelectEditor.refreshValue()` re-arms the baseline too now (ea3bb6f79f).

handsontable/src/utils/valueAccessors.ts
budnix: You're right — gated on the cell's own `type` in ea3bb6f79f, and the same carve-out now covers a dropdown or autocomplete offering `''` in `source`.

handsontable/src/utils/valueAccessors.ts
budnix: Good point — it stores whatever `emptyValue` is set to now, not just `null` (ea3bb6f79f).

handsontable/src/utils/valueAccessors.ts
budnix: Agreed, undo has to restore verbatim — `UndoRedo.*` sources skip the mapping in ea3bb6f79f.

handsontable/src/dataMap/metaManager/metaSchema.ts
budnix: Good spot — the docs now say loading bypasses the option and name `setSourceDataAtCell` alongside `setDataAtCell` (ea3bb6f79f).

handsontable/src/dataMap/metaManager/metaSchema.ts
budnix: You're right, that row contradicted the text above it — reworded in ea3bb6f79f, and it now says the mapping is one-way.

handsontable/src/__tests__/core/settings.types.ts
budnix: Nice catch, that pin proved nothing — it reads `ColumnSettings['emptyValue']` back out now (ea3bb6f79f).

docs/content/guides/getting-started/binding-to-data/binding-to-data.md
```diff
@@ -849,6 +849,72 @@ hot.populateFromArray(1, 1, newValues, 2, 2);
...
+
+::: tip
+
+Opening a cell editor and confirming it without typing anything never changes the cell, whatever
+`emptyValue` is set to.
+
+On a cell without a [`validator`](@/api/options.md#validator), nothing is written at all and no
+[`afterChange`](@/api/hooks.md#afterchange) hook fires. On a validated cell the value is written back
+unchanged, so the validator still runs and [`allowInvalid`](@/api/options.md#allowinvalid) keeps
+behaving as it always has. Either way the stored value stays exactly as it was.
```
Haxikowy: `ea3bb6f79` removed the write-back, so this paragraph now describes the opposite of what the code does: nothing is written on either branch, the validator runs through `validateCell()` rather than through a write, and no `afterChange` fires for a validated cell either. It was written in `8eabc84f` to document the write-back and needs the same edit that `#finishUnchangedEdit()` just got.

Two smaller leftovers from the same commit: `metaSchema.ts:7328` still declares `@type {''|null}` though `core/settings.ts` widened the option to `CellValue` and the table above it now says "Any other value", and that table row does not cover `emptyValue: undefined`, which the implementation treats as unset rather than storing `undefined`.

budnix: You're right on all three — the tip, the `@type` tag and the missing `undefined` row are fixed in 646c1f80b2, with a unit test pinning the `undefined` case. Also bumped the option to `@since 18.2.0`.

handsontable/src/editors/baseEditor/baseEditor.ts
```diff
@@ -367,8 +413,19 @@ export class BaseEditor {
         return;
       }
 
+      // Ctrl/Meta + Enter over a real range copies the edited value into every other cell of the
+      // selection, so an unchanged editor still has work to do there. Over a single cell it writes
+      // only the edited cell, which makes it the same gesture as a plain Enter. `!== true` rather
+      // than `=== false`, so a missing range reads as "does not fill other cells" instead of falling
+      // into the fill branch.
+      const fillsOtherCells = ctrlDown === true &&
+        this.hot.getSelectedRangeActive()?.isSingle() !== true;
```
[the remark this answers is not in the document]
Haxikowy: Confirmed, and it's my fault: my earlier comment asked for `!== true` and had the logic backwards. A missing range needs `fillsOtherCells` to be `false` so the guard stays armed, which is what the original `=== false` did; `(getSelectedRangeActive()?.isSingle() ?? true) === false` states it explicitly, and the comment at lines 418-420 currently describes the opposite of what the line does.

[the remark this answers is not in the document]
budnix: Good catch, and thanks for tracing it back to your own note — fixed in b9f6274, with an E2E test that takes the range away for one Ctrl+Enter confirm.

handsontable/src/utils/valueAccessors.ts
```diff
@@ -1,20 +1,68 @@
...
+function isEmptyStringConfigured(cellMeta: Record<string, unknown>): boolean {
+  const { type } = cellMeta;
+
+  if (type === 'checkbox') {
+    return cellMeta.checkedTemplate === '' || cellMeta.uncheckedTemplate === '';
+  }
+
+  if (type === 'autocomplete' || type === 'dropdown') {
+    return Array.isArray(cellMeta.source) && cellMeta.source.includes('');
+  }
```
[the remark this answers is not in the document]
budnix: This assumes `null` reaches the `source` lookup, and it never does. `autocompleteValidator` turns `null` into `''` before the `strict` branch, then answers with `allowEmpty`, so `null` and `''` validate identically. Checked against the built validator: with `strict: true` and `source: ['', 'a', 'b']` both are valid; with `allowEmpty: false` both are invalid; only a real non-empty value missing from `source` fails. `dropdownValidator` calls the same function.

The array-only limit is real though, and was undocumented. It is now on the option and in the guide, with a unit test pinning the function-source case, in 84b9143.

Haxikowy (dismissed):

demtario (dismissed):

handsontable/src/utils/valueAccessors.ts
```diff
@@ -1,20 +1,83 @@
...
+  // ends up empty lands on the same value whichever path emptied it - the editor, a paste, a fill or
+  // `setDataAtCell()` - and so a custom `valueSetter` returning `''` still means "empty" here.
+  if (newValue !== '' || emptyValue === '' || emptyValue === undefined) {
+    return newValue;
+  }
+
+  // Undo and redo restore what the cell held before, verbatim. Remapping here would make an `''` that
+  // legitimately predates the setting - `loadData` never passes through this function - impossible to
+  // restore, so the user would watch undo produce a different value than the one they undid to.
+  if (typeof source === 'string' && source.startsWith('UndoRedo.')) {
```
demtario, line 68: **High: the UndoRedo `source` carve-out never reaches `setSourceDataAtCell`, so undoing a row/column removal can silently overwrite legitimate `''` values.**

This exemption (`source.startsWith('UndoRedo.')`) is fed by the one call site in `processChanges()` (`core.ts:2397`), which now passes `source`. But `setSourceDataAtCell()` has two more calls to `getValueSetterValue()` — `core.ts:4375` and `core.ts:4391` — and both still call it with only `(changeValue, cellMeta)`, even though `source` is a function parameter already in scope there (`core.ts:4348`).

`undoRedo/actions/removeRow.ts:152` and `removeColumn.ts:214` restore data through exactly that path: `hot.setSourceDataAtCell(changes, undefined, undefined, 'UndoRedo.undo')`. So: configure a column with `emptyValue: null`, put a real `''` in a cell, delete the row, undo the deletion — the restored value becomes `null`, not the `''` it actually held. This is row/column-wide, not single-cell, and directly contradicts the comment right above this line ("Undo and redo restore what the cell held before, verbatim").

Not caught because the only unit coverage of this carve-out calls `getValueSetterValue()` directly with `'UndoRedo.undo'` supplied by the test (`valueAccessors.unit.js:136-138`) — it can't catch a caller that never passes `source` through. No test drives a real `hot.undo()` after a row/column removal with `emptyValue` set.

Fix: pass `source` through at both `core.ts:4375` and `core.ts:4391`, same as the fix already applied in `processChanges()`.

handsontable/src/editors/baseEditor/baseEditor.ts
```diff
@@ -377,6 +435,17 @@ export class BaseEditor {
         value = this.cellProperties.valueParser(value, this.cellProperties);
       }
 
+      // The editor still holds exactly what it was opened with, so the user confirmed without changing
+      // anything. Writing the editor's stringified value back over the cell is what turned a `null`
+      // into `''` (#3927), so an unchanged confirm must not go through the normal save path.
+      const isUnchanged = !fillsOtherCells && this.#hasValueBeforeEdit && value === this.#valueBeforeEdit;
```
demtario, line 441: **High (flag for explicit sign-off, not necessarily a bug): this "unchanged confirm writes nothing" guard changes default behavior for every existing grid, unconditionally.**

`isUnchanged` applies regardless of whether `emptyValue` is configured — it's not gated behind the new opt-in setting. Per `.changelogs/3927.json`, this is classified `"type": "fixed"`, `"breaking": false`: after this change, confirming an editor without altering its value no longer fires `afterChange`/`beforeChange`, for every grid on upgrade, with no way to opt back into the old behavior.

That's a defensible classification — the old behavior (rewriting the cell and firing a change event on a no-op confirm) was arguably the actual bug. But it's bundled into a PR whose headline feature is an opt-in `emptyValue` setting, and this repo's breaking-changes policy asks that a behavior change like this be a deliberate, visible call rather than a side effect. Any consumer using `afterChange` as a per-confirm counter (dirty-flag tracking, autosave triggers) will observe fewer events post-upgrade.

Not asking for a code change here — asking that this be confirmed as an intentional, signed-off classification.

demtario (commented):
2 inline notes below.

handsontable/src/utils/valueAccessors.ts
budnix: Good spot, and thanks for tracing it to the exact call sites — both now pass `source` through in 9cac15d, with an E2E test that drives a real undo after a row removal.

handsontable/src/editors/baseEditor/baseEditor.ts
budnix: Confirmed and signed off — the old no-op write was the actual bug, so it stays `fixed` and non-breaking. One thing worth knowing: it is slightly broader than the diff you read, since bdec8a6 also stopped `beforeChange` firing on such a confirm, and both will be called out in the 18.1 to 18.2 migration guide.

demtario (dismissed):

Issue #4646, opened 2017-12-01
`average` type od SummaryCalculations shows `NaN` when there is no data


AMBudnik:
### Description

`average` type od SummaryCalculations shows `NaN` when there is no data

as mentioned at https://forum.handsontable.com/t/average-summary-column-gives-incorrect-results/1611

> Average summary column NAN when all values set to 0 or blank cell. Instead of NAN shouldn't this be 0 or blank and not ERROR. Also if you enter 1 in one average cell and set the rest to zero. the average comes out at 1 and not 0.2 for example.
> 
> in the example below modify the E column to all zeros to get NAN. Then change one cell in E to 1 or 23 etc to get the wrong average result. In fact any blank or zero column is ignored in the average calculation
> 
> ie
> 
> 10 10 10 10 0
> 
> should be 8 as 5 cells but results at 10

### Steps to reproduce

1. Erase values from `E1` to `E5`

### Demo

http://jsfiddle.net/handsoncode/Lmow8a37/

### Your environment
* Handsontable version: 1.14.3
* Browser Name and version: Chrome 62
* Operating System: Windows 10

AMBudnik:
Renaming the topic to `change` and `to be discussed`. I would recommend changing `NaN` to `Not enough data` like other calculations are doing.

chris601:
Defiantly a bug.  Don't think you've looked at this properly;  Also title not relaying the issue that I raised.  The average does not work correctly !!  I had to code my own solution.

AMBudnik:
Thank you for your feedback @chris601 

Do you think that we should add something or create additional issue report?

AMBudnik:
Just to update the topic - it works the same in `7.1.1` but the previous demo got broken from missing resources. Here's a new one http://jsfiddle.net/handsoncode/g9ohb7zj/

AMBudnik:
It works the same for v8.1.0 http://jsfiddle.net/AMBudnik/qpnh3aou/ 
And I still request on changing the `NaN` to `Not enough data`

![Zrzut ekranu 2020-10-8 o 12 03 54](https://user-images.githubusercontent.com/10757813/95444673-5d30ff00-095e-11eb-981c-6a61e10b7cc2.png)

budnix:
Hi,

I tested this on the current development version and I can confirm the report.

**The `NaN` result**

[...]

PR #13299, opened 2026-08-28
DEV-2691: Show "Not enough data" for an empty average column summary

budnix:
### Context

The PR fixes the `average` column summary showing `NaN` when the summarized range holds no values, reported in #4646 (open since 2017).

I first checked whether the issue still reproduces on `develop` (18.0.0, Chrome 148). It does, using the issue's own steps: select the source cells and press Delete. Delete writes `null` into the cells (`emptySelectedCells`, `handsontable/src/core.ts:2810`), `calculateAverage` was `calculateSum / countEntries`, and `countEntries` subtracts the blank cells — so an all-blank range divided `0 / 0`.

`average` was the only summary type without a guard for this. Measured on the same empty column:

| type | before | after |
|---|---|---|
| `min` | `Not enough data` | unchanged |
| `max` | `Not enough data` | unchanged |
| `count` | `0` | unchanged |
| `sum` | `0` | unchanged |
| `average` | **`NaN`** | **`Not enough data`** |

So this aligns `average` with what `min` and `max` have always done, which is also what was requested on the issue in 2018 and again in 2020.

**This is a user-visible behavior change.** An empty range now yields the string `Not enough data` instead of the number `NaN`, so code reading the destination cell programmatically gets a string. That is the same contract `min`/`max` already had. The `calculateAverage` return type widened from `number` to `number | string`, matching `calculateMinMax`. It is **not breaking** under `.ai/BREAKING-CHANGES.md`: the method is `@private` and absent from the public API reference.

The shared `'Not enough data'` literal moved into a `NOT_ENOUGH_DATA` constant, since three summary types use it now.

#### Not reproducible, so not changed

The issue also quotes a forum post claiming `10 10 10 10 0` averages to `10` instead of `8`. That does **not** reproduce — it returns `8`. Zeros are counted correctly. Whatever caused it in 1.14.3 is long gone.

#### Deliberately left out of scope

Two pre-existing quirks found while investigating. Both would drag `min`/`max` along, so they belong in their own tickets:

1. A cell holding `''` counts as the number `0`, while `null` counts as blank (`isNaN('')` is `false`). So an all-`''` column averages to `0` while an all-`null` column reports `Not enough data`.
2. `Not enough data` is a hardcoded English string that does not go through i18n. That has been true for `min`/`max` since they were written.

### How has this been tested?

Three cases added to the existing `calculateAverage` block in `columnSummary.spec.js`: an all-blank range, the issue's actual gesture (clearing populated cells with Delete), and a guard that the normal path did not move — a range with *some* blanks still averages over the non-blank cells only.

The two new `Not enough data` cases were run with the fix reverted first, and both failed with `Expected NaN to be 'Not enough data'` — so they genuinely catch the bug rather than merely passing.

```bash
# E2E — 61 specs
npm --prefix handsontable run test:e2e -- --testPathPattern=columnSummary

# Unit — 4067 tests

# Types — caught the widened return type; the type test was updated to match calculateMinMax

# Lint and build
```

All pass. The full legacy E2E suite was not run locally — this machine has ~51 viewport specs that fail from display state regardless of the change; CI covers those.

### Types of changes

### Related issue(s):

1. #4646
2. DEV-2691

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

- [x] My change requires a change to the documentation. The summary-functions table in the column summary guide said nothing about empty columns, for any function — this PR adds a line covering all five.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2691

---

> [!NOTE]
> **Low Risk**
> Localized column-summary display fix with no auth or data-pipeline impact; only behavioral change is destination cells showing a string instead of NaN when the range has no values.
> 
> **Overview**
> Fixes the **`average`** column summary showing **`NaN`** when every cell in the range is empty (e.g. after clearing with Delete). **`calculateAverage`** now checks that **`countEntries`** is finite and positive before dividing; otherwise it returns **`Not enough data`**, matching **`min`** / **`max`**. The shared message is centralized in **`NOT_ENOUGH_DATA`**, and the private return type is **`number | string`**.
> 
> Docs now state that **`min`**, **`max`**, and **`average`** show **`Not enough data`** on empty columns while **`sum`** and **`count`** stay **`0`**. New specs cover all-empty ranges, Delete-to-clear, partial blanks, default ranges, and invalid range bounds.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9c1f398ae747beaab4353b890b957d4f45300470. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changelogs/13299.json
+++ b/.changelogs/13299.json
@@ -1,5 +1,5 @@
 {
-  "issuesOrigin": "private",
+  "issuesOrigin": "public",
   "title": "Fixed a bug where the `average` column summary showed `NaN` once the summarized range held no values, for example after you cleared the source cells with the Delete key. It now shows `Not enough data`, the same result the `min` and `max` summaries already returned for an empty range. Ranges that still hold at least one value are unaffected, and empty cells stay excluded from both the sum and the divisor.",
   "type": "fixed",
   "issueOrPR": 13299,
--- a/.changelogs/13299.json
+++ b/.changelogs/13299.json
@@ -1,8 +1,8 @@
 {
-  "issuesOrigin": "public",
+  "issuesOrigin": "private",
   "title": "Fixed a bug where the `average` column summary showed `NaN` once the summarized range held no values, for example after you cleared the source cells with the Delete key. It now shows `Not enough data`, the same result the `min` and `max` summaries already returned for an empty range. Ranges that still hold at least one value are unaffected, and empty cells stay excluded from both the sum and the divisor.",
   "type": "fixed",
-  "issueOrPR": 4646,
+  "issueOrPR": 13299,
   "breaking": false,
   "framework": "none"
 }
--- a/.changelogs/4646.json
+++ b/.changelogs/4646.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed a bug where the `average` column summary showed `NaN` once the summarized range held no values, for example after you cleared the source cells with the Delete key. It now shows `Not enough data`, the same result the `min` and `max` summaries already returned for an empty range. Ranges that still hold at least one value are unaffected, and empty cells stay excluded from both the sum and the divisor.",
+  "type": "fixed",
+  "issueOrPR": 4646,
+  "breaking": false,
+  "framework": "none"
+}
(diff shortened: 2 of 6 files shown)


qunabu:
Task linked: [DEV-2691 Fix `average` column summary showing NaN for an empty range (#4646)](https://app.clickup.com/t/9015210959/DEV-2691)

adrianspdev (dismissed):

adrianspdev (approved):

Issue #5833, opened 2019-03-04
Undo of remove_row doesn't restore data if dataSchema is a function


gentoo90:
### Description
If `dataSchema` is a function, UndoRedo plugin doesn't save the data from a removed row and when undo action is run, this row is restored empty.

### Steps to reproduce
1. Remove a row
2. Undo the last action

### Demo
http://jsfiddle.net/nf4ewdt6/
![remove_row_undo](https://user-images.githubusercontent.com/1297574/53758170-f025cf80-3ec5-11e9-9ab3-36de6585db37.gif)

### Your environment
* Handsontable version: 6.2.2
* Browser Name and version: Chrome 72, Firefox 65
* Operating System: Windows 8.1

AMBudnik:
Thank you for sharing the issue report @gentoo90 

I have replicated the issue and you're right when we use `dataSchema` the returned row is blank.

AMBudnik:
Data remains blank after updating the example to `8.0.0-beta1` http://jsfiddle.net/ok4ptb1a/

AMBudnik:
Filling up the subject with some investigation.

When we use 'undo' we get a console error
![Zrzut ekranu 2020-04-17 o 10 09 18](https://user-images.githubusercontent.com/10757813/79546928-8483cf00-8093-11ea-9552-31377ce1c068.png)
![Zrzut ekranu 2020-04-17 o 10 12 07](https://user-images.githubusercontent.com/10757813/79547245-ee03dd80-8093-11ea-8b43-835e62fc1651.png)

pnowak:
This issue is still replicable using v8.0.0. Here's an updated demo - https://jsfiddle.net/Lrevxqkw/

AMBudnik:
This issue works the same in v8.2.0 - https://jsfiddle.net/cd1vmpou/1/
We can also replicate the same behavior using https://handsontable.com/docs/8.2.0/tutorial-data-sources.html#page-data-schema

budnix:
Hi @gentoo90

I tested your demo and the bug still reproduced. It is now fixed in #13300.

There were two separate faults behind it.

The row came back empty because the snapshot held nothing restorable when the columns use function `data` accessors, which is the pattern the "Function data source and schema" guide documents. The removed row's accessor values are now captured and written back through the accessor after the row is re-inserted.

[...]

PR #13300, opened 2026-08-28
DEV-447: Restore row data on undo when dataSchema is a function (#5833)

adrianspdev:
### Context

The PR fixes undoing a row removal when `dataSchema` is a function that returns non-plain objects (GitHub #5833, ClickUp DEV-447).

Before this change, `RemoveRowAction.undo()` wrote every captured key of the removed row back through `Object.defineProperty`. Two documented setups broke:

- **Constructor rows with a read-only or non-configurable member** (the reporter's fiddle: `Object.defineProperty(this, 'sum', { get })`). The undo threw `TypeError: Cannot redefine property: sum`. The throw escaped before the plugin reset `ignoreNewActions`, so UndoRedo silently stopped recording for the rest of the session.
- **The "Function data source and schema" guide pattern** (constructor `dataSchema` + function `columns[].data` accessors). The snapshot held nothing restorable, so the row came back empty (the original 2019 report). Worse, a render between the index-mapper update and the data splice called the user accessor with `undefined`, aborting `alter()` half-way (`countRows()` 4 vs source 3) and leaving `getData()` throwing.

What changed:

- `setProperty` writes through a setter or plain assignment when the own property is non-configurable, and leaves read-only members untouched instead of throwing.
- `UndoRedo.undo()` / `redo()` reset `ignoreNewActions` when an action throws, then rethrow.
- `DataMap.get()` no longer calls a function accessor with a missing source row.
- `DataSource.setAtCell()` / `getAtCell()` and `Core#setSourceDataAtCell()` / `getSourceDataAtCell()` accept a `ColumnDataGetterSetterFunction` column (parameter widening only; the type is now re-exported from `handsontable` and `handsontable/base`).
- `RemoveRowAction` captures accessor-column values in a new `accessorValues` field and restores them through the accessor after the row is re-inserted. Function-valued own keys are dropped from the row snapshot: a function is never a cell value, and restoring one aliased the removed row's closure onto the row the schema creates on undo.
- Public read types (`HotInstance.colToProp`, `BaseEditor.prop`) are unchanged; `src/__tests__/core/undoRemoveRowAccessor.types.ts` locks the consumer patterns that would otherwise break.

Behavior change to be aware of: a row object carrying a per-row function as an own property (at any depth) that `dataSchema` does not recreate no longer survives undo (previously it was written back by reference).

Review round (commit 3270713) — all findings from the first review addressed:

- `setProperty` writes own accessors through their setter and never redefines them into data properties, so a configurable object-literal getter keeps deriving; `writeOwnProperty` also guards `__proto__`/`constructor`/`prototype` itself (CodeQL).
- `DataMap.get()`/`set()` route function-accessor reads and writes through the `modifyRowData`-aware row and skip the write when the source row is missing.
- `RemoveRowAction.undo()` disarms the `afterViewRender` settle callback when the restoring write throws, so a half-undone action can no longer land on the redo stack; the function-key sweep in the snapshot is recursive; the accessor-column scan runs once per removal.
- `setSourceDataAtCell` resolves a function prop through the accessor-aware `propToCol`, so accessor columns read their own column meta (`valueSetter`, `sourceDataValidator`) instead of the table-meta fallback.
- `GridSettings['modifySourceData']` and the hook JSDoc now type `column` as `number | string | ColumnDataGetterSetterFunction` (method syntax keeps existing narrow handlers assignable).
- `ColumnDataGetterSetterFunction` documented in the TypeScript-types guide table and referenced from binding-to-data.

Out of scope, pre-existing: ES-class prototype getters render `null` regardless of undo (`DataMap.get()` reads own properties only). Remaining follow-ups noted during review: `tableView.ts` casts `colToProp()` to `string`; `HotInstance.propToCol` does not accept a function prop.

### How has this been tested?

- Jest: `npm --prefix handsontable run test:unit` (4070 tests) — new units in `helpers/__tests__/object.unit.ts`, `plugins/undoRedo/__tests__/undoRedo.unit.js`, `plugins/undoRedo/__tests__/actions/removeRow.unit.js`, `dataMap/__tests__/dataMap.unit.js`, `dataMap/__tests__/dataSource.unit.js`.
- Types: `npm --prefix handsontable run test:types` — new `src/__tests__/core/undoRemoveRowAccessor.types.ts`.
- Playwright: `cd tests && npx playwright test e2e/undo-remove-row-schema.spec.ts` — 2 tests × 6 projects (theme × bundle), both scenarios through the context menu and Ctrl/Cmd+Z, asserting restored values, no page errors, `countRows() === getSourceData().length`, and a second remove→undo cycle.
- Legacy Jasmine/Puppeteer: `npm --prefix handsontable run test:e2e` — 9807 specs, 0 failures (the `removeRow.spec.js` exact-shape expectation gains `accessorValues`).
- Lint/build: `npm --prefix handsontable run eslint`, `run stylelint`, `run build`.
- Manual: reporter's fiddle config and the docs `example7` config, remove row → undo → second cycle, on the built `dist/`.

### Types of changes

### Related issue(s):

1. DEV-447
2. #5833

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-447

---

> [!NOTE]
> **Medium Risk**
> Changes core undo/redo and source-data write paths with subtle index-mapper timing; behavior shifts for function-valued own properties on restored rows, but the change is well covered by new unit, type, and e2e tests.
> 
> **Overview**
> Fixes **undo of row removal** when rows come from a **function `dataSchema`** or columns use **function `data` accessors** — cases that previously threw, restored empty rows, or left UndoRedo stuck with `ignoreNewActions` true.
> 
> **`RemoveRowAction`** now snapshots **`accessorValues`** per removed row and reapplies them via `setSourceDataAtCell` after re-inserting the row; row clones **strip nested function properties** so undo does not alias old row closures. **`setProperty`** writes through accessor setters and **skips read-only / non-configurable** targets instead of redefining them.
> 
> The **data layer** routes function accessors through **`modifyRowData`-aware** `dataRow` reads/writes and **no-ops** when a mapped row has no backing source yet (mid-undo). **`setSourceDataAtCell` / `getSourceDataAtCell`** accept accessor functions; **`ColumnDataGetterSetterFunction`** is exported and documented.
> 
> **UndoRedo** resets **`ignoreNewActions`** and **discards partial actions** when undo/redo throws; remove-row undo **disarms** the `afterViewRender` settle hook if the restoring write fails.
> 
> Docs note that undo restores accessor columns automatically; Playwright and extensive Jest coverage were added.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 76b9c14a532696dac1f5d4f1c5669752c83af603. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-447 Undo of remove_row doesn't restore data if dataSchema is a function](https://app.clickup.com/t/9015210959/DEV-447)

handsontable/src/helpers/object.ts
```diff
@@ -366,6 +366,38 @@ export function getProperty<T = unknown>(object: Record<string | symbol, unknown
...
+ * a derived getter on a constructor-created row – is left untouched instead of throwing.
+ *
+ * @param {object} target The object that owns the property.
+ * @param {string} propName The own property name to write.
+ * @param {*} value The value to write.
+ */
+function writeOwnProperty(target: Record<string, unknown>, propName: string, value: unknown): void {
+  const descriptor = Object.getOwnPropertyDescriptor(target, propName);
+
+  if (descriptor && descriptor.configurable === false) {
```
demtario: **medium:** `writeOwnProperty` guards only `configurable === false`, so a *configurable* own accessor is still replaced by a data property and stops deriving.

An object-literal getter has the descriptor `{ get, enumerable: true, configurable: true }`, so it skips the new branch and reaches the unconditional `Object.defineProperty` below. Verified:

```js
const o = { a: 1, b: 2, get sum() { return this.a + this.b; } };
// after the defineProperty path: sum === 99, getter gone, o.a = 100 leaves sum at 99
```

That is a row shape this PR is fixing: `captureRowData`'s `deepClone` uses `Object.entries`, which *does* capture an own enumerable getter's value, and `undo()` writes it back through `setProperty`. Reachable with `dataSchema: () => ({ a: null, b: null, get sum() { ... } })`. A class-prototype getter is non-enumerable and so is genuinely safe, but an own literal getter is not.

Not a regression, the pre-PR code clobbered it identically. Raising it because this helper exists to protect accessors on the restore path and covers only the non-configurable half, and all four new `object.unit.ts` cases use non-configurable descriptors.

Suggestion: branch on descriptor *kind* before configurability. If `descriptor.get || descriptor.set`, call the setter and skip when getter-only; then handle the non-configurable data-property case; then fall through to `defineProperty`.

Second, smaller point on the same helper (line 441): the non-configurable read-only fall-through turns a former `TypeError` into a silent no-op for *every* `setProperty` caller. The only non-test write-path caller is `dataSource.setAtCell` (:299, :304), which `core.setSourceDataAtCell` reaches at `core.ts:4384`, so `hot.setSourceDataAtCell(0, 'readOnlyProp', 'x')` previously threw and now returns having written nothing. The changelog scopes the read-only handling to "undoing a row removal", so that half of the change is undocumented. Either warn once from the discard branch, or widen the changelog sentence.

handsontable/src/base.ts
```diff
@@ -459,8 +459,8 @@ export {
 // Note: CellCoords and CellRange are already exported as runtime values above.
 export type { GridSettings, Events, SanitizerContext } from './core/settings';
 export type {
-  CellValue, CellChange, RowObject, SourceRowData, ChangeSource, CellMeta, CellProperties,
-  ColumnSettings, RemoveIndexSignature
+  CellValue, CellChange, ColumnDataGetterSetterFunction, RowObject, SourceRowData, ChangeSource,
```
demtario, line 463: **medium:** `ColumnDataGetterSetterFunction` is the only symbol this PR adds to the public type surface (here and at `index.ts:217`), and it is not documented.

`docs/content/guides/tools-and-building/typescript-types/typescript-types.md` has a "Data types" table that enumerates every sibling type exported from these two entries -- `CellValue`, `CellChange`, `RowObject`, `ChangeSource`, `SourceRowData`, `SelectOptionsObject`. The new one is absent, and `grep -rn "ColumnDataGetterSetterFunction" docs/` returns nothing. The one-line `binding-to-data.md` change documents the undo *behavior*, not the type.

The Typedoc half of the docs requirement is covered -- JSDoc was updated on `colToProp`, `getSourceDataAtCell` and `setSourceDataAtCell`, and the interface itself gained a block. The gap is the hand-maintained guide table.

Add a row next to `SourceRowData`. Worth referencing from that guide's "Identify changed columns in hooks" section too, which already teaches that `prop` is the accessor function.

handsontable/src/dataMap/dataSource.ts
```diff
@@ -255,6 +275,15 @@ class DataSource {
 
     const dataRow = this.modifyRowData(normalizedRow);
 
+    if (typeof column === 'function') {
```
demtario, line 278: **medium:** This branch hands an accessor function to `modifySourceData`, whose contract still says the argument is a number.

`core/hooks/constants.ts:2081` documents `@param {number} column Physical column index or property name.` and `core/settings.ts:659` types it `column: number`. Both legs are now affected: the undo write here (the hook fires at :269, just above) and `captureAccessorValues`' read via `getSourceDataAtCell` -> `getAtPhysicalCell:339`.

I checked the blast radius, and it is bounded -- no crash. In-repo `formulas` survives: its handler bails on `ioMode !== 'get'` for the write, and on the read leg `propToCol`'s cache hit returns a real visual column. Even on a cache *miss* (`dataMap.ts:301-303` returns the accessor unchanged) `toPhysicalColumn(fn)` is `null`, so `getCellType` returns `'EMPTY'` and the handler returns early. Userland handlers doing string or index arithmetic on `column` are the ones that would notice.

The PR widened `setSourceDataAtCell`, `setAtCell`, `getAtCell`, `colToProp` and `propToCol` but stopped at the hook. Widen the `@param` and the type to `number | string | ColumnDataGetterSetterFunction` so the documented contract matches the rest of the chain.

handsontable/src/plugins/undoRedo/actions/removeRow.ts
```diff
@@ -130,6 +184,23 @@ export class RemoveRowAction extends BaseAction {
       });
     });
 
+    // Accessor-column values live behind a function and are invisible to `data`; restore them
+    // through the accessor itself, the same way `dataSource.setAtCell` writes through it. The guard
+    // also bails out when the column has no visual index left – a column trimmed since the removal
+    // makes `toVisualColumn` return `null`, and `colToProp(null)` echoes `null` back, which is not
```
demtario, line 234: **medium:** This comment makes four behavioral claims about trimmed and hidden columns during undo, and no test pins any of them.

The `colToProp(null)` echo checks out by reading `dataMap.ts:262-270`, so this is a coverage gap rather than a defect. Worth splitting the claims by how reachable they are:

- **Hidden columns** -- `hiddenColumns` is a shipped plugin, so "a hidden column keeps its visual index, so it restores like any other" is reachable by any user and is the claim worth a test.
- **Trimmed columns** -- no in-repo plugin trims columns, so that half is only reachable through a custom `TrimmingMap` on `columnIndexMapper`. Note that it also encodes a real asymmetry: an accessor column trimmed at removal time is never captured and its value is unrecoverable, while a prop-based column survives because `getSourceDataAtRow` walks every own key. The comment states that limitation, so it reads as an accepted tradeoff, but it is worth being sure that is deliberate.

Add a unit test that removes and undoes a row with a column hidden, so the assertion this comment makes cannot rot silently.

demtario (commented):
8 inline notes below.

handsontable/src/plugins/undoRedo/undoRedo.ts
```diff
@@ -264,10 +264,18 @@ export class UndoRedo extends BasePlugin {
...
+    try {
+      (action as { undo: (hot: HotInstance, callback: () => void) => void }).undo(this.hot, () => {
+        this.ignoreNewActions = false;
+        this.undoneActions.push(action);
+      });
+
+    } catch (error) {
+      // An action that throws never reaches its settle callback. Without this reset every later
+      // user action would be silently dropped from the stack for the rest of the session.
       this.ignoreNewActions = false;
```
demtario, line 278: **low:** The reset fixes the wedged plugin, but the action itself is still lost. It was popped at `:249`, and this `catch` neither pushes it back onto `doneActions` nor reaches the `undoneActions.push` in the settle callback, so a throwing undo drops one history entry and also skips `afterRedoStackChange` and `afterUndo`. Same shape at `:340` for redo.

Not a regression -- the pre-PR code lost the action too. Raising it because this is new hardening code and the two new `undoRedo.unit.js` tests assert only `ignoreNewActions`, so the stack behavior is left unspecified either way.

Either push `action` back onto the stack it came from before rethrowing, or state in the comment that the entry is deliberately discarded because a partially applied action cannot be safely replayed.

handsontable/src/core.ts
```diff
@@ -4315,20 +4318,24 @@ export default function Core(
...
     const input = setDataInputToArray(row, column, value);
     const isThereAnySetSourceListener = instance.hasHook('afterSetSourceDataAtCell');
     const changesForHook: Array<Array<unknown>> = [];
-    const getCellProperties = (changeRow: number, changeProp: string | number): CellProperties => {
+    const getCellProperties = (
+      changeRow: number, changeProp: string | number | ColumnDataGetterSetterFunction
+    ): CellProperties => {
       const visualRow = instance.toVisualRow(changeRow);
-      const visualColumn = instance.toVisualColumn(changeProp as number);
+      const visualColumn = typeof changeProp === 'function' ? null : instance.toVisualColumn(changeProp as number);
```
demtario: **low:** Forcing `null` for a function prop sends every accessor-column restore down the `tableMeta` fallback, so it skips the column's own cell meta.

That meta feeds both `getValueSetterValue(changeValue, cellMeta)` and `runSourceDataValidator(newValue, cellMeta, ...)` at :4372-4382. So a column-level `valueSetter` or `sourceDataValidator` is honored on a normal edit to that column but bypassed when undo restores it -- an asymmetry introduced by this fix, since accessor props never reached this path before.

The information needed is already captured: `removeRow.ts:195` destructures `physicalColumn` from the `accessorValues` tuple. Passing a resolvable column alongside the accessor would let `getCellProperties` call `getCellMetaTransient` for accessor columns instead of falling back.

.changelogs/13300.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "private",
```
demtario, line 2: **low, question rather than a finding:** which issue origin does the team want here?

The two most recent entries of this exact shape -- a `DEV-xxx` ID *plus* a public issue in the PR title -- both used the public form: `072bfbe3d` added `.changelogs/6154.json` with `"issuesOrigin": "public"`, `"issueOrPR": 6154`, and `884cd6502` added `.changelogs/5933.json` with `public`/`5933`. This PR's subject has the same shape, `... (#5833)`.

But `.changelogs/README.md:76` reads the other way: "Work tracked in a private ClickUp task is `private` -- that covers every change with a `DEV-xxx` ID", and DEV-447 is exactly that. So the repo's own docs are self-inconsistent and this is one of two defensible readings, not a mistake.

Worth settling because it is user-visible: with `private`, the generated `CHANGELOG.md` link resolves to `/pull/13300` and public issue #5833 never appears in the release notes. If `public` is intended, the file also needs renaming to `.changelogs/5833.json` per the README's filename rule.

Separately, on line 3 the title's trailing clause misparses. "a function returning objects with read-only members **or accessor-based columns**" attaches `or` to `returning`, but accessor columns come from `columns[].data`, not from what `dataSchema` returns. Suggested: "... and stopped recording further actions when `dataSchema` was a function, either because the row objects had read-only members or because the columns used function `data` accessors."

handsontable/src/plugins/undoRedo/actions/removeRow.ts
```diff
@@ -75,24 +135,18 @@ export class RemoveRowAction extends BaseAction {
...
-          // `nestedRows` manages its `__children` tree restoration on undo
-          // independently; writing `__children` back via `setSourceDataAtCell`
-          // would clash with the plugin's internal state.
-          if (rowData && typeof rowData === 'object' && !Array.isArray(rowData)) {
-            delete (rowData as Record<string, unknown>).__children;
-          }
-
-          removedData.push(rowData);
+          removedData.push(captureRowData(hot, physicalRowIndex + i));
+          removedAccessorValues.push(captureAccessorValues(hot, physicalRowIndex + i));
```
demtario: **low:** `captureAccessorValues` repeats row-invariant work inside the per-row loop -- it rescans every column for each removed row, though the set of accessor columns does not depend on the row.

To be clear about the scale, because my first read overstated it: this is **not** a new complexity class and not a perf regression. The adjacent `removedCellMetas: getCellMetas(hot, physicalRowIndex, lastRowIndex, 0, hot.countCols() - 1)` on line 155 already walks removed-rows x columns, and `utils.ts:14-26` shows it calls `hot.getCellMeta` per cell -- the allocating variant, heavier per cell than `colToProp` + `toPhysicalColumn`. So the existing pass dominates.

It is still redundant work with a trivially correct fix: hoist the scan above the `for (let i = 0; i < amount; i++)` loop, build the `[physicalColumn, prop]` pairs once, bail out early when the list is empty, then per row only invoke each accessor. Reads more clearly too, since it makes the row-independence explicit. Fine to skip if you would rather not touch it.

handsontable/src/dataMap/dataMap.ts
```diff
@@ -897,7 +910,13 @@ class DataMap {
       value = out;
 
     } else if (typeof prop === 'function') {
-      value = (prop as (row: unknown) => unknown)(this.dataSource!.slice(physicalRow, physicalRow + 1)[0]);
+      const sourceRow = this.dataSource![physicalRow];
+
+      // A row index that is mapped but not (yet) backed by source data – e.g. mid-way through an
+      // undo of a row removal – must read as empty instead of handing `undefined` to user code.
+      if (sourceRow !== undefined) {
+        value = prop(sourceRow);
```
budnix: The guard is right, but this branch reads a different row than the rest of the method. Eight lines up, `dataRow` holds the result of the `modifyRowData` hook, and every other branch reads through it. This one goes back to `this.dataSource[physicalRow]` and skips the hook, so with `nestedRows` -- or any user `modifyRowData` listener -- an accessor column and a string-prop column in the same row can show data from two different rows. Reading `dataRow` covers the missing-row case too, since it is the raw source row whenever no listener swaps it.

```suggestion
      // A row index that is mapped but not (yet) backed by source data – e.g. mid-way through an
      // undo of a row removal – must read as empty instead of handing `undefined` to user code.
      if (dataRow) {
        value = prop(dataRow);
```

handsontable/src/dataMap/dataMap.ts
```diff
@@ -981,7 +1000,7 @@ class DataMap {
     const { dataDotNotation } = this.hot!.getSettings();
 
     // try to set value under property `prop` (includes dot)
-    if (dataRow && hasOwnProperty(dataRow, prop)) {
+    if (dataRow && typeof prop !== 'function' && hasOwnProperty(dataRow, prop)) {
```
budnix, line 1004: `get()` now skips the accessor when the source row is missing, but `set()` still calls it. This line picked up the `typeof prop !== 'function'` half of the change; the function branch further down still does `prop(this.dataSource.slice(physicalRow, physicalRow + 1)[0], newValue)`, and that first argument is `undefined` in exactly the state the new `dataMap.unit.js` test builds by hand -- a row mapper one index longer than the source array, which is what `RemoveRowAction.undo` creates between `setIndexesSequence` and `alter`. A `setDataAtCell` that lands in that window calls the user's accessor with `undefined` and throws inside their code, which is the crash the read path was just fixed for. Would be good to mirror the guard here.

handsontable/src/plugins/undoRedo/undoRedo.ts
```diff
@@ -264,10 +264,18 @@ export class UndoRedo extends BasePlugin {
...
 
-    (action as { undo: (hot: HotInstance, callback: () => void) => void }).undo(this.hot, () => {
+    try {
+      (action as { undo: (hot: HotInstance, callback: () => void) => void }).undo(this.hot, () => {
+        this.ignoreNewActions = false;
+        this.undoneActions.push(action);
+      });
+
+    } catch (error) {
+      // An action that throws never reaches its settle callback. Without this reset every later
```
budnix, line 274: Small correction to this comment, and it cuts against the other thread on this line: for `RemoveRowAction` the settle callback *is* still reached. `removeRow.ts:222` arms `addHookOnce('afterViewRender', undoneCallback)` right before the `setSourceDataAtCell` on the next line, so when that write throws the hook is still armed. It fires on the next render and runs `undoneActions.push(action)`, putting an action whose undo only half finished onto the redo stack -- later, after this `throw` already left the plugin. So the entry is not dropped, it arrives late and on the wrong stack, and Ctrl+Y then redoes a removal that never fully came back. Disarming the hook when the write throws would make the failure clean.

handsontable/src/plugins/undoRedo/actions/removeRow.ts
```diff
@@ -130,6 +184,23 @@ export class RemoveRowAction extends BaseAction {
...
     });
 
+    // Accessor-column values live behind a function and are invisible to `data`; restore them
+    // through the accessor itself, the same way `dataSource.setAtCell` writes through it. The guard
+    // also bails out when the column has no visual index left – a column trimmed since the removal
+    // makes `toVisualColumn` return `null`, and `colToProp(null)` echoes `null` back, which is not
+    // an accessor. A hidden column keeps its visual index, so it restores like any other. A column
+    // trimmed at removal time is never captured either, because `captureAccessorValues` iterates
+    // `countCols()`, which does not count trimmed columns.
+    this.accessorValues.forEach((rowValues, rowIndexDelta) => {
```
budnix, line 238: Can I ask about the order here -- the data pass above pushes every own key of the snapshot into `changes`, and this pass appends after it, so both go into one `setSourceDataAtCell` batch and the data keys are written first. In the pattern the guide you just updated describes, the row object is `{ index: n }` and the accessor writes `source[columnIndex][row.index]`. On undo, `afterCreateRow` restamps `index` to the row's new slot, then this batch writes the captured pre-removal `index` back over it, and only then do the accessors read `row.index`. In the plain case the two values agree, so it works. Did you check a case where they don't -- a manual row move or a sort between the removal and the undo?

handsontable/src/plugins/undoRedo/actions/removeRow.ts
```diff
@@ -2,7 +2,57 @@ import type { HookCallback } from '../../../core/hooks/bucket';
...
+  const rowData = deepClone(hot.getSourceDataAtRow(physicalRow));
+
+  if (isPlainObject(rowData)) {
+    delete rowData.__children;
+
+    Object.keys(rowData).forEach((key) => {
+      if (typeof rowData[key] === 'function') {
+        delete rowData[key];
+      }
+    });
```
budnix: `deepClone` keeps functions at every level -- at `object.ts:127` a function is not an array, a `Date`, or an object, so it falls through to `return obj` -- but this loop only walks the top level. So `{ onSelect: fn }` is dropped and `{ meta: { onSelect: fn } }` is kept, even though the aliasing reason in the JSDoc applies to both. Worth a line in the changelog too: a row's own function property used to come back on undo and now does not, which shows up for anyone reading `getSourceDataAtRow` after an undo.

handsontable/src/helpers/object.ts
adrianspdev: Fixed in 3270713. `writeOwnProperty` now branches on descriptor kind first: an own accessor (configurable or not) is written through its setter and never redefined into a data property; a getter-only accessor is skipped; the non-configurable data-property handling stays as it was. Unit tests added for the configurable object-literal getter (asserts it keeps deriving after a later write to `a`) and a configurable setter-backed accessor.

On the second point: the changelog title now states the `setSourceDataAtCell()` read-only no-op explicitly, so that half of the change is documented rather than console-warned.

handsontable/src/base.ts
adrianspdev: Added in 3270713 — a `ColumnDataGetterSetterFunction` row in the Data types table next to `SourceRowData`, plus a reference from the binding-to-data guide's "Identify changed columns in hooks" section.

handsontable/src/plugins/undoRedo/actions/removeRow.ts
adrianspdev: Test added in 3270713: remove + undo with `hiddenColumns` active on an accessor column, asserting the value is captured and restored. The trimmed-column asymmetry stays as an accepted tradeoff — a trimmed accessor column's value is unreachable at capture time (`countCols()` does not see it), and no in-repo plugin trims columns; the code comment records the limitation.

handsontable/src/plugins/undoRedo/undoRedo.ts
adrianspdev: Went with the documented-discard option (3270713): a partially applied action cannot be safely replayed in either direction, so both catch blocks now state the discard is deliberate, and the two unit tests assert the throwing action lands on neither stack. The related leak on this line — the still-armed `afterViewRender` settle callback budnix found — is fixed in `RemoveRowAction.undo`; see that thread.

handsontable/src/dataMap/dataSource.ts
adrianspdev: Widened in 3270713: the hook's `@param` in `core/hooks/constants.ts` and `GridSettings['modifySourceData']` both take `number | string | ColumnDataGetterSetterFunction` now. The interface entry deliberately switched to method syntax so existing `(row, column: number, ...)` handlers stay assignable under `strictFunctionTypes`; a type regression in `undoRemoveRowAccessor.types.ts` pins both the widened shape and the pre-widening narrow handler.

handsontable/src/core.ts
adrianspdev: Fixed in 3270713. `getCellProperties` resolves a function prop through the accessor-aware `propToCol` cache, so an accessor column reads its own column meta; only an unresolvable accessor falls back to the table meta. Unit test asserts a column-level `valueSetter` runs when undo restores an accessor column.

.changelogs/13300.json
adrianspdev: Decision: the entry stays `private`/13300, following the `.changelogs/README.md` rule (work tracked in a private ClickUp task is `private`, which covers every `DEV-xxx` change) — so the release-notes link will point at the PR. The title misparse is fixed with your suggested wording in 3270713, and the title now also covers the `setSourceDataAtCell` read-only no-op and the function-property snapshot change.

handsontable/src/plugins/undoRedo/actions/removeRow.ts
adrianspdev: Hoisted in 3270713 — `collectAccessorColumns()` builds the `[physicalColumn, accessor]` pairs once per removal, and the per-row capture maps over them.

handsontable/src/dataMap/dataMap.ts
adrianspdev: Applied your suggestion in 3270713 — the branch reads through `dataRow`, and the guard covers the missing-row case. Unit test added with a `modifyRowData` listener swapping the row.

handsontable/src/dataMap/dataMap.ts
adrianspdev: Mirrored in 3270713 — `set()`'s function branch now goes through the `modifyRowData`-aware `dataRow` and skips the write when the row is missing. Unit test builds the same mapper-longer-than-source window and asserts the accessor is never called with `undefined`.

handsontable/src/plugins/undoRedo/undoRedo.ts
adrianspdev: Verified — the callback stayed armed and settled the half-undone action late onto the redo stack, exactly as you describe. Fixed in 3270713: `RemoveRowAction.undo` wraps the `setSourceDataAtCell` call and removes the `afterViewRender` hook before rethrowing. Unit test: after a throwing undo, a later `hot.render()` leaves the redo stack empty.

handsontable/src/plugins/undoRedo/actions/removeRow.ts
adrianspdev: Checked it — the two values cannot disagree through a row move or a sort, because both only remap visual indexes: the physical `rows` array (and its `index` stamps) never reorders, and `undo()` restores the pre-removal index sequence before `alter`, so the restored row re-enters at physical `this.index` — the same slot `afterCreateRow` restamps. Writing the captured `index` back over the restamp is therefore a no-op in every reachable ordering. What would desync them is a structural change between the removal and the undo, but undo is LIFO, so that action gets unwound first.

Pinned in 3270713 with two unit tests on the guide's exact columnar fixture, one with an interleaved reorder between removal and undo.

handsontable/src/plugins/undoRedo/actions/removeRow.ts
adrianspdev: Fixed in 3270713 — the sweep is now recursive (`stripFunctionValues`), so `{ meta: { onSelect: fn } }` is dropped along with the top level, and the changelog title states that a row's own function-valued properties no longer come back on undo. Unit test asserts the nested key is absent from the snapshot and after the undo.

.changelogs/13300.json
```diff
@@ -0,0 +1,8 @@
...
+  "title": "Fixed a bug where undoing a row removal threw an error, restored the row without its data, and stopped recording further actions when `dataSchema` was a function returning objects with read-only members or accessor-based columns.",
+  "type": "fixed",
+  "issueOrPR": 13300,
```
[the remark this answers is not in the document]
adrianspdev: Keeping `issuesOrigin: private` / `issueOrPR: 13300` by decision — `.changelogs/README.md` states that work tracked in a private ClickUp task is `private`, which covers every change with a `DEV-xxx` ID (this one is DEV-447).

demtario (approved):

PR #13301, opened 2026-08-28
DEV-2681: Expose cache update change sources

demtario:
### Context

The PR fixes DEV-2681 by exposing the source of an index sequence mutation on cache-update hooks. EditorManager now uses that direct signal to distinguish structural changes from rearrangements, removing the count-based heuristic.

### How has this been tested?

- `npm --prefix handsontable run eslint`
- `npm --prefix handsontable run stylelint`
- `npm --prefix handsontable run build`
- `npm --prefix handsontable run test:types`
- `npm --prefix handsontable run test:unit -- --testPathPattern=indexMapper`
- `HOT_TEST_PORT=8124 npx playwright test --project=e2e-main e2e/editor-trimmed-row.spec.ts`
- Runtime repro demo remains instrumented; post-fix manual log verification is pending.

### Types of changes

### Related issue(s):

1. DEV-2681

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

[skip changelog] DEV-2652 is unreleased, so this corrects a regression before it reaches users.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2681

---

> [!NOTE]
> **Medium Risk**
> Changes index-mapper cache hook contracts and editor reconciliation on structural vs rearrangement updates; behavior is well-tested but touches critical editing and index-translation paths.
> 
> **Overview**
> Adds **`indexesChangeSource`** (`init` | `remove` | `insert` | `move` | `update`) to the `IndexMapper` `cacheUpdated` payload and to the public **`afterRowSequenceCacheUpdate`** / **`afterColumnSequenceCacheUpdate`** hooks (typed via new **`IndexesChangeSource`**). The mapper keeps a **`#cacheUpdateSource`** so the source survives deferred cache flushes during batched operations; nested hook-driven cache updates clear the source, and **`insertIndexes`** no-ops when inserting zero indexes.
> 
> **EditorManager** stops inferring structural row/column changes from physical index **counts** and instead treats **`insert`** / **`remove`** as structural when repairing an open editor—addressing DEV-2681 and edge cases like cancelled inserts and same-length data swaps.
> 
> Tests and hook JSDoc are updated to assert the new field across init, move, insert, remove, update, and batch paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e6b398516479bd95cbdef15e23e6036b0584cc0a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2681 Expose indexesChangeSource in the cacheUpdated payload instead of inferring structural changes from index counts](https://app.clickup.com/t/9015210959/DEV-2681)

handsontable/src/translations/indexMapper.ts
```diff
@@ -900,11 +911,13 @@ export class IndexMapper {
         indexesSequenceChanged: this.indexesSequenceChanged,
         trimmedIndexesChanged: this.trimmedIndexesChanged,
         hiddenIndexesChanged: this.hiddenIndexesChanged,
+        indexesChangeSource: this.#cacheUpdateSource,
       });
 
       this.indexesSequenceChanged = false;
       this.trimmedIndexesChanged = false;
       this.hiddenIndexesChanged = false;
+      this.#cacheUpdateSource = undefined;
```
budnix: The source is cleared after the hook fires, so a listener that starts another cache update inside that same call still sees the old source. Pagination does exactly that - its cache-update handler rewrites the hiding map through `batchExecution` and can call `updateCache(true)` - so a hiding-only update can go out labelled `'insert'`. Capturing the value and clearing it before the emit keeps the payload honest. The three booleans have the same ordering, but flipping those is a bigger behavior call, so this only moves the new field.

```js
      const indexesChangeSource = this.#cacheUpdateSource;

      this.#cacheUpdateSource = undefined;

      this.runLocalHooks('cacheUpdated', {
        indexesSequenceChanged: this.indexesSequenceChanged,
        trimmedIndexesChanged: this.trimmedIndexesChanged,
        hiddenIndexesChanged: this.hiddenIndexesChanged,
        indexesChangeSource,
      });

      this.indexesSequenceChanged = false;
      this.trimmedIndexesChanged = false;
      this.hiddenIndexesChanged = false;
```

handsontable/src/translations/indexMapper.ts
```diff
@@ -203,6 +213,7 @@ export class IndexMapper {
   constructor() {
     this.indexesSequence.addLocalHook('change', () => {
       this.indexesSequenceChanged = true;
+      this.#cacheUpdateSource = this.indexesChangeSource;
```
budnix, line 218: `insertIndexes()` sets the source before calling `indexesSequence.insert()`, and `IndexMap#insert()` fires `change` even when the inserted list is empty. `fitToLength()` takes that branch on every `updateData()` where the count does not change, so a swap of equal length now reports `'insert'`. That flips the branch `tests/e2e/editor-trimmed-row.spec.ts:673` was written to pin - its comment says that case "routes to the rearrangement repair", which is no longer true. Worth a quick check whether a zero-amount insert should fire `change` at all.

handsontable/src/editorManager.ts
```diff
@@ -166,15 +127,27 @@ class EditorManager {
...
+   *
+   * @param {object} indexesChangesState The state object of the index mapper's cache update.
+   * @param {'init'|'remove'|'insert'|'move'|'update'} [indexesChangesState.indexesChangeSource] The sequence change source.
+   * @returns {boolean}
+   */
+  #isStructuralChange(indexesChangesState: {
+    indexesChangeSource?: 'init' | 'remove' | 'insert' | 'move' | 'update';
+  }): boolean {
+    return indexesChangesState.indexesChangeSource === 'insert' ||
+      indexesChangesState.indexesChangeSource === 'remove';
```
budnix, line 150: `'update'` and `'init'` can also change the number of physical indexes, and neither counts as structural here. `dataMap.ts` calls `setIndexesSequence([])` at 615 and 674 when the last row or column goes, which reports `'update'`, and `loadData()` runs `initToLength()` and reports `'init'` - neither closes an open editor. The old count check caught both; now they reach `#reconcileEditorWithIndexMaps()`, which resolves a physical index captured against the old data. `'update'` is the awkward one, since `columnSorting` uses it for a same-length permutation too, so keeping the count as a second signal for `'update'` and `'init'` may be the safer shape.

handsontable/src/editorManager.ts
```diff
@@ -166,15 +127,27 @@ class EditorManager {
...
+  }
   /**
    * Applies the repair one axis's cache update calls for, then lets the hidden-cell guard run.
    *
```
budnix, line 154: Just below this, `@param {boolean} isStructuralChange` still says "Whether that axis's physical index count just changed", and `#reconcileEditorWithIndexMaps()`'s JSDoc still says the two handlers "discriminate on the physical index count". `tests/e2e/editor-trimmed-row.spec.ts:521` repeats it. Nothing counts indexes anymore, so those send the next reader to a mechanism that is gone.

handsontable/src/core/settings.ts
```diff
@@ -317,6 +317,7 @@ export interface GridSettings {
   afterCustomBordersUpdate?: () => void;
   afterColumnSequenceCacheUpdate?: (indexesChangesState: {
     indexesSequenceChanged: boolean; trimmedIndexesChanged: boolean; hiddenIndexesChanged: boolean;
+    indexesChangeSource?: 'init' | 'remove' | 'insert' | 'move' | 'update';
```
budnix: The same union appears again at line 434, six more times in `editorManager.ts`, and twice in the hook JSDoc - nine copies with nothing tying them together. One exported `IndexesChangeSource` type would make a sixth source a one-line change. Also `indexMapper.ts:144` still types the field it comes from as plain `string | undefined`, so nothing checks the emitted value against what is published here.

handsontable/src/__tests__/core/core.unit.js
```diff
@@ -132,4 +132,30 @@ describe('Core', () => {
...
+      data: [['a', 'b'], ['c', 'd']],
+      afterRowSequenceCacheUpdate,
+      afterColumnSequenceCacheUpdate,
+    });
+
+    core.init();
+    afterRowSequenceCacheUpdate.calls.reset();
+    afterColumnSequenceCacheUpdate.calls.reset();
+
+    core.rowIndexMapper.insertIndexes(1, 1);
```
budnix, line 149: This calls `insertIndexes()` straight on the mapper, so it pins the payload but not the routing the change exists for - nothing asserts that `'insert'`/`'remove'` still go to the recapture and everything else to reconciliation. A batched `moveIndexes()` is not covered either, and that is the case the new field was added to survive; the batched test uses `setIndexesSequence`, which never goes through a move. Would be good to add both.

handsontable/src/core/hooks/constants.ts
```diff
@@ -334,6 +334,7 @@ export const REGISTERED_HOOKS = [
    * @param {boolean} indexesChangesState.indexesSequenceChanged Indicates if the sequence of indexes has changed.
    * @param {boolean} indexesChangesState.trimmedIndexesChanged Indicates if the trimmed indexes have changed.
    * @param {boolean} indexesChangesState.hiddenIndexesChanged Indicates if the hidden indexes have changed.
+   * @param {'init'|'remove'|'insert'|'move'|'update'} [indexesChangesState.indexesChangeSource] Indicates what caused a sequence change.
```
budnix, line 337: This adds a field to two public hook payloads, so the changelog gate will want an entry in `.changelogs/`. Nothing in the diff adds one, and the PR touches `handsontable/src/**`, so the docs/test/CI exemption does not cover it.

docs/superpowers/plans/2026-08-28-dev-2681-cache-update-source.md
```diff
@@ -0,0 +1,205 @@
+# DEV-2681 cache-update source Implementation Plan
```
budnix: Did this end up in the commit on purpose? `docs/` is tracked, so a working plan with agent instructions and checkboxes ships with the repo and goes stale as soon as the design moves. Its architecture section already misses the `'init'` and `'update'` cases noted in the other comments.

handsontable/src/translations/__tests__/indexMapper.unit.ts
```diff
@@ -746,6 +746,75 @@ describe('IndexMapper', () => {
...
+      ];
+
+      testCases.forEach(({ source, change }) => {
+        const indexMapper = new IndexMapper();
+        const cacheUpdatedCallback = jasmine.createSpy('cacheUpdated');
+
+        indexMapper.addLocalHook('cacheUpdated', cacheUpdatedCallback);
+        change(indexMapper);
+
+        expect(cacheUpdatedCallback).toHaveBeenCalledWith(jasmine.objectContaining({
```
Haxikowy: `toHaveBeenCalledWith` passes when any recorded call matches, and every case runs `initToLength(3)` first, so the spy always holds at least two calls. A stale or leaked source on the final emission would not fail this, and the `init` row passes trivially. Asserting `cacheUpdatedCallback.calls.mostRecent().args[0]` pins the emission you mean, and `it.each` would name the failing source instead of reporting one anonymous title for all five.

Haxikowy (commented):
Two things I did not see covered in the existing review.

`dataMap.ts:405` calls `rowIndexMapper.insertIndexes()` before the `numberOfCreatedRows > 0` guard on line 421. An `alter('insert_row')` that creates nothing, because `maxRows` is reached or `beforeCreateRow` vetoes, therefore emits source `'insert'` and repairs an open editor as though rows shifted. The heuristic this PR removes was immune to a vetoed insert and its JSDoc said so, so this is a second site for the zero-amount insert problem already raised on `indexMapper.ts:216`, and a fix there would cover it.

`#cacheUpdateSource` is last-write-wins, and `hot.batch()` suspends both mappers, so two sequence mutations in one batch collapse to a single emission carrying only the later source. `hot.batch(() => { trimRow(0); alter('insert_row_above', 0) })` emits one update with `trimmedIndexesChanged: true` and source `'insert'`, so only the recapture branch runs and the trim goes unreconciled. Reasoned from the code path, not executed. Today the structural sources survive by accident, because `isBatched` is a boolean rather than a counter and `insertIndexes()`/`removeIndexes()` call `resumeOperations()` unconditionally, which ends the outer batch early. Making the latch refuse to overwrite `'insert'`/`'remove'` with a rearrangement source would keep the signal once that is fixed.

handsontable/src/core/settings.ts
demtario: Fixed in e6b3985. IndexesChangeSource is now defined once in IndexMapper and used by the mapper, GridSettings, and EditorManager; generated type tests pass.

handsontable/src/__tests__/core/core.unit.js
demtario: Expanded coverage on the branch: public row and column cache-update hooks assert every source, and mapper tests include a nested cache update plus source-specific test names. The editor regression suite covers structural and rearrangement routing.

handsontable/src/translations/indexMapper.ts
demtario: Confirmed with runtime tracing. A zero-amount insert fired the sequence change and emitted insert; e6b3985 now returns before preparing the mutation when the amount is zero. The regression test covers this case.

handsontable/src/editorManager.ts
demtario: Fixed in e6b3985. The editor and regression-test comments now describe source-based classification rather than the removed physical-count heuristic.

handsontable/src/core/hooks/constants.ts
demtario: No changelog entry is required here: DEV-2681 corrects unreleased DEV-2652 behavior, and the PR description contains the documented [skip changelog] exception.

handsontable/src/translations/indexMapper.ts
demtario: Fixed in e6b3985: capture and clear the pending source before notifying cache-update listeners, so a reentrant update receives no inherited source.

docs/superpowers/plans/2026-08-28-dev-2681-cache-update-source.md
demtario: Removed in e6b3985. The implementation plan was working material and should not be committed with the product change.

handsontable/src/editorManager.ts
demtario: Validated the data-swap paths after the no-op insertion fix: both equal-length and reduced-length editor regressions pass. The setIndexesSequence([]) path only runs after the opposing axis becomes empty, leaving no editable cell. I left update/init on reconciliation rather than restoring count-based classification.

handsontable/src/translations/__tests__/indexMapper.unit.ts
demtario: Addressed in e6b3985: setup runs before the listener is registered, each case asserts exactly one emission, the expectation reads mostRecent().args[0], and sources have distinct test titles.

budnix (approved):

Haxikowy (approved):

PR #13304, opened 2026-08-28
DEV-2680: Keep selection with moved editor

demtario:
### Context

The PR fixes edits being discarded or written through stale visual coordinates when a trimming map moves the record under an open editor. It reconciles the active selection with the editor's rebound coordinates and avoids re-entering the refresh path when no coordinates changed.

`IndexMapper` now fires `beforeCacheUpdate` / `afterCacheUpdate` around a cache rebuild (`afterCacheUpdate` guaranteed through `try/finally`), because `updateCache()` rebuilds every cache *before* `cacheUpdated`, so the pre-update visual space cannot be recovered afterwards. `Core` uses that lifecycle on both mappers: for a **trim-only** update (`trimmedIndexesChanged` and not `indexesSequenceChanged`), `Selection` snapshots every layer in **physical** coordinates before the swap and restores them after it, shrinking a partially trimmed range onto its surviving records and dropping a dead non-active layer without firing selection hooks. A sequence permutation is deliberately excluded: reordering can make one rectangle's records non-contiguous, which `CellRange` cannot describe.

The snapshot is taken only while an editor is open, so the sibling repair for a selection with **no** editor (merged in #13310) keeps its own half untouched: it drops a highlight a trim left pointing at another record, this one rebases the selection the editor owns.

#### Merged develop, and what the two repairs had to agree on

#13310 landed on develop after this branch opened and rewrote the same three areas: the cache-update hooks in `core.ts`, the trim docblock in `editorManager.ts`, and the `EditorTrimmedRowPage` fixture. The merge keeps develop's frame (`IndexesChangesState`, the `hadOpenEditor` sampling, `repairSelection()`) and splices the snapshot hooks into it, with the restore running **before** `commit()` so the ranges are back on their surviving records before the highlight is walked off hidden cells.

Two integration defects the clean auto-merge of `selection.ts` did not surface, both fixed here:

- **Layer-keyed state was left behind by the restore.** develop's `#rowExtentSpansGrid` / `#columnExtentSpansGrid` are keyed by layer index, and the restore re-indexes the layers it keeps. Both sets now go through `#remapLayerSet()` (renamed from `#remapHeaderLayers()`, which four sets share) via the snapshot. Unfixed, a surviving layer inherits a dropped one's grid-tracking flag, and a full-column selection is then judged as naming records, or the reverse.
- **A restore-time drop stopped announcing itself.** The merge routes one shape develop already handled through the restore instead of `deselectIfHighlightStranded()`, and the restore cleared the selection with `clear()`, so the public `afterDeselect` stopped firing there. It cannot be sent from the restore: `Core` answers `afterDeselect` with `EditorManager#closeEditor()`, which **saves** the pending value, and the restore runs before the public cache-update hooks — closing the editor there would commit the edit through the very coordinates the drop abandons, appending records. `#clearForRestore()` therefore owes the notification and `Selection#notifyDeferredDeselect()` sends it after those hooks, where the editor is already discarded and closing is a no-op. It stays silent when anything re-selected in between, which `Filters#filter()` does.

One behavior question the two repairs answered differently, settled in favor of develop's rule: when a trim removes **every** record of the active range, the range parks on its focus's pre-update slot only while the trim leaves that slot **in range** — the editor is discarded, and the next keystroke re-prepares against the record now sitting there, so the write lands on a real record. When the slot addresses nothing the whole selection is cleared instead: never clamped inward onto a record the trim merely slid into view, and never replaced by promoting a bystander layer into the active slot. That is the same "drop, never clamp" rule the no-editor repair applies, and both ends are pinned as tests.

- Unit tests added/modified (`*.unit.js`): `handsontable/src/translations/__tests__/indexMapper.unit.ts` — the new `beforeCacheUpdate` / `afterCacheUpdate` hooks.
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`): `tests/e2e/editor-trimmed-row.spec.ts`
- Type tests (`*.types.ts`) updated if public API changed: none — no public API changed.
- For a bug fix — the spec that fails without this fix: `commit paths that read the selection after trimming` in `tests/e2e/editor-trimmed-row.spec.ts`
- Merge regressions pinned: `drops a fully trimmed active range whose own slot the trim left out of range` and `collapses a fully trimmed active range onto one parked cell` (the two ends of the parking rule, both asserting `afterDeselect`), plus develop's own `tests/e2e/selection-trimmed-row.spec.ts` and `tests/e2e/selection-trim-shapes.spec.ts` kept green unchanged.
- Demo page / recorded trace (for UI changes): local repro validated before removing diagnostics.

### Commands run

- `npm --prefix handsontable run build` — passed.
- `handsontable/node_modules/.bin/tsc --noEmit -p handsontable/tsconfig.json` — passed.
- `npx eslint handsontable/src/core.ts handsontable/src/editorManager.ts handsontable/src/selection/selection.ts` — passed.
- `npm --prefix handsontable run stylelint` — passed.
- `npm --prefix handsontable run test:unit` — 4212 passed.
- `npm --prefix handsontable run test:e2e` — 9868 specs, 0 failures.
- `HOT_TEST_PORT=8124 npx playwright test --project=e2e-main` — 745 passed, 1 skipped (full suite, after the merge).

### Types of changes

### Related issue(s):
1. DEV-2680

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

---

> [!NOTE]
> **High Risk**
> Touches index-map cache lifecycle, selection repair, and editor commit ordering on every trim/hide cache update; regressions could mis-route fills, deselect, or nested mapper updates.
> 
> **Overview**
> Fixes **selection-dependent edits being lost or written to the wrong rows** when a trimming map (e.g. `trimRows`, Filters) runs while a cell editor is open. Commits that read the **active selection**—`DropdownEditor` and **Ctrl+Enter** fill—now follow the records the user edited instead of stale visual coordinates or silent discards.
> 
> **`IndexMapper#updateCache()`** now fires **`beforeCacheUpdate`** (pre-rebuild mappings) and **`afterCacheUpdate`** (always, via `finally`) around the cache swap, documented in `.ai/INDEX-MAPPING.md`.
> 
> **`Core`** hooks both row/column mappers: on **trim-only** updates (`trimmedIndexesChanged` without `indexesSequenceChanged`), **`Selection`** **captures** every layer in physical coordinates before the rebuild and **restores** them on `cacheUpdated` (before `commit()`). Sort/move permutations are excluded. Restore-time full clears **defer `afterDeselect`** until after public cache hooks so closing the editor does not save through abandoned coordinates. Layer-keyed state (`…ExtentSpansGrid`, header markers) is **remapped** after restore.
> 
> **`Selection`** adds the snapshot stack (per axis, for nested updates), shrink-onto-survivors logic, grid-tracking vs record-named extents, and **`#coversForeignRecords()`** when an untrim reveals rows inside a range.
> 
> E2E in `editor-trimmed-row.spec.ts` and unit tests for the new mapper hooks; changelog **#13304**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 53e107c41d68633124382793e700482a1d66fa3a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2680 An open editor's selection does not follow its record, so dropdown and Ctrl+Enter edits are silently lost](https://app.clickup.com/t/9015210959/DEV-2680)

handsontable/src/editorManager.ts
```diff
@@ -693,8 +693,41 @@ class EditorManager {
...
+      const { from, to, highlight } = activeSelectionRange;
+
+      if (from.row !== null && from.row >= 0) {
+        from.row += rowOffset;
+      }
+      if (from.col !== null && from.col >= 0) {
+        from.col += columnOffset;
+      }
+      if (to.row !== null && to.row >= 0) {
+        to.row += rowOffset;
```
Haxikowy: A trim inside the selected span makes this offset wrong for `to`: with rows 0..9, the range 3..7, the editor at 5 and row 4 trimmed, the range becomes 2..6 and Ctrl+Enter writes the value into row 2, which the user never selected. A sequence permutation is worse, because `moveRow` or `columnSorting` can collapse a 5-row range to a single cell once `refresh()` clamps the overshoot, so each corner needs re-translating from its own captured physical index rather than inheriting the editor's delta.

handsontable/src/editorManager.ts
```diff
@@ -693,8 +693,41 @@ class EditorManager {
...
+
     editor.row = visualRow;
     editor.col = visualColumn;
+
+    const activeSelectionRange = this.selection.getActiveSelectedRange();
+
+    if (activeSelectionRange) {
+      const { from, to, highlight } = activeSelectionRange;
+
+      if (from.row !== null && from.row >= 0) {
```
Haxikowy: The `>= 0` test reads the value before the shift, so a negative offset can still land below zero: select rows 0..4, edit row 4, then `trimRows([0, 1])` gives `from.row === -2`, and `refresh()` clamps only the upper bound (`selection.ts:1661`) so it survives into `getSelected()`. One trim less lands on exactly -1 and the range then reads as a column selection with `selectedByColumnHeader` empty, so clamp each shifted corner to `>= 0`, ideally in one helper going through `CellCoords#assign()` instead of six copies of this rule.

handsontable/src/editorManager.ts
```diff
@@ -693,8 +693,41 @@ class EditorManager {
...
+      if (to.col !== null && to.col >= 0) {
+        to.col += columnOffset;
+      }
+      if (highlight.row !== null && highlight.row >= 0) {
+        highlight.row += rowOffset;
+      }
+      if (highlight.col !== null && highlight.col >= 0) {
+        highlight.col += columnOffset;
+      }
+      this.selection.refresh();
```
Haxikowy: This runs `prepareEditor()` synchronously inside the mapper's `cacheUpdated` unwind, because `refresh()` calls `finish()` before `markEndSource()` and `afterSelectionFinished` at `core.ts:906` fires on the `'refresh'` source, which rebinds `editor.row` and `editor.col` from the highlight and supersedes lines 703-704. When the record lands on a `readOnly` column or beside a hidden row, `isCellEditable()` fails and `clearActiveEditor()` nulls the reference on an editor that is still open, leaving its value unreachable and `#closeEditorWhenCellHidden()` with nothing to close, so defer this out of the cache update the way the sibling method at :803 documents.

Haxikowy: Nothing marks a selection source here, and `refresh()` self-marks `'refresh'` only when the source is already `'unknown'`, so a trim arriving while the source is `'mouse'` or `'keyboard'` falls outside the exclusion list at `core.ts:872` and closes the editor being repaired. `'refresh'` is also absent from `ignoreScrollSources` (`core.ts:841`), so filtering or sorting with an editor open now jumps the viewport; bracketing the call with `markSource('shift')` and `markEndSource()` fixes both, since `'shift'` is already in both lists.

handsontable/src/editorManager.ts
```diff
@@ -693,8 +693,41 @@ class EditorManager {
       return;
     }
 
+    if (visualRow === editor.row && visualColumn === editor.col) {
```
Haxikowy: This returns before the selection repair whenever the editor's own coordinates happen not to move, so trimming rows below the edited record leaves `to.row` past the last row and Ctrl+Enter appends records, the corruption 13272's changelog names. It is also load-bearing against re-entrancy now, since `IndexMapper#updateCache` resets `indexesSequenceChanged` only after the hook and a nested update would otherwise shift the range twice, so key the repair off the range's own corners instead of the editor's and say so where the guard lives.

handsontable/src/editorManager.ts
```diff
@@ -693,8 +693,41 @@ class EditorManager {
...
+      return;
+    }
+
+    const rowOffset = visualRow - (editor.row ?? visualRow);
+    const columnOffset = visualColumn - (editor.col ?? visualColumn);
+
     editor.row = visualRow;
     editor.col = visualColumn;
+
+    const activeSelectionRange = this.selection.getActiveSelectedRange();
```
Haxikowy: `getActiveSelectedRange()` returns one layer while `refresh()` re-applies all of them, so a ctrl-click selection ends up with the active block following the data and the rest sitting on whatever slid into their rows. `Selection.shiftRows()` and `shiftColumns()` already do this job with bounds clamping, sentinel handling, `snapToNearestVisible` and `markSource('shift')`, and `nestedRows.ts:1123` already calls `selection.refresh()` from the plugin layer for the same stated reason, so this belongs in Selection rather than writing integers into its `CellCoords` from here.

tests/e2e/editor-trimmed-row.spec.ts
```diff
@@ -616,13 +616,33 @@ test.describe('an index-map change on the column axis', () => {
...
+   * record moves, only real visual coordinates may move; shifting the sentinel makes the selection
+   * invalid.
+   */
+  test('preserves a column-header selection when the edited record moves',
+    async({ page, theme, bundle }) => {
+      const grid = new EditorTrimmedRowPage(page, theme, bundle);
+
+      await grid.goto();
+      await grid.openEditorAndType(0, 0, 'EDITED');
+      await grid.setActiveRangeStartRow(-1);
```
Haxikowy: `setActiveRangeStartRow(-1)` builds a range `selectColumns()` can never emit (`from=[-1,0]`, `to=[0,0]`, `selectedByColumnHeader` empty), so this exercises one guard line and nothing about column-header selection. The assertion that flips on develop is `to.row` moving 0 to 4 rather than the sentinel the name claims to preserve, and there is no `sourceData()` or `sourceRowCount()` call despite the file's rule at :20-21; a real `selectColumns(0)` would also push `to.row` past a 5-row grid, which this shape cannot reach.

tests/fixtures/pages/EditorTrimmedRowPage.ts
```diff
@@ -394,6 +401,15 @@ export class EditorTrimmedRowPage {
     await this.page.keyboard.press('Enter');
   }
 
+  /**
+   * Commits a multi-cell edit with the platform-specific Control or Meta modifier.
+   */
+  async commitWithCtrlOrMetaEnter(): Promise<void> {
+    const modifier = await this.page.evaluate(() => navigator.platform.includes('Mac') ? 'Meta' : 'Control');
```
Haxikowy: Playwright's `ControlOrMeta` covers this and is already the convention here, documented at `SelectionFeaturesPage.ts:227` and used throughout `clipboard.spec.ts`, while `navigator.platform` is deprecated. Every Playwright leg runs on `ubuntu-latest`, so the `Meta` branch never executes in CI: `await this.page.keyboard.press('ControlOrMeta+Enter')` drops both the evaluate and the dead branch.

.changelogs/13304.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Fixed cell edits being lost after rows are trimmed.",
```
Haxikowy: This lands in the same unreleased `Fixed` section as `.changelogs/13272.json`, whose title says the edit "is discarded when a filter or trim takes that record out of view", so the two entries read as contradicting each other. The title also names only trimming while the tests cover `manualRowMove` and a dropdown commit, and the change users observe is that the selection now moves with the record, so amending 13272's entry may be cleaner than adding this one.

Haxikowy (commented):
Five findings with no line in the diff to attach to.

`selection.ts:1651-1672`, reached from `editorManager.ts:729`: `refresh()` never restores the active layer, unlike `importSelection` at :1597. `setRangeEnd` sets it to `getLayerLevel()`, so with ctrl-clicked ranges the active layer becomes the last range and the following commit reads a different layer than the editor's, which is the class of bug this PR fixes.

`selection.ts:1646-1672`: the `selectedByRowHeader` and `selectedByColumnHeader` sets are restored after the range loop, so every hook fired inside it sees them empty. `core.ts:851-871` strips `ht__selection--rows` and `--columns` from the root element and never re-adds them, pagination skips its `isSelectedByColumnHeader()` off-page snap, and `viewportScroller` picks the wrong strategy. The new test's scenario is a header selection, so this sits on the happy path of the change.

`editorManager.ts:588-620` and `tests/e2e/editor-trimmed-row.spec.ts:606-617`: both doc blocks still state that the selection does not follow the record, and both name DEV-2680 as future work that "inverts these cases". This PR is DEV-2680. The JSDoc at :615-617 ("An index-map change does NOT adjust the selection") is now false for this path, and the describe name `commit paths the rebind cannot reach` no longer fits two of its three tests.

Missing coverage: no unit test pins the offset arithmetic, and `from.col`, `to.col` and both `highlight` branches have no assertion anywhere in the suite, since plain Enter reads the editor's own coordinates and the column-axis tests commit that way. `handsontable/src/selection/__tests__/utils.unit.ts` already tests pure range transforms with no DOM, and would take a table test for the -1 arrival, the `to.row` overshoot, and the column mirror.

`selection.ts:1610-1612`: `exportSelection()` returns `activeRange` by reference while cloning `ranges`, and `dialog.ts:383` and `emptyDataState.ts:567` hold that snapshot. Mutating `from`, `to` and `highlight` in place therefore edits a live snapshot whose sibling clones stay frozen. I did not prove an index-map change lands inside that window, so treat reachability as open; routing through the Selection API removes the aliasing regardless.

handsontable/src/editorManager.ts
demtario: Fixed in b80ab07. Selection now snapshots every range corner in physical coordinates before the mapper rebuilds and translates each independently afterward, including arbitrary permutations.

handsontable/src/editorManager.ts
demtario: Fixed in b80ab07. Header coordinates are preserved as sentinels rather than offset or clamped; real coordinates are translated through the relevant mapper.

handsontable/src/editorManager.ts
demtario: Fixed in b80ab07. Reconciliation moved into Selection and directly reapplies highlights after the cache update. It does not call refresh() or enter the selection hook/editor-preparation pipeline while the mapper unwinds.

demtario: Fixed in b80ab07. The restore path does not call refresh(), preserves the existing selection source, and the regression test asserts no afterSelection hook is fired during the rebase.

handsontable/src/editorManager.ts
demtario: Fixed in b80ab07. The capture/rebase is keyed to every selection range and occurs for nonstructural mapper changes even when the editor coordinates do not move; it no longer relies on that early return.

handsontable/src/editorManager.ts
demtario: Fixed in b80ab07. The logic now belongs to Selection: it snapshots and restores every layer, active-layer index, header state, and selection source without mutating CellCoords from EditorManager.

tests/e2e/editor-trimmed-row.spec.ts
demtario: The implementation no longer shifts header corners, but the test setup is still synthetic rather than exercising selectColumns(0). This is not resolved by b80ab07; it needs that test revision.

tests/fixtures/pages/EditorTrimmedRowPage.ts
demtario: Confirmed. The current helper still selects the modifier through navigator.platform. ControlOrMeta is the established Playwright convention and gives CI coverage without the deprecated platform branch; this remains to be corrected.

.changelogs/13304.json
demtario: Updated in b80ab07 to “Fixed cell selections not following edited records after index changes.” Entry 13272 covers the case where the edited record is removed; this entry documents the complementary selection-dependent commits when the record survives a trim or permutation.

handsontable/src/selection/selection.ts
```diff
@@ -1616,6 +1631,88 @@ class Selection {
...
+
+    const activeRange = this.getActiveSelectedRange();
+
+    this.highlight.useLayerLevel(this.#activeSelectionLayer);
+    this.highlight.getFocus().clear();
+
+    if (activeRange && this.highlight.isEnabledFor(FOCUS_TYPE, activeRange.highlight)) {
+      this.highlight.getFocus().add(activeRange.highlight).commit().syncWith(activeRange);
+    }
+  }
```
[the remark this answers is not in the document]
demtario: Fixed in 720dfce. restorePhysicalSelection() now clears selection state when either visible axis is empty, without firing afterDeselect and triggering an editor commit. The new regression trims all rows during an open edit and verifies that no selection remains and source data is unchanged.

handsontable/src/selection/selection.ts
```diff
@@ -1616,6 +1631,94 @@ class Selection {
...
+      return;
+    }
+
+    if (this.tableProps.countRows() === 0 || this.tableProps.countCols() === 0) {
+      this.clear();
+
+      return;
+    }
+
+    this.selectedRange.ranges = snapshot.ranges.map(({ physical, visual }) => this.tableProps.createCellRange(
```
budnix: Each corner is mapped on its own. That works for a trim, but not for a move or a sort, where rows change their order relative to each other. With rows A,B,C,D,E select rows 0-2, open the editor, then `moveRow(4, 1)`: `from` follows A to visual 0 and `to` follows C to visual 3, so the range grows to cover a row the user never picked — and `Ctrl+Enter` writes into it, since `BaseEditor#saveValue()` reads exactly these corners. The same split mapping can leave `highlight` outside `from..to`, and `DropdownEditor#finishEditing()` discards the commit in that case, which is the loss this PR set out to fix. Both new `moveRow` tests use a 1x1 range where all three corners move together, so would be good to add one where the moved row crosses the selected span.

handsontable/src/selection/selection.ts
```diff
@@ -1616,6 +1631,94 @@ class Selection {
...
+      this.#createVisualCoords(physical.highlight, visual.highlight),
+      this.#createVisualCoords(physical.from, visual.from),
+      this.#createVisualCoords(physical.to, visual.to),
```
budnix: `selectColumns(0)` sets `to.row = countRows - 1` as a real index, not a header sentinel, so it gets snapshotted and rebased like any other coordinate. After `moveRow(0, 4)` that physical row sits at visual 3, and the "whole column" selection quietly loses its last row — while `selectedByColumnHeader` is restored as-is and still claims a header selection, so `isEntireColumnSelected()` and that set now disagree and the header highlight drops. Pinning `to.row` to the new last row when the layer is in `selectedByColumnHeader` would keep the two in step. Worth noting the test above passes only by luck: `setActiveRangeStartRow(-1)` leaves `to.row = 0`, and physical 0 lands on the last row after the move, so a real `selectColumns(0)` range would fail it.

handsontable/src/selection/selection.ts
```diff
@@ -1678,6 +1781,104 @@ class Selection {
...
+    indexMapper: IndexMapper,
+    fallbackIndex: number | null,
+    count: number,
+  ): number | null {
+    if (index !== null && index < 0) {
+      return index;
+    }
+
+    return (index === null ? null : indexMapper.getVisualFromPhysicalIndex(index)) ??
+      clamp(fallbackIndex ?? 0, 0, Math.max(count - 1, 0));
```
budnix: When a record was trimmed away `getVisualFromPhysicalIndex()` returns `null`, and this falls back to the old visual index clamped into range — which is now a different record. So the range is not dropped, it silently moves onto someone else's row, and a later `Ctrl+Enter` or fill writes there. The JSDoc says editor reconciliation cleans this up afterwards, but that only covers the active layer's editor, not the other layers. Curious whether returning `null` and dropping the range would be safer than picking a neighbour.

handsontable/src/selection/selection.ts
```diff
@@ -1616,6 +1631,94 @@ class Selection {
...
+   * Captures every selection layer in physical coordinates before an index mapper replaces its
+   * visual-to-physical cache. Header coordinates remain untouched because they are sentinels, not
+   * indexes. The snapshot is intentionally limited to a live editor because only that path needs
+   * its selection to follow a record through a trim or permutation.
+   */
+  capturePhysicalSelection(): void {
+    const activeRange = this.getActiveSelectedRange();
+
+    if (!this.tableProps.isEditorOpened() || !activeRange ||
+        !this.#hasResolvedPhysicalCoords(activeRange.highlight)) {
```
budnix: The guard checks only the active range's highlight, but the snapshot below captures every layer. A non-active layer holding coordinates that no longer resolve still gets captured, its indexes come back as `null`, and `#getVisualIndex` clamps them onto real rows — so that layer ends up on records it never covered, and `applyAndCommit` paints a highlight there. Checking each range as it is captured and skipping the ones that do not resolve would keep the guard and the snapshot talking about the same thing.

handsontable/src/selection/selection.ts
```diff
@@ -1616,6 +1631,94 @@ class Selection {
...
+      this.#physicalSelectionSnapshot = null;
+
+      return;
+    }
+
+    this.#physicalSelectionSnapshot = {
```
budnix: The snapshot is only consumed by the matching `cacheUpdated`, and a fair amount runs in between — both collection `updateCache()` calls and the `hidingChangesObservable` emit. If a user map throws in there, `cacheUpdated` never fires and this field keeps its pre-update ranges, so the next qualifying update restores coordinates from a mapping state that is two updates old. A `try/finally` around the body of `IndexMapper#updateCache` would close that. Separate thought: both mappers write this one field, so if anything on that path ever triggers the other axis's update, the inner capture overwrites the outer one — I could not find a plugin that does it today, but one slot per axis costs nothing.

handsontable/src/editorManager.ts
```diff
@@ -586,38 +586,18 @@ class EditorManager {
...
-   * record, which is a larger change than this repair: DEV-2680.
-   *
-   * Both exceptions are pinned by cases in `tests/e2e/editor-trimmed-row.spec.ts` under `commit paths
-   * the rebind cannot reach`. Those cases assert the LOSS on purpose, so a regression back to a write
-   * fails – they are not a statement that losing the edit is the desired end state. DEV-2680 inverts
-   * them.
+   * `Selection` separately snapshots every layer in physical coordinates before the cache rebuild and
+   * restores it afterwards. Keeping that operation out of this manager prevents selection hooks from
+   * preparing an editor while the mapper is still unwinding. It lets editor-specific commit paths,
+   * including DropdownEditor and Ctrl+Enter, keep targeting the moved record.
```
budnix: This now promises the commit paths keep targeting the moved record with no limits attached, but that only holds for trims — under a sort or a move the corner-by-corner rebase can widen the range or push the focus out of it, so both named paths still misbehave. The old text was careful about which paths were covered and which were not; keeping that shape and naming reorders as the remaining gap would stop the next reader from assuming it is all handled. Also, the removed paragraph about an index-map change with no editor open was the only place that defect was written down, and `capturePhysicalSelection()` still returns early unless `isEditorOpened()` — so it is still there. Was dropping it deliberate?

handsontable/src/core.ts
```diff
@@ -793,22 +793,48 @@ export default function Core(
 
   this.selection = selection;
 
-  const onIndexMapperCacheUpdate = ({ hiddenIndexesChanged }: { hiddenIndexesChanged: boolean }) => {
+  let rowIndexCount = this.rowIndexMapper.getNumberOfIndexes();
```
budnix: These two counters are the same "did the physical space change size" check `EditorManager` already keeps in `#lastRowIndexCount`/`#lastColumnIndexCount`, and its JSDoc calls out that the seeding point matters — it is built at core.ts:1778, after `updateSettings()`, so its first cache update is a real one. These seed at line 796, before `updateSettings()` at 1773, when the mappers hold nothing yet, so the two copies start from different places and can answer the same question differently. The blind spot comes along too: remove and insert the same number of rows in one flushed update and this reads as a reorder, so the rebase maps captured physical indexes through renumbered ones. `EditorManager` documents that at lines 133-136 and points at DEV-2681 as the fix that deletes both copies — could this reuse the counter that already exists rather than adding a second one?

handsontable/src/selection/selection.ts
demtario: Fixed in 20ac8d1. Unresolved non-active ranges are dropped instead of clamped onto another record. If only a former active range is dropped, surviving layers remain and one becomes active. The only fallback retained is when the active highlight itself was trimmed: EditorManager cancels that editor immediately, while the visual slot remains selected so the existing continue-editing behavior is preserved.

handsontable/src/selection/selection.ts
demtario: Fixed in 20ac8d1. The test now creates the range with the real selectColumns() API and verifies both the full visual bounds and isEntireColumnSelected() after trimming. Whole-column and whole-row layers pin the opposite-axis end to the current last index. Sequence moves are no longer physically rebased, so they cannot shorten a full-column range.

handsontable/src/selection/selection.ts
demtario: Fixed in 20ac8d1. Selection rebasing is now limited to pure trimming updates. Moves and sorts keep the visual range unchanged, so a permutation cannot widen a rectangular range onto unrelated records. Added an E2E regression where moving a row across a multi-cell selection leaves its bounds unchanged, and documented permutations as the remaining limitation.

handsontable/src/selection/selection.ts
demtario: Fixed in 20ac8d1. capturePhysicalSelection() now validates each layer on the axis being rebuilt and snapshots only ranges whose highlight, from, and to coordinates all resolve. Original layer indexes are retained so active and header-layer state can be remapped after invalid layers are removed.

handsontable/src/selection/selection.ts
demtario: Fixed in 20ac8d1. Row and column snapshots now use separate slots. IndexMapper passes the pending change state to beforeCacheUpdate and always fires afterCacheUpdate from a finally block; Core discards that axis snapshot there, so a thrown cache rebuild or nested update cannot strand or overwrite stale selection state. Unit coverage pins the exception path.

handsontable/src/editorManager.ts
demtario: Fixed in 20ac8d1. The JSDoc and E2E describe block now state that selection reconciliation applies to pure trimming updates only. They explicitly document why permutations remain unsupported for selection-dependent commits and restore the note that selection changes with no open editor remain tracked by DEV-2679.

handsontable/src/core.ts
demtario: Fixed in 20ac8d1. The duplicate row/column counters were removed from Core. Selection restoration now keys directly off a pure trimming state (trimmedIndexesChanged && !indexesSequenceChanged), which excludes inserts, removals, moves, and sorts without another seeded count or the equal-size structural blind spot. EditorManager keeps its existing counters only for editor-record repair.

qunabu:
Task linked: [DEV-2679 Typing into a selection left past the last row by a trim grows the source data](https://app.clickup.com/t/9015210959/DEV-2679)

handsontable/src/selection/selection.ts
```diff
@@ -1678,6 +1835,179 @@ class Selection {
...
+    const highlight = this.#createVisualCoords(
+      physicalRange.highlight, visualRange.highlight, axis, undefined, useVisualFallback);
+    const from = this.#createVisualCoords(
+      physicalRange.from, visualRange.from, axis, fromBoundary, useVisualFallback);
+    const to = this.#createVisualCoords(
+      physicalRange.to, visualRange.to, axis, toBoundary, useVisualFallback);
+
+    if (!highlight || !from || !to) {
+      return null;
+    }
```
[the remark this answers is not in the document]
demtario: Confirmed and fixed in fe8e3ac411 — this was real, and worse than the description: the inherited layer is filled across its full width.

**Reproduced first.** Select physical rows 1-3 in column A, type into row 3, trim row 1, press Ctrl+Enter. With a second layer on physical row 4 spanning both columns, the commit wrote `EDITED` into **`A4` and `B4`** while the edited records `A2`/`A3` stayed untouched. `BaseEditor#saveValue()` reads `getSelectedRangeActive()` under `ctrlDown`, so dropping the active layer hands the commit to whichever layer inherits the active slot.

I also had a test (`keeps surviving layers when the active range loses a corner`) that pinned the drop as correct. It asserted `sourceData()` was untouched, which held only because it never committed — so it documented the orphaned state without exercising it. That test was wrong and has been replaced.

**Fix.** A range whose focus survives now shrinks onto its surviving records instead of being dropped; a range is dropped only when its focus itself was trimmed. This is exact rather than approximate: trimming only removes records, so the survivors of a contiguous range stay contiguous in the new visual space and a `CellRange` still describes them precisely. `capturePhysicalSelection()` records the physical indexes the range covers along the axis being updated, and `#restorePhysicalRange()` resolves a trimmed corner to the nearest surviving record scanning inward from that end.

Two paths are deliberately unchanged. Whole-row and whole-column layers skip the span capture entirely — they re-pin to the axis boundaries on restore, so walking their span would cost a pass over the grid for a result that is never read. And when the *active* layer's own focus is trimmed, all three corners still fall back to the clamped visual slot, which is what leaves the editor somewhere `EditorManager` can cancel it.

**Verification.** The new case `commits to the surviving selected records when the active range loses a corner` asserts the whole source data set after a real `Ctrl+Enter`. I reverted just the `selection.ts` change and rebuilt to confirm it fails without the fix (`A4`/`B4` = `EDITED`) and passes with it. Full runs after the fix: 192/192 Playwright cases in `editor-trimmed-row.spec.ts`, and 9865 Jasmine specs with 0 failures.

budnix (dismissed):

budnix (dismissed):

budnix (approved):

PR #13305, opened 2026-08-28
DEV-2693: Document which options work at which configuration level

### LINKED ISSUE #8431: Documentation: what options apply to what level (table, column, cell)
### Description
The documentation describes [cascading options](https://docs.handsontable.com/pro/2.0.0/Options.html) (a great feature!): 

1. Table options through the `constructor` and `updateSettings`
2. Column options provided through `options.columns`
3. Cell options provided through `cellProperties`

However, for any given option, it doesn't say which contexts it works in. Some of them you can work out intuitively (like `columns` is probably not a valid option *for a column* -- but maybe it could be??). It would be helpful if each option simply listed the contexts (table, column, cell) that the option can be applied to. Even better if the documentation could filter down to a single context.

### Reference
https://docs.handsontable.com/pro/2.0.0/Options.html

### LINKED ISSUE #8476: Specify meta data in documentation
@alvarotrigo commented on [Thu Apr 16 2015](https://github.com/handsontable/handsontable/issues/2377)

I can see the methods `getCellMeta` and `setCellMeta`, but there's nowhere in the documentation where the `meta` array is specified or explained.

Not even the possible values for the `key` parameter are specified in the `setCellMeta` docs. Is it some kind of mysterious secret?

[I've just discovered a way (not sure if its the proper one) to set a cell to invalid state](http://stackoverflow.com/questions/24595389/handsontable-how-to-set-cell-to-invalid-manually) but I found no reference to it in the docs and I couldn't guess it from the returned object from `getCellMeta`.

---

@AMBudnik commented on [Mon Apr 20 2015](https://github.com/handsontable/handsontable/issues/2377#issuecomment-94387177)

Thank you @alvarotrigo for your support. We'll see what we can do.

budnix:
### Context

The PR documents which configuration options take effect at which level, closing two issues open since 2018 that ask the same thing.

An option can be *written* at any level — the cascade is a prototype chain, so putting `readOnly` inside `columns` never throws. Whether it *takes effect* is a different question, decided by where the grid reads it. Nothing in the docs stated that per option.

`metaSchema.ts` already carried a scope sentence for 112 of 169 options. The missing 57 were exactly the ones people ask about: `readOnly`, `type`, `renderer`, `editor`, `className`, `width`, `source`, `locale`, `placeholder`. The guide even promised information it could not deliver:

> Unless an option's reference entry states otherwise, you can set it at any cascading level.

#### Approach

`metaSchema.ts` becomes the single source of truth. A new `@configScope` JSDoc tag on all 169 options lists the levels each takes effect at. Each value was resolved from the read site, which is mechanical rather than a judgement call:

| The code reads | Meaning | Levels | Count |
| --- | --- | --- | --- |
| `getSettings()` / `tableMeta` | `TableMeta` is a sibling of the column branch, so column and cell values never reach it | grid | 105 |
| `getColumnMeta()` | Column meta inherits from global meta | grid + columns | 2 |
| `cellProperties` / `getCellMeta()` | Cell meta inherits from column, which inherits from global | every level | 58 |
| the raw `columns` setting | never goes through the meta chain | columns | 1 |
| a plugin that merges a per-column entry | `getMergedPluginSettings()` reads `columns[n][pluginKey]` | grid + columns | 2 |

That is 168. All 169 options carry a tag, but `preventWheel` is `@private`, so jsdoc drops it from the API reference and the matrix skips it too — otherwise its row would link to an anchor that does not exist.

Every plugin reads its own toggle through `getSettings()[PLUGIN_KEY]`, which is why no plugin can be switched on per column or per cell. That single rule accounts for most of the grid-only set. A plugin can still let you override its *sub-options* per column: `columnSorting` and `multiColumnSorting` merge `columns[n][pluginKey]`, so both are `grid columns` with a note saying the toggle itself stays grid level.

#### What ships

- **`@configScope` tag** on all 169 options, registered for ESLint in `.eslintrc.js`.
- **The convention is recorded** in `handsontable/src/dataMap/metaManager/AGENTS.md`: how to pick the value from the read site, the two traps, and the regenerate step.
- **Generator** `handsontable/scripts/generate-option-levels.mjs` (+ the pure lib `scripts/utils/option-levels.js`), wired as `npm run generate:option-levels`. It writes `option-levels.json` and injects the matrix into the guide page between marker comments, following the existing `measure-module-sizes.mjs` pattern. It reads `src/` directly, so it needs no core build.
- **New guide page**, "Option levels", with the 168-option matrix, a search box and per-level filters, both mirrored in the URL so a filtered view can be linked.
- **Level badges** under every option in the Options API reference, via an `applyConfigScope` pre-processor.
- **New "Configuration" sidebar category** holding the two configuration pages, placed after Getting started. The two pages had overlapping content; the split is now clean — "Setting options" covers *how* to set an option at each level, "Option levels" covers *where* each option takes effect. Four duplicated sections were removed from the new page in favour of links.

#### Renames

The sidebar read `Configuration › Configuration options › Configuration option levels`, repeating the same words three times. The pages are now titled **Setting options** and **Option levels**, with `metaTitle` aligned to match.

Worth knowing: `metaTitle` renders nowhere on this site. Only `content.config.ts` declares it; `<title>`, `og:title` and the `h1` all come from `title`. So the rename does change the browser tab and the search-result headline. What protects the existing SEO is the permalink, and that is unchanged — `/configuration-options` still resolves, and `canonicalUrl` still points at it.

#### Fixed along the way

- **An orphan entry in the Options reference "On this page" list.** Option names render as `### <name>`, and three option *descriptions* used `###` for their own subsections, making them siblings of the options rather than children. `How 'auto' differs from leaving height unset` therefore appeared as if it were an option. All three are now `####`, which nests them under their option and keeps them out of the table of contents (Starlight covers h2–h3). Pre-existing — the `height` one dates to #12011 — and the two in `className` had not surfaced yet only because `docs:api` reads `handsontable/tmp`.
- **A dead link in the matrix.** `preventWheel` is `@private`; the matrix listed it and linked to a non-existent anchor. `parseOptions` now skips `@private` blocks.
- **Deep links into the matrix page were broken.** The URL sync rebuilt the address from the path and search params only, dropping the fragment, and it runs on mount — so landing on a section anchor stripped the hash and cancelled the browser's jump to it.
- **The URL sync is debounced** (200ms). Rows still filter on every keystroke, but Safari ignores `replaceState` after roughly 100 calls per 30 seconds, which typing an option name crosses on its own. The pending write is cancelled on every call, so reverting a keystroke inside the window cannot leave a stale query in the URL.

#### Notes for the reviewer

- **Page URLs are unchanged.** Routes come from each page's `permalink` frontmatter, not the file path, so moving the two folders is URL-safe.
- **The badges render below the heading, never inside it.** An earlier attempt put them in the `<h3>`; markdown derives a heading's anchor from its text, so every option anchor changed (`#readonly` became `#readonly-grid-columns-cells-cell`), breaking 656 in-page links and every inbound link. `applyConfigScope.test.mjs` asserts on `hot-header.hbs` that the option heading renders no config-scope markup. It deliberately does not read the generated reference: `content/api/` is gitignored and the `plugins` CI job runs no `docs:api` step, so a check that read the generated file would pass by doing nothing.
- **`scripts/utils/option-levels.js` is CommonJS on purpose**, matching `browser-targets.js`. As `.mjs` it would need a `transform` override in `jest.config.js`, and any such override — even one restating Jest's default pattern verbatim — breaks `updateSettingsExternalDataSource.unit.js`.
- **Two options are declared as method shorthand** (`isEmptyRow`, `isEmptyCol`). The parser accepts both declaration forms; matching only `name:` silently merges a method's doc block into the next option's and misattributes its tag. The unit test pins the parsed count to the number of `@memberof Options#` markers, minus `@private` ones, so this cannot regress.
- Only `title` is not settable at the grid level, because it is read straight off the raw `columns` setting in `getColHeader`.

#### Known gaps, deliberately left for follow-up

- The eight caveats in the generator's `NOTES` map reach the matrix page's Notes column but **not** the API reference badge, which renders levels only. So `search` badges as all four levels without the "the toggle is grid level" caveat. Closing this means moving the caveats into `metaSchema.ts` as a second tag the badge can render.
- `LEVELS` is declared in three places: the core lib, the docs page script, and the four hand-written filter checkboxes. Adding a fifth level would need all three updated, and no test ties them together.
- `test-handsontable-unit` does not run on `docs/**`, so a hand-edit to the generated block between the markers would not be caught by the sync test until an unrelated core PR goes red.

### How has this been tested?

Core unit suite, including 10 assertions that pin the tags, the generator, and the committed page together:

```bash
# 330 suites, 4077 tests passed

npm run test:unit --prefix handsontable -- --testPathPattern=optionLevels
# 10 passed
```

Docs pipeline tests, including the heading-anchor guard and the duplicate-tag guard:

```bash
npm run docs:test:plugins --prefix docs
# 250 passed
```

Lint and a full docs build:

```bash
npm run build --prefix docs
```

Manual verification in Chromium against the built site:

- the Options reference "On this page" list runs `height` → `hiddenColumns` with no orphan entry, and the three demoted subsections still render as `h4` under their option;
- the matrix renders 168 rows as a real table, with no `preventWheel` row;
- loading `#options-matrix` keeps the fragment and scrolls to it;
- typing a character and deleting it inside the debounce window leaves the URL clean;
- typing `wordwrap` settles to `?q=wordwrap#options-matrix`, keeping query and fragment together;
- loading `?q=title&level=columns` restores the search box, the checkbox and a single row;
- no console errors.

### Types of changes

### Related issue(s):

1. DEV-2693
2. Closes #8431
3. Closes #8476

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2693

---

> [!NOTE]
> **Low Risk**
> Changes are documentation, generated content, and JSDoc metadata; grid runtime behavior is unchanged aside from updated doc links in source comments.
> 
> **Overview**
> Adds **`@configScope`** on every public option in **`metaSchema.ts`** as the source of truth for where each setting actually takes effect (grid, `columns`, `cells`, `cell`), distinct from where it can be written in the cascade.
> 
> A new **`npm run generate:option-levels`** pipeline parses those tags and refreshes **`option-levels.json`** plus the injected matrix on the **Option levels** guide, with client-side search, level filters, and shareable URL query params. The Options API reference now shows **“Can be set at”** badges via an **`applyConfigScope`** JSDoc pre-processor (badges sit below headings so option anchors stay stable).
> 
> Docs IA shifts configuration into a **Configuration** sidebar: the former getting-started guide is renamed **Setting options**, links site-wide point at the new paths, and the cascading guide text no longer implies every option works at every level.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d58dda990fc88bf1e8b7e53f4869e94100a7a96c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2693 Document which configuration options work at which level](https://app.clickup.com/t/9015210959/DEV-2693)

handsontable/src/dataMap/metaManager/metaSchema.ts
```diff
@@ -1141,6 +1163,7 @@ export default (): Record<string, unknown> => {
      * @type {boolean|object}
      * @default undefined
      * @category ColumnSorting
+     * @configScope grid
```
demtario: **High:** `columnSorting` is tagged `@configScope grid`, but the plugin also reads per-column configuration, so the matrix and the new API badge contradict a shipped guide.

`getMergedPluginSettings()` reads the raw `columns` array off the column-meta prototype and merges the per-column entry (`src/plugins/columnSorting/columnSorting.ts:620-636`, via `getPluginColumnConfig` at `:601-611`). `docs/content/guides/rows/rows-sorting/rows-sorting.md:305` documents it ("You can also override `columnSorting` options per column, using the `columns` configuration"), and `columnSorting.spec.js:2886-2909` asserts it.

Suggested fix: `@configScope grid columns`, plus a `NOTES` entry. The note is the substance here rather than an extra. The plugin toggle itself is grid level; only `indicator`, `headerAction`, `sortEmptyCells`, and `compareFunctionFactory` are overridable per column. A bare `grid columns` puts this option in the same bucket as `headerClassName` and `data`, which reads as "sorting can be enabled per column" and is not true. The note is also the place to say that this resolves through the raw `columns` setting, like `title`, not through `getColumnMeta()` like the other two `grid columns` options.

handsontable/src/dataMap/metaManager/metaSchema.ts
```diff
@@ -4663,6 +4761,7 @@ export default (): Record<string, unknown> => {
      * @type {boolean|object}
      * @default undefined
      * @category MultiColumnSorting
+     * @configScope grid
```
demtario: **High:** Same defect as `columnSorting`.

`MultiColumnSorting extends ColumnSorting` (`src/plugins/multiColumnSorting/multiColumnSorting.ts:71`) with its own `pluginKey`, so it inherits `getMergedPluginSettings` and resolves `columns[n].multiColumnSorting` the same way. Asserted by `multiColumnSorting.spec.js:2984` and `:3026`.

Same fix: `@configScope grid columns` plus the matching `NOTES` entry.

docs/scripts/jsdoc-convert/renderer/preProcessors/__tests__/applyConfigScope.test.mjs
```diff
@@ -0,0 +1,94 @@
...
+  // Regression guard. The badges must render BELOW the heading, never inside it: a
+  // markdown heading's anchor is derived from its text, so markup in the heading rewrote
+  // every option anchor (`#readonly` -> `#readonly-grid-columns-cells-cell`) and broke
+  // 656 in-page links plus every inbound link to the API reference.
+  const optionsMd = resolve(currentDir, '../../../../../content/api/options.md');
+  let source;
+
+  try {
+    source = readFileSync(optionsMd, 'utf8');
+  } catch {
```
demtario: **Low:** This regression guard never executes, and the assertion below it cannot simply be tightened.

`docs/.gitignore:25` ignores `/content/api/*`, and the `plugins` job in `.github/workflows/docs.yml` runs `docs:test:plugins` with no `docs:api` step, so `readFileSync` throws and this `catch` returns on every CI run. The PR description's claim that the test "now fails if any `###` heading contains markup" does not hold today.

The second half: once the reference is generated, `assert.doesNotMatch(heading, /<|>/)` conflicts with `hot-header.hbs:4`, which puts `<span class="tag-deprecated">Deprecated</span>` inside the `### {{{name}}}` heading. `metaSchema.ts` has no `@deprecated` option today, so the test would start failing the day one is added, on markup this PR did not introduce.

Suggested fix: generate the reference before the assertion (or move the check into the `docs:api` pipeline), and scope the regex to config-scope markup only.

handsontable/src/dataMap/metaManager/metaSchema.ts
```diff
@@ -688,26 +702,27 @@ export default (): Record<string, unknown> => {
...
-     * The `cell` option overwrites the [top-level grid options](@/guides/getting-started/configuration-options/configuration-options.md#set-grid-options),
+     * The `cell` option overwrites the [top-level grid options](@/guides/configuration/configuration-options/configuration-options.md#set-grid-options),
      * and the [`columns`](#columns) options.
      *
      * Each entry's `row` and `col` are **visual** indexes. This differs from the [`cells`](#cells)
      * option, whose `row` and `column` are physical indexes.
      *
      * Read more:
-     * - [Configuration options: Setting cell options](@/guides/getting-started/configuration-options/configuration-options.md#set-cell-options)
+     * - [Configuration options: Setting cell options](@/guides/configuration/configuration-options/configuration-options.md#set-cell-options)
```
demtario: **Low:** The label rename sweep is incomplete. 18 sites still read "Configuration options".

Bare `[Configuration options](...)` labels were renamed to `[Setting options]`, but every `[Configuration options: <section>]` label was left in place, on lines this PR already touched: `metaSchema.ts` 714, 763, 764, 903, 1077, 1486, 2662, 5451, 5614, 5687, 5770, 6310, 6345, 7170, 7325, and `docs/content/recipes/rendering-styling/conditional-row-coloring/conditional-row-coloring.md:304`.

Either rename them all to "Setting options: ...", or leave all of them on the old label.

demtario (commented):
4 inline notes below.

### Off-diff

**High: the new `@configScope` convention is not recorded in any `AGENTS.md`.**

`handsontable/scripts/utils/option-levels.js:72-74` throws when an option carries no `@configScope` tag, and `handsontable/test/__tests__/optionLevels.unit.js` turns that into a red `test:unit`. So every future option must carry the tag, and any tag change must be followed by `npm run generate:option-levels --prefix handsontable`. Nothing records that: `git diff --stat -- '*AGENTS.md'` is empty on this branch, and `grep -rn configScope` across the `AGENTS.md` files exits 1. The root checklist requires the update; the lowest correct scope looks like `handsontable/src/dataMap/metaManager/AGENTS.md`.

handsontable/src/dataMap/metaManager/metaSchema.ts
budnix: Good spot — retagged `grid columns` in d58dda990f, with a note that only the sub-options are per column and that it resolves through the raw `columns` setting. Also recorded the whole `@configScope` convention in the metaManager AGENTS.md, which I had missed.

handsontable/src/dataMap/metaManager/metaSchema.ts
budnix: You are right, it inherits the same merge — fixed in d58dda990f alongside `columnSorting`, with the matching note.

docs/scripts/jsdoc-convert/renderer/preProcessors/__tests__/applyConfigScope.test.mjs
budnix: Yep, my mistake — it was skipping on every CI run. In d58dda990f the guard asserts on `hot-header.hbs` instead, which always exists, and only looks for config-scope markup so a future `@deprecated` option will not trip it. I checked it by putting the bug back and watching it fail.

handsontable/src/dataMap/metaManager/metaSchema.ts
budnix: Fair point — finished the sweep in d58dda990f, all 18 now read "Setting options: ...".

demtario (approved):

PR #12804, opened 2026-06-19
DEV-1926: Fix context menu inserting two rows on iPad Safari

GabberPL:
### Context

On iPad Safari, selecting **Insert row above** or **Insert row below** from the context menu inserts **two rows** instead of one. Mac, iPhone, and Android are not affected.

**Root cause:** iPadOS Safari reports a desktop user-agent, so `isMobileBrowser()` returns `false`. However `isTouchSupported()` returns `true`. Walkontable's `registerEvents()` therefore calls both `initTouchEvents()` and `initMouseEvents()`, registering listeners for both event families on the same element.

When the user taps a menu item:
1. `touchend` fires → `onTouchEnd()` directly calls `this.onMouseUp(event)` → `onCellMouseUp` → `beforeOnCellMouseUp` → `executeCommand()` **(first insert)**
2. The browser then synthesizes a `mouseup` event → the `mouseup` listener registered in `initMouseEvents()` calls `onMouseUp(event)` again → `executeCommand()` **(second insert)**

**Fix:** Stamp `#lastTouchMouseUpAt = Date.now()` immediately before `onTouchEnd` calls `onMouseUp`. The `mouseup` listener in `initMouseEvents` skips calling `onMouseUp` if fewer than 500 ms have elapsed since that timestamp, suppressing the duplicate synthesized event.

### Related

- GitHub issue: https://github.com/handsontable/handsontable/issues/12803
- ClickUp: https://app.clickup.com/t/9015210959/DEV-1926

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature or improvement
- [ ] Breaking change
- [ ] Translation

### Affected project(s)

- [x] `handsontable`
- [ ] `@handsontable/angular-wrapper`
- [ ] `@handsontable/react-wrapper`
- [ ] `@handsontable/vue3`

### Checklist

- [x] The fix is targeted at the correct layer (Walkontable event deduplication)
- [ ] E2E test added (requires BrowserStack / real iPad — cannot run in headless CI)

---
_Generated by [Claude Code](https://claude.ai/code/session_0184DoCLi8qW26muYz5g1cDa)_

---

> [!NOTE]
> **Medium Risk**
> Touches core pointer event routing in Walkontable; the 500 ms window could theoretically affect rapid touch-then-real-mouse sequences on hybrid devices, though scope is limited to duplicate suppression after touch `mouseup`.
> 
> **Overview**
[...]

Issue #12803, opened 2026-06-19
Context menu "Insert row above/below" inserts two rows instead of one on iPad (Safari)


GabberPL:
### Description

When using the context menu to **Insert row above** or **Insert row below** on an iPad running Safari, **two rows are inserted** instead of one.

The bug does **not** reproduce on:
- Mac (Safari or Chrome)
- iPhone (Safari)
- Android (Chrome)

It is **iPad-specific** and appears to be a touch-event handling issue — likely a `touchend` + synthesized `click` double-firing the context menu item action.

### Steps to reproduce

1. Open Handsontable on an iPad using Safari.
2. Long-press or tap a cell to open the context menu.
3. Tap **Insert row above** (or **Insert row below**).

### Expected behavior

One row is inserted.

### Actual behavior

Two rows are inserted.

### Environment

- Device: iPad (all models tested)
- Browser: Safari (iPad OS)
- Other browsers/platforms: not affected

### Additional context

Investigated in relation to DEV-1703 / PR #12587 (scroll jump fix) — that change only modifies `ignoreScrollSources` in `core.js` and cannot cause this. This is a pre-existing or unrelated bug triggered by Safari on iPadOS firing both `touchend` and a synthetic `click` event on context menu items, causing the action to be executed twice.

Issue #11659, opened 2025-06-05
On the mobile side, each scroll will re-trigger the cell selection. Under normal circumstances, it should be distinguished whether the user wants to move or selecte. If it is move, the selection cannot be re-triggered.


JXUnx1:
### Describe the bug

On the mobile side, each scroll will re-trigger the cell selection. Under normal circumstances, it should be distinguished whether the user wants to move or selecte. If it is move, the selection cannot be re-triggered.

### Handsontable version

last version

### Your environment

chrome moblie

adrianspdev:
Hi, thanks for reporting this issue and for the clear reproduction!

This has already been resolved in https://github.com/handsontable/handsontable/pull/12450.

PR #13306, opened 2026-08-28
DEV-2687: Make double-tap-to-edit deterministic on devices with touch and mouse listeners

adrianspdev:
### Context

The PR fixes double-tap-to-edit on devices that register both touch and mouse listeners — iPad Safari and Chrome (desktop UA since iPadOS 13), Windows touchscreens. Since 18.1.0-rc1 the editor opened intermittently ("sometimes after 3 taps, sometimes never"); 18.0.0 opened it every time.

**Regressing change:** `9ca060fe6` — DEV-1926 "Fix context menu inserting two rows on iPad Safari" (#12804, fixing #12803), as identified in the ticket. It stamps `#lastTouchMouseUpAt` on a touch-driven `onMouseUp` and drops the browser-synthesized `mouseup` that follows a tap — but only the `mouseup`.

**Verified mechanism** (differs from the ticket's theory):
- **18.0.0**: the unsuppressed synthesized `mousedown`+`mouseup` pair after a tap completed a second double-click match, so a tap on an already-selected cell opened the editor instantly — the same duplication that caused #12803. Reliable editing was a side effect.
- **18.1 rc**: the synthesized `mouseup` is dropped, so the editor opened only when two `touchend`s landed inside the 500 ms mouseup pairing slot. iOS swallows taps under ~250 ms, so the working window was ~250–500 ms — a coin flip at a human 300–600 ms cadence.

**Fix (as shipped, after review rounds):**
- `helpers/dom/inputOrigin.ts` (deliberately NOT spread onto `Handsontable.dom`): `getMouseEventTouchOrigin(event)` — Blink's `sourceCapabilities.firesTouchEvents`, `undefined` when the engine is silent — plus the shared `TOUCH_SYNTHESIZED_MOUSE_WINDOW` (500 ms) fallback window used by both Walkontable and TableView (same window, deliberately different policies).
- Walkontable `event.ts`: the mouse listeners drop the browser-synthesized pair **symmetrically and consume-once** — an engine report of "not touch" always wins (a real mouse or pen is never dropped); otherwise only the first `mousedown`/`mouseup` pair after a touch-driven `onMouseUp` is dropped, inside the 500 ms ceiling. The pending pair is cleared when a scroll-classified gesture ends; `touchcancel` resets the per-gesture state so a cancelled gesture cannot leak real mouse events into the touch paths.
- Touch double-taps are paired by an independent detector (`#handleTouchTap`): two taps on the same cell **coordinates** (not TD identity — Walkontable recycles elements) within `TOUCH_DBLTAP_TIMEOUT` (1000 ms) fire the double-click. Touch never arms the mouse `#dblClickOrigin` slots, so a mouse click after a tap cannot pair with it; mouse double-click timing (1000/500 ms) is unchanged. Context-menu commands still run once (#12803).
- `tableView.ts`: `#isSyntheticMouseEvent` reuses the shared helper (`getMouseEventTouchOrigin(event) ?? #recentTouchEnd`) — same fallback window, deliberately laxer policy (a Blink-flagged pair must never run the outside-click handling that closes editors).
- Walkontable `AGENTS.md` and `tests/AGENTS.md`: the dual-listener rules, the residuals, and how to emulate the setup in Playwright.

Behavior note vs 18.0.0: the touch double-tap window is 1 s on all touch devices (was 500 ms), and a single tap on an already-selected cell no longer opens the editor (that was the duplicated-events side effect; recorded in the changelog entry).

- Unit tests added/modified (`*.unit.js`): `handsontable/src/helpers/dom/__tests__/inputOrigin.unit.ts` (3 cases for `getMouseEventTouchOrigin`)
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`): `tests/e2e/touch-tap-to-edit.spec.ts` — a desktop-UA + `hasTouch` describe (the iPad listener setup; Chromium synthesizes the same compat sequence) and a mobile-UA describe (iPhone emulation; touch-only listeners), all timers driven by `page.clock`; fixture `tests/fixtures/demo/touch-tap-to-edit.html`, page object `tests/fixtures/pages/TouchTapToEditPage.ts`, `tests/fixtures/pages/windowTypes.ts`.
- For a bug fix — the spec that fails without this fix: `touch-tap-to-edit.spec.ts › a double-tap at a 700 ms cadence opens the editor` and the origin-fallback spec (both RED on the pre-fix engine, verified by rebuilding `dist/` from `develop`'s `event.ts`); every review-round change carries its own red→green proof (consume-once, per-gesture reset, gate placement, touchcancel).
- Demo page / recorded trace (for UI changes): local `handsontable/dev-pr.html?touch=1` (gitignored) — scripted iOS-like tap sequences: gap 600/900 ms → editor opens, 1200 ms → does not

**Real-device verification:** tested on a real iPad Pro 11" (Safari) by @adrianspdev — tap once to select, then double-tap at a normal cadence opens the editor every time; the context-menu command runs once. This is the only environment that exercises the WebKit timing fallback on trusted synthesized events (Chromium reports `sourceCapabilities`, so the Playwright spec covers the fallback with script-dispatched events instead).

### Commands run

```
npm --prefix handsontable run eslint          # 0 errors
npm --prefix handsontable run typecheck       # ✓ tsc --noEmit
npm --prefix handsontable run test:unit       # 330 suites, 4070+ tests passed
npm --prefix handsontable run test:walkontable # 816 specs, 0 failures
cd tests && npx playwright test e2e/touch-tap-to-edit.spec.ts --project=e2e-main --project=e2e-main-min   # all tests × 2 legs
cd tests && npx playwright test e2e/mobile-selection-handles.spec.ts e2e/shadow-dom.spec.ts --project=e2e-main   # 16 passed
```

### Types of changes

### Related issue(s):

1. DEV-2687
2. Regressing: DEV-1926 / #12804 (fixing #12803)
3. QA tracking: RELEASE-869

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

- [x] This change needs a manual QA pass — on a real iPad (Safari and Chrome): tap a cell once, then double-tap it at a normal 300–600 ms cadence → the editor opens every time; long-press → context menu → tap "Insert row above" → exactly one row is inserted. Automation cannot emit trusted WebKit-synthesized mouse events, which is the path the timing fallback guards. Already verified by the author on an iPad Pro 11".

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2687

---

> [!NOTE]
> **Medium Risk**
> Changes low-level pointer event routing for touch-capable desktops and tablets; regressions could affect selection, editing, or context menus, though coverage is broad and the fix is scoped to synthesized-mouse handling.
> 
> **Overview**
> Fixes **intermittent cell editing on iPad and other devices** that register both touch and mouse listeners (desktop UA + touch). After #12804 only suppressed the synthesized `mouseup`, leaked `mousedown` events broke double-click pairing and could duplicate `onCellMouseDown` / context-menu actions.
> 
> **Walkontable `event.ts`** now drops the browser’s first post-tap **`mousedown`/`mouseup` pair** via `#isTouchSynthesizedMouseEvent()` (Blink `sourceCapabilities` when available, otherwise a consume-once `#synthesizedPairPending` flag inside a shared **500 ms** window). **Touch double-taps** are detected separately in `#handleTouchTap()` (same **cell coordinates** within **1 s**), without arming the mouse double-click slots; **`touchcancel`** resets gesture state so real mouse input cannot be misrouted.
> 
> Adds internal **`helpers/dom/inputOrigin.ts`** (`getMouseEventTouchOrigin`, `TOUCH_SYNTHESIZED_MOUSE_WINDOW`). **`tableView.ts`** reuses the helper and window for outside-click / editor-close guards (policy stays stricter than Walkontable’s).
> 
> **Behavior vs 18.0.0:** editing requires an intentional **double-tap** (not a single tap on an already-selected cell from duplicate mouse events). Changelog entry **#13306**.
> 
> **Tests:** unit tests for `inputOrigin`, Playwright **`touch-tap-to-edit.spec.ts`** (desktop Chrome + `hasTouch`, `page.clock`), fixture and page object; AGENTS docs for dual-listener rules.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7c3b47d283dfc1893d1203c69787e349b28d2514. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-1926 Context menu "Insert row above/below" inserts two rows on iPad Safari](https://app.clickup.com/t/9015210959/DEV-1926)
Task linked: [RELEASE-869 App - Touch Devices](https://app.clickup.com/t/9015210959/RELEASE-869)
Task linked: [DEV-2687 [18.1.0-rc1+] Double-tap-to-edit unreliable on iPad (desktop-UA WebKit) — #12804's one-sided mouseup suppression desyncs dblclick pairing](https://app.clickup.com/t/9015210959/DEV-2687)

handsontable/src/helpers/dom/event.ts
```diff
@@ -50,6 +50,42 @@ export function isMiddleClick(event: Event): boolean {
...
+ * Browsers that fire touch events synthesize a `mousedown`/`mouseup`/`click` sequence after
+ * `touchend` for compatibility. Only Blink exposes the origin (`sourceCapabilities.firesTouchEvents`);
+ * WebKit and Firefox do not, and script-dispatched events carry `null`. In those cases the function
+ * returns `undefined` so the caller can fall back to a timing heuristic.
+ *
+ * @param {Event} event The mouse event object.
+ * @returns {boolean|undefined} `true` when synthesized from touch, `false` when it comes from a
+ *                              non-touch device, `undefined` when the browser does not tell.
+ */
+export function isTouchSynthesizedMouseEvent(event: Event): boolean | undefined {
```
danielzytohoc: **Low (review note):** this export automatically becomes **public API** — `src/index.ts` spreads every export of `helpers/dom/event.ts` onto `Handsontable.dom` (index.ts:113 → 132), and the new unit test already addresses it as `Handsontable.dom.isTouchSynthesizedMouseEvent`, so it looks intended. Flagging it for a conscious yes: once 18.1 ships, the tri-state `boolean | undefined` return contract is locked in under the breaking-changes policy. The name is collision-free against `develop` and reads well, so if public is the intent there's nothing to do; if it's meant to stay internal, it needs to live outside this module before release.

handsontable/src/3rdparty/walkontable/src/event.ts
```diff
@@ -490,9 +554,13 @@ class Event {
         clearTimeout(this.#dblClickTimeout[1]);
       }
 
+      // A touch tap arms both slots at `touchend`, so the mouseup slot decides how long a
+      // double-tap may take. Give touch the same window as the mousedown slot (DEV-2687).
+      const pairingTimeout = this.touchApplied ? TOUCH_DBLTAP_TIMEOUT : DBLCLICK_MOUSEUP_TIMEOUT;
```
budnix: Curious about a side effect of holding both slots for a full second. A tap arms slot 0 and slot 1 at the same instant, so on a Windows touchscreen a single real mouse click on the same cell any time in the next second now completes the double-click and opens the editor.

Before, slot 1 expired at 500 ms and the old time gate dropped the mouseup below 500 ms, so no single click could ever pair with a tap — now the whole 0–1000 ms range does on Blink, where `firesTouchEvents` is false for a real mouse. The JSDoc on `TOUCH_SYNTHESIZED_MOUSE_WINDOW` covers the WebKit side of this trade-off but not this one, so worth a check on a touchscreen laptop with a mouse plugged in.

handsontable/src/3rdparty/walkontable/src/event.ts
```diff
@@ -197,19 +234,25 @@ class Event {
...
-          if (Date.now() - this.#lastTouchMouseUpAt < 500) {
+          if (this.#isTouchSynthesizedMouseEvent(event)) {
             return;
           }
           this.onMouseUp(event);
         });
       this.#deps.eventManager.addEventListener(this.#deps.wtTable.holder, 'mousedown',
-        (event: MouseEvent) => this.onMouseDown(event));
+        (event: MouseEvent) => {
+          if (this.#isTouchSynthesizedMouseEvent(event)) {
```
budnix, line 273: This drops both halves whenever the browser says the event came from touch, but it never checks that the touch path actually handled the gesture. If a tap drifts more than 10 px, `onTouchEnd` marks it as a scroll and skips both `onMouseDown` and `onMouseUp` — and when the grid has nothing left to scroll, the browser still sends the compat mouse pair.

On Blink that pair is now dropped, so the cell is never selected; on WebKit the 500 ms fallback reads a stale `#lastTouchMouseUpAt` and lets it through, so the cell is selected. Gating on "the touch path fired `onMouseUp` for this gesture" rather than the browser report alone would keep the two engines in step.

handsontable/src/tableView.ts
```diff
@@ -1675,16 +1681,11 @@ class TableView {
...
    * @returns {boolean}
    */
-  #isSyntheticMouseEvent(event: Event & { sourceCapabilities?: { firesTouchEvents: boolean } }) {
-    if (event.sourceCapabilities) {
-      return event.sourceCapabilities.firesTouchEvents === true;
-    }
-
-    return this.#recentTouchEnd;
+  #isSyntheticMouseEvent(event: Event): boolean {
+    return isTouchSynthesizedMouseEvent(event) ?? this.#recentTouchEnd;
```
budnix: Good to see the Blink check shared between the two layers. The timing fallback stayed split, though — Walkontable waits 500 ms after the last touch-driven mouseup, while this one clears `#recentTouchEnd` after 400 ms.

On WebKit and Firefox a synthesized mouseup landing between those two numbers is dropped by Walkontable but still runs the outside-click and `unlisten()` logic here. Pulling the window into one shared constant next to the helper would stop the two layers from disagreeing.

tests/fixtures/pages/TouchTapToEditPage.ts
```diff
@@ -0,0 +1,111 @@
...
+    return this.page.evaluate(counter => window.hookCounts[counter], name);
+  }
+
+  /**
+   * Assert a hook counter has reached exactly the expected value (a duplicate invocation lands
+   * synchronously with the tap, so it would already push the count past the value on the first
+   * read).
+   */
+  async expectHookCount(name: HookCounterName, expected: number): Promise<void> {
+    await expect.poll(() => this.hookCount(name)).toBe(expected);
```
budnix: `expect.poll(...).toBe(n)` returns on the first matching sample, so it cannot fail on a count that goes past `n` a moment later — which is exactly the regression these tests exist to catch. The JSDoc says a duplicate lands synchronously with the tap, but Chromium dispatches the compat mousedown/mouseup after `locator.tap()` resolves, so that is not guaranteed.

A second read after the poll settles would at least give the duplicate a chance to fail the test (each `evaluate` is a real round-trip, so pending browser tasks get to run):

```suggestion
  async expectHookCount(name: HookCounterName, expected: number): Promise<void> {
    await expect.poll(() => this.hookCount(name)).toBe(expected);

    // A duplicate can land a task later, after the first matching sample, so re-read
    // the counter — otherwise an over-count would never fail this assertion.
    expect(await this.hookCount(name)).toBe(expected);
  }
```

.changelogs/13306.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Changed the touch double-tap gesture that opens the cell editor to accept taps up to 1 second apart, so editing by double-tap works reliably on iPad and other devices that report both touch and mouse input.",
```
budnix: The entry covers the new 1-second window, but not the part an iPad user will actually notice: in 18.0.0 a single tap opened the editor on these devices, and from 18.1 it takes a double-tap.

`#12804` only ever shipped in the 18.1 RCs, so `changed` is the right type here — it is the 18.0.0 behavior that goes away, not a released bug being fixed. Worth one more sentence naming that, so nobody reads the silence as "single tap still works".

handsontable/src/3rdparty/walkontable/src/event.ts
```diff
@@ -14,10 +14,44 @@ import { getCellCoordsFromMousePosition } from './utils/pointerToCoords';
...
+ * browser-synthesized compatibility sequence when the engine does not report their origin
+ * (WebKit, Firefox, script-dispatched events). Invariants: it must exceed the longest delay the
+ * browser takes to deliver the synthesized sequence (WebKit lands it within ~350 ms even with
+ * double-tap-zoom detection), and it stays shorter than `TOUCH_DBLTAP_TIMEOUT`, so a mouse pair that
+ * arrives 500–1000 ms after a tap on the same cell is processed and completes the double-click
+ * against the still-armed tap. Trade-offs on this fallback path: a real mouse click within the
+ * window after a tap is dropped, and a pair straddling the boundary yields one unpaired half
+ * (mirror image of the pre-DEV-2687 behavior, where only mouseup was dropped).
+ */
+const TOUCH_SYNTHESIZED_MOUSE_WINDOW = 500;
```
demtario: **High:** `TOUCH_SYNTHESIZED_MOUSE_WINDOW` (500) being shorter than `TOUCH_DBLTAP_TIMEOUT` (1000) leaves a 500-1000 ms band in which the browser-synthesized mouse pair that follows a **single** tap completes `onCellDblClick` and opens the editor.

Before this PR both windows were 500 ms, and their coinciding was the safety property: a mouseup that survived the gate always found `#dblClickOrigin[1]` already null. Now a tap arms slot `[0]` for 1000 ms and slot `[1]` for 1000 ms, so a pair accepted at t=600 ms finds both slots set.

This is the fallback path, and iPad Safari and Firefox never expose `sourceCapabilities`, so they are always on it. The PR states both sides of the contradiction in the same commit: [`AGENTS.md` L191](https://github.com/handsontable/handsontable/blob/cae20342ce5b911d26c99b71a61ffacb313958af/handsontable/src/3rdparty/walkontable/AGENTS.md#L191) says 18.0.0's single-tap-opens-editor "was a side effect, not a feature; do not 'restore' it", while [`touch-tap-to-edit.spec.ts` L99-L101](https://github.com/handsontable/handsontable/blob/cae20342ce5b911d26c99b71a61ffacb313958af/tests/e2e/touch-tap-to-edit.spec.ts#L99-L101) asserts it happens.

Reachability is not theoretical: `#lastTouchMouseUpAt` is stamped at [L647](https://github.com/handsontable/handsontable/blob/cae20342ce5b911d26c99b71a61ffacb313958af/handsontable/src/3rdparty/walkontable/src/event.ts#L647) **before** `this.onMouseUp(event)` on the next line, so the tap's own selection and render work is charged against the 500 ms budget. A 300 ms render leaves ~200 ms for a browser this JSDoc says needs ~350 ms, and that number carries no source. Where does it come from, and was it measured on device?

Fix: arm a consume-once "expect one synthesized pair" flag at `touchend`, consumed by the first mousedown+mouseup pair, and keep 500 ms as a ceiling. The ceiling stays load-bearing because a first tap on an unselected cell is `preventDefault`-ed and produces no pair, so a bare flag would stay armed and swallow the next genuine mouse pair. Do **not** simply set this constant to `TOUCH_DBLTAP_TIMEOUT`: that makes the dropped-mousedown problem below strictly worse. Either fix requires inverting the last spec case, which is itself a sign that test was written to the implementation rather than to the stated invariant.

handsontable/src/3rdparty/walkontable/src/event.ts
demtario, line 273: **Medium:** `mousedown` was never gated before this PR, and on the fallback path this gate also drops a **real** mouse mousedown within 500 ms of a tap. The JSDoc's trade-off note says "a real mouse click within the window after a tap is dropped", which understates the fill-handle case: the whole drag gesture dies, not one click.

On the fallback path, within 500 ms of a tap:

- Fill-handle drag: the `hasClass(realTarget, 'corner')` branch never runs, so [`autofill.ts` L861](https://github.com/handsontable/handsontable/blob/cae20342ce5b911d26c99b71a61ffacb313958af/handsontable/src/plugins/autofill/autofill.ts#L861)'s `mouseDownOnCellCorner` is never set and the drag silently does nothing. Tap to select, then grab the handle with a Magic Keyboard trackpad, is a natural iPad interaction.
- Drag-selection: `#mouseDown` is never set, so `onMouseMove` will not extend the selection.
- Right-click: the `contextmenu` listener is ungated so the menu opens, but the dropped mousedown means the selection is not moved first, so the command acts on the previously selected cell.

Blink is unaffected: a physical mouse, a Windows pen, and hybrid-laptop trackpads all report `firesTouchEvents: false`.

Fix: the consume-once scheme from the other note bounds this to the ceiling. At minimum, name the fill-handle case in the trade-off note at L46-L56 so the cost is recorded.

tests/fixtures/pages/TouchTapToEditPage.ts
```diff
@@ -0,0 +1,111 @@
...
+  async hookCount(name: HookCounterName): Promise<number> {
+    return this.page.evaluate(counter => window.hookCounts[counter], name);
+  }
+
+  /**
+   * Assert a hook counter has reached exactly the expected value (a duplicate invocation lands
+   * synchronously with the tap, so it would already push the count past the value on the first
+   * read).
+   */
+  async expectHookCount(name: HookCounterName, expected: number): Promise<void> {
```
demtario: **Medium:** `expect.poll(...).toBe(expected)` resolves on the first read that matches and never observes a later increment, so this is not an "exactly N" assertion. That matters because [test 3](https://github.com/handsontable/handsontable/blob/cae20342ce5b911d26c99b71a61ffacb313958af/tests/e2e/touch-tap-to-edit.spec.ts#L55-L66) is the only test that proves the symmetric mousedown drop, which is this PR's headline change: if the synthesized mousedown were still processed the counter would reach 3, and the poll can go green on the transient 2.

The JSDoc here justifies the pattern with "a duplicate invocation lands synchronously with the tap", but the spec's own model and the code comment at [`event.ts` L239](https://github.com/handsontable/handsontable/blob/cae20342ce5b911d26c99b71a61ffacb313958af/handsontable/src/3rdparty/walkontable/src/event.ts#L239) both say the browser synthesizes the pair ~0-50 ms later, i.e. after the tap call resolves.

`expectEditorClosed()` at L85 has the same shape and is hollow by construction: the editor starts closed, so the poll returns `false` on the first read and can never catch an editor that opens 50 ms later. It is the assertion in tests 1 and 3 that is supposed to prove no spurious edit.

Related: test 1 asserts `count === 1`, which by its own title is a tautology - if a `preventDefault`-ed first tap produces no synthesized pair, the count is 1 whether or not the drop logic works.

Fix: add a `click` counter to the fixture (`click` is the last event of the synthesized compat sequence), poll **that** to its expected value as the settle barrier, then read the mousedown/mouseup counters with a plain `expect`. A `clickCount` of 0 vs 1 also distinguishes "no pair synthesized" from "pair dropped", which is exactly what test 1's title claims.

handsontable/src/helpers/dom/event.ts
demtario: **Medium:** an `is*` predicate returning `boolean | undefined` breaks the convention this file establishes, and it cannot be narrowed later.

In this same file `isImmediatePropagationStopped`, `isRightClick`, `isLeftClick`, `isMiddleClick`, `isTouchEvent`, and `hasTouchList` all return a strict `boolean`, while the nullable results are named `get*` (`getFirstChangedTouch`, `getTouchPointById`). Every export here lands on the public `Handsontable.dom` surface via [`index.ts` L25](https://github.com/handsontable/handsontable/blob/cae20342ce5b911d26c99b71a61ffacb313958af/handsontable/src/index.ts#L25) and [`base.ts` L104](https://github.com/handsontable/handsontable/blob/cae20342ce5b911d26c99b71a61ffacb313958af/handsontable/src/base.ts#L104), so the tri-state is permanent public API: a consumer writing `if (Handsontable.dom.isTouchSynthesizedMouseEvent(e))` silently reads "the browser did not tell us" as "not synthesized", which is the opposite of what both internal call sites do. Narrowing the signature afterwards is a breaking change.

Fix: rename to a `get*`-shaped accessor, e.g. `getMouseEventTouchOrigin(event): boolean | undefined`. Both call sites stay behaviorally unchanged.

.changelogs/13306.json
demtario: **Medium:** the entry records the gain and omits the loss. Per [`AGENTS.md` L191](https://github.com/handsontable/handsontable/blob/cae20342ce5b911d26c99b71a61ffacb313958af/handsontable/src/3rdparty/walkontable/AGENTS.md#L191) added in this PR, 18.0.0 opens the editor from a tap on these devices; #12804, which removed that path, is only in the 18.1.0 RC line, so users upgrading from stable 18.0.0 are coming from the tap-to-edit behavior and 18.1.0 would ship with no changelog line telling them it is gone.

Fix: extend the existing title rather than adding a second entry. Two precision points for the wording: it is a tap on an **already-selected** cell that opened the editor (a first tap on an unselected cell is `preventDefault`-ed and gets no synthesized pair), and the claim rests on the AGENTS.md note rather than on a verified 18.0.0 run, so phrase it accordingly. If the note is wrong, that line needs fixing instead.

`type: "changed"` and `breaking: false` both look right to me - a gesture-behavior change is not in the breaking table's API/CSS/defaults rows.

demtario (commented):
5 inline notes below.

handsontable/src/helpers/dom/event.ts
adrianspdev: Kept internal. Moved `isTouchSynthesizedMouseEvent` (with its type guard) to `helpers/dom/inputOrigin.ts`, which `src/index.ts` does not spread onto `Handsontable.dom` (same pattern as `helpers/dom/cellCoords.ts`). Verified on the built UMD: `'isTouchSynthesizedMouseEvent' in Handsontable.dom === false`. The shared 500 ms window constant lives there too now.

handsontable/src/3rdparty/walkontable/src/event.ts
adrianspdev: Confirmed, and fixed by taking touch out of the mouse slots entirely: a tap no longer arms `#dblClickOrigin`; touch double-taps are paired by a separate `#handleTouchTap()` (`#lastTapTD`/`#lastTapAt`, 1000 ms). Mouse pairing is back to its pre-PR 1000/500 ms semantics and never sees a tap, so a real click after a tap is a single click on every engine. New spec case: `a tap followed by a real mouse click on the same cell does not open the editor`. The late-pair assertion in the fallback test flipped to "editor closed" for the same reason.

handsontable/src/3rdparty/walkontable/src/event.ts
adrianspdev: Agreed. The gate now keys on the touch path having handled the gesture: dropped only within 500 ms of a touch-driven `onMouseUp` (no stamp for a scroll-classified gesture ⇒ processed), and the browser report is used only as a veto (`firesTouchEvents === false` ⇒ always processed). Both engines now select the cell after a drifted tap. New spec case: `a drifted tap that the touch path treats as a scroll still lets the browser mouse pair select the cell` (script-dispatched touch gesture, so it runs the fallback path; the Blink `true` path shares the same code).

handsontable/src/tableView.ts
adrianspdev: Unified: `TOUCH_SYNTHESIZED_MOUSE_WINDOW = 500` exported from `helpers/dom/inputOrigin.ts`, used by Walkontable's mouse listeners and by TableView's `#recentTouchEnd` timeout (was 400).

tests/fixtures/pages/TouchTapToEditPage.ts
adrianspdev: Applied your re-read first, then went further after demtario's note on the same helper: the fixture now counts the synthesized `click` (last event of the sequence), the tests settle on the expected click count, and the mousedown/mouseup counters are read with a plain `expect` — exact, not first-match. The "synchronous" claim in the JSDoc was mine and wrong; removed.

.changelogs/13306.json
adrianspdev: Reworded (single `changed` entry): a double-tap opens the editor (taps up to 1 s apart), and a single tap on an already-selected cell no longer opens it as it did in 18.0.0, where the duplicated mouse events after the tap triggered the double-click. Verified against the published 18.0.0 bundle: first tap selects, a tap on the selected cell + its synthesized pair opened the editor.

handsontable/src/3rdparty/walkontable/src/event.ts
adrianspdev: The band is gone structurally rather than by tuning: touch taps no longer arm `#dblClickOrigin` at all — touch double-taps are paired by a separate `#handleTouchTap()` (same cell within 1 s), so a mouse pair that survives the gate is a plain mouse click and the last spec case now asserts the editor stays closed. On the stamping order: agreed, `#lastTouchMouseUpAt` is now taken after `onMouseUp()` returns so the tap's own render is not charged against the window. The "~350 ms" figure had no source and is removed; the constant is documented as a ceiling. Consume-once adopted as well (see the mousedown thread).

handsontable/src/3rdparty/walkontable/src/event.ts
adrianspdev: Adopted your consume-once scheme: only the first mousedown+mouseup pair after a touch-driven `onMouseUp` is dropped (`#synthesizedPairPending`), 500 ms stays as the ceiling. Once the pair is consumed, a real fill-handle grab / drag-selection / right-click inside the ceiling is processed. Residual, now named in the JSDoc: a first tap on an unselected cell is `preventDefault`-ed and synthesizes nothing, so a real mouse event inside the ceiling right after such a tap is still dropped on engines without `sourceCapabilities`. Spec: the fallback case dispatches a third script pair inside the ceiling and asserts it is processed (red when consumption is disabled).

tests/fixtures/pages/TouchTapToEditPage.ts
adrianspdev: Switched the spec to your model: the fixture counts `click` (last event of the synthesized sequence) and the tests settle on the expected click count, then read the mousedown/mouseup counters with a plain `expect`. Test 1 now asserts `click === 0` explicitly (nothing synthesized after a `preventDefault`-ed first tap) and is retitled; test 3 settles on `click === 1` before asserting mousedown/mouseup are exactly 2 — that is the assertion that would read 3 without the symmetric drop.

handsontable/src/helpers/dom/event.ts
adrianspdev: Renamed to `getMouseEventTouchOrigin()`; it also moved to `helpers/dom/inputOrigin.ts`, which `index.ts` does not spread onto `Handsontable.dom`, so the tri-state is internal. The Walkontable predicate that consumes it stays `#isTouchSynthesizedMouseEvent(): boolean`.

.changelogs/13306.json
adrianspdev: Title extended (single entry, `changed`, not breaking) and made precise per your note: a single tap on an already-selected cell no longer opens the editor as in 18.0.0, where duplicated mouse events after the tap triggered the double-click. The AGENTS.md line was corrected the same way (already-selected cell, first tap is `preventDefault`-ed). The 18.0.0 claim was reproduced against the published 18.0.0 bundle with the demo page (`dev-latest.html?touch=1`): the tap on the already-selected cell opened the editor every time.

handsontable/src/3rdparty/walkontable/src/event.ts
```diff
@@ -126,13 +166,25 @@ class Event {
...
+   * `touchend`. The whole sequence must be ignored – dropping only one half fires
+   * `onCellMouseDown` a second time per tap and re-arms the double-click pairing slot (DEV-2687).
+   * Used as the fallback for engines that do not expose `sourceCapabilities`
+   * (see `#isTouchSynthesizedMouseEvent`).
    *
    * @type {number}
    */
   #lastTouchMouseUpAt: number = 0;
+  /**
+   * `true` between a touch-driven `onMouseUp` and the browser-synthesized `mouseup` that follows it.
```
demtario, line 183: **Medium:** `#synthesizedPairPending` is never reset when a new gesture starts, so the flag armed by one tap survives into the next gesture and swallows *its* compatibility pair.

`onTouchStart` sets `touchApplied`, `#touchWasMoved`, `#longPressFired` and `#deferredTouchStartEvent`, but not this flag. Trace:

1. Tap cell A. It is an unselected cell, so `onTouchEnd` `preventDefault`s it and the browser synthesizes nothing — but the tap branch still ran, so `#lastTouchMouseUpAt = T1` and `#synthesizedPairPending = true`.
2. At T1+200 ms a gesture on cell B drifts past `LONG_PRESS_MOVE_THRESHOLD`. `onTouchEnd` takes the `else` branch, so no new stamp and the flag is untouched.
3. The browser's compatibility pair for that gesture arrives at ~T1+250 ms: inside the ceiling, flag still armed, so `#isTouchSynthesizedMouseEvent()` drops both halves and cell B is never selected.

This is not WebKit-only. Blink's own compatibility pair carries `sourceCapabilities.firesTouchEvents === true`, so the `=== false` veto does not fire and the pair falls through to the pending+ceiling test. The new spec case passes because it drags from a clean state, with no preceding tap inside the ceiling.

Three places now disagree about the residual. This JSDoc says a fill-handle drag or drag-selection right after a tap "works" on engines that do not report the origin, and `AGENTS.md` says the residual gap is "only the FIRST synthesized pair" and that a drifted gesture's pair "is processed and selects the cell on every engine alike". `helpers/dom/inputOrigin.ts` states it correctly: a real mouse event arriving inside the ceiling **before** a pair that never comes is still dropped, and a `preventDefault`-ed first tap and a long-press are both cases where no pair comes.

Fix: reset `#synthesizedPairPending = false` in `onTouchStart`. A new gesture means the previous gesture's pair can no longer arrive, so the reset is safe, it bounds the residual to a single gesture, and it makes both over-broad claims true. Then reword this JSDoc and the `AGENTS.md` bullet to match, or point them at the `inputOrigin.ts` text.

handsontable/src/3rdparty/walkontable/src/event.ts
```diff
@@ -492,7 +586,50 @@ class Event {
...
+  #handleTouchTap(event: MouseEvent | TouchEvent, cell: ParentCell): void {
+    if (this.#longPressFired || !cell.TD) {
+      this.#lastTapTD = null;
+
+      return;
+    }
+
+    const now = Date.now();
+
+    if (cell.TD === this.#lastTapTD && now - this.#lastTapAt < TOUCH_DBLTAP_TIMEOUT) {
```
demtario: **Medium:** pairing taps by TD element identity goes against an explicit core rule, and a recycled TD lets a single tap open the editor.

`handsontable/AGENTS.md` L84: "when de-duplicating per-tick selection updates, compare **coordinates, not the resolved `td`**: Walkontable reuses the same elements across scrolls, so the element under a stationary pointer is identical on every tick while the cell it represents changes."

Reachable path: tap cell A, then scroll with the mouse wheel or a scrollbar drag — neither goes through `onTouchEnd`, so the `#lastTapTD = null` reset in its `else` branch never runs — then tap within `TOUCH_DBLTAP_TIMEOUT` at a position that resolves to the same recycled TD. The two taps pair, `#fireDblClick()` runs, and the editor opens from what the user experienced as one tap: the regression class this PR removes elsewhere. A re-render from a row insert or remove reshuffles TDs the same way.

Fix: store `cell.coords.row` / `cell.coords.col` instead of the element and compare those. It also drops the retained DOM node that `#lastTapTD` holds indefinitely (no timer clears it, unlike the `#dblClickOrigin` slots).

The mouse `#dblClickOrigin` slots compare TD identity too, but that is pre-existing; this line is new.

handsontable/src/3rdparty/walkontable/src/event.ts
```diff
@@ -576,8 +713,17 @@ class Event {
...
     if (!wasScrolled || this.#longPressFired) {
-      this.#lastTouchMouseUpAt = Date.now();
+      // The stamp is taken AFTER the tap is handled, so the selection/render work onMouseUp does
+      // is not charged against the TOUCH_SYNTHESIZED_MOUSE_WINDOW ceiling. Everything onMouseUp
+      // triggers is synchronous; the browser only dispatches the synthesized mousedown/mouseup
+      // sequence after this touchend handler returns.
       this.onMouseUp(event);
+      this.#lastTouchMouseUpAt = Date.now();
+      this.#synthesizedPairPending = true;
+    } else {
```
demtario, line 735: **Low:** this `else` branch is new logic with no test.

Nothing in `tests/e2e/touch-tap-to-edit.spec.ts` puts a scroll gesture between two taps: `dispatchTouchDrag()` was added in this same push and is used once, in the drifted-gesture case. Three lines cover it with the page object as it stands: `tapCell(1, 1)` -> `dispatchTouchDrag(1, 1, 20)` -> `tapCell(1, 1)` -> `expectEditorClosed()`.

Note the interaction with the TD-identity comment above: switching `#lastTapTD` to coords largely removes this branch's reason to exist, so decide that first rather than adding both.

handsontable/src/3rdparty/walkontable/src/event.ts
```diff
@@ -126,13 +166,25 @@ class Event {
   #deferredTouchStartEvent: TouchEvent | null = null;
   /**
    * Timestamp (ms) of the most recent `onMouseUp` call that originated from a touch gesture.
-   * Used to suppress the browser-synthesized `mouseup` event that fires after `touchend` on
-   * devices that register both touch and mouse listeners (e.g. iPad Safari), which would
-   * otherwise trigger `onCellMouseUp` — and context-menu commands — a second time.
+   * On devices that register both touch and mouse listeners (iPad with a desktop UA, Windows
+   * touchscreens) the browser synthesizes a `mousedown`/`mouseup`/`click` sequence after
+   * `touchend`. The whole sequence must be ignored – dropping only one half fires
+   * `onCellMouseDown` a second time per tap and re-arms the double-click pairing slot (DEV-2687).
```
demtario: **Low:** stale clause. "re-arms the double-click pairing slot" no longer describes the code — `onMouseDown` now arms `#dblClickOrigin` only when `!this.touchApplied`, so a touch tap never touches the mouse slots. The `onCellMouseDown`-fires-twice half of the sentence is still accurate. Drop or reword the second half.

handsontable/src/tableView.ts
```diff
@@ -315,8 +321,10 @@ class TableView {
   #mouseDownLastPos: {row: number, col: number} | null = null;
   /**
    * Flag indicating that a touch interaction just ended. Set to `true` on
-   * `touchend` and reset asynchronously via `_registerTimeout`. Used together with
-   * `sourceCapabilities.firesTouchEvents` (Chrome/Blink) to detect synthetic
+   * `touchend` and reset asynchronously via `_registerTimeout` after
+   * `TOUCH_SYNTHESIZED_MOUSE_WINDOW` — the same fallback window Walkontable's mouse
```
demtario: **Low:** "so both layers agree on which events are synthesized" overstates what sharing the constant buys, and contradicts guidance added in this same push.

Sharing `TOUCH_SYNTHESIZED_MOUSE_WINDOW` makes the two ceilings equal. The policies stay different on purpose: Walkontable is veto -> pending-pair -> ceiling, while `#isSyntheticMouseEvent()` here is a plain `getMouseEventTouchOrigin(event) ?? this.#recentTouchEnd`. The Walkontable `AGENTS.md` bullet in this push says so explicitly — "the two gates are not to be 'harmonized'" — so a future reader gets opposite instructions from the two files.

Fix: "so both layers use the same fallback window", and at the second occurrence (the `#recentTouchEndTimeout` comment) add the one-line reason the policies differ.

demtario (commented):
5 inline notes below.

tests/fixtures/pages/TouchTapToEditPage.ts
demtario: Re-checked at `63c0068`. The `click` counter plus `settleOnClicks()` is a real barrier and it closes this in the cases that use it, so most of the note is addressed. One case is still uncovered.

The `#12803` context-menu test has no barrier at all. `page.clock` is installed and then paused in `beforeEach`, so `runFor(100)` advances virtual time only — it never waits on the browser's input-dispatch pipeline. And the counter is attached to `container`, which the ContextMenu's own Handsontable root sits outside, so `settleOnClicks()` cannot be used there, as the comment says. The sequence is therefore poll (first match) -> fake-clock tick -> exact read, and a duplicate `afterCreateRow` landing a browser task later passes all three. `rowCount() === 6` is a single snapshot read with the same exposure. That leaves the duplicate-command guard, which is the whole point of #12803, without a settle barrier.

Fix that removes the special case instead of adding a real-time wait: attach the click counter to `document` in capture phase rather than to `container`. The menu root is then covered, existing counts do not change (a click inside the grid still counts once), `settleOnClicks()` becomes usable in this test, and the comment explaining why it cannot be used goes away.

Also, with the barrier in place, `expectHookCount()`'s poll-then-reread has no remaining job — the only two call sites already sit behind `settleOnClicks()`. Worth collapsing to `expectHookCountExactly()` and deleting the helper, so there is one assertion shape rather than two with overlapping guarantees.

handsontable/src/3rdparty/walkontable/src/event.ts
adrianspdev: Confirmed, including the Blink reachability. `onTouchStart` now resets `#synthesizedPairPending`, bounding the residual to a single gesture, and the JSDoc/AGENTS.md claims were reworded to match the `inputOrigin.ts` text (which you're right was the only accurate one). New spec case proves it: a tap that synthesizes nothing followed by a drifted gesture whose Blink-flagged pair must be processed — red with the reset removed, green with it.

handsontable/src/3rdparty/walkontable/src/event.ts
adrianspdev: Switched `#lastTapTD` to `#lastTapCoords` (row/col comparison), per the core rule; the recycled-TD single-tap path is gone and the detector no longer retains a DOM node. The stored coordinates are a clone — parentCell()'s border branches hand out references into the live selection CellRange, which normalize() mutates in place. The mouse `#dblClickOrigin` slots stay TD-based as pre-existing behavior outside this PR's scope.

handsontable/src/3rdparty/walkontable/src/event.ts
adrianspdev: Kept the reset (coordinates fix the recycling bug; the reset is still what keeps a scroll from semantically bridging two taps) and added your three-line test: tap → drifted gesture → tap → editor stays closed.

handsontable/src/3rdparty/walkontable/src/event.ts
adrianspdev: Reworded: the leaked pair now "acts as a phantom mouse click"; the fires-twice half stayed.

handsontable/src/tableView.ts
adrianspdev: Reworded to "both layers use the same fallback window", and the `#recentTouchEndTimeout` comment now carries the one-line reason the policies deliberately differ (a Blink-flagged pair must never run the outside-click handling), pointing the reader the same way as the Walkontable AGENTS.md bullet.

handsontable/src/3rdparty/walkontable/src/event.ts
```diff
@@ -511,6 +658,11 @@ class Event {
     this.#longPressFired = false;
     this.#deferredTouchStartEvent = event;
 
+    // A new gesture means the previous gesture's synthesized pair can no longer arrive; without
+    // this reset a pair armed by a `preventDefault`-ed tap (which synthesizes nothing) would
+    // swallow THIS gesture's compatibility pair (DEV-2687 review).
+    this.#synthesizedPairPending = false;
```
demtario: **High:** follow-on from the `#synthesizedPairPending` reset (the thread that asked for it is resolved; this is about where the reset sits, not a reversal).

Clearing the flag on `touchstart` un-gates a *delayed* compatibility pair from the previous gesture, and `touchApplied` stays `true` for the whole next gesture, so that leaked `mouseup` reaches `#handleTouchTap()` and can fire a phantom double-tap.

Chain, all at this head:

- flag cleared here, so `#isTouchSynthesizedMouseEvent()` returns `false` for the late pair (L373-377)
- `onMouseUp()` routes to `#handleTouchTap()` whenever `touchApplied` (L568-572), set at L656 and cleared only at L744
- `#handleTouchTap()` finds `#lastTapCoords` still matching inside `TOUCH_DBLTAP_TIMEOUT` and calls `#fireDblClick()` (L607-627)

So: tap cell A (already selected, nothing `preventDefault`-ed), start the next gesture, A's pair lands mid-gesture, the editor opens before that gesture's own `touchend`. Pre-reset that pair was dropped by `pending && withinCeiling`, so the path is new in this push.

The premise stated at L660-663 ("a new gesture means the previous gesture's synthesized pair can no longer arrive") is contested inside this same file: L258 says the pair arrives ~0-50 ms later, L704-706 cites WebKit's native ~300 ms delay. Rather than settle that, put the reset where it holds under either reading: `onTouchEnd`'s scroll branch, next to the existing `#lastTapCoords = null` (L735-741).

```js
} else {
  this.#lastTapCoords = null;
  // This gesture armed nothing, so a still-pending flag belongs to an earlier gesture whose
  // pair never came; clear it so THIS gesture's compatibility pair is processed.
  this.#synthesizedPairPending = false;
}
```

Under that placement a leaked pair necessarily arrives with `touchApplied === false`, so it cannot reach `#handleTouchTap()` at all, and the worst residual is one extra `onCellMouseDown`, which a single pair cannot escalate into a double-click (L585-596 only sets `#dblClickOrigin[1]`). It still closes what the earlier thread asked for: a tap that armed nothing, followed by a scroll gesture, clears at that gesture's `touchend`, before its own pair arrives. Both new tests pass unchanged.

On testing it: `dispatchTouchDrag()` fires touchstart/touchmove/touchend atomically inside one `evaluate()`, so there is no seam to dispatch a mouse event mid-gesture. A spec for this needs the three touch events split into separate page-object calls.

Docs move with the code: `walkontable/AGENTS.md` L189 ("the flag is reset on every `touchstart`, so the residual never outlives one gesture") and `helpers/dom/inputOrigin.ts` L22-23 ("the pending pair stays armed for the same ceiling after it") both describe the current placement. A pair delayed past the *next* gesture's `touchend` leaks under either placement, so neither doc should claim the residual is bounded to one gesture.

tests/e2e/touch-tap-to-edit.spec.ts
```diff
@@ -0,0 +1,208 @@
...
+
+  test('a pair armed by a tap that synthesized nothing does not swallow the next gesture\'s pair', async ({ page }) => {
+    // First tap on an unselected cell: preventDefault-ed, no pair arrives, but the gate was armed.
+    await grid.tapCell(1, 1);
+    await page.clock.runFor(200);
+
+    // A drifted gesture on another cell, still inside the armed ceiling; the browser's
+    // compatibility pair for it carries firesTouchEvents === true (Blink) — no veto. The
+    // touchstart of this gesture must have disarmed the stale pair, so the pair is processed.
+    await grid.dispatchTouchDrag(3, 1, 20);
```
demtario: **Medium:** this scenario cannot occur in real Blink, so the test models a pair the engine never sends.

`dispatchTouchDrag(3, 1, 20)` targets (3,1) while (1,1) is the cell selected by the preceding tap, so `onTouchEnd` takes the `!selectedCellWasTouched(target)` branch (`event.ts` L713-718) and calls `event.preventDefault()`. Blink then synthesizes no compatibility pair for that gesture at all, which is the opposite of what the comment above states.

Run the drag on the selected cell instead, keeping the mouse pair and the assertion on (3,1):

```js
await grid.dispatchTouchDrag(1, 1, 20);
await grid.dispatchMouseEvent(3, 1, 'mousedown', true);
await grid.dispatchMouseEvent(3, 1, 'mouseup', true);
```

On (1,1) `selectedCellWasTouched()` is `true`, nothing is `preventDefault`-ed, and real Blink does emit the pair. The test still discriminates: without the reset the flag is armed and only 200 ms of virtual time has passed against the 500 ms ceiling, so the (3,1) pair is dropped and `expectSelectedCell(3, 1)` fails.

demtario (commented):
2 inline notes below.

handsontable/src/3rdparty/walkontable/AGENTS.md
```diff
@@ -182,6 +182,15 @@ geometry (`isRendered`, no boxes), the styles handler asks whether `getComputedS
...
...chstart`, so the residual never outlives one gesture. The `mouseup` half consumes the pair (`#synthesizedPairPending = false`), so any later real mouse event inside the ceiling is treated as real — a fill-handle grab, a drag-selection, or a right-click started with a mouse right after a tap works on engines that do not report the input origin (WebKit, Firefox), where the residual gap is only the FIRST synthesized pair itself. Dropping only `mouseup` (the original #12804 fix) fired `onCellMouseDown` twice per tap and re-armed the pairing slot and made iPad double-tap-to-edit nondeterministic. `TOUCH_SYNTHESIZED_MOUSE_WINDOW` lives in `helpers/dom/inputOrigin.ts`, alongside `getMouseEventTouchOrigin()`, and is shared with `tableView.ts` (its `#recentTouchEndTimeout` uses the same constant), so both layers agree on which events are synthesized. `tableView.ts`'s own gate (`#isSyntheticMouseEvent()`) deliberately keeps its older `reported ?? #recentTouchEnd` policy instead of adopting `event.ts`'s consume-once order: a Blink-flagged pair (`firesTouchEvents === true`) must never be allowed to close the editor through the outside-click handler, so the two gates are not to be "harmonized".
```
demtario: **Low, re-raise of [this thread](https://github.com/handsontable/handsontable/pull/13306#discussion_r3881269404), which I am leaving open:** the source comments were fixed, the normative doc was not.

`tableView.ts` L325 and L639 now read "both layers use the same fallback window", with an explicit note that the two policies deliberately differ. This line still carries the old claim: "is shared with `tableView.ts` (its `#recentTouchEndTimeout` uses the same constant), so both layers agree on which events are synthesized", while its own next sentence says the two gates "are not to be 'harmonized'". `helpers/dom/inputOrigin.ts` L11 repeats the same clause.

Since this file is what a future agent reads as authoritative, it is the one place the wording matters most. Fix: "so both layers use the same fallback window", here and at `inputOrigin.ts` L11. What is shared is the constant, not the policy.

danielzytohoc (commented):
QA review (deep pass on the head, plus adversarial verification of the gate's state machine, the callback consumers, and the legacy suites). The architecture is right and the earlier review rounds clearly made it better — consume-once symmetric drop, coords-based tap pairing, per-gesture state resets. Findings below are only what is **not already covered by the open threads**; I endorse all three of @demtario's open threads (notes on two of them included).

### 1. Low — no `touchcancel` handling: a cancelled gesture leaves `touchApplied` stuck, and real mouse mouseups then route into `#handleTouchTap`

`event.ts` registers no `touchcancel` listener (only dragToScroll does, and its AGENTS.md notes browsers cancel gestures often — `touchend` then never fires). `touchApplied` is set in `onTouchStart` and cleared only in `onTouchEnd`, so after a cancelled gesture (system gesture, dialog, browser claiming the touch) it stays `true` until the next `touchstart`. In that state, `onMouseUp`'s new `if (this.touchApplied) { this.#handleTouchTap(...); return; }` routes **every real mouse mouseup** into the tap detector — which has no `button` filter and pairs by coords only. Two right-clicks on the same cell within 1 s, or two drag-selections merely **ending** on the same cell within 1 s, fire a phantom double-click; meanwhile left mousedowns never arm `#dblClickOrigin`.

The stuck-flag hole is pre-existing (pre-PR it made `button === 0 || touchApplied` arm the slots for any button), so this is not a regression in magnitude — but the PR re-shapes it (drag-**end** pairing is new; pre-PR required start-cell === end-cell), and the new AGENTS.md claim "a real mouse click right after a tap cannot complete a double-click by pairing with it" does not hold while the flag is stuck (or while a finger is physically held down during mouse use). Cheap close, in-PR or as a follow-up: a `touchcancel` listener mirroring dragToScroll's pattern — reset `touchApplied`, `#deferredTouchStartEvent`, `#lastTapCoords`, and cancel the long-press timer.

### 2. Addendum to the open spec-realism thread: the same unreachable pair is modeled in the *other* drift spec, and the AGENTS.md drift sentence needs narrowing

The `preventDefault` block in `onTouchEnd` is not gated on `isTap` — it runs for any cancelable touchend over a cells-range target (verified on the head, `event.ts` ~695–723). So `'a drifted tap that the touch path treats as a scroll still lets the browser mouse pair select the cell'` has the same modeling issue as the last spec: its drift target (2,1) is unselected, `!selectedCellWasTouched` → `preventDefault()` → real engines synthesize **no** pair for that gesture at all. Both drift specs should drift over the **already-selected** cell (as proposed in the open thread for the last spec), and the new AGENTS.md sentence "its browser-synthesized compatibility pair is processed and selects the cell" should be narrowed to that reachable case — a drift ending over an unselected cell simply produces nothing.

### 3. On the open High (reset placement): endorse the move; one nuance on reachability

I verified the chain line-by-line on the head — it is mechanically real. On real-engine reachability my adversarial pass leans the other way: Blink queues the compatibility pair during touchend processing (it precedes the next hardware touchstart), and on WebKit either the fast-tap path delivers the pair ~0–50 ms after touchend, or the legacy ~350 ms regime converts the second touchstart into double-tap-zoom and never dispatches the first tap's pair; a preventDefault-ed touchend synthesizes nothing. That said, the proposed scroll-branch placement is strictly safer under *either* timing reading, costs nothing, keeps both new tests green, and script-dispatched events reach the current path trivially — so the move should land regardless, along with dropping the "bounded to one gesture" claim from the docs (a pair delayed past the next gesture's touchend leaks under either placement).

### 4. Low — PR body drifted from the shipped design

The **Fix** section still describes the pre-review iteration: helper in `helpers/dom/event.ts`, unit tests in `event.unit.ts`, and "touch-originated taps arm the mouseup pairing slot for `TOUCH_DBLTAP_TIMEOUT`". The shipped design is `helpers/dom/inputOrigin.ts` + `inputOrigin.unit.ts`, and touch taps never arm the mouse slots (independent `#handleTouchTap`). Since squash-merge makes this body the permanent record, please update it before merging.

### 5. CI state — neither red check is this PR

- **Unit / test**: the failure is `src/utils/__tests__/Interval.unit.ts` — wall-clock `setTimeout` timing assertions on an untouched file (spy-call counts at 50/150/250 ms), classic loaded-runner flake; the PR has zero overlap with it. Needs a re-run. (A separate de-flake task for that suite is queued on our side.)
- **Manual QA / verify**: by design — waits for a non-author `/manual-qa passed`. QA will sign off after the on-device pass on the next rc (RELEASE-869 resume protocol: double-tap cadence protocol, #12804 one-row check, selection handles, iPhone recheck).

### Verified clean (for reviewer confidence)

- **Callback consumers**: `onCellDblClick`/`onCellCornerDblClick` receiving the touchend `TouchEvent` is the *pre-existing* contract (pre-PR, pairing could only ever complete on the touch-driven `onMouseUp`, since the synthesized mouseup was already suppressed). editorManager, autofill's corner handler (no-arg), tableView's `preventDefault` — all TouchEvent-safe. The mergeCells `#onBeforeBeginEditing` non-MouseEvent early-return (veto bypass on touch) is pre-existing, unchanged here.
- **Corner double-tap**: reachable and correct — parentCell's `wtBorder`+`current` branch resolves the focus highlight; coords pairing even survives the TD recycling the old TD-identity pairing broke on.
- **Legacy suites, structurally**: the core Jasmine/Puppeteer environment has no `hasTouch`, so `isTouchSupported()` is false and the new mouse gate is never even installed; the legacy Walkontable suite contains zero touch simulation; `tableView.spec.js`'s touch groups were traced spec-by-spec (same-tick dispatch → identical gate outcome; per-instance state cannot leak across specs). Mobile Jasmine double-tap specs now pair through `#handleTouchTap` and only get more deterministic.
- **Mouse-only paths**: `!touchApplied && button === 0` ≡ old `button === 0 || touchApplied` for every desktop event; dblclick timing constants keep the literal 1000/500 values.
- Conventions: no banned APIs, JSDoc style consistent, changelog entry (post-reword) honestly records the lost single-tap-opens-editor side effect, helper correctly kept off `Handsontable.dom`.

**Verdict**: approve-ready once the open High (reset placement) and the drift-spec realism fixes land; the `touchcancel` gap can ride along or be filed as a follow-up. Release logistics reminder: base is `develop` — 18.1 needs the maintainer cherry-pick to `release/18.1.0` and an rc9 for the on-device QA sign-off.

handsontable/src/3rdparty/walkontable/src/event.ts
adrianspdev: Moved as suggested: the reset now lives in `onTouchEnd`'s scroll branch, next to the `#lastTapCoords` reset, with your comment; `onTouchStart` no longer touches the flag. So a delayed pair from the previous gesture stays gated by pending+ceiling, and one that arrives after a scroll gesture's `touchend` lands with `touchApplied === false` — it can't reach `#handleTouchTap`, worst case one extra `onCellMouseDown`. All three doc sites dropped the "bounded to one gesture" claim (a pair delayed past the next gesture's `touchend` leaks under any placement, as you said). Both F-round tests pass unchanged; the reworked one (below) is red with the scroll-branch reset removed.

tests/e2e/touch-tap-to-edit.spec.ts
adrianspdev: Applied your rework verbatim: the drift runs on the selected cell (1,1) — nothing is `preventDefault`-ed there, so real Blink does emit the pair — and the `firesTouchEvents: true` pair + `expectSelectedCell` assertion stay on (3,1). Re-proved it discriminates: with the scroll-branch reset removed, the pair is dropped against the still-armed 500 ms ceiling and the selection assertion fails.

handsontable/src/3rdparty/walkontable/AGENTS.md
adrianspdev: Fixed at both sites — walkontable AGENTS.md and the `inputOrigin.ts` constant JSDoc now say "so both layers use the same fallback window"; the "not to be harmonized" sentence stays. The constant is shared, the policy is not.

tests/fixtures/pages/TouchTapToEditPage.ts
adrianspdev: Applied exactly as proposed: the fixture's click counter moved to `document` (capture phase) — the ContextMenu's root is covered, grid clicks still count once — so the #12803 test now settles on the observed click count before exact reads of `afterCreateRow` and the row count, and the "counter can't see the menu" comment is gone. With every counter read behind a real barrier, `expectHookCount` (poll + re-read) had no remaining job: deleted; `expectHookCountExactly` is the one assertion shape left.

adrianspdev:
Re: the QA review (5064105131) — All four actionable points landed (plus the two open demtario threads you endorsed, which were fixed in the previous push):

1. **touchcancel** — taken in-PR: `onTouchCancel()` resets `touchApplied`, the deferred touchstart, the tap coords, the move/long-press flags, and cancels the long-press timer (dragToScroll's pattern), and releases the mouse-down flag (a long-press fires onMouseDown from its timer, and no mouseup ever follows a cancel). It deliberately leaves `#synthesizedPairPending`/`#lastTouchMouseUpAt` alone — a cancelled gesture never ran `onMouseUp`, so anything pending belongs to the previous tap and clearing it here would reopen the reset-placement leak. New spec case: script touchstart+touchcancel, then two real right-clicks on the same cell inside the double-tap window → no editor (red with the listener removed).
2. **Second drift spec** — reworked to the reachable shape: select first, drift over the selected cell, then the no-origin fallback pair must be processed (exact hook count). The AGENTS.md drift sentence is narrowed: a drift over an unselected cell is preventDefault-ed and synthesizes nothing.
3. **Reset placement** — landed in the previous push (`ce10395d57`), with the docs' "bounded to one gesture" claim dropped.
4. **PR body** — rewritten to the shipped design (inputOrigin.ts, consume-once gate, independent tap detector, coords pairing, per-gesture resets).
5. **CI** — agreed on both: Interval is an untouched-file wall-clock flake (re-run), Manual QA waits for the on-device pass per RELEASE-869.

The "verified clean" sweep matches our internal review's callback-contract and legacy-suite traces — useful to have them independently confirmed on the record.

budnix (dismissed):

demtario (commented):
2 inline notes below.

tests/e2e/touch-tap-to-edit.spec.ts
```diff
@@ -0,0 +1,225 @@
...
+  test('a drifted gesture over the selected cell lets the browser mouse pair through on the fallback path', async ({ page }) => {
+    // Select the cell first; its own tap is preventDefault-ed and synthesizes nothing.
+    await grid.tapCell(2, 1);
+    await page.clock.runFor(200);
+
+    // Drift over the SELECTED cell — nothing is preventDefault-ed there, so real engines do
+    // synthesize a compatibility pair for this gesture; its touchend (scroll branch) cleared the
+    // pending flag, so the pair, carrying no origin information, must be processed.
+    await grid.dispatchTouchDrag(2, 1, 20);
+    const mousedownsBefore = await grid.hookCount('beforeOnCellMouseDown');
```
demtario: **Medium:** the relative baseline is read *after* the drift, so it absorbs the regression the absolute assertion used to catch.

If `dispatchTouchDrag()` ever started firing cell mouse hooks, `mousedownsBefore` would read 2 instead of 1 and `mousedownsBefore + 1` would still pass. The version this replaced asserted `expectHookCountExactly('beforeOnCellMouseDown', 0)` right after the drift, and that was the only place in the suite proving the deferred-mousedown scroll classification (#11659, `event.ts` L672-682) still holds. I walked the rest of the file: no other test asserts it now.

`beforeOnCellMouseUp` also went unasserted, which matters for this test specifically, since its title claims *both* halves of the pair get through and the gate applies to `mousedown` and `mouseup` symmetrically (`event.ts` L369-386).

Absolute counts keep both properties, and the numbers are fixed by the preceding `tapCell(2, 1)`, which contributes exactly one of each:

```js
await grid.dispatchTouchDrag(2, 1, 20);
await grid.expectHookCountExactly('beforeOnCellMouseDown', 1);
await grid.expectHookCountExactly('beforeOnCellMouseUp', 1);

await grid.dispatchMouseEvent(2, 1, 'mousedown');
await grid.dispatchMouseEvent(2, 1, 'mouseup');

await grid.expectHookCountExactly('beforeOnCellMouseDown', 2);
await grid.expectHookCountExactly('beforeOnCellMouseUp', 2);
```

handsontable/src/3rdparty/walkontable/src/event.ts
```diff
@@ -576,15 +725,50 @@ class Event {
...
+   * after a cancel no `mouseup` ever follows, which would leave mouse-move selection dragging
+   * armed until the next click.
+   *
+   * @private
+   */
+  onTouchCancel() {
+    this.#cancelLongPressTimer();
+
+    this.touchApplied = false;
+    this.#mouseDown = false;
```
demtario, line 769: **Medium:** this line and the `#cancelLongPressTimer()` call at L762 ship with no test that can fail if either is deleted.

`LONG_PRESS_DELAY` is 500 (L19) and the new touchcancel spec advances 300 + 100 = 400 virtual ms under a paused `page.clock`, so the long-press timer never fires whether or not the cancel path cleared it. `#mouseDown` is read only by `onMouseMove` on the way to `onCellMouseOverOutside` (L488), which that spec never triggers. The test covers the `touchApplied` reset and nothing else in this handler.

Two cheap ways to close it, both worth having:

- In `tests/e2e/touch-tap-to-edit.spec.ts`, right after `dispatchTouchCancel()`, advance past the long-press delay and assert nothing fired: `await page.clock.runFor(600);` then the context menu is not visible and `beforeOnCellMouseDown` is still 0. That is what kills the timer-cancel mutant.
- For `#mouseDown` and `#deferredTouchStartEvent`, a case in Walkontable's own pipeline (`handsontable/src/3rdparty/walkontable/test/spec/event.spec.js`, which already carries touch cases) asserting the post-cancel state directly, since the flag has no reachable behavior in the Playwright fixture.

Separately, while reading this handler: it is the only gesture terminator that does not clear `#synthesizedPairPending`, although a cancelled gesture arms nothing, which is exactly the reason the scroll branch gives when it clears the flag at L738-741. A preventDefault-ed tap arms it with no pair coming; if the next gesture is cancelled rather than ended, the flag survives and the next real mouse pair inside the remaining ceiling is dropped on engines that do not report the input origin. One line here, same rationale, and it would make `AGENTS.md` L190 ("`touchcancel` resets the per-gesture state") accurate.

tests/e2e/touch-tap-to-edit.spec.ts
adrianspdev: Fixed in 3787d4f73b — the absolute counts are restored, in the shape you suggested: 1/1 for both hooks after the drift (the tap's contribution, the drift adds none), 2/2 after the script mouse pair. That re-pins the scroll classification (#11659) and brings `beforeOnCellMouseUp` back under assertion.

Determinism check before committing to absolute numbers: `dispatchTouchDrag()` is script-dispatched `TouchEvent`s, and the browser never synthesizes a compatibility pair for script events — the comment above the drift describes what real engines do, not what this fixture produces — so the counts cannot flake on a Chromium-synthesized pair. Verified by mutation: with the drift firing cell mouse hooks the 1/1 assertion fails, which the relative baseline indeed absorbed.

handsontable/src/3rdparty/walkontable/src/event.ts
adrianspdev: All addressed in 3787d4f73b:

- `onTouchCancel()` now clears `#synthesizedPairPending`, with the scroll-branch rationale in the JSDoc, and the AGENTS.md sentence now lists everything the cancel resets. A new spec pins it: tap arms the gate, the next gesture is cancelled, and a script mouse pair inside the ceiling must move the selection.
- The cancelled-gesture spec now advances 600 virtual ms right after the cancel and asserts `beforeOnCellMouseDown` is still 0 and the context menu stayed closed — that kills the `#cancelLongPressTimer()` mutant.
- `#mouseDown` went to Playwright rather than `event.spec.js`: the walkontable Jasmine suite is frozen for new cases (the presence gate blocks them; new coverage goes to Playwright), and the flag turned out reachable there after all — the long-press timer fires `onMouseDown`, and Walkontable listens for `mousemove` document-wide, so a post-cancel mousemove far outside the viewport drives the `onCellMouseOverOutside` drag path. The new spec long-presses, cancels, dispatches that mousemove, and asserts the selection did not extend.
- `#deferredTouchStartEvent` stays untested deliberately: `onTouchStart` unconditionally overwrites it at the next gesture, so the clear is defensive with no reachable behavior; a state-poking unit case would test the assignment, not a behavior.

Mutation-verified as a set: a build with all three cancel-path lines deleted fails exactly the three new/changed tests, one per line; the restored build is 12/12 green.

demtario (approved):

danielzytohoc (approved):

PR #13307, opened 2026-08-28
DEV-2701: Apply rowHeights and colWidths over manually resized rows and columns (#4371)

### LINKED ISSUE #4371: hot.updateSettings({rowHeights: []}) work strange.
### Description
When setting `manualRowResize: true`, `hot.updateSettings({rowHeights: []})` works strange.

### Steps to reproduce
See the demo
1. Click button `150` or `50`, it works well;
2. Drag the first row to change `height`;
3. Click button `150` or `50` again. **Sometimes but not always** it dosen't work.

### Demo
https://jsfiddle.net/ssbunny/omyd85x6/

### Your environment
* Handsontable version: 0.32.0
* Browser Name and version: Safari 9.0.2
* Operating System: OS X 10.11.2

budnix:
### Context

The PR fixes [#4371](https://github.com/handsontable/handsontable/issues/4371), open since 2017 and confirmed to still reproduce on 18.0.0 and `develop`.

After a user resizes a row by dragging, that row ignores every later `rowHeights` value. `updateSettings({ rowHeights })` resizes every row **except** the ones the user dragged, which is what the reporter described as "sometimes it does not work". The same applies to `colWidths` and `ManualColumnResize`.

Measured on `develop` with 5 rows and `manualRowResize: true`:

| Step | row 0 | rows 1-4 |
|---|---|---|
| `updateSettings({ rowHeights: 50 })` | 50 | 50 |
| drag row 0 to 120 | **120** | 50 |
| `updateSettings({ rowHeights: 150 })` | **120** | 150 |
| `updateSettings({ rowHeights: 50 })` | **120** | 50 |

`Core#getRowHeight` reads the `rowHeights` setting, then runs the `modifyRowHeight` hook and returns whatever the hook gives back. `ManualRowResize` returns its stored height from that hook, and nothing cleared the index map holding it: the plugin used the default `SETTING_KEYS`, so a `rowHeights`-only update never re-ran `updatePlugin()`. Even a full plugin cycle kept the height, because `disablePlugin()` saves the map into a private `#config` and `#onMapInit()` replays it.

A second defect surfaced while investigating. The `rowHeights` docs claimed the option sets the **minimum** height the resize plugins can apply. That held only with `AutoRowSize` enabled — that branch uses `Math.max`. Without it, the manual size overrode the option in both directions, so the same option behaved differently depending on an unrelated plugin.

### What changed

Passing `rowHeights` or `colWidths` to `updateSettings()` now re-declares the sizes and discards the ones stored by dragging.

The render-time priority is **unchanged**: between updates a dragged size still wins over the option, exactly as `Core#getRowHeight` documents. What changed is that an explicit size option in an `updateSettings()` call re-declares the sizes.

- Both resize plugins list their size option in `SETTING_KEYS`, so passing it runs `updatePlugin()`.
- The clear runs **after** `enablePlugin()`. It has to: `disablePlugin()` saves the map into `#config` and `#onMapInit()` replays it, so clearing earlier lets the plugin resurrect the sizes it just saved.
- A grid whose plugin option is an array keeps that array. The check reads the **merged** settings, not the call payload, so it covers both an array passed in the call and one set when the grid was built.
- A function-valued size option never clears. It states no fixed size, and a framework wrapper rebuilds an inline one on every render.
- New API: `ManualRowResize#clearManualSize()`, matching the method `ManualColumnResize` already had, plus `clearManualSizes()` on both plugins.

### Wrapper changes

The core change alone would have regressed React. `SettingsMapper.getSettings()` copies every prop into the payload and the effect that calls `updateSettings()` runs on **every commit**, so `rowHeights` arrived unchanged on each re-render and would have discarded the user's drag every time. Angular does the same on any settings change. Vue already diffed each key against the current settings and was unaffected.

No core-side heuristic can separate a spurious re-pass from a deliberate call, so the wrappers are the right layer to fix. React and Angular now skip an unchanged size value, following the pattern Vue already used.

### Behavior change

An app that passes `rowHeights` on a timer, or inside a broad settings blob, will now lose the sizes its users produced by dragging. Three ways to keep them:

- leave the size option out of the `updateSettings()` call;
- pass `manualRowResize` / `manualColumnResize` as an array in the same call;
- manage the sizes yourself with `clearManualSizes()`.

This is not marked as a breaking change: it restores the documented behavior of `updateSettings()`, and the previous behavior contradicted the `rowHeights` documentation. Flagging it here so the reviewer can weigh that call.

### How has this been tested?

New Playwright spec that drives a **real pointer drag** on the resize handle, so the stored size is produced the way a user produces it. Verified red without the fix (the two bug cases fail, the four guard cases still pass) and green with it.

```bash
# New E2E spec (HOT_TEST_PORT only needed if another checkout holds 8123)
cd tests && HOT_TEST_PORT=8451 npx playwright test --project=e2e-main \
  e2e/manual-resize-size-option-reset.spec.ts

# Core
npm run test:e2e --prefix handsontable -- --testPathPattern='autoRowSize|autoColumnSize|manualRowResize|manualColumnResize'

# Wrappers
```

Results: core E2E 9839 specs, unit 4073, resize + auto-size E2E 259, React 97, Vue 39, Angular 298, new Playwright spec 6. Types and lint clean.

The existing test from #11849 (`manualRowResize` array must not shrink rows below the `AutoRowSize` content height) still passes, and the inline-start clone stays aligned with the master table.

### Documentation

- Corrected the `rowHeights` "minimum" sentence in `metaSchema.ts`, and the `colWidths` counterpart.
- Documented the reset in the `Core#getRowHeight` JSDoc and in both the row-height and column-width guides.

### Types of changes

### Related issue(s):

1. DEV-2701
2. Fixes #4371

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2701

---

> [!NOTE]
> **Medium Risk**
> Changes sizing precedence and `updatePlugin` lifecycle for manual resize plugins; apps that pass `rowHeights`/`colWidths` on every `updateSettings` or re-render will now clear user-dragged sizes unless wrappers skip unchanged values or callers use the documented escape hatches.
> 
> **Overview**
> Fixes **#4371**: after a user drags to resize a row or column, later `updateSettings({ rowHeights })` or `updateSettings({ colWidths })` now applies to **every** row/column instead of skipping the dragged ones.
> 
> **Core:** `ManualRowResize` and `ManualColumnResize` register `rowHeights` / `minRowHeights` / `colWidths` in `SETTING_KEYS` and use `redeclaresManualSizes()` to clear the internal size map when those options are re-declared (with exceptions for function values, non-empty `manual*Resize` arrays, and `undefined`). `updatePlugin()` avoids a full disable/enable cycle on size-only updates so preset arrays are not replayed over user drags. New **`clearManualSizes()`** on both plugins and **`clearManualSize(row)`** on `ManualRowResize` (parity with columns).
> 
> **Wrappers:** React and Angular no longer forward unchanged `rowHeights` / `colWidths` on re-render, so manual sizes are not wiped on every commit.
> 
> Docs, changelogs, unit tests for `redeclaresManualSizes`, and Playwright E2E with real handle drags are included. Minor doc tweak: `textExtractor` marked grid-level in option-levels metadata.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 33afcb241c40c1e81238f51f92d267382117edca. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2701 Apply rowHeights and colWidths over manually resized rows and columns (#4371)](https://app.clickup.com/t/9015210959/DEV-2701)

handsontable/src/plugins/manualResize/utils.ts
```diff
@@ -73,3 +73,42 @@ export function shouldRefreshHandleAfterAutoResize(
...
+ * @param {*} pluginSetting The plugin option, `manualRowResize` or `manualColumnResize`, read from
+ *                          the merged settings.
+ * @returns {boolean}
+ */
+export function redeclaresManualSizes(
+  newSettings: Record<string, unknown> | undefined,
+  sizeOptionKey: string,
+  pluginSetting: unknown,
+): boolean {
+  if (!newSettings || Array.isArray(pluginSetting)) {
```
Haxikowy: `Array.isArray(pluginSetting)` is true for `[]`, so a grid built with `manualRowResize: []` never honors `rowHeights` again: `isEnabled()` is `!![]` so the plugin runs, but the clear is suppressed forever. `[]` is a normal way to spell "enabled, no presets".

```suggestion
  if (!newSettings || (Array.isArray(pluginSetting) && pluginSetting.length > 0)) {
```

handsontable/src/plugins/manualRowResize/manualRowResize.ts
```diff
@@ -53,6 +54,17 @@ export class ManualRowResize extends BasePlugin {
...
 
+  /**
+   * Returns the setting keys that trigger a plugin update after an `updateSettings()` call. The
+   * `rowHeights` option is listed alongside the plugin's own key, so that re-declaring the row
+   * heights discards the heights kept from earlier manual resizing.
+   *
+   * @returns {string[]}
+   */
+  static get SETTING_KEYS(): string[] {
+    return [PLUGIN_KEY, 'rowHeights'];
```
Haxikowy: `core.ts:5334` reads `tableMeta.rowHeights ?? tableMeta.minRowHeights`, so `updateSettings({ minRowHeights: 50 })` re-declares heights exactly as `rowHeights` does, but the plugin never updates and the dragged rows still win. Needs adding here, at the `redeclaresManualSizes` call on line 237, and to `SIZE_SETTING_KEYS` in the Angular wrapper.

handsontable/src/plugins/manualRowResize/manualRowResize.ts
```diff
@@ -196,12 +208,37 @@ export class ManualRowResize extends BasePlugin {
...
+      // `BasePlugin#onUpdateSettings` feeds `updatePluginSettings()` with `newSettings[PLUGIN_KEY]`,
+      // which a `rowHeights`-only update does not carry. Restore the option from the merged settings
+      // so `getSetting()` keeps reporting it.
+      this.updatePluginSettings(this.hot.getSettings()[PLUGIN_KEY]);
+    }
 
-    super.updatePlugin();
+    // Runs after the re-initialization, so that the heights replayed on the map's `init` hook are
+    // discarded too.
+    if (redeclaresManualSizes(newSettings, 'rowHeights', this.hot.getSettings()[PLUGIN_KEY])) {
```
Haxikowy: `updateSettings({ manualRowResize: false })` then `{ rowHeights: 30 }` then `{ manualRowResize: true }` brings every dragged height back: `onUpdateSettings` returns after `disablePlugin()` without reaching `updatePlugin()`, `disablePlugin()` snapshots the live map into `#config` (line 253), and `#onMapInit` replays it on the next enable. Clearing `#config` on the enable path fixes it, and it wants a spec since it contradicts the changelog.

handsontable/src/dataMap/metaManager/metaSchema.ts
```diff
@@ -5871,8 +5884,21 @@ export default (): Record<string, unknown> => {
...
-     * The `rowHeights` option also sets the minimum row height that can be set
-     * via the {@link ManualRowResize} and {@link AutoRowSize} plugins (if they are enabled).
+     * When the {@link AutoRowSize} plugin is enabled, `rowHeights` sets the minimum row height, so a
+     * row still grows to fit its content.
+     *
+     * A row resized by dragging keeps its height until the height is re-declared. Passing
+     * `rowHeights` to [`updateSettings()`](@/api/core.md#updatesettings) re-declares the heights and
+     * discards the ones stored by the {@link ManualRowResize} plugin. To keep those instead, leave
+     * `rowHeights` out of the call, or pass
+     * [`manualRowResize`](#manualRowResize) as an array in the same call.
```
Haxikowy: "To keep those instead ... pass `manualRowResize` as an array in the same call" does the opposite: the spec at `manual-resize-size-option-reset.spec.ts:67` drags row 0 to ~120, passes `{ manualRowResize: [90 x 5], rowHeights: 50 }`, and asserts every row lands on 90. The array replaces the dragged size and only an array set at build time keeps it, so the same wording at `:1258`, in the changelog title, and in the PR body needs the fix too. Both guides already say "replace" correctly.

handsontable/src/dataMap/metaManager/metaSchema.ts
```diff
@@ -5871,8 +5884,21 @@ export default (): Record<string, unknown> => {
      * | A function  | Set row heights dynamically,<br>on each render                                                      | `rowHeights(visualRowIndex) { return visualRowIndex * 10; }` |
      * | `undefined` | Used by the [modifyRowHeight](@/api/hooks.md#modifyRowHeight) hook,<br>to detect row height changes | `rowHeights: undefined`                                      |
      *
-     * The `rowHeights` option also sets the minimum row height that can be set
-     * via the {@link ManualRowResize} and {@link AutoRowSize} plugins (if they are enabled).
+     * When the {@link AutoRowSize} plugin is enabled, `rowHeights` sets the minimum row height, so a
+     * row still grows to fit its content.
+     *
+     * A row resized by dragging keeps its height until the height is re-declared. Passing
```
Haxikowy: With AutoRowSize enabled, `#onModifyRowHeight` takes `Math.max(stored, incoming)` (`manualRowResize.ts:829`), so a taller `rowHeights` or taller content overrides the dragged height on the next render with no re-declaration at all. Scoping this to "When AutoRowSize is disabled" keeps it true; the `colWidths` twin at `:1254` is accurate as written, since the column plugin has no `Math.max` branch.

handsontable/src/plugins/manualColumnResize/manualColumnResize.ts
```diff
@@ -287,11 +324,38 @@ export class ManualColumnResize extends BasePlugin {
...
    */
   clearManualSize(column: number): void {
+    // The map only exists while the plugin is enabled, and a disabled plugin stores no widths.
+    if (!this.enabled) {
+      return;
+    }
+
     const physicalColumn = this.hot.toPhysicalColumn(column);
 
     this.#columnWidthsMap.setValueAtIndex(physicalColumn, null);
```
Haxikowy: The new row twin guards `physicalRow !== null` (`manualRowResize.ts:339`) and this one does not, so an out-of-range visual index writes `indexedValues['null']` and fires a spurious `invalidateColumnWidthCache()`. The JSDoc at `:322` also still reads "Clears the cache for the specified column index" while the row version got full prose and an `@example`.

```suggestion
    if (physicalColumn !== null) {
      this.#columnWidthsMap.setValueAtIndex(physicalColumn, null);
    }
```

handsontable/src/plugins/manualRowResize/manualRowResize.ts
```diff
@@ -277,6 +314,56 @@ export class ManualRowResize extends BasePlugin {
...
+   *
+   * @example
+   * ```js
+   * const resizePlugin = hot.getPlugin('manualRowResize');
+   *
+   * resizePlugin.clearManualSizes();
+   * hot.render();
+   * ```
+   */
+  clearManualSizes(): void {
```
Haxikowy, line 359: `clearManualSizes()` on both plugins and the new `clearManualSize()` appear only in the two `.types.ts` files, which prove the signature compiles and nothing more. These are what the guides tell users to call when the function-option or wrapper case leaves sizes stuck, so they want a drag plus clear plus render assertion, and one for the new `!this.enabled` no-op.

wrappers/react-wrapper/src/settingsMapper.ts
```diff
@@ -32,13 +41,21 @@ export class SettingsMapper {
...
-      if (!isInit && initOnlySettingKeys.includes(key)) {
+      // Checked before the value comparison below. The two lists do not overlap today, but a
+      // changed deep-comparable key used to return early and skip this guard entirely, which would
+      // have forwarded an init-only key to `updateSettings` as soon as one joined both lists.
+      if (initOnlySettingKeys.includes(key)) {
         return true;
       }
+
+      if (DEEP_COMPARABLE_SETTINGS.includes(key)) {
+        return areEquivalentSettingsValue(prevProps[key], properties[key]);
```
Haxikowy: This compares `prevProps` against the incoming props, while Angular (`hot-table.component.ts:238`) and Vue (`helpers.ts:112`) compare against the live grid settings. After anything changes `rowHeights` through the instance ref, a React re-render with the unchanged prop no longer re-asserts it, where before this PR it always did.

docs/content/guides/rows/row-height/row-height.md
```diff
@@ -192,6 +192,31 @@ When you set a row height programmatically with [`ManualRowResize#setManualSize(
...
+```js
+// every row is 50px again, including the rows the user dragged
+hot.updateSettings({ rowHeights: 50 });
+```
+
+To resize rows without discarding what the user dragged, leave `rowHeights` out of the call. To replace the stored heights with your own, pass [`manualRowResize`](@/api/options.md#manualrowresize) as an array. You can also clear them yourself with [`ManualRowResize#clearManualSize()`](@/api/manualRowResize.md#clearmanualsize) for one row, or [`clearManualSizes()`](@/api/manualRowResize.md#clearmanualsizes) for all of them, followed by `hot.render()`.
+
+::: tip
+
+Two cases never discard the stored heights, so passing `rowHeights` does not reset anything:
```
Haxikowy: A grid built with `manualRowResize` as an array is also never reset, since `redeclaresManualSizes` short-circuits on `Array.isArray`, and the PR tests exactly that at `spec.ts:79`. Vue 3 belongs in the wrapper bullet too, because `prepareSettings` drops unchanged keys via `simpleEqual` and Vue users hit the same case. Same two gaps at `column-width.md:263`.

tests/e2e/manual-resize-size-option-reset.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+    // silently revert the row to 90 - neither the dragged height nor the requested one.
+    await grid.applySettings('rowsArray', { rowHeights: SHORT });
+
+    const heights = await grid.rowHeights('rows-array');
+
+    expect(heights[0]).toBe(dragged);
+    expect(heights.slice(1)).toEqual(new Array(4).fill(ARRAY_HEIGHT));
+  });
+
+  test('applies colWidths to a column that was resized by dragging', async () => {
```
Haxikowy, line 111: Rows get five cases, columns get two, and the fixture has no `manualColumnResize`-array grid at all, so the array short-circuit and the re-init-versus-restore branch are proved only on the row copy of logic duplicated verbatim in `manualColumnResize.ts:228`. No case enables AutoRowSize or AutoColumnSize either, though that different `Math.max` branch is the PR's stated premise.

Haxikowy (commented):
Three things worth deciding before this lands.

**The breaking-change call.** `.ai/BREAKING-CHANGES.md` lists "changing API signatures or behavior" as breaking, and its only exemption covers APIs outside the public reference. `updateSettings`, `rowHeights` and `colWidths` are all in it. A plain-JS app that re-passes a settings snapshot now destroys every size its users dragged, with no warning and no opt-out beyond dropping the key. The React and Angular guards in this PR exist precisely because that vector is real, and jQuery callers get no equivalent. "It restores documented behavior" argues for making the change and for the `fixed` type, not for `breaking: false`. Either flag it with a step in the existing 18.1 to 18.2 migration guide, or record the team's decision to accept it in bold in the description.

**`handsontable/src/helpers/wrappers/jquery.ts:32`** forwards the caller's whole settings object to `updateSettings()` on the re-init path, so `$el.handsontable({...})` against a live instance wipes drags exactly as an unguarded React render would. It is the only internal caller that re-passes a full config; `columnSorting.ts:212` and the init call at `core.ts:1747` are both safe.

**Changelog.** One `fixed` entry carries three new public methods, and `framework: none` despite user-visible React and Angular changes. `.changelogs/README.md` permits a second file when the `type` differs, so `clearManualSize`/`clearManualSizes` belong in `added` or they never reach the Added section of the release notes.

Smaller: `handsontable/AGENTS.md` is not updated for two conventions future work will get wrong (`SETTING_KEYS` may carry a key that is not the plugin's own option; the clear must run after the disable/enable because `#onMapInit` replays `#config`). The PR body has stale claims: the spec now holds seven tests, not six; the `colWidths` block never contained a "minimum" sentence to correct; and the `#11849` spec cited as still passing is not in the tree under that name.

handsontable/src/plugins/manualResize/utils.ts
budnix: Good spot — fixed in `4bd7b1a2f7`, the guard now needs a non-empty array, with a fixture grid built from `[]` to prove it.

handsontable/src/plugins/manualRowResize/manualRowResize.ts
budnix: You're right — its own docs call it an alias for `rowHeights`, so both keys now share `ROW_SIZE_OPTIONS` and go through the same path in `4bd7b1a2f7`, wrappers included.

handsontable/src/plugins/manualRowResize/manualRowResize.ts
budnix: Fair point, but this one predates the PR — the plugin gets no callback at all while it is disabled, so the same sequence replays the drags on develop too. Clearing `#config` on enable would drop the restore-on-re-enable behavior it has always had, so I would rather do it as its own ticket.

handsontable/src/dataMap/metaManager/metaSchema.ts
budnix: Yep, my mistake — the array replaces the dragged sizes, and only an array set at build time keeps them. Reworded both option blocks and the changelog title in `4bd7b1a2f7`.

handsontable/src/dataMap/metaManager/metaSchema.ts
budnix: Agreed — scoped that sentence to AutoRowSize being disabled in `4bd7b1a2f7`, and added a test for the `Math.max` case. You were right about the `colWidths` twin too, the column plugin has no such branch.

handsontable/src/plugins/manualColumnResize/manualColumnResize.ts
budnix: Nice catch — added the null guard and brought the JSDoc up to match the row twin in `4bd7b1a2f7`.

handsontable/src/plugins/manualRowResize/manualRowResize.ts
budnix: You are right, the type tests proved nothing on their own. Added drag-then-clear cases for one row, for all of them, for the column twin, and for the disabled-plugin no-op in `4bd7b1a2f7`.

wrappers/react-wrapper/src/settingsMapper.ts
budnix: Good point — the size keys now diff against the grid's live settings like Angular and Vue do, in `4bd7b1a2f7`, so an imperative change through the ref is re-asserted again.

docs/content/guides/rows/row-height/row-height.md
budnix: Agreed — added the array-built grid to the list and Vue 3 to the wrapper bullet, in both guides and both option blocks, in `4bd7b1a2f7`.

tests/e2e/manual-resize-size-option-reset.spec.ts
budnix: Fair point — added a `manualColumnResize` array grid with the column drag-and-re-declare case, a column clear case, and an AutoRowSize grid. The spec went from 7 tests to 15 in `4bd7b1a2f7`.

docs/content/guides/columns/column-width/column-width.md
```diff
@@ -249,6 +249,31 @@ When you set a column width programmatically with [`ManualColumnResize#setManual
...
+```
+
+To resize columns without discarding what the user dragged, leave `colWidths` out of the call. To replace the stored widths with your own, pass [`manualColumnResize`](@/api/options.md#manualcolumnresize) as an array. You can also clear them yourself with [`ManualColumnResize#clearManualSize()`](@/api/manualColumnResize.md#clearmanualsize) for one column, or [`clearManualSizes()`](@/api/manualColumnResize.md#clearmanualsizes) for all of them, followed by `hot.render()`.
+
+::: tip
+
+Two cases never discard the stored widths, so passing `colWidths` does not reset anything:
+
+- `colWidths` set to a function. A function states no fixed width, so the stored widths are left alone.
+- The React and Angular wrappers, where a `colWidths` prop whose value did not change is not forwarded to `updateSettings()`. Re-applying the same value is not an update.
```
demtario: **medium**: this lists only React and Angular as skipping an unchanged `colWidths` prop, but the `colWidths` JSDoc added in this same PR (`metaSchema.ts`) says React, Angular, **and Vue**. Vue's `prepareSettings`/`simpleEqual` diff already skips forwarding an unchanged value, so this guide under-reports the exception and could send a Vue user to `clearManualSizes()` unnecessarily. Add "and Vue" to match the JSDoc.

docs/content/guides/rows/row-height/row-height.md
```diff
@@ -192,6 +192,31 @@ When you set a row height programmatically with [`ManualRowResize#setManualSize(
...
+
+- `rowHeights` set to a function. A function states no fixed height, so the stored heights are left alone.
+- The React and Angular wrappers, where a `rowHeights` prop whose value did not change is not forwarded to `updateSettings()`. Re-applying the same value is not an update.
```
demtario: **medium**: same gap as the `colWidths` guide — this lists only React and Angular, but the `rowHeights` JSDoc added in this PR (`metaSchema.ts`) says React, Angular, **and Vue**. Add "and Vue" here too.

handsontable/src/plugins/manualColumnResize/utils.ts
```diff
@@ -1,6 +1,9 @@
 export {
+  COLUMN_SIZE_OPTIONS,
+  ROW_SIZE_OPTIONS,
```
demtario: **medium**: `ROW_SIZE_OPTIONS` is re-exported here only so `manualColumnResize/__tests__/utils.unit.js` can test row-specific `redeclaresManualSizes` behavior under a `describe('manualColumnResize/utils', ...)` block — `manualColumnResize.ts` itself never imports `ROW_SIZE_OPTIONS`. That's row logic tested under a "column" suite name, plus a dead re-export to support it. Move the `redeclaresManualSizes`/`*_SIZE_OPTIONS` tests to a new `manualResize/__tests__/utils.unit.js` and drop this re-export.

handsontable/src/plugins/manualColumnResize/manualColumnResize.ts
```diff
@@ -282,14 +320,55 @@ export class ManualColumnResize extends BasePlugin {
   }
 
   /**
-   * Clears the cache for the specified column index.
+   * Clears the width stored for the specified column, so the column falls back to the width coming
+   * from the [`colWidths`](@/api/options.md#colwidths) option or from the theme. Call `render()`
```
demtario: **low**: this doc (and the identical wording on `clearManualSizes()` further below) says a cleared column falls back to "the width coming from the `colWidths` option or from the theme." `Core#getColWidth` actually falls back to the hardcoded `DEFAULT_COLUMN_WIDTH` constant, not anything theme-derived — unlike `getRowHeight()`, which genuinely reads `stylesHandler.getDefaultRowHeight()`. Drop "or from the theme" for the column variant (or say "the built-in default width").

demtario (commented):
4 inline notes below. 3 more have no changed line to anchor to:

- No unit/E2E test for the new `physicalColumn`/`physicalRow !== null` guard in `clearManualSize()` (both plugins) — only index `0` is ever exercised, so the guard has zero coverage.
- `Core#getColWidth()`'s JSDoc has no equivalent of the new `getRowHeight()` paragraph explaining that `updateSettings({ rowHeights })`/`{ colWidths }` discards the sizes stored by manual resizing.
- No column equivalent of the `rows-empty-array` (`manualRowResize: []`) fixture/E2E/unit test, despite `redeclaresManualSizes` being logic shared with columns.

docs/content/guides/columns/column-width/column-width.md
budnix: Good spot — added Vue in `33afcb241c`, plus the array-grid case this guide was also missing next to it.

docs/content/guides/rows/row-height/row-height.md
budnix: You're right — same fix in `33afcb241c`, and the array-grid bullet that was missing here too.

handsontable/src/plugins/manualColumnResize/utils.ts
budnix: Agreed — the `redeclaresManualSizes` tests moved to a new `manualResize/__tests__/utils.unit.js` and the re-export is gone in `33afcb241c`.

handsontable/src/plugins/manualColumnResize/manualColumnResize.ts
budnix: Nice catch — checked it, `getColWidth` ends at `DEFAULT_COLUMN_WIDTH` while only `getRowHeight` reads the theme. Both column methods now say the built-in default width, in `33afcb241c`.

demtario (approved):

PR #13308, opened 2026-08-28
DEV-2702: Add the textExtractor option so headers export as displayed text

### LINKED ISSUE #4088: Custom headers are being shown in exported CSV files
### Description
See Demo, when export as file, the headers are exported as HTML. Generally we don't want to see HTML tags in the header. Is there a way to strip away all the HTML tags from the headers? 

### Demo
http://jsfiddle.net/2zjsuuda/2/

budnix:
### Context

The PR adds a new grid option, `textExtractor`, so that column and row headers reach a file or the clipboard as the text the grid displays rather than as their raw HTML.

This resolves GitHub issue #4088, open since February 2017. A header setting is also its display string: a header of `'<b>Total</b>'` renders as **Total** in the grid, while the export writes the value it was given, so the file receives the literal `<b>Total</b>`. The screen and the file disagree.

The problem is wider than the issue reports. It was verified still reproducing on `develop` across `colHeaders`, `columns[].title`, `rowHeaders`, `nestedHeaders`, XLSX (which did not exist in 2017 and inherited the behavior by reusing the same data provider), and copy/paste.

**No existing option fixes it.** `sanitizeValues` is a CSV-injection guard — `true` only prefixes `'` to values starting with `= + - @`, so it does nothing to `<b>`. The grid-level `sanitizer` option lists headers among its covered surfaces, but it applies to DOM writes only: with a tag-stripping sanitizer configured, the grid shows `Total` while CSV, XLSX and the clipboard still carry `<b>Total</b>`, so a correctly configured sanitizer *widens* the gap. Routing export values through `sanitizer` does not work either, because a sanitizer returns HTML *source* — plain headers come back entity-encoded (`R&D` becomes `R&amp;D`), and the allowlist style the `sanitizer` docs show first returns `<b>Bold</b>` unchanged.

**The option.** `textExtractor` is the counterpart to `sanitizer`, and the pair reads as one idea:

- `sanitizer` — how content is written **to the screen** (HTML in, HTML out).
- `textExtractor` — how content becomes **text everywhere else** (HTML in, text out).

```js
textExtractor?: true | ((content: string, context: TextExtractorContext) => string);
```

The default is `undefined`, which passes content through untouched, so every existing export produces the same bytes as before. `true` selects the built-in extraction: it runs a configured `sanitizer` first, under its existing `'header'` surface, then reduces the result to text. A function gives full control per surface.

Running the sanitizer first is not decoration — a sanitizer may delete text rather than unwrap it (an allowlist filter drops `<script>alert()</script>` whole), so extracting from the unsanitized value would put content in a file that the grid never displayed. Parsing is also what turns entities back into the characters they stand for, so a header shown as `Tom & Jerry` is written as `Tom & Jerry` and not `Tom &amp; Jerry`.

**Deliberately generic.** `src/utils/textExtractor.ts` mirrors `src/utils/sanitizer.ts`, so any future consumer — a printer, a search index, an assistive label — opts in with a single `extractText(hot, value, 'Plugin.surface')` call and no change to the option or to core. `TextExtractorContext` keeps the `(string & {})` member, so a third-party plugin can pass a surface of its own.

**Scope.** Header surfaces only: `'ExportFile.columnHeader'`, `'ExportFile.rowHeader'`, `'CopyPaste.columnHeader'`. Cell surfaces are excluded on purpose — a cell exports its data value and renderer markup never reaches a file, so parsing cell data as HTML would destroy a legitimate value such as `a<b`. There is a test pinning that.

Two implementation notes worth a reviewer's attention:

1. `copyPaste.getRangedData()` feeds `text/plain`, `text/html` **and** the `beforeCopy` / `afterCopy` payloads. The projection therefore happens in `#setClipboardData()` and maps rather than mutates, so the hooks still receive exactly what they always did.
2. `_dataToHTML()` escapes the values it writes, so an unprojected header reaches a rich-text target as the visible characters `<b>Bold</b>`. Both clipboard flavours are projected; only the source-data flavour stays raw, so a grid-to-grid copy still round-trips object-valued cells.

### How has this been tested?

New unit tests, type tests and E2E cases. Every suite below was run before this PR was opened.

```bash
# Unit tests for the new util (30 tests)
npm run test:unit --prefix handsontable -- --testPathPattern=textExtractor

# Full unit suite (4093 tests, 330 suites)

# E2E for the wired plugins (538 specs)
npm run test:e2e --prefix handsontable -- --testPathPattern="copyPaste|exportFile"

# Type tests, including the Required<GridSettings> gate

# Lint
```

Coverage added:

- `src/utils/__tests__/textExtractor.unit.ts` — both option forms, non-string pass-through, entity decoding, sanitizer-first ordering, and the fast-path guard.
- `src/__tests__/core/textExtractor.types.ts` — the `true` shorthand, call arity, the wrapper mapped-type chain, and literal exhaustiveness.
- `src/plugins/exportFile/__tests__/types/csv.spec.js` — including a case asserting the **unset** default still exports raw markup (no behavior change), and one asserting cell data is untouched.
- `src/plugins/exportFile/__tests__/types/xlsx/headers.spec.js` — the unset default and the projected header.
- `src/plugins/copyPaste/__tests__/copy.spec.js` — both clipboard flavours and a `beforeCopy` payload guard.

One thing a reviewer should know: the built-in extraction skips parsing for strings it cannot change, and the guard is `/[<&]/` rather than the existing `HTML_CHARACTERS` regex. `HTML_CHARACTERS` requires a closing `;` for an entity, but the HTML parser decodes the legacy named references without one — `'A &amp B'` parses to `'A & B'` and `'&ltx'` to `'<x'`. Using it would have written to a file something the grid never displayed. Both cases are pinned by tests.

### Types of changes

### Related issue(s):

1. DEV-2702
2. Resolves #4088

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

Documentation updated in this PR: the option's JSDoc in `metaSchema.ts` (the API reference), a new "Export headers as plain text" section in the export-to-CSV guide, and notes in `handsontable/AGENTS.md` and `handsontable/.ai/CONCERNS.md` so the next contributor routes new non-DOM consumers through `utils/textExtractor.ts` rather than through `sanitizer`.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2702

---

> [!NOTE]
> **Medium Risk**
> Changes export and clipboard output when enabled; entity decoding can write formula-like header text into spreadsheets (documented—pair with `sanitizeValues` for untrusted headers). Default-off preserves existing export bytes.
> 
> **Overview**
> Adds a grid-level **`textExtractor`** option (`true` or a custom function) so column/row/nested headers can leave the grid as **plain text that matches what users see**, instead of raw HTML strings. Default behavior is unchanged when the option is unset.
> 
> A new **`utils/textExtractor.ts`** helper (`extractText`, `getTextExtractor`) mirrors the sanitizer pattern: built-in mode runs an optional **`sanitizer`** on the `'header'` surface first, then parses markup in an inert `<template>` (aligned with `HTML_CHARACTERS` / `fastInnerHTML`). Cell **data** is never projected.
> 
> **ExportFile** routes column, row, and nested header labels through `extractText` (CSV and XLSX). **CopyPaste** projects only column-header rows into `text/plain` and `text/html` after `beforeCopy`, using row **identity** so hooks still see the original arrays and reshaped copies do not mis-treat data rows as headers.
> 
> Ships **`TextExtractorContext`** typing, docs (export guide, TypeScript types, AGENTS/CONCERNS), changelog entry **#13308**, and broad unit/E2E/type tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 50fa927d000367fda13c152b38b23fdb7f7601bc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2702 Add the `textExtractor` option so headers export as displayed text (GH #4088)](https://app.clickup.com/t/9015210959/DEV-2702)

adrianspdev:
@budnix One thing to settle in review: the TS type is `true | function`, but the JSDoc in `metaSchema.ts` says `boolean | function`. So JS users can pass `false` (reads as "off", unit-tested), but TS users can't write `textExtractor: false` or `textExtractor: flag` — it doesn't compile. Deliberate for now; if you prefer the type to match the docs, I'll widen it to `boolean | function`. No runtime difference either way.

budnix:
@adrianspdev Good point — widened it to `boolean | function` in `50fa927d00`, so a plain flag works now and the type matches the JSDoc.

adrianspdev (approved):

qunabu:
Heads-up: this landed without a `@configScope` tag on `textExtractor`, so `develop` is red on `Unit / test`:

```
FAIL test/__tests__/optionLevels.unit.js
  ● configuration-option levels › encountered a declaration exception
    Option `textExtractor` has no @configScope tag in metaSchema.ts.
      at parseOptions (scripts/utils/option-levels.js:79:13)
```

Because `Visual`, `Build`, `E2E`, `Integration`, `Performance` and `Walkontable` are all gated on `!failure()`, this skips the entire rest of the pipeline on every PR opened against `develop` — not just the unit job.

The fix looks like one line. The option's own JSDoc already states the scope:

> This option can only be set at the grid level. It has no effect when set in the `columns`, `cells`, or `cell` options.

which matches `sanitizer` — the counterpart named in the same block:

```js
     * @configScope grid
     */
    textExtractor: undefined,
```

Flagging rather than patching it, since it is your feature and you may want it in the same changelog entry.

budnix:
> Heads-up: this landed without a `@configScope` tag on `textExtractor`, so `develop` is red on `Unit / test`:
> 
> ```
> FAIL test/__tests__/optionLevels.unit.js
>   ● configuration-option levels › encountered a declaration exception
>     Option `textExtractor` has no @configScope tag in metaSchema.ts.
>       at parseOptions (scripts/utils/option-levels.js:79:13)
> ```
> 
> Because `Visual`, `Build`, `E2E`, `Integration`, `Performance` and `Walkontable` are all gated on `!failure()`, this skips the entire rest of the pipeline on every PR opened against `develop` — not just the unit job.
> 
> The fix looks like one line. The option's own JSDoc already states the scope:
> 
> > This option can only be set at the grid level. It has no effect when set in the `columns`, `cells`, or `cell` options.
> 
> which matches `sanitizer` — the counterpart named in the same block:
> 
> ```js
>      * @configScope grid
>      */
>     textExtractor: undefined,
> ```
> 
> Flagging rather than patching it, since it is your feature and you may want it in the same changelog entry.

Fixed in c394b365aeb208a2d476e3e892046f0ef565b8f4

PR #13310, opened 2026-08-28
DEV-2679: Drop a selection a trim left pointing at another record

demtario:
### Context

The PR fixes silent data corruption: a trimming index map (Filters, `trimRows`, `nestedRows`) could leave the selection pointing at a record that is no longer under it, and a write through that stale coordinate made `applyChanges()` append records to the source data.

`core.ts` adjusted the selection for hidden indexes only, calling `selection.commit()` under `hiddenIndexesChanged`. The asymmetry is structural rather than an oversight in the flag names: a hidden row keeps its place in the visual space, so a coordinate over it stays addressable, while a trimmed row leaves that space entirely. `commit()` recomputes only the renderable projection and never touches visual indexes, so it cannot repair a trim.

Measured in a browser with 5 rows, `selectCell(3, 0)` and then `trimRows([0, 1, 3])`, which leaves 2 visible rows while the highlight still reads visual row 3:

| write path | before |
|---|---|
| Ctrl+V | 5 records become 7, the typed value on an invented row |
| `populateFromArray` (Ctrl+Enter, autofill) | 5 records become 7 |
| `setDataAtCell(...getSelected()[0])` | 5 records become 7 |

A multi-cell range is worse: drag-select A1:A5, trim rows 2 to 4, paste once, and 5 records become 8. The paste sizes its fill loop from the range corners, so it writes down to `to.row` even while the highlight itself still addresses a live record.

Note for the reviewer: the task describes typing into the stranded cell as the trigger, and that turned out not to reproduce. `prepareEditor()` bails when the cell has no rendered `TD`, so no editor is ever prepared over a stranded coordinate and no keystroke commits. The reachable entry points are the ones that read the selection corners directly, which makes **paste the realistic user action**.

The fix drops the selection when a trimming cache update leaves it stale, on three shapes: the highlighted record is gone, the highlight is past the last row, or a range corner is past the last row. Dropping rather than clamping is deliberate: clamping moves the highlight onto a neighboring record, so the next paste silently overwrites live data, which is the same defect class with a smaller blast radius.

To ask "is the record still there", `Selection` tracks the highlight's physical coordinates. They cannot be recovered after the fact, because `IndexMapper#updateCache()` rebuilds every cache before it fires `cacheUpdated`.

Four details worth a reviewer's attention:

1. **The editor state is read twice, and which reading matters is the subtlety.** An editor still open after the public hooks was rebound onto its record and owns the selection, and `CopyPaste` refuses to paste while an editor is open, so the repair leaves it alone. An editor that was open before and is gone now was discarded because the trim took its record away, and nothing owns the selection any more, so the repair runs on the narrower half: a coordinate that addresses nothing.
2. **A row or column insert raises `trimmedIndexesChanged` exactly as a filter does.** Only the size of the physical index space tells them apart, so the repair discriminates on that count and re-reads the captured record instead of deselecting.
3. **A permutation (a sort, a row move) inverts the record test if left unhandled**, dropping a healthy selection and keeping a genuinely stranded one, so it re-reads the record too.
4. **`Filters#filter()` captures the highlighted column before it writes the trimming map**, and re-reads the live selection at the end, preferring it. Both halves are needed: a selection can arrive during the call, because `emptyDataState` restores the one it stashed when the grid emptied, and a selection can disappear during it when the Core drops it.

An axis trimmed away entirely is excluded. There the whole visual space is gone rather than the selection being stale, and selecting headers over it stays meaningful, which existing copy-with-headers behavior depends on.

One shape is deliberately left open and documented in the JSDoc: a trim **above** the highlight that leaves its coordinate in range while shifting the record out from under it. Closing it means deselecting on every trim above the selection, which is a wider behavior change than this fix takes on. It needs a follow-up task.

This is user-observable: a selection can now disappear after a trim where it previously survived out of range. It replaces silent data corruption, and there is precedent on this same path, since Filters already deselects when a filter leaves no rows. One consequence worth naming: after a drop, arrow keys do nothing until the user selects a cell again, because `transformStart`/`transformEnd` both bail when nothing is selected.

### How has this been tested?

New Playwright coverage in `tests/e2e/selection-trimmed-row.spec.ts` (14 cases). No editor takes part in most of them, so they pin selection behavior rather than the editor's. Each case asserts both halves, that the data did not grow **and** where the selection landed, because a record count alone passes under a clamp, a deselect and a follow-the-record alike.

The suite was run against the fix stashed out: exactly the deselect cases failed and the control cases passed both ways, so none of them is vacuous.

```bash
cd tests && npx playwright test                      # all six legs
npm --prefix handsontable run test:e2e               # full Jasmine suite
npm --prefix wrappers/react-wrapper run test
npm --prefix wrappers/vue3 run test
```

One existing case changed contract: `nested-rows-update-data.spec.ts` asserted that a highlight survives the collapse replay, and it now asserts the drop. The compensating `selection.refresh()` in `nestedRows` is kept, since it is dead only when the repair fires.

Unrelated to this PR, `fill-handle-formulas.spec.ts:14` fails deterministically on the horizon legs. It fails identically with this branch's source changes stashed out, so it is pre-existing on `develop` and worth its own task.

Manual: the two generated demo pages, comparing the released build against this branch. Select the cell in visual row 3, trim rows 0, 1 and 3, then press Ctrl+V. The released build grows the data from 5 records to 7; this branch leaves 5 and no selection.

### Types of changes

### Related issue(s):

1. DEV-2679

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2679

---

> [!NOTE]
> **Medium Risk**
> Touches core selection repair on every index-map cache update and filter re-selection, with a deliberate UX change (selection may clear after a trim). Risk is mitigated by extensive e2e tests and narrow deselect-vs-clamp rules, but regressions could still affect paste, filtering, and stash/restore flows.
> 
> **Overview**
> Fixes **silent row growth** when paste, Ctrl+Enter, or similar writes used a selection whose visual index no longer pointed at the same record after **Filters**, **`trimRows`**, or other **trimming index maps**.
> 
> **Core** now runs a **`repairSelection`** protocol on row/column `cacheUpdated`: capture whether an editor was open *before* public hooks, treat **structural** index-count changes and **permutations** (sort/move) as recapture-only, then on real trims call **`Selection#deselectIfHighlightStranded()`** (drop stale cell/range selections; **clamp** full-column / select-all via new **`rowExtentSpansGrid` / `columnExtentSpansGrid`** metadata in export/import). Hidden rows still go through **`commit()`** + conditional **`recaptureHighlightRecord()`**.
> 
> **Selection** tracks the highlight’s **physical** row/column at layout time and exports the grid-span flags so **dialog** / **emptyDataState** restore via spread doesn’t lose clamp behavior.
> 
> **Filters** snapshots the highlighted column **before** writing the trim map and re-selects that column afterward (with fallbacks when the core drops a stranded selection).
> 
> Adds **AGENTS.md** guidance, changelog **13310**, and broad **Playwright** coverage (`selection-trimmed-row`, `selection-trim-shapes`); **nested-rows** test expectation updates to deselect instead of an out-of-range highlight.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 81f7f7f5ec1f044eededb2f73e253d93276913c0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2679 Typing into a selection left past the last row by a trim grows the source data](https://app.clickup.com/t/9015210959/DEV-2679)

handsontable/src/core.ts
```diff
@@ -793,24 +793,140 @@ export default function Core(
 
   this.selection = selection;
 
-  const onIndexMapperCacheUpdate = ({ hiddenIndexesChanged }: { hiddenIndexesChanged: boolean }) => {
+  let hadOpenEditorOnCacheUpdate = false;
+
+  const onIndexMapperCacheUpdate = (
+    { hiddenIndexesChanged, trimmedIndexesChanged }:
+    { hiddenIndexesChanged: boolean; trimmedIndexesChanged?: boolean }
```
budnix: `trimmedIndexesChanged` is optional here but required in `repairSelection` and in both hook callbacks, for the same payload. `IndexMapper#updateCache()` always sends all three flags, so the optional form does not match what the mapper actually does - and it invites a caller to fire the local hook with a partial payload that `repairSelection` would destructure to `undefined` and silently skip the repair. One shared type across the four sites would keep them honest.

```suggestion
    { hiddenIndexesChanged: boolean; trimmedIndexesChanged: boolean }
```

handsontable/src/core.ts
```diff
@@ -793,24 +793,140 @@ export default function Core(
...
 
+    // Sampled HERE, before the public cache-update hooks run, because `EditorManager` discards a
+    // stranded editor inside those hooks - by the time the selection repair reads this, an editor
+    // that was open when the trim landed is already gone.
+    //
+    // One window is knowingly left open: a hook consumer that trims from inside
+    // `afterRowSequenceCacheUpdate` nests another cache update, which rewrites this flag before the
+    // outer repair reads it. The nested update samples the same editor state a moment later, so the
+    // two disagree only if that consumer also opened or closed an editor in between.
+    hadOpenEditorOnCacheUpdate = editorManager?.isEditorOpened() === true;
```
budnix: This one flag is shared by the row and the column `cacheUpdated` hooks, since both go through `onIndexMapperCacheUpdate`. If a column update fires while the row hooks are still running, it rewrites the flag - and by then the stranded editor is already gone, so it writes `false`. The outer row repair then runs the full record test instead of `unresolvableOnly` and drops a selection `EditorManager` deliberately left in place. The comment covers the same-axis nesting; worth covering this one too, or giving each axis its own flag.

handsontable/src/core.ts
```diff
@@ -793,24 +793,140 @@ export default function Core(
...
+
+    if (!trimmedIndexesChanged) {
+      return;
+    }
+
+    if (editorManager?.isEditorOpened()) {
+      return;
+    }
+
+    this.selection.deselectIfHighlightStranded({ unresolvableOnly: hadOpenEditorOnCacheUpdate });
```
budnix: When `deselectIfHighlightStranded()` keeps the selection, nothing re-reads the captured physical index, so it points at the old record from then on. Trim row 0 with (2,0) selected: the selection is kept (the documented "trim above the highlight" gap), visual row 2 now holds A3, but the capture still says A2 - and every later trim is judged against a record the highlight no longer addresses. Trimming A2 next drops a healthy selection; trimming A3 keeps one that a paste writes into A4. Re-reading whenever we did not deselect keeps the documented gap and stops the drift - the `editorManager?.isEditorOpened()` return four lines up needs the same.

```suggestion
    if (!this.selection.deselectIfHighlightStranded({ unresolvableOnly: hadOpenEditorOnCacheUpdate })) {
      this.selection.recaptureHighlightRecord();
    }
```

budnix (commented):
A few notes on the selection repair. The two I would look at first are the missing re-read when the selection survives a trim (`core.ts`), and the range branch in `#isAxisStranded` that looks unreachable, which would make `unresolvableOnly` a no-op. The rest are smaller.

handsontable/src/core.ts
```diff
@@ -793,24 +793,140 @@ export default function Core(
...
+  this.rowIndexMapper.addLocalHook('cacheUpdated', (indexesChangesState: {
+    hiddenIndexesChanged: boolean; trimmedIndexesChanged: boolean; indexesSequenceChanged: boolean;
+  }) => {
     onIndexMapperCacheUpdate(indexesChangesState);
 
     this.runHooks('afterRowSequenceCacheUpdate', indexesChangesState);
+
+    const indexCount = this.rowIndexMapper.getNumberOfIndexes();
+
+    repairSelection(indexCount !== lastRowIndexCount, indexesChangesState);
```
budnix: The row count is read after `afterRowSequenceCacheUpdate` has run, so whatever a hook consumer does to the row count is folded into the structural-versus-trim answer. A consumer that calls `alter('insert_row_above')` from that hook makes a plain trim look structural, `repairSelection` takes the early return, and the fix never runs. The count is already final when `cacheUpdated` fires, so reading it first makes the answer independent of consumers. Same shape in the column hook above.

```suggestion
    const indexCount = this.rowIndexMapper.getNumberOfIndexes();

    this.runHooks('afterRowSequenceCacheUpdate', indexesChangesState);

    repairSelection(indexCount !== lastRowIndexCount, indexesChangesState);
```

handsontable/src/plugins/filters/filters.ts
```diff
@@ -1037,15 +1049,20 @@ export class Filters extends BasePlugin {
...
-      }
+    // The selection is read again here, and the captured value is only a FALLBACK. Both halves
+    // earn their place. A selection can arrive during the call - `emptyDataState` restores the one
+    // it stashed when the grid emptied, which is why the state at entry cannot be the only source -
+    // and a selection can disappear during it, dropped by the Core when the trim strands it, which
+    // is why the state at exit cannot be either.
+    const currentHighlightColumn = this.hot.getSelectedRangeActive()?.highlight.col;
+    const columnToSelect = currentHighlightColumn ?? selectedHighlightColumn;
+
+    if ((this.hot.selection.isSelected() || wasSelected) && !isSelectionDropped &&
```
budnix: Curious whether the fallback should win here. A consumer that deliberately deselects during the filter (from `afterRowSequenceCacheUpdate` or `afterTrimRow`) gets overruled, because `columnToSelect` falls back to the column captured at entry - the old `isSelected()`-at-exit check honored that deselect. Separately, `(isSelected() || wasSelected)` looks fully implied by the `columnToSelect` checks: `selectedHighlightColumn` is only defined when `wasSelected` was true, and a selection that arrives during the call sets `currentHighlightColumn`, so the boolean clause never changes the outcome.

handsontable/src/selection/selection.ts
```diff
@@ -974,6 +1005,69 @@ class Selection {
...
+    if (count === 0) {
+      return false;
+    }
+
+    if (!unresolvableOnly && physicalIndex !== null &&
+        indexMapper.getVisualFromPhysicalIndex(physicalIndex) === null) {
+      return true;
+    }
+
+    return visualIndex > count - 1;
```
budnix: Can I ask - is this line reachable? `isCornerOutOfRange` above already deselects whenever `from.row` or `to.row` is past `maxRow`, and the highlight always sits between those two corners, so `visualIndex > count - 1` can only be true after that check has fired. That makes `unresolvableOnly: true` a no-op: it skips the record branch, this branch never returns true, and `isRowStranded`/`isColumnStranded` are always false. The "editor is open" case passes on the corner check alone, so if anyone later narrows that check the editor protection goes with it and nothing turns red.

tests/e2e/selection-trimmed-row.spec.ts
```diff
@@ -0,0 +1,270 @@
...
+    await grid.selectRangeWithFocusAt([0, 0, 1, 0], 0, 0);
+    // Only rows below the range are trimmed, so neither corner moves out of range.
+    await grid.trimRows([4]);
+
+    expect(await grid.selected()).toEqual([[0, 0, 1, 0]]);
+  });
+
+  test('drops every layer when the active highlight is the stranded one', async() => {
+    await grid.selectCell(0, 0);
+    await grid.selectRangeWithFocusAt([3, 0, 3, 0], 0, 0);
```
budnix: This one is named for multiple layers but only ever has one. `selectRangeWithFocusAt` calls `hot.selectCells([targetRange])`, which replaces the selection rather than adding to it, so the `selectCell(0, 0)` above is gone before the trim happens - leaving this identical to the first test in the file. It would still pass if `deselect()` were changed to clear only the active layer, so the "when it fires the whole selection goes" rule in the JSDoc has no coverage. `selectCells` takes several ranges in one call, so passing both would give it teeth.

adrianspdev:
Review findings, by priority. No critical or high issues found — the repair logic holds up under the batch, reentrancy, and editor-gating scenarios I traced. The items below are behavior sign-offs and fail-open gaps.

**Medium**

1. **Every full-column and select-all selection now drops after any row trim** — `handsontable/src/selection/selection.ts:1470`
   A full-column selection always has `to.row` at the last row, so after any trim `to.row > maxRow` holds and the corner check deselects. Consistent with the corner rule, but it is the widest user-visible change in the PR — worth an explicit sign-off and release-note visibility, not only the PR description.

2. **A batch that combines a structural change with a trim skips the stranded test entirely** — `handsontable/src/core.ts:880`
   When one cache update carries both an insert/remove and a trim (`hot.batch()` around `alter()` + `trimRows`), the index count changes, `repairSelection` takes the structural early return, and the trim is never tested. The corruption window stays open there. Fail-open to pre-existing behavior, so no regression — but it belongs in the same follow-up task as the documented trim-above-highlight gap.

3. **A stranded corner in a non-active layer of a multi-layer selection survives** — `handsontable/src/selection/selection.ts:1435`
   The repair reads the active layer only (documented in the JSDoc). A paste through a stranded non-active layer still appends records. Follow-up material.

**Low**

4. **`hadOpenEditorOnCacheUpdate` can be rewritten before the outer repair reads it** — `handsontable/src/core.ts:806`
   The flag is shared by both axes' `cacheUpdated` hooks, so a nested cache update fired from inside the public hooks rewrites it before the outer `repairSelection` reads it. The stranded editor is already discarded by then, so it writes `false` and the outer repair runs the full record test instead of `unresolvableOnly`.

5. **Hiding + trim in one update with an editor open leaves the capture stale** — `handsontable/src/core.ts:815`
   In that combination the recapture is skipped (trim present) and the repair is skipped (editor open), so `commit()`'s highlight move desyncs the captured record. The next trim is then judged against a record the highlight no longer addresses. Very narrow — needs an editor open across two trim events.

6. **The Filters fallback re-selects over a deliberate mid-filter deselect** — `handsontable/src/plugins/filters/filters.ts:1060`
   A consumer that deselects on purpose from `afterRowSequenceCacheUpdate` or `afterTrimRow` during `filter()` gets overruled, because `columnToSelect` falls back to the column captured at entry. The old `isSelected()`-at-exit check honored that deselect.

None of these blocks the merge.

handsontable/src/core.ts
```diff
@@ -793,24 +793,131 @@ export default function Core(
...
+  ) => {
+    if (!this.selection.isSelected()) {
+      return;
+    }
+
+    if (isStructuralChange || (indexesSequenceChanged && !trimmedIndexesChanged)) {
+      this.selection.recaptureHighlightRecord();
+
+      return;
+    }
```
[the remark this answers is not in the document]
demtario: Confirmed and fixed in `45e9a3a`. Reproduced in a browser first, and both directions inverted as described:

| batched case | expected | before the fix |
|---|---|---|
| sort + trim `A1` (not the highlighted record) | keep | dropped |
| sort + trim `A3` (the highlighted record) | drop | kept |

A permutation now recaptures and falls through instead of returning, so the trim test that follows judges the record the selection is actually on. Two cases added to `selection-trimmed-row.spec.ts`, verified to fail against the old gating.

handsontable/src/core.ts
demtario: Fixed in `9071d67`. All four sites now share one `IndexesChangesState` type with all three flags required, so a partial payload is a compile error rather than a silently skipped repair.

handsontable/src/core.ts
demtario: Fixed in `9071d67`. `onIndexMapperCacheUpdate()` now returns the sampled value instead of writing a shared flag, and each hook passes its own into `repairSelection()`. No shared mutable state, so a nested column update can no longer rewrite what the outer row repair reads.

handsontable/src/core.ts
demtario: Confirmed and fixed in `9071d67`. I reproduced it before changing anything, exactly as you described:

```
selectCell(2, 0)      -> highlight on A2
trimRows([0])         -> kept (the documented trim-above gap), visual 2 now holds A3,
                         but the capture still names A2
trimRows([2])         -> selection DROPPED   <- healthy selection, wrong verdict
```

Every path that leaves the selection in place now re-reads the record, including the `isEditorOpened()` return you pointed at. Covered by a new case in `selection-trimmed-row.spec.ts`, which I verified fails without the change.

handsontable/src/core.ts
demtario: Fixed in `9071d67` — the count is read before the hook runs.

Worth recording the follow-on, since it came straight out of this change: reading early but writing that pre-hook snapshot back *after* the hook rewound a nested `alter()`'s counter, so the next update looked structural and skipped the stranded test. Bugbot caught it, and `527b47cb` records the count before `runHooks()` instead.

handsontable/src/plugins/filters/filters.ts
demtario: Second half: you are right that the boolean clause was fully implied by the `columnToSelect` checks. Removed in `9071d67`, along with the now-unused `wasSelected`.

First half: kept deliberately, and documented in the code. The fallback is what restores the user's column when the Core drops a selection this filter's own trim stranded — without it, filtering with a doomed row highlighted leaves no selection at all, which is a regression against the pre-PR behavior. The cost is the one you name, so a consumer that wants a deselect to stick should do it from `afterFilter`, which runs last. Happy to revisit if you would rather honor the mid-filter deselect and accept the lost selection.

handsontable/src/selection/selection.ts
demtario: Good catch — it was a no-op, and the editor protection was riding entirely on the corner check.

Restructured in `9071d67` into two predicates: `#isAxisOutOfRange()` (applied to the highlight **and** both corners) and `#isAxisRecordGone()` (the highlight's record). `unresolvableOnly` now gates only the second, so the editor exemption is explicit and the range test can no longer be narrowed out from under it.

The highlight is still tested in its own right rather than assumed to sit inside the corners — nothing in `Selection` enforces that, and a highlight the corners do not cover is exactly the coordinate a commit writes through.

tests/e2e/selection-trimmed-row.spec.ts
demtario: Correct — `selectCells()` replaces rather than adds, so the case was a duplicate of the first test in the file. Fixed in `9071d67`: it now passes both ranges in one call and asserts both layers exist before the trim, so the "the whole selection goes" rule actually has coverage.

handsontable/src/core.ts
```diff
@@ -793,24 +793,167 @@ export default function Core(
...
 
-  this.rowIndexMapper.addLocalHook('cacheUpdated', (indexesChangesState: { hiddenIndexesChanged: boolean }) => {
-    onIndexMapperCacheUpdate(indexesChangesState);
+  this.rowIndexMapper.addLocalHook('cacheUpdated', (indexesChangesState: IndexesChangesState) => {
+    const hadOpenEditor = onIndexMapperCacheUpdate(indexesChangesState);
+    const indexCount = this.rowIndexMapper.getNumberOfIndexes();
 
     this.runHooks('afterRowSequenceCacheUpdate', indexesChangesState);
+
+    repairSelection(indexCount !== lastRowIndexCount, indexesChangesState, hadOpenEditor);
```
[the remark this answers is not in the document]
demtario: Confirmed and fixed in `527b47cb`. The count is now recorded before `runHooks()`, so a nested `alter()` that advances the counter is not rewound by the outer hook writing a stale snapshot back.

handsontable/src/selection/selection.ts
```diff
@@ -1279,6 +1384,143 @@ class Selection {
...
+      // Clamping is what such a selection already means, and it closes the append just as well as
+      // dropping would, so the user keeps the column they were working in.
+      //
+      // `isEntireColumnSelected()` cannot answer this: it compares the range height against the
+      // CURRENT row count, which the trim has already changed, so it reads false exactly here. The
+      // header anchor is structural and survives the trim.
+      if ((from.row ?? 0) < 0 || (from.col ?? 0) < 0) {
+        this.refresh();
+
+        return false;
```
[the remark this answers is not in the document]
demtario: Confirmed and fixed in `0a7256e7`. The anchor test was per-selection when it needed to be per-axis: a full-row selection is anchored in the row header, so `from.col < 0` sent a stranded **row** coordinate down the clamp path, and `refresh()` would slide it onto a surviving row for the next paste to overwrite.

Each axis is now judged by its own anchor — a row-axis overflow clamps only when the row extent is column-header-anchored (a full column or select-all), and vice versa. A full-row selection whose row a trim stranded is dropped. New case in `selection-trimmed-row.spec.ts`, verified to fail against the per-selection test.

demtario:
Thanks — this was useful, and two of the six were already fixed by the time you posted. Going through them in your order:

**1. Full-column and select-all selections dropping.** Confirmed, and changed rather than signed off. `selectColumns(0)` followed by any row trim was dropping the whole selection. A header-anchored extent tracks the grid rather than naming records, so those now **clamp** instead (`527b47cb`), which keeps the user's column and closes the append just as well.

One trap worth recording: `isEntireColumnSelected()` cannot answer this question, because it compares the range height against the *current* row count, which the trim has already changed — it reads `false` exactly when the repair runs. The header anchor is structural and survives.

Bugbot then caught that my first version of this was per-selection when it had to be per-axis: a full-row selection is anchored in the row header, so a stranded **row** coordinate was taking the clamp path and would have slid onto a neighbor. Fixed per-axis in `0a7256e7`.

**2. Batched structural change plus trim.** Agreed it is open, and I deliberately left it. Running the stranded test on the structural path is not safe: `alter()` emits its cache update *before* `selection.shiftRows()`, so the selection is legitimately out of range mid-flight and about to be repaired — testing there would deselect prematurely. That is the same window `EditorManager` guards with its stranded-in-current-task latch. Follow-up material, as you say.

**3. Non-active layer.** Real limitation, but I could not reach corruption through it, and I think the phrasing needs narrowing before it goes in the follow-up. With a stranded non-active layer and a safe active one, `CopyPaste` reads `getSelectedRangeActive()`, so the paste wrote only through the active layer and the record count was unchanged; `emptySelectedCells()` did not append either. So: the stranded layer does survive, but I have no write path that turns it into appended rows.

**4 and 5.** Both already fixed in `9071d67`, which landed just before your review. The editor flag is now returned per axis rather than shared, and every path that keeps the selection re-reads the record — including the hiding-plus-trim-with-editor-open combination you describe in 5.

**6. Filters fallback.** Same point budnix raised; kept deliberately and now documented in the code. The fallback is what restores the column when the Core drops a selection the filter's own trim stranded — without it, filtering with a doomed row highlighted leaves no selection at all, a regression against pre-PR behavior. A consumer wanting a deselect to stick should do it from `afterFilter`, which runs last.

The follow-up task therefore carries three items: the trim-above-highlight gap, the batched structural-plus-trim case, and non-active layers. And per your first point, the clamp behavior is worth a line in the release notes.

budnix (dismissed):

budnix (dismissed):

budnix (approved):

PR #13312, opened 2026-08-31
DEV-2709: Keep the trial hard-stop block under the no-ui-warns flag

budnix:
### Context

The PR fixes the `no-ui-warns` flag switching off the trial hard-stop block.

`mountBrandingSurface` returned early on `!channels.ui` *before* it reached the `LOCK_CONTENT` routing table, so the flag removed the blocking lock screen along with the badge and the bar. Measured in a real browser before the fix: a trial licence past its grace period, on a key carrying `no-ui-warns`, printed a console error and left the grid fully clickable and editable.

The flag is meant to suppress license **warnings** — that is what keeps a licensed SaaS application from showing license copy to its own end users. The hard stop is the enforcement of a licence that has stopped, not a warning, so it must survive the flag.

Because external/SaaS keys carry `no-ui-warns` by default, this meant external **trials** were unblockable — and they are the only externally-issued keys that can be blocked at all, since a lapsed subscription (`usage_hard_stop`) has no `LOCK_CONTENT` entry and `release_expired` never blocks.

**The fix:** route `LOCK_CONTENT` before the channel gate, leaving only the badge behind it.

Only `trial_hard_stop` changes in practice. `invalid` and `missing` describe keys whose flags could not be read, so both already resolved to open channels and reached the lock either way.

**Not a breaking change.** Entitlement keys ship with this release; the published 18.0.0 rejects every one of them as invalid, so no existing integration depends on this behaviour.

Documentation updated alongside the code: `handsontable/AGENTS.md`, plus the flag's two definition sites — `NO_UI_WARNS_FLAG` in `entitlementLicenseKey/constants.ts` and `LicenseChannels` in `types.ts` — both of which described the flag as closing the lock screen.

### How has this been tested?

**Unit** — `licenseBranding.unit.js`. The `no-ui-warns` block is split in two: warning states (`trial_valid`, `trial_notice`, `trial_soft_stop`) still render nothing when the UI channel is closed, and lock states now assert the lock **is** mounted. The second half fails if the `channels.ui` gate is ever moved back in front of the `LOCK_CONTENT` routing.

**E2E** — `license-branding.spec.ts`, two new cases driving a genuine trial + `no-ui-warns` key end to end, through the payload read, `resolveChannels` and the branding routing. The unit test mocks `_getLicenseState`, so it cannot see a regression in flag resolution.

- At the hard stop: the lock is still mounted.
- At the soft stop: the badge and bar are compared against an unflagged `trial` key. This is the instant where badge suppression actually bites — at the hard stop they are absent for every key, so asserting their absence there would prove nothing.

Needed a new `trial-external` fixture key.

```bash
npm run test:unit --prefix handsontable          # 329 suites, 4068 tests
npm --prefix tests test -- license-branding.spec.ts license-bar.spec.ts   # 174 tests
```

All green. The Playwright legs need the minified bundles — run `build:umd.min` and `build:styles.min` first, or the `-min` projects fail with a misleading "element(s) not found".

Also verified in Chromium against a recorded run of every license key variant: only `trial_hard_stop` + `no-ui-warns` changes, from *usable* to *blocked*. Every other state is unchanged.

### Types of changes

### Related issue(s):

1. DEV-2709

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2709

---

> [!NOTE]
> **Medium Risk**
> Changes license enforcement routing for externally issued trial keys at hard stop; behavior is intentional and narrowly scoped, but it alters whether end users can interact with the grid when the flag is set.
> 
> **Overview**
> Fixes a bug where **`no-ui-warns`** on an entitlement key hid the **trial hard-stop lock** as well as badge/bar warnings, leaving an expired external trial grid fully usable.
> 
> **`mountBrandingSurface`** now evaluates **`LOCK_CONTENT`** and mounts the lock **before** the `channels.ui` gate; only trial badge/popover paths stay behind that gate. Docs and type comments are updated to state that **`no-ui-warns` silences warning UI only**, not enforcement (DEV-2709).
> 
> Tests add a **`trial-external`** fixture key, split unit coverage for warning vs lock states under a closed UI channel, and add E2E cases for hard-stop blocking and soft-stop badge/bar suppression with the flag.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 549d307192ae87b5213f3847973859ec609b8ce3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2709 Fix two license notification defects found in the 18.1 license key audit](https://app.clickup.com/t/9015210959/DEV-2709)

tests/e2e/license-branding.spec.ts
```diff
@@ -201,6 +201,36 @@ test.describe('entitlement license key branding', () => {
...
+    // for the same rule mocks `_getLicenseState`, so it cannot see a regression in flag resolution.
+    test('still blocks when the key carries no-ui-warns', async () => {
+      await license.goto(INSTANT.trialHardStop, { key: 'trial-external' });
+
+      await expect(license.lock).toBeVisible();
+      await expect(license.lock).toContainText('Your Handsontable trial license key expired on 2026-09-26.');
+    });
+  });
+
+  test.describe('a trial issued for external use', () => {
```
demtario: Medium: this new `describe('a trial issued for external use', …)` block re-parents two pre-existing, unrelated tests — `is unaffected by the app using the Dialog plugin for its own dialogs` (line 234) and `takes the lock down with the instance` (line 247) — which both use the default `trial` key, never `trial-external`. The prior `'a trial past its grace period'` describe's closing `});` was moved up instead of down past those two tests.

Non-functional (single top-level `beforeEach`, no scoped-hook difference), but it mislabels CI failures: a red run on either test would point at the external-key path when it has nothing to do with it.

Fix: move the closing `});` of `'a trial past its grace period'` to after `takes the lock down with the instance` (currently line 255), then place this new describe block after that so it only wraps its own test.

budnix: Good spot — already fixed in `549d307192`; the external-use describe now sits after `takes the lock down with the instance` so it only wraps its own test.

demtario (approved):

Issue #13316, opened 2026-08-31
[Bug] MultiSelect cell type missing dropdown arrow icon and single-click open behavior


arjuno-ima:
### Describe the bug

The MultiSelect cell type does not render a dropdown indicator icon (arrow/caret) and does not open upon a single click. This creates an inconsistent user experience compared to native dropdown and autocomplete cell types.

### Video/Screenshots

**Multiselect Cell Type**
<img width="705" height="299" alt="Image" src="https://github.com/user-attachments/assets/52c583fa-9182-4fcd-95ba-ee8ce29c05a7" />

**Autocomplete Cell Type**
<img width="371" height="158" alt="Image" src="https://github.com/user-attachments/assets/c4f6edda-2d84-405d-8f8c-26ba9380d4a3" />

### Handsontable version

18.0.0

### Framework version

React 18

### Your environment

Mac Chrome

qunabu:
Thanks for the report! This has been logged for triage.

---
_Generated by [Claude Code](https://claude.ai/code)_

budnix:
Thanks for the report, and for the side-by-side screenshots — they made this quick to pin down.

I reproduced both parts of it, on 18.0.0 and on the development branch. You are right on both counts: a multiselect cell renders no dropdown indicator, and there is no single-click way to open its list.

One clarification on the second part. Autocomplete and dropdown cells do not open when you click the cell body either — they open when you click the small arrow at the edge of the cell. That arrow is the click target. Multiselect had no arrow, so it had no single-click path at all. So this is really one problem rather than two, and adding the indicator fixes both.

I also checked it against the original design for this cell type. The indicator was part of that design, so this is a gap in the implementation rather than a deliberate difference.

The fix is now open as #13318. Multiselect cells will show the indicator — including when the cell is empty, which today renders completely blank — and a single click on it opens the list. It will ship in an upcoming releases.

Until then, you can open the list by double-clicking the cell, or by selecting the cell and pressing <kbd>Enter</kbd>.

One note in case you style cells yourself: the multiselect indicator has its own CSS class, `ht-multi-select-arrow`. Autocomplete and dropdown cells use `htAutocompleteArrow`.

PR #11981, opened 2025-12-30
Introduce a new multi-select cell type.

jansiegel:
### Context
This PR introduces a new **MultiSelect cell type** (editor, renderer, and validator).

The MultiSelect cell type allows storing multiple values (or key/value pairs) in a single cell.

#### Overview

- ##### Editor
    The editor is a dropdown with selectable items.  
    Items can be selected by clicking them or by using the keyboard:
    - navigate with arrow keys,
    - select using `SPACE` or `ENTER` (configurable via an option).

    Each selection immediately updates the underlying cell data (unlike the other Handsontable editors).

- ##### Data format
    The cell value is always an array:
    - an array of strings, or
    - an array of objects with `key` / `value` properties,  
    depending on how the `source` property is defined in the cell meta.

- ##### Renderer
    The renderer displays the selected values as a list of closable tags (“chips”), reflecting the underlying cell data.

- ##### Validator
    The validator verifies that:
    - the cell value matches the expected data type,
    - the stored values are a subset of the values defined in the cell’s `source`.

#### Available options
*(Option names subject to change; discussion required)*

- **`allowEmpty`** – allows empty values to pass validation (default: `true`)
- **`placeholder`** – placeholder text shown in the cell renderer when the cell is empty
- **`visibleRows`** – maximum number of visible rows in the dropdown
- **`maxSelections`** – maximum number of selectable entries
- **`sourceSortFunction`** – custom sort function for dropdown entries
- **`enterCommits`** – defines whether pressing `ENTER` closes the editor
- **`searchInput`** – toggles the search input inside the dropdown
- **`filteringCaseSensitive`** – if `true`, search filtering is case sensitive

#### Still left to do

1. **CSS variables**  
   New UI elements require new CSS variables. Temporary variables are currently used, as agreed with @sl01k.

2. **Performance and refactoring (Phase 2)**  
   There are potential performance improvements to be addressed (e.g., the dropdown may re-render more often than necessary). No visible issues were observed during regular usage.  
   Unifying the autocomplete, dropdown, and select editors will likely require a small refactor.

3. **Accessibility**  
   ARIA attributes will be updated after Wojtek’s review.

#### Demo:
- https://j27yll.csb.app/ (based on `0.0.0-next-a4c3bb2-20260211`)

-------

| :exclamation:  Discrepancies between the final implementation and the original specification are documented in this gist: https://gist.github.com/jansiegel/be52e96243da30743ebda26f03eb69b0 .   |
|----------------------------------------------|

[...]

PR #13318, opened 2026-09-01
DEV-2716: Add a dropdown indicator to the multiselect cell type

budnix:
### Context

The PR fixes the `multiselect` cell type rendering no dropdown indicator and not opening its list on a single click, unlike `autocomplete`, `dropdown` and `handsontable`. Reported in #13316 on 18.0.0, reproduced on released 18.0.0 **and** on current `develop`. It is not a regression — the indicator has never existed, having shipped that way in #11981.

**The two reported symptoms are one defect.** A Handsontable list cell does not open on a single click of the *cell* — it opens on a single click of *the arrow*. Measured before the fix:

| Action | Editor state after | List opens? |
| --- | --- | --- |
| single click — `multiselect` cell (anywhere) | `STATE_VIRGIN` | no |
| double click — `multiselect` cell | `STATE_EDITING` | yes |
| single click — `dropdown` cell **body** | `STATE_VIRGIN` | no |
| single click — `dropdown` **arrow** | `STATE_EDITING` | yes |

The arrow *is* the click target, so `multiselect` having none left it with no single-click path at all. Adding the indicator resolves both symptoms together.

**Why this is a bug and not an enhancement.** Two independent design sources require the indicator:

1. The spec ([ClickUp doc `8cnjcyf-6575`](https://app.clickup.com/9015210959/v/dc/8cnjcyf-6575/8cnjcyf-7595)) never mentions an arrow in prose, but all four of its embedded design images show one — including the chips mockup the shipped renderer was built from, per the implementing PR author's [own notes](https://gist.github.com/jansiegel/be52e96243da30743ebda26f03eb69b0). Those notes enumerate every other deliberate deviation from the spec and do not list the arrow.
2. Figma ([Product-Features `360-846`](https://www.figma.com/design/JXZaX61arXLhlAi0WK3YNj/Product-Features?node-id=360-846), "↳ Multi-Select Cell Type") bakes the caret into a shared `cell / cell-type / select` component instantiated by all 17 cell mockups on the board — so it is present in all three renderer proposals that were considered, all three alignment variants, and in the open state as well.

The shipped renderer reproduces the chosen mockup faithfully — chips, `×` remove button, `+N` overflow badge — and omitted only the caret.

### What changed

- **`multiSelectRenderer`** builds a `ht-multi-select-arrow` span in **all three** render paths (chips, empty values, placeholder), inserted as the cell's first child so its float clears the content. It does not de-duplicate: every branch already clears the cell first — the two data branches through `empty()`, and the placeholder branch through `fastInnerText`, whose fast lane needs a lone text node as `firstChild` and therefore always falls through to `empty()` while an indicator sits there. A unit case seeds a stale indicator into a reused `TD` and asserts the fresh one is the survivor, in each of the three branches.
- **Single-click open** via one delegated `mousedown` listener per instance. It resolves the cell from the indicator's own `data-row`/`data-col`, not from a closure captured at render time — `autocompleteRenderer` captures `row`/`col` from whichever cell rendered first, and that is one reason its class is not reused here. Left button only, matching the `htAutocompleteArrow` path (DEV-2677), and it calls `onCellDblClick` with a `null` event so the list still opens after a shift-drag range. An `afterDestroy` hook tears the listener down, matching what `autocompleteRenderer` pairs with the same listener.
- **A separate class**, not `htAutocompleteArrow`: that class carries autocomplete's own styling *and* its global listener, so reusing it would route multiselect clicks into autocomplete's stale-coordinate handler whenever a grid has both column types.
- **Icon** added to the `selectArrow` group in the committed codegen input (`scripts/themes/figma/templates/iconsMap.ts` plus its runtime mirror `utils/helpers/iconsMap.mjs`) and mirrored into the six generated files under `src/themes/static/`, so a later `npm run generate:themes` reproduces it rather than dropping it. `scripts/themes/figma/__tests__/iconsMap.test.mjs` asserts the template and the mirror stay identical.
- **Chips overflow** now subtracts the indicator's measured outer width, so the last visible chip and the `+N` badge stop short of it. This narrows the overlap band rather than closing it: the budget is derived from `getColWidth`, a border-box width, and the cell's own horizontal padding is still not subtracted — a pre-existing gap called out in a code comment. Closing it means measuring the container's content box instead, which changes how many chips every multiselect cell shows, so it is deliberately left out of this PR.
- **RTL** handled the way `_autocomplete-renderer.scss` does it, with a `[dir="rtl"]` float override.

### One unrelated test file had to change

`tests/fixtures/pages/EditorHiddenCellPage.ts` selected cells with a centred `cell.click()`. That broke on all six Playwright legs once the indicator existed, and the cause is worth recording because it is not the arrow:

| | before | after |
| --- | --- | --- |
| auto-sized column width | 68px | **93px** (+25 = 16px icon + 8px margin + 1px) |
| cell midpoint | x=92 | **x=105** |
| chip's `×` button spans | x=96–112 | x=96–112 |

The indicator is measured by `autoColumnSize` — correctly, since it occupies real space; not measuring it would let chips render under the arrow in auto-sized columns. That widened the column by 25px and moved the midpoint 13px right, onto the chip's `×`. That `×` suppresses selection by design (`beforeOnCellMouseDown` + `stopImmediatePropagation`), so `hot.getSelected()` stayed `undefined` and the page object's poll timed out with no editor and no selection to explain it. The spec had been passing by 4 pixels.

The page object now clicks 4px from the cell's leading edge, which is padding in every cell type. `tests/AGENTS.md` already warned against centred clicks on list cells; it now also records the chip-`×` hazard and this measurement, since the existing note only covered the arrow.

### Also in this branch

- **Pre-commit hook fix.** Staging `handsontable/scripts/themes/figma/templates/iconsMap.ts` made `scripts/lint-staged.mjs` block: the file sits inside the hook's `handsontable/scripts/` scope but `handsontable/.eslintignore` excludes it, and the hook runs ESLint from the repo root where that ignore does not apply. ESLint then answers with a *parsing* error, which is exit 1 (a lint error) rather than the exit 2 config gap `runEslint` already tolerates. `lintable()` now drops package-ignored paths — every `handsontable/.eslintignore` entry that falls inside a hook scope, so the vendored `test/lib/jquery*.js` and the built `test/dist` bundles cannot block a commit either. This is the invariant `scripts/lint-files.mjs` already documents for itself: "a strict SUBSET of CI … never a false block on files CI deliberately does not lint". Covered in `scripts/__tests__/lint-files.test.mjs`.

### How has this been tested?

Unit — new cases in the existing renderer suite covering all three render paths, the coordinate dataset, `aria-hidden` under `ariaTags`, and a stale indicator not surviving a re-render of a reused `TD` in each branch. Two pre-existing stubs in that file needed `rootElement` and `addHookOnce`, which the renderer now touches; real instances have both.

```bash
npm run test:unit --prefix handsontable                       # 4144 passed, 331 suites
```

E2E — `multiselect` added as a fourth cell type to the existing dropdown-arrow spec, so the affordance every list cell type shares has one home. Two new cases apply to all four types: a click on the cell **body** must not open the editor (which would otherwise let a change that made the whole cell a click target pass every other case), and the indicator must render in an **empty** cell.

```bash
cd tests
npx playwright test e2e/cell-dropdown-arrow-button.spec.ts    # 96 passed (all six legs)
npx playwright test e2e/editor-hidden-cell.spec.ts            # 78 passed (all six legs)
npx playwright test --project=e2e-main                        # 563 passed, 1 skipped
```

Legacy Jasmine suite (no new specs added there — it is frozen):

```bash
npm run test:e2e --prefix handsontable -- --testPathPattern=multiSelect   # 127 specs, 0 failures
```

Codegen and tooling:

```bash
cd handsontable && node --test scripts/themes/figma/__tests__/*.test.mjs  # 16 passed
npm run test:tooling                                                     # 146 passed
```

Full local gate: `eslint` 0 errors, `stylelint` clean, `tests` package lint clean, `test:types` clean.

Manual — verified in Chrome against a released-18.0.0 vs local-build harness. Before: multiselect 0 indicator elements, dropdown and autocomplete 4 each. After: 4 each across all three columns, the indicator present on the empty row, no collision with the `+2` badge, and clicking row 3's indicator opens the editor at `[3, 0]` rather than the first-rendered cell.

### Known follow-ups, deliberately not in this PR

1. The chips overflow budget ignores the cell's horizontal padding (see above) — pre-existing, and this PR narrows rather than widens the band.
2. The indicator duplicates `autocompleteRenderer`'s arrow in both SCSS and listener mechanics. A shared helper would remove the duplication and also fix autocomplete's closure-captured `row`/`col`, but that changes decade-old behavior on three other cell types.
3. `getDropdownIndicatorReserve` reads `getComputedStyle` once per multiselect cell per draw. Worth caching per instance; note `recalculateChipsVisibility` already did one such read per cell before this change.

### Types of changes

### Related issue(s):

1. #13316
2. DEV-2716

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

The multiselect guide gains an "Opening the dropdown" section, which names both indicator classes rather than implying the two cell families share one element. `tests/AGENTS.md` records that `multiselect` now carries an indicator too, that it is a different element and listener from `htAutocompleteArrow`, and the chip-`×` hazard above.

Changelog entry: `.changelogs/13316.json` (already in this branch — it is keyed by the public issue number, so it did not need the PR number).

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2716 — supersedes the auto-triaged duplicate DEV-2715, now closed.

---

> [!NOTE]
> **Low Risk**
> Scoped to multiselect rendering, styling, and tests; behavior of other cell types is unchanged aside from safer E2E cell clicks.
> 
> **Overview**
> Fixes **#13316** by giving `multiselect` cells the same list-cell affordance as autocomplete and dropdown: a trailing **`ht-multi-select-arrow`** indicator (not `htAutocompleteArrow`, so mixed column types do not share autocomplete’s listener) on every render path—chips, empty, and placeholder.
> 
> A **delegated left-click `mousedown`** on that arrow opens the editor via `onCellDblClick`, using `data-row`/`data-col` on the indicator and `closest('td')` so frozen overlay clones work; chip overflow layout **reserves the arrow’s width** so chips and `+N` stay inside the cell. Theme/icon maps and SCSS mirror autocomplete styling, including RTL.
> 
> Docs add an **“Opening the dropdown”** section; E2E coverage extends the dropdown-arrow suite to multiselect (body click must not open, empty cells, frozen column, overflow sweep). **`EditorHiddenCellPage`** selects via leading-edge clicks so auto-sized columns are not thrown off by the new indicator/chip `×`. Pre-commit **`lintable()`** skips paths in `handsontable/.eslintignore` so figma template stencils do not block commits.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0e806309014d88090ccf7732764df0ef5fca7712. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2716 [GH-13316] MultiSelect cell type is missing the dropdown arrow icon and single-click-to-open behavior](https://app.clickup.com/t/9015210959/DEV-2716)
Task linked: [DEV-2677 Autocomplete/dropdown: clicking an empty cell then pressing Enter closes the editor on the same keypress](https://app.clickup.com/t/9015210959/DEV-2677)

qunabu:
Task linked: [DEV-2715 [GH-13316] MultiSelect cell type is missing the dropdown arrow icon and single-click-to-open behavior](https://app.clickup.com/t/9015210959/DEV-2715)

handsontable/src/renderers/multiSelectRenderer/utils/utils.ts
```diff
@@ -201,7 +317,14 @@ function recalculateChipsVisibility(
...
 ): void {
-  const containerWidth = columnWidth;
+  // Subtract the indicator so the last visible chip and the `+N` badge stop short of it. This
+  // narrows the overlap band rather than closing it: `columnWidth` is a border-box width from
+  // `getColWidth`, and the cell's own horizontal padding is still not subtracted — a pre-existing
+  // gap. Closing it means measuring the container's content box instead of deriving the budget from
+  // the column width, which changes how many chips every multiselect cell shows.
+  const containerWidth = columnWidth === null
+    ? null
+    : columnWidth - getDropdownIndicatorReserve(chipsContainer, rootDocument);
```
demtario, line 328: **medium**: The new overflow-reserve math (`getDropdownIndicatorReserve` / its use in `recalculateChipsVisibility`) has no test coverage. No unit test calls `handleChipsOverflow` (and jsdom's `offsetWidth`/`getComputedStyle` return 0 there anyway), and the fixture that reaches multiselect (`tests/fixtures/demo/cell-dropdown-arrow.html`) only ever seeds a single chip, so no scenario reaches the overflow branch.

Fix: add a Playwright E2E case with enough chips to overflow a narrow multiselect column, asserting the last visible chip and the `+N` badge stop clear of the indicator — the scenario this calculation exists to fix.

handsontable/src/renderers/multiSelectRenderer/utils/utils.ts
```diff
@@ -193,6 +285,30 @@ export function cacheColumnWidthAndRegisterResizeHook(
...
+ *
+ * Both margins are summed, so the physical `marginLeft`/`marginRight` pair works for LTR and RTL
+ * alike and does not depend on logical-property support.
+ *
+ * @param {HTMLElement} chipsContainer The chips container whose cell holds the indicator.
+ * @param {Document} rootDocument The document that owns the grid.
+ * @returns {number} The reserved width in pixels, or `0` when no indicator is rendered.
+ */
+function getDropdownIndicatorReserve(chipsContainer: HTMLElement, rootDocument: Document): number {
+  const indicator = chipsContainer.parentElement?.querySelector<HTMLElement>(`.${ARROW_CLASS}`);
```
demtario: **low**: `chipsContainer.parentElement?.querySelector(...)` uses optional chaining on a value that's guaranteed non-null by the call contract — `chipsContainer` is always appended to the `TD` before this runs.

Fix: `chipsContainer.parentElement!.querySelector(...)`.

handsontable/src/renderers/multiSelectRenderer/utils/utils.ts
```diff
@@ -101,6 +103,96 @@ export function createOverflowIndicator(rootDocument: Document, count: number):
...
+
+    // The `null` event is load-bearing, not laziness: `EditorManager#openEditor` only applies its
+    // "no editor for a multi-cell selection" default when the event is a `MouseEvent`, so forwarding
+    // the real one would stop the indicator from opening the list after a shift-drag range.
+    hotInstance.view._wt.getSetting(
+      'onCellDblClick',
+      null,
+      hotInstance._createCellCoords(visualRow, visualColumn),
+      td
+    );
```
demtario, line 192: **low, informational**: `hotInstance.view._wt.getSetting(...)` plus `hotInstance._createCellCoords(...)` reaches through private members / a deep property chain. Not a new issue though — it's a verbatim copy of the existing call shape in `autocompleteRenderer.ts` (line 69). Flagging for awareness/consistency, not a blocker.

demtario (commented):
3 inline notes below.

handsontable/src/renderers/multiSelectRenderer/utils/utils.ts
budnix: Good spot — nothing covered it, so 0e80630901 adds a width-sweep case (a single width has slack and passes either way). Worth knowing: asserting clearance of the indicator actually fails today because the budget never subtracts the cell padding, so the case asserts the cell box instead and I will raise the rest separately.

handsontable/src/renderers/multiSelectRenderer/utils/utils.ts
budnix: You are right — switched to `parentElement!` in 0e80630901 and wrote the call contract into the JSDoc.

handsontable/src/renderers/multiSelectRenderer/utils/utils.ts
budnix: Agreed — left it matching `autocompleteRenderer` on purpose so the two stay easy to diff; giving it a real API would touch the other three cell types, so I would rather do that separately.

demtario (approved):

PR #13319, opened 2026-09-01
DEV-2708: Fix removeChild error on context menu over a resize handle

demtario:
### Context

The PR fixes an uncaught `NotFoundError` thrown out of the `ManualRowResize` and `ManualColumnResize` context-menu handlers:

```
Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
```

It was reported by Sentry from the demos site, with the failing frame inside the plugin's own handler and no page code in the stack.

Both plugins detach their handle and their guide from the root element when a `contextmenu` lands on the handle, but the two elements are attached at different moments:

| Element | Attached in | When |
|---|---|---|
| handle | `setupHandlePosition()` | `mouseover` over a header |
| guide | `setupGuidePosition()` | only once a `mousedown` over the handle reaches the plugin's own root-element listener |

`hideHandleAndGuide()` only toggles the `active` class, and the context-menu handler is the only place in the core that ever detaches either element. So after a hover-only interaction the guide has never been a child of the root element, and the second `removeChild` throws. The throw also skips the rest of the handler, which is what left the handle re-appearing right after the menu opened.

A plain desktop right-click never hit it, because its own `mousedown` attaches the guide first, which is exactly what the existing legacy spec simulates. Two paths do reach the handler with no `mousedown` handled in between, and both are reproduced in the new tests:

1. host-page code that stops `mousedown` from reaching the grid (an overlay, a drag library, a sandboxed editor),
2. a synthetic `contextmenu` (an automation harness, a long-press shim, host code opening its own menu).

The fix detaches both elements with `remove()`, which is a no-op on an element that has no parent. It is preferred over a parentage check because it also covers a detached handle, which a guide-only guard would miss.

Note on the original hypothesis in the task: the handle is **not** detached by a re-render or a row removal. Nothing in the core detaches it other than this handler, so the fix does not belong in teardown.

### How has this been tested?

New Playwright spec, verified **red before the fix** with the exact error message above and green after:

```bash
cd tests && npx playwright test e2e/manual-resize-contextmenu.spec.ts
```

Five cases: a synthetic context menu over a hovered row handle and over a hovered column handle, a real right-click on each handle while the host page swallows `mousedown`, and a control case asserting that a real resize drag still gets both elements detached. The assertion is the absence of an uncaught page error (`page.on('pageerror')`); an exception raised inside a listener never reaches whoever dispatched the event, so a `not.toThrow()` around the event would be green on the unfixed code.

Regression runs, all green:

```bash
npm run test:e2e --prefix handsontable -- --testPathPattern='manual(Row|Column)Resize'   # 147 specs
npm run test:e2e --prefix handsontable -- --testPathPattern='contextMenu'                # 596 specs
```

The new spec was also run across all six theme x bundle legs, repeated, to rule out flake.

### Types of changes

### Related issue(s):

1. DEV-2708

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2708

---

> [!NOTE]
> **Low Risk**
> Narrow DOM teardown change in resize plugins’ context-menu path, with targeted e2e regression coverage and no auth or data-handling impact.
> 
> **Overview**
> Fixes an uncaught DOM error when opening a context menu on a **row or column resize handle** that was only hovered, not dragged. In `ManualRowResize` and `ManualColumnResize`, `#onContextMenu` used to call `rootElement.removeChild` on both the handle and guide; the **guide is only appended after `mousedown`**, so hover-then-context-menu could throw and abort the rest of the handler.
> 
> Both plugins now detach handle and guide with **`Element.remove()`**, which is safe when the node was never attached. A changelog entry documents the user-facing fix.
> 
> **Tests:** New Playwright coverage (`manual-resize-contextmenu.spec.ts`) with a dedicated fixture and page object—synthetic `contextmenu`, right-click with swallowed `mousedown`, and post-drag cases—asserting no uncaught `pageerror`. Shared `dragResizeHandle` in `tests/fixtures/gestures.ts` replaces duplicated drag logic in `ManualResizeSizeOptionResetPage`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7a4b3d1fcf6c797d788c164bbc1185269fd39c00. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2708 Handsontable core: removeChild throws in ManualRowResize.onContextMenu2 (Sentry DEMOS-2W)](https://app.clickup.com/t/9015210959/DEV-2708)

handsontable/src/plugins/manualRowResize/manualRowResize.ts
```diff
@@ -783,8 +783,12 @@ export class ManualRowResize extends BasePlugin {
...
   #onContextMenu() {
     this.hideHandleAndGuide();
-    this.hot.rootElement.removeChild(this.#handle);
-    this.hot.rootElement.removeChild(this.#guide);
+    // Both elements are detached with `remove()`, which is a no-op on an element that has no
+    // parent. The guide is attached only once a "mousedown" over the handle reaches `#onMouseDown`,
+    // so a context menu opened over a merely hovered handle reaches a guide that was never
+    // attached, and `removeChild` threw there (DEV-2708).
+    this.#handle.remove();
+    this.#guide.remove();
```
budnix, line 791: The `remove()` swap kills the crash cleanly. One thing I noticed reading around it: `hideHandleAndGuide()` only strips the `active` class and `#onMouseUp` never detaches, so after a normal drag both divs sit in the root element for the life of the grid — the new spec pins that as its premise with `toHaveCount(1)`. `disablePlugin()` leaves them behind too. Was moving the detach into `hideHandleAndGuide()`, so mouseup and disable share it, looked at and left out of scope? Same applies to `manualColumnResize.ts`.

tests/fixtures/demo/manual-resize-contextmenu.html
```diff
@@ -0,0 +1,76 @@
...
+    const container = document.querySelector('[data-testid="grid"]');
+
+    container.className = `ht-theme-${window.htTheme}`;
+
+    window.grid = new Handsontable(container, {
+      data: Handsontable.helper.createSpreadsheetData(5, 5),
+      colHeaders: true,
+      rowHeaders: true,
+      manualRowResize: true,
+      manualColumnResize: true,
```
budnix, line 67: The fixture turns on both resize plugins but not `contextMenu`, so the ContextMenu plugin never runs in any of the six tests. The changelog says "an error thrown when a context menu was opened over a row or column resize handle", and the Sentry demos have it on — with it enabled, a second `contextmenu` listener runs after the handle is already detached, and the menu re-renders the grid on top of it. Worth enabling here so the tests cover the shape users actually hit?

```suggestion
      manualColumnResize: true,
      contextMenu: true,
```

tests/e2e/manual-resize-contextmenu.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+    await grid.hoverRowHeader(2);
+
+    await grid.rightClick(grid.rowHandle);
+
+    // With the `mousedown` swallowed the guide was never attached, which is what used to throw.
+    // Whether the handle stays detached afterwards is not asserted: the right-click leaves the
+    // pointer ON the handle, so detaching it changes the element under the cursor and the browser
+    // sends a fresh "mouseover" that re-reveals it. Whether that lands before or after the
+    // suppression microtask is up to the browser, and no native menu covers the grid here.
+    await expect(grid.rowGuide).toHaveCount(0);
```
budnix, line 88: This `toHaveCount(0)` is true on both the fixed and the unfixed code — the guide is never attached on this path — so `pageErrors` is doing all the detecting in these two tests. That works today, but it rests on `mouse.down/up({ button: 'right' })` producing a real `contextmenu` event, which nothing here checks. If that ever stops happening these two stay green while covering nothing, and they are the only tests for the actual Sentry path. Could you add one positive signal, like a `contextmenu` counter on the fixture root asserted above zero?

tests/e2e/manual-resize-contextmenu.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+ * inside a listener never reaches whoever dispatched the event, so a `not.toThrow()` around the
+ * event would be green on the unfixed code.
+ */
+test.describe('Manual resize handle and the context menu', () => {
+  let grid: ManualResizeContextMenuPage;
+  let pageErrors: string[];
+
+  test.beforeEach(async ({ page, theme, bundle }) => {
+    pageErrors = [];
+    page.on('pageerror', error => pageErrors.push(error.message));
```
budnix, line 30: The listener goes in here, but every `expect(pageErrors).toEqual([])` sits mid-test instead of at the end — in the first two tests there is a re-hover after the check and nothing looks at errors again. `#onMouseDown` also arms a 500 ms autoresize timeout, so the drag tests finish with that still pending. A `test.afterEach` check would cover the whole body including the tail, and it drops six copies of the same line. Curious whether inline was on purpose.

tests/fixtures/pages/ManualResizeContextMenuPage.ts
```diff
@@ -0,0 +1,211 @@
...
+  async parkPointer(): Promise<void> {
+    await this.page.mouse.move(0, 0);
+  }
+
+  /**
+   * Right-clicks an element with a real pointer, the way a user opens a context menu.
+   *
+   * @param {Locator} target The element to right-click.
+   */
+  async rightClick(target: Locator): Promise<void> {
```
budnix, line 134: This is Playwright's own right-click, minus its actionability and hit-target checks. Every other page object in the suite uses the built-in (`GridPage.ts:88`, `MenuScrollPage.ts:94`, and four more), which is also decent evidence that the built-in really does fire `contextmenu` here. The difference that matters: with no hit-target check, a handle covered by another element on one theme leg gets pressed through the cover instead of failing loudly, so the test goes quiet rather than red. The sibling page object hand-rolls *drags* because Playwright has no drag primitive, but a right-click is not that case — the whole body becomes `await target.click({ button: 'right' });`.

tests/fixtures/pages/ManualResizeContextMenuPage.ts
```diff
@@ -0,0 +1,211 @@
...
+    await expect(this.columnHandle).toBeVisible();
+  }
+
+  /**
+   * Resizes a row by dragging its resize handle, which is what attaches the resize guide.
+   *
+   * @param {number} row The visual row index.
+   * @param {number} deltaY How far to drag, in CSS pixels.
+   */
+  async dragRowHandle(row: number, deltaY: number): Promise<void> {
```
budnix, line 101: `dragRowHandle` and `dragColumnHandle` are near-copies of the same two methods in `ManualResizeSizeOptionResetPage.ts:64` and `:98` — same hover, same `boundingBox()` null check, same "The row resize handle has no layout box." message, same "Several moves, because..." comment. A later fix to the gesture then has to land in two files, and the copy that gets missed keeps passing until the geometry shifts. Would it be worth pulling the gesture into a shared helper under `tests/fixtures/`? The two page objects differ only in how they find the grid container.

budnix (commented):
A few notes from a close read of the fix and the new tests.

handsontable/src/plugins/manualRowResize/manualRowResize.ts
demtario: Yes, looked at and deliberately left out of scope, and you are right about the extent of it: after any normal drag both divs stay in the root element for the life of the grid, and `disablePlugin()` and `destroy()` leave them too.

Sharing the detach through `hideHandleAndGuide()` is not a drop-in change, which is why it is not in this PR:

- `#onMouseUp` re-runs `setupHandlePosition()` right after, which re-attaches the handle but never the guide.
- `setupHandlePosition()` early-returns on several paths (`shouldSkipResizeHandlePositioning`, a header with no resolvable coordinates, a null current row), so a shared detach can leave the handle detached mid-drag where today it stays attached.
- `setupGuidePosition()` reads `outerWidth()` and the inline style off the handle, which measure zero while it is detached, so the handle-first call order in `#onMouseDown` becomes load-bearing rather than incidental.

That is a behavior change on lines this diff does not touch, and it needs its own coverage plus the same treatment in `manualColumnResize.ts`. Filed as a separate ticket, with the constraints above written down and `destroy()` included in its scope.

tests/fixtures/demo/manual-resize-contextmenu.html
demtario: Good catch, applied. The fixture now sets `contextMenu: true`, so every test runs with the second `contextmenu` listener the reporting grids have, and the comment above the option says why it is there.

Green on all six theme x bundle legs with it on, repeated, so nothing in the fix depended on the plugin being absent.

tests/e2e/manual-resize-contextmenu.spec.ts
demtario: Agreed, and this was the sharpest of the notes. On this path `toHaveCount(0)` on the guide is true before and after the fix, so the error check was the only detector, resting on an assumption nothing verified.

The fixture now counts every `contextmenu` that reaches the document (capture phase, no `preventDefault`, no `stopPropagation`), the page object exposes the count, and both real-gesture tests assert it above zero.

Verified the new assertion is not itself vacuous: swapping the gesture to a plain left click turns exactly those two tests red and leaves the other four green.

tests/e2e/manual-resize-contextmenu.spec.ts
demtario: Not on purpose, and your reasoning is better than mine was. Moved to a `test.afterEach`, which drops the six duplicated lines and covers the tail of every test: the re-hover the first two end on, and the 500 ms autoresize timeout the drag tests leave armed.

tests/fixtures/pages/ManualResizeContextMenuPage.ts
demtario: Applied, the whole body is now `await target.click({ button: 'right' });`.

Accepted for the failure mode you name rather than for the precedent: without the hit-target check, a handle covered on one theme leg is pressed through the cover and the test goes quiet instead of red, which is the same class of problem as the note above about a test that stays green covering nothing. Your point about the sibling page object hand-rolling drags only because Playwright has no drag primitive is right, and that distinction is now recorded in the shared helper.

tests/fixtures/pages/ManualResizeContextMenuPage.ts
demtario: Extracted. The gesture now lives in `tests/fixtures/gestures.ts` as `dragResizeHandle(page, handle, { x?, y? })`, and both page objects call it.

The split follows the difference you identified: how a grid and its header are addressed stays in each page object, which reveals the handle and passes it in, while the `boundingBox()` null check, the midpoint and the multi-step press exist once. The reason the drag is spelled out at all, that Playwright has no drag primitive, is written down there too, so the next reader does not replace it with a built-in that does not exist.

Both affected specs re-run across all six legs: 138 passed.

budnix (approved):

Issue #4980, opened 2018-03-30
fragmentSelection not working for fixed columns with editor: false


CorneH:
When a handsontable has fixedColumnsLeft and has set fragmentSelection to true, the setting works correctly for all columns, except for the fixedColumns.

Some further investigation learns that the difference between fixed and not-fixed column is in the "isChildOfTableBody"-variable in TableView.prototype.isTextSelectionAllowed.

AMBudnik:
Thanks for sharing @CorneH 

You're right. We are not able to select a fragment of cell's content when the cell is a part of the fixed area.

Here's a demo for test purposes http://jsfiddle.net/handsoncode/hakt2fhd/

![gif](https://user-images.githubusercontent.com/10757813/38135286-a6804bda-3417-11e8-917d-97c9657f0e70.gif)

AMBudnik:
Please update to [v `5.0.0`](http://jsfiddle.net/gx1u4a0s/).

![sgif](https://user-images.githubusercontent.com/10757813/44987972-5ff3e580-af89-11e8-9e1d-3be229bd0a4c.gif)

swk777:
it's still not working when editor is false

AMBudnik:
Thank you for the suggestion @swk777 

I have replicated the issue http://jsfiddle.net/handsoncode/1twy0q3e/

AMBudnik:
Still replicable in `7.2.1,` http://jsfiddle.net/vL1cbgxq/

pnowak:
This issue is still replicable using v8.0.0. Here's an updated demo - https://jsfiddle.net/w6Lnk5gp/

xiaodong-zhu:
in addition to set editor: false, fragmentSelection: true, disableVisualSelection: true
**ALSO ADD** a CSS rule:
textarea.HandsontableCopyPaste {
    display: none !important;
}

by doing this, .HandsontableCopyPaste does not get focused and thus the browser native selection remains

xiaodong-zhu:
In addition to set editor: false, disableVisualSelection: true, fragmentSelection: true
ALSO ADD a CSS rule:
textarea.HandsontableCopyPaste {
    display: none !important;
}

by doing this, textarea.HandsontableCopyPaste does not get focused and thus the browser native selection remains

dtomaszewski:
Hi @AMBudnik any news here ?

AMBudnik:
Hi @dtomaszewski this issue hasn't been listed for the next milestone. 
[...]

PR #13320, opened 2026-09-01
DEV-2720: Allow fragmentSelection in frozen rows and columns (#4980)

budnix:
### Context

The PR fixes `fragmentSelection` inside frozen rows and columns.

With `fragmentSelection` enabled, you cannot select a fragment of a cell's text anywhere in a frozen area — a frozen column (`fixedColumnsStart`), a frozen row (`fixedRowsTop`/`fixedRowsBottom`), or either start corner. It affects both `fragmentSelection: true` and `fragmentSelection: 'cell'`.

Two things about the issue as filed are worth correcting:

- **`editor: false` is irrelevant.** The issue title blames it. Tested with and without it — identical result either way.
- **It was never fixed.** The thread says it was fixed in 5.0.0. That could not be reproduced: 5.0.0 and 12.0.0 both fail with a passing control in the same run, and `git log -S isChildOfTableBody` shows the responsible line unchanged since 2015. No spec in `fragmentSelection.spec.js` has ever set a `fixed*` option, which is why it stayed silent.

Reproduced on `develop` and on published 18.0.0.

#### Cause

`TableView.isTextSelectionAllowed` tested membership of the **master** table only:

```ts
const isChildOfTableBody = isChildOf(el, this._wt.wtTable.spreader);
```

A frozen cell you see and click is not in the master table — it is rendered in an overlay clone that is a *sibling* of the master, not a descendant. So this was always `false` for a frozen cell, and every `fragmentSelection` branch is gated behind it. The `mousedown` handler then called `clearTextSelection()` and `preventDefault()`, so the gesture never started.

Confirmed by forcing the guard open on a fresh page and nothing else — the frozen cell then selected normally. No CSS involved (`user-select` is `auto` on both the clone and master cells).

#### Fix

Two parts, both in `handsontable/src/tableView.ts`.

**1. Resolve the table that owns the element, then test its selectable area.**

```ts
#isSelectableTableArea(el) {
  const { spreader } = this.#getOwningWt(el).wtTable;

  if (!isChildOf(el, spreader)) {
    return false;
  }

  return closest(el, ['TH'], spreader) === null;
}
```

`#getOwningWt` wraps `wtOverlays.getParentOverlay(el) ?? this._wt` — the idiom this file already used in two other places, now shared by all three. `activeWt` was the other candidate and was rejected: the root-element `mousedown` listener can run before Walkontable swaps it, so it is not reliable here.

The rule is "anything the owning table renders, except a header". Two details earn their place, both found by review:

- **It cannot be just the cells.** The selection borders sit in the spreader beside the table, and a multi-cell drag lands a `mousemove` on one. An earlier version of this fix tested for a `TD` and so rejected them, which cancelled the gesture: a drag across three cells in the master selected **0 characters where 18.0.0 selects 82** — that is `fragmentSelection: true`'s headline behavior, "text selection in multiple cells at a time", broken. Allowing the whole spreader keeps borders working exactly as before.
- **Headers must still be excluded, and `isChildOf` has to come first.** Column headers sit in the `THEAD` and row headers are `TH` elements inside the `TBODY`'s own rows. Since `shouldRenderTopOverlay`/`shouldRenderInlineStartOverlay` are true whenever headers exist at all, allowing them made **header labels selectable on any grid with headers — no frozen rows or columns needed**. Measured: on a grid with only `colHeaders` + `rowHeaders`, dragging a column header selected 26 characters where 18.0.0 selected none. The `isChildOf` check runs first because it is what bounds the `closest` walk — `closest` runs past an `until` argument that is not an ancestor, which would leave the grid entirely and match a `TD` on the host page, making a header selectable whenever the grid sits inside a table cell.

**2. Confine a text selection to the overlay it started in.**

Part 1 on its own made a boundary-crossing drag *worse*. Instead of nothing, a leftward drag out of the master into the frozen columns produced a 1716-character selection containing `R1C3`, `R1C4` and further columns — cells to the *right* of a leftward drag — because the overlay tables sit next to each other in the DOM in an order that does not follow the visual layout. So `mousedown` records the overlay it started in and `mousemove` treats leaving that overlay as disallowed.

The confinement is gated on `fragmentSelection` being set and skipped for an input, so the editor/IME path is untouched: the editor's textarea lives outside every clone, and without that exemption a drag reaching it would cancel a gesture `isTextSelectionAllowed` explicitly permits via its `isInput` branch.

The drag state is also cleared on the `documentElement` mouseup, not only the `rootElement` one. A release outside the grid never reaches the latter, and stale state plus the new confinement would let a plain hover over the grid wipe whatever the user has selected on the host page.

#### What this deliberately does not change

A drag that starts in a normal cell and crosses into a frozen area still ends with no selection. That is unchanged from published 18.0.0 — verified side by side, both grow to 23 characters in the master and drop to 0 on entering the frozen area. A single native range cannot meaningfully span the master table and an overlay clone, so making that gesture select the visually-swept cells needs a selection layer that does not exist today. Left as it was rather than replaced with a wrong answer.

This is now deliberate behavior, so the `fragmentSelection` JSDoc (`metaSchema.ts`, which feeds the API reference) states both that the selection is *cleared* when the pointer leaves its area — not clamped — and that headers are not selectable. The `selection` guide only links the option and carries no prose about it, so nothing there needed changing.

### How has this been tested?

New Playwright spec — 13 tests, run across all 3 themes and both bundles (78 legs):

```bash
cd tests && npx playwright test e2e/fragment-selection-frozen.spec.ts
```

Coverage: all five frozen areas `getParentOverlay` can resolve (frozen columns, frozen top rows, frozen bottom rows, and both start corners), a master-table control, `fragmentSelection: 'cell'`, a `fragmentSelection: false` guard, the boundary-crossing clear, a companion case that stops short of the seam, a multi-cell drag in the master and inside the frozen columns, and a header guard.

**Verified the tests actually catch the bugs** — three builds, three outcomes:

| Build | Result |
|---|---|
| `develop` (unfixed) | 7 failed / 6 passed — every frozen area red, all controls and guards green |
| First attempt, spreader with no header exclusion | 1 failed — the header guard, exactly the case it exists for |
| Second attempt, `TD`-only | 7 passed / 6 failed against the frozen areas' own multi-cell case |
| Final | 13 passed |

Regression runs on the final build:

```bash
npm --prefix handsontable run test:unit    # 4199 tests, 336 suites
npm --prefix handsontable run test:e2e -- --testPathPattern='(selection|copyPaste|fragmentSelection|Editor|editors|Header|headers)'  # 2699 specs
npm --prefix tests run lint
```

All green (`2699 specs, 0 failures`). One earlier run of that legacy sweep reported 19 failures; identical re-runs on the same build reported 0, and the header specs alone report 0 of 494 — the legacy suite's known environmental flakiness rather than a regression.

**One gap worth naming.** Whether the selection border ends up under the pointer during a drag depends on timing. Six gesture variants failed to hit it reliably inside the Playwright fixture, so the two multi-cell tests are honest coverage of multi-cell selection but are **not** a guard for the border case; that one rests on the measurements above. The test comment says so, rather than looking like it covers more than it does.

Two harness rules the new fixture encodes, both learned from false results while writing it:

- The grid is built on `load`, not during parsing. A container that still measures zero when Handsontable is constructed goes down the hidden-init path, where the first render waits on an observer that never fires for an element that was visible all along. That produced a 1-in-28 flake where the grid was simply empty.
- `dragAcrossText` throws if the drag would leave the grid's visible box. A clipped element still reports a layout box, so a drag aimed at it silently lands on whatever is really at those pixels — which is how an earlier measurement ended up driving the wrong grid entirely.

### Types of changes

### Related issue(s):

1. DEV-2720
2. #4980

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2720

---

> [!NOTE]
> **Medium Risk**
> Changes global mousemove/mousedown text-selection logic in TableView, but overlay confinement and extra DOM lookups run only when fragmentSelection is enabled; behavior without that option should stay the same aside from shared overlay resolution helpers.
> 
> **Overview**
> Fixes **`fragmentSelection`** so in-cell text selection works in **frozen rows, columns, and corners**, not only in the master viewport. Selection used to be blocked because **`isTextSelectionAllowed`** only treated elements inside the master spreader as selectable; frozen cells live in overlay clones beside the master.
> 
> **Walkontable** adds **`getParentOverlayByRenderedArea`**, resolving the overlay from each clone’s **spreader** (cells plus selection borders), not just the `TABLE`. **`TableView`** uses that to decide selectable body area (still excluding **`TH`** headers) and, when **`fragmentSelection`** is on, **records the starting overlay on mousedown** and **clears the selection** if the drag leaves that overlay—avoiding bogus cross-overlay ranges. Drag state is also reset on **document mouseup** so hovers after an outside release do not wipe host-page selection. **`fragmentSelection`** docs note the overlay boundary and that headers stay non-selectable.
> 
> Adds a **Playwright** fixture, page object, and **e2e** coverage for frozen areas, multi-cell drags, boundary clearing, and header guards.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fd27acfe2786f75cc9fe1419fc756387d05583bf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2720 fragmentSelection does not work in frozen rows and columns](https://app.clickup.com/t/9015210959/DEV-2720)

demtario (commented):
5 inline notes below.

handsontable/src/tableView.ts
```diff
@@ -1669,6 +1703,63 @@ class TableView {
...
+   *
+   * Everything the owning table renders counts, not just the cells. The selection borders sit in the
+   * spreader beside the table, and a multi-cell drag passes over them: rejecting one cancels a
+   * selection that is still inside the same area, which is what `fragmentSelection: true` exists to
+   * allow. Headers are the one exception — see below.
+   *
+   * @param {HTMLElement} el The element to check.
+   * @returns {boolean}
+   */
+  #isSelectableTableArea(el: HTMLElement) {
```
demtario, line 1757: **High**: selection-border elements inside a **frozen** overlay clone are misattributed to the master table here (and in `#hasLeftTextSelectionOverlay` below), so the confinement/permission logic this PR adds only works correctly for the master table's own borders.

`Border` (`handsontable/src/3rdparty/walkontable/src/selection/border/border.ts:356`) appends border divs into `this.wot.wtTable.spreader` — a sibling of `TABLE`, using the overlay's own `wot`. `getParentOverlay` (`overlays.ts:953`) matches an element to an overlay only via `overlay.clone.wtTable.TABLE.contains(element)`, which is never true for a border div. So `#getOwningWt` always falls back to `this._wt` (master) for border elements, regardless of which overlay they actually belong to.

Concretely: a border inside `inline_start` resolves as master's. `#isSelectableTableArea` then checks `isChildOf(borderEl, master_spreader)` → `false` → the drag is cancelled. `#hasLeftTextSelectionOverlay` resolves the same border to overlay `'master'`, which mismatches `#textSelectionOverlay` and reads as "left the overlay", clearing the selection mid-gesture even though the pointer never left the frozen area.

This is the same class of bug the PR's own third commit ("Keep selection borders selectable so multi-cell drags survive") fixed — but only for the master table, not for frozen rows/columns/corners.

Suggested fix: resolve the owning table for this specific check via `wtTable.spreader.contains(el)` per overlay (checking each clone, then master), rather than reusing `getParentOverlay`'s `TABLE`-based semantics — other callers (`updateCellHeader`, the public `getElementOverlayName`) depend on that method's current behavior, so it shouldn't change.

tests/e2e/fragment-selection-frozen.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+    await grid.dragAcrossCells('master', 1, 0, 2);
+
+    expect(await grid.selectedText()).toContain('R1C1');
+  });
+
+  test('selects across two cells inside the frozen columns', async () => {
+    await grid.initGrid({ fixedColumnsStart: 2, fragmentSelection: true });
+    await grid.clearTextSelection();
+
+    await grid.dragAcrossCells('inlineStart', 1, 0, 1);
```
demtario: **Medium**: `dragAcrossCells`'s `trail` array (`FragmentSelectionPage.ts:207-222`) is built specifically to prove the drag crossed the `wtBorder` element between the two cells, but neither this call nor the one at line 135 ever inspects it — both discard the return value. As written, the suite can't detect a regression on exactly the border-crossing case (see the High finding on `tableView.ts`), because the assertion (`toContain('\t')`) stays green whether or not the pointer actually landed on the border during the synthetic drag.

Suggested fix: assert `trail.includes('border')` in at least one of the two tests (with tolerance for coalesced Playwright moves), or drop the unused trail/`elementUnder` machinery if it's not meant to be asserted.

.changelogs/13320.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Fixed text selection not working inside frozen rows and columns when `fragmentSelection` is enabled.",
```
demtario: **Low**: the title covers only the frozen-rows/columns fix. This PR also makes header text unselectable under `fragmentSelection: true` (see the new `metaSchema.ts` JSDoc: "Headers are not selectable, whether or not they are frozen") — a second behavior change for existing users of that option, not mentioned here. Non-blocking; `"breaking": false` is still correct since no default setting changed.

tests/fixtures/demo/fragment-selection.html
```diff
@@ -0,0 +1,136 @@
...
+    // Theme and bundle come from the ?theme=/?bundle= query params set by the
+    // Playwright projects, mapped through fixed allowlists to LITERALS so a
+    // crafted param can never be reflected into the document — no XSS. An
+    // ABSENT param keeps the default (main / plain UMD) so the fixture stays
+    // browsable by hand; an unknown value THROWS instead of falling back — a
+    // typo in the project config must be one red leg, never a silently
+    // mislabeled green one (nothing else asserts which file actually loaded).
+    const htParams = new URLSearchParams(location.search);
+
+    window.htTheme = ({ main: 'main', horizon: 'horizon', classic: 'classic' })[htParams.get('theme') ?? 'main'];
```
demtario: **Low / nit**: the theme/bundle allowlists (here and line 24) are plain object literals, so a `?theme=` value naming an inherited `Object.prototype` key (`constructor`, `toString`, `hasOwnProperty`, ...) resolves to a truthy non-string instead of `undefined`, silently bypassing the fixture's own "throw on unknown value" fail-loud contract. Not an XSS vector — no attacker-controlled string ever reaches the DOM either way, only one of the fixed literals or a built-in's own `toString()`. Optional fix: `Object.create(null)` for the maps, or guard with `Object.hasOwn()` before indexing.

handsontable/src/tableView.ts
```diff
@@ -574,6 +602,12 @@ class TableView {
         return;
```
demtario, line 602: **Low**, for awareness only — not a new regression: the new `this.#textSelectionOverlay = null;` reset in this `documentElement` mouseup handler sits after the synthetic-touch-event early return above, so it's skipped for a synthetic (touch-emulated) mouseup. This mirrors the identical pre-existing pattern already present for `#selectionMouseDown` on this same handler, so the new field just inherits that existing gap rather than introducing a new one. No action required unless the team wants to fix both together.

handsontable/src/tableView.ts
budnix: Nice catch — and you were right about the mechanism too. Fixed in fd27acfe27 with a new `getParentOverlayByRenderedArea` that matches on the spreader; `getParentOverlay` is untouched, so `updateCellHeader` and `getElementOverlayName` keep their behavior.

tests/e2e/fragment-selection-frozen.spec.ts
budnix: Good spot, it was dead. In fd27acfe27 I dropped the return but kept the hit test, since it paces the drag and that is what puts the border under the pointer. Asserting on it does not work (it samples step positions, the border is hit between them), so the guard is now a three-cell frozen drag that goes red without the fix.

.changelogs/13320.json
budnix: Good point on the title — widened in fd27acfe27 to mention the confinement. On headers though, those were already unselectable in 18.0.0: I measured both builds and the header drag selects 0 either way, because headers render into a clone and that failed the old master-spreader check too. The JSDoc line documents behavior I kept, not a new one.

tests/fixtures/demo/fragment-selection.html
budnix: Agreed — fixed in fd27acfe27. `filters-value-list.html` already guards it the same way, so I copied that form.

handsontable/src/tableView.ts
budnix: Fair point — I put it there on purpose to match how `#selectionMouseDown` already behaves, so I am leaving both alone rather than growing this PR. Worth its own ticket if the team wants them fixed together.

tests/e2e/fragment-selection-frozen.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+    await grid.initGrid({ fixedColumnsStart: 3, colWidths: 150, fragmentSelection: true });
+    await grid.clearTextSelection();
+
+    await grid.dragAcrossCells('inlineStart', 1, 0, 2);
+
+    const selected = await grid.selectedText();
+
+    // The drag ends at the third cell's edge, so it takes text from the first two and stops there.
+    // What matters is that it survived at all: without the fix this same gesture selected nothing.
+    expect(selected).toContain('R1C1');
```
demtario: **Medium**: this rewritten test can pass without ever exercising the border-attribution bug it's meant to guard. The pre-fix bug only triggers when a `mousemove` lands on the `wtBorder` div (misattributed to master); a step landing only on plain `TD`s was never broken even before this fix. The fixture's own `window.elementUnder` hit-test (`fragment-selection.html`) already classifies each of the 20 drag steps as `'border' | 'cell' | 'other'`, but `dragAcrossCells` (`FragmentSelectionPage.ts:211-216`) now discards that result entirely, along with the dropped `trail` array. The only remaining assertion (`toContain('R1C1')`) proves a boundary was crossed, not that the border was actually hit mid-drag — the one thing the bug depended on. Whether one of the 20 fixed-fraction step positions actually lands on the border is pixel-geometry-dependent (cell padding/border width can vary by theme), so this could silently stop proving anything on some theme/bundle leg without failing red.

Suggested fix: assert the hit-test result includes `'border'` as a precondition before checking `selectedText()`.

handsontable/src/3rdparty/walkontable/src/overlay/overlays.ts
```diff
@@ -925,6 +925,43 @@ class Overlays {
...
+        return;
+      }
+
+      if (overlay.clone && overlay.clone.wtTable.spreader.contains(element)) { // todo demeter
+        result = overlay.clone;
+      }
+    });
+
+    return result;
+  }
```
demtario, line 965: **Low**: `getParentOverlayByRenderedArea` is a near-verbatim copy of `getParentOverlay` just below it — identical 5-element overlay array and `arrayEach` walk, differing only in `spreader.contains()` vs `TABLE.contains()`. A future overlay type added to one list and not the other would silently desync the two lookups.

Suggested fix: extract a private helper that takes the containment predicate as a parameter, and have both public methods call it.

demtario (commented):
2 inline notes below.

tests/e2e/fragment-selection-frozen.spec.ts
budnix: You are right, and the fix was simpler than I thought — done in e9c102dcea. The fixture now counts the real `mousemove` targets, so both tests assert the border was hit before checking the text. Sampling the step positions was what could not see it.

handsontable/src/3rdparty/walkontable/src/overlay/overlays.ts
budnix: Good point — extracted `#findParentOverlay` in e9c102dcea, so both lookups share one overlay list and just pass in the containment test.

demtario (approved):

PR #13321, opened 2026-09-01
DEV-2629: Stop the perf comment publishing unsupportable regressions

demtario:
### Context

The PR fixes the sticky performance comment posted on every pull request, which reports regressions that are not real. This is the third PR on this task, and the last one that is scoped. PR 1 (#13255) fixed how the suite measures; PR 2 (#13317) fixed which baseline it compares against. Both landed, and the comment is still wrong, for reasons neither of them touched.

Replaying the develop goldens published on `gh-pages` through PR 2's own median selection — with the baseline drawn only from the trailing window, never from the run under test — gives the current false-positive rate. Every one of these comparisons measures develop against develop, so every delta is noise by construction:

```
threshold   red callouts fired on no-change comparisons
5 (today)     35%  (47/135)
10            13%  (17/135)
15             3%   (4/135)
20             3%   (4/135)
worst absolute no-change delta: 53.2%
```

(19 goldens, 2026-08-27 through 2026-09-01, 135 scenario comparisons. The exact counts move as develop accumulates goldens; the shape does not. Reproduce with the command below.)

More than a third of no-change comparisons currently fire a red callout. That is the number this PR is aimed at.

The remaining cause is CI runner speed. Across those goldens the per-run speed factor spans **0.63x to 1.10x**, and it moves all nine scenarios together — the `2026-08-28T10-43-13Z` run reads low on every scenario at once. Normalizing it away arithmetically was considered and rejected: a median-of-ratios estimator is blind to the highest-value regression class. Injecting a synthetic uniform +20% slowdown across all nine scenarios, raw detection is 69-81% of rows and normalized detection is 6-10%.

**How that residual gets fixed is now an open question, and this PR does not depend on the answer.** The plan had been to measure the merge base and the PR head on the same runner in one job, so runner speed would cancel. That premise was tested before any of it was built, and it failed: running the suite twice in one job against a byte-identical build, the second run was slower on **8 of 9 scenarios**, median **+9.2%**, worst **+22.5%**. There is a systematic second-position penalty, so that design would have made every PR look like a regression rather than fixing anything. A dedicated always-on machine is being evaluated instead. Either way it is a change to how the baseline is obtained, and nothing in this PR is coupled to that.

**This PR owns the reporting layer only.** It changes nothing about how the suite measures or which baseline it loads. Its job is to stop the comment publishing numbers it cannot support, and to surface the reliability data the tooling already computes and throws away. Every file it touches had zero tests before this PR.

#### Three corrections to the scope carried into this PR

The scope was written before #13255 and #13317 landed. Three items no longer described reality, and following them literally would have shipped dead code or a regression in the tool.

**1. The CV gate as filed is inert.** It was written against pre-#13255 data where intra-run CV was 58-97%. Post-#13255 it is 0.9-4.0% on seven of nine scenarios, so a 15% gate on it never fires. The signal that matters now is run-to-run spread across the median window (11-19% per scenario), which #13317 already assembles. The comment needs both, labeled distinctly.

**2. No suppression gate was added.** The scope asked for one. Simulated: threshold 10 plus "suppress when the window CV exceeds 15%" gets false callouts to 4%, but it also suppresses **51 of 89 genuine +25% regressions**. It buys accuracy by going half-blind. Raising the threshold alone is strictly better, and the knee is sharp:

```
threshold   no-change(false)   real +25%   real +50%   real +100%
5             35%                79%         91%         100%
10            13%                76%         90%          99%
15             3%                69%         88%          99%
20             3%                61%         87%          98%
```

15 is the knee. Above it, sensitivity drops with no false-positive gain.

**3. The threshold had to be split, not raised.** `REGRESSION_CALLOUT_THRESHOLD` gated heap as well as timing. Heap's run-to-run CV is 0.4-3.6% against timing's 11-19%, so one constant cannot serve both:

```
threshold   heap no-change(false)   real +10% leak   real +25% leak
2             14%                     99%              100%
3             12%                     97%              100%
5 (today)      4%                     93%              100%
10             0%                     60%              100%
15             0%                      9%              100%
```

Raising the shared constant to 15 would have taken detection of a genuine +10% heap leak from 93% to 9% — a regression in the tool, introduced by the fix for it. Heap stays at 5, which is already at its knee. Only the timing threshold moves.

The timing number is **interim, and correct only for the baseline it was measured against** — today's cross-runner median. Whenever the comparand changes, it has to be re-derived, because a quieter baseline makes 15 far too permissive to catch anything. `performance-tests/scripts/replay-goldens.mjs` is committed here so that is a command, not a judgement call.

### What changed

**One rule decides whether a delta may be published.** `comparability(baselineCategories, currentCategories, isCrossWindow)` in `thresholds.mjs` is consulted by every site in both builders that renders a percentage. Three things make a comparison unpublishable:

| Reason | Rendered as | What is withheld |
|---|---|---|
| The baseline recorded 0 ms for a category this run did record | `baseline incomplete` | the total, and that category |
| This run recorded 0 ms for a category the baseline did record | `capture incomplete` | the total, and that category |
| The two sides were measured over different trace windows | `window mismatch` | everything derived from the trace, heap and DOM counters included |

The first is the filed defect: ~4 ms of real work divided against a baseline that never measured it, published as `+115.7% 🔴`. The check is **symmetric**, because the mirror case is equally unsupportable — a capture failure on this run's side publishes a fake green `-23%` a reader cannot tell from a real win. The label names the side and the categories (`this run captured no rendering or painting`), which is the detail that says whether re-running develop would fix it. A scenario in this state is listed under "Total delta not assessed" rather than folded into "within tolerance", and counted separately in the HTML dashboard rather than as Neutral.

Categories both sides captured are still published: one failed capture does not hide the rest of the row.

**One CV implementation.** `fmtCv()` had zero callers repo-wide; `html-report-builder.mjs` carried a private near-duplicate `calcCv`; both hardcoded `15`. Now one exported `calcCv()` plus `CV_WARNING_THRESHOLD`. It uses the **sample** standard deviation (n-1): at three iterations the population form understates spread by about 18%, and this number now gates how much a reader trusts the delta beside it.

**Two reliability numbers, labeled distinctly**, as `CV run / base`. The first is the spread across this run's three iterations; the second is how far apart the develop runs behind the median baseline sit. Baseline spread was not previously available to the report layer at all — the per-iteration values are stripped before a golden is saved — so it is computed in `median-snapshot.mjs`, which already holds the whole window. It renders `n/a`, never `0.0%`, in the three paths with no history window (golden mode, the self-compare fallback, a thin history); a rendered `0.0%` would claim a perfectly stable baseline, the same class of lie this PR exists to remove. Hook-timing CV is surfaced too, from `_iterationValues.hookTiming`, collected and never read.

**Colour bands derive from the callout constants.** `classifyChange` banded at ±10 while the callout fired at a different number, and the client script carried a second copy of those bands. Left alone, the HTML report would paint a row red at +12% while the comment said "within tolerance" — precisely the "two numbers describing the same scenario" complaint on this task. The bands now come from the split constants, serialized into the payload rather than restated.

**A self-comparison is no longer reported as a develop baseline.** The no-golden fallback stamps a fresh timestamp, and both builders branched on the timestamp merely being present — so a run compared against itself printed `Baseline: single develop run <now>` and cleared every scenario. It is marked `isSelfCompare`, and the tolerance claim is suppressed.

**The Trace window row is informational and never coloured.** After #13255 it is mark-to-mark harness wall clock: on the four scroll scenarios it sits at ~16700 ms with a run-to-run CV of 0.0-0.2% regardless of what the grid does, because it is 500 sequential wheel round trips. It was still getting a red badge above 10%. It is the one percentage still printed on a window mismatch, deliberately — it is the size of the two windows, so it explains the mismatch the other rows are withheld for.

**Provenance in both reports.** `isMedian`, `medianWindowSize`, `medianSourceTimestamps` reached the CI log only, and the commit/run metadata reached the snapshot but not the report. Neither artifact said whether a delta was computed against one develop run or five. Both now do. `PERF_COMMIT_SHA` is plumbed through both CI entry points, because on the `pull_request` path the default `GITHUB_SHA` is the ephemeral merge commit, which exists on no branch.

**Script-block containment.** The payload is serialized into an inline `<script>`, and `JSON.stringify` does not escape `<`, so a branch name containing `</script>` closed the block early — and the branch name comes from `GITHUB_HEAD_REF`, which a fork pull request chooses. Pre-existing rather than introduced here, but this PR touches exactly that serialization. Escaped, with the same hole closed in the `<title>` interpolation.

**First tests for the reporting layer**, plus the two stale docs (`performance-tests/README.md` and the performance-testing skill both listed 7 scenarios at the wrong fixture sizes; reality is 9).

### Review rounds

Four automated review passes, across six commits. They found real defects each time, and the pattern is worth recording for whoever reviews this next: every miss was the same shape — gating one render site instead of the class.

- **Round 1** found 8, two serious. `buildDetailedMetrics` published the withheld `+115.6%` in the expanded detail table while the card badge above it read `baseline incomplete`. And the self-comparison-as-develop-baseline defect above.
- **Round 2** found that the cross-window heap fix had landed in one of three places, so one run produced three verdicts: a red heap cell in the table, "within tolerance" in the callout below it, and `regressions: 1` in the linked report. Also that cross-window per-category deltas were still published, in colour, under a badge saying the baseline was incomplete.
- **Enumerating the class myself** rather than patching the reported instance then found a third site: `buildMemoryMetrics`. Heap min/max, node and listener counts are all extrema over the same window.
- **Round 3** found that making the guard symmetric had opened a new hole: per-category deltas were gated on the window alone, so a run that missed a category badged itself incomparable and then published a green `-100.0%` for that category one row below. `pctChange`'s zero-denominator guard had masked the mirror case, which is why only the new direction leaked. It also found the label misattributing a current-side failure to the baseline on two of three surfaces.

That is what `comparability()` exists for: the rule is stated once, and the per-category exemption comes from the verdict rather than from a condition re-derived at each call site.

### Follow-up

Whatever replaces the comparand also carries the `cell-editing` `selectCell` change. That change removes the 20 `page.evaluate` round trips from the measured window, and is deliberately **not** here: it redefines what the scenario measures, so for the next five develop pushes the median window would hold a mix of old-definition and new-definition snapshots, reading as a large sustained improvement that could mask a real regression. It is cheapest to land alongside a baseline change that invalidates the window anyway.

Two things reviewed and deliberately left:

- `updateCounters.documentsMin/Max` are parsed and medianed but reach no report. Pre-existing, and adding a Documents row is a display decision rather than a correctness fix.
- The symmetric guard means a genuine optimisation that drives a category to 0 ms is withheld as a capture failure rather than reported as a win. The replay bounds that cost: across 135 single develop runs standing in for a PR, none recorded exactly 0 ms in any active category. Note the bound is one-directional — the baseline side of a replay is a median, which dilutes a single zero capture, so it does not bound the baseline-side rate.

**This PR does not close the task.** It stops the comment publishing unsupportable numbers, which is worth having on its own — every fix here is about not asserting more than the measurement supports, and none of it changes when the baseline strategy is settled. The 3% residual and the runner-speed cause remain, and the task stays open for them.

### Reading this PR's own performance comment

No scenario golden is invalidated by this change, so any large delta on this PR's own comment is either a genuine finding or the runner-speed noise described above. Do not read a single run of it, green or red, as validating the change.

`[skip changelog]` — this PR only touches `performance-tests/**`, two CI files and two documentation files. No published package changes.

### How has this been tested?

- **New unit tests**, the first for any of these files: `thresholds.test.mjs` (48 cases), `report-builder.test.mjs` (28), `html-report-builder.test.mjs` (22). Between them they cover the comparability rule in all four states and at the boundaries, per-category gating in both directions, sample-vs-population CV at n=3, the bands tracking the callout constants, `n/a` degradation on all three no-history paths, the self-comparison path, cross-window suppression of timing, heap and memory, script-block containment, and the provenance footer. `median-snapshot.test.mjs` gains 2 cases for the new per-scenario spread.

  ```bash
  npm run test:tooling
  ```

  283 tests pass, up from 179.

- **Mutation-checked**, so the tests are not hollow. Five separate reversions each turn the suite red: population standard deviation in `calcCv`; a shared timing/heap threshold; dropping the heap threshold argument at the `Δ Heap` call site; ungating the memory deltas; ungating `categoryMetric`.

- **Rendered in a real browser.** The HTML report was served over HTTP and loaded in Chromium: no console errors, both script blocks parse, the provenance line renders, the incomparable card shows the right label, `Total active` carries a CV where it was hardcoded `null`, `Trace window (harness wall clock)` is labeled and uncoloured, and only a spread above the threshold is flagged.

- **The threshold derivation is reproducible** and committed, rather than a number chosen by eye:

  ```bash
  git fetch origin gh-pages
  node performance-tests/scripts/replay-goldens.mjs --since 2026-08-27
  ```

  It prints both curves and marks the value currently in force. It applies the same comparability rule the reports use and checks `windowSource` on the PR side, so it cannot count a comparison the shipped code would refuse to publish. Re-run after the symmetric-guard change: 0 skipped, curve unchanged, 15 still the knee.

- `npx eslint` clean on every touched file.

### Types of changes

### Related issue(s):

1. DEV-2629

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

---

> [!NOTE]
> **Low Risk**
> Changes are confined to the performance-tests package, CI env for reports, and documentation—no Handsontable runtime or published package behavior.
> 
> **Overview**
> **Stops the sticky perf comment and HTML report from publishing percentage deltas when the comparison cannot support them.** A shared `comparability()` rule in `thresholds.mjs` gates every delta site for incomplete baselines, incomplete current captures, and trace **window mismatch** (including heap and memory extrema), with explicit labels instead of fake regressions like `+115.7%`.
> 
> **Makes reporting more honest and aligned across surfaces.** Timing vs heap regression callouts use separate thresholds (15% / 5%) with colour bands serialized from the same constants; the markdown table adds **CV run / base** (intra-run vs median-baseline spread from `median-snapshot.mjs`); provenance and `PERF_COMMIT_SHA` in CI fix cited commits on PRs; self-compare runs are labeled and no longer read as “within tolerance”; HTML payload/title escaping hardens fork branch names.
> 
> **Adds `scripts/replay-goldens.mjs`** to re-derive threshold false-positive rates from gh-pages goldens, plus large unit-test coverage for `thresholds`, `report-builder`, and `html-report-builder`. Docs/skills are updated for nine scenarios, thresholds, and withheld-delta behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ea3f1db383f080bf24361137390d7cf6935bd0ee. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/performance-tests/README.md
+++ b/performance-tests/README.md
@@ -56,13 +56,15 @@ Each scenario measures a specific user interaction pattern:
 
 | Scenario | Grid size | Action | Notes |
 |---|---|---|---|
-| **scroll-down** | 5000 x 10 | `mouse.wheel(0, 350)` x 500 | Vertical scroll from top |
-| **scroll-up** | 5000 x 10 | `mouse.wheel(0, -350)` x 500 | Pre-scrolls to bottom, then scrolls up |
+| **scroll-down** | 10000 x 50 | `mouse.wheel(0, 350)` x 500 | Vertical scroll from top |
+| **scroll-up** | 10000 x 50 | `mouse.wheel(0, -350)` x 500 | Pre-scrolls to bottom, then scrolls up |
 | **scroll-right** | 10 x 5000 | `mouse.wheel(350, 0)` x 500 | Horizontal scroll from left |
 | **scroll-left** | 10 x 5000 | `mouse.wheel(-350, 0)` x 500 | Pre-scrolls to right, then scrolls left |
-| **filtering** | 1000 x 1000 | `filters.addCondition()` + `filter()` | Hook timing: beforeFilter -> afterFilter |
-| **sorting** | 1000 x 1000 | `columnSorting.sort()` | Hook timing: beforeColumnSort -> afterColumnSort |
+| **filtering** | 100000 x 100 | `filters.addCondition()` + `filter()` | Hook timing: beforeFilter -> afterFilter |
+| **sorting** | 100000 x 100 | `columnSorting.sort()` | Hook timing: beforeColumnSort -> afterColumnSort |
 | **cell-editing** | 5000 x 10 | selectCell + Enter + type + Enter x 20 | Sequential cell edits |
+| **initial-load** | 100000 x 100 | `new Handsontable(...)` | Grid construction only |
+| **source-data-validator-load** | 100000 x 100 | `new Handsontable(...)` with `sourceDataValidator` | Same fixture as initial-load plus the one option |
 
 Each scenario runs **1 warmup iteration** (discarded) followed by **3 measured iterations** with CDP tracing.
 
@@ -168,7 +170,64 @@ Additional metrics from `UpdateCounters`:
 - DOM node count (min/max)
 - Event listener count (min/max)
 
-**CV%** (coefficient of variation) is shown per metric. Values above 15% are flagged with `!!!` -- these indicate unstable measurements that may not be reliable for comparison.
+### Reading a delta
+
+Two spreads appear beside every delta, and they answer different questions. Both are a coefficient
+of variation over the sample standard deviation, and both are flagged above `CV_WARNING_THRESHOLD`
+(15%).
+
+| Column | What it measures | Typical value |
+|---|---|---|
+| **CV run** | How much the three iterations of this one run disagreed | under 4% on most scenarios |
+| **CV base** | How far apart the develop runs behind the median baseline sit | 11-19% |
+
+A tight `CV run` says the run measured itself consistently. It says nothing about whether the run is
+comparable to the baseline, which is what `CV base` reports. When the second number is wide, the
+delta beside it is measured against a moving target.
+
+`n/a` means the spread is genuinely unknown, not zero. The baseline spread is absent in golden mode,
+in the self-compare fallback, and whenever the history window is too thin for a median.
+
+### Thresholds
+
+Timing and heap are called out on separate thresholds, in `lib/thresholds.mjs`:
+
+| Constant | Value | Why |
+|---|---|---|
+| `REGRESSION_CALLOUT_THRESHOLD_TIMING` | 15 | Timing's run-to-run CV is 11-19%, so a lower bar fires mostly on noise |
+| `REGRESSION_CALLOUT_THRESHOLD_HEAP` | 5 | Heap's run-to-run CV is 0.4-3.6%, so it can be judged an order of magnitude tighter |
+
+Neither number is a guess, and neither should be changed by eye. Both come from replaying every
+develop golden against a median of the goldens before it: every such comparison measures develop
+against develop, so every delta it produces is noise, and the rate at which a threshold fires on
+that set is its false-positive rate. Reproduce it with:
+
+```bash
+git fetch origin gh-pages
+node scripts/replay-goldens.mjs --since 2026-08-27
+```
+
+The script prints the curve for both metrics and marks the value currently in force. Re-run it after
+any change to how the suite measures or to which baseline it compares against, and move each
+constant to the knee of its curve.
+
+### What the report refuses to publish
+
+A percentage is withheld, and rendered as `baseline incomplete`, when the comparison cannot support
+one:
+
+- The baseline recorded zero for an active category that the current run did record. That is a
+  failed capture, not a cheap operation, and dividing into it is what produced the `+115.7%`
+  regressions this reporting layer was rewritten to stop.
+- The two sides were measured through different trace windows (`windowSource` disagrees). After the
+  measurement fix this is a safety net rather than a common case.
+
+Such a scenario is listed under "Not assessed" rather than folded into "within tolerance": it was
+neither cleared nor flagged, and the comment says so.
+
+The **Trace window** row is informational and never coloured. After the measurement fix it is
+mark-to-mark harness wall clock, which on the scroll scenarios is 500 sequential wheel round trips
+sitting at a run-to-run CV of 0.0-0.2% regardless of what the grid does.
 
 ## Adding a new scenario
 
--- a/performance-tests/README.md
+++ b/performance-tests/README.md
@@ -186,7 +186,9 @@ comparable to the baseline, which is what `CV base` reports. When the second num
 delta beside it is measured against a moving target.
 
 `n/a` means the spread is genuinely unknown, not zero. The baseline spread is absent in golden mode,
-in the self-compare fallback, and whenever the history window is too thin for a median.
+in the self-compare fallback, in the single-run `latest.json` fallback (fewer than
+`MIN_VALID_SNAPSHOTS` qualifying develop snapshots, so `computeMedianSnapshot` returns `null`), and
+whenever the history window is too thin for a median.
 
 ### Thresholds
 
@@ -213,17 +215,21 @@ constant to the knee of its curve.
 
 ### What the report refuses to publish
 
-A percentage is withheld, and rendered as `baseline incomplete`, when the comparison cannot support
-one:
+A percentage is withheld when the comparison cannot support one. The label names the cause --
+`baseline incomplete`, `capture incomplete`, or `window mismatch` -- and the reasons are:
 
-- The baseline recorded zero for an active category that the current run did record. That is a
-  failed capture, not a cheap operation, and dividing into it is what produced the `+115.7%`
-  regressions this reporting layer was rewritten to stop.
-- The two sides were measured through different trace windows (`windowSource` disagrees). After the
-  measurement fix this is a safety net rather than a common case.
+- The baseline recorded zero for an active category that the current run did record
+  (`baseline incomplete`). That is a failed capture, not a cheap operation, and dividing into it is
+  what produced the `+115.7%` regressions this reporting layer was rewritten to stop.
+- This run recorded zero for an active category that the baseline did record (`capture incomplete`).
+  The mirror case: a capture failure on this run's side is no more supportable than one on the
+  baseline's, and would otherwise publish a fake win a reader cannot tell from a real one.
+- The two sides were measured through different trace windows (`windowSource` disagrees, rendered
+  `window mismatch`). After the measurement fix this is a safety net rather than a common case.
 
-Such a scenario is listed under "Not assessed" rather than folded into "within tolerance": it was
-neither cleared nor flagged, and the comment says so.
+Such a scenario is listed under "Not assessed" in the HTML report (the markdown comment says "Total
+delta not assessed") rather than folded into "within tolerance": it was neither cleared nor flagged,
+and the comment says so.
 
 The **Trace window** row is informational and never coloured. After the measurement fix it is
 mark-to-mark harness wall clock, which on the scroll scenarios is 500 sequential wheel round trips
--- a/performance-tests/lib/html-report-builder.mjs
+++ b/performance-tests/lib/html-report-builder.mjs
@@ -6,7 +6,6 @@ import {
   REGRESSION_CALLOUT_THRESHOLD_TIMING,
   REGRESSION_CALLOUT_THRESHOLD_HEAP,
   CV_WARNING_THRESHOLD,
-  BASELINE_INCOMPLETE_LABEL,
   activeTotalsPerIteration,
   calcCv,
   classifyChange,
@@ -158,7 +157,7 @@ function buildPayload(scenarioResults, goldenScenarios, hasGolden, meta, goldenS
         painting: categoryMetric('painting', gCats, cCats, verdict),
         total: { current: currentTotal, baseline: goldenTotal || 0, change: totalChange },
       },
-      detailedMetrics: buildDetailedMetrics(current, golden, baselineIncomplete, verdict),
+      detailedMetrics: buildDetailedMetrics(current, golden, verdict),
       memory: buildMemoryMetrics(current, golden, isCrossWindow),
       hookTiming: buildHookTiming(current, golden),
       heap,
@@ -206,7 +205,6 @@ function buildPayload(scenarioResults, goldenScenarios, hasGolden, meta, goldenS
       heap: REGRESSION_CALLOUT_THRESHOLD_HEAP,
       cvWarning: CV_WARNING_THRESHOLD,
     },
-    baselineIncompleteLabel: BASELINE_INCOMPLETE_LABEL,
     summary: {
       total: scenarios.length,
       regressions,
@@ -219,15 +217,24 @@ function buildPayload(scenarioResults, goldenScenarios, hasGolden, meta, goldenS
   };
 }
 
-function buildDetailedMetrics(current, golden, baselineIncomplete = false, verdict = null) {
+/**
+ * @param {object} current
+ * @param {object | null} golden
+ * @param {{ comparable: boolean, reason: string | null, incompleteCategories: string[] }} verdict
+ *   -- required, and deliberately not defaulted: a caller passing the old boolean argument shape
+ *   would silently un-gate every per-category delta rather than fail
+ * @returns {Array<object>}
+ */
+function buildDetailedMetrics(current, golden, verdict) {
   const rows = [];
   const cCats = current.categories || {};
   const gCats = golden?.categories || {};
-  const incompleteCategories = verdict?.incompleteCategories ?? [];
+  const baselineIncomplete = !!golden && !verdict.comparable;
+  const { incompleteCategories } = verdict;
   // Only the active categories participate in the comparability verdict. The others (loading,
   // other, experience, idle) are reported but never summed into a total, so a window mismatch is
   // the only thing that invalidates them.
-  const isCrossWindow = verdict?.reason === 'window-mismatch';
+  const isCrossWindow = verdict.reason === 'window-mismatch';
 
   for (const key of ['scripting', 'rendering', 'painting', 'loading', 'other', 'experience', 'idle']) {
     const c = cCats[key];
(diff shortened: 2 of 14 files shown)


qunabu:

budnix (commented):
Five notes, none blocking. The direction is right — I read the whole reporting layer and the withhold logic holds up. These are the spots where the new behavior stops one step short of what the surrounding comment promises.

Checked and found fine, so you can skip re-explaining them: the `&lt;` escape in `serializePayload`, the dashboard bucket split (it cannot double-count), and the duplicated `PERF_COMMIT_SHA`.

performance-tests/lib/html-report-builder.mjs
```diff
@@ -534,7 +670,9 @@ a:hover { text-decoration: underline; }
 .metrics-table tr:last-child td { border-bottom: none; }
 .metrics-table .bold { font-weight: 600; }
 .metrics-table .num { font-variant-numeric: tabular-nums; text-align: right; }
-.metrics-table .cv-warn { color: #cf222e; font-weight: 600; }
+/* Not scoped to .metrics-table: the baseline-spread note in a card header carries this class too,
+   and a table-only rule made that flag a silent no-op. */
+.cv-warn { color: #cf222e; font-weight: 600; }
```
budnix, line 682: Unscoping this from `.metrics-table` was the right call, but the baseline-spread note still does not turn red. `.card-header .baseline-spread` (line 552) sets `color: #656d76` at two-class specificity, and this rule has one. The grey wins, so all that gets through is the bold weight.

`buildScript()` line 937 is the caller that adds the class to that element, so the no-op is on the exact path this comment describes.

```suggestion
.cv-warn { color: #cf222e; font-weight: 600; }
.card-header .baseline-spread.cv-warn { color: #cf222e; }
```

performance-tests/lib/report-builder.mjs
```diff
@@ -154,11 +311,68 @@ function buildRegressionCallouts(results, goldenScenarios) {
     callouts.push(lines.join('\n'));
   }
 
+  // A scenario whose baseline is unusable was neither cleared nor flagged, so say so rather than
+  // letting it fall silently into "within tolerance". Scoped to the total, because heap may still
+  // have been assessed for the same scenario -- otherwise the note would contradict a callout
+  // standing directly above it.
+  const note = skipped.length > 0
```
budnix: The comment above says heap may still have been assessed, and that is true for a `capture incomplete` scenario — but not for a window mismatch. Line 285 sets `heapPct` to `null` whenever `isCrossWindow`, so for those scenarios heap was not assessed either, and a reader of "Total delta not assessed" concludes heap passed.

`skipped` already carries the label, so the two cases can be told apart without new state — the entries ending in `(window mismatch)` are exactly the ones where heap is out too.

Worth knowing before you touch it: the wording is pinned by `performance-tests/lib/__tests__/report-builder.test.mjs`, so a fix needs the test updated in the same commit.

performance-tests/lib/html-report-builder.mjs
```diff
@@ -111,30 +179,69 @@ function buildPayload(scenarioResults, goldenScenarios, hasGolden, meta) {
...
+      heap: REGRESSION_CALLOUT_THRESHOLD_HEAP,
+      cvWarning: CV_WARNING_THRESHOLD,
+    },
     summary: {
       total: scenarios.length,
       regressions,
       improvements,
-      neutral: scenarios.length - regressions - improvements,
+      notAssessed,
+      neutral: scenarios.length - regressions - improvements - notAssessed,
```
budnix: A scenario with no baseline falls through to `neutral` here, because it is counted as the remainder. "Neutral" claims the scenario is flat against its baseline, which is the one thing not known about it — there is nothing to be flat against.

The markdown report has the mirror gap at `performance-tests/lib/report-builder.mjs:271`: `if (!golden) continue;` drops the scenario before `skipped` is built, so it is not listed as skipped there either.

A fourth count (`noBaseline`) subtracted out of the remainder would keep "Neutral" meaning what it says, and give a new scenario somewhere honest to land.

performance-tests/lib/teardown.mjs
```diff
@@ -130,9 +130,15 @@ function collectIterationValues(parsedResults) {
...
-      .filter(v => typeof v === 'number');
+    // Index-aligned to the iteration, never compacted. Dropping the entries where a category
+    // recorded nothing would both understate that category's own spread and, once the arrays are
+    // recombined into an active total, pair one iteration's scripting with another's painting.
+    // An iteration that recorded no time in a category recorded zero of it.
+    values.categories[key] = parsedResults.map((r) => {
+      const value = r.categories?.[key];
+
+      return typeof value === 'number' && Number.isFinite(value) ? value : 0;
+    });
```
budnix, line 141: The zero-fill is right for the CV, and the reasoning above it is sound. The mismatch is with the number printed next to it: `averageParsedTraces` in `performance-tests/trace-parser.mjs:1009` filters the missing values out before the mean, so the average and the spread describe different sets of iterations.

Concrete case — painting seen in 2 of 3 runs at 30ms each:

- Total shows `30` (mean of `[30, 30]`)
- CV is computed from `[30, 30, 0]`, so it reads as a wildly unstable category

Same underlying data, two denominators. Whichever rule is right, both sides should use it. That file is not in this diff, so this is a heads-up rather than a change request.

performance-tests/README.md
```diff
@@ -168,7 +170,64 @@ Additional metrics from `UpdateCounters`:
...
+```
+
+The script prints the curve for both metrics and marks the value currently in force. Re-run it after
+any change to how the suite measures or to which baseline it compares against, and move each
+constant to the knee of its curve.
+
+### What the report refuses to publish
+
+A percentage is withheld, and rendered as `baseline incomplete`, when the comparison cannot support
+one:
```
budnix: `baseline incomplete` is one of three labels this path can render, not the only one. `lib/thresholds.mjs:30-33` maps `current-incomplete` and `both-incomplete` to `capture incomplete`, and `window-mismatch` to `window mismatch` — and the second bullet below is a window mismatch, so it does not render as `baseline incomplete` at all.

```suggestion
A percentage is withheld when the comparison cannot support one. The label names the cause --
`baseline incomplete`, `capture incomplete`, or `window mismatch` -- and the reasons are:
```

Two smaller gaps in the same block: the `n/a` spread list does not mention the single-run `latest.json` fallback, and "Not assessed" is HTML-only wording — the markdown comment says "Total delta not assessed".

performance-tests/lib/html-report-builder.mjs
```diff
@@ -144,12 +244,18 @@ function buildDetailedMetrics(current, golden) {
...
+    const incomplete = isCrossWindow || incompleteCategories.includes(key);
+
     rows.push({
       label: categoryLabel(key),
       key,
       current: c ?? 0,
       baseline: g ?? 0,
-      change: pctChange(g, c),
+      change: incomplete ? null : pctChange(g, c),
+      incomplete,
```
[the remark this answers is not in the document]
demtario: Fixed in b03f074eac (landed before this review round): `NO_BASELINE_VERDICT` gives `categoryMetric`/`buildDetailedMetrics` `comparable: true` and empty `incompleteCategories` when there's no golden at all, so a brand-new scenario (or one the median omitted) reads `--` in the detail table, matching the markdown's early return. Covered by `buildHtmlReport -- scenario absent from the baseline` in html-report-builder.test.mjs.

performance-tests/lib/html-report-builder.mjs
```diff
@@ -111,30 +172,69 @@ function buildPayload(scenarioResults, goldenScenarios, hasGolden, meta) {
...
+      heap: REGRESSION_CALLOUT_THRESHOLD_HEAP,
+      cvWarning: CV_WARNING_THRESHOLD,
+    },
     summary: {
       total: scenarios.length,
       regressions,
       improvements,
-      neutral: scenarios.length - regressions - improvements,
+      notAssessed,
+      neutral: scenarios.length - regressions - improvements - notAssessed,
```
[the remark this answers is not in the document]
demtario: Fixed in b03f074eac: the Neutral filter excludes `!s.notAssessed`. Just extended it further (this round) to also exclude `!s.hasBaseline`, so a scenario missing from the baseline entirely doesn't leak into Neutral either -- see the noBaseline bucket below.

performance-tests/lib/html-report-builder.mjs
demtario: Fixed with your suggested rule: `.card-header .baseline-spread.cv-warn { color: #cf222e; }` at two-class specificity now beats the grey `.card-header .baseline-spread` rule.

performance-tests/lib/report-builder.mjs
demtario: Fixed: the "Total delta not assessed" note now appends "; heap also not assessed" specifically for a window-mismatch entry, and stays as-is for a plain baseline-incomplete one where heap is still independently assessed. Updated `report-builder.test.mjs` in the same commit ('says heap is also not assessed when the total is withheld for a window mismatch' / 'does not say heap is also not assessed for a baseline-incomplete comparison').

performance-tests/lib/html-report-builder.mjs
demtario: Fixed: added a fourth `noBaseline` bucket -- dashboard counter + filter in html-report-builder.mjs, and a "No baseline yet" note in report-builder.mjs that fixes the mirror gap you flagged at report-builder.mjs:271. Neutral now excludes it on both surfaces.

performance-tests/lib/teardown.mjs
demtario: Noted, leaving as-is since it's outside this diff, as you said. Flagging for whoever next touches `trace-parser.mjs`/`averageParsedTraces`.

performance-tests/README.md
demtario: Fixed with your suggested wording, plus named the current-side (`capture incomplete`) case explicitly rather than leaving it implicit. Also added the single-run `latest.json` fallback to the n/a spread list, and noted the HTML says "Not assessed" where the markdown says "Total delta not assessed".

budnix (approved):

Issue #7031, opened 2020-06-18
Methods `colToProp` and `propToCol` should work properly


wszymanski:
### Description
As described in the title, it seems that methods [colToProp](https://handsontable.com/docs/7.4.2/Core.html#colToProp) and [propToCol](https://handsontable.com/docs/7.4.2/Core.html#propToCol) may not work properly. They don't return `null` values for indexes beyond the table boundaries. It doesn't fit changes done within https://github.com/handsontable/handsontable/pull/5945. Since `8.0.0-beta.1` release methods like `toPhysicalRow`, `toVisualColumn`, etc. return `null` for non-existent rows/columns.

Maybe the type of indexes handled by methods isn't proper too. It may have an influence on the work of the `setDataAtRowProp`. Please keep in mind that this method currently has some problems. It is described within https://github.com/handsontable/handsontable/issues/4118.

### Your environment
* Handsontable version: `8.0.0-beta.2`

budnix:
Hi @wszymanski,

I tested this on the current `develop` and it still reproduces, so the report is still valid.

Both methods hand the argument back when it names no column. That answer cannot be told apart from a real one. When the visual and physical column order have drifted apart, the number returned points at a *different* column than the one you asked about:

```js
// 4 columns, physical column 1 trimmed
toVisualColumn(1);          // null — no visual index
propToCol(1);               // 1    — claims the column sits at visual 1
getDataAtCol(propToCol(1)); // reads physical column 2
```

I have a fix open in #13330. Both methods now return `null` when the argument names no column that exists and is visible. That lines them up with `toVisualColumn()` and the other index translators, which is what you asked for.

Two details worth knowing:

- A property name your data set does not use is still handed back unchanged. Only indexes resolve.
- `propToCol()` now answers the same way for a trimmed column whether its property is cached or not. Those two paths used to disagree.

This changes what the methods return, so it is a breaking change. It will ship in the next major release, not in an 18.x patch. A migration guide with before and after examples ships with it.

On the `setDataAtRowProp()` link: that one (#4118) was a separate bug and is already fixed.

Issue #4118, opened 2017-03-07
Issue with setDataAtRowProp() - bad column index translation


zoladkow:
### Description
I want to make changes on the grid, by way of passing an array of changes [[row, prop, value], ... ]
I expect those changes to be passed to `afterChange` handler as: [[row, prop, oldValue, newValue], ... ]

What I actually get is the changes array where oldValue is from a different column/prop/index.

While debugging I found the problem is at `getAtCell()` function where the prop I pass is translated using `this.colToProp(column)` which is strange since I call a `setDataAtRow**Prop**()` and that **prop** is never translated to **col** along the way.

Accidentally my prop is actually an index in datasource array - I use array of arrays not objects as datasource (less data overhead) so I can see the confusion. But I'd expect all column/prop translation to be done consequently. 

Also - the changes I pass to `setDataAtRowProp()` are rendered in correct cells, so it seems it is only the `afterChange()` collection which gets the wrong oldValues.

But if I tried to pass a changes array with doing `propToCol()` - I get the expected `afterChange` changes, but things displayed on the grid get messed up.

### Demo
not so simple to create a simple demo, instead an example of changes data, input and output:
Contains part of columns collection to show how each grid column maps to datasource column

[sample-data.txt](https://github.com/handsontable/handsontable/files/823823/sample-data.txt)

edit: [simple demo](http://jsfiddle.net/zoladkow/5jzke7Lz/)

1. select a row on the grid
2. **WITHOUT** changes in the inputs, click first `Save ` button
3. again, **WITHOUT** changes in the inputs, click second `Save` button

### Your environment
* Handsontable version: 0.31.0
* Browser Name and version: Chrome 55
* Operating System: Win 10

zoladkow:
Oh, one more thing - this also impacts the undo feature - the changes are badly calculated

AMBudnik:
Hi @zoladkow 

Thank you for the detailed description and the demo.
You are right it is a bit complicated example. Can you tell me where and what exactly happens wrong and what would be an expected result?

Thanks in advance.

zoladkow:
Sure, here is what goes wrong:
![2017-03-13_15-31-51](https://cloud.githubusercontent.com/assets/11553616/23858971/7cec3bd6-0802-11e7-980e-04e1c2a6d3a7.gif)

And the log of changes passed to `afterChange`:
```
[loadData] null
[...]

PR #13322, opened 2026-09-01
DEV-2721: Read setDataAtRowProp old value by prop, not visual column

budnix:
### Context

The PR fixes the old value reported by `setDataAtRowProp()` and `setSourceDataAtCell()`, and with it the data corruption that undo caused afterwards. GitHub issue [#4118](https://github.com/handsontable/handsontable/issues/4118) has been open since 2017 and still reproduces on `develop`.

Both methods are addressed by **prop** — a key for object data, a source index for array data. To build the change list they read the previous value with `dataSource.getAtCell()`, which does not take a prop: it takes a **visual column index** and resolves it with `colToProp()`. The prop was therefore translated a second time.

`colToProp()` returns its argument unchanged when it is not an integer, so **string props were never affected and numeric props always were** — array-shaped data breaks, object data does not. This is the opposite of what the issue thread assumes.

The write was never affected (`applyChanges` uses `datamap.set()`, which addresses by prop), so the value landed in the right cell while the reported old value came from another column. Measured on a 6 × 15 grid with values `r<row>c<col>`:

| Setup | Call | Reported | Correct |
|---|---|---|---|
| `columns: [{data:5},…{data:14}]` | `setDataAtRowProp(1, 5, 'NEW')` | `r1c10` | `r1c5` |
| `columns: [{data:2},{data:0},{data:1}]` | `setDataAtRowProp(1, 0, 'NEW')` | `r1c2` | `r1c0` |
| `manualColumnMove: [4,3,2,1,0]` | `setDataAtRowProp(1, 3, 'NEW')` | `r1c1` | `r1c3` |

`hiddenColumns` was never affected, because hiding does not renumber physical columns.

**Undo then corrupted the data.** The undo plugin is correct — it faithfully wrote back the bad old value it was handed, so an undo replaced the cell with another column's value instead of restoring it:

```
source row 1 BEFORE  [... "r1c4","r1c5","r1c6" ... "r1c10" ...]
setDataAtRowProp([[1, 5, 'NEW']])
  afterChange           [[1,5,"r1c10","NEW"]]     <- old value already wrong
source row 1 AFTER   [... "r1c4","NEW","r1c6" ...]  <- write was always correct
undo()
source row 1 UNDONE  [... "r1c4","r1c10","r1c6" ...]  <- column 5 now holds column 10's value
```

No change was needed in the undo plugin.

### The change

`DataSource` gains `getAtCellByProp()` — `getAtPhysicalCell()` without the `colToProp()` step — and the two prop-addressed call sites use it (`core.ts` `setDataAtRowProp`, `setSourceDataAtCell`). `getAtCell()` is untouched: `getSourceDataAtCell()` documents its column as visual and has its own long-standing TODO.

Two cases in `UndoRedo.spec.js` (`should undo single change`, `should redo single change`) passed **because of** the bug. They called `setDataAtRowProp(0, 0, …)` on object data, where the write created a literal `"0"` key while the old value was read from `name`, so `expect(getDataAtCell(0, 0)).toBe('Timothy')` passed without `name` ever changing. Both now address the object's own key, like every other test in the file, and assert the whole source row so the round trip is real.

The `handsontable-playwright-e2e` skill also gains the port-8123 `reuseExistingServer` trap, which is what made the first run of the new spec exercise another checkout's build. Hit while writing this PR's tests.

- Unit tests added/modified (`*.unit.js`): none — the defect is in how two Core methods address the data source, which needs a real grid with a real `columnIndexMapper`; covered by the E2E spec below.
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`): `tests/e2e/set-data-at-row-prop-old-value.spec.ts` (11 tests), with `tests/fixtures/demo/set-data-at-row-prop.html` and `tests/fixtures/pages/SetDataAtRowPropPage.ts`.
- Modified in the frozen Jasmine suite: `handsontable/src/plugins/undoRedo/__tests__/UndoRedo.spec.js` — the two vacuous cases described above.
- Type tests (`*.types.ts`) updated if public API changed: none — `getAtCellByProp()` is on the private `DataSource` class; no public signature changed.
- For a bug fix — the spec that fails without this fix: all six remapping cases in `set-data-at-row-prop-old-value.spec.ts`. With the two call sites reverted the run is **6 failed / 5 passed**, and the 5 that still pass are exactly the controls (no remap, hidden columns, the two string-prop cases). Verified before committing.

### Commands run

```
$ cd tests && HOT_TEST_PORT=8131 npx playwright test --project=e2e-main set-data-at-row-prop-old-value
11 passed (1.8s)

# same command with the two call sites reverted, to prove the tests are real:
6 failed
5 passed (2.3s)

$ npm run test:e2e --prefix handsontable -- --testPathPattern=UndoRedo
195 specs, 0 failures

$ npm run test:e2e --prefix handsontable -- --testPathPattern="setDataAt|sourceData|dataProvider|valueSetter|valueGetter"
257 specs, 0 failures

```

Check port 8123 before running locally — a second checkout listening there silently serves its own build (`lsof -nP -i :8123 | grep LISTEN`).

### Types of changes

Not a breaking change under the policy — the reported value was wrong, not contractual. Worth calling out in review that the `oldValue` delivered to `beforeChange` / `afterChange` for numeric-prop calls does change, and undo now restores correctly. Users who built a workaround (one commenter on #4118 rewrote `oldValue` in `beforeChange`) will see different values. The changelog entry says so.

### Related issue(s):

1. DEV-2721
2. #4118

Found alongside two siblings, filed separately: DEV-2722 (`sourceDataValidator` reads one cell and blanks another) and DEV-2723 (#7031, `colToProp`/`propToCol` out-of-range docs).

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2721

---

> [!NOTE]
> **Medium Risk**
> Touches core change/undo data paths and changes reported `oldValue` for numeric props under column remapping; behavior is a correctness fix but may differ from workarounds in `beforeChange`.
> 
> **Overview**
> Fixes incorrect **previous values** (and undo corruption) when `setDataAtRowProp()` or `setSourceDataAtCell()` use a **numeric prop** while columns are remapped via `columns[].data` or `manualColumnMove`. Those APIs are prop-addressed, but the old read went through `dataSource.getAtCell()`, which treats its second argument as a **visual column** and runs `colToProp()` again—so hooks saw another column’s value even though the write was correct.
> 
> Adds **`DataSource.getAtCellByProp()`** (direct physical read, no column translation) and switches the two core call sites to use it. **`oldValue` in `beforeChange` / `afterChange` / `afterSetSourceDataAtCell`** now matches the cell being updated; **undo** restores the real prior value. String props were unchanged; numeric props on object rows now report the value at the numeric key (typically `undefined`), aligned with where the write lands.
> 
> Coverage: unit tests for `getAtCellByProp`, Playwright E2E for remap/move/undo scenarios, and tightened `UndoRedo` object-data undo/redo cases that previously masked the bug. Changelog entry and Playwright skill note on port **8123** / `HOT_TEST_PORT` for local E2E.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 35358b135da043107748d51c94d6b45f25e0892f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changelogs/13322.json
+++ b/.changelogs/13322.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Fixed a bug where `setDataAtRowProp()` and `setSourceDataAtCell()` reported the previous value of the wrong column when the property was a number and the columns were moved or remapped through `columns[].data`. The `oldValue` passed to `beforeChange`, `afterChange` and `afterSetSourceDataAtCell` now comes from the addressed cell, and undoing such a change restores the original value instead of overwriting the cell with another column's value. Data addressed by a string property was never affected.",
+  "type": "fixed",
+  "issueOrPR": 13322,
+  "breaking": false,
+  "framework": "none"
+}
--- a/.changelogs/13322.json
+++ b/.changelogs/13322.json
@@ -1,6 +1,6 @@
 {
   "issuesOrigin": "private",
-  "title": "Fixed a bug where `setDataAtRowProp()` and `setSourceDataAtCell()` reported the previous value of the wrong column when the property was a number and the columns were moved or remapped through `columns[].data`. The `oldValue` passed to `beforeChange`, `afterChange` and `afterSetSourceDataAtCell` now comes from the addressed cell, and undoing such a change restores the original value instead of overwriting the cell with another column's value. Data addressed by a string property was never affected.",
+  "title": "Fixed a bug where `setDataAtRowProp()` and `setSourceDataAtCell()` reported the previous value of the wrong column when the property was a number and the columns were moved or remapped through `columns[].data`. The `oldValue` passed to `beforeChange`, `afterChange` and `afterSetSourceDataAtCell` now comes from the addressed cell, and undoing such a change restores the original value instead of overwriting the cell with another column's value. Data addressed by a string property was never affected. A numeric property against object-shaped data now reports the value stored under that numeric key (usually `undefined`), which is where the write itself lands.",
   "type": "fixed",
   "issueOrPR": 13322,
   "breaking": false,
--- a/handsontable/src/plugins/undoRedo/__tests__/UndoRedo.spec.js
+++ b/handsontable/src/plugins/undoRedo/__tests__/UndoRedo.spec.js
@@ -1660,17 +1660,20 @@ describe('UndoRedo', () => {
     }
 
     describe('undo', () => {
+      // Addressed by the object's own key. Passing `0` here used to write a literal `"0"` key while
+      // the old value was read from `name`, so the assertions passed without `name` ever changing.
       it('should undo single change', async() => {
         handsontable({
           data: createObjectData()
         });
 
-        await setDataAtRowProp(0, 0, 'Pearce');
+        await setDataAtRowProp(0, 'name', 'Pearce');
 
-        expect(getDataAtRowProp(0, 0)).toBe('Pearce');
+        expect(getDataAtRowProp(0, 'name')).toBe('Pearce');
 
         getPlugin('undoRedo').undo();
         expect(getDataAtCell(0, 0)).toBe('Timothy');
+        expect(getSourceDataAtRow(0)).toEqual({ name: 'Timothy', surname: 'Dalton' });
       });
 
       it('should undo creation of a single row', async() => {
@@ -2108,14 +2111,15 @@ describe('UndoRedo', () => {
     });
 
     describe('redo', () => {
+      // Addressed by the object's own key, for the same reason as the matching `undo` test above.
       it('should redo single change', async() => {
         handsontable({
           data: createObjectData()
         });
 
-        await setDataAtRowProp(0, 0, 'Pearce');
+        await setDataAtRowProp(0, 'name', 'Pearce');
 
-        expect(getDataAtRowProp(0, 0)).toBe('Pearce');
+        expect(getDataAtRowProp(0, 'name')).toBe('Pearce');
 
         getPlugin('undoRedo').undo();
 
@@ -2123,7 +2127,8 @@ describe('UndoRedo', () => {
 
         getPlugin('undoRedo').redo();
 
-        expect(getDataAtRowProp(0, 0)).toBe('Pearce');
+        expect(getDataAtRowProp(0, 'name')).toBe('Pearce');
+        expect(getSourceDataAtRow(0)).toEqual({ name: 'Pearce', surname: 'Dalton' });
       });
 
       it('should redo creation of a single row', async() => {
--- a/tests/e2e/set-data-at-row-prop-old-value.spec.ts
+++ b/tests/e2e/set-data-at-row-prop-old-value.spec.ts
@@ -18,8 +18,8 @@ import { SetDataAtRowPropPage } from '../fixtures/pages/SetDataAtRowPropPage';
 test.describe('setDataAtRowProp old value', () => {
   let grid: SetDataAtRowPropPage;
 
-  test.beforeEach(async ({ page, theme }) => {
-    grid = new SetDataAtRowPropPage(page, theme);
+  test.beforeEach(async ({ page, theme, bundle }) => {
+    grid = new SetDataAtRowPropPage(page, theme, bundle);
     await grid.goto();
   });
 
@@ -32,6 +32,7 @@ test.describe('setDataAtRowProp old value', () => {
     });
 
     expect(result.oldValue).toBe('r1c5');
+    expect(result.reportedProp).toBe(5);
     expect((result.after as unknown[])[5]).toBe('NEW');
   });
 
@@ -102,8 +103,20 @@ test.describe('setDataAtRowProp old value', () => {
     });
 
     expect(result.oldValue).toBe('r1c3');
+    expect(result.reportedProp).toBe('p3');
     expect((result.after as Record<string, unknown>).p3).toBe('NEW');
     expect(result.undone).toEqual(result.before);
+  });
+
+  test('reads the numeric key a numeric prop writes to when the data is object-shaped', async () => {
+    // A numeric prop has no meaning for object data: the write lands on the literal `0` key, so the
+    // read has to land there too. It used to go through `colToProp(0)`, report `p0` as the old value
+    // and let undo overwrite `p0` with it, even though `p0` was never touched.
+    const result = await grid.setDataAtRowProp({ dataKind: 'object', prop: 0 });
+
+    expect(result.oldValue).toBeUndefined();
+    expect((result.after as Record<string, unknown>).p0).toBe('r1c0');
+    expect((result.after as Record<string, unknown>)[0]).toBe('NEW');
   });
 
   test('keeps string props working when columns[].data selects a subset', async () => {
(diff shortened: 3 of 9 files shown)


qunabu:
Task linked: [DEV-2721 setDataAtRowProp reports the wrong oldValue for numeric props, and undo then corrupts the data](https://app.clickup.com/t/9015210959/DEV-2721)
Task linked: [DEV-2722 sourceDataValidator reads one cell and blanks a different one when columns are moved or remapped](https://app.clickup.com/t/9015210959/DEV-2722)

tests/e2e/set-data-at-row-prop-old-value.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+ * array-shaped data. Undo then wrote the wrong old value back into the source data, which is the
+ * costliest symptom and the reason the undo assertions below matter.
+ *
+ * Cell values are `r<row>c<col>`, so a wrong old value names the column it actually came from.
+ */
+test.describe('setDataAtRowProp old value', () => {
+  let grid: SetDataAtRowPropPage;
+
+  test.beforeEach(async ({ page, theme }) => {
+    grid = new SetDataAtRowPropPage(page, theme);
```
demtario: **high** — this `beforeEach` (and the matching one at line 125) destructures only `{ page, theme }`, dropping `bundle`. `SetDataAtRowPropPage`'s constructor defaults `bundle = 'umd'`, so the `-min` CI legs of this spec load `handsontable.js` instead of `handsontable.full.min.js` — they silently re-test the same build as the `umd` legs, per the bundle-threading rule in `tests/AGENTS.md` ("fixture allowlist → page-object constructor → goto() query params → spec destructure. Miss one link and a leg silently tests the wrong build."). Fix: `async ({ page, theme, bundle }) => { grid = new SetDataAtRowPropPage(page, theme, bundle); }` in both blocks.

handsontable/src/plugins/undoRedo/__tests__/UndoRedo.spec.js
```diff
@@ -1660,17 +1660,20 @@ describe('UndoRedo', () => {
...
     describe('undo', () => {
+      // Addressed by the object's own key. Passing `0` here used to write a literal `"0"` key while
+      // the old value was read from `name`, so the assertions passed without `name` ever changing.
       it('should undo single change', async() => {
         handsontable({
           data: createObjectData()
         });
 
-        await setDataAtRowProp(0, 0, 'Pearce');
+        await setDataAtRowProp(0, 'name', 'Pearce');
```
demtario, line 1670: **medium** — before this fix, `setDataAtRowProp(0, 0, ...)` against object data exercised the buggy path (per the comment above). Switching this test — and the matching redo test at line 2120 — to address by `'name'` is the right call for these two tests, but it removes the only coverage of "numeric prop against object data." Under the new `getAtCellByProp`, that shape hits `dataRow[0]` on an object row and returns `undefined` — a real, newly-defined behavior with no test anywhere now (the E2E fixture's object-data scenarios only use string props). Suggest one case asserting the `undefined` result.

.changelogs/13322.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Fixed a bug where `setDataAtRowProp()` and `setSourceDataAtCell()` reported the previous value of the wrong column when the property was a number and the columns were moved or remapped through `columns[].data`. The `oldValue` passed to `beforeChange`, `afterChange` and `afterSetSourceDataAtCell` now comes from the addressed cell, and undoing such a change restores the original value instead of overwriting the cell with another column's value. Data addressed by a string property was never affected.",
```
demtario: **low** — this states "Data addressed by a string property was never affected," true for array data, but doesn't mention that a numeric prop against **object** data now returns `undefined` instead of the old (also wrong) value — see the `UndoRedo.spec.js` comment above. Minor completeness gap, not blocking.

tests/fixtures/pages/SetDataAtRowPropPage.ts
```diff
@@ -0,0 +1,86 @@
...
+  /** Visual row index. Defaults to 1. */
+  row?: number;
+}
+
+/**
+ * What the fixture reports back after driving one scenario.
+ */
+export interface PropResult {
+  oldValue: unknown;
+  reportedProp?: unknown;
```
demtario, line 27: **nit** — `reportedProp` is computed by the fixture (`set-data-at-row-prop.html`) and typed here, but no test asserts it. Given the bug class is prop mis-resolution, asserting the echoed prop matches the prop passed in would be a cheap, on-target addition — otherwise drop the field.

demtario (commented):
4 inline notes below. Two more findings have no changed line to anchor to:

- **medium** — the `setSourceDataAtCell` describe block (`tests/e2e/set-data-at-row-prop-old-value.spec.ts`) only covers the `columns[].data` offset case and a no-remap control; missing a `manualColumnMove` case, which reaches the prop-translation bug through a different mechanism (`toPhysicalColumn`) than `columns[].data` (`colToPropCache`). The second swapped call site (`core.ts:4406`, inside `setSourceDataAtCell`) is under-covered for that path.
- **low** — no direct Jest unit test for the new `DataSource.getAtCellByProp()` (`handsontable/src/dataMap/dataSource.ts`). It's pure logic with no `columnIndexMapper` dependency, so the "needs a real grid" justification for E2E-only coverage doesn't apply to this method itself. Sibling methods (`getAtCell`, `setAtCell`) have unit coverage in `dataSource.unit.js` with a lightweight mock that would cover this cheaply, including the object-data-numeric-prop shape noted in the `UndoRedo.spec.js` comment.

tests/e2e/set-data-at-row-prop-old-value.spec.ts
budnix: You're right — `bundle` is threaded through both blocks now, and I added the `manualColumnMove` case for `setSourceDataAtCell` in the same file (35358b135d). Rebuilding without the fix now turns the `-min` leg red, so it really is loading its own bundle.

handsontable/src/plugins/undoRedo/__tests__/UndoRedo.spec.js
budnix: Good catch — that case now lives in the Playwright spec instead (object data, `prop: 0`), asserting `undefined` plus that the write and the read both land on the literal `0` key. I also gave `getAtCellByProp()` its own unit block in `dataSource.unit.js`, six cases — both in 35358b135d.

.changelogs/13322.json
budnix: Fair point — added a sentence in 35358b135d. I worded it as the value under that numeric key (usually `undefined`), since that is where the write lands too.

tests/fixtures/pages/SetDataAtRowPropPage.ts
budnix: Good spot — asserted in 35358b135d, `toBe(5)` in the offset test and `toBe('p3')` in the object-data test, so the field earns its place.

demtario (approved):

PR #13323, opened 2026-09-01
DEV-2722: Blank the source cell the validator actually read

budnix:
### Context

The PR fixes `sourceDataValidator` clearing the wrong cell. Found while analysing [#4118](https://github.com/handsontable/handsontable/issues/4118) — the same read/write addressing mismatch, but this one destroys data on its own rather than only misreporting a value. `sourceDataValidator` ships in released 18.0.0, so this affects released code.

`sourceDataValidator.ts` addressed the same cell two different ways:

- It **read** the value with `dataSource.getAtCell(row, physicalColumn)`. `getAtCell()` expects a **visual** column index and resolves it through `colToProp()`, so the physical column was translated as if it were visual.
- When the value was invalid and `allowInvalid: false`, it **blanked** with `dataSource.setAtCell(physicalRow, physicalColumn, null)` — no translation at all.

Whenever visual and physical disagree, the validator judged one source cell and nulled a different one. Measured on a 3 × 6 grid with a validator that rejects everything:

| Setup | Validator on column | Value it was handed | Cell it nulled |
|---|---|---|---|
| no remap (control) | 2 | `r0c2` | index 2 — correct |
| `manualColumnMove: [4,3,2,1,0]` | 0 | `r0c4` | index 0 |
| `columns: [{data:2},{data:3},{data:4},{data:5}]` | 0 | `r0c2` | index 0 |

In the last two rows the grid silently wiped a column the validator never inspected.

### The change

The column's **source address** is resolved once from its visual index (`resolveProp()`), and that single address is used for both the read and the blanking write. The physical column index is kept only for the aggregated warning, which reports coordinates rather than addressing data.

Both scan paths are covered — the batched one (a row-independent validator, one meta per column) carries the resolved prop on its `ColumnValidator` entry; the per-cell one resolves it alongside the meta it already computes.

Note the corrected expectation in the moved-columns case: `columns[0]` is the meta of **physical** column 0, whose value lives at source index 0 however the move shuffles the display. The old code reached source 4 — the column that merely happens to be shown first — and then blanked index 0 anyway, so neither half was right.

- Unit tests added/modified (`*.unit.js`): `handsontable/src/dataMap/__tests__/sourceDataValidator.unit.js` — three new cases pinning the address on both scan paths (batched and per-cell) and for string props, plus the existing mock updated to model the prop-addressed read.
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`): `tests/e2e/source-data-validator-address.spec.ts` (3 tests), with `tests/fixtures/demo/source-data-validator-address.html` and `tests/fixtures/pages/SourceDataValidatorPage.ts` — a real grid with a rejecting validator and `allowInvalid: false`, asserting both the value handed to the validator and which source cell ends up `null`.
- Type tests (`*.types.ts`) updated if public API changed: none — the change is internal to `sourceDataValidator.ts`; no public signature changed.
- For a bug fix — the spec that fails without this fix: the three new unit cases. Reverting the blanking address alone turns the run to **3 failed / 17 passed**, and the three failures are exactly those cases. Verified before committing.

### Commands run

```
$ npm run test:unit --prefix handsontable -- --testPathPattern=sourceDataValidator
Tests: 20 passed, 20 total

# same command with the blanking address reverted, to prove the tests are real:
Tests: 3 failed, 17 passed, 20 total

$ npm run test:unit --prefix handsontable -- --testPathPattern="dataMap/__tests__"
Test Suites: 3 passed, 3 total
Tests:       40 passed, 40 total

$ npm run test:e2e --prefix handsontable -- --testPathPattern="sourceData|validator|Validator"
241 specs, 0 failures

$ cd tests && HOT_TEST_PORT=8131 npx playwright test --project=e2e-main source-data-validator-address
3 passed (871ms)

```

Check port 8123 before running locally — a second checkout listening there silently serves its own build (`lsof -nP -i :8123 | grep LISTEN`).

### Types of changes

### Related issue(s):

1. DEV-2722

Sibling of DEV-2721 (PR #13322), which fixes the same addressing mistake in `setDataAtRowProp()` and `setSourceDataAtCell()`. The two PRs touch different files and can land in either order.

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2722

---

> [!NOTE]
> **Medium Risk**
> Touches source-data validation and direct writes to user data on load when `allowInvalid` is false; scope is narrow but incorrect addressing could still corrupt grids with column moves or custom `columns[].data`.
> 
> **Overview**
> Fixes a **data-loss bug** in `sourceDataValidator` when `allowInvalid` is `false`: validation could **read** one source cell but **clear** another if columns were moved or remapped via `columns[].data` (or accessor functions).
> 
> **Runtime behavior:** Column mappings now carry each column’s **source address** from `colToProp(visualColumn)`. Both batched and per-cell scan paths read values with `getAtCellByProp(row, sourceColumn, …)` and still blank with `setAtCell(row, sourceColumn, null)`, so read and write use the same address. Physical column index is only used for warning metadata.
> 
> **`DataSource`:** `getAtCellByProp` accepts an optional pre-resolved `dataRow` so validators resolve `modifyRowData` **once per row** (not per column).
> 
> **Tests & docs:** New unit coverage for remap/move/accessor cases and `modifyRowData` call counts; Playwright fixture/spec for end-to-end regression; changelog entry; `tests/AGENTS.md` notes on waiting for the bundle before fixture `ready`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3c988f84efd8c6b12ff723c2f8b2b65516543c1b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2721 setDataAtRowProp reports the wrong oldValue for numeric props, and undo then corrupts the data](https://app.clickup.com/t/9015210959/DEV-2721)
Task linked: [DEV-2722 sourceDataValidator reads one cell and blanks a different one when columns are moved or remapped](https://app.clickup.com/t/9015210959/DEV-2722)

demtario (commented):
3 inline notes below.

- No test exercises a function-valued `columns[].data` accessor (`DataAccessorFn`) through the read+blank path. `resolveProp()`'s declared return type includes `DataAccessorFn`, and `DataSource#getAtPhysicalCell`/`#setAtCell` both branch on `typeof column === 'function'`, but only the numeric-remap and string-prop cases are covered in `sourceDataValidator.unit.js` / `source-data-validator-address.spec.ts`.

handsontable/src/dataMap/sourceDataValidator.ts
```diff
@@ -180,9 +200,12 @@ function validatePerCell(
         continue;
       }
 
-      const value = dataSource.getAtCell(row, col);
+      // The prop is the source address of this physical column; `getAtCell()` would resolve `col`
+      // as a *visual* index and read another column whenever the two differ.
+      const prop = resolveProp(hotInstance, visualColumn);
+      const value = dataSource.getAtPhysicalCell(row, prop, dataSource.modifyRowData(row));
```
demtario: **medium**: This (and the equivalent call at line 246 in `validateBatched`) reach into two `@private`-tagged `DataSource` methods (`getAtPhysicalCell`, `modifyRowData`) from outside the class — this is the first external caller of either. The two-line combination `dataSource.getAtPhysicalCell(row, prop, dataSource.modifyRowData(row))` is also literally `getAtCell()`'s body inlined and duplicated at both sites.

Suggest adding a small `DataSource` method, e.g. `getAtSourceProp(row, prop) { return this.getAtPhysicalCell(row, prop, this.modifyRowData(row)); }`, and calling that once from each site here. Keeps the `@private` internals inside their own class and removes the duplication.

handsontable/src/dataMap/sourceDataValidator.ts
```diff
@@ -218,10 +241,11 @@ function validateBatched(
     }
 
     for (let i = 0; i < columns.length; i += 1) {
-      const { physicalColumn, cellMeta } = columns[i];
-      const value = dataSource.getAtCell(row, physicalColumn);
+      const { physicalColumn, prop, cellMeta } = columns[i];
+      // Read by the column's source address, for the same reason as in `validatePerCell()`.
+      const value = dataSource.getAtPhysicalCell(row, prop, dataSource.modifyRowData(row));
```
demtario: **low**: `dataSource.modifyRowData(row)` only depends on `row`, but is called once per column inside this loop rather than once per row. Not a regression — the pre-fix `getAtCell()` call already invoked `modifyRowData` internally on every call — but since this line is already being rewritten here, it's a low-cost opportunity to hoist `const dataRow = dataSource.modifyRowData(row);` above the loop and pass `dataRow` into `getAtPhysicalCell` for every column, avoiding `columns.length`x redundant `modifyRowData` hook invocations per row.

handsontable/src/dataMap/sourceDataValidator.ts
budnix: Good point — added `DataSource#getAtSourceProp` in fc669d5dba and both sites go through it, plus the accessor-function tests your body note asked for. `modifyRowData` still runs from the validator, once per row, because of your other comment.

handsontable/src/dataMap/sourceDataValidator.ts
budnix: Nice catch — hoisted it in fc669d5dba, so it runs once per row on both paths, and the per-cell path now skips it entirely for rows it never reads.

demtario (dismissed):
Conflicts

budnix:
@demtario Good spot — merged develop in 4ff822877e. #13073 had landed the same fix in the meantime, so the validator now keeps its `collectColumnMappings` shape and both reads go through `getAtSourceProp` as you asked.

demtario (approved):

PR #5945, opened 2019-04-11
New architecture for record translator

wszymanski:
### Problems
As mentioned in https://github.com/handsontable/handsontable/issues/5112 even simple connections between plugins make them not working properly (part of them works, but they are handled by workarounds). From now, also a few working simultaneously plugins should work properly.

### Context
Introducing `global index mapper`, which manages the index mappings provided by smaller maps called `index map(s)`.

1. Global index mapper may be used by `rowIndexMapper` or `columnIndexMapper` properties on the Handsontable instance.

https://github.com/handsontable/handsontable/blob/ebf7764cab9608f6cec2207e9317e2f17e149040/src/plugins/manualRowMove/manualRowMove.js#L181

2. The mapper includes three main objects for storing indexes:

a) map for the sequence of indexes. It is registered by default and may be used from API methods like `getIndexesSequence`, `setIndexesSequence`, `getNumberOfIndexes` etc.

https://github.com/handsontable/handsontable/blob/24f352228743779d3cd6deaa57f6ea50e64e6e36/src/translations/indexMapper.js#L25-L30

https://github.com/handsontable/handsontable/blob/ebf7764cab9608f6cec2207e9317e2f17e149040/src/plugins/columnSorting/columnSorting.js#L590-L596

b) collection for different skip maps. Indexes marked as skipped in any map won't be rendered. The API methods like `getNotSkippedIndexes`, `getNotSkippedIndexesLength`, `isSkipped` etc. return information about all not rendered rows, basing on values inside all maps stored in the collection.

https://github.com/handsontable/handsontable/blob/24f352228743779d3cd6deaa57f6ea50e64e6e36/src/translations/indexMapper.js#L36

https://github.com/handsontable/handsontable/blob/ebf7764cab9608f6cec2207e9317e2f17e149040/src/plugins/manualColumnMove/manualColumnMove.js#L235

c) collection for another kind of mappings.

https://github.com/handsontable/handsontable/blob/24f352228743779d3cd6deaa57f6ea50e64e6e36/src/translations/indexMapper.js#L42

3. There are three main kinds of index maps which may be registered in the collections. 

User may work with:

- `VisualIndexToPhysicalIndexMap` - which provides mappings from visual index to physical index and update the physical index on remove/add row or column action.
- `PhysicalIndexToValueMap` - which provides mappings from physical index to any value and doesn't update the value on remove/add row or column action.
[...]

PR #13325, opened 2026-09-01
DEV-2723: Document colToProp/propToCol out-of-range return

budnix:
### Context

The PR corrects what the API reference says `colToProp()` and `propToCol()` return when the index or property matches no column, and adds the specs that pin it. It is the patch-level half of [#7031](https://github.com/handsontable/handsontable/issues/7031); the behavior change that issue asks for is breaking and is not attempted here.

Neither method signals an unknown column. Both hand the argument straight back — `colToProp(999)` is `999`, `propToCol(999)` is `999` — while `toVisualColumn()` and the other index translators return `null` in the same situation. The API reference did not say so, so the asymmetry was invisible to anyone reading the docs.

It used to. PR [#5945](https://github.com/handsontable/handsontable/pull/5945) introduced the pass-through deliberately, in the same commit that gave the index mappers their `null`, and documented it as `@returns {String|Number} Column property, physical column index or passed argument.` That sentence was lost in a later rewrite while the code kept the behavior.

**`null` is still a possible result**, and the docs now say so rather than implying an argument always comes back:

| Case | Result |
|---|---|
| `propToCol('name')` where that column is trimmed, and the property is cached | `null` — the cached branch returns `toVisualColumn()` with no fallback |
| `propToCol(1)` where that column is trimmed, on plain array data | `1` — the uncached branch falls back to the passed index instead |
| `colToProp(0)` with `columns: [{ data: null }]` | `null` — `isDefined(null)` is `true`, so the cache hit wins, on a valid index |
| `colToProp(null)` | `null` — the non-integer echo, which `undoRedo/actions/removeRow.ts` depends on |

All measured on the built bundle. The consequence for callers is in the docs: validate before use, because `Number.isInteger()` alone lets the second row through and a `countCols()` comparison alone lets `null` through (`null` compares as `0`).

Rows 1 and 2 are the same question with two different answers, and the second is wrong rather than merely unhelpful — the index it returns addresses a different column. That is a defect in its own right, filed separately as **DEV-2726**; this PR documents the split as it stands today and pins it with a test, so the fix has a tripwire that forces the docs to be simplified along with the code.

`propToCol()` also gets a note that its TypeScript declaration narrows the result to `number`, so neither a returned property name nor `null` is visible to the type checker. The same caveat now sits on the `CoreMethods` declaration in `core/types.ts`, which is what a TypeScript consumer actually resolves to.

**No runtime code changed.** The diff is two JSDoc blocks, one interface comment, and specs.

### Why not fix the behavior here

Verified against the code and the history rather than the issue label, and it does not qualify as a patch:

- The pass-through was deliberate, added by #5945 together with its documentation.
- #7031 was filed by the author of #5945 — this is the team revisiting its own design, filed as `enhancement` + `Breaking change`, not as a bug.
- Core depends on it. `applyChanges()` grows the grid with `while (Number(datamap.propToCol(prop)) > countCols() - 1)`; `Number(null)` is `0`, so a `null`-returning `propToCol()` silently stops auto column growth. Measured: `setDataAtCell(0, 6, …)` on a 3-column array grid grows it to 7 columns, and pasting past the edge of a 2-column grid grows it to 4.
- `plugins/undoRedo/actions/removeRow.ts:233-240` uses the echo as a bail-out signal, with a comment saying so. Roughly 78 call sites across 26 files would each need a null-safety pass.

Note that #7031 asks about indexes **beyond the table boundaries**, which is a different case from a trimmed column the mapper still knows about. DEV-2726 covers the second one and is separable — `columnIndexMapper.getNumberOfIndexes()` tells them apart, so fixing it leaves the pass-through above untouched.

- Unit tests added/modified (`*.unit.js`): none — these two methods are covered by the E2E specs below, which run against a real grid and a real `columnIndexMapper`.
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`): none — this PR touches no runtime behavior, so the coverage it needs is the two existing Jasmine specs for the methods themselves (`handsontable/src/__tests__/core/colToProp.spec.js`, `handsontable/src/__tests__/core/propToCol.spec.js`), edited in place as the frozen-suite rules allow.
- Type tests (`*.types.ts`) updated if public API changed: none — `core/types.ts` gained a comment; no signature changed.
- For a bug fix — the spec that fails without this fix: `colToProp.spec.js` › "should return proper value after calling the function when columns was reorganized (data is array of objects)" contained `expect(propToCol(10)).toBe(10)` inside `describe('Core.colToProp')` — a copy-paste that left `colToProp` out of range asserted **nowhere**, so the contract this PR publishes could have changed with every spec still green. Now `expect(colToProp(10)).toBe(10)`, plus dedicated cases for the pass-through, the `null` echo, the `{ data: null }` column, and both sides of the trimmed-column split.

Also replaced: the two `// I'm not sure if this should return result like that by design.` comments beside those assertions, with what the archaeology settled — it is by design, and `applyChanges()` depends on it.

### Commands run

```
$ npm run test:e2e --prefix handsontable -- --testPathPattern="colToProp|propToCol"
17 specs, 0 failures

$ npm run test:types --prefix handsontable
✓ tsc -p ./test/types

$ npm run build --prefix handsontable
✓ node scripts/prepare-package-for-publish.mjs
```

[skip changelog]

### Types of changes

The template has no documentation-only option. This is filed as a bug fix because two real defects are corrected — an API reference that stated the wrong return contract, and a spec asserting a different method than the one it was written for. No runtime behavior changes.

### Related issue(s):

1. DEV-2723
2. DEV-2726 (the trimmed-column defect this PR documents but does not fix)
3. #7031

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

- [x] My change requires a change to the documentation. This PR is that change — the API reference is generated from these JSDoc blocks.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2723

---

> [!NOTE]
> **Low Risk**
> Documentation and test-only changes with no production code paths modified.
> 
> **Overview**
> **Documents** how `colToProp()` and `propToCol()` behave when the index or property does not map to a column: both **echo the argument** (by design since #5945), unlike `toVisualColumn()` / other mappers that return `null`. The JSDoc in `core.ts` and a note on `CoreMethods` in `core/types.ts` also spell out when **`null` is a real result** (trimmed cached properties, `{ data: null }` columns, non-integer/`null` echoes) and warn callers to validate results.
> 
> **Tests** expand `colToProp.spec.js` and `propToCol.spec.js`: fix a copy-paste that asserted `propToCol(10)` inside the `colToProp` suite, add cases for out-of-range pass-through, non-integer echo, and trimmed-column behavior, and **pin known defect DEV-2726** (array data + trimmed physical index returns the wrong index instead of `null`).
> 
> **No runtime implementation changes** — behavior is unchanged; docs and specs are aligned with the current bundle.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a6c49de1e59ce199e470a1939deb99b0571511b6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/handsontable/src/__tests__/core/propToCol.spec.js
+++ b/handsontable/src/__tests__/core/propToCol.spec.js
@@ -80,7 +80,10 @@ describe('Core.propToCol', () => {
     expect(propToCol('name')).toBe(null);
   });
 
-  it('should hand back a bare physical index whose column is trimmed (array data)', async() => {
+  // Not a specification of correct behavior — this pins a defect so the JSDoc and the code cannot
+  // drift apart. See the comment on the assertion below.
+  it('pins known defect DEV-2726: a bare physical index whose column is trimmed comes back ' +
+    'unchanged instead of `null` (array data)', async() => {
     const hot = handsontable({
       data: [
         ['a0', 'a1', 'a2', 'a3'],
--- a/handsontable/src/__tests__/core/propToCol.spec.js
+++ b/handsontable/src/__tests__/core/propToCol.spec.js
@@ -33,7 +33,9 @@ describe('Core.propToCol', () => {
     columnIndexMapper().setIndexesSequence([4, 3, 2, 1, 0]);
 
     expect(propToCol(0)).toBe(4);
-    expect(propToCol(10)).toBe(10); // I'm not sure if this should return result like that by design.
+    // By design: an unmatched property comes straight back (introduced with the index mappers in
+    // #5945, and `applyChanges()` depends on it to grow the grid past the last column).
+    expect(propToCol(10)).toBe(10);
   });
 
   it('should return proper value after calling the function when columns was reorganized (data is array of objects)', async() => {
@@ -51,6 +53,30 @@ describe('Core.propToCol', () => {
 
     expect(propToCol('id')).toBe(3);
     expect(propToCol(0)).toBe(3);
-    expect(propToCol(10)).toBe(10); // I'm not sure if this should return result like that by design.
+    // By design: an unmatched property comes straight back (introduced with the index mappers in
+    // #5945, and `applyChanges()` depends on it to grow the grid past the last column).
+    expect(propToCol(10)).toBe(10);
+  });
+
+  it('should return `null` for a property whose column is trimmed', async() => {
+    const hot = handsontable({
+      data: [
+        { id: 1, name: 'Ted', lastName: 'Right' },
+        { id: 2, name: 'Frank', lastName: 'Honest' },
+      ]
+    });
+
+    expect(propToCol('name')).toBe(1);
+
+    const trimmingMap = hot.columnIndexMapper.createAndRegisterIndexMap('spec-trim', 'trimming');
+
+    trimmingMap.setValueAtIndex(1, true);
+
+    await render();
+
+    // A cached property resolves through `toVisualColumn()`, which has no visual index left to
+    // return. So `null` is a real result here — the unmatched-property echo does not cover it, and
+    // a `result < countCols()` guard would let this `null` through as column 0.
+    expect(propToCol('name')).toBe(null);
   });
 });
--- a/handsontable/src/__tests__/core/propToCol.spec.js
+++ b/handsontable/src/__tests__/core/propToCol.spec.js
@@ -79,4 +79,32 @@ describe('Core.propToCol', () => {
     // a `result < countCols()` guard would let this `null` through as column 0.
     expect(propToCol('name')).toBe(null);
   });
+
+  it('should hand back a bare physical index whose column is trimmed (array data)', async() => {
+    const hot = handsontable({
+      data: [
+        ['a0', 'a1', 'a2', 'a3'],
+        ['b0', 'b1', 'b2', 'b3'],
+      ]
+    });
+
+    const trimmingMap = hot.columnIndexMapper.createAndRegisterIndexMap('spec-trim-array', 'trimming');
+
+    trimmingMap.setValueAtIndex(1, true);
+
+    await render();
+
+    expect(hot.toVisualColumn(1)).toBe(null);
+
+    // KNOWN DEFECT, tracked as DEV-2726. Plain array data caches no properties, so this takes the
+    // uncached branch, which falls back to the passed index instead of returning `null` like the
+    // cached branch above. The index it returns addresses a DIFFERENT column — visual 1 is physical
+    // 2 here — so the caller gets a wrong answer rather than an unknown one.
+    //
+    // Pinned to keep the documented behavior and the code in step. When DEV-2726 lands this
+    // expectation becomes `null`, and the `propToCol` JSDoc in `core.ts` has to be simplified with
+    // it.
+    expect(propToCol(1)).toBe(1);
+    expect(hot.toPhysicalColumn(1)).toBe(2);
+  });
 });
--- a/handsontable/src/core.ts
+++ b/handsontable/src/core.ts
@@ -4094,17 +4094,20 @@ export default function Core(
    * Returns the property name that corresponds with the given column index.
    * If the data source is an array of arrays, it returns the columns index.
    *
-   * When the column index does not point at an existing column, the method returns the argument you
-   * passed, unchanged. It does not return `null`, unlike {@link Core#toVisualColumn} and the other
-   * index translators. Check the index against {@link Core#countCols} first if you need to tell an
-   * out-of-range call apart from a real result.
+   * When the column index points at no existing column, the method hands the argument back
+   * unchanged. It does not signal an unknown column, so the result on its own never tells you
+   * whether that column exists.
+   *
+   * The result can also be `null`, in two cases: an argument that is not an integer comes straight
+   * back, and a column declared as `{ data: null }` resolves to `null` for an index that is
+   * perfectly valid. Test the result before you use it as a property name.
    *
    * @memberof Core#
    * @function colToProp
    * @param {number} column Visual column index.
-   * @returns {string|number} Column property, physical column index, or the passed argument. When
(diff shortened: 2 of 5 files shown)


qunabu:

qunabu:
Task linked: [DEV-2726 propToCol returns a wrong visual column for a trimmed column, and that also grows the grid spuriously](https://app.clickup.com/t/9015210959/DEV-2726)

handsontable/src/core.ts
```diff
@@ -4108,10 +4116,28 @@ export default function Core(
   /**
    * Returns column index that corresponds with the given property.
    *
+   * When the property matches no column, the method hands the argument back unchanged, so the
+   * result on its own never tells you whether that column exists.
+   *
+   * The result can also be `null`, and for a **trimmed** column which of the two you get depends on
+   * how the property is declared. A property held in the column cache – object data, or one named
+   * by a `columns[].data` entry – resolves through {@link Core#toVisualColumn} and comes back
```
demtario: **High:** this new JSDoc uses `{@link Core#toVisualColumn}` and `{@link Core#countCols}` (also line 4129). This repo's JSDoc convention mandates `[[LINK]]` syntax, not `{@link ...}`. Suggest: `[[Core#toVisualColumn]]` / `[[Core#countCols]]`.

handsontable/src/core.ts
```diff
@@ -4108,10 +4116,28 @@ export default function Core(
...
+   * `null` compares as `0`.
+   *
+   * The TypeScript declaration narrows the result to `number`, so neither a returned property name
+   * nor `null` is visible to the type checker.
+   *
    * @memberof Core#
    * @function propToCol
-   * @param {string|number} prop Property name or physical column index.
-   * @returns {number} Visual column index.
+   * @param {string|number|Function} prop Property name, physical column index, or a `columns[].data`
```
demtario, line 4140: **Medium:** `@param {string|number|Function} prop` documents the parameter as accepting a `columns[].data` accessor function, but the actual TS signature stays `prop: string | number` (same here and in `core/types.ts:125`) — `Function` isn't accepted and doesn't type-check. A caller following this doc and passing an accessor gets a compile error. Suggest narrowing the `@param` back to `{string|number}`, or adding a caveat sentence symmetric to the one already written for the return type's narrowing.

handsontable/src/core.ts
```diff
@@ -4094,12 +4094,20 @@ export default function Core(
    * Returns the property name that corresponds with the given column index.
    * If the data source is an array of arrays, it returns the columns index.
    *
+   * When the column index points at no existing column, the method hands the argument back
+   * unchanged. It does not signal an unknown column, so the result on its own never tells you
+   * whether that column exists.
+   *
+   * The result can also be `null`, in two cases: an argument that is not an integer comes straight
```
demtario, line 4101: **Low:** this prose says a non-integer argument comes back unchanged, but the `@param {number} column` tag two lines below stays untouched (still typed as plain `number`). Runtime behavior described is accurate — just a minor doc inconsistency between the `@param` type and the prose.

handsontable/src/__tests__/core/propToCol.spec.js
```diff
@@ -51,6 +53,58 @@ describe('Core.propToCol', () => {
...
+
+    // KNOWN DEFECT, tracked as DEV-2726. Plain array data caches no properties, so this takes the
+    // uncached branch, which falls back to the passed index instead of returning `null` like the
+    // cached branch above. The index it returns addresses a DIFFERENT column — visual 1 is physical
+    // 2 here — so the caller gets a wrong answer rather than an unknown one.
+    //
+    // Pinned to keep the documented behavior and the code in step. When DEV-2726 lands this
+    // expectation becomes `null`, and the `propToCol` JSDoc in `core.ts` has to be simplified with
+    // it.
+    expect(propToCol(1)).toBe(1);
```
demtario, line 110: **High:** this test pins a self-acknowledged "KNOWN DEFECT, tracked as DEV-2726" as a passing assertion (`expect(propToCol(1)).toBe(1)`, which the comment itself says addresses the wrong column). This conflicts with `handsontable/.ai/TESTING.md:286`: "a test that executes it while asserting nothing, or asserting the *buggy* output, is worse than none." Consider dropping the specific-value assertion (the JSDoc change alone documents the defect) or reframing it as an explicit regression tripwire rather than a spec of correct behavior.

demtario (commented):
4 inline notes below.

**Off-diff:** `handsontable/src/__tests__/core/colToProp.spec.js:52` still carries the stale "I'm not sure if this should return result like that by design." comment on an equivalent out-of-range case, two tests above the one this PR rewrote with the same explanation. Not blocking, but worth a follow-up cleanup for consistency.

handsontable/src/core.ts
budnix: Good spot — both are `[[Core#toVisualColumn]]` and `[[Core#countCols]]` now, fixed in a6c49de1e5.

handsontable/src/core.ts
budnix: Good point — kept `{string|number|Function}` since the accessor does resolve at runtime, and added the parameter-narrowing caveat in a6c49de1e5.

handsontable/src/core.ts
budnix: Fair point — added the echo to the `@param` description in a6c49de1e5, the type tag stays `{number}`.

handsontable/src/__tests__/core/propToCol.spec.js
budnix: You're right — renamed the test in a6c49de1e5 so it reads as a defect pin, not a spec.

demtario (approved):

PR #13326, opened 2026-09-01
DEV-2725: Fix the release workflow's prod-docs branch and merge-PR fallback

budnix:
### Context

The PR fixes the release workflow's docs branch and its merge-to-develop fallback. The 18.1.0 release ([run 33503319432](https://github.com/handsontable/handsontable/actions/runs/33503319432)) reported two failures on Slack:

- `Docs: branch update failed. Manual intervention required`
- `GitHub: merge to develop failed. Check the Actions run and merge manually.`

Every docs step in `stable-publish` carries `continue-on-error: true`. That sets each step's `conclusion` to `success` and leaves the real result only in `outcome`, so the Actions UI showed a fully green step list while three steps had failed. The same breakage shipped on 17.1.0 and 18.0.0 — 18.0.0 was repaired by hand in `d969d9e051` and `82b0f96284`, and 17.1.0 was never repaired at all, so `prod-docs/17.1` still carries the stray tarballs and has only the three whitelisted API files.

#### What was broken

1. **The `.gitignore` surgery never matched.** The step grepped `^/content/api/$`, but `docs/.gitignore` holds `/content/api/*` plus three `!` negations. Nothing was removed (39 → 39 lines), the line-count guard exited 1, and `docs/content/api/` stayed ignored — so the generated API reference could never be committed. The `.gitignore` edit is gone entirely; the commit step force-adds the directory instead, which needs no edit and cannot drift.
2. **`cd docs && npm ci` could never work.** `docs/` is a pnpm workspace package and ships no `package-lock.json`, so this always died with `EUSAGE`. The `Install dependencies` step above it already populates `docs/node_modules` from the root, so the step now runs `npm --prefix docs run docs:api`.
3. **A dead script name.** `docs:scripts:generate-legacy-docs-versions` no longer exists in `docs/package.json` — this workflow was its last reference in the repo. Removed.
4. **The `docs/package.json` guard compared line counts.** `@handsontable/angular-wrapper` is absent on the release tag, so adding it always grows the file by one line. The guard failed *after* the write had already succeeded. It now asserts the value that was written.
5. **A bare `git add .`** in the workspace that also holds the downloaded release artifacts committed four `_tmp_packs/*.tgz` (5.5 MB+) onto the docs branch every release. Replaced with explicit paths.
6. **The docs branch never deployed.** It was pushed with `secrets.GITHUB_TOKEN`, and GitHub does not trigger workflows from such pushes, so `docs-production.yml` never ran for `prod-docs/18.1` at all. The docs steps now use the release App token. Confirmed against 18.0.0: the bot's push produced no docs-production run, and the first run matches the human repair commit.
7. **A stale lockfile broke the deploy.** The bot adds `@handsontable/angular-wrapper` to `docs/package.json` without touching `pnpm-lock.yaml`, but `docs-production.yml` installs with `pnpm install --frozen-lockfile`, so the deploy dies on `ERR_PNPM_OUTDATED_LOCKFILE` before it builds anything. A new `Update lockfile for the docs devDependency` step closes the gap.

Plus two supporting changes:

- **`gh pr create` failures no longer erase the merge PR.** On 18.1.0 the fallback pushed `merge/18.1.0-to-develop` correctly, then `gh pr create` failed with `Resource not accessible by integration (createPullRequest)`. `pr-url` stayed empty, so Slack printed the generic "merge failed" instead of naming the branch a human needed to open a PR from. The step now falls back to a compare URL, writes it to the job summary, and Slack's wording is accurate for both cases. The `gh pr list` lookup above it also got `|| true`, so a token with no PR access at all cannot kill the step before the branch is even pushed.
- **Failures are named in the job summary.** `Set docs result output` lists exactly which docs steps failed, so a broken docs branch stops looking green.

#### Turning the token on made the safety gate mandatory

Fixing 6 has a consequence worth calling out for review: once the push uses the App token, it **triggers the deploy**. Combined with `continue-on-error` on every docs step, the old unconditional commit would have pushed whatever half-assembled state the failures left, and the deploy would have run on it.

So `Commit and push docs branch` is now gated on all eight preceding steps:

```yaml
if: >-
  steps.release-token.outcome == 'success'
  && steps.docs-changelog-content.outcome == 'success'
  && steps.docs-deps.outcome == 'success'
  && steps.docs-branch.outcome == 'success'
  && steps.docs-api.outcome == 'success'
  && steps.docs-angular.outcome == 'success'
  && steps.docs-lockfile.outcome == 'success'
  && steps.docs-changelog.outcome == 'success'
```

A skipped push leaves the branch untouched and reviewable, which is strictly better than pushing a branch that deploys and fails.

For the same reason, the two guards that `exit 1` now revert the file they guard first (`git checkout -- pnpm-lock.yaml` / `docs/package.json`). Without that, the commit step stages the bad file anyway and the error message telling the operator to "land the change on develop, then re-run" would be false — it would already have shipped.

The App token is also minted **late**, right before the docs steps, rather than at the top of the job. At the top it puts the npm publish behind a docs-only dependency — a rotated key or a 5xx from the token endpoint would fail the job before `Publish` and cost the release its npm upload — and it burns the 60-minute installation-token lifetime before `docs-commit`, the last step that uses it. `Publish` is now step 5, the token step 10.

#### About the new `--lockfile-only` call

`release-lockfile.test.mjs` deliberately pins the number of `pnpm install --lockfile-only` calls in `publish.yml` so a new one cannot appear unnoticed (DEV-2667). This adds a third kind, and the constant moves from 5 to 6 with the reasoning recorded in the test:

- it is not a version bump, so `lockfile-float-gate.mjs` does not apply — that gate demands *zero* change, and this step must change the file;
- it is not a merge-conflict resolution;
- it is a real dependency addition, confined to a `prod-docs/*` branch that never reaches a registry.

The step asserts two things instead:

1. **The intended change landed.** `pnpm-workspace.yaml` sets `linkWorkspacePackages: true`, so `~X.Y.0` can resolve to the in-repo wrapper as a `link:`, which adds nothing under `packages:`. The float check alone would therefore pass identically whether pnpm wrote the importer entry or not — and this step is the only thing between the release and `ERR_PNPM_OUTDATED_LOCKFILE`. So it greps the `docs:` importer for `specifier: ~${VERSION}`.
2. **Nothing else moved.** It compares the set of resolved `name@version` identities under `packages:` and blocks on any **removal** other than the wrapper's. A DEV-2667 float re-resolves existing packages, so it always removes their old identity (core-js 3.37 → 3.50 removes `core-js@3.37.0`). Checking removals alone catches that while leaving room for the wrapper to pull in a genuinely new transitive dependency, which only ever adds.

Line counts cannot express this, which is worth spelling out because the first version of this PR got it wrong. The docs branch name drops the patch (`18.1.0` and `18.1.1` both map to `prod-docs/18.1`), so a patch release **rewrites** the devDependency rather than adding it — measured at 14 added / 2 deleted lines. Any "no deletions" rule would have failed every patch release's docs step.

Repairing the existing `prod-docs/18.1` branch is separate manual work and is not part of this PR.

### How has this been tested?

This is a CI-only change, so the checks are the tooling suite plus direct rehearsal of the shell logic against real repository state.

```bash
# The pinned lockfile-call test and the rest of the tooling suite
npm run test:tooling

# Every workflow file still parses
node -e "const y=require('js-yaml'),f=require('fs');
  f.readdirSync('.github/workflows').filter(n=>/\.ya?ml$/.test(n))
   .forEach(n=>y.load(f.readFileSync('.github/workflows/'+n,'utf8')));"
```

Result: **200/200 tooling tests pass, all 30 workflow files parse, all 19 `run` blocks in `stable-publish`, `stable-merge` and `stable-notify` pass `bash -n`.**

The float guard was exercised over six scenarios, built from real lockfiles generated by `pnpm install --lockfile-only`:

| scenario | expected | result |
|---|---|---|
| first add (develop → +dep) | passes | passes |
| patch bump 18.1.0 → 18.1.1 | passes | passes |
| DEV-2667 float (300 packages moved) | blocks | blocks (297 identities removed) |
| no change | passes | passes |
| float **and** the legitimate dep add together | blocks | blocks |
| wrapper gains a new transitive dependency | passes | passes |

The last two matter most: a float is still caught when the legitimate change is present too, and a future transitive dependency will not trip the guard.

The positive importer assertion was checked against three real lockfiles: found on a registry resolve, found on a patch bump, not found when the dependency is absent.

Also rehearsed against the real repository:

- **The tarball leak is reproduced and fixed.** With four fake `_tmp_packs/*.tgz` present, the old `git add .` stages all four plus any other untracked root file. The new explicit paths plus `git add -f docs/content/api` stage 104 generated API files and zero tarballs.
- **`npm --prefix docs run docs:api` works with no `npm ci`**, generating 106 files from `handsontable/tmp` — confirming `docs/scripts/jsdoc-convert/configuration.mjs` reads `pathToSource: '../../../handsontable/tmp'`, which the `Unpack handsontable tarball` step populates.
- **The `docs-result` loop** reproduces 18.1.0 exactly: `failed: docs-branch docs-api docs-angular`. Worth noting for anyone rehearsing shell locally — the first pass was validated under zsh, which does not word-split unquoted expansions, and silently reported no failures. Everything above was re-run under `bash`.

#### Known limitation

The docs pipeline is proven only up to the next stable release run. `docs:api` was tested against a locally built `handsontable/tmp`; in CI it reads the **published tarball's** contents, a path no release has ever reached because `npm ci` always died first. It should be equivalent, and the new `handsontable/tmp/base.js` sentinel plus the failure summary make any gap loud rather than silent, but it is not verified end-to-end yet.

#### Deliberately not changed

The deploy regenerates the API reference itself (`docs:docker:build:production` → `npm run build` → `docs:api`), so the committed `docs/content/api` copies are a snapshot the deploy never reads. Dropping them would be reasonable, but `prod-docs/18.0` carries 102 such files, so changing the branch shape is a team decision rather than something to fold into a bug fix.

### Types of changes

### Related issue(s):

1. DEV-2725

**Needs a settings change outside this repo:** the release GitHub App still has no `pull_requests: write`, so the merge PR will keep failing to open until an org admin grants it. This PR makes that failure legible and recoverable rather than silent.

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

CI/tooling only — `.github/workflows/publish.yml` and `.github/scripts/__tests__/release-lockfile.test.mjs`. No package source is touched, so the changelog gate passes on path. [skip changelog]

### Checklist:

ClickUp task: https://app.clickup.com/t/123kvxe920d

---

> [!NOTE]
> **Medium Risk**
> Changes only CI, but alters the stable npm publish job’s docs deploy path, GitHub App token usage, and lockfile rules—mistakes could block releases or push broken prod-docs branches that still trigger deploys.
> 
> **Overview**
> **Fixes the stable release `prod-docs/*` pipeline and hardens merge-to-develop when automatic PR creation fails** (DEV-2725).
> 
> In **`stable-publish`**, the docs branch update no longer relies on broken steps: it drops `.gitignore` surgery and `npm ci` in `docs/`, runs **`docs:api`** from an unpacked published tarball, pushes with a **late-minted release App token** (so `docs-production.yml` can run), adds **`docs-lockfile`** to sync `@handsontable/angular-wrapper` into `pnpm-lock.yaml` with targeted checks (not the float gate), stages only intended paths (no `_tmp_packs` tarballs), and **gates** `Commit and push docs branch` on all prior `continue-on-error` step **`outcome`s**. **`Set docs result output`** now lists failed step ids in the job summary.
> 
> In **`stable-merge`**, `gh pr list` / `gh pr create` failures no longer hide a pushed merge branch: fallback **compare URL**, step summary, and clearer Slack text when the App lacks `pull_requests: write`.
> 
> **`release-lockfile.test.mjs`** pins the sixth `--lockfile-only` call and adds tests for docs outcome wiring and `$VERSION` assignment per shell. **`AGENTS.md`** documents the third ungated lockfile-only case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f22edf2ef8a21b07a66ca9141402cf8c5f364529. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2725 Fix the release workflow's prod-docs branch and merge-PR fallback](https://app.clickup.com/t/9015210959/DEV-2725)
Task linked: [DEV-2667 [18.1.0-rc1+] Release-cut regenerated pnpm-lock.yaml, floating deps — production-bundle E2E legs deterministically red on release/18.1.0](https://app.clickup.com/t/9015210959/DEV-2667)

.github/workflows/publish.yml
```diff
@@ -1771,55 +1820,180 @@ jobs:
...
+          git diff --stat -- pnpm-lock.yaml
+
+          # First, the change must actually have landed. `linkWorkspacePackages: true`
+          # means `~X.Y.0` can resolve to the in-repo wrapper as a `link:`, which adds
+          # nothing under `packages:` -- so the float check below would pass identically
+          # whether pnpm wrote the importer entry or not. This step is the only thing
+          # between the release and ERR_PNPM_OUTDATED_LOCKFILE, so assert the positive.
+          if ! sed -n '/^  docs:$/,/^  [a-zA-Z]/p' pnpm-lock.yaml \
+               | grep -A 2 "'@handsontable/angular-wrapper':" \
+               | grep -q "specifier: ~${VERSION}"; then
```
demtario, line 1865: **High.** This step references `${VERSION}` but never assigns it — each `run:` block is a separate shell process, and every sibling step in this job (`docs-branch`, `docs-angular`, `docs-commit`) re-declares `VERSION` locally; this one doesn't. So `$VERSION` expands empty here, and the check silently degrades from "does the docs importer record `@handsontable/angular-wrapper` at exactly the intended version" to "does it record *any* `~`-prefixed specifier at all."

On a patch release (e.g. 18.1.0 → 18.1.1) a stale `~18.1.0` specifier left in place by a prior partial failure would pass this guard, get committed by `docs-commit`, and reproduce the exact `ERR_PNPM_OUTDATED_LOCKFILE` failure this step's own comment says it exists to prevent ("the only thing between the release and ERR_PNPM_OUTDATED_LOCKFILE"). The added test (`release-lockfile.test.mjs`) only counts occurrences of the `--lockfile-only` string, so it can't catch this.

Fix: add `VERSION='${{ needs.stable-build.outputs.version }}'` as the first line of this step's `run:` block, matching the sibling steps.

.github/workflows/publish.yml
```diff
@@ -1689,7 +1710,12 @@ jobs:
           done
           echo "All packages published successfully."
 
-      - name: Unpack handsontable tarball for CDN purge
+      # `handsontable/tmp` feeds two later steps: the CDN purge, and `docs:api` --
+      # docs/scripts/jsdoc-convert reads `pathToSource: '../../../handsontable/tmp'`.
+      # Unpacking the published tarball documents exactly what shipped, and is far
+      # cheaper than rebuilding from source.
+      - name: Unpack handsontable tarball
+        id: unpack-tarball
```
demtario, line 1722: **Medium.** `id: unpack-tarball` is added here but never consulted — it's absent from `docs-commit`'s `if:` step-outcome gate and from `docs-result`'s `OUTCOMES` list. `docs-api`'s own guard (`if [ ! -f handsontable/tmp/base.js ]`) only catches total extraction failure; if `tar -xzf` fails partway (e.g. a truncated download) after `base.js` already landed, this step correctly fails (continue-on-error catches it) but nothing downstream checks its outcome, so API docs can still be generated from a partial `handsontable/tmp` tree and pushed. Undercuts this PR's own goal ("never push a docs branch assembled from failed steps").

Fix: add `steps.unpack-tarball.outcome == 'success'` to both `docs-commit`'s `if:` condition and `docs-result`'s `OUTCOMES` env block.

.github/scripts/__tests__/release-lockfile.test.mjs
```diff
@@ -199,8 +206,11 @@ test('no new --lockfile-only site appears ungated', () => {
     EXPECTED_LOCKFILE_ONLY_CALLS,
     `publish.yml: expected ${EXPECTED_LOCKFILE_ONLY_CALLS} \`pnpm install --lockfile-only\` `
     + `calls, found ${found}. Three follow a version bump and are gated; two resolve a `
-    + 'pnpm-lock.yaml merge conflict in `stable-merge` and legitimately rewrite the file. '
-    + 'A new one needs a deliberate decision about which kind it is (DEV-2667).'
+    + 'pnpm-lock.yaml merge conflict in `stable-merge` and legitimately rewrite the file; '
+    + 'one adds the docs devDependency on the `prod-docs/*` branch and is checked for an '
+    + 'no other package losing its resolved identity. A new one needs a deliberate '
```
demtario: **Low.** "...is checked for an no other package losing its resolved identity..." — grammar error in this test-failure message ("checked for an no other" reads broken; e.g. "checked for it, and no other"). Cosmetic, not logic.

demtario (commented):
3 inline notes below.

**Off-diff:** the monorepo gotchas paragraph in root `CLAUDE.md`/`AGENTS.md` (around the `pnpm install --lockfile-only` / lockfile-float-gate discussion) still says there are two ungated `--lockfile-only` call sites (the two in `stable-merge`). This PR adds a third ungated category — the new `docs-lockfile` step's real-dependency-addition check — but the doc wasn't updated. Per the repo's own mandatory checklist ("Update AGENTS.md" when a change introduces a new convention/constraint), this paragraph should be revised to describe the third category and the updated ungated-call count.

.github/workflows/publish.yml
budnix: Nice catch — fixed in f22edf2ef8: the step sets `VERSION` first now, and a new test fails any step that reads `$VERSION` without assigning it.

.github/workflows/publish.yml
budnix: Good point — `unpack-tarball` is now in both the `docs-commit` gate and `docs-result`'s OUTCOMES, in f22edf2ef8.

.github/scripts/__tests__/release-lockfile.test.mjs
budnix: Good spot — reworded in f22edf2ef8, and the root AGENTS.md lockfile paragraph now lists the docs call as the third ungated kind.

demtario (approved):

PR #13330, opened 2026-09-02
DEV-2723: Return null from colToProp/propToCol for an unknown column (#7031)

budnix:
### Context

The PR makes [`colToProp()`](https://handsontable.com/docs/javascript-data-grid/api/core/#coltoprop) and [`propToCol()`](https://handsontable.com/docs/javascript-data-grid/api/core/#proptocol) return `null` when their argument names no column, closing [#7031](https://github.com/handsontable/handsontable/issues/7031). **This is a breaking change and is not releasable in a minor** — see "Release constraint" below.

Both methods used to hand the argument straight back. That answer could not be told apart from a real one, and where the visual and physical column order had diverged it named a *different* column than the caller asked about:

```
// 4 columns, physical column 1 trimmed
toVisualColumn(1)          = null        <- no visual index
propToCol(1)               = 1           <- claims it is at visual 1
...but visual column 1 is  = physical 2
getDataAtCol(propToCol(1)) = ["a2","b2"] <- physical column 2's data
```

After this change both answer `null`, matching `toVisualColumn()` and the other index translators.

A property name the data set does not use is still handed back unchanged — only indexes resolve. `colToProp()` also still returns a non-integer argument unchanged, which `undoRedo/actions/removeRow.ts` relies on.

### The scope decision

The break is confined to what the **two methods return**. Every internal consumer keeps behaving exactly as before, because widening the blast radius past what #7031 asks for would break far more than it fixes. Concretely:

- **Auto column growth** no longer routes through `propToCol()`. `applyChanges()` reads the bound from the numeric prop directly — `Number(propToCol(prop))` would now be `Number(null)` = `0`, silently ending growth. The comparison is unchanged for every index that reaches it, because an in-range prop always resolves below `countCols()` and never entered the loop.
- **The `*ByProp` selection hooks** keep emitting the negative header sentinel. `selectRows()`, `selectColumns()` and `selectAll()` all fire with a negative column today, and that is a sentinel, not an out-of-range index. `columnToPropForSelection()` passes anything `< 0` through — deliberately `< 0` rather than `=== -1`, since nested headers produce further negative levels.
- **`cellMeta.prop`** keeps carrying the index for a column auto-growth has not created yet, so `beforeValidate` / `afterValidate` / `postAfterValidate` and every cell function still receive what they did.
- **`getDataAtCell()`**, the copyable getters, the `beforeChange` / `afterChange` payloads, `validateChanges`, `processChanges`, the editor's `prop`, the ghost-table renderer args and the undo/redo replay all fall back to the index they were given before.

The one hook whose payload does change is **`modifyData`**: a *numeric* prop naming no column now arrives as `null` rather than the index. Documented in the migration guide.

### Two latent bugs this exposed

Both were found by the test suite, not by reading:

1. `dataSource.getAtRow()` filtered columns with `!Number.isInteger(prop)`. A column excluded by the `columns` option used to arrive as its index (rejected); as `null` it started passing, so `getSourceData()` returned columns the `columns` option had filtered out. Now tested with `prop !== null && !Number.isInteger(prop)`.
2. `getDataAtProp()` built a range from `propToCol()`. With `null` at both ends the range collapses to column `0`, so an unknown property returned column 0's data. It now returns an empty array.

### The type change

```ts
// Before
propToCol(prop: string | number): number;
colToProp(column: number): string | number;

// After
propToCol(prop: string | number): number | null;
colToProp(column: number): string | number | null;
```

This is the more visible half of the break: under `strictNullChecks`, `const col: number = hot.propToCol(prop)` stops compiling. `undoRemoveRowAccessor.types.ts` existed specifically to pin the old shape — it now pins the narrowed shape a consumer has to write, which is the same migration the guide documents.

The internal audit was driven by the compiler: widening the declaration surfaced 25 call sites across 8 files (formulas, dataProvider, filters, mergeCells, editorManager, ghostTable, metaSchema, dataMap). Each was resolved by an explicit check or a documented fallback — no `as` casts were used to silence one.

### Release constraint

**This cannot ship in 18.2.** `develop` is on the 18.x line; the change needs a major. The migration guide is filed as `migrating-from-18.2-to-19.0` on that assumption — if the next major is numbered differently, the directory, the two `permalink`s, `sidebar.js` and `docs/tests/paths.js` all need renaming together. Flagging rather than guessing.

- Unit tests added/modified (`*.unit.js`): none — the behavior needs a real `columnIndexMapper` and a real grid, so it is covered by the E2E specs below.
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`): none — the two methods under test have existing Jasmine specs, edited in place as the frozen-suite rules allow.
- Modified in the frozen Jasmine suite: `handsontable/src/__tests__/core/colToProp.spec.js` and `propToCol.spec.js`. Four assertions that read `expect(propToCol(10)).toBe(10)` — each carrying a `// I'm not sure if this should return result like that by design.` comment — now assert `null`, and that comment is replaced with what the change settled. New cases pin: the unknown index, the trimmed column on both branches, the preserved non-integer pass-through, the preserved unknown-property pass-through, auto column growth surviving, and the `-1` header sentinel still reaching `afterSelectionByProp`.
- Type tests (`*.types.ts`) updated if public API changed: `handsontable/src/__tests__/core/undoRemoveRowAccessor.types.ts`.
- For a bug fix — the spec that fails without this fix: n/a; this is a deliberate contract change, and the four flipped assertions above are what would fail if it were reverted.

### Commands run

```
$ npm run test:e2e --prefix handsontable
9872 specs, 0 failures, 8 pending specs

$ npm run test:types --prefix handsontable
(exit 0)

$ npm run test:unit --prefix handsontable
Tests: 5 failed, 4194 passed, 4199 total
```

The 5 unit failures are pre-existing and unrelated — `Handsontable.version` reads `18.0.0` against a `hot.config.js` bumped to `18.1.0` by the `Merge release/18.1.0 to develop` (#13324) that landed mid-branch, plus the license-expiry tests that read the same build date. Verified by running the same spec on a clean `origin/develop` checkout, where it fails identically.

Three regressions were found and fixed during the run rather than shipped: the `getAtRow` filter above, `dropdownValidator` (`cellMeta.prop` reaching `afterValidate` as `null`), and an undo replay that called `setDataAtCell` with a `null` column.

### Types of changes

### Related issue(s):

1. DEV-2723
2. #7031

Related: DEV-2726 covers the trimmed-column defect on `propToCol()`'s uncached branch as a **patch**, so the 18.x line can be fixed without waiting for this. That fix is subsumed here — if this lands first, DEV-2726 is already resolved; if DEV-2726 lands first, this PR needs a rebase but no rework. PR #13325 (DEV-2723 docs) touches the same two JSDoc blocks and the same two spec files, so whichever merges second needs a conflict resolution.

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

- [x] My change requires a change to the documentation. The JSDoc for both methods is updated and a migration guide is added.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2723

---

> [!NOTE]
> **High Risk**
> This is a documented breaking API change on core index translation used across data access, hooks, plugins, and TypeScript consumers; incorrect null handling could mis-address cells or break upgrades under strictNullChecks.
> 
> **Overview**
> **Breaking change (19.0):** `colToProp()` and `propToCol()` now return **`null`** when the argument names no existing, visible column (out-of-range index, trimmed column), instead of echoing the input. TypeScript signatures add `| null`. Unknown **property names** are still returned unchanged.
> 
> Internal paths that must keep prior behavior use **`colToPropOrIndex`** (index fallback only when there is no physical column) or **`propToCol(...) ?? prop`** so auto column growth, validation, undo/redo, selection, copy, and change hooks stay compatible. **`afterSelectionByProp` / `afterSelectionEndByProp`** bypass `colToProp` for negative header sentinels via `columnToPropForSelection`.
> 
> **Related behavior fixes:** `getDataAtProp()` returns **`[]`** for a numeric prop with no column; **`modifyData`**’s column argument is **`null`** in that case. Docs add a **18.2 → 19.0 migration guide**, changelog entry, and comment examples that guard on `propToCol === null`.
> 
> Specs and type tests are updated; plugins (formulas, filters, dataProvider, etc.) explicitly treat `null` from `propToCol`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 12abe185839ca2343debe26f848ba3aa26adb982. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2726 propToCol returns a wrong visual column for a trimmed column, and that also grows the grid spuriously](https://app.clickup.com/t/9015210959/DEV-2726)

PR #8727, opened 2021-09-17
Issue 7545: Vue3 Wrapper

Siemienik:
### Context
The PR adds support for Vue3 in the form of a new NPM package `@handsontable/vue3`.

Currently, the implementation supports the same functionality as the wrapper for Vue2 except for declarative renderers and editors.

Things to do:
- [x] ~Declarative renderers (`hot-renderer="true"`);~ *
- [x] ~Declarative editors (`hot-editors="true"`);~ *
- [x] Code and tests clean up;
- [x] Extensive testing all features and components with an external app (`@vue/cli`);
- [x] Documentation. Developed [here](https://github.com/handsontable/handsontable/pull/8991), it will be released/merged separately to this PR;

*&nbsp;the feature will not be available in the initial release

Additional minor things that I found during development:
- [x] One of the test show that changing the dataset by reference does not update the `HotTable` component;
- [x] On VSC IDE the definition files for Vue components cannot be found thus even though the package passes the tests (including TS), and published files are correct from the developer's perspective (its IDE) there is something wrong.

_**For reviewer**: There is one job that will fail. It is a quality code checker that fails because it detects code deduplication from other wrappers. I think that the current measure can be ignored as after merging to the `develop` branch the rate will drop to ~1.4% (acceptance level is 3%). See https://sonarcloud.io/summary/new_code?id=handsontable_handsontable&pullRequest=8727_

#### `hot-table` component example
```vue
<template>
  <div>
    <hot-table
      license-key="non-commercial-and-evaluation"
      :id="id"
      :class="className"
      :style="style"
      :rowHeaders="rowHeaders"
      :colHeaders="colHeaders"
      :columnSorting="columnSorting"
      :contextMenu="true"
      :readOnly="readOnly"
      :data="data"
      >
    </hot-table>
  </div>
</template>

<script lang="ts">
import { defineComponent } from 'vue';
import { registerAllModules } from 'handsontable/registry';
import { HotTable } from '@handsontable/vue3';

registerAllModules();

export default defineComponent({
  name: 'MyComponentName',
  data() {
    return {
      data: [[0, 1, 2, 3, 4], [5, 6, 7, 8, 9]],
      readOnly: false,
      rowHeaders: true,
      colHeaders: true,
      columnSorting: true,
      id: 'my-custom-id',
      className: 'my-custom-classname',
      style: 'height: 142px; overflow: hidden; border: 1px solid red;'
    }
  },
  components: {
    HotTable,
  }
});
</script>
[...]

PR #13331, opened 2026-09-02
fix(vue3): emit and publish the TypeScript declarations

demtario:
### Context

`@handsontable/vue3` declares `"types": "./index.d.ts"` (and `exports["."].types`) but has published **no `.d.ts` at all** since `17.1.0-rc1`. Consumers get `TS7016`, which under `noImplicitAny` is a hard error rather than merely untyped, and `HotTable` is implicitly `any` so nothing a Vue or Nuxt template binds is checked. It is also why the Nuxt starter had to annotate from `handsontable/settings` instead of from the wrapper, while the Angular starter could use `@handsontable/angular-wrapper`'s own `GridSettings`.

Published `.d.ts` count per version (`npm pack` + `tar tzf`):

| 14.6.1 | 15.3.0 | 16.0.1 | 17.0.0 | 17.0.1 | 17.1.0-rc1 | 17.1.0 | 18.0.0 | 18.1.0 |
|---|---|---|---|---|---|---|---|---|
| 15 | 15 | 15 | 15 | 15 | **0** | **0** | **0** | **0** |

The published `package.json` of 17.0.1 and 18.1.0 differ only in `version` and the `handsontable` peer range. `files`, `.npmignore` and `scripts` are byte-identical. Packaging was never the problem; **emission** is.

**Root cause.** The declarations came from `rollup-plugin-typescript2` inside the ES bundle build. #12237 ("chore: patch critical/high dependency vulnerabilities") added `.ts`/`.vue` to `nodeResolve`'s `extensions` and `@babel/preset-typescript` to the babel plugin in `wrappers/vue3/.config/base.js`, which silently stopped that emission. Rollup still exits 0. **The same commit made the same `base.js` change to `wrappers/react-wrapper` and covered it there**, adding `scripts/prepare-types.mjs` with a standalone `tsc --emitDeclarationOnly` pass wired into `build`. vue3 was left behind.

**Why `vue-tsc` and not plain `tsc`.** `src/vue.d.ts` shims `*.vue` as `DefineComponent<{}, {}, any>`. Plain `tsc` therefore succeeds and emits an `index.d.ts` that is **textually identical** to the good one, while typing `HotTable` as `any`: declarations that silence `TS7016` and check nothing, and no `HotTable.vue.d.ts` at all. That is one of the two failure modes the new type test is designed to catch, and it is measured below, not assumed.

**Why `strictNullChecks`.** The declaration build ran without it, so the checker folded `Handsontable | null` down to `Handsontable` and the emit claimed `hotInstance` is never `null`. It is `null` before `hotInit()` and after the grid is destroyed, so `hotTableRef.value.hotInstance.getData()` would have compiled and then thrown. Since the package has never published declarations there is no type contract to keep, and adding the `| null` after a release would be breaking for every TypeScript consumer, so the flag is on and the nullable members are annotated truthfully (`null as Handsontable | null`, `null as HotTableProps[] | null`, and the matching setter widening that TS2380 requires). Runtime behavior is unchanged: `columnSettings` was already assigned `newColumnSettings || null` and read behind a truthiness guard, and `for (const key in null)` is a no-op.

**Also fixed:** the entry point now exports its types. Even the releases that *did* ship declarations re-exported values only, and `exports` has no `./types` subpath, so `HotTableProps` was never reachable by a consumer.

**Prevention.** No wrapper package was subject to *any* declaration validation: `emitted-types.yml` (TS 5.1.6 typecheck + publint + attw) `cd`s into `handsontable/` at every step, and `scripts/verify-bundles.mjs` only compared embedded version strings. That is why this shipped in four consecutive releases unnoticed. Two guards added, plus the contract test, all three exercised in *Commands run*.

Ticket also asked about `@handsontable/vue` (Vue 2): **nothing to do.** The package was removed from the repo in `ba5bea1cbb` (#12015), and its last publish, `16.2.0`, ships its 6 `.d.ts` correctly.

- Unit tests added/modified (`*.unit.js`): none. `scripts/__tests__/typePointers.test.mjs` (10 cases, `node --test`) covers the new pure guard helper, including the exact `@handsontable/vue3@18.1.0` manifest shape
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`): none. This is a packaging and type-surface defect with no runtime behavior to drive in a browser
- Type tests (`*.types.ts`) updated if public API changed: `wrappers/vue3/test/types/vue3.types.ts` (new), run by the new `test:types` script
- For a bug fix, the spec that fails without this fix: `wrappers/vue3/test/types/vue3.types.ts`, red against both a plain-`tsc` emit and a non-null-checked emit, and `scripts/verify-bundles.mjs` exits 1 on the pre-fix state. All three A/Bs are reproduced below.

The type test does not assert "declarations exist". That assertion would have passed with the bug half-fixed. It asserts they mean something, on two axes:

- not `any`, via `type IsAny<T> = 0 extends (1 & T) ? true : false`, plus structural checks on the members `HotTable.vue` declares
- still nullable where the source is nullable, via `type IsNullable<T> = null extends T ? true : false`, asserted positively so a build that loses `strictNullChecks` goes red instead of shipping the unsound shape

### Commands run

**1. Reproduce (before the fix).** Build exits 0, emits `es/`, `commonjs/`, `dist/`, and zero declarations:

```
$ npm run build --prefix wrappers/vue3     # exit 0
$ find wrappers/vue3 -maxdepth 1 -name '*.d.ts'
(empty)
```

**2. After the fix.** 5 declarations, same filenames 17.0.1 shipped, and `HotTable.vue.d.ts` carries real members, not `DefineComponent<{}, {}, any>`:

```
$ npm run build --prefix wrappers/vue3
Running vue-tsc to generate declaration files...
TypeScript declarations prepared successfully.
$ tar tzf handsontable-vue3-18.1.0.tgz | grep -c '\.d\.ts$'
5
$ grep -E 'hotInstance|columnSettings' wrappers/vue3/HotTable.vue.d.ts
    __hotInstance: Handsontable | null;
    columnSettings: HotTableProps[] | null;
    hotInstance: Handsontable | null;
```

(17.0.1's count of 15 was rpt2 duplicating the same 5 files into `commonjs/` and `dist/`; nothing references those, and react-wrapper ships flat too.)

**3. The type test is not hollow, on either axis.**

Plain `tsc` emit, whose `index.d.ts` is byte-identical to the good one:

```
$ pnpm exec tsc -p tsconfig.json --emitDeclarationOnly --declarationDir ./types-tsc
$ npm run test:types --prefix wrappers/vue3
test/types/vue3.types.ts(25,7): error TS2322: Type 'false' is not assignable to type 'true'.
... 6 assertions red
```

Non-null-checked emit, which is what this branch shipped before the review round:

```
$ pnpm exec vue-tsc -p tsconfig.json --emitDeclarationOnly --strictNullChecks false --declarationDir /tmp/nonstrict
$ cp /tmp/nonstrict/*.d.ts wrappers/vue3/ && npm run test:types --prefix wrappers/vue3
test/types/vue3.types.ts(54,7): error TS2322: Type 'true' is not assignable to type 'false'.
test/types/vue3.types.ts(60,7): error TS2322: Type 'true' is not assignable to type 'false'.
```

Against the `vue-tsc` emit with the flag on, it is green.

**4. The declaration script fails closed.** A type error injected into `src/helpers.ts`, then the script run standalone after a full build:

```
$ node scripts/prepare-types.mjs; echo "EXIT=$?"
Running vue-tsc to generate declaration files...
src/helpers.ts(201,7): error TS2322: Type 'string' is not assignable to type 'number'.
Error: vue-tsc failed to emit the declaration files. See the diagnostics above.
EXIT=1
$ ls wrappers/vue3/*.d.ts
HotColumn.vue.d.ts  HotTable.vue.d.ts  helpers.d.ts  index.d.ts  types.d.ts
```

Non-zero exit, diagnostics printed raw rather than through `util.inspect`, and the previous build's five declarations left intact, so the tree is never handed back in the state this ticket describes.

**5. The publish guard fails on the pre-fix state and passes after** (`scripts/verify-bundles.mjs`):

```
$ mv wrappers/vue3/index.d.ts /tmp/ && node scripts/verify-bundles.mjs; echo $?
ERROR: Broken type declaration pointer: @handsontable/vue3 - pkg.types is ./index.d.ts but the file does not exist in .../wrappers/vue3.
ERROR: Broken type declaration pointer: @handsontable/vue3 - pkg.exports["."].types is ./index.d.ts but the file does not exist in .../wrappers/vue3.
1
$ mv /tmp/index.d.ts wrappers/vue3/ && node scripts/verify-bundles.mjs; echo $?
All packages have the expected version number and resolvable type declarations.
0
```

**6. End-to-end from a consumer's seat.** Same `tsconfig` (`strict`, `noImplicitAny`), same file, two installs:

```
# published @handsontable/vue3@18.1.0
consumer.ts(1,26): error TS7016: Could not find a declaration file for module '@handsontable/vue3'.
consumer.ts(2,36): error TS7016: Could not find a declaration file for module '@handsontable/vue3'.
TypeScript: 2 errors in 1 files

# the tarball built from this branch
TypeScript: No errors found
```

The second install also type-checks `import type { HotTableProps } from '@handsontable/vue3'`, which was unresolvable in every previous release.

**7. Suites and gates:**

```
$ npm run test --prefix wrappers/vue3        -> Test Suites: 5 passed, 5 total / Tests: 39 passed, 39 total
$ npm run test:types --prefix wrappers/vue3  -> clean
$ npm run lint --prefix wrappers/vue3        -> clean
$ npm run eslint                             -> clean
$ npm run test:tooling                       -> # tests 213 / # pass 213 / # fail 0
$ npm run build --prefix wrappers/react-wrapper   -> exit 0, 11 declarations flattened
$ npm run build --prefix wrappers/angular-wrapper -> exit 0
$ GATE_BASE=origin/develop node .github/scripts/test-presence-gate.mjs -> Pass.
$ node scripts/verify-bundles.mjs            -> All packages have the expected version number and resolvable type declarations.
$ pnpm install --frozen-lockfile             -> exit 0, lockfile unchanged
```

**8. publint / attw measured on all three wrappers before wiring CI.** publint exits 0 for all three from the package root (it honors `publishConfig.directory`, so one invocation also covers the Angular wrapper's `dist/hot-table` tree), and exits 1 with `pkg.types is ./index.d.ts but the file does not exist` on the pre-fix state. It is a precise detector for this defect.

**`attw` is deliberately not wired.** On a package with no declarations it short-circuits with `This package does not contain types` and **exits 0**, measured bare, with `--ignore-rules untyped-resolution`, and with `--entrypoints .`. It is blind to exactly this bug, so adding it here would have looked like coverage without being any. It stays a follow-up for interop shape, where it needs per-package ignore rules for pre-existing findings.

### Notes for the reviewer

- **`strictNullChecks`, not full `strict`, and that is measured.** `strictNullChecks` alone yields 4 errors, all fixed here. Full `strict` yields 19, and two groups are not mechanical: `HotTableProps[]` is not assignable to `ColumnSettings[]` (`HotTableProps extends GridSettings`, whose `data` is `unknown[][] | object[]`, against `ColumnSettings['data']` of `string | number | ColumnDataGetterSetterFunction`), and five `TS2571 Object is of type 'unknown'` from `HotColumn.vue`'s `inject()` calls, which need typed injection keys. The remainder are 4 implicit-any parameters in `helpers.ts` and `miscCache.currentSourceColumns` inferring `null`. Matching `wrappers/react-wrapper`'s `"strict": true` is worth doing, as its own task.
- `skipLibCheck: true` in `wrappers/vue3/tsconfig.json` is required, not cosmetic: vue 3.5.35's own `.d.ts` use `NoInfer` (TS 5.4) and `ToggleEvent` (TS 5.5 `lib.dom`), and this package pins `typescript@4.9.5`. Without it `vue-tsc` fails on six errors inside `node_modules` and emits nothing. `vue-tsc@^1.8.27` is the version that pairs with TS 4.9.5; the lockfile resolves `@vue/language-core@1.8.27(typescript@4.9.5)`. v2.x would force a TS major bump. The same flag in `test/types/tsconfig.json` also hides diagnostics inside the `.d.ts` under test, which is now stated in a comment there rather than left to be discovered.
- **The React wrapper got the half of this hardening that its TypeScript pin allows.** `clean` now removes the `types/` staging directory, `prepare-types.mjs` always regenerates rather than flattening whatever `types/` it finds, and a missing or empty emit throws where it used to `process.exit(0)`. Its `tsc` exit code must still be swallowed: that package pins `typescript@3.8.2`, which cannot parse `handsontable/tmp`'s inline type modifiers (`import { type EngineContext }`, TS 4.5+) or `@types/react`'s `ts5.0` tree, so it reports 32 `TS1005` and 5 `TS1110` from files outside `src/` on every run while still emitting all 11 `src/` declarations. Removing the swallow would red `npm run build` there permanently. The reason and the removal condition are recorded in `wrappers/react-wrapper/AGENTS.md`. Note this makes React's `prepare:types` fail-closed on an empty emit, a behavior change to a package this ticket does not otherwise touch.
- `wrappers/vue3/scripts/prepare-types.mjs` sweeps the flattened root `*.d.ts` only **after** a successful emit. Sweeping first meant a failed standalone run left `es/`, `commonjs/` and `dist/` in place with `types` advertised and no `index.d.ts`, which `npm pack` would have packed happily. `vue-tsc` does write into `types/` even when it reports errors, so the throw is what keeps that bad emit from reaching the package root; the staging directory itself never packs, since `.npmignore` is `*` with only root `*.d.ts` and the three bundle directories re-included.
- The dead `build:esDts` script is removed. `NODE_ENV=esDts` was never a key in `.config/factory.js`, so `rollup -c` fell through to a bare `baseConfig` with no output file, and its `rimraf ./buildTemp/` targeted a directory nothing creates. It has never been referenced by `build` or by CI since the wrapper landed in #8727.
- `/types` in `.gitignore` is anchored on purpose. Unanchored, `types` also matches `test/types/`, which silently excludes the new type test.
- The new `Type-surface contract test` CI step uses `--if-present`, so it is inert for the wrappers that have no such script yet. It cannot live in the (build-less) `Types` workflow because it reads the declarations the wrapper build emits.
- `scripts/verify-bundles.mjs` reads the *publish* manifest (honoring `publishConfig.directory`), not the source one. The Angular wrapper's source manifest declares no `types` at all, so reading it would have checked nothing.
- The `pnpm-lock.yaml` diff is additions-only. Six `semver` resolutions that an install had re-satisfied from their floating `^7` ranges were reverted, per the lockfile rule in the root `CLAUDE.md`.

Follow-ups deliberately out of scope: full `strict` in `wrappers/vue3/tsconfig.json` (19 errors, detailed above), the `false-cjs` masquerade publint warns about on **all three** wrappers (`types` names a `.d.ts` while `exports["."].import` names a `.mjs`; the fix is splitting into `.d.mts`/`.d.cts`, a shared change), removing the now-inert rpt2 `declaration: true` block from both wrappers' `.config/es.js`, upgrading react-wrapper off `typescript@3.8.2` so its script can throw like this one, and the `examples` repo switching its Vue and Nuxt starters to the wrapper's own type once this is released.

### Types of changes

### Related issue(s):
1. DEV-2732
2. #12237 (the commit that introduced the regression)

### Affected project(s):
- [ ] `handsontable`
- [x] `@handsontable/vue3`

### Checklist:

🤖 Generated with [Claude Code](https://claude.com/claude-code)


qunabu:
Task linked: [DEV-2727 Angular demo example not starting after release of 18.1.0](https://app.clickup.com/t/9015210959/DEV-2727)
Task linked: [DEV-2732 @handsontable/vue3 declares "types": "./index.d.ts" but ships no type declarations](https://app.clickup.com/t/9015210959/DEV-2732)

wrappers/vue3/tsconfig.json
```diff
@@ -8,8 +8,9 @@
     "sourceMap": true,
     "baseUrl": ".",
     "experimentalDecorators": true,
+    "skipLibCheck": true,
```
budnix, line 11: **The published types say `hotInstance` is never `null`, but it often is.**

This config sets no `strict` and no `strictNullChecks`. `vue-tsc` uses it for the emit, so the declarations are built with null checks off.

Two members lose their `| null`:

- `src/HotTable.vue:88` declares `get hotInstance(): Handsontable | null`, and it really does return `null` before `hotInit()` and after the grid is destroyed.
- `src/HotTable.vue:85` holds `columnSettings: null as HotTableProps[]`.

With null checks off, the checker folds `Handsontable | null` down to `Handsontable`, and that is what ships.

The green CI run is the proof. `test/types/vue3.types.ts:41` does `instance.hotInstance.getData` under `"strict": true`. If the emitted type still carried `| null`, that line would be TS18047. It passes, so the `| null` is gone.

For a user this means `hotTableRef.value.hotInstance.getData()` before mount compiles clean and then throws at runtime. `wrappers/react-wrapper/tsconfig.json` already sets `"strict": true`.

Worth fixing now rather than later: the package has never published declarations, so there is no type contract yet to keep. Ship `hotInstance: Handsontable` today and adding the `| null` back later becomes a breaking change for every TypeScript user - they would have to add `!` or null guards. This PR is the one free moment to get the shape right.

I know it is not a one-line change: turning `strict` on also makes the two `null as X` casts at lines 80 and 85 errors, so those need real types first. But the point of the PR is declarations that mean something, and right now the new contract test locks the unsound shape in rather than catching it.

wrappers/vue3/scripts/prepare-types.mjs
```diff
@@ -0,0 +1,89 @@
...
+
+console.log('Running vue-tsc to generate declaration files...');
+
+await execAsync(
+  'pnpm exec vue-tsc -p tsconfig.json --emitDeclarationOnly --declaration --declarationDir ./types',
+  // vue-tsc's diagnostics are the whole point of a failure here, and Node's default 1 MB
+  // pipe would abort the child with ENOBUFS before they arrive — the same trap the test
+  // hooks avoid with TEST_RUN_MAX_BUFFER. See .ai/LOCAL-ENFORCEMENT.md.
+  { cwd: root, maxBuffer: 64 * 1024 * 1024 }
+);
```
budnix: **`vue-tsc` errors get cut off when this step fails.**

`vue-tsc` writes its diagnostics to **stdout**, not stderr. This `await` has no `try`/`catch`, so a failure turns into an unhandled rejection. Node then prints the whole error object through `util.inspect`, and that helper cuts any string at 10,000 characters by default (checked on Node 24: `util.inspect.defaultOptions.maxStringLength` is `10000`). Past that you only get `... N more characters`.

So `maxBuffer: 64 * 1024 * 1024` keeps the output in memory, but the printing step throws most of it away. A build with many type errors shows only the first slice, wrapped in quotes and `+` joins.

Catching the error and writing the raw streams out fixes both the cut and the quoting:

```suggestion
try {
  await execAsync(
    'pnpm exec vue-tsc -p tsconfig.json --emitDeclarationOnly --declaration --declarationDir ./types',
    // vue-tsc's diagnostics are the whole point of a failure here, and Node's default 1 MB
    // pipe would abort the child with ENOBUFS before they arrive — the same trap the test
    // hooks avoid with TEST_RUN_MAX_BUFFER. See .ai/LOCAL-ENFORCEMENT.md.
    { cwd: root, maxBuffer: 64 * 1024 * 1024 }
  );

} catch (error) {
  // vue-tsc prints its diagnostics on stdout. An unhandled rejection would render them
  // through util.inspect, which cuts strings at 10000 characters, so write them out here.
  process.stdout.write(error.stdout ?? '');
  process.stderr.write(error.stderr ?? '');

  throw new Error('vue-tsc failed to emit the declaration files. See the diagnostics above.');
}
```

wrappers/vue3/scripts/prepare-types.mjs
```diff
@@ -0,0 +1,89 @@
+/**
+ * Cross-platform script to generate TypeScript declaration files for the Vue 3 wrapper.
```
budnix, line 2: **The React wrapper has the same script, and it never got this hardening.**

This file is a near copy of `wrappers/react-wrapper/scripts/prepare-types.mjs`. Only the Vue 3 copy was fixed. The React one still does the two things this PR names as the cause of DEV-2732:

1. It runs the compiler only when `types/` is missing (`if (!await exists(typesDir))`). And `wrappers/react-wrapper/scripts/clean.mjs` removes `es/`, `commonjs/`, `dist/` and the root `*.d.ts` - but not `types/`. So a leftover `types/` from an earlier build gets flattened into the package as it is.
2. It swallows a failed `tsc` in an empty `catch {}` and carries on with whatever landed on disk.

The React wrapper does ship declarations today, and the new publint step catches a total miss. But a stale or partial emit still passes both new gates, because the files are there.

Could the two wrappers share one script? The only real difference is the binary (`tsc` vs `vue-tsc`). If that is too much for this PR, adding the same two guards to the React copy would still close the gap.

wrappers/vue3/scripts/prepare-types.mjs
```diff
@@ -0,0 +1,89 @@
...
+// into the package, so there is deliberately no "skip when the directory already exists"
+// shortcut; and the flattened root `.d.ts` are cleared too, because the move below only
+// overwrites what vue-tsc re-emits. Without this, running `prepare:types` on its own after
+// renaming a source file leaves the old file's declaration at the package root. The `build`
+// script's `clean` step covers this case, but this script is also its own entry point.
+await rm(typesDir, { recursive: true, force: true });
+
+const staleDeclarations = (await readdir(root)).filter(entry => entry.endsWith('.d.ts'));
+
+await Promise.all(staleDeclarations.map(entry => rm(join(root, entry), { force: true })));
```
budnix: **A failed standalone run leaves a package you can still pack.**

These lines delete `types/` and every root `*.d.ts` *before* `vue-tsc` runs. The comment says this script is also its own entry point, so picture that case: someone runs `npm run prepare:types` after a full build, and `vue-tsc` then fails on a type error.

What is left behind: `es/`, `commonjs/` and `dist/` from the earlier build, `package.json` still saying `types: "./index.d.ts"`, and no `index.d.ts`. `npm run create-package` is a plain `npm pack` with no `prepack` hook, so it happily builds a tarball in exactly the DEV-2732 state.

A safer order: let `vue-tsc` emit into a clean `types/` first, and sweep the root `*.d.ts` only once it has succeeded. That keeps the "never flatten a stale `types/`" rule and never leaves the package worse than it was.

wrappers/vue3/test/types/tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+    "noEmit": true,
+    "baseUrl": "./",
+    "target": "ES2017",
+    "module": "ESNext",
+    "moduleResolution": "node",
+    "lib": ["ES2017", "dom"],
+    "esModuleInterop": true,
+    "allowSyntheticDefaultImports": true,
+    "types": [],
+    "skipLibCheck": true
```
budnix, line 20: **Worth a comment here: this hides errors in the files under test.**

Every file this test checks is a `.d.ts`, and `skipLibCheck` turns off all diagnostics inside `.d.ts`. The `IsAny` assertions still catch a collapse to `any`, but an emitted declaration that names a type `vue-tsc` could not express would ship broken and this test would stay green.

I can see it is forced by the same TypeScript 4.9.5 vs vue 3.5.x skew as the tsconfig one, so I am not asking you to flip it. Core's equivalent (`handsontable/test/types/tsconfig.json`) sets `"skipLibCheck": false`, so a short comment here pointing at the AGENTS.md note and the TypeScript 5 upgrade would stop someone later from either trusting this test too far or "fixing" the flag and going red.

wrappers/vue3/test/types/vue3.types.ts
```diff
@@ -0,0 +1,85 @@
...
+const getColumnSettingsIsNotAny: IsAny<HotTableInstance['getColumnSettings']> = false;
+const hotInit: () => void = instance.hotInit;
+const refreshColumns: () => void = instance.refreshColumns;
+
+// The public types must be importable from the entry point — before this fix
+// `index.d.ts` re-exported values only, so `HotTableProps` was unreachable even
+// in the releases that did ship declarations.
+const settings: HotTableProps = { data: [[1, 2]], rowHeaders: true, licenseKey: 'non-commercial-and-evaluation' };
+const nestedSettings: HotTableProps['settings'] = { colHeaders: true };
+const vueProps: VueProps<{ readOnly: boolean }> = { readOnly: 'anything' };
```
budnix, line 74: **`VueProps` can collapse to `any` and this file still passes.**

`HotTableProps` is protected: the `@ts-expect-error` further down turns into an unused-directive error if it becomes `any`. `VueProps` has no such guard. This line compiles both when `VueProps<T>` is the real mapped type and when `VueProps` is `any`, so it proves nothing.

One extra line closes the hole (and add `vuePropsIsNotAny` to the `export` list at the bottom):

```suggestion
const vuePropsIsNotAny: IsAny<VueProps<{ a: 1 }>> = false;
const vueProps: VueProps<{ readOnly: boolean }> = { readOnly: 'anything' };
```

scripts/verify-bundles.mjs
```diff
@@ -2,10 +2,14 @@ import '@angular/compiler';
 import globalJsdom from 'global-jsdom';
 import chalk from 'chalk';
 import glob from 'glob';
+import { existsSync, readFileSync } from 'fs';
+import { dirname, join, resolve } from 'path';
+import { fileURLToPath } from 'url';
```
budnix: **New `.mjs` imports are missing the `node:` prefix.**

`.claude/skills/node-scripts-dev/SKILL.md:70` says: "Always use the `node:` protocol prefix for built-in modules." These three added imports are bare.

```suggestion
import { existsSync, readFileSync } from 'node:fs';
import { dirname, join, resolve } from 'node:path';
import { fileURLToPath } from 'node:url';
```

The same applies to the five built-in imports at the top of `wrappers/vue3/scripts/prepare-types.mjs` (`child_process`, `fs/promises`, `path`, `url`, `util`).

pnpm-lock.yaml
```diff
@@ -21509,7 +21579,7 @@ snapshots:
       postcss-modules-scope: 3.2.1(postcss@8.5.12)
       postcss-modules-values: 4.0.0(postcss@8.5.12)
       postcss-value-parser: 4.2.0
-      semver: 7.7.1
+      semver: 7.8.4
```
budnix: **Six unrelated `semver` resolutions moved with this change.**

The only deletions in this lockfile diff are six `semver: 7.7.1` lines. They sit under `css-loader`, `istanbul-lib-instrument`, `npm-install-checks`, `npm-package-arg`, `npm-pick-manifest` and `postcss-loader`, and they all became `semver: 7.8.4`.

Adding `vue-tsc` to one workspace package cannot legitimately re-resolve those. It looks like a plain `pnpm add` / `pnpm install` re-satisfied the floating `^7` ranges.

Nothing new was fetched - 7.7.1, 7.7.2 and 7.8.4 were already all in the tree - so the risk here is low. But the repo's lockfile rule (root `CLAUDE.md`, DEV-2667) exists because this kind of drift is invisible downstream, and every `vue-tsc` entry the PR adds is pinned. Could you revert just those six lines?

wrappers/vue3/tsconfig.json
demtario: Fixed in 9a816ff. `strictNullChecks` is on, and the members behind it are annotated truthfully now:

- `__hotInstance: null as Handsontable | null`
- `columnSettings: null as HotTableProps[] | null`
- `set hotInstance(hotInstance: Handsontable | null)` (TS2380: the setter has to accept the getter's return type)
- `helpers.ts:101`, where `additionalHotSettingsInProps` is genuinely `null` whenever `props.settings` is unset

The emitted `HotTable.vue.d.ts` now reads:

```
    __hotInstance: Handsontable | null;
    columnSettings: HotTableProps[] | null;
    hotInstance: Handsontable | null;
```

No runtime change. `columnSettings` was already assigned `newColumnSettings || null` at `HotTable.vue:250` and read behind a truthiness guard at line 113, and `for (const key in null)` is a no-op, so the `| null` only writes down what the code already did. 39/39 Jest tests still pass.

I took `strictNullChecks` rather than the full `strict` that `wrappers/react-wrapper/tsconfig.json` sets, and that is a measurement rather than a preference. `strictNullChecks` alone: 4 errors, all of them the ones listed above. Full `strict`: 19, and two of those are not mechanical.

1. `HotTable.vue:113` and `:251`: `HotTableProps[]` is not assignable to `ColumnSettings[]`. `HotTableProps extends GridSettings`, whose `data` is `unknown[][] | object[]`, while `ColumnSettings['data']` is `string | number | ColumnDataGetterSetterFunction`. That is a genuine core-vs-wrapper model mismatch, and casting it away would be the exact kind of silencing this PR exists to stop.
2. `HotColumn.vue`: five `TS2571 Object is of type 'unknown'` from the `inject()` calls, which need typed injection keys.

The rest are 4 implicit-any parameters in `helpers.ts` and `miscCache.currentSourceColumns` inferring `null`. Happy to file that as its own task.

On the contract test locking the unsound shape in: it did, and it no longer does. Both members are asserted positively with `type IsNullable<T> = null extends T ? true : false`, kept next to the existing `IsAny` guard since `null extends any` also holds. The A/B against a deliberately non-strict emit:

```
$ pnpm exec vue-tsc -p tsconfig.json --emitDeclarationOnly --strictNullChecks false --declarationDir /tmp/nonstrict
$ cp /tmp/nonstrict/*.d.ts . && npm run test:types
test/types/vue3.types.ts(54,7): error TS2322: Type 'true' is not assignable to type 'false'.
test/types/vue3.types.ts(60,7): error TS2322: Type 'true' is not assignable to type 'false'.
```

`wrappers/vue3/AGENTS.md` now records why the flag has to stay, so it does not get read as cosmetic later.

wrappers/vue3/scripts/prepare-types.mjs
demtario: Applied as suggested in 9a816ff, and then triggered the path to check the claim holds. Injected a type error into `src/helpers.ts` and ran the script standalone:

```
$ node scripts/prepare-types.mjs; echo "EXIT=$?"
Running vue-tsc to generate declaration files...
src/helpers.ts(201,7): error TS2322: Type 'string' is not assignable to type 'number'.
Error: vue-tsc failed to emit the declaration files. See the diagnostics above.
EXIT=1
```

The diagnostic arrives raw and unquoted, so `error.stdout` does carry it, and the throw inside the top-level-await module does exit 1 rather than printing and continuing.

wrappers/vue3/scripts/prepare-types.mjs
demtario: Took the smaller option, and split it by what the measurement allows. In 9a816ff the React wrapper gets:

- `clean.mjs`: `types` added to the `dirs` array, so the staging directory can no longer survive a clean.
- `prepare-types.mjs`: the `if (!await exists(typesDir))` skip is gone. It always clears `types/` and recompiles, so a leftover directory is never flattened as it is.
- `prepare-types.mjs`: the silent `process.exit(0)` on "no declaration directory found" is now a throw, plus two more checks that fail on the outcome rather than the exit code: an empty emit, and no `index.d.ts` at the root after flattening.
- `/types` added to `.gitignore`.

Point 2 (the swallowed `tsc` failure) I did not remove, and I want to be explicit about why rather than just calling it out of scope. That swallow is load-bearing in this package. It pins `typescript@3.8.2`, which cannot parse the declarations it has to read:

```
$ pnpm exec tsc -p tsconfig.json --emitDeclarationOnly --declaration --declarationDir ./types-probe; echo $?
../../handsontable/tmp/3rdparty/walkontable/src/core/_base.d.ts(13,15): error TS1005: ',' expected.
...
../../node_modules/.pnpm/@types+react@18.3.31/node_modules/@types/react/ts5.0/index.d.ts(3367,50): error TS1110: Type expected.
1
```

32 x TS1005 (from `handsontable/tmp`'s inline type modifiers, `import { type EngineContext }`, TS 4.5+) and 5 x TS1110 (from `@types/react`'s `ts5.0` tree). Every one of them comes from a file outside `src/`, and `tsc` still emits all 11 `src/` declarations. Removing the swallow would turn `npm run build` red here permanently, for a package this PR does not otherwise touch. The three outcome checks above are what stands in for it: a missing or empty emit now fails, so the shape from this ticket cannot ship silently, while a mangled-but-present emit stays uncaught. That limitation is now written into `wrappers/react-wrapper/AGENTS.md`, next to the existing pitfall row about TS 3.8.2 mangling modern type helpers, with the swallow marked as removable once that TypeScript pin is lifted.

Flagging one consequence: React's `prepare:types` is now fail-closed on an empty emit where it previously exited 0. It is green today (verified by a full `npm run build --prefix wrappers/react-wrapper`, exit 0, all 11 declarations flattened), but that is a behavior change to a package outside this PR's stated scope, so it is your call whether it belongs here or in a follow-up.

On sharing one script: I would rather not in this PR. The binary is not the only difference any more, since the Vue 3 copy throws on a non-zero exit and the React copy cannot, and a single script would put both build pipelines behind one change.

wrappers/vue3/scripts/prepare-types.mjs
demtario: Reordered in 9a816ff, to the order you describe: clear `types/`, run `vue-tsc`, and only once that succeeded sweep the root `*.d.ts` and move the fresh files up. The sweep cannot move any earlier, because the move only overwrites what `vue-tsc` re-emitted, so a renamed source file would otherwise leave its old declaration at the package root.

Verified against your scenario. Full build first, then a type error injected into `src/helpers.ts`, then the script standalone:

```
$ node scripts/prepare-types.mjs; echo "EXIT=$?"
Running vue-tsc to generate declaration files...
src/helpers.ts(201,7): error TS2322: Type 'string' is not assignable to type 'number'.
Error: vue-tsc failed to emit the declaration files. See the diagnostics above.
EXIT=1

$ ls *.d.ts
HotColumn.vue.d.ts  HotTable.vue.d.ts  helpers.d.ts  index.d.ts  types.d.ts
```

All five declarations from the previous build are still there, so the tree is left no worse than it was rather than packable in the broken state.

One detail worth recording: `vue-tsc` does emit into `types/` even when it reports errors, so the failed run leaves five declarations staged there. They never reach a tarball, because `.npmignore` is `*` with only root `*.d.ts` and `commonjs/`, `dist/`, `es/` re-included, and step 1 of the next run clears the directory anyway. The throw is what keeps that bad emit from being flattened to the root.

wrappers/vue3/test/types/tsconfig.json
demtario: Comment added in 9a816ff, with the limitation stated rather than implied:

```jsonc
    "types": [],
    // `skipLibCheck` is a stand-in for the TypeScript 4.9.5 vs vue 3.5.x skew described in
    // `wrappers/vue3/AGENTS.md`, not a free choice: without it the checker fails inside vue's
    // own `.d.ts`. The cost is that it also silences diagnostics inside the files under test,
    // which are all `.d.ts` — so an emitted declaration naming a type `vue-tsc` could not
    // express would ship broken and this suite would stay green. The `IsAny` assertions below
    // still catch a collapse to `any`. Core's equivalent (`handsontable/test/types/tsconfig.json`)
    // sets `false`; set it here too when this package moves to TypeScript 5.
    "skipLibCheck": true,
```

Checked that the JSONC comment parses, since no other `tsconfig.json` in the repo carries one: `tsc --showConfig -p test/types/tsconfig.json` is clean and `npm run test:types` runs normally.

wrappers/vue3/test/types/vue3.types.ts
demtario: Right, that line held either way. Added in 9a816ff, with `vuePropsIsNotAny` in the `export` list:

```ts
const vuePropsIsNotAny: IsAny<VueProps<{ a: 1 }>> = false;
const vueProps: VueProps<{ readOnly: boolean }> = { readOnly: 'anything' };
```

`VueProps<T> = { [P in keyof T]: any }`, so the mapped type itself is not `any` and the guard is `false` on a real emit while flipping on a collapsed one.

scripts/verify-bundles.mjs
demtario: Done in 9a816ff. Both sets carry the prefix now: the 3 in `scripts/verify-bundles.mjs` and the 5 in `wrappers/vue3/scripts/prepare-types.mjs`.

Also applied to the two React wrapper scripts touched in the same commit (`prepare-types.mjs`, `clean.mjs`), since they were bare too. `scripts/utils/typePointers.mjs` and its test needed nothing: the helper module imports no built-ins, and the test already used `node:test` and `node:assert/strict`.

pnpm-lock.yaml
demtario: Reverted in 9a816ff. All six patched by line number rather than by pattern, so the legitimate `semver: 7.8.4` entries elsewhere in the tree are untouched:

```
$ git diff origin/develop -- pnpm-lock.yaml | grep '^-' | grep -v '^---'
(empty)
$ git diff origin/develop --stat -- pnpm-lock.yaml
 pnpm-lock.yaml | 88 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 1 file changed, 88 insertions(+)
```

The lockfile diff is additions-only now, which is the shape a pinned devDependency should produce. `pnpm install --frozen-lockfile` exits 0 against it and rewrites nothing.

budnix (approved):

PR #13332, opened 2026-09-02
DEV-2739: Close the editor strand window when alter() finishes its work

danielzytohoc:
### Context

The PR fixes a residual data-corruption path left open by DEV-2652 (#13272), and with it the dominant CI flake of the past week (`editor-trimmed-row.spec.ts` "a removal that strands the editor", 15 red legs across 9 branches since 2026-08-28 — investigated under DEV-2668).

DEV-2652's "strand window" protects an editor that a structural `alter()` stranded while that `alter()`'s own `selection.shiftRows()` re-prepare is still coming. The window was closed by `hot._registerTimeout(..., 0)` — a plain `setTimeout(0)`, i.e. the **next macrotask** — although its JSDoc claimed it closes "when the task that opened it ends". While the window leaks past `alter()`'s return:

- `#reconcileEditorWithIndexMaps()` skips the discard of a stranded editor, and
- a trimming change that lands before the timer fires (`Filters#filter()` in the same task — one line of application code away — or in the next CDP evaluate on a loaded CI runner) commits through the stale visual coordinates, and `applyChanges()` **appends records** to reach them.

`hot.alter('remove_row', 1); hot.getPlugin('filters').filter();` reproduced the corruption deterministically with #13272 merged.

The fix closes the window at the true boundary the JSDoc always described: the end of `alter()`'s synchronous work. `grid.alter()`'s common tail now calls the new `EditorManager#closeStrandWindow()` (after `adjustRowsAndCols()`, i.e. after `selection.shiftRows()` had its chance to re-prepare). The zero-delay timeout stays as a backstop for openers no `alter()` tail closes (a structural cache update fired outside `alter()`, or an `alter()` aborted by a throwing hook). The field is renamed `#strandedInCurrentTask` → `#strandedInCurrentAlter` to match the real contract, and its JSDoc now states the mechanism, the boundary, and why the boundary matters.

The protected case is unchanged: a plugin trimming from `afterRemoveRow` (an index-map change nested **inside** the running `alter()`) still sees the window open, and the existing spec pinning that behavior stays green.

Note for the release notes: this also turns the probabilistic CI failure of the existing "discards rather than appending when a later filter would commit through it" case into a deterministic pass — the spec was honest; the product raced.

### How has this been tested?

- New Playwright case `discards when the filter lands in the same task as the removal` (`tests/e2e/editor-trimmed-row.spec.ts`) with a new page-object driver `removeRowThenFilterSameTask()` that runs `alter()` and `filter()` in one synchronous `page.evaluate` block. Verified **red on unfixed develop for the right reason** (`committedChangeCount` 1 vs 0, source grown to 8 records), deterministically, before applying the fix.
- Green after the fix, including a 15× soak of the whole strand family:
  - `cd tests && npx playwright test --project=e2e-main e2e/editor-trimmed-row.spec.ts e2e/editor-hidden-cell.spec.ts e2e/selection-trimmed-row.spec.ts`
  - `cd tests && npx playwright test --project=e2e-main e2e/editor-trimmed-row.spec.ts -g "strands the editor|nested inside a removal" --repeat-each=15` → 45/45 passed.
- `npm run test:types --prefix handsontable`, `npx eslint src/editorManager.ts src/core.ts` (handsontable) and `npx eslint` on the two changed test files — clean.
- `npm run build --prefix handsontable` — clean.

### Types of changes

### Related issue(s):

1. DEV-2739
2. DEV-2668 (the CI-flakiness investigation that isolated this defect)
3. DEV-2652 / #13272 (the change that introduced the strand window)

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2739

---

> [!NOTE]
> **Medium Risk**
> Changes core `alter()`/`updateData()` and editor reconciliation timing around structural edits and open editors, which can affect data integrity and plugin hooks, though behavior is scoped and backed by targeted e2e tests.
> 
> **Overview**
> Fixes a case where **source rows were appended** if a filter ran right after row removal while a cell editor was still open (DEV-2739).
> 
> The editor manager no longer relies on a **zero-delay timeout** to decide when a “stranded” open editor should be discarded. It uses **depth-counted `suspendStrandDiscards()` / `resumeStrandDiscards()` scopes** around structural work in **`alter()`** and **`updateData()`** (after index-map updates, through selection repair). While a scope is open, reconcile tolerates a temporarily unresolvable editor; when the outermost scope closes synchronically, a follow-up trim/filter in the **same task** can discard the pending edit instead of committing through stale visual coordinates.
> 
> E2E coverage adds assertions for filter-after-remove and a deterministic **`removeRowThenFilterSameTask`** helper; changelog **13332** documents the user-facing fix.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7160d409ccb40f6365df2870c2a636a95418210c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2739 Editor strand window closes on setTimeout(0): a same-task trim/filter commits through a stranded editor and appends rows](https://app.clickup.com/t/9015210959/DEV-2739)

handsontable/src/editorManager.ts
```diff
@@ -208,6 +218,24 @@ class EditorManager {
...
+   * The core calls this at the end of `alter()`'s synchronous work, once `selection.shiftRows()`
+   * has had its chance to re-prepare the stranded editor. From this point on a trimming change
+   * that finds the editor unresolvable discards the pending edit instead of tolerating it – which
+   * is what keeps a `Filters#filter()` following the `alter()` in the SAME task from committing
+   * through the stranded coordinates and appending records (DEV-2739).
+   *
+   * An `alter()` nested inside another one (fired from a hook) closes the outer window on its own
+   * tail too. That early close narrows the outer call's protection, but only toward a discard of
+   * the pending edit – never toward the appending commit this window exists to prevent – and the
+   * outer re-prepare, when it comes, re-derives the editor's state independently of the window.
```
demtario: **medium**: `#strandedInCurrentAlter` is a single non-reentrant boolean, and this doc comment's claim that a nested `alter()` closing the outer window early narrows protection "only toward a discard of the pending edit – never toward the appending commit" doesn't hold in all cases.

`datamap.removeRow()` (core.ts:1225) fires `afterRemoveRow` synchronously, before the outer `alter()` reaches its own `selection.shiftRows()` (core.ts:1287) or its own `closeStrandWindow()` tail (core.ts:1398). If that hook synchronously calls `instance.alter(...)` again (a realistic pattern – `EditorTrimmedRowPage.removeRowTrimmingFrom` already shows hooks firing structural work from `afterRemoveRow`), the nested call's own tail clears the *shared* flag before the outer call's `shiftRows()`/`prepareEditor()` has had its chance to legitimately re-derive the stranded editor's coordinates. A same-task reconcile landing in that gap then discards an edit the outer `alter()` would have correctly recovered – which is exactly the regression `tests/e2e/editor-trimmed-row.spec.ts:871-878`'s existing invariant warns against ("doing anything else here is a regression rather than a gap").

No shipped core plugin triggers this today, so it's not blocking, but the doc comment overclaims safety. Consider a reentrancy-safe close (depth counter, or track which `alter()` call opened the window), or at minimum narrow the JSDoc claim and add a regression test nesting `alter()` inside `afterRemoveRow`.

handsontable/src/core.ts
```diff
@@ -1390,6 +1390,13 @@ export default function Core(
         grid.adjustRowsAndCols(); // makes sure that we did not add rows that will be removed in next refresh
       }
 
+      // The alter's synchronous work is done: `selection.shiftRows()` has had its chance to
+      // re-prepare an editor the removal stranded. Closing the strand window here rather than on
+      // the zero-delay backstop timeout is what lets a trimming change that follows in the SAME
```
demtario: **low**: this comment (and the matching JSDoc at `editorManager.ts:98-109,224-225`) attributes the re-prepare exclusively to `selection.shiftRows()`, but `closeStrandWindow()` sits after the whole `switch`, so it equally covers `remove_col`/`insert_col`, whose counterpart is `selection.shiftColumns()` (core.ts:1201,1354) – and `#recaptureEditedRecord()` opens the window on a null physical *column* too (editorManager.ts:522-523,531). Consider broadening the wording to name both axes so a future reader doesn't assume the protection is row-only.

demtario (commented):
2 inline notes below. One more finding has no changed line to anchor to:

- **medium**: `Core#updateData()` (`handsontable/src/core.ts:3357-3373`) hits the same structural-stranding mechanism as `alter()` (`fitToLength` → `removeIndexes` → `afterRowSequenceCacheUpdate` → `#recaptureEditedRecord()`), but its completion callback never calls `editorManager.closeStrandWindow()`. `hot.updateData(shorterData)` followed same-task by `filters.filter()` reproduces the same DEV-2739 append-corruption this PR fixes, just via a different entry point. Consider calling `closeStrandWindow()` from `updateData()`'s completion path too, or moving the close to a shared choke point so every structural-change caller gets it for free.

handsontable/src/editorManager.ts
danielzytohoc: Fixed properly rather than narrowing the claim: the boolean is gone, replaced with depth-counted scopes in the house `suspend/resume` shape (`EditorManager#suspendStrandDiscards()` / `resumeStrandDiscards()`, `#strandDiscardsSuspended` counter). `grid.alter()` opens a scope at entry and closes it at its tail, so a nested `alter()` can only end its own level — the outer call's protection survives by construction. A zero-delay timeout still heals a scope a thrown hook aborts, bounding a leaked suspension to the current task.

Two empirical notes from driving your scenario live (instrumented against the pre-fix bundle):

1. An `afterRemoveRow`-nested `alter()` cannot reach the gap at all — that hook dispatches before the mapper's cache updates flush, so the nested tail runs before the strand is even marked.
2. The deepest reachable point, a nested `alter()` fired from `afterRowSequenceCacheUpdate` itself (after `#recaptureEditedRecord()`), also produced no discard on the old boolean — probe log: the nested call's own `shiftRows()` re-derived the stranded editor's coordinates (row 4 → 3, resolvable) *before* its tail cleared the flag. The rescue and the flag-clear travel together inside any nested `alter()`, so the overclaim was emergently true — but only emergently.

The depth counter makes it structural, and the new spec case (`keeps the edit when a nested alter() ends inside the removal that stranded it`, driving exactly that cache-update-nested shape) pins the end-to-end invariant so a refactor that reorders the nested tail ahead of the selection step fails the suite instead of shipping the discard regression. Soaked 75/75 on the strand family, 393/393 across the matrix with the adjacent editor/selection specs.

handsontable/src/core.ts
danielzytohoc: Broadened as suggested: the tail comment now names `shiftColumns()` and notes `#recaptureEditedRecord()` opens the scope on either axis, and the field/method JSDoc in `editorManager.ts` describes both repairs (`shiftRows()`/`shiftColumns()` for `alter()`, `selection.refresh()` for `updateData()`).

danielzytohoc:
Re the review-body finding (updateData shares the stranding mechanism with no window close) — **confirmed and fixed**. `hot.updateData(shorterData); filters.filter();` in one task reproduced the append exactly as you predicted (new spec case went red on the previous HEAD with `committedChangeCount` 1 and the appended rows, before the fix). `updateData()`'s completion callback now wraps its structural span in the same scope pair — `suspendStrandDiscards()` before `fitToLength()`, `resumeStrandDiscards()` after `selection.refresh()` (its re-prepare chance) — and the case is green: the same-task filter discards instead of appending.

On the choke-point alternative: there is no shared point that knows when a structural operation *finished* — the opener is hook-driven (`#recaptureEditedRecord`), but only the operation knows where its selection repair ends, which is exactly why DEV-2652 fell back to a timer. The depth-counted scope is the generalized answer: operations opt in with a pair, and a structural cache update fired with **no** scope open now gets no suspension at all — the next reconcile discards the stranded editor immediately, which loses the pending edit but can never append (the safe failure direction for any exotic caller core does not wrap). `loadData()` was checked and left unscoped on purpose: it resets the index maps through the `init` source, which routes to reconciliation rather than recapture, so it never opens a strand.

All three findings addressed in c0f6eb9fe; full spec 28/28, strand family soaked 75/75, matrix 393/393.

handsontable/src/editorManager.ts
danielzytohoc: Follow-up on the regression-test half, because the picture changed with data: I implemented the nesting test in both shapes — `afterRemoveRow` (your suggested minimum) and directly on `afterRowSequenceCacheUpdate` (the deepest reachable point) — and hammered them focused across all six legs. Both shapes hit a **pre-existing develop defect**, not the scope mechanics: the nested call's selection machinery and the outer `shiftRows()` adjust the open editor twice, and the commit lands one record up (deterministically without a trim in the nested span; ~50% under load with one, because the trim's selection repair sometimes corrects the landing). Reproduced identically on a bundle built from unchanged `origin/develop`, so it is neither caused nor fixed here — filed as **DEV-2755** with the repro, and the nested-alter test moved there with it (a pin here would be permanently racy against someone else's defect, which `failOnFlakyTests` rightly forbids). The depth-counted scopes stand on their own: they make the JSDoc claim true by construction instead of emergent.

danielzytohoc:
**Correction to my updateData reply above — the "green" claim needs qualifying.** The scope pair in `updateData()`'s completion callback is in and does restore the calm-path behavior (same-task `updateData(shorter); filter()` discards instead of appending — this was deterministic-red before the pair). But a focused hammer (`--repeat-each=15`, all six legs) showed the end state is still load-dependent on unchanged develop: ~50% of loaded runs commit through and append anyway, which points at `replaceData`'s completion-callback timing or `selection.refresh()`'s render-state-dependent re-prepare racing the scope's resume — deeper than this PR's mechanism. Filed as **DEV-2756** with the repro and load data; the deterministic-assert test was descoped there (7160d409c) so this PR doesn't ship a coin-flip.

Two incidental data points from the same soaking, relevant to you as the DEV-2680 author: (1) DEV-2755 (also just filed) — a nested `alter()` during a removal's cache update lands the open editor's commit one record up, deterministic in the no-trim shape, on unchanged develop; (2) the `commit paths that read the selection after trimming` / edge-case families from #13304 fail ~135 of 738 executions under a `--repeat-each=3` full-matrix soak locally — and pristine develop scores an indistinguishable 146/738, so it's not this branch. Worth a look before those families meet a loaded CI runner; they'll surface in the DEV-2668 flake data otherwise.

Current PR state: depth-counted strand scopes (reentrancy-safe by construction), `updateData()` scope pair, both-axes wording, same-task regression test — 41/41 full spec, 360/360 focused hammer of the PR-owned cases across all legs.

demtario (approved):

PR #10086, opened 2022-11-03
Implement ResizeObserver to the table's root element

budnix:
### Context

The PR fixes the issue related to the UI glitches that happened while expanding Handsontable's root element from the outside. Or expanding the window size that is caused by the context menu or dropdown menu within the iframe.

_[skip changelog]_

### Types of changes

- [x] New feature or improvement (non-breaking change which adds functionality)

### Related issue(s):
1. fixes #9580

### Affected project(s):
- [x] `handsontable`

### Checklist:

- [x] I have reviewed the guidelines about [Contributing to Handsontable](https://github.com/handsontable/handsontable/blob/master/CONTRIBUTING.md) and I confirm that my code follows the code style of this project.
- [x] I have signed the [Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLScpMq4swMelvw3-onxC8Jl29m0fVp5hpf7d1yQVklqVjGjWGA/viewform?c=0&w=1)
- [x] My change requires a change to the documentation.

PR #13334, opened 2026-09-02
DEV-2744: Migrate the flaky RefreshDimensions specs to Playwright

danielzytohoc:
### Context

The PR migrates the flakiest legacy spec family on the CI-flakiness investigation's ledger — the `{before,after}RefreshDimensions` Jasmine specs — to Playwright (`tests/e2e/refresh-dimensions.spec.ts`), executing the frozen-suite attrition policy. This family caused 5 red legs in the investigation's window and 7 more in the 08-27→08-31 extension: the iframe group's fixed `await sleep(50)` races the resize-event→rAF→hook pipeline, and two tests asserted synchronously after their mutation, so the call they counted was the ResizeObserver's **initial** delivery, not the resize — load-sensitive since 2022 (PR #10086), made tighter by the 300→50 ms cut in PR #11463.

The migration replaces every fixed wait with polling on an interleaved hook log, and strengthens the coverage:

- `before`/`after` are asserted together through one log, which also pins their **order** — the per-hook legacy files never could.
- The initial-delivery behavior the old synchronous asserts accidentally depended on is now pinned **explicitly** (one delivery on observe, equal dimensions, `actionPossible: false`).
- The legacy rAF-sync contract ("a resize never fires the hooks synchronously") is kept in its observable form: the resize driver returns the hook-log length read in the same task as the mutation, asserted `0`, before the poll for the deferred delivery — no `requestAnimationFrame` stub needed.
- Blocking (`beforeRefreshDimensions → false`) is covered on **both** pipelines and asserts the **observable effect**, not just spy counts: on the main-document ResizeObserver path the scroll holder ignores the new root width; on the iframe window-resize path the grid carries 20 columns against a 500 px viewport so horizontal virtualization exposes the refresh — an unblocked shrink renders fewer columns (positive control), a blocked one keeps the count.
- Every trigger that can deliver more than once (unhiding a root or body, a pinned iframe grid's window resize) is asserted by **polling the latest entry of a given hook** (`lastEntry('after')`), never by reading the array's last element in a separate round trip — a further before/after pair landing between the two reads would otherwise fail a correct grid.
- The hidden-root/hidden-body suppression tests pair the bounded negative settle with a positive control (unhiding delivers the size the root grew to while hidden), so the negative half cannot pass vacuously.
- The iframe assertions compare the hook payload to what the iframe **measures at assertion time** (viewport `clientWidth`, root height) instead of the legacy's theme-dependent constants — which is why the spec runs identically on all six theme × bundle legs.

Legacy attrition: `beforeRefreshDimensions.spec.js` is deleted; `afterRefreshDimensions.spec.js` keeps **only** the dvh loop-guard spec, with a header comment explaining why — it cannot be made honestly green until the ResizeObserver loop-guard product ticket fixes the guard in `resizeMonitor.ts`, and it should migrate together with that fix. The file's `no-fixed-sleep-in-spec` debt drops from 23 warnings to 2 (both inside the kept dvh test).

### How has this been tested?

- `cd tests && npx playwright test --project=e2e-main e2e/refresh-dimensions.spec.ts` — 7/7.
- Full matrix soak after the review round: `npx playwright test e2e/refresh-dimensions.spec.ts --repeat-each=10` (all six `e2e-*` legs) — **420/420 passed**, including the `-min` legs local gates never cover.
- Trimmed legacy file still runs: `npm run test:e2e --testPathPattern=afterRefreshDimensions` — 0 failures (dvh passes on an idle machine; its CI red is the ResizeObserver loop-guard product defect).
- `npx eslint` on the changed test files — clean (2 pre-existing warn-level sleeps remain in the kept dvh test, tracked by the loop-guard and sleep-debt tickets).

### Types of changes

### Related issue(s):

1. DEV-2744
2. The CI-flakiness investigation task (linked in ClickUp) — the flake ledger this executes
3. The ResizeObserver loop-guard product ticket (linked in ClickUp) — blocks migrating the remaining dvh spec

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2744

---

> [!NOTE]
> **Low Risk**
> Test-only refactor with no production code changes; risk is limited to test coverage gaps if the migration missed an edge case.
> 
> **Overview**
> Moves flaky Jasmine coverage for **`beforeRefreshDimensions`** / **`afterRefreshDimensions`** into Playwright so CI stops racing fixed `sleep(50)` against the resize → rAF → hook pipelines.
> 
> **`beforeRefreshDimensions.spec.js`** is removed. **`afterRefreshDimensions.spec.js`** keeps only the **`dvh`** ResizeObserver loop-guard case (documented as blocked on DEV-2740). New **`tests/e2e/refresh-dimensions.spec.ts`**, **`refresh-dimensions.html`**, and **`RefreshDimensionsPage`** drive main-document ResizeObserver and iframe window-resize paths via a polled interleaved hook log (including **`before`/`after` order**), synchronous-then-deferred resize behavior, blocking with DOM effects (holder width / virtualized column count), hidden root/body suppression, and runtime-measured iframe dimensions instead of theme literals.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b6c17242d4acb781e4659da7123735fc74dd8a7b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2740 ResizeObserver loop guard never trips on a loaded machine, and a legitimate trip disconnects resize reactivity permanently](https://app.clickup.com/t/9015210959/DEV-2740)
Task linked: [DEV-2744 Migrate the flaky RefreshDimensions legacy specs (iframe + root resize) to Playwright](https://app.clickup.com/t/9015210959/DEV-2744)
Task linked: [DEV-2184 Escalation burn-down 2: fixed-sleep debt in specs (~562 occurrences), flip no-fixed-sleep-in-spec to error](https://app.clickup.com/t/9015210959/DEV-2184)

tests/e2e/refresh-dimensions.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+    await grid.afterAnimationFrames(3);
+
+    expect(await grid.hookLogLength()).toBe(0);
+
+    // Positive control: unhiding delivers the size the root grew to while hidden.
+    await grid.setRootDisplay('');
+    await expect.poll(() => grid.hookLogLength()).toBeGreaterThanOrEqual(2);
+
+    expect((await grid.hookLog()).at(-1)).toEqual(
+      entry('after', { width: 120, height: 100 }, { width: 200, height: 100 }, true),
```
demtario, line 114: **High.** The preceding poll (`toBeGreaterThanOrEqual(2)`, not `toBe(2)`) already admits more than one before/after pair can land here - the comment above even says so ("unhiding delivers the size the root grew to", not "exactly one delivery"). But this assertion takes `.at(-1)` from a *separate* `page.evaluate` round trip and pins it to one specific `after` entry. If a further pair lands between the poll resolving and this read (the ResizeObserver genuinely stays attached here, since `observe()` isn't skipped for a main-document/`isScrollOnWindow`-false grid), `.at(-1)` is `'before'` or a stale `after` with different values, and the assertion breaks despite correct grid behavior.

Fix: filter for the last `after` entry instead of the array's last element, e.g. `(await grid.hookLog()).filter(e => e.hook === 'after').at(-1)`.

Same pattern repeats at lines 124-129 for the body-hidden variant.

This reproduces the exact loaded-runner flake class (passes on an idle CI runner, reds on a loaded one) this migration exists to eliminate.

tests/e2e/refresh-dimensions.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+    await grid.setBodyDisplay('');
+    await expect.poll(() => grid.hookLogLength()).toBeGreaterThanOrEqual(2);
+
+    expect((await grid.hookLog()).at(-1)).toEqual(
+      entry('after', { width: 200, height: 100 }, { width: 240, height: 100 }, true),
+    );
+  });
+});
+
+test.describe('a grid inside an iframe (window-resize pipeline)', () => {
```
demtario, line 133: **Medium.** The iframe/window-resize pipeline lost its block-refresh coverage. The main-document pipeline gets a dedicated test asserting `beforeRefreshDimensions` returning `false` blocks the refresh (lines 62-89), but no equivalent exists in this `describe` block, even though `buildIframeGrid()` and the fixture already support `{ blockRefresh: true }` (`RefreshDimensionsPage.ts:59`, `refresh-dimensions.html:88`) - it's just never called with it. The deleted `beforeRefreshDimensions.spec.js` (iframe `describe`, lines 178-193) had this exact scenario.

Fix: add a test calling `buildIframeGrid({ blockRefresh: true })` + `setIframeWidth(50)`, asserting only `before` logs and that the grid's measured size doesn't adopt the new viewport.

tests/e2e/refresh-dimensions.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+  test('fires when the iframe window shrinks, reporting the measured viewport', async({ page, theme, bundle }) => {
+    const grid = new RefreshDimensionsPage(page, theme, bundle);
+
+    await grid.goto();
+    await grid.buildIframeGrid();
+    await grid.clearHookLog();
+
+    await grid.setIframeWidth(50);
+
+    await expect.poll(() => grid.hookLogLength()).toBe(2);
```
demtario, line 147: **Low.** Same assertion shape as the finding at line 112 (exact `toBe(2)` poll, then a separate full-log read asserted precisely) - flagging for consistency, though unlike that case I couldn't find a concrete second-trigger path here: the iframe's window-scroll grid never attaches a ResizeObserver (gated off by `isScrollOnWindow` in `nativeScrollInput.ts`), and only one `resize` event is dispatched by `setIframeWidth(50)`. Worth a second look if this ever flakes, but not a confirmed bug.

tests/e2e/refresh-dimensions.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+/**
+ * Migrated from the frozen Jasmine specs `{before,after}RefreshDimensions.spec.js` (DEV-2744,
+ * flake ledger of DEV-2668). Their fixed `sleep(50)` waits raced two pipelines that this spec
+ * polls instead: the ResizeObserver of a sized grid (which also delivers ONCE on observe - the
+ * call the legacy synchronous asserts were actually counting), and the window-resize path of a
+ * grid whose document is an iframe's.
+ *
+ * `before` and `after` are asserted together through one interleaved log, which also pins their
+ * ORDER - something the legacy per-hook files never could. The dvh loop-guard spec stays in the
+ * legacy suite until DEV-2740 fixes the guard it tests.
```
demtario, line 20: **Nit.** The deleted legacy spec's rAF-sync test (`spyOn(window, 'requestAnimationFrame')`, asserting the hook count is still 0 immediately after resize and only becomes 1 after the rAF tick - `beforeRefreshDimensions.spec.js:121-137`) has no direct replacement. The "resize doesn't fire the hook synchronously, only after one rAF" contract is now only implicit via `expect.poll`. Optional - only worth restoring if that timing guarantee is considered a contract worth its own regression test.

demtario (commented):
4 inline notes below.

tests/e2e/refresh-dimensions.spec.ts
danielzytohoc: Fixed in 58cfbe36c, and you're right that it was exactly the class this migration exists to kill. Both hidden-then-revealed phases now poll the **latest `after` entry** through a new page-object probe (`lastEntry(hook)`: `htHookLog.filter(e => e.hook === which).at(-1)`), asserted directly with `expect.poll(...).toEqual(entry('after', …))` — no separate length poll, no second round trip to race. The observer indeed stays attached on the main-document grid, so the latest completed refresh is the only thing worth pinning there.

tests/e2e/refresh-dimensions.spec.ts
danielzytohoc: Restored in 58cfbe36c as its own test (`returning false from beforeRefreshDimensions blocks the window-resize refresh`). One wrinkle worth knowing: for a window-scroll grid the "measured size doesn't adopt the viewport" half has no DOM observable — the root and holder widths follow the iframe viewport through CSS layout whether the refresh runs or not (probed: both go 500→50 with the refresh entirely starved). So the fixture's `buildIframeGrid()` gained a `columns` option: 20 columns against the 500 px viewport makes horizontal virtualization the observable. The unblocked shrink test now asserts the rendered column count **drops** (the positive control), and the blocked test asserts exactly one `before` entry (prev `{500,0}`, `curr.width` equal to the measured viewport, action `true`), no `after` after a bounded settle, and the rendered column count **unchanged** — the block visibly reached the view. 420/420 on the full matrix at `--repeat-each=10`.

tests/e2e/refresh-dimensions.spec.ts
danielzytohoc: Applied the same shape in 58cfbe36c for consistency: the pinned-grid case polls `lastEntry('after')` and then reads `lastEntry('before')`, instead of a length poll plus a whole-log read. Agreed there's no concrete second trigger on that path today (no observer on a window-scroll grid, one `resize` per width change) — one idiom for every multi-delivery-capable trigger is cheaper than remembering which ones are safe.

tests/e2e/refresh-dimensions.spec.ts
danielzytohoc: Restored in 58cfbe36c in its observable form rather than with a `requestAnimationFrame` stub: `resizeRoot()` now returns the hook-log length read synchronously in the same `evaluate` as the mutation, and the resize test asserts it is `0` before polling for the deferred pair. That pins the contract the legacy test meant — a resize never fires the hooks in its own task — without asserting the scheduling primitive that happens to implement it.

demtario (approved):

PR #13335, opened 2026-09-02
DEV-2747: Add per-plugin AGENTS.md knowledge files

budnix:
### Context

The PR gives every plugin under `handsontable/src/plugins/` its own `AGENTS.md` knowledge file, with `CLAUDE.md` symlinked to it — the convention already used across the monorepo.

Before this change only 10 of the 44 plugin directories had one. Everything else an agent needed to know about a plugin either lived in the monorepo-scoped `handsontable/AGENTS.md`, which cannot hold per-plugin depth, or nowhere at all. Two pieces of shared knowledge had no home anywhere: the plugin contract itself, and the `PLUGIN_PRIORITY` table — whose ordering is load-bearing in four places, with nothing recording that fact.

**34 new `AGENTS.md` files + 34 `CLAUDE.md` symlinks. Coverage is now 44/44.** The 10 pre-existing files were not touched.

Two of the 44 directories are not plugins, and they now carry the knowledge that was previously homeless:

- **`plugins/base/AGENTS.md`** — the plugin contract: the five statics and what each one actually controls, the lifecycle in the order it runs, the four-branch `onUpdateSettings` state machine, the hard-conflict registry with both registration sites, and the full `PLUGIN_PRIORITY` table for all 42 plugins — including the four orderings that are load-bearing, the free slots for a new plugin, and the distinction between priority ordering `enablePlugin()` and hook-callback order.
- **`plugins/manualResize/AGENTS.md`** — the helpers the two resize plugins share, plus the issue #4371 `SETTING_KEYS` rules and the load-bearing one-pixel tolerance in the scale math. That directory holds no plugin at all, which is itself worth stating.

Each file follows the shape of the existing ones: what the plugin owns and what it does not, the traps with the reason behind each, cross-links, and the targeted `testPathPattern` commands.

#### Knowledge moved down, not copied

Five bullets in `handsontable/AGENTS.md` were plugin-owned. The detail moved into the plugin that owns it and a one-line pointer stayed behind, so nothing is duplicated and nothing is lost:

| Was in `handsontable/AGENTS.md` | Now lives in |
|---|---|
| Formulas undo/redo bypass, and the `validateDependentCells` open defect (2 bullets) | `formulas/`, `undoRedo/` |
| `SETTING_KEYS` manual-resize consequences | `manualResize/` |
| Dialog `template.id` and the button-type render sites | `dialog/`, `emptyDataState/`, `notification/`, `loading/` |
| `textExtractor` traps | `exportFile/`, `copyPaste/` |

The merged-cells and Filters one-liners were kept at the top level, since they are worth having there, and extended with pointers to the fuller treatment.

`.ai/CONCERNS.md` was deliberately left intact. It is a cross-cutting inventory by design, so the plugin files cite the relevant section rather than absorb it.

#### Conventions recorded, and the file made mandatory for new plugins

The convention is written down where each kind of reader will meet it:

- A paragraph under Plugin Lifecycle in `handsontable/AGENTS.md`, the `src/plugins/` entry and the "New Plugin" checklist in `handsontable/.ai/STRUCTURE.md`, and a routing row in the root `AGENTS.md` so the lookup is one hop instead of two.
- **Root `AGENTS.md` checklist item 3 now states two unconditional cases.** It used to be conditional — "if a change introduces new conventions… update the `AGENTS.md` at the correct scope" — which never actually required the file. It now says a **new plugin must ship its own `AGENTS.md` with `CLAUDE.md` symlinked to it**, and that a trap found while changing an existing plugin belongs in *that plugin's* file rather than the monorepo-wide ones. This file loads every session, so that sentence is where the rule bites.
- **The `handsontable-plugin-dev` skill carries the how.** A new `## Knowledge file (required)` section lists the sections every file carries and says explicitly that **length follows the plugin, not a template** — `stretchColumns` is three bullets and complete, `base` is long because it holds the contract for 42 plugins. Two new Registration Checklist steps: write the file and `ln -s AGENTS.md CLAUDE.md` (relative target, so it resolves in a worktree; mode `120000`, so a regular file in `git status` means you copied), and **add your row to the `PLUGIN_PRIORITY` table in `base/AGENTS.md`** — nothing enforces that table, so it drifts silently otherwise.
- Two corrections while in that skill: `PLUGIN_PRIORITY` was described as "Execution order (higher = later)", which invites the assumption that it orders hook callbacks. It orders `enablePlugin()`, and hook order only transitively. And the `SETTING_KEYS` row now shows `[PLUGIN_KEY]` as *the default*, so a reader knows to declare it only when they need something else.
- `handsontable-dev`'s file-layout block lists the two files as required.

**No source change — this PR touches only `.md` files and symlinks to them.** Both gates classify it that way on path, so neither a test nor a `Refactor-only:` trailer is required:

- The **test-presence gate** does not fire: its `SOURCE` patterns require `.(ts|js|tsx)$` (`.github/scripts/lib/presence-gate.mjs`). Confirmed green on push — `✔️ test-gate`.
- The **changelog gate** passes automatically: `/\.md$/i` is in its never-relevant list (`.github/scripts/lib/changelog-gate.mjs`), so **no `.changelogs/*.json` entry is needed and no skip marker is used**.

- Unit tests added/modified (`*.unit.js`): none — no source changed
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`): none — no source changed
- Type tests (`*.types.ts`) updated if public API changed: none — no public API changed
- For a bug fix — the spec that fails without this fix: n/a

What was verified instead, since correctness here means the documents are accurate and navigable:

1. **Coverage is complete** — 44 directories, 44 `AGENTS.md`, 44 `CLAUDE.md`, every symlink resolving to the relative `AGENTS.md`.
2. **The symlinks match the repo convention byte for byte** — the 34 new `CLAUDE.md` entries are mode `120000` with blob `47dc3e3d`, identical to the pre-existing `filters/CLAUDE.md` and `nestedHeaders/CLAUDE.md`.
3. **Every navigational cross-link resolves** — all `../<plugin>/AGENTS.md`, `../../../AGENTS.md`, `../../../.ai/*.md`, `../../core/layout/`, `../../translations/AGENTS.md` and `../../../../tests/AGENTS.md` paths were checked from each file's own directory.
4. **The `PLUGIN_PRIORITY` table is complete and correct** — diffed against every `export const PLUGIN_PRIORITY` in the tree: 42 plugins, no entry missing, no entry that is not a plugin.
5. **Nothing reaches the npm package** — `handsontable.copy` in `package.json` pulls only `dist/`, `styles/`, `languages/` and three root files, so no `src/**` markdown ships.
6. **Facts were sourced from the code, not inferred.** Four claims were corrected on review, including the strict-vs-loose bind-map semantics, which was verified against `strictBindsMap.ts` rather than taken from the option name — strict never renumbers existing headers, so a removal leaves a gap in the sequence.

### Commands run

```
# Coverage + symlink integrity (from handsontable/src/plugins)
for d in */; do d=${d%/}; [ "$d" = "__tests__" ] && continue
  [ -f "$d/AGENTS.md" ] || echo "MISSING AGENTS: $d"
  [ -L "$d/CLAUDE.md" ] && [ "$(readlink "$d/CLAUDE.md")" = "AGENTS.md" ] || echo "BAD LINK: $d"
done
# → no output; dirs=44 agents=44

# Symlink mode/blob matches the existing convention
git ls-files -s handsontable/src/plugins/{filters,base}/CLAUDE.md
# → both 120000 47dc3e3d863cfb5727b87d785d09abf9743c0a72

# PLUGIN_PRIORITY table vs. reality — both diffs empty
comm -23 /tmp/tbl.txt /tmp/act.txt   # in table but not a plugin
comm -13 /tmp/tbl.txt /tmp/act.txt   # a plugin but missing from table
```

Pre-push output: `pre-push: ok` / `✔️ test-gate (0.39 seconds)`.

Lint and build were not run: ESLint and Stylelint have no `.md` inputs, and no source file changed, so the build output is unaffected.

### Types of changes

None of the four boxes applies. This is agent-facing documentation only — no shipped code, no public API, no user-visible behavior.

### Related issue(s):

1. DEV-2747

### Affected project(s):

- [x] `handsontable` — repository paths only; no shipped code or published artifact changes
- [ ] `@handsontable/vue3`

### Checklist:

- [ ] My change requires a change to the documentation. — this PR *is* the documentation change; no user-facing guide or API reference is affected.

### Note for the reviewer

Two things worth a look:

1. **The `.ai/STRUCTURE.md` edit is slightly wider than the task.** While adding the knowledge-file convention I also corrected the "New Plugin → E2E tests" line, which still pointed at the frozen Jasmine path instead of Playwright. It matches the root `AGENTS.md`, but it is easy to drop if you would prefer a narrower diff.
2. **"Required" here is documented, not machine-enforced.** Everything else in the mandatory checklist has a gate behind it, and this does not — a new plugin can still land without its `AGENTS.md` and nothing will complain. A `handsontable/test/__tests__/pluginKnowledgeFiles.unit.js` asserting that every `src/plugins/*/` with a `PLUGIN_KEY` has an `AGENTS.md`, a `CLAUDE.md` symlink pointing at it, and a row in `base/AGENTS.md`'s priority table would close that, in about 25 lines, and matches the idiom already used by `esTarget.unit.js` and `previewPackaging.unit.js`. The table-membership half is the part that earns its place, since that is what drifts. Happy to add it in this PR if you want it — left out for now because it was outside the ask.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2747

---

> [!NOTE]
> **Low Risk**
> Documentation and symlink-only changes; no runtime, API, or published package artifacts are modified.
> 
> **Overview**
> **Adds per-plugin agent knowledge files** so every directory under `handsontable/src/plugins/` has an `AGENTS.md` (with `CLAUDE.md` symlinked to it). This closes the gap where only 10 of 44 plugin folders had local guidance; coverage is now **44/44**, including **`base/AGENTS.md`** (shared plugin contract, `onUpdateSettings` branches, hard conflicts, full **`PLUGIN_PRIORITY`** table) and **`manualResize/AGENTS.md`** (shared resize helpers and `SETTING_KEYS` traps).
> 
> **Refactors monorepo docs** instead of duplicating facts: long plugin-specific bullets in `handsontable/AGENTS.md` are shortened to pointers into the owning plugin’s file (formulas/undo, dialog/notification, export/copyPaste, manual resize, etc.). Root `AGENTS.md`, `handsontable/.ai/STRUCTURE.md`, and the **`handsontable-plugin-dev`** / **`handsontable-dev`** skills now **require** a new plugin to ship `AGENTS.md` + symlink and to register its priority row in `base/AGENTS.md`; the skill also clarifies that **`PLUGIN_PRIORITY` orders `enablePlugin()`**, not hook callbacks directly, and softens bulk-rewrite guidance for existing inline hook wrappers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 54ed4488bff1e456e9d67f72d983fb031a338fdf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2747 Add per-plugin AGENTS.md knowledge files](https://app.clickup.com/t/9015210959/DEV-2747)

handsontable/src/plugins/copyPaste/AGENTS.md
```diff
@@ -0,0 +1,119 @@
...
+the markup is read. **Never `importNode` those nodes into the live document** — that makes them live again.
+Background in `../../../.ai/CONCERNS.md`.
+
+**`parseFromString` is itself a Trusted Types sink, so the source-data parse is wrapped in a `try`/`catch`
+that degrades instead of failing** (DEV-2617). Under `require-trusted-types-for 'script'` it throws unless
+the value came from a policy — which it did not when no `sanitizer` is configured, or when one is configured
+and returns a plain string. Losing the source-data flavor costs object-key fidelity on an internal paste;
+letting the throw escape would kill the paste outright, so the catch warns
+(`#warnClipboardParseRefused`) and the `text/html` branch carries the paste. Do not "tidy" that catch away,
+and do not widen it to the `text/html` parse — that one has no fallback left.
```
demtario: **High:** this claim inverts the actual code. `parseFromString` for `text/html` is already wrapped in a `try`/`catch` with a `readPlainText` fallback (`copyPaste.ts` ~1052-1067), so "that one has no fallback left" is wrong — it has one. Please correct or drop this sentence so future edits to the Trusted Types handling here aren't misled.

handsontable/src/plugins/columnSummary/AGENTS.md
```diff
@@ -0,0 +1,88 @@
...
+
+## The refresh pass caches every endpoint, not just the matched ones
+
+`cacheSummaryDestinations(endpoints)` is called with **all** endpoints even though only the matched ones are
+refreshed. A summary result must stay excluded from the ranges of the endpoints being refreshed, whatever
+its own source column is.
+
+## `afterCreateRow` ordering with TrimRows
+
+TrimRows' `afterCreateRow` fires *after* this plugin's, so the range recreation is deferred to the next
```
demtario: **Medium:** `trimRows.ts` registers zero `addHook` calls (no hooks at all), so it can't be firing an `afterCreateRow` listener after this plugin's. This looks lifted from a stale/misattributed comment in `endpoints.ts:393` — `nestedRows.ts` (also a trimmer) does register `afterCreateRow`, TrimRows doesn't. Please correct or remove.

handsontable/src/plugins/emptyDataState/AGENTS.md
```diff
@@ -0,0 +1,73 @@
...
+
+- **`wheelDeltaX` is non-standard** (Safari). It is the fallback when `event.deltaX` is `NaN`, and the sign
+  is inverted (`-1 * wheelDeltaX`).
+- **Only forward when the table actually has a horizontal scroll and is not window-scrollable**
+  (`hasHorizontalScroll() && !isHorizontallyScrollableByWindow()`), then `preventDefault()`. Without those
+  guards the page stops scrolling.
+
+## Selection on hide
+
+Hiding the placeholder re-selects `(0, 0)` with scrolling suppressed, then fires
```
demtario: **Medium:** this only describes the fallback path. `#hide()` (`emptyDataState.ts` ~580-600) first restores the selection captured on show via `importSelection` when one exists; `selectCell(0, 0, …)` only runs when there was none. As written this implies hiding always discards the prior selection.

handsontable/src/plugins/copyPaste/AGENTS.md
```diff
@@ -0,0 +1,119 @@
...
+  Identity is what survives that — do not switch the check to a content comparison.
+- `modifyCopyableRange` is the hook for constraining what may be copied (DEV-844).
+
+Paste sizing is inherently two-phase: the plugin tries to populate all copied data, or repeat it within the
+selection, but **it cannot know up front whether the populated data exceeds the selection** — some cells
+reject values, and that is only known after reading their cell meta.
+
+## Header copying
+
+Four options control it — `copyColumnHeaders`, `copyColumnGroupHeaders`, `copyColumnHeadersOnly`, and
```
demtario: **Medium:** `pasteMode` doesn't belong in this list — it only controls how pasted data is inserted relative to the selection (`populateFromArray(..., this.pasteMode)`, `copyPaste.ts:752`) and has no connection to header copying.

handsontable/src/plugins/autofill/AGENTS.md
```diff
@@ -0,0 +1,77 @@
...
+A fill area may start or end on a hidden cell. The highlighted selection stores **renderable** indexes only
+(that is Walkontable's contract), so there is no record of the hidden endpoints — the plugin extrapolates
+where the start and end are. Any change to range computation has to keep that extrapolation, or a fill that
+touches a hidden row silently loses a row.
+
+## Dragging past the table's edge (DEV-2024)
+
+Drag counting is driven by `beforeOnCellMouseOver`, and that hook **does not fire once the pointer leaves
+the table element** — dragging the last column's fill handle at a slight angle is enough. The drag is
+therefore counted a second time on the mouse-up path. Without it, a fill made with such a drag is never
```
demtario: **Medium:** the fallback increment (`handleDraggedCells += 1`) that handles a pointer leaving the table lives in `#onMouseMove` (`autofill.ts` ~908-910), not `#onMouseUp` — `#onMouseUp` only reads the already-accumulated count. Attributing it to "the mouse-up path" points a future debugger at the wrong function.

handsontable/src/plugins/base/AGENTS.md
```diff
@@ -0,0 +1,215 @@
...
+  disabled, and the reason belongs in a comment.
+- **Prefer arrow function class fields, passed directly**: `this.addHook('afterX', this.#onAfterX)`. This is
+  the house rule (`../../../AGENTS.md` states it as Required) and it is what makes a listener nameable,
+  greppable and individually removable with `removeHooks(name)`. Two corrections to how it is usually
+  justified, both worth knowing before you enforce it:
+  - **The plugin's own `addHook` does not care about identity.** It pushes whatever reference it was handed
+    into `#hooks`, and `removeHooks` removes that same reference — so an inline wrapper registered through
+    `this.addHook` *is* cleaned up by `disablePlugin()`. Identity only matters when you register on
+    `this.hot.addHook` and later call `this.hot.removeHook` with a separately built function.
+  - **About 27 inline-wrapper sites already exist**, across 11 plugins (autoColumnSize, autoRowSize,
```
demtario: **Low:** says "across 11 plugins" but the parenthetical names 12 (autoColumnSize, autoRowSize, columnSummary, formulas, emptyDataState, dragToScroll, comments, customBorders, mergeCells, dialog, contextMenu, dropdownMenu).

demtario (commented):
6 inline notes below.

handsontable/src/plugins/copyPaste/AGENTS.md
budnix: You're right, that one has a fallback of its own. Fixed in bf6c06b32d — both catches are described now, including the empty `text/plain` guard.

handsontable/src/plugins/columnSummary/AGENTS.md
budnix: Good spot — `trimRows.ts` registers no hooks at all. Fixed in bf6c06b32d, which now names `nestedRows` and flags the `endpoints.ts:393` comment as stale.

handsontable/src/plugins/emptyDataState/AGENTS.md
budnix: Good point — the `importSelection` path was missing. Fixed in bf6c06b32d; `selectCell(0, 0)` now reads as the fallback it is.

handsontable/src/plugins/copyPaste/AGENTS.md
budnix: Agreed, `pasteMode` has nothing to do with headers. Fixed in bf6c06b32d — three header options, with `pasteMode` split out as the paste-side one.

handsontable/src/plugins/autofill/AGENTS.md
budnix: Yep, my mistake — the increment is in `#onMouseMove`. Fixed in bf6c06b32d, which now also covers the `countDragStep` gate and the scroll replay.

handsontable/src/plugins/base/AGENTS.md
budnix: Nice catch — both numbers were wrong, it's at least 38 sites across 16 plugins. Fixed in bf6c06b32d, where I also dropped a false claim in that same sentence about two of them staying bound past `disablePlugin()`.

demtario (approved):

PR #12161, opened 2026-03-13
E2e test sleep replacement

qunabu:
### Context
This change introduces a new `waitForNameAnimationFrames()` helper to e2e tests, aiming to replace non-deterministic `sleep()` calls with frame-based waits for improved test speed and reliability. Initially, a global replacement of `sleep()` was attempted but caused regressions due to timing sensitivities in various tests. The current approach adds the new helper and explicitly migrates `sleep()` calls in specific e2e specs where frame-based waiting is safe and beneficial, while preserving `sleep()`'s original time-based behavior for other tests.

### How has this been tested?
- **Unit test:** `BABEL_ENV=commonjs npx env-cmd -f ../hot.config.js npx jest --testPathPattern=src/helpers/__tests__/function.unit.js` (11 passed)
- **E2E (migrated hooks):** `npm run test:e2e.dump --testPathPattern="hooks/(afterOnCellMouseOver|beforeOnCellMouseOver|afterChange).spec.js" && npm run test:e2e.puppeteer` (11 specs, 0 failures)
- **E2E (sanity check on timing-sensitive validator):** `npm run test:e2e.dump --testPathPattern=settings/validator.spec.js && npm run test:e2e.puppeteer` (51 specs, 0 failures)

### Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature or improvement (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Additional language file or change to the existing one (translations)

### Related issue(s):
1.
2.
3.

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular-wrapper`
- [ ] `@handsontable/react-wrapper`
- [ ] `@handsontable/vue3`

### Checklist:
- [x] I have reviewed the guidelines about [Contributing to Handsontable](https://github.com/handsontable/handsontable/blob/master/CONTRIBUTING.md) and I confirm that my code follows the code style of this project.
- [x] I have signed the [Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLScpMq4swMelvw3-onxC8Jl29m0fVp5hpf7d1yQVklqVjGjWGA/viewform?c=0&w=1)
- [ ] My change requires a change to the documentation.

[Slack Thread](https://handsoncode.slack.com/archives/D0AKSFNH1JQ/p1773423492213519?thread_ts=1773423492.213519&cid=D0AKSFNH1JQ)

[...]

PR #13336, opened 2026-09-02
DEV-2745: Migrate the display-none rerender flaky specs to Playwright

danielzytohoc:
### Context

The PR migrates the second family on the CI-flakiness investigation's flake ledger to Playwright: the four `Core_init` "rerender after `display` changes from `none`" variants (Flake 4 — twice red on Integration/ESM trunk runs, and since 08-28 also on a PR leg) and the `observeVisibilityChangeOnce` DOM-helper group (a Flake 5 single — the same mechanism tested one level lower; both were created together in PR #10130).

Root cause on the test side: `await waitForNextAnimationFrames(2)` — ~32 ms, with the frame count silently capped at 2 in `test/helpers/common.js` — used as a proxy for IntersectionObserver delivery, which is not frame-bound. The waits were shortened from `sleep(100)` by the mass conversion in PR #12161 (2026-03).

The new `tests/e2e/hidden-init-rerender.spec.ts` (+ fixture + page object) polls the observable effect instead:

- The four hiding shapes (inline style vs stylesheet rule × root vs parent) are one parametrized loop against a `?hidden=` fixture allowlist. Each test pins the premise first (`offsetParent === null` — without it the scenario silently degrades into a plain visible init), reveals, polls the top-clone holder gaining a layout box, then asserts the legacy geometry contract (holder covers the header row; stretched header widths equal the first data row's; holder spans the cloned table) plus one strengthening: every stretched column must exceed the 50 px unstretched default, so a rerender that painted without measuring fails.
- The helper group runs through a counting IntersectionObserver **wrapper** (the real observer still runs): fires exactly once despite same-frame flickers; disconnects rather than unobserves after the first delivery; no observer leak with a zero-height body. Positive polls precede every bounded once-ness settle.

Legacy attrition: the four `init.spec.js` tests and the `observeVisibilityChangeOnce` describe in `element.spec.js` are removed, each replaced by a pointer comment. `hidden-init-destroy.spec.ts` (the destroy-race spec) is untouched — it covers the destroy race, not the reveal rerender.

Worth flagging for reviewers: during authoring, the spec's own `goto()` initially used a default (rAF-polled) `waitForFunction` and timed out 3 times in ~700 executions under parallel workers — never in isolation, never sequentially hammered (60×), never with 6 concurrent pages (120×). Switching the readiness wait to **timer polling** and making the fixture build fail loud (constructor throws are captured and rethrown with a page snapshot) produced 420/420 across the full matrix at `--repeat-each=10`. If it ever recurs in CI, the failure now names itself (build error vs page snapshot) instead of being a mute timeout. The rAF-polled `waitForFunction` default may be a tier-wide latent flake source worth a follow-up lint/skill rule — noted in DEV-2745's ClickUp comment.

### How has this been tested?

- `cd tests && npx playwright test --project=e2e-main e2e/hidden-init-rerender.spec.ts` — 7/7.
- Full matrix soak: `npx playwright test e2e/hidden-init-rerender.spec.ts --repeat-each=10` (all six legs) — **420/420 passed**.
- Trimmed legacy files under the Puppeteer runner: `npm run test:e2e --testPathPattern='core/init|dom/__tests__/element'` — the only failures are the two pre-existing `has{Vertical,Horizontal}Scrollbar` tests, which fail identically on **pristine develop** on macOS (overlay scrollbars; verified by stash-and-rerun) and are green on CI Linux. Untouched by this PR.
- `npx eslint` on all five changed/new files — clean.

### Types of changes

### Related issue(s):

1. DEV-2745
2. The CI-flakiness investigation task (linked in ClickUp) — the flake ledger this executes

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2745


qunabu:
Task linked: [DEV-2745 Migrate the display-none rerender flaky specs (Core_init + observeVisibilityChangeOnce) to Playwright](https://app.clickup.com/t/9015210959/DEV-2745)
Task linked: [DEV-2210 Disconnect the hidden-init visibility observer on destroy and guard its callback](https://app.clickup.com/t/9015210959/DEV-2210)

demtario (approved):

PR #13337, opened 2026-09-02
DEV-2746: Deflake the Interval unit tests and the autocomplete case-sensitivity wait

danielzytohoc:
### Context

The PR removes the last two test-side items on the CI-flakiness investigation's flake ledger (part B of that plan, after the two migration PRs, #13334 and #13336):

1. **`Interval.unit.ts` — 3 CI incidents in 4 days (08-28 → 08-31).** The timing tests asserted real-`setTimeout` wall-clock checkpoints (`expect(spy).not.toHaveBeenCalled()` at 50 ms against a 100 ms interval), so any ≥50 ms event-loop stall on a loaded runner flipped them. Rewritten on Jest fake timers: `Interval` ticks on `requestAnimationFrame` (read from `window` at call time, so the mocked one) gated by `Date.now()`, and `advanceTimersByTime()` moves both in lockstep. The assertions became **exact** — the old stop/start test's "2 or 3 calls, timing can vary" tolerance is now `toBe(2)`, so the tests got stronger, not just quieter. Jest tier is not frozen; this is an in-place fix by policy.

2. **AutocompleteEditor "default filtering should be case sensitive when filteringCaseSensitive is true" — 1 incident.** Its `waitForNextAnimationFrames(2)` (~32 ms) raced the editor's deferred query and the inner grid's render. Routed as a frozen-suite edit (the ledger shows one hit, not a recurrence, and the ticket allowed this route): the fixed waits become `waitUntil()` polls on the asserted list state itself — settled at 9 rows after the lowercase query, emptied after the case-sensitive uppercase one — so the poll condition cannot mask the behavior under test.

This adds the **`waitUntil(condition, timeout)` helper to `test/helpers/common.js`** (auto-mounted as a spec global like its siblings): the condition-based replacement the frozen tier never had, which is why fixed sleeps and frame-count waits kept being the path of least resistance (the enforcement audit behind the CI-flakiness investigation task flagged the missing "legal alternative" explicitly). It rejects with a named reason on timeout instead of letting a stale assertion fail later.

### How has this been tested?

- `npm run test:unit --testPathPattern=Interval` — 6/6 in 0.6 s (previously ≥1 s of real waiting), exact call counts.
- `npm run test:e2e --testPathPattern=autocompleteEditor` — 140 specs, 0 failures.
- `npx eslint` on all three files — 0 errors (12 pre-existing warn-level debt findings elsewhere in the autocomplete file, untouched).

### Types of changes

### Related issue(s):

1. DEV-2746
2. The CI-flakiness investigation task (linked in ClickUp) — the flake ledger this closes out, test-side

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2746

### DIFF
--- a/handsontable/src/editors/autocompleteEditor/__tests__/autocompleteEditor.spec.js
+++ b/handsontable/src/editors/autocompleteEditor/__tests__/autocompleteEditor.spec.js
@@ -2426,7 +2426,9 @@ describe('AutocompleteEditor', () => {
       editorInput.val('e');
 
       await keyDownUp('e'); // e
-      await waitForNextAnimationFrames(2);
+      // Every choice contains a lowercase 'e', so the queried list settles at the full 9 rows.
+      // Poll for that state instead of guessing how long the editor's deferred query takes.
+      await waitUntil(() => getActiveEditor().htEditor?.getData().length === 9);
 
       {
         const ac = getActiveEditor();
@@ -2449,7 +2451,9 @@ describe('AutocompleteEditor', () => {
         await keyDownUp('e'); // E (same as 'e')
       }
 
-      await waitForNextAnimationFrames(2);
+      // The case-sensitive query for an uppercase 'E' matches nothing - the emptied list IS the
+      // asserted behavior, so polling for it cannot mask a failure.
+      await waitUntil(() => getActiveEditor().htEditor?.getData().length === 0);
 
       {
         const ac = getActiveEditor();
--- a/handsontable/src/utils/__tests__/Interval.unit.ts
+++ b/handsontable/src/utils/__tests__/Interval.unit.ts
@@ -1,6 +1,18 @@
 import Interval from '../interval';
 
+// Fake timers drive the whole loop deterministically: `Interval` ticks on `requestAnimationFrame`
+// (read from `window` at call time, so the mocked one) and gates on `Date.now()` elapsed time,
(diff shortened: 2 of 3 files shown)


qunabu:
Task linked: [DEV-2744 Migrate the flaky RefreshDimensions legacy specs (iframe + root resize) to Playwright](https://app.clickup.com/t/9015210959/DEV-2744)
Task linked: [DEV-2745 Migrate the display-none rerender flaky specs (Core_init + observeVisibilityChangeOnce) to Playwright](https://app.clickup.com/t/9015210959/DEV-2745)
Task linked: [DEV-2746 Deflake the Interval unit tests (fake timers) and the AutocompleteEditor case-sensitivity wait](https://app.clickup.com/t/9015210959/DEV-2746)

demtario (approved):

danielzytohoc:
Fresh data point for this PR's Interval half: the wall-clock `Interval.unit.ts` tests just failed the `Unit / test` gate on an unrelated branch again (run 33636723237, PR #13334's develop-merge push — 2 failed / 4311, the `start`/`stop` checkpoint asserts). That's the 4th incident since 08-28 and the first to block a PR pipeline outright; the branch does not contain this PR's fake-timers rewrite. Worth landing this one early — it removes the cause rather than re-running around it.

PR #13338, opened 2026-09-02
DEV-2749: Re-run the visual gate when visual-approved is added

budnix:
### Context

The PR makes the `visual-approved` label do the whole job, instead of being step one of two.

Today, adding the label changes nothing on its own. The check stays red until someone opens the workflow run and presses **Re-run jobs** by hand — and nothing on the pull request says so except one line in the bot comment.

The label itself was never broken. The gate in `visual.yml` ("Visual approval gate") reads the labels **live** from the API, precisely so that a re-run after labelling can see them. The gap is that only a *new attempt* runs the gate at all, and nothing starts one. So a labelled pull request keeps its red `Visual / Compare`, and with it a red `CI Gate`, indefinitely.

Seen on #13325: attempt 2 of run `33610031096` had exactly two failed jobs, `Visual / Compare` and `CI Gate`. Attempt 3 — a manual re-run, no code change — was green.

`.github/workflows/visual-approval-rerun.yml` closes that gap. It fires on `pull_request: [labeled]` and re-runs the failed jobs for you.

Two things make it more than a single API call:

1. **It has to wait.** The sticky comment that prompts the approval is posted by `Compare`, which is not the last job in the pipeline. So at label time the run is usually still `in_progress`, and `rerun-failed-jobs` rejects a run that has not finished. It polls (60s, up to 90 minutes) until the run completes.
2. **It re-reads everything on every poll** — the label, the head commit, the pull request state, the run and the job. A push during the wait cancels the pending re-run rather than approving screenshots nobody reviewed.

Design notes worth flagging for review:

- **The push check compares against `LABELED_HEAD_SHA`, the head as the `labeled` event saw it — not against the run's own `head_sha`.** This is load-bearing and easy to get wrong: the script looks the run up *by* the live head, so `run.head_sha === prHeadSha` always holds and a check against it is dead code that would silently follow the new commit. The first revision of this PR had exactly that bug.
- **`rerun-failed-jobs`, not a single-job re-run.** `CI Gate` lives in the calling workflow and is failed too, so this endpoint picks both up with their `needs` order intact. Whether a single-job re-run reaches across the reusable-workflow boundary to the caller's dependent job is not documented, and it is not worth betting the merge gate on. The cost is that an unrelated failed job is re-run alongside, and a red job that has nothing to do with screenshots keeps `CI Gate` red — the wording in the bot comment and the README says so rather than promising a green pipeline.
- **The concurrency group is keyed by label name as well as pull request number.** `labeled` fires for *every* label, and a run's concurrency group is resolved when the run is created, before the job's `if:` skips it. Keyed on the number alone, someone adding `bug` or `docs` mid-wait would cancel the waiter and then skip its own job — nothing re-runs, and nothing reports why.
- **Fork and Dependabot pull requests are excluded, and Dependabot needs two clauses.** On a `labeled` event `github.actor` is whoever *applied the label*, so the canonical guard's `github.actor != 'dependabot[bot]'` half does not exclude a Dependabot pull request — a maintainer labels it and the guard passes. The re-run would be spent for nothing, because a re-run keeps the run's original actor, so `visual.yml`'s `IS_UNTRUSTED` stays true and the gate still refuses the label. The job therefore also carries `github.event.pull_request.user.login != 'dependabot[bot]'`, and keeps the canonical clause because `fork-guards.test.mjs` asserts that shape. Any future `labeled`-triggered workflow needs the same pair.
- **The run lookup filters on `head_branch`, not `pull_requests`.** GitHub leaves `run.pull_requests` **empty** on same-repo `pull_request` runs (verified: run `33610031096` returns `[]`), so a filter on it is a no-op that reads like a check. `head_branch` is always populated and separates two pull requests raised from *different* branches at the same commit. It does not separate one branch raised against two bases — that residual case is documented in the script rather than pretended away; the worst outcome there is a re-run spent on the sibling pull request, since each run's gate reads its own pull request's labels.
- **Closed pull requests are skipped.** `pr-cleanup.yml` purges `pr-<n>/` from R2 on close and fires only on `closed`, so a later re-run would republish screenshots that nothing will ever purge again.

All branching lives in `.github/scripts/lib/visual-approval-rerun.mjs`, which is pure and unit-tested; the wrapper only talks to the API and sleeps. This follows the `visual-gate.mjs` split already used next door.

### How has this been tested?

```bash
npm run test:tooling
```

358 pass, 0 fail. 19 of those are new (`.github/scripts/__tests__/visual-approval-rerun.test.mjs`), covering: re-run on a failed gate, wait on an unfinished run, skip on a stripped label, skip on a closed pull request, skip when the head moved after labelling, skip on a gate that passed or was cancelled/timed out, and the job-name suffix match for the nested `publish.yml` path.

Several are drift locks rather than logic tests:

1. `COMPARE_JOB` is composed back out of `test.yml`'s `visual` job name and `visual.yml`'s `compare` job name. Renaming either would otherwise stop every approval silently, with a green run and nothing in the log.
2. The workflow's `timeout-minutes` must exceed the script's `WAIT_TIMEOUT_MINUTES`, so a job timeout cannot kill the script mid-wait and replace its message with a bare cancellation.
3. The concurrency group must contain `github.event.label.name`.
4. The workflow must pass `LABELED_HEAD_SHA`, and must carry both Dependabot clauses.
5. `a run's own head_sha cannot stand in for the labelled commit` pins the dead-guard bug directly: it feeds the state the wrapper actually produces (run agrees with the live head) and asserts the decision still skips.

Verified against the live GitHub API, read-only:

- Ran the script in `DRY_RUN` mode against #13325 and against this pull request. It read the pull request, the live labels, state and head SHA, found the newest `test.yml` run, paginated the job list, matched `Visual / Compare`, and decided correctly each time — including the closed-pull-request skip, which fired for real once #13325 merged.
- Replayed #13325's **real** attempt-2 job payload (`Visual / Compare` = `failure`, `CI Gate` = `failure`) through `decide()`. It returns `rerun` — so this would have fired on exactly the state that prompted the task.
- Confirmed the push guard is load-bearing rather than decorative: with `labeledHeadSha` absent the same input returns `rerun` (the old, buggy behaviour); with it, `skip`.
- Confirmed `run.actor` survives a re-run (run where `actor` is `handsontable-release-bot[bot]` and `triggering_actor` is the person who pressed re-run). That is what makes `visual.yml`'s `IS_UNTRUSTED` still true on a re-run of a Dependabot pull request — so the extra author clause prevents a *wasted* re-run, not an unsafe approval.
- Confirmed `head_branch` on this pull request's own run equals `pull_request.head.ref`, so the run lookup resolves.

**Not exercised: the `rerun-failed-jobs` POST itself.** Triggering it means spending a real build on an open pull request, so it is left for the first live use. The next pull request with a red `Visual / Compare`: add the label and watch the **Visual approval rerun** run. Everything up to that call is verified above.

Also ran `npx eslint` on the changed `visual-tests/lib/**` — clean. `.github/scripts/**` is outside ESLint's default scope, the same as the existing scripts there.

### Types of changes

### Related issue(s):

1. DEV-2749
2. Prompted by the label doing nothing on #13325.

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

CI and tooling only, plus the `handsontable-visual-tests` package. No published package changes.

### Checklist:

Documentation is updated in the same commits: `visual-tests/README.md` (the approval steps, the fork caveat, the mermaid flow and the manual re-run section), `visual-tests/AGENTS.md` (the approval bullet plus a new one on the `labeled`/`github.actor` trap), and the gate's own pull request comment in `visual-tests/lib/visual-gate.mjs`, which no longer tells people to re-run the job — and which now branches on `seeded`, so a fork or Dependabot contributor is told the label is ignored there instead of being promised a re-run that never starts.

### Follow-up (not in this PR)

`manual-qa.yml` has the identical shape: "a reviewer comments `/manual-qa passed`, **then re-run this job**". The same mechanism would close it. Left out to keep this change scoped.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2749

---

> [!NOTE]
> **Medium Risk**
> Changes merge-gate behavior and uses `actions: write` to re-run workflows; logic is heavily tested but the live `rerun-failed-jobs` path is not exercised in CI.
> 
> **Overview**
> Adds **`visual-approval-rerun`** automation so applying **`visual-approved`** re-triggers failed CI jobs instead of requiring a manual **Re-run jobs**. A new workflow on `pull_request: labeled` runs a polling script that waits for the Tests pipeline to finish, then calls **`rerun-failed-jobs`** when **`Visual / Compare`** failed—while re-reading labels, PR state, and head SHA on each poll so pushes or label removal cancel the re-run. Safety checks include **`LABELED_HEAD_SHA`** (approval must match the commit that was labeled), fork/Dependabot guards (with an extra author check for `labeled` events), and concurrency keyed by label name.
> 
> **Fixes a latent `visual.yml` fallback bug** by running **`build:types`** and **`downlevel:types`** when ES/CJS artifacts are missing, plus a new **`partial-packaging.test.mjs`** guard so any job using **`postbuild:partial`** must emit type declarations.
> 
> Contributor-facing copy in **`visual-gate.mjs`**, **`visual-tests/README.md`**, and **`AGENTS.md`** now describes label-only approval on trusted runs and omits misleading label instructions on fork/Dependabot runs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bebbff8548478b4268eb3f15a52a0e1945b5e970. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2749 Re-run the visual gate automatically when visual-approved is added](https://app.clickup.com/t/9015210959/DEV-2749)

.github/scripts/visual-approval-rerun.mjs
```diff
@@ -0,0 +1,266 @@
...
+  //
+  // It does not separate one branch raised against two bases (a backport opened
+  // against `develop` and a release branch): those runs agree on every field the
+  // run object exposes. `run.pull_requests` would settle it, but GitHub leaves
+  // it EMPTY on same-repo `pull_request` runs — verified against run
+  // 33610031096, which returns `pull_requests: []` — so filtering on it is a
+  // no-op dressed up as a check. In that residual case the worst outcome is a
+  // re-run spent on the sibling pull request's build; no approval crosses over,
+  // because each run's gate reads its own pull request's labels.
+  return runs.find(run => run.head_branch === headRef) ?? null;
```
demtario: Medium: the file header claims "all branching lives in `./lib/...`, which is pure and unit-tested; this wrapper only talks to the API and sleeps," but `readRun()`'s `runs.find(run => run.head_branch === headRef)` is real branching logic living here, untested — and it's the exact predicate this PR's own last commit had to fix (multi-PR-same-commit disambiguation). Suggest extracting a pure `selectRun(runs, headRef)` into `lib/visual-approval-rerun.mjs` with a unit test for the multi-PR-same-commit case already documented in the comment above.

.github/scripts/visual-approval-rerun.mjs
```diff
@@ -0,0 +1,266 @@
...
+  throw lastError;
+}
+
+/**
+ * Read the pull request's live head commit and labels.
+ *
+ * The event payload carries both, but it is a snapshot of the moment the label
+ * was applied — the very staleness this script exists to work around.
+ *
+ * @returns {Promise<{headSha: string, labels: string[]}>} The live state.
```
demtario: Low: `readPullRequest()`'s `@returns` JSDoc is stale — documents `{headSha, labels}` but the function returns (and the caller uses) `{headSha, headRef, state, labels}`. Update the type.

demtario (commented):
2 inline notes below.

.github/scripts/visual-approval-rerun.mjs
budnix: Nice catch — pulled it out as `selectRun()` in the lib in bebbff8548, with a unit test for the two-PRs-same-commit case.

.github/scripts/visual-approval-rerun.mjs
budnix: Good spot — updated in bebbff8548.

.github/scripts/visual-approval-rerun.mjs
demtario: Confirmed fixed in bebbff8548.

.github/scripts/visual-approval-rerun.mjs
demtario: Confirmed fixed in bebbff8548.

demtario (approved):

PR #13339, opened 2026-09-02
DEV-2728: Replace TypeDoc [[X]] doc links with resolvable {@link} tags

demtario:
### Context

The PR fixes API reference pages that publish `[[methodName]]` as literal text instead of a link.

`[[Target]]` is TypeDoc's legacy cross-reference syntax. The API reference is not built with TypeDoc — it is built with `jsdoc-to-markdown` and `dmd` (`docs/scripts/jsdoc-convert/`), and neither one has ever handled `[[Target]]`:

- `dmd@6.2.3` has no handler for it.
- `docs/scripts/jsdoc-convert/renderer/postProcessors/jsdocLinksFixer.mjs` only rewrites real markdown links, `[label](target)`.
- There is no remark or Astro plugin for it in `docs/src/plugins/` or `docs/astro.config.mjs`.
- `git log -S` over `docs/scripts/` shows support never existed.

So the shorthand is passed straight through to markdown and rendered as plain text. It is not failing on "some entries" — it fails on every one of them, and the leak is only visible where the JSDoc block actually reaches the generator. Ten published pages were affected on `latest`, and nine of them on `/docs/18.0/` as well:

`api/core`, `api/hooks`, `api/options`, `api/data-provider`, `api/export-file`, `api/empty-data-state`, `api/manual-column-move`, `api/manual-row-move`, `api/multi-column-sorting`, `api/trim-rows`

Plus one guide page, `server-side-data-crud.md`, which had the same shorthand in prose.

The working syntax is an inline `link` tag, which `jsdocLinksFixer.mjs` rewrites into an `@/api/<page>.md#<anchor>` link. This PR converts every reference to it, in four cases:

- **Already qualified** (`[[Options#dataProvider]]`, `[[ExportFile#exportAsBlobAsync]]`, `[[ContextMenu]]`) — converted verbatim.
- **Bare Core members** (`[[getSettings]]`, `[[getCellMeta]]`, `[[setCellMeta]]`, `[[getColumnMeta]]`) — qualified as `Core#...`, matching the 23 existing precedents on that page.
- **Targets the API reference does not document** (`DataProviderQueryParameters`, `handleBeforeAlterForCrud`, `#previousConditionStack`, `getParsedNumber`, …) — converted to inline code, **not** to a link tag. An unresolvable link tag silently emits an `@/api/<name>.md` link to a page that does not exist, which is worse than the raw text because nothing flags it.
- **No anchor on the target page** — `[[Hooks#run]]` in `core/hooks/constants.ts` was rewritten as prose. `Hooks#run` lives in `core/hooks/index.ts`, which `docs/scripts/jsdoc-convert/parser/parser.mjs` never reads, so the hooks page has no `#run` anchor.

Fixing this in the pipeline instead was rejected: `@example` blocks contain over 200 legitimate array literals (`[[1, 2], [3, 4]]`, `[['A1']]`, `[[null]]`), and a `[[X]]` → link-tag transform would mangle them.

#### Why the syntax spread

`handsontable/.ai/CONVENTIONS.md` documented `[[MY_LINK]]` as the rule — under a heading reading "Typedoc Formatting Rules", in a repository that has never used TypeDoc. The JSDoc template two paragraphs above it taught the same thing by example (`@fires [[eventName]]`, `@throws [[ErrorType]]`), and `.claude/skills/handsontable-code-review/references/code-quality.md` repeated it, so a reviewer following the skill would flag correct code.

All three are corrected here. Without that, a documented convention would contradict the new gate. The template's `@fires` and `@throws` examples are fixed to the form the source actually uses — bare `@fires Hooks#afterChange` and `@throws {Error}` — because the `**Emits**:` and `**Returns**:` lines those produce are auto-linked by `typesLinkingFixers.mjs` and need no link tag. A bullet in `handsontable/AGENTS.md` **Critical Rules** states the rule where it auto-loads, since `CONVENTIONS.md` needs an explicit read and the gate is Jest rather than ESLint.

#### Guard test

`handsontable/test/__tests__/jsdocLinkSyntax.unit.js` scans core JSDoc comment lines and `docs/content` pages for identifier-shaped `[[Target]]` references, so the syntax cannot come back. It walks the filesystem rather than shelling out to `git grep`, so it sees a brand-new unstaged file and does not depend on running inside a git checkout.

`docs/content/api` is skipped as generated output, except the three files its `.gitignore` un-ignores (`introduction.md`, `plugins.md`, `sidebar.js`) — those are hand-authored, tracked, and the likeliest place for a hand-written API cross-reference. `[[toc]]` directives and array literals are exempt. The scan covers `handsontable/src` and `docs/content`, the two surfaces where the shorthand was found; `[[X]]` written anywhere else is not caught. `//` line comments are deliberately out of scope: jsdoc parses `/**` blocks only, so a reference there cannot reach a page, and including them would flag commented-out code such as `// data: [[ISO_DATE]]`.

#### Unrelated fix taken while in the file

`core.ts` carried a byte-identical duplicated paragraph in the `updateSettings` JSDoc, which published the same three sentences twice on `api/core` (verified live). Pre-existing, but both copies were changed lines in this PR, so it is removed here rather than left half-rewritten.

### How has this been tested?

Regenerated the whole API reference and verified the output directly:

```bash

# No identifier-shaped [[Target]] survives (only [[toc]] and array literals):
grep -rnE '\[\[[A-Za-z_#]' docs/content/api/ | grep -v '\[\[toc\]\]'
```

All 107 generated pages were then checked for links pointing at a page or anchor that does not exist. Every one of the 27 targets introduced by this PR resolves. The 12 broken targets the checker reports (`@/api/layoutManager.md`, `hooks.md#run`, `hooks.md#add`, `options.md#pastemode`, …) are pre-existing, come from `@see` tags this PR does not touch, and are unchanged by it.

Spot-checked the rendered result on the page the report named — `docs/content/api/core.md` now contains `[Core#getCellMeta](@/api/core.md#getcellmeta)`, matching the label style of the `[Core#selectAll]` links already on the hooks page. The `updateSettings` duplication went from 2 occurrences to 1 after the rebuild.

New guard test:

```bash
npm run test:unit --prefix handsontable -- --testPathPattern=jsdocLinkSyntax
```

Every behavior it claims was confirmed by making it fail, not by reading it:

| Injected | Expected | Result |
|---|---|---|
| `[[SomeTarget]]` in a new `.ts` under `handsontable/src` | red | red, file named |
| `[[Hooks#afterChange]]` in a new page under `docs/content/guides` | red | red, file named |
| `/** Single-line block with [[SomeTarget]]. */` | red | red, file named |
| `[[SomeApiMethod]]` in tracked `docs/content/api/introduction.md` | red | red, file named |
| `[[SomeGeneratedTarget]]` in generated `docs/content/api/core.md` | green (excluded) | green |

Each fixture was removed afterwards; `git status` is clean.

Full suites:

```bash
npm run test:unit --prefix handsontable   # 4313 passed, 341 suites
npm run stylelint --prefix handsontable   # clean
```

No E2E run: the change is comments, one markdown paragraph, and a test. No runtime code is touched.

### Types of changes

### Related issue(s):

1. DEV-2728

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

[skip changelog] — documentation only. The change is JSDoc comments, agent-facing convention files, one guide paragraph, and a test; no runtime behavior changes. The changelog gate fires because the paths are under `handsontable/src/**`.

Once this lands, it needs cherry-picking to `prod-docs/18.1` and `prod-docs/18.0`, which is where those two branches' published pages are generated from. `core.ts` conflicts on both (10 references on 18.1 and 2 on 18.0, against 12 here), and `helpers/number.ts` conflicts on 18.0 (2 of 3). Running the guard test on each branch after resolving proves nothing was left behind.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2728

---

> [!NOTE]
> **Low Risk**
> Changes are limited to comments, documentation, and a static analysis test; no runtime or public API behavior is modified.
> 
> **Overview**
> Fixes **DEV-2728** by replacing TypeDoc-style `[[Target]]` cross-references in JSDoc and guides with syntax the API docs pipeline actually resolves: qualified `{@link Class#member}` tags, VuePress-style `[label](@/api/...)` in markdown, or inline code when there is no API page.
> 
> **Convention and agent docs** (`CONVENTIONS.md`, `AGENTS.md`, code-review skill) now state that references are built with `jsdoc-to-markdown` + `dmd`, not TypeDoc, and document `@fires` / `@throws` as bare targets.
> 
> **Source comments** across `core.ts`, hooks, DataProvider-related plugins, and helpers are updated accordingly; duplicate `updateSettings` JSDoc in `core.ts` is removed. One guide paragraph (`server-side-data-crud.md`) switches from `[[Hooks#...]]` to a proper API link.
> 
> **Regression guard:** new unit test `jsdocLinkSyntax.unit.js` scans `handsontable/src` JSDoc comment lines and `docs/content` markdown for identifier-shaped `[[Target]]`, while exempting `[[toc]]`, array literals, and generated API pages.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7da671d809b1520a1e10029cad32e850e5ebf63a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2728 API Docs section: URLs corrupted](https://app.clickup.com/t/9015210959/DEV-2728)

handsontable/test/__tests__/jsdocLinkSyntax.unit.js
```diff
@@ -0,0 +1,139 @@
...
+      }
+
+      matches.push(`${file}:${index + 1}: ${line.trim()}`);
+    });
+  }
+
+  return matches;
+}
+
+describe('cross-reference link syntax', () => {
```
budnix, line 130: This guard stops the old `[[Target]]` syntax coming back, but nothing checks that a new `{@link Class#member}` actually points at a page that exists. The PR description already names 12 targets on develop that go nowhere, like `@/api/layoutManager.md` and `hooks.md#run`, and the 27 added here were verified by hand once. Curious whether a second `it()` that fails when a linked class has no generated API page is worth a follow-up ticket, since those 12 existing hits would have to be fixed or listed first.

handsontable/test/__tests__/jsdocLinkSyntax.unit.js
```diff
@@ -0,0 +1,139 @@
...
+  join('handsontable', 'src', '3rdparty', 'walkontable', 'dist'),
+];
+
+// `docs/content/api` is regenerated from the core source on every docs build, so scanning it
+// would only re-report what the core-source case already covers. Its `.gitignore` un-ignores
+// three files, though (`/content/api/*` followed by `!introduction.md`, `!plugins.md`,
+// `!sidebar.js`): those are hand-authored and tracked, they are the likeliest place for a
+// hand-written API cross-reference, and they must stay in the scan.
+const GENERATED_API_DIR = join('docs', 'content', 'api');
+const AUTHORED_API_FILES = ['introduction.md', 'plugins.md', 'sidebar.js'];
```
budnix: `sidebar.js` never reaches the scan. Line 134 asks `walk` for `.md` files only, so the file is dropped by extension before this list is ever checked, while the comment above says all three files must stay in the scan. Dropping the entry looks like the better fix than adding `.js` to the docs walk: that walk runs with `commentsOnly` false, so a `.js` file would be read as raw code, and `sidebar.js` is a data module that renders no prose anyway.

handsontable/test/__tests__/jsdocLinkSyntax.unit.js
```diff
@@ -0,0 +1,139 @@
...
+
+const TYPEDOC_LINK = /\[\[[A-Za-z_#][A-Za-z0-9_#+.]*\]\]/;
```
budnix: The character class has no `|`, so TypeDoc's labelled form `[[Target|label]]` slips past this. Nothing in the repo ever used that form, so this is hardening rather than a real hole, but it is a one-character fix while you are in here.
```suggestion
const TYPEDOC_LINK = /\[\[[A-Za-z_#][A-Za-z0-9_#+.|]*\]\]/;
```

handsontable/test/__tests__/jsdocLinkSyntax.unit.js
demtario: Correct, fixed in 7da671d809. Line 134 asks `walk` for `.md` only, so `sidebar.js` was dropped by extension before `AUTHORED_API_FILES` was ever consulted, and the comment claiming all three stay in the scan was wrong.

Took your fix rather than adding `.js` to the docs walk, for the reason you gave: that walk runs with `commentsOnly` false, so a `.js` file would be read as raw code. The comment now says why `sidebar.js` is deliberately absent, so the next reader does not re-add it.

handsontable/test/__tests__/jsdocLinkSyntax.unit.js
demtario: Good catch on the gap, but the suggested class does not close it. The label characters have to satisfy the same class, so it only matches a single-token label:

| Input | `[A-Za-z0-9_#+.|]*` |
|---|---|
| `[[Core#getCellMeta\|getter]]` | matches |
| `[[Core#getCellMeta\|the getter]]` | misses |
| `[[Core#getCellMeta\|getCellMeta()]]` | misses |

A real label holds spaces and punctuation, so the realistic case would still have leaked. Fixed in 7da671d809 with an optional label group that runs to the closing brackets instead:

```js
const TYPEDOC_LINK = /\[\[[A-Za-z_#][A-Za-z0-9_#+.]*(?:\|[^\]\n]*)?\]\]/;
```

The target stays identifier-shaped, which is what keeps `[[1, 2], [3, 4]]` and `[['A1']]` out, and `[^\]\n]` cannot cross a `]`, so it will not run away across a Markdown table row. Checked against 17 cases (labelled and plain forms that must match, array-literal shapes that must not), then confirmed end to end: a `[[Core#getCellMeta|the effective getter]]` fixture in a new `src` file turns the suite red and is named in the output.

handsontable/test/__tests__/jsdocLinkSyntax.unit.js
demtario: Agreed, and agreed it is a follow-up rather than something to land here. Your own reason is the blocker: the 12 dead targets come from `@see` tags this PR does not touch, so the check would have to ship red, and allowlisting them at the same time as introducing the check makes the allowlist the default state.

Worth knowing that the checker already exists as a throwaway script, so the follow-up is mostly promotion plus that cleanup. It reads the generated `docs/content/api/*.md`, collects each page's heading and `{#id}` anchors, then flags every `(@/api/<page>.md#<anchor>)` whose page or anchor is absent. That is how the 27 targets here were verified, and how the 12 pre-existing ones were found:

- Missing pages: `@/api/layoutManager.md` (from `core.md`), `@/api/cELLS_LIMIT.md` (from `moveCells.md`).
- Missing anchors: `hooks.md#add`, `#has`, `#once`, `#remove`, `#run` (all from `core.md`); `options.md#pastemode`, `#renderallcolumn`; `autoColumnSize.md#widths`; `autoRowSize.md#heights`; `conditionCollection.md#event`.

The `Hooks#*` five are the same root cause as the `[[Hooks#run]]` reference dropped in this PR: `parser.mjs` feeds only `core/hooks/constants.js` to the Hooks page, so the `Hooks` class methods have no anchors and every `@see Hooks#add` points nowhere.

One design note for whoever picks it up: it has to run after `docs:api`, so it belongs in the docs build rather than in `test:unit`, which does not regenerate the reference. Happy to file the ticket with the script and this list attached if you want it tracked.

budnix (approved):

Issue #5027, opened 2018-04-12
Dropdown option for just horizontal alignment?


aaronbeall:
### Description
I'm using `dropdownMenu` with the option `"alignment"` and it looks like this:

![image](https://user-images.githubusercontent.com/2973900/38702727-56cc7188-3e6f-11e8-94d5-83c268b4a556.png)

However I don't want to show the "Top, Middle, Bottom" vertical alignment options as they don't really do anything in my case. Is there a way to hide this?

### Demo

http://jsfiddle.net/qxdtq6jj/

### Your environment
* Handsontable version: pro-1.18.1
* Browser Name and version: Chrome 65
* Operating System: MacOS High Sierra

AMBudnik:
Hi @aaronbeall 

That is a good question!

In the https://docs.handsontable.com/pro/2.0.0/bower_components/handsontable-pro/dist/handsontable.full.js we can see in one of the comments that


    /**
     * Execute context menu command.
     *
     * You can execute all predefined commands:
     *  * `'row_above'` - Insert row above
     *  * `'row_below'` - Insert row below
     *  * `'col_left'` - Insert column left
     *  * `'col_right'` - Insert column right
     *  * `'clear_column'` - Clear selected column
     *  * `'remove_row'` - Remove row
     *  * `'remove_col'` - Remove column
     *  * `'undo'` - Undo last action
     *  * `'redo'` - Redo last action
     *  * `'make_read_only'` - Make cell read only
     *  * `'alignment:left'` - Alignment to the left
     *  * `'alignment:top'` - Alignment to the top
     *  * `'alignment:right'` - Alignment to the right
     *  * `'alignment:bottom'` - Alignment to the bottom
     *  * `'alignment:middle'` - Alignment to the middle
     *  * `'alignment:center'` - Alignment to the center (justify)
     *
     * Or you can execute command registered in settings where `key` is your command name.
     *
     * @param {String} commandName
     * @param {*} params
     */

But it doesn't change the string to a key for context menu nor dropdown menu.

You may can with a [custom dropdown menu configuration.](http://jsfiddle.net/handsoncode/bw4qvxp5/) 

I'm marking this issue as `assistance needed` as apparently there's something missing in the examples or in the docs.

AMBudnik:
It looks like this PR https://github.com/handsontable/handsontable/pull/4998 can help us to solve this case. It's not merged yet but if you have some time I highly encourage to test it.

swistach:
[...]

PR #13340, opened 2026-09-02
DEV-2758: Skip unresolvable menu item keys instead of rendering them

budnix:
### Context

The PR fixes a `contextMenu` / `dropdownMenu` item key that cannot be resolved being rendered as a menu row labelled with the **raw key string**, on a row that does nothing when clicked.

The sympathetic case needs no unusual config at all:

```js
contextMenu: ['row_above', 'borders'] // customBorders not enabled
```

That renders a row reading `borders`. The user took a documented key straight from the [context menu keys table](https://handsontable.com/docs/javascript-data-grid/context-menu/#context-menu-with-selected-options) and did not enable its plugin — and that table itself says *"Requires: CustomBorders"*, so the docs fully expect people to land here. Any typo does the same (`row_abvoe` renders a row reading `row_abvoe`), and so does a `parent:child` command name used as an item key (`alignment:left`).

This is the failure mode already documented in `handsontable/src/plugins/contextMenu/AGENTS.md` as the shape of #5429, where `freeze_column` rendered a dead row in `dropdownMenu` for seven years. #5429 was closed by fixing that one plugin; the placeholder that made it invisible stayed. Eight plugins still register on the context menu hook only, so their keys are all exposed the same way.

An unresolvable key is now **skipped**, with a `warnOnce()` naming it — quiet in the UI, loud in the console, instead of the other way round.

Two smaller defects in the same area go with it:

- `CommandExecutor#execute()` split the command name on `:` and looked up only the first segment, so a command registered under a full `a:b` key — which is what object-form `items` produce — was never found and threw `Menu command 'alignment' not exists.` on click. It now matches the whole name first, then falls back to the split.
- A subcommand name matching no submenu entry read `disabled` off `undefined` and threw a `TypeError`. It is now a no-op.

#### Scoping constraint worth knowing before review

The skip **must not** live inside `getItems()`. `prepareMenuItems()` calls it **twice** — once to build the list handed to `afterContextMenuDefaultOptions`, then `setPredefinedItems()`, then again. On the first pass every plugin key is unknown *by design*, and the placeholder emitted there is load-bearing: `nestedRows/ui/contextMenu.ts` runs `rangeEach(0, items.length - 1, …)` and inserts its entries only when the list is **non-empty**. Filtering unconditionally makes `contextMenu: ['add_child']` yield an empty first-pass list, nestedRows never inserts, and `add_child` vanishes — re-breaking #9894. The `#predefinedItemsSet` flag confines the skip to the second pass, and `itemsFactory.unit.js` pins both halves.

An array entry can also be a full item definition **object** rather than a key string, so the skip is guarded by `!isObject(name)`.

#### What this deliberately does not do

It does not resolve predefined subcommand keys at any menu level, which is the open request in [#5027](https://github.com/handsontable/handsontable/issues/5027). `alignment:left` is documented only as an `executeCommand()` **command name** — and it works there — while the documented list of *menu item keys* contains only top-level keys, so that expectation was an assumption rather than a broken promise. Building it would mean changing the shallow `extend` merge that #9894 depends on. Instead the docs now state the distinction and give the supported way to trim a submenu, which is what #5027 actually asked for:

```js
contextMenu: true,
afterContextMenuDefaultOptions(options) {
  const alignment = options.items.find(item => item.key === 'alignment');
  const verticalKeys = ['alignment:top', 'alignment:middle', 'alignment:bottom'];

  alignment.submenu.items = alignment.submenu.items
    .filter(item => !verticalKeys.includes(item.key));
}
```

`{ items: { 'alignment:left': { name: 'Left' } } }` therefore still does nothing on click — it just no longer throws.

### How has this been tested?

New unit coverage for both halves of the fix. Each new test was run against the **unfixed** source first: 3 of 10 `itemsFactory` tests and 7 of 10 `commandExecutor` tests fail there, and the ones that pass on both are the pre-existing behaviors they pin.

```bash
npm --prefix handsontable run test:unit -- --testPathPattern=itemsFactory      # 10 passed
npm --prefix handsontable run test:unit -- --testPathPattern=commandExecutor   # 10 passed
```

The Playwright spec for #5429 contained a test that **pinned the old placeholder behavior** (`expect(items).toEqual(['freeze_column', 'unfreeze_column'])`, commented as holding on the unfixed build by design). That block now pins the skip instead, plus the console warning:

```bash
cd tests && HOT_TEST_PORT=8213 npx playwright test --project=e2e-main e2e/dropdown-menu-freeze-column.spec.ts   # 21 passed
```

Regression runs over everything that contributes menu items, `nestedRows` first as the #9894 risk:

```bash
npm --prefix handsontable run test:e2e -- --testPathPattern=nestedRows      # 175 specs, 0 failures
npm --prefix handsontable run test:e2e -- --testPathPattern=contextMenu     # 596 specs, 0 failures
npm --prefix handsontable run test:e2e -- --testPathPattern=dropdownMenu    # 196 specs, 0 failures
npm --prefix handsontable run test:e2e -- --testPathPattern=filters         # 284 specs, 0 failures
npm --prefix handsontable run test:e2e -- --testPathPattern=customBorders   # 141 specs, 0 failures
npm --prefix handsontable run test:unit                                     # 4331 passed, 342 suites
npm --prefix handsontable run eslint                                        # 0 errors
npm --prefix handsontable run test:types                                    # clean
npm --prefix handsontable run build                                         # clean, incl. strict postbuild
```

Note for anyone re-running the Playwright leg: pass `HOT_TEST_PORT`. On the default 8123 `reuseExistingServer` attaches to whatever already listens, and another checkout's server silently serves its own build — that produced two confusing red runs here before the port was pinned.

### Types of changes

Behavior does change for a grid configured with a key that resolves to nothing: the row disappears instead of showing the raw key. Nothing can depend on clicking it — it carried no callback — so this is treated as a bug fix rather than a breaking change. The only conceivable reliance is abusing an arbitrary string as a static label, which is undocumented and better served by `{ name: '…', disabled: true }`.

### Related issue(s):

1. DEV-2758
2. Same defect class as #5429, which was closed by fixing one plugin rather than the placeholder.
3. Addresses the symptom in #5027 and documents its answer; the feature request in that thread is deliberately not built (see above).

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2758

---

> [!NOTE]
> **Medium Risk**
> Changes shared context/dropdown menu item building and command execution paths used on every menu open; behavior is intentionally different when keys are invalid, but regression coverage is strong around the two-pass hook contract and nestedRows.
> 
> **Overview**
> **Fixes dead menu rows** that showed the raw configuration key (e.g. `borders` without CustomBorders, or `freeze_column` in the column menu when the plugin does not contribute). After plugin hooks run, `ItemsFactory` now **drops** keys that still do not resolve and logs a **one-time console warning** instead of emitting a no-op placeholder. The skip is limited to the second `getItems()` pass via `#predefinedItemsSet`, so the first pass still keeps placeholders plugins like nestedRows need (#9894).
> 
> **`CommandExecutor`** gains `#findCommand()`: parent `parent:child` submenu lookup stays first; whole-name lookup fixes object-form keys like `'alignment:left'` (#5027). Misspelled subcommands warn and no-op; lookups use `hasOwnProperty` to avoid `Object.prototype` names. **DropdownMenu** treats colon-keyed commands as known when deciding whether to rebuild before `executeCommand`.
> 
> Docs add guidance on invalid keys, plugin requirements, and that `alignment:left` is a command name—not a menu key—with `afterContextMenuDefaultOptions` examples for trimming submenus. Unit and e2e tests cover both passes, command precedence, and the updated skip behavior (DEV-2758 / #5429 class).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fec8684727cbb5ac94f0f82ad23806527b1513ee. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/content/guides/accessories-and-menus/context-menu/context-menu.md
+++ b/docs/content/guides/accessories-and-menus/context-menu/context-menu.md
@@ -125,6 +125,69 @@ The `filter_by_condition`, `filter_by_condition2`, `filter_operators`, `filter_b
 `filter_action_bar` items build the filtering interface and take effect only in the dropdown (column)
 menu, not in the context menu. See [Filter menu items](@/guides/accessories-and-menus/column-menu/column-menu.md#filter-menu-items).
 
+The table above lists every key you can use. A key that is not on it is skipped, and Handsontable
+logs a console warning naming it. A key is also skipped when the plugin it needs is not enabled, so
+check the "Requires" column when an item does not appear.
+
+### Command names are not menu item keys
+
+Some commands are addressed with a `parent:child` name, such as `alignment:left`. Those names work
+with [`executeCommand()`](@/api/contextMenu.md#executecommand), which runs a command directly:
+
+::: only-for javascript
+```js
+hot.getPlugin('contextMenu').executeCommand('alignment:left');
+```
+:::
+
+::: only-for react
+```jsx
+hotRef.current.hotInstance.getPlugin('contextMenu').executeCommand('alignment:left');
+```
+:::
+
+They are not menu item keys. Listing `alignment:left` in `contextMenu` does not create a "Left"
+item, because only the keys in the table above are resolved.
+
+To show part of a submenu, keep the predefined submenu items and filter out the ones you don't
+want. The items you keep carry their own actions, so they keep working:
+
+::: only-for javascript
+```js
+const hot = new Handsontable(container, {
+  contextMenu: true,
+  afterContextMenuDefaultOptions(options) {
+    const alignment = options.items.find(item => item.key === 'alignment');
+    const verticalKeys = ['alignment:top', 'alignment:middle', 'alignment:bottom'];
+
+    alignment.submenu.items = alignment.submenu.items
+      .filter(item => !verticalKeys.includes(item.key));
+  },
+});
+```
+:::
+
+::: only-for react
+```jsx
+<HotTable
+  contextMenu={true}
+  afterContextMenuDefaultOptions={(options) => {
+    const alignment = options.items.find(item => item.key === 'alignment');
+    const verticalKeys = ['alignment:top', 'alignment:middle', 'alignment:bottom'];
+
+    alignment.submenu.items = alignment.submenu.items
+      .filter(item => !verticalKeys.includes(item.key));
+  }}
+/>
+```
+:::
+
+The alignment submenu then offers only the horizontal options. For the column menu, use
+`afterDropdownMenuDefaultOptions` instead.
+
+To build a submenu item of your own, give it a `name` and a `callback`. A custom item runs only the
+`callback` you write — it does not inherit an action from a predefined key of the same name.
+
 To see the context menu, right-click on a cell. On touch devices, long-press a cell to open the context menu.
 
 ::: only-for javascript
--- a/docs/content/guides/accessories-and-menus/context-menu/context-menu.md
+++ b/docs/content/guides/accessories-and-menus/context-menu/context-menu.md
@@ -126,67 +126,8 @@ The `filter_by_condition`, `filter_by_condition2`, `filter_operators`, `filter_b
 menu, not in the context menu. See [Filter menu items](@/guides/accessories-and-menus/column-menu/column-menu.md#filter-menu-items).
 
 The table above lists every key you can use. A key that is not on it is skipped, and Handsontable
-logs a console warning naming it. A key is also skipped when the plugin it needs is not enabled, so
-check the "Requires" column when an item does not appear.
-
-### Command names are not menu item keys
-
-Some commands are addressed with a `parent:child` name, such as `alignment:left`. Those names work
-with [`executeCommand()`](@/api/contextMenu.md#executecommand), which runs a command directly:
-
-::: only-for javascript
-```js
-hot.getPlugin('contextMenu').executeCommand('alignment:left');
-```
-:::
-
-::: only-for react
-```jsx
-hotRef.current.hotInstance.getPlugin('contextMenu').executeCommand('alignment:left');
-```
-:::
-
-They are not menu item keys. Listing `alignment:left` in `contextMenu` does not create a "Left"
-item, because only the keys in the table above are resolved.
-
-To show part of a submenu, keep the predefined submenu items and filter out the ones you don't
-want. The items you keep carry their own actions, so they keep working:
-
-::: only-for javascript
-```js
-const hot = new Handsontable(container, {
-  contextMenu: true,
-  afterContextMenuDefaultOptions(options) {
-    const alignment = options.items.find(item => item.key === 'alignment');
-    const verticalKeys = ['alignment:top', 'alignment:middle', 'alignment:bottom'];
-
-    alignment.submenu.items = alignment.submenu.items
-      .filter(item => !verticalKeys.includes(item.key));
-  },
-});
-```
-:::
-
-::: only-for react
-```jsx
-<HotTable
-  contextMenu={true}
-  afterContextMenuDefaultOptions={(options) => {
-    const alignment = options.items.find(item => item.key === 'alignment');
-    const verticalKeys = ['alignment:top', 'alignment:middle', 'alignment:bottom'];
-
-    alignment.submenu.items = alignment.submenu.items
-      .filter(item => !verticalKeys.includes(item.key));
-  }}
-/>
-```
-:::
-
-The alignment submenu then offers only the horizontal options. For the column menu, use
-`afterDropdownMenuDefaultOptions` instead.
-
-To build a submenu item of your own, give it a `name` and a `callback`. A custom item runs only the
-`callback` you write — it does not inherit an action from a predefined key of the same name.
+logs a console warning naming it. A key is also skipped when the plugin that provides it is not
+enabled, so check the required plugin named beside the key when an item does not appear.
 
 To see the context menu, right-click on a cell. On touch devices, long-press a cell to open the context menu.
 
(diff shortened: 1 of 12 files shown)


qunabu:
Task linked: [DEV-2758 Context menu: unresolvable item keys render as raw text instead of being skipped](https://app.clickup.com/t/9015210959/DEV-2758)

docs/content/guides/accessories-and-menus/context-menu/context-menu.md
```diff
@@ -170,6 +174,133 @@ To see the context menu, right-click on a cell. On touch devices, long-press a c
...
+
+::: only-for javascript
+```javascript
+const verticalKeys = ['alignment:top', 'alignment:middle', 'alignment:bottom'];
+
+const hot = new Handsontable(container, {
+  licenseKey: 'non-commercial-and-evaluation',
+  contextMenu: true,
+  afterContextMenuDefaultOptions(options) {
+    const alignment = options.items.find(item => item.key === 'alignment');
```
demtario, line 226: **high**: this "filter a submenu" example throws when copy-pasted, in all four framework variants (JS here, plus lines 238, 264, 285).

`afterContextMenuDefaultOptions`/`afterDropdownMenuDefaultOptions` call the hook with the items array itself as the first argument (see `contextMenu.ts:413`, `dropdownMenu.ts:283`, and the `GridSettings` typing in `core/settings.ts`), not an object with an `.items` property. This example does `options.items.find(...)`, which throws `Cannot read properties of undefined (reading 'find')`.

Fix: rename the parameter to `items` and call `items.find(...)` directly (drop `.items`) in all four variants.

docs/content/guides/accessories-and-menus/context-menu/context-menu.md
```diff
@@ -125,6 +125,10 @@ The `filter_by_condition`, `filter_by_condition2`, `filter_operators`, `filter_b
 `filter_action_bar` items build the filtering interface and take effect only in the dropdown (column)
 menu, not in the context menu. See [Filter menu items](@/guides/accessories-and-menus/column-menu/column-menu.md#filter-menu-items).
 
+The table above lists every key you can use. A key that is not on it is skipped, and Handsontable
```
demtario: **high**: this sentence overstates what's fixed. "A key that is not on it is skipped, and Handsontable logs a console warning naming it" isn't true for two cases:

- Object-form `items` (e.g. `contextMenu: { items: { myKey: 'not_a_real_key' } }`) still falls through to the old raw-placeholder bug with no warning at all — `getItems()`'s `isObject(pattern)` branch is unchanged by this PR.
- A key matching a hidden built-in `ITEMS` member (e.g. hidden via `allowInsertRow: false`) is dropped silently, no warning — per `itemsFactory.unit.js`.

Same overstatement in `handsontable/src/plugins/contextMenu/AGENTS.md:20` and in the changelog title.

Fix: scope this sentence to array-form keys outside the built-in `ITEMS` set, or extend `reportUnresolvedKey` to the object-form branch so the claim is universally true.

handsontable/src/plugins/dropdownMenu/dropdownMenu.ts
```diff
@@ -598,9 +598,16 @@ export class DropdownMenu extends BasePlugin {
...
     // a listener would see as noise. And never while the menu is open — it was just built, and a
     // rebuild would swap the items out from under the click that is running this command.
+    // A command can also be registered under a key that itself contains a colon, so both names
+    // count as known here — the same two lookups `CommandExecutor#execute()` makes. Testing only
+    // the primary name would rebuild the list on every colon-keyed command, which is exactly the
+    // per-call hook noise this check exists to avoid.
     const [primaryCommandName] = commandName.split(':');
+    const { commands } = this.commandExecutor;
+    const isCommandKnown = hasOwnProperty(commands, primaryCommandName) ||
+      hasOwnProperty(commands, commandName);
```
demtario: **high**: the new `hasOwnProperty(commands, commandName)` whole-name fallback (line 608) has no test coverage. Every e2e and legacy spec that calls `executeCommand`/`executeDropdownCommand` in this PR uses either an unqualified key or a submenu-parent-resolved key, both satisfied by the pre-existing `primaryCommandName` check on line 607 alone.

Fix: add a case with object-form `items: { 'alignment:left': {...} }`, call `executeCommand('alignment:left')` twice, and assert `afterDropdownMenuDefaultOptions` fires only once (from the initial build) — proving the fallback actually avoids the extra rebuild it claims to.

handsontable/src/plugins/dropdownMenu/dropdownMenu.ts
```diff
@@ -598,9 +598,16 @@ export class DropdownMenu extends BasePlugin {
     // Only an unknown one: rebuilding on every call would fire both item hooks per command, which
     // a listener would see as noise. And never while the menu is open — it was just built, and a
     // rebuild would swap the items out from under the click that is running this command.
+    // A command can also be registered under a key that itself contains a colon, so both names
+    // count as known here — the same two lookups `CommandExecutor#execute()` makes. Testing only
+    // the primary name would rebuild the list on every colon-keyed command, which is exactly the
+    // per-call hook noise this check exists to avoid.
     const [primaryCommandName] = commandName.split(':');
```
demtario: **high** (architecture): the "primary name, then whole name" resolution rule is now duplicated here and in `commandExecutor.ts#findCommand` (lines 105-146) — the comment above even says so ("the same two lookups `CommandExecutor#execute()` makes"). `AGENTS.md` flags the still-open half of issue #5027 (resolving predefined subcommand keys); when that lands, this copy of the rule silently goes stale since nothing keeps the two in sync.

Fix: expose a `hasCommand(name: string): boolean` (or similar) on `CommandExecutor` and have `DropdownMenu#executeCommand` call it instead of re-implementing the two-lookup rule.

handsontable/src/plugins/contextMenu/itemsFactory.ts
```diff
@@ -94,18 +103,44 @@ export class ItemsFactory {
...
+
+  /**
+   * Reports a menu item key that names nothing, once per grid instance per key.
+   *
+   * @param {string} name The unresolved menu item key.
+   */
+  #warnUnresolvedKey(name: string) {
+    warnOnce(
+      this.hot.rootElement,
+      `menu-unresolved-item-key:${name}`,
```
demtario: **medium**: `warnOnce()`'s dedup key is `menu-unresolved-item-key:${name}`, scoped to `hot.rootElement` — shared by both `contextMenu` and `dropdownMenu` on the same instance. A key that's unresolvable in both menus (e.g. `borders` with `CustomBorders` disabled, a scenario this PR's own `AGENTS.md` lists) warns exactly once, and the message names neither menu.

Fix: include the menu name (or plugin name) in both the dedup key (e.g. `menu-unresolved-item-key:${menuName}:${name}`) and the warning text, so the message is actionable and a second, independently useful warning from the other menu isn't silently suppressed.

handsontable/src/plugins/contextMenu/commandExecutor.ts
```diff
@@ -96,9 +91,60 @@ export class CommandExecutor {
...
-    params.unshift(commandSplit.join(':'));
+    params.unshift(commandName);
     arrayEach(callbacks, callback => callback.apply(this.hot, params));
   }
+
+  /**
+   * Resolves a command name to the descriptor that should run.
+   *
+   * @param {string} commandName Command id, optionally a `parent:child` subcommand name.
+   * @returns {object|undefined} The command, or `undefined` when a subcommand name matches no
```
demtario, line 117: **nit**: `#findCommand`'s JSDoc documents only the return type, not that it can throw. It calls `throwWithCause(...)` on an unresolvable parent (line 146) with no `@throws` tag. Low priority — the repo uses `@throws` exactly once codebase-wide, and the inline comments plus the `AGENTS.md` resolution ladder already document the throw path.

demtario (commented):
6 inline notes below.

- `.changelogs/13340.json` documents only the raw-key-placeholder fix (#5429), not the second, independently user-visible defect this PR also fixes: a colon-keyed command (`{ items: { 'alignment:left': ... } }`) used to throw `Menu command 'alignment' not exists.` on click and now works (#5027, covered by `commandExecutor.unit.js` and the new "Command names are not menu item keys" guide section). Consider a second bullet or an amended title so the changelog matches the shipped behavior change.

docs/content/guides/accessories-and-menus/context-menu/context-menu.md
budnix: Fair point on the type, but the hook is called with `{ items }` — see `contextMenu.ts:409`, and `nestedRows` and `manualColumnFreeze` both read `.items` — so the example runs as written. The declared array type in `core/settings.ts` is the real bug there and deserves its own fix.

docs/content/guides/accessories-and-menus/context-menu/context-menu.md
budnix: Good spot — scoped that sentence to array-form keys in c08b5a2e1a, and called out the two silent cases. Also amended the changelog title to cover the colon-key throw.

handsontable/src/plugins/dropdownMenu/dropdownMenu.ts
budnix: You're right, it was untested. Added a `colon-key` grid and a hook-call counter in c08b5a2e1a; the count goes 0 to 2 without the check.

handsontable/src/plugins/dropdownMenu/dropdownMenu.ts
budnix: Agreed — pulled the rule into `CommandExecutor#hasCommand()` in c08b5a2e1a, so the dropdown menu asks instead of re-implementing it.

handsontable/src/plugins/contextMenu/itemsFactory.ts
budnix: Nice catch — the menu name is now in both the dedup key and the text (c08b5a2e1a), passed in as the third argument to `ItemsFactory`.

handsontable/src/plugins/contextMenu/commandExecutor.ts
budnix: Good point — noted the throw in the JSDoc description in c08b5a2e1a, without adding a `@throws` tag.

demtario (approved):

PR #12255, opened 2026-04-02
fix(DEV-345): dataSchema defaults no longer block minSpareRows spare row detection

qunabu:
## Summary

Fixes [DEV-345](https://app.clickup.com/t/86c94yz5r) and GH #2409.

- `isEmptyRow` and `isEmptyCol` in `metaSchema.js` previously only treated `null`, `undefined`, and `''` as "empty". When `dataSchema` defined non-null defaults (e.g. `false` for a checkbox column), spare rows filled with schema defaults were never counted as empty.
- This caused `adjustRowsAndCols` to see 0 spare rows on every user action and keep appending rows indefinitely.
- Fix: when an explicit `dataSchema` is configured, a cell whose value equals the schema default for that column is now treated as "empty" — consistent with how object-typed values were already handled. Without an explicit `dataSchema` (duck-schema), behaviour is unchanged.

## Test plan

- [ ] `npm_config_testPathPattern=metaManager/__tests__/metaSchema pnpm --filter handsontable run test:unit` — 12 new unit tests all pass
- [ ] E2E regression in `test/e2e/Core_isEmpty.spec.js` covers `isEmptyRow`/`isEmptyCol` with `dataSchema` non-null defaults
- [ ] E2E regression in `test/e2e/settings/minSpareRows.spec.js` covers the checkbox + `minSpareRows` scenario (GH #2409)
- [ ] E2E regression in `test/e2e/settings/minSpareCols.spec.js` covers array `dataSchema` + `minSpareCols`

https://claude.ai/code/session_01KuVd6n7DAxZtiHyZWVeE8F

---

> [!NOTE]
> **Medium Risk**
> Touches core emptiness detection used by automatic spare row/col logic, which can subtly affect grid sizing behavior across configurations. The change is scoped to cases with an explicit `dataSchema` and is backed by new unit/E2E tests.
> 
> **Overview**
> Fixes a regression where `minSpareRows`/`minSpareCols` could keep appending rows/columns when `dataSchema` defined non-null defaults (e.g. `false`), by updating `metaSchema`’s `isEmptyRow`/`isEmptyCol` to treat *explicit* schema-default values (including primitives) as empty while preserving duck-schema behavior.
> 
> Adds targeted unit and E2E coverage for array/object schemas, object-valued defaults, and a loop short-circuit regression, plus a public changelog entry for GH #2409.
> 
[...]

Issue #2409, opened 2015-04-24
row creation bug when dataSchema is specified on checkbox column


steveAG:
When the dataSchema option is used and a checkbox column exists in the table it yields an interesting bug, every value change adds a new row to the table. If the dataSchema option is removed, or the checkbox column is removed everything functions as expected.

Here's a fiddle with a simple mock-up that replicates the issue: http://jsfiddle.net/96t06hvz/

AMBudnik:
Interesting case - let us take a closer look at it.

AMBudnik:
present in 0.24.1, demo: http://jsfiddle.net/96t06hvz/1/

madmax1025:
I was able to reproduce this issue when my default empty object had values for some properties.  I had set the checkbox columns underlying value to false in my object constructor.  When my default object had a null values for every property then no new rows were added with every change.

The datamap.createRow was triggered in the "adjustRowsAndCol" function in core.js.  The call to instance.countEmptyRows always returned a value of 0 when using a default object that had any property with a value other than null.

AMBudnik:
thanks for sharing investigation results @madmax1025

gentoo90:
Looks like all values in `dataSchema` have to be `null` to prevent this. `false` doesn't count as `null` so the row doesn't count as empty and after any change table will add new rows.

`0` also doesn't cont as `null`, but empty string `''` does.

AMBudnik:
Replicated using version 0.35.1/1.15.1

Updated demo http://jsfiddle.net/qyxeyjd7/
![gif](https://user-images.githubusercontent.com/10757813/36030188-7ac4b496-0da6-11e8-87c4-650a7c63f1e0.gif)

AMBudnik:
I have replicated the issue using v`2.0.0.` it happens to the `minSpareRows` being set. If we comment them the issue is also gone.

ps. I did not find any version where this case worked as expected.

warpech:
> Looks like all values in `dataSchema` have to be `null` to prevent this.

I can confirm that I have replicated this workaround in https://jsfiddle.net/warpech/ekayqv9t/5/

AMBudnik:
The issue remains the same in v 12.1.3 http://jsfiddle.net/yovr45m3/

qunabu:
This is fixed and will be release in 17.1.0

PR #13342, opened 2026-09-02
DEV-2760: Read dataSchema defaults along the column's dotted path (#5069)

### LINKED ISSUE #5069: default values in dataSchema break minSpareRows behavior
### Description

When setting default values via the dataSchema property, those values are set in new rows, which is great. However, when using the minSpareRow options, those rows are not recognized as empty. 
This leads to new rows being added after every single operation.

To check if a cell is empty should be evaluated against it's value in the schema, not just null.

This is probably the generalized case of #2409

### Steps to reproduce

1. set a dataSchema with default values
2. set minSpareRows > 1
3. change any cell => new rows are added

### Demo

https://jsfiddle.net/yhy5cmzo/67/

### Your environment
* Handsontable version: 0.36.0
* Browser Name and version: Chromium 64.0.3282.167 & Firefox 58.0.2
* Operating System: Ubuntu 17.10 (64-bit)

budnix:
### Context

The PR fixes the remaining half of GH issue [#5069](https://github.com/handsontable/handsontable/issues/5069) — open since 2018 — where `dataSchema` default values make `minSpareRows` append new rows after every single operation.

The **flat**-`dataSchema` half was already fixed by #12255 (DEV-345) and released in 17.1.0. That issue stayed open only because the changelog entry cited #12254 and #2409, never #5069 — nobody linked the fix back to the ticket.

A **nested** `dataSchema` default still reproduced on released 18.0.0 and on `develop`. Verified in Chrome against the 17.0.1, 18.0.0 and local builds: the grid grows 6 → 8 → 10 → 12 → 14 — two rows per edit, without bound.

```js
{
  data: [{ name: 'a', meta: { active: true } }, …],
  dataSchema: { name: null, meta: { active: false } },
  columns: [{ data: 'name' }, { data: 'meta.active', type: 'checkbox' }],
  minSpareRows: 2,
}
```

**Cause.** `isEmptyRow` / `isEmptyCol` decide "empty" by comparing each cell against its `dataSchema` default, but they looked that default up **flat** — `schema[prop]`. For a column bound to a nested property, `colToProp()` hands back a dot-separated path (`'meta.active'`), so `schema['meta.active']` resolved to `undefined`, never matched the leaf value the cell actually holds, and no spare row ever counted as empty. `countEmptyRows(true)` then returned `0`, and `adjustRowsAndCols` appended `minSpareRows` fresh rows on every change batch.

**Fix.** A module-level `getSchemaDefault(schema, prop, dataDotNotation, hasExplicitSchema)` in `metaSchema.ts` reads the default the same way the compared value is read, so the two are comparable.

The reader it mirrors is **`Core#getDataAtCell` -> `DataMap#get`**. That matters, and getting it wrong is what a review round caught here: `DataMap#get` tests `hasOwnProperty(row, prop)` **first** and only then walks the path, so an own key wins even a literal dotted one, and even while `dataDotNotation` is on. An earlier revision of this branch mirrored `DataSource#getAtPhysicalCell` instead — a different reader, without that rule — and walked unconditionally, which broke `dataSchema: { 'meta.active': false }`: the value side resolved the literal key to `false` while the default side read `schema.meta` and got `undefined`, reintroducing the exact unbounded growth this PR exists to stop. Both the own-key precedence and a regression test for it are now in place.

Four more details worth a reviewer's attention:

- The walk is **confined to an explicit `dataSchema`**. Without one, `getSchema()` returns the duck-schema, and resolving a dotted path inside it newly reported a real row whose nested object holds only `null` leaves as empty — changing `countEmptyRows()` for grids that never set `dataSchema` at all. That was also caught in review and is now gated, so duck-schema behavior is byte-identical to 18.0.0.
- It deliberately does **not** reuse `getProperty()`: that throws on a `null` intermediate (`null['x']`), and a schema may legitimately stop short of the depth a column asks for.
- The object-typed (`isObjectEqual`) branch needed the same path-aware read, or a column bound to a nested *object* (`data: 'meta.tags'`) keeps the original miss.
- `isEmpty()` already filters `null` / `''` / `undefined` before the comparison is reached, so an unresolvable path returning `undefined` can never accidentally match a cell value.

A function `columns[].data` accessor is read flat and resolves to `undefined`. The schema is a plain template with no accessor applied to it, so such a column is never treated as holding its default — unchanged from before this helper existed, and noted in the JSDoc and in `metaManager/AGENTS.md` so the gap is explicit rather than accidental.

`isObjectEqual(a, b)` is widened to `unknown` parameters. Its body was already `unknown`-safe (`stableStringify` handles objects, arrays, dates, primitives, `null` and `undefined`), while the old `object | unknown[]` signature forced an `as` cast at all 8 call sites. Widening it let this change add **no new casts** and drops two existing ones. `unknown` is a superset, so every existing caller still compiles — confirmed by `npm run test:types`. The six *other* call sites keep their now-redundant casts: stripping them is a mechanical cleanup across unrelated files, and I kept this diff to the bug. Happy to do it here if a reviewer would rather.

**Deliberately out of scope.** While verifying this, a second and arguably worse defect turned up: `DataMap#recursiveDuckColumns` registers a column property **only** for schema keys whose value is `null`. With a `dataSchema` carrying non-null defaults and no `columns` option, those columns get no property and render blank even though the data holds the values — a real row `{ id: 1, opt: 'o2', flag: true }` renders as `["1", "", ""]`. That is unchanged on 17.0.1 and 18.0.0, so it is not a regression, and fixing it would change which columns get a property and therefore shift column counts — breaking-change territory that needs sign-off. It is documented in `handsontable/src/dataMap/metaManager/AGENTS.md`, including the warning that it makes no-`columns` fixtures *look* like they pass a spare-row test, because those rows count as empty by accident.

### How has this been tested?

Every test here was confirmed to **fail without the code it covers**, by reverting the source and rebuilding — not written after the fact and assumed to bite.

- The original 8 cases: 3 unit failures and 4 Playwright failures against pre-fix code, with the exact expected numbers (5 rows expected, 7 received).
- The two regression cases added in review: both fail against the intermediate, buggy revision of this branch and pass against the current one.

The Playwright spec includes a deliberate complement case — filling a spare row **must** still append a fresh one — so the fix cannot pass by simply reporting every row as empty.

```bash
# Unit — 10 new cases
npm run test:unit --prefix handsontable -- --testPathPattern=metaSchema   # 30 passed

# New Playwright spec, per leg (the shape each CI job runs)
cd tests && npx playwright test --project=e2e-main e2e/min-spare-rows-nested-schema.spec.ts
# 5 passed; all six legs green individually

# Regression safety
npm run test:unit --prefix handsontable                                   # 4321 passed, 340 suites
npm run test:types --prefix handsontable                                  # passed
npm run build --prefix handsontable                                       # passed, incl. strict postbuild
npm run eslint --prefix handsontable                                      # clean

# The legacy specs #12255 added still pass
npm run test:e2e --prefix handsontable -- --testPathPattern="minSpare|Core_isEmpty"
# 35 specs, 0 failures
```

Manual verification in Chrome, with 17.0.1 as a positive control so a "does not reproduce" result could not pass vacuously:

| Scenario | 17.0.1 (pre-fix) | 18.0.0 | This branch |
|---|---|---|---|
| Flat object `dataSchema` + `columns` (as originally reported) | 7 → 9 → 11 → 13 → 15 | stable at 5 | stable at 5 |
| **Nested** `dataSchema`, `columns: [{ data: 'meta.active' }]` | 6 → 8 → 10 → 12 → 14 | 6 → 8 → 10 → 12 → 14 | **stable at 4** |
| Array `dataSchema` with non-null defaults | 7 → 9 → 11 → 13 → 15 | stable at 5 | stable at 5 |
| Literal dotted key `{ 'meta.active': false }` | stable at 4 | stable at 4 | stable at 4 |
| No `dataSchema`, `columns: [{ data: 'meta.tags' }]` | `isEmptyRow(1)` false | false | false |
| Control: nested schema with a `null` leaf default | stable | stable | stable |

The last two rows are the review regressions, measured against 18.0.0 to confirm this branch now matches it exactly.

Also measured on this branch: `isEmptyCol(1)` returns `true` when every value in a nested-default column equals its default (it returned `false` before), a column bound to the parent object (`data: 'meta'`) still compares by content, and `dataDotNotation: false` with a literal dotted key still resolves.

**One note on local test runs.** Row counts are read with `expect.poll`, because `adjustRowsAndCols` is deferred and a one-shot read cannot tell a stable count from one that has not grown yet. Separately, running all six legs of this spec concurrently on one local static server can exceed the 20s per-test timeout inside `page.waitForFunction` — five workers each pulling a freshly written ~6 MB bundle, the throughput trap `tests/AGENTS.md` documents. Each leg passes on its own, which is both the documented local gate (`--project=e2e-main`) and the shape CI runs, since CI gives every leg its own job. Flagging it rather than leaving a reviewer to hit it.

### Types of changes

### Related issue(s):

1. DEV-2760
2. Fixes #5069 (the nested-`dataSchema` half; the flat half shipped in 17.1.0 via #12255)

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

The changelog entry is `.changelogs/5069.json` rather than `<PR-number>.json`: this cites a real public GitHub issue, so per the `issuesOrigin: "public"` row in `.changelogs/README.md` the filename is the issue number. That matches #12255, whose entry for the flat half of this same bug is `.changelogs/2409.json`.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2760

---

> [!NOTE]
> **Medium Risk**
> Changes core empty-row/column detection used by `minSpareRows` and related APIs; behavior is tightly scoped with regression tests but edge cases around `dataDotNotation` and duck-schema grids need careful review.
> 
> **Overview**
> Fixes **GH #5069**: with nested `dataSchema` defaults and columns like `{ data: 'meta.active' }`, spare rows were never treated as empty because defaults were read with flat `schema[prop]` while cell values use a dotted path—so **`minSpareRows` kept appending rows on every change** and **`isEmptyRow` / `isEmptyCol` lied**.
> 
> Adds **`getSchemaDefault()`** in `metaSchema.ts` and wires **`isEmptyRow`** and **`isEmptyCol`** to compare cells against defaults using the same rules as **`DataMap#get`** (own literal keys before path walk, walk only with an explicit `dataSchema` and `dataDotNotation`, safe handling of `null` intermediates and function `data` accessors). **`isObjectEqual`** now accepts **`unknown`** so object-default comparisons need no extra casts.
> 
> Documentation and tests: expanded **`metaManager/AGENTS.md`**, many **`metaSchema` unit tests**, a Playwright **`min-spare-rows-nested-schema`** fixture/spec, and changelog **`.changelogs/5069.json`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5941ad8aaacc0e4b0b284912f846f243a96dceba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2760 Fix minSpareRows with nested dataSchema defaults (#5069)](https://app.clickup.com/t/9015210959/DEV-2760)
Task linked: [DEV-345 default values in dataSchema break minSpareRows behavior](https://app.clickup.com/t/9015210959/DEV-345)

.changelogs/5069.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed `minSpareRows` and `minSpareCols` repeatedly adding rows and columns when `dataSchema` sets a default value on a nested property, such as `{ data: 'meta.active' }`.",
```
demtario: Title overclaims a `minSpareCols` fix that this change can't reach. `Core#adjustRowsAndCols` gates column-appending on `canCreateSpareCols = minSpareCols > 0 && !tableMeta.columns && instance.dataType === 'array'` (`handsontable/src/core.ts:1534`) — array-of-arrays data with no `columns` option. A dotted `prop` (the only case this fix changes) requires object data with an explicit `columns[].data: 'meta.active'`, which makes `canCreateSpareCols` false. The PR's own E2E spec concedes this at `tests/e2e/min-spare-rows-nested-schema.spec.ts:73-74` ("`isEmptyCol` ... is reachable through the public API even where `minSpareCols` is not").

Fix: narrow the title to `minSpareRows` (optionally mention `isEmptyCol`/`countEmptyCols` API correctness instead of `minSpareCols` auto-growth).

demtario (commented):
1 inline note below, plus one off-diff note.

**Off-diff — missing test coverage.** `handsontable/src/dataMap/metaManager/metaSchema.ts` JSDoc and the new `AGENTS.md` bullet both document: a function-typed `columns[].data` accessor is read flat and resolves to `undefined`. No unit test in `handsontable/src/dataMap/metaManager/__tests__/metaSchema.unit.js` exercises `getSchemaDefault`/`isEmptyRow`/`isEmptyCol` with a function-typed `prop` — the one path where the runtime value falls outside the declared `string | number` type. A small addition beside the existing nested-schema mocks would pin the documented behavior.

.changelogs/5069.json
budnix: Good spot — you're right, `canCreateSpareCols` rules that out. Narrowed the title to `minSpareRows` plus the `isEmptyRow()`/`isEmptyCol()` correctness in 5941ad8aaa, and added the function-`columns[].data` unit tests you called out in both describe blocks.

demtario (approved):

Issue #5409, opened 2018-09-18
Pasting tabbed data can deviate from object data schema


cgatesman:
When initializing Handsontable using object data source (https://docs.handsontable.com/5.0.2/tutorial-data-sources.html#page-object), and a dataSchema for that object, I would expect Handsontable to stick with that object schema.

I found that if you paste tabbed data on a right most cell, Handsontable passes changes into `beforeChange`/`afterChange` with prop values as numbers instead of actual prop names because these undefined new columns don't exist.  Ultimately, it causes Handsontable to add additional properties to the data that should not exist in accordance to the `dataSchema`.

Here is a jsfiddle demonstrating this with some instructions on how to reproduce the issue: https://jsfiddle.net/cgatesman/zke1txdL/

As of now, I am correcting this problem with the following `beforeChange` hook, but, to me, this doesn't seem like it's the correct behavior for Handsontable when feeding it object data (not 2 dimensional array data).

```javascript
function (changes, source) {
  if (source === 'CopyPaste.paste' && is.array(changes)) {
    for (var i = changes.length - 1; i >= 0; i--) {
      if (is.number(changes[i][1])) {
        changes.splice(i, 1);
      }
    }
  }
}
```

AMBudnik:
Thank you for sharing the example.

```
{2: "Frank Honest", id: 1, name: "2", address: ""}
{id: 2, name: "Frank Honest", address: ""}
```

breaks the `dataSchema` so it this case the schema is useless.

AMBudnik:
Works the same in 7.3.0 https://jsfiddle.net/hr6g95tb/

![Zrzut ekranu 2020-01-17 o 15 15 01](https://user-images.githubusercontent.com/10757813/72618774-26343c80-393c-11ea-8137-2da8fc067a6b.png)

AMBudnik:
The behavior **does not change** in 8.0.0-beta2-rev15 https://jsfiddle.net/AMBudnik/4ftLyu3s/

pnowak:
This issue is still replicable using v8.0.0. Here's an updated demo - https://jsfiddle.net/zs47trqx/

budnix:
Hi @cgatesman,

Thanks for the report and the jsfiddle. It made this easy to confirm.

[...]

PR #13343, opened 2026-09-02
DEV-2762: Deprecate writing past the last column on object data (#5409)

budnix:
### Context

The PR documents and deprecates a behaviour that has been undocumented since 2018, and is the **non-breaking half** of GitHub issue #5409. The breaking half — actually dropping the write — is DEV-2763, kept separate on purpose.

On an **object** data source, a paste (or a `populateFromArray()` / `setDataAtCell()` call) that runs past the last column writes the overflow value under a property named after the **column index**:

```js
// dataSchema: { id: null, name: null }; copy row 2, paste into row 1's "Name" cell
{ 2: "Frank Honest", id: 1, name: "2", address: "" }
```

`createCol()` refuses to create a column for object data ("you can only have as much columns as defined in first data row, data schema or in the 'columns' setting"), so `applyChanges()` correctly skips creating one — and then falls through to the write anyway. The result is a positional key on a record whose other fields are named. It reproduces on released 18.1.0 and on develop, unchanged since the issue was filed.

**Why this PR does not remove the write.** The value is observable through six public read paths, measured on 18.1.0 after the paste:

| Read path | Returns |
|---|---|
| `getDataAtCell(0, 2)` | `"Frank Honest"` |
| `getDataAtRowProp(0, 2)` / `(0, "2")` | `"Frank Honest"` |
| `getSourceDataAtCell(0, 2)` | `"Frank Honest"` |
| `getSourceData()[0]` | `{2: "Frank Honest", id: 1, …}` |
| `countSourceCols()` | `4` (was 3) |
| `getData()` / `countCols()` | `[1, "2"]` / `2` — the only paths that exclude it |

`colToProp()` echoes the out-of-range index, so `dataMap.get()` finds the value under that key. Only *rendering* excludes it. Against the versioning policy (patch = backward-compatible bug fixes, major = any backward-incompatible change to the public API), removing the write is a major. So this PR ships everything that is **not** backward-incompatible:

1. **Docs.** Nothing described this behaviour anywhere. `setDataAtCell()` JSDoc now states when a column is created and what happens when it cannot be; the `allowInsertColumn` JSDoc bullet no longer conflates column creation with the write; the clipboard guide gains a "Pasting wider than the grid" section.
2. **A one-time deprecation warning** via `deprecatedWarnOnce()`, naming 19.0.0 as the removal release and `setDataAtRowProp()` as the supported way to write a field the grid shows no column for. **The write itself is unchanged.**

Array data sources are deliberately not warned about: there the index names a real array slot rather than a positional key on a named record, so extending the row is type-consistent, and `applyChanges()` creates the column outright when the source allows it.

### How has this been tested?

Four new unit tests in `handsontable/src/plugins/copyPaste/__tests__/copyPaste.unit.ts` — the warning fires, fires **once**, does not fire for an array data source, and the write still happens (so the deprecation is a warning, not a silent behaviour change).

```shell
# new + existing unit tests for this area (10 passed)
npm run test:unit --prefix handsontable -- --testPathPattern='copyPaste.unit'

# E2E for every area that writes past the last column (708 specs, 0 failures)
npm run test:e2e --prefix handsontable -- --testPathPattern='copyPaste|setDataAtCell|formulas'

# types + lint
```

Manual: the original jsfiddle from the issue, re-run in Chrome with a real <kbd>Ctrl</kbd>+<kbd>C</kbd> / <kbd>Ctrl</kbd>+<kbd>V</kbd> — the warning appears once in the console and the paste behaves exactly as before.

### Types of changes

### Related issue(s):

1. DEV-2762
2. Fixes the documentation and deprecation half of #5409
3. Breaking half tracked in DEV-2763

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2762

---

> [!NOTE]
> **Low Risk**
> Non-breaking deprecation and documentation only; writes still occur until 19.0.0, with a console warning on a narrow data-mutation path.
> 
> **Overview**
> Documents and **deprecates** (18.2.0) writing past the last column when the grid cannot add columns—especially **object** (and function `dataSchema`) data—where overflow values still land on a **numeric property key** that no column shows. **Runtime behavior is unchanged** for now; **`setDataAtCell`** emits a **one-time** `deprecatedWarnOnce` pointing to **19.0.0** removal and **`setDataAtRowProp()`** as the supported alternative.
> 
> Docs are updated in the **clipboard** guide (“Pasting wider than the grid”), **`setDataAtCell`** JSDoc, and **`allowInsertColumn`** (clarifies that column creation vs. whether overflow reaches source data are separate). A changelog entry records the deprecation. **CopyPaste** unit tests cover the warning (once per session), continued write until 19.0.0, function `dataSchema`, array growth (no warn), and empty `data: []` bootstrap (no warn).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 50b5da7d90329a4526ce823b7afd9b5830599dab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2762 Document and deprecate the schema-breaking write past the last column (#5409)](https://app.clickup.com/t/9015210959/DEV-2762)
Task linked: [DEV-2763 Stop writing past the last column on an object data source (#5409) — breaking](https://app.clickup.com/t/9015210959/DEV-2763)

docs/content/guides/cell-features/clipboard/clipboard.md
```diff
@@ -325,6 +326,31 @@ The [`beforePaste`](@/api/hooks.md#beforepaste) and [`afterPaste`](@/api/hooks.m
...
+property named after the column index, so no column displays them.
+[`getSourceData()`](@/api/core.md#getsourcedata) returns those properties and
+[`countSourceCols()`](@/api/core.md#countsourcecols) counts them.
+
+::: tip
+
+On an object data source this write is deprecated as of 18.2.0 and will be ignored from 19.0.0 on,
+because the value can never become a column there - it only adds a key your
+[`dataSchema`](@/api/options.md#dataschema) does not declare. Set
+[`allowInsertColumn`](@/api/options.md#allowinsertcolumn) to `false` to drop such values today, or
```
demtario: **High:** this tip says to set `allowInsertColumn` to `false` to drop these values on an object data source today, but that's not true. `applyChanges()` in core.ts only consults `allowInsertColumn` inside the `dataType === 'array'` branch; for object/function data, `datamap.set()` always writes the value regardless of this option. This also directly contradicts the metaSchema.ts text added in this same PR ("Whether a write past the last column reaches the source data at all is **not** controlled by this option"). Suggest dropping the `allowInsertColumn` clause from this tip — currently there's no lever that prevents the write on an object/function data source.

handsontable/src/plugins/copyPaste/__tests__/copyPaste.unit.ts
```diff
@@ -120,3 +127,132 @@ describe('CopyPaste ragged clipboard', () => {
...
+    // slot rather than a property the schema never declared.
+    expect(pastLastColumnWarnings()).toHaveLength(0);
+    expect(hot.countCols()).toBe(3);
+
+    hot.destroy();
+  });
+
+  it('should not warn for a grid that declares no columns at all', () => {
+    const hot = new Handsontable(document.createElement('div'), {
+      data: [],
```
demtario: **Medium:** this new `describe` block tests `Core.setDataAtCell`'s deprecation branch, but it's added to the CopyPaste plugin's unit test file. This case in particular doesn't exercise paste at all — it calls `hot.setDataAtCell()` directly. A core-level Jest home already exists for this (`handsontable/src/__tests__/core/core.unit.js`), with no `setDataAtCell` coverage yet. Suggest moving the pure-API assertions there; paste-specific cases can stay here if paste-path coverage is also wanted.

handsontable/src/core.ts
```diff
@@ -2684,6 +2697,31 @@ export default function Core(
...
       if (visualColumnIndex >= this.countCols()) {
+        // No column exists at this index, and on an object-rowed data source none ever can -
+        // `createCol()` refuses one ("you can only have as much columns as defined in first data
+        // row, data schema or in the 'columns' setting"), which is why `applyChanges()` skips
+        // creating it. The index then travels on as the property name, so `dataMap.set()` mints a
+        // positional key on a row whose other fields are named: `{ 2: 'x', id: 1 }` (#5409). No
+        // column renders it, yet it reaches every consumer that serializes the row. Deprecated in
+        // 18.2.0; the write is skipped from 19.0.0 on.
+        //
+        // The predicate mirrors `applyChanges()`'s own creation gate, which is `=== 'array'` - so
```
demtario: **Low:** the comment's justification is inaccurate for the function-`dataSchema` case. `isColumnModificationAllowed()` is `!(dataType === 'object' || tableMeta.columns)` — it doesn't check for `'function'`, so for a function schema with no `columns` option it returns `true`, and `createCol()` does **not** refuse. The end behavior is still correct because `applyChanges()`'s column-creation branch is itself gated on `dataType === 'array'` directly (not just `!== 'array'`, but a 3-way AND with `!tableMeta.columns` and `tableMeta.allowInsertColumn`), so `createCol()` is never reached for non-array data regardless of `isColumnModificationAllowed()`. Worth correcting the stated reasoning.

handsontable/src/core.ts
```diff
@@ -2650,6 +2650,19 @@ export default function Core(
    * Set new value to a cell. To change many cells at once (recommended way), pass an array of `changes` in format
    * `[[row, col, value],...]` as the first argument.
    *
+   * Writing past the last column creates the missing columns only where the grid can create them: an
+   * array-of-arrays [`data`](@/api/options.md#data) source with no [`columns`](@/api/options.md#columns) option and
+   * [`allowInsertColumn`](@/api/options.md#allowinsertcolumn) left on. In every other configuration the column count
+   * is fixed, and the value is instead written to a property named after the column index. That property is not part
+   * of your [`dataSchema`](@/api/options.md#dataschema) and no column displays it, but
+   * [`getSourceData()`](@/api/core.md#getsourcedata) returns it and
+   * [`countSourceCols()`](@/api/core.md#countsourcecols) counts it.
```
demtario: **Low:** "`countSourceCols()` counts it" overstates — `countSourceCols()` → `countFirstRowKeys()` only inspects `data[0]`. A write past the last column on row 1+ still lands as a phantom key on that row (and `getSourceData()` returns it), but `countSourceCols()` won't reflect it unless row 0 also has the key. Same overstatement appears in the new clipboard.md text (lines ~339-340).

demtario (commented):
4 inline notes below.

docs/content/guides/cell-features/clipboard/clipboard.md
budnix: Fair point that those two texts disagree, but it's the option note that was wrong — a paste breaks out at `core.ts:1794` before anything is written, and only a direct `setDataAtCell()` writes past. Rewrote that note in fcaf6dabd7 and pinned the paste case with a test.

handsontable/src/plugins/copyPaste/__tests__/copyPaste.unit.ts
budnix: Good spot — moved the pure-API case to `core.unit.js` in fcaf6dabd7, with direct `setDataAtCell()` coverage for the other branches too.

handsontable/src/core.ts
budnix: You're right, `isColumnModificationAllowed()` never sees `'function'`. The comment now cites `applyChanges()`'s own gate instead, in fcaf6dabd7.

handsontable/src/core.ts
budnix: Agreed, it only reads the first row's keys. Fixed in both `core.ts` and `clipboard.md` in fcaf6dabd7.

demtario (approved):

PR #13344, opened 2026-09-02
DEV-2763: Stop writing past the last column on object data (#5409)

### LINKED ISSUE #5409: Pasting tabbed data can deviate from object data schema

budnix:
### Context

The PR fixes GitHub issue #5409 and is the **breaking half** of it. The docs and the one-time deprecation warning are #13343 (DEV-2762), which this branch is stacked on and which should land first.

On an **object** data source, a paste (or a `populateFromArray()` / `setDataAtCell()` call) that runs past the last column wrote the overflow value under a property named after the **column index**:

```js
// dataSchema: { id: null, name: null }; copy row 2, paste into row 1's "Name" cell
{ 2: "Frank Honest", id: 1, name: "2", address: "" }
```

`createCol()` refuses to create a column for object data ("you can only have as much columns as defined in first data row, data schema or in the 'columns' setting"), so `applyChanges()` correctly skips creating one — and then fell through to the write anyway. The result was a positional key on a record whose other fields are named. Reproduced on released 18.1.0 and on develop, unchanged since 2018.

The fix is three lines in `Core#setDataAtCell`: an index at or past `countCols()` skips the change when the data source is object-based. One guard closes every object-data write path, because `grid.populateFromArray` **ends by calling** `setDataAtCell`.

### Breaking change and migration

The dropped value was observable through six public read paths, measured on 18.1.0 after the paste:

| Read path | Before | After |
|---|---|---|
| `getDataAtCell(0, 2)` | `"Frank Honest"` | `null` |
| `getDataAtRowProp(0, 2)` / `(0, "2")` | `"Frank Honest"` | `undefined` |
| `getSourceDataAtCell(0, 2)` | `"Frank Honest"` | `null` |
| `getSourceData()[0]` | `{2: "Frank Honest", id: 1, …}` | `{id: 1, …}` |
| `countSourceCols()` | `4` | `3` |
| `getData()` / `countCols()` | `[1, "2"]` / `2` | unchanged |

Under the [versioning policy](https://handsontable.com/docs/javascript-data-grid/versioning-policy/) (patch = backward-compatible bug fixes, major = any backward-incompatible change to the public API) that makes this a **major**. Migration guide entry added at `docs/content/guides/upgrade-and-migration/migrating-from-18.2-to-19.0/`, with before/after examples and the `setDataAtRowProp()` replacement.

> **Note on the migration guide file.** PR #13330 (DEV-2723, also on hold for this major) creates the same page. This branch writes it with only its own section, matching that PR's frontmatter and heading style exactly, so whichever lands first the other rebases into a clean concatenation. No content is lost either way.

### Three things deliberately left uncapped — each has a test that fails otherwise

1. **Array data, even when `columns` narrows it.** The index names a real array slot rather than a positional key on a named record, so extending the row is type-consistent. Widening the guard to `!isColumnModificationAllowed()` (also false whenever `columns` is set) reads well, and was the first version of this fix — it failed exactly **1 spec out of 9878**: `formulas.spec.js`'s "should not throw an error while performing actions on HOT with renamed sheet", which writes to column 1 of a one-`columns` array grid and reads the value back. All five targeted suites were green with the wrong guard; only the full run caught it. Note *reachability* does not separate the two cases — `getDataAtCell(row, outOfRangeCol)` reads the value back on both, because `colToProp()` echoes the index.
2. **`setDataAtRowProp()`.** The caller names the field to write, so a numeric prop there is a real request. `tests/e2e/set-data-at-row-prop-old-value.spec.ts:111` pins the literal `0` key it creates.
3. **`colToProp()` / `propToCol()`.** The out-of-range pass-through is by design (#5945) and `applyChanges()` grows an array grid from it, so the cap sits at the write site rather than in either translation method. This also keeps the change compatible with #13330, which deliberately keeps the index in the `beforeChange` / `afterChange` payload — this guard drops the change entirely, so no payload entry is reported for a value that never landed.

### How has this been tested?

Four unit tests in `copyPaste.unit.ts` (two fail without the guard; the two array-data tests pass either way, which is what makes them regression guards), two cases in `setDataAtCell.spec.js` next to the existing `maxCols` skip precedent, and one strengthened assertion in `paste.spec.js` — which previously asserted only `getData()`, bounded by `countCols()`, and so could not tell a dropped value from one parked past the end of the row.

```shell
# unit (10 passed)
npm run test:unit --prefix handsontable -- --testPathPattern='copyPaste.unit'

# full legacy E2E — 9878 specs, 0 failures

# the Playwright discriminator — 13 passed, including the numeric-key case
cd tests && npm run test:e2e -- e2e/set-data-at-row-prop-old-value.spec.ts

# types + lint
```

Manual: the original jsfiddle from the issue, re-run in Chrome with a real <kbd>Ctrl</kbd>+<kbd>C</kbd> / <kbd>Ctrl</kbd>+<kbd>V</kbd> — row 0 stays `{ id: 1, name: "2", address: "" }` and `afterChange` reports only the one change that landed.

### Types of changes

### Related issue(s):

1. DEV-2763
2. Fixes #5409
3. Stacked on #13343 (DEV-2762), the docs and deprecation half
4. Shares the 19.0 migration guide page with #13330 (DEV-2723)

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2763

---

> [!NOTE]
> **High Risk**
> Breaking public API: integrators who relied on overflow values in `getSourceData()` or change hooks must migrate; the change touches all bulk write paths through `setDataAtCell`.
> 
> **Overview**
> **Breaking (19.0):** On object-rowed data, `setDataAtCell` (and anything that ends there—paste, `populateFromArray`, autofill) no longer writes overflow values as numeric property keys the schema never declared. Those cells are **skipped**: nothing is persisted and **`beforeChange` / `afterChange` omit them**. Array data, empty grids with no columns, and `setDataAtRowProp` are unchanged.
> 
> The guard lives in **`Core#setDataAtCell`**: column index ≥ `countCols()` when `dataType !== 'array'` and `countCols() > 0`. The 18.2 deprecation warning is removed. If every requested change is skipped, the method returns early so an open editor in another cell is not closed.
> 
> Docs add a **18.2 → 19.0 migration guide**, update the clipboard “pasting wider than the grid” section, and register a **breaking changelog** for #5409. Tests cover object vs array behavior, hook payloads, and editor preservation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 517b21f6ca7bff8419750c04ea36e90f6027bbf3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2762 Document and deprecate the schema-breaking write past the last column (#5409)](https://app.clickup.com/t/9015210959/DEV-2762)
Task linked: [DEV-2763 Stop writing past the last column on an object data source (#5409) — breaking](https://app.clickup.com/t/9015210959/DEV-2763)

PR #13345, opened 2026-09-02
DEV-2714: Keep the release from re-resolving example lockfiles

demtario:
### Context

The PR stops the release from re-resolving the nine tracked example lockfiles, and reports the delta when one changes anyway.

`publish.yml`'s `first-rc-build` is the only automated writer of `examples/next/{docs,visual-tests}/<framework>/package-lock.json`. It installs the examples and then commits the whole tree with a bare `git add .` (`publish.yml:350`). Nothing validated those nine files: no `npm ci`, no `--frozen-lockfile`, no audit step, no Dependabot entry, and `lockfile-float-gate.mjs` hardcodes `pnpm-lock.yaml`. So once per release the example dependency graph was re-resolved and committed with nobody told.

**The cause was not the `latest` pins.** 104 leaf specifiers are `latest`, but npm honours a locked edge for a dist-tag spec, so with the lockfile present `npm install` re-resolves nothing. `examples/scripts/clean-subpackages.mjs` was deleting all nine lockfiles before the install ran, which left the resolver nothing to reuse. That is the defect this PR removes.

The follow-up question on the earlier review asked whether this needs an `overrides` block in the example manifests or a new mode on the float gate. Neither, as posed:

- **No `overrides`.** `install-subpackages.mjs`'s `alignTypescriptWithAngular` *writes* `devDependencies.typescript` into every Angular leaf manifest at install time, and the `latest` pins are deliberate. A hand-maintained pin list would fight it and still would not express "the set CI tested".
- **No new float-gate site.** `--e` means `--exclude` (`scripts/translate-to-native-npm.mjs:20-21`), so `rc-build`, `stable-prepare`, `stable-build` and `stable-publish` all opt *out* of examples. `examples:install` appears exactly once in the workflow, so there is one place to cover, not six. The gate script is untouched and its pinned call count stays at 6.

The new check is **report-only on purpose**: these trees build documentation demos and visual-test fixtures and ship in no bundle, so a transitive patch bump must not red a release candidate. It writes the diff, any untracked lockfile no ignore rule covers, and each changed file's `npm audit` counts to `$GITHUB_STEP_SUMMARY`, which needs no token and works on forks.

#### Keeping the Angular TypeScript alignment on the same source of truth

`alignTypescriptWithAngular` resolved the TypeScript range from a live `npm view @angular-devkit/build-angular@latest`, which was right while the lockfiles were deleted on every install and `latest` decided what got installed. With the lockfiles reused, the lockfile decides, so reading the range from `latest` would break the coupling: on the day Angular bumps its supported TypeScript range, the leaf manifests would get `~6.1.0` while the lockfile still installed builder `22.1.5`, `--legacy-peer-deps` would swallow the peer conflict, and `ng build` would fail on the compiler's own version check inside `first-rc-build`.

So the alignment now reads the range out of the framework's own `package-lock.json`. `lockfileVersion: 3` records `peerDependencies` per package, so the locked builder supplies it with no registry call at all (`22.1.5` declares `>=6.0 <6.1`). When there is no lockfile, which is exactly the post-`--reset-lockfiles` state, it falls back to `npm view` as before, so a deliberate refresh still moves Angular and TypeScript together.

Three smaller fixes ride along, all in the same "committed unchecked" class:

- `examples/.gitignore` named `angular`, `react` and `vue` where the directories are `angular-wrapper`, `react-wrapper` and `vue3`, and had no `visual-tests` patterns at all, so 5 of the 23 per-example lockfile paths fell through into `git add .`'s reach (the other 18 were covered by per-leaf `.gitignore` files). All 23 are covered now, the old names are kept alongside the new ones because the install globs still accept them, and the nine shared lockfiles stay tracked.
- `clean-subpackages.mjs` refuses `--reset-lockfiles` without a version instead of quietly cleaning the workspace root and leaving all nine lockfiles untouched.
- `examples/README.md` documents that the shared lockfiles are reused rather than recomputed, and gives the deliberate-refresh recipe.

### How has this been tested?

Automated, in `.github/scripts/__tests__/release-lockfile.test.mjs` (5 new cases, 13 total in the file, 367 in the tooling suite):

```bash
node --test .github/scripts/__tests__/release-lockfile.test.mjs
npm run test:tooling
npm run eslint
```

1. `clean-subpackages.mjs` never deletes a lockfile unconditionally (both removals sit inside the `--reset-lockfiles` guard, and the opt-in still exists).
2. `first-rc-build` runs the report, in the order report → float gate → commit, and the report step does not name the gate's path.
3. The report script exits 0 on a dirty tree, names the changed tracked lockfile and an untracked lockfile at an uncovered path, and stays quiet about a properly ignored per-example one.
4. On a checkout it cannot diff (an unborn `HEAD`), the report exits 0 and says it could not determine the answer rather than printing the all-clear.
5. `--reset-lockfiles` with no version exits 1 with a message naming the missing argument.

Tests (1) and (2) were mutation-checked: unguarding the deletions fails (1), and moving the report step below the float gate fails (2) along with the pre-existing `every release commit is preceded by the gate`.

Manually, the central claim is that a preserved lockfile makes the install reproducible. All nine trees, installed in the checkout against their committed locks:

```bash
for d in examples/next/docs/* examples/next/visual-tests/*; do
  case "$d" in *angular*) extra=--legacy-peer-deps;; *) extra=;; esac
  ( cd "$d" && npm install --no-audit --ignore-scripts $extra >/dev/null 2>&1; echo "$? $d" )
done
git status --porcelain -- examples/next
```

Nine zeros and a clean status, including `docs/react-wrapper` (whose lock carries hand-pruned `@handsontable/pikaday` entries) and both Angular trees. Also verified separately:

- A second `npm install` over a symlinked `node_modules/handsontable`, which is what `link-packages.mjs` leaves behind and what `publish.yml:292`'s `prebuild` install then runs against: exit 0, lock untouched.
- `docs/js` with a fresh `npm_config_cache`, to rule out the cache answering instead of the registry: lock untouched.
- `node ./scripts/clean-subpackages.mjs next` keeps all nine; adding `--reset-lockfiles` deletes exactly those nine.
- The report script against a real dirty lockfile, with the audit leg live: `5 high, 3 moderate` for `docs/js`, plus the untracked entry, exit 0. Against a directory that is not a checkout it reports "could not determine" instead of the all-clear, and an unwritable `$GITHUB_STEP_SUMMARY` no longer exits non-zero.
- `git check-ignore` over all 23 per-example lockfile paths: 23 ignored (was 18), and the 9 tracked shared lockfiles still unignored.
- The lock-derived TypeScript alignment resolves `~6.0.0` for all five Angular examples, matching what their manifests already pin, so it is a no-op today; a directory with no lockfile resolves `null` and takes the `npm view` fallback. Running the real `install-subpackages.mjs` over `docs/angular-wrapper` (alignment, install and `link-packages` included) left both the manifests and the lockfile untouched.

Not verified locally, and it cannot be: `first-rc-build` end to end. That job runs once per release and only on a real RC cut, so the wiring is asserted by test (2) rather than exercised.

### Types of changes

### Related issue(s):

1. DEV-2714

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

`[skip changelog]` — CI and tooling only. No package source changes (`handsontable/src/**`, `wrappers/**`), so the changelog gate passes on path; this states it explicitly.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2714

---

> [!NOTE]
> **Medium Risk**
> Changes release automation and example install/clean behavior on the only job that runs `examples:install` and commits examples; mistakes could still let unreviewed lockfiles land on the release branch, though report-only visibility and tests reduce that risk.
> 
> **Overview**
> Stops **first RC** cuts from silently re-resolving the nine tracked `examples/next/.../package-lock.json` files and committing the result via `git add .`.
> 
> **Root fix:** `clean-subpackages.mjs` no longer removes framework lockfiles on every `examples:install`; removal is opt-in with `--reset-lockfiles` (and that flag now fails without a version). `examples/.gitignore` ignores all example lockfiles except the shared framework-level ones, fixing gaps for `visual-tests` and misnamed framework dirs.
> 
> **Visibility:** New report-only step in `first-rc-build` runs `example-lockfile-report.mjs` (always exit 0) before the pnpm float gate and commit. It surfaces lockfile/manifest diffs vs `HEAD`, untracked shared lockfiles, and optional `npm audit` summaries to the job log/summary without blocking the RC.
> 
> **Angular install alignment:** `alignTypescriptWithAngular` derives the TypeScript tilde range from the framework lockfile (locked `typescript` or builder peer deps) instead of `npm view ...@latest`, so reused locks stay consistent with leaf manifest writes during install.
> 
> **Guardrails:** `release-lockfile.test.mjs` adds tests for clean-script guards, workflow step order, and report-script behavior; `AGENTS.md` and `examples/README.md` document the workflow.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cb7ab07a07fdab877797031d1c402621900a391c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2714 Guard example lockfiles against re-float at release time](https://app.clickup.com/t/9015210959/DEV-2714)

examples/.gitignore
```diff
@@ -2,9 +2,28 @@ tmp/
...
 */docs/vue/*/package-lock.json
+*/docs/vue3/*/package-lock.json
+*/visual-tests/angular/*/package-lock.json
+*/visual-tests/angular-wrapper/*/package-lock.json
+*/visual-tests/js/*/package-lock.json
+*/visual-tests/react/*/package-lock.json
+*/visual-tests/react-wrapper/*/package-lock.json
+*/visual-tests/ts/*/package-lock.json
+*/visual-tests/vue/*/package-lock.json
+*/visual-tests/vue3/*/package-lock.json
```
budnix: The list below names each framework directory one at a time, but two of the names the install globs accept are wildcards, and `link-packages.mjs` line 23 already special-cases `angular-9` and `angular-10`. I checked with `git check-ignore`: a future `angular-ssr` or `vue2` directory is not covered, so its leaf lockfiles land back inside the release job's `git add .`, which is the same gap this block is closing for `angular`, `react` and `vue`. A glob plus a negation covers every name and every category, and it also picks up the `examples/package-lock.json` at the root that nothing ignores today.

```suggestion
# Per-example lockfiles, created by running `npm install` inside an example folder. Only the
# shared framework-level lockfile is tracked. See the explanation in /examples/README.md.
#
# Ignore every lockfile under `examples/`, then re-include the shared framework-level ones --
# the only lockfiles that sit at exactly `<version>/<category>/<framework>/`. Written as a pair
# so a new framework or category directory needs no edit here: the install globs accept any
# `angular-*` and `vue*` name, and a name with no rule puts its leaf lockfiles back within
# reach of the release job's bare `git add .`.
**/package-lock.json
!*/*/*/package-lock.json
```

examples/scripts/install-subpackages.mjs
```diff
@@ -69,7 +112,12 @@ const alignTypescriptWithAngular = async (frameworkUrl) => {
       continue;
     }
 
-    const typescriptRange = await resolveAngularTypescriptRange(builderSpec);
+    const exampleDir = path.basename(path.dirname(packageJsonPath));
+    const lockedRange = lockedTypescriptRange(frameworkUrl, exampleDir);
+    // Follow the lockfile when there is one, and the registry only when there is not.
+    const typescriptRange = lockedRange
```
budnix: `lockedTypescriptRange` returns `null` both when there is no lockfile and when the lockfile has no builder under either key it tries, so the second case quietly falls back to `npm view ...@latest` -- the one answer the comment above says is wrong while a lockfile exists. Reading `packages['node_modules/typescript'].version` would skip the key guessing altogether: it is `6.0.3` in both Angular lockfiles, which gives `~6.0.0`, exactly what all five leaf manifests already pin. It would also leave both helpers returning the same kind of string, so a later edit cannot feed `~6.0.0` back through the range regex and silently get `null`. One small thing while you are in here -- the lockfile is read and parsed again for every example, about 2.6 MB of JSON across the five Angular ones, so hoisting the read to the top of `alignTypescriptWithAngular` would be nice.

.github/scripts/example-lockfile-report.mjs
```diff
@@ -0,0 +1,210 @@
...
+ */
+import { execFileSync } from 'node:child_process';
+import { appendFileSync } from 'node:fs';
+import path from 'node:path';
+import { pathToFileURL } from 'node:url';
+
+// `:(glob)` asks for strict pathname semantics: a single `*` stops at a `/`, and only `**` spans
+// directory levels. Without it git's default matching lets `*` swallow separators, so a pattern
+// meant for one level quietly matches any depth.
+const LOCKFILE_PATHSPEC = ':(glob)examples/**/package-lock.json';
```
budnix, line 38: The pathspec covers lockfiles only, but `alignTypescriptWithAngular` writes `devDependencies.typescript` straight into the Angular leaf `package.json` files at install time, and the same `git add .` commits those. So a real manifest edit can reach the release branch with nothing mentioning it, while the lockfile sitting next to it would have been reported. A second diff line for `package*.json` would catch both, though it needs to stay out of the audit loop, since a leaf directory has no lockfile of its own. Curious whether the manifests were meant to be out of scope here.

.github/scripts/example-lockfile-report.mjs
```diff
@@ -0,0 +1,210 @@
...
+      + 'failed (see the job log for the error). Treat this as unknown, not as unchanged, and '
+      + 'check `git status -- examples` on the release branch before trusting the commit.';
+  }
+
+  if (changed.length === 0 && untracked.length === 0) {
+    return 'The tracked example lockfiles are unchanged. This release installs the dependency '
+      + 'set CI tested.';
+  }
+
+  const out = ['### Example lockfiles changed during this cut', ''];
```
budnix: I ran this against a tree where nothing had changed but one lockfile sat at a path no rule ignores, and the output came back as "Example lockfiles changed during this cut" followed by advice to run `--reset-lockfiles` and review the diff. Nothing had changed and the real fix was an ignore pattern, so both the heading and the closing paragraph send the release operator the wrong way. This is the one case the untracked scan exists for, so it seems worth branching the heading and the remedy on whether `changed` is empty.

.github/scripts/__tests__/release-lockfile.test.mjs
```diff
@@ -351,3 +358,199 @@ test('every step that reads $VERSION also sets it', () => {
...
+  );
+
+  // Guarded means indented deeper than the removals that run unconditionally, which is what
+  // sitting inside the `if (resetLockfiles)` block looks like. Text-based on purpose: the
+  // regression to catch is a future edit moving these two lines back out of the block.
+  const baseline = Math.min(...removals.filter(match => !isLockfile(match)).map(([, indent]) => indent.length));
+
+  for (const [, indent, call] of lockfileRemovals) {
+    assert.ok(
+      indent.length > baseline,
```
budnix: Both checks in this test still pass if the guard gets flipped to `if (!resetLockfiles)` -- the two lines stay indented deeper than the unconditional ones, and the `args.includes('--reset-lockfiles')` line is untouched. That is exactly the regression the message right below describes, so the test reads as stronger than it is. Matching the guard's own text would close it cheaply, something like asserting `if (resetLockfiles) {` appears between the last unconditional push and the first lockfile one.

.github/workflows/publish.yml
```diff
@@ -331,6 +331,23 @@ jobs:
...
+      # must not red a release candidate. It must not be invisible either.
+      # Must stay ABOVE the float gate: `release-lockfile.test.mjs` asserts the
+      # gate is the step immediately before `Commit and push`.
+      #
+      # Runs even after a failed step, because the lockfiles are a candidate
+      # cause: if the examples install or build broke, the diff is the first
+      # thing the operator needs, and a step that only runs on the happy path
+      # would have nothing to say exactly when it matters. The job still fails.
+      - name: Report example lockfile changes
+        if: ${{ !cancelled() }}
```
budnix, line 348: Running this after a failed step makes sense for the diff, which is the part an operator actually needs. The audit leg is the awkward half: a half-finished `examples:install` can leave all nine lockfiles dirty, and each gets its own `npm audit` with a 60 second timeout, so a job that is already red can sit here for several more minutes producing counts from a broken tree. Setting `SKIP_AUDIT` when the job status is no longer success would keep the useful half and drop the rest.

examples/.gitignore
demtario: Taken as suggested, in cb7ab07. You are right that the per-name list could not keep up: `internal.framework_dirs` accepts `angular-*` and `vue*`, so the list was always one directory behind.

Verified with `git check-ignore` after the change:

- all 23 per-example lockfile paths ignored, and the 9 shared ones still not ignored
- `docs/angular-ssr/demo`, `docs/vue2/demo`, a new `e2e` category, and a versioned `18.2.0/docs/svelte/demo` all covered by `**/package-lock.json`
- `examples/package-lock.json` at the root now ignored, which nothing did before
- a hypothetical future shared lockfile at `18.2.0/docs/svelte/package-lock.json` correctly re-included by the negation, so the pair works in both directions

One consequence worth recording, because it changed the script next door: with every lockfile ignored except the shared level, the report's untracked scan can no longer surface a leaf lockfile at all. The only thing it can now find is a shared lockfile that was never committed, so "add an ignore rule" would be half the story. Its remedy now says to either commit it on `develop` as a reviewed resolution or ignore it there. Noted in `AGENTS.md` too.

examples/scripts/install-subpackages.mjs
demtario: Right on all three counts, fixed in cb7ab07, with one deliberate difference from what you sketched.

The locked `typescript` version is now the primary answer, for the reason you give: it is what actually installs, and npm resolved it under the locked builder's peer range in the first place. Both Angular lockfiles have `typescript@6.0.3`, which gives `~6.0.0` and matches all five leaf manifests, so the alignment is a no-op today.

I kept the builder's `peerDependencies.typescript` as an explicit second tier rather than deleting it, for the case of a lockfile that pins the builder but not TypeScript. That case is unlikely, but the fallback after it is `npm view ...@latest`, which is the answer this whole change exists to avoid while a lockfile is present, so I would rather exhaust the lockfile first. Both tiers are explicit `if (spec) return spec` checks, so neither can silently fall through the way the old single `null` did.

The regex trap is closed as well: the helper is now `typescriptSpecFrom` and matches `(\d+)\.(\d+)` instead of requiring `>=`, so it accepts a bare version and a range alike and both return the same kind of string. That was the specific thing that would have turned `~6.0.0` back into `null` on a later edit.

And the read is hoisted to the top of `alignTypescriptWithAngular`, so it is once per framework directory instead of once per example.

.github/scripts/example-lockfile-report.mjs
demtario: Out of scope by oversight, not by design. You are right that it is the same hazard: the alignment writes `devDependencies.typescript` into the leaf manifests and the same `git add .` commits them, so a manifest edit could reach the release branch while the lockfile beside it would have been reported.

Added in cb7ab07. `MANIFEST_PATHSPEC` is a second `--name-only` diff plus a shared `--stat`, reported in its own section, and kept out of the audit loop exactly as you say, since a leaf directory has no lockfile for `npm audit` to read.

Verified by pointing a real manifest at `~6.1.0`, which is the shape the alignment would write on the day Angular bumps its range:

```
### Example lockfiles changed during this cut

examples/next/docs/angular-wrapper/demo/package.json | 4 ++--

Manifests changed (`alignTypescriptWithAngular` writes these at install time):

- `examples/next/docs/angular-wrapper/demo/package.json`
```

There is a test for it now too, and following your point about wrong remedies below: a manifests-only change gets a different closing line, because `--reset-lockfiles` would not reproduce it. It points at `examples:install` instead.

.github/scripts/example-lockfile-report.mjs
demtario: Confirmed and fixed in cb7ab07. Both halves were wrong for that case: nothing had changed, and the fix was an ignore rule, not a refresh.

The heading and the closing paragraph now branch on whether anything tracked actually moved. Untracked-only gets `### Untracked example lockfiles found during this cut` and a remedy about `examples/.gitignore`; a real change keeps the old heading and the refresh recipe.

Your point generalized further than the case you hit, so thank you for it. A manifests-only change had the same defect from the opposite direction: it is a real change, so it took the "changed during this cut" branch, but `--reset-lockfiles` does not reproduce a manifest edit either. That closing line now points at `npm run examples:install next` instead.

And because of the `.gitignore` pair above, the untracked branch's advice had to change once more: with every lockfile ignored except the shared level, an untracked hit can only be a shared lockfile that was never committed, so it now says commit it as a reviewed resolution or ignore it, rather than assuming an ignore rule is what is missing.

Two tests cover the wording, asserting the right heading appears and the wrong remedy does not.

.github/scripts/__tests__/release-lockfile.test.mjs
demtario: Correct, and this was the sharpest one. The indentation check could not see a negated guard, so the test read as stronger than it was.

Fixed in cb7ab07 along the lines you suggest: it now takes the last unconditional `removes.push(...)` and the first lockfile one, and asserts `if (resetLockfiles) {` appears in the source between them. The indentation check stays as a secondary assertion.

Mutation-checked both ways, since a test I only assert about is worth as little as the one you caught:

- guard flipped to `if (!resetLockfiles)`: `not ok 9 - the examples clean step keeps the lockfiles unless asked` (it passed before this change, which is your point exactly)
- guard restored: 15/15 pass

I did consider a behavioral test instead, running the script with and without the flag against a fixture and asserting the lockfiles survive. It cannot run in the `tooling tests` job: that job installs only the root workspace, so `examples`' `rimraf` is absent, which is why the argument validation in this script now happens before the dependency loads. So the source assertion is the strongest form available here.

.github/workflows/publish.yml
demtario: Agreed, and taken as suggested in cb7ab07:

```yaml
env:
  SKIP_AUDIT: ${{ job.status != 'success' && '1' || '' }}
```

The script already treats any truthy `SKIP_AUDIT` as "skip", and an empty string is falsy, so the happy path is unchanged and a red job keeps only the diff. Nine dirty lockfiles at up to 60s each was the worst case you describe, and the counts would have been taken from a half-installed tree anyway.

The 60s timeout itself is worth keeping for the green path, since without it a hanging registry stalls the job with nothing to catch it.

budnix (approved):

demtario:
cursor review

PR #13346, opened 2026-09-02
DEV-2719: Detach the resize handle and guide on plugin teardown

demtario:
### Context

The PR fixes both manual resize plugins leaving their handle and their guide inside the application's own container after the plugin is turned off.

`ManualRowResize` and `ManualColumnResize` create the two elements in the constructor and append them to `hot.rootElement` lazily — the handle in `setupHandlePosition()` on `mouseover` over a header, the guide in `setupGuidePosition()` on `mousedown` over the handle. Nothing detached them again except the context-menu handler: `hideHandleAndGuide()` only stripped the `active` class, `disablePlugin()` detached nothing, and `destroy()` only called `super.destroy()`.

The handle is the harmful half, and the symptom is not the one it looks like. It is `opacity: 0` at rest, so nothing looks wrong, but it keeps `z-index: 210`, `pointer-events: auto`, `cursor: row-resize` and `opacity: 1` on `:hover`. So after `updateSettings({ manualRowResize: false })` the orphan lights up with a resize indicator and a resize cursor for a feature that is off, **and it swallows the click on the header underneath it**, because the core resolves the cell from `event.target` and the target is the orphan `div`. Measured on `handsontable@18.0.0`, clicking the strip's own coordinates with the plugin disabled:

| | `hot.getSelected()` |
|---|---|
| orphan present | `undefined` (nothing selected) |
| orphan removed by hand | `[[1, -1, 1, 4]]` (row selected) |

The guide is inert by comparison — `display: none` without `active`, so 0×0, no paint and no hit test — but it is detached on the same paths for symmetry.

Both plugins gain a private `#detachHandleAndGuide()`, called from `disablePlugin()`, `destroy()` and the context-menu handler, so one teardown serves all three.

#### Three deliberate choices worth a reviewer's attention

**The teardown is not wired into `hideHandleAndGuide()`**, which is what the review of #13319 proposed. `#onMouseUp()` calls `hideHandleAndGuide()` and then re-runs `setupHandlePosition()`, and `shouldSkipResizeHandlePositioning()` returns `true` when the click count is above one — exactly the second `mouseup` of a double-click. A detach inside `hideHandleAndGuide()` would therefore leave the handle **detached** for the 500ms until `afterMouseDownTimeout()` restores it, i.e. a visible flicker on every double-click autosize. `keeps the row handle attached through a double-click autoresize` pins that path.

**The pressed flag is not reset inside the shared teardown**, only at the context-menu call site where aborting the drag is the point. `updatePlugin()` runs `disablePlugin(); enablePlugin();` on any `updateSettings()` carrying the plugin's own key, which is what a framework wrapper sends on every re-render — so clearing the flag in the shared helper made the `mouseup` ending an in-flight drag take the idle branch, dropping the drag with no `afterRowResize`/`afterColumnResize` and the dragged size never confirmed. With the reset kept at its call site, `#onContextMenu()` is byte-equivalent to before.

**A completed drag deliberately still leaves both elements attached.** Detaching the guide at `mouseup` was implemented and then reverted: two specs added by #13319 take `expect(grid.rowGuide).toHaveCount(1)` after a drag as their documented premise, because a completed drag is the only gesture that attaches the guide, and they exist to prove the context-menu handler still detaches one that the working path attached. Since an orphaned guide is inert, the change bought nothing user-visible and cost that regression coverage. `keeps both row elements attached while the plugin stays on` states the kept behavior explicitly so it is not mistaken for an oversight.

#### Behavior change to be aware of

`updatePlugin()` runs `disablePlugin(); enablePlugin();` whenever an `updateSettings()` call carries the plugin's own key, so `updateSettings({ manualRowResize: true })` or `{ manualRowResize: [40, 50] }` now also detaches the handle until the next `mouseover`. Previously it survived, at a position that may already have been stale. A `rowHeights`/`colWidths`-only update takes the other branch of `updatePlugin()` and does not re-initialize, so the common size-update path keeps the handle. This is not a breaking change: no default value moves and no API is renamed.

### How has this been tested?

New Playwright spec `tests/e2e/manual-resize-teardown.spec.ts` (11 tests) with its fixture and page object. The shared locators and reveal gestures were extracted from `ManualResizeContextMenuPage` into a `ManualResizePage` base, which both page objects now extend — a structural change only, no behavior touched.

Nothing here can be covered in jsdom: the handle exists only once the header has a layout box.

Attachment is asserted with `toHaveCount`, never with a visibility matcher. Both matchers are blind on these elements — the guide is already `display: none`, and the handle has a real box at `opacity: 0`, which Playwright reports as *visible* — so neither can tell attached-but-inert from detached, and a spec built on either would be green on the unfixed code. The swallowed-click test reads the handle's own `boundingBox()` at runtime and clicks inside it, because the handle is a 10px band on the header's bottom edge whose height is theme-dependent, and a centred header click would miss it and pass unfixed.

**Red-to-green evidence.** With both plugin source files reverted to `origin/develop` and the bundle rebuilt, **6 of the 11 tests fail**: the four detach cases, the swallowed click, and the plugin-level destroy. The other five are pins, and each says so in the spec where it could otherwise be mistaken for coverage it does not provide:

- `leaves nothing in the container after the grid is destroyed` cannot fail — `Core#destroy()` runs `empty(rootContainer)` *before* it iterates the plugins, so the container is already clean by the time `plugin.destroy()` is reached. What it guards is that ordering. Observing the `destroy()` override at all needs `detaches the handle when the plugin itself is destroyed`, which destroys the plugin alone with the grid left alive; reverting just the two `destroy()` overrides reds exactly that test and nothing else.
- `keeps the row handle attached through a double-click autoresize` reaches the 500ms boundary through the only observable that cannot be true before it. The timeout resets the click count, and until it does `shouldSkipResizeHandlePositioning()` makes every `mouseover` early-return, so the handle cannot move off its row. Polling the handle count would have resolved on its first evaluation and observed nothing, and the autoresize is no use either — it sets the row to `max(startHeight, renderedHeight)`, which single-line content leaves unchanged.

```bash

# new spec + the #13319 spec it now shares a base page object with, all six theme x bundle legs
cd tests && npx playwright test manual-resize-teardown manual-resize-contextmenu   # 102/102

# legacy Jasmine, both plugins, on each theme
npm --prefix handsontable run test:e2e -- --testPathPattern='manual(Row|Column)Resize'                    # 147 specs, 0 failures
npm --prefix handsontable run test:e2e -- --testPathPattern='manual(Row|Column)Resize' --theme=classic    # 147 specs, 0 failures
npm --prefix handsontable run test:e2e -- --testPathPattern='manual(Row|Column)Resize' --theme=horizon    # 147 specs, 0 failures

# cross-cutting legacy sweep
npm --prefix handsontable run test:e2e -- --testPathPattern='manual(Row|Column)Resize|nestedHeaders|autoRowSize|autoColumnSize|core/destroy'   # 485 specs, 0 failures

npm --prefix handsontable run test:unit      # 4311 passed
npm --prefix handsontable run test:types     # pass
npm --prefix handsontable run eslint         # 0 errors
npm --prefix handsontable run stylelint      # pass

# the change adds a destroy() code path, and the jsdom observer stubs are duplicated per package
npm --prefix wrappers/react-wrapper run test     # 99 passed
npm --prefix wrappers/vue3 run test              # 39 passed
npm --prefix wrappers/angular-wrapper run test   # 299 passed
```

### Types of changes

### Related issue(s):

1. DEV-2719

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2719

---

> [!NOTE]
> **Low Risk**
> Localized DOM lifecycle fix in the two manual resize plugins with explicit guards for autoresize timing and in-flight drags; behavior change is limited to plugin disable/destroy and is covered by new e2e tests.
> 
> **Overview**
> Fixes **orphaned manual resize handles** left in the grid container after `manualRowResize` / `manualColumnResize` are turned off, which could still capture pointer events and block header clicks.
> 
> Both plugins now share **`#detachHandleAndGuide()`**, invoked from **`disablePlugin()`**, **`destroy()`**, and the existing context-menu path (replacing inline `remove()` calls). **`hideHandleAndGuide()`** is unchanged so completed drags and double-click autoresize do not flicker the handle. **`afterMouseDownTimeout()`** exits early when the plugin is already disabled so a pending 500ms autoresize timer cannot re-append the handle or touch an unregistered size map.
> 
> Adds Playwright coverage (`manual-resize-teardown.spec.ts`), a teardown fixture, and a shared **`ManualResizePage`** base used by the context-menu page object. **`AGENTS.md`** documents lazy attach/teardown pitfalls for future edits.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4a846ef40a7ca502740682cd8bafb5797a85c78f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2719 Handsontable core: manual resize handle and guide are never detached on teardown](https://app.clickup.com/t/9015210959/DEV-2719)

handsontable/src/plugins/manualColumnResize/manualColumnResize.ts
```diff
@@ -253,6 +253,8 @@ export class ManualColumnResize extends BasePlugin {
    * Disables the plugin functionality for this Handsontable instance.
    */
   disablePlugin() {
+    this.#detachHandleAndGuide();
```
budnix, line 256: The detach here doesn't cover the 500 ms `#autoresizeTimeout` that a double-click leaves pending. `hot._registerTimeout` is only cleared in `Core#destroy()`, so if an app calls `updateSettings({ manualColumnResize: false })` inside that window the timer still fires, and `afterMouseDownTimeout()` ends with `setupHandlePosition()`, which appends the handle straight back into the container - the exact orphan this PR removes. A `this.enabled` guard inside `afterMouseDownTimeout()` would close it, as long as it still resets `#autoresizeTimeout` and `#dblclick` on the way out, since `#onMouseDown` only re-registers the timer while `#autoresizeTimeout` is null. Same gap in `manualRowResize.ts`.

handsontable/src/plugins/manualColumnResize/manualColumnResize.ts
```diff
@@ -548,6 +550,29 @@ export class ManualColumnResize extends BasePlugin {
...
+   * Detaches the resize handle and the resize guide from the root element and clears their active
+   * state. Shared by the context menu handler, `disablePlugin()` and `destroy()`, so a plugin that
+   * is turned off leaves nothing of its own in the container.
+   *
+   * Both elements are detached with `remove()`, which is a no-op on an element that has no
+   * parent. The guide is attached only once a "mousedown" over the handle reaches
+   * `#onMouseDown`, so a context menu opened over a merely hovered handle reaches a guide that
+   * was never attached, and `removeChild` threw there (DEV-2708).
+   *
+   * The pressed flag is deliberately NOT reset here. `updatePlugin()` runs
```
budnix: This holds for the `updatePlugin()` cycle, but not for a real disable: `super.disablePlugin()` clears the events, so the `mouseup` never arrives and `#pressed` stays latched true. After a later re-enable, `#onMouseMove` then reads plain pointer movement as a drag and writes sizes from a stale `startX`. That part is pre-existing, but the docblock now reads as if the flag is safe either way - an `event.buttons === 0` check in `#onMouseMove` would cover both cases. Curious what you think.

handsontable/src/plugins/manualColumnResize/manualColumnResize.ts
```diff
@@ -548,6 +550,29 @@ export class ManualColumnResize extends BasePlugin {
...
+   * `disablePlugin(); enablePlugin();` on any `updateSettings()` carrying the plugin's own key,
+   * which is what a framework wrapper sends on every re-render - clearing the flag there would
+   * make the "mouseup" that ends an in-flight drag take the idle branch, so the drag would be
+   * dropped with no `afterRowResize`/`afterColumnResize` and the dragged size never confirmed.
+   * The context menu handler resets it at its own call site, where aborting the drag is the point.
+   */
+  #detachHandleAndGuide() {
+    this.hideHandleAndGuide();
+    this.#handle.remove();
```
budnix, line 584: Because `#pressed` stays true on purpose, a drag that is in flight when `updatePlugin()` fires now loses both elements from the DOM. `enablePlugin()` doesn't put them back, and `#onMouseOver` early-returns while `#pressed` is true, so the user keeps dragging with no handle and no guide until mouseup calls `setupHandlePosition()` again. The resize itself still lands, so this is visual only - worth a quick check that it's an accepted trade for wrapper re-renders.

handsontable/src/plugins/manualRowResize/manualRowResize.ts
```diff
@@ -532,6 +534,29 @@ export class ManualRowResize extends BasePlugin {
     removeClass(this.#guide, 'active');
   }
 
+  /**
```
budnix, line 537: Everything this fix uncovered lives only in this docblock, copied into the two plugin files. Develop now carries `handsontable/src/plugins/manualRowResize/AGENTS.md` and its column twin (added in #13335, which this branch sits 7 commits behind), and the repo rule puts a trap found while changing a plugin in that file. Worth adding the three points after the rebase: `hideHandleAndGuide()` never detached anything, an orphaned handle swallows the header click underneath it, and why `#pressed` has to survive the update cycle.

tests/fixtures/pages/ManualResizePage.ts
```diff
@@ -0,0 +1,156 @@
...
+   */
+  async parkPointer(): Promise<void> {
+    await this.page.mouse.move(0, 0);
+  }
+
+  /**
+   * Navigate and wait for the fixture's grid to have rendered - a real DOM condition, never a
+   * sleep.
+   */
+  abstract goto(): Promise<void>;
```
budnix, line 170: The base class exists so a gesture fix can't land in one page object and be missed by the other, but `goto()` stayed abstract - and the new bundle wait only ended up in `ManualResizeTeardownPage`. `ManualResizeContextMenuPage.goto()` still goes straight to `expect(...ht_clone_top).toBeVisible()`, so on a cold server it can time out on visibility for a reason that has nothing to do with the clone. An `awaitBundle()` step here that both `goto()`s call would keep the two in step.

tests/e2e/manual-resize-teardown.spec.ts
```diff
@@ -0,0 +1,213 @@
...
+    await expect(grid.columnGuide).toHaveCount(1);
+    await expect(grid.columnGuide).toBeHidden();
+  });
+
+  test('keeps resizing after the plugin is turned off and on again', async () => {
+    const startHeight = await grid.renderedRowHeight(2);
+
+    await grid.parkPointer();
+    await grid.setResizeOption('manualRowResize', false);
+    await expect(grid.rowHandle).toHaveCount(0);
```
budnix, line 164: This test never reveals the handle, so `toHaveCount(0)` is already true from page load and stays green with the fix reverted. The four detach tests above set the premise up first - worth doing the same here so the line actually watches the teardown.
```suggestion
    await grid.hoverRowHeader(2);
    await expect(grid.rowHandle).toHaveCount(1);

    await grid.parkPointer();
    await grid.setResizeOption('manualRowResize', false);
    await expect(grid.rowHandle).toHaveCount(0);
```

tests/e2e/manual-resize-teardown.spec.ts
```diff
@@ -0,0 +1,213 @@
...
+
+      const box = await grid.rowHandle.boundingBox();
+
+      return box !== null && Math.abs(box.y - boxBefore!.y) > 2;
+    }, { timeout: 5000 }).toBe(true);
+
+    await expect(grid.rowHandle).toHaveCount(1);
+  });
+
+  test('detaches the handle when the plugin itself is destroyed', async () => {
```
budnix, line 208: This test and the swallowed-click one above only cover `manualRowResize`. The two plugins are edited in lockstep, so right now the column `destroy()` override and its orphan-swallows-click path have no assertion at all. Would be good to mirror at least the destroy case for `manualColumnResize`.

handsontable/src/plugins/manualColumnResize/manualColumnResize.ts
demtario: Confirmed, and it turned out worse than the read: on its way to `setupHandlePosition()` the pending callback also runs the resize hooks and writes through `setManualSize()` into a size map that `disablePlugin()` has already unregistered.

Fixed the way you suggested. `afterMouseDownTimeout()` now opens with an `if (!this.enabled)` bail that still resets `#autoresizeTimeout` and `#dblclick`, so `#onMouseDown` can arm a fresh timer after a re-enable.

Covered by two new tests, one per plugin, in their own describe because the clock has to be installed before the page loads. They cross the 500ms boundary with `page.clock.runFor()` rather than a sleep: `toHaveCount(0)` resolves the instant it is true, so simply giving it a longer timeout would have observed nothing. Both verified red with the bail removed and the bundle rebuilt.

handsontable/src/plugins/manualColumnResize/manualColumnResize.ts
demtario: You are right that the docblock overstated it. Rewritten to say plainly that keeping the flag is a trade rather than a safe default, with both known costs named - the latch you describe here, and the in-flight drag from your next comment.

On the `event.buttons === 0` check, I tried it rather than reasoned about it, because it is the fix I would want too. It reds **41 of the 147 specs** across the two plugin suites. The frozen Jasmine helpers simulate `mousemove` with only `clientX`/`clientY` - `resizeRow`/`resizeColumn` in `test/helpers/common.js`, plus dozens of inline `simulate('mousemove', ...)` calls in both plugin specs - so `event.buttons` is 0 on every synthetic drag and the guard aborts all of them. Those helpers are shared with the nestedHeaders and autoRowSize specs, so making it work means a sweep well past this PR's blast radius.

Left out here and recorded in `manualColumnResize/AGENTS.md` with the measurement, so the next person does not rediscover the 41. Happy to raise it as a follow-up task if you agree that is the right split.

handsontable/src/plugins/manualColumnResize/manualColumnResize.ts
demtario: Confirmed, and yes - accepted trade. The resize still lands, so it is visual only, and the alternative (skipping the detach while `#pressed`) reopens the orphan this PR exists to close, on exactly the path a real disable takes.

The actual defect was that the diff implied it was free rather than stating it. The `#detachHandleAndGuide()` docblock now names this alongside the latch.

handsontable/src/plugins/manualRowResize/manualRowResize.ts
demtario: Done. The three points, plus the timeout bail from your first comment, are now in `manualColumnResize/AGENTS.md` - that is where the row file says the shared behaviour is documented - and the row file gained a pointer naming them so nobody reads only that half.

One deviation: I merged develop rather than rebasing. The branch is bound to an open PR and the repo forbids force-pushing those, so a rebase was not available. The merge was clean and brought in the per-plugin AGENTS.md files from #13335.

tests/fixtures/pages/ManualResizePage.ts
demtario: Good catch, and it was a deliberate omission that turned out to be the wrong call. I had stripped the wait out of `ManualResizeContextMenuPage.goto()` to keep the base-class extraction behaviour-neutral, which left exactly the asymmetry the base class exists to prevent.

`awaitBundle()` now lives on `ManualResizePage`, with the reasoning in its docblock, and both `goto()`s open with it.

tests/e2e/manual-resize-teardown.spec.ts
demtario: Confirmed - the handle was never revealed, so that line restated a fact that was already true at page load. Applied your suggestion.

tests/e2e/manual-resize-teardown.spec.ts
demtario: Done, and mirrored the swallowed-click case as well as destroy - you are right that both column paths had no assertion at all.

The column click point is biased like the row one: the handle sits at `headerStart - 6 + headerWidth`, so its centre is a pixel inside the column and rounding can put it on the next one.

budnix (approved):

demtario:
cursor review

PR #13348, opened 2026-09-02
DEV-2766: Flag matcher downgrades and widened tolerances in the weakening gate

danielzytohoc:
### Context

The PR adds matcher-aware detections to the test-weakening gate, which until now only compared assertion **counts** per file. Counting misses two real shapes the enforcement-tightening spec calls out: a PR that replaced `toHaveBeenCalledTimes(300)` with `toBeGreaterThanOrEqual(300)` while **adding** a third assertion (count rose, gate silent), and a PR that deleted a committed-value `toBe` while adding other assertions (count flat, gate silent). The review round added a third shape: deleting a whole failing `it()` while growing a survivor (count rises, histogram flat, gate silent).

`.github/scripts/lib/test-weakening.mjs` now exports:

- `EXACT_MATCHERS` — the Jest/Jasmine value pins (`toBe`, `toEqual`, `toStrictEqual`, `toHaveBeenCalledTimes`, `toHaveLength`, `toHaveBeenCalledWith`, `toHaveBeenLastCalledWith`) plus the Playwright locator/page assertions that pin a value (`toHaveText`, `toHaveValue`, `toHaveCount`, `toHaveAttribute`, `toHaveClass`, `toHaveCSS`, `toHaveId`, `toHaveJSProperty`, `toHaveURL`, `toHaveTitle`) — `tests/e2e/*.spec.ts` is the tier that grows, and it holds ~500 Playwright matcher calls that both tables used to miss.
- `BOUNDED_MATCHERS` (`toBeGreaterThanOrEqual`, `toBeLessThanOrEqual`, `toBeGreaterThan`, `toBeLessThan`, `toBeCloseTo`, `toBeTruthy`, `toBeFalsy`, `toBeDefined`, `toContain`, `toContainEqual`, `toMatch`, `toMatchObject`, `toHaveProperty`, `toHaveBeenCalled`, `toContainText`, `toContainClass`) — `toHaveBeenCalled` is the loose form of `toHaveBeenCalledTimes`; the two Playwright entries are the substring and class-subset checks.
- `THROW_MATCHERS` (`toThrow`, `toThrowError`, `toThrowWithCause`) — exact with an argument, bounded when bare; the bare call is its own histogram label (`toThrow()`), so `toThrowError('msg')` → `toThrow()` diffs as a downgrade. Negation flips the pair: a bare `not.toThrow()` pins the one outcome "does not throw" and is **exact** (179 bare calls in `handsontable/` — 172 `not.toThrow()` and 7 `not.toThrowError()` — none with an argument), while `not.toThrow('msg')` rules one error out and is bounded.
- `NEGATION_PINS` (`toHaveBeenCalled`, `toBeDefined`) — a negated call (`.not.toBe(0)`) is its own label (`not.toBe`) and counts as **bounded** (it rules one value out), except these two, whose negation pins a single value (`not.toHaveBeenCalled()` is zero calls, `not.toBeDefined()` is `undefined`), and the bare throw negation above, whose rule depends on the argument and lives in `matcherKind`. So `toBe(5)` → `not.toBe(0)` is a finding, and `toHaveBeenCalledTimes(0)` → `not.toHaveBeenCalled()` (466 vs 100 uses in `handsontable/`) is not. Pinning those negations as exact has a price, recorded in the blind spots and pinned by a test: an exact → exact swap such as `expect(r).toBe(5)` → `expect(() => f()).not.toThrow()` is invisible to `matcher-downgrade`.
- `matcherKind(label)` — the single classifier (`'exact'` | `'bounded'` | `null`) behind the detectors and the evals scorer. Playwright's state-only assertions (`toBeVisible`, `toBeHidden`, `toBeAttached`, `toBeEnabled`, `toBeChecked`, …) are in no table: they assert a state, not a value, so they are neither — a documented blind spot.
- `matcherHistogram(src)` — per-label call counts; `toBe(` does not swallow `toBeGreaterThan(`, `toContain(` does not swallow `toContainText(`, multi-line chains count.
- `countTestBlocks(src)`, `countTableRows(src, start)`, and `TEST_CALL_RE` — the `it()`/`test()` opener regex moved here from `evals/score.mjs` (whose `extractTestBlocks` now imports it), so the detector and the scorer count the same blocks. The regex also matches the parameterized openers (`it.each(`, `test.each(`, modifier chains such as `test.skip.each(`) and exposes an `each` group; a table counts one block per row of its array literal — the tests the runner registers — so folding two `it()` into a two-row table is quiet and deleting a row is not. A table the scanner cannot read as an array literal (a variable, a `.map()` call) counts as one; a tagged-template table (`it.each\`…\``) is not counted (no spec uses that form). The repo has 23 `it.each` tables across 10 unit specs; before this, they counted zero blocks. `TEST_CALL_RE` and `countSkipFocus` are built from one opener grammar (`openerRe(names, { memberAccess })`), so a focused or skipped opener the block count accepts is always one marker — `xit.each(`, `fit.each(`, `test.concurrent.only(`, `test.concurrent.skip(`, `test.concurrent.only.each(` included; the marker counter widens the grammar to suites and allows a member access because Playwright spells a focused suite `test.describe.only(`.
- `detectMatcherDowngrade(before, after)` — pure; `{ kind: 'matcher-downgrade', exactDrops, boundedRises }` only when at least one exact label **lost** calls, at least one bounded label **gained** calls, **and the total exact count did not rise**. Either half alone is not a finding (a plain removal is already `assertions-removed`; adding a relational assertion is the documented pattern for values no token derives). The totals rule keeps a rename toward a *more* exact matcher quiet — the #13266 calibration hit renamed 42 `toBeGreaterThan` to `toBe` beside 11 `toEqual` to `toContain` and read as a downgrade when the labels were compared one by one; its price (a downgrade beside a larger addition of exact matchers hides behind the rising total) is documented in the blind spots.
- `detectPrecisionWidening(before, after)` — pure; `{ kind: 'precision-widened', widenings: [{ from, to }] }` when a `toBeCloseTo(x, digits)` argument went down. An omitted argument counts as the framework default of 2; the argument scanner skips string literals **and line/block comments** (a `// relaxed` between the arguments used to drop a real widening, and an apostrophe in it opened a "string" that ran to end of file); a non-literal argument is never judged, and neither is a call whose argument list holds a regex literal in argument or operand position — at the start of an argument, after an operator, an opening bracket, or a keyword such as `return` — because a misread bracket used to truncate the list and turn a scan failure into an invented `5 → 2` finding. A `/` right after a closing `)` is ambiguous and reads as a division.
- `formatFinding(finding)` — the one-line renderer the gate uses for console and `$GITHUB_STEP_SUMMARY`, extracted so the output is unit-tested.

`detectWeakening()` now reports **five** kinds: `assertions-removed`, `tests-removed` (new — the block count dropped while the file still exists), `skip-or-focus-added`, `matcher-downgrade`, `precision-widened`. Every finding carries a `kind` discriminator (the two pre-existing kinds were keyed `type`; the only consumer is the gate renderer, and the tests were updated). The gate stays **warn-only** (exit 0) — this is reviewer signal.

Scorer: `evals/score.mjs` classifies through `matcherKind` (so a negated-only spec no longer suppresses the warning, and a spec whose only matcher is `not.toThrow()` no longer raises it), `countMatchers` returns the histogram it already built (one regex pass instead of two), and `toBeCloseTo` no longer trips `loose-matchers-only` — pinning a float to ten decimal places is not loose; the detector's `precision-widened` owns that loosening. `extractTestBlocks` skips an `it.each` table and reads the title and body from the second call (every table used to come back untitled and hollow), carries `rows`, and the score's `tests` sums them, so it agrees with `countTestBlocks`. The `matchers: { exact, bounded }` field and the warning-only signal are otherwise unchanged; the three fixture references still score `meaningful` with no warning.

Docs: `.ai/LOCAL-ENFORCEMENT.md` describes the five finding kinds, the tables, the negation and throw rules, the parameterized-table rule, the blind spots, and the replay numbers below; `evals/README.md` documents the `matchers` field and the row-aware `tests` count.

**Calibration on real history** (read-only replay of the detector over the last 300 first-parent commits of `develop` at `69dab641b`, 481 spec files compared; the replay mirrors the gate's diff logic): `assertions-removed` 21, `skip-or-focus-added` 2, **`matcher-downgrade` 1, `precision-widened` 0, `tests-removed` 16**. The one downgrade hit is #13242 (`afterRefreshDimensions.spec.js`: `toHaveBeenCalledTimes` 2 → 1, `toBeGreaterThanOrEqual` 0 → 1 — the exact shape from the spec). Before the totals rule the kind fired twice; the second hit, #13266 (`cellAlignment.spec.js`: `toEqual` 11 → 0, `toContain` 0 → 27, but `toBeGreaterThan` 42 → 0 and `toBe` 0 → 42 in the same file — 74 → 77 exact overall), was a false positive, so the honest figure for the matcher kinds is **one for one** on replayed history, not two for two. Of the 16 `tests-removed` hits, 15 sit on spec migrations to Playwright and the skipped-test burn-down, and one (#13244, `templateLiteralTag.unit.ts`, 6 → 5) is a unit test deleted together with the helper it covered — all legitimate deletions; 14 of them are on files `assertions-removed` already flagged, 2 (the `copy`/`cut` specs, dead skips removed) are reached only by the new kind. The fourth commit (row-aware `it.each` counting, the `not.toThrow()` flip, the keyword regex start) leaves the replay unchanged: same 481 files, same five totals, same hits. One matcher finding in 300 PRs is the intended noise level for a warn-tier gate; a `tests-removed` finding mostly says "a test was deleted here", which the reviewer then judges.

**Review fixups (second commit):**

- `splitTopLevelArgs` had a dead escape branch: `i += ch === '\\' ? 2 : 1` compared the *opening quote* to a backslash, so an escaped quote inside the first `toBeCloseTo` argument (`f('a\'b')`) ended the literal early, the stray quote opened a second "string" that ran to end of file, the list never closed, and the widening was silently dropped (a false negative). Now `src[i] === '\\'`. Regression test added, red before the fix (`null` instead of the `5 → 2` widening).
- The `pre-push.mjs` step-3 comment now names all finding kinds.
- The lib's module JSDoc carries a **Known blind spots** section (mirrored on the affected functions).

**Review fixups (third commit)** — one commit, since the scorer imports the new exports:

- Playwright matchers in both tables; the throw family; `.not.` as its own label; `matcherKind` as the shared classifier; `NEGATION_PINS`.
- The totals rule in `detectMatcherDowngrade` (total exact count must not rise).
- Comment skipping in the argument scanner; a regex literal in a `toBeCloseTo` argument list makes the call unjudgeable (never a default).
- The fifth kind, `tests-removed`, with `countTestBlocks` / `TEST_CALL_RE` shared with the scorer.
- Scorer: `toBeCloseTo` exempt from `loose-matchers-only`; `countMatchers` returns its histogram.
- Docs and the gate/pre-push wording updated; the calibration replay re-run and the honest numbers recorded.

**Review fixups (fourth commit)** — the Cursor Bugbot thread and the audit of the third commit:

- `TEST_CALL_RE` matches `it.each` / `test.each` and the modifier chains; `countTestBlocks` counts a table by its rows through the new `countTableRows` (a table it cannot read counts as one); `countSkipFocus` reads `it.only.each(` as a marker. Before: `it.each` matched nothing, so deleting a table was invisible to `tests-removed` and folding two `it()` into one table read as test blocks 2 → 0.
- Scorer: `extractTestBlocks` skips the table, reads the title and body from the second call, and reports `rows`; `tests` sums them.
- `matcherKind`: a bare negated throw matcher is exact (`not.toThrow()`), a negated throw matcher with an argument is bounded — the bare negation classified bounded, a scorer regression against the pre-PR state, where `toThrow` was in no table.
- `REGEX_START_RE` recognises a regex literal after a keyword (`return`, `typeof`, `case`, …), so a `toBeCloseTo` holding `return /\)/` is not judged; the promise is worded as "argument or operand position".
- Docs: the `tests-removed` replay sentence corrected to 15 + 1 (see above); the module blind spots, `LOCAL-ENFORCEMENT.md`, and `evals/README.md` carry the parameterized-table rule.

**Review fixups (fifth commit)** — the second Cursor Bugbot thread and the audit of the fourth commit:

- `countSkipFocus` had its own regex (`.each` only after `.skip`/`.only`, no modifier chain, no prefix before `.each`), while `TEST_CALL_RE` accepted `[xf](it|test).each` and `(?:\.MOD)*` chains — so `xit.each(`, `fit.each(`, `test.concurrent.only(`, `test.concurrent.skip(`, and `test.concurrent.only.each(` each counted one test block and zero markers, and a new spec born with any of them raised neither `skip-or-focus-added` nor the scorer's `skip-or-focus` signal. Both regexes now come from one builder, `openerRe(names, { memberAccess })`, which exposes `prefix`, `chain`, and `each` groups: `TEST_CALL_RE = openerRe('it|test')`; the marker counter reads `openerRe('it|test|describe|context', { memberAccess: true })` and counts an opener whose `prefix` is set or whose `chain` holds `skip`/`only`. A parity test walks the grammar (372 openers: both names, both prefixes, every modifier chain up to two deep, with and without `.each`) through both counters; the five probes are pinned as `skip-or-focus-added` findings on a new spec and as the scorer's `skip-or-focus` signal (the scorer already shares `countSkipFocus`, so it has no regex of its own to mirror).
- Docs: the `not.toThrow()` figure corrected in `LOCAL-ENFORCEMENT.md`, the test comment, and this description — 179 bare calls in `handsontable/` (172 `not.toThrow()`, 7 `not.toThrowError()`), none with an argument; the earlier 220 also counted `wrappers/`. The module header now names both comment-skipping scanners (the `toBeCloseTo` argument reader and the `it.each` row counter, both on `splitTopLevelArgs`) and states the asymmetry: a commented-out `it(` still counts as a live block, a commented-out table row is not counted (2 → 1), a table commented out whole counts as one — pinned by a test. `LOCAL-ENFORCEMENT.md` carries the "`assertions-removed` is not row-aware" clause and what `skip-or-focus-added` counts; the blind-spots list (header and `LOCAL-ENFORCEMENT.md`) records the negation trade-off above.

No changelog entry: tooling/test-infrastructure only (no `handsontable/src/**` or `wrappers/**` change), so the path-aware changelog gate passes on its own.

Observation, out of scope: the root `eslint .github/scripts/ …` script lints exactly one file under `.github/scripts/` — ESLint's default dot-directory ignore swallows the rest — and `evals/` is in no lint target either. The touched files were linted explicitly (see below); the pre-existing `max-len`/`brace-style`/`arrow-parens` debt on the `evals/` lines adjacent to this change was cleared in passing.

### How has this been tested?

Test-first: the new tests were written before the implementation and failed with `does not provide an export named 'BOUNDED_MATCHERS'` (detector) and `matchers` undefined / missing warning (scorer, 3 failing), then passed after implementation. The review-fixup regression test was likewise red first (`detectPrecisionWidening` returned `null` for the escaped-quote input) and green after the one-character fix. For the third commit, every reviewer probe was first reproduced against the committed detector with a scratch script (all eight held: `toHaveText` → `toContainText`, `toHaveBeenCalledTimes` → `toHaveBeenCalled`, `toThrowError('msg')` → `toThrow()`, and `toBe(5)` → `not.toBe(0)` all silent; the comment-with-bracket case invented a `5 → 2` widening; the comment-between-arguments cases dropped a real one; the deleted-`it()` case came back clean), then pinned by a test that passes on the new code. For the fourth commit the same probe script confirmed each claim against the third commit first (`countTestBlocks("it.each([1, 2])('n %s', fn)")` → 0, two `it()` → one `it.each` → `tests-removed 2 → 0`, a table deleted beside a growing `it()` → no finding, the real `classify.unit.js` → 18 with its two tables uncounted, `matcherKind('not.toThrow()')` → `bounded` with `loose-matchers-only` raised, the `return /\)/` shape → an invented `5 → 2`), then flipped after the change (2, `assertions-removed` only, `tests-removed 3 → 1`, 25, `exact` with no warning, `null`). For the fifth commit the probe script confirmed the five openers against the fourth commit first (each: blocks 1, markers 0, `detectWeakening('', src)` → `[]`; scorer `gamingSignals` → `[]`), then flipped after the change (markers 1, `skip-or-focus-added` 0 → 1, `skip-or-focus` 5); the new detector test file was also run against the committed lib and was red (3 of 40 failing — the extended `countSkipFocus` case, the parity walk, and the five-probe findings), while the comment-asymmetry and negation-trade-off cases pass on both revisions, as pins of documented behavior.

- Unit tests added/modified: `.github/scripts/__tests__/test-weakening.test.mjs` (40 cases — the two original downgrade shapes, pure addition → none, helper rename → none, exact → exact → none, bounded-only addition → none, exact drop without bounded rise → left to the count detector, Playwright `toHaveText` → `toContainText`, `toHaveBeenCalledTimes` → `toHaveBeenCalled`, `toThrowError('msg')`/`toThrow('msg')` → `toThrow()` and the tightening direction → none, `not.toThrow()` → `not.toThrow('boom')` flagged, `toBe(5)` → `not.toBe(0)` flagged while `toHaveBeenCalledTimes(0)` → `not.toHaveBeenCalled()` and `toBe(undefined)` → `not.toBeDefined()` are not, the totals rule on the #13266 shape and its documented price, `toBeCloseTo` 5 → 2 flagged, tightening / unchanged / non-literal → none, omitted digits = 2, escaped quotes, a line/block comment and an apostrophe between the arguments still read, a bracket in a comment or a regex literal — including one after `return` — never becomes a finding while a division and a `newValue / 2` are still judged, unterminated comments → null, `tests-removed` on the deleted-`it()` shape with count rising, `countTestBlocks` opener forms and look-alikes, a parameterized table counted by rows (primitive, object, nested, trailing-comma rows; modifier chains; `describe.each` as a suite; a variable, a call, an unterminated or empty table as one; the tagged-template blind spot), two `it()` folding into a table quiet on blocks, a deleted table and a deleted row flagged beside a growing survivor, block additions and new specs quiet, a commented-out row not counted while a commented-out `it()` is, the opener-grammar parity walk over both counters, the five prefixed-`.each` / deep-`skip`-`only` openers as `skip-or-focus-added` on a new spec, `toBe(5)` → `not.toThrow()` invisible (the documented price), `matcherHistogram` labels including `not.` and `toThrow()`, `matcherKind` for every rule including the negated throw pair, `formatFinding` for every kind), `evals/__tests__/score.test.mjs` (27 cases; new since the first commit: `countMatchers` returns its histogram and the score keeps the two totals, a negated-only spec raises `loose-matchers-only`, a `toBeCloseTo`-only spec and a `not.toThrow()`-only spec do not, an `it.each` table reads as one titled block with its own assertions and `rows`, `tests` agrees with `countTestBlocks`, a table with no call after it is hollow, the five prefixed-`.each` / deep-`skip`-`only` openers raise `skip-or-focus` 5 and `tests` 5).
- E2E tests: none — no runtime change.
- Type tests: n/a.

Commands run (all from the repo root):

```
node --test .github/scripts/__tests__/test-weakening.test.mjs   → # tests 40 / pass 40 / fail 0
node --test evals/__tests__/score.test.mjs                        → # tests 27 / pass 27 / fail 0
npm run test:tooling                                              → # tests 401 / pass 401 / fail 0
node evals/run-eval.mjs                                           → References: 3/3 meaningful — harness self-test PASSED.
npx eslint evals/__tests__/score.test.mjs evals/score.mjs scripts/pre-push.mjs → clean
npx eslint --no-ignore --rule '{"no-restricted-syntax":"off","import/extensions":"off","no-restricted-globals":"off","no-console":"off"}' \
  .github/scripts/lib/test-weakening.mjs .github/scripts/test-weakening-gate.mjs .github/scripts/__tests__/test-weakening.test.mjs
                                                                  → clean (root config with the same relaxations the scripts/** override grants)
env GATE_BASE=HEAD~30 node .github/scripts/test-weakening-gate.mjs → 6 files flagged on recent develop (assertions-removed and tests-removed on the spec migrations), 0 matcher/precision findings, exit 0
calibration replay (300 develop commits, before → after the third round)  → matcher-downgrade 2 → 1, precision-widened 0 → 0, tests-removed n/a → 16, assertions-removed 21 → 21, skip-or-focus-added 2 → 2
calibration replay (third → fourth commit)                        → identical: 481 files, 21 / 16 / 1 / 2 / 0, same hits
npm run eslint (root; covers .github/scripts)                      → clean
```

### Types of changes

Tooling / CI change — no product code.

### Related issue(s):

1. DEV-2766

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

None — repo tooling only (`.github/scripts/`, `scripts/`, `evals/`, `.ai/`).

### Checklist:

- [x] My change requires a change to the documentation. (`.ai/LOCAL-ENFORCEMENT.md`, `evals/README.md` — updated in this PR.)
- [ ] This change needs a manual QA pass — not needed; fully covered by the unit tests above.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2766

---

> [!NOTE]
> **Low Risk**
> Changes are limited to CI/tooling heuristics and eval scoring; the gate remains non-blocking (exit 0), so product runtime behavior is unaffected.
> 
> **Overview**
> Extends the **warn-only** diff-based test-weakening gate so it catches loosening that assertion counts alone miss, while keeping **Handsontable product code** untouched (tooling under `.github/scripts/`, `evals/`, docs, and hook comments).
> 
> **`detectWeakening`** now emits **five** finding kinds (findings use **`kind`** instead of **`type`**): existing assertion and skip/focus deltas, plus **`tests-removed`** (fewer `it`/`test` blocks, including row-aware **`it.each`** tables), **`matcher-downgrade`** (exact matchers drop while bounded matchers rise, with a total-exact-count guard and Playwright/Jest tables, negation, and throw handling), and **`precision-widened`** for relaxed **`toBeCloseTo`** digit arguments. Shared logic lives in **`test-weakening.mjs`** (`matcherHistogram`, `matcherKind`, `openerRe` / **`TEST_CALL_RE`**, **`formatFinding`**, argument scanning with comment/regex safeguards).
> 
> The **evals scorer** imports the same primitives: row-aligned **`tests`** counts, **`matchers: { exact, bounded }`**, and a warning-only **`loose-matchers-only`** signal. **`test-weakening-gate.mjs`** and **pre-push** surface all kinds via **`formatFinding`**; **`.ai/LOCAL-ENFORCEMENT.md`** and **`evals/README.md`** document behavior and blind spots. Coverage is a large expansion of **`test-weakening.test.mjs`** and scorer unit tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 54c000f801ad3b11557082d798eb733e43f0ae8d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2766 Weakening detector: flag matcher downgrades (exact → bounded) instead of counting assertions only](https://app.clickup.com/t/9015210959/DEV-2766)

.github/scripts/lib/test-weakening.mjs
```diff
@@ -4,17 +4,101 @@
...
+ */
+
+/**
+ * Matchers that pin a value, a count, or a call exactly. Trading one of these
+ * for a `BOUNDED_MATCHERS` entry keeps the assertion count but loosens what the
+ * test proves. Names only — the call paren is matched by `MATCHER_CALL_RE`.
+ *
+ * @type {ReadonlyArray<string>}
+ */
+export const EXACT_MATCHERS = Object.freeze([
```
budnix, line 97: Both tables only hold Jest and Jasmine matcher names, but `SPEC_RE` also picks up `tests/e2e/*.spec.ts`, where the Playwright matchers live. I counted 498 Playwright matcher calls across 69 specs there (`toBeVisible` 206, `toHaveCount` 136, `toHaveText` 76) and none are in either table, so `matcherHistogram` returns `{}` for every one of those files. I tried `toHaveText('A1')` becoming `toContainText('A')` and got no finding at all. Since new E2E has to be Playwright now, that is the tier that will grow — worth either adding those names or saying so in the blind spots list.

.github/scripts/lib/test-weakening.mjs
```diff
@@ -4,17 +4,101 @@
...
+
+/**
+ * Matchers that accept a range, a partial shape, or mere presence. Legitimate
+ * on their own (a relational assertion is the documented pattern for values no
+ * token derives — see `handsontable/.ai/TESTING.md`), suspicious only when one
+ * appears where an exact matcher disappeared.
+ *
+ * @type {ReadonlyArray<string>}
+ */
+export const BOUNDED_MATCHERS = Object.freeze([
```
budnix, line 127: `toHaveBeenCalled` is missing here, and it is the natural loose form of `toHaveBeenCalledTimes` — the repo has 594 bare `toHaveBeenCalled()` calls against 811 `toHaveBeenCalledTimes()`. Swapping one for the other stays completely silent: the assertion count is flat so `assertions-removed` says nothing, and with no bounded rise `detectMatcherDowngrade` returns null. The `toThrow` family has the same hole (`toThrow` 300 uses, `toThrowWithCause` 106, `toThrowError` 58), so `toThrowError('exact msg')` becoming `toThrow()` also reports nothing.

.github/scripts/lib/test-weakening.mjs
```diff
@@ -43,6 +127,224 @@ export function countSkipFocus(src) {
...
+ * @param {string} before The spec contents at the base revision.
+ * @param {string} after The spec contents at the head revision.
+ * @returns {{kind: 'matcher-downgrade', exactDrops: {matcher: string, from: number, to: number}[],
+ *   boundedRises: {matcher: string, from: number, to: number}[]}|null} The finding, or null.
+ */
+export function detectMatcherDowngrade(before, after) {
+  const histogramBefore = matcherHistogram(before);
+  const histogramAfter = matcherHistogram(after);
+  const exactDrops = histogramDeltas(
+    EXACT_MATCHERS, histogramBefore, histogramAfter, delta => delta.to < delta.from,
```
budnix: This filter runs per matcher name, so a plain rename from one exact matcher to another reads as a drop even when the file ends up more exact than before. The PR's own calibration hit shows it: in `3d8e32005b` (#13266, `cellAlignment.spec.js`) the total exact count went 74 to 77 and the total bounded count went 42 to 27, because 42 `toBeGreaterThan` calls became `toBe` — and the gate still reports a matcher downgrade on `toEqual` 11 to 0 against `toContain` 0 to 27. `assertions-removed` already flags that file, so the new kind only added a wrong reason there. Requiring the total exact count not to rise would drop that one without losing the two shapes your tests cover.

.github/scripts/lib/test-weakening.mjs
```diff
@@ -43,6 +127,224 @@ export function countSkipFocus(src) {
...
+
+  for (const match of src.matchAll(/\.\s*toBeCloseTo\s*\(/g)) {
+    const args = splitTopLevelArgs(src, match.index + match[0].length);
+
+    if (args === null) {
+      continue;
+    }
+
+    if (args.length === 1) {
+      digits.push(CLOSE_TO_DEFAULT_DIGITS);
```
budnix, line 662: When `splitTopLevelArgs` stops early on a bracket it misread, the truncated list comes back with one entry, and this branch reads that as "digits omitted" and pushes the default 2. That turns a scan failure into an invented finding: `toBeCloseTo(y, 5)` becoming `toBeCloseTo(y /* keep )*/, 5)` gets reported as a widening from 5 to 2 digits, and `toBeCloseTo(f(/\)/), 5)` does the same. It is the one direction this module says it never goes — the docstring above promises an unbalanced list yields null so the caller treats it as no finding. Gating the default on the argument text having no comment or regex start would keep that promise.

.github/scripts/lib/test-weakening.mjs
```diff
@@ -43,6 +127,224 @@ export function countSkipFocus(src) {
...
+  }
+
+  return { kind: 'matcher-downgrade', exactDrops, boundedRises };
+}
+
+/**
+ * Split a call's argument list at its top-level commas, skipping string literals
+ * and nested brackets. A heuristic scanner (a regex literal holding an unbalanced
+ * bracket can confuse it), which is acceptable here: an unbalanced list yields
+ * null, and the caller treats "cannot judge" as "no finding".
```
budnix: The scanner skips string literals but not comments, so this promise does not hold for a fairly ordinary shape. A plain `toBeCloseTo(w, // relaxed` with the `2)` on the next line makes the second argument text ` // relaxed\n  2`, which fails the integer check, so a real 5 to 2 widening is dropped instead of reported. An apostrophe in that comment is worse — `// that's the width` opens a string that runs to the end of the file. This is the same class as the escaped-quote miss you fixed in the second commit, and that one got a regression test, so it feels worth the same treatment.

.github/scripts/lib/test-weakening.mjs
```diff
@@ -43,6 +127,224 @@ export function countSkipFocus(src) {
...
+ * @returns {Record<string, number>} Matcher name → number of calls.
+ */
+export function matcherHistogram(src) {
+  const histogram = {};
+
+  if (!src) {
+    return histogram;
+  }
+
+  for (const match of src.matchAll(MATCHER_CALL_RE)) {
```
budnix, line 488: Because the regex only needs a dot before the name, `.not.toBe(` lands in the histogram as `toBe`, so adding `.not.` is invisible to all four detectors. `expect(x).toBe(5)` becoming `expect(x).not.toBe(0)` gives no finding and no severity change, even though the second version proves almost nothing. The new test at line 93 of the spec file pins that behavior, so it reads as intended rather than as a gap. Counting a `.not.`-prefixed exact matcher as bounded would catch it, and the same fix would stop the scorer suppressing `loose-matchers-only` on a spec whose only exact matcher is negated.

.github/scripts/lib/test-weakening.mjs
```diff
@@ -83,14 +387,26 @@ export function detectWeakening(before, after, context = {}) {
...
+    findings.push({ kind: 'skip-or-focus-added', before: sBefore, after: sAfter });
+  }
+
+  const downgrade = detectMatcherDowngrade(before, after);
+
+  if (downgrade) {
+    findings.push(downgrade);
+  }
+
+  const widening = detectPrecisionWidening(before, after);
```
budnix, line 807: Nothing here looks at how many test blocks the file has, so deleting a whole failing `it()` while adding assertions to the survivor comes back clean. I tried two tests with one `toBe` each, then dropped the first and grew the second to three assertions: the count rose, `toBe` stayed flat, and the lone `toContain` rise is not a finding on its own, so the result was no findings and severity ok. That is the most direct way to reach green of all of them, and the new blind spots block does not mention it. `extractTestBlocks` in `evals/score.mjs` already counts blocks if you want a cheap fifth kind.

evals/score.mjs
```diff
@@ -440,9 +461,24 @@ export function scoreTestSource(src, options = {}) {
...
   const relevance = options.diff ? assessRelevance(src, options.diff) : null;
+  const assertions = countAssertions(src);
+  const matchers = countMatchers(src);
   const problems = [];
   const warnings = [];
 
+  // Warning-only: a relational assertion is the documented pattern for values no
+  // token derives. The bound-count must account for EVERY assertion, because a
+  // helper such as `grid.expectCell()` may pin a value the scorer cannot see.
+  if (matchers.exact === 0 && matchers.bounded > 0 && matchers.bounded >= assertions) {
```
budnix: `toBeCloseTo` counts as bounded, so a spec that pins a float to ten decimal places trips this and gets told nothing pins an exact value. I ran `expect(w).toBeCloseTo(123.4567890123, 10)` and it scored `{exact: 0, bounded: 1}` with the warning raised, which is the opposite of what that spec does. There are 167 `toBeCloseTo` calls in `handsontable/`, many in dimension and export specs where it is the only sensible matcher, so this will fire a fair bit. Leaving `toBeCloseTo` out of this one check and letting `precision-widened` own it would read better.

evals/score.mjs
```diff
@@ -440,9 +461,24 @@ export function scoreTestSource(src, options = {}) {
...
+    const histogram = matcherHistogram(src);
```
budnix: `countMatchers` on line 465 already built this exact histogram on line 443 and then threw it away, so this is a second full regex pass over the same file. Small per file, but the scorer runs over every fixture and candidate in the eval loop. Returning the histogram from `countMatchers` would save the pass and keep the two uses in step.

.ai/LOCAL-ENFORCEMENT.md
```diff
@@ -41,6 +41,31 @@ executable lines the unit tests cover (`.github/scripts/diff-coverage-gate.mjs`,
...
+`toBeLessThan`, `toBeCloseTo`, `toBeTruthy`, `toBeFalsy`, `toBeDefined`,
+`toContain`, `toContainEqual`, `toMatch`, `toMatchObject`, `toHaveProperty`),
+live in `.github/scripts/lib/test-weakening.mjs` and are the single source of
+truth for the evals scorer as well. Either half alone is not a finding: a plain
+removal is already `assertions-removed`, and adding a relational assertion is the
+documented pattern for values no token derives. **Precision widening** is a
+`toBeCloseTo(x, digits)` argument going down; an omitted argument counts as the
+framework default of 2, and an argument that is not an integer literal is never
+judged. Replayed over 300 `develop` commits (494 spec files) the two new kinds
+fired on 2 files, so a finding is worth a sentence in review, not a reflex.
```
budnix: Reading this, I take "fired on 2 files" to mean two real weakenings, but one of the two is a false positive. In #13266 the spec ended up more exact overall (exact matcher calls 74 to 77, bounded 42 to 27) and the flag comes from the per-name comparison I mentioned in `detectMatcherDowngrade`. So on the replayed history the new kinds are one for one rather than two for two, which is still fine for a warn tier — worth saying though, so a reviewer does not trust a flag more than the data supports.

.github/scripts/lib/test-weakening.mjs
danielzytohoc: Done in 1e91dbcc1: EXACT_MATCHERS now carries the ten Playwright locator/page assertions that pin a value (toHaveText, toHaveValue, toHaveCount, toHaveAttribute, toHaveClass, toHaveCSS, toHaveId, toHaveJSProperty, toHaveURL, toHaveTitle) and BOUNDED_MATCHERS gains toContainText and toContainClass (.github/scripts/lib/test-weakening.mjs, the two tables near the top). Your probe now reports: toHaveText('A1') → toContainText('A') is a matcher-downgrade with toHaveText 1 → 0 beside toContainText 0 → 1, pinned by the 'flags a Playwright toHaveText → toContainText swap' test. The state-only assertions (toBeVisible, toBeHidden, …) stay outside both tables on purpose, and the module header plus LOCAL-ENFORCEMENT.md now name that as a blind spot, so a toHaveText → toBeVisible swap is documented as invisible rather than silently missed.

.github/scripts/lib/test-weakening.mjs
danielzytohoc: Both holes are closed in 1e91dbcc1. toHaveBeenCalled joined BOUNDED_MATCHERS as the loose form of toHaveBeenCalledTimes, and the throw family got its own THROW_MATCHERS table: a call with an argument counts as exact, a bare call is labelled toThrow() in the histogram and counts as bounded, so the two forms diff apart. Reproduced on the new code: toHaveBeenCalledTimes(3) → toHaveBeenCalled() reports toHaveBeenCalledTimes 1 → 0 beside toHaveBeenCalled 0 → 1, and toThrowError('exact msg') → toThrow() reports toThrowError 1 → 0 beside toThrow() 0 → 1, while the tightening direction stays quiet; all four shapes are in the 'flags toHaveBeenCalledTimes → toHaveBeenCalled and toThrowError(msg) → toThrow()' test. 559cb3303 then settled the negated pair: a bare not.toThrow() pins the one outcome "does not throw" and counts as exact (179 bare calls in handsontable/, none with an argument), while not.toThrow('msg') rules one error out and is bounded (matcherKind, with its own test rows).

.github/scripts/lib/test-weakening.mjs
danielzytohoc: Adopted in 1e91dbcc1: detectMatcherDowngrade now also requires the file's total exact count not to rise, so a rename toward a more exact matcher stays quiet. Replaying 3d8e32005b on cellAlignment.spec.js against the new lib gives exact 74 → 77, no matcher-downgrade, and only the assertions-removed 120 → 115 row that was already there; the 300-commit replay drops from two downgrade hits to the single real one (#13242). The price you would expect — a downgrade beside a larger addition of exact matchers hides behind the rising total — is pinned in the 'quiet when the file ends up MORE exact overall' test and written into the module header and LOCAL-ENFORCEMENT.md.

.github/scripts/lib/test-weakening.mjs
danielzytohoc: Fixed in 1e91dbcc1 on both fronts: splitTopLevelArgs now skips line and block comments, so the bracket inside /* keep )*/ never ends the scan, and closeToDigits refuses to judge any call whose argument list holds a regex literal (REGEX_START_RE — at the start of an argument, after an operator or opening bracket, and since 559cb3303 after a keyword such as return), so a misread list can no longer fall through to the default of 2. Your two probes now return null — toBeCloseTo(y /* keep )*/, 5) and toBeCloseTo(f(/\)/), 5) both report nothing — while a genuine toBeCloseTo(y, 5) → toBeCloseTo(y) still reports 5 → 2 and width / 2 is still read as a division; all of that is in the 'never turns a scan it cannot trust into a finding' test. One residual is written into the module header: a `/` right after a closing `)` (`if (x) /re/.test(s)`) is ambiguous with `(a) / 2` and reads as a division.

.github/scripts/lib/test-weakening.mjs
danielzytohoc: Given the same treatment as the escaped-quote miss in 1e91dbcc1: splitTopLevelArgs now skips line and block comments the way it skips strings, and an unterminated comment yields null. Your shapes now read correctly — toBeCloseTo(w, // relaxed followed by 2) reports the 5 → 2 widening, // that's the width no longer opens a string (a later toBeCloseTo in the same file is still judged), and /* was 5 */ 2 reports 5 → 2 as well. All three are pinned by the 'reads the digits argument past a comment between the arguments' test, with the unterminated variants in the sibling 'never turns a scan it cannot trust into a finding' test.

.github/scripts/lib/test-weakening.mjs
danielzytohoc: Done in 1e91dbcc1: a negated call is now its own histogram label (not.toBe) and matcherKind counts it as bounded, except the negations that pin a single outcome — not.toHaveBeenCalled() is zero calls and not.toBeDefined() is undefined (NEGATION_PINS), and since 559cb3303 a bare not.toThrow() is "does not throw". So toBe(5) → not.toBe(0) now reports toBe 1 → 0 beside not.toBe 0 → 1, toHaveBeenCalledTimes(0) → not.toHaveBeenCalled() stays quiet, and the scorer no longer lets a negated-only spec suppress loose-matchers-only — all three pinned in the new tests, and the histogram test you pointed at now asserts toBe and not.toBe as separate rows.

.github/scripts/lib/test-weakening.mjs
danielzytohoc: Added as the fifth kind in 1e91dbcc1: tests-removed fires when the it()/test() block count dropped while the file still exists. countTestBlocks lives in the lib and reuses TEST_CALL_RE, which moved out of evals/score.mjs so the detector and the scorer's extractTestBlocks count the same blocks (and, since 559cb3303, count an it.each table by its rows, so folding tests into a table is quiet and deleting a row is not). Your probe now returns a single tests-removed 2 → 1 finding at warn severity (flag when source also changed), pinned by the 'flags a deleted test block even when the assertion count ROSE' test. On the 300-commit replay the kind fires 16 times, 14 of them on files assertions-removed already flagged and two (the copy/cut specs) reached only by the new kind; the counts are recorded in LOCAL-ENFORCEMENT.md, and the module header lists the kind's blind spots — a test moved to another file reads as removed, and one deleted describe reads as every test it held.

evals/score.mjs
danielzytohoc: Agreed and done in 1e91dbcc1: the scorer now treats toBeCloseTo as pinning for the purpose of loose-matchers-only (it stays bounded in the detector's table, where precision-widened owns its loosening). Your probe expect(w).toBeCloseTo(123.4567890123, 10) now scores {exact: 0, bounded: 1} with no warning, while a toBeGreaterThan-only spec still warns; both are pinned in evals/__tests__/score.test.mjs and the matchers row in evals/README.md documents the exemption.

evals/score.mjs
danielzytohoc: Done in 1e91dbcc1: countMatchers returns the histogram it already built alongside the two totals (evals/score.mjs:477-490), and scoreTestSource destructures it, so the warning branch reads the per-label detail without a second regex pass — matcherHistogram is now called once in evals/score.mjs. The score object still carries only { exact, bounded }; the 'countMatchers returns the histogram behind its totals' test pins both the shape and the classification through matcherKind.

.ai/LOCAL-ENFORCEMENT.md
danielzytohoc: Rewritten in 1e91dbcc1 with the honest figures: the paragraph now says matcher-downgrade fired once on the 300-commit replay (#13242, a real loosening), precision-widened never, tests-removed 16 times, and that before the totals rule the downgrade kind fired twice with the second hit (#13266) being the rename toward a more exact matcher — a false positive — so a matcher finding is one for one on replayed history. I re-ran the replay independently at 69dab641b (481 spec files) with the lib at 244393a78 and got the same counts, and two downgrade hits with the pre-totals-rule lib, so the numbers in the doc match the data.

budnix (dismissed):

budnix (approved):

PR #13349, opened 2026-09-02
DEV-2764: Lint setTimeout and frame-count waits as fixed sleeps

danielzytohoc:
### Context

The PR changes the determinism lint so that the three enforcement points — the Playwright tier's `no-restricted-syntax` bans, the frozen tier's `handsontable/no-fixed-sleep-in-spec` rule, and the evals scorer — agree on what a "fixed wait" is. Until now a `setTimeout` was the one fixed delay none of them saw: the Playwright tier banned `waitForTimeout` but a timer inside `page.evaluate` passed lint, and the frozen tier's rule flagged `sleep()` while `setTimeout(fn, 100)` and the 1,300+ `waitForNextAnimationFrames()` calls (a delay measured in frames) went uncounted. The enforcement-tightening spec asked for all three to be closed in one PR, now that the frozen tier has a legal replacement: the `waitUntil(condition, timeout)` spec global in `handsontable/test/helpers/common.js`.

What changes:

- **Playwright tier (`tests/.eslintrc.cjs`), at `error`:** a new `no-restricted-syntax` entry bans the global timer — bare `setTimeout(`, `window.setTimeout(`, and `globalThis.setTimeout(` — so the usual disguise (a timer inside `page.evaluate`) is caught too. The member form is pinned to the global object on purpose: Playwright's `test.setTimeout(ms)` and `testInfo.setTimeout(ms)` set a budget, not a wait, and stay legal (a `callee.property.name` match alone flagged them — caught in review, fixed here). The ban reaches **page objects** as well as specs, and the message says so: the lint script covers `e2e` and `fixtures`, and a timer moved from a spec into the page object it drives is the same fixed wait one file further away (the `.html` fixtures that drive timers deliberately are not linted). The message spells the exception form — `// eslint-disable-next-line no-restricted-syntax -- <owning task id>: <why>`, the same convention `test.fixme` uses — and what qualifies: a 0 ms scheduling barrier for a negative assertion, never a duration. The one existing site, `macrotaskBarrier()` in `e2e/customBorders.spec.ts`, is a chain of three 0 ms macrotasks that lets a **negative** assertion prove no stale progressive-borders batch fires — `expect.poll` cannot prove a negative, so it keeps the barrier under the disable line. That line now names **DEV-2764**: the change that owns the spec (#13166) carries no ClickUp task id, so the id on the line is the task that introduced the ban and accepted this one exception to it, with a TODO for the fixture probe that will replace it. The rule was not lowered.
- **Frozen tier (`handsontable/no-fixed-sleep-in-spec`), still at `warn`:** two new shapes with their own message IDs — `noSetTimeout` for a global `setTimeout` (bare, `window.`, or `globalThis.` — a `setTimeout` method on any other object is that object's contract and is not judged) whose delay is a **non-zero** numeric literal, and `noFrameWait` for any `waitForNextAnimationFrames()` call **except a literal `0`**, which resolves at once (the helper's `totalFramesToWait === 0` branch) — the same zero-duration hand-off as `setTimeout(fn, 0)`, and 8 live sites. Every message names `waitUntil()` as the replacement. A computed delay (`setTimeout(fn, delay)`), a bare hand-off (`setTimeout(done)`), and a literal `0` (`setTimeout(done, 0)` — one macrotask, the same scheduling barrier the Playwright tier sanctions, not a timed wait) stay silent on purpose: the rule judges only what it can judge statically. The override that turns the rule on now names **`*.unit.ts`** alongside `*.unit.js` and `*.spec.js`: the 217 `*.unit.ts` files sat outside it (review found `src/helpers/__tests__/function.unit.ts` carrying eleven `sleep()` calls the rule never saw), and `handsontable/test/__tests__/frozenTierLintScope.unit.js` pins the override's reach in both directions through ESLint's own config resolution (on, at `warn`, for `.unit.ts` / `.unit.js` / `.spec.js` under `src/**/__tests__`, walkontable's `test/`, and SheetClip's `test/`; off for source files and `test/helpers/common.js`). `common.js` loses `waitForNameAnimationFrames`, a deprecated alias forwarding to `waitForNextAnimationFrames` that no spec used and neither the rule nor the scorer matched — the cheapest way around both gates is gone rather than special-cased.
- **RuleTester coverage** for the rule — the first for any custom rule in the repo — in `handsontable/.config/plugin/eslint/__tests__/`, run under `node --test` by pointing ESLint's `RuleTester` at `node:test`'s `describe`/`it`. The test imports `eslint`, so it runs through a new `handsontable/scripts/tasks.json` task exposed as `npm run test:eslint-rules` (a `package.json` shim), from CI's `Lint / core` job (`lint.yml`) — the job that already installs the core package and whose scope (`outside-ci`) covers a rule edit. It is deliberately **not** in the root `test:tooling` glob: that script backs the `Checks / tooling tests` job, which is checkout + Node only, and the first push had the test there, red on `ERR_MODULE_NOT_FOUND 'eslint'` for every PR. The task names the test file as a **literal path, no glob**: `node --test` exits 1 (`Could not find`) only when nothing in the list matched *and* no pattern is a glob — a quoted glob that matches nothing prints `tests 0` and exits 0, and a second literal file that still exists masks a renamed first one the same way — so a rename would have left `Lint / core` green with nothing run. `handsontable/test/__tests__/eslintRulesTask.unit.js` pins the list in both directions (every file in `__tests__` named, every named file present) and bans glob characters. A `handsontable/.eslintrc.js` override lets that ESM test import the CJS rule with its extension (`import/extensions` in the plugin's `pattern` + `ignorePackages` form — eslint-plugin-import compares per-extension values with `===`, so the array value the first push used was silently ignored); the root `.eslintrc.js` override for `scripts/**` and `evals/**` carried the same broken array form for years and enforced neither direction, so it is fixed in passing and now reports a missing extension in both trees (probe below).
- **Evals scorer (`evals/score.mjs`):** two new `determinismSmells` signals, `set-timeout` and `fixed-frame-wait`, still zero-dep, and they **mirror the rule's exemptions**, not just its bans: the scorer reads each call's arguments and flags only the global timer (bare, `window.`, `globalThis.`) with a non-zero numeric-literal delay, and a frame wait with anything but a literal `0` — so `test.setTimeout(30_000)`, `testInfo.setTimeout(ms)`, a `setTimeout` method on any other object, `setTimeout(fn, 0)` (what `macrotaskBarrier()` uses), and a computed `setTimeout(fn, delay)` all pass, as they do under lint. The signal list is exported as `DETERMINISM_SIGNALS`. `run-eval.mjs` gains a `counterexamples/` role: a fixture the scorer **must** mark `suspect`, which turns the harness self-test two-sided (references must pass the bar, counterexamples must fail it). A counterexample declares its smell in its file name — `<scenario>.<smell>.spec.ts` (contract in `evals/lib/counterexamples.mjs`) — and counts as caught only when that smell is the **only** smell and the only problem found: a bare `suspect` verdict is also reached by an empty file, a hollow test, or a `.skip` with zero determinism smells, and a stray `README.md` in the folder is now a fixture error instead of a "caught" counterexample. Five counterexamples — one per signal (`sleep-call`, `wait-for-timeout`, `network-idle`, `set-timeout`, `fixed-frame-wait`) — land under `evals/fixtures/e2e-escape-cancels-edit/counterexamples/`, each the reference plus exactly one smell; the scorer test compares the fixtures against the exported `DETERMINISM_SIGNALS`, so a sixth signal cannot land without its fixture.
- **Docs:** `.ai/LOCAL-ENFORCEMENT.md` (a per-tier table of the determinism rules, plus the machinery rules learned here — which CI job runs which machinery tests, why the root glob must stay dependency-free, the literal-path rule for `node --test`, and the narrower hard-link note below; "every scorer signal has a fixture" is now, accurately, every **determinism-smell** signal — the hollow-test and gaming signals are covered by inline-source unit tests only, as `evals/README.md` says), `handsontable/.ai/CONVENTIONS.md`, `handsontable/.ai/TESTING.md` (the "Waiting" rule now shows `waitUntil()`, the spy example uses it, the helper list carries it, `*.unit.ts` and the two zero-duration hand-offs are named), `tests/AGENTS.md`, `evals/README.md`, and the Playwright skill's `references/determinism.md`.

**The sleep-debt the burn-down inherits (frozen tier, `eslint --format json` over `handsontable/src/**/__tests__`, `handsontable/test`, walkontable's `test/spec` and `test/unit`, and SheetClip's `test/`, `*.spec.js` + `*.unit.js` + `*.unit.ts`, 1,373 files, 0 lint errors):**

| messageId | warnings | files | where |
|---|---|---|---|
| `noSleep` (pre-existing baseline) | 568 | 67 | 505 in `src/**/__tests__` specs (61 files), 52 in `walkontable/test` (5), 11 in `src/helpers/__tests__/function.unit.ts` — the `*.unit.ts` file the override used to miss |
| `noFrameWait` (**new**) | 1,317 | 153 | 1,316 in `src/**/__tests__` specs (152 files), 1 in `function.unit.ts` |
| `noSetTimeout` (**new**) | 22 | 4 | all in `src/**/__tests__` specs |

New delta: **+1,339 warnings** (1,317 + 22), 0 new errors — the existing debt does not block anything, locally or in CI. Against the previous count (1,154 files, `*.unit.ts` outside the override, no `waitForNextAnimationFrames(0)` exemption): `noSleep` +11 (all `function.unit.ts`), `noFrameWait` −7 (the 8 `waitForNextAnimationFrames(0)` sites now pass, `function.unit.ts` adds 1), `noSetTimeout` unchanged; the 187 other `*.unit.ts` files under `src/**/__tests__` and the 30 under walkontable/SheetClip carry no fixed wait. The heaviest `noFrameWait` files: `autocompleteEditor.spec.js` (151), `notification.spec.js` (140), `numericEditor.spec.js` (61), `numericFormat.spec.js` (61), `settings/validator.spec.js` (55). `noSetTimeout` sits in four files: `settings/validator.spec.js` (17), `core/onKeyDown.spec.js` (2), `validators/validatorCorrection.spec.js` (2), `undoRedo/actions/dataChange.spec.js` (1).

One gotcha found on the way and now documented in `.ai/LOCAL-ENFORCEMENT.md` and `handsontable/.ai/CONVENTIONS.md`, narrowed in review: `eslint-plugin-handsontable` is a `file:` dependency, and pnpm materializes it under `node_modules/.pnpm/` as **hard links** to the source files (same inode, link count 2 — `stat` shows it), not a symlink to the directory and not a plain copy. An edit that writes the file in place is therefore live at once; the trap is the write that replaces the inode — an editor's atomic save-then-rename, `sed -i`, a branch switch that rewrites the file, a fresh worktree, and always a brand-new rule file. After any of those, every lint run (the hooks, `npm run lint`, a `--format json` debt count) executes the rule as it was at the last install while the RuleTester tests, which import the source, already pass — that is how the first debt count came back with 0 new warnings. `pnpm install --frozen-lockfile --offline` relinks it in under a minute.

This is a tooling / test-infrastructure change — no source under `handsontable/src/**` or `wrappers/**` changes, so no changelog entry.

### How has this been tested?

Failing-test-first throughout. First push: both new test files were run before the implementation and failed for the right reason (7 of 21 RuleTester cases: "Should have 1 error but had 0" on the unflagged shapes; 3 of 22 scorer tests: smells not detected, counterexamples not `suspect`, `run-eval` without a counterexample role). Review fix-ups, round 1: the three new 0 ms `valid` cases were red first ("Should have no errors but had 1") until the rule exempted a literal `0`; the five-fixture scorer test was red first (`ENOENT` on the first missing counterexample, and the `>= 5` count) until the three fixtures landed. Round 2: `waitForNextAnimationFrames(0)` as a `valid` case was red until the rule's `isZeroLiteral` landed; the scorer's exemption tests (`test.setTimeout(30_000)`, `setTimeout(fn, 0)`, `setTimeout(fn, delay)` not flagged; `setTimeout(fn, 100)`, `window.setTimeout(fn, 100)`, the timer inside `page.evaluate` flagged) were red against the old regex; `frozenTierLintScope.unit.js` is red on exactly its three `.unit.ts` cases with the override change stashed (verified) and green with it.

- `npm --prefix handsontable run test:eslint-rules` (the dispatcher task; same as `node --test handsontable/.config/plugin/eslint/__tests__/no-fixed-sleep-in-spec.test.mjs`) → `# tests 29 / # pass 29 / # fail 0` (18 valid incl. `setTimeout(done, 0)`, `window.setTimeout(resolve, 0)`, a 0 ms promise, `waitForNextAnimationFrames(0)`, and `setTimeout` methods on non-global objects such as `scheduler.setTimeout(callback, 100)`; 11 invalid incl. `setTimeout(() => done(), 1)`, `globalThis.setTimeout(resolve, 50)`, `waitForNextAnimationFrames(frames)` — a computed count cannot be judged — and one body carrying all three shapes with distinct message IDs and line numbers). Rename probe: with the test file renamed the task exits 1 (`Could not find`), restored afterwards; the `node --test` matrix behind the literal-path rule, on Node 22.23: literal missing → exit 1; `"nomatch/*.mjs"` + literal missing → `tests 0`, exit 0; existing literal + missing literal → exit 0.
- `node --test evals/__tests__/*.test.mjs` → `# tests 33 / # pass 33 / # fail 0` (`score.test.mjs` 27: the two smells, the look-alikes `_registerTimeout`/`clearTimeout`/`waitUntil`/`expect.poll` are not smells, the six exemption/flag cases above plus nested-timer and `globalThis.` spellings, `waitForNextAnimationFrames(0)` passes while `()`, `(2)`, `(frames)` are flagged, each of the five counterexamples carries exactly its one smell and is caught for it, the fixture table equals `DETERMINISM_SIGNALS`, and an integration test that runs `run-eval.mjs --json` and asserts every reference is `meaningful` and the counterexamples between them declare every signal; `counterexamples.test.mjs` 6: the file-name contract, a stray `README.md`/unknown smell rejected, `missReason` null only for the sole-smell sole-problem case).
- `npm run test:tooling` (root) → `# tests 377 / # pass 377 / # fail 0`, and `grep -l "from 'eslint'"` over every directory in that glob finds nothing — the glob is dependency-free, which is what `Checks / tooling tests` needs.
- `node evals/run-eval.mjs` → `References: 3/3 meaningful; counterexamples: 5/5 caught — harness self-test PASSED.` exit 0.
- Jest, `handsontable/`: `test/__tests__/eslintRulesTask.unit.js` (4) + `test/__tests__/frozenTierLintScope.unit.js` (9) → 13 passed; the pre-push gate re-ran both on this push.
- `npm --prefix tests run lint` → exit 0. Selector proof with throwaway probes (not committed): a spec carrying `test.setTimeout(60000)`, `testInfo.setTimeout(60000)`, and `window.`/bare/`globalThis.setTimeout(resolve, 10)` inside `page.evaluate` reports exactly 3 errors, on the three real timers; a **page object** under `fixtures/pages/` carrying `test.setTimeout(30000)`, `setTimeout(fn, 10)`, and `window.setTimeout(fn, 0)` reports exactly 2, on the two timers, with the "spec or a page object" message. `grep -rn "no-restricted-syntax -- DEV-" tests/` now finds both exception sites (the `test.fixme` precedent and the barrier). No other spec or page object in `tests/e2e` / `tests/fixtures` calls `setTimeout`.
- The pre-push gate ran `e2e/customBorders.spec.ts` (the one spec this PR touches; the round-2 change is comment-only) with `npx playwright test --project=e2e-main` against this checkout's own build → green, `pre-push: ok`, and recorded the green-run cache entry for the spec's current content. Earlier rounds: `57 passed (3.5s)` locally and on the first push's pre-push re-run against a fresh `npm --prefix handsontable run build`.
- `npm run eslint` (root: `.github/scripts/`, `bin/`, `scripts/`) → exit 0 with `import/extensions` now enforcing; `npx eslint --ext .mjs evals/` → exit 0. Override proof with throwaway probes: in `scripts/` and in `evals/`, `import … from './lint-files'` / `'./score'` / `'./lib/counterexamples'` / `'../.github/scripts/lib/test-weakening'` each report `Missing file extension "mjs"`, the `.mjs`-suffixed imports pass (so the resolver does resolve `.mjs`); before the fix the same probes passed both ways. `npx eslint --ext .js,.mjs .config/plugin` (core config, the rule + its test) → exit 0; a throwaway `__tests__/*.mjs` importing `'../rules/no-fixed-sleep-in-spec'` without its extension reports `Missing file extension "js"` while `import eslint from 'eslint'` stays clean (`ignorePackages`). The override must be probed through the directory argument: ESLint 8 ignores a dot-folder file passed on its own ("File ignored by default"), but lints it when `.config/plugin` is the argument, which is how `lint:eslint` runs it.
- Hard-link proof: `stat` on `handsontable/.config/plugin/eslint/rules/no-fixed-sleep-in-spec.js` and on its `node_modules/.pnpm/eslint-plugin-handsontable@file+…/rules/` counterpart reports the same inode with link count 2 in this checkout; in a sibling checkout that has switched branches since its install the two are separate inodes and `cmp` shows the store copy is stale — the branch-switch case the note now names.
- Debt count (plugin verified in sync with the source via the shared inode): `npx eslint --no-error-on-unmatched-pattern -f json -o … "src/**/__tests__/**/*.spec.js" "src/**/__tests__/**/*.unit.js" "src/**/__tests__/**/*.unit.ts" "test/**/*.spec.js" "test/**/*.unit.js" "test/**/*.unit.ts" "src/3rdparty/walkontable/test/**/*.spec.js" "src/3rdparty/walkontable/test/unit/**/*.unit.ts" "src/3rdparty/SheetClip/test/*.unit.ts"` from `handsontable/` → 1,373 files, 0 errors, warnings per `messageId` as in the table above.

### Types of changes

### Related issue(s):

1. DEV-2764

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

(Tooling and test infrastructure only: lint configs, the custom ESLint plugin, the evals harness, the Playwright tier's config, one CI workflow step, two Jest pins for the lint machinery, one removed dead test-helper alias, and one comment-level exception in a Playwright spec.)

### Checklist:

- [x] My change requires a change to the documentation (done in this PR: `.ai/LOCAL-ENFORCEMENT.md`, `handsontable/.ai/CONVENTIONS.md`, `handsontable/.ai/TESTING.md`, `tests/AGENTS.md`, `evals/README.md`, the Playwright skill's determinism reference).

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2764

---

> [!NOTE]
> **Low Risk**
> Tooling and test-lint only—no product runtime changes. Frozen-tier new matches are warnings and surface existing debt; CI gains an explicit custom-rule test step.
> 
> **Overview**
> Aligns what counts as a **fixed wait** across Playwright ESLint, the frozen-tier `handsontable/no-fixed-sleep-in-spec` rule, and `evals/score.mjs`, closing the gap where `setTimeout` and frame-count waits slipped past enforcement.
> 
> **Playwright (`tests/.eslintrc.cjs`, error):** bans global `setTimeout` (including inside `page.evaluate`) while keeping `test.setTimeout` / `testInfo.setTimeout` as budgets; applies to page objects as well as specs. Documents the existing 0ms `macrotaskBarrier()` exception in `customBorders.spec.ts` with an eslint-disable tied to DEV-2764.
> 
> **Frozen tier (warn):** the custom rule now flags non-zero literal global `setTimeout` and `waitForNextAnimationFrames` except literal `0`, points authors at `waitUntil()`, and applies to `*.unit.ts` (not only `*.spec.js` / `*.unit.js`). Adds RuleTester coverage, `npm run test:eslint-rules`, a CI step on **Lint / core**, and Jest guards so the task list cannot silently skip tests. Removes the unused `waitForNameAnimationFrames` alias.
> 
> **Evals:** extends determinism scoring with `set-timeout` and `fixed-frame-wait` (matching lint exemptions), adds `counterexamples/` fixtures and `evals/lib/counterexamples.mjs` so `run-eval.mjs` fails if a reference passes or a named counterexample is not caught for exactly one smell.
> 
> **Docs / misc:** determinism tables in `.ai/LOCAL-ENFORCEMENT.md` and testing guides; fixes root `import/extensions` for `scripts/**` and `evals/**` (pattern form that was previously ineffective).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d5c136a468037ab32048d5042c3175f62dae2816. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2764 Determinism lint: ban setTimeout in Playwright specs, flag setTimeout and frame-count waits in the frozen suite, mirror both in the evals scorer](https://app.clickup.com/t/9015210959/DEV-2764)

evals/__tests__/score.test.mjs
```diff
@@ -136,6 +140,83 @@ test('determinism smells: sleep(, waitForTimeout, and networkidle are detected',
...
+    'escape-cancels-edit-network-idle.spec.ts': 'network-idle',
+    'escape-cancels-edit-set-timeout.spec.ts': 'set-timeout',
+    'escape-cancels-edit-frame-wait.spec.js': 'fixed-frame-wait',
+  };
+
+  assert.deepEqual(
+    Object.values(expectedByFile).sort(),
+    findDeterminismSmells('sleep( waitForTimeout( networkidle setTimeout( waitForNextAnimationFrames(')
+      .map(smell => smell.type).sort(),
+    'the table above names every signal the scorer knows',
```
budnix: This guard can only ever report the five signals that are already in the table, because the string it scans is hardcoded. I added a sixth pattern to `findDeterminismSmells` with no fixture and this assertion stayed green, so the exact case it exists to catch still gets through. Comparing against a signal list exported from `score.mjs` would make it fail instead.

evals/run-eval.mjs
```diff
@@ -174,13 +196,24 @@ if (parsed.json) {
 
 const references = results.filter(result => result.role === 'reference');
 const failedReferences = references.filter(result => result.score.verdict !== 'meaningful');
+const counterexamples = results.filter(result => result.role === 'counterexample');
+// A counterexample that scores meaningful means the scorer no longer sees the smell it carries.
+const missedCounterexamples = counterexamples.filter(result => result.score.verdict !== 'suspect');
```
budnix: A counterexample counts as caught as soon as it is not `meaningful`, and `suspect` is reached by any problem at all — I checked, and an empty file, a hollow test, and a `.skip` all score `suspect` with zero determinism smells. `listFixtureFiles` scores every file in the folder, so a stray `README.md` would also count, and the `noisy` filter above hides the output that would have shown it. Asserting the expected smell type per file, not just the verdict, would keep the count honest.

handsontable/.config/plugin/eslint/rules/no-fixed-sleep-in-spec.js
```diff
@@ -1,31 +1,94 @@
...
+          return;
+        }
+
+        if (isGlobalTimerCall(callee) && isFixedDuration(node.arguments[1])) {
+          context.report({ node, messageId: 'noSetTimeout' });
+
+          return;
+        }
+
+        if (callee && callee.type === 'Identifier' && callee.name === 'waitForNextAnimationFrames') {
```
budnix: `waitForNextAnimationFrames(0)` resolves right away — with `framesToWait` at 0 the helper takes its `totalFramesToWait === 0` branch and `minimumElapsedTime` is 0, so no time passes. That is the same zero-duration hand-off `isFixedDuration` exempts a few lines up for `setTimeout(fn, 0)`, yet here it still warns, and there are 8 live sites where it tells the author to swap a barrier for `waitUntil()`. Exempting a literal 0 would line the two shapes up.
```suggestion
        if (callee && callee.type === 'Identifier' && callee.name === 'waitForNextAnimationFrames'
          && !(node.arguments[0] && node.arguments[0].type === 'Literal' && node.arguments[0].value === 0)) {
```

tests/e2e/customBorders.spec.ts
```diff
@@ -872,10 +872,17 @@ test.describe('CustomBorders progressive application (customBordersProgressive)'
...
+   * it carries the tier's setTimeout exception: a negative ("no further batch fired") cannot be
+   * polled for, so `expect.poll` is not a replacement here.
+   *
+   * TODO (owned by the CustomBorders progressive-application work, #13166): expose a "batch
+   * queue is empty" probe on the fixture and assert on it, then delete this barrier and its
+   * eslint exception.
    */
   async function macrotaskBarrier(page: Page): Promise<void> {
     await page.evaluate(() => new Promise((resolve) => {
+      // eslint-disable-next-line no-restricted-syntax -- CustomBorders progressive-application work: a 0ms macrotask barrier for negative assertions, not a duration wait (see the JSDoc above)
```
budnix: This line names the work in prose but leaves out the DEV id, and both the message on `tests/.eslintrc.cjs:61` and `tests/AGENTS.md:167` ask for the `-- DEV-1234: <why>` form. `grep -rn "no-restricted-syntax -- DEV-" tests/` finds the `test.fixme` precedent in `editor-window-scroll.spec.ts:29` but skips this site, so it is not counted or attributable the way the convention intends. The `#13166` in the JSDoc above is a PR number, so the task id still needs to land on this line.

.ai/LOCAL-ENFORCEMENT.md
```diff
@@ -89,7 +114,8 @@ presence gate or the test requirement. Do not use it to dodge writing tests.
...
+- **A custom ESLint rule is live only after `pnpm install`.** `eslint-plugin-handsontable` is a `file:` dependency, and pnpm materializes those as a *copy* under `node_modules/.pnpm/`, not a symlink. Until you reinstall (`pnpm install --frozen-lockfile --offline`, under a minute), every lint run — the hooks, `npm run lint`, a `--format json` debt count — executes the rule as it was at the previous install, while the RuleTester tests (which import the source file) already pass. Verify with `diff` against the copy before trusting a lint result that disagrees with the tests.
```
budnix: pnpm hard-links this file rather than copying it — on the installed tree the source and the path under `node_modules/.pnpm/` both report inode 289527719 with a link count of 2. So an edit that writes in place is live with no reinstall, and a reader who tries that will see the rule already working and stop trusting the note. The trap is real but narrower: it bites on an editor's save-then-rename, on `sed -i`, on a branch switch or a fresh worktree, and always for a brand-new rule file.

handsontable/.eslintrc.js
```diff
@@ -184,6 +184,25 @@ module.exports = {
         'import/no-relative-packages': 'off',
       }
     },
+    // The custom ESLint rules' own RuleTester tests run as ESM under `node --test`
+    // (`npm run test:eslint-rules`; CI's `Lint / core` job), so a relative import must carry the
+    // extension Node's ESM resolver requires — the same allowance the root config gives
```
budnix: The root config block this points at is itself the broken form described two lines below — `.eslintrc.js:197` still has `js: ['error', 'always']`. I ran both against the same probe file: the string form here reports `Missing file extension "js"`, the array form there reports nothing, so `import/extensions` enforces neither direction for `scripts/**` or `evals/**` right now, including the new `evals/` files in this PR. Might be worth fixing in passing, or pointing this sentence at a precedent that works. Worth a quick check on the `.mjs` side too — my `evals/` probe passed both ways, so the resolver's extension list may also be in play.

handsontable/.eslintrc.js
```diff
@@ -232,9 +251,11 @@ module.exports = {
         'handsontable/restricted-module-imports': 'off',
         'handsontable/require-async-in-it': 'error',
         // Determinism guards for the frozen Jasmine suite. WARN, not error: the
-        // existing sleep()/it.flaky() debt must surface without red-walling CI.
-        // Escalation to error happens in the flip-to-blocking task once the
-        // debt is burned down. New E2E belongs in Playwright (tests/e2e).
+        // existing sleep()/setTimeout(fn, <ms>)/waitForNextAnimationFrames()/it.flaky()
```
budnix: This override registers the rule for `*.unit.js` and `*.spec.js` only, so all 217 `*.unit.ts` files sit outside it. It already misses real sites — `src/helpers/__tests__/function.unit.ts` has eleven `sleep()` calls and one frame wait — and the same probe source gives one warning saved as `.unit.js` and zero saved as `.unit.ts`. Adding `*.unit.ts` to `files` would pull them in, and would nudge the debt numbers in the description.

evals/score.mjs
```diff
@@ -248,8 +248,10 @@ export function findGamingSignals(src) {
 }
 
 /**
- * Detect determinism smells — fixed sleeps and load-state waits that make a
- * test timing-dependent instead of condition-based (web-first waits).
+ * Detect determinism smells — fixed sleeps, fixed timers, frame-count waits, and
+ * load-state waits that make a test timing-dependent instead of condition-based
+ * (web-first waits, `expect.poll`, the frozen tier's `waitUntil()`). Mirrors the
```
budnix, line 449: Adding to the point already raised on line 265: the pattern also catches two shapes the frozen-tier rule deliberately lets through — a literal zero delay like `setTimeout(fn, 0)`, which is the exemption this PR just added and exactly what `macrotaskBarrier()` uses, and a computed delay like `setTimeout(fn, delay)`. I ran both and each comes back as `set-timeout`. So this "Mirrors the lint bans" line claims a bit more than the pattern does today.

evals/score.mjs
```diff
@@ -259,6 +261,9 @@ export function findDeterminismSmells(src) {
     ['sleep-call', /\bsleep\s*\(/g],
     ['wait-for-timeout', /\bwaitForTimeout\s*\(/g],
     ['network-idle', /\bnetworkidle\b/g],
+    // The `\b` keeps `_registerTimeout(` and `clearTimeout(` out; `window.setTimeout(` is in.
+    ['set-timeout', /\bsetTimeout\s*\(/g],
+    ['fixed-frame-wait', /\bwaitForNextAnimationFrames\s*\(/g],
```
budnix: `common.js:157` still exports `waitForNameAnimationFrames`, a deprecated alias that forwards straight to `waitForNextAnimationFrames` with the same padded delay, and neither this pattern nor the rule's `noFrameWait` branch matches it. No spec uses it today, so nothing slips through right now, but typing the old name is the cheapest way around both new gates. Matching both names covers it, or the alias could just go since nothing depends on it.
```suggestion
    ['fixed-frame-wait', /\bwaitForNe(?:xt|ame)AnimationFrames\s*\(/g],
```

handsontable/scripts/tasks.json
```diff
@@ -136,6 +136,12 @@
       "deps": ["downlevel:types"],
       "mode": "inherit"
     },
+    "test:eslint-rules": {
+      "cmd": "node --test \".config/plugin/eslint/__tests__/*.test.mjs\"",
```
budnix: `node --test` with a quoted glob exits 0 and prints `tests 0` when nothing matches, so a rename here, or a later rule's test landing with a different suffix, leaves the new `Lint / core` step green with nothing run. It works today — I get 27 passing — but a gate that can quietly stop guarding feels like the kind of thing this repo normally pins. Naming the file alongside the glob is enough, since an explicit missing path exits 1.

tests/.eslintrc.cjs
```diff
@@ -35,6 +36,14 @@ module.exports = {
...
         message: 'No fixed sleep() delay — wait for a condition instead. See .claude/skills/handsontable-playwright-e2e/references/determinism.md.',
       },
+      {
+        // The global timer only: bare `setTimeout(`, `window.setTimeout(`, and `globalThis.setTimeout(` —
+        // the usual disguise is a timer inside `page.evaluate` once `waitForTimeout` is banned. The member
+        // form is pinned to the global object on purpose: Playwright's `test.setTimeout(ms)` and
+        // `testInfo.setTimeout(ms)` set a budget, not a wait, and must stay legal.
+        selector: "CallExpression[callee.name='setTimeout'], CallExpression[callee.object.name='window'][callee.property.name='setTimeout'], CallExpression[callee.object.name='globalThis'][callee.property.name='setTimeout']",
+        message: 'No setTimeout() in a spec — a fixed timer is not a wait. Poll a data probe with expect.poll, or use a web-first assertion. A justified exception (a scheduling barrier, never a duration) carries the same eslint-disable line as test.fixme below, naming the owning task, so it stays counted and attributable. See tests/AGENTS.md.',
```
budnix: `no-restricted-syntax` sits at the top level with no `files` scoping and the lint script covers `e2e` and `fixtures`, so this reaches page objects too, yet the message says "in a spec" and asks for an owning task, which does not really fit fixture plumbing. Three `.html` fixtures already drive timers on purpose, and `fixtures/pages/HiddenInitPage.ts:9` describes exactly that rAF and timer ordering, so a real timer moving into a page object is plausible. Curious whether reaching `fixtures/` was intended — if it was, the wording could say so; if not, scoping the selector to `e2e/**/*.spec.ts` would match the intent.

.ai/LOCAL-ENFORCEMENT.md
```diff
@@ -73,6 +73,31 @@ Machine-enforced by the presence gate; full decision rules in
...
+the Playwright tier, the `waitUntil(condition, timeout)` spec global
+(`handsontable/test/helpers/common.js`) on the frozen tier. The Playwright tier's
+only sanctioned exception is a 0ms scheduling barrier that lets a negative
+assertion prove "nothing else fired"; it carries the same eslint-disable line as
+`test.fixme`, naming the owning task (`tests/AGENTS.md`). The frozen tier stays
+at **warn** until the debt is burned down — the count is the burn-down's backlog,
+not a gate. The rule's RuleTester coverage lives in
+`handsontable/.config/plugin/eslint/__tests__/` and runs through
+`npm run test:eslint-rules` in `handsontable/` (CI: the `Lint / core` job). Every
+scorer signal has a fixture under `evals/fixtures/<case>/counterexamples/` that
```
budnix: This says every scorer signal has a counterexample fixture, but the five new files only cover the determinism family — `hollowTests` and `gamingSignals` have none. Someone reading it later would think a hollow-test regression is already fixture-covered, when it is only covered by the inline-source unit tests, which is a weaker guarantee. The narrower wording in `evals/README.md` is the accurate one, so matching it here would help.

evals/score.mjs
```diff
@@ -259,6 +261,9 @@ export function findDeterminismSmells(src) {
     ['sleep-call', /\bsleep\s*\(/g],
     ['wait-for-timeout', /\bwaitForTimeout\s*\(/g],
     ['network-idle', /\bnetworkidle\b/g],
+    // The `\b` keeps `_registerTimeout(` and `clearTimeout(` out; `window.setTimeout(` is in.
+    ['set-timeout', /\bsetTimeout\s*\(/g],
```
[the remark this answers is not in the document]
danielzytohoc: Narrowed to the global timer: the pattern now rejects any member-access setTimeout, so test.setTimeout(30_000), testInfo.setTimeout(ms), and another object's setTimeout method are no longer smells, while bare, window., and globalThis. still are. Covered by new cases in evals/__tests__/score.test.mjs.

evals/__tests__/score.test.mjs
danielzytohoc: The scorer now exports DETERMINISM_SIGNALS, derived from its own signal table, and the guard compares the counterexample fixtures against that list instead of a hardcoded string. Adding a sixth pattern to findDeterminismSmells without a fixture now fails this test and the run-eval integration test.

evals/run-eval.mjs
danielzytohoc: A counterexample now declares its smell in its file name (escape-cancels-edit.set-timeout.spec.ts, contract in evals/lib/counterexamples.mjs) and counts as caught only when the scorer reports exactly that smell and no other problem, so an empty file, a hollow test, or a .skip no longer passes as caught. A file in the folder that names no known smell, a stray README included, is reported as a fixture error and fails the run. The missed-counterexample line now prints the reason instead of the verdict.

handsontable/.config/plugin/eslint/rules/no-fixed-sleep-in-spec.js
danielzytohoc: Done: a literal 0 frame count is exempt now, matching the setTimeout(fn, 0) hand-off, and the 8 live sites stop warning. Anything else, a computed count included, still reports noFrameWait; both shapes are RuleTester cases. The recount is in the description: noFrameWait went from 1,324 to 1,317 (the 8 exempt sites, plus 1 from the newly covered function.unit.ts).

tests/e2e/customBorders.spec.ts
danielzytohoc: The line now carries the task-id form. The PR that owns this spec has no ClickUp task behind it (its branch and related issues are GitHub-only), so the line names DEV-2764, the task that introduced the setTimeout ban and accepted this one exception; the JSDoc above states that and keeps the TODO for the fixture probe that will replace the barrier. grep for the convention now finds both sites.

.ai/LOCAL-ENFORCEMENT.md
danielzytohoc: Rewritten to the narrower claim: pnpm hard-links the plugin's files (same inode, link count 2, checkable with stat), so an in-place edit is live at once, and the trap is the write that replaces the inode: an editor's atomic save, sed -i, a branch switch, a fresh worktree, and always a brand-new rule file. I confirmed both halves here: this checkout shows the shared inode, while a sibling checkout that has switched branches since its install holds a separate, stale store copy. The same correction landed in handsontable/.ai/CONVENTIONS.md and the PR description.

handsontable/.eslintrc.js
danielzytohoc: Fixed the root config in passing rather than repointing: the scripts/** and evals/** override now uses the pattern form with ignorePackages, and this override matches it. Probes in both trees now report Missing file extension "mjs" for an extensionless relative import and pass with the extension, so the resolver does resolve .mjs and the earlier both-ways pass was the array form alone; npm run eslint at the root and eslint over evals/ stay clean with the rule enforcing. The comment here now describes that history instead of pointing at the broken block.

handsontable/.eslintrc.js
danielzytohoc: Added *.unit.ts to the override, and a Jest test (test/__tests__/frozenTierLintScope.unit.js) now pins the rule's reach through ESLint's own config resolution: on at warn for .unit.ts, .unit.js and .spec.js under src/**/__tests__, walkontable's test/ and SheetClip's test/, off for source files. The recount over all 1,372 files is in the description: noSleep 568 in 67 files (the 11 new ones are all function.unit.ts), noFrameWait 1,317 in 153, noSetTimeout 22 in 4, still 0 errors.

evals/score.mjs
danielzytohoc: The scorer now reads the call's arguments and mirrors the frozen rule's exemptions: setTimeout(fn, 0), a computed setTimeout(fn, delay), and waitForNextAnimationFrames(0) pass, while setTimeout(fn, 100), window.setTimeout(fn, 100), and the timer inside page.evaluate are still flagged. Each of those shapes is a unit test, and the comment and README now say the mirroring includes the exemptions.

evals/score.mjs
danielzytohoc: Took the delete route: waitForNameAnimationFrames is gone from test/helpers/common.js, and a repo-wide grep finds no spec, typing, or doc that names it, so there is no second spelling left for either the rule or the scorer to match.

handsontable/scripts/tasks.json
danielzytohoc: Pinned, with one correction to the suggested shape: on Node 22 node --test exits 1 only when nothing in the list matched and no pattern is a glob, so a literal path next to the glob still exits 0 with tests 0 when the literal is missing, and a second literal file that still exists masks a renamed first one the same way. The task now names the file as its only, literal entry (a rename makes it exit 1, verified), and test/__tests__/eslintRulesTask.unit.js pins the list in both directions and bans glob characters so a second rule test cannot reintroduce the gap.

tests/.eslintrc.cjs
danielzytohoc: Reaching fixtures/ was intended and stays: a timer moved from a spec into the page object it drives is the same fixed wait one file further away, and a probe page object confirms the two real timers are reported while test.setTimeout is not. The message now says a spec or a page object, spells the disable-line form, and describes what qualifies as the one justified exception (a 0 ms scheduling barrier for a negative assertion, never a duration); the config header, tests/AGENTS.md and the skill reference state the scope, and the .html fixtures that drive timers deliberately remain unlinted.

.ai/LOCAL-ENFORCEMENT.md
danielzytohoc: Narrowed to match evals/README.md: every determinism-smell signal has a counterexample fixture that must score suspect for exactly that smell, while the hollow-test and gaming signals are covered by the inline-source unit tests in evals/__tests__/score.test.mjs only. The machinery rule further down uses the same wording.

budnix (approved):

PR #13350, opened 2026-09-02
DEV-2765: Ratchet warn-level determinism rules on added spec lines

danielzytohoc:
### Context

The PR adds a diff-scoped **determinism ratchet** to the enforcement machinery, and wires it into pre-push and CI.

Today the frozen Jasmine suite's determinism rules — `handsontable/no-fixed-sleep-in-spec`, `handsontable/no-new-it-flaky`, `handsontable/no-skipped-test` — are `warn` in `handsontable/.eslintrc.js`, and nothing consumes warnings: no `--max-warnings`, `scripts/pre-push.mjs` and `scripts/lint-staged.mjs` block on errors only, and the CI lint job exits 0 with warnings. The existing debt (66 spec files, ~560 `sleep()` calls, 13 `it.flaky()`, 10 skips) makes a flip to `error` impossible, so a new `sleep()` today reaches `develop` with no gate in its way — the sleep-debt burn-down cannot converge while new debt keeps arriving.

The ratchet closes the gap without red-walling the debt. A warning from one of the three rules on a line the branch **added** is an error; a pre-existing occurrence on an unchanged line stays a warning. The debt can only shrink.

What ships:

- `.github/scripts/lib/lint-ratchet.mjs` — pure logic, mirroring the presence gate's lib/CLI split: `RATCHETED_RULES` (the single source of truth), `RATCHETED_FILES` + `selectRatchetedFiles()` (the trees the rules apply to; the Playwright tier is excluded because its own config bans `sleep()` at `error`; `*.unit.ts` is a candidate and is covered the moment the frozen-tier override in `handsontable/.eslintrc.js` names it — until then ESLint reports none of the three rules there and a candidate with no findings never blocks), `parseAddedLines()` (`-U0` hunks, omitted counts, renames under the new path, deletions dropped, header detection only outside a hunk so an added line starting with `++ ` cannot be mistaken for a file header, raw paths — the CLI runs git with `core.quotePath=false`, so a non-ASCII name keys the same file as the `-z` name list, and a header git still C-quotes is left unattributed rather than decoded — CRLF), `selectRatchetedFindings()` (the intersection; a file with a fatal parse message is skipped whole; severity is not consulted, so an `error` from a ratcheted rule is reported too), and `formatReport()` (names each finding as `file:line — rule: message`, the fix path, and the moved-line trade-off: re-indented or moved lines count as added, a renamed file adds nothing on its own).
- `.github/scripts/lint-ratchet.mjs` — the CLI. Base from `--base` / `GATE_BASE` / the `origin/develop` merge-base, always the **merge-base** of that ref with HEAD and never the ref itself: an unknown ref, or one with no merge-base (a clone too shallow to reach the fork point), skips with exit 0 and a notice naming the ref — diffing against a non-ancestor would read every line the base gained since the fork as the branch's own. Collects the changed `handsontable/{src,test}/**/*.{spec,unit}.js` and `*.unit.ts`, diffs the **whole branch** with `-M` and filters the parsed result to those candidates (a pathspec naming only a renamed spec's new path hides its old path from rename detection, so a plain `git mv` would read as a new file with every legacy line added), exits 0 at once when none changed or none gained a line, runs the **package's own** ESLint (`handsontable/node_modules/eslint/bin/eslint.js`, cwd `handsontable/`, `--format json`) on exactly those files so the custom plugin and `.eslintignore` resolve as in CI, intersects, prints the Markdown report, and exits 1 **only** when the list is non-empty. Every tooling gap — no usable base, a failed `git diff`, ESLint missing, exit 2, unparsable output, a killed child, an unexpected throw — prints a notice and exits 0.
- `scripts/pre-push.mjs` — a blocking step 2b after the ESLint step, spawned only when `needsDeterminismRatchet(changed)` (a pure, tested seam) says a ratcheted file changed; a killed child is classified by the existing `isSpawnInfraFailure()` and does not block.
- `.github/workflows/lint.yml` — the ratchet is a **step of the existing `core` job** (the install is already warm, and `build` needs `lint`, so a second job would sit on the critical path), PR-only and scope-gated through a new `run-ratchet` input that `test.yml` derives from a new `lint-ratchet` path filter in `checks.yml` (`handsontable/src/**`, `handsontable/test/**` — the same trees as `RATCHETED_FILES`). The `core` checkout takes `fetch-depth: 0` only on the runs where the step executes (the expression carries the step's own condition). **The base is the PR base branch's live tip**: the step fetches `refs/heads/<base.ref>` and passes `GATE_BASE=origin/<base.ref>` — never the event payload's frozen `base.sha`. The checkout is `refs/pull/N/merge`, and that SHA is already an ancestor of it, so `merge-base(base.sha, HEAD)` is `base.sha` itself; a "Re-run all jobs" after the base moved, or a branch that merged the base after it forked, would then read every later base commit as a line the PR added — the first `GATE_BASE` gate that blocks cannot afford that. A failed refresh of the base branch is a `::warning::` and the CLI diffs from the checkout's copy. No token or secret, so no fork guard is needed. Local and CI run the same script with the same rule set: local scope == CI scope.
- Docs: `.ai/LOCAL-ENFORCEMENT.md` (the new gate: what blocks — re-indented or moved lines count as added, a `git mv` does not; how to satisfy it — a condition wait or `// eslint-disable-next-line <rule> -- <ticket>: <reason>`; **which base** — pre-push uses the `origin/develop` merge-base while CI uses the merge-base with the base branch's live tip, so on a branch targeting `release/*` or stacked on another feature branch pre-push may report the parent's lines, confirm with `--base <parent>`; `*.unit.ts` coverage; and when it skips), a root `AGENTS.md` gotcha (a blocking diff-scoped PR gate diffs from the base branch's live tip, never `base.sha`), `handsontable/.ai/CONVENTIONS.md` (the ratchet paragraph below the rules table), `handsontable/.ai/TESTING.md`, the rule-level comments in `handsontable/.eslintrc.js` (both the sleep rules and the skip rule point at the ratchet), and the `lefthook.yml` step comment.

Design trade-off, documented in the test, the report, and `LOCAL-ENFORCEMENT.md`: a **re-indented or moved** line counts as added. A unified diff cannot tell a move or a reformat from an addition, and ignoring re-added text would let a genuinely new `sleep()` hide behind any deleted one, so the ratchet stays strict — a reformat that touches a `sleep()` is the moment to replace it. A **renamed file** is the one case rename detection resolves: a plain `git mv` adds no line and does not block; only the lines changed inside it count.

Known follow-up (a separate backlog ticket, not this PR): nothing lints `.github/scripts/**` today — the root `eslint` script passes directories with no `--ext`, no `overrides[].files` glob names that tree, and `lintable()` in `scripts/lint-files.mjs` drops dot-segment paths. The ratchet's files are clean under the config that would apply (checked by feeding ESLint the root config with `.github/scripts/**/*.mjs` added to the scripts override, see the commands below), so wiring the directory in will not turn them red.

This is a tooling / test-infrastructure change: nothing under `handsontable/src/**` or `wrappers/**` changes, so **no changelog entry** is required (the changelog gate is path-aware and passes automatically).

- Unit tests added/modified (`*.unit.js`): none — the machinery has its own `node --test` suites:
  - `.github/scripts/__tests__/lint-ratchet.test.mjs` (25 tests on the pure lib): rule-set pin; every ratcheted rule resolves to `warn` **per path** — one `.spec.js` and one `.unit.js` path in each tree, resolved through the overrides the way ESLint applies them (basename globs, ordered, last one wins; the resolver has its own test), so a `files` glob that stops matching the frozen tier fails the test where a flattened id→level map could not; file selection; `-U0`/omitted-count/rename/deletion/context/`+++`-content/CRLF diff parsing; a raw non-ASCII path is keyed as-is and a still-quoted header is left unattributed without leaking its hunk into the next file; added sleep reported, pre-existing sleep not, moved line reported, fatal parse skipped, non-ratcheted rule ignored, error-severity reported, absolute→relative path match, sort order; report shape incl. the moved-line note.
  - `.github/scripts/__tests__/lint-ratchet-cli.test.mjs` (12 tests on the CLI seam, each against a throwaway git repository built with `execFileSync` — the script and its lib copied under `.github/scripts/` so `repoRoot()` resolves the fixture, a stub ESLint that reports **every** `sleep(` line so the intersection is provably the CLI's): rename-only commit of a spec carrying a legacy `sleep()` → exit 0 on the fast path; rename + one new `sleep()` → exactly the new line, exit 1; `GATE_BASE` resolves like `--base`; unknown ref → exit 0 with a notice naming it; a base with no merge-base (orphan history) → exit 0 with a notice instead of a diff against it; **the CI shape** — a detached `refs/pull/N/merge`-style merge commit diffed from the live base tip reads only the branch's lines (exit 0, ESLint ran) while the frozen fork SHA attributes the base's `sleep()` to the branch (exit 1 at the base's line); **a branch that merged the base after it forked** — same two verdicts, pinning that a head-SHA checkout alone would not have helped; a spec named `café.spec.js` is reported at its line (the test first asserts git's default output quotes the name); the three **tooling-gap skips** — ESLint not installed, ESLint exiting 2 (the notice carries its first stderr line), unparsable JSON — each starting from the commit the in-place control test proves is blocked, so exit 0 there is a skip of a real finding; in-place addition control → exit 1 at its line.
  - `.github/scripts/__tests__/lint-ratchet-workflow.test.mjs` (5 tests, text-based like `fork-guards.test.mjs`): the ratchet is a step of `core` and `lint.yml` has no other job than `core` and `visual-tests`; the step keeps the `run-ratchet && pull_request` scope gate; no expression in `lint.yml` reads `base.sha`, the step fetches `refs/heads/${BASE_REF}` from `base.ref` and passes `GATE_BASE="origin/${BASE_REF}"`; a failed refresh is a `::warning::`; the checkout's conditional `fetch-depth` carries the step's exact condition.
  - `scripts/__tests__/pre-push.test.mjs` (+2 tests for the new step's skip path and its run path).
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`): none — no runtime behavior changes.
- Type tests (`*.types.ts`) updated if public API changed: n/a.
- For a bug fix — the spec that fails without this fix: `lint-ratchet-cli.test.mjs` was written first and run against the pre-fix CLI: 5 of the first 6 failed for the intended reasons — rename-only and `GATE_BASE` exited 1 on the untouched legacy `sleep()` (line 5), rename + new sleep reported lines 3 **and** 6, the unrelated base exited 1 on the legacy line, the unknown-ref notice did not name the ref; only the in-place control passed. The review-round tests were checked the same way: the stale-base halves (`GATE_BASE: fork` → exit 1 at `landedOnTheBase.spec.js:5`) are the defect the live tip fixes, run against real merge commits; the `café.spec.js` test fails on the pre-fix CLI (the per-byte decode keyed the file as `cafÃ©.spec.js`, so no finding and exit 0). The lib test file was originally run before the lib existed and failed with `ERR_MODULE_NOT_FOUND`; the `+++`-content parsing case was checked against the existing naive header-first parser (`diff-coverage.mjs` `addedLinesByFile`), which mis-attributes those lines to `not-a-file.spec.js` — the new parser does not.
- Demo page / recorded trace (for UI changes): n/a.

### Commands run

```
$ node --test .github/scripts/__tests__/lint-ratchet.test.mjs .github/scripts/__tests__/lint-ratchet-cli.test.mjs .github/scripts/__tests__/lint-ratchet-workflow.test.mjs
# tests 42 / # pass 42 / # fail 0

$ npm run test:tooling            # every enforcement-machinery suite, incl. the pre-push tests
# tests 406 / # pass 406 / # fail 0 / duration_ms 2401

$ npm run eslint                  # root scope: .github/scripts/ bin/changelog bin/lib scripts/
(clean, exit 0)

# The config that WOULD apply to .github/scripts/** once it is wired into lint:
# a throwaway probe.cjs at the repo root = require('./.eslintrc.js') with
# '.github/scripts/**/*.mjs' appended to every override's `files`.
$ npx eslint --no-ignore --no-eslintrc -c probe.cjs .github/scripts/lib/lint-ratchet.mjs .github/scripts/lint-ratchet.mjs .github/scripts/__tests__/lint-ratchet*.test.mjs
(clean, exit 0 — the pre-fix lib reported 4 errors: 3× jsdoc/valid-types on `readonly string[]` / `readonly RegExp[]`, 1× no-nested-ternary on the sort comparator)

$ node .github/scripts/lint-ratchet.mjs                       # this branch (no spec changed)
Determinism ratchet: no changed spec/unit file under handsontable/ — nothing to ratchet.   exit=0

$ GATE_BASE=origin/develop node .github/scripts/lint-ratchet.mjs   # the CI contract
Determinism ratchet: no changed spec/unit file under handsontable/ — nothing to ratchet.   exit=0

$ node .github/scripts/lint-ratchet.mjs --base no-such-ref-xyz   # unknown base ref
Determinism ratchet: no merge-base between "no-such-ref-xyz" and HEAD (unknown ref, or a clone too shallow to reach the fork point) — skipped; CI's lint is authoritative.   exit=0
```

Proof on this branch, with throwaway scratch commits (reset afterwards, not in the PR), on `handsontable/src/renderers/numericRenderer/__tests__/contextMenu.spec.js`, which already carries a `sleep(304)` (line 38), against the real ESLint:

1. Added a new `await sleep(50);` plus a comment line. `pre-commit` (lint-staged) surfaced **2 warnings and let the commit through** — the gap. The ratchet:
   ```
   ## Determinism ratchet
   ❌ 1 warn-level determinism finding(s) sit on lines this branch **added**. Pre-existing occurrences stay warnings; new ones block.
   - `handsontable/src/renderers/numericRenderer/__tests__/contextMenu.spec.js:40` — `handsontable/no-fixed-sleep-in-spec`: Do not use a fixed sleep() delay. …
   exit=1
   ```
   Only line 40 (the new one) is listed; the pre-existing `sleep(304)` (now line 39) is not. `node scripts/pre-push.mjs` on the same state: presence gate pass → ESLint 2 warnings → ratchet block, `exit=1`.
2. Reset; added only a comment line to the same debt-laden file: `Determinism ratchet: no new sleep() / it.flaky() / skip on the lines this branch added.` `exit=0` — existing debt does not newly block.
3. Reset; added a new `sleep(50)` behind `// eslint-disable-next-line handsontable/no-fixed-sleep-in-spec -- <ticket>: …`: `exit=0` — the documented exception works.
4. Reset; `git mv contextMenu.spec.js contextMenuRenamed.spec.js`, nothing else (the reviewer's repro). Fixed CLI: `Determinism ratchet: the changed spec/unit files gained no line — nothing to ratchet.` `exit=0` — the fast path, ESLint never spawned. The pre-fix CLI on the same state: `❌ 1 … contextMenuRenamed.spec.js:38 — handsontable/no-fixed-sleep-in-spec` `exit=1` — the untouched legacy line, blocked.
5. On top of the rename, inserted `await sleep(50);` at line 38 (the legacy `sleep(304)` shifts to 39). `pre-commit` surfaced 2 warnings (38, 39) and let it through. Fixed CLI: `❌ 1 … contextMenuRenamed.spec.js:38 — handsontable/no-fixed-sleep-in-spec` `exit=1` — the new line only; line 39 is not listed.

The wiring reads: `lint.yml` job `core`, step `Fail on new sleep() / it.flaky() / skip in the changed specs`, `if: inputs.run-ratchet && github.event_name == 'pull_request'`, `BASE_REF: ${{ github.event.pull_request.base.ref }}`, `GATE_BASE="origin/${BASE_REF}"` after a `git fetch` of that branch, and the checkout's `fetch-depth` set to `0` under the same condition (`lint-ratchet-workflow.test.mjs` pins all of it); `checks.yml` filter `lint-ratchet: ["./handsontable/src/**", "./handsontable/test/**"]` exported as an output; `test.yml` passes `run-ratchet` from it.

### Types of changes

### Related issue(s):

1. DEV-2765
2. The enforcement-tightening spec (sleep-debt burn-down, CI-flakiness investigation) — tooling only, no library issue.

### Affected project(s):

- [x] `handsontable` (tooling, docs, and a config comment only — no library source)
- [ ] `@handsontable/vue3`

### Checklist:

- [x] My change requires a change to the documentation. (Done in this PR: `.ai/LOCAL-ENFORCEMENT.md`, `AGENTS.md`, `handsontable/.ai/CONVENTIONS.md`, `handsontable/.ai/TESTING.md`.)
- [ ] This change needs a manual QA pass — no; the gate is proven by its own tests and the scratch runs above. Note: this PR edits no file under `handsontable/src/**` or `handsontable/test/**`, so its own ratchet step in `Lint / core` is skipped by the scope gate; the first live CI run is the next PR that edits a spec or unit file.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2765

---

> [!NOTE]
> **Medium Risk**
> Changes merge-blocking CI and pre-push behavior for spec edits; misconfigured bases or diff parsing could block or miss findings, though the design explicitly skips on tooling gaps.
> 
> **Overview**
> Adds a **determinism ratchet** so new test anti-patterns cannot slip in while legacy Jasmine debt stays at ESLint `warn`. For changed `handsontable` `*.spec.js` / `*.unit.{js,ts}` files, `lint-ratchet.mjs` diffs added lines against ESLint warnings for `no-fixed-sleep-in-spec`, `no-new-it-flaky`, and `no-skipped-test`; intersections **block** pre-push and CI even though full `npm run lint` still exits 0 on existing debt.
> 
> The gate is shared end-to-end: pure logic in `lint-ratchet.mjs`, CLI with merge-base diffs, rename-aware git, and **fail-open** skips when tooling cannot run. **Pre-push** runs it after ESLint when ratcheted paths change; **Lint / core** adds a PR-only, path-scoped step that fetches the base branch live tip (`GATE_BASE=origin/<base.ref>`, not payload `base.sha`) and uses full `fetch-depth` only when the ratchet runs. `checks.yml` exports a `lint-ratchet` scope flag; docs and ESLint comments describe moved-line vs rename behavior and how to fix or exempt findings.
> 
> Ships with broad `node --test` coverage (lib, CLI fixtures, workflow text assertions) and updates to `pre-push` tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f44e5801f91d4154cc86b9e57bc43939ab419d40. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2765 Diff-scoped determinism ratchet: warn-level fixed waits on ADDED spec lines fail pre-push and CI](https://app.clickup.com/t/9015210959/DEV-2765)

.github/workflows/lint.yml
```diff
@@ -32,3 +37,24 @@ jobs:
...
+  # base, and needs the base's history (fetch-depth 0), not a build. No token,
+  # no secret: safe on fork and Dependabot PRs without a guard.
+  determinism-ratchet:
+    name: determinism ratchet
+    runs-on: ubuntu-latest
+    if: inputs.run-ratchet && github.event_name == 'pull_request'
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # https://github.com/actions/checkout/releases/tag/v6.0.2
+        with:
+          fetch-depth: 0
```
budnix: On a pull_request run this takes `refs/pull/N/merge`, and `base.sha` is already an ancestor of that merge commit, so the CLI's `merge-base(base.sha, HEAD)` returns `base.sha` unchanged and the diff ends up being base-tip to merge-commit. That equals the branch's own changes only while `base.sha` is still the develop tip the merge ref was built from, and a "Re-run all jobs" replays the old payload against a merge ref GitHub has since rebuilt, so every develop commit in between then reads as a line this branch added. The other three `GATE_BASE` gates share this pattern, but they are all warn-only, so this is the first one where a stale base can actually block someone. Curious whether pinning the head SHA is all it takes here, since `merge-base` would then find the real fork point no matter how stale `base.sha` is.

```suggestion
      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # https://github.com/actions/checkout/releases/tag/v6.0.2
        with:
          ref: ${{ github.event.pull_request.head.sha }}
          fetch-depth: 0
```

.github/workflows/lint.yml
```diff
@@ -32,3 +37,24 @@ jobs:
...
         run: npm run in visual-tests lint
+
+  # The frozen Jasmine suite's determinism rules (no-fixed-sleep-in-spec,
+  # no-new-it-flaky, no-skipped-test) are `warn`, and `npm run lint` above exits
+  # 0 with warnings. This job is the ratchet: the same rules on a line the PR
+  # ADDED fail; pre-existing debt stays a warning. The same script blocks in the
+  # pre-push hook, so local scope == CI scope. PR-only — it diffs against the PR
+  # base, and needs the base's history (fetch-depth 0), not a build. No token,
+  # no secret: safe on fork and Dependabot PRs without a guard.
+  determinism-ratchet:
```
budnix: This job pairs a full-history checkout with `setup-workspace`, so a whole workspace `pnpm install` runs just to hand ESLint a handful of spec files, and the `core` job 20 lines up already has that install warm and only lacks history. Because `build` has `needs: [lint]`, the extra job also sits on the critical path for the rest of the DAG. Adding `fetch-depth: 0` to `core` and folding the ratchet in as a step there would win back one runner and one install, at the price of a deeper clone on every lint run. Did you already weigh that, or was the separate job mostly about keeping the scope gate independent?

.github/scripts/lib/lint-ratchet.mjs
```diff
@@ -0,0 +1,256 @@
...
+ * `lint:eslint` task covers — so the local run stays a subset of CI's lint
+ * scope (`dist/` and `tmp/` are build output and never linted). The Playwright
+ * tier (`tests/e2e/*.spec.ts`) is deliberately absent: its own config bans
+ * `sleep()` / `waitForTimeout()` at `error`, so there is nothing to ratchet.
+ *
+ * @type {readonly RegExp[]}
+ */
+export const RATCHETED_FILES = Object.freeze([
+  /^handsontable\/(src|test)\/.*\.(spec|unit)\.js$/,
+  /^handsontable\/(src|test)\/.*\.unit\.ts$/,
```
budnix, line 52: I put the same seven-line probe at `probe.unit.js` and `probe.unit.ts` and ran the package's own ESLint over both: the `.js` copy reported `no-fixed-sleep-in-spec` and `no-skipped-test`, and the `.ts` copy reported nothing at all, with no parse error either. The override that turns those three rules on is `files: ['*.unit.js', '*.spec.js']` in `handsontable/.eslintrc.js` and `*.unit.ts` is not in it, so a new `sleep()` in any of the 217 `.unit.ts` files under `handsontable/{src,test}/` is not blocked, while LOCAL-ENFORCEMENT.md says it is. Dropping `.unit.ts` from here and from the docs is the simpler fix; adding `*.unit.ts` to that override also works, at the cost of 11 new warnings in `function.unit.ts` where the fixed wait is the point of the test. Whichever way you go, the `every ratcheted rule is configured` test flattens every override into one map, so it only proves the ids exist somewhere and cannot catch a file-glob mismatch like this one.

.github/scripts/lib/lint-ratchet.mjs
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  return p.slice(1, -1).replace(/\\([0-7]{3}|[\\"tnr])/g, (_, esc) => {
+    switch (esc) {
+      case 't': return '\t';
+      case 'n': return '\n';
+      case 'r': return '\r';
+      case '"': return '"';
+      case '\\': return '\\';
+      default: return String.fromCharCode(parseInt(esc, 8));
```
budnix: `\303\251` is one UTF-8 character spread over two bytes, so decoding each escape on its own gives `Ã©` instead. I tried it against real git: `parseAddedLines` keyed the file as `cafÃ©.spec.js` while `git diff --name-only -z` gave `café.spec.js`, so `candidateSet.has(file)` missed and every added line in that file was dropped. Nothing is broken today since no spec has a non-ASCII name, but `-c core.quotePath=false` on the git calls makes git print the raw path, which I checked too and the key then matched, and it lets `unquotePath` and the unquoting call in `newSidePath` go away completely. The test at line 207 only covers `\t`, which is ASCII, so it stays green either way.

.github/scripts/lib/lint-ratchet.mjs
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    for (const m of messages) {
+      if (m?.ruleId && ruleSet.has(m.ruleId) && added.has(m.line)) {
+        findings.push({ file, line: m.line, ruleId: m.ruleId, message: m.message });
+      }
+    }
+  }
+
+  return findings.sort((a, b) => (a.file < b.file ? -1 : a.file > b.file ? 1 : a.line - b.line));
```
budnix: Worth knowing that nothing lints this tree right now: `eslint .github/scripts/ -f json` reports exactly one file, `check-changelog.js`, because the root script passes directories with no `--ext` and ESLint 8 only picks up `.mjs` where an `overrides[].files` glob names it, which `scripts/**/*.mjs` does and `.github/scripts/**` does not. `lintable()` in `scripts/lint-files.mjs` drops dot-segment paths as well, so pre-commit and pre-push skip it too, which means the roughly 1,150 new lines in this PR are unlinted and the `npm run eslint` line in the description does not reach them. Fed the config that would apply, this file has four errors: `no-nested-ternary` on this sort comparator and three `jsdoc/valid-types` on the `readonly string[]` types. Clearly a follow-up rather than yours to fix here, since the whole directory has 163 errors on develop today, but adding `.github/scripts/**/*.mjs` to that existing override would close it in one line.

.github/scripts/__tests__/lint-ratchet-cli.test.mjs
```diff
@@ -0,0 +1,299 @@
...
+
+  assert.equal(run.status, 0, `exit ${run.status}\n${run.stdout}${run.stderr}`);
+  assert.match(run.stdout, /no merge-base/);
+  assert.match(run.stdout, /skipped/);
+  assert.deepEqual(reportedLocations(run.stdout), [], 'the legacy sleep() must not be read as added');
+});
+
+// --- control: the gate still bites on a plain in-place addition ---
+
+test('a new sleep() added in place (no rename) is reported at its line', () => {
```
budnix: The base-resolution branches are covered well here, but the three tooling-gap branches are not: ESLint binary missing, ESLint exiting 2, and unparsable JSON. Those are the ones carrying the "never a false block" promise, and they break invisibly, since flipping the `existsSync` guard or moving the `status === 2` check below `JSON.parse` would make every push fail with no test to catch it. `createFixtureRepo()` already writes the stub at `handsontable/node_modules/eslint/bin/eslint.js`, so deleting that file, swapping in a stub that exits 2, and one that prints a bare `{` would cover all three on top of the commit this test already builds. Would be good to pin them, since LOCAL-ENFORCEMENT.md section 2 holds the enforcement machinery to the same test rule as everything else.

handsontable/.eslintrc.js
```diff
@@ -235,6 +235,10 @@ module.exports = {
         // existing sleep()/it.flaky() debt must surface without red-walling CI.
         // Escalation to error happens in the flip-to-blocking task once the
         // debt is burned down. New E2E belongs in Playwright (tests/e2e).
+        // A NEW occurrence on a line a branch adds is already blocked: the
+        // diff-scoped ratchet (.github/scripts/lint-ratchet.mjs, pre-push + CI
+        // lint) fails on these three warn rules wherever they hit an added line.
+        // Keep its RATCHETED_RULES list in step with the levels here.
```
budnix, line 241: Six lines below this, the skip comment still says new skips are caught by the diff-based test-weakening detector. That detector is non-blocking in both pre-push and CI and works off a before/after count, so it never stops a push. Now that `no-skipped-test` is in `RATCHETED_RULES`, an added `it.skip(` line exits 1 at pre-push and reds the CI job, which is a much bigger surprise for whoever hits it, and the same one-sentence pointer you added above the sleep rules would cover it. Commenting up here because line 247 falls outside this hunk.

.github/workflows/lint.yml
danielzytohoc: Right on the shape — and pinning the head SHA alone would not have covered it: a branch that merged develop after the payload's base.sha still has that SHA as an ancestor, so merge-base(base.sha, head) is base.sha again. The step now fetches refs/heads/<base.ref> and passes GATE_BASE=origin/<base.ref>; against the live tip the merge-base is the merge ref's base parent (or the last base commit the branch merged in), so the diff is the PR's own lines in both shapes, and the checkout can stay refs/pull/N/merge so `npm run lint` keeps linting the merge result. lint-ratchet-cli.test.mjs pins both shapes against real merge commits (live tip → exit 0, frozen fork SHA → exit 1 on the base's line), lint-ratchet-workflow.test.mjs pins that lint.yml hands the CLI that tip and nothing else, and a failed refresh of the base branch is a ::warning:: rather than a red job.

.github/workflows/lint.yml
danielzytohoc: It was about keeping the scope gate independent, and that does not need a job: the ratchet is now a step of `core` behind the same `inputs.run-ratchet && github.event_name == 'pull_request'` condition, and the checkout takes `fetch-depth: 0` only under that same expression, so a lint run with no spec change keeps the shallow clone. One runner and one install saved, and the step sits after `npm run lint` in the job that `build` already waits for. lint-ratchet-workflow.test.mjs pins the step, its gate, the live-tip base and that the depth condition matches the step's condition verbatim.

.github/scripts/lib/lint-ratchet.mjs
danielzytohoc: Kept `*.unit.ts` in the candidate set — with no override naming it ESLint reports none of the three rules there, so it can never block — and fixed the docs and the JSDoc to say it is covered the moment the frozen-tier override names `*.unit.ts` (that override change is in flight in a separate PR, so it is not touched here). The config test now resolves each ratcheted rule per path, the way ESLint applies overrides (basename globs, in order, last one wins, excludedFiles honored), for a .spec.js and a .unit.js path in each tree, and asserts `warn`; the resolver has its own test. A `files` glob that stops matching the frozen tier fails it, which the flattened map could not see.

.github/scripts/lib/lint-ratchet.mjs
danielzytohoc: Confirmed against real git: the pre-fix CLI keyed `café.spec.js` as `cafÃ©.spec.js` and exited 0 with 'gained no line'. Git now runs with `-c core.quotePath=false`, `unquotePath` and its call site are gone, and a header git still C-quotes (a control character, quote or backslash) is left unattributed rather than decoded — never a false block. Added a CLI test with `café.spec.js` that first asserts git's default output quotes the name, then that the ratchet reports `café.spec.js:5` and exits 1, plus lib tests for the raw key and the still-quoted header.

.github/scripts/lib/lint-ratchet.mjs
danielzytohoc: Reproduced your four with the root config plus `.github/scripts/**/*.mjs` on the scripts override (`npx eslint --no-ignore --no-eslintrc -c probe.cjs .github/scripts/lib/lint-ratchet.mjs`, the `--no-ignore` because ESLint skips dot-directories by default): 3× jsdoc/valid-types, 1× no-nested-ternary. Fixed — `ReadonlyArray<string>` / `ReadonlyArray<RegExp>` in the JSDoc types and a named `byFileThenLine` comparator — and the CLI and all three ratchet test files are clean under that config too, so wiring the directory into lint (the separate backlog ticket) will not turn them red.

.github/scripts/__tests__/lint-ratchet-cli.test.mjs
danielzytohoc: Added the three: the stub deleted (skips with 'not installed'), a stub that writes an ESLint-style stderr and exits 2 (skips, and the notice carries the first stderr line), and a stub printing a bare `{` (skips with 'no parsable JSON') — each asserting exit 0, the skip message and no reported location. Each starts from the in-place `sleep(50)` commit the control test proves is blocked, so the exit 0 is provably a skip of a real finding rather than an empty run.

handsontable/.eslintrc.js
danielzytohoc: Updated the skip comment: a new skip on a line a branch adds is blocked by the same diff-scoped ratchet as the sleep rules above (exit 1 at pre-push, red in the CI lint job), replacing the pointer to the non-blocking weakening detector.

budnix (dismissed):

budnix (approved):

PR #13351, opened 2026-09-02
DEV-2767: Add advisory presence-gate warnings and scorer smells

danielzytohoc:
### Context
The PR adds a WARN-only advisory layer to the test-presence gate and two heuristic smells to the eval scorer. Nothing here blocks: the gate's exit code is unchanged in both `GATE_MODE`s (pre-push `block` included), and the advisories print below the verdict as a "non-blocking" Markdown section — plus `::warning` annotations in GitHub Actions.

Why: the gate proves a test *exists*, but three shapes slipped through it unseen. A modified `*.spec.js` counts as coverage, so `it` blocks appended to the frozen Jasmine suite were invisible. RTL logic landed in source with no RTL-aware test. Walkontable engine changes were "covered" by a core-tier spec that never pins the overlay math. And the PR template's "spec that fails without this fix" line stayed empty on bug fixes.

**Presence-gate advisories** (`.github/scripts/lib/presence-warnings.mjs`, pure functions; CLI wiring in `test-presence-gate.mjs`):
1. **frozen-suite-growth** — ≥3 new `it(`/`it.each(`/`fit(` blocks on added lines of *modified* frozen Jasmine specs (a *new* spec is already blocked). Message steers to `tests/e2e` and asks for a justification.
2. **red-spec-field** — the "Bug fix" box is ticked and the template's red-spec line is empty/placeholder. An answer written on the line below counts (free text or a nested list item); a sibling template item or the next heading does not. CI only: the `presence` job now fetches the *live* PR body (same mechanism as the Manual QA job, read-only so it works on the fork/Dependabot token) into `GATE_PR_BODY_FILE`; that step carries `continue-on-error: true`, so an action-runtime failure skips one advisory check and can never fail the job (or, through `test.yml`'s `needs`, the CI Gate). Locally the check is skipped silently. HTML comments are stripped with an index scanner that re-scans from the splice seam, so `<!<!-- -->--` leaves no `<!--` behind (CodeQL `js/incomplete-multi-character-sanitization`).
3. **rtl-correlation** — added source lines match `isRtl|layoutDirection` and no added *test-side* line matches `/rtl|layoutDirection/i`. Test-side is a file the gate classifies as `test` plus anything under `tests/**` — a Playwright page object or helper is `neither` to the gate, yet it is where `initGrid({ layoutDirection: 'rtl' })` usually lands. The CLI's diff pathspec is built from `isAdvisoryPath()` (classifier `source`/`test` **plus `tests/**`**), so those files reach the detector: a review probe showed the earlier classifier-only pathspec dropped every page object (19 `tests/**` code files across the last 40 develop commits), and the warning fired on a change the page object had paired. `presence-gate-cli.test.mjs` runs the CLI against a throwaway git repository with the env the `presence` job sets and pins both directions — a page-object RTL line silences the warning, a silent page object does not — plus that a page object is still not coverage for the verdict (`❌` and exit 1 in block mode, unchanged).
4. **walkontable-routing** — `handsontable/src/3rdparty/walkontable/src/**` changed with nothing under `walkontable/test/**` or `tests/e2e/walkontable/**`. Deliberately requires `classify(path) === 'source'` and ignores a `D` status (a deletion needs no new coverage) — stated in the lib header.

Every gap in the input (no body, no diff, a diff that will not fit the buffer) is silence, never a finding. The diff read is pathspec-limited to source, test, and `tests/**` files with `--unified=0` — no build, no full-tree diff.

**Scorer smells** (`evals/score.mjs`):
- `theme-sensitive-viewport` (in `determinismSmells`) — a rendered-count read inside a describe whose grid setup pins no viewport. A read is one of the legacy helpers **by exact name** (`countVisibleRows()`/`countVisibleCols()`, `countRenderedRows()`/`countRenderedCols()`, `getRenderedRowsCount()` — a look-alike such as `countVisibleCustomBorders()` counts drawn borders and does not read) or a `:visible` selector that something counts (`toHaveCount(`/`.count()` on the selector or on the locator it is captured into — a `:visible` click or `.first()` counts nothing). A pin is `scrollViewportTo`, or a **top-level `width`/`height` of an options object** handed to a call — `handsontable({ … })`, `grid.initGrid({ … })`, `new Handsontable(host, { … })`, or a local passed to one whole or spread (`initGrid(ROOMY_VIEWPORT)`, `initGrid({ ...ROOMY_VIEWPORT })`). A nested `width` (`border: { width: 2 }`, `columns: [{ width: 100 }]`) is not the grid's size, an expected value inside an assertion (`toEqual({ width: 2, color: '#548235' })`) is not a setup, and a TypeScript type (`Promise<{ width: number }>`) is not a value. An outer describe's setup covers nested ones; comments are blanked first. Measured over the 69 specs in `tests/e2e/`: before, 1 file / 5 reads (`customBorders.spec.ts` — 31 `countVisibleCustomBorders()` matches, 26 of them hidden by `width:` false pins — 14 by the `border: { width: 2, color: '#548235' }` on line 442 alone — and 5 surfacing and marking the file suspect); after, 0 files / 0 reads.
- `unasserted-capture` (new `structureSmells` field) — `const x = await …` in a test body whose value never reaches `expect(…)`/`assert…(…)`, its matcher chain, or an `x.expect…(` helper — following **one level of derivation**: a later `const y = <expression using x>` counts as a use when `y` is asserted (`const tokens = String(className).split(' '); expect(tokens)…`; `const borders = await lab.cellBorders(2, 2); expect(borders?.top)…`). Each capture is reported once, under its innermost test — a regex literal with a paren (`/inset\(/`) makes the bracket scanner run one body over the tests after it, which double-counted their captures before. Reported by name and test title. The identifier is escaped for every regex metacharacter before it becomes a pattern (CodeQL `js/incomplete-sanitization`). Measured over `tests/e2e/`: before, 7 files / 15 captures, each marking a shipped spec suspect; after derivation and the dedupe, 3 files / 4 captures (`column-move-sorting` `label`, `dropdown-height` `optionCount`, `overlay-scrollbar-clearance` `box` and `point`) — every one a value fetched to drive an action (a bounding box for a pointer move, a count for a keyboard loop) whose outcome the test asserts by other means. Those are false positives for the smell's purpose, so **it feeds `warnings`, not `problems`, until its precision is measured**; suspect files in `tests/e2e/` went from 20 to 14, and the remaining 14 are hollow tests, a `fixme`, and a `sleep(`, all pre-existing signals.

`run-eval.mjs` scores each fixture's `counterexamples/` directory (files named `<scenario>.<smell>.spec.ts`, the convention the setTimeout-scorer change introduced on `develop`) and fails the harness self-test when a counterexample does not trip exactly the smell its name declares — a problem, or a warning-tier smell such as `unasserted-capture` — so each documented smell is proven to fire, not just described. Two fixtures added (`e2e-rendered-rows-viewport`, `e2e-unasserted-capture`), each with a clean reference and a smelly counterexample; the three existing references still score clean.

Docs: `.ai/LOCAL-ENFORCEMENT.md` (the advisory layer, the `tests/**` pathspec and its CLI test, and how to add a detector), `evals/README.md` (fields, fixture layout, the warning-tier smell and its measured numbers), `evals/fixtures/e2e-unasserted-capture/case.md`, `.claude/skills/handsontable-e2e-testing/SKILL.md` (the frozen-suite-growth advisory).

Tooling/test-infrastructure only — no changelog entry (the changelog gate is path-aware and passes automatically).

- Unit tests added/modified (`*.unit.js`): none — no `handsontable/src` or `wrappers` change
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`): none — the two new `evals/fixtures/**/*.spec.ts` are scorer fixtures, scored statically
- Type tests (`*.types.ts`) updated if public API changed: n/a
- For a bug fix — the spec that fails without this fix: n/a (tooling change; each detector's `node --test` case was written first and failed with a missing-module / missing-export error before the implementation)

Machinery tests: `.github/scripts/__tests__/presence-warnings.test.mjs` (new, 27 cases on synthetic diffs/bodies/name-status lists, including `isAdvisoryPath`, plus one text-based case pinning the `presence` job's body step: `continue-on-error: true` and `GATE_PR_BODY_FILE` handed to the gate), `.github/scripts/__tests__/presence-gate-cli.test.mjs` (new, 4 cases running the CLI against a throwaway git repository with CI's env: page-object pairing through the real `git diff` path, its silent control, a docs mention that does not pair, and the unchanged blocking verdict), `evals/__tests__/score.test.mjs` (+16 cases: the customBorders shapes, nested and expected-value widths, the options-object contexts, derivation, the dedupe, the warning tier), `evals/run-eval.mjs` self-test (5/5 references clean, 2/2 counterexamples caught).

**Merge note (develop merged after the matcher-downgrade detector landed):** the scorer's `testBodies()` — which both new smells scan — now skips an `it.each` table the same way `extractTestBlocks()` does, because the shared `TEST_CALL_RE` from develop matches a parameterized opener at the table's `(`; without the skip an `it.each` body was scanned as its table, so an awaited capture inside it went unreported and a `:visible` capture fell back to describe scope. Two tests pin it (both fail on the raw auto-merge). Nothing else in the merge goes beyond a textual union.

### How has this been tested?
Each detector's `node --test` case was written first and watched fail (`ERR_MODULE_NOT_FOUND` / "does not provide an export named …") before the implementation; the viewport counterexample then exposed a comment-matching false negative that run-eval caught, fixed with a regression test. The review fixes followed the same order: the `stripHtmlComments` case failed on the missing export, the next-line red-spec answer read as empty (`true` where `false` was expected), the page-object pairing still warned, the `continue-on-error` shape test failed on the missing key, and the bare `:visible` click test counted 3 reads where 0 were expected — each went green only with its fix. The `$`-identifier capture test was green before and after (an identifier can carry no other metacharacter); it pins the case the full escape must keep working.

Second review round: the CLI test's pairing case and verdict case were run against a filter that drops `tests/fixtures/**` and failed (`not ok 1`, `not ok 4`) before the `isAdvisoryPath` pathspec made them pass; the derivation, dedupe, and options-object cases were written from the shapes in `tests/e2e/hiding-array-classname.spec.ts`, `overlay-scrollbar-clearance.spec.ts`, `customBorders.spec.ts`, and `selection-handles.spec.ts`, and the TypeScript-type case caught a regex backtracking hole (`\s*` outside the lookahead let `{ width: number }` pin) that was fixed before it went green. Both smells were measured over `tests/e2e/*.spec.ts` before and after (numbers above). Commands run and their final lines:
```
node --test .github/scripts/__tests__/presence-warnings.test.mjs   → # pass 27 / # fail 0
node --test .github/scripts/__tests__/presence-gate-cli.test.mjs   → # pass 4 / # fail 0
node --test .github/scripts/__tests__/presence-gate.test.mjs        → # pass 18 / # fail 0
node --test evals/__tests__/score.test.mjs                          → # pass 34 / # fail 0
npm run test:tooling                                                → # tests 409 / # pass 409 / # fail 0
node evals/run-eval.mjs                                             → References: 5/5 meaningful — harness self-test PASSED. Counterexamples: 2/2 caught (a problem, or a warning-tier smell). (exit 0)
npm run eslint                                                      → clean (exit 0)
GATE_BASE=<develop~40> GATE_MODE=block GITHUB_ACTIONS=true node .github/scripts/test-presence-gate.mjs
  → "✅ Pass." + advisory section (frozen-suite-growth: 38 new `it` blocks in 9 specs; walkontable-routing: 2 engine files)
    + two ::warning annotations on stderr → exit 0  (existing debt surfaces, never blocks)
GATE_BASE=$(git merge-base origin/develop HEAD) GATE_MODE=warn GITHUB_ACTIONS=true GATE_PR_BODY_FILE=<live body> node .github/scripts/test-presence-gate.mjs
  → this branch, the way the presence job runs it: "✅ Pass.", no advisory, exit 0 — and the same in GATE_MODE=block
GATE_BASE=origin/develop GATE_MODE=block GITHUB_ACTIONS=true GATE_PR_BODY_FILE=<body> node .github/scripts/test-presence-gate.mjs
  → body with the answer on the next line behind a `<!<!-- -->--` placeholder: "✅ Pass.", no advisory, exit 0
  → body with the template placeholder only: "✅ Pass." + red-spec-field advisory + one ::warning, exit 0
```
Workflow change: the `presence` job ran green with the new body-reading step on this PR's own run (`Checks / test presence`, run 33656148793); the step now also carries `continue-on-error: true`, pinned by the text-based test above.

### Types of changes

### Related issue(s):
1. The enforcement-tightening spec (advisory warnings, never blocking)
2.
3.

### Affected project(s):
- [ ] `handsontable`
- [ ] `@handsontable/vue3`

(none — CI/tooling and the eval harness only)

### Checklist:
- [x] My change requires a change to the documentation. (done in this PR: `.ai/LOCAL-ENFORCEMENT.md`, `evals/README.md`, `evals/fixtures/e2e-unasserted-capture/case.md`, `.claude/skills/handsontable-e2e-testing/SKILL.md`)

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2767

---

> [!NOTE]
> **Low Risk**
> Changes are CI/tooling and eval harness only; advisory paths fail open on missing inputs and cannot block merges.
> 
> **Overview**
> Adds **non-blocking advisory warnings** to the test-presence gate: after the existing pass/fail verdict, the CLI prints heuristic nudges (frozen Jasmine suite growth, empty bug-fix “red spec” field in CI, RTL source/test mismatch, Walkontable engine changes without engine-tier tests). Exit codes stay unchanged; CI can emit `::warning` annotations and reads the **live** PR body for the red-spec check via `GATE_PR_BODY_FILE`.
> 
> Implements detectors in **`presence-warnings.mjs`** and widens the advisory `git diff` pathspec with **`isAdvisoryPath()`** so `tests/**` helpers/page objects count for RTL pairing without treating them as coverage for the blocking verdict.
> 
> Extends the **eval scorer** with **`theme-sensitive-viewport`** (unpinned rendered-row counts) and warning-tier **`unasserted-capture`**, plus two new fixture cases and updated counterexample harness rules for structure smells. Docs/skills (`LOCAL-ENFORCEMENT`, `evals/README`, legacy E2E skill) describe the new behavior; **`node --test`** coverage includes unit tests for detectors and CLI plumbing against throwaway repos.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b48a77eec330d99070c1a7bffc4361cae5080abe. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2767 Presence-gate warns and scorer heuristics: frozen-suite growth, red-spec field, RTL correlation, walkontable routing, viewport and unasserted-capture smells](https://app.clickup.com/t/9015210959/DEV-2767)

.github/scripts/test-presence-gate.mjs
```diff
@@ -52,6 +64,74 @@ function readTrailers(range) {
...
+ *
+ * @param {string} range The diff range, e.g. `<base>...HEAD`.
+ * @param {{status: string, oldPath: string, path: string}[]} changes Parsed diff entries.
+ * @returns {string} The unified diff, empty when nothing relevant changed.
+ */
+function readDiff(range, changes) {
+  const paths = new Set();
+
+  for (const change of changes) {
+    if (change.status !== 'D' && classify(change.path) !== 'neither') {
```
budnix: `classify()` returns `'neither'` for every non-spec file under `tests/**`, so this filter drops page objects and helpers from the pathspec and they never reach `parseUnifiedDiff`. That is the one file `isTestSide()` in `presence-warnings.mjs` was written to accept - the test at `presence-warnings.test.mjs:268` and the `tests/**` line in `LOCAL-ENFORCEMENT.md` both promise that a page-object change pairs an RTL source change, but in CI the detector never sees it and warns anyway. Over the last 40 commits on develop, 19 `tests/**` code files changed and every one would be dropped here, so this looks worth a fix before the RTL warning goes live.
```suggestion
    if (change.status !== 'D' && (classify(change.path) !== 'neither' || change.path.startsWith('tests/'))) {
```

evals/score.mjs
```diff
@@ -247,9 +247,334 @@ export function findGamingSignals(src) {
...
+ * `test.describe.serial(`). `describe.each([…])(` is excluded — its first paren
+ * holds the table, not the suite body.
+ */
+const DESCRIBE_CALL_RE = /(?<![\w$.])(?:[xf]describe|(?:test\.)?describe(?:\.(?!each\b)\w+)*)\s*\(/g;
+
+/**
+ * A rendered-count read through the legacy-suite helpers: `countVisibleRows()`,
+ * `countRenderedCols()`, `getRenderedRowsCount()`.
+ */
+const RENDERED_COUNT_HELPER_RE = /(?<![\w$])(?:countVisible\w*|countRendered\w*|getRendered\w*Count)\s*\(/g;
```
budnix: `countVisible\w*` matches any helper name that starts with `countVisible`, not just row and column counts. In `tests/e2e/customBorders.spec.ts` it matches 31 times, and every match is `countVisibleCustomBorders()` - a count of drawn borders, which has nothing to do with theme row height. `VIEWPORT_PIN_RE` on line 282 then leans the other way and reads the `border: { width: 2, color: '#548235' }` on line 442 as a pinned viewport, which hides 14 of them on its own; 5 of the 31 still come out as `theme-sensitive-viewport` and the file scores `suspect`.
```suggestion
const RENDERED_COUNT_HELPER_RE = /(?<![\w$])(?:countVisible(?:Rows|Cols|Columns)|countRendered(?:Rows|Cols|Columns)|getRendered(?:Rows|Cols|Columns)Count)\s*\(/g;
```

evals/score.mjs
```diff
@@ -468,6 +821,13 @@ export function scoreTestSource(src, options = {}) {
     });
   }
 
+  if (structureSmells.length > 0) {
+    problems.push({
+      type: 'structure-smells',
```
budnix, line 1040: `findUnassertedCaptures` follows the captured name only, so a value that gets reshaped into a second local and then checked looks unasserted. `tests/e2e/hiding-array-classname.spec.ts:62` does exactly that - `const className = await ...` becomes `tokens`, and `tokens` is what `expect()` receives. Running the detector over the 72 specs in `tests/e2e/` flags 7 files and 15 captures, and because this goes into `problems` rather than `warnings` each of those shipped specs now scores `suspect`. Would you consider following one level of derivation, or leaving this as a warning until the precision is measured?

.github/scripts/test-presence-gate.mjs
danielzytohoc: Confirmed — 19 tests/** code files across the last 40 develop commits, and every one was dropped before parseUnifiedDiff. The pathspec now comes from a new exported isAdvisoryPath() (classifier source/test plus tests/**, the same set isTestSide() accepts), and a new presence-gate-cli.test.mjs runs the CLI against a throwaway git repository with the presence job's env: a page-object layoutDirection: 'rtl' line silences the warning, a silent page object keeps it, a docs mention does not pair, and a page object is still not coverage for the verdict (❌ and exit 1 in block mode). Against a filter that drops tests/fixtures/** the pairing and verdict cases fail, so the test is falsifiable; presence-gate.test.mjs is untouched and green.

evals/score.mjs
danielzytohoc: Reproduced the 31/14/5 on customBorders.spec.ts and the file scoring suspect. Both regexes are tightened: the helper read is the exact-name alternation you suggested (shared with COUNT_SHAPE_RE), and the pin is no longer any width: key but a top-level width/height of an object literal handed to a call — handsontable({…}), grid.initGrid({…}), new Handsontable(host, {…}), or a local passed whole or spread — with anything inside an assertion span (toEqual({ width: 2, color: '#548235' })), a nested border/columns width, or a TypeScript type excluded. Measured over the 69 specs in tests/e2e: 1 file / 5 reads before (all customBorders, the other 26 hidden by width: false pins), 0 / 0 after; tests cover each customBorders shape and the harness self-test stays 5/5 references clean, 2/2 counterexamples caught.

evals/score.mjs
danielzytohoc: Reproduced the 7 files / 15 captures. The detector now follows one level of derivation — a later const whose initializer uses the capture counts when that local is asserted, which clears className→tokens, layouts→overflowing, and the five customBorders lab handles — and reports each capture once under its innermost test (a /inset\(/ regex in overlay-scrollbar-clearance made one body swallow the next tests, so box and point were counted twice). That leaves 3 files / 4 captures (label, optionCount, box, point), each a value fetched to drive a pointer move or keyboard loop whose outcome the test asserts otherwise, so unasserted-capture is a warning until precision is measured: no shipped spec is suspect because of it (20 → 14 suspect files, the rest hollow/fixme/sleep), and run-eval still fails when a counterexample trips nothing.

budnix (dismissed):

budnix (approved):

PR #11021, opened 2024-06-12
Prevent the table from creating an endless ResizeObserver loop when the container parents are using `dvh` as size values.

jansiegel:
### Context
This PR:
- Adds logic to disconnect the `ResizeObserver` when there were too many triggers from it in direct succession
    It also adds a warning message describing the potential issue.
    The window resizing is excluded from that logic.
- Adds a test case for the new logic

### How has this been tested?
Added a test case for the new logic.

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature or improvement (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Additional language file or change to the existing one (translations)

### Related issue(s):
1. handsontable/dev-handsontable#1898

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/angular`
- [ ] `@handsontable/react`
- [ ] `@handsontable/vue`
- [ ] `@handsontable/vue3`

### Checklist:

- [x] I have reviewed the guidelines about [Contributing to Handsontable](https://github.com/handsontable/handsontable/blob/master/CONTRIBUTING.md) and I confirm that my code follows the code style of this project.
- [x] I have signed the [Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLScpMq4swMelvw3-onxC8Jl29m0fVp5hpf7d1yQVklqVjGjWGA/viewform?c=0&w=1)
- [ ] My change requires a change to the documentation.

PR #13355, opened 2026-09-03
DEV-2740: Anchor the ResizeObserver loop guard to frames

demtario:
### Context

The PR fixes the runaway-`ResizeObserver` guard in `handsontable/src/3rdparty/walkontable/src/overlay/resizeMonitor.ts`, which had two defects.

**It could not trip on a loaded machine.** The guard counted `ResizeObserver` deliveries in direct succession and zeroed that count after 100 ms of wall-clock quiet. A `ResizeObserver` delivers at most once per rendering frame, so reaching the threshold required 300 consecutive frames each shorter than 100 ms. Under CPU contention Chrome stretches frames past 100 ms, the count reset before the threshold, and a genuine `dvh` feedback loop ran forever — the guard was disabled in exactly the machine state it exists for. The count is now reset only when a whole rendering frame passes with **no** delivery, which a self-sustaining loop cannot produce, so the trip depends on delivery cycles rather than on CPU speed.

The delivery has to be recorded **synchronously in the observer callback**. Within one frame the browser runs the animation-frame callbacks, then style and layout, then delivers the `ResizeObserver` callbacks; and a callback registered during a frame's animation-frame phase lands in the *next* frame's list, in registration order. So the quiet-frame watchdog — itself an animation-frame callback — runs before a callback registered by that same frame's observer delivery, and a flag set from the deferred part of the observer callback would be read one frame too early, calling a busy frame quiet. Only the `onContainerElementResize` fire stays deferred.

**The disconnect was permanent.** Nothing re-observed after the guard fired, so a grid legitimately resized 300+ times in a gap-free stream — dragging a splitter around the grid for a few seconds occupies every frame too — silently lost container-resize reactivity for the instance's lifetime. That shipped with the guard's origin (#11021, 2024-06). The disconnect is now a cooldown: the observer is observed again after a delay that doubles on every trip no quiet frame separated from the last one, capped at 30 s and returned to base on the first quiet frame. A legitimate grid resumes reacting; a page with a real loop is throttled to a bounded rate instead of frozen. The warning is printed once per instance.

Three details the review surfaced, all fixed here:

1. **The deferred `onContainerElementResize` fire was never cancelled on destroy.** Its animation-frame id was thrown away, so `destroy()` cancelled two of the three handles that outlive a task boundary. `getSetting()` on a function-valued key invokes the setting synchronously, so a delivery landing one frame before the teardown refreshed the dimensions of a torn-down grid; only core's own `isDestroyed` guard in `tableView.ts` kept that from throwing. All three handles are now cancelled, backed by an `#isDestroyed` flag — an observer entry carries the state from its own snapshot, so a delivery can be dispatched after the disconnect.
2. **A reconnect that found the wrapper detached gave up silently.** `observe()` is a no-op without a parent element, so a host that parks the grid's subtree outside the document — a framework re-render, a `keep-alive` cache, a tab or accordion that caches its panel — could land its detach exactly on the reconnect timer and restore the permanent disconnect through DOM timing alone. The reconnect now schedules another attempt when it cannot attach, riding the same backoff, so a wrapper that stays detached costs one no-op attempt per capped interval rather than a spinning timer.
3. **The warning text is reworded**, since it described the behavior that is being replaced. It now ends "The observer will be disconnected and reconnected after a short delay."; the source and both pinning specs moved together in one commit.

`observe()` deliberately does **not** reset the accumulated backoff, which `resetResizeCount()` does. The two are not symmetrical, and the JSDoc now says so: a window resize is evidence about the *cause* of the deliveries, so the succession they belong to can be forgotten, while a change of scrollable element says only which element to watch and nothing about whether the loop went away.

**Not a breaking change** — the reconnect adds recovery and removes nothing. One behavior change is worth a reviewer's attention: a grid that today survives on a loaded machine only because the wall-clock reset kept saving it will now trip and print a console warning it never printed before, then recover after the cooldown. That is the intended fix.

Two smaller notes:

- Timers and animation frames moved from the globals onto the injected `rootWindow`, matching `scrollbarVisibility.ts`. That makes the class unit-testable and correct for a grid whose document is an iframe's. Walkontable cannot use core's `_registerTimeout` — the architecture boundary forbids reaching `hot`.
- `nativeScrollInput.ts` still calls `resetResizeCount()` only on a 200 ms *trailing* debounce, so a multi-second window drag-resize accumulates the count and can trip the guard. Resetting on every `resize` event as well would remove that false-positive path, but mobile browsers fire `resize` while scrolling, so it needs its own decision. Left out of this PR, worth a follow-up.

The dvh loop-guard spec moved from the frozen Jasmine suite into Playwright, which is what its pointer comment in `afterRefreshDimensions.spec.js` asked for: it could not be made honestly green before this fix, because its only success signal is the guard tripping and there was no bound on how long a loaded runner takes to produce 300 gap-free deliveries. Its fixed `sleep(6000)` is gone — the spec polls until the warning appears, checks the count is frozen across a bounded run of frames inside the cooldown, then polls for the recovery.

### How has this been tested?

New unit coverage drives the observer, the animation frames and the timers by hand, so every case is deterministic:

```bash
npm run test:unit --prefix handsontable -- --testPathPattern=resizeMonitor
```

17 cases: the trip with every frame gap at 150 ms (the case the old guard fails), a quiet frame breaking the succession, recovery after the cooldown, warn-once plus the doubling backoff, the backoff returning to base after a quiet frame and after `resetResizeCount()`, the cancel/re-arm handoff when a window resize clears a succession, `observe()` cancelling a pending reconnect, the reconnect retrying while the wrapper is detached and attaching once it is back, and teardown leaving no reconnect, no watchdog and no deferred fire armed.

Every defect was reproduced against the unfixed code before its fix was accepted:

- With the 100 ms wall-clock reset put back, 5 of the unit cases fail, including the load case.
- With `resizeMonitor.ts` reverted to `develop` and the bundle rebuilt, the new Playwright spec fails on the recovery assertion (`expect(received).toBeGreaterThan(300)`) — the permanent disconnect.
- With the destroy guards removed, the three teardown cases fail.
- With the detached-wrapper retry removed, the detach case fails.

Full gates, all green on the final tree:

```bash
npm run test:unit --prefix handsontable            # 4328/4328
npm run test:walkontable --prefix handsontable     # 816/816
npm run test:e2e --prefix handsontable             # 9848 specs, 0 failures
cd tests && npx playwright test e2e/refresh-dimensions.spec.ts   # 48/48, all six theme x bundle legs
npm run eslint --prefix handsontable               # 0 errors
```

The Playwright spec was also run with `--repeat-each=3` on `e2e-main` (24/24) on top of the full six-leg run, since its assertions are timing-sensitive.

Manual check of the real defect: open the fixture page (`tests/fixtures/demo/refresh-dimensions.html`, `buildDvhGrid()`), throttle the CPU 6x in DevTools, and confirm the warning appears and the refresh rate collapses. On `develop` at that throttle the warning never appears at all.

### Types of changes

### Related issue(s):

1. DEV-2740

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2740

---

> [!NOTE]
> **Medium Risk**
> Changes hot-path container resize observation and loop-breaking behavior; grids that only avoided tripping due to the old wall-clock reset will now warn and enter cooldown, though they should recover after reconnect.
> 
> **Overview**
> **Fixes the Walkontable `ResizeMonitor` endless-loop guard** so it trips under CPU load and restores container resize handling after a trip instead of disconnecting forever.
> 
> The guard no longer resets its succession counter on a **100 ms wall-clock idle** timer (which failed when frames stretched past 100 ms). It now resets only after a **full rendering frame with no `ResizeObserver` delivery**, using a quiet-frame watchdog; deliveries are recorded **synchronously** in the observer callback so frame ordering does not false-clear busy frames. On trip (still **300** successive deliveries), the observer **disconnects temporarily**, warns **once** per instance (`warnOnce`), then **re-observes** after a cooldown (**2 s** base, doubling up to **30 s**); a quiet frame or `resetResizeCount()` (window resize) resets backoff.
> 
> **Lifecycle and realm fixes:** timers, animation frames, and `ResizeObserver` are taken from injected **`rootWindow`** (iframe-correct). `destroy()` cancels reconnect, watchdog, and deferred `onContainerElementResize` rAF; **`#isDestroyed`** guards late callbacks. **`observe()`** cancels pending reconnects, and reconnect retries when the wrapper parent is detached.
> 
> **Tests/docs:** large **`resizeMonitor.unit.ts`** harness; dvh loop-guard case moved from Jasmine to **Playwright** with fixture sampling; legacy **`afterRefreshDimensions.spec.js`** removed. **`rootWindow`** typing is **`Window & typeof globalThis`** in engine types. Changelog **#13355** and **AGENTS.md** document the guard.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bbbfdbcec8ec9ea09dd97ddaf7e6bbae4253e879. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2740 ResizeObserver loop guard never trips on a loaded machine, and a legitimate trip disconnects resize reactivity permanently](https://app.clickup.com/t/9015210959/DEV-2740)

tests/e2e/refresh-dimensions.spec.ts
```diff
@@ -219,3 +220,57 @@ test.describe('a grid inside an iframe (window-resize pipeline)', () => {
...
+
+      // The guard trips on the 300th successive delivery, and the frames already queued at that
+      // moment still run their deferred hook fire - so the total is at least the threshold, never
+      // exactly it.
+      expect(refreshesWhenTripped).toBeGreaterThanOrEqual(300);
+
+      // The observer is disconnected for the whole cooldown, so a bounded run of frames inside it
+      // must add nothing. The poll above is this negative check's positive control: the pipeline
+      // reached 300 deliveries a moment ago, so a frozen count here can only be the disconnect.
+      await grid.afterAnimationFrames(30);
```
budnix: Curious whether this 30-frame wait always lands inside the 2 s cooldown. By the time the loop trips, `expect.poll` is already on its 1000 ms interval, so the warning can be read up to a second late, and 30 frames in a row on a loaded runner add about another second. If those two add up past `RESIZE_LOOP_GUARD_RECONNECT_DELAY` the observer is back before the next line runs, and the `toBe` goes red on exactly the busy machine this PR is about. Recording the time alongside the warning in the fixture and checking it stayed under the cooldown would make that a real precondition.

handsontable/src/3rdparty/walkontable/src/overlay/resizeMonitor.ts
```diff
@@ -102,32 +168,212 @@ export class ResizeMonitor {
...
+   * not by a timer that polls for the whole session.
+   *
+   * It deliberately does NOT reset the accumulated backoff, which `resetResizeCount()` does. The two
+   * are not symmetrical: a window resize is evidence about the CAUSE of the deliveries, so the
+   * succession they belong to can be forgotten, while a change of scrollable element says only which
+   * element to watch and nothing about whether the loop went away.
    */
   observe() {
-    const parentElement = this.#deps.wtTable.wtRootElement.parentElement;
+    const wasInCooldown = this.#reconnectTimeoutId !== null;
```
budnix, line 203: `observe()` is the one public method that does not check `#isDestroyed`, so a call after `destroy()` attaches the observer again and quietly undoes the disconnect. Nothing throws, because the callback bails out on the flag, but the observer stays live and keeps holding the parent element for the rest of the page's life. The destroy tests cover the reconnect timer and the queued frames, not a direct `observe()`, so this path has no cover today.
```suggestion
    if (this.#isDestroyed) {
      return;
    }

    const wasInCooldown = this.#reconnectTimeoutId !== null;
```

handsontable/src/3rdparty/walkontable/src/overlay/resizeMonitor.ts
```diff
@@ -102,32 +168,212 @@ export class ResizeMonitor {
...
+      this.#resetSuccession();
+    });
+  }
+
+  /**
+   * Disconnects the observer for a cooldown and warns once. The counter is cleared here rather than on
+   * reconnect so that `resetResizeCount()` and a quiet frame during the cooldown stay meaningful.
+   */
+  #tripLoopGuard() {
+    if (!this.#hasWarned) {
```
budnix: `helpers/console` - the same module this file already imports `warn` from - has a `warnOnce(scope, key, ...args)` that keeps the already-warned state for you in a `WeakMap`. `formulas.ts` passes `this` as the scope, which gives the same once-per-instance behavior as `#hasWarned`, so the field and the `if` could both go. A spy on `console.warn` still sees exactly one call, so the specs would not need to change. Worth a quick look.

handsontable/src/3rdparty/walkontable/src/overlay/resizeMonitor.ts
```diff
@@ -62,32 +141,19 @@ export class ResizeMonitor {
...
-      if (this.#containerDomResizeCount === 300) {
-        warn('The ResizeObserver callback was fired too many times in direct succession.' +
-          '\nThis may be due to an infinite loop caused by setting a dynamic height/width (for example, ' +
-          'with the `dvh` units) to a Handsontable container\'s parent. ' +
-          '\nThe observer will be disconnected.');
+    this.#registerDelivery();
 
-        this.#resizeObserver.disconnect();
-      }
+    this.#settingFrameId = this.#deps.rootWindow.requestAnimationFrame(() => {
```
budnix: The timers and frames moved onto the injected `rootWindow` here, but the observer a few lines up is still built from the page-global `ResizeObserver`. `rootWindow` is `rootElement.ownerDocument.defaultView`, so for a container-scrolled grid inside an iframe those two come from different windows and only half the iframe fix lands. Moving the construction into the constructor is what it takes, since a class field initializer runs before `#deps` is assigned, and the unit harness would swap its `window.ResizeObserver` patch for a property on the fake `rootWindow`. Reads like a consistency gap rather than a correctness bug, but worth a look given the stated reason for the move.

handsontable/src/3rdparty/walkontable/src/overlay/resizeMonitor.ts
demtario: Fixed in bbbfdbcec8 — you were right, and this was the worst of the four. `observe()` now returns early on `#isDestroyed`.

The silence is what made it bad: every callback bails out on the flag, so a post-teardown `observe()` throws nothing and the observer just stays live holding the parent element. Covered by `should refuse to observe after it was destroyed`, which asserts the observer never attaches a second time.

tests/e2e/refresh-dimensions.spec.ts
demtario: Good catch, and it does not always land inside the cooldown — this is the same flake class the PR exists to remove, so I reworked it rather than adding a precondition (a precondition that can fail on a loaded runner is still a flake).

The measurement moved into the page. The fixture starts a per-frame sampling series the moment it sees the guard's warning and stops before 1500 ms, checking elapsed time *before* taking each sample, so every sample is strictly inside the 2000 ms cooldown however long an individual frame turns out to be. Neither the poll interval nor the frame duration is in the path any more; the spec just reads the series.

Your suggestion to record the time alongside the warning is what made it work — it just ended up driving the sampling instead of asserting a precondition after the fact.

One thing the rework surfaced, by failing: the first assertion came back `299`, not `>= 300`. The guard warns synchronously while it counts, so the sampling frame is registered *before* the tripping delivery's own deferred hook fire — the first sample is one fire short of the threshold. The comparison now starts at the third sample and the threshold assertion sits on the settled value, with the ordering written down in a comment.

Verified: 48/48 across the six legs, 144/144 with `--repeat-each=3`, and the reworked spec still fails when `resizeMonitor.ts` is reverted to `develop`.

handsontable/src/3rdparty/walkontable/src/overlay/resizeMonitor.ts
demtario: Adopted. `#hasWarned` and the `if` are both gone; it is `warnOnce(this, LOOP_GUARD_WARN_KEY, ...)` now, scoped on the instance exactly as `formulas.ts` does it, so a second grid on the page still gets its own warning.

The specs did not need changing, as you expected — both still assert a single `console.warn` call across two trips.

handsontable/src/3rdparty/walkontable/src/overlay/resizeMonitor.ts
demtario: Fixed, and I would rate this higher than a consistency gap — it is a correctness bug for this specific design.

The guard now counts *frames*: a delivery arms a watchdog on `rootWindow.requestAnimationFrame`, and the succession is broken when a frame passes with no delivery. If the observer belongs to a different window, its observations are delivered on that window's rendering loop while the watchdog runs on the grid's, so the two cadences can drift apart and the watchdog can read a busy frame as quiet (or the reverse). That is the one-delivery-per-frame invariant the whole fix rests on, and it does not hold across a realm boundary.

Construction moved into the constructor, as you noted it would have to. Rather than cast, I widened `rootWindow` to `Window & typeof globalThis` — the idiom `src/core/types.ts:267` already uses for the same reason. Four declarations, no cascade beyond the `CoreAbstract` getter.

The unit harness now puts the fake constructor on its fake `rootWindow` and no longer patches `window.ResizeObserver`, which is its own proof the construction really moved: jsdom's no-op stub would make every delivery in that file a no-op if the class still reached for the global.

budnix (approved):

PR #13358, opened 2026-09-03
DEV-2772: Fix the high-severity Dependabot alerts

qunabu:
### Context

The PR fixes the high-severity [Dependabot alerts](https://github.com/handsontable/handsontable/security/dependabot). The repository had **229 open alerts — 104 high severity, 0 critical**.

**81 of the 104 high alerts are closed here**, plus 77 medium/low ones as a side effect (158 of 229 total). The 23 that remain are listed at the bottom; none of them ship in a published package, and each needs a decision beyond a dependency bump.

Nearly every alert lands on a *transitive* dependency, so bumping a declaration is rarely the lever. The change works through two mechanisms:

| Where | Change |
|---|---|
| Root `pnpm.overrides` | brace-expansion 1/2/5, browserslist, fast-uri, http-proxy-middleware 3, immutable, js-yaml 3/4, nanoid 3, pacote, piscina 4, postcss (`^8.5.23` — see below), sharp, shell-quote, svgo 4, `external-editor>tmp`, vite 6 |
| `docs` | `@playwright/test` → `catalog:` (1.62.1) — drops the last Playwright vulnerable to GHSA-7mvr-c777-76hp. This branch first pinned `1.61.x` by hand; develop's #13248 then moved the whole monorepo onto the catalog, which does the same job and also lands the migration `pnpm-workspace.yaml` asks for, so the merge takes `catalog:` and drops the manual pin |
| `docs/angular-type-check/package-lock.json` | Deleted and gitignored |
| `examples/next` | vite → `^6.4.3` in all 13 example manifests that pin it, plus an `overrides` block in the two `react-wrapper` framework roots |

Each override is scoped as narrowly as its advisory — a bare name where a single major is installed, `name@<major>` where several are, `parent>child` where only one consumer is affected.

#### The deleted lockfile

This was the single biggest win. `docs/angular-type-check/` is a **pnpm workspace member** listed in `pnpm-workspace.yaml`; it is installed by the root `pnpm install` and its dependencies resolve from `pnpm-lock.yaml`. Nothing ever runs `npm ci` there — `docs-angular-typecheck.yml` and the matching composite action both just run `npm run typecheck --prefix docs/angular-type-check` after a root `pnpm install`.

So the committed `package-lock.json` was never read. It had drifted to Angular 21.2.11 against a declared `^19.2.20`, which is not even a satisfiable resolution, and Dependabot scanned it anyway — it was the only source of **11 high alerts**.

#### Why vite 6 and not 7 or 8

`vite-plugin-static-copy` (peer `^5.0.0 || ^6.0.0`) and `@vitejs/plugin-vue` (peer `^5.0.0 || ^6.0.0`) cap the examples at vite 6. `^6.4.3` is the lowest release that clears both GHSA-c27g-q93r-2cwf and GHSA-fx2h-pf6j-xcff, and it is the highest major the existing plugins support.

#### Two collateral fixes

`react-scripts@5.0.1` freezes a large transitive tree, so minimatch 9, postcss, serialize-javascript and underscore sit below their patched versions with no reachable range. The npm `overrides` block in the two `react-wrapper` framework roots is the only lever there.

Both `react-wrapper` demos also move from `typescript@4.2.4` to `^5.0.2`, matching their `accessibility` sibling. `react-scripts` type-checks the grid's own sources, and 4.2.4 cannot parse the inline `type` import modifier they now use — the build broke as soon as the lockfile floated `handsontable` from 18.0.0 to 18.1.0 through its `latest` specifier. This is latent on `develop` today: any `npm install` in that example hits it, including the one `install-subpackages.mjs` runs on every examples build.

#### The postcss BOM — fixed at the source, so there is no cap

The first push of this branch turned **all six `UMD.min` E2E legs** red while every plain `UMD` leg stayed green. The cause was the postcss bump, and the mechanism is worth recording because nothing in the build complains about it.

sass writes `@charset "UTF-8"` into `handsontable.css`, because the themes contain a non-ASCII glyph. From **postcss 8.5.24** onward, the `css-loader` → `CssExtractRspackPlugin` → `LightningCssMinimizerRspackPlugin` chain turns that at-rule into a **UTF-8 BOM** instead — and it lands **after** the preserved `/*!` license banner, at byte 1662, not at offset 0. That position is what makes it dangerous: a *leading* BOM is stripped by the CSS decoder, but one at a non-zero offset is stripped by nothing, and U+FEFF is an ident code point in CSS rather than whitespace. It fuses with the selector that follows, so the first rule becomes `\uFEFF.ht-root-wrapper` and silently stops matching. The grid's root wrapper loses `display:flex`, `position:relative` and `height:100%`; cells stop receiving pointer events (`#grid … intercepts pointer events` in the Playwright call log) and popovers mis-anchor.

**There are two consumers, and both break** — the earlier revision of this description named only the first:

1. `.config/plugin/rspack/css-to-js-export-plugin.js` reads the file with `fs.readFileSync(cssPath, 'utf8')`, which does **not** strip a BOM in Node, and `JSON.stringify`s it into `src/styles/handsontableStyles.{js,ts}`. Every bundle that inlines the runtime theme string carries it.
2. `handsontable/styles/handsontable.min.css` itself ships corrupted, to anyone who loads the minified stylesheet directly. Confirmed in Chrome against a real 8.5.28 build: with the BOM, `.ht-root-wrapper` computes `display:block; position:static; flex-direction:row`; without it, `display:flex; position:relative; flex-direction:column`.

Only the E2E legs that load the inlined string went red, which is why a CSS regression first presented as three unrelated `dblclick` timeouts plus one popover-position assertion on one bundle axis.

**The fix is ours, not a version cap.** `.config/plugin/rspack/strip-css-bom-plugin.js` removes the BOM from the emitted asset in `.config/styles-production.js`, registered *before* `addCssToJsExport` — both are `afterEmit` taps, so registration order decides which one sees it — and the export plugin repeats the strip as a belt to those braces. Verified by building at **postcss 8.5.28** in a throwaway worktree: `handsontable.min.css` comes out byte-identical to the 8.5.23 output (modulo build date), with no `EF BB BF` at any offset, and both inlined exports are clean.

So the override is `postcss: '^8.5.23'` — `>=8.5.18` is what GHSA-r28c-9q8g-f849 requires and `>=8.5.23` what GHSA-fxqj-rqcc-2cmp raised it to. **A `<8.5.24` cap is deliberately not used**: it admits exactly one published version, so the next postcss advisory would leave no satisfiable range.

Widening the range moves nothing today — pnpm keeps a lock entry that still satisfies the range, so `postcss` stays resolved at 8.5.23 across the tree and the only lockfile delta is the recorded override string. `handsontable/test/__tests__/cssBomStrip.unit.js` pins the strip, the plugin ordering, and the absence of a cap; it fails if the strip is removed.

The two `examples/next/*/react-wrapper` npm overrides deliberately stay at `^8.5.18`: those trees build through Create React App's postcss chain, never through lightningcss, and all four react examples build clean on 8.5.26.

#### No toolchain float

`core-js`, `hyperformula`, `typescript`, `esbuild`, `@rspack/core`, `lightningcss`, `sass` and `@swc/core` are all unchanged against develop's lockfile — this is not a DEV-2667-style float. `browserslist` moves only because GHSA-73wf-gq98-2v4g requires it, and `BROWSERS_LIST` is pinned integers, so the accompanying caniuse-lite data cannot shift the compile target. The 32 other differences are the npm-CLI family (`cacache`, `make-fetch-happen`, `ssri`, `node-gyp`, `nopt`, `which`, `npm-*`, …) that the `pacote` 20 → 21 override pulls in; none of it reaches a bundle.

#### Merging develop (two merges)

**First merge** — `docs/package.json` and `pnpm-lock.yaml` conflicted, and develop won both. The docs one is covered in the table above. `pnpm-lock.yaml` was settled the way `publish.yml`'s `stable-merge` does it: take develop's file, then `pnpm install --lockfile-only` to re-apply this branch's `pnpm.overrides` on top, so develop's resolutions stay the base and nothing of develop's floats. A side benefit — `postcss` is now a single 8.5.23 (the stale 8.5.26 peer entry the earlier re-resolve had left behind is gone) and `playwright` resolves to one version across the tree instead of three.

**Second merge** — only `AGENTS.md` conflicted, a both-sides-added-bullets clash at the same insertion point in the monorepo-gotchas list. Nothing was genuinely in contention, so both sides are kept, with develop's new DEV-2714 bullet leading (it is a sibling of the release-lockfile bullet it sits next to).

The two bullets did overlap in *advice*, so the merge reconciles them — see the open question below.

#### `Visual / Compare` — green

An earlier revision of this branch (`252f0e7c`) reported 6 changed of 1646, all filters/menu-state captures, and this section argued at length that they were antialiasing noise rather than a rendering change. On the current head the gate reports **Changed items: 0 / New: 0 / Deleted: 0 / Passed: 1646**, with `visual-approved` absent — so the diffs are gone, no baseline was rewritten, and no label was applied. The analysis is left out of the description now that there is nothing to explain away.

The supporting measurement is still worth keeping, because it says something about the build: built from a clean worktree and diffed against develop with timestamps normalised, all 18 stylesheets are byte-identical and exactly one file differs — `dist/languages/all.min.js`, where the only deltas are rspack module IDs. The bundles differ solely in `Handsontable.buildDate`. `handsontable/tmp/` likewise. This change does not alter rendering.

One observation from that earlier run is still open and is filed as a follow-up: develop's own baseline run `33747512147` reported 4 changed items against the previous `base/develop`, which suggests the `thresholdPixel: 150` tolerance is under-set for menu-state captures. `visual-tests/AGENTS.md` says not to change it without re-measuring.

#### Open question: this branch floated the example lockfiles, and DEV-2714 now argues against that

`DEV-2714` (#13345) landed while this branch was in review, and its finding is that the example-lockfile float came from **deleting** those lockfiles, not from the `latest` pins — npm honours a locked edge for a dist-tag spec, so with the lockfile present `npm install` re-resolves nothing.

This branch regenerated seven of the nine example lockfiles from scratch, because npm ignores a *new* `overrides` block while a valid lockfile is on disk. Measured against develop, that floated **61–211 packages per lockfile**, including `handsontable` 18.0.0 → 18.1.0 — which is precisely what broke the two react demos' pinned `typescript@4.2.4` on the grid's own inline `type` imports and forced the `^5.0.2` bump in this PR.

Worth noting what is and is not affected by DEV-2714's rules: this branch's `overrides` sit in the **framework-level** `package.json`, not the leaf manifests, so they do not collide with `alignTypescriptWithAngular` (which only writes leaves declaring `@angular-devkit/build-angular`). And the `vite` bumps are direct leaf dependencies, so they need no re-resolve at all — a plain `npm install` with the lockfile left in place picks them up.

So there is a cheaper version of this PR, and it is the reviewer's call:

- **As it stands** — seven lockfiles refreshed, `handsontable` floated to 18.1.0, the two `typescript` bumps required, and the four `react-wrapper` transitive overrides (minimatch 9, postcss, serialize-javascript, underscore) in effect. Closes ~8 more high alerts. Tested green.
- **Minimal alternative** — keep develop's lockfiles, land only the 13 `vite` manifest bumps via a plain `npm install`, drop the `react-wrapper` `overrides` and both `typescript` bumps. Much smaller diff, no float, but leaves those ~8 alerts open in the two CRA example trees — which stay flagged regardless, since `svgo` there is unfixable until they migrate off Create React App.

I have left it as it stands because it is tested and closes more alerts, but the minimal version is more in keeping with DEV-2714's intent, and these trees ship in no bundle (which is why their audit is report-only by design). Say which you prefer and I will reshape it.

Nothing under `handsontable/src/**` or `wrappers/**` is touched, so `check-changelog.js` passes automatically. The build-config change (the BOM strip) does ship with a unit test; the rest is a dependency and lockfile change, verified by re-running the existing suites and by building every artifact the changed dependencies feed.

- Unit tests added/modified (`*.unit.js`): `handsontable/test/__tests__/cssBomStrip.unit.js` — drives `strip-css-bom-plugin` against a real file with a BOM after the banner, asserts the plugin is registered before `addCssToJsExport`, and asserts no `<8.5.24` cap returns. Verified failing with the strip removed. The full existing suite was re-run alongside it
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`): none — no user-facing behavior change
- For a bug fix — the spec that fails without this fix: `cssBomStrip.unit.js` for the BOM strip. For the dependency work the verification is the alert count and the green builds below
- Demo page / recorded trace (for UI changes): not applicable

### Commands run

```bash
# Core
npm --prefix handsontable run build                  # exit 0
npm --prefix handsontable run test:unit              # Test Suites: 345 passed, Tests: 4398 passed
npm run test:tooling                                 # exit 0
npm run eslint                                       # ESLint: No issues found

# Lockfile integrity
pnpm install --frozen-lockfile                       # exit 0

# Wrappers
npm --prefix wrappers/angular-wrapper run build      # exit 0
npm --prefix wrappers/react-wrapper run build        # exit 0
npm --prefix wrappers/vue3 run build                 # exit 0

# Test-workspace lint
npm --prefix visual-tests run lint                   # exit 0
npm --prefix tests run lint                          # exit 0
npm --prefix performance-tests run lint              # exit 0

# Docs site — NOT verified locally; left to CI
# `npm --prefix docs run build` gets through docs:api, validate-highlights and
# the vite client bundle, then prerenders thousands of API routes at several
# seconds each. Left running rather than blocked on. The docs-relevant risk here
# is sharp 0.34.5 -> 0.35.4 (astro's image service), smoke-tested directly:
node -e "const s=require('sharp');console.log(s.versions.sharp,s.versions.vips)"
# -> 0.35.4 8.18.6, and a PNG/WebP encode round-trip succeeds

# The postcss regression — bundle invariants after the fix
grep -c $'\xef\xbb\xbf' handsontable/styles/handsontable.min.css   # 0 (122906 bytes, == develop)
grep -o '\\uFEFF' handsontable/dist/handsontable.full.min.js | wc -l  # 1 (HyperFormula only)

# The two specs that failed, on all three -min themes
cd tests && npx playwright test e2e/dropdown-height.spec.ts e2e/license-branding.spec.ts \
  --project=e2e-classic-min   # PASS (29) FAIL (0)
  --project=e2e-horizon-min   # PASS (29) FAIL (0)
  --project=e2e-main-min      # PASS (29) FAIL (0)

# And the whole min leg
cd tests && npx playwright test --project=e2e-classic-min   # PASS (817) FAIL (0)

# All 18 affected examples, production builds — every one OK
for e in docs/js/{basic-example,demo,data-provider,accessibility} \
         docs/react-wrapper/{accessibility,basic-example,demo,data-provider} \
         docs/ts/demo docs/vue3/{basic-example,data-provider,demo} \
         visual-tests/js/{basic-example,demo} \
         visual-tests/vue3/{basic-example,demo} \
         visual-tests/react-wrapper/{basic-example,demo}; do
  (cd examples/next/$e && npm run build)
done
```

Alert coverage was measured by resolving every open alert's vulnerable version range against the versions each lockfile now resolves, rather than by eyeballing the diff.

The docs site build was not run to completion locally (it prerenders thousands of API routes), but `Docs / preview / deploy` has since run the full `astro build` through to `Complete!` and deployed green, so that gap is closed on CI.

### Types of changes

### Related issue(s):
1. DEV-2772
2.
3.

### Affected project(s):
- [ ] `handsontable`
- [ ] `@handsontable/vue3`

None of the four published packages changes its source. The edits are confined to the root workspace manifest and lockfile, `handsontable/.config/` (the BOM strip) and its unit test, `docs/`, `examples/next/` and `AGENTS.md`. The BOM strip is a no-op against the currently resolved postcss 8.5.23 — the emitted stylesheet is byte-identical with and without it — so no published artifact changes. The root overrides do feed every package's build, which is why all three wrapper builds and the core build were re-run above.

### The 23 high alerts left open

Each needs a decision beyond a dependency bump, and none ships in a published package:

- **extract-zip (9 alerts)** — reached through `@puppeteer/browsers`. GHSA-jmr9-qjv8-65gv has **no patched release at all**; 2.0.1 is the newest version published. Dev and test only.
- **image-size (2)** — an optional dependency of `less`. Both advisories have **no patched release**; the newest published version is still in range.
- **xlsx (4)** — 0.18.5 is the last SheetJS build ever published to npm, and both advisories are unpatched there. It is kept only so the TypeScript docs examples type-check; the browser examples load it from a CDN `<script>`. Fixing it means either re-sourcing from the SheetJS CDN — a non-registry dependency, which works against the air-gapped install support in `AGENTS.md` — or moving the one remaining example to ExcelJS, which most docs examples already use.
- **`@angular/*` 19.2.x (6)** — Angular has not backported these to the 19 line. They clear only when the wrapper's support matrix moves to 20/21, which is a support decision, not a patch.
- **svgo 1.3.2 (2)** — capped at `^1.2.2` by `@svgr/plugin-svgo@5` under `react-scripts@5.0.1`. Forcing svgo 2 breaks the SVG loader at build time, and the advisory is a `removeScripts` weakness that only matters for untrusted SVG input, which these examples do not have. The real fix is migrating the four Create React App examples to Vite.

Suggested follow-ups, in order of value:

1. Migrate the CRA `react-wrapper` examples to Vite — clears `svgo` and drops a large unmaintained transitive tree.
2. Take the puppeteer major (25.x) to clear the nine `extract-zip` alerts.
3. Decide how to source `xlsx` for the docs examples.
4. Decide whether the Angular wrapper support matrix moves off 19.x.
5. Add a PR job that builds `examples/next/docs/**`. Raised in review: today `examples:install` / `examples:build` run only in `publish.yml`'s `first-rc-build`, so nothing on a pull request builds those apps — a path filter alone would have no job to trigger. This PR's vite 4 → 6 jump is covered by the 18 hand-run production builds; the next one would land unwatched.
6. Re-measure the visual `thresholdPixel` for menu-state captures. (The BOM half of this follow-up is done in this PR — the strip removes the corruption at emit rather than asserting against it later.)

Nothing was dismissed in the Dependabot UI. All 25 manifest paths carrying an open alert still exist on `develop`, so no alert pointed at removed code — `examples/prs/` was removed earlier, but Dependabot had already auto-closed those.

### Checklist:

`AGENTS.md` gained three bullets rather than user-facing docs, since nothing here is user-visible: the BOM mechanism, both of its consumers and the reason a version cap is the wrong tool for it; the two override mechanisms and the npm trap where a new `overrides` block is silently ignored while a valid `package-lock.json` or any `node_modules/` is on disk (npm prints `up to date` and leaves the vulnerable version pinned), pointing at `clean-subpackages.mjs next --reset-lockfiles` as the sanctioned way into that state; the four permanently unfixable alert families, so they are not chased again; and why `docs/angular-type-check/` must not carry a lockfile.

ClickUp task: https://app.clickup.com/t/123kvxe9h5m

---

> [!NOTE]
> **Medium Risk**
> Root overrides and postcss caps affect every package build; a wrong postcss upper bound would silently break minified CSS and full-bundle E2E. Example lockfile floats change resolved versions in non-shipped demos only.
> 
> **Overview**
> Closes most open high-severity Dependabot alerts by tightening **root `pnpm.overrides`**, bumping **Vite to `^6.4.3`** across affected `examples/next` JS/React demos, and adding **npm `overrides`** in the two `react-wrapper` framework roots (CRA transitive fixes). **`postcss` stays capped below 8.5.24** so minified CSS does not pick up a mid-file BOM that breaks inlined theme strings and `.min` E2E legs.
> 
> **`docs/angular-type-check/`** drops its committed **`package-lock.json`** and gitignores it, since the folder is a **pnpm workspace member** resolved from `pnpm-lock.yaml`—the stale npm lock was scanned by Dependabot but never used in CI.
> 
> **`AGENTS.md`** gains maintainer guidance on BOM/postcss invariants, the split between pnpm vs npm override trees (including `performance-tests/`), unfixable alert families, and the angular-type-check lockfile rule.
> 
> Collateral in the example lock refresh: **`handsontable` `latest` → 18.1.0** and **`typescript` `^5.0.2`** on CRA react demos so type-checking survives the grid’s inline `type` imports.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ae377b7050d7432dcd481d94e5e1cb9f377205b9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2772 Fix the high-severity Dependabot alerts](https://app.clickup.com/t/9015210959/DEV-2772)
Task linked: [DEV-2667 [18.1.0-rc1+] Release-cut regenerated pnpm-lock.yaml, floating deps — production-bundle E2E legs deterministically red on release/18.1.0](https://app.clickup.com/t/9015210959/DEV-2667)

qunabu:
Task linked: [DEV-2714 Guard example lockfiles against re-float at release time](https://app.clickup.com/t/9015210959/DEV-2714)

package.json
```diff
@@ -104,7 +104,25 @@
       "cacache>tar": "^7.5.19",
```
budnix: Same shape as the brace-expansion entries below: GHSA-r292-9mhp-454m is a high advisory patched in tar 7.5.21, and `^7.5.19` still lets 7.5.19 and 7.5.20 through. Its two siblings a few lines up, `pacote>tar` and `node-gyp>tar`, carry the same floor. The lock resolves 7.5.22 today so nothing is broken, but since this PR is raising floors anyway these three seem worth pulling up with the rest.

```suggestion
      "cacache>tar": "^7.5.21",
```

package.json
```diff
@@ -104,7 +104,25 @@
...
       "typed-rest-client": "^3.1.0",
       "undici": "^6.28.0",
-      "@astrojs/mdx": "^6.0.1"
+      "@astrojs/mdx": "^6.0.1",
+      "brace-expansion@1": "^1.1.16",
+      "brace-expansion@2": "^2.1.2",
+      "brace-expansion@5": "^5.0.7",
```
budnix: These three floors all sit one release below the patched version — GHSA-rgw5-rvv9-x895 patches at 1.1.18, 2.1.4 and 5.0.9, so `^1.1.16`, `^2.1.2` and `^5.0.7` still let a vulnerable version through. The lock happens to resolve 1.1.18, 2.1.4 and 5.0.9 today because pnpm picks the highest match, so nothing is broken right now, but a re-resolve narrowed by a peer range could quietly drop back. Raising the floors makes the guard hold on its own.

```suggestion
      "brace-expansion@1": "^1.1.18",
      "brace-expansion@2": "^2.1.4",
      "brace-expansion@5": "^5.0.9",
```

package.json
```diff
@@ -104,7 +104,25 @@
...
+      "browserslist": "^4.28.7",
+      "fast-uri": "^3.1.6",
+      "http-proxy-middleware@3": "^3.0.7",
+      "immutable": "^5.1.8",
+      "js-yaml@3": "^3.15.1",
+      "js-yaml@4": "^4.3.1",
+      "nanoid@3": "^3.3.18",
+      "pacote": "^21.5.1",
+      "piscina@4": "^4.9.3",
+      "postcss": ">=8.5.18 <8.5.24",
```
budnix: There's a newer postcss advisory, GHSA-fxqj-rqcc-2cmp, that patches at 8.5.23 — it's only medium so it sits outside this PR's high-severity scope, but it means 8.5.18 through 8.5.22 are still affected. Since the BOM cap already blocks everything from 8.5.24 up, 8.5.23 is the only version this range can actually pick, so tightening the floor costs nothing and makes that explicit. If you do change it, the AGENTS.md sentence about the lower bound would need a matching tweak — it reads correctly as written today.

```suggestion
      "postcss": ">=8.5.23 <8.5.24",
```

examples/next/docs/ts/demo/package.json
```diff
@@ -11,7 +11,7 @@
   },
   "dependencies": {
     "handsontable": "latest",
-    "vite": "4.4.3"
+    "vite": "^6.4.3"
```
budnix, line 14: Regenerating this lock floated `handsontable` from 18.0.0 to 18.1.0, which is the same thing that pushed the two react-wrapper demos off TypeScript 4.2 — but this package still pins `typescript: 4.2.3` a few lines down. I tried it: `tsc --noEmit --skipLibCheck` (this package's own `check-types` script) against handsontable 18.1.0 gives a dozen `TS1005: ',' expected` errors from inside `node_modules/handsontable/3rdparty/walkontable/`, because the shipped typings use inline `type` import modifiers and `skipLibCheck` doesn't skip syntax errors. TypeScript 5.9.3 passes the same check clean. Was this one left out on purpose, or did it just get missed alongside the react demos?

examples/next/docs/js/demo/package.json
```diff
@@ -25,7 +25,7 @@
     "jasmine": "3.7.0",
     "jasmine-console-reporter": "3.1.0",
     "puppeteer": "^24.2.1",
-    "vite": "4.4.3"
+    "vite": "^6.4.3"
```
budnix, line 28: Vite jumps two majors here, and I couldn't find a CI path filter that covers `examples/next/docs/**` — `checks.yml` only lists `./examples/next/visual-tests/**`, so the visual-tests copies get built but these don't. The PR description says all 18 examples were built by hand, which covers this change, but the next vite major would land with nothing watching. Would it be worth adding a path filter for the docs examples so this stays covered?

AGENTS.md
```diff
@@ -226,6 +226,11 @@ Follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) format. Changel
...
...k in a throwaway directory holding only the framework `package.json` and its examples' `package.json` files, then copy the result back. **That full re-resolve is the expensive option, so weigh it against the bullet above**: it floats the whole tree (61-211 packages per lock in DEV-2772, `handsontable` 18.0.0 -> 18.1.0 among them, which then broke the react demos' pinned `typescript@4.2.4` on the grid's own inline `type` imports), and those trees ship in no bundle, which is exactly why their audit is report-only. A **direct** dependency bump in a leaf manifest — the usual security fix there, e.g. `vite` — needs none of this: a plain `npm install` with the lockfile left in place picks it up and re-resolves nothing else. Reach for the throwaway re-resolve only when the fix genuinely has to travel through `overrides`, and say in the PR why the float was worth it. After any lockfile work, confirm the compile floor did not move: `core-js`, `hyperformula`, `typescript` and `esbuild` must be byte-identical, and `browserslist`/`caniuse-lite` may only move when an advisory demands it (`BROWSERS_LIST` is pinned integers, so caniuse data cannot shift the target — see the ES-floor entry above).
```
budnix: There's a third tracked npm lockfile this bullet doesn't mention: `performance-tests/package-lock.json`. It's npm-managed with its own committed lock, so it inherits nothing from `pnpm.overrides`, same as the `examples/next` trees. It resolves patched versions today, so nothing is wrong now — but someone following this bullet would fix two trees and miss the third.

AGENTS.md
```diff
@@ -226,6 +226,11 @@ Follows [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) format. Changel
...
+- **Four high-severity Dependabot alerts have no fix and must not be chased.** `extract-zip@2.0.1` (via `@puppeteer/browsers`) and `image-size@0.5.5` (an optional dependency of `less`) have no patched release at all. `xlsx@0.18.5` is the last SheetJS build ever published to npm — patched versions exist only on the SheetJS CDN, and it is kept here purely so the TypeScript docs examples type-check, while the browser examples load it from a CDN `<script>`. `svgo@1.3.2` in the two `react-wrapper` example trees is capped by `@svgr/plugin-svgo@5` under `react-scripts@5.0.1`; forcing svgo 2 breaks the SVG loader, so it clears only when those Create React App examples migrate to Vite. The `@angular/*` 19.2.x alerts are the same shape: Angular has not backported them, so they clear only when the wrapper's support matrix moves.
```
budnix: The extract-zip part reads as permanent, but it looks like there is a way out. `puppeteer@25.9.0` ships `@puppeteer/browsers@3.2.1`, whose only dependencies are `yargs` and `modern-tar` — extract-zip is gone. The `^24.2.1` pin in `handsontable/package.json` keeps us on `@puppeteer/browsers@2.13.2`, which still pulls `extract-zip ^2.0.1`. Taking a puppeteer major is a real decision, so deferring is fair, but wording it as "deferred until we move to puppeteer 25" would stop the next person from crossing it off the list forever.

package.json
qunabu: Applied, all three — thanks. Confirmed GHSA-r292-9mhp-454m is high and patched in 7.5.21, so `^7.5.19` admitted 7.5.19 and 7.5.20. `cacache>tar`, `pacote>tar` and `node-gyp>tar` are now `^7.5.21`. The lock does not move (it resolved 7.5.22 already), so the diff is confined to the recorded `overrides` block — which is exactly your point: the guard now holds on its own rather than relying on pnpm picking the highest match.

package.json
qunabu: Good catch, and it is worse than a floor being one release low — GHSA-rgw5-rvv9-x895 is a *second* high advisory on brace-expansion, and it specifically bypasses the CVE-2026-14257 mitigation that GHSA-3jxr-9vmj-r5cp addressed. So the floors I derived from that first advisory were stale the moment I wrote them. Raised to `^1.1.18` / `^2.1.4` / `^5.0.9`. No resolved version changes.

package.json
qunabu: Applied: `>=8.5.23 <8.5.24`. GHSA-fxqj-rqcc-2cmp patches at 8.5.23, so my `>=8.5.18` was admitting five affected versions. As you note, with the BOM cap already at `<8.5.24` the range now resolves the single version that satisfies everything, so it is effectively a pin until the BOM is fixed upstream — and the AGENTS.md sentence now says that, listing all three bounds (8.5.18 from GHSA-r28c-9q8g-f849, raised to 8.5.23 by GHSA-fxqj-rqcc-2cmp, capped below 8.5.24 by the BOM).

examples/next/docs/ts/demo/package.json
qunabu: Missed, not deliberate — you are right. I had looked at this file, saw `check-types` was not wired into any CI job, and moved on; I should have fixed it anyway since it is the same defect as the react demos. Reproduced your result exactly: on 4.2.3 it fails inside `node_modules/handsontable/3rdparty/walkontable/` with TS1005, because `skipLibCheck` does not skip *syntax* errors. Now `^5.0.2` like its siblings, and `npm run check-types` exits 0 with no errors (one caveat for anyone repeating it: a stale `dist/` makes `tsc` report an unrelated TS6059 about `rootDir`, so clean it first).

examples/next/docs/js/demo/package.json
qunabu: Verified and it is worse than a missing filter: there is no PR job to filter *to*. `examples:install` / `examples:build` appear only in `publish.yml` (`first-rc-build`), so nothing on a pull request builds `examples/next/docs/**` at all. The `visual-tests/**` entries you found are there because the visual suite builds those particular apps; the docs examples have no equivalent. So adding a path filter alone would trigger nothing — closing this properly means adding a job that builds the docs examples, which I would rather not fold into a security PR. Worth its own ticket, and I have flagged it in the PR description. For this change the coverage is the 18 hand-run production builds; the first automated exercise is the next RC cut.

AGENTS.md
qunabu: Added — the bullet now names `performance-tests/` as a third npm-managed tree with its own committed lockfile, equally outside `pnpm.overrides`, and says explicitly that a sweep fixing only the `examples/next` trees has missed one. Confirmed it resolves patched versions today, so nothing is wrong now; the omission was in the guidance, which is the part that would have bitten the next person.

AGENTS.md
qunabu: You are right and my wording was overconfident — I put extract-zip in a bullet headed "no fix and must not be chased", which invites exactly the permanent cross-off you describe. Confirmed `puppeteer@25.9.0` ships `@puppeteer/browsers@3.2.1` with only `yargs` and `modern-tar`, extract-zip gone, and that the `^24.2.1` pin in `handsontable/package.json` is what holds us on `@puppeteer/browsers@2.13.2`. Reworded: `image-size` stays as genuinely unpatched, extract-zip is now called out as escapable and **deferred until we move to puppeteer 25**, not closed, with the pin named as the thing holding it. The PR description follow-up list says the same.

budnix (dismissed):

demtario (dismissed):

danielzytohoc (commented):
Solid work, and the write-up is unusually honest about its own tradeoffs — the postcss/BOM bisection in particular is the kind of thing that normally gets lost. I checked the load-bearing claims rather than taking the description at face value; they hold up. Five things I'd like changed or re-run before this goes in.

## What I verified

- **No toolchain float.** I did a resolved-version set diff across both lockfiles (2,200 → 2,196 packages, 89 changed names). `core-js`, `hyperformula`, `typescript`, `esbuild`, `@rspack/core`, `lightningcss`, `sass` and `@swc/core` are all absent from the changed set, and every one of the 89 falls into the sharp / pacote-21-npm-CLI / browserslist / advisory buckets. The claim is accurate.
- **The overrides actually took effect.** `postcss` consolidated 8.5.12 + 8.5.15 + 8.5.16 into a single 8.5.23; the two CRA trees resolve minimatch 9.0.9, postcss 8.5.26, serialize-javascript 7.1.1, underscore 1.13.8. I also confirmed empirically that npm does not record an `overrides` block in the lockfile root, so its absence from `packages[""]` is normal rather than a sign of a lock that drifted from its manifest.
- **vite is genuinely clear.** 6.4.3, 7.3.5 and 8.1.3 all clear GHSA-fx2h-pf6j-xcff and GHSA-v6wh-96g9-6wx3, and every example lock resolves 6.4.3.
- **The lockfile deletion is safe.** `docs/angular-type-check` is a pnpm workspace member (`pnpm-workspace.yaml:12`) and the only thing ever run there is `npm run typecheck --prefix`. `Docs / Angular types` is green.
- **The Node floor is fine.** The pacote-21 family declares `^20.17.0 || >=22.9.0`; every `.nvmrc` in the repo is `22`.
- **The docs gap you flagged is already closed.** `Docs / preview / deploy` ran the full `astro build` through to `Complete!` and deployed, green in 8m24s, on sharp 0.35.4 and postcss 8.5.23. You can drop that caveat from the description.

## Findings

### 1. The postcss cap is a one-version pin, and the root cause is one line in this repo

`>=8.5.23 <8.5.24` admits exactly 8.5.23 — 8.5.24 through 8.5.28 are all published. The next postcss advisory makes this range unsatisfiable-with-a-fix, and we'd be choosing between a known CVE and a corrupted stylesheet with no third option.

But the defect is ours, not upstream's. `handsontable/.config/plugin/rspack/css-to-js-export-plugin.js:29` reads with `'utf8'`, which does not strip a BOM in Node:

```js
const css = fs.readFileSync(cssPath, 'utf8').replace(/^\uFEFF/, '');
```

That makes the cap unnecessary and is robust to whatever the toolchain does next. Follow-up 6 proposes *asserting* the absence of a BOM; stripping it is strictly better and cheaper than both the assertion and the pin. I'd do it here and widen the override to `^8.5.23`.

### 2. There is no visual verdict on the current head

The "6 changed of 1646, antialiasing noise" analysis is from run `33748810662` (`252f0e7c`), where cross-browser was green in 8m30s. On the current head (`6c9ae1fe`) it never got that far: `Visual / cross-browser` hit the 6-hour job ceiling at *Build visual test examples* and was cancelled — the orphan processes at kill were `npm run build` → `ng build` → `esbuild`. Every comparison step was skipped, `Visual / Compare` failed at *Download cross-browser screenshots*, and `CI Gate` failed off that.

I don't think it's your change: `Visual / multi-framework` ran the identical step in 10 minutes on the same commit, and I found no comparable hang across the last 40 `Tests` runs, so it looks like a stuck runner. A re-run is in flight as I write this. Worth noting so nobody reads the current red as the antialiasing story — and the `visual-approved` decision can't be made until the gate actually compares something.

### 3. The new `AGENTS.md` bullet documents a procedure we already have

The Dependabot bullet says to generate the lock in a throwaway directory holding only the framework and leaf manifests, then copy it back. `clean-subpackages.mjs next --reset-lockfiles` already removes both the framework lockfiles and every `node_modules` under those trees (`examples/scripts/clean-subpackages.mjs:52-60`) — which is exactly the no-lock, no-`node_modules` state in which npm honours a new `overrides` block.

Please point the bullet at the existing flag. As written we end up with two documented ways to do one operation, and DEV‑2714's sanctioned path reads as though it were insufficient for this case when it isn't.

### 4. The description is stale on its own headline number

The postcss section says the override is `>=8.5.18 <8.5.24`; the code and `AGENTS.md` both say `>=8.5.23 <8.5.24`. For a trap this subtle the PR body is the record people will come back to, so worth correcting.

### 5. Dead override entry

`minimatch@9: ^9.0.7` in `examples/next/visual-tests/react-wrapper/package.json` is a no-op — that tree resolves only minimatch 3.1.5 and 5.1.9. It looks like a copy of the docs-tree block. Drop it, or add a word saying it's a deliberate forward guard.

## On the open question

**Keep it as it stands.** DEV‑2714's rule exists to stop the release job from floating those trees silently; a deliberate, reviewed, tested refresh landed on `develop` is precisely the path that bullet describes as the sanctioned one. The minimal alternative leaves roughly eight highs open in trees that can't be fully cleared anyway while svgo is capped by CRA, and buys a smaller diff in exchange for coverage. The one change that would make it properly consistent with DEV‑2714 is finding 3 — regenerate through `--reset-lockfiles` so the documented path is also the real one.

## One process note

The body carries `DEV-2714` five times and `DEV-2667` once, and `252f0e7c`'s commit message carries `DEV-2714` three more. Squash-merge concatenates commit messages, so ClickUp's autolink will fire on both tasks and can auto-transition or auto-close them. DEV‑2714 is already merged so that one is cheap; I haven't checked DEV‑2667's current state. Worth referring to them as `#13345` and by title in the squash message.

Nit: the new bullets use ASCII `->` where the file's convention is `→`.

qunabu:
Thanks — all five actioned in `ae377b705`. One correction, on finding 1.

**Stripping at the read site isn't enough.** The BOM lands at byte 1662, *after* the preserved `/*!` banner, not at offset 0 — and that position is the bug. A leading BOM is stripped by the CSS decoder; one further in is stripped by nothing, and U+FEFF is a CSS ident code point rather than whitespace, so it fuses with the selector that follows. Built at 8.5.28 in a throwaway worktree and measured in Chrome:

| `handsontable.min.css` | display | position | flex-direction |
|---|---|---|---|
| with BOM | `block` | `static` | `row` |
| control | `flex` | `relative` | `column` |

So the shipped stylesheet breaks too, not just the inlined copy — my original write-up was wrong to pin it on the inlined string alone. `css-to-js-export-plugin` stripping on read would have left every direct consumer of the minified stylesheet with a dead `.ht-root-wrapper`.

Fixed at the emitted asset instead: `strip-css-bom-plugin.js`, registered before `addCssToJsExport` (both are `afterEmit` taps, so order decides which sees it). Your read-site strip is kept as the second line. At 8.5.28 the stylesheet now comes out byte-identical to the 8.5.23 output with both exports clean — which is what actually licenses the widen to `^8.5.23`.

The widen moves nothing today: pnpm keeps a lock entry that still satisfies the range, so postcss stays resolved at 8.5.23 everywhere and the only lockfile delta is the recorded override string — your no-float diff still holds. `cssBomStrip.unit.js` pins the strip, the ordering and the absence of a cap; verified red with the strip removed.

The rest:

2. Agreed, and the re-run went green on the same head — **Changed 0 / Passed 1646**, `visual-approved` absent. Nothing was approved and no baseline rewritten. The antialiasing section is out of the description; the under-set `thresholdPixel` stays as follow-up 6.
3. Done — the bullet now points at `clean-subpackages.mjs next --reset-lockfiles`. Confirmed it clears both levels of `node_modules` too, so it is genuinely the same state.
4. Fixed, along with the stale docs-build caveat and visual section.
5. Dropped. That tree resolves only minimatch 3.1.5 and 5.1.9, and npm doesn't record overrides in the lock, so nothing needed regenerating.
- Nit: fixed.

**Not done:** re-running the example lockfiles through `--reset-lockfiles`. It would produce a fresh float different from the one already tested, so the committed lockfiles stand and the docs point at the flag for the next refresh. Say the word if you'd rather the documented path be the literal one.

Process note taken — the squash body needs hand-editing at merge: this commit carries no task IDs, but `252f0e7c` still has three.

Local: unit 4398/4398 (345 suites), tooling 387/387, core eslint clean.

danielzytohoc (commented):
Re-reviewed `ae377b705`. All five actioned, and **you're right on finding 1 — my suggestion was wrong.** I anchored the strip at `/^\uFEFF/`, so it would have matched nothing at byte 1662. Your diagnosis is also better than the original write-up's: since a leading BOM is stripped by the decoder but a mid-file one isn't, the shipped stylesheet breaks too, which explains why the plain `UMD.min` legs went red rather than only the bundles that inline the string.

The fix is the right shape. I confirmed the ordering argument holds structurally, not just by comment: `stripCssBomPlugin` is pushed inside `config.forEach`, `addCssToJsExport` runs after the loop, and `afterEmit` is an `AsyncSeriesHook`, so registration order is execution order.

Verified rather than taken on trust:

| Claim | Result |
|---|---|
| Visual green on the prior head | Changed 0 / New 0 / Deleted 0 / Passed 1646, `approved=false`, no labels |
| Same on this head | Changed 0 / Passed 1646; the whole run is green including `CI Gate` |
| The widen floats nothing | Only the recorded override string moves (14 lines); postcss stays a single `8.5.23` |
| The new test actually runs | `PASS test/__tests__/cssBomStrip.unit.js`; suites 340 → 346, tests 4311 → 4430 |
| 3, 4, 5 and the nit | `--reset-lockfiles` referenced, throwaway procedure gone, `minimatch@9` dropped, no ASCII arrows left |

Two things left, neither blocking.

### 1. The strip's coverage is a property of current content, not of the build

`themes-css-production.js` runs the identical `css-loader` → `CssExtractRspackPlugin` → `LightningCssMinimizerRspackPlugin` chain and emits `../styles/[name].min.css`, with no strip registered. I checked all eight built minified stylesheets: only `handsontable.min.css` carries non-ASCII (one line), so it is the only asset that can acquire a charset today and the gap is not live.

But it holds only while the themes stay pure ASCII. The first `content:` arrow or non-ASCII icon fallback added to a theme gives that stylesheet a BOM at postcss ≥ 8.5.24, in the same position, with nothing to catch it — and the cap that previously made this class impossible is now gone. Registering the same plugin in `themes-css-production.js` is two lines and makes the invariant structural. Happy for it to be a follow-up, but I'd rather it not be left to be rediscovered.

### 2. `cssBomStrip.unit.js`'s last assertion is stronger than its name

`expect(pnpm.overrides.postcss).toBe('^8.5.23')`, titled *"should not cap postcss below 8.5.24"*, will go red on the next legitimate floor raise — which is exactly the event the range is designed to accommodate. `expect(pnpm.overrides.postcss).not.toMatch(/</)` states the actual invariant and survives a bump, with no new dependency.

The first two cases are properly behavioral. Cases 3 and 4 assert source text, so a rename reds them while a push into a dead branch would pass them; both could be behavioral cheaply. I'm not pressing that one — `esTarget.unit.js`, `previewPackaging.unit.js`, `optionLevels.unit.js` and `jsdocLinkSyntax.unit.js` all pin config invariants the same way, so it is the house style here.

### One correction for the record

"A re-run on the same commit produced the same six, so it is deterministic rather than random" no longer holds. Six changed on `252f0e7c`, zero on both `6c9ae1fe` and `ae377b705` — across commits that cannot affect rendering. That is run-to-run nondeterminism, not a stable sub-threshold delta, which turns follow-up 6 from an open question into a known problem. Worth capturing so the determinism conclusion isn't re-derived from two samples.

Agreed on not re-running the example lockfiles through `--reset-lockfiles` — it would trade a tested float for an untested one.

danielzytohoc (approved):

PR #13359, opened 2026-09-03
DEV-459: Apply a columns array before updateSettings' data phase

### LINKED ISSUE #5543: Error after update settings rendering twice and picking wrong renderer functions
### Description
Hi, I m having issues when I try to update Handsontable settings removing columns. 
After update settings, Handsontable renders two times, first time using wrong renderer functions and the second time using the right renderer functions.

### The expected behavior is:
- Update settings (even removing columns).
- Render one time
- Pick the right renderer function

### Steps to reproduce

1. Init displaying two columns with different renderer functions (eg. number, string)
2. Remove the first one (in this case number)
3. Handsontable try to render string column using number renderer function.
4. If the renderer logic between number & string renderer functions is incompatible, then the app was broken.

### Demo
v.1.16.0 http://jsfiddle.net/angelfragap/brd6mc8n/
v.6.1.1 http://jsfiddle.net/angelfragap/qgzmbf0c/

- Both demos contain a button to apply the update. 
- First, render produce the console output like this: 

image
- It s fine

- After clicking on the button and apply the changes, the console displays like this.

image

### Environment
* Handsontable version: at least from [1.16.0 ... 6.1.1 ]
* Browser Name and version: All
* Operating System: All

I would like to allow the users to choose which columns want to display. Therefore the table columns config should change on the fly.

Now the workaround is to check if the incoming columnKey should use the current renderer function.

If it is really necessary to render two times, then that is no a problem, but at least using the right renderer the first time could be nice.

Debugging a little bit, I found out that inside the `updateSettings`, `priv.columnSettings` is not being recalculated before the first render, then old columns settings are applied to render the first time.

demtario:
### Context

The PR fixes `updateSettings()` applying a shortened `columns` array too late, so one render and one round of source-data validation ran against the previous column configuration.

`updateSettings()` merges the payload into `tableMeta` up front, so `tableMeta.columns` already holds the new value, but the column meta layer was rebuilt much later (`core.ts`, the clear at the old line 3921 and the loop at 3933). When the same payload also carries `data`, `instance.updateData` reaches `dataMap/replaceData.ts`, which renders at `:164` and runs `runSourceDataValidators` at `:157`. Both resolve per-column configuration from the column meta layer, and both ran before it was rebuilt. The column count was already new by then, because `updateData` fits the index mapper from `getInitialColumnCount()` (`core.ts:3465`), which reads the merged setting.

Two effects, both reproduced in a real browser against `18.1.0`:

1. A cell renderer was called for a column that had moved into that index, with a value belonging to another column. The reporter's application threw on it. A second render repairs the DOM, which is why the bug survived since 2018: the settled grid looks correct.
2. The removed column's `sourceDataValidator` was applied to the surviving column. With `allowInvalid: false` the rejected values were written back as `null` (`dataMap/sourceDataValidator.ts:89`), so the surviving column's source data was silently blanked. No later render repairs that.

The fix extracts the existing clear-and-replay and the `columns` rebuild loop into two local helpers, and runs them before the data phase when the payload names `columns` and the setting is an array. An array's `length` is a physical column count that needs no dataset, so it is correct there whatever `data` does in the same call. The pre-existing late pass is kept and stays authoritative: it covers the `columns` function form (whose column count is genuinely unknown that early, and which must not be probed twice per call), `maxCols`, and any column the data phase itself created. Its clear is skipped when the early pass already did it.

The number of renders and the firing order of every hook are unchanged.

The change also repairs a second case that fell out of the same ordering: the late loop is bounded by `countCols()`, the not-trimmed count, while `columns` addresses physical columns. On a grid with a trimming map that bound stopped short, and the last physical column kept the default renderer instead of its own.

- Unit tests added/modified (`*.unit.js`): `handsontable/src/__tests__/core/updateSettings.unit.js`
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`): `tests/e2e/update-settings-column-shrink.spec.ts`, with `tests/fixtures/demo/update-settings-column-shrink.html` and `tests/fixtures/pages/UpdateSettingsColumnShrinkPage.ts`
- Type tests (`*.types.ts`) updated if public API changed: none, the public API is unchanged
- For a bug fix, the spec that fails without this fix: three unit cases (`should apply the new column configuration before the data phase reads it`, `should validate the source data against the new column configuration`, `should apply the columns array to every physical column when one is trimmed`) and two Playwright cases (`when data travels along`, `when a theme travels along`). Verified by reverting `core.ts` and re-running: 3 unit failures and 2 Playwright failures. The remaining cases pass either way on purpose, as guards.
- Demo page / recorded trace (for UI changes): `handsontable/dev-pr.html` and `dev-latest.html`, generated by the demo-page skill. Each renderer records its calls and flags a value whose shape belongs to another column, because the settled DOM is correct even on the broken build. Before: `pass 1: col 0 (prop "string") -> numberRender, first value "Alpha" <<< VIOLATION`, `violations: 4`, and the second grid's source values read `[null,null,null,null]`. After: `violations: 0` and `["Alpha","Bravo","Charlie","Delta"]`.

### Commands run

```
npm --prefix handsontable run eslint                # 0 errors (580 pre-existing warnings)
npm --prefix handsontable run test:types            # pass
npm --prefix handsontable run test:unit             # Tests: 4317 passed, 4317 total
npm --prefix handsontable run test:e2e              # 9849 specs, 0 failures, 8 pending
cd tests && npm run test:e2e                        # 820 passed
npm --prefix wrappers/react-wrapper run test        # Tests: 99 passed, 99 total
npm --prefix wrappers/vue3 run test                 # Tests: 39 passed, 39 total
npm --prefix wrappers/angular-wrapper run test      # Tests: 299 passed, 299 total
```

### Types of changes

### Related issue(s):

1. #5543
2. DEV-459

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

### Follow-ups, deliberately not in this PR

1. The late pass still passes a visual index to `metaManager.updateColumnMeta()`, which takes a physical one. The early pass is physical-correct and covers the reported shapes; the residual is a payload that omits `columns` on a grid with a non-identity column mapper.
2. The clear at the old line 3921 is gated on the payload naming `cell`, `cells` or `columns`, while the rebuild loop always runs off `tableMeta.columns`, and `columnMeta.updateMeta` only extends. Widening that gate drops plugin-written cell meta that the `_userDefinedMetaProps` replay does not restore, so it needs its own reasoning and tests.
3. Columns created by `grid.adjustRowsAndCols()` for `minCols` and `minSpareCols` still receive no column-meta pass in the same call.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-459

---

> [!NOTE]
> **Medium Risk**
> Touches core `updateSettings` ordering for a common wrapper pattern (`data` + `columns`); behavior change is intentional but `columns` function form still has the known early-ordering gap noted in the PR.
> 
> **Overview**
> Fixes **#5543**: when `updateSettings()` sends a **shorter `columns` array with `data`**, `tableMeta.columns` was already new but column meta was rebuilt **after** `updateData` rendered and ran `sourceDataValidator`. That briefly used the **previous** column’s renderer and validator (wrong DOM paint; with `allowInvalid: false`, surviving source values could be nulled).
> 
> **`updateSettings` in `core.ts`** now runs an **early pass** for **array** `columns`: shared `resetMetaCaches()` + `applyColumnMeta()`, bounded by **`columns.length`** (physical indexes, not `countCols()`), **before** the data load. The existing late pass still handles `columns` as a **function** and updates that omit `columns`; when the early pass ran, the late clear/apply is skipped to avoid double work on wrapper re-renders.
> 
> Adds unit coverage in `updateSettings.unit.js`, Playwright e2e + fixture for renderer mismatches during the bad intermediate render, changelog **5543**, and an **AGENTS.md** pitfall documenting the two-pass behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ce579aab9a6a20756f949df320d3a792cb2d0581. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-459 Error after update settings rendering twice and picking wrong renderer functions](https://app.clickup.com/t/9015210959/DEV-459)

handsontable/src/core.ts
```diff
@@ -3873,6 +3872,68 @@ export default function Core(
...
+    // has no such count before the data phase (`getInitialColumnCount()` would read the previous
+    // dataset width), and probing it here would also call the user's function a second time per
+    // column, per call. The loop further down stays authoritative for both forms.
+    const earlyColumnsSetting = settings.columns !== undefined && Array.isArray(tableMeta.columns)
+      ? tableMeta.columns
+      : null;
+
+    if (earlyColumnsSetting !== null) {
+      resetMetaCaches();
+      applyColumnMeta(Math.min(earlyColumnsSetting.length, tableMeta.maxCols));
```
budnix: `maxCols` caps `countCols()`, which is a count of columns on screen, but this loop walks physical column indexes. Set `maxCols: 2` with `columns: [A, B, C]` and trim physical column 0: the columns on screen are physical 1 and 2, but both passes stop at physical 1, so physical 2 renders with nothing from `columns`. That is the same gap the `TrimmingMap` test further down was added to close, and the test named "in agreement under `maxCols`" currently pins the clamp as intended. Worth a look at whether the bound should just be the array's own length.
```suggestion
      applyColumnMeta(earlyColumnsSetting.length);
```

handsontable/src/core.ts
```diff
@@ -3913,38 +3974,17 @@ export default function Core(
...
-            : columnSetting[j];
-
-          if (column) {
-            metaManager.updateColumnMeta(j, column);
-          }
-        }
-
-        j += 1;
-      }
+      applyColumnMeta(clen);
```
budnix, line 3995: For an array `columns`, `getInitialColumnCount()` returns `columns.length`, so `clen` can never be larger than the early bound and every column this touches was already written a few lines up. The React and Angular wrappers re-send `columns` on every commit, so a wide grid pays `extend()` plus `extendByMetaType()` twice per column on each render. Skipping this when the early pass ran would still leave the function form and the `columns`-absent payload working. Curious whether that was considered and ruled out.

handsontable/src/core.ts
```diff
@@ -3873,6 +3872,68 @@ export default function Core(
...
+    // would otherwise paint the new data with the PREVIOUS column's renderer, and validate it against
+    // the previous column's rule, blanking the source value when `allowInvalid` is `false`
+    // (GitHub issue #5543).
+    //
+    // Only the array form is resolved this early. Its length is a physical column count that needs no
+    // dataset, so it is correct here whatever `data` does in the same payload. A `columns` FUNCTION
+    // has no such count before the data phase (`getInitialColumnCount()` would read the previous
+    // dataset width), and probing it here would also call the user's function a second time per
+    // column, per call. The loop further down stays authoritative for both forms.
+    const earlyColumnsSetting = settings.columns !== undefined && Array.isArray(tableMeta.columns)
```
budnix, line 3928: The array-only gate leaves the same corruption in place for `columns: (i) => ({ ... })`, since the data phase still runs `runSourceDataValidators` and the render against the old column meta, so a removed column's `sourceDataValidator` can still blank the surviving column's source data under `allowInvalid: false`. The comment above explains why the function form cannot be resolved this early, and that reasoning makes sense. Did this end up as a follow-up ticket, or is the plan to leave the function form as it is? Either way it seems worth saying somewhere a user will see, because the changelog reads as a full fix.

handsontable/src/__tests__/core/updateSettings.unit.js
```diff
@@ -193,4 +194,235 @@ describe('Core#updateSettings', () => {
...
+      core.columnIndexMapper.updateCache(true);
+
+      expect(core.countCols()).toBe(2);
+
+      core.updateSettings({
+        data: [{ a: 1, b: 2, c: 3 }],
+        columns: buildColumns(),
+      });
+
+      const metaManager = core._getMetaManager();
```
budnix, line 363: The `countCols()` check above runs before the update, but the thing this test pins is that the late pass is bound by `countCols()` after it. `updateData` calls `columnIndexMapper.fitToLength(3)` and `initIndexMappers()` in between, and nothing here proves the trimming map came through at its old value. If it ever stops surviving that, `clen` becomes 3, the late pass covers physical 2 on its own, and this test goes green on unfixed code too.
```suggestion
      // The late pass is bound by `countCols()`, so pin it again after the update.
      expect(core.countCols()).toBe(2);

      const metaManager = core._getMetaManager();
```

tests/e2e/update-settings-column-shrink.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+    await grid.expectSurvivingColumn();
+  });
+
+  test('keeps every renderer on its own column when a theme travels along', async({ page, theme, bundle }) => {
+    const grid = new UpdateSettingsColumnShrinkPage(page, theme, bundle);
+
+    await grid.goto();
+    await grid.shrinkWithTheme();
+
+    // A theme in the payload paints from `useTheme()`, before the data phase. That
```
budnix: The fixture sets the container to `ht-theme-<theme>` and then sends `theme: 'ht-theme-<theme>'`, so `currentThemeName !== settings.theme` is false and `useTheme()` never runs. `applyThemeOverrides()` also returns at its first guard because the payload has no `colorScheme` and no `density`, so no extra paint happens and this test does the same work as the first one. Sending a theme that differs from the one on the container would make it match what the comment describes. Would be good to check the horizon project too, since it has the same shape.

tests/fixtures/pages/UpdateSettingsColumnShrinkPage.ts
```diff
@@ -0,0 +1,110 @@
...
+  renderer: string;
+  col: number;
+  prop: string;
+  violation: boolean;
+};
+
+declare global {
+  interface Window {
+    htRendererCalls: RendererCall[];
+    hot: { countCols(): number };
```
budnix: `windowTypes.ts:145` already augments `Window` with `hot: FixtureHotInstance`, and declaring it here with a different type is a TS2717 duplicate. `EmptyValuePage.ts` ran into the same thing and leaves `hot` out on purpose, with a comment saying why. Playwright transpiles without type checking so CI stays quiet, but an editor with both files open goes red. Adding `countCols(): number` to `FixtureHotInstance` and dropping this line keeps one home for the augmentation.
```suggestion
    htRendererCalls: RendererCall[];
```

.changelogs/5543.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed `updateSettings()` applying a `columns` array too late when the same call also changed `data`, which could hand a cell renderer a value belonging to another column and let a removed column's `sourceDataValidator` blank the surviving column's source data. A `columns` array now also reaches every physical column on a grid with trimmed columns.",
```
budnix: The second sentence says a `columns` array now reaches every physical column on a grid with trimmed columns, but with the `maxCols` clamp in `core.ts` that only holds when `maxCols` is unset. The whole title runs about 350 characters and leans on `sourceDataValidator` and "physical column", which a release-notes reader has nothing to hang those on. Cutting it to the user-visible effect would read better, something like: a shortened `columns` array sent together with `data` no longer renders or validates a column with the previous column's settings.

handsontable/src/core.ts
demtario: You are right, and this was a real defect rather than a style point. Fixed in ce579aab9a: the bound is now `earlyColumnsSetting.length`.

Two AI reviewers and I had all looked at this clamp and cleared it. One of them raised it as a HIGH finding on the grounds that `clen` is unclamped, which is wrong (`countCols()` is `Math.min(tableMeta.maxCols, ...)` at core.ts:5790-5795), so I checked `maxCols` with a longer `columns` array and no trimming, measured identical behavior before and after the change, and closed it. Your case adds the trimming, which is exactly what makes the physical and on-screen spaces diverge, and my check had ruled it out by construction.

Verified rather than reasoned this time. The `maxCols` test is rewritten as your scenario (`maxCols: 2`, three `columns` entries, physical column 0 trimmed) and asserts all three physical columns keep their own renderer. It is red against the clamped version I had pushed:

    ✕ should apply the `columns` array by physical index when `maxCols` caps a trimmed grid
    Tests: 1 failed, 14 passed

The comment at the call site now records why the bound is the array's length and not `countCols()`, with your scenario spelled out, so the next reader does not have to re-derive it.

handsontable/src/core.ts
demtario: Considered and kept deliberately, which was the wrong call. Adopted in ce579aab9a: the late pass now runs only when the early one did not.

My reasoning had been that both passes are `extend`, so the duplicate is idempotent and keeping it made the late pass the single authority for every payload shape. That ignored the cost you point at, and the redundancy is provable rather than incidental: for an array `getInitialColumnCount()` returns `columns.length`, the mapper is fitted to exactly that, and `countCols()` is `Math.min(maxCols, notTrimmedIndexesLength)`, so `clen` can never exceed the bound the early pass already used.

The function form and the `columns`-absent payload still go through it, which is what the "should re-read a `columns` function on an update that does not name it" test covers. The `minCols`/`minSpareCols` columns are unaffected either way, since `grid.adjustRowsAndCols()` runs after both passes and never had a column-meta pass in the same call. That gap is listed as a follow-up in the PR description.

handsontable/src/core.ts
demtario: Correct, and the plan is to leave the function form as it is for now. Your point about the changelog overclaiming is fair, so that is fixed rather than argued.

Why it cannot come along: before the data phase the function form has no usable column count. `getInitialColumnCount()` reads `countSourceCols()` for array-typed data and `datamap.colToPropCache.length` for object- and function-typed data, both of which still describe the previous dataset at that point, so the early pass would probe the wrong index range. It would also call the user's function a second time per column per call, and a `columns` function is documented as free to read external state, so a double call is observable.

What changed for visibility:

- The changelog title no longer reads as a full fix. It is now scoped to the reported shape: "Fixed a shortened `columns` array, sent together with `data`, rendering and validating a column with the previous column's settings."
- `handsontable/AGENTS.md` records the gap explicitly, in the bullet that documents the two-pass ordering: "Still open, deliberately: the function form keeps the original defect, because it cannot be resolved this early."

No ClickUp ticket yet. I would rather not open one without the maintainer deciding whether it belongs as its own task or as a note on DEV-459, so say which and I will file it. Fixing it properly means giving the function form a trustworthy pre-data-phase column count, which is its own change with its own reasoning about how many times the callback may run.

handsontable/src/__tests__/core/updateSettings.unit.js
demtario: Right, the pre-update check pinned the wrong moment. Added in ce579aab9a, with the reason in a comment so it does not get tidied away later as a duplicate assertion:

```js
// Pinned again after the update: `updateData` runs `fitToLength()` and `initIndexMappers()` in
// between, so the count this test depends on has to be the one that holds afterwards. Without
// this, a trimming map that stopped surviving the update would take `countCols()` to 3 and let
// the assertions below pass on unfixed code.
expect(core.countCols()).toBe(2);
```

The `maxCols` test that grew out of your first comment takes the same lesson: its only `countCols()` assertion sits after the update, for the same reason.

tests/e2e/update-settings-column-shrink.spec.ts
demtario: Confirmed and fixed in ce579aab9a. The test was doing the same work as the first one, exactly as you describe: the fixture set the container to `ht-theme-<theme>` and then sent `theme: 'ht-theme-<theme>'`, so `getCurrentThemeName() !== settings.theme` was false and `useTheme()` never ran.

The fixture now sends a theme that differs from the one on the container, picked per leg so it differs on all six:

```js
const otherTheme = window.htTheme === 'main' ? 'horizon' : 'main';
```

Rather than trust that this restored the extra paint, the fixture counts render passes on `beforeRender` and the spec asserts the count. Measured first, then asserted:

    data:        passes=2  calls=20
    theme:       passes=3  calls=35
    columnsOnly: passes=2  calls=12

So the `data` case asserts 2 and the theme case asserts 3, and neither can silently become the other again. Ran on the full matrix rather than `e2e-main` alone, per your note about horizon: 18 passed (3 cases across all six projects, `-min` legs included).

tests/fixtures/pages/UpdateSettingsColumnShrinkPage.ts
demtario: Fixed in ce579aab9a. `windowTypes.ts` already carries `countCols(): number` on `FixtureHotInstance` (line 118), so the declaration was pure duplication.

Went with the `EmptyValuePage.ts` pattern you point at rather than extending `FixtureHotInstance`, since it is the one already documented in the tree, and left its reasoning in the file so the next page object does not re-add the augmentation. `Window` now gets only the two fixture globals, and the grid slice is a local interface used at the call site.

Checked with `tsc --noEmit` over the spec together with `windowTypes.ts` and `EmptyValuePage.ts`, since the duplicate only shows up when both files are in one program: clean, no TS2717. Worth noting for anyone verifying this the same way, a single-file `tsc` run is misleading here. On its own the page object reports `TS2339: Property 'hot' does not exist`, because `windowTypes.ts` is not pulled in, and adding it to the program is what reproduces what the editor sees.

.changelogs/5543.json
demtario: Both points taken. Shortened in ce579aab9a to the user-visible effect, close to your wording:

> Fixed a shortened `columns` array, sent together with `data`, rendering and validating a column with the previous column's settings.

That drops `sourceDataValidator` and "physical column" from the release notes, and drops the second sentence entirely. The sentence was doubly wrong: your first comment removed the `maxCols` clamp it was hedging around, and the trimmed-column behavior it described is an internal consequence rather than something a release-notes reader can act on. It is recorded in the `handsontable/AGENTS.md` bullet and pinned by two unit tests instead.

One entry, not two. The trimmed-column repair is not a separately reachable user-facing change: it falls out of bounding the pass by the array's own length, and both halves are the same defect seen from two angles, so splitting it would put two overlapping lines in the same `CHANGELOG.md` section.

budnix (approved):

demtario:
cursor review

PR #13360, opened 2026-09-03
DEV-465: Unmerge the destination when pasting over a merged cell

### LINKED ISSUE #5092: Allow to unmerge cells while copy/paste
I see the code，and I think you may forgot to remove the mergedCellsCollection.

demtario:
### Context

The PR fixes pasting a multi-cell block over a merged cell. The merge stayed in place, so the pasted values landed in the source data of cells the merge keeps `hidden: true` and were never rendered: the grid showed one value while `getData()` and every export returned nine. The paste did not silently do nothing — it silently wrote data nobody could see.

Reported in 2018 as #5092, confirmed there against 2.0.0 and 8.0.0, and still reproducible on `develop`. A maintainer confirmed in that thread that Excel and Google Sheets unmerge the destination in this case and that Handsontable should follow the same scheme.

#### Root cause

Three gaps on the merge/clipboard boundary; the first two are fixed here.

1. **`MergeCells` never observed a paste.** It registered no `beforePaste`/`afterPaste`, and its `afterChange` listener returned early for every source except `Autofill.fill`. A `CopyPaste.paste` write therefore reached `populateFromArray` → `setDataAtCell` and landed under the merge untouched.

2. **Selection expansion made the whole merge the paste target.** `beforeSelectionHighlightSet` expands any selection touching a merge to the merge's full rectangle, and `CopyPaste#populateValues` then tiles the clipboard across `max(clipboard extent, selection extent)`. So even a single pasted value arrived as one change per covered cell — nine changes for a 3x3 merge. Unmerging without addressing this would have turned one Ctrl+V into nine *visible* copies of one value, so the two are one decision, not two.

#### Behavior

| Case | Behavior |
|---|---|
| Multi-cell clipboard over a merged range | Every merged range the write reaches is unmerged; all pasted values become visible |
| Single value over a merged range | Merge is kept; only its top-left cell is written, covered cells stay empty |
| Undo | The geometry travels with the paste's own undo entry, so one `undo()` restores the values and the merges together rather than in two steps |

`beforeChange` records the affected ranges while they are still in the collection, and `afterChange` drops them once the data has landed. Recording before the write is what lets the geometry ride inside the paste's own `DataChangeAction`.

#### Three timing traps this had to survive, each with a spec

All three were found in review, and each one made the fix silently wrong rather than merely incomplete. They are the parts most worth a reviewer's attention.

1. **`afterPaste` fires before `afterChange` on a validated column.** As soon as any pasted cell carries a validator — `type: 'numeric'`, `type: 'date'`, `dropdown` and `autocomplete` all install one — `validateChanges` defers `applyChanges` to a microtask, so the order is `beforeChange` → `afterPaste` → `afterChange`. An earlier revision cleared the record on `afterPaste`, which made the whole fix inert on any validated column while the undo action still recorded the geometry, so redo dropped a merge the paste never dropped. Nothing may clear the record outside the `beforeChange`/`afterChange` pair.

2. **A paste's validation window carries other changes, and none of them may inherit its geometry.** A validator that corrects a value writes through `setDataAtCell` with a `*Validator` source (the documented convention in `validators/__tests__/validatorCorrection.spec.js`), once per corrected cell, and each of those raises a `beforeChange` of its own. `collectMergedCellsDestroyedByChange` therefore takes the `source` and returns nothing unless it owns the snapshot. Measured on a nine-cell paste with a correcting validator: without the check, all **ten** stack entries carried the geometry, and undoing a single correction re-formed the merge on top of the eight values the paste had already written — the same "shown value is not the stored value" defect this PR exists to remove, reproduced mid-undo.

3. **The record is measured after other listeners have had their say.** `beforeChange` is registered at `orderIndex: 900`, after every ordinary listener and before `DataChangeAction`'s at 1000. A listener that vetoes part of a paste by nulling its entries then shrinks — or empties — what gets recorded, instead of leaving a snapshot that describes a write that never happened. Without this, a fully vetoed paste left a stale record that the next unrelated edit's undo entry picked up, so undoing that edit re-merged a range the user had since unmerged by hand.

Two smaller notes:

- **`countPartials` is on, and each range is unmerged through its own range.** The written rectangle can reach past the selection and clip a merged range whose top-left corner the selection never touched, and `unmergeRange` matches on that corner.
- The geometry type and its range builder live in the core `src/utils/mergeAreas.ts` rather than in either plugin, following the precedent for `movableMeta.ts`: registering only `UndoRedo` must not pull `MergeCells` into the bundle.

#### Deliberately out of scope

- Copy does not carry the merge — a merged range still copies as its top-left value plus empty cells.
- Pasted HTML `rowspan`/`colspan` still flattens without creating a merge. `htmlToGridSettings` already computes a `mergeCells` array and `CopyPaste` still discards it.

Both are documented in the guide.

#### Correction to an earlier version of this description

An earlier revision of this PR and of the guide claimed that `pasteMode: 'shift_down'` / `'shift_right'` leaves merges alone, on the reasoning that `populateFromArray` recurses without forwarding `source`. Measured on a live grid, rebuilt from scratch per mode: **all three paste modes unmerge the destination**, every change arrives with source `CopyPaste.paste`, and no row or column shift occurs. Both halves of the claim were wrong, so the claim was removed from the guide rather than replaced with another spot-check. How these modes interact with merges beyond this case is not characterized here.

### How has this been tested?

Every new spec was verified to fail without the corresponding piece of the fix, not just to pass with it.

```bash
# unit + type + lint

# the areas this touches
npm run test:e2e --prefix handsontable -- --testPathPattern=mergeCells
npm run test:e2e --prefix handsontable -- --testPathPattern=copyPaste
npm run test:e2e --prefix handsontable -- --testPathPattern='undoRedo|UndoRedo'

# full suites
```

Added:

- `mergeCells/__tests__/mergeCells.spec.js` — new `cooperation with the CopyPaste plugin` block: multi-cell paste unmerges and every value renders; a single value keeps the merge and writes only the anchor; a merged range reached only by the paste overflow is unmerged too; a paste touching no merge leaves the collection untouched; HTML spans still flatten with the plugin disabled. Two further cases repeat the first two on a validated column, where the hook order differs.
- `mergeCells/__tests__/undoRedo.spec.js` — new `paste over a merged cell` block: one undo restores data and geometry and the stack goes 1 → 0; redo re-applies both; a single-value paste records one change and never touches the merge; a paste/undo/redo round-trip on a validated column; and a correcting validator's own changes do not inherit the geometry.
- `mergeCells/__tests__/utils.unit.ts` — new file covering `getRangeFromChanges` (min/max derivation, prop-keyed columns, entries another listener nullified) and `toMergeAreaKey`.
- `undoRedo/__tests__/actions/dataChange.spec.js` — the existing action-shape assertion now includes the new `mergedCells: []` field.

Manual, in a browser against the local build: copied `G3:I5`, selected the `G6:I8` merge, pasted — the merge is dropped and all nine values render. One Ctrl+Z restores `G6` in a re-merged range; Ctrl+Shift+Z re-applies the paste. A single-value paste leaves the merge intact. Each of the three timing traps above was reproduced and then re-checked after its fix, on a grid with a plain validator, with a correcting validator, and with a `beforeChange` listener that vetoes the paste.

### Types of changes

### Related issue(s):

1. #5092
2. DEV-465

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-465

---

> [!NOTE]
> **Medium Risk**
> Touches core change/paste/undo pipelines and hook ordering (validators, async applyChanges); behavior change is intentional but affects common edit flows when merges and clipboard interact.
> 
> **Overview**
> Fixes **#5092**: pasting more than one cell onto a merged range now **drops that merge** (and any other merge the paste touches), so every pasted value is visible—matching Excel/Google Sheets. A **single-value** paste still keeps the merge and updates only the top-left cell (extra `beforeChange` logic stops the tiled “nine copies of one value” behavior).
> 
> **MergeCells** hooks into `CopyPaste.paste` via prioritized `beforePaste` / `beforeChange` / `afterChange`: snapshot affected merges before the write, unmerge after data lands, with careful handling for validated columns, paste overflow past the selection, and skipped/read-only cells.
> 
> **Undo/redo** stores destroyed merge geometry on `DataChangeAction` (`mergedCells`), scoped to source `CopyPaste.paste` so validator corrections don’t inherit it; one undo restores values and merges together. Shared **`mergeAreas`** utils support geometry without bundling MergeCells into UndoRedo-only builds.
> 
> Docs and changelog added; broad unit/e2e coverage for paste, undo, and edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 47c417a78be4cb1e06ee3e6ab63cca94379cfe86. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-465 Allow to unmerge cells while copy/paste](https://app.clickup.com/t/9015210959/DEV-465)

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -1943,19 +1994,162 @@ export class MergeCells extends BasePlugin {
...
+      return [];
+    }
+
+    const { from, to } = getRangeFromChanges(this.hot, changes);
+    const topStart = this.hot._createCellCoords(from.row, from.column);
+    const bottomEnd = this.hot._createCellCoords(to.row, to.column);
+    const range = this.hot._createCellRange(topStart, topStart, bottomEnd);
+
+    return this.mergedCellsCollection
+      .getWithinRange(range, true)
```
budnix: Can I ask whether a merge that receives no change at all should survive here? `populateFromArray` drops the change for every cell that is `readOnly` (`core.ts:1790`), `skipRowOnPaste` or `skipColumnOnPaste`, so a merge sitting only on such cells is still inside the min/max box of the changes and gets unmerged with nothing written into it. Matching against the real change coordinates instead of the box would still catch the overflow clip the comment above describes, since those cells do produce changes. A spec that pastes across a `skipColumnOnPaste` column holding a merge would pin this down.

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -1943,19 +1994,162 @@ export class MergeCells extends BasePlugin {
...
+      return;
+    }
+
+    // `unmergeRange` renders on its own, so a multi-merge paste would otherwise redraw once per
+    // merge.
+    this.hot.batchRender(() => {
+      snapshot.forEach((mergeArea) => {
+        this.unmergeRange(toMergeAreaRange(this.hot, mergeArea), true);
+      });
+    });
```
budnix: `batchRender` has no `try`/`finally` (`core.ts:3206`), and `unmergeRange` still runs `beforeUnmergeCells` and `afterUnmergeCells` when `auto` is on — `auto` is only passed along, it does not skip the hooks. `runHooks` does not catch (`core/hooks/index.ts:254`), so one throwing user listener skips `resumeRender()` and the grid never draws again for the rest of its life. `undoRedo/utils.ts` has the same shape at lines 128 and 158, with `mergeRange`'s two hooks on top.

```suggestion
    this.hot.suspendRender();

    try {
      snapshot.forEach((mergeArea) => {
        this.unmergeRange(toMergeAreaRange(this.hot, mergeArea), true);
      });
    } finally {
      this.hot.resumeRender();
    }
```

docs/content/guides/cell-features/merge-cells/merge-cells.md
```diff
@@ -371,6 +371,61 @@ hot.updateSettings({
 hot.getDataAtCell(0, 1); // -> null, cleared as usual
 ```
 
+## Copying and pasting over merged cells
+
+Pasting a block of more than one cell over a merged range unmerges that range, and every pasted value becomes visible. Excel and Google Sheets behave the same way: a block with its own rows and columns cannot fit inside a single merged cell, so the merge gives way.
```
budnix, line 376: This is the first mention of Google Sheets on the page, and `.ai/DOC-STANDARDS.md` asks for the expanded trademark line once a page names both it and Excel. The page has no disclaimer at all right now, so this would be a good moment to add one at the end. `undo-redo.md` line 235 has the exact wording to copy.

handsontable/src/plugins/mergeCells/mergeCells.ts
demtario: Good catch, and it reproduced. Pasting a 2x3 block at `G6` across a `skipColumnOnPaste` column 7 that holds a merge:

```
changes:  5,6  5,8  5,9  6,6  6,8  6,9   <- column 7 skipped entirely
merge at (5,7):  gone
column 7 data:   "H6", null              <- never written
```

The min/max box spanned columns 6-9 and swallowed it. Switched to your suggestion: each change is now resolved on its own through `MergedCellsCollection#get` (`#mergeAtChange`), so only merges that actually receive a write are recorded. That covers `readOnly` and `skipRowOnPaste` too, and the overflow clip is unaffected because `get` matches from any covered cell rather than the top-left one.

Two specs added, both of which fail with the box restored while the overflow spec keeps passing:

- `should keep a merged cell that the paste skipped entirely (skipColumnOnPaste)`
- `should keep a merged cell built entirely of read-only cells`

One knock-on worth flagging: with the box gone, `getRangeFromChanges` had only its original autofill caller again, so I reverted that extraction. `calculations/autofill.ts` is back to its `develop` state and the `utils.unit.ts` I had added is removed. This round is a net -187/+169.

handsontable/src/plugins/mergeCells/mergeCells.ts
demtario: Confirmed on all three points: `batchRender` is `suspendRender()` / call / `resumeRender()` with no `try`/`finally` (`core.ts:3206`), and `unmergeRange` runs `beforeUnmergeCells` / `afterUnmergeCells` regardless of `auto` - `auto` is only forwarded into them. So one throwing user listener would strand the grid with rendering suspended for the rest of its life.

Applied your suggestion here, and the same shape at both sites in `undoRedo/utils.ts` (`remergeCellsGeometryOnly`, `unmergeCellsGeometryOnly`), where `mergeRange`'s own two hooks are exposed as well. Each call site now carries a comment naming why `batchRender` is not usable, so this does not quietly regress back to it.

docs/content/guides/cell-features/merge-cells/merge-cells.md
demtario: Added, copied verbatim from `undo-redo.md:235`:

> Microsoft and Excel are registered trademarks of Microsoft Corporation. Google Sheets is a trademark of Google LLC.

You are right that the page had no disclaimer at all before this - it named Excel only in the new section, so the requirement arrived with this PR.

budnix (approved):

demtario:
`visually-approved` because the diff is only the antialiasing issue, not actual difference

Issue #5661, opened 2018-12-12
The `cell` settings are erased after loading new `columns` via `updateSettings`


AMBudnik:
### Description

The `cell` settings are erased after loading new `columns` via `updateSettings`.

### Steps to reproduce

1. Click the button to see comments and `readOnly` state being erased

### Demo

https://jsfiddle.net/4jyq6wan/

### Your environment
* Handsontable version: 6.2.1
_It doesn't see to be version related_
* Browser Name and version: Chrome 71
* Operating System: Windows 10

inform [forum](https://forum.handsontable.com/t/gh-5661-update-setting-add-comments/2739)

AMBudnik:
Works the same in 8.0.0-beta1 https://jsfiddle.net/okeg3zsh/

budnix:
Thank you for reporting this.

I tested it on the latest release (18.1.0) and on `develop`, and it still reproduces. The `cell` settings are lost after `updateSettings({ columns: ... })`, exactly as your demo shows.

I also tested a few variants to map out the scope:

| Call | `cell` settings kept? |
|---|---|
| `updateSettings({ columns: [...] })` | no |
| `updateSettings({ cells() {...} })` | no |
| `updateSettings({ cell: [...] })` | yes |
| `updateSettings({ rowHeaders: true })`, or any other option | yes |
| `updateSettings({ data: [...] })` | yes |

What happens: `updateSettings()` clears the cell metadata cache when you pass `columns` or `cells`. It then re-applies the `cell` option only if that same call also passed `cell`. The button in the demo passes only `columns`, so nothing puts the comments and the `readOnly` flag back. The option itself is still there — `getSettings().cell` still returns the original array — but its effect on the cells is gone.

I have prepared a fix in #13361. Values declared through `cell` now survive `updateSettings()` calls that do not mention it, and they stay with their row through sorting and row moves. Passing `cell` again still replaces the previous entries, so `cell: []` still removes them all.

The fix will ship in one of the upcoming releases.

Until then, you can work around it by passing `cell` together with `columns` in the same call:

```js
hot1.updateSettings({
  columns: column2,
  cell: cellSettings, // the same array you passed at initialization
});
```

Issue #4446, opened 2017-08-03
`readOnly` state is being reset by `updateSettings`.


AMBudnik:
### Description

`readOnly` state is being reset by `updateSettings`.

### Steps to reproduce

1. Select cells from `0, 0` to `1, 1`
2. Use context menu to unselect `readOnly` option
3. Click the button

`readOnly` is still available for all cells.

If we add `setCellMeta` to set `readOnly: false` to a cell is it also skipped.

1. Go to demo2
2. Select cells from `1, 0` to `1, 1`
3. Use context menu to unselect `readOnly` option
4. Click the button

demo2: http://jsfiddle.net/9a1f461t/1/

### Demo

http://jsfiddle.net/9a1f461t/

### Your environment
* Handsontable version: 1.13.0
* Browser Name and version: Chrome 60
* Operating System: Windows 10

I have tried it with a few versions before without any change.

AMBudnik:
The main and secondary issues remain the same in version 0.35.1 - updated demo http://jsfiddle.net/wmtcpcyq/

m2a2x:
I'm getting same issue. 
We haveЖ
`cells: function (row, col, prop) {  var source = this.toVisualRow(row); (here always returns null)  return readOnly: true | false }`  
It's setting up ok, but after filtering indexes mess up.

AMBudnik:
Hi @m2a2x 
can you share a demo and a bug scenario? My demo doesn't apply filters. Your case may be different.

AMBudnik:
Works the same in v 5.0.1

AMBudnik:
Similar case

> 1. I use setCellMeta method store a id in cell
> 2. then i insert some row. I dynamic calculate the height,then update height property use updateSettings
> 3. my id in cell all lost

reported at [forum](https://forum.handsontable.com/t/updatesettings-clear-cellmeta-any-way-solve-it/2506)

aninde:
Still broken with v12.0.0 http://jsfiddle.net/rhq9g0uz/

demtario:
Fix was just merged in https://github.com/handsontable/handsontable/pull/12811, Im closing this issue.

PR #13361, opened 2026-09-03
DEV-2774: Keep the `cell` option across updateSettings

budnix:
### Context

The PR fixes GitHub issue #5661, open since 2018: values declared through the declarative [`cell`](https://handsontable.com/docs/api/options/#cell) option are erased by `updateSettings({ columns })` and `updateSettings({ cells })`. Reproduced on released **18.1.0** and on **develop**.

The reporter's [demo](https://jsfiddle.net/4jyq6wan/) declares two comments and a read-only cell through `cell`, then swaps `columns` on a button click — after the click all three are gone.

What makes this a bug rather than a documented quirk: after the wipe, `getSettings().cell` still returns the original array. The option value survives, only its effect on the cells is thrown away, so `getSettings()` reports something the grid is no longer doing.

**Cause.** `updateSettings` clears the whole cell-meta cache whenever `columns`, `cells`, or `cell` is passed, and then re-applies the `cell` option only when *that same call* passed it. The replay sitting next to it, added for #4446, cannot help: the option's writes are made with user-defined recording suspended **on purpose**, so that `updateSettings({ cell: [] })` can still delete previously declared entries — which excludes them from that snapshot by construction. #4446 is released and works; it simply covers a different origin.

**Fix.** Every `setMeta` write is now filed under exactly one origin, and the origin decides what survives the clear:

| Origin | Scope open at write time | Survives the clear |
| --- | --- | --- |
| imperative `setCellMeta` | none | yes — replayed (already true, #4446) |
| the declarative `cell` option | `startCellOptionMetaRecording()` | yes — replayed, unless the call restates `cell` |
| plugin-declarative (`Core#_setCellMetaDeclarative`) | `disableUserDefinedMetaRecording()` | no — dropped, the plugin re-applies it |

Both replays live in `resetMetaCaches()` and use the **physical** coordinates read from the storage keys, so a value returns to the record it was resolved to rather than to whatever record now sits at the option's visual position. The newest write for a key moves it between buckets, so an imperative override of a `cell`-declared value survives with no special case.

Two simpler-looking fixes were rejected, and both are worth stating because they are the first thing a reader reaches for:

1. **Falling back to `tableMeta.cell` at the re-application site.** It re-applies by *visual* index. That is harmless on the `columns` path, where the index mappers have just been reset, but on the `cells` path nothing resets them — so on a sorted grid the read-only cell would visibly jump to a different record.
2. **Deriving the bucket as `_persistedMetaProps` minus `_userDefinedMetaProps`.** That sweeps in `_setCellMetaDeclarative`, whose writes `ColumnSummary` *depends* on being wiped: it re-applies `readOnly` and `columnSummaryResult` from its endpoints on `afterUpdateSettings`, and a replayed value would strand a stale summary class on a cell the new endpoints no longer cover. Hence the separate third bucket.

Four further trade-offs are deliberate, and are recorded in `handsontable/src/dataMap/metaManager/AGENTS.md` so they are not re-litigated: the bucket test being an equality of the two scope counters rather than `> 0`; a plugin write on a `cell`-declared key dropping that key from the replay; the replay not firing `beforeSetCellMeta`/`afterSetCellMeta`; and the replay not being bounded to the live grid (which the imperative replay has never been either).

**Behavior changes**, both improvements over losing the values outright:

1. `cell` persists across `updateSettings()` calls that do not mention it.
2. A `cell` value later overridden with `setCellMeta` keeps the override across those calls.

Unchanged and pinned by specs: restating `cell` replaces every previously declared entry (`cell: []` removes them all); a restated value still wins a direct conflict; `loadData` still resets cell states while `updateData` does not.

This is not a breaking change — the only callers affected are ones that relied on the values disappearing.

Two incidental fixes came out of review, both in the code this PR already touches. `updateSettings({ cell: null })` used to clear the meta cache and then throw on `null.forEach`, leaving the instance with its cache wiped and the rest of the update never run (`isDefined(null)` is `true`; the test is now `Array.isArray`). And the entry's `row`/`col` were stored as cell meta — the entry's *visual* coordinates on an object `getCellMeta` stamps with *physical* ones — so they are now stripped before the write.

Documentation updated in the same change: the public `cell` option Typedoc, the `updateSettings` JSDoc (which documented the old contract, and carried the same paragraph twice), the `_setCellMetaDeclarative` JSDoc that claimed it behaved "exactly like the `cell` option", `handsontable/src/dataMap/metaManager/AGENTS.md`, and `handsontable/.ai/META-MANAGEMENT.md`.

### How has this been tested?

16 new E2E specs in a `(#5661)` block next to the existing `(#4446)` one, and 12 new unit tests across the two meta-layer suites. The new coverage includes the reporter's exact scenario, the restate-replaces and `cell: []` rules, the imperative-override rule, a `manualRowMove` case proving the replay is physical rather than visual, a `columns`-value-still-reaches-a-replayed-cell case, `cell` and `columns` in a single payload, the `loadData` / `updateData` contrast, `ColumnSummary`'s meta landing in no bucket, and the two scope-nesting rules.

```bash
npm run test:e2e --prefix handsontable -- --testPathPattern=setCellMeta
npm run test:e2e --prefix handsontable -- --testPathPattern=columnSummary
npm run typecheck --prefix handsontable
```

Results:

- Core unit suite: **4323 passed, 0 failed**
- Full core E2E suite: **9865 passed, 0 failed**, 8 pending
- `--testPathPattern=setCellMeta`: **37 passed, 0 failed**
- `--testPathPattern=columnSummary`: **63 passed, 0 failed** — the integration guard for the third bucket
- `tsc --noEmit`: clean; `eslint`: 0 errors

`develop` has since been merged in twice. The second merge conflicted with DEV-459 (#13359), which refactored this exact block: the `cell`-option snapshot and replay now live inside the `resetMetaCaches()` helper that refactor introduced, so both of its call sites — the new early pass and the guarded later one — carry the option across the clear. `tsc --noEmit` and the `setCellMeta` specs were re-run after that merge; CI covers the rest.

### Types of changes

### Related issue(s):

1. #5661
2. DEV-2774

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2774

https://claude.ai/code/session_018jufsuFHJPJ1AR7Lxg7D6d

---

> [!NOTE]
> **Medium Risk**
> Changes core meta lifecycle during `updateSettings` and `clearCache`, with subtle precedence and plugin-declarative edge cases; regression risk is mitigated by extensive tests but affects all grids that mix `cell`, `columns`, and imperative meta.
> 
> **Overview**
> Fixes **#5661**: declarative **`cell`** meta was wiped when **`updateSettings`** passed **`columns`** or **`cells`**, even though **`getSettings().cell`** still held the array.
> 
> **MetaManager** now tracks three write origins. Imperative **`setCellMeta`** and **`cell`**-option writes each get their own replay bucket; plugin writes via **`_setCellMetaDeclarative`** stay untracked so cache clears still drop them (e.g. ColumnSummary). On meta cache reset, **`getCellOptionCellMetas()`** is replayed at **physical** coordinates unless the call **restates** **`cell`** (including **`cell: []`**). Precedence is unchanged: restated **`cell`** wins; otherwise a later imperative override beats a preserved declaration.
> 
> **`updateSettings`** also treats **`cell: null`** as omitted (no throw), rejects non-array **`cell`** with a one-time warning without updating settings, and strips **`row`/`col`** before applying entries so they are not stored as meta.
> 
> Docs, changelog **5661**, and broad E2E/unit coverage document the behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0329c931a5fedd3dcf90646c7550aa9ad0345a7e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2774 Fix the `cell` option being erased by updateSettings (GH #5661)](https://app.clickup.com/t/9015210959/DEV-2774)

qunabu:
Task linked: [DEV-459 Error after update settings rendering twice and picking wrong renderer functions](https://app.clickup.com/t/9015210959/DEV-459)

demtario (commented):
3 inline notes below.

handsontable/src/core.ts
```diff
@@ -3974,16 +3979,49 @@ export default function Core(
...
 
+    // The `cell` option is restated when this call passes an array - including an empty one, which is how a
+    // caller removes every previously declared entry. A restatement replaces what the option declared
+    // before, so the earlier entries are neither replayed nor merged.
+    //
+    // The test is `Array.isArray`, not `isDefined`: `isDefined(null)` is true, so a `cell: null` used to
+    // clear the meta cache and then throw on `null.forEach`, leaving the instance with its cache wiped and
+    // the rest of the update - render, dimensions, hooks - never run. A non-array is now ignored, which
+    // leaves the option exactly as it was.
+    const isCellOptionRestated = Array.isArray(settings.cell);
```
demtario, line 4004: **medium**: `Array.isArray(settings.cell)` silently no-ops on *any* non-array `cell` value, not just the `null` case this line was meant to fix. A common typo like `cell: { row: 0, col: 0, readOnly: true }` (forgetting to wrap in an array) now silently does nothing instead of surfacing the mistake. The comment above only justifies the `null`-crash fix, but the guard is broader than that.

Fix: either throw/warn when `settings.cell` is defined, non-null, and not an array, or narrow the silent-ignore behavior (and its comment) to `null` specifically.

handsontable/src/__tests__/core/setCellMeta.spec.js
```diff
@@ -317,4 +334,276 @@ describe('Core.setCellMeta', () => {
...
+      });
+
+      await removeCellMeta(0, 0, 'readOnly');
+
+      await updateSettings({ columns: [{}, {}, {}, {}, {}] });
+
+      expect(getCellMeta(0, 0).readOnly).toBe(false);
+    });
+
+    it('should apply a restated `cell` option passed together with `columns` in one call', async() => {
```
demtario, line 468: **low**: No test covers a restated `cell` option overriding a *pre-existing imperative `setCellMeta` override* on the same key, even though this precedence is explicitly documented (core.ts JSDoc: "a value re-stated through `cell` takes precedence over the preserved value"; metaSchema.ts: "unless the call restates `cell`, which is applied last and wins over everything"). Verified the behavior is implemented correctly (`resetMetaCaches` replays user-defined first, then the `isCellOptionRestated` block runs last and overwrites) - this is a missing regression test, not a bug.

Suggested addition: init with `cell: [{row:0,col:0,readOnly:true}]` -> `setCellMeta(0,0,'readOnly',false)` -> `updateSettings({cell:[{row:0,col:0,readOnly:true}]})` -> assert `getCellMeta(0,0).readOnly === true`.

handsontable/src/__tests__/core/setCellMeta.spec.js
```diff
@@ -317,4 +334,276 @@ describe('Core.setCellMeta', () => {
       expect(afterCellMetaReset).toHaveBeenCalled();
     });
   });
+
+  describe('preserving the declarative `cell` option across updateSettings (#5661)', () => {
```
demtario, line 338: **low** (advisory, not a violation): all new coverage for this user-visible behavior (the `cell` option surviving `updateSettings`) landed in the frozen Jasmine `*.spec.js` suite. Allowed under the monorepo policy (this edits an existing spec file rather than adding a new one), but per `handsontable/.ai/TESTING.md`'s rule that new E2E for user-visible behavior should be Playwright, worth considering whether a `tests/e2e/*.spec.ts` case is also warranted here. Author/maintainer judgment call, not blocking.

handsontable/src/core.ts
budnix: Nice catch — a non-array now warns once per instance instead of passing in silence, in `1fb9a3e80f`. `null` stays quiet, since it reads as "not passed".

handsontable/src/__tests__/core/setCellMeta.spec.js
budnix: Good spot — added that exact case in `1fb9a3e80f`, plus one for the new warning.

handsontable/src/__tests__/core/setCellMeta.spec.js
budnix: Fair point — I kept these next to the existing #4446 block so the whole rule stays in one place, and they assert cell-meta state rather than browser interaction.

qunabu (dismissed):
LGTM but @demtario got some nice catches

handsontable/src/core.ts
```diff
@@ -3974,16 +3979,59 @@ export default function Core(
...
+    // the rest of the update - render, dimensions, hooks - never run. `null` reads as "not passed" and is
+    // ignored silently, the way the other nullable options are cleared.
+    const isCellOptionRestated = Array.isArray(settings.cell);
+
+    // Any other non-array is a mistake worth surfacing, most often a single entry that was not wrapped in
+    // an array. Ignoring it in silence would leave the previously declared entries in place and look like
+    // the call did nothing at all. Warned once per instance, because a wrapper that re-sends its settings
+    // on every render would otherwise repeat it on every commit.
+    if (settings.cell !== undefined && settings.cell !== null && !isCellOptionRestated) {
+      warnOnce(instance.rootElement, 'Core.invalidCellOption',
```
demtario, line 4013: **medium**: The warning says the passed `cell` value was ignored. For `getSettings()`, it is not.

`updateSettings` copies every own key onto `globalMeta` first (`globalMeta[i] = settings[i]` a few lines above). Constructor settings, including `cell`, already live on that same prototype (`MetaManager` → `globalMeta.updateMeta`). A typo like `cell: { row, col, readOnly }` therefore warns `"The passed value was ignored"`, still replaces `getSettings().cell` with the object, and still skips applying it as meta (the previous entries survive via replay). `null` is the same shape: treated as "not passed" for application, but still written onto `globalMeta`. The new spec asserts cell meta and the warning count, not `getSettings().cell`. The Typedoc line in `metaSchema.ts` repeats the same "ignored" claim.

Fix: do not assign `cell` onto `globalMeta` unless `Array.isArray(settings.cell)`. Optionally assert `getSettings().cell` is still the previous array after a malformed / `null` payload. Align the warning and Typedoc with that.

budnix: You're right, the warning was lying — `1fb9a3e80f` only stopped the value being applied. Fixed in `0329c931a5`: it is now dropped in the settings loop and in the constructor filter too, so `getSettings().cell` keeps the previous array.

demtario (approved):

PR #13362, opened 2026-09-03
DEV-2775: Batch the mergeCells init render (#5687)

### LINKED ISSUE #5687: Having merge cells populated causes 2 additional table rendering (both renderer and afterRender)
### Description
If mergeCells is populated two additional render cycles (renderer and afterRender) are exectued. 
Is this expected behavior or is there any workaround because in cases of a bigger table (even 20x20 with some styles applied) this has significant impact on performance. For a table of 20x20 that is almost additional 800 calls.

### Steps to reproduce

1. Create a table 
2. Populate mergeCells or set it to true

### Demo
https://jsfiddle.net/w7kx2epy/

### Your environment
* Handsontable version: 6.2.2
* Browser Name and version: Chrome Version 71
* Operating System: Ubuntu 16.04

budnix:
### Context

The PR fixes [#5687](https://github.com/handsontable/handsontable/issues/5687) — open since 2018 — where enabling `mergeCells` made the grid draw itself **two extra times** while it initialized.

Measured on released 18.1.0 and on `develop`, with a 10x10 grid and no headers so one full draw is exactly 100 `afterRenderer` calls:

| Configuration | Draws before | Draws after | Cell renders |
| --- | --- | --- | --- |
| no `mergeCells` | 1 | 1 | 100 → 100 |
| `mergeCells: true` (no areas) | 2 | **1** | 200 → 100 |
| `mergeCells: [2 areas]` | 3 | **2** | 300 → 200 |

Both extra draws came from the plugin's `#onAfterInit` handler, and stack traces taken inside `afterRender` account for all three:

1. `core.ts` — the grid's own first draw.
2. `generateFromSettings()` clears the cells each merge area covers through `setDataAtCell()`, which renders on its own.
3. The handler then called `this.hot.render()` again.

Comparing the master table's full `outerHTML` at the end of every draw confirmed one draw was always wasted: byte-identical with declared areas, and DOM-equal (`isEqualNode`, only attribute order differing) with `mergeCells: true`.

`#onAfterInit` now applies the merges between an explicit `suspendRender()` / `resumeRender()` pair, and skips the work entirely when no merge area is declared. The one remaining extra draw is unavoidable: `afterInit` only runs once the grid has already been drawn.

Three parts of the change look removable but are not, and all three are now recorded in `handsontable/src/plugins/mergeCells/AGENTS.md`:

- **The `cells` length check sits outside the suspend/resume pair.** `resumeRender()` always draws once the pair is entered, so not entering it is the only way to spend zero draws on a grid with nothing to merge.
- **The inner `this.hot.render()` stays.** `resumeRender()` draws through `TableView#render`, which picks fast-vs-full from `forceFullRender`, and only `Core#render` sets that flag. `generateFromSettings()` returns early without writing when every covered cell is already `null`, so without it the batched draw can be a *fast* draw that skips the cell renderers and the spans never appear. An **async validator** on any column also reverses the order of the two init draws — making that render the one which puts the merges on screen — so it could not simply be deleted either.
- **The suspend/resume is written by hand, not via `Core#batchRender`.** That helper is `suspendRender(); fn(); resumeRender();` with no `finally`, and the clearing write runs user code. A throw from a `beforeChange` handler or a validator would skip `resumeRender()` and leave `renderSuspendedCounter` above zero for the rest of the instance's life, making every later `render()` a silent no-op. Confirmed in a browser: `batchRender` + throw leaves `isRenderSuspended()` `true`, the explicit `try`/`finally` leaves it `false`. **`Core#batchRender` and `Core#batch` still have this shape** and are worth fixing separately — that is not done here.

#### Observable behavior change

`afterRender` and `afterRenderer` now fire fewer times during init on a grid with `mergeCells` enabled — once instead of three times with declared areas, once instead of twice for plain `mergeCells: true`. No API, option, or rendered output changes; the final DOM is identical (that is how the wasted draws were identified). The entry is filed `breaking: false` on the basis that render counts are not a public contract — flagging it explicitly so a reviewer can overrule that if the team disagrees.

Not in scope: `updatePlugin()` (the `updateSettings` path) is untouched, since `updateSettings()` renders at the end regardless. The guard counts *declared* areas rather than *valid* ones, so a config whose areas are all invalid still pays the one draw — no regression, and narrowing it would mean either running `validateSetting()` twice (double-warning) or restructuring `generateFromSettings()`. Applying the merges before the grid's *first* draw would remove the last extra draw too, but needs the work moved off `afterInit`.

### How has this been tested?

New Playwright spec, written before the fix and confirmed failing on the unfixed build (`2 ≠ 1` and `3 ≠ 2`):

```bash
cd tests && npx playwright test --project=e2e-main --project=e2e-horizon --project=e2e-classic e2e/merge-cells-init-renders.spec.ts
```

12 passed across the three themes. It covers four cases: no `mergeCells`, `mergeCells: true`, declared areas (counts plus the rendered `rowspan`/`colspan`), and the async-validator ordering — which waits for the deferred write's own draw before re-checking the spans, rather than sampling a value already asserted.

Full suites, re-run after the review fixes:

```bash
npm run test:e2e --prefix handsontable                                        # 9849 specs, 0 failures
npm run test:e2e --prefix handsontable -- --testPathPattern=mergeCells        # 523 specs, 0 failures
npm run test:unit --prefix handsontable -- --testPathPattern=mergeCells       # 66 passed
npm run test:e2e --prefix handsontable -- --testPathPattern=modifySinglePassLayout  # 9 passed
npm run test:types --prefix handsontable                                      # clean
npm run eslint --prefix handsontable                                          # 0 errors
```

### Types of changes

### Related issue(s):

1. DEV-2775
2. #5687

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2775

https://claude.ai/code/session_013WmHQiTcgajen48hx2Lfm8

---

> [!NOTE]
> **Medium Risk**
> Changes mergeCells init timing against the core render/suspend pipeline and user hooks (validators, beforeChange), though behavior is heavily regression-tested and final output is intended to stay the same.
> 
> **Overview**
> Fixes [#5687](https://github.com/handsontable/handsontable/issues/5687) by changing how the **mergeCells** plugin applies configured merges on init. **`#onAfterInit`** now runs **`generateFromSettings()`** only when declared **`cells`** exist, wraps that work in **`suspendRender()` / `resumeRender()`** with **`try`/`finally`**, and keeps an explicit **`hot.render()`** so merges still paint correctly (including with async validators). **`mergeCells: true`** with no areas skips the path entirely so init no longer triggers a redundant full draw.
> 
> **`afterRender` / `afterRenderer`** fire fewer times during construction; the final grid and API are unchanged. **`AGENTS.md`** documents the init-render constraints; **`.claude/skills/pr-creation/SKILL.md`** clarifies public vs private changelog filenames. New Playwright coverage (**`merge-cells-init-renders.spec.ts`**, fixture, page object) locks draw counts and merge DOM for several setting shapes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bf5c7f13f4dfbc14e62f8b5dca8b5e24a7c14a55. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changelogs/5687.json
+++ b/.changelogs/5687.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed the `mergeCells` plugin rendering the grid two extra times while it initialized.",
+  "type": "fixed",
+  "issueOrPR": 5687,
+  "breaking": false,
+  "framework": "none"
+}
--- a/.changelogs/5687.json
+++ b/.changelogs/5687.json
@@ -1,6 +1,6 @@
 {
   "issuesOrigin": "public",
-  "title": "Fixed the `mergeCells` plugin rendering the grid two extra times while it initialized.",
+  "title": "Fixed the `mergeCells` plugin drawing the grid more times than needed while it initialized.",
   "type": "fixed",
   "issueOrPR": 5687,
   "breaking": false,
--- a/handsontable/src/plugins/mergeCells/AGENTS.md
+++ b/handsontable/src/plugins/mergeCells/AGENTS.md
@@ -56,11 +56,12 @@ mirroring the old comparison against `undefined`.
 - **The `TR` `background` property is modified so it can be changed asynchronously later.** Only the alpha
   changes, so it is invisible — the TDs' own background covers it. Do not remove it as dead styling.
 
-## The init draw is batched, and two things about it are load-bearing
+## The init draw is batched, and three things about it are load-bearing
 
-`#onAfterInit` applies the declared merges inside `hot.batchRender()` (#5687). Before that it drew the
-grid twice — `generateFromSettings()` clears the cells each area covers through `setDataAtCell()`, which
-renders, and the handler then rendered again. Three rules come out of it, and each has a measured reason.
+`#onAfterInit` applies the declared merges between a `suspendRender()` / `resumeRender()` pair (#5687).
+Before that it drew the grid twice — `generateFromSettings()` clears the cells each area covers through
+`setDataAtCell()`, which renders, and the handler then rendered again. Three rules come out of it, and
+each has a measured reason.
 
 - **Keep the explicit `this.hot.render()` inside the batch.** `resumeRender()` draws through
   `TableView#render`, which picks fast-vs-full from `hot.forceFullRender`, and only `Core#render` sets that
@@ -73,7 +74,14 @@ renders, and the handler then rendered again. Three rules come out of it, and ea
   write would leave such a grid unmerged until validation resolves.
 - **Skip the work entirely when no area is declared.** `mergeCells: true` with no `cells` has nothing to
   apply, and the initial render already shows the final grid, so a draw there repaints an identical table.
-  `batchRender()` always draws on resume, so this has to be a check *around* it, not inside.
+  `resumeRender()` always draws once the pair is entered, so this has to be a check *around* it, not
+  inside.
+
+Do not collapse the pair back into `hot.batchRender()`. That helper is `suspendRender(); fn();
+resumeRender();` with **no `finally`** (`core.ts`), and the clearing write runs user code — a
+`beforeChange` handler, a sync validator. A throw there would skip `resumeRender()` and leave
+`renderSuspendedCounter` above zero for the rest of the instance's life, so every later `render()`
+silently does nothing. The explicit `try`/`finally` here is that guard.
 
 Coverage: `tests/e2e/merge-cells-init-renders.spec.ts` pins the counts and the async-validator ordering.
 `updatePlugin()` is a separate path and is deliberately untouched — `updateSettings()` renders at the end
--- a/handsontable/src/plugins/mergeCells/AGENTS.md
+++ b/handsontable/src/plugins/mergeCells/AGENTS.md
@@ -56,6 +56,29 @@ mirroring the old comparison against `undefined`.
 - **The `TR` `background` property is modified so it can be changed asynchronously later.** Only the alpha
   changes, so it is invisible — the TDs' own background covers it. Do not remove it as dead styling.
 
+## The init draw is batched, and two things about it are load-bearing
+
+`#onAfterInit` applies the declared merges inside `hot.batchRender()` (#5687). Before that it drew the
+grid twice — `generateFromSettings()` clears the cells each area covers through `setDataAtCell()`, which
+renders, and the handler then rendered again. Three rules come out of it, and each has a measured reason.
+
+- **Keep the explicit `this.hot.render()` inside the batch.** `resumeRender()` draws through
+  `TableView#render`, which picks fast-vs-full from `hot.forceFullRender`, and only `Core#render` sets that
+  flag. `generateFromSettings()` returns early without writing anything when every covered cell is already
+  `null`, so without the explicit call the batched draw can be a *fast* one — it skips the cell renderers,
+  and the spans never appear.
+- **Never gate that render on "the clearing write already rendered."** Put an async `validator` on any
+  column and the clearing write's own draw lands *after* `afterInit` returns, which **reverses** the order
+  of the two init draws: the handler's render becomes the one that puts the merges on screen. Gating on the
+  write would leave such a grid unmerged until validation resolves.
+- **Skip the work entirely when no area is declared.** `mergeCells: true` with no `cells` has nothing to
+  apply, and the initial render already shows the final grid, so a draw there repaints an identical table.
+  `batchRender()` always draws on resume, so this has to be a check *around* it, not inside.
+
+Coverage: `tests/e2e/merge-cells-init-renders.spec.ts` pins the counts and the async-validator ordering.
+`updatePlugin()` is a separate path and is deliberately untouched — `updateSettings()` renders at the end
+regardless.
+
 ## `cellCoords.ts` handles six structural cases
 
 Adding rows/columns, removing rows/columns, removing the whole merge, removing partially-including-the-start,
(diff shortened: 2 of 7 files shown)


qunabu:
Task linked: [DEV-2775 MergeCells renders the grid two extra times on init (#5687)](https://app.clickup.com/t/9015210959/DEV-2775)

qunabu (dismissed):

handsontable/src/plugins/mergeCells/AGENTS.md
```diff
@@ -56,6 +56,37 @@ mirroring the old comparison against `undefined`.
...
+`#onAfterInit` applies the declared merges between a `suspendRender()` / `resumeRender()` pair (#5687).
+Before that it drew the grid twice — `generateFromSettings()` clears the cells each area covers through
+`setDataAtCell()`, which renders, and the handler then rendered again. Three rules come out of it, and
+each has a measured reason.
+
+- **Keep the explicit `this.hot.render()` inside the batch.** `resumeRender()` draws through
+  `TableView#render`, which picks fast-vs-full from `hot.forceFullRender`, and only `Core#render` sets that
+  flag. `generateFromSettings()` returns early without writing anything when every covered cell is already
+  `null`, so without the explicit call the batched draw can be a *fast* one — it skips the cell renderers,
+  and the spans never appear.
```
demtario: **medium**: This bullet's justification doesn't hold for the call site it documents. `#onAfterInit` calls `generateFromSettings()` with no argument, so `alreadyAppliedMerges` is always an empty `Set`, `isReapplied` is always `false`, and the null-skip guard (`if (!isReapplied || this.#getStoredValueAt(r, c) !== null)`) always evaluates `true` here. The "returns early when every covered cell is already null" behavior belongs to `updatePlugin()`'s re-apply path, not to this init call. The actual load-bearing reason to keep the explicit `render()` is the one correctly described in the next bullet: an async validator defers the clearing write's own render past `afterInit`, so nothing else sets `forceFullRender` before `resumeRender()` draws.

Fix: reword this bullet to point at the async-validator deferral, or fold it into the next one.

handsontable/src/plugins/mergeCells/AGENTS.md
```diff
@@ -56,6 +56,37 @@ mirroring the old comparison against `undefined`.
 - **The `TR` `background` property is modified so it can be changed asynchronously later.** Only the alpha
   changes, so it is invisible — the TDs' own background covers it. Do not remove it as dead styling.
 
+## The init draw is batched, and three things about it are load-bearing
```
demtario: **nit**: Heading says "three things... load-bearing," but a fourth rule (don't collapse the pair back into `hot.batchRender()`) follows unbulleted at line 80. Either fold it into the bulleted list and update the count, or call it out separately.

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -778,10 +778,31 @@ export class MergeCells extends BasePlugin {
...
+      this.hot.suspendRender();
+
+      try {
+        this.generateFromSettings();
+        // Load-bearing: `resumeRender()` draws through the view, which picks fast-vs-full from
+        // `forceFullRender`, and only `Core#render` sets it. Without this the batched draw can be
+        // a fast one, which skips the cell renderers that apply the spans.
+        this.hot.render();
+      } finally {
+        // Suspend/resume by hand rather than through `Core#batchRender`, which has no `finally`:
```
demtario, line 799: **nit**: This comment frames the `try`/`finally` as guarding "the instance" generally, but it only restores `renderSuspendedCounter`. A throw in `generateFromSettings()` still propagates past this block, so `this.#initialized = true` and `#captureMergeAnchors()` never run either way (unchanged from before this PR, not a regression). Consider narrowing the wording to "restores the render-suspend counter."

.changelogs/5687.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed the `mergeCells` plugin rendering the grid two extra times while it initialized.",
```
demtario: **low**: The title reads as "the extra draws are gone," but for the common case (`cells` declared, no async validator) one extra draw is still emitted by design — down from two, not to zero. Consider: "Fixed the `mergeCells` plugin drawing the grid more times than needed while it initialized."

demtario (commented):
4 inline notes below.

Off-diff: the object-form `mergeCells: { cells: [...] }` setting resolves through a different `getSetting()` branch (`isPlainObject` vs `Array.isArray` in `base.ts`) than the array form the new fixture exercises. Verified harmless, but the new `cells.length > 0` guard's behavior on that shape is untested by this PR's E2E spec.

handsontable/src/plugins/mergeCells/AGENTS.md
budnix: You're right, that early return can't fire on the init path — reworded the bullet to the async-validator reason in `bf5c7f13f4`.

handsontable/src/plugins/mergeCells/AGENTS.md
budnix: Good spot — it's a fourth bullet now and the heading says four, in `bf5c7f13f4`.

handsontable/src/plugins/mergeCells/mergeCells.ts
budnix: Fair point — narrowed the wording in `bf5c7f13f4`, it now says only the counter is restored and the throw still propagates.

.changelogs/5687.json
budnix: Agreed, that read as if they were all gone. Used your wording in `bf5c7f13f4`.

demtario (approved):

Issue #5905, opened 2019-03-22
count error?


ByboyCn:
Hello, developer:
I found some problems with the count calculation:
First, I set the count column to a numeric type. The total number of displays is four. 
https://jsfiddle.net/8hc2boqk/4/
Then I set this as a letter or other character. He displays 0.
https://jsfiddle.net/8hc2boqk/3/
 but I read the explanation in the document. "Counts the Non-empty values ​​in the specified column and row range. "
My understanding is that any non-empty column can be counted, but the case of the letter can not be counted.
help look.
thank you ~

AMBudnik:
Hey @ByboyCn 

thank you for investigating this issue.

The `Average` case has been reported repeatedly [here](https://github.com/handsontable/handsontable/issues?utf8=%E2%9C%93&q=is%3Aopen+is%3Aissue+average) for many use cases.

When it comes to `Count` formula I guess that it works as expected (like Excel) but the description is wrong. The Office package docs https://support.office.com/en-us/article/count-function-a59cd7fc-b623-4d93-87a4-d23bf411294c tells us that `Count` should tell us how many cells with numbers are in the range. So if you have types non-numeric values it should not count them.

I will discuss this with our CTO.

budnix:
Hi @ByboyCn,

Thanks for the report, and apologies for the long wait.

I tested both of your examples on the current version. Here is what I found.

**Your original case works as intended.** The `count` summary counts cells that hold a number, so a column of letters returns 0. That part is deliberate. The documentation was wrong to promise "the number of all non-empty cells" — that sentence is what misled you, and I have corrected it. If you want to summarize a column of text that starts with a number, such as `3 kg`, use the `forceNumeric` option.

**While testing it, I found a real bug underneath.** A cell holding an empty string was read as the number zero. That matters because the two ways of clearing a cell store different things: the Delete key stores `null`, which was ignored correctly, but clearing a cell in the editor stores an empty string. Data loaded from a backend carries `""` the same way. So the result changed depending on how a cell became blank.

Measured on a column of `10`, `20`, `30` with one cell cleared:

| Summary | Cleared with Delete | Cleared in the editor | Correct |
|---|---|---|---|
| `min` | 10 | 0 | 10 |
| `count` | 3 | 4 | 3 |
| `average` | 20 | 15 | 20 |

`sum` was the only summary type that stayed correct, because adding zero changes nothing.

Both the fix and the documentation correction are in #13363, and they will ship in an upcoming release.

[...]

PR #13363, opened 2026-09-03
DEV-2776: Stop ColumnSummary reading an empty cell as zero

budnix:
### Context

The PR fixes the `columnSummary` plugin reading a cell that holds an empty string (`''`) as the number **zero**. A `null` cell is correctly treated as empty, so until now the summary result depended on how a cell became blank.

Two everyday routes put `''` into a grid:

- **Data loaded with empty strings.** JSON from a backend, or a CSV import, routinely carries `""` for a blank numeric field. The summary is wrong from the first render, with no user action at all.
- **A user clearing a cell.** The Delete key writes `null`, but opening the editor and deleting the contents writes `''`. Verified in a real browser on both a plain column and a `type: 'numeric'` column — the numeric editor does not normalize the empty string away.

A column of `10`, `20`, `30` with one cell cleared, measured on released 18.1.0 and on develop:

| Summary | Cleared with Delete key (`null`) | Cleared in the editor (`''`) | Correct |
|---|---|---|---|
| `min` | 10 | **0** | 10 |
| `max` | 30 | 30 | 30 |
| `sum` | 60 | 60 | 60 |
| `count` | 3 | **4** | 3 |
| `average` | 20 | **15** | 20 |

`max` looks safe only because the numbers are positive. On `-10`, `-20`, `-30` it returned **0** instead of `-10`. `sum` is the only genuinely safe type, because adding zero changes nothing.

**Cause.** The plugin's calculation gate (now named `holdsNoNumber()`, previously `isNullishOrNaN()`) decided emptiness with the global `isNaN()`, which converts its argument with `Number()` first. `Number('')` is `0`, not `NaN`, so an empty string was reported as a real value, and the two summing sites then ran `Number(rawValue)` and turned it into an actual zero. `' '` and `'  '` coerce the same way.

The plugin already got the empty string right on another path: with `forceNumeric: true` it runs `parseFloat('')`, which **is** `NaN`, so blanks were excluded there. The default path was the outlier on that point, so the fix follows precedent set inside the plugin rather than inventing a rule. (`forceNumeric` is not a general reference implementation - `parseFloat(true)` is `NaN`, so it skips booleans, which the default path deliberately counts. That split is documented in the plugin's `AGENTS.md`.)

**Booleans deliberately keep coercing.** A `checkbox` column stores `true`/`false`, and a `sum` summary over one is the established way to count ticked boxes, so `true` still counts as `1`. There is a test pinning that, and a note in the plugin's `AGENTS.md`, so it does not get "finished" later.

#### Behavior change, and why it is not breaking

One case has a visible before/after: a range in which **every** cell holds an empty string. `min`, `max` and `average` now return `Not enough data` instead of `0`, and `count` returns `0` instead of the number of blank cells. That is the correction, not a side effect.

It is classed as a bug fix rather than a breaking change on the same grounds as DEV-2691 / #13299, which changed `average` from `NaN` to `Not enough data` and shipped as a non-breaking fix: no documented behavior said a blank cell is zero, `forceNumeric` already disagreed with it, and `min` returning `0` for a column that contains no zero is not a contract anyone could have relied on deliberately.

#### Documentation

GitHub issue #5905 reports a second thing: `count` returns `0` for a column of letters, while the guide promised "the number of all non-empty cells". That half is **working as designed** — the plugin summarizes numbers, and `getCellValue()` already raises a "not in a numeric format" error for a letter when `suppressDataTypeErrors` is off. The guide wording was what was wrong, so the `count` and `average` rows now say "cells that hold a number", followed by a paragraph that spells out which cells are skipped, notes that checkbox cells still count, and points at `forceNumeric` for a cell that *starts* with a number, such as `3 kg`.

#### Also in this PR

`.claude/skills/pr-creation/SKILL.md` told me the changelog file must always be named after the PR number. That holds only for `issuesOrigin: private`. A `public` entry cites an issue number and is named after it, and this PR got that wrong before review caught it - the entry is now `.changelogs/5905.json`. The skill carries the two-row table from `.changelogs/README.md` so the next author does not repeat it.

### How has this been tested?

Every new test was confirmed **red before the fix and green after**. Reverting the one function alone turns 2 of the new unit tests and 8 of the 11 new E2E cases red; the other 3 E2E cases are regression guards that must stay green either way (`sum` unchanged, a cell holding `0` still counted, booleans still summed).

Two of the E2E cases drive the path the reporter actually takes - the empty string arriving through an edit and recalculated by `afterChange` - rather than seeding it in the initial data.

```bash
# Unit tests for the helper, including the boolean and zero guards
npm run test:unit --prefix handsontable -- --testPathPattern=columnSummary

# E2E for the plugin
npm run test:e2e --prefix handsontable -- --testPathPattern=columnSummary

# Full gates
```

Results: unit 4322/4322, `columnSummary` E2E 74/74, full E2E 9860/0 (8 pre-existing pending), types and ESLint clean (0 errors).

Also verified by hand in Chrome against both the released 18.1.0 build and a local develop build, driving the real editor (double-click, select all, Backspace, Enter) rather than writing values through the API.

### Types of changes

### Related issue(s):

1. DEV-2776
2. #5905

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2776

---

> [!NOTE]
> **Medium Risk**
> Changes numeric aggregation semantics for a widely used plugin; behavior shifts for empty-string cells but is covered by extensive tests and aligns with documented intent.
> 
> **Overview**
> Fixes **ColumnSummary** treating cells with `''` (or whitespace-only text) as numeric **zero**, which made `min`, `max`, `count`, and `average` disagree with blanks cleared via Delete (`null`).
> 
> The built-in paths now share a renamed gate, **`holdsNoNumber`**, that explicitly skips empty/whitespace strings before coercion. **`sum`**, **`min`/`max`**, **`count`**, and **`average`** all use it; literal **`0`** and checkbox **`true`/`false`** still participate. Ranges that are only empty strings now yield **`Not enough data`** (or **`count` → 0**) instead of **`0`**.
> 
> Docs clarify that **`count`** / **`average`** count cells that hold a number; tests cover edits via **`afterChange`**, **`forceNumeric`** vs default behavior, and a public changelog entry for #5905.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a01cc159f58bb16a884b6a7a070f07d93f9c26a3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2776 Fix ColumnSummary treating an empty string as the number zero](https://app.clickup.com/t/9015210959/DEV-2776)
Task linked: [DEV-2691 Fix `average` column summary showing NaN for an empty range (#4646)](https://app.clickup.com/t/9015210959/DEV-2691)

.claude/skills/pr-creation/SKILL.md
```diff
@@ -162,7 +162,14 @@ When asked to update, fix, or re-fill a PR description, use the same temp-file a
...
 
-Every PR that changes source code needs a changelog entry in `.changelogs/`. The filename **must** be `{PR-number}.json`, using the PR number returned by `gh pr create` in the previous step. See the `changelog-creation` skill for the JSON schema and title-writing rules.
+Every PR that changes source code needs a changelog entry in `.changelogs/`. The filename **must** match `issueOrPR`, and which number that is depends on `issuesOrigin`:
+
+| `issuesOrigin` | `issueOrPR` | Filename |
+|---|---|---|
+| `private` — the default, and almost always correct | the PR number from `gh pr create` | `{PR-number}.json` |
+| `public` — only when the entry cites a real public GitHub issue | the **issue** number | `{issue-number}.json` |
+
+The field picks the link path in the generated `CHANGELOG.md`, so `public` paired with a PR number publishes a link to `/issues/{PR-number}` — a wrong link that also hides the entry from the duplicate check. Work tracked only in a private ClickUp task is always `private`. See `.changelogs/README.md` for the authoritative table, and the `changelog-creation` skill for the JSON schema and title-writing rules.
```
demtario: **Medium**: This line claims a `public`/PR-number mismatch "hides the entry from the duplicate check," which overstates the effect. `findRepublishedEntries` in `bin/lib/published-entries.js` matches a republished entry by the number captured from the `[#N]` link text, not from the `pull`/`issues` URL segment, so the check still finds the entry regardless of `issuesOrigin`. What the mismatch actually does is drop the check's severity from `error` to `warning` (unless the title also matches, in which case it still escalates to `error`) — it never fully hides the entry.

Fix: reword to describe the real effect, e.g. "a wrong link that also weakens the republish-duplicate check from a hard error to a warning."

demtario (commented):
3 findings below (1 inline, 2 off-diff).

**Off-diff notes:**

- `handsontable/src/plugins/columnSummary/AGENTS.md` — the new section documents two specific claims (`forceNumeric: true` throws on an empty cell when `suppressDataTypeErrors` is `false`; `forceNumeric: true` sums a checkbox column to `0`) that are accurate but untested — no case in `columnSummary.spec.js` combines `forceNumeric: true` with an empty string or a boolean. Not a regression (the code path is untouched), but worth a follow-up test now that the trap is called out by name.
- `docs/content/guides/columns/column-summary/column-summary.md` — the page content changed substantively (two new paragraphs, two rewritten table rows) but the frontmatter has no `menuTag: updated`. Minor, easy to add in a follow-up or amend.

qunabu (dismissed):
LGTM but resolve @demtario issues please

.claude/skills/pr-creation/SKILL.md
budnix: You're right — reworded in `708b59e722`: the check still finds the entry, it just drops from error to warning.

demtario (dismissed):

demtario (approved):

PR #13347, opened 2026-09-02
DEV-2768: Teach skills and agent docs the determinism rules the flake work earned

danielzytohoc:
### Context

The PR adds the determinism rules the CI-flakiness investigation earned to the skills and agent docs that dispatch test work, so the next migration inherits them instead of re-learning them run by run. Skills, agent docs, `AGENTS.md`, and `.ai/` files, plus one `node --test` check over the review dispatch surfaces; no product code, no changelog entry (tooling/test-infrastructure change).

The text describes the tree after the three sibling lint PRs land, and they merge first: #13349 (the global `setTimeout` at error in specs and page objects, frame waits flagged in the frozen tier), #13350 (the diff-scoped ratchet), and #13364 (every `waitForFunction()` passes `{ polling }`, and `tests/.eslintrc.cjs` errors on one without it). This PR merges develop before landing. Where it overlaps #13364 in `tests/AGENTS.md` (the fixture-contract `goto()` bullet), the wording is byte-identical to that branch.

Each rule lands where the dispatch already points readers, with the one-line WHY drawn from the measured incident:

- **Playwright mechanics** (`handsontable-playwright-e2e` + `references/determinism.md`, `tests/AGENTS.md`): the six page-object rules lint cannot see — lint reads source text, so a timer in a fixture's inline script or a string-form `evaluate`, and the state a wait ends on, stay out of its reach. A `setTimeout` that runs in the browser is a fixed wait wherever it sits; `page.waitForFunction()` passes an explicit `{ polling }` (the rAF default starved under worker load: 3 mute timeouts in ~700 runs of `hidden-init-rerender.spec.ts`, 0 after timer polling — every page object passes it since #13364); a scroll/mutate method ends on a render-state probe, never `scrollTop` (reference `OverlaysPage.scrollToEnd()`, which ends on the last cell being rendered; `FrozenTallCellPage.scrollVerticallyTo()` is the counter-example behind the `frozen-column-row-heights` gap); a multi-delivery trigger is asserted by polling the latest entry of its kind; a fixture build fails loud through a captured constructor throw; a negative assertion needs a positive control. `references/determinism.md` carries the full rules and the incident behind each; `tests/AGENTS.md` carries a one-sentence summary per rule and points there, so there is one copy of the prose to keep current. Its fixture-contract bullet spells out `{ polling: 100 }`.
[...]

PR #13364, opened 2026-09-03
DEV-2769: Give every waitForFunction() an explicit polling interval, and lint for it

danielzytohoc:
### Context

The PR fixes a latent mute-timeout flake in the Playwright tiers and adds the lint that keeps it fixed.

`page.waitForFunction()` polls on `requestAnimationFrame` by default. Under parallel-worker load a healthy page can have its rAF callbacks starved past the test timeout, so the wait times out with nothing wrong on the page. Measured twice now: 3 mute `goto()` timeouts in ~700 runs of `hidden-init-rerender.spec.ts` (0 after that page object moved to `{ polling: 100 }`), and today two `editor-trimmed-row.spec.ts` tests timing out on `waitForFunction(() => 'Handsontable' in window)` in a 56-test local run while the machine was also building — both green on the re-run, which is the signature of this class, not proof of health.

Twelve `waitForFunction()` calls in `tests/` still ran on the default when the branch was cut, and eleven more arrived on `develop` while the PR was open; all twenty-three are swept here, and the review round extended the sweep to the thirteen calls in `visual-tests/` and `performance-tests/`. The skill rule that says "an explicit polling interval, always" sat beside eleven counterexamples.

Changes:

- **Helper** (review round): the bundle wait every page object's `goto()` opens with now lives in one place, `tests/fixtures/bundle.ts` — `awaitBundle(page)` with the interval in `BUNDLE_POLLING_MS`. Twenty-one page objects each kept their own copy of the same line, which is why the lint could stop one from forgetting the option but nothing kept the value from drifting. Every identical wait in `tests/fixtures/pages/**` is now `await awaitBundle(this.page)`; the per-file comments that only restated the helper's two reasons (why `waitForFunction` rather than `expect`, why a timer interval rather than rAF) are gone, the file-specific ones stay, and `ManualResizePage`'s private wrapper is replaced by the shared function. It is a separate module rather than `gestures.ts` because that file is pointer gestures and a bundle wait is not one. The waits on different conditions are untouched: `HiddenInitRerenderPage`'s `htBuildError`-aware wait, `IncrementalRenderPage`'s `htReady` wait, `MergeCellsInitRendersPage`'s and `MergeCellsNestedRowsCollapsePage`'s fixture-state waits, and the two overlay-scrollbar fade waits.
- **Sweep**: every `waitForFunction()` in `tests/fixtures/pages/**` and `tests/e2e/**` passes an explicit polling interval — through the helper for the twenty-one bundle waits (ten from the branch point, plus `InvalidMarkUpdateSettingsPage`, `ManualResizePage`, `SortingFilteredSpareRowsPage`, `UpdateSettingsColumnShrinkPage`, `EditorPreventCloseElementPage`, `IncrementalRenderPage`, `walkontable/ExactRowHeightsPage`, `HidingArrayClassNamePage`, `NumericArrayClassNamePage`, and `walkontable/InlineStartBorderRefreshPage`, which reached `develop` afterwards and were caught by this PR's own lint on each `develop` merge), and inline for `IncrementalRenderPage`'s `htReady` wait and the two overlay-scrollbar fade waits (`walkontable/OverlaysPage`, `overlay-scrollbar-clearance.spec.ts`), which add `polling: 100` to their existing `{ timeout: 5000 }`. `HiddenInitRerenderPage` already had it.
- **Other tiers** (review round): `visual-tests/src/test-runner.ts`'s scrollbar-clearance settle — the same condition and the same `timeout: 5000` as `OverlaysPage` — and all twelve `performance-tests/` waits (the pre-trace `__hot` / `__build` waits in nine scenario specs, `cell-editing`'s `resetFn` wait, and the two index-mapper waits in `lib/scroll-utils.mjs`) pass `{ polling: 100 }`, keeping their existing timeouts. None of the performance waits sits inside a measured span: `runTracedScenario` opens the window after `MEASURE_START_MARK` and closes it after `settle()`, while the swept waits run before the first trace, in `setupFn`, or in `resetFn` after `stopTracing()` — so no number moves. The performance README and the `performance-testing` skill show the polled form.
- **Lint** (`no-restricted-syntax` at `error`, three selectors for the three shapes of the miss, now in all three Playwright tiers): a `waitForFunction()` call with no options argument; an options object literal without a `polling` property (`polling`, `'polling'`, and `['polling']` keys all count); and that literal wrapped in a type assertion (`as`, `satisfies`, `<T>`). Only a plain options **variable** is not judged, matching how the frozen-tier rule treats what it cannot see; a spread (`{ ...opts }`, `...args`) is flagged, because the rule asks the call site to state the interval. `polling: 'raf'` passes: the rule asks for an explicit choice at the call site, and reviewers see it. The message names the rAF-starvation reason and points at `tests/AGENTS.md`. `tests/.eslintrc.cjs` is the source; `visual-tests/.eslintrc.js` and `performance-tests/.eslintrc.cjs` carry a verbatim copy of the block, spread over the airbnb `no-restricted-syntax` list their `.ts` files already get (a rule setting replaces the inherited one, it does not merge) and over the tier's own list for the performance `.mjs` helpers, where the TypeScript-only third selector is inert. The selectors are quoted single-outside, double-inside in all three files, because `visual-tests` lints its own `.eslintrc.js` under the root `quotes` rule; esquery reads both the same.
- **Docs**: `tests/AGENTS.md` shows `await awaitBundle(this.page)` as the form to copy in the `goto()` paragraph, keeps the two reasons, and says why not to inline a copy; the Determinism list carries the ban with the reason and the judged/not-judged boundary.

Merge order: land this PR before #13347 (skills and agent docs). #13347's review round replaces its two "known debt" sentences about unpolled page objects — the Determinism bullet in `tests/AGENTS.md` and item 2 of `handsontable-playwright-e2e/references/determinism.md` — with the post-sweep statement that cites this PR, so it must merge after the sweep it describes. The `goto()` paragraph of `tests/AGENTS.md` is no longer byte-identical to the one #13347 lands (it now names the helper), so #13347 takes this PR's version on its rebase. Nothing in the repo runs `waitForFunction()` on the rAF default any more: `tests/`, `visual-tests/`, and `performance-tests/` are all swept and all linted for it, and all three lint scripts run in CI through the root `npm run lint` in `Lint / core` (`npm run all lint` reaches `visual-tests`, and the root script calls `npm run lint --prefix performance-tests` and `--prefix tests` by name).

- **CI fallback fix** (review round, CI-only): this PR's shape — `tests/`, `visual-tests/`, and `performance-tests/` changed, `handsontable/` untouched — runs `Build / UMD` but scope-skips `Build / ES + CJS`, so `visual.yml` composes `tmp/` itself. That fallback built the type declarations but not the language packs, while the UMD artifact carries `dist/` and `styles/` only, so `postbuild:partial` failed on the `./languages/all` require target (`languages/all.js`) and both visual jobs went red on a PR that touches no grid code. The fallback now runs `build:languages` and `build:languages.min` like `build.yml`'s ES+CJS job; `.github/scripts/__tests__/partial-packaging.test.mjs` asserts it per shell block (it fails against the previous `visual.yml`, since the UMD fallback in the same job already named `build:languages` and would have masked the gap), and the root `AGENTS.md` gotcha records the trigger.

### How has this been tested?

- **Rule red on the pre-sweep tree, green after.** With the rule added and the tree unswept, `npx eslint --ext .ts e2e fixtures` (from `tests/`) reported exactly the 12 sites (11 page objects + `overlay-scrollbar-clearance.spec.ts:283`). After the sweep: 0 problems. Each later `develop` merge turned the rule red again on exactly the page objects added there in the meantime (four, then four more), which is the rule doing its job; swept each time, 0 problems again.
- **Selector semantics probe** (a throwaway file linted, then deleted), 16 shapes. Flagged: no options; `undefined` only; `{ timeout: 5000 }`; the same literal wrapped in `as Opts`, `satisfies Opts`, and `<Opts>`; `{ ...opts }`; a `page.mainFrame().waitForFunction()` receiver; a multi-line call whose literal lacks `polling`. Not flagged: `{ polling: 100 }`; `{ 'polling': 100 }`; `{ ['polling']: 100 }`; `{ polling: 100 } as Opts`; `{ timeout, polling: 'raf' }`; an options variable, bare or cast; a call whose *second* argument is an object literal while the third carries `polling`; `page.waitForSelector(x, { timeout })`.
- **Review round, rule active in every tier.** A throwaway probe in each tier (no options; `{ timeout: 5000 }`; the same literal `as { timeout: number }`; and a polled call) run through the tier's own lint script: `npm run lint` in `visual-tests` flagged the three unpolled shapes and passed the polled one; `npm run lint` in `performance-tests` did the same for a `.ts` probe under `scenarios/` and flagged both plain-JS shapes in a `.mjs` probe under `lib/`; `npm run lint` in `tests` still flags all three after the quoting swap. Probes deleted; with the real tree each of the three lint scripts reports 0 problems, and `npm run typecheck` in `performance-tests` passes.
- **Adversarial review** (three independent lenses, each finding re-verified by a refuter): 14 raw findings, 7 confirmed, all nits, all addressed in the second commit — the wrapped-literal bypass, the string-key false positive, the "spread is not judged" wording, the message pointing at a skill file that does not mention polling on `develop`, the `tests/AGENTS.md` paragraph colliding with #13347, and the Determinism list omitting the new ban.
- **CI seam checked**: `Lint / core` runs the root `npm run lint`, which fans out to `npm run lint --prefix tests`, `npm run lint --prefix performance-tests`, and (through `npm run all lint`) the `visual-tests` lint, so the rule blocks in CI for all three tiers; `scripts/lint-files.mjs` scopes include `tests/(e2e|fixtures)/`, so pre-push and lint-staged reach the functional tier too.
- **The 24 specs that use a swept page object** run green on `e2e-main` locally against the worktree's own build, 336 tests (`autocomplete-async-source`, `autocomplete-stale-highlight`, `cell-dropdown-arrow-button`, `editor-hidden-cell`, `editor-prevent-close-element`, `editor-trimmed-row`, `header-markup-gate`, `incremental-render`, `invalid-mark-update-settings`, `layout-slot-focus`, `manual-resize-contextmenu`, `manual-resize-size-option-reset`, `manual-resize-teardown`, `merge-cells-init-renders`, `merge-cells-nested-rows-collapse`, `min-spare-rows-nested-schema`, `refresh-dimensions`, `selection-trim-shapes`, `selection-trimmed-row`, `sorting-filtered-spare-rows`, `source-data-validator-address`, `trusted-types`, `update-settings-column-shrink`, `walkontable/exact-row-heights`). For the record, the first attempt was 332/336: three `refresh-dimensions` iframe tests failed because the worktree had no `styles/handsontable.min.css` built yet, and one `autocomplete-async-source` `beforeEach` hit the 20 s test timeout inside `awaitBundle()` on that same cold first-start server — a slow bundle load on a timer poll, not the rAF class this PR removes. With the stylesheet built, 336/336. The pre-push hook re-ran `overlay-scrollbar-clearance.spec.ts` (5 passed).

```bash
cd tests && npx eslint --ext .ts e2e fixtures
cd visual-tests && npm run lint
cd performance-tests && npm run lint && npm run typecheck
```

`[skip changelog]` — test-tier and lint-config change only; no shippable source touched.

### Types of changes

### Related issue(s):

1. DEV-2769

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2769

---

> [!NOTE]
> **Low Risk**
> Test-tier, lint, and CI packaging changes only; no core grid or wrapper runtime behavior.
> 
> **Overview**
> Fixes **mute Playwright timeouts** caused by `waitForFunction()` defaulting to `requestAnimationFrame`, which parallel workers can starve on an otherwise healthy page.
> 
> **Determinism sweep:** Every `waitForFunction()` in `tests/`, `visual-tests/`, and `performance-tests/` now passes an explicit `{ polling: 100 }` (or adds `polling` beside existing `timeout`). Functional E2E centralizes the bundle gate in **`tests/fixtures/bundle.ts`** (`awaitBundle(page)`, `BUNDLE_POLLING_MS`); page objects call that instead of duplicating the wait. Scroll helpers, performance scenario specs, visual scrollbar settle, and a few inline waits (`htReady`, overlay fade) are updated the same way.
> 
> **Lint:** `no-restricted-syntax` in all three Playwright tiers flags missing or unpolluted options (mirrored from `tests/.eslintrc.cjs`). Docs/skills (`tests/AGENTS.md`, performance README, performance-testing skill) describe the rule.
> 
> **CI:** `visual.yml`'s ES+CJS **artifact fallback** now runs `build:languages` / `build:languages.min` so `postbuild:partial` can satisfy `./languages/all` when only UMD build runs; **`partial-packaging.test.mjs`** asserts `build:languages` per shell block that runs `build:es`. Root **`AGENTS.md`** records that gap.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d13bbf35545a889d35f1db1f37718dcffc5d8f69. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:

tests/fixtures/pages/ManualResizePage.ts
```diff
@@ -160,7 +160,7 @@ export abstract class ManualResizePage {
    * pulls its own copy, which outlasts the 10s `expect` timeout.
    */
   async awaitBundle(): Promise<void> {
-    await this.page.waitForFunction(() => 'Handsontable' in window);
+    await this.page.waitForFunction(() => 'Handsontable' in window, undefined, { polling: 100 });
```
budnix: This class already wraps the bundle wait in `awaitBundle()`, but the other fifteen page objects each keep their own copy of the same line. That is why this PR touches sixteen files, and why a later commit had to sweep four more that landed on develop meanwhile. Would it be worth lifting this into `tests/fixtures/gestures.ts`, next to `dragResizeHandle`, so the 100 ms interval lives in one spot? The new rule stops a page object from forgetting the option, but it cannot keep the value from drifting apart.

tests/.eslintrc.cjs
```diff
@@ -51,6 +56,25 @@ module.exports = {
         selector: "CallExpression[callee.property.name='fixme']",
         message: 'test.fixme() parks a known product bug and is allowed ONLY with an eslint-disable line naming the tracking task (`// eslint-disable-next-line no-restricted-syntax -- DEV-1234: <why>`), so the exception stays counted and attributable. See tests/AGENTS.md.',
       },
+      {
+        // Three shapes of the same miss: no options argument, an options literal without `polling`, and
+        // that literal wrapped in a type assertion (`as`, `satisfies`, `<T>`). Only a plain options
+        // VARIABLE is not judged; a spread is flagged too, because the rule asks the call site to be
+        // explicit about the interval.
+        selector: "CallExpression[callee.property.name='waitForFunction'][arguments.length<3]",
```
budnix: This config only covers `tests/`, so the ban stops at this package. `visual-tests/` and `performance-tests/` also run Playwright and have thirteen `waitForFunction` calls between them with no polling option, and both already have their own eslint config to copy this into. The clearest one is `visual-tests/src/test-runner.ts`, which waits on the same scrollbar clearance condition with the same `timeout: 5000` you just fixed in `OverlaysPage`. Was leaving those for a follow-up deliberate?

tests/fixtures/pages/ManualResizePage.ts
danielzytohoc: Done in 91207f957 and 084d72cee: the wait now lives in tests/fixtures/bundle.ts as awaitBundle(page) with the interval in BUNDLE_POLLING_MS (bundle.ts:7 and :28), and all twenty-one page objects that carried the inline copy call `await awaitBundle(this.page)` instead, including ManualResizePage's private wrapper (removed) and the three page objects develop added meanwhile. It is a separate module rather than gestures.ts because that file is pointer gestures and a bundle wait is not one; `git grep "'Handsontable' in window"` now hits only bundle.ts:28, and the waits on other conditions (HiddenInitRerenderPage, IncrementalRenderPage's htReady, the two scrollbar-fade waits) keep their own { polling: 100 }. tests/AGENTS.md:39-48 shows the helper as the form to copy, `npx eslint --ext .ts e2e fixtures` reports 0 problems, and the 27 specs that use a swept page object pass on e2e-main (378/378).

tests/.eslintrc.cjs
danielzytohoc: Not deliberate — fixed in 2e939bad3. visual-tests/.eslintrc.js:7-27 and performance-tests/.eslintrc.cjs:7-27 carry the same three selectors and message as tests/.eslintrc.cjs, spread over the airbnb no-restricted-syntax list the *.ts overrides inherit and, for the performance .mjs helpers, over the tier's own list (performance-tests/.eslintrc.cjs:89-96). All thirteen calls now pass { polling: 100 }: visual-tests/src/test-runner.ts:40 (the scrollbar-clearance settle, timeout: 5000 kept), the nine pre-trace waits in performance-tests/scenarios/*/*.spec.ts, cell-editing's resetFn wait (cell-editing.spec.ts:48), and the two index-mapper waits in lib/scroll-utils.mjs:19 and :37 — none sits inside runTracedScenario's marked window (they run before the start mark or after stopTracing), so no measured number moves. `npm run lint` in each of the three tiers reports 0 problems, a throwaway unpolled probe in each tier is flagged (3 in tests, 3 in visual-tests, 5 in performance-tests across a .ts and a .mjs probe), and both `Lint / core` (root npm run lint reaches all three tiers) and `Lint / visual tests` are green on the head.

budnix (approved):

Issue #7553, opened 2019-05-30
setState on afterValidate hooks


pnowak:
### Description
We have many questions about using `setState` in hooks `afterValidate` - users want to change ` state` after validation. We need investigations why it doesn't work. And fix it if it's a mistake. If it's not a mistake, it's enough to create a working demo.

### Steps to reproduce

1. https://github.com/handsontable/support/issues/116#issuecomment-497224262

### Your environment
* React wrapper version:
* Handsontable version:
* Browser Name and version:
* Operating System:

swistach:
I close this issue due to we put that problem into a scope of #8372.

warpech:
Reopening, because it is not linked to #8372 as we assumed.

Rather, as @jansiegel explains in https://github.com/handsontable/handsontable/issues/8372#issuecomment-1048665298, it:

> Seems to be a problem with the way `afterValidate` hook is triggered - using the (undocumented) `postAfterValidate` hook seems to resole the problem.

rmlevangelio:
Any update on this one? Basically I would also love to know how can I set a state to the parent component after a change has been done

adrianszymanski89:
Hi @rmlevangelio 

This issue isn't on our roadmap yet. As for the second question. Can you please describe your requirements more or share your current implementation?

budnix:
Hi @simkimsia, @rmlevangelio, and everyone following this thread — thanks for your patience on this one.

I looked into it and reproduced it. Here is what I found.

The hook was never the problem. Calling `setState` inside `afterValidate` is fine. The highlight disappeared because of `updateSettings()`: when it receives `cell`, `cells`, or `columns`, it clears the cell metadata cache, and the failed-validation flag was not restored afterwards. The cell kept the invalid value — only the red mark went away.

That is why React users ran into it so often. A grid with `HotColumn` children re-sends `columns` on every render, so any re-render dropped the marks. Calling `setState` from `afterValidate` simply causes a re-render at the exact moment a cell has turned invalid, which is why it looked like a hook problem. Angular is affected too, on every change to the `settings` input. Cells still being checked by an asynchronous validator were affected as well.

Two notes for the record:

- The cause originally reported here was fixed in 12.0.0, which is why some setups stopped seeing it. A second cause remained, and that is what stayed reproducible.
- `postAfterValidate` is not a workaround for this. It only appears to help in examples that have no `cells` prop and no `HotColumn` children — plain `afterValidate` works in those too.

[...]

PR #13365, opened 2026-09-03
DEV-2778: Preserve the invalid cell mark across updateSettings

budnix:
### Context

The PR fixes a bug where a cell marked as invalid silently loses its highlight while keeping the invalid value. It closes GitHub issue #7553, open since 2019.

`updateSettings` clears the whole cell-meta cache when its payload carries `cell`, `cells` or `columns` (`core.ts`). Cell meta written through `setCellMeta` is snapshotted and replayed around that clear — the #4446 fix. The failed-validation flag is not, because the validation flow writes `valid` straight onto the meta object rather than through `setCellMeta`, so it is never recorded as user-defined meta.

The smoking gun: in one and the same `updateSettings({ cells })` call, a `setCellMeta`-written `className` survives and the invalid mark does not.

**Why it reached users as an `afterValidate` problem.** `createNewGlobalSettings` assigns the `HotColumn`-derived `columns` *after* the skip-if-unchanged gate (`hotTableInner.tsx:127`), so a React grid with `HotColumn` children re-sends `columns` on every render. Calling `setState` inside `afterValidate` is simply the most common way to trigger a render at the moment a cell has just gone invalid. The hook itself is not involved — a control with no `afterValidate` listener, re-rendered from a button long after validation finished, loses the mark just the same, and memoizing does not help.

All three wrappers can trigger it, on different cadences. React re-sends `columns` on **every** render of a grid with `HotColumn` children, with no gate at all. Angular re-sends on every `settings` input change. Vue only re-sends when the column settings actually changed — it compares by reference and order first.

Two things the issue thread concluded are worth correcting for the record:

- `postAfterValidate` is not a workaround. It passes only in demos that have no `cells` prop and no `HotColumn` child — so does plain `afterValidate`.
- The originally reported mechanism (`data` → `loadData` → `clearCellsCache`) was already fixed in 12.0.0, when `updateSettings` switched to `updateData`. Bisected empirically: 7.0.0 lost the mark, 11.1.0 lost it, 12.0.0 kept it. That fix never reached the `HotColumn` setup, which is why the issue stayed reproducible.

### The change

A second snapshot around the same cache clear: `CellMeta#getInvalidMetas()` and `restoreInvalidMetas()`, exposed through `MetaManager` as `getInvalidCellMetas()` / `restoreInvalidCellMetas()`.

Three decisions worth a reviewer's attention:

1. **The restore is a direct property write, never `setCellMeta`.** Routing it through `setCellMeta` would record `valid` as a user-defined property, and every later cache clear would then replay a stale `false` onto a cell the user has since corrected. There is a test that pins this.
2. **Only `valid === false` is preserved, never `valid === true`.** This matches the rule `evictRow()` already applies, for the reason its comment gives: a passing result has no rendered state, and every core reader branches on `!== false`, so `undefined` and `true` are indistinguishable.
3. **All three trigger keys preserve it, `columns` included.** A changed `columns` can change the validator, which is an argument for dropping the result there. It is preserved anyway, because `columns` is exactly the key the wrappers re-send with nothing having changed. A cell whose validator genuinely changed is re-validated on the next edit or on `validateCells()`, as before.

Two related paths are covered as well:

- **A validator still in flight when the cache is cleared.** `validateChanges` hands `validateCell` the stored meta *by reference*, so the clear detaches it and a write on it reaches nothing. Every result now goes through one `persistValidationResult()` helper, in **both** directions. Three symptoms, one cause: a failure that never lands leaves the cell unmarked; a pass that never lands leaves behind the `false` that `restoreInvalidCellMetas` re-applied, so a corrected value keeps its red mark; and the `allowInvalid: false` cancel path writes `valid = true` on that same detached reference, so a *rejected* edit — where the cell keeps its old, valid value — stays flagged. The helper peeks with `getCellMetaIfExists()` rather than resolving, so a passing result never materializes meta on its own and `_validateCells` keeps its O(invalid cells) bound; only a failure materializes, because only a failure has to be rendered. Physical coordinates come off the meta object rather than being re-derived from the visual pair, which a row insert or move in the same window would shift.
- **A `valid` flag inherited from a higher meta layer.** `valid` is an ordinary meta key, so a grid can declare `valid: false` on the column or global layer. The snapshot reads own properties only; reading through the prototype chain reported every materialized cell of such a grid as invalid and stamped the flag on permanently.

`evictRow()` and `getInvalidMetas()` share one `isFlaggedInvalid()` predicate, so the two keep rules cannot drift apart.

`getInvalidMetas()` walks the materialized cell metas once. `getUserDefinedCellMetas()` already makes the identical walk on the line above, and the loop body is a single own-property `=== false` check. The read path (`getCellMeta`) is untouched — no new branch there.

### How has this been tested?

```bash
npm run test:e2e --prefix handsontable                                  # 9860 specs, 0 failures
npm run test:unit --prefix handsontable                                 # 4402 tests, 344 suites, 0 failures
npm run test --prefix wrappers/react-wrapper                            # 99 tests, 16 suites, 0 failures
npm run eslint --prefix handsontable                                    # 0 errors
npm run build --prefix handsontable                                     # clean
```

Targeted runs while developing:

```bash
npm run test:e2e --prefix handsontable -- --testPathPattern=setCellMeta
npm run test:unit --prefix handsontable -- --testPathPattern=metaManager
```

Tests added:

- `handsontable/src/dataMap/metaManager/__tests__/metaManager.unit.ts` — the passthroughs, `valid === false` only, own-property-only (a flag inherited from the column layer is ignored, an own flag is reported even when the column says otherwise), physical rather than visual coordinates, survival across a clear, and the assertion that the restored `valid` lands in neither `_userDefinedMetaProps` nor `_persistedMetaProps`.
- `handsontable/src/__tests__/core/setCellMeta.spec.js` — a `#7553` describe block beside the existing `#4446` one: `cells`, `cell`, `columns`, no other cell affected, repeated calls, the shifted position after a row insert, the out-of-range column checked against the `setCellMeta` baseline, an async validator failing after the clear, an async *correction* resolving after the clear, a rejected `allowInvalid: false` edit across the clear, an inherited `valid` flag, and the correction path.

**Red runs.** Each fix was disabled in turn and only its own tests failed: without the snapshot/restore, 7 of the `#7553` cases; without the write-through, exactly the async failure case; restricting the write-through to failures only, exactly the async-correction and rejected-edit cases; without the own-property guard, exactly the inherited-flag case and its unit counterpart. The over-fix guard (a corrected cell must not get its mark back) passes either way, by design.

**Manual.** A React harness with 10 scenarios, driven both through the API and by real keyboard typing, run against the local build and released 18.1.0, with both a synchronous and an asynchronous validator. On 18.1.0, 5 of the 10 lose the mark in both modes — including `HotColumn` with and without the hook. On this branch all 10 keep it in both modes, and correcting the value clears the mark without it returning on the next render.

**A note on one test run.** A full E2E pass reported 21 failures once; re-running the identical tree gave 9858/0, and the validation specs were green in both. It was environment flakiness in the legacy suite, not a regression — recorded here so the number is not a surprise if it reappears.

### Review follow-ups

A code review raised twelve points; nine are addressed above or in the tests. A later automated pass raised three more on the async write-through, two of which were real regressions in this branch and are fixed above (the third — coordinates shifting under a concurrent row move — is handled by reading the physical pair off the meta object).

Three points from the first review are deliberately declined:

- **Generalizing the two guards into a `_directMetaProps` set.** It would collapse the snapshot pair and the eviction keep condition into one mechanism, which is the right shape, but it is a write-path change to the meta layer for a single flag. That deserves its own task and sign-off rather than riding along here. The `AGENTS.md` note now points at the idea so the next person finds the door. (`_persistedMetaProps` cannot be reused: declarative `cell`-option writes live there and must not be replayed.)
- **Folding the two snapshot walks into one pass.** The second walk is one own-property comparison per materialized meta, on a path that already does a full plugin re-init and re-render. Merging them means changing `getUserDefinedCellMetas()`'s return shape for a gain that does not show up next to that cost.
- **The `waitForNextAnimationFrames(7)` barrier in the specs.** It is the established idiom in `validator.spec.js` and its neighbours. Matching the surrounding suite beats introducing a novel barrier in a frozen legacy suite. The one genuinely timing-sensitive new test avoids the question entirely — its validator hands the callback to the test, so the ordering is exact rather than merely likely.

### Out of scope

`hotTableInner.tsx:127` still bypasses the deep-compare gate for the `HotColumn`-derived `columns`, so every render of such a grid rebuilds the whole meta cache. That is a performance issue rather than a correctness one now, and it deserves its own task. Fixing it alone would not have closed this hole anyway: `areEquivalentSettingsValue` treats two different function references as changed, so a `columns` prop carrying an inline `validator` defeats the gate regardless.

### Types of changes

### Related issue(s):

1. DEV-2778
2. #7553

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2778

---

> [!NOTE]
> **Medium Risk**
> Changes validation persistence and `updateSettings` meta-cache replay in core paths used by all wrappers; behavior is narrow and heavily tested, but mistakes could mis-flag cells or leak stale invalid state.
> 
> **Overview**
> Fixes **#7553**, where cells could keep an invalid value but lose the `htInvalid` styling after `updateSettings()` with `cell`, `cells`, or `columns` (common on React `HotColumn` re-renders and Angular `settings` updates). Failed validation sets `valid` with a **direct meta write**, so it was not covered by the existing user-defined / `cell`-option snapshots around `clearCache()`.
> 
> `updateSettings` now **snapshots and restores** cells with own-property `valid === false` via `getInvalidCellMetas()` / `restoreInvalidCellMetas()` (direct restore, not `setCellMeta`, so a later fix is not replayed as still invalid). `CellMeta` shares an **`isFlaggedInvalid()`** predicate with viewport eviction so scroll and cache-clear behavior stay aligned.
> 
> **Async validation** is handled with `persistValidationResult()`: results are written onto the **current** stored meta (`getCellMetaIfExists`), including passes and the `allowInvalid: false` cancel path, when the validator’s meta object was detached by a cache clear. **`getStructureVersion()`** drops late results after row/column structure changes or `loadData`/`clearCellsCache`, but not after `updateSettings`’s `clearCache()`.
> 
> Adds unit coverage, Playwright e2e for the three trigger keys and async edge cases, changelog **7553**, and meta-layer docs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 12735d4ede66fcc81fa529e94c824ef7fd864f53. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2778 Invalid cell mark is lost after updateSettings with cell/cells/columns (GH #7553)](https://app.clickup.com/t/9015210959/DEV-2778)

demtario (commented):
2 inline notes below.

handsontable/src/__tests__/core/setCellMeta.spec.js
```diff
@@ -317,4 +317,243 @@ describe('Core.setCellMeta', () => {
       expect(afterCellMetaReset).toHaveBeenCalled();
     });
   });
+
+  describe('preserving failed validation results across updateSettings (#7553)', () => {
```
demtario: **high** New E2E coverage was added to the frozen Jasmine/Puppeteer suite (`describe` at this line, several cases).

Core `AGENTS.md` requires new E2E in Playwright (`tests/e2e/*.spec.ts`) and allows only routine edits to existing Jasmine specs. The Jest tests in `metaManager.unit.ts` already cover the metadata-layer restore; the user-visible `#7553` regression belongs in Playwright.

Move those cases to `tests/e2e/*.spec.ts` and drop the new Jasmine `describe`.

handsontable/src/core.ts
```diff
@@ -2515,6 +2518,58 @@ export default function Core(
...
+   * re-derived from the visual ones, which a row insert or move in the same window would shift.
+   *
+   * @param {object} cellProperties The cell meta object the validator was handed.
+   * @param {boolean} valid The validation result to persist.
+   */
+  function persistValidationResult(
+    cellProperties: Record<string, unknown> & { row?: number, col?: number, visualRow?: number, visualCol?: number },
+    valid: boolean
+  ) {
+    const physicalRow = cellProperties.row;
```
demtario, line 2568: **medium** `persistValidationResult` looks up the stored meta by `cellProperties.row` / `.col`. Those fields are stamped when the meta is read and are not updated when `LazyFactoryMap` shifts keys on insert/remove. After a cache clear, the write-through can land on a different cell (or materialize `valid: false` there) while the captured object has already moved.

Direct assignment on the captured object follows the shift; this helper does not. Physical coordinates on the object are not more stable than visual ones in that window.

Resolve the current map key by object identity (or rebase/invalidate the target on structural change), discard the result if the original cell is gone, and add a regression that runs a validator across an insert/remove.

handsontable/src/__tests__/core/setCellMeta.spec.js
budnix: You're right, that belonged in Playwright — moved to `tests/e2e/invalid-mark-update-settings.spec.ts` in baa8f320bd (with its own fixture and page object) and dropped the Jasmine block.

handsontable/src/core.ts
budnix: Good catch — the stamped `row`/`col` really do go stale, and I reproduced it marking the wrong cell. Fixed in baa8f320bd with a `getStructureVersion()` check that drops the write when rows or columns moved.

demtario (dismissed):
Conflicts

demtario (approved):

PR #13366, opened 2026-09-03
DEV-443: Count spare rows instead of assuming them when sorting a filtered grid

### LINKED ISSUE #5983: Sorting after filtering with minSpareRows set
### Description
There seems to be an issue with sorting after filtering if minSpareRows is true or set to a number.

For example, if the setting is minSpareRows: 1, then when you filter, then sort, the sort seems to be off by one rows. If its minSpareRows: 3, then the sort is off by 3 row.

### Steps to reproduce

1. Filter using the button I've added to the bottom of the table. (uses the Filtering API, it also breaks if you manually filter anything)
2. Sort descending the "A" column, and you can see how the sorting is off by however many extra blank rows you want added. See below in the image how the A column should be descending, but the bottom three rows are out of order (minSpareRows: 1).

### Demo
https://jsfiddle.net/3b7qw456/

### environment
* Handsontable version: 7.0.2
* Browser Name and version: Chrome, Version 73.0.3683.103 (Official Build) (64-bit)
* Operating System: Windows 8

demtario:
### Context

The PR fixes sorting a filtered grid that has `minSpareRows` set. With `columnSorting` (or `multiColumnSorting`), `filters` and `minSpareRows: N`, applying a filter and then sorting left the last N rows unsorted at the bottom of the grid — "the sort is off by however many extra blank rows you want added", as the report puts it.

`ColumnSorting#getNumberOfRowsToSort()` subtracted `minSpareRows` from the visible row count to keep the trailing spare rows out of the sortable range. That assumed the spare rows are on screen. They are not once a filter is applied: a spare row is empty, so an ordinary condition trims it away like any other non-matching row, and nothing re-creates it while the filter is on (`grid.adjustRowsAndCols()` runs after alters, data changes and `updateSettings`, never on `trimmedIndexesChanged`). The subtraction then excluded that many **real data rows**, which `sortByPresetSortStates()` re-appended below the sorted block in their original order.

The fix counts the trailing empty rows actually present instead of trusting the option, capping both the result and the walk at `minSpareRows`. The count can never exceed what was subtracted before, so behavior with the spare rows present is unchanged; the walk is capped rather than going through `countEmptyRows(true)`, whose scan is unbounded and reads every column of each row it visits — `adjustRowsAndCols()` caps its own `minSpareCols` count the same way, for the same reason.

`MultiColumnSorting` extends `ColumnSorting` and does not override the method, so both plugins were affected and both are covered.

One related overlap is **deliberately left alone**: a spare row sits inside the band `fixedRowsBottom` reserves, so with `minSpareRows: 1, fixedRowsBottom: 1` both terms describe the same row and two rows leave the sortable range. That is pre-existing, untouched by this change, and the existing `fixedRowsTop`/`fixedRowsBottom`/`minSpareRows` spec passes because of it. It is now recorded in the plugin's `AGENTS.md` so whoever corrects it knows that expectation has to be revisited first.

### How has this been tested?

New Playwright coverage in `tests/e2e/sorting-filtered-spare-rows.spec.ts` — 5 cases run against both sorting plugins (10 tests): the reported filter-then-sort flow, the same with `minSpareRows: 3`, a present spare row staying out of the sort, trailing empty rows beyond `minSpareRows` taking part in it, and the spare row returning below the data once the filter is cleared.

The two complement cases use `sortEmptyCells: true` with a single ascending click on purpose. Under the default (`false`) the compare function pushes empty values to the end in both sort orders, so a spare row lands last whether or not the guard excluded it — assertions written that way pass with the guard deleted and prove nothing. With the option on, an empty value sorts first ascending, so the spare row staying last is evidence of the guard. Verified by patching the guard out of the built bundle: those cases fail without it and pass with it.

```bash

# new coverage, all four local legs
cd tests && npx playwright test sorting-filtered-spare-rows \
  --project=e2e-main --project=e2e-main-min --project=e2e-horizon --project=e2e-classic

# full functional leg

# the frozen Jasmine suites for the touched plugins
npm run test:e2e --prefix handsontable -- --testPathPattern='columnSorting'
npm run test:e2e --prefix handsontable -- --testPathPattern='multiColumnSorting'
npm run test:e2e --prefix handsontable -- --testPathPattern='filters'

```

Results: new spec 40 passed across the four legs; full `e2e-main` 837 passed / 1 skipped; Jasmine columnSorting 274, multiColumnSorting 137, filters 284, all 0 failures; unit 4374 passed; types, ESLint and Stylelint clean.

### Types of changes

### Related issue(s):

1. #5983
2. DEV-443

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes sortable row-range logic used on every sort for columnSorting and multiColumnSorting; behavior is well covered by new e2e tests but touches core grid interaction with filters and spare rows.
> 
> **Overview**
> Fixes **#5983**: on a **filtered** grid with `minSpareRows`, column/multi-column sorting left the last N visible data rows unsorted at the bottom.
> 
> **`getNumberOfRowsToSort()`** no longer subtracts the `minSpareRows` setting from the row count. It now walks from the bottom and counts trailing empty rows with `isEmptyRow()`, stopping at `minSpareRows`, because filters remove empty spare rows from the view and they are not recreated until the filter clears. Only rows actually present as spare are excluded from the sort; extra trailing empty rows beyond the cap still participate in sorting.
> 
> Adds Playwright regression coverage (both sorting plugins), a demo fixture and page object, a changelog entry, and **AGENTS.md** notes on the counting rule and an existing `minSpareRows` + `fixedRowsBottom` overlap left unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 21558c1ef2290962a7fe91f212ea7d9b1d73e847. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/handsontable/src/plugins/columnSorting/AGENTS.md
+++ b/handsontable/src/plugins/columnSorting/AGENTS.md
@@ -73,6 +73,30 @@ same place.
 The sorted rows are arrays of the form `[rowIndex, ...values]`, so the only sorted column's value sits at
 index **1**.
 
+## Spare rows are counted, not assumed (#5983)
+
+`getNumberOfRowsToSort()` keeps the trailing spare rows out of the sortable range, and it decides
+how many there are with `hot.countEmptyRows(true)` capped at `minSpareRows`, never from the option
+alone. `minSpareRows` says how many trailing empty rows the Core tops the grid up to, not how many
+are on screen: a filter trims an empty spare row exactly like any other non-matching row, and
+nothing re-creates it while the filter is on (`adjustRowsAndCols()` runs after alters, data changes
+and `updateSettings`, never on `trimmedIndexesChanged`). Subtracting the option from a filtered
+row count therefore excluded the last N **real data rows** from the sort and re-appended them
+unsorted at the bottom. The cap matters in the other direction too: trailing empty rows beyond
+`minSpareRows` are ordinary data and stay in the sort.
+
+`MultiColumnSorting` does not override the method, so both plugins share this and any change to it.
+
+**Known overlap, deliberately left alone.** The count and `fixedRowsBottom` are subtracted
+independently, but a spare row is appended at the end of the data, so it sits *inside* the band
+`fixedRowsBottom` already reserves. With `minSpareRows: 1, fixedRowsBottom: 1` and no filter both
+terms describe the same row, two rows leave the sortable range, and the last real data row is left
+unsorted, which is the #5983 symptom reached through a different option. The fix above did not
+change that (the counted value is never larger than `minSpareRows`, so the subtraction is exactly
+what it always was), and `__tests__/columnSorting.spec.js`'s "should respect `fixedRowsTop`,
+`fixedRowsBottom`, and `minSpareRows` together" currently passes *because* of the double
+subtraction - anyone correcting the overlap has to revisit that expectation first.
+
 ## The sort indicator reserves room on the header *container*
 
 `has-sort-indicator` goes on the header container, not the label. Padding on the label would enlarge the
--- a/handsontable/src/plugins/columnSorting/AGENTS.md
+++ b/handsontable/src/plugins/columnSorting/AGENTS.md
@@ -76,8 +76,10 @@ index **1**.
 ## Spare rows are counted, not assumed (#5983)
 
 `getNumberOfRowsToSort()` keeps the trailing spare rows out of the sortable range, and it decides
-how many there are with `hot.countEmptyRows(true)` capped at `minSpareRows`, never from the option
-alone. `minSpareRows` says how many trailing empty rows the Core tops the grid up to, not how many
+how many there are by counting them with `isEmptyRow()`, capped at `minSpareRows`, never from the
+option alone. The walk stops at the cap on purpose, so do not swap it for `countEmptyRows(true)`:
+that one is uncapped, reads every column of every row it visits, and this runs on every sort.
+`minSpareRows` says how many trailing empty rows the Core tops the grid up to, not how many
 are on screen: a filter trims an empty spare row exactly like any other non-matching row, and
 nothing re-creates it while the filter is on (`adjustRowsAndCols()` runs after alters, data changes
 and `updateSettings`, never on `trimmedIndexesChanged`). Subtracting the option from a filtered
@@ -92,10 +94,17 @@ independently, but a spare row is appended at the end of the data, so it sits *i
 `fixedRowsBottom` already reserves. With `minSpareRows: 1, fixedRowsBottom: 1` and no filter both
 terms describe the same row, two rows leave the sortable range, and the last real data row is left
 unsorted, which is the #5983 symptom reached through a different option. The fix above did not
-change that (the counted value is never larger than `minSpareRows`, so the subtraction is exactly
-what it always was), and `__tests__/columnSorting.spec.js`'s "should respect `fixedRowsTop`,
-`fixedRowsBottom`, and `minSpareRows` together" currently passes *because* of the double
-subtraction - anyone correcting the overlap has to revisit that expectation first.
+change that: the counted value is never larger than `minSpareRows`, so the subtraction is exactly
+what it always was.
+
+Nothing in either sorting suite covers the overlap. `__tests__/columnSorting.spec.js`'s "should
+respect `fixedRowsTop`, `fixedRowsBottom`, and `minSpareRows` together" is the only test in
+`columnSorting/__tests__` or `multiColumnSorting/__tests__` that sets both options, and it is
+blind to it - its `Total` row holds the largest value in the sorted column, so it lands on the
+same row whether or not it took part in the sort, and all four assertions hold either way
+(verified by removing the spare-row term from a built bundle: identical output). Whoever fixes
+the overlap has to give that spec a `Total` value that is not the extreme one before it can say
+anything.
 
 ## The sort indicator reserves room on the header *container*
 
--- a/handsontable/src/plugins/columnSorting/columnSorting.ts
+++ b/handsontable/src/plugins/columnSorting/columnSorting.ts
@@ -673,7 +673,31 @@ export class ColumnSorting extends BasePlugin {
       return Math.max(0, (settings.maxRows ?? 0) - fixedRowsBottom);
     }
 
-    return Math.max(0, numberOfRows - (settings.minSpareRows ?? 0) - fixedRowsBottom);
+    const minSpareRows = settings.minSpareRows ?? 0;
+    // The spare rows are COUNTED, never assumed from the option. `minSpareRows` says how many
+    // trailing empty rows the Core tops the grid up to, not how many are on screen right now:
+    // a filter trims an empty spare row away like any other non-matching row, and nothing
+    // re-creates it while the filter is on. Subtracting the option there pinned the last N real
+    // data rows below the sortable range and left them unsorted (#5983). Counting also caps the
+    // other direction - trailing empty rows beyond `minSpareRows` are ordinary data and stay in
+    // the sort.
+    //
+    // The count stops at `minSpareRows` rather than going through `countEmptyRows(true)`, whose
+    // walk is uncapped: verifying that a row is empty reads every column of it, and this runs on
+    // every sort, so a grid with a large block of trailing empty rows (`minRows`, or a dataset
+    // that simply ends blank) would pay for rows the cap then discards. `adjustRowsAndCols()`
+    // caps its `minSpareCols` count the same way, for the same reason.
+    let spareRows = 0;
+
+    for (let row = numberOfRows - 1; row >= 0 && spareRows < minSpareRows; row--) {
+      if (!this.hot.isEmptyRow(row)) {
+        break;
+      }
+
+      spareRows += 1;
+    }
+
+    return Math.max(0, numberOfRows - spareRows - fixedRowsBottom);
   }
 
   /**
--- a/tests/e2e/sorting-filtered-spare-rows.spec.ts
+++ b/tests/e2e/sorting-filtered-spare-rows.spec.ts
@@ -1,0 +1,136 @@
+import { test, expect } from '../fixtures/test';
+import { SortingFilteredSpareRowsPage, type SortingPlugin } from '../fixtures/pages/SortingFilteredSpareRowsPage';
+
+/**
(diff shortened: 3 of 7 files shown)


qunabu:
Task linked: [DEV-443 Sorting after filtering with minSpareRows set](https://app.clickup.com/t/9015210959/DEV-443)

handsontable/src/plugins/columnSorting/AGENTS.md
```diff
@@ -73,6 +73,30 @@ same place.
 The sorted rows are arrays of the form `[rowIndex, ...values]`, so the only sorted column's value sits at
 index **1**.
 
+## Spare rows are counted, not assumed (#5983)
+
+`getNumberOfRowsToSort()` keeps the trailing spare rows out of the sortable range, and it decides
+how many there are with `hot.countEmptyRows(true)` capped at `minSpareRows`, never from the option
```
budnix: This line says the count goes through `hot.countEmptyRows(true)`, but the code never calls it. The comment in `columnSorting.ts` says the loop stops at `minSpareRows` on purpose, because that walk is uncapped. A future reader could trust this file and put the uncapped call back, which is the exact thing the code comment set out to avoid.
```suggestion
how many there are by counting them with `isEmptyRow()`, capped at `minSpareRows`, never from the option
```

handsontable/src/plugins/columnSorting/AGENTS.md
```diff
@@ -73,6 +73,30 @@ same place.
...
+`MultiColumnSorting` does not override the method, so both plugins share this and any change to it.
+
+**Known overlap, deliberately left alone.** The count and `fixedRowsBottom` are subtracted
+independently, but a spare row is appended at the end of the data, so it sits *inside* the band
+`fixedRowsBottom` already reserves. With `minSpareRows: 1, fixedRowsBottom: 1` and no filter both
+terms describe the same row, two rows leave the sortable range, and the last real data row is left
+unsorted, which is the #5983 symptom reached through a different option. The fix above did not
+change that (the counted value is never larger than `minSpareRows`, so the subtraction is exactly
+what it always was), and `__tests__/columnSorting.spec.js`'s "should respect `fixedRowsTop`,
+`fixedRowsBottom`, and `minSpareRows` together" currently passes *because* of the double
```
budnix: I worked through that spec and it looks like it passes either way. It has 7 rows with `fixedRowsTop: 1`, `fixedRowsBottom: 1` and `minSpareRows: 1`, so a single subtraction gives a bound of 6 and `Total` (111) joins the sort. But 111 is the largest value, so it still lands on row 5 and all four assertions hold. That matters because whoever fixes the overlap will look up the guard named here, find it green, and think the overlap is already handled.

handsontable/src/plugins/columnSorting/columnSorting.ts
```diff
@@ -673,7 +673,31 @@ export class ColumnSorting extends BasePlugin {
...
+    // data rows below the sortable range and left them unsorted (#5983). Counting also caps the
+    // other direction - trailing empty rows beyond `minSpareRows` are ordinary data and stay in
+    // the sort.
+    //
+    // The count stops at `minSpareRows` rather than going through `countEmptyRows(true)`, whose
+    // walk is uncapped: verifying that a row is empty reads every column of it, and this runs on
+    // every sort, so a grid with a large block of trailing empty rows (`minRows`, or a dataset
+    // that simply ends blank) would pay for rows the cap then discards. `adjustRowsAndCols()`
+    // caps its `minSpareCols` count the same way, for the same reason.
+    let spareRows = 0;
```
budnix, line 690: The comment above says this loop avoids `countEmptyRows(true)` because that walk is uncapped, which makes sense for a path that runs on every sort. Core's own `minSpareRows` branch in `adjustRowsAndCols()` still calls the uncapped version after every change batch, which is a hotter path than sorting. Curious whether a capped `countEmptyRows(ending, limit)` in Core was on the table, so both places share one walk instead of keeping three variants around.

tests/e2e/sorting-filtered-spare-rows.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+// `multiColumnSorting` extends `ColumnSorting` and inherits the range calculation untouched,
+// so every case has to hold for both.
+for (const plugin of SORTING_PLUGINS) {
+  test.describe(`sorting a filtered grid with minSpareRows (${plugin})`, () => {
+    let grid: SortingFilteredSpareRowsPage;
+
+    test.beforeEach(async({ page, theme, bundle }) => {
+      grid = new SortingFilteredSpareRowsPage(page, theme, bundle);
+      await grid.goto();
+      await grid.useSortingPlugin(plugin);
```
budnix: The fixture builds a grid on load, this line destroys it and builds a second, and four of the five tests then build a third. Only the first test uses the grid this line makes. Times two plugins and six theme/bundle legs, that is a lot of grid construction nothing reads. Letting the page object do the first build instead of the fixture would drop one of the three.

tests/fixtures/pages/SortingFilteredSpareRowsPage.ts
```diff
@@ -0,0 +1,139 @@
...
+    return this.header(col).locator('span.colHeader');
+  }
+
+  /**
+   * Click a column header's sorting label, the gesture the bug report describes, and wait for
+   * the indicator so the follow-up assertions run against a sorted grid.
+   */
+  async sortByHeader(col: number): Promise<void> {
+    await this.sortLabel(col).click();
+    await expect(this.sortLabel(col)).toHaveClass(/ascending|descending/);
```
budnix: After the first sort the label already carries `ascending`, so this regex is true the moment the second click is dispatched and the wait returns without checking anything. Tests 1, 2 and 5 all click twice in a row, so the second click is never really awaited here. It works today only because `expect.poll` on the values retries a few lines later. Passing the expected order in and asserting that would make a lost click fail where it happens, instead of as a timeout further down.

tests/fixtures/demo/sorting-filtered-spare-rows.html
```diff
@@ -0,0 +1,90 @@
...
+
+      window.hot = new Handsontable(container, Object.assign({
+        // 13 rows, so a plain-text sort of column A puts A13 first descending - the value
+        // the report names as stuck at the bottom.
+        data: Handsontable.helper.createSpreadsheetData(13, 5),
+        colHeaders: true,
+        rowHeaders: true,
+        columnSorting: true,
+        filters: true,
+        dropdownMenu: true,
```
budnix: Is `dropdownMenu` needed here? The page object drives filtering through the plugin API, and `Filters` guards its dropdown wiring, so the menu looks unused. It does add a `.changeType` button right after the `span.colHeader` that `sortByHeader` clicks, and that button carries a negative margin, so it is one more thing sitting near the click target on a 50px column. Worth dropping if you agree it is not doing anything.

handsontable/src/plugins/columnSorting/AGENTS.md
demtario: Fixed, applied your wording. I also added the inverse instruction to the section, so the file itself says not to swap the loop for `countEmptyRows(true)`, rather than leaving that reasoning only in the source comment.

handsontable/src/plugins/columnSorting/AGENTS.md
demtario: You are right, and I checked it rather than taking it on the arithmetic: I removed the spare-row term from a built bundle and ran that spec's exact setup, and the output is identical with and without it. `Total` holds the largest value in the sorted column, so it lands on row 5 either way and all four assertions hold.

The paragraph now says the spec is blind to the overlap, and that fixing it starts with giving `Total` a value that is not the extreme one. I also grepped both sorting suites: it is the only test that sets `fixedRowsBottom` and `minSpareRows` together, so nothing else covers it either, and the text says that too.

handsontable/src/plugins/columnSorting/columnSorting.ts
demtario: Core has already answered this for columns: the `minSpareCols` branch of `adjustRowsAndCols()` is an inline capped loop whose comment gives the same reasoning (an uncapped count "paid O(empty columns * rows) per edit"), and the `countEmptyCols(true)` call it replaced is still there as public API, unchanged. This fix follows that precedent instead of adding a `limit` parameter, mainly because a public `countEmptyRows(ending, limit)` returns a number that is not the count, which is a poor public contract to take on for an internal walk.

That leaves the third variant you are pointing at, and it is the one worth closing: the `minSpareRows` branch of `adjustRowsAndCols()` still calls `countEmptyRows(true)` uncapped, on a hotter path than sorting, and it can take the same inline treatment its `minSpareCols` sibling already has, with no public API change. I would rather do that as its own change with its own coverage than fold a core hot-path edit into a sorting fix. Happy to open it if you agree.

tests/e2e/sorting-filtered-spare-rows.spec.ts
demtario: Reduced: the plugin selection moved out of `beforeEach`, so every test now builds exactly once instead of the four that built on top of it.

I kept the fixture's own build on load. Every `fixtures/demo/*.html` with an `init*Grid` helper auto-builds, it keeps the page browsable by hand, and `goto()`'s `expect(cell(0, 0)).toBeVisible()` wait needs a grid to exist. So it is two builds per test rather than one, and dropping the third needed no new URL parameter.

tests/fixtures/pages/SortingFilteredSpareRowsPage.ts
demtario: Fixed. `sortByHeader(col, expectedOrder)` now takes the order and asserts that exact indicator, and the argument is required so the no-op cannot come back by omission.

Verified it bites: flipping one call to the wrong order makes the test fail inside `sortByHeader` at the click, instead of surfacing as a value mismatch further down.

tests/fixtures/demo/sorting-filtered-spare-rows.html
demtario: Agreed, dropped. Filtering goes through the plugin API, so the menu was doing nothing, and a comment in its place now records why it is absent: the `.changeType` button sat right next to the `span.colHeader` the sort clicks.

budnix (approved):

PR #13341, opened 2026-09-02
DEV-2759: Document what render() covers, costs, and when to call it

budnix:
### Context

The PR closes the documentation gap that GitHub issue [#8462](https://github.com/handsontable/handsontable/issues/8462) has tracked since 2016.

An audit against 18.1.0 found the issue roughly half solved. DEV-457 (#12882) had already added the `setCellMeta()` to `render()` cross-reference, which answered "when must I call it myself?", and the batch operations guide replaced the reporter's `setTimeout` workaround. What was still missing is the reporter's main question — what `render()` actually does:

- **What a render covers was never documented.** The JSDoc said "all cell renderers are recalled", which reads as the whole data set. In fact `hot.render()` sets `forceFullRender`, `TableView.render()` turns that into `_wt.draw(false)`, and the draw cycle redraws only the rendered band: the viewport plus the `viewportRowRenderingOffset` / `viewportColumnRenderingOffset` buffer.
- **The cost of a render was stated nowhere.**
- **The `render()` JSDoc contradicted itself.** "Calling this method manually is not recommended" sat two sentences before "call this method to apply them".
- **Why hand-applied classes disappear was never explained.** `walkontable/src/render/cells.ts:93` clears the cell's `class` attribute and removes its inline `style`, `dir`, and all `aria-*` / `role` attributes before each renderer runs, so only what a renderer writes back survives.
- **No guide-level explanation of rendering existed.** A maintainer proposed a "short tutorial on `render()`" in 2020; it was never written.
- **Nothing compared** `render()`, `updateSettings()`, and `refreshDimensions()`.

The confusion keeps arriving through support rather than being answered by a page: issue #11171 (2024) and forum threads 4623 and 5459 are the same question.

**What this PR adds**

1. A new guide, `docs/content/guides/optimization/rendering/rendering.md` ("Understanding rendering"), registered first in the Optimization sidebar. It covers what a render does, what it covers, what it costs, when Handsontable renders on its own, when you call `render()` yourself, why direct DOM changes disappear, how to observe renders via `beforeRender` / `afterRender` and their `isForced` flag, that no per-cell or per-row render exists, and a comparison table of the three repaint methods.
2. A rewritten `render()` JSDoc, plus cross-links on `refreshDimensions()`, `updateSettings()`, and `setCellMeta()`. The API page is generated from this JSDoc, so the API reference is the file that had to change.
3. A "Changes made outside a renderer do not survive" section in the cell renderer guide.
[...]

Issue #5769, opened 2019-02-06
EcoRenderers feature


budnix:
### Description

This issue is about implementing a new feature internally called as `EcoRenderers`. The idea of this is to increasing Handsontable performance by re-rendering cells only when their value has changed. I assume that after implementing this feature issues related to scroll performance or refresh table lag after cell edit should disappear (or at least less annoying).

This feature is currently in the early development stage (researching and prototyping). Any important changes or news will be updated in this post.

aaronbeall:
This is of great interest to me!

Also please keep in mind cell dependencies. I really like how MobX and similar libraries [automatically track dependencies by access](https://mobx.js.org/best/react.html), it would be amazing if calling `getSourceData()` and such from a renderer would track the dependency, so that for example a triggered re-render of one cell would also re-render any renderers that had called `getSourceData()` on that cell. We make heavy use of renderers that have computed rendering (diffs, totals, conditional formatting).

tsohr:
+1. holymoly, Can't wait for this.

AMBudnik:
Inform [`forum`](https://forum.handsontable.com/t/handsontable-causing-performance-issue/2983)

rayokao:
I have noticed that Handsontable internally **reads** plenty from a DOM-property called _offsetWidth_ (and _offsetHeight_). Oddly, **reading** this property forces browser to reflow. This issue is known to cause rendering issues, jerky scrolling for example. The issue is explained [in this Sitepoint article](https://www.sitepoint.com/10-ways-minimize-reflows-improve-performance/). According to my own profiling efforts I would say that the way the grid dimensions are calculated is the issue, not the renderers.

tkane2000:
Hi @budnix!  

We're having major problems with scroll performance, so thanks for addressing the issue!

It sounds like the ecoRenderer solution is for tables in which the viewportColumnRenderingOffset and viewportRowRenderingOffset are set to the full length of columns and rows(?).  We have large tables (100 x 50,000) so we're setting both to auto and table cells will be rendered as you scroll.

Am I correct that this won't help in that scenario or am I missing something?

Thanks!

CC @AMBudnik

budnix:
[...]

PR #13367, opened 2026-09-03
DEV-2779: Add a recipe for repainting a single cell (#9614)

### LINKED ISSUE #9614: renderer cell only once
### Description

Hello author, I need to ask you a question. I just need to modify the value of a cell. Is there any way to trigger the renderer method of a cell only? Or which API can be completed?
thanks!

### Your environment
Handsontable version: 6.2.2
Browser name and version: laster
Operating system: mac m1

budnix:
### Context

The PR adds a documentation recipe answering GitHub issue #9614 — *"renderer cell only once"* — open since 2022. The reporter asked for a way to re-run one cell's renderer instead of a full `render()`. It was deferred to #5769 (EcoRenderers), and that issue has since been **closed and moved to the community forum**, so the deferral no longer points anywhere.

**The saving is real.** On a 6-column grid with an expensive renderer, one `setDataAtCell()` runs the renderer for every cell on screen — 66 calls, ~16 ms — while a per-cell repaint runs it once, ~0.8 ms. The cost is driven by renderer complexity times the number of cells on screen, never by how many records you hold.

**A public per-cell API is still the wrong answer.** Every constraint the technique carries is a caller promise the engine cannot verify, so a public method would either corrupt the layout silently or degrade to a full render in the common cases. DEV-2759 (#13341) documented the absence of a per-cell render as a deliberate trade; this PR does not reopen that decision, and the new cross-links are worded so they do not contradict it.

#### What the recipe covers

`recipes/performance/repaint-single-cell` (JavaScript only). Cancel the render with `beforeViewRender`'s `skipRender`, **and repaint inside that same hook** — before Handsontable applies the selection classes at the end of the draw. The recipe reproduces the three things Handsontable does per cell that are not public API: the `td` reset, the `valueFormatter` precedence, and the base-renderer chaining guarded by `_isBaseRendererCalled`.

#### Limitations, each measured

1. **Row heights are not re-measured.** `markOversizedRows` runs only inside the band render. A wrapping value left the main table **240 px out of alignment** with its frozen columns. Self-heals on the next full render.
2. **`autoRowSize` / `autoColumnSize` must be off.** They sample by running the renderer in an off-screen `GhostTable`, which the cancelled render never reaches — with them on, the repaint still cost **36** renderer calls instead of 1.
3. **Cells in frozen areas need a full render.** There is no per-axis rule: a cell is also in the main table iff its index is at or after the main table's first rendered index, which moves with the frozen count *and* the scroll position, symmetrically on both axes. At `fixedRowsTop: 1, fixedColumnsStart: 1`, cell (0,0) exists in **four** tables and `getCell()` reaches only two, so the gate hands every frozen-area cell to a normal render.
4. **Renderers that read other cells go stale.** There is no dependency graph between cells.
5. **Structural changes must fall through** to a full render.
6. **`afterViewRender` does not fire on the skip path.** This is why the recipe is JavaScript only: the React wrapper clears its cell-portal cache on `beforeViewRender` and restores it on `afterViewRender` (`hotTableInner.tsx:192`), so cancelling a render on a grid with React component renderers would clear the portals and never restore them. A React example was written and then deleted rather than shipped unverified.

#### Cross-links, for discoverability

- The [Understanding rendering](https://github.com/handsontable/handsontable/blob/develop/docs/content/guides/optimization/rendering/rendering.md) guide, at *"There is no per-cell or per-row render"*.
- The `render()` JSDoc, after *"There is no API for rendering a single cell"*.
- The `beforeViewRender` hook JSDoc, which also gains a description of what `skipRender` does and does not do — that was not written down anywhere.
- `handsontable/AGENTS.md`, noting that the recipe mirrors `renderCell`, `formatCellValue` and the `td` reset by hand, so the three must be updated together.

#### Follow-up worth filing

Export `renderCell()` and `formatCellValue()` from the package. They already exist and are shared by `TableView`, `GhostTable` and the auto-size samplers, but are not public — which is why the recipe restates the base-renderer chaining by hand. Exporting them is small, non-breaking, and would make every future recipe correct by construction.

#### Second commit — what a code review caught

The first commit's verification exercised the UMD full bundle, not the `handsontable/base` entry the example imports. Three defects followed, all fixed in `bb167d678e`:

- **The example did not run.** `handsontable/base` never defines `Handsontable.renderers` (only the full entry does), so `Handsontable.renderers.TextRenderer` threw on the first draw. Now imports `textRenderer` and `baseRenderer` directly, the established idiom for base-entry examples.
- **The repaint wiped the selection.** Repainting from `afterChange` ran after the selection classes had been applied, so editing a selected cell — the common case — lost its highlight. The repaint moved into the `beforeViewRender` handler, restoring the engine's own reset → render → selections order.
- **The frozen-row versus frozen-column asymmetry was wrong** and is retracted; see limitation 3 above. The first commit message states the old claim — please read this description instead.

Also fixed: resolve the renderer from the cell metadata already held rather than from row and column (the row/column form re-ran the `cells` function the recipe exists to avoid), decide from the change's `source` rather than a flag around `setDataAtCell` (an async validator closes that window before the decision is made), compare with `!= null` since `getCell()` can answer `undefined`, correct the prose that inverted `getCell()`'s default, and wrap the example in `::: only-for javascript` as both sibling performance recipes do.

### How has this been tested?

Verified through an import map against the published 18.1.0 **ESM** build, so the `handsontable/base` path is the one actually exercised:

- Full render: **66 renderer calls, ~16 ms**. Per-cell repaint: **1 call, ~0.8 ms**. Stable across repeated runs.
- Console clean — no errors or warnings.
- Selection class preserved on the edited cell after a repaint (`current` before and after).
- Base-renderer chaining verified: `htDimmed` survives a repaint on a read-only cell, and `_isBaseRendererCalled` is reset so the next full render still applies the base renderer.
- Accessibility attributes (`role`, `tabindex`, `aria-colindex`) restored by the repaint.
- The frozen-copy matrix in limitation 3 was measured by counting every `td` in the DOM across six `fixedRowsTop` / `fixedColumnsStart` combinations, at the top-left scroll position and scrolled.
- Limitations 1, 2 and 6 were each reproduced and measured, not inferred.

`docs/tests/recipeConsoleErrors.spec.ts` picks the new page up automatically and asserts a clean console.

```bash
npx eslint src/core.ts src/core/hooks/constants.ts --prefix handsontable   # 0 errors
```

Documentation and JSDoc only — no runtime code changed, so no unit or E2E run applies. Both commits carry a `Refactor-only:` trailer and the pre-push test-presence gate passed on each.

### Types of changes

### Related issue(s):

1. DEV-2779
2. #9614
3. #5769 (closed, moved to the forum — context for why #9614 was never answered)

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

[skip changelog]

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2779

---

> [!NOTE]
> **Low Risk**
> Documentation and JSDoc only; no runtime or public API changes. The recipe documents advanced internal hooks that can mis-render if misapplied, but that risk is for adopters of the pattern, not for the library itself.
> 
> **Overview**
> Adds a **JavaScript-only performance recipe** that shows how to use `beforeViewRender` + `skipRender` to cancel the post-edit viewport render and repaint only the changed cell(s) by calling the same internal `cellRenderer` path Handsontable uses—without shipping a public per-cell API.
> 
> The recipe includes a working demo (full render vs targeted repaint with renderer-call counts), safety gates (change `source`, max cells, frozen areas, row/column shape, visible `td`, viewport band offset), and explicit limitations (fixed row heights, no auto-sizing, wrapper/React portal cache).
> 
> **Discoverability:** links from the rendering guide, performance recipe index, recipes sidebar, `render()` JSDoc, and expanded `beforeViewRender` hook docs. **Contributor docs** in `AGENTS.md` and Walkontable `AGENTS.md` document the `getCell()` + `skipRender` trap and tie the recipe to `renderCell.ts` maintenance.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7e1cf9a753959b72f842a72a1d41043b97a23b5b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2759 Document the render() method: what it covers, what it costs, and when to call it](https://app.clickup.com/t/9015210959/DEV-2759)
Task linked: [DEV-2779 Recipe: repaint a single cell instead of the whole viewport (GH #9614)](https://app.clickup.com/t/9015210959/DEV-2779)

PR #13369, opened 2026-09-03
DEV-412: Keep frozen overlays above a rotated master layer

### LINKED ISSUE #6223: [transform(rotate) on mobile] table renders with multiple problems.
### Steps to reproduce

1. get the html file from here: https://github.com/DongShi/issues-demo/tree/master/handsontable/rotate
2. serve it with a http server
3. visit the page on a mobile phone and do scrolling

### Your environment
* Handsontable version: 7.0.0
* Browser Name and version: safari, chrome on phone.
* Operating System: ios android

Scenario:

- In our mobile app, we use HOT to render tables.  However, in portrait mode, user can see just a few columns of data, due to the limited screen width. So user want to read the table in a 'landscape' view.

- Our mobile app does not support change orientation, so we decide to add a button along the hot table and once the button is clicked, we transform the DOM of HOT by rotating it 90 deg. This way, we mimic a landscape view and user can see more columns of data at first sight.

There are two major problems:
1. The width is clipped:
image

I think this one may be similar to #5050 #6201

2. when you scrolling the rotated table, you will see the major table (ht_master) goes above the left fixed columns (ht_clone_left)
image

Please get them fixed, earlier is better.

demtario:
### Context

The PR fixes frozen-column overlap when a host applies `transform: rotate()` (GitHub #6223, DEV-412). Without a stacking context on `.ht_master`, a browser-promoted master scroll layer can paint above the frozen overlays after scroll. The change gives `.ht_master` `z-index: 0` so that layer stays below the clones.

- Unit tests added/modified (`*.unit.js`): none — covered by `tests/e2e/css-transform-rotate.spec.ts`
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`): `tests/e2e/css-transform-rotate.spec.ts`, `tests/e2e/fill-handle-frozen-panes.spec.ts`
- Type tests (`*.types.ts`) updated if public API changed: none
- For a bug fix — the spec that fails without this fix: `tests/e2e/css-transform-rotate.spec.ts` (`keeps the frozen columns above a promoted master scroll layer`)
- Demo page / recorded trace (for UI changes): fixture `tests/fixtures/demo/css-transform.html`

### Commands run

```
HOT_TEST_PORT=8137 npx playwright test e2e/css-transform-rotate.spec.ts
# PASS (12) FAIL (0)

HOT_TEST_PORT=8137 npx playwright test e2e/mobile-selection-handles.spec.ts e2e/fill-handle-frozen-panes.spec.ts --project=e2e-main
# PASS (9) FAIL (0)

npm run lint --prefix tests
```

### Types of changes

### Related issue(s):

1. DEV-412
2. #6223

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

- [ ] MANUAL QA NEEDED — 

ClickUp task: https://app.clickup.com/t/9015210959/DEV-412

---

> [!NOTE]
> **Medium Risk**
> Touches overlay z-ordering and mobile selection-handle positioning on the master table; wrong stacking or handle math could affect hit-testing at freeze/header seams, though E2E targets the main regressions.
> 
> **Overview**
> Fixes frozen columns painting above the scrollable master when the grid sits under a CSS `transform` (e.g. rotate): **`.ht_master` now uses `z-index: 0`** so the master scroll layer stays in a stacking context below overlay clones (`walkontable.scss`, `_z-index-map.scss`).
> 
> Because that blocks the legacy mobile-handle trick of **`z-index: 9999` escaping clones**, **`border.ts` repositions** top/bottom touch handles when a `top`, `inline_start`, or `bottom` clone would cover them—keyed on **whether those overlays render** (headers count even with `fixedRowsTop: 0`), not only on frozen-pane counts.
> 
> Adds **Playwright coverage** for rotated hosts (`css-transform-rotate.spec.ts`) and expanded mobile-handle cases (freeze lines, RTL, headers on/off), plus fixture/page-object updates.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4ed0cd62da90e1c754e2ab85a61abec2c76cfcef. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-412 `transform` (rotate) on mobile table renders with multiple problems.](https://app.clickup.com/t/9015210959/DEV-412)

handsontable/src/styles/base/_z-index-map.scss
```diff
@@ -4,15 +4,14 @@
...
-// frozen overlays draw their own copy of every affordance, so none of them needs to outrank a clone
-// to be visible inside a pane.
+// The master table starts a stacking context at 0, so its descendants stay below the frozen
+// overlays even when the browser promotes its scroll holder to a composited layer. The frozen
+// overlays draw their own copy of every affordance.
 // The clone values are declared twice: here, and in
 // `handsontable/src/3rdparty/walkontable/css/walkontable.scss`. Keep both in sync.
 
 $z-indexes: (
+  "ht_master": 0,
```
budnix, line 14: This also caps the mobile selection handles. `_mobile.scss` gives them `z-index: 9999` and `border.ts` sets it inline again when a selection edge lands on a freeze line (#9850), and until now that 9999 cleared `.ht_clone_inline_start` at 120 because the master opened no stacking context. Nothing would tell us if that changed, since the mobile specs only assert the value is still `9999` and `mobile-handles.html` sets no `fixedColumnsStart`, so the frozen case never renders. Would be good to check this by hand on a mobile grid with frozen columns and pin whichever behavior is the right one.

tests/e2e/css-transform-rotate.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+test.describe('a grid rotated by its host element', () => {
+  test('uses the untransformed layout box for the scroll viewport', async ({ page, theme, bundle }) => {
+    const grid = new CssTransformGridPage(page, theme, bundle);
+
+    await grid.goto();
+
+    const { width, height } = await grid.viewportSize();
+
+    expect(width).toBeGreaterThan(height);
+    expect(width).toBeGreaterThan(450);
```
budnix: Can I ask - does this one fail on `develop`? `clientWidth` and `clientHeight` read the layout box, which a CSS `transform` never changes, and this PR only touches `z-index`, so I think it passes either way. DEV-412 lists several symptoms, so maybe it is here to pin the viewport-size one - if so a short comment saying that would help, because right now it reads like a guard that cannot go red.

tests/fixtures/demo/css-transform.html
```diff
@@ -0,0 +1,56 @@
...
+    const params = new URLSearchParams(location.search);
+
+    window.htTheme = ({ main: 'main', horizon: 'horizon', classic: 'classic' })[params.get('theme') ?? 'main'];
+    window.htBundle = ({ umd: 'handsontable.js', 'full-min': 'handsontable.full.min.js' })[
+      params.get('bundle') ?? 'umd'
+    ];
+
+    if (!window.htTheme || !window.htBundle) {
+      throw new Error('Unknown theme or bundle.');
+    }
```
budnix, line 21: When a project config has a typo this throws without saying which param was wrong or what it held, so the red leg only says "Unknown theme or bundle". Every other fixture splits the two checks and puts the bad value in the message, which is what makes a mislabeled leg easy to spot. Worth matching them here.

```suggestion
    window.htTheme = ({ main: 'main', horizon: 'horizon', classic: 'classic' })[params.get('theme') ?? 'main'];
    if (!window.htTheme) {
      throw new Error('Unknown ?theme= value: ' + JSON.stringify(params.get('theme')));
    }

    window.htBundle = ({ umd: 'handsontable.js', 'full-min': 'handsontable.full.min.js' })[
      params.get('bundle') ?? 'umd'
    ];
    if (!window.htBundle) {
      throw new Error('Unknown ?bundle= value: ' + JSON.stringify(params.get('bundle')));
    }
```

handsontable/src/styles/base/_z-index-map.scss
demtario: Confirmed: the new master stacking context made the frozen overlay own the entire mobile handle hit area. I preserved the interaction by repositioning both the visible top handle and its 40px hit area into the scrollable pane when the selection starts on a fixed-row or fixed-column boundary. The E2E fixture now includes both frozen axes, and the regression is covered for multi-cell ranges in LTR and RTL.

tests/fixtures/demo/css-transform.html
demtario: Updated to the split, fail-loud checks with the invalid query value in each message. I also aligned the mobile-handles fixture with the same theme/bundle contract while extending its coverage.

tests/e2e/css-transform-rotate.spec.ts
demtario: You are right: this assertion passed on develop because clientWidth/clientHeight use the untransformed layout box. I removed the test and the now-unused page-object method; the remaining spec only pins the stacking regression.

handsontable/src/3rdparty/walkontable/src/selection/border/border.ts
```diff
@@ -628,12 +638,20 @@ class Border {
...
+    }px`;
 
-    topHitAreaStyles.top = `${parseInt(String(top - ((hitAreaSize / 4) * 3)), 10)}px`;
-    topHitAreaStyles[inlinePosProperty] = `${parseInt(String(left - ((hitAreaSize / 4) * 3)), 10)}px`;
+    topHitAreaStyles.top = `${
+      parseInt(String(isAtFrozenRowBoundary ? top : top - ((hitAreaSize / 4) * 3)), 10)
+    }px`;
+    topHitAreaStyles[inlinePosProperty] = `${
+      parseInt(String(isAtFrozenColumnBoundary ? left : left - ((hitAreaSize / 4) * 3)), 10)
+    }px`;
```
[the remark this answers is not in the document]
demtario: Good catch, this reproduced. With `fixedRowsTop: 0`, selecting any row-0 cell satisfied `fromRow === fixedRowsTop` and took the inset path: the handle bottom landed 15px below the cell top instead of resting on the corner.

Fixed by routing the positioning through the existing guarded helper, `isFrozenBoundaryEdge`, which requires the fixed count to be greater than 0. The `zIndex` block keeps its original unguarded comparison, so no pre-existing behavior changes there.

Covered by a new spec asserting the handle stays on the cell outer corner for a row-0 selection without frozen panes. I confirmed it fails on the previous commit and passes now, and the frozen panes are opt-in in the fixture so the original specs run against their unfrozen configuration again.

budnix (dismissed):

budnix (approved):

PR #6157, opened 2019-07-28
Improve custom borders performance (SVG)

warpech:
### Context

The goal for this PR is to improve CustomBorders feature so that even complex custom border setup does not have a significant impact on performance of Handsontable.

It started out as a solution to https://github.com/handsontable/handsontable/issues/6064, but during the course of fixing it turned out that the reason is a lot of redundant actions in `border.js`, `selection.js` and `customBorders.js`.

Border edge lines are no longer rendered by the `Border` class. Instead, they are rendered using SVG, by a new low-level function `getSvgPathsRenderer`, managed in a new class `BorderRenderer`. There is one SVG per overlay. SVG is rendered atop the table, with `pointer-events: none`. Every stroke style is two `<path>`s in SVG (one `horizontal`, one `vertical`). If you have million borders, but all in one style, that's just one path element. Paths are windowed (= only the part of the border that corresponds to the rendered cells is updated at each scroll).

The actual order and position of borders is implemented according to the spec in KB: https://kb.handsontable.com/display/HAN/Borders

Class `Border` was renamed to `SelectionHandle`, because border rendering is removed from it. All `SelectionHandle` does now is rendering the corner or the selection handles on mobile.

Breaking changes:

1. 🚨 The visual position of borders is now different. The borders used to be rendered inside of a cell. Now they are entered on the gridlines.
2. 🚨 Overwriting border styles using CSS is no longer possible. The only way to set border styles is through the settings object. You can see in a commit that changes the selection border of the custom menu from being defined in CSS into being defined in `menu.js`: https://github.com/handsontable/handsontable/commit/1b78d48ef7190d913d1da666a6953ad90cd2c9b6
2. 🚨 In the past, border edges would overlap the header cell edges. Now they don't, just like in desktop spreadsheet apps. Old:  
image  
New:  
image

### How has this been tested?

Manually using such config: https://gist.github.com/warpech/81686e38fa0bc5c8a35e7e2dfbb2e408

This config has 100x100 cells with 4 borders each. It caused HOT 7.1.0, as well as the current develop (2531da4e24b30f16d37ca5734eab80d74cf91396) to freeze. After this PR, such config loads in 4.7s in Chrome.

The borders should looks sharp in every browser at 100% zoom level. They might be blurry at other zoom levels that are not multiplies of 100.

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature or improvement (non-breaking change which adds functionality)
[...]

PR #13371, opened 2026-09-03
DEV-395: Fix the first column rendering 1px narrower with row headers

### LINKED ISSUE #6673: If there are row headers, the first column has a wrong width
### Description

The width of the first column depends on the visibility of the row headers, and this is causing an annoying inconsistency: If the row headers are present, the width of the first column excluding the gridline is 1px smaller than the widths of any other column.

The problem appears with and without `colWidths` attribute, however it is more annoying with `colWidths`, because when I set a specific width for a column, I expect that it will be applied in a consistent way to all columns.

Consider the example, in which all columns are set to 75. In rendering, the width of the first column excluding gridlines is 73, while other columns are 74. 

I would expect that the width of the column excluding the gridline should be the same for all columns, so I would expect value 74 everywhere.

We can argue if it is correct that the gridline is included in the column width; however the line separating the headers and the cells should not be considered a gridline. In the popular desktop spreadsheet application, column widths include the gridline but not the line that separates the header. So, this supports the claim that 74 is the correct width excluding the gridlines for all columns.

widths

### Steps to reproduce

1. Create a table that is configured to have row headers and equal column widths
2. Verify that the column widths excluding the gridline are not equal

### Demo
https://codesandbox.io/s/brave-edison-s6s6x

### Your environment
* Handsontable version: 7.3.0
* Browser Name and version: any
* Operating System: any

demtario:
### Context

The PR fixes the first data column rendering 1px narrower than every other column whenever row headers are enabled. With `colWidths: 75` and `rowHeaders: true`, column 0's content box measured 73px while every other column measured 74px, so one declared width produced two different rendered widths.

**The linked issue says this was already fixed. It was not.** The commit it cites (`59b71c2942`) belongs to PR #6157, which was closed **unmerged** on 2022-01-13; the commit exists only on abandoned `feature/issue-6064*` branches and never reached `develop`. The defect was reproduced on current `develop` in all three themes before any work started.

**Root cause.** `handsontable/src/styles/base/_base.scss` gave `td:first-of-type` a `border-inline-start` on top of the `border-inline-end` that every cell carries, and `box-sizing: border-box` took both out of the same `<col>` width. `:first-of-type` matches the first `td` whether or not a row-header `th` precedes it, so column 0 was asked to fit two gridlines into one column width.

To stop the seam from doubling, the row header then gave up its own `border-inline-end` at horizontal offset 0, and a `correctHeaderWidth` flag in the engine widened the row header by 1px the moment the grid scrolled — with matching compensations in `inlineStartOverlay#scrollTo`, the hider width, and both column calculators (`viewportWidth + 1`). **The whole table therefore changed width by being scrolled.**

**Fix.** The gridline between the row header and column 0 now belongs to the row header at every scroll position, and no cell standing behind a row header draws an inline-start border. Every column then has the same content width, and the row header's width no longer depends on the scroll offset, so all four compensations are removed rather than rebalanced.

Removing the calculator `+ 1` also tightens grids **without** row headers by one pixel when scrolled: a column whose end sits 1px past the viewport edge is no longer reported fully visible. That pixel was latent slack introduced alongside `correctHeaderWidth`; keeping it would have created a fresh off-by-one for the row-header case now that `getViewportWidth()` is exact at every offset.

While verifying, one defect this change introduced was found and fixed in the same PR: `BaseEditor#getEditedCellRect` cancelled its 1px inline-start shift for `renderableColumn <= 0`, on the assumption that column 0 owns its own start border. With row headers that left the editor 51px wide but flush with the cell's left edge, overhanging into the next column. The offset now keys off the same computed border that already sizes the editor, so which cells get the shift is read from the DOM rather than from an index.

**Deliberately out of scope, and pinned as a control case:** with no row header in front of it, column 0 is the first cell of its row and still draws the grid's own inline-start frame inside its declared width. It stays 1px narrower than the rest, exactly as before.

**Also deliberately untouched:** the row axis (`innerBorderTop` / `innerBorderBottom`, `columnHeaderBorderCompensation`) has the identical design and the identical latent problem. Bringing it into line is a separate change; a follow-up task should be filed.

The `innerBorderInlineStart`, `innerBorderLeft` and `emptyColumns` classes are still stamped on `.ht_master` for backward compatibility. No stylesheet reads them any more, and a test pins that they are still published.

### How has this been tested?

New Playwright spec, plus updated expectations in the existing suites wherever a number encoded the removed pixel.

**New spec** — `tests/e2e/row-header-border-ownership.spec.ts` (fixture + page object alongside it). Covers the reported case, the frozen-column clone, RTL, nested headers, invariance of the row-header width and the scroll range across a horizontal scroll, the legacy classes, and the no-row-headers control.

```bash
cd tests && npx playwright test e2e/row-header-border-ownership.spec.ts
```

**Verified the spec is not vacuous.** The source was reverted to the base commit, rebuilt, and the spec re-run: **8 of its 10 tests fail without the fix.** The 2 that pass are the intentional guards (the legacy classes, and the no-row-headers control), which must hold in both states.

| Suite | Result |
|---|---|
| `tests/e2e` Playwright, new spec, all 6 theme × bundle legs | 60/60 passed |
| `tests/e2e` Playwright, full suite (`e2e-main`) | 855 passed, 1 skipped |
| `npm run test:e2e --prefix handsontable` | 9848 specs, 0 failures |
| `npm run test:walkontable --prefix handsontable` | 816 specs, 0 failures |
| `npm run test:unit --prefix handsontable` | 4393 passed |
| `npm run test --prefix wrappers/react-wrapper` | 99 passed |
| `npm run test --prefix wrappers/vue3` | 39 passed |
| `npm run test --prefix wrappers/angular-wrapper` | 299 passed |
| `npm run eslint` + `stylelint --prefix handsontable` | 0 errors |

**Manually verified in a real browser** across `main`, `horizon` and `classic`, with and without row headers, frozen columns, RTL and nested headers: every column's content box is equal; the row header stays 50px and the hider width stays constant across a horizontal scroll; the seam renders exactly once (no missing and no doubled gridline); the last column is fully reachable at maximum scroll; and the editor aligns identically for column 0 and every other column.

**Updated existing expectations.** Every changed number was checked against the browser first, not adjusted until green:

- Horizontal scroll positions that embedded the compensation, each 1px lower (`core/viewportScroll/__tests__/**` incl. RTL, `scrollViewportTo`, `scrollToFocusedCell`, `beforeViewportScrollHorizontally`, `nestedHeaders/navigation`, `comments`, `dropdownMenu`). Two of those specs' own comments named the removed term as "+ 1 header border compensation".
- `focusSelection.spec.js` no longer needs the 1px of container slack that existed to absorb the first column's border, so scroll positions land on exact `colWidths` multiples again on all three themes.
- Editor position specs across all cell-type editors: column 0 now follows the same convention as every other column. The RTL helper reads the border off the cell instead of taking a flag from the caller.
- The two frozen-column seam specs in `autoRowSize` and `manualColumnMove` asserted the seam on the cell's `border-left`; they now assert it on both sides of the seam, so a missing gridline and a doubled one are both failures.
- The Walkontable overlay spec's `+ 1px (header border left width)` term is gone; the `drawCycleHooks` spec that pinned `correctHeaderWidth` now pins the invariant that replaced it (the row header width is unchanged across a skipped horizontal-scroll draw).

**Visual tests will change.** Every screenshot with row headers shifts by 1px: the row header's content box is 1px narrower, column 0's is 1px wider, and a selection border on column 0 moves with it. This is the intended effect and needs a reg-suit review plus the `visual-approved` label.

### Types of changes

### Related issue(s):

1. #6673
2. DEV-395

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-395

---

> [!NOTE]
> **Medium Risk**
> Touches core Walkontable layout, scrolling, frozen overlays, RTL, and editor/selection geometry; behavior is well-tested but regressions would show as 1px misalignment or wrong visible columns when scrolled.
> 
> **Overview**
> Fixes **#6673**: with row headers enabled, column 0 no longer renders 1px narrower than other columns, and the row header no longer grows by 1px when the grid is scrolled horizontally.
> 
> **Border ownership** moves the gridline between the row header and column 0 to the row header’s `border-inline-end` at every scroll offset. Data cells behind row headers no longer get an extra inline-start border (`:first-of-type` → `:first-child` in Walkontable and theme SCSS). Seam color rules under `.htRowHeaders` keep overlays and active-header accents consistent.
> 
> **Engine cleanup** drops the scroll-dependent `correctHeaderWidth` path, hider horizontal `+1`, `expandHiderHorizontallyBy`, and the `viewportWidth + 1` column calculators. Horizontal scroll/viewport expectations and editor/selection positioning now use DOM border reads (`BaseEditor`, `Border#appear`, `editorInlineStartOffset` helper) instead of column index.
> 
> Adds Playwright coverage in `tests/e2e/row-header-border-ownership.spec.ts` and updates broad unit/e2e expectations (typically 1px lower scroll positions). `innerBorderInlineStart` / `innerBorderLeft` remain on the DOM for compatibility but no longer affect layout.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 30243c3535d0442db5acc8164c9cca12eaac0860. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
+++ b/handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
@@ -20,8 +20,7 @@ import type { ColumnsCalculationType, RowsCalculationType } from '../calculator/
 
 /**
  * The state that describes what is actually rendered in the TBODY/THEAD right now: the rendered
- * row/column bands (held by the viewport), the render filters derived from them, and the
- * `correctHeaderWidth` flag that describes the row-header width the DOM was rendered with. Captured
+ * row/column bands (held by the viewport) and the render filters derived from them. Captured
  * before the draw resolves the new band, so the master's `skipRender` gate can put it back — see
  * {@link restoreRenderedStateIfSafe}.
  */
@@ -30,7 +29,6 @@ interface RenderedState {
   columnsRenderCalculator: ColumnsCalculationType | null;
   rowFilter: RowFilter | null;
   columnFilter: ColumnFilter | null;
-  correctHeaderWidth: boolean;
 }
 
 /**
@@ -149,17 +147,6 @@ function runMasterDrawCycle(table: Table, ctx: DrawContext): void {
     stationaryBands: wtOverlays.isScrollDrivenDraw && wtViewport.allowsStationaryBands(),
   });
 
-  if (ctx.rowHeadersCount && !wtSettings.getSetting('fixedColumnsStart')) {
-    const leftScrollPos = wtOverlays.inlineStartOverlay.getScrollPosition();
-    const previousState = table.correctHeaderWidth;
-
-    table.correctHeaderWidth = leftScrollPos !== 0;
-
-    if (previousState !== table.correctHeaderWidth) {
-      ctx.runFastDraw = false;
-    }
-  }
-
   if (ctx.runFastDraw) {
     wtOverlays.refresh(true);
     // Fast (scroll) draws skip the full-render header-height pass below, so the master/top
@@ -270,8 +257,10 @@ function runMasterDrawCycle(table: Table, ctx: DrawContext): void {
   if (ctx.positionChanged) {
     if (ctx.performRedraw) {
       // It refreshes the cells borders caused by a 1px shift (introduced by overlays which add or
-      // remove `innerBorderTop` and `innerBorderInlineStart` CSS classes to the DOM element. This happens
-      // when there is a switch between rendering from 0 to N rows/columns and vice versa).
+      // remove the `innerBorderTop` / `innerBorderBottom` CSS classes on the DOM element. This
+      // happens when there is a switch between rendering from 0 to N rows and vice versa). The
+      // inline-start class is still toggled for backward compatibility but no longer shifts the
+      // layout — the row header owns its inline-end border at every scroll position (#6673).
       wtOverlays.refreshAll(); // `refreshAll()` internally already calls `refreshSelections()` method
     } else {
       // A skipped render must not run the nested reconciliation draw above (`refreshAll` is
@@ -362,9 +351,9 @@ function buildRenderFilters(table: Table, ctx: DrawContext): { rowFilter: RowFil
 }
 
 /**
- * Captures the rendered row/column bands, the render filters, and the `correctHeaderWidth` flag,
- * i.e. everything that describes which source rows and columns the current DOM holds and how it was
- * rendered. Master-only: the `skipRender` gate it serves is master-only.
+ * Captures the rendered row/column bands and the render filters, i.e. everything that describes
+ * which source rows and columns the current DOM holds. Master-only: the `skipRender` gate it serves
+ * is master-only.
  *
  * @param {Table} table The master table.
  * @param {Viewport} wtViewport The viewport that owns the rendered bands.
@@ -376,7 +365,6 @@ function captureRenderedState(table: Table, wtViewport: Viewport): RenderedState
     columnsRenderCalculator: wtViewport.columnsRenderCalculator,
     rowFilter: table.rowFilter,
     columnFilter: table.columnFilter,
-    correctHeaderWidth: table.correctHeaderWidth,
   };
 }
 
@@ -408,10 +396,6 @@ function captureRenderedState(table: Table, wtViewport: Viewport): RenderedState
  *   just-built filters are kept instead, because filter consumers (`getRowHeader`, `getTrForRow`,
  *   the selection scanner's header loop) read `rowFilter!` unguarded and must never see `null`
  *   after a completed draw.
- * - `correctHeaderWidth` is restored whenever no render happened, regardless of the totals gates:
- *   the flag describes the row-header width the DOM was rendered with, and that DOM did not change.
- *   Leaving it advanced would make the next draw see "no change" and never re-render the corrected
- *   header width.
  * - The visible-row/column calculators are deliberately NOT restored: they describe what the user
  *   can see (scroll position), not what the DOM holds. After a skipped draw the visible band may
  *   therefore extend past the rendered band — unlike a fast draw, which guarantees the visible band
@@ -434,8 +418,6 @@ function restoreRenderedStateIfSafe(
   if (wtViewport.renderCycleSeq !== renderCycleSeqBeforeHook) {
     return;
   }
-
-  table.correctHeaderWidth = state.correctHeaderWidth;
 
   if (state.rowFilter === null || state.rowFilter.total === wtSettings.getSetting<number>('totalRows')) {
     wtViewport.rowsRenderCalculator = state.rowsRenderCalculator;
--- a/handsontable/src/3rdparty/walkontable/test/spec/table/drawCycleHooks.spec.js
+++ b/handsontable/src/3rdparty/walkontable/test/spec/table/drawCycleHooks.spec.js
@@ -311,7 +311,7 @@ describe('Table.draw() lifecycle hooks (characterization for the drawCycle refac
     }
   });
 
-  it('should keep the row header width unchanged across a horizontal scroll whose render is skipped', async() => {
+  it('should keep the column rollback across a horizontal scroll whose render is skipped', async() => {
     let skipNextRender = false;
     const wt = walkontable({
       data: getData,
@@ -331,17 +331,35 @@ describe('Table.draw() lifecycle hooks (characterization for the drawCycle refac
 
     const rowHeaderWidth = () => wt.wtTable.TABLE.querySelector('tbody tr th').offsetWidth;
     const widthBeforeScroll = rowHeaderWidth();
+    const renderedColumnsBeforeScroll = wt.wtTable.TBODY.firstChild.querySelectorAll('td').length;
 
     expect(widthBeforeScroll).toBeGreaterThan(0);
 
-    // The row header owns its inline-end border at every scroll position, so its width is scroll
-    // independent - a skipped draw cannot leave the DOM describing a width the engine no longer
-    // reports (#6673). Before that change the header grew by 1px once the table scrolled, and the
-    // flag that tracked it had to be rolled back here.
+    // The column axis is what the rollback still protects here. Without it the advanced column band
+    // describes columns the skipped draw never rendered, so the filter stops matching the DOM and
+    // `getCell` reads past it.
     skipNextRender = true;
     wt.scrollViewportHorizontally(3, 'end');
     wt.draw();
 
+    // The render was really skipped - the TBODY still holds the pre-scroll column band.
+    expect(wt.wtTable.TBODY.firstChild.querySelectorAll('td').length).toBe(renderedColumnsBeforeScroll);
+
+    // A rolled-back band describes the stale DOM, so every cell it resolves really holds that
+    // column's value. Without the rollback the band names the post-scroll columns while the TBODY
+    // still holds the pre-scroll ones, and `getCell` maps a new index onto an old cell - an element,
+    // but the wrong one, which is why an `instanceof HTMLElement` check cannot see this.
+    expect(wt.wtTable.getFirstRenderedColumn()).toBeGreaterThan(-1);
+
+    for (let col = wt.wtTable.getFirstRenderedColumn(); col <= wt.wtTable.getLastRenderedColumn(); col++) {
+      expect(wt.wtTable.getCell(new Walkontable.CellCoords(0, col)).textContent)
+        .toBe(`${getData(0, col)}`);
+    }
+
+    // And the row header width is scroll independent, so neither the skipped draw nor the one that
+    // follows it can leave the DOM describing a width the engine no longer reports. It used to grow
+    // by 1px once the table scrolled, which is what the removed `correctHeaderWidth` flag tracked
+    // and what this spec had to roll back (#6673).
     expect(rowHeaderWidth()).toBe(widthBeforeScroll);
 
     skipNextRender = false;
--- a/handsontable/src/3rdparty/walkontable/test/spec/table/drawCycleHooks.spec.js
+++ b/handsontable/src/3rdparty/walkontable/test/spec/table/drawCycleHooks.spec.js
@@ -311,7 +311,7 @@ describe('Table.draw() lifecycle hooks (characterization for the drawCycle refac
     }
   });
 
-  it('should restore the `correctHeaderWidth` flag when the render is skipped', async() => {
+  it('should keep the row header width unchanged across a horizontal scroll whose render is skipped', async() => {
     let skipNextRender = false;
     const wt = walkontable({
       data: getData,
@@ -329,16 +329,25 @@ describe('Table.draw() lifecycle hooks (characterization for the drawCycle refac
 
     wt.draw();
 
-    expect(wt.wtTable.correctHeaderWidth).toBe(false);
+    const rowHeaderWidth = () => wt.wtTable.TABLE.querySelector('tbody tr th').offsetWidth;
+    const widthBeforeScroll = rowHeaderWidth();
 
-    // The flag flips before the `beforeDraw` gate, but the header it describes never re-renders on
-    // a skipped draw. Left advanced, the next draw would see "no change" and keep the stale header
-    // width forever - so the rollback must put the flag back with the rest of the rendered state.
+    expect(widthBeforeScroll).toBeGreaterThan(0);
+
+    // The row header owns its inline-end border at every scroll position, so its width is scroll
+    // independent - a skipped draw cannot leave the DOM describing a width the engine no longer
+    // reports (#6673). Before that change the header grew by 1px once the table scrolled, and the
+    // flag that tracked it had to be rolled back here.
     skipNextRender = true;
     wt.scrollViewportHorizontally(3, 'end');
     wt.draw();
 
-    expect(wt.wtTable.correctHeaderWidth).toBe(false);
+    expect(rowHeaderWidth()).toBe(widthBeforeScroll);
+
+    skipNextRender = false;
+    wt.draw();
+
+    expect(rowHeaderWidth()).toBe(widthBeforeScroll);
   });
 
   it('should fire `beforeDraw` but SKIP `onDraw` when beforeDraw sets skipRender', async() => {
--- a/handsontable/src/editors/textEditor/__tests__/textEditor.spec.js
+++ b/handsontable/src/editors/textEditor/__tests__/textEditor.spec.js
@@ -140,7 +140,12 @@ describe('TextEditor', () => {
 
     await keyDownUp('F2');
 
-    expect(editor.offset()).toEqual($(getCell(0, 0)).offset());
+    // Every cell here sits behind a row header, which owns the gridline on its inline-start
+    // side, so the editor starts 1px before the cell in order to cover it (#6673).
+    expect({
+      top: editor.offset().top,
+      left: editor.offset().left + 1,
+    }).toEqual($(getCell(0, 0)).offset());
   });
 
   it('should render the editor in the expected position when stepping top-to-bottom with top and bottom overlays', async() => {
@@ -174,15 +179,22 @@ describe('TextEditor', () => {
 
     await keyDownUp('enter');
 
-    expect(editor.offset()).toEqual($(getCell(0, 0, true)).offset());
+    // Every cell here sits behind a row header, which owns the gridline on its inline-start
+    // side, so the editor starts 1px before the cell in order to cover it (#6673).
+    const editorOffsetAtRowEdge = () => ({
+      top: editor.offset().top,
+      left: editor.offset().left + 1,
+    });
+
+    expect(editorOffsetAtRowEdge()).toEqual($(getCell(0, 0, true)).offset());
 
     await keyDownUp('enter');
     await keyDownUp('enter');
 
-    // Cells that do not touch the edges of the table have an additional top border.
+    // Cells that do not touch the edges of the table also have an additional top border.
     const editorOffset = () => ({
       top: editor.offset().top + 1,
-      left: editor.offset().left,
+      left: editor.offset().left + 1,
     });
 
     expect(editorOffset()).toEqual($(getCell(1, 0, true)).offset());
@@ -206,7 +218,7 @@ describe('TextEditor', () => {
     await keyDownUp('enter');
 
     // The first row of the bottom overlay has different position, influenced by `innerBorderTop` CSS class.
-    expect(editor.offset()).toEqual($(getCell(5, 0, true)).offset());
+    expect(editorOffsetAtRowEdge()).toEqual($(getCell(5, 0, true)).offset());
 
     await keyDownUp('enter');
     await keyDownUp('enter');
@@ -235,16 +247,17 @@ describe('TextEditor', () => {
 
     await keyDownUp('enter');
 
-    expect(editor.offset()).toEqual($(getCell(0, 0, true)).offset());
-
-    await selectCell(0, 1);
-    await keyDownUp('enter');
-
-    // Cells that do not touch the edges of the table have an additional left border.
+    // Every cell here sits behind a row header, which owns the gridline on its inline-start
+    // side, so the editor starts 1px before the cell in order to cover it (#6673).
     const editorOffset = () => ({
       top: editor.offset().top,
       left: editor.offset().left + 1,
     });
+
+    expect(editorOffset()).toEqual($(getCell(0, 0, true)).offset());
+
+    await selectCell(0, 1);
+    await keyDownUp('enter');
 
     expect(editorOffset()).toEqual($(getCell(0, 1, true)).offset());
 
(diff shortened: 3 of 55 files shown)


qunabu:
Task linked: [DEV-395 If there are row headers, the first column has a wrong width](https://app.clickup.com/t/9015210959/DEV-395)

demtario:
`visually-approved` because of intended 1px shift

handsontable/src/styles/base/_base.scss
```diff
@@ -180,8 +180,14 @@
...
 
+      // A body cell draws the grid's inline-start frame only when it IS the first cell of its row,
+      // i.e. when no row-header `th` precedes it. With row headers the `th` owns the separator to
+      // column 0 (see the header rules further down), so the cell keeps the plain 0/1 border every
+      // other column has and its content box is the same width as theirs (#6673). `:first-child`,
+      // not `:first-of-type`: the latter also matched the first `td` sitting behind a `th`, which is
+      // what made column 0 carry two borders inside one `col` width.
       &:not(.htFirstDatasetColumnNotRendered) {
-        :where(#{$grid-body-row}) > td:first-of-type {
+        :where(#{$grid-body-row}) > td:first-child {
```
budnix, line 190: With row headers on, column 0 no longer has an inline-start border, so `Border#appear` skips its `+1` correction and the master's selection start edge lands at `rowHeaderWidth - 1`. That pixel sits inside `.ht_clone_inline_start`, which is z-index 120 against the border div's 10, so the left edge of a selection on column 0 looks like it ends up hidden behind the row header. Would be good to open a grid with `rowHeaders: true` and no frozen columns and check that the focus and area borders still show on column 0 - the customBorders specs only count visible elements, so the suite would not catch it.

handsontable/src/styles/base/_base.scss
```diff
@@ -663,33 +672,16 @@
...
-    .ht_master:not(.innerBorderInlineStart):not(.emptyColumns) {
-      :where(table.htCore) > tbody > tr > th,
-      :where(table.htCore) > thead > tr > th:first-child,
-      ~ .handsontable:not(.htGhostTable) :where(table.htCore) > tbody > tr > th,
-      ~ .handsontable:not(.ht_clone_top):not(.htGhostTable) :where(table.htCore) > thead > tr > th:first-child {
-        border-inline-end-width: 0;
-        border-inline-start-width: 1px;
-      }
-    }
+    // Inline-start axis border ownership (#6673). A row header `th` always carries its 1px
```
budnix, line 708: Now that the row header keeps its inline-end border, the seam's color comes from whichever `th` rule wins on sibling position: `--ht-border-color` when the `th` is `:last-child` in its clone, and `--ht-cell-horizontal-border-color` otherwise. Those two differ in the horizon theme, where `--ht-cell-horizontal-border-color` is transparent, so `rowHeaders: true` with no frozen columns now draws a line that develop did not, while the same grid with `fixedColumnsStart: 1` stays transparent. An explicit `border-inline-end-color` on the row-header `th` would make it the same either way. Curious whether the horizon screenshots already show this.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -270,8 +257,10 @@ function runMasterDrawCycle(table: Table, ctx: DrawContext): void {
   if (ctx.positionChanged) {
     if (ctx.performRedraw) {
       // It refreshes the cells borders caused by a 1px shift (introduced by overlays which add or
-      // remove `innerBorderTop` and `innerBorderInlineStart` CSS classes to the DOM element. This happens
-      // when there is a switch between rendering from 0 to N rows/columns and vice versa).
+      // remove the `innerBorderTop` / `innerBorderBottom` CSS classes on the DOM element. This
+      // happens when there is a switch between rendering from 0 to N rows and vice versa). The
+      // inline-start class is still toggled for backward compatibility but no longer shifts the
```
budnix, line 262: Since the class no longer shifts anything, does `InlineStartOverlay#resetFixedPosition` still need to report a position change when it toggles? That result is ORed into `ctx.positionChanged`, so every scroll crossing horizontal offset 0 runs `refreshAll()`, a nested `wot.draw(true)` over the master and every clone, for a class that moves no pixel. Most grids pay it, because `usesLayoutSnapshotForCalculators()` needs uniform row heights and uniform column widths, so a `colWidths` array or `manualColumnResize` drops off the pre-render path. Returning `positionChanged: false` from the inline-start branch of `#computeHeaderBordersState` looks like it would be enough.

tests/fixtures/pages/RowHeaderBorderOwnershipPage.ts
```diff
@@ -0,0 +1,248 @@
...
+  async scrollHorizontallyTo(name: string, testId: string, column: number): Promise<void> {
+    await this.page.evaluate(
+      ([gridName, col]) => (window as unknown as {
+        scrollHorizontallyTo: (name: string, column: number) => void
+      }).scrollHorizontallyTo(gridName as string, col as number),
+      [name, column]
+    );
+
+    await expect.poll(() => this.grid(testId).locator('.ht_master .wtHolder')
+      .evaluate(holder => holder.scrollLeft)).toBeGreaterThan(0);
```
budnix: The doc above says the wait is on the scroll offset leaving zero, but the check is `toBeGreaterThan(0)`, and those two differ in RTL. `setScrollPosition` negates the target, so an RTL holder's `scrollLeft` goes below zero and never above it - `RtlWindowPage` calls out the same thing. Nothing scrolls the `rtl` grid today so this passes, but the first RTL scroll test added here would sit until the poll times out.
```suggestion
    await expect.poll(() => this.grid(testId).locator('.ht_master .wtHolder')
      .evaluate(holder => holder.scrollLeft)).not.toBe(0);
```

handsontable/src/3rdparty/walkontable/test/spec/table/drawCycleHooks.spec.js
```diff
@@ -311,7 +311,7 @@ describe('Table.draw() lifecycle hooks (characterization for the drawCycle refac
     }
   });
 
-  it('should restore the `correctHeaderWidth` flag when the render is skipped', async() => {
+  it('should keep the row header width unchanged across a horizontal scroll whose render is skipped', async() => {
```
budnix: This one can no longer fail for the thing it is named after. The old version asserted that `restoreRenderedStateIfSafe` rolled `correctHeaderWidth` back; the row header width is now scroll-independent by construction, so you could delete the rollback body and this still passes. In a characterization file that reads as an empty check - would it be worth pointing it at the filters or calculators the rollback still restores?

handsontable/src/editors/textEditor/__tests__/textEditor.spec.js
```diff
@@ -140,7 +140,12 @@ describe('TextEditor', () => {
 
     await keyDownUp('F2');
 
-    expect(editor.offset()).toEqual($(getCell(0, 0)).offset());
+    // Every cell here sits behind a row header, which owns the gridline on its inline-start
+    // side, so the editor starts 1px before the cell in order to cover it (#6673).
+    expect({
```
budnix, line 143: This `left + 1` literal and its comment are repeated about thirty times across nine editor spec files. The next geometry change has to touch every copy, and any copy that gets missed keeps asserting the old offset without anyone noticing. `handsontable/test/helpers/themeLayoutFromTokens.js` already exists for this and gained a helper in this same diff, so one shared `editorOffsetBehindRowHeader(editor)` would replace all of them.

handsontable/src/styles/base/_base.scss
demtario: Confirmed, and fixed in a4c544c456. Measured on a develop build against this branch, `main` and `horizon`, selection on (1,0):

| | develop | this branch (before) | now |
|---|---|---|---|
| left edge `style.left` | 50px | 49px | 50px |
| absolute rect | 70..72 | **69..71** | 70..72 |
| inline-start clone right edge | 70 | 70 | 70 |

So exactly as you described: `rowHeaderWidth - 1`, and one of the two border pixels was painted over by the clone at z-index 120.

One nuance worth recording, because it changes what the right fix is. The `-1` position was *geometrically* correct after the ownership change: the gridline really is the row header's now, so straddling it is what every other cell does. It is occlusion that makes it wrong, not arithmetic. So the fix is not a different reading of the border but a second test alongside it - `Border#appear` now shifts the edge when the cell owns an inline-start border **or** its previous sibling is a row-header `th`. Read from the DOM rather than an index, so RTL and `htFirstDatasetColumnNotRendered` come along for free.

You were right that the suite could not catch it, and for the reason you gave. New coverage measures geometry instead of existence: `selectionInlineStartEdgeOffset()` asserts the edge sits at the cell's own boundary (`0`) for a cell behind a row header and still straddles the shared gridline (`-1`) for a cell with another cell beside it, and `selectionEdgeHiddenBehindRowHeader()` asserts no master edge overlaps the overlay. Three tests, covering LTR, RTL and the frozen clone (where both the master's and the clone's copies are checked).

Negative control: all three fail on the previous branch head with `Expected: 0, Received: -1`.

handsontable/src/styles/base/_base.scss
demtario: You were right, on both halves. The horizon screenshots did show it. Fixed in a4c544c456.

Measured in the browser, `horizon`, row-header `th` inline-end color:

| shape | develop @0 | this branch (before) | now |
|---|---|---|---|
| `rowHeaders: true` | transparent | **`rgb(231,231,233)`** | transparent |
| `+ fixedColumnsStart: 1` | transparent | transparent | transparent |
| `columns: []` | `rgb(231,231,233)` | `rgb(231,231,233)` | `rgb(231,231,233)` |

The corner `th` behaved identically, so the header row grew the same line. Tokens confirm the mechanism: `horizon` sets `--ht-cell-horizontal-border-color: var(--ht-colors-transparent)` while `--ht-border-color` is a real color; `main` and `classic` map both to the same value, which is why only one leg of the matrix could ever see this.

Rather than an explicit color on the row header alone, ownership is now stated for both elements that own a seam, since the corner has the same problem in the header row:

```scss
&.htRowHeaders {
  .ht_master:not(.emptyColumns),
  .ht_master:not(.emptyColumns) ~ .handsontable:not(.htGhostTable) {
    :where(#{$grid-head-row}, #{$grid-body-row}) > th:first-child:not(.ht__active_highlight-prev) {
      border-inline-end-color: var(--ht-cell-horizontal-border-color);
    }
  }
}
```

Two carve-outs, both load-bearing:

- **`.emptyColumns`** - with no data columns the row header really *is* the grid's inline-end edge, and develop draws the frame color there (verified: `be: 1px`, `rgb(231,231,233)`). Dropping the carve-out makes it transparent in horizon, which would be a fresh regression.
- **`ht__active_highlight-prev`** - on develop an active column-0 header drew its own inline-start accent, but that border is 0 now, so the accent has to come from the corner's `-prev` tag. Verified it lands: `rgb(46,86,252)` on the corner with column 0 selected.

The target color is whatever develop renders at scroll 0, so the PR's visual delta stays the 1px shift it advertises. Making horizon gain a gridline would be a theme change and needs design sign-off, not a bug-fix PR.

Coverage: one test asserts the seam color is identical across the plain and frozen shapes and equal to a plain body cell's inline-end color (compared element-to-element, since a literal only holds on one theme), and unchanged after a horizontal scroll. A second test pins the `emptyColumns` complement. Both fail on the previous head on `e2e-horizon` (`Expected "rgba(255, 255, 255, 0)", Received "rgb(231, 231, 233)"`) and both pass there on `e2e-main`, which is worth knowing: `e2e-main` is what the local gates run, so this class of defect is only visible on a CI leg.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
demtario: Agreed on the diagnosis: the class moves no pixel now, so the `refreshAll()` it triggers on every scroll crossing offset 0 is pure waste, and your point about `usesLayoutSnapshotForCalculators()` is what makes it common rather than rare - any `colWidths` array or active `manualColumnResize` drops off the pre-render path.

I am deliberately not landing it here. It is slow output, not wrong output, whereas getting it wrong means stale overlays on scroll, and this PR already carries 1587 unreviewed visual diffs - a render-pipeline short-circuit is the last thing that should ride along with those. It is also the same shape as the row-axis work that this PR left alone on purpose (`innerBorderTop` / `innerBorderBottom` / `columnHeaderBorderCompensation` still work the old way), so the two belong together.

Filed as a follow-up in the current sprint, with your `#computeHeaderBordersState` suggestion recorded as the starting point and a note to verify the overlays still refresh where they must.

tests/fixtures/pages/RowHeaderBorderOwnershipPage.ts
demtario: Taken verbatim in a4c544c456. The doc now names the reason rather than leaving the reader to match it against the code:

> It is `not.toBe(0)`, not `toBeGreaterThan(0)`: `setScrollPosition` negates the target in RTL, so an RTL holder's `scrollLeft` goes NEGATIVE and a greater-than poll would sit until it timed out.

Still nothing scrolls the `rtl` grid - the RTL test added for the first thread only selects a cell - so this remains a trap disarmed ahead of use rather than a live bug. Which is the point: it would have cost whoever added the first RTL scroll a timeout with no hint as to why.

handsontable/src/3rdparty/walkontable/test/spec/table/drawCycleHooks.spec.js
demtario: You were right, and I confirmed it by measurement rather than by reading: neutering the column half of `restoreRenderedStateIfSafe` left the spec green. Repointed in ca1c94f143.

Worth recording that my first attempt was **also** vacuous. I pointed it at `expect(columnFilter.offset).toBe(getFirstRenderedColumn())`, which can never fail: `getFirstRenderedColumn()` reads `columnsRenderCalculator.startColumn` and `columnFilter.offset` reads the filter, and the rollback restores both together - so with the rollback neutered they advance together too and still match. The sibling row spec above has the same weakness in its equivalent line; what carries that one is its `getCell` check, and only because a shrunk dataset makes the stale coordinate resolve to `-2`.

`instanceof HTMLElement` is not enough for the scroll case either. With the band advanced past the stale DOM, `sourceToRendered` maps the new index onto rendered index 0 and hands back a real element - the wrong cell, but an element. So the assertion now compares content:

```js
for (let col = wt.wtTable.getFirstRenderedColumn(); col <= wt.wtTable.getLastRenderedColumn(); col++) {
  expect(wt.wtTable.getCell(new Walkontable.CellCoords(0, col)).textContent)
    .toBe(`${getData(0, col)}`);
}
```

With the column rollback neutered that fails `Expected '0' to be 'c'`. The row-header width assertion stays as a secondary check, since it is cheap and still states the scroll-independence the removed flag used to fight, but it is no longer what the spec is named after.

handsontable/src/editors/textEditor/__tests__/textEditor.spec.js
demtario: Done in 7bc20eff65. It was worse than thirty: **86 sites across ten spec files** (`selectEditor` too, and the two `intl*` editors with one each), plus 98 lines of repeated explanation. All of it now goes through one helper, so the sweep is net -69 lines.

```js
expect({
  top: editor.offset().top,
  left: editorInlineStartOffset(editor),
}).toEqual($(getCell(0, 0)).offset());
```

Three deliberate choices in it:

- **It reads the rule off the edited cell's computed border**, the same value `getEditedCellRect` keys on, instead of hardcoding `+ 1`. So no spec restates *which* columns get the shift - which is exactly what went wrong in the editor in the first place, where the offset was keyed on an index while the width was keyed on the border.
- **It landed in `test/helpers/common.js`, not `themeLayoutFromTokens.js`.** That file derives everything from theme tokens; this measures a DOM element, and `getActiveEditor` (which the helper needs to reach the edited `TD`) already lives in `common.js`.
- **The top axis stays with the caller.** `getEditedCellRect` decides that one from the row's position in the overlays (`hasColumnHeaders && renderableRow <= 0 || renderableRow === firstRowIndexOfTheBottomOverlay`), not from a border, so there is no single value to fold in - and the sites that carry `top: editor.offset().top + 1` keep their own comment, which is still accurate.

It throws in RTL rather than answering, since it corrects the physical left edge and that is the inline *start* only in LTR; the RTL specs mirror the offset themselves through their local `offsetForRtl`.

Re-ran everything the sweep touches: core E2E 9848 specs / 0 failures on all three themes, unit 4393/0, ESLint 0 errors.

budnix (dismissed):

handsontable/src/styles/base/_base.scss
```diff
@@ -367,6 +375,25 @@
...
+    // an active column-0 header still gets its inline-start accent from the corner, which is where
+    // that pixel now lives.
+    &.htRowHeaders {
+      .ht_master:not(.emptyColumns),
+      .ht_master:not(.emptyColumns) ~ .handsontable:not(.htGhostTable) {
+        :where(#{$grid-head-row}, #{$grid-body-row}) > th:first-child:not(.ht__active_highlight-prev) {
+          border-inline-end-color: var(--ht-cell-horizontal-border-color);
+        }
+      }
+    }
```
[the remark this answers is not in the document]
demtario: Correct, and fixed in 30243c3535. The specificity analysis holds: the rule was `(0,6,1)` against `th:last-child.ht__active_highlight`'s `(0,3,1)`, so it won and forced the seam to the cell-border token.

Measured in a browser, `horizon`, row 1 selected, row-header `th` inline-end:

| | width | color |
|---|---|---|
| develop, scroll 0 | `0px` | — nothing drawn by the `th` |
| develop, scrolled | `1px` | `rgb(34, 34, 34)` accent |
| this branch, before | `1px` | `rgba(255, 255, 255, 0)` |
| now | `1px` | `rgb(34, 34, 34)` |

One correction to the report, which does not change the verdict. At scroll 0 develop drew the seam from the *cell* side, not the header's: `td` at column 0 carried `border-inline-start: 1px color(srgb 0.83 0.83 0.836)`, the `.highlight.area` selection border. So what disappeared was that line, and only once scrolled did develop show the accent you describe. Either way develop drew something at every offset and this branch drew nothing, so it is a regression.

`ht__active_highlight` now joins the exclusions rather than getting an explicit accent colour, which lets the existing header rules decide. The result is better than parity: the accent is the same colour at every scroll position and in every overlay shape, where develop flipped between the area colour at offset 0 and the accent once scrolled - the border being `0px` there was the whole reason.

Your observation about column 0 keeping its accent while the row header lost it is also what showed the fix is right rather than a suppression. The same inconsistency existed *within* the branch: with two row headers the second one is not `:first-child`, so it was never matched and kept its accent, while the single-row-header shape lost it.

Pinned by a new test that reads the expected value from the accent the column axis already uses - the corner's `ht__active_highlight-prev` edge with column 0 selected - rather than from a literal, and asserts it across all four shapes plus after a horizontal scroll. It fails on the previous commit on the `main` and `horizon` legs; `classic` maps the accent and the cell-border colour to one value, so that leg cannot see it.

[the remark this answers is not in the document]
demtario: Correct, and fixed in 30243c3535. This was the better of the two findings: it is the same defect the rule was written to fix, one column over.

First, the premise checks out - more than one row header is a supported shape, not a curiosity. `afterGetRowHeaderRenderers` is a documented public hook that appends renderers (`tableView.ts:1053`), and `autoRowHeaderSize` measures each of them one by one. Built a grid with a second renderer and got two `th` per body row.

Measured, `horizon`, scroll 0, no selection, body-row row headers in the inline-start clone:

| grid | develop | branch before | now |
|---|---|---|---|
| 2 row headers | `0px`, `0px` | `1px transparent`, **`1px` frame** | `1px transparent`, `1px transparent` |
| 2 row headers + `fixedColumnsStart` | `0px`, `0px` | `1px transparent`, `1px transparent` | unchanged |

So the branch drew a visible frame-coloured line at the seam where develop drew none, and the two shapes disagreed - exactly the divergence the rule exists to remove.

The body half is fixed by dropping `:first-child`: every `th` in a body row *is* a row header, so all of them are matched. The inner ones need no override, because they are never `:last-child` and the base cell rule already gives them this colour - which is why widening the selector cannot overreach.

**The head row is a genuine limitation and I left it.** CSS cannot count how many corner cells precede the first column header, so that half stays keyed on `:first-child`, and with two or more row headers the head-row seam keeps the frame colour. That is not new: develop does the same, because the block this PR deleted only matched `thead > tr > th:first-child` while matching every `tbody > tr > th`. Verified on a develop build - corner clone `th[1]`, `1px rgb(231, 231, 233)`, against `1px transparent` for the same seam in `.ht_clone_top`. Fixing it needs a marker class from the engine on the last row header, which is engine work and a new public class, so it is recorded in AGENTS.md and in the follow-up task rather than improvised here.

Pinned by a new test asserting the last row header owns the seam in both shapes, with the row-header count asserted as its own precondition so the case cannot silently stop being two. It fails on the previous commit on the `horizon` leg.

budnix (approved):

PR #13373, opened 2026-09-03
DEV-430: Span a merged cell over its visible rows when rows are trimmed

### LINKED ISSUE #7686: Merge cells is not working with nested rows in handsontable
We are trying to implement nested rows in handson. Along with the nested rows we also need to merge first column. But when I try to collapse the nested row, the merging is shifting towards the next cells. Below is the js fiddle. To reproduce just click collapse on the first cell. You will see that the merge is messed up. Can you please look into this.
[https://jsfiddle.net/zr6w2u7y/](https://jsfiddle.net/zr6w2u7y/)

beforeCollapsing
afterCollapsing

demtario:
### Context

The PR fixes merged cells drawing on top of the rows below them once rows inside the merge are removed from view.

With `nestedRows` and one merge per group, collapsing a parent left its merge spanning all six of its rows although five had been trimmed away, so it overlapped the next group and the grid rendered two merged cells over the same rows. Expanding the parent again wiped the recovered rows from the lookup matrix, leaving them as empty visible cells beside the restored merge.

The cause is in the trimming support added in #12792. That change re-anchors a merge onto a cached physical top-left on every `rowIndexMapper` `cacheUpdated`, but deliberately keeps `rowspan` unchanged. Trimming compresses the visual row space -- a trimmed row has no visual index at all -- so a merge that keeps its declared span reaches past its own rows and onto whatever sits below. A `nestedRows` collapse is simply the loudest case; a filter hiding one row inside a merge shows the same defect.

The fix makes the merge's **physical rows** the authoritative description of it, and derives the visual `row` / `col` / `rowspan` from that list on every cache update. The merge sits on the **topmost** visual position among the rows of its own that are still visible, the span is how many of them there are, and both are restored in full once the rows come back. The rows are held as an explicit list rather than a start-plus-length range, because merging on a sorted grid (or over a row a filter has hidden) produces a set that is not consecutive -- the previous range model was already inaccurate there.

Six further changes fall out of that, each with a test:

- **Row insert and remove remap the anchors before the shift.** The index mapper emits its cache update *before* `afterCreateRow` / `afterRemoveRow`, so a re-anchor has already run against a grid whose row count changed while the merges had not been shifted yet. Deriving the anchors from the merges there shrank the span twice, which an existing undo/redo spec caught.
- **Removing the last *visible* row of a partly trimmed merge no longer deletes the trimmed rows with it.** `shiftCollections` takes an optional predicate, and a merge left owning no row at all is dropped by the hook itself rather than by the shift -- the shift rules on the merge's visual coordinates, which are stale for a merge purged while all of its rows were trimmed.
- **A column move or freeze no longer truncates a merge.** `translateAfterAxisMove` replaces every merge object and the anchors are keyed on object identity, so the replacements were re-anchored from their post-move visual coordinates -- which, under a filter, describe only the visible part. It now reports which merges replaced which, and the anchors are carried across.
- **A row move under a filter translates every row the merge owns**, not just the visible ones.
- **A row insert decides from the visual order whether it lands inside a merge**, not by comparing physical indexes against the insertion point. A merge's physical rows ascend only while nothing has reordered them.
- **Undo of an unmerge performed on a collapsed group no longer re-creates a one-cell "merge".** A single cell is not a merge, and creating one would shrink a six-row merge to a single row for good. `canMergeRange` now rejects it on the automatic path too; the automatic callers restore areas that were merges when they were recorded, so a single cell reaching them only ever means trimmed geometry.

**Semantics change worth a reviewer's attention, on two axes.** #12792 shipped in 18.1.0 with the documented behavior "merges stay whole while rows are trimmed". This PR replaces that with "a merge spans its visible rows, and is restored whole when they return". Nothing is lost -- the full span lives in the anchor -- but three existing specs asserted the old behavior and were updated, and a fourth was retargeted to filter on a column outside the merge (merging clears the cells it covers, so the old condition values were hiding the merge's own rows rather than the rows above it).

The second axis came out of review. Taking the **topmost** visible row rather than the first one the anchor lists also changes where a merge anchors on a **sorted** grid, which has nothing to do with trimming. On a descending sort a merge over physical rows 2-4 used to anchor on physical row 2 at visual row 6 and span down over visual 6-8, which are physical 2, 9 and 1 -- two rows it does not own. It now anchors at visual row 4 and covers visual 4-6, exactly its own three rows. Three further existing specs asserted the old anchor and were updated. The change fixes the top-left only: a merge whose visible rows are non-contiguous in the visual order still spans one continuous block, so it can still cover foreign rows, and the guide says so.

Not a breaking change under the policy: no default setting value changes, and no public API changes shape. Render-only clipping was considered and rejected -- the matrix overlap and the selection and paste geometry all read the model `rowspan`, so a render-only fix leaves the collection corrupt.

**Known limitation, documented in the code:** when `manualRowMove` splits a merge into fragments while rows are trimmed, a trimmed row cannot be attributed to a fragment, so a split merge is re-anchored from what is visible. It is worth a follow-up.

### How has this been tested?

Full suites, all green:

```bash
npm run test:unit --prefix handsontable          # 4424 passed
npm run test:e2e --prefix handsontable           # 9905 specs, 0 failures
npm run build --prefix handsontable && cd tests && npx playwright test   # 5184 passed, all six theme/bundle legs
./node_modules/.bin/tsc --noEmit -p tsconfig.json  # clean, from handsontable/
```

Targeted:

```bash
npm run test:e2e --prefix handsontable -- --testPathPattern='mergeCells|nestedRows|undoRedo|columnSorting|manualRowMove'
cd tests && npx playwright test merge-cells-nested-rows-collapse
```

Tests added:

- **Playwright** (`tests/e2e/merge-cells-nested-rows-collapse.spec.ts` plus its fixture and page object) -- the reported scenario, asserted on the rendered first column: every group is one merged block, a collapsed group shrinks to its parent row and leaves the group below intact, expanding restores every merge whole, and two groups collapse and expand back in reverse order.
- **Jasmine** (`mergeCells.spec.js`) -- a new `nested rows` describe covering collapse, expand, two parents in reverse order, the undo limitation, and a merge whose rows are all removed being dropped with no merge meta left on any surviving row; plus, in the existing trimming describe, the column-move and column-freeze regressions, the removal of a merge's last visible row, the two row-insert cases, and a merge anchoring on its topmost visible row after an insert reordered the anchor's list. The sorted-grid spec now also asserts the rendered result -- the spanning cell, the two cells it covers, and the ordinary cells either side.
- **Unit** (`cellCoords.unit.ts`, `cellsCollection.unit.ts`) -- `relocate` with and without a new span; `relocateInMatrix` shrinking, growing, and no longer leaving a footprint on the merge below; and `dropMerges` removing by identity, including a merge whose stale coordinates name another merge's slot.

The two column-move tests, the two row-insert tests and the topmost-row test were each confirmed to fail without the fix and pass with it, rather than being assumed to cover it.

Manually verified in a browser against a local build: the issue's own configuration, the equivalent `filters` and `trimRows` cases, and row insert and remove while a group is collapsed. Selection, cell meta, and editing on the collapsed cell all behave.

### Types of changes

### Related issue(s):

1. https://github.com/handsontable/handsontable/issues/7686

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

The merge-cells guide gains a section on what happens when rows inside a merge are removed from view (`filters`, `trimRows`, a `nestedRows` collapse), how that differs from `hiddenRows`, and the undo limitation, plus a paragraph on how a merged cell follows its rows through a sort. The plugin's `AGENTS.md` is updated with the anchor model, the topmost-row rule and the hook-ordering trap.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-430

---

> [!NOTE]
> **Medium Risk**
> Large, behavior-changing changes in the core mergeCells plugin (trimming, sort, moves, undo) affect rendering and the lookup matrix; risk is mitigated by extensive tests but regressions in edge cases remain plausible.
> 
> **Overview**
> Fixes **#7686**: merged cells no longer paint over rows below when **filters**, **`trimRows`**, or a collapsed **`nestedRows`** parent remove rows inside the merge.
> 
> The mergeCells plugin now treats each merge’s **physical row list** (plus column) as authoritative and **re-derives** visual `row` / `col` / **`rowspan`** on every row-index-mapper update: anchor on the first still-visible owned row, span only the visible owned rows, hide from the lookup matrix when none are visible, and **restore the full span** when rows return. **`relocate` / `relocateInMatrix`** accept an updated rowspan; **`dropMerges`** and a **`shiftCollections` `canRemove`** guard partly trimmed merges on row delete; **axis moves** carry anchors across replaced merge objects.
> 
> Related behavior updates include **row insert/remove** remapping anchors before shifts, **sorted-grid** top-left from anchor visual order (not “smallest visual index”), **undo** rejecting one-cell “merges” after unmerge on a collapsed group, plus docs/changelog and broad unit, Jasmine, and Playwright coverage for nested-row collapse.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a3711728baa357da1284dacfe7af510674d73514. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-430 Merge cells is not working with nested rows in handsontable](https://app.clickup.com/t/9015210959/DEV-430)

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -1014,17 +1174,26 @@ export class MergeCells extends BasePlugin {
...
-      // the merge is fully visible).
-      for (let offset = 1; visualColumn !== null && visualRow === null && offset < merge.rowspan; offset++) {
-        visualRow = this.hot.toVisualRow(anchor.physicalRow + offset);
+          if (rowIndex === null) {
+            return;
+          }
+
+          if (visualRow === null) {
+            visualRow = rowIndex;
+          }
```
budnix, line 1262: This picks the first visible row in list order, which is only the real top-left while the anchor list is in visual order. `#remapRowAnchorsAfterInsert` appends the inserted rows at the end of that list, so after an insert inside a partly trimmed merge the order no longer matches. I walked the arithmetic for a merge over physical rows 1-4: trim row 2, insert 2 rows inside, then trim row 1 too, and the merge lands at visual row 3 spanning 4 rows on a grid that only has visual rows 0-5, instead of sitting at 1-4. Taking the smallest visual index makes this independent of list order and also fixes the top-left on a sorted grid, which would let the append-versus-sort note above go away.
```suggestion
          if (visualRow === null || rowIndex < visualRow) {
            visualRow = rowIndex;
          }
```

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -947,27 +984,141 @@ export class MergeCells extends BasePlugin {
...
+   * @param {number[]} physicalRows Physical indexes of the removed rows.
    */
-  #remapRowAnchors(mapPhysicalRow: (physicalRow: number) => number) {
+  #remapRowAnchorsAfterRemove(physicalRows: number[]) {
+    const removed = new Set(physicalRows);
+
+    this.#remapAnchors(anchoredRows => anchoredRows
+      .filter(physicalRow => !removed.has(physicalRow))
+      .map(physicalRow => physicalRow -
+        physicalRows.reduce((shift, removedRow) => shift + (removedRow < physicalRow ? 1 : 0), 0)));
```
budnix: The `reduce` sits inside the `map`, so this walks `physicalRows` once for every row a merge owns. On develop the same `reduce` ran once per merge, and only while trimming was on, so removing a few thousand rows on a grid with many merges is a lot more work now than it was. Sorting the removed rows once and binary searching for how many sit above a row keeps it near-linear, and you are already building `removed` here anyway.

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -1867,21 +2044,31 @@ export class MergeCells extends BasePlugin {
...
 
-    this.#captureMergeAnchors();
+    this.mergedCellsCollection.shiftCollections('up', row, count, (mergedCell) => {
+      if (!remapped) {
+        return true;
+      }
+
+      const anchor = this.#mergeAnchors.get(mergedCell);
+
+      return !anchor || anchor.physicalRows.length === 0;
```
budnix: An empty anchor only lets the shift drop the merge - `shift()` still makes the call from the merge's own visual coords, and for a fully trimmed merge those are whatever it had before it was purged. So the two can disagree and a merge can stay in `mergedCells` owning no rows at all: every later re-anchor purges it again, but a column move still runs it through `translateAfterAxisMove` and puts it back in the matrix at that stale visual row. Would it be cleaner to drop the merge outright once its anchor is empty, instead of routing that decision through the shift? Curious what you think.

handsontable/src/plugins/mergeCells/mergeCells.ts
demtario: You are right, and I took the suggestion as written. `889dd55`.

I re-walked your arithmetic and reproduced it as a spec: merge over physical rows 1-4, trim row 2, insert 2 rows inside, then trim row 1 too. Against the old comparison the merge lands at visual row 3; with the smallest visual index it sits at 1 spanning 4, which is where its rows are. It is in the trimming describe as `should anchor a merge on its topmost visible row after an insert reordered its rows`.

Two things worth flagging, because this reaches further than the insert case.

**It changes the sorted-grid anchor, which three specs on `develop` assert.** On a descending sort a merge over physical 2-4 has its rows at visual 6, 5 and 4. List order anchored it at 6 and spanned down over visual 6-8, which are physical 2, 9 and 1 -- two rows the merge does not own. Topmost anchors at 4 and covers 4-6, exactly its own three rows. So this is better, but it is a second semantics change on top of the trimming one, and it has nothing to do with trimming. I updated the three specs, rescoped the changelog entry, the guide and the plugin's `AGENTS.md`, and added a paragraph to the PR description for it.

**It fixes the top-left, not the span.** A merge whose visible rows are non-contiguous in the visual order (a partial sort, a row move -- never trimming alone) still spans one continuous block downwards from that top-left, so it can still cover rows it does not own. I kept that caveat in the JSDoc rather than letting "also fixes the top-left on a sorted grid" read as "fixes sorted grids".

The sorted-grid spec also had no rendered assertion, only collection state, so I added one: `rowspan="3"` on the topmost cell, `hidden` on the two it covers, plain cells on the rows either side.

The append-versus-sort note is gone, as you expected. The append itself stays -- only its justification was the ordering assumption.

handsontable/src/plugins/mergeCells/mergeCells.ts
demtario: Fixed in `889dd55`.

The removed rows are now sorted once into a local copy, and the shift is a binary search for how many of them sit above a given row:

```ts
const removed = new Set(physicalRows);
const ascending = [...physicalRows].sort((rowA, rowB) => rowA - rowB);

this.#remapAnchors(anchoredRows => anchoredRows
  .filter(physicalRow => !removed.has(physicalRow))
  .map(physicalRow => physicalRow - countBelow(ascending, physicalRow)));
```

A copy rather than the argument itself: `afterRemoveRow` makes no promise about the order it reports the removed rows in, and sorting the caller's array in place would be a side effect on a hook payload.

Kept the remap unconditional. Your "only while trimming was on" observation is about `develop`, where the anchor was a cache that trimming invalidated; the anchors are the authoritative description of a merge now, so that guard cannot come back -- only the per-row cost was worth removing.

handsontable/src/plugins/mergeCells/mergeCells.ts
demtario: Agreed, and done in `889dd55` -- but leading with the negative result, because I could not build the state you describe.

I tried five sequences aimed at the disagreement, all with `nestedRows`, since it is the only plugin that removes trimmed rows at all (a filter or `trimRows` can only ever remove rows you can see, so the anchor cannot empty while any row is trimmed):

1. merge over one group's children, collapse, remove that parent
2. merge over the second group's children, collapse both, remove the first parent then the second
3. merge spanning children of two groups, collapse both, remove both parents
4. shrink the stale span first with removals below it, then remove the owning group
5. collapse the last group, remove the groups above it one at a time

Every one ended with the merge dropped. The reason looks structural rather than lucky: a `nestedRows` removal always takes the whole subtree, which is exactly the set of trimmed rows, so by the time an anchor empties the shift's stale span has been covered as well. If you can see a shape I missed I would like to know -- I may well have stopped one step short.

So I took your suggestion as the simplification you framed it as, not as a fix for a live defect. `#dropMergesWithoutRows()` now drops those merges right after the remap, and the shift's predicate reduces to "only a merge with no anchor at all". They go through a new `MergedCellsCollection#dropMerges()`, which matches by identity: `remove(row, column)`'s coordinate lookup is exactly what cannot find a merge whose coordinates describe nothing, or worse, finds the merge that has since taken that slot. There is a unit test for that case.

One thing a new deletion path invites, so I checked it rather than assuming: no cell meta is stranded. After the drop no surviving row carries `hidden`, `rowspan` or `colspan` -- every row the merge owned was removed from the data, so its meta went with it. That is asserted in the new spec. Worth knowing the same spec passes without `#dropMergesWithoutRows`, since the shift drops the merge too; it pins the meta, not the new branch.

budnix (dismissed):

budnix (approved):

PR #13375, opened 2026-09-04
DEV-2780: Reset the fill handle drag state after a double-click fill-down

### LINKED ISSUE #13370: Double-clicking the fill handle leaves mouseDownOnCellCorner stuck, ghost fill border follows the mouse
### Describe the bug

Double-clicking the fill handle (fill-down) leaves the Autofill plugin in a permanent "dragging" state. After the double-click, a fill ("ghost") border follows whatever cell the mouse hovers, as if the fill handle were still held down. Clicking the fill handle once resets it.

The bug only triggers when the double-click fill actually applies (i.e. selectAdjacent() finds an extent to fill down to). When there is nothing to fill, cleanup runs normally.

Steps to reproduce

1. Open the demo below.
2. Click cell B4.
3. Double-click the fill handle. The column fills down as expected.
4. Move the mouse around the grid without pressing any button.

Actual: a ghost fill border follows the hovered cells; hot.getPlugin('autofill').mouseDownOnCellCorner is stuck on true.

Expected: no fill border after the double-click gesture ends. The flag resets to false.

Demo
https://demos.handsontable.com/?docs=guides%2Fcell-features%2Fautofill-values%2Fjavascript%2Fexample3.js&v=18.1.0

Root cause

On the second click's mouseup, two of the plugin's own handlers race in the same bubble phase:

### Video/Screenshots

Image

_No response_

### Handsontable version

18, 18.1

### Framework version

_No response_

### Your environment

Windows

demtario:
### Context

The PR fixes the Autofill plugin leaving its drag state armed after a double-click on the fill handle (the copy-down gesture). Once the fill-down applied, the fill border kept following the pointer with no button held, and `getPlugin('autofill').mouseDownOnCellCorner` stayed `true` until the next click on the fill handle. Reported in #13370.

Root cause: two `mouseup` listeners run on the same native event. Walkontable synthesizes the corner double-click from its `mouseup` listener on the table holder and fires `afterOnCellCornerDblClick`, whose `fillIn()` zeroes `handleDraggedCells`. The plugin's own `mouseup` listener on `documentElement` runs after it (bubble order) and gated its teardown on `if (this.handleDraggedCells)`, so on an applied fill-down the teardown was skipped and `mouseDownOnCellCorner` never cleared. `#onMouseMove` then redrew the fill border on every pointer move. When the double-click had nothing to fill, `fillIn()` never ran, the counter stayed at 1, and cleanup worked, which is why the bug showed up only when the fill applied.

The fix gates `#onMouseUp` on the gesture flag `mouseDownOnCellCorner` instead of the step counter. The flag is always set whenever the counter is non-zero (both are set together on `afterOnCellCornerMouseDown`), so the drag-fill path is unchanged; the only new path is the double-click case, where the counter is already 0 and `fillIn()` is a no-op on the empty fill highlight.

The plugin's `AGENTS.md` gains a section on the two drag-state fields and the holder-before-document `mouseup` order.

### How has this been tested?

- New Playwright spec `tests/e2e/fill-handle-double-click.spec.ts` with a real-pointer double-click on the fill handle. Case 1 asserts the fill-down applied, then hovers other cells and asserts no visible `.wtBorder.fill` edge and `mouseDownOnCellCorner === false`. Case 2 asserts a normal drag-fill still works after a double-click fill-down. Case 1 failed without the fix (4 fill edges visible) and passes with it; 12/12 across all six theme × bundle legs:
  ```bash
  cd tests && npx playwright test e2e/fill-handle-double-click.spec.ts
  ```
- Legacy Jasmine autofill suite (181 specs, 0 failures):
  ```bash
  npm --prefix handsontable run test:e2e -- --testPathPattern=autofill
  ```
- `npm --prefix handsontable run test:types`, ESLint on the changed files.
- Manual reproduction of the issue steps against the 18.1.0 demo confirmed the ghost border before the fix.

### Types of changes

### Related issue(s):

1. Closes #13370
2. DEV-2780

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2780

---

> [!NOTE]
> **Low Risk**
> Localized Autofill mouseup teardown with documented event-order edge case; drag-fill path unchanged and covered by new e2e tests.
> 
> **Overview**
> Fixes a bug where **double-clicking the fill handle** (copy-down) left the Autofill plugin thinking a drag was still active: the dashed fill border followed the pointer and `mouseDownOnCellCorner` stayed true until the next corner press.
> 
> **`#onMouseUp`** now tears down on the gesture flag **`mouseDownOnCellCorner`** instead of **`handleDraggedCells`**. On a successful copy-down, Walkontable’s holder **`mouseup`** runs before the plugin’s document listener and **`fillIn()`** already resets the step counter, so the old guard skipped cleanup. Drag-fill behavior is unchanged: **`fillIn()`** still runs only when **`handleDraggedCells > 1`**.
> 
> Adds contributor notes in **`AGENTS.md`**, a changelog entry, Playwright coverage with real-pointer double-click helpers on **`SelectionFeaturesPage`**, and a JSDoc type fix for **`handleDraggedCells`** (`number`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bd36b8ecf8a322ba1107e29afbaa5dcd5ae166c6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2780 [GH-13370] Fix Autofill: double-click fill handle leaves mouseDownOnCellCorner stuck](https://app.clickup.com/t/9015210959/DEV-2780)

handsontable/src/plugins/autofill/AGENTS.md
```diff
@@ -44,6 +44,20 @@ last pointer position with `countDragStep = false`. That is the other half of th
 shifts a new cell under a resting pointer has to redraw the border without registering a drag that never
 happened.
 
+## Two drag-state fields, one sentinel (GitHub #13370)
+
+`mouseDownOnCellCorner` is the **gesture flag** (set on `afterOnCellCornerMouseDown`, cleared in
+`#onMouseUp`), `handleDraggedCells` is the **step counter**. Only the flag says whether a corner gesture
```
budnix, line 50: While you are writing down the flag versus counter split, the JSDoc on `handleDraggedCells` in `autofill.ts:115` still says `@type {boolean}` even though the field holds a number. That tag is the exact mix-up this PR fixes, so someone who reads the field instead of this file gets the wrong idea. Worth a quick one-word fix.

handsontable/src/plugins/autofill/AGENTS.md
```diff
@@ -44,6 +44,20 @@ last pointer position with `countDragStep = false`. That is the other half of th
...
+is in progress – `fillIn()` legitimately zeroes the counter through `resetSelectionOfDraggedArea()`.
+
+That matters because the corner double-click is not a native `dblclick`: Walkontable synthesizes it
+from the mousedown/mouseup pairs and fires it from its `mouseup` listener on the **holder**, which runs
+before this plugin's `mouseup` listener on the `documentElement` (bubble order). On a copy-down that
+applies, `fillIn()` has therefore already zeroed the counter when `#onMouseUp` runs. Gating the teardown
+on the counter skipped it and left `mouseDownOnCellCorner` stuck, so `#onMouseMove` kept redrawing the
+fill border under a released pointer. Gate any teardown on the flag, never on the counter.
```
budnix, line 58: Can I ask — should this rule cover `disablePlugin()` as well? It only calls `super.disablePlugin()`, and the base class clears the event manager, so `updateSettings({ fillHandle: false })` between the corner mousedown and the mouseup removes the `documentElement` mouseup listener while `mouseDownOnCellCorner` is still true. Switching the fill handle back on re-adds the mousemove listener, and the border follows a released pointer again, which is the same symptom as #13370. `updatePlugin()` is safe because it re-adds the listeners before the mouseup, so this looks like a narrow follow-up rather than something to fix here.

handsontable/src/plugins/autofill/autofill.ts
```diff
@@ -873,7 +873,12 @@ export class Autofill extends BasePlugin {
    * On mouse up listener.
    */
   #onMouseUp() {
-    if (this.handleDraggedCells) {
+    // Gate on the gesture flag, not on the drag-step counter. The corner double-click is
+    // synthesized by Walkontable from the holder's `mouseup`, which runs before this
+    // `documentElement` listener, and its `fillIn()` has already zeroed `handleDraggedCells`
+    // by then – a counter-based guard would skip the teardown and leave the fill border
+    // following the pointer (GitHub #13370).
```
budnix: The comment repeats the new AGENTS.md section almost word for word, and two copies of the same explanation tend to drift apart. Since the section already carries the full story, one line here plus a pointer would keep them in sync. Feel free to skip this if you would rather have the reason sitting right at the guard.
```suggestion
    // Gate on the gesture flag, not on the drag-step counter. See this plugin's AGENTS.md
    // ("Two drag-state fields, one sentinel") for why the counter is already 0 here (GitHub #13370).
```

handsontable/src/plugins/autofill/AGENTS.md
demtario: Good catch, and you are right that it is the same mix-up. Fixed in 0fc6322: the tag on `handleDraggedCells` now reads `@type {number}`.

handsontable/src/plugins/autofill/autofill.ts
demtario: Agreed, one source is better than two that drift. Applied your suggestion verbatim in 0fc6322.

handsontable/src/plugins/autofill/AGENTS.md
demtario: Checked it, and the path is real. `disablePlugin()` only delegates to the base class, the base tears down the event manager, and nothing resets the drag state, so a `fillHandle: false` update landing between the corner mousedown and the mouseup strands `mouseDownOnCellCorner` at `true`. Re-enabling re-adds the mousemove listener and the border follows a released pointer again. Your read on `updatePlugin()` matches mine: it re-adds the listeners synchronously, so the mouseup still finds them.

Taking your recommendation and keeping it out of this PR. It needs its own spec and it touches a lifecycle path this change does not, so I have created a follow-up task in the current sprint for it. Left it unlinked here on purpose.

The fix there should also cover `addingStarted` and `mouseDragOutside`, which this plugin's AGENTS.md already says belong in the teardown and which no path currently clears.

budnix (approved):

PR #13376, opened 2026-09-04
DEV-2781: Fix useHotEditor setValue not updating the cell

### LINKED ISSUE #13374: useHotEditor setValue not work
### Describe the bug

useHotEditor setValue not work

### Video/Screenshots

_No response_

https://demos.handsontable.com/?docs=guides%2Fcell-functions%2Fcell-editor%2Freact%2Fexample1.tsx&v=18.1.0

### Handsontable version

18.1

### Framework version

_No response_

### Your environment

chorme

demtario:
### Context

The PR fixes React custom editors not committing `setValue` after a click on the editor UI (GH-13374 / DEV-2781).

In 18.1, PR #13194 started treating a document `mouseup` outside the grid as a deselect. React editors still portaled to `document.body`, which core `#isPathWithinGrid` does not treat as inside the grid. A click on a `useHotEditor` control therefore closed the editor with the original value, and the following `setValue` + `finishEditing` ran on a VIRGIN editor and did nothing. The official cell-editor React demo is the reproduction.

The wrapper now mounts a stable editor host in `hot.rootPortalElement` (the same contract native portal UI already uses) and returns `useHotEditor`'s `value` immediately instead of through `useDeferredValue`.

- Unit tests added/modified (`*.unit.js`): none — covered by `wrappers/react-wrapper/test/hotTable.spec.tsx` and `wrappers/react-wrapper/test/hotColumn.spec.tsx`
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`): none — the functional Playwright suite has no React runtime; the regression is covered in the React wrapper Jest suite with a full pointer sequence (`mousedown` / `mouseup` / `click`)
- For a bug fix — the spec that fails without this fix: `should commit setValue after a real pointer sequence on the editor control` in `wrappers/react-wrapper/test/hotTable.spec.tsx` (cell stayed `A1`); also `should mount the editor component inside the Handsontable root portal` and the matching HotColumn portal spec
- Demo page / recorded trace (for UI changes): official repro is the cell-editor React demo at v18.1.0

### Commands run

```
npm --prefix wrappers/react-wrapper run test
```

Result: React wrapper Jest suite 102/102 passed (including the three new cases that failed on the unfixed code).

### Types of changes

### Related issue(s):

1. #13374
2. DEV-2781

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

- [ ] MANUAL QA NEEDED — 

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2781

---

> [!NOTE]
> **Medium Risk**
> Changes where React editor DOM mounts and how `useHotEditor` value syncs, which directly affects whether cell edits commit on pointer interaction; scope is limited to `@handsontable/react-wrapper` with targeted tests.
> 
> **Overview**
> Fixes **React custom editors not saving** when users click controls in the editor UI (GH-13374). After 18.1’s outside-click deselect, editors still portaled to `document.body` were treated as outside the grid, so the editor closed before `setValue` / `finishEditing` could apply.
> 
> The wrapper now keeps a stable **`.hot-wrapper-editor-portal-host`** and portals component editors there, then **moves that host into `hot.rootPortalElement`** after init so core sees editor clicks as inside the grid. `createEditorPortal` requires that host (no `document.body` fallback); `HotColumn` gets `getEditorPortalHost` via context. **`useHotEditor`** drops `useDeferredValue` so **`value` updates in the same render as `setValue`**.
> 
> Docs, changelogs, `AGENTS.md`, and Jest specs cover portal placement, pointer-commit regression, and immediate `value` display.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 99bcf2cf8918f9c3da58a190d71fe208849ca00a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2781 [GH-13374] Fix useHotEditor setValue not updating the edited cell (React wrapper)](https://app.clickup.com/t/9015210959/DEV-2781)

wrappers/react-wrapper/src/hotTableInner.tsx
```diff
@@ -57,6 +58,12 @@ const HotTableInner = forwardRef<
    */
   const globalEditorClassInstance = useRef<Handsontable.editors.BaseEditor | null>(null);
 
+  /**
+   * Stable host for React editor portals. Appended to `rootPortalElement` after init
+   * so core treats editor UI as inside the grid (see `#isPathWithinGrid`).
```
budnix, line 64: This holds for the path checks, but there is a second gate it does not cover. `isOutsideInput` is `isInput(el) && !('hotInput' in el.dataset)`, so a focused plain `<input>` inside a React editor still makes every document mouseup call `hot.unlisten()`, no matter where the editor sits. That was already true before this PR, but both new specs use a `<button>`, where `isInput` is false, so nothing here would catch it. Curious whether stamping `data-hot-input` on inputs inside the host belongs in this PR or in its own ticket.

wrappers/react-wrapper/src/hotTableInner.tsx
```diff
@@ -108,6 +115,28 @@ const HotTableInner = forwardRef<
     return null;
   }, [hotElementRef]);
 
+  /**
+   * Get or create the stable editor portal host.
+   *
+   * @returns {HTMLElement | null} The host element, or `null` before a document is available.
+   */
+  const getEditorPortalHost = useCallback((): HTMLElement | null => {
```
budnix, line 124: The host is created here during render, but it only gets attached in the mount effect further down, and React runs child effects before parent ones. So an editor component's own mount effects run while its DOM is still detached: `focus()` does nothing, `getBoundingClientRect()` returns zeros, and `getComputedStyle()` comes back empty. On `document.body` all three worked, so an editor that autofocuses or measures itself on mount quietly loses that. Appending the host to `doc.body` right here and then moving it into `rootPortalElement` in the effect would keep the old behavior.

wrappers/react-wrapper/src/hotTableInner.tsx
```diff
@@ -108,6 +115,28 @@ const HotTableInner = forwardRef<
...
+    const doc = getOwnerDocument();
+
+    if (!doc) {
+      return null;
+    }
+
+    if (!editorPortalHostRef.current) {
+      const host = doc.createElement('div');
+
+      host.className = DEFAULT_CLASSNAME;
```
budnix: This gives the host the same class `createEditorPortal` already puts on the portal's own wrapper, so `hot-wrapper-editor-container` now names two nested divs. Anyone calling `querySelector('.hot-wrapper-editor-container')` gets the host instead of the editor, and any page CSS on that class applies twice. A `position: relative` in that CSS is the bad case, since the editor positions itself with page coordinates and would land in the wrong spot. A separate class for the host keeps the two apart.

wrappers/react-wrapper/src/hotTableInner.tsx
```diff
@@ -203,6 +232,13 @@ const HotTableInner = forwardRef<
 
     __hotInstance.current.init();
 
+    const portalHost = editorPortalHostRef.current;
+    const rootPortalElement = __hotInstance.current.rootPortalElement;
+
+    if (portalHost && rootPortalElement && portalHost.parentNode !== rootPortalElement) {
+      rootPortalElement.appendChild(portalHost);
```
budnix: Moving the host in here also means editors inherit whatever the theme sets on `rootPortalElement`, and the theme classes carry `color-scheme`. That property is inherited, so a `useHotEditor` editor with an `<input>`, `<select>` or checkbox renders those controls in the grid's scheme instead of the host page's. It is arguably the better default, but it changes how existing React editors look and neither the changelog nor the docs mention it. Worth a line somewhere so people are not surprised.

wrappers/react-wrapper/src/helpers.tsx
```diff
@@ -205,7 +210,7 @@ export function createEditorPortal(doc: Document | null, Editor: HotTableProps['
     <div {...containerProps}>
       {editorElement}
     </div>
-    , doc.body);
+    , portalHost ?? doc.body);
```
budnix: This fallback cannot actually run. Both call sites pass `getOwnerDocument()` as `doc` and `getEditorPortalHost()` as `portalHost`, and the host is null only when the document is null, which the guard above already returned on. The `@param` line says it falls back to `doc.body` before the table mounts, and that is the exact state this PR fixes, so leaving it in invites someone to bring the bug back. The spot that does need a fallback is the append block in `hotTableInner`, where a missing `rootPortalElement` leaves the host detached and the editor never shows up, with no warning.

wrappers/react-wrapper/src/hotEditor.tsx
```diff
@@ -261,7 +257,7 @@ export function useHotEditor<T>(
   return useMemo(
     () => ({
       get value(): T | undefined {
-        return deferredValue;
+        return editorValue;
```
budnix, line 260: Can I ask what this half of the fix is for? The commit path never reads React state - `setValue` writes straight to the editor instance and `finishEditing` reads it back with `getValue()` - so the deferred value could not have caused the reported miss. Dropping it does change when `useHotEditor().value` updates for every consumer, and the comment you removed said the deferral was there for performance. If it stays, a changelog line or a doc note would help, since it is a timing change in a public hook.

wrappers/react-wrapper/test/hotTable.spec.tsx
```diff
@@ -518,7 +521,63 @@ describe('Editor configuration using React components', () => {
...
+    expect(hotInstance.getDataAtCell(0, 0)).toEqual('A1');
+
+    await act(async () => {
+      simulatePointerActivation(document.querySelector('#pointerCommitEditor button'));
+    });
+
+    expect(hotInstance.getDataAtCell(0, 0)).toEqual('new-value');
+  });
+
+  it('should update the hook value in the same turn as setValue', async () => {
```
budnix, line 560: I do not think this one goes red without the `useDeferredValue` change. `act()` drains the transition lane that `useDeferredValue` schedules its catch-up render on, so the span already reads `typed` on the old code - I checked with a small React 18.3.1 plus jsdom script and it printed `typed` either way. The pointer spec does not cover it either, since `setValue` writes straight to the native editor. Would be good to confirm by putting the two `useDeferredValue` lines back and running just this spec.

wrappers/react-wrapper/AGENTS.md
```diff
@@ -35,5 +35,6 @@
...
 | Trusting a green `npm run build` for types, or `npx eslint src/*.tsx` | The type build hides `tsc` errors — inspect the emitted `types.d.ts` after any `src/` type change. This package has no lint script/config, so direct `npx eslint` misparses TS/JSX; use the monorepo lint. |
+| Mounting a React editor on `document.body` | Core `#isPathWithinGrid` treats only `rootElement` / `rootWrapperElement` / `rootPortalElement` as inside the grid. After the 18.1 document-mouseup deselect, a body-mounted editor is a foreign click: `setValue` + `finishEditing` run on a closed editor and the cell does not change. Always portal into `hot.rootPortalElement` (the stable host in `HotTableInner`). Keep `mousedown` `stopPropagation` on the editor root as defense in depth. |
```
budnix: `#isPathWithinGrid` does not look at `rootWrapperElement` - it checks `rootElement` and `rootPortalElement`, then falls back to the shadow host chain. `rootWrapperElement` shows up in a different expression, `isPathThroughGridUi`, which feeds the focus check rather than the deselect gate, and the document `mousedown` handler uses `#isPathWithinGrid` on its own. Since the point of this row is to stop someone repeating the bug, naming a third element that does not count could send them the wrong way.
```suggestion
| Mounting a React editor on `document.body` | Core `#isPathWithinGrid` treats only `rootElement` / `rootPortalElement` (plus the grid's shadow hosts) as inside the grid. After the 18.1 document-mouseup deselect, a body-mounted editor is a foreign click: `setValue` + `finishEditing` run on a closed editor and the cell does not change. Always portal into `hot.rootPortalElement` (the stable host in `HotTableInner`). Keep `mousedown` `stopPropagation` on the editor root as defense in depth. |
```

wrappers/react-wrapper/src/hotTableInner.tsx
demtario: Tracked as a follow-up. Left out of this PR because it predates the portal change and the new specs use a button.

wrappers/react-wrapper/src/hotTableInner.tsx
demtario: The host now appends to `document.body` when it is created, then moves into `rootPortalElement` after init. After StrictMode cleanup it reconnects to `document.body` if it is detached.

wrappers/react-wrapper/src/hotTableInner.tsx
demtario: The host now uses `hot-wrapper-editor-portal-host`. The portal wrapper keeps `hot-wrapper-editor-container`. The portal specs assert the host is not the editor class.

wrappers/react-wrapper/src/hotTableInner.tsx
demtario: Documented in the cell-editor guide and in the `changed` changelog entry for this PR.

wrappers/react-wrapper/src/helpers.tsx
demtario: `createEditorPortal` now returns null without a host. If `rootPortalElement` is missing, the append block warns and leaves the host on `document.body`.

wrappers/react-wrapper/src/hotEditor.tsx
demtario: Kept. The commit path is still the portal move. `value` now updates in the same render as `setValue`; that is documented in the cell-editor guide, custom-cells, and the `changed` changelog entry.

wrappers/react-wrapper/test/hotTable.spec.tsx
demtario: Confirmed: this spec stayed green with `useDeferredValue` still in place. The it-block now states it documents the public hook contract, not the GH-13374 commit path.

wrappers/react-wrapper/AGENTS.md
demtario: Applied the suggested wording.

budnix (approved):

PR #13377, opened 2026-09-04
DEV-2783: Keep Emitted Types from canceling CI Gate

demtario:
### Context

The PR fixes the Emitted Types job timing out at 15 minutes and reporting as `cancelled`, which CI Gate treats as a failure and which blocked #13373 and #13375.

The job itself was fine: artifacts downloaded, `tsc@5.1.6` and `publint` passed. GitHub then killed it because `attw --pack .` walks every exports path (about 200 plugins, languages, themes, and `dist/` files) × several resolution modes — about 8 minutes with no output until the table prints — and the three `npx -y` tool downloads have no cache. On a slow npm day, `tsc` and `publint` alone took 4–7 minutes each and left attw no time. A timeout is `cancelled`, not `failure`, so the check name is "Cancelled after 15m".

The same job on #13373 succeeded the day before in 8m 36s. This is a budget problem, not a type error in those PRs.

Three changes, all in `emitted-types.yml`:

1. Raise `timeout-minutes` from 15 to 30.
2. Install `typescript@5.1.6`, `publint@0.3.21`, and `@arethetypeswrong/cli@0.18.3` once into `$RUNNER_TEMP` and cache `~/.npm`. The check steps then call the local binaries.
3. Limit attw to the typed public roots (`.`, `base`, `registry`, `settings`). `tsc` still type-checks every emitted `.d.ts`. `publint` still lints the full exports map.

[skip changelog]

### How has this been tested?

- New pin tests in `.github/scripts/__tests__/emitted-types.test.mjs` (timeout ≥ 30, one cached install, attw `--entrypoints . base registry settings`).
- Full GitHub tooling suite: `node --test .github/scripts/__tests__/*.test.mjs` — 112 passed, 0 failed.

- Unit tests added/modified (`*.unit.js`): none — CI/tooling only
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`): none
- Type tests (`*.types.ts`) updated if public API changed: none
- For a bug fix — the spec that fails without this fix: `.github/scripts/__tests__/emitted-types.test.mjs`
- Demo page / recorded trace (for UI changes): none

### Commands run

```
node --test .github/scripts/__tests__/*.test.mjs
# 112 passed, 0 failed
```

### Types of changes

### Related issue(s):

1. DEV-2783
2. Blocks #13373
3. Blocks #13375

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2783

---

> [!NOTE]
> **Low Risk**
> CI/workflow-only changes with no product runtime impact; PR attw scope is narrower but full checks remain on trunk via `attw-full`.
> 
> **Overview**
> Fixes **Emitted Types** hitting the 15-minute limit (reported as `cancelled`, which fails CI Gate) by giving the job more time and cutting cold-start work.
> 
> The reusable **`emitted-types.yml`** workflow now uses a **30-minute** timeout, installs **typescript**, **publint**, and **attw** once into `$RUNNER_TEMP` with a cached **`~/.npm`** install (no per-step **`npx -y`**), and runs **attw** via **`TOOLS_BIN`**. On PRs, **attw** is limited to typed public entrypoints (`.`, `base`, `registry`, `settings`, `themes`); **`attw-full`** (wired from **`run-all`** in **`test.yml`** and **`develop.yml`**) restores the full export walk on trunk.
> 
> **`checks.yml`** adds **`emitted-types.yml`** to the **`verify-types`** path filter so workflow-only PRs still run the job. New **`emitted-types.test.mjs`** locks in these constraints (timeout floor, install/cache shape, reusable-workflow env rules, attw scope).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2cf6d068396ef899301dd6c4427e5e15f86a9736. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/scripts/__tests__/emitted-types.test.mjs
+++ b/.github/scripts/__tests__/emitted-types.test.mjs
@@ -7,66 +7,172 @@ import { repoRoot } from '../lib/repo-root.mjs';
 // Emitted Types is on CI Gate's `needs` list. GitHub reports a job timeout as
 // `cancelled`, and the gate treats that as a failure, so a 15-minute budget
 // that attw regularly exceeded blocked otherwise-green PRs (#13373, #13375).
-// The workflow comments record the three constraints this file pins: the
-// timeout floor, a single cached install instead of three `npx -y` downloads,
-// and attw limited to the typed public roots. A revert of any of those puts
-// the race back.
+// The workflow comments record the constraints this file pins: the timeout
+// floor, a single cached install instead of three `npx -y` downloads, attw
+// limited to the typed public roots on PRs, and no reusable-workflow contexts
+// in job-level env. A revert of any of those puts the race — or a 0-job
+// parse failure — back.
 
-const WORKFLOW = readFileSync(
-  path.join(repoRoot(), '.github/workflows/emitted-types.yml'),
-  'utf8'
-);
-const withoutComments = WORKFLOW.replace(/#[^\n]*/g, '');
+const root = repoRoot();
+const read = (rel) => readFileSync(path.join(root, rel), 'utf8');
+const WORKFLOW = read(path.join('.github', 'workflows', 'emitted-types.yml'));
+const CHECKS = read(path.join('.github', 'workflows', 'checks.yml'));
+const TEST_YML = read(path.join('.github', 'workflows', 'test.yml'));
+const DEVELOP_YML = read(path.join('.github', 'workflows', 'develop.yml'));
+const withoutComments = (source) => source.replace(/#[^\n]*/g, '');
+const workflow = withoutComments(WORKFLOW);
+
+/**
+ * Return the first `@version` pin for `name` in the comment-stripped workflow.
+ *
+ * @param {string} name Package name as it appears before `@` (escaped for RegExp).
+ * @returns {string} The pinned version.
+ */
+function pin(name) {
+  const match = workflow.match(new RegExp(`${name}@([\\d.]+)`));
+
+  assert.ok(match, `missing ${name}@<version> pin in emitted-types.yml`);
+
+  return match[1];
+}
+
+/**
+ * Escape a semver string for use inside a RegExp.
+ *
+ * @param {string} value A version like `5.1.6`.
+ * @returns {string} The escaped pattern.
+ */
+function escapeDots(value) {
+  return value.replace(/\./g, '\\.');
+}
 
 test('the Emitted Types job budget is at least 30 minutes', () => {
-  const timeout = Number(/timeout-minutes:\s*(\d+)/.exec(WORKFLOW)?.[1]);
+  const timeout = Number(/\n    timeout-minutes:\s*(\d+)/.exec(workflow)?.[1]);
 
   assert.ok(
-    timeout >= 30,
-    `timeout-minutes (${timeout}) must be >= 30 — 15 minutes cancelled on a cold npx + full-export attw`
+    Number.isFinite(timeout) && timeout >= 30,
+    `job-level timeout-minutes (${timeout}) must be >= 30 — 15 minutes cancelled on a cold npx + full-export attw`
   );
 });
 
 test('type-check tools are installed once, not via three cold npx -y calls', () => {
+  const ts = pin('typescript');
+  const publint = pin('publint');
+  const attw = pin('@arethetypeswrong\\/cli');
+
   assert.match(
-    WORKFLOW,
+    workflow,
     /name: Install type-check tools/,
     'expected a single Install type-check tools step'
   );
   assert.match(
-    WORKFLOW,
-    /actions\/cache@/,
-    'expected an npm cache so the install is not a cold download every run'
+    workflow,
+    /path:\s*~\/\.npm/,
+    'expected the npm cache path so the install is not a cold download every run'
   );
+  assert.match(
+    workflow,
+    new RegExp(
+      `key:\\s*emitted-types-npx-\\$\\{\\{\\s*runner\\.os\\s*\\}\\}-ts${escapeDots(ts)}-publint${escapeDots(publint)}-attw${escapeDots(attw)}`
+    ),
+    'cache key must name each installed version — a stale key is a permanent miss'
+  );
+  assert.match(
+    workflow,
+    /restore-keys:\s*emitted-types-npx-\$\{\{\s*runner\.os\s*\}\}-/,
+    'expected a prefix restore-key so a one-tool bump still hits the other tarballs'
+  );
+  assert.match(
+    workflow,
+    /npm install --no-package-lock --prefer-offline --no-audit --no-fund/,
+    'pinned install must not revalidate metadata or audit on a warm ~/.npm'
+  );
+  assert.match(
+    workflow,
+    /tsc" --version \| grep -F 'Version 5\.1\.6'/,
+    'install step must assert the root tsc is the pinned 5.1.x, not attw\'s nested compiler'
+  );
+  assert.match(
+    workflow,
+    /TOOLS_BIN=/,
+    'call tsc/publint/attw via TOOLS_BIN — do not prepend the whole .bin to GITHUB_PATH'
+  );
+  assert.doesNotMatch(
+    workflow,
+    /GITHUB_PATH/,
+    'do not put the tool .bin on GITHUB_PATH — it also fronts semver/marked/highlight'
+  );
+  assert.match(workflow, /"\$TOOLS_BIN\/tsc"/);
+  assert.match(workflow, /"\$TOOLS_BIN\/publint"/);
+  assert.match(workflow, /"\$TOOLS_BIN\/attw"/);
   assert.equal(
-    (withoutComments.match(/npx\s+-y/g) || []).length,
+    (workflow.match(/npx\s+-y/g) || []).length,
     0,
     'do not bring back per-step `npx -y` — each one is an uncached registry fetch'
   );
 });
 
-test('job-level env does not use the runner context', () => {
-  // Reusable workflows reject ${{ runner.* }} in jobs.<id>.env. The caller
-  // then fails at parse time with 0 jobs, so CI Gate never reports
-  // (DEV-2783, runs 33853301602 / 33853913419).
-  const jobEnv = withoutComments.match(/\n    env:\n([\s\S]*?)\n    steps:/)?.[1] ?? '';
+test('job-level env does not use contexts GitHub rejects on a reusable workflow', () => {
+  // Reusable workflows reject ${{ runner.* }} (and env/steps/job) in
+  // jobs.<id>.env. The caller then fails at parse time with 0 jobs, so CI
+  // Gate never reports (DEV-2783, runs 33853301602 / 33853913419). The
+  // block must sit above `steps:` so this regex can see it — moving it
+  // below is valid YAML and the same parse failure, so fail closed.
+  const jobEnv = workflow.match(/\n    env:\n([\s\S]*?)\n    steps:/)?.[1];
 
+  assert.ok(
+    jobEnv != null && jobEnv.trim() !== '',
+    'job-level env must sit above steps: — a reusable workflow rejects runner/env/steps/job there'
+  );
   assert.doesNotMatch(
     jobEnv,
-    /\$\{\{\s*runner\./,
-    'do not use ${{ runner.* }} in job-level env of a reusable workflow'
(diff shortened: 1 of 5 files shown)


qunabu:
Task linked: [DEV-2783 Give Emitted Types enough time so attw does not cancel CI Gate](https://app.clickup.com/t/9015210959/DEV-2783)

.github/workflows/emitted-types.yml
```diff
@@ -70,16 +102,23 @@ jobs:
...
+          # attw walks every exports path × several resolution modes. The
+          # package publishes ~200 of those (every plugin, language file,
+          # dist/ asset). That analysis alone took ~8 minutes and, with a
+          # cold npx, blew the old 15-minute timeout. tsc already type-checks
+          # every emitted .d.ts and publint already lints the full exports
+          # map, so attw only needs the typed public roots — the dual-package
+          # pair plus settings and themes. false-export-default is Babel CJS
+          # interop.
+          attw --pack . --format table \
+            --entrypoints . base registry settings themes \
```
budnix: Can I ask how you picked these five? Running attw on the published 18.1.0 finds 275 entrypoints, and 113 of them are typed JS subpaths like ./plugins/*, ./renderers/* and ./cellTypes/* - those lose their type-resolution check here. The comment above says tsc and publint already cover them, but tsc only globs `tmp/**/*.d.ts` so it never opens the .d.mts wrappers, and its `moduleResolution: node` never reads the exports map, so neither tool sees a subpath whose import condition points at a CJS-shaped declaration. Since the 30-minute budget and the cached install in this same PR already leave room for the full run, would it work to keep the five roots on PRs and let the develop push do full auto-discovery?

.github/workflows/emitted-types.yml
```diff
@@ -15,9 +15,20 @@ jobs:
...
     runs-on: ubuntu-latest
-    timeout-minutes: 15
+    # 15 minutes was lost on #13373 / #13375: GitHub reports a job timeout as
+    # `cancelled`, and CI Gate treats that as a failure. attw on every export
+    # path took ~8 minutes even when `npx -y` was fast; a cold npm day then
+    # spent another 4–7 minutes per tool. 30 minutes covers a slow download
+    # plus the analysis, and the non-prebuilt fallback (pnpm install + a full
+    # local `npm run build`) that runs when Build is scope-skipped. Do not
+    # lower this without re-timing a cold run of that fallback too.
+    timeout-minutes: 30
```
budnix, line 29: Worth a quick check - this job does not run on this PR. `verify-types` in checks.yml filters on `./handsontable/**` only and `run-all` is false on a PR ref, so `gh pr checks 13377` reports Emitted Types as skipping. That means the new install step, the cache, the entrypoint list and the two dropped ignore rules all land on develop untested. Adding `./.github/workflows/emitted-types.yml` to that filter would run it here, and it would also time the non-prebuilt fallback the comment above says was never timed.

.github/workflows/emitted-types.yml
```diff
@@ -53,6 +64,27 @@ jobs:
...
+      # Three sequential `npx -y` downloads have no cache (setup-node caches
+      # pnpm, not ~/.npm). On a slow npm day each one took several minutes and
+      # left attw no time inside the old 15-minute budget. One install into
+      # $RUNNER_TEMP, plus an npm cache keyed on the three pins, is one
+      # download on a miss and a local reuse after that.
+      - name: Cache npm for the type-check tools
+        uses: actions/cache@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # https://github.com/actions/cache/releases/tag/v5.0.3
+        with:
+          path: ~/.npm
+          key: emitted-types-npx-${{ runner.os }}-ts5.1.6-publint0.3.21-attw0.18.3
```
budnix, line 80: This key is exact with no `restore-keys`, so bumping any one of the three tools is a full cold miss and all three tarballs download again on exactly the slow-registry day the cache is here for. `setup-workspace/action.yml` pairs its exact key with a prefix fallback for the same reason.
```suggestion
          key: emitted-types-npx-${{ runner.os }}-ts5.1.6-publint0.3.21-attw0.18.3
          restore-keys: emitted-types-npx-${{ runner.os }}-
```

.github/workflows/emitted-types.yml
```diff
@@ -53,6 +64,27 @@ jobs:
...
+      - name: Install type-check tools
+        run: |
+          tools="$RUNNER_TEMP/emitted-types-tools"
+          mkdir -p "$tools"
+          cd "$tools"
+          npm init -y >/dev/null
+          npm install --no-package-lock \
```
budnix: A warm `~/.npm` still lets npm revalidate package metadata against the registry, and the quick audit request fires too - both happen before any cached tarball is read. On the slow-registry day this step is meant to survive, that gives back part of what the cache buys. The pins are exact, so skipping those is safe.
```suggestion
          npm install --no-package-lock --prefer-offline --no-audit --no-fund \
```

.github/workflows/emitted-types.yml
```diff
@@ -53,6 +64,27 @@ jobs:
...
+            typescript@5.1.6 \
```
budnix, line 89: The old form named the compiler at the call site (`npx -y -p typescript@5.1.6 tsc`) so it could not drift. Now the version lives here and the step below just calls `tsc`, which resolves through whatever npm put in the root `.bin`. It works today because the direct dependency wins the root slot and attw's own 5.6.1-rc nests under it, but if this line ever changes the job would quietly check the oldest-supported-TS contract on a modern compiler. A `tsc --version` assertion in this step would close that.

.github/workflows/emitted-types.yml
```diff
@@ -53,6 +64,27 @@ jobs:
...
+            publint@0.3.21 \
+            @arethetypeswrong/cli@0.18.3
+          echo "$tools/node_modules/.bin" >> "$GITHUB_PATH"
```
budnix: This folder holds seven binaries after the install, not three - attw, publint, tsc, tsserver, semver, marked and highlight - and GITHUB_PATH prepends, so every later step in the job picks those up first. Nothing after this calls `semver` or `marked` today, so it is fine right now, but it sets a trap for whoever adds the next step. Writing `TOOLS_BIN` to `$GITHUB_ENV` and calling each tool by path would keep the effect inside the three steps that want it.

.github/scripts/__tests__/emitted-types.test.mjs
```diff
@@ -0,0 +1,72 @@
...
+    0,
+    'do not bring back per-step `npx -y` — each one is an uncached registry fetch'
+  );
+});
+
+test('job-level env does not use the runner context', () => {
+  // Reusable workflows reject ${{ runner.* }} in jobs.<id>.env. The caller
+  // then fails at parse time with 0 jobs, so CI Gate never reports
+  // (DEV-2783, runs 33853301602 / 33853913419).
+  const jobEnv = withoutComments.match(/\n    env:\n([\s\S]*?)\n    steps:/)?.[1] ?? '';
```
budnix: I tried two regressions against this and both pass. Move the `env:` block after `steps:` while keeping a `${{ runner.temp }}` in it - valid YAML, same parse failure on the caller, but the regex matches nothing and `?? ''` turns that into a green assertion. The second one is `COPY: ${{ env.HOT_PREBUILT }}`, which `jobs.<id>.env` rejects the same way it rejects `runner`, but the guard only looks for `runner.`. Failing when the block is not found, and banning `env`, `steps` and `job` alongside `runner`, would catch both.

.github/scripts/__tests__/emitted-types.test.mjs
```diff
@@ -0,0 +1,72 @@
...
+
+test('type-check tools are installed once, not via three cold npx -y calls', () => {
+  assert.match(
+    WORKFLOW,
+    /name: Install type-check tools/,
+    'expected a single Install type-check tools step'
+  );
+  assert.match(
+    WORKFLOW,
+    /actions\/cache@/,
```
budnix: This only checks that the string `actions/cache@` appears somewhere, so both ways the cache can quietly stop working stay untested. I changed the key to a stale value and then changed `path: ~/.npm` to something else, and the test stayed green each time. The three versions are written twice, once in the install step and once in the cache key, so bumping attw and forgetting the key gives you a permanent cache miss with a passing test. Asserting that the key names each installed version, and that `path: ~/.npm` is present, would make this hold.

.github/scripts/__tests__/emitted-types.test.mjs
```diff
@@ -0,0 +1,72 @@
...
+// the race back.
+
+const WORKFLOW = readFileSync(
+  path.join(repoRoot(), '.github/workflows/emitted-types.yml'),
+  'utf8'
+);
+const withoutComments = WORKFLOW.replace(/#[^\n]*/g, '');
+
+test('the Emitted Types job budget is at least 30 minutes', () => {
+  const timeout = Number(/timeout-minutes:\s*(\d+)/.exec(WORKFLOW)?.[1]);
```
budnix: This reads `WORKFLOW` with comments still in, while lines 42, 63 and 68 read `withoutComments`. It also takes the first `timeout-minutes:` match in the file, so a step-level `timeout-minutes:` added above the job's own, or a comment that spells a number in that form, would make the check read the wrong value. Lines 32 and 37 have the same mix. Switching all three to `withoutComments` keeps one rule for the whole file.

.github/workflows/emitted-types.yml
demtario: PRs still check the five typed roots. test.yml and develop.yml pass attw-full from run-all, so develop / master / the publish.yml release path walk every exports path and keep the old untyped-resolution / no-resolution ignores. e6d6fc1

.github/workflows/emitted-types.yml
demtario: Added ./.github/workflows/emitted-types.yml to the verify-types filter. This push should run the job, including the non-prebuilt install + full-build fallback.

.github/workflows/emitted-types.yml
demtario: Added the prefix restore-key.

.github/workflows/emitted-types.yml
demtario: Install now uses --prefer-offline --no-audit --no-fund.

.github/workflows/emitted-types.yml
demtario: The install step now greps tsc --version for Version 5.1.6.

.github/workflows/emitted-types.yml
demtario: TOOLS_BIN goes to GITHUB_ENV. Later steps call $TOOLS_BIN/{tsc,publint,attw} and nothing is prepended to GITHUB_PATH.

.github/scripts/__tests__/emitted-types.test.mjs
demtario: The test now fails when the job env block is missing (env after steps) and bans env / steps / job alongside runner. Both of those mutations are red.

.github/scripts/__tests__/emitted-types.test.mjs
demtario: The key must name each installed version and path must be ~/.npm. A stale key or a different path is now red.

.github/scripts/__tests__/emitted-types.test.mjs
demtario: Timeout, install, and cache assertions all read the comment-stripped file. timeout-minutes is the job-level (4-space) match.

budnix (dismissed):

.github/scripts/__tests__/emitted-types.test.mjs
```diff
@@ -0,0 +1,178 @@
...
+}
+
+/**
+ * Escape a semver string for use inside a RegExp.
+ *
+ * @param {string} value A version like `5.1.6`.
+ * @returns {string} The escaped pattern.
+ */
+function escapeDots(value) {
+  return value.replace(/\./g, '\\.');
```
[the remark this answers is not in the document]
demtario: Removed escapeDots. The pin test now compares the three versions as plain strings, so nothing is interpolated into a RegExp. 2cf6d06

budnix (approved):

Issue #3119, opened 2016-01-12
Issues with scrolling in a parent element


ohadb1:
We use handsontable for a few years as our main input table widget. We have multiple forms where we need to have synchronized scrolling with frozen row headers.
To achieve that we used to generate the tables inside a div with a designated width and overflow-x, while each table is created within a child div with a large width and all columns rendered.
This setup stopped working in version 0.20.3. with the following phenomena:
1. While the height of the hider in the left container is correct (fit the rows) the height of the holders in each of the clones is wrong. We got over that by enforcing the hider height on all holders at the afterRender hook callback.
2. When you scroll (horizontally), the afterScroll hook is fired only once, at the first scroll event but not on subsequent scrolls, although the onTableScroll of the overlay is fired everytime. It just gets to an execution flow without a refreshAll() call that drives the transformation of the left clone to the correct coordinates. This eliminates the header freezing effect.
3. When the parent horizontal scrolling is not zero, the positioning of the editor container div has wrong offset as it doesn't compensate for the scrolling status of the parent element (which is calculated correctly in other places in the code.

We are open to other ways of setting synchronized scrolling of multiple tables, including with the built-in scroll bar, but there are other instances where the wrapping element is scrolled (often vertically), and then the editor positioning is wrong.

AMBudnik:
Welcome @ohadb1 can you create a demo code where we can track all the changes that are important to you?

ohadb1:
https://jsfiddle.net/ecn0njcs/5/

AMBudnik:
Thank you for demo.
I tried to check scrolling issue in comparison with older versions but I can't see the difference - although in none of them it was working fine till you use a new settings option called `preventOverflow` (since [0.20.3](https://github.com/handsontable/handsontable/releases/tag/0.20.3)), when I added it to your instances' settings it provided correct behaviour: https://jsfiddle.net/ecn0njcs/6/

ohadb1:
It was working on an older version. Anyway, this fixed onlt the height of the hider.
There are still two issues: 
1. the scrolling in the outer div doesn't fire the callback that sets the transform3d property of the left clone to compensate for the scrolling and float the row headers.
2. The positioning of the editor container div doesn't compensate for scrolling, so if you have a non-zero scroll, the editor will appear to the left with the scroll offset instead of above the edited cell.

AMBudnik:
Don't worry, I'll try to bring you answers to all issues.
[...]

PR #13379, opened 2026-09-04
DEV-2789: Resolve the trimming container per scroll axis

budnix:
### Context

The PR gives each scroll axis of the Walkontable engine its own trimming container, so a grid with a definite `width` and no `height` scrolls its columns inside its own box while the window scrolls its rows.

Today `getTrimmingContainer()` names one container for both axes and deliberately skips a single-axis `clip` (DEV-1025). Core writes `overflow-x: clip` on the root for a definite `width` with no sized `height`, so the engine hands both axes to the window and the columns past the width are hidden with no scrollbar. The documented workaround is "also set `height`". `preventOverflow: 'horizontal'` is that mixed layout already, but its horizontal predicate reports the window, the workspace width falls back to the document width when the content is wider, `hasHorizontalScroll()` is `false`, the bottom overlay scrolls with the holder instead of the window, and the vertical scroll hooks never fire.

This is Parts B1 (engine) and B2 (core contract) of the `width`/`height` options unification (DEV-2789). They must land before the core part that changes `height: 'auto'`, because that change drops the both-axes clip and would leave a `height: 'auto', width: 500` grid with no way to reach its columns.

**Part B1 – engine per-axis trimming:**

- `getTrimmingContainer(base, axis)` overload in `helpers/dom/element.ts`. The public no-axis form is unchanged, including the DEV-1025 exemption.
- `overlay/axisOwner.ts` resolves the owner of one axis; `preventOverflow` becomes an alias that forces one axis's owner to the root's parent.
- Top and bottom overlays carry the vertical owner, the inline-start overlay the horizontal one; `isVerticallyScrollableByWindow()` / `isHorizontallyScrollableByWindow()` read those. Every decision about the other axis goes through the viewport predicates (`rootSized`, scrollbar clearance, scrollbar subtraction).
- `MasterTable#alignOverlaysWithTrimmingContainer` lays the holder out per axis (split mode: owner box on an element-owned axis, DOM-driven on a window-owned one). A vertical owner with no intrinsic height gets `height: auto`, the #3119 guard the element mode already had.
- `Overlays#beforeDraw` re-resolves the owners on every full draw. They were re-resolved only inside `adjustElementsSize`, which the draw cycle gates on a position or spreader change, so a removed clip left an overlay on the element while the master resolved the window.
- `ScrollSync` reads the per-axis scroll positions off each overlay's own scrolling element and rebinds once when an owner's identity moves.
- Test harness: the Jasmine bridge reporter (both copies) now describes a failed expectation's `expected`/`actual` when it cannot be serialized. A failing spec whose value was cyclic (`toBe(window)`, `toEqual([overlay])`) used to throw inside the reporter and vanish from the run, leaving it green one spec short. That hid the `getOverlays` spec, which has been failing since #12951 changed the overlay order; the spec now asserts the real order (the only consumer, MergeCells, iterates the whole list).

**Part B2 – the core contract for `height: 'auto'`:**

- `core.ts` now writes the `overflow-x: clip` longhand for `height: 'auto'` with a definite width, the same contract an unset height writes. `'auto'` still writes the `overflow: clip` shorthand, so this changes nothing readable today; it is what keeps every column reachable once `'auto'` stops writing the shorthand (the next PR). Only an unset height clears the longhand: for `'auto'` with a relative width the shorthand stays whole, so the clip is not silently reduced to the vertical axis.
- The core consumers of the two predicates are verified in the Playwright suite: the root's `htHasScrollX` / `htVerticallyScrollableByWindow` classes, `stretchH: 'all'` filling the root instead of the page, and `updateSettings({ width })` on a height-less grid moving the horizontal owner to the root without a height change (carried by the engine's owner resync).
- The ordering canary is pinned as user behavior: `height: 'auto'` keeps every column reachable through the holder. It passes today through the element mode and after the `'auto'` change through the split mode, and fails if the shorthand is dropped before the engine can split the axes.

Docs: the "Per-axis trimming containers" section in `handsontable/src/3rdparty/walkontable/AGENTS.md`, the gate rows in `.ai/RENDERING-LIFECYCLE.md`, a paragraph in `.ai/ARCHITECTURE.md`, a coverage note in `.ai/CONCERNS.md`, the `width` JSDoc tip (a definite width scrolls horizontally on its own), and the traps in the `walkontable-dev`, `walkontable-testing`, and `handsontable-e2e-testing` skills.

- Unit tests added/modified (`*.unit.js`): `handsontable/src/helpers/dom/__tests__/element.unit.ts` (per-axis cases), `handsontable/src/3rdparty/walkontable/test/unit/overlay/scrollSync.unit.ts` (`resyncScrollableElementsWithOwners`)
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`): `tests/e2e/width-window-scroll.spec.ts` (new, with `tests/fixtures/demo/width-window-scroll.html` and `tests/fixtures/pages/WidthWindowScrollPage.ts`): holder scrolls the columns, page has no horizontal scrollbar, frozen columns pinned on a holder scroll, frozen rows pinned on a window scroll, both axes virtualized, root scroll classes, `height: 'auto'` canary, `stretchH: 'all'` against the root width, owner move through `updateSettings({ width })`, RTL twin, legacy `preventOverflow: 'horizontal'` leg with the vertical scroll hook firing
- Walkontable specs (own runner): `viewport.spec.js`, `table/table/master.spec.js` (split mode, clip removed, two-instance #3119 case), `overlay.spec.js`, `settings/preventOverflow.spec.js`
- Legacy Jasmine edits: `handsontable/src/__tests__/settings/width.spec.js` (the "known limitation" comment removed, horizontal scroll asserted; `height: 'auto'` + definite width writes the longhand, `height: 'auto'` + relative width keeps the shorthand)
- Type tests (`*.types.ts`) updated if public API changed: `handsontable/src/helpers/dom/__tests__/dom.types.ts`
- For a bug fix — the spec that fails without this fix: `width-window-scroll.spec.ts` "owns the horizontal axis in the root and leaves the vertical one to the window"; `preventOverflow.spec.js` "should scroll the frozen bottom rows with the window when `horizontal` is passed"; `master.spec.js` "should hand both axes back to the window when the clip is removed"

### Commands run

```
npm run test:walkontable --prefix handsontable         # 835 specs, 0 failures
npm run test:e2e --prefix handsontable                  # 9895 specs, 0 failures, 8 pending (main, B1)
npm run test:e2e --prefix handsontable -- --testPathPattern="settings/width|settings/height|emptyDataState"   # 122 specs, 0 failures (B2)
npm run test:e2e --prefix handsontable -- --testPathPattern="colHeader|nestedHeaders/__tests__/rowspan|columnSorting/__tests__/columnSorting|manualColumnResize/__tests__/manualColumnResize|autocompleteEditor|a11y/attributes|core/count|core/view|settings/preventOverflow|core/scrollViewportTo|stretchColumns"   # 666 specs, 0 failures (B2)
npm run test:e2e --prefix handsontable -- --testPathPattern="settings/width|settings/preventOverflow|emptyDataState|core/view|core/scrollViewportTo" --theme=horizon   # 306 specs, 0 failures
npm run test:e2e --prefix handsontable -- --testPathPattern="settings/width|settings/preventOverflow|emptyDataState|core/view|core/scrollViewportTo" --theme=classic   # 306 specs, 0 failures
cd tests && npx playwright test --project=e2e-main      # 865 passed, 1 skipped (one start-of-run timeout in auto-row-header-size, 6/6 alone)
cd tests && npx playwright test width-window-scroll --project=e2e-main   # 10 passed
npm run test:unit --prefix handsontable                 # 4429 passed
npm run test:types --prefix handsontable                # ok
npm run eslint --prefix handsontable                    # clean on all changed files
```

### Types of changes

### Related issue(s):

1. DEV-2789
2. DEV-1025 (the single-axis clip exemption, kept on the public form)
3. #3119 (the intrinsic-height guard, now shared with the split mode)

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

- [ ] MANUAL QA NEEDED — 

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2789

---

> [!NOTE]
> **High Risk**
> Changes core Walkontable scroll sync, overlay positioning, and trimming-container resolution across frozen rows/columns, `preventOverflow`, and cross-realm iframes—areas where small mistakes cause silent layout or scroll regressions.
> 
> **Overview**
> Walkontable now resolves **horizontal and vertical scroll ownership separately**, so a grid with a definite `width` and no sized `height` scrolls columns inside the root while the page scrolls rows. That replaces the old single `getTrimmingContainer()` answer that ignored single-axis `overflow-x: clip` and left extra columns unreachable.
> 
> The engine adds **`resolveAxisOwner()`** and a per-axis **`getTrimmingContainer(base, axis)`** overload; region overlays store the owner for their axis, and **`isHorizontallyScrollableByWindow()` / `isVerticallyScrollableByWindow()`** drive overlay sizing, scrollbar clearance, and workspace measurement. **`MasterTable`** lays out the holder in split, element, or full-window modes (including clearing stale pixel sizes when leaving split mode and the #3119 intrinsic-height guard). **`ScrollSync`** reads each axis from its overlay’s scrolling element, rebinds when owners change, and uses **`isHTMLElement`** instead of realm-bound `instanceof` for scroll offsets and wheel handling.
> 
> **`preventOverflow: 'horizontal'`** is modeled as forcing the horizontal owner to the grid root, fixing frozen bottom rows, workspace width, and vertical scroll hooks in that layout. **Core** writes **`overflow-x: clip`** for **`height: 'auto'`** with a definite width so the same contract applies before a follow-up change drops the both-axes clip shorthand.
> 
> Test infrastructure: the Jasmine bridge reporter **sanitizes unserializable expectation values** so failed specs are not dropped from Puppeteer runs. Docs, skills, `width` JSDoc, and broad Walkontable/core tests cover split mode, owner moves, and RTL.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7617f6715fb4119c082ab283e650110936b34204. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:

handsontable/src/3rdparty/walkontable/src/overlay/scroll/scrollSync.ts
```diff
@@ -351,7 +375,9 @@ export class ScrollSync {
     const inlineStartOverlay = this.#deps.getInlineStartOverlay();
     const masterScrollable = topOverlay.mainTableScrollableElement;
```
demtario: **high** `syncScrollWithMaster` still reads both axes off `topOverlay.mainTableScrollableElement`. In split mode that is the window, so `isHTMLElement` returns and the holder's `scrollLeft` is never copied onto a newly shown top/bottom clone.

`syncScrollPositions` already reads per axis ([L307–L308](https://github.com/handsontable/handsontable/blob/ccab4b397b71a58a8fdba8e0d067835531574c16/handsontable/src/3rdparty/walkontable/src/overlay/scroll/scrollSync.ts#L307-L308)). Mirror that here: `scrollLeft` from `inlineStartOverlay.mainTableScrollableElement`, `scrollTop` from `topOverlay.mainTableScrollableElement`, copy each only when that owner is an element. Add a spec that scrolls the holder, then enables `fixedRowsBottom` (or `fixedRowsTop` on a header-less grid) and asserts clone `scrollLeft` matches.

handsontable/src/3rdparty/walkontable/src/overlay/scroll/scrollSync.ts
```diff
@@ -576,13 +651,20 @@ export class ScrollSync {
...
-        const overflow = geometryReader
-          .getComputedStyle(tableParentNode as Element).getPropertyValue('overflow');
+        const style = geometryReader.getComputedStyle(tableParentNode as Element);
+        const traps = (value: string) => value === 'hidden' || value === 'clip';
 
-        return overflow === 'hidden' || overflow === 'clip';
+        return traps(style.getPropertyValue('overflow-x')) || traps(style.getPropertyValue('overflow-y'));
       })();
 
     return isOverflowClip ? wtTable.holder : getScrollableElement(wtTable.TABLE);
```
demtario, line 682: **high** Keeping `scrollableElement` as the holder when either axis is clipped is documented, but `#onCloneWheel` then translates onto that holder and `preventDefault`s if any axis moved. A clone wheel that moves `scrollLeft` cancels the window-owned vertical scroll — the layout this PR advertises.

Drive `scrollVertically` / `scrollHorizontally` from each overlay's `mainTableScrollableElement` (holder on X, `rootWindow.scrollBy` on Y in split mode) and `preventDefault` only for the consumed axis.

demtario (commented):
2 inline notes below.

Off-diff:
- `overlays.ts:683` — `scrollHorizontally` / `scrollVertically` still use realm-bound `instanceof HTMLElement`, so overlay wheel cannot scroll columns in the iframe+width-only path. Use `isHTMLElement()`.

handsontable/src/3rdparty/walkontable/src/overlay/scroll/scrollSync.ts
budnix: Good spot — fixed in 8f1b6ddedd. There was a second blocker under it: a nested draw's `beforeDraw` was wiping the render-state flag before the sync ran, so `setRenderingStateChanged` now latches until consumed.

handsontable/src/3rdparty/walkontable/src/overlay/scroll/scrollSync.ts
budnix: You're right — fixed in 8f1b6ddedd. Both methods now drive each axis's own owner, with `rootWindow.scrollBy` for a window-owned axis, which also takes care of the `instanceof` you flagged off-diff at `overlays.ts:683`.

demtario (commented):
1 note with no changed line to anchor to:

- `nativeScrollInput.ts:221` — `#onTableScroll` still uses realm-bound `instanceof HTMLElement` on `mainTableScrollableElement`, so a key-press holder scroll in the iframe+width-only path skips `syncScrollPositions`. Use `isHTMLElement()`. Extend `iframe-cross-realm-scroll.spec.ts` with an arrow-key holder scroll that asserts the top clone stays aligned.

demtario (dismissed):

budnix:
Agreed — fixed in 35f3fc904d. The arrow-key test you asked for also turned up a second one: `getScrollTop` returned `undefined` for a cross-realm window, so the iframe grid was rendering its last rows at page top. Both are covered in `iframe-cross-realm-scroll.spec.ts` now.

demtario (dismissed):

handsontable/src/3rdparty/walkontable/src/overlay/regions/bottomOverlay.ts
```diff
@@ -226,11 +225,10 @@ export class BottomOverlay extends Overlay {
...
-    const preventOverflow = this.wtSettings.getSetting<boolean | string>('preventOverflow');
 
-    // Both strips need this overlay to be laid out against the scrollport; when the window anchors it
-    // instead, `repositionOverlay` never runs and clipping would expose the master for nothing.
-    const rootSized = this.trimmingContainer !== rootWindow || preventOverflow === 'horizontal';
+    // Width is a horizontal question: sized against the scrollport whenever an element owns the
+    // horizontal axis (see `TopOverlay#adjustRootElementSize`).
+    const rootSized = !wtViewport.isHorizontallyScrollableByWindow();
     // Clip and band together, or not at all - see `TopOverlay#adjustRootElementSize`.
     const clearanceApplies = holderOwnsScrollbars(this.trimmingContainer, rootWindow);
```
demtario: **high** One `clearanceApplies` is taken from this overlay's vertical `trimmingContainer`, then reused for the horizontal bottom strip. In split mode that owner is the window, so `holderOwnsScrollbars` is false and `#bottomClearance` stays 0. Frozen bottom rows then paint over the holder's horizontal scrollbar.

Walkontable AGENTS.md requires a clearance strip to read the owner of the axis the strip lies on. The inline-start clone already does that ([inlineStartOverlay.ts#L204](https://github.com/handsontable/handsontable/blob/baf367583c3774f2afeb40aa6918d287758b2bdb/handsontable/src/3rdparty/walkontable/src/overlay/regions/inlineStartOverlay.ts#L204)) and publishes a non-zero horizontal strip, so the bottom band notches at the frozen-column / frozen-row join (#10370). The Playwright spec only asserts X alignment after `fixedRowsBottom` is enabled.

Split the predicates: keep the vertical strip on `this.trimmingContainer`, and key `#bottomClearance` on the horizontal owner (`inlineStartOverlay.trimmingContainer` / `!isHorizontallyScrollableByWindow()`). Apply that bottom strip only when this clone actually sits on the holder's bottom edge — the window-Y path never runs `repositionOverlay`, and `hasVerticalScroll()` at 271 is the wrong stand-in for that.

handsontable/src/3rdparty/walkontable/src/overlay/regions/bottomInlineStartCornerOverlay.ts
```diff
@@ -119,7 +122,9 @@ export class BottomInlineStartCornerOverlay extends Overlay {
     //
     // A touch-only device has no pointer that could reach the scrollbar - see `canGrabScrollbar`.
     // Clip and band together, or not at all - see `TopOverlay#adjustRootElementSize`.
-    const clearanceApplies = holderOwnsScrollbars(this.trimmingContainer, this.deps.rootWindow);
+    // Read off the frozen-bottom-rows overlay, whose axis this strip belongs to, so the two gates
+    // cannot disagree.
+    const clearanceApplies = holderOwnsScrollbars(this.bottomOverlay.trimmingContainer, rootWindow);
```
demtario: **high** The corner's bottom (horizontal) strip is keyed off `bottomOverlay.trimmingContainer`, which is the vertical owner. Same split layout, same covered scrollbar. The new comment says that container is "the axis this strip belongs to." It is not.

Read the horizontal owner (`inlineStartOverlay.trimmingContainer` or `!isHorizontallyScrollableByWindow()`), and keep the gate identical to `BottomOverlay`'s bottom strip so the two cannot notch. Do not copy the bottom overlay's vertical `trimmingContainer`.

demtario (commented):
2 inline notes below.

handsontable/src/3rdparty/walkontable/src/overlay/regions/bottomOverlay.ts
budnix: Nice catch — fixed in bc1f587288. The bottom strip now reads the horizontal owner, and `#restsOnHolderBottomEdge` replaces the `hasVerticalScroll()` stand-in you flagged.

handsontable/src/3rdparty/walkontable/src/overlay/regions/bottomInlineStartCornerOverlay.ts
budnix: You're right, that comment was plain wrong. Fixed in bc1f587288 — the corner now reads the bottom overlay's finished strip through `getBottomClearance()` instead of computing its own.

demtario (approved):

Issue #8688, opened 2021-09-01
Incorrect DropdownEditor's height if we set height as 'auto' for a table


mrpiotr-dev:
### Description
If we initialize Handsontable with height set as `auto`, then DropdownEditor's height will be much shorter than it should be.

### Steps to reproduce
1. Go to the demo
2. Click on the DropdownEditor's button in one of these three cells

Result: Editor is open, but options are not visible.

Now remove `height: 'auto'` from the initialization object and repeat the above steps.

### Demo
https://jsfiddle.net/6j43qwz5/

### Related issues

- #3828

### Your environment
* Handsontable version: any
* Browser Name and version: any
* Operating System: any

Inform [forum](https://forum.handsontable.com/t/dropdown-issue-with-fewer-rows/5493)

AMBudnik:
Testing this behavior on the newest Chrome 96 on macOS Monterey 12.0.1 using reported Handsontable v9.0.0 I also see the editor being cut and visually altered

![Kapture 2021-11-29 at 10 16 53](https://user-images.githubusercontent.com/10757813/143840272-cbc35c37-bc92-4637-bf43-482ca9b12904.gif)


If there are only two rows the list of choices does not show

![Zrzut ekranu 2021-11-29 o 10 19 04](https://user-images.githubusercontent.com/10757813/143840566-b8b824ff-3e20-4d78-9505-e188a16ca0ad.png)

Behavior also replicable in Handsontable v11 (updated demo https://jsfiddle.net/14kgumxb/3/)

aninde:
Removed from v12.1.0.

warpech:
We had a fix for this, but we reverted it. We will make another attempt. Some details in the internal discussion: https://handsoncode.slack.com/archives/C01SYSTAGCT/p1655993690103669?thread_ts=1655915253.916219&cid=C01SYSTAGCT

adrianszymanski89:
Inform [Forum](https://forum.handsontable.com/t/excess-space-is-showing-bottom-of-the-table/6490/2)

anlagno:
Hello, Handsontable team!
Is there any workaround to fix this issue? Or any update in which version it will be fixed?

adrianszymanski89:
Hi @anlagno 

The current workaround for this is either not to set the height option at all or to set it to a fixed value. Then the dropdown menu will be rendered correctly.

ronwang01:
Any update on this issue?

adrianszymanski89:
[...]

Issue #1319, opened 2014-02-18
Date column doesn't allow manual input


MonchoDev:
Tried different configurations for the date type column but it doesn't allow me to do a manual entry, it will auto load the date, and will only write an extra character at the end of the date.

PR #13381, opened 2026-09-04
DEV-2789: Unify the width and height options, height auto as CSS auto

budnix:
### Context

The PR changes the grid `width` / `height` options so both axes share one contract: a defined set of accepted values, `'auto'` meaning what CSS `auto` means on a `<div>`, and a symmetric `null` reset. It is Part A of the width/height unification (DEV-2789) and builds on #13379 (Parts B1 and B2, the engine's per-axis trimming, merged): that split-axis engine mode is what keeps a `height: 'auto', width: 500` grid's columns reachable once `'auto'` stops clipping.

**Breaking change: `height: 'auto'` no longer clips the grid.**

Before, `height: 'auto'` wrote `height: auto; overflow: clip;` on the root element. The clip made the root the element the grid scrolled in, so every row rendered at once (5000 `<tr>` instead of ~36, #1319) and a dropdown editor on the last row was cut off at the grid's edge (#8688 / DEV-271). Now it writes `height: auto` and nothing else: the grid behaves like a plain block element, the nearest scrolling ancestor or the page scrolls it, and the rows stay virtualized. The reported `'auto'` case of #8688 is fixed by this; the fixed-`height` case still needs the editor portal (DEV-1656).

Migration, before and after (full entry: `## 17.` in the 18.1 → 19.0 migration guide):

```js
// Before: an internal clip, every row rendered, the page never scrolled the grid.
height: 'auto'

// After: the page (or the nearest scrolling ancestor) scrolls the grid, rows are virtualized.
height: 'auto'

// To keep an internal vertical scrollbar, size the grid or its parent:
height: 500                       // or '50%', '75vh', 'calc(100% - 40px)'
// or: <div style="height: 300px; overflow: auto"><div id="grid"></div></div> with height: 'auto'
```

What the options accept now, for both axes:

| Value | Result |
| --- | --- |
| a number, `'500'`, `'500px'` | `<n>px`, the grid scrolls inside its box; a pixel height has the root wrapper's top and bottom bars (pagination, sheets bar, license notification) subtracted |
| `'50%'`, `'75vh'`, and any other CSS length or expression the browser reads (`'20em'`, `'calc()'`, `'var()'`, `'cqw'`) | passed through as written |
| `'auto'` | inline `auto`, no clip |
| a function | called, then resolved as above |
| `null` | restores that axis's initial inline value, leaves the other axis alone |
| anything else (`'abc'`, `''`, `-100`, `true`), and 13 CSS keywords: the ones that set no size (`'inherit'`, `'initial'`, `'unset'`, `'revert'`, `'revert-layer'`, `'none'`, `'normal'`), the content-sizing ones (`'min-content'`, `'max-content'`, `'fit-content'`, `fit-content()`), and the fill ones that read as a fixed size (`'stretch'`, `'-webkit-fill-available'`, `'-moz-available'`) | ignored: one `console.warn` per grid and value, no stray clip, and `getSettings()` keeps the previous value (the default on init) |

Two smaller behavior changes ride along: `width: null` clears the inline width (it wrote `width: nullpx`), and `height: null` restores the height only (it replayed the whole initial `style` attribute, wiping a `width` set through the option). A `width` set in the same call as that restore still clips the horizontal axis. The overflow the root writes is derived from the DOM state after both axes are applied: a sized height clips both axes; a free height (`'auto'` or unset) with a definite width clips the horizontal axis only; a free height with a free or container-driven width (`%`, viewport and container-query units, `var()`, `env()`) clips nothing. That last rule replaces a regex on the literal string that clipped `var(--w)` and `50cqw`.

Implementation:

- `src/utils/rootSize.ts`: `resolveRootSize()` (grammar first, `CSS.supports` as the tail oracle, so jsdom resolves the documented forms without it) and `classifyInlineSize()`. `describeValue()` moved to `src/utils/describeValue.ts`, shared with the header-size warning.
- `src/core/rootSize.ts`: `applyRootSize()` replaces the root size block in `updateSettings` and is the only writer of the root's inline `height`, `width` and `overflow*`. Validation runs on the value `beforeHeightChange` / `beforeWidthChange` return. A pixel height gets the edge-slot reservation here (`reserveEdgeSlotsHeight`), and the slot `ResizeObserver` re-applies it through `reapplyPixelRootHeight`. `applyRootSize()` reports the axes it ignored, and `updateSettings` puts the previous stored value back for them, so neither `getSettings()` nor the re-apply reads a size the grid never used. The scrollable-element re-pick rule is a superset of the old one (it also fires when the height moves between free and sized).
- Walkontable: a window-owned vertical axis always resolves to the window, a window-owned horizontal one only while the vertical axis is window-owned too (`Overlay#ownsWindowScroll()`); per-axis scrollbar clearance (`holderOwnsAxisScrollbar`), where the horizontal strip asks which element scrolls the columns (`inlineStartOverlay.mainTableScrollableElement`), so the reverse split clears the holder's scrollbar under frozen columns and frozen bottom rows; the clone scroll sync writes only the axes an element owns; a wheel gesture moves each axis exactly once (`NativeScrollInput#onWheel` scrolls both axes itself and always consumes the event); the bottom overlay and bottom corner subtract the table overhang only while the window owns the vertical axis.
- `GridSettings.width` / `height` widen to `number | 'auto' | (string & {}) | null | (() => number | string | null)`.
- Docs: `height` / `width` JSDoc and the grid-size guide name the rejected keywords and why, and spell out that any `height` other than `'auto'` clips both axes while a `width` clips only when it is definite. Migration `## 17.` has before/after code and three smaller changes (the ignored values and keywords, `var()` and container-query widths no longer clipped under a free height, `width: null`). The root-size and wheel rules are in `handsontable/AGENTS.md` and the Walkontable `AGENTS.md`. The three recipe families that pair `height: 'auto'` with `width: '100%'` (frozen summary row, freeze columns, server-side) were reviewed and need no edit: that pair scrolls with the page on both axes, before and after.

Visual diffs: the PR tier (Chromium, `main` and `main-dark`) reports 2 changed screenshots, `empty-data-state-no-columns-1` on both themes, waiting on `Visual / approve`. The only difference is the text antialiasing of the "No data available" message (subpixel versus grayscale); no element moves (1,856 and 1,846 pixels, max channel delta 96 and 93, measured 2026-09-17). The cross-browser leg, which the PR tier does not run, shows the same antialiasing change on the `height: 'auto'` grid of `basic-two-tables-demo` (measured 2026-09-07).

Known consequence, not changed here: `height: 'auto'` without a definite `width` puts both axes in window mode. In that mode, a `stretchH` grid in a container narrower than its columns' base widths stretches the columns to the document width (1280px in a 400px container) and the page scrolls sideways. A grid with no `height` does the same on `develop` (measured against develop's own build). `stretch-columns-container-resize.spec.ts` pins its below-base-sum case on a pixel height for that reason.

- Unit tests added/modified (`*.unit.js`): `handsontable/src/utils/__tests__/rootSize.unit.ts` (the resolver: numbers, pixel strings, `auto`, every unit family, functions with and without the oracle, every rejected keyword, unreadable values; container-driven detection through `classifyInlineSize`); `handsontable/src/core/__tests__/rootSize.unit.ts` (jsdom applier: inline-style contract per case, one warning per distinct bad value, `null` per property, initial-style restore, an overflow the module does not own, an ignored value keeping the previous stored setting and the default on init, a definite width clipping beside a `height: null` restore); `handsontable/src/__tests__/core/layoutSlotReservation.unit.js` (slot reservation on a pixel height, overflow read through the longhands, the slot observer re-applying the last readable height after an ignored update); `walkontable/test/unit/overlay/scrollSync.unit.ts` (clone sync per axis owner); `walkontable/test/unit/overlay/scrollbarClearance.unit.ts` (`holderOwnsAxisScrollbar` per axis); `helpers/dom/__tests__/element.unit.ts` (`isHTMLElement` for an element from another realm)
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`): `tests/e2e/root-size-options.spec.ts` (new, with `tests/fixtures/demo/root-size-options.html` and `tests/fixtures/pages/RootSizeOptionsPage.ts`): `'auto'` writes `height: auto` only, virtualizes the rows exactly like an unset height, beats a stylesheet height, clips X only with a definite width and keeps every column reachable, clips nothing with `%` or `var()`, fills a bounded parent and scrolls inside it, flips clip and scroll owner both ways through `updateSettings`, the #8688 dropdown on the last row shows all 10 options; sized values (`300`, `'300'`, `'300px'`, `'50%'`, `calc()`); five unreadable values plus `width: 'inherit'`, each one warning naming option and value, grid unchanged; `null` per axis. `tests/e2e/width-window-scroll.spec.ts`: a diagonal wheel moves each axis exactly once and scrolls the page itself, a horizontal-only wheel moves the columns once and leaves the page still, and a `width`-only change re-picks the scroll owner. `tests/e2e/bottom-slot-sizing.spec.ts`: with `height: 'auto'` in a fixed CSS box the page owns the vertical axis and the bar still follows the last row. `tests/e2e/stretch-columns-container-resize.spec.ts`: the `'auto'` cases pin window mode (an unsized top clone root, a grow that stays inside the viewport, and a grow past it that gives the page a real horizontal scrollbar), the clone-root width check runs on a pixel height, and the below-base-sum case uses a pixel height. `tests/e2e/overlay-scrollbar-clearance.spec.ts`: in the reverse split the bottom strip is drawn and the frozen bottom rows, frozen columns, and bottom corner are clipped out of it; the case fails on the build without the gate fix. `tests/fixtures/demo/auto-height-scrollbar-zoom.html`: a definite `width` keeps the holder a content-sized scroll box on both axes; all four assertions are unchanged, and removing the column header height fraction from the hider sum still fails the 90% and 80% cases
- Legacy Jasmine edits: `settings/height.spec.js` (pixel strings, `'auto'` contract and virtualization, clip flip both ways, unreadable values, pass-through, `null`), `settings/width.spec.js` (the `'auto'` contract, `null`, `var()`/`cqw`), `hooks/beforeHeightChange.spec.js` and `beforeWidthChange.spec.js` (validation after the hook), `core/count.spec.js` (`'auto'` renders the window's band, not every row)
- Type tests (`*.types.ts`) updated if public API changed: `src/__tests__/core/settings.types.ts` (new string forms), new `src/__tests__/core/rootSize.types.ts` (accepted forms, rejected `true`/array, the `'auto'` literal pinned by exhaustiveness)
- Wrapper tests: `wrappers/react-wrapper/test/hotTable.spec.tsx` (`height="auto"` contract; the same bad value re-sent on three commits warns once), `wrappers/vue3/test/hotTable.spec.ts` (same contract; after a bad `height`, two unrelated prop changes make the wrapper re-send it, so it goes out 3 times and warns once)
- For a bug fix — the spec that fails without this fix: `root-size-options.spec.ts` "virtualizes the rows the same way an unset height does" and "shows every option of a dropdown editor on the last row (#8688)"; `count.spec.js` "should return the window's band of rows, not every row, with `height: 'auto'`"; `layoutSlotReservation.unit.js` "should re-apply the last readable height when a slot resizes after an ignored update"; `overlay-scrollbar-clearance.spec.ts` "clears the holder's horizontal scrollbar in the reverse split"

### Commands run

CI run [35324286591](https://github.com/handsontable/handsontable/actions/runs/35324286591) on `b3280b58f6` (the current head) was in progress when this was written. The previous head, `faf07b9422`, was green on every job in run [35233454208](https://github.com/handsontable/handsontable/actions/runs/35233454208), with `Visual / approve` waiting on the two antialiasing diffs above.

Local runs on `cce07ef2a7`, the last commit before the latest develop merge (that merge brings #13509 and #13556 and is covered by CI only):

```
npm run test:types --prefix handsontable                                   # ok
handsontable/node_modules/.bin/eslint <changed files>                     # 0 errors
npm run test:unit --prefix handsontable                                    # 403 suites, 5546 passed
npm run test:walkontable --prefix handsontable                             # 851 specs, 0 failures
npm run test:e2e --prefix handsontable -- '--testPathPattern=settings/(height|width)'                      # 48 specs, 0 failures
cd tests && HOT_TEST_PORT=8124 npx playwright test --project=e2e-main                                      # 1545 passed, 1 skipped
cd tests && HOT_TEST_PORT=8124 npx playwright test stretch-columns-container-resize overlay-scrollbar-clearance --project=e2e-main --project=e2e-horizon --project=e2e-classic   # 48 passed
npm --prefix wrappers/vue3 run test                                        # 41 passed
```

### Types of changes

### Related issue(s):

1. DEV-2789
2. #8688 / DEV-271 (the `height: 'auto'` case; the fixed-height case stays with DEV-1656)
3. #1319 (every row rendered under `'auto'`)
4. DEV-613 (define how `'auto'` should work), DEV-1102 (documentation of the size options)

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

- [ ] MANUAL QA NEEDED — 

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2789

---

> [!NOTE]
> **High Risk**
> Breaking layout and scroll ownership for `height: 'auto'` and revised width/height validation affects every grid using those options, with deep Walkontable scroll/wheel changes.
> 
> **Overview**
> **Breaking:** `height: 'auto'` now sets only inline `height: auto` (no `overflow: clip`), so scrolling moves to the page or nearest scroll ancestor and **row virtualization stays on**—the opposite of the old “render every row / internal clip” behavior.
> 
> Grid `width` and `height` are unified behind **`core/rootSize.ts`** and **`utils/rootSize.ts`**: one writer for root inline size and overflow, shared value resolution (pixels, CSS lengths, `'auto'`, functions, **`null`** per-axis reset), invalid values ignored with a one-time warning while **`getSettings()` keeps the last applied size**, and edge-slot height reservation moves here (`reapplyPixelRootHeight` on slot resize).
> 
> Overflow clipping is derived from classified sizes: sized height clips both axes; free height + definite width clips X only; free height + `%` / `var()` / `cqw` clips nothing. Walkontable follows with **`#ownsWindowScroll()`**, per-axis scrollbar clearance (`holderOwnsAxisScrollbar`), split-mode clone scroll sync, wheel handling so each axis scrolls once (no double page scroll), and bottom-corner positioning gated on vertical window ownership.
> 
> Docs, migration **§17**, changelog **13381**, types, and broad unit/E2E coverage accompany the behavior change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b3280b58f698cd99cda1362a971e40513eb79f0f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-271 incorrect `DropdownEditor`'s height if we set height as 'auto' for a table](https://app.clickup.com/t/9015210959/DEV-271)
Task linked: [DEV-613 Define how `auto` in `height` and `width` should work. ](https://app.clickup.com/t/9015210959/DEV-613)
Task linked: [DEV-1102 The `height` option does not have `auto` mentioned as an option](https://app.clickup.com/t/9015210959/DEV-1102)

handsontable/src/3rdparty/walkontable/src/overlay/regions/bottomOverlay.ts
```diff
@@ -275,23 +275,25 @@ export class BottomOverlay extends Overlay {
...
     // `TopOverlay#adjustRootElementSize`.
-    const inlineEndClearanceApplies = holderOwnsScrollbars(this.trimmingContainer, rootWindow);
-    const bottomClearanceApplies = holderOwnsScrollbars(
-      this.wot.wtOverlays.inlineStartOverlay.trimmingContainer, rootWindow
-    ) && this.#restsOnHolderBottomEdge();
+    const verticalClearanceApplies =
+      holderOwnsAxisScrollbar(wtViewport.isVerticallyScrollableByWindow(), rootWindow);
+    const horizontalClearanceApplies =
+      holderOwnsAxisScrollbar(wtViewport.isHorizontallyScrollableByWindow(), rootWindow) &&
+      this.#restsOnHolderBottomEdge();
```
demtario: **high** — Horizontal scrollbar-clearance gate keys off owner *identity* (`isHorizontallyScrollableByWindow()`) rather than the `ownsWindowScroll`-corrected answer, so the reverse split reintroduces the #10370 notch one mode over.

Reverse split (vertical = element, horizontal = window; e.g. `preventOverflow: 'vertical'`, or an ancestor clipping only the vertical axis): `Overlay#ownsWindowScroll('x')` is `false`, so `inlineStartOverlay.mainTableScrollableElement` is the holder — the holder really scrolls the columns and paints a horizontal scrollbar at its bottom edge. But `isHorizontallyScrollableByWindow()` is `true` (trimming container is the window), so `holderOwnsAxisScrollbar(true, rootWindow)` is `false`, `horizontalClearanceApplies` is false, `#bottomClearance` resolves to 0, and the frozen bottom / frozen-column overlays cover the holder's scrollbar with no clearance strip. `hasHorizontalScroll()` takes the measure path here and returns true, so the branch is reachable — only the owner-identity gate suppresses the strip.

The asymmetry encoded in `ownsWindowScroll()` (horizontal is window-scrolled only while the vertical axis is also window-owned) reached `#computeMainScrollableElement` but not the clearance gates. Feed the horizontal gates that corrected value. Same gate at `inlineStartOverlay.ts:202` and, off-diff, `bottomInlineStartCornerOverlay.ts` L142. Medium confidence: the logic gap is certain, the visible notch was not browser-verified and the mode is an edge.

docs/content/guides/upgrade-and-migration/migrating-from-18.1-to-19.0/migrating-from-18.1-to-19.0.md
```diff
@@ -696,3 +696,41 @@ columns: [
...
+the parent and scrolls inside it.
+
+Two smaller changes ship with this one:
+
+- A `width` or `height` value the browser cannot read as a size (`'abc'`, `-100`, `true`), or a CSS
+  keyword that would collapse the grid (`'min-content'`, `'inherit'`), is ignored with a one-time
+  console warning. It used to be written to the root element as it was.
+- `width: null` clears the inline width, the way `height: null` clears the height. It used to write
+  `width: nullpx`. Both resets restore their own property only, so `height: null` no longer removes a
+  `width` set through the option.
```
demtario, line 809: **high** — Missing migration note for a breaking sub-change. The changelog title names a change this section never documents: a `var()` / container-query `width` (e.g. `'var(--w)'`, `'50cqw'`) is no longer clipped, replacing the old literal-string regex that did clip it. §16 lists only two smaller changes (the one-time warning and `width: null`). A user who relied on the old clip for an internal horizontal scrollbar gets no guidance. Add a third bullet with who is affected and how to migrate, matching the treatment the other two get.

wrappers/vue3/test/hotTable.spec.ts
```diff
@@ -521,3 +521,47 @@ describe('Non-HOT based CRUD actions', () => {
...
+    const { rootElement } = testWrapper.getComponent(HotTable).vm.hotInstance;
+
+    expect(rootElement.style.height).toBe('auto');
+    expect(rootElement.style.overflowX).toBe('');
+    expect(rootElement.style.overflowY).toBe('');
+
+    testWrapper.unmount();
+  });
+
+  it('should ignore an unreadable size with a warning and keep the height', async() => {
```
demtario: **high** — This test calls `setProps({ height: 'abc' })` exactly once, then asserts one warning, so it does not prove the cross-commit warning-dedup contract. The React counterpart re-sends the same bad value across three commits (`hotTable.spec.tsx:1256-1258`) to prove the guard fires once across repeated identical values; a Vue wrapper missing that guard entirely would still pass this test, because there is only one apply. Re-send the same `'abc'` two or three times before reading `sizeWarnings` and assert the length stays 1.

handsontable/src/dataMap/metaManager/metaSchema.ts
```diff
@@ -3845,30 +3845,45 @@ export default (): Record<string, unknown> => {
...
      * You can set the `height` option to one of the following:
      *
-     * | Setting                                                                    | Example                    |
-     * | -------------------------------------------------------------------------- | -------------------------- |
-     * | A number of pixels                                                         | `height: 500`              |
-     * | A string with a [CSS unit](https://www.w3schools.com/cssref/css_units.asp) | `height: '75vw'`           |
-     * | `'auto'`                                                                   | `height: 'auto'`           |
-     * | A function that returns a valid number or string                           | `height() { return 500; }` |
+     * | Setting                                                                    | Example                              |
+     * | -------------------------------------------------------------------------- | ------------------------------------ |
```
demtario, line 3849: **medium** — Content-sizing keywords are now rejected but the docs mischaracterize why. The old code wrote any string straight to inline style (`style.width = isNaN(width) ? \`${width}\` : \`${width}px\``), so `max-content`, `fit-content`, `stretch`, `-webkit-fill-available`, `-moz-available`, `none`, `normal` all rendered before and are now ignored — a breaking change. This JSDoc (and the migration guide) describe the rejected set as "a CSS keyword that would collapse the grid," which is false for the content-sizing ones. Name the content-sizing keywords explicitly and give the real clip-safety reason from the resolver (`utils/rootSize.ts:71-77`).

.changelogs/13381.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Unified the grid `width` and `height` options: `height: 'auto'` now writes inline `height: auto` only, so the page (or the nearest scrolling ancestor) scrolls the grid and the rows stay virtualized instead of the grid clipping itself and rendering every row; both options accept a number, a pixel string, a CSS length or expression, `'auto'`, or a function, and ignore an unreadable value or a collapsing keyword with a one-time warning; `null` resets its own axis only (`width: null` no longer writes `nullpx`); and a `var()` or container-query width is no longer clipped.",
```
demtario: **medium** — The `title` is a ~600-character, four-clause, semicolon-chained sentence written at implementation level ("writes inline `height: auto`", "no longer writes `nullpx`"). `bin/changelog` renders `title` verbatim into the version-comparison UI, where a paragraph-length title is a real usability problem. All structural fields are correct. Shorten to one clause naming the user-facing headline (e.g. "`height: 'auto'` no longer clips the grid, so the page or nearest scrolling ancestor scrolls it and rows stay virtualized.") and let the migration guide carry the detail.

docs/content/guides/upgrade-and-migration/migrating-from-18.1-to-19.0/migrating-from-18.1-to-19.0.md
```diff
@@ -696,3 +696,41 @@ columns: [
...
+- the [`EmptyDataState`](@/api/emptyDataState.md) or [`StretchColumns`](@/api/stretchColumns.md)
+  plugins, which read where the grid scrolls
+- code that reads `hot.view.isVerticallyScrollableByWindow()`, which now returns `true` for
+  `height: 'auto'`
+
+### How to migrate
+
+For the common case, nothing. The grid still grows to fit its rows and the page still scrolls it.
+
+To keep an internal vertical scrollbar, set a numeric `height`, or place the grid inside a parent
```
demtario: **medium** — §16 is prose-only. The file's own convention (§15) pairs the workaround with a `js` code block, and this workaround ("set a numeric `height`, or place the grid inside a parent with a fixed height and `overflow: auto`") is exactly the kind that needs a before/after snippet. Add two short blocks under "How to migrate": the old `height: 'auto'` config next to the numeric-height or `overflow: auto`-parent replacement.

docs/content/guides/getting-started/grid-size/grid-size.md
```diff
@@ -356,18 +356,40 @@ const hotSettings = ref({
...
-| `height: <number>` | `height: <n>px; overflow: clip;`                            | The grid itself                                                                | Enabled                     |
-| `height` unset     | None, Handsontable does not touch the root inline styles    | Nearest ancestor with `overflow: auto` or `overflow: hidden`, else the window  | Depends on the scroll parent |
+| Setting                         | Inline styles on root                                       | Scroll parent                                                                  | Row virtualization |
+| ------------------------------- | ----------------------------------------------------------- | ------------------------------------------------------------------------------ | ------------------ |
+| `height: 'auto'`                | `height: auto;`                                             | Nearest ancestor with `overflow: auto` or `overflow: hidden`, else the window  | Enabled            |
+| `height: <number or CSS length>` | `height: <value>; overflow: clip;`                          | The grid itself                                                                | Enabled            |
```
demtario, line 362: **medium** — The page states `width` and `height` accept the same set of values (L375) but then documents different clip semantics for that same set. The height table here says a `height` given as a number or CSS length (which by the page's own list includes `%` and viewport units) gets `overflow: clip` and becomes its own scroll parent, while L391 classifies the width equivalents (`'100%'`, `'80vw'`, `'var(--w)'`) as "relative," not clipping and leaving overflow to the page. As written, `height: '50%'` clips but `width: '50%'` does not, right after the "same set of values" claim. Either add the definite-vs-relative split to the height table, or caveat that clip/scroll semantics differ per axis.

handsontable/src/core/rootSize.ts
```diff
@@ -0,0 +1,374 @@
...
+
+  if (!isFreeHeight(heightState)) {
+    if (!heightRestored) {
+      if (isOwnedOverflow(inlineOverflow(style, 'overflowX'))) {
+        style.overflowX = 'clip';
+      }
+      if (isOwnedOverflow(inlineOverflow(style, 'overflowY'))) {
+        style.overflowY = 'clip';
+      }
+    }
```
demtario, line 280: **low** — Nested grid, definite initial inline height: on `updateSettings({ height: null, width: <definite> })`, `heightRestored` is true, so the sized branch's `if (!heightRestored)` skips overflow writing and the newly set width never gets `overflow-x: clip`. Root instances get a fresh `<div>` (free branch) and never hit this. Write the width's owned overflow even when the height was restored, or scope the `heightRestored` skip to the axis that was actually restored.

handsontable/src/utils/rootSize.ts
```diff
@@ -0,0 +1,217 @@
...
+  kind: RootSizeKind;
+  /**
+   * The value to write to the root's inline style, or `null` when the value is invalid.
+   */
+  cssValue: string | null;
+  /**
+   * `true` when the size resolves against something outside the grid (the container, the viewport,
+   * a custom property), so the grid cannot know the pixel box up front.
+   */
+  isContainerDriven: boolean;
```
demtario: **low** — `isContainerDriven` is computed and exported on `RootSizeResolution`, but no production code reads it: `applyAxis` reads only `.kind`/`.cssValue`, and `applyOverflow` re-derives the classification via `classifyInlineSize`. The two classifiers also disagree on `'auto'`, so the unused field is a latent trap. Drop it, or document it as test-only.

handsontable/src/core/rootSize.ts
```diff
@@ -0,0 +1,374 @@
...
+  if (isOwnedOverflow(inlineOverflow(style, 'overflowY'))) {
+    style.overflowY = '';
+  }
+
+  if (isOwnedOverflow(inlineOverflow(style, 'overflowX'))) {
+    style.overflowX = widthState === 'definite' ? 'clip' : '';
+  }
+}
+
+/**
```
demtario, line 294: **low** — `hasScrollOwnerChanged` JSDoc claims a superset of the old rule and warns that "a missed one leaves a stale scroller," but omits that width is deliberately not read here and is re-resolved by the engine every draw (`Overlays#beforeDraw` / `ScrollSync`). That omission is what made the Bugbot thread read as a live width bug. Add one sentence naming the per-draw re-resolution, as the E2E comment at `width-window-scroll.spec.ts:366-372` already does.

handsontable/src/3rdparty/walkontable/AGENTS.md
```diff
@@ -611,6 +611,20 @@ fields. Three rules follow.
   gates that disagree leave a notch where the frozen columns stop and the frozen rows carry on
   (#10370). The draw cycle positions the bottom overlay before the corner, which is what makes the
   read safe.
+- **A wheel gesture must move each axis exactly once, so the grid scrolls BOTH axes itself and then
+  always consumes the event.** `Overlays#scrollVertically` / `scrollHorizontally` move whatever owns
+  the axis, and for a window owner that means `rootWindow.scrollBy({ behavior: 'instant' })` — the
+  page is scrolled by the grid, not by the browser. So `NativeScrollInput#onWheel` calls
```
demtario: **low** — This line cites `NativeScrollInput#onWheel`; the method is `#onCloneWheel`, and it early-returns at the `getScrollableElement() === rootWindow` test before reaching the translation, so "in every mode" is also false (full window mode is excluded). Per the cite-by-symbol convention, correct the symbol name and the "every mode" qualifier.

demtario (commented):
11 inline notes below.

- **low (test coverage, off-diff)** `tests/e2e/stretch-columns-container-resize.spec.ts` — after the migration of the grow case to a pixel height, "grow the container past the viewport in window-owned mode" is no longer exercised anywhere in this file. The `'auto'`/no-width path now puts the horizontal axis in window mode, and that grow direction (a real horizontal scrollbar appearing against the viewport) is left untested. Consider one case that keeps window mode and grows past the viewport width.

Context: the three unresolved Bugbot threads on `core/rootSize.ts` (L313 width re-pick, L279 restored overflow, L230 invalid sizes) were each re-verified against the current code and are correct as resolved — not re-reported here.

handsontable/src/3rdparty/walkontable/src/overlay/regions/bottomOverlay.ts
budnix: Nice catch — fixed in cce07ef2a7: both horizontal gates now read `inlineStartOverlay.mainTableScrollableElement`, and a reverse-split case in `overlay-scrollbar-clearance.spec.ts` fails without it.

docs/content/guides/upgrade-and-migration/migrating-from-18.1-to-19.0/migrating-from-18.1-to-19.0.md
budnix: Good point — added a third bullet for `var()` and container-query widths in cce07ef2a7, with who is affected and how to migrate.

wrappers/vue3/test/hotTable.spec.ts
budnix: You're right, one `setProps` proved nothing. In cce07ef2a7 two more prop changes make the wrapper re-send `'abc'`, and the test checks it went out 3 times with one warning.

handsontable/src/dataMap/metaManager/metaSchema.ts
budnix: Fair point — the JSDoc, the grid-size guide, and §16 now name all three keyword groups with the real reason for each, in cce07ef2a7.

.changelogs/13381.json
budnix: Agreed — shortened it to the headline in cce07ef2a7 and left the detail to the migration guide.

docs/content/guides/upgrade-and-migration/migrating-from-18.1-to-19.0/migrating-from-18.1-to-19.0.md
budnix: Good spot — §16 now has before/after blocks for a numeric height and for the `overflow: auto` parent, in cce07ef2a7.

docs/content/guides/getting-started/grid-size/grid-size.md
budnix: Yep, my mistake — cce07ef2a7 now says outright that any height other than `'auto'` clips both axes, while a width clips only when it is definite.

handsontable/src/core/rootSize.ts
budnix: Good find — in cce07ef2a7 a width written in the same call now clips beside a restored height, with a unit test for the nested case.

handsontable/src/utils/rootSize.ts
budnix: Makes sense — dropped it in cce07ef2a7; its test checks now go through `classifyInlineSize()`, which production actually uses.

handsontable/src/core/rootSize.ts
budnix: Good call — added the per-draw re-resolution note to the JSDoc in cce07ef2a7.

handsontable/src/3rdparty/walkontable/AGENTS.md
budnix: True — fixed the name to `#onCloneWheel` in cce07ef2a7 and noted that full window mode returns early and leaves the wheel to the browser.

demtario (approved):

handsontable/src/core/rootSize.ts
```diff
@@ -0,0 +1,379 @@
...
+  const relativeExample = axis === 'height' ? '\'75vh\'' : '\'75vw\'';
+
+  warnOnce(
+    rootElement,
+    `invalid-root-size-${axis}-${described}`,
+    `Handsontable: the \`${axis}\` option expects a number of pixels, such as \`500\`, \`'500'\`, ` +
+    `or \`'500px'\`, a CSS length such as \`'50%'\` or \`${relativeExample}\`, or \`'auto'\`. ` +
+    `The value ${described} cannot be read as a size, so it is ignored and the grid's ${axis} ` +
+    'is left as it was.'
+  );
```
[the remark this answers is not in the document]
budnix: Not changing this one: every grid made through `Handsontable()` or `new Handsontable.Core()` gets a fresh `<div>` as `rootElement` (`core.ts:468`), so a new grid on the same container does warn again – checked in jsdom, the count went from 1 to 2. The only grid that reuses its element is the `handsontable` cell editor, which rebuilds its inner grid on the same `htContainer` on every `open()` (`handsontableEditor.ts:110-122`): scoped to the element, a bad `height` there warns once (1 warning over 3 opens and 3 inner grids), while scoped to the instance it would warn on every open. `rootElement` is also the scope the `warnOnce` JSDoc asks for, and the one `Core.invalidCellOption` uses.

Issue #4021, opened 2017-01-25
Row heigths can't be smaller than row header's text height.


KodjoSuprem:
Row heigths can't be smaller than rowheader text's height.

Both with Manual row resize plugin and HT rowHeights option.

Rows header content should be overlapped if needed.

Exemple on Excel:
![image](https://cloud.githubusercontent.com/assets/4376260/22296921/f8309ece-e31b-11e6-8d5c-40fac4fb062f.png)

AMBudnik:
Hi @KodjoSuprem 
How you would like it to look like? Like hidden with an extra second border?

KodjoSuprem:
It is like their is a minimum row size that cannot be overriden, with the rowHeights options.

Take a look at this exemple : http://jsfiddle.net/f75hboxj/
Row n°2 should be extra thin.
But it keeps a minimum size

AMBudnik:
You're right. Even when there are no headers like here: http://jsfiddle.net/handsoncode/1uzs1jj9/ you still will get min height of 22px for a row. And the smallest value you can set is 4px http://jsfiddle.net/handsoncode/9p1nmjdp/
Currently Handsontable does not allow to cut text vertically.

KodjoSuprem:
Ok so can you propose as workaround?
Currently I remove the row header innerHTML rows smaller than 20 px just like you did in your exemple.

But it is not the best solution.

AMBudnik:
So basically you would like this row to be hidden (not deleted)? Can you create a draft (can be a drawing) with a desired look and a index structure that you would like to get? I guess that there's something we can do.

KodjoSuprem:
Basically the minimum row height of 22px is critical.
Tables with headers should accept smallers height 

Exemple:
MS excel allow small row heights, header and row text are half visible (rows overlap)

![image](https://cloud.githubusercontent.com/assets/4376260/22590284/bd430334-ea0e-11e6-8c35-358b72cad13e.png)

For legal reasons document layouts have to be very specific so row size matter a lot for our clients. A lot of official reports use empty lines as separators. Those lines have a height smaller than 22px.

AMBudnik:
Hi @KodjoSuprem 

thank you for additional information.

I tried to find a workaround however the `beforeRowResize` works exactly as `beforeColumnResize` with the reported bug: https://github.com/handsontable/handsontable/issues/3328

I will mark this issue as a request for a change.

balmychan:
👍

[...]

PR #13382, opened 2026-09-04
DEV-2791: Add an exact row-height mode to Walkontable

budnix:
### Context

The PR adds an exact row-height mode to the rendering engine (Walkontable). It is part 1 of 2 for GitHub issue #4021 (a row cannot be shorter than its content, because `rowHeights` works as a minimum).

A provided row height has always been a floor. Four things hold a row up, and the exact mode has to defeat each one:

- **The engine measures the rendered rows and keeps the larger value.** `RowUtils#isExact` decides per row whether the provided height is a floor or the final height. `markOversizedRows` skips exact rows before the geometry read, and a uniform exact band skips the measurement walk outright. `RowUtils` returns the provided height without the `Math.max` against the oversized records, and an overlay that reports no height for an exact row falls back to the row's own height, so no overlay renders it differently from what the row-height cache holds.
- **The height is written to one cell only.** `render/exactRowHeight.ts` marks the row with `htExactRow` and the stylesheet releases the cells' minimum height through that class, so the height still goes on a single cell — the first that spans one row and is rendered. A merged cell never carries it (the span sizes it), and a cell MergeCells covers is `display: none`, so it is skipped too. When no cell can carry it, the height goes on the `tr` instead. The marker sits on the row because the cell renderers reset a painted cell's class and style, and a per-cell marker cost +2.2 ms of style recalculation per draw on 462 cells.
- **A table cell's `height` is a minimum in CSS.** `overflow: hidden` on the cell does not shrink it, and a `height: 100%` child resolves to the content height. The only thing that works is taking the content out of flow: each data cell's content moves into an absolutely positioned `div.htCellClip` that clips, and the cell drops its own padding (a border-box cell cannot be shorter than its padding plus border — 17px in `horizon`) while the wrapper's insets carry it. Row headers keep their `.relative` wrapper and the header's own `span.rowHeader` clips, so the active-row accent bar keeps the 1px overhangs it needs to meet the gridlines. `htMiddle`/`htBottom` map onto the wrapper with `align-content`, never `display: flex`, because the indicator renderers float their arrows.
- **The renderers rewrite the cell.** `getCellContentRoot(TD)` is new in the DOM helpers. `fastInnerText`, `fastInnerHTML`, and the checkbox, autocomplete and multiSelect renderers write through it, so the wrapper survives a redraw and no in-flow content lands beside it (which would grow the row back).

**The PR is inert.** The mode defaults to `'min'` and Handsontable never sets it, so no DOM, CSS or behavior changes for any user. Part 2 (separate task, breaking) flips `rowHeights` to the exact mode while `minRowHeights` keeps the floor semantics, and updates ManualRowResize, AutoRowSize, mergeCells, the Formulas hyperlink renderer, the editor, docs and the migration guide.

### Performance

The mode must cost nothing for the grids that never use it, so the cost is measured rather than argued.

Every loop over rows asks `RowUtils#mayHaveExactRows()` once per draw and skips the per-row probe when it is `false`. Without that, a default grid paid one settings read per rendered row, per table, per draw:

| Default grid, 30 rendered rows | Mode reads per draw |
|---|---|
| Per-row probe | 61 |
| Asked once per draw | 3 |

Draw time against the pre-change bundle, on a 100k-row grid with a 570-cell viewport, medians of three rounds:

| Bundle | Full draw | Scroll step |
|---|---|---|
| Before this change | 2.22 ms | 10.20 ms |
| This branch | 2.24 ms | 10.05 ms |
| This branch, exact mode | 2.23 ms | 10.47 ms |

The row-height hook chain runs exactly as often as before: 3 `modifyRowHeight` calls per rendered row in `min` mode, the same as the bundle without the mode, and 5 in `exact` mode. What remains for every grid is one property read plus an integer compare per cell in the text writer, which exits at once for a plain text cell. A custom renderer that wipes its cell every draw pays one wrapper rebuild per draw, on exact rows only.

The mode also survives `renderMode: 'onChange'`: the height pass runs for every rendered row whatever the cell painter decided, re-applying is idempotent, and the painter reads its own stamps without inspecting cell DOM.

- Unit tests added/modified (`*.unit.js`):
  - `handsontable/src/3rdparty/walkontable/test/unit/axisSizing/rowHeightMode.unit.ts` (new, 19 tests)
  - `handsontable/src/3rdparty/walkontable/test/unit/renderer/exactRowHeight.unit.ts` (new, 22 tests)
  - `handsontable/src/helpers/dom/__tests__/element.unit.ts` (`fastInnerText` wrapper lane, `getCellContentRoot`)
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`): `tests/e2e/walkontable/exact-row-heights.spec.ts` (new, 10 tests on all six theme × bundle legs) with `tests/fixtures/demo/walkontable/exact-row-heights.html` and `tests/fixtures/pages/walkontable/ExactRowHeightsPage.ts`. The fixture flips the mode on the engine directly. The `min`-mode test is the precondition that a short provided height alone does not shrink a row; the option-only variant asserts the engine sees a uniform size source AND a uniform mode, which is the only path through the uniform-band shortcut and the row-height cache's uniform strategy.
- Type tests (`*.types.ts`) updated if public API changed: none — no public Handsontable API changed.
- For a bug fix — the spec that fails without this fix: n/a (inert engine feature; the exact-mode tests fail without the change).
- Demo page / recorded trace (for UI changes): none committed (`handsontable/dev-pr.html` is a local, gitignored demo with a mode toggle).

Docs: `handsontable/src/3rdparty/walkontable/AGENTS.md` gains an "Exact row heights" section, `handsontable/AGENTS.md` a renderer pitfall, and the `handsontable-renderer-dev` skill two rules.

Changelog: `.changelogs/4021.json` (`fixed`, breaking, public issue #4021) describes the whole `rowHeights` change this and part 2 deliver together, so part 2 adds no second entry.

Two things for the reviewers to decide:

1. `htExactRow` (on the row) and `htCellClip` (the wrapper) become public CSS surface once part 2 lands, and the wrapper changes what a custom renderer or a user stylesheet sees inside a cell of an exact row (`TD.firstChild` is the wrapper; `td > .x` selectors stop matching). That belongs in part 2's migration guide next to the clipping itself.
2. `getCellContentRoot` and `CELL_CLIP_CLASS` are exported on `Handsontable.dom` by the full build (custom renderers will need the function; the constant could stay internal).

### Commands run

```
npm --prefix handsontable run test:unit          → Test Suites: 351 passed, Tests: 4511 passed
npm --prefix handsontable run test:walkontable   → 816 specs, 0 failures
cd tests && npx playwright test                  → 5446 passed, 6 skipped (full theme x bundle matrix)
cd tests && npx playwright test e2e/walkontable/exact-row-heights.spec.ts → 60 passed
npm --prefix handsontable run eslint / stylelint / test:types → clean
```

The unit run reports one failure in `Core#updateSettings with external data source`, a DataProvider test that touches nothing here and passes on its own — it times out under parallel load.

### Types of changes

### Related issue(s):
1. DEV-2791
2. #4021 (part 1 of 2; the issue closes with part 2)

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:
- [ ] MANUAL QA NEEDED — 

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2791


qunabu:
Task linked: [DEV-2791 Walkontable: exact row-height mode (issue #4021, part 1 of 2)](https://app.clickup.com/t/9015210959/DEV-2791)

demtario (approved):

PR #13383, opened 2026-09-04
DEV-2629: Key the perf baseline on browser and harness, name the run shift

demtario:
### Context

The PR changes how the performance suite picks and describes the baseline a pull request is compared against, so the sticky comment stops reporting shifts that come from the CI environment as if they came from the code under review.

Replaying the 42 marks-valid develop goldens on gh-pages (2026-08-27 to 2026-09-04) showed two mechanisms behind the "−20% on one run, +20% on the rerun" reports, neither of them the PR under test:

1. **Develop-side step changes leak through the median window.** The Playwright 1.58 → 1.62 bump changed the Chromium build and shifted initial-load by −18% and sorting by −20% in a single develop push; filtering moved +49% one push later. The median-of-5 baseline carried the old browser's numbers for five more pushes, so unrelated PRs got Filtering +25% to +45% red callouts. Nothing recorded which browser had produced a golden, so nothing could refuse the comparison.
2. **A per-run CI runner speed factor moves every scenario together.** It spans 0.63x to 1.12x across develop goldens; removing it takes the scroll scenarios' run-to-run CV from ~13% to ~3%. It is what makes a whole table read −25% green or +12% yellow.

A third gap made the first one invisible where it happened: a golden (develop-push) run never compared against history, so a develop regression surfaced only on the next five PRs.

What ships:

- **Provenance.** `lib/setup.mjs` (Playwright `globalSetup`) records the Chromium build, CPU model and count, memory, platform and GitHub runner image into `output/environment.json`. The teardown stamps `environment` and `harnessVersion` on every snapshot; both report footers print them; the gh-pages history index gains commit, Chromium and CPU columns.
- **Baseline compatibility key** (`lib/environment.mjs`): `{ chromium, harnessVersion }` per snapshot and `measurementVersion` per scenario. The median draws only on goldens with the same key, and within a scenario only on entries at the same version. Goldens recorded before provenance existed are excluded by the absence of the fields, the way `windowSource` already excludes pre-marks goldens. With fewer than two compatible goldens the comment says "no comparable baseline" and names what changed (`Chromium 138.0.7204.23 -> 140.0.7339.16`), instead of publishing deltas against the wrong browser. `HARNESS_VERSION` lives in `trace-runner.mjs`; bump it when the runner changes what a window contains. Bump a scenario's `measurementVersion` when only that spec's definition changes. This is the discriminator the earlier follow-up notes said did not exist for a scenario redefinition.
- **Golden run self-comparison.** The workflow fetches history in golden mode too. The develop run's report goes to the job summary and each regressed scenario becomes a `::warning` annotation on that run. The saved snapshot is untouched: the teardown saves before it loads, and the single-file fallback is disallowed in golden mode, which also closes a latent case where a golden run compared against the file it had just written.
- **Run shift.** A `Δ vs shift` column and a `Run shift` footer line name the median delta across scenarios and remove it per row as a ratio. Informational only: callouts still fire on the raw `Δ Total`, because a change that slows every scenario alike is exactly what a median across scenarios cannot see. Callout prose and the `::warning` annotations carry the same "+80% relative to this run's shift of −20%" context.
- **Heap threshold per scenario.** `heapThresholdFor(name)`: 10 for scroll-left and scroll-right (their peak heap is GC timing across 500 wheel renders, run-to-run CV 4.4 to 6.3%), 5 elsewhere. Both builders read it; the replay script prints a per-scenario heap table.
- `scripts/replay-goldens.mjs` groups goldens by key before replaying, so a replay across a Chromium change never medians two environments together.

Not in this PR (planned as a follow-up under the same task): GC between iterations and more iterations for the short scenarios, and a post-GC heap field. Those change measured numbers and rely on the `HARNESS_VERSION` mechanism introduced here.

**Operational note.** After merge, PR comments read "no comparable baseline" until two develop pushes on a perf-scoped path have recorded goldens with provenance. Expected and stated in the comment itself; not a tooling regression. The replay against today's goldens is unchanged (all 42 sit in one "no provenance" group), so `REGRESSION_CALLOUT_THRESHOLD_TIMING = 15` stays; re-run `node performance-tests/scripts/replay-goldens.mjs --since <date>` once a week of keyed goldens exists.

### How has this been tested?

- `npm run test:tooling` — 453 pass (performance-tests suite 220, of which 61 new: `environment.test.mjs`, `report-builder-provenance.test.mjs`, `html-report-builder-provenance.test.mjs`, plus new cases in `median-snapshot`, `snapshot-store` and `thresholds`).
- `cd performance-tests && npm run lint && npm run typecheck` — clean.
- End to end on this machine with Handsontable built from the branch, `npx playwright test --grep "scroll-right|scroll-down"`:
  - compare mode against the real gh-pages history: legacy goldens refused, footer reads "the develop goldens were recorded on a different environment (Chromium unknown -> 151.0.7922.34, harness unversioned -> 1); deltas resume once two develop pushes have run on the current one";
  - compare mode against history stamped with this machine's provenance: median of 5 selected, footer names the baseline Chromium and the current environment, deltas render;
  - golden mode: snapshot carries `harnessVersion: 1` and the environment block; no self-comparison against the just-saved file.
- Two independent automated review passes (general code review and the Handsontable reviewer): no critical or high findings; the two medium findings (shift context on the `::warning` annotations, no-baseline reason missing from the golden-mode footer) are fixed and covered by tests.

### Types of changes

### Related issue(s):

1. DEV-2629

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

Tooling only: `performance-tests/`, `.github/workflows/performance-tests.yml`, the `performance-testing` skill, `performance-tests/README.md`, one row in `AGENTS.md`. No shipped package changes, so no changelog entry.

### Checklist:

---

> [!NOTE]
> **Low Risk**
> Changes are confined to performance-tests tooling and CI; no production grid or wrapper code paths are modified.
> 
> **Overview**
> The performance suite now **records run provenance** (Chromium build, CPU, runner image via Playwright `globalSetup`) and **only medians develop goldens that match** the current `{ chromium, harnessVersion }` key and each scenario's `measurementVersion`. Incompatible or legacy goldens are **refused with an explicit reason** instead of producing misleading PR deltas after Playwright/Chromium bumps.
> 
> **Reports and CI** add a **`Δ vs shift`** column and **run shift** footer (median across scenarios, informational only—regression callouts still use raw `Δ Total`), per-scenario **heap thresholds** for horizontal scroll scenarios, and **develop golden runs** that compare against trailing history (job summary + `::warning` per regressed scenario) without rewriting the saved snapshot. The workflow fetches golden history in both modes, always ensures `result.md`, and documents the behavior in the performance-testing skill and README.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d83b025313aa5024bb4a3cba80d7a832c74c3f71. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:

performance-tests/lib/teardown.mjs
```diff
@@ -193,28 +289,60 @@ export default async function teardown() {
...
+  const versionOf = scenario => scenario.measurementVersion ?? DEFAULT_MEASUREMENT_VERSION;
+  // A scenario redefined since the baseline was recorded (another `measurementVersion`) is the same
+  // situation as a window mismatch -- the two sides measured different quantities under one name --
+  // and is withheld through the same path. The median already filters these out per scenario; this
+  // catches the single-file fallback, which is one whole golden and cannot.
   const crossWindow = (current, baseline) => Object.keys(current)
-    .filter(name => baseline?.[name] && windowSourceOf(baseline[name]) !== windowSourceOf(current[name]));
+    .filter(name => baseline?.[name] && (
+      windowSourceOf(baseline[name]) !== windowSourceOf(current[name])
+      || versionOf(baseline[name]) !== versionOf(current[name])
```
budnix: This filter now fires for two different reasons, but `comparability()` still knows only one of them. A scenario withheld because its `measurementVersion` changed gets the `window mismatch` label and the note "the two sides were measured over different trace windows", and both sides actually used marks, so both statements are false. Giving `comparability()` its own `version-mismatch` reason would name the real cause instead of sending someone to look for a window bug that isn't there.

performance-tests/lib/environment.mjs
```diff
@@ -0,0 +1,205 @@
...
+  if (current.harnessVersion !== baseline.harnessVersion) {
+    differs.push(
+      `harness ${baseline.harnessVersion ?? 'unversioned'} -> ${current.harnessVersion ?? 'unversioned'}`
+    );
+  }
+
+  const change = differs.length > 0 ? differs.join(', ') : 'provenance differs';
+
+  return `the develop goldens were recorded on a different environment (${change}); `
+    + 'deltas resume once two develop pushes have run on the current one';
```
budnix: I traced this with one compatible golden in `history/` plus a compatible `latest.json`: the median is refused, the single-file fallback passes the key check, and deltas render against that one run. So deltas resume on the first develop push after the change, not the second, and only the median needs two. Worth softening the wording so the wait isn't overstated.
```suggestion
    + 'deltas resume with the next develop push, and as a median once two have run';
```

performance-tests/lib/snapshot-store.mjs
```diff
@@ -109,17 +219,46 @@ async function loadMedianFromHistory(historyDir, goldenPathExists) {
...
+    }
 
-  if (!median && snapshots.length > 0) {
+    return { snapshot: median, incompatibleExample: null, versionsMismatch: false };
+  }
+
+  const keyed = compatibleWith && isCompleteKey(compatibleWith.key);
+  // Enough goldens matched the key, yet no median came out: the per-scenario version filter
+  // emptied it. That is a different message from "wrong browser", and it must not be silent.
+  const versionsMismatch = !!keyed
```
budnix: This infers "the scenarios were redefined" from a count, and the count has a second way of getting there: two compatible goldens with different scenario sets also empty the median, and then the comment blames a redefinition that never happened. The mirror case is line 242, where the `incompatibleExample` filter runs over all snapshots rather than the marks-valid ones, so a history where most goldens failed their marks reads as a browser change. Both messages send a maintainer to the wrong place. Having `computeMedianSnapshot` report why it refused would remove the guessing.

performance-tests/lib/snapshot-store.mjs
```diff
@@ -109,17 +219,46 @@ async function loadMedianFromHistory(historyDir, goldenPathExists) {
...
+    && snapshots.filter(s => isValidForMedian(s) && isCompatibleBaseline(s, compatibleWith.key)).length
+      >= MIN_VALID_SNAPSHOTS;
+  // The newest golden that was refused for its key, so the report can name what changed.
+  const incompatibleExample = keyed && !versionsMismatch
+    ? snapshots
+      .filter(s => !isCompatibleBaseline(s, compatibleWith.key))
+      .sort((a, b) => Date.parse(b.timestamp || 0) - Date.parse(a.timestamp || 0))[0] ?? null
+    : null;
+
+  if (snapshots.length > 0) {
```
budnix: When the history fetch produces no files at all, this warning never runs and `reasonFor` also returns null, because there is no incompatible golden to point at. Golden mode has no single-file fallback and no self-compare, so the job summary shows raw numbers with nothing saying the comparison did not happen — the silent case this PR is closing. A `snapshots.length === 0` reason would cover it, and I don't see a test on that path yet.

performance-tests/lib/snapshot-store.mjs
```diff
@@ -44,34 +59,127 @@ export async function saveSnapshots(scenarioResults, metadata = {}, goldenDir =
...
+
+  return Object.entries(snapshot.scenarios || {}).some(([name, entry]) => {
+    if (!(name in scenarioVersions)) {
+      // A scenario this run did not measure has no version to disagree with; it is neither a match
+      // nor a mismatch, so it does not decide the answer.
+      return false;
+    }
+
+    return (entry?.measurementVersion ?? DEFAULT_MEASUREMENT_VERSION) === scenarioVersions[name];
+  }) || Object.keys(snapshot.scenarios || {}).every(name => !(name in scenarioVersions));
```
budnix: The trailing `every` is true for an empty array, so a `latest.json` cut short by a failed deploy (`scenarios: {}` with the key fields still present) comes back as the baseline. In the teardown `golden` is then truthy, the self-compare branch is skipped, `hasGolden` is false, and the footer stays empty because `baselineUnavailable` is null, so the comment carries raw numbers with no baseline line at all. Refusing an empty `scenarios` object separately from the disjoint-names case this clause is written for would close it.

performance-tests/lib/snapshot-store.mjs
```diff
@@ -44,34 +59,127 @@ export async function saveSnapshots(scenarioResults, metadata = {}, goldenDir =
...
+}
+
+/**
+ * Backwards-compatible shape: the snapshot alone, or null.
+ *
+ * @param {string} [goldenDir]
+ * @param {object} [options] -- as for loadBaseline
+ * @returns {Promise<object | null>}
+ */
+export async function loadSnapshots(goldenDir = DEFAULT_GOLDEN_DIR, options = {}) {
```
budnix: Did this end up with a caller? The teardown imports `loadBaseline`, and the only uses of `loadSnapshots` I can find are in `snapshot-store.test.mjs`. If it exists only for the tests, saying that in the JSDoc would stop the next reader from treating it as a supported second entry point.

performance-tests/lib/median-snapshot.mjs
```diff
@@ -109,6 +110,10 @@ function medianScenario(entries) {
   const hookTimingValues = entries.map(entry => entry.hookTiming).filter(v => v !== undefined && v !== null);
 
   return {
+    // When computeMedianSnapshot was given a key it filtered the entries to one version, so the
+    // first one's is the median's. Without a key (the replay's no-provenance group) the entries
+    // may mix versions and this is only the newest entry's; no consumer reads it off a median today.
```
budnix: The teardown does read this off a median — `versionOf(baseline[name])` in its `crossWindow` predicate. I checked it with three unkeyed goldens, newest at version 2 and the older two at version 1: the median mixes both definitions, reports version 2, matches the current run, and nothing gets withheld. That is the silent mixing the key exists to stop, so this line reads as a promise that no longer holds.

performance-tests/lib/report-builder.mjs
```diff
@@ -57,14 +67,51 @@ export function buildReport(allScenarioResults, goldenSnapshots, meta = {}) {
...
+ * consumer can say how far a row sits from the rest of the run the way the callout prose does.
+ *
+ * @param {Record<string, object>} allScenarioResults
+ * @param {object | null} goldenSnapshots
+ * @param {object} [meta]
+ * @returns {Array<{ name: string, title: string, totalPct: number | null, heapPct: number | null,
+ *   timingRegressed: boolean, heapRegressed: boolean, shift: number | null,
+ *   relativePct: number | null }>}
+ */
+export function collectRegressions(allScenarioResults, goldenSnapshots, meta = {}) {
```
budnix: The teardown calls this right after `buildReport` with the same three arguments, so `assessScenarios` and `runShift` both run twice over the same data. `assessScenarios`'s own doc comment says that deciding this per render site is how the table and the callouts once disagreed on one run, and this brings a second copy of that decision back. Returning the assessed rows from `buildReport` and letting the annotations read them would keep it to one.

performance-tests/lib/report-builder.mjs
```diff
@@ -142,63 +232,66 @@ function reliabilityCell(current, golden) {
   return `${fmtCvValue(calcCv(iterationTotals))} / ${fmtCvValue(golden?.spread)}`;
 }
 
-function buildSummaryTable(results, goldenScenarios, hasGolden, crossWindow) {
+function buildSummaryTable(assessments, hasGolden, shift) {
   const headers = hasGolden
     ? [
-      'Scenario', 'Scripting', 'Rendering', 'Painting', 'Total', '\u0394 Total',
-      'CV run / base', 'JS Heap', '\u0394 Heap',
+      'Scenario', 'Scripting', 'Rendering', 'Painting', 'Total', 'Δ Total', 'Δ vs shift',
```
budnix: The header adds this column whenever there is a golden, but the legend below it is only added when `shift != null`. On a self-comparison, or on a run where fewer than five scenarios produced a comparable delta, every cell in the column reads `--` with nothing explaining what it is. Building the header from the same condition as the legend would keep the two in step.

performance-tests/lib/html-report-builder.mjs
```diff
@@ -955,6 +1014,14 @@ function buildScript() {
       right.appendChild(spread);
     }
 
+    // The delta with the run's common shift removed, beside the raw one the badge carries. Neutral
+    // styling on purpose: it informs the reading of the badge, it does not compete with it.
+    if (data.hasBaseline && scenario.totalChangeVsShift != null && !scenario.badgeIsHeap) {
```
budnix: This gate on `!scenario.badgeIsHeap` has no counterpart in the markdown builder — `report-builder.mjs` fills its `Δ vs shift` cell for every row. So a scenario that regressed only on heap gets shift-relative context in the sticky comment and none in the linked report. Since the rule stated in this file is that the two renderings must not present the same row differently, it would be good to pick one side and apply it to both.

performance-tests/lib/html-report-builder.mjs
```diff
@@ -1229,11 +1296,12 @@ function buildScript() {
...
         );
-        // Memory is banded on the heap threshold, which is an order of magnitude tighter than the
-        // timing one because heap barely moves run to run.
+        // Memory is banded on the scenario's heap threshold, which is an order of magnitude tighter
+        // than the timing one because heap barely moves run to run -- except on the scenarios whose
+        // peak heap depends on GC timing, which carry a wider band (heapThresholdFor).
         changeTd.style.color = row.incomplete
           ? statusColor('neutral')
-          : statusColor(classifyChangeCss(row.change, data.thresholds.heap));
+          : statusColor(classifyChangeCss(row.change, scenario.heapThreshold));
```
budnix, line 1311: Now that this reads `scenario.heapThreshold`, nothing in `buildScript()` reads `data.thresholds.heap` any more, yet it is still serialized at line 246 and `REGRESSION_CALLOUT_THRESHOLD_HEAP` is imported only to fill it. The next person adding a heap band will reach for `data.thresholds.heap`, get 5 for `scroll-left` and `scroll-right`, and paint red on the GC noise the override exists to tolerate. Either drop the field or label it as the shared default only.

performance-tests/lib/build-history-index.mjs
```diff
@@ -24,19 +27,59 @@ const runs = entries
...
+const rows = [];
+
+for (const name of runs) {
   // Parse timestamp back to readable format
   // 2026-04-08T10-15-30Z → 2026-04-08 10:15:30 UTC
   const readable = name
     .replace('Z', '')
     .replace('T', ' ')
     .replace(/ (\d{2})-(\d{2})-(\d{2})$/, ' $1:$2:$3 UTC');
+  const { commit, chromium, cpu } = await provenanceOf(name);
```
budnix: Each call parses a whole snapshot to pull three short strings, and the loop awaits one before starting the next. `performance-reports/develop/` gains a directory per develop push and nothing prunes it, so this grows into a few hundred sequential reads inside the deploy step of every push. `await Promise.all(runs.map(provenanceOf))` keeps the same output and drops the serial part.

performance-tests/lib/build-history-index.mjs
```diff
@@ -24,19 +27,59 @@ const runs = entries
   .sort()
   .reverse(); // newest first
 
-const rows = runs.map((name) => {
+/**
+ * @param {string} value
+ * @returns {string}
+ */
+function escapeHtml(value) {
```
budnix: This is the same function as `escapeHtml` in `html-report-builder.mjs`, same four replaces, and neither one escapes `'`. The two now have to change together, so adding the apostrophe in one and not the other would leave one of the generated pages behind. Both files already import from `lib/`, so this could sit next to `exists()` in a shared helper.

.github/workflows/performance-tests.yml
```diff
@@ -135,6 +140,15 @@ jobs:
           retention-days: 14
           if-no-files-found: warn
 
+      # Token-free, both modes: on the develop push this is where the run's comparison against
+      # the trailing median is read (the sticky comment below is pull-request only).
+      - name: Write the performance summary to the job summary
+        if: always()
+        run: |
+          if [ -f performance-tests/output/result.md ]; then
```
budnix: The composite has an "Ensure PR comment body exists" step in front of its equivalent, which writes a placeholder when `result.md` is missing, so a run that dies before the teardown still says "No result.md was generated. Check the workflow log". Here the file check just skips, so the develop run gets an empty job summary in exactly the case where the summary matters most. Reusing the composite's shape would keep the two paths saying the same thing.

.claude/skills/performance-testing/SKILL.md
```diff
@@ -227,11 +232,21 @@ Build artifacts (`handsontable/dist/` and `handsontable/styles/`) must exist bef
...
+
+- **Provenance.** `lib/setup.mjs` (Playwright `globalSetup`) records the Chromium build (`browser.version()`), CPU model and count, memory, platform and the GitHub runner image into `output/environment.json`. The teardown stamps it on the snapshot (`environment`, `harnessVersion`) and both reports print it in their footer.
+- **Compatibility key.** `{ chromium, harnessVersion }` at the snapshot level, `measurementVersion` per scenario. `computeMedianSnapshot(..., { compatibleWith })` draws only on goldens with the same key, and within a scenario only on entries at the same `measurementVersion`. A golden recorded before provenance existed has no key and is excluded by the absence of the fields, like a pre-marks golden is by `windowSource`. With fewer than 2 compatible goldens the comment says "no comparable baseline" and why (`describeKeyMismatch`), and deltas resume two develop pushes later. **Bump `HARNESS_VERSION` in `trace-runner.mjs`** when the runner changes what a window contains (settle, GC between iterations, work moved in or out); **bump a scenario's `measurementVersion`** when only that spec's definition changes.
```
budnix: This says the comment reports "no comparable baseline" with fewer than two compatible goldens, but with exactly one it falls through to `latest.json`, which is that same golden, and deltas render against a single run. I called `loadBaseline` that way and got the single snapshot back with `unavailableReason: null`. The same off-by-one is in `README.md` line 51 and in `describeKeyMismatch`, so all three want the milder wording.

performance-tests/README.md
```diff
@@ -43,12 +43,16 @@ PERF_MODE=golden node scripts/run.mjs
...
+The suite measures whatever Chromium the `@playwright/test` pin ships, and every
+snapshot records which one (`environment.chromium`, captured by the Playwright
+`globalSetup` in `lib/setup.mjs`). The baseline is a median over develop goldens
+with the **same Chromium build and harness version** only, so after a Playwright
+bump the first PR runs report "no comparable baseline" with the reason, and deltas
+resume once two develop pushes have run on the new engine. The comment's `Δ vs
+shift` column and `Run shift` footer name how much faster or slower the CI runner
+was than the baseline's, which every scenario shares; callouts still fire on the
+raw delta. A golden (develop-push) run is compared against the trailing median
+too, and annotates its own run with a `::warning` per regressed scenario.
```
budnix: The new `::warning` annotations are documented here, but not how often they will be noise. `thresholds.mjs` records that 3% of no-change comparisons fire a callout at the 15% threshold, and the suite runs nine scenarios per push, so roughly one develop push in four should carry a warning that measures runner speed rather than code. One sentence naming that rate and pointing at `scripts/replay-goldens.mjs` would stop the first meaningless one from being read as a real regression.

performance-tests/lib/teardown.mjs
demtario: Done in d83b025. `comparability()` now takes `'window-mismatch' | 'version-mismatch'` (the boolean still means the window one), the teardown classifies the two separately (`meta.versionMismatchScenarios`), and a redefined scenario reads "measurement changed" with "the scenario was redefined since the baseline was recorded (different measurementVersion)". Heap is withheld under the same label. Tests in thresholds and report-builder-provenance.

performance-tests/lib/environment.mjs
demtario: Done. The wording is one constant now, `RESUME_NOTE` in environment.mjs: "deltas resume with the next develop push, and as a median once two have run". Every refusal appends it; SKILL.md and the README say the same.

performance-tests/lib/snapshot-store.mjs
demtario: Done. `explainMedianRefusal()` in median-snapshot.mjs returns a discriminated reason next to the selection it explains: `empty-history`, `no-marks-valid`, `incompatible-key` (with the newest marks-valid incompatible golden as the example, so a history of failed marks is not read as a browser change), `version-mismatch`, `disjoint-scenarios`. snapshot-store maps them to text through `BASELINE_REFUSALS` and infers nothing from counts. One test per branch, including the two you named.

performance-tests/lib/snapshot-store.mjs
demtario: Done. `empty-history` is a stated refusal now, and it reaches the golden-mode footer through the no-baseline path. Test: "an empty history directory is a stated refusal, not a silent absence".

performance-tests/lib/snapshot-store.mjs
demtario: Done. `loadBaseline` refuses a golden with `scenarios: {}` on its own (`empty-golden` reason) before the key and version checks, so it can never come back truthy and empty. Test added; the disjoint-names clause is unchanged.

performance-tests/lib/snapshot-store.mjs
demtario: Only the tests call it. The JSDoc now says so: kept for the existing tests and any caller that predates `loadBaseline`; new code calls `loadBaseline` and reads the reason with the snapshot.

performance-tests/lib/median-snapshot.mjs
demtario: Done. `computeMedianSnapshot` filters entries to a single version in the unkeyed case too, the newest golden's, so a median never mixes two definitions and the version it reports is the one it measured. Test: "without a key, a scenario is still medianed at a single version". It also reports `scenarioEntryCounts`, so a per-scenario shortfall against `medianWindowSize` is visible.

performance-tests/lib/report-builder.mjs
demtario: Done. `assessReport()` decides once; `buildReport` and `collectRegressions` take the assessment as a parameter and the teardown passes the same object to both. A test pins that rendering from a precomputed assessment equals computing its own.

performance-tests/lib/report-builder.mjs
demtario: Done. The header adds the column on the same condition as the legend (`hasGolden && shift != null`). Tests assert it is absent on a self-comparison and on a table with fewer than five comparable rows.

performance-tests/lib/html-report-builder.mjs
demtario: Done. The `!scenario.badgeIsHeap` gate is gone; the HTML shows the shift-relative figure for every row that has one, as the markdown does.

performance-tests/lib/html-report-builder.mjs
demtario: Done. `thresholds.heap` is dropped from the payload along with the import; the test now asserts `heap` is absent from `thresholds` and that every scenario carries `heapThreshold`.

performance-tests/lib/build-history-index.mjs
demtario: Done, `Promise.all` over the runs.

performance-tests/lib/build-history-index.mjs
demtario: Done. `lib/html-utils.mjs` is the one escaper, apostrophe included, and both pages import it. Test added.

.github/workflows/performance-tests.yml
demtario: Done. "Ensure the report body exists" runs in both modes ahead of the job summary with the composite's placeholder text, and the pull_request-only copy is folded into it.

.claude/skills/performance-testing/SKILL.md
demtario: Done in SKILL.md, README.md and `RESUME_NOTE`: with no compatible golden the comment says "no comparable baseline" and why; with exactly one, the single-file fallback serves and the footer reads "single develop run"; the median needs two.

performance-tests/README.md
demtario: Done. One paragraph with the replay numbers: across 42 develop goldens 8% of no-change rows crossed the 15% threshold, so with nine rows per push roughly two develop pushes in five carry a warning that measured the runner, read against `Run shift`, re-derive with `scripts/replay-goldens.mjs`. I used the current replay's 8% rather than the 3% in thresholds.mjs, which was measured on 19 goldens.

budnix (approved):

PR #13384, opened 2026-09-04
DEV-2792: Add the renderMode option and a diffing selection pass

budnix:
### Context

The PR adds the `renderMode` option, which lets a grid, a column, or a cell skip repainting cells that did not change since their last render, together with the `markCellChanged()` and `markAllCellsChanged()` methods, and makes the selection pass update only the cells whose selection state changed. It comes from the GitHub issue #9614 request ("renderer cell only once"): with a slow custom renderer, every `setDataAtCell` repainted the whole rendered band.

`renderMode` cascades (`cell` → `column` → `global`) and defaults to `'always'`, today's behavior. Under `'onChange'` a cell is painted only when the element it lands in showed something else after its last paint: another cell (scroll), another value, another renderer, a changed cell meta, or a structural change (sort, filter, hide, insert, remove, data reload, settings update). What the engine cannot see is documented as the option's limit: a value object mutated in place, and state outside the grid that a renderer reads. `renderMode: 'always'` on such cells, or `markCellChanged()` before the render, covers it.

How it works:

- `src/core/incrementalRender/` – `RenderChangeTracker` (one epoch per instance), `cellPaintStamps` (a `WeakMap` of what each TD showed after its last paint), `CellPainter` (the per-cell decision and the paint, extracted from `TableView`). The per-cell version is an own property on the meta object, written by every meta write, the validation writes, and `markCellChanged()`; the dynamic extension goes through a new non-tracking `CellMeta#extendMeta`.
- Walkontable asks a new `shouldPaintCell` setting before resetting and painting a cell element; the default answers `true`.
- The selection pass now diffs against what each element carried after the previous pass (`selection/appliedSelection.ts`) and caches the cell-range scan per layer and overlay (`selection/scanCache.ts`). Both are always on. On 40000 rendered cells, a select-all followed by a full draw went from ~1000 ms to ~200 ms with skipping, and a selection move on a fast draw from 335 ms to 48 ms in every mode.
- Search, MergeCells, and Formulas mark the cells whose DOM they write from their own state.

### Measurements

Old = the `develop` bundle, new = this branch after the merge with `develop`. Chrome, a 10000×60 grid in a 1800×1000 viewport, fixed column widths, one frozen row and column, `columnSorting` on; a full draw paints 864 cells. All six configurations ran interleaved in one tab, twice, 20 iterations each, best median. Values are milliseconds of script per operation.

**Text renderer**

| scenario | old (ms) | new, `always` (ms) | new, `onChange` (ms) | cells painted, old → `onChange` | speedup, old → `onChange` |
|---|---|---|---|---|---|
| `render()`, nothing changed | 4.5 | 4.4 | 2.5 | 864 → 0 | 1.8× |
| `setDataAtCell`, 1 cell | 4.7 | 4.5 | 2.9 | 864 → 1 | 1.6× |
| `setDataAtCell`, 50 cells | 4.9 | 5.1 | 3.5 | 864 → 50 | 1.4× |
| `setDataAtCell`, 500 cells | 5.8 | 5.9 | 4.7 | 864 → 231 | 1.2× |
| `selectCell`, move | 1.7 | 1.8 | 1.8 | 0 | 0.9× |
| `selectCells`, 10×10 range | 4.7 | 2.4 | 2.4 | 0 | 2.0× |
| `selectColumns` | 4.9 | 3.2 | 3.1 | 0 | 1.6× |
| `selectRows` | 4.1 | 2.5 | 2.5 | 0 | 1.6× |
| `selectAll` | 14.3 | 13.1 | 13.4 | 0 | 1.1× |
| 1 cell change, 10×10 range selected | 8.0 | 8.3 | 4.7 | 864 → 1 | 1.7× |
| 500 cell changes, `selectAll` active | 24.2 | 23.5 | 13.6 | 864 → 231 | 1.8× |
| sort toggle (structural) | 15.7 | 16.1 | 16.0 | 864 → 864 | 1.0× |

**Custom renderer, about 0.1 ms per cell**

| scenario | old (ms) | new, `always` (ms) | new, `onChange` (ms) | speedup, old → `onChange` |
|---|---|---|---|---|
| `setDataAtCell`, 1 cell | 90.6 | 90.6 | 3.0 | 30× |
| `setDataAtCell`, 50 cells | 91.2 | 91.2 | 7.8 | 12× |
| `setDataAtCell`, 500 cells | 92.0 | 91.8 | 27.1 | 3.4× |
| 1 cell change, 10×10 range selected | 93.7 | 93.7 | 4.8 | 20× |
| 500 cell changes, `selectAll` active | 107.7 | 106.7 | 35.0 | 3× |
| sort toggle (structural) | 101.5 | 100.6 | 101.4 | 1× |

- The default mode is never slower: data changes and structural changes are equal within noise, and the selection shapes are faster in every mode because the selection diff is always on.
- `onChange` scales with the number of changed cells, not with the viewport. The 500-cell case paints 231 cells because the change list hits some cells twice.
- The remaining `onChange` floor is about 2.5 ms per full draw: resolving the meta, value, and renderer of every rendered cell to decide whether it changed. Sorting and scrolling keep the full-repaint cost by design, because every cell changes there.

**Note for reviewers:** the rendering guide stated under DEV-2759 that there is no per-cell render by design. This PR reverses that statement for an opt-in mode; the default is unchanged.

### How has this been tested?

- `tests/e2e/incremental-render.spec.ts` (Playwright, 13 cases): each case checks which cells were painted and that the tables are byte-identical to a forced full repaint (`markAllCellsChanged()` + `render()`): data and meta changes, structural changes, scrolling, `markCellChanged()`/`markAllCellsChanged()`, a column-level override, every selection shape (single, rows, columns, all, overlapping layers, deselect), frozen overlays with merged cells, formula dependents and a `HYPERLINK` whose URL lives in another cell, Search results, and the `'always'` default. Run on `e2e-main` and `e2e-main-min`.
- Unit: `src/core/incrementalRender/__tests__/`, `src/3rdparty/walkontable/test/unit/selection/appliedSelection.unit.ts`; full `npm run test:unit`.
- `npm run test:walkontable` (816 specs), full `npm run test:e2e` (legacy suite), `npm run test:types`, ESLint.

### Types of changes

### Related issue(s):

1. DEV-2792
2. #9614

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2792

---

> [!NOTE]
> **Medium Risk**
> Touches the core render and selection paths for all grids (selection diff is always on); default `renderMode: 'always'` preserves prior paint behavior, but `'onChange'` and plugin marking gaps could cause stale DOM if misused.
> 
> **Overview**
> Adds **`renderMode`** (`'always'` by default, opt-in **`'onChange'`**) at grid/column/cell level so renders can skip repainting unchanged viewport cells, plus **`markCellChanged()`** / **`markAllCellsChanged()`** for external or undetectable updates. Core tracks a render **epoch** and per-cell meta **`_renderVersion`**; **`CellPainter`** in `incrementalRender/` compares paint stamps and gates Walkontable’s new **`shouldPaintCell`** hook before cell elements are reset.
> 
> Walkontable’s **selection pass** now diffs classes/attributes per element (`appliedSelection`, cached cell scans keyed by band + epoch) instead of bulk-resetting on fast draws; cell/header renderers call **`clearAppliedSelection`** when they wipe DOM. Meta writes bump render version; dynamic `cells`/`type` extension uses **`extendCellMeta`** so unchanged re-applies don’t force repaints.
> 
> Docs/changelog cover limits (in-place value/meta mutation, external renderer state). Plugins that paint from their own state (**Search**, **MergeCells**, **Formulas** `HYPERLINK`, **Comments** on disable, **Filters** nested list focus) call the new markers or deselect where skipping broke behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d61eafc1113e852c9520dd10b60eb3b30d966ba0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2759 Document the render() method: what it covers, what it costs, and when to call it](https://app.clickup.com/t/9015210959/DEV-2759)
Task linked: [DEV-2792 Explore a skip-unchanged-cells render mode (prototype for GH #9614)](https://app.clickup.com/t/9015210959/DEV-2792)

handsontable/src/plugins/formulas/formulas.ts
```diff
@@ -2143,6 +2170,8 @@ export class Formulas extends BasePlugin {
    * @param {string} [source] Source of the call.
    */
   #onAfterLoadData = (sourceData: unknown[], initialLoad: boolean, source = '') => {
+    this.#hyperlinkCells.clear();
```
demtario, line 2203: **high** `#onAfterLoadData` (also hooked as `afterUpdateData`) clears `#hyperlinkCells` here. Same-size `updateData` does not advance the render epoch: [`loadData`](https://github.com/handsontable/handsontable/blob/a76ef0bdd80e4cc5c34f73b7b6fcf9df9b8d4098/handsontable/src/core.ts#L3770) calls `renderChangeTracker.markAllChanged()`, but the [`updateData` callback](https://github.com/handsontable/handsontable/blob/a76ef0bdd80e4cc5c34f73b7b6fcf9df9b8d4098/handsontable/src/core.ts#L3710-L3728) only `fitToLength`s, and `insertIndexes(..., 0)` returns without `cacheUpdated`. The following `render()` then skips unchanged HYPERLINK labels under `onChange`, so `#onAfterRenderer` never re-registers them. Later URL-argument edits no-op in `#invalidateHyperlinkCells` and leave a stale `href`.

After the `clear()`, call `this.hot.markAllCellsChanged()`, and/or bump the epoch in the `updateData` callback the way `loadData` already does.

demtario (commented):
1 inline note below.

Off-diff:

- **medium** `handsontable/src/plugins/search/search.ts` (`disablePlugin`): the cleanup still relies on a one-shot `beforeRenderer` to strip `htSearchResult`. `query()` now calls `markAllCellsChanged()`; this path does not. A direct `disablePlugin()` + `render()` under `onChange` skips those cells, so the class stays. `updateSettings({ search: false })` is fine because core already advances the epoch. Call `this.hot.markAllCellsChanged()` in `disablePlugin()` next to that hook.

handsontable/src/plugins/formulas/formulas.ts
budnix: Good catch — fixed in 02e05b4111: `updateData` now advances the render epoch like `loadData` does, with a unit case and a Playwright step on the HYPERLINK scenario. The search disable path is fixed in the same commit; its hook also had to read the plugin state instead of the setting, or the one-shot pass re-added the class.

demtario (commented):
2 inline notes below.

Off-diff:

- **medium** `handsontable/src/plugins/comments/comments.ts` (`disablePlugin`): only unregisters shortcuts. Under `renderMode: 'onChange'`, skipped cells keep `htCommentCell`. `updateSettings({ comments: false })` is fine because core already advances the epoch; `setComment` / `removeComment` are covered via `setCellMeta`. Call `this.hot.markAllCellsChanged()` in `disablePlugin()`, same as Search.

handsontable/src/plugins/formulas/formulas.ts
```diff
@@ -404,9 +414,25 @@ export class Formulas extends BasePlugin {
...
     ));
 
+    this.#invalidateHyperlinkCells();
     this.hot.runHooks('afterFormulasValuesUpdate', exportedChanges);
   };
 
+  /**
+   * Marks every cell rendered as a hyperlink as changed, so its `href` is rebuilt on the next render.
+   */
+  #invalidateHyperlinkCells() {
```
demtario, line 454: **high** `#invalidateHyperlinkCells` only walks keys already in `#hyperlinkCells`. A cell that becomes a `HYPERLINK` while keeping the same formatted label does not change value, renderer, version, or epoch, so `shouldPaint` is false and `#onAfterRenderer` never wraps it. No anchor appears.

On an engine write, also `markCellChanged` for any address whose `#getHyperlinkHref` is now non-null, not only keys already in the set.

handsontable/src/plugins/search/search.ts
```diff
@@ -174,6 +174,9 @@ export class Search extends BasePlugin {
     this.hot.addHookOnce('afterViewRender', () => {
       this.hot.removeHook('beforeRenderer', beforeRendererCallback);
     });
+    // The one-shot hook above strips the result class while the cells paint, so under
+    // `renderMode: 'onChange'` the next render has to reach every cell.
+    this.hot.markAllCellsChanged();
```
demtario, line 176: **medium** `markAllCellsChanged()` only forces a paint of the current band. The one-shot `beforeRenderer` is then removed. Off-screen matches keep `htSearchResult` on stored `className` (`#onBeforeRenderer` writes the union back; [`baseRenderer`](https://github.com/handsontable/handsontable/blob/02e05b4111e18bf9e9d1492e8d456780b8d096f4/handsontable/src/renderers/baseRenderer/baseRenderer.ts#L39-L40) reapplies it). Scrolling those cells in shows the highlight again.

On disable, clear `isSearchResult` and strip the class from stored meta (same as a query that matches nothing), then mark all. Do not rely on a one-shot paint of the viewport.

handsontable/src/plugins/formulas/formulas.ts
budnix: Good spot — fixed in d61eafc111: the engine update now also marks every changed cell whose hyperlink lookup is non-null, not only the ones already in the set.

handsontable/src/plugins/search/search.ts
budnix: Agreed — this landed in 7eec493f28, which strips the class from every stored meta instead of relying on a paint. The `isSearchResult` flag stays on purpose: `updatePlugin()` is disable plus enable, and clearing it would make `updateSettings({ search })` drop the results.

demtario (approved):

PR #13385, opened 2026-09-04
DEV-2629: Force a GC between perf iterations and record the live heap

demtario:
### Context

**Stacked on #13383.** Base is `feature/DEV-2629_Perf-baseline-provenance`; retarget to `develop` once that PR merges. Review only the commits above the base.

The PR changes what the performance runner does between and after iterations, so the four short-window scenarios stop reporting GC timing as grid time.

The 2026-09-04 replay of develop goldens left filtering, sorting, initial-load and sourceDataValidator with a run-to-run CV of 11 to 20% **after** removing the CI runner speed factor, against 2.5 to 3.6% on the scroll scenarios. Their windows are 50 to 150 ms on a 300 MB heap; GC lands under `scripting`, and a single pause moves a mean of three by 10 to 20%. initial-load's third iteration read about 50% slower than its first two on every run because two destroyed 100000x100 grids were still uncollected when it started.

What ships:

- **Forced GC before every measured iteration** (`HeapProfiler.collectGarbage` over a control CDP session, separate from the per-iteration tracing session). Garbage from the previous reset is collected on nobody's clock instead of inside the next window.
- **Live heap after every end mark.** After `afterActionFn`, the runner forces a GC and reads `Runtime.getHeapUsage`, records it per iteration to `heap-after-gc.json` (new `lib/heap-after-gc.mjs`, same pattern as `hook-timing.json`), and the teardown folds the average into `updateCounters.jsHeapAfterGcBytes`. The median baseline medians it over the entries that carry it; the HTML memory table shows it as `JS heap after GC`. Informational for now: the gate stays on the windowed maximum until enough goldens carry the field to derive a threshold from with `scripts/replay-goldens.mjs`. It is the metric that should eventually replace `jsHeapMaxBytes`, whose peak on the scroll scenarios depends on where V8 scheduled a GC rather than on what the grid retains.
- **Iterations 3 → 5** for filtering, sorting, initial-load and source-data-validator-load. Their iterations cost seconds next to the fixture load; the scroll scenarios stay at 3 because each iteration is 500 wheel round trips.
- **`HARNESS_VERSION` 1 → 2**, with a history block in `trace-runner.mjs`. Both changes alter the number a scenario publishes for the same grid code, so the median window restarts cleanly through the compatibility key from #13383 instead of averaging two harness definitions for five develop pushes.

Aggregation stays the arithmetic mean: median-of-3 would change the published number on exactly the noisy scenarios and buys little at n=5 once GC is controlled.

**Operational note.** After merge, PR comments read "no comparable baseline: the develop goldens were recorded on a different environment (harness 1 -> 2)" until two develop pushes on a perf-scoped path have recorded harness-2 goldens. Expected. Re-derive `REGRESSION_CALLOUT_THRESHOLD_TIMING` with `node performance-tests/scripts/replay-goldens.mjs --since <date>` once about ten harness-2 goldens exist; 15 was measured against harness-1 noise and is likely too permissive now.

### How has this been tested?

- `node --test performance-tests/lib/__tests__/*.test.mjs` — 230 pass (10 new: `heap-after-gc.test.mjs`, a median case for `jsHeapAfterGcBytes`, four HTML memory-table cases including one pinning that the gate still runs on the windowed maximum).
- `cd performance-tests && npm run lint && npm run typecheck` — clean.
- End to end on this machine, `PERF_MODE=golden npx playwright test --grep sorting`: 5 `iteration-N.json` files plus `heap-after-gc.json` with five readings (269.1 to 269.5 MB, spread 0.15%); the saved snapshot carries `harnessVersion: 2`, `runs: 5`, and `jsHeapAfterGcBytes` beside the windowed min and max; scripting CV across the five iterations 3.4%; the report refuses the harness-1 history with the reason "harness 1 -> 2".

### Types of changes

### Related issue(s):

1. DEV-2629

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

Tooling only: `performance-tests/`, the `performance-testing` skill, `performance-tests/README.md`. No shipped package changes, so no changelog entry.

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes published perf numbers and baseline compatibility (harness bump), affecting CI comparisons and golden history until new develop goldens accumulate; no product runtime impact.
> 
> **Overview**
> Bumps the performance harness to **`HARNESS_VERSION` 2** so short-window scenarios stop billing **GC from the previous reset** into the traced action. Before each measured iteration the runner **forces a full GC** (separate control CDP session); after the end mark it GCs again and records **`jsHeapAfterGcBytes`** via `heap-after-gc.json`, folded into snapshots and shown in the HTML report as **“JS heap after GC”** (informational only—the regression gate stays on windowed **max** heap).
> 
> **Filtering, sorting, initial-load, and source-data-validator-load** go from **3 → 5** measured iterations to tame run-to-run spread on ~50–150 ms windows; scroll scenarios stay at 3. Shared **`sidecar.mjs`** unifies hook-timing and heap sidecars; median baselines **median the post-GC heap** when enough goldens carry it. Docs and **`scenario-configs.test.mjs`** pin iteration counts and versioning rules (`measurementVersion` now covers iteration-count changes).
> 
> Until two develop runs record harness-2 goldens, PR compare mode will report **no comparable baseline** (harness 1 → 2)—expected.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a098fd3d0f7ff9bffa928cad9d798c5b28abd2ad. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.claude/skills/performance-testing/SKILL.md
+++ b/.claude/skills/performance-testing/SKILL.md
@@ -204,7 +204,7 @@ These combine `scrollViewportTo()` with a deterministic `waitForFunction` that c
 | initial-load | 100000x100 | `new Handsontable(...)` | Grid construction only |
 | source-data-validator-load | 100000x100 | `new Handsontable(...)` with `sourceDataValidator` | initial-load's fixture plus the one option |
 
-Iterations: 3 for the scroll and cell-editing scenarios, 5 for filtering, sorting, initial-load and source-data-validator-load. Those four have 50 to 150 ms windows on a 300 MB heap, where one GC pause inside the window moves a mean of three by 10 to 20%, and their iterations are cheap next to the fixture load. Do not raise the scroll scenarios: each iteration is 500 wheel round trips.
+Iterations: 3 for the scroll and cell-editing scenarios, 5 for filtering, sorting, initial-load and source-data-validator-load. Each of those four states its own reason in its `scenario.config.mjs` (short windows on a 300 to 350 MB heap where one GC pause moves a mean of three by 10 to 20%; for `source-data-validator-load`, a 20% run-to-run spread after removing the runner factor), and their iterations are cheap next to the fixture load. `lib/__tests__/scenario-configs.test.mjs` pins the counts, so change the config and the test together. Do not raise the scroll scenarios: each iteration is 500 wheel round trips.
 
 ## Run Commands
 
@@ -266,8 +266,9 @@ Everything published describes the slice between the two `performance.mark`s tha
   `scrollToRow`/`scrollToColumn` report trimming, not scroll position, so their
   `waitForFunction` returns before the scroll has rendered.
 
-A category measured as exactly `0`, or a CV of `141.42%` (the CV of `{x, 0, 0}` at three
-iterations), means the window is wrong -- not that the operation was cheap.
+A category measured as exactly `0`, or a CV of `sqrt(n - 1) × 100%` (one nonzero iteration among
+zeros: `141.42%` at three iterations, `200%` at five), means the window is wrong -- not that the
+operation was cheap.
 
 1. **Spec** calls `runTracedScenario()` -> forced GC (over a control CDP session), CDP `Tracing.start`, start mark, action, settle, end mark, `afterActionFn`, forced GC + `Runtime.getHeapUsage` readback, `Tracing.end` -> raw JSON per iteration, plus `heap-after-gc.json` for the scenario. The GC before tracing keeps the previous reset's garbage out of the window (initial-load's third iteration read ~50% slower than its first two before it); the readback after the end mark is the live set, recorded as `updateCounters.jsHeapAfterGcBytes`. Both are part of `HARNESS_VERSION` 2.
 2. **Teardown** (`lib/teardown.mjs`) discovers `output/*/iteration-*.json`, calls `parseTrace()` from `trace-parser.mjs`
--- a/.claude/skills/performance-testing/SKILL.md
+++ b/.claude/skills/performance-testing/SKILL.md
@@ -204,6 +204,8 @@ These combine `scrollViewportTo()` with a deterministic `waitForFunction` that c
 | initial-load | 100000x100 | `new Handsontable(...)` | Grid construction only |
 | source-data-validator-load | 100000x100 | `new Handsontable(...)` with `sourceDataValidator` | initial-load's fixture plus the one option |
 
+Iterations: 3 for the scroll and cell-editing scenarios, 5 for filtering, sorting, initial-load and source-data-validator-load. Those four have 50 to 150 ms windows on a 300 MB heap, where one GC pause inside the window moves a mean of three by 10 to 20%, and their iterations are cheap next to the fixture load. Do not raise the scroll scenarios: each iteration is 500 wheel round trips.
+
 ## Run Commands
 
 ```bash
@@ -267,7 +269,7 @@ Everything published describes the slice between the two `performance.mark`s tha
 A category measured as exactly `0`, or a CV of `141.42%` (the CV of `{x, 0, 0}` at three
 iterations), means the window is wrong -- not that the operation was cheap.
 
-1. **Spec** calls `runTracedScenario()` -> CDP `Tracing.start`, start mark, action, settle, end mark, `Tracing.end` -> raw JSON per iteration
+1. **Spec** calls `runTracedScenario()` -> forced GC (over a control CDP session), CDP `Tracing.start`, start mark, action, settle, end mark, `afterActionFn`, forced GC + `Runtime.getHeapUsage` readback, `Tracing.end` -> raw JSON per iteration, plus `heap-after-gc.json` for the scenario. The GC before tracing keeps the previous reset's garbage out of the window (initial-load's third iteration read ~50% slower than its first two before it); the readback after the end mark is the live set, recorded as `updateCounters.jsHeapAfterGcBytes`. Both are part of `HARNESS_VERSION` 2.
 2. **Teardown** (`lib/teardown.mjs`) discovers `output/*/iteration-*.json`, calls `parseTrace()` from `trace-parser.mjs`
 3. **trace-parser.mjs** categorizes events into DevTools categories (scripting, rendering, painting, loading, system, idle), measures the window `runTracedScenario()` marked around the action (falling back to the auto-zoomed window only for traces recorded without marks), synthesizes ProfileCall scripting from CPU profile data
 4. **Teardown** averages across iterations via `averageParsedTraces()`, collects per-iteration values for CV% calculation, strips internal fields (`_iterationValues`, `_debug`) from saved snapshots
--- a/.claude/skills/performance-testing/SKILL.md
+++ b/.claude/skills/performance-testing/SKILL.md
@@ -59,11 +59,14 @@ export default {
   name: 'my-scenario',
   warmupRuns: 1,
   iterations: 3,
-  // Bump when this spec changes what the marked window contains: the median baseline only draws on
-  // develop goldens recorded at the same version (see lib/environment.mjs).
+  // Bump when this spec changes what the marked window contains, or when `iterations` changes: the
+  // median baseline only draws on develop goldens recorded at the same version (see
+  // lib/environment.mjs).
   measurementVersion: 1,
 };
 ```
+
+`measurementVersion` covers everything scenario-local that changes the published number or its spread: the window's contents, and the iteration count (a mean of five and a mean of three have the same expected value but different spreads, and `CV run` would mean something different per row). `HARNESS_VERSION` covers runner-wide changes. When both change in one PR, the harness bump already restarts the whole golden pool and the scenario bump is not needed on top; say so in the config comment.
 
 The `name` must match the directory name -- it determines the `output/<name>/` subdirectory, and the teardown reads `measurementVersion` from the config by that name. Add a comment documenting the grid size and why it was chosen.
 
@@ -266,9 +269,9 @@ Everything published describes the slice between the two `performance.mark`s tha
   `scrollToRow`/`scrollToColumn` report trimming, not scroll position, so their
   `waitForFunction` returns before the scroll has rendered.
 
-A category measured as exactly `0`, or a CV of `sqrt(n - 1) × 100%` (one nonzero iteration among
-zeros: `141.42%` at three iterations, `200%` at five), means the window is wrong -- not that the
-operation was cheap.
+A category measured as exactly `0`, or a CV of `sqrt(n) × 100%` (one nonzero iteration among
+zeros, with the sample standard deviation `calcCv` uses: `173.21%` at three iterations, `223.61%`
+at five), means the window is wrong -- not that the operation was cheap.
 
 1. **Spec** calls `runTracedScenario()` -> forced GC (over a control CDP session), CDP `Tracing.start`, start mark, action, settle, end mark, `afterActionFn`, forced GC + `Runtime.getHeapUsage` readback, `Tracing.end` -> raw JSON per iteration, plus `heap-after-gc.json` for the scenario. The GC before tracing keeps the previous reset's garbage out of the window (initial-load's third iteration read ~50% slower than its first two before it); the readback after the end mark is the live set, recorded as `updateCounters.jsHeapAfterGcBytes`. Both are part of `HARNESS_VERSION` 2.
 2. **Teardown** (`lib/teardown.mjs`) discovers `output/*/iteration-*.json`, calls `parseTrace()` from `trace-parser.mjs`
--- a/performance-tests/README.md
+++ b/performance-tests/README.md
@@ -158,9 +158,9 @@ Two rules follow for anyone writing or changing a scenario:
   `scrollToColumn` wait on trimming, not scroll position, so they return before the scroll
   has rendered; the settle is what makes those resets deterministic.
 
-A category measured as exactly `0`, or a CV of `sqrt(n - 1) × 100%` (one nonzero iteration among
-zeros: `141.42%` at three iterations, `200%` at five), means the window is wrong -- not that the
-operation was cheap.
+A category measured as exactly `0`, or a CV of `sqrt(n) × 100%` (one nonzero iteration among
+zeros, with the sample standard deviation `calcCv` uses: `173.21%` at three iterations, `223.61%`
+at five), means the window is wrong -- not that the operation was cheap.
 
 ### Golden baseline workflow
 
--- a/performance-tests/README.md
+++ b/performance-tests/README.md
@@ -83,7 +83,7 @@ Each scenario measures a specific user interaction pattern:
 | **initial-load** | 100000 x 100 | `new Handsontable(...)` | Grid construction only |
 | **source-data-validator-load** | 100000 x 100 | `new Handsontable(...)` with `sourceDataValidator` | Same fixture as initial-load plus the one option |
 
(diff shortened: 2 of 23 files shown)


qunabu:

performance-tests/README.md
```diff
@@ -152,8 +152,9 @@ Two rules follow for anyone writing or changing a scenario:
   `scrollToColumn` wait on trimming, not scroll position, so they return before the scroll
   has rendered; the settle is what makes those resets deterministic.
 
-A category measured as exactly `0`, or a CV of `141.42%` (the CV of `{x, 0, 0}` at three
-iterations), means the window is wrong -- not that the operation was cheap.
+A category measured as exactly `0`, or a CV of `sqrt(n - 1) × 100%` (one nonzero iteration among
+zeros: `141.42%` at three iterations, `200%` at five), means the window is wrong -- not that the
+operation was cheap.
```
budnix: This formula does not match what `calcCv` actually computes. It divides by `values.length - 1`, the sample standard deviation, so one nonzero iteration among zeros comes out at `sqrt(n) x 100%` -- I ran it and got 173.21% at three iterations and 223.61% at five, not the 141.42% and 200% written here. That matters because someone who hits a genuinely short capture will look for these numbers, not find them, and read a broken window as a real measurement. The old 141.42% was already the population value, so generalizing it carried the error forward.

```suggestion
A category measured as exactly `0`, or a CV of `sqrt(n) × 100%` (one nonzero iteration among
zeros: `173.21%` at three iterations, `223.61%` at five), means the window is wrong -- not that the
operation was cheap.
```

.claude/skills/performance-testing/SKILL.md
```diff
@@ -264,10 +266,11 @@ Everything published describes the slice between the two `performance.mark`s tha
   `scrollToRow`/`scrollToColumn` report trimming, not scroll position, so their
   `waitForFunction` returns before the scroll has rendered.
 
-A category measured as exactly `0`, or a CV of `141.42%` (the CV of `{x, 0, 0}` at three
-iterations), means the window is wrong -- not that the operation was cheap.
+A category measured as exactly `0`, or a CV of `sqrt(n - 1) × 100%` (one nonzero iteration among
+zeros: `141.42%` at three iterations, `200%` at five), means the window is wrong -- not that the
+operation was cheap.
```
budnix: Same numbers as in `performance-tests/README.md`, and the same fix -- `calcCv` uses the sample standard deviation, so this is `sqrt(n) x 100%`: 173.21% at three iterations and 223.61% at five. Worth changing both files together so the two docs stay in step.

```suggestion
A category measured as exactly `0`, or a CV of `sqrt(n) × 100%` (one nonzero iteration among
zeros: `173.21%` at three iterations, `223.61%` at five), means the window is wrong -- not that the
operation was cheap.
```

performance-tests/lib/trace-runner.mjs
```diff
@@ -224,46 +264,74 @@ export async function runTracedScenario({
...
+        if (afterActionFn) {
+          await afterActionFn();
+        }
 
-    process.stdout.write(' stopping');
-    const traceJson = await stopTracing(cdp);
+        // Also outside the window (the UpdateCounters extrema are taken over the marked samples
+        // only), and after the readbacks so the GC does not compete with them: what the action left
+        // alive, per iteration. Informational, so a failed readback records null and moves on.
+        heapAfterGc.push(await heapUsedAfterGc(control));
```
budnix: This runs before `stopTracing`, so a full major GC on a 300 to 350 MB heap is recorded inside every saved trace. The marked window keeps it out of the timing categories and the UpdateCounters extrema, but not out of the auto-zoom fallback: when an iteration loses its marks, `calculateWindow` picks the busiest region, and a forced major GC is a strong candidate for that -- which is the defect the marks exist to prevent. Moving this call to just after `stopTracing(cdp)` costs nothing, since stopping the trace does not touch page state.

performance-tests/lib/trace-runner.mjs
```diff
@@ -224,46 +264,74 @@ export async function runTracedScenario({
...
+      } finally {
+        clearInterval(heartbeat);
+      }
 
-    const outPath = join(outputDir, `iteration-${i}.json`);
+      process.stdout.write(' stopping');
+      const traceJson = await stopTracing(cdp);
```
budnix: `stopTracing` only sends `Tracing.end`, so the session `startTracing` creates each iteration is never detached -- a five-iteration run ends with five still attached, plus `control`. If `actionFn` or `settle` throws, this line never runs at all, so that session keeps an active trace while the outer `finally` detaches only `control`. Since this PR added exactly that cleanup for the new session a few lines down, extending it to the one holding the recording would be good.

performance-tests/lib/heap-after-gc.mjs
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * @param {Array<number | null>} values -- per-iteration used heap after GC, in bytes; null where
+ *   the readback failed for that iteration
+ * @returns {{ averageBytes: number | null, values: Array<number | null> }}
+ */
+export function summarizeHeapAfterGc(values) {
+  const finite = (values || []).filter(v => typeof v === 'number' && Number.isFinite(v));
+
+  return {
+    averageBytes: finite.length > 0 ? finite.reduce((a, b) => a + b, 0) / finite.length : null,
```
budnix, line 24: This averages only the readings that came back, and `saveHeapAfterGc` writes whenever at least one did, so four failed readbacks out of five publish a single-sample number as the run's `jsHeapAfterGcBytes`. Nothing reaches the snapshot to say how many samples were behind it, because teardown reads only `averageBytes`. The marks check in `teardown.mjs` throws in golden mode for the same reason -- one bad iteration folded into a mean that nothing later can tell apart -- and this field is meant to become the gate, so a `readCount` on the summary carried onto `updateCounters` would close the same hole.

performance-tests/lib/heap-after-gc.mjs
```diff
@@ -0,0 +1,45 @@
...
+ * @param {string} outputDir -- scenario output directory
+ * @param {Array<number | null>} values
+ */
+export async function saveHeapAfterGc(outputDir, values) {
+  const summary = summarizeHeapAfterGc(values);
+
+  if (summary.averageBytes === null) {
+    return;
+  }
+
```
budnix, line 44: This `mkdir` can never do anything, since `runTracedScenario` creates `outputDir` before the first iteration and `saveHeapAfterGc` only runs after the loop. Harmless, but `lib/hook-timing.mjs` already writes the same shape of per-iteration sidecar, and `teardown.mjs` now reads two near-identical ones in a row. Curious whether one shared save-and-read helper is worth it here -- it would also put `hook-timing.json` behind a named constant the way `HEAP_AFTER_GC_FILE` now is.

performance-tests/lib/median-snapshot.mjs
```diff
@@ -91,13 +91,20 @@ function medianScenario(entries) {
   if (updateCounterEntries.length > 0) {
     const jsHeapMinBytes = median(updateCounterEntries.map(uc => uc.jsHeapMinBytes));
     const jsHeapMaxBytes = median(updateCounterEntries.map(uc => uc.jsHeapMaxBytes));
+    // Absent on goldens recorded before the runner read the live heap; median() skips those, so a
+    // window mixing old and new entries medians only the ones that have it.
+    const jsHeapAfterGcBytes = median(updateCounterEntries.map(uc => uc.jsHeapAfterGcBytes));
```
budnix: `median()` skipping absent values is right for old goldens, but it also means this field gets medianed over however many entries happen to carry it, with no floor. If four of the five goldens in the window had failing readbacks, the baseline's `jsHeapAfterGcBytes` is one run's number sitting beside a true median of five for `jsHeapMaxBytes`, and the report still labels the whole row a median of five. That is the case `MIN_VALID_SNAPSHOTS` exists to refuse, so applying the same floor per field would keep the label honest.

performance-tests/lib/teardown.mjs
```diff
@@ -151,6 +152,25 @@ async function collectScenarioResults() {
...
+    // updateCounters beside the windowed extrema it is meant to eventually replace as the gate.
+    const heapAfterGcPath = join(scenarioDir, HEAP_AFTER_GC_FILE);
+
+    if (await exists(heapAfterGcPath)) {
+      const heapData = JSON.parse(await readFile(heapAfterGcPath, 'utf8'));
+      const bytes = typeof heapData.averageBytes === 'number' ? heapData.averageBytes : null;
+
+      // The per-iteration readings stay in heap-after-gc.json (uploaded with the artifact); no
+      // report reads them yet, so they are not carried on _iterationValues.
+      if (bytes !== null) {
```
budnix: When the window caught no `UpdateCounters` samples, `averageParsedTraces` leaves `updateCounters` as null, and this spread turns it into a two-field object. That object then gets past the `if (!cUc) return []` guard in `buildMemoryMetrics`, so the report grows a memory table with `--` in every other row, and `medianScenario` stamps `sampleCount: 0` on it. These are 50 to 150 ms windows, so it is reachable, and guarding on the existing object keeps the shape honest.

```suggestion
      if (bytes !== null && averaged.updateCounters) {
```

performance-tests/lib/html-report-builder.mjs
```diff
@@ -362,6 +362,9 @@ function buildMemoryMetrics(current, golden, isCrossWindow = false) {
   const pairs = [
     ['Min JS heap', 'jsHeapMinLabel', 'jsHeapMinBytes'],
     ['Max JS heap', 'jsHeapMaxLabel', 'jsHeapMaxBytes'],
+    // The live set after a forced GC (lib/heap-after-gc.mjs). Informational until enough goldens
+    // carry it to derive a threshold; the row is skipped for runs recorded before it existed.
+    ['JS heap after GC', 'jsHeapAfterGcLabel', 'jsHeapAfterGcBytes'],
```
budnix: The memory table paints every row with `classifyChangeCss(row.change, scenario.heapThreshold)`, so this new row turns red at 5% (10% on the scroll scenarios) exactly like the row that actually gates. No threshold has been derived for the live set yet, so a 7% move between two runs will read as a regression sitting next to a flat `jsHeapMaxBytes`. The detailed-metrics table already honors `row.neutral` for informational rows, so setting that here would keep the coloring in step with the comment right above this line.

performance-tests/lib/__tests__/html-report-builder-provenance.test.mjs
```diff
@@ -158,6 +158,79 @@ describe('buildHtmlReport -- environment and refusal', () => {
...
+    const data = payloadOf(buildHtmlReport(
+      { a: withAfterGc(55_000_000, '55.0 MB') }, { timestamp: 't', scenarios: { a: golden() } }, {}
+    ));
+    const row = data.scenarios[0].memory.find(r => r.label === 'JS heap after GC');
+
+    assert.equal(row.currentDisplay, '55.0 MB');
+    assert.equal(row.baselineDisplay, '--');
+    assert.equal(row.change, null);
+  });
+
```
budnix, line 203: These three cases cover both sides present, current only, and neither, but not the current run missing the field while a harness-2 baseline carries it. That one renders as baseline `50.0 MB`, current `--`, change null, which looks the same as a metric nobody measured even though the capture actually failed. `thresholds.mjs` already exports `INCOMPARABLE_LABELS['current-incomplete']` as 'capture incomplete' for exactly that difference, so a case pinning it would be a good addition.

performance-tests/scenarios/source-data-validator-load/scenario.config.mjs
```diff
@@ -5,7 +5,9 @@
 export default {
   name: 'source-data-validator-load',
   warmupRuns: 1,
-  iterations: 3,
+  // Five, not three: run-to-run spread of the validator pass over 10 million cells was 20% after
+  // removing the runner factor. The iterations are cheap next to the fixture load.
```
budnix, line 9: Going from three to five iterations changes neither what the marked window contains nor the runner, so `measurementVersion` stayed at 1 here and in the other three configs. That is safe this time only because `HARNESS_VERSION` 2 already invalidates the whole golden pool. The next iteration-count change on a single scenario has no bump path, and would median three-sample and five-sample means together in one window -- same expected value, different spread, and a CV run column that means something different per row. Worth saying which of the two versions is meant to cover this.

performance-tests/README.md
demtario: Done, suggestion applied, plus a note that this is the sample form `calcCv` uses.

.claude/skills/performance-testing/SKILL.md
demtario: Done, same wording as the README.

performance-tests/lib/trace-runner.mjs
demtario: Done. The readback now runs after `stopTracing`, so no saved trace contains the forced GC and the auto-zoom fallback cannot land on it.

performance-tests/lib/trace-runner.mjs
demtario: Done. Each iteration's tracing session is detached in a `finally` that wraps the marked section and `stopTracing`, so a throwing `actionFn` releases it too; the control session keeps its own `finally`.

performance-tests/lib/heap-after-gc.mjs
demtario: Done. `summarizeHeapAfterGc` reports `readCount`; the teardown carries it as `updateCounters.jsHeapAfterGcSamples` and warns when fewer iterations than ran produced a reading. A warning rather than the golden-mode throw the marks check uses, because the field is informational until it gates; the throw belongs at the commit that makes it the gate.

performance-tests/lib/heap-after-gc.mjs
demtario: Done. `lib/sidecar.mjs` (`writeSidecar` / `readSidecar`) is used by hook-timing.mjs, heap-after-gc.mjs and the teardown, and `hook-timing.json` sits behind `HOOK_TIMING_FILE`. The `mkdir` lives in the shared writer, where it is generic rather than dead.

performance-tests/lib/median-snapshot.mjs
demtario: Done. `jsHeapAfterGcBytes` is medianed only when at least `MIN_VALID_SNAPSHOTS` entries carry it, otherwise omitted from the median. Test added.

performance-tests/lib/teardown.mjs
demtario: Done, suggestion applied.

performance-tests/lib/html-report-builder.mjs
demtario: Done. The row carries `neutral: true` and the memory table paints neutral rows unbanded, like the trace-window row.

performance-tests/lib/__tests__/html-report-builder-provenance.test.mjs
demtario: Done. When the baseline carries the field and this run does not, the row is incomplete with the `capture incomplete` label, rendered per row. Test added.

performance-tests/scenarios/source-data-validator-load/scenario.config.mjs
demtario: Done. `measurementVersion` covers an iteration-count change (same expected value, different spread, and `CV run` would mean something different per row). All nine config comments and the skill say so, and the four bumped configs note that `HARNESS_VERSION` 2 covered this instance.

budnix (dismissed):

budnix (approved):

PR #13387, opened 2026-09-04
DEV-2785: Keep an editor open when focus moves into preventCloseElement

demtario:
### Context

The PR fixes a regression reported by a customer: in 18.1.0 the documented `preventCloseElement` contract for custom cell editors stopped working, so the flatpickr recipe from the docs no longer saves the date picked in the calendar. The cell keeps its old value. The same setup works on 18.0.0.

`preventCloseElement` is public API. The custom-cells guide states that clicks on that element are not treated as a click outside, and two shipped recipes rely on it (flatpickr, color picker).

**Root cause.** PR #13194 added a second outside-click path in `tableView.ts`: a focus-based verdict on the document's `mouseup`. `isFocusLostToOutside` is built on `FocusGridManager#isForeignFocusTarget()`, which answers `true` for any focused element outside the grid's root element. flatpickr appends its calendar to `document.body` and moves the browser focus into it, so the next `mouseup` anywhere read as a focus loss to the outside. That deselected the cell, `afterDeselect` closed the editor, and the editor's **pre-edit** value was committed before flatpickr had applied the newly clicked day. The recipe's own `onClose` handler then called `finishEditing()` on an editor that was already finished, which is a no-op.

`preventCloseElement`'s only mechanism was a `mousedown` listener in `editorFactory` that stops propagation, which reaches only the older of the two paths. Worse, because that mousedown handler never runs, `#outsideClickHandled` stays `false`, so the `!wasOutsideClickHandled` guard let the mouseup deselect through.

The measurement that isolated it, on the unmodified recipe: open the editor, then click the calendar's next-month arrow, which selects no date, so any state change is the grid's doing.

| Version | After clicking the next-month arrow |
|---|---|
| 18.0.0 | Editor open, calendar open, `calendarContainer` holds focus |
| 18.1.0 and `develop` | Editor destroyed, cell deselected |

The calendar takes the focus on **both** versions, so focus behavior is not the difference. 18.1.0 started acting on it.

**The fix.** The open editor's `preventCloseElement` and its subtree now count as part of the editor, on the event-path side (`#isPathWithinGrid`) and on the focused-element side (a guard on `isFocusLostToOutside`). Both halves are needed: the path guard alone still lets `unlisten()` fire off `isFocusLostToOutside` on its own. The surface lookup is gated on `isOpened()`, because editor instances are cached per class per grid, so a picker left in the document body by an earlier edit must not suppress genuine outside clicks for the rest of the instance's life.

`preventCloseElement` also moves from `editorFactory`'s internal type onto `BaseEditor`, so an editor written by hand can declare one. It is deliberately left uninitialized, so an editor that never sets one keeps reading `undefined`, which is what the existing unit test asserts.

Two notes for the reviewer:

1. The **color picker recipe was never affected**, which is worth knowing because it explains the shape of the tests. Pickr suppresses the focus move, so `document.activeElement` stays the editor's own input and `tableView.ts` returns early before the focus verdict. The trigger is the focus moving into the third-party UI, not merely clicking it, so every test case here asserts the focus move as its own precondition. Without that, a case goes green on unfixed code and pins nothing.
2. The `#isPathWithinGrid` half is **partly hardening beyond the reported bug**, and deliberately so. The mousedown path was not broken: `editorFactory`'s `stopPropagation` still suppresses it. But that listener is wired exactly once, right after the editor's `init`/`afterInit` returns, so it never exists for a `preventCloseElement` assigned later in `beforeOpen`/`afterOpen` (both documented hooks), nor for an editor that rebuilds its picker element on every open. For those two shapes this branch is the only mousedown-side protection. The resulting behavior matches what `stopPropagation` already produces, so nothing observable changes for existing editors.

Three things were considered and deliberately **not** done, each recorded in the code comments:

- `FocusGridManager#isForeignFocusTarget()` is untouched. It mirrors the focus-scope-manager rule and has a much wider blast radius.
- No `mouseup` `stopPropagation` was added. That handler also runs `selection.finish()` and clears `#mouseDown` / `#selectionMouseDown` / `#textSelectionOverlay`, so suppressing it would leave stale drag state.
- `hot.isListening()` still goes `false` when the focus sits in an out-of-grid picker. That is pre-existing behavior: the focus scope manager unlistens whenever a focus event lands outside every scope, and it is unchanged since 18.0.0 (verified against the diff). An editor in that situation commits through its own callback, the way the recipe's `onClose` does, never through a keystroke. The tests are written to that reality.

One known limitation is left open, noted in `#isPathWithinGrid`'s JSDoc: the shadow-host fallback is keyed on the root element's host chain, so it covers a surface in the grid's own shadow tree but not one in a different shadow tree under a path-filtering host. That combination is unverifiable without such a host, so no guard was written on reasoning alone. The focus-side test is what carries the reported case.

### How has this been tested?

**New Playwright coverage** in `tests/e2e/editor-prevent-close-element.spec.ts`, with a flatpickr-free fixture: a custom `editorFactory` editor whose panel is appended to `document.body` and declared as `preventCloseElement`. Five cases, and **three of them are red when only `tableView.ts` is reverted** (verified by reverting, rebuilding and re-running):

1. Survives the browser focus moving into the panel. Asserts through the committed-change log, not the cell value, because the forced commit wrote the value the editor was opened with, so the cell looks untouched.
2. Commits the value the panel reported, ending the edit from the panel's own callback the way the recipe's `onClose` does. This is the end-to-end symptom.
3. Survives the focus moving into an `<input>` inside the panel. This is the branch flatpickr's year field and month `<select>` take (`isOutsideInput`, evaluated before the focus check).
4. Survives the focus move with `outsideClickDeselects: false`. Already green before the fix; regression coverage for that setting's own path.
5. Negative control: a genuine click outside both the grid and the panel still ends the edit, so the guard cannot be overbroad. The point is derived from both measured boxes and asserted with `elementFromPoint`, so a later fixture edit cannot make it quietly press the grid.

**Manual verification against the real recipe**, running the unmodified `docs/content/recipes/cell-types/flatpickr/javascript/example1.js` on a local build. Before: the editor is destroyed by any click in the calendar. After: `Release (EU)` goes from `15/03/2025` to `01/03/2025` and the linked `Release (US)` cell updates with it, since both columns read the same `releaseDate` field. Also walked the year `<input>`, then the month `<select>`, then a day click in one session: the editor stays `STATE_EDITING` throughout and commits `2025-05-20`.

The reporter's initial hypothesis, that the new hidden-cell editor guard was closing the editor, was **ruled out** rather than assumed away: the recipe registers no hiding index map (`autoRowSize` registers a `physicalIndexToValue` map, and there is no Pagination, `hiddenRows` or `hiddenColumns`), and `rowIndexMapper.isHidden(0)` and `columnIndexMapper.isHidden(1)` both read `false` in the browser, so that guard returns before doing anything.

Commands:

```bash

# the new spec, and the whole functional tier
cd tests && npx playwright test editor-prevent-close-element --project=e2e-main

```

Results: full legacy Jasmine suite **9894 specs, 0 failures**. Jest unit **4425 passed**. Playwright `e2e-main` **897 passed**, with the new spec also green on `e2e-horizon`, `e2e-classic` and `e2e-main-min`. `eslint` and `stylelint` clean, `tsc --noEmit` clean.

One pre-existing regression was caught and fixed on the way: initializing the new field to `null` broke `editors/__tests__/index.spec.js`, which asserts `preventCloseElement` reads `undefined` when the editor does not set one. Leaving the field uninitialized keeps that contract intact.

The change was also reviewed by two independent reviewers before this PR. Neither found a blocking issue; their notes on a docstring imprecision and a hardcoded test coordinate are already applied above.

### Types of changes

### Related issue(s):

1. DEV-2785

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2785

---

> [!NOTE]
> **Medium Risk**
> Changes core outside-click and focus handling in `tableView.ts` for all grids with open editors, though behavior is scoped to editors that set `preventCloseElement` while opened.
> 
> **Overview**
> Fixes a **18.1 regression** where custom editors with an out-of-grid `preventCloseElement` (e.g. flatpickr) closed and committed the **pre-edit** value when the picker took browser focus. The older outside-click path only used `mousedown` `stopPropagation` from `editorFactory`; the newer **document `mouseup` focus-loss** path still treated the picker as outside the grid.
> 
> **`tableView.ts`** now treats the open editor’s `preventCloseElement` subtree as part of the editor on **both** paths: `#isPathWithinGrid()` includes the surface in the event path, and `#isFocusWithinEditorSurface()` blocks false “focus lost outside” on `mouseup`. Lookup is gated on **`isOpened()`**, ignores surfaces that **contain the grid** (with a one-time warning), and uses **`closest()`** for shadow-DOM focus.
> 
> **`preventCloseElement`** is promoted to a public optional field on **`BaseEditor`** (removed from the factory-only extended type). Docs, changelog, and **AGENTS.md** describe the dual outside-click paths. **Playwright** coverage exercises focus moves, `afterOpen` assignment, drag-release outside the panel, and invalid `body` surfaces.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 114ec42890a4e85519d8895a65fb075ce02f03ed. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2785 [18.1.0+] Flatpickr recipe: date picker selection not saved to cell ](https://app.clickup.com/t/9015210959/DEV-2785)

handsontable/src/tableView.ts
```diff
@@ -1850,6 +1869,47 @@ class TableView {
...
+   * loss to the outside and deselect the cell, committing the editor's pre-edit value.
+   *
+   * @private
+   * @param {HTMLElement|null} element The deepest reachable focused element.
+   * @returns {boolean}
+   */
+  #isFocusWithinEditorSurface(element: HTMLElement | null): boolean {
+    const surface = this.#getActiveEditorSurface();
+
+    return surface !== null && element !== null && surface.contains(element);
```
budnix: The element handed in here comes from `getDeepActiveElement()`, which walks into shadow roots, but `contains()` stops at a shadow boundary. So a picker built as a web component puts the focus on a node inside its own shadow root, this returns false, and the cell gets deselected with the pre-edit value committed again, which is the same symptom the PR fixes just with a different picker library. `closest()` is already imported in this file and it follows `.host` across those boundaries.
```suggestion
    return surface !== null && element !== null && closest(element, [surface]) !== null;
```

tests/fixtures/demo/editor-prevent-close-element.html
```diff
@@ -0,0 +1,161 @@
...
+        panelInput.setAttribute('data-testid', 'panel-input');
+
+        panel.appendChild(focusTarget);
+        panel.appendChild(setValueTarget);
+        panel.appendChild(commitTarget);
+        panel.appendChild(panelInput);
+        editor.hot.rootDocument.body.appendChild(panel);
+
+        editor.picker = panel;
+        editor.preventCloseElement = panel;
```
budnix: Assigning the panel here in `init` means the factory wires its own `mousedown` stopPropagation listener, so a press on the panel never reaches the document handler at all. That leaves the new `#isPathWithinGrid` branch untouched by every spec, and it also blocks the mouseup deselect twice over, so deleting either half of the fix keeps all five cases green. Would be good to add a case that assigns the surface in `afterOpen`, where no factory listener exists, so the two halves get told apart.

tests/fixtures/pages/EditorPreventCloseElementPage.ts
```diff
@@ -0,0 +1,238 @@
...
+    await expect.poll(() => this.isEditorOpen()).toBe(true);
+    await expect(this.panel).toBeVisible();
+  }
+
+  /**
+   * Clicks the panel's focusable element and reports whether the browser focus actually landed
+   * inside the panel.
+   *
+   * The precondition matters more than it looks. A popup that keeps the focus in the editor's own
+   * input never reaches the focus-driven outside-click path at all - `tableView`'s `mouseup`
```
budnix: That early return needs `!isOutsideInput(el)`, and `isOutsideInput` is only false when the element carries `data-hot-input`, which `editorFactory` never stamps on `editor.input`. So a factory editor's own input is an outside input and the handler does not return early. The color-picker recipe survives for a different reason: its input sits inside `rootElement`, so `isForeignFocusTarget()` is false. Worth fixing the wording, since the next person will reason from it.

tests/fixtures/pages/EditorPreventCloseElementPage.ts
```diff
@@ -0,0 +1,238 @@
+import { type Locator, type Page, expect } from '@playwright/test';
+
+interface HandsontableFixture {
+  getSelected(): number[][] | undefined;
+  getActiveEditor(): { isOpened(): boolean; state: string } | undefined;
+  getDataAtCell(row: number, col: number): unknown;
+}
+
+interface FixtureWindow extends Window {
```
budnix: `windowTypes.ts` already declares `hot` globally with the full instance type, so extending `Window` and narrowing `hot` to three members here is a TS2430 conflict. Nothing catches it because `tests/` has no tsconfig and only runs eslint. `CellDropdownArrowPage.ts` carries a comment about exactly this, and `EditorHiddenCellPage.ts` keeps `extends Window` only for its extra prop, so either shape works.
```suggestion
interface FixtureWindow {
```

handsontable/src/editors/baseEditor/baseEditor.ts
```diff
@@ -92,6 +92,18 @@ export class BaseEditor {
    * @type {HTMLTableCellElement}
    */
   TD: HTMLTableCellElement | null = null;
+  /**
+   * An element the editor renders OUTSIDE its own container – a dropdown, popover or third-party
+   * picker appended to the document body. The grid treats this element and its whole subtree as a
+   * part of the editor, so neither a click landing in it nor the browser focus moving into it
+   * counts as a click or a focus loss outside the grid.
+   *
+   * Left UNINITIALIZED on purpose: an editor that never sets one must keep reading `undefined`,
```
budnix: This paragraph ships straight into the public API reference, where "the factory's type alone" means nothing to someone writing a custom editor. The `undefined` contract is worth keeping, just as a plain code comment rather than published JSDoc. The `@type {HTMLElement}` three lines down also does not match the declared `HTMLElement | null`, so the generated docs will state the wrong type.

handsontable/src/editors/baseEditor/baseEditor.ts
```diff
@@ -92,6 +92,18 @@ export class BaseEditor {
...
+   * which is what it read while the property was declared on the factory's type alone.
+   *
+   * @type {HTMLElement}
+   */
+  preventCloseElement?: HTMLElement | null;
```
budnix, line 106: This moves `preventCloseElement` onto the public `BaseEditor` type, which is the part users feel, since the recipes currently declare their own type intersections just to get an assignment to compile. `editors.types.ts` already types `init(editor: BaseEditor)`, so one assignment in there would pin the new surface and stop a later refactor from quietly dropping it. Curious whether that file should get a line.

handsontable/src/tableView.ts
```diff
@@ -1850,6 +1869,47 @@ class TableView {
...
+   *
+   * Gated on `isOpened()` on purpose. `getActiveEditor()` also answers for an editor that is merely
+   * PREPARED (`prepareEditor()` runs on every cell selection), and editor instances are cached per
+   * class per grid – so a picker parked in the document body by an earlier edit would otherwise
+   * suppress genuine outside clicks for the rest of the instance's life.
+   *
+   * @private
+   * @returns {HTMLElement|null}
+   */
+  #getActiveEditorSurface(): HTMLElement | null {
```
budnix, line 1893: Nothing here checks where the surface actually sits, so an editor that sets `preventCloseElement` to `document.body`, which is easy to do when a picker library hands you its root instead of its popup, makes every click count as inside the grid while that editor is open. Outside-click deselect then stops working with no warning and nothing obvious to blame. The `isOpened()` gate limits how long that lasts, but not how wide it reaches. A `warnOnce` when the surface contains `rootElement` would at least make the mistake visible.

handsontable/src/tableView.ts
```diff
@@ -1824,22 +1825,40 @@ class TableView {
...
+   * written for it – the focus-side test (`#isFocusWithinEditorSurface()`) is what carries the
+   * reported case.
+   *
    * @param {EventTarget[]} eventPath The event propagation path (`event.composedPath()`).
    * @private
    * @returns {boolean}
    */
   #isPathWithinGrid(eventPath: EventTarget[]): boolean {
     const { rootElement, rootPortalElement } = this.hot;
+    const editorSurface = this.#getActiveEditorSurface();
```
budnix: The surface lookup runs before the two cheap `includes` checks that answer almost every call, so every document mousedown pays for a `getActiveEditor()` plus the cross-realm `isHTMLElement` resolution even when the first `includes` would have returned true. Moving it below those two checks keeps the common path as cheap as it was. The same mouseup then resolves the surface up to three times, once in `#isFocusWithinEditorSurface` and once per `#isPathWithinGrid` call, so passing it down would be worth a look.

handsontable/src/tableView.ts
```diff
@@ -619,8 +619,9 @@ class TableView {
       const eventPath = event.composedPath();
       const isPathThroughGridUi = eventPath.includes(rootWrapperElement ?? rootElement) ||
         (this.hot.rootPortalElement && eventPath.includes(this.hot.rootPortalElement));
-      const isFocusLostToOutside = this.hot.getFocusManager().isForeignFocusTarget(activeHTMLElement) ||
-        (!wasInsideGridClick && !this.hot.getFocusManager().hasBrowserFocus() && !isPathThroughGridUi);
+      const isFocusLostToOutside = !this.#isFocusWithinEditorSurface(activeHTMLElement) &&
```
budnix, line 622: Curious whether `isForeignFocusTarget()` was considered as the home for this. It already holds `#hot`, so it could read the open editor's surface itself and every caller would get the same answer. Reading it again though, that on its own would not carry the reported case: the second clause here is true on a panel click too, because `#mouseDown` is only set from the `rootElement` listener and `hasBrowserFocus()` goes false as soon as the focus leaves the grid's focus roots. So the guard would still be needed here, and negating the whole expression looks like the right call.

handsontable/src/tableView.ts
demtario: Verified and applied. `closest()` follows `.host` at `element.ts:122-123` and is already imported here, so the suggestion is a drop-in. The hole was real rather than theoretical: the element does come from `getDeepActiveElement()`, so a picker built as a web component would reproduce this exact symptom with a different library. I also wrote down why `contains()` is wrong here, so it does not get switched back.

Addressed in c66c728f6b.

tests/fixtures/demo/editor-prevent-close-element.html
demtario: You were right, and it was worse than described. I reverted each half separately and rebuilt: **either half alone kept all five cases green**, so the suite proved the fix worked without proving either piece was needed.

Two cases now separate them, each verified red against its own half and green against the other:

- surface assigned in `afterOpen`, so no factory listener exists and the press reaches the document handler. Red without the `#isPathWithinGrid()` branch.
- press inside the panel released outside it (a picker's slider dragged past its edge). The release's event path misses the panel while the focus is still in it. Red without the focus guard.

The fixture takes a `surfaceAssignedIn` param for the first one.

Addressed in c66c728f6b.

tests/fixtures/pages/EditorPreventCloseElementPage.ts
demtario: Correct, my comment was wrong. `isOutsideInput` is `isInput(el) && !('hotInput' in el.dataset)`, and only `selectEditor` and `passwordEditor` stamp `dataset.hotInput` - `editorFactory` never does, so a factory editor's own input reads as an outside input and the handler does not return early.

Rewrote it to name the real reason (the input sits inside `rootElement`, so `isForeignFocusTarget()` reads `false`) and to say explicitly why the early-return reading is wrong, since that is the part the next person would repeat. The `AGENTS.md` entry had the same mistake and is corrected too.

Addressed in c66c728f6b.

tests/fixtures/pages/EditorPreventCloseElementPage.ts
demtario: Verified: `windowTypes.ts:155` does declare `hot` globally with the full instance type. Took the suggestion, and copied `CellDropdownArrowPage`'s comment across so the reason travels with the shape instead of looking like an arbitrary choice.

Addressed in c66c728f6b.

handsontable/src/editors/baseEditor/baseEditor.ts
demtario: Agreed on both counts. The rationale is now a plain comment above the block, and `@type` reads `{HTMLElement|null}`.

One thing worth flagging: `TD` a few fields up has the same mismatch (`@type {HTMLTableCellElement}` against `HTMLTableCellElement | null`), and so do several of its neighbours. I left them alone, so this field now diverges from the local pattern on purpose. Happy to bring the neighbours in line in a follow-up if you would rather they match.

Addressed in c66c728f6b.

handsontable/src/editors/baseEditor/baseEditor.ts
demtario: Added, in the `init` callback that already types `BaseEditor`. It pins both forms the field accepts (an element and `null`), so a refactor that drops the property or narrows it fails the type test rather than going quiet.

Addressed in c66c728f6b.

handsontable/src/tableView.ts
demtario: Taken, and one step further than a warning: a surface that contains `rootElement` is now **refused** as well as reported. A warning on its own still leaves outside-click deselect dead for as long as that editor is open, and the option names an element the editor renders *outside* its container, so an ancestor of the grid can never be a legitimate answer.

`warnOnce` keyed on `rootElement`, which this file already imports and uses. A new spec case pins it: with `document.body` assigned, an outside click still ends the edit and the warning is emitted.

Addressed in c66c728f6b.

handsontable/src/tableView.ts
demtario: Reordered - the surface is resolved only after the two `includes` checks, so a press inside the grid costs exactly what it did before. It has to stay above the `ShadowRoot` bail-out though, since a complete path can carry the surface, and I noted that where it sits.

I left the three-resolutions-per-`mouseup` part alone. Threading the value through `#isPathWithinGrid()`'s signature couples its two call sites for a saving at gesture rate, which did not feel worth the API change. Say so if you disagree and I will do it.

Addressed in c66c728f6b.

handsontable/src/tableView.ts
demtario: Your second reading matches what I found. `hasBrowserFocus()` goes `false` as soon as the focus leaves the grid's focus roots and `#mouseDown` is only set from the `rootElement` listener, so the second clause is true on a panel click and moving the test into `isForeignFocusTarget()` would not carry the case on its own. Kept the guard at the call site, negating the whole expression.

Leaving that method untouched is deliberate for a second reason: it mirrors the focus-scope-manager rule, so every other caller would inherit the change.

Addressed in c66c728f6b.

handsontable/src/tableView.ts
```diff
@@ -1843,13 +1860,88 @@ class TableView {
...
+      warnOnce(
+        this.hot.rootElement,
+        'TableView.preventCloseElementContainsGrid',
+        'The editor\'s `preventCloseElement` contains the grid, so it cannot be told apart from ' +
+        'the page. Assign the picker\'s own popup element instead of an ancestor of the grid. The ' +
+        'element is ignored, and clicks outside the grid keep closing the editor.'
+      );
+
+      return null;
+    }
```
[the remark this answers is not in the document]
demtario: Confirmed and fixed in 71bf6cba06. This was a real bug in the guard, not a false positive - and the same mistake `closest()` had already replaced one helper further down this file, which is the annoying part.

Reproduced it before fixing it: with the grid moved into a shadow root, `document.body.contains(rootElement)` reads `false`, no warning is emitted, and `preventCloseElement` stays set to `document.body` - so `composedPath()` carries it on every click and outside-click deselect is off for the whole edit, exactly as described. The check now walks from the ROOT upwards with `closest()`, which follows `.host` across the boundary, and the reason it cannot be `contains()` is written next to it.

The test almost let this through twice. A shadow-boundary case was added and **passed against the defect it was written for**: the press was resolving to `<html>`, and `composedPath()` for an event on `<html>` does not carry `document.body`, so the wrongly honored surface never turned up in the path. The fixture now has a dedicated out-of-grid target inside `body` whose identity is asserted through `elementFromPoint` before the press, and the case is verified red against `contains()` and green against `closest()`.

tests/fixtures/pages/EditorPreventCloseElementPage.ts
```diff
@@ -0,0 +1,294 @@
...
+      const target = document.elementFromPoint(box.x + (box.width / 2), box.y + (box.height / 2));
+
+      return target === outside;
+    });
+
+    if (!landsOutside) {
+      throw new Error('The out-of-grid target is covered by something else');
+    }
+
+    await this.page.getByTestId('outside-target').click();
```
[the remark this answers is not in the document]
demtario: Half right, and the half that is right is worth knowing - but the conclusion does not hold. I measured it on the fixture rather than reasoning about it, with counting listeners on `documentElement` and an `afterDeselect` hook.

**Correct:** `editorFactory`'s `mousedown` `stopPropagation` is already bound to whatever the editor named, refusing the surface in the view does not unbind it, and on an ancestor of the grid it swallows every press on the page. The document `mousedown` listener count stayed at **0** throughout, for the cell press as well as the outside press.

**Not correct:** the editor does not stay open, so the case is not vacuous. Pressing `#outside-target` blurs the editor's input - it is a plain `div`, so the focus lands on `body` - which drops `hasBrowserFocus()` and trips the *second* clause of `isFocusLostToOutside`, the one that does not go through `isForeignFocusTarget()`. Measured after the press: `afterDeselect` fired once, `getActiveEditor().isOpened()` is `false` and its state is `STATE_VIRGIN`, `getSelected()` is `null`, `document.activeElement` is `BODY`, and the warning is in the log. The advertised contract holds; the close just arrives one event later than it would without the swallowed `mousedown`.

Independently of the reasoning, the case does discriminate: it is red against `contains()` and green against `closest()`, which it could not be if the editor stayed open either way.

No behavior change, but the finding earned a comment: both the guard and the spec case now say that the close comes from the `mouseup` path and why, so the next reader does not assume the `mousedown` path is doing the work. 114ec42890.

budnix (approved):

demtario:
`visually-approved` because of flaky test

Issue #13388, opened 2026-09-04
CopyPaste receives no clipboard events under Salesforce Lightning Web Security (LWC) — works in plain Shadow DOM


GreenFlux:
### Describe the bug

Inside a Salesforce Lightning Web Component running under Lightning Web Security (LWS), the CopyPaste plugin never receives native `copy`/`cut`/`paste` events: `Ctrl/Cmd+C` and `Ctrl/Cmd+V` do nothing in the grid, both for in-grid copy→paste and for pasting external clipboard content. Keydown-based shortcuts on the same selection (e.g. `Ctrl/Cmd+Z` / `Ctrl/Cmd+Y`) work, so keyboard events reach the grid — only the clipboard events don't.

#13194 (DEV-1619, shipped in 18.1.0) states that CopyPaste "binds its clipboard listeners to the grid's shadow root next to the document ones" and that copy/paste works under LWS without wrapper-side glue code. In a real org it does not.

**Steps to reproduce**

1. Upload the 18.1.0 full dist (`handsontable.full.min.js` + CSS) as a static resource in a Salesforce org (any Developer Edition, LWS enabled — the default).
2. Create a minimal LWC that loads it via `platformResourceLoader` and renders a grid into a `lwc:dom="manual"` container:

   ```js
   // hotGrid.js (trimmed to essentials)
   import { LightningElement } from 'lwc';
   import { loadScript, loadStyle } from 'lightning/platformResourceLoader';
   import HOT from '@salesforce/resourceUrl/handsontable';

   export default class HotGrid extends LightningElement {
       initialized = false;
       renderedCallback() {
           if (this.initialized) return;
           this.initialized = true;
           Promise.all([
               loadScript(this, HOT + '/handsontable/handsontable.full.min.js'),
               loadStyle(this, HOT + '/handsontable/handsontable.min.css')
           ]).then(() => {
               new window.Handsontable(this.template.querySelector('.grid'), {
                   data: [['a', 'b'], ['c', 'd']],
                   colHeaders: true,
                   copyPaste: true,
                   undoRedo: true,
                   licenseKey: 'non-commercial-and-evaluation'
               });
           });
       }
   }
   ```

   ```html
   <template>
       <div class="grid" lwc:dom="manual"></div>
   </template>
   ```

3. Place the component on a Lightning page, click a cell, press `Ctrl/Cmd+C`, click another cell, press `Ctrl/Cmd+V`.

[...]

Issue #605, opened 2013-04-26
updateSettings does not work with fixedColumnsLeft


bgrohman:
The updateSettings method does not work with the fixedColumnsLeft setting. When I start with a grid configured with two fixed columns and then call 

``` javascript
$('#grid').handsontable('updateSettings', {fixedColumnsLeft: 0});
```

, the number of fixed columns does not change.

PR #13389, opened 2026-09-07
DEV-2795: Fix copy and paste in a Salesforce Lightning Web Component

budnix:
### Context

The PR fixes copy, cut and paste doing nothing when the grid runs inside a Salesforce Lightning Web Component. Reported in #13388 with a byte-identical build verified across three environments: a plain page (works), a native open shadow root without Salesforce (works), and an LWC under Lightning Web Security (fails).

Lightning Web Security delivers a clipboard event only to listeners bound **at or below the element the grid renders into**. `CopyPaste` bound its `copy`/`cut`/`paste` listeners in two places — `rootDocument` and, since #13194, the grid's shadow root. Both sit above that line, so under LWS neither listener ever ran and the shortcuts silently did nothing. Keydown-based shortcuts such as undo kept working, which is why the grid otherwise looked healthy.

The shadow-root binding added in #13194 looks like it should already cover this. It does not — that is the substance of the report, and the reporter isolated it to LWS specifically.

The fix adds a third binding, on `this.hot.rootElement`, through the same `dedupe()` wrapper the other two already use. It is **additive**:

- The document binding stays. It carries the events that target `document.body` — outside the grid, where nothing bound below would see them — and it is what keeps clipboard working on Chrome 133 and lower (DEV-2277).
- Outside a sandbox the new listener simply sees an in-grid event before the others do, and they skip it through the existing `#processedClipboardEvents` registry. No behavior changes on a normal page or in a plain shadow root.

Two invariants this leans on, both checked and now written down in the plugin's `AGENTS.md`: `EventManager`'s `extendEvent()` mutates and returns the *same* event object, so the identity-keyed `WeakSet` dedupe holds across bindings; and `BasePlugin#disablePlugin()` clears the event manager, so `updateSettings()` does not stack duplicate listeners.

One consequence worth flagging for review: **that registry is now load-bearing for every grid**, not just one inside a shadow tree, because the document and `rootElement` listeners both see every in-grid event. A test now pins it (below).

The three repeated `addEventListener` triples are folded into one `bindClipboardListeners()` helper. The behavioral change is a single added line.

The shadow-root binding is now effectively a duplicate of the `rootElement` one — `rootElement` is always a descendant of its own shadow root, all three listeners are bubble-phase, and the events only the shadow-root listener can see are rejected by `isInternalElement()` anyway. It is **kept deliberately**: removing it is a behavior change in exactly the environment nobody here can test. `AGENTS.md` says so, so a future reader neither deletes it nor re-invents an independent justification for it.

**Changelog:** `.changelogs/13388.json` is already committed here. It is a `public` entry, so per `.changelogs/README.md` it is named after the GitHub issue number rather than the PR number, and needs no post-PR round-trip.

The last commit is unrelated to the fix and can be reviewed on its own: three files pointed at a `Manual QA required` label that does not exist (the real one is `Requires Manual QA`), so a human following the PR template would have created a near-duplicate red label.

- Unit tests added/modified (`*.unit.js`): none — this is DOM event delivery, which the Playwright specs below cover directly
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`): `tests/e2e/shadow-dom.spec.ts`, with `tests/fixtures/demo/shadow-dom.html` and `tests/fixtures/pages/ShadowGridPage.ts`
- For a bug fix — the spec that fails without this fix: `grid whose clipboard events arrive the way LWS delivers them › copies and pastes between cells with keyboard shortcuts`, and the `cuts and pastes…` case beside it

The fixture gains a `?delivery=lws-shape` mode that reproduces the two things LWS does that this plugin feels. Both are needed — with only the first, the tests never exercise the branch that actually carries a paste in an org:

1. **Delivery**, via `stopPropagation` on the container, so the plugin's document and shadow-root listeners never see the event while listeners **on** the container still fire. It must not be `stopImmediatePropagation` (that would cut off the plugin's own container listener) and it must not sit on the shadow root (the plugin binds there too, so the test would pass without the fix).
2. **A collapsed `composedPath()`**, replaced in the capture phase with the shadow host chain. This is what sends `#resolveClipboardEventTarget` down its second branch, onto the retargeted `event.target` — the branch that carries the paste in an org. With an intact path the first branch answers and that half of the method goes untested.

Both new specs were **run against a build without the fix and failed** (target cell unchanged), then passed with it. The discrimination check was re-run after the harness was strengthened, by disabling only `bindClipboardListeners(this.hot.rootElement)`: exactly those 2 specs failed and the other 14 passed.

Three assertions guard against a test that looks green for the wrong reason:

- The fixture records every clipboard event that reaches the document. The LWS-shape specs assert that list is **empty**; the default-delivery spec asserts it **contains `paste`**, so the empty list is not a claim about a recorder that never worked.
- The default-delivery spec asserts `afterPaste` fired **exactly once**. That is the mode where all three listeners see the event, so it is where a broken registry would surface. No value assertion could catch it: under the default `overwrite` mode a double paste writes the same values twice and looks identical. `pasteMode: 'shift_down'` is where it would actually corrupt data.
- The LWS-shape copy spec copies `C3`, a value no other spec in the file copies. The browser clipboard outlives a test (each test gets a fresh context, not a fresh clipboard), so reusing a value an earlier spec copied would let a broken copy still paste the leftover and pass.

**What this does and does not prove.** It verifies the shape of LWS, not LWS: a real org also runs the grid behind a sandbox membrane that no fixture here stands in for — the same caveat #13227 carried. That is why MANUAL QA is ticked below.

### Commands run

```bash
HOT_TEST_PORT=8199 npm --prefix tests run test:e2e -- shadow-dom.spec.ts
#   16 passed (3.4s)          — and 2 failed / 14 passed with the fix line disabled

HOT_TEST_PORT=8199 npm --prefix tests run test:e2e
#   862 passed, 1 skipped (1.2m)

npm --prefix handsontable run test:e2e -- --testPathPattern=copyPaste
#   260 specs, 0 failures

#   Test Suites: 345 passed, 345 total
#   Tests:       4416 passed, 4416 total

node --test .github/scripts/__tests__/manual-qa-gate.test.mjs
#   pass 3, fail 0            — the PR-template edit does not touch the router's regex contract

#   ✖ 578 problems (0 errors, 578 warnings)   — all pre-existing, in unrelated spec files

npm --prefix tests run lint
#   clean
```

`HOT_TEST_PORT` is not decoration. A static server left listening on the default 8123 by another checkout is reused silently (`reuseExistingServer`), and it served that checkout's fixtures and `dist/` — which surfaced as a fixture probe being "not a function" rather than as anything pointing at the port. The runs above are on a private port; `lsof -i :8123` is the check.

### Types of changes

### Related issue(s):

1. #13388
2. DEV-2795
3. #13194 (DEV-1619) — added the shadow-root binding this issue finds insufficient under LWS; #13227 (DEV-2596) — the same class of problem for the Comments plugin

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

- [x] My change requires a change to the documentation. The Shadow DOM guide is updated in this PR: it described the shadow-root check as what copy and paste rely on under LWS, which this change makes untrue.
- [x] MANUAL QA NEEDED — confirm copy, cut and paste in a real Salesforce org under Lightning Web Security; no fixture can reproduce the LWS membrane, so no automated leg here can judge it.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2795

---

> [!NOTE]
> **Medium Risk**
> Changes clipboard listener wiring for all grids; regression would surface as silent failure in LWS or double-paste if dedupe broke, though new e2e guards the latter.
> 
> **Overview**
> Fixes **copy, cut, and paste doing nothing** when the grid runs inside a Salesforce Lightning Web Component under Lightning Web Security. LWS only delivers clipboard events to listeners on or below the grid’s container; listeners on `document` and the shadow root never ran, so shortcuts were silent.
> 
> **CopyPaste** now registers `copy`/`cut`/`paste` on **`rootElement`** as well (via shared `bindClipboardListeners()`), while keeping the document binding (Chrome ≤133, body-targeted events) and the shadow-root binding. **`#processedClipboardEvents`** dedupe is now load-bearing on every grid because document and `rootElement` both see in-grid events.
> 
> **Tests** add a `?delivery=lws-shape` fixture mode (stopped propagation + collapsed `composedPath()`), new shadow-dom e2e cases, and **`afterPaste`** hook counts so a double-paste cannot pass under default `overwrite` mode. **Docs** update the Shadow DOM guide and plugin `AGENTS.md`.
> 
> Also renames contributor guidance from the non-existent **`Manual QA required`** label to the repo’s **`Requires Manual QA`** label (template, `pr-creation` skill, CONTRIBUTING). Changelog: `.changelogs/13388.json`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5a220b3d790cfeff8d811a470d5ab1a2b7e10ca1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2795 [GH-13388] CopyPaste ignores clipboard events under Salesforce Lightning Web Security](https://app.clickup.com/t/9015210959/DEV-2795)
Task linked: [DEV-2277 [#605] Can't get DateTime to format correctly](https://app.clickup.com/t/9015210959/DEV-2277)
Task linked: [DEV-2596 Fix the comment tooltip not opening on hover inside Shadow DOM (#8624)](https://app.clickup.com/t/9015210959/DEV-2596)

Haxikowy (commented):
**Label rename missed two files.** `CONTRIBUTING.md:85` and `.ai/LOCAL-ENFORCEMENT.md:83` still say to apply the `Manual QA required` label. This PR renames it to `Requires Manual QA` in the pr-creation skill, the PR template, `manual-qa.yml`, and root `AGENTS.md`, but these two files point at a label that no longer exists.

**No plain-page paste-count test.** `tests/e2e/clipboard.spec.ts` checks final cell values under the default overwrite mode, which look identical whether the dedupe registry fires once or pastes twice. `rootElement` is now bound on every grid, not just shadow-hosted ones, so that registry is load-bearing everywhere — a future dedupe regression would silently double-paste on an ordinary page and nothing in the suite would catch it.

budnix:
@Haxikowy good spot on both — fixed in `5a220b3d79`: `CONTRIBUTING.md` and `.ai/LOCAL-ENFORCEMENT.md` now say `Requires Manual QA`, and `clipboard.spec.ts` has a plain-page paste-count test. I checked the new test goes red when I disable the dedupe registry, so it would really catch a double paste.

Haxikowy (approved):

PR #13391, opened 2026-09-07
DEV-2782: Reset the fill handle drag state when the plugin is disabled

demtario:
### Context

The PR fixes the fill handle staying armed when the Autofill plugin is disabled in the middle of a pointer gesture.

`Autofill.disablePlugin()` only delegated to the base class. The base class clears the event manager, which removes the `documentElement` `mouseup` listener that owns the drag teardown. So when `updateSettings({ fillHandle: false })` lands between the corner `mousedown` and the `mouseup`, `#onMouseUp()` never runs and `mouseDownOnCellCorner` stays `true`. Re-enabling the fill handle re-adds the `mousemove` listener, `#onMouseMove` sees the stale flag, and the fill border follows a released pointer.

`updatePlugin()` was never affected: it calls `disablePlugin()` and `enablePlugin()` synchronously, so the listeners are back before the `mouseup` arrives. The setting path is different, because `BasePlugin.onUpdateSettings` calls `disablePlugin()` alone when the option turns falsy.

The fix extracts the teardown into a private `#resetDragState()` shared by `#onMouseUp` and `disablePlugin()`. It clears `handleDraggedCells` and the fill preview border (through `resetSelectionOfDraggedArea()`, so a disable does not leave the dashed border painted), `mouseDownOnCellCorner`, `mouseDragOutside`, `#currentDragDirection` and the last pointer position.

Two notes for review:

1. `addingStarted` is deliberately left out of the reset. It mirrors a pending `addRow()` timeout registered through `_registerTimeout`; that timeout still fires after the plugin is disabled and clears the flag itself. Resetting it early would let a re-enabled drag schedule a second `addRow()` and insert two rows. The reasoning is recorded in the plugin's `AGENTS.md`.
2. `updatePlugin()` also calls `disablePlugin()`, so the reset had to be kept away from it. `BasePlugin#isRelevantToSettings()` tests whether a `SETTING_KEYS` entry is *present* in the payload, not whether its value changed, so `updateSettings({ fillHandle: true })` reaches `updatePlugin()` with the value unchanged — and the React wrapper re-sends every declared prop on every re-render (`fillHandle` is not in its `DEEP_COMPARABLE_SETTINGS`). An unconditional reset therefore cancelled a live drag on any ordinary re-render, which would have been a worse regression than the bug being fixed. `#isReconfiguring` marks the disable/enable pair inside `updatePlugin()` and the reset is skipped while it is set: the `mouseup` listener is re-registered in the same tick, so the gesture is genuinely still live. The third E2E case pins that, and it fails against a build with the unconditional reset.

### How has this been tested?

New Playwright spec `tests/e2e/fill-handle-disable-mid-drag.spec.ts`, three cases driven with a real pointer:

- press the fill handle, disable it through `updateSettings`, release, re-enable, then hover with no button held. Asserts no visible `.wtBorder.fill` and `getPlugin('autofill').mouseDownOnCellCorner === false`. Red before the fix (4 fill borders), green after.
- a follow-up drag-fill still works after the disable/re-enable round trip, so a teardown that over-resets would be caught. Not regression coverage on its own (a corner `mousedown` re-arms the state either way), and the spec says so in a comment.
- a drag survives `updateSettings({ fillHandle: true })` re-sending the unchanged value mid-gesture, and the fill commits over the whole dragged extent. Verified red against a build carrying the unconditional reset (the fill never committed, `A3` instead of `A1`).

```bash
# new spec plus the neighbouring fill-handle specs
cd tests && npx playwright test --project=e2e-main \
  e2e/fill-handle-disable-mid-drag.spec.ts \
  e2e/fill-handle-double-click.spec.ts \
  e2e/fill-handle-frozen-panes.spec.ts \
  e2e/fill-handle-formulas.spec.ts
# 12 passed

# the legacy Jasmine autofill suite
npm run build:umd --prefix handsontable
npm run test:e2e.dump --prefix handsontable -- --testPathPattern=plugins/autofill
npm run test:e2e.puppeteer --prefix handsontable -- --testPathPattern=plugins/autofill
# 158 specs, 0 failures, 1 pending

npm run build:types --prefix handsontable
```

### Types of changes

### Related issue(s):

1. DEV-2782

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2782

---

> [!NOTE]
> **Medium Risk**
> Touches Autofill gesture lifecycle and `disablePlugin`/`updatePlugin` interaction; incorrect reset timing could break mid-drag fills on React re-renders or leave stuck drag state.
> 
> **Overview**
> Fixes the fill handle staying armed (and the dashed preview border following the pointer) when **`fillHandle` is turned off via `updateSettings()`** while the corner is still held. Disabling the Autofill plugin removes the document `mouseup` listener before release, so teardown now goes through a shared **`#resetDragState()`** used by **`#onMouseUp`** and **`disablePlugin()`**.
> 
> **`updatePlugin()`** also calls `disablePlugin()`, but framework re-renders often re-send unchanged `fillHandle`, so a **`#isReconfiguring`** flag skips reset during the synchronous disable/enable pair. If the **resolved** fill rules actually change (`directions` or `autoInsertRow`), the gesture is still cleared so release cannot commit under stale constraints.
> 
> Adds Playwright coverage (`fill-handle-disable-mid-drag.spec.ts`) and page helpers for mid-drag settings updates; **`AGENTS.md`** documents the lifecycle rules (including leaving **`addingStarted`** out of reset).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5a8b93616eb4a2294ec3c005f54af807bcdac456. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/handsontable/src/plugins/autofill/autofill.ts
+++ b/handsontable/src/plugins/autofill/autofill.ts
@@ -136,6 +136,12 @@ export class Autofill extends BasePlugin {
    */
   #currentDragDirection: string | null = null;
   /**
+   * Whether the plugin is being torn down and rebuilt by `updatePlugin()` rather than genuinely
+   * disabled. `updateSettings()` reaches `updatePlugin()` whenever `fillHandle` is merely *present*
+   * in the payload, unchanged value included, so this must not end a live gesture.
+   */
+  #isReconfiguring = false;
+  /**
    * Last mouse client position. Stays `null` until the first `mousemove` of a drag, so a scroll
    * that happens after pressing the fill handle but before any drag move is not replayed with a
    * stale or default position.
@@ -203,8 +209,15 @@ export class Autofill extends BasePlugin {
    *  - [`fillHandle`](@/api/options.md#fillhandle)
    */
   updatePlugin(): void {
-    this.disablePlugin();
-    this.enablePlugin();
+    this.#isReconfiguring = true;
+
+    try {
+      this.disablePlugin();
+      this.enablePlugin();
+    } finally {
+      this.#isReconfiguring = false;
+    }
+
     super.updatePlugin();
   }
 
@@ -215,7 +228,12 @@ export class Autofill extends BasePlugin {
     // The base class drops the `documentElement` `mouseup` listener that owns the drag teardown,
     // so a gesture that is in progress when the plugin is disabled would otherwise stay armed
     // until the next `enablePlugin()` picks it up (DEV-2782, the lifecycle twin of GitHub #13370).
-    this.#resetDragState();
+    // A reconfiguration re-registers that listener in the same tick, so the gesture it belongs to
+    // is still live and must survive - see `updatePlugin()`.
+    if (!this.#isReconfiguring) {
+      this.#resetDragState();
+    }
+
     super.disablePlugin();
   }
 
--- a/handsontable/src/plugins/autofill/autofill.ts
+++ b/handsontable/src/plugins/autofill/autofill.ts
@@ -212,6 +212,10 @@ export class Autofill extends BasePlugin {
    * Disables the plugin functionality for this Handsontable instance.
    */
   disablePlugin(): void {
+    // The base class drops the `documentElement` `mouseup` listener that owns the drag teardown,
+    // so a gesture that is in progress when the plugin is disabled would otherwise stay armed
+    // until the next `enablePlugin()` picks it up (DEV-2782, the lifecycle twin of GitHub #13370).
+    this.#resetDragState();
     super.disablePlugin();
   }
 
@@ -880,10 +884,20 @@ export class Autofill extends BasePlugin {
         this.fillIn();
       }
 
-      this.handleDraggedCells = 0;
-      this.mouseDownOnCellCorner = false;
-      this.#currentDragDirection = null;
+      this.#resetDragState();
     }
+  }
+
+  /**
+   * Ends the corner gesture: clears the drag flag, the step counter, the drag direction, and the
+   * fill preview border. Shared by the `mouseup` teardown and `disablePlugin()`.
+   */
+  #resetDragState() {
+    this.resetSelectionOfDraggedArea();
+    this.mouseDownOnCellCorner = false;
+    this.mouseDragOutside = false;
+    this.#currentDragDirection = null;
+    this.#lastMouseClientPosition = null;
   }
 
   /**
--- a/handsontable/src/plugins/autofill/autofill.ts
+++ b/handsontable/src/plugins/autofill/autofill.ts
@@ -4,7 +4,7 @@ import type { CellProperties } from '../../settings';
 import { BasePlugin } from '../base';
 import { Hooks } from '../../core/hooks';
 import { offset, outerHeight, outerWidth } from '../../helpers/dom/element';
-import { isObject } from '../../helpers/object';
+import { isObject, isObjectEqual } from '../../helpers/object';
 import { arrayEach, arrayMap } from '../../helpers/array';
 import { isEmpty } from '../../helpers/mixed';
 import { getCellCoordsFromMousePosition } from '../../helpers/dom/cellCoords';
@@ -139,6 +139,8 @@ export class Autofill extends BasePlugin {
    * Whether the plugin is being torn down and rebuilt by `updatePlugin()` rather than genuinely
    * disabled. `updateSettings()` reaches `updatePlugin()` whenever `fillHandle` is merely *present*
    * in the payload, unchanged value included, so this must not end a live gesture.
+   *
+   * @type {boolean}
    */
   #isReconfiguring = false;
   /**
@@ -209,6 +211,9 @@ export class Autofill extends BasePlugin {
    *  - [`fillHandle`](@/api/options.md#fillhandle)
    */
   updatePlugin(): void {
+    const previousDirections = this.directions;
+    const previousAutoInsertRow = this.autoInsertRow;
+
     this.#isReconfiguring = true;
 
     try {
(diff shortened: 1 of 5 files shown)


qunabu:
Task linked: [DEV-2782 Autofill: disablePlugin() leaves the corner drag state armed mid-gesture](https://app.clickup.com/t/9015210959/DEV-2782)

handsontable/src/plugins/autofill/autofill.ts
```diff
@@ -203,15 +209,31 @@ export class Autofill extends BasePlugin {
...
   /**
    * Disables the plugin functionality for this Handsontable instance.
    */
   disablePlugin(): void {
+    // The base class drops the `documentElement` `mouseup` listener that owns the drag teardown,
+    // so a gesture that is in progress when the plugin is disabled would otherwise stay armed
+    // until the next `enablePlugin()` picks it up (DEV-2782, the lifecycle twin of GitHub #13370).
+    // A reconfiguration re-registers that listener in the same tick, so the gesture it belongs to
+    // is still live and must survive - see `updatePlugin()`.
+    if (!this.#isReconfiguring) {
```
budnix, line 250: Can I ask - was it deliberate to skip the reset for every `updatePlugin()` pass, not just the one the AGENTS.md note is about? That note describes a narrow case, `updateSettings` re-sending the same `fillHandle` value, but a pass where the value really changed skips the reset too. Select A1, drag the handle down to A3, call `updateSettings({ fillHandle: 'horizontal' })`, then release without moving the mouse: the gesture stays alive, `#currentDragDirection` is still `'vertical'`, and `#onMouseUp` fills A2:A3 vertically even though only horizontal is allowed now. This was already the behavior before the PR, but the AGENTS.md diff turns it into a rule for future changes, so comparing the old and new `fillHandle` value with `isObjectEqual` may be worth a look.

tests/e2e/fill-handle-disable-mid-drag.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+    grid = new SelectionFeaturesPage(page, theme);
+    await grid.goto();
+    await grid.initGrid({ data: DATA });
+  });
+
+  test('ends the drag state when the fill handle is disabled between mousedown and mouseup', async () => {
+    await grid.selectCells(0, 0, 0, 0);
+    await grid.pressFillHandle();
+
+    await grid.setFillHandleEnabled(false);
```
budnix, line 32: On this path `#resetDragState()` clears the gesture flags and also clears the fill border through `resetSelectionOfDraggedArea()`, but this test only covers the flags. The handle is disabled right after `pressFillHandle()` and before any pointer move, so `redrawBorders()` never ran and the fill highlight is already empty when `clear()` reaches it. None of the three tests paints a border before disabling, and that is the more damaging shape of the bug, since the old code left the border on screen as well as the gesture armed. Would be good to drag one cell first so the clear actually has something to do:
```suggestion
    await grid.dragPointerToCell(2, 0);
    await grid.setFillHandleEnabled(false);
```

handsontable/src/plugins/autofill/autofill.ts
```diff
@@ -135,6 +135,12 @@ export class Autofill extends BasePlugin {
    * @type {string|null}
    */
   #currentDragDirection: string | null = null;
+  /**
+   * Whether the plugin is being torn down and rebuilt by `updatePlugin()` rather than genuinely
+   * disabled. `updateSettings()` reaches `updatePlugin()` whenever `fillHandle` is merely *present*
+   * in the payload, unchanged value included, so this must not end a live gesture.
+   */
+  #isReconfiguring = false;
```
budnix, line 145: Small thing - this field's JSDoc has no `@type` tag, while the three fields right around it all have one. `mouseDownOnCellCorner` carries `@private @type {boolean}`, `#currentDragDirection` has `@type {string|null}`, and `#lastMouseClientPosition` spells out its object shape. Adding it keeps the class consistent.
```suggestion
   *
   * @type {boolean}
   */
  #isReconfiguring = false;
```

handsontable/src/plugins/autofill/autofill.ts
demtario: Not deliberate, and you are right that it was too broad. The flag was aimed at the no-op re-send and ended up covering every `updatePlugin()` pass, including one where the value really changed. Your repro reproduces exactly as described: `fillIn()` sizes the fill from the highlight rather than from `directions`, so the abandoned vertical drag commits under a setting that now allows only `horizontal`.

Fixed in 5a8b936. `updatePlugin()` captures the resolved configuration around its disable/enable pair and resets the drag state when it moved:

```js
const previousDirections = this.directions;
const previousAutoInsertRow = this.autoInsertRow;
// ... disable + enable under #isReconfiguring ...
if (this.autoInsertRow !== previousAutoInsertRow || !isObjectEqual(previousDirections, this.directions)) {
  this.#resetDragState();
}
```

I compared the resolved config rather than the old and new `fillHandle` values, because the raw values differ in shape for the same configuration: `'vertical'` and `{ direction: 'vertical' }` resolve identically and neither should end a drag. The resolution in `enablePlugin()` also folds in the rule that a horizontal-only direction forces `autoInsertRow` to `false`, which a raw comparison would miss.

Added an E2E case for your scenario, and mutation-tested both directions so neither case is passing by accident: dropping the post-enable reset reds only the new direction-change case, and making the reset unconditional again reds only the same-value re-send case.

The AGENTS.md paragraph now scopes the carve-out to a reconfiguration that changes nothing, and states the direction-change rule explicitly.

tests/e2e/fill-handle-disable-mid-drag.spec.ts
demtario: Agreed, and taken in 5a8b936. The first case now drags one cell before disabling, asserts the border is genuinely painted, then asserts it is gone after the disable:

```js
await grid.dragPointerToCell(2, 0);
await expect(grid.visibleFillBorders()).not.toHaveCount(0);

await grid.setFillHandleEnabled(false);
await expect(grid.fillHandle()).toHaveCount(0);
await expect(grid.visibleFillBorders()).toHaveCount(0);
```

So the `resetSelectionOfDraggedArea()` half of the teardown is now covered rather than reached with nothing to clear.

Following your point through turned up a second problem in the same area: the case that was supposed to prove a same-value re-send keeps the drag alive was not testing that at all. This fixture builds the grid with `fillHandle: { autoInsertRow: false }`, so sending `true` is a real configuration change, not a re-send. There is now a `resendFillHandleSetting()` helper that reads the grid's current value and sends it back verbatim, which is the shape a wrapper produces on an unrelated re-render.

handsontable/src/plugins/autofill/autofill.ts
demtario: Added in 5a8b936, matching the neighbouring fields.

budnix (approved):

PR #13392, opened 2026-09-07
DEV-2525: Stop an auto-height grid growing scrollbars below 100% zoom

budnix:
### Context

The PR fixes a grid configured with `height: 'auto'` growing scrollbars it should never have, whenever the browser renders below 100% — a zoom level under 100%, or Windows display scaling under 100%.

A grid told to size itself to its rows must not scroll itself; it grows and the page scrolls instead. Below 100% it did. The hider (the grid's scroll box) came out a fraction shorter than the table inside it, the browser answered with a vertical scrollbar, that bar took ~15px of width away from columns already stretched to the full width by `stretchH`, and a horizontal scrollbar followed it in. Both bars are full size however small the fraction that summoned them.

**This is a regression in the 18.x line, not long-standing behavior.** Measured across five versions and twelve zoom levels:

| zoom | 16.2.0 | 17.1.0 | 18.0.0 | 18.1.0 | develop (before) | this PR |
|---|---|---|---|---|---|---|
| 100% | clean | clean | clean | clean | clean | clean |
| 90% | clean | clean | both bars | both bars | both bars | clean |
| 85% | clean | clean | both bars | both bars | clean | clean |
| 80% | clean | clean | both bars | both bars | clean | clean |
| 75% | clean | clean | both bars | both bars | both bars | clean |
| 70% | clean | clean | both bars | both bars | clean | clean |
| 67% | clean | clean | both bars | both bars | both bars | clean |
| 60 / 55 / 50 / 110 / 125% | clean | clean | clean | clean | clean | clean |

The levels that pass on the broken builds are the ones where the rounding happens to come out right, which is why users report the problem as intermittent.

DEV-2657 / #13269 already took the bulk of this off on `develop` (it removed a per-row accumulation worth up to 18px), but it landed after the 18.1.0 RC branch was cut, so it is not in any released version. What it left behind is a **constant** sub-pixel shortfall — identical for a 5-row grid and a 40-row one — and that is what this PR removes.

#### Root cause

With `height: 'auto'` the holder keeps `style.height = 'auto'` and simply resolves to its child, so the `.wtHider` height decides everything. It is a sum, written in `SpreaderSize#adjustElementsSize`, and two of its terms dropped their sub-pixel part:

- **the cells' bottom border**, entered as the integer literal `1`. A browser cannot paint a border thinner than one device pixel, so a declared `1px` resolves to `1.11111px` at 90% and `1.49254px` at 67%, and every row renders that fraction taller than the sum accounted for;
- **the column header height**, read through the whole-pixel `offsetHeight`.

Added together those account for essentially the whole gap:

| zoom | border inflation lost | header fraction lost | sum | measured gap |
|---|---|---|---|---|
| 90% | 0.1111 | 0.0972 | 0.2083 | 0.2254 |
| 80% | 0.2500 | 0.2188 | 0.4688 | 0.4690 |
| 75% | 0.3333 | 0.3333 | 0.6667 | 0.6875 |
| 67% | 0.4925 | 0.4776 | 0.9701 | 0.9799 |

#### The change

| File | Change |
|---|---|
| `axisSizing/hiderCompensation.ts` (new) | `getHiderHeightCompensation()` adds the amount by which the browser **inflated** the cells' bottom border past the whole pixel the row heights were summed with, read from the cached style snapshot (no DOM read). `addContentHeightSlack()` adds a fixed 0.05px to a fractional total. |
| `viewport/workspaceSize.ts` | A cached `getColumnHeaderHeightFraction()` twin supplies the sub-pixel part `getColumnHeaderHeight()` rounds away. The integer getter is untouched — the calculators depend on it. |
| `overlay/spreaderSize.ts` | Folds in both terms. `expandHiderVerticallyBy` reads the height back with `parseFloat`, not `parseInt`, which would truncate the correction straight off again. |
| `viewport/boxLayout/gatherLayoutInput.ts` | The single-pass solver re-derives the same total independently, so it folds in the identical terms and its prediction keeps matching what is written to the DOM. |
| `types.ts`, `test/helpers/common.js` | `getStyleForTD` declared optional on the engine's minimal `StylesHandler` interface, and implemented in the Puppeteer harness stub. |

**Nothing changes at 100% zoom with no display scaling.** Every term is 0 there and the slack is skipped for a whole-pixel total, which is what the green suites below assert.

The horizontal scrollbar needed no separate fix for the reported case — it was only ever a consequence of the unwanted vertical one, and the measured horizontal overflow is 0 at every level once the vertical bar is gone.

#### Three decisions a reviewer should check

1. **Add the inflation to the historical `1`; never return the border itself.** The menu grids and the Filters by-value list drop all four borders on `td:first-child`, and `StylesHandler`'s probe cell *is* a `td:first-child` inside them, so the border reads `0px` there — at 100% zoom, on every platform. Returning it would have resized those grids by a whole pixel at the default zoom. The same trap runs the other way at 50% (`2px`) and above 100% (`0.8px`), where the row sum already carries the whole pixel. The `> Math.round(…)` gate is deliberately the one `StylesHandler#calculateRowHeight` uses, so the two agree about which borders the rows already account for.
2. **A fixed sub-pixel slack, not a rounding up to the device-pixel grid.** The device-pixel version reads as the principled choice, but this total is also what the browser weighs against a fixed `height`: a 264px box holding 264.22px of rows at 80% zoom shows no bar because the browser rounds the deficit away, and inflating the content by most of a device pixel pushes it over into a full 15px bar. Measured, then replaced. Deciding it from the holder's own height cannot work either — with `height: 'auto'` the holder resolves *from* the hider being written, so the test would read the previous draw.
3. **`getBoundingClientRect()` and `offsetHeight` are in different coordinate spaces.** A rect is scaled by an ancestor's CSS `zoom`; `offsetHeight` and the computed style are not. The header fraction therefore comes from `getComputedStyle().height`, and is clamped to ±1px — the computed height is the content box while `offsetHeight` is the border box, so a host stylesheet putting a border on `thead` would otherwise turn the "fraction" into several whole pixels of real box, subtracted straight out of the hider.

Two smaller notes: `stylesHandler` is a user-supplied setting that defaults to `null`, so `getStyleForTD` is optional and guarded — calling it unguarded threw inside the draw and took out 695 of 816 walkontable specs. And `getColumnHeaderHeight()`'s cache is reset per draw by `createColumnsCalculator`, so the extra computed-style read runs every draw; it adds a resolved style, not a second reflow, because the `outerHeight` above it has already flushed layout.

All of the above is recorded in `handsontable/src/3rdparty/walkontable/AGENTS.md`.

#### Known limitation, deliberately out of scope

A zoom applied **after** the grid is built is still not followed. `StylesHandler` reads the cells' border once from a snapshot taken at init, so a grid built at 100% and then zoomed out keeps its original row heights — the limitation DEV-2657 documented in its own comment. Measured on a 40-row grid, vertical overflow with the zoom applied post-load:

| zoom | 18.1.0 | this PR |
|---|---|---|
| 90% | 4px | 3px |
| 80% | 9px | 8px |
| 67% | 18px | 17px |

Better at every level, never worse, but not fixed — the dominant term there is the frozen row-height cache. Chrome remembers zoom per origin, so a returning user loads at their zoom level and gets the fixed path; the gap is the first live Ctrl− press. It needs the row-height cache and the style snapshot invalidated on a zoom change, and wants its own ticket.

### How has this been tested?

New tests, both verified failing-first. With the fix reverted, all eight `the scroll box holds the whole table` assertions fail.

- `handsontable/src/3rdparty/walkontable/test/unit/axisSizing/hiderCompensation.unit.ts` — 31 cases over the two pure helpers: the borderless / 50% / 125% cases that must stay at the historical `1`, the partial-`stylesHandler` case that caused the 695 failures, and the whole-pixel totals the slack must leave untouched.
- `tests/e2e/auto-height-scrollbar-zoom.spec.ts` plus its fixture and page object — eight cases across four zoom levels and two row counts, one navigation each. The file documents which assertion discriminates: the scroll-box-vs-table one, since this fixture drives the zoom through CSS `zoom` (which leaves `devicePixelRatio` at 1, where Chrome rounds a sub-pixel overflow away instead of painting a bar). The scrollbar assertions are the user-visible backstop.

The twelve-device-scale evidence above comes from a standalone harness that cannot run inside these Playwright projects: `--force-device-scale-factor` needs `viewport: null`, and every project's `devices['Desktop Chrome']` pins `deviceScaleFactor` instead — the constraint `tests/AGENTS.md` documents. Playwright's context `deviceScaleFactor` and CDP `Emulation.setDeviceMetricsOverride` both report the ratio faithfully but never inflate the border, so a test built on either passes on the unfixed code.

Full suites, all green:

```bash
npm run test:walkontable --prefix handsontable   # 816 specs, 0 failures
npm run test:unit --prefix handsontable          # 4456 tests, 0 failures
npm run test:e2e --prefix handsontable           # 9894 specs, 0 failures, 8 pending
npm run test:types --prefix handsontable         # pass
npm run stylelint --prefix handsontable          # 0 errors
npm run build --prefix handsontable              # pass, incl. the strict postbuild gate

cd tests && npx playwright test --project=e2e-main   # 910 passed, 1 skipped, 0 failed
```

Chrome only, per the project's browser-testing rule. Edge shares the Blink layout engine, so the pixel-snapping behavior carries.

### Types of changes

### Related issue(s):

1. DEV-2525
2. Builds on DEV-2657 / #13269, which is on `develop` but missed the 18.1.0 release cut.

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2525

---

> [!NOTE]
> **Medium Risk**
> Changes the master hider height calculation and layout input path in Walkontable; behavior is unchanged at 100% zoom but affects scroll/overflow at sub-100% scaling, with broad unit and e2e coverage mitigating regression risk.
> 
> **Overview**
> Fixes **DEV-2525 / #13392**: grids with `height: 'auto'` no longer grow vertical (and follow-on horizontal) scrollbars when the browser renders below 100% zoom or display scaling.
> 
> Walkontable now sizes the master **hider** with sub-pixel-aware totals instead of a whole-pixel sum. A new **`hiderCompensation`** module supplies border inflation compensation (historical `1px` plus device-pixel inflation read via optional `stylesHandler.getStyleForTD`) and a **0.05px slack** on fractional totals for the DOM write only. **`SpreaderSize#adjustElementsSize`** folds in **`getColumnHeaderHeightFraction()`** and uses **`parseFloat`** when expanding the hider so fractional heights are not truncated. **`gatherLayoutInput`** applies the same compensation and header fraction so layout prediction stays aligned, but deliberately omits the slack.
> 
> Supporting changes: optional **`getStyleForTD`** on the engine `StylesHandler` type with harness stubbing, expanded **AGENTS.md** guidance, unit tests for the helpers, and Playwright e2e coverage across zoom levels and row counts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 207e2c26f6c2af68dea8ec5da63f1da0cafb71cc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changelogs/13392.json
+++ b/.changelogs/13392.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Fixed a grid with `height: 'auto'` showing a vertical scrollbar, and a horizontal one behind it, when it was rendered below 100% browser zoom or system display scaling.",
+  "type": "fixed",
+  "issueOrPR": 13392,
+  "breaking": false,
+  "framework": "none"
+}
--- a/handsontable/src/3rdparty/walkontable/AGENTS.md
+++ b/handsontable/src/3rdparty/walkontable/AGENTS.md
@@ -124,6 +124,49 @@ Three more things that pass every functional test and only show up in a profile 
 
 When you add a new content-driven measurement, ask which tables actually render the content — measuring the master alone is the trap both of these exist to work around.
 
+## The hider height is fractional below 100% zoom, and two coordinate spaces meet there
+
+`SpreaderSize#adjustElementsSize` writes the master hider's height, and below 100% browser zoom or
+display scaling it writes a **fractional** value (`"1210.45px"`). That element is what decides
+whether the grid shows a scrollbar: with `height: 'auto'` the holder keeps `style.height = 'auto'`
+and simply resolves to the hider, so a hider a fraction shorter than the table inside it is a
+vertical scrollbar on a grid that must never scroll — and, because `stretchH` already sized the
+columns against the full width, a horizontal one behind it (DEV-2525).
+
+Four rules come out of that, and each of them was a defect first.
+
+- **Never `parseInt` a value read back off `hider.style.height`.** It truncates the fraction and
+  hands back the shortfall. `expandHiderVerticallyBy` does `parseFloat`. (`topOverlay.ts`'s
+  `parseInt(holderParent.style.height, 10)` reads the *clone's* holder parent, a different element,
+  and is not this.)
+- **The sum's fixed terms must carry their fraction too.** `getHiderHeightCompensation`
+  (`axisSizing/hiderCompensation.ts`) reads the cells' *rendered* `border-bottom-width` — a browser
+  cannot paint a border thinner than one device pixel, so a declared `1px` resolves to 1.11111px at
+  90% and 1.49254px at 67% — and `Viewport#getColumnHeaderHeightFraction()` supplies what
+  `getColumnHeaderHeight()`'s integer `offsetHeight` rounded away. Both are 0 at 100%.
+- **`getBoundingClientRect()` and `offsetHeight` are NOT in the same coordinate space.** A rect is
+  scaled by an ancestor's CSS `zoom`; `offsetHeight`, `clientHeight` and the computed style are not.
+  Subtracting one from the other to get a fraction looks right under
+  `--force-device-scale-factor` (where they agree) and yields a large **negative** number under CSS
+  `zoom`, shrinking the hider instead of growing it. Take the fractional height from
+  `getComputedStyle(el).height`, which matches `offsetHeight`'s space at any zoom. The same trap
+  ruins measurements in tests — never compare a rect against a `clientHeight`.
+- **`gatherLayoutInput` re-derives the same total independently**, so both sites fold in the same two
+  terms and the same device-pixel rounding. A prediction built on whole pixels while the DOM carries
+  fractional ones disagrees exactly at the knife edge, which is a scrollbar the solver said would not
+  be there.
+
+`snapUpToDevicePixel()` rounds the finished total up to the next device pixel, because summing the
+rows reproduces the browser's own sub-pixel snapping only to about 0.02px and that is enough to
+summon a full-size bar on a small grid at 67%. It is a no-op at 100%, where the totals are whole
+pixels already.
+
+**`stylesHandler` is a user-supplied setting that defaults to `null`, and a standalone Walkontable
+host may implement only part of it** — the engine's own Puppeteer harness (`test/helpers/common.js`)
+does. `getStyleForTD` is therefore declared optional on the `StylesHandler` interface in `types.ts`,
+and calling it unguarded threw inside the draw and took out 695 of 816 specs. Guard every method you
+add a dependency on, and teach the harness stub the same method.
+
 ## Column-axis border ownership: the row header owns its gridline
 
 The two axes are NOT symmetric, and the column axis is the settled one. On the column axis a row
--- a/handsontable/src/3rdparty/walkontable/AGENTS.md
+++ b/handsontable/src/3rdparty/walkontable/AGENTS.md
@@ -140,10 +140,20 @@ Four rules come out of that, and each of them was a defect first.
   `parseInt(holderParent.style.height, 10)` reads the *clone's* holder parent, a different element,
   and is not this.)
 - **The sum's fixed terms must carry their fraction too.** `getHiderHeightCompensation`
-  (`axisSizing/hiderCompensation.ts`) reads the cells' *rendered* `border-bottom-width` — a browser
-  cannot paint a border thinner than one device pixel, so a declared `1px` resolves to 1.11111px at
-  90% and 1.49254px at 67% — and `Viewport#getColumnHeaderHeightFraction()` supplies what
+  (`axisSizing/hiderCompensation.ts`) adds the amount by which the browser *inflated* the cells'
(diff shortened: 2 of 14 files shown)


qunabu:
Task linked: [DEV-2525 Browser zoom issue in v16.2](https://app.clickup.com/t/9015210959/DEV-2525)
Task linked: [DEV-2657 Fix the grid rendering taller than its own container below 100% zoom](https://app.clickup.com/t/9015210959/DEV-2657)

demtario (approved):

PR #13393, opened 2026-09-07
DEV-2786: Drop the dead position-change report on the inline-start axis

demtario:
### Context

The PR removes a nested reconciliation draw that a horizontal scroll paid for nothing.

`ctx.positionChanged` in `walkontable/src/table/drawCycle.ts` exists for one purpose: an `innerBorder*` class toggle shifts the layout by 1px, so the draw has to reconcile the overlays against the shifted layout by calling `refreshAll()` — a nested `wot.draw(true)` over the master and every clone.

Since #6673 (PR #13371) the inline-start axis no longer shifts by that pixel. The row header carries its `border-inline-end` at every scroll position and no cell behind it carries a `border-inline-start`, so `innerBorderInlineStart` (and `innerBorderLeft`, `emptyColumns`) drive no geometry and no stylesheet reads them; they are stamped for backward compatibility only. `InlineStartOverlay#resetFixedPosition` nevertheless kept reporting a position change whenever it toggled that class, and `placeFixedOverlays` ORed the result into the flag. Every scroll crossing horizontal offset 0 therefore ran the full reconciliation to fix a shift that cannot happen.

The overlay now reports `false` unconditionally and `placeFixedOverlays` no longer reads its result; only the top and bottom overlays feed the flag. The class stamping is untouched, so the public DOM surface is unchanged.

Grids on the single-pass layout path did not pay this, because `prepareHeaderBorders` applies the class before the cells render and the post-render toggle then found it already in place. The cost landed on the grids that drop off that path, and there are **two** ways to do that, both of which paid: breaking the uniform-size requirement (a `colWidths` array, or an active `manualColumnResize`), and breaking the element-mode requirement (a window-scrolled grid). The window shape is the harsher one, since `prepareHeaderBorders` bails on `trimmingContainer === rootWindow` outright and so could never have pre-applied the class whatever the other settings said. Both have a leg in the spec.

This is item 2 of DEV-2786. The row axis keeps its 1px shift and its compensations, and is the scope of a separate PR; the `AGENTS.md` bullet describing it is deliberately left in place.

### Why this is user-facing

An earlier revision of this description claimed `[skip changelog]` on the grounds that no released version paid the cost. That was wrong, and the review caught it. The released `18.1.0` tag still carries `ctx.positionChanged = wtOverlays.inlineStartOverlay.resetFixedPosition() || ctx.positionChanged` at `drawCycle.ts:554`, so released grids do run this nested draw on every crossing. What #13371 changed is that the draw stopped fixing anything — my `git tag --contains` check proved the *waste* is unreleased, not the *draw*.

Since #13371 and this change ship together, the net effect against 18.1.0 is a genuine scroll improvement, so this carries a `changed` changelog entry rather than a skip.

### How has this been tested?

New Playwright coverage in `tests/e2e/walkontable/inline-start-border-refresh.spec.ts`, with a fixture and page object. It measures the draw cost of one round trip across horizontal offset 0, on the single-pass path and off it, and separately asserts that the overlays stay correctly sized and aligned and that the compatibility classes are still stamped.

Counting `refreshAll` outright cannot see this: `ScrollSync` calls it once per scroll event as the normal scroll response, so the count is at least 1 per crossing either way and that baseline hides the difference. The reconciliation draw is the **re-entrant** call — the scroll-driven one runs `wot.draw(true)`, and a draw that sees `positionChanged` calls `refreshAll` again from inside it — so the fixture counts by nesting depth and reports the two kinds separately. The depth is counted on `refreshAll` itself rather than on `draw`: measured, `wtOverlays.wot !== hot.view._wt`, and a `draw` patched on the instance `hot.view._wt` hands you counts zero calls even for a `refreshAll()` invoked directly. `ScrollSync` holds `refreshAll` as a late-bound closure, so replacing the method is observed through its call too.

The negative control is the two off-path reconciliation assertions — the `colWidths` array and the window-scrolled shape — both of which fail `Expected: 0, Received: 1` against `origin/develop`'s source with the bundle rebuilt. The uniform-width case is labelled in the spec as a control rather than a regression, because that path already cost nothing. The overlay-geometry assertions pass with and without the fix; they guard the `wtOverlays.adjustElementsSize()` call that shared the disconnected branch, and are not themselves proof of the fix.

The file also carries one **positive** assertion, so it cannot pass vacuously: a vertical crossing off the single-pass path must cost exactly **one** re-entrant `refreshAll`. Without it, every reconciliation expectation here is `toBe(0)` and a later change that deleted `ctx.positionChanged` and its whole branch would keep the whole file green. That expectation also marks this change's boundary — when the row axis is brought into line it becomes 0 like the others.

Two shapes the metrics could not otherwise reach are covered explicitly: a crossing whose cell render is cancelled via `skipRender` (reachable outside tests through `NestedRows`, and the case that used to get the master `adjustElementsSize()` from the removed branch), and the backward-compatibility classes still being stamped in both scroll modes.

```bash
# the new spec, all six theme x bundle legs
cd tests && npx playwright test walkontable/inline-start-border-refresh

# every walkontable Playwright spec, all six legs
cd tests && npx playwright test walkontable

# the Jasmine walkontable suite

# core E2E around scrolling, overflow and column resizing
npm run test:e2e --prefix handsontable -- --testPathPattern='scroll|preventOverflow|manualColumnResize'

cd tests && npm run lint
```

Results: 48 passed (new spec, 6 legs) · 192 passed (all walkontable Playwright, 6 legs) · 816 specs / 0 failures (Jasmine walkontable) · 386 specs / 0 failures (core E2E subset) · eslint, stylelint, `test:types` and the Playwright-tier lint all clean.

### Types of changes

### Related issue(s):

1. DEV-2786

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Touches the master draw cycle’s `positionChanged` → `refreshAll` reconciliation path; incorrect narrowing could skip needed overlay refreshes, though e2e tests assert geometry and that vertical (row-axis) reconciliation still runs once.
> 
> **Overview**
> Walkontable no longer treats toggling **`innerBorderInlineStart`** on the inline-start overlay as a 1px layout change. **`InlineStartOverlay#resetFixedPosition`** and **`adjustHeaderBordersPosition`** always return **`false`**, and **`placeFixedOverlays`** in **`drawCycle.ts`** stops OR-ing that overlay into **`ctx.positionChanged`**. Only top/bottom overlays still drive the nested **`refreshAll()`** reconciliation path.
> 
> **Backward-compatibility class stamping on `.ht_master` is unchanged**; the fix only removes an extra nested **`wot.draw(true)`** that ran when horizontal scroll crossed offset 0 (especially on layouts off the single-pass path, e.g. varied **`colWidths`** or window scrolling).
> 
> Adds Playwright coverage (**`inline-start-border-refresh.spec.ts`**) that counts re-entrant vs scroll-driven **`refreshAll`** calls and checks overlay geometry, plus changelog and Walkontable lifecycle docs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bfc99696ef62bd587e7b6d82dac3f8dadde7b60c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/handsontable/src/3rdparty/walkontable/AGENTS.md
+++ b/handsontable/src/3rdparty/walkontable/AGENTS.md
@@ -149,6 +149,31 @@ Consequences worth knowing:
   reads them any more. `innerBorderInlineStart` also only toggles when the grid has row headers and NO
   frozen columns, so it is not a usable "has scrolled" signal in a test — poll the holder's
   `scrollLeft` instead.
+- **`InlineStartOverlay#resetFixedPosition` reports `false` unconditionally**, and `placeFixedOverlays`
+  no longer ORs it into `ctx.positionChanged` — only the top and bottom overlays feed that flag now.
+  The class shifts no layout, so the flag's one job (reconciling a 1px shift) had nothing to do, yet
+  every scroll crossing horizontal offset 0 ran `refreshAll()`: a nested `wot.draw(true)` over the
+  master and every clone. Grids on the single-pass path did not pay it, because `prepareHeaderBorders`
+  applies the class before the cells render and the post-render toggle then finds it already in place.
+  Those two gates line up exactly, which is worth knowing before you assume a window-scrolled grid slips
+  between them: `usesLayoutSnapshotForCalculators()` requires `!isHorizontallyScrollableByWindow()`,
+  that predicate IS `inlineStartOverlay.trimmingContainer === rootWindow`, and `prepareHeaderBorders`
+  bails on the same test against the same overlay — so anything passing the single-pass gate is
+  element-scrolled and never bails. `preventOverflow` does not enter into it; it changes how
+  `resetFixedPosition` uses the trimming container, not what the container is. The cost therefore landed
+  on the grids that drop off that path, which `colWidths` as an array or an active `manualColumnResize`
+  is enough to do.
+- **Counting `refreshAll` cannot measure a reconciliation draw on its own.** `ScrollSync` calls
+  `refreshAll` once per scroll event as the normal response to a scroll (`overlay/scroll/scrollSync.ts`),
+  so a per-crossing count is at least 1 whether or not anything reconciled, and that baseline hides the
+  difference. The reconciliation is the **re-entrant** call: the scroll-driven one runs `wot.draw(true)`,
+  and a draw that sees `positionChanged` calls `refreshAll` again from inside it. Count by nesting depth
+  **on `refreshAll` itself**, which is reachable: `ScrollSync` holds it as a late-bound closure
+  (`refreshAll: () => overlays.refreshAll()`), so replacing the method is observed through that call too.
+  Do **not** try to get the depth by patching `draw` on the instance `hot.view._wt` hands you:
+  `wtOverlays.wot !== hot.view._wt`, so a patched `_wt.draw` counts **zero** calls even for a
+  `refreshAll()` invoked directly, which definitely runs `this.wot.draw(true)`.
+  `tests/e2e/walkontable/inline-start-border-refresh.spec.ts` measures it this way.
 - **The row axis is unchanged.** `innerBorderTop` / `innerBorderBottom` still shift the layout by 1px,
   which is what `positionChanged` and the reconciliation draw in `table/drawCycle.ts` exist for, and
   `columnHeaderBorderCompensation` in `topOverlay`/`spreaderSize` is the vertical twin that stayed.
--- a/handsontable/src/3rdparty/walkontable/AGENTS.md
+++ b/handsontable/src/3rdparty/walkontable/AGENTS.md
@@ -160,9 +160,13 @@ Consequences worth knowing:
   that predicate IS `inlineStartOverlay.trimmingContainer === rootWindow`, and `prepareHeaderBorders`
   bails on the same test against the same overlay — so anything passing the single-pass gate is
   element-scrolled and never bails. `preventOverflow` does not enter into it; it changes how
-  `resetFixedPosition` uses the trimming container, not what the container is. The cost therefore landed
-  on the grids that drop off that path, which `colWidths` as an array or an active `manualColumnResize`
-  is enough to do.
+  `resetFixedPosition` uses the trimming container, not what the container is. The cost therefore
+  landed on the grids that drop off that path, and there are TWO ways to do that, both of which paid:
+  breaking the uniform-size requirement (`colWidths` as an array, or an active `manualColumnResize`),
+  and breaking the element-mode requirement (a window-scrolled grid, i.e. no `width`/`height`). The
+  window shape is the harsher one, since `prepareHeaderBorders` bails on it outright and so could
+  never have pre-applied the class whatever the other settings said. Both have a leg in
+  `tests/e2e/walkontable/inline-start-border-refresh.spec.ts`.
 - **Counting `refreshAll` cannot measure a reconciliation draw on its own.** `ScrollSync` calls
   `refreshAll` once per scroll event as the normal response to a scroll (`overlay/scroll/scrollSync.ts`),
   so a per-crossing count is at least 1 whether or not anything reconciled, and that baseline hides the
--- a/handsontable/src/3rdparty/walkontable/src/overlay/regions/inlineStartOverlay.ts
+++ b/handsontable/src/3rdparty/walkontable/src/overlay/regions/inlineStartOverlay.ts
@@ -2,7 +2,6 @@ import type { TableDeps } from '../../table/baseTable';
 import {
   addClass,
   getScrollLeft,
-  hasClass,
   removeClass,
   setOverlayPosition,
   resetCssTransform,
@@ -60,7 +59,8 @@ export class InlineStartOverlay extends Overlay {
   /**
    * Updates the left overlay position.
    *
-   * @returns {boolean}
+   * @returns {boolean} Always `false` - this overlay's header-border classes shift no layout, so it
+   * never reports a position change. See `InlineStartOverlay#adjustHeaderBordersPosition`.
    */
   resetFixedPosition() {
     const wtTable = this.deps.getWtTable();
@@ -88,11 +88,11 @@ export class InlineStartOverlay extends Overlay {
       resetCssTransform(overlayRoot);
     }
 
-    const positionChanged = this.adjustHeaderBordersPosition(overlayPosition);
+    this.adjustHeaderBordersPosition(overlayPosition);
 
     this.adjustElementsSize();
 
-    return positionChanged;
+    return false;
   }
 
   /**
@@ -410,9 +410,10 @@ export class InlineStartOverlay extends Overlay {
   }
 
   /**
-   * Pre-applies the `innerBorderInlineStart` class before the cell render (single-pass gated
-   * path), so the post-render `resetFixedPosition` toggle is a no-op and the nested re-draw is
-   * skipped. Element mode only — see `TopOverlay#prepareHeaderBorders`.
+   * Pre-applies the `innerBorderInlineStart` class before the cell render (single-pass gated path),
+   * so a `beforeViewRender` listener sees the class in the state this draw ends in. Unlike the top
+   * overlay's copy, this one saves no re-draw: the class shifts no layout, so the post-render toggle
+   * reports nothing either way. Element mode only — see `TopOverlay#prepareHeaderBorders`.
    */
   prepareHeaderBorders() {
     if (!this.needFullRender || !this.shouldBeRendered() ||
@@ -425,10 +426,17 @@ export class InlineStartOverlay extends Overlay {
   }
 
   /**
-   * Adds css classes to hide the header border's header (cell-selection border hiding issue).
+   * Stamps the `innerBorderInlineStart` / `innerBorderLeft` / `emptyRows` classes on the master.
+   *
+   * Always reports `false`: the row header carries its inline-end border at every scroll position
+   * and no cell behind it carries an inline-start border (#6673), so none of these classes moves the
+   * layout by a pixel and no stylesheet reads them. They are stamped for backward compatibility
+   * only. Reporting a position change made every scroll crossing horizontal offset 0 run
+   * `refreshAll()` - a nested `wot.draw(true)` over the master and every clone - to reconcile a
+   * 1px shift that cannot happen.
    *
    * @param {number} position Header X position if trimming container is window or scroll top if not.
-   * @returns {boolean}
+   * @returns {boolean} Always `false`.
    */
   adjustHeaderBordersPosition(position: number) {
     const masterParent = this.deps.getWtTable().holder.parentNode as HTMLElement;
@@ -447,45 +455,38 @@ export class InlineStartOverlay extends Overlay {
       removeClass(masterParent, 'innerBorderLeft innerBorderInlineStart');
     }
 
-    return state.positionChanged;
+    return false;
   }
 
   /**
    * Computes the inline-start overlay's header-border state without mutating the DOM. Pure: reads
-   * settings and the current class state only. Splitting the decision from the write lets the
-   * single-pass draw resolve the `innerBorderInlineStart` toggle before rendering, instead of after.
+   * the settings only. Splitting the decision from the write lets the single-pass draw resolve the
+   * `innerBorderInlineStart` toggle before rendering, instead of after.
+   *
+   * Reports no position change, which is why it does not read the current class state either: the
+   * class shifts nothing. See `InlineStartOverlay#adjustHeaderBordersPosition`.
    *
    * @param {number} position The overlay offset that decides whether the inline-start border is shown.
-   * @returns {{ hasEmptyRows: boolean, innerBorder: string, positionChanged: boolean }}
+   * @returns {{ hasEmptyRows: boolean, innerBorder: string }}
    */
   #computeHeaderBordersState(position: number) {
     const { wtSettings } = this;
-    const masterParent = this.deps.getWtTable().holder.parentNode as HTMLElement;
     const rowHeaders = wtSettings.getSetting('rowHeaders') as ((...args: unknown[]) => unknown)[];
     const fixedColumnsStart = wtSettings.getSetting<number>('fixedColumnsStart');
     const totalRows = wtSettings.getSetting<number>('totalRows');
     const preventVerticalOverflow = wtSettings.getSetting('preventOverflow') === 'vertical';
     const hasEmptyRows = !totalRows;
     let innerBorder = 'keep';
-    let positionChanged = false;
 
     if (!preventVerticalOverflow) {
       if (fixedColumnsStart && !rowHeaders.length) {
         innerBorder = 'add';
 
       } else if (!fixedColumnsStart && rowHeaders.length) {
-        const previousState = hasClass(masterParent, 'innerBorderInlineStart');
-
-        if (position) {
-          innerBorder = 'add';
-          positionChanged = !previousState;
-        } else {
-          innerBorder = 'remove';
-          positionChanged = previousState;
-        }
+        innerBorder = position ? 'add' : 'remove';
       }
     }
 
-    return { hasEmptyRows, innerBorder, positionChanged };
+    return { hasEmptyRows, innerBorder };
   }
 }
--- a/handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
+++ b/handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
@@ -523,9 +523,9 @@ function renderActiveSelections(table: Table, runFastDraw: boolean): void {
  *
  * Only the TOP and BOTTOM overlays contribute to the flag. The corners never did, and the
  * inline-start overlay no longer does: its `innerBorderInlineStart` class shifts no layout since
- * #6673, so it always reported `false` while still forcing the OR - and on the draws where the class
- * did toggle, a nested reconciliation draw over the master and every clone for a 1px shift that
- * cannot happen. It is still repositioned here, its result is just not read.
+ * #6673, so its report meant nothing — `false` on most draws, and `true` on the draws where the
+ * class toggled, which cost a nested reconciliation draw over the master and every clone for a 1px
+ * shift that cannot happen. It is still repositioned here, its result is just not read.
  *
  * @param {Table} table The master table.
  * @param {DrawContext} ctx The per-draw scratch (receives `positionChanged`).
--- a/handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
+++ b/handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
@@ -123,9 +123,10 @@ function runMasterDrawCycle(table: Table, ctx: DrawContext): void {
   // (`innerBorderTop` / `innerBorderInlineStart`) BEFORE resolving the snapshot and rendering the
   // cells, from the pre-render scroll position + settings. Cells then render in their final
   // position, so the post-render `resetFixedPosition` toggle is a no-op (`positionChanged` stays
-  // `false`) and the nested `wot.draw(true)` re-render never fires. Element mode only (guaranteed
-  // by the gate); the border box is thus present when `beginDrawLayout` measures the workspace,
-  // matching a steady-state scrolled draw.
+  // `false`) and the nested `wot.draw(true)` re-render never fires. That saving is `innerBorderTop`'s
+  // alone — the inline-start class shifts no layout and reports nothing whenever it is applied
+  // (#6673). Element mode only (guaranteed by the gate); the border box is thus present when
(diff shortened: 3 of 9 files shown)


qunabu:

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -518,8 +519,13 @@ function renderActiveSelections(table: Table, runFastDraw: boolean): void {
...
  * Master-only fixed-position pass: repositions the top / bottom / inline-start / corner overlays and
- * records in `ctx.positionChanged` whether an `innerBorder*` toggle shifted the layout by 1px (the
- * corners do not contribute to the flag, matching the original).
+ * records in `ctx.positionChanged` whether an `innerBorder*` toggle shifted the layout by 1px.
+ *
+ * Only the TOP and BOTTOM overlays contribute to the flag. The corners never did, and the
+ * inline-start overlay no longer does: its `innerBorderInlineStart` class shifts no layout since
+ * #6673, so it always reported `false` while still forcing the OR - and on the draws where the class
+ * did toggle, a nested reconciliation draw over the master and every clone for a 1px shift that
+ * cannot happen. It is still repositioned here, its result is just not read.
```
budnix: On develop this overlay returned `true` whenever the class toggled, so saying it always reported `false` reads as the opposite of the bug this PR removes. The part after the dash is also missing its verb. Reworded below so the point lands as the report being meaningless rather than always false.
```suggestion
 * Only the TOP and BOTTOM overlays contribute to the flag. The corners never did, and the
 * inline-start overlay no longer does: its `innerBorderInlineStart` class shifts no layout since
 * #6673, so its report meant nothing — `false` on most draws, and `true` on the draws where the
 * class toggled, which cost a nested reconciliation draw over the master and every clone for a 1px
 * shift that cannot happen. It is still repositioned here, its result is just not read.
```

handsontable/src/3rdparty/walkontable/src/overlay/regions/inlineStartOverlay.ts
```diff
@@ -410,9 +410,10 @@ export class InlineStartOverlay extends Overlay {
   }
 
   /**
-   * Pre-applies the `innerBorderInlineStart` class before the cell render (single-pass gated
-   * path), so the post-render `resetFixedPosition` toggle is a no-op and the nested re-draw is
-   * skipped. Element mode only — see `TopOverlay#prepareHeaderBorders`.
+   * Pre-applies the `innerBorderInlineStart` class before the cell render (single-pass gated path),
+   * so a `beforeViewRender` listener sees the class in the state this draw ends in. Unlike the top
+   * overlay's copy, this one saves no re-draw: the class shifts no layout, so the post-render toggle
```
budnix, line 415: This fixes the claim here, but the same sentence also sits in `Overlays#prepareHeaderBorders` in `overlay/overlays.ts` near line 403, where it still says the pre-applied class makes the post-render toggle a no-op and skips the nested `wot.draw(true)`. That file is not in the diff, so I could not put a comment on it directly. Worth updating there too, since it is the wrapper that calls this method and so the copy a reader hits first.

tests/e2e/walkontable/inline-start-border-refresh.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+      const draws = await wt.countDrawsAcrossOffsetZero();
+
+      expect(draws.leavingOffsetZero.reconciliation).toBe(0);
+      expect(draws.returningToOffsetZero.reconciliation).toBe(0);
+      expect(draws.leavingOffsetZero.scrollDriven).toBeGreaterThanOrEqual(1);
+      expect(draws.returningToOffsetZero.scrollDriven).toBeGreaterThanOrEqual(1);
+    });
+  });
+
+  test.describe('with a colWidths array (off the single-pass layout path)', () => {
```
budnix, line 90: A `colWidths` array is one way off the single-pass path, but a window-scrolled grid is another, and `prepareHeaderBorders` bails outright on `trimmingContainer === rootWindow`, so that shape toggled the class after the render on every crossing. Curious whether you left it out on purpose, since `rtl-window-scroll.spec.ts` already has the fixture shape for it. Either a third leg or one line here saying why window mode is out of scope would make the coverage story clear.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -533,7 +539,7 @@ function placeFixedOverlays(table: Table, ctx: DrawContext): void {
     ctx.positionChanged = wtOverlays.bottomOverlay.resetFixedPosition() || ctx.positionChanged;
   }
 
-  ctx.positionChanged = wtOverlays.inlineStartOverlay.resetFixedPosition() || ctx.positionChanged;
+  wtOverlays.inlineStartOverlay.resetFixedPosition();
```
budnix, line 542: Dropping this one from the OR leaves `placeFixedOverlays` looking lopsided, five `resetFixedPosition()` calls with two of them feeding the flag, while `_base.ts:231` still declares the method as returning `boolean`. Both corner overlays return a hardcoded `true` (`topInlineStartCornerOverlay.ts:157`, `bottomInlineStartCornerOverlay.ts:140`), so anyone who later ORs one of those in to restore the symmetry makes `positionChanged` true on every draw of a grid with row headers and frozen columns, and every draw then pays a nested `refreshAll()`. A short line here saying the corner results are constants and must never be read would close that off, or the base could stop promising a boolean it no longer gets.

handsontable/src/3rdparty/walkontable/AGENTS.md
```diff
@@ -149,6 +149,31 @@ Consequences worth knowing:
   reads them any more. `innerBorderInlineStart` also only toggles when the grid has row headers and NO
   frozen columns, so it is not a usable "has scrolled" signal in a test — poll the holder's
   `scrollLeft` instead.
+- **`InlineStartOverlay#resetFixedPosition` reports `false` unconditionally**, and `placeFixedOverlays`
```
budnix, line 152: This block lands the new behavior here, but `walkontable/.ai/RENDERING-LIFECYCLE.md` around line 279 still says the `resetFixedPosition()` calls, inline-start included, "OR-together into `positionChanged`". That file is not in the diff so I could not comment on it, and it is half-updated already since it does mention the class drives no geometry since #6673, so a reader now gets two answers. The root CLAUDE.md sends rendering-engine questions to `.ai/` first, so that is the copy most readers will hit.

handsontable/src/3rdparty/walkontable/AGENTS.md
```diff
@@ -149,6 +149,31 @@ Consequences worth knowing:
...
+  no longer ORs it into `ctx.positionChanged` — only the top and bottom overlays feed that flag now.
+  The class shifts no layout, so the flag's one job (reconciling a 1px shift) had nothing to do, yet
+  every scroll crossing horizontal offset 0 ran `refreshAll()`: a nested `wot.draw(true)` over the
```
budnix, line 155: Worth a second look at the skip-changelog reason in the description, which says no released version ever paid this cost. The `18.1.0` tag still has `ctx.positionChanged = wtOverlays.inlineStartOverlay.resetFixedPosition() || ctx.positionChanged` at `drawCycle.ts:554`, so released grids did run this nested draw, and what #13371 changed is that the draw stopped fixing anything. The "nothing user-facing" call may well still be right, but as written the reason does not hold, and this is a real scroll win for anyone on 18.1.0.

tests/e2e/walkontable/inline-start-border-refresh.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    test('costs no reconciliation draw when the grid crosses horizontal offset 0', async () => {
+      // THE regression assertion. Off the single-pass path the class is toggled after the cells
+      // render, so the overlay reported a position change and the draw ran the 1px reconciliation:
+      // one re-entrant `refreshAll` on each crossing, a full nested draw over the master and every
+      // clone, for a shift that cannot happen since #6673. This is the assertion that fails without
+      // the source change.
+      const draws = await wt.countDrawsAcrossOffsetZero();
+
+      expect(draws.leavingOffsetZero.reconciliation).toBe(0);
```
budnix, line 110: Every reconciliation assertion in this file is `toBe(0)`, and I could not find a single positive one anywhere in the repo, since `overlay.spec.js:576` and `:586` are both `not.toHaveBeenCalled()` and `drawCycleHooks.spec.js` only pins the skipped-render case. So a later change that deleted `ctx.positionChanged` and its whole branch would keep this spec, and every other one, green. A row-axis twin here, where a vertical crossing on the measured path must cost exactly one re-entrant `refreshAll`, would pin that the flag is still alive, and the fixture already has the counter for it.

tests/e2e/walkontable/inline-start-border-refresh.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+          expect(await wt.masterBorderClasses())
+            .toEqual({ innerBorderInlineStart: true, innerBorderLeft: true });
+
+          await wt.scrollToOffsetZero();
+
+          expect(await wt.masterBorderClasses())
+            .toEqual({ innerBorderInlineStart: false, innerBorderLeft: false });
+        });
+
+        test('leaves every overlay correctly sized and aligned after a round trip', async () => {
```
budnix, line 180: These metrics only ever come from draws that actually rendered, which leaves one case uncovered. A draw whose `beforeViewRender` listener set `skipRender` and which crosses offset 0 used to take the `positionChanged` branch and get the master `adjustElementsSize()` at `drawCycle.ts:283`, and now takes the plain `else` instead, so the stale-scrollbar note on that very line is what gets narrowed. It is reachable outside tests, since `nestedRows.ts:757` turns `#skipRender` on and clears it inside a `_registerTimeout`, so a horizontal scroll landing in that window is a skipped draw. Likely harmless because a horizontal crossing changes no master size after #6673, but that is exactly what this test proves only for draws that rendered.

tests/fixtures/demo/walkontable/inline-start-border-refresh.html
```diff
@@ -0,0 +1,245 @@
...
+    // Theme and bundle come from the ?theme=/?bundle= query params set by the
+    // Playwright projects, mapped through fixed allowlists to LITERALS so a
+    // crafted param can never be reflected into the document — no XSS. An
+    // ABSENT param keeps the default (main / plain UMD) so the fixture stays
+    // browsable by hand; an unknown value THROWS instead of falling back — a
+    // typo in the project config must be one red leg, never a silently
+    // mislabeled green one (nothing else asserts which file actually loaded).
+    const htParams = new URLSearchParams(location.search);
+
+    window.htTheme = ({ main: 'main', horizon: 'horizon', classic: 'classic' })[htParams.get('theme') ?? 'main'];
```
budnix: These lookups are plain object literals, so a prototype key slips past the guard below. `?theme=constructor` returns the `Object` constructor, which is truthy, so the throw never fires and the stringified function lands in the href. Nothing attacker-controlled reaches the document so the "no XSS" part holds in effect, but "a crafted param can never be reflected" promises more than the code does. The one that would actually bite is `?colWidths=valueOf`, which passes the guard and then matches neither branch, so a leg labelled varied would quietly measure the uniform grid, which is the mislabeled green this comment says it wants to prevent.
```suggestion
    const htThemes = { main: 'main', horizon: 'horizon', classic: 'classic' };
    const htThemeParam = htParams.get('theme') ?? 'main';

    window.htTheme = Object.hasOwn(htThemes, htThemeParam) ? htThemes[htThemeParam] : undefined;
```

tests/fixtures/demo/walkontable/inline-start-border-refresh.html
```diff
@@ -0,0 +1,245 @@
...
+
+    const holder = () => container.querySelector('.ht_master .wtHolder');
+    const masterParent = () => container.querySelector('.ht_master');
+
+    /**
+     * Resolves once `predicate` holds, then after two more frames so the rAF-batched draw has
+     * landed. Throws rather than hanging if the state never arrives - a silent timeout here would
+     * read as "zero reconciliation draws" and pass the very assertion under test.
+     */
+    const waitFor = async (predicate, what) => {
```
budnix, line 142: The doc comment says this resolves after the rAF-batched draw has landed, but the predicate it gets from `scrollToColumn` is already true on the first pass, because `setScrollPosition` writes `scrollEl.scrollLeft` synchronously. So what you get is two frames after the scroll was issued, while the draw arrives later from an async scroll event through `ScrollSync` (`scrollSync.ts:320`). If a loaded runner slips past those two frames, `scrollDriven` comes back 0 and the leaving crossing's draws get counted against the returning one, so the failure reads as a bad count rather than a timing miss. A counter bumped from a real `afterViewRender` in the fixture would make the wait mean what the comment claims.

tests/fixtures/demo/walkontable/inline-start-border-refresh.html
```diff
@@ -0,0 +1,245 @@
...
+        return {
+          leavingOffsetZero: leaving,
+          returningToOffsetZero: {
+            scrollDriven: scrollDriven - leaving.scrollDriven,
+            reconciliation: reconciliation - leaving.reconciliation,
+          },
+        };
+
+      } finally {
+        wtOverlays.refreshAll = original;
```
budnix: This does not put the original method back. `original` is `wtOverlays.refreshAll.bind(wtOverlays)` from line 153, so the instance ends up carrying an own, pre-bound property that shadows `Overlays.prototype.refreshAll` instead of falling through to it. Anything that later spies on the prototype method, or calls `refreshAll` with a different receiver, no longer sees the real one on this instance. Harmless right now because every test re-`goto`s a fresh page, but the `try/finally` reads like a real restore.
```suggestion
        delete wtOverlays.refreshAll;
```

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
demtario: Taken verbatim in `e70a793`. You are right that it read backwards: `false` on most draws with `true` on the toggling ones is the bug, and my wording described the state after the change as though it were the state before it.

handsontable/src/3rdparty/walkontable/src/overlay/regions/inlineStartOverlay.ts
demtario: Fixed in `e70a793`. `Overlays#prepareHeaderBorders` now says the skipped re-render is `innerBorderTop`'s alone, and that pre-applying the inline-start class only keeps a `beforeViewRender` listener from seeing a stale value. Agreed it is the copy a reader hits first, since it is the wrapper that calls the per-overlay method.

tests/e2e/walkontable/inline-start-border-refresh.spec.ts
demtario: Not deliberate, and you were right that it paid. Covered in `e70a793`: the fixture takes `?scroll=element|window` and there is a window-mode block in the spec.

Worth recording how it verified, because my first attempt said the opposite. I reverted the source with `git stash`, but the fix was already committed, so only my uncommitted doc edits came back and the whole file passed. Re-run against `origin/develop`'s source instead, the window leg fails `Expected: 0, Received: 1` exactly as the `colWidths` leg does, so it is a real regression test rather than decoration.

One wrinkle in the fixture: window mode drives the page scroll directly instead of `scrollViewportTo`. Asking that API for column 0 there targets the grid's own document origin, not `scrollX === 0` — the body margin and the heading above the grid put the table a little way in — so the offset the border state derives from never returned to zero and the crossing never happened.

Separately, I had earlier talked myself out of a related concern by answering the wrong question. A window-scrolled grid cannot be *on* the single-pass path, since the gate requires `!isHorizontallyScrollableByWindow()` and that predicate is the same trimming-container test `prepareHeaderBorders` bails on. That is true and irrelevant to your point, which is that window scrolling is a second way *off* the path. My `AGENTS.md` list of ways off it named only the uniform-size causes; it now names both.

tests/fixtures/demo/walkontable/inline-start-border-refresh.html
demtario: Real bug, fixed in `bfc9969`. The maps are read through `Object.hasOwn` now, via a shared `pick(map, raw, name)` used by all four params.

Measured the four keys against both forms, and the old truthiness guard admitted every one of them: `valueOf`, `constructor`, `toString` and `__proto__` all pass `!!map[k]` and all fail `Object.hasOwn`, while a real key still resolves. So `?colWidths=valueOf` would indeed have gone through and matched neither branch, which is the mislabeled green that block exists to prevent.

tests/fixtures/demo/walkontable/inline-start-border-refresh.html
demtario: Correct, fixed in `bfc9969` with `delete wtOverlays.refreshAll`. Assigning the pre-bound function back left an own property shadowing the prototype method rather than falling through to it, so the `try/finally` was not the restore it looked like.

tests/fixtures/demo/walkontable/inline-start-border-refresh.html
demtario: Right, and the wait now requires a real draw. The fixture counts `afterViewRender` and `scrollToColumn` waits on the offset AND the counter advancing, so it no longer returns two frames after the scroll was merely issued.

One case needed handling: the opening scroll to column 0 on a fresh page is already at target, so nothing scrolls and no draw is coming - requiring one there would hang. That path settles and returns without waiting for a render.

Checked that `afterViewRender` actually fires on these crossings before relying on it, since `onDraw` sits inside the full-render branch and would not fire on a fast draw. The band changes on both directions here, so it does.

tests/e2e/walkontable/inline-start-border-refresh.spec.ts
demtario: Good catch, and taken. There is now a positive assertion: a vertical crossing off the single-pass path must cost exactly **one** re-entrant `refreshAll`. The counter is generalised over the axis, so the row-axis case reuses it.

It currently reads 1, so it would fail at 0 if the flag and its branch were deleted rather than narrowed - which was your point. I have also noted in the spec that this expectation becomes 0 when the row axis is brought into line, so it doubles as the boundary marker for that work.

tests/e2e/walkontable/inline-start-border-refresh.spec.ts
demtario: Covered in `bfc9969`. The fixture can now cross offset 0 with the cell render cancelled through a one-shot `beforeViewRender` listener, and the spec asserts the master's sizes are unchanged across it - hider width and height, both scroll ranges, and the row-header width.

Only the size fields: a skipped render rolls the rendered band back, so row offsets within the table legitimately differ. `skipped >= 1` is asserted as a precondition, or the rest would pass on a draw that rendered normally. The wait there is on the offset alone by necessity, since no draw renders and `afterViewRender` never fires.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
demtario: Fixed in `bfc9969` with the comment, at the call site. It names both corner overlays as returning a constant, says never to OR either in to make the pass look symmetrical, and gives the consequence: `positionChanged` would be `true` on every draw of any grid with row headers and frozen columns, and every draw would pay a nested `refreshAll()`.

I left the base class declaring `boolean` rather than narrowing it. The row-axis follow-up removes `ctx.positionChanged` and this whole branch, so reshaping the contract now would be churn against a signature that is about to go.

handsontable/src/3rdparty/walkontable/AGENTS.md
demtario: Fixed in `bfc9969`. Phase G now says only the top and bottom results OR together, that the inline-start overlay reports `false` unconditionally and its result is not read at all, and that the corner overlays return a constant which must never be ORed in. Agreed on the routing - the root guide sends rendering-engine questions to `.ai/` first, so a half-updated copy there is worse than none.

handsontable/src/3rdparty/walkontable/AGENTS.md
demtario: You are right and I have replaced the skip with a `changed` entry (`.changelogs/13393.json`), plus a section in the description explaining the correction.

Verified your reading directly: `git show 18.1.0:.../drawCycle.ts` has the OR at line 554, so released grids do run this nested draw. My `git tag --contains` check tested the wrong thing - it proved the *waste* is unreleased, not the *draw*. Since #13371 and this change ship together, the net effect against 18.1.0 is a real scroll win, which is what the entry now says.

budnix (approved):

PR #13394, opened 2026-09-07
DEV-2784: Keep a merge's trimmed rows when a row move splits it

demtario:
### Context

The PR fixes a merged cell losing the rows a filter or `trimRows` had removed from view, permanently, when `manualRowMove` split that merge into fragments.

A merged cell is described by its **anchor**: the physical rows it owns plus its physical left column. Its visual `row` / `col` / `rowspan` are derived from that anchor whenever the row index mapper's cache updates, so a row removed from view stays in the anchor and the merge comes back whole once the row returns.

A row move goes through `MergedCellsCollection#translateAfterAxisMove`, which replaces every merge object with new ones, splitting a merge into one fragment per contiguous run. Anchors are keyed on object identity, so they have to be carried onto the replacements — and `#transferAnchorsAfterAxisMove` gave up whenever a row move produced more than one fragment, because a row with no visual position cannot be placed in a visual run. Each fragment was then re-anchored from what was visible, the removed rows were dropped, and clearing the filter no longer restored them.

**The rule this PR introduces:** a row removed from view travels within its **run of the row index sequence** — a maximal set of the merge's *own* rows with no other row between them — and each run's removed rows go to that run's first visible row. A run with no visible row cannot be placed at all, and its rows are dropped, which is what happened to every one of them before.

The sequence, rather than the physical index, is what defines a run: `IndexMapper#moveIndexes` removes the moved rows from the full sequence and re-inserts them at a position computed among the rows that are *not* trimmed, so every row it did not move keeps its slot — which makes the sequence exactly the order the rows take once they come back. Physical distance ties in the ordinary case and does not see the move at all.

Two properties of the run rule are worth stating, because the obvious alternative fails on both:

- **A removed row is never paired across a row the merge does not own.** The span is one continuous block from its top-left, so an anchor pairing rows across a foreign row cannot be drawn: once trimming is lifted the block covers that foreign row and stops short of the merge's own row beyond it. A plain nearest-in-sequence rule shipped exactly that defect during review, caught by a reviewer, and the spec named below pins it.
- **Distance inside a run is deliberately not measured.** A run holds no foreign row, so its visible rows are visually adjacent and always land in the same fragment. A nearest-carrier rule would compute a choice nothing can observe, and its tie-break would be a branch no test could pin.

Three details around it are load-bearing, and each is documented in the plugin's `AGENTS.md`:

- **The attribution is computed in `afterRowMove`, not `beforeRowMove`**, because it reads the sequence the move produced. It is skipped entirely unless a merge exists and a row is trimmed, since building the sequence position map is real allocation per drop on a large grid.
- **Whether the merge was split is read from a run count, not from how many fragments came back.** A merge broken into a real run plus a single cell returns one fragment, because the collection drops the single cell; carrying the whole anchor onto that survivor would hand it the dropped fragment's rows too. The count goes through the same `detectContiguousRuns` helper the split itself uses, so the two cannot disagree.
- **A merge whose rows a sort has scattered is excluded from both the split and the retention.** Its span is one block reaching over rows it does not own, so a fragment can be cut from foreign rows. Such a merge keeps exactly the behavior it had before this change.

`translateAfterAxisMove` also gained an optional `retainedIndexes` parameter, so a fragment that draws a single cell **only because its other rows are removed from view** survives the singleton drop; it is a merge again as soon as they return. It is keyed per merge rather than as one flat set, because two merges in different columns can cover the same rows, and a shared set would keep the genuinely-single fragment of the merge that carries nothing, leaving a phantom `1x1` entry in the lookup matrix. The retention is not conditional on the merge having been split, which also fixes a second defect found along the way: a merge trimmed down to one visible cell was deleted by **any** row move, including one that never touched it.

No public API or default setting changes. The new parameter is optional with a default, and the two other call sites (column move, column freeze) are unaffected — a column reorder never changes which rows a merge covers.

### How has this been tested?

Seven E2E cases added to the existing row-trimming `describe` in `mergeCells.spec.js`, and two unit cases to `cellsCollection.unit.ts`. They fall into two groups, which is worth separating:

**Five pin behavior this PR changes, and each was confirmed to fail against the unpatched source:**

- a split that leaves a removed row above the break, and the mirror case below it (`Expected $[0].rowspan = 1 to equal 2.`)
- a single-column merge split so the surviving cell still owns a removed row, and a merge trimmed to one visible cell hit by an unrelated row move (`Expected $.length = 0 to equal 1.`)
- `should not send a trimmed row across a row the merge does not own when a move splits it` — the reachability case, which failed against the first version of this branch with `Expected $[0].rowspan = 3 to equal 4.`

**Two pin behavior this PR deliberately leaves alone,** so they pass against `develop` as well and exist to stop a future change from quietly widening the new path into a shape it cannot reason about: the two sort-scattered cases, one per `colspan`. Each fails if its guard is removed — dropping `#describesOwnRows` from the split turns the first red, and dropping it from the retention turns the second red.

The two unit cases cover the retention, including that one merge's retained index cannot keep another merge's single-cell fragment.

```bash
# the new cases
npm run test:e2e --prefix handsontable -- --testPathPattern='mergeCells/__tests__/mergeCells.spec'
npm run test:unit --prefix handsontable -- --testPathPattern='cellsCollection'

# full local gates, all green
npm run test:e2e --prefix handsontable     # 9918 specs, 0 failures
npm run test:unit --prefix handsontable    # 4435 passed
npm run eslint --prefix handsontable       # 0 errors
```

### Types of changes

### Related issue(s):

1. DEV-2784

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2784

---

> [!NOTE]
> **Medium Risk**
> Touches core merge anchor transfer and row-move hooks with subtle edge cases (sort-scattered merges, per-merge retention), but behavior is gated on active row trimming and heavily spec’d; no public API change.
> 
> **Overview**
> Fixes **mergeCells** losing physical rows that **filters** / **`trimRows`** had hidden when **`manualRowMove`** split the merge or when any row move hit a merge that only showed one visible row.
> 
> Row moves now **plan trimmed-row attribution** from the post-move row index sequence (runs of the merge’s own rows, first visible carrier per run, no pairing across foreign rows) and **re-anchor split fragments** with those rows. **`translateAfterAxisMove`** accepts optional per-merge **`retainedIndexes`** so a `1×1` fragment that still owns trimmed rows is not dropped as a singleton—including unrelated row moves on a merge trimmed to one visible cell. **Sort-scattered** merges stay on the old path via **`#describesOwnRows`**.
> 
> Docs and **`AGENTS.md`** describe the new rules and clarify that **undo/redo** restores order, not merges. **E2E** and **unit** tests cover the new behavior and retention edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 727d13e3444bbe68f3600d699496a773978a187c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2784 Merge cells: a manualRowMove split drops the trimmed rows of the merge](https://app.clickup.com/t/9015210959/DEV-2784)

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -1165,6 +1191,231 @@ export class MergeCells extends BasePlugin {
...
+   * sequence, which asks for carriers that are not visually adjacent to begin with — the scattered
+   * shape {@link MergeCells#describesOwnRows} keeps out of both the split and the retention. Keep the
+   * rule anyway: that reasoning holds for the moves that exist today, not for every one that could be
+   * added.
+   *
+   * @param {number[]} physicalRows The anchor's physical rows.
+   * @param {number[]} rowSequencePositions Physical row -> its slot in the row index sequence.
+   * @returns {Map<number, number[]>} Visible physical row -> the trimmed physical rows it carries.
+   */
+  #attributeTrimmedRows(physicalRows: number[], rowSequencePositions: number[]): Map<number, number[]> {
```
Haxikowy, line 1269: Nearest-in-sequence has no contiguity check, and `#describesOwnRows` cannot catch the gap because it reads the pre-move drawn block, so a move that lands a foreign row between two carriers sends a trimmed row across it and builds a non-contiguous anchor. Reproduced on this branch: 12x3 grid, `mergeCells: [{ row: 2, col: 0, rowspan: 5, colspan: 2 }]`, `trimRows([3,4,5])`, `moveRow(6, 3)`, `untrimAll()` yields `[{row:2,rowspan:3,colspan:2},{row:5,rowspan:2,colspan:2}]`, where the second fragment covers physical row 9 (`A10` is visibly inside the merge) and leaves the merge's own physical row 6 outside it.

Walking outward through the sequence and refusing a carrier not reachable without crossing an unowned row would fix it; falling back to the old lossy path is better than emitting a non-contiguous anchor. Three trimmed rows is the minimum, at two the tie-break happens to save it.

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -1165,6 +1191,231 @@ export class MergeCells extends BasePlugin {
...
+
+    if (visibleRows.length === 0) {
+      return trimmedRowsByCarrier;
+    }
+
+    physicalRows
+      .filter(physicalRow => this.hot.toVisualRow(physicalRow) === null)
+      .forEach((trimmedRow) => {
+        const trimmedPosition = rowSequencePositions[trimmedRow];
+        const carrierRow = visibleRows.reduce((nearest, candidate) => {
```
Haxikowy: The `visibleRows.reduce` runs once per trimmed row, so attribution is O(rowspan^2) for a single merge. Measured in plain Node against transcribed method bodies: rowspan 500 costs 0.112 ms, 5000 costs 10.9 ms, 50000 costs 1006 ms. Sorting the merge's rows by sequence position once and sweeping with two pointers gives O(n log n).

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -1165,6 +1191,231 @@ export class MergeCells extends BasePlugin {
...
+   * Decides which visible row of a merge each of its trimmed rows travels with when a row move splits
+   * the merge: the visible owned row nearest to it in the row index sequence, the row above winning a
+   * tie.
+   *
+   * Physical distance cannot answer this. A trimmed row usually sits one physical index away from two
+   * of its neighbors, so the distance ties, and it says nothing about the move — `moveIndexes` leaves
+   * every row it did not move where it was in the sequence and re-inserts the moved ones among the
+   * rows that are not trimmed, so the sequence is exactly the order the rows come back in.
+   *
+   * The tie-break is there to make the result independent of the order the rows are listed in, and no
```
Haxikowy: The docblock says a tie needs carriers that are not visually adjacent, and that no spec depends on which side it picks. Both are wrong: with two trimmed rows between adjacent carriers the distance ties at 2 and `isCandidateAbove` is the only thing preventing the corruption in my other comment, verified on the branch. Nothing pins it, so a future reader can delete or flip the tie-break and every spec stays green.

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -2300,8 +2577,10 @@ export class MergeCells extends BasePlugin {
       return;
     }
 
+    const { context, carriers } = this.#planRowMoveTranslation(snapshot, this.#buildRowSequencePositions());
```
Haxikowy: `#buildRowSequencePositions()` is evaluated as an argument, so it runs on every drop even when the collection is empty or nothing is trimmed, and in both cases the array is never read. On a 1M-row grid that is 8.9 ms of allocation per drop (measured, plain Node). Return early on `mergedCells.length === 0` and gate the rest on `#isRowTrimmingActive()`.

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -1165,6 +1191,231 @@ export class MergeCells extends BasePlugin {
...
+      if (!anchor || !physicalRows) {
+        return;
+      }
+
+      const trimmedRowsByCarrier = this.#attributeTrimmedRows(anchor.physicalRows, rowSequencePositions);
+      // The snapshot is the merge's drawn block, which for a merge whose rows a sort scattered reaches
+      // over rows it does not own. Its fragments would then be anchored onto those foreign rows, so
+      // such a merge is reported as unsplit and left to the guarded single-fragment path, exactly as
+      // it was before the split learned to distribute trimmed rows.
+      const describesOwnRows = this.#describesOwnRows(physicalRows, trimmedRowsByCarrier);
```
Haxikowy, line 1370: When `describesOwnRows` is false and the merge carries trimmed rows, those rows are lost permanently and nothing tells the user: the row-move path calls `collection.add(info, true)` directly, so no hook fires and no warning prints. The guide's advice to un-sort first cannot reach someone mid-drag, and the undo path will not bring the rows back either.

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -1165,6 +1191,231 @@ export class MergeCells extends BasePlugin {
...
+    fragments: MergedCellCoords[],
+    anchor: MergeAnchor,
+    plan: RowMoveTranslation,
+    rowSequencePositions: number[]
+  ) {
+    fragments.forEach((fragment) => {
+      const physicalRows: number[] = [];
+
+      rangeEach(fragment.rowspan - 1, (offset) => {
+        const physicalRow = this.hot.toPhysicalRow(fragment.row + offset);
```
Haxikowy, line 1435: `toPhysicalRow` can return `null` here and the value is pushed unguarded, after which `rowSequencePositions[null]` makes the comparator return `NaN` and the sort order becomes engine-dependent, which `#reanchorMergesToVisibleRows` then reads the head of as the top-left. `#captureAnchorOf:1016` filters nulls at the equivalent site.

Two neighbours have the same shape: `rowSequencePositions[trimmedRow]` at 1244 is unguarded (an undefined slot silently dumps every trimmed row on `visibleRows[0]`), and `#describesOwnRows` returns true vacuously on an empty `presentRows`.

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -1165,6 +1191,231 @@ export class MergeCells extends BasePlugin {
...
+   *
+   * The distinction matters because the fragments of a split are cut from the drawn block. Only when
+   * the block is the merge's own rows is a fragment's content its own by construction, which is what
+   * lets {@link MergeCells#reanchorFragmentsAfterSplit} anchor a fragment from the rows it draws.
+   *
+   * @param {number[]} drawnRows Physical rows the merge drew before the move.
+   * @param {Map<number, number[]>} trimmedRowsByCarrier Keyed by the merge's visible own rows.
+   * @returns {boolean}
+   */
+  #describesOwnRows(drawnRows: number[], trimmedRowsByCarrier: Map<number, number[]>): boolean {
```
Haxikowy, line 1313: `physicalRow >= 0` is dead here, and at 1299 and `cellsCollection.ts:797`: `getVisualFromPhysicalIndex` maps both `undefined` and the `-1` sentinel to `null` (`indexMapper.ts:421`), and `toPhysicalRow` does the same. Line 1235 spells the same predicate as `!== null` alone, so the file now carries three spellings and the reader has to check which one is right.

handsontable/src/plugins/mergeCells/cellsCollection.ts
```diff
@@ -799,7 +811,8 @@ class MergedCellsCollection {
           [otherSpanProp]: merge[otherSpanProp],
         };
 
-        if (replacement.colspan === 1 && replacement.rowspan === 1) {
+        if (replacement.colspan === 1 && replacement.rowspan === 1
+          && !retained?.has(toPhysical(run.start))) {
```
Haxikowy, line 815: Retention is keyed on carriers, so when a split leaves two single-cell fragments and only one holds a trimmed row, that one survives as a merge and its sibling is destroyed. Which fragment survives is decided by the attribution rather than by anything the user did, and nothing documents or pins the rule.

docs/content/guides/cell-features/merge-cells/merge-cells.md
```diff
@@ -448,7 +448,7 @@ Setting `virtualized` to `true` also removes the range expansion, but it is a pe
...
 
 - **Auto-split**: if the move bisects a merge so the underlying cells are no longer contiguous in the new visual order, the merge is split into separate merges, one per contiguous run. The cross-axis span (`rowspan` for column moves, `colspan` for row moves) is preserved on every fragment.
-- **Silent drop of single-cell fragments**: any resulting fragment that ends up as a single cell (`rowspan === 1 && colspan === 1`) is removed, because a single cell is no longer a merge. The [`afterMergeCells`](@/api/hooks.md#aftermergecells) hook is not fired for the dropped fragment.
+- **Silent drop of single-cell fragments**: any resulting fragment that ends up as a single cell (`rowspan === 1 && colspan === 1`) is removed, because a single cell is no longer a merge. The [`afterMergeCells`](@/api/hooks.md#aftermergecells) hook is not fired for the dropped fragment. A fragment that shows a single cell only because the rest of its rows are removed from view is kept, and spans them again when they come back.
```
Haxikowy: This promises unconditionally that a fragment showing one cell is kept and spans its rows again, but the retention is gated on `#describesOwnRows` (`mergeCells.ts:1352`). For a sort-scattered single-column merge the fragment is still dropped, which the spec at `mergeCells.spec.js:3808` asserts. The sentence predates commit `d94c8ad4` and was not updated with it.

docs/content/guides/cell-features/merge-cells/merge-cells.md
```diff
@@ -459,6 +459,8 @@ Some features remove rows from the grid entirely: [`filters`](@/api/options.md#f
...
 - When none of its rows is shown, the merged cell is not displayed.
 - When the rows come back, the merged cell spans them again. Nothing about the merge is lost while its rows are away. A merged cell you create while rows are already removed from view covers only the rows you could see, and does not grow when the rest come back.
+- When a row move splits a merged cell while some of its rows are removed from view, each removed row stays with the fragment that holds the row nearest to it in the grid's row order, which is the order the rows take when they come back. When two fragments are equally near, the row joins the one above.
+- A merged cell whose rows a sort has separated is the exception. It still spans one block, so that block reaches over rows the merged cell does not cover, and a row move that breaks the block cannot tell which fragment owns which row. Each fragment then covers only the rows it shows, and the rows removed from view are not restored. Sort the column back, or clear the filter, before you move the rows.
```
Haxikowy: "Each fragment then covers only the rows it shows" understates it for a single-column merge: `mergeCells.spec.js:3808` asserts `merges()` is `[]` both after the move and after `untrimAll()`, so the merged cell disappears outright. Worth saying that a fragment left as a single cell is dropped in this case, since a merge vanishing on a drag is the part a user will notice.

docs/content/guides/cell-features/merge-cells/merge-cells.md
```diff
@@ -448,7 +448,7 @@ Setting `virtualized` to `true` also removes the range expansion, but it is a pe
...
 
 - **Auto-split**: if the move bisects a merge so the underlying cells are no longer contiguous in the new visual order, the merge is split into separate merges, one per contiguous run. The cross-axis span (`rowspan` for column moves, `colspan` for row moves) is preserved on every fragment.
-- **Silent drop of single-cell fragments**: any resulting fragment that ends up as a single cell (`rowspan === 1 && colspan === 1`) is removed, because a single cell is no longer a merge. The [`afterMergeCells`](@/api/hooks.md#aftermergecells) hook is not fired for the dropped fragment.
+- **Silent drop of single-cell fragments**: any resulting fragment that ends up as a single cell (`rowspan === 1 && colspan === 1`) is removed, because a single cell is no longer a merge. The [`afterMergeCells`](@/api/hooks.md#aftermergecells) hook is not fired for the dropped fragment. A fragment that shows a single cell only because the rest of its rows are removed from view is kept, and spans them again when they come back.
 
 [`undo`](@/api/options.md#undo) and [`redo`](@/api/options.md#redo) restore the pre-move state, including any merges that were split or dropped by the reorder.
```
Haxikowy: `RowMoveAction` records only `{rows, finalIndex}` and replays the inverse move (`undoRedo/actions/rowMove.ts:60`), so no merge geometry is restored and a split merge never comes back whole. The claim was already wrong on develop, but this PR changes the outcome for colspan-1 merges and edits the two bullets directly above, so it now reads as reviewed and kept.

Haxikowy (commented):
`#planRowMoveTranslation` reads `getIndexesSequence()`, which nestedRows never updates: `rowMoveController.ts:154` fires `afterRowMove` by hand after reordering the source data, and its `beforeRowMove` returns `false` so `moveIndexes` never runs. I traced the consequence and nothing splits under that path, so the stale attribution is never read, but the plan is computed against an order the move did not produce and no spec covers it.

Two things in the description no longer match the branch. It says five E2E cases were added and head has six, and "every one of them was confirmed to fail against the unpatched source" cannot hold for `should not retain a single-cell fragment of a merge whose rows a sort scattered`, which passes against develop because the fragment is dropped there anyway. That sixth spec pins the guard added in `d94c8ad4`, which is worth saying plainly since the sentence is what a reviewer leans on.

Line 15 of the description has "behaviour"; `.ai/DOC-STANDARDS.md` rule 4 covers PR descriptions.

handsontable/src/plugins/mergeCells/AGENTS.md
```diff
@@ -85,6 +85,67 @@ re-deriving them from the merges. They have to: the index mapper emits its cache
 `afterCreateRow`/`afterRemoveRow`, so by the time those hooks run a re-anchor has already gone round once
 against a grid whose row count changed while the merges had not been shifted yet.
 
+## A row move that splits a merge distributes its trimmed rows
+
+A row move can leave a merge's rows in more than one contiguous visual run, and `translateAfterAxisMove`
+then replaces the merge with one fragment per run. A trimmed row has no visual index, so it belongs to no
+run, and the rule that places it is: **the fragment holding the visible owned row nearest to it in the row
+index sequence, the row above winning a tie.**
```
budnix: This writes the nearest-in-sequence rule down as the plugin's contract, and there is already a comment on `mergeCells.ts:1233` with a case where that rule picks a carrier on the far side of a row the merge does not own, leaving a fragment whose rows are no longer next to each other in the sequence. If the attribution rule changes, this section and the tie-break note below it have to change with it. Worth doing in the same commit, since this is the file the next person reads before touching `#attributeTrimmedRows`.

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -1165,6 +1191,225 @@ export class MergeCells extends BasePlugin {
...
+      plans.set(merge, { trimmedRowsByCarrier, isSplit });
+
+      const carrierRows = this.#carrierRowsOf(trimmedRowsByCarrier);
+
+      // Not gated on `isSplit`: a merge trimmed down to one visible cell draws a single cell without
+      // any help from the move, and the collection would drop it as a singleton on a move that does
+      // not even touch it.
+      if (carrierRows.size > 0) {
+        carriers.set(merge, carrierRows);
+      }
```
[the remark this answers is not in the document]
demtario: Fixed in `d94c8ad4`. `carriers` is now gated on `#describesOwnRows`, the same guard the split uses, so a scattered merge's single cell is left to be dropped instead of retained.

Pinned by `should not retain a single-cell fragment of a merge whose rows a sort scattered` — removing the gate turns it red with `Expected $.length = 1 to equal 0.` both after the move and after `untrimAll()`.

handsontable/src/plugins/mergeCells/mergeCells.ts
demtario: Confirmed and fixed in `d36ef5a8`. Your fixture reproduced exactly: `[{row:2,rowspan:3,colspan:2},{row:5,rowspan:2,colspan:2}]`, the second fragment over physical 9 with the merge's own physical 6 outside it.

`#groupRowsBySequenceRuns` now cuts the merge's rows into runs — maximal sets of its *own* rows with nothing else between them in the sequence — and a run's trimmed rows go to that run's first visible row, so a row is never paired across a row the merge does not own. A run with no visible row cannot be placed and is dropped, which is the fallback you suggested.

Your case is now a spec: `should not send a trimmed row across a row the merge does not own when a move splits it`. It fails against the previous version of this branch with `Expected $[0].rowspan = 3 to equal 4.`

handsontable/src/plugins/mergeCells/mergeCells.ts
demtario: Fixed by the same rewrite in `d36ef5a8`. The per-trimmed-row `reduce` is gone: the merge's rows are sorted by sequence position once, cut into runs, and each run is walked once, so the pass is O(n log n) dominated by that sort.

handsontable/src/plugins/mergeCells/mergeCells.ts
demtario: You were right that the note was wrong, and chasing it showed the branch should not exist at all.

With run grouping, a run contains no foreign row, so its visible rows are visually adjacent and always land in the *same* fragment. Which carrier inside a run receives a trimmed row is therefore unobservable, and the tie-break was a branch no test could ever pin — which is what you were pointing at from the other side.

Removed the distance arithmetic entirely in `d36ef5a8`. The docblock now states that distance inside a run is deliberately not measured, and why.

handsontable/src/plugins/mergeCells/mergeCells.ts
demtario: Fixed in `d36ef5a8`. The plan is skipped unless `mergedCells.length > 0` and `#isRowTrimmingActive()`, so no position map is built on a drop that has nothing to attribute.

Skipping is behavior-preserving rather than a shortcut: with no row trimmed there are no trimmed rows to distribute, and the path it falls back to gives the same answer — a fragment draws exactly the rows it owns.

handsontable/src/plugins/mergeCells/mergeCells.ts
demtario: Left as is, and I would rather flag that than decide it quietly.

It is pre-existing: on `develop` the same shape loses those rows too, because every fragment is re-anchored from what is visible. What this PR changes is how *often* it is reachable, not that it is silent.

I did not add a `warn()` because it would fire on an ordinary drag in a legitimate configuration (sorted grid, active filter, merged cells), which felt like a call to make deliberately rather than fold into a fix. For now it is documented on both sides: the guide says to sort back or clear the filter before moving the rows, and AGENTS.md records why the shape is excluded.

Happy to add a warning or a hook in this PR if you would prefer the user-visible signal now.

handsontable/src/plugins/mergeCells/mergeCells.ts
demtario: All three fixed in `d36ef5a8`.

- A `null` from `toPhysicalRow` is skipped rather than pushed, so it can no longer reach the comparator and leave the sort order to the engine.
- `#groupRowsBySequenceRuns` filters out rows with no sequence position, which covers the unguarded `rowSequencePositions[trimmedRow]` read you pointed at — that expression no longer exists, since the rewrite dropped the distance arithmetic.
- `#describesOwnRows` now requires a non-empty drawn block instead of passing vacuously on an empty one.

handsontable/src/plugins/mergeCells/mergeCells.ts
demtario: Fixed my two sites in `d36ef5a8` — both read `!== null` only now.

I left `cellsCollection.ts:797` alone because it predates this branch, so the count is down from three spellings to two rather than one. Say the word and I will unify it here.

handsontable/src/plugins/mergeCells/cellsCollection.ts
demtario: Documented in `d36ef5a8`. AGENTS.md now states it plainly rather than leaving it to be inferred: the sibling is a genuine single cell and the survivor is a merge whose other rows are away, so which one survives is decided by where the trimmed rows are reachable from, never by anything about the drag.

The pair is pinned by `should keep a single-column merge alive through a split when the surviving cell owns a trimmed row`, which asserts both halves — the survivor at physical 2 and the sibling at physical 4 gone.

docs/content/guides/cell-features/merge-cells/merge-cells.md
demtario: Fixed in `d36ef5a8`. The sentence now ends with the exception and points at the next section, where the scattered case is described.

docs/content/guides/cell-features/merge-cells/merge-cells.md
demtario: Fixed in `d36ef5a8`. The bullet now says a single-column merged cell disappears altogether in this case, because every fragment it leaves behind is a single cell — agreed that a merge vanishing on a drag is the part a user actually notices.

docs/content/guides/cell-features/merge-cells/merge-cells.md
demtario: Verified and fixed in `d36ef5a8`. `RowMoveAction` stores only `{rows, finalIndex}` and its `undo` replays inverse `moveRow` calls, so no merge geometry is recorded or restored; there is also no spec covering merges across a row-move undo.

The guide now says undo and redo restore the row and column order but not merged cells: a merged cell the reorder split stays split, one it dropped stays dropped, and you have to merge again to get the original back.

handsontable/src/plugins/mergeCells/AGENTS.md
demtario: Done in the same commit, `d36ef5a8`.

The attribution rule changed in that commit — the case in the other thread was real, so nearest-in-sequence is gone — and this section changed with it: the run rule, why a trimmed row is never paired across a row the merge does not own, and why distance inside a run is deliberately not measured. The tie-break note below it is gone too, because the branch is gone.

demtario:
Thanks — the reachability case in the line comments was real, and it is fixed in `d36ef5a8` along with the rest. Answering the three points from the review body here, since they have no line to hang on.

**`nestedRows`.** Agreed, and it stays narrowed rather than closed. The plan is now skipped unless a merge exists and a row is trimmed, but that does not cover your case: `rowMoveController` fires `afterRowMove` by hand with `beforeRowMove` returning `false`, so `moveIndexes` never runs and the sequence a plan would read there is the pre-move one. I recorded that in AGENTS.md, together with your finding that nothing splits on that path today so the stale plan is never read, and an instruction not to start depending on it without a spec for the path. I did not add that spec.

**The description.** Corrected. It said five E2E cases and head has seven, and the blanket falsification claim was wrong, so the two groups are now separated: five pin behavior this PR changes and were each confirmed to fail against the unpatched source (with the failure messages), and two pin behavior it deliberately leaves alone, pass against `develop`, and exist so that a future change cannot quietly widen the new path into a shape it cannot reason about — each of those fails if its own guard is removed. The description also now states the run rule rather than the nearest-in-sequence one it was describing.

**"behaviour".** Fixed.

docs/content/guides/cell-features/merge-cells/merge-cells.md
```diff
@@ -448,9 +448,9 @@ Setting `virtualized` to `true` also removes the range expansion, but it is a pe
...
 
 - **Auto-split**: if the move bisects a merge so the underlying cells are no longer contiguous in the new visual order, the merge is split into separate merges, one per contiguous run. The cross-axis span (`rowspan` for column moves, `colspan` for row moves) is preserved on every fragment.
-- **Silent drop of single-cell fragments**: any resulting fragment that ends up as a single cell (`rowspan === 1 && colspan === 1`) is removed, because a single cell is no longer a merge. The [`afterMergeCells`](@/api/hooks.md#aftermergecells) hook is not fired for the dropped fragment.
+- **Silent drop of single-cell fragments**: any resulting fragment that ends up as a single cell (`rowspan === 1 && colspan === 1`) is removed, because a single cell is no longer a merge. The [`afterMergeCells`](@/api/hooks.md#aftermergecells) hook is not fired for the dropped fragment. A fragment that shows a single cell only because the rest of its rows are removed from view is kept, and spans them again when they come back, unless a sort has separated the merged cell's rows (see the next section).
```
budnix: The retention this sentence promises is wired to the row move only, but the sentence sits under **"Behavior during row/column reorder and column freeze"**, which also covers `manualColumnMove` and `manualColumnFreeze`.

At d36ef5a8d3:

- `mergeCells.ts:2628` (row move) passes `plan?.carriers`.
- `mergeCells.ts:2563` (column move) and `mergeCells.ts:2667` (column freeze/unfreeze) call `translateAfterAxisMove('column', snapshot)` with no retention map, so `retained` is `undefined` at the guard in `cellsCollection.ts:814`.

The chain that makes that bite:

1. Trimming rows inside a merge shrinks the merge object's own `rowspan` to the visible count — `mergeCells.ts:1668` passes `visualRowspan` into `relocate()`.
2. A column translation builds each fragment's `rowspan` from that shrunk value (`otherSpanProp`).
3. So a merge with exactly one visible row reaches the guard with `rowspan === 1`. When the fragment's `colspan` is 1 as well, it is dropped — and `translateAfterAxisMove` has already cleared `mergedCells`, so nothing brings it back when the rows return.

Two shapes hit it:

- A single-column merge (`colspan: 1`) trimmed to one visible row. Any column move or column freeze deletes it, with no split involved.
- A wide merge trimmed to one visible row that a column move splits. Every `colspan: 1` fragment is dropped.

A merge with *no* visible row is unaffected: the re-anchor returns before `relocate()`, so its `rowspan` stays above 1.

Worth noting that the singleton drop itself is not new here — this PR adds the escape hatch and wires it to one of the three paths. So the narrowest fix is probably this sentence: scope it to row moves, the way the sort caveat next to it is scoped. Extending the retention to the column axis looks cheap if you would rather close the gap instead, since every fragment there shares the same rows and there is no nearest-row attribution to work out.

I traced this in the code and have not run a spec for it, so please treat the two repro shapes as unverified.

demtario: Good catch, and your trace holds — I ran all three shapes rather than leaving them unverified.

**1. An unrelated column move deletes the merge.** `colspan: 1` merge over physical rows 2 and 3, `trimRows([3])`, then `moveColumn(4, 1)` — a column the merge does not cover. The collection is empty immediately after the move and still empty after `untrimAll()`.

**2. A column freeze deletes it too.** Same shape at `col: 2`, `freezeColumn(2)` empties the collection.

**3. The split case is a partial loss.** A `colspan: 3` merge trimmed to one visible row, bisected by a column move, comes out as one `colspan: 2` fragment and loses the `colspan: 1` one — the wider fragment clears the guard, so it is not the total loss the other two are.

Your reading of why is right: the trimming re-anchor shrinks the merge's own `rowspan` to the visible count, a column translation copies that onto every fragment, and `rowspan === 1 && colspan === 1` then reaches the guard with no retention map behind it.

I took the narrow half of your suggestion here and filed the rest as its own task. The guide bullet is now scoped, and it says what actually happens rather than only what does not: retention is a row-move behavior, and a column move or freeze drops such a merged cell **even when the column you move is not one of its own** — which is the part a user would never guess. `AGENTS.md` carries the same note next to the row-axis hatch, so the next person to touch this sees the asymmetry.

The follow-up task has all three reproductions, the chain through the shrunk `rowspan`, and your point that the column axis needs no attribution work since every fragment there covers the same rows. It is pre-existing on `develop`, on a different axis from this task, and wants its own specs for both paths — so it seemed better as its own review than as a fourth round on this one. Say the word if you would rather have it in here.

Haxikowy (approved):

budnix (approved):

PR #13395, opened 2026-09-07
DEV-2618: Finish the className normalization family

budnix:
### Context

The PR finishes the `className` normalization family that [DEV-2604](https://app.clickup.com/t/9015210959/DEV-2604) / #13235 (GitHub issue #7427) started. That PR fixed `hiddenRows` and `hiddenColumns` and deliberately left three items out of scope. This is those three.

`className` is publicly typed `string | string[]` (`handsontable/src/settings.ts`) and `addClass()` accepts both. #13235 added the shared `normalizeClassNames()` helper in `helpers/dom/element` and moved two call sites onto it. Two hand-rolled copies remained, one of them carrying a real bug.

**1. `numericRenderer` mutated the user's array (the only user-visible defect here).** Line 59 did `classArr = cellProperties.className as string[]`, taking the array **by reference**, and then pushed `htRight` and `htNumeric` into it. A grid-level or column-level `className: ['a', 'b']` is **one array instance shared by every cell** through the cell meta prototype chain (`dataMap/metaManager/metaLayers/cellMeta.ts`), so those two classes landed in the user's own array — and every other cell reading `className` through that prototype picked them up. A text column sharing the array rendered right-aligned and numeric-styled.

Confirmed in Chrome before the fix. The text column's cell rendered `class="htNumeric htRight mark shared"`, and the source array came back as `['shared', 'mark', 'htRight', 'htNumeric']`.

No hiding plugin is involved in the fixture, and that is deliberate: until this PR, `hiddenRows`/`hiddenColumns` rewrote every array `className` they read into a string, which **masked** this leak — so the defect only reproduced without one.

**2. `search` hand-rolled the same normalize, but had no bug.** It built a fresh array with `push(...className)`, so it never aliased. That half is a call-site migration. Its `cellProperties` is also retyped from `Record<string, unknown>` to `CellProperties` — the type `beforeRenderer` already declares in `core/settings.ts:653`, and the exact anti-pattern `handsontable/AGENTS.md` names, since `Record<string, unknown>` erases every property to `unknown` and is what let the original `as string` casts compile.

**3. `contextMenu`'s alignment needed no change — it was already fixed.** The task asked to harden `applyAlignClassName` (`plugins/contextMenu/utils.ts`), which cast `cellMeta.className as string` and passed it to helpers calling `.replace()`. [DEV-2654](https://app.clickup.com/t/9015210959/DEV-2654) / #13266 (GitHub issue #7122) had already rewritten `prepareAlignClass` onto `normalizeClassNames()` and widened the cast to `string | string[]`. It is also already tested, with no hiding plugin in the fixture: `contextMenu/__tests__/utils.unit.js:77` and `:110` pass arrays into `prepare{Horizontal,Vertical}AlignClass`, and `__tests__/predefinedItems/alignment.spec.js:498` right-clicks and aligns a cell whose `className` is an array. Verified rather than assumed; no code in this PR.

**4. The `hiddenRows`/`hiddenColumns` removal-branch write, which #13235 declined.** It was declined **because** it depended on item 3, and item 3 has since landed.

The removal branch ran its full normalize, join, compare and write for every cell that has *any* `className` — on a grid with a grid-level value, that is every cell in it, almost none of them anywhere near a hidden row or column. Finding the marker now gates the join, the compare and the write. (The `normalizeClassNames()` call itself still runs first, because the token test needs the array; a raw-string `includes()` pre-filter would skip that allocation too and is recorded in the plugin's `AGENTS.md` as the next step.)

The correctness half matters more than the speed. Rewriting a cell you have no marker on puts an own string `className` on it that shadows the column-level and grid-level cascade, and for a value set through `setCellMeta` (recorded in `_userDefinedMetaProps`) that string **permanently replaces the user's array**, because `getUserDefinedMetas()` re-reads `className` at `updateSettings` time. Compare-before-assign cannot prevent either, because an array is always `!==` its joined form.

Two things made it safe to take now. A sweep of `handsontable/src/**` found **no** consumer that string-operates a cell-meta `className` — the `.split is not a function` crash class from #7427 is gone from `src/` entirely. And more decisively: a grid with **no** hiding plugin already exposes arrays to every one of those consumers, so leaving arrays in place only makes hiding-plugin grids consistent with the default.

One caveat on that sweep, which was scoped to `src/`: `predefinedItems/alignment.ts` still casts `className as string` at seven call sites, and the value reaches the public `beforeCellAlignment` payload. A **user** handler typed against `string` and calling `.split(' ')` would throw the #7427 shape. That was already reachable on a grid without a hiding plugin; this PR makes it reachable on hiding-plugin grids too. Nothing in `src/` breaks, so it is tracked in DEV-2803 rather than fixed here.

Measured on an isolated re-implementation of the two hook bodies (3M calls, node 22, **not** in-situ, no GC pressure captured): ~106 ns/call before #13235, ~79–84 ns/call as #13235 merged, ~38 ns/call with the write gated.

#### One existing assertion inverts, deliberately

`tests/e2e/hiding-array-classname.spec.ts` had a test named *"normalizes the array to a string on the cells it touches"*, asserting `hasOwnClassName(GRID_LEVEL, 2, 0) === true`. It is now `false`, and the test is renamed to say so: the grid-level array keeps cascading instead of being replaced by an own string copy. This also makes the array path symmetric with the string path's existing *"leaves the cascade intact on rows it does not mark"* test, which asserted `false` all along.

### How has this been tested?

Both new tests were verified **red before green**. The unit test reported the source array as `['foo', 'bar', 'htRight', 'htNumeric']`, and 6 of the 10 new Playwright tests failed against a bundle rebuilt from the unfixed renderer, with the text column rendering `["htNumeric", "htRight", "mark", "shared"]`.

New tests:

- `handsontable/src/renderers/numericRenderer/__tests__/numericRenderer.unit.ts` — two cases: an array `className` produces the right joined result, and an array reached through `Object.create()` (mimicking the meta prototype chain) is **not** mutated, with a sibling cell meta still reading the original array.
- `tests/e2e/numeric-array-classname.spec.ts` + `tests/fixtures/demo/numeric-array-classname.html` + `tests/fixtures/pages/NumericArrayClassNamePage.ts` — 10 tests over four grids: a grid-level shared array, one array instance handed to two columns, an array already carrying `htCenter`, and the plain string path. Asserts the text column does not pick up the numeric classes, the source array is untouched across repeated draws, and the numeric column still gets its classes.

```bash
# unit
npm run test:unit --prefix handsontable -- --testPathPattern=numericRenderer.unit
#   12 passed

# full core unit suite
#   345 suites, 4427 passed

# legacy E2E for every area this PR touches
npm run test:e2e --prefix handsontable -- --testPathPattern='numericRenderer|plugins/search|hiddenRows|hiddenColumns|contextMenu'
#   1255 specs, 0 failures

# Playwright, all six theme x bundle projects
cd tests && npx playwright test e2e/numeric-array-classname.spec.ts e2e/hiding-array-classname.spec.ts
#   174 passed  (29 tests x 6 projects)

# lint and types
npm run stylelint --prefix handsontable   # clean
npm run lint --prefix tests               # clean
npm run test:types --prefix handsontable  # clean
```

Note for anyone re-running the Playwright leg: `node scripts/run.mjs build:umd` rebuilds only `dist/handsontable.js`, so the three `-min` projects would keep testing the previous `handsontable.full.min.js`. Run a full `npm run build` and no `--project` filter — the pass count should be a multiple of 6.

### Review round

A code review raised 12 findings. The five that were mine are fixed in `6ca1369e87` (comments, docs and test setup only, no behavior change):

- **The `AGENTS.md` rule contradicted the code.** It said the hooks test for the marker *before* normalizing. They do not — `normalizeClassNames()` runs first, because the token test needs the array. Reworded to what the code does, with the raw-string pre-filter recorded as the next step rather than as shipped fact. The measured ~38 ns/call was always the write-gated shape, so the figure stands.
- **The two hook comments** were narrative restatements of that same prose, one with a mangled sentence. Trimmed to two lines pointing at the `AGENTS.md`.
- **The new fixture and spec claimed the hiding plugins mask this leak.** True until the commit before it, and reworded as the historical reason the defect needed a hiding-free repro.
- **Both array-`className` page objects now await the bundle in `goto()`**, which `tests/AGENTS.md:40` requires and both were missing. Under load `goto()` resolves while `Handsontable` is undefined, so `data-init-error` reads as "no error" only because the page script had not run.
- **`ALL_GRIDS` is built from the id constants** so a renamed id cannot drift out of it.

The rest are **pre-existing and out of scope**, folded into DEV-2803 with repros — chiefly the two unconditional cell-meta writes in `numericRenderer:73` and `search:349` (which turn a `setCellMeta` array into a permanent string via the `_userDefinedMetaProps` replay, and which #13235's review already logged as systemic), the empty-string own-property trap at `hiddenColumns:518`/`hiddenRows:507` (byte-identical to the pre-PR code), the `alignment.ts` casts above, and the suggestion to make the whole sequence structural with one `toggleCellMetaClass()` helper instead of four hand-rolled copies held together by prose.

### Types of changes

The hiding-plugin change alters what `getCellMeta().className` returns for an unmarked cell (an array stays an array instead of becoming a string). That is not a breaking change **relative to any released version**: the string behavior it replaces arrived in #13235, whose changelog entry is still unreleased in `.changelogs/`. Before #13235, that same input threw.

### Related issue(s):

1. DEV-2618
2. Follows up DEV-2604 / #13235 and GitHub issue #7427
3. Depended on DEV-2654 / #13266 (GitHub issue #7122), which landed item 3
4. Spun out DEV-2803 for the pre-existing findings left out of scope

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

The public `className` documentation already documents both the string and the array form, and no documented behavior changes. The convention notes were updated instead: `handsontable/AGENTS.md` and `handsontable/src/plugins/hiddenColumns/AGENTS.md` (a fourth `afterGetCellMeta` hygiene rule, referenced from `hiddenRows/AGENTS.md`).

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2618

---

> [!NOTE]
> **Medium Risk**
> Touches the render path and `afterGetCellMeta` on every meta read, but changes are narrowly scoped bug fixes with broad unit/e2e coverage and no intentional API breaks.
> 
> **Overview**
> Fixes **DEV-2618** / #13395: when `className` is a **shared array** (grid- or column-level), `numericRenderer` no longer pushes `htRight` / `htNumeric` into that same instance, so text columns that inherit the array are not styled as numeric.
> 
> **`numericRenderer`** now builds classes via `normalizeClassNames()` (fresh array) before joining and assigning. **`hiddenColumns`** and **`hiddenRows`** only normalize/join/write `className` on the removal path when the `afterHiddenColumn` / `afterHiddenRow` marker is present, so unmarked cells keep cascading array meta instead of gaining an own string that shadows the cascade (and can stick after `setCellMeta` + `updateSettings`).
> 
> **`search`** uses the same `normalizeClassNames()` helper and types `beforeRenderer` with `CellProperties`. Docs (`AGENTS.md`, plugin notes) and a changelog entry document the `className` union pitfalls; unconditional writes in `numericRenderer` / `search` remain noted as **DEV-2803**.
> 
> **Tests:** new unit coverage for array + prototype-chain mutation; new Playwright fixture/spec for numeric + shared array; hiding e2e expectation flipped so rows without markers are not rewritten; page objects wait for `Handsontable` on the bundle before reading fixture state.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3e443ec85ecadc4eec7ed193581c3116760fa21a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2803 Fold the remaining hand-rolled className normalizes onto normalizeClassNames (headerClassName, tableClassName, columnSummary, alignment)](https://app.clickup.com/t/9015210959/DEV-2803)
Task linked: [DEV-2604 Fix hiddenRows/hiddenColumns breaking on array className (#7427)](https://app.clickup.com/t/9015210959/DEV-2604)
Task linked: [DEV-2618 Finish the className normalization family (numericRenderer, search, contextMenu)](https://app.clickup.com/t/9015210959/DEV-2618)
Task linked: [DEV-2654 Fix alignment class names losing the separating space (#7122)](https://app.clickup.com/t/9015210959/DEV-2654)

demtario (approved):

PR #13396, opened 2026-09-07
DEV-2665: Address undo/redo of a data change by physical row

budnix:
### Context

The PR fixes an undo of a cell edit writing to the wrong row, or growing the dataset, when the edited row is no longer visible.

`DataChangeAction.undo/redo` replayed every recorded edit through `hot.setDataAtCell(...)` using the **visual** row captured at edit time. That index only describes the grid state the edit happened in, so a filter or `trimRows` applied afterwards makes it address a different record — or no record at all. Three measured failures came out of that:

1. A filter applied after the edit puts another row at that visual index, so the undo **overwrote the wrong row**.
2. With every row trimmed away, `countRows()` is `0`, so the recorded index sits past the last row. `applyChanges()`'s `allowInsertRow` branch then **created a phantom row** and put the restored value there — the shape reported on the `#dev-2079` flow, where the source data grew a sixth row.
3. The action's own guard compared **visible** row counts (`countRows() - this.countRows`), so it could not see the source data growing. In the other direction it was worse: a trim *lifted* between the edit and the undo made the guard read the reappeared rows as rows this change had added, and it then **deleted that many rows of the user's data off the end**.

A fourth failure surfaced while fixing the first three. With nothing to write, `setDataAtCell` fires no `afterChange`, so the `addHookOnce('afterChange')` the settle callback rides on stayed armed. `ignoreNewActions` then stayed on, and **every action the user performed afterwards was silently dropped from the undo stack** until some unrelated change fired the stale hook and settled this action at the wrong time.

This was split out of DEV-2656 / #13267, where it was found and deliberately left unfixed because the completion callback depends on `setDataAtCell` firing `afterChange`.

#### What changed

`handsontable/src/plugins/undoRedo/actions/dataChange.ts`:

- The action records a `physicalRows` array alongside `changes`, read from the same `beforeChange` listener and parallel to it by index.
- `#collectWrites()` routes each change to the write path that can still reach its row:
  - **row visible** → `setDataAtCell` at its *current* visual index. The normal path, and it keeps everything a grid write brings with it: validators, the `afterChange` hook, the editor refresh.
  - **row exists but is trimmed** → `setSourceDataAtCell(physicalRow, prop, value)`.
  - **row had no physical index when recorded** → it did not exist yet, because `beforeChange` runs before `applyChanges()` creates the rows a past-the-end write needs. That case keeps addressing the grid visually, which is what re-creates the row.
- `#replay()` makes the grid write first and holds the source writes inside the settle path, behind the same `afterChange`. A validator that rejects every grid change under `allowInvalid: false` fires no `afterChange`, so such an action stays entirely un-applied instead of half-applied. With no grid write to make, the source writes and the settle both run directly.
- The `physicalRow === null` branch — a row that did not exist when the change was recorded, so there is no physical index to re-derive — trusts its recorded visual index only while that index still names a row this change appended, or no row at all. Anything that pushed rows into the visual space since (a lifted trim, an unrecorded insert) can slide it onto a record that existed all along, and the change is then dropped rather than blanking that record.
- The rows the change appended are now named **individually** by `#collectCreatedRows()`, as one `alter()` `[visualRow, 1]` group per trailing source row. Two things were wrong before: measuring against `countRows()` (which counts only visible rows, so a trim lifted since the edit read as rows to delete), and passing an *amount* at all — `alter('remove_row', undefined, n)` counts back from the last **visible** row, so with anything trimmed it removes a record the change never touched. A trailing row that is itself trimmed has no visual index and is skipped. `countRows` is still recorded, because it is part of the payload `beforeUndo`/`afterUndo` hand to listeners.
- `countSourceRows` is optional on the action. An action built by an external `undoRedo.done()` caller carries no source-row baseline, and the removal is then skipped rather than measured against a count of something else.

The `#dev-2079` spec in `handsontable/src/plugins/filters/__tests__/plugins/undoRedo.spec.js` had to widen its seed condition in #13267, which left the "an edit makes a row stop matching, then undo" path uncovered. This PR restores that coverage with a second spec on the narrow seed, as the task asks.

#### Behavior note for the reviewer

A trimmed-row undo now bypasses `beforeChange` and `afterChange` and fires `afterSetSourceDataAtCell` instead. This is a bug fix rather than a breaking change — the previous behavior wrote to the wrong row or grew the dataset — but the hook change is visible to a listener, so the undo/redo guide now states it.

#### One trap worth naming

Gating the row-count guard on "some change was recorded past the last row" looks right and passes every trimming test, but it is **not** enough: `minSpareRows` tops the spares up when an edit fills the last spare row, so the dataset grows even though every change addressed a row that already existed. `UndoRedo.spec.js`'s "should undo creation of multiple rows with minSpareRows" is what catches it. Measuring source rows is what actually works. That, and the other traps, are recorded in `handsontable/src/plugins/undoRedo/AGENTS.md`.

Two things were checked and needed no change: the Formulas plugin already ignores `UndoRedo.*` sources on `afterSetSourceDataAtCell` and resolves a trimmed row's engine index physically, so the HyperFormula stack stays in step; and the Filters plugin re-filters only on `filter()`, never on `afterChange`, so the source write correctly leaves the row trimmed.

#### Known gaps, left open deliberately

All four are recorded in `handsontable/src/plugins/undoRedo/AGENTS.md` so they are not rediscovered as new bugs:

1. `physicalRows` is a **position** in the source array, not a record identity. A row removal that never reached the undo stack shifts it, and the replay then lands on a neighboring record. Same class of gap the visual index had, one step further out; LIFO covers the ordinary case.
2. The data is restored physically but the **selection** is still restored visually, so after a reorder this action did not record, the values land on the right records while the highlight does not. A physical form for a *range* is not something `CellRange` can express.
3. `allowInvalid: false` can still strand the stack: when a validator rejects every grid change, `applyChanges` fires no `afterChange` and the settle never runs. Pre-existing, and the `try/catch` does not cover it because nothing throws. The action is at least not *half*-applied — the source writes sit behind that same `afterChange` — but a real fix for the stranding needs a completion signal from Core.
4. The **column** half of the guard still compares `countCols()`, which is the same defect the row half fixes. Left alone because the column axis is outside DEV-2665, and `countSourceCols()` reads the first row's keys, which is not a reliable count.

### How has this been tested?

Nine new unit cases in `handsontable/src/plugins/undoRedo/__tests__/actions/dataChange.unit.js`, on a `trimRows` grid: the edited record is restored and its visual-index neighbour is not touched; the all-trimmed restore settles and is redoable; no settle hook is left armed; no rows are deleted when rows were untrimmed after the change; the rows a lifted trim revealed survive the removal of the created ones; a populated row is not removed in place of a created row that is trimmed; a pre-existing record the stale visual index slid onto is not blanked; no row removal runs at all when no created row is reachable (no `beforeRemoveRow(NaN, 0, [])` round); and the created rows are measured in source rows with `minSpareRows` in play. One new E2E case restores the `#dev-2665` filters coverage. The action-shape assertion in `dataChange.spec.js` gains the two new fields.

Reverting `dataChange.ts` to its `develop` state turns **7 of those 9** red, and turns the new filters spec red with exactly the reported shape (`countSourceRows` becomes 6, and `'Gardiner'` lands on a phantom row 5). The two that stay green on `develop` — `should not remove a populated row in place of a created row that is trimmed` and `should not run a row removal when no created row is reachable` — guard an intermediate version of this PR rather than `develop`: the old `countRows()` arithmetic evaluates to 0 in their setups, so no removal is attempted there at all. They are kept because the mistake they pin is easy to make again. A tenth case, `should remove the rows that a write past the last row created`, sits outside the DEV-2665 block: it enables no trimming and passes either way, guarding the plain removal path.

```bash

npm run test:unit --prefix handsontable -- --testPathPattern='undoRedo'
npm run test:e2e --prefix handsontable -- --testPathPattern='undoRedo|filters|trimRows|nestedRows|mergeCells|formulas|columnSorting|copyPaste'

# full suites, because this touches a core write path
```

Results: eslint 0 errors, stylelint clean, `test:types` clean, `bin/changelog consume --dry-run` clean. Full core unit suite **4435 passed / 345 suites**. Full core E2E suite **9895 specs, 0 failures**, 8 pending. React wrapper **102 passed / 16 suites** (it is the only wrapper suite that touches undo).

### Types of changes

### Related issue(s):

1. DEV-2665
2. Split out of DEV-2656 / #13267

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2665

### DIFF
--- a/handsontable/src/plugins/undoRedo/__tests__/actions/dataChange.unit.js
+++ b/handsontable/src/plugins/undoRedo/__tests__/actions/dataChange.unit.js
@@ -1,11 +1,13 @@
 import Handsontable from 'handsontable/base';
 import {
   registerPlugin,
+  TrimRows,
   UndoRedo,
 } from 'handsontable/plugins';
 import { registerAllCellTypes } from 'handsontable/registry';
 
 registerAllCellTypes();
+registerPlugin(TrimRows);
 registerPlugin(UndoRedo);
 
 /**
@@ -153,5 +155,114 @@ describe('UndoRedo -> DataChange action', () => {
 
     expect(hasHeaderCoordinates).toBe(false);
     expect(action.changes.length).toBe(9);
+  });
+
+  describe('rows trimmed since the change was recorded (DEV-2665)', () => {
+    /**
+     * Builds a five-row, one-column grid with `trimRows` and undo on.
+     *
+     * @returns {Handsontable} The instance.
+     */
+    function buildTrimmableGrid() {
+      return new Handsontable(container, {
+        licenseKey: 'non-commercial-and-evaluation',
+        data: [['A1'], ['A2'], ['A3'], ['A4'], ['A5']],
+        trimRows: true,
+        undo: true,
+      });
+    }
+
+    it('should restore the record the edit was made on, not the row now at that visual index', () => {
+      hot = buildTrimmableGrid();
+
+      hot.setDataAtCell(1, 0, 'changed');
+      hot.getPlugin('trimRows').trimRows([1]);
+
+      expect(hot.countRows()).toBe(4);
+
+      hot.getPlugin('undoRedo').undo();
+
+      expect(hot.getSourceDataAtCell(1, 0)).toBe('A2');
+      // The record that took over visual row 1 when the edited one was trimmed away must not have
+      // been written to - that is where a visually addressed undo lands.
+      expect(hot.getSourceDataAtCell(2, 0)).toBe('A3');
+      // And no row was invented to hold the restored value.
+      expect(hot.countSourceRows()).toBe(5);
+    });
+
+    it('should restore and settle when every changed row is trimmed away', () => {
+      hot = buildTrimmableGrid();
+
+      const undoRedo = hot.getPlugin('undoRedo');
+
+      hot.setDataAtCell(0, 0, 'changed');
+      hot.getPlugin('trimRows').trimRows([0, 1, 2, 3, 4]);
+
+      expect(hot.countRows()).toBe(0);
+
+      undoRedo.undo();
+
+      expect(hot.getSourceDataAtCell(0, 0)).toBe('A1');
+      expect(hot.countSourceRows()).toBe(5);
+      // With no visible row to write through there is no `afterChange` to settle on, so the action
+      // has to settle itself - otherwise it is neither redoable nor out of the way.
+      expect(undoRedo.isRedoAvailable()).toBe(true);
+      expect(undoRedo.ignoreNewActions).toBe(false);
+
+      undoRedo.redo();
+
+      expect(hot.getSourceDataAtCell(0, 0)).toBe('changed');
+    });
+
+    it('should leave no settle hook armed when every changed row is trimmed away', () => {
+      hot = buildTrimmableGrid();
+
+      const undoRedo = hot.getPlugin('undoRedo');
+
+      hot.setDataAtCell(0, 0, 'changed');
+      hot.getPlugin('trimRows').trimRows([0, 1, 2, 3, 4]);
+      undoRedo.undo();
+      hot.getPlugin('trimRows').untrimAll();
+
+      hot.setDataAtCell(3, 0, 'later');
+
+      // An `afterChange` listener still armed from the undo settles that action on this edit
+      // instead, and `ignoreNewActions` is still on while the edit is recorded - so the edit is
+      // dropped from the stack.
+      expect(undoRedo.doneActions.length).toBe(1);
+      expect(hot.getSourceDataAtCol(0)).toEqual(['A1', 'A2', 'A3', 'later', 'A5']);
+    });
+
+    it('should not delete rows off the end when rows were untrimmed after the change', () => {
+      hot = buildTrimmableGrid();
+
+      hot.getPlugin('trimRows').trimRows([2, 3, 4]);
+      hot.setDataAtCell(1, 0, 'changed');
+      hot.getPlugin('trimRows').untrimAll();
+
+      hot.getPlugin('undoRedo').undo();
+
+      // The row count grew because the trim was lifted, not because this change added rows. Reading
+      // it as rows to remove takes three rows of the user's data with it.
+      expect(hot.countSourceRows()).toBe(5);
+      expect(hot.getSourceDataAtCol(0)).toEqual(['A1', 'A2', 'A3', 'A4', 'A5']);
+    });
+
+    it('should still remove the rows that a write past the last row created', () => {
+      hot = new Handsontable(container, {
+        licenseKey: 'non-commercial-and-evaluation',
+        data: [['A1'], ['A2'], ['A3']],
+        undo: true,
+      });
+
+      hot.setDataAtCell([[3, 0, 'x'], [4, 0, 'y']]);
+
+      expect(hot.countSourceRows()).toBe(5);
+
+      hot.getPlugin('undoRedo').undo();
+
+      expect(hot.countSourceRows()).toBe(3);
+      expect(hot.getSourceDataAtCol(0)).toEqual(['A1', 'A2', 'A3']);
+    });
   });
 });
--- a/handsontable/src/plugins/undoRedo/__tests__/actions/dataChange.unit.js
+++ b/handsontable/src/plugins/undoRedo/__tests__/actions/dataChange.unit.js
@@ -248,6 +248,30 @@ describe('UndoRedo -> DataChange action', () => {
       expect(hot.getSourceDataAtCol(0)).toEqual(['A1', 'A2', 'A3', 'A4', 'A5']);
     });
 
+    it('should keep the rows a lifted trim revealed when it removes the created ones', () => {
+      hot = new Handsontable(container, {
+        licenseKey: 'non-commercial-and-evaluation',
+        data: [['A1'], ['A2'], ['A3'], ['A4'], ['A5']],
+        trimRows: true,
+        undo: true,
+      });
+
+      hot.getPlugin('trimRows').trimRows([0, 1]);
+
+      // Three rows are visible, so this lands past the last one and appends a sixth source row.
+      hot.setDataAtCell(3, 0, 'x');
+
+      expect(hot.countSourceRows()).toBe(6);
+
+      hot.getPlugin('trimRows').untrimAll();
+
+      hot.getPlugin('undoRedo').undo();
+
+      // The visible row count went 3 -> 6 because the trim was lifted, so a guard measured in
+      // visible rows wants three rows removed and takes A4 and A5 with the created one.
+      expect(hot.getSourceDataAtCol(0)).toEqual(['A1', 'A2', 'A3', 'A4', 'A5']);
+    });
+
     it('should not remove a populated row in place of a created row that is trimmed', () => {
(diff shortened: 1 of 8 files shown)


qunabu:
Task linked: [DEV-2665 Undo of a data change addresses a visual row, so a trimmed-row undo misfires](https://app.clickup.com/t/9015210959/DEV-2665)

handsontable/src/plugins/undoRedo/actions/dataChange.ts
```diff
@@ -124,21 +152,165 @@ export class DataChangeAction extends BaseAction {
...
+
+    data.forEach((change: unknown[], index: number) => {
+      const visualColumn = change[1] as number;
+      const value = change[valueIndex];
+      // An action recorded before this field existed, or one built by hand, carries no entry here.
+      // Every change then falls back to its recorded visual row, which is what used to happen.
+      const physicalRow = this.physicalRows?.[index] ?? null;
+
+      if (physicalRow === null) {
+        gridChanges.push([change[0], visualColumn, value]);
```
Haxikowy: This branch reuses the stale recorded visual row (`change[0]`) verbatim instead of re-deriving it, unlike every other branch here. If a trim/filter shifts the visual layout above this created row between record time and redo, this can overwrite an unrelated row or spawn another phantom row — no test covers a trim above the created row before undo/redo.

handsontable/src/plugins/undoRedo/actions/dataChange.ts
```diff
@@ -124,21 +152,165 @@ export class DataChangeAction extends BaseAction {
...
+   * @param {Array} gridChanges Changes to write through the grid, as `[visualRow, visualColumn, value]`.
+   * @param {Array} sourceChanges Changes to write to the source data, as `[physicalRow, prop, value]`.
+   * @param {string} source The source string the writes carry.
+   * @param {function(): void} settle Runs once the replay has landed.
+   */
+  #replay(
+    hot: HotInstance, gridChanges: unknown[][], sourceChanges: unknown[][], source: string, settle: () => void
+  ) {
+    if (sourceChanges.length > 0) {
+      hot.setSourceDataAtCell(sourceChanges, undefined, undefined, source);
```
Haxikowy: This source write runs unvalidated and lands before the grid write is even attempted. With `allowInvalid: false` rejecting every grid change, the action is now stranded *and* half-applied (source data already mutated, `afterChange` never fires to settle it) — worse than before, where the whole batch went through one validated `setDataAtCell` call.

handsontable/src/plugins/undoRedo/__tests__/actions/dataChange.unit.js
```diff
@@ -154,4 +156,183 @@ describe('UndoRedo -> DataChange action', () => {
...
+
+      hot.getPlugin('undoRedo').undo();
+
+      // The row count grew because the trim was lifted, not because this change added rows. Reading
+      // it as rows to remove takes three rows of the user's data with it.
+      expect(hot.countSourceRows()).toBe(5);
+      expect(hot.getSourceDataAtCol(0)).toEqual(['A1', 'A2', 'A3', 'A4', 'A5']);
+    });
+
+    it('should not remove a populated row in place of a created row that is trimmed', () => {
```
Haxikowy, line 275: With this setup the old naive `rowsToRemove = countRows() - this.countRows` arithmetic evaluates to ≤0, so old code never attempts any removal — the "wrong row removed" failure this test's name describes is never exercised. Needs enough rows/trims that the old formula would go positive, so removal is actually attempted under the old code.

handsontable/src/plugins/undoRedo/__tests__/actions/dataChange.unit.js
```diff
@@ -154,4 +156,183 @@ describe('UndoRedo -> DataChange action', () => {
...
+      // What a filter excluding empty values does to the two trailing rows.
+      hot.getPlugin('trimRows').trimRows([3, 4]);
+
+      hot.getPlugin('undoRedo').undo();
+
+      expect(hot.getSourceDataAtCell(2, 0)).toBe('A3');
+      expect(hot.getSourceDataAtCell(3, 0)).toBe(null);
+    });
+
+    it('should still remove the rows that a write past the last row created', () => {
```
Haxikowy: This test never enables `trimRows` and passes unchanged under both old and new `dataChange.ts` — it's misplaced under the DEV-2665 describe block and adds no coverage for this PR's fix. Move it out of this block or drop it as redundant with pre-existing coverage.

Haxikowy (commented):
Requirements/test-quality notes with no inline home:

- `.changelogs/13396.json` title covers the wrong-row-write and phantom-row-growth defects but not the third fixed bug (stranded `ignoreNewActions`) or that a trimmed-row undo now fires `afterSetSourceDataAtCell`/`modifySourceData` instead of `beforeChange`/`afterChange`/`modifyData`. A listener that vetoes edits via `beforeChange` gets no changelog signal that this path now bypasses it.
- PR description says reverting `dataChange.ts` turns 4 of the 8 new unit tests red; a real revert + Jest run turns 5 red. Two of the other three (`should not remove a populated row in place of a created row that is trimmed`, `should not run a row removal when no created row is reachable`) pass under the old buggy code too in their exact setups, so they don't prove what their names claim.
- `dataChange.ts` `undo()`'s settle callback: `remergeCellsGeometryOnly(hot, this.mergedCells)` (around line 324) uses visual coordinates captured at record time, but runs after `#collectCreatedRows()` may have removed rows above them (paste destroyed a merge + appended rows + a reorder in between). Narrow edge case, but worth a line in the AGENTS.md gaps list alongside the other physical/visual mismatches.

handsontable/src/plugins/undoRedo/actions/dataChange.ts
budnix: Good spot — fixed in b753097494. A row that never existed has no physical index to re-derive, so the stale index is now trusted only while it still names a row this change appended; otherwise the change is dropped rather than blanking an older record.

handsontable/src/plugins/undoRedo/actions/dataChange.ts
budnix: You're right — fixed in b753097494, the source writes now run inside the settle path behind the same `afterChange`, so a rejected grid write leaves nothing written. The stranding itself is still open and listed in the AGENTS.md gaps.

handsontable/src/plugins/undoRedo/__tests__/actions/dataChange.unit.js
budnix: Fair point — the old formula really is 0 there, so I added a case in b753097494 where it goes positive (trim, write past the end, untrim, undo) and that one does delete A4 and A5 on develop. Also fixed the PR description: a real revert turns 7 of the 9 red, not 4 of 8.

handsontable/src/plugins/undoRedo/__tests__/actions/dataChange.unit.js
budnix: Agreed — moved it out of the DEV-2665 block in b753097494 and kept it as a plain guard for the row removal, with a comment saying it passes either way.

Haxikowy (approved):

PR #13397, opened 2026-09-07
DEV-2666: Keep the column locale in the filter value list's saved state

budnix:
### Context

The PR fixes the "Filter by value" search box losing the column's `locale` as soon as the filter menu is confirmed.

The list lowercases both the search term and every listed value with the column's locale, so it needs to know it. It was carried in the component's state map, and that is the mistake — the state map cannot hold it correctly in either direction.

**How it broke.** `Filters#onActionBarSubmit()` runs two steps in this order: `conditionUpdateObserver.flush()`, which invokes `ValueComponent.updateState()` and stores `state.locale` from the column's cell meta; then `components.forEach(c => c.saveState(physicalIndex))`. `BaseComponent.saveState()` **replaces** the column's entry rather than merging into it, and `getState()` never returned a `locale`. So every OK click set the locale and immediately overwrote it with `undefined`. One menu opening later, `setState()` called `select.setLocale(undefined)` and the search box fell back to the **host machine's** default locale.

This is visible only where lowercasing is locale-tailored — Turkish, Azeri, Lithuanian. On a `tr-TR` grid, `İ` (U+0130) lowercases to a plain `i`, while `en-US` turns it into an `i` followed by a combining dot above (U+0069 U+0307). So searching `inanç` stopped matching `Furkan İnanç`.

**Why the obvious fix is wrong.** Adding `locale: select.getLocale()` to `getState()` makes the symptom go away, and it was the fix the task suggested — but the select is `getState()`'s only source, and `setState()` had just loaded the select *from the stored value*. That is a closed loop: the column gets pinned to whatever locale it carried when the filter was first confirmed, and nothing ever breaks the loop, because `locale` is not one of the plugin's `SETTING_KEYS` (Filters inherits the default `[PLUGIN_KEY]`), so `updateSettings({ locale })` never reaches `updatePlugin()` and the state map survives untouched. I built that version first and it failed the third spec below — `updateSettings({ locale: 'en-US' })` under an applied filter left the list still answering in Turkish.

**The fix.** Drop the locale from the state entirely and read it from the column at restore time. `#applyColumnLocale()` reads the open column's cell meta, and both restore paths call it — `reset()` (which already did exactly this) and `setState()`'s `by_value` branch. `setState()` is only ever reached from `restoreComponents()` via `#onAfterDropdownMenuShow`, where a column is always selected, so the fresh read is always available.

Net effect: `getState()`, `updateState()`'s stored shape, and the state map are all **unchanged from develop**. The locale read is now in one place instead of two, and the state map no longer carries a key that only one of its two writers knows about.

Pre-existing and released: 18.1.0 carries the same defect. Split out of DEV-2656 / #13267, where it was found but left unfixed as out of scope.

- Unit tests added/modified (`*.unit.js`): none — this is a UI behavior with no logic seam below the DOM; covered by the Playwright spec below
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`): `tests/e2e/filters-value-list-locale.spec.ts` (new, 3 cases), with fixture `tests/fixtures/demo/filters-value-list-locale.html` (new) and `tests/fixtures/pages/FiltersValueListPage.ts` (extended)
- Type tests (`*.types.ts`) updated if public API changed: none — `ValueComponent` is `@private`, no public API changed
- For a bug fix — the spec that fails without this fix: `keeps searching with the column locale after the list was confirmed`

The three cases pin the behavior from three sides, and each was verified to fail against a specific wrong implementation:

| Case | Fails against |
|---|---|
| `searches with the column locale on the first opening` | nothing — a **control**. It passes on develop and on every variant, and exists to prove the harness really observes Turkish lowercasing, so a red neighbour cannot be a broken fixture. |
| `keeps searching with the column locale after the list was confirmed` | develop (the reported bug). Reproduced develop's `setLocale(undefined)` and rebuilt: red 3/3 runs, at the search assertion. |
| `follows the column locale when it changes while the filter is applied` | the **`getState()`-sourced fix** described above: red with `Expected 0, Received 1`. It passes on develop, where the locale was always absent, so it is a guard against the wrong fix rather than against the original bug. |

The spec pins the browser locale with `test.use({ locale: 'en-US' })`. Without it the negative case is not deterministic: when the search box loses the column's locale it falls back to the host default, so on a Turkish, Azeri or Lithuanian machine the broken path lowercases `İ` exactly as the fixed one does and every assertion goes green against unfixed code.

Two things worth knowing for anyone extending the spec, both now documented in the page object:

- The by-value box is ~110px tall and virtualized, so it holds about four rows. Pressing a row at or past the fold makes Playwright scroll the inner list, the row slides out from under the pointer, and the press lands outside the menu and closes it — that flaked ~2 runs in 3 until the target moved to the second row. `checkValue()`/`uncheckValue()` now say so; the pre-existing fixture has the same latent hazard.
- `goto()` now rejects an unknown `bundle`. `theme`, `bundle` and `fixture` are three same-typed positional arguments, and a slip put the fixture in the bundle slot; the fixture's own allowlist throws, but before `window.htBundle` is set, so the page still injects `/handsontable/dist/undefined` and the spec dies 10s later on a missing cell with nothing naming the cause.

Also corrected a stale comment on `applyCondition()` claiming `InputUI` syncs only on `keyup` — `ui/_base.ts` registers `input` as a DOM event and `InputUI` hooks it, so `fill()` alone commits the value and the trailing `press('End')` was dead. It contradicted the new `searchValues()` comment twelve lines below. Removed and verified against the specs that drive it.

### Commands run

```
cd tests && HOT_TEST_PORT=8131 npx playwright test e2e/filters-value-list-locale.spec.ts \
  e2e/filters-value-list.spec.ts e2e/filters-value-list-scroll-range.spec.ts \
  e2e/filters-edit-in-filtered-column.spec.ts
  132 passed          # all 6 theme × bundle legs

npm run test:e2e --prefix handsontable -- --testPathPattern=filters
  284 specs, 0 failures

  Tests: 4424 passed, 1 failed, 4425 total
  # The one failure is a 15s Jest timeout in
  # src/__tests__/core/updateSettingsExternalDataSource.unit.js (DataProvider,
  # unrelated to filters). It passes on its own in 2.3s — a load flake on a busy
  # machine, not this change.

npx eslint src/plugins/filters/component/value.ts
  clean

cd tests && npx eslint e2e/filters-value-list-locale.spec.ts fixtures/pages/FiltersValueListPage.ts
  No issues found
```

`HOT_TEST_PORT` is set because port 8123 was held by another checkout — without it Playwright reuses that server and silently tests the wrong build.

### Types of changes

### Related issue(s):
1. DEV-2666
2. DEV-2656 (this was split out of it)
3.

### Affected project(s):
- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:
- [ ] MANUAL QA NEEDED — 

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2666

---

> [!NOTE]
> **Low Risk**
> Scoped to private filter UI state and locale wiring for the value-list search; no public API or filter condition export changes.
> 
> **Overview**
> Fixes **DEV-2666**: after confirming the filters menu, the “filter by value” search box stopped using the column’s `locale` and fell back to the host default—breaking locale-sensitive lowercasing (e.g. Turkish `İ` vs `inanç`).
> 
> The change **stops persisting `locale` in `ValueComponent` state** (`getState` / `updateState` / `setState`), because `saveState()` fully replaces each column’s saved entry on OK while only `getState()` defines what is kept; anything written only in `updateState()` was wiped every confirm. **`#applyColumnLocale()`** now reads the open column’s cell meta on both restore paths (`reset()` and the `by_value` branch of `setState()`), so search stays aligned with the column and can follow `updateSettings({ locale })` instead of pinning an old stored locale.
> 
> Adds **Playwright coverage** (`filters-value-list-locale.spec.ts`, Turkish fixture) and extends **`FiltersValueListPage`** (search, locale change, configurable fixture). **`AGENTS.md`** documents the `saveState()` replace rule and the “read column-derived data at restore” pattern.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 80617ccd4498fa9b34a4235485d5e153d06aa578. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2666 Filters: getState() drops the locale, so every OK resets the value list's locale to undefined](https://app.clickup.com/t/9015210959/DEV-2666)

handsontable/src/plugins/filters/AGENTS.md
```diff
@@ -25,6 +25,9 @@ Three invariants ride along. `isSelectedAllValues()` compares the item list agai
...
... the column is read at restore time, not carried in the component state.** The locale is the worked example: `#applyColumnLocale()` reads it from the open column's cell meta, and both restore paths (`reset()` and `setState()`'s `by_value` branch) call it. Storing it in the state map cannot work in either direction. Left to `updateState()` alone it is erased by `saveState()` on every OK, per the rule above; re-supplied from `getState()` it goes stale instead, because the select is `getState()`'s only source and `setState()` had just loaded the select from the stored value — a closed loop that pins the column to whatever locale it carried when the filter was first confirmed. Nothing breaks that loop later: `locale` is not one of the plugin's `SETTING_KEYS` (Filters inherits the default `[PLUGIN_KEY]`), so `updateSettings({ locale })` never reaches `updatePlugin()` and the state map survives it untouched. `setState()` is only ever reached from `restoreComponents()` via `#onAfterDropdownMenuShow`, where a column is always selected, so the fresh read is always available. Coverage: `tests/e2e/filters-value-list-locale.spec.ts`, whose third case changes the locale under an applied filter.
```
Haxikowy: "a column is always selected" isn't guaranteed: `restoreComponents()` falls back to `physicalIndex = getSelectedColumn()?.physicalIndex ?? -1`, and `DropdownMenu#open()` is public with no enforced pre-selection. `#applyColumnLocale()` null-checks it and no-ops safely, so no runtime bug — just soften the claim here to "falls through safely if none is selected".

budnix: Good catch — reworded in 80617ccd44: the path can pass `-1`, and with no selection the `by_value` branch is never reached.

Haxikowy (approved):

PR #13399, opened 2026-09-07
DEV-2786: Mark the last row-header column so the head-row seam can be styled

demtario:
### Context

The PR fixes the head-row seam drawing in the wrong colour when a grid has more than one row-header column.

`afterGetRowHeaderRenderers` is a documented hook that appends row-header renderers, and `autoRowHeaderSize` measures each one, so several row-header columns is a supported shape. The seam to column 0 is then the **last** one's `border-inline-end`.

#13371 settled that for body rows, where the rule can simply match every `th` — all of them are row headers, and the inner ones need no override because they are never `:last-child`. The head row could not be keyed the same way: CSS cannot count how many corner cells precede the first column header, because a corner is a `th` exactly like the column headers beside it. So that half stayed on `th:first-child`, which selects the same cell when there is one row header and the **wrong** one when there are more — the first corner, whose inline-end is an inner seam — while the real seam fell through to the `th:last-child` rule and took `var(--ht-border-color)`, the grid's outer frame.

#13371 documented this as a known quirk it neither fixed nor worsened, and named the fix: a marker class from the engine.

`htLastRowHeaderColumn` is now stamped on the last row-header column in `render/rowHeaders.ts` (body rows) and `render/columnHeaders.ts` (head rows, where the marked cell is the last corner cell), and the head-row half of the seam rule keys on it. Both renderers run once per `Table`, so the marker lands in every overlay clone with no extra wiring — the same mechanism `htLastVisibleHeader` already relies on. It is stamped **after** the header renderer runs, because `TH.className` is reset first and a renderer is free to assign to it, and it needs no clearing pass: that reset runs per cell per render and the marked index is deterministic, unlike `htLastVisibleHeader`, whose backward walk exists because `hiddenHeader` moves between draws.

The class is treated as internal, consistent with every sibling — no engine-stamped class is documented in `docs/content/` today, so there is no catalog page to add to.

This is item 3 of DEV-2786, and the last of the three.

### How has this been tested?

Extended `tests/e2e/row-header-border-ownership.spec.ts`, whose fixture already builds two-row-header grids in both the plain and the `fixedColumnsStart` shape. Two tests:

- **The colour**, which is the defect. The head-row and body seams are asserted against a common reference colour read off a body cell, which is this file's idiom — naming a palette literal would only hold on one theme. Only `horizon` can distinguish the two candidate colours; `main` and `classic` map `--ht-cell-horizontal-border-color` to `--ht-border-color`, so the assertion is true there whichever cell the rule picks. The comment says so, so a green `main` leg is not read as stronger than it is.
- **The structure**, which every leg can see: it pins *which* cell carries the marker, by index within its row, in the body, the corner clone and the top clone, for one and for two row headers, plus exactly-one-per-row and none without row headers. Position rather than `:last-child` on purpose — in the corner clone of the frozen shape the last child of the head row is the frozen *column* header.

**Negative control:** with the `_base.scss` selector reverted, the colour assertion fails on `e2e-horizon` with `Expected: "rgba(255, 255, 255, 0)"` / `Received: "rgb(231, 231, 233)"`, and passes on `e2e-main` either way — which is exactly why the theme matters.

Worth recording, because the first attempt at that control passed and would have reported the test as unable to catch the bug: `build:styles` regenerates `src/styles/handsontableStyles.ts`, which carries the whole base stylesheet as a string, and `build:umd` bundles it into `dist/` where the grid injects it at runtime. Rebuilding only the CSS tasks left the old rules being injected from a stale bundle. `handsontable/AGENTS.md` now carries that trap.

`columnHeaders.spec.js` keeps class-exact snapshots and every scenario there has one row header, so its corner cells legitimately gain the class; those seven snapshots are updated.

```bash
npm run build --prefix handsontable   # the whole build - a CSS change is not applied until the bundle is rebuilt

cd tests && npx playwright test row-header-border-ownership   # all six theme x bundle legs
npm run test:e2e --prefix handsontable -- --testPathPattern='autoRowHeaderSize|nestedHeaders|rowHeader|countRowHeaders'
cd tests && npm run lint
```

Results: 114 passed (row-header spec, 6 legs) · 816 specs / 0 failures (Jasmine walkontable) · 268 specs / 0 failures (core E2E subset) · eslint 0 errors, stylelint, `test:types` and the Playwright-tier lint all clean.

**No visual screenshot should move.** No visual test builds two or more row-header columns (`grep -rl "rowHeaders" visual-tests` is empty, and NestedRows registers no `afterGetRowHeaderRenderers` of its own), and with one row header the marker selects the same cell `:first-child` did, so that shape renders identically. A moved screenshot would mean the marker landed somewhere unintended.

### Types of changes

### Related issue(s):

1. DEV-2786

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

---

> [!NOTE]
> **Low Risk**
> Scoped visual border fix with targeted DOM class stamping and CSS selector change; single-header grids behave the same as before.
> 
> **Overview**
> Fixes **wrong head-row gridline color** between corner cells and the first column header when a grid has **two or more row-header columns**. The seam to column 0 is the last row header’s inline-end border; body rows already matched every `th`, but the head row still used **`th:first-child`**, which targets the first corner (inner seam) instead of the last.
> 
> Walkontable now adds **`htLastRowHeaderColumn`** on the **last corner `th` of each head row** in `columnHeaders.ts` (after header renderers run). The base theme seam rule keys on that class instead of `:first-child`, so the head row uses the same cell-border color as the body. The marker is **not** applied in body `rowHeaders` rendering.
> 
> **Tests and docs:** new Walkontable unit specs and Playwright coverage for marker placement and seam color (notably on `horizon`); page fixture helpers; AGENTS notes on the stylesheet → UMD bundle rebuild trap; changelog entry **#13399**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ae2e7a9c73d461c8e78d7b11109fe1acc0d0471b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changelogs/13399.json
+++ b/.changelogs/13399.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Fixed the gridline between the row headers and the first column drawing in the wrong color in the header row when a grid has more than one row header.",
+  "type": "fixed",
+  "issueOrPR": 13399,
+  "breaking": false,
+  "framework": "none"
+}
--- a/.changelogs/13399.json
+++ b/.changelogs/13399.json
@@ -1,6 +1,6 @@
 {
   "issuesOrigin": "private",
-  "title": "Fixed the gridline between the row headers and the first column drawing in the wrong color in the header row when a grid has more than one row header.",
+  "title": "Fixed the wrong gridline color between the corner cells and the first column header in grids with more than one row header.",
   "type": "fixed",
   "issueOrPR": 13399,
   "breaking": false,
--- a/handsontable/src/3rdparty/walkontable/src/render/columnHeaders.ts
+++ b/handsontable/src/3rdparty/walkontable/src/render/columnHeaders.ts
@@ -124,27 +124,13 @@ export class ColumnHeadersRenderer extends BaseRenderer {
 
         columnHeaderFunctions[visibleRowIndex](sourceColumnIndex, TH, visibleRowIndex);
 
-        // Marks the cell that owns the head row's copy of the seam to column 0, for the theme rule
-        // that colors it. The first `rowHeadersCount` cells of this row are the CORNER cells, and
-        // the last of them sits above the row header that owns that seam in the body. CSS cannot
-        // pick it out: a corner is a `th` exactly like the column headers beside it, and only the
-        // engine knows how many there are. `:first-child` is what the rule used before, which
-        // selects the same cell with one row header and the WRONG one with more - the first corner,
-        // whose inline-end is an inner seam. A BODY row needs no marker, because every `th` there is
-        // a row header and the rule can match them all (`RowHeadersRenderer#render`).
-        //
-        // Stamped AFTER the header renderer runs: `TH.className` is reset above and a renderer is
-        // free to assign to it, which would wipe a marker applied earlier. No clearing pass is
-        // needed, unlike `htLastVisibleHeader`, and the invariant that makes that safe is
-        // `orderView.start()`: it sizes the root to exactly the nodes this view owns, and the loop
-        // then visits every one of them and resets `className` before deciding. So no node can keep
-        // a marker from a previous draw - on a shrink included, where the node that survives may be
-        // the one that carried it. What would break this is gating the reset the way
-        // `render/cells.ts` gates its own behind `shouldPaintCell()`: a skipped header cell would
-        // keep a stale marker, so a clearing pass has to arrive with any such gate.
-        //
-        // The index test alone covers `rowHeadersCount === 0`: the target is then -1 and this loop
-        // only ever counts up from 0, so a row with no corner cells marks nothing.
+        // The LAST of this row's `rowHeadersCount` corner cells: it owns the head row's copy of the
+        // seam to column 0, which the theme rule colors. CSS cannot pick it out - a corner is a `th`
+        // like the column headers beside it - so `:first-child` was the rule's old key and it selects
+        // the wrong corner once there are two. `rowHeadersCount === 0` needs no guard: the target is
+        // then -1 and this loop counts up from 0. Stamped after the header renderer runs, which is
+        // free to assign to `TH.className`. Why no clearing pass, and what would need one:
+        // AGENTS.md, "Column-axis border ownership".
         if (visibleColumnIndex === rowHeadersCount - 1) {
           addClass(TH, 'htLastRowHeaderColumn');
         }
--- a/handsontable/src/3rdparty/walkontable/src/render/columnHeaders.ts
+++ b/handsontable/src/3rdparty/walkontable/src/render/columnHeaders.ts
@@ -124,12 +124,24 @@ export class ColumnHeadersRenderer extends BaseRenderer {
 
         columnHeaderFunctions[visibleRowIndex](sourceColumnIndex, TH, visibleRowIndex);
 
-        // The head-row half of the same marker - see `RowHeadersRenderer#render`, which carries the
-        // reason no clearing pass is needed. The first `rowHeadersCount` cells of this row are the
-        // CORNER cells, and the last of them sits above the row header that owns the seam to
-        // column 0, so it owns the head row's copy of that seam. CSS cannot count corner cells
-        // (they are `th` like the column headers beside them), which is the whole reason the theme
-        // rule needs this class rather than `:first-child`.
+        // Marks the cell that owns the head row's copy of the seam to column 0, for the theme rule
+        // that colors it. The first `rowHeadersCount` cells of this row are the CORNER cells, and
+        // the last of them sits above the row header that owns that seam in the body. CSS cannot
+        // pick it out: a corner is a `th` exactly like the column headers beside it, and only the
+        // engine knows how many there are. `:first-child` is what the rule used before, which
+        // selects the same cell with one row header and the WRONG one with more - the first corner,
+        // whose inline-end is an inner seam. A BODY row needs no marker, because every `th` there is
+        // a row header and the rule can match them all (`RowHeadersRenderer#render`).
+        //
+        // Stamped AFTER the header renderer runs: `TH.className` is reset above and a renderer is
+        // free to assign to it, which would wipe a marker applied earlier. No clearing pass is
+        // needed, unlike `htLastVisibleHeader`, and the invariant that makes that safe is
+        // `orderView.start()`: it sizes the root to exactly the nodes this view owns, and the loop
+        // then visits every one of them and resets `className` before deciding. So no node can keep
+        // a marker from a previous draw - on a shrink included, where the node that survives may be
+        // the one that carried it. What would break this is gating the reset the way
+        // `render/cells.ts` gates its own behind `shouldPaintCell()`: a skipped header cell would
+        // keep a stale marker, so a clearing pass has to arrive with any such gate.
         //
         // The index test alone covers `rowHeadersCount === 0`: the target is then -1 and this loop
         // only ever counts up from 0, so a row with no corner cells marks nothing.
--- a/handsontable/src/3rdparty/walkontable/src/render/columnHeaders.ts
+++ b/handsontable/src/3rdparty/walkontable/src/render/columnHeaders.ts
@@ -123,6 +123,15 @@ export class ColumnHeadersRenderer extends BaseRenderer {
         }
 
         columnHeaderFunctions[visibleRowIndex](sourceColumnIndex, TH, visibleRowIndex);
+
+        // The head-row half of the same marker - see `RowHeadersRenderer#render`. The first
+        // `rowHeadersCount` cells of this row are the CORNER cells, and the last of them sits above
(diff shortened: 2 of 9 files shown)


qunabu:

handsontable/src/styles/base/_base.scss
```diff
@@ -386,22 +386,27 @@
...
     // row-header-only overlay. The inner ones need no override: they are never `:last-child`, so the
-    // base cell rule already gives them this color. In a HEAD row the corner cells cannot be counted
-    // in CSS, so that half stays keyed on `:first-child` and a grid with two or more row headers
-    // keeps drawing its head-row seam in the frame color, exactly as it does today — same value as
-    // before this change, not a new divergence.
+    // base cell rule already gives them this color.
+    // The HEAD row keys on `.htLastRowHeaderColumn` instead, because CSS cannot pick the last corner
+    // cell out of a head row — a corner is a `th` like the column headers beside it, and only the
+    // engine knows how many of them there are. `render/rowHeaders.ts` and `render/columnHeaders.ts`
+    // stamp that class on the last row-header column in every table. It used to key on
```
budnix: This names `render/rowHeaders.ts` as one of the stampers, but the last commit took the marker out of that file — it now carries a comment saying it is left out on purpose. The `in every table` part reads as every row too, and the class only lands on head rows. Since this sits right above the selector, it sends a reader to the wrong file first.
```suggestion
    // engine knows how many of them there are. `render/columnHeaders.ts` stamps that class on the
    // last corner cell of each head row, and only there. It used to key on
```

handsontable/src/3rdparty/walkontable/test/spec/renderer/columnHeaders.spec.js
```diff
@@ -62,12 +62,12 @@ describe('Walkontable.Renderer.ColumnHeadersRenderer', () => {
     expect(rootNode.outerHTML).toMatchHTML(`
       <thead>
         <tr>
-          <th class="">1</th>
+          <th class="htLastRowHeaderColumn">1</th>
```
budnix, line 65: Every expectation this commit updates runs with `rowHeadersCount = 1`, so the marker always lands on index 0 — the same cell `:first-child` used to pick. This suite would still pass if the check were `rowHeadersCount > 0 && visibleColumnIndex === 0`, which is the old buggy rule, so the two-row-header case the fix exists for is pinned only in the Playwright specs. `htLastVisibleHeader` already shows the pattern in this file: one spec pins it and the one at line 274 proves it moves on the next render cycle. A case with `rowHeadersCount = 2` asserting index 1, plus one that re-renders after dropping back to 1, would also be the direct proof of the no-clearing-pass claim in the renderer comment.

handsontable/src/3rdparty/walkontable/src/render/columnHeaders.ts
```diff
@@ -123,6 +123,31 @@ export class ColumnHeadersRenderer extends BaseRenderer {
         }
 
         columnHeaderFunctions[visibleRowIndex](sourceColumnIndex, TH, visibleRowIndex);
+
+        // Marks the cell that owns the head row's copy of the seam to column 0, for the theme rule
```
budnix: This block says nearly the same thing as walkontable `AGENTS.md` lines 212-232 and the comment above the selector in `_base.scss`. Three copies is what let one of them drift — the `_base.scss` version still names `render/rowHeaders.ts` as a stamper after the last commit removed it. May I suggest keeping this one to what a reader at this line needs, which cell and why not `:first-child` and why after the renderer runs, and leaving the long version in `AGENTS.md`?

handsontable/src/3rdparty/walkontable/src/render/rowHeaders.ts
```diff
@@ -109,6 +109,13 @@ export class RowHeadersRenderer extends BaseRenderer {
         }
 
         rowHeaderFunctions[visibleColumnIndex](sourceRowIndex, TH, visibleColumnIndex);
+
+        // No marker is stamped here on purpose. The theme rule that colors the seam to column 0
```
budnix: This comment is the only change in the file, and walkontable `AGENTS.md` lines 218-221 already makes the same point in almost the same words. Curious whether it needs to sit inside the loop body — that runs for every row header of every rendered row, and someone scanning the render path hits six lines about code that is not there. The method JSDoc would hold it just as well if you want to keep it.

tests/fixtures/pages/RowHeaderBorderOwnershipPage.ts
```diff
@@ -209,6 +209,60 @@ export class RowHeaderBorderOwnershipPage {
...
+   */
+  lastCornerHeaderCell(testId: string, level = 0): Locator {
+    return this.grid(testId)
+      .locator('.ht_clone_top_inline_start_corner table.htCore > thead > tr').nth(level)
+      .locator('th.htLastRowHeaderColumn');
+  }
+
+  /**
+   * How many corner cells a head row renders, counted by the engine's marker being present on
+   * exactly one of them.
```
budnix: The summary says corner cells, but this returns how many `th`s carry `htLastRowHeaderColumn` — at most one, and never the corner count. Both call sites use it that way, asserting 1 to prove the seam rule cannot color two gridlines at once. The name is fine, it is just the first line that does not match.
```suggestion
   * How many `th`s in a head row carry the engine's marker. Exactly one when the grid has row
   * headers, or the seam rule would color more than one gridline.
```

.changelogs/13399.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Fixed the gridline between the row headers and the first column drawing in the wrong color in the header row when a grid has more than one row header.",
```
budnix: In a header row there are no row headers — that gridline runs between the corner cells and the first column header. Row headers only exist in body rows, so this names the wrong pair of cells for the row the fix is about. It is also one 27-word sentence, where the doc standards ask for short ones.

```suggestion
  "title": "Fixed the wrong gridline color between the corner cells and the first column header in grids with more than one row header.",
```

handsontable/src/3rdparty/walkontable/test/spec/renderer/columnHeaders.spec.js
demtario: This was the one that mattered, and you are right that the suite was vacuous. Fixed in `87daf20`, following the `htLastVisibleHeader` pattern you pointed at: one spec asserts index 1 with two row headers, and one drops back to one row header on the next render cycle and asserts the marker moved with nothing left behind.

That second one is also the direct proof of the no-clearing-pass claim, as you suggested, so the comment no longer asserts something only prose backed.

Verified both directions rather than trusting it: substituting the old rule you named, `rowHeadersCount > 0 && visibleColumnIndex === 0`, fails exactly those two specs and nothing else, and the correct rule passes all 818.

handsontable/src/styles/base/_base.scss
demtario: Taken verbatim in `87daf20`. The drift is exactly what you describe: the body-row stamp came out in the commit before, and this copy still sent a reader to that file first.

handsontable/src/3rdparty/walkontable/src/render/columnHeaders.ts
demtario: Agreed, and your diagnosis is borne out by the `_base.scss` copy having been the one that rotted. Cut in `87daf20` to which cell, why not `:first-child`, that `rowHeadersCount === 0` needs no guard, and that it is stamped after the renderer runs, with a pointer to `AGENTS.md` for the invariant behind the missing clearing pass.

handsontable/src/3rdparty/walkontable/src/render/rowHeaders.ts
demtario: Moved to the method JSDoc in `87daf20`. You are right that the loop body is the wrong place - it runs per row header per rendered row, and the note is about code that is not there. Kept rather than deleted only because it is what stops someone completing the symmetry and re-adding the stamp; the Playwright spec now asserts the marker's ABSENCE in body rows for the same reason.

tests/fixtures/pages/RowHeaderBorderOwnershipPage.ts
demtario: Taken verbatim in `87daf20`. The summary was left over from an earlier version of the helper that did count corner cells.

.changelogs/13399.json
demtario: Taken verbatim in `ae2e7a9`. The old title was self-contradictory rather than just loose - it named the body row's pair of cells while saying "in the header row", and a header row has no row headers at all. 22 words in one sentence now, against 27.

budnix (approved):

Haxikowy (approved):

PR #13400, opened 2026-09-07
DEV-2787: Keep the grid listening for an editor's own input

demtario:
### Context

The PR fixes a custom editor losing its own keyboard shortcuts after a click into its input field.

On every `mouseup` on the document, `tableView` decides whether the focused input belongs to the grid, and it read that from the `data-hot-input` attribute alone. The grid stamps the inputs it builds itself (the text editor's textarea, the select editor's `select`, the filters and pagination controls), but an editor supplied by a user carries none: `editorFactory` does not stamp `editor.input`, and neither does the React or Angular component editor. A focused editor field therefore read as a page input, and the first disjunct of the unlisten condition called `hot.unlisten()` — which blocks every `table`-scoped shortcut context, the `editor` one included. The editor loses its own Enter, Escape and Tab.

A left press hides how bad that is: the focus scope manager re-listens on the `click` that closes the gesture, so the grid is deaf only between the two events. A **right** press ends in `contextmenu` and no `click`, so nothing re-listens and the grid stays deaf for the rest of the edit — right-click in an editor's field, and Escape no longer discards what you typed.

The fix keeps the attribute as one way to prove ownership and adds containment as the other. `#isForeignInput()` now answers the question at both call sites: an unstamped input is the page's unless it sits in the grid's own DOM while a cell is being edited (`#isWithinOpenEditorDom()`), or inside the open editor's `preventCloseElement` subtree. `isOutsideInput()` itself is unchanged — it has four other call sites, and `FocusGridManager#focusCell()` blurs the element it answers `true` for, so widening the helper would blur a component editor's field on every selection change.

The `isCellEdited()` gate is what keeps the change narrow. An `<input>` a custom renderer puts inside a TD must keep counting as the page's, otherwise the arrow keys would move the selection while they move the caret; with no editor open, nothing about that changes, and the checkbox renderer's own `setTimeout(instance.listen, 10)` is untouched and still load-bearing.

One widening over the reported case is deliberate and documented in the code: the containment test is not per-cell, so a Shift+Click that extends a selection onto a checkbox cell (the selection source `'shift'` skips `closeEditor()`) leaves the editor open while the focus moves onto the checkbox's unstamped input, which then reads as the grid's. The outcome is the same either way — before, that `mouseup` unlistened and the checkbox renderer listened again 10ms later; now the round trip is skipped.

No public API, option, hook or default changes.

### How has this been tested?

Two new real-browser cases, both verified **red** on unfixed code (bundle rebuilt without the fix) and green with it:

```bash
# the new spec: no unlisten over a click into the editor's own field, and Escape
# still discards after a right click in that field
npx playwright test --project=e2e-main e2e/editor-focused-input-listening.spec.ts --config tests/playwright.config.ts
```

The listening state is asserted by counting `afterUnlisten` calls over one named gesture, not by reading `isListening()` at the end of it — the focus scope manager re-listens on the following `click`, so an end-of-gesture read goes green against the defect. That trap is now written down in `tests/AGENTS.md`.

A React wrapper case covers the same verdict reached from `rootPortalElement`, where the wrapper's editor portal host lives:

```bash
npm run test --prefix wrappers/react-wrapper   # "should not unlisten the grid when a pointer gesture lands in the editor's own input"
```

Full local runs, all green after the change:

```bash
npm run eslint --prefix handsontable            # 0 errors
npm run typecheck --prefix handsontable
npm run test:unit --prefix handsontable         # 4479 passed / 352 suites
npm run test:e2e --prefix handsontable          # 9911 specs, 0 failures
npx playwright test --config tests/playwright.config.ts   # 5616 passed, 6 skipped (all six legs)
npm run test --prefix wrappers/react-wrapper    # 103 passed
npm run test --prefix wrappers/vue3             # 39 passed
npm run test --prefix wrappers/angular-wrapper  # 299 passed
```

Not covered, and stated rather than implied: the `closest()`-over-`contains()` choice in the new containment test is there so a field inside a shadow root nested under the grid still resolves, and no fixture exercises that shape — the existing shadow variant puts the *grid* in a shadow root, which a plain parent walk already reaches. The code comment says so, so a later switch to `contains()` is not mistaken for a safe simplification.

### Types of changes

### Related issue(s):

1. DEV-2787

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2787

---

> [!NOTE]
> **Medium Risk**
> Touches document-level focus/unlisten logic on every mouseup; containment is gated on `isCellEdited()` to preserve renderer-owned in-cell inputs, with new e2e specs guarding regressions.
> 
> **Overview**
> Fixes **custom editors losing Enter, Escape, and Tab** after the user clicks the editor’s own unstamped `<input>`.
> 
> On document `mouseup`, `tableView` used `isOutsideInput()` (`data-hot-input` only), so component/`editorFactory` fields were treated as page inputs and **`hot.unlisten()`** ran—disabling all table-scoped shortcuts including the editor context. A left click often “healed” on the following `click`; a **right click** in the field did not, so Escape could stop discarding edits.
> 
> **`#isForeignInput()`** treats an unstamped input as the grid’s when it sits inside the open editor’s `preventCloseElement` or, while a cell is edited, inside **`rootWrapperElement` / `rootPortalElement`** (`#isWithinOpenEditorDom()`). `isOutsideInput()` is unchanged elsewhere. The document `mouseup` path also caches editor-surface and grid-root checks via **`#getGridUiRoots()`**.
> 
> Adds Playwright coverage (unlisten counter + right-click Escape), checkbox/regression cases, React/Angular wrapper specs, changelog **#13400**, and docs in `AGENTS.md` / `tests/AGENTS.md` on outside-click paths and observing `unlisten()`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ffa93a0bd711de291616d53276a3b755f257e29e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2787 Stamp data-hot-input on inputs inside the React editor portal host](https://app.clickup.com/t/9015210959/DEV-2787)

handsontable/src/tableView.ts
```diff
@@ -1913,6 +1916,119 @@ class TableView {
...
+   * checkbox renderer's own `setTimeout(instance.listen)` exists to re-listen after exactly that.
+   * With no editor open, none of that changes. With one open, an input inside the grid's own DOM
+   * is the grid's, and the grid must keep listening for the editor's sake - whichever root the
+   * editor mounted into: `editorFactory` appends the container to `rootPortalElement` for
+   * `position: 'portal'` and to `rootElement` otherwise, the React wrapper's editor portal host
+   * lives in `rootPortalElement`, and the Angular adapter appends its placeholder to
+   * `rootElement`. The test is deliberately not per-cell: it does not try to prove the field
+   * belongs to the edited cell, only that it is not the page's.
+   *
+   * That is a real widening over the reported case, and the reachable shape is a Shift+Click:
```
Haxikowy: This widening is real but untested: no case covers Shift+Click onto a checkbox cell while another cell's editor is open. Add a Playwright case asserting `isListening()`/`afterUnlisten` stay stable for that gesture, reusing the counter fixture already added.

wrappers/react-wrapper/test/hotTable.spec.tsx
```diff
@@ -557,6 +558,61 @@ describe('Editor configuration using React components', () => {
     expect(hotInstance.getDataAtCell(0, 0)).toEqual('new-value');
   });
 
+  it('should not unlisten the grid when a pointer gesture lands in the editor\'s own input', async () => {
```
Haxikowy, line 561: Only React got a regression test, but Angular's component editor mounts its placeholder in `rootElement` too (`base-editor-adapter.ts`) and depends on the same containment path with nothing pinning it. Add an Angular test asserting the grid keeps listening after a click into that placeholder's input.

handsontable/src/tableView.ts
```diff
@@ -620,20 +620,23 @@ class TableView {
...
-      if (activeHTMLElement !== null && isInput(activeHTMLElement) && !isOutsideInputElement) {
+      if (activeHTMLElement !== null && isInput(activeHTMLElement) && !isForeignInputElement) {
         return;
       }
 
       const eventPath = event.composedPath();
-      const isPathThroughGridUi = eventPath.includes(rootWrapperElement ?? rootElement) ||
-        (this.hot.rootPortalElement && eventPath.includes(this.hot.rootPortalElement));
-      const isFocusLostToOutside = !this.#isFocusWithinEditorSurface(activeHTMLElement) &&
+      const isPathThroughGridUi = this.#getGridUiRoots().some(root => eventPath.includes(root));
```
budnix: `#getGridUiRoots()` builds a fresh array here on every document `mouseup`, and a second one inside `#isWithinOpenEditorDom()` whenever an editor is open. The comment a few lines up hoists `isFocusInEditorSurface` for exactly this reason, so allocating twice right after feels a bit off. Hoisting the roots into one local and handing them to `#isForeignInput()` would keep it to one array per event.

handsontable/src/tableView.ts
```diff
@@ -1913,6 +1916,119 @@ class TableView {
...
+   * The stamp is therefore treated as one of two ways to prove ownership, containment being the
+   * other, and `isOutsideInput()` itself is left alone: it has four other call sites, and
+   * `FocusGridManager#focusCell()` BLURS the element it answers `true` for, so widening the helper
+   * would blur a component editor's field on every selection change.
+   *
+   * Answering `false` also takes the early return above, which skips the `outsideClickDeselects`
+   * block further down - the same treatment the guard has always given the grid's own stamped
+   * textarea. That block was already unreachable in every case this predicate newly answers
+   * `false` for, so the deselect is unchanged rather than newly suppressed: it needs
+   * `isFocusLostToOutside`, and the focus sits either inside a grid focus root (where
```
budnix: `isForeignFocusTarget()` checks whether the element sits inside the grid with `.contains()` and `element.closest('.handsontable')`, and both of those stop at a shadow boundary. So an editor field inside a web component in the grid reads `true` there, not `false` — and that is the same shape `#isWithinOpenEditorDom()` says it picked the `closest()` helper for. The conclusion still holds, just for a different reason: the `mousedown` handler sets `#outsideClickHandled` on any press outside the grid, so this deselect pass is skipped there already. Would it read better to lead with that guard instead?

handsontable/src/tableView.ts
```diff
@@ -1913,6 +1916,119 @@ class TableView {
...
+    return closest(element, this.#getGridUiRoots()) !== null;
+  }
+
+  /**
+   * The elements that hold the grid's own UI: the root wrapper (the grid, plus whatever a plugin
+   * renders into its layout slots and overlay layer) and the portal layer (menus, dialogs, every
+   * `position: 'portal'` editor container, the React wrapper's editor portal host).
+   *
+   * Shared by the two `mouseup` tests that ask "is this the grid's own UI" of an element or an
+   * event path, so a new mount root has one place to be added. Deliberately NOT shared with
```
budnix: The same list already exists in `focusManager/grid.ts` line 134, with the same `rootWrapperElement ?? rootElement` fallback and the same `isHTMLElement` filter. So a new mount root has two places to update, not one. That copy is what `isForeignFocusTarget()` and `hasBrowserFocus()` read from, and those decide the other half of this same `mouseup` verdict, so the two drifting apart would be easy to miss. Worth either sharing one list or softening the claim here.

handsontable/src/tableView.ts
```diff
@@ -1913,6 +1916,119 @@ class TableView {
...
+   * With no editor open, none of that changes. With one open, an input inside the grid's own DOM
+   * is the grid's, and the grid must keep listening for the editor's sake - whichever root the
+   * editor mounted into: `editorFactory` appends the container to `rootPortalElement` for
+   * `position: 'portal'` and to `rootElement` otherwise, the React wrapper's editor portal host
+   * lives in `rootPortalElement`, and the Angular adapter appends its placeholder to
+   * `rootElement`. The test is deliberately not per-cell: it does not try to prove the field
+   * belongs to the edited cell, only that it is not the page's.
+   *
+   * That is a real widening over the reported case, and the reachable shape is a Shift+Click:
+   * `core.ts` skips `editorManager.closeEditor()` when the selection source is `'shift'`, so
```
budnix: `core.ts` line 1142 skips `closeEditor()` for five sources, not one — `shift`, `refresh`, `loadData`, `updateData` and `deselect`. So Shift+Click is the reachable shape you tested, but not the only one that leaves an editor open while the focus moves; a `refresh`- or `loadData`-sourced selection change does the same. The widening argument still holds for each of them, it is just narrower here than the code is.

wrappers/react-wrapper/test/hotTable.spec.tsx
```diff
@@ -557,6 +558,61 @@ describe('Editor configuration using React components', () => {
...
+
+    await act(async () => {
+      simulatePointerActivation(field);
+    });
+
+    // The precondition the case rests on: without the focus actually sitting in the field, the
+    // verdict never reaches the `isOutsideInput` branch and the case would pin nothing.
+    expect(document.activeElement).toBe(field);
+    expect(unlistenCount).toBe(0);
+    expect(hotInstance.isListening()).toBe(true);
```
budnix, line 618: `simulatePointerActivation` dispatches `mousedown`, `mouseup` and `click`, all bubbling, so this reads the state after the `click` has already had a chance to re-listen — which is the exact effect the new `tests/AGENTS.md` note in this PR describes. `expect(unlistenCount).toBe(0)` on the line above is what actually pins the fix; this line would stay green even if the `mouseup` unlistened. Same at line 597. Keeping it as a sanity check is fine, it just should not be read as the assertion carrying the case.

handsontable/src/tableView.ts
demtario: Writing this case disproved the premise it rests on, mine included, so thank you for pushing on it.

`markSource('shift')` in `selection.ts` is the row/column **shift** an insert or a remove performs. It has nothing to do with the Shift key, so `core.ts` never spares an editor on a Shift+Click, and the paragraph I had written here was wrong.

What three probes in a real browser showed:

- At the moment of the `mouseup` the focus **is** the checkbox renderer's unstamped `<input>`, but `getActiveEditor().isOpened()` is already `false`.
- The closer is the selection change itself: `EditorManager.closeEditor()` runs from the same `Selection` local hook that fires `afterSelection`.
- So `isCellEdited()` is false when the verdict runs, and the containment test is never consulted for that input. The grid unlistens exactly as it did before this change, and `checkboxRenderer`'s own `setTimeout(instance.listen, 10)` restores listening as it always did.

The widening is therefore not reachable by a pointer gesture at all: any press that moves the focus onto another cell's input also changes the selection, which closes the editor first.

Rather than drop the request, I turned it into coverage of that reason: `tests/e2e/editor-open-checkbox-focus.spec.ts` with its own fixture and page object. Two cases, one with an editor open and one without, both reading the focus and the editor state **at** `mouseup` through a capture-phase listener (neither value survives the gesture: the focus is back on `body` once the `click` has run). Both go red if an editor ever survives that selection change, which is exactly when this would need another look.

The fixture's editor column uses a bare `editorFactory` editor rather than the default one on purpose. `textEditor` also ends its edit on `focusout`, which would have closed it for a second, unrelated reason and left the cases proving less than they claim.

The JSDoc here is corrected to say all of the above. In `6dc9e67e83`.

wrappers/react-wrapper/test/hotTable.spec.tsx
demtario: Added, in `6dc9e67e83`: `wrappers/angular-wrapper/projects/hot-table/src/lib/editor/editor-input-listening.spec.ts`.

A component editor whose template is a plain `<input>`, wired through `BaseEditorAdapter` the way `base-editor-adapter.spec.ts` wires one, with the grid appended to `document.body` (a detached container never moves `document.activeElement`, which is exactly what the verdict reads). It opens the editor, focuses the field, dispatches `mousedown`/`mouseup`, and asserts `afterUnlisten` never fires.

You were right that nothing pinned that half: Angular's placeholder goes to `rootElement` (`base-editor-adapter.ts`), React's portal host to `rootPortalElement`, and only the second had a test.

Verified in both directions rather than just green: with `handsontable/src/tableView.ts` reverted to the base commit and `tmp/` rebuilt, the case fails with `unlistenCount` 1; restored and rebuilt, it passes. Full Angular suite 300 passed / 14 suites.

handsontable/src/tableView.ts
demtario: Fair, and the inconsistency with the hoist right above it is the part that convinces me. Done in `fcf99354ed`: the roots are resolved once per event into a local, and handed to `#isForeignInput()`, which passes them on to `#isWithinOpenEditorDom()`. `isPathThroughGridUi` reads the same local, so it is one array per `mouseup` now instead of one or two.

handsontable/src/tableView.ts
demtario: You are right on both counts, and I checked before changing it.

`isForeignFocusTarget()` reaches the grid through `isInternalElement()`, which is `element.closest('.handsontable')` — the native one, so it stops at a shadow boundary — plus `contains()` on the wrapper and the portal, same limitation. For a field inside a web component under the grid, all of those miss, the shadow-host-chain fallback does not match either, and the method returns `true`. So my paragraph reasoned from the opposite of what that code does, for precisely the shape the `closest()` helper elsewhere in this file exists to handle.

Rewritten to lead with the guards you point at, which need no boundary reasoning at all: a press outside the grid has already set `#outsideClickHandled` on the `mousedown` (that path does its own deselect or `destroyEditor()`), which this handler reads back as `wasOutsideClickHandled`; and a press inside the grid fails the block's own `!#isPathWithinGrid(eventPath)` test. Either way the deselect pass is skipped without the focus position entering into it.

In `fcf99354ed`.

handsontable/src/tableView.ts
demtario: Confirmed — `focusManager/grid.ts` line 134 builds the same pair with the same `rootWrapperElement ?? rootElement` fallback and the same `isHTMLElement` filter, so "one place to be added" was wrong.

I softened the claim rather than sharing the list, and said why it matters: that copy is what `isForeignFocusTarget()` and `hasBrowserFocus()` read from, and those answer the other half of this same `mouseup` verdict, so the two drifting apart would split the verdict against itself. The comment now names the other site explicitly so a new mount root is added in both.

Sharing one list across the two modules is the better end state, but it moves a `FocusGridManager` internal into a shared helper, which is more than this fix should carry. Happy to do it here if you would rather see it now, or leave it as a follow-up.

In `fcf99354ed`.

handsontable/src/tableView.ts
demtario: Correct, and the comment now names all five: `shift`, `refresh`, `loadData`, `updateData` and `deselect`.

It also says what they have in common, which is what keeps the conclusion intact: every one of them is data-driven, so none coincides with the press that would have to move the browser focus onto another cell's input in the first place. The wording is scoped to a pointer gesture now, rather than implying the tested shape is the only one the code allows.

In `fcf99354ed`.

wrappers/react-wrapper/test/hotTable.spec.tsx
demtario: Agreed, and it is the trap this PR documents in `tests/AGENTS.md`, so leaving the line unlabelled here was inconsistent of me.

Both lines are annotated in `fcf99354ed`: the one before the gesture as a starting-state check, and this one as an end-state sanity check, with the reason spelled out — `simulatePointerActivation` ends in a bubbling `click`, the focus scope manager re-listens on that, so it stays green even when the `mouseup` unlistened. `expect(unlistenCount).toBe(0)` is what carries the case, and it is the assertion that fails on unfixed code.

Kept rather than removed: it still catches an end state that no longer heals, which is a different failure from the one the counter sees.

Haxikowy (approved):

budnix (approved):

PR #13402, opened 2026-09-07
DEV-2626: Shift cell meta when a nested row is detached from its parent

budnix:
### Context

The PR fixes cell meta being left behind when a row is detached from its parent through the
**"Detach from parent"** context menu item.

`DataManager#detachFromParent()` restructures the tree by hand: it splices `__children` and fires the
row hooks itself, so it never reaches `DataMap#createRow`/`removeRow` — the only two methods that keep
`MetaManager` in step. Detaching a child moves it out of its parent's subtree to a new position, so
every row between the old and the new position shifts, but their stored cell meta (comments,
`className`, everything else) stayed on the old row indexes and ended up on the wrong cells.

This is the same symptom class and the same context menu as #7727, which DEV-2605 / #13233 fixed for
the *insert* side. Detaching needs both halves of the move, so this adds a `moveCellsMeta()` helper
next to the existing `shiftCellsMeta()`: the block is removed at its old physical index and
re-inserted at its new one. Pre-existing, not a regression.

Three details are worth a reviewer's attention:

1. **The destination is read, never computed.** `getRowIndex(element)` is called after the final
   `rewriteCache()`, rather than reusing `movedElementRowIndex`, which is derived arithmetically from
   the grandparent position. A sibling that owns descendants breaks any `parentIndex + n` formula.
2. **The raw `getRowIndex()` result is kept out of the `?? 0` fallback** that the hook arguments use.
   As a meta index, that `0` would splice meta from the top of the grid whenever the cache does not
   know the row object.
3. **A move to the same index is skipped.** Detaching the last child of a last child re-parents it
   without moving any row, and a remove plus re-insert there would blank meta that is still on the
   right cell.

### Known limitation (deliberate)

The detached block's **own** meta is reset rather than carried across. `LazyFactoryMap` has no move
primitive, and the alternative — snapshotting through `getCellMetas()` — takes visual indexes,
materializes meta for every column, and fires `afterSetCellMeta` once per cell.

This is still strictly better than the current behavior: today that meta stays on the old physical
row, which by then holds a different record, so it corrupts an innocent row's cells. After this
change it is simply blank. Preserving it needs a move primitive in `LazyFactoryMap`, which is its own
change with its own unit tests — happy to open a follow-up ticket if you agree it is worth it.

### Noticed but deliberately not touched

All of the below is **pre-existing** and none of it is changed here. It is on the record so it is not
rediscovered later, and it is why `moveCellsMeta()` reads its destination index rather than reusing
`movedElementRowIndex`.

**1. The row-creation hooks report the wrong index in the grandparent branch.** Both are computed
arithmetically, which is exactly what this ticket warned against:

- `movedElementRowIndex = grandparentRowIndex + countChildren(grandparent)`, reported as `- 2`.
- `getChild(grandparent, countChildren(grandparent) - 1)` indexes direct `__children` with a
  **recursive** descendant count, so with any grandchildren present it reads past the end,
  `getRowIndex(null) ?? 0` yields `0`, and `beforeCreateRow` fires with `1`.

On `[{G, __children: [{P, __children: [{E}]}, {S}]}]`, detaching `E` reports `afterCreateRow` at
index `0` while `E` lands at row 3. This is not inert: `UndoRedo` builds a `CreateRowAction` straight
from `afterCreateRow`, and `UndoRedo#done()` blocks only the `UndoRedo.undo`, `UndoRedo.redo` and
`auto` sources — not `nestedRows`. So Ctrl+Z after a detach can run `alter('remove_row', 0, 1)` and
take out `G` with its whole subtree. Fixing it changes public hook argument values and undo behavior,
so it needs its own ticket rather than a quiet ride along with a cell-meta fix.

**2. `getRowIndexWithinParent()` returns `-1`, never `null`/`undefined`.** The existing guard tests
for the latter two, so a `-1` passes it and `parent.__children.splice(-1, 1)` removes the *last*
child instead of the element. The new meta code deliberately mirrors that guard through a flag set
inside the branch, rather than adding a second, differently-worded copy of the condition.

**3. `childRowIndex` keeps its `?? 0` fallback**, which feeds `removedRowIndexes` and the
`beforeRemoveRow`/`afterRemoveRow` arguments. On a `null` lookup it reports the removal at rows
`0..n`, and listeners that shift their own collections from that index (`customBorders`,
`mergeCells`) move the wrong entries. Unreachable through the context menu today, since
`getDataObject()` returns a live cached object.

**4. `RowMoveController#moveCellsMeta()` mixes the two index spaces.** It reads with
`getCellMetaAtRow()` (physical) and writes with `hot.spliceCellsMeta()` (visual, and it translates
again internally), passing the same indexes to both — so the row-move path puts meta on the wrong
rows as soon as a collapsed or trimmed row makes visual and physical diverge. Same rule this PR's
`AGENTS.md` entry now spells out. Worth its own ticket; the entry documents it in the meantime so
nobody copies the pattern.

### One existing spec changed, deliberately

`Formulas > preserveTextValue > should keep preserved text values escaped after detaching a row with
the Nested Rows plugin` (from #13213) asserted the **old** behavior, and its comment said so
outright: "The `cell` markings stay with their physical coordinates, they do not follow the moved
row." That is precisely what this PR changes, so the spec had to move with it.

Its data marks physical row 3 (`parent2`) with `preserveTextValue`, detaches the child at row 1, and
the child is appended last — landing on row 3. It previously expected the child to come out escaped
as `'0123456`, because it inherited the marking stranded on the index it happened to land on. Now
the marking follows `parent2` from row 3 to row 2 and keeps its escape, while the child gets fresh
meta at row 3 and is not escaped. Escaping a value the user never marked, purely because it inherited
a vacated physical index, is the same defect class this PR fixes.

The spec still discriminates rather than being weakened: it now **fails against the base branch**
(`Expected ''0123456' to equal '0123456'`), so it is a regression test for this change instead of an
assertion of the bug. Its title is unchanged and still accurate — the marked cell's preserved text
value does stay escaped across the detach.

Worth a reviewer's eye: `cell` is re-applied at physical coordinates on the next `updateSettings`
(`core.ts:4240`), so `cell`-declared markings remain coordinate-anchored on that path while they now
follow rows on a detach. Flagging in case the team wants that reconciled separately.

### How has this been tested?

Four regression tests were added to the existing `alter.spec.js`, in a describe block next to the
DEV-2605 ones. Each marks a cell **above** the moved range and one **below** it, and asserts the one
above did not move — a shift at the wrong index moves both.

- context menu "Detach from parent" on `getSimplerNestedData()`
- a deep subtree below the detached row, on `getMoreComplexNestedData()`
- a detached row that carries its own descendants, so the block `amount` is exercised (3, not 1)
- the same-index case, which pins detail 3 above

Verified that the first three **fail without the source change** (`Expected undefined to be
'below-cell'`) and pass with it. The fourth passes either way by design — it guards the fix against
over-reaching, and its comment says so.

```bash
npm run test:e2e --prefix handsontable --testPathPattern=nestedRows   # 180 specs, 0 failures
npm run test:e2e --prefix handsontable --testPathPattern=formulas     # 420 specs, 0 failures
npm run eslint --prefix handsontable                                  # 0 errors
```

### Types of changes

### Related issue(s):

1. DEV-2626
2. Follow-up from the review on DEV-2605 / #13233 (the insert side, GitHub issue #7727)

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2626

### DIFF
--- a/.changelogs/13402.json
+++ b/.changelogs/13402.json
@@ -1,6 +1,6 @@
 {
   "issuesOrigin": "private",
-  "title": "Fixed a bug where detaching a row from its parent left cell metadata, such as comments and custom class names, on the wrong rows.",
+  "title": "Fixed a bug where detaching a row from its parent left cell metadata, such as comments and custom class names, on the wrong rows. The detached row's own metadata is now cleared instead of being left behind on another row.",
   "type": "fixed",
   "issueOrPR": 13402,
   "breaking": false,
--- a/.changelogs/13402.json
+++ b/.changelogs/13402.json
@@ -1,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Fixed a bug where detaching a row from its parent left cell metadata, such as comments and custom class names, on the wrong rows.",
+  "type": "fixed",
+  "issueOrPR": 13402,
+  "breaking": false,
+  "framework": "none"
+}
--- a/handsontable/src/plugins/formulas/__tests__/formulas.spec.js
+++ b/handsontable/src/plugins/formulas/__tests__/formulas.spec.js
@@ -4581,9 +4581,11 @@ describe('Formulas general', () => {
       ]);
       expect(formulasPlugin.engine.getSheetValues(formulasPlugin.sheetId)[1]).toEqual([123456]);
 
-      // Detaching the child moves it below its former parent's block, so it becomes physical row 3
-      // – the preserved text cell. The `cell` markings stay with their physical coordinates, they do
-      // not follow the moved row.
+      // Detaching the child shifts every row below it up by one, and since DEV-2626 the cell meta
+      // moves with them: the marking follows `parent2` from physical row 3 to row 2, so its value
+      // stays escaped. The detached child is appended last and lands on physical row 3, but it was
+      // never marked and now gets fresh meta there, so it is NOT escaped – before DEV-2626 it
+      // inherited the marking left behind on the index it happened to land on.
       getPlugin('nestedRows').dataManager.detachFromParent(
         getPlugin('nestedRows').dataManager.getDataObject(1)
       );
@@ -4592,7 +4594,7 @@ describe('Formulas general', () => {
         ['parent1'],
         ['child2'],
         ['\'parent2'],
-        ['\'0123456'],
+        ['0123456'],
       ]);
     });
 
--- a/handsontable/src/plugins/nestedRows/AGENTS.md
+++ b/handsontable/src/plugins/nestedRows/AGENTS.md
@@ -108,6 +108,24 @@ They are written in different places and can drift. Keep this in mind:
   `TypeError: Cannot read properties of undefined`, which is what forced one customer to retry inside
   `requestAnimationFrame`. Never re-introduce the non-null assertion, and never feed the result
   straight into the trimming map — filter `null` out first.
+- **The hand-built tree operations have to shift the cell meta themselves.** `MetaManager` is kept in
+  step only by `DataMap#createRow`/`removeRow`, and `addChild`, `addChildAtIndex` and
+  `detachFromParent` reach neither — they splice `__children` and fire the row hooks by hand. So
+  stored meta (comments, `className`, everything) stays on the old physical rows and lands on the
+  wrong cells: #7727 for the insert side, DEV-2626 for the detach side, same context menu. Use
+  `dataManager.shiftCellsMeta()` for an insert and `moveCellsMeta()` for a move. Both take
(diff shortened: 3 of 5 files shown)


qunabu:
Task linked: [DEV-2605 Fix cell meta not shifting when NestedRows inserts a row via the context menu](https://app.clickup.com/t/9015210959/DEV-2605)
Task linked: [DEV-2626 NestedRows: detachFromParent moves rows without shifting cell meta](https://app.clickup.com/t/9015210959/DEV-2626)

demtario (approved):

PR #13403, opened 2026-09-07
DEV-2805: Keep a merge trimmed to one visible row across a column move or freeze

demtario:
### Context

The PR fixes a merged cell being deleted for good by a column move or a column freeze when its rows are trimmed down to one visible row.

When rows inside a merged cell are trimmed (Filters, `trimRows`, collapsed nested rows), the merge re-anchors onto its visible rows and its own `rowspan` shrinks to the visible count. Trimmed to a single visible row, the merge object is `rowspan: 1`. `MergedCellsCollection#translateAfterAxisMove` drops any fragment that comes out `rowspan === 1 && colspan === 1`, unless the caller lists that fragment in `retainedIndexes`. Only the row-move handler passed that map. The two column call sites (`#onAfterColumnMove`, `#onAfterColumnFreeze`) passed nothing, so a `colspan: 1` merge in that state was dropped by any `manualColumnMove` or `manualColumnFreeze`, including one that touched none of its own columns. The collection had already been cleared by then, so the rows coming back restored nothing.

The fix adds `#planColumnMoveRetention(snapshot)`: for every merge that owns more than one row, at least one of them trimmed, all of its physical columns are retained. Both column call sites pass it. The column axis needs no trimmed-row attribution and no ownership guard, because a column reorder never changes which rows a merge covers, so every fragment inherits the whole anchor either way. A merge that owns a single row is never retained: with nothing to come back, its one-column fragment is a genuine single cell and is dropped as before.

A second, related defect surfaced while addressing review feedback. `translateAfterAxisMove` rebuilds the lookup matrix from every replacement, so a merge whose rows are all trimmed (and therefore purged from the matrix) was written back at its stale visual coordinates, over whatever physical rows now sit there, and the replacement object lost the purged flag that re-adds it once its rows return. This drew a phantom merge over foreign rows after a column move, a column freeze, or a row move. It was pre-existing for a merge trimmed in one go (`rowspan >= 2`), and the new retention would have extended it to a merge trimmed one row at a time (`rowspan: 1`). All three axis-move handlers now re-run the re-anchor after the anchors are carried over, gated on row trimming being active, which purges and re-flags every merge without a visible top-left.

The merge-cells guide bullet that documented the old behavior as a row-move-only guarantee is rewritten, and the plugin `AGENTS.md` gains the traps above.

### How has this been tested?

- Three new Jasmine specs in the trimming describe of `handsontable/src/plugins/mergeCells/__tests__/mergeCells.spec.js`, one per path: an unrelated column move, a column freeze, and a column-move split that leaves a single-column fragment. Each also asserts the full span comes back after `untrimAll()`. All three were red before the fix and are green after it.
- Three more Jasmine specs in the same describe for the phantom: a merge purged by an incremental trim across a column move, a fully trimmed merge across a column move, and a fully trimmed merge across a row move. Each asserts the merge is absent from the lookup matrix and the DOM after the move and comes back whole after `untrimAll()`. All three were red before the second commit.
- One Jasmine spec for the one-row floor: a `rowspan: 1, colspan: 3` merge whose row is trimmed, split by a column move, keeps only its two-column fragment after `untrimAll()`. Red without the floor (two merges came back), green with it.
- One new unit test in `cellsCollection.unit.ts` for `retainedIndexes` on the `'column'` axis.
- `npm run test:e2e --testPathPattern='mergeCells|manualColumnFreeze|manualColumnMove|manualRowMove|trimRows|nestedRows|hiddenRows|hiddenColumns|filters/__tests__/filters'`: 1800 specs, 0 failures.
- `npm run test:unit --testPathPattern=cellsCollection`: 45 passed.
- `npm run typecheck`, `npm run eslint`: clean.

### Types of changes

### Related issue(s):

1. DEV-2805

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2805

---

> [!NOTE]
> **Medium Risk**
> Touches mergeCells axis-move and trimming interaction paths (column move, freeze, row move); behavior is heavily spec’d but regressions could affect grids with merges plus filters/trimRows.
> 
> **Overview**
> Fixes **merged cells being permanently lost** when `trimRows`/filters leave only one visible row and the user runs **column move** or **column freeze**. Those paths now pass `#planColumnMoveRetention` into `translateAfterAxisMove`, so `1×1` fragments that still own trimmed rows are not treated as disposable singletons (same idea as row move). **One-row merges** with no extra rows to restore still drop single-column fragments when split.
> 
> Also fixes **phantom merges** after row/column reorder: `translateAfterAxisMove` could put fully trimmed merges back into the lookup matrix at stale coordinates. **Column move, column freeze, and row move** handlers now run `#reanchorMergesToVisibleRows()` again when trimming is active, after anchors are transferred.
> 
> User docs and plugin `AGENTS.md` are updated to describe **trimmed-row retention on both axes** and the post-move re-anchor. Coverage adds integration specs for column move/freeze/split, the one-row floor, phantom-matrix cases, and a unit test for column-axis `retainedIndexes`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit afcd1696d10dc411cdc49e1f8e7a39800c5d6a45. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2805 Merge cells: a column move or column freeze deletes a merge trimmed to one visible row](https://app.clickup.com/t/9015210959/DEV-2805)

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -2559,8 +2619,10 @@ export class MergeCells extends BasePlugin {
     }
 
     this.#transferAnchorsAfterAxisMove(
-      this.mergedCellsCollection.translateAfterAxisMove('column', snapshot), 'column');
+      this.mergedCellsCollection.translateAfterAxisMove(
+        'column', snapshot, this.#planColumnMoveRetention(snapshot)), 'column');
     this.#captureMergeAnchors();
+    this.#purgeInvisibleMergesAfterAxisMove();
```
Haxikowy: This new call runs unconditionally on every column move and freeze/unfreeze, even with zero row trimming active and nothing in `#purgedMerges` — unlike `#planColumnMoveRetention`, it isn't gated by `#isRowTrimmingActive()`. Same at `#onAfterColumnFreeze` (~L2732). It's a no-op when untrimmed, but it's an unmeasured extra `#reanchorMergesToVisibleRows()` pass over every merge on a frequent UI interaction — worth gating the same way.

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -2626,6 +2688,7 @@ export class MergeCells extends BasePlugin {
     this.#transferAnchorsAfterAxisMove(
       this.mergedCellsCollection.translateAfterAxisMove('row', snapshot, plan?.carriers), 'row', plan?.context);
     this.#captureMergeAnchors();
+    this.#purgeInvisibleMergesAfterAxisMove();
```
Haxikowy: This reanchor now runs twice per row move: once via the row index mapper's `cacheUpdated` hook (which can't see the post-move replacement objects, per this plugin's own AGENTS.md) and again here. Intentional and not a bug, just doubled `O(merges)` cost on the previously-working row-move path — flagging for awareness.

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -1388,6 +1388,66 @@ export class MergeCells extends BasePlugin {
...
+    const retained = new Map<MergedCellCoords, Set<number>>();
+
+    if (!this.#isRowTrimmingActive()) {
+      return retained;
+    }
+
+    snapshot.forEach((physicalColumns, merge) => {
+      const anchor = this.#mergeAnchors.get(merge);
+      const ownsTrimmedRow = anchor?.physicalRows
+        .some(physicalRow => this.hot.toVisualRow(physicalRow) === null) ?? false;
```
budnix: Can I ask whether a merge that owns only one physical row was meant to be retained here? For a `rowspan: 1, colspan: 3` merge whose single row is trimmed, `ownsTrimmedRow` is true, so a column move that splits it keeps the one-column fragment — and when the row comes back the re-anchor leaves that fragment at 1×1, in both `mergedCells` and the lookup matrix, where develop dropped it. The row path never hits this because it retains only when a still-visible carrier row exists, so a floor on the row count would line the two axes up. Would be good to verify with `{ row: 2, col: 1, rowspan: 1, colspan: 3 }`, `trimRows([2])`, `moveColumn(1, 4)`, `untrimAll()` — a later merge over that leftover cell should silently do nothing, since `add()` bails on `alreadyExists`.

```suggestion
      const anchorRows = anchor?.physicalRows ?? [];
      const ownsTrimmedRow = anchorRows.length > 1
        && anchorRows.some(physicalRow => this.hot.toVisualRow(physicalRow) === null);
```

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -1388,6 +1388,66 @@ export class MergeCells extends BasePlugin {
...
+   * once its rows return is lost with the old one. Nothing else repairs that after a column reorder,
+   * which never touches the row index mapper, so the re-anchor is run here by hand: it purges every
+   * merge without a visible top-left (every row trimmed, or the anchor column hidden), flags it, and
+   * leaves the visible ones alone, since their replacements already sit on the coordinates the
+   * re-anchor would derive.
+   *
+   * Run after {@link MergeCells#captureMergeAnchors}, so the re-anchor reads the carried anchors.
+   */
+  #purgeInvisibleMergesAfterAxisMove() {
+    this.#reanchorMergesToVisibleRows();
```
budnix: `#purgeInvisibleMergesAfterAxisMove()` only forwards to `#reanchorMergesToVisibleRows()`, which five other places in this file already call directly. So the file now has two names for one operation, and grepping for the re-anchor's call sites misses these three. The doc block above is the real value here — folding it into `#reanchorMergesToVisibleRows()` and calling that method straight from the three handlers would keep it without the extra hop.

handsontable/src/plugins/mergeCells/mergeCells.ts
```diff
@@ -1388,6 +1388,45 @@ export class MergeCells extends BasePlugin {
...
+    }
+
+    snapshot.forEach((physicalColumns, merge) => {
+      const anchor = this.#mergeAnchors.get(merge);
+      const ownsTrimmedRow = anchor?.physicalRows
+        .some(physicalRow => this.hot.toVisualRow(physicalRow) === null) ?? false;
+
+      if (ownsTrimmedRow) {
+        retained.set(merge, new Set(physicalColumns));
+      }
```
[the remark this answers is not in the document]
demtario: Confirmed and fixed in 568d2a2, and the hole was wider than the 1x1 case: `translateAfterAxisMove` rebuilds the matrix from every replacement, so any fully trimmed merge (`rowspan >= 2` too) came back at its stale coordinates after a column move, a column freeze, or a row move, and the replacement object lost the `#purgedMerges` flag. All three handlers now re-run `#reanchorMergesToVisibleRows()` after the anchors are carried over. Three specs pin it, including your incremental-trim shape.

handsontable/src/plugins/mergeCells/mergeCells.ts
demtario: Gated on `#isRowTrimmingActive()` in afcd169, on all three handlers. Without trimming nothing can be purged, so the pass is skipped.

handsontable/src/plugins/mergeCells/mergeCells.ts
demtario: Intentional, and now gated the same way as the column path. The second pass is needed because the mapper's pass ran before `afterRowMove` and saw the old objects. Without it a fully trimmed `rowspan: 2` merge stays in the matrix over foreign rows after a row move; `should keep a fully trimmed merge out of the lookup matrix across a row move and bring it back whole` was red before the call was added.

handsontable/src/plugins/mergeCells/mergeCells.ts
demtario: Right, thanks. Took the `length > 1` floor in afcd169 and added your scenario as `should drop the single-column fragment of a one-row merge whose row is trimmed when a column move splits it`. Confirmed red without the floor: two merges came back after `untrimAll()`.

handsontable/src/plugins/mergeCells/mergeCells.ts
demtario: Folded in afcd169. The three handlers call `#reanchorMergesToVisibleRows()` directly, under the trimming gate, and the explanation now lives in its JSDoc.

Haxikowy (approved):

budnix (approved):

PR #13404, opened 2026-09-07
DEV-2812: Re-measure row heights after AutoRowSize#clearCache()

budnix:
### Context

The PR fixes row headers drifting out of alignment with their rows after an application calls `AutoRowSize#clearCache()`.

`clearCache()` empties every measured row height and schedules nothing to replace them. Rows are only measured by `calculateVisibleRowsHeight()`, which is registered on the `beforeRender` hook — the full-render hook — and it measures just the visible band. Scrolling performs a *view* render, so scrolling measures nothing. Every row that was not on screen during the next full render therefore stays unmeasured for good and falls back to the default row height.

That stays invisible until the grid is scrolled sideways and a wide wrapping column comes into view. The data cell then renders at its content height, because a cell can never render shorter than its own text, while the row header — which has nothing to push it taller — honors the default height it was given. The two tables disagree, and because each row's gap adds to the ones below it, the offset grows the further down the grid you look. Measured on `develop` with the reporting client's own demo: 40px per row, 540px cumulative.

`clearCache()` now leaves the plugin owing a full recalculation, and the next render honors it the way `#onInit` already does through `calculateAllRowsHeight()`. Deferring rather than measuring inside `clearCache()` keeps repeated calls down to a single recalculation and leaves `clearCache()` as cheap as it has always been.

Four rules make that safe, and each of them has a concrete failure behind it:

- **The full sweep runs after `calculateVisibleRowsHeight()`, never instead of it.** `calculateAllRowsHeight()` measures only up to `syncLimit` rows synchronously and hands the rest to an idle sweep that asks for no redraw, so on a grid scrolled past that limit the band on screen would draw unmeasured — re-creating the defect on the frame meant to repair it.
- **The flag is held, not consumed, when the render cannot measure.** `recalculateAllRowsHeight()` is a no-op on a hidden grid, and a column-less grid still reports itself visible — the sweep would then write a near-empty height for every row that nothing later corrects. `calculateVisibleRowsHeight()` already bails out on that grid for the same reason.
- **The flag is consumed before the sweep and re-owed if it throws.** The ghost table runs the real renderers. Holding it across the call would let a re-entrant render recurse; spending it outright would leave every unmeasured row at the default height for the instance's life.
- **`measuredRows` is left alone.** The public `isNeedRecalculate()` slices the height map by it, so zeroing it makes that method answer "nothing to recalculate" at the exact moment every height was dropped. `AutoColumnSize#clearCache()` leaves its counterpart alone too.

The selective forms — `clearCache([rows])` and `clearCacheByRange()` — had the identical below-the-fold hole and now close it differently: they queue exactly the rows they cleared onto the refresh queue, which `#calculateSpecificRowsHeight()` reads from the data, so only those rows are re-measured rather than the whole grid.

`calculateAllRowsHeight()` also cancels any sweep still in flight before starting its own. Two running together doubled the work and raced on `inProgress` — whichever finished first cleared it while the other was still writing heights, and `#onBeforeRender` reads that flag to decide whether the refresh queue is safe to drain.

Two notes for reviewers, both of which contradict the original report in DEV-2718:

1. **This is not a regression.** 16.2.0, 17.0.0, 18.0.0 and `develop` all break identically. The bisection table in DEV-2718 does not hold up — it appears to have measured the frame right after a scroll, before the repaint, which is briefly out of step on every version.
2. **The trigger is horizontal scrolling, not vertical**, and the minimal 3-column repro attached to DEV-2718 does not reproduce on any version: it has no horizontal scroll, so every row is measured correctly.

`renderAllColumns: true` does not help, which matches what the client reported.

### How has this been tested?

**New Playwright E2E** in `tests/e2e/auto-row-size-clear-cache.spec.ts`, on a fixture whose grid is narrower than its columns so the wrapping columns start off-screen. Five tests: the row headers stay aligned after a full wipe; no rendered row is left unmeasured; the two selective forms re-measure rows they cleared below the fold; and — as a control that the added render-path work does not break the ordinary case — a grid whose cache was never touched still lines up.

**New unit tests** in `handsontable/src/plugins/autoRowSize/__tests__/rowHeightsInteraction.unit.ts` for the parts that are logic rather than geometry: `isNeedRecalculate()` still reports work owed after a wipe, a column-less render does not spend the scheduled recalculation, and a sweep that throws re-owes it.

Every one of these was confirmed to **fail on the unfixed build for the right reason** and pass on the fixed one — the E2E at 519px drift and 12 unmeasured rows, the selective-form tests at 5 unmeasured rows, the column-less guard and the throw case by their own assertions. The below-the-fold behavior is deliberately covered in E2E rather than unit: jsdom reports no layout, so every row falls inside the rendered band and there is no fold to test against.

```bash
# new E2E coverage (30 = 5 tests x 6 theme/bundle projects)
cd tests && npx playwright test auto-row-size-clear-cache

# regression surfaces for the plugin and its neighbours
npm run test:e2e --prefix handsontable -- --testPathPattern=autoRowSize          # 55 specs, 0 failures
npm run test:e2e --prefix handsontable -- --testPathPattern=manualRowResize      # 75 specs, 0 failures
npm run test:e2e --prefix handsontable -- --testPathPattern=autoRowHeaderSize    # 5 specs, 0 failures
npm run test:unit --prefix handsontable -- --testPathPattern=rowHeightsInteraction  # 8 passed
npm run test:unit --prefix handsontable -- --testPathPattern=ghostTable             # 22 passed

npm run test:types --prefix handsontable  # clean
```

Also verified by hand against the client's own demo from the DEV-2718 attachment: the worst row-header drift goes from 540px to 0px, at every horizontal scroll position.

### Types of changes

### Related issue(s):

1. DEV-2812
2. DEV-2718 — the client report this was diagnosed from

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2812

---

> [!NOTE]
> **Medium Risk**
> Changes the render and async measurement pipeline for a core sizing plugin; mistakes could affect scrollbars, overlays, or performance on large grids after cache clears.
> 
> **Overview**
> Fixes **row headers drifting away from data rows** after `AutoRowSize#clearCache()` (and the selective `clearCache(rows)` / `clearCacheByRange()` APIs). Clearing heights used to leave off-screen rows at the default height while only the visible band was re-measured on render, so wrapped cells and row headers could disagree once wide columns scrolled into view.
> 
> A **full** `clearCache()` now sets `#fullRecalculationScheduled`; the next full render runs the visible-band pass first, then `recalculateAllRowsHeight()` (with guards for hidden, column-less, or empty grids, and re-owing the flag if measurement throws). **`measuredRows` is not reset** so `isNeedRecalculate()` stays truthful.
> 
> **Selective** clears enqueue affected rows on the existing refresh queue via `#drainRowRefreshQueue()`, which also waits while a sweep is in flight or there are no columns. Sweeps are **single-flight** (`#idleSweepTimer` cancellation), with `#abandonSweep()` resetting `inProgress`, cleaning the ghost table, and scheduling retry on renderer failures.
> 
> Adds **unit tests** for these edge cases and **Playwright E2E** for alignment and below-the-fold re-measurement in a real layout.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4364cf9ed6121649ab4915fd1bc975c5bfb831bd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/handsontable/src/plugins/autoRowSize/autoRowSize.ts
+++ b/handsontable/src/plugins/autoRowSize/autoRowSize.ts
@@ -558,10 +558,19 @@ export class AutoRowSize extends BasePlugin {
         return;
       }
 
-      this.calculateRowsHeight({
-        from: current,
-        to: Math.min(current + AutoRowSize.CALCULATION_STEP, length)
-      }, colRange, overwriteCache);
+      try {
+        this.calculateRowsHeight({
+          from: current,
+          to: Math.min(current + AutoRowSize.CALCULATION_STEP, length)
+        }, colRange, overwriteCache);
+      } catch (error) {
+        // This runs inside an idle task, so the throw escapes whatever called
+        // `calculateAllRowsHeight()` - no caller's try/catch can see it. Leave the plugin
+        // recoverable before it goes.
+        this.#abandonSweep();
+
+        throw error;
+      }
 
       current = current + AutoRowSize.CALCULATION_STEP + 1;
 
@@ -586,7 +595,14 @@ export class AutoRowSize extends BasePlugin {
 
     // sync
     if (syncLimit >= 0) {
-      this.calculateRowsHeight({ from: 0, to: syncLimit }, colRange, overwriteCache);
+      try {
+        this.calculateRowsHeight({ from: 0, to: syncLimit }, colRange, overwriteCache);
+      } catch (error) {
+        this.#abandonSweep();
+
+        throw error;
+      }
+
       current = syncLimit + 1;
     }
     // async
@@ -594,8 +610,29 @@ export class AutoRowSize extends BasePlugin {
       loop();
     } else {
       this.inProgress = false;
+
+      // The whole grid fitted inside the sync limit, so `loop()` never ran and its completion
+      // branch - the other place the queue is drained - was never reached.
+      this.#drainRowRefreshQueue();
+
       this.hot.view.adjustElementsSize();
     }
+  }
+
+  /**
+   * Puts the plugin back into a state the next render can recover from, after a measurement threw
+   * part way through a sweep.
+   *
+   * Without it `inProgress` stays `true` for the instance's life, which does more than leave the
+   * heights half measured: `#drainRowRefreshQueue()` refuses to run while a sweep is in flight, so
+   * the refresh queue is silently disabled too. The full recalculation is re-owed, so the next
+   * render starts over.
+   */
+  #abandonSweep(): void {
+    cancelIdleTask(this.#idleSweepTimer);
+    this.#idleSweepTimer = 0;
+    this.inProgress = false;
+    this.#fullRecalculationScheduled = true;
   }
 
   /**
@@ -847,7 +884,10 @@ export class AutoRowSize extends BasePlugin {
       const visualRow = this.hot.toVisualRow(physicalRow);
 
       // A row outside the dataset, or one a trimming map hides, has no visual index to measure.
-      if (visualRow !== null) {
+      // Duplicates are skipped, the way the other producers of this queue do it: while the queue is
+      // held back (a column-less grid, or a sweep in flight) overlapping calls would otherwise pile
+      // the same row up and measure it once per copy.
+      if (visualRow !== null && !this.#visualRowsToRefresh.includes(visualRow)) {
         this.#visualRowsToRefresh.push(visualRow);
       }
     });
@@ -910,7 +950,9 @@ export class AutoRowSize extends BasePlugin {
     // The flag is held rather than consumed whenever this render cannot measure: while the grid is
     // hidden `recalculateAllRowsHeight()` is a no-op, and with no columns the measurement writes a
     // near-empty height for every row that nothing would ever correct (the same reason
-    // `calculateVisibleRowsHeight()` bails out on a column-less grid).
+    // `calculateVisibleRowsHeight()` bails out on a column-less grid). The row count is checked for
+    // the same shape of reason, though nothing is at stake: a sweep over no rows measures nothing,
+    // so holding the flag just keeps the work owed until there is something to measure.
     if (this.#fullRecalculationScheduled &&
         this.hot.countCols() > 0 &&
         this.hot.countRows() > 0 &&
--- a/handsontable/src/plugins/autoRowSize/autoRowSize.ts
+++ b/handsontable/src/plugins/autoRowSize/autoRowSize.ts
@@ -334,13 +334,17 @@ export class AutoRowSize extends BasePlugin {
    *
    * Only a render measures rows, and it measures just the visible band, so an emptied cache leaves
    * every row below the fold unmeasured for good. Such a row falls back to the default height,
-   * which its own cell then refuses to honour once a wide wrapping column scrolls into view - a
+   * which its own cell then refuses to honor once a wide wrapping column scrolls into view - a
    * cell never renders shorter than its text - while the row header, having nothing to push it
-   * taller, does honour it. The two tables drift apart from there.
+   * taller, does honor it. The two tables drift apart from there.
    *
    * The flag makes the next render restore the measurements the way `#onInit` does. Deferring
    * rather than measuring inside `clearCache()` keeps repeated calls down to one recalculation and
    * leaves `clearCache()` as cheap as it has always been.
+   *
+   * "The next render" means a full one: `beforeRender` is raised by `TableView#render()`, not by
+   * the engine's scroll draw, so a caller that clears the cache and then only scrolls gets nothing.
+   * Every documented use of `clearCache()` pairs it with a redraw.
    *
    * @type {boolean}
    */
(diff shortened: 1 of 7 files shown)


qunabu:
Task linked: [DEV-2718 Row header (inline-start overlay) body row heights out of sync with master table — regression since v17.0.0](https://app.clickup.com/t/9015210959/DEV-2718)
Task linked: [DEV-2812 AutoRowSize#clearCache() leaves rows unmeasured, so row headers drift out of sync](https://app.clickup.com/t/9015210959/DEV-2812)

handsontable/src/plugins/autoRowSize/autoRowSize.ts
```diff
@@ -781,10 +899,35 @@ export class AutoRowSize extends BasePlugin {
...
+        this.hot.countCols() > 0 &&
+        this.hot.countRows() > 0 &&
+        this.hot.view.isVisible()) {
+      // Consumed before the call, not after: the sweep resizes the overlays, and a re-entrant
+      // render reaching this branch with the flag still set would recurse. It is re-owed if the
+      // sweep throws - the ghost table runs the real renderers, so a renderer that throws would
+      // otherwise leave every unmeasured row at the default height for the instance's life.
+      this.#fullRecalculationScheduled = false;
+
+      try {
```
demtario, line 973: **Medium.** This try/catch only guards the *synchronous* call to `recalculateAllRowsHeight()`. A renderer that throws during the idle-scheduled continuation (`loop()` re-invoked via `requestIdleTask` — a real macrotask, outside this try/catch) is uncaught: `#fullRecalculationScheduled` stays `false` and `inProgress` stays `true` forever, silently reintroducing the exact below-the-fold defect this PR fixes, on any grid larger than `syncLimit`. The new unit test (`rowHeightsInteraction.unit.ts`, "should still owe the full recalculation after a render whose measurement threw") mocks `recalculateAllRowsHeight()` with a synchronous throw and never reaches the real async path, so it can't catch this gap.

Fix: catch inside `loop()` itself, resetting `inProgress`/`#idleSweepTimer` and re-setting `#fullRecalculationScheduled` on failure - or at minimum narrow the comment/AGENTS.md wording to state the guarantee covers only the synchronous phase.

handsontable/src/plugins/autoRowSize/autoRowSize.ts
```diff
@@ -715,24 +792,65 @@ export class AutoRowSize extends BasePlugin {
...
+   *
+   * @param {number[]} physicalRows Physical row indexes whose heights were cleared.
+   */
+  #queueClearedRowsForRefresh(physicalRows: number[]): void {
+    physicalRows.forEach((physicalRow) => {
+      const visualRow = this.hot.toVisualRow(physicalRow);
+
+      // A row outside the dataset, or one a trimming map hides, has no visual index to measure.
+      if (visualRow !== null) {
+        this.#visualRowsToRefresh.push(visualRow);
```
demtario, line 898: **Low.** Unlike its sibling queue producers - `#onBeforeChange` collects into a `Set`, `#onAfterFormulasValuesUpdate` checks `indexOf` before pushing - `#queueClearedRowsForRefresh` pushes every physical row unconditionally. Overlapping `clearCache`/`clearCacheByRange` calls landing while the queue is held back (column-less grid, or a sweep in flight) accumulate duplicate visual-row entries in `#visualRowsToRefresh`. Not wrong (the ghost table tolerates duplicates and just re-measures/re-writes the same height), just wasted work and inconsistent with the established pattern. Consider deduping here too.

handsontable/src/plugins/autoRowSize/autoRowSize.ts
```diff
@@ -781,10 +899,35 @@ export class AutoRowSize extends BasePlugin {
...
+    // synchronously and leaves the rest to an idle sweep, so on a grid scrolled past that limit the
+    // band on screen would otherwise draw unmeasured, which is the very defect this repairs.
+    //
+    // The flag is held rather than consumed whenever this render cannot measure: while the grid is
+    // hidden `recalculateAllRowsHeight()` is a no-op, and with no columns the measurement writes a
+    // near-empty height for every row that nothing would ever correct (the same reason
+    // `calculateVisibleRowsHeight()` bails out on a column-less grid).
+    if (this.#fullRecalculationScheduled &&
+        this.hot.countCols() > 0 &&
+        this.hot.countRows() > 0 &&
```
demtario, line 965: **Nit.** This guard has a fourth condition, `this.hot.countRows() > 0`, that neither the comment above nor the matching AGENTS.md bullet documents (both only explain the hidden-grid and column-less cases). Harmless - `calculateAllRowsHeight()` is a no-op on 0 rows regardless - but worth a clause since this PR is otherwise thorough about documenting every guard's rationale.

demtario (commented):
3 inline notes below.

**Off-diff:** `handsontable/src/plugins/autoRowSize/autoRowSize.ts:597` — when a grid's whole row count fits within `syncLimit`, `calculateAllRowsHeight()` never calls `loop()`, so its top-level (pre-existing, unchanged) else branch skips `#drainRowRefreshQueue()` — unlike `loop()`'s own completion branch, which this PR made call it. Practically benign here (`overwriteCache=true` already remeasures those rows), but it breaks `#drainRowRefreshQueue()`'s own documented "called from the render, and again when a sweep finishes" contract.

handsontable/src/plugins/autoRowSize/autoRowSize.ts
budnix: Good spot — both phases now recover through `#abandonSweep()` in `f405498bbc`, and the top-level else branch you flagged off-diff drains the queue there too. The new test drives the idle continuation and does fail without it.

handsontable/src/plugins/autoRowSize/autoRowSize.ts
budnix: Agreed — deduped in `f405498bbc`, it now skips a row that is already in the queue.

handsontable/src/plugins/autoRowSize/autoRowSize.ts
budnix: Fair point — documented in `f405498bbc`, in the comment and in the matching AGENTS.md bullet.

demtario (approved):

PR #13409, opened 2026-09-07
PRO-370: Add the SheetsBar plugin

Haxikowy:
### Context

The PR adds the `SheetsBar` plugin — a tab bar below the grid for switching between the sheets of a multi-sheet workbook. Each sheet owns its data, optional settings overrides, and its runtime view state (scroll, selection, sort, filters, hidden/trimmed indexes, manual sizes, merges, borders, tracked cell meta), captured on the way out and restored on the way back. Sheets cooperate with the `Formulas` plugin: with one shared HyperFormula instance and a `sheetName` per sheet, cross-sheet references such as `=Rates!B1` work, the plugin registers every sheet in the engine up front, and a tab rename renames the engine sheet and rewrites the references to it.

The bar registers a focus scope with its own shortcut context (Enter/Space activate the focused tab, arrows and Home/End walk the strip, Escape abandons a drag), drags tabs with pointer capture and FLIP animation, pages an overflowing strip, and offers a per-tab menu and a checkable all-sheets menu built on the shared `Menu` component. All visible strings are translated in 21 languages; ARIA follows the `ariaTags` setting; the rename editor is IME-safe with a grapheme-cluster length limit. Ships with the grid-level `sheetsBar` option, twelve hooks, a public API, `--ht-sheets-bar-tab-*` design tokens and generated icons for all three themes, a guide with examples for all four frameworks, and a live cross-sheet formulas demo.

Stacked on #13410, which carries the plugin-lifecycle fixes (Comments re-enable, tracked hooks for Formulas/Comments/Loading) this plugin's per-sheet settings switching depends on — merge that one first. A docs-only commit here also gives fenced code blocks the same block margin as the embedded examples.

One caveat for reviewers: the theme generator could not be run locally (`scripts/themes/figma/tokens.json` is not on this machine), so the generated CSS was reproduced with the generator's own code. Someone with the Figma export should run `npm run generate:themes` once and confirm an empty diff.

### How has this been tested?

- Unit: `npm run test:unit --prefix handsontable` — full suite green, including 156 SheetsBar tests (view-state restore ordering, settings baseline, formulas seeding/rename/duplicate, drag flicker regression, a11y attributes, IME/grapheme name limits) and hook-release tests for Formulas/Comments/Loading.
- E2E: `cd tests && npx playwright test sheets-bar --project=e2e-main` — 47 specs (keyboard model, menus, drag with pointer capture, paging, RTL, touch drag).
[...]

PR #13410, opened 2026-09-07
PRO-370: Let plugins disable cleanly: Comments, Formulas, and Loading

Haxikowy:
### Context

The PR fixes three plugin-lifecycle defects found while building the SheetsBar plugin (stacked under #13409, which needs them for its per-sheet settings switching):

- The `Comments` plugin threw `The "plugin:comments" context is already registered` when enabled a second time, and disabling it while its editor had the focus left an empty shortcut context holding the keyboard until a click. It now reuses the context, clears only its own shortcut group, hides the editor, and hands the grid context back.
- `Formulas`, `Comments`, and `Loading` registered hooks in `enablePlugin()` through the untracked `hot.addHook()`, so the hooks kept firing after the plugin was disabled — a disabled `Formulas` kept calling into a dead engine (`getSheetWidth` errors). They register tracked hooks now, and a new lint rule (`handsontable/require-tracked-hook-in-enable`, with autofix) keeps it that way.
- The hook-release test this adds caught `Loading` registering its `afterDialogFocus` hook only on the first enable, so any `updateSettings()` call silently dropped it; the hook registers on every enable now.

### How has this been tested?

- `npm run test:unit --prefix handsontable --testPathPattern="comments|hooksReleasedOnDisable|formulas|loading"` — including a test that counts each plugin's hooks before and after a disable, and Comments re-enable/keyboard-handback tests. Mutation-checked: each test fails with its fix reverted.
- `npm run test:eslint-rules` in `handsontable/` — RuleTester coverage for the new rule (runs in the Lint core CI job; the rule test imports `eslint`, so it stays out of the dependency-free root `test:tooling` job).

### Types of changes

### Related issue(s):

1. PRO-370

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

---

> [!NOTE]
> **Medium Risk**
> Touches core plugin enable/disable paths (Comments shortcuts, Formulas engine sync hooks, Loading/dialog focus); regressions would show as duplicate hooks, broken re-enable, or stale listeners after disable.
> 
> **Overview**
> Fixes plugin lifecycle bugs when toggling **Comments**, **Formulas**, or **Loading** off and on via `updateSettings()`.
> 
> **Comments** no longer throws on a second enable (`getOrCreateContext` for `plugin:comments`), duplicate shortcuts after re-enable are avoided by clearing the plugin shortcut group on disable, focus returns to `grid` when disabled with the editor open, and hook/local-listener registration is split correctly between “every enable” (`afterSetTheme` via tracked `this.addHook`) vs “create once” (`DisplaySwitch` local hooks).
> 
> **Formulas** and **Loading** (and related **Comments** hooks) now register instance hooks through **`this.addHook()`** so `disablePlugin()` tears them down; **Loading**’s `afterDialogFocus` is re-registered on each enable instead of being lost after the first settings update.
> 
> Adds ESLint rule **`handsontable/require-tracked-hook-in-enable`**, unit tests for hook counts and Comments lifecycle, and changelog **#13410**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b1b5483c92d7300242d0ea87ab79dd30a994d9c6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/handsontable/.config/plugin/eslint/rules/require-tracked-hook-in-enable.js
+++ b/handsontable/.config/plugin/eslint/rules/require-tracked-hook-in-enable.js
@@ -8,7 +8,7 @@ module.exports = {
       recommended: false,
     },
 
-    fixable: 'code',
+    hasSuggestions: true,
 
     messages: {
       useTrackedAddHook: 'Use this.addHook() rather than this.hot.addHook() inside enablePlugin(). ' +
@@ -16,6 +16,7 @@ module.exports = {
         'instance keeps running after the plugin is disabled — usually against state the plugin ' +
         'has already torn down. Register it in the constructor or init() when it genuinely has to ' +
         'outlive a disable.',
+      replaceWithTrackedAddHook: 'Replace this.hot.addHook() with the tracked this.addHook().',
     },
 
     schema: [],
@@ -62,9 +63,12 @@ module.exports = {
         context.report({
           node,
           messageId: 'useTrackedAddHook',
-          fix(fixer) {
-            return fixer.replaceText(callee.object, 'this');
-          },
+          suggest: [{
+            messageId: 'replaceWithTrackedAddHook',
+            fix(fixer) {
+              return fixer.replaceText(callee.object, 'this');
+            },
+          }],
         });
       },
     };
--- a/handsontable/.config/plugin/eslint/rules/require-tracked-hook-in-enable.js
+++ b/handsontable/.config/plugin/eslint/rules/require-tracked-hook-in-enable.js
@@ -1,0 +1,72 @@
+module.exports = {
+  meta: {
+    type: 'problem',
+
+    docs: {
+      description: 'Requires plugins to register hooks through the tracked BasePlugin#addHook inside enablePlugin',
+      category: 'Custom',
+      recommended: false,
+    },
+
+    fixable: 'code',
+
+    messages: {
+      useTrackedAddHook: 'Use this.addHook() rather than this.hot.addHook() inside enablePlugin(). ' +
+        'BasePlugin only removes the hooks it registered itself, so a hook added straight to the ' +
+        'instance keeps running after the plugin is disabled — usually against state the plugin ' +
+        'has already torn down. Register it in the constructor or init() when it genuinely has to ' +
+        'outlive a disable.',
+    },
+
+    schema: [],
+  },
+
+  create(context) {
+    /**
+     * Reports whether a node sits inside the body of an `enablePlugin` method.
+     *
+     * @param {object} node The node to test.
+     * @returns {boolean} `true` when an `enablePlugin` method encloses it.
+     */
+    function isWithinEnablePlugin(node) {
+      let current = node.parent;
+
+      while (current) {
+        if (current.type === 'MethodDefinition' && current.key && current.key.name === 'enablePlugin') {
+          return true;
+        }
+
+        current = current.parent;
+      }
+
+      return false;
+    }
+
+    return {
+      CallExpression(node) {
+        const { callee } = node;
+
+        if (callee.type !== 'MemberExpression' || callee.property.name !== 'addHook') {
+          return;
+        }
+
+        const target = callee.object;
+        const isInstanceCall = target.type === 'MemberExpression'
+          && target.object.type === 'ThisExpression'
+          && target.property.name === 'hot';
+
+        if (!isInstanceCall || !isWithinEnablePlugin(node)) {
+          return;
+        }
+
+        context.report({
+          node,
+          messageId: 'useTrackedAddHook',
+          fix(fixer) {
+            return fixer.replaceText(callee.object, 'this');
+          },
+        });
+      },
+    };
+  },
+};
--- a/handsontable/AGENTS.md
+++ b/handsontable/AGENTS.md
@@ -17,11 +17,13 @@ This is the core data grid package. **TypeScript** - source files in `src/` are 
 - No direct cross-plugin imports - use hooks for inter-plugin communication, or `hot.getPlugin('Name')` if API access is required
 - Never use raw `setTimeout` - use `this.hot._registerTimeout(fn, delay)` instead; it auto-clears on `hot.destroy()`, preventing memory leaks
 - DRY: reuse existing helpers and mixins; if code repeats, extract a generic helper rather than duplicating
+- Plugin file doing more than two things → split it (see `src/plugins/sheetsBar/` for the sliced layout)
 - Method ordering: public methods first, then private listeners
 - In text and comments, always write `Handsontable`, never `HOT` (a `hot` variable holding an instance is fine)
 - Never call `String.prototype.toLocaleLowerCase`/`toLocaleUpperCase` directly — use `localeLowerCase()` from `helpers/string` (faster, locale-correct, crash-safe). Enforced by `no-restricted-syntax`. See `.ai/CONVENTIONS.md`.
 - Never call a JavaScript method newer than `../browser-targets.js` (Chrome >= 130, Edge >= 130, Firefox >= 132, Safari >= 18.2, iOS >= 18.2 — the pinned Baseline 2024 set). swc lowers syntax only and adds no core-js polyfills, so such a call throws `X is not a function` on a supported browser. Two gates catch this: `tsconfig.json` pins `lib` to `ES_TARGET` (`es2024`) from `../browser-targets.js`, so an above-floor built-in is a type error in `.ts` files; `no-restricted-syntax` covers the same methods plus the 12 remaining `.js` files in `src/` (`allowJs` is off). That rule bans one name, `with` → `arr.slice()` plus an index assignment, and it is above the floor on two engines at once: `core-js-compat` puts `Array#with` at Firefox 140 (against our 132) and `TypedArray#with` at Safari/iOS 26.0 (against our 18.2). Verify any new method's floor against `core-js-compat`'s `data.json` before using it, and add it to the rule when it sits above the floor.
 - `throwWithCause()` assigns `cause` after construction instead of passing the constructor options bag. The options-bag overload is ES2022 (Chrome 94 / Firefox 91 / Safari 15.0) and now sits well inside the floor, so this is no longer a compatibility requirement — it is kept because an engine that accepts the options bag and drops it silently leaves `error.cause` undefined and breaks `error.cause?.handsontable === true` detection with no visible failure. Do not "simplify" it back to `new Error(msg, { cause })` without adding a test that would catch that.
+- `IndexMapper#getIndexesSequence()` returns the mapper's live internal array, not a copy — always copy it with `.slice()` before storing or caching it (`src/plugins/sheetsBar/viewState.ts`), or a later reorder silently corrupts the stored snapshot.
 
 ## Plugin Lifecycle
 
@@ -172,6 +174,7 @@ Source `.ts` files (`src/**/*.ts`, excluding walkontable and test/type files) an
 | `window.scrollTo(...)` | `this.hot.rootWindow.scrollTo(...)` |
 | `document.querySelector(...)` | `this.hot.rootDocument.querySelector(...)` |
 | `console.warn(...)` | `import { warn } from 'helpers/console'; warn(...);` |
+| `this.hot.addHook(...)` inside `enablePlugin()` | `this.addHook(...)` — tracked hooks are removed by `disablePlugin()`; enforced by `handsontable/require-tracked-hook-in-enable` (lexical match only; aliases/helpers are not caught). Rule tests: `npm run test:eslint-rules` in `handsontable/` (run by the Lint core CI job — RuleTester imports `eslint`, so the dependency-free root `test:tooling` job cannot run them) |
 | Missing JSDoc comment (`jsdoc/require-jsdoc`) on a class/method/field/function | Add a multiline block above it with a blank line before `/**` and after `*/` — `/**` on its own line, then ` * Description.`, then ` */`; no `@private` tag on `#`-fields; no `@param`/`@returns` in `.ts` files |
 
 ## Build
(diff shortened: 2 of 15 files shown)


qunabu:
Task linked: [PRO-370 [Tabs Bar] Sheets switcher](https://app.clickup.com/t/9015210959/PRO-370)

handsontable/src/plugins/comments/comments.ts
```diff
@@ -317,7 +317,7 @@ export class Comments extends BasePlugin {
       this.#editor = new CommentEditor(this.hot.rootDocument, this.hot.isRtl(), this.hot.rootPortalElement);
       this.#editor?.addLocalHook('resize',
         (width: number, height: number) => this.#onEditorResize(width, height));
-      this.hot.addHook('afterSetTheme', (themeName: string, firstRun: boolean) => {
+      this.addHook('afterSetTheme', (themeName: string, firstRun: boolean) => {
```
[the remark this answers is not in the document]
Haxikowy: Confirmed and fixed — the hook now registers on every enable, outside the editor's first-create guard. Regression test added (fails on the guarded version): keeps hiding the editor on theme changes after a disable/enable round trip.

handsontable/src/plugins/comments/comments.ts
```diff
@@ -317,13 +317,17 @@ export class Comments extends BasePlugin {
...
 
+    // Registered on every enable, not only when the editor is first built: tracked hooks are
+    // removed by disablePlugin() while the editor instance survives it, so a hook inside the
+    // first-create guard would be gone for good after one disable/enable round trip.
+    this.addHook('afterSetTheme', (themeName: string, firstRun: boolean) => {
+      if (!firstRun) {
+        this.hide();
+      }
+    });
+
```
budnix, line 330: The `if (!this.#displaySwitch)` block just below creates the switch once, but the two `#displaySwitch?.addLocalHook(...)` calls on lines 346-347 sit outside it, and `addLocalHook` in `mixins/localHooks.ts` pushes without checking for a duplicate. The switch survives a disable and is only torn down in `destroy()`, so every disable/enable round trip leaves one more `hide` and `show` handler behind — after five toggles a single hover runs `showAtCell()` five times. It is the same pile-up the new comment in `unregisterShortcuts()` describes for the shortcut context, and the editor's `resize` hook already avoids it by sitting inside its own create guard. Worth a quick check whether those two lines can move inside the guard too.

handsontable/AGENTS.md
```diff
@@ -17,11 +17,13 @@ This is the core data grid package. **TypeScript** - source files in `src/` are
 - No direct cross-plugin imports - use hooks for inter-plugin communication, or `hot.getPlugin('Name')` if API access is required
 - Never use raw `setTimeout` - use `this.hot._registerTimeout(fn, delay)` instead; it auto-clears on `hot.destroy()`, preventing memory leaks
 - DRY: reuse existing helpers and mixins; if code repeats, extract a generic helper rather than duplicating
+- Plugin file doing more than two things → split it (see `src/plugins/sheetsBar/` for the sliced layout)
```
budnix: `src/plugins/sheetsBar/` does not exist on this branch or on develop — a repo-wide grep finds it only as a fake path inside a pre-push test fixture. Line 26 points at `src/plugins/sheetsBar/viewState.ts` the same way. Since #13409 is the PR that adds the plugin and it sits on top of this one, this merges first and leaves two dead paths in the guide. Would it make sense to move both lines over to #13409?

handsontable/src/plugins/__tests__/hooksReleasedOnDisable.unit.js
```diff
@@ -0,0 +1,62 @@
...
+
+    const before = hookCount(hookName);
+
+    hot.updateSettings(settings);
+
+    expect(hookCount(hookName)).toBeGreaterThan(before);
+
+    hot.updateSettings({ [pluginKey]: false });
+
+    expect(hookCount(hookName)).toBe(before);
```
budnix, line 60: This stops at the disable, so it catches a hook that is never removed but not a hook that never comes back. A hook left inside a first-create guard and switched to `this.addHook` passes here and still disappears for good on the second enable — which is exactly what got caught in Comments. Comments has its own round-trip test now, but Loading and Formulas do not, and one more enable at the end would cover all three.
```suggestion
    expect(hookCount(hookName)).toBe(before);

    hot.updateSettings(settings);

    expect(hookCount(hookName)).toBeGreaterThan(before);
```

handsontable/.config/plugin/eslint/rules/require-tracked-hook-in-enable.js
```diff
@@ -0,0 +1,72 @@
...
+  meta: {
+    type: 'problem',
+
+    docs: {
+      description: 'Requires plugins to register hooks through the tracked BasePlugin#addHook inside enablePlugin',
+      category: 'Custom',
+      recommended: false,
+    },
+
+    fixable: 'code',
```
budnix: An autofix here changes when the hook is torn down, and `npm run eslint:fix` is a real script in this package, so it can rewrite a hook that was meant to outlive a disable without anyone reading the message. The message itself says to move such a hook to the constructor or `init()`, which is the opposite of what the fix does. `meta.hasSuggestions` would keep the one-click fix but make someone pick it. Not a blocker since nothing in `src/` matches today — curious what you think.

handsontable/.config/plugin/eslint/rules/require-tracked-hook-in-enable.js
```diff
@@ -0,0 +1,72 @@
...
+      }
+
+      return false;
+    }
+
+    return {
+      CallExpression(node) {
+        const { callee } = node;
+
+        if (callee.type !== 'MemberExpression' || callee.property.name !== 'addHook') {
```
budnix, line 50: This only matches `addHook`, so `this.hot.addHookOnce()` inside `enablePlugin()` walks straight past — `formulas.ts:667` still has one. `BasePlugin` has no tracked `addHookOnce`, so there is nothing to auto-fix it to, but the same callback can still fire after a disable. The AGENTS.md row lists aliases and helpers as the known gaps; would be good to name this one there too.

handsontable/src/plugins/__tests__/hooksReleasedOnDisable.unit.js
demtario, line 60: **medium** The shared test asserts the hook count returns to baseline after disable, then stops. A hook moved to `this.addHook` but left inside a first-create guard still passes — the Comments `afterSetTheme` failure this PR already fixed in https://github.com/handsontable/handsontable/blob/3a92f01e2e294808501eea0ebc82d2506b8bfc3e/handsontable/src/plugins/comments/__tests__/comments.unit.js#L76-L90. After the disable assertion, enable again and assert `hookCount(hookName)` is greater than `before`.

handsontable/AGENTS.md
demtario: **low** These two Critical Rules bullets point at `src/plugins/sheetsBar/`, which does not exist on this branch or on `develop`. Drop both bullets here (they belong with #13409). Keep the `this.hot.addHook` → `this.addHook` table row.

demtario (commented):
2 inline notes below.

Findings with no changed line to anchor to:

- **high** `handsontable/src/plugins/loading/AGENTS.md` still says `afterDialogFocus` is wired once inside the `#dialogPlugin === null` guard — the pattern this PR removed. Keep resolving `#dialogPlugin` once; say the hook is `this.addHook(...)` on every `enablePlugin()` so `disablePlugin()` can drop it and a later enable can put it back.
- **high** `handsontable/src/plugins/base/AGENTS.md` still lists Loading's `afterDialogFocus` as a deliberate `this.hot.addHook` that must survive `disablePlugin()`, and tells agents not to convert it. Drop Loading from that exception. Keep the auto-size example. The old once-only registration was to avoid doubling on `updatePlugin()`'s disable/enable cycle, which tracked `this.addHook` now handles.
- **medium** `handsontable/src/plugins/comments/AGENTS.md` never records that `afterSetTheme` must be registered on every enable (outside the editor-create guard), or that the shortcut context is immortal and must be reused and cleared by group. Extend the theme bullet with the re-register rule. Add: `getOrCreateContext('plugin:comments')`, and `unregisterShortcuts()` must `removeShortcutsByGroup` on that context as well as `grid`.

handsontable/src/plugins/comments/comments.ts
Haxikowy: Confirmed. `addLocalHook` pushes without a dedupe check and the switch survives a disable, so both listeners doubled on every round trip. Moved them inside the create guard in d16f3120a, next to the editor `resize` hook. The unit test spies `DisplaySwitch.prototype.addLocalHook` and counts one `hide` and one `show` after two round trips; it goes red on the old placement.

handsontable/AGENTS.md
Haxikowy: Both bullets are out of this branch (f721e5cdb). They go in with the plugin in #13409.

handsontable/src/plugins/__tests__/hooksReleasedOnDisable.unit.js
Haxikowy: Added in the shape you suggested (c491149fd). Checked it against the real defect: move `afterDialogFocus` back inside the first-create branch in `loading.ts` and the loading row fails.

handsontable/.config/plugin/eslint/rules/require-tracked-hook-in-enable.js
Haxikowy: Agreed, the fix and the message pull in opposite directions. Now `hasSuggestions: true` with a `replaceWithTrackedAddHook` suggestion instead of `fixable: code` (da6ecc39c), so `eslint:fix` walks past it and someone has to pick the rewrite. RuleTester asserts `output: null` plus the suggestion output.

handsontable/.config/plugin/eslint/rules/require-tracked-hook-in-enable.js
Haxikowy: Named in the AGENTS.md row (f721e5cdb). Left `formulas.ts:667` alone: it is a `once` hook on `init`, so it fires at most once at startup and never against state a later disable tore down. Matching `addHookOnce` in the rule would mean reporting with no replacement to point at, so the guide entry is the right place for it today.

handsontable/src/plugins/__tests__/hooksReleasedOnDisable.unit.js
Haxikowy: Added (c491149fd), with the re-enable assertion after the disable one. Verified against the `Loading` defect rather than the fixed `Comments` one: put `afterDialogFocus` back inside the first-create branch and the loading row fails.

handsontable/AGENTS.md
Haxikowy: Both bullets dropped (f721e5cdb); they belong with the plugin in #13409. The `this.hot.addHook` row stays.

budnix (dismissed):

demtario (commented):
The three off-diff notes from https://github.com/handsontable/handsontable/pull/13410#pullrequestreview-5138285302 are still open on `f721e5cdb`:

- **high** [`handsontable/src/plugins/loading/AGENTS.md`](https://github.com/handsontable/handsontable/blob/f721e5cdb5a3dfe68bb60b8cb9ddc3073ba38b8f/handsontable/src/plugins/loading/AGENTS.md#L17) still says `afterDialogFocus` is wired once inside the `#dialogPlugin === null` guard — the pattern this PR removed. Keep resolving `#dialogPlugin` once; say the hook is `this.addHook(...)` on every `enablePlugin()` so `disablePlugin()` can drop it and a later enable can put it back.
- **high** [`handsontable/src/plugins/base/AGENTS.md`](https://github.com/handsontable/handsontable/blob/f721e5cdb5a3dfe68bb60b8cb9ddc3073ba38b8f/handsontable/src/plugins/base/AGENTS.md#L70) still lists Loading's `afterDialogFocus` as a deliberate `this.hot.addHook` that must survive `disablePlugin()`, and tells agents not to convert it. Drop Loading from that exception. Keep the auto-size example. The old once-only registration was to avoid doubling on `updatePlugin()`'s disable/enable cycle, which tracked `this.addHook` now handles.
- **medium** [`handsontable/src/plugins/comments/AGENTS.md`](https://github.com/handsontable/handsontable/blob/f721e5cdb5a3dfe68bb60b8cb9ddc3073ba38b8f/handsontable/src/plugins/comments/AGENTS.md#L85) never records that `afterSetTheme` must be registered on every enable (outside the editor-create guard), or that the shortcut context is immortal and must be reused and cleared by group. Extend the theme bullet with the re-register rule. Add: `getOrCreateContext('plugin:comments')`, and `unregisterShortcuts()` must `removeShortcutsByGroup` on that context as well as `grid`.

Haxikowy:
The three off-diff notes are handled in c947168e6.

- `loading/AGENTS.md`: the bullet now separates the two. The dialog reference is still resolved once behind the `#dialogPlugin === null` guard; `afterDialogFocus` is a tracked `this.addHook(...)` on every `enablePlugin()`, and the doc says why the guard used to be there and what it cost.
- `base/AGENTS.md`: Loading is out of the exception, and out of the site list along with Comments, since neither has an untracked hook left. The auto-size example stays, with a note that inside `enablePlugin()` the lint rule now enforces the tracked form.
- `comments/AGENTS.md`: new section on what survives a disable. It covers the `afterSetTheme` re-registration, the opposite rule for the `DisplaySwitch` local hooks (`addLocalHook` does not dedupe, so those belong inside the create guard), and the immortal shortcut context with `getOrCreateContext` plus `removeShortcutsByGroup` on both contexts and the handback to `grid`. The theme bullet under "Where to look next" links to it.

demtario (approved):

PR #12841, opened 2026-06-24
fix(selection): prevent active highlight from jumping when Ctrl+clicking a selected cell (DEV-1771)

GabberPL:
## Summary

- **Bug**: Ctrl/Cmd+clicking an already-selected cell in a multi-cell selection caused the active highlight to jump to a different cell instead of staying on the re-clicked cell.
- **Root cause**: `mouseUp()` in `mouseEventHandler.ts` had a "second-click dedup" feature that called `removeLayers()` to remove all selection layers matching the re-clicked cell. When the user re-clicked cell B (which existed as layer 1 and was added again as layer 3), `removeLayers([1, 3])` removed **both** — leaving only other cells, so the active highlight jumped.
- **Fix**: Removed the `removeLayers()`/`pop()` calls from `mouseUp()`. The `refresh()` call is kept (it snaps hover-extended ranges back to their single-cell bounds). Ctrl+clicking an already-selected cell now adds a duplicate layer, making that cell the active focus without touching any other selections.

Note: a previous incomplete fix (PR #12638) only bypassed dedup when `disableVisualSelection` was set — the default case still had the bug.

## Test plan

- [x] E2E tests in `secondClickDeselects.spec.js` (main + MergeCells) rewritten to verify new behavior — 29/29 pass
- [x] Manual repro: select A1, Ctrl+click A2, Ctrl+click A2 again → highlight stays on A2 ✓
- [x] Before/after demos: `demos/DEV-1771/`

ClickUp: DEV-1771

https://claude.ai/code/session_01QQpUhCAtdiayW5zazRjg4a

---
_Generated by [Claude Code](https://claude.ai/code/session_01QQpUhCAtdiayW5zazRjg4a)_

---

> [!NOTE]
> **Medium Risk**
> Changes core mouse-up selection layering used on every Ctrl/Cmd+click in multiple mode, including edge cases with hidden rows/columns and empty selection; behavior is heavily test-covered but still user-visible.
> 
> **Overview**
> Fixes **Ctrl/Cmd+click** on an already-selected cell in **multiple** selection mode so the active highlight no longer jumps to another layer.
> 
[...]

Issue #9594, opened 2022-06-21
Ctrl + left click should deselect the second time.


jzju:
Now it selects the cell multiple times on click.

<img width="263" alt="sel" src="https://user-images.githubusercontent.com/9774380/174800780-6766a0bd-345d-49f1-907b-965aa1cb6177.png">

adrianszymanski89:
Hi @jzju 

Thank you for reporting the issue. Indeed there is a problem here. I'm marking it as a regression.

V11: https://jsfiddle.net/aszymanski/vrf2mx8a/ (working correctly)
V12: https://jsfiddle.net/aszymanski/vrf2mx8a/1/ (select is overwritten on each mouse selection)

jzju:
V11 Don't work correctly for me on Brave Version 1.40.105 Chromium: 103.0.5060.53

adrianszymanski89:
Hi @jzju 

Yes, you are correct. When I tested it previously on v. 11, it seemed to be working as you wanted, but it must have been some error as now it works the same as in v. 12.  However, that's the correct behavior. You can find more information about the selection mode here: https://handsontable.com/docs/api/options/#selectionmode.

The good news is that you can change this setting by using `selectionMode: 'range'`

jzju:
There are different modes in which you can use this plugin. Choose between selecting a single cell, a range of adjacent cells, and multiple non-adjacent ranges of cells.

I don't think that multiple should make it possible to select the same cell multiple times, just that the cells don't have to be continuous. You could try this out in Google docs or excel to see the difference.

AMBudnik:
Hi @jzju 

We have good news. This behavior was changed. You can test it at https://handsontable.com/demo.

We introduced it within v16.0.0. Here https://handsontable.com/docs/react-data-grid/changelog/#_16-0-0 is a full list of changes.

PR #13412, opened 2026-09-08
DEV-2822: Fix the Ctrl+click highlight jump in a multiple selection

budnix:
### Context

The PR fixes the cell highlight jumping to a previously clicked cell when you <kbd>Cmd</kbd>/<kbd>Ctrl</kbd> + click a cell that is already selected.

With `selectionMode: 'multiple'`, that click removed **every** copy of the clicked cell's layer, whatever the highlight was doing. The cell the user just clicked therefore stopped being selected, and the highlight fell back to whichever layer happened to remain — reading, to the user, as the highlight jumping to a cell they clicked one step earlier. Moving the highlight around inside a multi-cell selection was not possible at all.

Reproduced on 16.0.0, on released 18.1.0 and on `develop`; 15.3.0 is clean.

**Correction to the investigation ticket.** DEV-2761 records that PR #12841 (DEV-1771) "targeted a different mechanism". It did not — that PR is the code that jumps. `git log -S "duplicateLayerIndexes"` points at its squash commit `79a463456e`, whose message shows the author changed direction inside the PR: an early commit removed the layer-removal, writing that it "caused ALL matching layers to be removed, making the active highlight jump to an unrelated cell", and a later commit in the same PR re-added it as a "toggle for single-cell layers". That final state shipped in 18.1.0.

#### What changes

No new configuration option. The rule now reads the highlight to tell two intents apart:

- the clicked cell **does not** hold the highlight → the highlight moves to it and it stays selected; only the stale copy of that layer goes;
- the clicked cell **already** holds the highlight → it is deselected, as before. Clicking the last remaining cell still clears the selection.

Two <kbd>Cmd</kbd>/<kbd>Ctrl</kbd> + clicks on the same cell therefore still deselect it, so the v16.0.0 behavior from #9594 is preserved.

The pre-click highlight cannot be recovered in `mouseUp` — the mousedown has already added the clicked layer and moved the focus onto it — so it is captured on mousedown into a `WeakMap` keyed by the `Selection` instance. It is read from `getActiveSelectedRange()` rather than the last layer, because keyboard navigation can rotate the focus onto a layer that is not the top one.

The rule deliberately ignores how fast the clicks arrive, so a <kbd>Cmd</kbd>/<kbd>Ctrl</kbd> + double-click does exactly what the two <kbd>Cmd</kbd>/<kbd>Ctrl</kbd> + clicks it consists of would do. An earlier revision of this PR keyed a carve-out on `event.detail` to stop a double-click deselecting a cell; review found that it also suppressed a *deliberate* fast second click, because `detail` cannot tell the two apart. That made "two clicks deselect" depend on the OS double-click threshold — measurably: the fast pair left the cell selected while the slow pair cleared it. The carve-out is gone, and `tests/e2e/ctrl-click-highlight.spec.ts` now pins both cadences landing in the same place.

#### What changes for existing code

No public API changes: no renamed option, no changed signature, no removed hook, and no changed default. What changes is one interaction path that was released in 16.0.0 through 18.1.0, and it is the defective one. A migration-guide section is added anyway at `docs/content/guides/upgrade-and-migration/migrating-from-18.1-to-18.2/`, next to the other behavior changes already documented there, so the new rule is written down somewhere a reader will find it.

Given a selection of `(0,0)`, `(1,0)`, `(1,1)` with the highlight on `(1,1)`, and a <kbd>Ctrl</kbd> + click on `(1,0)`:

**Before:**

```js
// (1,0) is removed; the highlight falls back to (1,1) — the cell clicked one step earlier.
hot.getSelected();  // [[0, 0, 0, 0], [1, 1, 1, 1]]
```

**After:**

```js
// (1,0) keeps the highlight; nothing else is dropped.
hot.getSelected();  // [[0, 0, 0, 0], [1, 1, 1, 1], [1, 0, 1, 0]]
```

Code that removed a selection layer in response to such a click should drop that workaround — the grid no longer removes the layer for you.

#### Not in scope

The reporting client also says <kbd>Ctrl</kbd> + double-click to start editing is impossible. That is a separate defect and is not fixed here: modifier + double-click has never opened the editor, on 15.3.0 either. It needs its own task and a check against a real Windows <kbd>Ctrl</kbd> key.

### How has this been tested?

New Playwright spec, `tests/e2e/ctrl-click-highlight.spec.ts` (8 tests), covering the reported four-step sequence, alternating adjacent cells, moving the highlight inside a row selection, both deselect paths, the two-click deselect, a focus rotated onto a non-top layer by keyboard navigation, and the same deselect at both click cadences.

```bash
cd tests && npx playwright test --grep "Ctrl\+click inside a multiple selection"
```

Every test was verified to fail on the code it guards, not just to pass on the fix:

- 4 of the 8 go red on `develop` — exactly the four that pin the reported defect.
- `deselects the only selected cell whether the two clicks are fast or slow` and `treats a ctrl+double-click as the two ctrl+clicks it is` both go red against the `detail`-guarded revision.
- `reads the focus from the active layer, not the last one` goes red, and alone, when the capture is switched from `getActiveSelectedRange()` to `current()`.

The last one matters because `locator.click()` pins `clickCount` per call, so two Playwright clicks never look like a double-click however fast they land. Reaching a real cadence needs `page.mouse.down/up({ clickCount })` by hand — `SelectionFeaturesPage#ctrlClickTwiceFast()`.

```bash
# Full matrix, all six theme × bundle legs
cd tests && npx playwright test                                  # 6018 passed, 6 skipped, 0 failed

# Legacy suites most likely to be affected
npm run test:e2e --prefix handsontable --testPathPattern=secondClickDeselects  # 29 specs, 0 failures
npm run test:e2e --prefix handsontable --testPathPattern=selection             # 1270 specs
npm run test:e2e --prefix handsontable --testPathPattern=mergeCells            # 561 specs, 0 failures

npm run test:unit --prefix handsontable                          # 4550 passed
npm run eslint --prefix handsontable                             # 0 errors
npm run stylelint --prefix handsontable                          # clean
```

Seven of the 21 existing `secondClickDeselects` cases flipped and are updated here — the "first cell" case plus all six `disableVisualSelection` cases. Every case that <kbd>Ctrl</kbd> + clicks the *focused* cell is untouched, as are all four MergeCells cases (a merged range is not a single cell, so it never matched the rule). Two test titles that were already wrong before this change are corrected: both read "should keep the re-clicked cell as the active focus" while asserting the opposite, left over from the reversal inside #12841.

One legacy spec, `Selection running in iframe should set correct wtBorder top position of selected cell`, is an order- and display-dependent flake on this machine: it passes in isolation on this branch *and* on the base commit, and came and went across four runs of the same suite with identical numbers. It measures a border's pixel position inside an iframe and touches no selection-layer logic.

### Types of changes

### Related issue(s):

1. DEV-2822
2. DEV-2761 — the investigation ticket this implements
3. DEV-1771 / #12841 — the PR whose final commit introduced this behavior
4. #9594 — the second-click deselect feature, preserved here

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

Documentation updated in this PR: the <kbd>Ctrl</kbd> + click rule under `selectionMode` in the selection guide, and section 7 of the 18.1 → 18.2 migration guide. Two traps found on the way are recorded in `handsontable/AGENTS.md`: the focus does not always sit on the last selection layer, and `event.detail` is always absent in the legacy Jasmine helpers.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2822

---

> [!NOTE]
> **Medium Risk**
> Changes default multi-selection mouse behavior and `getSelected()` layer counts for Ctrl+click on non-focused selected cells; integrators that assumed one-click deselect or hook on every such click may see different results.
> 
> **Overview**
> Fixes **Ctrl/Cmd+click** in `selectionMode: 'multiple'` so re-clicking an already-selected cell no longer makes the highlight jump to a different layer.
> 
> **`mouseEventHandler.ts`** stores the pre-click highlight on mousedown (`getActiveSelectedRange()`, not the top layer) in a per-`Selection` `WeakMap`. On mouseup, when duplicate single-cell layers exist: if the click was on the **focused** cell, all copies are removed (deselect); otherwise only **stale** duplicate layers are stripped so the clicked cell keeps the highlight. Behavior does not depend on `event.detail`, so fast double-clicks match two separate clicks.
> 
> Docs add the rule to the selection guide and **section 8** of the 18.1→18.2 migration guide; changelog **#13412**; `AGENTS.md` notes on active focus vs last layer and avoiding `event.detail`.
> 
> **Tests:** new Playwright `ctrl-click-highlight.spec.ts`; legacy `secondClickDeselects` expectations updated; `SelectionFeaturesPage` gains ctrl-click helpers and `focusCell()` uses `getSelectedRangeActive()`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 14b9e54d50f89e48e40edaeb961cb48538595c48. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2761 Jumping issue: Ctrl+click jumps back to a previous cell instead of staying where they last clicked](https://app.clickup.com/t/9015210959/DEV-2761)
Task linked: [DEV-2822 Fix the Ctrl+click highlight jump inside a multiple selection](https://app.clickup.com/t/9015210959/DEV-2822)
Task linked: [DEV-1771 *Eccam* Selection when used with CMD key jumps between cells](https://app.clickup.com/t/9015210959/DEV-1771)

handsontable/src/selection/mouseEventHandler.ts
```diff
@@ -184,6 +233,26 @@ export function mouseUp({ isLeftClick, selection, cellRangeMapper }: MouseUpOpti
...
+        clickedCell?.isSet() === true &&
+        focusedBefore.isEqual(clickedCell);
+
+      if (!isTogglingOff) {
+        // `findAll` walks the layers in order, so every copy but the last one is stale. Keeping
+        // the last one keeps the focus on the cell the user clicked.
+        selectionRange.removeLayers(duplicateLayerIndexes.slice(0, -1));
+        refreshLayers(selection);
+
+        return;
```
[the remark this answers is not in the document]
budnix: Thanks — the remap gap you point at is real, but this sequence behaves *better* here than on released 18.1.0, so it is not introduced by this PR.

I ran your exact steps against both builds — ctrl+click (2,0), ctrl+click row header 0, ctrl+click (2,1), then ctrl+click (2,0) again:

| | active layer | `selectedByRowHeader` | `isSelectedByRowHeader()` |
|---|---|---|---|
| released 18.1.0 | 1 — a **cell** layer | `[1]` | **true** (wrongly flagged) |
| this branch | 2 — the kept cell | `[1]` | **false** |

The old path removed *both* copies, so the active layer landed on index 1 and collided with the stale flag. The keep-last-layer path leaves the clicked layer last and active at index 2, which is outside the stale set — so the donated flag is never read, and no full-row extension follows.

The underlying issue is genuine: `removeLayers()` re-indexes the layers while `selectedByRowHeader`, `selectedByColumnHeader` and the two extent sets are not remapped. It is pre-existing and already has a `TODO` at `selection.ts:1489`. Fixing it means a remap helper at every `removeLayers` call site, so it is getting its own ticket rather than being folded into this fix.

qunabu (dismissed):

handsontable/AGENTS.md
```diff
@@ -91,6 +91,8 @@ Gotcha: Filters `conditionCollection` uses physical indexes, `getDataAtCol()` us
...
...d click from a lone click — but *not* from a user deliberately clicking twice quickly, and the boundary between them is the OS double-click threshold, which the user configures. A rule that reads it therefore changes with click speed: DEV-2761 shipped a `detail > 1` guard for one review round, and it made "two Ctrl+clicks deselect a cell" silently stop working when the two clicks were fast, because the second one reported `detail: 2` and skipped the toggle. Both suites are blind to this. The legacy Jasmine helpers (`test/helpers/mouseEvents.js`) build events through jQuery's `$.Event(type)`, which never sets `detail`, so `jquery.simulate.js` passes its own default of **0** to `initMouseEvent` — `mouseDoubleClick` is two plain click pairs, indistinguishable from two separate clicks. Playwright is no better: `locator.click()` and `locator.dblclick()` pin `clickCount` per call, so two `click()` calls always look like two unrelated clicks however fast they land. Reaching the real cadence needs `page.mouse.down/up({ clickCount })` by hand — `SelectionFeaturesPage#ctrlClickTwiceFast()` is the worked example, and `tests/e2e/ctrl-click-highlight.spec.ts` pins that the rule ignores cadence.
```
Haxikowy: Checked against `playwright-core@1.45.3`'s server `Mouse.click()`: `dblclick()` calls it with `clickCount:2`, which loops `cc=1..2` doing `down({clickCount:cc}); up({clickCount:cc})` — the same escalating sequence `ctrlClickTwiceFast()` hand-rolls, so `.dblclick()` already reports a genuine `detail: 2` second click. Worth re-checking before this ships as a trap: either the claim needs narrowing to `locator.click()` only, or the helper is redundant with `ctrlDoubleClickCell()`.

.changelogs/13412.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Fixed the cell highlight jumping to a previously clicked cell when you Ctrl/Cmd+click a cell that is already selected. Ctrl/Cmd+click now moves the highlight to that cell and deselects it only when it already holds the highlight.",
```
Haxikowy: Two full sentences in one `title` — denser than this project's usual changelog style. The first sentence alone already names the behavior change; consider trimming.

handsontable/AGENTS.md
budnix: Good spot — measured it and you're right on both counts: `dblclick()` gives `detail [1,2]` while two `click()` calls give `[1,1]`. Narrowed the claim to `locator.click()` and dropped the now-redundant helper in `c41f8e6da0`.

.changelogs/13412.json
budnix: Agreed — trimmed it down to the first sentence in `c41f8e6da0`.

Haxikowy (dismissed):

Haxikowy (approved):

PR #13413, opened 2026-09-08
DEV-2823: Rebuild the row height cache when the row order changes

budnix:
### Context

The PR fixes a rendering regression introduced in 18.1.0: with custom row heights (`autoRowSize`, `rowHeights`, `manualRowResize`) and a column sorted, the grid renders too few rows while you scroll, leaving a chunk of blank white space below the last rendered row. Scrolling further compounds the offset, which is what makes rows look repeated or out of order.

Reported by a client through DEV-2757. Verified as a regression: **17.1.0 clean, 18.0.0 clean, 18.1.0 broken, `develop` broken.**

**Two things in the original report turned out to be wrong**, and are worth recording:

1. **It is not "the 4th click".** The bug tracks whether the grid is *currently sorted*, not how many times the header was clicked. Clicks 1, 2, 4 and 5 (sorted) all break it; clicks 0, 3 and 6 (sort cleared) are clean. The very first click already breaks it.
2. **The `` in the cells is not required.** Stripping every image still reproduces it.

Ruled out by measurement rather than by reading: #12951 (the `modifySinglePassLayout` opt-out does not help), #13184 (the bug follows the sorted *state*, not the click mechanism), and the `oversizedRows` map (empty in this scenario).

### Root cause

`Viewport#rowHeightCache` / `#columnWidthCache` are prefix-sum `PositionCache`s keyed by **render** index, and they are what the viewport calculators read to map a scroll offset onto a band. The sizes behind them resolve per **physical** index — `modifyRowHeight` (so `AutoRowSize`, `ManualRowResize`) and the per-column `width` through `getCellMeta`.

`PositionCache#isCurrent()` re-checks only the item **count**. So any update that changes *which* physical index each render index points at, while leaving *how many* alone, is invisible to it and the cached offsets keep describing the previous layout.

`AutoRowSize` could not catch it either: it invalidates through `observeMapChange(this.rowHeightsMap, …)`, and `rowHeightsMap` is keyed by physical row — a sort changes no value in it, so the observer never fires.

Three shapes were measured stale on the unfixed build:

| Operation | Cache | Cached offset | Truth | Drift |
|---|---|---|---|---|
| Sort (rows) | `getOffset(3)` | 1859 | 1039 | **820px** |
| Same-count hide swap (rows) | `getOffset(12)` | 7473 | 7593 | **120px** |
| Column move | `getOffset(2)` | 977 | 360 | **617px** |

The 1859 is telling: it is 3 × ~620px, the row heights from the **pre-sort** order. A plain `render()` does not fix it; only invalidating the cache does.

### The fix

`onIndexMapperCacheUpdate(state, axis)` in `core.ts` invalidates the axis's size cache whenever **any** of the three change flags is set. One shared per-axis seam, so both axes are covered.

The gate is exactly as wide as it needs to be, and both ways of getting it wrong were hit while building this:

- **Too narrow** — gating on `indexesSequenceChanged` alone (the first version pushed here) fixes sorting but leaves the same-count trim/hide swap broken. Caught in review, then reproduced: 120px of drift.
- **Too wide** — dropping the gate makes `updateCache(force = true)`, which arrives with every flag false and nothing rearranged, pay a cache *and layout* invalidation on every forced no-op. That reproducibly timed out `Core#updateSettings with external data source › should keep ignoring static data after a DataProvider updatePlugin cycle`.

This does not touch the per-render caching #13078 added — that is about renders, edits and scroll steps, none of which come through this hook.

### How has this been tested?

New Playwright E2E (`tests/e2e/sorted-row-height-cache.spec.ts`), 4 cases on a window-scrolled fixture (no `height`, `autoRowSize` on, wrapped-text heights scrambled against the row index):

1. No blank space while scrolling, before sorting and after asc/desc.
2. The same through the **header-click** path — the gesture the ticket actually reports, which since #13184 sorts on mouse up over the label only.
3. Cached row offsets stay true across a sort **and** across a same-count hide swap.
4. Cached column offsets stay true across a column move.

**Each case was verified to fail without the fix.** With the invalidation disabled in the built bundle all 4 fail; restored, all 4 pass. The specs also assert their own preconditions (row heights genuinely vary, and the sort genuinely reorders), so they cannot go green vacuously if the fixture stops producing varied heights — and the gap probe now throws rather than reporting "no gap" when it cannot find the grid.

The 2px tolerance on the offset assertions is the first rendered row's documented 1px border-top compensation, not slack — the real drift is 120–820px.

```bash
npm run eslint --prefix handsontable                     # 0 errors
npm run test:unit --prefix handsontable                  # 4550 passed
cd tests && npx playwright test --project=e2e-main       # 1000 passed
npm --prefix handsontable run test:e2e -- --testPathPattern='(columnSorting|autoRowSize|manualColumnMove|hiddenRows|trimRows|filters)'
                                                         # 1247 specs, 0 failures
```

Also verified by hand against the client's own reproduction (1000 rows, sampled `autoRowSize`, image-bearing cells): every gap goes to 0 across the entire 495k-pixel scroll range, for every sort state, and with `viewportRowRenderingOffset` at both `0` and `auto`.

### Known gap, deliberately not fixed here

`wtViewport.oversizedRows` is reorder-blind in the same way — it is keyed by renderable row, and `resetOversizedRows` wipes only the rendered band, so on a reorder every index outside that band keeps the previous order's measurement. That matters when walkontable measures rows itself rather than a `modifyRowHeight` provider supplying them. Pre-existing, not reproduced, and out of scope here; recorded in `src/3rdparty/walkontable/AGENTS.md` so the next person does not assume the position cache was the only one.

### Types of changes

### Related issue(s):

1. DEV-2823
2. DEV-2757

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2823

---

> [!NOTE]
> **Medium Risk**
> Touches core viewport sizing on every index-mapper cache update (sort, filter, hide, trim, move); gated to real mapping changes but affects scroll/render hot paths for custom-sized grids.
> 
> **Overview**
> Fixes an **18.1.0 regression** where grids with variable row heights or column widths (e.g. `autoRowSize`, manual resize) could render too few tracks after **sorting, column moves, or same-count hide/trim swaps**, leaving blank space below or beside the painted table and making rows look duplicated or out of order while scrolling.
> 
> **`onIndexMapperCacheUpdate` in `core.ts`** now drops the Walkontable **row-height and column-width prefix-sum caches** whenever the index mapper reports **any** mapping change—`indexesSequenceChanged`, `trimmedIndexesChanged`, or `hiddenIndexesChanged`—not sequence alone. Those caches are keyed by render index but sizes come from physical indexes, and `PositionCache#isCurrent()` only notices count changes, so permutations and same-count hide/trim swaps used to keep stale offsets.
> 
> Documentation in **`AGENTS.md`** and **`walkontable/AGENTS.md`** records the invalidation rules and the separate `oversizedRows` blind spot (out of scope here). A **changelog** entry and new **Playwright E2E** coverage (`sorted-row-height-cache.spec.ts` plus fixture/page object) assert no blank gaps while scrolling after sort, offset drift after hide/trim swaps and column moves, and header-click sorting.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 866b7370b60c87181ef9ae654e0ea5916a1ddf9d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2757 [18.1.0] The 4th sorting action removes rows](https://app.clickup.com/t/9015210959/DEV-2757)
Task linked: [DEV-2823 The row-height cache goes stale after sorting with autoRowSize, leaving blank gaps on scroll](https://app.clickup.com/t/9015210959/DEV-2823)

tests/e2e/sorted-row-height-cache.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+    await grid.waitForStableScrollRange();
+
+    expect(Math.abs(await grid.offsetDrift('row', 12))).toBeLessThanOrEqual(2);
+
+    // This one does not: it swaps WHICH rows are hidden while keeping HOW MANY the same. It is the
+    // shape the first version of this fix missed, because it was gated on `indexesSequenceChanged`.
+    await grid.swapHiddenRows(9, 3);
+    await grid.waitForStableScrollRange();
+
+    expect(Math.abs(await grid.offsetDrift('row', 12))).toBeLessThanOrEqual(2);
```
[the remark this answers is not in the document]
budnix: Good spot — the swap really was a no-op, fixed in `5faa7b23ba`: it now hides row 3 first, then swaps 9 and 3, with a count check to keep it honest. One correction: `hideRow` does not force-flush inside `batchExecution(fn, true)` — the real cause was `swapHiddenRows(3, 3)` canceling itself out.

budnix:
## Visual diffs — I checked all 14, and they are a fix

I went through every one of the 14 changed screenshots and compared each pair pixel by pixel. I am confident these show the fix working, not a regression.

**11 of them are pagination** (`auto-page-size-table-without-defined-size`, steps 2/5/6/7, on chromium plus main, main-dark, horizon and horizon-dark).

In all 11, the only part that changes is a band at the bottom of the image, where the last row meets the pagination bar. They show one of two things:

- **A blank strip.** In the old baseline there is empty space between the last row and the pagination bar. It is clearest on `chromium-theme-horizon-dark/...-6`: two rows, then a gap, then the bar sitting on its own at the bottom. With the fix the bar sits right under the last row.
- **A clipped row.** In the old baseline the last row is cut in half by the pagination bar. On `chromium-theme-horizon/...-2` row 16 "LL Bottom Bracket" is sliced through the middle. On `chromium/...-6` the Name cell reads "HL Road" with the rest cut off; with the fix it draws at its real height and wraps to "HL Road Frame".

The page numbers are identical in both versions everywhere I looked — "1 - 2 of 44 / Page 1 of 17", "9 - 16 of 44 / Page 2 of 5", "43 - 44 of 44 / Page 17 of 17". So the page size did not change. Only the row height got corrected.

That blank strip is exactly the bug this PR fixes: a stale row height cache let the grid place content in the wrong spot. The old screenshots were recording the bug.

**3 of them are merged cells** (`scroll-merged-cells-demo-1` on chromium, firefox and webkit). These are only a small vertical scroll offset — 1px on chromium and firefox, 2px on webkit. When I shift the new image by that amount, the difference drops from about 7% to 0.8%, and what is left is text antialiasing. The rows and their content are the same. It comes from the 1px border compensation on the first rendered row, so I read it as harmless.

My conclusion: the new screenshots are correct and the baseline should be updated.

handsontable/src/core.ts
```diff
@@ -899,6 +899,39 @@ export default function Core(
...
+    // rebuild on its own anyway, so those cost nothing.
+    //
+    // It is a gate rather than an unconditional call because `updateCache(force = true)` reaches
+    // here with every flag false and nothing rearranged. `invalidateRowHeightCache()` also drops the
+    // layout, so paying that on each forced no-op update is what pushed the DataProvider's
+    // repeated `updateSettings` cycles past their timeout.
+    //
+    // This does not touch the per-render caching #13078 added - that is about renders, edits and
+    // scroll steps, none of which come through here.
+    const mappingChanged = indexesChangesState.indexesSequenceChanged ||
```
Haxikowy, line 927: No unit or E2E test isolates `trimmedIndexesChanged` true with the other two flags false, e.g. a plain Filters/TrimRows trim with no sort. A regression on that branch ships green.

handsontable/src/core.ts
```diff
@@ -899,6 +899,39 @@ export default function Core(
...
+    // leaving blank space past the last rendered row or column.
+    //
+    // The gate is "did the mapping actually change", NOT `indexesSequenceChanged` - gating on the
+    // sequence alone was the first version of this fix and it left the same-count trim/hide swap
+    // broken. All three flags have to invalidate; where the count also changed the cache would
+    // rebuild on its own anyway, so those cost nothing.
+    //
+    // It is a gate rather than an unconditional call because `updateCache(force = true)` reaches
+    // here with every flag false and nothing rearranged. `invalidateRowHeightCache()` also drops the
+    // layout, so paying that on each forced no-op update is what pushed the DataProvider's
```
Haxikowy: This timeout claim doesn't reproduce: unconditional invalidation on `updateSettingsExternalDataSource.unit.js`'s DataProvider test ran ~2.8s, same as gated, both far under Jest's 5s timeout, since `invalidate()` itself is O(1). Re-verify with a real repro or soften the claim to "avoids paying invalidation on no-ops".

Haxikowy (commented):
A second `updateCache(force = true)` call site exists in `pagination.ts:837`, beyond the DataProvider one analyzed in this PR. It's safe in every traced path today, but nothing pins that invariant if pagination's batching logic changes.

handsontable/src/core.ts
budnix: Good catch — added a same-count trim swap case in `50173f9817`, and it is the only test that fails when `trimmedIndexesChanged` is dropped.

handsontable/src/core.ts
budnix: You are right — it does not reproduce; the full suite passes either way, so I dropped that claim in `50173f9817` and now name both forced call sites (pagination and DataProvider) instead. My earlier failures were machine load from a concurrent Playwright run.

tests/fixtures/pages/SortedRowHeightCachePage.ts
```diff
@@ -0,0 +1,305 @@
...
+          getPlugin: (n: string) => { hideRow: (r: number) => void, showRow: (r: number) => void },
+        },
+      }).hot;
+      const plugin = hot.getPlugin('hiddenRows');
+
+      hot.batchExecution(() => {
+        plugin.hideRow(toHide);
+        plugin.showRow(toShow);
+      }, true);
+    }, [hide, show]);
```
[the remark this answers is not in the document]
budnix: You are right, and I was wrong to push back on this last time — the swap fired two cache updates (hide alone at count 118, show at 119), so the count invalidated it and the case never ran. Fixed in `8d5679b51e` by showing before hiding, which keeps both writes in one flush; dropping `hiddenIndexesChanged` now fails this test and nothing else.

tests/e2e/sorted-row-height-cache.spec.ts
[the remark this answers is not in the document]
budnix: Correction to my earlier reply here: `hideRow` does force-flush inside `batchExecution`, so you were right and I was not. I had read it off `isCurrent()` being false, which a count change explains just as well; the event trace settled it, and the fix is in `8d5679b51e`.

Haxikowy (dismissed):

Haxikowy (approved):

PR #11790, opened 2025-08-21
Add Classic (modern) theme and information about Classic (legacy) depracation

adrianspdev:
### Context
This PR introduces new Classic (Modern) theme CSS variables, imported from Figma. The documentation has been updated accordingly, including information about the deprecation.

### How has this been tested?
Locally

### Types of changes

- [x] New feature or improvement (non-breaking change which adds functionality)

### Related issue(s):
1. https://github.com/handsontable/dev-handsontable/issues/2812

### Affected project(s):
- [x] `handsontable`

### Checklist:

- [x] I have reviewed the guidelines about [Contributing to Handsontable](https://github.com/handsontable/handsontable/blob/master/CONTRIBUTING.md) and I confirm that my code follows the code style of this project.
- [x] I have signed the [Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLScpMq4swMelvw3-onxC8Jl29m0fVp5hpf7d1yQVklqVjGjWGA/viewform?c=0&w=1)
- [x] My change requires a change to the documentation.

[skip changelog]

PR #13414, opened 2026-09-08
DEV-2824: Highlight the 18.1 features on the changes-between-versions page

demtario:
### Context

The PR adds the editorial highlights for 18.1 to the "Changes between versions" page, so the release's headline features render as featured cards instead of plain changelog bullets:

- Shadow DOM and Salesforce support ([#13194](https://github.com/handsontable/handsontable/pull/13194))
- Single-pass rendering, standing in for the release's performance work ([#12951](https://github.com/handsontable/handsontable/pull/12951))
- Resize and move a selection by dragging ([#13076](https://github.com/handsontable/handsontable/pull/13076))
- The new `intl-datetime` cell type ([#13151](https://github.com/handsontable/handsontable/pull/13151))

`docs/content/data/version-highlights/18.1.json` was an empty scaffold until now, so the 18.1 group had no cards at all.

Two changes in `VersionComparison.tsx` were needed to make those four cards actually reach the page the task links to (`?from=15.0&to=18.1`, which resolves to the **New** tab):

1. **Every highlight is now promoted onto the New tab**, not only breaking ones. The previous rule (`SHOW_BREAKING_HIGHLIGHTS_ON_NEW`) worked for every earlier release because each non-`added` highlight happened to also be breaking. 18.1 is the first release whose headline feature is filed under `#### Fixed` (Shadow DOM) with the performance work under `#### Changed`, and no data-only mapping can put those on the New tab, because no `added` bullet exists for them. The constant is kept and renamed (`SHOW_HIGHLIGHTS_ON_NEW`) so the rule stays revertable in one line.
2. **Featured cards are deduplicated by PR number.** One PR can be cited by two changelog bullets in different sections: 18.1's #12951 is both an `added` hook and a `changed` render path, and 16.1's #11790 is both `added` and `deprecated`. Without this, change 1 would render "Classic theme" twice on the New tab of the 16.1 group, and #12951 twice on 18.1. The losing bullet keeps its place in the compact list on tabs its own category belongs to, and is dropped on the New tab so a promoted highlight cannot drag an unrelated-category bullet onto it.

Three notes for the reviewer:

- A cherry-pick to `prod-docs/*` needs **both** files. `18.1.json` alone leaves the Shadow DOM card off the default New tab, because the promotion rule lives in `VersionComparison.tsx`.
- The tab counts come from `matchesFilter` alone, so the promoted Shadow DOM card shows on **New** without being counted in the New tab's number. That mismatch already existed for promoted breaking highlights, so it is not a regression, but it is now visible on every release.
- The `codeBefore` / `codeAfter` fields are validated and loaded but not rendered by `FeaturedEntry` today, so the new entries omit them.

### How has this been tested?

Docs-only change. Verified in a browser against a local dev server (`npx astro dev` in `docs/`) at `/docs/javascript-data-grid/changes-between-versions/?from=15.0&to=18.1`:

- The default **New** tab renders four featured cards for 18.1.0: #12951, #13076, #13151 (New pill) and #13194 (Fixed pill).
- The 16.1.0 group still renders exactly one "Classic theme" card on both **New** and **All** (no duplicate).
- On **All**, #12951 renders one card, with its second bullet in the compact list.

```bash
npm --prefix docs run docs:validate-highlights   # All highlight files are valid.
npm --prefix docs run docs:test:plugins          # 257 pass, 0 fail
npm --prefix docs run docs:lint
docs/node_modules/.bin/tsc --noEmit --jsx react-jsx --module esnext --moduleResolution bundler \
  --target es2022 --strict --skipLibCheck --types react \
  docs/src/components/VersionComparison/VersionComparison.tsx
```

`[skip changelog]`

### Types of changes

### Related issue(s):

1. DEV-2824

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

- [x] I have signed the [Contributor License Agreement](https://cla.handsontable.com/sign) - one signature covers both Handsontable and HyperFormula; the `cla/signed` check on this PR confirms it.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2824

---

> [!NOTE]
> **Low Risk**
> Docs-only data and presentation logic for the version comparison page; no grid runtime or auth/data paths affected.
> 
> **Overview**
> Adds **four editorial featured cards** for release **18.1** (Shadow DOM/Salesforce, single-pass rendering, selection drag resize/move, `intl-datetime`) by filling `docs/content/data/version-highlights/18.1.json`, which was previously an empty scaffold.
> 
> Updates the **Changes between versions** UI so those cards show on the default **New** tab even when changelog bullets are under **Fixed** or **Changed**: the old “breaking-only” promotion becomes **`SHOW_HIGHLIGHTS_ON_NEW`**, which surfaces **every** highlighted PR on New (still revertable in one constant).
> 
> Refactors featured/compact splitting into **`featuredEntries.ts`** and adds **`partitionEntries`** to **dedupe featured cards by PR number** when one PR appears in multiple changelog sections; secondary bullets stay in the compact list only on tabs that match their category. **`VersionComparison.tsx`** wires in the module, and new **unit tests** lock in promotion, deduplication, and tab behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 23a7f41a1b5451ae88797c4502779f7506c4ca46. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/content/data/version-highlights/18.1.json
+++ b/docs/content/data/version-highlights/18.1.json
@@ -1,4 +1,25 @@
 {
   "version": "18.1",
-  "highlighted": []
+  "highlighted": [
+    {
+      "prNumber": 13194,
+      "tagline": "Shadow DOM and Salesforce support",
+      "whyItMatters": "The grid now works inside Shadow DOM and sandboxed hosts such as Salesforce Lightning Web Components: the cell editor stays open when clicked, copy and paste work, focus is released when it leaves the grid, an outside click clears the selection, and the grid's z-index values no longer paint over the host page's UI."
+    },
+    {
+      "prNumber": 12951,
+      "tagline": "Single-pass rendering",
+      "whyItMatters": "The grid predicts whether scrollbars will appear instead of rendering, measuring, and re-rendering, which cuts the work of the initial render and of every scroll step. It is one of over 20 performance changes in 18.1 that also cover sorting, filtering, index translation, and cell metadata memory - use the new `modifySinglePassLayout` hook to keep a table on the previous layout path."
+    },
+    {
+      "prNumber": 13076,
+      "tagline": "Resize and move a selection by dragging",
+      "whyItMatters": "The `selectionHandles` option shows draggable handles at each edge midpoint of a selected range, so you can resize the selection with a pointer. The `moveCells` option lets you drag a selection's border to move its cells to another location."
+    },
+    {
+      "prNumber": 13151,
+      "tagline": "New intl-datetime cell type",
+      "whyItMatters": "The `intl-datetime` cell type edits dates and times in one cell with a native date-time picker, and displays them through `Intl.DateTimeFormat` with the new `dateTimeFormat` option. Column sorting, filtering, and XLSX export work with it out of the box."
+    }
+  ]
 }
--- a/docs/content/data/version-highlights/18.1.json
+++ b/docs/content/data/version-highlights/18.1.json
@@ -9,7 +9,7 @@
     {
       "prNumber": 12951,
       "tagline": "Single-pass rendering",
-      "whyItMatters": "The grid predicts whether scrollbars will appear instead of rendering, measuring, and re-rendering, which cuts the work of the initial render and of every scroll step. It is one of over 20 performance changes in 18.1 that also cover sorting, filtering, index translation, and cell metadata memory - use the new `modifySinglePassLayout` hook to keep a table on the previous layout path."
+      "whyItMatters": "The grid predicts whether scrollbars will appear instead of rendering, measuring, and re-rendering, which cuts the work of the initial render and of every scroll step. It is one of over 20 performance changes in 18.1 that also cover sorting, filtering, index translation, and cell metadata memory. Use the new `modifySinglePassLayout` hook to keep a table on the previous layout path."
     },
     {
       "prNumber": 13076,
--- a/docs/src/components/VersionComparison/VersionComparison.tsx
+++ b/docs/src/components/VersionComparison/VersionComparison.tsx
@@ -142,22 +142,26 @@ function matchesFilter(entry: VersionEntry, filter: FilterKind): boolean {
   }
 }
 
-// Breaking highlights are also surfaced on the New tab for now, so a major
-// release's headline features (often breaking, e.g. a TypeScript migration or a
-// new layout system) stay visible on the default landing view. Set this to false
-// to revert to showing breaking highlights only on the Breaking and All tabs.
-const SHOW_BREAKING_HIGHLIGHTS_ON_NEW = true;
+// Every highlight is also surfaced on the New tab, so a release's headline
+// features stay visible on the default landing view whichever changelog section
+// they were filed under. An author picks at most five highlights per release
+// (see docs/scripts/validate-version-highlights.mjs), and the section a feature
+// lands in is an implementation detail of the changelog: 18.1's Shadow DOM
+// support sits under Fixed and its single-pass rendering under Changed. Set this
+// to false to revert to showing a highlight only on All and on the tab matching
+// its own category.
+const SHOW_HIGHLIGHTS_ON_NEW = true;
 
 // Whether a highlighted entry renders as a featured card under the active filter.
 // A highlight always shows on All and on the tab matching its own category
 // (a deprecated highlight on Deprecated, a breaking one on Breaking, and so on),
 // so it never leaks onto an unrelated tab. The one exception is the revertable
-// rule above that also promotes breaking highlights onto New.
+// rule above that also promotes every highlight onto New.
 function isFeatured(entry: VersionEntry, filter: FilterKind): boolean {
   if (!entry.highlighted) return false;
   if (filter === 'all') return true;
   if (matchesFilter(entry, filter)) return true;
-  return SHOW_BREAKING_HIGHLIGHTS_ON_NEW && filter === 'new' && entry.breaking;
+  return SHOW_HIGHLIGHTS_ON_NEW && filter === 'new';
 }
 
 interface FilterTabsProps {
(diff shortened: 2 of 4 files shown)


qunabu:
Task linked: [DEV-2824 DOCS - Changes between versions - highlight 18.1 boxes](https://app.clickup.com/t/9015210959/DEV-2824)

docs/src/components/VersionComparison/VersionComparison.tsx
```diff
@@ -296,8 +300,27 @@ const COMPACT_THRESHOLD = 5;
...
+  // the compact list - but only on a tab its own category belongs to, so a
+  // promoted highlight does not drag a second bullet onto the New tab.
+  const featured: VersionEntry[] = [];
+  const compact: VersionEntry[] = [];
+  const featuredPrNumbers = new Set<number>();
+
+  for (const entry of entries) {
+    const alreadyFeatured = entry.prNumber !== null && featuredPrNumbers.has(entry.prNumber);
+
+    if (isFeatured(entry, filter) && !alreadyFeatured) {
```
budnix: This loop keeps whichever bullet comes first as the featured card, so the card's pill comes from that bullet's changelog section. Today `#### Added` sits above `#### Changed`, which is why #12951 shows a New pill; if those sections ever swap, the New tab would show a Changed pill instead and drop the `added` bullet. A test would pin this down, and `docs:test:plugins` already runs `src/components/__tests__/*.test.mjs` for exactly this kind of logic. The CSS import on line 1 blocks the transpile-and-import trick those tests use, so `isFeatured` and this loop would need to move into a plain `.ts` sibling first.

docs/src/components/VersionComparison/VersionComparison.tsx
```diff
@@ -296,8 +300,27 @@ const COMPACT_THRESHOLD = 5;
...
+      if (entry.prNumber !== null) featuredPrNumbers.add(entry.prNumber);
+      featured.push(entry);
+    } else if (!alreadyFeatured || matchesFilter(entry, filter)) {
```
budnix: `!alreadyFeatured ||` never changes the outcome here. `ReleaseGroup` only receives entries that already passed `isFeatured(e, filter) || matchesFilter(e, filter)` in the `visible` memo, so reaching this branch with `alreadyFeatured` false means `isFeatured` was false, and that forces `matchesFilter` to be true. When `alreadyFeatured` is true the condition reduces to `matchesFilter` anyway, so the plain check states the rule more directly.

```suggestion
    } else if (matchesFilter(entry, filter)) {
```

docs/content/data/version-highlights/18.1.json
```diff
@@ -1,4 +1,25 @@
...
+  "highlighted": [
+    {
+      "prNumber": 13194,
+      "tagline": "Shadow DOM and Salesforce support",
+      "whyItMatters": "The grid now works inside Shadow DOM and sandboxed hosts such as Salesforce Lightning Web Components: the cell editor stays open when clicked, copy and paste work, focus is released when it leaves the grid, an outside click clears the selection, and the grid's z-index values no longer paint over the host page's UI."
+    },
+    {
+      "prNumber": 12951,
+      "tagline": "Single-pass rendering",
+      "whyItMatters": "The grid predicts whether scrollbars will appear instead of rendering, measuring, and re-rendering, which cuts the work of the initial render and of every scroll step. It is one of over 20 performance changes in 18.1 that also cover sorting, filtering, index translation, and cell metadata memory - use the new `modifySinglePassLayout` hook to keep a table on the previous layout path."
```
budnix: The second sentence carries two ideas at once: how many performance changes 18.1 has, and the hook that keeps a table on the old layout path. The docs voice guide asks for one idea per sentence, and this card is where a reader first meets single-pass rendering, so three shorter sentences land better.

```suggestion
      "whyItMatters": "The grid predicts whether scrollbars will appear instead of rendering, measuring, and re-rendering, which cuts the work of the initial render and of every scroll step. It is one of over 20 performance changes in 18.1 that also cover sorting, filtering, index translation, and cell metadata memory. Use the new `modifySinglePassLayout` hook to keep a table on the previous layout path."
```

docs/src/components/VersionComparison/VersionComparison.tsx
demtario: Extracted `isFeatured` and this loop into `featuredEntries.ts` (plain `.ts`, so `ts.transpileModule` + data-URL import works like `toc-keyboard-nav.test.mjs` does with `a11y-utils.ts`), and added `src/components/__tests__/version-comparison-featured.test.mjs`.

The first-bullet-wins rule is now pinned: one case feeds both #12951 bullets and asserts the featured card carries the `added` pill while the `changed` bullet falls to the compact list on All and is absent on New. `partitionEntries`' JSDoc states that the card's pill comes from whichever section appears first in the changelog. 8 cases total; I checked they are not hollow by flipping `SHOW_HIGHLIGHTS_ON_NEW` to false (2 fail) and by removing the dedupe guard (1 fail).

docs/src/components/VersionComparison/VersionComparison.tsx
demtario: Correct, the `visible` memo makes that half dead. Applied your suggestion.

Since `partitionEntries` now lives in its own module and can be imported without going through the page, I added the precondition to its JSDoc: it expects entries already narrowed by `isFeatured(e, filter) || matchesFilter(e, filter)`, and drops anything outside that set rather than listing it. The test feeds its fixtures through that same filter.

docs/content/data/version-highlights/18.1.json
demtario: Applied as suggested.

budnix (approved):

PR #13415, opened 2026-09-08
DEV-2790: Check and document changelog API reference links

demtario:
### Context

The PR documents and checks a docs convention that lapsed silently in April 2024, and it is the first of two PRs for DEV-2790. The second one backfills the missing links.

Changelog entries under `Added` used to link every new option, hook, method, and plugin to its reference page, like `` [`renderAllColumns`](@/api/options.md#renderallcolumns) ``. They stopped: a reader of `changelog-15` through `changelog-18` gets a GitHub PR number and nothing else, so finding out what `searchMode` or `minRowHeights` does means leaving the page and searching the docs by hand.

The links were never produced by tooling. They came from a manual editorial pass in a dedicated post-release PR titled `Docs: Edit <version> docs`, run 14 times between 2021 and 2024. The last one is `45225f48c4` (2024-03-05, the 14.2.0 page); the next release, `f48236aedc` (2024-04-16, 14.3.0), pasted the root `CHANGELOG.md` section verbatim with no follow-up pass, and every release since has done the same. `git log --all -S '@/api/' -- .changelogs/` and the same over `CHANGELOG.md` are both empty on every ref, so the pipeline never carried such a link. The practice lived only in one person's habit, with no checklist, script, or check encoding it, and ended when they did.

Note that the regression starts at **14.3.0**, not at 15.0 as the task describes it. The break sits inside `changelog-14.md`; 15 through 18 inherit it.

Two things this PR settles:

1. **The rule cannot live in a `.changelogs/*.json` entry title.** `bin/changelog` renders `title` verbatim into four destinations, and only the docs page resolves `@/api/` links. On GitHub the root `CHANGELOG.md` and the release body would link to a literal `@/api/...` path and 404, and the version-comparison UI drops the link and keeps the text. So linking stays a docs-page step, which is exactly where the lost pass sat. An entry title that needs a docs link uses an absolute `https://handsontable.com/docs/...` URL, the form `CHANGELOG.md` already uses 82 times.
[...]

PR #13416, opened 2026-09-08
DEV-2790: Link new API to its reference in the changelog pages

demtario:
### Context

The PR restores the reference links in the changelog pages, and it is the second of two PRs for DEV-2790. The first one, #13415, documents the convention and adds the check whose report produced this worklist.

Entries under `Added` used to link every new option, hook, method, and plugin to its reference page. That stopped at 14.3.0 (April 2024), when the manual post-release editing pass that produced the links ended, so a reader of `changelog-15` through `changelog-18` gets a GitHub PR number and nothing else. #13415 carries the full root-cause trail; this PR is the content fix.

Scope is 14.3.0 and later, which is where the regression actually starts. The 14.0.0, 14.1.0, and 14.2.0 sections already carry their links from the old pass and are untouched. Only `Added` sections are edited, and only to add links, so the text still matches the root `CHANGELOG.md` line for line and the diff stays reviewable.

What is deliberately left unlinked, because it has no reference-page anchor: theme tokens (`paginationButton*`), CSS class names and style variants (`no-icons`, `icons-only`), TypeScript type names (`SanitizerContext`), external APIs (`Intl.NumberFormat`, `Intl.DateTimeFormat`), object-shape keys that are not API members (`key`/`value`), wrapper package names, and the plugin-nested settings that the generated reference does not document as members (`searchMode`, `roundFloat`, `style`, `template`, `rowspan`, `visibleWhen`). Where an entry names such a setting, the plugin around it is linked instead, which is what the reader can actually navigate to. A link to a page that does not document the name is worse than no link.

Two judgement calls worth flagging in review:

- `changelog-14.md` line 141 says "the `manualRowResize` plugin", so it links to `@/api/manualRowResize.md`, the plugin page, not to the option of the same name. The check suggested the option; the sentence means the plugin.
- `changelog-15.md` named a `beforeCompositionstart` hook. The registered hook is `beforeCompositionStart` (`handsontable/src/core/hooks/constants.ts:453`), so the link would never have resolved. **The same typo is in the root `CHANGELOG.md`, and this PR corrects it there too**, so the two stay identical. That is the only change to `CHANGELOG.md`, and the only word changed anywhere in this PR.

One real gap this surfaced, not fixed here: `docs/content/api/dataProvider.md` is generated and reachable at `/api/data-provider`, but `docs/content/api/sidebar.js` does not list it, so the `DataProvider` reference is absent from the API navigation. The link added in `changelog-17.md` resolves; the navigation entry is a separate change.

`[skip changelog]`

### How has this been tested?

- Every `@/api/` link in `changelog-14` through `changelog-18` was verified against a **really generated** reference, not against a name list: the core package was built and `docs:api` run, then each link's target page and anchor were checked to exist as a heading.
  ```bash
  # 67 distinct links checked -> all resolve (this also covers the pre-existing 14.0-14.2 links)
  ```
  The anchor check matters because `framework-loader.mjs` resolves only the `.md` half of the link and passes `#anchor` through untouched, so a wrong anchor renders as a working link to the wrong place.
- The check from #13415, run before and after: **44 findings before, 13 after**, and all 13 remaining are outside this PR's scope (`changelog-8`, `changelog-11`, and the 14.0.0/14.1.0 sections).
- No line was added or removed in any changelog page, only rewritten with links.

A second commit followed a review of #13415, which widened the check to `Core` members. That surfaced three method names this PR had missed: the 18.0.0 `NestedRows` entry names `updateSettings()`, `loadData()`, and `updateData()`, now linked to `@/api/core.md`. The plugin methods in that same entry (`collapseAll()` and the twelve others) were already linked by hand, because the check cannot attribute a bare method name to one plugin.

### Types of changes

### Related issue(s):

1. DEV-2790

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2790


qunabu:
Task linked: [DEV-2790 Changelogs from 15.0 onward stopped linking new API to reference pages](https://app.clickup.com/t/9015210959/DEV-2790)

budnix (approved):

Issue #3828, opened 2016-10-18
preventOverflow and dropdown, autocomplete render bug


vkomphasouk:
### Description

When you expand the dropdown or autocomplete list, only a small part of the list is rendered with a vertical scrollbar instead of the full list. It seems the height calculation is of when using preventOverflow = 'horizontal'
### Steps to reproduce

...
preventOverflow: 'horizontal'
columns: [
  {
    type: 'autocomplete',
    filter: false,
    source: ...
  },
  {
    type: 'dropdown',
    source: ...
  }
]
...
### Demo

[jsfiddle](http://jsfiddle.net/csjnsftz/)

### Related issue

- #8688

### Your environment
- Version: 
- Browser Name and version: IE 11 & Chrome
- Operating System: Win 10

AMBudnik:
Thanks for sharing @vkomphasouk 
It doesn't look as desired.

AMBudnik:
Still replicable in version 0.31.0

![example23](https://cloud.githubusercontent.com/assets/10757813/23359343/1f374034-fce6-11e6-9b3a-ef18052579b6.gif)

krzysztofspilka:
Also, according to the information shared in the support channel, it's reproducible when one instance of Handsontable is placed right below the other.

AMBudnik:
Hi @vkomphasouk 
I have pumped the version to 1.15.1 and it seems to work better. Can you check it, please?

AMBudnik:
Please take a look at https://github.com/handsontable/handsontable/issues/5032
Where @djcooke provides a workaround

ynte:
The workaround still gives undesirable results when you're working with multiple rows: [https://jsfiddle.net/7c1fqrto/](https://jsfiddle.net/7c1fqrto/). The dropdown does not open in the right place.

AMBudnik:
inform [forum](https://forum.handsontable.com/t/gh-4172-3828-dropdown-not-showing-all-values-due-of-last-record-due-to-scroll-bar/3821)

aninde:
This issue will be fixed by the upcoming v12.1.0 http://jsfiddle.net/3otc524p/

AMBudnik:
Hi @vkomphasouk 

I have great news! We just released Handsontable v12.1.0 where this issue is solved.

If it won't work for you after the update please let me know at support@handsontable.com

Issue #6045, opened 2019-06-07
Opening dropdown on fixed row opens wrong way


jeremy-smith-maco:
### Description

The dropdown on a fixed row will open downwards if not scrolled to the bottom of the table. It should be opening upwards.

### Steps to reproduce

1. Set `minSpareRows` above 0
2. Have a `dropdown` type cell
3. Open dropdown cell

### Demo

http://jsfiddle.net/tv174upw/1/

### Your environment
* Handsontable version: 6.0.1
* Browser Name and version: Chrome Version 69.0.3497.100
* Operating System: Windows 10

AMBudnik:
Thank you for sharing the issue report. You're right. Now it looks like the dropdown is broken. The 96-row data needs to be visible to get the list of options.

pnowak:
This issue is still replicable using v8.0.0. Here's an updated demo - https://jsfiddle.net/ot8k9cgb/

AMBudnik:
It works the same in v8.2.0 http://jsfiddle.net/handsoncode/kondga26/. 
We can only see the list of options when we scroll all the way down.

![image](https://user-images.githubusercontent.com/10757813/105356080-3e940280-5bf3-11eb-8305-de3fae9c7f36.png)

ddooley:
It will be great to have a fix on this.  Any ETA?

AMBudnik:
Hi @ddooley, I wanted to reintroduce the issue to the team, but it seems that the situation has improved.

Could you please check this demo https://jsfiddle.net/14t6bnzs/

Here's my outcome on Chrome 142 (np zoom/scale) macOS 26.0. I'm able to see the option on every element with our without scrolling to the bottom of the table.

<img width="779" height="450" alt="Image" src="https://github.com/user-attachments/assets/8a7f92df-0336-4048-956c-a80faf7175c8" />

budnix:
The issue is fixed as mentioned by @AMBudnik in https://github.com/handsontable/handsontable/issues/6045#issuecomment-3630697303

ddooley:
Ah, I just noticed your response.  I followed the link you sent just now, but I'm not seeing that behaviour in my Chrome 144.0.7559.135  browser running Mac OS Sequoia 15.3 - the a,b,c,d menu doesn't show above input. Perhaps an upgrade to Chrome would get me there though, I'll see.

[...]

PR #13417, opened 2026-09-08
DEV-2825: Position dropdown editor lists against the viewport

### LINKED ISSUE #8688: Incorrect DropdownEditor's height if we set height as 'auto' for a table

budnix:
### Context

The PR fixes #8688: the dropdown editors' option lists are cut off by the grid's edge.

The lists of the `autocomplete`, `dropdown`, `handsontable`, and `multiselect` editors were positioned inside the grid's root element. Any set `height` makes that root clip its own content, so a list opened near the last row was cut off at the grid's edge. The editor tried to compensate by trimming the list to the rows that fit the space left inside the grid, which on a short grid left as few as two of eight choices visible. The issue has been open since 2021 with `priority: high`, and the standing advice on it was a workaround: do not set a `height`.

Measured on the exact grid from the fiddle in the issue (3 rows, one `dropdown` column, 8 options, `height: 'auto'`):

| | Before | After |
| --- | --- | --- |
| `getWorkspaceHeight()` | 117 | 800 |
| list height written | `58px` | `233px` |
| options a pointer can reach | 2 of 8 | 8 of 8 |

**Two mechanisms caused it, and both are keyed on the grid's own box:**

1. `overflow: clip` on the root element cut anything below the grid's bottom edge. This is the decisive one: with the height clamp disabled, the full-height list rendered and 6 of 8 options landed outside the clip, where `document.elementFromPoint()` did not even reach the editor. So no fix that keeps the list inside that root can work.
2. `flipDropdownVerticallyIfNeeded()` and `MultiSelectEditor#getAvailableSpace()` measured free space with `getWorkspaceHeight()`, so `limitDropdownIfNeeded()` clamped the list to the rows that fit inside the grid.

**Relationship to #13381.** That PR removes the clip for `height: 'auto'` and fixes the reported shape only. Measured on its head, two other shapes still showed 2 of 8: a fixed numeric `height`, and `height: 'auto'` inside a parent with a fixed height and `overflow: auto`. This PR closes those, so #8688 can be closed on it rather than left partly open against DEV-1656. #13381 has since squash-merged, and this PR targets `develop` and carries it directly.

### The fix

The lists are positioned `position: fixed`. They stay children of the grid root in the DOM, so focus, ARIA, and outside-click handling are untouched, but a `fixed` box is laid out outside the normal flow, so neither the root's clip nor a scrolling ancestor can cut them.

**A `fixed` box resolves against the viewport only while no ancestor is its containing block.** Any host-page `transform`, `translate`, `rotate`, `scale`, `perspective`, `filter`, `backdrop-filter`, `contain: paint|layout|strict|content`, `container-type`, or a `will-change` naming one of those properties or `contain` takes over, and `top`/`left` then resolve against that ancestor's padding box. A grid in a centered modal is the everyday case: positioned from raw viewport coordinates, the list landed **313px below and 384px across** from its cell there, with **1 of 8 options reachable**. Handsontable's own DOM sets none of those properties, so the host page decides. Every `will-change` value the code treats as establishing one was measured in Chrome to anchor a fixed child; `will-change: container-type` does not, and is left out.

`getFixedContainingBlockRect()` (`helpers/dom/element.ts`) returns the box the list is actually laid out in, falling back to the viewport so callers use it unconditionally. Every coordinate, the flip decision and the clamp resolve against it. A list taller than that box is trimmed to fit and scrolls, because CSS gives a fixed box no way out of such an ancestor. The bound is never `innerHeight`: that counts scrollbar gutters and, on mobile, the area under browser chrome and the software keyboard, which is up precisely because a text editor has focus.

**The list follows its cell.** A `fixed` box adds no scrollable overflow, so a list that stays where it opened ends up past the box's edge with its options out of reach. On every scroll that moves the cell - the grid's own, the page's, or an ancestor's - and on a window resize, the list is moved to the cell, its vertical flip is decided again, and its height is clamped again.

- A capture-phase document `scroll` listener handles the page and its ancestors. It skips the grid's own scroll, which the `afterScroll*` hooks already handle and which fires before the holder moves. Both listeners guard `hot.isDestroyed`.
- `MultiSelectEditor` routes both paths through one `#followCell()`. It decides the vertical flip both ways (filtering keeps the flip as it is), keeps the list's own scroll position across the clamp (a wheel past the list's last entry scrolls the grid or the page), and skips the clamp while the edited row is not rendered, which is when `refreshDimensions()` hides the list.
- `MultiSelectEditor#destroy()` tears down its event manager. The scroll listener is the first one that manager holds; without the teardown it outlives every grid, and `MemoryLeakTest` counted 88 of them.
- `MultiSelectEditor` keeps the horizontal flip from DEV-1198 (#13548): the flipped offset is relative to the grid root like `start`, so it goes through the same containing-block conversion.

Four decisions worth reviewing:

- **Only the vertical axis moved off the grid's workspace.** The horizontal flip deliberately still measures `getWorkspaceWidth()`. #8688 is a vertical defect, a list kept inside the grid's width is inside the viewport anyway, and widening it would only let the list hang off the grid's side. Making both axes viewport-based broke 8 legacy specs that pin the sideways placement, for no user-visible gain.
- **Coordinates resolve against the editor holder's rect, not the cell's.** `TextEditor#refreshDimensions()` has already placed the holder at exactly the offset the old `absolute` rules resolved against, border compensation included. That compensation is conditional: `BaseEditor#getEditedCellRect()` cancels a 1px shift for a cell that draws its own inline-start border, which depends on row headers and on which columns are rendered. A hardcoded `cell.left - 1` lands the list a pixel off on some columns and not others. `MultiSelectEditor` derives it the same way.
- **The same physical property the old rules wrote is written:** `left` under LTR, `right` under RTL, resolved against the containing block's right edge. Mirroring RTL into a `left` value has to assume the holder is exactly as wide as the cell, which put the flipped list 2px off its edge. The two properties also carry opposite signs, since `right: offset` against the holder moves the list's right edge left.
- **The list is not portaled.** The context menu escapes the clip by portaling into `rootPortalElement` with document coordinates, which makes the containing-block problem impossible. Moving the list out of the grid root breaks outside-click detection, focus management and ARIA, so that stays scoped as DEV-1656. The one known gap points at it: a page reflow that moves the grid without scrolling or resizing anything leaves the list where it was.

The traps are documented in `handsontable/AGENTS.md`, `tests/AGENTS.md`, and the `handsontable-editor-dev` skill.

### Breaking change

A list can now paint outside the grid's box, and whether it opens above or below the cell follows the space left in **the box the list is laid out in** - the browser window, or the nearest ancestor that is the containing block for a fixed element - rather than the space left inside the grid.

```js
// Before: on a short grid with a set height, the list flipped above the cell to fit
// inside the grid, and was trimmed to the rows that fit - as few as two choices.
// After: the list opens downwards, overhangs the grid's bottom edge, and shows every
// choice that fits on screen. It still flips above the cell when the space below it in
// that box is too short, and is still trimmed when the box itself cannot hold the list.
height: 150
columns: [{ type: 'dropdown', source: eightOptions }]
```

Sideways placement is unchanged. `visibleRows` still caps the list. A list still paints inside its ancestor stacking context, so a host that puts the grid in its own layer keeps its own chrome on top. Migration guide: section `## 18.` of the 18.1 to 19.0 guide.

Five legacy vertical-flip specs encoded "the list must stay inside the grid" and are updated to the new contract, with a comment on each pointing at the Playwright spec that now covers the flip.

- Unit tests added/modified: `handsontable/src/helpers/dom/__tests__/element.unit.ts` - 14 cases for `getFixedContainingBlockRect()`: the viewport fallback; the padding box of an ancestor with `transform`, `filter`, `contain: paint`, or a `will-change` naming `transform`, `translate`, `rotate`, `scale`, `backdrop-filter` or `contain`; a `contain` value and a `will-change` hint that must NOT establish one; crossing a shadow boundary; and the nearest of two nested ancestors. jsdom drops `perspective`, `backdrop-filter` and `container-type` from its cascade, so those three have no unit case.
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`):
  - `tests/e2e/dropdown-editor-clip.spec.ts` (new), with its own fixture `tests/fixtures/demo/dropdown-editor-clip.html` and page object `DropdownEditorClipPage.ts`; its in-page calls are typed in `windowTypes.ts`. 12 cases: four where the list is no longer bounded by the grid (the reported `height: 'auto'` shape, a fixed `height` that clips its root, `height: 'auto'` inside a scrolling parent, a `multiselect` grid); two containing-block ancestors (a `transform`ed modal, `contain: paint`); and six on following the cell (a page scroll, the flip decided again on a page scroll, a `multiselect` list clamped again on the grid's own scroll, its flip cleared once the cell has room below again, no error when a page scroll unrenders its row, and the list's own scroll position kept). Every reachability check hit-tests each option with `document.elementFromPoint()` rather than counting rendered cells, because under a clip an option exists at full size while nothing on screen can reach it, so `toBeVisible()` and `toHaveCount()` both pass on the broken build.
  - `tests/e2e/dropdown-height.spec.ts` and `tests/fixtures/pages/DropdownHeightPage.ts` (modified): #8872's flex-squeezed grid no longer trims the list, so the spec asserts every option renders; the page object's clipping-ancestor walk stops at a `fixed` element and clamps to the viewport, or it intersects the option with the grid root and calls a readable option clipped.
  - Legacy Jasmine edits: `autocompleteEditor.spec.js` (2 cases), `handsontableEditor/__tests__/positioning.spec.js` (1), `multiSelectEditor.spec.js` (1, parametrized twice).
- Type tests (`*.types.ts`): none - no public API change.
- For a bug fix - the spec that fails without this fix: the eight #8688 cases in `dropdown-editor-clip.spec.ts`. The four `multiselect` follow cases guard the follow itself: without their fixes, three fail on all three themes (0-1 of 8 options reachable after a grid scroll; the flip still set after the cell returns to the top; `Cannot read properties of null (reading 'getBoundingClientRect')`), and the scroll-position case fails (`200` → `0`) when only the scroll restore is removed.
- Demo page / recorded trace: two local demo pages, released 18.1.0 from the CDN against this build. Five grids, each with an 8-option list in the last row, counted by hit-testing every option:

| Case | Grid | Released | This PR |
| --- | --- | --- | --- |
| A | `height: 'auto'`, 3 rows (the reported shape) | 2 of 8 | 8 of 8 |
| B | fixed `height: 150` | 2 of 8 | 8 of 8 |
| C | `height: 'auto'` in a 130px scrolling parent | 3 of 8 | 8 of 8 |
| D | `multiselect`, fixed height | 1 of 8 | 8 of 8 |
| E | inside a `transform`ed ancestor | 2 of 8 | 8 of 8 |

RTL was also checked by hand in Chrome: the `right` property is written, the list's right edge aligns with the holder's to the pixel, and all 8 options are reachable.

### Commands run

Run on the head commit, after a full build of both bundles. The full unit, legacy E2E and Playwright suites run in this PR's CI.

```bash
cd tests && HOT_TEST_PORT=8127 npx playwright test \
  e2e/dropdown-editor-clip.spec.ts e2e/multiselect-open-left.spec.ts e2e/dropdown-height.spec.ts \
  --project=e2e-main --project=e2e-horizon --project=e2e-classic \
  --project=e2e-main-min --project=e2e-horizon-min --project=e2e-classic-min
```
```
126 passed
```

```bash
npm --prefix handsontable run test:e2e -- --testPathPattern=editors/multiSelectEditor
npm --prefix handsontable run test:e2e -- --testPathPattern=editors/multiSelectEditor --theme=horizon
```
```
95 specs, 0 failures
95 specs, 0 failures
```

```bash
npm --prefix handsontable run test:unit -- --testPathPattern=editors/
npm --prefix handsontable run test:unit -- --testPathPattern=helpers/dom/__tests__/element.unit
cd handsontable && ./node_modules/.bin/tsc --noEmit -p tsconfig.json
npm --prefix tests run lint
```
```
Tests: 37 passed, 37 total
Tests: 177 passed, 177 total
tsc: no errors
tests lint: 0 errors
```

### Types of changes

### Related issue(s):

1. DEV-2825
2. #8688 / DEV-271
3. #3828 (the `preventOverflow: 'horizontal'` shape, already closed; verified still correct here)
4. #8872 / DEV-1656 (the flex-squeezed grid, whose list is no longer trimmed)
5. #13381 / DEV-2789 (the base branch)
6. #13548 / DEV-1198 (the `multiselect` horizontal flip, combined with the `fixed` positioning)

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

- [ ] MANUAL QA NEEDED — 

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2825

---

> [!NOTE]
> **Medium Risk**
> Breaking, user-visible editor placement for four common cell types across scroll, modals/transform ancestors, and RTL; mitigated by broad unit/E2E coverage and migration notes.
> 
> **Overview**
> Fixes **#8688** by changing how `autocomplete`, `dropdown`, `handsontable`, and `multiselect` option lists are placed: lists stay in the grid DOM but use **`position: fixed`**, so grid `overflow: clip` and scrolling parents no longer cut them off.
> 
> **Placement logic** now measures vertical space against the viewport (or the nearest CSS containing block via new **`getFixedContainingBlockRect()`**), flips above/below based on that box instead of grid workspace height, clamps list height to free space (including base `HandsontableEditor`), and repositions on page/ancestor scroll and resize. Horizontal flip still uses grid workspace width. Coordinates are derived from the editor holder rect; LTR/RTL use `left`/`right` as before.
> 
> **`MultiSelectEditor`** gains scroll-follow (`#followCell`), bidirectional vertical flip in `DropdownController`, document scroll/resize listeners with destroy guards, and **`eventManager.destroy()`** to fix listener leaks. **`AutocompleteEditor`** trims/restores height using wanted height on scroll and gates duplicate height writes.
> 
> **Breaking:** lists can extend past the grid; vertical flip follows viewport space, not “fit inside grid.” Documented in migration guide **§19**, changelog **8688**, and updated legacy/E2E specs (adjacency assertions, `elementFromPoint` reachability, new `dropdown-editor-clip` Playwright suite).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2dc7a16c373ecedb2769511c229b950cc1b49a4c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2825 Dropdown editor lists are cut off by the grid's edge](https://app.clickup.com/t/9015210959/DEV-2825)
Task linked: [DEV-271 incorrect `DropdownEditor`'s height if we set height as 'auto' for a table](https://app.clickup.com/t/9015210959/DEV-271)

demtario:
Related ClickUp backlog task: [DEV-419](https://app.clickup.com/t/86c950xfh) — *Opening dropdown on fixed row opens wrong way* (public [#6045](https://github.com/handsontable/handsontable/issues/6045)).

Same family as this PR: vertical flip / available-space for dropdown editors on fixed rows (and when the list would open the wrong way / get clipped). Worth tracking against this change when it lands.

demtario:
Related ClickUp: [DEV-419](https://app.clickup.com/t/86c950xfh) — *Opening dropdown on fixed row opens wrong way* (public [#6045](https://github.com/handsontable/handsontable/issues/6045)).

DEV-419 itself is being closed as fixed via `flipDropdownVerticallyIfNeeded()` / #6045. Residual list clipping when the root height clips the dropdown list is tracked by this PR (#8688 / DEV-2825).

qunabu:
Task linked: [DEV-1198 When not enough space multiselect should open to the left](https://app.clickup.com/t/9015210959/DEV-1198)

handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
```diff
@@ -763,12 +764,20 @@ export class AutocompleteEditor extends HandsontableEditor {
...
+      // clamp it to that box - both of which need the NEW height, and neither of which a
+      // hand-edited `top` can express. Reading the flag rather than re-deciding: the caller has
+      // just decided the flip from the untrimmed height, and re-deciding here on the trimmed one
+      // could disagree with the space figure it passed in.
       if (this.isFlippedVertically) {
-        this.htEditor.rootElement.style.top =
-          `${parseInt(this.htEditor.rootElement.style.top, 10) + dropdownHeight - height}px`;
+        this.flipDropdownVertically();
+      } else {
+        this.unflipDropdownVertically();
```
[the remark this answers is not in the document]
budnix: The growth half is fixed in aa76a2040e: `limitDropdownIfNeeded()` now measures `getTargetEditorHeight()` rather than the list's current height, so a list trimmed at the edge grows back when the cell gains room again — a new case in `dropdown-editor-clip.spec.ts` failed on all three themes before it. Not changing the `scrollTop` half: the sub-grid keeps its holder scroll across that `updateSettings()`. Measured on `main` and `horizon` with the list parked at `scrollTop` 40 and a page scroll that re-clamped it (main's height went 116 → 145) — the scroll read 40 both times.

handsontable/src/editors/autocompleteEditor/autocompleteEditor.ts
```diff
@@ -763,12 +749,24 @@ export class AutocompleteEditor extends HandsontableEditor {
       const rowsThatFit = Math.max(Math.ceil(spaceAvailable / rowHeight) - 1, 1);
       const height = rowsThatFit * rowHeight;
 
-      if (this.isFlippedVertically) {
-        this.htEditor.rootElement.style.top =
-          `${parseInt(this.htEditor.rootElement.style.top, 10) + dropdownHeight - height}px`;
-      }
-
```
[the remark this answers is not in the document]
budnix: The trim half is fixed in 20f2f70d91: `setDropdownHeight()` now shares the base editor's `appliedDropdownHeight` gate, and `updateDropdownDimensions()` records the height it writes so the next trim to that same number is not skipped as a false match. A trimmed `dropdown` list was writing the identical height twice per scroll event — 20 writes of `203` across 10 page scrolls, now 1, and that one is a real change.

Not changing the restore path. Its check compares `getDropdownHeight()` against the height the list WANTS, not against a height just written, so the pixel or two of `offsetHeight` drift does not defeat it: 235 against a wanted 233 reads false and the branch does not run. Measured 0 height writes across 10 scrolls on an untrimmed list, for both `dropdown` and `handsontable`.

docs/content/guides/upgrade-and-migration/migrating-from-18.1-to-19.0/migrating-from-18.1-to-19.0.md
```diff
@@ -872,3 +872,56 @@ if (typeof value === 'string') {
...
+- One ancestor property changes where the list is anchored, and the grid reads it for you. An
+  element with any of these becomes the box the list is laid out in, instead of the browser
+  window:
+
+  - `transform`, `translate`, `rotate`, or `scale`
+  - `perspective`
+  - `filter` or `backdrop-filter`
+  - `contain` set to `paint`, `layout`, `strict`, or `content`
+  - `container-type`
+  - `will-change` naming any of the above, or naming `contain`
```
demtario: **medium**: this bullet reads as "`will-change` naming any of the above", and "the above" includes `container-type` (the line directly above). But the code deliberately excludes it: `FIXED_CONTAINING_BLOCK_WILL_CHANGE` in `helpers/dom/element.ts` omits `container-type`, with a comment stating that a `will-change` naming it does not establish a containing block. In a breaking-change guide this misleads anyone debugging why their `will-change: container-type` modal fails to anchor the list. Reword to exclude it, e.g. "`will-change` naming any of the properties above except `container-type`, or naming `contain`."

handsontable/AGENTS.md
```diff
@@ -122,6 +122,7 @@ Gotcha: Filters `conditionCollection` uses physical indexes, `getDataAtCol()` us
...
... the horizontal flip deliberately still measures `getWorkspaceWidth()`, because a list inside the grid's width is inside the viewport anyway. Coordinates resolve against the editor HOLDER's rect, not the cell's, so the conditional 1px border compensation in `getEditedCellRect()` is preserved, and they are written to the same physical property the old rules used (`left` under LTR, `right` under RTL, which carry opposite signs). Escaping the clip does **not** guarantee painting above neighbouring page content: a `fixed` box still paints inside its ancestor stacking context, and `.ht-root-wrapper.ht-shadow-dom` carries `isolation: isolate` precisely to keep an embedded grid below its host's chrome. A document `scroll` listener follows the cell, skips the grid's own scroll (which the `afterScroll*` hooks already handle) and MUST guard `hot.isDestroyed`; an editor gaining its first `eventManager` listener must also have a `destroy()` that tears the manager down (`MultiSelectEditor` did not, and leaked 88). `HandsontableEditor#htContainer` is the sub-grid's own `rootElement`, so the flip methods and `applyRootSize()` write the same element. Full note: the `handsontable-editor-dev` skill.
```
demtario: **medium**: this containing-block list is out of sync with the code it cites (`helpers/dom/element.ts`) in two ways. It drops `translate`, `rotate`, and `scale` (all three are in `FIXED_CONTAINING_BLOCK_PROPS` next to `transform`), and its will-change clause ("naming one of those") omits `contain`, which `FIXED_CONTAINING_BLOCK_WILL_CHANGE` includes. Align it with the two constants; the `SKILL.md` bullet already states this correctly and can be copied.

handsontable/src/helpers/dom/element.ts
```diff
@@ -1145,6 +1145,165 @@ function elementTrapsAxis(el: HTMLElement, axis: OverflowAxis, rootWindow: Windo
...
+  const willChange = style.willChange ?? '';
+
+  if (FIXED_CONTAINING_BLOCK_WILL_CHANGE.some(token => willChange.split(/[\s,]+/).includes(token))) {
+    return true;
+  }
+
+  // A container-query container contains its layout, which has the same effect.
+  const containerType = (style as unknown as Record<string, string>).containerType ?? '';
+
+  return containerType !== '' && containerType !== 'normal';
```
demtario: **medium**: `container-type` is matched with a denylist (`!== '' && !== 'normal'`), while the sibling `contain` check above uses an allowlist. `container-type: scroll-state` (Chrome 133, above the Chrome-130 floor) sets up a scroll-state query container but applies no layout containment, so it does not create a containing block for `position: fixed` descendants. This treats it as one anyway, so the list anchors to that ancestor's padding box instead of the viewport, reintroducing the offset-drift this PR fixes. `size`/`inline-size` are correctly caught; only `scroll-state` (and any future non-`normal` value) is wrong. Fix: tokenize and allowlist `size`/`inline-size`, mirroring the `contain` handling. Low real-world likelihood today, so non-blocking.

.changelogs/8688.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "public",
+  "title": "Fixed the `autocomplete`, `dropdown`, `handsontable`, and `multiselect` editors' lists being cut off by the grid's edge or by a scrolling parent, and trimmed to as few as two choices to fit inside it: the lists are now positioned against the box they are laid out in - the browser window, or the nearest ancestor that is the containing block for a fixed element - so every choice that fits on screen is shown. A list can therefore paint outside the grid's box, and whether it opens above or below the cell now follows the space left in that box rather than the space left inside the grid.",
```
demtario: **medium**: the title is 107 words / 588 characters, roughly 3x every other current `breaking: true` entry (35-52 words). It buries the one-sentence headline (lists are positioned against the viewport / containing block instead of absolute-inside-grid, so nothing is cut off) under mechanism detail that duplicates migration-guide section 19 almost verbatim, making the rendered `CHANGELOG.md` / release entry hard to scan. Trim to one tight sentence and leave the mechanics to section 19.

**low**: the same line uses plain hyphens as clause separators (`...laid out in - the browser window...` and `...for a fixed element - so...`). DOC-STANDARDS rule 11 lists changelog entries as non-site text that takes en dashes (–); the hyphen convention is docs-site only. Replace both.

tests/fixtures/pages/DropdownEditorClipPage.ts
```diff
@@ -0,0 +1,348 @@
...
+
+  /**
+   * Records every height written to the list's own sub-grid from now on.
+   *
+   * The scroll follow re-clamps the list on every scroll event, and the free space usually changes
+   * by less than one row between two of them - so the clamp keeps landing on the height already
+   * applied. Each such write is a full sub-grid `updateSettings()` render that changes nothing,
+   * which is why the editors gate the write on the height they last wrote.
+   */
+  async startListHeightWriteRecorder(): Promise<void> {
```
demtario, line 286: **medium**: this recorder is what pins the height-gate fix (the test asserts no two consecutive `updateSettings({height})` writes repeat), and the author's revert-and-rerun confirms it fails without the gate. But it has no positive control that the patch is actually intercepting anything: it only checks that scroll events fired (`scrollCount() > 0`). If a later refactor renames `htEditor`, moves the sub-grid reference, or routes the height write through a different call, the patch silently stops capturing, `htListHeightWrites` stays empty for an unrelated reason, and the test passes forever having verified nothing. Add a positive control: an early full-row scroll step expected to change the trimmed height, asserting `expect(written.length).toBeGreaterThanOrEqual(1)` before the steady-state no-repeat loop.

demtario (commented):
9 inline notes below.

handsontable/src/helpers/dom/__tests__/element.unit.ts
```diff
@@ -2064,4 +2065,163 @@ describe('DomElement helper', () => {
...
+    }
+
+    it('should fall back to the viewport when no ancestor establishes a containing block', () => {
+      const parent = document.createElement('div');
+      const child = document.createElement('div');
+
+      parent.appendChild(child);
+      document.body.appendChild(parent);
+
+      const rect = getFixedContainingBlockRect(child);
```
demtario, line 2102: **low**: this fallback test's four assertions pass identically whether the function returns the viewport or wrongly returns `parent`'s padding box. In jsdom `documentElement.clientWidth/Height` are 0 and an un-stubbed `parent.getBoundingClientRect()` is all zeros, so the test named for the fallback cannot fail on it. (The fallback branch itself is distinguishingly covered by the `contain: size` and `will-change: opacity` negatives, so this is test-robustness, not a coverage hole.) Fix: `stubRect(parent, { top: 100, left: 200, width: 400, height: 300 })` before the call and keep the `top === 0` / viewport-size assertions, matching those two negatives.

tests/fixtures/pages/windowTypes.ts
```diff
@@ -146,6 +146,14 @@ export interface FixtureHotInstance {
     isOpened(): boolean,
     beginEditing(): void,
     finishEditing(restoreOriginalValue?: boolean): void,
+    /** The `<td>` currently being edited. Throws once the grid is destroyed. */
+    getEditedCell(): HTMLTableCellElement,
+    /** The option list's container, on the `handsontable` / `autocomplete` / `dropdown` family. */
+    htContainer?: HTMLElement,
```
demtario: **low**: `htContainer?: HTMLElement` was added to `getActiveEditor()`'s return type but is never read anywhere in `tests/`; meanwhile `DropdownEditorClipPage` re-derives the same node through a page-wide `document.querySelector('.handsontableEditor')`. Either drop the unused field, or switch the page object to read `editor.htContainer` directly (it is the same node).

docs/content/guides/upgrade-and-migration/migrating-from-18.1-to-19.0/migrating-from-18.1-to-19.0.md
```diff
@@ -872,3 +872,56 @@ if (typeof value === 'string') {
   hot.setDataAtCell(0, 1, value.trim());
 }
 ```
+
+## 19. Dropdown editor lists are no longer confined to the grid
+
+The lists of the [`autocomplete`](@/guides/cell-types/autocomplete-cell-type/autocomplete-cell-type.md),
```
demtario, line 878: **low**: `handsontable` and `multiselect` are in plain backticks here while `autocomplete` and `dropdown` are linked to their guide pages, though both target pages exist. Link all four consistently: `[\`handsontable\`](@/guides/cell-types/handsontable-cell-type/handsontable-cell-type.md)` and `[\`multiselect\`](@/guides/cell-types/multiselect-cell-type/multiselect-cell-type.md)`.

docs/content/guides/upgrade-and-migration/migrating-from-18.1-to-19.0/migrating-from-18.1-to-19.0.md
budnix: Good point — I took `container-type` off that list entirely in 2dc7a16c37, since it does not anchor the list in any current engine, so the `will-change` line no longer points at it.

handsontable/AGENTS.md
budnix: Agreed — synced it with the two constants in 2dc7a16c37, and dropped `container-type` there too.

handsontable/src/helpers/dom/element.ts
budnix: Good spot, and it reached further than `scroll-state`: measured in Chromium, Firefox, and WebKit, no `container-type` value makes a containing block, `size` and `inline-size` included, so I removed the check rather than allowlisting in 2dc7a16c37. It was putting the list 208px off its cell inside any container-query wrapper.

.changelogs/8688.json
budnix: Fair point on both — trimmed it to 48 words with a pointer to the guide, and switched to an en dash, in 2dc7a16c37.

tests/fixtures/pages/DropdownEditorClipPage.ts
budnix: You're right, an empty list would have passed. In 2dc7a16c37 the recorder now throws if it finds no sub-grid, and the test takes a 60px step and checks a write was recorded before the no-repeat check.

handsontable/src/helpers/dom/__tests__/element.unit.ts
budnix: Nice catch — stubbed the parent's rect in 2dc7a16c37, and checked it: the test now fails when the helper wrongly returns the parent's box.

tests/fixtures/pages/windowTypes.ts
budnix: Yep, my mistake — dropped `htContainer` in 2dc7a16c37 and declared the `htEditor` the recorder actually reads.

docs/content/guides/upgrade-and-migration/migrating-from-18.1-to-19.0/migrating-from-18.1-to-19.0.md
budnix: Good call — linked all four in 2dc7a16c37, in that sentence and in the intro, which had the same gap.

demtario (approved):

PR #13418, opened 2026-09-08
DEV-2786: Let the column header own the gridline below it at every scroll position

demtario:
### Context

The PR fixes the row axis of the grid's border ownership, the twin of what #13371 settled on the column axis for issue #6673. It is the last of three parts; #13393 and #13399 are merged.

With column headers on, the first rendered body row was given a 1px `border-top` on top of the `border-bottom` every cell carries, and `box-sizing: border-box` took both out of the same row height. So that one row's content box was 1px shorter than every other row's: the reported symptom is a first row that renders taller or shorter than its neighbours, `[30, 29, 29, 29]` where `[29, 29, 29, 29]` was declared.

The column header answered by dropping its own `border-bottom` at vertical offset 0 so the seam did not double, and the engine gave it back (`innerBorderTop`) the moment the grid scrolled, with matching compensations in the scroll target, the hider height, the layout snapshot's scrollbar prediction and the row-height sums. Measured on `main` before the change:

| reading | at offset 0 | scrolled to row 12 |
|---|---|---|
| `.ht_master` classes | none | `innerBorderTop` |
| header `th` `border-bottom` | 0px | 1px |
| thead height / `getColumnHeaderHeight()` | 29 | 30 |
| `getViewportHeight()` | 271 | 270 |
| hider height | 1770 | 1771 |
| first rendered body `tr` | 30 | 30 |
| second body `tr` | 29 | 29 |

So the header, and the whole table, changed height by being scrolled. And because that 1px shift landed after the overlay positions had been computed, every crossing of vertical offset 0 paid a reconciliation draw: `ctx.positionChanged` triggered `wtOverlays.refreshAll()`, a nested `wot.draw(true)` over the master and every clone.

After this change the gridline below the last head row belongs to that header at every scroll position, no body row abutting a header draws a `border-top`, every row has the same content height, and the header's height no longer depends on the scroll offset. `ctx.positionChanged` has no contributors left, so the flag, the reconciliation branch, `prepareHeaderBorders`, `columnHeaderBorderCompensation`, `expandHiderVerticallyBy` and `isScrolledBeyondHiderHeight` are all deleted.

**Two things a reviewer should read before the diff.**

**1. The CSS gate is `thead:not(:empty) + tbody`, not the `:not(.htColumnHeaders)` the approved plan named.** The plan proposed gating on the root's `htColumnHeaders` class, mirroring the `htRowHeaders` gate the column axis uses. That turns out to be wrong here, and the difference is not cosmetic. Every overlay clone carries its own `thead`, empty unless that clone renders the head row, so the per-table form matches exactly the tables whose first body row abuts a header: the master and the inline-start clone (no frozen top rows), the top clone and the top corner (frozen top rows, where the frozen row is the one under the header). The bottom clones' `thead` is always empty, so their first row keeps its `border-top`. There the pixel is the bottom-freeze seam, not a header seam, and the grid-wide gate would have taken it away. It also makes `fixedRowsTop` work with no extra rule. A spec pins each half.

**2. `columnHeaderHeight` semantics move by a pixel, exactly as `rowHeaderWidth` did in #13371.** The value is written as a border-box height, and the header now carries a `border-bottom` it did not carry at offset 0, so `columnHeaderHeight: 40` yields a 38px content box where it used to yield 39 at offset 0 and 38 once scrolled. The option is now consistent instead of scroll-dependent, which is the point, but it is a visible change for anyone who tuned the number against the unscrolled grid.

**Deliberately out of scope: a grid with no column headers keeps its 1px first-row border.** There the border is the grid's own top frame, and the +1px size compensation survives, now gated on the header being absent rather than applied unconditionally. That mirrors #13371 exactly, which left "no row headers means column 0 stays 1px narrower" out of scope and pinned it as a control case. Both are pinned as control cases here too.

**`colHeaders: false` does not mean "no head row", and the predicate had to learn that.** NestedHeaders replaces the column-header renderers through the documented `afterGetColumnHeaderRenderers` hook whenever it holds a layer, and never consults `colHeaders`. Such a grid renders a `thead`, so the CSS rule fires and its first body row draws no top border, while `hasColHeaders()` still reports `false`. Keyed on the setting, every row-height sum in the core would be a pixel long: measured on that shape, the row rendered at 29px while `getDefaultRowHeight(0)` returned 30. The predicate therefore reads the RENDERED header count (`view.getColumnHeadersCount()`), which is the same quantity the engine side already reads (the length of the resolved `columnHeaders` array), falling back to the setting only before the view exists. `ManualRowMove` routes through the predicate instead of calling `hasColHeaders()` itself. Covered by a fixture grid and a unit test, both red before the fix.

**The `+ 1` is now asked as a question, not answered four times.** `StylesHandler#getDefaultRowHeight`, `AutoRowSize` and `ManualRowMove`'s drop guideline each carried their own copy; they route through the new `StylesHandler#firstRenderedRowDrawsTopBorder()`. In the engine the same question is `getFirstRowBorderCompensation(externalRowCalculator, hasColumnHeaders)` in `axisSizing/boxModel.ts`, read by `SpreaderSize#adjustElementsSize` and by `gatherLayoutInput`. Those two must agree to the pixel or the grid predicts a scrollbar it does not get, which is how a StretchColumns spec failed mid-change on a layout snapshot describing a scrollbar the DOM never grew.

**Selection had to follow the pixel.** `.wtBorder` paints at z-index 10 and `.ht_clone_top` at 160, so a selection edge left centred on the shared gridline is geometrically right and completely hidden behind the header. `standsBelowColumnHeader()` is the row-axis twin of `standsBehindRowHeader` and moves the edge onto the cell's own boundary. Separately, the row-0 active-header accent used to come from that row's own `border-top`, which is now 0px wide, so `SelectionManager` tags the head row's corner cells instead.

**Two test-harness fixes ride along, both caused by this change.**

- `rtl-window-scroll.spec.ts` hardcoded `cell-79-29`; the band that reaches the document end moved by a row, so it resolved to nothing. It now reads the row out of the DOM, per the rule in `tests/AGENTS.md`.
- `frozen-column-row-heights.spec.ts` compared the master and the frozen-column clone straight after `scrollVerticallyTo(0)`, which waits only for `scrollTop` (satisfied the moment it is assigned) while the redraw it triggers is coalesced into a later animation frame. Under full-suite load that sampled the two tables mid-draw. Now polled.
- `fill-handle-formulas.spec.ts` grabbed the fill handle at its geometric centre, and the 1px shift moved that centre into Chromium's overlay-scrollbar strip, which takes no layout space and so cannot be subtracted from any bounds arithmetic. `FormulasGridPage` now asks `elementFromPoint` for a point the browser will actually deliver the press to; the full reasoning is in the code comment.

### How has this been tested?

New Playwright coverage in `tests/e2e/column-header-border-ownership.spec.ts` (15 tests) over a 9-grid fixture: plain column headers, `fixedRowsTop`, `fixedRowsBottom`, nested headers, `autoRowSize`, no-headers control, no-rows, two row headers, and `preventOverflow: 'horizontal'` (the one shape the engine never stamped `innerBorderTop` on, so it was the shape where nothing gave the header a bottom border at any offset).

Every assertion is geometric, and it has to be: `--ht-cell-vertical-border-color` resolves to `--ht-border-color` in `main`, `horizon` and `classic` alike, so a colour assertion on this seam would be true on all six legs whichever element drew it. For the same reason the accent test's negative control asserts the class, not the colour.

Per-property negative controls, both run:

- Reverting only the CSS fails 8 of the 15.
- Reverting only the selection files fails exactly the 3 selection and accent tests.

```bash
npm run build --prefix handsontable        # a CSS change is not applied until the bundle is rebuilt
npm run test:e2e --prefix handsontable     # the full suite, not a subset
cd tests && npx playwright test            # all six theme x bundle legs
```

Unit coverage added for both new predicates: `StylesHandler#firstRenderedRowDrawsTopBorder` (both branches, plus that `getDefaultRowHeight` stops compensating with headers on) and `getFirstRowBorderCompensation` (all four input combinations).

**One number in `walkontable/test/spec/overlay.spec.js` moves and is worth reading rather than skimming.** At the window-scroll bottom edge the master's `bottom` goes from `documentClientHeight` to `documentClientHeight - 1`. That is not a tolerated overshoot: `top` is unchanged and the table is genuinely one pixel shorter, because its first body row no longer draws a `border-top`. The neighbouring `newY += 1` in `TopOverlay#scrollTo` is deliberately untouched; it is row-border accounting, not header-border accounting, and removing it fails 22 walkontable specs including "sticks to the bottom edge (without headers)", which lands a row short. The spec carries that reasoning inline.

**And the window-scroll axis owner now has its own Playwright coverage**, in `tests/e2e/column-header-window-scroll.spec.ts`. The element-scroll fixture cannot reach it: every grid there declares a `width` and a `height`, so the browser clamps `scrollTop` to the holder's range and an overshooting scroll target is absorbed before anything can observe it. The new fixture declares neither, so the window owns both axes and nothing clamps. It pins the user-visible contract there: a bottom-edge snap reveals its row rather than clipping it, the scroll still reaches the very last row, and the header's height does not move when the window scrolls. Its own header comment is explicit that it does NOT guard the `newY += 1` term (it passes with that term removed, because in window mode the pixel changes only how far past flush the target lands, not which row is revealed); the walkontable scroll specs guard that.

**Review cost to expect: every visual screenshot with column headers moves**, the mirror of what #13371 did to every shot with row headers. That is the single largest cost in this PR and the reason it lands on its own.

### Types of changes

### Related issue(s):

1. DEV-2786

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

- [x] I have signed the [Contributor License Agreement](https://cla.handsontable.com/sign) ; one signature covers both Handsontable and HyperFormula; the `cla/signed` check on this PR confirms it.

---

> [!NOTE]
> **Medium Risk**
> Changes core Walkontable rendering, hider/viewport math, and documented `columnHeaderHeight` sizing semantics; risk is mitigated by broad e2e/unit coverage but any custom layout tuned to the old scroll-dependent header pixel may need a 1px adjustment.
> 
> **Overview**
> **Column headers now keep their bottom border at every scroll position**, mirroring the column-axis fix in #6673. CSS uses a per-table `thead:not(:empty) + tbody > tr:first-child` rule so the first body row under a header no longer draws an extra `border-top`, which stops the first data row from rendering 1px taller than its neighbors and stops **table/header/hider height from changing when you scroll**.
> 
> The Walkontable draw cycle **drops the scroll-driven `innerBorderTop` / `innerBorderBottom` layout trade**: `ctx.positionChanged`, `prepareHeaderBorders()`, nested `refreshAll()` reconciliation draws, `columnHeaderBorderCompensation`, `expandHiderVerticallyBy`, and scroll-based hider height tweaks are removed. **`getHiderHeightCompensation` returns 0 when column headers are rendered**, and row measurement (`oversizedRows`, `RenderSizeProbe`) only subtracts the first-row top border on tables with no head row.
> 
> **`columnHeaderHeight` is documented and applied consistently as border-box** (includes the header’s 1px top and bottom borders), so the same value no longer differs by 1px between offset 0 and scrolled. Call sites that used to special-case the first body row route through **`StylesHandler#firstRenderedRowDrawsTopBorder()`** (also used by AutoRowSize and ManualRowMove). **Selection borders and active-header accents** are adjusted so edges and highlights stay visible when the shared gridline sits under the top overlay (`standsBelowColumnHeader`, corner seam classes).
> 
> Docs/changelog and a large test refresh cover overlay dimensions, scroll bands, resize helpers (`getColumnHeaderBandHeight`), and the no–nested-draw behavior on legacy layout paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bd181baa074f11af28de2b3f85d38849a0ee769e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:

handsontable/src/renderSizeProbe.ts
```diff
@@ -169,8 +169,14 @@ export class RenderSizeProbe {
...
         // Only single-row header cells carry a content-driven per-level height.
+        // `outerHeight`, not `innerHeight`: the value measured here goes back into the
+        // `columnHeaderHeight` funnel and `adjustColumnHeaderHeights` writes it as `style.height` on
+        // a `box-sizing: border-box` header cell, so it has to be a border-box height or the round
+        // trip loses the cell's own borders. It loses exactly one pixel since DEV-2786, where the
+        // header gained a permanent `border-bottom` - which left the master's content-driven header
+        // row a pixel taller than the height written to the frozen overlays' corner cell.
         if (cell instanceof HTMLTableCellElement && cell.rowSpan <= 1) {
-          const cellHeight = innerHeight(cell);
+          const cellHeight = outerHeight(cell);
```
budnix, line 180: The JSDoc four lines up still says this method matches `Table.markIfOversizedColumnHeader`, which reads `innerHeight` of each header cell, so it now describes the opposite of what the code does. That name is also gone from the engine - the closest things I can find are `adjustColumnHeaderHeights` and `syncOversizedColumnHeadersWithFrozenOverlays`. Worth rewriting that first paragraph, because a doc block this precise is the one the next person will trust over the code.

handsontable/src/3rdparty/walkontable/src/axisSizing/hiderCompensation.ts
```diff
@@ -45,6 +45,25 @@ export function getHiderHeightCompensation(wtSettings: Settings): number {
     return 0;
   }
 
+  // The pixel the declared `1` below stands for is the FIRST rendered body row's own `border-top`,
```
budnix, line 52: The doc block on this function says the `1` is the cell bottom border that the internal row-height calculator misses, and this comment says it is the first rendered body row's `border-top`. Only the second reading makes the early return below correct, so someone who starts from the JSDoc will read this as a bug and try to revert it. Would be good to bring the doc block in line so both tell the same story.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
```diff
@@ -191,6 +176,15 @@ function runMasterDrawCycle(table: Table, ctx: DrawContext): void {
...
       restoreRenderedStateIfSafe(table, wtViewport, renderedStateBeforeDraw, renderCycleSeqBeforeHook);
+      // The only path through a master draw that never reaches `wtOverlays.refresh()`, which is what
+      // sizes the master hider/spreader on every other path (the per-overlay `adjustElementsSize()`
+      // calls in `placeFixedOverlays` size the clone roots, not the master hider). The bands this
+      // reads are current whether or not the cell render ran, so without it a skipped draw keeps a
+      // stale scrollbar until an unrelated full draw. It used to be reached from the
+      // `ctx.positionChanged` branch below, which DEV-2786 removed — and only on the draws where an
+      // `innerBorder*` class happened to toggle, so the coverage is now complete rather than
+      // incidental.
+      wtOverlays.adjustElementsSize();
```
budnix, line 187: This makes every skipped render size the master hider, where before only the draws that toggled a border class did. The Playwright leg that covers it, "leaves the master sizes alone when the crossing draw skips its render", still carries a comment saying a skipped draw now takes the plain else with no sizing call - true before this change, not after. The assertions there still hold, so it is only the note that needs a line.

handsontable/src/utils/stylesHandler.ts
```diff
@@ -167,6 +168,41 @@ export class StylesHandler {
...
+   * per table: an overlay clone that renders no head row keeps the border on its own first row, and
+   * `axisSizing/oversizedRows` resolves that from each table's `THEAD`.
+   *
+   * It asks how many header rows are actually RENDERED, not whether `colHeaders` is on, because the
+   * two disagree: `afterGetColumnHeaderRenderers` lets a plugin add head rows to a grid that
+   * declared none, and NestedHeaders does exactly that whenever it holds a layer, without consulting
+   * `colHeaders`. Such a grid has a non-empty `thead`, so the CSS rule fires and its first body row
+   * draws no top border, while `hasColHeaders()` still reports `false` - and every caller then
+   * compensates for a pixel that is not there. The engine side asks the same question the same way
+   * (`getFirstRowBorderCompensation` takes the length of the resolved `columnHeaders` array), which
```
Haxikowy: `getFirstRowBorderCompensation` doesn't exist anywhere in the tree — `boxModel.ts` only exports `getRowBorderCompensation`/`getBoxAdjustedRowHeight`. The engine-side twin is `getHiderHeightCompensation` in `hiderCompensation.ts`; fix the pointer or the next "don't add a fifth copy" search comes up empty.

handsontable/src/3rdparty/walkontable/AGENTS.md
```diff
@@ -245,18 +248,41 @@ Consequences worth knowing:
...
+  (frozen top rows). The bottom clones' `thead` is always empty, so their first row keeps its
+  `border-top` — there the pixel is the bottom-freeze seam, not a header seam, and a grid-wide gate on
+  the root's `htColumnHeaders` class would have taken it away. A grid with no column headers keeps the
+  border everywhere for the same reason: it is the grid's own top frame, deliberately left as-is, the
+  mirror of column 0 staying 1px narrower on a grid with no row headers.
+- **The `+ 1` that goes with it is asked as a question, not answered per call site.** It used to be
+  reimplemented independently in `StylesHandler#getDefaultRowHeight`, `AutoRowSize` and
+  `ManualRowMove`'s drop guideline; they all now route through
+  `StylesHandler#firstRenderedRowDrawsTopBorder()`. In the engine the same question is
+  `getFirstRowBorderCompensation(externalRowCalculator, hasColumnHeaders)` in `axisSizing/boxModel.ts`,
```
Haxikowy: Same stale pointer as `stylesHandler.ts`: `getFirstRowBorderCompensation`/`axisSizing/boxModel.ts` don't exist. Rename to `getHiderHeightCompensation` in `hiderCompensation.ts` — this doc's whole job here is stopping a fifth copy.

.changelogs/13418-changed.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "The `columnHeaderHeight` option now resolves to the same height at every scroll position. It is a border-box height that includes the header's 1px top and bottom borders, so `columnHeaderHeight: 40` leaves a 38px content box instead of 39px on an unscrolled grid.",
+  "type": "changed",
+  "issueOrPR": 13418,
+  "breaking": false,
```
Haxikowy: This changes `columnHeaderHeight`'s resolved value for anyone who tuned it against an unscrolled grid, with no deprecation path or migration step — a literal read of `.ai/BREAKING-CHANGES.md`'s "changing API behavior" row calls that breaking. Mirrors the same call on `rowHeaderWidth` in #13371, so probably intentional precedent, but worth an explicit sign-off here too.

.changelogs/13418.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Fixed the first row rendering a pixel taller than the rest, and the grid's height changing when it was scrolled, on grids with column headers.",
+  "type": "fixed",
+  "issueOrPR": 13418,
+  "breaking": false,
+  "framework": "none"
+}
```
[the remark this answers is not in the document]
demtario: Agreed, and this is now covered. It landed in `49af2d090`, which postdates the `80be8dfd` this review ran against:

- `docs/content/guides/columns/column-header/column-header.md` gained a paragraph on the border-box semantics.
- The `columnHeaderHeight` JSDoc in `handsontable/src/dataMap/metaManager/metaSchema.ts` says the same thing at the option itself.
- `.changelogs/13418-changed.json` is a second entry, type `changed`, and its title names `columnHeaderHeight` explicitly with the `40` to 38px example. It is separate from `13418.json` because a file carries one `type` and `type` picks the section.

handsontable/src/renderSizeProbe.ts
demtario: Good catch, fixed in `3573ada9b`.

Confirmed `markIfOversizedColumnHeader` appears nowhere in the tree except that sentence, so the pointer was dead as well as wrong. The block now names `adjustColumnHeaderHeights` as the consumer and says the value is a BORDER-BOX height, with the `outerHeight`-versus-`innerHeight` reasoning left where it belongs, at the measurement itself.

handsontable/src/3rdparty/walkontable/src/axisSizing/hiderCompensation.ts
demtario: You are right, and you picked the correct one of the two: the pixel is the first rendered body row's `border-top`, which is what makes the early return correct. Fixed in `3573ada9b`.

The doc block now states that, names both consumers, and lists the two cases that return 0 in the order the code checks them.

I also renamed `DECLARED_BOTTOM_BORDER_WIDTH` to `DECLARED_UNCOUNTED_GRIDLINE_WIDTH`, since the old name was carrying half the confusion on its own. The paragraph about the sub-pixel inflation still speaks of the cells' bottom border, and deliberately so: that is the border the code actually measures, as a proxy for how much the browser inflates any declared `1px`. The whole pixel and the inflation are now clearly two different things in the text.

handsontable/src/3rdparty/walkontable/src/table/drawCycle.ts
demtario: Correct on both counts, and the assertions do still hold. Fixed in `3573ada9b`.

Two comments in that spec were stale, not one. The first said a skipped draw takes the plain `else`; it now says the branch was removed and the call hoisted into the skip path, so every skipped draw sizes the master. The second said dropping the extra `adjustElementsSize()` was safe; nothing was dropped, so it now says the now-unconditional call has to be a no-op on a horizontal crossing, which is what makes running it every time safe.

Re-ran that spec on all six theme and bundle legs after the edit: 78 passed.

handsontable/src/utils/stylesHandler.ts
demtario: Right, that pointer went stale when the function was folded into the engine's own compensation module during the develop merge. Fixed in `7f987049f`.

It now reads `getHiderHeightCompensation` in `axisSizing/hiderCompensation.ts`, which is the function that takes the resolved `columnHeaders` length. Grepped the tree afterwards: no occurrence of the old name is left outside the generated `walkontable/dist` bundle.

handsontable/src/3rdparty/walkontable/AGENTS.md
demtario: Fixed in `7f987049f`, same commit as the `stylesHandler.ts` one.

The bullet now names `getHiderHeightCompensation(wtSettings)` in `axisSizing/hiderCompensation.ts` and says it reads the `externalRowCalculator` and `columnHeaders` settings itself, rather than taking them as arguments the way the old signature did. Agreed on why it matters: a "do not add a fifth copy" note is worth nothing if the search for the fourth one comes up empty.

.changelogs/13418-changed.json
demtario: Fair challenge, and I am deliberately not resolving this one myself. Leaving `breaking: false` as it stands and flagging it for an explicit call.

What the change actually is, so the sign-off is made on the facts: the total header height is unchanged. What moves is which element paints the shared gridline, so a declared `columnHeaderHeight` now resolves to the same content box at every scroll offset instead of being a pixel taller while the grid sat at offset 0. Put another way, the old value was not stable to begin with: the same setting produced 39px unscrolled and 38px scrolled. Anyone who tuned against the unscrolled figure was tuning against a number that changed under them as soon as the user scrolled.

That is why I read it as a 1px behavior correction rather than an API break, and why the entry is `changed` rather than `breaking`. It is the same shape as the `rowHeaderWidth` call in #13371, so the precedent points the same way, but you are right that precedent is not sign-off.

Happy to flip `breaking` to `true` and add a migration-guide step if the team wants the stricter reading. Whoever makes the call, could you record it here so the next person reading `.ai/BREAKING-CHANGES.md` against this row finds the reasoning attached to a decision rather than to an assumption.

Haxikowy (dismissed):

budnix (approved):

demtario:
`visually-approved` label: every one of the 755 diffs is the intended row-axis border-ownership move, in which the column header takes over the gridline below it, so the seam repaints and body content shifts by one pixel while the headers themselves and the bottom frozen rows stay untouched.

Haxikowy (approved):

PR #13420, opened 2026-09-08
DEV-2650: Build the context menu check mark as a DOM node

budnix:
### Context

The PR fixes the last surface where Handsontable wrote its own markup into an `innerHTML` sink, so the grid now needs no Trusted Types policy at all for markup it authors itself.

A menu item that shows a check mark built the mark as an HTML string and baked it into its own label:

```js
`<span class="selected">${String.fromCharCode(10003)}</span>${label}`
```

`menuItemRenderer` then wrote that label through `fastInnerHTML`, which is a Trusted Types sink. On a page under `require-trusted-types-for 'script'` with no `default` policy and no policy-backed `sanitizer`, opening the menu threw and took the whole menu down:

```text
TypeError: Failed to set the 'innerHTML' property on 'Element':
           This document requires 'TrustedHTML' assignment.
```

This is the grid's own markup, not user data, so the "no policy needed" claim DEV-2617 shipped did not cover it.

**Six call sites, not the two the task named.** There were two helper functions, and `markSelected()` in `customBorders/utils.ts` is a byte-for-byte copy of `markLabelAsSelected()` in `contextMenu/utils.ts`. A bordered selection threw exactly the way a read-only one did.

| Helper | Items |
| --- | --- |
| `markLabelAsSelected()` | `contextMenu/predefinedItems/readOnly.ts`, `comments/contextMenuItem/readOnlyComment.ts` |
| `markSelected()` | `customBorders/contextMenuItem/{top,bottom,left,right}.ts` |

**The fix — the task's option 2.** A menu item carries its state on the item instead of inside the label string:

1. `menu/utils.ts` gains `checked?: boolean | (() => boolean)` on `MenuItemLike` plus an `isItemChecked(item, hot)` predicate that resolves it the way `isItemDisabled` does — comparing against `true`, so a truthy string is never mistaken for a state.
2. `menu/menuItemRenderer.ts` writes the label, then prepends a `<span class="selected">` built with `createElement` + `textContent`. Order matters: `fastInnerHTML` replaces whatever the wrapper holds, so a span added before it would be wiped. The span still ends up first in the DOM.
3. The six items return a plain `name()` and declare `checked()`.
4. Both string helpers are now unused. They are **kept as legacy exports** rather than deleted: their modules ship a `.d.ts` next to them in `tmp/`, so `moduleResolution: node` resolves them whatever the package `exports` map says, and a caller cannot be assumed not to exist. That is what keeps this PR non-breaking.
5. **Declaring `checked` also makes the item checkable**, so it is announced as `menuitemcheckbox` and carries `aria-checked`. This is not cosmetic — see the accessibility note below.

**`checked` is also a new public option** on a custom menu item, added to `MenuItemConfig`, to the context menu guide's property table, and to the type test.

**The rendered DOM is unchanged** — `[span.selected, text]` inside `div.htItemWrapper`, same class, same character. The four legacy positioning specs that measure `span.selected`'s offset are the regression check on that.

#### The accessibility half, which the first round of this PR got wrong

Only `make_read_only` declared `checkable`. The other five items would have drawn a visible tick as a plain `role="menuitem"` with no `aria-checked` — and `aria-checked` is invalid on a plain menu item, so their checked and unchecked states would have been **indistinguishable to a screen reader**. That is worse than what they did before this PR, where the marker at least reached the accessible name (as raw markup, since a non-checkable item takes its `aria-label` from the item value).

So `isItemCheckable()` now also returns `true` for an item declaring `checked`, and the checkbox branch labels from `ariaLabel ?? itemValue` (these five declare no `ariaLabel`, and reading one unconditionally would have written the string `"undefined"` as the accessible name). `readOnly.ts` drops its now-duplicate `ariaLabel` and its `ariaChecked`, computing the read-only check once instead of three times.

The test is by **declared type**, not by presence and not by resolved value. By presence, someone's existing item carrying an unrelated property named `checked` (a string, say) would change role on upgrade. By resolved value, an item would be a checkbox only while it happened to be checked, and unchecking it would drop the `aria-checked="false"` that conveys the state.

#### Why not the other two options the task listed

Option 1 (a sentinel the renderer recognizes inside a string) keeps a marker inside a string, which is the shape DEV-2617 spent nine surfaces removing. Option 3 (a CSS `::before`) deletes the `<span class="selected">` element that consumers may style or assert on, which the breaking-changes policy discourages and which four in-repo specs read.

#### Also in this PR

`checkSelectionBorders()` moved from `getCellMeta` to `getCellMetaTransient`. It walks the whole selection on every menu draw, and the materializing read retained one meta object per visited cell for the grid's life — the pattern the core `AGENTS.md` bans for exactly this shape. Pre-existing and unchanged in frequency — it ran from `name()` before and runs from `checked()` now — but fixed while here, because `readOnly.ts` in this same PR uses the transient read and the two sibling helpers should not disagree.

**The three places that stated the limitation are updated together**, as the task required: the Trusted Types section of `docs/content/guides/security/security/security.md`, the `.changelogs/13244.json` entry title, and the DOM-templates rule in `handsontable/AGENTS.md` (which went from "three sinks survive" to two). Editing the 13244 entry is safe — it is not in `CHANGELOG.md`, so it has not shipped. The trap itself is recorded in `handsontable/src/plugins/contextMenu/AGENTS.md`.

Two changelog entries, which `.changelogs/README.md` allows only because they land in different sections: `13420.json` (`fixed`, the Trusted Types bug) and `13420-added.json` (`added`, the new `checked` option). A new public option buried in the Fixed section would be missed by anyone scanning for new API.

### How has this been tested?

Two new unit suites:

- `menu/__tests__/menuItemRenderer.unit.js` (19 cases) — the predicate, the DOM the renderer builds, that the label never reaches the sanitizer, the custom-renderer branch, and the whole `aria-checked` / role derivation including both narrowing decisions above.
- `menu/__tests__/checkedMenuItems.unit.js` (25 cases) — imports the six real item factories and asserts, per item, that it reports checked and unchecked correctly, is announced as a checkbox, and keeps markup out of its label. Without this a `checked` block dropped from any of the five non-`readOnly` items would ship green: the legacy positioning specs only ever query `Read only`, and the renderer suite builds synthetic items.

In `tests/e2e/trusted-types.spec.ts`, the case that asserted `MENU-THREW` now asserts `MENU: 1 checkmark` plus `expectNoViolations()`, under real CSP enforcement with no `default` policy and no Handsontable policy in the allowlist. It was flipped rather than deleted, as the task asked. The count is load-bearing: a fix that dropped the mark, or rendered the label but lost the span, fails there rather than passing as "no violation". The two menu cases moved out of the `surviving sink: header content` block — the menu is no longer a sink — and folded into one test that runs with and without a sanitizer and asserts the two agree, which is the actual user-visible claim. Keeping them separate would have left a title-only duplicate that stayed green even if the sanitizer plumbing were deleted.

```bash
# Unit

# Types

# Legacy E2E — the four plugins this touches
npm run test:e2e --prefix handsontable -- --testPathPattern="contextMenu|dropdownMenu|customBorders|comments"

# Playwright — the CSP enforcement suite, all theme x bundle legs
cd tests && npx playwright test e2e/trusted-types.spec.ts
```

Results on this branch:

| Gate | Result |
| --- | --- |
| `tsc --noEmit` + `test:types` | clean |
| `eslint` / `stylelint` | 0 errors |
| `test:unit` | 4594 passed, 357 suites |
| legacy E2E, the four plugins | 1020 specs, 0 failures |
| legacy E2E, full suite | 0 failures |
| Playwright `trusted-types.spec.ts` | 96 passed (theme x bundle legs) |
| `scripts/__tests__/changelog-published-entries.test.mjs` | 18 passed |

### Types of changes

### Related issue(s):

1. DEV-2650
2. DEV-2617 / #13244 — the change this completes

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2650

---

> [!NOTE]
> **Medium Risk**
> Reserved `checked` on custom menu items can change role and visuals on upgrade; behavior is documented but easy to miss. Core menu/CSP path is well tested; impact is limited to context/column menus and TT-enforced pages.
> 
> **Overview**
> Fixes the last Handsontable-authored markup path that hit a Trusted Types sink: context menu check marks for **Read only**, **Read-only comment**, and border sides no longer embed `<span class="selected">` in the item `name`. Those six built-ins now expose plain labels plus a **`checked`** option (`boolean` or function); **`menuItemRenderer`** writes the label through **`fastInnerHTML`**, then prepends the tick with **`createElement`**, so the menu works under **`require-trusted-types-for 'script'`** without a policy or sanitizer for that mark.
> 
> **`checked`** is also public API on context/column menu items. Declaring it (when the value is boolean or function) draws the mark, sets **`menuitemcheckbox`** / **`aria-checked`**, and can collide with custom items that already used a property named **`checked`**—the migration guide documents renaming or dropping duplicate marks in **`name`**.
> 
> Docs and changelogs drop the “context menu still needs TT” caveat; **`markLabelAsSelected`** / **`markSelected`** remain as legacy exports. Border selection checks use **`getCellMetaTransient`** during menu draws to avoid materializing meta for every cell in the range.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d62a77d678841aa72be21c9d207a50a8c4ad51a9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2617 Trusted Types: documented support does not work, and cannot be a default sanitization strategy](https://app.clickup.com/t/9015210959/DEV-2617)
Task linked: [DEV-2650 Context menu still builds markup as a string: markLabelAsSelected reaches innerHTML](https://app.clickup.com/t/9015210959/DEV-2650)

demtario (commented):
4 inline notes below.

.changelogs/13420-added.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Added a `checked` option to a context menu and column menu item. It draws the check mark Handsontable uses for its own toggles and announces the item as a checkbox to assistive technology. It takes a boolean or a function returning one, called with the Handsontable instance as `this`. Note that `checked` is now a reserved item property: a custom item that already carries one holding a boolean or a function is treated as a checkbox from now on.",
```
demtario: **medium** The upgrade note covers the ARIA half of `checked` becoming reserved, but not the visible half: the renderer now also draws a `span.selected` into such an item.

`isItemCheckable()` returns `true` for a boolean or function `checked` (`menu/utils.ts:209-211`), and once `isItemChecked` resolves `true` the renderer inserts the mark (`menuItemRenderer.ts:113-126`). So an existing item shaped `{ name: () => '<span class="selected">✓</span>My toggle', checked: myState }` gets a **second** mark prepended on top of its own. Drawing the tick inside `name` was the documented route: `context-menu.md:384` still says `name` "Supports HTML", and it is what the public `markLabelAsSelected()` export did.

Two things to add:

1. Extend this title to say the item is announced as a checkbox **and** drawn with the check mark, so a reader knows to rename the property if the mark is unwanted.
2. Add a step to `docs/content/guides/upgrade-and-migration/migrating-from-18.1-to-18.2/migrating-from-18.1-to-18.2.md`. That guide already carries the sibling #13244 work as sections 4, 5 and 6, so it is the file this change skipped.

tests/e2e/trusted-types.spec.ts
```diff
@@ -228,4 +207,30 @@ test.describe('Trusted Types enforcement', () => {
...
   });
+
+  test.describe('the converted sink: the context menu check mark', () => {
+    // DEV-2650. This block asserted a throw until the check mark stopped being a `<span>` baked
+    // into a menu item's label: the label reached `fastInnerHTML` carrying the grid's own markup,
+    // so a read-only selection took the whole menu down under enforcement. The item now declares
+    // `checked` and the renderer builds the span with `createElement`, so the label is plain text
+    // and never reaches the sink.
+
+    test('marks an item as checked whether or not a sanitizer is configured', async () => {
```
demtario, line 218: **medium** The five items this PR newly turns into `menuitemcheckbox` (the four `customBorders` items plus `commentsReadOnly`) have no assertion in a real browser, for either the role and state or the check mark itself.

- The fixture this case drives configures `contextMenu: ['make_read_only']` only (`tests/fixtures/demo/trusted-types.html:228`), and its status string counts `span.selected` inside the first `.htItemWrapper` (`:239`).
- `expectNoViolations()` reports CSP violations, not axe results (`tests/fixtures/pages/TrustedTypesPage.ts:127-131`), so it asserts nothing about ARIA.
- The existing browser-level a11y specs assert `menuitemcheckbox` for `Read only` alone and pin the count at 1 (`contextMenu/__tests__/a11y/attributes.spec.js:78-95`, `dropdownMenu/__tests__/a11y/attributes.spec.js:64-82`), and the four positioning specs that measure `span.selected` query `Read only` only.
- `checkedMenuItems.unit.js` does reach these five, but its stub returns one meta object for every cell (`:29-37`), so it pins the direction key names rather than the renderer integration.

Suggest one Playwright case that opens a menu with `customBorders` and `comments` enabled on a bordered and read-only-comment selection, asserting `role="menuitemcheckbox"`, `aria-checked`, and the `span.selected` count on those items.

handsontable/src/plugins/comments/contextMenuItem/readOnlyComment.ts
```diff
@@ -11,12 +11,12 @@ export default function readOnlyCommentItem(plugin: Comments) {
   return {
     key: 'commentsReadOnly',
     name(this: HotInstance): string {
-      const label: string = this.getTranslatedPhrase(C.CONTEXTMENU_ITEMS_READ_ONLY_COMMENT);
-      const areReadOnly = checkSelectionConsistency(this.getSelectedRange() ?? [], (row: number, col: number) => {
+      return this.getTranslatedPhrase(C.CONTEXTMENU_ITEMS_READ_ONLY_COMMENT);
+    },
+    checked(this: HotInstance) {
+      return checkSelectionConsistency(this.getSelectedRange() ?? [], (row: number, col: number) => {
         return !!(plugin.getCommentMeta(row, col, META_READONLY));
```
demtario: **low** This walk still goes through a materializing meta read, unlike the two siblings the PR converted.

`plugin.getCommentMeta()` resolves through `this.hot.getCellMeta()` (`comments.ts:923`), and `checkSelectionConsistency` breaks early only on a hit (`contextMenu/utils.ts:133-152`), so a selection with no read-only comment is walked in full and retains one meta object per visited cell. `menuItemRenderer.ts:67` computes `isItemChecked` unconditionally, before the separator and custom-renderer branches, so it runs on every menu draw even though this item is disabled unless the highlighted cell has a comment.

Not a regression in frequency, since `name()` did the same walk on develop. It is the consistency gap the PR description names ("the two sibling helpers should not disagree") applied to the third sibling, and it is the pattern `handsontable/AGENTS.md` bans by default.

Fix: read the flag transiently here, either via a transient variant of `getCommentMeta` on `Comments` or inline as `this.getCellMetaTransient(row, col)[META_COMMENT]?.[META_READONLY]`. Worth dropping `getCellMeta` from the stub at `checkedMenuItems.unit.js:33` too, since returning the same object for both methods is why neither the switch nor a regression away from it is observable.

handsontable/src/plugins/contextMenu/__tests__/contextMenu.types.ts
```diff
@@ -88,6 +91,7 @@ new Handsontable(document.createElement('div'), {
         key: 'name',
         hidden: false,
         disabled: false,
+        checked: true,
```
demtario, line 94: **low** These two added lines compile identically with and without the new `checked?: boolean | (() => boolean)` declaration, so deleting the declaration would leave `test:types` green.

`MenuItemConfig` carries `[key: string]: unknown` (`contextMenu.ts:48`), so `checked: true` and `checked() {...}` were already assignable on develop. And these literals never reach `MenuItemConfig` at all: `contextMenu?: boolean | object | string[]` (`core/settings.ts:266`) checks them against bare `object`.

The explicit declaration takes precedence over the index signature, so pinning it works both ways: `const checkedItem: MenuItemConfig = { name: 'x', checked: true };` plus a `@ts-expect-error` on `checked: 'pending'`, which is also what makes the by-type rather than by-presence narrowing testable at the type level.

Noting the severity: the neighboring `hidden` and `disabled` lines at `:44-48` are equally unpinned, so this follows the file's existing pattern rather than introducing it.

.changelogs/13420-added.json
budnix: Good spot — the entry now says the mark gets drawn too, and section 7 of the 18.1 to 18.2 migration guide covers the rename and the double-mark case in 064ae691ac.

tests/e2e/trusted-types.spec.ts
budnix: You're right — added `context-menu-checked-items.spec.ts` in 064ae691ac, it reads role, `aria-checked` and the mark count on all five in a real menu.

handsontable/src/plugins/comments/contextMenuItem/readOnlyComment.ts
budnix: Nice catch — it reads through `getCellMetaTransient` now in 064ae691ac, and I dropped `getCellMeta` from the stub so a slip back would fail.

handsontable/src/plugins/contextMenu/__tests__/contextMenu.types.ts
budnix: Fair point, it proved nothing — pinned against `MenuItemConfig` with a `@ts-expect-error` in 064ae691ac; deleting the declaration now turns `test:types` red.

tests/fixtures/pages/ContextMenuCheckedItemsPage.ts
```diff
@@ -0,0 +1,110 @@
...
+  /**
+   * Navigate to the fixture. The theme and bundle travel as query params so the fixture loads the
+   * matching stylesheet and Handsontable build — miss either and the leg silently tests something
+   * other than what its name says.
+   */
+  async goto(): Promise<void> {
+    await this.page.goto(
+      `/tests/fixtures/demo/context-menu-checked-items.html?theme=${this.theme}&bundle=${this.bundle}`);
+    await expect(this.cell(0, 0)).toBeVisible();
+  }
```
demtario, line 57: **medium** `goto()` waits only on `expect(this.cell(0, 0)).toBeVisible()`, with no `await this.page.waitForFunction(() => 'Handsontable' in window)` first.

`tests/AGENTS.md` documents this exact failure mode: a fixture's own ready state does not prove the bundle loaded, and `expect` is the wrong tool for that wait. `dist/handsontable.js` is ~6MB uncompressed and every worker pulls its own copy, so a cold or busy server can outlast the 10s `expect` timeout (`tests/playwright.config.ts:85`) before the bundle even loads, while `waitForFunction` polls against the full 20s test budget instead. Every sibling page object added around this PR (e.g. `WidthWindowScrollPage.ts`, `IncrementalRenderPage.ts`) follows the `waitForFunction` pattern for this reason; this file is the outlier.

Fix: add `await this.page.waitForFunction(() => 'Handsontable' in window);` before the `toBeVisible()` call in `goto()`.

tests/fixtures/pages/ContextMenuCheckedItemsPage.ts
```diff
@@ -0,0 +1,110 @@
...
+  readonly page: Page;
+  readonly theme: string;
+  readonly bundle: string;
+  readonly grid: Locator;
+
+  constructor(page: Page, theme = 'main', bundle = 'umd') {
+    this.page = page;
+    this.theme = theme;
+    this.bundle = bundle;
+    this.grid = page.getByTestId('grid');
```
demtario, line 33: **nit** `readonly grid: Locator` is assigned here but never read - `cell()` builds its own `.ht_master` locator instead of scoping through `this.grid`.

Fix: drop the unused `grid` field, or route `cell()`/`goto()` through it if it was meant to scope lookups.

demtario (commented):
2 inline notes below.

tests/fixtures/pages/ContextMenuCheckedItemsPage.ts
budnix: Good spot on 064ae691ac — the develop sync added `awaitBundle()` there in dcf0988d3f, which is that wait plus the `{ polling }` the new lint rule needs.

tests/fixtures/pages/ContextMenuCheckedItemsPage.ts
budnix: Fair point — it reads it now for the `data-init-error` check in dcf0988d3f, and I routed `cell()` through it as you suggested in 4bd8287bf3.

demtario (dismissed):

.changelogs/13420-added.json
```diff
@@ -0,0 +1,8 @@
+{
+  "issuesOrigin": "private",
+  "title": "Added a `checked` option to a context menu and column menu item. It draws the check mark Handsontable uses for its own toggles and announces the item as a checkbox to assistive technology. It takes a boolean or a function returning one, called with the Handsontable instance as `this`. Note that `checked` is now a reserved item property: a custom item that already carries one holding a boolean or a function is now drawn with the check mark and announced as a checkbox. Rename the property to opt out, and drop any mark you drew inside `name` yourself or the item shows two. See the migration guide.",
+  "type": "added",
+  "issueOrPR": 13420,
```
demtario: Multiple changelogs of the same PR number

budnix: You're right — folded it into 13420.json in d62a77d678; `sync` and the republish check both key on the number alone, so the second file would have been dropped or blocked a release.

demtario (approved):

PR #13421, opened 2026-09-08
DEV-2769: Give the merged page fixtures an explicit polling interval

demtario:
### Context

The PR fixes a red `develop`. `Lint / core` currently fails on the trunk itself, so every pull request opened from it inherits the failure regardless of what it changes.

Two pull requests that were each green in isolation landed close together and conflict semantically, not textually:

- #13364 added the `waitForFunction() needs an explicit polling interval` ESLint rule.
- #13392 added four page fixtures under `tests/fixtures/pages/` that call `waitForFunction()` with no options argument.

Neither PR's CI could see the other, and git merged both cleanly, so the rule and its five violations arrived on `develop` together. Reproduced at `bd953c1dad`, the current trunk tip, with the exact command the failing job runs:

```
✖ 5 problems (5 errors, 0 warnings)
tests/fixtures/pages/AutoHeightScrollbarZoomPage.ts
tests/fixtures/pages/EditorOpenCheckboxFocusPage.ts
tests/fixtures/pages/IframeWidthWindowScrollPage.ts
tests/fixtures/pages/WidthWindowScrollPage.ts
```

The fix is the one the rule's own message prescribes: pass `undefined, { polling: 100 }` as the options argument at all five call sites, matching `IncrementalRenderPage.ts`, `HiddenInitRerenderPage.ts`, and `MergeCellsInitRendersPage.ts`. Four are the identical `'Handsontable' in window` bundle probe; the fifth is the iframe page's `ready` flag.

This is not a lint-appeasement change. The rule exists because the default polls on `requestAnimationFrame`, which parallel Playwright workers starve, so a healthy page times out with nothing wrong on it. These four fixtures are exactly the shape the rule was written for: each waits on a ~6 MB UMD bundle to finish parsing while every other worker pulls its own copy.

Found while chasing a `Lint / core` failure on #13419, a one-line docs change that touches no TypeScript at all.

`[skip changelog]`

### How has this been tested?

- The exact command the failing CI step runs, before and after, at the trunk tip:
  ```bash
  npm run lint
  # before: exit 1, 5 problems (5 errors, 0 warnings)
  # after:  exit 0, 0 polling errors
  ```
  The 1928 remaining warnings are the pre-existing `sleep()` advisories, which are warn-only and unchanged by this PR.
- No test behavior changes: `polling: 100` alters how often the predicate is evaluated, not what it asserts.

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
[...]

PR #13422, opened 2026-09-08
DEV-2769: Wait for the bundle through the shared helper, not an inline copy

demtario:
### Context

The PR corrects a shortcut taken in #13421. That PR fixed a red `develop` by giving five `waitForFunction()` calls an explicit polling interval, which cleared the lint error but wrote `{ polling: 100 }` inline five times over. #13364 had already shipped the right tool for it, `awaitBundle()` in `tests/fixtures/bundle.ts`, and `tests/AGENTS.md` says so plainly:

> Do not inline a copy: the lint catches a missing `{ polling }`, but only the helper keeps the value from drifting between page objects.

So the lint went green on code that violated the convention the lint exists to serve. Roughly ten page objects already call the helper; these four were the only ones left reaching for the literal, which is exactly the drift the helper was introduced to end.

Four `goto()` methods now call `awaitBundle(this.page)`. The helper carries the reasoning those inline comments were duplicating - why `waitForFunction` and not `expect` (the plain UMD bundle is ~6 MB and every worker pulls its own copy, so a cold server outlasts the 10s `expect` timeout), and why a timer interval and not the `requestAnimationFrame` default (parallel workers starve rAF callbacks, so the default times out on a healthy page) - so each site keeps only the part specific to its own fixture.

The fifth call is not a bundle wait: `IframeWidthWindowScrollPage` also waits on the fixture's own `ready` flag, which signals that the iframe's stylesheets applied and the first render ran. That one stays a direct `waitForFunction`, now with `BUNDLE_POLLING_MS` rather than a second copy of the number.

No behavior changes: `BUNDLE_POLLING_MS` is 100, the value the inline calls used.

`[skip changelog]`

### How has this been tested?

- The command whose failure started this thread, at the current trunk tip:
  ```bash
  npm run lint    # exit 0, 0 errors
  ```
  The 1928 remaining warnings are the pre-existing `sleep()` advisories, warn-only and untouched.
- The four converted `goto()` methods wait on the same predicate at the same interval as before, so the E2E legs that drive them are unaffected.

### Types of changes

### Related issue(s):

1. DEV-2769

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2769

---

> [!NOTE]
> **Low Risk**
> Test fixture page-object refactors only; no production or runtime behavior change.
> 
> **Overview**
> Replaces duplicated inline `waitForFunction` bundle waits in four Playwright page objects with the shared **`awaitBundle()`** helper from `tests/fixtures/bundle.ts`, matching the convention documented in `tests/AGENTS.md` and aligning the last holdouts with ~10 other page objects.
> 
> **`AutoHeightScrollbarZoomPage`**, **`EditorOpenCheckboxFocusPage`**, and **`WidthWindowScrollPage`** now call `awaitBundle(this.page)` in `goto()` instead of `{ polling: 100 }` on `'Handsontable' in window`. Comments at those sites defer polling/`expect` rationale to the helper.
> 
> **`IframeWidthWindowScrollPage`** uses `awaitBundle` for the bundle and **`BUNDLE_POLLING_MS`** for the fixture’s separate `ready` wait (still a direct `waitForFunction`), so the interval is not duplicated. Behavior is unchanged because `BUNDLE_POLLING_MS` is 100.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d98226eef7cc3d7b21ba8c4b275f406a4a62f705. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/fixtures/pages/AutoHeightScrollbarZoomPage.ts
+++ b/tests/fixtures/pages/AutoHeightScrollbarZoomPage.ts
@@ -1,4 +1,5 @@
 import { type Page, type Locator, expect } from '@playwright/test';
+import { awaitBundle } from '../bundle';
 
 /**
  * Page Object for the auto-height-scrollbar-zoom fixture (DEV-2525): a `height: 'auto'` grid with
@@ -29,12 +30,11 @@ export class AutoHeightScrollbarZoomPage {
       `/tests/fixtures/demo/auto-height-scrollbar-zoom.html` +
       `?theme=${this.theme}&bundle=${this.bundle}&zoom=${zoom}&rows=${rows}`
     );
-    // Wait for the bundle itself before anything else. The `document.write`-injected script and the
+    // Wait for the bundle itself before anything else: the `document.write`-injected script and the
     // block that constructs the grid are separate, so a page can look ready while `Handsontable` is
     // still undefined, and the failure then surfaces inside a later `page.evaluate` far from its
-    // cause. `waitForFunction`, not `expect`: the plain UMD bundle is ~6 MB and every worker pulls
-    // its own copy, which outlasts the 10s `expect` timeout on a cold server.
-    await this.page.waitForFunction(() => 'Handsontable' in window, undefined, { polling: 100 });
+    // cause. The helper owns the reasoning and the polling interval.
+    await awaitBundle(this.page);
     await expect(this.cell(0, 0)).toBeVisible();
     await expect(this.lastRow()).toBeAttached();
   }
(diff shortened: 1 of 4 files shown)


qunabu:

budnix (approved):

PR #13424, opened 2026-09-08
DEV-2827: Poll the boundary row's height after the scroll-back instead of reading it once

danielzytohoc:
### Context

The PR fixes the one Playwright flake in the DEV-2668 ledger that had not been understood: `tests/e2e/walkontable/frozen-column-row-heights.spec.ts`, "keeps the panes aligned when the band boundary moves onto an already-tall row", which failed on three unrelated branches across three legs in August, always at the last assertion after the scroll back to the top, with `Expected: 69, Received: 30`.

**What it is.** A read race in the test, not a rendering transient. `locator.boundingBox()` resolves the element in one round trip and reads its box in another, and Walkontable recycles the same `<tr>` nodes across a re-render. At the boundary the tall row is the band's first `<tr>`; after the scroll back that same node shows row 0. A node resolved before the scroll-driven draw and measured after it therefore reports a normal row's height (30) for the tall row (69). The original spec's two sequential reads reproduced this 13 times in 80 runs, while a synchronous in-page snapshot taken immediately afterwards showed 69/69 every time.

**What the engine does.** The draw that moves the band is one synchronous task and leaves both tables at the resting height: the frozen overlay is measured and the master's rows are re-sized inside that same draw (`syncOversizedRowsWithFrozenOverlays`, then the sync's own re-apply, then `applyToDOM`). Traces of every view-render hook and every DOM mutation batch around 148 scroll-backs never showed the two tables apart at any task boundary. No engine change is warranted. The first version of this PR blamed a "one draw later" transient in the engine; the review (thanks, @budnix) asked which draw path that was, and the answer is none – the description and the comment in the spec are corrected accordingly.

**The fix.**

- `FrozenTallCellPage` reads geometry inside one `evaluate` on the table's root (never recycled): `rowHeight()`, `rowOffsetWithinTable()`, and two paired readers, `rowHeights(row)` (master and overlay in one evaluation) and `rowOffsetDrift(rows)`. `settledTallRowHeight(row)` establishes a tall row's resting height by polling until both tables agree above a normal row's height.
- The boundary assertions are `expect.poll`s pinned to explicit values instead of compares between two reads: both tables at `restingHeight + 1` at the boundary (the band's first `<tr>` carries 1px of top border), `[0, 0, 0]` offset drift for the rows below, "agreed and tall" for the test that grows the cell at the boundary, and both tables at `restingHeight` after the scroll back.
- The finding is recorded where the next spec author looks: `tests/AGENTS.md` (Determinism) and the `handsontable-playwright-e2e` skill's determinism reference.

**Interaction with DEV-2786 (#13418, merged to develop while this PR was in review).** That change hands the gridline under a column header to the header, so with this fixture's column headers the band's first `<tr>` no longer carries a 1px `border-top` – at the boundary the tall row measures exactly its resting height on both sides (CI on the merge commit: 65/65 on `classic`, 69/69 on `main`). The first version of the pinned boundary assertion expected `restingHeight + 1`, the pre-DEV-2786 value both the review and I assumed; it now pins `restingHeight`, and the two boundary tests' premise comments describe the new border ownership (the flip survives only on head-less tables, the bottom clones). #13418 also touched this spec's final assertion (a poll on master = clone plus a single read); the merge keeps this PR's pinned, atomic version.

Unrelated to DEV-2718 (an AutoRowSize `clearCache()` case). `FrozenTallCellPage.scrollVerticallyTo()` still ends on `scrollTop` on purpose; the polls after it are what absorb a read that lands before the draw.

Test-only change; no shippable source touched. `[skip changelog]`

- Unit tests added/modified (`*.unit.js`): none — test-only change.
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`): `tests/e2e/walkontable/frozen-column-row-heights.spec.ts`, `tests/fixtures/pages/walkontable/FrozenTallCellPage.ts`.
- Type tests (`*.types.ts`) updated if public API changed: none.
- For a bug fix — the spec that fails without this fix: the original one-shot compare, reproduced locally at 3 of 150 runs (all six legs, `--repeat-each=25`) and, with a synchronous snapshot beside it, at 13 of 80 runs on `e2e-main` and `e2e-horizon` — every one of them a Playwright read of a recycled node while the DOM was already consistent.
- Demo page / recorded trace (for UI changes): none.

### Commands run

```bash
cd tests && npx playwright test e2e/walkontable/frozen-column-row-heights.spec.ts --repeat-each=5 --workers=6
```
`540 passed (49.9s)` — the whole file, all six legs (engine built from develop before DEV-2786).
```bash
cd tests && npx playwright test e2e/walkontable/frozen-column-row-heights.spec.ts -g "boundary|first row the master renders" --repeat-each=30 --workers=6
```
`360 passed (1.1m)` — the two boundary tests, all six legs (same engine).
```bash
cd handsontable && npm run build:styles && npm run build:themes-css && npm run build:themes-css.min && npm run build:umd && npm run build:umd.min
cd tests && npx playwright test e2e/walkontable/frozen-column-row-heights.spec.ts -g "boundary|first row the master renders" --repeat-each=10 --workers=6
cd tests && npx playwright test e2e/walkontable/frozen-column-row-heights.spec.ts --repeat-each=2 --workers=6
```
`120 passed (20.4s)` and `216 passed (20.6s)` — after the merge with develop, on an engine rebuilt from the current develop (with DEV-2786).
```bash
cd tests && npx eslint --ext .ts e2e/walkontable/frozen-column-row-heights.spec.ts fixtures/pages/walkontable/FrozenTallCellPage.ts
```
Clean.

### Types of changes

### Related issue(s):

1. DEV-2827

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

- [ ] MANUAL QA NEEDED — 

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2827

---

> [!NOTE]
> **Low Risk**
> Changes are limited to Playwright fixtures, one spec, and test-authoring docs; no runtime Handsontable behavior is modified.
> 
> **Overview**
> Fixes a flaky Walkontable E2E case where **paired `boundingBox()` reads** could straddle a re-render: recycled `<tr>` nodes made a tall boundary row look like a normal row (`Expected: 69, Received: 30`) even when the DOM was already consistent.
> 
> **`FrozenTallCellPage`** now measures geometry in a **single in-page `evaluate`** on stable roots (`rowHeight`, `rowOffsetWithinTable`), adds **`rowHeights`** / **`rowOffsetDrift`** for master+overlay in one shot, and **`settledTallRowHeight`** to poll until both panes agree above a normal row baseline. Missing rows return **`NaN`** instead of `0`.
> 
> **`frozen-column-row-heights.spec.ts`** boundary scenarios use **`expect.poll`** pinned to **`restingHeight`** (and zero drift) instead of comparing two sequential reads; comments are updated for **DEV-2786** column-header border ownership (0px compensation with headers, symmetry still asserted).
> 
> **`tests/AGENTS.md`** and the Playwright skill **determinism** reference document the “no two geometry reads across a draw” rule. Test-only; no product code changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 321fabe7bf0776a389436994e3376443b8bfecea. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2718 Row header (inline-start overlay) body row heights out of sync with master table — regression since v17.0.0](https://app.clickup.com/t/9015210959/DEV-2718)
Task linked: [DEV-2827 Reproduce the frozen-column-row-heights flake across all six legs: missing wait, or a dropped frozen row height?](https://app.clickup.com/t/9015210959/DEV-2827)

tests/e2e/walkontable/frozen-column-row-heights.spec.ts
```diff
@@ -199,11 +199,20 @@ test.describe('walkontable row heights with frozen columns', { tag: '@walkontabl
         .toBe(await wt.rowOffsetWithinTable(wt.inlineStartOverlay, row));
```
budnix: This offset compare and the height compare at 194-195 are both single reads taken right after the band moved onto the boundary, which is the same race the new poll below fixes. By the reasoning in the PR body a master-versus-overlay compare can also pass while both tables sit at the wrong height, and at the boundary nothing pins them to the expected `restingHeight + 1`. Was fixing only the scroll-back assertion deliberate, or did these simply never flake? The test above has the same pair at 169-170.

tests/e2e/walkontable/frozen-column-row-heights.spec.ts
```diff
@@ -199,11 +199,20 @@ test.describe('walkontable row heights with frozen columns', { tag: '@walkontabl
...
     }
 
-    // And back off the boundary again — the 1px must not be left behind.
+    // And back off the boundary again — the 1px must not be left behind. Read it web-first: a
+    // frozen-derived height reaches the master one draw AFTER the band moves (the master renders
+    // its band before the frozen overlays are measured, and `drawCycle.ts` documents that draw as
```
budnix: Can I ask which draw path this describes? In `drawCycle.ts` the frozen sync is followed by `applyToDOM` in the same full draw, and the "self-corrects on the next draw" line there is about the visible row range, which that same branch rebuilds before the draw ends. A fast scroll draw does skip the row-height twin (157-159), so a real lag is plausible, but for a different reason than the one written here. Worth a quick check, since this comment is what tells the next reader that no engine change is warranted.

tests/e2e/walkontable/frozen-column-row-heights.spec.ts
```diff
@@ -199,11 +199,20 @@ test.describe('walkontable row heights with frozen columns', { tag: '@walkontabl
...
+    // here failed 3 of 150 runs on two legs. Both heights are polled against the resting height, so
+    // a transient equality (both tables mid-update) cannot pass either.
     await wt.scrollVerticallyTo(0);
 
-    expect(await wt.rowHeight(wt.master, BOUNDARY_ROW))
-      .toBe(await wt.rowHeight(wt.inlineStartOverlay, BOUNDARY_ROW));
+    await expect.poll(async() => ({
+      master: await wt.rowHeight(wt.master, BOUNDARY_ROW),
+      overlay: await wt.rowHeight(wt.inlineStartOverlay, BOUNDARY_ROW),
+    })).toEqual({ master: restingHeight, overlay: restingHeight });
```
budnix: `restingHeight` now decides what both tables must reach, but it is still a single read at 187 taken right after `goto`. If the same lag you describe can hit that read, the captured value would be the provided height instead of the tall one - line 189 would then fail rather than pass wrongly, so it is a flake source and not a silent hole. Would it be worth polling until the master and the overlay agree before capturing it?

tests/e2e/walkontable/frozen-column-row-heights.spec.ts
danielzytohoc: Fair point, and it turned out to be the whole story (see the next thread). Fixed: the boundary compares are now `expect.poll`s pinned to explicit values — both tables at `restingHeight + 1` at the boundary, `[0, 0, 0]` offset drift for the rows below, and "agreed and taller than a normal row" for the test that grows the cell at the boundary (`settledTallRowHeight()`). Each pair is read in one evaluation (`rowHeights()`, `rowOffsetDrift()`), which matters more than the poll.

tests/e2e/walkontable/frozen-column-row-heights.spec.ts
danielzytohoc: You were right to ask: there is no such draw path, and the comment was wrong. I traced every view-render hook and every DOM mutation batch around the scroll-back over 148 runs — one synchronous full draw, and both tables at the resting height at every task boundary; the sync's re-apply and `applyToDOM` run inside it, exactly as you read it. The 30 came from the reader: `locator.boundingBox()` resolves the node in one round trip and reads its box in another, and Walkontable recycles the same `<tr>` across the re-render, so a node resolved as the boundary row before the draw is row 0 after it — 30 is a normal row's height. The two sequential reads reproduced it 13 times in 80 runs while a synchronous snapshot taken right after showed 69/69 every time. The page object now reads geometry inside one `evaluate` on the table root, the comment describes this, the PR description is corrected, and the finding is in `tests/AGENTS.md` (Determinism) and the skill's determinism reference.

tests/e2e/walkontable/frozen-column-row-heights.spec.ts
danielzytohoc: Yes — `restingHeight` is now `settledTallRowHeight(BOUNDARY_ROW)`: polled until the master and the inline-start overlay agree on a height above a normal row's, then captured. With the atomic reads the single read was no longer racing anything, but establishing the reference rather than catching it is the right shape regardless.

tests/fixtures/pages/walkontable/FrozenTallCellPage.ts
```diff
@@ -113,11 +113,57 @@ export class FrozenTallCellPage {
...
 
-    return box?.height ?? 0;
+      settled = master;
+
+      return master === overlay && master > normalHeight;
+    }, {
+      message: `row ${row} to settle at one tall height in both the master and the inline-start overlay`,
+    }).toBe(true);
+
+    return settled;
```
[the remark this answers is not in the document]
danielzytohoc: Good catch. `normalRowHeight()` reads row 1, which leaves the band after the boundary scroll, and `rowHeight()` answered 0 for an unrendered row, so the check degenerated to `> 0`. Two changes: `rowHeight()` and `rowHeights()` now return `NaN` for a row the table does not render (no comparison accepts it, so a missing row fails loudly instead of passing as 0), and `settledTallRowHeight(row, normalRow = row + 1)` baselines on an explicitly rendered normal row — the one after the tall row, rendered whenever the tall row is. Re-verified: the whole file 108/108 and the two boundary tests 120/120 across the six legs.

danielzytohoc:
Heads-up for the re-review: while this was in review, DEV-2786 (#13418) landed on develop and changed what the boundary test measures — under a column header the band's first `<tr>` no longer carries a 1px `border-top`, so the tall row at the boundary now measures exactly its resting height on both sides (65/65 on classic, 69/69 on main in CI). The pinned boundary assertion expected `restingHeight + 1` (the pre-DEV-2786 value), which failed two legs on the merge commit; it now pins `restingHeight`, and the two boundary tests' premise comments describe the new border ownership. The merge also kept this PR's pinned, atomic final assertion over #13418's compare-plus-single-read on the same lines. Re-verified on an engine rebuilt from current develop: 120/120 (boundary tests ×10) and 216/216 (whole file ×2) across the six legs.

qunabu:

budnix (approved):

PR #13425, opened 2026-09-08
DEV-2831: Close the strand-discard scope on alter()'s early returns

budnix:
### Context

The PR fixes a deterministic append: `alter()` opens the editor's structural-change scope and then leaks it on three of its four exits, so a filter running later in the same task commits a stranded editor and adds records that were never asked for.

`grid.alter()` calls `editorManager.suspendStrandDiscards()` as its first statement, and the matching `resumeStrandDiscards()` sat at the function's tail. **Four** exits leave the suspended region and three of them never reached it: a `beforeAlter` veto's `return`, the `insert_row_*` `maxRows` `return`, and the switch's default `throwWithCause()`. Any hook that throws does the same.

While such a scope is open, the next trimming change reads it as open, skips the unresolvable-editor discard in `#reconcileEditorWithIndexMaps()`, and that change's own selection work commits through the stale coordinates — which `applyChanges()` satisfies by appending records. This is the DEV-2739 append reopened through a different door.

The `suspendStrandDiscards()` self-heal timeout does not cover any of it. A zero-delay timer runs in the **next** macrotask, while the `Filters#filter()` that skips the discard runs in **this** one — the leak only ever needed the rest of the current task. So the timer healed nothing that mattered, while retaining one timeout handle per call for the instance's life (`_registerTimeout` never prunes `this.timeouts`). It is removed here.

The fix is a `try`/`finally` around `alter()`'s body, which covers all four exits at once, and the same bracket around `updateData()`'s scope — its `adjustRowsAndCols()` runs the row-creation hooks and its `selection.refresh()` the selection hooks, either of which can throw. The `finally` sits where the old resume did, before `adjustElementsSize()`/`render()`, so ordering on the success path is unchanged.

An earlier structural change is needed for any of this to bite, because neither a veto nor a throw writes an index map: the captured record survives and guides the repair on its own. The removal in the repro is what clears it and leaves the editor's stale coordinates as the only thing left to read.

#### Why DEV-2756 is also in here

This was found while investigating DEV-2756, which reported the same append shape for `updateData()` + `filter()` as nondeterministic (~50% "under load", `45 failed / 90`). **That nondeterminism does not exist.** The 45/90 was a stale minified bundle: `npm run build:umd` writes `dist/handsontable.js` only, so the three `-min` legs kept running the pre-fix code, which the ticket itself says appends deterministically. 3 legs × 15 repeats = 45. It also explains the other half of the report — the local gates are pinned to `e2e-main`, the bundle a partial build *does* refresh, so whole-suite runs were genuinely green.

So this PR carries three things: the `alter()` fix (DEV-2831), a regression pin for DEV-2756's `updateData()` invariant so it cannot silently break, and the stale-bundle trap written down in `tests/AGENTS.md` and the `handsontable-playwright-e2e` skill.

#### Notes for the reviewer

- **The `updateData()` pin is not vacuous, but it does not pin what you might expect.** Removing `updateData()`'s scope pair entirely still passes — with no scope open the reconcile discards immediately, which loses the edit but can never append, the safe direction. Replacing the inline `resumeStrandDiscards()` with a deferred one fails 3/3. The test therefore pins *"the scope closes inline"*, which is the actual invariant.
- **The `maxRows` exit has no test of its own, and that is now a coverage note rather than a risk.** With the `finally` there is no hand-written resume left to forget, and the throw test exercises the identical mechanism. For the record, I could not construct a repro that reaches the `maxRows` exit with a live strand: making the counts meet needs `updateSettings({ maxRows })`, which re-captures the edited record so the filter takes the ordinary rebind branch, and a column-axis strand instead gets rescued by `selection.shiftColumns()`.
- The `return`s inside the `remove_row` / `remove_col` per-group `arrayEach` callbacks are not `alter()` exits and need nothing.
- **`alter()`'s body is reindented by two spaces.** That is most of the diff. `git diff -w` shows the real change.
- **One assertion is deliberately absent from the `updateData` test**, and the reason is worth knowing: after the discard, `getActiveEditor()` is `null` on the plain UMD bundle and a freshly prepared editor at row 0 on `full.min`. Asserting the editor reference therefore splits along the bundle axis on a verified full build — a genuine content difference, not staleness. The test asserts the observable outcome instead, and `tests/AGENTS.md` records both shapes.

#### Two things in here that came from develop, not from me

`develop` was merged in (not rebased — the branch is bound to this PR) because **DEV-2812 / #13404 landed 9 minutes before this branch's `Visual / Compare` ran**. It re-measures row heights, this branch did not have it, and exactly one screenshot of 1646 differed as a result. The merge makes the comparison apples-to-apples.

That merge also brought a **pre-existing `Lint / core` error on develop**: `tests/fixtures/pages/AutoRowSizeClearCachePage.ts`, added by #13404, calls `page.waitForFunction()` with no explicit `{ polling }`, which DEV-2769 / #13364 had already made a lint error before #13404 merged. `develop` is red on that check right now, and every branch merging it inherits the failure. It is fixed here in its own commit (`ddd85ad81d`), using the shared `awaitBundle()` helper that `tests/AGENTS.md` prescribes rather than an inline interval. **Happy to drop that commit if you would rather land it straight on develop** — it is unrelated to this PR's subject and is only here because CI Gate cannot go green without it.

### How has this been tested?

Three new Playwright E2E cases in `tests/e2e/editor-trimmed-row.spec.ts`, with their page-object methods. Each was verified to fail on the unfixed code before being made to pass.

```bash

# all six theme x bundle legs
cd tests && npx playwright test e2e/editor-trimmed-row.spec.ts

# determinism check, each new case
cd tests && npx playwright test e2e/editor-trimmed-row.spec.ts --grep "vetoed alter" --repeat-each=15
cd tests && npx playwright test e2e/editor-trimmed-row.spec.ts --grep "strands the editor" --repeat-each=15

npm --prefix handsontable run test:e2e -- --testPathPattern=alter
npm --prefix handsontable run test:e2e -- --testPathPattern=undoRedo
```

Results:

- `editor-trimmed-row.spec.ts`, all six legs: **264 passed**
- the two `alter()` cases, `--repeat-each=15` across all six legs: **180 passed** — the veto one fails **3/3** on unmodified `develop`, the throw one fails **3/3** with the per-exit resumes in place
- `discards when the filter lands in the same task as the swap`, same: **270 passed** (that grep also picks up the sibling removal case)
- Legacy Jasmine E2E, the suites that drive `alter()` hardest: `--testPathPattern=alter` **79 specs, 0 failures**; `--testPathPattern=undoRedo` (it catches around its own `alter()` calls) **201 specs, 0 failures**
- Core unit suite: **4592 passed / 4592**, 356 suites
- `eslint` clean on every changed file

A full `npm --prefix handsontable run build` is required before the cross-leg run — a partial build is exactly what produced DEV-2756's phantom flake.

### Types of changes

### Related issue(s):

1. DEV-2831 — the `alter()` scope leak this PR fixes.
2. DEV-2756 — where it was found; its own report is a measurement artifact, and the regression pin for its invariant is in this PR.
3. DEV-2739 — the append this leak reopens, and where the scope pair landed.

### Affected project(s):

- [x] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2831

---

The first iteration of this PR added a hand-written resume at each of the two `return`s and a comment claiming the timer already covered a throw. A `/code-review` pass caught that the timer fires a macrotask too late to cover anything, so the throw exits leaked as well. That is verified now (`hot.alter('no_such_action')` inside a `try`/`catch` appends 3/3) and pinned by the third test.

---

> [!NOTE]
> **Medium Risk**
> Touches core `alter()`, `updateData()`, and editor reconciliation on structural changes; behavior change is intentional bug-fix with E2E pins, but mistakes here affect editing commits and selection lifecycle.
> 
> **Overview**
> Fixes a bug where **vetoed or failed structural/data changes** could leave the editor’s “strand-discard” suspension open for the rest of the synchronous task, so a later filter/trim could **commit a stranded editor into wrong rows** or, separately, leave selection stuck on `updateData` and **block editors from opening**.
> 
> **`alter()`** now wraps its body in `try`/`finally` so `resumeStrandDiscards()` always runs on `beforeAlter` veto, `maxRows` early return, and thrown errors—not only on the happy path. **`updateData()`** uses the same pattern for the strand scope, and nests `try`/`finally` so `selection.markEndSource()` still runs if `selection.refresh()` throws; **`loadData()`** gets the same `markEndSource` pairing.
> 
> **`editorManager.suspendStrandDiscards()`** drops the zero-delay self-heal timeout (it ran in the next task, too late for same-task `Filters#filter()`), and docs stress that callers must pair suspend with `finally`.
> 
> Adds Playwright coverage for same-task `updateData` + filter, vetoed/throwing `alter` before filter, and throwing refresh after data swap; extends the editor-trimmed-row page object. **Test docs** warn that partial UMD builds leave `-min` legs on stale bundles (misread as flakes).
> 
> Changelog entry for issue **13425**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a74a1ede3c68e01390f178b443c2b221d06d6cb1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2739 Editor strand window closes on setTimeout(0): a same-task trim/filter commits through a stranded editor and appends rows](https://app.clickup.com/t/9015210959/DEV-2739)
Task linked: [DEV-2756 updateData under an open editor: a same-task filter still appends records under load](https://app.clickup.com/t/9015210959/DEV-2756)
Task linked: [DEV-2831 alter() leaks its structural-change scope on an early return, so a same-task filter appends records](https://app.clickup.com/t/9015210959/DEV-2831)

qunabu:
Task linked: [DEV-2812 AutoRowSize#clearCache() leaves rows unmeasured, so row headers drift out of sync](https://app.clickup.com/t/9015210959/DEV-2812)

budnix:
The two visual diffs look unrelated to this PR.

Both are the same screenshot in two browsers: `cross-browser/{chromium,firefox}/selection-arabic-rtl-demo-2.png`.

- **The code here cannot run on that page.** `arabic-rtl-demo` never calls `alter()` or `updateData()`, and those are the only two callers of `suspendStrandDiscards()`. No editor is opened either, which the discard branch also requires. So the scope this PR touches stays closed for the whole test.
- **The identical shipped code passed this check earlier.** Visual / Compare was green on ddd85ad81d, and the diff of `handsontable/src/core.ts` and `handsontable/src/editorManager.ts` between that commit and HEAD is empty. Everything that landed since is docs plus one test-only page object.

The diff itself is geometry: the last row is clipped in `actual` but complete in `expected`, and the row-header strip loses its selection highlight. Both browsers moved together, which fits a sub-pixel row-height shift on a layout sitting right on the clip boundary rather than a per-browser flake.

Applying `visual-approved`.

handsontable/src/core.ts
```diff
@@ -3699,15 +3707,20 @@ export default function Core(
...
+        // `finally`, for the same reason `alter()` uses one: `adjustRowsAndCols()` runs the row
+        // creation hooks and `selection.refresh()` the selection hooks, so a hook that throws
+        // would otherwise leave the scope open for the rest of the task (DEV-2831).
+        try {
+          instance.columnIndexMapper.fitToLength(this.getInitialColumnCount());
+          instance.rowIndexMapper.fitToLength(this.countSourceRows());
 
-        editorManager.resumeStrandDiscards();
+          grid.adjustRowsAndCols();
+          selection.markSource('updateData');
```
Haxikowy, line 3718: `selection.markEndSource()` never runs if `adjustRowsAndCols()` or `selection.refresh()` throws, so `#selectionSource` stays stuck at `'updateData'`. Same failure mode as the strand-discard leak this PR fixes, one line above it: worth wrapping in try/finally too.

tests/e2e/editor-trimmed-row.spec.ts
```diff
@@ -1227,13 +1227,124 @@ test.describe('an index-map change nested inside a removal', () => {
...
+
+      await grid.updateDataThenFilterSameTask(
+        [['X0', 'Y0'], ['X1', 'Y1'], ['X2', 'Y2']], 0, ['X2']
+      );
+
+      // The filter's trimming map really was written, so the discard was reached. Without this the
+      // assertions below are all satisfied by "no editor was ever open" - the premise pinned above
+      // rules that out going in, and this rules out the trigger never firing.
+      expect(await grid.sawTrimmingCacheUpdate('row')).toBe(true);
+      // Deliberately NOT asserting the editor reference is gone. Whether a fresh editor is
```
Haxikowy, line 1265: The umd/full-min `getActiveEditor()` divergence after a discard (`null` vs a fresh editor at row 0) is a real behavioral inconsistency, not just a test-scoping detail. Worth a follow-up ticket rather than leaving it as an unasserted comment.

Haxikowy (commented):
Commit `ddd85ad81d` fixes a lint error that originated in DEV-2769/#13364, not DEV-2831. The commit message attributes it to this ticket; worth a one-line amend noting it's an unrelated CI fix. Not blocking.

handsontable/src/core.ts
budnix: Nice catch — paired it with a `finally` in a74a1ede3c, and gave `loadData()` the same treatment. One detail: `adjustRowsAndCols()` runs before `markSource()`, so only a `refresh()` throw can strand the source, and when it does `afterSetRangeEnd` then skips the scroll, the editor close and `prepareEditor()` for good — the grid stops taking typing, so this was worth more than a label fix.

tests/e2e/editor-trimmed-row.spec.ts
budnix: Good spot — filed [DEV-2862](https://app.clickup.com/t/123kvxea2pf) and pointed the comment at it in a74a1ede3c. You were right that it is real rather than test scoping: the fixture enables no Formulas, so the settings are identical and only the bundle differs.

Haxikowy (approved):

PR #13426, opened 2026-09-08
DEV-2832: Redirect the legacy docs.handsontable.com host to handsontable.com

budnix:
### Context

The PR fixes `docs.handsontable.com` serving the whole documentation tree with **HTTP 200** instead of redirecting to `handsontable.com`, so the docs were live on two hostnames at once. Every path was affected, not one page.

Both hosts return an identical `etag` (`26470690e7c996927c451360912ebba1`), so the same Cloudflare Pages project (`handsontable-docs`) backs both. `docs.handsontable.com` is a legacy custom domain on that project, left from before the move to `handsontable.com/docs`, and it was not listed in the domain table in `docs/README-DEPLOYMENT.md`.

**Cause.** All docs routing lives in `docs/cloudflare/_worker.js`, and once a `_worker.js` exists Cloudflare Pages ignores `_redirects`, so that file is the only thing deciding routing. Two facts combined:

1. **There was no host rule at all.** Every rule matches on the path; nothing read `url.hostname`, so nothing ever moved a request between hostnames.
2. **Every redirect was built from the hostname the request arrived on:**

   ```js
   function abs(path, base) {
     return `${base.origin}${path}`;
   }
   ```

So a legacy URL either answered 200 outright, or redirected to another legacy URL. There was no way out — which is why `docs.handsontable.com/` 301'd to `docs.handsontable.com/docs`.

### The fix, and why it sits where it does

Two changes:

1. **`abs()` swaps in the canonical origin for a legacy host**, so every path rule above the cut doubles as a one-hop cross-host redirect that resolves what the path actually *means*.
2. **Rule 12a**: on a legacy host, `GET`/`HEAD` gets a `301` to the same path on `handsontable.com`.

Rule 12a's position is load-bearing in **both** directions, which is the whole design:

**It cannot go earlier.** The worker does not serve the whole main domain — only `/docs/*` on `handsontable.com` reaches it, and everything else is the marketing site. Measured:

| Path | On docs host | On main host |
|---|---|---|
| `/` | 301 → `/docs` | 200 (marketing homepage) |
| `/0.8.0/` | 301 → changelog | **404** |
| `/demo/foo` | 301 → `/demo` | **404** |
| `/customers/foo` | 301 → `/customers/` | **404** |

A host-first collapse would have dumped `docs.handsontable.com/` on the marketing homepage and `/0.8.0/` on a 404 — the most common legacy bookmark shapes. Rules 2/2a/2d/2e stay ahead of the cut so those resolve to their real destinations.

**It cannot go later.** This is what the review on the first commit caught, and it mattered. Rules 13–17 answer with a **302**, because their destination depends on the `docs_fw` cookie and must not be cached as permanent. That is correct on the canonical host and useless on a legacy one: a 302 is not a canonicalisation signal, so those URLs never leave the index. With the collapse done only at the end of the pipeline, the most commonly linked legacy URLs took that path:

```
BEFORE (first commit)                    AFTER (rule 12a)
/docs                 302  ->            /docs                 301
/docs/                302  ->            /docs/                301
/docs/14.4            302  ->            /docs/14.4            301
/docs/installation    302  ->            /docs/installation     301
/docs/binding-to-data 302  ->            /docs/binding-to-data  301
```

Only deep content pages got the 301. Those same rules also read `docs_fw` from the **legacy host's** cookie jar, and cookies are host-scoped — so a preference saved on `handsontable.com` was invisible and a React reader silently landed on the JavaScript page. Both problems have one cause and one fix: stop deciding anything cookie-shaped on the legacy host. Hand the bare path over with a 301 and let the canonical host, which owns the cookie, choose the framework.

Placing the cut above the asset fallback also removes the discarded `env.ASSETS.fetch()` subrequest every unmatched legacy hit used to pay, and the 500 an unhealthy asset binding would have turned the 301 into. The end-of-pipeline catch-all is now unreachable for `GET`/`HEAD` and is **deleted**.

`GET`/`HEAD` only, so the rule 18a POST mock still answers on that host: a 301 is downgraded to GET by every client and loses the body, and a POST is never indexed.

Rule 18 sits below the cut and so takes two 301 hops (`→ handsontable.com/docs/14.4/react-installation → …/react-data-grid/installation/`). That is cheaper than reordering it above rule 13, which would change which rule wins for a `.html` path.

**Also from the review:** `isLegacyDocsHost()` now strips a trailing dot — the URL parser lowercases a host but preserves it, so `docs.handsontable.com.` bypassed the exact-match `Set` and served 200. And `loadWorker()` uses `replaceAll`: the `__LATEST_DOCS_VERSION__` placeholder also appears in the comment above the declaration, so a single `replace()` substituted the prose and left **rule 1b untestable**. Production uses `sed … /g`, so it was never affected.

**Drive-by:** the header rule list was stale — two rule 3s, and 13–19 did not match the code. `docs/cloudflare/README.md` declares that list authoritative, and the drift produced a wrong finding in the review itself. It is re-synced against the `// -- N.` markers and now states where a content-serving rule must go.

### Why the host list is an exact-match allowlist

`LEGACY_DOCS_HOSTS` is a `Set` of exact hostnames, deliberately not "any host that is not handsontable.com". The same worker serves staging and every PR preview from `handsontable-docs-staging.pages.dev` and its per-branch subdomains, so a negative match would 301 all of them into production and break docs review with no visible error. Verified against this PR's own live preview.

### Also repaired along the way

`/demo/*` and `/customers/*` were dead-ending at **404** on the docs host, because those rules were written for the main host. And `docs.handsontable.com/robots.txt` returned 404, so the main site's robots directives never applied to the subdomain at all.

### On the SEO framing

Worth being precise so this is not oversold. The docs host already emitted a correct canonical tag pointing at `handsontable.com/docs/...`, and its sitemap already listed only `handsontable.com` URLs. But a canonical tag is a *hint*, not a directive: the duplicate host answered 200, so crawlers could keep both URLs indexed and spend crawl budget on each copy. The 301 makes it unambiguous and consolidates inbound link equity. This is a crawl-hygiene and link-consolidation fix, not a live ranking split.

**Deploy note:** this only changes live behavior on the next **production** docs deploy (`docs-production.yml`, from `prod-docs/*`). Merging to `develop` alone will not move `docs.handsontable.com`.

No changelog entry: the change is `docs/`-only, and the changelog gate passes those automatically.

Docs-site change — no `handsontable/src/**` or `wrappers/**` files touched, so the unit/E2E/type tiers do not apply. The worker's own suite is the relevant coverage, and per `docs/cloudflare/README.md` it is the **only** automated coverage for this file.

- Unit tests added/modified (`*.unit.js`): none — this file is covered by `docs/cloudflare/__tests__/_worker.test.mjs` (`node:test`), which goes from 16 to **33** cases
- E2E tests added/modified (Playwright `tests/e2e/*.spec.ts`): none — the worker is not reachable from the grid E2E harness
- For a bug fix — the specs that fail without this fix: `the legacy docs host 301s a content page to the same path on the canonical host` for the original bug, and `the cookie-reading rules 13-17 answer the legacy host with a 301, not their usual 302` for the review finding
- Demo page / recorded trace (for UI changes): n/a — no UI change

A `requestOn(host, path)` helper was added because the existing `request()` helper hardcodes `https://handsontable.com` and therefore cannot exercise host behavior at all.

Coverage includes: content-path collapse, query-string preservation, one-hop path resolution, the four non-`/docs` legacy paths, the 302-family paths now answering 301, cookie-independence on the legacy host, rule 1b, rule 18's two-hop chain, trailing-dot and uppercase hostnames, the POST mock surviving the method guard, the external HyperFormula redirect staying external, the canonical host being untouched, and five staging/preview/apex hosts serving their own content with no redirect.

**Verified the tests catch the real regressions, three ways:**

- Removing the `docs.handsontable.com` entry from `LEGACY_DOCS_HOSTS` reddens **12** cases.
- **Moving rule 12a below the cookie rules** — i.e. reintroducing exactly the bug the review found — reddens **3** cases: the 302-family test, the cookie-independence test, and the rule 18 test.
- The 304 test is kept as a **placement sentinel**: rule 12a now returns before `ASSETS` is consulted, so move it below the fallback and the 304 escapes (it carries no `Location`) and the test goes red.

Live behavior was confirmed directly rather than assumed. The 304:

```
curl -sSI -H 'If-None-Match: "26470690e7c996927c451360912ebba1"' \
  https://docs.handsontable.com/docs/angular-data-grid/row-parent-child/
HTTP/2 304
```

And this PR's own preview deployment, which is the case the allowlist exists to protect:

```
pr-13426.handsontable-docs-staging.pages.dev/docs/angular-data-grid/row-parent-child/  200, no redirect
pr-13426.handsontable-docs-staging.pages.dev/docs/next/javascript-data-grid/           301 -> same preview host
pr-13426.handsontable-docs-staging.pages.dev/                                          301 -> same preview host /docs
```

### Commands run

```
$ cd docs && node --test cloudflare/__tests__/_worker.test.mjs
ℹ tests 33
ℹ pass 33
ℹ fail 0

$ npx eslint docs/cloudflare/_worker.js docs/cloudflare/__tests__/_worker.test.mjs
(no output — clean)

$ node --check docs/cloudflare/_worker.js
(clean)
```

CI's `Docs / plugins` job runs the same suite with dependencies installed and passed on the first commit.

### Types of changes

### Related issue(s):

1. DEV-2832

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

None of the above — the change is confined to the documentation site (`docs/cloudflare/_worker.js`, its test, `docs/cloudflare/README.md`, and `docs/README-DEPLOYMENT.md`), which is not one of the listed packages.

### Checklist:

- [x] My change requires a change to the documentation. — done in this PR. `docs/README-DEPLOYMENT.md` documents `docs.handsontable.com` as a legacy custom domain that 301s on `GET`/`HEAD`, which was missing from the domain table; `docs/cloudflare/README.md` gains the legacy-host collapse in its "concepts worth knowing" list, beside the framework cookie and `LATEST_VERSION`.

### Out of scope — flagged for a decision

1. **Two `*.pages.dev` hosts also serve the docs at 200 with no `noindex`:** `handsontable-docs.pages.dev` (production apex) and `handsontable-docs-staging.pages.dev` (staging apex). Both carry the correct canonical, neither is blocked from indexing, and staging is the worse of the two since unreleased content is crawlable. Left alone deliberately — the production apex is the documented way to verify a production deploy, and redirecting the staging apex would break docs review. The right fix is a `noindex` header or a `robots.txt` on `*.pages.dev` hosts, which is a separate change.
2. **`/i18n/*` still redirects to a marketing 404.** `handsontable/src/i18n/utils.ts:79` hardcodes `https://docs.handsontable.com/i18n/missing-language-code` in a console warning, and that URL **already 404s today** — a dead Netlify-era path, so pre-existing rather than a regression here. Giving it a real destination is its own decision, and it needs the string in `i18n/utils.ts` updated too.
3. **The CSP `script-src` still allowlists `https://docs.handsontable.com`**, which is now unreachable by construction. Left in place on purpose: dropping it would block that load on staging and preview hosts, where neither `'self'` nor the list covers `handsontable.com`.

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2832

---

> [!NOTE]
> **Medium Risk**
> Changes production docs routing and SEO canonicalisation for a live custom domain; mis-placing rule 12a or broadening the host allowlist could break previews or leave duplicate/302 legacy URLs indexed.
> 
> **Overview**
> Stops **`docs.handsontable.com`** from serving duplicate docs with **200** by canonicalising traffic to **`handsontable.com`** in the Cloudflare Pages worker (`docs/cloudflare/_worker.js`).
> 
> **`abs()`** now uses the canonical origin for allowlisted legacy hosts so path rules above the cut can **301 in one hop** with the right destination. **Rule 12a** issues a **GET/HEAD 301** to the same path on the canonical host for anything that did not match earlier rules, placed **before** cookie-based **302** rules so legacy URLs are not left indexable and framework choice stays on the host that owns the **`docs_fw`** cookie. **`LEGACY_DOCS_HOSTS`** is an exact-match list (staging/PR previews are excluded).
> 
> Redirects on the legacy host still get **HSTS**; document-only security headers are not applied to redirects. Worker tests grow with a **`requestOn(host)`** helper, **`replaceAll`** for the latest-version placeholder in tests, and deployment docs describe the legacy domain behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 89515ff23febbff65adc0d93e963ac282eda95ce. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


qunabu:
Task linked: [DEV-2832 Redirect docs.handsontable.com to handsontable.com with a 301](https://app.clickup.com/t/9015210959/DEV-2832)

docs/cloudflare/__tests__/_worker.test.mjs
```diff
@@ -33,6 +38,23 @@ function request(path, cookie, method = 'GET') {
...
+/**
+ * Same shape as `request()`, but on an arbitrary hostname. The host-collapse
+ * rule is the only rule that reads `url.hostname`, so it is the only one that
+ * cannot be exercised through the handsontable.com-pinned helper above.
+ *
+ * @param {string} host
+ * @param {string} path
+ * @returns {object} A minimal Request stand-in.
+ */
+function requestOn(host, path) {
```
demtario: **Low** — `requestOn(host, path)` hardcodes `method: 'GET'`, so no test exercises rule 12a's HEAD branch on the legacy host. The GET-or-HEAD predicate is the crux of the rule's method restriction, but only half of it is covered. Fix: add an optional `method` param to `requestOn()` (mirroring `request()`), and one assertion that a HEAD request to `docs.handsontable.com` also collapses to a 301.

demtario (commented):
1 inline note below.

Off-diff, not blocking:
- `docs/cloudflare/_worker.js:831-843` (`withSecurityHeaders`) passes 3xx through untouched. Every GET/HEAD on `docs.handsontable.com` is now a 3xx post-merge, so `Strict-Transport-Security` stops being emitted there and the existing HSTS pin (max-age 1yr) would age out with no renewal. Worth a one-line check on whether DNS/apex config sets HSTS independently for that host.
- `docs/cloudflare/_worker.js:819` (`CONTENT_SECURITY_POLICY`) still lists `https://docs.handsontable.com` in `script-src`. Pre-existing line, now dead weight since that host never serves 200 again. Optional cleanup.

docs/cloudflare/__tests__/_worker.test.mjs
budnix: Nice catch — requestOn() takes a method now and there is a HEAD case in 89515ff23f; narrowing the rule to GET reddens it, so that half is covered. Both off-diff notes are fixed there too: redirects now carry HSTS (the apex does set includeSubDomains, but this stops the host leaning on that), and the dead docs.handsontable.com CSP entry is gone — my earlier reason for keeping it was simply wrong.

demtario (approved):

PR #12195, opened 2026-03-20
PRO-1063: Migrate docs from VuePress 1 to Astro Starlight

qunabu:
## Summary

- Replaces VuePress 1 with Astro 5 + Starlight as the documentation framework
- Ports all guides and content from VuePress markdown format to Astro-compatible markdown
- Adds Starlight component overrides to match the existing docs UI: breadcrumbs, breadcrumbs with "Copy as Markdown", header with versions dropdown, support dropdown, Headway widget, GitHub stars counter, page footer with social links and live status dot, InfoBox "What's new" banner, and sidebar `menuTag` badges (`new`/`updated`)
- Adds interactive example UX: loading shimmer, source code toggle, "Edit on StackBlitz" button, "See on GitHub" link
- Fixes `Handsontable.dom.addClass/removeClass` undefined errors in demo examples by replacing with native `classList` calls
- Removes all `docs/.vuepress/` files (322 files deleted)
- Updates `docs/.eslintrc.js` to a standalone config without VuePress/Vue parser dependencies

## Test plan

- [ ] Run `npm run dev` in `docs/` — server starts without errors
- [ ] Visit a guide page — breadcrumbs appear above the `h1`, "Copy as Markdown" copies valid markdown to clipboard
- [ ] Visit a page with a live example — loading shimmer shows while example loads, "Source code" toggle reveals code, "Edit on StackBlitz" opens the example in StackBlitz, "See on GitHub" links to the correct file
- [ ] Visit the demo page for React (`/docs/react-data-grid/demo/`) — no `classList` / `dom.removeClass` errors in console
- [ ] Verify sidebar shows "New" (green) and "Updated" (amber) badges on pages that have `menuTag` frontmatter
- [ ] Check the InfoBox "What's new" banner appears above the sidebar, and closes/hides on subsequent visits
- [ ] Verify versions dropdown, support dropdown, Headway bell, and GitHub stars badge render in the header
- [ ] Run `npm run docs:lint` in `docs/` — passes with no errors

[skip changelog]

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it replaces the docs build/deploy pipeline (Astro build + Netlify packaging) and removes large amounts of VuePress-specific tooling, which could break documentation previews, linting, or deployment paths if any assumptions remain.
> 
> **Overview**
[...]

PR #13427, opened 2026-09-08
DEV-2833: Fix "undefined" prefix in the Open in AI prompt links

budnix:
### Context

The PR fixes the **Open in ChatGPT** and **Open in Claude** buttons in the docs right sidebar, which open the AI vendor with the literal word `undefined` in front of the page URL. It affects every docs page.

Reproduced on the live site with a plain `curl`:

```bash
curl -sL "https://handsontable.com/docs/angular-data-grid/row-parent-child/" \
  | grep -o 'https://chatgpt.com/?q=[^"]*'
# https://chatgpt.com/?q=undefined%20https%3A%2F%2Fhandsontable.com%2Fdocs%2Fangular-data-grid%2Frow-parent-child%2F
```

So the AI chat opens prefilled with:

```
undefined https://handsontable.com/docs/angular-data-grid/row-parent-child/
```

**Cause.** `docs/src/components/PageSidebar.astro` builds the prompt from the `starlight-page-actions` config:

```js
const prompt = userPrompt?.includes('{url}')
  ? userPrompt?.replace('{url}', currentUrl)
  : `${userPrompt} ${currentUrl}`;
```

`starlightPageActions()` is called in `docs/astro.config.mjs` with no options, and the plugin's `defaultConfig` has no `prompt` key — only `actions` and `share`. `actions.chatgpt` and `actions.claude` both default to `true`, so the buttons render, but `userPrompt` is `undefined`. The optional chain makes the ternary take the `else` branch, and the template literal stringifies `undefined` straight into the query string.

The plugin's own `overrides/PageTitle.astro` has the guard this component is missing:

```js
const promptTemplate = userPrompt ?? Astro.locals.t("prompt.default");
```

`PageSidebar.astro` re-implements that block but dropped that one line. The docs site overrides both `PageTitle` and `PageSidebar`, so the plugin's guarded version never renders here. Present since the VuePress → Astro Starlight migration (#12195).

**Fix.** Restore the fallback. With no `prompt` configured, the links now carry the plugin's injected translation:

```
Read https://handsontable.com/docs/angular-data-grid/row-parent-child/. I want to ask questions about it.
```

The `{url}` placeholder and the append-the-URL branch both keep working, so configuring a `prompt` in `astro.config.mjs` later needs no further change.

### How has this been tested?

New regression test at `docs/src/components/__tests__/page-sidebar-ai-prompt.test.mjs`. It extracts the component's prompt-building statements from the `.astro` source and runs them against a stubbed config, so it asserts behavior rather than matching source text. It covers all three branches: no configured prompt, a prompt with `{url}`, and a prompt without `{url}`.

Verified it actually catches this bug — reverted the component to the old code and the first case went red with the exact reported string:

```
AssertionError: The input was expected to not match /undefined/. Input:
'undefined https://handsontable.com/docs/angular-data-grid/row-parent-child/'
```

```bash
npm run docs:test:plugins --prefix docs   # 300/300 pass (296 before, +4 new)
npm run docs:lint --prefix docs           # clean
```

Also confirmed against a real render, not just the diff — `astro dev` plus a `curl` of the same page:

```bash
curl -s "http://localhost:4321/docs/angular-data-grid/row-parent-child/" \
  | grep -o 'https://claude.ai/new?q=[^"]*'
# https://claude.ai/new?q=Read%20http%3A%2F%2Flocalhost%3A4321%2Fdocs%2Fangular-data-grid%2Frow-parent-child%2F.%20I%20want%20to%20ask%20questions%20about%20it.
```

That render step mattered: it is the only way to confirm `Astro.locals.t('prompt.default')` resolves to the English string rather than returning the key itself.

### Second commit: `docs/AGENTS.md`

Verifying this against a real render in a fresh worktree hit two dev-server traps that cost more time than the fix did, so they are written down in `docs/AGENTS.md` §2.12:

- `npm run dev` always dies with `Dev server failed to start within 30s.` in an agent session. Astro 7 detects the agent, force-enables background mode, and SIGTERMs the child it spawned when this config does not boot inside its hardcoded 30s. `ASTRO_DEV_BACKGROUND=1 npx astro dev` runs the real server in the foreground.
- A fresh worktree 500s on every docs page with `[postcss] ENOENT ... handsontable/styles/handsontable.min.css`, because `git worktree` materializes tracked files only and that stylesheet is a build output. The postcss cache holds the failure, so the server needs a restart after the file appears.

### Note for the reviewer

Production docs build from `prod-docs/<major>.<minor>`, which cherry-picks from `develop` selectively. This fix does not reach handsontable.com/docs until it is cherry-picked there.

Optional follow-up, deliberately not done here: the prompt points the AI at the HTML page. The docs already serve clean Markdown for the same page (`/docs/_md/<slug>.md` returns `200 text/markdown`, and the "Copy Markdown" button uses it). Pointing the AI links at that instead would give the model the page content without the nav and grid markup. That is a product decision, so it is left out of a bug fix.

### Types of changes

### Related issue(s):

1. DEV-2833

### Affected project(s):

- [ ] `handsontable`
- [ ] `@handsontable/vue3`

### Checklist:

ClickUp task: https://app.clickup.com/t/9015210959/DEV-2833

### DIFF
--- a/docs/AGENTS.md
+++ b/docs/AGENTS.md
@@ -564,6 +564,20 @@ The custom loader (`src/plugins/framework-loader.mjs`) renders every source page
 - `npm run dev` must work with the default Node heap. Do not add `NODE_OPTIONS=--max-old-space-size` workarounds; shrink the data store instead.
 - Plugin unit tests run with `node --test src/plugins/__tests__/*.test.mjs`.
 
+### Starting the dev server from an agent session
+
+`npm run dev` dies with `Dev server failed to start within 30s.` and exit 1 in every agent session, and the message is a lie - nothing in the config is broken. Astro 7's `astro dev` calls `isRunByAgent()`, and on a hit it force-enables **background mode**: it spawns a detached child, polls for that child's lock file for 30 seconds, then SIGTERMs the child and exits non-zero. This docs config never boots that fast, so the wrapper always kills a server that was on its way up. Run the server process itself instead:
+
+```bash
+ASTRO_DEV_BACKGROUND=1 npx astro dev
+```
+
+That env var is what the spawned child receives, so setting it makes the foreground process *be* the server and skips the timeout wrapper entirely. Do not raise the timeout (it is hardcoded in `astro/dist/cli/dev/background.js`) and do not conclude the config is at fault.
+
+### The dev server 500s in a fresh worktree until the core CSS exists
+
+Every page returns HTTP 500 with `[postcss] ENOENT: no such file or directory, open '../../../handsontable/styles/handsontable.min.css'`. `src/styles/handsontable-import.css` imports the **built** stylesheet from the core package, and `git worktree` materializes tracked files only, so a new worktree has no `handsontable/styles/`. Build the core package, or copy `handsontable/styles/` in from a checkout that already has it. Then **restart the dev server** - postcss caches the resolution failure, so a running server keeps 500ing after the file appears.
+
 ---
 
 ## 2.13 Example-Runner Error Handling
--- a/docs/src/components/PageSidebar.astro
+++ b/docs/src/components/PageSidebar.astro
@@ -20,10 +20,14 @@ const { actions, prompt: userPrompt } = config;
 const { pageActions } = (Astro.locals as any).starlightRoute.entry.data;
 const showActions = pageActions || typeof pageActions === 'undefined';
 
-const prompt = userPrompt?.includes('{url}')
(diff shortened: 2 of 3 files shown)


qunabu:
Task linked: [DEV-2833 Docs: "Open in ChatGPT" and "Open in Claude" prefill the prompt with "undefined"](https://app.clickup.com/t/9015210959/DEV-2833)

qunabu (approved):

demtario (approved):

